PR #11526, opened 2018-09-14
Put 'static' first in declarations

nsajko:


libr/crypto/crypto.c
```diff
@@ -5,8 +5,9 @@
 
 R_LIB_VERSION (r_crypto);
 
-struct { const char *name; ut64 bit; }
-static const crypto_name_bytes[] = {
+static const struct {
+	const char *name; ut64 bit;
```
radare: move ut64 bit after name, one field per line

PR #11527, opened 2018-09-14
libr/util/buf.c: Fix related to unsigned numbers

nsajko:
Subtraction of unsigned numbers always yields a positive number, thus
R_MAX (0, b->length - b->offset) is always b->length - b->offset,
instead of the intended.

### DIFF
--- a/libr/util/buf.c
+++ b/libr/util/buf.c
@@ -205,16 +205,23 @@ R_API const ut8 *r_buf_buffer (RBuffer *b) {
 	return (b && !b->sparse)? b->buf: NULL;
 }
 
+static ut64 remainingBytes(ut64 limit, ut64 length, ut64 offset) {
+	if (offset >= length ) {
+		return 0;
+	}
+	return R_MIN (limit, length - offset);
+}
+
 R_API ut64 r_buf_size (RBuffer *b) {
 	if (!b) {
 		return 0LL;
 	}
 	if (b->iob) {
 		RIOBind *iob = b->iob;
-		return R_MIN (b->limit, R_MAX (0, iob->fd_size (iob->io, b->fd) - b->offset));
+		return remainingBytes (b->limit, iob->fd_size (iob->io, b->fd), b->offset);
 	}
 	if (b->fd != -1) {
-		return R_MIN (b->limit, R_MAX (0, b->length - b->offset));
+		return remainingBytes (b->limit, b->length, b->offset);
 	}
 	if (b->sparse) {
 		ut64 max = 0LL;
@@ -223,7 +230,7 @@ R_API ut64 r_buf_size (RBuffer *b) {
 		}
 		return 0LL;
 	}
-	return b->empty? 0: R_MIN (b->limit, R_MAX (0, b->length - b->offset));
+	return b->empty? 0: remainingBytes (b->limit, b->length, b->offset);
 }
 
 // rename to new?


libr/util/buf.c
```diff
@@ -223,7 +234,7 @@ R_API ut64 r_buf_size (RBuffer *b) {
 		}
 		return 0LL;
 	}
-	return b->empty? 0: R_MIN (b->limit, R_MAX (0, b->length - b->offset));
+	return b->empty? 0: minimumZeroMaximum (b->limit, b->length, b->offset);
```
MaskRay: Can you just inline it `R_MIN (b->limit, b->offset < b->length ? b->length-b->offset : 0)` instead of inventing a new function?

nsajko: But that would mean code triplication, because I would need to inline it at three places.

Also, I hate the ternary operator.

libr/util/buf.c
```diff
@@ -205,16 +205,27 @@ R_API const ut8 *r_buf_buffer (RBuffer *b) {
...
+	ut64 d;
+	if (length <= offset) {
+		return 0;
+	}
+	d = length - offset;
+	if (limit <= d) {
+		return limit;
+	}
+	return d;
+}
```
radare: empty line between functions

libr/util/buf.c
```diff
@@ -205,16 +205,27 @@ R_API const ut8 *r_buf_buffer (RBuffer *b) {
 	return (b && !b->sparse)? b->buf: NULL;
 }
 
+static ut64 minimumZeroMaximum(ut64 limit, ut64 length, ut64 offset) {
```
radare: confusing function name

libr/util/buf.c
radare: having a separate function makes the logic more clear and avoids duplication of inlined code. (the compiler can decide). But i dont think this change makes the code more readable because the function name is a bit confusing

libr/util/buf.c
nsajko: Do you maybe have a suggestion for a better name?

libr/util/buf.c
```diff
@@ -205,16 +205,28 @@ R_API const ut8 *r_buf_buffer (RBuffer *b) {
...
+	ut64 d;
+	if (length <= offset) {
```
radare: maybe offset >= length is more readable.

libr/util/buf.c
```diff
@@ -205,16 +205,28 @@ R_API const ut8 *r_buf_buffer (RBuffer *b) {
...
+		return 0;
+	}
+	d = length - offset;
```
radare: define and assign ut64 d = ... here

libr/util/buf.c
```diff
@@ -205,16 +205,28 @@ R_API const ut8 *r_buf_buffer (RBuffer *b) {
...
+	if (limit <= d) {
+		return limit;
+	}
+	return d;
```
radare: What about just this? `return R_MIN(d, limit);`

libr/util/buf.c
MaskRay: Alternatively, you can add a check `if (b->offset >= b->length) return 0; ` then you don't need the helper function at all because the logic is simple enough and can be inlined.

nsajko: > Alternatively, you can add a check if (b->offset >= b->length) return 0; then you don't need the helper function at all because the logic is simple enough and can be inlined.

That would still mean code duplication.

PR #11528, opened 2018-09-14
libr/util/uleb128.c: Fix left shift of negative value

nsajko:
Those are formally an error (undefined behavior).


nsajko:
Correction: shifts of **negative** values are undefined behavior.

MaskRay:
Let me change it further: when RHS is within bounds, left shifts of negative values are undefined, right shifts are implementation defined, in C and C++.

PR #11530, opened 2018-09-14
libr/core/cmd.c: Fix related to unsigned numbers

nsajko:
Subtraction of unsigned numbers always yields a positive number. Adds
casts to st64 to give the intended result.

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -2636,7 +2636,7 @@ repeat_arroba:
 			if (ptr[2] == '.') { // "@.."
 				if (ptr[3] == '.') { // "@..."
 					ut64 addr = r_num_tail (core->num, core->offset, ptr + 4);
-					r_core_block_size (core, R_ABS (addr - core->offset));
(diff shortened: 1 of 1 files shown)


libr/core/cmd.c
```diff
@@ -2636,7 +2636,7 @@ static int r_core_cmd_subst_i(RCore *core, char *cmd, char *colon, bool *tmpseek
 			if (ptr[2] == '.') { // "@.."
 				if (ptr[3] == '.') { // "@..."
 					ut64 addr = r_num_tail (core->num, core->offset, ptr + 4);
-					r_core_block_size (core, R_ABS (addr - core->offset));
+					r_core_block_size (core, R_ABS ((int)addr - (int)core->offset));
```
XVilka: It is wrong because you are reducing 64bit integer to 32bit integer on some systems. Please using 64bit integer here.

nsajko: It is OK, because int is the type of the bsize parameter anyway; so I am not causing much additional harm.

XVilka:
You do. (int)(x - y) != (int)x - (int)y

PR #11531, opened 2018-09-15
Add user defined field in RBreakpointItem

Wenzel:
This PR adds a `void*` field named `user` in `RBreakpointItem` struct in order for the user implementing a plugin to associate its own data to the breakpoint.

The use case I have is for `r2vmi` debugger plugin, i need to associate a `vmi_event_t` struct with each `breakpoint`.
My current implementation is to use a `GHashTable`, [associating a breakpoint virtual address](https://github.com/Wenzel/r2vmi/blob/master/io_vmi.h#L22) to the `vmi_event_t`, but i would rather directly the breakpoint struct if possible:
https://github.com/Wenzel/r2vmi/blob/master/debug_vmi.c#L543

What do you think ?
Thanks.

### DIFF
--- a/libr/include/r_bp.h
+++ b/libr/include/r_bp.h
@@ -60,6 +60,7 @@ typedef struct r_bp_item_t {
 	char *data;
 	char *cond; /* used for conditional breakpoints */
 	char *expr; /* to be used for named breakpoints (see r_debug_bp_update) */
+	void *user; /* user defined data associated with the breakpoint */
 } RBreakpointItem;
 
 struct r_bp_t;


radare:
Shouldn't hurt to hold that in there, but we will probably need mechanisms to free it when done. Also, are you using glib just for a hashtable? because in r2land we have hashtables too

Wenzel:
@radare yes i'm using `glib` as i was already familiar with interface and my project has a dependency on it.

I think freeing this field is the responsability of the user, not radare2.
You cannot take for granted that:
1. the user will put a pointer to a struct on the heap there (could be just to store a int between breakpoints)
2. maybe this struct will be freed by another fonction (which is my case for example, a `vmi_event_t` is being freed by calling [vmi_clear_event](https://github.com/libvmi/libvmi/blob/master/libvmi/events.h#L684)

radare:
This seems reasonable, but can be problematic because we can't hook when a breakpoint is removed. Can be also a problem if the user changes the debugger backend with the dL command.

Im fine for merging this because i guess it works for your case and shuoldnt be problematic for most use cases, but those are drawbacks i can imagine..

ret2libc:
Do you need a `user` different for each `RBreakpointItem`? 
Do you need a `user` different from the one in `RBreakpoint`?

Also, are you sure you can get a reference to the `RBreakpointItem->user` every time an Item is deleted? I guess you'd need it to free the referenced object. If that's not always possible, you probably need a `free_user` callback as well.

Wenzel:
@ret2libc 
> Do you need a user different for each RBreakpointItem?

Yes, because each one of them defines a specific breakpoints, therefore a VMI event associated with it

> Do you need a user different from the one in RBreakpoint?

I could have used that field, but i would have stored a hashtable in it, better use RBreakpointItem directly.

> Also, are you sure you can get a reference to the RBreakpointItem->user every time an Item is deleted? 


When do you delete an item ?
My use case is using it in my `__breakpoint` function, when a breakpoint is either set or unset.

> I guess you'd need it to free the referenced object

That's right.
The question is how to pass a callback to radare2, when should radare2 call it ?

Should we simply rely on the `breakpoint` function ?
Like when you are unsetting a breakpoint, you expect the user to free the `user` field associated with the `RBreakpointItem` ?

Thanks for looking into it.

radare:
i think that storing custom info inside the bpitems is a bad idea because bps are not linked to a single debugger backend, so it may result in bad behaviours and problems that you cant control without lot of work on top of it, so its just easier to keep track of that extra data in the debugger backend when the breakpoint is hit imho.

thats why i closed the issue, but maybe you have a better proposal or solution

radare:
Yeah this brings many complicated scenaarios we probably dont want to go in. Having a hashtable is probably the cleanest way to do this.

Issue #11399, opened 2018-09-05
Use-after-free in symbols()


fumfel:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Ubuntu 18.04 x64
| File format of the file you reverse (mandatory)      | WASM
| Architecture/bits of the file (mandatory)            | N/A
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.0.0-git 19358 @ linux-x86-64 git.2.9.0-9-g38d9170 commit: 38d9170f5da82ece05d1b90e366763999e633e34 build: 2018-09-05__15:07:23

### Expected behavior

Disassembly of file or error message.

### Actual behavior

Use-after-free in ASAN build.

### Steps to reproduce the behavior 
- Download file: [r2_uaf_symbols.zip](https://github.com/radare/radare2/files/2352890/r2_uaf_symbols.zip)
- Unpack and run radare with command: `r2 -A r2_uaf_symbols`

### Additional Logs, screenshots, source-code,  configuration dump, ...

```
==8161==ERROR: AddressSanitizer: heap-use-after-free on address 0x60300002fcb0 at pc 0x7fe43eb3276a bp 0x7fff828bdd60 sp 0x7fff828bdd50
READ of size 8 at 0x60300002fcb0 thread T0
    #0 0x7fe43eb32769 in symbols XYZ/radare2/libr/..//libr/bin/p/bin_wasm.c:168
    #1 0x7fe43e794c00 in r_bin_object_set_items XYZ/radare2/libr/bin/obj.c:173
    #2 0x7fe43e7975a6 in r_bin_object_new XYZ/radare2/libr/bin/obj.c:95
    #3 0x7fe43e78f66f in r_bin_file_new_from_bytes XYZ/radare2/libr/bin/bfile.c:515
    #4 0x7fe43e75d811 in r_bin_load_io2 XYZ/radare2/libr/bin/bin.c:485
    #5 0x7fe43e75e411 in r_bin_load_io XYZ/radare2/libr/bin/bin.c:42
    #6 0x7fe43fc9386c in r_core_file_do_load_for_io_plugin XYZ/radare2/libr/core/cfile.c:385
    #7 0x7fe43fc9386c in r_core_bin_load XYZ/radare2/libr/core/cfile.c:543
    #8 0x55c1e7df930b in main XYZ/radare2/binr/radare2/radare2.c:1093
[...]

PR #11533, opened 2018-09-15
Fix a use-after-free and a memory leak in handling of WASM binaries 

pageflt:
This PR addresses the following problems in handling of WASM binaries:

- **Fix issue #11399: use-after-free in symbols():**

Commit 7e083b57f introduced the issue #11399. The list referenced by
`codes` in `entries()`, is the same list that `bf->g_codes` is pointing at.
By freeing it, we introduce a use-after-free in a subsequent call within
`symbols()`, where we try to iterate over the list that `bf->g_codes` is
supposed to be referencing.

- **Fix memleak during loading of WASM binaries:**

A memory leak is reported by ASAN when handling WASM binaries. It is
caused by multiple allocations of `RBinInfo` structure. First, `RBinInfo`
is allocated within a call to `size()` from `r_bin_object_set_items()`. Then
there is another, explicit allocation of a `RBinInfo` structure through
a call to the `info()` callback of the WASM `RBinPlugin`. This causes loss
of reference to the initial structure, and subsequently a leak.

There are no apparent uses of `RBinInfo` structure inbetween these two
points, and the `size()` result is in no way dependent on this structure,
therefore I resolved the memory leak by removing the allocation from
within `size()`.

After addressing these issues, I experience no crashes or ASAN errors
when working with the reproducer of #11399 or other WASM binaries:

```
$ r2 -A samples/11399/r2_uaf_symbols
err: beach sections
[x] Analyze value pointers (aav)
[x] Value from 0x00000000 to 0x0001c862 (aav)
[x] 0x00000000-0x0001c862 in 0x0-0x1c862 (aav)
[x] Constructing a function name for fcn.* and sym.func.* functions (aan)
[x] Type matching analysis for all functions (afta)
 -- Wrong argument
[0x00000000]> q
$
```

### DIFF
--- a/libr/bin/p/bin_wasm.c
+++ b/libr/bin/p/bin_wasm.c
@@ -269,10 +269,7 @@ static RBinInfo *info(RBinFile *bf) {
 }
 
 static ut64 size(RBinFile *bf) {
-	if (!bf->o->info) {
-		bf->o->info = info (bf);
-	}
-	if (!bf->o->info) {
+	if (!bf || !bf->buf) {
 		return 0;
 	}
 	return bf->buf->length;
--- a/libr/bin/p/bin_wasm.c
+++ b/libr/bin/p/bin_wasm.c
@@ -72,7 +72,6 @@ static RList *entries(RBinFile *bf) {
 		}
 		if (!addr) {
 			r_list_free (ret);
-			r_list_free (codes);
 			return NULL;
 		}
 	}


libr/bin/p/bin_wasm.c
```diff
@@ -72,7 +72,6 @@ static RList *entries(RBinFile *bf) {
 		}
 		if (!addr) {
 			r_list_free (ret);
-			r_list_free (codes);
```
radare: yeah, this is wrong, the list is owned by the bin object, not returning a copy

PR #11535, opened 2018-09-15
Underscore cmd

radare:


libr/cons/cons.c
```diff
@@ -687,21 +687,44 @@ R_API void r_cons_context_break(RConsContext *context) {
...
+		if (!CTX(lastMode)) {
+			// snapshot of the output
+			free (CTX(lastOutput));
+			CTX(lastOutput) = calloc (CTX(buffer_len) + 1, 1);
+			if (CTX(lastOutput)) {
+				CTX(lastLength) = CTX(buffer_len);
+				memcpy (CTX(lastOutput), CTX(buffer), CTX(buffer_len));
+			}
+		}
+		CTX(lastMode) = false;
```
dsh2: For the else-branch the assignment seems trivial. So it actually belongs to the if-branch?

radare: Agree. Will change before merge

libr/cons/cons.c
```diff
@@ -687,7 +688,25 @@ R_API void r_cons_context_break(RConsContext *context) {
...
-R_API void r_cons_flush() {
+#define CTX(x) I.context->x
+
+R_API void r_cons_last(void) {
+	if (!CTX (lastEnabled)) {
+		return;
+	}
+	CTX (lastMode) = true;
+	r_cons_memcat (CTX (lastOutput), CTX (lastLength));
+	//r_cons_print (lastOutput);
```
xarkes: Don't merge useless commented lines, people can git log.

radare:
Not gonna merge this yet i have more things to improve and test: will do tonight

libr/cons/cons.c
```diff
@@ -696,12 +715,25 @@ R_API void r_cons_flush() {
 		r_cons_reset ();
 		return;
 	}
+ 	if (lastMatters ()) {
+		if (!CTX (lastMode)) {
+			// snapshot of the output
+			free (CTX (lastOutput));
+			CTX (lastOutput) = calloc (CTX (buffer_len) + 1, 1);
```
ret2libc: `R_NEWS0`?

libr/cons/cons.c
```diff
@@ -696,12 +715,25 @@ R_API void r_cons_flush() {
...
+			if (CTX (lastOutput)) {
+				CTX (lastLength) = CTX (buffer_len);
```
ret2libc: I don't think you need this, since `lastLength` is always equal to `buffer_len` you can just use the latter.

libr/cons/cons.c
```diff
@@ -696,12 +715,25 @@ R_API void r_cons_flush() {
 		r_cons_reset ();
 		return;
 	}
+ 	if (lastMatters ()) {
+		if (!CTX (lastMode)) {
```
ret2libc: just do `if (lastMatters () && !CTX (lastMode))` and set `lastMode = true` inside the inner if.

radare:
the use of this command is the same as in perl, python or awk. the _ variable contains the output of the last command. many people use to do a long disassembly or listing of things from a remote side like frida agent, which may take a while to be processed and printed.

Using _ the user can just grep for different things or reuse the last output for different purposes without having to run the command again. I was thinking in also adding a numeric argument to _ to choose a different slot, like multiple clipboards.

ret2libc:
Got it. Does it keep the same output if you do multiple greps?

libr/cons/cons.c
radare: thats not the same

radare:
yes, the output captured is before any grep is applied. so you can do:

```
> pd 1000
> _~call
> _~pop
```

radare:
Cmd.repeat have a different usage. When enabled if u press x it will run x if u type an empty command, but in the next block.

Same goes for other commands like pd or ds

libr/cons/cons.c
radare: refactored the code, now its cleaner, thanks for the hint

libr/cons/cons.c
radare: im now using malloc when needed, much faster than before, but still handy to have scr.last to disable it at runtime

libr/cons/cons.c
radare: i need it for the r_cons_last() which prints the last buffer to be flushed laster, in that case lastMode is set so its not captured again and the same cache is used.

Issue #10633, opened 2018-07-05
ARM 32-bit Emulation BL/BLX error


SpenserStyles:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Mac 10.13.5 64-bit
| File format of the file you reverse (mandatory)      | BIN
| Architecture/bits of the file (mandatory)            | arm 32-bit cortex
| r2 -v full output, **not truncated** (mandatory)     | radare2 2.6.9 18640 @ darwin-x86-64 git.2.6.9 commit: 159d16b34550f4a83ee496839981c8316339ac4d build: 2018-07-03__09:48:34

### Expected behavior
On BL/BLX store correct return address into LR.

### Actual behavior
BL/BLX stores the expected return address plus 4 which is incorrect. 

### Steps to reproduce the behavior 
Not sure why this is occuring because when I ae the command it produces the expected result but when I single step it does not.  
Example:
PC = 0x800199a
BL 0x8001CA5
=> This causes LR to be 0x80019A2 when it should be 0x800199E

My current fix was to edit the anal_arm_cs.c code to always subtract 4 from LR after computation

```
case ARM_INS_BL:
	case ARM_INS_BLX:
		//r_strbuf_appendf (&op->esil, "4,pc,+,lr,=");
		r_strbuf_appendf (&op->esil, "4,pc,+,lr,=,4,lr,-=,%s,pc,=", ARG(0)); // FIX TO CORRECTLY COMPUTE LR
```

radare:
Dont paste code here just press the pencil button and make a pullreq so we can review and comment

Maijin:
Use https://help.github.com/articles/editing-files-in-your-repository/

Maijin:
Created a PR with the diff patch - https://github.com/radare/radare2/pull/11539

PR #11539, opened 2018-09-16
Fix - ARM 32-bit Emulation BL/BLX error #10633

Maijin:
See @SpenserStyles - https://github.com/radare/radare2/issues/10633

### DIFF
--- a/libr/anal/p/anal_arm_cs.c
+++ b/libr/anal/p/anal_arm_cs.c
@@ -1598,7 +1598,7 @@ r4,r5,r6,3,sp,[*],12,sp,+=
 		break;
 	case ARM_INS_BL:
 	case ARM_INS_BLX:
-		r_strbuf_appendf (&op->esil, "4,pc,+,lr,=,4,lr,-=,%s,pc,=", ARG(0));
+		r_strbuf_appendf (&op->esil, "pc,lr,=,%s,pc,=", ARG(0));
 		break;
 	case ARM_INS_MOVT:
 		r_strbuf_appendf (&op->esil, "16,%s,<<,%s,|=", ARG(1), REG(0));
--- a/libr/anal/p/anal_arm_cs.c
+++ b/libr/anal/p/anal_arm_cs.c
@@ -1598,7 +1598,7 @@ r4,r5,r6,3,sp,[*],12,sp,+=
 		break;
 	case ARM_INS_BL:
 	case ARM_INS_BLX:
-		r_strbuf_appendf (&op->esil, "4,pc,+,lr,=,%s,pc,=", ARG(0));
+		r_strbuf_appendf (&op->esil, "4,pc,+,lr,=,4,lr,-=,%s,pc,=", ARG(0));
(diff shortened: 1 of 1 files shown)


Maijin:
image

Maijin:
Updated screenshot:

image

radare:
the new output looks wrong to me, do you have a test case to proof your patch?

radare:
new out put looks wrong to me

libr/anal/p/anal_arm_cs.c
```diff
@@ -1598,7 +1598,7 @@ r4,r5,r6,3,sp,[*],12,sp,+=
 		break;
 	case ARM_INS_BL:
 	case ARM_INS_BLX:
-		r_strbuf_appendf (&op->esil, "4,pc,+,lr,=,%s,pc,=", ARG(0));
+		r_strbuf_appendf (&op->esil, "4,pc,+,lr,=,4,lr,-=,%s,pc,=", ARG(0));
```
mrmacete: isn't this redundant? why not just
```
r_strbuf_appendf (&op->esil, "pc,lr,=,%s,pc,=", ARG(0));
```
it would be consistent with arm64 implementation.

and yeah all of this assume that `pc` is already pointing to the next instruction.

Maijin: See the message here https://github.com/radare/radare2/issues/10633

Maijin:
```
Not sure why this is occuring because when I ae the command it produces the expected result but when I single step it does not.
Example:
PC = 0x800199a
BL 0x8001CA5
=> This causes LR to be 0x80019A2 when it should be 0x800199E
```

mrmacete: ok so probably there's a wrong assumption on what the value of `pc` should be when stepping on the ESIL code

radare:
Use $$ instead of pc maybe?

mrmacete:
yes, so when using `$$` then the ESIL representation of the instruction must take into account the current instruction's size to initialize `lr` properly i guess

mrmacete:
ok, so if i understood the code correctly, when doing ESIL step the `pc` register is guaranteed to point to the next instruction: https://github.com/radare/radare2/blob/master/libr/core/cmd_anal.c#L3481-L3484

and that's very cool for cases like setting `lr` because in this way ESIL implementations of instructions don't have to care about instruction size at all.

so this PR is ok for me (maybe i'd simplify that expression to get rid of sum and subtraction of 4 which cancel out).

The difference in behaviour when using `ae` vs. `aes` it's exactly for how the `pc` gets initialized but that sounds correct. `$$` can be used instead when referring to the address of the current instruction which in general is different from `pc`.

Maijin:
Looks like I can merge right away, no red this time 👍

mrmacete:
ok i'm going to add tests for this in r2r anyways

Maijin:
Great!

PR #11546, opened 2018-09-16
Introduce new R_* macros to define direction, null, defaults, etc in function signatures

radare:
- Helps to identify some bugs at compile time
- Useful for git grep and understand the usage of the functions
- Improve readability
- Fix memleak introduced in previous anal/var pr
- Help to find unnecessary null checks and start to use r_assert
- Deprecate R_TRUFAE R_TRUE R_FALSE and R_ERROR

### DIFF
--- a/libr/anal/var.c
+++ b/libr/anal/var.c
@@ -68,11 +68,9 @@ R_API bool r_anal_var_display(RAnal *anal, int delta, char kind, const char *typ
 	return true;
 }
 
-R_API bool r_anal_var_add(RAnal *a, ut64 addr, int scope, int delta, char kind, const char *type, int size,
-		bool isarg, const char *name) {
-	if (!a) {
-		return false;
-	}
+R_API bool r_anal_var_add(RAnal *a, ut64 addr, int scope, int delta, char kind, R_IFNULL("int") const char *type, int size, bool isarg, R_NONNULL const char *name) {
+	r_return_val_if_fail (a, false);
+	r_return_val_if_fail (name, false);
 	if (!kind) {
 		kind = R_ANAL_VAR_KIND_BPV;
 	}
--- a/libr/include/r_types.h
+++ b/libr/include/r_types.h
@@ -13,6 +13,20 @@
 #undef __UNIX__
 #undef __WINDOWS__
 
+#define R_IN /* do not use, implicit */
+#define R_OWN /*own*/
+#define R_OUT /*out*/
+#define R_INOUT /*inout*/
+#define R_NULL /*null*/
+#define R_IFNULL(x) /*ifnull*/
+#ifdef __GNUC__
+#define R_NONNULL __attribute__((__nonnull__))
+#define R_UNUSED __attribute__((__unused__))
+#else
+#define R_NONNULL /*nonnull*/
+#define R_UNUSED /*unused*/
+#endif
+
 #ifdef R_NEW
 #undef R_NEW
 #endif
@@ -199,14 +213,6 @@ extern "C" {
 
 #ifndef __packed
 #define __packed __attribute__((__packed__))
-#endif
-
-#ifndef UNUSED
-#ifdef __GNUC__
-#define UNUSED __attribute__((__unused__))
-#else
-#define UNUSED
-#endif
 #endif
 
 typedef int (*PrintfCallback)(const char *str, ...);
--- a/libr/include/r_types.h
+++ b/libr/include/r_types.h
@@ -14,17 +14,17 @@
 #undef __WINDOWS__
 
 #define R_IN /* do not use, implicit */
-#define R_OWN /*own*/
-#define R_OUT /*out*/
-#define R_INOUT /*inout*/
-#define R_NULL /*null*/
-#define R_IFNULL(x) /*ifnull*/
+#define R_OWN /* pointer ownership is transferred */
+#define R_OUT /* parameter is written, not read */
+#define R_INOUT /* parameter is read and written */
+#define R_NULLABLE /* pointer can be null */
+#define R_IFNULL(x) /* default value for the pointer when null */
 #ifdef __GNUC__
 #define R_NONNULL __attribute__((__nonnull__))
 #define R_UNUSED __attribute__((__unused__))
 #else
-#define R_NONNULL /*nonnull*/
-#define R_UNUSED /*unused*/
+#define R_NONNULL /* nonnull */
+#define R_UNUSED /* unused */
 #endif
 
 #ifdef R_NEW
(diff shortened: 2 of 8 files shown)


condret:
R_TRUFAE must stay, it always makes me smile :D

libr/anal/var.c
```diff
@@ -68,11 +68,9 @@ R_API bool r_anal_var_display(RAnal *anal, int delta, char kind, const char *typ
 	return true;
 }
 
-R_API bool r_anal_var_add(RAnal *a, ut64 addr, int scope, int delta, char kind, const char *type, int size,
-		bool isarg, const char *name) {
-	if (!a) {
-		return false;
-	}
+R_API bool r_anal_var_add(RAnal *a, ut64 addr, int scope, int delta, char kind, R_IFNULL("int") const char *type, int size, bool isarg, R_NONNULL const char *name) {
```
ret2libc: shall we add `R_IFNULL(R_ANAL_VAR_KIND_BPV) char kind` too? maybe we need to split the signature in 2 lines since it's becoming long :/

libr/include/r_types.h
```diff
@@ -13,6 +13,20 @@
 #undef __UNIX__
 #undef __WINDOWS__
 
+#define R_IN /* do not use, implicit */
+#define R_OWN /*own*/
```
ret2libc: I guess that when it's used on a return value of a function, it means that the caller is responsible for freeing the element.

What about when it's used as argument to a function? What does that mean? Shall we be more descriptive in the comment to document the behaviour? Currently I'm not sure I understand 100% this one

libr/include/r_types.h
```diff
@@ -13,6 +13,20 @@
...
+#define R_OUT /*out*/
+#define R_INOUT /*inout*/
+#define R_NULL /*null*/
```
ret2libc: shall we mark this as default behaviour? otherwise it's unclear whether a dev should expect null or not when nothing is used.

libr/include/r_util/r_ctypes.h
```diff
@@ -36,9 +36,9 @@ R_API char *r_type_format(Sdb *TDB, const char *t);
 R_API int r_type_func_exist(Sdb *TDB, const char *func_name);
 R_API const char *r_type_func_cc(Sdb *TDB, const char *func_name);
 R_API const char *r_type_func_ret(Sdb *TDB, const char *func_name);
-R_API int r_type_func_args_count(Sdb *TDB, const char *func_name);
-R_API char *r_type_func_args_type(Sdb *TDB, const char *func_name, int i);
-R_API char *r_type_func_args_name(Sdb *TDB, const char *func_name, int i);
+R_API int r_type_func_args_count(Sdb *TDB, R_NONNULL const char *func_name);
+R_API char *r_type_func_args_type(Sdb *TDB, R_NONNULL const char *func_name, int i);
```
ret2libc: R_OWN here

libr/util/ctype.c
```diff
@@ -482,7 +482,7 @@ R_API char *r_type_func_args_name(Sdb *TDB, const char *func_name, int i) {
 
 #define MIN_MATCH_LEN 4
 
-static char *type_func_try_guess(Sdb *TDB, char *name) {
+static R_OWN char *type_func_try_guess(Sdb *TDB, R_NONNULL char *name) {
```
ret2libc: are we going to use these macros in `static` functions too? I'd be to use them only in R_API functions, static functions can change often.

ret2libc:
Where is the part of code for `Fix memleak introduced in previous anal/var pr`?

radare:
the anal/var pr wasnt merged, he fixed it and i rebased, so my fix was removed from this pr

libr/util/ctype.c
radare: yes

libr/include/r_types.h
radare: maybe R_NULLABLE is a better name :? not sure which should be the default but i would expect nothing can be null unless specified

libr/anal/var.c
radare: The function signature is too complex imho, we should rethink it maybe. But having it in one line is the only sane way to make it work with grep.

R_IFNULL(kind) makes no sense because char cant be null, unless we assume that a nul char is the same as null. but it looks inconsistent to me to use R_IFNULL with char.

libr/include/r_types.h
MaskRay: The comments should be more descriptive. I'll assume `R_OWN` as argument annotation means the callee will take ownership of whatever the argument owns, resembling the following:
```cpp
void sink(std::unique_ptr<int> p)
void sink(std::vector<int>&& p)
```

libr/include/r_types.h
ret2libc: I like `R_NULLABLE` and I'm fine with assuming everything should be non-null, unless specified.

libr/util/ctype.c
ret2libc: kk

libr/include/r_types.h
radare: btw, what about R_OWNED instead of R_OWN

libr/include/r_types_base.h
```diff
@@ -59,13 +59,11 @@ typedef struct _utX{
...
 
-#define R_ERROR -2
 #define R_FAIL -1
-#define R_FALSE 0
-#define R_TRUE 1
-#define R_TRUFAE 2
+// must be deprecated
 #define R_NOTNULL (void*)(size_t)1
 #define R_EMPTY { 0 }
+#define R_EMPTY2 {{ 0 }}
```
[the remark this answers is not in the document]
radare: `$ git grep '{{0}}'`

libr/anal/var.c
ret2libc: You are right.

libr/include/r_util/r_ctypes.h
```diff
@@ -36,10 +36,10 @@ R_API char *r_type_format(Sdb *TDB, const char *t);
 R_API int r_type_func_exist(Sdb *TDB, const char *func_name);
 R_API const char *r_type_func_cc(Sdb *TDB, const char *func_name);
 R_API const char *r_type_func_ret(Sdb *TDB, const char *func_name);
-R_API int r_type_func_args_count(Sdb *TDB, const char *func_name);
-R_API char *r_type_func_args_type(Sdb *TDB, const char *func_name, int i);
-R_API char *r_type_func_args_name(Sdb *TDB, const char *func_name, int i);
-R_API char *r_type_func_guess(Sdb *TDB, char *func_name);
+R_API int r_type_func_args_count(Sdb *TDB, R_NONNULL const char *func_name);
```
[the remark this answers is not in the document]
radare: yeah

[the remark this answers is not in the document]
radare: but it doesnt hurts to make that implicit either :D #watdu

PR #11547, opened 2018-09-16
Fix #11460 - Show help message for unhandled f subcommands

### LINKED ISSUE #11460: Nonexistent subcommands do not issue an error
### Expected behavior

```
[0x00000000]> h
|ERROR| Invalid command 'h' (0x68)
[0x00000000]> fu
|ERROR| Invalid command 'fu' (0x6675)
[0x00000000]> 
```

### Actual behavior
```
[0x00000000]> h
|ERROR| Invalid command 'h' (0x68)
[0x00000000]> fu
[0x00000000]> 
```

radare:


radare:
I used this behaviour because it seems consistent with other commands, but maybe we should specify a one line error when command doesnt exist instead of show the command help

radare:
But this should change in all the commands, not just 'f'

ret2libc:
yeah... This should be handled when we'll refactor command parsing

PR #11548, opened 2018-09-16
Fix #11455 - Implement internal less pager with scr.pager=..

### LINKED ISSUE #11455: Implement native scr.pager in RCons

radare:


libr/cons/cons.c
```diff
@@ -707,6 +707,7 @@ static bool lastMatters() {
 }
 
 R_API void r_cons_flush(void) {
+	static bool pageable = true;
```
ret2libc: what about adding it as a variable to `RCons`? I'm not a big fan of static vars.

libr/cons/cons.c
```diff
@@ -730,10 +731,18 @@ R_API void r_cons_flush(void) {
 
 	if (I.is_interactive && I.fdout == 1) {
 		/* Use a pager if the output doesn't fit on the terminal window. */
-		if (I.pager && *I.pager && I.context->buffer_len > 0 && r_str_char_count (I.context->buffer, '\n') >= I.rows) {
+		if (pageable && I.pager && *I.pager && I.context->buffer_len > 0 && r_str_char_count (I.context->buffer, '\n') >= I.rows) {
 			I.context->buffer[I.context->buffer_len - 1] = 0;
-			r_sys_cmd_str_full (I.pager, I.context->buffer, NULL, NULL, NULL);
-			r_cons_reset ();
+			if (*I.pager == '.') {
```
ret2libc: I'd use `strcmp` to avoid considering `scr.pager=./mybin` as internal less.

radare:
better name for scr.pager=. ?

libr/cons/cons.c
radare: i think . is a pretty bad name, in fact i was thikning in using .. like in ~.. but wanted to discuss without overengineering. also scr.pager=? doesnt shows anything

ret2libc: Ok, but in any case it makes sense to use `strcmp`. The compiler is anyway smart enough to reduce it to just a 2bytes check i think.

libr/cons/cons.c
radare: yes, maybe useful for tasks because it swaps that global rconscontext struct, so it makes sense. changing

libr/cons/cons.c
```diff
@@ -715,6 +715,7 @@ R_API void r_cons_flush(void) {
 		r_cons_reset ();
 		return;
 	}
+	CTX (pageable) = true;
```
ret2libc: mmm this seems wrong, otherwise it will always enter the `if (CTX (pageable)`

libr/cons/cons.c
```diff
@@ -727,13 +729,19 @@ R_API void r_cons_flush(void) {
...
 		/* Use a pager if the output doesn't fit on the terminal window. */
-		if (I.pager && *I.pager && I.context->buffer_len > 0 && r_str_char_count (I.context->buffer, '\n') >= I.rows) {
+		if (CTX (pageable) && I.pager && *I.pager && I.context->buffer_len > 0 && r_str_char_count (I.context->buffer, '\n') >= I.rows) {
 			I.context->buffer[I.context->buffer_len - 1] = 0;
-			r_sys_cmd_str_full (I.pager, I.context->buffer, NULL, NULL, NULL);
-			r_cons_reset ();
+			if (!strcmp (I.pager, "..")) {
+				char *str = r_str_ndup (CTX (buffer), CTX (buffer_len));
+				CTX (pageable) = false;
+				r_cons_less_str (str, NULL);
```
ret2libc: where do you set `CTX (pageable)` back to true?

radare: oops

radare:
Related to https://github.com/radare/radare2/issues/11552

radare:
i have fixed r_cons_less in this pr to make it fill the whole screen, this may be related. let me fix before merge.

ret2libc:
> i have fixed r_cons_less in this pr to make it fill the whole screen, this may be related. let me fix before merge. 

Yeah, though even with this PR the last line is still not printed.

PR #11555, opened 2018-09-17
Add undefined behaviour tests in the default sys/asan.sh

radare:


XVilka:
Not the only, but probably not covered by tests. Lets keep it. For example I spotted it before in ESIL and PDB.

PR #11561, opened 2018-09-17
ESIL interrupts: add aeli and aelir

vifino:
no help yet. aeli- and more to come later.

paging dr @condret

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -2,6 +2,7 @@
 
 #include "r_util.h"
 #include "r_core.h"
+#include "r_anal.h"
 
 static const char *help_msg_a[] = {
 	"Usage:", "a", "[abdefFghoprxstc] [...]",
@@ -101,6 +102,9 @@ static const char *help_msg_ae[] = {
 	"aeip", "", "initialize ESIL program counter to curseek",
 	"aek", " [query]", "perform sdb query on ESIL.info",
 	"aek-", "", "resets the ESIL.info sdb instance",
+	"aeli", "", "list loaded ESIL interrupts",
+	"aeli", " [file]", "load ESIL interrupts from shared object",
+	"aelir" " [interrupt number]", "remove ESIL interrupt and free it if needed",
 	"aep", "[?] [addr]", "manage esil pin hooks",
 	"aepc", " [addr]", "change esil PC to this address",
 	"aer", " [..]", "handle ESIL registers like 'ar' or 'dr' does",
@@ -4220,6 +4224,8 @@ static bool cmd_aea(RCore* core, int mode, ut64 addr, int length) {
 	return true;
 }
 
+
+
 static void cmd_aespc(RCore *core, ut64 addr, int off) {
 	RAnalEsil *esil = core->anal->esil;
 	int i, j = 0;
@@ -4274,6 +4280,13 @@ static void cmd_aespc(RCore *core, ut64 addr, int off) {
 	r_reg_set_value (core->dbg->reg, r, curpc);
 	r_core_esil_step (core, curpc + instr_size, NULL, NULL);
 	r_core_seek (core, oldoff, 1);
+}
+
+static const char _handler_no_name[] = "<no name>";
+static int _aeli_iter(dictkv* kv, void* ud) {
+	RAnalEsilInterrupt* interrupt = kv->u;
+	r_cons_printf ("%3x: %s\n", kv->k, interrupt->handler->name ? interrupt->handler->name : _handler_no_name);
+	return 0;
 }
 
 static void cmd_anal_esil(RCore *core, const char *input) {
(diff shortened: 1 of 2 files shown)


libr/core/cmd_anal.c
```diff
@@ -4562,6 +4575,31 @@ static void cmd_anal_esil(RCore *core, const char *input) {
...
+		case 'i': // aeli interrupts
+			switch (input[2]) {
+			case ' ': // "aeli" with arguments
+				if (!r_anal_esil_load_interrupts_from_lib (esil, input + 3)) {
+					eprintf ("Failed to load interrupts from '%s'.", input + 3);
+				}
+				break;
+			case 0: // "aeli" with no args
+				if (esil && esil->interrupts) {
+					dict_foreach (esil->interrupts, _aeli_iter, NULL);
```
radare: arent we suposed to kill dict_*¿

libr/core/cmd_anal.c
```diff
@@ -101,6 +102,9 @@ static const char *help_msg_ae[] = {
 	"aeip", "", "initialize ESIL program counter to curseek",
 	"aek", " [query]", "perform sdb query on ESIL.info",
 	"aek-", "", "resets the ESIL.info sdb instance",
+	"aeli", "", "list loaded ESIL interrupts",
+	"aeli", " [file]", "load ESIL interrupts from shared object",
+	"aelir" " [interrupt number]", "remove ESIL interrupt and free it if needed",
```
radare: use `aeli-` instead

libr/core/cmd_anal.c
```diff
@@ -4276,6 +4282,13 @@ static void cmd_aespc(RCore *core, ut64 addr, int off) {
 	r_core_seek (core, oldoff, 1);
 }
 
+static const char _handler_no_name[] = "<no name>";
+static int _aeli_iter(dictkv* kv, void* ud) {
+	RAnalEsilInterrupt* interrupt = kv->u;
+	r_cons_printf ("%3x: %s\n", kv->k, interrupt->handler->name ? interrupt->handler->name : _handler_no_name);
+	return 0;
```
radare: shouldnt return 1?

radare: (or bool?)

ret2libc:
yes, it's probably for another PR since dict was already added by @condret in the initial PR. But yeah, it should be killed in favour of just one hashtable implementation.

libr/core/cmd_anal.c
vifino: i disagree, `aeli-` should in the future unload whole libraries, not specific interrupts

libr/core/cmd_anal.c
vifino: not sure, think not, i think the iter returns 1 when it found the element so it stops calling it to exit early

libr/core/cmd_anal.c
vifino: (i see `cmd-` as undoing the effects of `cmd` in a way, so `aeli` would undo the loading of a file, therefore unload it, which aelir doesn't do, it just removes an interrupt)

libr/core/cmd_anal.c
vifino: another pr, maybe?
i tried making the most minimal amount of changes

condret:
I see no point in using sdb, for ints as keys. Yes, sdb is a multi purpose ht. You can use multipurpose tools for every task, but if you have specific tools for specific tasks available, which work more efficient, then you should use the more efficient tools

libr/core/cmd_anal.c
radare: isnt this ael- vs aeli- ? imho the 'r' should not be used to remoev things because - is used in all the other commands.

ret2libc:
In short, the reasons are:
1. SdbHt should handle both `char *` and `int` and whatever as key
2. dict doesn't allow you to have 0 as key AFAIK, because 0 == MHTNO, so you'll always have one value which is considered "special" and it will cause random failures in the moment those values are used (and this is just a design decision, so to remove that you should rewrite it entirely)
3. dict doesn't grow, so it's just like having multiple lists (in particular, at esil_interrupt.c:r_anal_esil_interrupts_init the esil->interrupts dict is initialized with size = sizeof(ut32), which means that you just have 4 lists :D )
4. dict uses the simple `k % m->size` to compute the bucket, which is not ideal because it's not well distributed (m->size is even provided by the user...)
5. dict interface is weird, it's definitely not what you'd expect from an hashtable (e.g. `bool dict_set(dict *m, dicti k, dicti v, void *u)` if k is the key, and v is the value, what is `u`?)

SdbHt was designed to be generic, it can easily support int as keys, it has already support for growing (though it was disabled because ... boh! :D) it uses prime numbers to reduce collisions, it has a very simple clear API and no special values that are treated differently AFAIK.

That said, I absolutely agree that for this PR dict_foreach is perfectly fine since dict was already used.

condret:
the second point kills dict, unless we do ugly hacks, which I'm not a fan of.
Interrupt 0 is valid and actually used on some platforms.
The problem of sdb for me is, that you always need to do itoa for hashing, which sucks. I didn't look at SdbHt yet, but if we can avoid using itoa on the ut32 key, I'd prefere that, just bc I don't want ugly hacks here

ret2libc:
SdbHt is not related to sdb. It comes from sdb source code, but that's it. It does have `char *` as key, but I think we can improve it a bit to handle other types as well.

ret2libc:
BTW, please double check but I'm pretty sure point 2 still applies since last time I used `dict`.

libr/core/cmd_anal.c
radare: uhm yeah.. its a bit confusing, lets merge for now and ill find time to think better constructions here

Issue #11141, opened 2018-08-19
Kill R_ASM_BUFSIZE


radare:
<img width="982" alt="screen shot 2018-08-19 at 05 35 23" src="https://user-images.githubusercontent.com/917142/44305178-bb12bf00-a371-11e8-80e7-95b39e2ff5dc.png">

MaskRay:
Use `sizeof op->txt` in these cases?

radare:
Nope. There must be no fixed size buffers at all. They are evil and limit us to do nice things. I mean they make sense for some things but not for that case. Because is better to use accessor methods to hide this complexity like i did in my pr

radare:
cc @wargio

wargio:
i'm ok with that.

radare:
just use RStrBuf instead of fixed stack buffers

PR #11563, opened 2018-09-17
replaced R_ASM_BUFSIZE w/ RStrBuf as per #11141

rfc2119:


radare:
Nope. thats not how you use the RStrBuf API instead of using fixed size array buffers.

kazarmy:
Since this clearly won't compile, I've cancelled the Travis build.

rfc2119:
oh, forgive my idiocy! I didn't know `RStrBuf` was a typedef. Should I do  `sizeof(RStrBuf)` instead ?

PR #11565, opened 2018-09-17
Fix #11554, by checking the return code of r_bin_mdmp_init_directory_…

### LINKED ISSUE #11554: Issue 10403 in oss-fuzz: radare2/ia_fuzz: Timeout in radare2_ia_fuzz
[clusterfuzz-testcase-minimized-ia_fuzz-5756355915284480.dms.zip](https://github.com/radare/radare2/files/2387911/clusterfuzz-testcase-minimized-ia_fuzz-5756355915284480.dms.zip)

mscherer:
…entry

### DIFF
--- a/libr/bin/format/mdmp/mdmp.c
+++ b/libr/bin/format/mdmp/mdmp.c
@@ -700,7 +700,7 @@ static bool r_bin_mdmp_init_directory_entry(struct r_bin_mdmp_obj *obj, struct m
 		break;
 	default:
 		eprintf ("[WARN] Invalid or unsupported enumeration encountered %i\n", entry->stream_type);
-		return false;
+		break;
 	}
 	return true;
 }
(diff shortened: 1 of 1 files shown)


libr/bin/format/mdmp/mdmp.c
```diff
@@ -720,7 +720,9 @@ static bool r_bin_mdmp_init_directory(struct r_bin_mdmp_obj *obj) {
 		ut32 delta = i * sizeof (struct minidump_directory);
 		int r = r_buf_read_at (obj->b, rvadir + delta, (ut8*) &entry, sizeof (struct minidump_directory));
 		if (r) {
-			r_bin_mdmp_init_directory_entry (obj, &entry);
+			if (!r_bin_mdmp_init_directory_entry (obj, &entry)) {
+				return false;
+			};
```
radare: unnecessary ;

radare:
so much xx

mscherer:
Ok so it seems that the tests have some unsupported enumeration, so I guess the right fix is either to ignore them (but that's a hack), or to add support for it. I think support is cleaner, but then, what about future format change, do we want to declare a file using a unsupported enumeration valid or not ?

XVilka:
See also https://github.com/radare/radare2/issues/10992
see `om` output that all maps do not even have "read" attribute, after setting it everything works.

PR #11567, opened 2018-09-17
WIP: various dict/ht improvements

ret2libc:


radare:
Why? If we use SdbHt, the logical api to use is sdb_ht with it. I understand the sdb ht must be internal to sdb. But i dont like the naming changes here

Maijin:
it's closed already

ret2libc:
This is already closed, yes, but the why is that sdb_ht is just a "subclass" of ht and there are no reasons to use sdb_ht here, since you don't need neither `cas` nor `expire` or anything else.

radare:
Agree on that. But its just a naming confusion that i want to clarify befoe renaming more apis. Anyway the change lgtm

ret2libc:
I won't merge this thing for now. I closed it indeed. Before i want to do some work on the ht itself.
And yeah, the names are really bad at the moment :/

PR #11570, opened 2018-09-17
Default Layout is supported in Load Layout menu

Vane11ope:
Thought it'd be nice that user can select and go back to the default layout whenever they want so I added the functionality as second level of Load Layout menu.

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -57,6 +57,12 @@ static const char *menus_ReOpen[] = {
 };
 static const int reopenNum = ((int)sizeof (menus_ReOpen) / (int)sizeof (const char *)) - 1;
 
+static const char *menus_loadLayout[] = {
+	"Saved", "Default",
+	NULL
+};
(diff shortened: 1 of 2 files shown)


libr/core/panels.c
```diff
@@ -57,6 +57,12 @@ static const char *menus_ReOpen[] = {
 };
 static const int reopenNum = ((int)sizeof (menus_ReOpen) / (int)sizeof (const char *)) - 1;
 
+static const char *menus_loadLayout[] = {
+	"Saved", "Default",
+	NULL
+};
+static const int loadLayoutNum = ((int)sizeof (menus_loadLayout) / (int)sizeof (const char *)) - 1;
```
XVilka: There are a lot of constructs like this, maybe it worth to make them a macro?

radare: change it here, we'll try to reduce the amount of commits fixing things from previous commits, that will make release notes easier

Vane11ope:
done.

PR #11572, opened 2018-09-18
r_types: Change R_NOTNULL to ~0

### LINKED ISSUE #11553: Kill R_NOTNULL

davidpolverari:
Closes #11553

### DIFF
--- a/libr/include/r_types_base.h
+++ b/libr/include/r_types_base.h
@@ -61,7 +61,7 @@ typedef struct _utX{
 
 #define R_FAIL -1
 // must be deprecated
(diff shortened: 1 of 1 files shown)


davidpolverari:
If this is not the right approach, I can replace all R_NOTNULL occurrences with the desired value.

radare:
Why? ~0 is a number and the current impl is a pointer. So nope its not the same

radare:
also, this is not fixing the issue at all, the issue aims to remove all the use cases of that variable because it must be removed, not changed

PR #11576, opened 2018-09-18
Fix - don’t assume ELF (arm) main address is physical

mrmacete:
Fixes https://github.com/radare/radare2/issues/11569
Depends on fixed test here: https://github.com/radare/radare2-regressions/pull/1490


libr/bin/format/elf/elf.c
```diff
@@ -1731,6 +1731,11 @@ ut64 Elf_(r_bin_elf_get_main_offset)(ELFOBJ *bin) {
 		*/
 		if (*addr > text && *addr < (text_end)) {
```
kazarmy: If this check is wrong, it should be removed

mrmacete: yeah i agree. except that i don't know the history of this code, so maybe it's correct for some binaries.

mrmacete: 👍

PR #11577, opened 2018-09-18
r_types: Kill R_NOTNULL

### LINKED ISSUE #11553: Kill R_NOTNULL

davidpolverari:
Closes #11553


XVilka:
This is wrong, see "which is a bad idea and shouldn't be used at all" part of the issue.

PR #11582, opened 2018-09-18
Add support for token information from minidump

mscherer:
While token stream is not documented on MSDN at the moment (or I didn't found it), the headers of Visual Studio and various projects do have the structure ( for example, mimikatz: https://github.com/gentilkiwi/mimikatz/blob/master/inc/DbgHelp.h#L4204 ), and from there, basic usage can be inferred.


XVilka:
A bit of red here.

mscherer:
Yup, gonna fix the test since that change the output, but no time before tonight at best :/

XVilka:
Please use the same name for the branch here and in radare2-regressions, it will use then the test from that branch.

PR #11587, opened 2018-09-19
Create aav.flagname when no flags have “.” at that address

mrmacete:
In this way it will be picked up by relsub and such, like in the case of “main” when it’s not also a symbol

Fixes https://github.com/radare/radare2/issues/11585
Related test: https://github.com/radare/radare2-regressions/pull/1493

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -6900,6 +6900,27 @@ void _CbInRangeAav(RCore *core, ut64 from, ut64 to, int vsize, bool asterisk, in
 			char *name = r_str_newf ("aav.0x%08"PFMT64x, to);
 			r_flag_set (core->flags, name, to, vsize);
 			free (name);
+		} else {
+			bool has_dot = false;
+			const char *flag_name = NULL;
+			const RList *flags = r_flag_get_list (core->flags, to);
+			RListIter *iter;
+			RFlagItem *f;
+			r_list_foreach (flags, iter, f) {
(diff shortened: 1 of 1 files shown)


mrmacete:
Compare the screenshots at offset `0x000083d8` (and `0x000083ec`)

Before:

screen shot 2018-09-19 at 22 14 12

After:

screen shot 2018-09-19 at 22 12 27

mrmacete:
The reason for this is that relsub code filters out all flags without a `'.'` in the name: https://github.com/radare/radare2/blob/4a722e80d8f7924cfbe2ddf72a55e2f36f1e23ac/libr/parse/parse.c#L107

radare:
Is there a problem if you just catch the first flag even if it contains a dot? Flags without dot can also be registers which is totally undesirable behaviour for this case. Maybe main should be named different

mrmacete:
oh yeah that's a good point. probably better to throw this PR away and name "main" with a dot in the first place... but how? "fcn.main" ?

mrmacete:
ok, so this PR is wrong anyways, because it produces `aav.main` === `main`. going to close this and go  back to the drawing board

radare:
Aav.main is not correct because aav.* means a place containing a dwords poiting to *.

As long as ther csn be multiple places pointing to the same place u need to prefix the thing with an address.

Maybe aav.0x6666.main would be better

But in still worried for cases where a register is used instead of a symbol name so im not sure about this dot check

PR #11609, opened 2018-09-21
Honor search.maxhits in /m

radare:


radare:
testcase. @Maijin can you add those? we probably are not honoring maxhits in many other places

screen shot 2018-09-21 at 13 41 17

PR #11610, opened 2018-09-21
Few improvements to type inference

### LINKED ISSUE #11445: Improve type inference
There are few things i would implement after r2con , so this is like self-note :D

* Flagspace analysis 

Infer `char *` for direct string assigment
 ```
mov qword [local_8h], str.Welcome_to_r2con2k18 
```

* Better pointer inference 

b2

b1

Until now only TIE is able to detect type properly for above example (tested with latest IDA too :} )

More suggestion are welcomed !

### PR DESCRIPTION
**Pull request by sivaramaaa**
Fixes #11445 

### Before 

before

### After 

after

Added a few test for this in r2r [#1495](https://github.com/radare/radare2-regressions/pull/1495)

### DIFF
--- a/libr/core/anal_tp.c
+++ b/libr/core/anal_tp.c
@@ -113,9 +113,12 @@ static void var_retype(RAnal *anal, RAnalVar *var, const char *vname, char *type
 		//type *ptr => type *
 		strncat (ntype, " *", len);
 	}
-	if (ref && r_str_endswith (ntype, "*")) {
-		// char * => char **
-		strncat (ntype, "*", len);
+	if (ref) {
+		if (r_str_endswith (ntype, "*")) { // type * => type **
+			strncat (ntype, "*", len);
+		} else {   //  type => type *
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by XVilka** [comment]
Please also add a tests for non-x86 architectures too - ARM, MIPS, PowerPC and AVR.

File: libr/core/anal_tp.c [hunk 1]
```diff
@@ -113,9 +113,12 @@ static void var_retype(RAnal *anal, RAnalVar *var, const char *vname, char *type
 		//type *ptr => type *
 		strncat (ntype, " *", len);
 	}
-	if (ref && r_str_endswith (ntype, "*")) {
-		// char * => char **
-		strncat (ntype, "*", len);
+	if (ref) {
+		if (r_str_endswith (ntype, "*")) { // type * => type **
+			strncat (ntype, "*", len);
```
radare: dont use strncat,, and if u do it do it properly. because this can overflow in many ways. use RStrBuf instead

PR #11621, opened 2018-09-22
RStrBuf: add r_strbuf_vappendf() and r_strbuf_vsetf()

pelijah:
Lets make code from `r_strbuf_setf()` and `r_strbuf_appendf()` reusable to avoid buggy implementations: https://github.com/radare/radare2/blob/master/libr/asm/p/asm_cris_gnu.c#L67


pelijah:
It is not my code. I think each `vsnprintf()` requires own `va_list`.

ret2libc:
Yeah I see. I wasn't aware of this thing!

pelijah:
It is just my assumption.

ret2libc:
Apparently it is right ;)

PR #11622, opened 2018-09-22
Fix #10843 - Make pipe lang plugin sleep while waiting for pipe

### LINKED ISSUE #10843: Running (python) scripts does not play well with tasks
I'm trying to use this new background task feature for python, but it's simply blocking waiting for python to return instead of actually backgrounding.

For instance:
```
& #!python script.py
```

The above will just block until script returns.

thestr4ng3r:
Fix #10843

### DIFF
--- a/libr/lang/p/pipe.c
+++ b/libr/lang/p/pipe.c
@@ -143,7 +143,9 @@ static int lang_pipe_run(RLang *lang, const char *code, int len) {
 				break;
 			}
 			memset (buf, 0, sizeof (buf));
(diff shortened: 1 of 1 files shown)


libr/lang/p/pipe.c
```diff
@@ -143,7 +143,9 @@ static int lang_pipe_run(RLang *lang, const char *code, int len) {
 				break;
 			}
 			memset (buf, 0, sizeof (buf));
+			void *bed = r_cons_sleep_begin ();
```
radare: ❤️

PR #11625, opened 2018-09-22
Fix GNU-based RAsm plugins to use RStrBuf API properly

pelijah:
Yeah looks ugly but it is better than to have 14 different implementations of the same function. The best way is to implement `generic_fprintf()` and `generic_print_address()` functions but they requires address of `RStrBuf`. Can we reuse `stream` field of `disassemble_info` struct too access it (see example below)? Now `stream` value is set to `stdout` but it seems that is not used anywhere.
```c
int generic_fprintf(void *stream, const char *format, ...) {
	...
	ret = r_strbuf_vappendf ( (RStrBuf *)stream, format, ap);
	...
}

void generic_print_address(bfd_vma address, struct disassemble_info *info) {
	...
	r_strbuf_appendf ((RStrBuf *)info->stream, "0x%08"PFMT64x, (ut64)address);
}

static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
	struct disassemble_info disasm_obj;
	...
	disasm_obj.print_address_func = &generic_print_address;
	disasm_obj.fprintf_func = &generic_fprintf;
        disasm_obj.stream = (void *)op->asm_buf;
        ...
}
```

### DIFF
--- a/libr/asm/arch/include/disas-asm.h
+++ b/libr/asm/arch/include/disas-asm.h
@@ -280,7 +280,7 @@ extern int print_insn_little_mips	(bfd_vma, disassemble_info *);
 extern int print_insn_little_nios2	(bfd_vma, disassemble_info *);
 extern int print_insn_little_powerpc	(bfd_vma, disassemble_info *);
 extern int print_insn_riscv		(bfd_vma, disassemble_info *);
-extern int print_insn_little_score      (bfd_vma, disassemble_info *); 
+extern int print_insn_little_score      (bfd_vma, disassemble_info *);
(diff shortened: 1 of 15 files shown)


radare:
As long as all gnu plugins share the same structure we can have a .inc file implementing all the common part or provide some macros to avoid repeating code.

I have plans for another refactoring to merge asm and anal, and there are plans to move some of those to extras too. But maybe is better to wait for future releases to do that

radare:
-350LOC !

pelijah:
What does it mean?

radare:
See the RAsm+RAnal project in github projects for more details

PR #11631, opened 2018-09-23
linux_heap_glibc: debug heap at core file

morbith-dqtz:
also avoid using symbols assuming  the max size for global_max_fast


morbith-dqtz:
Some error at 32 bits, closing for test

PR #11632, opened 2018-09-23
linux_heap_glibc: debug heap at core file 

morbith-dqtz:
and avoid using symbols for global_max_fast assuming max size for the arch
fix r_malloc_state_64

### DIFF
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -86,19 +86,27 @@ static bool GH(update_main_arena)(RCore *core, GHT m_arena, MallocState *main_ar
 	return true;
 }
 
-static void GH(update_global_max_fast)(RCore *core, GHT g_max_fast, GHT *global_max_fast) {
-	(void)r_io_read_at (core->io, g_max_fast, (ut8 *)global_max_fast, sizeof (GHT));
-}
-
 static void GH(get_brks)(RCore *core, GHT *brk_start, GHT *brk_end) {
-	RListIter *iter;
-	RDebugMap *map;
-	r_debug_map_sync (core->dbg);
-	r_list_foreach (core->dbg->maps, iter, map) {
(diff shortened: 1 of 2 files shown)


libr/core/linux_heap_glibc.c
```diff
@@ -86,19 +86,27 @@ static bool GH(update_main_arena)(RCore *core, GHT m_arena, MallocState *main_ar
...
+			if (strstr (map->name, "[heap]")) {
+				*brk_start = map->addr;
+				*brk_end = map->addr_end;
+				break;
+			}
+		}
+	} else {
+		RIOSection *s;
+		SdbListIter *iter1, *iter2;
+		ls_foreach_safe (core->io->sections, iter1, iter2, s) {
```
ret2libc: why did you use `ls_foreach_safe`? This is usually used when you need to delete elements while iterating over the list. Usually, you can just use `ls_foreach`

morbith-dqtz: Because were I read to base the code was done this way. I'll change to ls_foreach_safe. Thx for the tip.

I don't understand what you mean with not tested, I did with centos 32/64, debian 32/64 and gentoo 32/64 with more than one test case, and it works well. Any further test could be grate.

This code intends to add the ability to debug the heap from a core file, you could check what I mean in this asciinema : 

https://asciinema.org/a/R2yTyYY1zCcsuo6iuKEnNxBJK

Also intends to avoid the use of debug library and symbol offsets to parse the main_arena and the heap.

ret2libc:
What I mean by not tested is that we don't have a lot of tests in https://github.com/radare/radare2-regressions . It would be great if you could add a test there, simply doing what you are manually trying on your systems.

XVilka:
Agree with @radare. Eventually we will work on testing debugger, but for now it is better to stick with core files analysis.

PR #11633, opened 2018-09-23
Remove three unused functions

nsajko:


libr/core/cmd.c
```diff
@@ -1322,27 +1322,6 @@ static int cmd_pipein(void *user, const char *input) {
 	return 0;
 }
 
-static void task_test(RCore *core, int usecs) {
```
ret2libc: since you are removing this, please also remove the `#if 0` code in `core/cmd.c` where task_test is "used".

radare:
@thestr4ng3r can you check this test function? Do you want it to have it in the unit tests?

radare:
About the disasm ones... i guess its fine, but will be good to git blame to see at which point we stopped using them and if thats not introducing any regression maybe @kazarmy can answer that

kazarmy:
sorry no time to look at this at the moment ... i prefer `#if 0` instead of removing the functions outright

radare:
Can you update the pr according to these comments? Thanks!

nsajko:
> Can you update the pr according to these comments? Thanks!

I will wait until @kazarmy or someone else has time to decide if the disasm.c functions are needed or should be removed.

radare:
Just if0. He will be offline a bunch of days. We have more if0 code to review and decide probably in another pr

nsajko:
> Just if0.

Done.

PR #11641, opened 2018-09-24
Remove the subarch callback from the RAsmPlugin

radare:


radare:
afaik nobody is using that

PR #11643, opened 2018-09-24
libr/asm/arch/msp430: optimize some array lengths

nsajko:
Also kills MSP430_INSTR_MAXLEN, fixes cmd->operands being uninitialized
in the case of an invalid opcode, and some code demultiplication is
done.


nsajko:
Does somebody understand the OS X errors? (The commit fails only on OS X.)

nsajko:
Oh, I can not use a macro in snprintf arguments, because of how snprintf is implemented as a macro on Darwin. ;(

libr/asm/arch/msp430/msp430_disas.c
```diff
@@ -98,79 +98,79 @@ static int decode_emulation(ut16 instr, struct msp430_cmd *cmd)
 	opcode = get_twoop_opcode(instr);
 
 	if (opcode == MSP430_ADDC && as == 0 && src == MSP430_R3) {
-		snprintf(cmd->instr, MSP430_INSTR_MAXLEN - 1, "%s", bw ? "adc.b" : "adc");
-		snprintf(cmd->operands, MSP430_INSTR_MAXLEN - 1, "r%d", dst);
+		snprintf(cmd->instr, R_ARRAY_SIZE (cmd->instr), "%s", bw ? "adc.b" : "adc");
```
radare: imho the only reason to use R_ARRAY_SIZE instead of sizeof() is to ensure that the type is an array and not a pointer. but the R_ARRAY_SIZE macro is not checking for this at all.

Anyway, feels like using snprintf so many times is kind of wrong.

and i also see that the whole file misses a space before the ( in the calls

nsajko: Well, in this case sizeof and R_ARRAY_SIZE are equivalent, because the size of the array element is 1; I can change it to sizeof if you want.

nsajko: > Anyway, feels like using snprintf so many times is kind of wrong.
>
> and i also see that the whole file misses a space before the ( in the calls

Agreed, but that is a matter for another pull request.

radare:
yeah. Shorter and cleaner imho. So its not hiding what its doing

PR #11645, opened 2018-09-25
Enabling adding the menu items more intuitively.

Vane11ope:
Before you had to pass the parent menu to add the child menu items to the menu tree like parent->sub[0] and not very readable but now, for example, if you wanna hang the new menu item like "Default" under the parent menus such as "File" -> "Load Layout" then you can pass the parent menu like "File.Load Layout" so that the codhe gets more readable.

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -173,14 +173,14 @@ static int gameCb(void *user);
 static int licenseCb(void *user);
 static int aboutCb(void *user);
 static int quitCb(void *user);
-static void addMenu(RPanelsMenuItem *parent, const char *name, RPanelsMenuCallback cb);
+static void addMenu(RCore *core, const char *parent, const char *name, RPanelsMenuCallback cb);
 static void removeMenu(RPanels *panels);
 static int file_history_up(RLine *line);
 static int file_history_down(RLine *line);
 static char *getPanelsConfigPath();
 static bool init(RCore *core, RPanels *panels, int w, int h);
 static void initSdb(RPanels *panels);
-static bool initPanelsMenu(RPanels *panels);
+static bool initPanelsMenu(RCore *core);
 static bool initPanels(RCore *core, RPanels *panels);
 static RStrBuf *drawMenu(RPanelsMenuItem *item);
 static void moveMenuCursor(RPanelsMenu *menu, RPanelsMenuItem *parent);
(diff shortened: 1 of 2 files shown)


libr/include/r_cons.h
```diff
@@ -1037,6 +1037,7 @@ typedef struct r_panels_t {
 	bool isZoom;
 	RPanelsMenu *panelsMenu;
 	Sdb *db;
+	Sdb *mdb;
```
ret2libc: Why don't you just use a `SdbHash` (which is an hashtable)? This is what, it seems, you really need here. You don't need all the things of `sdb`, but you just want to store a number for a given string key.

Vane11ope: woo thanks and that's exactly it.

...tbh i didn't even know such a thing, my bad. i will fix this using that shortly later :)

ret2libc: It's still not perfect in its implementation and API (it only supports `char *` as key for now) but it will be better soon :)

Vane11ope: cool and that should be enough for me anyway here :)

Vane11ope:
done.

libr/core/panels.c
```diff
@@ -1910,17 +1910,27 @@ static int openMenuCb (void *user) {
...
 
-static void addMenu(RPanelsMenuItem *parent, const char *name, RPanelsMenuCallback cb) {
+static void addMenu(RCore *core, const char *parent, const char *name, RPanelsMenuCallback cb) {
+	RPanels *panels = core->panels;
 	RPanelsMenuItem *item = R_NEW0 (RPanelsMenuItem);
+	RPanelsMenuItem *p_item;
+	if (parent) {
+		ut64 addr = r_num_math (core->num, sdb_ht_find (panels->mht, parent, NULL));
+		p_item = (RPanelsMenuItem *)addr;
+		sdb_ht_insert (panels->mht, sdb_fmt ("%s.%s", parent, name), sdb_fmt ("%p", item));
```
ret2libc: sorry the naming of the api/data structures is really confusing... you should use the `ht_*` API defined in `sdb/src/ht.h`, otherwise you'll end up using sdb anyway :) Sorry about it, anyway, at least you know now!

Vane11ope:
hopefully this time it should satisfy your criteria :)

PR #11654, opened 2018-09-25
configure: unset CAPSTONE_* vars when it is not used (fix #10266)

### LINKED ISSUE #10266: Broken r_core.pc causes oddities (crashes, wrong analysis) in R2 library user programs
### Work environment

| OS/arch/bits (mandatory)                             | Debian testing x86_64
| Architecture/bits of the file (mandatory)            | ELF, x86/64
| r2 -v full output, **not truncated** (mandatory)     | radare2 2.7.0-git 18308 @ linux-x86-64 git.2.6.0-132-gcc3a5ddfe commit: cc3a5ddfe7ee5a533a6aa78bda73f529fdcadff6 build: 2018-06-05__14:43:50

### Expected behavior

I have a program that depends on the _r_core_ library and uses _pkg-config_ when building. I expected to load the provided binary, perform analysis on it and not crash.

### Actual behavior

My program crashes when analyzing the provided binary (while the R2 CLI does not).

- Make a Radare2 user installation via `sys/user.sh` (I used a different `HOME` because I wanted it to be installed elsewhere).
- Download [my_program.zip](https://github.com/radare/radare2/files/2072526/my_program.zip) and extract it somewhere. It contains:
  - `binary`: A binary that my program will be analyzing (via the `r_core` library).
  - `env.sh`: Environment specifying R2 installation prefix (needs to be changed).
  - `build.sh`: Build script for my program. It uses pkg-config files from the R2 installation.
  - `run.sh`: Run script for my program.
  - `Main.cpp`: Source code of my program that sets up R2 core, loads `binary` and calls `aaa` for auto-analysis.
- Change `R2_INSTALL_PREFIX` in `env.sh`.
- Run `build.sh`.
- Run `run.sh`.
- Crash.

- GDB backtrace: [gdb_backtrace.txt.zip](https://github.com/radare/radare2/files/2072566/gdb_backtrace.txt.zip)
- The R2 CLI seems to work fine when analyzing the binary.
- This was originally reported to the Cutter project: https://github.com/radareorg/cutter/issues/512. Cutter is also affected by this issue.

### Solution
I have actually tracked this down to `r_core.pc` (pkg-config file) adding dependency on my system capstone via _-lcapstone_ for some reason. I believe the built-in capstone from Radare2 and the system capstone forced to be linked against my program were somehow conflicting. Removing capstone from `r_core.pc` solves this issue.

ret2libc:
Signed-off-by: Riccardo Schirone <sirmy15@gmail.com>


configure.acr
```diff
@@ -34,6 +34,8 @@ IFAND WITH_CAPSTONE HAVE_PKGCFG_CAPSTONE {
 	USE_CAPSTONE = 1 ;
 }{
 	USE_CAPSTONE = 0 ;
+	CAPSTONE_CFLAGS = '' ;
```
radare: shouldnt this be done by acr itself? in the PKGCFG directive? it will override user defined env vars, but at least it will be consistent

ret2libc: i have no idea :) i don't know acr, but i'm not sure how PKGCFG could do this. The problem is: you have capstone installed, but you still want to use embedded capstone.

I *think* this means that:
1) PKGCFG finds that capstone is installed, so it sets CAPSTONE_CFLAGS and CAPSTONE_LDFLAGS
2) WITH_CAPSTONE is false, so it goes in the else case and it sets USE_CAPSTONE=0, but still, CAPSTONE_CFLAGS and CAPSTONE_LDFLAGS have been already set.

radare:
fixed in https://github.com/radare/acr/commit/4b1b07e3972ff7101af9580f929f0b7f416952aa

ret2libc:
The acr patch doesn't fix the problem here. If you do have capstone installed, but you don't want to use it (e.g. `--without-syscapstone`), with your patch CAPSTONE_CFLAGS and CAPSTONE_LDFLAGS would be set anyway. With my patch they would be unset in case USE_CAPSTONE=0

radare:
you may split the logic of having capstone, using capstone and having capstone cflags from pkgconfig and which flags use for capstone instead. so, this is making all that stuff “const” instead of overwriting their values

radare:
so, use PKGCFG_CS_CFLAGS, USER_CS_CFLAGS and CS_CFLAGS for example

ret2libc:
@radare please look at this now and see if you like it.

radare:
Better rename to WITH_SYSCAPSTONE. Because the other way implies it will build r2 against cs or not. And we dont have such option yet

ret2libc:
> Better rename to WITH_SYSCAPSTONE

at that point also USE_CAPSTONE should become USE_SYSCAPSTONE, USE_ZIP should be USE_SYSZIP, USE_MAGIC -> USE_SYSMAGIC, etc.... my 2 cents, let's just keep it as it is now. When we'll need to compile without capstone support at all we will review these. (btw, the user-visible text is already very clear about it, so it would be just a rename of a variable that would be inconsistent with all the others)

radare:
Ok then. But we may want to fix that later. I’ll do a new release of acr so u can update the configure script properly.

ret2libc:
Done, taking acr from master. If travis is green I'll merge.

radare:
ive released acr 1.7.0, feel free to update the pr to get the right version in the acr header

PR #11659, opened 2018-09-26
Fix #11649 - ESIL for zeroing RAX when MOV EAX

### LINKED ISSUE #11649: Bug when analyzing access information about registers
### Work environment

| OS/arch/bits (mandatory)                             | Max OS
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.0.0-git 19896 @ darwin-x86-64 git.2.9.0-200-g17f2f3abf commit: 17f2f3abf1fed2f2966a17a19f24c85ce95ae2d9 build: 2018-09-25__12:45:28

### Expected behavior

```bash
> s 0x00400aa0
> pd 1
    0x00400aa0      b8bf406000     mov eax, 0x6040bf
> aear # should be no result
```

### Actual behavior

```bash
> s 0x00400aa0
> pd 1
    0x00400aa0      b8bf406000     mov eax, 0x6040bf
> aear # generate wrong results
    rax
```

- Simply type `r2 mcf_strip` 
- Following the steps I mentioned in behavior. 

The bug caused by ESIL, IMO. For ESIL, `mov eax, 0x6040bf` will be translated into `6308031,eax,=,0xffffffff,rax,&=`, which indicates the instruction has read the data from `rax`. But actually the high bits of `rax` are cleared by CPU, without any reading actions. I am not so sure how ESIL is generated, but `0x0,rax,=,6308031,eax,=` might be a better solution.

The binary is attached. Actually, I think there is no need for a specific binary. I guess all these instruction should have this problems.

[mcf_strip.zip](https://github.com/radare/radare2/files/2416231/mcf_strip.zip)

XVilka:
Blind quickfix

### DIFF
--- a/libr/anal/p/anal_x86_cs.c
+++ b/libr/anal/p/anal_x86_cs.c
@@ -741,10 +741,10 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 				src = getarg (&gop, 1, 0, NULL, SRC_AR);
 				dst = getarg (&gop, 0, 0, NULL, DST_AR);
 				const char *dst64 = r_reg_32_to_64 (a->reg, dst);
-				esilprintf (op, "%s,%s,=", src, dst);
 				if (a->bits == 64 && dst64) {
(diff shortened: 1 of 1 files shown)


radare:
Theres a pr already for that

PR #11660, opened 2018-09-26
Clangification

ret2libc:
some random style fixes just for testing the clang-format file :)

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -7002,7 +7002,7 @@ static int cmd_anal_all(RCore *core, const char *input) {
 		break;
 	case 'f': // "aaf"
 	{
-		int analHasnext = r_config_get_i (core->config, "anal.hasnext");
+		const int analHasnext = r_config_get_i (core->config, "anal.hasnext");
 		r_config_set_i (core->config, "anal.hasnext", true);
 		r_core_cmd0 (core, "afr@@c:isq");
 		r_config_set_i (core->config, "anal.hasnext", analHasnext);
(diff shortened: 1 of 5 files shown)


libr/core/cmd_anal.c
```diff
@@ -7084,8 +7086,8 @@ static int cmd_anal_all(RCore *core, const char *input) {
 			// Usefull when running commands with ";" or via r2 -c,-i
 			run_pending_anal (core);
 			dh_orig = core->dbg->h
-					? strdup (core->dbg->h->name)
-					: strdup ("esil");
+					  ? strdup (core->dbg->h->name)
```
radare: Why those little spaces after the tabs? My script was fixing them with a regex

libr/core/cmd_anal.c
```diff
@@ -6999,22 +6999,22 @@ static int cmd_anal_all(RCore *core, const char *input) {
...
-		break; // "aab"
+		break;
 	case 'f': // "aaf"
-		{
-			int analHasnext = r_config_get_i (core->config, "anal.hasnext");
-			r_config_set_i (core->config, "anal.hasnext", true);
-			r_core_cmd0 (core, "afr@@c:isq");
-			r_config_set_i (core->config, "anal.hasnext", analHasnext);
-		}
+	{
```
radare: Not sure if i prefer the indented braces or not..

libr/core/cmd_anal.c
```diff
@@ -6999,22 +6999,22 @@ static int cmd_anal_all(RCore *core, const char *input) {
...
+		int analHasnext = r_config_get_i (core->config, "anal.hasnext");
```
radare: This can be const

libr/core/cmd_print.c
```diff
@@ -4060,13 +4060,13 @@ static int cmd_print(void *data, const char *input) {
 				} else {
 					int ret, bufsz;
 					RAnalOp aop = {
-						0
+						0,
```
radare: This comma is unnecessary

libr/core/cmd_anal.c
ret2libc: I think not indented is better. Otherwise you have 2 indentations for almost nothing. That said, the blocks shouldn't even be needed in a `switch-case`, but radare2 source code is full of super-messy switch/cases... :(

And, again, I'm happy as long as a tool automatically do that :D

radare:
Did you saw the whitelist i did for indenting?

ret2libc:
> Did you saw the whitelist i did for indenting?

yes, what about it? Do you want me to check this clang format on those files?

radare:
Well we should increase that list by adding the files we have verified that can be indented properly. At some point we can switch to a blacklist to skip all the gnu code and such

PR #11661, opened 2018-09-26
StepIn and StepOver works properly as a menu item

Vane11ope:
For now Debug->StepIn || StepOver does not work properly in the menu in the Panels mode and now it does with this.


Vane11ope:
So the commit message for this should be just 

"Fixed a few broken menu items"

PR #11662, opened 2018-09-26
Distinguish C/C++

devnexen:
More based on presence of mangled naming.


radare:
Lgtm

Maijin:
That's not correct though, the test should have been added here https://github.com/radare/radare2/blob/master/libr/bin/mangling/cxx.c, can you make the move @devnexen ?

PR #11663, opened 2018-09-26
 Detect function prefixed with __libc_ and __GI_ 

### LINKED ISSUE #11647: Function starting with __GI_ and __libc_ to get function detection
### Work environment

| OS/arch/bits (mandatory)                             | Mac OS 64
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.0.0-git 19672 @ darwin-x86-64 git.2.9.0-165-g40c076334 commit: 40c076334d5f167a077074836cf208f1b0e715c1 build: 2018-09-20__23:47:30

### Expected behavior

```
bl sym.__GI_perror               ;[3]   ; void perror(const char *s)
bl sym.__libc_perror               ;[3]   ; void perror(const char *s)
```

### Actual behavior

```
bl sym.__GI_perror               ;[3]   
bl sym.__libc_perror               ;[3]  
```

File provided in Private message/telegram (Mirai.zip)

Both prefixes are commonly used in glibc via a #define:

- `__libc_`: https://github.com/lattera/glibc/blob/master/include/libc-symbols.h#L896
- `__GI_`: https://github.com/lattera/glibc/blob/master/include/libc-symbols.h#L521

sivaramaaa:
Fixes #11647 

* After fix , we can detect function arguments for func with prefix __GI_ and __libc_

```
|           0x00018824      3c300be5       str r3, [ptr]
|           0x00018828      80301be5       ldr r3, [nbytes]
|           0x0001882c      0100a0e3       mov r0, 1       ; int fd
|           0x00018830      3c101be5       ldr r1, [ptr]   ; void *ptr
|           0x00018834      0320a0e1       mov r2, r3      ; size_t nbytes
|           0x00018838      3d2300eb       bl sym.__GI_write      ; ssize_t write(int fd, void *ptr, size_t nbytes)
```

### DIFF
--- a/libr/anal/d/types-linux.sdb.txt
+++ b/libr/anal/d/types-linux.sdb.txt
@@ -86,6 +86,53 @@ func.prctl.arg.3=unsigned long,v4
 func.prctl.arg.4=unsigned long,v5
 func.prctl.ret=int
 
+sigaction=func
+func.sigaction.args=3
+func.sigaction.arg.0=int,signum
+func.sigaction.arg.1=const struct sigaction *,act
+func.sigaction.arg.2=struct sigaction *,oldact
+func.sigaction.ret=int
+
+select=func
+func.select.args=5
+func.select.arg.0=int,nfds
+func.select.arg.1=fd_set *,readfds
+func.select.arg.2=fd_set *,writefds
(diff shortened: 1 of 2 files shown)


radare:
Lgtm

PR #11668, opened 2018-09-26
Use internal constant to update flag register for x86

ZhangZhuoSJTU:
Original code uses both normal constant number and internal constant number to update flag register for x86. It makes difficulty for some analysis based on ESIL. This is a quick fix to standardize the usage.

### DIFF
--- a/libr/anal/p/anal_x86_cs.c
+++ b/libr/anal/p/anal_x86_cs.c
@@ -811,7 +811,7 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 			src = getarg (&gop, 1, 0, NULL, SRC_AR);
 			dst_r = getarg (&gop, 0, 0, NULL, DST_R_AR);
 			dst_w = getarg (&gop, 0, 1, NULL, DST_W_AR);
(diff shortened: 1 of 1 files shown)


libr/anal/p/anal_x86_cs.c
```diff
@@ -859,16 +859,16 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 		esilprintf (op, "eax,rax,=,31,rax,>>,?{,0xffffffff00000000,rax,|=,}");
 		break;
 	case X86_INS_AAA:
-		esilprintf (op, "0,cf,=,0,af,=,9,al,>,?{,10,al,-=,1,ah,+=,1,cf,=,1,af,=,}");
+		esilprintf (op, "$0,cf,=,$0,af,=,9,al,>,?{,10,al,-=,1,ah,+=,$1,cf,=,$1,af,=,}");
```
radare: why this $0 -> $1. change? looks wrong to me

ZhangZhuoSJTU: I thought this is `0 -> $0` and `1 -> $1`. Did I do some mistypes?

radare:
Ah right its fine. Numbers starting with $ doesnt have side effects. So lgtm

PR #11670, opened 2018-09-27
fix axt when asm.var.sub=false

mytbk:


ret2libc:
Thanks for your contribution! If you'd like to add a test for this in radare2-regressions repository it would be great :)

PR #11678, opened 2018-09-27
Fix #11377 - Fix the way we find the entrypoint in ELF

### LINKED ISSUE #11377: r2 mis-identifies binary entrypoint
# This template is meant for bug reports, if you have a feature request, please be as descriptive as possible and delete the template

Make sure you are testing using the latest git version of radare2 before submitting any issue.

*If you would like to report a bug, please fill the template bellow*

### Work environment

| OS/arch/bits (mandatory)                             | Debian 9.5 amd64
| Architecture/bits of the file (mandatory)            | ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=a5b3decf60235cc1bae86e7a9a02d3ba853729d4, not stripped
| r2 -v full output, **not truncated** (mandatory)     | radare2 2.9.0-git 1 @ linux-x86-64 git.2.9.0-git commit: 5c0eaa71f1f7c20f05bf0302eecd8da3b8d7663d build: 2018-09-01__01:13:41

### Expected behavior
- `s entry0`
- View/print/whatever the actual entrypoint

### Actual behavior
- `s entry0`
- nothing there, because entrypoint is mis-defined

- `r2 ./ls.patched`
- `ie` -- This shows incorrect entrypoint

rabin2 and readelf agree on the right entrypoint

```
$ readelf -h ./ls.patched
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              EXEC (Executable file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x87f02b
  Start of program headers:          517932 (bytes into file)
  Start of section headers:          882472 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         10
  Size of section headers:           64 (bytes)
  Number of section headers:         40
  Section header string table index: 37

$ rabin2 -H ./ls.patched
0x00000000  ELF64       0x464c457f
0x00000010  Type        0x0002
0x00000012  Machine     0x003e
0x00000014  Version     0x00000001
0x00000018  Entrypoint  0x0087f02b
0x00000020  PhOff       0x0007e72c
0x00000028  ShOff       0x000d7728
```

r2 console does not
```
[0x0067fff1]> ie
[Entrypoints]
vaddr=0x0068002b paddr=0x0008002b baddr=0x00400000 laddr=0x00000000 haddr=0x00000018 type=program
```

[ls.patched.zip](https://github.com/radare/radare2/files/2346286/ls.patched.zip)

radare:


XVilka:
Breaks Thumb binaries.

ret2libc:
@radare please see https://github.com/radare/radare2/pull/11683 . I think that's the right fix for the issue though there are other errors in other tests. but i think we need to fix the other things, the function seems right as i fixed it.

PR #11679, opened 2018-09-28
include/r_types: remove R_NOTNULL

### LINKED ISSUE #11553: Kill R_NOTNULL

davidpolverari:
First step of refactoring. load_bytes() first parameter may not be
needed because of bf, depending on its use cases. Needs further testing.

Closes #11553

### DIFF
--- a/libr/bin/obj.c
+++ b/libr/bin/obj.c
@@ -57,9 +57,8 @@ R_API RBinObject *r_bin_object_new(RBinFile *binfile, RBinPlugin *plugin, ut64 b
 		if (sz < bsz) {
 			bsz = sz;
 		}
-		o->bin_obj = plugin->load_bytes (binfile, bytes + offset, sz,
-						 loadaddr, sdb);
-		if (!o->bin_obj) {
+		if (!plugin->load_bytes (binfile, &o->bin_obj, bytes + offset, sz,
+					 loadaddr, sdb)) {
 			bprintf (
 				"Error in r_bin_object_new: load_bytes failed "
 				"for %s plugin\n",
--- a/libr/bin/p/bin_art.c
+++ b/libr/bin/p/bin_art.c
@@ -63,19 +63,20 @@ static Sdb *get_sdb(RBinFile *bf) {
 	return ao? ao->kv: NULL;
 }
 
-static void *load_bytes(RBinFile *bf, const ut8 *buf, ut64 sz, ut64 la, Sdb *sdb){
+static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut64 la, Sdb *sdb){
 	ArtObj *ao = R_NEW0 (ArtObj);
 	if (!ao) {
-		return NULL;
+		return false;
 	}
 	ao->kv = sdb_new0 ();
 	if (!ao->kv) {
 		free (ao);
-		return NULL;
+		return false;
 	}
 	art_header_load (&ao->art, bf->buf, ao->kv);
 	sdb_ns_set (sdb, "info", ao->kv);
-	return ao;
+	*bin_obj = ao;
+	return true;
 }
 
 static bool load(RBinFile *bf) {
--- a/libr/bin/p/bin_bf.c
+++ b/libr/bin/p/bin_bf.c
@@ -5,8 +5,8 @@
 #include <r_lib.h>
 #include <r_bin.h>
 
-static void *load_bytes(RBinFile *bf, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
-	return R_NOTNULL;
+static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
+	return true;
 }
 
 static bool load(RBinFile *bf) {
--- a/libr/bin/p/bin_fs.c
+++ b/libr/bin/p/bin_fs.c
@@ -53,18 +53,15 @@ static bool check_bytes(const ut8 *buf, ut64 length) {
 	return p != NULL;
 }
 
-static void * load_bytes(RBinFile *bf, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
-	if (check_bytes (buf, sz)) {
-		return R_NOTNULL;
-	}
-	return NULL;
+static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
+	return check_bytes (buf, sz);
 }
 
 static bool load(RBinFile *bf) {
 	const ut8 *bytes = bf ? r_buf_buffer (bf->buf) : NULL;
 	ut64 sz = bf ? r_buf_size (bf->buf): 0;
 	ut64 la = (bf && bf->o) ? bf->o->loadaddr: 0;
-	return load_bytes (bf, bytes, sz, la, bf? bf->sdb: NULL) != NULL;
+	return load_bytes (bf, &bf->o->bin_obj, bytes, sz, la, bf? bf->sdb: NULL);
 }
 
 static int destroy(RBinFile *bf) {
--- a/libr/bin/p/bin_java.c
+++ b/libr/bin/p/bin_java.c
@@ -74,22 +74,22 @@ static Sdb *get_sdb(RBinFile *bf) {
 	return NULL;
 }
 
-static void *load_bytes(RBinFile *bf, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
-	struct r_bin_java_obj_t *bin_obj = NULL;
+static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
+	struct r_bin_java_obj_t *tmp_bin_obj = NULL;
 	RBuffer *tbuf = NULL;
-	void *res = NULL;
 	if (!buf || sz == 0 || sz == UT64_MAX) {
-		return NULL;
+		return false;
 	}
 	tbuf = r_buf_new ();
 	r_buf_set_bytes (tbuf, buf, sz);
-	res = bin_obj = r_bin_java_new_buf (tbuf, loadaddr, sdb);
-	add_bin_obj_to_sdb (bin_obj);
+	tmp_bin_obj = r_bin_java_new_buf (tbuf, loadaddr, sdb);
+	*bin_obj = tmp_bin_obj;
+	add_bin_obj_to_sdb (tmp_bin_obj);
 	if (bf && bf->file) {
-		bin_obj->file = strdup (bf->file);
+		tmp_bin_obj->file = strdup (bf->file);
 	}
 	r_buf_free (tbuf);
-	return res;
+	return true;
 }
 
 static bool load(RBinFile *bf) {
@@ -102,7 +102,7 @@ static bool load(RBinFile *bf) {
 		return false;
 	}
 
-	bin_obj = load_bytes (bf, bytes, sz, bf->o->loadaddr, bf->sdb);
+	load_bytes (bf, (void **) &bin_obj, bytes, sz, bf->o->loadaddr, bf->sdb);
 
 	if (bin_obj) {
 		if (!bf->o->kv) {
--- a/libr/bin/p/bin_mbn.c
+++ b/libr/bin/p/bin_mbn.c
@@ -71,15 +71,15 @@ static bool check_bytes(const ut8 *buf, ut64 bufsz) {
 	return false;
 }
 
-static void * load_bytes(RBinFile *bf, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
-	return (void*)(size_t)check_bytes (buf, sz);
+static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
+	return check_bytes (buf, sz);
 }
 
 static bool load(RBinFile *bf) {
 	if (bf && bf->buf) {
 		const ut8 *bytes = r_buf_buffer (bf->buf);
 		ut64 sz = r_buf_size (bf->buf);
-		return load_bytes (bf, bytes, sz, bf->o->loadaddr, bf->sdb) != NULL;
+		return load_bytes (bf, &bf->o->bin_obj, bytes, sz, bf->o->loadaddr, bf->sdb);
 	}
 	return false;
 }
--- a/libr/bin/p/bin_menuet.c
+++ b/libr/bin/p/bin_menuet.c
@@ -62,15 +62,15 @@ static bool check_bytes(const ut8 *buf, ut64 length) {
 	return false;
 }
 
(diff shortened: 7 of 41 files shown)


XVilka:
Note that it will also require patching the Python r2lang interface and r2book documentation: 
 - https://github.com/radare/radare2-bindings/blob/master/libr/lang/p/python/bin.c 
 - https://github.com/radare/radare2book/blob/master/plugins/dev-bin.md
 - https://github.com/radare/radare2book/blob/master/plugins/python.md

libr/bin/obj.c
```diff
@@ -55,8 +55,11 @@ R_API RBinObject *r_bin_object_new(RBinFile *binfile, RBinPlugin *plugin, ut64 b
 		if (sz < bsz) {
 			bsz = sz;
 		}
-		o->bin_obj = plugin->load_bytes (binfile, bytes + offset, sz,
+		//o->bin_obj = plugin->load_bytes (binfile, bytes + offset, sz,
```
ret2libc: please just remove old code

libr/bin/p/bin_art.c
```diff
@@ -63,19 +63,20 @@ static Sdb *get_sdb(RBinFile *bf) {
...
 	ao->kv = sdb_new0 ();
 	if (!ao->kv) {
 		free (ao);
-		return NULL;
+		return false;
 	}
 	art_header_load (&ao->art, bf->buf, ao->kv);
 	sdb_ns_set (sdb, "info", ao->kv);
-	return ao;
+	bin_obj = ao;
```
ret2libc: what's the purpose of this line?

davidpolverari: The old function prototype returned a void* pointer to a bin_obj. I changed the prototype to return a bool, and the pointer is now returned in the first parameter of the function.

I thought about not adding that parameter and just use the bf parameter to set bf->o->bin_obj, as in most cases, that is the one pointer that is set by the function. I didn’t do it because I wasn’t sure whether there were other use cases that could differ.

This line was meant to keep the old behavior.

ret2libc: please see my other comment about pointer uses. This line is not doing anything except changing the value of the *local variable* `bin_obj`, which is not reflected outside this function.

davidpolverari: This is all in response to issue #11553.

ret2libc: I see, thanks! Yeah so either you add a `bool *loaded` to the arguments and you keep `void *` as return argument, or you make `bool` the return value and you add a `void **` as argument. I wouldn't set `o->bin_obj` directly in the `load_bytes` function because in some cases you may not need that.

libr/bin/p/bin_bf.c
```diff
@@ -5,8 +5,8 @@
 #include <r_lib.h>
 #include <r_bin.h>
 
-static void *load_bytes(RBinFile *bf, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
-	return R_NOTNULL;
+static bool load_bytes(void **bin_obj, RBinFile *bf, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
```
radare: RBinFile should be the first argument, because it's the "this" of those methods. having this bin_obj or `bo` as 2nd argument looks good to me

davidpolverari: Ok. I'll swap their order then. Thanks!

davidpolverari: No problem. I'll swap the first 2 args order, rebase and push it again. Thanks for the review!

davidpolverari:
No problem. I'll swap the first 2 args order, rebase and push it again. Thanks for the review!

libr/bin/obj.c
davidpolverari: Old code removed.

libr/bin/p/bin_art.c
davidpolverari: Added void ** parameter to load_bytes().

davidpolverari:
I updated the PR, rebased and pushed it, but Travis is complaining and I don't have a clue why...

libr/bin/p/bin_fs.c
```diff
@@ -53,18 +53,15 @@ static bool check_bytes(const ut8 *buf, ut64 length) {
...
+static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
+	return check_bytes (buf, sz);
 }
 
 static bool load(RBinFile *bf) {
 	const ut8 *bytes = bf ? r_buf_buffer (bf->buf) : NULL;
 	ut64 sz = bf ? r_buf_size (bf->buf): 0;
 	ut64 la = (bf && bf->o) ? bf->o->loadaddr: 0;
-	return load_bytes (bf, bytes, sz, la, bf? bf->sdb: NULL) != NULL;
+	return load_bytes (bf, &bf->o->bin_obj, bytes, sz, la, bf? bf->sdb: NULL);
```
ret2libc: ?

libr/bin/p/bin_java.c
```diff
@@ -74,22 +74,23 @@ static Sdb *get_sdb(RBinFile *bf) {
 	return NULL;
 }
 
-static void *load_bytes(RBinFile *bf, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
-	struct r_bin_java_obj_t *bin_obj = NULL;
+static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
+	struct r_bin_java_obj_t *tmp_bin_obj = NULL;
 	RBuffer *tbuf = NULL;
-	void *res = NULL;
+	//void *res = NULL;
```
ret2libc: do not comment, just remove the line.

libr/bin/p/bin_java.c
```diff
@@ -102,7 +103,7 @@ static bool load(RBinFile *bf) {
 		return false;
 	}
 
-	bin_obj = load_bytes (bf, bytes, sz, bf->o->loadaddr, bf->sdb);
+	load_bytes (bf, (void **) &bin_obj, bytes, sz, bf->o->loadaddr, bf->sdb);
```
ret2libc: is the cast really needed?

libr/bin/p/bin_mbn.c
```diff
@@ -71,15 +71,15 @@ static bool check_bytes(const ut8 *buf, ut64 bufsz) {
...
+static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
+	return check_bytes (buf, sz);
 }
 
 static bool load(RBinFile *bf) {
 	if (bf && bf->buf) {
 		const ut8 *bytes = r_buf_buffer (bf->buf);
 		ut64 sz = r_buf_size (bf->buf);
-		return load_bytes (bf, bytes, sz, bf->o->loadaddr, bf->sdb) != NULL;
+		return load_bytes (bf, &bf->o->bin_obj, bytes, sz, bf->o->loadaddr, bf->sdb);
```
ret2libc: ?

libr/bin/p/bin_menuet.c
```diff
@@ -62,15 +62,15 @@ static bool check_bytes(const ut8 *buf, ut64 length) {
...
+static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
+	return check_bytes (buf, sz);
 }
 
 static bool load(RBinFile *bf) {
 	const ut8 *bytes = bf ? r_buf_buffer (bf->buf) : NULL;
 	ut64 sz = bf ? r_buf_size (bf->buf): 0;
 	ut64 la = (bf && bf->o) ? bf->o->loadaddr: 0;
-	return load_bytes (bf, bytes, sz, la, bf? bf->sdb: NULL) != NULL;
+	return load_bytes (bf, &bf->o->bin_obj, bytes, sz, la, bf? bf->sdb: NULL);
```
ret2libc: ?

libr/bin/p/bin_p9.c
```diff
@@ -13,15 +13,15 @@ static bool check_bytes(const ut8 *buf, ut64 length) {
...
+static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb){
+	return check_bytes (buf, sz);
 }
 
 static bool load(RBinFile *bf) {
 	const ut8 *bytes = bf? r_buf_buffer (bf->buf): NULL;
 	ut64 sz = bf? r_buf_size (bf->buf): 0;
 	ut64 la = (bf && bf->o)? bf->o->loadaddr: 0;
-	return load_bytes (bf, bytes, sz, la, bf? bf->sdb: NULL);
+	return load_bytes (bf, &bf->o->bin_obj, bytes, sz, la, bf? bf->sdb: NULL);
```
ret2libc: ?

libr/bin/obj.c
```diff
@@ -55,8 +55,9 @@ R_API RBinObject *r_bin_object_new(RBinFile *binfile, RBinPlugin *plugin, ut64 b
 		if (sz < bsz) {
 			bsz = sz;
 		}
-		o->bin_obj = plugin->load_bytes (binfile, bytes + offset, sz,
+		plugin->load_bytes (binfile, &o->bin_obj, bytes + offset, sz,
```
ret2libc: you have to get the `bool result` from load_bytes and check it at https://github.com/radare/radare2/pull/11679/files#diff-ce3aeea66b7b1f6e50aaa345b1963acdR61, instead of checking o->bin_obj.

libr/bin/p/bin_fs.c
davidpolverari: Problem is that `load_bytes()` is used through a function pointer declared in `r_bin.h`. Thus, I _think_ I can't use different function signatures on their definitions on `bin_*.c`, even though they are static, because it seems to offer some kind of polymorphism to the binary loading logic.

libr/bin/p/bin_java.c
davidpolverari: Ok.

libr/bin/p/bin_java.c
davidpolverari: I though it was needed, because the code (that I kept as close as possible to the original), uses `struct r_bin_java_obj_t *bin_obj = NULL` and `load_bytes()` expects a RBinFile*.

libr/bin/p/bin_mbn.c
davidpolverari: I didn't change the signature of `load()`. So I took advantage of `load_bytes()` already returning a bool and just returning it to `load()`'s caller, instead of the previous definition where the result of `void *` returned by `load_bytes()` was checked against NULL, returning a bool using the old behavior.

libr/bin/p/bin_menuet.c
davidpolverari: Same as above.

libr/bin/p/bin_p9.c
davidpolverari: Still the same as `bin_p9.c` and `bin_menuet.c`.

libr/bin/obj.c
davidpolverari: Ok, I'll change that. I didn't change it before because I was avoiding to make those changes in this PR, that already changed a lot, and to keep most of the code close to the original.

ret2libc: Yes but this is causing the issues you are seeing in travis

libr/bin/p/bin_fs.c
ret2libc: I'm saying that you shouldn't pass bf->o->bin_obj (and you should check for nulls in load_bytes before assigning it). In the original code, bf->o->bin_obj was not set, in your code it *may* be  (if check_bytes will set it in the future)

libr/bin/p/bin_java.c
ret2libc: Oh i see.

radare:
Ping.

libr/bin/p/bin_fs.c
davidpolverari: As `load_bytes()` doesn't change `bf->o->bin_obj` and it probably will once it grows to do more than call `check_bytes()`, isn't it reasonable leaving it there in order to ease further extension?

davidpolverari:
I will just rebase and send the changes.

libr/bin/obj.c
davidpolverari: I did change it, but travis is still red.

ret2libc:
> I did change it, but travis is still red.

Did you try to run the tests that are failing on your machine? In particular, if you do 
```
$ r2 radare2-regressions/bins/mdmp/hello.dmp
```
you'll see there's a Segmentation fault. If you attach gdb to it, you'll quickly see that `bf->o->bin_obj` is NULL in bin_mdmp.c and that's because you didn't set it in `load-bytes`.

davidpolverari:
Thanks, @ret2libc , I fixed that issue and now it's green. I was lost before on reading the radare2-regressions output, and I didn't see before that the error was in bin_mdmp.c. Now learned one more lesson :).

libr/bin/p/bin_fs.c
davidpolverari: I understood about null checking before setting the value of `bf->o->bin_obj`. But I didn't understand about not passing it `load_bytes()`, as we have to keep the function signature.

radare:
Green!

libr/bin/obj.c
```diff
@@ -57,9 +58,10 @@ R_API RBinObject *r_bin_object_new(RBinFile *binfile, RBinPlugin *plugin, ut64 b
 		if (sz < bsz) {
 			bsz = sz;
 		}
-		o->bin_obj = plugin->load_bytes (binfile, bytes + offset, sz,
+		success = plugin->load_bytes (binfile, &o->bin_obj, bytes + offset, sz,
```
radare: no need to have a variable that is only used once you can just` if (!plugin->load_bytes( ...)`

davidpolverari: I'm going to remove it, rebase and send the updated PR.

libr/bin/p/bin_fs.c
davidpolverari: I still didn't change those functions to check `bf->o->bin_obj` for null because I thought this PR is already too big, and I was afraid of changing too much at once. I can do it in a future PR or in this one, if you deem appropriate.

ret2libc: ok no problem

davidpolverari: No problem if I defer that or no problem if I change more? Should I wait for this to be merged and work on those enhancements on a next PR or keep on working on it?

davidpolverari: Ok, then I'll prepare another PR to further refactor those things after the merge of this one.

radare:
n fact... the binobj pointer should be inside the binfile, isnt? because i think we can remove one more arg from this function doing this. Imho its a bit confusing that we have binfile, binobj, .etc.. this is something to discuss and address in a separate PR, but wanted to mention it here first :)

radare:
yes please. rbin needs a lot of love. cleanup and refactorings. any reduction or clarification of code is welcome :)

PR #11686, opened 2018-09-28
Fix pae/pade (fix #11677)

### LINKED ISSUE #11677: pa sub commands are buggy
There are a few issues with the `pa` sub commands:

- The help messages are not consistent with the expected behaviour (based on standalone `rasm2` and its help messages).
If I understand the intentions of these commands correctly, and for reasons of clarity and comprehensibility, the help should read something like this:
```
Usage: pa[edD] [asm|hex]  print (dis)assembled
| pa [assembly]    print hexpairs of the given assembly expression
| pad [hexpairs]   print assembly expression from hexpairs
| paD [hexpairs]   print assembly expression from hexpairs and show hexpairs
| pae [assembly]   print ESIL expression of the given assembly expression
| pade [hexpairs]  print ESIL expression from hexpairs
```
- `pa 'mov eax, ebx;pop ebp;nop'` should show the hexpairs like `rasm2 'mov eax, ebx;pop ebp;nop'` does
- `pae 'mov eax, ebx;pop ebp;nop'` should show ESIL expression
- the ESIL expression of `pade 89d85d90` **does not** match the ESIL expression of `rasm2 -E 89d85d90`

### Expected behavior
rasm2_expected

### Actual behavior
r2_pa_bug

jhscheer:
- fixed an issue where `pae`/`pade` only printed the first line of the ESIL expression
- improved help messages for `pa*` sub commands
 
see https://github.com/radare/radare2/issues/11677

### DIFF
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -13,11 +13,11 @@
 
 static const char *help_msg_pa[] = {
 	"Usage: pa[edD]", "[asm|hex]", "print (dis)assembled",
-	"pa", " [assembly]", "print hexpairs of the given assembled instruction",
-	"paD", " [assembly]", "print assembled from hexparis",
-	"pad", " [hexpairs]", "print disassembled from hexpairs",
-	"pae", " [assembly]", "print ESIL after assembling given instruction",
-	"pade", " [hexpairs]", "print ESIL expression assembled ESIL from hexpairs",
+	"pa", " [assembly]", "print hexpairs of the given assembly expression",
+	"paD", " [hexpairs]", "print assembly expression from hexpairs and show hexpairs",
+	"pad", " [hexpairs]", "print assembly expression from hexpairs",
+	"pade", " [hexpairs]", "print ESIL expression from hexpairs",
+	"pae", " [assembly]", "print ESIL expression of the given assembly expression",
 	NULL
 };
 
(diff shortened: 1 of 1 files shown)


ret2libc:
I'm going to review this in the next days. Thanks!

libr/core/cmd_print.c
```diff
@@ -4027,51 +4027,69 @@ static int cmd_print(void *data, const char *input) {
...
+					int printed = 0;
+                    			int bufsz;
+					RAnalOp aop = { 0 };
 					char *hex_arg = calloc (0, strlen (arg));
 					if (hex_arg) {
 						bufsz = r_hex_str2bin (arg + 1, (ut8 *)hex_arg);
-						ret = r_anal_op (core->anal, &aop, core->offset,
-							(const ut8 *)hex_arg, bufsz, R_ANAL_OP_MASK_ESIL);
-						if (ret > 0) {
+						while (printed < bufsz) {
```
ret2libc: this loop seems very very similar to the one for `pae`. Please refactor it in a new function to reduce duplication.

libr/core/cmd_print.c
```diff
@@ -4088,8 +4106,11 @@ static int cmd_print(void *data, const char *input) {
 				}
 				break;
 			}
+			case '?': // "pad"
```
ret2libc: fix `"pad"` -> `"pad?"`

libr/core/cmd_print.c
jhscheer: You are of course right and I also don't like this duplication.
However, for the time being, I would rather leave it like this since:
a) I honestly have no idea where a good place for this function would be
(the same functionality is not only used in this file, but also i.e. here
https://github.com/radare/radare2/blob/d2ed1809d023d26ee055cdbf7d593910afb36e8c/binr/rasm2/rasm2.c#L295
and maybe other places I'm not aware of.
b) Considering that this is by far not the only place that would improve by reducing duplication, I don't feel like going down this rabbit hole now.

libr/core/cmd_print.c
jhscheer: done

libr/core/cmd_print.c
ret2libc: Ok np ;) thanks

libr/core/cmd_print.c
```diff
@@ -13,11 +13,11 @@
...
-	"pa", " [assembly]", "print hexpairs of the given assembled instruction",
-	"paD", " [assembly]", "print assembled from hexparis",
-	"pad", " [hexpairs]", "print disassembled from hexpairs",
-	"pae", " [assembly]", "print ESIL after assembling given instruction",
-	"pade", " [hexpairs]", "print ESIL expression assembled ESIL from hexpairs",
+	"pa", " [assembly]", "print hexpairs of the given assembly expression",
+	"pad", " [hexpairs]", "print assembly expression from hexpairs",
+	"paD", " [hexpairs]", "print assembly expression from hexpairs and show hexpairs",
+	"pae", " [assembly]", "print ESIL expression of the given assembly expression",
+	"pade", " [hexpairs]", "print ESIL expression from hexpairs",
```
radare: alphabetically sort pls

libr/core/cmd_print.c
```diff
@@ -13,11 +13,11 @@
...
 static const char *help_msg_pa[] = {
 	"Usage: pa[edD]", "[asm|hex]", "print (dis)assembled",
-	"pa", " [assembly]", "print hexpairs of the given assembled instruction",
-	"paD", " [assembly]", "print assembled from hexparis",
-	"pad", " [hexpairs]", "print disassembled from hexpairs",
-	"pae", " [assembly]", "print ESIL after assembling given instruction",
-	"pade", " [hexpairs]", "print ESIL expression assembled ESIL from hexpairs",
+	"pa", " [assembly]", "print hexpairs of the given assembly expression",
+	"pad", " [hexpairs]", "print assembly expression from hexpairs",
+	"paD", " [hexpairs]", "print assembly expression from hexpairs and show hexpairs",
```
radare: "print disassembly expression" looks wrong to me. something like "disassemble given hexpairs" is more concise imho

libr/core/cmd_print.c
```diff
@@ -4027,51 +4027,69 @@ static int cmd_print(void *data, const char *input) {
...
+							if (r_anal_op (core->anal, &aop, core->offset,
+								    (const ut8 *)hex_arg + printed, bufsz - printed, R_ANAL_OP_MASK_ESIL) > 0) {
+								const char *str = R_STRBUF_SAFEGET (&aop.esil);
+								r_cons_println (str);
+							} else {
+								eprintf ("Cannot decode instruction\n");
+								break;
+							}
+							if (aop.size < 1) {
+								eprintf ("Invalid\n");
```
radare: invalid wat, better use the same error string as above

libr/core/cmd_print.c
jhscheer: I don't mind the terminology as long as it is consistent over all commands.
I thought combining {"assembled instruction", "disassembled", "after assembling given instruction"} to just "assembly expression" might be a good idea, since there was already the term "ESIL expression".

However, to keep it short & consistent, omitting "expression" and both verbs assemble/disassemble would probably be best.

What do you think about this suggestion:
```
pa [assembly]:   print hexpairs (asm -> hex)
pad [hexpairs]:  print assembly (hex -> asm)
paD [hexpairs]:  print assembly and hexpairs (hex -> asm & hex)
pade [hexpairs]: print ESIL (hex -> esil)
pae [assembly]:  print ESIL (asm -> esil)
```

radare: the shorter the better. btw "print hexpairs" is assemble, and "print assembly" is "disassemble"

radare:
thanks

Issue #11687, opened 2018-09-28
Memory Leak in rabin2 path variable


bannsec:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Ubuntu x86 64
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86/64
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.0.0-git 19571 @ linux-x86-64 git.2.9.0-235-g354a40836 commit: 354a40836aa12df0ed3fb3ffa90e800e6dd49294 build: 2018-09-28__20:26:49

### Expected behavior
No memory leak

### Actual behavior
Memory leak

### Steps to reproduce the behavior 
Simply running rabin2 encounters a memory leak. Specifically, line 577 in `/binr/rabin2/rabin2.c`.

```c
path = r_sys_getenv (R_LIB_ENV);
```

That path variable is malloc'd but never free'd.

radare:
can you send a PR fixing this? its easier than reporting

bannsec:
Yep, it's under PR #11688. Was using this for tracking purposes.

PR #11688, opened 2018-09-28
Memleak

bannsec:
Guessing this is where the free should go... Ref #11687


binr/rabin2/rabin2.c
```diff
@@ -577,6 +577,7 @@ int main(int argc, char **argv) {
 		path = r_sys_getenv (R_LIB_ENV);
 		if (path && *path) {
 			r_lib_opendir (l, path);
+			free(path);
```
ret2libc: I think executing it always after the `if` is better. Otherwise, if path is not NULL, but it has length == 0, you won't free it.

PR #11698, opened 2018-09-30
Add golang symbols detection in aaa Fix #10959 ##anal

### LINKED ISSUE #10959: Integrate Go-Helpers natively
Integrate Go-Helpers from zlowram natively, https://github.com/zlowram/radare2-scripts/tree/master/go_helpers.

According to telegram user, the script need to be updated by changing the size calculation

Maijin:


Maijin:
The binary has only 2100 symbols detected.

I don't understand how comes `aa` is faster than `af @@ sym.*` but is supposed to do the same according to the help, this can also be observed with `aa*` vs `aa` time.

radare:
please rebase

libr/core/cmd_anal.c
```diff
@@ -6826,7 +6826,7 @@ static void rowlog(RCore *core, const char *str) {
 	}
 }
 
-static void rowlog_done(RCore *core) {
+R_API void rowlog_done(RCore *core) {
```
radare: dont make non r_* apis public with R_API, this should be static, or refactored into r_print

radare:
can you share the binary?

Maijin:
@radare - Sure it's https://github.com/radare/radare2-regressions/blob/master/bins/elf/analysis/go_stripped2

Maijin:
But the aa*, aa analysis difference can be seen using `/bin/ls` or any binary

Maijin:
For reference, I'm now using aF instead of af which does the trick for this case

libr/util/print.c
```diff
@@ -1968,3 +1968,30 @@ R_API void r_print_hex_from_bin (RPrint *p, char *bin_str) {
...
+
+R_API char* r_print_rowlog(RPrint *print, char *str) {
+	int use_color = print->flags & R_PRINT_FLAGS_COLOR;
+	bool verbose =  print->scr_prompt;
+	char* oldstr = str;
+	if (!verbose) {
+		return oldstr;
+	}
+	if (use_color) {
+		eprintf ("[ ] "Color_YELLOW"%s\r["Color_RESET, str);
```
xarkes: Why `eprintf` here?
Isn't eprintf only used to print debug info on `stderr`?

Maijin: This was like that before probably for a reason

Maijin:
I'm rebasing because tests got executed after merge of tests

radare:
the pr lgtm, the time diff between aa and aa* should be tracked in a separate issue

libr/util/print.c
```diff
@@ -1968,3 +1968,30 @@ R_API void r_print_hex_from_bin (RPrint *p, char *bin_str) {
 	p->cb_printf ("\n");
 	free (buf);
 }
+
+R_API char* r_print_rowlog(RPrint *print, char *str) {
```
radare: use const char *

libr/util/print.c
```diff
@@ -1968,3 +1968,30 @@ R_API void r_print_hex_from_bin (RPrint *p, char *bin_str) {
...
+	int use_color = print->flags & R_PRINT_FLAGS_COLOR;
+	bool verbose =  print->scr_prompt;
+	char* oldstr = str;
```
radare: use const char *

libr/util/print.c
```diff
@@ -1968,3 +1968,30 @@ R_API void r_print_hex_from_bin (RPrint *p, char *bin_str) {
...
+	}
+	if (use_color) {
+		eprintf ("[ ] "Color_YELLOW"%s\r["Color_RESET, str);
+	} else {
+		eprintf ("[ ] %s\r[", str);
+	}
+	return oldstr;
+}
+
+R_API void r_print_rowlog_done(RPrint *print, char *oldstr) {
```
radare: use const char *

radare:
please fix the comments

Maijin:
Fixed!

PR #11699, opened 2018-09-30
Refactor cpp symbols detection.

devnexen:
Merging them into one and returning similar value.
"Fix" for #11662

### DIFF
--- a/libr/bin/p/bin_elf.inc
+++ b/libr/bin/p/bin_elf.inc
@@ -1078,22 +1078,6 @@ static bool has_sanitizers(RBinFile *bf) {
 	return ret;
 }
 
-static bool is_cpp(RBinFile *binfile) {
-	bool ret = false;
-	RList* imports_list = imports (binfile);
-	RListIter *iter;
-	RBinImport *import;
-	r_list_foreach (imports_list, iter, import) {
-		if (strncmp (import->name, "_ZS", 3) == 0 ||
-                    strncmp (import->name, "_ZN", 3) == 0) {
-			ret = true;
-			break;
-		}
-	}
-	r_list_free (imports_list);
-	return ret;
-}
-
 static bool is_golang(RBinFile *binfile) {
 	bool ret = false;
 	RList* section_list = sections (binfile);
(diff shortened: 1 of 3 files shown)


radare:
Dont think this is a c++ bin but i didnt checked

libr/bin/p/bin_elf.inc
```diff
@@ -1118,7 +1102,7 @@ static RBinInfo* info(RBinFile *bf) {
 	if (!(ret = R_NEW0 (RBinInfo))) {
 		return NULL;
 	}
-	ret->lang = is_cpp (bf) ? "c++" : "c";
+	ret->lang = r_bin_is_cxx (bf) ? "cxx" : "c";
```
ret2libc: why did you change the string from `c++` to `cxx`? I'd keep it as it was if there is no particular reason.

devnexen: It s to be "in line" with the mangling feature change. But I can put it back if it triggers annoyance.

ret2libc: I see

libr/include/r_bin.h
```diff
@@ -820,6 +820,7 @@ R_API void r_bin_filter_sections (RList *list);
 R_API void r_bin_filter_classes (RList *list);
 R_API bool r_bin_strpurge(RBin *bin, const char *str, ut64 addr);
 R_API bool r_bin_string_filter(RBin *bin, const char *str, ut64 addr);
+R_API bool r_bin_is_cxx (RBinFile *binfile);
```
ret2libc: style: no space before `(`.

libr/bin/p/bin_elf.inc
ret2libc: I think this change actually changes the semantic of the check... before we were just checking imports, now you are checking all symbols, right? Is that what you wanted?

Also, shall we consider C binaries compiled with ASAN (and that for this reasons have some mangled names) C++ binaries? Or shall we filter those ASAN symbols out?

devnexen: No after second thoughts I prefer not considering C sanitized binaries be seen as.

radare:
Merging, but at some point we must improve the language detectors to avoid performance costs when loading bins with lot of symbols

PR #11707, opened 2018-10-01
Fixing few leaks.

devnexen:


binr/rahash2/rahash2.c
```diff
@@ -649,6 +649,7 @@ int main(int argc, char **argv) {
 		}
 	}
 	if (optind >= argc) {
+		free (iv);
```
ret2libc: since you are at it, could you `free(iv)` in all relevant paths?

shlr/java/class.c
```diff
@@ -3598,7 +3598,11 @@ R_API ut64 r_bin_java_signature_attr_calc_size(RBinJavaAttrInfo *attr) {
 R_API RBinJavaAttrInfo *r_bin_java_enclosing_methods_attr_new(ut8 *buffer, ut64 sz, ut64 buf_offset) {
 	ut64 offset = 6;
 	RBinJavaAttrInfo *attr = r_bin_java_default_attr_new (buffer, sz, buf_offset);
-	if (!attr || sz < 10) {
+	if (!attr) {
+		return NULL;
+	}
+	if (sz < 10) {
+		free (attr);
```
ret2libc: if you want you could keep just one condition as it was before and just add `free (attr)`. It will work for both cases, since free works well on NULL too.

PR #11709, opened 2018-10-01
Fix pd-x, tests pd -x and pd--x too

radare:


radare:
We need tests for pd-2 vs pd -2

PR #11710, opened 2018-10-01
WIP: Initial import of the alloc api (testing only in str.c)

radare:


radare:
strdup() is a problem too

libr/util/alloc.c
```diff
@@ -0,0 +1,51 @@
...
+R_API void r_alloc_init () {
+	r_alloc_hooks (malloc, calloc, realloc, free);
+}
+
+R_API void *r_malloc(size_t sz) {
+	return _r_malloc? _r_malloc (sz): malloc (sz);
+}
+
+R_API void *r_calloc(size_t sz) {
+	return _r_malloc? _r_malloc (sz): malloc (sz);
```
ret2libc: ? shouldn't you use `r_calloc`?

libr/util/alloc.c
```diff
@@ -0,0 +1,51 @@
...
+}
+
+R_API void *r_realloc(void *p, size_t sz) {
+	if (_r_realloc) {
```
ret2libc: very very minor stuff, but... why the `if` here and the ternary operator everywhere else?

libr/include/r_util/r_alloc.h
```diff
@@ -0,0 +1,39 @@
...
+R_API void r_alloc_hooks(RMalloc m, RCalloc c, RRealloc r, RFree f);
+
+#if R_MALLOC_GLOBAL
+R_API RMalloc *r_malloc;
+R_API RCalloc *r_calloc;
+R_API RRealloc *r_realloc;
+R_API RFree *r_free;
+#else
+R_API void *r_malloc(size_t sz);
+R_API void *r_calloc(size_t sz);
```
ret2libc: wrong signature?

libr/util/alloc.c
```diff
@@ -0,0 +1,51 @@
...
+#if R_MALLOC_GLOBAL
+R_API void r_alloc_hooks(RMalloc m, RCalloc c, RRealloc r, RFree f) {
+	r_malloc = m? m: malloc;
+	r_calloc = c? c: calloc;
+	r_realloc = r? r: realloc;
+	r_free = f? f: free;
+}
+
+#else
+static RMalloc *_r_malloc = malloc;
```
ret2libc: can't you just reuse the global pointers that you use when `R_MALLOC_GLOBAL`?

radare:
> On 2 Oct 2018, at 09:55, Riccardo Schirone <notifications@github.com> wrote:
> 
> @ret2libc commented on this pull request.
> 
> I don't understand what's the big difference between R_MALLOC_GLOBAL == 0 and == 1. The things seem pretty similar to me.
> 
MallocGlobal == 1:

Call function pointers. Which fails to build and is how capstone solves this

MallocGlobal == 0:

Calls a function that checks for those symbols as in a wrap

radare:
Yes. But i want to keep them hidden from the global view

ret2libc:
And is it really that important to have the distinction between MallocGlobal==1/0? I don't see a lot of differences and I'm not sure about the advantages of having both of them.

libr/include/r_util/r_alloc.h
```diff
@@ -0,0 +1,39 @@
...
+R_API void *r_realloc(void *p, size_t sz);
+R_API void r_free(void *p);
+#endif
+
+#else
+
+#define r_malloc(x) malloc((x))
+#define r_calloc(x,y) calloc((x),(y))
```
ret2libc: one thing i hate in the `R_NEW` macros is that i always forget what is nmemb and what is sz. Could you use more meaningful names for the arguments?

radare:
Its for testing and discussing. Obv not to merge

radare:
Ypu can also commit whatever you want in this pr. This is for discussion and playground

ret2libc:
> Its for testing and discussing. Obv not to merge

perfect. One thing that would help to understand the usefulness of this is to have the alternative implementation of malloc/calloc/etc.

Maybe one alternative implementation may be one that exit with an error as soon as malloc/calloc/etc. returns NULL. It may be used for debugging or similar...

radare:
To use different heap than the app using r2api itself

Choose different alloc impl for benchmarking

Allow to run in kernel, bootloader...

XVilka:
I think it is even possible to use dmalloc http://dmalloc.com/
I used it in one of my embedded projects, it turned up to be VERY useful to spot the bugs.

libr/util/alloc.c
radare: uh?

libr/include/r_util/r_alloc.h
radare: what? swapping nmemb and sz results in the same results,  in fact it's documented, that it doesnt matter at all if you swap them. so i dont see that important at all

ret2libc:
I'm perfectly fine with having this PR in. But before merging it requires some fixes in the actual implementation, because I think there are some unnecessary things and some bugs (wrong calloc signatures, etc.) as noticed in comments.

radare:
yep. i have to address those comments, but the main problem i see is that not everything will be allocated via r_malloc, so we can have problems when using alternative implementations, this is just the base not the whole transition

libr/util/alloc.c
radare: yes

libr/util/alloc.c
radare: maybe we should just assert because there's no sense in  just setting some o f thoe pointers and not others

radare:
pls review again

radare:
if we use r_malloc we can do checks like if (allocSize > MB(16)) { ...

radare:
@ret2libc what do you think about using static inline instead? actually we shuold be able to configure the malloc-wrapper-level. but im still unsure about the benefits:

pros:
* use different allocation engine
* detect who tries to do HUGE allocations
* ability to implement our memory profiler/logs
* ability to run in kernel/grub by changing the malloc callbacks

cons:
* we should change zillions of calls to malloc/calloc/free
* easy that any other plugin uses libc apis inatead of r_alloc ones and will crash
* require global vars which are evil

PR #11715, opened 2018-10-02
Rbin ELF plugin r_bin_elf_resize_section was not resizing the ELF segment

wdworks:
I beleive this fixes it
See this [issue](https://github.com/radare/radare2/issues/11708)

### DIFF
--- a/libr/bin/format/elf/elf_write.c
+++ b/libr/bin/format/elf/elf_write.c
@@ -140,7 +140,10 @@ ut64 Elf_(r_bin_elf_resize_section)(struct Elf_(r_bin_elf_obj_t) *bin, const cha
 			if (phdrp->p_paddr) {
 				phdrp->p_paddr += delta;
 			}
-		}
+		} else if (phdrp->p_offset + phdrp->p_filesz >= rsz_offset + rsz_osize) {
+			phdrp->p_filesz += delta;
+			phdrp->p_memsz += delta;
+		} 
 		off = ehdr->e_phoff + i * sizeof (Elf_(Phdr));
 		if (r_buf_write_at (bin->b, off, (ut8 *)phdrp, sizeof (Elf_ (Phdr))) == -1) {
 			perror ("write (phdr)");


libr/bin/format/elf/elf_write.c
```diff
@@ -141,6 +141,10 @@ ut64 Elf_(r_bin_elf_resize_section)(struct Elf_(r_bin_elf_obj_t) *bin, const cha
 				phdrp->p_paddr += delta;
 			}
 		}
+		else if( phdrp->p_offset + phdrp->p_filesz >= rsz_offset + rsz_osize ){
```
ret2libc: Please check https://github.com/radare/radare2/blob/master/DEVELOPERS.md for the coding style. We usually do:
```
if (..) {
<tab>...
} else if (...) {
<tab>...
}
```

ret2libc:
Just out of curiosity, since I don't usually use this command.. how are you testing whether this works? I've tried with:
```
$ cp /bin/ls /tmp/ls
$ rabin2 -O r/.text/80000 /tmp/ls
$ /tmp/ls
```

Do you expect the binary to still work after increasing the .text section? Am I doing anything wrong?

wdworks:
With the text section it of course won't work  because all the data/reloc offsets have to be fixed and that's usually too hard.
Common theme is to increase data section(because it is mapped after text and most of the PROGBITS segments) in order to place the code there and execute it by copying to executable page.
This is usefull when you want to obfuscate the compiled binary and you have to place your code somewhere.
```
$ cp /bin/ls /tmp/
$ rabin2 -S /tmp/ls | grep '\.data' 
20 0x00023040  2616 0x00223040  2616 -rw- .data.rel.ro
23 0x00024000   616 0x00224000   616 -rw- .data
$ rabin2 -O r/.data/680 /tmp/ls 
delta: 64
-> elf section ()
-> elf section (.interp)
-> elf section (.note.ABI-tag)
-> elf section (.gnu.hash)
-> elf section (.dynsym)
-> elf section (.dynstr)
-> elf section (.gnu.version)
-> elf section (.gnu.version_r)
-> elf section (.rela.dyn)
-> elf section (.init)
-> elf section (.plt)
-> elf section (.plt.got)
-> elf section (.text)
-> elf section (.fini)
-> elf section (.rodata)
-> elf section (.eh_frame_hdr)
-> elf section (.eh_frame)
-> elf section (.init_array)
-> elf section (.fini_array)
-> elf section (.jcr)
-> elf section (.data.rel.ro)
-> elf section (.dynamic)
-> elf section (.got)
-> elf section (.data)
-> elf section (.bss)
-> elf section (.shstrtab)
-> program header (0x00000040)
-> program header (0x00000238)
-> program header (0x00000000)
-> program header (0x00023028)
-> program header (0x00023a78)
-> program header (0x00000254)
-> program header (0x0001e730)
-> program header (0x00000000)
-> program header (0x00023028)
COPY FROM 0x00024268
COPY TO 0x000242a8
Shifted 64 byte(s)
$ rabin2 -S /tmp/ls | grep '\.data' 
20 0x00023040  2616 0x00223040  2616 -rw- .data.rel.ro
23 0x00024000   680 0x00224000   680 -rw- .data
$ /tmp/ls 
/bin  boot  dev	etc  home  include  lib  lib64	lost+found  media  mnt	opt  proc  r2pipe  root  run  sbin  sys  tmp  usr  var	virt

```

libr/bin/format/elf/elf_write.c
wdworks: Sorry about that.
Changed my despicable else if.

ret2libc:
Right! Thanks for the explanation.

wdworks:
> Right! Thanks for the explanation.

That interesting that apearently either almost no one used the feature or i am wrong here somewhere.
The only way it is possible that this is present in the ELF Rbin plugin which is arguably the most used one. And r2 is like 10 yrs and counting?

wdworks:
How to interpret this  codecov patch failure? Should i somehow merge it in a single commit to be rigth?

ret2libc:
Don’t worry about codecov report

wdworks:
I've just got a though all of a sudden that if we resize segment we should respect the segment alignment. Or the hell with it? Not 100% sure about the fatality of it in elf. For the most part it will work but are there edge case scenarios?

radare:
Please rebase

libr/bin/format/elf/elf_write.c
```diff
@@ -140,7 +140,10 @@ ut64 Elf_(r_bin_elf_resize_section)(struct Elf_(r_bin_elf_obj_t) *bin, const cha
 			if (phdrp->p_paddr) {
 				phdrp->p_paddr += delta;
 			}
-		}
+		} else if ( phdrp->p_offset + phdrp->p_filesz >= rsz_offset + rsz_osize ) {
```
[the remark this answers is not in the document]
wdworks: Squashed em and removed the trial spaces within ().
Hopefully that what was expected).

PR #11717, opened 2018-10-02
Add rust binary detection support.

devnexen:
Basic for non stripped, using symbols lookup
to avoid additional lookups.


radare:
Check this out this PR, thats the idea i have to improve the rbin parsing performance.

https://github.com/radare/radare2/pull/11712

PR #11718, opened 2018-10-02
detecting rust binaries for Mach-O

devnexen:


radare:
STsart commits in uppercase

PR #11723, opened 2018-10-03
WIP: ptrace-wrap

thestr4ng3r:


ret2libc:
Apart from the actual content of the PR which I didn't look at yet... I don't like the `shlr` approach. I think you should create a separate repository, maybe with your own unit tests, etc. Otherwise, code that is created in shlr just for r2 will never be reusable and it would be just easier to directly include it in r_debug.

thestr4ng3r:
The original repo is here: https://github.com/thestr4ng3r/ptrace-wrap
I am not a fan of copying dependencies into shlr too, but since it's so tiny I think it is probably the most practical approach here.

ret2libc:
Oh cool! I didn't see it. Is it mentioned anywhere? Shall we create a target for Makefile like `sdb-sync` or similar? Also, maybe it's worth automatically adding a comment at the beginning of every .c/.h file with something like `do not edit this file directly, modify upstream source code at XXXXXX`.

thestr4ng3r:
Yes, that is a very good idea. I'll add a target.

libr/io/p/io_debug.c
```diff
@@ -465,69 +465,82 @@ static char *get_and_escape_path (char *str) {
...
+	bool runprofile;
+	const char *cmd;
+} fork_child_data;
+
+static void fork_child_callback(void *user) {
+	fork_child_data *data = user;
+	if (data->runprofile) {
+		char **argv = r_str_argv (data->cmd, NULL);
+		if (!argv) {
+			exit(1);
```
radare: space

shlr/sdb/src/base64.c
```diff
@@ -10,7 +10,7 @@
 static const char cb64[]="ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 static const char cd64[]="|$$$}rstuvwxyz{$$$$$$$>?@ABCDEFGHIJKLMNOPQRSTUVW$$$$$$XYZ[\\]^_`abcdefghijklmnopq";
 
-static void b64_encode(const ut8 in[3], char out[4], int len) {
+static void sdb_b64_encode(const ut8 in[3], char out[4], int len) {
```
radare: i dont think you should update sdb in here, isnt sdb in master already? do you need to rebase maybe?

libr/include/r_io.h
```diff
@@ -474,6 +501,11 @@ R_API bool r_io_addr_is_mapped(RIO *io, ut64 vaddr);
 R_API bool r_io_read_i (RIO* io, ut64 addr, ut64 *val, int size, bool endian);
 R_API bool r_io_write_i (RIO* io, ut64 addr, ut64 *val, int size, bool endian);
 
+#if HAVE_PTRACE
+R_API long r_io_ptrace(RIO *io, r_ptrace_request_t request, pid_t pid, void *addr, r_ptrace_data_t data);
```
radare: in linux/ppc the last two args are swapped, i guess this change doesnt affects the behaviour. but just as a random comment

libr/io/p/io_debug.c
```diff
@@ -465,69 +465,82 @@ static char *get_and_escape_path (char *str) {
...
+			argv[0] = r_str_replace (argv[0], "\\ ", " ", true);
+		}
+		if (!argv) {
+			free (path_escaped);
+			free (_cmd);
+			return;
+		}
+		if (argv && *argv) {
+			int i;
+			for (i = 3; i < 1024; i++) {
```
radare: uhm

libr/include/r_io.h
thestr4ng3r: Hmm ok, is that documented anywhere? It doesn't really matter here if the types of the args are not swapped.

shlr/sdb/src/base64.c
thestr4ng3r: yes, needs a rebase.

libr/io/p/io_debug.c
thestr4ng3r: 🙈

libr/io/p/io_debug.c
thestr4ng3r: Not my code, I just moved it as it was. But I don't really understand the sense of closing [3;1024[ as well.

thestr4ng3r:
> why is r_debug_ptrace and r_io_ptrace ? sounds to me like dupped apis

r_debug_ptrace is just a wrapper for the io one from io bind, because ptrace is so important in debug, so you don't have to write `dbg->iob.ptrace (dbg->iob.io, ...` all the time.

libr/include/r_io.h
radare: its documented in the manpager of the ppclinux ptrace syscall. but i dont have any linux-ppc now to test so meh :P

PR #11726, opened 2018-10-03
Fix debugging in FreeBSD 11.2

saucec0de:
In FreeBSD 11.2 (stable) radare2 has issues with debugging - it does not disable/enable breakpoints correctly leading to wrong analysis and also to unusable debugging. This patch fixes this issue.

### DIFF
--- a/libr/debug/p/debug_native.c
+++ b/libr/debug/p/debug_native.c
@@ -220,7 +220,9 @@ static int r_debug_native_continue_syscall (RDebug *dbg, int pid, int num) {
 	return ptrace (PTRACE_SYSCALL, pid, 0, 0);
 #elif __BSD__
 	ut64 pc = r_debug_reg_get (dbg, "PC");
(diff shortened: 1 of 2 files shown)


libr/debug/p/debug_native.c
```diff
@@ -220,7 +220,9 @@ static int r_debug_native_continue_syscall (RDebug *dbg, int pid, int num) {
 	return ptrace (PTRACE_SYSCALL, pid, 0, 0);
 #elif __BSD__
 	ut64 pc = r_debug_reg_get (dbg, "PC");
-	return ptrace (PTRACE_SYSCALL, pid, (void*)(size_t)pc, 0);
+	errno = 0;
+	int retVal = ptrace (PTRACE_SYSCALL, pid, (void*)(size_t)pc, 0);
+	return 0==retVal;
```
radare: use `return retval == 0`. (first var and then value), also use shorter and non-camelcased variable names

PR #11731, opened 2018-10-04
Added main detection for linux64 endbr64

xarkes:


radare:
can you the bin as a test? thanks

PR #11747, opened 2018-10-05
Fix CF on X86_INS_SHL

xtrm0:


libr/anal/p/anal_x86_cs.c
```diff
@@ -777,6 +777,31 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 		}
 		break;
 	case X86_INS_SHL:
+	{
```
radare: ths seems like spaguetti with the other PR, can you generalize the patch to reuse this switch case?

xtrm0:
Ok, then let me make a new PR with the change, to make mergine easier (and also because october fest)

xtrm0:
I think I will just move the switch case, because SHL and SAL are exactly the same instruction in x86 afaik

xtrm0:
Also shouldn't there be some unit tests for the ESIL + ESIL emulator? Like for instance setting values on registers on qemu, then running a few steps, and then comparing some values with the values of the same operations on the ESIL?

radare:
can we close this ?

PR #11764, opened 2018-10-07
Kill anal.bb.align variable

radare:


radare:
Why remove if its red? It was just a test to see how many xx was happening if i remove this key

radare:
maybe we should kill this variable, but keep the behaviour that makes it green by default

PR #11769, opened 2018-10-08
Fix leak in cmd_open

F3real:


libr/core/cmd_open.c
```diff
@@ -1229,6 +1229,7 @@ static int cmd_open(void *data, const char *input) {
 					eprintf ("%d\n", fd);
 				}
 				r_core_bin_load (core, argv0, baddr);
+				free (file);
```
radare: file is not a simple heap pointer. but RCoreFile. this fix is hiding the real problem

ret2libc:
Also... do we have a ASAN stack trace? How did you reproduce the leak?

XVilka:
Looks red.

PR #11727, opened 2018-10-04
Make up lines even more beautiful

jdz:
Use dashed line from the same Unicode block as other line drawing
characters.  This ensures that they all are present in the same font,
are the same width and align perfectly.

PR #11770, opened 2018-10-08
Make up lines even more beautiful

jdz:
Use dashed line from the same Unicode block as other line drawing
characters.  This ensures that they all are present in the same font,
are the same width and align perfectly.

Re-opening #11727 according to guidelines.

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -37,7 +37,7 @@ static const char* r_vline_u[] = {
 	"│", // LINE_VERT
 	"├", // LINE_CROSS
 	"─", // LINE_HORIZ
-	"⁝", // LINE_UP
+	"╎", // LINE_UP
 	// "↑", // LINE_UP
 	//"┌", // LUP_CORNER
 	"┘", // LUP_CORNER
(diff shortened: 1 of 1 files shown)


jdz:
3/4 tests failed because r2r PR was (initially) not rebased on top of master.

radare:
without curvy
screenshot 2018-10-09 at 00 22 26

radare:
before was not looking great either..

screenshot 2018-10-09 at 00 23 47

jdz:
What font is this?

It looks like it does not have the "box drawing" Unicode block (the lines are way lighter than the font itself).  It seems to me that the missing corner and vertical line characters are pulled from one font, and this font does not have dashed vertical line, so it is pulled from yet another font.

In #11727 I included screenshots with 4 different monospace fonts that do have characters for the whole "box drawing" Unicode block.  You might have noticed that I've removed some of the screenshots from the original comment.  That's because I later found out that in the other screenshots the box drawing characters were taken from DejaVu Sans Mono.

To summarise: not all fonts have the box drawing characters in the font, but when they do, there's a great chance the whole Unicode block is included. If your favourite font does not have this unicode block then the font substitution must be configured.  There's ample information on how to do it on Linux with fontconfig; not sure about MacOS.

I still stand by this change and think it's the right thing to do.

jdz:
I use Input: http://input.fontbureau.com/.

XVilka:
Maybe open a bug in fonts missing those characters?

radare:
Font makers are strange copyright creatures. Not sure if we can make it change

PR #11771, opened 2018-10-08
Avoid centering graph nodes too big for the current terminal size

### LINKED ISSUE #11740: Graph improvements: not centered in the middle of the node
screenshot 2018-10-05 at 10 39 45

cyanpencil:
closes #11740


radare:
Doesnt seems to work fine in all cases. Try with a random dex with very large method calls

PR #11772, opened 2018-10-08
NetBSD/OpenBSD build fix

devnexen:


devnexen:
Might be ... if this then it might be a recent change I built in openbsd about 2 days ago. Will try later if necessary.

PR #11773, opened 2018-10-08
Print marks on flags  if asm.offset=false

### LINKED ISSUE #11062: Console UI issue with asm.offset and asm.cmt.right
`e asm.offset = false` displays weird space and `e asm.tabs = <number_of_tab>` does not work with `e asm.cmt.right = false`
### Work environment

| OS/arch/bits (mandatory)                             | Parrot Security OS 4.1 amd64
| r2 -v full output, **not truncated** (mandatory)     | radare2 2.9.0-git 19081 @ linux-x86-64 git.2.8.0-70-g8a2397e4b
commit: 8a2397e4b1b6bdbfb23c84f02a5abe6a5fa80149 build: 2018-08-13__14:46:52

1. `e asm.offset = false` display
-  radare2 display with `e asm.offset = true`
screenshot at 2018-08-13 15-08-03

- radare2 display with `e asm.offset = false`
screenshot at 2018-08-13 15-11-09
In my opinion, disasm code should display from left side without space (green highlight)
- VV mode does not have this issue
screenshot at 2018-08-13 15-13-47

2. `e asm.tabs` does not work with `e asm.cmt.right = false`
- Here is how radare2 display with `e asm.cmt.right = true` and `e asm.tabs = 16`
screenshot at 2018-08-13 15-15-24

- Set `asm.cmt.right = false`, display space will disappear. (It looks super weird without  offset and cmt.right)
screenshot at 2018-08-13 15-16-14
(`VV` mode set `asm.cmt.right = false` by default i think so nothing changes in this mode)

cyanpencil:
closes #11062 

Now if offset is disabled and marks are enabled, mark space will be printed on flags too.
This avoid ugly alignment like this:
```
/ (fcn) fcn.00401000 31
|   fcn.00401000 ();
|           ; var int local_8h @ ebp-0x8
|           ; var int local_4h @ ebp-0x4
|                push ebp
|                mov  ebp, esp
|                sub  esp, 8
|                nop
|                mov  eax, 0x402000
|                mov  dword [local_4h], eax
```
and substitutes with this
```
/ (fcn) fcn.00401000 31
|   fcn.00401000 ();
|               ; var int local_8h @ ebp-0x8
|               ; var int local_4h @ ebp-0x4
|                push ebp
|                mov  ebp, esp
|                sub  esp, 8
|                nop
|                mov  eax, 0x402000
|                mov  dword [local_4h], eax
```
This will duplicate marks on flags, though. I didn't do it, but if it's a problem I can easily change it to whitespace only

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -439,17 +439,6 @@ static void ds_comment_lineup(RDisasmState *ds) {
 	_ALIGN;
 }
 
-static void ds_begin_comment(RDisasmState *ds) {
-	if (ds->show_comment_right) {
-		_ALIGN;
-	} else {
-		if (ds->use_json) {
-			ds_begin_json_line (ds);
-		}
-		ds_pre_xrefs (ds, false);
-	}
-}
-
 static void ds_comment_(RDisasmState *ds, bool align, bool nl, const char *format, va_list ap) {
 	if (ds->show_comments) {
 		if (ds->show_comment_right && align) {
@@ -1408,6 +1397,17 @@ static void ds_pre_xrefs(RDisasmState *ds, bool no_fcnlines) {
 	ds->line = tmp;
 }
 
+static void ds_begin_comment(RDisasmState *ds) {
+	if (ds->show_comment_right) {
+		_ALIGN;
+	} else {
+		if (ds->use_json) {
+			ds_begin_json_line (ds);
+		}
+		ds_pre_xrefs (ds, false);
+	}
+}
+
 static int var_comparator(const RAnalVar *a, const RAnalVar *b){
 	if (a && b) {
 		return a->delta > b->delta;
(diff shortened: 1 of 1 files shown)


ret2libc:
From testing, this seems to improve the situation (so I'm ok merging this), though there are many quirks with `asm.offset=false`.
Two things in particular:
1) the lines are not aligned correctly
2) the comments like `; CODE XREF from main` are still not aligned correctly.

radare:
ping

cyanpencil:
> the lines are not aligned correctly 

I think you are referring to this behaviour:
```
        ,=<      7418           je 0x50b0                              ;[2]                                            
        |        488b0511af21.  mov rax, qword [reloc._ITM_registerTMCloneTable]       ; [0x21ffb0:8]=0                
        |        4885c0         test rax, rax                                                                          
       ,==<      740c           je 0x50b0                              ;[2]                                            
       ||        5d             pop rbp                                                                                
       ||        ffe0           jmp rax                                                                                
       ||        660f1f840000.  nop word [rax + rax
```
but this is because of the space for the marks... I can't really think of a clean solution here :(

> the comments like ; CODE XREF from main are still not aligned correctly.

You are right, missed them, now they should be fixed

radare:
solve the cornflakes^Wconflicts

PR #11774, opened 2018-10-08
Supports clock_nanosleep on FreeBSD/NetBSD.

devnexen:


libr/include/r_util/r_sys.h
```diff
@@ -9,6 +9,12 @@
 #define R_SYS_DEVNULL "/dev/null"
 #endif
 
+#if __linux__ || __FreeBSD__ || __NetBSD__
```
radare: maybe add OpenBSD too?

devnexen: We do not have this :-)

Issue #11692, opened 2018-09-29
Memory Leak in rabin -H


bannsec:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             |  Ubuntu x86 64
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86/64
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.0.0-git 19569 @ linux-x86-64 git.2.9.0-240-g2582b36a7 commit: 2582b36a77b7b754000b071fe1f15fc4797d28e3 build: 2018-09-29__15:43:50

### Expected behavior
No memory leak

### Actual behavior
Memory leak

### Steps to reproduce the behavior 
- Compile r2 with ASAN/LeakSanitizer
- Run `rabin2 -H test_file` where `test_file` is simply a blank file

```
$ ASAN_OPTIONS=detect_odr_violation=0 rabin2 -H blerg

=================================================================
==42476==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1 byte(s) in 1 object(s) allocated from:
    #0 0x56098bb3d308 in calloc (/home/angr/opt/radare2/binr/rabin2/rabin2+0xe4308)
    #1 0x7fb6327002e9 in r_bin_load_io2 /home/angr/opt/radare2/libr/bin/bin.c:441:15
    #2 0x7fb6326ffef1 in r_bin_load_io /home/angr/opt/radare2/libr/bin/bin.c:42:12
    #3 0x7fb63270173b in r_bin_load /home/angr/opt/radare2/libr/bin/bin.c:261:9
    #4 0x7fb632722c89 in r_bin_open /home/angr/opt/radare2/libr/bin/open.c:32:6
    #5 0x56098bb77ffb in main /home/angr/opt/radare2/binr/rabin2/rabin2.c:1014:7
    #6 0x7fb6309c1b96 in __libc_start_main /build/glibc-OTsEL5/glibc-2.27/csu/../csu/libc-start.c:310

SUMMARY: AddressSanitizer: 1 byte(s) leaked in 1 allocation(s).
```

Note: This leak happens with other examples as well, this is just the easiest to reproduce, for example:

```
=================================================================
==42618==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32768 byte(s) in 1 object(s) allocated from:
[...]

PR #11775, opened 2018-10-08
Catching leaked mem

bannsec:
Ref #11692. I did put a conditional in this free since at line 459 it gets freed and not necessarily recreated:

```
 459                         R_FREE (buf_bytes);
 460                         if (tfd < 0) {
 461                             tfd = iob->fd_open (io, fname, R_PERM_R, 0);
 462                         }
 463                         sz = iob->fd_size (io, tfd);
 464                         if (sz != UT64_MAX) {
 465                             buf_bytes = calloc (1, sz + 1);
 466                             if (buf_bytes) {
 467                                 (void) iob->fd_read_at (io, tfd, 0, buf_bytes, sz);
 468                             }
 469                         }
```

Does radare have a conditional free? Similar to R_FREE but only frees if the value is non-null?

### DIFF
--- a/libr/bin/bin.c
+++ b/libr/bin/bin.c
@@ -486,7 +486,7 @@ R_API bool r_bin_load_io2(RBin *bin, int fd, ut64 baseaddr, ut64 loadaddr, int x
 			bin, fname, buf_bytes, sz, file_sz, bin->rawstr,
 			baseaddr, loadaddr, fd, name, NULL, offset, true);
 	} else {
-		R_FREE (buf_bytes);
+		free (buf_bytes);
 	}
 	return binfile? r_bin_file_set_cur_binfile (bin, binfile): false;
 }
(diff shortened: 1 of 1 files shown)


radare:
Conditional free? Thats implicit. All free implementations handle null properly theres no need to add an extra if

bannsec:
Oh right! lol Yeah, I'm suuper rusty on C. Removed the conditional.

bannsec:
I see what you're saying. So you're right, the `r_bin_file_new_from_bytes` call does definitely "steal" the bytes. The `r_bin_file_xtr_load_bytes` actually calls `r_bin_file_create_append` with `steal_ptr` set to `False`, so that should be fine.

I think if we just change the R_FREE into an else->R_FREE, that should correctly catch the leak while keeping the memory there for the call that needed it.

bannsec:
Pushed the change. Basically, that last if statement appears to be the only place that possibly steals the pointer. Thus, if we don't follow that branch, we should be able to safely free the buffer.

PR #11786, opened 2018-10-10
Fix #10324 - Implement RLog API for leveled logging ##core (#11786)

### LINKED ISSUE #10324: A debug print / log level system for output
Right now it seems that eprintf is the only method we have for outputting to the console that isn't intended for 'proper' program use. So when developing something or debugging something I'm seeing eprintf being used for debug output and then commented out or removed before commit.

I think a syslog style logging system would be useful for debugging/development. Could add a extra argument flag to the rbin programs to enable file output in a syslog format or to the console itself. Radare itself could use a scr.loglevel variable to select output to the console (not sure how it would work in visual mode).

Then from a programming standpoint we could have some sort of LOG singleton with LOG.warn/error/info ect. which would output (if matching loglevel) to the console or a file.

Let me know if theres any qualms or better ideas. I'm really just wanting a good way to enable debug level output temporarily for development or debugging issues.

mcd1992:
Resolves #10324 

This should also fit all the criteria in https://hackmd.io/s/r1I08jpHm for Cutter support as well.

I haven't changed any of the eprintfs over to the new system yet. I'm planning to do it library by library to make sure nothing catches fire.

Hopefully I didn't miss anything. Also :jack_o_lantern: :shirt:


mcd1992:
I'm half expecting to see some jenkins/CI errors, I didn't test with Windows / OSX...

libr/include/r_util/r_assert.h
```diff
@@ -78,28 +76,28 @@ R_API void r_assert_log(RLogLevel level, const char *fmt, ...);
 #define r_return_if_fail(expr) \
 	do { \
 		if (!(expr)) { \
-			H_LOG_ (R_LOG_WARNING, "%s: assertion '%s' failed (line %d)\n", R_FUNCTION, #expr, __LINE__); \
+			R_LOG_WARN ("%s: assertion '%s' failed (line %d)\n", R_FUNCTION, #expr, __LINE__); \
```
ret2libc: this should remain `H_LOG_`

ret2libc: and all the others too

libr/util/logging.c
```diff
@@ -0,0 +1,157 @@
+/* radare - LGPL - Copyright 2007-2013 - pancake */
+
+#define LOGGING_CONFIGSTR_SIZE 1024
+#define LOGGING_OUTPUTBUF_SIZE 8192
+
+#include <r_core.h>
+#include <stdarg.h>
+
+static RLoggingFuncdef cb_logfunc_print = NULL; // Function to call when outputting log string
```
ret2libc: globals are not really thread-safe.

libr/util/logging.c
```diff
@@ -0,0 +1,157 @@
...
+	}
+	return true;
+}
+
+// cconfig.c configuration callback functions below
+int cb_logging_config_level(void *coreptr, void *nodeptr) {
+	RConfigNode *node = (RConfigNode*)nodeptr;
+	char *value = node->value;
+	int value_len = strnlen (value, LOGGING_CONFIGSTR_SIZE);
+	if (!str_is_digits (value, value_len))
```
ret2libc: just use `strtol` instead of str_is_digits + atoi

libr/util/logging.c
```diff
@@ -0,0 +1,157 @@
...
+ * \param funcname Contains the function name of the calling function
+ * \param filename Contains the filename that funcname is defined in
+ * \param lineno The line number that this log call is being made from in filename
+ * \param lvl Logging level for output
+ * \param tag String that will be prepended to the log line (describes lvl)
+ * \param fmtstr A printf like string
+ * This function is used by the R_LOG_* preprocessor macros for logging/debugging
+*/
+void _r_logging_internal(const char *funcname, const char *filename,\
+ unsigned int lineno, unsigned int level, const char *tag, const char *fmtstr, ...) {
```
ret2libc: since `level` and `tag` are always the same, I see no reasons to have both arguments. Just compute the `tag` from the level, as we were doing in the previous log implementation.

mcd1992: I wasn't sure if there would ever be a use-case where someone wanted to set a custom tag. That way they can throw something at the error level but set a custom tag like `GREPFORME` or whatever they want to kinda make it stand out. I guess they could always just have it in the string itself so it would be `ERROR: GREPFORME: text`

libr/util/logging.c
mcd1992: The cconfig callbacks need to return false if the passed value is invalid. If I use strtol I can't tell if the user passed 'asdf' or 0/LONG_MAX/LONG_MIN. Although re-reading the manpage I can use errno to figure it out with strtol. I'll take a peek.

libr/include/r_util/r_assert.h
mcd1992: Ah didn't realize I changed the H_LOG_ define as well as removed all usage of it! Thanks.

libr/util/logging.c
mcd1992: My C knowledge is pretty limited when it comes to async/threaded. Is there an existing example in R2 I can look at to see how its done? Quick google points me to mutexing or thread-local storage. I'm not seeing any TLS stuff in libr at a glance though...

mcd1992:
Right now its just the basic features: https://gist.github.com/mcd1992/309c37879b481aff1b692027dc5148e9

I used to have a filter config var but all it was going to do was strncmp with the whole output string which was kinda lazy. The current setup will allow for a new config var like cfg.logging.filefilter which can allow you to filter down to a single file. We would have to do some weird tricks to get per-library filtering since the default preprocessor macros only give us the function_name filename:lineno like in the test output `WARN: cmd_info in cmd_info.c:325: hellotest`

libr/util/logging.c
ret2libc: yes you can AFAIK. you can also use `endptr` argument to know if there are other chars after the parsed number (e.g. `1234abcd` endptr should end up on the `a` and you'd know the input wasn't completely numeric)

libr/util/logging.c
mcd1992: Moved the buffer into _r_logging_internal. I'll ask around if its worth using thread local `__thread`/`__declspec(thread)` for the other things. We'd need to made defines for it and make sure its available on all supported platforms.

mcd1992:
Alrighty that should fix what has been pointed out.

The clang-format-diff script pointed out a lot of false positives. I fixed the switch indentation one but everything else seemed incorrect. The only ones I think might be relevant are me lining up the equal signs on enum declaration and similar. I see it done elsewhere in R2 but I can remove the alignment whitespace if needed. The other being long lines and multiline function declarations. I don't see any other examples of it in R2 so I stuck with my single space indent.

Thanks for reviewing!

ret2libc:
> The clang-format-diff script pointed out a lot of false positives

I think there are only few false positives. Some things that needs to be fixed and that the script reports:
```
-       const char *tag = R_BETWEEN (0, level, sizeof(level_tags)-1) ? level_tags[level] : "";
+       const char *tag = R_BETWEEN (0, level, sizeof (level_tags) - 1) ? level_tags[level] : "";      
```
```
-void _r_logging_internal(const char *funcname, const char *filename,
- unsigned int lineno, unsigned int level, const char *fmtstr, ...) {
+void _r_logging_internal (const char *funcname, const char *filename,
+       unsigned int lineno, unsigned int level, const char *fmtstr, ...) {
```
```
 #if (defined(_WIN32) || defined(_WIN64)) && !defined(__GNUC__)                                                        
-       #define MACRO_LOG_FUNC __FUNCTION__                                                                            
-       #define MACRO_WEAK_SYM                                                  
-       // TODO: Windows weak symbols?            
+#define MACRO_LOG_FUNC __FUNCTION__                                                                                   
+#define MACRO_WEAK_SYM                                                                                                
+// TODO: Windows weak symbols?                                                  
 #else                                             
-       #define MACRO_LOG_FUNC __func__                                          
-       #define MACRO_WEAK_SYM __attribute__((weak))
+#define MACRO_LOG_FUNC __func__                                                     
+#define MACRO_WEAK_SYM __attribute__ ((weak))                                                                          
 #endif                  
```
```
-       p = r_sys_getenv("R2_LOGCOLORS");           
-       SETCB ("cfg.logging.colors", p? p: "false", cb_logging_config_colors, "Should the logging output use colors (TODO)");
-       free(p);                                                                                                        
+       p = r_sys_getenv ("R2_LOGCOLORS");                                                                             
+       SETCB ("cfg.logging.colors", p ? p : "false", cb_logging_config_colors, "Should the logging output use colors (TODO)");
+       free (p);                 
```
```
-       RConfigNode *node = (RConfigNode*)nodeptr;                                                                
+       RConfigNode *node = (RConfigNode *)nodeptr;                 
```

This is not super important at this stage anyway (I expect @radare will require some changes when he'll look at it). But at least you know what's the expected style for future changes.

libr/util/logging.c
```diff
@@ -0,0 +1,123 @@
...
+	cfg_loglvl = level;
+}
+
+void r_logging_set_traplevel(RLoggingLevel level) {
+	cfg_logtraplvl = level;
+}
+
+void r_logging_set_file(const char *filename) {
+	int value_len = strnlen (filename, LOGGING_CONFIGSTR_SIZE) + 1;
+	strncpy (&cfg_logfile, filename, value_len);
```
ret2libc: no `&`, just cfg_logfile.

libr/util/logging.c
```diff
@@ -0,0 +1,123 @@
...
+	if (cfg_logfile[0] != 0x00) {
+		FILE *file = r_sandbox_fopen (cfg_logfile, "a+"); // TODO: Optimize (static? Needs to remake on cfg change though)
+		if (!file) {
+			file = r_sandbox_fopen (cfg_logfile, "w+");
+		}
+		if (file) {
+			fprintf (file, "%s", output_buf);
+			fclose (file);
+		} else {
+			eprintf ("_r_log_internal failed to write to file: r_sandbox_fopen returned %i for file %s\n", (int)file, cfg_logfile);
```
ret2libc: you probably want to use `fileno`, right?

libr/util/logging.c
```diff
@@ -0,0 +1,123 @@
...
+
+void r_logging_set_colors(bool show_info) {
+	cfg_logcolors = show_info;
+}
+
+/**
+ * \brief Set the logging callback function for printing
+ * \param cbfunc RLoggingFuncdef style function to be called
+*/
+void r_logging_set_callback(RLoggingFuncdef cbfunc) {
```
ret2libc: can you add `R_API` at the beginning of the definition? I'm not sure *why* we use it everywhere, but we are used to do that so just at least for consistency...

ret2libc:
Thanks a lot! You'll probably receive another review by @radare after 3.0 release ;)

libr/include/r_util/r_logging.h
```diff
@@ -0,0 +1,71 @@
...
+#define MACRO_LOG_FUNC __FUNCTION__
+#define MACRO_WEAK_SYM
+// TODO: Windows weak symbols?
+#else
+#define MACRO_LOG_FUNC __func__
+#define MACRO_WEAK_SYM __attribute__ ((weak))
+#endif
+
+typedef void (*RLoggingFuncdef)(const char *output, const char *funcname,
+	const char *filename, unsigned int lineno, unsigned int level, const char *fmtstr, ...);
```
thestr4ng3r: Please add a `void *user` argument to the callback, we will need that at least in Cutter. Then also add a variable to contain the value that should be passed to it next to the places where a callback is stored.

libr/util/logging.c
```diff
@@ -0,0 +1,123 @@
...
+#include <r_core.h>
+#include <stdarg.h>
+
+// TODO: Use thread-local storage to make these variables thread-safe
+static RLoggingFuncdef cb_logfunc_print = NULL; // Function to call when outputting log string
+static int cfg_loglvl = R_LOGLVL_ERROR; // Logging level output
+static int cfg_logtraplvl = R_LOGLVL_FATALTRAP; // Logging trap level
+static bool cfg_logsrcinfo = false; // Print out debug source info with the output
+static bool cfg_logcolors = false; // Output colored log text based on level
+static char cfg_logfile[LOGGING_CONFIGSTR_SIZE] = ""; // Output text to filename
```
thestr4ng3r: I personally would move these options (except the callback of course) to RCore and do the logging to a file for example in a callback there. Basically have a default logging callback inside core, which is set for each new task by default and does some of the logic that is now implemented in `_r_logging_internal()`. What do you think?

libr/include/r_util/r_assert.h
```diff
@@ -18,14 +16,14 @@ R_API void r_assert_log(RLogLevel level, const char *fmt, ...);
 
 #define r_warn_if_reached() \
 	do { \
-		r_assert_log (R_LOG_WARNING, "(%s:%d):%s%s code should not be reached\n", \
+		R_LOG_WARN ("(%s:%d):%s%s code should not be reached\n", \
```
radare: i would prefer to have no macros for this.

`r_log (R_LOG_WARN, "....");`

libr/include/r_util/r_logging.h
```diff
@@ -0,0 +1,71 @@
...
+	R_LOGLVL_VERBOSE   = 2,
+	R_LOGLVL_INFO      = 3,
+	R_LOGLVL_WARN      = 4,
+	R_LOGLVL_ERROR     = 5,
+	R_LOGLVL_FATAL     = 6,
+	R_LOGLVL_FATALTRAP = 7, // This will call r_sys_breakpoint() and trap the process for debugging!
+	R_LOGLVL_NONE      = 0xFF
+} RLoggingLevel;
+
+#define R_LOG(lvl, fmtstr, ...) _r_logging_internal (MACRO_LOG_FUNC, __FILE__,\
```
radare: i would prefer to avoid the use of macros

radare: nvm

libr/core/cconfig.c
```diff
@@ -2290,6 +2290,63 @@ static int cb_dbgsnap(void *user, void *data) {
 	return true;
 }
 
+static int cb_logging_config_level(void *coreptr, void *nodeptr) {
+	RConfigNode *node = (RConfigNode *)nodeptr;
+	char *value = node->value;
+	char *bad_data = NULL;
+	long int ival = strtol (value, &bad_data, 10);
```
radare: use r_num_math

libr/cons/cons.c
```diff
@@ -367,6 +368,18 @@ R_API bool r_cons_enable_mouse(const bool enable) {
 #endif
 }
 
+// Stub function that cb_logfunc_print gets assigned to in logging.c by default
+// This allows Cutter to set per-task logging redirection
+R_API void r_cons_logging_stub(const char *output, const char *funcname, const char *filename,
+ unsigned int lineno, unsigned int level, const char *fmtstr, ...) {
+	if (!I.null && I.context->logging_callback) {
+		va_list args;
+		I.context->logging_callback (output, funcname, filename, lineno, level, fmtstr, args);
```
radare: can we have this logging callback as an RList? so we can hook many listeners? this can be used for having events that update the UI and keep different instances of r2 in sync

libr/core/cconfig.c
```diff
@@ -2290,6 +2290,63 @@ static int cb_dbgsnap(void *user, void *data) {
...
+		return false;
+	r_logging_set_level (ival);
+	return true;
+}
+
+static int cb_logging_config_traplevel(void *coreptr, void *nodeptr) {
+	RConfigNode *node = (RConfigNode *)nodeptr;
+	char *value = node->value;
+	char *bad_data = NULL;
+	long int ival = strtol (value, &bad_data, 10);
```
radare: use r_num_math()

radare:
Also, please rebase and squash

libr/include/r_util/r_assert.h
ret2libc: Personally, I'm not against these kind of macros (or inline functions, if you prefer)... they are very useful and used in other projects as well.

libr/cons/cons.c
ret2libc: > can we have this logging callback as an RList? so we can hook many listeners?

this seems useful to save both to screen and to file or other places

> this can be used for having events that update the UI and keep different instances of r2 in sync

what do you mean with this?

radare:
See https://github.com/radareorg/cutter/issues/834#issuecomment-430160278

ret2libc:
I wouldn't use logging for that. For that we need a publish/subscriber pattern or something like that, logging is not the right tool IMHO.

radare:
why not just reuse the same logic and apis for this? otherwise we will end up with duplicated apis for different purposes. can you make a proposal?

ret2libc:
> why not just reuse the same logic and apis for this? otherwise we will end up with duplicated apis for different purposes. can you make a proposal?

I could make a proposal, but the problem is that log is very different from publish/subscribe. Think about creating a new function. You could have multiple calls to the log api to log different things done during function creation, but you would want to "publish" a `function-created` event only at the end of the function creation process. Also, you may want to attach an object to the event, so that subscribers can get details about the event. I do not think this can be cleanly done with the same API and it would create a mess as the two things evolve differently.

I'd imagine something like:
```C
struct function_created_event_t *evt = RNEW0(..);
evt->fcn_name = "xxxx";
evt->fcn_addr = 0xdeadbeef;
r_event_publish("function-created", evt);
```
```C
void fcn_created_cb(const char *evt_name, void *evt_data) {
    struct function_created_event_t *data = (struct function_created_event_t *)evt_data;
    ....
}

r_event_subscribe("function-created", fcn_created_cb);
```

If you want to talk more, we may continue on the right github issue, but I think it's much cleaner and flexible to keep the API separated and try to merge them *only if/when* we realize they are equal (I personally doubt it will ever happen because they are two very different concepts)

radare:
I'm thinking about events in just a string, if we start doing several binary data structures this will be a big mess imho. with binary incompetilibilites and crashes between different versions of r2, and this cant be catched at compile time. so i would just go for “tag” + “message” and tag can be a log level that is always delivered to all the listeners. this way we can even log them to file. without having to create new apis to duplicate all this.

About the subscription we will need to unsubscribe, and we want this api to be thread safe, so sharing pointers in events shuldnt be an option.

This event/log/chat thing (see the T command) may be also something to look at.

libr/include/r_util/r_logging.h
mcd1992: Whats the use case for this? Would it normally contain the RCore instance or?

libr/util/logging.c
mcd1992: Hrm, I thought we wanted all this in r_util. Could you update the hackmd with the new concept, I'm not fully following.

mcd1992:
I'm might not fully following the big picture here but if y'all are wanting to have some sort of C level API/IPC for things like Cutter I'd vote to save some sanity and sacrifice speed for something like a MQTT server builtin since you want publish/subscribe. Maybe even reuse the tcp/http stuff that r2pipe does with websockets/protobuf or something. 

I do agree that this doesn't seem relevant to the current logging stuff though, this should be some separate API/library spec'd out in a different issue. When that issue gets solved/implemented it wouldn't be hard at all to have this logging stuff publish to it.

radare:
Yes we want this in rutil

thestr4ng3r:
> Yes we want this in rutil

Why? I would think writing stuff to a logfile, especially since it is controlled by eval vars, which are set from core anyway, would fit better into core than util.

That's also what I wrote here in the pad:
> By default, core should register some default logging function in the RConsContext for each task (or at least the main task). This function should then handle logging like Aaron implemented, i.e. respecting the eval vars for user configuration, logging to a file and so on.

Anyway, I don't think it would be much work to just move the options to RCore and the implementation to a default callback.

libr/include/r_util/r_logging.h
thestr4ng3r: For the default one, probably. But for the task-local stuff I it will be important.

radare:
Util is accessible from core. And is something that must be used from all modules not just core

radare:
But i understand the thread, cons singleton problem which can be only solved by attaching it to rcoretasks..

thestr4ng3r:
@radare The idea of this is the following: since you want logging to be called from any part of the code, the functions you call for logging have to be in **util**. However, the actual heavy stuff that writes the log to a file or whatever should be in core and called by util, which eventually calls the callback in the RConsContext. See what I mean?

radare:
This can be solved by sharing a void* with the associated core instance for the specific log message, this can be set from core and switched when task changes if needed. and this void* will be used when a callback is executed in the log api, and this callback will be implemented in core. but the log api must be in util. this sounds good to you?

thestr4ng3r:
Uhm, sorry I don't get what you mean. Especially this part:
> sharing a void* with the associated core instance for the specific log message

radare:
please solve the conflicts and rebase

radare:
ping

libr/cons/cons.c
```diff
@@ -367,6 +368,18 @@ R_API bool r_cons_enable_mouse(const bool enable) {
 #endif
 }
 
+// Stub function that cb_logfunc_print gets assigned to in logging.c by default
+// This allows Cutter to set per-task logging redirection
+R_API void r_cons_logging_stub(const char *output, const char *funcname, const char *filename,
+ unsigned int lineno, unsigned int level, const char *fmtstr, ...) {
```
radare: use ut32 instead of. unsigned int

libr/cons/cons.c
radare: passing args without va_start? this seems like UB

libr/cons/cons.c
```diff
@@ -367,6 +368,18 @@ R_API bool r_cons_enable_mouse(const bool enable) {
...
+	if (!I.null && I.context->logging_callback) {
+		va_list args;
+		I.context->logging_callback (output, funcname, filename, lineno, level, fmtstr, args);
+	} else {
+		r_cons_printf ("%s", output);
```
radare: r_cons_strcat, there's no need to format

libr/core/cconfig.c
radare: use r_num_math()

libr/core/cconfig.c
```diff
@@ -2290,6 +2290,65 @@ static int cb_dbgsnap(void *user, void *data) {
...
+	if (*bad_data) {
+		return false;
+	}
+	r_logging_set_traplevel (ival);
+	return true;
+}
+
+static int cb_logging_config_file(void *coreptr, void *nodeptr) {
+	RConfigNode *node = (RConfigNode *)nodeptr;
+	char *value = node->value;
```
radare: that can be const char *

libr/include/r_util.h
```diff
@@ -41,7 +41,7 @@ int gettimeofday (struct timeval* p, void* tz);
 #include "r_util/r_des.h"
 #include "r_util/r_file.h"
 #include "r_util/r_hex.h"
-#include "r_util/r_log.h"
+#include "r_util/r_logging.h"
```
radare: i would keep r_log.h as well as the api names

libr/util/logging.c
```diff
@@ -0,0 +1,127 @@
+/* radare - LGPL - Copyright 2007-2013 - pancake */
+
+#define LOGGING_CONFIGSTR_SIZE 512
```
radare: why set a limit for this? just use the heap

libr/util/logging.c
```diff
@@ -0,0 +1,127 @@
...
+	// Setup varadic arguments
+	va_list args;
+	va_start (args, fmtstr);
+
+	// TODO: Colors
+
+	// Build output string with src info, and formatted output
+	char output_buf[LOGGING_OUTPUTBUF_SIZE] = ""; // Big buffer for building the output string
+	const char *tag = R_BETWEEN (0, level, sizeof (level_tags) - 1) ? level_tags[level] : "";
+	int offset = snprintf (output_buf, LOGGING_OUTPUTBUF_SIZE, "%s", tag);
```
radare: use r_str_newf and make output_buf in the heap instead of global "big" buffer

libr/util/logging.c
ret2libc: I agree with this. Only thing I could think of is that in this way we could use logging even when the memory is finished (if we care about this).

libr/util/logging.c
```diff
@@ -0,0 +1,127 @@
...
+	[R_LOGLVL_VERBOSE]   = "VERBOSE: ",
+	[R_LOGLVL_DEBUG]     = "DEBUG: ",
+	[R_LOGLVL_INFO]      = "INFO: ",
+	[R_LOGLVL_WARN]      = "WARNING: ",
+	[R_LOGLVL_ERROR]     = "ERROR: ",
+	[R_LOGLVL_FATAL]     = "FATAL: ",
+	[R_LOGLVL_FATALTRAP] = "FATAL: "
+};
+
+/* TODO: Might want to create a r_logging_use_envvars function so that programs
```
ret2libc: I think this could be removed. Programs that do not use core just need to call `r_logging_set_*` functions with the arguments they want and they should be goood to go.

libr/util/logging.c
```diff
@@ -0,0 +1,127 @@
...
+	const char *tag = R_BETWEEN (0, level, sizeof (level_tags) - 1) ? level_tags[level] : "";
+	int offset = snprintf (output_buf, LOGGING_OUTPUTBUF_SIZE, "%s", tag);
+	if (cfg_logsrcinfo) {
+		offset += snprintf (output_buf + offset, LOGGING_OUTPUTBUF_SIZE - offset, "%s in %s:%i: ", funcname, filename, lineno);
+	}
+	vsnprintf (output_buf + offset, LOGGING_OUTPUTBUF_SIZE - offset, fmtstr, args);
+
+	// Actually print out the string with our callbacks
+	if (cb_logfunc_print) {
+		cb_logfunc_print (output_buf, funcname, filename, lineno, level, fmtstr, args);
```
ret2libc: do we need to pass the `output_buf` given that we are passing anyway all the other arguments?

ret2libc: also, AFAIK when you reuse a va_list more than one time, you should copy it and use the copy for the second time, otherwise the first call to vsnprintf may "exhaust" the va_list.

libr/include/r_util.h
mcd1992: Should we rename core/log.c to something more descriptive then to avoid conflicts? There's also a lot of existing usage of `r_core_log` elsewhere related to it. Don't want to make it too confusing having a `r_core_log` and `r_log`.

libr/util/logging.c
mcd1992: Ah yea this comment was before I did the r_logging_set changes. Thanks.

libr/util/logging.c
mcd1992: The output_buf param is so you don't have to re-format the args into a string since its already been done in `_r_logging_internal`.

libr/util/logging.c
ret2libc: Reason why I would keep using snprintf instead of r-str functions is because in this way you can use rasserts and logs in r-str without worrying about infinite recursion or similar issues

ret2libc:
You should keep a separate function for rasserts logs, as it was before, because the macros in rassert.h are (of course) inlined and adding the call to rsys breakpoint there would increase a lot the code size

mcd1992:
What seems to be the verdict on r_util vs r_core split?

My thinking is that everything that *can* go into r_util should. This way it can be used without a r_core/r_cons dependency at minimum. If r_core is loaded the r_cons_new() call in r_core_init() will change the default callback in util/logging.c to point to r_cons_logging_stub which will handle the per-task callback[s]. I don't see too much of a reason to split it between r_core and r_util unless its for organization in which case it will be kinda weird to have a core/logging.c/h and a util/logging.c/h.

Is there something I may be overlooking?

Also TODO:
- [ ] Move callbacks to RList with context/*user support - https://pastebin.com/QPhvTmfy
- [ ] Avoid undefined behavior with va_args
- [x] Use radare's ut32 types
- [x] Replace r_cons_printf with r_cons_strcat
- [ ] Use heap for buffer instead of stack (r_str_newf)
- [ ] r_assert trap only on R_DEBUG_ASSERT

mcd1992:
@ret2libc I'll do that; bring back assert.c and just make it re-use the loglevel in logging.h.

mcd1992:
Just squashing to try and cleanup a bit, not done yet of course.

radare:
See my revent api as reference for having an instance in core and let other modules call this log api with the instance created from core as an idea

libr/util/logging.c
ret2libc: @mcd1992 please address the second comment about va_list/va_copy...

libr/core/cconfig.c
```diff
@@ -2632,6 +2691,37 @@ R_API int r_core_config_init(RCore *core) {
...
+	/* cfg.logging */
+#if DEBUG
+#define R_LOGGING_DEFAULTLEVEL R_LOGLVL_WARN
+#else
+#define R_LOGGING_DEFAULTLEVEL R_LOGLVL_ERROR
+#endif
+	// R2_LOGLEVEL / cfg.logging.level
+	p = r_sys_getenv ("R2_LOGLEVEL");
+	SETICB ("cfg.logging.level", p ? atoi(p) : R_LOGGING_DEFAULTLEVEL, cb_logging_config_level, "Output logging level "\
+	 "(0:SILLY, 1:VERBOSE, 2:DEBUG, 3:INFO, 4:WARN, 5:ERROR, 6:FATAL, 7:FATALTRAP)"
```
radare: silly?

radare: fataltrap?

libr/include/r_util/r_logging.h
```diff
@@ -0,0 +1,74 @@
+#ifndef R_LOGGING_H
```
radare: rename to r_log.h

libr/util/assert.c
```diff
@@ -4,10 +4,10 @@
  * It prints a message to the log and it provides a single point of entrance in
  * case of debugging. All r_return_* functions call this.
  */
-R_API void r_assert_log(RLogLevel level, const char *fmt, ...) {
+R_API void r_assert_log(RLoggingLevel level, const char *fmt, ...) {
```
radare: RLogLevel

libr/util/logging.c
radare: Hurd dislikes fixed size buffers for paths, but yeah i assume there must be a sane limit. even more for logging

libr/util/logging.c
```diff
@@ -0,0 +1,123 @@
...
+static char cfg_logfile[LOGGING_CONFIGSTR_SIZE] = ""; // Output text to filename
+static const char *level_tags[] = { // Log level to tag string lookup array
+	[R_LOGLVL_SILLY]     = "SILLY: ",
+	[R_LOGLVL_VERBOSE]   = "VERBOSE: ",
+	[R_LOGLVL_DEBUG]     = "DEBUG: ",
+	[R_LOGLVL_INFO]      = "INFO: ",
+	[R_LOGLVL_WARN]      = "WARNING: ",
+	[R_LOGLVL_ERROR]     = "ERROR: ",
+	[R_LOGLVL_FATAL]     = "FATAL: ",
+	[R_LOGLVL_FATALTRAP] = "FATAL: "
```
radare: wrong text

libr/util/logging.c
```diff
@@ -0,0 +1,123 @@
...
+};
+
+// cconfig.c configuration callback functions below
+R_API void r_logging_set_level(RLoggingLevel level) {
```
radare: i. would have a single function. r_log_set_options(enum RLogOption, RLogLevel level)

libr/util/logging.c
```diff
@@ -0,0 +1,123 @@
...
+	cfg_loglvl = level;
+}
+
+R_API void r_logging_set_traplevel(RLoggingLevel level) {
+	cfg_logtraplvl = level;
+}
+
+R_API void r_logging_set_file(const char *filename) {
+	int value_len = strnlen (filename, LOGGING_CONFIGSTR_SIZE) + 1;
```
radare: use r_str_nlen, strnlen is. not portable

libr/util/logging.c
radare: ok, but this is just an r_str_ncpy

libr/util/logging.c
```diff
@@ -0,0 +1,123 @@
...
+		offset += snprintf (output_buf + offset, LOGGING_OUTPUTBUF_SIZE - offset, "%s in %s:%i: ", funcname, filename, lineno);
+	}
+	vsnprintf (output_buf + offset, LOGGING_OUTPUTBUF_SIZE - offset, fmtstr, args);
+
+	// Actually print out the string with our callbacks
+	if (cb_logfunc_print) {
+		cb_logfunc_print (output_buf, funcname, filename, lineno, level, fmtstr, args);
+	} else {
+		if (level < R_LOGLVL_ERROR) {
+			fprintf (stdout, "%s", output_buf);
```
radare: imho. all log messages should go to stderr

radare: but. maybe we should have a callback to handle how to. print log messages (from core where this thing will be instantiated) and allow to send via udp sockets, use r_log, or whatever (cutter?) decides

libr/util/logging.c
```diff
@@ -0,0 +1,123 @@
...
+	if (cfg_logfile[0] != 0x00) {
+		FILE *file = r_sandbox_fopen (cfg_logfile, "a+"); // TODO: Optimize (static? Needs to remake on cfg change though)
+		if (!file) {
+			file = r_sandbox_fopen (cfg_logfile, "w+");
+		}
+		if (file) {
+			fprintf (file, "%s", output_buf);
+			fclose (file);
+		} else {
+			eprintf ("_r_log_internal failed to write to file: %s\n", cfg_logfile);
```
radare: use __FUNCTION__ instead of hardcoding the funciton name here, but in fact this should be an assert line

libr/util/logging.c
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}
+
+	va_end (args);
```
radare: not sure if its fine to va_end before the end of the function..

libr/util/meson.build
```diff
@@ -25,7 +25,7 @@ files = [
   'json_indent.c',
   'lib.c',
   'list.c',
-  'log.c',
+  'logging.c',
```
radare: keep it as log.c

libr/util/logging.c
ret2libc: This is the fallback. There is already the callback to choose to print it however you want, but it still makes sense to have a fallback solution when there's no callback setup (for example if you are using this before the callback is set or in other projects where r_core is not used)

libr/util/logging.c
ret2libc: it should be AFAIK, as long as you don't reuse args.

radare:
if its fallback, then just use stderr

ret2libc:
> What seems to be the verdict on r_util vs r_core split?

My *personal* opinion is that what you did is good enough. I see it being easily useful from RCore and from outside RCore.

> TODO:

I think the PR is good enough as it is + fixes to comments done by @radare and fix the va_copy thing. The other improvements that you mentioned in the TODO list, IMHO, could be done in a second PR because they don't prevent us from start using the feature.

libr/core/cconfig.c
mcd1992: SILLY was from zf_log I just kept it. FATAL was so a dev could have a fatal error but let it continue on, FATALTRAP being the FATAL that calls r_sys_breakpoint. I guess now that I added in cfg.logging.traplevel it would be possible to make FATAL always breakpoint and a dev/user could just raise the traplevel cfg to stop it from breakpoint-ing.

Would you rather I do that?

libr/core/cconfig.c
```diff
@@ -2632,6 +2691,37 @@ R_API int r_core_config_init(RCore *core) {
 	SETPREF ("cfg.wseek", "false", "Seek after write");
 	SETCB ("cfg.bigendian", "false", &cb_bigendian, "Use little (false) or big (true) endianness");
 
+	/* cfg.logging */
+#if DEBUG
```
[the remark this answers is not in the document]
ret2libc: @radare ? how can @mcd1992 distinguish between a debug build and a release one? Is there already something to do that? Shall we just default to WARNING for this PR and change it in a following PR?

mcd1992:
@radare what do you prefer for the `logging` to `log` rename. Right now there is a core/log.c that I want to avoid 'collision' with. It has a config var `cfg.log` that I avoided by making mine `cfg.logging.*` and the actual file itself `core/log.c` might become an issue if we split the code between core and util.

My thoughts are to rename the existing `core/log.c` to `core/corelog.c` and its config var to `cfg.corelog`, this seems redundant but makes sense to me since all of its R_API exports are `r_core_log_`. This way if we do decide to split the logging stuff into core and util we'll have `core/log.c` available and `util/log.c` will be related to it. If 'corelog' seems weird then maybe we could use 'textlog', 'chatlog' or 'tlog' instead if that makes more sense (just going by what I see in `T?`).

Right now since I've got everything in util (and cons) there's no technical issue with having `core/log.c` and `util/log.c` being different APIs other than the confusion it might cause. That and the config var `cfg.logging.*` vs `cfg.log`.

Any preference?

radare:
i dont see why core/log should conflict with util/log. the api prefix is completely different, and in any case we should. rebrand the core/log apis because its more about timestamped messages in a queue than a raw appending log

> On 25 Oct 2018, at 17:38, Aaron <notifications@github.com> wrote:
> 
> @radare <https://github.com/radare> what do you prefer for the logging to log rename. Right now there is a core/log.c that I want to avoid 'collision' with. It has a config var cfg.log that I avoided by making mine cfg.logging.* and the actual file itself core/log.c might become an issue if we split the code between core and util.
> 
> 

i dont see any collision in there and this cfg.log variable is only used once, and i. bet nobody. knew or even cared about this.

> My thoughts are to rename the existing core/log.c to core/corelog.c and its config var to cfg.corelog, this seems redundant but makes sense to me since all of its R_API exports are r_core_log_. This way if we do decide to split the logging stuff into core and util we'll have core/log.c available and util/log.c will be related to it. If 'corelog' seems weird then maybe we could use 'textlog', 'chatlog' or 'tlog' instead if that makes more sense (just going by what I see in T?).
> 
> 

name it clog.c if you want, but there’s no problem in having the same filename in. 2 different modules
> Right now since I've got everything in util (and cons) there's no technical issue with having core/log.c and util/log.c being different APIs other than the confusion it might cause. That and the config var cfg.logging.* vs cfg.log.
> 
> 
your api should be r_log_,  and replace the current util/log.c. which. is currently not used anywhere

radare:
There’s no difference between debug and release builds. it was at some point to avoid git pulling capstone.

Just honor the same config option as the one you added for the asserts

mcd1992:
Okay that should fix most of the review points other than:
- [ ] Use heap for buffer instead of stack (r_str_newf)
- [ ] Move callback system to RVector with context/*user support - https://pastebin.com/QPhvTmfy
- [ ] Use similar system to REvent API
- [ ] Add r_log_set_options to set all options with one function

I'll try and get those done next week.

libr/util/assert.c
```diff
@@ -7,7 +7,7 @@
 R_API void r_assert_log(RLogLevel level, const char *fmt, ...) {
 	va_list args;
 	va_start (args, fmt);
-	r_vlogf (level, fmt, args);
+	R_LOG (level, NULL, fmt, args);
```
ret2libc: this is wrong because:
1) it uses NULL where there should be `fmt` (which makes the function segfault)
2) _r_log_internal called by R_LOG does not accept a va_list but variadic arguments.

The thing should be refactored a bit so to provide a R_VLOG version as it was before, that accepts a va_list instead of multiple arguments, and `r_assert_log` should call the R_VLOG instead of R_LOG. But... We've already asked many things to you in this PR and I'm willing to clean this up myself when merging if you don't mind. :)

mcd1992: I changed R_LOG back to `R_LOG(lvl, tag, fmtstr, ...)` so you can specify a custom tag again with it. The other macros just pass NULL so they are still `(fmtstr, ...)`.

Also I guess I need to read up on how the varadic stuff works in C, I thought the `...` was some sort of inferred va_list.

PR #11787, opened 2018-10-10
Support of Apple/LLVM blocks extension

devnexen:
Apple toolchain and LLVM provides closure extension for C/C++,
using as possible the most reliable identifier rather than
functions owners where names can appear too generics.


radare:
Cool! can you add a test?

PR #11788, opened 2018-10-10
Fix memleaks in ragg2

bannsec:
Some ragg2 memleaks.


libr/egg/egg.c
```diff
@@ -512,6 +512,7 @@ R_API void r_egg_finalize(REgg *egg) {
 	RBuffer *b;
 	RListIter *iter;
 	if (!egg->bin->buf) {
+		free (egg->bin);
```
radare: use r_buf_free

PR #11792, opened 2018-10-11
Fix few code paths.

devnexen:


libr/asm/arch/cris/gnu/cris-dis.c
```diff
@@ -201,39 +201,39 @@ get_opcode_entry (unsigned int insn,
...
       memset (bdapq_m2_prefixes, 0, 65536 * sizeof (bdapq_m2_prefixes[0]));
 
       bdapq_m4_prefixes
-	= malloc (65536 * sizeof (const struct cris_opcode **));
+	= malloc (65536 * sizeof (bdapq_m4_prefixes[0]));
       if (!bdapq_m4_prefixes) {
 	      return NULL;
       }
 
       memset (bdapq_m4_prefixes, 0, 65536 * sizeof (bdapq_m4_prefixes[0]));
```
radare: use calloc nstead of malloc+memset

radare: anyway this is gnu code, dont touch it too much because it will break when we sync again

libr/core/cmd.c
```diff
@@ -2260,7 +2260,6 @@ static int r_core_cmd_subst_i(RCore *core, char *cmd, char *colon, bool *tmpseek
 							r_cons_printf ("%s\n", res);
 							free (res);
 						}
-						free (res);
```
radare: lgtm

PR #11801, opened 2018-10-12
syscall: do not reload sdb database if nothing changes

### LINKED ISSUE #10737: r2 periodically tries to open sysregs sdb that doesn't exist
https://github.com/radare/radare2/blob/master/libr/syscall/syscall.c#L120

ret2libc:
fixes #10737

### DIFF
--- a/libr/syscall/syscall.c
+++ b/libr/syscall/syscall.c
@@ -157,8 +157,6 @@ R_API bool r_syscall_setup(RSyscall *s, const char *arch, int bits, const char *
 		if (dbName) {
 			s->db = openDatabase (s->db, dbName);
 			free (dbName);
-		} else {
-			eprintf ("Couldn't create dbName string.\n");
 		}
 	}
 
@@ -169,8 +167,6 @@ R_API bool r_syscall_setup(RSyscall *s, const char *arch, int bits, const char *
 			sdb_free (s->srdb);
 			s->srdb = openDatabase (NULL, dbName);
 			free (dbName);
-		} else {
-			eprintf ("Couldn't create dbName string.\n");
 		}
 	}
 	if (s->fd) {
--- a/libr/syscall/syscall.c
+++ b/libr/syscall/syscall.c
@@ -38,6 +38,8 @@ R_API void r_syscall_free(RSyscall *s) {
 		sdb_free (s->srdb);
 		sdb_free (s->db);
 		free (s->os);
+		free (s->cpu);
+		free (s->arch);
 		free (s);
 	}
 }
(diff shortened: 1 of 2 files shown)


libr/syscall/syscall.c
```diff
@@ -65,17 +67,49 @@ static Sdb *openDatabase(Sdb *db, const char *name) {
 	return db;
 }
 
+static inline bool syscall_reload_needed(RSyscall *s, const char *os, const char *arch, int bits) {
+	bool is_equal = true;
+	is_equal &= (s->os && !strcmp (s->os, os));
```
radare: Thats innefficient. If one conditional fails we dont need to go further

libr/syscall/syscall.c
```diff
@@ -113,17 +147,20 @@ R_API bool r_syscall_setup(RSyscall *s, const char *arch, int bits, const char *
...
 	}
 
-	char *dbName = r_str_newf (R_JOIN_2_PATHS ("syscall", "%s-%s-%d"),
-		os, arch, bits);
-	s->db = openDatabase (s->db, dbName);
-	free (dbName);
+	if (syscall_changed) {
+		char *dbName = r_str_newf (R_JOIN_2_PATHS ("syscall", "%s-%s-%d"),
+			os, arch, bits);
+		s->db = openDatabase (s->db, dbName);
```
radare: Check for null

libr/syscall/syscall.c
```diff
@@ -113,17 +147,20 @@ R_API bool r_syscall_setup(RSyscall *s, const char *arch, int bits, const char *
...
 
-	dbName = r_str_newf (R_JOIN_2_PATHS ("sysregs", "%s-%d-%s"),
-		arch, bits, cpu ? cpu: arch);
-	sdb_free (s->srdb);
-	s->srdb = NULL;
-	s->srdb = openDatabase (s->srdb, dbName);
-	free (dbName);
+	if (sysregs_changed) {
+		char *dbName = r_str_newf (R_JOIN_2_PATHS ("sysregs", "%s-%d-%s"),
+			arch, bits, cpu);
```
radare: Check for null

libr/syscall/syscall.c
```diff
@@ -113,17 +151,28 @@ R_API bool r_syscall_setup(RSyscall *s, const char *arch, int bits, const char *
...
-		os, arch, bits);
-	s->db = openDatabase (s->db, dbName);
-	free (dbName);
+	if (syscall_changed) {
+		char *dbName = r_str_newf (R_JOIN_2_PATHS ("syscall", "%s-%s-%d"),
+			os, arch, bits);
+		if (dbName) {
+			s->db = openDatabase (s->db, dbName);
+			free (dbName);
+		} else {
```
radare: Remove this eprintf. We should put an assert line in r_str_newf instead of printing an error in every single place that something can fail. Just keep the if to avoid crashes when the heap is under pressure or buggy

libr/syscall/syscall.c
```diff
@@ -113,17 +151,28 @@ R_API bool r_syscall_setup(RSyscall *s, const char *arch, int bits, const char *
...
-	free (dbName);
+	if (sysregs_changed) {
+		char *dbName = r_str_newf (R_JOIN_2_PATHS ("sysregs", "%s-%d-%s"),
+			arch, bits, cpu);
+		if (dbName) {
+			sdb_free (s->srdb);
+			s->srdb = openDatabase (NULL, dbName);
+			free (dbName);
+		} else {
+			eprintf ("Couldn't create dbName string.\n");
```
radare: Same here

libr/syscall/syscall.c
ret2libc: No.... assert are not meant to handle runtime errors. Removing this print just means that in case something bad happens you won’t know what went bad.

radare:
Why are we getting tons of aasert log messages at runtime then? If we add a string for every single failing alocation:

* console will be filled with crap everytime this happens
* if so is better to just close because the console will be useless
* r2 will be much bigger in size
* asserts are suposed to be used to ensure those situations and it fits perfectly with the assert api you did. But just using it as post condition instead of precondition
* vala does that too

ret2libc:
@radare removed eprintf.

PR #11802, opened 2018-10-12
disasm.c: Add ds_comment_start/middle/end + use case

kazarmy:


ret2libc:
Image with results?

kazarmy:
huh, no results ... it's just a way to split the following ds_comment calls: https://github.com/radare/radare2/blob/3ccc90d4de32f269cd9e94446a7bf3f600604f3a/libr/core/disasm.c#L3671-L3680 into 3 parts so that I don't have to repeat myself like the above

PR #11803, opened 2018-10-12
fix asm.describe=1 when asm.cmt.right=1

rfc2119:
I don't know precisely where should I write asm description to screen. Please remove/edit wherever you feel like.


radare:
i fixed the commit message because it was wrong. tested and merged. thanks

PR #11804, opened 2018-10-13
Do not show types/vars comments in the hexadecimal mode view

### LINKED ISSUE #11427: Do not show types/vars comments in the hexadecimal mode view
```
r2 /bin/ls
aaa; s main; VpP
```
then press '|' and type `pd` then press enter.

They do not make any sense in the hexdump mode and can clog the output:
image

sivaramaaa:
Fixes #11427 

comment

### DIFF
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -1505,7 +1505,7 @@ static void annotated_hexdump(RCore *core, const char *str, int len) {
 
 		if (core->print->use_comments) {
 			for (j = 0; j < nb_cols; j++) {
-				const char *comment = core->print->get_comments (core->print->user, addr + j);
+				const char *comment = core->print->get_comments (core->print->user, addr + j, false);
 				if (comment) {
 					r_cons_printf (" ; %s", comment);
 				}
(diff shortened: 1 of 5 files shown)


radare:
why not use px instead of pxc instead of making pxc not show comments?

sivaramaaa:
> why not use px instead of pxc instead of making pxc not show comments?

no px doesn't show comment at all , but i need to remove only type related command and keep other comments like section flags etc !

**Before**

sss

radare:
The issue was complaining that comments was overlapping the disasm at the right. So this wont solve the problem.

radare:
Lgtm as an improvement to pxc but i dont think this fixes the problem

sivaramaaa:
> The issue was complaining that comments was overlapping the disasm at the right. So this wont solve the problem.

Ahh now i get it , let me see , what can be done to fix the issue !

XVilka:
Seems there is one failing test @sivaramaaa - `[XX] db/esil/apple iphone-arm-esil pdsf  208`

sivaramaaa:
I am closing this PR right now , it think it's related to panels, so @cyanpencil or @Vane11ope could do a better fix or I could get help from them to fix it !

PR #11813, opened 2018-10-14
Avoid hardcoding prefix in global.mk

radare:


EvanCarroll:
This does nothing because it's the default.

The problem is you override that in build.sh. Not because the default is explicitly mentioned in global.mk.

Issue #9240, opened 2018-01-25
./sys/install.sh to /usr/bin not /usr/local/bin (violation of FHS)


EvanCarroll:
### Work environment

Using Linux, after I run `./configure` I get

```
Final report:
 - PREFIX = /usr/local
 - HAVE_LIB_GMP = 0
 - HAVE_OPENSSL = 0
 - USE_CAPSTONE = 0
 - HAVE_FORK = 1
 - VERSION = 2.3.0-git
 - USE_LIB_ZIP = 0
 - USE_LIB_MAGIC = 0
 - DEBUGGER = 1
 - CC = gcc
 - USERCC = gcc
 - USEROSTYPE = gnulinux
 - LIBVERSION = 2.3.0-git
 - BUILD = x86_64-unknown-linux-gnu
 - HOST = x86_64-unknown-linux-gnu
 - TARGET = x86_64-unknown-linux-gnu
```

### Expected behavior

Installs to /usr/local/bin

### Actual behavior

Installs to /usr/bin

This creates problems with sdbs aren't found,

> Cannot find /usr/local/share/radare2/2.3.0-git/format/dll/mySdb.sdb

If I don't use `./sys/install.sh` and I 

    ./configure
    ./make
    ./make install

Everything works fine.

radare:
did you git clean -xdf or make clean/mrproper before running ./configure ?

just tested this and the behaviour is correct if you run configure without arguments it will be installed in /usr/local, as any other autotools program will do.

sys/install.sh works fine too, i think your problem is because you iddnt cleaned the builddir, which is the reason why i did sys/install.sh

also you can pass the prefix as argument to sys/install.sh

davidpolverari:
@EvanCarroll, did it work for you after git clean -xdf or make clean/mrproper before running ./configure? Maybe this can be closed by now.

EvanCarroll:
Nope this is still a problem. Radare is installing into `/usr/bin`.

EvanCarroll:
Here is the problem in `build.sh` you have,

```
 27 else
 28   DEFAULT_PREFIX=/usr
 29   [ -n "${PREFIX}" -a "${PREFIX}" != /usr ] && \
 30     CFGARG="${CFGARG} --with-rpath"
 31 fi
```

that should be,

```
 27 else
 28   DEFAULT_PREFIX=/usr/bin
 29   [ -n "${PREFIX}" -a "${PREFIX}" != /usr ] && \
[...]

PR #11814, opened 2018-10-15
Remove all traces of prefix override

EvanCarroll:
This removes all traces of the improperly set `PREFIX`. If `PREFIX` is set wrong for your platform or distro, file a bug and we'll fix it in the `./configure` script. However the default *should* be correct in all cases, as that's the explicit goal of autotools and how **EVERY** GNU utility works. This can actually be set in your distro too, so if for some reason your whole distro wants its users to install somewhere other than `/usr/local` they can simply [update the copy of `general.m4` with a different `ac_default_prefix`](https://github.com/autotools-mirror/autoconf/blob/487d6aaaa4fe45001d30c439ea6240fc74f1b5d7/lib/autoconf/general.m4#L405)

Either way, [randomly guessing and asserting that Radare install into `/usr/bin` is just wrong](https://github.com/radare/radare2/issues/9240#issuecomment-404976781). Moreover, if the distro installs to `/usr/bin` and manages those files (as most do) Radare installing there (via `build.sh`) can actually break the system entirely, as in the case of Ubuntu and Debian. Almost as bad, this also leaves the distro free to fuck up radare with it's own copy or with files named poorly that conflict with radare's own poorly named files.

* When the user installs utilities for the system, they should be installed in `/usr/local/bin`.
* When the system installs utilities for the system, they should be installed in `/usr/bin`

This brings `/sys/build.sh` which `install.sh` calls, in alignment with the result of `./configure; make; make install;`

### DIFF
--- a/sys/build.sh
+++ b/sys/build.sh
@@ -59,7 +59,6 @@ A=$(dirname "$PWD/$0")
 cd "$A" && cd .. || exit 1
 
 if [ "`uname`" = Darwin ]; then
-	DEFAULT_PREFIX=/usr/local
 	# purge previous installations on other common paths
 	if [ -f /usr/bin/r2 ]; then
 		type sudo || NOSUDO=1
@@ -67,16 +66,11 @@ if [ "`uname`" = Darwin ]; then
 		[ -n "${NOSUDO}" ] && SUDO=
 		# purge first
 		echo "Purging r2 installation..."
-		./configure --prefix=/usr > /dev/null
+		./configure > /dev/null
 		${SUDO} ${MAKE} uninstall > /dev/null
 	fi
-else
-	DEFAULT_PREFIX=/usr
-	[ -n "${PREFIX}" -a "${PREFIX}" != /usr ] && \
-		CFGARG="${CFGARG} --with-rpath"
 fi
 
-[ -z "${PREFIX}" ] && PREFIX="${DEFAULT_PREFIX}"
 
 case "$1" in
 -h)
@@ -91,7 +85,6 @@ case "$1" in
 	CFGARG="${CFGARG} $*"
 	;;
 *)
-	PREFIX="$1"
 	;;
 esac
 
@@ -140,7 +133,7 @@ if [ -z "${KEEP_PLUGINS_CFG}" ]; then
 fi
 unset DEPS
 pwd
-./configure ${CFGARG} --prefix="${PREFIX}" || exit 1
+./configure ${CFGARG} || exit 1
 ${MAKE} -s -j${MAKE_JOBS} MAKE_JOBS=${MAKE_JOBS} || exit 1
 if [ "`uname`" = Darwin ]; then
 	./sys/macos-cert.sh


EvanCarroll:
See #9240  for more information

sys/build.sh
```diff
@@ -59,24 +59,18 @@ A=$(dirname "$PWD/$0")
...
 		# purge first
 		echo "Purging r2 installation..."
-		./configure --prefix=/usr > /dev/null
+		./configure > /dev/null
 		${SUDO} ${MAKE} uninstall > /dev/null
 	fi
-else
-	DEFAULT_PREFIX=/usr
-	[ -n "${PREFIX}" -a "${PREFIX}" != /usr ] && \
-		CFGARG="${CFGARG} --with-rpath"
```
radare: this will break the behaviour in many linux distros

radare:
also, in r2 we dont use autotools, so, all those m4 comments are totally unrelated

ret2libc:
What about this? Why wasn't it merged? I think it makes sense to *not assume* the default prefix is `/usr`, for the reasons explained in https://github.com/radare/radare2/pull/11814#issue-222764653 .

@radare ?

radare:
Yes i also agree on this. But this pr is breaking other things. As i said the prefix thing is not the real problem

ret2libc:
So if there are no actions on this, I'd say let's close it. It is not reviewed, it is not modified...

radare:
it is reviewed, i got no feedback, because he got angry. so i'll try to take care of this, but not for 3.4,  i have already changed some stuff in the build system that will be helpful to avoid problems with multiple installations of r2. and i idont want to push many changes in the build system on a single release to avoid stacking problems. also i iwill add more comments here to clarify

sys/build.sh
```diff
@@ -59,24 +59,18 @@ A=$(dirname "$PWD/$0")
...
-	DEFAULT_PREFIX=/usr/local
 	# purge previous installations on other common paths
 	if [ -f /usr/bin/r2 ]; then
 		type sudo || NOSUDO=1
 		[ "$(id -u)" = 0 ] || SUDO=sudo
 		[ -n "${NOSUDO}" ] && SUDO=
 		# purge first
 		echo "Purging r2 installation..."
-		./configure --prefix=/usr > /dev/null
+		./configure > /dev/null
```
radare: the reason for that is to purge the installation found in /usr before installing further, the problem is that if u have r2 installed from the distro it will basically purge it, so this change makes sense to avoid killing the linux installation.. problem iis when you dont use linux, this will basically remove the previous installation of r2, or basically remove the one from the pkgsystem, because bsd and mac install in /usr/local which iis the default, therefor this is moving the problem from linux to bsd

sys/build.sh
radare: the rpath is necessary only when building outside the default path.. this is, if u install in /usr/local, you NEED rpath, so this change is wrong

sys/build.sh
```diff
@@ -140,7 +133,7 @@ if [ -z "${KEEP_PLUGINS_CFG}" ]; then
 fi
 unset DEPS
 pwd
-./configure ${CFGARG} --prefix="${PREFIX}" || exit 1
```
radare: why are you removing the ability to the user to specify a different prefix?

sys/build.sh
EvanCarroll: > problem iis when you dont use linux, this will basically remove the previous installation of r2, or basically remove the one from the pkgsystem, because bsd and mac install in /usr/local which iis the default, therefor this is moving the problem from linux to bsd

Linux installs into /usr/local too. This doesn't change anything. `make install` and `make uninstall` will just work. That's the point. FreeBSD installs ports packages to `/usr/local` too (afaik). But the point is without touching `--prefix` it you don't move the problem, you get rid of it.

> because bsd and mac install in /usr/local

Even if so, why is this a problem.

sys/build.sh
condret: WAT, I always run with --prefix, don't remove this

radare:
Hey, have a look at my PR, can you test it? https://github.com/radareorg/radare2/pull/15891

im closing this issue because there are conflicts, but we can continue the discussion in the other PR, thanks

PR #11816, opened 2018-10-15
Fix build for GNU/Hurd

radare:
Avoid the use of PATH_MAX


radare:
the thing is.. is solaris realpath() working with NULL as argument?

radare:
confirmed this behaviour is consistent on solaris, openbsd, macos and gnu/linux

PR #11820, opened 2018-10-15
WIP: ARM/ESIL: Add and fix emulation of addw, add, adr

alvarofe:
before the patch:

0x00026186      7944           add r1, pc

after the patch:

0x0002618e      7944           add r1, pc   ; "sqlite3_backup_finish"

This will help to find more xref in arm binaries

### DIFF
--- a/libr/anal/p/anal_arm_cs.c
+++ b/libr/anal/p/anal_arm_cs.c
@@ -1484,6 +1484,7 @@ static int analop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len
 		break;
 	case ARM_INS_SADD16:
 	case ARM_INS_SADD8:
+	case ARM_INS_ADDW:
 	case ARM_INS_ADD:
 		op->type = R_ANAL_OP_TYPE_ADD;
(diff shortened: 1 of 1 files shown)


radare:
Can we have a test?

alvarofe:
I believe also the instruction `mov r0, pc` is not correctly emulated due to PC register, it does not take into account pipelining of arm. 

```
add r0, pc, 0 -> 0,8,$$,+,+,0xfffffffc,&,r0,=
mov r0, pc -> pc,r0,=
```
I will try to collect more cases and then send a PR with different commits for each instruction/emulation. Regarding the test cases, I need to create from scratch (it shouldn't be difficult) since I cannot share files.

radare:
looks red

alvarofe:
```
0x8700eca8 ? add    r5, pc, #1
0x8700ecac ? ldr    pc, [r5]
0x8700ecb0 ? tsthi    r4, #933888    ; 0xe4000
>>> p/x $r5
$2 = 0x8700ecb1
```

It seems instructions such as `add r0, pc, imm` does not take into account whether or not the result is align to 4 bytes - I checked on arm device. Therefore, reverting that change.

All in all, this PR adds `addw` which was not contemplated by radare and adding the correct esil expression for `adr`. 

I tested locally and it should be green now.

PR #11821, opened 2018-10-15
Fix static build with PIC

thestr4ng3r:


radare:
Wtf? Is the plugin working without those symbols?

radare:
i dont think this is a good fix, see how its solved by other code shsared between asm and anal

thestr4ng3r:
Should be fine if it compiles, `pic_midrange.o` is already in asm, see https://github.com/radare/radare2/blob/master/libr/asm/p/pic.mk
This is where the duplicate symbols came from.

radare:
Just changing the visibility to private should be enough. Because making asm depend on anal just to get this symbol is wrong. Mainly because its annissue with only one plugin

thestr4ng3r:
No, asm doesn't depend on anal, it is the other way around. And as far as I can tell, this is the case for other plugins as well.

thestr4ng3r:
wtf, no! The change is in anal. Asm still contains everything. How does removing stuff from **ANAL** make **ASM** depend on **ANAL**?

Anyway, I suppose the error on macOS is why this line was added in the first place. I just wonder why it appears. `pic_midrange_get_opcode` is defined just normal with R_API, so it should be exported.

thestr4ng3r:
Ok, anal doesn't link against asm too. Then I suppose just removing the R_APIs will do the trick.

radare:
without the R_API it was failing too but you can try

thestr4ng3r:
It does fix the static build for me on both Linux and macOS. Let's see what Travis has to say.

radare:
oops, we solved it in the same way :( its fixed in master. lets wait for googlefuzz to try again

PR #11822, opened 2018-10-15
Add zign hash using sha256

radare:


radare:
tests should be fixed in https://github.com/radare/radare2-regressions/pull/1502

libr/include/r_sign.h
```diff
@@ -43,12 +51,14 @@ typedef struct r_sign_item_t {
 	RSignGraph *graph;
 	ut64 offset;
 	RList *refs;
+	RSignHash *hash;
```
radare: why not just have char *hash here? keeping a struct for just one field is unnecessary

Maijin:
ping @cmatthewbrooks

cmatthewbrooks:
Thanks - I'm gonna try to repackage my code soon and will test this out then. The goal will be a separate repo with code and sigs with easy output for other disassemblers.

PR #11827, opened 2018-10-16
Add sys/release-notes.sh and define commit message conventions ·rules

radare:
screenshot 2018-10-16 at 03 24 09

### DIFF
--- a/sys/release-notes.sh
+++ b/sys/release-notes.sh
@@ -1,0 +1,63 @@
+#!/bin/sh
+#
+# This script parses the commit messages between the last 2 git tags
+# if the last commit contains the word "Release", otherwise it shows
+# all the changes between the last tag and HEAD.
+#
+# Commits are grouped in different sections specified in the commit
+# message with the middle-dot character '·'. The section names are
+# arbitrary and we may be careful to use them properly but having
+# in mind that this may change.
+#
+# Commits without any middle·dot in the message are discarted and
+# displayed in the "to review" section.
+#
+# The tool prints Markdown, no plans to support other formats.
+#
+# --pancake
+
+if [ -n "`git log -n 1 | grep Release`" ]; then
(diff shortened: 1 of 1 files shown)


radare:
The idea for this PR is to define commit message conventions. This proposal is pretty simple and i think it heavily simplifies the time we need to spend to write the release notes

radare:
This is.. we should append a middle-dot + tagname on all the commits with things that must appear in the release notes.

radare:
Go uses this char too. And its available in the catalan layout, but its accessible on all layouts using the altgr+.

radare:
All other chars i though could appear as r2 commands therefor not valid for commit messsages

ret2libc:
Altgr+. seems to work on IT layout, but not on EN-US layot.

ret2libc:
Altgr+. seems to work on IT layout, but not on EN-US layout.

radare:
https://github.com/marcioAlmada/yay/issues/1

it’s \xc2\xb7

https://middot.net

$ printf "·•" |hexdump -C
00000000  c2 b7 e2 80 a2

radare:
but im open for other suggestions

ret2libc:
What about `##` or `>>` or `<<` or just a particular line in the full commit log?
Like:
```
Add sys/release-notes.sh and define commit message conventions

Changelog-Section: rules
```

radare:
## is good, that was my first though, but i was afraid of github managing it somehow,. but seems like its not. so i can change to ## if you want

ret2libc:
After thinking a bit about this, I think using a separate line in the git commit message may be better because otherwise you are going to steal useful characters from the first line of the commit message, which usually shouldn’t be longer than 78 chars.

radare:
Its ok because we want short statements in the changelogs and hashtags will be short and only one per commit

PR #11829, opened 2018-10-16
r_util: use char * instead of void *

ret2libc:
On some architectures it is not valid to dereference a void * (I'm
getting this error when trying to compile on ppc64le and s390x), so
let's switch to char *, which is valid.


ret2libc:
@radare I think we should release 3.0.1 with this patch, otherwise it won't compile on some architectures.

radare:
Sgtm but i bet more distros will complain with things so better wait 1-2 days for this minor release

ret2libc:
sure ;)

PR #11830, opened 2018-10-16
Fix typo, use R_PACKED and use r2 types in the PDB parser

XVilka:


radare:
Are you sure this R_PACKED thing is not reintroducing the UBSAN problem? the PR lgtm

radare:
Why it wasnt like this two days ago?

XVilka:
Not much people saw the code I think.

PR #11835, opened 2018-10-16
Fix #11834 clock_nanosleep/FreeBSD.

### LINKED ISSUE #11834: HAS_CLOCK_NANOSLEEP crash the build (radare2 v 3.0 in #FreeBSD 10.4 RELEASED)
### Work environment

| OS/arch/bits (mandatory)                             | FREEBSD 10.4-RELEASED - AMD 64
| File format of the file you reverse (mandatory)      |  -
| Architecture/bits of the file (mandatory)            |  -
| r2 -v full output, **not truncated** (mandatory)     | (compile error)

### Git commit version
```asm
Tue Oct 16 18:52:37 JST 2018
heads/master-0-g647a6fa43
HEAD is now at 647a6fa43 Clarify BIOS section mapping in comments
```
### Expected behavior
(compile success)

### Actual behavior
(compile failed)

- use FreeBSD 10.4 (latest) on AMD 64 and compile r2 via git

### Problem source:
- This commit is the reason: https://github.com/radare/radare2/commit/a018c02cb703b8c11524e9fe2c3fc83f1e0bf5df

### Solution
- basically reverting `libr/include/r_util/r_sys.h` and `libr/util/sys.c` to previous code as per shown in commit https://github.com/radare/radare2/commit/a018c02cb703b8c11524e9fe2c3fc83f1e0bf5df will fix this.
- I pulled request the fix code: https://github.com/radare/radare2/pull/11836

### Additional Logs (report and compile error log w/ details OS / compiler info)
https://github.com/unixfreaxjp/dev/blob/master/r2v3freebsdx11-x64.md?fbclid=IwAR3C5ZDHUqCw6EbLDQfpHGaEn2CKfGuSs8hvWCAkhTiUKlB2xtXqjpv7HFE

Reported by @unixfreaxjp

devnexen:
Appears only from 11.1 release.

### DIFF
--- a/libr/include/r_util/r_sys.h
+++ b/libr/include/r_util/r_sys.h
@@ -9,7 +9,7 @@
 #define R_SYS_DEVNULL "/dev/null"
 #endif
 
-#if __linux__ || (__FreeBSD__ && __FreeBSD_version >= 1101000) || __NetBSD__
+#if __linux__ || (__FreeBSD__ && __FreeBSD_version >= 1101000) || (__NetBSD__ && __NetBSD_Version__ >= 700000000)
 #define HAS_CLOCK_NANOSLEEP 1
 #else
 #define HAS_CLOCK_NANOSLEEP 0
--- a/libr/include/r_util/r_sys.h
+++ b/libr/include/r_util/r_sys.h
@@ -9,7 +9,7 @@
 #define R_SYS_DEVNULL "/dev/null"
 #endif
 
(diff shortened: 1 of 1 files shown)


radare:
I assume we should add the same version check for netbsd and openbsd

devnexen:
OpenBSD does not have this function but NetBSD if we want to be cautious why not, they have since 7.0 (2015).

PR #11836, opened 2018-10-16
Fixing issue #11834 (FreeBSD usleep with HAS_CLOCK_NANOSLEEP macro flagging bug)

unixfreaxjp:
Considering the FreeBSD9 CC (not clang) version that is actually unaffected to this bug  HAS_CLOCK_NANOSLEEP, see report log.

### DIFF
--- a/libr/include/r_util/r_sys.h
+++ b/libr/include/r_util/r_sys.h
@@ -9,12 +9,6 @@
 #define R_SYS_DEVNULL "/dev/null"
 #endif
 
-/* #if __linux__ || __FreeBSD__ || __NetBSD__
-#define HAS_CLOCK_NANOSLEEP 1
-#else
-#define HAS_CLOCK_NANOSLEEP 0
-#endif */ 
-
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/libr/include/r_util/r_sys.h
+++ b/libr/include/r_util/r_sys.h
@@ -9,11 +9,11 @@
 #define R_SYS_DEVNULL "/dev/null"
 #endif
 
-#if __linux__ || __FreeBSD__ || __NetBSD__
+/* #if __linux__ || __FreeBSD__ || __NetBSD__
 #define HAS_CLOCK_NANOSLEEP 1
 #else
 #define HAS_CLOCK_NANOSLEEP 0
-#endif
+#endif */ 
 
 #ifdef __cplusplus
 extern "C" {
--- a/libr/util/sys.c
+++ b/libr/util/sys.c
@@ -249,7 +249,7 @@ R_API void r_sys_backtrace(void) {
 }
 
 R_API int r_sys_sleep(int secs) {
-#if HAS_CLOCK_NANOSLEEP
+#if __linux__
 	struct timespec rqtp;
 	rqtp.tv_sec = secs;
 	rqtp.tv_nsec = 0;
@@ -263,7 +263,7 @@ R_API int r_sys_sleep(int secs) {
 }
 
 R_API int r_sys_usleep(int usecs) {
-#if HAS_CLOCK_NANOSLEEP
+#if __linux__
 	struct timespec rqtp;
 	rqtp.tv_sec = usecs / 1000000;
 	rqtp.tv_nsec = (usecs - (rqtp.tv_sec * 1000000)) * 1000;


unixfreaxjp:
I didn't see your PR until now, I saw that we have different approach.
10.4-RELEASE (currently stable and in production) doesn't have it and many of legend versions with self compiled kernels are running too, thus BSD embedded too with CC new compiler too affected. I strongly suggest avoiding  HAS_CLOCK_NANOSLEEP for avoiding build error, with this pull request (tested).
```asm
$ r2 -v
radare2 3.0.0 19916 @ freebsd-x86-64 git.2.9.0-385-g0e62aee60
commit: 0e62aee6018dfa0f1c8ee8664c4ad1fd54c3ed11 build: 2018-10-16__21:51:30
```

unixfreaxjp:
As per suggested, I add comments in here:

radare2 is well known for stability & compatibility in UNIX environment for long. This is why many UNIX users, like myself, started on using radare.
`clock_nanosleep` was introduced to FreeBSD since version 11.1, while many users are mostly stick with 10 due to differences between 10 and 11 are big, and a lot of legacy hardened products, NAS products, and embed products are using previous tampered versions. A lot of other BSD users are also seek the codes in the FreeBSD `/usr/ports` for packages and that includes compilers obviously. So out there there are so many various of running boxes that under version `11.1` and not necessary have the specific minor versions to aim.

Further, `clock_nanosleep` is good, but we doesn't have any problem with current usleep() using previous logic until end of 2.9. Suddenly I tested BSD when v3 released and boom! several errors and crash appears, PS* the test is still in progress.. some Issues to come from me will be..

We had no big problem with or without `clock_nanosleep` before, I am asking, why we should ruin the compatibility and stability as our existing quality with this additional? I just run the test and the first bug I hate and I must face now is if I can not compile r2, and this what had actually happened.

We had similar issues with several previous bugs I found in FreeBSD too (it is all in my Issue list), which in the end we decided to bring up compatibility and stability as stronger consideration. And here we are, here we go again. Not that I am against technology, but I am against the degradation of quality of radare2 in BSD platforms that will be impacted. The current usleep function is good enough for previous version, no fatal urgency or error or any damage that make us change it right away...

If the `clock_nanosleep` was coded in a way that won't disturb my tests I am okay with it, the case is build error, once you can't compile r2 the matter is becoming URGENT. I think you need to test more thoroughly too. If I can find the bug in my first run test and you didn't see that coming, means the current check scheme still need to be improved, sorry for put this out loud.

More, if we don't be stern about these stability and compatibility in testing or our beloved product, coders will do more and more innovation without much caring people who were using the r2 in previous boxes or environments since naturally people who codes will check only for environments that they can reach or used. What will happen is we may ending up to workaround rabbit hole or more works in adjusting due to minor version used as flag may be changed, we don't know that.. since we're not BSD gods. We need to test, learn about the alive boxes out there, seek the roadmap of the current running boxes too, to implement something different than before. Shortly, radare2 is much way older than FreeBSD 11.1 itself, do we have to abandon all users who got affected?

I am willing to aside my happiness for finally doing a PR (again) on big "THAT I FOUND", if that is for the r2 good. Gladly will do that. So it is up to the leader and the rest of r2dev to decide. 
BSD is my life. Many of us are out there too.

Enough ranting, so, as suggestion, if you that "eager" to code `clock_nanosleep` in the `usleep`, let the legacy FREEBSD system REALLY died first, until maybe wheｎ FreeBSD v13 is released, or similar, FreeBSD12 will keep some of version 10 alive a lot too I guess, I guess the current legacy will be dead by then. 
Right now? The timing is incorrect, this is why I coded my PullRequest for this matter.

XVilka:
Just check the FreeBSD version in macros

XVilka:
See https://www.freebsd.org/doc/en/books/porters-handbook/porting-versions.html

radare:
The reason for switching to nanosleep instead of usleep is because it's thread safe and handles signals properly (according to the documentation), so it's not just a matter of rolling back to usleep because it works everywhere. We must use nanosleep where it is available. So it is better to just check for the OS and version.

Breaking support for non-main-platforms is expected as long as we dont have any farm running those operating systems. This PR was merged in 9th October, this is one week before the release, and usually people doesn't test r2 until I do a release. This happened ALWAYS since i have started the radare project. So i don't expect this to change, even if I do alpha/beta releases people keeps ignoring them until the final release is out. So it's impossible to get good feedback from all targets and distros before a release.

Specifying a 1-2 week freeze time before each release can be good to avoid such issues, but it's not a perfect solution either.

Btw I was already planning to release 3.0.1 tomorrow with few bug fixes that have been spotted by other people. (Like compilation on s930x architecture). So this fix will be included there.

Can you confirm the fix in https://github.com/radare/radare2/pull/11835 works for you?

radare:
We can also move this check to the configure step and just verify if the symbol exists in libc. This will be safer than checking for operating system macros.. and maybe add a commandline flag to disable the use of this, but this is probably unnecessary because it is desired to use it by default.

unixfreaxjp:
> Can you confirm the fix in #11835 works for you?

Still at work now, I will do that after.

radare:
I have merged the version check fix, please confirm that it's working in your systems. I'll do the check at configure time in case it doesn't.

Thanks!

unixfreaxjp:
Checked and the problem is solved:
```asm
$ r2 -v
radare2 3.0.0 19737 @ freebsd-x86-64 git.2.9.0-394-gd34b5a57e
commit: d34b5a57e4db671e8a696f7c68f36f50c133d9a2 build: 2018-10-18__07:21:20


$ r2 -v
radare2 3.0.0 19782 @ freebsd-x86-32 git.2.9.0-394-gd34b5a5
commit: d34b5a57e4db671e8a696f7c68f36f50c133d9a2 build: 2018-10-18__16:29:23

```
Thank you very much for the cooperation.

radare:
👍

PR #11840, opened 2018-10-17
Fix cyclomatic complexity computation when jump tables are involved ##anal

radare:
* Add asserts and checks to ensure the graph is correct before analysis
* Use correct E-N+2P formulae
* Run afbe right after af to get the jmptbl info loaded asap

### DIFF
--- a/libr/anal/bb.c
+++ b/libr/anal/bb.c
@@ -1,4 +1,4 @@
-/* radare - LGPL - Copyright 2010-2016 - pancake, nibble */
+/* radare - LGPL - Copyright 2010-2018 - pancake, nibble */
 
 #include <r_anal.h>
 #include <r_util.h>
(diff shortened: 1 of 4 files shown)


radare:
As for reference, this is the binary. and the command to test:

```
$ r2 -qs 0x140002510 -c 'af;afi~complex' rabin2.exe
```

[rabin2.zip](https://github.com/radare/radare2/files/2488246/rabin2.zip)

funnily this function is wrongly analyzed by default so we have to do this:

```
$ r2 -e anal.endsize=false -qs 0x140002510 -c 'af;afi~complex' rabin2.exe
```

we should remove this anal.endsize variable, or change it to false by default. and here wer have a test to verify that the analysis good with endsize=0, but we have other bins in the testsuite that fail to be analyzed without changing this variable

PR #11841, opened 2018-10-17
Fix #11838 - Handle op.val when finding xrefs ##anal

### LINKED ISSUE #11838: Axt missing xrefs
### Work environment

| OS/arch/bits (mandatory)                             | OSX 64 bits
| File format of the file you reverse (mandatory)      | ELF32
| Architecture/bits of the file (mandatory)            | x86 32 bits
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.0.0 19943 @ darwin-x86-64 git.2.9.0-380-g218cdd7ba commit: 218cdd7babd73866f69052e28b1f333c1b6c2dd7 build: 2018-10-16__11:41:19

### Expected behavior

```
axt @ str.the_ripper
main 0x08048321 [DATA] mov dword [ebp - 0x10], str.the_ripper
```

### Actual behavior

`axt @ str.the_ripper`

[ch2(1).zip](https://github.com/radare/radare2/files/2486529/ch2.1.zip)

Not caused by the ht optimization, per @ret2libc

radare:


astuder:
PS: Or to fix the issue at hand, use asm.var.submin. That way axt will be in sync with the disassembly.

radare:
On intel immediates can be memory addresses too. Theres a check that verifies if the value is pointing to something meaningful before going further so i dont see much problem in here. This wasnt catched because ms compiler uses mov instead of lea and things like this

We can change the analop for the mov.. but yhen we will break other usecases.

radare:
I agree that for firmwares or stuff that is mapped at address 0 can rwsult on false positives. But seems like we have 0 tests for this. Do you have any reproducer? Becayse travis is green

radare:
Yeah. Thats another topic :)

I would merge it like this. And add tests. Then we can play with finding false positives and this min thing. But we should probably add a mode to be more strict on the imm/addr analysis which will be useful for 8bit archs.

astuder:
The 8051 interrupt vector table before the PR:
before-pr-11841

After the PR:
after-pr-11841

radare:
Can you add a test in r2r?

radare:
I’ll get the bin from there and fix the pr before merging

astuder:
I can't share this binary, but I will see if I can add an 8051 hello world to r2r.

astuder:
Created a regression test using AVR and Arduino.
https://github.com/radare/radare2-regressions/pull/1524

XVilka:
Agree with @astuder on:
1. `Using asm.var.submin`
2. `substitution in disasm should rely on anal results (i.e. xrefs) instead of making its own guesses`

radare:
should be fixed now we need tests to confirm

astuder:
works with my project

radare:
Awesome. Thanks for testing. Do unhave time to write a test for the bin it was not finding the ref and update your pr? So i can merge everything at once?

astuder:
I think the test for Majins scenario is already in r2r:
https://github.com/radare/radare2-regressions/commit/638dfc51670cf0e260f3bbe18ae35d5f735be868

PR #11843, opened 2018-10-17
Allow multiple packages as arguments for r2pm -i,-c,-ci ##tools

radare:
which hashtag would be good for this? ##tools?


ret2libc:
Yeah `##tools` seems good enough.

PR #11846, opened 2018-10-18
Fix UBSAN unaligned struct issue in the tcc parser

radare:


radare:
Imho the PDB parser should be reimplemented as a bin plugin, and the idp command should be removed because everything will be managed the by rbin api, like its done for dwarf.

PR #11853, opened 2018-10-18
Fix ptrace musl compability

jnbr:
musl libc doesn't use `enum __ptrace_request`. This causes two problems when trying to compile for it:

`libr/include/r_io.h` defaults to `__enum ptrace_request`, inverting the logic and only using the enum when building for glibc fixes this.

`shlr/ptrace_wrap` uses `enum __ptrace_request`. I see no option to disable `ptrace_wrap` at configure time without patching the meson or configure files. Using the same `typedef` as in `libr/include/r_io.h` fixes `ptrace_wrap` for musl.

### DIFF
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -26,12 +26,12 @@
 #include <sys/ptrace.h>
 #endif
 
-#if defined(__APPLE__) || __sun || __FreeBSD__ || __OpenBSD__
-typedef int r_ptrace_request_t;
-typedef int r_ptrace_data_t;
-#else
+#if defined(__GLIBC__) && defined(__linux__)
 typedef enum __ptrace_request r_ptrace_request_t;
 typedef void * r_ptrace_data_t;
+#else
+typedef int r_ptrace_request_t;
+typedef int r_ptrace_data_t;
 #endif
 #endif
 


radare:
shouldnt this fix go into the ptrace-wrap repo?

shlr/ptrace-wrap/include/ptrace_wrap.h
```diff
@@ -22,6 +22,12 @@
 #include <semaphore.h>
 #include <sys/ptrace.h>
 
+#ifndef __GLIBC__
```
radare: i prefer positive conditionals #ifdef ...

libr/include/r_io.h
```diff
@@ -26,7 +26,7 @@
 #include <sys/ptrace.h>
 #endif
 
-#if defined(__APPLE__) || __sun || __FreeBSD__ || __OpenBSD__
+#ifndef __GLIBC__
```
radare: what about gnu/kfreebsd ?

shlr/ptrace-wrap/include/ptrace_wrap.h
jnbr: changed it

libr/include/r_io.h
jnbr: covered now by only using the enum on glibc on linux

thestr4ng3r:
requires https://github.com/radare/radare2/pull/11862

PR #11854, opened 2018-10-18
Fix VV esil function comment no-newline

kazarmy:
Fix for the following problem:

31c95e8a-c4e3-4304-a003-16e0f8087ad4

specifically [line 4331](https://github.com/radare/radare2/compare/master...kazarmy:VV-esil-func-cmt-no_nl-fix?expand=1#diff-7b45b46dd8a84a924b91f34c2b1e6309R4331) should move `blx sym.imp.close` to a new line.


radare:
screenshot 2018-10-18 at 17 21 07

PR #11863, opened 2018-10-18
libr/util/list.c: remove assert in r_list_length

### LINKED ISSUE #11844: rlist asserts in i commands
screenshot 2018-10-18 at 02 44 38

davidpolverari:
r_list_length was not returning 0 if RList* was null, because assert
prevented it.

Closes #11844

### DIFF
--- a/libr/util/list.c
+++ b/libr/util/list.c
@@ -52,8 +52,7 @@ R_API void r_list_init(RList *list) {
 }
 
 R_API int r_list_length(const RList *list) {
-	r_return_val_if_fail (list, 0);
-	return list->length;
+	return list? list->length: 0;
 }
 
 /* remove all elements of a list */


radare:
asserts are there for a reason, we must fix the code to avod those asserts to happen

davidpolverari:
so, for example [in this line](https://github.com/radare/radare2/blob/136db63c05cfae30d5ab145e8a697c9598f561bc/libr/core/cmd_info.c#L512) `obj->libs` could never be `NULL`? I thought that [r_list_length](https://github.com/radare/radare2/blob/7677ef924668a7e7742eb77c9086a4149696e577/libr/util/list.c#L54) should deal with a null RList*, as for example in the case stated in issue #11844, r2 with malloc://1024 will have a obj->libs==NULL, right?

ret2libc:
> Should in this case a NULL pointer never be passed to r_list_length?

That's right. It shouldn't. `r_list_length` should return the length of a list and NULL is not a list, so you are calling it wrongly. In particular in https://github.com/radare/radare2/issues/11844, we should check for obj->libs, etc. and provide a meaningful message before calling `r_list_length`. For example, something like `The opened file does not provide binary information`.

Example (very simple but just to get the idea) for https://github.com/radare/radare2/blob/136db63c05cfae30d5ab145e8a697c9598f561bc/libr/core/cmd_info.c#L510:
```
case 'l': { // "il"
			RBinObject *obj = r_bin_cur_object (core->bin);
                        if (!obj || !obj->libs) {
                                    eprintf("The opened file does not provide binary or libs info\n");
                                    break;
                        }
			RBININFO ("libs", R_CORE_BIN_ACC_LIBS, NULL, r_list_length (obj->libs));
			break;
}
```

This shows, for example, that we are dealing with `i` commands in the wrong way. Why would you want to get empty results from `i` commands when there's no binary? If you call `il` and you just get no output, does that mean the binary has no libraries or that you have opened a file with no binary/lib information? These asserts help us catch those cases.

radare:
i always prefer to have positive conditionals. so do if (obj && obj->libs) { … } else eprintf

PR #11866, opened 2018-10-19
libr/asm: Use memcmp instead of strcmp in r_asm_add

lzutao:


lzutao:
Ooops, closed becaused using C99 variable declaration.

PR #11867, opened 2018-10-19
libr/asm: Use memcmp instead of strcmp in r_asm_add

lzutao:


lzutao:
Build failed ! Checking it !

radare:
Thats wrong. What are you trying to do here?

lzutao:
Sorry, I tried to reuse the `r_asm_is_valid` function in ` r_asm_add`, and use `memcmp` instead of `strcmp` but I forgot to +1 in result of `strlen`.

radare:
there’s no point in using memcmp to compare strings, strcmp should be used

lzutao:
There is a TODO in `asm.c` say that "use memcmp instead of strcmp". 

Btw, what about using `strncmp` instead?

Maijin:
image

Maijin:
Quite old commit though

Maijin:
@lzutao if you're looking for contributions use the issue tracker ;)

PR #11868, opened 2018-10-19
RBin refactoring ##refactor

ret2libc:


libr/bin/bfile.c
```diff
@@ -324,13 +324,13 @@ R_API RBinFile *r_bin_file_new(RBin *bin, const char *file, const ut8 *bytes, ut
 	}
 	int res = r_bin_file_set_bytes (binfile, bytes, sz, steal_ptr);
 	if (!res && steal_ptr) { // we own the ptr, free on error
-		free((void*) bytes);
+		free ((void *)bytes);
 	}
 	binfile->rbin = bin;
-	binfile->file = file? strdup (file): NULL;
+	binfile->file = file ? strdup (file) : NULL;
```
radare: nope. spacing goes after the ?, not before, isnt?

libr/bin/obj.c
```diff
@@ -264,6 +260,7 @@ R_API int r_bin_object_set_items(RBinFile *binfile, RBinObject *o) {
 }
 
 R_API RBinObject *r_bin_object_get_cur(RBin *bin) {
+	r_return_val_if_fail (bin, NULL);
```
radare: we should kill the concept of "current binobj"

libr/core/cfile.c
```diff
@@ -314,21 +313,20 @@ static int r_core_file_do_load_for_debug(RCore *r, ut64 baseaddr, const char *fi
...
+	RBinOptions bo = {
+		.offset = 0LL,
+		.baseaddr = baseaddr,
+		.rawstr = false,
+		.xtr_idx = xtr_idx,
+		.iofd = fd,
+		.loadaddr = UT64_MAX,
+	};
 
-	bo = r_bin_options_new (0LL, baseaddr, false);
```
radare: should we kill the options_new then? its usually handy for bindings to have methods to construct structs rather than having to build them by hand., from C its fine like this imho. but then we should probably specify that the binoptions is const in the rbinopen signature

libr/io/io.c
```diff
@@ -529,10 +529,9 @@ R_API bool r_io_set_write_mask(RIO* io, const ut8* mask, int len) {
 	return true;
 }
 
-R_API int r_io_bind(RIO* io, RIOBind* bnd) {
-	if (!io || !bnd) {
-		return false;
-	}
+R_API int r_io_bind(RIO *io, RIOBind *bnd) {
```
radare: this is bool

libr/bin/bfile.c
ret2libc: I think we have both cases in our codebase... I was always used to ` ? ` and clang-format supports that.

libr/bin/obj.c
ret2libc: that's a good point i wanted to ask you indeed. what's the point in having multiple rbinobj in a rbinfile? I though as well we could remove this thing that just complicates stuff.

libr/core/cfile.c
ret2libc: i removed the options_new api, yes. and yes, i will add `const`.

ret2libc:
> there's no refactoring at all in this PR. its just code cleanup

well, 1) it's still WIP 2) this "code cleanup" according to wikipedia is refactoring :P

radare:
the idea behind this was to have common shared information between different loadings, so for example

You have one rbinfile which references a file (fd), and then the rbinobjects which reference a view of that file. the purpose of this was to have support for fatbins, this is 1 file with N binobjects. or the same goes for zip files, you have one zip file.

But i agree that this is just an unnecesssary abstraction that is making the understanding of the api more difficult.

it reminds me a bit to the rcorefile.. so we should probably kill it yes.

libr/bin/obj.c
```diff
@@ -371,13 +374,13 @@ R_API int r_bin_object_delete(RBin *bin, ut32 binfile_id, ut32 binobj_id) {
 }
 
 R_API void r_bin_object_set_baddr(RBinObject *o, ut64 baddr) {
-	if (!o || baddr == UT64_MAX) {
-		return;
-	}
+	r_return_if_fail (o && baddr != UT64_MAX);
 	o->baddr_shift = baddr - o->baddr;
```
ret2libc: XXX: this is not setting `o->baddr` anymore. It should, given the name of the API. All changes to baddr should come through this API. Also, `baddr` is many time initialized with `0xfffffffffffffffffff`, but why? A baddr is always defined, even it is not set explicitly (where, I think, it is assumed to be zero. So dealing with 0xffffff..fff just makes things harder)

radare:
no, pic bins have no baddr, object files have no baddr either. ffff means that r2 should specify one of its own (by rbinldr or specified by the user with -B)

ret2libc:
> no, pic bins have no baddr, object files have no baddr either

that's just an internal detail i think, since even when you load those things, baddr gets set to `0` after a bunch of set and re-set. You can test it by loading the PIC binary (or the library) and check `i~baddr`.

radare:
baddr 0 is a valid base address. -1 have a different meaning

libr/include/r_bin.h
```diff
@@ -320,7 +320,6 @@ typedef struct r_bin_t {
 	ut64 filter_rules;
 	bool demanglercmd;
 	bool verbose;
-	bool io_owned;
```
ret2libc: @radare i removed this, it wasn't ever used. It was set in r_bin_load but that path was never reached actually...

libr/bin/bin.c
```diff
@@ -226,48 +235,27 @@ R_API void r_bin_string_free(void *_str) {
...
-	}
-	// ALIAS?	return r_bin_load_as (bin, file, baseaddr, loadaddr,
-	// xtr_idx, fd, rawstr, 0, file);
+	r_return_val_if_fail (bin, false);
+
 	RIOBind *iob = &(bin->iob);
-	if (!iob) {
-		return false;
-	}
-	if (!iob->io) {
```
ret2libc: i removed this

radare:
Because its used when you create rbin outside rcore. Therefor rbin instantiates its own io and owns it to avoid leak/dblfree

ret2libc:
> Because its used when you create rbin outside rcore. Therefor rbin instantiates its own io and owns it to avoid leak/dblfree

Yes, but in radare2 code it is never used outside rcore. And even if it was, the simplest solution is to just make sure that the caller allocates the RIO itself. This would make the code much simpler. Anyway, that's not needed because it seems all calls to r_bin_load do use a RBin coming from RCore.

radare:
Let me check the projects that are using this...

libr/bin/bin.c
```diff
@@ -32,17 +32,9 @@ static RBinPlugin *bin_static_plugins[] = { R_BIN_STATIC_PLUGINS, NULL };
...
-	// (MIN_LOAD_SIZE)
-	// if it fails, the whole file is loaded.
-	const ut64 MAX_LOAD_SIZE = 0;  // 0xfffff; //128 * (1 << 10 << 10);
-	int res = r_bin_load_io2 (bin, fd, baseaddr,
-		loadaddr, xtr_idx, offset, name, MAX_LOAD_SIZE);
-	return res? res: r_bin_load_io2 (bin, fd, baseaddr,
-			loadaddr, xtr_idx, offset, name, UT64_MAX);
+R_API bool r_bin_load_io(RBin *bin, int fd, ut64 baseaddr, ut64 loadaddr, int xtr_idx, ut64 offset, const char *name) {
+	r_return_val_if_fail (bin, false);
+	return r_bin_load_io2 (bin, fd, baseaddr, loadaddr, xtr_idx, offset, name, 0);
```
ret2libc: the second call to r_bin_load_io2 has been removed because of what's explained in https://github.com/radare/radare2/pull/11868/commits/5a151c0782e574328cc7ab7f4771415a847e9d87

ret2libc:
Ok I think this PR is already big enough as it is, so I'm going to stop here for the moment... Let's see what Travis says. It's mainly about cleaning code, removing unused functions (e.g. r_bin_load_as) and removing some unused fields in RBinOptions(e.g. name) and in RBin (e.g. is_owned)

libr/bin/open.c
```diff
@@ -2,33 +2,17 @@
...
 	int iofd = -1, rawstr = 0, xtr_idx = 0;
-	if (bo) {
-		baddr = bo->baseaddr;
-		laddr = bo->loadaddr;
-		xtr_idx = bo->xtr_idx;
-		iofd = bo->iofd;
-		rawstr = bo->rawstr;
-	}
+
+	r_return_val_if_fail (bo, -1);
```
radare: add asserts for bin and filename

radare: those cant be null either

radare:
lgtm, agree, this is_owned was deprecated already but not removed. it's fine to kill it

radare:
The isowned thing seems unused. Projects i knew used it switched to r2pipe, so its fine to merge imho

ret2libc:
It’s red just because of the hvaddr thing

radare:
We can merge it then because ive reverted this red thing

ret2libc:
We can merge if you want, but I'd prefer after 3.0.1. So we have some time to test.

Issue #3593, opened 2015-10-26
Reduce memory usage


radare:
This is a work-in-progress task that aims to track the refactoring task to reduce memory usage without affecting performance.

There are many places where there memory is abused. for example in `r_bin.h`:

<img width="389" alt="screen shot 2015-10-26 at 03 48 04" src="https://cloud.githubusercontent.com/assets/917142/10720178/88eb4cc0-7b94-11e5-8490-2cb59378a9a0.png">

Same situations can be found in `r_asm.h` `r_anal.h` and others `r_core.h` where many variables can be defined as bool instead of int, and reduce buffer sizes or just use the heap.

Keeping, or enhancing speed is important. But also, we should find and fix memory leaks.

```
$ valgrind --dump-instr=yes --trace-jump=yes --tool=callgrind r2 /bin/ls
...
$ kcachegrind callgrind*
```

radare:
Maybe a good idea to follow the approach in RStrBuf or maybe just go for the appraoch of a string pool and just store pointers with a reference to this single allocation place. the reason for not doing that was to try to avoid the use of the heap. but i guess a single allocation for all the strings on each struct should be fine for this case

radare:
See also r_str_const

radare:
This is fixed

PR #11872, opened 2018-10-19
Refactor RBinSection

F3real:
Part of #3593


radare:
ping, make travis happy, rebase this pr and review the XX

libr/bin/format/mdmp/mdmp_pe.c
```diff
@@ -159,7 +159,10 @@ RList *PE_(r_bin_mdmp_pe_get_sections)(struct PE_(r_bin_mdmp_pe_bin) *pe_bin) {
 			break;
 		}
 		if (sections[i].name[0]) {
-			strncpy (ptr->name, (char*)sections[i].name, R_BIN_SIZEOF_STRINGS);
+			ptr->name = strdup ((char*)sections[i].name);
+		}
```
radare: else goes in the same line as }

libr/bin/format/mdmp/mdmp_pe.c
```diff
@@ -159,7 +159,10 @@ RList *PE_(r_bin_mdmp_pe_get_sections)(struct PE_(r_bin_mdmp_pe_bin) *pe_bin) {
...
+		else {
+			ptr->name = strdup ("");
```
radare: probabl name = UNLL is fine

libr/bin/p/bin_te.c
```diff
@@ -106,14 +106,7 @@ static RList *sections(RBinFile *bf) {
...
 		}
-		if (sections[i].name[sizeof (sections[i].name) - 1]) {
-			memcpy (ptr->name, sections[i].name,
-				sizeof (sections[i].name));
-			ptr->name[sizeof (sections[i].name)] = 0;
-		} else {
-			strncpy (ptr->name, (char *) sections[i].name,
-				R_BIN_SIZEOF_STRINGS);
-		}
+		ptr->name = strdup ((char*)sections[i].name);
```
ret2libc: mmm probably a strndup would be better here. It seems like sections[i].name may not be null terminated. and in that case strdup would cause bad behaviour.

libr/bin/p/bin_wasm.c
```diff
@@ -104,7 +104,7 @@ static RList *sections(RBinFile *bf) {
 			r_list_free (ret);
 			return NULL;
 		}
-		strncpy (ptr->name, (char*)sec->name, R_BIN_SIZEOF_STRINGS);
+		ptr->name = strdup ((char*)sec->name);
```
ret2libc: strndup

libr/bin/format/mdmp/mdmp_pe.c
```diff
@@ -159,7 +159,9 @@ RList *PE_(r_bin_mdmp_pe_get_sections)(struct PE_(r_bin_mdmp_pe_bin) *pe_bin) {
 			break;
 		}
 		if (sections[i].name[0]) {
-			strncpy (ptr->name, (char*)sections[i].name, R_BIN_SIZEOF_STRINGS);
+			ptr->name = strdup ((char*)sections[i].name);
```
ret2libc: strndup here too?

libr/bin/format/omf/omf.c
```diff
@@ -751,9 +751,9 @@ int r_bin_omf_send_sections(RList *list, OMF_segment *section, r_bin_omf_obj *ob
 
 		// if index == 0, it's mean there is no name
 		if (section->name_idx && section->name_idx - 1 < obj->nb_name) {
-			snprintf (new->name, R_BIN_SIZEOF_STRINGS, "%s_%d", obj->names[section->name_idx - 1], ct_name++);
+			new->name = r_str_newf ("%s_%d", obj->names[section->name_idx - 1], ct_name++);
```
ret2libc: @radare is there a `r_str_nnewf` that would limit the maximum size of the string (e.g. similar to snprintf)? I'm not sure we can assume obj->names[..] is null terminated. can we?

radare:
merging my only concern is about using "NULL" when no section name is there, but as long as its geren, and the refactor looks legit, we can review and discuss that later

radare:
thanks

radare:
I think this situation only happens in PE. Would be good to have a test. But i think we null terminate them in the interna representation too so it may be fine i think

PR #11876, opened 2018-10-20
Fix #11869 by using r_bin_elf_p2v_new

kazarmy:
Oops.


radare:
Why did u merged this red pr? Also it was wrong

ret2libc:
It is green

PR #11885, opened 2018-10-21
core/cmd_info.c: check for NULL before calling r_list_length()

### LINKED ISSUE #11844: rlist asserts in i commands

davidpolverari:
There are other cases that are harder to fix. For example, with `iI`, trying to check RBinObj->bin_obj is null works for most of RBin plugins, but some won't set bin_obj and still load binary information, making it harder to deal with it.

Closes #11844


davidpolverari:
Several tests depend on the the text returned by the current behavior, so any textual changes will break them.

PR #11887, opened 2018-10-21
ie: set hvaddr of elf entry0 from r2 ehdr segment ##bin

kazarmy:
Continuation of https://github.com/radare/radare2-regressions/pull/1527#issuecomment-431599115.

### DIFF
--- a/libr/bin/p/bin_elf.inc
+++ b/libr/bin/p/bin_elf.inc
@@ -397,7 +397,20 @@ static RList* entries(RBinFile *bf) {
 	ptr->paddr = Elf_(r_bin_elf_get_entry_offset) (obj);
 	ptr->vaddr = Elf_(r_bin_elf_p2v) (obj, ptr->paddr);
 	ptr->hpaddr = 0x18;  // e_entry offset in ELF header
-	ptr->hvaddr = Elf_(r_bin_elf_p2v) (obj, ptr->hpaddr);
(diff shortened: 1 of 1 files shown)


libr/bin/p/bin_elf.inc
```diff
@@ -397,7 +397,20 @@ static RList* entries(RBinFile *bf) {
 	ptr->paddr = Elf_(r_bin_elf_get_entry_offset) (obj);
 	ptr->vaddr = Elf_(r_bin_elf_p2v) (obj, ptr->paddr);
 	ptr->hpaddr = 0x18;  // e_entry offset in ELF header
-	ptr->hvaddr = Elf_(r_bin_elf_p2v) (obj, ptr->hpaddr);
+	ptr->hvaddr = UT64_MAX;
+	if (bf->o->sections) {
+		RListIter *iter;
+		RBinSection *section;
+		r_list_foreach_prev (bf->o->sections, iter, section) {
```
radare: why prev?

kazarmy: The ehdr section is the last section added:
https://github.com/radare/radare2/blob/37b4a61718532c7b2a0afbb6d9e6aec4ec660ee7/libr/bin/p/bin_elf.inc#L287
but there's no guarantee it'll stay that way.

PR #11888, opened 2018-10-21
other RBin cleanup

ret2libc:


libr/bin/bfile.c
```diff
@@ -723,19 +723,11 @@ R_API int r_bin_file_set_cur_by_name(RBin *bin, const char *name) {
...
-		binfile->o->plugin = plugin;
-		return true;
-	}
-	return false;
+	return binfile ? binfile->o : NULL;
 }
 
 R_API int r_bin_file_deref_by_bind(RBinBind *binb) {
-	RBin *bin = binb? binb->bin: NULL;
+	RBin *bin = binb ? binb->bin : NULL;
```
radare: dat indent

libr/bin/open.c
```diff
@@ -20,17 +20,3 @@ R_API int r_bin_open(RBin *bin, const char *filename, RBinOptions *bo) {
 	}
 	return -1;
 }
-
-R_API RBinFile *r_bin_get_file(RBin *bin, int bd) {
-	return r_id_storage_take (bin->ids, bd);
-}
-
-R_API bool r_bin_close(RBin *bin, int bd) {
```
radare: uhm i think we dont want to remove this API because the idea was to make all rbinobj apis use the idstorage instead of returning pointers like condret did for siol.

before removing this i would like to discuss about the api design

libr/core/cmd_open.c
```diff
@@ -152,7 +152,7 @@ static const char *help_msg_oonn[] = {
 	NULL
 };
 
-static inline ut32 find_binfile_id_by_fd (RBin *bin, ut32 fd) {
+static inline ut32 find_binfile_id_by_fd(RBin *bin, ut32 fd) {
 	RListIter *it;
 	RBinFile *bf;
 	r_list_foreach (bin->binfiles, it, bf) {
```
radare: idstorage is suposed to replace this

libr/include/r_bin.h
```diff
@@ -14,6 +14,10 @@ extern "C" {
 
 R_LIB_VERSION_HEADER (r_bin);
 
+#define MODE_PRINT 0x000
```
radare: this is used in core mainly, but its a pretty generic "types", so i would kill it or make it into r_util and reuse it everywhere, but basically we handle 'j' for json, 'r' for r2 (or '*', 0 for normal, 'q' for quiet (simple)

ret2libc: Yeah probably moving to r_util is the best thing... I see they are dupped in r_core.h (e.g. `R_CORE_BIN_PRINT`, etc.)

ret2libc: Yeah probably moving to r_util is the best thing... I see they are dupped in r_core.h (e.g. `R_CORE_BIN_PRINT`, etc.)

libr/bin/open.c
radare: if you keep r_bin_open dont kill r_bin_close

libr/include/r_bin.h
radare: there's echo? pls move this and the ones from core.. i would probably put that in r_types or something even if its not a basic enum, it is for the whole r2, and all the printing modes should be implemented in core. but we are far from this point too :P

libr/bin/open.c
ret2libc: ok

ret2libc:
This is mainly cleaning and moving stuff around.. like renaming R_CORE_BIN_PRINT/SET/JSON/etc. to R_MODE_PRINT/SET/JSON in r_types.h . Also removed some unused code and simplified some bits in binr (like removing the redundant r_load_io2).

radare:
Please use proper commit message (capitalized, concise and descriptibe and use the ## )

PR #11894, opened 2018-10-21
check for NULL on obj member before calling r_list_length

minhnxd:
obj members pointer are default to NULL, so the member needed to be checked before calling r_list_length().
#11844


ret2libc:
@radare did you read my comment?

radare:
no

PR #11906, opened 2018-10-23
bin/bin.c: check if o->libs is NULL before calling r_list_length

### LINKED ISSUE #11881: r_list_length assertion failed
### Work environment

| OS/arch/bits (mandatory)                             | Windows/x86/64
| File format of the file you reverse (mandatory)      | none
| Architecture/bits of the file (mandatory)            | none
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.0.0 1 @ windows-x86-64 git. commit: 7d32c05f9a4134e78b5b7d5737ff9bcbcdfca54d build: Sat 10/20/2018__10:35:45.06

### Expected behavior
no warnings/errors
### Actual behavior
`WARNING: r_list_length: assertion 'list' failed (line 55)`
run `r2 -c ia -`

davidpolverari:
The "iI" (and consequently "ia") commands were failing on assert because
of that.

Fixes #11881

### DIFF
--- a/libr/bin/bin.c
+++ b/libr/bin/bin.c
@@ -945,7 +945,7 @@ R_API int r_bin_is_stripped(RBin *bin) {
 R_API int r_bin_is_static(RBin *bin) {
 	r_return_val_if_fail (bin, false);
 	RBinObject *o = r_bin_cur_object (bin);
-	if (o && r_list_length (o->libs) > 0) {
+	if (o && o->libs && r_list_length (o->libs) > 0) {
 		return R_BIN_DBG_STATIC & o->info->dbg_info;
 	}
 	return true;


ret2libc:
As written in a very similar thread: 'I actually think we should print some error message when trying to execute `i` commands but `obj` or `obj->symbols/imports/etc.` are NULL, because that means there were some errors and that's different from not having symbols at all. As I said in another thread, 0 imports is not the same as "there's not even a binary" and is not the same as "i couldn't get imports because malloc failed/header corrupted/whtever else". @radare ?'

ret2libc:
How do you distinguish in this case between a static binary, that has no libraries, and a bin plugin that does not provide those info?

radare:
Why do you need such difference

ret2libc:
Well, usability. User friendliness.

ret2libc:
Sorry, actually the error message would fit in the `i` commands, from where this function r_bin_is_static is called. So it is fine in this case to just check like it's done in the PR, but in the PR you merged yesteday (https://github.com/radare/radare2/pull/11894) it would have made sense to print a message IMHO.

radare:
Imho it will be just messy and fill the logs of many people using r2pipe with useless data when they can just check if output is empty vs empty list. From the user perspective i dont see a reason to have such messages either because if i type il and get nothing i understand that theres no libs to list. Having an error saying there are no libs is redundant. Totally not unix

ret2libc:
I'm not saying you should get an error when there are no libs. I'm saying there should be an error when even "asking for libs" does not have any sense. Like if you open `r2 -`, it doesn't make sense to say that there are no libs, there is not even a binary!

davidpolverari:
I agree with @radare about the errors messages. Anyways, if this is still a moot point, maybe we should open another issue to discuss it, and separate that discussion of bugfixing.

radare:
for me its fine to continue using bin->verbose for those eprintfs, i would have this var unset by default

PR #11913, opened 2018-10-24
Visual panel mode: swap panel position with the one to the right/left

xnand:
Implemented two functions to swap the currently selected panel in visual panel mode with the adjacent ones.

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -132,6 +132,8 @@ static const char *help_msg_panels_window[] = {
 	"|",        "split the current panel vertically",
 	"-",        "split the current panel horizontally",
 	"' '",      "(space) toggle graph / panels",
+	"}",        "swap current panel position with the one on the right",
+	"{",        "swap current panel position with the one to the left",
 	"e",        "change title and command of current panel",
 	"g",        "show graph in the current panel",
 	"hjkl",     "move around (left-down-up-right)",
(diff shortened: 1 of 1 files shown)


Vane11ope:
curnode + 1 isnt necessarily the juxtaposed panel so u need to get the panel more carefully. also code stylings don't match in many places. 

if you want the feature then i can implement though, but im not very sure if the feature is worth implementing really tbh.

xnand:
my bad about code stylings, should have fixed them now locally. about the feature i just felt it was missing to customize the layout a bit more easily, but it's up to you whether to integrate it or not, you sure know more than me about it. will work more on it tomorrow in case.

libr/core/panels.c
```diff
@@ -1424,6 +1428,62 @@ static void resizePanelDown(RPanels *panels) {
...
+	cy = pc.pos.y;
+	cw = pc.pos.w;
+	ch = pc.pos.h;
+	pc.pos.x = panel[prevnode].pos.x;
+	pc.pos.y = panel[prevnode].pos.y;
+	pc.pos.w = panel[prevnode].pos.w;
+	pc.pos.h = panel[prevnode].pos.h;
+	panel[prevnode].pos.x = cx;
+	panel[prevnode].pos.y = cy;
+panel[prevnode].pos.w = cw;
```
radare: wrong indent

radare:
see @Vane11ope comments, i fixed the wrongly indented line

Vane11ope:
actually, to get the panel that is juxtaposed against the one you are on, you need to carefully loop whole panels and nothing to do with indexes for panels. because now you can split the panels vertically with | and horizontally with -, so this PR would not do the thing you wanna do. 

but about you telling me "it should be more flexible to change each panel position", then i completely agree on that. so i will do something about it.

Vane11ope:
i saw your comments in irc and again, to manipulate the panels' positions, you can't do anything with indices of panels, but have to loop panels and check their positions each.

and in my opinion, swapping panels is not effective itself i'm afraid anyway. 

thank you for this and i will add some useful features to let users change panels positions more flexibly soon enough anyway :)

PR #11914, opened 2018-10-24
Initial import of the REvent API ##refactor

radare:


radare:
See https://github.com/skypjack/eventpp for discussion

radare:
Other questions to discuss:

* event type can be `const char *`... maybe more versatile than int, but slower to process
* eprintf if no event listener? imho nope
* thread safetiness (how)

libr/include/r_util/r_event.h
```diff
@@ -0,0 +1,33 @@
...
+
+#ifndef R_EVENT_H
+#define R_EVENT_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct r_event_t {
+	void *user;
```
ret2libc: I think this is not necessary. Each event should have all the right data passed through `data`

libr/include/r_util/r_event.h
```diff
@@ -0,0 +1,33 @@
...
+
+typedef enum {
+	R_EVENT_META_SET = 0,
+} REventType;
+
+typedef struct r_event_meta_t {
+	int type;
+	ut64 addr;
+	const char *string;
+} REventMeta;
```
ret2libc: This is specific for each REventType, isn't it?

libr/include/r_util/r_event.h
```diff
@@ -0,0 +1,33 @@
...
+
+typedef void (*REventCallback)(REvent *ev, REventType type, void *data);
+
+R_API REvent *r_event_new(void *user);
+R_API void r_event_free(REvent *ev);
+R_API void r_event_hook(REvent *ev, REventCallback cb);
+R_API void r_event_unhook (REvent *ev, REventCallback cb);
```
ret2libc: if you want to try the sys/pre-commit-indent.sh in your git hooks.. :D

libr/util/event.c
```diff
@@ -0,0 +1,37 @@
+/* radare2 - MIT - Copyright 2018 - pancake */
+
+#include <r_util.h>
+
+R_API REvent *r_event_new(void *user) {
+	REvent *ev = R_NEW0 (REvent);
+	if (ev) {
+		ev->callbacks = r_list_newf (NULL);
```
ret2libc: more than a list, I think we need an hashtable (you can use `ht_new`). If a listener register itself for the event R_EVENT_META_SET, there's no reason why it should be called for other kind of events too.

libr/include/r_util/r_event.h
```diff
@@ -0,0 +1,33 @@
...
+	int type;
+	ut64 addr;
+	const char *string;
+} REventMeta;
+
+typedef void (*REventCallback)(REvent *ev, REventType type, void *data);
+
+R_API REvent *r_event_new(void *user);
+R_API void r_event_free(REvent *ev);
+R_API void r_event_hook(REvent *ev, REventCallback cb);
```
ret2libc: this should have a `REventType type` as well. You want to register a listener for a particular type of event, not one that handles everything.

ret2libc:
> event type can be const char *... maybe more versatile than int, but slower to process

Mmm, I like it as you did, personally.

> eprintf if no event listener? imho nope

I agree, nope!

> thread safetiness

I think we can just start simple with a big giant lock on the entire REvent thing. every method would lock it and unlock it when it finishes. Given our low-concurrency level in r2codebase, it shouldn't be a big problem and it can be easily improved in the future, if needed.

libr/include/r_util/r_event.h
radare: yes. the void* could be a union

libr/include/r_util/r_event.h
radare: The idea is that REvent will be associated to RCore, this is the instance associated, this will be important to handle tasks and such. Mainly because who will send the event doesnt need to be n RCore, but the code that handles the events must be in RCore (as core plugins, whatever..)

libr/include/r_util/r_event.h
radare: the problem i see is that we shold define ALL the event structs and types in util, and we cant register new types at runtime or over versions. this makes the event system dependant on the version of r2, which is not really a problem because we are already breaking apis frequently to ensure sanity :D

libr/include/r_util/r_event.h
radare: sounds like a plan

libr/util/event.c
radare: what if you want to hook all events? my idea was to filter the event type in the hook itself. because i dont think we'll need to add more than 10 event hooks, unless we abuse the event system for other things like rbinpull or so

libr/include/r_util/r_event.h
radare: my idea was to handle the event type inside the callback. what other ppl think about this? this way its more versatile and you can do everything in one hook, and not having to create 200 hooks to handle all the possible events and not be able to detect when. you are missing an event

radare:
i think that if we have RCore in void *user we can just use the r_task api from there and handle the safetness in the hook itself instead of inside the REvent api. cc @thestr4ng3r

radare:
.

libr/include/r_util/r_event.h
ret2libc: How many times do you think you'll have the same code that handles both R_EVENT_META_SET and, for example, R_EVENT_FUNCTION_ADD? I think these two pieces of code will be very different, so if you put all together in one hook it will just be a lot of "if (type == ...)". IMHO it's better to have X different functions that just do one thing, instead of one that does a lot of things. We already went through the `switch (X) { case A: case B: ...` path with commands, and it's now a mess, where more granular function would have been better. This is mostly the same IMHO.

libr/util/event.c
thestr4ng3r: What about a bitmask with different event types? This obviously limits us to 32/64 types, but maybe some can be put together to an event "scope".

ret2libc: we could just use a ALL type during r_event_hook and just register the listenere in all type of events.

libr/include/r_util/r_event.h
ret2libc: ok

ret2libc:
> i think that if we have RCore in void *user we can just use the r_task api from there and handle the safetness in the hook itself instead of inside the REvent api.

I'm not sure this would be safe against concurrent accesses to the REvent API

libr/util/event.c
thestr4ng3r: Also, I think instead of a list I would intuitively use RVector. I wouldn't expect callbacks to be added or removed randomly all the time, but you will iterate the list very often, so RVector will probably be the better choice.

radare: i think we will have more than 64 types of events at some point.. unless we do sub-types

libr/include/r_util/r_event.h
radare: this reminds me to the discussion on merging callback handlers in the esil vm :P

radare:
no, concurrent access to the event api is not solved ... we should forbid such thing to avoid infinite loops, so maybe a bool and assert if false when trying to send a message should be fine

thestr4ng3r:
Concurrent (threaded) access to REvent would not be safe, but basically nothing in r2 is at the moment anyway. If you use tasks, you should have no problem for now. If we will have real threads at some point, we can add a lock then I would say.

libr/util/event.c
ret2libc: Why not just the hashtable? It's really simple to create/iterate/whatever. It doesn't limit the number of events we can have, it's fast, it's fast to get the listeners from a given type...

radare:
with a bool we can avoid r_event_send to be called from another r_event_send, that's not black magic

libr/util/event.c
thestr4ng3r: My point not about the hashtable, what I meant was if the question is RList vs. RVector, then RVector is probably the better choice here.

How would you do the hashtable exactly? REventType -> List of Callbacks? Then how do we handle the case that you want to subscribe to multiple or all events?

ret2libc:
```C
int add_callback(void *user, const char *key, const void *value) {
    RVector *v = (RVector *)value;
    REventCallback *cb = (REventCallback *)user;
    r_vector_push (v, cb);
    return 1;
}

R_API void r_event_hook(REvent *ev, REventType type, REventCallback cb) {
    if (type == R_EVENT_ALL) {
        ht_foreach (ev->callbacks, add_callback, cb);
    } else {
        RVector *v = ht_find (ev->callbacks, type, NULL);
        r_vector_push (v, cb);
    }
}
```

thestr4ng3r: Yep, lgtm too. This should make it quite efficient to call the callbacks.

libr/include/r_util/r_event.h
thestr4ng3r: Yes, and you can still hook one callback to multiple event types and do a switch if it makes sense for the case, or hook separate callbacks for each type.

thestr4ng3r:
> with a bool we can avoid r_event_send to be called from another r_event_send, that's not black magic

I think there is no general solution for this. A callback should definitely be able to send events in certain cases. The two options to solve it are blocking events for some time (the bool you mentioned) or unregistering the own callback. I would say we should have both options and the respective callbacks should do the one that is appropriate. This is also the approach that is commonly done in Qt with signals/slots.

XVilka:
Some systems define some small set of priorities, so callbacks with lower priority can be interrupted by callbacks with higher priorities.

libr/util/event.c
radare: agree on vector

radare:
@ret2libc can you update the PR (or open another one). with your rvector + hashtable proposal?

ret2libc:
Sure, give me a couple of days though.

PR #11924, opened 2018-10-24
Fix leak in anal fcn

devnexen:


radare:
thats not how to free a hint.  use r_anal_hint_free instead of free

devnexen:
Correct, missed the sub allocations.

PR #11930, opened 2018-10-24
Stop pdR when encountering a ret

astuder:
This PR stops recursive disassembly with pdR when encountering a ret op-code. This prevents disassembly of unrelated code.


radare:
Cc @skuater

radare:
Maybe we want to trace also the push; ret..

astuder:
Thinking about it: one scenario where stopping at ret is undesirable is a ret skipped by a conditional jump. E.g. logic checking input parameters. With this pull request, pdR would stop at the ret, whereas previously it would continue.

radare:
merged, we can think on an emulated-friendly pdR version with support for that stuff :)

Issue #11925, opened 2018-10-24
RList asserts in /R


radare:
`r2 -qc/R /bin/ls`

problem:

in. cmd_search.c:971

```
-> 971 			r_list_append (localbadstart, (void *) (intptr_t) idx);
```

using intptr_t is a bad usage of the api because if idx==0 the element will not be added

ret2libc:
Ok I will solve this with the `ht` implementation, since actually `badstart` is just a map to check which addresses needs to be skipped and the ht data structures handle well NULLs/0s. Also, it will be much more efficient.

ret2libc:
fixed with https://github.com/radare/radare2/commit/08422752f2e5eeca09a1bb70f3571833ccbe426a

PR #11932, opened 2018-10-25
if idx==0 return before r_list_append

minhnxd:
Fixes: #11925


minhnxd:
Closing, travis isn't happy with this change

radare:
The fix is to not use rlist to store intptr and not to use nullptr as 0. So use rvector or any other data structure that allows any arbitrary number to be stored without having to abuse the heap

PR #11944, opened 2018-10-26
Refactoring and optimization for the RBin.language identification ##bin

radare:
This commit can be benchmarked:
```
              swift app  libr_core
    old code:   9549      37147
    new code:    23       17232
```

This means that now identifying language in a binary is from 2 to 400 times faster depending on the binary.

For DVIA iOS app this is 1s less


libr/bin/obj.c
```diff
@@ -101,10 +101,39 @@ R_API RBinObject *r_bin_object_new(RBinFile *binfile, RBinPlugin *plugin, ut64 b
 	return o;
 }
 
+static void r_bin_filter_classes(RBinFile *bf, RList *list) {
```
ret2libc: this should be R_API or we should reveme the `r_bin_` prefix.

libr/bin/blang.c
```diff
@@ -0,0 +1,194 @@
...
+				swiftIsChecked = true;
+			}
+			if (hasswift || check_swift (sym)) {
+				info->lang = "swift ";
+				return R_BIN_NM_SWIFT;
+			}
+		}
+		if (!cantbe.cxx) {
+			bool hascxx = false;
+			if (!ft || (!strstr (ft, "mach") && !strstr (ft, "elf"))) {
```
ret2libc: this check seems independent of the current symbol, so it could be done before the loop, just one time, I think.

libr/bin/blang.c
```diff
@@ -0,0 +1,194 @@
...
+				}
+				cxxIsChecked = true;
+			}
+			if (hascxx || check_cxx (sym)) {
+				canBeCxx = true;
+				cantbe.cxx = true;
+			}
+		}
+		if (!cantbe.objc) {
+			if (!ft || (!strstr (ft, "mach") && !strstr (ft, "elf"))) {
```
ret2libc: this check seems independent of the current symbol, so it could be done before the loop, just one time, I think.

libr/bin/blang.c
```diff
@@ -0,0 +1,194 @@
...
+				continue;
+			}
+			if (check_objc (sym)) {
+				info->lang = "objc";
+				return R_BIN_NM_OBJC;
+			}
+		}
+		if (!cantbe.dlang) {
+			bool hasdlang = false;
+			if (!ft || (!strstr (ft, "mach") && !strstr (ft, "elf") && !strstr (ft, "pe"))) {
```
ret2libc: this check seems independent of the current symbol, so it could be done before the loop, just one time, I think.

libr/bin/mangling/rust.c
```diff
@@ -0,0 +1,76 @@
+/* radare - LGPL - Copyright 2011-2018 - pancake */
+
+#include <r_bin.h>
+
+#define RS(from, to) (replace_seq ((const char **)&in, &out, (const char *)(from), to))
+
+static bool replace_seq (const char **in, char **out, const char *seq, char value) {
```
ret2libc: check with `./sys/clang-format-diff.py`... spaces before `(` etc.

Maijin:
Faster but does it work - https://travis-ci.org/radare/radare2/jobs/446477504#L1154 ?

ret2libc:
> Faster but does it work - https://travis-ci.org/radare/radare2/jobs/446477504#L1154 ? 

it seems it does and it improves the situation, because it demangle some symbols that were not before. Actually... even the symbols with `imp.` should be demangled IMHO, but that's another story.

Maijin:
Yeah you a right, it looks better

ret2libc:
Of course, before merging please remember to remove the `#if 0`/`#if 1` you added :D

radare:
yeah  and ill cleanup a bunch more things, dont worry ;) its still wip

libr/bin/obj.c
radare: done

libr/bin/blang.c
radare: agree. fixed

libr/bin/mangling/rust.c
radare: sys/indent with clang-format does it wrong.

1) space before ? and : 
2) indents functions with tabs + 2 spaces wtf

radare:
imports are also demangled. the problem is that before it was only demangling some. and for some file formats, now its doing for all of them. So i would probably go for fixing the test and continue the optimizations in another pr when this one. is merged

libr/bin/mangling/rust.c
radare: im fine with the current clang-format output, but we should probably fix that in a separate PR, otherwise reviewing the changes will be a mess

radare: r2pm -i r2i

PR #11949, opened 2018-10-26
Remove all S commands ##refactoring

Maijin:
Let see what travis says


Maijin:
ok so recap of this test:

@pancake  From travis perspective there are only 24 tests related to S that are failing when removing S completly.

On those 24 tests 
- 23 tests are only making use of S for display so they can be replaced by om :)
- 1 test is related to the S adding “map” functionnality (the screenshot). This functionnality need to be verified working with om and test moved to db/io

image

PR #11952, opened 2018-10-26
Add asm.section.{perm,name}

radare:


radare:
We need an O(1) api in rbin and stop using io-sections, which is the only reason for using it right now. all this info must be held and queried from rbin directly.

the if 0 code will be removed when travis agrees

PR #11935, opened 2018-10-25
fix x86 cmp [mem], imm not getting data xrefs to mem

ulidtko:
Hey, I noticed that one specific instruction doesn't generate data xrefs as it should: `cmp [mem], imm`. The  `cmp reg, [mem]` form is fine, but with mem location on the left -- nope.

The fix seem to drop a stray line over there. No idea what it was supposed to do, it doesn't look intentional at all. Why assume that the immediate operand will be an address? Why the elaborate `op->ptr` calculations before? Why the difference between INSOP(0)/INSOP(1) cases?

Verified via `pij`, works more correctly now AFAICT.

PR #11954, opened 2018-10-26
Use op->val instead of op->ptr for fetching immediate in cmp [mem],imm instr

sivaramaaa:
This may fix "x86 cmp [mem], imm not getting data xrefs to mem" issue , without affecting type constraints ! (suggested by this PR [#11935](https://github.com/radare/radare2/pull/11935))


sivaramaaa:
```
[  ]  33  cmd_types: struct offset for dst operand with ta
[XX]
Regression: run_tests.sh
Command:  /home/travis/build/radare/radare2/install/bin/radare2 -e scr.color=0 -N -q -i /tmp/r2-regressions//cmd_types-OORU9x/rad   - > /tmp/r2-regressions//cmd_types-OORU9x/out 2> /tmp/r2-regressions//cmd_types-OORU9x/err
File: -
Script: e asm.bits = 32
"td struct foo {int a; int b ; int c;}"
wx 8b40048951048d500883780800
?e
ta foo.b @ 0x00000000
ta foo.b @ 0x00000003
ta foo.c @ 0x00000006
ta foo.c @ 0x00000009
pd 4
Diff: --- /tmp/r2-regressions//cmd_types-OORU9x/exp	2018-10-26 13:38:46.716035161 +0000
+++ /tmp/r2-regressions//cmd_types-OORU9x/out	2018-10-26 13:38:46.744035225 +0000
@@ -2,4 +2,4 @@
             0x00000000      8b4004         mov eax, dword [eax + foo.b] ; [0x4:4]=0x508d0451
             0x00000003      895104         mov dword [ecx + foo.b], edx
             0x00000006      8d5008         lea edx, [eax + foo.c]      ; 0x8
-            0x00000009      83780800       cmp dword [eax + foo.c], 0
+            0x00000009      83780800       cmp dword [eax + foo.c], 0  ; [0x8:4]=0x8788308
```
@radare , @Maijin  what do u guys think about this change ?

radare:
foo.c is in the offset 8 of the struct foo. so the static analysis takes that 8 and uses it as address to be resolved. the problem here is that statically we dont have the value of eax. so this comment shuodlnt appear unless we allow the static to use rreg and let the user define the values of the registers with ar or by using an eval var like its done for the gp register in. mips

sivaramaaa:
> the problem here is that statically we dont have the value of eax. so this comment shuodlnt appear 

yup exactly i agree with that , but the problem is we won't be able to distinguish
 `cmp dword [eax + foo.c], 0  ` with `cmp [0xdeadbeef], val` instr

ulidtko:
Hey, let's join our effort and collaborate. That `patch-1` branch in my fork has some test fixes. I think you can cherry-pick the commit via `git remote add`, rather than reproducing all the steps I've already done. Travis has to get green, anyway.

In this PR, there's a thing worth changing IMO, that is to add the same `if imm` block below, symmetrically handling the case of [mem] in INSOP(1) in the exact same way as [mem] in INSOP(0).

libr/anal/p/anal_x86_cs.c
```diff
@@ -2172,7 +2172,9 @@ static void anop(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len, csh
 				op->stackptr = regsz;
 				op->type |= R_ANAL_OP_TYPE_REG;
 			}
-			op->ptr = INSOP(1).imm;
+			if (INSOP(1).type == X86_OP_IMM) {
+				op->val = INSOP(1).imm;
+			}
 			break;
 		default:
 			switch (INSOP(1).type) {
```
ulidtko: Here, below, the exact same handling of `case X86_OP_MEM` as for `switch (INSOP(0).type)`.
Suggest adding 
```
		if (INSOP(0).type == X86_OP_IMM) {
			op->val = INSOP(0).imm;
		}
```

here too, before `break`.

sivaramaaa:
` 0x00000000      8b4004         mov eax, dword [eax + foo.b] ; [0x4:4]=0x508d0451`

@radare IMHO we should fix the testes for now , since the issue u are talking about existed even before my fix (see the above ^ example of `mov`, i have only made changes to `cmp`)

sivaramaaa:
> Hey, let's join our effort and collaborate,  you can cherry-pick the commit 

yup done :+1:

ulidtko:
@radare hi, I asked you on telegram, and the question remains. Why doesn't Travis apply r2r overlay before running the r2r tests?

See, [these lines](https://travis-ci.org/radare/radare2/jobs/447309067#L614-L617):

> You are not in master, overlay will not be applied

I can see `overlay.sh` is scripted that way; the `apply` handler explicitly checks `if [ "$B" = master ]` and refuses to continue if it's not.

I think it should not do that! `overlay.sh apply` should work in PR branches too. Otherwise, how is it even possible to synchronize test changes with related r2 changes in a pull request?

Please relax that check, or correct me if I'm wrong.

PR #11965, opened 2018-10-27
WIP: Initial import of the scr.lock ##visual

radare:
DEMO: https://asciinema.org/a/P6210DmJoUFgGtb68sKThQKGW

Known issues:

* [x] Block ^Z
* [x] Current input method doesn't handles backspace
* [x] Password is saved in the console history, so it will be better to move from eval to a command and ask for password after running the command outside the prompt loop.

Wishlist/Future:

* [ ] Launch it after N minutes of inactivity
* [ ] noecho|icanon is required for iterm to show the key emoji, but read seems to fail that way
* [ ] Not tested on windows (probably not working)


radare:
Known issues:

* [ ] Not tested on windows (probably not working)
* [ ] r_cons_password() should be implemented instead of harcoding the logic in core
* [ ] Current input method doesn't handles backspace
* [ ] Password is saved in the console history, so it will be better to move from eval to a command and ask for password after running the command outside the prompt loop.
* [ ] noecho|icanon is required for iterm to show the key emoji, but read seems to fail that way

radare:
solved 2/4 issues mentioned

PR #11967, opened 2018-10-27
Remove all Sl commands ##refactoring

Maijin:


radare:
Kill it. I think this was done to save/restore contents of sections. But i think we can do the same dump/load actions with just wff and wtf.. but we probably need to add some helpers to avoid having to specify address and size of each. Which i think we have them already in ?$?. Do u think we shohld have om subcommands for this?

Maijin:
Would be nice to have a way to dump a map yep, something like `omd` or in `wt` idk.

PR #11971, opened 2018-10-27
Fix r_log when used for r_assert

ret2libc:
- Split r_log in r_vlog/r_log and make r_assert_log use r_vlog because it
needs to provide a va_list.
- Rename _r_internal_logging to r_log because that's a R_API function
and it needs a good name
- Make R2_CHECKS_LEVEL dependent on 'buildtype' option
- Make R_LOGLVL_WARN the default log level when in debug mode (default)


ret2libc:
Yes but that's a good thing! Now that asserts are printed with RCons they are caught by r2r and there are many failures. It could just be that some asserts are simply wrong, but it may also be we should rethink why some things are called in a certain way... I'll look at them, but those are a separate problem from this PR.

ret2libc:
* introduce ASSERT_STDOUT macro to print assert failures to stdout
* use ASSERT_STDOUT in the Travis env with asan
* use list of callbacks in RLog instead of just one.

PR #11973, opened 2018-10-28
asm.bb.middle: Realign disasm on bb start

kazarmy:
Fixes the problem shown in https://github.com/radare/radare2/issues/4007#issuecomment-175016985. Currently for disasm only. Not sure whether defaulting it to true is ok performance-wise.


radare:
is this a fix for the jump-in-the-middle issues in the default analysis?

kazarmy:
not sure what you mean by that ... any r2 issues to look at?

kazarmy:
26f31f8714859a37f607a75e9a4c6abd4e243436 Reverted optimization since it is unsound if basic blocks can nest or overlap.

PR #11994, opened 2018-10-30
Modernize RFlag with asserts, boolified and more ##refactor

radare:


libr/flag/flag.c
```diff
@@ -178,20 +168,20 @@ R_API RFlagItem *r_flag_item_clone(RFlagItem *item) {
...
-		free (item->color);
-		free (item->comment);
-		free (item->alias);
-		/* release only one of the two pointers if they are the same */
-		if (item->name != item->realname) {
-			free (item->name);
-		}
-		free (item->realname);
-		free (item);
+	r_return_if_fail (item);
```
ret2libc: On free methods we said we don’t want to use rreturn because it is alright to use null in these cases

PR #12009, opened 2018-10-30
/Cd search x509 certs

wargio:
before it was searching for an invalid `magic` value, but instead it should be searching for a bitmasked magic


wargio:
this search is limited to certificates that are not bigger than 0x7fff

Issue #11910, opened 2018-10-23
Visual Panels crashes on Windows


BlaCoiso:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Windows, x86, 64 bits
| File format of the file you reverse (mandatory)      | none
| Architecture/bits of the file (mandatory)            | none
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.1.0-git 1 @ windows-x86-64 git. commit: 6941e5bd6649228024ba9964a2c4cc01a799629f build: Tue 10/23/2018__ 9:57:18.34

### Expected behavior
Entering Visual Panels mode works

### Actual behavior
radare just crashes when entering the mode

### Steps to reproduce the behavior 
`radare2 -c V! -` or just do `V!` inside r2

radare:
can you provide a backtrace?

BlaCoiso:
```
Access violation - code c0000005 (first chance)
radare2!addMenu+0x1c9:
00007ff6`e9de8a79 8b00            mov     eax,dword ptr [rax] ds:00000000`00000012=????????
```
backtrace:
```
00 radare2!addMenu+0x1c9 [d:\radare2\libr\core\panels.c @ 1874] 
01 radare2!initPanelsMenu+0x2ff [d:\radare2\libr\core\panels.c @ 1955] 
02 radare2!r_core_visual_panels+0xb1 [d:\radare2\libr\core\panels.c @ 2744] 
03 radare2!r_core_visual_cmd+0xad5 [d:\radare2\libr\core\visual.c @ 2014] 
04 radare2!r_core_visual+0x16c [d:\radare2\libr\core\visual.c @ 3244] 
05 radare2!cmd_visual+0x7b [d:\radare2\libr\core\cmd.c @ 1317] 
06 radare2!r_cmd_call+0x23d [d:\radare2\libr\core\cmd_api.c @ 239] 
07 radare2!r_core_cmd_subst_i+0x30c4 [d:\radare2\libr\core\cmd.c @ 2929] 
08 radare2!r_core_cmd_subst+0x4ec [d:\radare2\libr\core\cmd.c @ 1937] 
09 radare2!r_core_cmd+0x425 [d:\radare2\libr\core\cmd.c @ 3650] 
10 radare2!r_core_prompt_exec+0x4a [d:\radare2\libr\core\core.c @ 2656] 
[...]

PR #12014, opened 2018-10-30
Fix crashes with visual panels on Windows

BlaCoiso:
This should fix issue #11910

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -1860,12 +1860,12 @@ static void addMenu(RCore *core, const char *parent, const char *name, RPanelsMe
 		return;
 	}
 	if (parent) {
-		ut64 addr = r_num_math (core->num, ht_find (panels->mht, parent, NULL));
+		ut64 addr = ht_find (panels->mht, parent, NULL);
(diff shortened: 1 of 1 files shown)


libr/core/panels.c
```diff
@@ -2342,7 +2342,7 @@ static bool init(RCore *core, RPanels *panels, int w, int h) {
 	panels->isResizing = false;
 	panels->can = createNewCanvas (core, w, h);
 	panels->db = sdb_new0 ();
-	panels->mht = ht_new ((DupValue)strdup, (HtKvFreeFunc)mht_free_kv, (CalcSize)strlen);
+	panels->mht = ht_new (NULL, (HtKvFreeFunc)mht_free_kv, (CalcSize)strlen);
```
radare: I think you have to keep this strdup

BlaCoiso: It has to be removed, since the value on the hashtable is not a string (it's a pointer to a panel item, I removed because it was trying to duplicate the item pointer as if it was a string)

radare: oh i see, lgtm

PR #12020, opened 2018-10-31
Change ubase64.c to base64.c ##refactoring

Maijin:


radare:
it was changed because there was another base64.o coming from sdb that is bundled inside the same library archive and creates a symbol conflict. so imho -1 for this change

Maijin:
mokay.

Issue #11075, opened 2018-08-13
ood command can't reopen files with spaces (windows) or if an argument has a `/` (linux)


BlackYoup:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Exherbo / Windows x86_64
| File format of the file you reverse (mandatory)      | PE, ELF
| Architecture/bits of the file (mandatory)            | x86
| r2 -v full output, **not truncated** (mandatory)     | radare2 2.9.0-git 19121 @ linux-x86-64 git.2.7.0-400-g915b3deeb commit: 915b3deeb19767f1b2fabd20a884752e6c1b4688 build: 2018-08-14__00:13:45 (same commit for the Windows binary)

### Expected behavior

`ood` should be able to:
- reopen binaries that have spaces in them or in the path (Windows)
- reopen binaries with a `/` in the given arguments (Linux)

### Actual behavior

Currently, `ood`
- Fails at parsing the path of the binary and gives a wrong path to `CreateProcess` (Windows)
- Fails at parsing the path of the binary and gives a wrong path to `execvp` (Linux)

#### Windows

- open a binary that's in a folder having spaces (open it in debug or not, doesn't matter)
- try to `ood`
- `CreateProcess` will fail because it can't find the file

##### Linux

- open a binary
- try to `ood /tmp/wedontcare`
- `execvp` will fail at finding the binary 

### Additional Logs, screenshots, source-code,  configuration dump, ...

`ood` works for most cases but some edge cases are not correctly handled, mostly because of https://github.com/radare/radare2/blob/915b3deeb19767f1b2fabd20a884752e6c1b4688/libr/core/cmd_open.c#L880

Merging the `path` and `arguments` here lead to further (quite hard) parsing which could be avoided. I think it  would make more sense to pass both the current debugged file and arguments in two separate arguments instead of merging them into one `cmd`. Something like `static int fork_and_ptraceme(RIO *io, int bits, const char *binpath, char **arguments) {}` for example (not sure about the signature, you get the idea).

Unless something is preventing this and this is why it's currently done like this? That's something I'm willing to do in any case if I can help

radare:
we merged a PR fixing this, considering it correct now, pls complain if not

PR #9353, opened 2018-02-09
fix ood whitespace (#9302)

leberus:
This aims to fix #9302 

<pre>
$ pwd
/home/osalvador/lab/r2/test1 2
$ r2 -d ./test
Process with PID 15663 started...
= attach 15663 15663
bin.baddr 0x00400000
Using 0x400000
Unknown DW_FORM 0x06
asm.bits 64
 -- Finnished a beer
[0x7f1ab197a1f0]> ood arg arg2
Wait event received by different pid 15663
Process with PID 15666 started...
File dbg:///home/osalvador/lab/r2/test1 2/test  arg arg2 reopened in read-write mode
= attach 15666 15666
Unknown DW_FORM 0x06
15666
[0x7f802224a1f0]> dc
argv[0]: /home/osalvador/lab/r2/test1 2/test
argv[1]: arg
argv[2]: arg2
[0x7f8021f604a9]>
</pre>

<pre>
$ r2 -d ~/lab/r2/test1\ 2/test
Process with PID 15694 started...
= attach 15694 15694
bin.baddr 0x00400000
Using 0x400000
Unknown DW_FORM 0x06
asm.bits 64
 -- Disable these messages with 'e cfg.fortunes = false' in your ~/.radare2rc
0x7ff36a4381f0]> ood arg3 arg4
Wait event received by different pid 15694
Process with PID 15696 started...
File dbg:///home/osalvador/lab/r2/test1 2/test  arg3 arg4 reopened in read-write mode
= attach 15696 15696
Unknown DW_FORM 0x06
15696
[0x7f63b85a91f0]> dc
argv[0]: /home/osalvador/lab/r2/test1 2/test
argv[1]: arg3
argv[2]: arg4
[0x7f63b82bf4a9]>
</pre>

I tried to test it as much as possible.

PR #12023, opened 2018-10-31
Fix args escape and unescape bugs when debug

bet4it:
e973deadca3c6c417825bcb04b02e2a38202f143 is useless after 41cb6ef0d2f39469cd30961fe81b6f7836dd504b, because the file path we get with `fd_get_name` is already an absolute path:
https://github.com/radare/radare2/blob/7b01b96f9070bbce45407f16b6382924c64ec479/libr/bin/bin.c#L217

What has been done in #9353 is totally wrong. It's impossible to separate unescaped args in a string. We should escape args before, separate and unescape them when we need them.

I think the Linux part of #11075 has been fixed.

By the way, I'm curious that is it proper to use the absolute path when we use `ood` to debug bins, even if we use relative path when we debug it in the first time? At least `argv[0]` has been changed and it may have an effect on running.

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -880,8 +880,8 @@ R_API void r_core_file_reopen_debug(RCore *core, const char *args) {
 		return;
 	}
 	int bits = core->assembler->bits;
-	char *oldname = r_file_abspath (binpath);
-	char *newfile = r_str_newf ("dbg://%s %s", oldname, args);
+	char *escaped_path = r_str_arg_escape (binpath);
+	char *newfile = r_str_newf ("dbg://%s %s", escaped_path, args);
 	char *newfile2 = strdup (newfile);
 	desc->uri = newfile;
 	desc->referer = NULL;
@@ -905,7 +905,7 @@ R_API void r_core_file_reopen_debug(RCore *core, const char *args) {
 	}
 #endif
 	r_core_cmd0 (core, "sr PC");
-	free (oldname);
+	free (escaped_path);
 	free (binpath);
 	free (newfile);
 }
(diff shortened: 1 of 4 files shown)


ret2libc:
LGTM, though we don't have proper debugging tests so... :)

libr/core/cmd_open.c
```diff
@@ -880,8 +880,8 @@ R_API void r_core_file_reopen_debug(RCore *core, const char *args) {
 		return;
 	}
 	int bits = core->assembler->bits;
-	char *oldname = r_file_abspath (binpath);
-	char *newfile = r_str_newf ("dbg://%s %s", oldname, args);
+	char *escaped_path = r_str_arg_escape (binpath);
+	char *newfile = r_str_newf ("dbg://%s %s", escaped_path, args);
```
radare: maybe we can use a different char than space to separate program name and arguments (as a hack)

libr/io/p/io_debug.c
```diff
@@ -382,7 +381,7 @@ static int fork_and_ptraceme_for_mac(RIO *io, int bits, const char *cmd) {
 		}
 		// XXX: this is a workaround to fix spawning programs with spaces in path
 		if (strstr (argv[0], "\\ ")) {
-			argv[0] = r_str_replace (argv[0], "\\ ", " ", true);
+			argv[0] = r_str_arg_unescape (argv[0]);
```
radare: wrong uasge, this returns. int, just ignore the result because this function modifies the argv[0] already

radare: also, the function signature of this function is not in the .h so it segfaults hard

libr/core/cmd_open.c
bet4it: If we do this, we don't need to escape the file path.
We generate a URL which `radare2` can recognise here. As we don't specify special char to separate URL, it's better for us not to change it.

PR #12027, opened 2018-10-31
Improve function param detection by removing func-number from function name

sivaramaaa:
This fixes the issue reported by @Maijin 

* Before 

before

* After

after


Maijin:
Great!

Maijin:
@ITAYC0HEN @xarkes

PR #12029, opened 2018-10-31
Remove aae dup and enable anal.types.constraint in aaaa ##anal

Maijin:


XVilka:
We need proper way to access/modify constraints information to properly test them: https://github.com/radare/radare2/issues/11828

Maijin:
@XVilka adding them in aaaa is a good way to let people try them though.

PR #12031, opened 2018-10-31
Following up on porting self:// to BSD.

devnexen:
Here the OpenBSD support, unfortunately not storing the path
but the proc mapping lookup is simpler.


XVilka:
We have so much ifdefs here. What about splitting that to a separate functions and call them on ifdef? In my opinion it will be easier to read. @radare @ret2libc

devnexen:
Or maybe not necessarily separate functions but distinct full blocks instead ?

PR #12036, opened 2018-11-01
anal.jmpmid: Add proper overlapping bb support (supports loops) ## anal

### LINKED ISSUE #10420: Graph view doesn't handle unaligned opcodes
### Work environment

| OS/arch/bits (mandatory)                             | Arch Linux / x86_64 / 64
| r2 -v full output, **not truncated** (mandatory)     | radare2 2.7.0-git 18486 @ linux-x86-64 git.2.6.0-308-g3feb95079
commit: 3feb95079e7a28a43e0f5bbabc04405e0337da94 build: 2018-06-20__03:53:51

### Expected behavior

Expected behavior is for jumps/calls to unaligned opcodes to display the correct opcodes in the graph view.

### Actual behavior

Although doing `s <address_jumped_to>` works great in visual mode, in the graph view, the node after the jump shows the closest aligned opcode.

```
radare2 888
aaaa
VV
gAb
# The selected node contains the incorrect opcode
<Space>
# Visual mode shows what it should be
```
[888.zip](https://github.com/radare/radare2/files/2117856/888.zip)

kazarmy:
Fixes #10420. Now defaulting anal.jmpmid to true.

### DIFF
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -328,7 +328,7 @@ R_API void r_anal_fcn_free(void *_fcn) {
 	free (fcn);
 }
 
-// TODO: use this function if anal.jmpmid == false
+// Use this function if anal.jmpmid == false
 static RAnalBlock *bbget(RAnalFunction *fcn, ut64 addr) {
 	RListIter *iter;
 	RAnalBlock *bb;
@@ -342,6 +342,30 @@ static RAnalBlock *bbget(RAnalFunction *fcn, ut64 addr) {
 		}
 	}
 	return NULL;
+}
+
+// Use this function if anal.jmpmid == true
+static RAnalBlock **bbget_all(RAnalFunction *fcn, ut64 addr) {
+	RListIter *iter;
+	RAnalBlock *bb;
+	RAnalBlock **bbs = malloc (sizeof (RAnalBlock *) * (r_list_length (fcn->bbs) + 1));
+	RAnalBlock **ptr;
+	r_return_val_if_fail (bbs, NULL);
+	ptr = bbs;
+	r_list_foreach (fcn->bbs, iter, bb) {
+		ut64 eaddr = bb->addr + bb->size;
+		if ((bb->addr >= eaddr && addr == bb->addr)
+		    || (addr >= bb->addr && addr < eaddr)) {
+			*ptr++ = bb;
+		}
+	}
+	if (ptr == bbs) {
+		free (bbs);
+		return NULL;
+	} else {
+		*ptr = NULL;
+		return bbs;
+	}
 }
 
 static RAnalBlock *appendBasicBlock(RAnal *anal, RAnalFunction *fcn, ut64 addr) {
@@ -866,6 +890,7 @@ static int fcn_recurse(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut8 *buf, ut6
 	const int noncode = anal->opt.noncode;
 	const int addrbytes = anal->iob.io ? anal->iob.io->addrbytes : 1;
 	RAnalBlock *bb = NULL;
+	RAnalBlock **bbs = NULL;
 	RAnalBlock *bbg = NULL;
 	int ret = R_ANAL_RET_END, skip_ret = 0;
 	int overlapped = 0;
@@ -912,15 +937,19 @@ static int fcn_recurse(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut8 *buf, ut6
 	if (r_anal_get_fcn_at (anal, addr, 0)) {
 		return R_ANAL_RET_ERROR; // MUST BE NOT FOUND
 	}
-	bb = bbget (fcn, addr); // TODO: bbget_all()
+	if (anal->opt.jmpmid && x86) {
+		bbs = bbget_all (fcn, addr);
+		bb = bbs ? r_anal_bb_op_starts_at_which (bbs, addr) : NULL;
+		free (bbs);
+	} else {
+		bb = bbget (fcn, addr);
+	}
 	if (bb) {
-		if (!anal->opt.jmpmid || !x86 || r_anal_bb_op_starts_at (bb, addr)) {
-			r_anal_fcn_split_bb (anal, fcn, bb, addr);
-			if (anal->opt.recont) {
-				return R_ANAL_RET_END;
-			}
-			return R_ANAL_RET_ERROR; // MUST BE NOT DUP
+		r_anal_fcn_split_bb (anal, fcn, bb, addr);
+		if (anal->opt.recont) {
+			return R_ANAL_RET_END;
 		}
+		return R_ANAL_RET_ERROR; // MUST BE NOT DUP
 	}
 
 	bb = appendBasicBlock (anal, fcn, addr);
@@ -975,9 +1004,14 @@ repeat:
 			r_anal_hint_set_bits (anal, op.jump, op.hint.new_bits);
 		}
 		if (idx > 0 && !overlapped) {
-			bbg = bbget (fcn, addr + idx); // TODO: bbget_all()
-			if (bbg && bbg != bb
-			    && (!anal->opt.jmpmid || !x86 || r_anal_bb_op_starts_at (bbg, addr + idx))) {
+			if (anal->opt.jmpmid && x86) {
+				bbs = bbget_all (fcn, addr + idx);
+				bbg = bbs ? r_anal_bb_op_starts_at_which (bbs, addr + idx) : NULL;
+				free (bbs);
+			} else {
+				bbg = bbget (fcn, addr + idx);
+			}
+			if (bbg && bbg != bb) {
 				bb->jump = addr + idx;
 				overlapped = 1;
 				VERBOSE_ANAL eprintf("Overlapped at 0x%08"PFMT64x "\n", addr + idx);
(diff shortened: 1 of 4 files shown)


kazarmy:
The code xrefs are a known problem.

libr/anal/fcn.c
```diff
@@ -344,6 +344,31 @@ static RAnalBlock *bbget(RAnalFunction *fcn, ut64 addr) {
...
+		ut64 eaddr = bb->addr + bb->size;
+		if ((bb->addr >= eaddr && addr == bb->addr)
+		    || (addr >= bb->addr && addr < eaddr)) {
+			*ptr++ = bb;
+		}
+	}
+	if (ptr == bbs) {
+		free (bbs);
+		return NULL;
+	} else {
```
radare: Dint else after return

libr/anal/fcn.c
```diff
@@ -912,15 +938,19 @@ static int fcn_recurse(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut8 *buf, ut6
 	if (r_anal_get_fcn_at (anal, addr, 0)) {
 		return R_ANAL_RET_ERROR; // MUST BE NOT FOUND
 	}
-	bb = bbget (fcn, addr); // TODO: bbget_all()
+	if (anal->opt.jmpmid && x86) {
+		bbs = bbget_all (fcn, addr);
```
radare: Chk fir nul

libr/anal/fcn.c
```diff
@@ -344,6 +344,31 @@ static RAnalBlock *bbget(RAnalFunction *fcn, ut64 addr) {
...
 }
 
+// Use this function if anal.jmpmid == true
+static RAnalBlock **bbget_all(RAnalFunction *fcn, ut64 addr) {
+	RListIter *iter;
+	RAnalBlock *bb;
+	// TODO: remove need for malloc
+	RAnalBlock **bbs = malloc (sizeof (RAnalBlock *) * (r_list_length (fcn->bbs) + 1));
+	RAnalBlock **ptr;
+	r_return_val_if_fail (bbs, NULL);
```
ret2libc: no please. See https://github.com/radare/radare2/blob/master/libr/include/r_util/r_assert.h#L68 . `r_return_val_if_fail` should not be used for runtime errors and a failing malloc is a runtime error, something that could happen. Just use the usual `if (!bbs) return NULL`

libr/include/r_anal.h
```diff
@@ -1315,6 +1315,7 @@ R_API bool r_anal_bb_set_offset(RAnalBlock *bb, int i, ut16 v);
 R_API ut16 r_anal_bb_offset_inst(RAnalBlock *bb, int i);
 R_API ut64 r_anal_bb_opaddr_at(RAnalBlock *bb, ut64 addr);
 R_API bool r_anal_bb_op_starts_at(RAnalBlock *bb, ut64 addr);
+R_API RAnalBlock *r_anal_bb_op_starts_at_which(RAnalBlock **bbs, ut64 addr);
```
ret2libc: I think this API is not as good as the rest. It accepts a `RAnalBlock**`.. this array comes from `bbget_all` and it's only used in that case. I see no reasons to have bbget_all and r_anal_bb_op_starts_at_which.

I think you can just merge the two things in something like:
```
R_API RAnalBlock *r_anal_bb_op_starts_at_which(RAnalFunction *fcn, ut64 addr) {
    r_list_foreach (fcn->bbs, iter, bb) {
		ut64 eaddr = bb->addr + bb->size;
		if ((bb->addr >= eaddr && addr == bb->addr) || (addr >= bb->addr && addr < eaddr)) {
			if (r_anal_bb_op_starts_at (bb, addr)) {
			         return bb;
		        }
		}
	}
}
```

And:
```
bbs = bbget_all (fcn, addr + idx);
bbg = bbs ? r_anal_bb_op_starts_at_which (bbs, addr + idx) : NULL;
free (bbs);
```
would become:
```
bbg = r_anal_op_stars_at_which (fcn, addr + idx);
```

(ofcourse this is not tested and it doesn't compile, but you get the idea)

kazarmy: I know ... both bbget_all and r_anal_bb_op_starts_at_which are slated to be removed in a subsequent PR but the code now currently works and I want to keep this code for now so that I have something to compare against later.

radare:
Bbs

kazarmy:
It's checked for NULL [on the next line.](https://github.com/radare/radare2/pull/12036/files/5b707262afc6089b72cf0c67908ba8a0f453bfe4#diff-49a52329a1b3f5e5f62d571577391074R943)

libr/anal/fcn.c
kazarmy: ok 74673d503052f381378bd873f84a5b445eb299dd

libr/anal/fcn.c
kazarmy: ok a571473e63e26d036a86bed0a8c6c87d86d21dd0

libr/include/r_anal.h
ret2libc: Just do a comparison in local between the two versions. This API as it is is almost useless now, because it's just used in that place and it won't be used anywhere else, because it requires RAnalBlock**. Otherwise, make it static, but IMHO it should just be done well from the beginning. It's not such a big change.

We all know the "we'll fix it later" often becomes a "it will remain that way for at least few months/years".

kazarmy: ok 2814f3492cb34109931e787b86a3235d27520ce3

> We all know the "we'll fix it later" often becomes a "it will remain that way for at least few months/years".

Actually, this one has higher priority than most because ... ok I was just being silly

ret2libc: Thanks a lot, I appreciate :) It's much better now.

libr/anal/fcn.c
```diff
@@ -1859,14 +1857,16 @@ R_API int r_anal_fcn_add_bb(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 siz
...
 	if (addr == UT64_MAX) {
 		return 0;
 	}
-	r_list_foreach (fcn->bbs, iter, bbi) {
+	r_list_foreach (fcn->bbs, iter, bbi) { // TODO: bbi should already been given as a parameter
 		if (addr == bbi->addr) {
 			return R_ANAL_RET_DUP;
 		}
-		if (addr > bbi->addr && addr < bbi->addr + bbi->size) {
+		if (addr > bbi->addr && addr < bbi->addr + bbi->size
```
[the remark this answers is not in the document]
kazarmy: can't do that because the first comparison is `>` and not `>=`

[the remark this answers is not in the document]
ret2libc: Well, you could because the `addr == bbi->addr` is already checked in the previous lines.

libr/anal/fcn.c
```diff
@@ -328,16 +328,14 @@ R_API void r_anal_fcn_free(void *_fcn) {
...
+static RAnalBlock *bbget(RAnalFunction *fcn, ut64 addr, bool jumpmid) {
 	RListIter *iter;
 	RAnalBlock *bb;
 	r_list_foreach (fcn->bbs, iter, bb) {
 		ut64 eaddr = bb->addr + bb->size;
-		if (bb->addr >= eaddr && addr == bb->addr) {
-			return bb;
-		}
-		if ((addr >= bb->addr) && (addr < eaddr)) {
+		if (((bb->addr >= eaddr && addr == bb->addr)
```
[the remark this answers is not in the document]
kazarmy: ok c5d5a29bf319d710f0ce508493e1e552f02c72ca

> ... in both cases, I think there shouldn't be any check here

There might be some (buggy) arch code that forgot to set bb->size (and btw bb->size is an `int` so it can be negative), and I don't know enough to remove the check.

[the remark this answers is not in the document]
ret2libc: it's ok. as said, it's not strictly related to this PR but this check is either useless or it is a sign of something wrong ;)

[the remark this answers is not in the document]
ret2libc: Ah, the way we can solve these "might be some buggy arch code.." nowadays is by using `r_return_val_if` or `r_warn_if_reached` or `r_warn_if_fail`. They are good checks for things that we assume should always be true but may not be in case of some programming errors.

libr/anal/fcn.c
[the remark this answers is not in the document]
kazarmy: Logically yes, but in English terms it is not 100% correct to use `r_anal_bb_is_in_offset`.

libr/anal/fcn.c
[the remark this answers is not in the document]
ret2libc: yeah, random comment indeed ;)

libr/core/cconfig.c
```diff
@@ -2447,7 +2447,7 @@ R_API int r_core_config_init(RCore *core) {
 	SETCB ("anal.jmpabove", "true", &cb_anal_jmpabove, "Jump above function pointer");
 	SETCB ("anal.datarefs", "false", &cb_anal_followdatarefs, "Follow data references for code coverage");
 	SETCB ("anal.brokenrefs", "false", &cb_anal_brokenrefs, "Follow function references as well if function analysis was failed");
-	SETCB ("anal.jmpmid", "false", &cb_anal_jmpmid, "Continue analysis after jump to middle of instruction (x86 only) [WIP no loops]");
+	SETCB ("anal.jmpmid", "true", &cb_anal_jmpmid, "Continue analysis after jump to middle of instruction (x86 only)");
```
radare: woah

PR #12037, opened 2018-11-01
other r_bin API cleanup

ret2libc:
This is just a small step. Most of the functions I've moved into `bin_private.h` are only internal details that are not used outside of the RBin module and probably shouldn't. I think with some API refactoring there could even be other functions to be removed...

This will help me understanding what are the APIs that I actually need to test and what is the interface used by other modules when interacting with RBin. Moreover it makes it easier/clearer which parts we should be careful about when changing and which are just details.


libr/bin/bfile.c
```diff
@@ -669,7 +670,7 @@ R_API RBinFile *r_bin_file_find_by_name(RBin *bin, const char *name) {
 	return bf;
 }
 
-R_API RBinFile *r_bin_file_find_by_name_n(RBin *bin, const char *name, int idx) {
+RBinFile *r_bin_file_find_by_name_n(RBin *bin, const char *name, int idx) {
```
radare: We have R_IPI for the internal apis

libr/bin/Makefile
```diff
@@ -27,7 +27,6 @@ OBJS+=mangling/demangler.o
 OBJS+=mangling/microsoft_demangle.o
 OBJS+=mangling/objc.o mangling/cxx.o mangling/msvc.o
 OBJS+=mangling/swift.o mangling/swift-sd.o
-OBJS+=mangling/dlang.o
```
radare: Why?

libr/bin/bfile.c
```diff
@@ -1,6 +1,7 @@
 /* radare2 - LGPL - Copyright 2009-2018 - pancake, nibble, dso */
 
 #include <r_bin.h>
+#include <bin_private.h>
```
radare: Maybe we should have an i/ directory to put the private/internal includes. The bin_ prefiz is implicit so we can use a different name for this

libr/bin/bin_private.h
```diff
@@ -0,0 +1,47 @@
+#ifndef _BIN_PRIVATE_H_
```
radare: Dont start defines with lowerdash

libr/bin/mangling/cxx.c
```diff
@@ -1,6 +1,7 @@
 /* radare - LGPL - Copyright 2013-2018 - pancake */
 
 #include <r_bin.h>
+#include <bin_private.h>
```
radare: We can use ./ like the line below and remove the -I.

libr/bin/mangling/dlang.c
```diff
@@ -1,46 +0,0 @@
-/* radare - LGPL - Copyright 2015 - pancake */
-
```
radare: Dead code after my refactor, right?

libr/bin/Makefile
ret2libc: It seems it's not used anywhere... maybe it was just removed in one of the recent related commit?

libr/bin/mangling/dlang.c
ret2libc: i guess so yeah

libr/bin/mangling/cxx.c
ret2libc: Yes we could. I did this just to avoid having different "../../" in different parts of the libr/bin codebase (e.g. sometimes it's used in libr/bin/bin.c, others in libr/bin/d/, etc.)

libr/bin/bfile.c
ret2libc: Do you mean in `libr/bin/i`?

libr/bin/bfile.c
ret2libc: Mmm I see R_IPI is re-defined in some files... is it safe to use?

radare:
Yes

libr/bin/Makefile
radare: yeah its fine to kill i t, its now in blang.c

libr/bin/bfile.c
radare: yes

libr/bin/mangling/cxx.c
radare: do that then

ret2libc:
@radare i've done what asked. Is it ok to skip the R_IPI thing? what value does it add?

radare:
Adds no value its just to ease grepping

ret2libc:
> Adds no value its just to ease grepping

Done, I think this is ready. Let's see if it compiles everywhere. I had to fix a bit the R_IPI handling because it used almost nowhere and in weird ways... now R_IPI is defined only one time in r_types.h

radare:
Lgtm

PR #12039, opened 2018-11-01
Improve array indexing in disassembly ##anal

sivaramaaa:
* Before

before

* After

after


ret2libc:
That's great @sivaramaaa ! Thanks! I didn't look at the code yet, but I've noticed that it does work only when the string is indexed through a register.

Something like `lea rax, [rbp - 0x70]` or `movzx eax, byte [rbp - 0x6f]` is not interpreted as `dest + 3` (for example). Is this intended? Do you want me to fill a separate issue?

ret2libc:
@radare btw I've noticed some time ago that RParse is becoming a mess... :(

sivaramaaa:
> Something like lea rax, [rbp - 0x70] or movzx eax, byte [rbp - 0x6f] is not interpreted as dest + 3 , Do you want me to fill a separate issue?

yup it would be very helpful if u can fill an issue !

libr/core/disasm.c
```diff
@@ -911,6 +911,17 @@ static void ds_highlight_word(RDisasmState * ds, char *word, char *color) {
 	ds->opstr = asm_str? asm_str: source;
 }
 
+static char *get_op_ireg (void *user, ut64 addr) {
+	RCore *core = (RCore *)user;
+	char *res = NULL;
+	RAnalOp *op = r_core_anal_op (core, addr, R_ANAL_OP_MASK_ESIL);
+	if (op && op->ireg) {
+		res = strdup(op->ireg);
```
XVilka: Space before `(`.

PR #12043, opened 2018-11-02
fix shift-arrows for st and rxvt-unicode terms

gordon-quad:
Since radare2 does not use termcap to process special keys, running it on different terminal emulators might break some key shortcuts.

This PR fixes shift+arrow shortcuts for st and rxvt-unicode terminal emulators.

### DIFF
--- a/libr/cons/input.c
+++ b/libr/cons/input.c
@@ -156,6 +156,23 @@ R_API int r_cons_arrow_to_hjkl(int ch) {
 		case '1':
 			ch = r_cons_readchar ();
 			switch (ch) {
+			// Support st/st-256color term and others
+			// for shift+arrows
+			case ';': // arrow+mod
+				ch = r_cons_readchar ();
+				switch (ch) {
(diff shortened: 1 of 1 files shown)


gordon-quad:
I am using Colemak keyboard layout on a daily basis and it is not generally a problem with radare2 because all its shortcuts are mnemonic, all except hjkl keys. I found no easy way of changing visual keybindings so the only option for me is to use arrow keys as an alternative. Unfortunately block selection commands did not work with shift-arrow shortcuts on two differnet terminal emulators ([st](https://st.suckless.org/) and [rxvt-unicode](http://software.schmorp.de/pkg/rxvt-unicode.html)) that I use, so I fixed input r_cons_arrow_to_hjkl to accept terminal codes that my terminal emulators produce. I quite unfamiliar with radare2 code base so if you think it is not right place or way to fix described problem please feel free to point that out, also if you think that it require some regression testing, I kindly ask you to point me to an appropriate place where I can add it.

PR #12049, opened 2018-11-04
capstone.sh: Fix shellcheck warning

lzutao:
- Use $( ... ) instead of legacy \` ... \`
- Use git rev-parse instead of grepping and parsing git branch
- Prefer [ p ] && [ q ] as [ p -a q ] is not well defined
- Clone until our needed TIP commit

### DIFF
--- a/shlr/capstone.sh
+++ b/shlr/capstone.sh
@@ -1,4 +1,4 @@
-#! /bin/sh
+#!/bin/sh
 CS_URL="$1" # url
 CS_BRA="$2" # branch name
 CS_TIP="$3" # tip commit
@@ -6,13 +6,13 @@ CS_REV="$4" # revert
 CS_DEPTH_CLONE=10
 
 fatal_msg() {
-	>&2 printf '[capstone] %s\n' "$1"
+	printf '[capstone] %s\n' "$1" >&2
 	exit 1
 }
 
 patch_capstone() {
 	for patchfile in ../capstone-patches/*.patch ; do
-		yes n | patch --reverse --strip 1 --input "${patchfile}"
+		yes n | patch -R -p 1 -i "${patchfile}"
 	done
 }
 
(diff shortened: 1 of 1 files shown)


shlr/capstone.sh
```diff
@@ -1,58 +1,84 @@
-#!/bin/sh
+#! /bin/sh
```
XVilka: why space is here?

lzutao: Space is [allow](https://unix.stackexchange.com/questions/276751/is-space-allowed-between-and-bin-bash-in-shebang) after `#!`. But if you prefer not having space after `#!`, I will change it.

radare:
Never seen that space before anywhere. Kill it

shlr/capstone.sh
```diff
@@ -1,58 +1,84 @@
...
-		HEAD="`git log|head -n1 | awk '{print $2}'`"
-	fi
-	BRANCH="`git branch | grep '*' | cut -d'*' -f2 | cut -d' ' -f2`"
-
-	if [ "${HEAD}" = "${CS_TIP}" ]; then
-		echo "[capstone] Already in TIP, no need to update from git"
-		exit 0
-	fi
+fatal_msg() {
+	>&2 printf '[capstone] %s\n' "$1"
```
radare: Dafuq?

shlr/capstone.sh
```diff
@@ -1,58 +1,84 @@
...
 }
 
-if [ -d capstone -a ! -d capstone/.git ]; then
-	echo "[capstone] release with no git?"
-	cd capstone
-	for PATCH in ../capstone-patches/* ; do
-		yes n | patch -Rp1 < $PATCH
+patch_capstone() {
+	for patchfile in ../capstone-patches/*.patch ; do
+		yes n | patch --reverse --strip 1 --input "${patchfile}"
```
radare: That line is full of gnuisms. It will break in all the bsds and macs

lzutao: :+1:  forgot that, will fix.

shlr/capstone.sh
lzutao: ? See [this](https://stackoverflow.com/a/23550347/5456794). But it's just a matter of style, so I'll change it as you want.

PR #12059, opened 2018-11-05
BSD code cleanups.

devnexen:
Dead code/proper debug macros types.

### DIFF
--- a/libr/io/p/io_ptrace.c
+++ b/libr/io/p/io_ptrace.c
@@ -43,8 +43,13 @@ static int __waitpid(int pid) {
 	return (waitpid (pid, &st, 0) != -1);
 }
 
+#if defined(__GLIBC__) && defined(__linux__)
 #define debug_read_raw(io,x,y) r_io_ptrace((io), PTRACE_PEEKTEXT, (x), (void *)(y), NULL)
 #define debug_write_raw(io,x,y,z) r_io_ptrace((io), PTRACE_POKEDATA, (x), (void *)(y), (void *)(z))
+#else
+#define debug_read_raw(io,x,y) r_io_ptrace((io), PTRACE_PEEKTEXT, (x), (void *)(y), 0)
+#define debug_write_raw(io,x,y,z) r_io_ptrace((io), PTRACE_POKEDATA, (x), (void *)(y), (int)(z))
+#endif
 #if __OpenBSD__ || __KFBSD__
 typedef int ptrace_word;   // int ptrace(int request, pid_t pid, caddr_t addr, int data);
 #else
--- a/libr/core/cfile.c
+++ b/libr/core/cfile.c
@@ -265,6 +265,7 @@ R_API int r_core_bin_reload(RCore *r, const char *file, ut64 baseaddr) {
 	return result;
 }
 
+#if __linux__ || __APPLE__
 static bool setbpint(RCore *r, const char *mode, const char *sym) {
 	RBreakpointItem *bp;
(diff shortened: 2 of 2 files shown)


libr/io/p/io_ptrace.c
```diff
@@ -43,8 +43,13 @@ static int __waitpid(int pid) {
 	return (waitpid (pid, &st, 0) != -1);
 }
 
+#if defined(__GLIBC__) && defined(__linux__)
 #define debug_read_raw(io,x,y) r_io_ptrace((io), PTRACE_PEEKTEXT, (x), (void *)(y), NULL)
 #define debug_write_raw(io,x,y,z) r_io_ptrace((io), PTRACE_POKEDATA, (x), (void *)(y), (void *)(z))
+#else
+#define debug_read_raw(io,x,y) r_io_ptrace((io), PTRACE_PEEKTEXT, (x), (void *)(y), 0)
+#define debug_write_raw(io,x,y,z) r_io_ptrace((io), PTRACE_POKEDATA, (x), (void *)(y), (int)(z))
```
ret2libc: what about always using `r_ptrace_data_t`?

libr/io/p/io_ptrace.c
```diff
@@ -43,8 +43,13 @@ static int __waitpid(int pid) {
 	return (waitpid (pid, &st, 0) != -1);
 }
 
+#if defined(__GLIBC__) && defined(__linux__)
 #define debug_read_raw(io,x,y) r_io_ptrace((io), PTRACE_PEEKTEXT, (x), (void *)(y), NULL)
 #define debug_write_raw(io,x,y,z) r_io_ptrace((io), PTRACE_POKEDATA, (x), (void *)(y), (void *)(z))
+#else
+#define debug_read_raw(io,x,y) r_io_ptrace((io), PTRACE_PEEKTEXT, (x), (void *)(y), 0)
```
ret2libc: is there no `NULL` on BSD?

libr/io/p/io_ptrace.c
devnexen: Sure ... the outcome will be the same just little code reduction.

libr/io/p/io_ptrace.c
devnexen: There is but as you pointed out there is r_ptrace_data_t and they have different underlying types Linux and non Linux for a reason.

ret2libc: I see. Then you could use the already existing `#if defined...` in r_io.h and do something like:
```C
#if defined(__GLIBC__) && defined(__linux__)
typedef enum __ptrace_request r_ptrace_request_t;
typedef void * r_ptrace_data_t;
#define R_PTRACE_NODATA NULL
#else
typedef int r_ptrace_request_t;
typedef int r_ptrace_data_t;
#define R_PTRACE_NODATA 0
#endif
```
```C
#define debug_read_raw(io,x,y) r_io_ptrace((io), PTRACE_PEEKTEXT, (x), (void *)(y), R_PTRACE_NODATA)
```
and avoid all together the new `#if defined...` you are introducing in this PR. This way things will be more "declarative" and there will be less `#if..` to follow.

devnexen: Ok did a PR for this little code simplification.

ret2libc: (ops I didn't see the PR was already merged...)

devnexen: That s ok there is no functional change nor bug introduction ;-)

PR #12062, opened 2018-11-05
Change argument type to ut8 before passing to ctype.h function

lzutao:
The behavior of the `<ctype.h>` functions is undefined for negative
arguments (other than `EOF`). In such a situation, the argument should
be cast to `unsiged char` for safety.

References:
 - C Programming: A Modern Approach, 2nd Edition: page 612, chapter 23.5


XVilka:
Everything is fine, but please do not touch mpc (or send PR to it instead, and update r2 version), since it is an external project: https://github.com/orangeduck/mpc/

PR #12063, opened 2018-11-05
Use ht to query strings for the cfstrings ##bin

radare:
0m47s vs 4m56s

### DIFF
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -78,6 +78,9 @@ static int string_scan_range(RList *list, RBinFile *bf, int min,
 	int count = 0, i, rc, runes;
 	int str_type = R_STRING_TYPE_DETECT;
 
+	// if list is null it means its gonna dump
+	r_return_val_if_fail (bf, -1);
+
 	if (type == -1) {
 		type = R_STRING_TYPE_DETECT;
 	}
@@ -91,10 +94,12 @@ static int string_scan_range(RList *list, RBinFile *bf, int min,
 		free (buf);
 		return -1;
 	}
-	RBinSection *s = r_bin_get_section_at (bf->o, from, false);
 	st64 vdelta = 0;
-	if (s) {
-		vdelta = s->vaddr - from;
+	if (bf->o) {
+		RBinSection *s = r_bin_get_section_at (bf->o, from, false);
+		if (s) {
+			vdelta = s->vaddr - from;
+		}
 	}
 	r_buf_read_at (bf->buf, from, buf, len);
 	// may oobread
@@ -243,7 +248,9 @@ static int string_scan_range(RList *list, RBinFile *bf, int min,
 			bs->string = r_str_ndup ((const char *)tmp, i);
 			if (list) {
 				r_list_append (list, bs);
-				ht_insert (bf->o->strings_db, K(bs->vaddr), bs);
+				if (bf->o) {
+					ht_insert (bf->o->strings_db, K(bs->vaddr), bs);
+				}
 			} else {
 				print_string (bf, bs);
 				r_bin_string_free (bs);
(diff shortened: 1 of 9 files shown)


radare:
Changing from dict to the new insert64 would be easier than switching to sdb_fmt, and i guess that sdb api change will come soon.

radare:
When this new api is done i think we can kill dict completely, as long as i have an api to do k=v with ut64 without having to mess with strings im fine. The implementation is not that problematic for me right now. Also we need to think in having a big k=v in each RBinObject to make indexable all the data loaded by rbin

libr/bin/bfile.c
```diff
@@ -885,6 +878,24 @@ static int is_data_section(RBinFile *a, RBinSection *s) {
 	return strstr (s->name, "_const") != NULL;
 }
 
+// XXX : This is O(n) for strings, we need to query strings by offset in the rbinobj directly
+static RBinString *find_string_at (RBinObject *bo, RList *ret, ut64 addr) {
+	if (addr != 0 && addr != UT64_MAX) {
+#if 1
```
ret2libc: remember to clean this before merging (after travis is green :)

radare: yep. its there for benchmarking reasons :)

radare:
cause yes?

libr/bin/bfile.c
radare: i will kill this ifdef when this pr gets green

radare:
should be green now

libr/bin/bfile.c
```diff
@@ -7,12 +7,21 @@
 #define R_STRING_SCAN_BUFFER_SIZE 2048
 #define R_STRING_MAX_UNI_BLOCKS 4
 
-static void print_string(RBinString *string, RBinFile *bf) {
-	if (!string || !bf) {
-		return;
+#define K(x) sdb_fmt("%"PFMT64x, x)
+
+static RBinString *find_string_at (RBinFile *bf, RList *ret, ut64 addr) {
+	if (addr != 0 && addr != UT64_MAX) {
```
ret2libc: why this check? i think you can just avoid it and look it up in the hashtable... if there is a value for those keys, we should get them, otherwise no.

libr/bin/bfile.c
```diff
@@ -794,34 +799,33 @@ R_IPI RList *r_bin_file_get_strings(RBinFile *a, int min, int dump, int raw) {
 				if (section->size > a->size) {
 					continue;
 				}
+				ut8 *sbuf = malloc (section->size);
+				if (!sbuf) {
+					continue;
+				}
+				r_buf_read_at (a->buf, section->paddr + cfstr_offs, sbuf, section->size);
```
ret2libc: why reading the entire section instead of doing what was done before?

libr/bin/obj.c
```diff
@@ -148,6 +149,7 @@ R_IPI RBinObject *r_bin_object_new(RBinFile *binfile, RBinPlugin *plugin, ut64 b
 	}
 	o->obj_size = bytes && (bytes_sz >= sz + offset)? sz: 0;
 	o->boffset = offset;
+	o->strings_db = ht_new (NULL, NULL, NULL);
```
ret2libc: nope! please provide a `free_kv` function that does this:
```C
static void free_kv(HtKv *kv) {
     free (kv->key);
}
```

by default the ht strdups the key, so you need to free it. we should probably have a default for this, i'm going to work on it but in the meantime please implement it. there are other places where this is already done and i will replace them once i'll change things in sdb.

libr/bin/bfile.c
```diff
@@ -82,6 +91,11 @@ static int string_scan_range(RList *list, RBinFile *bf, int min,
 		free (buf);
 		return -1;
 	}
+	RBinSection *s = r_bin_get_section_at (bf->o, from, false);
```
ret2libc: why `false` for the `va` parameter? it seems you are using the `s->vaddr` in the next lines, so you are probably dealing with virtual addresses...

libr/bin/bfile.c
radare: to speedup when cfstring is invaild or filled with zeros or ffffs. in theory none of those addresses should be available in the hashtable. but yeah we can skip that unnecessary check that will be matching in the 99.9% of cases

libr/bin/bfile.c
radare: its faster

libr/bin/bfile.c
radare: nope those are phisical addresses. some string sections are not virtually mapped.

libr/bin/obj.c
radare: done

libr/bin/bfile.c
ret2libc: But you are using s->vaddr few lines below this...

libr/bin/bfile.c
```diff
@@ -82,6 +94,13 @@ static int string_scan_range(RList *list, RBinFile *bf, int min,
 		free (buf);
 		return -1;
 	}
+	st64 vdelta = 0;
+	if (bf->o) {
```
ret2libc: Is there any valid reason for a binfile to not have a binobject? It seems a r-return-Val at the beginning of the function would be better

radare:
I think this kind of issues are related to thw problwm of strings not working in fat binaries. I have opened an issue with a reproducer. If i change this to true there are 28 XX

radare:
What i have noticed is that binfiles are created even when running with r2 -n. But in this case the ->o is null. Which has been spotted by a bunch of asserts i added in this pr.

We will have to look at this more carefully but in separate pr. This is getting addictive

libr/bin/bfile.c
ret2libc: yes but it could also be a loooot of data (think about a data section with many many resources inside..)

radare:
Yes it can be big but the loop will process it complrtelly so ideally we should nest another loop in here to read with big chunks but its fine right now to be 64MB

PR #12064, opened 2018-11-05
WIP: Add callbacks API in RFlag, for future optimizations ##flags

radare:
The idea behind this new feature of RFlag is to allow RCore to attach a callback that uses RBin to resolve the name of an offset and the offset for a name). This way we can save a lot of time when loading the binary because we wont need to create any flag.

To do this we have to switch from RList to another data structure (or add companion hashtables) to allow such queries by filtering with strncmp (flagname, "sym.imp.", 8); ...

### DIFF
--- a/libr/bin/bflag.c
+++ b/libr/bin/bflag.c
@@ -1,0 +1,34 @@
+/* radare2 - LGPL - Copyright 2018 - pancake, nibble, dso */
+
+
+#include <r_bin.h>
+
+// primitives for r_flag_set_callbacks()
+
+R_API const char *r_bin_flag_i(RBin *bin, ut64 addr) {
+	r_return_val_if_fail (bin, NULL);
+	if (bin->cur && bin->cur->o) {
+		RListIter *iter;
+		RBinSymbol *s;
+		r_list_foreach (bin->cur->o->symbols, iter, s) {
+			if (s->vaddr == addr) {
+				return s->name;
+			}
+		}
+	}
+	return NULL;
+}
+
+R_API ut64 r_bin_flag(RBin *bin, const char *name) {
+	// TODO: we should iterate over all the BinFiles
+	if (!strncmp (name, "sym.", 4)) {
+		RListIter *iter;
+		RBinSymbol *s;
+		r_list_foreach (bin->cur->o->symbols, iter, s) {
+			if (!strcmp (name + 4, s->name)) {
+				return s->vaddr;
+			}
+		}
+	}
+	return UT64_MAX;
+}
--- a/libr/flag/item.c
+++ b/libr/flag/item.c
@@ -2,8 +2,6 @@
 
 #include <r_flag.h>
 
-#define ISNULLSTR(x) (!(x) || !*(x))
-
 R_API RFlagItem *r_flag_item_new (void) {
 	return R_NEW0 (RFlagItem);
 }
@@ -45,14 +43,14 @@ R_API void r_flag_item_free(RFlagItem *item) {
 R_API void r_flag_item_set_alias(RFlagItem *item, const char *alias) {
 	r_return_if_fail (item);
 	free (item->alias);
-	item->alias = ISNULLSTR (alias)? NULL: strdup (alias);
+	item->alias = R_STR_ISEMPTY (alias)? NULL: strdup (alias);
 }
 
 /* add/replace/remove the comment of a flag item */
 R_API void r_flag_item_set_comment(RFlagItem *item, const char *comment) {
 	r_return_if_fail (item);
 	free (item->comment);
-	item->comment = ISNULLSTR (comment) ? NULL : strdup (comment);
+	item->comment = R_STR_ISEMPTY (comment) ? NULL : strdup (comment);
 }
 
 /* add/replace/remove the realname of a flag item */
@@ -61,5 +59,5 @@ R_API void r_flag_item_set_realname(RFlagItem *item, const char *realname) {
 	if (item->name != item->realname) {
 		free (item->realname);
 	}
-	item->realname = ISNULLSTR (realname) ? NULL : strdup (realname);
+	item->realname = R_STR_ISEMPTY (realname) ? NULL : strdup (realname);
 }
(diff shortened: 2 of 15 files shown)


ret2libc:
Are you sure the time is really spent creating flags? I don't see that many improvements in time (maybe in terms of memory) by using this thing. Instead of querying RFlags you'll end up querying RBin. Now, about the time spent... it would all depends on how we implement these lookup functions. I'm not that sure it will be faster or better. Personally, I'd like to see some tests/benchmarks about this before starting going down this road.

radare:
Yes, from a random bin im using for testing, which takes 50s to load, 30s of them is spent creating flags from the bin info. which is stupid. ebcause this info must be get rid when the binobj is unloaded or rebased.

This proposal aims to define the base to allow rflags query rbin to resolve addr from name and name(s) from addr.

And yes. it is important to. have a fast and well defined way to query this info to avoid more bottlenecks in the future. i have proposed many solutions to this, but as long as i get no. feedback i preferred to write a PR to have something to discuss and test.

The last commit in the pr adds a bin.flags option to change the behaviour, so it should be green and shouldn't be breaking anything by default for now

ret2libc:
I still don't like it, sorry. We should understand why RFlag is so slow in insertion (it may be because on every r_flag_set we call r_flag_get_list). Moreover, this would bring a lot of complexity IMHO. If we use flags, it's because they bring some value. In the moment you don't use them anymore, you have to mess with RBin and RBin would have to replicate some of the things currently done by RFlag. 

Moreover, you have to join results from both RBin and RFlag, you can't delete "bin related" flags anymore, etc. etc. We will create a lot of "if (flag is bin related ) do this else do that"...

Can you share the binary so I can also see where exactly it's the bottleneck in RFlag?

radare:
Rflag is not slow on insertion. its just an unnecessary thing that we shouldnt do because this information doesnt belongs to rflag, because it have different properties, and main difference is ownership. i don't want to list those flags, because they are grouped in.

and yes. i wrote code merging the list returned from rbin_flag and rflag itself, but the problem here is ownership of. the returned list. so it will leak or make the whole thing a copy which is unnecessarily slow., but i have to find a solution. to this before merging this pr

libr/bin/bflag.c
```diff
@@ -0,0 +1,37 @@
...
+#include <r_bin.h>
+
+// primitives for r_flag_set_callbacks()
+
+R_API const char *r_bin_flag_i(RBin *bin, ut64 addr) {
+	r_return_val_if_fail (bin, NULL);
+	if (bin->cur && bin->cur->o) {
+		RListIter *iter;
+		RBinSymbol *s;
+		r_list_foreach (bin->cur->o->symbols, iter, s) {
```
radare: this should be O(1) using the generic k/v storage inside every rbinobj

libr/bin/bflag.c
```diff
@@ -0,0 +1,37 @@
...
+	}
+	return NULL;
+}
+
+R_API ut64 r_bin_flag(RBin *bin, const char *name) {
+	// TODO: we should iterate over all the BinFiles
+	if (!strncmp (name, "sym.", 4)) {
+		RListIter *iter;
+		RBinSymbol *s;
+		r_list_foreach (bin->cur->o->symbols, iter, s) {
```
radare: same goes here

radare:
my other solution to this problem was to have another RFlag inside each RBinObject, and just delegate everything as recursvie calls from core->flags into core->bin->obj[]->flags

libr/flag/item.c
```diff
@@ -0,0 +1,65 @@
+/* radare - LGPL - Copyright 2007-2018 - pancake */
+
+#include <r_flag.h>
+
+#define ISNULLSTR(x) (!(x) || !*(x))
```
ret2libc: just use `R_STR_ISEMPTY` from r_str.h

ret2libc:
Can we close this? Apart from the initial concerns we discussed before, I think flag code is much different now, so it's probably worth checking again with current code base to see if time is still a problem.

radare:
Afaik your changes made rflags slower to insert , so loading times would be even worst than before for that case. But i didnt had time to test too much and this pr was just a poc, and it was just for testing the theory that r2 will load much faster if we just use rbin to resolve addr name instead of duplicating the info in flags

radare:
Totally broken at this point but the idea remains for me. Any comments @thestr4ng3r ?

thestr4ng3r:
I do like the idea of making flags partially callback based, not necessarily only because of loading times (this has to be profiled carefully again of course) but also because of memory usage. Currently we at least double the memory used by symbol names and sometimes even more from realnames and demangling.

I didn't read the diff at all, so no idea how it would be done there but this would be my idea to implement it:
Have RFlag being the object that overall manages all the flags and provides interfaces to query them.
Have multiple RFlagStorage objects that contain the callbacks to actually query any flags. This would be like a typical interface in OOP.
RBin implements one RFlagStorage, RAnal also implements one for functions and there can be another one that implements user-editable flags, which would basically contain the implementation of RFlag as we have it right now.
But I think this should be discussed in a dedicated issue, not this pr.

radare:
The overall problem
With this prs is that flags handled via callbacks can be processed with the coverage/range things

radare:
I like your proposal, i also agree that its also a memory usage problem. I didnt measured it but it should be a big difference too.

The way my code is implemented just adds one callback that is initialized by core it doesnt supports “plugins” to attach more than one flag backend. Which works as a poc and we may evaluate the possibility to introduce this at least as an option at the beginning

ret2libc:
Shall we close this?

ret2libc:
Closing. This had no review, no changes, no nothing for a long time. I'm closing the PR as there is no benefit in keeping this open. The branch will still be active, so if @trufae wants he can continue working on this when he has time. No need to keep it in the backlog.

trufae:
This is more like an IDEA than a PR because it breaks the concept of flags by adding fake entities managed by callbacks, it results in faster times, and i think we end up agreeing on that have no sense and we may improve rbin apis instead of making rflag inconsistent.

I dont plan to continue working on this PR but i wanted to have some broad discussion, and IDEAS is probably a better place for this right now

PR #12071, opened 2018-11-06
remove other RBin unused functions and make others static

ret2libc:
It also introduces a WARNING when plugins do not implement the load_buffer method

### DIFF
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -312,8 +312,8 @@ R_API bool r_bin_file_object_new_from_xtr_data(RBin *bin, RBinFile *bf, ut64 bas
 		plugin = r_bin_get_binplugin_any (bin);
 	}
 	r_buf_free (bf->buf);
-	bf->buf = r_buf_new_with_bytes ((const ut8*)bytes, data->size);
-	//r_bin_object_new append the new object into binfile
+	bf->buf = r_buf_new_with_bytes ((const ut8 *)bytes, data->size);
+	// r_bin_object_new append the new object into binfile
 	o = r_bin_object_new (bf, plugin, baseaddr, loadaddr, offset, sz);
 	// size is set here because the reported size of the object depends on
 	// if loaded from xtr plugin or partially read
@@ -457,15 +457,6 @@ R_IPI RBinFile *r_bin_file_find_by_id(RBin *bin, ut32 binfile_id) {
 		binfile = NULL;
 	}
 	return binfile;
-}
-
-R_IPI int r_bin_file_object_add(RBinFile *binfile, RBinObject *o) {
-	if (!o) {
-		return false;
-	}
-	r_list_append (binfile->objs, o);
-	r_bin_file_set_cur_binfile_obj (binfile->rbin, binfile, o);
-	return true;
 }
 
 R_API int r_bin_file_delete_all(RBin *bin) {
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -260,54 +260,6 @@ static char *swiftField(const char *dn, const char *cn) {
 	return NULL;
 }
 
-R_IPI RList *r_bin_classes_from_symbols(RBinFile *bf, RBinObject *o) {
-	RBinSymbol *sym;
-	RListIter *iter;
-	RList *symbols = o->symbols;
-	RList *classes = o->classes;
-	if (!classes) {
-		classes = r_list_newf ((RListFree)r_bin_class_free);
-	}
-	r_list_foreach (symbols, iter, sym) {
-		if (sym->name[0] != '_') {
-			continue;
-		}
-		const char *cn = sym->classname;
-		if (cn) {
-			RBinClass *c = r_bin_class_new (bf, sym->classname, NULL, 0);
-			if (!c) {
-				continue;
-			}
-			// swift specific
-			char *dn = sym->dname;
-			char *fn = swiftField (dn, cn);
-			if (fn) {
-				// eprintf ("FIELD %s  %s\n", cn, fn);
-				RBinField *f = r_bin_field_new (sym->paddr, sym->vaddr, sym->size, fn, NULL, NULL);
-				r_list_append (c->fields, f);
-				free (fn);
-			} else {
-				char *mn = strstr (dn, "..");
-				if (mn) {
-					// eprintf ("META %s  %s\n", sym->classname, mn);
-				} else {
-					char *mn = strstr (dn, cn);
-					if (mn && mn[strlen(cn)] == '.') {
-						mn += strlen (cn) + 1;
-						// eprintf ("METHOD %s  %s\n", sym->classname, mn);
-						r_list_append (c->methods, sym);
-					}
-				}
-			}
-		}
-	}
-	if (r_list_empty (classes)) {
-		r_list_free (classes);
-		return NULL;
-	}
-	return classes;
-}
-
 R_IPI RBinFile *r_bin_file_new(RBin *bin, const char *file, const ut8 *bytes, ut64 sz, ut64 file_sz, int rawstr, int fd, const char *xtrname, Sdb *sdb, bool steal_ptr) {
 	RBinFile *binfile = R_NEW0 (RBinFile);
 	if (!binfile) {
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -411,6 +411,15 @@ R_API bool r_bin_file_object_new_from_xtr_data(RBin *bin, RBinFile *bf, ut64 bas
 	return true;
 }
 
+static RBinFile *file_create_append(RBin *bin, const char *file, const ut8 *bytes, ut64 sz, ut64 file_sz, int rawstr, int fd, const char *xtrname, bool steal_ptr) {
+	RBinFile *bf = r_bin_file_new (bin, file, bytes, sz, file_sz, rawstr,
+		fd, xtrname, bin->sdb, steal_ptr);
+	if (bf) {
+		r_list_append (bin->binfiles, bf);
+	}
+	return bf;
+}
+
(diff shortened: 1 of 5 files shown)


libr/bin/obj.c
```diff
@@ -8,17 +8,135 @@
...
+					}
+				}
+			}
+		}
+	}
+	if (r_list_empty (classes)) {
+		r_list_free (classes);
+		return NULL;
+	}
+	return classes;
```
radare: Maybe we need to fill o->classes directly here instead of returning a list, because this function is appending them and returning which is conceptually wrong and may lead to dupped data

libr/bin/bfile.c
```diff
@@ -885,6 +744,61 @@ static int is_data_section(RBinFile *a, RBinSection *s) {
 	return strstr (s->name, "_const") != NULL;
 }
 
+static void get_strings_range(RBinFile *bf, RList *list, int min, int raw, ut64 from, ut64 to) {
+	RBinPlugin *plugin = r_bin_file_cur_plugin (bf);
+	RBinString *ptr;
+	RListIter *it;
+
+	if (!bf || !bf->buf) {
```
radare: R return if fail this

libr/bin/bfile.c
```diff
@@ -885,6 +744,61 @@ static int is_data_section(RBinFile *a, RBinSection *s) {
 	return strstr (s->name, "_const") != NULL;
 }
 
+static void get_strings_range(RBinFile *bf, RList *list, int min, int raw, ut64 from, ut64 to) {
+	RBinPlugin *plugin = r_bin_file_cur_plugin (bf);
```
radare: Do this after checking for null

libr/bin/bin.c
```diff
@@ -1285,6 +1285,20 @@ R_IPI void r_bin_class_free(RBinClass *c) {
 	free (c);
 }
 
+static RBinClass *class_get(RBinFile *binfile, const char *name) {
+	r_return_val_if_fail (binfile && binfile->o && name, NULL);
+
+	RBinClass *c;
+	RListIter *iter;
+	RList *list = binfile->o->classes;
```
radare: Unnecessary variable

libr/bin/bin.c
```diff
@@ -1285,6 +1285,20 @@ R_IPI void r_bin_class_free(RBinClass *c) {
...
+	r_list_foreach (list, iter, c) {
```
radare: Add a warning message because this is o(n) and we shohld use a kv to resolve this faster

libr/bin/bfile.c
```diff
@@ -953,67 +867,12 @@ R_IPI RList *r_bin_file_get_strings(RBinFile *a, int min, int dump, int raw) {
 		}
 	} else {
 		if (a) {
-			r_bin_file_get_strings_range (a, ret, min, raw, 0, a->size);
```
radare: Ideally this functionality should be implemented in rsearch. But this change doesnt fits in this pr. Just keep it in mind for future changes

libr/bin/obj.c
```diff
@@ -8,17 +8,135 @@
 	if (binfile->rbin->verbose) \
 	eprintf
 
+static void mem_free(void *data) {
+	RBinMem *mem = (RBinMem *)data;
+	if (mem && mem->mirrors) {
+		mem->mirrors->free = mem_free;
+		r_list_free (mem->mirrors);
+		mem->mirrors = NULL;
```
radare: No need to null

libr/bin/obj.c
```diff
@@ -8,17 +8,135 @@
...
+	}
+	free (mem);
+}
+
+static void object_delete_items(RBinObject *o) {
```
radare: This is called only once from objfree. What about moving all this code in there and remove the function?

libr/bin/bfile.c
```diff
@@ -885,6 +744,61 @@ static int is_data_section(RBinFile *a, RBinSection *s) {
...
+		return;
+	}
+	if (raw != 2) {
+		ut64 size = to - from;
+		// in case of dump ignore here
+		if (bf->rbin->maxstrbuf && size && size > bf->rbin->maxstrbuf) {
+			if (bf->rbin->verbose) {
+				eprintf ("WARNING: bin_strings buffer is too big "
+					"(0x%08" PFMT64x
+					")."
```
radare: Ugly indent

libr/bin/bfile.c
ret2libc: ok good to know

libr/bin/obj.c
ret2libc: Are you sure? I'm just keeping the code as it was, I didn't really looked at it, just moved the function from one place to another.

libr/bin/obj.c
ret2libc: I'm actually ok having a separate function. I tink it makes things simpler to read.

libr/bin/obj.c
ret2libc: actually... it seems like classes will be always NULL :S I don't see any r_list_ on the `classes` variable in `classes_from_symbols`. Do you see where this is filled? It seems all this thing with classes is useless :/ I'd just keep the r_list_foreach in the body and remove the rest...

ret2libc:
Ok, I think this is enough for 1 PR. Let's see what travis says and if you @radare have other comments.

Some things to note:
- i've added a r_assert to be sure that "any" plugin is always available
- implemented load_buffer in the "any" plugin, by creating an empty r_buf that is never used

libr/bin/obj.c
radare: ok

libr/bin/obj.c
```diff
@@ -8,21 +8,148 @@
...
+			if (r) {
+				*r = 0;
+			}
+			return q;
+		}
+	}
+	return NULL;
+}
+
+static RList *classes_from_symbols(RBinFile *bf, RBinObject *o) {
```
radare: Just pass RBinFile, and dereference the RBinObject from the bf.

at some point we shuold merge those two structs to make it less confusing

libr/bin/obj.c
```diff
@@ -8,21 +8,148 @@
...
+	RBinSymbol *sym;
+	RListIter *iter;
+	RList *symbols = o->symbols;
+	RList *classes = o->classes;
```
radare: what about not taking this and always create a new list?, the one who calls this function should be in charge to fill o->classes?

For consistency maybe it should be like this,otherwise the function name may change. If we decide to just updae the binobj inside each of those calls instead of returning a list. im fine too. but it can be useful at some point to have access to those apis, not just having them internally.

libr/bin/obj.c
```diff
@@ -8,21 +8,148 @@
...
 	}
 	free (o->regstate);
 	r_bin_info_free (o->info);
-	r_bin_object_delete_items (o);
+	object_delete_items (o);
 	R_FREE (o);
 }
 
+static char *swiftField(const char *dn, const char *cn) {
+	char *p = strstr (dn, ".getter_");
```
radare: dn and cn can be null

libr/bin/obj.c
```diff
@@ -8,21 +8,148 @@
...
+}
+
+static bool file_object_add(RBinFile *binfile, RBinObject *o) {
+	r_return_val_if_fail (binfile && o, false);
+	r_list_append (binfile->objs, o);
+	r_bin_file_set_cur_binfile_obj (binfile->rbin, binfile, o);
+	return true;
+}
+
 R_IPI RBinObject *r_bin_object_new(RBinFile *binfile, RBinPlugin *plugin, ut64 baseaddr, ut64 loadaddr, ut64 offset, ut64 sz) {
```
radare: what do you think about renaming r_bin_object_ into r_bin_obj_ or r_binobj_ ?

libr/bin/bfile.c
```diff
@@ -411,110 +341,48 @@ R_API bool r_bin_file_object_new_from_xtr_data(RBin *bin, RBinFile *bf, ut64 bas
...
 
-R_IPI RBinFile *r_bin_file_new_from_bytes(RBin *bin, const char *file, const ut8 *bytes, ut64 sz, ut64 file_sz, int rawstr, ut64 baseaddr, ut64 loadaddr, int fd, const char *pluginname, const char *xtrname, ut64 offset, bool steal_ptr) {
-	ut8 binfile_created = false;
-	RBinPlugin *plugin = NULL;
-	RBinXtrPlugin *xtr = NULL;
-	RBinObject *o = NULL;
-	if (sz == UT64_MAX) {
-		return NULL;
+static RBinPlugin *get_plugin(RBin *bin, const char *pluginname, const ut8 *bytes, ut64 sz) {
+	RBinPlugin *plugin;
```
radare: defiene and assign in the same line

libr/bin/obj.c
```diff
@@ -8,21 +8,148 @@
...
+	r_list_free (o->strings);
+	r_list_free (o->symbols);
+	r_list_free (o->classes);
+	r_list_free (o->lines);
+	sdb_free (o->kv);
+	if (o->mem) {
+		o->mem->free = mem_free;
+	}
+	r_list_free (o->mem);
+	o->mem = NULL;
```
radare: remove all those = NULL

libr/bin/obj.c
```diff
@@ -8,21 +8,148 @@
...
+	o->libs = NULL;
+	o->relocs = NULL;
+	o->sections = NULL;
+	o->strings = NULL;
+	o->symbols = NULL;
+	o->classes = NULL;
+	o->lines = NULL;
+	o->info = NULL;
+	o->kv = NULL;
+	for (i = 0; i < R_BIN_SYM_LAST; i++) {
```
radare: and this loop too

radare: i mean the = NULL inside the loop

ret2libc:
@radare I think i've addressed all comments. I *think* i've also fixed the `classes_from_symbols` function.

radare:
Rbinfile holds one rbinobj inside. So they are basically the same. I mean you cannot think in them as separate identities. We should always use one of them and imho it shoukd ve rbinfile, even if we wanr to rename it later

ret2libc:
Perfect, that's one more reason to do the renaming later rather than now.
Are you ok merging this?

PR #12072, opened 2018-11-06
Rewrite r_num_abs as function (fix #12060) and r_num_units

### LINKED ISSUE #12060: Multiple evaluation in r_num_abs(x) macro
https://github.com/radare/radare2/blob/865ec593b18b874924cae43f2d46bd2448262095/libr/include/r_util/r_num.h#L5

Macro `r_num_abs(x)` evaluated `x` two times. Consider that if we called  `r_num_abs(x++)`, it would expand to `x++>0?x++:-x++`?

Why not use [`llabs`](https://www.freebsd.org/cgi/man.cgi?query=llabs&sektion=3&apropos=0&manpath=FreeBSD+11.2-RELEASE+and+Ports) function instead?

Note that this macro is only used in `r_io_shift`.

lzutao:
r_num_abs:

- Make r_num_abs static inline
- Use unsigned long long type when defining TB macro
- Use st64 instead of long long

r_num_units:

- Add len argument for buf in r_num_units
- Add PB and EB for r_num_units
- Always display one number after decimal point

  * This change simplifies the code
- Use long double type as assigning from ut64 to double cause data loss

r_num_tail_base:

- Use isxdigit (fast) instead of isHexDigit (slower)
- Rename nth to get_nth_nibble

Update all functions to use with new r_num_units:

- Remove side effect when using in other functions like cb_printf

### DIFF
--- a/libr/include/r_util/r_num.h
+++ b/libr/include/r_util/r_num.h
@@ -2,7 +2,6 @@
 #define R_NUM_H
 
 #define R_NUMCALC_STRSZ 1024
-#define r_num_abs(x) x>0?x:-x
 
 #ifdef __cplusplus
 extern "C" {
@@ -51,7 +50,7 @@ typedef const char *(*RNumCallback2)(struct r_num_t *self, ut64, int *ok);
 
 R_API RNum *r_num_new(RNumCallback cb, RNumCallback2 cb2, void *ptr);
 R_API void r_num_free(RNum *num);
-R_API char *r_num_units(char *buf, ut64 num);
+R_API char *r_num_units(char *buf, size_t len, ut64 number);
 R_API int r_num_conditional(RNum *num, const char *str);
 R_API ut64 r_num_calc(RNum *num, const char *str, const char **err);
 R_API const char *r_num_calc_index(RNum *num, const char *p);
@@ -79,6 +78,10 @@ R_API int r_num_str_len(const char *str);
 R_API int r_num_str_split(char *str);
 R_API RList *r_num_str_split_list(char *str);
 R_API void *r_num_dup(ut64 n);
+
+static inline st64 r_num_abs(st64 num) {
+	return num < 0 ? -num : num;
+}
 
 #ifdef __cplusplus
 }
--- a/libr/util/unum.c
+++ b/libr/util/unum.c
@@ -75,32 +75,43 @@ R_API void r_num_free(RNum *num) {
 	R_FREE (num);
 }
 
-#define KB (1024)
-#define MB (1024*KB)
-#define GB (1024*MB)
-#define TB (1024*GB)
+#define KB (1ULL << 10)
+#define MB (1ULL << 20)
+#define GB (1ULL << 30)
+#define TB (1ULL << 40)
+#define PB (1ULL << 50)
+#define EB (1ULL << 60)
 
-R_API char *r_num_units(char *buf, ut64 num) {
+/**
+ * Convert size in bytes to human-readable string
+ *
+ * Result is stored in buf (buf should be at least 8 bytes in size).
+ * If buf is NULL, memory for the new string is obtained with malloc(3),
+ * and can be freed with free(3).
+ *
+ * On success, returns a pointer to buf. It returns NULL if
+ * insufficient memory was available.
+ */
+R_API char *r_num_units(char *buf, size_t len, ut64 num) {
+	long double fnum;
 	char unit;
-	int tnum;
-	double fnum = num;
 	if (!buf) {
-		buf = malloc (32);
+		buf = malloc (len + 1);
 		if (!buf) {
 			return NULL;
 		}
 	}
-	//if (num>=TB) { unit = 'T'; fnum = num/TB; } else
-	if (num>=GB) { unit = 'G'; fnum = fnum/GB; } else
-	if (num>=MB) { unit = 'M'; fnum = fnum/MB; } else
-	if (num>=KB) { unit = 'K'; fnum = fnum/KB; } else
-		{ unit = 0; fnum = (double)num; }
-	tnum = (int)((double)(fnum - (int)fnum)*10);
-	if (tnum) {
-		snprintf (buf, 31, "%.1f%c", fnum, unit);
-	} else {
-		snprintf (buf, 31, "%.0f%c", fnum, unit);
+	fnum = num;
+	if (num >= EB) { unit = 'E'; fnum /= EB; } else
+	if (num >= PB) { unit = 'P'; fnum /= PB; } else
+	if (num >= TB) { unit = 'T'; fnum /= TB; } else
+	if (num >= GB) { unit = 'G'; fnum /= GB; } else
+	if (num >= MB) { unit = 'M'; fnum /= MB; } else
+	if (num >= KB) { unit = 'K'; fnum /= KB; } else {
+		snprintf (buf, len, "%" PFMT64u, num);
+		return buf;
 	}
+	snprintf (buf, len, "%.1" LDBLFMT "%c", fnum, unit);
 	return buf;
 }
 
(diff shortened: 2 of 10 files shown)


libr/util/unum.c
```diff
@@ -75,32 +75,37 @@ R_API void r_num_free(RNum *num) {
...
+// TODO: Use SI system
+// https://metacpan.org/pod/Number::Bytes::Human
 R_API char *r_num_units(char *buf, ut64 num) {
 	char unit;
 	int tnum;
 	double fnum = num;
-	if (!buf) {
-		buf = malloc (32);
-		if (!buf) {
+#undef R_NUM_UNITS_BUFLEN
```
ret2libc: i don't like too much the `#define` in the middle of the function

ret2libc: yes but since you are also suggesting the "at least 32 bytes in size" in the documentation on top of `r_num_units` it would be better to say "at least R_NUM_UNITS_BUFLEN bytes in size" and it's very likely users of that API would want to allocate a buffer by using `char buf[R_NUM_UNITS_BUFLEN]`. By defining the constant here you are making it private, while it would probably fit better in r_num.h so that users of RNum could use it.

radare: you can use `const int unitsBuflen = 32`. no need for a define here

libr/util/unum.c
```diff
@@ -672,18 +677,18 @@ R_API ut64 r_get_input_num_value(RNum *num, const char *input_value){
...
-static bool isHexDigit (const char _ch) {
-	const char ch = tolower ((ut8)_ch);
-	if (IS_DIGIT (ch)) {
-		return true;
-	}
-	return (ch >= 'a' && ch <= 'f');
-}
-
+/**
+ * \brief Get i_th nibble from integer n
```
radare: rename the function if its not clear what it does instead of adding 8 lines of comments to explain 3 lines of C.

libr/util/unum.c
```diff
@@ -853,3 +858,8 @@ R_API void *r_num_dup(ut64 n) {
 	*hn = n;
 	return (void*)hn;
 }
+
+R_API long long r_num_abs(long long num) {
```
radare: use st64 instead of long long

libr/util/unum.c
lzutao: I think @ret2libc is right. User could use R_NUM_UNITS_BUFLEN to allocate a buffer by using char `buf[R_NUM_UNITS_BUFLEN+1]` later.

radare:
ULL implies ut64

radare:
Passing buffers with implicit sizes is wrong by definition

lzutao:
@radare Do you mean that user imports `<r_util.h>` so definition of `R_NUM_UNITS_BUFLEN` is not there.

lzutao:
Pushed some changed in r2r to fix build, must rebuilt again for Travis to be green.

libr/include/r_util/r_num.h
```diff
@@ -2,7 +2,7 @@
 #define R_NUM_H
 
 #define R_NUMCALC_STRSZ 1024
-#define r_num_abs(x) x>0?x:-x
+#define R_NUM_UNITS_BUFLEN 31 /* used in r_num_units() */
```
radare: imho its wrong to have such define here, because the size of the buffer constructed to do this is internal to the function call. if you want to fix this, just add (and pass) the buffer length to the function instead of asuming the caller have allocated X bytes. also 31 is just a random guess, nothing scientific at all

libr/include/r_util/r_num.h
```diff
@@ -51,6 +51,21 @@ typedef const char *(*RNumCallback2)(struct r_num_t *self, ut64, int *ok);
 
 R_API RNum *r_num_new(RNumCallback cb, RNumCallback2 cb2, void *ptr);
 R_API void r_num_free(RNum *num);
+/**
+ * \brief Convert size in bytes to human-readable string (upto Exabyte)
```
radare: Don't put comments in include files. i dont think you need 15 lines to explain what the function does. Im not a big fan of comments as you can see :P and in fact the comments were suposed to be added only in the vapi files, so we can use valadoc to get documentation. for all the bindings at the same time, not just C.

libr/util/unum.c
```diff
@@ -75,31 +75,37 @@ R_API void r_num_free(RNum *num) {
...
+// TODO: Use SI system
+// https://metacpan.org/pod/Number::Bytes::Human
 R_API char *r_num_units(char *buf, ut64 num) {
 	char unit;
 	int tnum;
 	double fnum = num;
-	if (!buf) {
-		buf = malloc (32);
-		if (!buf) {
+	if (buf == NULL) {
```
radare: == NULL is. not necessary. just use !buf like it was before

libr/util/unum.c
```diff
@@ -75,31 +75,37 @@ R_API void r_num_free(RNum *num) {
...
+#define KB (((ut64)1) << 10)
+#define MB (((ut64)1) << 20)
+#define GB (((ut64)1) << 30)
+#define TB (((ut64)1) << 40)
+#define PB (((ut64)1) << 50)
+#define EB (((ut64)1) << 60)
+
+// TODO: Use SI system
+// https://metacpan.org/pod/Number::Bytes::Human
 R_API char *r_num_units(char *buf, ut64 num) {
```
radare: add int buflen

libr/util/unum.c
```diff
@@ -853,3 +851,8 @@ R_API void *r_num_dup(ut64 n) {
 	*hn = n;
 	return (void*)hn;
 }
+
+R_API st64 r_num_abs(st64 num) {
```
radare: why not put this function. in the .h as static inline? i cant think of any. reason to have this primitive not inlined, and shouldn't be available to the bindings. imho

libr/include/r_util/r_num.h
```diff
@@ -79,6 +94,10 @@ R_API int r_num_str_len(const char *str);
 R_API int r_num_str_split(char *str);
 R_API RList *r_num_str_split_list(char *str);
 R_API void *r_num_dup(ut64 n);
+/**
+ * \brief Compute the absolute value of an integer
```
radare: i think its clear enough. by the function name, no need to have such comment

radare:
in fact this was used wrongly because of this conceptual bad design:

```
libr/cons/cons.c:                       char *buflen = r_num_units (buf, I.context->buffer_len);
```

libr/util/unum.c
lzutao: [FreeBSD prefer testing explicit for NULL](https://www.freebsd.org/cgi/man.cgi?query=style&sektion=9&manpath=freebsd-release-ports).

Quoted:

```
Test pointers against NULL, e.g., use:
     (p	= f()) == NULL
not:
     !(p = f())

Do not use ! for tests unless it is a boolean, e.g. use:
     if	(*p == '\0')
not:
     if	(!*p)
```

I think it's more readable. But I change it as you ask.

libr/util/unum.c
lzutao: agreed

radare:
Yes.

lzutao:
@radare Should we make use of [`humanize_number(3)`](https://www.freebsd.org/cgi/man.cgi?query=humanize_number&apropos=0&sektion=0&manpath=FreeBSD+11.2-RELEASE+and+Ports&arch=default&format=html) on FreeBSD?

radare:
for consistency i think its fine to avoid depending on humanize_number

radare:
please solve those last comments and ill merge

libr/util/unum.c
ret2libc: As with any project, unfortunately, each one has its own set of rules and style. In r2 we do use `!` to check for NULL. You may find it less readable (I do prefer `== NULL` as well, actually) but the important thing is to be consistent with the rest of the code, which use `!`.

libr/include/r_util/r_num.h
lzutao: I guess that `r_num_units` need at least 6 bytes: 4 bytes for number (0-1023), 1 byte for unit and 1 for NULL.

lzutao: oh I forget the decimal number, so we need at least 8 bytes.

lzutao:
I'am going to remove floating-point number and use integer division instead (with remainder). 
So do I need to use locale to specify the decimal point or just use `'.'`?

radare:
I would just use .

lzutao:
Looks like the build failed because of non-matching unit. 
In new r_num_units, it coverts 1025 to 1.0K.
@radare Should 1025 be converted to 1.0K or 1K?
IMHO, 1.0K is more correct but looks weird.

radare:
1K imho

radare:
Sgtm

lzutao:
Does r2 support `long double` type? The conversion from `u64t` to `double` may 
cause data loss with integers have more 53 bits.

radare:
Thats up to the compiler not r2

radare:
see comments in here https://github.com/radare/radare2-regressions/pull/1544#pullrequestreview-174137877

radare:
i guess appveyour build will be green if you rebase again

radare:
please remove this trailing .0

radare:
Thanks!

PR #12080, opened 2018-11-07
remove r_bin_load and just use r_bin_open

ret2libc:
* remove r_bin_open because it was just a (useless) wrapper around r_bin_load
* rename r_bin_close to r_bin_file_close because actually what it does is to remove a binfile and not closing RBin as a whole
* set the BinFile ID in the r_bin_load_io instead of just in r_bin_open so that it is set in a more consistent way.


radare:
...

radare:
did u even read the reasoning of the rbinopen api? because its rbinload who should die

libr/bin/bin.c
```diff
@@ -395,7 +395,14 @@ R_API bool r_bin_load_io(RBin *bin, int fd, ut64 baseaddr, ut64 loadaddr, int xt
 	} else {
 		free (buf_bytes);
 	}
-	return binfile? r_bin_file_set_cur_binfile (bin, binfile): false;
+
+	if (!binfile || !r_bin_file_set_cur_binfile (bin, binfile)) {
+		return false;
+	}
+
+	int bd = bin->cur->id;
```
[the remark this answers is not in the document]
radare: what about just using binfile->fd ? which doesnt makes imiplicit the need for bin->cur to point to the newly. created binfile?

libr/core/cfile.c
```diff
@@ -314,21 +314,11 @@ static int r_core_file_do_load_for_debug(RCore *r, ut64 baseaddr, const char *fi
...
-		.offset = 0LL,
-		.baseaddr = baseaddr,
-		.rawstr = false,
-		.xtr_idx = xtr_idx,
-		.iofd = fd,
-		.loadaddr = UT64_MAX,
-	};
-
-	if (r_bin_open (r->bin, filenameuri, &bo) == -1) {
+	int fd = cf? cf->fd: -1;
```
radare: we prefer rbinopen/close than rbbinload. so this is a regression

libr/bin/bin.c
```diff
@@ -300,102 +328,109 @@ R_API bool r_bin_load_io(RBin *bin, int fd, ut64 baseaddr, ut64 loadaddr, int xt
...
 
-	r_return_val_if_fail (bin && io && fd >= 0 && (st64)sz >= 0, false);
+	bool is_debugger = iob->fd_is_dbg (io, opt->fd);
+	const char *fname = iob->fd_get_name (io, opt->fd);
 
-	bool is_debugger = iob->fd_is_dbg (io, fd);
-	const char *fname = iob->fd_get_name (io, fd);
-	if (loadaddr == UT64_MAX) {
-		loadaddr = 0;
+	bin->rawstr = opt->rawstr;
```
ret2libc: this was done before in r_bin_load and I moved it to r_bin_open_io to be sure it's always set, but... I'll need to fix this in the future because it doesn't make sense to change the RBin rawstr property because you open a new BinFile.

ret2libc:
@radare should be better now

ret2libc:
this is ready to be merged IMHO. Please review again.

libr/bin/bfile.c
```diff
@@ -837,3 +835,13 @@ R_IPI RList *r_bin_file_get_strings(RBinFile *a, int min, int dump, int raw) {
 R_API ut64 r_bin_file_get_baddr(RBinFile *binfile) {
 	return binfile? r_bin_object_get_baddr (binfile->o): UT64_MAX;
 }
+
+R_API bool r_bin_file_close(RBin *bin, int bd) {
```
radare: this function always returns false. you should return true when bf is not null

libr/bin/bin.c
```diff
@@ -116,6 +116,14 @@ R_API RList *r_bin_dump_strings(RBinFile *bf, int min, int raw) {
 	return r_bin_file_get_strings (bf, min, 1, raw);
 }
 
+R_API void r_bin_options_init(RBinOptions *opt, ut64 baseaddr, ut64 loadaddr, int fd, int rawstr) {
+	memset (opt, 0, sizeof (*opt));
+	opt->baseaddr = baseaddr;
+	opt->loadaddr = loadaddr;
+	opt->fd = fd;
```
radare: maybe fd should be the first arg

libr/core/cmd_search.c
```diff
@@ -258,7 +258,17 @@ static void cmd_search_bin(RCore *core, RInterval itv) {
...
+					.pluginname = plug->name,
+					.offset = core->offset,
+					.baseaddr = 0,
+					.loadaddr = 0,
+					.sz = 4096,
+					.xtr_idx = 0,
+					.rawstr = core->bin->rawstr,
+					.fd = core->file->fd,
+				};
+				r_bin_open_io (core->bin, &opt);
```
radare: there shoulkd be no different between RBin.open and RBin.open_io, the last one should die, but in a diferent PR

ret2libc: I think we should just have a `r_bin_open_buffer`. Bin should have nothing to do with IO. Everything should be done through IO and I think RBin shouldn't even depend on RIO.

Issue #11988, opened 2018-10-29
ob inconsistencies


radare:
* [ ] ob file -> should open a file, and obf must die
* [ ] obp bd -> should do what "ob bd" does now
* [ ] ob. -> the same as oba $$
* [ ] kill obb
* [ ] rename objid to binfd in ob?

radare:
* [ ] Add `obaa`: command to open a bin for all the opened files 'o'

radare:
Maybe we should make `io` command to be an alias for `ob` and kill what its doing right now

Maijin:
The whole oa, ob, oc, of, on, o for opening files is very confusing

XVilka:
I think it is better to remove `io` whatsoever.

Maijin:
I agree on removing `io` too

radare:
why

radare:
Also, ob should display the name of the plugin

c-urly:
How many of this task got over can anybody check the boxes above?

Maijin:
@c-urly none

trufae:
bringing back this discussion to the table. comments out there? anyone in the mood to do any of the changes enumerated there?

alex-bender:
```]> ob?
Usage: ob   # List open binary files backed by fd
| ob [bfid]              switch to open given objid
| ob                     list opened binary files and objid
| ob*                    list opened binary files and objid (r2 commands)
| ob *                   select all bins (use 'ob bfid' to pick one)
| ob-*                   delete all binfiles
| ob-[objid]             delete binfile by binobjid
| ob. ([addr])           show bfid at current address
| ob=                    show ascii art table having the list of open files
| obL                    same as iL or Li
| oba [addr] [baddr]     open file and load bin info at given address
| oba [addr] [filename]  open file and load bin info at given address
| oba [addr]             open bin info from the given address
| obf ([file])           load bininfo for current file (useful for r2 -n)
| obj                    list opened binary files and objid (JSON format)
| obn [name]             select binfile by name
| obo [fd]               switch to open binfile by fd number
| obr [baddr]            rebase current bin object
```
looks like `ob bd` does nothing now
kill obb -- looks like it's done

trufae:
moving to 6.x because enotime to do all the changes and the last release is delayed for more than 6 month

PR #12092, opened 2018-11-08
Rename objid to binfd in ob? ##Rbin 

Maijin:
Part of #11988

### DIFF
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -3636,7 +3636,7 @@ static int r_core_bin_file_print(RCore *core, RBinFile *binfile, int mode) {
 			RBinInfo *info = obj->info;
 			ut8 bits = info ? info->bits : 0;
 			const char *arch = info ? info->arch : "unknown";
-			r_cons_printf ("{\"objid\":%d,\"arch\":\"%s\",\"bits\":%d,\"binoffset\":%"
+			r_cons_printf ("{\"binfd\":%d,\"arch\":\"%s\",\"bits\":%d,\"binoffset\":%"
 					PFMT64d",\"objsize\":%"PFMT64d"}",
 					obj->id, arch, bits, obj->boffset, obj->obj_size);
 			if (iter->n) {
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -60,19 +60,19 @@ static const char *help_msg_oa[] = {
 
 static const char *help_msg_ob[] = {
 	"Usage:", "ob", " # List open binary files backed by fd",
-	"ob", "", "List opened binary files and objid",
-	"ob*", "", "List opened binary files and objid (r2 commands)",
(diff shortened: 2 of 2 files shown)


radare:
@ret2libc maybe objid is not even the good name for this. any other proposal to refer to them? also we are refering tot he binfile, but if the binfile have no binobj inside then its not listed by ob.. this seems inconsistent, but we will work out slowly with those refactorings

ret2libc:
`binfd` as done by @Maijin is not good?

Maijin:
I just changed objid to binfd like you asked in the issue @radare

radare:
The thing is... an RBinFile can have only one RBinObject. there was a time were we decided to have more than one rbinobject per rbinfile, for fatmach0 files and such, but for simplicity we are just squashing them into a single 1 concept, an rbinfile can have no rbinobject too.. so the id is refering right now to the RBinFile.

We should be consistent in concepts, api names and types to handle this here Thats why im asking, because io fds are named "fds", but here we should use probably "bds" (bin descriptor), i would remove the 'f' in the concept because there's no need to have a file in the game

ret2libc:
I don't know.. I think we are a bit too ahead with this PR. It's still not clear (at least to me) where we are going with RBinFile/RBinObject/etc. so I'd wait a bit before choosing.

radare:
can we continue this pls? pls @ret2libc think about it

libr/core/cbin.c
```diff
@@ -3636,7 +3636,7 @@ static int r_core_bin_file_print(RCore *core, RBinFile *binfile, int mode) {
 			RBinInfo *info = obj->info;
 			ut8 bits = info ? info->bits : 0;
 			const char *arch = info ? info->arch : "unknown";
-			r_cons_printf ("{\"objid\":%d,\"arch\":\"%s\",\"bits\":%d,\"binoffset\":%"
+			r_cons_printf ("{\"binfd\":%d,\"arch\":\"%s\",\"bits\":%d,\"binoffset\":%"
```
radare: i think we should use bd, as in bin descriptor. we have fd for file descriptors in rio, so it makes sense to use just 'bd' or maybe 'id' but that can be confusing because there can be other ids.

radare:
its not about changing names is about deciding one and using it across all the places, because right now that concept is a bit "vague"

and again do we want to kill the rbinobject or the rbinfile. and if so, we want a single rbinobject that is assoicated with an fd no matter if we load more than one rbin from the same bin, this makes sense to me and its simpler, but things must be discussed before going forward.

I have an idea for supporting multiple rbinobj, by  having a functon that returns the last binobj (following the io fd priority rules) that covers a specific memory address

ret2libc:
I think this can be closed. This is about a much larger refactoring involving RBin, so it's still too early for this change. Useless keeping the PR open IMHO.

radare:
I dont think its early for this change at all. The rbin refactor is coming really late

PR #12093, opened 2018-11-08
Add CCf*

thestr4ng3r:
I don't see a reason for the rad 'f'. At least I didn't find anything that would use the offset, but not pass 'f'.


thestr4ng3r:
Hmm, or wait, now it just prints everything if you are not at the start of a function...

thestr4ng3r:
This should be better. I also changed it to use `r_anal_get_fcn_in()` instead of `..._at()`, so you can call it from anywhere inside a function.

radare:
ok to merge?

thestr4ng3r:
I think so.

PR #12101, opened 2018-11-08
Fix incorrect set of asm.cpu

bet4it:
I don't understand what problems was fixed in d9fb5713db4b09b41135935e7fa81382069b5ddc.
`core->assembler->cpu` is cleared by `r_asm_use` and set by `r_config_set` if necessary, and `asm_cpu` is the old value of `asm.cpu`, we should't set it again, otherwise it will lead to a mess.

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -468,9 +468,6 @@ static int cb_asmarch(void *user, void *data) {
 		eprintf ("asm.arch: cannot find (%s)\n", node->value);
 		return false;
 	}
-	//we should strdup here otherwise will crash if any r_config_set
-	//free the old value
-	char *asm_cpu = strdup (r_config_get (core->config, "asm.cpu"));
 	if (core->assembler->cur) {
 		const char *newAsmCPU = core->assembler->cur->cpus;
 		if (newAsmCPU) {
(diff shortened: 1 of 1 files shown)


radare:
47 xx

radare:
screenshot 2018-11-13 at 02 14 11

radare:
ping

bet4it:
I will try to figure it out.

bet4it:
Reopen and merge this pull request? @radare

radare:
?

bet4it:
radare/radare2-regressions#1551 has resolved all of the `47 xx` mentioned above.
And if we don't merge this pull request, many tests will fail on the latest `radare2-regressions`. (At least on my local environment. I don't understand why it doesn't happen on travis)
For example, we can test https://github.com/radare/radare2-regressions/blob/master/new/db/asm/mips_v2_64 manually:
```
$ r2 -
 -- Use 'rabin2 -ris' to get the import/export symbols of any binary.
[0x00000000]> e asm.arch=mips;e asm.cpu=v2;e asm.bits=64;s 0
[0x00000000]> e cfg.bigendian=false;pad 000000cc
pref 0, (zero)
[0x00000000]> e asm.arch=mips;e asm.cpu=v2;e asm.bits=64;s 0
[0x00000000]> e cfg.bigendian=false;pad 000000ec
invalid
```

bet4it:
@radare

PR #12106, opened 2018-11-09
Fix uninitialised section name

F3real:
Fix for #11872

### DIFF
--- a/libr/bin/p/bin_pe.inc
+++ b/libr/bin/p/bin_pe.inc
@@ -167,6 +167,8 @@ static RList* sections(RBinFile *bf) {
 		}
 		if (sections[i].name[0]) {
 			ptr->name = strdup ((char*)sections[i].name);
+		} else {
+			ptr->name = strdup ("");
 		}
 		ptr->size = sections[i].size;
 		if (ptr->size > bin->size) {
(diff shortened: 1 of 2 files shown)


libr/bin/p/bin_pe.inc
```diff
@@ -167,6 +167,8 @@ static RList* sections(RBinFile *bf) {
 		}
 		if (sections[i].name[0]) {
 			ptr->name = strdup ((char*)sections[i].name);
+		} else {
+			ptr->name = strdup ("NULL");
```
radare: we should probably  use "" instead of "NULL" for such cases

F3real: @radare You suggested NULL in previous commit as default value. Should we change both places to "", or leave it as NULL ?
https://github.com/radare/radare2/pull/11872#discussion_r230454467

radare: Yes, i suggested an actual NULL, not a "NULL" string. I saw that in your PR but didn't wanted to delay the merge more time. And as long as it was green i didnt wanted to mess it more. A bunch of regressions happened, but none of them was catched by our testsuite, so i was waiting to start doing the "NULL" -> NULL transition before the release

F3real: ah, I didn't understand you well last time then. Changed both occurrences to _""_ now. Currently setting section name to NULL will just crash radare, so probably separate PR should be made adding  checks.

radare:
Yeah sorry if i didnt explained my position well. Yep that plan sounds perfect

PR #12107, opened 2018-11-09
Mdmp asserts

ret2libc:


libr/bin/p/bin_mdmp.c
```diff
@@ -488,6 +502,7 @@ RBinPlugin r_bin_plugin_mdmp = {
 	.libs = &libs,
 	.load = &load,
 	.load_bytes = &load_bytes,
```
radare: is there any path where load_bytes is necessary  when having load_buffer?

ret2libc: not that i'm aware of, but since it's still present in many other plugins I've kept it there. for now there are many plugins that don't implement load_buffer, but there is now a WARNING so i guess in the next 2 releases we are going to fix all of them and at that point we can completely remove `load_bytes` and `load`.

PR #12115, opened 2018-11-10
bin/filter: correctly allocate strings when filtering them

ret2libc:
Directly reallocate strings when filtering them, because it is very hard for
the caller to know before hand the correct size of the string otherwise. The
patch uses r_str_ APIs to makes this easy.

This fixes an ASAN detected overflow : https://travis-ci.org/radare/radare2/jobs/453150808#L1039 .

### DIFF
--- a/libr/bin/filter.c
+++ b/libr/bin/filter.c
@@ -3,54 +3,54 @@
 #include <r_bin.h>
 #include "i/private.h"
 
-static void hashify(char *s, ut64 vaddr) {
-	if (!s) {
-		return;
-	}
+static char *hashify(char *s, ut64 vaddr) {
+	r_return_val_if_fail (s, NULL);
+
+	char *os = s;
 	while (*s) {
 		if (!IS_PRINTABLE (*s)) {
+			free (os);
 			if (vaddr && vaddr != UT64_MAX) {
-				sprintf (s, "_%" PFMT64d, vaddr);
+				return r_str_newf ("_%" PFMT64d, vaddr);
 			} else {
 				ut32 hash = sdb_hash (s);
-				sprintf (s, "%x", hash);
+				return r_str_newf ("%x", hash);
 			}
-			return;
 		}
 		s++;
 	}
+	return os;
 }
 
 // TODO: optimize this api:
 // - bin plugins should call r_bin_filter_name() before appending
-R_API void r_bin_filter_name(RBinFile *bf, Sdb *db, ut64 vaddr, char *name, int maxlen) {
+R_API char *r_bin_filter_name(RBinFile *bf, Sdb *db, ut64 vaddr, char *name) {
+	r_return_val_if_fail (db && name, NULL);
+
 	const char *uname;
+	char *resname = strdup (name);
 	ut32 vhash, hash;
 	int count;
-	if (!db || !name) {
-		return;
-	}
-	uname = sdb_fmt ("%" PFMT64x ".%s", vaddr, name);
+
+	uname = sdb_fmt ("%" PFMT64x ".%s", vaddr, resname);
 	vhash = sdb_hash (uname); // vaddr hash - unique
-	hash = sdb_hash (name);   // name hash - if dupped and not in unique hash must insert
+	hash = sdb_hash (resname); // name hash - if dupped and not in unique hash must insert
 	count = sdb_num_inc (db, sdb_fmt ("%x", hash), 1, 0);
 	if (sdb_exists (db, sdb_fmt ("%x", vhash))) {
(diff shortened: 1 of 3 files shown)


libr/bin/filter.c
```diff
@@ -3,54 +3,54 @@
...
 }
 
 // TODO: optimize this api:
 // - bin plugins should call r_bin_filter_name() before appending
-R_API void r_bin_filter_name(RBinFile *bf, Sdb *db, ut64 vaddr, char *name, int maxlen) {
+R_API char *r_bin_filter_name(RBinFile *bf, Sdb *db, ut64 vaddr, char *name) {
+	r_return_val_if_fail (db && name, NULL);
+
 	const char *uname;
+	char *resname = strdup (name);
```
F3real: I think that `name` is leakead here

ret2libc: thanks! you are right

radare:
Offtopic: do you think it would be useful to preallocate hashtables to a specific size when filling them when we know the amount to avoid unnecessary resizes

libr/bin/filter.c
```diff
@@ -3,54 +3,54 @@
...
 		}
 		s++;
 	}
+	return os;
 }
 
 // TODO: optimize this api:
 // - bin plugins should call r_bin_filter_name() before appending
-R_API void r_bin_filter_name(RBinFile *bf, Sdb *db, ut64 vaddr, char *name, int maxlen) {
+R_API char *r_bin_filter_name(RBinFile *bf, Sdb *db, ut64 vaddr, char *name) {
```
radare: Returning an arg that is sometimes allocated in the heap may result in leaks

libr/bin/obj.c
```diff
@@ -241,7 +241,7 @@ static void filter_classes(RBinFile *bf, RList *list) {
 		char *namepad = malloc (namepad_len + 1);
 		if (namepad) {
 			strcpy (namepad, cls->name);
-			r_bin_filter_name (bf, db, cls->index, namepad, namepad_len);
+			namepad = r_bin_filter_name (bf, db, cls->index, namepad);
```
radare: If it returns null it leaks

libr/bin/filter.c
```diff
@@ -3,54 +3,54 @@
...
-		hashify (name, vaddr);
+		resname = hashify (resname, vaddr);
 	}
 	if (count > 1) {
-		int namelen = strlen (name);
-		if (namelen > maxlen) {
-			name[maxlen] = 0;
-		}
-		strcat (name, sdb_fmt ("_%d", count - 1));
+		resname = r_str_appendf (resname, "_%d", count - 1);
```
radare: This leaks because its not freeing the prev resname

ret2libc:
> Offtopic: do you think it would be useful to preallocate hashtables to a specific size when filling them when we know the amount to avoid unnecessary resizes

Sure, if you already know for sure the amount of elements, yes. If it's not sure or they could be less than what you think, just go without specifying any size and you won't notice any big difference IMHO.

libr/bin/obj.c
```diff
@@ -240,8 +240,12 @@ static void filter_classes(RBinFile *bf, RList *list) {
 		int namepad_len = strlen (cls->name) + 32;
 		char *namepad = malloc (namepad_len + 1);
 		if (namepad) {
+			char *p;
 			strcpy (namepad, cls->name);
-			r_bin_filter_name (bf, db, cls->index, namepad, namepad_len);
+			p = r_bin_filter_name (bf, db, cls->index, namepad);
+			if (p) {
+				namepad = p;
```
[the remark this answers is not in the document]
ret2libc: if necessary `r_bin_filter_name` automatically frees the `name`.

libr/bin/filter.c
```diff
@@ -3,54 +3,59 @@
...
 		s++;
 	}
+	return os;
 }
 
-// TODO: optimize this api:
-// - bin plugins should call r_bin_filter_name() before appending
-R_API void r_bin_filter_name(RBinFile *bf, Sdb *db, ut64 vaddr, char *name, int maxlen) {
+// - name should be allocated on the heap
+R_API char *r_bin_filter_name(RBinFile *bf, Sdb *db, ut64 vaddr, char *name) {
```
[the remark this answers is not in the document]
ret2libc: it can't be `const char *` because `name` could be freed in the function.

libr/bin/obj.c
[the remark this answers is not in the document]
radare: ah i see the r_str_append now. lgtm

PR #12123, opened 2018-11-11
Open radare2 project after loading plugins and don't lose -e flags

hewittc:
Defer project loading until after plugins have been loaded. Also addresses FIXME regarding loss of -e flags from getopt loop.

### DIFF
--- a/binr/radare2/radare2.c
+++ b/binr/radare2/radare2.c
@@ -675,9 +675,6 @@ int main(int argc, char **argv, char **envp) {
 				return 0;
 			}
 			r_config_set (r.config, "prj.name", optarg);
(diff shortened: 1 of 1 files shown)


radare:
red

binr/radare2/radare2.c
```diff
@@ -855,6 +852,11 @@ int main(int argc, char **argv, char **envp) {
 
 	r_bin_force_plugin (r.bin, forcebin);
 
+	if (r_config_get (r.config, "prj.name")) {
```
radare: dont get this var twice, just save in a local var and then use it. but this will be always true. because it will be "", not NULL, so you have to check for (*prj_name) instead

PR #12126, opened 2018-11-12
Fix leaks in bin.c

F3real:


ret2libc:
red on appveyor because of master

radare:
Because not rebased you mean? Master is green

ret2libc:
> Because not rebased you mean? Master is green

well... yes, master was red until 2h ago.

radare:
It was only 1 commit red because of a race with r2r merge. And this pr is red because of appveyour so totally unrelated to travis

PR #12130, opened 2018-11-12
forward to gmtime on windows platforms

lowlyw:
windows doesn't have gmtime_r, but it's already threadsafe, so do some stuff to make it so

### DIFF
--- a/libr/include/r_types.h
+++ b/libr/include/r_types.h
@@ -132,6 +132,22 @@
   #define __WINDOWS__ 1
 #endif
 
+#if defined(_WIN32)
+#ifndef gmtime_r
+#include <time.h>
+struct tm *gmtime_r(const time_t *t, struct tm *r) {
+	// gmtime is threadsafe in windows because it uses TLS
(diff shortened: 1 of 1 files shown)


XVilka:
```
 : error LNK2005: gmtime_r already defined in radiff2.c
```

pelijah:
https://stackoverflow.com/questions/19051762/difference-between-gmtime-r-and-gmtime-s
So we can replace gmtime_r with gmtime_s for msvc as it done in #12129

lowlyw:
actually it's not a drop in replacement.

on linux/mac

```c
     struct tm *
     gmtime_r(const time_t *clock, struct tm *result);
```

on windows

```c
errno_t gmtime_s(  
   struct tm* _tm,  
   const __time_t* time  
);  
```

notice the windows version returns an errno, and the posix version returns a struct tm *

pelijah:
Sad because I missed this thing and my PR was already merged.

pelijah:
Second try #12135

radare:
May we close this now?

PR #12131, opened 2018-11-12
Fix grub minix plugin build.

devnexen:
Using wrapper instead, and might be safe anyway
in case allocators impl change.


radare:
see https://github.com/radare/radare2/issues/12141

Issue #12112, opened 2018-11-10
Invalid instruction errors


davidpolverari:
# Invalid instruction errors

Since the last builds, r2 is issuing several `Invalid instruction (possibly truncated) at 0x...` errors when ran with `-dA` command line switches. When either analysis or debugging only is specified, such errors do not occur.

### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Debian testing (Buster) x86-64
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86/64
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.1.0-git 19992 @ linux-x86-64 git.3.0.1-180-g6c7bfd846 commit: 6c7bfd846a28c3494c01850943aa2bf5e84c5800 build: 2018-11-09__23:51:56

### Expected behavior
```
$ r2 -dA /bin/true
...
(analysis happens with no errors)
...
```

### Actual behavior
```
$ r2 -dA /bin/true
Process with PID 29696 started...
= attach 29696 29696
bin.baddr 0x55731f490000
Using 0x55731f490000
asm.bits 64
[Invalid instruction (possibly truncated) at 0x55731f499060
Invalid instruction (possibly truncated) at 0x55731f490131
Invalid instruction (possibly truncated) at 0x55731f490131
[x] Analyze all flags starting with sym. and entry0 (aa)
[x] Analyze function calls (aac)
[x] Analyze len bytes of instructions for references (aar)
...

```
### Steps to reproduce the behavior 
- Use `r2 -dA` with any ELF64 binary under Debian testing (Buster) x86-64.

radare:
Works without the -A?

davidpolverari:
Using only `-d` (without `-A`) it doesn't show any errors. But as soon as an `aaaa` is issued, the same errors are shown.

davidpolverari:
If only debugging or only analysis is attempted, everything is fine.

radare:
Thats because theres one analysis command being executed by aaa that is enabling emulation and breaking the process state. An if would be enough to cover this

Cc @maijin

davidpolverari:
Those errors didn't happen before 4a5048f7f4bf233653029e0b60bd5d2ab8517b0e

radare:
Can we have a static test? So @kazarmy can check it out and test it to avoid future reressions?

Thanks

[...]

PR #12132, opened 2018-11-12
No invalid instruction error on debug

kazarmy:
"Fix" #12112 by restoring previous behavior on debug. This doesn't break [this test](https://github.com/radare/radare2-regressions/blob/abc85cf8ceb387402fb3676fdf36ffad7ea23e09/new/db/cmd/cmd_af#L165-L167) in `cmd_af`.

### DIFF
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -958,7 +958,9 @@ repeat:
 			return R_ANAL_RET_ERROR;
 		}
 		if ((oplen = r_anal_op (anal, &op, addr + idx, buf + addrbytes * idx, len - addrbytes * idx, R_ANAL_OP_MASK_ALL)) < 1) {
-			eprintf ("Invalid instruction (possibly truncated) at 0x%"PFMT64x"\n", addr + idx);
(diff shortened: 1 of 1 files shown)


libr/anal/fcn.c
```diff
@@ -958,7 +958,9 @@ static int fcn_recurse(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut8 *buf, ut6
 			return R_ANAL_RET_ERROR;
 		}
 		if ((oplen = r_anal_op (anal, &op, addr + idx, buf + addrbytes * idx, len - addrbytes * idx, R_ANAL_OP_MASK_ALL)) < 1) {
-			eprintf ("Invalid instruction (possibly truncated) at 0x%"PFMT64x"\n", addr + idx);
+			if (!anal->coreb.core || !anal->coreb.cfggeti || !anal->coreb.cfggeti (anal->coreb.core, "cfg.debug")) { // HACK
```
radare: see the cfg.debug callback, it's setting this:

                core->io->debug = node->i_value;

as well as this:

                core->bin->is_debugger = true;

radare: better check one of those instead of calling cfg.get() all the time

PR #12138, opened 2018-11-13
Fix use after free in core.

devnexen:
The tasks are already freed by the reference counting.

### DIFF
--- a/libr/core/task.c
+++ b/libr/core/task.c
@@ -242,12 +242,13 @@ R_API void r_core_task_decref (RCoreTask *task) {
 		return;
 	}
 	TASK_SIGSET_T old_sigset;
-	tasks_lock_enter (task->core, &old_sigset);
+	RCore *core = task->core;
+	tasks_lock_enter (core, &old_sigset);
 	task->refcount--;
 	if (task->refcount <= 0) {
 		task_free (task);
 	}
-	tasks_lock_leave (task->core, &old_sigset);
+	tasks_lock_leave (core, &old_sigset);
 }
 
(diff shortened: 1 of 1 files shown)


radare:
Cc @th

radare:
Cc @thestr4ng3r

libr/core/task.c
```diff
@@ -182,6 +182,7 @@ static void task_free (RCoreTask *task) {
 	r_th_lock_free (task->dispatch_lock);
 	r_cons_context_free (task->cons_context);
 	free (task);
+	task = NULL;
```
thestr4ng3r: What is the point of setting the parameter to NULL?

libr/core/core.c
```diff
@@ -2450,7 +2450,6 @@ R_API RCore *r_core_fini(RCore *c) {
 	r_list_free (c->files);
 	r_list_free (c->watchers);
 	r_list_free (c->scriptstack);
-	r_list_free (c->tasks);
```
thestr4ng3r: The free function saved in core->tasks is r_core_task_decref, freeing the list should just decrease the refcount for each task as you would expect. If this leads to a uaf in some case, the problem is somewhere else, maybe somewhere the refcount is not increased where it should. If you have a case where this happens, please elaborate.

Also, you still have to free the RList itself or it will leak.

devnexen: you re right this was the wrong approach but there is genuinely an UAF case. With the other diff you can see why there is.

thestr4ng3r: Yes, you are absolutely right. Thanks for noticing, the fix is now correct.

PR #12147, opened 2018-11-13
Fix minor leaks

dav1901:
fix CID 1396670, 1396671 and 1396674 from #12141 .


radare:
Use free instead of R_FREE

PR #12148, opened 2018-11-14
Fix minor issues

dav1901:
fix CID 1396469 and 1396673.


libr/bin/filter.c
```diff
@@ -9,11 +9,12 @@ static char *hashify(char *s, ut64 vaddr) {
 	char *os = s;
 	while (*s) {
 		if (!IS_PRINTABLE (*s)) {
-			free (os);
 			if (vaddr && vaddr != UT64_MAX) {
+				free (os);
 				return r_str_newf ("_%" PFMT64d, vaddr);
 			} else {
```
radare: Dont else after return

PR #12150, opened 2018-11-14
do not call r_anal_cc_arg if we don't have cc info

ret2libc:


libr/core/anal_tp.c
```diff
@@ -707,21 +706,26 @@ R_API void r_core_anal_type_match(RCore *core, RAnalFunction *fcn) {
...
-		r_list_foreach (list2, iter2, bp_var) {
-			if (bp_var->isarg) {
-				const char *query = sdb_fmt ("fcn.0x%08"PFMT64x".arg.%d", fcn->addr, (bp_var->delta - 8));
-				char *type = (char *) sdb_const_get (anal->sdb_fcns, query, NULL);
-				if (type) {
-					var_retype (anal, bp_var, NULL, type, fcn->addr, false, false);
+	if (fcn->cc) {
+		const char *place = r_anal_cc_arg (anal, fcn->cc, 1);
+		if (place && !strncmp (place, "stack", 5)) {
+			RList *list2 = r_anal_var_list (anal, fcn, R_ANAL_VAR_KIND_BPV);
```
radare: Can be null?

ret2libc: I guess in theory it could be, but it shouldn't matter since r_list_foreach accepts NULL :( and r_list_free works nice with NULL.

PR #12152, opened 2018-11-14
core/cmd: macro->cmd should use a wrapper because of different types

ret2libc:


libr/core/cmd.c
```diff
@@ -3999,6 +3999,10 @@ static void cmd_descriptor_init(RCore *core) {
 	}
 }
 
+static int core_cmd0_wrapper(void *core, const char *cmd) {
```
radare: just for the void* -> RCore* so its not really a crashy problem. i was about to do a typedef and cast instead of doing this

ret2libc: I did this given what we said in the past about cast being unsafe in case something changes in the signature.. the cast will just work silently, the wrapper will give you an error back. But I'm ok with the cast too.

radare:
Yep casts can be dangerous. But.. maybe we can handle this with a proper signature instead of void*? Because cmd is no longer independent from core.. or we can make them take an rcorebind

Anyway the warning is gone amd the fix is good so meh

radare:
the XX seems totally unrelated can you rebase?

PR #12154, opened 2018-11-14
Fix some mdmp overflows spotted by googlefuzz

radare:


radare:
After merging this we can close https://github.com/radare/radare2/issues/12134 and commit all the fuzzed files in r2r

PR #12155, opened 2018-11-14
bin/elf: Do not filter 0, because of external imports

ret2libc:
external imports like __libc_start_main do not usually have a physical
address in the file and the virtual address is set to 0 because it is
not known statically (it is set at runtime by the loader). Other
disassemblers (e.g. IDA) create a special fake section where they put
those things one after the other, probably just as a way for users to
see those things linearly instead of having all these symbols at the
same address 0.

This patch improves the way PHDR and SHDR symbols/imports are mixed
together, by checking more than just the offset and by not filtering for
toffset == 0, which is a valid value.


radare:
Seems like a hashtable will make this faster and refavtoring out this chk lgtm. Waiting for travis

ret2libc:
Yes I thought about that but I’ll do it in a later PR because I first wanted to check if this thing breaks something.

PR #12158, opened 2018-11-14
implement load_buffer for MZ parser

p2mate:
Signed-off-by: p2 <p2@psychaos.be>


radare:
Will be good to remove the load bytes from all plugins and just have one callback named load with the buffer

p2mate:
you mean load_buffer and remove load and load_bytes?

radare:
yes, and then rename load_buffer to load() theres no need to have more than one way to load a bin when wehave rbuffer which abstracts the access to the data and solves overflow and memory usage issues

PR #12163, opened 2018-11-15
WIP:  Fix r_core_anal_graph_to and abt commands

kamou:
a tentative fix for r_core_anal_graph_to

### DIFF
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2753,7 +2753,7 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
 	return;
 }
 
-static bool anal_path_exists(RCore *core, ut64 from, ut64 to, RList *bbs, int depth) {
+static bool anal_path_exists(RCore *core, ut64 from, ut64 to, RList *bbs, int depth, RList *state) {
 	RAnalBlock *bb = r_anal_bb_from_offset (core->anal, from);
 	RListIter *iter = NULL;
 	RList *refs = NULL;
@@ -2769,11 +2769,12 @@ static bool anal_path_exists(RCore *core, ut64 from, ut64 to, RList *bbs, int de
 	if (!bb || !bbs) {
 		return false;
 	}
+	r_list_append(state, from);
 
 	// try to find the target in the current function
 	if (r_anal_bb_is_in_offset (bb, to) ||
-		(anal_path_exists (core, bb->jump, to, bbs, depth - 1)) ||
-		(anal_path_exists (core, bb->addr + bb->size, to, bbs, depth - 1))) {
+			((!r_list_contains(state, bb->jump) && anal_path_exists (core, bb->jump, to, bbs, depth - 1, state))) ||
+			((!r_list_contains(state, bb->addr + bb->size) && anal_path_exists (core, bb->addr + bb->size, to, bbs, depth - 1, state)))) {
 		r_list_prepend (bbs, bb);
 		return true;
 	}
@@ -2793,7 +2794,7 @@ static bool anal_path_exists(RCore *core, ut64 from, ut64 to, RList *bbs, int de
 			r_list_foreach (refs, iter, refi) {
 				if (refi->type == R_ANAL_REF_TYPE_CALL) {
 					if (r_anal_bb_is_in_offset (bb, refi->at)) {
-						if (refi->at != refi->addr && anal_path_exists (core, refi->addr, to, bbs, depth - 1)) {
+						if (refi->at != refi->addr && anal_path_exists (core, refi->addr, to, bbs, depth - 1, state)) {
 							r_list_prepend (bbs, bb);
 							return true;
 						}
@@ -2808,13 +2809,14 @@ static bool anal_path_exists(RCore *core, ut64 from, ut64 to, RList *bbs, int de
 
 R_API RList* r_core_anal_graph_to(RCore *core, ut64 addr, int n) {
 	RList *list = r_list_new ();
+	RList *state = r_list_new ();
 	if (!list) {
 		return NULL;
 	}
 
-	if (anal_path_exists (core, core->offset, addr, list, 1024)) {
+	if (anal_path_exists (core, core->offset, addr, list, 1024, state)) {
 		return list;
-    }
+	}
 
 	return NULL;
 }
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2843,7 +2843,7 @@ static RList* anal_graph_to(RCore *core, ut64 addr, int depth) {
 
 R_API RList* r_core_anal_graph_to(RCore *core, ut64 addr, int n) {
 	int depth = r_config_get_i (core->config, "anal.depth");
-	return anal_graph_to(core, addr, depth * 4);
+	return anal_graph_to (core, addr, depth * 4);
 }
 
 R_API int r_core_anal_graph(RCore *core, ut64 addr, int opts) {
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2827,7 +2827,7 @@ R_API RList* r_core_anal_graph_to(RCore *core, ut64 addr, int n) {
 		ht_up_free (state);
 		state = ht_up_new0 ();
 		if (xref->type == R_ANAL_REF_TYPE_CALL) {
-			if (anal_path_exists (core, core->offset, xref->at, list, depth * 4, state)) {
+			if (anal_path_exists (core, xref->addr, addr, list, depth * 4, state)) {
 				ht_up_free (state);
 				return list;
 			}
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2767,7 +2767,6 @@ static bool anal_path_exists(RCore *core, ut64 from, ut64 to, RList *bbs, int de
 	RListIter *iter = NULL;
 	RList *refs = NULL;
 	RAnalRef *refi;
-	RAnalFunction *fcn = NULL;
 	RAnalFunction *cur_fcn = NULL;
 	ut64 *frp = R_NEW0 (ut64);
 
@@ -2795,12 +2794,7 @@ static bool anal_path_exists(RCore *core, ut64 from, ut64 to, RList *bbs, int de
 	}
 
 	// find our current function
-	r_list_foreach (core->anal->fcns, iter, fcn) {
-		if (r_anal_fcn_in (fcn, from)) {
-			cur_fcn = fcn;
-			break;
-		}
-	}
+	cur_fcn = r_anal_get_fcn_in (core->anal, from, 0);
 
 	// get call refs from current basic block and find a path from them
 	if (cur_fcn) {
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2848,7 +2848,6 @@ R_API RList* r_core_anal_graph_to(RCore *core, ut64 addr, int n) {
 	int depth = r_config_get_i (core->config, "anal.graph_depth");
 	RList *path, *paths = r_list_new ();
 	HtUP *avoid = ht_up_new0 ();
-	n = 5;
 	while (n) {
 		path = anal_graph_to (core, addr, depth, avoid);
 		if (path) {
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2852,7 +2852,6 @@ R_API RList* r_core_anal_graph_to(RCore *core, ut64 addr, int n) {
     while (n) {
         path = anal_graph_to (core, addr, depth, avoid);
         if (path && (r_list_length (path) >= 2)) {
-			printf("found a path :)\n");
 			RAnalBlock *last = r_list_get_n (path, r_list_length (path) - 2);
             r_list_append (paths, path);
             ht_up_update (avoid, last->addr, last);
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2851,12 +2851,15 @@ R_API RList* r_core_anal_graph_to(RCore *core, ut64 addr, int n) {
 	n = 5;
 	while (n) {
 		path = anal_graph_to (core, addr, depth, avoid);
-		if (path && (r_list_length (path) >= 2)) {
-			RAnalBlock *last = r_list_get_n (path, r_list_length (path) - 2);
+		if (path) {
 			r_list_append (paths, path);
-			ht_up_update (avoid, last->addr, last);
-			n--;
-			continue;
+			if (r_list_length (path) >= 2) {
+				RAnalBlock *last = r_list_get_n (path, r_list_length (path) - 2);
+				ht_up_update (avoid, last->addr, last);
+				n--;
+				continue;
+			}
+			break;
 		}
 		// no more path found
 		break;
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2766,7 +2766,7 @@ static bool anal_path_exists(RCore *core, ut64 from, ut64 to, RList *bbs, int de
 		return false;
 	}
 
-	if (!bb){
+	if (!bb || !bbs){
 		return false;
 	}
 
@@ -2787,14 +2787,20 @@ static bool anal_path_exists(RCore *core, ut64 from, ut64 to, RList *bbs, int de
 	}
 
 	// get call refs from current basic block and find a path from them
-	refs = r_anal_fcn_get_refs (core->anal, cur_fcn);
-	r_list_foreach (refs, iter, refi) {
-		if (refi->type == R_ANAL_REF_TYPE_CALL) {
-			if (r_anal_bb_is_in_offset (bb, refi->at))
-				if (refi->at != refi->addr && anal_path_exists (core, refi->addr, to, bbs, depth - 1)) {
-					r_list_prepend (bbs, bb);
-					return true;
+	if (cur_fcn) {
+		refs = r_anal_fcn_get_refs (core->anal, cur_fcn);
+		if (refs) {
+			r_list_foreach (refs, iter, refi) {
+				if (refi->type == R_ANAL_REF_TYPE_CALL) {
+					if (r_anal_bb_is_in_offset (bb, refi->at)) {
+						if (refi->at != refi->addr && anal_path_exists (core, refi->addr, to, bbs, depth - 1)) {
+							r_list_prepend (bbs, bb);
+							return true;
+						}
+
+					}
 				}
+			}
 		}
 	}
 
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2846,21 +2846,21 @@ static RList *anal_graph_to(RCore *core, ut64 addr, int depth, HtUP *avoid) {
 
 R_API RList* r_core_anal_graph_to(RCore *core, ut64 addr, int n) {
 	int depth = r_config_get_i (core->config, "anal.graph_depth");
-    RList *path, *paths = r_list_new ();
-    HtUP *avoid = ht_up_new0 ();
+	RList *path, *paths = r_list_new ();
+	HtUP *avoid = ht_up_new0 ();
 	n = 5;
-    while (n) {
-        path = anal_graph_to (core, addr, depth, avoid);
-        if (path && (r_list_length (path) >= 2)) {
+	while (n) {
+		path = anal_graph_to (core, addr, depth, avoid);
+		if (path && (r_list_length (path) >= 2)) {
 			RAnalBlock *last = r_list_get_n (path, r_list_length (path) - 2);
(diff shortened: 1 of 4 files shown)


kamou:
this is WIP,
for now it only finds paths which does not require to recurse into function calls.

libr/core/canal.c
```diff
@@ -2753,39 +2753,28 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+		return true;
 	}
-	return list;
+
+	return false;
+}
+
+R_API RList* r_core_anal_graph_to(RCore *core, ut64 addr, int n) {
+	RList *list = r_list_new ();
+	if (r_core_anal_path_exists (core, core->offset, addr, list, 1024))
```
radare: check (`if (list &&`) and add missing braces {}

radare: why not make this function return the list or null in case of fail? instead of passing a reference as arg?

libr/core/canal.c
```diff
@@ -2753,39 +2753,28 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+		return list;
+	else return NULL;
```
radare: dont else after return

libr/core/cmd_anal.c
```diff
@@ -7499,13 +7499,10 @@ static int cmd_anal(void *data, const char *input) {
 		} else if (input[1] == 't') {
```
ret2libc: seems like `abt` is not listed in the help message :( would you mind adding it?

libr/core/canal.c
```diff
@@ -2753,39 +2753,28 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
 	return;
 }
 
-R_API RList* r_core_anal_graph_to(RCore *core, ut64 addr, int n) {
-	RAnalBlock *bb, *root = NULL, *dest = NULL;
-	RListIter *iter, *iter2;
-	RList *list2 = NULL, *list = NULL;
-	RAnalFunction *fcn;
+bool r_core_anal_path_exists(RCore *core, ut64 from, ut64 to, RList *bbs, int depth) {
```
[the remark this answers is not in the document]
kamou: cone

libr/core/cmd_anal.c
kamou: done

kamou:
> However... @radare creating a graph with all basic blocks of all functions in a binary may create a very large graph.. are we really sure we do want to do this analysis inter-function and not just intra-function?

It could be possible to only allow this if a particular option is set.
There could also be an option in order to control the depth variable. What do you think ?

kamou:
and I am not really creating a graph, rather a list of all basic blocks which should be reached in order to get to the target.

ret2libc:
> and I am not really creating a graph, rather a list of all basic blocks which should be reached in order to get to the target.

I know you are not, I did suggest however to directly work on the graph as that's the best way to analyze a graph IMHO. Moreover, your approach does not work in general (e.g. try to get from a basic block inside a loop to another one inside the same loop that comes *before* and you'll probably never get any path or if you get some it will be full of duplicated nodes and very likely not correct)

> There could also be an option in order to control the depth variable. What do you think ?

I think the depth variable is just the result of working recursively on a graph in the wrong way. There are proper algorithms to visit all nodes of a graph without relying on `depth` to interrupt the recursion. The recursive way you are analyzing things is too fragile (e.g. `radare2 -A -c 's main; abt 0x000039e0' ./r2r/bins/elf/analysis/ls-alxchk` will generates tons of "going too deep" but actually the address is just 3 blocks from the main one)

radare:
other conventions we use in r2land:

* start commits in capital
* if its wip state that in the PR title

libr/core/canal.c
```diff
@@ -2753,39 +2753,76 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+
+	if (depth < 1) {
+		eprintf ("going too deep\n");
+		return false;
+	}
 
+	if (!bb || !bbs) {
+		return false;
+	}
+	r_list_append(state, (void *)from);
```
radare: missing space

libr/core/canal.c
```diff
@@ -2753,39 +2753,76 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+	RAnalRef *refi;
+	RAnalFunction *fcn = NULL;
+	RAnalFunction *cur_fcn = NULL;
+
+	if (depth < 1) {
+		eprintf ("going too deep\n");
+		return false;
+	}
 
+	if (!bb || !bbs) {
```
radare: r_return_val_if_fail?

libr/core/canal.c
```diff
@@ -2753,39 +2753,76 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+	}
+
+	// get call refs from current basic block and find a path from them
+	if (cur_fcn) {
+		refs = r_anal_fcn_get_refs (core->anal, cur_fcn);
+		if (refs) {
+			r_list_foreach (refs, iter, refi) {
+				if (refi->type == R_ANAL_REF_TYPE_CALL) {
+					if (r_anal_bb_is_in_offset (bb, refi->at)) {
+						if (refi->at != refi->addr && !r_list_contains(state, (void *)refi->addr) && anal_path_exists (core, refi->addr, to, bbs, depth - 1, state)) {
```
radare: void* an ut64 is a bad idea, also missing a space

libr/core/canal.c
```diff
@@ -2769,12 +2779,17 @@ static bool anal_path_exists(RCore *core, ut64 from, ut64 to, RList *bbs, int de
...
+	r_list_append (state, (void *)frp);
+
+	ut64 tj = bb->jump;
+	ut64 fj = bb->addr + bb->size;
 
 	// try to find the target in the current function
 	if (r_anal_bb_is_in_offset (bb, to) ||
-			((!r_list_contains (state, (void *)bb->jump) && anal_path_exists (core, bb->jump, to, bbs, depth - 1, state))) ||
-			((!r_list_contains (state, (void *)(bb->addr + bb->size)) && anal_path_exists (core, bb->addr + bb->size, to, bbs, depth - 1, state)))) {
+			((!r_list_find (state, (void *)&(tj), cmp_addr) && anal_path_exists (core, bb->jump, to, bbs, depth - 1, state))) ||
```
radare: may be better to use r_ht_up, there seems to be many places where this list is iterated, now that hashtables are smaller we can evaluate using them in here. also, im a bit afraid of this hacky void*/ut64 usage. which would is solved if rewritten with the hashtable

ret2libc: yes the hash table is the best thing for this use case. See `sdb/ht_uu.h` and `sdb/ht_uu.c`

libr/core/canal.c
```diff
@@ -2753,39 +2753,90 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+
+	// try to find the target in the current function
+	if (r_anal_bb_is_in_offset (bb, to) ||
+		((!r_list_find (state, (void *)&(tj), cmp_addr) && anal_path_exists (core, bb->jump, to, bbs, depth - 1, state))) ||
+		((!r_list_find (state, (void *)&(fj), cmp_addr) && anal_path_exists (core, bb->addr + bb->size, to, bbs, depth - 1, state)))) {
+		r_list_prepend (bbs, bb);
+		return true;
+	}
 
+	// find our current function
```
ret2libc: you can use `R_API RAnalFunction *r_anal_get_fcn_in(RAnal *anal, ut64 addr, int type);` to get the function `from` belongs to.

libr/core/canal.c
ret2libc: r_return_val_if_fail only for `bbs`. `bb` needs to be checked with `if` as you've already done, because it is a runtime error that needs to be checked.

libr/core/canal.c
kamou: ok ! will do that.

libr/core/canal.c
kamou: I looked for something similar, but didn't look well enough apparently. thanks for the tip. will do that.

libr/core/canal.c
```diff
@@ -2753,39 +2753,90 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
-		eprintf ("Unable to find source or destination basic block\n");
+
+	return false;
+}
+
+R_API RList* r_core_anal_graph_to(RCore *core, ut64 addr, int n) {
+	RList *list = r_list_new ();
+	RList *state = r_list_new ();
+
+	r_return_val_if_fail (list && state, NULL);
```
ret2libc: nope! You should just handle the possible NULL values as usual. r_return_ functions should be used for programming errors only, not for runtime error conditions. If `r_list_new()` returns NULL that's probably because we are out of memory and that is something the programmer cannot do anything about, so it's not his fault.

As written in `r_assert.h`
```
 * If @expr evaluates to %FALSE, the current function should be considered to
 * have undefined behaviour (a programmer error). The only correct solution
 * to such an error is to change the module that is calling the current
 * function, so that it avoids this incorrect call.
```

libr/core/canal.c
```diff
@@ -2753,39 +2753,90 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
-				dest = bb;
-			}
-			if (r_anal_bb_is_in_offset (bb, core->offset)) {
-				root = bb;
-				r_list_append (list, list2);
+	}
+
+	// get call refs from current basic block and find a path from them
+	if (cur_fcn) {
+		refs = r_anal_fcn_get_refs (core->anal, cur_fcn);
```
ret2libc: I think you are not considering the "backward" edge of a call. The "ending" basic block of a function should have a link to the basicblocks that have a call to it.

libr/core/canal.c
kamou: just to make sure you didn't miss that change, I stopped casting ut64 to void* by using `r_list_find` instead of r_list_contains. and allocating a new ut64 each time I need to...
https://github.com/radare/radare2/blob/db2a8619761c76dcc9f6aa3bef9c5dde4e2e7315/libr/core/canal.c#L2772
https://github.com/radare/radare2/blob/db2a8619761c76dcc9f6aa3bef9c5dde4e2e7315/libr/core/canal.c#L2783
https://github.com/radare/radare2/blob/db2a8619761c76dcc9f6aa3bef9c5dde4e2e7315/libr/core/canal.c#L2756

but no problem can switch to hash tables, but isn't there an equivalent of sets instead ? because I just need unique values. not k/v pairs.

ret2libc: No we don't have `set` yet, but you can easily use `HtUU` easily to simulate a set. You just need to put your elements as keys (no need to allocate them on the heap) and a dummy value, just to mark them as present.

libr/core/canal.c
kamou: right ! thanks

kamou: correct me if I am wrong, but I think this is in fact not needed.
If you have a look at the code:
https://github.com/radare/radare2/blob/db2a8619761c76dcc9f6aa3bef9c5dde4e2e7315/libr/core/canal.c#L2791
you see that it is only once all blocks are analysed (when we reach a leaf block) that anal_path_exists will start following calls.

libr/core/canal.c
kamou: done

libr/core/canal.c
kamou: done

libr/core/canal.c
kamou: done

libr/core/canal.c
kamou: using the hashtable now.

libr/core/canal.c
kamou: I can't here, because I get a lot of warnings with this. I just want to silently ignore non predictable branches.

kamou: oh sorry, I did not see your comment @ret2libc, I can do that !

kamou: done

libr/core/canal.c
ret2libc: I'm not sure I follow you... what I'm saying is: suppose you have function A that calls function B. Your seek is at the last basic block of function B. Now you do `abt 0xYYYYYYY` where 0xYYYYYYY is an address in the function A, a few basic blocks after the call to function B. I would expect `abt` to give me the path that includes the `ret` instruction in the last basic block of function B. Not sure if it's clearer :)

kamou: pretty clear !

kamou: fixed

kamou:
I think I broke something.

libr/core/canal.c
```diff
@@ -2753,39 +2753,106 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+		return false;
+	}
+
+	r_return_val_if_fail (bbs, false);
+
+	if (!bb) {
+		return false;
+	}
+
+	if (bb->addr != from) {
```
ret2libc: why is this thing necessary?

libr/core/canal.c
```diff
@@ -2753,39 +2753,106 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+	xrefs = r_anal_xrefs_get (core->anal, cur_fcn->addr);
+
+	// forward search
+	if (anal_path_exists (core, core->offset, addr, list, depth - 1, state)) {
+		ht_up_free (state);
+		return list;
+	}
+
+	// backward search
+	if (xrefs) {
```
ret2libc: IMHO this seems like a big hack... Again, as said in one of my first messages, the "right" thing to do here would be to create a graph that connects all the basic blocks and then look for a path that connects two nodes in the graph with DFS.

radare:
Why do you want to create a dfg every time we query this instead of changing the storage of basic blocks to have all this already in place. But this seems like an intermediate solution.

ret2libc:
> Why do you want to create a dfg every time we query this instead of changing the storage of basic blocks to have all this already in place. But this seems like an intermediate solution.

Well, first because changing the RAnalBlock structure would be a real big thing. second, because what you need is an inter-function graph and not a intra-function graph. So... it's really hard IMHO to keep such data structure updated. For now I think it's just way easier to create this data structure on the fly.

But yes, I do agree we can keep this intermediate solution to start with.

libr/core/canal.c
kamou: It was just a useless idea, removing it.

libr/core/canal.c
kamou: I understand your point, I just needed this feature quickly. and I didn't have enough time to look at the DFS api and understand what is missing for the graph to be able to generate a supergraph of all functions. I can have a look at it, but later.

ret2libc: Yep it’s fine as is for now :) , I’ll re-look at the pr on Monday and test it. Thanks!!

kamou: cool !

radare:
thanks!

radare:
will be good to add tests for this after merged.

btw you should define and assign in the same line more frequently and use r_return_if_fail

libr/core/canal.c
```diff
@@ -2753,39 +2753,100 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
-			}
-			if (r_anal_bb_is_in_offset (bb, core->offset)) {
-				root = bb;
-				r_list_append (list, list2);
+	if (depth < 1) {
+		eprintf ("going too deep\n");
+		return false;
+	}
+
+	r_return_val_if_fail (bbs, false);
```
radare: this should go at the begining of the function, before the r_anal_bb_from_offset in the line 2757

libr/core/canal.c
```diff
@@ -2753,39 +2753,100 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+	// try to find the target in the current function
+	if (r_anal_bb_is_in_offset (bb, to) ||
+		((!ht_up_find (state, bb->jump, NULL) && anal_path_exists (core, bb->jump, to, bbs, depth - 1, state))) ||
+		((!ht_up_find (state, bb->fail, NULL) && anal_path_exists (core, bb->fail, to, bbs, depth - 1, state)))) {
+		r_list_prepend (bbs, bb);
+		return true;
+	}
+
+	// find our current function
+	cur_fcn = r_anal_get_fcn_in (core->anal, from, 0);
```
radare: deifne variable here instead of making it = NULL on top

libr/core/canal.c
```diff
@@ -2753,39 +2753,100 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+
+	// get call refs from current basic block and find a path from them
+	if (cur_fcn) {
+		refs = r_anal_fcn_get_refs (core->anal, cur_fcn);
```
radare: define refs here, reduce the scope

libr/core/canal.c
```diff
@@ -2753,39 +2753,100 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+
+	return false;
+}
+
+static RList* anal_graph_to(RCore *core, ut64 addr, int depth) {
+	RAnalFunction *cur_fcn = r_anal_get_fcn_in (core->anal, core->offset, 0);
+	RList *xrefs;
+	RList *list = r_list_new ();
+	HtUP *state = ht_up_new0 ();
+	RListIter *iter = NULL;
```
radare: no need to set to null

libr/core/canal.c
```diff
@@ -2753,39 +2753,100 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+	RAnalRef *xref = NULL;
+
+	if (!list || !state || !cur_fcn) {
+		return NULL;
+	}
+	xrefs = r_anal_xrefs_get (core->anal, cur_fcn->addr);
```
radare: define here

libr/core/canal.c
```diff
@@ -2753,39 +2753,100 @@ R_API void r_core_recover_vars(RCore *core, RAnalFunction *fcn, bool argonly) {
...
+		return list;
+	}
+
+	// backward search
+	if (xrefs) {
+		r_list_foreach (xrefs, iter, xref) {
+			if (xref->type == R_ANAL_REF_TYPE_CALL) {
+				ut64 offset = core->offset;
+				core->offset = xref->addr;
+				r_list_free(list);
```
radare: missing space

libr/core/canal.c
kamou: done

libr/core/canal.c
kamou: done

libr/core/canal.c
kamou: done

libr/core/canal.c
kamou: done

libr/core/canal.c
kamou: done

libr/core/canal.c
kamou: done

radare:
Lets wait for travis but lgtm

kamou:
how can I write tests for this ?
I tried adding a few things in new/db/anal/others (in radare2-regressions), but doesn't seem to be executed, not sure I'm doing things correctly...
how can I only test my test and not the whole test suite ?

libr/core/cmd_anal.c
```diff
@@ -7071,6 +7078,64 @@ static void cmd_anal_aav(RCore *core, const char *input) {
...
+			RList *path;
+			RListIter *pathi;
+			RListIter *bbi;
+			bool first_path = true, first_bb = true;
+			if (json) {
+				r_cons_printf ("[");
+			}
+			r_list_foreach (paths, pathi, path) {
+				if (json) {
+					if (!first_path)
```
radare: add braces {}

libr/core/cmd_anal.c
```diff
@@ -7071,6 +7078,64 @@ static void cmd_anal_aav(RCore *core, const char *input) {
...
+						r_cons_printf (", ");
+				}
+				r_cons_printf ("[");
+				r_list_foreach (path, bbi, bb) {
+					if (json && !first_bb)
```
radare: add braces

PR #12173, opened 2018-11-15
Fix another assert when printing blocks

ret2libc:
From the git commit comment:
An empty list is different than a NULL return value. The caller has to
call r_list_free anyway on the returned list, so there's no real
advantage in returning NULL, it just makes the caller's life harder
because it cannot understand if NULL is returned because the list was
empty (e.g. there are no IO maps, etc.) or because there were some
errors.

### DIFF
--- a/libr/core/cmd_help.c
+++ b/libr/core/cmd_help.c
@@ -427,7 +427,6 @@ static int cmd_help(void *data, const char *input) {
 		k = r_str_trim_ro (input + 1);
 		tmp = r_core_get_boundaries_prot (core, -1, k, "search");
 		if (!tmp) {
-			R_LOG_WARN ("Cannot get boundaries\n");
 			return false;
 		}
 		r_list_foreach (tmp, iter, map) {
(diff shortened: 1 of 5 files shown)


libr/core/cmd_help.c
```diff
@@ -426,6 +426,10 @@ static int cmd_help(void *data, const char *input) {
 	case 'B': // "?B"
 		k = r_str_trim_ro (input + 1);
 		tmp = r_core_get_boundaries_prot (core, -1, k, "search");
+		if (!tmp) {
+			R_LOG_WARN ("Cannot get boundaries\n");
```
radare: why not put this warnin in r_core_get_boundaries_prot?

ret2libc: because you don't always want to do this. Apparently there are cases where you have a fallback solution (e.g. in cmd_anal.c or in cmd_print.c)

ret2libc: Ok i have removed all those R_LOG_WARN. NULL is returned only if allocation fails and we don't usually prints messages when allocation fails... should be good to merge i think (let's wait for travis though)

radare:
yeah better like this

PR #12178, opened 2018-11-15
Enable pdrj ##print

kazarmy:


radare:
It is documented in the help message?

kazarmy:
I don't think the `j` suffix needs to be explicitly documented here.

PR #12183, opened 2018-11-15
Do not use a fake RBinFile/RBinObject when creating a file

ret2libc:
The fact that we are creating a file, means there is no
RBinFile/RBinObject and creating a fake one just to pass arch/bits seems
a bit overkill, plus the fact that all r_binfile/r_binobj APIs wouldn't
work well because the fake binfiles wouldn't contain anything.

### DIFF
--- a/binr/rabin2/rabin2.c
+++ b/binr/rabin2/rabin2.c
@@ -898,12 +898,8 @@ int main(int argc, char **argv) {
 		if (!bits) {
 			bits = 32;
 		}
-		if (!r_bin_use_arch (bin, arch, bits, create)) {
-			eprintf ("Cannot set arch\n");
-			r_core_fini (&core);
-			return 1;
-		}
-		b = r_bin_create (bin, code, codelen, data, datalen);
+
+		b = r_bin_create (bin, code, codelen, data, datalen, arch, bits, create);
 		if (b) {
 			if (r_file_dump (file, b->buf, b->length, 0)) {
 				eprintf ("Dumped %"PFMT64d" bytes in '%s'\n", b->length, file);
@@ -913,7 +909,7 @@ int main(int argc, char **argv) {
 			}
 			r_buf_free (b);
 		} else {
-			eprintf ("Cannot create binary for this format '%s'.\n", create);
+			eprintf ("Cannot create binary for this format '%s/%d/%s'.\n", arch, bits, create);
 		}
 		r_core_fini (&core);
 		return 0;
--- a/binr/ragg2/ragg2.c
+++ b/binr/ragg2/ragg2.c
@@ -68,21 +68,17 @@ static void list(REgg *egg) {
 
 static int create(const char *format, const char *arch, int bits, const ut8 *code, int codelen) {
 	RBin *bin = r_bin_new ();
-	RBuffer *b;
-	if (!r_bin_use_arch (bin, arch, bits, format)) {
-		eprintf ("Cannot set arch\n");
-		r_bin_free (bin);
-		return 1;
-	}
-	b = r_bin_create (bin, code, codelen, NULL, 0); //data, datalen);
+	int ret = 0;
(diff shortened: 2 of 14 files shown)


binr/ragg2/ragg2.c
```diff
@@ -68,21 +68,17 @@ static void list(REgg *egg) {
 
 static int create(const char *format, const char *arch, int bits, const ut8 *code, int codelen) {
```
radare: boolify this

binr/rabin2/rabin2.c
```diff
@@ -898,12 +898,8 @@ int main(int argc, char **argv) {
...
 			bits = 32;
 		}
-		if (!r_bin_use_arch (bin, arch, bits, create)) {
-			eprintf ("Cannot set arch\n");
-			r_core_fini (&core);
-			return 1;
-		}
-		b = r_bin_create (bin, code, codelen, data, datalen);
+
+		b = r_bin_create (bin, code, codelen, data, datalen, arch, bits, create);
```
radare: we need a better solution to share the arch/cpu/bits/os information across all the modules from a single place. I dont like the idea of adding more parameters to that function, because we will need to set more options in the futre like OS, or any other parameters to create an executable, which are a lot.

Thats why it was using a fake rbinobj. this is similar approach from what i started with the RBinOptions and the open/close api model. we have to configure the bin plugin or pass an RBinOptions as argument with all that metadata/hints to create the file.

im also worried about the code+codelen,data+datalen parameters which may be probably better to use rbuffer and allow to pass multiple sections and different paddings, etc. so this should change also

ret2libc: Yeah I was thinking the same about arch/bits/os/pluginname.... Do you have a suggested name for this new structure? I cannot think of any at the moment.

radare:
RBinTarget ? RBinArchOptions?

if we want to use the same across all the modules this should be a pointer to a pointer and put the definition in rutil instead

libr/bin/bin.c
```diff
@@ -1220,19 +1201,25 @@ R_API void r_bin_bind(RBin *bin, RBinBind *b) {
...
-	RBinFile *a = r_bin_cur (bin);
-	RBinPlugin *plugin = r_bin_file_cur_plugin (a);
+	const ut8 *data, int datalen, const char *arch, int bits, const char *pluginname) {
+	r_return_val_if_fail (pluginname, NULL);
+
+	RBinPlugin *plugin = r_bin_get_binplugin_by_name (bin, pluginname);
+	if (!plugin) {
+		R_LOG_ERROR ("No plugin named '%s'\n", pluginname);
+		return NULL;
+	} else if (!plugin->create) {
```
radare: no else after return

binr/rabin2/rabin2.c
radare: wtf ror some reason my  answer didnt reached github wtf

i proposed RBinArchOptions or as long as it needs to be reused across all modules it should live in util and be named like RArchOptions

ret2libc: It did reach github, I saw it. Anyway, why should it live in r_util? I didn't see any usage of arch/bits/etc. outside of RBin(or things that depend on RBin). I think arch/bits/etc. are very bin-specific concepts, so it makes sense to define them there.

radare:
shuold be in util because we need it from asm, anal, bin, core and probably in other places too. and its something that have no more dependencies than the system itself, because its just a wayt o define an architecture.

ret2libc:
`from asm, anal, bin, core`. Those are all things that depend on RBin. And I think in theory it's a concept that lives in "binary". But I guess we'll see when trying to implement it where this concept is actually useful.

radare:
no, asm and anal dont depend on rbin at all

ret2libc:
Ops, seems like so. Ok

PR #12187, opened 2018-11-15
WIP: Initial cleanup of the jemalloc library to actually compile

radare:


ret2libc:
What's the status of this? Can it be closed? I'd like to clean super-old PRs that have not been touched for a long time.

radare:
Needs to be finished and tested. Problem is we dont have any test for jemalloc

XVilka:
@ret2libc it is also quite related to:
- https://github.com/radare/radare2/issues/11886
- https://github.com/radare/radare2/issues/4328
- https://github.com/radare/radare2/issues/5390

Basically reliance on the host platform should be removed and the code should be refactored to be more portable, for example allowing heap navigation for remote gdb targets of a different architecture, bitness and endianess.

radare:
without having tests for jemalloc refactoring this is just going to break things, closing because in 2 years nobody cared to maintain this

PR #12195, opened 2018-11-17
Fix #12166 attempt.

### LINKED ISSUE #12166: ts* (without args) is not working
```
[0x1000011ec]> "td struct foo {int a;int b;};"
[0x1000011ec]> ts* foo
pf.foo dd a b
[0x1000011ec]> ts*
[0x1000011ec]> ts
foo
[0x1000011ec]>
```

devnexen:
Display structs list.


radare:
Free the list at the end

PR #12197, opened 2018-11-17
Fix #11611 - Do not assemble with wrong plugin

### LINKED ISSUE #11611: x86 Asm plugins error
### Work environment

| OS/arch/bits (mandatory)                             | Alpine Linux v3.8 x86_64
| r2 -v full output, **not truncated** (mandatory)     | radare2 2.9.0 74407 @ linux-x86-64 git.2.9.0 commit: 3e91ad8828fbd7388f0af964347da2e401b7c8eb build: 2018-09-07__15:54:33

### Expected behavior

The below command should work:
```r2
% rasm2 -a x86 -b 32 'call dword [0x11223344]'
ff1544332211   
```

### Actual behavior

```r2
% rasm2 'call dword [0x11223344]'
Cannot assemble 'call dword [0x11223344]' at line 3
invalid
```

### Additional information

The opposite operation works:
```r2
% rasm2 -a x86 -b 32 -d 'ff1544332211'
call dword [0x11223344]
```

`x86.as` and `x86.nasm` plugins works well:
```r2
% rasm2 -a x86.nasm -b 32 'call dword [0x11223344]'
ff1544332211   
% rasm2 -a x86.as -b 32 'call dword [0x11223344]'
ff1548332211
```

lzutao:


radare:
The selection of the assembler plugin is the first that matches the name and have a callback. This will make r2 more unusable

lzutao:
@radare Do you have any suggestion to fix #11611? That issue is really annoying when sometime you can assemble with x86 plugin and sometime you cannot.

radare:
The solution is to fix the default assembler.

lzutao:
AFAIK, the default x86 plugin of rasm2 is Capstone, which is just a disassembler.
So to choose appropriate default assembler, we could loop through plugins that 
have a name starting with `x86`, then choose the first has assembler callback.
But this hides information from user.

radare:
The default assembler is x86.nz and yes imwhat you describe is what r2 does by default. You can select another one with the asm.assembler var

Maijin:
Greetings,

x86.nz (x86 handmade assembler) is one of the many x86 assemblers available in radare2 and is updated frequently.  **Please add a test case so we can fix x86.nz in https://github.com/radare/radare2-regressions/tree/master/new/db/asm** (See https://github.com/radare/radare2-regressions/blob/master/new/README.md) You can do so with github web editor without even cloning the repo.

You can contribute to x86.nz by completing the following file https://github.com/radare/radare2/blob/master/libr/asm/p/asm_x86_nz.c.
You can also use keystone(http://keystone-engine.org) within radare2/rasm2 by installing the radare2 plugin via r2pm : 

```
	r2pm -i keystone-lib
	r2pm -i keystone
then
	rasm2 -a x86.ks…
or in radare2 session
	e asm.assembler = x86.ks
```

Others x86 assemblers are also available trough r2pm or master (see rasm2 -L list):

```
a___  16 32 64   x86.as      LGPL3   Intel X86 GNU Assembler
a___  16 32 64   x86.nasm    LGPL3   X86 nasm assembler
a___  16 32 64   x86.nz      LGPL3   x86 handmade assembler
ad__  32         x86.olly    GPL2    OllyDBG X86 disassembler
```

PR #12203, opened 2018-11-17
Minor code cleanups

dav1901:


libr/debug/p/debug_gdb.c
```diff
@@ -575,7 +575,6 @@ static const char *r_debug_gdb_reg_profile(RDebug *dbg) {
 			"seg	gs	.32	60	0\n"
 			);
 		}
-		break;
```
radare: Wtf

libr/debug/p/debug_gdb.c
```diff
@@ -726,7 +725,6 @@ static const char *r_debug_gdb_reg_profile(RDebug *dbg) {
 #endif
 			);
 		}
-		break;
```
radare: Wtf

libr/util/sys.c
```diff
@@ -861,7 +860,6 @@ R_API int r_sys_run(const ut8 *buf, int len) {
 	if (!pid) {
 		ret = cb ();
 		exit (ret);
-		return ret;
```
radare: Nope. Imagine exit fails, because the symbol is replaced or something we want to continue gracefully

libr/debug/p/debug_gdb.c
dav1901: Every if-else condition in that case has a return in it so the break is unnecessary.

libr/debug/p/debug_gdb.c
radare: it doesnt hurts because the statements are pretty long, and refactoring or commenting code for testing can become in unexpected results just because the programmer will assume there must be a break on each of these cases. im fine with cases with just return,but if that is a matter of 1-2 lines.

libr/socket/r2pipe.c
```diff
@@ -266,7 +266,6 @@ R_API R2Pipe *r2p_open(const char *cmd) {
 		}
 		r2p_close (r2p);
 		exit (rc);
-		return NULL;
```
radare: keep those returns in case of exit fails

libr/socket/proc.c
```diff
@@ -57,7 +57,6 @@ R_API struct r_socket_proc_t *r_socket_proc_open(char* const argv[]) {
 		dup2 (sp->fd1[1], 1);
 		execv (argv[0], argv);
 		exit (1);
-		break;
```
radare: keep it

libr/asm/p/asm_x86_vm.c
```diff
@@ -36,10 +36,6 @@ void decompile_vm(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
 			buf_asm = sdb_fmt ("vpcext %xh, %xh", buf[2], buf[3]);
 		}
 		op->size = 4;
-	} else if (len > 4 && buf[0] == 0x0F && buf[1] == 0xC6 && buf[2] == 0x28 && buf[3] == 1 && buf[4] == 0) {
```
radare: dupe?

PR #12205, opened 2018-11-18
XREF in local variables ##anal

### LINKED ISSUE #11865: XREF in local variables
### Work environment

| OS/arch/bits (mandatory)                             | Ubuntu x64
| Architecture/bits of the file (mandatory)            | x86 
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.0.0 19966 @ darwin-x86-64 git.2.9.0-395-g9beda2f60 commit: 9beda2f60f97a35f3e20c2c5f6933241bb78afbf build: 2018-10-18__11:04:28

### Expected behavior

- Get file [ch2.zip](https://github.com/radare/radare2/files/2494822/ch2.zip)
- aaa
- axt @ local_ch # But needs to be able to select function though

Should display something like:

```
main 0x0804831a [DATA] mov dword [local_ch], str.john
main 0x08048366 [DATA] mov eax, dword [local_ch]
```

### Actual behavior

Display nothing

sivaramaaa:
fixes #11865 

**Updated** 

screenshot from 2018-11-20 11-24-28

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -5540,6 +5540,43 @@ static void cmd_anal_ucall_ref (RCore *core, ut64 addr) {
 	}
 }
 
+static char *get_buf_asm(RCore *core, ut64 addr, RAnalFunction *fcn, bool color) {
+	int has_color = core->print->flags & R_PRINT_FLAGS_COLOR;
+	char str[512];
+	const int size = 12;
+	ut8 buf[12];
+	RAsmOp asmop = {0};
+	char *buf_asm = NULL;
+	bool asm_varsub = r_config_get_i (core->config, "asm.var.sub");
+	core->parser->pseudo = r_config_get_i (core->config, "asm.pseudo");
+	core->parser->relsub = r_config_get_i (core->config, "asm.relsub");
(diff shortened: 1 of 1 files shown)


radare:
That wont work if the fcn have a dot in its name?

sivaramaaa:
@radare yup , what about a solution like this , split this input with "."  and use last one as var name

for example `func.a.b.c.var_name` ==> [func,a,b,c,var_name] 

what do u think @Maijin , @XVilka

Maijin:
We need a way to do this, so any solution you can come with I'm all in 👍

sivaramaaa:
> axt main local_10h

sounds great , will do it !

radare:
Only problem i see is that current arg can contain spaces as a math expression ans with this change we enforce no spaces in those expressions

XVilka:
What's wrong with taking the last dot? We have the same approach for local
flags. So it will be consistent.

radare:
imageina f ucntion named fcn.main.loop and another one named fcn.main with  a local v ar named "loop".

using a space is safer here

XVilka:
Well, then lets go for space.

PR #12215, opened 2018-11-19
Add reloc name, wip to fix dupped relocs

radare:


libr/core/cbin.c
```diff
@@ -1492,7 +1492,9 @@ static int bin_relocs(RCore *r, int mode, int va) {
 			}
 			r_cons_printf ("vaddr=0x%08"PFMT64x" paddr=0x%08"PFMT64x" type=%s",
 				addr, reloc->paddr, bin_reloc_type_name (reloc));
-			if (reloc->import && reloc->import->name[0]) {
+			if (reloc->name) {
```
ret2libc: remove the whole `if` and just do a single `r_cons_printf (" %s", name);`

radare:
yeah, its wip, and iknow its gonna be red, but its not even solving the issue of the libhellojni bin, but didnt wanted to trash  or  lose my changes

radare:
I moved this to 3.2 because it requires some discussions and more  changes in rbin, so lets just focus on stability for the release

radare:
@ret2libc as long as you are working on elf and bin, wanna inherit this pr?

ret2libc:
Sure, but what's the problem exactly? Do you have an Issue, a test or a binary to reproduce the problem?

radare:
i dont have any bin now. but can be probably easy to craft one

ret2libc:
well this seems fixed in one of the recent PRs about relocs on ELF. Can we close?

radare:
moving forward, dont expect/want to get changes in here a week before the rls

radare:
did you tried setting two relocs in the same address? do we have a test related to this pr

radare:
this will increase memory usage a lot, and this pr have leaks

radare:
Closing, i guess this was not even a complete/proper fix, and now have conflicts and nobody seems to complain about tricked reloc hacks in here, we may handle this when we get a sample

PR #12236, opened 2018-11-20
Added ttc and tec in "tt?" and "te?" respectively. 

a-meth:


libr/core/cmd_type.c
```diff
@@ -824,7 +826,7 @@ static int cmd_type(void *data, const char *input) {
 		case 'b': // "teb"
 			res = r_type_enum_member (TDB, name, member_name, 0);
 			break;
-		case 'c' : { // "tec"
+		case 'c' : {
```
XVilka: Don't remove these comments, since they are required for searching the command implementation.

a-meth: Done!

a-meth:
> I noticed you added `ttc` command - can you please add a test in radare2-regressions for these commands emitting C output? So it wont be broken accidentally during some later changes.

Will look into it.

PR #12237, opened 2018-11-20
r_anal_bb_op_starts_at() uses binary search

kazarmy:


kazarmy:
I think linear seach is better than binary search here due to it being less complicated and also due to a low average block size, so I'm closing this.

PR #12241, opened 2018-11-20
Add RBinArchOptions type

ret2libc:
For now I put the RBinArchOptions in r_bin.h. It can be easily moved if it needs to be used elsewhere in the future. The reasons I did that is because I'm not sure we will be able to easily extend this structure and use it everywhere... Some places do require `os`, others don't. Some places do require `cpu`, others do not, etc... So I think the best thing is to keep it as simple as possible for now and choose in the future when we have a better view.

### DIFF
--- a/binr/rabin2/rabin2.c
+++ b/binr/rabin2/rabin2.c
@@ -899,18 +899,9 @@ int main(int argc, char **argv) {
 			return 1;
 		}
 		codelen = r_hex_str2bin (p, code);
-		if (!arch) {
-			arch = R_SYS_ARCH;
-		}
-		if (!bits) {
-			bits = 32;
-		}
-		if (!r_bin_use_arch (bin, arch, bits, create)) {
-			eprintf ("Cannot set arch\n");
-			r_core_fini (&core);
-			return 1;
-		}
-		b = r_bin_create (bin, code, codelen, data, datalen);
+		RBinArchOptions opts;
+		r_bin_arch_options_init (&opts, arch, bits);
+		b = r_bin_create (bin, create, code, codelen, data, datalen, &opts);
 		if (b) {
 			if (r_file_dump (file, b->buf, b->length, 0)) {
 				eprintf ("Dumped %"PFMT64d" bytes in '%s'\n", b->length, file);
--- a/libr/bin/bin.c
+++ b/libr/bin/bin.c
@@ -125,9 +125,8 @@ R_API void r_bin_options_init(RBinOptions *opt, int fd, ut64 baseaddr, ut64 load
 }
 
 R_API void r_bin_arch_options_init(RBinArchOptions *opt, const char *arch, int bits) {
-	memset (opt, 0, sizeof (*opt));
 	opt->arch = arch? arch: R_SYS_ARCH;
-	opt->bits = bits? bits: 32;
+	opt->bits = bits? bits: R_SYS_BITS;
 }
 
 R_API void r_bin_info_free(RBinInfo *rb) {
@@ -1217,7 +1216,8 @@ R_API RBuffer *r_bin_create(RBin *bin, const char *plugin_name,
 	if (!plugin) {
 		R_LOG_WARN ("Cannot find RBin plugin named '%s'.\n", plugin_name);
 		return NULL;
-	} else if (!plugin->create) {
+	}
+	if (!plugin->create) {
 		R_LOG_WARN ("RBin plugin '%s' does not implement \"create\" method.\n", plugin_name);
 		return NULL;
 	}
--- a/libr/bin/bin.c
+++ b/libr/bin/bin.c
@@ -956,39 +956,21 @@ R_API int r_bin_use_arch(RBin *bin, const char *arch, int bits, const char *name
 	r_return_val_if_fail (bin && arch, false);
 
 	RBinFile *binfile = r_bin_file_find_by_arch_bits (bin, arch, bits);
-	RBinObject *obj = NULL;
-	if (binfile) {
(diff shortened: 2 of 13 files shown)


XVilka:
Good idea actually. Probably even worth generalizing and using similar approach in RAnal or RDebug for example.

radare:
Why not do the same in util and remove the Bin from the name? We need this in debug, asm, anal, core, bin, syscall, ...

ret2libc:
Because I don't want to overengineer right now. Are you sure you don't need/want `os` or `cpu` or `model` or whatever in RDebug/RAnal/etc.? I'm not, actually. So I prefer that whoever will do the work to make this general afterwards will move it if he sees fit. With the current set of changes, this seems like the right solution, I don't know about the future :)

radare:
Then i see no benefit in doing this because its just the same of all the time just packed in a struct.

ret2libc:
> Then i see no benefit in doing this because its just the same of all the time just packed in a struct.

It is just the same, but packed in a struct, so that you can easily add `os`, `cpu`, etc. in the future. This is all according to your comment at https://github.com/radare/radare2/pull/12183#discussion_r234023143 . Moreover, it doesn't create anymore a fake RBinFile/RBinObject, which is a big win for me, since it simplifies those concepts.

I'm not saying we shouldn't add cpu/os/etc. to this structure or we shouldn't move it to RUtil in the future. I'm saying that right now it's not worth it because we don't really know for sure if/how we will need this structure in those libs. As said in the first comment, in many places where you do use arch/bits, you do not care about bin format. In others you do. In other places you do not care about os, in other you do. I think things should be make generic when there's a necessity and not before, otherwise you just risk to overengineer something that at the end may be not even used.

binr/rabin2/rabin2.c
```diff
@@ -899,18 +899,9 @@ int main(int argc, char **argv) {
...
-		}
-		if (!r_bin_use_arch (bin, arch, bits, create)) {
-			eprintf ("Cannot set arch\n");
-			r_core_fini (&core);
-			return 1;
-		}
-		b = r_bin_create (bin, code, codelen, data, datalen);
+		RBinArchOptions opts;
+		r_bin_arch_options_init (&opts, arch, bits);
+		b = r_bin_create (bin, create, code, codelen, data, datalen, &opts);
```
radare: those options of arch+bits is something that can live inside RBin, like it does in RDebug, RAsm, RAnal, RSyscall, .... iknow this stuff is only going to be used in RBinCreate and in RBinObjects and should be separate instances.. but im just trying to throw some comments to make this more consistent

libr/bin/bin.c
```diff
@@ -124,6 +124,12 @@ R_API void r_bin_options_init(RBinOptions *opt, int fd, ut64 baseaddr, ut64 load
 	opt->rawstr = rawstr;
 }
 
+R_API void r_bin_arch_options_init(RBinArchOptions *opt, const char *arch, int bits) {
+	memset (opt, 0, sizeof (*opt));
```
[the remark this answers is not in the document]
ret2libc: ok, going to remove this. It actually was overengineered, because initially I was planning to include os/cpu/etc.... but as said, I ended up leaving that for the future.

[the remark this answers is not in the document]
ret2libc: Ah no wait, I actually did this to ensure there is always a default value. See the `opt->arch = arch? arch: R_SYS_ARCH;` and the `opt->bits = bits? bits: 32;`. It's true you could do this with the C syntax, but you'd need to do it in every call site.

radare:
why 32? it looks wrong to assume this

ret2libc:
> why 32? it looks wrong to assume this

It was already done in https://github.com/radare/radare2/pull/12241/files#diff-815e6cc1a39b4e492b2616d4209eb380L906

radare:
can we move to 3.2?

libr/bin/bin.c
```diff
@@ -1219,20 +1207,28 @@ R_API void r_bin_bind(RBin *bin, RBinBind *b) {
...
+	const ut8 *code, int codelen, const ut8 *data, int datalen,
+	RBinArchOptions *opt) {
+
+	r_return_val_if_fail (bin && plugin_name && opt, NULL);
+
+	RBinPlugin *plugin = r_bin_get_binplugin_by_name (bin, plugin_name);
+	if (!plugin) {
+		R_LOG_WARN ("Cannot find RBin plugin named '%s'.\n", plugin_name);
+		return NULL;
+	} else if (!plugin->create) {
```
radare: dont return after else

libr/bin/bin.c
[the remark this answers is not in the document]
radare: use R_SYS_BITS then

radare: you are setting all the fields of the struct, no need for a memset

libr/bin/bin.c
ret2libc: I'm just `else` after return, I'm `else if`, which is different. I'm just dealing with the possible things that could happen..
1) plugin may be NULL
OR
2) if plugin was not NULL, plugin->create may be null

If I don't use the `else if` I'd just have to put the `if` on the following line, losing a bit of meaning to the mentioned code.

libr/bin/bin.c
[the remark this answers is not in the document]
ret2libc: ok

ret2libc: ok

libr/bin/bin.c
radare: the behaviour is the  same and its clear what it does. anything after return

PR #12246, opened 2018-11-20
Use -static in CFLAGS too

ret2libc:
This should fix googlefuzz build.


radare:
why

ret2libc:
don't know, but it seems to fix the issue when building on googlefuzz (at least on my machine)

PR #12248, opened 2018-11-21
rbin: simplify load_bytes() signature ##refactor

davidpolverari:
load_bytes() has a void **bin_obj parameter which is only used to set bf->o->bin_obj of RBinFile. As it already receives the relevant RBinFile parameter, this patch sets bin_obj directly on bf->o->bin_obj.

### DIFF
--- a/libr/bin/obj.c
+++ b/libr/bin/obj.c
@@ -182,7 +182,7 @@ R_IPI RBinObject *r_bin_object_new(RBinFile *binfile, RBinPlugin *plugin, ut64 b
 		if (sz < bsz) {
 			bsz = sz;
 		}
-		if (!plugin->load_bytes (binfile, &o->bin_obj, bytes + offset, sz,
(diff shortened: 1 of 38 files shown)


davidpolverari:
The only place where usage is a little bit different is at https://github.com/radare/radare2/pull/12248/commits/846332bfcbce4cbb362a3750f5f8a699d0eeff10#diff-ce3aeea66b7b1f6e50aaa345b1963acdL185. I think this one is causing problems, but still don't know the best way to refactor it.

radare:
why closed?

radare:
btw load_bytes() must die, we shoulkd only use load_buffer, but a simplification is always welcome

ret2libc:
Looks ok, but as said by @radare load_bytes should die, so it's not worth spending too much time making it cool.

davidpolverari:
I think I’m going to close this PR, as Travis is not green, and it won’t be that useful. What do you think?

radare:
Ok

PR #12255, opened 2018-11-21
Fix some COVs

ret2libc:


ret2libc:
fixes https://github.com/radare/radare2/issues/12026

ret2libc:
It should also fix https://github.com/radare/radare2/issues/11998, which I think were mostly false positives because of the `if (cfg->ht)` check, which probably made coverity things cfg->ht could be null. However, it is not possible and it should not happen, so the elements created by r_config_set are actually stored in the ht/list and are later freed from there.

PR #12259, opened 2018-11-21
Replace r_str_utf16_encode() for escaping strings for json

### LINKED ISSUE #11395: r_str_encode_utf16 doesn't work as expected
```
file     /tmp/Здравствуйте
```
```
"file":"/tmp/\u00d0\u0097\u00d0\u00b4\u00d1\u0080\u00d0\u00b0\u00d0\u00b2\u00d1\u0081\u00d1\u0082\u00d0\u00b2\u00d1\u0083\u00d0\u00b9\u00d1\u0082\u00d0\u00b5"
```
While it should be something like:
```
"file":"/tmp/\u0417\u0434\u0440\u0430\u0432\u0441\u0442\u0432\u0443\u0439\u0442\u0435"
```
at least for the json output.

Please if I don't fix it, add unit tests to that. Thanks.
This is more or less related: https://github.com/radareorg/cutter/issues/589

thestr4ng3r:
Fix #11395

I added the function `r_str_escape_utf8_to_json()`, which can be used to explicitly escape a utf-8 string for json.
`r_str_utf16_encode()` looks very wrong to me. All it does is go through the string byte by byte and print non-printable bytes as `\u00xx`, which is nonsense, because it corresponds to no real encoding and especially not to utf-16.

### DIFF
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -63,7 +63,7 @@ static void pair_str(const char *key, const char *val, int mode, int last) {
 		if (!val) {
 			val = "";
 		}
-		char *encval = r_str_utf16_encode (val, -1);
+		char *encval = r_str_escape_utf8_to_json (val, -1);
 		if (encval) {
 			char *qs = r_str_newf ("\"%s\"", encval);
 			pair (key, qs, mode, last);
@@ -1677,7 +1677,7 @@ static int bin_imports(RCore *r, int mode, int va, const char *name) {
 		} else if (IS_MODE_SIMPLE (mode)) {
 			r_cons_println (symname);
 		} else if (IS_MODE_JSON (mode)) {
-			str = r_str_utf16_encode (symname, -1);
+			str = r_str_escape_utf8_to_json (symname, -1);
 			str = r_str_replace (str, "\"", "\\\"", 1);
 			r_cons_printf ("%s{\"ordinal\":%d,"
 				"\"bind\":\"%s\","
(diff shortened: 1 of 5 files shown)


radare:
what about _for_ instead of _to_ in the namings? lgtm

kazarmy:
The UTF-8 encoding of `r_str_escape_utf()` can't be used to escape for json, since it escapes non-printable single-byte chars as `\xhh` and 4-byte chars as `\Uxxxxxxxx`, and json doesn't support either escape sequence.

> r_str_utf16_encode() looks very wrong to me. All it does is go through the string byte by byte and print non-printable bytes as \u00xx, which is nonsense, because it corresponds to no real encoding and especially not to utf-16.

The method name could be better, but apparently r_str_utf16_encode() is like that because json only supports `\uxxxx` for escaping non-printable bytes beyond some control chars.

thestr4ng3r:
Yes, that is correct indeed. However, if you have single, non printable bytes, you should probably not print it as a string at all and just use base64 I guess. For 4-byte unicode, not having \U is actually a pretty stupid limitaion... Perhaps we should not escape unicode at all and just print utf-8?

In any case, having a central function for escaping strings for json is a good thing since we can fix the things there now. So feel free to enhance/fix the implementation of `r_str_escape_utf8_to_json()` (or `r_str_escape_utf8_for_json()`) any time!

kazarmy:
> Perhaps we should not escape unicode at all and just print utf-8?

Yes but [RFC 8259](https://tools.ietf.org/html/rfc8259) says:
```
unescaped = %x20-21 / %x23-5B / %x5D-10FFFF
```
so can't run away from escaping.

> So feel free to enhance/fix the implementation of r_str_escape_utf8_to_json() (or r_str_escape_utf8_for_json()) any time!

Possibly for normal use cases and current r2 behavior this situation is fine.

PR #12265, opened 2018-11-22
Improve pf to support unions and improve visual xrefs interaction

radare:


radare:
pls @Maijin check behaviour of VpXx

radare:
cc @XVilka

radare:
Is this a quantum answer?

radare:
If i swap the if/else blocks then it makes no sense. Whats your proposal?

Maijin:
It is not quantum, current behavior in master is the following:

You are at line where there is a "lea rdi, str.COLUMNS". 

image

You press `X` you get the location of the string that you can reach:

image

You are in the data section on the string:
image

You press `x` and you get the xref to this string:

image

Your fix changes this behavior to show all the xref from the function both for x and X which is wrong and very confusing, it should only affects cursor mode so it gets the same behavior as above.

radare:
so you want swapped behaviour? aka:

* if cursor is set, get refs/xrefs from function
* if cursor is not set: get refs/xrefs from current offset

imho it makes sense to keep the current behaviour but is not consistent because cursor implies to be more precise.

Maijin:
This is my proposition:

For Cursor:
- Cursor is set and current selection contains XREF: get refs/xrefs from current offset
- Cursor is  set but the current selection does not contains XREF: do nothing
- Cursor is set on the function name: get refs/xrefs from function

For Non-cursor
- The current line contains XREF: get refs/xrefs from current offset
- The current line does not contains XREF: do nothing
- The current line is set on the function name: get refs/xrefs from function

radare:
i see no difference in cursor vs no cursor in your proposal

Maijin:
Because there shouldn't be any

radare:
any wat. i changed the pr according to your comment, so imho we can merge it and discuss this later

ret2libc:
3.2 IMHO. Not enough testing before release... which is in less then 2 days.

radare:
fixed an underflow and rebased to master to make it green

PR #12269, opened 2018-11-22
restore fd when loading symbols from file with oba

Wenzel:
the command `oba` with a filename was erasing the original fd:
screenshot_20181122_101858

This is fixed by this PR.

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -276,6 +276,7 @@ static void cmd_open_bin(RCore *core, const char *input) {
 			char *arg = strdup (input + 3);
 			char *filename = strchr (arg, ' ');
 			if (filename) {
+				int saved_fd = r_io_fd_get_current (core->io);
 				RIODesc *desc = r_io_open (core->io, filename + 1, R_PERM_R, 0);
 				if (desc) {
 					*filename = 0;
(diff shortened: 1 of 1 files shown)


libr/core/cmd_open.c
```diff
@@ -276,6 +276,7 @@ static void cmd_open_bin(RCore *core, const char *input) {
 			char *arg = strdup (input + 3);
 			char *filename = strchr (arg, ' ');
 			if (filename) {
+				int saved_fd = r_io_fd_get_current(core->io);
```
radare: Space before (

libr/core/cmd_open.c
```diff
@@ -285,6 +286,7 @@ static void cmd_open_bin(RCore *core, const char *input) {
 					r_bin_open_io (core->bin, &opt);
 					r_io_desc_close (desc);
 					r_core_cmd0 (core, ".is*");
+					r_io_use_fd(core->io, saved_fd);
```
radare: Same here. Also did u checked if issue happens too in the else filename branch?

Wenzel: no issue when using `oba 0x804d7000`

radare:
A better solution to this would be to make io->desc a stack instead of just a pointer, but this brings more problems because this should be an fd stack, not a desc* one, otherwise we can spot many uaf

radare:
cc  @ret2libc

radare:
@ret2libc im merging, but i want you to discuss this io->desc as a stack vs killing it or keeping fd instead of void* etc same with @condret 

imho we should kill this concept of fd priorizations. but i understand its useful sometimes.. i just dont see why it should fail if there's only one fd mapped on the whole memoryspace and then fail..

PR #12277, opened 2018-11-22
Reset cursor when using refs/xrefs in visual, otherwise it gets summed

ret2libc:


libr/core/visual.c
```diff
@@ -1222,7 +1222,10 @@ R_API int r_core_visual_refs(RCore *core, bool xref) {
 	} else if (IS_DIGIT (ch)) {
 		refi = r_list_get_n (xrefs, ch - 0x30);
 		if (refi) {
-			r_core_cmdf (core, "s 0x%"PFMT64x, refi->addr);
+			if (core->print->cur_enabled) {
+				core->print->cur = 0;
+			}
+			r_core_cmdf (core, "s 0x%" PFMT64x, refi->addr);
```
radare: would be awesome to use the api instead of rcorecmd :P

Issue #12278, opened 2018-11-22
gcc7 warning in libmagic


radare:
this code comes from openbsd.. so probably good to fix and report or sync if fixed there

[491/1127] Compiling C object 'libr/magic/libr@magic@@r_magic@sha/apprentice.c.o'.
In function 'parse',
    inlined from 'load_1' at ../libr/magic/apprentice.c:522:8:
../libr/magic/apprentice.c:955:10: warning: 'memset' writing 3200 bytes into a region of size 8 overflows the destination [-Wstringop-overflow=]
    (void)memset(p, 0, sizeof (*mp) * ALLOC_INCR);
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

PR #12290, opened 2018-11-22
Further meson warning build fixes.

devnexen:
While here attempt fixing #12278 old version of file/libmagic.


radare:
Can you confirm the warning is gone with this change?

devnexen:
Yes redid a full build with meson and gcc 8.

PR #12293, opened 2018-11-23
Allow oo reopen file in debugger when appropriate

### LINKED ISSUE #12292: r_bin_object_set_baddr: assertion 'o' failed (line 486)
### Work environment

| OS/arch/bits (mandatory)                             | MacOS Mojave / x86 / 64
| File format of the file you reverse (mandatory)      | mach064
| Architecture/bits of the file (mandatory)            | x86 / 64
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.1.0-git 20184 @ darwin-x86-64 git.3.0.1-339-g11fcc3bad commit: 11fcc3badc6cd2c418cdb21dbadb70ecc92ab43b build: 2018-11-22__14:49:48

### Expected behavior
- I expected the file to reopen without error

### Actual behavior
```
WARNING: r_bin_object_set_baddr: assertion 'o' failed (line 486)
```

**Note:** The file still opens correctly and I don't see any obvious side effects.

- Open in debugging mode - `r2 -d <any binary>`
- Reopen the file - `oo`

davidpolverari:
This patch allows oo behave as ood when r2 is in debugger mode.

Fix #12292

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -1468,7 +1468,9 @@ static int cmd_open(void *data, const char *input) {
 			}
 			break;
 		case '\0': // "oo"
-			if (core && core->io && core->io->desc) {
+			if (core->dbg) {
+				r_core_file_reopen_debug (core, input + 2);
+			} else if (core && core->io && core->io->desc) {
 				//does not work for debugging
 				int fd;
 				if ((ptr = strrchr (input, ' ')) && ptr[1]) {


davidpolverari:
If that is not the right approach, we can warn the user to use ood instead.

radare:
nope, you should use ood, otherwise you cant go back to normal file from debugger

davidpolverari:
Then I'll just alert the user when he tries to do oo when debugging to use ood instead, and do nothing in this case.

radare:
This change was a request from the cutter team

davidpolverari:
Which one? This:
https://github.com/radare/radare2/pull/12293#issuecomment-44113950? No problem! 😊It is already fixed, right?

PR #12296, opened 2018-11-23
Fix two minor warnings

dav1901:


libr/asm/p/asm_x86_nz.c
```diff
@@ -949,7 +949,7 @@ static int opdec(RAsm *a, ut8 *data, const Opcode *op) {
 	int reg = 0;
 	int rm;
 	bool use_sib = false;
-	int sib;
+	int sib = 0;
```
radare: i  bet i fixed this already  O_o maybe i lost the commit with the over 9000  branches i have in local. thanks

PR #12298, opened 2018-11-23
Further cleanup-code/less "optimistic" assumptions.

devnexen:


libr/util/file.c
```diff
@@ -63,7 +63,12 @@ R_API bool r_file_truncate (const char *filename, ut64 newsize) {
 #ifdef _MSC_VER
         _chsize (fd, newsize);
```
XVilka: See https://msdn.microsoft.com/en-us/library/dk925tyb.aspx - can you please share the error handling code between them?

devnexen: Yes hesitated not having a Win32 env at hand, no pb.

PR #12304, opened 2018-11-23
Fix cursor when following refs/xrefs with \n in visual mode

ret2libc:
Commit cc9d361dd317d3679d6af7ff0809744c8d82c945 only partially solved
the issue. The same should have been done for \n, \r, ' '. This patch
refactors the code to be sure both cases do the same thing.


radare:
lgtm

PR #12308, opened 2018-11-24
Fix #12307 - Escape ansi chars in cbin ##visual

### LINKED ISSUE #12307: ansi injection
Some strings coming from the binary are not filtered properly. This binary have a symbol and a section with special ansi chars that make rabin2 -s or r2 -cis a.out trash the console

- [ ] is
- [ ] pd @$S

[a.out.zip](https://github.com/radare/radare2/files/2611987/a.out.zip)

devnexen:
Let's try to display only displayable characters.


radare:
Nope. This issue is not related to macho. The filtering must be done in the printing code not in the parsing one otherwise we lose information

PR #12316, opened 2018-11-25
Better ansi injection fix

radare:


radare:
@kazarmy can you have a look at yhis issue too? Its opening yhe ansi macho bin and scrollingup few lines. I will check that later when at home

PR #12320, opened 2018-11-25
Fetch stuct name from the end in typdef deleration ##types

### LINKED ISSUE #12272: typedefed structs fail to load with to
``` 
$ cat findus.h

typedef struct {
    const char* name;
    const char* signature;
    void*       fnPtr;
} JNINativeMethod;
$ r2 -
[0x00000000]> t-*
[0x00000000]> to findus.h
[0x00000000]> tk
JNINativeMethod=typedef
struct.{=name,signature,fnPtr
struct.{.fnPtr=void *,8,0
struct.{.fnPtr.meta=4
struct.{.name=const char *,0,0
struct.{.name.meta=4
struct.{.signature=const char *,4,0
struct.{.signature.meta=4
typedef.JNINativeMethod=struct <anonymous>
{=struct
[0x00000000]> ts
{
[0x00000000]>
```

sivaramaaa:
Fixes #12272 

fix

### DIFF
--- a/shlr/tcc/tcc.h
+++ b/shlr/tcc/tcc.h
@@ -354,6 +354,7 @@ typedef struct ParseState {
     int line_num;
     int tok;
     CValue tokc;
+    uint8_t *p;
 } ParseState;
 
 /* used to record tokens */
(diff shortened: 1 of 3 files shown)


radare:
looks red

radare:
The jni.h is now not loading fine while its a valid include for gcc. It was working before, so i consider this a regression. Also exposed the __R2TINYC__ define in master, so you may rebase. im moving this to 3.2 because there's no time to test it or fix it properly now.

PR #12322, opened 2018-11-25
Types: some minor improvements to `ta`

AlexKornitzer:
Not all pointers to structs start at the beginning, thanks Microsoft for your implementation of list entries (LIST_ENTRY). Thus I have tweaked `ta` so that it can now take an additional argument `offset`. So for example we can now do the following: `ta _LDR_DATA_TABLE_ENTRY.DllBase 0x10`, as we are pointing not at the start of the structure but actually at offset 0x10 (the InInitializationOrderLinks field).
I have also extended the r_type_get_struct_memb function so that it will now parse inline structs rather than incorrectly bailing out. So below we can see that BaseDllName (UNICODE_STRING) is correctly resolved to the inlined structure.

Example of new improvements:
image

NOTE: Annoyingly cmd_type.c is formatted badly so there are lots of additions due to linting!

### DIFF
--- a/libr/core/cmd_type.c
+++ b/libr/core/cmd_type.c
@@ -41,7 +41,7 @@ static const char *help_msg_t_minus[] = {
 static const char *help_msg_ta[] = {
 	"Usage: ta[...]", "", "",
 	"tas", " <offset>", "List all matching structure offsets",
-	"ta", " <struct member>", "Change immediate to structure offset",
+	"ta", " <struct member> [offset]", "Change immediate to structure offset",
 	"taa", " [fcn]", "Analyze all/given function to convert immediate to linked structure offsets (see tl?)",
 	"ta?", "", "show this help",
 	NULL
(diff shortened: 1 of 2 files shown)


libr/core/cmd_type.c
```diff
@@ -41,7 +41,7 @@ static const char *help_msg_t_minus[] = {
 static const char *help_msg_ta[] = {
 	"Usage: ta[...]", "", "",
 	"tas", " <offset>", "List all matching structure offsets",
-	"ta", " <struct member>", "Change immediate to structure offset",
+	"ta", " <struct member> [offset]", "Change immediate to structure offset",
```
radare: isnt this [struct.member] ?

AlexKornitzer: Probably but did not want to change more than necessary

AlexKornitzer:
Have I handled adding optional args correctly, I just copied it from elsewhere but it is done differently in different commands!

radare:
no i mean the help message was wrong

AlexKornitzer:
Do you want me to change it to the suggestion above?

libr/core/cmd_type.c
```diff
@@ -1017,7 +1016,13 @@ static int cmd_type(void *data, const char *input) {
 			free (off);
 		} break;
 		case ' ': {
-			char *type = strdup (input + 2);
+			const char *off = NULL;
```
AlexKornitzer: Nah I meant, is this the correct way to handle optional arguments?

libr/core/cmd_type.c
```diff
@@ -1017,7 +1016,13 @@ static int cmd_type(void *data, const char *input) {
...
+			char *space, *type = strdup (r_str_trim_ro (input + 2));
+			space = strchr (type, ' ');
```
radare: define and assign in the same line. its easier to read

PR #12325, opened 2018-11-26
Show segment comments if there are no sections (elf only)

kazarmy:


libr/bin/p/bin_elf.inc
```diff
@@ -209,6 +210,7 @@ static RList* sections(RBinFile *bf) {
 				ptr->perm |= R_PERM_R;
 			}
 			r_list_append (ret, ptr);
+			bf->o->num_pure_sections++;
```
radare: this can be bool. and will be good to make this change for all the bin formats, not just in elf.

radare:
i also noticed a lot of inconsistencies with segments vs sections in rbin and rabin2. we shuold make both accessible easily

libr/bin/p/bin_elf.inc
kazarmy: ok bool in f8847b6635e1637f149996196e788249bb976d89 ... have to start somewhere regarding those bin formats

kazarmy:
this is better done with test bins on hand

libr/include/r_bin.h
```diff
@@ -224,6 +224,7 @@ typedef struct r_bin_object_t {
 	ut64 size;
 	ut64 obj_size;
 	RList/*<RBinSection>*/ *sections;
+	bool segments_only;
```
ret2libc: is it really necessary adding this field in the RBinObject? Can't you compute it in the `bin_sections` function since you are already iterating over all sections? If I understand it correctly, it would just be a matter of checking if there's any section with `is_segment == false`. That way, you don't need to set segments_only in each binplugin.

kazarmy: ofc not, yes b89bb4c206f45f1ca619c6c57328d134d8479e49

libr/core/cbin.c
```diff
@@ -2277,6 +2278,12 @@ static int bin_sections(RCore *r, int mode, ut64 laddr, int va, ut64 at, const c
 		r_cons_printf ("Nm Paddr       Size Vaddr      Memsz Perms %sName\n",
                    chksum ? "Checksum          " : "");
 	}
+	r_list_foreach (sections, iter, section) {
```
ret2libc: yeah seems buch better now. only small thing: you could do this only when IS_MODE_SET, because otherwise it's computed every time you print things.

kazarmy: huh? and where should I store the value of `segments_only`?

> otherwise it's computed every time you print things

2n is still O(n)

ret2libc: where you are storing it now. I know 2n is still O(n), but it's still double the time for average sizes.

ret2libc: anyway, as said it's a small thing, but I bet it would make @radare happy :-D

kazarmy: as you wish d6aae2d9914e9ad40323916ee0cdb0806e75cdc5 ... assumes that sections can't change after loading, but that's probably true currently since `S` has been killed

> I know 2n is still O(n), but it's still double the time for average sizes.

here it's not going to be double the time ... it's going to be more like going from `kn` to `(k+m)n` where `m` is a much smaller value than `k`, since the additional loop is very simple

ret2libc: Why did you turn segments_only into a static variable? there's no need for that.

kazarmy: oops 447cdd005581cde970d7852943eba07ecf47b10f

radare:
i dnt like the idea of mixing segments and sections information, but if ret2lbc aproves its fine to merge i guess, ill try to fix this stuff in rbin properly after the 3.1.1 release

ret2libc:
Well, we know the ideal solution would be to have RBinSection and RBinSegment properly separated. That's not the current situation unfortunately, with RBinSection that just represents both concepts :(

radare:
Yes thats what i mean..

PR #12326, opened 2018-11-26
No-debugger fixes

ret2libc:
s390x still cannot compile with debugger. I have really no interest in fixing it, but I think we should have an option to enable/disable the debugger, since apparently that's just an `#if` away. At least I don't need to patch radare2 on fedora to compile on s390x.


ret2libc:
@radare ok to merge?

radare:
Try to do proper commit messages and use the ## thing

ret2libc:
Sorry again :(

PR #12341, opened 2018-11-27
Simplification self plugin for BSD.

devnexen:
For 3 of them since we do not do mappings bookeeping and
displaying permission bits.

### DIFF
--- a/libr/io/p/io_self.c
+++ b/libr/io/p/io_self.c
@@ -546,48 +546,34 @@ bool bsd_proc_vmmaps(RIO *io, int pid) {
 		}
 		ut8 *p_start = p;
 		ut8 *p_end = p + size;
-		int n = 0;
-
-		while (p_start < p_end) {
-			struct kinfo_vmentry *entry = (struct kinfo_vmentry *)p_start;
-			size_t sz = entry->kve_structsize;
-			if (sz == 0) {
-				break;
-			}
-			p_start += sz;
-			n ++;
-		}
-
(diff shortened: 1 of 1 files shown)


libr/io/p/io_self.c
```diff
@@ -546,48 +546,34 @@ bool bsd_proc_vmmaps(RIO *io, int pid) {
...
-			goto exit;
-		}
-
-		p_start = p;
 
 		while (p_start < p_end) {
 			struct kinfo_vmentry *entry = (struct kinfo_vmentry *)p_start;
 			size_t sz = entry->kve_structsize;
 			int perm = 0;
+			char pstr[4] = {0};
```
radare: this wont  work because initial  string  should be "---". anyway this is implemented in r_str_rwx, better use this function

libr/io/p/io_self.c
```diff
@@ -622,23 +607,28 @@ bool bsd_proc_vmmaps(RIO *io, int pid) {
...
 		if (entry.kve_protection & KVE_PROT_EXEC) {
 			perm |= R_PERM_X;
+			pstr[2] = 'w';
 		}
 
-		io->cb_printf (" %p - %p [off. %zu]\n",
+		io->cb_printf (" %p - %p %3s [off. %zu]\n",
 				(void *)entry.kve_start,
 				(void *)entry.kve_end,
+				pstr,
```
radare: same here

PR #12348, opened 2018-11-28
Set the core offset right and update the panels correctly when scrolling disassembly##panels

Vane11ope:
Removed the "refresh" parameter from Panel so that all the panels will be refreshed all the time. 
The core offset was not set right in some places, now they are fixed.
Some fixes and refactoring are additional.


radare:
Isnt this making it slow ?

Vane11ope:
@radare not critically but i might have second thought. im getting it back wait.

Vane11ope:
@radare done. this is better. please edit the commit message like the title of this PR which i edited when you squash and merge. thank you :)

Vane11ope:
oh thought i messed up but actually master was red.

radare:
appveyour build is unhappy

radare:
please rebase master is green

PR #12349, opened 2018-11-28
Minor fixes/cleanups

dav1901:


libr/asm/arch/msp430/msp430_disas.c
```diff
@@ -180,7 +180,7 @@ static int decode_emulation(ut16 instr, struct msp430_cmd *cmd)
 /* return #byte of instruction */
 static int decode_addressing_mode(ut16 instr, ut16 op1, ut16 op2, struct msp430_cmd *cmd)
 {
-	int ret, srcOperInCodeWord = 0;
+	int ret = 0, srcOperInCodeWord = 0;
```
radare: will be better to update this file from gnu

PR #12353, opened 2018-11-29
Little type correction/panels.

devnexen:


devnexen:
Just realized there is build bug with meson due to missing clock nanosleep check, probably radare already working on it though.

PR #12356, opened 2018-11-29
Cut flags inbytes string if too long

### LINKED ISSUE #12294: Flags in bytes not working well
Sometimes flags in bytes are not positioned correctly:
img_20181123_032309_676

cyanpencil:
Fix #12294 

if the `flagstr` is too long (longer than ds->nb - 1, which is the number of cells available for opcode bytes, that depends on asm.nbytes), cut it with a `\0`


radare:
isnt this chopping the flags too early? in the screenshot of the issue seems like there maybe enough space to print that flag in there

PR #12357, opened 2018-11-29
Fix few leaks

dav1901:
Fix few leaks in astr.


libr/util/astr.c
```diff
@@ -43,7 +43,11 @@ R_API RASN1String *r_asn1_concatenate_strings (RASN1String *s0, RASN1String *s1,
 		r_asn1_free_string (s0);
 		r_asn1_free_string (s1);
 	}
-	return r_asn1_create_string (str, true, len);
+	RASN1String* asn1str = r_asn1_create_string (str, true, len);
```
radare: this seems to be a pretty bad design api. the ownership of str shouldn't be transferred conditionally

cc @wargio

radare:
Ping @wargio

wargio:
actually i don't see why the string has to be handled here in case of failure.
on normal libs you check the api output and if it fails you dealloc the string, not the library

libr/util/astr.c
wargio: RASN1String is meant for that.

radare:
pls rebase and solve conflicts

libr/util/astr.c
```diff
@@ -280,7 +301,11 @@ R_API RASN1String *r_asn1_stringify_oid (const ut8* buffer, ut32 length) {
 		}
 		++i;
 	} while (X509OIDList[i].oid && X509OIDList[i].name);
-	return r_asn1_create_string (str, true, ASN1_OID_LEN);
+	RASN1String* asn1str = r_asn1_create_string (str, true, ASN1_OID_LEN);
+	if (!asn1str) {
```
radare: as i said, r_asn1_create_string() should take const char * as 1st argument. and the caller should always free the str. dont make ownership behaviour conditional because that confuses developers and makes it easier to introduce bugs

dav1901: `r_asn1_create_string` returns `NULL` on failure and doesn't free the string, since `str` is allocated locally in this function, not freeing it in that case would cause a leak in memory.

radare:
...

i know, and thats WRONG behaviour. just fix the asn1_create_string to have a consistent behaviour, taking the first argument as const char* and never freeing the arg passed.

in fact the function takes const char* which is making this behaviour even more confusing, because the string is not const at all.

radare:
anyway, im merging, i will fix that code later, tired of waiting

PR #12362, opened 2018-11-30
Refline generation ignores comments

kazarmy:


kazarmy:
I think all this malloc'ing and free'ing is a bit slow, so don't merge yet.

kazarmy:
ok can merge now if green

PR #12363, opened 2018-11-30
Fixed a performance issue##panels 

Vane11ope:
Refreshing all the panels when scrolling disassembly causes some performance issue so fixed it.


radare:
Told ya :D

PR #12365, opened 2018-12-01
libr: debug: p: inline small functions

0xXA:


radare:
there's no way any compiler would inline a function that is only referenced by a struct member

PR #12387, opened 2018-12-03
Fix double-free in classes-from-symbols

radare:


radare:
No need to return them because the list is the o->classes and this is another bad design change. I will clone the struct and see if that makes it green. But this needs proper clean redesign. I have also fixed many issues introduced by the load buffer change in the nso plugin and there are still more to fix. I would like to go more radical for next release.

ret2libc:
No I explained myself wrongly probably... if you don't r_list_append, the o->classes is never changed in that function, so the entire function is useless. And you leak memory because the classes are allocated but never append to the list. Could you provide the reproducer to get the double free?

radare:
the r_list_append happens in another place. the pr is green, do you have a test that makes this fail?

PR #12394, opened 2018-12-04
Fix few clang warnings

dav1901:


libr/core/cmd_type.c
```diff
@@ -814,7 +814,7 @@ static int cmd_type(void *data, const char *input) {
 				ls_free (l);
 			} else { // "tej ENUM"
 				RListIter *iter;
-				RTypeEnum *member = R_NEW0 (RTypeEnum);
+				RTypeEnum *member;
```
devnexen: The scope could be slighty reduced possibly.

PR #12395, opened 2018-12-04
Creating a panel from the menu also needs to ask the user if they wan…

Vane11ope:
…t cache or not ##panels

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -475,7 +475,8 @@ static int layoutSidePanel(void *user) {
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


libr/core/panels.c
```diff
@@ -475,7 +475,8 @@ static int layoutSidePanel(void *user) {
 			}
 		}
 	}
-	addPanelFrame (core, panels, child->name, cmd, 0);
+	int caching = r_cons_yesno ('y', "Cache the result? (Y/n)");
```
radare: isnt this bool?

radare: fixed in master tired of watiing

Issue #12340, opened 2018-11-27
ARM aav interprets data as address and break thumb mode


sylvainpelissier:
*If you would like to report a bug, please fill the template bellow*

### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Linux arm 32
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | ARM, EABI5
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.1.0 20233 @ linux-x86-64 git.3.1.0-14-g13d5fcdd8 commit: 13d5fcdd87a7f824e4dde826b142264a135a10b2 build: 2018-11-27__17:43:56

### Steps to reproduce the behavior 

Compile
```
const int a = 0x000103c9;

int main()
{
		int b;
		b = 2;
		b = b + a;
        return 0;
}
```

with arm-linux-gnueabi-gcc

- Binary:
[thumb.zip](https://github.com/radare/radare2/files/2621715/thumb.zip)

`r2 -aa thumb` will work fine. But after `aav` command, the main function is interpreted in thumb instructions.

```
  :::::::   ;-- main:
  :::::::   ; UNKNOWN XREF from entry0 (+0x34)
  :::::::   ; UNKNOWN XREF from main (+0x38)
  :::::::   ; UNKNOWN XREF from section..rodata (+0x4)
  :::::::   0x000103c8      04b0           add sp, 0x10
  :`======< 0x000103ca      2de5           b 0xfe28
  : :::::   0x000103cc      00b0           add sp, 0
  :,======< 0x000103ce      8de2           b 0x108ec
  ========< 0x000103d0      0cd0           beq 0x103ec
  ========< 0x000103d2      4de2           b 0x10870
  :|:::::   0x000103d4      0230           adds r0, 2
  ========< 0x000103d6      a0e3           b 0x10b1a
  :|:::::   0x000103d8      0830           adds r0, 8
  :|::`===< 0x000103da      0be5           b 0xfdf4
  :|:: ::   0x000103dc      1c20           movs r0, 0x1c
  `=======< 0x000103de      9fe5           b 0xff20
   |:: ::   0x000103e0      0830           adds r0, 8
   |`=====< 0x000103e2      1be5           b 0xfe1c
   | : ::   0x000103e4      0230           adds r0, 2
   | :,===< 0x000103e6      83e0           b 0x104f0
[...]

PR #12398, opened 2018-12-04
Remove logic from _CbInRangeAav when arch equal to arm/thumb

alvarofe:
It was assumed all the values were functions but that might not be
true and then wrong hints would be set.

This should be handled by ESIL which is the only one it will know whether
a call is gonna be made with the data under analysis.

The issue #12340 shows how data is wrongly interpreted.

const int a = 0x000103c9;
```
int main()
{
	int b;
	b = 2;
	b = b + a;
	return 0;
}
```
It gets translated to
```
┌ (fcn) main 56
│   main (int argc, char **argv, char **envp);
│           ; UNKNOWN XREF from entry0 (+0x34)
│           0x000103c8      04b02de5       str fp, [sp, -4]!
│           0x000103cc      00b08de2       add fp, sp, 0
│           0x000103d0      0cd04de2       sub sp, sp, 0xc
│           0x000103d4      0230a0e3       mov r3, 2
│           0x000103d8      08300be5       str r3, [local_8h]          ; 8
│           0x000103dc      1c209fe5       ldr r2, aav.0x000103c9       ; [0x10400:4]=0x103c9 aav.0x000103c9
│           0x000103e0      08301be5       ldr r3, [local_8h]          ; 8
│           0x000103e4      023083e0       add r3, r3, r2
│           0x000103ec      0030a0e3       mov r3, 0
│           0x000103f0      0300a0e1       mov r0, r3
│           0x000103f4      00d08be2       add sp, fp, 0
│           0x000103f8      04b09de4       pop {fp}
└           0x000103fc      1eff2fe1       bx lr
            ; DATA XREF from main (0x103dc)
            0x00010400      .dword 0x000103c9 ; main
```
There are other cases where they should be handled elsewhere like below
```
|       #   0x000102f8      0c009fe5       ldr r0, [0x0001030c]        ; [0x1030c:4]=0x103c8 main
|       #   0x000102fc      0c309fe5       ldr r3, aav.0x00010404       ; [0x10310:4]=0x10404 aav.0x00010404
|       #   0x00010300      ebffffeb       bl sym.imp.__libc_start_main ;[1]   ; int __libc_start_main(func main, int argc, char **ubp_av, func init, func fini, func rtld_fini, void *stack_end)
        #   0x00010304      f0ffffeb       bl sym.imp.abort            ;[2]   ; void abort(void)
```

r2 should handle __libc_start_main to detect those functions but aav
should not make those assumptions

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -6987,46 +6987,20 @@ static void cmd_anal_aad(RCore *core, const char *input) {
 	r_list_free (list);
 }
 
-
-static bool archIsArmOrThumb(RCore *core) {
+bool archIsMips(RCore *core) {
 	RAsm *as = core ? core->assembler : NULL;
-	if (as && as->cur && as->cur->arch) {
-		if (r_str_startswith (as->cur->arch, "mips")) {
-			return true;
-		}
-		if (r_str_startswith (as->cur->arch, "arm")) {
-			if (as->bits < 64) {
-				return true;
-			}
-		}
+	if (as && as->cur && as->cur->name) {
+		return strstr (as->cur->name, "mips");
 	}
 	return false;
 }
 
-bool archIsMips (RCore *core) {
-	return strstr (core->assembler->cur->name, "mips");
-}
-
 void _CbInRangeAav(RCore *core, ut64 from, ut64 to, int vsize, bool asterisk, int count) {
-	bool isarm = archIsArmOrThumb (core);
-	if (isarm) {
-		if (to & 1) {
-			// .dword 0x000080b9 in reality is 0x000080b8
-			to--;
-			r_anal_hint_set_bits (core->anal, to, 16);
-			// can we assume is gonna be always a function?
-		} else {
-			r_core_seek_archbits (core, from);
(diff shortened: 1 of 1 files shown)


radare:
Red

alvarofe:
I removed the fix branch on r2r because I was getting XX in other places unrelated with the PR

radare:
[31m[XX][39m db/anal/others [33maav thumb detection[39m  91

$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -aarm -b32 -i /tmp/tmp-15969tT6mHwORyTad.tmp ../bins/firmware/armthumb.bin
aav
f


[35m-0x0000000c 4 aav.0x0000000c[39m
[35m[39m[32m+0x0000000d 4 aav.0x0000000d[39m
[32m[39mEXPECT64=MHgwMDAwMDAwZCA0IGFhdi4weDAwMDAwMDBkCg==
[0K

alvarofe:
`0xc` vs `0xd`, that it will be because of considering it as thumb - previous logic. I will update the test, but before that I will look at it more closely tomorrow to see if other action is needed.

PR #12403, opened 2018-12-04
Add scr.wideoff to make offsets and reg values depend on asm.bits ##visual

radare:
TODO:

* [x] honor scr.wideoff in hexdump and disasm

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -366,6 +366,13 @@ static int cb_scrlast(void *user, void *data) {
 	return true;
 }
 
+static int cb_scr_wideoff(void *user, void *data) {
+	RCore *core = (RCore *) user;
+	RConfigNode *node = (RConfigNode *) data;
+	core->print->wide_offsets = node->i_value;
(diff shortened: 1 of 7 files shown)


mrmacete:
lgtm 👍, it can't do harm. apart from r2r tests to adjust i mean.

radare:
we should probably apply this change to the prompt, the offset in disasm, the visual title... and it can become a bit cluttered. Comments on this?

XVilka:
@radare sounds reasonable change in prompt, disasm offset, visual title.

radare:
ive done that change. but im not happy to see that depending on asm.bits. so i  would prefer to h ave a separate variable, which probably depends on asm.bits, because there's no sense in showing 64bit address when working on a 16bit bootloader

radare:
also, maybe we want to have  a single place to generate that offset string.

proposals and ideas are welcome

wargio:
Yes and no. If the profile has a reg with nbits > asm.bits then show nbits

PR #12406, opened 2018-12-04
Update meson

lzutao:
Changelog:
```

* meson: Use install_man to install manpages
| 
| Also, let install zsh completion files.
| 
* meson: Check for explicit_memset
| 
* meson: Remove redundant calls to get_option
| 
| No need to join prefix with other project directories.
| 
* meson: Install shlr/www only if use_webui built option is true
| 
| * Add use_webui build option
| 
| Rationale: Debian doesn't install shlr/www by default, give them an option
| to enable/disable its installation.
| 
* meson: Update to 0.47
  
  Use a fallback version when sys/version.py failed.

```


xarkes:
cc @pelijah

meson.build
```diff
@@ -448,27 +445,28 @@ else
...
 man1_files = run_command(glob_cmd + ['man/*.1']).stdout().strip().split(';')
-install_data(man1_files,
-  install_dir: join_paths(get_option('mandir'), 'man1')
-)
-
 man7_files = run_command(glob_cmd + ['man/*.7']).stdout().strip().split(';')
-install_data(man7_files,
-  install_dir: join_paths(get_option('mandir'), 'man7')
-)
+install_man(man1_files, man7_files)
```
pelijah: Does it create `man1`/`man7` subdirs?

meson_options.txt
```diff
@@ -23,3 +26,5 @@ option('use_sys_xxhash', type: 'boolean', value: false)
 option('use_sys_openssl', type: 'boolean', value: false)
 option('use_libuv', type: 'boolean', value: true)
 option('debugger', type: 'boolean', value: true)
+
+option('use_webui', type: 'boolean', value: false, description: 'install different WebUIs for radare2')
```
pelijah: False by default?

meson.build
```diff
@@ -171,21 +175,23 @@ if host_machine.system() == 'windows'
   endforeach
 else
   r2_prefix = get_option('prefix')
-  r2_libdir = join_paths(r2_prefix, get_option('libdir'))
-  r2_incdir = join_paths(r2_prefix, get_option('includedir'), 'libr')
-  r2_datdir = join_paths(r2_prefix, get_option('datadir'))
+  r2_libdir = get_option('libdir')
+  r2_incdir = join_paths(get_option('includedir'), 'libr')
+  r2_datdir = get_option('datadir')
```
pelijah: R2_DATDIR, R2_INCDIR, R2_LIBDIR must be absolute

meson.build
```diff
@@ -1,16 +1,20 @@
-project('radare2', 'c', meson_version: '>=0.46.0')
+project('radare2', 'c', license: 'LGPL3', version: '3.1.2', meson_version: '>=0.47.0')
 
-py3_exe = import('python3').find_python()
+py3_exe = import('python').find_installation()
```
pelijah: We need exactly `python3`

meson_options.txt
ret2libc: me neither actually... also in fedora we don't ship it. For me it's ok to have it false by default. @radare ?

XVilka:
Seems a safe bet, since meson-0.46 is rarely the latest in most distributions already: https://repology.org/metapackage/meson/versions

meson_options.txt
radare: we need to update the webui, but nobody seems to be in the mood to contribute, for me its ok to disable it because some distros get anoyed with licence issues because of being distribuetd in binary form

radare:
ok to merge for me if you all are happy with those changes

PR #12429, opened 2018-12-07
Fix few more clang warnings

dav1901:


libr/io/p/io_rap.c
```diff
@@ -31,7 +31,7 @@ static int rap__write(RIO *io, RIODesc *fd, const ut8 *buf, int count) {
 	r_write_be32 (tmp + 1, count);
 	memcpy (tmp + 5, buf, count);
 
-	ret = r_socket_write (s, tmp, count + 5);
+	r_socket_write (s, tmp, count + 5);
```
radare: if we are ignoring the return value, which we probably shouldnt.. we must prefix with (void)

dav1901: It is ignored the same way across the file without the void prefix, should I add (void) to all or leave it as is?

radare:
Add the void or check the value in all the places

PR #12432, opened 2018-12-07
Better Hexdump support ##panels

Vane11ope:
Might need some refactoring but good for now.

I implemented this because I wanted it and I thought it would be useful too in some case generally.


Vane11ope:
betterhexdump
works like this.

PR #12435, opened 2018-12-09
Fix #8054 - Update flag registers in ADD/SUB for ARM/Thumb ##anal

### LINKED ISSUE #8054: [esil] flag updates for three address instructions
tl;dr: current flag update behavior for ESIL (update only on `=` or `cmp`) seems to make it difficult to implement three address instructions.

## Example

Let's look at the [ADD (register)](https://www.meriac.com/archex/AArch32_v82A_ISA/add_r.xml) instruction from the ARMv7 instruction set. Most encodings (A1, T1 and T3) support `Rd`, `Rn` and `Rm` to be independently specified. This requires the use of the `+` and `=` instructions for the ESIL translation, e.g.:

```
> rasm2 -a arm 'add r1, r2, r3' | rasm2 -a arm -E -
r3,r2,+,r1,=
```

If one source and the destination register are the same, the `+=` instruction is used in the translation:

```
> rasm2 -a arm 'add r1, r1, r2' | rasm2 -a arm -E -
r2,r1,+=
```

Let's compare this with the semantically equivalent instruction `add r1, r2, r1`:

```
> rasm2 -a arm 'add r1, r2, r1' | rasm2 -a arm -E -
r1,r2,+,r1,=
```

On the first look, `r2,r1,+=` (`esil_addeq`) and `r1,r2,+,r1,=` (`esil_add` followed by `esil_eq`) seem to do the same thing. For the simple `add` instruction that is actually true.

## Updating Flags

However, if we want to implement the `adds` instruction which updates the ARM status flags (carry `C`, negative `N`, overflow `V` and zero `Z`), we will notice that there is an important difference between calling `esil_addeq` and `esil_add` followed by `esil_eq` (all functions are found in `libr/anal/esil.c`).

`src,dest,+=` updates the `cur` and `old` values that are used to calculate ESIL's built in flags (`$c31`, `$s`, `$o`, `$z` etc.) like this (from `esil_addeq`):

```.c
esil->old = d;       // d is the value of the destination register
esil->cur = d + s;   // s is the source value
```

`src,dest,=` however does the following (from `esil_eq`):

```.c
esil->cur = num2;  // num2 is the source value
esil->old = num;   // num is the old destination value
```

## Problems

1.) it is very confusing that `r2,r1,+=` and `r1,r2,+,r1,=` are not semantically equivalent in ESIL
2.) if we do not use a combined binop and assignment instruction (`+=`, `-=`, `/=` etc.) there does not seem to be straight forward way to access the overflow or carry flags [0]

[0]: `$s` and `$z` only depend on the result, so they can be calculated from `esil->cur` only, which is the same for both instruction sequences.

## Solution?

In general I would suggest that flags should be updated whenever a binary (`+`, `-`, ...) or unary (`++`, `!`, ...) operation is performed, not when a register is assigned.
This would allow for `r2,r1,+=` and `r1,r2,+,r1,=` to have the same semantics.

However, I can see that this would cause some problems with `movs` like instructions. Thus one might have to update the flags on `=` as well. In this case we would end up with ESIL translations like: `r1,r2,+,$c,C,=,$s,N,=,$o,V,=,$z,Z,=,r1,=` or `r1,r2,+,$c,C,=,$o,V,=,r1,=,$s,N,=,$z,Z,=`.

Also, changing such a fundamental thing as how the flags are updated would probably introduce a lot of new bugs in stable ESIL translators. I have mostly looked at the ARM translator which is in dire need of repair, but I think that other implementations are actually a lot more complete and well tested.

I would love to hear your ideas on how to implement flag updates for three address instructions in ESIL.

condret:
lets wait for travis' hateRED against me

Fixes https://github.com/radare/radare2/issues/8054

### DIFF
--- a/libr/anal/p/anal_arm_cs.c
+++ b/libr/anal/p/anal_arm_cs.c
@@ -1422,6 +1422,7 @@ static int analop64_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int l
 
 #define MATH32(opchar) arm32math(a, op, addr, buf, len, handle, insn, pcdelta, str, opchar, 0)
 #define MATH32_NEG(opchar) arm32math(a, op, addr, buf, len, handle, insn, pcdelta, str, opchar, 1)
+#define MATH32AS(opchar) arm32mathaddsub(a, op, addr, buf, len, handle, insn, pcdelta, str, opchar)
 
 static void arm32math(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len, csh *handle, cs_insn *insn, int pcdelta, char (*str)[32], const char *opchar, int negate) {
 	const char *dest = ARG(0);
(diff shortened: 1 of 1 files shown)


radare:
Le red

condret:
ofc red, what else expect from travis

condret:
can someone pls check if the results are correct now, I'm not that much an arm person

radare:
Seems like all the XX are correct, so this pr is wrong

radare:
your code is removing the & 0xffffff mask, as well as making the expressions less readable.

compare this:

```
8            04108de2   sp,r1,=,4,r1,+=
```
vs
``` 
8            04108de2   4,sp,+,0xffffffff,&,r1,=
```

thats an add r1, sp, 4 from FileDP

radare:
also, the issuew as complaining about the lack of updating flags. this patch doesnt change this. @mrmacete can you have a look at this too?

radare:
Still 16 xx

radare:
greeeeeeeeen

alvarofe:
sorry I didn't see until now. Was not possible to reuse arm32math to form the esil expression for the operation and then appending at the end the flags changes? saying that because to me `arm32addsub` does not support as much as `arm32math` 

````
arm32addsub() {
    arm32math();
    update_flags();
}
```
Should I look this with real samples and the spec to be really sure, although you gave the first step ;)

PR #12439, opened 2018-12-10
Fiximg little clang warnings in various places.

devnexen:


radare:
Can you fix them all in a singlepr instead of 90000 different ones?

devnexen:
It was to avoid colliding with dav1901 possibly but i think that s probably one of the last.

Issue #11756, opened 2018-10-07
Implement c output format for all t commands


radare:
So we can tc > foo.h and then to foo.h

See my previous commits in tu, ts, te.. to handle json

* [x] ttc
* [x] tec
* [x] tsc
* [x] tuc

radare:
Cc @sivar

radare:
Cc @sivaramaaa

ret2libc:
@radare 3.2 for this?

radare:
I would love to have tsc before the release (sunday) can you @a-meth implement it in time?

thanks!

radare:
@ret2libc 3.1 is totally fine. tuc is a copypasta of tsc just changing the type name. the only  issue i can imagine is recursive struct definitions, but printing a struct shouldnt recurse at all

radare:
No pr yet so moving to 3.2

Maijin:
Reopening as it is missing `tc` and the help for `tsc` and `tuc`

radare:
Which command to use to list callcint cnventions? Do we have any dupped? @sivaramaa

Maijin:
@sivaramaaa *

radare:
too late, i have already checked the code and replied and merged

PR #12444, opened 2018-12-10
Implemented tc

a-meth:
Issue #11756

### DIFF
--- a/libr/core/cmd_type.c
+++ b/libr/core/cmd_type.c
@@ -137,6 +137,7 @@ static const char *help_msg_ts[] = {
 	"tsj", " [type]", "Show pf format string for given struct in json",
 	"ts*", "", "Show pf.<name> format string for all loaded structs",
 	"ts*", " [type]", "Show pf.<name> format string for given struct",
+	"tsc", "", "List all loaded structs in C output format", 
 	"tss", " [type]", "Display size of struct",
 	"ts?", "", "show this help",
 	NULL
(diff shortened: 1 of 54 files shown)


libr/core/cmd_type.c
```diff
@@ -745,7 +745,10 @@ static int cmd_type(void *data, const char *input) {
 			r_core_cmd0 (core, "k anal/cc/*");
 			break;
 		case 0:
-			r_core_cmd0 (core, "k anal/cc/*~=cc[0]");
```
radare: you are breaking current tc, without changing the help message or providing an alternative command to achieve the listing o f calling conventions. i think there's another issue for this

radare: i think tc is already  in afcl, so i think this PR is fine, but you should fix the help message first

a-meth: Fixed.

libr/core/cmd_type.c
```diff
@@ -745,7 +745,10 @@ static int cmd_type(void *data, const char *input) {
...
+			r_core_cmd0 (core, "tuc");
```
radare: we can probably simplify this  into 1 single line:

r_core_cmd0 (core, "tuc;tsc;ttc;tec");

radare:
Thanks!!

XVilka:
@a-meth can you please add a few tests for tc/tsc/etc?

XVilka:
By the way PR was wrong. See `help_msg_tc()` which is unfixed, also the actual `tc` implementation.

a-meth:
Can I just remove the help_msg_tc()?

XVilka:
I sent a proper fix already: https://github.com/radare/radare2/pull/12451

PR #12459, opened 2018-12-12
Add relocation search in disassembly jump substituion.

n00bSec:
This PR adds relocation lookups for `e asm.jmpsub=true` to give better disassembly using relocations.

While disassembling Pwnable.KR's softmmu kernel module (http://pwnable.kr/bin/softmmu
), the disassembly of `sym.mmuinit` previously failed to substitute in `printk` instead for improperly decoded calls.


n00bSec:
In the same binary there's relocations within `mov` operands as well that resolve to strings present in flags.

PR #12460, opened 2018-12-12
Snow is supported in Panels because it is winter ##panels

Vane11ope:
Useful too to see which panel you are on currently. 
Activate the mode by '(' like the way in visual mode.

Code is still kind of messy but works okay for now. I am on refactoring too so.
gifrecord_2018-12-12_100306


EvanCarroll:
The snow is obviously throwing off the color of the labels though. I'm holding on my breath on Radare3 to finally resolve these rough edges.

Maybe can we get a decent ncurses replacements with opacity too. http://blessedjs.org/

PR #12469, opened 2018-12-12
Fix so -N when no anal info available

kazarmy:


radare:
+1 for killing dupped code if it works :) we probably need more tests

radare:
well you added a test for it :) thanks!

PR #12471, opened 2018-12-13
Set cortex when arm/16 on elf and macho ##bin

radare:


radare:
This is breaking some tests because when asm.cpu=cortex blx results in invalid W T F, probably capstone bug.

```
00f04ae8 blx sym.imp.printf -> invalid
```

PR #12477, opened 2018-12-13
Following-up on meta display from on an offset

devnexen:
this time we display all metadata types.

### DIFF
--- a/libr/anal/meta.c
+++ b/libr/anal/meta.c
@@ -728,31 +728,32 @@ beach:
 }
 
 R_API void r_meta_list_offset(RAnal *a, ut64 addr, char input) {
-	RAnalMetaUserItem ui = { a, 0, 0, NULL, NULL, 0, NULL };
-	SdbList *ls = sdb_foreach_list (DB, true);
-	SdbListIter *lsi;
-	SdbKv *kv;
+	const int types[] = {
+		R_META_TYPE_VARTYPE,
+		R_META_TYPE_HIGHLIGHT,
+		R_META_TYPE_RUN,
+		R_META_TYPE_COMMENT,
+		R_META_TYPE_HIDE,
+		R_META_TYPE_MAGIC,
+		R_META_TYPE_FORMAT,
+		R_META_TYPE_STRING,
+		R_META_TYPE_CODE,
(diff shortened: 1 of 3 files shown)


libr/anal/meta.c
```diff
@@ -727,6 +727,34 @@ static int meta_print_item(void *user, const char *k, const char *v) {
 	return 1;
 }
 
+R_API void r_meta_list_offset(RAnal *a, ut64 addr, char input) {
+	RAnalMetaUserItem ui = { a, 0, 0, NULL, NULL, 0, NULL };
```
radare: ?

libr/anal/meta.c
```diff
@@ -727,6 +727,34 @@ static int meta_print_item(void *user, const char *k, const char *v) {
...
+	SdbList *ls = sdb_foreach_list (DB, true);
+	SdbListIter *lsi;
+	SdbKv *kv;
+
+	ls_foreach (ls, lsi, kv) {
```
radare: this is  very slow. see how the meta is accessed in other functions here, to reduce the scope of iterations it shifts the offset and uses that as a key to resolve the list of values that match a range of offsets.

radare: just do r-meta-find

radare: uhm. chk this:
```

R_API RAnalMetaItem *r_meta_find_in(RAnal *a, ut64 at, int type, int where) {
        char *res = meta_inrange_get (a, at, 1);
```

libr/anal/meta.c
```diff
@@ -747,6 +776,8 @@ R_API int r_meta_list_cb(RAnal *a, int type, int rad, SdbForeachCallback cb, voi
 	isFirst = true; // TODO: kill global
 	ls_foreach (ls, lsi, kv) {
 		if (type == R_META_TYPE_ANY || (strlen (sdbkv_key (kv)) > 5 && sdbkv_key (kv)[5] == type)) {
+			if (strstr(sdbkv_key(kv), "83fc4"))
```
XVilka: This was forgotten to be removed?

devnexen: Yes :-) ....

PR #12494, opened 2018-12-16
Implement support for nword in ahi ##anal

radare:
- ahi0 == ahi
- any digit from 0 to 9 is valid
- This breaks some behaviour in visual mode that needs to be fixed in Vdi

	$ r2 -qcq - << EOF
	wx c7458843c6ff00
	pd 1
	ahi 10
	pd 1
	ahi-
	ahi1 10
	pd 1
	EOF

	mov dword [rbp - 0x78], 0xffc643
	mov dword [rbp - 120], 0xffc643
	mov dword [rbp - 0x78], 16762435


radare:
tests in https://github.com/radare/radare2-regressions/pull/1583

PR #12500, opened 2018-12-17
Fix more build warnings

devnexen:


libr/cons/cons.c
```diff
@@ -138,7 +138,8 @@ static inline void r_cons_write(const char *buf, int len) {
 	if (I.fdout < 1) {
 		I.fdout = 1;
 	}
-	(void) write (I.fdout, buf, len);
+	int r = write (I.fdout, buf, len);
+	(void)r;
```
radare: wtf is this dead code

devnexen: Is to silent warnings ... (void)write ... is not enough in some cases.

devnexen: Fair enough, I was not too proud of it either...

shlr/spp/spp.h
```diff
@@ -5,6 +5,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdarg.h>
+#include <errno.h>
```
radare: this fix should go into the spp repo.. but is this already a warning?

radare:
send a  PR fixing spp repo, we can sync that later into r2

PR #12502, opened 2018-12-17
Support r2core.js for Emscripten >=v1.37.24

iamahuman:
Emscripten v1.37.24 excludes `cwrap` from default exports and requires setting EXTRA_EXPORTED_RUNTIME_METHODS to put it back. Also, it seemed like `malloc` crashed when trying to do `e asm.arch=x86`, so I gave it some more memory (which worked). 33554432 is the least power of 2 that is greater than the default (16777216).

The third commit was because r2 emits the following (kinda ugly) warning every time the core gets created:

    > var r2core = require('./radare2.js');
    > var r2i = r2core._r2_asmjs_new();
    r_config_get: variable 'asm.arch' not found
    anal.arch: cannot find 'wasm'
    asm.arch: cannot find (wasm)
    e asm.bits: Cannot set value, no plugins defined yet
    e asm.bits: Cannot set value, no plugins defined yet
    asm.arch: cannot find (wasm)
    > r2i != 0  // the instance does get successfully created, despite above warnings.
    true

Feel free to omit it if you think it is unnecessary. btw is https://www.npmjs.com/package/r2core no longer maintained?

### DIFF
--- a/binr/radare2/Makefile
+++ b/binr/radare2/Makefile
@@ -20,7 +20,7 @@ ifeq (${COMPILER},emscripten)
 # -s INLINING_LIMIT=1
 # --memory-init-file 0
 CFLAGS+=-s EXPORTED_FUNCTIONS='["_r2_asmjs_new", "_r2_asmjs_cmd", "_r2_asmjs_free", "_r2_asmjs_openurl"]'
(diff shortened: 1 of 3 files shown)


radare:
The wasm “fix” is wrong i fixed it in master

PR #12505, opened 2018-12-17
Fix pd -N start addr to be equal to so -N

kazarmy:


radare:
rebase pls

radare:
looks red  thats a regression

kazarmy:
fixed

PR #12513, opened 2018-12-18
Using libbsd for Linux, for now only for unum api.

devnexen:
Redefining a constant to avoid to mess with meson build handling.


devnexen:
There is a subset of additional functions from BSD (as arc4random family for example).

ret2libc:
> There is a subset of additional functions from BSD (as arc4random family for example). 

And what is exactly that you need from there?

devnexen:
For now using arc4random in the unum api.

radare:
the fallback is to use rand() from libc  which is imho far more than enough.

the only reason to use arc4random is to get better statistical entropy results, also this arc4random is implemented in openssl which we already support as dependency, but we dont use it at all. and in fact openbsd have sitched to chacha20 instead of rc4 for security reasons. so i still dont see much point in overcomplicating the build system with  extra dependencies for something nobody will notice

devnexen:
Alright good point.

PR #12521, opened 2018-12-19
Improve io map lookup performance in presence of many maps ##performance

mrmacete:
- add the “skyline shadow” concept
- add `r_io_map_is_mapped` which binary-searches the skyline shadow to tell if a pointer belongs to any map
- rewrite `r_io_map_get` so that it binary-searches the skyline instead of linearly scanning the whole maps list

The "skyline shadow" is a vector of sorted coalesced intervals taken from the skyline. It's useful in presence of thousands of io maps to efficiently determine if a pointer belongs to any map (without caring about which map). Here's a depiction of this concept:

skyline-shadow

### DIFF
--- a/libr/io/ioutils.c
+++ b/libr/io/ioutils.c
@@ -139,6 +139,9 @@ R_API bool r_io_is_valid_offset(RIO* io, ut64 offset, int hasperm) {
 		return false;
 	}
 	if (io->va) {
+		if (hasperm == 0) {
+			return r_io_map_is_mapped (io, offset);
+		}
 		if ((map = r_io_map_get (io, offset))) {
 			return ((map->perm & hasperm) == hasperm);
 		}
--- a/libr/io/map.c
+++ b/libr/io/map.c
@@ -70,6 +70,7 @@ void io_map_calculate_skyline(RIO *io) {
 	struct map_event_t *ev;
 	bool *deleted = NULL;
 	r_pvector_clear (&io->map_skyline);
+	r_pvector_clear (&io->map_skyline_shadow);
 	r_pvector_init (&events, free);
 	if (!r_pvector_reserve (&events, ls_length (io->maps) * 2) ||
 			!(deleted = calloc (ls_length (io->maps), 1))) {
@@ -139,8 +140,39 @@ void io_map_calculate_skyline(RIO *io) {
 			last_map = map;
 		}
 	}
-
 	r_binheap_clear (&heap);
+
+	const RPVector *skyline = &io->map_skyline;
+	RPVector *shadow = &io->map_skyline_shadow;
+	RInterval *cur_itv = NULL;
+
+	for (i = 0; i < r_pvector_len (skyline); i++) {
+		const RIOMapSkyline *part = r_pvector_at (skyline, i);
+		ut64 part_from = part->itv.addr;
+		ut64 part_size = part->itv.size;
+
+		if (!cur_itv) {
+			cur_itv = R_NEW (RInterval);
+			cur_itv->addr = part_from;
+			cur_itv->size = part_size;
+			continue;
+		}
+
+		if (part_from == r_itv_end (*cur_itv)) {
+			cur_itv->size += part_size;
+		} else {
+			if (!r_pvector_push (shadow, cur_itv)) {
+				R_FREE (cur_itv);
+				break;
+			}
+			cur_itv = NULL;
+		}
+	}
+	if (cur_itv) {
+		if (!r_pvector_push (shadow, cur_itv)) {
+			R_FREE (cur_itv);
+		}
+	}
 out:
 	r_pvector_clear (&events);
 	free (deleted);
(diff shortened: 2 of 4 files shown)


mrmacete:
oops, breaks stuff - i'll dig into that

radare:
We can probably add some memoization and cache invalidation because most reads are done close to the last readed address. But we have been looking for high level optimizations first. Thats why the skyline worked fine to simplify this.

You may probably want to have a look at the search.in=io.maps because we may probably want an io.skyline to avoid searching the same memory twice for example. But for some search commands, permissions matter and thats what broke rop search fex.

Good work

libr/io/map.c
```diff
@@ -139,8 +140,41 @@ void io_map_calculate_skyline(RIO *io) {
 			last_map = map;
 		}
 	}
-
 	r_binheap_clear (&heap);
+
+	const RPVector *skyline = &io->map_skyline;
+	RPVector *shadow = &io->map_skyline_shadow;
+	RInterval *cur_itv = NULL;
```
radare: no need to hold this in the heap. `RInterval cur_itv = {0}`

libr/io/map.c
```diff
@@ -139,8 +140,41 @@ void io_map_calculate_skyline(RIO *io) {
...
+
+	for (i = 0; i < r_pvector_len (skyline); i++) {
+		const RIOMapSkyline *part = r_pvector_at (skyline, i);
```
radare: possible null  deref

libr/io/map.c
```diff
@@ -139,8 +140,41 @@ void io_map_calculate_skyline(RIO *io) {
...
+		ut64 part_from = part->itv.addr;
+		ut64 part_size = part->itv.size;
+
+		if (!cur_itv) {
+			cur_itv = R_NEW (RInterval);
```
radare: null deref. better use r_itv_new()  (which doesnt exist yet. but feel free to add it)

radare: like: cur_itv = r_itv_new (part_from, part_size);

libr/io/map.c
```diff
@@ -139,8 +140,41 @@ void io_map_calculate_skyline(RIO *io) {
...
+			cur_itv->size = part_size;
+		}
+	}
+	if (cur_itv) {
+		if (!r_pvector_push (shadow, cur_itv)) {
+			R_FREE (cur_itv);
+		}
+	}
 out:
 	r_pvector_clear (&events);
```
radare: leaking cur_itv?

libr/io/map.c
radare: ok nvm. you are pushing it into a pvector. so its legit to keep it in the heap

libr/io/map.c
```diff
@@ -305,6 +339,26 @@ R_API RIOMap* r_io_map_get(RIO* io, ut64 addr) {
 	return NULL;
 }
 
+R_API bool r_io_map_is_mapped(RIO* io, ut64 addr) {
+	if (!io) {
+		return false;
+	}
+#define CMP(addr, itv) \
+	(((addr) < r_itv_end (*(RInterval *)(itv))) ? -1 : (((addr) > r_itv_end (*(RInterval *)(itv))) ? 1 : 0))
```
radare: is the cast necessary? isnt *itv enough?

libr/io/map.c
```diff
@@ -305,6 +339,26 @@ R_API RIOMap* r_io_map_get(RIO* io, ut64 addr) {
 	return NULL;
 }
 
+R_API bool r_io_map_is_mapped(RIO* io, ut64 addr) {
+	if (!io) {
```
radare: use r_return_val_if_fail here

libr/io/map.c
mrmacete: why leaking? it's pushed in the vector and it'll be freed when clearing the vector, am i assuming that right?

libr/io/ioutils.c
```diff
@@ -139,6 +139,9 @@ R_API bool r_io_is_valid_offset(RIO* io, ut64 offset, int hasperm) {
 		return false;
 	}
 	if (io->va) {
+		if (hasperm == 0) {
```
radare: if (!hasperm) or ha_sperm

libr/io/map.c
```diff
@@ -305,6 +339,26 @@ R_API RIOMap* r_io_map_get(RIO* io, ut64 addr) {
...
+		return false;
+	}
+#define CMP(addr, itv) \
+	(((addr) < r_itv_end (*(RInterval *)(itv))) ? -1 : (((addr) > r_itv_end (*(RInterval *)(itv))) ? 1 : 0))
+	const RPVector *shadow = &io->map_skyline_shadow;
+	size_t i, len = r_pvector_len (shadow);
+	r_pvector_lower_bound (shadow, addr, i, CMP);
+#undef CMP
```
radare: undef only works for variables, not macros. but i always missed that in CPP :D

libr/io/ioutils.c
mrmacete: LOL

libr/io/map.c
mrmacete: yup, that seems necessary because the compiler complains:
```
map.c:357:42: error: passing 'void' to parameter of incompatible type 'RInterval' (aka 'struct r_interval_t')
        r_pvector_lower_bound (shadow, addr, i, CMP_END_GT);
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
/Users/ftamagni/src/mrmacete-radare2/libr/include/r_vector.h:230:9: note: expanded from macro 'r_pvector_lower_bound'
                        if ((cmp (x, ((void **)(vec)->v.a)[m])) > 0) { \
                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
map.c:15:24: note: expanded from macro 'CMP_END_GT'
        (((addr) < r_itv_end (*(itv)) ? -1 : (((addr) > r_itv_end (*(itv))) ? 1 : 0))
                              ^~~~~~
/Users/ftamagni/src/mrmacete-radare2/libr/include/r_util/r_itv.h:38:40: note: passing argument to parameter 'itv' here
static inline ut64 r_itv_end(RInterval itv) {
```

libr/include/r_util/r_itv.h
```diff
@@ -18,6 +18,15 @@ typedef struct r_interval_t {
...
 typedef RInterval r_itv_t;
 
+static inline RInterval *r_itv_new(ut64 addr, ut64 size) {
+	RInterval *itv = R_NEW (RInterval);
+	if (itv) {
+		itv->addr = addr;
+		itv->size = size;
+	}
+	return itv;
+}
```
radare: add r_itv_free too for consistency

PR #12542, opened 2018-12-22
 meson: Do not use glob with meson

lzutao:
**Note**: This PR should be merge after updating the sdb vendor (After merging radare/sdb#181). 

Fix deprecated warnings in Meson 0.49.0, examples:

```
libr/socket/meson.build:30: DEPRECATION: Library r_socket was passed to the "libraries" keyword argument of a previous call to generate() method instead of first positional argument. Adding r_socket to "Requires" field, but this is a deprecated behaviour that will change in a future version of Meson. Please report the issue if this warning cannot be avoided in your case.
```

### DIFF
--- a/libr/anal/meson.build
+++ b/libr/anal/meson.build
@@ -145,8 +145,7 @@ r_anal = library('r_anal', files,
 r_anal_dep = declare_dependency(link_with: r_anal,
                                 include_directories: r_anal_inc)
 
-pkgconfig_mod.generate(
-  libraries: [r_anal],
+pkgconfig_mod.generate(r_anal,
   subdirs: 'libr',
   version: r2_version,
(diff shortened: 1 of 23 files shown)


lzutao:
**(NOT SOLVED)** Currently, there is an issue (Not quite bad for users, but for developers):
When we changing `meson.build` file and type `ninja`, although we have built r2
successfully before changing, it is still rebuilding everything. I have thought
that using `globbing` in Meson is the problem, but it is not. We should explore it
more carefully

radare:
lgtm, can we merge this?

pelijah:
Ok if you will keep all those file lists up to date.

lzutao:
I think if the meson build job failed, it should be easy to fix.

PR #12546, opened 2018-12-23
Fix #12528 - Removed suffix modifier for 'j' in cmd_print.c

### LINKED ISSUE #12528: pv1j vs pvj1
for consistency it should be pv1j, because j is a suffix. not a modifier

pmb:~ pancake$ r2 -qc'pv1j' /bin/ls
{"value":6215344901283465301,"string":""}
pmb:~ pancake$ r2 -qc'pvj1' /bin/ls
{"value":85,"string":""}

GustavoLCR:
j had a modifier that ignored the one from v, instead use only the one from v. #12528

### DIFF
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -2413,12 +2413,10 @@ static void cmd_print_pv(RCore *core, const char *input, const ut8* block) {
 		input++;
 		break;
 	default:
-		if (input[1] != 'j') {
-			fixed_size = false;
-		} else {
-			n = 8;
+		if (input[1] == 'j') {
 			input++;
 		}
+		fixed_size = false;
 		break;
 	}
 	// variables can be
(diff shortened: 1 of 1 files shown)


libr/core/cmd_print.c
```diff
@@ -2413,7 +2413,12 @@ static void cmd_print_pv(RCore *core, const char *input, const ut8* block) {
 		input++;
 		break;
 	default:
-		fixed_size = false;
+		if (input[1] != 'j') {
+			fixed_size = false;
+		} else {
+			n = 8;
```
radare: Dont assume qwords here. Better use asm.bits/8 if no size is specified

radare:
Hello?

radare:
Can you answer my question?

GustavoLCR:
Sorry, yes. This is to remove pvj4 and make pv4j work

PR #12558, opened 2018-12-24
Fix missing exported symbols

blenk92:
Hi there,
when analyzing shared libraries some of the exported symbols get "lost".  This happens if multiple symbols point to the same location. This patch aims to fix this behaviour. Tests will be added soonish.

Questions | Answers
-- | --
OS/arch/bits (mandatory) | Manjaro x86 64
File format of the file you reverse (mandatory) | ELF
Architecture/bits of the file (mandatory) | x86/64
r2 -v full output, not truncated (mandatory) | radare2 3.2.0-git 20508 @ linux-x86-64 git.3.1.3-139-ge7451c6a0 commit: e7451c6a0b074a4a1602f3ffcac1cc58ed25ecf3 build: 2018-12-24__16:44:50

### Expected behavior
```
$ r2 /usr/lib/libc.so.6
[0x00023af0]> is~getpid
3577 0x000c9420 0x000c9420  LOCAL   FUNC   12 __GI___getpid
3807 0x000c9420 0x000c9420  LOCAL   FUNC   12 __GI_getpid
6267 0x000c9420 0x000c9420   WEAK   FUNC   12 getpid
8102 0x000c9420 0x000c9420 GLOBAL   FUNC   12 __getpid
[0x00024330]> f~getpid
0x000c9420 12 sym.__GI___getpid
0x000c9420 12 sym.__GI_getpid
0x000c9420 12 sym.getpid
0x000c9420 12 sym.__getpid
```
Behaviour shall be similar to readelf:
```
$ readelf -s /usr/lib/libc.so.6 | grep getpid
   581: 00000000000c9420    12 FUNC    WEAK   DEFAULT   14 getpid@@GLIBC_2.2.5
   965: 00000000000c9420    12 FUNC    GLOBAL DEFAULT   14 __getpid@@GLIBC_2.2.5
  3577: 00000000000c9420    12 FUNC    LOCAL  DEFAULT   14 __GI___getpid
  3807: 00000000000c9420    12 FUNC    LOCAL  DEFAULT   14 __GI_getpid
  6267: 00000000000c9420    12 FUNC    WEAK   DEFAULT   14 getpid
  8102: 00000000000c9420    12 FUNC    GLOBAL DEFAULT   14 __getpid
```
However, I don't fully understand why the \@\@GLIBC symbols are not just added to the list of symbols in general, as from my understanding a program might very well link against those symbols. Maybe this could also be clarified here.

### Actual behavior
```
$ r2 /usr/lib/libc.so.6 
[0x00024330]> is~getpid
3577 0x000c9420 0x000c9420  LOCAL   FUNC   12 __getpid
3807 0x000c9420 0x000c9420  LOCAL   FUNC   12 __getpid
6267 0x000c9420 0x000c9420   WEAK   FUNC   12 __getpid
8102 0x000c9420 0x000c9420 GLOBAL   FUNC   12 __getpid
[0x00024330]> f~getpid
0x000c9420 12 sym.__getpid
```

see above

[libc.so.6.zip](https://github.com/radare/radare2/files/2707547/libc.so.6.zip)

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -2962,7 +2962,7 @@ static int Elf_(fix_symbols)(ELFOBJ *bin, int nsym, int type, RBinElfSymbol **sy
 			while (!p->last) {
 				if (p->offset && d->offset == p->offset) {
 					p->in_shdr = true;
-					if (*p->name && strcmp (d->name, p->name)) {
+					if (*p->name && *d->name && (r_str_startswith (d->name, p->name) || r_str_startswith (d->name, "$"))) {
 						strcpy (d->name, p->name);
 					}
 				}


blenk92:
Test failures are caused by tests that check wrong behaviour. I added the PR https://github.com/radare/radare2-regressions/pull/1595 to fix these tests and add a test related to this PR.

radare:
Looks redish

blenk92:
> Looks redish

Yeah it does :-D
This however is imho because two tests are wrong. see https://github.com/radare/radare2-regressions/pull/1595

radare:
This strstr does exactly the same as strncmp

blenk92:
> This strstr does exactly the same as strncmp

Is that a hint to use strncmp instead of strstr?
If so, yeah it is possible and I can fix that, although one would have to determine the length of the search-string first.

Or do you want to say that the strstr statement has no impact because it does exactly what the strcmp already does?
If so, I disagree. The `strcmp (d->name, p->name)` simply checks if d->name and p->name are exactely the same. It will return values < 0 and > 0 but not 0 if the strings differ, with means the strcmp clause will be true as long a both strings are not already the same. Due to that for instance `__getpid` is overwritten with `getpid` (btw this would also be true if both string are the same in the beginning and differ later on)
The `strstr(d->name, p->name) == d->name` is intended to check if d->name contains p->name at index 0. Which means `__getpid` is not overwritte with `getpid`.

radare:
> On 26 Dec 2018, at 00:02, blenk92 <notifications@github.com> wrote:
> 
> This strstr does exactly the same as strncmp
> 
> Is that a hint to use strncmp instead of strstr?
> If so, yeah it is possible and I can fix that, although one would have to determine the length of the search-string first.
> 
> Or do you want to say that the strstr statement has no impact because it does exactly what the strcmp already does?
> If so, I disagree. The strcmp (d->name, p->name) simply checks if d->name and p->name are exactely the same. It will return values < 0 and > 0 but not 0 if the strings differ, with means the strcmp clause will be true as long a both strings are not already the same. Due to that for instance __getpid is overwritten with getpid (btw this would also be true if both string are the same in the beginning and differ later on)
> 
No not this

> The strstr(d->name, p->name) == d->name is intended to check if d->name contains p->name at index 0. Which means __getpid is not overwritte with getpid.
> 

Thats exactly the same as strncmp

radare:
Anyway your pr breaks 2 tests. Did you checked them?

blenk92:
okay, I modified the commit and used strncmp. I hope that is acceptable now...

> Anyway your pr breaks 2 tests. Did you checked them?

Yes. Please have a look at https://github.com/radare/radare2-regressions/pull/1595. That PR fixes the tests and shows that the current behaviour of the tests is wrong ;)

radare:
Use the same branch name for r2 and r2r and travis may be green, otherwise we can test that

libr/bin/format/elf/elf.c
```diff
@@ -2962,7 +2962,8 @@ static int Elf_(fix_symbols)(ELFOBJ *bin, int nsym, int type, RBinElfSymbol **sy
 			while (!p->last) {
 				if (p->offset && d->offset == p->offset) {
 					p->in_shdr = true;
-					if (*p->name && strcmp (d->name, p->name)) {
+					if (*p->name && ((strlen(p->name) != strlen(d->name) && strncmp(d->name, p->name, strlen(p->name)) == 0) ||
+							(d->name && d->name[0] == '$'))) {
```
radare: I dont like this strlenncomparison, also internally this code is doing several more strlen calls which are totally unnecessary also as long as this conditional is becoming a bit complex move that into a static function with proper name.

Also always add a space before the (

Another issue i see is that you are checking for d->name not being null after dereferencing it

You can use r_str_startswith instead of that strncmo+strlen

blenk92:
> Use the same branch name for r2 and r2r and travis may be green, otherwise we can test that

i've created a new pull request using the same branch name. You can find it here https://github.com/radare/radare2-regressions/pull/1597

PR #12560, opened 2018-12-25
Optimize ds_align_comment() ##disasm

thestr4ng3r:
`ds_align_comment()` is currently the bottleneck for `agJ`. The idea here is to save the position in the cons buffer of the start of the current line and calculate the line length starting from there, excluding ansi escape sequences and respecting utf-8.

## Comparison
Before:
```
[florian@florian-desktop radare2]$ time r2 /bin/ls -qc "s main;af;agJ" > /dev/null

real	0m0,440s
user	0m0,428s
sys	0m0,012s
```

After:
```
[florian@florian-desktop radare2]$ time r2 /bin/ls -qc "s main;af;agJ" > /dev/null

real	0m0,153s
user	0m0,139s
sys	0m0,013s
```

Here is the comparison of a seek in cutter from outside a function to inside a function, which causes a call to `agJ`:
Before:
bildschirmfoto vom 2018-12-25 12-50-25

After:
bildschirmfoto vom 2018-12-25 12-50-42

### DIFF
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -588,6 +588,10 @@ R_API const char *r_cons_get_buffer() {
 	return I.context->buffer_len? I.context->buffer : NULL;
 }
 
+R_API int r_cons_get_buffer_len() {
+	return I.context->buffer_len;
+}
+
 R_API void r_cons_filter() {
 	/* grep */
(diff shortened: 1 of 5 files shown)


radare:
💥

radare:
Good optimization

PR #12568, opened 2018-12-27
Fix no-anal asm.bbline when there is backwards ref to and ref from addr

kazarmy:


radare:
this was ##disasm

radare:
or maybe ##glitch

kazarmy:
ok ... i think ##disasm is better since it shows affected component

PR #12571, opened 2018-12-27
Fix no-anal asm.bbline when there is only backwards ref from addr

kazarmy:


libr/anal/reflines.c
```diff
@@ -403,7 +403,8 @@ R_API char* r_anal_reflines_str(void *_core, ut64 addr, int opts) {
 			r_list_free (lvls);
 			return NULL;
 		}
-		if (in_refline (addr, ref) && !(middle_after && ref->to == addr && ref->direction >= 0)) {
+		if (in_refline (addr, ref) && !(middle_after && ((ref->to == addr && ref->direction >= 0)
```
radare: seems like this conditional can be improved by  moving the logic into a separate function, and making the ref->to/from depend on ref->direction instead of comparing both cases. but its fine to merge it like this for now

PR #12574, opened 2018-12-28
Use refline_kept() in r_anal_reflines_str() instead

kazarmy:
For https://github.com/radare/radare2/pull/12571#discussion_r244233783.


radare:
why appveyour red

Issue #12474, opened 2018-12-13
rabin2 -qqzzz /bin/ls doesnt work


radare:

Maijin:
<img width="400" alt="image" src="https://user-images.githubusercontent.com/1408600/49938324-d766d880-fed9-11e8-84f1-6ea84373c0ed.png">
Wut?

radare:
uhm maybe it was the other binary..

i found another issue:

$ rabin2 -qqzzz /bin/ls

radare:
ping @AdamVanScyoc

AdamVanScyoc:
Disregard last commit; still working on it.

radare:
https://github.com/radare/radare2/pull/12653 <-

PR #12578, opened 2018-12-29
Add handler for simplest mode #12474

AdamVanScyoc:
From `rabin2 --help`:
` -zzz            dump raw strings to stdout (for huge files)`


radare:
looks red, please rebase, and chk the 3 xx tests
screenshot 2018-12-29 at 11 18 43

AdamVanScyoc:
I made some changes to the code (https://github.com/AdamVanScyoc/radare2/commit/10f96ceb5a94bac3c1e0a5d93768e77801129262#diff-f790c0bb172c20a97e46d7735fc12236) and it now should be dumping directly to stdout exactly as radare2-regressions expects.

When I run `r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N bins/elf/analysis/hello-linux-x86_64` and at the t2 shell type e.g. `izzz~?` it give the predicted output: in this case, `0`.

But when I run the radare2-regressions, theses tests still fail with the outputs from before I made these latest changes (i.e. `82`). Any ideas?

radare:
dont merge, better rebase otherwise the commit history gets fuckedup and the patch files are messy

libr/anal/class.c
```diff
@@ -613,7 +613,6 @@ R_API RAnalClassErr r_anal_class_base_get(RAnal *anal, const char *class_name, c
 	cur = next;
 	if (!cur) {
 		free (content);
-		free (base->class_name);
```
radare: regressions because of merge

libr/bin/bfile.c
```diff
@@ -59,6 +59,9 @@ static void print_string(RBinFile *bf, RBinString *string) {
 		free (f_name);
 		break;
 	}
+	case R_MODE_SIMPLEST:
+		printf ("%s\n", string->string);
```
radare: never use printf in r2land. use r_cons_printf instead

PR #12594, opened 2018-12-30
Fix infinite loop bug related to anal.nopskip

radare:


radare:
This change the behaviour of the analysis when starting to analyze a function in a code like this, was resulting in infinite loop. anal.hasnext=true seems to be working still fine, but some functions may not be analyzed. lets just trust travis. we need to add a test for this

PR #12595, opened 2018-12-31
Minor code cleanups/fixes

dav1901:


shlr/java/class.c
```diff
@@ -8376,42 +8388,48 @@ R_API char *r_bin_java_resolve_b64_encode(RBinJavaObj *BIN_OBJ, ut16 idx) {
...
 		str = malloc (34);
-		out = malloc (34);
-		memset (out, 0, 34);
 		if (str) {
 			snprintf (str, 34, "0x%"PFMT64x, r_bin_java_raw_to_long (item->info.cp_long.bytes.raw, 0));
-			r_base64_encode (out, (const ut8 *) str, strlen (str));
+			out = malloc (34);
+			if (out) {
+				memset (out, 0, 34);
+				r_base64_encode (out, (const ut8 *) str, strlen (str));
```
radare: use r_base64_encode_dyn() which does the malloc internally. the memset is unnecessary

radare: (because it can be done with just a calloc). but use the api

shlr/java/class.c
```diff
@@ -8376,42 +8388,48 @@ R_API char *r_bin_java_resolve_b64_encode(RBinJavaObj *BIN_OBJ, ut16 idx) {
...
 		str = malloc (1000);
-		out = malloc (1000);
-		memset (out, 0, 1000);
 		if (str) {
 			snprintf (str, 1000, "%f", r_bin_java_raw_to_double (item->info.cp_double.bytes.raw, 0));
-			r_base64_encode (out, (const ut8 *) str, strlen (str));
+			out = malloc (1000);
+			if (out) {
+				memset (out, 0, 1000);
+				r_base64_encode (out, (const ut8 *) str, strlen (str));
```
radare: same here

shlr/java/class.c
dav1901: done

PR #12616, opened 2019-01-03
Deprecate ib - seems very related to oob and its actually buggy ##bin

radare:
- 2ib = segfault


radare:
Seems like it's still used.. so we may  probably fix the crash first..

screenshot 2019-01-03 at 12 37 00

PR #12617, opened 2019-01-03
Initial WIP implementation of sync using http and T ##sync

radare:


radare:
There's a lot more work to be done, but that's just a PoC, and fits within r2

XVilka:
For the reference - this was the original issue: https://github.com/radare/radare2/issues/7410

libr/core/cmd_log.c
```diff
@@ -180,15 +250,52 @@ static int cmd_log(void *data, const char *input) {
...
 		break;
-	case 'T':
+	case 'T':  // Ts ? as ms?
 		if (r_config_get_i (core->config, "scr.interactive")) {
 			textlog_chat (core);
 		} else {
 			eprintf ("Only available when the screen is interactive\n");
 		}
 		break;
+	case '=':
```
ret2libc: please add `// T=` just to help grepping it

libr/core/cmd_log.c
```diff
@@ -180,15 +250,52 @@ static int cmd_log(void *data, const char *input) {
...
+				}
+				r_cons_break_pop ();
+			} else {
+				// TODO: Sucks that we can't enqueue functions, only commands
+				eprintf ("Background thread syncing with http.sync started.\n");
+				RCoreTask *task = r_core_task_new (core, true, "T=&&", NULL, core);
+				r_core_task_enqueue (core, task);
+			}
+		} else {
+			if (atoi (input + 1) > 0 || (input[1] == '0')) {
```
ret2libc: how is `T=` supposed to work? Does it expect a number as parameter? It's not clear from the help msg.

radare:
this pr can become quite big.. because it dpeends other features and changes.. so i dont wanna make it much bigger. so pls review this partial impl and ill continue with separate PRs

libr/core/cmd_log.c
radare: yes, it expects a number but it's optional, in that case it pulls from the last id retrieved.

PR #12618, opened 2019-01-03
Fix #12498 - Handle ^C in agf

### LINKED ISSUE #12498: agf is very slow and cant be stopped
try r2 -c 'af;agf' /bin/ls

radare:


radare:
Solved in  
https://github.com/radare/radare2/pull/12618

PR #12624, opened 2019-01-03
Implement /s to search for sections depending on entropy ##search

radare:


radare:
We want /s* too to flag the regions

Issue #12393, opened 2018-12-04
aa subcommand changes


radare:
* afta -> aaft
* aftm -> aft
* aaE -> aafe
* aad -> is never called

radare:
pls suggest more changes to a subcommands in this issue

radare:
https://github.com/radare/radare2/pull/12628

PR #12631, opened 2019-01-03
Rename aftm -> aft and afta -> aaft

dav1901:
#12393.


radare:
This change only makes sense if the `afta -> aaft` is done too. afta is pretty popular out there, thas why i would like to get feedback, but go for it and i'll merge if no one complains and travis gets green (some tests will not be passing)

dav1901:
travis should fail, will fix tests later.

PR #12636, opened 2019-01-04
 Fix #12623 - Make Grep Highlights Case Insensitive

### LINKED ISSUE #12623: scr.highlight should be case insensitive
main problem is that we are using r_str_replace() which is case sensitive and we dont have an insensitive api to do that. it's easy to fix, but not done

image 2019-01-03 13 41 36

GustavoLCR:


libr/cons/grep.c
```diff
@@ -589,13 +589,18 @@ R_API int r_cons_grepbuf(char *buf, int len) {
...
 					if (cons->grep_highlight && grep->str) {
 						char *str = r_str_ndup (tline, ret);
-						char *newstr = r_str_newf (Color_INVERT"%s"Color_RESET, grep->str);
+						char *grepstr = r_str_ndup (tline, ret);
+						char *old_grepstr = grepstr;
+						grepstr = (char *) r_str_casestr (grepstr, grep->str);
+						grepstr[strlen (grep->str)] = '\0';
+						char *newstr = r_str_newf (Color_INVERT"%s"Color_RESET, grepstr);
 						if (str && newstr) {
-							str = r_str_replace (str, grep->str, newstr, 1);
```
radare: better implement r_str_replace_icase() and put all the logic inside this function

libr/cons/grep.c
```diff
@@ -589,13 +589,18 @@ R_API int r_cons_grepbuf(char *buf, int len) {
 				if (show) {
 					if (cons->grep_highlight && grep->str) {
 						char *str = r_str_ndup (tline, ret);
-						char *newstr = r_str_newf (Color_INVERT"%s"Color_RESET, grep->str);
+						char *grepstr = r_str_ndup (tline, ret);
```
radare: grepstr can be null

libr/cons/grep.c
```diff
@@ -589,13 +589,18 @@ R_API int r_cons_grepbuf(char *buf, int len) {
...
+						char *old_grepstr = grepstr;
+						grepstr = (char *) r_str_casestr (grepstr, grep->str);
```
radare: you must look to find all the matches, not just the 1st one

libr/cons/grep.c
```diff
@@ -589,13 +589,18 @@ R_API int r_cons_grepbuf(char *buf, int len) {
...
+						grepstr[strlen (grep->str)] = '\0';
+						char *newstr = r_str_newf (Color_INVERT"%s"Color_RESET, grepstr);
 						if (str && newstr) {
-							str = r_str_replace (str, grep->str, newstr, 1);
+							str = r_str_replace (str, grepstr, newstr, 1);
```
radare: that will make the string lowercase, not just the word to replace

libr/util/str.c
```diff
@@ -979,6 +979,59 @@ R_API char* r_str_replace(char *str, const char *key, const char *val, int g) {
...
+	slen = strlen (str);
+	for (i = 0; i < slen;) {
+		p = (char *)r_str_casestr (str + i, key);
+		if (!p) {
+			break;
+		}
+		off = (int)(size_t) (p - str);
+		scnd = strdup (p + klen);
+		slen += vlen - klen;
+		// HACK: this 32 avoids overwrites wtf
```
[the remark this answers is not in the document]
GustavoLCR: This came from a copy+paste from r_str_replace(). Will remove

libr/util/str.c
```diff
@@ -979,6 +979,59 @@ R_API char* r_str_replace(char *str, const char *key, const char *val, int g) {
...
+			free (scnd);
+			break;
+		}
+		str = newstr;
+		p = str + off;
+
+		if (keep_case) {
+			char *str_case = r_str_ndup (p, klen);
+			tmp_val = strdup (val);
+			str_case[klen] = '\0';
```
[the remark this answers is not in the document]
GustavoLCR: It will only call itself once, so it shouldn't be a problem.

PR #12639, opened 2019-01-04
Add free_all to RVector ##util

thestr4ng3r:
The idea is that you can do things like having a single buffer like the following:
`"hello\0r2\0is\0cool\0"`
Then saving pointers to each individual string in an RPVector and freeing the buffer once automatically when the RPVector is freed. This is especially useful when fetching sdb arrays, where you can just allocate the whole string once and set each ',' to zero.

Tests: https://github.com/radare/radare2-regressions/pull/1620

### DIFF
--- a/libr/include/r_vector.h
+++ b/libr/include/r_vector.h
@@ -38,6 +38,7 @@ extern "C" {
 
 typedef int (*RPVectorComparator)(const void *a, const void *b);
 typedef void (*RVectorFree)(void *e, void *user);
+typedef void (*RVectorFreeAll)(void *user);
 typedef void (*RPVectorFree)(void *e);
 
 typedef struct r_vector_t {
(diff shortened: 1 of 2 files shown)


ret2libc:
Is it really something we want in RVector? I mean, I would create a new structure RVectorSdb or similar and in the r_vector_sdb_free I would call both r_vector_free and free(buf). Does that make sense?

ret2libc:
I agree I don't particularly like the change either...

thestr4ng3r:
Yeah we could do something like
```
typedef struct {
  RPVector v;
  void *buf;
} RBufVector;
```
Having a callback for the free all function would have the advantage that we can be more flexible about how to free the global buffer. But just from the top of my head I can not really think of any actual practical uses for this other than a buffer with strings.

PR #12642, opened 2019-01-04
Fix #12434 - Fixes for exports and anal behaviour

### LINKED ISSUE #12434: name section not handled in wasm
[main.zip](https://github.com/radare/radare2/files/2659143/main.zip)

```
$ cat main.c
#include <emscripten.h>
#include <stdbool.h>
#include <stdint.h>

EMSCRIPTEN_KEEPALIVE
bool getTrue() {
   return true;
}

EMSCRIPTEN_KEEPALIVE
bool getFalse() {
   return false;
}
$ emcc -s WASM=1 -o main.js -Oz -g2 main.c
```
and then:

```
$ rabin2 -s main.wasm
[wasm] error: beach symtab entries
[Symbols]
Num Paddr      Vaddr      Bind     Type Size Name
000 0x00000024 0x00000024   NONE   FUNC    3 fcn.0
001 0x00000029 0x00000029   NONE   FUNC    3 fcn.1
$ wasm-objdump -d main.wasm

main.wasm:	file format wasm 0x1

Code Disassembly:

000022 <_getTrue>:
 000024: 41 01                      | i32.const 1
 000026: 0b                         | end
000027 <_getFalse>:
 000029: 41 00                      | i32.const 0
 00002b: 0b                         | end
$
```

wargio:
Fixed loop behaviour (now jumps to correct ends, etc..)
Fixed Exports (aka symbols)


libr/anal/p/anal_wasm.c
```diff
@@ -34,7 +42,22 @@ static ut64 get_cf_offset(RAnal *anal, const ut8 *data) {
 	return UT64_MAX;
 }
 
-static ut64 find_if_else(ut64 addr, const ut8 *data, int len, bool is_loop) {
+/**
+ * Searches where the opcode ends.
```
radare: omg comments

wargio: People asked them, so i added them.

radare:
imho comments should be in public apis only  and in functions that the function name is not making clear what they do. also documenting functions implies that we need a tool to write navigable documented code, and this was suposed to be done in the vapis like i did 8 years ago and nobody cared. also  i prefer one  line  comments instead of java style ones

wargio:
ok, i'll remove them :>

PR #12647, opened 2019-01-04
Add RBufVector

thestr4ng3r:


thestr4ng3r:
ping @radare @ret2libc

thestr4ng3r:
Same as the free_all thing would have been.
You can have a single big buffer and let the vector elements point to substrings of it.

libr/util/vector.c
```diff
@@ -297,3 +297,23 @@ static void quick_sort(void **a, size_t n, RPVectorComparator cmp) {
 R_API void r_pvector_sort(RPVector *vec, RPVectorComparator cmp) {
 	quick_sort (vec->v.a, vec->v.len, cmp);
 }
+
```
ret2libc: no need 2 spaces

thestr4ng3r:
Yes, I have a place where I need it. I can show you after I have implemented it.

radare:
why not just add void*user in RVector?

thestr4ng3r:
That would be the same as in https://github.com/radare/radare2/pull/12639, but without the extra free function. So what solution do you want?

thestr4ng3r:
`r_anal_class_get_all()` is the use case and here is a test for it: https://github.com/radare/radare2-regressions/pull/1623/files#diff-dc2281e83ab9f673093d0172b8435e2c

thestr4ng3r:
ping @radare @ret2libc

radare:
r_anal_class_get_all is reachable?

thestr4ng3r:
`r_anal_class_get_all()` is defined as `R_API` and declared in the header, so it is reachable. Or what do you mean?

radare:
It is not called by anyone else

thestr4ng3r:
The test does: https://github.com/radare/radare2-regressions/pull/1623/files#diff-dc2281e83ab9f673093d0172b8435e2c
But I primarily plan to call it from Cutter.

libr/anal/class.c
```diff
@@ -127,6 +127,25 @@ R_API bool r_anal_class_exists(RAnal *anal, const char *name) {
 	return r;
 }
 
+R_API RBufVector/*<const char *>*/ *r_anal_class_get_all(RAnal *anal) {
```
radare: this seems not very efficient, sdb can return a list and you can convert it to vector for faster nth access in cutter. in r2land i dont see the need to have such access need.

PR #12649, opened 2019-01-04
WIP: Refactoring of get_import_addr in format/elf

ret2libc:
Before the patch, the relocations in .rel.plt (or the others) were
traversed every time an imported symbol needed to be found. With the
refactoring, relocations are parsed only one time, placed in a hashtable
for fast lookup and stored in a cache.

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -25,6 +25,8 @@
 
 #define bprintf if(bin->verbose) R_LOG_WARN
 
+#define MAX_REL_RELA_SZ (sizeof (Elf_(Rel)) > sizeof (Elf_(Rela))? sizeof (Elf_(Rel)): sizeof (Elf_(Rela)))
+
 #define READ8(x, i) r_read_ble8((x) + (i)); (i) += 1
 #define READ16(x, i) r_read_ble16((x) + (i), bin->endian); (i) += 2
 #define READ32(x, i) r_read_ble32((x) + (i), bin->endian); (i) += 4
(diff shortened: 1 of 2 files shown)


libr/bin/format/elf/elf.c
```diff
@@ -1236,314 +1238,349 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
+			if (len < 4) {
+				goto out;
+			}
+			if (bin->endian) {
+				ut64 base = r_read_be32 (buf);
+				base -= (nrel * 16);
+				base += (rel->k * 16);
+				plt_addr = base;
+			} else {
+				// FIXME: this does not seem to work as
```
ret2libc: this needs to be fixed somehow, but I know nothing about PPC64. However, as mentioned in https://github.com/radare/radare2/issues/9825#issuecomment-451498683 it seems this was introduced to fix something that is actually not fixed, so...

ret2libc:
As expected, the only test failing is `db/formats/elf/elf-mosquitoes`, which is anyway already wrong. I would mark it as broken and let somebody else with experience in PPC to fix it.

ret2libc:
See https://travis-ci.org/radare/radare2/jobs/475481322#L2500

libr/bin/format/elf/elf.c
```diff
@@ -1236,314 +1238,352 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
+				plt_addr = base;
+			}
+			return plt_addr;
+		}
+		break;
+	}
+	case EM_SPARC:
+	case EM_SPARCV9:
+	case EM_SPARC32PLUS:
+		plt_addr = Elf_(r_bin_elf_get_section_addr) (bin, ".plt");
```
radare: isnt this the same as plt_section->offset ? no need to find twice

libr/bin/format/elf/elf.c
```diff
@@ -1236,314 +1238,352 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
+			return -1;
+		}
+		if (REL_TYPE == R_386_PC16) {
+			plt_addr += rel->k * 12 + 20;
+			// thumb symbol
+			if (plt_addr & 1) {
+				plt_addr--;
+			}
+			return plt_addr;
+		} else {
```
radare: no need to  else after return

libr/bin/format/elf/elf.c
```diff
@@ -1236,314 +1238,352 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
+		case R_ARM_JUMP_SLOT: {
+			plt_addr += rel->k * 12 + 20;
+			// thumb symbol
+			if (plt_addr & 1) {
+				plt_addr--;
+			}
+			return plt_addr;
+			break;
+		}
+		case 1026: // arm64 aarch64
```
radare: add constant with R_ARM64_JUMP_SLOT name?

libr/bin/format/elf/elf.c
radare: ppc-linux is a bit weird in many aspects xD

radare:
screenshot 2019-01-04 at 23 49 14

radare:
ok, mark the ppc test as broken and ill merge, pls rebase

libr/bin/format/elf/elf.c
ret2libc: yeah :/ And I really don't know the architecture at all :(

ret2libc:
r2r PR that mark the PPC mosquito test as broken (it was already wrong anyway..) https://github.com/radare/radare2-regressions/pull/1633

PR #12665, opened 2019-01-05
Add basic authorization for the Web interface.

devnexen:
- For post 3.2 only.
- For the moment no user DB but just one from the config.
- Basic auth with all its known little culprits,
credentials carried for the whole session and so on ...
but that might suffice for a start.

### DIFF
--- a/binr/r2agent/r2agent.c
+++ b/binr/r2agent/r2agent.c
@@ -29,7 +29,7 @@ static int usage (int v) {
 	"  -h        show this help message\n"
 	"  -s        run in sandbox mode\n"
 	"  -u        enable http Authorization access\n"
-	"  -t        user:password token\n"
+	"  -t        user:password authentification file\n"
 	"  -p [port] specify listening port (defaults to 8080)\n");
 	return !v;
 }
@@ -41,14 +41,16 @@ static int showversion() {
 
 int main(int argc, char **argv) {
 	RSocket *s;
+	RSocketHTTPOptions so;
 	RSocketHTTPRequest *rs;
-	int c, timeout = 3;
+	int c;
 	int dodaemon = 0;
 	int dosandbox = 0;
 	bool listenlocal = true;
 	const char *port = "8080";
-	bool httpauth = false;
-	const char *authtoken = NULL;
+	const char *httpauthfile = NULL;
+	char *pfile = NULL;
+	memset (&so, 0, sizeof (so));
 
 	while ((c = getopt (argc, argv, "adhup:t:sv")) != -1) {
 		switch (c) {
@@ -66,10 +68,10 @@ int main(int argc, char **argv) {
 		case 'v':
 			return showversion ();
 		case 'u':
-			httpauth = true;
+			so.httpauth = true;
 			break;
 		case 't':
-			authtoken = optarg;
+			httpauthfile = optarg;
 			break;
 		case 'p':
 			port = optarg;
@@ -82,9 +84,21 @@ int main(int argc, char **argv) {
 		return usage (0);
 	}
 
-	if (httpauth && authtoken) {
-		if (strlen (authtoken) < 3 || !strchr (authtoken, ':')) {
-			eprintf ("Invalid authorization token '%s'\n", authtoken);
+	so.accept_timeout = 0;
+	so.timeout = 3;
+
+	if (so.httpauth) {
+		if (!httpauthfile) {
+			eprintf ("No authentification user list set\n");
+			return usage (0);
+		}
+
+		int sz;
+		pfile = r_file_slurp (httpauthfile, &sz);
+		if (pfile) {
+			so.authtokens = r_str_split_list (pfile, "\n");
+		} else {
+			eprintf ("Empty list of HTTP users\\n");
 			return usage (0);
 		}
 	}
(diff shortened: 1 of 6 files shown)


devnexen:
@radare nothing urgent when you have time :-)

radare:
yay, wasnt expecting a PR that quickly! will look at this after the release :) im focusing on stability for 2 days

libr/socket/http_server.c
```diff
@@ -72,6 +74,23 @@ R_API RSocketHTTPRequest *r_socket_http_accept (RSocket *s, int accept_timeout,
 			} else
 			if (!strncmp (buf, "Content-Length: ", 16)) {
 				content_length = atoi (buf+16);
+			} else
+			if (!strncmp (buf, "Authorization: Basic ", 21) && httpauth) {
+				char *authtoken = buf + 21;
+				size_t authlen = strlen (authtoken);
+				char *decauthtoken = calloc (4, authlen + 1);
```
radare: check for null

libr/socket/http_server.c
```diff
@@ -72,6 +74,23 @@ R_API RSocketHTTPRequest *r_socket_http_accept (RSocket *s, int accept_timeout,
 			} else
 			if (!strncmp (buf, "Content-Length: ", 16)) {
 				content_length = atoi (buf+16);
+			} else
+			if (!strncmp (buf, "Authorization: Basic ", 21) && httpauth) {
```
radare: use r_str_startswith()

libr/socket/http_server.c
```diff
@@ -72,6 +74,23 @@ R_API RSocketHTTPRequest *r_socket_http_accept (RSocket *s, int accept_timeout,
 			} else
 			if (!strncmp (buf, "Content-Length: ", 16)) {
 				content_length = atoi (buf+16);
+			} else
```
radare: if and else in the same line

libr/socket/http_server.c
```diff
@@ -72,6 +74,23 @@ R_API RSocketHTTPRequest *r_socket_http_accept (RSocket *s, int accept_timeout,
 			} else
 			if (!strncmp (buf, "Content-Length: ", 16)) {
 				content_length = atoi (buf+16);
```
radare: space around +

libr/socket/http_server.c
```diff
@@ -72,6 +74,23 @@ R_API RSocketHTTPRequest *r_socket_http_accept (RSocket *s, int accept_timeout,
...
+			if (!strncmp (buf, "Authorization: Basic ", 21) && httpauth) {
+				char *authtoken = buf + 21;
+				size_t authlen = strlen (authtoken);
+				char *decauthtoken = calloc (4, authlen + 1);
+				if (r_base64_decode ((ut8 *)decauthtoken, authtoken, authlen) == -1) {
+					eprintf ("Could not decode authorization token\n");
+					free (decauthtoken);
+				} else {
+					if (!strcmp (decauthtoken, expauthtoken)) {
+						hr->auth = 1;
```
radare: use bool?

binr/r2agent/r2agent.c
```diff
@@ -104,10 +120,13 @@ int main(int argc, char **argv) {
 		char *result_heap = NULL;
 		const char *result = page_index;
 
-		rs = r_socket_http_accept (s, 0, timeout);
+		rs = r_socket_http_accept (s, 0, timeout, httpauth, authtoken);
```
radare: i think it will be better to refactor this into RSocketHttpOptions struct and put the timeout, and other configuration options in there.

or maybe we can just add a callback which is used to process the authentication check.

radare: otherwise we cant have more than one user, also will be good to define privilege levels to different users and make r2 multiuser

libr/core/cconfig.c
```diff
@@ -3018,6 +3018,8 @@ R_API int r_core_config_init(RCore *core) {
 	SETPREF ("http.upget", "false", "/up/ answers GET requests, in addition to POST");
 	SETPREF ("http.upload", "false", "Enable file uploads to /up/<filename>");
 	SETPREF ("http.uri", "", "Address of HTTP proxy");
+	SETPREF ("http.auth", "false", "Enable/Disable HTTP Authentification");
+	SETPREF ("http.authtok", "r2admin:r2admin", "HTTP Authentification user:password token");
```
radare: what about http.auth.file or cmd.httpauth?

radare:
fix conflicts after rebase

binr/r2agent/r2agent.c
devnexen: That s true it s better having multi user even though as stated if the original PR comment, it was purposely mono-user as a "first step" but will try to come up with something along ...

libr/core/cconfig.c
devnexen: Yes in multi user context having a file makes more sense. If you insist I ll make the necessary change for multi user usage (see original PR comment).

libr/socket/http_server.c
devnexen: Fair point ... even though your solution need your suggestion implies calculating a known constant length (and I was following the general state of the rest of the workflow).

libr/socket/http_server.c
devnexen: In fact i did for style consistency sake even though this particular part of the code is a bit inconsistent with r2 as a whole. I may then refactor a bit the even unchanged parts then ...

devnexen:
Seems a little issue with the darwin environment ...

radare:
I don't think it's WIP anymore, lgtm

PR #12669, opened 2019-01-06
Meson: Update capstone

pelijah:
Related: #12659


radare:
cant we have a single place t odefine this for make and meson?

PR #12683, opened 2019-01-07
Initial move of RCons->pal into RCons->context->pal to be task safe

radare:


radare:
its still crashing in other places, and there are many more issues:

* cons->context should not be resolved multiple times because that can result on the wrong context resolution
* r_cons_singleton()  should be done only once, and try to use core->cons if p ossible
* See other vars from  RCons that are not saved in context that can result in crash

radare:
more work on https://github.com/radare/radare2/pull/12815

PR #12694, opened 2019-01-08
Fix relocations in ET_REL ELF

ret2libc:
Fix a bunch of things on ET_REL ELF:
- when a section symbol is found, get the name of the section, instead of just print empty stuff like readelf does
- do not print warning messages (e.g. no program headers, no dyn segment, etc.) when the type is ET_REL, because it's a regular thing for that kind of files to not have those things
- fix relocation patching for type R_X86_64_PLT32 (there are still some weird stuff there, wrt to that fake .got.r2 section)


ret2libc:
Before:
```
[0x08000040]> ir
[Relocations]
vaddr=0x08000088 paddr=0x0000001a type=SET_32
vaddr=0x08000309 paddr=0x00000024 type=SET_32 printf - 0x08000068
vaddr=0x08000088 paddr=0x00000033 type=SET_32 0x00000012
vaddr=0x08000309 paddr=0x0000003d type=SET_32 printf - 0x08000081
vaddr=0x08000040 paddr=0x00000020 type=SET_32 - 0x080000f8

5 relocations

[0x08000063]> pdi 3 @ 0x08000059
0x08000059           bf88000008  mov edi, str.Good_morning__s
0x0800005e           b800000000  mov eax, 0
0x08000063           e8b5020000  call 0x800031d
```
After:
```
[0x08000040]> ir
[Relocations]
vaddr=0x08000088 paddr=0x0000001a type=SET_32 .rodata
vaddr=0x08000309 paddr=0x00000024 type=SET_32 printf - 0x08000068
vaddr=0x08000088 paddr=0x00000033 type=SET_32 .rodata + 0x00000012
vaddr=0x08000309 paddr=0x0000003d type=SET_32 printf - 0x08000081
vaddr=0x08000040 paddr=0x00000020 type=SET_32 .text - 0x080000f8

5 relocations

[0x08000040]> pdi 3 @ 0x08000059
0x08000059           bf88000008  mov edi, str.Good_morning__s
0x0800005e           b800000000  mov eax, 0
0x08000063           e8a1020000  call reloc.printf
```

TODO in separate PR: fix the addend of the relocations, which are still wrong...

ret2libc:
Ok maybe the solution I implemented is not the best because it adds a new symbol for each section listed in the symbol table (which, in theory, would be the right thing to do, but since we already have flags like `section..text` it's not ok). Marking this as WIP.

ret2libc:
Actually I wouldn't know how to solve this otherwise... I think the main problem (that we also have in many other places) is that when there are multiple flags in the same address, a random one is selected. Shall we introduce a sort of "priority" for flags? User defined symbols always have higher priority, while the one that we set in the code can have a priority that we define manually (e.g. section flags would have a priority of 99, relocs flags 80, string flags 50 and symbol flags 30 (random numbers :D ))

radare:
ping conflicts

ret2libc:
Yeah let's not care about this for some time, I need to fix some things in flags before fixing this.

ret2libc:
this messes up with all the tests because some flags get the priority over others (thus the flags priority issue i want to solve first). Because it does not have any sense that just because i'm adding some symbols (that should be rightfully added since they are in the symbol table, even if they are symbol's that reference the sections) radare2 gives a different name to functions or other stuff.

radare:
please rebase

ret2libc:
I need https://github.com/radare/radare2/pull/13138 to continue this.

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1731

ret2libc:
@radare please review when you can. I think this is ready. See the r2r PR (https://github.com/radare/radare2-regressions/pull/1731/files) to quickly understand what's changed. Also, see https://github.com/radare/radare2/pull/12694#issue-242929523 for an explanation of the fixes.

radare:
conflicts, pls rebase and ill review + merge

PR #12698, opened 2019-01-08
WIP: Improvements and code cleanup in util/str ##core

radare:


radare:
there's a glitch that some lines are not fully written when terminal is small, even if its green is not perfect

kazarmy:
fixed by #12838

PR #12699, opened 2019-01-08
Improve rap:// code ##io

radare:


radare:
im starting to not be sure about this change because it implies a lot of changes in the documentation and locks to 'r2' when the protocol is pretty generic.

PR #12702, opened 2019-01-08
Make afcf also print args from vars (unified with disasm)

thestr4ng3r:
Please don't squash. I think the commits are clean enough.

Example:
```
[0x00005ae0]> s main
[0x00004070]> af
[0x00004070]> pd 1
┌ (fcn) main 4886
│   main (int argc, char **argv, char **envp);
│           ; arg int arg_5h @ rbp+0x5
│           ; var int local_8h @ rsp+0x8
│           ; var int local_17h @ rsp+0x17
│           ; var int local_28h @ rsp+0x28
│           ; var int local_30h @ rsp+0x30
│           ; var int local_32h @ rsp+0x32
│           ; var int local_38h @ rsp+0x38
│           ; var int local_45h @ rsp+0x45
│           ; var int local_46h @ rsp+0x46
│           ; var int local_47h @ rsp+0x47
│           ; var int local_48h @ rsp+0x48
│           ; arg int argc @ rdi
│           ; arg char **argv @ rsi
│           ; DATA XREF from main (0x5b01)
│           0x00004070      4157           push r15
[0x00004070]> afcf
int main (int argc, char **argv, char **envp);
[0x00004070]> afcfj
[{"name":"main","return":"int","args":[{"name":"argc","type":"int"},{"name":"argv","type":"char **"},{"name":"envp","type":"char **"}],"count":3}]
[0x00004070]> s 0x00015bb0
[0x00015bb0]> af
[0x00015bb0]> pd 1
┌ (fcn) fcn.00015bb0 21
│   fcn.00015bb0 (int arg1, int arg2);
│           ; arg int arg1 @ rdi
│           ; arg int arg2 @ rsi
│           0x00015bb0      488b4738       mov rax, qword [rdi + 0x38] ; [0x38:8]=0x1800190040000b ; '8' ; arg1
[0x00015bb0]> afcf
fcn.00015bb0 (int arg1, int arg2);
[0x00015bb0]> afcfj
[{"name":"fcn.00015bb0","args":[{"name":"arg1","type":"int"},{"name":"arg2","type":"int"}],"count":2}]
```

Both disasm and afcf now use the same code for formatting the function signature.

### DIFF
--- a/libr/anal/var.c
+++ b/libr/anal/var.c
@@ -965,3 +965,143 @@ R_API void r_anal_var_list_show(RAnal *anal, RAnalFunction *fcn, int kind, int m
 	}
 	r_list_free (list);
 }
+
+R_API void r_anal_fcn_vars_cache_init(RAnal *anal, RAnalFcnVarsCache *cache, RAnalFunction *fcn) {
+	cache->bvars = r_anal_var_list (anal, fcn, 'b');
+	cache->rvars = r_anal_var_list (anal, fcn, 'r');
+	cache->svars = r_anal_var_list (anal, fcn, 's');
+	r_list_sort (cache->bvars, (RListComparator)var_comparator);
+	r_list_sort (cache->rvars, (RListComparator)var_comparator);
+	r_list_sort (cache->svars, (RListComparator)var_comparator);
+}
+
+R_API void r_anal_fcn_vars_cache_fini(RAnalFcnVarsCache *cache) {
+	if (!cache) {
+		return;
+	}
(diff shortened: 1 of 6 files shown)


radare:
why  the return type is not printed in pd?

```
│   fcn.00015bb0 (int arg1, int arg2);
```

thestr4ng3r:
The return type was never printed before in pd. Though it would probably make sense to print it now. Do you think we should add it?

XVilka:
@thestr4ng3r yes, I think we should.

PR #12703, opened 2019-01-08
Remove flag list in RFlags and just use hashtable

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1646


radare:
Flags2? There are some missing null checks after allocation

ret2libc:
> Flags2? There are some missing null checks after allocation

That needs to be removed, it's redundant since we already have the hashtable that can be used to iterate over elements.

radare:
yep, ill wait for that masacre

ret2libc:
> yep, ill wait for that masacre

yeah :D

radare:
there are 30 XX

radare:
please rebase and squash, merging branchs in pullreqs is a bad practice and its a bit anoying to test and read the history

ret2libc:
I'm using skiplist to implement the foreach functions, so we get sort by offset for free.

ret2libc:
> please rebase and squash, merging branchs in pullreqs is a bad practice and its a bit anoying to test and read the history

No need. I said this many time. Since we end up squashing the entire PR, it doesn't really matter whether the contributor rebase or merge, because at the end the history will be the same, so the result doesn't change. 

It's just easier for the contributor to merge, instead of rebasing, because:
1) you don't need to force-push
2) in case of conflicts between multiple commits, you just have to solve them one time and not many times for each rebased commit.

ret2libc:
@radare if you want you can start having a look at the patch. The red tests are just a matter of sorting and I'll fix them if you think the PR is ok. The PR mainly switch to a more declarative `foreach` approach, which requires working with some callback data, but apart from that is much easier to understand what the code does and it hides the internal details related to how flags are stored (e.g. if one day we'll use a trie, we could make all the code work easily). In this way I could easily remove the `flags` list in RFlag (as said, the details are now hidden, so it was easy), which was redundant since we already have the hashtable by name and the skiplist by offset and we can use the skiplist to iterate the flags in a sorted way.

radare:
the refactoring looks legit, ill look into the pr now

libr/core/vmenus.c
```diff
@@ -858,20 +864,22 @@ R_API bool r_core_visual_hudclasses(RCore *core) {
 	return res? true: false;
 }
 
+static bool hudstuff_append(RFlagItem *fi, void *user) {
+	RList *list = (RList *)user;
+	char *s = r_str_newf ("0x%08"PFMT64x"  %s", fi->offset, fi->name);
+	r_list_append (list, s);
```
radare: null chk

libr/flag/flag.c
```diff
@@ -725,15 +778,52 @@ R_API void r_flag_bind(RFlag *f, RFlagBind *fb) {
...
-		if (r_str_glob (flag->name, glob)) {
-			count++;
-		}
-	}
+
+	r_flag_foreach_glob (f, glob, flag_count_foreach, &count);
 	return count;
 }
+
+#define FOREACH_BODY(condition) \
```
radare: what about making a small static function implementing the condition and make a function instead of a multiline macro here?

libr/core/cmd_flag.c
```diff
@@ -168,45 +167,56 @@ static void cmd_fz(RCore *core, const char *input) {
...
+		u->next = fi->offset;
 	}
+	return true;
 }
 
 static int flag_to_flag(RCore *core, const char *glob) {
-	RFlagItem *flag;
-	RListIter *iter;
-	ut64 next = UT64_MAX;
+	r_return_val_if_fail (glob, 0);
```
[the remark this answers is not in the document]
ret2libc: Yes/No. They are anyway a way to ensure your assumption about your code are correct, so they are still useful in static code too. I'd keep them there too, unless you have very strong opinion about it.

libr/core/core.c
```diff
@@ -1140,23 +1140,23 @@ static void autocomplete_breakpoints(RLine* line, const char* str) {
...
+	tmp_argv[(*i)++] = fi->name;
+	if (*i == TMP_ARGV_SZ - 1) {
+		return false;
+	}
+	return true;
+}
+
 static void autocomplete_flags(RLine* line, const char* str) {
 	RCore *core = line->user;
 	if (!core || !str) {
```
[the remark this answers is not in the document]
ret2libc: why?

libr/flag/flag.c
ret2libc: Mmm... I'm not sure it's possible. The condition needs to access some parameters that are different for each function, so I would need to do things in a weird way. The macro makes it easier. Do you have a solution for this?

ret2libc: Please share how you would do that. I cannot think of a way to do that without complicating things for no reasons.

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1646

ret2libc:
This seems ready to me @radare . There are only a couple of your comments I have not addressed, reply to those comments if you care about them.

PR #12705, opened 2019-01-08
Have arm pseudo's replace() use the same replacement style as x86's

jcreigh:
Fix `movt r0, 0x1234` from incorrectly showing up as `r0 = 0x1234 << r0`


radare:
Would be good to have some common code to avoid repeating the same parser if its enough generic. Do you have an example of how this improves the disasm? A test?

Thanks!

jcreigh:
> Would be good to have some common code to avoid repeating the same parser if its enough generic. Do you have an example of how this improves the disasm? A test? Thanks!

Might be worth splitting out into something common. Right now, each parser is providing its own replacement function which all basically do the same thing.

In this particularly case, it just fixes the `movt` instruction, which initially has `<< 16` in it, which improperly got replaced with `<< r0` due to the `1`. It was either that, change the numbers to capital letters, or introduce an escape to prevent the number from being used as a variable. Since x86's is ostensibly the most used parser, I figured it'd be fine to copy its style. I did it with a script instead of by hand so there wouldn't be any typos, and I literally copied x86's without changes. I'm not sure of the best way to test it.

If a particular scheme could be agreed upon, I think it'd be a worthy goal to convert all the parsers to using a common implementation.

radare:
travis complains

screenshot 2019-01-09 at 00 53 04

libr/parse/p/parse_arm_pseudo.c
```diff
@@ -133,26 +136,40 @@ static int replace(int argc, const char *argv[], char *newstr) {
...
+			if (argc - 1 != ops[i].narg) {
 				continue;
 			}
 		}
 		if (!strcmp (ops[i].op, argv[0])) {
-			for (j = k = 0; ops[i].str[j] != '\0'; j++, k++) {
-				if (IS_DIGIT(ops[i].str[j])) {
-					int idx = ops[i].str[j]-'0';
-					if (idx < argc) {
+			if (newstr != NULL) {
```
radare: != NULL is implicit, just do if (newstr)

libr/parse/p/parse_arm_pseudo.c
```diff
@@ -133,26 +136,40 @@ static int replace(int argc, const char *argv[], char *newstr) {
...
+						}
+						int idx = ops[i].args[d];
+						d++;
+						if (idx <= 0) {
+							// XXX Shouldn't ever happen...
+							continue;
+						}
 						const char *w = argv[idx];
-						if (w) {
+						if (w != NULL) {
```
radare: != NULL is unnecessary just do if (w)

PR #12709, opened 2019-01-09
fix a little bug in wasm type section parsing

XYlearn:


XYlearn:
before this patch. r2 will not parse entire was type section

PR #12712, opened 2019-01-09
Add initial support for dyldcache v3 ##bin

mrmacete:
This is for arm64e caches

### DIFF
--- a/libr/bin/p/bin_dyldcache.c
+++ b/libr/bin/p/bin_dyldcache.c
@@ -11,6 +11,8 @@
 #include "../format/mach0/mach0.h"
 #include "objc/mach0_classes.h"
 
+#define R_IS_PTR_AUTHENTICATED(x) ((x) & 0x8000000000000000ULL)
+
 typedef struct {
 	ut8 version;
 	ut64 slide;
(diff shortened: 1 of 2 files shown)


libr/bin/p/bin_dyldcache.c
```diff
@@ -466,7 +543,7 @@ static RDyldRebaseInfo *get_rebase_info(RBinFile *bf, RDyldCache *cache) {
 		rebase_info->start_of_data = start_of_data;
 		rebase_info->one_page_buf = one_page_buf;
 		rebase_info->page_size = 4096;
-		rebase_info->slide = estimate_slide (bf, cache, 0xffffffffffffffffL);
+		rebase_info->slide = estimate_slide (bf, cache, 0xffffffffffffffffULL);
```
Maijin: Can't you use defines like UT64_MAX like defined in https://github.com/radare/radare2/blob/master/libr/include/r_types_base.h ?

mrmacete: you're right: https://github.com/radare/radare2/pull/12720

Maijin: great thanks 👍

PR #12713, opened 2019-01-09
Fix #12654 - resolv reloc symbols

### LINKED ISSUE #12654: Missing reloc symbols
Hi there,
I came a across the fact that some reloc symbols are not recognized by radare2. I'm already trying to debug this for some time now, and I fear I'm might missing some detail.  So I would appreciate any help on this.

### Work environment

| OS/arch/bits (mandatory)                             | Manjaro X86 64
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.2.0-git 20627 @ linux-x86-64 git.3.1.3-217-g399fcc24a commit: 399fcc24aad065fe08fefc6d74bebdbc53e0181a build: 2019-01-04__23:00:04

### Expected behavior
```
r2 /usr/lib/libc.so.6
[0x00024330]> aa
...
[0x00024330]> sf sym.__run_exit_handlers
[0x0003a320]> agcj~{}
[
  {
    "name": "sym.__run_exit_handlers",
    "size": 588,
    "imports": [
      "sym.__lll_lock_wait_private",
      "sym.__lll_unlock_wake_private",
      "reloc.free",
      "rsp",
      "sym._Exit",
      "sym.__call_tls_dtors"
    ]
  }
]
```
From BinaryNinja:
screenshot from 2019-01-04 23-22-32

### Actual behavior
```
r2 /usr/lib/libc.so.6
[0x00024330]> aa
...
[0x00024330]> sf sym.__run_exit_handlers
[0x0003a320]> agcj~{}
[
  {
    "name": "sym.__run_exit_handlers",
    "size": 588,
    "imports": [
      "sym.__lll_lock_wait_private",
      "sym.__lll_unlock_wake_private",
      "unk.0x1bdfa8",
      "rsp",
      "sym._Exit",
      "sym.__call_tls_dtors"
    ]
  }
]
```
Visual Mode shows same behaviour:
screenshot from 2019-01-04 23-23-33

Execute steps as shown above using (extracted) [libc.so.6.zip](https://github.com/radare/radare2/files/2728760/libc.so.6.zip)

blenk92:
Another try for https://github.com/radare/radare2/issues/12654

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -2943,14 +2943,14 @@ static RBinElfSymbol* get_symbols_from_phdr(ELFOBJ *bin, int type) {
 		bool is_vaddr = false;
 		// zero symbol is always empty
 		// Examine entry and maybe store
-		if (type == R_BIN_ELF_IMPORTS && sym[i].st_shndx == SHT_NULL) {
+		if (type == R_BIN_ELF_IMPORT_SYMBOLS && sym[i].st_shndx == SHT_NULL) {
 			if (sym[i].st_value) {
 				toffset = sym[i].st_value;
 			} else if ((toffset = get_import_addr (bin, i)) == -1){
 				toffset = 0;
 			}
 			tsize = 16;
-		} else if (type == R_BIN_ELF_SYMBOLS) {
+		} else if (type == R_BIN_ELF_ALL_SYMBOLS) {
 			tsize = sym[i].st_size;
 			toffset = (ut64) sym[i].st_value;
 			is_sht_null = sym[i].st_shndx == SHT_NULL;
@@ -3018,14 +3018,14 @@ done:
 		ret = p;
 	}
 	ret[ret_ctr].last = 1;
-	if (type == R_BIN_ELF_IMPORTS && !bin->imports_by_ord_size) {
+	if (type == R_BIN_ELF_IMPORT_SYMBOLS && !bin->imports_by_ord_size) {
 		bin->imports_by_ord_size = ret_ctr + 1;
 		if (ret_ctr > 0) {
 			bin->imports_by_ord = (RBinImport * *) calloc (ret_ctr + 1, sizeof (RBinImport*));
 		} else {
 			bin->imports_by_ord = NULL;
 		}
-	} else if (type == R_BIN_ELF_SYMBOLS && !bin->symbols_by_ord_size && ret_ctr) {
+	} else if (type == R_BIN_ELF_ALL_SYMBOLS && !bin->symbols_by_ord_size && ret_ctr) {
 		bin->symbols_by_ord_size = ret_ctr + 1;
 		if (ret_ctr > 0) {
 			bin->symbols_by_ord = (RBinSymbol * *) calloc (ret_ctr + 1, sizeof (RBinSymbol*));
@@ -3048,7 +3048,7 @@ static RBinElfSymbol *Elf_(r_bin_elf_get_phdr_symbols)(ELFOBJ *bin) {
 	if (bin->phdr_symbols) {
 		return bin->phdr_symbols;
 	}
-	bin->phdr_symbols = get_symbols_from_phdr (bin, R_BIN_ELF_SYMBOLS);
+	bin->phdr_symbols = get_symbols_from_phdr (bin, R_BIN_ELF_ALL_SYMBOLS);
 	return bin->phdr_symbols;
 }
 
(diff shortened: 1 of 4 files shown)


blenk92:
@ret2libc Please review

blenk92:
wait a second, I forgot to free some memory.  Let me fix this first...

blenk92:
Now the memory gets freed...

libr/bin/p/bin_elf.inc
```diff
@@ -1042,9 +1050,9 @@ static void lookup_symbols(RBinFile *bf, RBinInfo *ret) {
 				ret->lang = "rust";
 			}
 		}
-		symbols_list->free = r_bin_symbol_free;
-		r_list_free (symbols_list);
 	}
+	symbols_list->free = r_bin_symbol_free;
```
ret2libc: I think this was in the if because otherwise you could access NULL

blenk92: :+1: 

> I think this was in the if because otherwise you could access NULL

:+1:

libr/bin/format/elf/elf.c
```diff
@@ -3199,6 +3203,20 @@ static RBinElfSymbol* Elf_(_r_bin_elf_get_symbols_imports)(ELFOBJ *bin, int type
 					if (st_name < 0 || st_name >= maxsize) {
 						ret[ret_ctr].name[0] = 0;
 					} else {
+						bool found = false;
```
ret2libc: why this new chunk of code? what does it fix?

ret2libc:
I tried a bit the PR and it seems to work well! Great!

However, I'm not sure it's necessary to have g_dyn_symbols. AFAIK symbols_by_ord is used only in reloc_convert and patch_relocs and both require  the dynamic symbols, not the symtab ones. For this reason I think we could just fill out the symbols_by_ord with the dynamic symbol entries in `_r_bin_elf_get_symbols_imports`, without the necessity to call `r_bin_elf_get_dynamic_symbols` before analyzing relocs.

blenk92:
> I tried a bit the PR and it seems to work well! Great!
> 
> However, I'm not sure it's necessary to have g_dyn_symbols. AFAIK symbols_by_ord is used only in reloc_convert and patch_relocs and both require the dynamic symbols, not the symtab ones. For this reason I think we could just fill out the symbols_by_ord with the dynamic symbol entries in `_r_bin_elf_get_symbols_imports`, without the necessity to call `r_bin_elf_get_dynamic_symbols` before analyzing relocs.

That is good news.

Yeah you are right. Let me try to fix that. 
Thanks for your review!

libr/core/cbin.c
```diff
@@ -1262,7 +1262,7 @@ static char *get_reloc_name(RBinReloc *reloc, ut64 addr) {
 		}
 		r_str_replace_char (reloc_name, '$', '_');
 	} else if (reloc->symbol && reloc->symbol->name) {
-		reloc_name = sdb_fmt ("reloc.%s_%d", reloc->symbol->name, (int)(addr & 0xff));
+		reloc_name = sdb_fmt ("reloc.%s", reloc->symbol->name);
```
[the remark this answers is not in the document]
blenk92: However, i want to point out that this solution also is not bullet-proof. If the last to bytes of the address match, this what you said will happen anyway.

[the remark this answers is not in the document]
blenk92: Hmm, actually reverting this change introduces odd behaviour. The original example now changes to
```
r2 /usr/lib/libc.so.6
[0x00024330]> aa
[0x00024330]> sf sym.__run_exit_handlers
[0x0003a320]> agcj~{}
[
  {
    "name": "sym.__run_exit_handlers",
    "size": 588,
    "imports": [
      "sym.__lll_lock_wait_private",
      "sym.__lll_unlock_wake_private",
      "reloc.free_168",
      "rsp",
      "sym.__GI__exit",
      "sym.__GI___call_tls_dtors"
    ]
  }
]
```
while the list of relocation symbols still contains the symbol without the address appended:
```
[0x0003a320]> ir~free
(...)
vaddr=0x001bdfa8 paddr=0x001bcfa8 type=SET_64 free
```

blenk92:
i guess the build was failing because of an invalid read. At least there is a invalid read when i try to execute the actions locally. The PR https://github.com/radare/radare2/pull/12730 aims to fix this invalid read.
@radare could you retrigger the test-execution if the other pr has been merged?

ret2libc:
So... The code does seem a bit confused *to me* (so maybe i'm just looking at it the wrong way :) BUT it works and... let's just merge it and allow others to test it as well. Everything seems better with this PR, so we can work on improving the code, if necessary, later.

Thanks a lot.

PR #12723, opened 2019-01-10
meson: Do not use reversed keyword of meson

lzutao:
[`files`](https://mesonbuild.com/Reference-manual.html#files) is a reversed keyword in meson. Try to avoid using it.


Maijin:
@pelijah

PR #12728, opened 2019-01-10
Add vars in zignatures ##sign

radare:


radare:
please add tests and verify it is working. the results should be improved quite a bit because we have  more metrics to compare.

XVilka:
Awesome, along with missing storing return types (and var types) and it will be in full parity with FLIRT.

Issue #12207, opened 2018-11-18
Increase size column in iS 1 char at least


radare:
most sections are >100K, so we need 1 more char to show that without 

```
[0x00000000]> iS
[Sections]
Nm Paddr       Size Vaddr      Memsz Perms Name
00 0x00000000 65537 0x00000000 65537 -rwx TEXT
01 0x00010000 16385 0x00010000 16385 -rwx DATA
02 0x00014000 360449 0x00014000 360449 -rwx LLVM
03 0x0006c000 131073 0x0006c000 131073 -rwx LINKEDIT
```

radare:
There are also other issues to handle vaddr/paddr columns, so we should probably use the table api to render this info

```
[0x1000011ec]> iS
[Sections]
Nm Paddr        Size Vaddr      Memsz Perms Name
00 0x00000f0c  13588 0x100000f0c 13588 -r-x 0.__TEXT.__text
01 0x00004420    456 0x100004420   456 -r-x 1.__TEXT.__stubs
02 0x000045e8    776 0x1000045e8   776 -r-x 2.__TEXT.__stub_helper
03 0x000048f0    504 0x1000048f0   504 -r-x 3.__TEXT.__const
04 0x00004ae8   1151 0x100004ae8  1151 -r-x 4.__TEXT.__cstring
05 0x00004f68    144 0x100004f68   144 -r-x 5.__TEXT.__unwind_info
06 0x00005000     16 0x100005000    16 -rw- 6.__DATA.__nl_symbol_ptr
07 0x00005010     40 0x100005010    40 -rw- 7.__DATA.__got
08 0x00005038    608 0x100005038   608 -rw- 8.__DATA.__la_symbol_ptr
09 0x000052a0    552 0x1000052a0   552 -rw- 9.__DATA.__const
10 0x000054d0     40 0x1000054d0    40 -rw- 10.__DATA.__data
11 0x00000000      0 0x100005500   224 -rw- 11.__DATA.__bss
12 0x00000000      0 0x1000055e0   140 -rw- 12.__DATA.__common
```

radare:
```
pmb:elf pancake$ r2 hello.ppc
 -- git blind --hard
[0x10000308]> ii
[Imports]
Num  Vaddr       Bind      Type Name
   1 0x4743430a  GLOBAL    FUNC printf
   2 0x4743431a    WEAK  NOTYPE __gmon_start__
   3 0x4743432a  GLOBAL    FUNC __libc_start_main

```

[...]

PR #12736, opened 2019-01-10
Increased size column for iS 

a-meth:
Issue #12207


radare:
the problem with this is that many tests will break and we need a flexible table implementation to render all those results otherwise we will be hacking with spaces all the time. So i prefer to close this PR until the table API is implemented

PR #12739, opened 2019-01-10
Store Anal Classes list in own Namespace ##anal ##classes

thestr4ng3r:


radare:
can you access it via the `k`  command?

thestr4ng3r:
It's in sdb, so yes of course. Or what do you mean?

radare:
its nested inside another sdb referenced from the root one?

thestr4ng3r:
The classes list is in `anal/classes` (for every class `ClassName=c`), the attrs are in `anal/classes/attrs`.

radare:
dont use more than one ##tag

PR #12749, opened 2019-01-11
Fix #12743

### LINKED ISSUE #12743: Undefined behaviour in ESIL
[esil.c:1183:24: runtime error: left shift of 1 by 63 places cannot be represented in type 'long long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior esil.c:1183:24 in

r4dix10:


radare:
We should check if this issue is related to capstone or its esil representation. 

Because maybe arm64 allows to do >63 bit shifts and maybe the behaviour is undefined. This patch will save us from UB by asan but i wohld like to clarify this and test it in real arm64 hw

r4dix10:
> We should check if this issue is related to capstone or its esil representation.
> 
> Because maybe arm64 allows to do >63 bit shifts and maybe the behaviour is undefined. This patch will save us from UB by asan but i wohld like to clarify this and test it in real arm64 hw

write tests ?

radare:
i think we should move this conditional above, with an eprintf to warn the user about this invalid esil expression, and then keep the shift code in the }else branch

radare:
i have fixed the condition and merged. thanks! we shuold probably check this kind of bug in more places in r2.

PR #12761, opened 2019-01-13
Prevent runaway scrolling in Visual mode ##visual

kazarmy:
... caused by holding down a movement key. Tested on Linux only.


radare:
interesting this was maybe the solution to the fflush(stdin) discussion we had a month ago

PR #12764, opened 2019-01-13
Flush Visual stdin only if printable, \t or \n

kazarmy:
Partially fixes the runaway scrolling issue but shouldn't disable any key/mouse input.

### DIFF
--- a/libr/core/visual.c
+++ b/libr/core/visual.c
@@ -4039,7 +4039,7 @@ dodo:
 				ch = r_cons_readchar ();
 			}
 #ifndef __WINDOWS__
-			if (!r_config_get_i (core->config, "scr.wheel")) {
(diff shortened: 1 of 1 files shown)


kazarmy:
Tried to prevent the runaway scrolling issue for arrow keys but testing revealed a scrolling slowdown after some time mashing the up and down arrow keys. Happens for both `tcflush (STDIN_FILENO, TCIFLUSH);` and `while (r_cons_readchar_timeout (0) != -1) ;`. Strangely, pressing the ':' key and hitting Enter resets scrolling speed. Don't know whether the problem is in r2, glibc, the kernel or the VM keyboard driver.

kazarmy:
Hmm there seems to be a workaround in https://github.com/radare/radare2/pull/12823.

PR #12769, opened 2019-01-13
HTTP/ reuse port option feature.

devnexen:
Allow multiple instances, might be of use
but is optional.


wargio:
i would suggest to extend the check on `#ifdef SO_REUSEPORT/#endif` in all the files, but i would wait @radare to confirm this.

radare:
what's the actual use case for this? im not sure what will benefit us from having this

devnexen:
@radare for now probably not much use, was thinking for possible evolution of this embedded web server.

radare:
i dont see the need to add new cmdline flags, config vars and such for something we dont know why would be useful for.

devnexen:
All right then.

PR #12770, opened 2019-01-13
Fix iR timestamp + consistent timestamp format output (pt*, iI~comp, iR)

### LINKED ISSUE #12327: Wrong PE resource Timestamp
### Work environment

| OS/arch/bits (mandatory)                             | OSX 64
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.1.0 20218 @ darwin-x86-64 git.3.1.0 commit: c033496ebc7034e52a84be9cdb2d2dfad6a4cfac build: 2018-11-26__10:29:06

### Expected behavior

[e15040beca2635a66eb395162f39db9f23468e7bf9f00c9c62dd5913cd5f3850.zip](https://github.com/radare/radare2/files/2615829/e15040beca2635a66eb395162f39db9f23468e7bf9f00c9c62dd5913cd5f3850.zip)
(It's malware, password is "infected")

`iR`

```
Resource 0
  name: 1
  timestamp: Sat May 19 09:15:04 2012
  vaddr: 0x004a2f04
  size: 308
  type: CURSOR
  language: LANG_ENGLISH
```

### Actual behavior

`iR`

```
Resource 0
  name: 1
  timestamp: Tue May 25 14:28:02 2004
  vaddr: 0x004a2f04
  size: 308
  type: CURSOR
  language: LANG_ENGLISH
```

koffiedrinker:
Fixes #12327 .

```
[koffiedrinker@ctf radare2]$ ./binr/radare2/radare2 ../e15040beca2635a66eb395162f39db9f23468e7bf9f00c9c62dd5913cd5f3850
[x] Constructing a function name for fcn.* and sym.func.* functions (aan)
[x] Enable constraint types analysis for variables
 -- 255 shades of (truecolor) grey
[0x004b1660]> iR
Resource 0
  name: 1
  timestamp: Sat May 19 09:15:04 2012
  vaddr: 0x004a2f04
  size: 308
  type: CURSOR
  language: LANG_ENGLISH
Resource 1
...
```

Basically, I take the timestamp from the image file header (which is always POSIX afaik) and compare with it to figure out whether the resource timestamp is POSIX or DOS. Added code to do the DOS timestamp conversion (I found one in date.c but I wasn't sure if I'm supposed to use that one/refactor it) and then I pipe it through the existing `_time_stamp_to_str()` to get the correct output.


Maijin:
Looks great, would be nice to have to find a non-malware binary to add a test for this

libr/bin/format/pe/pe.c
```diff
@@ -537,6 +537,43 @@ static char *_time_stamp_to_str(ut32 timeStamp) {
 #endif
 }
 
+static char*_time_stamp_to_dos_str(ut32 timeStamp) {
```
radare: uhm have you seen libr/util/date.c ?

radare: this code is not PE specific

koffiedrinker: Ok, will refactor.

koffiedrinker:
So, I refactored date.c to have the same timestamp for all pt* commands as well as iI~comp and iR:

Before:
```
[koffiedrinker@ctf radare2]$ ./binr/radare2/radare2 -
 -- You see it, you fix it!
[0x00000000]> b 4
[0x00000000]> ptd
1980-00-00 0:0:0
[0x00000000]> wx e249b340
[0x00000000]> pt
2004-05-25 13:28:02 +0000
[0x00000000]> ptd
2012-05-19 9:15:4
[0x00000000]>
[0x00000000]> pth
2070-05-26 13:28:02 +0000

[0x00000000]> pt.
Tue Jan 15 20:17:27 2019
[0x00000000]> ? 1547579847 + 11644473600
int64   13192053447
[0x00000000]> ? 13192053447 * 10000000
hex     0x1d4ad06f3a9fd80
[0x00000000]> b 8
[0x00000000]> wx 80fda9f306add401
[0x00000000]> ptn
2019-01-15 19:17:27 +0000
```

After:
```
[koffiedrinker@ctf radare2]$ r2 -
 -- What about taking a break? Here, have this nice 0xCC.
[0x00000000]> b 4
[0x00000000]> ptd
Tue Jan  1 00:00:00 1980
[0x00000000]> pt
Thu Jan  1 01:00:00 1970
[0x00000000]> wx e249b340
[0x00000000]> pt
Tue May 25 15:28:02 2004
[0x00000000]> ptd
Sat May 19 09:15:04 2012
[0x00000000]> pth
Mon May 26 15:28:02 2070
[0x00000000]> b 8
[0x00000000]> wx 80fda9f306add401
[0x00000000]> ptn
Tue Jan 15 20:17:27 2019
[0x00000000]>
[0x00000000]> q
```

* All pt* commands now use the same base function which I moved from pe.c to date.c (it was already there for the parsing the PE header timestamp). The iR command also uses this function.
* For "ptd" which is "print time dos" I fixed the output it would give for 0x00000000 (1980-00-00 doesn't make sense) or null bytes/invalid timestamps in general.

The only open point there still is, is that ptd/iR doesn't take into account DST. All others are localized correctly (I think).

radare:
screenshot 2019-01-16 at 01 03 24

koffiedrinker:
What do you want me to do?

I can change the regression to fit the output it will now return. Or I can change the output of all the timestamps.

Having it in a test does make it difficult since the timestamps are localized so depending on the testing setup timezone, you might fail the check or not. Maybe it's better to not localize all timestamps? Would fix the DST issue we have with iR and ptd.

radare:
welll. . it's `Fri Feb` vs `2009-02-13 23:31:30` ... the "new" output is pretty short and imprise :P im fine for using the new date format, but all the time info should be there, not just month and weekday

koffiedrinker:
> welll. . it's `Fri Feb` vs `2009-02-13 23:31:30` ... the "new" output is pretty short and imprise :P im fine for using the new date format, but all the time info should be there, not just month and weekday

That's because of the "cut" command after the rax2 -t command that cuts the end off? https://github.com/radare/radare2-regressions/blob/64ff39f8c51c84d18fd76e5423b42f3600d64a9c/new/db/tools/rax2#L317 (I'll test it this evening to be sure)

radare:
ok then the pr can be merged, but the test must be fixed

radare:
Cc @maijin pls can you fix the test and merge?

Maijin:
Done.

PR #12778, opened 2019-01-14
Fix #12773 - Support setting multiple configs at the same time

### LINKED ISSUE #12773: Support to set more than one config variable at the same time
For example:

`e search.from = anal.from = 0x808080`

GustavoLCR:
This fixes #12773.

### DIFF
--- a/libr/config/config.c
+++ b/libr/config/config.c
@@ -593,21 +593,17 @@ beach:
 }
 
 R_API int r_config_eval(RConfig *cfg, const char *str) {
-	char *ptr, *a, *b, name[1024];
+	char *ptr, *config, *val, names[1024];
 	unsigned int len;
 	if (!str || !cfg) {
 		return false;
 	}
 	len = strlen (str) + 1;
-	if (len >= sizeof (name)) {
+	if (len >= sizeof (names)) {
 		return false;
 	}
-	memcpy (name, str, len);
-	str = r_str_trim (name);
-
-	if (!str) {
-		return false;
-	}
+	memcpy (names, str, len);
+	str = r_str_trim (names);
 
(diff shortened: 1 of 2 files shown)


libr/config/config.c
```diff
@@ -593,21 +593,17 @@ R_API RConfigNode* r_config_set_i(RConfig *cfg, const char *name, const ut64 i)
 }
 
 R_API int r_config_eval(RConfig *cfg, const char *str) {
-	char *ptr, *a, *b, name[1024];
+	char *ptr, *config, *val, names[1024];
```
radare: why 1024, didnt tested the patch, but will be good to have some tests, does it support more than one assignment?

GustavoLCR: Yes, it does suppport multiple assignments. 
i.e.
`e a = b = c = d = some_val`

radare:
Im thinking on a possible issue where we want to actually use the '=' char as the value. Maybe we should check if the command is in quote mode (this is, last char is '"' (or check a var that is set already) to make `"e cmd.prompt=e scr.color=3"` work too

radare:
can you rebase? the broken  tests look unrelated

libr/config/config.c
```diff
@@ -593,43 +593,62 @@ R_API RConfigNode* r_config_set_i(RConfig *cfg, const char *name, const ut64 i)
...
-	ptr = strchr (str, '=');
-	if (ptr) {
+	val = strrchr (names, '=');
+	if (val) {
 		/* set */
-		ptr[0] = '\0';
-		a = r_str_trim (name);
-		b = r_str_trim (ptr + 1);
-		(void) r_config_set (cfg, a, b);
+		ptr = strrchr (names, '"');
```
radare: maybe better to just use r_str_endswith()

Issue #7250, opened 2017-04-11
rasm2/armass missing many opcodes (some of v4, most of v5+)


endrift:
Here's a few that I noticed quickly:
- Basic:
  - [x] mul
  - [x] smull
  - [x] smlal
  - [x] smla< x >< y >
  - [x] smlaw< y >
  - [ ] umull
  - [ ] umlal
  - [ ] msr
  - [ ] mrs
  - [ ] clz
- Memory:
  - [ ] pld
  - [ ] ldm (all forms)
  - [ ] stm (all forms)
- Saturating arithmetic
  - All of it
- Misc
  - [ ] rev (all forms)

And a bunch of stuff from ARMv6 and ARMv7 that I haven't bothered checking yet

trufae:
all done, or at least i think so. chk https://github.com/radareorg/radare2/pull/24593 and if u miss any other one just fill a ticket and ill take care of them. thanks for reporting and sorry for the delay

PR #12788, opened 2019-01-15
Fix arm 64 assemble for msr/mrs OPs.

SimonTheCoder:
Added:
MSR support
MSR (immediate) support
MRS support


SimonTheCoder:
> @SimonTheCoder Hello o/ Maybe you want to check - #7250

OK. Thanks.

PR #12790, opened 2019-01-15
Fix afns and add JSON support

ITAYC0HEN:
The following pull request fixes the broken `afns`. 
`afns` is being used to print all the strings referenced in a function. Previously, when executed `afns` it used to print a single flag, which wasn't necessarily a string. This was because the `break;` keyword prevents the loop to iterate over the referenced flags. This PR fixes it.
Moreover, it implements a proper JSON support for `afnsj`.

```
[0x00001347]> afns
0x00001356 0x00002004 str..::_Megabeets_::.
0x00001362 0x00002019 str.Think_you_can_make_it
0x0000138b 0x00002030 str.Success
0x00001399 0x0000203a str.Nop__Wrong_argument.

[0x00001347]> afnsj~{}
[
  {
    "addr": 4950,
    "ref": 8196,
    "flag": "str..::_Megabeets_::."
  },
  {
    "addr": 4962,
    "ref": 8217,
    "flag": "str.Think_you_can_make_it"
  },
  {
    "addr": 5003,
    "ref": 8240,
    "flag": "str.Success"
  },
  {
    "addr": 5017,
    "ref": 8250,
    "flag": "str.Nop__Wrong_argument."
  }
]
```

### DIFF
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -313,49 +313,78 @@ static bool blacklisted_word(char* name) {
 	return false;
 }
 
-static char *anal_fcn_autoname(RCore *core, RAnalFunction *fcn, int dump) {
+static char *anal_fcn_autoname(RCore *core, RAnalFunction *fcn, int dump, int mode) {
 	int use_getopt = 0;
 	int use_isatty = 0;
+	bool first = true; // is first iteration of the JSON loop
 	char *do_call = NULL;
 	RAnalRef *ref;
 	RListIter *iter;
 	RList *refs = r_anal_fcn_get_refs (core->anal, fcn);
+	if (mode == 'j') {
+		r_cons_printf ("[");
+	}
 	r_list_foreach (refs, iter, ref) {
 		RFlagItem *f = r_flag_get_i (core->flags, ref->addr);
 		if (f) {
-			if (dump) {
-				r_cons_printf ("0x%08"PFMT64x" 0x%08"PFMT64x" %s\n", ref->at, ref->addr, f->name);
-			}
-			if (blacklisted_word (f->name)) {
-				break;
-			}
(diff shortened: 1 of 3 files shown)


libr/core/canal.c
```diff
@@ -313,49 +313,78 @@ static bool blacklisted_word(char* name) {
...
+				// take only strings flags
+				if (!strncmp (f->name, "str.", 4)) {
+					if (mode == 'j') {
+						if (first) {
+							first = false;
+						} else {
+							r_cons_printf (",");
+						}
+						// print output in JSON format
+						r_cons_printf ("{\"addr\":%"PFMT64u", \"ref\":%"PFMT64u",\"flag\": \"%s\"}", ref->at, ref->addr, f->name);
```
radare: use the pj api and just do a single rconsprintf() at the end of the loop. also, this will solve the problem of the %s non escaped here

libr/core/canal.c
```diff
@@ -313,49 +313,78 @@ static bool blacklisted_word(char* name) {
...
+					}
```
radare: else starts in the same line as }

libr/core/canal.c
```diff
@@ -313,49 +313,78 @@ static bool blacklisted_word(char* name) {
...
+					else {
+						r_cons_printf ("0x%08"PFMT64x" 0x%08"PFMT64x" %s\n", ref->at, ref->addr, f->name);
+					}
+				}
+ 			}
+			// break if a proper autoname found and not in dump mode
+			else if (do_call) {
```
radare: same  here

radare:
thanks!

Maijin:
@ITAYC0HEN please add a test in new/json

PR #12798, opened 2019-01-15
Issue #12738

### LINKED ISSUE #12738: afi should display if the function has been executed or not
we have trace information at instruction level using the `dt` command. but will be good to have this info available in `afll` for grepping /sorting/coloring those who have been executed during an emulation/run.

r4dix10:
Fix #12738 - afi displays if function is executed.

Should this be extended to display function coverage ? as discussed in https://github.com/radare/radare2/issues/12738#issuecomment-453823452

### DIFF
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2519,6 +2519,22 @@ static int fcn_print_detail(RCore *core, RAnalFunction *fcn) {
 	return 0;
 }
 
+static bool is_fcn_traced(RDebugTrace *traced, RAnalFunction *fcn) {
+	int tag = traced->tag;
+	RListIter *iter;
+	RDebugTracepoint *trace;
+
+	r_list_foreach (traced->traces, iter, trace) {
+		if (!trace->tag || (tag & trace->tag)) {
+			if (r_anal_fcn_in (fcn, trace->addr)) {
+				r_cons_printf ("%d\n", trace->times);
+				return true;
+			}
+		}
+	}
+	return false;
+}
+
 static int fcn_print_legacy(RCore *core, RAnalFunction *fcn) {
 	RListIter *iter;
 	RAnalRef *refi;
@@ -2528,26 +2544,10 @@ static int fcn_print_legacy(RCore *core, RAnalFunction *fcn) {
 
 	// execution times:
 	r_cons_printf ("execution state: ");
-	if (core->dbg->trace->enabled == 0) {
+	if (!core->dbg->trace->enabled) {
 		r_cons_printf ("(dbg.trace unset)\n");
-	} else if (core->dbg->trace->enabled == 1) {
-		unsigned long long max_addr = fcn->addr;
-		max_addr += (ut64)r_anal_fcn_size (fcn);
-
-		bool flag = false;
-		int tag = core->dbg->trace->tag;
-		RListIter *iter;
-		RDebugTracepoint *trace;
-		r_list_foreach (core->dbg->trace->traces, iter, trace) {
-		  if (!trace->tag || (tag & trace->tag)) {
-			if (trace->addr >= fcn->addr && trace->addr <= max_addr) {
-			  r_cons_printf ("%d\n", trace->times);
-			  flag = true;
-			  break;
-			}
-		  }
-		}
-		if (!flag) {
+	} else if (core->dbg->trace->enabled) {
+		if (!is_fcn_traced (core->dbg->trace, fcn)) {
 			r_cons_println ("Not Executed");
 		}
 	}
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2542,16 +2542,6 @@ static int fcn_print_legacy(RCore *core, RAnalFunction *fcn) {
 	int ebbs = 0;
 	char *name = r_core_anal_fcn_name (core, fcn);
 
-	// execution times:
-	r_cons_printf ("execution state: ");
-	if (!core->dbg->trace->enabled) {
-		r_cons_printf ("(dbg.trace unset)\n");
-	} else if (core->dbg->trace->enabled) {
-		if (!is_fcn_traced (core->dbg->trace, fcn)) {
-			r_cons_println ("Not Executed");
-		}
-	}
-
 	r_cons_printf ("#\noffset: 0x%08"PFMT64x"\nname: %s\nsize: %"PFMT64d,
 			fcn->addr, name, (ut64)r_anal_fcn_size (fcn));
 	r_cons_printf ("\nis-pure: %s", r_anal_fcn_get_purity (core->anal, fcn) ? "true" : "false");
@@ -2633,6 +2623,17 @@ static int fcn_print_legacy(RCore *core, RAnalFunction *fcn) {
 		}
 	}
 	free (name);
+
+	// execution times:
+	r_cons_printf ("\nexecution state: ");
+	if (core->dbg->trace->enabled) {
+		if (!is_fcn_traced (core->dbg->trace, fcn)) {
+			r_cons_println ("Not Executed");
+		}
+	} else {
+		r_cons_printf ("(dbg.trace unset)\n");
+	}
+
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


libr/core/canal.c
```diff
@@ -2525,6 +2525,33 @@ static int fcn_print_legacy(RCore *core, RAnalFunction *fcn) {
 	RList *refs, *xrefs;
 	int ebbs = 0;
 	char *name = r_core_anal_fcn_name (core, fcn);
+
+	// execution times:
+	r_cons_printf ("execution state: ");
+	if (core->dbg->trace->enabled == 0) {
+		r_cons_printf ("(dbg.trace unset)\n");
+	} else if (core->dbg->trace->enabled == 1) {
```
radare: this is boolean and can only contain 2 values,  no need to == 0  and == 1, just if{}else{}

libr/core/canal.c
```diff
@@ -2525,6 +2525,33 @@ static int fcn_print_legacy(RCore *core, RAnalFunction *fcn) {
...
+		max_addr += (ut64)r_anal_fcn_size (fcn);
+
+		bool flag = false;
+		int tag = core->dbg->trace->tag;
+		RListIter *iter;
+		RDebugTracepoint *trace;
+		r_list_foreach (core->dbg->trace->traces, iter, trace) {
+		  if (!trace->tag || (tag & trace->tag)) {
+			if (trace->addr >= fcn->addr && trace->addr <= max_addr) {
+			  r_cons_printf ("%d\n", trace->times);
```
radare: use  tabs

libr/core/canal.c
```diff
@@ -2525,6 +2525,33 @@ static int fcn_print_legacy(RCore *core, RAnalFunction *fcn) {
...
+		unsigned long long max_addr = fcn->addr;
+		max_addr += (ut64)r_anal_fcn_size (fcn);
+
+		bool flag = false;
+		int tag = core->dbg->trace->tag;
+		RListIter *iter;
+		RDebugTracepoint *trace;
+		r_list_foreach (core->dbg->trace->traces, iter, trace) {
+		  if (!trace->tag || (tag & trace->tag)) {
+			if (trace->addr >= fcn->addr && trace->addr <= max_addr) {
```
radare: thats not how you c heck if an offset is inside a function. there's an api  for that. functions are not  linear

libr/core/canal.c
```diff
@@ -2525,6 +2525,33 @@ static int fcn_print_legacy(RCore *core, RAnalFunction *fcn) {
...
+		r_cons_printf ("(dbg.trace unset)\n");
+	} else if (core->dbg->trace->enabled == 1) {
+		unsigned long long max_addr = fcn->addr;
+		max_addr += (ut64)r_anal_fcn_size (fcn);
+
+		bool flag = false;
+		int tag = core->dbg->trace->tag;
+		RListIter *iter;
+		RDebugTracepoint *trace;
+		r_list_foreach (core->dbg->trace->traces, iter, trace) {
```
radare: i was suggesting to walk the instructions of the function and chk the trace info from there, but maybe for some cases this reversed approach is faster. im  just caring about feature right now, so i would merge if it works and makes sense the way its organized. we can find faster paths later

libr/core/canal.c
```diff
@@ -2519,12 +2519,39 @@ static int fcn_print_detail(RCore *core, RAnalFunction *fcn) {
...
+
 static int fcn_print_legacy(RCore *core, RAnalFunction *fcn) {
 	RListIter *iter;
 	RAnalRef *refi;
 	RList *refs, *xrefs;
 	int ebbs = 0;
 	char *name = r_core_anal_fcn_name (core, fcn);
+
+	// execution times:
+	r_cons_printf ("execution state: ");
```
radare: i think this info should be printed below the other lines, because 'offset' should be the first row for each function  display

radare:
Please change commit message and PR description with "Fix #number - descrption"

radare:
also rebase.

libr/core/canal.c
```diff
@@ -2519,12 +2519,39 @@ static int fcn_print_detail(RCore *core, RAnalFunction *fcn) {
...
+	if (!core->dbg->trace->enabled) {
```
radare: just use } else { below

libr/core/canal.c
```diff
@@ -2519,12 +2519,39 @@ static int fcn_print_detail(RCore *core, RAnalFunction *fcn) {
 	return 0;
 }
 
+static bool is_fcn_traced(RDebugTrace *traced, RAnalFunction *fcn) {
+	int tag = traced->tag;
+	RListIter *iter;
+	RDebugTracepoint *trace;
+
+	r_list_foreach (traced->traces, iter, trace) {
```
radare: just use this api `R_API RDebugTracepoint *r_debug_trace_get(RDebug *dbg, ut64 addr); ` to check if the first instruction has been traced or not. this should be enough to determine if the function ist raced.

libr/core/canal.c
```diff
@@ -2519,12 +2519,39 @@ static int fcn_print_detail(RCore *core, RAnalFunction *fcn) {
...
+		if (!trace->tag || (tag & trace->tag)) {
```
radare: to check the code coverage we need to deteremine how many instructions of the function has been traced and how many the function contains. we can probably split this problem in a separate pr

libr/core/canal.c
r4dix10: https://github.com/radare/radare2/pull/12798#discussion_r248018901

r4dix10: > just use this api `R_API RDebugTracepoint *r_debug_trace_get(RDebug *dbg, ut64 addr); ` to check if the first instruction has been traced or not. this should be enough to determine if the function ist raced.

`e dbg.trace=1`
`db 0x4004d7`
`dc` -> hits breakpoint
`dso`

in this above case checking only the first instruction won't work. we need to go through the debug trace list of addresses and check if it exists in the fcn, if yes then fcn is being traced.

libr/core/canal.c
```diff
@@ -2519,12 +2519,39 @@ static int fcn_print_detail(RCore *core, RAnalFunction *fcn) {
...
 	RAnalRef *refi;
 	RList *refs, *xrefs;
 	int ebbs = 0;
 	char *name = r_core_anal_fcn_name (core, fcn);
+
+	// execution times:
+	r_cons_printf ("execution state: ");
+	if (!core->dbg->trace->enabled) {
+		r_cons_printf ("(dbg.trace unset)\n");
+	} else if (core->dbg->trace->enabled) {
```
radare: Swap conditional bodies and remove the second else which is the same as the inverse above

r4dix10:
@radare any more changes ?

libr/core/canal.c
```diff
@@ -2519,12 +2519,29 @@ static int fcn_print_detail(RCore *core, RAnalFunction *fcn) {
...
 
+static bool is_fcn_traced(RDebugTrace *traced, RAnalFunction *fcn) {
+	int tag = traced->tag;
+	RListIter *iter;
+	RDebugTracepoint *trace;
+
+	r_list_foreach (traced->traces, iter, trace) {
+		if (!trace->tag || (tag & trace->tag)) {
+			if (r_anal_fcn_in (fcn, trace->addr)) {
+				r_cons_printf ("%d\n", trace->times);
```
radare: Should eb acumulati g all the times all the instructions covered by a function has been executed. Also the way you iterate doesnt allows u to identify the function coverage %

radare:
Rename “execution state” with “traced” and dont show it if theres nothing to show. That will make travis green.

libr/core/canal.c
r4dix10: https://github.com/radare/radare2/pull/12798#discussion_r248486907

should fcn coverage be implemented in this pr only ?

>Also the way you iterate doesnt allows u to identify the function coverage %

the above code will be slightly modified for storing total size (traced_fcn_size) of all the instructions of function being traced. then we will take (traced_fcn_size / fcn_size) to show coverage %. Is this approach good ?

radare:
Thanks!

PR #12801, opened 2019-01-15
Rebase Breakpoints for PIEs ##debug

fcasal:
Previsouly the breakpoints set while not debugging would not be hit in the case of a PIE binary.


libr/debug/debug.c
```diff
@@ -1673,3 +1673,19 @@ R_API ut64 r_debug_get_baddr(RDebug *dbg, const char *file) {
 	return 0LL;
 #endif
 }
+
+R_API void r_debug_bp_rebase(RDebug *dbg, ut64 baddr) {
+	RBreakpointItem *bp;
+	RListIter *iter;
+	r_list_foreach (dbg->bp->bps, iter, bp) {
+		bp->addr += baddr;
```
radare: this seems like a bad idea. maybe the breakpoints should store more information in order to hold the reference to the associated map or bin, as long as this can be very different between versions it can be tricky.

this code can:

* overflow the addr value
* if called twice will result on invalid results

What about holding bp->oaddr and make bp->addr = bp->oaddr + baddr ? we will still need a way to reload the breakpoints to the right places

libr/core/cmd_open.c
```diff
@@ -1479,6 +1480,10 @@ static int cmd_open(void *data, const char *input) {
 						r_config_set (core->config, "cfg.debug", "false");
 						ut64 orig_baddr = sdb_num_get (core->sdb, "orig_baddr", 0);
 						r_bin_set_baddr (core->bin, orig_baddr);
+						ut64 baddr = r_config_get_i (core->config, "bin.baddr");
```
radare: bin.baddr should die, we should avoid using it

fcasal: Is there an alternative?

libr/debug/debug.c
fcasal: That seems a better way to do it, yes. If fixed the issue when calling this twice would invalidate. When unrebasing I update the orig_addr of all of them but it should be the same for those that were rebased, and the correct one for those bps added during the debug

radare:
I will introduce new apis to solve this later today. Hopefully we can remove some stuff from rbin soon

libr/core/cmd_open.c
radare: not yet

radare: i mean you can get the current bin and check the baddr of it by using the r_bin_get_cur() api, its probably better than relay on bin.baddr, the thng is that having a variable is useful for the user as input to change the baddr to any other value. we should probably decide what to do in here before , tehre are other 16  places in core that this var is used

radare:
comments on this @ret2libc @thestr4ng3r ?

fcasal:
Btw, there are still some issues with the PR, I will redo it tomorrow.

libr/debug/debug.c
```diff
@@ -1673,3 +1673,20 @@ R_API ut64 r_debug_get_baddr(RDebug *dbg, const char *file) {
...
+
+R_API void r_debug_bp_rebase(RDebug *dbg, ut64 baddr) {
+	RBreakpointItem *bp;
+	RListIter *iter;
+	r_list_foreach (dbg->bp->bps, iter, bp) {
+		bp->addr = bp->orig_addr + baddr;
+	}
+}
+
+R_API void r_debug_bp_unrebase(RDebug *dbg, ut64 baddr) {
```
ret2libc: Is this function really necessary? If you "rebase" all breakpoints, to "unrebase" them just means to "rebase" with baddr=0 IMHO.

radare:
we have the same rebasing problem in rbin when ood or reloading the p rocess. so we shuold solve it in the same or similar way

xarkes:
Might fix https://github.com/radareorg/cutter/issues/1053

libr/debug/debug.c
radare: we can do a quickfix by merging this, but this definively needs to be rethinked and done in  sync with rbin

fcasal:
Don't merge yet, there are still issues with the PR I haven't had time to fix yet.

radare:
if  you dont want me to merge something please prefix your pr  title with "WIP"

fcasal:
I fixed the issues I had found with the previous commit. 
I added a baddr field to RBreakpoint so that we can keep track of the "unrebased" address when adding a breakpoint.

libr/bp/bp.c
```diff
@@ -167,6 +167,9 @@ static RBreakpointItem *r_bp_add(RBreakpoint *bp, const ut8 *obytes, ut64 addr,
 	}
 	b = r_bp_item_new (bp);
 	b->addr = addr + bp->delta;
+	if (bp->baddr > addr)
```
radare: missing braces

libr/bp/bp.c
```diff
@@ -167,6 +167,9 @@ static RBreakpointItem *r_bp_add(RBreakpoint *bp, const ut8 *obytes, ut64 addr,
...
+		eprint ("base addr should not be larger than the breakpoint address.");
```
radare: do you mean eprintf with a newline?

radare: why this condition? and why this condition is done before using the variables involved?

fcasal: This condition should never happen but if it does an error should be printed

radare:
As a wip change it works if u only setup breakpounts in the program. But we need a solution to make this work for any module not just the main executable. Because each module can be loaded at different places

xarkes:
So? Waiting for a fix to also rebase breakpoints in modules?

libr/bp/bp.c
```diff
@@ -167,6 +167,10 @@ static RBreakpointItem *r_bp_add(RBreakpoint *bp, const ut8 *obytes, ut64 addr,
 	}
 	b = r_bp_item_new (bp);
```
radare: nullchk

libr/bp/bp.c
```diff
@@ -167,6 +167,10 @@ static RBreakpointItem *r_bp_add(RBreakpoint *bp, const ut8 *obytes, ut64 addr,
...
 	b->addr = addr + bp->delta;
+	if (bp->baddr > addr) {
+		eprintf ("base addr should not be larger than the breakpoint address.\n");
+	}
+	b->orig_addr = addr - bp->baddr;
```
radare: this is more like a delta

radare: so i would find a better name for this. also it will be good to see some feedback from @ret2libc because this baddr thing is very related to the rbin refactoring

fcasal:
Changed the name to delta and added a null check.
And indeed this would not work properly with breakpoints which are not in the main binary.

radare:
ok, i think im gonna merge this, it should fix many common use cases that people is complaining and the correct fix requires too much refactorings and discussions for now and its similar to the support for projects

libr/core/cmd_open.c
ret2libc: this does not seem to be used at all, am i right?

fcasal: You are right! fixed

radare:
Please rebase, appveyour should be green

radare:
Maybe if we define the size of the current map we can determine if the bp is inside the program region or not, and at least show a warning to the user about a speicifc breakpoint not being rebased, otherwise im afraid to merge this and introduce more unexpected behaviours

xarkes:
Then don't merge, I think it's not good to merge unstable stuff.
Also, I think we need to reintroduce the debugger tests in the testsuite, that could help catching those undefined behaviors.

radare:
we have 3 weeks before the release. can we have a good solution for this pr in a week imho, but we can move to 3.5 and focus on debug then..

btw, imho the solution can be just dpeending on the debug maps. tracking the name+permissions as unique-id to determine the base address of the current map to get the delta for the bp.

can you try to implement this? i think that may work in all situations.. except when the map is not there, then we should disable the bp and show a warning

radare:
Please check this issue https://github.com/radare/radare2/issues/13753 and comment it

radare:
Related to https://github.com/radare/radare2/issues/13753

radare:
Can you please rebase and fix the conflicts?

fcasal:
I can do it tonight

radare:
Good!

fcasal:
I have not looked at the code in a long time, but I see that you now rebase flags, functions, etc in the rebase_everything function. Do you want me to rebase the breakpoints in there as well?

radare:
👍

ITAYC0HEN:
ready to be merged? :)

PR #12802, opened 2019-01-15
Fix missing comma bug in pj.c fixes invalid json in afcfj

radare:


radare:
Ths fixes r2dec support

PR #12810, opened 2019-01-16
File hash refactoring

devnexen:
First step of computing md5/sha1 from core to bin pov.
Still using config to store for now.

### DIFF
--- a/binr/radare2/radare2.c
+++ b/binr/radare2/radare2.c
@@ -1319,7 +1319,7 @@ int main(int argc, char **argv, char **envp) {
 			if (has_project) {
 				r_config_set (r.config, "bin.strings", "false");
 			}
-			if (iod && r_core_hash_load (&r, iod->name) == false) {
+			if (iod && r_bin_file_hash_load (r.bin, r.config, iod->name) == false) {
 				//eprintf ("WARNING: File hash not calculated\n");
 			}
 			nsha1 = r_config_get (r.config, "file.sha1");
--- a/binr/radare2/radare2.c
+++ b/binr/radare2/radare2.c
@@ -1314,12 +1314,13 @@ int main(int argc, char **argv, char **envp) {
 			const char *npath, *nsha1;
 			char *path = strdup (r_config_get (r.config, "file.path"));
 			char *sha1 = strdup (r_config_get (r.config, "file.sha1"));
+			ut64 limit = r_config_get_i (r.config, "cfg.hashlimit");
 			has_project = r_core_project_open (&r, r_config_get (r.config, "prj.name"), threaded);
 			iod = r.io ? r_io_desc_get (r.io, fh->fd) : NULL;
 			if (has_project) {
 				r_config_set (r.config, "bin.strings", "false");
 			}
-			if (iod && r_bin_file_hash_load (r.bin, r.config, iod->name) == false) {
+			if (iod && r_bin_file_hash (r.bin, limit, iod->name) == false) {
 				//eprintf ("WARNING: File hash not calculated\n");
 			}
 			nsha1 = r_config_get (r.config, "file.sha1");
--- a/libr/bin/Makefile
+++ b/libr/bin/Makefile
@@ -2,7 +2,7 @@ include ../config.mk
 include ../../global.mk
 
 NAME=r_bin
-DEPS=r_util r_io r_socket r_magic r_hash
+DEPS=r_util r_io r_socket r_magic r_hash r_config
 
 .PHONY: pre
 
--- a/libr/bin/Makefile
+++ b/libr/bin/Makefile
@@ -2,7 +2,7 @@ include ../config.mk
 include ../../global.mk
 
 NAME=r_bin
-DEPS=r_util r_io r_socket r_magic r_hash r_config
+DEPS=r_util r_io r_socket r_magic r_hash
 
 .PHONY: pre
 
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -884,7 +884,7 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
 	r_hash_do_end (ctx, R_HASH_SHA1);
 	p = hash;
 	r_hex_bin2str (ctx->digest, R_HASH_SIZE_SHA1, p);
-	r_strbuf_appendf (o->info->hashes, " sha1 %s", hash);
+	r_strbuf_appendf (o->info->hashes, "\nsha1 %s", hash);
 	r_hash_free (ctx);
 	free (buf);
 	return true;
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -878,6 +878,7 @@ R_API int r_bin_file_hash_load(RBin *bin, RConfig *config, const char *file) {
 		p += 2;
 	}
 	*p = 0;
+	bf->o->info->file_hash[R_BIN_FILE_HASH_MD5] = r_str_ndup (hash, R_HASH_SIZE_MD5);
 	r_config_set (config, "file.md5", hash);
 	r_hash_free (ctx);
 	ctx = r_hash_new (true, R_HASH_SHA1);
@@ -888,6 +889,7 @@ R_API int r_bin_file_hash_load(RBin *bin, RConfig *config, const char *file) {
 		p += 2;
 	}
 	*p = 0;
+	bf->o->info->file_hash[R_BIN_FILE_HASH_SHA1] = r_str_ndup (hash, R_HASH_SIZE_SHA1);
 	r_config_set (config, "file.sha1", hash);
 	r_hash_free (ctx);
 	free (buf);
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -864,10 +864,12 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
 	}
 
 	buf_len = r_io_desc_size (iod);
-	if (buf_len > limit)
-		buf_len = limit;
-	buf = calloc (1, buf_len);
-	r_io_desc_read (iod, buf, buf_len);
+	// By SLURP_LIMIT normally cannot compute ...
+	if (buf_len > limit) {
+		eprintf ("Cannot compute hash\n");
+		return -1;
+	}
+	buf = r_file_slurp (file, &buf_len);
 	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1);
 #define BLK_SIZE_OFF 1024
 	for (i = 0; i < buf_len; i += BLK_SIZE_OFF) {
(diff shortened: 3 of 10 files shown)


radare:
see https://github.com/radare/radare2/issues/12648

devnexen:
Fair point I planned to remove the config in another commit, I guess it might require quite some unit test changes ...

devnexen:
If you like the actual form, I may start changing the various rabin2/r2 i tests ...

binr/radare2/radare2.c
```diff
@@ -1319,7 +1319,7 @@ int main(int argc, char **argv, char **envp) {
 			if (has_project) {
 				r_config_set (r.config, "bin.strings", "false");
 			}
-			if (iod && r_core_hash_load (&r, iod->name) == false) {
+			if (iod && r_bin_file_hash_load (r.bin, r.config, iod->name) == false) {
```
radare: bad naming, r_bin_file expects to be something using the RBinFile . also _load() expects to parse the headers and such. i would say.. `r_bin_file_hash()` and use the current RBinObject loaded to recompute the checksums.

libr/bin/Makefile
```diff
@@ -2,7 +2,7 @@ include ../config.mk
 include ../../global.mk
 
 NAME=r_bin
-DEPS=r_util r_io r_socket r_magic r_hash
+DEPS=r_util r_io r_socket r_magic r_hash r_config
```
radare: uhm nope. dont do that

libr/bin/bfile.c
```diff
@@ -847,3 +848,50 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
 	}
 	return false;
 }
+
+R_API int r_bin_file_hash_load(RBin *bin, RConfig *config, const char *file) {
```
radare: bad naming

libr/bin/bfile.c
```diff
@@ -847,3 +848,50 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+	RBinFile *bf = bin->cur;
+	if (!file && bf) {
+		file = bf->file;
+	}
+	
+	limit = r_config_get_i (config, "cfg.hashlimit");
+	if (bf && bf->size > limit) {
+		return false;
+	}
+	buf = (ut8 *) r_file_slurp (file, &buf_len);
```
radare: dangerous, ths can eat all memory.

you shuold use the io api via the RIOBind instance using the fd associated to the RBinFIle selected (dont use bin->cur, take that as argument of the function)

libr/bin/bfile.c
```diff
@@ -847,3 +848,50 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+	if (!buf) {
+		return false;
+	}
+	ctx = r_hash_new (true, R_HASH_MD5);
+	md5 = r_hash_do_md5 (ctx, buf, buf_len);
```
radare: this is very slow because it requires a lot of memory. use the update method and hash per blocks to finalize at the end. this is the only way to get it memory and cpu efficient and

libr/bin/bfile.c
```diff
@@ -847,3 +848,50 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+	p = hash;
+	for (i = 0; i < R_HASH_SIZE_MD5; i++) {
```
radare: What about using RStrBuf and store the hashes as a string in `name  value` , like `md5 038943939485..`.

this way we are not limited to only one hash

libr/bin/bfile.c
```diff
@@ -847,3 +848,50 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+		sprintf (p, "%02x", md5[i]);
+		p += 2;
+	}
+	*p = 0;
+	bf->o->info->file_hash[R_BIN_FILE_HASH_MD5] = r_str_ndup (hash, R_HASH_SIZE_MD5);
+	r_config_set (config, "file.md5", hash);
```
radare: dont use rconfig api. the hash info must be exposed to core via the api or rbinobj fields.

libr/bin/bfile.c
radare: we can also store the results in a hashtable but thats proabbly a bit overengineered for just 2-4 values

libr/bin/bfile.c
```diff
@@ -847,3 +848,50 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+	ctx = r_hash_new (true, R_HASH_SHA1);
+	sha1 = r_hash_do_sha1 (ctx, buf, buf_len);
+	p = hash;
+	for (i = 0; i < R_HASH_SIZE_SHA1; i++) {
+		sprintf (p, "%02x", sha1[i]);
+		p += 2;
+	}
+	*p = 0;
+	bf->o->info->file_hash[R_BIN_FILE_HASH_SHA1] = r_str_ndup (hash, R_HASH_SIZE_SHA1);
+	r_config_set (config, "file.sha1", hash);
```
radare: dont use rconfig

libr/bin/bin.c
```diff
@@ -125,6 +125,8 @@ R_API void r_bin_info_free(RBinInfo *rb) {
 		return;
 	}
 	free (rb->intrp);
+	free (rb->file_hash[R_BIN_FILE_HASH_MD5]);
+	free (rb->file_hash[R_BIN_FILE_HASH_SHA1]);
```
radare: drop. just use a string

libr/bin/meson.build
```diff
@@ -147,7 +147,8 @@ pkgconfig_mod.generate(r_bin,
   requires: pkgconfig_magic_requires + [
     'r_util',
     'r_io',
-    'r_socket'
+    'r_socket',
+    'r_config'
```
radare: nope

libr/core/cbin.c
```diff
@@ -698,6 +698,8 @@ static int bin_info(RCore *r, int mode, ut64 laddr) {
 		pair_ut64 ("binsz", r_bin_get_size (r->bin), mode, false);
 		pair_str ("bintype", info->rclass, mode, false);
 		pair_int ("bits", info->bits, mode, false);
+		pair_str ("md5hash", info->file_hash[R_BIN_FILE_HASH_MD5], mode, false);
```
radare: print the string containing all the hash=values

libr/bin/bfile.c
```diff
@@ -1,7 +1,10 @@
 /* radare2 - LGPL - Copyright 2009-2018 - pancake, nibble, dso */
 
 #include <r_bin.h>
+#include <r_hash.h>
 #include "i/private.h"
+#include "hash/md5.h"
```
radare: use r_hash api which i s generic and works for all the hash algorithsm  at the same time, no need to compute them by separate

radare:
i think that if we make the hashes appear in a different place than `i`  we wll get 0 broken tests.

another idea is to make the file.hash  variable empty  by  default and not show any hash. this way it  will be green, but keep the functionality in there

libr/bin/bfile.c
```diff
@@ -847,3 +848,55 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+	}
+
+	buf_len = r_io_desc_size (iod);
+	if (buf_len > limit)
+		buf_len = limit;
+	buf = calloc (1, buf_len);
+	r_io_desc_read (iod, buf, buf_len);
+	ctx = r_hash_new (false, R_HASH_MD5);
+	for (i = 0; i < buf_len; i ++) {
+		(void)r_hash_do_md5 (ctx, &buf[i], 1);
```
radare: dont do byte per  byte, thats ultra slow, use blocks like 1024 or 4096

libr/bin/bfile.c
```diff
@@ -847,3 +848,55 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+	for (i = 0; i < R_HASH_SIZE_MD5; i++) {
+		sprintf (p, "%02x", ctx->digest[i]);
+		p += 2;
+	}
+	*p = 0;
+	char fmt[256];
+	snprintf (fmt, sizeof (fmt), "md5:%s", hash);
+	o->info->hashes = r_strbuf_new (fmt);
+	r_hash_free (ctx);
+	ctx = r_hash_new (false, R_HASH_SHA1);
```
radare: you can set more than one  hash  in  this r_hash_new(), the  loop will  update all the defined hashes and fill the results for all of them at the end.  no  need to iterate twice

libr/bin/bfile.c
```diff
@@ -847,3 +848,55 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+	for (i = 0; i < buf_len; i ++) {
+		(void )r_hash_do_sha1 (ctx, &buf[i], 1);
+	}
+	r_hash_do_end (ctx, R_HASH_SHA1);
+	p = hash;
+	for (i = 0; i < R_HASH_SIZE_SHA1; i++) {
+		sprintf (p, "%02x", ctx->digest[i]);
```
radare: we have a function  to r_bin2hexstr

libr/bin/bfile.c
```diff
@@ -847,3 +848,55 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+		p += 2;
+	}
+	*p = 0;
+	snprintf (fmt, sizeof (fmt), ",sha1:%s", hash);
```
radare: snprintf i s evil. you have r_strbuf_appendf()

radare: not sure if we should store this in json.. maybe  this way will work seamlessly in ij

libr/bin/bfile.c
```diff
@@ -847,3 +848,41 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+	}
+
+	buf_len = r_io_desc_size (iod);
+	if (buf_len > limit)
+		buf_len = limit;
+	buf = calloc (1, buf_len);
+	r_io_desc_read (iod, buf, buf_len);
+	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1);
+	for (i = 0; i < buf_len; i += 1024) {
+		(void)r_hash_do_md5 (ctx, &buf[i], 1024);
```
radare: use define or const to define this 1024 insead of copypasta

libr/bin/bfile.c
```diff
@@ -847,3 +848,41 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+		(void)r_hash_do_sha1 (ctx, &buf[i], 1024);
+	}
```
radare: the last block (if file is not % 1024) is not hashed

radare: err i mean it will hash more bytes than the ones available in the buffer, you must R_MIN(1024, buf_len-i)

libr/bin/bfile.c
```diff
@@ -847,3 +848,41 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
...
+
+	buf_len = r_io_desc_size (iod);
+	if (buf_len > limit)
+		buf_len = limit;
+	buf = calloc (1, buf_len);
+	r_io_desc_read (iod, buf, buf_len);
+	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1);
+	for (i = 0; i < buf_len; i += 1024) {
+		(void)r_hash_do_md5 (ctx, &buf[i], 1024);
+		(void)r_hash_do_sha1 (ctx, &buf[i], 1024);
```
radare: use buf+i, is easier to read

libr/bin/bfile.c
devnexen: Ah right ...

devnexen:
> do you want to do the support for many hashes in this pr? instead of hardcoded md5/sha1

For now ... as is.

devnexen:
If you guys want an end of line for each hash type, no problems I can ... :-)

Maijin:
Yes please, otherwise can't use grep anymore

Maijin:
Fix what @radare found, click on resolve conversation when done and I'll fix the tests

libr/bin/bfile.c
devnexen: I tend to disagree on this one :)

radare:
imho it will be better to not have somethng wip that breaks 70 tests, so i would vote for having those hashes in eprintf (stderr) to not break any  test, or add a different command to show those bin hashes.. and we can compute them from a command in r2, unlinking that from rbin too

devnexen:
Ok so you do not want them by `i` no pb ... I ll remove it from there.

PR #12815, opened 2019-01-16
Move Palette to RConsContext

### LINKED ISSUE #12644: Crash  in bg webserver
`$ r2 -c'=h&' -cVp /bin/ls`

thestr4ng3r:
TODO:
- Copy palette instead of initializing default when creating a new context, so a new task inherits the palette from the one it is started from
- Kill the global `curtheme` (and maybe `getNext`) in cmd_eval.c and move it to RConsContext

fixes #12644

### DIFF
--- a/libr/config/config.c
+++ b/libr/config/config.c
@@ -523,6 +523,18 @@ R_API int r_config_rm(RConfig *cfg, const char *name) {
 	return false;
 }
 
+R_API void r_config_node_value_format_i(char *buf, size_t buf_size, const ut64 i, R_NULLABLE RConfigNode *node) {
+	if (node && node->flags & CN_BOOL) {
+		r_str_ncpy (buf, r_str_bool ((int) i), buf_size);
+		return;
+	}
+	if (i < 1024) {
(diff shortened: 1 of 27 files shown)


radare:
Do you want to split this process in different prs?

thestr4ng3r:
Ok well I just weren't aware of your pr. But mine is much further already anyway.

So right now the state of this pr should be pretty much stable (when the few feat_arroba tests are fixed) and the only undesired behaviour is the following:
- When starting a new task, it always has the default palette loaded.
- When switching the color theme with `eco` from different tasks, then calling `eco` from any task, the list just shows the last theme switched to, not the actual current one from the current task.

So just minor issues. Depends if you are ok with merging it like this.

thestr4ng3r:
wtf why is travis not using this r2r pr: https://github.com/radare/radare2-regressions/pull/1652

radare:
if its wip change the pr title

thestr4ng3r:
That's the question: Do we consider it wip and continue to work on this pr or merge it now and make new prs for the changes mentioned?

PR #12823, opened 2019-01-17
Prevent runaway scrolling in Visual when using arrow keys ##visual

kazarmy:


libr/core/visual.c
```diff
@@ -4052,6 +4052,21 @@ R_API int r_core_visual(RCore *core, const char *input) {
 #ifndef __WINDOWS__
 			if (IS_PRINTABLE (ch) || ch == '\t' || ch == '\n') {
 				tcflush (STDIN_FILENO, TCIFLUSH);
+			} else if (ch == 0x1b) {
+				char chrs[2];
+				chrs[0] = r_cons_readchar ();
+				if (chrs[0] == '[') {
+					chrs[1] = r_cons_readchar ();
+					if (chrs[1] >= 'A' && chrs[1] <= 'D') { // arrow keys
+						tcflush (STDIN_FILENO, TCIFLUSH);
```
radare: this should be a bug in  the terminal. n mac i was having different behavour, which terminal are you testing this on?  try in tty

kazarmy: it's xfce4-terminal ... bug doesn't repro on a tty virtual console nor on gnome-terminal

still, we can't assume that the user is using a non-buggy terminal

radare:
yeah just to clarify :)

PR #12854, opened 2019-01-20
Printkey json cb show format pj 

a-meth:


libr/core/cmd_type.c
```diff
@@ -268,7 +277,10 @@ static int printkey_cb(void *user, const char *k, const char *v) {
 }
 
 static int printkey_json_cb(void *user, const char *k, const char *v) {
-	r_cons_printf ("\"%s\"", k);
+	PJ *pj = pj_new ();
```
a-meth: What should the nullchk here return?

radare: should return 1 otherwise the foreach will stop. but well. maybe if the memory allocation fails is better to return  0  and stop the rest of iterations from happening

radare: so yeah, return 0

PR #12858, opened 2019-01-20
Speed up RBinReloc retrieval ##optimization

ret2libc:
Use RBTree to speed up relocation search, in particular during disassembly.

On a Linux Kernel with debug info which has 17569271 relocations, this is a huge performance improvement. Without this patch, it was very slow to scroll in visual mode while disassembling.

### DIFF
--- a/libr/bin/bin.c
+++ b/libr/bin/bin.c
@@ -752,32 +752,13 @@ R_API RList *r_bin_get_libs(RBin *bin) {
 	return o ? o->libs : NULL;
 }
 
-R_API RList *r_bin_patch_relocs(RBin *bin) {
+R_API RBNode *r_bin_patch_relocs(RBin *bin) {
 	r_return_val_if_fail (bin, NULL);
-	static bool first = true;
 	RBinObject *o = r_bin_cur_object (bin);
-	if (!o) {
-		return NULL;
-	}
-	// r_bin_object_set_items set o->relocs but there we don't have access
-	// to io
-	// so we need to be run from bin_relocs, free the previous reloc and get
-	// the patched ones
-	if (first && o->plugin && o->plugin->patch_relocs) {
-		RList *tmp = o->plugin->patch_relocs (bin);
-		first = false;
-		if (!tmp) {
-			return o->relocs;
-		}
-		r_list_free (o->relocs);
-		o->relocs = tmp;
-		REBASE_PADDR (o, o->relocs, RBinReloc);
-		first = false;
-	}
-	return o->relocs;
+	return o? r_bin_object_patch_relocs (bin, o): NULL;
 }
 
-R_API RList *r_bin_get_relocs(RBin *bin) {
+R_API RBNode *r_bin_get_relocs(RBin *bin) {
 	r_return_val_if_fail (bin, NULL);
 	RBinObject *o = r_bin_cur_object (bin);
 	return o ? o->relocs : NULL;
--- a/libr/bin/obj.c
+++ b/libr/bin/obj.c
@@ -18,6 +18,17 @@ static void mem_free(void *data) {
 	free (mem);
 }
 
+static int reloc_cmp(const void *a, const RBNode *b) {
+	RBinReloc *ar = (const RBinReloc *)a;
+	RBinReloc *br = container_of (b, const RBinReloc, vrb);
+	return ar->vaddr - br->vaddr;
+}
+
+static void reloc_free(RBNode *rbn) {
+	RBinReloc *r = container_of (rbn, RBinReloc, vrb);
+	free (r);
+}
+
 static void object_delete_items(RBinObject *o) {
 	ut32 i = 0;
 	r_return_if_fail (o);
@@ -26,7 +37,7 @@ static void object_delete_items(RBinObject *o) {
 	r_list_free (o->fields);
 	r_list_free (o->imports);
 	r_list_free (o->libs);
-	r_list_free (o->relocs);
+	r_rbtree_free (o->relocs, reloc_free);
 	r_list_free (o->sections);
 	r_list_free (o->strings);
 	ht_up_free (o->strings_db);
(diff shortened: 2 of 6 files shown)


libr/bin/obj.c
```diff
@@ -18,6 +18,23 @@ static void mem_free(void *data) {
 	free (mem);
 }
 
+static int reloc_cmp(const void *a, const RBNode *b) {
+	const RBinReloc *ar = (const RBinReloc *)a;
+	const RBinReloc *br = container_of (b, const RBinReloc, vrb);
+	if (ar->vaddr > br->vaddr) {
+		return 1;
+	} else if (ar->vaddr < br->vaddr) {
```
radare: im not a fan of those else after return

libr/bin/obj.c
```diff
@@ -18,6 +18,23 @@ static void mem_free(void *data) {
...
+		return -1;
+	} else {
+		return 0;
+	}
+}
+
+static void reloc_free(RBNode *rbn) {
+	RBinReloc *r = container_of (rbn, RBinReloc, vrb);
+	free (r);
```
radare: just free (container_of(..) in 1 line

libr/core/disasm.c
```diff
@@ -3857,22 +3857,37 @@ static void ds_print_ptr(RDisasmState *ds, int len, int idx) {
...
+
+static int getreloc_tree(const void *user, const RBNode *n) {
+	struct getreloc_t *gr = (struct getreloc_t *)user;
+	const RBinReloc *r = container_of (n, const RBinReloc, vrb);
+	if ((r->vaddr >= gr->vaddr) && (r->vaddr < (gr->vaddr + gr->size))) {
+		return 0;
+	}
+
+	if (gr->vaddr > r->vaddr) {
+		return 1;
```
radare: no need to else

libr/core/disasm.c
```diff
@@ -3857,22 +3857,37 @@ static void ds_print_ptr(RDisasmState *ds, int len, int idx) {
...
-	list = r_bin_get_relocs (core->bin);
-	r_list_foreach (list, iter, r) {
-		if ((r->vaddr >= addr) && (r->vaddr < (addr + size))) {
-			return r;
-		}
-	}
-	return NULL;
+	RBNode *relocs = r_bin_get_relocs (core->bin);
+	struct getreloc_t gr = { .vaddr = addr, .size = size };
+	RBNode *res = r_rbtree_find (relocs, &gr, getreloc_tree);
```
radare: relocs can be null, you can return null earlier

libr/core/disasm.c
```diff
@@ -3857,22 +3857,37 @@ static void ds_print_ptr(RDisasmState *ds, int len, int idx) {
...
+	return res? container_of (res, RBinReloc, vrb): NULL;
```
radare: this container_of api is rbtree-specific, imho it should be named differently

radare: with r_rbtree_ prefix

libr/bin/bin.c
```diff
@@ -752,32 +752,13 @@ R_API RList *r_bin_get_libs(RBin *bin) {
...
-		o->relocs = tmp;
-		REBASE_PADDR (o, o->relocs, RBinReloc);
-		first = false;
-	}
-	return o->relocs;
+	return o? r_bin_object_patch_relocs (bin, o): NULL;
 }
 
-R_API RList *r_bin_get_relocs(RBin *bin) {
+R_API RBNode *r_bin_get_relocs(RBin *bin) {
```
radare: this breaks the api. what about adding the r_bin_get_relocs_rbtree() and keep the list?

radare: also, rbtree is not in bindings, so it will make more apis unaccessilbe from valabind unless we update the vapis

libr/bin/bin.c
```diff
@@ -752,32 +752,13 @@ R_API RList *r_bin_get_libs(RBin *bin) {
 	return o ? o->libs : NULL;
 }
 
-R_API RList *r_bin_patch_relocs(RBin *bin) {
+R_API RBNode *r_bin_patch_relocs(RBin *bin) {
```
radare: add r_bin_patch_relocs_list and make this function return bool. we already have the get_relocs() to get the rbnode

libr/bin/bin.c
radare: i think we should use RBTree instead of RBNode, and yep, the typedef must be changed.

radare:
maybe we want an api to get a list from an rbtree?

radare:
the tests are faling because of the order, i think the ordinal should be kept and list them in the order specified by the file

ret2libc:
> the tests are faling because of the order, i think the ordinal should be kept and list them in the order specified by the file

I'm not sure. At least on ELF (and all failing tests are ELFs) relocations are one of the few kinds of entries that do not have an "ordinal", so I think listing them by vaddr is the best thing.

libr/core/disasm.c
ret2libc: I don't think so. `container_of` is really rbtree specific. It's just a macro to get the address of an outer struct from the address of a field. It is used heavily in the linux kernel.

We should probably move it to r_util.h or in r_types.h instead of defining it in r_rbtree.h.

radare:
Thatvwill work for me too

libr/bin/bin.c
ret2libc: no, as this function is used right now it needs to return a RList/RBNode. We may think of whether this is good or not in additional refactoring IMHO.

radare:
RBTree is a typedef for RBNode. And yes i knowbthats how it is right now but its something to change imho

ret2libc:
> RBTree is a typedef for RBNode. And yes i knowbthats how it is right now but its something to change imho

I'm ok with this change and you are right. I was referring to making `r_bin_patch_relocs` return a bool instead of a RBNode/RBTree.

radare:
patch_relocs() is patching. theres no  need to get the same list as the one returned from get_relocs(), why it shouldnt return bool?

ret2libc:
> patch_relocs() is patching. theres no need to get the same list as the one returned from get_relocs(), why it shouldnt return bool?

ok yes you are right. But please let me do it in a separate PR, this has nothing to do with the change from rlist to rbtree.

ret2libc:
WRT RBNode -> RBTree, let's do it too in a separate thing. Shall we change all APIs to use RBTree instead of RBNode? Or only some? It may get a bit confusing if we don't change everything.

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1656

PR #12884, opened 2019-01-23
Add new command: pifc and pifcj

ITAYC0HEN:
This command will print all the `CALLS` from a given function. The API is now also used by `aflm` which was introduced last week and was adapted in this PR. 
While `aflm` will print all the calls from all the functions, the new `pifc` command would do that only for a specific function.

```
[0x00001347]> pifc
0x0000135d call sym.imp.puts
0x00001369 call sym.imp.puts
0x00001382 call sym.beet
0x00001392 call sym.imp.puts
0x000013a0 call sym.imp.puts
```

and in a JSON format:
```
[0x00001347]> pifcj~{}
[
  {
    "dest": "sym.imp.puts",
    "addr": 4160,
    "at": 4957
  },
  {
    "dest": "sym.imp.puts",
    "addr": 4160,
    "at": 4969
  },
  {
    "dest": "sym.beet",
    "addr": 4822,
    "at": 4994
  },
  {
    "dest": "sym.imp.puts",
    "addr": 4160,
    "at": 5010
  },
  {
    "dest": "sym.imp.puts",
    "addr": 4160,
    "at": 5024
  }
]
```

Screenshot:
image

### DIFF
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2341,6 +2341,26 @@ static int fcn_list_default(RCore *core, RList *fcns, bool quiet) {
 	return 0;
 }
 
+// for a given function returns an RList of all functions that were called in it
+R_API RList *r_core_anal_fcn_get_calls (RCore *core, RAnalFunction *fcn) {
+	RList *refs = NULL;
+	RAnalRef *refi;
+	RListIter *iter;
+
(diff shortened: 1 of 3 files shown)


Maijin:
@ITAYC0HEN don't forget to add tests

PR #12886, opened 2019-01-23
WIP: Fixes and tests to find out why the debugger is broken for Android

radare:


radare:
with this PR i broke the linux debugger in the same way the android debugger is, so its "easier" to test :P

radare:
looks like fixed now!

PR #12901, opened 2019-01-24
Shorten local var and type name ##types

### LINKED ISSUE #12656: Shorten arguments and locals
It probably makes sense to shorten local variable names from `local_8h` to `loc_8h` or `var_8h`,
also it makes sense to use `uint8_t` or even `u8`/`ut8` instead of full `unsigned int` or even worse `unsigned long long`.

See how it is too long for some cases:
screenshot_20190105_184540

sivaramaaa:
Fix #12656 

shortvar


sivaramaaa:
Still have to  update many test , so don't expect tarvis to be grin :)

radare:
u probably need to rebase because i broke master a bit ago

Maijin:
That is the format that got voted in the poll

sivaramaaa:
> That is the format that got voted in the poll

Yup , but I think both @XVilka and @radare are voting for even shorter name , if that's the case we have to reconsider the option

But afair Ida , r2dec follows `uint32_t` convention , also imho it's more clear than `ut32` option

ITAYC0HEN:
`uint32_t` is cleaner imo
 and well... people voted and I am quite sure that this question was quite clear :))

XVilka:
Lets stay with `uint8_t`, and merge this once ready. If we change our mind we can always change it later. Anyway `uint64_t` is already an improvement compared to `unsigned long long`...

libr/core/anal_tp.c
```diff
@@ -57,7 +57,7 @@ static void var_rename(RAnal *anal, RAnalVar *v, const char *name, ut64 addr) {
 	if (!*name || !strcmp (name , "...")) {
 		return;
 	}
-	bool is_default = (!strncmp (v->name, "local_", 6)
+	bool is_default = (!strncmp (v->name, "var_", 6)
```
xarkes: Why not reuse VARPREFIX macro here?

libr/core/anal_tp.c
```diff
@@ -120,6 +120,13 @@ static void var_retype(RAnal *anal, RAnalVar *var, const char *vname, char *type
 			r_strbuf_append (sb, " *");
 		}
 	}
+
+	char* tmp1 = r_strbuf_get (sb);
+	if (r_str_startswith(tmp1, "unsigned long long")) {
+		r_strbuf_set (sb, "uint64_t");
+	} else if (r_str_startswith(tmp1, "unsigned")) {
```
xarkes: Missing spaces before function call

radare:
Agree

radare:
so rebase and update tests and we merge

sivaramaaa:
Yup will do it today evening !

PR #12904, opened 2019-01-24
Rewrite RSpaces to use RBTree and remove use of indices everywhere

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1665


libr/util/spaces.c
```diff
@@ -1,213 +1,185 @@
...
 }
 
-static int r_space_count (RSpaces *s, int n) {
-	if (s->count_for) {
-		return s->count_for (s, n);
-	}
-	return 0;
+static int space_cmp(const void *incoming, const RBNode *rb) {
+	const RSpace *a = (const RSpace *)incoming;
+	const RSpace *b = container_of (rb, const RSpace, rb);
```
radare: though this container_of would be renamed before using it in more places

libr/util/spaces.c
```diff
@@ -1,213 +1,185 @@
...
 }
 
-R_API int r_space_list(RSpaces *s, int mode) {
-	const char *defspace = NULL;
-	int count, len, i, j = 0;
-	if (mode == 'j') {
-		s->cb_printf ("[");
+R_API RSpace *r_spaces_add(RSpaces *sp, const char *name) {
+	r_return_val_if_fail (sp, NULL);
+	if (!name || *name == '*') {
```
radare: what about !*name

libr/util/spaces.c
ret2libc: we agreed it was enough to move it to r_util.h. container_of is a very generic function, used also in the kernel. I really think the name is perfect. We can create an additional macro specific for RSpace (that uses container_of) if we want.

libr/util/spaces.c
ret2libc: right

ret2libc:
TODO: add some short comments to describe the API

radare:
solve conflicts and imho its ready to merge.

libr/core/core.c
```diff
@@ -2205,7 +2208,38 @@ static char *get_comments_cb(void *user, ut64 addr) {
...
+	}
+	if (mode == '*' && r_spaces_current (sp)) {
+		r_cons_printf ("%s %s # current\n", sp->name, r_spaces_current_name (sp));
+	}
+	if (mode == 'j') {
+		r_cons_printf ("]\n");
+	}
+}
+
+static void cb_event_handler(REvent *ev, int event_type, void *user, void *data) {
```
ret2libc: Completely unrelated to this PR, but @radare this is wrong. You should create separate functions for each event_type, instead of using a switch/case. It's much much cleaner. We all know how we ended up with switch/cases for commands :D

ret2libc:
@radare I think this can be merged.

radare:
We have already discused this a month ago . Do u wanna discuss it again?

radare:
Kk will review in few hours! Thanks

PR #12905, opened 2019-01-24
Remove invalid sleep from cmd_open()

thestr4ng3r:
This is more than just wrong. Sleep should only be used when absolutely no global stuff is accessed while active. And what is done here is the exact opposite of that.
Maybe something inside `r_core_file_open()` or `r_core_bin_load()` could be wrapped in sleep, but definitely not the whole functions.


radare:
r_core_file and r_core_bin must die, and open() can take a while or even block

thestr4ng3r:
Yes, then the sleep can later be added around open.

radare:
ok merge

Issue #12931, opened 2019-01-28
ragg2 not working on Arch and MacOSX


deepakchethan:
# This template is meant for bug reports, if you have a feature request, please be as descriptive as possible and delete the template

Make sure you are testing using the latest git version of radare2 before submitting any issue.

*If you would like to report a bug, please fill the template bellow*

### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Arch Linux x86_64
| File format of the file you reverse (mandatory)      | ELF 
| Architecture/bits of the file (mandatory)            | x86/32
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.2.0-git 20551 @ linux-x86-64 git.3.1.3-198-gbfa346286
commit: bfa3462866670931cb2efe9ae87f57ed2c477063 build: 2019-01-10__10:19:00

### Expected behavior
ragg2 hi.c must give a tiny binary
### Actual behavior
It is giving me an error
```
Buffer truncated at 124 because of the RAsmOp abuse.
Input: eb0e662e0f1f8400000000000f1f400050bf01000000488d359f000000ba06000000e81900000031ff89442404e85e00000031d289042489d059c30f1f440000897c24fc48897424f0895424ec8b5424fc895424dc488b7424f048897424d08b5424ec895424cc8b7c24dc488b7424d08b5424ccb8010000000f05488
Segmentation fault (core dumped)
```
### Steps to reproduce the behavior 
- Please share the binary if it is shareable by drag and dropping it here in a zip archive (mandatory)
- Use asciinema to describe the issue and share the link here (mandatory if u can't share the binary) 
- Use code markdown `CODE` to make your code visible
- Or even better, create a Pull-Request for the https://github.com/radare/radare2-regressions containing the test case examples can be found in the t/ folder see for example https://github.com/radare/radare2-regressions/blob/master/t/cmd_search#L7

The source code of hi.c
```
main() {
    write(1, "Hello", 6);
    exit(0);
}
```
and used the ragg2 command as
```
ragg2 hi.c
```

### Additional Logs, screenshots, source-code,  configuration dump, ...

Drag and drop zip archives containing the Additional info here, don't use external services or link.
The zip file contains the binary and the screenshot of the error message.
[ragg.zip](https://github.com/radare/radare2/files/2804274/ragg.zip)

[...]

PR #12935, opened 2019-01-29
Increased the RStrBuf size to avoid segmentation fault for ragg2

deepakchethan:
For issue #12931 
One of the bug fix is to increase the bufferSize. Which I did this time.

Other way I could think of is to handle it within the r_asm_psuedo_hex method and return if the the assumption fails. 

Please let me know, if I should handle it in any other way.


deepakchethan:
Okay, I'll start working on it.

Issue #12924, opened 2019-01-28
Segmentation fault while listing symbols in PDB


Quentin01:
# Segmentation fault while listing symbols in PDB

### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits                         | Debian arm 64, Ubuntu x86 32
| File format     | PDB
|File output|  `libmkv_plugin.pdb: MSVC program database ver 7.00, 1024*1707 bytes`
| r2 -v    | `radare2 3.3.0-git 0 @ linux-x86-64 git.3.3.0-git commit: HEAD build: 2019-01-25__10:51:29`

### Expected behavior

Rabin2 listing the symbols of the PDB when launching it like that: `LD_LIBRARY_PATH="./lib" ./bin/rabin2 -P libmkv_plugin.pdb`.

### Actual behavior

```
$ LD_LIBRARY_PATH="./lib" ./bin/rabin2 -P libmkv_plugin.pdb
WARNING: r_list_get_n: assertion 'list' failed (line 442)
WARNING: r_list_get_n: assertion 'list' failed (line 442)
WARNING: r_list_get_n: assertion 'list' failed (line 442)
WARNING: r_list_get_n: assertion 'list' failed (line 442)
WARNING: r_list_get_n: assertion 'list' failed (line 442)
WARNING: r_list_get_n: assertion 'list' failed (line 442)
WARNING: r_list_get_n: assertion 'list' failed (line 442)
...
WARNING: r_list_get_n: assertion 'list' failed (line 442)
WARNING: r_list_get_n: assertion 'list' failed (line 442)
WARNING: r_list_get_n: assertion 'list' failed (line 442)
WARNING: r_list_get_n: assertion 'list' failed (line 442)
WARNING: r_list_get_n: assertion 'list' failed (line 442)
Segmentation fault
```

### Steps to reproduce the behavior 

Just list the symbols of the PDB (`libmkv_plugin.pdb`).

### Additional Logs, screenshots, source-code,  configuration dump, ...

[libmkv_plugin.pdb.zip](https://github.com/radare/radare2/files/2802157/libmkv_plugin.pdb.zip)

radare:
the  crash is fixed, but we are not closing this issue because it requires a better fix. cc @AdamVanScyoc

XVilka:
Better fix wouldn't arrive in one day, moving ahead.

PR #12939, opened 2019-01-29
Replace calls to corrupted stream free functions #12924

AdamVanScyoc:


AdamVanScyoc:
Just realized my fix causes an enormous memory leak; don't merge, I'll fix it

radare:
Oops. read the comment after merging. At least its not segfaulting, but it's not the correct way to solve it, because those structs shouldnt be corrupted at all, i'll wait for another PR instead of reverting this one, ok?

PR #12946, opened 2019-01-29
Kill RIOSection for good##refactor

ret2libc:


libr/bin/p/bin_elf.inc
```diff
@@ -877,22 +877,24 @@ static RList* patch_relocs(RBin *b) {
...
 	if (!g) {
 		return NULL;
 	}
-	n_off = g->paddr + g->size;
-	n_vaddr = g->vaddr + g->vsize;
+	n_vaddr = g->itv.addr + g->itv.size;
 	//reserve at least that space
 	size = bin->reloc_num * 4;
-	if (!b->iob.section_add (io, n_off, n_vaddr, size, size, R_PERM_R, ".got.r2", 0, io->desc->fd)) {
+	RIOMap *gotr2map = b->iob.map_add (io, io->desc->fd, R_PERM_R, 0, n_vaddr, size);
```
ret2libc: a `null://` map should be added here, not the `io->desc->fd`.

ret2libc:
@radare this is ready to be reviewed!

libr/bin/p/bin_elf.inc
```diff
@@ -877,22 +877,31 @@ static RList* patch_relocs(RBin *b) {
...
+	free (muri);
+	if (!gotr2desc) {
 		return NULL;
 	}
+
+	RIOMap *gotr2map = b->iob.map_get (io, n_vaddr);
+	if (!gotr2map) {
+		return NULL;
+	}
+	gotr2map->name = strdup (".got.r2");
```
radare: fishy

ret2libc: in which way? this was already like this. It is to create the "fake" got section used to patch relocations.

PR #12952, opened 2019-01-30
Add reloc indicator to json disassembly to explain wheter an address is relocated

ITAYC0HEN:
This PR is adding an indicator for each instruction, saying whether it is in the relocation table.
This will be shown to the user in `pdj`, `pij` etc

image


radare:
pls update tests and merge

radare:
What about fixing the actual disasm

PR #12953, opened 2019-01-30
WIP: Bincurat

radare:
Random related comments:

* Should be r_bin_file_current()... but.. we really want to have that? i mean, a default binfile? and we should use descriptors, not pointers
* same happens with R_IO which assumes there's a current "filedescriptor".. this is in siol-speak used only when working in non-va mode

This function spots a bunch of bad design decisions in RBin and its not used anywhere yet, lets keep it as is for now.

### DIFF
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -937,3 +937,23 @@ R_API bool r_bin_file_hash(RBin *bin, ut64 limit, const char *file, RList/*<RBin
 	r_hash_free (ctx);
 	return true;
 }
+
+R_API RBinFile *r_bin_file_at(RBin *bin, ut64 addr) {
+	RBinFile *bf;
+	RListIter *iter;
+	RBinFile *obf = bin->cur; // XXX this must die
+	r_list_foreach (bin->binfiles, iter, bf) {
+		bin->cur = bf;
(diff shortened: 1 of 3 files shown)


radare:
this function spots a bunch of bad design decisions in RBin and its not used anywhere yet, lets keep it as is for now

libr/bin/bfile.c
```diff
@@ -904,3 +906,24 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
+	RListIter *iter;
+	RBinFile *obf = bin->cur; // XXX this must die
+	r_list_foreach (bin->binfiles, iter, bf) {
+		bin->cur = bf;
+		RList *sections = r_bin_get_sections (bin);
+		RBinSection *s;
+		r_list_foreach (sections, iter, s) {
+			ut64 from = s->vaddr;
+			ut64 to = from + s->vsize;
+			if (R_BETWEEN (from, addr, to)) {
```
ret2libc: the bad about R_BETWEEN is that it includes both the start and the end, so i think you should use `to - 1`

radare:
this is important to support multidex

radare:
rebased

radare:
Its now merged an already working implementation. `e bin.at=true`

PR #12962, opened 2019-01-31
Eliminated the assumption in r_asm_psuedo_hex

deepakchethan:
Fix for issue #12931 

After making sure that the input size is sanitized. The segmentation fault is eliminated. I have used to internal API's of radare2 as suggested. 

EDIT: I have changed the way I resolved this based on the suggestion of @radare


libr/asm/asm.c
```diff
@@ -65,9 +65,11 @@ static inline int r_asm_pseudo_org(RAsm *a, char *input) {
 
 // wtf isnt this the same as r_asm_op_set_hex() ??
 static inline int r_asm_pseudo_hex(RAsmOp *op, char *input) {
-	// assume input fits inside op->buf
-	int len = r_hex_str2bin (input, (ut8*)r_strbuf_get (&op->buf));
-	r_asm_op_set_hex (op, r_str_trim_head_tail (input));
+    	ut8 *opBuf = r_asm_op_get_buf (op);
```
radare: chk for null

libr/asm/asm.c
```diff
@@ -65,9 +65,11 @@ static inline int r_asm_pseudo_org(RAsm *a, char *input) {
...
+	// Truncate the input based on the size of the opBuf
+    	input = r_str_newlen (r_str_trim_head_tail (input), sizeof (RStrBuf));
```
radare: this is wrong. sizeof(RStrBuf) is not a restriction for anything to care of. also you are reusing a variable that is passed to the function instead of creaing one , so changing its meaning and therefor breaking the logic, and i dont see why you want to truncate that input

libr/asm/asm.c
deepakchethan: Okay, will do.

libr/asm/asm.c
deepakchethan: Okay now after I saw the r_asm_op_set_hex method. I realized the problem was r_hex_str2bin arguments were improperly used. I will revert and fix that.

libr/asm/asm.c
```diff
@@ -65,9 +65,11 @@ static inline int r_asm_pseudo_org(RAsm *a, char *input) {
...
+    	int len = r_hex_str2bin (input, opBuf);
+    	r_asm_op_set_hex (op, input);
```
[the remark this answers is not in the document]
deepakchethan: r_asm_op_set_bin() is currently not present in the api.
But like you said r_asm_op_set_hex is implicitly setting the bin.
Also r_hex_str2bin is being used to get the length so that we can throw an error and return it.
r_asm_op_set_hex() api does not return the length. I didn't want to modify it since it might affect the other usages of that method.

deepakchethan:
I have made the changes that were asked and I am not getting the segmentation error anymore.

radare:
this fix is not correct, the issue lays in several other places in r2, not just egg, and your fix is truncating the buffer even more, so it will be generating the wrong input. im about to take a 18h offline flight, so i will be working on the right fix for this because it affects asm, egg and r_util.

the problem is that massemble splits the input into lines which have a limit defined by R_ASM_BUFSIZE which is something that must be deprecated. that truncated buffer causes an odd (uneven) hexpair string which util returns a negative value to indicate that to the caller which wrongly assumes that and ends up having an non-null terminated string that makes it crash, but this crash is just the tip of a bunch of several issues introduced after the RStrBuf refactoring in RAsmOp that its wrongly used in that case.

So there's no simple fix for this without the proper refactoring and code cleanup of all those parts involved.

PR #12965, opened 2019-01-31
Fix assembler for Windows

GustavoLCR:
This fixes all failing assembler tests for Windows (I think).


GustavoLCR:
AppVeyor ignored me :(

PR #12966, opened 2019-02-01
Fix asan crash in PE parser (oobread)

radare:


radare:
[Key4.exe.zip](https://github.com/radare/radare2/files/2823177/Key4.exe.zip)

PR #12969, opened 2019-02-02
Fix thumb early

radare:


radare:
Cherry-picked all those commits into master

PR #12971, opened 2019-02-02
Add user * to REvent Hooking

thestr4ng3r:
Unit tests for REvent: https://github.com/radare/radare2-regressions/pull/1677

After these changes, the ht uses RVector<REventCallbackHook> as the key instead of RPVector. Also, hooked callbacks are referred by a unique handle.

### DIFF
--- a/libr/anal/anal.c
+++ b/libr/anal/anal.c
@@ -26,35 +26,35 @@ R_API void r_anal_unset_limits(RAnal *anal) {
 }
 
 static void meta_unset_for(REvent *ev, int type, void *user, void *data) {
-	RSpaces *s = (RSpaces *)user;
+	RSpaces *s = (RSpaces *)ev->user;
 	RAnal *anal = container_of (s, RAnal, meta_spaces);
 	RSpaceEvent *se = (RSpaceEvent *)data;
 	r_meta_space_unset_for (anal, se->data.unset.space);
 }
 
 static void meta_count_for(REvent *ev, int type, void *user, void *data) {
-	RSpaces *s = (RSpaces *)user;
+	RSpaces *s = (RSpaces *)ev->user;
 	RAnal *anal = container_of (s, RAnal, meta_spaces);
 	RSpaceEvent *se = (RSpaceEvent *)data;
 	se->res = r_meta_space_count_for (anal, se->data.count.space);
 }
 
 static void zign_unset_for(REvent *ev, int type, void *user, void *data) {
-	RSpaces *s = (RSpaces *)user;
+	RSpaces *s = (RSpaces *)ev->user;
 	RAnal *anal = container_of (s, RAnal, zign_spaces);
 	RSpaceEvent *se = (RSpaceEvent *)data;
 	r_sign_space_unset_for (anal, se->data.unset.space);
 }
 
 static void zign_count_for(REvent *ev, int type, void *user, void *data) {
-	RSpaces *s = (RSpaces *)user;
+	RSpaces *s = (RSpaces *)ev->user;
 	RAnal *anal = container_of (s, RAnal, zign_spaces);
 	RSpaceEvent *se = (RSpaceEvent *)data;
 	se->res = r_sign_space_count_for (anal, se->data.count.space);
 }
 
 static void zign_rename_for(REvent *ev, int type, void *user, void *data) {
-	RSpaces *s = (RSpaces *)user;
+	RSpaces *s = (RSpaces *)ev->user;
 	RAnal *anal = container_of (s, RAnal, zign_spaces);
 	RSpaceEvent *se = (RSpaceEvent *)data;
 	r_sign_space_rename_for (anal, se->data.rename.space,
@@ -167,13 +167,13 @@ R_API RAnal *r_anal_new() {
 	anal->opt.depth = 32;
 	anal->opt.noncode = false; // do not analyze data by default
 	r_spaces_init (&anal->meta_spaces, "CS");
-	r_event_hook (anal->meta_spaces.event, R_SPACE_EVENT_UNSET, meta_unset_for);
-	r_event_hook (anal->meta_spaces.event, R_SPACE_EVENT_COUNT, meta_count_for);
+	r_event_hook (anal->meta_spaces.event, R_SPACE_EVENT_UNSET, meta_unset_for, NULL);
+	r_event_hook (anal->meta_spaces.event, R_SPACE_EVENT_COUNT, meta_count_for, NULL);
 
 	r_spaces_init (&anal->zign_spaces, "zs");
-	r_event_hook (anal->zign_spaces.event, R_SPACE_EVENT_UNSET, zign_unset_for);
-	r_event_hook (anal->zign_spaces.event, R_SPACE_EVENT_COUNT, zign_count_for);
-	r_event_hook (anal->zign_spaces.event, R_SPACE_EVENT_RENAME, zign_rename_for);
+	r_event_hook (anal->zign_spaces.event, R_SPACE_EVENT_UNSET, zign_unset_for, NULL);
+	r_event_hook (anal->zign_spaces.event, R_SPACE_EVENT_COUNT, zign_count_for, NULL);
+	r_event_hook (anal->zign_spaces.event, R_SPACE_EVENT_RENAME, zign_rename_for, NULL);
 	anal->sdb_fcns = sdb_ns (anal->sdb, "fcns", 1);
 	anal->sdb_meta = sdb_ns (anal->sdb, "meta", 1);
 	anal->sdb_hints = sdb_ns (anal->sdb, "hints", 1);
--- a/libr/include/r_util/r_event.h
+++ b/libr/include/r_util/r_event.h
@@ -15,7 +15,7 @@ typedef struct r_event_t {
 	bool incall;
 	HtUP *callbacks;
 	RVector *all_callbacks;
-	int hook_handle_next;
+	int next_handle;
 } REvent;
 
 typedef struct r_event_callback_handle_t {
--- a/libr/include/r_util/r_event.h
+++ b/libr/include/r_util/r_event.h
@@ -13,7 +13,16 @@ typedef struct r_event_t {
 	void *user;
 	bool incall;
 	HtUP *callbacks;
+	int hook_handle_next;
 } REvent;
+
+typedef void (*REventCallback)(REvent *ev, int type, void *user, void *data);
+
+typedef struct r_event_callback_hook_t {
+	REventCallback cb;
+	void *user;
+	int handle;
+} REventCallbackHook;
 
 typedef enum {
 	R_EVENT_ALL = 0,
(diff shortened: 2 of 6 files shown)


radare:
why travis red? can you rebase? i merged some stuff

thestr4ng3r:
pr travis is green, it's only push again because of missing r2r changes

thestr4ng3r:
rebased

libr/include/r_util/r_event.h
```diff
@@ -13,8 +13,17 @@ typedef struct r_event_t {
 	void *user;
 	bool incall;
 	HtUP *callbacks;
+	int hook_handle_next;
```
radare: whats the purpose of this variable?

libr/util/event.c
```diff
@@ -34,50 +37,60 @@ R_API void r_event_free(REvent *ev) {
...
 }
 
-static bool del_hook(void *cb, const ut64 k, const void *v) {
-	RPVector *cbs = (RPVector *)v;
+static bool del_hook(void *user, const ut64 k, const void *v) {
+	int handle = (int)(intptr_t)user;
+	RVector *cbs = (RVector *)v;
 	r_return_val_if_fail (cbs, false);
-	r_pvector_remove_data (cbs, cb);
+	for (size_t i=0; i<cbs->len; i++) {
```
radare: spaces around =, aand <, also, dont define variables inside the size_t

libr/util/event.c
```diff
@@ -34,50 +37,60 @@ R_API void r_event_free(REvent *ev) {
...
+		RPVector *cbs = ht_up_find (ev->callbacks, (ut64)type, NULL);
+		add_hook (&hook, 0, cbs);
 	}
+	return hook.handle;
 }
 
-static bool del_hook(void *cb, const ut64 k, const void *v) {
-	RPVector *cbs = (RPVector *)v;
+static bool del_hook(void *user, const ut64 k, const void *v) {
+	int handle = (int)(intptr_t)user;
```
radare: this intptr-t thing is scary

libr/util/event.c
```diff
@@ -34,50 +37,60 @@ R_API void r_event_free(REvent *ev) {
...
+	}
 	return true;
 }
 
-R_API void r_event_unhook(REvent *ev, int type, REventCallback cb) {
+R_API void r_event_unhook(REvent *ev, int type, int handle) {
 	r_return_if_fail (ev);
 	if (type == R_EVENT_ALL) {
-		ht_up_foreach (ev->callbacks, del_hook, cb);
+		ht_up_foreach (ev->callbacks, del_hook, (void *)(intptr_t )handle);
```
radare: uhmf casting int to void* is ... fine on 32bits, but its a bad practice

libr/anal/anal.c
```diff
@@ -167,13 +167,13 @@ R_API RAnal *r_anal_new() {
 	anal->opt.depth = 32;
 	anal->opt.noncode = false; // do not analyze data by default
 	r_spaces_init (&anal->meta_spaces, "CS");
-	r_event_hook (anal->meta_spaces.event, R_SPACE_EVENT_UNSET, meta_unset_for);
-	r_event_hook (anal->meta_spaces.event, R_SPACE_EVENT_COUNT, meta_count_for);
+	r_event_hook (anal->meta_spaces.event, R_SPACE_EVENT_UNSET, meta_unset_for, NULL);
```
radare: not super fan of adding more parameters

libr/include/r_util/r_event.h
ret2libc: unsure about this too...

thestr4ng3r: It's counted up to give every registered callback a unique handle, so you can unhook it again. Unhooking only by the function pointer will not work anymore because the same function may be hooked with a different user pointer.

ret2libc:
Btw I just realized the way I'm using REvent in RSpaces is potentially wrong as of right now...

libr/anal/anal.c
ret2libc: you need to pass the right user here though

libr/include/r_util/r_event.h
```diff
@@ -13,8 +13,17 @@ typedef struct r_event_t {
...
 } REvent;
 
+typedef void (*REventCallback)(REvent *ev, int type, void *user, void *data);
+
+typedef struct r_event_callback_hook_t {
```
ret2libc: move this in event.c, this does not need to be public-facing API

libr/include/r_util/r_event.h
ret2libc: AFAIK RIDStorage should be the right thing for these kind of operations, but... I've never used it.

libr/include/r_util/r_event.h
```diff
@@ -29,12 +38,11 @@ typedef struct r_event_meta_t {
...
 } REventMeta;
 
-typedef void (*REventCallback)(REvent *ev, int type, void *user, void *data);
 
 R_API REvent *r_event_new(void *user);
 R_API void r_event_free(REvent *ev);
-R_API void r_event_hook(REvent *ev, int type, REventCallback cb);
-R_API void r_event_unhook(REvent *ev, int type, REventCallback cb);
+R_API int r_event_hook(REvent *ev, int type, REventCallback cb, void *user);
+R_API void r_event_unhook(REvent *ev, int type, int handle);
```
ret2libc: is the type really needed at this point? I think it's enough to use the handle

libr/util/event.c
ret2libc: please just use `&handle` and get a `int *handle = (int *)user` in the `del_hook` function.

libr/include/r_util/r_event.h
ret2libc: Or it's RIDPool :D i'm not sure

libr/include/r_util/r_event.h
thestr4ng3r: Yeah right I just noticed that too while trying to remove it. I am just not sure whether the caller of the function should be responsible for specifying the handle.

libr/anal/anal.c
ret2libc: Actually, you don't since you've changed callbacks to use `ev->user`. Sorry for the noise.

libr/util/event.c
thestr4ng3r: `r_vector_foreach()` doesn't work here with the current vector api since I need to delete an element and we only have `r_vector_remove_at()` which needs the index. In `r_event_send()` I did use the foreach though.

libr/include/r_util/r_event.h
thestr4ng3r: RIDPool would work, but looks like overkill to me.

libr/util/event.c
```diff
@@ -34,50 +37,60 @@ R_API void r_event_free(REvent *ev) {
...
+static bool del_hook(void *user, const ut64 k, const void *v) {
+	int handle = (int)(intptr_t)user;
+	RVector *cbs = (RVector *)v;
 	r_return_val_if_fail (cbs, false);
-	r_pvector_remove_data (cbs, cb);
+	for (size_t i=0; i<cbs->len; i++) {
+		REventCallbackHook *hook = r_vector_index_ptr (cbs, i);
+		if (hook->handle == handle) {
+			r_vector_remove_at (cbs, i, NULL);
+			break;
```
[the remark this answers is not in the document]
thestr4ng3r: `del_hook()` is only called in `ht_up_foreach()` when unregistering from `R_EVENT_ALL`. In that case, all elements do have to be iterated.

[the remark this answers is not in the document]
ret2libc: ok right the same `handle` can be in different vectors.

libr/include/r_util/r_event.h
ret2libc: however, i feel like `next_handle` or something like that would be better.

libr/util/event.c
```diff
@@ -15,9 +21,12 @@ R_API REvent *r_event_new(void *user) {
 	// and if used in hook/unhook/send APIs will raise a warning
 	ut64 i;
 	for (i = 1; i < R_EVENT_MAX; ++i) {
-		ht_up_insert (ev->callbacks, (ut64)i, r_pvector_new (NULL));
+		ht_up_insert (ev->callbacks, (ut64)i, r_vector_new (sizeof (REventCallbackHook), NULL, NULL));
 	}
 	ev->user = user;
+
```
ret2libc: remove these spaces around the assignment, I don't see much value in this case.

libr/util/event.c
```diff
@@ -34,50 +43,64 @@ R_API void r_event_free(REvent *ev) {
...
 
-static bool del_hook(void *cb, const ut64 k, const void *v) {
-	RPVector *cbs = (RPVector *)v;
+static bool del_hook(void *user, const ut64 k, const void *v) {
+	int handle = *((int *)user);
+	RVector *cbs = (RVector *)v;
 	r_return_val_if_fail (cbs, false);
-	r_pvector_remove_data (cbs, cb);
+	size_t i;
+	for (i = 0; i < cbs->len; i++) {
```
ret2libc: what about a `r_vector_enumerate(vector, it, i)`? it would be a bit more declarative.

libr/util/event.c
```diff
@@ -34,50 +43,64 @@ R_API void r_event_free(REvent *ev) {
...
 	}
+	handle.handle = hook.handle;
+	handle.type = type;
+	return handle;
 }
 
-static bool del_hook(void *cb, const ut64 k, const void *v) {
-	RPVector *cbs = (RPVector *)v;
+static bool del_hook(void *user, const ut64 k, const void *v) {
+	int handle = *((int *)user);
```
ret2libc: no need for the external parenthesis.

libr/util/event.c
```diff
@@ -34,50 +43,64 @@ R_API void r_event_free(REvent *ev) {
...
+	hook.cb = cb;
+	hook.user = user;
+	hook.handle = ev->hook_handle_next++;
 	if (type == R_EVENT_ALL) {
-		ht_up_foreach (ev->callbacks, add_hook, cb);
+		ht_up_foreach (ev->callbacks, add_hook, &hook);
 	} else {
-		RPVector *cbs = ht_up_find (ev->callbacks, type, NULL);
-		add_hook (cb, 0, cbs);
+		RPVector *cbs = ht_up_find (ev->callbacks, (ut64)type, NULL);
```
ret2libc: `s/RPVector/RVector` ?

libr/util/event.c
```diff
@@ -34,50 +43,64 @@ R_API void r_event_free(REvent *ev) {
...
+R_API void r_event_unhook(REvent *ev, REventCallbackHandle handle) {
 	r_return_if_fail (ev);
-	if (type == R_EVENT_ALL) {
-		ht_up_foreach (ev->callbacks, del_hook, cb);
+	if (handle.type == R_EVENT_ALL) {
+		ht_up_foreach (ev->callbacks, del_hook, &handle.handle);
 	} else {
-		RPVector *cbs = ht_up_find (ev->callbacks, type, NULL);
-		del_hook (cb, 0, cbs);
+		RPVector *cbs = ht_up_find (ev->callbacks, (ut64)handle.type, NULL);
```
ret2libc: RVector ?

ret2libc:
This is ok for me @thestr4ng3r @radare

radare:
Not for travis

PR #12982, opened 2019-02-03
Fix Windows path with / in the last position

GustavoLCR:
Before, if you set for example:
`e dir.projects = myprojects/`
It would result in an invalid path when creating a file:
`C:\path\to\myprojects/\file`

### DIFF
--- a/libr/util/file.c
+++ b/libr/util/file.c
@@ -211,12 +211,9 @@ R_API char *r_file_abspath(const char *file) {
 			return strdup (file);
 		}
 		if (cwd && !strchr (file, ':')) {
-			int len = strlen (file) - 1;
-			char * nfile = file[len] == '/' ? r_str_ndup (file, len) : file;
+			char * nfile = r_str_ndup (file, strlen (file) - r_str_endswith (file, "/"));
 			ret = r_str_newf ("%s\\%s", cwd, nfile);
-			if (nfile != file) {
-				free (nfile);
-			}
+			free (nfile);
 		}
 #endif
 	}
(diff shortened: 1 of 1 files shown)


libr/util/file.c
```diff
@@ -211,7 +211,14 @@ R_API char *r_file_abspath(const char *file) {
 			return strdup (file);
 		}
 		if (cwd && !strchr (file, ':')) {
-			ret = r_str_newf ("%s\\%s", cwd, file);
+			int len = strlen (file) - 1;
+			char * nfile;
```
radare: Use ternary operator and make those 7 lines into one

libr/util/file.c
```diff
@@ -211,7 +211,14 @@ R_API char *r_file_abspath(const char *file) {
...
+			if (file[len] == '/') {
+				nfile = r_str_ndup (file, len);
+			} else {
+				nfile = file;
+			}
+			ret = r_str_newf ("%s\\%s", cwd, nfile);
```
radare: Nfile is leaking here

libr/util/file.c
```diff
@@ -211,7 +211,12 @@ R_API char *r_file_abspath(const char *file) {
 			return strdup (file);
 		}
 		if (cwd && !strchr (file, ':')) {
-			ret = r_str_newf ("%s\\%s", cwd, file);
+			int len = strlen (file) - 1;
```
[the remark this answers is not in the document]
GustavoLCR: smart 👍

radare:
👍

PR #12988, opened 2019-02-04
Fixed bin_mz plugin

Freakqency:
The bin_mz plugin overshadows the features and functionality of any ne, le or lx file format plugin. While developing a  ne file format plugin it was identified that the ne plugin was unable to interact with the binary due to the mz plugin, this was also observed when testing a le file format plugin. This pull request addresses this issue.


radare:
red

libr/bin/p/bin_mz.c
```diff
@@ -58,22 +58,20 @@ static bool check_bytes(const ut8 *buf, ut64 length) {
 		}
 
 		// Check for New Executable, LE/LX or Phar Lap executable
-		if (!memcmp (buf + new_exe_header_offset, "NE", 2) ||
-			!memcmp (buf + new_exe_header_offset, "LE", 2) ||
+		if (!memcmp(buf, "MZ", 2) &&
```
XVilka: But why do you check here for MZ again? It was checked before in this function already. This looks wrong to me.

XVilka: And where is LE check?

libr/bin/p/bin_mz.c
```diff
@@ -58,22 +58,20 @@ static bool check_bytes(const ut8 *buf, ut64 length) {
...
 		}
 	}
-
-	// Raw plain MZ executable (watcom)
+	
 	if (!checkEntrypoint (buf, length)) {
 		return false;
 	}
-
+
```
XVilka: You added unneeded tab here - please check the code for right indentation and trailing whitespaces before committing.

libr/bin/p/bin_mz.c
Freakqency: LE check was left out because of one of the binary in the r2-regressions repo cdogs is an LE but it is in the mz bins folder. due to which the build was failing.as you said the mz check was unnecessary and it was bad code on my part.

libr/bin/p/bin_mz.c
Freakqency: sorry, my bad. I will make sure to check the code for tabs and unnecessary spaces before committing next time.

libr/bin/p/bin_mz.c
Maijin: Are you sure it's an LE file? If yes then that means the test is wrong

Freakqency: In my analysis using file command, it showed up as LE file, for further analysis I made the bin_mz plugin return false at the end (line no: 73) and removed all the conditions in the if statement having only the `!memcmp (buf + new_exe_header_offset, "LE", 2)` condition and returned true for the statement.when I tested these conditions on the cdogs plugin it still showed up as mz. I also tried using a plugin for le which uses `r_read_ble16 () ` function to find the offset .this also returned as le

Maijin: Hey you actually right it is an LE file. So the test is wrong.

Maijin: Can you add back a check for LE and send a Pull request on radare2-regressions side adding BROKEN=1 on the faulty tests?

Freakqency: yeah I will clean up this code and create a pull request in r2 regressions repo

libr/bin/p/bin_mz.c
```diff
@@ -58,22 +58,20 @@ static bool check_bytes(const ut8 *buf, ut64 length) {
 		}
 
 		// Check for New Executable, LE/LX or Phar Lap executable
-		if (!memcmp (buf + new_exe_header_offset, "NE", 2) ||
-			!memcmp (buf + new_exe_header_offset, "LE", 2) ||
+		if (!memcmp(buf, "MZ", 2) && 
+			!memcmp (buf + new_exe_header_offset, "NE", 2) ||
 			!memcmp (buf + new_exe_header_offset, "LX", 2) ||
 			!memcmp (buf + new_exe_header_offset, "PL", 2)) {
-			if (!checkEntrypoint (buf, length)) {
```
radare: as long as this conditional is pretty long and repetitive is better to move this into a separate function and make this code more readable like this: if (knownHeader(buf+new_exe_header_offset)) {

radare:
awesome thanks!

libr/bin/p/bin_mz.c
Freakqency: okay sure

libr/bin/p/bin_mz.c
```diff
@@ -50,22 +67,9 @@ static bool check_bytes(const ut8 *buf, ut64 length) {
 	// See if there is a new exe header.
 	new_exe_header_offset = r_read_ble16 (buf + 0x3c, false);
 	if (length > new_exe_header_offset + 2) {
-		// check for PE
-		if (!memcmp (buf + new_exe_header_offset, "PE", 2) &&
-			(length > new_exe_header_offset + 0x20) &&
-			!memcmp (buf + new_exe_header_offset + 0x18, "\x0b\x01", 2)) {
+		if(!knownHeader(buf,new_exe_header_offset,length)){
```
radare: add space before (

libr/bin/p/bin_mz.c
```diff
@@ -15,6 +15,23 @@ static Sdb *get_sdb(RBinFile *bf) {
 	return NULL;
 }
 
+static bool knownHeader(const ut8 *buf,ut16 offset,ut64 length){
+	// check for PE
+		if (!memcmp (buf + offset, "PE", 2) &&
```
radare: wrong indentation

PR #12989, opened 2019-02-04
Fixing file hash computation.

devnexen:
Reading by small chunks instead of "slurping" all.

### DIFF
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -849,7 +849,7 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
 	return false;
 }
 
-R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
+R_API bool r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
 	char hash[128], *p;
 	RHash *ctx;
 	ut64 buf_len = 0, r = 0;
@@ -874,23 +874,30 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
 	buf_len = r_io_desc_size (iod);
 	// By SLURP_LIMIT normally cannot compute ...
 	if (buf_len > limit) {
-		eprintf ("Cannot compute hash\n");
+		eprintf ("Cannot compute hash: %zu/%zu\n", buf_len, limit);
 		return false;
 	}
-#define BLK_SIZE_OFF 4096
+#define BLK_SIZE_OFF 64000
 	ut8 *buf = malloc (BLK_SIZE_OFF);
 	if (!buf) {
 		eprintf ("Cannot allocate computation buffer\n");
 		return false;
 	}
 	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1);
-	while (r < buf_len) {
+	while (r+BLK_SIZE_OFF < buf_len) {
 		r_io_desc_seek (iod, r, R_IO_SEEK_SET);
-		size_t rem_len = R_MIN(buf_len-r, BLK_SIZE_OFF);
+		size_t rem_len = R_MIN (buf_len-r, BLK_SIZE_OFF);
 		int b = r_io_desc_read (iod, buf, rem_len);
 		(void)r_hash_do_md5 (ctx, buf, rem_len);
 		(void)r_hash_do_sha1 (ctx, buf, rem_len);
 		r += b;
+	}
+	if (r < buf_len) {
+		r_io_desc_seek (iod, r, R_IO_SEEK_SET);
+		size_t rem_len = R_MIN (buf_len-r, BLK_SIZE_OFF);
+		int b = r_io_desc_read (iod, buf, rem_len);
+		(void)r_hash_do_md5 (ctx, buf, rem_len);
+		(void)r_hash_do_sha1 (ctx, buf, rem_len);
 	}
 	r_hash_do_end (ctx, R_HASH_MD5);
 	p = hash;
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -852,6 +852,7 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
 R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
 	char hash[128], *p;
 	RHash *ctx;
+	RStrBuf *sbuf;
 	ut64 buf_len = 0, r = 0;
 	int i;
 	RBinFile *bf = bin->cur;
@@ -880,22 +881,24 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
 	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1);
 	while (r < buf_len) {
 #define BLK_SIZE_OFF 4096
-		ut8 buf[BLK_SIZE_OFF];
-		int b = r_io_desc_read (iod, buf, sizeof (buf));
-		(void)r_hash_do_md5 (ctx, buf, R_MIN (b, sizeof (buf)));
-		(void)r_hash_do_sha1 (ctx, buf, R_MIN (b, sizeof (buf)));
+		ut8 *buf = malloc (BLK_SIZE_OFF);
+		int b = r_io_desc_read (iod, buf, R_MIN (buf_len-r, BLK_SIZE_OFF));
+		(void)r_hash_do_md5 (ctx, buf, R_MIN (b, BLK_SIZE_OFF));
+		(void)r_hash_do_sha1 (ctx, buf, R_MIN (b, BLK_SIZE_OFF));
 		r += b;
 		r_io_desc_seek (iod, r, R_IO_SEEK_SET);
+		free (buf);
 	}
 	r_hash_do_end (ctx, R_HASH_MD5);
 	p = hash;
 	r_hex_bin2str (ctx->digest, R_HASH_SIZE_MD5, p);
-	o->info->hashes = r_strbuf_new ("");
-	r_strbuf_appendf (o->info->hashes, "md5 %s", hash);
+	sbuf = r_strbuf_new ("");
+	r_strbuf_appendf (sbuf, "md5 %s", hash);
 	r_hash_do_end (ctx, R_HASH_SHA1);
 	p = hash;
 	r_hex_bin2str (ctx->digest, R_HASH_SIZE_SHA1, p);
-	r_strbuf_appendf (o->info->hashes, "\nsha1 %s", hash);
+	r_strbuf_appendf (sbuf, "\nsha1 %s", hash);
+	o->info->hashes = r_strbuf_drain (sbuf);
 	r_hash_free (ctx);
 	return true;
 }
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -852,8 +852,7 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
 R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
 	char hash[128], *p;
 	RHash *ctx;
-	ut8* buf;
-	ut64 buf_len = 0;
+	ut64 buf_len = 0, r = 0;
 	int i;
 	RBinFile *bf = bin->cur;
 	if (!bf) {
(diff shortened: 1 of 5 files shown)


libr/bin/bfile.c
```diff
@@ -878,34 +877,25 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
-		r_hex_bin2str (ctx->digest, R_HASH_SIZE_SHA1, p);
-		r_strbuf_appendf (o->info->hashes, "\nsha1 %s", hash);
-		r_hash_free (ctx);
-	}
-	free (buf);
+	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1);
+	while (r < buf_len) {
+#define BLK_SIZE_OFF 4096
+		ut8 buf[BLK_SIZE_OFF];
+		int b = r_io_desc_read (iod, buf, sizeof (buf));
```
radare: what about the last chunk? not all files are 4K aligned in size

libr/bin/bfile.c
```diff
@@ -878,34 +877,25 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
-		p = hash;
-		r_hex_bin2str (ctx->digest, R_HASH_SIZE_SHA1, p);
-		r_strbuf_appendf (o->info->hashes, "\nsha1 %s", hash);
-		r_hash_free (ctx);
-	}
-	free (buf);
+	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1);
+	while (r < buf_len) {
+#define BLK_SIZE_OFF 4096
+		ut8 buf[BLK_SIZE_OFF];
```
radare: better use the heap for this, some systems have very small stack

libr/bin/bfile.c
```diff
@@ -878,34 +877,25 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
+		int b = r_io_desc_read (iod, buf, sizeof (buf));
+		(void)r_hash_do_md5 (ctx, buf, R_MIN (b, sizeof (buf)));
+		(void)r_hash_do_sha1 (ctx, buf, R_MIN (b, sizeof (buf)));
+		r += b;
+		r_io_desc_seek (iod, r, R_IO_SEEK_SET);
+	}
+	r_hash_do_end (ctx, R_HASH_MD5);
+	p = hash;
+	r_hex_bin2str (ctx->digest, R_HASH_SIZE_MD5, p);
+	o->info->hashes = r_strbuf_new ("");
```
radare: i think hashes can be just char *, so we construct the strbuf() in here and drain it in there when complete

libr/bin/bfile.c
```diff
@@ -878,34 +878,27 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
-		r_strbuf_appendf (o->info->hashes, "md5 %s", hash);
-		r_hash_do_end (ctx, R_HASH_SHA1);
-		p = hash;
-		r_hex_bin2str (ctx->digest, R_HASH_SIZE_SHA1, p);
-		r_strbuf_appendf (o->info->hashes, "\nsha1 %s", hash);
-		r_hash_free (ctx);
+	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1);
+	while (r < buf_len) {
+#define BLK_SIZE_OFF 4096
+		ut8 *buf = malloc (BLK_SIZE_OFF);
```
radare: put the malloc/free outside the while loop

libr/bin/bfile.c
```diff
@@ -878,34 +878,27 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
-		r_hex_bin2str (ctx->digest, R_HASH_SIZE_MD5, p);
-		o->info->hashes = r_strbuf_new ("");
-		r_strbuf_appendf (o->info->hashes, "md5 %s", hash);
-		r_hash_do_end (ctx, R_HASH_SHA1);
-		p = hash;
-		r_hex_bin2str (ctx->digest, R_HASH_SIZE_SHA1, p);
-		r_strbuf_appendf (o->info->hashes, "\nsha1 %s", hash);
-		r_hash_free (ctx);
+	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1);
+	while (r < buf_len) {
```
radare: if you do while (r + BLK_SIZE_OFF < buf_len) you can save all those R_MINs below and add the last remaining part if needed with if (r<buf_len) { } after this loop. so you dont have to compute the siez of the buffer becaus its known in each loop

libr/bin/bfile.c
radare: also chk for null

libr/bin/bfile.c
```diff
@@ -878,34 +878,27 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
+#define BLK_SIZE_OFF 4096
+		ut8 *buf = malloc (BLK_SIZE_OFF);
+		int b = r_io_desc_read (iod, buf, R_MIN (buf_len-r, BLK_SIZE_OFF));
+		(void)r_hash_do_md5 (ctx, buf, R_MIN (b, BLK_SIZE_OFF));
+		(void)r_hash_do_sha1 (ctx, buf, R_MIN (b, BLK_SIZE_OFF));
+		r += b;
+		r_io_desc_seek (iod, r, R_IO_SEEK_SET);
```
radare: why seek after read? imho it makes more sense to move this before the read operation

libr/bin/bfile.c
```diff
@@ -878,34 +878,27 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
+		free (buf);
 	}
-	free (buf);
+	r_hash_do_end (ctx, R_HASH_MD5);
+	p = hash;
+	r_hex_bin2str (ctx->digest, R_HASH_SIZE_MD5, p);
+	sbuf = r_strbuf_new ("");
```
radare: define and assign in the same line

libr/bin/bfile.c
```diff
@@ -852,8 +852,8 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
 R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
 	char hash[128], *p;
 	RHash *ctx;
-	ut8* buf;
-	ut64 buf_len = 0;
+	RStrBuf *sbuf;
```
radare: dont hold those vars here if you can initialize them in place below

radare:
but lets do that in a different pr

libr/core/cmd_info.c
```diff
@@ -458,7 +458,7 @@ static int cmd_info(void *data, const char *input) {
 		}
 		case 'F': // "iF"
 			if (input[1] == 'j') {
-				r_cons_printf ("{\"%s\"}", r_strbuf_get (core->bin->cur->o->info->hashes));
+				r_cons_printf ("{\"%s\"}", core->bin->cur->o->info->hashes);
```
radare: can be NULL, also this json looks wrong, use pj api

libr/bin/bfile.c
```diff
@@ -876,36 +875,34 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
-		return -1;
-	}
-	//  XXX should use io api not raw file slurping
-	int blen = 0;
-	buf = (ut8*)r_file_slurp (file, &blen);
-	if (!buf) {
 		return false;
 	}
-	buf_len = blen;
-	if (!o->info) {
```
Maijin: This is if is necessary, otherwise it crashes when a rbin plugin doesn't have info implemented

libr/bin/bfile.c
```diff
@@ -852,20 +852,19 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
 R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
```
radare: the function signature returns int, shouldnt this be bool instead?

libr/bin/bfile.c
radare: can you split this loop in 2 to avoid computing the minblksz on every iteration instead of just none?

libr/bin/bfile.c
```diff
@@ -876,36 +875,34 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
-	}
-	//  XXX should use io api not raw file slurping
-	int blen = 0;
-	buf = (ut8*)r_file_slurp (file, &blen);
-	if (!buf) {
 		return false;
 	}
-	buf_len = blen;
-	if (!o->info) {
+#define BLK_SIZE_OFF 4096
```
radare: i think 64K performs better in modern hw

libr/bin/bfile.c
```diff
@@ -876,36 +875,34 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
-		r_hash_do_end (ctx, R_HASH_SHA1);
-		p = hash;
-		r_hex_bin2str (ctx->digest, R_HASH_SIZE_SHA1, p);
-		r_strbuf_appendf (o->info->hashes, "\nsha1 %s", hash);
-		r_hash_free (ctx);
-	}
+	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1);
+	while (r < buf_len) {
+		r_io_desc_seek (iod, r, R_IO_SEEK_SET);
+		size_t rem_len = R_MIN(buf_len-r, BLK_SIZE_OFF);
```
radare: missing space

libr/bin/bfile.c
devnexen: yes good point.

libr/bin/bfile.c
devnexen: Sorry I forgot :-)

libr/bin/bfile.c
devnexen: Ok but what if it is not ?

devnexen:
Might be worthy to apply the request above then ... ie occuring only when requested.

libr/bin/bfile.c
```diff
@@ -875,37 +874,42 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
-		r_hash_do_end (ctx, R_HASH_SHA1);
-		p = hash;
-		r_hex_bin2str (ctx->digest, R_HASH_SIZE_SHA1, p);
-		r_strbuf_appendf (o->info->hashes, "\nsha1 %s", hash);
-		r_hash_free (ctx);
-	}
+	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1);
+	while (r+BLK_SIZE_OFF < buf_len) {
+		r_io_desc_seek (iod, r, R_IO_SEEK_SET);
+		size_t rem_len = R_MIN (buf_len-r, BLK_SIZE_OFF);
```
radare: there's no remaining necessary at all, inside this loop the blocks are always of BLK_SIZE_OFF (even if thats a bad name)

libr/bin/bfile.c
```diff
@@ -875,37 +874,42 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
+		int b = r_io_desc_read (iod, buf, rem_len);
+		(void)r_hash_do_md5 (ctx, buf, rem_len);
+		(void)r_hash_do_sha1 (ctx, buf, rem_len);
+		r += b;
+	}
+	if (r < buf_len) {
+		r_io_desc_seek (iod, r, R_IO_SEEK_SET);
+		size_t rem_len = R_MIN (buf_len-r, BLK_SIZE_OFF);
```
radare: no need to use R_MIN, this loop only contains rem_lens smaller than BLK_SIZE_OFF, so just rem_len = buf_len -r;

radare: also this is const

libr/bin/bfile.c
radare: this is const

libr/bin/bfile.c
```diff
@@ -875,37 +874,42 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
-	//  XXX should use io api not raw file slurping
-	int blen = 0;
-	buf = (ut8*)r_file_slurp (file, &blen);
-	if (!buf) {
+		eprintf ("Cannot compute hash: %zu/%zu\n", buf_len, limit);
 		return false;
 	}
-	buf_len = blen;
-	if (!o->info) {
+#define BLK_SIZE_OFF 64000
```
radare: better use a `const size_t blocksize = 64000;` no need to use a define which its scope is global

libr/bin/bfile.c
```diff
@@ -875,37 +874,42 @@ R_API int r_bin_file_hash(RBin *bin, ut64 limit, const char *file) {
...
 	// By SLURP_LIMIT normally cannot compute ...
 	if (buf_len > limit) {
-		eprintf ("Cannot compute hash\n");
-		return -1;
-	}
-	//  XXX should use io api not raw file slurping
-	int blen = 0;
-	buf = (ut8*)r_file_slurp (file, &blen);
-	if (!buf) {
+		eprintf ("Cannot compute hash: %zu/%zu\n", buf_len, limit);
```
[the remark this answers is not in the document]
devnexen: Useless was just debugging something...

PR #12993, opened 2019-02-04
Fix rabin2 extract path for Windows

GustavoLCR:


binr/rabin2/rabin2.c
```diff
@@ -188,8 +188,11 @@ static bool extract_binobj(const RBinFile *bf, RBinXtrData *data, int idx) {
 	if (!path) {
 		return false;
 	}
-	// XXX: Wrong for w32 (/)
 	ptr = strrchr (path, DIRSEP);
+#if __WINDOWS__
+	char *p = strrchr (path, '/');
```
radare: strrchr is not portable, better use r_str_rchr

radare: also this, with the other patch from yesterday seems like workarounds to me, we should probably use or implement more generic filesystem apis

GustavoLCR: yeah, didn't notice there was already one api for that.

PR #12995, opened 2019-02-04
Better fix for abspath for Windows

GustavoLCR:
This uses the native Windows API to do the parsing.
Fixes the problems it had with `../example` paths and `folder/file`
+Fix for copy command

### DIFF
--- a/libr/util/file.c
+++ b/libr/util/file.c
@@ -211,11 +211,11 @@ R_API char *r_file_abspath(const char *file) {
 			return strdup (file);
 		}
 		if (!strchr (file, ':')) {
-			const ut32 max_size = 32768;
-			char *abspath = malloc (max_size);
+			char *abspath = malloc (MAX_PATH);
 			if (abspath) {
-				GetFullPathName (file, max_size, abspath, NULL);
-				ret = abspath;
+				GetFullPathName (file, MAX_PATH, abspath, NULL);
+				ret = strdup (abspath);
+				free (abspath);
 			}
 		}
 #endif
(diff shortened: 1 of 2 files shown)


libr/util/file.c
```diff
@@ -210,10 +210,13 @@ R_API char *r_file_abspath(const char *file) {
 		if (!strncmp (file, "\\\\", 2)) {
 			return strdup (file);
 		}
-		if (cwd && !strchr (file, ':')) {
-			char * nfile = r_str_ndup (file, strlen (file) - r_str_endswith (file, "/"));
-			ret = r_str_newf ("%s\\%s", cwd, nfile);
-			free (nfile);
+		if (!strchr (file, ':')) {
+			const ut32 max_size = 32768;
+			char *abspath = malloc (max_size);
```
radare: why 32K? isnt there a MAXPATH for windows?

libr/util/file.c
```diff
@@ -1049,7 +1052,12 @@ R_API bool r_file_copy (const char *src, const char *dst) {
 #if HAVE_COPYFILE_H
 	return copyfile (src, dst, 0, COPYFILE_DATA | COPYFILE_XATTR) != -1;
 #elif __WINDOWS__
-	return r_sys_cmdf ("copy %s %s", src, dst);
+	char *s = r_file_abspath (src);
+	char *d = r_file_abspath (dst);
```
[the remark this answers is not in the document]
GustavoLCR: from what I checked path with spaces is not working in linux as well... maybe this require changes in another place

libr/util/file.c
```diff
@@ -1049,7 +1052,16 @@ R_API bool r_file_copy (const char *src, const char *dst) {
 #if HAVE_COPYFILE_H
 	return copyfile (src, dst, 0, COPYFILE_DATA | COPYFILE_XATTR) != -1;
 #elif __WINDOWS__
-	return r_sys_cmdf ("copy %s %s", src, dst);
+	char *esrc = r_str_replace (strdup (src), "\"", "^\"", 1);
+	char *edst = r_str_replace (strdup (dst), "\"", "^\"", 1);
+	char *s = r_file_abspath (esrc);
+	char *d = r_file_abspath (edst);
+	bool ret = r_sys_cmdf ("copy \"%s\" \"%s\"", s, d);
```
GustavoLCR: Is there a better way to do this?

radare: do this works? because last time i tried quoting things in  windows wasnt workng at all, will be better to have an r_sys_cmdv() with variable argument handling to get argv as separate parameters instead of passing just a string and then using spawn instead of system to explictly separate each argument

radare: but this function doesnt exist, so the fix requires more work

GustavoLCR: only double quotes work in windows cmd, I tested and the only problem now is paths with spaces. But thats a problem in linux as well

GustavoLCR: I fixed the path with spaces problem using r_str_argv(). can you please check?

radare: thats not how to use r_str_argv, imho the whole thing is getting overengineered and the solution will be simpler if done like i said at the begining

radare:
way to go!

PR #13006, opened 2019-02-06
Support typedef'd (anonymous) structs ##types

kazarmy:
Partial fix for #12272.

### DIFF
--- a/libr/core/cmd_type.c
+++ b/libr/core/cmd_type.c
@@ -270,11 +270,11 @@ static void save_parsed_type(RCore *core, const char *parsed) {
 static RCore *core_ = NULL; // HACK
 
 static int stdifstruct(void *user, const char *k, const char *v) {
+	r_return_val_if_fail (core_ && core_->anal, false);
 	if (!strncmp (v, "struct", strlen ("struct") + 1)) {
 		return true;
 	}
 	if (!strcmp (v, "typedef")) {
-		r_return_val_if_fail (core_ && core_->anal, false);
 		Sdb *TDB = core_->anal->sdb_types;
 		char key[136];
 		snprintf (key, sizeof (key), "typedef.%s", k);
--- a/libr/core/cmd_type.c
+++ b/libr/core/cmd_type.c
@@ -271,7 +271,7 @@ static Sdb *TDB_ = NULL; // HACK
 
 static int stdifstruct(void *user, const char *k, const char *v) {
 	r_return_val_if_fail (TDB_, false);
-	if (!strncmp (v, "struct", strlen ("struct") + 1)) {
+	if (!strcmp (v, "struct")) {
 		return true;
 	}
 	if (!strcmp (v, "typedef")) {
--- a/libr/core/cmd_type.c
+++ b/libr/core/cmd_type.c
@@ -277,7 +277,7 @@ static int stdifstruct(void *user, const char *k, const char *v) {
 	if (!strcmp (v, "typedef")) {
 		const char *typedef_key = sdb_fmt ("typedef.%s", k);
 		const char *type = sdb_const_get (TDB_, typedef_key, NULL);
-		if (type && !strncmp (type, "struct ", 7)) {
+		if (type && r_str_startswith (type, "struct ")) {
 			return true;
 		}
 	}
--- a/libr/core/cmd_type.c
+++ b/libr/core/cmd_type.c
@@ -276,9 +276,8 @@ static int stdifstruct(void *user, const char *k, const char *v) {
 	}
 	if (!strcmp (v, "typedef")) {
 		Sdb *TDB = core_->anal->sdb_types;
-		char key[136];
-		snprintf (key, sizeof (key), "typedef.%s", k);
-		const char *type = sdb_const_get (TDB, key, NULL);
+		const char *typedef_key = sdb_fmt ("typedef.%s", k);
+		const char *type = sdb_const_get (TDB, typedef_key, NULL);
 		if (type && !strncmp (type, "struct ", 7)) {
 			return true;
 		}
--- a/libr/core/cmd_type.c
+++ b/libr/core/cmd_type.c
@@ -272,7 +272,8 @@ static RCore *core_ = NULL; // HACK
 static int stdifstruct(void *user, const char *k, const char *v) {
 	if (!strncmp (v, "struct", strlen ("struct") + 1)) {
 		return true;
-	} else if (!strcmp (v, "typedef")) {
+	}
+	if (!strcmp (v, "typedef")) {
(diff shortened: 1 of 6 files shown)


kazarmy:
Partial fix because typedefs and struct tags live in different namespaces. Shown by:

```c
struct A {
  float num;
};

typedef struct {
  double num;
} A;

int main(void) {
  printf("%zu\n", sizeof(struct A));
  printf("%zu\n", sizeof(A));
}
```
Output:
```
4
8
```

libr/core/cmd_type.c
```diff
@@ -267,8 +267,22 @@ static void save_parsed_type(RCore *core, const char *parsed) {
 	}
 }
 
+static RCore *core_ = NULL; // HACK
+
 static int stdifstruct(void *user, const char *k, const char *v) {
-	return !strncmp (v, "struct", strlen ("struct") + 1);
+	if (!strncmp (v, "struct", strlen ("struct") + 1)) {
+		return true;
+	} else if (!strcmp (v, "typedef")) {
```
radare: no need to else after return

libr/core/cmd_type.c
```diff
@@ -267,8 +267,22 @@ static void save_parsed_type(RCore *core, const char *parsed) {
...
+		r_return_val_if_fail (core_ && core_->anal, false);
```
radare: those statements should be at the begining of the function not here

libr/core/cmd_type.c
```diff
@@ -267,8 +267,22 @@ static void save_parsed_type(RCore *core, const char *parsed) {
 	}
 }
 
+static RCore *core_ = NULL; // HACK
```
radare: globals are evil

libr/core/cmd_type.c
```diff
@@ -267,8 +267,22 @@ static void save_parsed_type(RCore *core, const char *parsed) {
...
+
 static int stdifstruct(void *user, const char *k, const char *v) {
-	return !strncmp (v, "struct", strlen ("struct") + 1);
+	if (!strncmp (v, "struct", strlen ("struct") + 1)) {
+		return true;
+	} else if (!strcmp (v, "typedef")) {
+		r_return_val_if_fail (core_ && core_->anal, false);
+		Sdb *TDB = core_->anal->sdb_types;
+		char key[136];
```
radare: use sdb_fmt

libr/util/ctype.c
```diff
@@ -423,6 +423,13 @@ R_API char *r_type_format(Sdb *TDB, const char *t) {
 		fmt = r_str_append (fmt, vars);
 		free (vars);
 		return fmt;
+	} else if (!strcmp (kind, "typedef")) {
+		snprintf (var2, sizeof (var2), "typedef.%s", t);
+		const char *type = sdb_const_get (TDB, var2, NULL);
+		// only supports struct atm
+		if (type && !strncmp (type, "struct ", 7)) {
+			return r_type_format (TDB, type + 7);
```
radare: isnt type +7 the same as just t?

XVilka:
I think it is a good approach because proper fix would require refactoring TCC code a lot. Good idea!

libr/core/cmd_type.c
XVilka: I don't like this part too.

radare:
In fact you only need the TDB. Not core at all

libr/util/ctype.c
kazarmy: only for a specific case that should be rare but it causes infinite recursion so fixed in cd4f7164169f640f79b066eec0994379af4aad88

XVilka:
Yes, problem is that typedefs and structures are totally different entities and we don't resolve typedefs when converting types into SDB representations.

libr/core/cmd_type.c
kazarmy: I don't like it either but doing away with the global apparently requires significant work in sdb -- there should be a way to pass a context to sdb_foreach_list_filter().

kazarmy:
> In fact you only need the TDB. Not core at all

Done in 055c05b56851f649eeb50219e534cf1404fcbff5.

> ... we don't resolve typedefs when converting types into SDB representations.

I don't think we should.

libr/core/cmd_type.c
kazarmy: done 2a9a424fa6527c0b43a725c75f6b26bd0be3f848

libr/core/cmd_type.c
kazarmy: done 466a38fc1987c892091597bff34a65ab5e60d6a2

libr/core/cmd_type.c
kazarmy: done 4f5b655e2fcd13cd14409f8acb4773c67b64ec7b

libr/core/cmd_type.c
kazarmy: That should be in a different PR imho.

libr/core/cmd_type.c
```diff
@@ -267,8 +267,21 @@ static void save_parsed_type(RCore *core, const char *parsed) {
 	}
 }
 
+static Sdb *TDB_ = NULL; // HACK
+
 static int stdifstruct(void *user, const char *k, const char *v) {
-	return !strncmp (v, "struct", strlen ("struct") + 1);
+	r_return_val_if_fail (TDB_, false);
+	if (!strncmp (v, "struct", strlen ("struct") + 1)) {
```
radare: uhm, isnt this the same as !strcmp ? because you are comparing the trailing 0 too

libr/core/cmd_type.c
```diff
@@ -267,8 +267,21 @@ static void save_parsed_type(RCore *core, const char *parsed) {
...
+		return true;
+	}
+	if (!strcmp (v, "typedef")) {
+		const char *typedef_key = sdb_fmt ("typedef.%s", k);
+		const char *type = sdb_const_get (TDB_, typedef_key, NULL);
+		if (type && !strncmp (type, "struct ", 7)) {
```
radare: for readability we can use r_str_startswith()

libr/core/cmd_type.c
kazarmy: yes ... fixed old code c8c0efe31672fbed619a054654d368f12da668eb

libr/core/cmd_type.c
kazarmy: done 6c632e94cc0bc88a603a036214ed81aa2b431deb

kazarmy:
Just be aware that usage of a counter for anon structs can cause problems, e.g. #13014, and with preexisting SDB entries.

Issue #12620, opened 2019-01-03
Rework `oLj` to actually print usable prefixes, not just IO backend names


xarkes:
For example, oLj would print `debug` while the `debug://` prefix is not usable, one has to use `dbg://`.
Also the `zip` io plugin allows multiple prefixes such as `zip://`, `apk://`, `zipall://` etc.

PR #13013, opened 2019-02-07
Added usable prefix field for oLj command

deepakchethan:
For issue #12620

Example output for `oLj` now is:

`{"IO_Plugins":[{"Permissions":"rw_","Name":"ar","Description":"Open ar/lib files [ar|lib]://[file//path]","License":"LGPL3","prefix":"ar://, lib://"},{"Permissions":"rw_","Name":"bfdbg","Description":"BrainFuck Debugger (bfdbg://path/to/file)","License":"LGPL3","prefix":"bfdbg://"}....}`

Please let me know if any more changes are to be done

### DIFF
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -158,6 +158,7 @@ typedef struct r_io_plugin_t {
 	char *author;
 	char *license;
 	void *widget;
+	const char *uris;
 	int (*listener)(RIODesc *io);
 	int (*init)(void);
 	RIOUndo undo;
--- a/libr/io/p/io_ar.c
+++ b/libr/io/p/io_ar.c
@@ -92,8 +92,9 @@ static int r_io_ar_close(RIODesc *fd) {
 
 RIOPlugin r_io_plugin_ar = {
 	.name = "ar",
-	.desc = "Open ar/lib files [ar|lib]://[file//path]",
+	.desc = "Open ar/lib files",
 	.license = "LGPL3",
+	.uris = "ar://,lib://",
 	.open = r_io_ar_open,
 	.open_many = r_io_ar_open_many,
 	.write = r_io_ar_write,
--- a/libr/io/p/io_http.c
+++ b/libr/io/p/io_http.c
@@ -104,12 +104,13 @@ static RIODesc *__open(RIO *io, const char *pathname, int rw, int mode) {
 
 RIOPlugin r_io_plugin_http = {
 	.name = "http",
-        .desc = "http get (http://rada.re/)",
+	.desc = "Make http get requests",
+	.uris = "http://",
 	.license = "LGPL3",
-        .open = __open,
-        .close = __close,
+	.open = __open,
+	.close = __close,
 	.read = __read,
-        .check = __plugin_open,
+	.check = __plugin_open,
 	.lseek = __lseek,
 	.write = __write,
 };
--- a/libr/io/p/io_zip.c
+++ b/libr/io/p/io_zip.c
@@ -650,7 +650,8 @@ static int r_io_zip_close(RIODesc *fd) {
 
 RIOPlugin r_io_plugin_zip = {
 	.name = "zip",
-	.desc = "Open zip files [apk|ipa|zip|zipall]://[file//path]",
+	.desc = "Open zip files",
+	.uris = "zip://,apk://,ipa://,jar://,zipall://,apkall://,ipaall://,jarall://",
 	.license = "BSD",
 	.open = r_io_zip_open,
 	.open_many = r_io_zip_open_many,
--- a/libr/io/plugin.c
+++ b/libr/io/plugin.c
@@ -85,6 +85,9 @@ R_API int r_io_plugin_list(RIO *io) {
 		io->cb_printf ("%s  %-8s %s (%s)",
 				str, plugin->name,
 			plugin->desc, plugin->license);
+		if (plugin->uris) {
+			io->cb_printf (" %s", plugin->uris);
+		}
 		if (plugin->version) {
 			io->cb_printf (" v%s", plugin->version);
 		}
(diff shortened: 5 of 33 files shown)


libr/io/plugin.c
```diff
@@ -158,3 +162,105 @@ R_API int r_io_plugin_write_at(RIODesc *desc, ut64 addr, const ut8 *buf, int len
 	}
 	return 0;
 }
+
+R_API char *r_io_plugin_get_uri_prefix (RIOPlugin *plugin) {
```
radare: wtf? why is this necessary? why is this hardcoded?

libr/io/plugin.c
```diff
@@ -158,3 +162,105 @@ R_API int r_io_plugin_write_at(RIODesc *desc, ut64 addr, const ut8 *buf, int len
...
+	char *prefix = NULL;
+	if (!strcmp ("ar", pluginName)) {
+		prefix = "ar://, lib://";
+	}
+	else if (!strcmp ("bfdbg", pluginName)) {
+		prefix = "bfdbg://";
+	}
+	else if (!strcmp ("bfdbg", pluginName)) {
+		prefix = "bfdbg://";
+	}
```
radare: else goes in the same line as the }

libr/io/plugin.c
radare: return type should be const char *

radare:
this information must be provided by the plugin, not hardcoded in any place

libr/io/plugin.c
deepakchethan: All the plugin prefixes are hardcoded in r_io_ar_plugin_open() method like
```
static bool r_io_ar_plugin_open(RIO *io, const char *file, bool many) {
	return !strncmp ("ar://", file, 5) || !strncmp ("lib://", file, 6);
}
````
and there are no internal fields that specify the prefix
I made another pr by adding an internal field inside the RIOPlugin and modifying 34 files. Since you told me not to modify internal fields a couple of times. I closed that PR and changed to this.

The PR in question is #13012. 

Thanks for taking the time to review. If both my ways are wrong. Feel free to close this issue with a message pointing the right direction.

**Comment by deepakchethan** [comment]
@radare thanks for nudging me in the right direction.

Since there was no internal field in RIOPlugin about the list of prefixes, I have created one. Initialized the same on all available plugins.

Could you review the code again ?

File: libr/io/p/io_ar.c [hunk 3]
```diff
@@ -94,6 +94,7 @@ RIOPlugin r_io_plugin_ar = {
 	.name = "ar",
 	.desc = "Open ar/lib files [ar|lib]://[file//path]",
 	.license = "LGPL3",
+	.prefixes = "ar://, lib://",
```
radare: what about making this field an actual array of char* instead of a plain string?

libr/include/r_io.h
```diff
@@ -158,6 +158,7 @@ typedef struct r_io_plugin_t {
 	char *author;
 	char *license;
 	void *widget;
+	char *prefixes;
```
radare: thats wrong, this buffer cant be modified, use const char *.. or const char ** to use {"foo","bar",NULL}

radare:
also the :// is called URI handler, not "prefix", just to clarify concepts and find a better name for that field

deepakchethan:
> also the :// is called URI handler, not "prefix", just to clarify concepts and find a better name for that field

Okay @radare, thanks for letting me know. So I'll change it to `uri_handlers` then?

libr/include/r_io.h
deepakchethan: I will change it to a const buffer.

libr/io/p/io_ar.c
deepakchethan: char ** type cannot be directly assigned like `char **a = {"one", "two", NULL}`. One way is using `char name[]` but even that requires using `strcpy` if we are assigning after declaration. 
Also since it has no other use other than directly displaying, is it really required?

libr/io/p/io_ar.c
```diff
@@ -94,6 +94,7 @@ RIOPlugin r_io_plugin_ar = {
 	.name = "ar",
 	.desc = "Open ar/lib files [ar|lib]://[file//path]",
 	.license = "LGPL3",
+	.uri_handlers = "ar://, lib://",
```
xarkes: Why don't you use an actual list rather than a plain `char *`?

libr/io/p/io_ar.c
xarkes: Sorry I didn't see that comment, ideally the json should yield a list for the available prefixes, so it's actually parsable :)
Something like:
```
{"IO_Plugins":[{"Permissions":"rw_","Name":"ar","Description":"Open ar/lib files [ar|lib]://[file//path]","License":"LGPL3","prefixex": ["ar://", "lib://"]}....}
```

deepakchethan: Okay xarxes, I will make the change and push them by tomorrow.

radare:
Const char* and it should work if u define a reference

radare:
I prefer just “uris”

libr/include/r_io.h
```diff
@@ -158,6 +158,7 @@ typedef struct r_io_plugin_t {
 	char *author;
 	char *license;
 	void *widget;
+	const char *uri_handlers;
```
radare: rename to uris

libr/io/p/io_ar.c
radare: const char ** should be fine, but we can also use comma separated values and use r_str_split_list, but then i would not add a space after the comma

radare:
btw for simplicity creating bindings or plugins in other languages (rust, vala, python ,...) having just char * instead of char ** would be easier, so maybe we can just go for the comma separated string, without a space

deepakchethan:
Okay @radare, I will make the changes

deepakchethan:
I have converted the field to `uris` and removed the spaces in the `uris` field on all plugins. In the json parsing code. I used the split method to split the uris and display them in the form of the array. 

The raw output now is:
`{"IO_Plugins":[{"permissions":"rw_","name":"ar","description":"Open ar/lib files [ar|lib]://[file//path]","license":"LGPL3","uris":["ar://","lib://"]},{"permissions":"rw_","name":"bfdbg","description":"BrainFuck Debugger (bfdbg://path/to/file)","license":"LGPL3","uris":["bfdbg://"]}..`


The parsed output is: 
screen shot 2019-02-10 at 10 54 50 pm

Ready for the review @radare

radare:
now we can probably remove the uris from the description, and show them in `r2 -L`

libr/io/plugin.c
```diff
@@ -100,28 +100,53 @@ R_API int r_io_plugin_list(RIO *io) {
...
+			char *uri;
+			char *uris = strdup (plugin->uris);
+			RList *plist = r_str_split_list (uris, ",");
+			RListIter *piter;
+			pj_k (pj, "uris");
+			pj_a (pj);
+			r_list_foreach (plist, piter, uri) {
+				pj_s (pj, uri);
+			}
+			pj_end (pj);
```
radare: r_list_free (plist); in here

radare: free (uris); its leaking here too

libr/io/plugin.c
```diff
@@ -100,28 +100,53 @@ R_API int r_io_plugin_list(RIO *io) {
...
+			pj_ks (pj, "author", plugin->author);
 		}
-		io->cb_printf ("}");
+		pj_end (pj);
 		n++;
 	}
-	io->cb_printf("]}");
+	pj_end (pj);
+	pj_end (pj);
+	io->cb_printf (pj_string(pj));
```
radare: missing space before (

libr/io/plugin.c
```diff
@@ -100,28 +100,53 @@ R_API int r_io_plugin_list(RIO *io) {
 R_API int r_io_plugin_list_json(RIO *io) {
 	RIOPlugin *plugin;
 	SdbListIter *iter;
+	PJ *pj = pj_new();
```
radare: missing space

libr/io/p/io_zip.c
```diff
@@ -651,6 +651,7 @@ static int r_io_zip_close(RIODesc *fd) {
 RIOPlugin r_io_plugin_zip = {
 	.name = "zip",
 	.desc = "Open zip files [apk|ipa|zip|zipall]://[file//path]",
```
radare: remove uri information in the description and improve the text

libr/io/p/io_http.c
```diff
@@ -104,7 +104,7 @@ static RIODesc *__open(RIO *io, const char *pathname, int rw, int mode) {
 
 RIOPlugin r_io_plugin_http = {
 	.name = "http",
-	.desc = "http get (http://rada.re/)",
+	.desc = "Make http get requests",
```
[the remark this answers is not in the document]
deepakchethan: Done radare. Anymore changes?

libr/io/plugin.c
```diff
@@ -100,28 +103,55 @@ R_API int r_io_plugin_list(RIO *io) {
...
+	PJ *pj = pj_new ();
+	if (!pj) {
+		return NULL;
+	}
+	
 	char str[4];
 	int n = 0;
-	io->cb_printf("{\"IO_Plugins\":[");
+	pj_o (pj);
+	pj_k (pj, "IO_Plugins");
```
radare: i dont like this camel snake case. maybe better to just use "io_plugins"

radare:
please rebase

libr/io/plugin.c
deepakchethan: I changed it.

deepakchethan:
I have squashed and rebased the commits @radare 

Thanks for helping me help with this issue.

PR #13014, opened 2019-02-07
Support typedef to anonymous struct in more than 1 header

kazarmy:


shlr/tcc/libtcc.c
```diff
@@ -355,7 +355,7 @@ static int tcc_compile(TCCState *s1)
 	preprocess_init (s1);
 
 	funcname = "";
-	anon_sym = SYM_FIRST_ANOM;
+	// anon_sym = SYM_FIRST_ANOM;
```
XVilka: just remove

kazarmy: ok 808c51232511407949adc57e2e47cd7c725fcfd4

PR #13016, opened 2019-02-07
afvsj refactor using pj.c

a-meth:


libr/anal/var.c
```diff
@@ -959,7 +987,9 @@ R_API void r_anal_var_list_show(RAnal *anal, RAnalFunction *fcn, int kind, int m
 		}
 	}
 	if (mode == 'j') {
-		anal->cb_printf ("]");
+		pj_end (pj);
+		anal->cb_printf ("%s\n", pj_string (pj));
```
Maijin: probably want to use println as pancake said already

a-meth: Yea sure

a-meth: println isn't available in anal->cb_ right?

Maijin: dunno, @radare ?

libr/anal/var.c
```diff
@@ -891,30 +905,44 @@ R_API void r_anal_var_list_show(RAnal *anal, RAnalFunction *fcn, int kind, int m
...
-						var->name, var->type, anal->reg->name[R_REG_NAME_SP],
-						(st64)R_ABS(var->delta));
+					pj_o (pj);
+					pj_ks (pj, "name", var->name);
+					pj_ks (pj, "kind", "var");
+					pj_ks (pj, "type", var->type);
+					pj_k (pj, "ref");
+					pj_o (pj);
+					pj_ks (pj, "base", anal->reg->name[R_REG_NAME_SP]);
+					char print_offset[32];
```
ret2libc: mmm is this "-XXX" really necessary? isn't it enough to print the number with `pj_ki`?

a-meth: I didn't want to change the output of the previous code. Should I use pj_kn instead?

ret2libc: @radare ^ ?

radare: pj_i /ki is for %d, aka 32bit, not 64. so it will require a cast in here, the var->delta shouldnt really require 64bit, so i would go for pj_ki (pi, "offset", (int)var->delta);

libr/anal/var.c
radare: nah, this is fine, noneed to expose more callbacks

libr/anal/var.c
```diff
@@ -872,17 +876,27 @@ R_API void r_anal_var_list_show(RAnal *anal, RAnalFunction *fcn, int kind, int m
...
-						var->name, var->type, anal->reg->name[R_REG_NAME_BP],
-						(st64)-R_ABS (var->delta));
+					pj_o (pj);
+					pj_ks (pj, "name" ,var->name);
+					pj_ks (pj, "kind", "var");
+					pj_ks (pj, "type", var->type);
+					pj_k (pj, "ref");
+					pj_o (pj);
+					pj_ks (pj, "base", anal->reg->name[R_REG_NAME_BP]);
+					pj_kn (pj, "offset", (st64)-R_ABS (var->delta));
```
radare: use pj_ki, because kn is unsigned, you should cast to int probably, but this will work without this -R_ABS magic

PR #13024, opened 2019-02-08
Compute hashes only with a new r2 flag

devnexen:


Maijin:
Could you also check the output of iFj and rewrite it with pj.c ? Right now the JSON produced is invalid and it's needed for cutter - https://github.com/radareorg/cutter/issues/1179. See example of pj rewrite here https://github.com/radare/radare2/pull/12822/files.

devnexen:
There is already another PR to address these.

xarkes:
I think there is the man page to update too.

binr/radare2/radare2.c
```diff
@@ -707,6 +709,9 @@ int main(int argc, char **argv, char **envp) {
 			threaded = true;
 			break;
 #endif
+		case 'T':
```
radare: why -T, if its iF? maybe we should use iT instead of iF? and keep -T here

Maijin:
iFj produces invalid json @devnexen, if you have time please have a look, otherwise I will send a PR later.
```
[0x1000011ec]> iFj
{"md5 dd111f5744e87c6eedd5f978833a0cdb
sha1 1461b5d708822709d40337192c84ba1bed4f6622"}
```

devnexen:
Did you try with last version ? It is not iF anymore anyway.

radare:
I did the change. Its it and rabin2 -t . So its now consistent

Maijin:
```
[0x1000011ec]> itj
{"values":"md5 dd111f5744e87c6eedd5f978833a0cdb\nsha1 1461b5d708822709d40337192c84ba1bed4f6622"}
[0x1000011ec]>
```

Still could be improved

philipmnelson:
I'm not seeing this using `itj`

Maijin:
Update your r2 again

devnexen:
> ```
> [0x1000011ec]> itj
> {"values":"md5 dd111f5744e87c6eedd5f978833a0cdb\nsha1 1461b5d708822709d40337192c84ba1bed4f6622"}
> [0x1000011ec]>
> ```
> 
> Still could be improved

"Problem" is hashes is just plain string as "informational" not a list or hashtable to be able to tweak the output that much.

Maijin:
Should be modify to output a proper JSON "md5": "", "sha1": "" etc.

radare:
Strings can be easily parsed

Maijin:
:/ That's not the point of outputing JSON if you have to parse it afterwards :p

devnexen:
> Strings can be easily parsed
> […](#)
> On 11 Feb 2019, at 10:36, David CARLIER ***@***.***> wrote: [0x1000011ec]> itj {"values":"md5 dd111f5744e87c6eedd5f978833a0cdb\nsha1 1461b5d708822709d40337192c84ba1bed4f6622"} [0x1000011ec]> Still could be improved "Problem" is hashes is just plain string as "informational" not a list or hashtable to be able to tweak the output that much. — You are receiving this because you modified the open/close state. Reply to this email directly, view it on GitHub, or mute the thread.

So much efforts for this :-) ... might be more straightforward to turn this into separate data.

radare:
Parsing this text is just 3 or 4 lines in C i dont see that much complication. Agree on that it should be saved in a more structured way but that wont be memory efficient or simpler than just a string so i dont see much value in this case to overengineer too much this

devnexen:
I sort of agree for the mem efficiency less for the possible improvements over time (eg more hashes and/or related infos).

philipmnelson:
@Maijin maybe I'm doing something wrong but this is the output of `itj`, on the latest build of r2.
```
{"core":{"type":"EXEC (Executable file)","file":"reversing/dropshot.exe.vir","fd":3,"size":195072,"humansz":"190.5K","iorw":false,"mode":"r-x","obsz":0,"block":256,"format":"pe"},"bin":{"arch":"x86","baddr":4194304,"binsz":195072,"bintype":"pe","bits":32,"canary":false,"retguard":false,"sanitiz":false,"class":"PE32","cmp.csum":"0x0003d4c1","compiled":"Mon Feb  8 01:15:47 1999","crypto":false,"dbg_file":"","endian":"little","havecode":true,"hdr.csum":"0x00000000","guid":"","intrp":"","laddr":0,"lang":"","linenum":false,"lsyms":false,"machine":"i386","maxopsz":16,"minopsz":1,"nx":true,"os":"windows","overlay":false,"pcalign":0,"pic":true,"relocs":false,"rpath":"","signed":false,"static":false,"stripped":true,"subsys":"Windows GUI","va":true,"checksums":{}}}

```

I don't see any hashes listed here. I'm probably doing something wrong but I just wanted to check.

Issue #12870, opened 2019-01-21
Merge/Kill quintuplicate JSOn implementation


Maijin:
Now we have:

- pj.c
- r_json
- sdbjson
- dsojson
- handmade printf json

We must only keep pj.c and likely have sdb uses pj.c too.

radare:
Sdb json api is for prsing. So i would just move pj.c into sdb and kill the rest

radare:
Rjson generates a memory structure so is usel too

thestr4ng3r:
> Rjson generates a memory structure so is usel too

What is rjson?

Maijin:
@thestr4ng3r it's no more ;) It was a duplicate of pi.c

PR #13025, opened 2019-02-09
Refactored rasm2 to use pj.c

deepakchethan:
For issue  #12870

### DIFF
--- a/binr/rasm2/rasm2.c
+++ b/binr/rasm2/rasm2.c
@@ -122,6 +122,10 @@ static int show_analinfo(RAsmState *as, const char *arg, ut64 offset) {
 	ut8 *buf = (ut8 *)strdup ((const char *)arg);
 	int ret, len = r_hex_str2bin ((char *)buf, buf);
 	PJ *pj = pj_new ();
+	if (!pj) {
+		return NULL;
+	}
+	
 	RAnalOp aop = { 0 };
 	
 	if (as->json) {
@@ -175,6 +179,9 @@ static void rasm2_list(RAsmState *as, const char *arch) {
 	RAsmPlugin *h;
 	RListIter *iter;
 	PJ *pj = pj_new ();
+	if (!pj) {
+		return;
+	}
 	if (as->json) {
 		pj_o (pj);
 	}
@@ -221,11 +228,14 @@ static void rasm2_list(RAsmState *as, const char *arch) {
 			if (as->quiet) {
 				printf ("%s\n", h->name);
 			} else if (as->json) {
-				const char *str_bits = "[32, 64]";
 				const char *license = "GPL";
 				pj_k (pj, h->name);
 				pj_o (pj);
-				pj_ks (pj, "bits", str_bits);
+				pj_k (pj, "bits");
+				pj_a (pj);
+				pj_i (pj, 32);
+				pj_i (pj, 64);
+				pj_end (pj);
 				pj_ks (pj, "license", license);
 				pj_ks (pj, "description", h->desc);
 				pj_ks (pj, "features", feat);
--- a/binr/rasm2/rasm2.c
+++ b/binr/rasm2/rasm2.c
@@ -62,7 +62,7 @@ static char *stackop2str(int type) {
 }
 
 // TODO: use pj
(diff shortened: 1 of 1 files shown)


binr/rasm2/rasm2.c
```diff
@@ -136,22 +135,23 @@ static int show_analinfo(RAsmState *as, const char *arg, ut64 offset) {
...
-			printf (",");
-		}
+		showanal (as, &aop, offset, buf + ret, len - ret, pj);
 		ret += aop.size;
 		r_anal_op_fini (&aop);
 	}
 	if (as->json) {
-		printf ("]");
+		pj_end (pj); 
+		printf ("%s\n", pj_string (pj));
```
Maijin: Use println

binr/rasm2/rasm2.c
```diff
@@ -239,7 +245,9 @@ static void rasm2_list(RAsmState *as, const char *arch) {
 		}
 	}
 	if (as->json) {
-		printf ("}\n");
+		pj_end (pj);
+		printf ("%s\n", pj_string (pj));
```
Maijin: same here

binr/rasm2/rasm2.c
```diff
@@ -124,9 +121,11 @@ static int showanal(RAsmState *as, RAnalOp *op, ut64 offset, ut8 *buf, int len)
 static int show_analinfo(RAsmState *as, const char *arg, ut64 offset) {
 	ut8 *buf = (ut8 *)strdup ((const char *)arg);
 	int ret, len = r_hex_str2bin ((char *)buf, buf);
+	PJ *pj = pj_new ();
```
Maijin: Add a check if !pj, check other part that uses pj for examples.

Maijin:
Also please add tests here if the features you modify aren't tested yet: https://github.com/radare/radare2-regressions/blob/master/new/db/tools/rax2

binr/rasm2/rasm2.c
```diff
@@ -220,10 +221,15 @@ static void rasm2_list(RAsmState *as, const char *arch) {
...
 				printf ("%s\n", h->name);
 			} else if (as->json) {
-				const char *str_bits = "32, 64";
+				const char *str_bits = "[32, 64]";
 				const char *license = "GPL";
-				printf ("\"%s\":{\"bits\":[%s],\"license\":\"%s\",\"description\":\"%s\",\"features\":\"%s\"}%s",
-					h->name, str_bits, license, h->desc, feat, iter->n? ",": "");
+				pj_k (pj, h->name);
+				pj_o (pj);
+				pj_ks (pj, "bits", str_bits);
```
radare: well you can use pj_a and push two 32 and 64 and pj_i instead of pass the hardcoded string.

this is not correct because that array will be quoted as a string and it should be an array

binr/rasm2/rasm2.c
radare: nope, this is fine, rasm2 doesnt uses RCons. so just using printf is fine, you can use fputs(pj_string(pj)) to avoid the format thing, but i'¡m fine with printf

binr/rasm2/rasm2.c
deepakchethan: Added the null check

binr/rasm2/rasm2.c
deepakchethan: I made the changes radare

deepakchethan:
> Also please add tests here if the features you modify aren't tested yet: https://github.com/radare/radare2-regressions/blob/master/new/db/tools/rax2

The rasm -Aj test is failing because the expected output is 
```
EXPECT='[{"opcode":0,"bytes":"55","type":"upush","stackop":"unknown","esil":"rbp,8,rsp,-,=[8],8,rsp,-=","stackptr":8}
]'
```
The modified one eliminates the newline between } and ].

I have made a pr requesting to change that test at https://github.com/radare/radare2-regressions/pull/1691.

Ready for review @Maijin

radare:
looks red

binr/rasm2/rasm2.c
Maijin: Feel free to click the "resolve conversation" when you fix something

Issue #10704, opened 2018-07-11
Implement dir.tmp and dir.home


radare:

ret2libc:
Already implemented. Closing.

PR #13026, opened 2019-02-09
WIP: Implemented dir.home and dir.tmp

deepakchethan:
For issue #10704


libr/core/cconfig.c
```diff
@@ -2921,6 +2921,8 @@ R_API int r_core_config_init(RCore *core) {
 	}
 	SETCB ("dir.source", "", &cb_dirsrc, "Path to find source files");
 	SETPREF ("dir.types", "/usr/include", "Default path to look for cparse type files");
+	SETPREF ("dir.home", r_sys_getenv (R_SYS_HOME), "Default path for the home directory");
```
radare: should have a callback that sets those env vars when the eval vars are modified, the getter should be a direct wrapper to r_sys_getenv... but thats gonna be tricky

radare:
ping

libr/core/cconfig.c
deepakchethan: Sorry @radare for replying late. I am going through the other example source to see how to do what you said. If this requires an urgent fix, you can close this PR or I really want to help with this.

radare:
Its fine like this. Not in a hurry but its simple to solve and having the WIP in the title is enough, no need to close

deepakchethan:
Ping Reply.

@radare now I have added callbacks. The output looks as follows.

screen shot 2019-02-11 at 2 49 51 pm

I am still looking about adding the wrapper instead of the r_get_env, It would help if you could tell me why it is required.

radare:
In the case the user modifies the environment using the % command like this: %HOME=...

(or any r2pipe script executed from inside r2 can change the environment variables, and the eval vars will not be in sync). Thats my only concern about not merging this, but imho its good to merge as long as it solves a bunch of portability issues and will work for most of the use cases.

radare:
Can  you please rebase to make it green?

deepakchethan:
Thanks for responding @radare, I have rebased the changes. It does sound tricky.

PR #13030, opened 2019-02-09
Added feature to return error while parsing c header file

m0nk3ydluffy:
Currently, `r_parse_c_file()` and `r_parse_c_string()` do not return the error's that occurred while parsing the header file or string. I have made the changes so that we can get to know if an error occurred and the details regarding the error.


XVilka:
LGTM except `errorfunc`.

m0nk3ydluffy:
@XVilka Please suggest some improvements regarding `errorfunc`?

PR #13036, opened 2019-02-09
iS= now accepts eco (theme) settings + highlighting of sections with w/x permissions

mkopec3:
Hi,

Here's a change proposal to allow iS= command interpreter eco (theme) settings.
Additionally, sections with w/x permissions are highlighted, what I feel might be useful feature for security researchers.

Check the pictures below to better understand the change (eco twilight was used).

**Before the change:**
iseq_before

**After the change:**
iseq_after


Maijin:
Maybe you should check om= too :)

mkopec3:
Ok, I'll investigate it...

PR #13042, opened 2019-02-10
om= now interprets theme (eco) settings

mkopec3:
Done according to Maijin's suggestion from PR #13036 

**Before the change:**
omeq_before

**After the change:**
omeq_after


Maijin:
Great! see also `o=` this one doesn't have color at all.

radare:
maybe it will be nice to also colorize the '#'

libr/core/cmd_open.c
```diff
@@ -205,10 +205,12 @@ static void list_maps_visual(RIO *io, ut64 seek, ut64 len, int width, int use_co
...
 				color_end = Color_RESET;
-				if (s->perm & R_PERM_X) { // exec bit
-					color = Color_GREEN;
+				if ((s->perm & R_PERM_X) && (s->perm & R_PERM_W)) { // exec & write bits
+					color = r_cons_singleton()->context->pal.comment;
+				} else if (s->perm & R_PERM_X) { // exec bit
+					color = r_cons_singleton()->context->pal.args;
 				} else if (s->perm & R_PERM_W) { // write bit
-					color = Color_RED;
+					color = r_cons_singleton()->context->pal.input;
```
radare: i think the colors should be:

is_writable = graph.false # red by default
is_executable = graph.true # green by default
is_exec+is_writable = graph.trufae # blue by default

mkopec3:
Done.
Adjusted this also for iS= (for consistency purpose)
However I would propose to leave _widget_sel_ for w/x on sections, just to highlight the fact that it's really bad when a section has those both permissions set together (e.g. executable stack)

mkopec3:
Regarding '#' colorization, I can adjust it later

radare:
ok

Issue #12996, opened 2019-02-04
Improve r2 -t


radare:
Use tasks instead of threads, because right now its crashing because of that.

Also do the analysis in background if -A is used

radare:
ping @thestr4ng3r

thestr4ng3r:
There are other things like task-local temprary eval vars that I would prioritize over -t.

radare:
ping @thestr4ng3r can you bring it back?

thestr4ng3r:
yes

PR #13048, opened 2019-02-11
Disable -t and deobfuscate -n variable

thestr4ng3r:
#12996

### DIFF
--- a/binr/radare2/radare2.c
+++ b/binr/radare2/radare2.c
@@ -1,6 +1,7 @@
 /* radare - LGPL - Copyright 2009-2018 - pancake */
 
 #define USE_THREADS 1
+#define ALLOW_THREADED 0
 #define UNCOLORIZE_NONTTY 0
 #ifdef _MSC_VER
 #ifndef WIN32_LEAN_AND_MEAN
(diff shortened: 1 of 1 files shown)


radare:
Why disable -t instead of making it work with tasks?

thestr4ng3r:
I don't have the time to do it right now since it requires a lot of refactoring, so I would like to disable it for now because it is completely broken anyway.

radare:
add a warning message at least

thestr4ng3r:
You mean a warning if you do -t while it is disabled?

radare:
yes

radare:
do that and i'll merge, but i dnt see much problem in just calling r_core_cmdf(core, "& aa"); shouldnt be hard to do and should work

thestr4ng3r:
No it's not that easy. -t should also load bin info in the background and there are other things between that and analysis. Also analysis is still not completely safe to run in tasks nevertheless, even if it works in practice.

radare:
opening the bin info can be "& oba;e io.va=1"

thestr4ng3r:
Yes, but there is more stuff in between and I don't won't to do any code duplication.

PR #13050, opened 2019-02-11
Fix x86_64 socket syscall

blenk92:
Currently syscall nr 41 is named socketcall. However on x86_64
syscall nr 41 is named socket and expects 3 arguments.

Infos taken from https://github.com/strace/strace/blob/master/linux/x86_64/syscallent.h

radare2 version:
radare2 3.3.0-git 20947 @ linux-x86-64 git.3.1.3-565-g7f8495dea
commit: 7f8495dea766b7b88931b253f7bc55f84198ef91 build: 2019-02-11__16:29:24


radare:
Can you add a test case showing how it looks good in the disasm?

blenk92:
> Can you add a test case showing how it looks good in the disasm?

https://github.com/radare/radare2-regressions/pull/1693
is this suitable?

PR #13055, opened 2019-02-12
Fix false positive string issues on arm64 when adrp+add ##disasm

radare:


radare:
maybe this should be an arm-specific check? or just for arm64? or maybe an option?  time will say because users will never say

PR #13066, opened 2019-02-12
Further Android detection via elf and symbols.

devnexen:


libr/bin/p/bin_elf.inc
```diff
@@ -966,6 +966,13 @@ static void lookup_symbols(RBinFile *bf, RBinInfo *ret) {
 				is_rust = true;
 				ret->lang = "rust";
 			}
+			if (!is_android && strstr (symbol->name, "android_")) {
```
Maijin: Not a fan of the strstr that's too generic imho

devnexen: That s not false ... it is just a little improvement for a difficult recognition :-) I wish Android had its own elf osabi :-)

Maijin: I'm not saying it's false, I'm saying that's too generic, the chance of getting a symbol containing  "android_" on a non-android executable is fairly high

devnexen:
After reconsidering, closing ...

Issue #13056, opened 2019-02-12
Add more examples for pf??


radare:

Maijin:
Done in https://github.com/radare/radare2/pull/13073

PR #13073, opened 2019-02-13
Added few examples of pf

deepakchethan:
for issue #13056


radare:
i think it will be interesting to add some examples using pf references like the ones in r2r/new/db/cmd/cmd_pf as well as real world use cases not just random format strings

deepakchethan:
I will add them now.

deepakchethan:
@radare sorry for taking so long. I was working on other minor PR. I updated the example list. Please check it when you are free :)

libr/core/cmd_print.c
```diff
@@ -361,14 +361,24 @@ static const char *help_detail2_pf[] = {
...
 	"pf", " {integer}? (bifc)", "Print integer times the following format (bifc)",
 	"pf", " [4]w[7]i", "Print an array of 4 words and then an array of 7 integers",
-	"pf", " ic...?i foo bar \"(pf xw yo foo)troll\" yo", "Print nested anonymous structres",
-	"pf", "n2", "print signed short (2 bytes) value. Use N insted of n for printing unsigned values",
+	"pf", " ic...?i foo bar \"(pf xw yo foo)troll\" yo", "Print nested anonymous structures",
+	"pf", " ;..x", "Print value located 6 bytes from current offset",
+	"pf", " [10]z[3]i[10]Zb", "Print an fixed size str, widechar, and var",
+	"pfj", " +F @ 0x14","Print the content at given offset with flag",
+	"pf", " n2", "print signed short (2 bytes) value. Use N insted of n for printing unsigned values",
+	"pf", " {2}? (plop)structname @ 0", "Print a format n times",
```
jb55: `{n}` struct printing is currently broken (see https://github.com/radare/radare2/issues/5641) and hasn't been fixed for 3 years.

users will most likely want `[2]? (plot) structname` because that appears to work and has the same effect. The only difference is that it prints an array of structs instead of just printing the struct n times.

deepakchethan: Thanks for the review. I updated like you said.

PR #13075, opened 2019-02-13
Various little code fixes.

devnexen:


libr/util/spaces.c
```diff
@@ -4,13 +4,7 @@
...
 R_API RSpaces *r_spaces_new(const char *name) {
 	RSpaces *sp = R_NEW0 (RSpaces);
-	if (!sp) {
-		return NULL;
-	}
-	if (!r_spaces_init (sp, name)) {
-		R_FREE (sp);
-		return NULL;
-	}
+	(void)r_spaces_init (sp, name);
```
radare: wtf?

devnexen: r_spaces_init free resources already if it fails at any stage.

radare: if sp is null its crashing inside r_spaces_init because there's an assert in there

radare: in case the resources are freed in r_spaces_init the sp is allocated but empty, aka not initialized, so it s wrong too

devnexen: That s why I used the R_FREE macro

radare: rfree sets the pointer to null, but you are passing this pointer as value not as reference, so it wont work

PR #13094, opened 2019-02-14
Fix aaF sort issue, honor a2f and fix big allocation in analysis ##anal

radare:


libr/anal/fcn.c
```diff
@@ -1862,12 +1862,21 @@ R_API RAnalFunction *r_anal_fcn_find_name(RAnal *anal, const char *name) {
 	return NULL;
 }
 
+#define MAXBBSIZE (1024*1024*1)
```
ret2libc: why? yet another MAXBBSIZE defined and yet another random limit... 

```
[775/1136] Compiling C object 'libr/anal/04634f8@@r_anal@sha/fcn.c.o'.
../libr/anal/fcn.c:1865: warning: "MAXBBSIZE" redefined
 #define MAXBBSIZE (1024*1024*1)
 
../libr/anal/fcn.c:20: note: this is the location of the previous definition
 #define MAXBBSIZE (16 * 1024)
```

libr/anal/fcn.c
```diff
@@ -1862,12 +1862,21 @@ R_API RAnalFunction *r_anal_fcn_find_name(RAnal *anal, const char *name) {
...
+R_API bool r_anal_fcn_add_bb(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 size, ut64 jump, ut64 fail, int type, RAnalDiff *diff) {
 	RAnalBlock *bb = NULL, *bbi;
 	RListIter *iter;
 	bool mid = false;
 	st64 n;
+	if (size == 0) { // empty basic blocks allowed?
+		eprintf ("warning: empty basic block at 0x%08"PFMT64x" is not allowed. pending discussion.\n", addr);
+		return false;
+	}
+	if (size > MAXBBSIZE) {
```
ret2libc: why? is 1MB so much that we cannot allocate 1MB of data?

libr/anal/fcn.c
```diff
@@ -1862,12 +1862,21 @@ R_API RAnalFunction *r_anal_fcn_find_name(RAnal *anal, const char *name) {
...
+#define MAXBBSIZE (1024*1024*1)
 /* rename RAnalFunctionBB.add() */
-R_API int r_anal_fcn_add_bb(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 size, ut64 jump, ut64 fail, int type, RAnalDiff *diff) {
+R_API bool r_anal_fcn_add_bb(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 size, ut64 jump, ut64 fail, int type, RAnalDiff *diff) {
 	RAnalBlock *bb = NULL, *bbi;
 	RListIter *iter;
 	bool mid = false;
 	st64 n;
+	if (size == 0) { // empty basic blocks allowed?
+		eprintf ("warning: empty basic block at 0x%08"PFMT64x" is not allowed. pending discussion.\n", addr);
```
ret2libc: use a `r_warn_if_reached`

libr/core/canal.c
```diff
@@ -4766,29 +4766,50 @@ static bool __cb(RFlagItem *fi, void *user) {
 static int __addrs_cmp(void *_a, void *_b) {
 	ut64 a = r_num_get (NULL, _a);
 	ut64 b = r_num_get (NULL, _b);
-        return a - b;
```
ret2libc: why? `a-b` is more compact and it's a standard way to implement cmp.

libr/anal/fcn.c
ret2libc: Also, I think there is already `anal.bb.maxsize` which should be used instead of hardcoding value.s

libr/core/canal.c
```diff
@@ -4766,29 +4766,50 @@ static bool __cb(RFlagItem *fi, void *user) {
...
 		ut64 a1 = r_num_get (NULL, addr2);
-		int sz = a1 - a0;
-		if (sz < 1) {
-			eprintf ("Warning: unsorted flag list\n");
+		if (a0 == a1) {
+			// ignore
+			continue;
+		}
+		if (a0 > a1) {
+			eprintf ("Warning: unsorted flag list 0x%llx 0x%llx\n", a0, a1);
```
ret2libc: use `r_warn_if_fail` instead of the `if (...) { eprintf }`

libr/core/canal.c
```diff
@@ -4766,29 +4766,50 @@ static bool __cb(RFlagItem *fi, void *user) {
...
+			continue;
+		}
+		st64 sz = a1 - a0;
+		if (sz < 1 || sz > MAXFCNSIZE) {
+			eprintf ("Warning: invalid flag range from 0x%08"PFMT64x" to 0x%08"PFMT64x"\n", a0, a1);
```
ret2libc: use `r_warn_if_fail` instead of the `if (... ) { eprintf }`.

libr/core/canal.c
```diff
@@ -4766,29 +4766,50 @@ static bool __cb(RFlagItem *fi, void *user) {
...
+		if (a0 == a1) {
+			// ignore
+			continue;
+		}
+		if (a0 > a1) {
+			eprintf ("Warning: unsorted flag list 0x%llx 0x%llx\n", a0, a1);
+			continue;
+		}
+		st64 sz = a1 - a0;
+		if (sz < 1 || sz > MAXFCNSIZE) {
```
ret2libc: why `MAXFCNSIZE` should be a limit? what's the issue in defining a function which is 1MB far from the last flag?

ret2libc:
This PR wasn't discussed at all and I think it introduces a lot of unnecessary changes (e.g. replacing `a- b` with a long if/else, if(...) eprintf instead of r_warn_if_fail, re-defines existing macros, adds random limits, etc.)

radare:
Will do all those changes in a pr now. But the actual limit helped me to identify a couple of issues in other analysis commands.

radare:
You told me to do all those changws in aaF already. And u dont use it at all, and its now working as expected so calm down your nerves and eat the soup

libr/anal/fcn.c
radare: its just an arbitrary size because 99.9% of the cases that will be a false positive, so better catch it earlier

ret2libc: I think it prevents proper analyses. When packers/obfuscators are used I often find very big basic blocks that are truncated because of limits like this. (e.g.  movfuscator)

libr/anal/fcn.c
radare: this macro doesnt allows me to add any custom format string. its not even clear what will happen if that assert is reached when no logs are shown (will assert, crash, ignore?)

libr/core/canal.c
radare: r_warn_if_fail doesnt allows me to show the offsets that are not sorted, same comments about the rwarn behaviour

libr/core/canal.c
radare: because some libc and sometimes kernels will decide to crash if you alloc 8 TB of ram instead of just tell you that thats probably wrong in your code

libr/anal/fcn.c
ret2libc: mmm as the name implies, it just warns if the line is reached. But even in this case it would be probably better to use `r_warn_if_fail`. We can change it to accept a custom string.

radare:
add a testcase for that movfuscator bin you mention, ive never looked at them, (i know the technique)

ret2libc:
https://github.com/radare/radare2-regressions/commit/26656df484acc86c6c5792f993bdc357eba13205 the test case. Note that here there's also another edge case to consider, which is the fact that there is no ending ret/jmp, so we should block the basic block when we reach the end of the segment.

PR #13096, opened 2019-02-14
Disable function autonaming to avoid confusions ##anal

radare:
screenshot 2019-02-14 at 04 12 08


radare:
10 XX

ITAYC0HEN:
I think it is better not to detect "jmp thunk table" functions and remove "aan". We should prioritize fixing this though. 
As a compromise, we can only apply "anal.autoname" only if:
1. Binary is PE
2. function of size < 10.
3. The referenced flag is "sym.imp."

In these cases, the functions are probably jmp thunks

Maijin:
In elf the elements are marked accordingly as dword or qword

ITAYC0HEN:
What do you mean? And what it. Implies about the solution? I thought that plt got detection in r2 is good

radare:
Feel free to improve aan with your suggestions. But the idea of this pr is to decide if we want it enabled or not by default

ITAYC0HEN:
I vote for disable.

radare:
♿️

ret2libc:
I'm fine with disabling it. It's not like it's that useful usually... it just gets the name of the first called function.

Maijin:
@ret2libc in the case I mentionned, the renaming is mandatory if you want to be able to understand something.

Maijin:
I have added other tools comparison in the issue, only binja is doing as bad as without aan.

radare:
It is very useful actually if u check the failing tests its important for windows binaries which use to have wrappers of wrappers of functions as well as for crackmes where methods get meaningful names related tonthe strings they contain.

radare:
Imho i would keep it enabled for now and wait for next release to do any change on this

ITAYC0HEN:
in my opinion, having something like this is worse than nothing:

image

radare:
i think we should make it an int and make it configurable to choose the level of autonaming rules to use. 0 for disable, 1 for direct trampolines, 2 for more stuff, ...

ITAYC0HEN:
can be merged now.

radare:
did you updated the tests?

radare:
no  wait after release? XD

PR #13098, opened 2019-02-14
Make ASAN green again

ret2libc:


libr/syscall/d/linux-x86-32.sdb.txt
```diff
@@ -363,4 +363,4 @@ pkey_free=0x80,382,1,
 statx=0x80,383,5,
 arch_prctl=0x80,384,2,
 io_pgetevents=0x80,385,6,
-rseq=0x80,386,4
+rseq=0x80,386,4,
```
radare: Not having this trailing comma shouldnt be abreason to crash or not to work

libr/core/cmd_anal.c
```diff
@@ -7470,16 +7470,18 @@ static void cmd_anal_abt(RCore *core, const char *input) {
...
+		ut64 addr = core->offset;
 		char *p;
 		int n = 1;
-		p = strchr (input + 1, ' ');
-		if (p) {
-			*p = '\0';
-			n = *(++p)? r_num_math (core->num, p): 1;
+		input++;
+		if (*input) {
+			p = strchr (input + 1, ' ');
```
radare: Inout is const char* that may segfault or missbehave

libr/core/cmd_anal.c
```diff
@@ -7470,16 +7470,18 @@ static void cmd_anal_abt(RCore *core, const char *input) {
...
+			if (p) {
+				*p = '\0';
+				n = *(++p)? r_num_math (core->num, p): 1;
```
radare: Preincrements are most of the time unnecessary. Better postincrement the expression above and just *p in here withoutbthe parenthesis

libr/syscall/syscall.c
```diff
@@ -184,7 +184,7 @@ R_API RSyscallItem *r_syscall_item_new_from_string(const char *name, const char
 	}
 	o = strdup (s);
 	int cols = r_str_split (o, ',');
-	if (cols < 3) {
+	if (cols < 4) {
```
radare: <3 should be fine too

ret2libc: No, because there is r_str_word_get0 (o, 3), which gets the 4th word. This is why there was an ASAN crash.

libr/core/cmd_anal.c
ret2libc: Ok. I didn't change this code it was already there as usual.

libr/syscall/d/linux-x86-32.sdb.txt
ret2libc: I know, indeed I fixed the reason of the crash (replaced `< 3` with `< 4` but since this is the only entry that is not compliant with the rest of the entries, I also fixed it as well.

libr/core/cmd_anal.c
ret2libc: Right, fixing it.

radare:
Just check the length before taking the 3th element because u know that at the moment of the split

ret2libc:
> Just check the length before taking the 3th element because u know that at the moment of the split

Why the same was not done for the first, the second and the third split? I'm just adapting to what's already there.

I don't see
```C
        if (cols > 0) {
 	    si->swi = (int)r_num_get (NULL, r_str_word_get0 (o, 0));
        }
        if (cols > 1) {
           	si->num = (int)r_num_get (NULL, r_str_word_get0 (o, 1));
         }
         if (cols > 2) {
        	si->args = (int)r_num_get (NULL, r_str_word_get0 (o, 2));
         }
```

radare:
Because syscall number and number of args are kind of more important if u want to have a meaningful query to resolve syscall names/numbers

ret2libc:
@radare ready to be merged IMO.

PR #13102, opened 2019-02-14
Address comments in #13094 related to maxbbsize and maxfcnsize ##anal

radare:


radare:
the a-b thing was wrong because you cant get a signed value out of unsigned values. also, comparing and resulting in -1,0,1 makes the code easier to maintain, i can make a ternary conditional if you want, but the a-b is wrong

ret2libc:
Ok, fine as it is then.

radare:
this "fix" introduces 3 regressions

```
[35m-/ (fcn) fcn.00560e67 6412[39m
[35m[39m[32m+/ (fcn) fcn.00560e67 1366[39m
[32m[39m |           0x00560e67      push esi
 |       ,=< 0x00560e68      jmp 0x560e7d
```

radare:
Still red

ret2libc:
It is red because the technique of using random limits is too fragile.
But btw I'm not even sure the current test is right. It says the function size is 1025, while the malloc is of 1024 bytes, so I'm not sure how the function can be 1025 bytes... At least after this patch the function is of size 1023, which makes more sense IMO.

radare:
the test fails because of the function boundary changed. but i dont think it was even correct before either. because it was 1018 and now its 1020 (dont know where u get the 1024, 1023, 1025 values)

im tempted to merge and update the test and forget about this until the rerlease

ret2libc:
yes please just update that test.

(1024 and 1024 comes from `afi` command. `fcn_last` is not exacty the size of the function... you can check the code)

PR #13103, opened 2019-02-14
Improved json for tsj/tuj and completed size calculator for struct/union

m0nk3ydluffy:
1. Improved `tsj` and `tuj`.
2. Modified function to calculate size of unions along with structs.

Before:
image

After:
image

### DIFF
--- a/libr/core/cmd_type.c
+++ b/libr/core/cmd_type.c
@@ -334,58 +334,17 @@ static void cmd_type_noreturn(RCore *core, const char *input) {
 	}
 }
 
-static int calculate_type_size(Sdb *sdb_types, char *name) {
+static void save_type_size(Sdb *sdb_types, char *name) {
 	char *type = NULL;
-	r_return_val_if_fail (sdb_types && name, 0);
+	r_return_if_fail (sdb_types && name);
 	if (!sdb_exists (sdb_types, name) || !(type = sdb_get (sdb_types, name, 0))) {
-		return 0;
+		return;
 	}
-	char *type_name = r_str_newf ("%s.%s", type, name);
 	char *type_name_size = r_str_newf ("%s.%s.%s", type, name, "size");
-	int size = 0;
-	if (sdb_exists (sdb_types, type_name_size)) { // size already exists so stop recursion
-		char *v = sdb_get (sdb_types, type_name_size, 0);
-		if (v) {
-			size = sdb_atoi (v);
-			free (v);
-		}
-	} else if (!strcmp (type, "struct") || !strcmp (type, "union")) {
-		int idx;
-		char *v = NULL;
-		//loop through the contents to find individual sizes
-		for (idx = 0; (v = sdb_array_get (sdb_types, type_name, idx, NULL)); ++idx) {
-			char *sub_name = sdb_array_get (sdb_types, sdb_fmt ("%s.%s", type_name, v), 0, NULL);
-			char *ptr = NULL;
-			int sz = 0;
-			if (strchr (sub_name, '*')) {
-				// It is a pointer variable so get its size;
-				sz = calculate_type_size (sdb_types, "void *");
-			} else if ((ptr = strchr (sub_name, ' '))) {
-				sz = calculate_type_size (sdb_types, ptr + 1);
-			} else {
-				sz = calculate_type_size (sdb_types, sub_name);
-			}
-			// count stores the number of elements in case of an array
-			int count = sdb_array_get_num (sdb_types, sdb_fmt ("%s.%s", type_name, v), 2, NULL);
-			if (count) {
-				sz *= count;
-			}
-			if (!strcmp (type, "struct")) {
-				// size of struct is sum of sizes of its elements
-				size += sz;
-			} else {
-				// size of union is max of the sizes of its elements
-				size = sz > size ? sz : size;
-			}
-			free (sub_name);
-			free (v);
-		}
-	}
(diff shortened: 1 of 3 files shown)


m0nk3ydluffy:
Some tests are failing because of modifications. I'll send PR in radare2-regressions fixing those tests with the new expected output.

libr/core/cmd_type.c
```diff
@@ -334,6 +334,78 @@ static void cmd_type_noreturn(RCore *core, const char *input) {
...
+	char *type = NULL;
+	if (!sdb_types || !name || !sdb_exists (sdb_types, name) || !(type = sdb_get (sdb_types, name, 0))) {
+		return 0;
+	}
+	char *type_name = r_str_newf ("%s.%s", type, name);
+	char *type_name_size = r_str_newf ("%s.%s.%s", type, name, "size");
+	int size = 0;
+	if (sdb_exists (sdb_types, type_name_size)) { // size already exists so stop recursion
+		char *v = sdb_get (sdb_types, type_name_size, 0);
+		size = sdb_atoi (v);
```
radare: add if (v)

libr/core/cmd_type.c
```diff
@@ -334,6 +334,78 @@ static void cmd_type_noreturn(RCore *core, const char *input) {
...
+		return;
+	}
+	char *str = strdup (parsed);
+	if (str) {
+		char *ptr = NULL;
+		int offset = 0;
+		while ((ptr = strstr (str + offset, "=struct\n")) || (ptr = strstr (str + offset, "=union\n"))) {
+			*ptr = 0;
+			char *name = ptr;
+			while (name - 1 >= str && *(name - 1) != '\n' && *(name - 1) != 0) {
```
radare: != 0 is implicit, the *name-1) things are scary, better not to do negative detlas with pointers its a bad practice

libr/core/cmd_type.c
```diff
@@ -334,6 +334,78 @@ static void cmd_type_noreturn(RCore *core, const char *input) {
...
+	}
+	sdb_set (sdb_types, type_name_size, sdb_fmt ("%d", size), 0); // store the size for fast access
+	free (type);
+	free (type_name);
+	free (type_name_size);
+	return size;
+}
+
+R_API void save_parsed_type_size(RCore *core, const char *parsed) {
+	if (!core || !core->anal || !parsed) {
```
radare: use r_return_if_fail

libr/core/cmd_type.c
```diff
@@ -334,6 +334,78 @@ static void cmd_type_noreturn(RCore *core, const char *input) {
 	}
 }
 
+static int calculate_type_size(Sdb *sdb_types, char *name) {
+	char *type = NULL;
+	if (!sdb_types || !name || !sdb_exists (sdb_types, name) || !(type = sdb_get (sdb_types, name, 0))) {
```
radare: mauybe better to use r_return_val_if_fail for sdb_types + name chk

libr/core/cmd_type.c
```diff
@@ -375,6 +448,23 @@ static int stdifstruct(void *user, const char *k, const char *v) {
 	return false;
 }
 
+static int print_struct_union_list_json_cb(void *p, const char *k, const char *v) {
+	RCore *core = (RCore *)p;
+	PJ *pj = pj_new ();
```
radare: chk for null

libr/core/cmd_type.c
```diff
@@ -375,6 +448,23 @@ static int stdifstruct(void *user, const char *k, const char *v) {
...
+	pj_o (pj);
+	Sdb *sdb = core->anal->sdb_types;
+	char *sizecmd = r_str_newf ("%s.%s.size", v, k);
+	char *size_s = sdb_querys (sdb, NULL, -1, sizecmd);
+	pj_ks (pj, "type", k);
+	pj_ki (pj, "size", size_s ? atoi (size_s) : 0);
+	pj_end (pj);
+	r_cons_printf ("%s", pj_string (pj));
```
radare: use "%s\n"

m0nk3ydluffy: This would result in newlines between json objects:
```
[{"type":"a","size":1400}
,{"type":"addr","size":1368}
]
```
rather than this:
`[{"type":"a","size":1400}, {"type":"addr","size":1368}]`
Should I still make the changes?

libr/core/cmd_type.c
m0nk3ydluffy: Okay! Thanks. Will keep the negative deltas in mind. For now I changed this to:
```
while (name - 1 >= str && *(name - 1) != '\n') {
	name--;
}
```
Ensuring `name - 1 >= str` before the other check should be safe for now.

libr/core/cmd_type.c
ITAYC0HEN: json object should be executed only at the end. Also, instead of "\n" use `r_cons_println (pj_string (pj));`

radare:
struct.0.size=0

i see all the structs sized as 0 in the failing tests

m0nk3ydluffy:
> struct.0.size=0�
> 
> i see all the structs sized as 0 in the failing tests

Yes! That is expected because in all those test cases, we are first removing all the types(including basic ones like char, int, long, etc.) by using `t-*`. Since calculation of the sizes of structs depends on the basic types, if we remove the basic types, there is no way to calculate the sizes of any data type thereby resulting in zero.

radare:
if you make a pr in r2r with the updated tests this pr will be green

m0nk3ydluffy:
@radare I closed the old PR in r2r and have made another PR in r2r with some new tests.

PR #13108, opened 2019-02-14
Do not escape path to Windows file

GustavoLCR:
Paths in windows contain '\\', escaping breaks things.

### DIFF
--- a/binr/radare2/radare2.c
+++ b/binr/radare2/radare2.c
@@ -1081,8 +1081,13 @@ int main(int argc, char **argv, char **envp) {
 				f = r_acp_to_utf8 (f);
 #	endif // __WINDOWS__
 				if (f) {
-					char *escaped_path = r_str_arg_escape (f);
-					pfile = r_str_append (pfile, escaped_path);
+					char *escaped_path = NULL;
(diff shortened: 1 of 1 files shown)


binr/radare2/radare2.c
```diff
@@ -1081,8 +1081,13 @@ int main(int argc, char **argv, char **envp) {
 				f = r_acp_to_utf8 (f);
 #	endif // __WINDOWS__
 				if (f) {
-					char *escaped_path = r_str_arg_escape (f);
-					pfile = r_str_append (pfile, escaped_path);
+					char *escaped_path = NULL;
+#		if !__WINDOWS__
```
radare: what about duplicating the whole body like this, looks nicer to read imho:

#if __WINDOWS__
char *escaped_path = NULL;
#else 
char *escaped_path = r_str_arg_escape (f);
#endif

radare: well in fact this can be 2 statements with pfile = r_str_append (pfile, r_str_arg_escape (f))

radare:
Why is travis red? Missing rebase?

GustavoLCR:
I think the regression is in master

PR #13111, opened 2019-02-15
Initial import of RAnalDataType and fix ref string comments in Dalvik ##anal

radare:


radare:
@kazarmy can  you review this pr?

libr/core/disasm.c
```diff
@@ -2755,6 +2755,22 @@ static void ds_instruction_mov_lea(RDisasmState *ds, int idx) {
 	case R_ANAL_OP_TYPE_CMOV:
 	case R_ANAL_OP_TYPE_MOV:
 		src = ds->analop.src[0];
+		if (ds->analop.ptr && !ds->analop.refptr) {
```
kazarmy: Printing of strings for MOV-type instructions with no flag at referenced addr is done somewhere else in disasm.c ... are Dalvik strings being filtered out?

kazarmy:
Printing of strings for MOV-type instructions with no flag at referenced addr is done somewhere else in disasm.c ... are Dalvik strings being filtered out?

libr/core/disasm.c
```diff
@@ -2755,6 +2755,22 @@ static void ds_instruction_mov_lea(RDisasmState *ds, int idx) {
...
+			ut64 addr = ds->analop.ptr;
+			if (r_io_is_valid_offset (core->io, addr, 0)) {
+				if (!r_flag_get_i (core->flags, addr)) {
+					char *s = r_core_cmd_strf (core, "ps @ 0x%"PFMT64x, addr);
+					if (s) {
+						r_str_trim (s);
+						_ALIGN;
+						if (*s) {
+							ds_comment (ds, true, "; \"%s\"", s);
```
kazarmy: Printing of strings for MOV-type instructions with no flag at referenced addr is done somewhere else in disasm.c ... are Dalvik strings being filtered out?

radare: Yep thats probably the reason, dalvik uses pascal style strings, so first byte defines the length. its probably ignored because of starting with a non printable char

kazarmy: I think you need to make the code arch-specific or controllable by config var

radare:
the strings problem is solved better here: https://github.com/radare/radare2/pull/13135

PR #13114, opened 2019-02-15
Fix a parsing issue of config in panels ##panels

Vane11ope:
way easier to deal with json for the panels config if the "caching" parameters were int and there's no problem doing it so.

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -1,4 +1,4 @@
-/* Copyright radare2 2014-2018 - Author: pancake, vane11ope */
+/* Copyright radare2 2014-2019 - Author: pancake, vane11ope */
 
 // pls move the typedefs into roons and rename it -> RConsPanel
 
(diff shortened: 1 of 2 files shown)


libr/core/panels.c
```diff
@@ -2946,7 +2946,7 @@ static void savePanelsLayout(RCore* core, bool temp) {
 		pj_kn (pj, "y", panel->pos.y);
 		pj_kn (pj, "w", panel->pos.w);
 		pj_kn (pj, "h", panel->pos.h);
-		pj_kb (pj, "caching", panel->caching);
+		pj_kn (pj, "caching", panel->caching);
```
radare: n expects ut64, this is int, and i dont see why you want caching to be int instead of bool

Vane11ope: so how do you pass char* "true" as bool true, i mean i can do it somehow but i can save it if i make it int at the first place

radare:
wat? pj_kb should use true,false from !0 and 0.

Vane11ope:
what do you even mean.

json is gonna look like {"Title":"Disassembly","Cmd":"pd $r","x":0,"y":1,"w":97,"h":78,"caching":false} then when i parse it, i get char* "false", how do i translate that to bool false, actual working code would help me understand a lot thanks.

radare:
Pls dont make json camel case

Vane11ope:
?????????

anyway a bug remains unless this gets merged or you show me alternative. 

i dont really understand why you stick to bool so much but i leave this here to you.

radare:
How do u parse this to get a string out of a bool? Also if u just want to show a number 0,1 bool is your friend. So just changing 1 line will be enough. Keep all the other lines intact

PR #13121, opened 2019-02-16
Disable ptrace-wrap for WSL

GustavoLCR:
First, this requires radare/acr#20, so that we can auto-generate a configure that can recognize 'wsl'


radare:
Run autogen.sh and commit the autogenerated configure too

GustavoLCR:
Had to edit the file because it was leaving `@VERSION@` in the configure

PR #13119, opened 2019-02-15
Autoname PE import thunks

pelijah:
Related: #11759

PR #13124, opened 2019-02-16
Fix #11759 - aac now names import trampolines

### LINKED ISSUE #11759: Fix PE files call to sym.imp if aan is disabled
In many PE the `aan` is absolute necessity to be able to analyze but it shouldn't be that way if the mechanism of propagating `sym.imp.xxx` would be working correctly.

`aaa` with `aan` (`e anal.autoname=true`):
image

`aaa` without `aan` (`e anal.autoname=false`):
image

But following the first call of this function the function is clearly identified:
image

The expected is to have the output without `aan` required:

`call sub.kernel32.dll_FindResourceA` for example

See example below:
[f15985a91b12e015812f7f960eb0abe03d0cd401f54d0088f75cb087776a297c.zip](https://github.com/radare/radare2/files/2453964/f15985a91b12e015812f7f960eb0abe03d0cd401f54d0088f75cb087776a297c.zip) (password: "infected")

Other tools for comparison:

image

image

image

kazarmy:
Fixes #11759.

### DIFF
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -767,12 +767,9 @@ static int core_anal_fcn(RCore *core, ut64 at, ut64 from, int reftype, int depth
 						fcnpfx = r_config_get (core->config, "anal.fcnprefix");
 					}
 					fcn->name = r_str_newf ("%s.%08"PFMT64x, fcnpfx, fcn->addr);
-					/* Autoname PE import thunk */
-					RBinInfo *info;
+					/* Import trampoline naming */
 					if (r_list_length (fcn->bbs) == 1
-					    && ((RAnalBlock *) r_list_first (fcn->bbs))->ninstr == 1
-					    && (info = r_bin_get_info (core->bin))
-					    && !strcmp (info->rclass, "pe")) {
+					    && ((RAnalBlock *) r_list_first (fcn->bbs))->ninstr == 1) {
 						RList *refs = r_anal_fcn_get_refs (core->anal, fcn);
 						if (refs && r_list_length (refs) == 1) {
 							RAnalRef *ref = r_list_first (refs);
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -767,9 +767,12 @@ static int core_anal_fcn(RCore *core, ut64 at, ut64 from, int reftype, int depth
 						fcnpfx = r_config_get (core->config, "anal.fcnprefix");
 					}
 					fcn->name = r_str_newf ("%s.%08"PFMT64x, fcnpfx, fcn->addr);
-					/* Import trampoline naming */
+					/* Autoname PE import thunk */
+					RBinInfo *info;
 					if (r_list_length (fcn->bbs) == 1
-					    && ((RAnalBlock *) r_list_first (fcn->bbs))->ninstr == 1) {
+					    && ((RAnalBlock *) r_list_first (fcn->bbs))->ninstr == 1
+					    && (info = r_bin_get_info (core->bin))
+					    && !strcmp (info->rclass, "pe")) {
 						RList *refs = r_anal_fcn_get_refs (core->anal, fcn);
 						if (refs && r_list_length (refs) == 1) {
 							RAnalRef *ref = r_list_first (refs);
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -767,6 +767,19 @@ static int core_anal_fcn(RCore *core, ut64 at, ut64 from, int reftype, int depth
 						fcnpfx = r_config_get (core->config, "anal.fcnprefix");
 					}
 					fcn->name = r_str_newf ("%s.%08"PFMT64x, fcnpfx, fcn->addr);
+					/* Import trampoline naming */
+					if (r_list_length (fcn->bbs) == 1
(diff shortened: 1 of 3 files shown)


radare:
i dont understand this failing test, also pls rebase

screenshot 2019-02-16 at 12 23 47

kazarmy:
> i dont understand this failing test, also pls rebase

the code was renaming a `loc.` function that was calling a noreturn function ... fixed now

libr/core/cmd_anal.c
```diff
@@ -7715,7 +7715,8 @@ static int cmd_anal_all(RCore *core, const char *input) {
 					goto jacuzzi;
 				}
 				if (r_config_get_i (core->config, "anal.autoname")) {
-					oldstr = r_print_rowlog (core->print, "Constructing a function name for fcn.* and sym.func.* functions (aan)");
+					oldstr = r_print_rowlog (core->print, "Speculatively constructing a function name "
+					                         "for fcn.* and sym.func.* functions (aan)");
 					r_core_anal_autoname_all_fcns (core);
```
ITAYC0HEN: I don't think it should be under "aan" because "aan" might have a use and has its own logic to auto-name functions by used flags. I believe "aan" should be an independent command outside "aaa" and analyze the jmp-thunk-table inside "aaa" regardless of "anal.autoname".

libr/core/canal.c
```diff
@@ -767,6 +767,21 @@ static int core_anal_fcn(RCore *core, ut64 at, ut64 from, int reftype, int depth
 						fcnpfx = r_config_get (core->config, "anal.fcnprefix");
 					}
 					fcn->name = r_str_newf ("%s.%08"PFMT64x, fcnpfx, fcn->addr);
+					/* Import trampoline naming */
+					if (r_list_length (fcn->bbs) == 1
```
ITAYC0HEN: too many nested `ifs` imo

ITAYC0HEN: is this necessary? Can't check whether the function has a single JMP instruction and that's it? number of instructions of a function should be available to you

ITAYC0HEN:
We have two Pull Requests that trying to solve this and I think none of them currently defined the jump-thunks structure. While this PR aims to detect wrappers, it doesn't necessarily detect jump-thunks created by the linker. Also, I'd keep "aan" as is, but remove it from "aaa" immediately. The jmp-thunks detection should be available under "aa"\"aaa" by default indeed, but not as part of "aan" which some people would find useful as is.

So it should be:
1. Binary is a PE
2. The imported function is referenced only once
3. The reference should be a JMP instruction to a dereferenced import
4. The reference should be the only instruction in the referencing function

```
if imported_function (fcn):
  for ref in references_to (fcn):
    if is_instruction_jmp_with_deref_of_imported_name (ref):
      make_name_the_same_as_imported_one_at (ref)
```

radare:
Yes i dint plan to kill aan. Just set anal.autoname to false by default

kazarmy:
> I think none of them currently defined the jump-thunks structure

What does this mean exactly?

> it doesn't necessarily detect jump-thunks created by the linker

Afaik, for PE the `sym.imp.` flags do correspond directly to import table entries as detected by the PE parser.

> 1. Binary is a PE

Ok. In r2, apparently ELF binaries have their `sym.imp.` flags marking their import thunks. Maybe we should follow the ELF way? Idk whether this is possible or easy though.

> `if imported_function (fcn):`

How does one determine whether a function is an imported_function or not?

ITAYC0HEN:
> What does this mean exactly?
The problems raised in the issue is the lack of detection of jmp-thunks. Which are what you called "wrappers" of a PE. This is kind of a "table" which includes JMPs to the Imported APIs. There is one "Table" for each DLL. This is Created by the linker to avoid filling tons of relocations everywhere. The structure is indeed functions that only contains JMP to a dereferenced API function in the IAT. I am not familiar with this in ELF.

When I said "defined" I meant that your conditions are not strictly limited for JMP-Thunks. I characterized the jmp-thunks later in my comment.

> Afaik, for PE the `sym.imp.` flags do correspond directly to import table entries as detected by the PE parser.

very true; didn't claim otherwise That said, jmp thunks arent sym.imp, but the JMPs to the sym.imp functions.

> How does one determine whether a function is an imported_function or not?
maybe if its flag is part of the "imports" flagspace

kazarmy:
I see that #13119 finds cross-references to flags in the `imports` flagspace. Should work just as well, I think, if not better, though I have no idea about the failing tests.

kazarmy:
Oh #13119 doesn't check whether the cross-reference is from a JMP instruction though.

ITAYC0HEN:
true :) doesn't check if it drefs as well neither num of instructions

radare:
Your solution seems more generic, not tied to PE. lgtm

libr/core/cmd_anal.c
radare: well. i think that this is actually an autonaming hack, what we should probably do is convert anal.autoname into 'int' instead of 'bool' and allow to define the level of aggressivity on autonaming things

kazarmy:
So it's mergeable now?

libr/core/cmd_anal.c
kazarmy: agree, though the process of defining the levels of aggressivity might be a political nightmare

PR #13125, opened 2019-02-16
Fix a parsing issue of the config ##panels

Vane11ope:


Vane11ope:
this also fixes a bug in sdb_json_get_bool too

PR #13126, opened 2019-02-16
Time to say goodbye to r_json

Vane11ope:
It's been quite a journey but now is good time to say goodbye to r_json once and for all.


radare:
Woah was not dxpecting this :)) killing duped apis is always good news

radare:
@wargio feels sad to remove such working code, what about moving it into its own separate repository for being reused by other people?

XVilka:
It is git - in case of emergency you can always break a glass and do git
checkout older revision.

Maijin:
+ the syntax for using r_json is awful with those: free(XXX(JSON(XXX(YYY)

image

Issue #13079, opened 2019-02-13
Folded function comments are shown below instead of above


radare:
That makes the disasm pretty confusing.

<img width="713" alt="screenshot 2019-02-13 at 14 28 31" src="https://user-images.githubusercontent.com/917142/52715021-abc8a300-2f9b-11e9-9348-49b8501c11b8.png">
<img width="654" alt="screenshot 2019-02-13 at 14 28 38" src="https://user-images.githubusercontent.com/917142/52715022-abc8a300-2f9b-11e9-84cc-60722da255aa.png">

radare:
this is not an RFE, its an actual regression and very anoying one if you use folded functions

PR #13128, opened 2019-02-16
Folded function comments are shown below instead of above #13079

rfc2119:
Why do we print comments if folded (not yet tested with r2r) ?


radare:
I think primting the comments of the first instruction of a folded function makes sense

radare:
and yep we need some tests for this at some point.

Issue #13072, opened 2019-02-13
Bad integer conversion in dalvik_disassemble


t00sh:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Debian 4.9.130-2, AMD 64
| File format of the file you reverse (mandatory)      | Dalvik dex file version 035
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.3.0-git 20837 @ linux-x86-64 git.3.2.1-324-gfb2682a16 commit: fb2682a16f9ab46eb32e79cdd5c8835cf76ae59e build: 2019-02-12__21:11:44

### Actual behavior

There is a bad integer conversion in `dalvik_disassemble` which can lead to memory corruption and segfault.

### Steps to reproduce the behavior 

- Open the  [binary](https://github.com/radare/radare2/files/2858237/dalvik-segfault.zip) in radare2
- Run `aa` and `pdf`

### Additional Logs, screenshots, source-code,  configuration dump, ...

ASan trace : 

```
==18593==ERROR: AddressSanitizer: SEGV on unknown address 0x613fd800e7b4 (pc 0x7ffff3ad7e00 bp 0x613fd800e7b4 sp 0x7fffffffc4c0 T0)

    #0 0x7ffff3ad7dff in dalvik_op /XXX/radare2-src-asan/libr/..//libr/anal/p/anal_dalvik.c:48
    #1 0x7ffff3cfcf21 in r_anal_op /XXX/radare2-src-asan/libr/anal/op.c:150
    #2 0x7ffff675dca2 in r_core_print_disasm /XXX/radare2-src-asan/libr/core/disasm.c:4842
    #3 0x7ffff64f69f6 in cmd_print /XXX/radare2-src-asan/libr/core/cmd_print.c:4885
```
The problem is a bit difficult to trace, but I think the main bug is in the `dalvik_disassemble` function : 

```C
                case 0x03: /* fill-array-data-payload */
                        // element_width = 2 bytes ushort little endian                                                               
                        // size = 4 bytes uint                                                                                        
                        // ([size*element_width+1)/2)+4                                                                               
                        if (len > 7) {
                                ut16 elem_width = buf[2] | (buf[3] << 8);
                                ut32 array_size = buf[4] | (buf[5] << 8) | (buf[6] << 16) | (buf[7] << 24);
[...]

PR #13133, opened 2019-02-17
Try a fix for #13072

t00sh:
I tried to add bound checks to avoid #13072. 

I don't really know if they are well placed. And the check in `dalvik_op` is to avoid a read out-of bounds, but I don't know if it's correct.


radare:
looks green for travis

PR #13135, opened 2019-02-17
WIP: Resolve pascal strings ##disasm

radare:


libr/core/disasm.c
```diff
@@ -3510,7 +3510,7 @@ static void ds_print_ptr(RDisasmState *ds, int len, int idx) {
 		if (ds->immstr) {
 			char *str = r_str_from_ut64 (r_read_ble64 (&v, core->print->big_endian));
 			if (str && *str) {
-				const char *ptr = str;
+				const char *ptr = str + 1;
```
kazarmy: First char is skipped ... why?

libr/core/disasm.c
```diff
@@ -3747,12 +3747,18 @@ static void ds_print_ptr(RDisasmState *ds, int len, int idx) {
...
+				m++;
+				l--;
+			}
+			kind = r_anal_data_kind (core->anal, refaddr, (const ut8*)m, l);
 			if (kind) {
 				if (!strcmp (kind, "text")) {
-					if (!string_printed && *msg) {
-						ds_print_str (ds, msg, len, refaddr);
+					if (!string_printed && *m) {
+						ds_print_str (ds, m, l, refaddr);
```
kazarmy: If non-printable chars are skipped, there should be some kind of visual indicator, e.g. `.."str"`

libr/core/disasm.c
radare: Because thats how pascal strings work. Dirst char is the length of the string and then theres the string

libr/core/disasm.c
radare: Those chars are not part of the string its jusy the length

libr/core/disasm.c
kazarmy: Only 1 byte for the length? If the length is 0xa (same value as newline), it's going to cause broken output.

libr/core/disasm.c
kazarmy: Ok but might cause issues if people expect the string to be at addr and not at addr+n

radare:
n shuold be bigger than 1 byte, but yes, those failing tests are there because of false positives, thats why we need a better way to dettermine this by using a config var set depending on the filetype or doing better checks..like we do for cfstrings and other type of strings

XVilka:
Are there any tests for this? You can reuse these https://github.com/radareorg/radeco-regressions/tree/master/freepascal and https://github.com/radareorg/radeco-regressions/tree/master/ada from radeco test suite.

XVilka:
Can you please add test and rebase?

radare:
@kazarmy can you have a look at this issue? im improving dalvik support, but busy in other places

kazarmy:
i've found this: https://source.android.com/devices/tech/dalvik/dex-format#mutf-8 ... where does it say that dalvik strings are in pascal format?

radare:
Closing for https://github.com/radare/radare2/pull/14525

radare:
Its pascal style but length is in uleb128, and yes,bthis modified utf8 thing is a mess too

kazarmy:
Oh you mean [string_data_item](https://source.android.com/devices/tech/dalvik/dex-format#string-data-item). It stores strings in mutf-8, which is very similar to utf-8.

Regarding mutf-8:

- "Code points in the range U+10000 … U+10ffff are encoded as a surrogate pair, each of which is represented as a three-byte encoded value."
  - `r_utf8_decode()` doesn't treat surrogate pairs any differently. They might be printed wrongly I suppose.
- "The final two items above make it simultaneously possible to include the code point U+0000 in a string and still manipulate it as a C-style null-terminated string."
  - Apart from almost being a contradiction, how many strings actually have embedded '`\0`'s in them??

PR #13137, opened 2019-02-17
Fix cursor scroll

radare:


radare:
This fix requires a better solution, but at least the cursor doesn't goes away and the interaction is not that painful. There are more issues to solve for folded functions

PR #13138, opened 2019-02-18
make function to allow grabbing the flags by order of importance in flagspace

lowlyw:
…lagspaces

r2r PR: https://github.com/radare/radare2-regressions/pull/1719


lowlyw:
seems like i have the wrong order in some of the use cases here....

i'm happy with the function, but maybe not with the usage in the filter funtion

radare:
screenshot 2019-02-18 at 06 22 40

libr/flag/flag.c
```diff
@@ -502,6 +502,36 @@ R_API RFlagItem *r_flag_get_i(RFlag *f, ut64 off) {
...
+	RFlagItem *flg = NULL;
+	RListIter *iter;
+	RSpace *myspace;
+	const char *spacename;
+	int i;
+	va_list ap;
+	va_start (ap, nspaces);
+	for (i = 0; i < nspaces; i++) {
+		spacename = va_arg (ap, const char *);
+		myspace = r_spaces_get (&(f->spaces), spacename);
```
ret2libc: use `r_flag_space_get (RFlag *, const char *name)`

libr/flag/flag.c
```diff
@@ -502,6 +502,36 @@ R_API RFlagItem *r_flag_get_i(RFlag *f, ut64 off) {
...
+		if (myspace) {
+			const RList *list = r_flag_get_list (f, off);
```
ret2libc: this does not need to be called on each loop, it can be done outside the outer loop.

libr/flag/flag.c
```diff
@@ -502,6 +502,36 @@ R_API RFlagItem *r_flag_get_i(RFlag *f, ut64 off) {
 	return list ? evalFlag (f, r_list_get_top (list)) : NULL;
 }
 
+/* return the first flag that matches an offset ordered by the order of
+ * operands to the function.
+ * XXX this is kind of super ugly! */
+R_API RFlagItem *r_flag_get_by_spaces(RFlag *f, ut64 off, int nspaces, ...) {
```
ret2libc: I don't know if it would be premature optimization, but I think we can make it much more efficient since it is basically a "compute max of a list of elements", where "max" however is not just a number but depends on the flagspace.

```
htspace = ht_uu_new0 ();
for i..nspaces
   space = r_flag_space_get(f, spacename);
   ht_uu_insert (htspace, (ut64)(uintptr_t)(space), i)

list = r_flag_get_list (f, off);
max_fi = Null
max = -1
for fi in list:
   ut64 pos = ht_uu_find (htspace, (ut64)(uintptr_t)fi->space, NULL)
   if pos > max:
       max = pos
       max_fi = fi

ht_uu_free (htspace)
return max_fi
```

libr/flag/flag.c
```diff
@@ -502,6 +502,36 @@ R_API RFlagItem *r_flag_get_i(RFlag *f, ut64 off) {
 	return list ? evalFlag (f, r_list_get_top (list)) : NULL;
 }
 
+/* return the first flag that matches an offset ordered by the order of
+ * operands to the function.
+ * XXX this is kind of super ugly! */
```
ret2libc: what's exactly the reason?

libr/include/r_flag.h
```diff
@@ -103,6 +103,7 @@ R_API void r_flag_list(RFlag *f, int rad, const char *pfx);
 R_API bool r_flag_exist_at(RFlag *f, const char *flag_prefix, ut16 fp_size, ut64 off);
 R_API RFlagItem *r_flag_get(RFlag *f, const char *name);
 R_API RFlagItem *r_flag_get_i(RFlag *f, ut64 off);
+R_API RFlagItem *r_flag_get_by_spaces(RFlag *f, ut64 off, int nspaces, ...);
```
ret2libc: we could remove `nspaces` by forcing the user to add a `NULL` at the end of the spaces, what do you think?

libr/parse/parse.c
```diff
@@ -262,14 +262,22 @@ static int filter(RParse *p, ut64 addr, RFlag *f, char *data, char *str, int len
 			}
 			if (f) {
 				RFlagItem *flag2;
-				flag = r_flag_get_i2 (f, off);
+				flag = r_flag_get_by_spaces (f, off, 4, "symbols", "imports",
+						"relocs", "strings");
+				if (!flag) {
+					flag = r_flag_get_i2 (f, off);
```
ret2libc: i believe this shouldn't be needed anymore if `r_flag_get_by_spaces` works well, should it?

libr/parse/parse.c
```diff
@@ -262,14 +262,22 @@ static int filter(RParse *p, ut64 addr, RFlag *f, char *data, char *str, int len
...
 				if (!flag) {
 					flag = r_flag_get_i (f, off);
 				}
 				if ((!flag || arm) && p->relsub_addr) {
 					computed = true;
-					flag2 = r_flag_get_i2 (f, p->relsub_addr);
+					flag2 = r_flag_get_by_spaces (f, p->relsub_addr, 4, "symbols",
+							"imports", "relocs", "strings");
+					if (!flag2) {
+						flag2 = r_flag_get_i2 (f, p->relsub_addr);
```
ret2libc: same: i believe this shouldn't be needed anymore if `r_flag_get_by_spaces` works well, should it?

libr/parse/parse.c
ret2libc: what about this? are there reasons to keep it?

libr/flag/flag.c
ret2libc: do you still think it's ugly? :) if so, please tell us we can try to find some other solutions!

libr/parse/parse.c
```diff
@@ -262,17 +262,13 @@ static int filter(RParse *p, ut64 addr, RFlag *f, char *data, char *str, int len
 			}
 			if (f) {
 				RFlagItem *flag2;
-				flag = r_flag_get_i2 (f, off);
+				flag = r_flag_get_by_spaces (f, off, "symbols", "imports",
```
ret2libc: i think you forgot `"functions"` before everything else.

ret2libc:
Some of the XX seem to be caused by the fact `_i2` was excluding flags shorter than 3 chars and that didn't have a `.`. However... I'm not sure that's the "right" way to do it.

ret2libc:
In particular the errors in the tests seem to be caused by `$a`, `$t`, `$d` symbols used in ARM binaries to identify data/thumb/regulard code. Those symbols are not showed by objdump and not even by IDA. objdump consider those symbols "--special-syms", which is an option that says `When displaying symbols include those which the target considers to be special in some way and which would not normally be of interest to the user. `. 

See also https://sourceware.org/binutils/docs-2.23/as/ARM-Mapping-Symbols.html

I think we should filter out those symbols as well or, maybe better, put them in a "special_symbols" flag space.

@radare what do you think?

radare:
Yes those $ symbols are important for hints to identify begining of functions, thumb/arm, dword types, etc, we should import this info as different types of symbols, maybe not setting flags for them but just using Cd etc

ret2libc:
TODO: filter out $t,$a,$d from ARM ELF symbols, because they are special symbols and should not be shown in the disassembly
TODO: use this new API instead of _i2 and see how it goes.

ret2libc:
See https://github.com/radare/radare2/pull/13173 for the ARM special symbol issues.

ret2libc:
See https://github.com/radare/radare2/pull/13175 to have a way to specify common flagspaces without hardcoding all strings everywhere.

ret2libc:
See https://github.com/radare/radare2/pull/13182 to split symbols in symbols, symbols.objects and symbols.sections flagspaces. This will allow us to get flags by a more fine-grained priorities (e.g. section symbols have less priority than regular symbols, while object symbols have higher priority, etc.)

libr/parse/parse.c
radare: yep. can you add this?  i'll merge after that

ret2libc:
@radare wait I’m still working on this.

radare:
can you rebase?

ret2libc:
Yes, I will tomorrow but as said I’m still working on this.

radare:
Ok

ret2libc:
@lowlyw @radare I think this is ready to be reviewed for good. Please tell me what you think.

Main changes with this PR: eliminate `r_flag_get_i2` which was messy and hardly configurable in favour of `r_flag_get_by_spaces` (thanks @lowlyw), which allows to get the first flagitem sorted by a configurable list of spaces. Moreover, it moves all code that handles flagspaces inside RCore, which is the only one module that should decide the names of flag spaces (see `r_core_flag_get_by_spaces`).

There are also tests attached https://github.com/radare/radare2-regressions/pull/1719/files. It seems the situation is slightly improved with the `s/sym.main/main` change.

radare:
can this be merged?

ret2libc:
If you think the change is ok, yes.

radare:
merged!

ret2libc:
thanks ;)

PR #13139, opened 2019-02-18
Implement self_regions for Windows

GustavoLCR:


libr/io/p/io_self.c
```diff
@@ -131,7 +132,32 @@ static int update_self_regions(RIO *io, int pid) {
...
 #ifdef _MSC_VER
-#pragma message ("Not yet implemented for this platform")
+	int perm;
+	MEMORY_BASIC_INFORMATION mbi;
+	HANDLE h = OpenProcess (PROCESS_QUERY_INFORMATION | PROCESS_VM_READ, 0, pid);
+	size_t size = VirtualQuery (0, &mbi, sizeof (mbi));
+	char *name = calloc (1024, sizeof (char));
+	while (size) {
+		ut64 to = (ut8 *) mbi.BaseAddress + mbi.RegionSize;
+		GetMappedFileName (h, (LPVOID) mbi.BaseAddress, name, 1024);
```
radare: check return value, otherwise name can be not initialized. also are we sure that this function null terminates the buffer always? otherwise we should memset instead of just name[0]=0

libr/io/p/io_self.c
```diff
@@ -131,7 +132,32 @@ static int update_self_regions(RIO *io, int pid) {
 	return bsd_proc_vmmaps(io, pid);
 #else
 #ifdef _MSC_VER
-#pragma message ("Not yet implemented for this platform")
+	int perm;
+	MEMORY_BASIC_INFORMATION mbi;
+	HANDLE h = OpenProcess (PROCESS_QUERY_INFORMATION | PROCESS_VM_READ, 0, pid);
+	size_t size = VirtualQuery (0, &mbi, sizeof (mbi));
+	char *name = calloc (1024, sizeof (char));
```
radare: check if allocation fails

radare: use a const int name_size = 1024; instead of hardcoding 1024 here and there

libr/util/lib.c
```diff
@@ -61,15 +61,20 @@ R_API void *r_lib_dl_open(const char *libname) {
...
 #endif
-	if (!libname || !*libname) {
-		return NULL;
-	}
 #if __WINDOWS__
 	LPTSTR libname_;
 
-	libname_ = r_sys_conv_utf8_to_utf16 (libname);
+	if (!libname || !*libname) {
+		libname_ = malloc (MAX_PATH);
```
radare: check allocation

libr/util/lib.c
```diff
@@ -61,15 +61,20 @@ R_API void *r_lib_dl_open(const char *libname) {
...
+		GetModuleFileName (NULL, libname_, MAX_PATH);
```
radare: check return value, null terminate buffer

libr/util/lib.c
radare: use calloc to ensure null terminated string?

GustavoLCR:
Those functions always terminates the string with a '\0' even if the string gets truncated when there is not enough space in the buffer (not considered a fail), so with these changes it should be fine.

libr/io/p/io_self.c
GustavoLCR: see comment below

PR #13142, opened 2019-02-18
Check if the unset space is the current one

ret2libc:


ret2libc:
Fixes https://github.com/radare/radare2/issues/13141#event-2146619779

PR #13149, opened 2019-02-18
Expose save_parsed_type() as an API call for Cutter

m0nk3ydluffy:


libr/core/cmd_type.c
```diff
@@ -383,7 +383,7 @@ static void save_parsed_type_size(RCore *core, const char *parsed) {
 	}
 }
 
-static void save_parsed_type(RCore *core, const char *parsed) {
+R_API void save_parsed_type(RCore *core, const char *parsed) {
```
radare: this function name cant be public. rename it to something meaningful like r_core_something, but this looks like a completely random function name to make it public. and also, this function is only used from the same file, so there's no need to make it public

libr/core/cmd_type.c
```diff
@@ -383,7 +383,7 @@ static void save_parsed_type_size(RCore *core, const char *parsed) {
...
 	if (!core || !core->anal || !parsed) {
```
radare: this shoulkd be r_return_val_if_fail

libr/include/r_core.h
```diff
@@ -497,6 +497,7 @@ R_API RList *r_core_get_func_args(RCore *core, const char *func_name);
 R_API void r_core_print_func_args(RCore *core);
 R_API char *resolve_fcn_name(RAnal *anal, const char * func_name);
 
+R_API void save_parsed_type(RCore *core, const char *parsed);
```
radare: if you need it from cutter. find a good name, same goes for that shitty resolve_fcn_name defined above

m0nk3ydluffy: resolve_fcn_name was the reason I though r_core_ prefix was not required. Sorry for that. Will rename to r_core_save_parsed_type

radare:
please update it now and i'll merge, i have a regression to fix in this exactly code and i would like to avoid conflicts

m0nk3ydluffy:
I changed the name to your liking. Done.

PR #13151, opened 2019-02-18
Add missing types-darwin.sdb and honor asm.os to reload them ##anal

radare:


libr/anal/d/types-darwin.sdb.txt
```diff
@@ -0,0 +1,159 @@
...
+_Exit=func
+func._Exit.args=1
+func._Exit.arg.0=int,status
+func._Exit.noreturn=true
+func._Exit.ret=void
+
+__libc_init=func
+func.__libc_init.args=3
+func.__libc_init.arg.0=int,argc
+func.__libc_init.arg.1=char **,argv
```
pelijah: So is this the right way to define pointer-arguments? Should we use it everywhere? For example, `types-windows.sdb.txt` contains such entries as `argtype*,argname` and `argtype,*argname`.

Maijin: @sivaramaaa would know

PR #13153, opened 2019-02-19
Compile in release mode when testing

ret2libc:


Maijin:
wow check the codecov

ret2libc:
@Maijin what about codecov?

Maijin:
Just interesting to see the huge change

ret2libc:
Mmm maybe it's better to run with -O0/debug symbols when using covscan

ret2libc:
This is not useful at all... Closing.

PR #13154, opened 2019-02-19
X86_INS_UD0 was introduced in capstone 4, so use it only there

ret2libc:


ret2libc:
This is necessary to build package on Fedora, because Fedora 29 still shipes capstone 3.0.5

PR #13158, opened 2019-02-19
Fix offline build w.r.t. capstone

Mic92:
If the user already provides a capstone tarball, the buildsystem should not try
to download it again.

### DIFF
--- a/shlr/Makefile
+++ b/shlr/Makefile
@@ -255,7 +255,6 @@ capstone-clean:
 	cd capstone ; $(MAKE) clean
 
 capstone: capstone-$(CS_VER).tar.gz
-	curl -o capstone-$(CS_VER).tar.gz $(CS_TAR)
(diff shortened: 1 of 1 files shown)


Mic92:
@radare can you clarify, why second curl is necessary? I don't understand your comment in https://github.com/radare/radare2/commit/f3bf16555fec45863a0cd63aeada89ecc250c39a#r32374990

Mic92:
Is it because a download might fail and leave garbage behind?

Mic92:
The current code does `wget` or fallback to `curl` if `wget` is not available:

```make
ifeq ($(shell type wget && echo 1),1)
WGET?=wget -c --no-check-certificate -O
else
WGET?=curl -o
endif
```

So it already uses `curl` and the second curl is not necessary.

radare:
Ok

Issue #9286, opened 2018-02-01
RAP Protocol Question


digitalstreamio:
I am trying to implement a RAP server within my project (c64 6502 emulator) to enable remote debugging. So far I have implemented Open/Read/Write/Seek/Close/Cmd operations and they seem to work fine.

However, I am struggling to figure out how to properly implement handling of Cmd operations that are issued by r2 debug module. For starters I have looked at handling of "drp" and "dr".

### DRP - debug register profile

currently i have it returning the following string (register profile for 6502 cpu):

```
=PC	pc
=SP	sp
gpr	a	.8	0	0
gpr	x	.8	1	0
gpr	y	.8	2	0
gpr	flags	.8	3	0
gpr	C	.1	.24	0
gpr	Z	.1	.25	0
gpr	I	.1	.26	0
gpr	D	.1	.27	0
gpr	V	.1	.30	0
gpr	N	.1	.31	0
gpr	sp	.8	4	0
gpr	pc	.16	5	0
```

### DR - debug registers

Option 1: string

```
a = 0x00
x = 0x00
y = 0x00
flags = 0x00
sp = 0x00
pc = 0x0000
```

Option 2: byte array w/ 7 elements that correspond to register values with offsets listed in DRP.

[ x x x x x x x ]

### Question 1

What is the format that r2 expects for 'dr' command? I tried the two options listed above but neither one seems to work.

I also tried tracing through the call stack for dr command:
```

libr/core/cmd_debug.c:  cmd_debug_reg()
libr/debug/dreg.c: r_debug_reg_sync()
libr/debug/p/debug_rap.c: r_debug_rap_reg_read()
libr/io/io.c: r_io_system()
libr/io/p/io_rap.c: rap__system()
```
but  I don't see how data received from my server in rap__system() call is actually copied into buffer set up by r_debug_reg_sync().
```
R_API int r_debug_reg_sync(RDebug *dbg, int type, int write) {
                ...
		} else {
			// int bufsize = R_MAX (1024, dbg->reg->size*2); // i know. its hacky
			int bufsize = dbg->reg->size;
			//int bufsize = dbg->reg->regset[i].arena->size;
			if (bufsize > 0) {
				ut8 *buf = calloc (1 + 1, bufsize);
				if (!buf) {
					return false;
				}
				//we have already checked dbg->h and dbg->h->reg_read above
				size = dbg->h->reg_read (dbg, i, buf, bufsize);
				// we need to check against zero because reg_read can return false
[...]

PR #13159, opened 2019-02-19
Fix debug rap reg profile setup

ScrimpyCat:
The operation was not being flushed prior to reading it, causing the `r_file_slurp` to not return anything.

This is part of the fix addressing some of the issues discussed in #9286 But it still is not enough to allow for remotely setting of the reg profile, due to it being replaced later (see my last comment in that issue).

So just thought I'd submit the PR now, as I'm not sure how to resolve the rest. If I can get some details about it I'd be happy to try and fix the remaining issue.


radare:
uhm, shouldnt this flushing happen inside the pipe_close?

ScrimpyCat:
Honestly I'm not sure, but adding the flush means `out = r_file_slurp (tf, NULL);` will actually contain the string (register profile) a server replies back with. Whereas without, it's just empty. Maybe the problem lies elsewhere?

This isn't a full solution to the problem though, this fix does allow for the register profile to be retrieved and then passed on so it's actually set. However afterwards the profile is being cleared again by the `r_config_set` (see the issue #9286 for more details regarding that). 

Basically in my use case I have a game I'm working on that I'm adding a RAP server implementation to ([WIP server](https://github.com/ScrimpyCat/Hacking-Game/blob/master/src/rap/RapServer.c)), as I thought I could leverage the RAP protocol (instead of adding a debug plugin) to allow for radare to debug programs in the game (I've added an [asm plugin](https://github.com/radare/radare2/compare/master...ScrimpyCat:master) to my own fork; I'm not going to try get any of this specific stuff into upstream though as it's just a toy/not of much use to anyone beyond the context of the game). If there's things that need to be fixed up with the RAP implementation in order to achieve working with custom RAP servers, then I'm interested in helping with that (though I'd need some pointers, since beyond this change I'm not sure how to get the rest to work). Or is there no intention to use the RAP protocol in this way (outside of radare to radare communication)?

radare:
Yes i understand the issue but i was suggesting to move that call info the pipe close but i think its fine right now, we may probably find other places where pipe_open is misused like this?

ScrimpyCat:
My bad didn't realise that's what you meant, would that have been more appropriate? Looking where else the `r_cons_pipe_open` is called, the only other file is https://github.com/radare/radare2/blob/73f76a99b29ea98fd286376b2fd470aa55e22666/libr/core/cmd.c but all of those pipes are being flushed before closing.

radare:
Then its finw but its a common mistake that we can avoid by moving the flush in there. What do you think?

ScrimpyCat:
That certainly would make it less error prone, so I think it could be a good idea. But I guess it just depends on whether you want the interface to closer resemble the regular close function (which isn't guaranteed to flush, but some devs might be familiar with close methods on other systems or languages which do flush).

PR #13161, opened 2019-02-19
Make Lisp fortune work with Clojure(Script)

wasamasa:
I've noticed that this fun one-liner looks very much like Clojure code, but doesn't actually work, so I took the liberty to fix it. It also works in ClojureScript as tested with lumo.


radare:
It was newlisp i think

wasamasa:
Interesting, I didn't think of this. However, the newlisp interpreter fails executing this.

PR #13162, opened 2019-02-19
Improve CHIP-8 analysis

wasamasa:
- Default to unknown type
- Detect RET
- Remove fail from always succeeding jump/call
- Move keypad commentary to LD Vx, K
- Annotate skips correctly
- Annotate more instructions

One thing I'm not sure about is the "JP V0, nnn" instruction. It jumps to the value of the embedded address, incremented by the contents of register `V0`. Therefore it's incorrect to annotate it to jump to `nnn` because the address might be anywhere between `nnn` and `nnn + V0`. Maybe it's better to not annotate it at all? Or is there support for annotating ranges?

### DIFF
--- a/libr/anal/p/anal_chip8.c
+++ b/libr/anal/p/anal_chip8.c
@@ -16,48 +16,38 @@ static int chip8_anop(RAnal *anal, RAnalOp *op, ut64 addr, const ut8 *data, int 
 	uint8_t kk = opcode & 0xFF;
 	op->size = 2;
 	op->addr = addr;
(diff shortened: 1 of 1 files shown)


radare:
can i merge this? you can continue improving chip8 support in other PRs

wasamasa:
Sure, I just hoped to get an answer with regards to jump instructions that have no accurate target, but more of a target range.

radare:
if i understood you correctly, the only way to do that is at runtime, because you need to know the vlaue of the register to perform that, right? if so, esil is the only way to go here

PR #13166, opened 2019-02-21
ASLR settings little refactor, available as util mean.

devnexen:


libr/util/sys.c
```diff
@@ -460,6 +460,37 @@ R_API int r_sys_chdir(const char *s) {
 	return r_sandbox_chdir (s)==0;
 }
 
+R_API bool r_sys_aslr(int val) {
+	bool ret = true;
+#if __linux__
+	static const char *rva = "/proc/sys/kernel/randomize_va_space";
+	char buf[2] = {0};
+	snprintf(buf, sizeof (buf), "%d\n", val != 0 ? 2 : 0);
+	int fd = open (rva, O_WRONLY);
```
radare: use r_sandbox apis otherwise you will be bypassing the sandbox

libr/util/sys.c
```diff
@@ -460,6 +460,37 @@ R_API int r_sys_chdir(const char *s) {
 	return r_sandbox_chdir (s)==0;
 }
 
+R_API bool r_sys_aslr(int val) {
+	bool ret = true;
+#if __linux__
+	static const char *rva = "/proc/sys/kernel/randomize_va_space";
+	char buf[2] = {0};
```
radare: the string you are trying to set in here is 3 byte long, not 2, this is overflowing

libr/util/sys.c
```diff
@@ -460,6 +460,37 @@ R_API int r_sys_chdir(const char *s) {
 	return r_sandbox_chdir (s)==0;
 }
 
+R_API bool r_sys_aslr(int val) {
+	bool ret = true;
+#if __linux__
+	static const char *rva = "/proc/sys/kernel/randomize_va_space";
```
radare: this string doesnt needs to be saved as static

libr/util/sys.c
```diff
@@ -460,6 +460,37 @@ R_API int r_sys_chdir(const char *s) {
...
+	char buf[2] = {0};
+	snprintf(buf, sizeof (buf), "%d\n", val != 0 ? 2 : 0);
```
radare: missing space before (

libr/util/sys.c
```diff
@@ -460,6 +460,37 @@ R_API int r_sys_chdir(const char *s) {
...
+	int fd = open (rva, O_WRONLY);
+	if (fd != -1) {
+		if (write (fd, buf, sizeof (buf)) != 2) {
```
radare: shoulkd be 3, and compare with sizeof(buf) not against this hardcoded 2

devnexen: I merely move the original code ... I was not treating as a C string but just to write to the file.

libr/util/sys.c
devnexen: Yes good point.

libr/util/sys.c
devnexen: Originally we write two bytes to the file but 3 works too ...

PR #13169, opened 2019-02-21
On ARM, labelling the unwind address segment a bit more accurately

devnexen:


libr/bin/p/bin_elf.inc
```diff
@@ -250,7 +262,7 @@ static RList* sections(RBinFile *bf) {
 				ptr->name = strdup ("NOTE");
 				break;
 			default:
-				ptr->name = strdup ("UNKNOWN");
+				ptr->name = setphname(mach, phdr[i].p_type);
```
radare: Missing space before (

libr/bin/p/bin_elf.inc
```diff
@@ -57,6 +57,17 @@ static void setimpord(ELFOBJ* eobj, ut32 ord, RBinImport *ptr) {
 	eobj->imports_by_ord[ord] = r_bin_import_clone (ptr);
 }
 
+static char *setphname(ut16 mach, Elf_(Word) ptyp) {
+	// TODO to complete over time
+	if (mach == EM_ARM) {
+		if (ptyp == SHT_ARM_EXIDX) {
+			return strdup ("EIXDX");
```
radare: We should evaluate the use of r_str_const to save some memory in here. But we can evolve the concept with a stringpool. Nothing to worry in this pr. But just to think about

libr/bin/p/bin_elf.inc
devnexen: again ... sorry I spend time jumping between projects requirements :)

PR #13173, opened 2019-02-22
Filter out special symbols in ELF

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1707


libr/bin/format/elf/elf.c
```diff
@@ -3210,6 +3210,31 @@ RBinSymbol *Elf_(_r_bin_elf_convert_symbol)(struct Elf_(r_bin_elf_obj_t) *bin,
...
+		return ELF_ST_TYPE (sym->st_info) == STT_NOTYPE &&
+			ELF_ST_BIND (sym->st_info) == STB_LOCAL &&
+			ELF_ST_VISIBILITY (sym->st_info) == STV_DEFAULT;
+	default:
+		return false;
+	}
+}
+
+static bool is_special_symbol(ELFOBJ *bin, Elf_(Sym) *sym, const char *name) {
+	switch (bin->ehdr.e_machine) {
```
radare: no need for a switch statement here, just an if is enough

ret2libc: I know but special symbols are by architecture so I have done this in preparation for future architectures. Compiler already compile to the best assembly anyway.

radare:
The a/t symbols indicate arm/thumb, so this is mainly a way to indicate a function starts there.. and we may make cbin.c spit out ahb lines.

The d is for data, so we can use Cd for them. So imho is better to use them instead of hiding them. but i agree those look anoying in the disasm

ret2libc:
I agree we should use them, but I think we should retrieve them in other ways and not convert them to RBinSymbols and RFlagItems. There should be a proper API for RBin to provide "help" to analysis or something like that.

radare:
We can just handle those special symbols as they are in the current cbin. No need to flag them. And u know that we can divide the loding time of rbin just skipping the flag creation? Thats what my pr was about to proof

ret2libc:
Ok so what about this:

I add a `bool is_special;` to `RBinSymbol` and each bin plugin will mark the right symbols as `is_special=true`. Then in cbin.c we do not create flags for those symbols. Does it seem better? That way we still have the symbols listed, but no flags for them.

radare:
Just use a differnet type no need to add a bool imho

ret2libc:
> Just use a differnet type no need to add a bool imho

Right, thanks.

libr/bin/format/elf/elf.c
radare: ive only seen this in thumb binaries, but its fair

radare:
also, having flags on where functions starts is useful for a variety of usecases, we i wont remove/filter those symbols

ret2libc:
IMHO they are not, because if you have those special symbols, you already have regular symbols which just overlap with them. If you don't have regular symbols, you most likely don't even have special symbols like `$t`, `$a`, etc. And I prefer seeing `my_function_name` instead of `$t` everywhere. So, that said, I don't think those flags add any value.

On the other hand, they do add a lot of value for anal hints and data definition, because you can set arm/thumb mode and where the data is. But that is not something that should be shown through flags, it should just be used by Cd.

Last thing, both objdump and IDA filter out those symbols.

ret2libc:
@radare you can review it again if you want! I've implemented what we said, so all those `$d`, `$t`, etc. are still visible through `is` but they are not in the flags.

I think it's a good improvement, though the best thing would be to use those info for hints and to define data.

libr/core/cbin.c
```diff
@@ -2048,6 +2053,9 @@ static int bin_symbols(RCore *r, int mode, ut64 laddr, int va, ut64 at, const ch
 		} else if (IS_MODE_SIMPLEST (mode)) {
 			const char *name = sn.demname? sn.demname: r_symbol_name;
 			r_cons_printf ("%s\n", name);
+		} else if (IS_MODE_RAD (mode) && is_special_symbol (symbol)) {
```
radare: Why chk twice just call is special inside the block below

Maijin:
I agree with ret2libc that those flags are very annoying as part of the disassembly. Having them in `is` or via analhint etc. sounds like a good compromise imho

libr/core/cbin.c
ret2libc: i did this to avoid a new level of indentation caused by the `if`, but... fixed it ;)

ret2libc:
@radare fixed. Any other concerns?

libr/bin/format/elf/elf.c
```diff
@@ -2817,6 +2817,31 @@ RBinElfSection* Elf_(r_bin_elf_get_sections)(ELFOBJ *bin) {
 	return ret;
 }
 
+static bool is_special_arm_symbol(ELFOBJ *bin, Elf_(Sym) *sym, const char *name) {
+	if (name[0] != '$') {
+		return false;
+	}
+	switch (name[1]) {
```
radare: Can we improve this check? i think its a bit weak that way, didnt checked whats in name+2, to avoid false positives

libr/bin/format/elf/elf.c
```diff
@@ -2817,6 +2817,31 @@ RBinElfSection* Elf_(r_bin_elf_get_sections)(ELFOBJ *bin) {
...
+	switch (bin->ehdr.e_machine) {
+	case EM_ARM:
+		return is_special_arm_symbol (bin, sym, name);
+	default:
+		return false;
+	}
+}
+
 static const char *bind2str(Elf_(Sym) *sym) {
 	switch (ELF_ST_BIND(sym->st_info)) {
```
radare: missing space before (

libr/core/cbin.c
```diff
@@ -2049,61 +2054,65 @@ static int bin_symbols(RCore *r, int mode, ut64 laddr, int va, ut64 at, const ch
...
+					} else {
+						// we dont want unnamed symbol flags
+					}
+				}
+				binfile = r_core_bin_cur (r);
+				plugin = r_bin_file_cur_plugin (binfile);
+				if (plugin && plugin->name) {
+					if (!strncmp (plugin->name, "pe", 2)) {
+						char *p, *module = strdup (r_symbol_name);
+						p = strstr (module, ".dll_");
```
radare: define and assign in the same line

libr/core/cbin.c
```diff
@@ -2049,61 +2054,65 @@ static int bin_symbols(RCore *r, int mode, ut64 laddr, int va, ut64 at, const ch
...
-				}
-				lastfs = 's';
-			}
-			if (r->bin->prefix) {
-				if (symbol->dup_count) {
-					r_cons_printf ("f %s.sym.%s_%d %u 0x%08"PFMT64x"\n",
-						r->bin->prefix, name, symbol->dup_count, symbol->size, addr);
+			/* Skip special symbols because we do not flag them and
+			 * they shouldn't be printed in the rad format either */
+			if (!is_special_symbol (symbol)) {
```
radare: is_special_symbol is checked in another conditional above. so this is doing nothing, i would rather remove the check above and keep it like this, but inverting the conditional to behave properly to set the ahb, Cd..

if (is_special_symbol(symbol)) {
  ...
} else {
...

libr/core/cbin.c
```diff
@@ -2049,61 +2054,65 @@ static int bin_symbols(RCore *r, int mode, ut64 laddr, int va, ut64 at, const ch
...
+								name, symbol->size, addr);
+						}
+					} else {
+						// we dont want unnamed symbol flags
+					}
+				}
+				binfile = r_core_bin_cur (r);
+				plugin = r_bin_file_cur_plugin (binfile);
+				if (plugin && plugin->name) {
+					if (!strncmp (plugin->name, "pe", 2)) {
```
radare: maybe r_str_startswith()

libr/core/cbin.c
ret2libc: No this is not done above. This is done above only if `IS_MODE_SET (mode)` is true, which means this won't get executed during printing in RAD mode.

ret2libc:
@radare I think we should merge this. I already have another branch that adds hints/metas  from special symbols, but I'd prefer to keep the two things separated.

radare:
Ok go for it

PR #13174, opened 2019-02-22
refactored cmd_print.c to use pj.c

deepakchethan:


libr/core/cmd_print.c
```diff
@@ -2741,26 +2741,29 @@ static int cmd_print_blocks(RCore *core, const char *input) {
...
+
 			if ((as->block[p].flags)
 				|| (as->block[p].functions)
 				|| (as->block[p].comments)
 				|| (as->block[p].symbols)
 				|| (as->block[p].perm)
 				|| (as->block[p].strings)) {
-				r_cons_printf ("\"offset\":%"PFMT64u ",", at);
-				r_cons_printf ("\"size\":%"PFMT64u ",", piece);
+					pj_kn (pj, "offset", at);
```
ret2libc: wrong indentation.

deepakchethan: fixed it :)

deepakchethan:
I have added a new method to pj.c library to print raw strings. The already present one was trying to escape utf8 and in doing so was adding additional `\`s.

libr/util/pj.c
```diff
@@ -133,6 +133,24 @@ R_API PJ *pj_kd(PJ *j, const char *k, double d) {
...
 
+R_API PJ *pj_krs(PJ *j, const char *k, const char *v) {
+	if (j && v && *v) {
+		pj_k (j, k);
+		pj_rs (j, v);
+	}
+	return j;
+}
+
+R_API PJ *pj_rs(PJ *j, const char *k) {
```
radare: thats wrong

libr/util/pj.c
```diff
@@ -133,6 +133,24 @@ R_API PJ *pj_kd(PJ *j, const char *k, double d) {
 	return j;
 }
 
+R_API PJ *pj_krs(PJ *j, const char *k, const char *v) {
```
radare: this too

libr/core/cmd_print.c
```diff
@@ -5102,21 +5121,22 @@ static int cmd_print(void *data, const char *input) {
...
 				}
-
-				r_cons_printf ("{\"string\":");
+				pj_o (pj);
 				str = r_str_utf16_encode ((const char *) core->block, len);
-				r_cons_printf ("\"%s\"", str);
-				r_cons_printf (",\"offset\":%"PFMT64u, core->offset);
-				r_cons_printf (",\"section\":\"%s\"", vaddr == UT64_MAX? "unknown": section->name);
-				r_cons_printf (",\"length\":%d", len);
+				pj_krs (pj, "string", str);
```
radare: just pass pj_ks and remove the r_str_utf16_encode which is wrong, the only thing you have to do is null terminate that core->block at the given len

deepakchethan: On making the changes that you said. I am not getting the right output. The new output is as follows:

screen shot 2019-02-24 at 4 57 06 pm

radare:
then we probably need to fix the json-utf8 call in the pj api, but it's a string, and there's only one type of string in json, adding a raw api to fill the string can be a root of problems

libr/util/pj.c
deepakchethan: Removed it.

libr/util/pj.c
deepakchethan: removed it

deepakchethan:
> then we probably need to fix the json-utf8 call in the pj api, but it's a string, and there's only one type of string in json, adding a raw api to fill the string can be a root of problems
> […](#)

For now I reverted to the string implementation for that particular instance. I'll try to change the json-utf method in another pr. If that is okay with you.

libr/core/cmd_print.c
```diff
@@ -5723,26 +5743,31 @@ static int cmd_print(void *data, const char *input) {
...
+
 						comma = ",";
 						// XXX: this only works in little endian
 						withref = 0;
 						if (core->print->hasrefs) {
 							char *rstr = core->print->hasrefs (core->print->user, val, true);
 							if (rstr && *rstr) {
 								char *ns = r_str_escape (rstr);
-								r_cons_printf (",\"ref\":\"%s\"}", *ns == ' ' ? ns + 1 : ns);
+								pj_ks (pj, "ref", *ns == ' ' ? ns + 1 : ns);
```
radare: just use r_str_trim_ro (ns);

radare:
i can merge after this fix, and yeah you can find a solution for the other problem later in a separate pr

libr/core/cmd_print.c
deepakchethan: Done

ret2libc:
Thanks!

PR #13175, opened 2019-02-22
Use macros for common flagspaces names instead of hardcoding them

ret2libc:


ret2libc:
In theory all those binding could be replaced in favour of REvent, but this is not the PR to do so...

PR #13176, opened 2019-02-22
WIP: Add rb_min_addr field to function, to make rbtree work

ret2libc:
This should fix https://github.com/radare/radare2/issues/13078


ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1708

ret2libc:
Closing this because it cannot work. The lowest value of the interval in the interval tree should be the key and not an additional value.

PR #13180, opened 2019-02-23
asm.strenc -> bin.str.enc

kazarmy:
Make it global.


radare:
We should group all the bin.str* into bin.str.*

we have this right now:

* bin.strfilter
* bin.strings
* bin.strpurge

so i suggest:

* bin.strings
* bin.str.{purge,filter,enc}

kazarmy:
sgtm

radare:
Can you do another pr with the changes i suggested?

kazarmy:
Yes

PR #13182, opened 2019-02-24
Add symbols.objects and symbols.sections flag spaces

ret2libc:


ret2libc:
This PR introduces 2 new flag spaces and it split the flags that were in the "symbols" space in other more fine-grained spaces. This is because we are going to prioritize flag resolution by space, so it makes sense to give more priority to "objects" or less priority to "section" symbols (it works only on ELF afaik)

radare:
red

ret2libc:
Yes I will fix tests but what do you think about the change?

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1710

ret2libc:
This is ready to be merged IMO.

radare:
about the change imho its a bit overengineered and i dont see much of the benefit out of that, but its also true that i didnt even had time to test it or check in detail or discuss with you about the plans. a bit busy these days O:)

libr/core/cbin.c
```diff
@@ -1873,6 +1873,18 @@ static bool isAnExport(RBinSymbol *s) {
 	return (s->bind && !strcmp (s->bind, R_BIN_BIND_GLOBAL_STR));
 }
 
+static void select_flag_space(RCore *core, RBinSymbol *symbol) {
+	if (!strncmp (symbol->name, "imp.", 4)) {
```
radare: why is "imp." inlined and not using a define? what about sym.imp.?

radare: i think that we can check if a symbol is an import by checking the irght propert¡y instead of asuming the name contains all the information

radare: i think symbol->type should be "import" or something likt that

ret2libc: this would require a lot of changes in all plugins... I'm not sure that's something worth at this point.

radare:
Ok merge it

radare:
btw what is symbols.sections?

ret2libc:
@radare I want to use `symbols.sections` for some symbols in the ELF that just reference the sections (see https://github.com/radare/radare2/pull/12694). Those symbols are needed to show meaningful output when you print relocations that are based on section base address.

But maybe we can avoid introducing `symbols.object` if we don't want to split `symbols` flagspace too much.

radare:
I wouldnt call them symbols. They are flags for the sections but not symbols

ret2libc:
They are symbols. They came from the symbol table together with all the other symbols.

radare:
ok then go merge it

PR #13189, opened 2019-02-24
Prerelease travis release

ret2libc:


Maijin:
I think we shouldn't use system libs for travis. The reason is that users are using ./sys/install.sh they don't use those -sysXXX=XXX options. If we test something that what users are doing, we're going to end-up not being able to spot problems of sys/install.sh

ret2libc:
That is only *one* of the tests. And it is useful to see possible bugs when radare2 is packaged in distributions like ubuntu or Fedora or others (where radare2 is not just ./sys/install.sh but it does use system libraries)

Maijin:
Ah right my mistake, i thought was for all the tests

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1715

ret2libc:
@radare i'm going to merge this. Is that ok?

PR #13193, opened 2019-02-25
Use fcn->meta.min instead of fcn->addr in fcn rbtree code

### LINKED ISSUE #13078: Function cannot be found if basic block is above function entry
### Work environment

| OS/arch/bits (mandatory)                             | voidlinux x86 64
| Architecture/bits of the file (mandatory)            | x86/32, x86/64 etc.

If a function has a basic block above the entry of the function the node will not be detected
as part of the function. For example tabbing throught the nodes in graph view will result in 'No function at ...' for the specific node.

https://github.com/radare/radare2-regressions/pull/1696

kazarmy:
Fixes #13078.

### DIFF
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -76,12 +76,37 @@ R_API void r_anal_fcn_update_tinyrange_bbs(RAnalFunction *fcn) {
 	}
 }
 
-// _fcn_tree_{cmp_addr,calc_max_addr,free,probe} are used by interval tree.
-static int _fcn_tree_cmp_addr(const void *a_, const RBNode *b_) {
+static void set_meta_min_if_needed(RAnalFunction *x) {
+	if (x->meta.min == UT64_MAX) {
+		if (r_list_length (x->bbs) == 0) {
+			x->meta.min = x->addr;
+		} else {
+			ut64 min = UT64_MAX;
+			ut64 max = UT64_MIN;
+			RListIter *bbs_iter;
+			RAnalBlock *bbi;
+			r_list_foreach (x->bbs, bbs_iter, bbi) {
+				if (min > bbi->addr) {
+					min = bbi->addr;
+				}
+				if (max < bbi->addr + bbi->size) {
+					max = bbi->addr + bbi->size;
+				}
+			}
+			x->meta.min = min;
+			x->_size = max - min; // HACK TODO Fix af size calculation
+		}
+	}
+}
+
+// _fcn_tree_{cmp,calc_max_addr,free,probe} are used by interval tree.
+static int _fcn_tree_cmp(const void *a_, const RBNode *b_) {
 	const RAnalFunction *a = (const RAnalFunction *)a_;
 	const RAnalFunction *b = FCN_CONTAINER (b_);
-	ut64 from0 = a->addr, to0 = a->addr + a->_size,
-		from1 = b->addr, to1 = b->addr + b->_size;
+	set_meta_min_if_needed ((RAnalFunction *)a);
+	set_meta_min_if_needed ((RAnalFunction *)b);
+	ut64 from0 = a->meta.min, to0 = a->meta.min + a->_size,
+		from1 = b->meta.min, to1 = b->meta.min + b->_size;
 	if (from0 != from1) {
 		return from0 < from1 ? -1 : 1;
 	}
@@ -94,7 +119,8 @@ static int _fcn_tree_cmp_addr(const void *a_, const RBNode *b_) {
 static void _fcn_tree_calc_max_addr(RBNode *node) {
 	int i;
 	RAnalFunction *fcn = FCN_CONTAINER (node);
-	fcn->rb_max_addr = fcn->addr + (fcn->_size == 0 ? 0 : fcn->_size - 1);
+	set_meta_min_if_needed (fcn);
+	fcn->rb_max_addr = fcn->meta.min + (fcn->_size == 0 ? 0 : fcn->_size - 1);
 	for (i = 0; i < 2; i++) {
 		if (node->child[i]) {
 			RAnalFunction *fcn1 = FCN_CONTAINER (node->child[i]);
@@ -124,8 +150,9 @@ static RBNode *_fcn_tree_probe(FcnTreeIter *it, RBNode *x_, ut64 from, ut64 to) 
 			x = y;
 			continue;
 		}
-		if (x->addr <= to - 1) {
-			if (from <= x->addr + (x->_size == 0 ? 0 : x->_size - 1)) {
+		set_meta_min_if_needed (x);
+		if (x->meta.min <= to - 1) {
+			if (from <= x->meta.min + (x->_size == 0 ? 0 : x->_size - 1)) {
 				return x_;
 			}
 			if ((y_ = x_->child[1])) {
@@ -141,15 +168,15 @@ static RBNode *_fcn_tree_probe(FcnTreeIter *it, RBNode *x_, ut64 from, ut64 to) 
 }
 
 R_API bool r_anal_fcn_tree_delete(RAnal *anal, RAnalFunction *data) {
-	return r_rbtree_aug_delete (&anal->fcn_tree, data, _fcn_tree_cmp_addr, _fcn_tree_free, _fcn_tree_calc_max_addr);
+	return r_rbtree_aug_delete (&anal->fcn_tree, data, _fcn_tree_cmp, _fcn_tree_free, _fcn_tree_calc_max_addr);
 }
 
 R_API void r_anal_fcn_tree_insert(RAnal *anal, RAnalFunction *fcn) {
-	r_rbtree_aug_insert (&anal->fcn_tree, fcn, &(fcn->rb), _fcn_tree_cmp_addr, _fcn_tree_calc_max_addr);
+	r_rbtree_aug_insert (&anal->fcn_tree, fcn, &(fcn->rb), _fcn_tree_cmp, _fcn_tree_calc_max_addr);
 }
 
 static void _fcn_tree_update_size(RAnal *anal, RAnalFunction *fcn) {
-	r_rbtree_aug_update_sum (anal->fcn_tree, fcn, &(fcn->rb), _fcn_tree_cmp_addr, _fcn_tree_calc_max_addr);
+	r_rbtree_aug_update_sum (anal->fcn_tree, fcn, &(fcn->rb), _fcn_tree_cmp, _fcn_tree_calc_max_addr);
 }
 
 #if 0
@@ -199,6 +226,7 @@ static void _fcn_tree_print_dot(RBNode *n) {
 }
 #endif
 
+#if 0
 // Find RAnalFunction whose addr is equal to addr
 static RAnalFunction *_fcn_tree_find_addr(RBNode *n, ut64 addr) {
 	while (n) {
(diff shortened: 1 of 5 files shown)


libr/anal/fcn.c
```diff
@@ -124,8 +150,9 @@ static RBNode *_fcn_tree_probe(FcnTreeIter *it, RBNode *x_, ut64 from, ut64 to)
 			x = y;
 			continue;
 		}
-		if (x->addr <= to - 1) {
-			if (from <= x->addr + (x->_size == 0 ? 0 : x->_size - 1)) {
+		set_meta_min_if_needed (x);
```
ret2libc: why? it's weird you need to adjust nodes while iterating over them.

libr/anal/fcn.c
```diff
@@ -237,11 +266,12 @@ static void _fcn_tree_iter_next(FcnTreeIter *it, ut64 from, ut64 to) {
 		}
 		x_ = it->path[--it->len];
 		x = FCN_CONTAINER (x_);
-		if (to - 1 < x->addr) {
+		set_meta_min_if_needed (x);
```
ret2libc: why? it's weird you need to adjust nodes while iterating over them.

kazarmy:
Reverted since https://github.com/radare/radare2/blob/2c2e9452ba69d8130ebd2ab903ec78a81922d898/libr/anal/fcn.c#L272-L275 can apparently cause the search to end prematurely if the rbtree is not keyed to meta.min.

ret2libc:
@kazarmy the key of the rbtree should be the same the lower value of the interval for the thing to work well. The approach of using meta.min as key may work, but then you have to live with slower queries for "give me function at address X".

IMHO a much better approach would be to directly store basicblocks and avoid the tinyrange search altogether.

kazarmy:
There can be 2 rbtrees, one keyed on meta.min and the other keyed on addr. The 2 rbtrees combined will likely take less space than a single rbtree based on basic blocks, depending on the average number of basic blocks per function.

ret2libc:
Yes, still it would be slower IMHO because of the tinyrange thing.

But yeah, we can use 2 rbtrees. It's just one more RBNode in each RAnalFunction.

radare:
Bear in mind that a single bb can be owned by more than one function

radare:
pls rebase and squash

ret2libc:
@radare this is still highly WIP.

kazarmy:
> pls rebase and squash

done

kazarmy:
> Bear in mind that a single bb can be owned by more than one function

This would mean that a single addr can be owned by more than one function, thus we should be using `RList *r_anal_get_fcns_in(addr)` instead of `RAnalFunction *r_anal_get_fcn_in(addr)`?

ret2libc:
> Bear in mind that a single bb can be owned by more than one function

I'm not sure we want to go that way (or if we are already there...) these kind of decisions have a big impact on how analysis is done. Another way I've seen this done is to assume an address is only part of a basic block and there are no shared basic blocks. If two functions happen to jump to the same basic block, then it's like they are calling a new function.

kazarmy:
yay it's green

radare:
❤️

libr/anal/fcn.c
```diff
@@ -200,9 +242,10 @@ static void _fcn_tree_print_dot(RBNode *n) {
 #endif
 
 // Find RAnalFunction whose addr is equal to addr
-static RAnalFunction *_fcn_tree_find_addr(RBNode *n, ut64 addr) {
+static RAnalFunction *_fcn_addr_tree_find_addr(RAnal *anal, ut64 addr) {
```
ret2libc: I think you can remove this function and directly use `r_rbtree_find`

libr/anal/fcn.c
```diff
@@ -76,12 +76,37 @@ R_API void r_anal_fcn_update_tinyrange_bbs(RAnalFunction *fcn) {
 	}
 }
 
+static void set_meta_min_if_needed(RAnalFunction *x) {
```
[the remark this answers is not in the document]
ret2libc: Have you looked at what I've suggested above?

libr/anal/fcn.c
kazarmy: Yes but:

1. `r_rbtree_find` requires a `RBComparator` function that accepts a `const void *` and a `const RBNode *`.
1. That first argument has to be castable to a `const ut64 *` for this case, since `_fcn_addr_tree_find_addr` only accepts addrs.
1. I can't reuse `_fcn_addr_tree_cmp` as the comparator function because it expects a `const RAnalFunction *` as its first argument.
1. I could do another comparator function, say  `_fcn_addr_tree_cmp_addr` but:
    1. That just replaces 1 function with another function.
    1. The first argument needs to be a pointer, and thus will cause unnecessary dereferencing.

So I don't think I should use `r_rbtree_find` here.

ret2libc: On the other hand you'd just use the provided API, no need to know the internal details of a RBTree and in case RBTree will ever be touched, you won't need to touch this code at all.

But ok for keeping it as is.

libr/anal/fcn.c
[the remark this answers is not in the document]
kazarmy: Yes but the problem is that you can't key the tree on `addr` and then test to end the search by checking `rb_min_addr` [here](https://github.com/radare/radare2/pull/13176/files#diff-49a52329a1b3f5e5f62d571577391074R197), because there is no guarantee that `rb_min_addr` will be larger (or more accurately, `rb_min_addr` will not be smaller) as you climb the tree.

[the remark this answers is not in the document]
kazarmy: > this is slow IMHO. In my PR i solved this by extending RTinyRange with a max_value, min_value fields.

Ok, lgtm

libr/anal/fcn.c
kazarmy: Ah the sacrifices made for code that appears fast.

ret2libc: It would still be fast. Avoiding dereferences, etc. are "tricks" to improve performances, but IMHO they are things that should be considered only when it was proven the slowness comes from there. I'd always prefer clean code before starting doing micro-improvements.

libr/anal/fcn.c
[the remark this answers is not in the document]
ret2libc: > Yes but the problem is that you can't key the tree on addr and then test to end the search by checking rb_min_addr here, because there is no guarantee that rb_min_addr will be larger (or more accurately, rb_min_addr will not be smaller) as you climb the tree.

What do you mean? This has nothing to do with how you key the tree I think. What I'm saying is that you can already compute `min_value`/`max_value` (consequently, you can also compute _size) by computing those values as you construct the RTinyRange.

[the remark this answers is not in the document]
kazarmy: Sorry, I was looking at all of the code at the given link, not just your suggestion.

libr/anal/fcn.c
kazarmy: > no need to know the internal details

That's where the documentation is.

ret2libc: :D

libr/anal/fcn.c
```diff
@@ -141,15 +179,19 @@ static RBNode *_fcn_tree_probe(FcnTreeIter *it, RBNode *x_, ut64 from, ut64 to)
 }
 
 R_API bool r_anal_fcn_tree_delete(RAnal *anal, RAnalFunction *data) {
-	return r_rbtree_aug_delete (&anal->fcn_tree, data, _fcn_tree_cmp_addr, _fcn_tree_free, _fcn_tree_calc_max_addr);
+	bool ret_min = !!r_rbtree_aug_delete (&anal->fcn_tree, data, _fcn_tree_cmp, _fcn_tree_free, _fcn_tree_calc_max_addr);
+	bool ret_addr = !!r_rbtree_aug_delete (&anal->fcn_addr_tree, data, _fcn_addr_tree_cmp, NULL, NULL);
```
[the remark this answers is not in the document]
kazarmy: ok 84d0c1fc2d774a6bfdc4b8464bed6d5d304f1a43

libr/anal/fcn.c
```diff
@@ -141,15 +179,19 @@ static RBNode *_fcn_tree_probe(FcnTreeIter *it, RBNode *x_, ut64 from, ut64 to)
...
+	r_return_val_if_fail (ret_min == ret_addr, false);
+	return ret_min;
 }
 
 R_API void r_anal_fcn_tree_insert(RAnal *anal, RAnalFunction *fcn) {
-	r_rbtree_aug_insert (&anal->fcn_tree, fcn, &(fcn->rb), _fcn_tree_cmp_addr, _fcn_tree_calc_max_addr);
+	r_rbtree_aug_insert (&anal->fcn_tree, fcn, &(fcn->rb), _fcn_tree_cmp, _fcn_tree_calc_max_addr);
+	r_rbtree_aug_insert (&anal->fcn_addr_tree, fcn, &(fcn->addr_rb), _fcn_addr_tree_cmp, NULL);
```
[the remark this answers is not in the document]
kazarmy: ok 84d0c1fc2d774a6bfdc4b8464bed6d5d304f1a43

kazarmy:
Re-WIPped for attempt(s) to remove `set_meta_min_if_needed`.

radare:
Status of this PR?

kazarmy:
In `set_meta_min_if_needed`:

- `meta.min` is computed only if it has the initial value `UT64_MAX` and this should happen only once, thus the function should not cause significant performance problems. However, if the correct value of `meta.min` changes after it has been inserted into the rbtree (say, through `afb+`), afaik this will not be reflected in the rbtree.
- `_size` is (re)computed if `meta.min` has the initial value `UT64_MAX`, to fix a bug triggered by 1 of the tests. This does mean that if some other code has already computed `meta.min` (e.g. `afll`), this fix will not be applied.

However, I think that this PR does improve strictly on the analysis code in master (i.e. makes analysis more correct and does not introduce regressions) and I would consider it mergable.

radare:
im +1 for merging

libr/anal/fcn.c
ret2libc: IMHO this is wrong... if meta.min would ever require to be recomputed, that means the search of the function is already wrong. You can't change the tree when you are in the middle of a traverse.

libr/anal/fcn.c
ret2libc: same as https://github.com/radare/radare2/pull/13193/files#r262920262

ret2libc:
I have just 2 small comments I just added. Let's wait for @kazarmy's reply. Apart from those 2 things, I think the PR is ok.

radare:
yay <3

libr/anal/fcn.c
kazarmy: ok removed 31325792d0959b735e913dbf83c7387d0e8844e6 they were added in a copy-and-paste fit, but are harmless

libr/anal/fcn.c
kazarmy: see https://github.com/radare/radare2/pull/13193#discussion_r262942662

kazarmy:
> I have just 2 small comments I just added. Let's wait for @kazarmy's reply. Apart from those 2 things, I think the PR is ok.

done ... can merge now?

ret2libc:
If Travis is still green, go ahead!

radare:
only one xx, probably caused by missing rebase :?

screenshot 2019-03-06 at 14 58 22

ret2libc:
yes, i think so. you can go ahead.

PR #13202, opened 2019-02-26
Support another type of jmptbl for x86-64 ##anal

radare:


radare:
tests in https://github.com/radare/radare2-regressions/pull/1713

PR #13205, opened 2019-02-27
Implement 'aesou' command (step over until) ##anal

radare:
Pending things..probably to do in a separate PR to not make it bigger

* [ ] Support syscall name/parameters resolution like its done in `as`
* [ ] Handle tailcall function parameters
* [ ] Implement aC* and aCj


libr/core/cmd_anal.c
```diff
@@ -79,6 +79,7 @@ static const char *help_msg_ab[] = {
 static const char *help_msg_abt[] = {
 	"Usage:", "abt", "[addr] [num] # find num paths from current offset to addr",
 	"abt", " [addr] [num]", "find num paths from current offset to addr",
+	"abte", " [addr]", "emulate from begining of function to the given address",
```
ret2libc: The beginning of which function?

radare:
The one that covers the address you give as argument

radare:
This thing needs to support syscall name resolution too (like `as` does)

PR #13209, opened 2019-02-27
Fix #13208 - Initial import of the asm.null plugin ##asm

### LINKED ISSUE #13208: Implement asm.null plugin
To be in sync with anal.null

radare:


radare:
we should include this plugin in more plugins cfg files

PR #13216, opened 2019-02-28
Fix va_arg issue in r_str_newf and r_str_appendf ##core

radare:
This patch improves performance and solves a random crash that was hard to catch

### DIFF
--- a/libr/util/str.c
+++ b/libr/util/str.c
@@ -895,7 +895,7 @@ R_API char *r_str_appendf(char *ptr, const char *fmt, ...) {
 	va_start (ap, fmt);
 	if (!strchr (fmt, '%')) {
 		va_end (ap);
-		return strdup (fmt);
+		return r_str_append (ptr, fmt);
 	}
 	va_copy (ap2, ap);
 	int ret = vsnprintf (NULL, 0, fmt, ap2);
--- a/libr/util/str.c
+++ b/libr/util/str.c
@@ -711,42 +711,24 @@ R_API char *r_str_trunc_ellipsis(const char *str, int len) {
 	return buf;
 }
 
-// Returns a new heap-allocated string that matches the format-string
-// specification.
 R_API char *r_str_newf(const char *fmt, ...) {
-	int ret, ret2;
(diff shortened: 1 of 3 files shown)


libr/util/str.c
```diff
@@ -711,42 +711,24 @@ R_API char *r_str_trunc_ellipsis(const char *str, int len) {
...
-			va_end (ap2);
-			va_end (ap);
-			return NULL;
-		}
-		tmp = r_str_new (p);
-		free (p);
-	} else {
-		tmp = r_str_new (string);
+	va_copy (ap2, ap);
+	int ret = vsnprintf (NULL, 0, fmt, ap2);
```
ret2libc: is this behaviour used everywhere?

libr/util/str.c
```diff
@@ -908,30 +890,29 @@ R_API char *r_str_append(char *ptr, const char *string) {
...
-	if (ret >= sizeof (string)) {
-		char *p = malloc (ret + 2);
-		if (!p) {
-			va_end (ap);
-			return NULL;
-		}
-		vsnprintf (p, ret + 1, fmt, ap);
+	if (!strchr (fmt, '%')) {
+		va_end (ap);
+		return strdup (fmt);
```
ret2libc: ? r_str_appendf should append, not replace AFAIK

libr/util/str.c
radare: wat

libr/util/str.c
radare: yes this was a regression .fixed

libr/util/str.c
radare: you mean if its portable to use null and 0 to get the size of the output buffer? its in the manpage at least, so i assume its standard and portable. we can discover that later, but at least this is faster than old code because its only writing the resulting string once, not twice in case it doesnt fit in the local buffer, which was an unnecessary work because then it is required to strdup again.

ret2libc: Ok, I can confirm it is in the linux man page. If it works on OS X too, we are 90% ok :)

PR #13226, opened 2019-03-01
Consistently use fcnIn() in ds_show_xrefs()

kazarmy:
Fixes some XX tests in #13193.


kazarmy:
Closing this because it merely hides but doesn't solve the problem.

PR #13232, opened 2019-03-02
Update Windows type db ##anal

pelijah:


xarkes:
"I have done until I letter" did you add that or was it here before?

pelijah:
I think it was there even before you joined the project.

pelijah:
LoadStringA returns int
https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-loadstringa

PR #13239, opened 2019-03-03
Pass RAnal instead of rbtree to fcn rbtree api

kazarmy:
For #13193 and also for greater flexibility in implementing the fcn rbtree.

### DIFF
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -140,16 +140,16 @@ static RBNode *_fcn_tree_probe(FcnTreeIter *it, RBNode *x_, ut64 from, ut64 to) 
 	}
 }
 
-R_API bool r_anal_fcn_tree_delete(RBNode **root, RAnalFunction *data) {
-	return r_rbtree_aug_delete (root, data, _fcn_tree_cmp_addr, _fcn_tree_free, _fcn_tree_calc_max_addr);
+R_API bool r_anal_fcn_tree_delete(RAnal *anal, RAnalFunction *data) {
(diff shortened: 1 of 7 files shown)


libr/anal/fcn.c
```diff
@@ -140,16 +140,16 @@ static RBNode *_fcn_tree_probe(FcnTreeIter *it, RBNode *x_, ut64 from, ut64 to)
...
+	return r_rbtree_aug_delete (&anal->fcn_tree, data, _fcn_tree_cmp_addr, _fcn_tree_free, _fcn_tree_calc_max_addr);
 }
 
-R_API void r_anal_fcn_tree_insert(RBNode **root, RAnalFunction *fcn) {
-	r_rbtree_aug_insert (root, fcn, &(fcn->rb), _fcn_tree_cmp_addr, _fcn_tree_calc_max_addr);
+R_API void r_anal_fcn_tree_insert(RAnal *anal, RAnalFunction *fcn) {
+	r_rbtree_aug_insert (&anal->fcn_tree, fcn, &(fcn->rb), _fcn_tree_cmp_addr, _fcn_tree_calc_max_addr);
 }
 
-static void _fcn_tree_update_size(RBNode *root, RAnalFunction *fcn) {
```
radare: Those are internals. Makes sense to keep them like this i think. Otherwise we are dereferencing twice unnecesarily

kazarmy: `_fcn_tree_update_size()` is called only by `r_anal_fcn_set_size()` (https://github.com/radare/radare2/pull/13239/files#diff-49a52329a1b3f5e5f62d571577391074L2255) in which the only argument possible is the root of the fcn rbtree. And #13193 will need to maintain 2 rbtrees so I think it's better that the updates to both is done in 1 function to prevent accidental half-updates.

PR #13240, opened 2019-03-03
Warn when invalid use of r_buf_buffer

radare:


libr/util/buf.c
```diff
@@ -402,7 +402,10 @@ R_API RBuffer *r_buf_new() {
 }
 
 R_API const ut8 *r_buf_buffer (RBuffer *b) {
-	return (b && !b->sparse)? b->buf: NULL;
+	if (b && !b->sparse && b->fd == -1 && !b->mmap) {
+		return b->buf;
+	}
+	r_return_val_if_fail (false, NULL);
```
ret2libc: Mmm why? Is it a programming error if one calls r_buf_buffer on a sparse RBuffer? This change would imply one has to know the internal details of the RBuffer before calling this function and it seems to go against the purpose of having a general RBuffer

radare:
Yes because its not linearly allocated same goes for fd based rbuffers and such. We may want to crash/assert earlier imho

radare:
Pls suggest improvements on this conditional because inthink i missed some cases but wnotime now. And if u are happy eith this usage of r_return ill merge after that

ret2libc:
So... Apparently `r_buf_buffer` is used only in all the `bin_*` plugins and in `io_gprobe`(which I have no idea what it is). In all the bin plugins it is used in the old `load` method, which is now deprecated in favor of `load_buffer`. I think `r_buf_buffer` should be deprecated as well(maybe add a comment on top of it, both in the .c and in .h). What do you think?

I think it should be deprecated because it's an operation that cannot be done/we don't want to do in the general case.

Unless.... we want to give a meaning to r_buf_buffer in the sparse case too and in the other cases, but I'm not sure it's worth it.

ret2libc:
I'm ok with the use of `r_return_*` if we want to deprecate `r_buf_buffer`. If instead we want to keep that API, I think we should implement it somehow for all the types of RBuffer.

radare:
Its not possible to implement a linear buffer for an infinite memory mapped or sparsed data or anything huge like a 2TB hard drive. Thats why its not recommended to use it unless you dont own the rbuffer instance.

ret2libc:
Perfect, in that case I think we should deprecate `r_buf_buffer`, as it's not that useful. All those `load` functions in the RBinPlugin could actually be removed soon, because `load_buffer` should be used instead. So there isn't a real need for `r_buf_buffer` to exist.

Are you ok with deprecating it and putting a comment on top of that function for now?

radare:
its useful because it save the need to copy the data to a new buffer when you already have it there, but its dangerous for all other usecases, which caused many problems in rbin because of bad usage

ret2libc:
I don't see any need for that API as of right now in the current codebase. All uses are in old `load` method of RBinPlugin. The only reason i can see for this PR is to interact with APIs that still use `ut8*` instead of RBuffer

radare:
Its used in other places but with legit use. At least we will catch if any plugin uses it wrongly

ret2libc:
I agree

PR #13246, opened 2019-03-03
Add eval for refptr #asm

Maijin:
Add eval for refptr and disable it by default, should reduce the volume of output in disassembly while leaving the user able to show this information back

```
Default Before:
0x100001237      c7058f420000.  mov dword [0x1000054d0], 0x50    ; 'P' ; section.10.__DATA.__data ; [0x1000054d0:4]=80 ; U"P."

Default After:
0x100001237      c7058f420000.  mov dword [0x1000054d0], 0x50    ; 'P' ; section.10.__DATA.__data ; U"P."
```

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -2747,6 +2747,7 @@ R_API int r_core_config_init(RCore *core) {
 	SETPREF ("asm.nodup", "false", "Do not show dupped instructions (collapse disasm)");
 	SETPREF ("asm.emu", "false", "Run ESIL emulation analysis on disasm");
 	SETPREF ("emu.pre", "false", "Run ESIL emulation starting at the closest flag in pd");
+	SETPREF ("asm.refptr", "false", "Show refpointer information in disasm");
 	SETPREF ("emu.lazy", "false", "Do not emulate all instructions with aae (optimization)");
 	SETPREF ("emu.stack", "false", "Create a temporary fake stack when emulating in disasm (asm.emu)");
 	SETCB ("emu.str", "false", &cb_emustr, "Show only strings if any in the asm.emu output");
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -2747,7 +2747,7 @@ R_API int r_core_config_init(RCore *core) {
 	SETPREF ("asm.nodup", "false", "Do not show dupped instructions (collapse disasm)");
 	SETPREF ("asm.emu", "false", "Run ESIL emulation analysis on disasm");
 	SETPREF ("emu.pre", "false", "Run ESIL emulation starting at the closest flag in pd");
-	SETPREF ("asm.refptr", "false", "Show refpointer information in disasm");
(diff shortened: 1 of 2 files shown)


Maijin:
To be discussed

Maijin:
It seems to have important usecases tested here:

image
image

I'm going to change the PR to only disable it for x86

Maijin:
`I'm going to change the PR to only disable it for x86`

But I don't know how to do that `asmarch->value` always return x86 in `cconfig.c` anyone?

ret2libc:
I often find those information not that useful (and sometimes not right), so that sounds good with me

radare:
Thats racist. Imho the condition shouldnt be x86

Maijin:
@radare Can you detail? On arm and MIPS the information brings something useful "sometimes", on x86 it's not never useful if you check the tests, only verbose. I have only setup the eval to false on x86 by default the users can still change it back manually if it fits their usecases

radare:
The d cision of considering this useful or not depends more on the text in the disasm than the arch itself

Maijin:
Yes imho  this is a good quick-fix for now because for ARM/MIPS if you check the tests have relevant output, x86 have 0 cases where the output is relevant without tweaking. We can always change this behavior later when there are better checks for the text itself which will requires much more work if not complete refactoring. The patch improves a lot of the readability in x86 disassembly

radare:
I would like to review the pr in r2r to understand this, but probably makes more sense to change this variable depending on x86, instead of making this thing behave different in case of x86

radare:
why its red? can you rebase?

libr/core/cconfig.c
```diff
@@ -195,6 +195,11 @@ static void rasm2_list(RCore *core, const char *arch, int fmt) {
 	}
 }
 
+static inline void __setshowrefptr(RConfig *cfg, const char *asmarch, int asmbits) {
+	int show_refptr = (strncmp (asmarch, "x86", 3));
```
radare: use bool

radare: i would make this config global instead of x86 specific first, otherwise there are many tests to update and we have no room for testing. maybe after release we can evaluate enabling it for some archs, etc

Maijin:
https://github.com/radare/radare2-regressions/pull/1728

libr/core/cconfig.c
```diff
@@ -2747,6 +2747,7 @@ R_API int r_core_config_init(RCore *core) {
 	SETPREF ("asm.nodup", "false", "Do not show dupped instructions (collapse disasm)");
 	SETPREF ("asm.emu", "false", "Run ESIL emulation analysis on disasm");
 	SETPREF ("emu.pre", "false", "Run ESIL emulation starting at the closest flag in pd");
+	SETPREF ("asm.refptr", "false", "Show refpointer information in disasm");
```
radare: set it to true by default, otherwise the default behaviour is changing,

PR #13253, opened 2019-03-03
Remove SLOW_IO from core/anal.c

radare:


radare:
This PR is also fixing a double-free

PR #13265, opened 2019-03-04
Fix esil negative numbers

aronsky:
esil's handling of negative numbers isn't perfect, but when subtraction is involved, it's basically non-existant.

For example, `ae 3,-4,+` results in `0xffffffffffffffff` (since esil treats all numbers as `ut64`, and it's tricky to know when to treat it as signed/unsigned). So, a bummer, but at least it works.

However, `ae 3,-4,-` returns nothing (as opposed to the expected `0xfffffffffffffff9`). Reason: the current implementation of `esil_sub` reads the parameters directly, and validates that they are numbers only by checking for digits (instead of using the more generic `r_anal_esil_get_parm`, used in `esil_add`, `esil_mul`, and `esil_div`, that supports negative numbers as well).

This PR fixes the issue - the code for `esil_sub` is copied from `esil_add`, with the actual operation, `s + d` replaced with `d - s`.

### DIFF
--- a/libr/anal/esil.c
+++ b/libr/anal/esil.c
@@ -1583,11 +1583,9 @@ static int esil_add(RAnalEsil *esil) {
 	ut64 s, d;
 	char *dst = r_anal_esil_pop (esil);
 	char *src = r_anal_esil_pop (esil);
-	if (src && r_anal_esil_get_parm (esil, src, &s)) {
-		if (dst && r_anal_esil_get_parm (esil, dst, &d)) {
+	if ((src && r_anal_esil_get_parm (esil, src, &s)) && (dst && r_anal_esil_get_parm (esil, dst, &d))) {
 			r_anal_esil_pushnum (esil, s + d);
 			ret = true;
-		}
 	} else {
 		ERR ("esil_add: invalid parameters");
 	}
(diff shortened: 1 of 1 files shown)


radare:
cc @condret

condret:
nice catch, yeah, this patch seems good. the old version of esil_sub is obviously broken, since esil_sub should not manipulate esil->old or esil->cur ... I guess that's the reason why there are tests.

radare:
if you rebase it should be green, if its fine for @condret im merging, i was waiting for him to review it as long as he is the strictiest guy in here related to esil specs and we had some discussions too

aronsky:
> if you rebase it should be green [...]

But that would require force-pushing the branch, won't it break the PR?

radare:
no, force pusihng a branch is fine, is the right way to update a pr in fact

aronsky:
OK, good to know :)

aronsky:
> I don't believe the Travis failure has anything to do with the change.

OK, so the failure *is* caused by my changes. Oops.

Working on it. Good thing there are regression tests...

aronsky:
> nice catch, yeah, this patch seems good. the old version of esil_sub is obviously broken, since esil_sub should not manipulate esil->old or esil->cur ... I guess that's the reason why there are tests.

@condret, I looked into the tests (which I've broken :) ), and in fact it does seem that other parts of esil assume that `esil_sub` should manipulate `cur` and `old` (specifically, the failing tests assume the borrow flag is affected by `esil_sub`). I updated the patch to include updates of `cur` and `old` (under the condition `src` is not internal, same as in the `esil_*eq` functions), and the tests now pass. However, if you assert that `esil_sub` should not manipulate `cur` and `old`, how you suggest we go around the fact that those tests fail?

This raises the question, by the way, regarding the functionality of other flags after `esil_add`/`esil_mul`/`esil_div` are executed - since those functions don't touch `cur`, `old`, and `lastsz`, accessing carry/overflow flags after calling those functions will not work correctly...

radare:
thanks!

PR #13269, opened 2019-03-04
Make -x behave like /x unless mask is provided

wasamasa:
`rafind2 -x f.f.` doesn't behave like `/x f.f.` in r2. This patch adjusts this behavior unless a mask is provided.

Please let me know if you consider this change to warrant mention in the manual or usage help.


radare:
Thanks! can you add some tests for this? i think we should find better names for those functions

Issue #13271, opened 2019-03-05
fcn_print_json must use PJ api


radare:
`static int fcn_print_json(RCore *core, RAnalFunction *fcn) {`

lives in libr/core/canal.c

deepakchethan:
Will start working on it :)

radare:
👍

deepakchethan:
This can be closed

PR #13272, opened 2019-03-05
Refactored canal to pj.c

deepakchethan:
for issue #13271

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -1163,11 +1163,20 @@ static int var_cmd(RCore *core, const char *str) {
 	switch (str[1]) { // afv[bsr]
 	case '\0':
 	case '*':
-	case 'j':
-		r_anal_var_list_show (core->anal, fcn, type, str[1]);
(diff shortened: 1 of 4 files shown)


radare:
The pr looks good, please verify ththe travis output. I know the old json contained newlines so maybe its this

libr/core/cmd_anal.c
```diff
@@ -1162,11 +1162,18 @@ static int var_cmd(RCore *core, const char *str) {
 	switch (str[1]) { // afv[bsr]
 	case '\0':
 	case '*':
-	case 'j':
-		r_anal_var_list_show (core->anal, fcn, type, str[1]);
+	case 'j': {
```
radare: handle \0 and * differently, passing NULL in the last argument instead of creating the pj instance. i think this is the reason isi creating those newlines in travis

deepakchethan: Yeah, you are right. This removed a lot of newlines in the output. I made the changes.

deepakchethan: Yeah, you are right. This removed a lot of new lines from the output. I made the changes.

PR #13273, opened 2019-03-05
Fix memory leak in strbuf.

aronsky:
`sb->ptr` was set to NULL without freeing the underlying data.

### DIFF
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -2301,13 +2301,13 @@ R_API ut32 r_anal_fcn_cost(RAnal *anal, RAnalFunction *fcn) {
 		return 0;
 	}
 	r_list_foreach (fcn->bbs, iter, bb) {
-		RAnalOp op = R_EMPTY;
+		RAnalOp op;
 		ut64 at, end = bb->addr + bb->size;
 		ut8 *buf = malloc (bb->size);
 		(void)anal->iob.read_at (anal->iob.io, bb->addr, (ut8 *) buf, bb->size);
 		int idx = 0;
 		for (at = bb->addr; at < end;) {
-			// memset (&op, 0, sizeof (op)); // Not needed, op = R_EMPTY
+			memset (&op, 0, sizeof (op));
 			(void) r_anal_op (anal, &op, at, buf + idx, bb->size - idx, R_ANAL_OP_MASK_BASIC);
 			if (op.size < 1) {
 				op.size = 1;
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -2301,13 +2301,13 @@ R_API ut32 r_anal_fcn_cost(RAnal *anal, RAnalFunction *fcn) {
 		return 0;
 	}
 	r_list_foreach (fcn->bbs, iter, bb) {
-		RAnalOp op;
+		RAnalOp op = R_EMPTY;
 		ut64 at, end = bb->addr + bb->size;
 		ut8 *buf = malloc (bb->size);
 		(void)anal->iob.read_at (anal->iob.io, bb->addr, (ut8 *) buf, bb->size);
 		int idx = 0;
 		for (at = bb->addr; at < end;) {
-			memset (&op, 0, sizeof (op));
+			// memset (&op, 0, sizeof (op)); // Not needed, op = R_EMPTY
 			(void) r_anal_op (anal, &op, at, buf + idx, bb->size - idx, R_ANAL_OP_MASK_BASIC);
 			if (op.size < 1) {
 				op.size = 1;
--- a/libr/anal/op.c
+++ b/libr/anal/op.c
@@ -14,14 +14,7 @@ struct VarUsedType {
 
 R_API RAnalOp *r_anal_op_new () {
 	RAnalOp *op = R_NEW0 (RAnalOp);
-	if (op) {
-		op->addr = UT64_MAX;
-		op->jump = UT64_MAX;
-		op->fail = UT64_MAX;
-		op->ptr = UT64_MAX;
-		op->val = UT64_MAX;
-		r_strbuf_init (&op->esil);
-	}
+	r_anal_op_init(op);
 	return op;
 }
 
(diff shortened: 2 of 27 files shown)


libr/anal/fcn.c
```diff
@@ -2301,13 +2301,13 @@ R_API ut32 r_anal_fcn_cost(RAnal *anal, RAnalFunction *fcn) {
...
 	r_list_foreach (fcn->bbs, iter, bb) {
-		RAnalOp op;
+		RAnalOp op = R_EMPTY;
 		ut64 at, end = bb->addr + bb->size;
 		ut8 *buf = malloc (bb->size);
 		(void)anal->iob.read_at (anal->iob.io, bb->addr, (ut8 *) buf, bb->size);
 		int idx = 0;
 		for (at = bb->addr; at < end;) {
-			memset (&op, 0, sizeof (op));
+			// memset (&op, 0, sizeof (op)); // Not needed, op = R_EMPTY
```
ret2libc: So, IMHO this shouldn't be needed at all ever, because `r_anal_op` should set all the necessary fields. However, setting `op = R_EMPTY` outside the loop is not the same as zeroing it at each iteration.

That said, if you want to remove the line, just remove it instead of uncommenting please.

libr/anal/op.c
```diff
@@ -14,14 +14,7 @@ struct VarUsedType {
 
 R_API RAnalOp *r_anal_op_new () {
 	RAnalOp *op = R_NEW0 (RAnalOp);
```
ret2libc: use R_NEW since you are doing `memset (..0, ..)` in `r_anal_op_init`.

libr/anal/reflines.c
```diff
@@ -80,15 +80,15 @@ R_API void r_anal_reflines_free (RAnalRefline *rl) {
...
+	RAnalOp op = R_EMPTY;
 	struct refline_end *el;
 	const ut8 *ptr = buf;
 	const ut8 *end = buf + len;
 	ut8 *free_levels;
 	int res, sz = 0, count = 0;
 	ut64 opc = addr;
 
-	memset (&op, 0, sizeof (op));
+	// memset (&op, 0, sizeof (op)); // Not needed, op = R_EMPTY
```
ret2libc: remove line.

libr/asm/op.c
```diff
@@ -3,7 +3,9 @@
 #include <r_asm.h>
 
 R_API RAsmOp *r_asm_op_new() {
-	return R_NEW0 (RAsmOp);
+	RAsmOp *op = R_NEW0 (RAsmOp);
```
ret2libc: use `R_NEW`, memset is already in the init.

libr/anal/fcn.c
aronsky: You're absolutely right, I did a search for all the `RAnalOp`s defined on the stack, and removed the following `memset`s where appropriate - didn't notice this one is in a loop. I will restore it, as well as remove the other `memset`s instead of commenting them out.

aronsky:
> The memory leak fix in strbuf looks ok.
> 
> However, were all the other changes for r_anal_op_init and r_asm_op_init really necessary? In any case, if we define them, I think it makes more sense to always call them instead of relying on R_EMPTY. R_EMPTY indeed sets all the Op fields to 0, which may not be what r_asm_op_init or r_anal_op_init do (e.g. r_anal_op_init does set some fields to UT64_MAX... in this particular case it won't probably matter, but we never know in the future what happens, so I'd be more inclined to always call r_*_op_init instead of using R_EMPTY).

I wanted `r_anal_op_init` and `r_asm_op_init` to be more in sync and parallel to each other. I agree that it isn't directly related to the PR's subject, per se - but the PR deals with the initialization of those objects on the stack, so I think it's relevant.

I also agree that using `R_EMPTY` might not be the best choice when `r_*_op_init` is available for initialization. However, the `r_*_op_init` functions are currently pretty redundant, and in-place initialization is far more readable.

If you think it's best to initialize them with `r_*_op_init`, I'll change that. I agree that it's better for future-proofing the code, and if anything relies on the RAnalOp fields being initialized to UT64_MAX, that'll take care of it, too.

ret2libc:
I'm thinking about it and... maybe r_anal_op_init and r_asm_op_init shouldn't be required at all. After all, r_asm_op and r_anal_op should ensure that `op` is always initialized to some sane defaults, even if there are errors. I think currently it's not clear whether the `memset(op,0)` are still required, so maybe it would be better to not touch them at all.

@radare what do you think?

radare:
when i did that i was initializing first in the plugin side, but then some plugins was not initializing it, so it was causing UB, so i endup memsetting that in the r_anal_op function.. i think thats the right place for this to happen.

also i think that r_asm_op should be a subset of r_anal_op, because r_anal_op allows to define the level of granularity in detail, so adding asm/dis in there shuold be enough to "kill" the duplicity between asm and anal.

libr/anal/op.c
```diff
@@ -14,14 +14,7 @@ struct VarUsedType {
...
-	if (op) {
-		op->addr = UT64_MAX;
-		op->jump = UT64_MAX;
-		op->fail = UT64_MAX;
-		op->ptr = UT64_MAX;
-		op->val = UT64_MAX;
-		r_strbuf_init (&op->esil);
-	}
+	r_anal_op_init(op);
```
radare: missing space before (

ret2libc:
Ok I agree. I think the initialization should be done in r_anal_op/r_asm_op.

radare:
Yep

ret2libc:
@aronsky sorry your PR became object of discussion, but the more you modify, the riskier it is :D

So, to wrap up: memset before using r_asm_disassemble/r_anal_op are not required and neither initializing op to R_EMPTY. Indeed, `r_asm_disassemble` and `r_anal_op` should work as a sort of initializer and be sure `op` always contain consistent data. If you look at `r_anal_op` code it actually already does the memset (though it would be better to move the memset before the `r_return_val_if_fail` line, and maybe use your r_anal_op_init instead), while `r_asm_disassemble` does not call anything (though it should call memset or, even better, your `r_asm_op_init`).

Does that sound ok?

aronsky:
> @aronsky sorry your PR became object of discussion, but the more you modify, the riskier it is :D

No problem, discussion is welcome!

I reverted most of the changes, and added a call to `r_asm_op_init`/`r_anal_op_init` in `r_asm_disassemble`/`r_anal_op`, as suggested.

I left the new `r_anal_op_init` function, as I think it's correct for it to exist :)

I also called the initialization functions prior to `r_return_val_if_fail`, as per your suggestion (they are guarded against NULL pointers), thus making sure the Op is returned in an initialized state even on any failure.

ret2libc:
Thanks a lot! This looks great now ;)

PR #13274, opened 2019-03-05
Update anal_m68k_cs.c

CrypticalCode0:
EXG (EXchanGe register) is an valid instruction for R_ANAL_OP_TYPE_MOV because it only moves from RegN to RegN.


ret2libc:
Thanks!

PR #13277, opened 2019-03-05
Little code cleanups

devnexen:


libr/anal/fcn.c
```diff
@@ -1133,7 +1133,7 @@ static int fcn_recurse(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut8 *buf, ut6
 			// if first byte in op.ptr is 0xff, then set leaddr assuming its a jumptable
 			{
 				ut8 buf[4];
-				anal->iob.read_at (anal->iob.io, op.ptr, &buf, sizeof (buf));
+				anal->iob.read_at (anal->iob.io, op.ptr, (ut8 *)&buf, sizeof (buf));
```
aronsky: `buf` is defined as an array of `ut8` literally in the line above, what's the added value of explicitly casting `&buf` to `ut8*`?

devnexen: Nothing :-) it s here to silent compilation warning with pedantic modern compiler.

radare: just pass buf without the & and the cast shouoldnt be necessary i think this patch is also in my kill-maxbb PR

mk/platform.mk
```diff
@@ -1,6 +1,10 @@
 ifeq ($(OSTYPE),auto)
 OSTYPE=$(shell uname | tr 'A-Z' 'a-z')
 endif
+ifneq (,$(findstring darwin,${OSTYPE}))
```
radare: use ifneq($(OSTYPE),darwin)

.. but why? i think this will break on BSDs. and this sounds more like a linuxism

devnexen: Hmmm in fact this change was accidental....

PR #13280, opened 2019-03-06
Fix #13252 - Fix path with spaces in projects

### LINKED ISSUE #13252: space in the file's path makes project saving impossible
### Work environment

| OS/arch/bits (mandatory)                             | Windows 8.1
| r2 -v full output, **not truncated** (mandatory)     | 

radare2 3.3.0 1 @ windows-x86-32 git.3.3.0
commit: 5a9127d2599c8ff61d8544be7d4c9384402e94a3 build: Tue 02/19/2019__12:55:07.31

- run radare2 'path that contains sapce'
- use Ps
- quit radare2
- radare2 -p project_name
radare fails to to load the binary

C:\Users\Rakach>radare2 "C:\Users\Rakach\Documents\RE\binary auditing\Splish.exe"
----------------------------------------------------------space is here:^
 -- Do not try to sploit that binary - that's impossible. Instead, only try to realize the truth: there is no binary.
[0x00401000]> pd 3
            ;-- entry0:
            ;-- section..text:
            ;-- eip:
            0x00401000      push 0                                     ; [00] -r-x section size 4096 named .text
            0x00401002      call 0x40178a
            0x00401007      mov dword [0x403480], eax                  ; [0x403480:4]=0
[0x00401000]> Ps splish
splish
[0x00401000]> q
Do you want to save the 'splish' project? (Y/n) y

C:\Users\Rakach>radare2 -p splish
Do you want to print 9874 lines? (y/N) n
 -- Use rabin2 to discover the real TRUTH
[0x00401000]> pd 3
            ;-- entry0:
            ;-- section..text:
            ;-- eip:
            0x00401000      invalid                                    ; [00] -r-x section size 4096 named .text
            0x00401001      invalid
            0x00401002      invalid
[0x00401000]> q
Do you want to save the 'splish' project? (Y/n) y`
`

as you can see I cant reuse the project and that only happens when the file's path is complex(contain spaces)

GustavoLCR:
Fix #13252

### DIFF
--- a/libr/core/project.c
+++ b/libr/core/project.c
@@ -521,7 +521,7 @@ static bool store_files_and_maps (RCore *core, RIODesc *desc, ut32 id) {
 	RListIter *iter;
 	RIOMap *map;
 	if (desc) {
-		r_cons_printf ("ofs %s %s\n", desc->uri, r_str_rwx_i (desc->perm));
+		r_cons_printf ("ofs \"%s\" %s\n", desc->uri, r_str_rwx_i (desc->perm));
 		if ((maps = r_io_map_get_for_fd (core->io, id))) {
 			r_list_foreach (maps, iter, map) {
 				r_cons_printf ("om %d 0x%"PFMT64x" 0x%"PFMT64x" 0x%"PFMT64x" %s%s%s\n", fdc,
--- a/libr/core/project.c
+++ b/libr/core/project.c
@@ -521,7 +521,7 @@ static bool store_files_and_maps (RCore *core, RIODesc *desc, ut32 id) {
 	RListIter *iter;
 	RIOMap *map;
 	if (desc) {
-		r_cons_printf ("ofs \"%s\" %s\n", desc->uri, r_str_rwx_i (desc->perm));
+		r_cons_printf ("\"ofs \\\"%s\\\" %s\"\n", desc->uri, r_str_rwx_i (desc->perm));
 		if ((maps = r_io_map_get_for_fd (core->io, id))) {
 			r_list_foreach (maps, iter, map) {
 				r_cons_printf ("om %d 0x%"PFMT64x" 0x%"PFMT64x" 0x%"PFMT64x" %s%s%s\n", fdc,


libr/core/project.c
```diff
@@ -521,7 +521,7 @@ static bool store_files_and_maps (RCore *core, RIODesc *desc, ut32 id) {
 	RListIter *iter;
 	RIOMap *map;
 	if (desc) {
-		r_cons_printf ("ofs %s %s\n", desc->uri, r_str_rwx_i (desc->perm));
+		r_cons_printf ("ofs \"%s\" %s\n", desc->uri, r_str_rwx_i (desc->perm));
```
radare: The whole command should be quoted instead but that wont work because you need to pass a second argument

libr/core/project.c
```diff
@@ -521,7 +521,7 @@ static bool store_files_and_maps (RCore *core, RIODesc *desc, ut32 id) {
 	RListIter *iter;
 	RIOMap *map;
 	if (desc) {
-		r_cons_printf ("ofs %s %s\n", desc->uri, r_str_rwx_i (desc->perm));
+		r_cons_printf ("\"ofs \\\"%s\\\" %s\"\n", desc->uri, r_str_rwx_i (desc->perm));
```
GustavoLCR: > The whole command should be quoted instead but that wont work because you need to pass a second argument

Like this?

radare: no.. you should use the api, i dont think there's a way for ofs to work with paths with spaces

GustavoLCR: It uses r_str_argv() so it works fine with just the inner quotes:

pic

I think the outer ones are unnecessary imho, should I revert it?

radare: ok

radare: then its fine

radare: as long as it works i will merge. we need to rework the command parser at some point to handler str_argv everywhere and such

radare:
Can you confirm? can we have a test?

GustavoLCR:
I can confirm it is working. Should I remove the outer quotes? It also works without those.
I'm having problems writing a test for it though:
`NAME='File contain spaces'`
`FILE='../bins/other/projects/hello world'`
`CMDS=<<EXPECT`
`e dir.projects = .tmp/`
`Ps spaces > /dev/null`
`Po spaces`
`p8 8`
`EXPECT=<<RUN`
`31ed4989d15e4889`
`RUN`
This should work, no? It works fine if I do it manually with r2.

PR #13307, opened 2019-03-07
Bring XNU kernelcache plugin into the core

mrmacete:
fixes https://github.com/radare/radare2/issues/13227


libr/bin/format/xnu/r_cf_dict.c
```diff
@@ -0,0 +1,624 @@
...
+
+RCFValueDict *r_cf_value_dict_parse(RBuffer *file_buf, ut64 offset, ut64 size, int options) {
+  RCFValueDict *result = NULL;
+  void *xml_buf;
+  yxml_t x;
+  int i, depth = 0;
+  char *content = NULL;
+  RList *stack = NULL;
+
+  xml_buf = malloc(XMLBUFSIZE);
```
radare: chk null define variable here and add space before (

libr/bin/format/xnu/r_cf_dict.c
```diff
@@ -0,0 +1,624 @@
...
+  yxml_init(&x, xml_buf, XMLBUFSIZE);
+
+  stack = r_list_newf ((RListFree)&r_cf_parse_state_free);
```
radare: define and assign here

libr/bin/format/xnu/r_cf_dict.c
```diff
@@ -0,0 +1,624 @@
...
+			next_state->value_type = R_CF_STRING;
+		} else if (!strcmp (x.elem, "integer")) {
+			next_state = r_cf_parse_state_new (R_CF_STATE_IN_SCALAR);
+			next_state->value_type = R_CF_INTEGER;
+		} else if (!strcmp (x.elem, "data")) {
+			if (options & R_CF_OPTION_SKIP_NSDATA) {
+				next_state = r_cf_parse_state_new (R_CF_STATE_IN_IGNORE);
+			} else {
+				next_state = r_cf_parse_state_new (R_CF_STATE_IN_SCALAR);
+				next_state->value_type = R_CF_DATA;
```
radare: chk for null?

libr/bin/format/xnu/r_cf_dict.c
```diff
@@ -0,0 +1,624 @@
...
+		RCFParseState *state = (RCFParseState*)r_list_get_top (stack);
+		if (state->phase == R_CF_STATE_IN_IGNORE) {
+			break;
+		}
+		if (!content) {
+			content = r_str_new (x.data);
+		} else {
+			content = r_str_append (content, x.data);
+		}
+
```
radare: empty line

libr/bin/format/xnu/r_cf_dict.c
```diff
@@ -0,0 +1,624 @@
...
+	default:
+		break;
+	}
+
+	if (result) {
+		break;
+	}
+  }
+
+  yxml_ret_t r = yxml_eof(&x);
```
radare: missing space, wrong indentation

libr/bin/format/xnu/r_cf_dict.c
```diff
@@ -0,0 +1,624 @@
...
+  if (stack) {
+	  r_list_free (stack);
+  }
+
+  return result;
+}
+
+static RCFParseState *r_cf_parse_state_new(RCFParsePhase phase) {
+	RCFParseState *state = R_NEW0 (RCFParseState);
+	if (!state) {
```
radare: swap condition and return state which is null already

libr/bin/format/xnu/r_cf_dict.c
```diff
@@ -0,0 +1,624 @@
...
+	}
+
+	dict->type = R_CF_DICT;
+	dict->pairs = r_list_newf ((RListFree)&r_cf_key_value_free);
+
+	return dict;
+}
+
+void r_cf_value_dict_free(RCFValueDict *dict) {
+	if (!dict) {
```
radare: use r_return_val_ifi_fail

and this functioni should be static i think

radare: if its not static it should be R_API, and maybe move this into util

libr/bin/format/xnu/scripts/machtraps.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/python
```
radare: uh

libr/bin/p/bin_mach064.c
```diff
@@ -5,11 +5,61 @@
...
 		if (!memcmp (buf, "\xfe\xed\xfa\xcf", 4) ||
 			!memcmp (buf, "\xcf\xfa\xed\xfe", 4)) {
-			return true;
+			RBuffer *tbuf = r_buf_new ();
+			if (!tbuf) {
+				return false;
+			}
+			struct MACH0_(obj_t) *test = NULL;
+			bool regular_mach0 = true;
+			r_buf_set_bytes (tbuf, buf, length);
```
radare: ouch, thats a bad idea, the check byets should just check, not fill the whole buffer, also assuming the buf+len will cover the whole file is not correct. we shold only use RBuffer in rbin plugins

mrmacete: ok reverting for now - then we'll have to find a way to fail here if it's a kernelcache, so we can avoid to use `-F kernelcache` all the time.

mrmacete: is there an amount of bytes it's guaranteed to have here?

radare: nope, you must use the r_buf_read_at to read the data you want , there's an implicit memory copy but it's the only way to be safe in case the backend doesnt provides a linear buffer

PR #13308, opened 2019-03-07
Add function signature here and there and solve (null) in fcn->cc ##visual

radare:
screenshot 2019-03-07 at 18 26 10

as requested by @XVilka

### DIFF
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -2456,22 +2456,35 @@ static int fcn_print_json(RCore *core, RAnalFunction *fcn, PJ *pj) {
 		return -1;
 	}
 	int ebbs = 0;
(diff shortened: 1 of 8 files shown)


XVilka:
Good, just need to be a different color I think, so will be easily distinctive from the 1t line

radare:
graph ttile cant use colors... yet i also prefer a different alignment and such, but that requires more work and didnt wanted to complicate the thing

radare:
maybe i should update the tests... or we should discuss a better title or hide title when not in interactivie

image 2019-03-07 19 03 04

radare:
actually the 2 line thing looks pretty bad, i would go for just showing the function signature and the offset.. and show the n/e/z/m.. in '?'. comments?

screenshot 2019-03-07 at 19 08 06

PR #13311, opened 2019-03-08
Initial Lazy Pass Manager setup

oddcoder:
Will add explanation soon, but this PR is very basic and incomplete in some sense, Still It would be great to have some feedback.
**EDIT**:
There are 2 design approaches I can think of:-
1-  keeping the pass manager as generic as it is and all the data it works with are of type `void *object`, These objects can be either Function or BBblock or that third thing that can be used to analyse the whole binary which I didn't decide yet, it will take some amount of care to write bug free code with everything stored as void *

2- The second approach would be having separate data types for each kind of pass manager, since they all share the same code but different object types, this calls for some template like crazy C pre processor, while this will make it easier to use the pass manager since everything will have its right type, but pass manager will be harder to maintain, and extend. I have seen similar tricks used at least once in Linux page table implementation and it is my least favorite approach. because reading through the code is so hard for newcomers.
**EDIT2**:
Unit tests can be found here https://github.com/radare/radare2-regressions/pull/1738


libr/include/passmanager.h
```diff
@@ -0,0 +1,73 @@
...
+	//TODO  passes can come with their own set of commands
+} Pass;
+
+#ifdef PASSMANAGER_IMPLEMENTATION_FOR_INTERNAL_USE_ONLY
+
+
+typedef struct pm{
+	bool log;
+	PassType t;
+	Sdb *passes; //(k,v) = (Pass->name, PassRunner)
```
ret2libc: Just use `HtPP` please.

ret2libc:
I would really really love to see this feature grow together with unit tests in r2r/unit. It will help design a better API and check that basic functions work (also it will explain how things should be used).

oddcoder:
yup that is the plan from the very beginning

libr/include/passmanager.h
oddcoder: HtPP is for pointer pointer hashtable but the key here is string, Am I missing something?

ret2libc: And what is a string if not a pointer to an array of chars? :D

See this test https://github.com/radare/sdb/blob/master/test/unit/test_hash.c#L251 to understand how to use it.

radare:
related pad https://hackmd.io/_CesDXyMTCGV2H5zjezIHg?both

alvarofe:
I would go for the second approach to impose some rules when using the passmanager and avoid misuses by providing the wrong type. Using always `void*` might end up in wonderland, each plugin using `void *` as they want, etc. To me the second approach is safer than the first one.

alvarofe:
Regarding the code, although is stilll WIP and there is not much to say by now, a few points (I will add some thoughts in the pad to discuss)

* For static declarations follow always the same name convention (camel case iirc the last time I was more active).
* Function whose parameter is a pointer, check for NULL dereference to avoid unexpected crashes. If on the contrary the function is supposed to be called with those arguments already checked, add a comment to mark it explicitly.
* I see a comment about implement a better hash table, is not enough the one shipped with r2/sdb? Or do you mean set HtPPOptions correctly for the passmanager?

oddcoder:
Sure, All comments will be taken into consideration and I will start by doing the template like thing, as for the better hashtable, I meant just better Hashfuntion in the HtPPOptions instead of the plain NULL function for hashing memory addresses.

alvarofe:
Did you rebase against master? why is red? if you consider it's finished let me know to check it out again

oddcoder:
it should be finished, and only 1 target is failing the cov one
image
I am not sure why is it, but I will rebase and push again and hope it works.

alvarofe:
+1 for me to merge - this the foundation of passmanager and yet r2 wouldn't be affected in the analysis because of this. It comes along with unit test. Keep up the good work @oddcoder

libr/anal/passmanager/passmanager.inc.c
```diff
@@ -0,0 +1,199 @@
...
+	}
+	pm->passes = ht_pp_new0 ();
+	//FIXME create better tuned hash table
+	HtPPOptions opt = { 0 };
+	pm->running = ht_pp_new_opt (&opt);
+	return pm;
+}
+
+R_API void PMFN_(set_anal)(PM_(TYPE_NAME) *pm, RAnal *anal) {
+	if (!pm)
```
ret2libc: just use `r_return_if_fail (pm);`

libr/anal/passmanager/passmanager.inc.c
```diff
@@ -0,0 +1,199 @@
...
+		return;
+	pm->parent = anal;
+}
+
+R_API RAnal *PMFN_(get_anal)(PM_(TYPE_NAME) *pm) {
+	if (!pm)
```
ret2libc: `r_return_val_if_fail (pm, NULL);`

libr/anal/passmanager/passmanager.inc.c
```diff
@@ -0,0 +1,199 @@
...
+	ht_pp_foreach (pm->passes, (HtPPForeachCallback)PMFN_(passrunner_free), NULL);
+	ht_pp_free (pm->passes);
+	ht_pp_foreach (pm->running, (HtPPForeachCallback)PMFN_(running_free), NULL);
+	ht_pp_free (pm->running);
+	free (pm);
+}
+
+R_API bool PMFN_(register_pass)(PM_(TYPE_NAME) *pm, PASS_(TYPE_NAME) *pass) {
+	bool pass_exist;
+	if (!pass || !pm) {
```
ret2libc: `r_return_val_if_fail (pass && pm, false);`

radare:
i dont see the purpose of this api yet, and its not yet used anywhere, i would suggest to continue this development in extras as a core plugin, and when its ready and usable merge it in core. too much boilerplate for nothing imho, that way you can experiment and improve that approach without waiting for design discussions here and it will be easier to test (r2pm -i lpm)

oddcoder:
It is true that this API is not used anywhere yet, but that is just me doing 1 thing at a time. The idea is that I favor having everything checked and tested over writing tons of untested features.
The plan was once this got merged aa would be then then diffused where the af* thing would be a module pass that always runs on binary loading (always as configurable and default to yes) and afva would be a function pass that would only run when the results are fetched. This should add some performance improvement and way cleaner data representation for both af and afva. But then again I wanted to make sure that first the API would land into r2 first. Also Because the way the API works it its usage in RAnal would be intrusive to RAnal and would lead to bunch of modifications and I want to avoid having to spend long time dealing with conflicts when merging back into r2 from r2extra.

XVilka:
Agree, I think it might even be merged right after the release, for better testing and adoption in r2 code. Otherwise it will end as SIOL.

radare:
i'm against adding apis that are not used, i dislike this design, i have some questions for many of the methods you are exposing here and the utility of them, and as long as its unused is dead code, there are no tests, allt he new commands that will be using this are not depending on anything to be modified inside r2, so it can be done as a plugin without any problem. i dont see which modifications you will need to do in anal, but i think we need to do omany refactorings still, this passmanager thing should depend on the events that ive been asking several times to add them. and then launch oother analysis (right now we do rcorecmd everywhere because the anal apis need love)

oddcoder:
well there are some tests https://github.com/radare/radare2-regressions/pull/1738, that should be very close to 100% coverage. It is done for 1 pass manager since all of them should behave exactly the same. As for the Plugin thing, The way RAnal would be like in the future after being completely integrated is that RAnalFunction, RBasicBlock and RAnal Data structure would kind of minimalist and would not hold much analysis results (with few exceptions), It would hold the minimal needed data to reconstruct binary. analysis passes (Lets say variable analysis) would store the extra data in its local data structure that is :
A- lazy and cached
B- calculated when only needed and accessed via get resultAPI
and that data structure would not be part of RAnalFunction or RAnal to start with but rather completely local to that pass. This is the kind of intrusive modification I was talking about. The pass manager would manage everything and the only thing that pass it self would need to worry about shall be the analysis only when requested to.
A key point is that this is not only restricted to aa but then this is just the start and expanding the Pass ideology to other implemented analyses would be carried on small steps

oddcoder:
I should also note that I will be doing the aa/aaa conversion to lazy as well just in case I never bought this up before

libr/anal/passmanager/passmanager.inc.c
```diff
@@ -0,0 +1,193 @@
...
+	pr->p = p;
+
+	//FIXME create better tuned hash table
+	HtPPOptions opt = { 0 };
+	pr->passResults = ht_pp_new_opt (&opt);
+	return pr;
+}
+
+static bool PMFN_(invalidatecb) (TYPE *object, const char *passname, const PR_(TYPE_NAME) *pr) {
+	void *x = NULL;
```
radare: define and assign in the same line

radare:
also i would like to see the use of this api, even if as pseudocode.

oddcoder:
An example of this api being used is available at the r2r unit test pull request here https://github.com/radare/radare2-regressions/pull/1738

libr/anal/passmanager/passmanager.inc.c
```diff
@@ -0,0 +1,193 @@
...
+
+R_API RAnal *PMFN_(get_anal)(PM_(TYPE_NAME) *pm) {
+	r_return_val_if_fail (pm, NULL);
+	return pm->parent;
+}
+
+R_API void PMFN_(free)(PM_(TYPE_NAME) *pm) {
+	ht_pp_foreach (pm->passes, (HtPPForeachCallback)PMFN_(passrunner_free), NULL);
+	ht_pp_free (pm->passes);
+	ht_pp_foreach (pm->running, (HtPPForeachCallback)PMFN_(running_free), NULL);
```
ret2libc: this shouldn't be needed if you create the HtPP with the right `freefn` Option. `ht_pp_free` already calls the free function on all elements.

libr/anal/passmanager/passmanager.inc.c
```diff
@@ -0,0 +1,193 @@
...
+	r_return_val_if_fail (pr != NULL, NULL);
+	r_return_val_if_fail (!PMFN_(is_running) (pm->running, pr->p->name, object), NULL);
+	PMFN_(add_running) (pm->running, pr->p->name, object);
+	result = pr->p->run (pm, pr->p, object);
+	ht_pp_insert (pr->passResults, object, result);
+	PMFN_(del_running) (pm->running, pr->p->name, object);
+	return result;
+}
+
+R_API void *PMFN_(get_cached_result)(PM_(TYPE_NAME) *pm, char *passName, TYPE *object) {
```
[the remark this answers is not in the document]
oddcoder: Yup agree, exporting this Function  isn't the best idea will have it removed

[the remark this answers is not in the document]
oddcoder: Oh! I remembered why it is needed, It is not to be used by a pass user, but rather the pass itself, passes are meant to be stateless so they don't keep track of their results. the main use of this function was in the invalidate routines where pass would modify one or more of its cached results. it is not the best way to do it but the simplest.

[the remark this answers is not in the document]
ret2libc: could you remove the `PMFN_` prefix and R_API from this function? it's used only inside the pass itself and in the unit tests. It shouldn't be exported IMHO.

libr/anal/passmanager/passmanager.inc.c
```diff
@@ -0,0 +1,185 @@
...
+	pm->parent = anal;
+}
+
+R_API RAnal *PMFN_(get_anal)(PM_(TYPE_NAME) *pm) {
+	r_return_val_if_fail (pm, NULL);
+	return pm->parent;
+}
+
+R_API void PMFN_(free)(PM_(TYPE_NAME) *pm) {
+	ht_pp_foreach (pm->passes, (HtPPForeachCallback)PMFN_(passrunner_free), NULL);
```
ret2libc: this is not needed. Create `pm->passes` by setting the right `freefn` function and when you call `ht_pp_free` all elements will be automatically freed.

libr/anal/passmanager/passmanager.inc.c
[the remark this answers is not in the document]
oddcoder: well both passes and pass manager users uses exactly the same minimal API, also passes aren't supposed to learn about pass manager internals. I tried to keep the API universal and minimal as possible

libr/anal/passmanager/passmanager.inc.c
oddcoder: fixed

libr/anal/passmanager/passmanager.inc.c
oddcoder: this one is not possible because i am using ready made htpp options and i can't have access to what is inside all is static

libr/anal/meson.build
```diff
@@ -34,6 +34,9 @@ r_anal_sources = [
   'var.c',
   'vtable.c',
   'xrefs.c',
+  'passmanager/functionpm.c',
```
radare: as i said, this is the only module having a subdirectory that is not d/ or p/, and imho the naming should be libr/anal/pm_fcn.c pm_bin.c pm_bb.c ... otherwise maybe just pm/

libr/anal/passmanager/bbpm.c
```diff
@@ -0,0 +1,4 @@
+#include <r_util.h>
+#include <r_passmanager/bbpm.h>
```
radare: thats wrong naming, r_passmanager is not a root module in r2land, so includename cant be like like this

libr/anal/passmanager/passmanager.inc.c
```diff
@@ -0,0 +1,185 @@
...
+	pr->p = p;
+
+	//FIXME create better tuned hash table
+	HtPPOptions opt = { 0 };
+	pr->passResults = ht_pp_new_opt (&opt);
+	return pr;
+}
+
+static bool PMFN_(invalidatecb) (TYPE *object, const char *passname, const PR_(TYPE_NAME) *pr) {
+	r_return_val_if_fail(pr && pr->p, false);
```
radare: missing space before (

libr/anal/passmanager/passmanager.inc.c
```diff
@@ -0,0 +1,185 @@
...
+ */
+typedef struct Pm_(TYPE_NAME) {
+	HtPP *passes;
+	HtPP *running;
+	RAnal *parent;
+} PM_(TYPE_NAME);
+
+static PR_(TYPE_NAME) *PMFN_(passrunner_new)(PASS_(TYPE_NAME) *p) {
+	PR_(TYPE_NAME) *pr = R_NEW0 (PR_(TYPE_NAME));
+	r_return_val_if_fail(pr, NULL);
```
radare: we dont check heap pointers with an asswer, but we do that with function arguments (*p), you should use a normal if() here. the reasoning is:

* the heap can fail at anytime, but we dont want r2 to die if the system is under pressure and cant allocate memory
* if a function of these gets a NULL as argument, thats something it shouldnt happen, so we can check that in developer builds with an assert message, but release builds will directly crash if that happens (but it shouldnt because we tested it in dev builds)

XVilka:
@oddcoder any updates on this?

radare:
abandoned?

oddcoder:
Final exams and thesis defense. Will get back to it when things are over

radare:
Cool :) good to know

XVilka:
Gentle ping.

oddcoder:
I am just little bit busy with stuff IRL, but will get back to it very
soon, sorry for the delay.

XVilka:
How is it going? Any updates?

oddcoder:
probably I will not be working on this pr in the foreseeable future

PR #13334, opened 2019-03-09
Compute file hashes on request via `it` and compare if has changed ##bin

radare:


libr/core/cmd_info.c
```diff
@@ -457,20 +457,48 @@ static int cmd_info(void *data, const char *input) {
...
+						char *s = r_str_prefix_all (old, "- ");
+						r_cons_printf ("%s\n", s);
+						free (s);
+						s = r_str_prefix_all (info->hashes, "+ ");
+						r_cons_printf ("%s\n", s);
+						free (s);
+						break;
+					}
+				}
+				if (input[1] == 'j') {
```
ITAYC0HEN: // "itj"

radare: Just press the pencil button

ITAYC0HEN:
what "it" stands for?

radare:
it was the only letter that was not used for anything, is a bad command, but the pr was stalled there for a month and nobody was reviewing or suggesting any better option. but just complaining because they were unable to get the hash of the file. so at least now it works :P

ITAYC0HEN:
problematic indeed. but that's what we have haha

PR #13335, opened 2019-03-09
Implement UJMP/UCALL color and update themes

oxcabe:


radare:
The only test that is failing is because of the ucall color. Can you update the test in a pr in r2r? Thanks!!

oxcabe:
Working on it right now

PR #13338, opened 2019-03-09
 Add support for UTF console input in Windows

GustavoLCR:
This helps #13305
@a1ext can you help test/review this?

R2R: radare/radare2-regressions#1748


radare:
looks red on windows

a1ext:
> looks red on windows

> [00:08:23] Cannot spawn children with code 3221225477

3221225477 = 0xc0000005 = Access Violation exception has thrown

@GustavoLCR have you tested your code?

libr/cons/dietline.c
```diff
@@ -942,17 +960,23 @@ R_API const char *r_line_readline_cb_win(RLineReadCallback cb, void *user) {
...
+#if UNICODE
+				hClipBoard = GetClipboardData (CF_UNICODETEXT);
+#else
 				hClipBoard = GetClipboardData (CF_TEXT);
+#endif
 				if (hClipBoard) {
 					clipText = GlobalLock (hClipBoard);
 					if (clipText) {
-						I.buffer.length += strlen (clipText);
+						char *txt = r_sys_conv_utf16_to_utf8 (clipText);
```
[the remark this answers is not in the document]
GustavoLCR: if `UNICODE` is not defined this is a macro for r_str_new(), so strdup()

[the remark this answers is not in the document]
a1ext: ok

libr/cons/dietline.c
```diff
@@ -774,8 +789,9 @@ R_API const char *r_line_readline_cb_win(RLineReadCallback cb, void *user) {
 			}
 			break;
 		case 39:// right arrow
-			I.buffer.index = I.buffer.index < I.buffer.length?
-					 I.buffer.index + 1: I.buffer.length;
+			I.buffer.index = I.buffer.index < I.buffer.length
```
radare: this check is probably not enough because its not considering the charsize. also the utf8_charsize should get the length to not override the buffer

libr/util/str.c
```diff
@@ -1882,6 +1882,53 @@ R_API bool r_str_char_fullwidth (const char* s, int left) {
 
 }
 
+/**
+ * Returns size in bytes of the utf8 char
+ * Returns 1 in case of ASCII
+ * str - Pointer to buffer
+ */
+R_API int r_str_utf8_charsize(const char *str) {
```
radare: int len , or check for null at the end, because i think all those strings are null terminated

libr/util/str.c
```diff
@@ -1882,6 +1882,53 @@ R_API bool r_str_char_fullwidth (const char* s, int left) {
...
+	return size;
+}
+
+/**
+ * Returns size in bytes of the previous utf8 char
+ * Returns 1 in case of ASCII
+ * str - Pointer to leading utf8 char
+ * len - Length in bytes of the buffer previous to str
+ */
+R_API int r_str_utf8_charsize_reverse(const char *str, int len) {
```
radare: i would handle if len < 0 perform strlen

radare: but... whats the point on having a reverse charsize here?

radare: len is inverse len, which is kind of confusing and errorprone

libr/util/str.c
GustavoLCR: isn't that what I have below?
`int length = strlen (str);`
`while (size < length)`

libr/util/str.c
GustavoLCR: the reverse charsize is for knowing how many bytes to the left you need to move the index so that it is always on the leading utf8 byte

libr/cons/utf8.c
```diff
@@ -255,7 +255,11 @@ R_API int r_cons_is_utf8() {
 }
 #else
 R_API int r_cons_is_utf8() {
+#if UNICODE && __WINDOWS__
+	return 1;
+#else
 	return 0;
+#endif
```
XVilka: By the way, I think we should determine this property on the fly.
See https://docs.microsoft.com/en-us/windows/console/getconsoleoutputcp and https://docs.microsoft.com/en-us/windows/desktop/api/winnls/nf-winnls-enumsystemcodepagesa

libr/cons/cons.c
```diff
@@ -469,6 +469,10 @@ R_API RCons *r_cons_new() {
 	I.null = 0;
 #if __WINDOWS__
 	I.ansicon = r_sys_getenv ("ANSICON");
+#if UNICODE
+	SetConsoleOutputCP (CP_UTF8);
```
XVilka: Same here, set only if UTF8 is listed in https://docs.microsoft.com/en-us/windows/desktop/api/winnls/nf-winnls-enumsystemcodepagesa

XVilka: Also please check for the error, since it might fail:

> If the function succeeds, the return value is nonzero.
> If the function fails, the return value is zero. To get extended error information, call `GetLastError()`.

libr/cons/cons.c
```diff
@@ -469,6 +469,10 @@ R_API RCons *r_cons_new() {
...
+	SetConsoleCP (CP_UTF8);
```
XVilka: Same here https://docs.microsoft.com/en-us/windows/console/setconsolecp

libr/cons/dietline.c
```diff
@@ -218,12 +217,18 @@ static int r_line_readchar_win(int *vch) { // this function handle the input in
...
+		free (buf);
 		return 0;
 	}
 	if (irInBuf.EventType == KEY_EVENT) {
 		if (irInBuf.Event.KeyEvent.bKeyDown) {
-			if (irInBuf.Event.KeyEvent.uChar.AsciiChar) {
-				*buf = irInBuf.Event.KeyEvent.uChar.AsciiChar;
+			if (irInBuf.Event.KeyEvent.uChar.UnicodeChar) {
+				free (buf);
+#if UNICODE
```
XVilka: I suggest to use function pointer here and set the proper function from the place where UNICODE/ASCII was determined.

GustavoLCR: reworked the r_sys_conv macros to do this

libr/cons/dietline.c
```diff
@@ -1040,19 +1056,25 @@ R_API const char *r_line_readline_cb_win(RLineReadCallback cb, void *user) {
...
+				}
+				while (size) {
+					I.buffer.data[i--] = buf[--size];
 				}
-				I.buffer.data[I.buffer.index] = buf[0];
 			} else {
-				I.buffer.data[I.buffer.length] = buf[0];
-				I.buffer.length++;
+				for (i = 0; i < R_LINE_BUFSIZE - 1 && buf[i]; i++) {
+					I.buffer.data[I.buffer.length + i] = buf[i];
```
a1ext: I think there will be an overflow of `I.buffer.data` in case when it already holds some data.

libr/cons/dietline.c
```diff
@@ -1040,19 +1056,25 @@ R_API const char *r_line_readline_cb_win(RLineReadCallback cb, void *user) {
...
+				}
+				I.buffer.length += strlen (buf);
```
a1ext: Here should be increment on actually copied length, not full length of `buf`

libr/cons/dietline.c
```diff
@@ -1040,19 +1056,25 @@ R_API const char *r_line_readline_cb_win(RLineReadCallback cb, void *user) {
...
 				if (I.buffer.length > (R_LINE_BUFSIZE - 1)) {
 					I.buffer.length--;
```
a1ext: Check is ok, but this code should be like 
```
I.buffer.lenght = R_LINE_BUFZIE - 1;
```

libr/cons/dietline.c
```diff
@@ -1040,19 +1056,25 @@ R_API const char *r_line_readline_cb_win(RLineReadCallback cb, void *user) {
...
 				}
 				I.buffer.data[I.buffer.length] = '\0';
 			}
-			I.buffer.index++;
+			I.buffer.index += strlen (buf);
```
a1ext: Should be incremented with actual copied length

XVilka:
LGTM now, @a1ext if you don't have any objections, I think it is time to merge and extensively test.

a1ext:
> LGTM now, @a1ext if you don't have any objections, I think it is time to merge and extensively test.
> Further fixes can be sent separately.

LGMT, Thanks @GustavoLCR

PR #13339, opened 2019-03-10
Fix r_type_func_guess

pelijah:
We can't rely on `strchr (str, '_')` because even some WinAPI DLL names contain `_` (e.g. WS_32.dll).

### DIFF
--- a/libr/util/ctype.c
+++ b/libr/util/ctype.c
@@ -609,14 +609,8 @@ R_API R_OWN char *r_type_func_guess(Sdb *TDB, R_NONNULL char *func_name) {
 	}
 	// some names are in format module.dll_function_number, try to remove those
 	// also try module.dll_function and function_number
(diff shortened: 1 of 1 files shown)


pelijah:
Hmm nice https://github.com/radare/radare2-regressions/blob/master/unit/test_util.c#L21. I think it is wrong test case.

ret2libc:
I don’t know about the test. I just fixed a mem leak in the test IIRC.

pelijah:
I thought you are the author of it https://github.com/radare/radare2-regressions/pull/1451.

ret2libc:
Ah mmm yes, apparently I am the author. I don't remember exactly how/when/why I did those tests, so if you feel like they are wrong, feel free to correct them (especially because I'm not very familiar with windows)

Maijin:
No worries we fixed this.

PR #13352, opened 2019-03-11
Remove redundant section lookup in get_strings_range

mrmacete:
It seems this loop is an unnecessary leftover from the past when `string_scan_range` didn't initialize string's `vaddr` field.

This improves performance dramatically, especially when there are thousands of sections (like for kernelcache and dyld cache).

### DIFF
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -314,17 +314,7 @@ static void get_strings_range(RBinFile *bf, RList *list, int min, int raw, ut64 
 			return;
 		}
 	}
-	if (string_scan_range (list, bf, min, from, to, -1, raw) < 0) {
-		return;
-	}
-	if (bf->o) {
-		r_list_foreach (list, it, ptr) {
-			RBinSection *s = r_bin_get_section_at (bf->o, ptr->paddr, false);
-			if (s) {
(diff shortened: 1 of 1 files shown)


radare:
Screenshot 2019-03-11 at 22 15 38

mrmacete:
yup, broke stuff. digging

mrmacete:
uhm... `iz` is correct, `izz` is wrong...
```
hugrid:new ftamagni$ r2 ../bins/pe/ioli/w32/crackme0x00.exe
 -- I am Pentium of Borg. Division is futile. You will be approximated.
[0x00401260]> izz~Password:
086 0x00002619 0x00002619  10  11 (.rdata) ascii Password:
[0x00401260]> iz~Password:
001 0x00002619 0x00404019  10  11 (.rdata) ascii Password:
[0x00401260]>
```

mrmacete:
yeah so now i figured out why that loop was there: because when doing `izz` it starts from address `0` and goes down to the end of the file, across all sections. so that's the only case in which we need to actually re-assign `vaddr` because the scan function is unaware of sections - let me see if i can cook something

mrmacete:
rebased

radare:
seems like your patch is fixing one test and the XX one is a bad test, should be updated

Screenshot 2019-03-12 at 01 34 24

PR #13353, opened 2019-03-11
Implement load_buffer/check_buffer in RBinPlugins

ret2libc:
Some of the bin plugins in https://github.com/radare/radare2/issues/13349

### DIFF
--- a/libr/bin/p/bin_any.c
+++ b/libr/bin/p/bin_any.c
@@ -6,40 +6,35 @@
 #include <r_bin.h>
 #include <r_magic.h>
 
-static char *get_filetype(RBinFile *bf) {
-	ut8 buf[4096] = {
-		0
-	};
+static char *get_filetype(RBuffer *obj) {
+	ut8 buf[4096] = { 0 };
 	char *res = NULL;
-	RMagic *ck;
-	if (!bf) {
+	RMagic *ck = r_magic_new (0);
+	if (!ck) {
 		return NULL;
 	}
-	ck = r_magic_new (0);
-	if (ck && bf && bf->buf) {
-		const char *tmp = NULL;
-		// TODO: dir.magic not honored here
-		char *pfx = r_str_newf (R_JOIN_2_PATHS ("%s", R2_SDB_MAGIC), r_sys_prefix (NULL));
-		r_magic_load (ck, R2_SDB_MAGIC);
-		r_buf_read_at (bf->buf, 0, buf, sizeof (buf));
-		tmp = r_magic_buffer (ck, buf, sizeof (buf));
-		if (tmp) {
-			res = strdup (tmp);
-		}
-		free (pfx);
+
+	const char *tmp = NULL;
+	// TODO: dir.magic not honored here
+	r_magic_load (ck, R2_SDB_MAGIC);
+	r_buf_read_at (obj, 0, buf, sizeof (buf));
+	tmp = r_magic_buffer (ck, buf, sizeof (buf));
+	if (tmp) {
+		res = strdup (tmp);
 	}
 	r_magic_free (ck);
 	return res;
 }
 
 static RBinInfo *info(RBinFile *bf) {
+	RBuffer *any_obj = bf->o->bin_obj;
 	RBinInfo *ret = R_NEW0 (RBinInfo);
 	if (!ret) {
 		return NULL;
 	}
 	ret->lang = "";
 	ret->file = bf->file? strdup (bf->file): NULL;
-	ret->type = get_filetype (bf);
+	ret->type = get_filetype (any_obj);
 	ret->has_pi = 0;
 	ret->has_canary = 0;
 	ret->has_retguard = -1;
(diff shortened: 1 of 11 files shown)


ret2libc:
Note that `check_buffer` is still not used, but it will be soon IMO, since we use load_buffer there's no reason to keep check_bytes.

radare:
cool

ret2libc:
I'd say let's start by reviewing/merging this if travis is green. I'll do the rest of the plugins soon.

radare:
sgtm

ret2libc:
Also, I'm rewriting the plugins to just  call `r_buf_ref` on the buffer passed to load_buffer. Does that sound like a problem to you? This way, we don't even create a new RBuffer, but we just use the one passed to load_buffer. The alternative may be to use a r_buf_new_with_bufref, which creates a new buffer structure but reuses the same data as the original buffer. What do you think?

Using r_buf_ref:
- pros:
  - no additional memory used, at all
- cons:
  - one single "buffer state", so if something changes the buffer state outside of the binplugin, it may affect the bin plugin as well, because the state is the same

Using r_buf_new_with_bufref:
- pros:
  - no additional memory for the data
  - separate buffer state for the bin plugin (e.g. seek, etc.)
- cons:
  - a little (very little, i'd say) of extra memory for the separate buffer state

radare:
yes, that was expected to happen too. copying buffers is not necessary

libr/bin/p/bin_any.c
```diff
@@ -6,40 +6,38 @@
...
-static char *get_filetype(RBinFile *bf) {
-	ut8 buf[4096] = {
-		0
-	};
+static char *get_filetype(RBuffer *obj) {
+	ut8 buf[4096] = { 0 };
 	char *res = NULL;
 	RMagic *ck;
-	if (!bf) {
+	ck = r_magic_new (0);
```
radare: define and assign in the same line

libr/bin/p/bin_any.c
```diff
@@ -6,40 +6,38 @@
...
-		r_buf_read_at (bf->buf, 0, buf, sizeof (buf));
-		tmp = r_magic_buffer (ck, buf, sizeof (buf));
-		if (tmp) {
-			res = strdup (tmp);
-		}
-		free (pfx);
+
+	const char *tmp = NULL;
+	// TODO: dir.magic not honored here
+	char *pfx = r_str_newf (R_JOIN_2_PATHS ("%s", R2_SDB_MAGIC), r_sys_prefix (NULL));
```
radare: wat, pfx is not used at all

ret2libc: apparently it never was :/

libr/bin/p/bin_any.c
```diff
@@ -6,40 +6,38 @@
...
 static RBinInfo *info(RBinFile *bf) {
+	RBuffer *any_obj = bf->o->bin_obj;
 	RBinInfo *ret = R_NEW0 (RBinInfo);
 	if (!ret) {
 		return NULL;
 	}
 	ret->lang = "";
 	ret->file = bf->file? strdup (bf->file): NULL;
-	ret->type = get_filetype (bf);
+	ret->type = get_filetype (any_obj);
```
radare: why call it any_obj instead of bin_obj

ret2libc: dunno, it is the `any` plugin, so `any_obj` :D

libr/bin/p/bin_art.c
```diff
@@ -130,8 +131,17 @@ static RBinInfo *info(RBinFile *bf) {
 	return ret;
 }
 
+static bool check_buffer(RBuffer *buf) {
+	char tmp[4];
+	int r = r_buf_read_at (buf, 0, (ut8 *)tmp, sizeof (tmp));
+	return r == 4 && !strncmp (tmp, "art\n", 4);
+}
+
 static bool check_bytes(const ut8 *buf, ut64 length) {
```
radare: this must be removed

libr/bin/p/bin_bf.c
```diff
@@ -64,32 +61,43 @@ static RBinInfo *info(RBinFile *bf) {
...
-				break;
-			default:
-				is_bf = 0;
-			}
+static bool check_buffer(RBuffer *buf) {
+	r_return_val_if_fail (buf, false);
+
+	ut8 tmp[16];
+	int read_length = r_buf_read_at (buf, 0, tmp, sizeof (tmp));
+	if (read_length <= 0) {
```
radare: what about != 16

radare: well != sizeof(tmp).

libr/bin/p/bin_art.c
ret2libc: not now, because the code still uses check_bytes. I plan to remove check_bytes in a future PR, but for now I'm making sure check_bytes rely on check_buffer.

libr/bin/p/bin_bf.c
ret2libc: I'm trying to keep the original semantic. And originally the check was `length > 0`, it wasn't really a requirement that 16 bytes were read.

ret2libc:
Ok let's wait for Travis. I think if green it can be merged.

PR #13359, opened 2019-03-12
Further usage of thread debug

devnexen:


radare:
did u fixed the warnings that appear in travis?

devnexen:
Found out on Linux _GNU_SOURCE is not enough to enable it, it is also a matter of glibc version. On BSD, headers are enough.

libr/core/project.c
```diff
@@ -354,6 +354,15 @@ R_API RThread *r_core_project_load_bg(RCore *core, const char *prjName, const ch
 	ps->rcPath = strdup (rcPath);
 	RThread *th = r_th_new (projectLoadBackground, ps, false);
 	r_th_start (th, true);
+#if 0
```
radare: can we remove this?

devnexen: sure ... in fact was not sure it would go without.

PR #13365, opened 2019-03-13
Rbin remove old loads

ret2libc:
Please have a double look at bin_dyldcache, because it was doing some weird things that I removed (e.g. r_buf_new_with_io and check_bytes inside load_buffer function.)


radare:
cc @mrmacete

mrmacete:
dyldcache LGTM 👍

PR #13366, opened 2019-03-13
bin_art: avoid double free

ret2libc:
ao->kv's ownership is moved to RBin through the get_sdb method of
RBinPlugin, thus the plugins should not free it themselves.


ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1742

PR #13375, opened 2019-03-14
Consider the map sorting when two map_events are the same.

ret2libc:
When two maps start at the same address, it's important to consider the
initial priority of the maps, which is given by the `id` field.

r2r PR: https://github.com/radare/radare2-regressions/pull/1745

### DIFF
--- a/libr/io/map.c
+++ b/libr/io/map.c
@@ -29,9 +29,12 @@ static int _cmp_map_event(const void *a_, const void *b_) {
 		return addr0 < addr1? -1: 1;
 	}
 	if (a->is_to != b->is_to) {
-		return a->is_to - b->is_to;
+		return !a->is_to? -1: 1;
 	}
-	return a->id - b->id;
+	if (a->id != b->id) {
+		return a->id < b->id? -1: 1;
+	}
+	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


ret2libc:
cc @condret

libr/io/map.c
```diff
@@ -26,9 +26,12 @@ static int _cmp_map_event(const void *a_, const void *b_) {
 	struct map_event_t *a = (void *)a_, *b = (void *)b_;
 	ut64 addr0 = a->addr - a->is_to, addr1 = b->addr - b->is_to;
 	if (addr0 != addr1) {
-		return addr0 < addr1 ? -1 : 1;
+		return addr0 < addr1? -1: 1;
 	}
-	return a->is_to - b->is_to;
```
radare: what about return -1, 0 or 1 instead of abusing the integer overflows here

ret2libc: ok

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1745

ret2libc:
I think this can be merged.

radare:
Me too but i didnt had time to test it yet.

radare:
Confirmed the fix , thanks!

PR #13382, opened 2019-03-15
Add support for TBH jmptbl ##anal

radare:


radare:
Merging for now, agreed with @mrmacete to do some more testing and add tests later on this

PR #13392, opened 2019-03-16
Fix issues when using file:// and ood together ##debug

radare:


radare:
Fixes https://github.com/radare/radare2/issues/13391

but this is kind of a workaround, ideally io and bin should differentiate uri and name properly because htats a bit inconsistent

libr/core/cmd_open.c
```diff
@@ -888,6 +886,13 @@ R_API void r_core_file_reopen_debug(RCore *core, const char *args) {
 		(void)r_core_cmd0 (core, "oo");
 		return;
 	}
+	bool is_file_uri = r_str_startswith (binpath, "file://");
+	if (is_file_uri) {
+		eprintf ("Trimming down the 'file://' URI handler.\n");
```
xarkes: Is it really useful to add a warning message? A comment is enough imho, a developer might care while the end user probably not.

libr/main/radare2.c
```diff
@@ -1119,6 +1119,7 @@ R_API int r_main_radare2(int argc, char **argv) {
 								if (r.file && iod && (iod->fd == r.file->fd) && iod->name) {
 									filepath = iod->name;
 								}
+								// XXX if using file:// iod->uri and iod->name are swapped :(
```
xarkes: Can't you swap them?

libr/core/cmd_open.c
radare: uhm yeah better to remove this

libr/main/radare2.c
radare: its fixed

PR #13399, opened 2019-03-17
 Show r_anal_fcn_add_bb() truncation warning only if verbose

kazarmy:


libr/anal/fcn.c
```diff
@@ -2000,6 +2000,10 @@ R_API bool r_anal_fcn_add_bb(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 si
 						fcn->addr, addr);
 			}
 			return false;
+		} else if (bb->size != size && (r_cons_is_interactive () || anal->verbose)) {
```
radare: why cons.interactive should matter here? imhoh this must be depending only in anal->verbose

radare:
i dont think we want to have those warnings, its more kind of debugging purposes, than really an issue. thats why i removed the warning message

libr/anal/fcn.c
kazarmy: well it's possible that the user makes a mistake while typing in an `afb+` command

kazarmy:
i'm curious as to why you say it's noisy ... couldn't repro with `a2f`

libr/anal/fcn.c
kazarmy: ok it depends only on anal->verbose now cd41dc00526a18c4ba3d3c9cbcf11926fdcd8f28

kazarmy:
yes it might be needed to debug tests ... `afb+` as i see it should give you a bb exactly the size requested, and not truncate it prematurely due to a jump or return ... perhaps it's best if the size param is removed from `afb+`

kazarmy:
but then again removing the size param might prevent manual adding of basic blocks that are larger than `anal.bb.maxsize`

radare:
Well that was happening with “aaa”.

radare:
Aaa in awk was spitting lots of of them

kazarmy:
Still can't repro :( ... anyway i'm checking only `anal->verbose` like you asked

radare:
With the mac’s awk

kazarmy:
Have to send the binary over ... don't think Homebrew has a binary repository

radare:
Its not a 3rd party pkg is the one that comes with the system. I will send it to u as soon as i arrive at home very late today

kazarmy:
> ... perhaps it's best if the size param is removed from afb+

I suppose this goal is achievable after fixing analysis when debugging and with judicious use of `aht`, so I'm closing this.

PR #13402, opened 2019-03-17
Fix asm.stackptr when disassembling in the middle of a function ##disasm

radare:


radare:
Funnily there's a bb->stackptr already.. so if its set properly at analysis time we shold honor it and add an afb subcommand to get and set this value

PR #13417, opened 2019-03-18
Fix invalid instruction warning

kazarmy:


radare:
how is this fixing the warning?

radare:
should honor anal.verbose, and whats wrong with showing the length?

kazarmy:
> whats wrong with showing the length?

well if the instruction is invalid, we'll never know for sure how long the instruction should be, and so showing the length is pointless

kazarmy:
... showing any length is pointless with regard to the invalid instruction

radare:
well it's useful as a hint because we were using fixed size buffers before.. not sure if that's still the case

kazarmy:
> should honor anal.verbose

ok 772e63280695788126d82b743b23d2aedfabeb1c ... if anal.verbose is honored, i think it's better that the filtering is removed ... might want to know where all the false positives are

radare:
yeah sounds reasonable

PR #13423, opened 2019-03-18
Improve arm TBH jump table ##anal

mrmacete:
- store cmpval and cmpreg in every basic block
- lookup cmpval from predecessors considering TBH’s index register

test here: https://github.com/radare/radare2-regressions/pull/1750

### DIFF
--- a/libr/anal/p/anal_arm_cs.c
+++ b/libr/anal/p/anal_arm_cs.c
@@ -2632,11 +2632,13 @@ jmp $$ + 4 + ( [delta] * 2 )
 		op->type = R_ANAL_OP_TYPE_UJMP;
 		op->cycles = 2;
 		op->ptrsize = 2;
+		op->ireg = r_str_get (cs_reg_name(handle, INSOP (0).mem.index));
 		break;
 	case ARM_INS_TBB: // byte jump table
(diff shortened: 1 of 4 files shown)


radare:
❤️

libr/anal/p/anal_arm_cs.c
```diff
@@ -2632,11 +2632,13 @@ jmp $$ + 4 + ( [delta] * 2 )
 		op->type = R_ANAL_OP_TYPE_UJMP;
 		op->cycles = 2;
 		op->ptrsize = 2;
+		op->ireg = r_str_get (cs_reg_name(handle, INSOP (0).mem.index));
 		break;
 	case ARM_INS_TBB: // byte jump table
 		op->type = R_ANAL_OP_TYPE_UJMP;
 		op->cycles = 2;
 		op->ptrsize = 1;
+		op->ireg = r_str_get (cs_reg_name(handle, INSOP (0).mem.index));
```
radare: Space before (

mrmacete:
well, no actually the lookup on predecessor is performed only just before walking the table - actually only when there's a TBH instruction right now (and that's already looking up stuff, so already o(n) not getting much worse than that)

PR #13426, opened 2019-03-18
[13363] Remove URI handler before looking for auxillary script

dutt:
When using debugger the file assumed path for the auxillary script
was dbg://...binary path.r2, without debugger it's just ...binary path.r2.

This patch cuts out the 'dbg://' prefix, if found, before looking for the
auxillary script.

Fix for https://github.com/radare/radare2/issues/13363

### DIFF
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -1328,8 +1328,14 @@ R_API int r_main_radare2(int argc, char **argv) {
 		r_flag_space_set (r.flags, NULL);
 		/* load <file>.r2 */
 		{
-			char f[128];
+			char f[256];
 			snprintf (f, sizeof (f), "%s.r2", pfile);
+			if(strncmp(f, "dbg://", 6) == 0) {
+				// cut out the uri handler part of the path
+				char tmp[256];
+				strncpy (tmp, f + 6, 250);
+				strncpy (f, tmp, 250);
+			}
 			if (r_file_exists (f)) {
 				// TODO: should 'q' unset the interactive bit?
 				bool isInteractive = r_cons_is_interactive ();
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -1329,20 +1329,16 @@ R_API int r_main_radare2(int argc, char **argv) {
 		/* load <file>.r2 */
 		{
 			char* f = r_str_newf ("%s.r2", pfile);
-			const int f_len = strlen (f);
-			const char *prefix = "dbg://";
-			const int prefix_len = strlen (prefix);
-			if (r_str_startswith (f, prefix)) {
-				memmove (f, f + prefix_len, f_len - prefix_len + 1);
-			}
-			if (r_file_exists (f)) {
+			const char *uri_splitter = strstr (f, "://");
+			const char *path = uri_splitter? uri_splitter + 3: f;
+			if (r_file_exists (path)) {
 				// TODO: should 'q' unset the interactive bit?
 				bool isInteractive = r_cons_is_interactive ();
-				if (isInteractive && r_cons_yesno ('n', "Do you want to run the '%s' script? (y/N) ", f)) {
-					r_core_cmd_file (&r, f);
+				if (isInteractive && r_cons_yesno ('n', "Do you want to run the '%s' script? (y/N) ", path)) {
+					r_core_cmd_file (&r, path);
 				}
 			}
-			free(f);
+			free (f);
 		}
 	} else {
 		r_core_block_read (&r);
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -1328,13 +1328,13 @@ R_API int r_main_radare2(int argc, char **argv) {
 		r_flag_space_set (r.flags, NULL);
 		/* load <file>.r2 */
 		{
-			char f[256];
-			snprintf (f, sizeof (f), "%s.r2", pfile);
-			if(strncmp(f, "dbg://", 6) == 0) {
-				// cut out the uri handler part of the path
-				char tmp[256];
-				strncpy (tmp, f + 6, 250);
-				strncpy (f, tmp, 250);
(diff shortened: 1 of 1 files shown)


dutt:
Fixes https://github.com/radare/radare2/issues/13363

libr/main/radare2.c
```diff
@@ -1328,8 +1328,14 @@ R_API int r_main_radare2(int argc, char **argv) {
 		r_flag_space_set (r.flags, NULL);
 		/* load <file>.r2 */
 		{
-			char f[128];
+			char f[256];
 			snprintf (f, sizeof (f), "%s.r2", pfile);
+			if(strncmp(f, "dbg://", 6) == 0) {
```
radare: add missing space before each ( and use ! instead of == 0

libr/main/radare2.c
```diff
@@ -1328,8 +1328,14 @@ R_API int r_main_radare2(int argc, char **argv) {
...
+				// cut out the uri handler part of the path
+				char tmp[256];
+				strncpy (tmp, f + 6, 250);
```
radare: you can use memmove instead of double strncpy

radare: also, use sizeof(x) instead of hardcoded constants here and there

libr/main/radare2.c
```diff
@@ -1328,8 +1328,14 @@ R_API int r_main_radare2(int argc, char **argv) {
 		r_flag_space_set (r.flags, NULL);
 		/* load <file>.r2 */
 		{
-			char f[128];
+			char f[256];
 			snprintf (f, sizeof (f), "%s.r2", pfile);
```
radare: we should use r_str_newf () instead here, having a 256 char limit here is not a good idea

libr/main/radare2.c
radare: imho its better to use f+=strlen ("dbg://")

radare: but keep a pointer to the orgiinal buffer to free it

libr/main/radare2.c
radare: you can also use if (r_str_startswith (f, "dbg://")) { its more readable

radare:
ping

dutt:
> ping

Yea, I'll fix it. Just busy with work, the dog other things for a bit :) planning to push a new version tomorrow.

libr/main/radare2.c
```diff
@@ -1328,8 +1328,14 @@ R_API int r_main_radare2(int argc, char **argv) {
 		r_flag_space_set (r.flags, NULL);
 		/* load <file>.r2 */
 		{
-			char f[128];
-			snprintf (f, sizeof (f), "%s.r2", pfile);
+			char* f = r_str_newf("%s.r2", pfile);
```
radare: f is never freed, so its leaking, also, missing sspace before (

libr/main/radare2.c
```diff
@@ -1328,8 +1328,14 @@ R_API int r_main_radare2(int argc, char **argv) {
...
+			int f_len = strlen (f);
+			const char* prefix = "dbg://";
+			int prefix_len = strlen (prefix);
+			if (r_str_startswith (f, prefix)) {
+				memmove (f, f + prefix_len, f_len - prefix_len);
```
radare: if you memmove one more byte you will get the trailing 0 for free

libr/main/radare2.c
```diff
@@ -1328,8 +1328,14 @@ R_API int r_main_radare2(int argc, char **argv) {
 		r_flag_space_set (r.flags, NULL);
 		/* load <file>.r2 */
 		{
-			char f[128];
-			snprintf (f, sizeof (f), "%s.r2", pfile);
+			char* f = r_str_newf("%s.r2", pfile);
+			int f_len = strlen (f);
```
radare: thihs is const

libr/main/radare2.c
```diff
@@ -1328,8 +1328,14 @@ R_API int r_main_radare2(int argc, char **argv) {
...
+			const char* prefix = "dbg://";
+			int prefix_len = strlen (prefix);
```
radare: thihs is const too

libr/main/radare2.c
dutt: It seems the syntax checker script is missing those.

libr/main/radare2.c
```diff
@@ -1328,15 +1328,21 @@ R_API int r_main_radare2(int argc, char **argv) {
 		r_flag_space_set (r.flags, NULL);
 		/* load <file>.r2 */
 		{
-			char f[128];
-			snprintf (f, sizeof (f), "%s.r2", pfile);
+			char* f = r_str_newf ("%s.r2", pfile);
+			const int f_len = strlen (f);
+			const char *prefix = "dbg://";
```
radare: i think it will be better to use : 

```
const char *peekaboo = strstr (pfile, "://");
char *path = peekaboo? peekaboo +3: pfile;
if (r_file_exisits(path)) { ...
```

radare: so we handle all the possible prefixes (may be useful for iihex:// mmap:// and others despite have some undesired behaviours on others, but as long asi ts asking before executing it should be fine

libr/main/radare2.c
```diff
@@ -1328,15 +1328,21 @@ R_API int r_main_radare2(int argc, char **argv) {
...
+				memmove (f, f + prefix_len, f_len - prefix_len + 1);
+			}
 			if (r_file_exists (f)) {
 				// TODO: should 'q' unset the interactive bit?
 				bool isInteractive = r_cons_is_interactive ();
 				if (isInteractive && r_cons_yesno ('n', "Do you want to run the '%s' script? (y/N) ", f)) {
 					r_core_cmd_file (&r, f);
 				}
 			}
+			free(f);
```
radare: missingi space before (

radare:
ping, please update the pr and ill merge.

PR #13428, opened 2019-03-18
FreeBSD get thread name support only from 12.x flavor

devnexen:


unixfreaxjp:
@radare do not merge this code yet

devnexen:
This is precisely to address this in fact ...

PR #13431, opened 2019-03-19
Improve Makefile on macOS

king6cong:
`OS`  can have already be set to other values on user machines, `OS=` fits better here.

### DIFF
--- a/shlr/sdb/config.mk
+++ b/shlr/sdb/config.mk
@@ -52,7 +52,7 @@ HAVE_VALA=#$(shell valac --version 2> /dev/null)
 # This is hacky
 HOST_CC?=gcc
 RANLIB?=ranlib
-OS?=$(shell uname)
+OS=$(shell uname)
 OSTYPE?=$(shell uname -s)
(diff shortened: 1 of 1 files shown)


radare:
Sdb is a separate project dont fix the issues in here because it will break after sync

radare:
i have pushed the fix in the sdb repo too

radare:
And btw, i use macos as main dev platform and travis is also testing this and nobody have any issue building r2 from git. So dunno whats the problem you are trying to fix in here

radare:
uhm, this patch is wrong. is breaking the mingw build, which is not used in r2, but stll suposed to be supported in sdb

king6cong:
> And btw, i use macos as main dev platform and travis is also testing this and nobody have any issue building r2 from git. So dunno whats the problem you are trying to fix in here

radare:
Ok i see

radare:
can you send a PR in sdb to update this, otherwise when we merge back it will regress again

PR #13436, opened 2019-03-19
 bin.str{purge,filter} -> bin.str.{purge,filter}

kazarmy:
For https://github.com/radare/radare2/pull/13180#issuecomment-466766072.


radare:
👍

PR #13441, opened 2019-03-20
Fixes #1179 - `itj` hashes calculation command has fixed

a1ext:
- changed the output json keys, now there will be `md5` and `sha1` instead of `values`
- small refactoring on hashes storage
- removed obsolete `file.sha1` usage

image

Related to radareorg/cutter#1179


radare:
red

PR #13443, opened 2019-03-20
Simplify WASM code by avoiding duplications

ret2libc:
I was doing another refactoring for RBuf and this wasm code was using it sooo much, so I ended up simplifying it a bit to avoid duplicated code.

### DIFF
--- a/libr/bin/format/wasm/wasm.c
+++ b/libr/bin/format/wasm/wasm.c
@@ -7,51 +7,45 @@
 
 #include "wasm.h"
 
+typedef size_t (*ConsumeFcn)(const ut8 *p, const ut8 *max, ut32 *out_value);
+typedef void *(*ParseEntryFcn)(RBuffer *b, ut64 max);
+
 // RBuffer consume functions
-static size_t consume_u32_r(RBuffer *b, ut64 max, ut32 *out) {
+static ut32 consume_r(RBuffer *b, ut64 max, size_t *n_out, ConsumeFcn consume_fcn) {
 	size_t n;
 	ut32 tmp;
-	if (!b || !b->buf || max >= r_buf_size (b) || r_buf_tell (b) > max) {
+	ut64 cur = r_buf_tell (b);
+	if (!b || max >= r_buf_size (b) || cur > max) {
 		return 0;
 	}
-	if (!(n = read_u32_leb128 (&b->buf[r_buf_tell (b)], &b->buf[max + 1], &tmp))) {
+	// 16 bytes are enough to store 128bits values
+	ut8 *buf = R_NEWS (ut8, 16);
+	if (!buf) {
 		return 0;
 	}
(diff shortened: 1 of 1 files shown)


libr/bin/format/wasm/wasm.c
```diff
@@ -7,51 +7,45 @@
 
 #include "wasm.h"
 
+typedef size_t (*ConsumeFcn)(const ut8 *p, const ut8 *max, ut32 *out_value);
```
wargio: i'm sure we will not always have `ut32*` pointer there.

libr/bin/format/wasm/wasm.c
```diff
@@ -266,527 +243,368 @@ static RList *r_bin_wasm_get_type_entries(RBinWasmObj *bin, RBinWasmSection *sec
...
+	if (!(consume_u32_r (b, max, &ptr->size))) {
+		goto beach;
 	}
-	return ret;
+	ptr->data = r_buf_tell (b);
+	r_buf_seek (b, ptr->size, R_IO_SEEK_CUR);
+	return ptr;
+
 beach:
-	eprintf ("[wasm] error: beach symtab entries\n");
```
wargio: Please keep `beach` errors.

ret2libc: I kept them in a single place at https://github.com/radare/radare2/pull/13443/files#diff-bc8cad383202ffde95806a51d3a5c748R259 . Is that ok?

libr/bin/format/wasm/wasm.c
ret2libc: Mmm I guess we'll take care of that when that's needed. Right now we don't need to complicate it. In any case, it will be easy to change if you need, for example, ut64 instead of ut32.

libr/bin/format/wasm/wasm.c
```diff
@@ -7,51 +7,45 @@
...
+typedef void *(*ParseEntryFcn)(RBuffer *b, ut64 max);
+
 // RBuffer consume functions
-static size_t consume_u32_r(RBuffer *b, ut64 max, ut32 *out) {
+static ut32 consume_r(RBuffer *b, ut64 max, size_t *n_out, ConsumeFcn consume_fcn) {
 	size_t n;
 	ut32 tmp;
-	if (!b || !b->buf || max >= r_buf_size (b) || r_buf_tell (b) > max) {
+	ut64 cur = r_buf_tell (b);
+	if (!b || max >= r_buf_size (b) || cur > max) {
```
radare: what about r_return..

libr/bin/format/wasm/wasm.c
wargio: ok.

libr/bin/format/wasm/wasm.c
wargio: ok

PR #13454, opened 2019-03-21
WIP: Support asm.pseudo=1 in pad command

wasamasa:
WIP because actually using this in an ASan build triggers crashes. This can be reproduced using any of these, showing that it's not the fault of the newly added chip8-parse plugin.  The `ASAN_OPTIONS` setting is to supress false positives for memory leaks.

```
ASAN_OPTIONS=detect_leaks=0 r2 -a x86 -b 32 -e asm.pseudo=1 -q -c 'pad 7400' -
ASAN_OPTIONS=detect_leaks=0 r2 -a chip8 -q -e asm.pseudo=1 -c 'pad c97f' -
```

See also https://github.com/radare/radare2/issues/13429

### DIFF
--- a/libr/asm/asm.c
+++ b/libr/asm/asm.c
@@ -640,7 +640,7 @@ R_API RAsmCode* r_asm_mdisassemble(RAsm *a, const ut8 *buf, int len) {
 	return acode;
 }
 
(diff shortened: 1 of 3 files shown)


wasamasa:
Ouch. I hoped to have a discussion whether parse plugins need a rewrite using strbuf to fix this issue.

radare:
Use irc/tg to discuss

PR #13456, opened 2019-03-21
Change 'sanitiz' position from bininfo to be according alphabetical order

RomeuG:
Hey, first contribution here. Very small one. 

When using the command `iI` in radare2 or `-I` in rabin2, `sanitiz` was before `class` and it seems that alphabetical order was intended, so I changed its position.

Thank you.


radare:
update the tests

PR #13471, opened 2019-03-23
fix printing consecutive type links in pd

condret:


radare:
Screenshot 2019-03-23 at 12 28 43

PR #13477, opened 2019-03-23
Remove obf #11988 ##Rbin

Maijin:
`o` already open and load the rbin info, obf is obsolete.


radare:
Its not dupe according to the help message

Maijin:
Yep according to help message, but o seem to already open and read the bininfo. Also in your issue you said obf must be killed. What do you propose?

radare:
obf loads the bininfo without opening a new file inside r2. well it does, but it closes it when done. that useful (and important) when debugging, otherweise you will be reading file instead of process memory

Screenshot 2019-03-24 at 20 03 02

radare:
@Maijin can we close this? any comments frmo someone else here?

Maijin:
yes yes sorry.

PR #13478, opened 2019-03-23
Fix anal/cc args to start from 0 instead of 1

### LINKED ISSUE #13077: anal/cc args starting by 1 instead of 0 wtf
screenshot 2019-02-13 at 13 03 52

sivaramaaa:
Fix #13077


radare:
Red

Maijin:
Now green 👍

libr/core/cmd_anal.c
```diff
@@ -2721,17 +2721,13 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
 			// if json_str initialize to NULL, it's possible for afcrj to output a (NULL)
 			// subvec_str and json_str should be valid until exiting this code block
 			bool json = input[3] == 'j'? true: false;
-			for (i = 0; i <= 11; i++) {
-				if (i == 0) {
+			for (i = -1; i <= 10; i++) {
+				if (i == -1) {
```
radare: uh? cc.-1.ret ? that looks bad to me

sivaramaaa: No it's not cc.-1 , it's cc.<calling convention name>.ret

But -1 is just used as a special index in that loop so that we don't have to dupe a huge portion of code !

Issue #13229, opened 2019-03-02
av changes


radare:
```
[0x1000011ec]> av?
Usage: av[?jr*]   C++ vtables and RTTI
| av                search for vtables in data sections and show results
| avj               like av, but as json
| av*               like av, but as r2 commands
| avr[j@addr]       try to parse RTTI at vtable addr (see anal.cpp.abi)
| avra[j]           search for vtables and try to parse RTTI at each of them
| avrr              recover class info from all findable RTTI (see aC)
| avrD [classname]  demangle a class name from RTTI
[0x1000011ec]> aC
Usage: aC[e] [addr-of-call] # analyze call args (aCe does esil emulation with abte)
[0x1000011ec]>
```

note the following:

* avra should be renamed to aavt (analyze all virtual tables)
* the aC command reference is wrong, it is ac
* avrr may be renamed to aavr maybe :?
* Is avra run from aaa? may we want to use the bin info for this?

thestr4ng3r:
"aavt (analyze all virtual tables)" is wrong, the command is about rtti, not just vtables, also it doesn't really analyze, it just reads data and prints it. Does that fit under aa? avrr does in any case though.

radare:
then i misunderstood. maybe is avrr the one that must be moved into aa? if those vtables thing are just printing commands why are they under analysis? PD: never used that

thestr4ng3r:
Well where would it fit other than anal? It's partially heuristic search and partially parsing data structures defined by C++ ABI.

radare:
Is av* executed from aaa? Which are the requirements for it to work? Can you fix the aC help message?

thestr4ng3r:
Right now there is nothing in aaa, but avrr is something could certainly be added. The requirement for it are xrefs for the vtable search (aar has to be executed before) and rtti in the binary. Maybe there should also be a fallback when there is no rtti, so it just detects vtables and adds classes for them.

Yes, I'll fix the help

radare:
Cc @maijin

Pls add this in aaa when it matters

Maijin:
@thestr4ng3r Do you have an example of binary in r2r I can use to test the aaa integration?

thestr4ng3r:
Not in r2r, but you can try with this: [cpp_sample.exe.gz](https://github.com/radare/radare2/files/2999628/cpp_sample.exe.gz)

thestr4ng3r:
Here's also the pdb for it: 
[...]

PR #13486, opened 2019-03-24
Add avrr in aaa #13229 #Ranal

Maijin:


libr/core/cmd_anal.c
```diff
@@ -7906,6 +7906,10 @@ static int cmd_anal_all(RCore *core, const char *input) {
 				oldstr = r_print_rowlog (core->print, "Check for objc references");
 				r_print_rowlog_done (core->print, oldstr);
 				cmd_anal_objc (core, input + 1, true);
+				static const char *oldstr = NULL;
```
radare: why static, it will leak, also, why set to null and in the next line assgn again? just do it in one line

libr/core/cmd_anal.c
```diff
@@ -7906,6 +7906,10 @@ static int cmd_anal_all(RCore *core, const char *input) {
...
+				oldstr = r_print_rowlog (core->print, "Check for vtables");
+				r_core_cmd0 (core, "avrr");
+				r_print_rowlog_done (core->print, oldstr);
```
radare: free the oldstr after this line, and dont make it static or even const

Maijin: Actually the `static const char *oldstr = NULL;` is a copy/pasta mistake, i've removed it

libr/core/cmd_anal.c
Maijin: Actually the static const char *oldstr = NULL; is a copy/pasta mistake, i've removed it

PR #13492, opened 2019-03-25
[WIP] Remove _size hack from set_meta_min_if_needed()

kazarmy:


kazarmy:
Closing ... will try a different approach that is less ambitious.

PR #13493, opened 2019-03-25
Optimize r_str_bounds by removing an unnecessary strdup ##optimization

radare:
Improves graph generation times. But it's still called too many times


radare:
the width is now computed without considering the newline a printable char. so imho the new output (one less column) is better and we have to update the tests and merge

Screenshot 2019-03-25 at 05 53 26

PR #13495, opened 2019-03-25
Improve aav ##anal

radare:
* Do not take as valid hits the words found inside a function (considered code)
* Honor anal.arch.align
* Call it after aac in aaa

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -7492,17 +7492,22 @@ static bool archIsMips(RCore *core) {
 	return false;
 }
 #endif
+static bool archIsThumbable(RCore *core) {
+	RAsm *as = core ? core->assembler : NULL;
+	if (as && as->cur && as->bits <= 32 && as->cur->name) {
+		return strstr (as->cur->name, "arm");
+	}
+	return false;
+}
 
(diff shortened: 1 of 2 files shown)


radare:
reduces lots of false positives, tested on x86-64, mips and arm, probably breaking some tests.

Maijin:
Only 2 XX, need to be checked manually to see if they're relevant

libr/core/cmd_anal.c
```diff
@@ -7484,46 +7484,51 @@ static void cmd_anal_aad(RCore *core, const char *input) {
...
-		r_core_cmdf (core, "ax 0x%"PFMT64x " 0x%"PFMT64x, to, from);
-		r_core_cmdf (core, "Cd %d @ 0x%"PFMT64x, vsize, from);
-		r_core_cmdf (core, "f+ aav.0x%08"PFMT64x "= 0x%08"PFMT64x, to, to);
-#endif
 	}
 }
 
 static void cmd_anal_aav(RCore *core, const char *input) {
 #define seti(x,y) r_config_set_i(core->config, x, y);
 #define geti(x) r_config_get_i(core->config, x);
```
XVilka: Maybe remove these macroses? They are used only twice, no need for them imho.

PR #13497, opened 2019-03-25
Fix #13485 - Crash in anal_tp (null deref)

### LINKED ISSUE #13485: aaaa then aaa crash
OS: Ubuntu x86_64
PE 64
r2 -v
radare2 3.4.0-git 21274 @ linux-x86-64 git.3.3.0-237-g41c7ba5cd
commit: 41c7ba5cd29ea972a5d0a70fe17145396d67dfa0 build: 2019-03-24__10:59:20

### Expected behavior
It does not crash.

### Actual behavior
It segfaults.

Run
> aaaa
then
> aaa
Segfaults on aaft I think.

[Efiflash.exe.zip](https://github.com/radare/radare2/files/3000510/Efiflash.exe.zip)

radare:


radare:
Please confirm the fix

PR #13510, opened 2019-03-26
Fix build on capstone3

ret2libc:


radare:
Didnt we agreed to drop cs3 support?

ret2libc:
Oh really? That would be a pity, many distros still ship capstone3. I think we should at least be able to compile radare2 on capstone3, though the user experience could be degraded of course because of missing functions.

radare:
We also killed py2, mingw32 and cygwin support.

Debian stable is still shipping cs3, but.. im aware there are a bunch of missing instructions in there so i dont know how much we want our users to experience that. Maybe we want to have a travis line for cs3 test?

ret2libc:
We already have :) And it's failing because of those recent additions (see https://travis-ci.org/radare/radare2/jobs/511350828)

ret2libc:
BTW, also Fedora 29 is still shipping capstone (and Fedora is usually one of the most updated distros, but capstone4 was just recently released and it will be shipped in Fedora 30, which is supposed to be released in few months AFAIR)

radare:
sounds reasonable

PR #13513, opened 2019-03-26
added an early-out optimization for r_isprint. 

roboman2444:
Reduces time spent in r_isprint from ~30% to ~2% in my tests.

Measured with valgrind cachegrind.
Appears to have minor real-world performance improvements (2% less time spent with a long r2pipe script).


radare:
Good catch

PR #13518, opened 2019-03-26
Rewrite kernelcache check_bytes

mrmacete:
Parse load commands to extract 3 metrics (kudos to @Siguza for suggesting):
- use of LC_UNIXTHREAD
- absence of any library linked against
- addresses in the upper half of the address space

If those are met, then it’s a kernelcache. This just needs to parse all load commands types and vmaddr from LC_SEGMENT_64 (but nothing more) from the bytes.

Plus: no big endian kernelcaches so far.

### DIFF
--- a/libr/bin/format/mach0/mach064_is_kernelcache.c
+++ b/libr/bin/format/mach0/mach064_is_kernelcache.c
@@ -1,0 +1,42 @@
+static bool is_kernelcache(const ut8 *buf, ut64 length) {
+	r_return_val_if_fail (length >= sizeof (struct MACH0_(mach_header)), false);
+
+	const ut8 *end = buf + length;
+	const ut8 *cursor = buf + sizeof (struct MACH0_(mach_header));
+	int ncmds = r_read_le32 (buf + 16);
+	bool has_unixthread = false;
+	bool has_negative_vaddr = false;
+
+	for (int i = 0; i < ncmds; i++) {
+		if (cursor >= end) {
+			return false;
+		}
+
+		ut32 cmdtype = r_read_le32 (cursor);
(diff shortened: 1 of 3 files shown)


radare:
Guess that conflicts with @ret2libc pr

libr/bin/format/mach0/mach064_is_kernelcache.c
```diff
@@ -0,0 +1,42 @@
+static bool is_kernelcache(const ut8 *buf, ut64 length) {
+	r_return_val_if_fail (length >= sizeof (struct MACH0_(mach_header)), false);
```
radare: as long as the input is tainted we dont want to use an assert in here because that will crash the whole program. It is better to add an if in here

libr/bin/format/mach0/mach064_is_kernelcache.c
```diff
@@ -0,0 +1,42 @@
+static bool is_kernelcache(const ut8 *buf, ut64 length) {
```
radare: use RBuffer

libr/bin/p/bin_mach064.c
```diff
@@ -4,29 +4,16 @@
 #include "bin_mach0.c"
 
 #include "objc/mach064_classes.h"
+#include "../format/mach0/mach064_is_kernelcache.c"
 
 static bool check_bytes(const ut8 *buf, ut64 length) {
```
radare: check_bytes sohuld be deprecated

ret2libc:
Just as a not for the future: we are trying to move to RBuffer, thus we should do all things with buffer (e.g. check_buffer should contain all the logic, while check_bytes, just for compatibility, should create a rbuffer and call check_buffer).

PR #13520, opened 2019-03-27
Fix integer overflow that leads to a crash in the DEX parser ##bin

radare:
clusterfuzz-testcase-minimized-ia_fuzz-5651766995058688.dms

### DIFF
--- a/libr/bin/p/bin_dex.c
+++ b/libr/bin/p/bin_dex.c
@@ -1528,9 +1528,10 @@ static void parse_class(RBinFile *binfile, RBinDexObj *bin, RBinDexClass *c,
 	if (!c || !c->class_data_offset) {
 		if (dexdump) {
 			rbin->cb_printf (
-				"  Static fields     -\n  Instance fields   "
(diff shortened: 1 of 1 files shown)


libr/bin/p/bin_dex.c
```diff
@@ -1544,15 +1544,34 @@ static void parse_class(RBinFile *binfile, RBinDexObj *bin, RBinDexClass *c,
...
 		p = bufbuf + c->class_data_offset;
-		// runtime error: pointer index expression with base 0x000000004402 overflowed to 0xfffffffffffffd46
-		p_end = p + (r_buf_size (binfile->buf) - c->class_data_offset);
+		// XXX may overflow
+		if (r_buf_size (binfile->buf) < c->class_data_offset) {
+			return;
+		}
+		ut32 p_size = (r_buf_size (binfile->buf) - c->class_data_offset);
+		p_end = p + p_size;
 		//XXX check for NULL!!
```
devnexen: What about this TODO ?

radare: fixed , thanks for pointing out

PR #13522, opened 2019-03-27
Use entrypoint from LC_UNIXTHREAD

mrmacete:
fixes https://github.com/radare/radare2/issues/13519

### DIFF
--- a/libr/bin/p/bin_xnu_kernelcache.c
+++ b/libr/bin/p/bin_xnu_kernelcache.c
@@ -723,6 +723,11 @@ static RList *entries(RBinFile *bf) {
 	}
 
 	RKernelCacheObj *kobj = (RKernelCacheObj*) obj->bin_obj;
+	ut64 entry_vaddr = kobj->mach0->entry;
+	ut64 entry_paddr = entry_vaddr - kobj->pa2va_exec;
+	RBinAddr *ba = newEntry (entry_paddr, entry_vaddr, 0);
+	r_list_append (ret, ba);
+
 	process_constructors (kobj, kobj->mach0, ret, 0, true, R_K_CONSTRUCTOR_TO_ENTRY, NULL);
 
 	return ret;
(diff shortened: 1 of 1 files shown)


libr/bin/p/bin_xnu_kernelcache.c
```diff
@@ -723,6 +723,11 @@ static RList *entries(RBinFile *bf) {
 	}
 
 	RKernelCacheObj *kobj = (RKernelCacheObj*) obj->bin_obj;
+	ut64 entry_vaddr = kobj->mach0->entry;
+	ut64 entry_paddr = entry_vaddr - kobj->pa2va_exec;
```
radare: can we assume entry_vaddr will be always be bigger than pa2va_exec ? can we have  a check?

libr/bin/p/bin_xnu_kernelcache.c
```diff
@@ -723,6 +723,11 @@ static RList *entries(RBinFile *bf) {
 	}
 
 	RKernelCacheObj *kobj = (RKernelCacheObj*) obj->bin_obj;
+	ut64 entry_vaddr = kobj->mach0->entry;
```
radare: kobj can be null?

libr/bin/p/bin_xnu_kernelcache.c
```diff
@@ -723,6 +723,11 @@ static RList *entries(RBinFile *bf) {
...
+	ut64 entry_paddr = entry_vaddr - kobj->pa2va_exec;
+	RBinAddr *ba = newEntry (entry_paddr, entry_vaddr, 0);
+	r_list_append (ret, ba);
```
radare: check for null if newEntry fails

libr/bin/p/bin_xnu_kernelcache.c
mrmacete: well, there's already a check right before

PR #13524, opened 2019-03-27
Call to same relocs must have same name instead of appending number

sivaramaaa:
### * Before

```
$ r2 -A ~/dev/karta/pngrutil.o 

[0x0800132c]> pdfs
0x0800132c call reloc.png_crc_read_45
0x0800133b call reloc.png_get_uint_32_60
0x08001352 call reloc.png_crc_read_83
0x08001361 call reloc.png_get_uint_32_98
0x0800138d:
0x08001398 call 0x8001399
0x080013a8 call reloc.png_crc_finish_169
0x080013b5:
0x080013cd call reloc.png_crc_read_206
0x080013dc call reloc.png_get_uint_32_221
0x080013f3 call reloc.png_crc_read_244
0x08001402 call reloc.png_get_uint_32_3
0x08001427 call 0x8001428
0x08001437 call reloc.png_crc_finish_56
0x08001444:
0x0800145c call reloc.png_crc_read_93
```

### * After 

```
$ r2 -A ~/dev/karta/pngrutil.o 

[0x0800132c]> pdfs
0x0800132c call png_crc_read
0x0800133b call png_get_uint_32
0x08001352 call png_crc_read
0x08001361 call png_get_uint_32
0x0800138d:
0x08001398 call png_warning
0x080013a8 call png_crc_finish
0x080013b5:
0x080013cd call png_crc_read
0x080013dc call png_get_uint_32
0x080013f3 call png_crc_read
0x08001402 call png_get_uint_32
0x0800141a cjmp 0x08001444
0x08001427 call png_warning
0x08001437 call png_crc_finish
0x08001444:
0x0800145c call png_crc_read

```

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -4610,16 +4610,17 @@ static char *ds_sub_jumps(RDisasmState *ds, char *str) {
 		name = fcn->name;
 	} else if (f) {
 		flag = r_core_flag_get_by_spaces (f, addr);
-		if (flag) {
-			name = flag->name;
-		} else {
-			RBinReloc *rel = getreloc (ds->core, addr, ds->analop.size);
+		RBinReloc *rel = getreloc (ds->core, addr, ds->analop.size);
+		if (rel) {
 			if (rel && rel->import && rel->import->name) {
 				name = rel->import->name;
+			} else if (rel && rel->symbol && rel->symbol->name) {
+				name = rel->symbol->name;
 			}
+		} else if (flag) {
+			name = flag->name;
 		}
-	}
-	if (name) {
+	} if (name) {
 		char *nptr, *ptr;
 		ut64 numval;
 		ptr = str;
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -4593,7 +4593,6 @@ static char *_find_next_number(char *op) {
 static char *ds_sub_jumps(RDisasmState *ds, char *str) {
 	RAnal *anal = ds->core->anal;
 	RFlag *f = ds->core->flags;
-	RFlagItem *flag;
 	const char *name = NULL;
 
 	if (!ds->jmpsub || !anal) {
@@ -4609,7 +4608,6 @@ static char *ds_sub_jumps(RDisasmState *ds, char *str) {
 	if (fcn) {
 		name = fcn->name;
 	} else if (f) {
-		flag = r_core_flag_get_by_spaces (f, addr);
 		RBinReloc *rel = getreloc (ds->core, addr, ds->analop.size);
 		if (rel) {
 			if (rel && rel->import && rel->import->name) {
@@ -4617,10 +4615,14 @@ static char *ds_sub_jumps(RDisasmState *ds, char *str) {
 			} else if (rel && rel->symbol && rel->symbol->name) {
 				name = rel->symbol->name;
 			}
-		} else if (flag) {
-			name = flag->name;
+		} else {
+			RFlagItem *flag = r_core_flag_get_by_spaces (f, addr);
+			if (flag) {
+				name = flag->name;
+			}
 		}
-	} if (name) {
+	}
+	if (name) {
 		char *nptr, *ptr;
 		ut64 numval;
 		ptr = str;


radare:
Why relocs > flags? This seems like its gonna be slow

sivaramaaa:
No it's only when asm.jmpsub is enabled , and it's disabled by default

radare:
I know but itay is proposing to enable it by default

sivaramaaa:
> I know but itay is proposing to enable it by default

probably we can enable once Rbin is optimised !

radare:
My PR proposal was to make RFlags use RBin as backend so we dont have to create flags for each rbin thing, therefor 2x speedup and /2 memory usage. but ret2libci didnt liked it.. and i agree iits a bad desgn because flags have diifferent properties.. but iit was nice poc :P

ret2libc:
Do consider that getreloc just takes O(nlogn) as it uses rbtree. So it's not that slow.

ret2libc:
Also, we could try to see if by using just flags instead of RBInRelocs we can get the same result.

radare:
Oh right you fixed that

sivaramaaa:
So shall we enable this by default ??

sivaramaaa:
> Also, we could try to see if by using just flags instead of RBInRelocs we can get the same result.

@ret2libc  No flags are the real culprit here , for every reloc call there is a flag created with a number appended since we can't have flag with same name in different offset

ret2libc:
Mh... AFAIK we can have flags with the same "displayed name" but different "real name".

BTW I meant O(logn), not O(nlogn)

libr/core/disasm.c
```diff
@@ -4610,16 +4610,17 @@ static char *ds_sub_jumps(RDisasmState *ds, char *str) {
 		name = fcn->name;
 	} else if (f) {
 		flag = r_core_flag_get_by_spaces (f, addr);
-		if (flag) {
-			name = flag->name;
-		} else {
-			RBinReloc *rel = getreloc (ds->core, addr, ds->analop.size);
+		RBinReloc *rel = getreloc (ds->core, addr, ds->analop.size);
```
radare: why resolve the reloc if we have a flag already? this is querying twice. and imho flags should go before the bin info

sivaramaaa: @radare  the main purpose of this PR is to use Rbin info first then only try to resolve using flags because currently for every call to same reloc function there is a flag created but with a number appended and it looks bad in disasm 

flag


ir

But yeah i have fixed querying twice issue

ret2libc: Can't we change the flags creation so that they don't contain the number suffix? As previously said, it should be possible to configure flagsitems to have a "displayed name" and a "real name" (that is unique), right?

sivaramaaa: Yup that could be done xD ,  I just wanted quick fix :) 

I won't have time until weekend,  but yeah let me try that way np :)

radare:
the reason for having the number is to avoid having all the relocs to the same symbol in the same offset, because it will miss a lot of uses

sivaramaaa:
> the reason for having the number is to avoid having all the relocs to the same symbol in the same offset, because it will miss a lot of uses


@ret2libc ping

ret2libc:
@sivaramaaa could you provide the binary you are testing? I don't understand why we need jmpsub to get flags in a call instruction.

sivaramaaa:
[pngrutil.zip](https://github.com/radare/radare2/files/3025711/pngrutil.zip)

@ret2libc  here you go ^^ 

steps to reproduce : 

```
r2 ./pngrutil.o 

> aaa
> s 0x0800132c 
> pdf~reloc.png_crc_read
```

XVilka:
Lets postpone this for 3.5.0

radare:
This was requested by cutter ppl i thibk we should make it for 3.4

ITAYC0HEN:
yup :) both for Cutter, r2 and for myself in the Karta plugin, the algorithm fails to spot all these as the same function and I don't want to write down ugly workaround to substring or to use the reloc table to verify these are the same. Simply, all these are the same function so when the user reads the disassembly in r2, the framework needs to show the functions as the same, instead of appending some numbers at the end.

```
[0x0800132c]> f~reloc.png_crc_read

0x0800019c 4 reloc.png_crc_read_156
0x080001cb 4 reloc.png_crc_read_203
0x08000847 4 reloc.png_crc_read_71
0x08000bf6 4 reloc.png_crc_read_246
0x0800112e 4 reloc.png_crc_read_46
0x0800132d 4 reloc.png_crc_read_45
0x08001353 4 reloc.png_crc_read_83
0x080013ce 4 reloc.png_crc_read_206
...
...
...
...

[0x0800132c]> # THIS SHOWS ZERO X-REFS
[0x0800132c]> axt reloc.png_crc_read
[0x0800132c]>
[0x0800132c]>
[0x0800132c]>

[0x0800132c]> axt reloc.png_crc_read_156
sym.png_crc_finish 0x800019b [CALL] call reloc.png_crc_read_156

[0x0800132c]> axt reloc.png_crc_read_239
sym.png_handle_oFFs 0x8002cee [CALL] call reloc.png_crc_read_239

[0x0800132c]> axt reloc.png_crc_read_110
sym.png_handle_tEXt 0x800356d [CALL] call reloc.png_crc_read_110
```

ITAYC0HEN:
So... 3.4 somehow?

ret2libc:
Personally I haven't wrapped my head around this yet... Still not sure if it's what we want or if we should change flags. But feel free to ignore me and go ahead if somebody else understands more about it and feels the change is good.

radare:
The solution would be to use f++ which dowsnt exist, and shoudl check if the new flag added have the same address, otherwise it will create a new name with that numeric suffix.

Sounds like a good fix?

libr/core/disasm.c
```diff
@@ -4609,17 +4608,20 @@ static char *ds_sub_jumps(RDisasmState *ds, char *str) {
...
+			}
+		} else {
+			RFlagItem *flag = r_core_flag_get_by_spaces (f, addr);
+			if (flag) {
+				name = flag->name;
 			}
 		}
-	}
-	if (name) {
+	} if (name) {
```
radare: } if ?, use different lines

radare:
please rebase

radare:
Write some tests pls

PR #13526, opened 2019-03-27
Add support to multiline assembly patching

guy-gal:
For example, this line 'OFFSET : nop;nop;nop' (provided in the example patchfile at binr/rabin2/patch.txt) now actually works


radare:
Better use the api than continue depending on the quotes here

guy-gal:
It fixes a behavior expected from 'wp' though, doesn't it?

radare:
Yup it does

PR #13537, opened 2019-03-28
Adding symbols for 32bit version of mfc140u.dll

oddcoder:
scripted


Maijin:
Can you sort it ?

oddcoder:
sure

oddcoder:
It will also be great if someone checked this before it get merged to verify its correctness, I am not sure if I did it right or not.

oddcoder:
the problem is while the vast majority of the ordinals are resolved, there few that are not captured, that makes me skeptical about the correctness of this  patch
image

Maijin:
Did you compare with other tools on some of the resolution?

oddcoder:
I checked some of the ordinals and x64dbg didn't resolve them as well. still being a bit paranoid about introducing new bug I thought I should mention that

Maijin:
Worst case you are adding wrong ordinals but that can be fixed later and it's usually easy to spot when you analyze something.

At first glances if x64db has the same output i think it's very likely good to go.

PR #13538, opened 2019-03-28
Fix pf not showing all output

oxcabe:
This fixes https://github.com/radare/radare2/issues/12041

Please review

### DIFF
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -1361,7 +1361,8 @@ static void cmd_print_format(RCore *core, const char *_input, const ut8* block, 
 		while (*fmt && IS_WHITECHAR (*fmt)) {
 			fmt++;
 		}
-		int struct_sz = r_print_format_struct_size (fmt, core->print, mode, 0) + 10;
+		int times = atoi (fmt);
+		int struct_sz = r_print_format_struct_size (fmt, core->print, mode, 0) * times;
 		int size = R_MAX (core->blocksize, struct_sz);
 		ut8 *buf = calloc (1, size);
 		if (!buf) {
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -1362,7 +1362,7 @@ static void cmd_print_format(RCore *core, const char *_input, const ut8* block, 
 			fmt++;
 		}
 		int times = atoi (fmt);
-		int struct_sz = r_print_format_struct_size (fmt, core->print, mode, 0) * times;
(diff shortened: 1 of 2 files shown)


libr/core/cmd_print.c
```diff
@@ -1361,7 +1361,8 @@ static void cmd_print_format(RCore *core, const char *_input, const ut8* block,
 		while (*fmt && IS_WHITECHAR (*fmt)) {
 			fmt++;
 		}
-		int struct_sz = r_print_format_struct_size (fmt, core->print, mode, 0) + 10;
+		int times = atoi (fmt);
```
radare: i think this "times" should be should be computed inside the r_print_format_struct_size, and i think this change s not solvinig the problem, (let me explain below)

libr/core/cmd_print.c
```diff
@@ -1361,7 +1361,8 @@ static void cmd_print_format(RCore *core, const char *_input, const ut8* block,
...
+		int struct_sz = r_print_format_struct_size (fmt, core->print, mode, 0) * times;
 		int size = R_MAX (core->blocksize, struct_sz);
```
radare: i think we should always read struct_sz and ignore the blocksize

libr/core/cmd_print.c
radare: oh wait, i think you are right, but the struct size should be computed inside the r_print_format_struct_size function, because the bug only affects when using a repeater

libr/core/cmd_print.c
oxcabe: should we put a limit to struct_sz?

radare:
nope, limits are bad

libr/core/cmd_print.c
radare: if struct is smaller than blocksize we will be readinig unnecessary bytes, so just remove the size variable and we should be fine i think

radare:
See the breaking tests

Screenshot 2019-03-29 at 13 22 39

oxcabe:
> i think we should always read struct_sz and ignore the blocksize

Everything broke in tests because in some cases (such as z, which prints a null terminated string) struct size is not controlled as tightly as in pq or po. I guess blocksize was being used as a quick-fix in this case, since it helped outputting correctly most formats.

We can fix it for each specific case, but that's gonna take some time,

radare:
i see, so then this is just a partial fix and computing the size of a null terminated string cant be done without parsing the actual struct in memory this is what was the R_MAX() thing suposed to do?

Issue #13259, opened 2019-03-04
asm.var.summary=true when numvars > screenheight? :D


radare:
at least something funky for scr.responsive

radare:
We may want to have asm.var.compact too and group vars by type like this:

int32_t var0, var1, var2, var3; ...

XVilka:
Very good idea! Totally agree.

radare:
any of you (the ones assigned) may be able to do this

Vane11ope:
if any of you guys don't do this then I probably can do this before the ROP gadgets improvement, but don't know shit about the code around this issue so I need time to read code first though.

kazarmy:
I'm currently on a slow and flaky internet connection and will be travelling soon so I'll wait

PR #13539, opened 2019-03-28
asm.var.summary should be on if the screen is too small ##visual

Vane11ope:
about this #13259

but still not thoroughly done probably and so much room to improve so you'd better keep this issue open just yet.

Screenshot from 2019-03-29 02-32-38

Screenshot from 2019-03-29 02-33-15

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -1687,8 +1687,9 @@ static void ds_show_functions(RDisasmState *ds) {
 	r_anal_fcn_vars_cache_init (core->anal, &vars_cache, f);
 
 	int o_varsum = ds->show_varsum;
-	if (!ds->show_varsum) {
-		int numvars = vars_cache.bvars->length + vars_cache.rvars->length + vars_cache.svars->length;
+	if (ds->interactive && !o_varsum) {
+		int padding = 10;
+		int numvars = vars_cache.bvars->length + vars_cache.rvars->length + vars_cache.svars->length + padding;
 		if (numvars > ds->l) {
 			ds->show_varsum = 1;
 		} else {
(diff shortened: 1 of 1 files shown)


libr/core/disasm.c
```diff
@@ -1685,6 +1685,17 @@ static void ds_show_functions(RDisasmState *ds) {
 	ds->stackptr = core->anal->stackptr;
 	RAnalFcnVarsCache vars_cache;
 	r_anal_fcn_vars_cache_init (core->anal, &vars_cache, f);
+
+	int o_varsum = ds->show_varsum;
```
radare: i think this is bool

libr/core/disasm.c
```diff
@@ -1685,6 +1685,17 @@ static void ds_show_functions(RDisasmState *ds) {
...
+	if (!ds->show_varsum) {
+		int numvars = vars_cache.bvars->length + vars_cache.rvars->length + vars_cache.svars->length;
+		if (numvars > ds->l) {
```
radare: if ds->l s the amount of lines of the screen.. maybe we want to do osome -5 or -10 for padding

radare: also i think we should only do this when in interactive mode, for visual

Vane11ope: i agree with both, 30 mins

Vane11ope: better 10 i guess for Vpp

Vane11ope: done.

radare:
waiting for travis, but it should be green

Vane11ope:
cool, take your time...with travis :)

PR #13547, opened 2019-03-29
Wip to kill *_bytes methods from RBin ##bin

radare:


libr/bin/bin.c
```diff
@@ -377,6 +378,8 @@ R_API bool r_bin_open_io(RBin *bin, RBinOptions *opt) {
 			return false;
 		}
 		ut64 seekaddr = is_debugger? opt->baseaddr: opt->loadaddr;
+		r_buf_seek (buf, seekaddr, 1);
+		buf->base_priv = seekaddr;
```
ret2libc: why? i think fat bins are a "slice" of a buffer. Why do you set `base_priv`?

radare:
Because a slice currently copies the bytes into a new malloc. Do we want that? Should be io based

ret2libc:
Ah you are right I think. Let’s wait for my refactoring of rbuf, because there I will support all operations on each type of rbuf

radare:
Yep

ret2libc:
Btw, right now slices are read only. Do we think readwrite buffers will be useful? If so, how do we want to use them? Shall A write on a slice be visible on the parent buffer as well? Or it should have its own copy of the data?

radare:
Readinly is fine for rbin

radare:
rebased after @ret2libc pr merged

radare:
Can you review @ret2libc ?

libr/bin/bin.c
```diff
@@ -383,39 +386,38 @@ R_API bool r_bin_open_io(RBin *bin, RBinOptions *opt) {
 		// change the name to something like
 		// <xtr_name>:<bin_type_name>
 		r_list_foreach (bin->binxtrs, it, xtr) {
-			if (xtr->check_bytes (buf_bytes, opt->sz)) {
+			if (!xtr->check_buffer) {
+				eprintf ("Missing check_buffer callback for '%s'\n", xtr->name);
```
ret2libc: Use `R_LOG_ERROR`

libr/bin/bin.c
```diff
@@ -383,39 +386,38 @@ R_API bool r_bin_open_io(RBin *bin, RBinOptions *opt) {
 		// change the name to something like
 		// <xtr_name>:<bin_type_name>
 		r_list_foreach (bin->binxtrs, it, xtr) {
-			if (xtr->check_bytes (buf_bytes, opt->sz)) {
+			if (!xtr->check_buffer) {
```
ret2libc: Many bin plugins do not have yet the check_buffer method :(

libr/bin/bin.c
```diff
@@ -383,39 +386,38 @@ R_API bool r_bin_open_io(RBin *bin, RBinOptions *opt) {
...
-								(void)iob->fd_read_at (io, tfd, 0, buf_bytes, opt->sz);
-							}
+							r_buf_seek (buf, 0, 0);
+							//buf->base_priv = 0;
 						}
 						// DOUBLECLOSE UAF : iob->fd_close (io, tfd);
 						tfd = -1; // marking it closed
-					} else if (opt->sz != file_sz) {
-						(void)iob->read_at (io, 0LL, buf_bytes, opt->sz);
+					//} else if (opt->sz != file_sz) {
```
ret2libc: ??

libr/bin/bin.c
```diff
@@ -383,39 +386,38 @@ R_API bool r_bin_open_io(RBin *bin, RBinOptions *opt) {
...
 						if (tfd < 0) {
 							tfd = iob->fd_open (io, fname, R_PERM_R, 0);
 						}
 						opt->sz = iob->fd_size (io, tfd);
 						if (opt->sz != UT64_MAX) {
-							buf_bytes = calloc (1, opt->sz + 1);
-							if (buf_bytes) {
-								(void)iob->fd_read_at (io, tfd, 0, buf_bytes, opt->sz);
-							}
+							r_buf_seek (buf, 0, 0);
```
ret2libc: use the right `R_BUF_*` macros.

libr/bin/bin.c
```diff
@@ -374,6 +375,8 @@ R_API bool r_bin_open_io(RBin *bin, RBinOptions *opt) {
 			return false;
 		}
 		ut64 seekaddr = is_debugger? opt->baseaddr: opt->loadaddr;
+		r_buf_seek (buf, seekaddr, 1);
```
ret2libc: use the right `R_BUF_*` macros

libr/bin/p/bin_xtr_fatmach0.c
```diff
@@ -13,13 +13,19 @@ static RBinXtrData * oneshot(RBin *bin, const ut8 *buf, ut64 size, int idx);
...
 	ut8 buf[4];
-	if (sz >= 0x300 && !memcmp (h, "\xca\xfe\xba\xbe", 4)) {
+	const ut64 sz = r_buf_size (b);
+	r_buf_read_at (b, 0, buf, 4);
+	if (sz >= 0x300 && !memcmp (buf, "\xca\xfe\xba\xbe", 4)) {
 		// XXX assuming BE
-		ut64 off = (ut64)r_read_at_be32 (h, 4 * sizeof (ut32));
+		// ut64 off = (ut64)r_read_at_be32 (h, 4 * sizeof (ut32));
+		// (ut64)r_read_at_be32 (h, 4 * sizeof (ut32));
+		r_buf_seek (b, 4 * sizeof (ut32), 0);
```
ret2libc: use `R_BUF_*` macros.

libr/bin/p/bin_xtr_fatmach0.c
```diff
@@ -13,13 +13,19 @@ static RBinXtrData * oneshot(RBin *bin, const ut8 *buf, ut64 size, int idx);
...
+		ut64 off = (ut64) r_buf_read_be32 (b);
```
ret2libc: no need to do seek+r_buf_read, you can use `r_buf_read_be32_at`.

libr/bin/bfile.c
```diff
@@ -447,12 +447,17 @@ static bool xtr_metadata_match(RBinXtrData *xtr_data, const char *arch, int bits
...
 		return false;
 	}
-
-	char *iter_arch = xtr_data->metadata->arch;
+	const char *iter_arch = xtr_data->metadata->arch;
 	int iter_bits = xtr_data->metadata->bits;
 	return bits == iter_bits && !strcmp (iter_arch, arch) && !xtr_data->loaded;
 }
 
+R_IPI RBinFile *r_bin_file_new_from_buffer(RBin *bin, const char *file, RBuffer *buf, ut64 file_sz, int rawstr, ut64 baseaddr, ut64 loadaddr, int fd, const char *pluginname, ut64 offset) {
```
ret2libc: ok, i guess this wrapper is just for now, right?

libr/bin/bin.c
radare: how should i do that with the new rbuffer api?

libr/bin/bin.c
radare: this is just for binxtr plugins, and i created a wrapper method that takes the buffer and calls the buf+sz method internally for now, we have to kill all the bytes methods before 3.6

libr/bin/bin.c
radare: we shouldnt be reading anything here, just pass the buffer to the check function

libr/bin/p/bin_xtr_fatmach0.c
radare: done

libr/bin/bfile.c
radare: yes. we must kill this when the PR gets merged. but at least this way we can kill all the public _bytes apis in here

libr/bin/bin.c
ret2libc: Ah right, if it's just for xtr there are only few of them

libr/bin/bfile.c
```diff
@@ -696,6 +701,12 @@ R_API void r_bin_file_free(void /*RBinFile*/ *bf_) {
 	free (a);
 }
 
+R_IPI RBinFile *r_bin_file_xtr_load_buffer(RBin *bin, RBinXtrPlugin *xtr, const char *filename, RBuffer *buf, ut64 file_sz, ut64 baseaddr, ut64 loadaddr, int idx, int fd, int rawstr) {
+	ut64 sz;
+	const ut8 *bytes = r_buf_data (buf, &sz);
```
ret2libc: mmm, `r_buf_data` returns the *whole buffer*, so even if you do a `r_buf_seek` before it returns the entire buffer anyway (making the previous r_buf_seek useless)... Do you want to get only the bytes from the current seek to the end of the buffer?

radare:
Even on io backends?

radare:
@ret2libc seems like rabin2 -G got broken. https://api.travis-ci.org/v3/job/535504547/log.txt how can i slice that rbuf with the new api?

libr/bin/bin.c
radare: we should probably do this rlogification in another pr, i alos want to do an RCoreIO killification. the RIO api is good enouogh to not require wrappers in core anymore for example

libr/bin/bin.c
radare: done

ret2libc:
> Even on io backends?

Yes, sure, with *all* backends the behaviour is the same: `r_buf_data` returns the whole buffer, from address 0 to address *max*.

> @ret2libc seems like rabin2 -G got broken. https://api.travis-ci.org/v3/job/535504547/log.txt how can i slice that rbuf with the new api?

You should use `r_buf_new_slice(original_buffer, starting_offset_of_the_slice, size_of_the_slice)`.

libr/bin/bfile.c
```diff
@@ -435,42 +427,34 @@ R_API bool r_bin_file_object_new_from_xtr_data(RBin *bin, RBinFile *bf, ut64 bas
...
-		return NULL;
-	}
-	// size is set here because the reported size of the object depends on
-	// if loaded from xtr plugin or partially read
-	if (!o->size) {
-		o->size = file_sz;
+	RBinFile *bf = r_bin_file_new (bin, file, NULL, r_buf_size (buf),
+		file_sz, rawstr, fd, pluginname, NULL, false);
+	if (bf) {
+		bf->buf = r_buf_ref (buf);
```
ret2libc: I think `bf->buf` should be set by `r_bin_file_new`.

libr/bin/bin.c
```diff
@@ -349,73 +347,56 @@ R_API bool r_bin_open_io(RBin *bin, RBinOptions *opt) {
...
-			eprintf ("Cannot allocate %u bytes.\n", asz);
-			return false;
-		}
-		ut64 seekaddr = is_debugger? opt->baseaddr: opt->loadaddr;
-		if (!iob->fd_read_at (io, opt->fd, seekaddr, buf_bytes, asz)) {
-			opt->sz = 0LL;
+	ut64 seekaddr = is_debugger? opt->baseaddr: opt->loadaddr;
+	if (seekaddr > 0 && seekaddr != UT64_MAX) {
+		// slice buffer if necessary
+		RBuffer *nb = r_buf_new_slice (buf, seekaddr, opt->sz);
```
ret2libc: why do we need to slice? i think we need it only when working on fat binaries, where you want to consider a binary that starts in the middle of another file.

radare:
Thats why this code exists. To make fat bins work

PR #13567, opened 2019-03-30
Fix a crash ##panels

Vane11ope:
go to menu ROP and open it then get out of V! once then comes back to V! then tries to get out and crash and now it is fixed though much improvement left undone for those menus like ROP such as better restoring algorithms etc

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -1588,11 +1588,14 @@ static void addPanelFrame(RCore *core, const char *title, const char *cmd, const
 		if (cmd) {
 			p->cmd = r_str_newf (cmd);
 		} else {
-			p->cmd = "";
+			p->cmd = r_str_newf ("");
 		}
-	} else {
+	} else if (cmd) {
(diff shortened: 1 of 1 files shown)


libr/core/panels.c
```diff
@@ -1588,11 +1588,14 @@ static void addPanelFrame(RCore *core, const char *title, const char *cmd, const
 		if (cmd) {
 			p->cmd = r_str_newf (cmd);
 		} else {
-			p->cmd = "";
+			p->cmd = r_str_newf ("");
```
radare: just use strdup(), no need to use a wrapper format functoin that does va_arg.. for just "" :P

Vane11ope: yeah in fact, just r_str_new should be better 'cuz that's wrapping strdup internally plus take care of if (!str) case. in any way, r_str_newf is totally unnecessary here and i certainly replace them all thanks :)

Vane11ope:
done.

PR #13574, opened 2019-03-31
anal_riscv: Implement get_reg_profile

laanwj:
Add `get_reg_profile` call for RISC-V. In this profile, include the integer and floating point registers. Use the ABI register naming instead of the architectural one as this was chosen in the disassembler.

This is straightforward as the user-level ISA only specifies these registers. In RV-I there are no flags, and only a few read-only counter CSRs are accessible from user mode which (I think) don't need to
be in this structure. In RV-F/D there is one 32-bit CSR containing a few flags.

This addition is not quite enough to get analysis or debugging working, but can be a step towards that, and prevents warnings about being unable to import the register profile from isil during launch on RISC-V.

Also update bits to `32|64` instead of `16|32`—there is no 16-bit RISC-V.

### DIFF
--- a/libr/anal/p/anal_riscv.c
+++ b/libr/anal/p/anal_riscv.c
@@ -141,43 +141,190 @@ static int riscv_op(RAnal *anal, RAnalOp *op, ut64 addr, const ut8 *data, int le
 	return op->size;
 }
 
-#if 0
-static int set_reg_profile(RAnal *anal) {
-	char *p =
-		"=pc	pc\n"
-		"=sp	sp\n"
+static char *get_reg_profile(RAnal *anal) {
+	const char *p = NULL;
+	switch (anal->bits) {
+	case 32: p =
+		"=PC	pc\n"
+		"=SP	sp\n" // ABI: stack pointer
+		"=LR	ra\n" // ABI: return address
+		"=BP	s0\n" // ABI: frame pointer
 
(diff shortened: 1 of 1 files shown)


libr/anal/p/anal_riscv.c
```diff
@@ -141,43 +141,190 @@ static int riscv_op(RAnal *anal, RAnalOp *op, ut64 addr, const ut8 *data, int le
...
+		"gpr	ft9	.64	488	0\n" // =f29
+		"gpr	ft10	.64	496	0\n" // =f30
+		"gpr	ft11	.64	504	0\n" // =f31
+		"gpr	fcsr	.32	512	0\n"
+		"flg	nx	.1	4096	0\n"
+		"flg	uf	.1	4097	0\n"
+		"flg	of	.1	4098	0\n"
+		"flg	dz	.1	4099	0\n"
+		"flg	nv	.1	4100	0\n"
+		"flg	frm	.3	4101	0\n"
```
laanwj: I'm slightly unsure about this part, is it allowed to define a 3-bit register field like this?

XVilka:
@radare @ret2libc I think it is a good addition to the 3.4.0 release, since a bugfix.

radare:
Im not sure about this .3 but should be supported if its not. So ok to merge and looking forward to get esil or real hw support

PR #13583, opened 2019-03-31
ROP gadgets work properly when in-n-out ##panels

Vane11ope:
ROP panel would be gone when V!->q->V!, and now it works properly like the settings will be saved.

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -1686,7 +1686,7 @@ static int saveLayoutCb(void *user) {
 
 static int copyCb(void *user) {
 	RCore *core = (RCore *)user;
-	addCmdfPanel (core, "How many bytes? ", "y %s");
+	addCmdfPanel (core, "How many bytes? ", "\"y %s\"");
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


libr/core/panels.c
```diff
@@ -1762,25 +1764,25 @@ static int systemShellCb(void *user) {
...
 	RCore *core = (RCore *)user;
-	addCmdfPanel (core, "search string: ", "\"/ %s\"");
+	addCmdfPanel (core, "search string: ", "/ %s");
 	return 0;
 }
 
 static int ropCb(void *user) {
 	RCore *core = (RCore *)user;
-	addCmdfPanel (core, "rop grep: ", "\"/R %s\"");
+	addCmdfPanel (core, "rop grep: ", "/R %s");
```
radare: uhmf wasnt those quotes necessary? or addCmdfPanel does the quoting by itself? otherwise can be a prooblem if u ropsearch for rax;ret for example

Vane11ope: great review because i wasn't familiar with the command just yet, need to check it after some sleep :P

also in that case, i don't know why i need this, but i need to unescape the cmd for savelayout and loadsavedlayout, which are used to save the cmd and restore by json.

otherwise backslash bothers the command to be carried out  actually. 

still, r_str_unescape didnt work properly for that....

radare:
I see.. but you should be able to save and run the quoted commands already

Vane11ope:
yeah i also thought so. i think there must be a bug in anywhere around that if that way doesn't work. let me track it down after work :)

Vane11ope:
done.

PR #13589, opened 2019-03-31
New command: 'wai' write assembly inside the current instruction ##asm

radare:


ITAYC0HEN:
Is this valid for non x86 or in general, instruction sets without the instruction "nop"?

And does it support more than one NOP to fill? Probably I'm too tired but didn't see any loop that can print more than one NOP.


Didn't test yet though, afk still

radare:
yes, wao is generic for all archs, and yes, also supports multinop

PR #13601, opened 2019-04-01
Fix #13585 - Changes to tl 

### LINKED ISSUE #13585: Changes in the tl command
* tl = only shw offset + type name
* tll = should be the current listing we have in tl (very verbose)
* missing space in help msg
* tl? is implicit imho, and usage string is empty

Screenshot 2019-03-31 at 20 13 31

GustavoLCR:
#13585

### DIFF
--- a/libr/core/cmd_type.c
+++ b/libr/core/cmd_type.c
@@ -115,15 +115,15 @@ static const char *help_msg_tt[] = {
 };
 
 static const char *help_msg_tl[] = {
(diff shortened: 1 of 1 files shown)


XVilka:
Please also send a PR to r2book, thank you!

radare:
I was expecting some broken tests.. but seems like we have 0 tests for this. Can you add some too? Thanks!

PR #13602, opened 2019-04-01
ec* shows bgcolor too for fgbg color

kazarmy:


XVilka:
Should be a part of 3.4.0 I think.

radare:
your wishes are orders to me

kazarmy:
thanks!

PR #13610, opened 2019-04-02
Add ELF Identification fields to elf32 pfo

kazarmy:


kazarmy:
Based on ELF spec at http://refspecs.linuxbase.org/elf/elf.pdf.

radare:
i added support for .h files in this directory. maybe those are more readable. but yeah thats the way to go :) i have a pad where i described all the ideas i have related about header parsing and types, structs... if you are in the mood i can share it with you and we can discuss some plans for 3.5

kazarmy:
is it this one? https://hackmd.io/3IvnIkxfSM6WNyeYLbESrA

radare:
yup

PR #13615, opened 2019-04-03
because i found this function useful at work ##panels

Vane11ope:
yup.


Vane11ope:
actually same kinda function exists in visual so I can cut out the common part of them tho, i don't know.

radare:
try to make better commit messages pls O:)

PR #13621, opened 2019-04-03
Fixed crash when doing aac and bin->cur is NULL

xarkes:


xarkes:
Waoh sorry about those spaces.

radare:
This makes me think about 3.4.2

Issue #13607, opened 2019-04-01
What about adding some newlines


radare:
<img width="1677" alt="Screenshot 2019-04-01 at 23 54 19" src="https://user-images.githubusercontent.com/917142/55362179-8a832e00-54d9-11e9-8de7-674f6707a852.png">

i understand the lack of newlines in function definitions. but for structs and enums it will be doable

StephaneLeveugle:
Hi, I'm going to work on this :)

radare:
Solved already

PR #13632, opened 2019-04-05
Add newlines to tsc/tuc/tec - #13607

StephaneLeveugle:
#13607

### DIFF
--- a/libr/core/cmd_type.c
+++ b/libr/core/cmd_type.c
@@ -412,6 +412,43 @@ static int print_struct_union_list_json(Sdb *TDB, SdbForeachCallback filter) {
 	return 1;
 }
 
+static void print_struct_union_in_c_format(Sdb *TDB, SdbForeachCallback filter) {
+	char *name = NULL;
+	SdbKv *kv;
+	SdbListIter *iter;
+	SdbList *l = sdb_foreach_list_filter (TDB, filter, true);
+	ls_foreach (l, iter, kv) {
+		if (name && !strcmp (sdbkv_value (kv), name)) {
+			continue;
+		}
+		free (name);
+		int n;
+		name = strdup (sdbkv_key (kv));
+		r_cons_printf ("%s %s {\n", sdbkv_value (kv), name);
+		char *p, *var = r_str_newf ("%s.%s", sdbkv_value (kv), name);
+		for (n = 0; (p = sdb_array_get (TDB, var, n, NULL)); n++) {
+			char *var2 = r_str_newf ("%s.%s", var, p);
+			if (var2) {
+				char *val = sdb_array_get (TDB, var2, 0, NULL);
+				if (val) {
+					r_cons_printf ("\t%s", val);
+					if (p && p[0] != '\0') {
+						r_cons_printf ("%s%s", strstr (val, " *")? "": " ", p);
+					}
+				}
+				r_cons_println (";");
+				free (val);
+			}
+			free (var2);
+			free (p);
+		}
+		free (var);
+		r_cons_println ("};");
+	}
+	free (name);
+	ls_free (l);
+}
+
 static int printkey_cb(void *user, const char *k, const char *v) {
 	r_cons_println (k);
 	return 1;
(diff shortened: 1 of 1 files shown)


radare:
can you update the tests? the new output looks correct. btw just noticed tfc is not consistent with tsc. 

| tfc <name>   Show function calling conventions
| tsc         List all loaded structs in C output format

imho we should find a better name for tfc (calling conventions) . comments on this @sivaramaaa @oxcabe ?

radare:
we also want a tf/ts/.. subcommand to open cmd.editor with this C output and then parse it back to reimport the new signature.

and it will be good to keep having oneliner version for exporting/importing as r2 commands (via *), im testing all the t subcommands and im seeing many inconsistencies and unexpected behaviours

StephaneLeveugle:
Sure! I'll do it over the weekend

oxcabe:
> imho we should find a better name for tfc (calling conventions)

I agree with changing tfc since it doesn't have to do with C output formatting. Maybe tsC? idk

> we also want a tf/ts/.. subcommand to open cmd.editor with this C output and then parse it back to reimport the new signature.

So, something like "to -"? We can open an issue for this

radare:
> On 5 Apr 2019, at 14:57, Óscar Carrasco <notifications@github.com> wrote:
> 
> imho we should find a better name for tfc (calling conventions)
> 
> I agree with changing tfc since it doesn't have to do with C output formatting. Maybe tsC? idk
> 
> 

i sent a PR to use tcc here,

> we also want a tf/ts/.. subcommand to open cmd.editor with this C output and then parse it back to reimport the new signature.
> 
> So, something like "to -"? We can open an issue for this
> 
> 

https://github.com/radare/radare2/issues/1364 <https://github.com/radare/radare2/issues/1364>

to - works already, but we want to edit, not to add. so my idea is taht we will need to specify which function/struct/enum/union/.. to edit in the command

radare:
Screenshot 2019-04-07 at 20 41 41

can you fix this last test, and also, add `d' subcommands to make that compatible with `"td .."` so, oneliners, becauses right now the ts* prints pf commands

radare:
you can do this in a separate pr, but its basically keeping the old behaviour in tsd, and make tsc multiline, feel free to change the test to use tsd instead of tsc instead of changing the output, and this way you get a test for both commands

radare:
ping

StephaneLeveugle:
Hey, couldn't answer sooner sorry about that.
Let me recap to make sure I understood what you need.

I need to create tsd/tud/ted whose functionality is the old behavior of tsc/tuc/tec, am I right?
If so, should I create new tests for these that repeat the tests for tsc/tuc/tec but in oneliners form or is it unnecessary?

radare:
> On 8 Apr 2019, at 12:45, Stephane Leveugle <notifications@github.com> wrote:
> 
> Hey, couldn't answer sooner sorry about that.
> Let me recap to make sure I understood what you need.
> 
> I need to create tsd/tud/ted whose functionality is the old behavior of tsc/tuc/tec, am I right?
> 

yes

> If so, should I create new tests for these that repeat the tests for tsc/tuc/tec but in oneliners form or is it unnecessary?
> 
> 
no need to create new tests just change the one you didnt updated to use tsd and dont update the output (because that would make it green)

StephaneLeveugle:
Alright, I'll do it tomorrow morning then :)
(european time)

radare:
Thanks!

StephaneLeveugle:
Something came up I wasn't able to do it this morning, tomorrow should be fine 
null

radare:
👍

radare:
tomorrow is today, any update?

StephaneLeveugle:
tonight :)

StephaneLeveugle:
Hey, I'm having an issue while updating the test.
It uses "tc" so if I turn it into "tsd" it won't work because it will only print structs and the test also expects typedef union and enum.
Should I update the output then? I can also update the tc command to use tsd/tud/ted instead of tsc/tuc/tec or create a new command altogether

radare:
why turn tc into tsd? and not tcd?

radare:
the idea is that there must be command modifiers to

* print any or all types in one line
* print any or all types in multiline

update the tests u like/need, as long as we test both implementations (oneliner, multiline)

StephaneLeveugle:
I ended up creating tcd which is the oneliner version of tc.
I updated the tests but travis is still failing and I'm not sure why.

I couldn't find any test failing due to my code but I can't say for sure.
My guess is that since I hadn't merged the changes from upstream in radare2-regressions but I had done it for my radare2 repo they were out of sync at the time of the build.

I since integrated the changes from upstream in radare2-regressions, can we manually launch a travis build to check if it changed anything?

radare:
Its failing because r2 moves fast and you have to rebase both prs to get it right. Can you do this now? Otherwise i will blind merge because i think all the failing tests are because of this

Thanks!

StephaneLeveugle:
I did it after seeing the build had failed, so ~2 hours ago.
Could you launch a build?

radare:
If you rebase and push -f the build should be restarted

radare:
thanks!

PR #13635, opened 2019-04-05
Showing function signature doesnt depend on asm.var ##disasm

radare:


radare:
We probably need a new config var for this like asm.fcnsig

PR #13636, opened 2019-04-05
 Initial import of the zdiff algorithm ##diff

radare:


radare:
thats how it looks right now, pretty far from being complete, but the skeleton is there, and there are a bunch of refactorings pending

Screenshot 2019-04-05 at 13 17 55

PR #13637, opened 2019-04-05
/at/af/.. list options instead of searching everything

radare:


radare:
maybe we dont want to change that behaviour because listing all the insturctions and their types can be useful..

radare:
tired of waiting for feedback. i have though about this and i think the current behaviour is fine

Issue #13620, opened 2019-04-03
Collapse switch/case in pds


radare:
![Screenshot 2019-04-03 at 19 47 01](https://user-images.githubusercontent.com/917142/55500800-4d3dae00-5649-11e9-8761-7c37711aedb9.png)

radare:
the swiitch/case lines doesnt provide any useful information to pds and they are very noisy. we may probably remove or collapse them

radare:
This is fixed now

PR #13643, opened 2019-04-05
Hide switch/case refs from pds ##print (#13620)

oxcabe:


libr/core/cmd_print.c
```diff
@@ -2341,7 +2341,9 @@ r_cons_pop();
 		if (asm_flags) {
 			str = strstr (line, ";-- ");
 			if (str) {
-				r_cons_printf ("%s\n", str);
+				if (!(strstr (line, "case") || strstr (line, "switch"))) {
```
radare: check with !strncmp because switch and case must be the first word after ";-- ", so we dont miss any false positive then

radare:
maybe we can keep the switch, because thats not really noisy, and allows us to see if the function contains a switch table... and how many of them.

radare:
and there's only 1 switch table.. but its printed many times (swithout your patch).. :<<<<

Screenshot 2019-04-05 at 17 39 22

oxcabe:
We could print switch table once and ignore the rest of them

oxcabe:
Review before merge please

libr/core/cmd_print.c
```diff
@@ -2341,7 +2342,9 @@ r_cons_pop();
 		if (asm_flags) {
 			str = strstr (line, ";-- ");
 			if (str) {
-				r_cons_printf ("%s\n", str);
+				if (strncmp(str + 4, "case", 4)) {
```
radare: use r_str_startswith()

radare: remember to use a space before (

libr/core/cmd_print.c
```diff
@@ -2431,10 +2434,15 @@ r_cons_pop();
...
-					if (show_offset) {
-						r_cons_printf ("%s0x%08"PFMT64x" ", use_color? pal->offset: "", addr);
+					if (strcmp(comment, switchcmp)) {
+						if (show_offset) {
+							r_cons_printf ("%s0x%08"PFMT64x" ", use_color? pal->offset: "", addr);
+						}
+						r_cons_printf ("%s%s\n", use_color? pal->comment: "", comment);
+					}
+					if (!strncmp (comment, "switch table", 12)) {
+						switchcmp = strdup (comment);
```
radare: you are not reusing the malloc, so its leaking twice

libr/core/cmd_print.c
```diff
@@ -2431,10 +2434,15 @@ r_cons_pop();
...
 				if (comment) {
-					if (show_offset) {
-						r_cons_printf ("%s0x%08"PFMT64x" ", use_color? pal->offset: "", addr);
+					if (strcmp(comment, switchcmp)) {
+						if (show_offset) {
+							r_cons_printf ("%s0x%08"PFMT64x" ", use_color? pal->offset: "", addr);
+						}
+						r_cons_printf ("%s%s\n", use_color? pal->comment: "", comment);
+					}
+					if (!strncmp (comment, "switch table", 12)) {
```
radare: you can use r_str_startswith()

libr/core/cmd_print.c
```diff
@@ -2431,10 +2434,15 @@ r_cons_pop();
 			if (show_comments) {
 				char *comment = r_core_anal_get_comments (core, addr);
 				if (comment) {
-					if (show_offset) {
-						r_cons_printf ("%s0x%08"PFMT64x" ", use_color? pal->offset: "", addr);
+					if (strcmp(comment, switchcmp)) {
```
radare: switchcmp may be null when not set, so check for null in here to avoid a crash

libr/core/cmd_print.c
```diff
@@ -2235,6 +2235,7 @@ static void disasm_strings(RCore *core, const char *input, RAnalFunction *fcn) {
 	const char *linecolor = NULL;
 	char *ox, *qo, *string = NULL;
 	char *line, *s, *str, *string2 = NULL;
+	char *switchcmp = malloc (core->blocksize);
```
radare: thats never freed. so its leaking.

you should set this var to NULL by default

libr/core/cmd_print.c
```diff
@@ -2431,10 +2434,15 @@ r_cons_pop();
...
+						}
+						r_cons_printf ("%s%s\n", use_color? pal->comment: "", comment);
+					}
+					if (!strncmp (comment, "switch table", 12)) {
+						switchcmp = strdup (comment);
 					}
-					r_cons_printf ("%s%s\n", use_color? pal->comment: "", comment);
 					R_FREE (comment);
 				}
 			}
```
radare: free it before leaving

radare:
also will be good to have a test with mac's ls's entrypoint

libr/core/cmd_print.c
oxcabe: This is what I was looking for. Thank you! :p

libr/core/cmd_print.c
oxcabe: switchcmp should be freed at the end of the for loop (or function), since I want it to retain its value each time a new comment is processed, right?

radare:
yes, free it when you dont need it anymore

oxcabe:
> also will be good to have a test with mac's ls's entrypoint

will do after commiting again

oxcabe:
Thank you for the code review @radare :P

I'll add tests after checking Travis

oxcabe:
Test added: https://github.com/radare/radare2-regressions/pull/1774

radare:
this patch is hiding all the comments, even the ones entered by the user. and it should only hide the switch and case ones

oxcabe:
weird I'll look what's going on here

test pr will be updated afterwards

oxcabe:
I roll-backed to a previous commit:

radare2 3.5.0-git 21414 @ linux-x86-64 git.3.3.0-380-gc72b66c00
commit: c72b66c0013e70d47a31e6b1aa733edeedc8b9dd build: 2019-04-06__01:04:12

and the behaviour is the same. I think it's not cause by this PR

oxcabe:
Also checked everything coming out from r_core_anal_get_comments. My comment wasn't there.

radare:
it was not like this in 0ee92a61b23916262f11377ca7e71feea289e19b, you fixed that behaviour

PR #13649, opened 2019-04-07
Add ELF Identification fields to elf64 pfo

kazarmy:


kazarmy:
Based on ELF64 header info at https://docs.oracle.com/cd/E19683-01/816-1386/6m7qcoblh/index.html.

Maijin:
don't forget to update the hardcoded -nn values too https://github.com/radare/radare2/blob/master/libr/bin/format/elf/elf.c#L144 etc.

kazarmy:
ok will do in next pr

radare:
-nn must die

kazarmy:
> don't forget to update the hardcoded -nn values too https://github.com/radare/radare2/blob/master/libr/bin/format/elf/elf.c#L144 etc.

well I'm not sure what -nn does but running the code above isn't one of them

kazarmy:
oh i see -nn calls rabin2

PR #13650, opened 2019-04-07
Enable asm.jmpsub by default

ITAYC0HEN:
This PR wishes to enable `asm.jmpsub` by default. This is as a continue to the conversation on Telegram and on PR #13524.

Please notice the calls in these two outputs:

image


sivaramaaa:
@radare as we have discussed earlier , this can be enabled by default since this works using getreloc and according to @ret2libc it's O(logn) , so must be good enough !

radare:
Lgtm. This was off before because sometimes jump destinations in each bb got a random flag and everything was so confusing. I think we are pretty safe now to switch it on by default

radare:
looks red btw

sivaramaaa:
Imo the test changes looks good , @radare if u can confirm it we can fix the test !

radare:
i confirm, but its red, fix the tests and ill merge

radare:
This is introducing several regressions. Mainly when using analysis with the debugger (jmp rip) and things like that

If we can get rid of those issues or solve the reloc problem differently

PR #13652, opened 2019-04-07
Fix pd large

radare:


radare:
imho the way to fix this is to make @x: and @s: to open a temporary malloc file, map it and then close it after running the command.

PR #13656, opened 2019-04-08
RBuffer rewriting

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1828


libr/bin/p/bin_ninds.c
```diff
@@ -33,13 +33,15 @@ static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut
...
-	ut64 sz = bf? r_buf_size (bf->buf): 0;
 	if (!bf || !bf->o) {
 		return false;
 	}
+	ut64 sz;
+	ut8 *bytes = r_buf_buffer (bf->buf, &sz);
 	load_bytes (bf, &bf->o->bin_obj, bytes, sz, bf->o->loadaddr, bf->sdb);
-	return check_bytes (bytes, sz);
+	bool res = check_bytes (bytes, sz);
+	free (bytes);
```
radare: wtf? r_buf_bnuffer is now returning a new buffer with all the bytes? this is gonna be a huge pain for big files

radare:
pls squash and rebase. this merge mess is kind of anoying :<

libr/bin/p/bin_ninds.c
ret2libc: yes :/ it's the only way.
`r_buf_buffer` should not be used! Consider that `r_buf_buffer` could be used only for "bytes buffer" and the buffer was already entirely allocated in memory. The problem is that r_buf_buffer could not be used for "file buffer" or other type of buffers. These changes allow us to implement `r_buf_buffer` for any kind of buffer, but of course you have to be very careful. Again, if you use `r_buf_buffer` you are defeating the very purpose of RBuffer, because you are asking to have the entire buffer allocated in memory (this was true before and it's still true. The only difference is that now you have a new allocated buffer, so that the method can be used for other types of buffers as well)

ret2libc:
> pls squash and rebase. this merge mess is kind of anoying :<

What's the issue with this? I'm still keeping things separated because it's easier for me in case something goes wrong I can revert and see what introduced the problem. We can squash+merge when we merge the PR into master. Until that moment, I see no issues in having some "Merge commit"

radare:
So this change is the step before removing this function?

radare:
The idea for it was to be able to avoid copying memory when we had tbuffer under control. This is we know whats in there the size and we can be sure its not fd allocated etc. Thats why i added the assert lines to avoid bad usage of it

But duplicating memory is the only way to do that. Imho is not a good change

ret2libc:
Removing `r_buf_buffer` would mean rewriting a lot of parts to directly use the RBuffer, which is a really huge work... IMO if it was fine to having X bytes in memory before, it is ok to have 2X as well, considering that the temporary allocation is freed almost immediately and that X should be small in the first place for what you said. Also, using r_buf_buffer is wrong in any case. If you just want to use r_buf_buffer, don't use RBuffer in the first place and just work with raw bytes.

ret2libc:
Anyway, if you don't want the double memory I could do it...

libr/anal/reflines.c
```diff
@@ -476,7 +476,8 @@ R_API char* r_anal_reflines_str(void *_core, ut64 addr, int opts) {
 		}
 	}
 	add_spaces (b, 0, pos, wide);
-	str = r_buf_free_to_string (b);
+	str = r_buf_to_string (b);
```
radare: What about r_buf_drain()

libr/bin/bfile.c
```diff
@@ -408,7 +406,7 @@ R_API bool r_bin_file_object_new_from_xtr_data(RBin *bin, RBinFile *bf, ut64 bas
 	ut64 sz = data->size;
 
 	RBinPlugin *plugin = get_plugin_with_buffer (bin, data->buf);
-	bf->buf = r_buf_new_with_bufref (data->buf);
+	bf->buf = r_buf_ref (data->buf);
```
radare: If this method returns a new instance it should be named “new” if its just increasing the ref ignore this msg

libr/bin/bin.c
```diff
@@ -1243,7 +1243,7 @@ R_API RBuffer *r_bin_package(RBin *bin, const char *type, const char *file, RLis
 		RListIter *iter;
 		ut32 num;
 		ut8 *num8 = (ut8*)&num;
-		RBuffer *buf = r_buf_new_file (file, true);
+		RBuffer *buf = r_buf_new_file (file, O_RDWR | O_CREAT, 0644);
```
radare: Wat. Why a buffer should use posix flags to create a rw file? Imho a buffer should work only in memory.

libr/bin/bin_write.c
```diff
@@ -37,7 +37,9 @@ R_API bool r_bin_wr_output(RBin *bin, const char *filename) {
 	if (!filename || !binfile || !binfile->buf) {
 		return false;
 	}
-	return r_file_dump (filename, r_buf_buffer (binfile->buf), r_buf_size (binfile->buf), 0);
+	ut64 tmpsz;
+	const ut8 *tmp = r_buf_buffer (binfile->buf, &tmpsz);
```
radare: Bad method name

libr/bin/format/dex/dex.c
```diff
@@ -36,7 +36,7 @@ RBinDexObj *r_bin_dex_new_buf(RBuffer *buf) {
 		goto fail;
 	}
 
-	r_buf_seek (bin->b, 0, 0);
+	r_buf_seek(bin->b, 0, R_BUF_SET);
```
radare: Imho we should reuse the io enums and add missing space

libr/bin/p/bin_dex.c
```diff
@@ -2072,7 +2077,7 @@ static RList *dex_fields(RBinFile *bf) {
 	r_list_append (ret, r_bin_field_new (addr, addr, siz, nam, sdb_fmt ("0x%08"PFMT64x, (ut64)val), fmt)); \
 	addr += siz;
 
-	r_buf_seek (bf->buf, 0, 0);
+	r_buf_seek(bf->buf, 0, R_BUF_SET);
```
radare: Space

libr/util/buf.c
```diff
@@ -383,391 +179,229 @@ R_API RBuffer *r_buf_new_sparse(ut8 Oxff) {
...
-	char *s;
-	if (!b) {
-		return strdup ("");
-	}
-	s = malloc (b->length_priv + 1);
-	if (s) {
-		memmove (s, b->buf_priv, b->length_priv);
-		s[b->length_priv] = 0;
+	ut64 sz = r_buf_size (b);
+	char *s = malloc (sz + 1);
```
radare: Check for null

libr/bin/format/pe/pe.c
```diff
@@ -2467,6 +2467,9 @@ static void bin_pe_get_certificate(struct PE_ (r_bin_pe_obj_t) * bin) {
 	bin->cms = NULL;
 	size = bin->data_directory[PE_IMAGE_DIRECTORY_ENTRY_SECURITY].Size;
 	vaddr = bin->data_directory[PE_IMAGE_DIRECTORY_ENTRY_SECURITY].VirtualAddress;
+	if (size == 0) {
```
radare: Unrelated

libr/io/p/io_zip.c
```diff
@@ -530,17 +531,17 @@ static ut64 r_io_zip_lseek(RIO *io, RIODesc *fd, ut64 offset, int whence) {
...
 		seek_val = (r_buf_size (zfo->b) < (offset + r_buf_tell (zfo->b)))? r_buf_size (zfo->b): offset + r_buf_tell (zfo->b);
 		io->off = seek_val;
-		r_buf_seek (zfo->b, seek_val, 0);
+		r_buf_seek(zfo->b, seek_val, R_BUF_SET);
 		return seek_val;
 	case SEEK_END:
 		seek_val = r_buf_size (zfo->b);
 		io->off = seek_val;
-		r_buf_seek (zfo->b, seek_val, 0);
+		r_buf_seek(zfo->b, seek_val, R_BUF_SET);
```
radare: Space

libr/bin/bfile.c
ret2libc: it is just increasing the ref.

libr/bin/bin.c
ret2libc: Mmmm... It was never like this. You could have a buffer that works on a file. However, before you could just say "true/false" to mean "create new file/do not create new file". Now I have given the possibility to the user to specify the flags.

libr/bin/format/dex/dex.c
ret2libc: I don't think so. If a library/app needs r_util it's not said it needs r_io as well. You should be able to use RBuffer without depending on r_io.h

libr/bin/format/pe/pe.c
ret2libc: unfortunately it is related, because i've changed some "size" type to `size_t`.

ret2libc:
I'm thinking about splitting this PR.

I can create a separate PR that just use in a better way `r_buf_buffer` and removes `r_buf_get_at`. Otherwise I guess this PR is too big, and it's not even finished...

libr/bin/format/pe/pe.c
ret2libc: and later on `size - 8` is passed to `r_buf_read_at`, so it would become a very large size_t number. Actually, this check should be `size < 8`.

radare:
pls add defaults for all those defines in r_userconf.h, in case the check is not done the default should be false
```
cp: './plugins.cfg' and 'plugins.cfg' are the same file
+ make -s -j2
ar: creating libr_z.a
ar: creating libr_zip.a
ar: creating librz.a
ar: creating libsdb.a
In file included from /home/travis/build/radare/radare2/libr/include/r_types.h:6:0,
                 from /home/travis/build/radare/radare2/libr/include/r_util.h:6,
                 from mem.c:3:
mem.c: In function 'r_mem_mmap_resize':
/home/travis/build/radare/radare2/libr/include/r_userconf.h:13:21: error: token "@" is not valid in preprocessor expressions
 #define HAVE_MREMAP @HAVE_MREMAP@
                     ^
mem.c:346:5: note: in expansion of macro 'HAVE_MREMAP'
 #if HAVE_MREMAP
     ^~~~~~~~~~~
make[2]: *** [mem.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [all] Error 2
make: *** [all] Error 2
+ exit 1
```

radare:
9 xx, im worried about the md5/sha1 ones but all the other issues seems to be regressions
Screenshot 2019-04-25 at 17 15 57

ret2libc:
Wow! Who did this lgtm thing? :D

libr/anal/reflines.c
ret2libc: Yes we *could* add it, but we need to keep a separate `r_buf_to_string` anyway. Moreover, we would need a `r_buf_drain_string` and a `r_buf_drain`. I'm not sure for now. Can we keep it this way for now and maybe think about it as we work with it?

libr/bin/format/pe/pe.c
```diff
@@ -672,7 +672,7 @@ static struct r_bin_pe_export_t* parse_symbol_table(struct PE_(r_bin_pe_obj_t)*
 		}
 	}
 	symctr = 0;
-	if (r_buf_read_at (bin->b, sym_tbl_off, (ut8*) buf, bufsz)) {
+	if (r_buf_read_at (bin->b, sym_tbl_off, (ut8*) buf, bufsz) > 0) {
```
radare: the loop below assumes the whole buffer have been read, maybe worth checking with == or chopping the loop below in case

ret2libc: I do agree. I actually had done it before, but then I chose to revert the changes because I wanted to do the minimal set of changes.

libr/include/r_util/r_buf.h
```diff
@@ -8,84 +8,104 @@ extern "C" {
...
-	ut8 *buf_priv;
-	ut64 length_priv;
-	st64 cur_priv;
-	// FIXME: some direct accesses to base_priv still exist unfortunately
-	ut64 base_priv;
-	RMmap *mmap_priv;
-	bool empty_priv;
-	bool ro_priv; // read-only
-	int fd_priv;
+	const RBufferMethods *methods;
```
radare: so much c++

libr/util/buf_bytes.c
```diff
@@ -0,0 +1,139 @@
...
+		priv->buf = t;
+		memset (priv->buf + priv->length, b->Oxff_priv, newsize - priv->length);
+	}
+	priv->length = newsize;
+	return true;
+}
+
+static int buf_bytes_read(RBuffer *b, ut8 *buf, size_t len) {
+	struct buf_bytes_priv *priv = get_priv_bytes (b);
+	ut64 real_len = priv->length < priv->offset? 0: R_MIN (priv->length - priv->offset, len);
```
radare: maybe size_t instead of ut64

libr/util/buf_bytes.c
```diff
@@ -0,0 +1,139 @@
...
+	memmove (buf, priv->buf + priv->offset, real_len);
```
radare: or cast it here

libr/util/buf_sparse.c
```diff
@@ -0,0 +1,220 @@
...
+			} else {
+				int l = R_MIN (c->to - priv->offset, len);
+				memcpy (buf, c->data + priv->offset - c->from, l);
+			}
+		}
+	}
+	if (priv->offset > max) {
+		return -1;
+	}
+	r = R_MIN (max - priv->offset, len);
```
radare: define and assign r here

libr/include/r_util/r_buf.h
ret2libc: eheh true. the alternative was to have a switch case in every function, which seemed more messy.

radare:
agree, its clenaer

radare:
fix those last comments and ill merge

unixfreaxjp:
I practically can not compile radare2 anymore in my FreeBSD older kernels after these commits by Mr @ret2libc :
https://github.com/radare/radare2/commit/917fcda0ad78ff0f444e9807838b1e4b86349cd0#diff-2e7dcceb4a48bc318baa54b718ae7b33 
It was perfectly working before until the latest of 3.4.1 I built as per shown in below version on May 11th 2019, I was busy for a week didn't notice these errors from cron email sooner, should've sent it to my mobile too next time:
```asm
$ r2 -v
radare2 3.5.0-git 22040 @ freebsd-x86-32 git.3.4.1-406-g8cb2ad4
commit: 8cb2ad402ba42d3e3239922e475be9352e36b425 build: 2019-05-11__02:14:27
```
Now is troubleshooting, I rollback to the above version.

ret2libc:
@unixfreaxjp if you could add the compile error you get we can try to fix it, thanks.

unixfreaxjp:
@ret2libc Ricardo, added. Thanks.

unixfreaxjp:
@ret2libc Ricardo, added. Thanks.

ret2libc:
@unixfreaxjp thanks for the log. Since I don't have a FreeBSD VM handy (and apparently we still don't have CI for FreeBSD :( ), could you quickly test if the below patch works? Thanks!

```patch
From 738a6c93ab0d9fc4f1ab02bde0b181b72660f437 Mon Sep 17 00:00:00 2001
From: Riccardo Schirone <sirmy15@gmail.com>
Date: Mon, 20 May 2019 10:58:43 +0200
Subject: [PATCH] Do not double define typedef RBuffer

It should be enough to define struct r_buf_t, since the typedef to
RBuffer is already specified.
---
 libr/include/r_util/r_buf.h | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/libr/include/r_util/r_buf.h b/libr/include/r_util/r_buf.h
index 1b16c4952..dd9057b84 100644
--- a/libr/include/r_util/r_buf.h
+++ b/libr/include/r_util/r_buf.h
@@ -38,14 +38,14 @@ typedef struct r_buffer_methods_t {
 	RBufferNonEmptyList nonempty_list;
 } RBufferMethods;
 
-typedef struct r_buf_t {
+struct r_buf_t {
 	const RBufferMethods *methods;
 	void *priv;
 	ut8 *whole_buf;
 	bool readonly;
 	int Oxff_priv;
 	int refctr;
-} RBuffer;
+};
 
 // XXX: this should not be public
 typedef struct r_buf_cache_t {
-- 
2.20.1

```

unixfreaxjp:
@ret2libc 
>  could you quickly test if the below patch works? 
I am patching it now, will add the result in here.

unixfreaxjp:
The patch has been released:  https://github.com/radare/radare2/pull/14112 and merged  in master https://github.com/radare/radare2/commit/ad0ccaf27e85b0fa27d52a2ef3e164d1d120820a

The patch merged version https://github.com/radare/radare2/commit/ad0ccaf27e85b0fa27d52a2ef3e164d1d120820a was tested successfully too after git pull:
```asm
   : 
cd "/usr/lib/radare2/" ;\
        rm -f last ; ln -fs 3.6.0-git last
cd "/usr/share/radare2/" ;\
        rm -f last ; ln -fs 3.6.0-git last
mkdir -p "/usr/share/radare2/3.6.0-git/"
/bin/sh sys/ldconfig.sh
/bin/sh ./configure-plugins --rm-static //usr/lib/radare2/last/
configure-plugins: Loading ./plugins.cfg ..
Removed 0 shared plugins that are already static
$ r2 -v
radare2 3.6.0-git 21802 @ freebsd-x86-64 git.3.5.1-48-gad0ccaf27
commit: ad0ccaf27e85b0fa27d52a2ef3e164d1d120820a build: 2019-05-20__10:03:55
```
This bug is FIXED.
Thank you @ret2libc & @radare

PR #13662, opened 2019-04-08
Increase autocompletion to 4096

radare:


radare:
Solves https://github.com/radare/radare2/issues/13651 but the proper fix requires more work

PR #13665, opened 2019-04-08
arm improvements/refactoring

daddesio:
My goal was to add a new glibc signature (so "s main" would work), and I ended up fixing other issues as well.

[gcc-linaro-arm-linux-gnueabi-2012.01-20120125_linux](https://releases.linaro.org/archive/12.01/components/toolchain/binaries/gcc-linaro-arm-linux-gnueabi-2012.01-20120125_linux.tar.xz) (GCC 4.6.3, glibc 2.13) produces binaries with the following entry point that require the new signature:

```
[0x000082bc]> pd20
            ;-- entry0:
            ;-- section..text:
            ;-- pc:
            ;-- r15:
            0x000082bc      4ff0000b       mov.w fp, 0                 ; [13] -r-x section size 220 named .text
            0x000082c0      4ff0000e       mov.w lr, 0
            0x000082c4      5df8041b       ldr r1, [sp], 4
            0x000082c8      6a46           mov r2, sp
            0x000082ca      4df8042d       str r2, [sp, -0x4]!
            0x000082ce      4df8040d       str r0, [sp, -0x4]!
            0x000082d2      dff814c0       ldr.w ip, [0x000082e8]      ; [0x82e8:4]=0x8395
            0x000082d6      4df804cd       str ip, [sp, -0x4]!
            0x000082da      0448           ldr r0, [0x000082ec]        ; [0x82ec:4]=0x8334
            0x000082dc      044b           ldr r3, [0x000082f0]        ; [0x82f0:4]=0x8351
            0x000082de      fff7daef       blx sym.imp.__libc_start_main
            0x000082e2      fff7e6ef       blx sym.imp.abort
            0x000082e6      0000           movs r0, r0
            0x000082e8      9583           strh r5, [r2, 0x1c]
            0x000082ea      0000           movs r0, r0
            0x000082ec      3483           strh r4, [r6, 0x18]
            0x000082ee      0000           movs r0, r0
            0x000082f0      5183           strh r1, [r2, 0x1a]
            0x000082f2      0000           movs r0, r0
            0x000082f4      034b           ldr r3, [0x00008304]        ; [0x8304:4]=0x81a8
```

### DIFF
--- a/libr/anal/p/anal_arm_cs.c
+++ b/libr/anal/p/anal_arm_cs.c
@@ -1474,17 +1474,19 @@ static int analop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len
 	case ARM_INS_NOP:
 		r_strbuf_setf (&op->esil, ",");
 		break;
+	case ARM_INS_BL:
+	case ARM_INS_BLX:
+		r_strbuf_append (&op->esil, "pc,lr,=,");
+		/* fallthrough */
 	case ARM_INS_BX:
 	case ARM_INS_BXJ:
-		{
-		const char *op1 = ARG (0);
-		if (!strcmp (op1, "pc")) {
-			r_strbuf_setf (&op->esil, "%d,$$,+,pc,=", pcdelta);
+	case ARM_INS_B:
+		if (ISREG(0) && REGID(0) == ARM_REG_PC) {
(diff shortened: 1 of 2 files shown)


daddesio:
Actually, patch #3 ("Cleanup br instruction") is wrong. In the emulator, we shouldn't be masking out the Thumb bit from PC because that will prevent the emulator from staying in Thumb mode.

radare:
Can you add some tests for those kind of binaries? Thanks for the pr!!

PR #13676, opened 2019-04-09
improve r_buf_buffer and remove r_buf_get_at

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1786

### DIFF
--- a/libr/bin/p/bin_mbn.c
+++ b/libr/bin/p/bin_mbn.c
@@ -77,14 +77,14 @@ static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut
 
 static bool load(RBinFile *bf) {
 	if (bf && bf->buf) {
-		const ut8 *bytes = r_buf_buffer (bf->buf);
-		ut64 sz = r_buf_size (bf->buf);
+		ut64 sz;
+		const ut8 *bytes = r_buf_buffer (bf->buf, &sz);
 		return load_bytes (bf, &bf->o->bin_obj, bytes, sz, bf->o->loadaddr, bf->sdb);
 	}
 	return false;
 }
 
-static int destroy (RBinFile *bf) {
+static int destroy(RBinFile *bf) {
 	return true;
 }
 
--- a/libr/bin/bin_write.c
+++ b/libr/bin/bin_write.c
@@ -37,7 +37,9 @@ R_API bool r_bin_wr_output(RBin *bin, const char *filename) {
 	if (!filename || !binfile || !binfile->buf) {
 		return false;
 	}
-	return r_file_dump (filename, r_buf_buffer (binfile->buf), r_buf_size (binfile->buf), 0);
+	ut64 tmpsz;
+	const ut8 *tmp = r_buf_buffer (binfile->buf, &tmpsz);
+	return r_file_dump (filename, tmp, tmpsz, 0);
 }
 
 R_API bool r_bin_wr_entry(RBin *bin, ut64 addr) {
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -40,10 +40,10 @@
 #define READWORD(x, i) READ32 (x, i)
 #endif
 
-#define BREAD8(x, i) r_read_ble8(r_buf_get_at (x, (i), NULL)); (i) += 1
-#define BREAD16(x, i) r_read_ble16(r_buf_get_at (x, (i), NULL), bin->endian); (i) += 2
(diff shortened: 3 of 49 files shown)


libr/bin/p/bin_mbn.c
```diff
@@ -77,14 +77,14 @@ static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut
 
 static bool load(RBinFile *bf) {
 	if (bf && bf->buf) {
-		const ut8 *bytes = r_buf_buffer (bf->buf);
-		ut64 sz = r_buf_size (bf->buf);
+		ut64 sz;
+		const ut8 *bytes = r_buf_buffer (bf->buf, &sz);
```
radare: i think r_buf_buffer is a really bad name

radare:
do you know that a buffer can be used as a frontend for io, and this can be 64bit in size? this is a bunch of terabytes.. i doubt you want to do that

radare:
yeah i cant think of a better name either

ret2libc:
> do you know that a buffer can be used as a frontend for io, and this can be 64bit in size? this is a bunch of terabytes.. i doubt you want to do that

I know. That's why you shouldn't use `r_buf_buffer` in the first place :) It's inherently wrong.

radare:
Fhen how do you read data without copying it entirely? :p when u know the data is there

ret2libc:
How do you know "the data is there"? You may know it only if you created the RBuffer in that same function, but... at that point there's no need to use RBuffer if you just want the entire buffer.

In all other cases, you can't be sure that "the data is there". The whole point of RBuffer is that you need to read small chunks when you need them, because the data may not be there.

radare:
> On 9 Apr 2019, at 21:49, Riccardo Schirone <notifications@github.com> wrote:
> 
> How do you know "the data is there"? You may know it only if you created the RBuffer in that same function, but... at that point there's no need to use RBuffer if you just want the entire buffer.
> 
> 

because if its a malloc, the data is there. any other option that is not malloc or mmap it shouldnt allow to access that raw buffer pointer+fixedsize

radare:
yes, reading small chunks, but not copying it entirely, for example a fatmacho file can contain 10 binaries of 10M each, and u only want to load one, theres no need to copy all that memory several times to finally get few bytes to check the header. the copy of a bfufer must be an explicit operation. which can be done in a constructor for example.

radare:
and instead of r_buf_buffer() (which maybe should be named "r_buf_weakdata()" or so.. there must be an r_buf_clone() oor r_buf_dup() to duplicate the buffer contents

ret2libc:
> because if its a malloc, the data is there. any other option that is not malloc or mmap it shouldnt allow to access that raw buffer pointer+fixedsize

My point is, if you have this:
```C
void myfunction(RBuffer *b, int a) {
    const ut8 *t = r_buf_buffer (b);
    ...
}
```

How can myfunction know that it will always receive a malloc/mmap-based RBuffer? It can't. So, either you make r_buf_buffer work with all kinds of RBuffers, or you can't use it at all. That said, you shouldn't use r_buf_buffer in any case, but you should always assume the data needs to be loaded from somewhere. If you don't want to do that, then I don't know why we are using Rbuffer.

radare:
Thats the thing:

* we dont want to use this api
* the api only makes sense on malloc cases where the user is in control of what it have created
* copying data is an unnecessary step because if u read previsous points its clear that its not going to be necessRy

ret2libc:
> * we dont want to use this api

I agree

> * the api only makes sense on malloc cases where the user is in control of what it have created

I partially agree. The point is: RBuffer can be of different types. The power of using RBuffer is that you have an API that behaves the same indipendently of how it was created. So if you have a RBuffer backed by IO, you can do the same operations as if you have a RBuffer backed by malloc or one backed by file or mmap.

In the moment you add some operations that are valid only for some type of RBuffers, you are defeating the purpose of RBuffer, and it means when the RBuffer will switch his type, nothing will work anymore.

And almost never you can be sure of what's the type of RBuffer. For example, the `RBinPlugin.load` method does accept a RBuffer, but you can't say whether it comes from malloc, mmap, io or whatever else.

> * copying data is an unnecessary step because if u read previsous points its clear that its not going to be necessRy

Please provide example with code. Try to create a function that accepts a RBuffer and needs to compare 10 bytes at offset 0x200 with an hard-coded string. How would you do that?

libr/bin/p/bin_mbn.c
ret2libc: `r_buf_data`?

ret2libc:
@radare I think this is ready to be reviewed. Somehow travis is still red, but I think it's because of master... as soon as master gets green, i can make sure this is green as well.

radare:
ok to merge

PR #13678, opened 2019-04-09
Few UAF fixes ##esil

devnexen:


libr/anal/esil.c
```diff
@@ -2736,6 +2736,7 @@ static int esil_set_jump_target(RAnalEsil *esil) {
 		ret = true;
 	} else {
 		free (src);
+		src = NULL;
```
radare: use R_FREE instead of free + null

PR #13681, opened 2019-04-09
Use higher value for the default esil stack address ##esil

radare:
* We need a better solution to avoid overlaps
* But this is solving problems on big so/o files

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -2706,7 +2706,7 @@ R_API int r_core_config_init(RCore *core) {
 	SETICB ("esil.gotolimit", core->anal->esil_goto_limit, &cb_gotolimit, "Maximum number of gotos per ESIL expression");
 	SETICB ("esil.stack.depth", 32, &cb_esilstackdepth, "Number of elements that can be pushed on the esilstack");
 	SETI ("esil.stack.size", 0xf0000, "Set stack size in ESIL VM");
-	SETI ("esil.stack.addr", 0x100000, "Set stack address in ESIL VM");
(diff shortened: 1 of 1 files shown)


condret:
new anti-r2: change va in an elf file, so that r2 users cannot aeim

radare:
@ret2libc feedback on this?

ret2libc:
I'm not the best person for ESIL stuff, sorry. I don't know anything about it and i don't use it

radare:
well thats just a matter of adding an api in IO to give us a address to place the stack. because we need to not overlap a ything that is already mapped. so assuming ahigher address by default is fine, but this can be also wrong

radare:
Cc @condret related to r_io_map_location()

radare:
cc @condret this PR requires the r_io_map_location() api to be fixed properly

radare:
this is probably easier to fix now that tests are in the same repo, but we need an api in IO to ask for a region to be allocated at a specific range that is not overlapping any map

radare:
@thestr4ng3r c0mments or ideas for this issue?

thestr4ng3r:
I think the stack addr can easily depend on a lot of things in practice (arch, bits <- this especially, the kernel, aslr, ...) so by default we may want to have `esil.stack.addr` to be set to 0 or UT64_MAX to indicate that it should be created automatically and implement some magic that finds a good addr given all the arch, os, etc. properties. In theory this could also be a callback in an anal plugin to override the builtin magic with custom magic.

radare:
The only missing piece to get this done is to have the function that given a size and a minaddr returns a valid memory range that is not covered by any io map. Anyone?

radare:
ive fixed this with -e esil.stack.addr=-1

PR #13687, opened 2019-04-10
more RPICK removal in avr esil

condret:


radare:
KILL IT!

Screenshot 2019-04-10 at 11 35 33

PR #13691, opened 2019-04-10
pfj: Remove errorneous ]}, after struct nested to 1 level

kazarmy:


radare:
what about using the pj api instead

kazarmy:
turn a 1-line fix into fixes all over the place? ... emm i think that should really be in a different pr

radare:
yes a quickfix is fine for now, but we should pjinify

PR #13704, opened 2019-04-11
HL for move horizontally by a page

Vane11ope:


libr/core/panels.c
```diff
@@ -3714,6 +3715,18 @@ R_API int r_core_visual_panels(RCore *core, RPanels *panels) {
...
 		break;
+	case 'H':
+		r_cons_switchbuf (false);
+		for (i = 0; i < PANEL_CONFIG_PAGE; i++) {
+			cur->model->directionCb (core, (int)LEFT);
+		}
+		break;
+	case 'L':
+		r_cons_switchbuf (false);
+		for (i = 0; i < PANEL_CONFIG_PAGE; i++) {
```
radare: uh? isn this PAGE thing a constant size? the width != height, and also maybe it should scroll half page instead of full page (same with JK).

radare: this is a constant of 10. maybe worth considering frame->w /2; or frame->h /2 ?

radare:
im merging because tired of waiting and that will require more changes in JK too, lets just not block in here and go on with more issues fixed :) thanks!

PR #13706, opened 2019-04-11
Replace pointer with an array of variable size

ret2libc:
The elements are directly after the headers, there is no pointer
involved.

Also, use r_buf_read_at instead of r_buf_get_at in mdmp.


libr/bin/format/mdmp/mdmp.c
```diff
@@ -424,29 +416,24 @@ static bool r_bin_mdmp_init_directory_entry(struct r_bin_mdmp_obj *obj, struct m
...
-			endOffset = left;
-		}
-		modules = (struct minidump_module *)(&(module_list->modules));
-		for (i = 0; i < module_list->number_of_modules; i++) {
-			nextOffset += sizeOfModule;
-			if (nextOffset > endOffset) {
-				eprintf ("[INFO] Invalid number of modules or truncated file\n");
+		offset = entry->location.rva + sizeof (module_list);
+		for (i = 0; i < module_list.number_of_modules; i++) {
+			struct minidump_module *module = R_NEW (struct minidump_module);
```
radare: check for null

PR #13709, opened 2019-04-11
Refactoring plus a little tweak for HL and some fixes ##panels

Vane11ope:
Fixed some bug introduced by the last major refactoring.

Basically it stopped loading the saved settings properly for that and now it's fixed. Additionally a little tweak for HL commands and refactoring too. 

A few glitches found mainly caused by ansi_crop and it's known to me so no worries even if you encounter one on your try.


Vane11ope:
hang on

Vane11ope:
yeah now JK feels better too. I will make them configurable anytime soon anyway like 

e panels.scroll.h
e panels.scroll.v

or something.

radare:
No need to make that configurable. All tui/fui tools have a consistent scrolling behaviour :) just take the size of the frame

PR #13718, opened 2019-04-12
Enable capstone5

radare:


radare:
Please @ret2libc can you review the failing tests now, this is a no-return point, because cs5 introduces many breaking changes that will make the tests not be compatible with cs4 (see the issue https://github.com/aquynh/capstone/issues/1456)

Imho we are good to go but important current changes are:

* arm/thumb v8 tests are broken, not sure if we should have asm.cpu=v8 by default because that is breaking some tests in cs5 (cs4 was incorrect)
* xchg args are swapped
* aad now gets an arg

radare:
I just added a configure flag to pick capstone5 at compile time, i'm merging this, also i can say for now X86, ARM, THUMB and ARM64 are good. But PPC have still some problems and the next branch will not be merged into master (or v5) until the other archs got updated, this can take a long while, and we cant wait that much to have PAC support (which is the main reason to use capstone5).

PR #13721, opened 2019-04-12
Added missing c2 command + theme support for comparison commands ##cons

mkopec3:
I'm considering #defining _r_cons_singleton()->context->pal.x_ in _r_core.h_ , to not obscure the code further with such long statements, and do refactor for this piece of code and the previous changes.
If it does make sense for you, then I will propose the change via subsequent PR

### DIFF
--- a/libr/core/cmd_cmp.c
+++ b/libr/core/cmd_cmp.c
@@ -7,6 +7,7 @@ static const char *help_msg_c[] = {
 	"c", " [string]", "Compare a plain with escaped chars string",
 	"c*", " [string]", "Same as above, but printing r2 commands instead",
 	"c1", " [addr]", "Compare 8 bits from current offset",
+	"c2", " [value]", "Compare a word from a math expression",
 	"c4", " [value]", "Compare a doubleword from a math expression",
 	"c8", " [value]", "Compare a quadword from a math expression",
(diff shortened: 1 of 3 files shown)


radare:
The change looks good :) i will merge as soon as travis finishes. Refactorings are always welcome. Thanks!

radare:
Screenshot 2019-04-12 at 23 28 47

radare:
seems like with your patch the color ansi chars are always there even with scr.color=0

mkopec3:
Thanks for pointing this out!
However making this work for **r_print_hexdiff** seems to be not so trivial.

I could potentially use something like:
```
bool useColor = r_config_get_i (core->config, "scr.color") != 0;
const char *color = useColor? r_cons_singleton()->context->pal.offset: "";
const char *colorEnd = useColor? Color_RESET: "";
```
and add color/colorEnd to cb_printf calls... however I can't do it without passing RCore *core to the function, to make the config read work.
Is there any way to pass 'core' without changing the function prototype?
I don't want to screw dependencies further...

radare:
You cannot use RCons from RUtil, that breaks the dependency graph.
```
/home/travis/build/radare/radare2/install/lib/libr_util.so: undefined reference to `r_cons_singleton'
```

you should reuse the info from the RPrint struct

PR #13735, opened 2019-04-14
preventing to little segfaults

rlaemmert:


libr/anal/fcn.c
```diff
@@ -2188,7 +2188,7 @@ R_API int r_anal_fcn_cc(RAnal *anal, RAnalFunction *fcn) {
 	}
 
 	int result = E - N + (2 * P);
-	if (result < 1 && anal->verbose) {
+	if (result < 1 && anal && anal->verbose) {
 		eprintf ("Warning: CC = E(%d) - N(%d) + (2 * P(%d)) < 1 at 0x%08"PFMT64x"\n", E, N, P, fcn->addr);
```
radare: this is on purpose.. because if you get such wraning it means the function is wrongly analyzed and the analysis code must be fixed i would prefer to keep that warning

libr/anal/sign.c
```diff
@@ -1579,7 +1579,7 @@ static int hashMatchCB(RSignItem *it, void *user) {
 	char *digest_hex = NULL;
 	bool retval = false;
 	digest_hex = r_sign_calc_bbhash (ctx->anal, ctx->fcn);
```
radare: assign and define in the same line

PR #13750, opened 2019-04-15
Fix #10568 - Implement f. command to show flag name in current offset ##core

### LINKED ISSUE #10568: f.
In the same sense as the `afl.`, it would be helpful to have `f.` to return the flags only at the current position, and `f.j` to return the flags in the current position as a json structure.

radare:


radare:
This breaks the usage of f. which currently existss and lists the local fufnction flags. and it gets even worst if we try to use f.* or f.j 

we probably need to rethink this a bit more. i'll just do it for fq. for now

PR #13764, opened 2019-04-15
Implement 'less' command and add 'l?' command ##print

radare:


radare:
why it takes 2 min more than master? hopefully travis issue

PR #13769, opened 2019-04-15
Fixing possible dangling pointer for str empty macro

devnexen:


libr/core/cbin.c
```diff
@@ -1545,7 +1545,7 @@ static int bin_relocs(RCore *r, int mode, int va) {
 			if (bin_demangle) {
 				char *mn = r_bin_demangle (r->bin->cur, NULL, name, addr);
 				if (mn && *mn) {
-					free (name);
+					R_FREE (name);
```
radare: this is unnecesary because name is assigned right after being freed. so the call to `free` is enough

devnexen: Yes you re right for this one.

PR #13770, opened 2019-04-15
Code refactoring for theme based colorization in cmp & map

mkopec3:
As promised in PR #13721


radare:
This refactoring doesnt changes the final code. i think it will be better to catch the pointer to pal at the begining of the function and avoid all those multiple dereferences, also add a space before every (

mkopec3:
2nd try

PR #13772, opened 2019-04-15
n/N splitting a panel vertically/horizontally with new user input cmd ##panels

Vane11ope:


radare:
It doesnt makes sense to see appveyour red here. i think its just random issues in their side

Issue #12099, opened 2018-11-08
Add an option to always confirm on quit


radare:
Sometimes we are doing stuff in many r2 sessions and you end up cloaing the wrong one by accident. Being able to “protect” some sessions from quitting by setting a var will be useful to avoid such accidents.

This happens when no project or debugger session.

r00tten:
Hey, I would like to implement this feature, if it is still wanted.

radare:
Imho the way to go here is to add a new config variable `scr.confirmquit` or `scr.quitprompt` or something like this (im open to suggestions here). Being set to false by default.

When the user goes to quit r2 (prompt_repl) it must chk if scr.interactive is set and if this var is set, then ask for confirmation (like its done when quitting a debugging sesion).

Yes. it's still in the wishlist :) looking forward the PR.

assigned to @r00tten

PR #13778, opened 2019-04-16
Confirm on quit

r00tten:
Implementation of the issue "Add an option to always confirm on quit #12099".

Add new configuration variable and according to the value take confirmation when user try to quit.

### DIFF
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -1507,6 +1507,12 @@ R_API int r_main_radare2(int argc, char **argv) {
 					}
 					free (question);
 				}
+
+				if (r_config_get_i (r.config, "scr.confirmquit")) {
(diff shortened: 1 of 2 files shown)


radare:
Not sure if 'y' should be the default

libr/main/radare2.c
```diff
@@ -1507,6 +1507,12 @@ R_API int r_main_radare2(int argc, char **argv) {
 					}
 					free (question);
 				}
+
+				if (r_config_get_i (r.config, "scr.confirmquit")) {
+					if (!r_cons_yesno ('y', "Do you want to quit? (Y/n)")) {
```
radare: change the default to 'n' and i'll merge. thanks!

r00tten:
I've changed the default value as requested

PR #13779, opened 2019-04-16
Add scr.square and implement the prc= command ala pz,p=.. ##print

radare:


radare:
Screenshot 2019-04-16 at 10 56 50
Screenshot 2019-04-16 at 10 44 30

PR #13793, opened 2019-04-17
Fix coredump PC not being considered

GustavoLCR:
Before ESIL would overwrite the PC to be on the entry0


libr/core/cfile.c
```diff
@@ -677,8 +677,16 @@ R_API bool r_core_bin_load(RCore *r, const char *filenameuri, ut64 baddr) {
 						stack_map = r_io_map_get (r->io, sp_addr);
 					}
 				}
+				regname = r_reg_get_name (r->anal->reg, R_REG_NAME_PC);
+				if (regname) {
+					RRegItem *reg = r_reg_get (r->anal->reg, regname, -1);
+					if (reg) {
+						ut64 seek = r_reg_get_value (r->anal->reg, reg);
+						r_core_cmdf (r, "s 0x%x", seek);
```
radare: use r_core_seek()

PR #13797, opened 2019-04-17
cfg.contaftererr: Continue after error when using -i

kazarmy:
This is mainly for r2r.

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -2934,6 +2934,7 @@ R_API int r_core_config_init(RCore *core) {
 
 	/* cfg */
 	SETPREF ("cfg.r2wars", "false", "Enable some tweaks for the r2wars game");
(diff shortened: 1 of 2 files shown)


libr/core/cconfig.c
```diff
@@ -2934,6 +2934,7 @@ R_API int r_core_config_init(RCore *core) {
 
 	/* cfg */
 	SETPREF ("cfg.r2wars", "false", "Enable some tweaks for the r2wars game");
+	SETPREF ("cfg.contaftererr", "false", "Continue after error when using -i");
```
radare: uhm not sure if we want to have such dangerous option. also there are more ways to interpret scripts than -i, so the help is misleading

kazarmy:
> uhm not sure if we want to have such dangerous option

ok closing

PR #13801, opened 2019-04-17
Fix oba command when called with only one argument ##bin

radare:


radare:
should be green now :D

PR #13803, opened 2019-04-17
Improvements for rabin2 -H and pfo in macho ##bin

radare:


libr/util/buf.c
```diff
@@ -779,33 +779,24 @@ R_API bool r_buf_append_buf_slice(RBuffer *b, RBuffer *a, ut64 offset, ut64 size
...
 			break;
 		}
 		sz += r;
-		r = r_buf_read_at (b, addr + sz, tmp, sizeof (tmp));
+		r = r_buf_read_at (b, addr + sz, tmp + sz, sizeof (tmp) - sz);
 	}
 	if (r < 0) {
 		return NULL;
 	}
-
```
radare: cc @ret2libc

ret2libc: what?

Maijin:
https://github.com/radare/radare2/blob/6e29eb8831308a26d1c699b1fab4f6807dd931ed/libr/bin/format/mach0/mach0.c#L131-L153 Don't forget to also update the hardcoded as long as `r2 -nn` is not killed as this is what users currently use the most and not `pfo` nor `-H`

libr/util/buf.c
radare: this r_buf api was broken, dunno if u touched it but the code was snot working because r_buf_read_at 1 instead of the length of the buffer read. now the behaviour is correct, but wanted u to chk and will be good to add a unit

radare:
imho those hardcoded things should be moved into the pfo thing, less dupes and easier way to maintain, im doing pfs. but i could do also a .h.. but im just trying to find black spots in the `pf` syntax

radare:
@Maijin im aware of that dupplicateness, im trying to find the best appraoch to improve pf to avoid the use of the hacky -nn

libr/util/buf.c
```diff
@@ -779,33 +779,24 @@ R_API bool r_buf_append_buf_slice(RBuffer *b, RBuffer *a, ut64 offset, ut64 size
...
+	int r = strlen ((const char *)tmp);
 	while (r >= 0) {
 		needle = r_mem_mem (tmp, r, (ut8 *)"\x00", 1);
 		if (needle) {
 			sz += (needle - tmp);
 			break;
 		}
 		sz += r;
-		r = r_buf_read_at (b, addr + sz, tmp, sizeof (tmp));
+		r = r_buf_read_at (b, addr + sz, tmp + sz, sizeof (tmp) - sz);
```
ret2libc: this seems wrong... the purpose of tmp was to read a possibly long string chunk-by-chunk, until you find the terminator char. At that point, once you know the length of the string, you can read exactly `sz` bytes from the buffer.

I think the change you made could result in overflows of `tmp`.

libr/util/buf.c
ret2libc: mmm.. how did you fixed `r_buf_read_at` ? i see you only changed `r_buf_get_string`

Maijin:
I said about using pfo from the start but you said you wanted it to be hardcoded as well for performances. User only work with r2 -nn and nobody knows about pfo what do? Shouldn't this code be replaced by a simple load pfo file for the given fileformat?

ret2libc:
oh no... did you already merged? :(

libr/util/buf.c
```diff
@@ -779,33 +779,24 @@ R_API bool r_buf_append_buf_slice(RBuffer *b, RBuffer *a, ut64 offset, ut64 size
 // length depends on the first '\0' found in the buffer.
 R_API char *r_buf_get_string(RBuffer *b, ut64 addr) {
 	const ut8 *needle = NULL;
-	ut8 tmp[16];
+	ut8 tmp[128] = {0};
```
ret2libc: no need to 0 the entire string, just the last byte.

libr/util/buf.c
```diff
@@ -779,33 +779,24 @@ R_API bool r_buf_append_buf_slice(RBuffer *b, RBuffer *a, ut64 offset, ut64 size
...
 	}
 	if (r < 0) {
 		return NULL;
 	}
-
-	char *res = R_NEWS (char, sz + 1);
-	if (!res) {
-		return NULL;
-	}
-	r = r_buf_read_at (b, addr + 20, (ut8 *)res, sz);
```
ret2libc: this was the only thing you needed to change... for a wrong copy-paste i put here `addr + 20`, but you should have changed that to `addr`.

PR #13807, opened 2019-04-18
Fix r_buf_get_string implementation

ret2libc:
r2r PR: https://github.com/radare/radare2-regressions/pull/1800


libr/util/buf.c
```diff
@@ -776,27 +776,38 @@ R_API bool r_buf_append_buf_slice(RBuffer *b, RBuffer *a, ut64 offset, ut64 size
 }
 
 // return an heap-allocated string read from the RBuffer b at address addr. The
-// length depends on the first '\0' found in the buffer.
+// length depends on the first '\0' found in the buffer. If there is no '\0' in
+// the buffer, there is no string, thus NULL is returned.
 R_API char *r_buf_get_string(RBuffer *b, ut64 addr) {
-	const ut8 *needle = NULL;
-	ut8 tmp[128] = {0};
+#define MIN_RES_SZ 64
```
radare: use const instead of define

PR #13810, opened 2019-04-18
Fix use after free of parser hint.

jbroadus:
In some cases, r_core_print_disasm calls ds_build_op_str with a RDisasmState
instance. This sets core->parser->hint to ds->hint. When the ds is freed, it
also frees the hint, leaving the parser with a pointer to invalid memory.

A crash was reproducible in some functions by doing a /c command after a pdf.

This change stores the parser's original hint upon entry to ds_build_op_str and
restores it before returning.


jbroadus:
Looks like this was fixed earlier today.

thestr4ng3r:
Thanks, but this has just been fixed in #13806

PR #13811, opened 2019-04-18
Comment works in Disassembly panel ##panels

Vane11ope:
Still needs to brush this up a bit more but good for now.


radare:
please dont use ## on every single commit you refactor something in panels, otherwise the changelogs at the end will be still huge and not much useful

PR #13820, opened 2019-04-19
t key instead of tab key is supported in Disassembly panel ##panels

Vane11ope:
Initial support for tab key, but just not tab key, 't' instead, because people already got used to using tab key as a key to move around tabs, so.


radare:
/home/travis/build/radare/radare2/libr/core/libr_core.so: undefined reference to `applyDisMode'

Vane11ope:
lol fml hold on

Issue #13816, opened 2019-04-19
Allow setting flag->realname from commands


Siguza:
**Is your feature request related to a problem? Please describe.**  
I'm trying to resymbolicate large C++ code bases, and the only way I know of importing symbols to r2, short of writing a custom plugin, is via flags. Now flags filter our special characters in their "normal" name, but have a "realname" to preserve those. Only sadly, it seems you cannot set this realname from the command line or from scripts, and hence this:

    OSOrderedSet::initWithCapacity(unsignedint,int(*)(OSMetaClassBaseconst*,OSMetaClassBaseconst*,void*),void*)

turns into this:

    sym.OSOrderedSet::initWithCapacity_unsignedint_int____OSMetaClassBaseconst__OSMetaClassBaseconst__void___void

**Describe the solution you'd like**  
I'd like an additional flag command, with which the `realname` field of a flag can somehow be set from the command line and scripts.

radare:
I would say fN or fnn command for this. Can you send a PR?

Siguza:
Closing this as #13824 was merged.

PR #13824, opened 2019-04-19
Add realname flag command

Siguza:
As per issue #13816.

### DIFF
--- a/libr/core/cmd_flag.c
+++ b/libr/core/cmd_flag.c
@@ -38,7 +38,8 @@ static const char *help_msg_f[] = {
 	"fm"," addr","move flag at current offset to new address",
 	"fn","","list flags displaying the real name (demangled)",
 	"fnj","","list flags displaying the real name (demangled) in JSON format",
-	"fN"," [[name]] [realname]","set flag realname (if no flag name current seek one is used)",
+	"fN","","show real name of flag at current address",
+	"fN"," [[name]] [realname]","set flag real name (if no flag name current seek one is used)",
 	"fo","","show fortunes",
 	"fO", " [glob]", "flag as ordinals (sym.* func.* method.*)",
 	//" fc [name] [cmt]  ; set execution command for a specific flag"
@@ -935,7 +936,13 @@ rep:
 		}
 		break;
 	case 'N':
-		if (input[1] == ' ' && input[2]) {
+		if (!input[1]) {
+			RFlagItem *item = r_flag_get_i (core->flags, core->offset);
+			if (item) {
+				r_cons_printf ("%s\n", item->realname);
+			}
+			break;
+		} else if (input[1] == ' ' && input[2]) {
 			RFlagItem *item;
 			char *name = str + 1;
 			char *realname = strchr (name, ' ');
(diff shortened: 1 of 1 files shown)


libr/core/cmd_flag.c
```diff
@@ -934,6 +935,30 @@ static int cmd_flag(void *data, const char *input) {
 			}
 		}
 		break;
+	case 'N':
+		if (input[1]==' ' && input[2]) {
```
radare: spaces around =

libr/core/cmd_flag.c
```diff
@@ -934,6 +935,30 @@ static int cmd_flag(void *data, const char *input) {
...
+			char *name, *realname;
```
radare: assign and define in the same line

libr/core/cmd_flag.c
```diff
@@ -934,6 +935,30 @@ static int cmd_flag(void *data, const char *input) {
...
+			} else {
+				realname = name;
+				item = r_flag_get_i (core->flags, core->offset);
+			}
+			if (item) {
+				r_flag_item_set_realname (item, realname);
+			} else {
+				eprintf ("Cannot find flag (%s)\n", name);
+			}
+		}
```
radare: show help on else

Siguza:
Done. I had originally copy-pasted the `fr` command code, but I changed that as well now.

libr/core/cmd_flag.c
```diff
@@ -930,7 +930,30 @@ static int cmd_flag(void *data, const char *input) {
...
+			} else {
+				realname = name;
+				item = r_flag_get_i (core->flags, core->offset);
+			}
+			if (item) {
+				r_flag_item_set_realname (item, realname);
+			} else {
+				eprintf ("Usage: fN [[name]] [realname]\n");
 			}
 		}
```
radare: wait.. if no arguments, we should show the realname of the flag in the current offset imho (or all of them if more than one)

radare:
also can you add some tests after this?

Siguza:
I'd rather keep `fN`, for two reasons:

- If `fN x` changes the current symbol's realname, I'd find it more intuitive for no arguments to show _only the current symbol's realname_, instead of all of them.
- The code case for `fn` looks deeply entangled with `f`, `f*`, `fj` and `fq`:

      case '\0':
      case 'n': // "fn" "fnj"
      case '*': // "f*"
      case 'j': // "fj"
      case 'q': // "fq"
      	if (input[1] == '.' && !input[2]) {
      	// ...

So if it's OK with you, I'd leave `fn` the way it is, but change `fN` once more so that:

- `fN` -> show realname at current offset
- `fN realname` -> change realname at current offset
- `fN name realname` -> change realname of flag "name"

radare:
the help of fn talks about realname. maybe the help is wrong.. we can have your prposal merged, and then (wiuthout having tests) improve the commands to avoid uppercase ones and dupes

Siguza:
Alright, I made the change to `fN`.

And yes, the help of `fn` does talk about realname, but there is one single implementation that handles all of `f`, `fj`, `fq` and `f*`.

PR #13826, opened 2019-04-19
Include aap in aaaa ##anal

radare:


Maijin:
We should try a build on aaa too

radare:
I prefer to merge this PR as is, and experiment on aaa in a separate PR, can you do that change?

Maijin:
Yes feel free to merge that after it's green and I will recheck later for aaa integration

PR #13829, opened 2019-04-20
Mark pf z strings that overflow into unallocated memory with "ovf"

kazarmy:


libr/util/format.c
```diff
@@ -1182,6 +1182,22 @@ static void r_print_format_nulltermstring(const RPrint* p, const int len, int en
 			return;
 		}
 	}
+	if (p->flags & R_PRINT_FLAGS_UNALLOC && !(p->iob.io->cached & R_PERM_R)) {
+		ut64 total_map_left = 0;
+		ut64 addr = seeki;
+		RIOMap *map;
+		while ((map = p->iob.io->va
+		        ? p->iob.map_get (p->iob.io, addr)
```
radare: what about using         RIOIsValidOff is_valid_offset; ?

kazarmy: because i need map size ... f5cdc8745153779b34f1a8dc552c69066c337792 checks map read permission

radare:
Ok to merge then

PR #13837, opened 2019-04-20
Make p= bars dont use the background color ##cons

radare:
the reason to not use the bg is beacuse it conflicts with Cons_INVERT if we want to highlight a row and also because it will look bad in panels.

before:

Screenshot 2019-04-21 at 01 27 41

after: 
Screenshot 2019-04-21 at 01 25 36

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -1987,6 +1987,18 @@ static int cb_iotrap(void *user, void *data) {
 	return true;
 }
 
+static int cb_scr_bgfill(void *user, void *data) {
+	RCore *core = (RCore*) user;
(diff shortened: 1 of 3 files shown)


radare:
the reason to not use the bg is beacuse it conflicts with Cons_INVERT if we want to highlight a row and also because it will look bad in panels.

radare:
maybe we can add an option to choose which rendering mode use when theres an option to fill with bg or with ascii art (scr.fill?)

radare:
ok to merge like this or people want to keep the old bgfill format

Screenshot 2019-04-21 at 02 10 06

PR #13844, opened 2019-04-21
Fix #13672: Display variables value in the disasm view

### LINKED ISSUE #13672: Display variables value in the disasm view
also the new behaviour that enables asm.var.summary if the lines is short its a problem for this, we should be able to have an asm.summaryr=2 or -1 to make that behaviour consistent up to the user

Screenshot 2019-04-09 at 01 50 38

deepakchethan:
`asm.var.summary` must equal -1 for the values to be displayed

Screen Shot 2019-04-21 at 5 05 29 PM

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -1512,6 +1512,7 @@ static void ds_show_functions_argvar(RDisasmState *ds, RAnalVar *var, const char
 		if (val) {
 			r_str_replace_char (val, '\n', '\0');
 			r_cons_printf (" = %s", val);
+			free (val);
 		}
 	}
 	r_strbuf_free (constr_buf);
@@ -1757,6 +1758,7 @@ static void ds_show_functions(RDisasmState *ds) {
 					if (val) {
 						r_str_replace_char (val, '\n', '\0');
 						r_cons_printf ("%s", val);
+						free (val);
 					}
 				}
 				}
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -1507,6 +1507,13 @@ static void ds_show_functions_argvar(RDisasmState *ds, RAnalVar *var, const char
 			cond? constr: "",
 			cond? "} ":"",
 			base, sign, delta);
+	if (ds->show_varsum == -1) {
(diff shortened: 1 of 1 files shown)


radare:
Cool! Looks good. My only concern is that maybe we want to have an option to toggle that feature

deepakchethan:
> Cool! Looks good. My only concern is that maybe we want to have an option to toggle that feature
> […](#)
> On 21 Apr 2019, at 13:56, dodococo ***@***.***> wrote: @deepakchethan requested your review on: #13844 Fix #13672: Display variables value in the disasm view. — You are receiving this because your review was requested. Reply to this email directly, view it on GitHub, or mute the thread.

Yes Sir, if it only works if `asm.var.summary = -1` or else we will be getting the older output. Is that what you mean by toggling?

radare:
Maybe make it depend on asm.slow too?

libr/core/disasm.c
```diff
@@ -1507,6 +1507,13 @@ static void ds_show_functions_argvar(RDisasmState *ds, RAnalVar *var, const char
 			cond? constr: "",
 			cond? "} ":"",
 			base, sign, delta);
+	if (ds->show_varsum == -1) {
+		char *val = r_core_cmd_strf (ds->core, ".afvd %s", var->name);
+		if (val) {
+			r_str_replace_char (val, '\n', '\0');
+			r_cons_printf (" = %s", val);
```
radare: free val in here, otherwise it's leaking

libr/core/disasm.c
```diff
@@ -1745,9 +1752,16 @@ static void ds_show_functions(RDisasmState *ds) {
 					COLOR_ARG (ds, color_func_var_type),
 					var->type, r_str_endswith (var->type, "*") ? "" : " ",
 					var->name, COLOR_ARG (ds, color_func_var_addr), i->name);
+				if (ds->show_varsum == -1) {
+					char *val = r_core_cmd_strf (ds->core, ".afvd %s", var->name);
+					if (val) {
+						r_str_replace_char (val, '\n', '\0');
+						r_cons_printf ("%s", val);
```
radare: free(val)

radare:
looks green! fix the memleaks

PR #13846, opened 2019-04-21
Assume filetype is unknown if `core->bin->cur` is null

deepakchethan:
Fix for comment at https://github.com/radare/radare2/issues/13825#issuecomment-485068356.

Previously, the method `is_unknown_file` used to return `false` if `core->bin->cur` or `core->bin->cur->o` are null. By changing that assumption this works for r2 -n as well.


radare:
the logic seems correct to me

PR #13852, opened 2019-04-21
Fix asan

Vane11ope:


Vane11ope:
sys/asan and V! then entropy/entropy Fire, keep pressing t would crash r2 and it's gonna be fixed with this.

libr/core/cmd_print.c
```diff
@@ -3489,7 +3489,7 @@ static void cmd_print_bars(RCore *core, const char *input) {
 					break;
 				case 'z':
 					if ((IS_PRINTABLE (p[j]))) {
-						if (p[j + 1] == 0) {
+						if (!(&p[j + 1])) {
```
radare: wtf this cant be null how to repro this because the fix looks wrong to me

Vane11ope: you gave me this error and i wrote how to repro in description.

radare:
i pushed the right fix. this is wrong

PR #13859, opened 2019-04-21
n/N commands are supported in window mode ##panels

### LINKED ISSUE #13857: nN dont work in window mode for panels

Vane11ope:
Fix #13857


radare:
wasnt that working with return false in the key handler?

PR #13867, opened 2019-04-22
Use RPVector in RLineAutocompletion

thestr4ng3r:
First refactoring to bring decent autocomplete to Cutter

### DIFF
--- a/libr/cons/dietline.c
+++ b/libr/cons/dietline.c
@@ -537,7 +537,7 @@ static void selection_widget_select() {
 
 static void selection_widget_update() {
 	int argc = r_pvector_len (&I.completion.args);
-	const char **argv = I.completion.args.v.a;
+	const char **argv = r_pvector_data (&I.completion.args);
 	if (argc == 0 || (argc == 1 && I.buffer.length >= strlen (argv[0]))) {
 		selection_widget_erase ();
 		return;
@@ -568,7 +568,7 @@ R_API void r_line_autocomplete() {
 		I.completion.opt = false;
 		I.completion.run (&I);
 		argc = r_pvector_len (&I.completion.args);
-		argv = I.completion.args.v.a;
+		argv = r_pvector_data (&I.completion.args);
 		opt = I.completion.opt;
 	}
 	if (I.sel_widget && !I.sel_widget->complete_common) {
--- a/libr/cons/dietline.c
+++ b/libr/cons/dietline.c
@@ -536,8 +536,8 @@ static void selection_widget_select() {
 }
 
 static void selection_widget_update() {
-	int argc;
-	const char **argv = r_line_completion_argv (&I.completion, &argc);
+	int argc = r_pvector_len (&I.completion.args);
+	const char **argv = I.completion.args.v.a;
 	if (argc == 0 || (argc == 1 && I.buffer.length >= strlen (argv[0]))) {
 		selection_widget_erase ();
 		return;
@@ -567,7 +567,8 @@ R_API void r_line_autocomplete() {
 	if (I.completion.run) {
 		I.completion.opt = false;
 		I.completion.run (&I);
-		argv = r_line_completion_argv (&I.completion, &argc);
+		argc = r_pvector_len (&I.completion.args);
+		argv = I.completion.args.v.a;
 		opt = I.completion.opt;
 	}
 	if (I.sel_widget && !I.sel_widget->complete_common) {
--- a/libr/cons/dietline.c
+++ b/libr/cons/dietline.c
@@ -536,8 +536,10 @@ static void selection_widget_select() {
 }
 
 static void selection_widget_update() {
-	if (I.completion.argc == 0 ||
-		(I.completion.argc == 1 && I.buffer.length >= strlen (I.completion.argv[0]))) {
+	int argc = r_line_completion_argc (&I.completion);
+	const char **argv = r_line_completion_argv (&I.completion);
+	if (argc == 0 ||
+		(argc == 1 && I.buffer.length >= strlen (argv[0]))) {
 		selection_widget_erase ();
 		return;
 	}
@@ -547,9 +549,9 @@ static void selection_widget_update() {
 	}
 	I.sel_widget->scroll = 0;
 	I.sel_widget->selection = 0;
-	I.sel_widget->options_len = I.completion.argc;
-	I.sel_widget->options = I.completion.argv;
-	I.sel_widget->h = R_MAX (I.sel_widget->h, I.completion.argc);
+	I.sel_widget->options_len = argc;
+	I.sel_widget->options = argv;
+	I.sel_widget->h = R_MAX (I.sel_widget->h, I.sel_widget->options_len);
 	selection_widget_draw ();
 	r_cons_flush ();
 	return;
@@ -566,8 +568,8 @@ R_API void r_line_autocomplete() {
 	if (I.completion.run) {
 		I.completion.opt = false;
 		I.completion.run (&I);
-		argc = I.completion.argc;
-		argv = I.completion.argv;
+		argc = r_line_completion_argc (&I.completion);
+		argv = r_line_completion_argv (&I.completion);
 		opt = I.completion.opt;
 	}
 	if (I.sel_widget && !I.sel_widget->complete_common) {
--- a/libr/cons/dietline.c
+++ b/libr/cons/dietline.c
@@ -536,10 +536,9 @@ static void selection_widget_select() {
 }
 
 static void selection_widget_update() {
-	int argc = r_line_completion_argc (&I.completion);
-	const char **argv = r_line_completion_argv (&I.completion);
-	if (argc == 0 ||
-		(argc == 1 && I.buffer.length >= strlen (argv[0]))) {
+	int argc;
+	const char **argv = r_line_completion_argv (&I.completion, &argc);
+	if (argc == 0 || (argc == 1 && I.buffer.length >= strlen (argv[0]))) {
 		selection_widget_erase ();
 		return;
 	}
@@ -568,8 +567,7 @@ R_API void r_line_autocomplete() {
 	if (I.completion.run) {
 		I.completion.opt = false;
 		I.completion.run (&I);
-		argc = r_line_completion_argc (&I.completion);
-		argv = r_line_completion_argv (&I.completion);
+		argv = r_line_completion_argv (&I.completion, &argc);
 		opt = I.completion.opt;
 	}
 	if (I.sel_widget && !I.sel_widget->complete_common) {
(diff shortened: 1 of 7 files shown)


libr/include/r_cons.h
```diff
@@ -878,8 +879,13 @@ typedef int (*RLineCallback)(RLine *line);
 
 typedef struct r_line_comp_t {
 	bool opt;
+
+	// private! use r_line_completion_argc(), r_line_completion_argv()
 	int argc;
-	const char **argv;
+	const char **argv; // weak reference to a whole list, if NULL, args will be used
```
radare: uhm isnt args replacing argc+argv ?

libr/include/r_cons.h
```diff
@@ -878,8 +879,13 @@ typedef int (*RLineCallback)(RLine *line);
...
+	RPVector args; /* <char *> */
+	bool args_weak; // whether args are weak refs or should be freed on clear
```
radare: isnt this something that should be tied to the rpvector constructor? instead of adding another var

libr/include/r_cons.h
```diff
@@ -944,6 +950,21 @@ R_API int r_line_set_hist_callback(RLine *line, RLineHistoryUpCb cb_up, RLineHis
 R_API int cmd_history_up(RLine *line);
 R_API int cmd_history_down(RLine *line);
 
+R_API void r_line_completion_init(RLineCompletion *completion);
+R_API void r_line_completion_fini(RLineCompletion *completion);
+R_API void r_line_completion_push_owned(RLineCompletion *completion, char *str);
+R_API void r_line_completion_push_weak(RLineCompletion *completion, const char *str);
```
radare: i would simplify and just have a push(), if you mix owned and weaks in the same pvector it will end up leaking

libr/include/r_cons.h
thestr4ng3r: args replaces argc+argv iff `argv == NULL`. This argv is for directly setting a whole list, like in `r_core_fgets()` without having to copy anything.

libr/include/r_cons.h
thestr4ng3r: No, for the RPVector, weak would just mean "no free function", but I need to check this value at some places and don't really want to check if the free function is null since that would be kind of confusing.

libr/cons/dietline.c
```diff
@@ -536,8 +536,10 @@ static void selection_widget_select() {
 }
 
 static void selection_widget_update() {
-	if (I.completion.argc == 0 ||
-		(I.completion.argc == 1 && I.buffer.length >= strlen (I.completion.argv[0]))) {
+	int argc = r_line_completion_argc (&I.completion);
+	const char **argv = r_line_completion_argv (&I.completion);
```
radare: rename to `argv = r_line_completion_argv (&I.completion, &argc);`

libr/cons/dietline.c
```diff
@@ -536,8 +536,10 @@ static void selection_widget_select() {
...
+	if (argc == 0 ||
```
radare: no newline

libr/cons/line.c
```diff
@@ -41,4 +43,56 @@ R_API char *r_line_get_prompt() {
...
+			weak_str = r_pvector_at (&completion->args, i);
+			r_pvector_set (&completion->args, i, strdup (weak_str));
+		}
+	}
+	r_pvector_push (&completion->args, str);
+	completion->args_weak = false;
+}
+
+R_API void r_line_completion_push_weak(RLineCompletion *completion, const char *str) {
+	if (!str) {
```
radare: use r_return_ in here

libr/cons/line.c
```diff
@@ -41,4 +43,56 @@ R_API char *r_line_get_prompt() {
...
+		return;
+	}
+	if (!completion->args_weak) {
+		str = strdup (str);
+	}
+	r_pvector_push (&completion->args, (void *)str);
+}
+
+R_API void r_line_completion_set_weak(RLineCompletion *completion, int argc, const char **argv) {
+	r_line_completion_clear (completion);
```
radare: r_return

libr/include/r_cons.h
thestr4ng3r: They don't leak as long as the correct functions are used. The way it works is this: An empty RLineCompletion always has its `args_weak == true`. As long as you push weak refs into it, this will stay. As soon as you push an owned string, all currently saved strings will be copied and `args_weak` set to true. When you then push a weak ref, it will be copied too. On clearing, depending on `args_weak`, either all strings are freed or left alone.

libr/include/r_cons.h
radare: yes i know, but if you construct the vecvtor with NULL free callback it will behave like this, you dont need an extra var to duplicate the info already exposed in vec->free

libr/cons/line.c
```diff
@@ -41,4 +43,56 @@ R_API char *r_line_get_prompt() {
...
+	completion->args_weak = true;
+	r_pvector_init (&completion->args, NULL);
+}
+
+R_API void r_line_completion_fini(RLineCompletion *completion) {
+	r_line_completion_clear (completion);
+}
+
+R_API void r_line_completion_push_owned(RLineCompletion *completion, char *str) {
+	if (!str) {
```
radare: r_return

radare: also chk for completion

libr/cons/line.c
```diff
@@ -41,4 +43,56 @@ R_API char *r_line_get_prompt() {
...
+		return;
+	}
+	if (completion->args_weak) {
+		// weak to owned => must strdup all currently saved strings
+		const char *weak_str;
+		size_t i;
+		for (i = 0; i < r_pvector_len (&completion->args); i++) {
+			weak_str = r_pvector_at (&completion->args, i);
```
radare: define and assign in the same line

libr/cons/line.c
thestr4ng3r: to check for what?

thestr4ng3r: nvm

libr/include/r_cons.h
thestr4ng3r: Yep that's what I said, but then I need to check the free function, or more precisely `vec->v.free_user`, which is kind of an internal implementation detail of RPVector. It would definitely work, but I wouldn't really like it.

thestr4ng3r:
So in general, this thing could be simplified with certain performance regressions:
* argc+argv could be removed and instead all entries copied into the vector.
* the weak pushing could be removed and all entries could be copied (could have quite a big negative impact with many entries)

I personally would keep it as it is now, but I'm also open for discussion, especially about the first point.

libr/include/r_cons.h
ret2libc: Mmm why do we need owned/weak? Can't we either have all strings weak or all strings owned? It seems a bit messy...

thestr4ng3r: See the last comment. Only weak can't work and only owned would need a lot of copying.

radare:
uhm but those new apis are not used anywhere yet.i guess you want this for cutter, and curreent code behaves the same because it uses argc/argv in r2, but iin cutter will do the rvector thing, right?

I understand the performance issues but i dont see why we want to have a weak vector to become owned when we just push one owned value, instead of failing at compile or run time.

about the ->is_owned, what about moving this logic into `r_vector_is_weak(x) (x->free != NULL)`

thestr4ng3r:
Actually the whole thing IS used in r2, in the regular autocompletion code in core.c. Cutter will eventually just let core fill the RLineCompletion and read out the values.

I wouldn't expect many copies of previously weak strings since in practice, usually only either weak or owned ones will be pushed. Before, this was handled by the global tmp_argv_heap and that was only changed in very few places. So if we assume the old code didn't double free or leak, this will be true. The difference though now is that RLineCompletion automatically manages copying if needed.

About r_vector_is_weak, yeah I think that would be a good idea.

ret2libc:
I think always copying the string is not that important from a perf point of view. After all I don’t expect to have millions of auto completion (even if there were, you can show only a small set of those on the screen so it’s quite useless to have all of them).

I think we shouldn’t distinguish between owned and weak , since this is really not a performance critical path. KISS

thestr4ng3r:
Well you are probably right. I changed it to all owned with a limit of 4096.

libr/cons/dietline.c
```diff
@@ -536,8 +536,9 @@ static void selection_widget_select() {
 }
 
 static void selection_widget_update() {
-	if (I.completion.argc == 0 ||
-		(I.completion.argc == 1 && I.buffer.length >= strlen (I.completion.argv[0]))) {
+	int argc = r_pvector_len (&I.completion.args);
+	const char **argv = I.completion.args.v.a;
```
ret2libc: Isn’t there a method to get this? I don’t like accessing internal fields in this way

libr/cons/dietline.c
```diff
@@ -566,8 +567,8 @@ R_API void r_line_autocomplete() {
 	if (I.completion.run) {
 		I.completion.opt = false;
 		I.completion.run (&I);
-		argc = I.completion.argc;
-		argv = I.completion.argv;
+		argc = r_pvector_len (&I.completion.args);
+		argv = I.completion.args.v.a;
```
ret2libc: Same here

libr/cons/dietline.c
thestr4ng3r: Now there is

libr/util/vector.c
```diff
@@ -230,6 +231,11 @@ R_API void r_pvector_free(RPVector *vec) {
 	free (vec);
 }
 
+R_API void r_pvector_set_free(RPVector *vec, RPVectorFree free) {
```
radare: uhm i think this should be only used ini the construcctor . changing the destructor of a vector after being initializied can result on undesired behaviours if we have added an item previously

libr/include/r_vector.h
```diff
@@ -158,6 +160,16 @@ static inline bool r_pvector_empty(RPVector *vec) {
 	return r_pvector_len (vec) == 0;
 }
 
+// returns a pointer to the offset inside the array where the element of the index lies.
+static inline void **r_pvector_index_ptr(RPVector *vec, size_t index) {
+	return ((void **)vec->v.a) + index;
```
radare: whhat if index exceeds thee vector size, what about returning *left?

thestr4ng3r: What is "left"? It would be the caller's responsibility to check the size here.

libr/util/vector.c
thestr4ng3r: Yeah we don't need it anymore anyway. I removed it.

PR #13871, opened 2019-04-23
Stop analyzing call to reloc and fix pifcj to honor asm.jumpsub

sivaramaaa:
### Before 

rebef

### After 

recaft

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -5648,6 +5648,10 @@ static void _anal_calls(RCore *core, ut64 addr, ut64 addr_end, bool printCommand
 						isValidCall = false;
 					}
 				}
+				RBinReloc *rel = r_core_getreloc (core, addr, op.size);
+				if (rel && (rel->import || rel->symbol)) {
+					isValidCall = false;
+				}
 				if (isValidCall) {
 #if JAYRO_03
 					if (!anal_is_bad_call (core, from, to, addr, buf, bufi)) {
--- a/libr/core/core.c
+++ b/libr/core/core.c
@@ -117,6 +117,42 @@ static void r_core_debug_syscall_hit(RCore *core) {
 	}
 }
 
+struct getreloc_t {
+        ut64 vaddr;
+        int size;
+};
+
+static int getreloc_tree(const void *user, const RBNode *n) {
+        struct getreloc_t *gr = (struct getreloc_t *)user;
+        const RBinReloc *r = container_of (n, const RBinReloc, vrb);
+        if ((r->vaddr >= gr->vaddr) && (r->vaddr < (gr->vaddr + gr->size))) {
+                return 0;
+        }
+
+        if (gr->vaddr > r->vaddr) {
+                return 1;
+        }
+        if (gr->vaddr < r->vaddr) {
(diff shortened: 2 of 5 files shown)


condret:
why stop analyzing calls to reloc?

sivaramaaa:
```
[0x08004b1a]> pd 1
            0x08004b1a      e8fcffffff     call 0x8004b1b  
```

Because in reality it's not an actual call to function , it just call to `$ + 1` ,  

The 0xff bytes in intruction `e8fcffffff` will be only be filled  dynamically

sivaramaaa:
@radare  , @condret and @ITAYC0HEN  , what do u guys think  ?? , 

imo we can stop analyzing this function while not in debug mode . and anlayze them only in debug mode !

ITAYC0HEN:
> imo we can stop analyzing the call to reloc while not in debug mode , and anlayze them only in debug mode !

Agree. The relocs point to functions which aren't in the binary - so I am all for not trying to analyze them as a function - because they are not here

libr/core/core.c
```diff
@@ -117,6 +117,42 @@ static void r_core_debug_syscall_hit(RCore *core) {
...
+        }
+        if (gr->vaddr < r->vaddr) {
+                return -1;
+        }
+        return 0;
+}
+
+// TODO: Use sdb in rbin to accelerate this
+// we shuold use aligned reloc addresses instead of iterating all of them
+R_API RBinReloc *r_core_getreloc(RCore *core, ut64 addr, int size) {
```
radare: this API should be in RBibn, there's nothing RCore specific in here

sivaramaaa: Ok will move it to Rbin !

radare:
lgtm but we are still asuming the legit use of relocs.. because in reality a reloc can modify ANY instruction, not just a call

radare:
maybe we should import the relocs as hints from rbin, and just honor RAnalOp to do whatever we want/need

radare:
If we import this reloc info as a hint we can get rid of all this code and get o(1)

sivaramaaa:
> If we import this reloc info as a hint we can get rid of all this code and get o(1)

sounds good to me , but for another PR :)

radare:
you mean to mrege this for now?

sivaramaaa:
I will make the change u asked on moving the API to Rbin and minor fix , then we can merge this 

And in another pr we can import reloc info as hint !

radare:
ok

libr/core/cmd_anal.c
```diff
@@ -5648,6 +5648,10 @@ static void _anal_calls(RCore *core, ut64 addr, ut64 addr_end, bool printCommand
 						isValidCall = false;
 					}
 				}
+				RBinReloc *rel = r_core_getreloc (core, addr, op.size);
+				if (rel && (rel->import || rel->symbol)) {
+					isValidCall = false;
+				}
 				if (isValidCall) {
 #if JAYRO_03
```
radare: i think this is what JAYRO_03 was  suposed to do, this was speedingi up the analysis on PE quite a biit afair. can you review that too because maybe we can remove this code now

PR #13877, opened 2019-04-23
A panel will be allocated for each help msg when it is called ##panels

Vane11ope:
I personally love this feature. 

people, including me, can use panels mode with the help msg being active all the time if they call it, and they can of course kill it anytime like any other panel.


radare:
wheres's the gif? :D

XVilka:
Better a WebP, since it is supported everywhere now, and without 256 colors
restriction.

radare:
Or mp4, wat...webp is now in mozilla too?

XVilka:
Yes

PR #13878, opened 2019-04-23
Demangle relocs and add asm.flags.{inline|limit} and improve jmpsub #disasm

radare:
IMAGE 2019-04-24 00:16:38
IMAGE 2019-04-24 00:16:42


radare:
all the testss should be updated and i would merge, comments?

PR #13885, opened 2019-04-24
Auto update of help panel synced with every mode change is supported ##panels

Vane11ope:


Vane11ope:
GIFrecord_2019-04-25_003826

PR #13887, opened 2019-04-24
Add disasm field to ao command output.

jace0x21:
Basically copied the code that was being used to generate the disasm field in the JSON output and added that to the standard output.

Addressing issue: https://github.com/radare/radare2/issues/13879

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -1624,9 +1624,9 @@ static void core_anal_bytes(RCore *core, const ut8 *buf, int len, int nops, int 
 		} else {
 		char disasm[128] = { 0 };
 		r_parse_varsub (core->parser, NULL,
-		core->offset + idx,
-		asmop.size, r_asm_op_get_asm (&asmop),
-		disasm, sizeof (disasm));
+			core->offset + idx,
+			asmop.size, r_asm_op_get_asm (&asmop),
+			disasm, sizeof (disasm));
 		ut64 killme = UT64_MAX;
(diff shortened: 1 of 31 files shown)


libr/core/cmd_anal.c
```diff
@@ -1622,6 +1622,21 @@ static void core_anal_bytes(RCore *core, const ut8 *buf, int len, int nops, int
 				r_anal_esil_stack_free (esil);
 			}
 		} else {
+			char disasm[128] = { 0 };
+			r_parse_varsub (core->parser, NULL,
+				core->offset + idx,
+				asmop.size, r_asm_op_get_asm (&asmop),
+				disasm, sizeof (disasm));
+				ut64 killme = UT64_MAX;
```
radare: wrong indent

libr/core/cmd_anal.c
```diff
@@ -1622,6 +1622,21 @@ static void core_anal_bytes(RCore *core, const ut8 *buf, int len, int nops, int
...
+				if (r_io_read_i (core->io, op.ptr, &killme, op.refptr, be)) {
```
radare: warning here, you should read specifying the endiannes , otherwise this wont be portable.

radare:
also update the tests with a pr in r2r

libr/core/cmd_anal.c
jace0x21: Not sure how to deal with endianness here. I thought it was being assumed that it was big endian hence the use of `be`. Any advice?

radare:
You have to commit first in r2r and then in r2.

Also, rebase both repos to get travis green

PR #13891, opened 2019-04-25
Initial work on refactoring RParse

radare:


radare:
i just pushed that commit into the changes pr because otherwise the changes were incomplete

PR #13893, opened 2019-04-25
Fix Warnings

thestr4ng3r:


libr/anal/var.c
```diff
@@ -750,7 +750,7 @@ R_API void r_anal_extract_rarg(RAnal *anal, RAnalOp *op, RAnalFunction *fcn, int
 				r_meta_set_string (anal, R_META_TYPE_VARTYPE, op->addr, vname);
 				free (name);
 				free (type);
-				*count++;
+				count++;
```
radare: thats wrong

libr/asm/asm.c
```diff
@@ -1098,15 +1098,15 @@ R_API RAsmCode* r_asm_rasm_assemble(RAsm *a, const char *buf, bool use_spp) {
 	if (use_spp) {
 		Output out;
 		out.fout = NULL;
-		out.cout = r_strbuf_new ("");
-		r_strbuf_init (out.cout);
+		out.cout = (SStrBuf *)r_strbuf_new ("");
+		r_strbuf_init ((RStrBuf *)out.cout);
```
radare: ideally, spp should have a Bind-like struct to replace this apis at runtime, or maybe compile it like we did before with an ifdef to avoid using that s_ api and use just r_ with some macros

libr/anal/var.c
thestr4ng3r: What is the point of the * in `*count++`?

thestr4ng3r: I guess it should be `(*count)++`?

radare:
y

thestr4ng3r:
Hmm I think those failing tests might actually be wrong, but I'm not sure. @sivaramaaa

radare:
i wanted to keep the strbuf warnings until i the proper fix comes.

radare:
and the regressions look wtf

radare:
ping

radare:
can you resolve  onflicts and update the pr to make it green?

libr/asm/asm.c
```diff
@@ -1110,15 +1110,15 @@ R_API RAsmCode* r_asm_rasm_assemble(RAsm *a, const char *buf, bool use_spp) {
...
 		struct Proc proc;
 		spp_proc_set (&proc, "spp", 1);
 
 		lbuf = replace_directives (lbuf);
 		spp_eval (lbuf, &out);
 		free (lbuf);
-		lbuf = strdup (r_strbuf_get (out.cout));
+		lbuf = strdup (r_strbuf_get ((RStrBuf *)out.cout));
```
radare: i would like to not merge those casts. the correct fix is to make spp use r_util when built inside r2, not to cast like this and in the future get segfaults without much hints to understand the reason. can you drop those casts from the pr? and i'll merge?  ill try to find some time to fix this spp warnings properly before sunday

thestr4ng3r: sure

PR #13899, opened 2019-04-25
Fix asm.cpu config loading

ickromwerk:
https://github.com/radare/radare2/issues/13888

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -505,10 +505,16 @@ static int cb_asmarch(void *user, void *data) {
 		if (newAsmCPU) {
 			if (*newAsmCPU) {
 				char *nac = strdup (newAsmCPU);
-				char *comma = strchr (nac, ',');
-				if (comma) {
-					*comma = 0;
-					r_config_set (core->config, "asm.cpu", nac);
+				if(asm_cpu) {
+					if(!strstr(nac, asm_cpu)) {
+						// set default cpu if asm.cpu is not set or
+						// not present in arch cpus
+						char *comma = strchr (nac, ',');
+						if (comma) {
+							*comma = 0;
+							r_config_set (core->config, "asm.cpu", nac);
+						}
+					}
 				}
 				free (nac);
 			} else {


libr/core/cconfig.c
```diff
@@ -505,10 +505,16 @@ static int cb_asmarch(void *user, void *data) {
 		if (newAsmCPU) {
 			if (*newAsmCPU) {
 				char *nac = strdup (newAsmCPU);
-				char *comma = strchr (nac, ',');
-				if (comma) {
-					*comma = 0;
-					r_config_set (core->config, "asm.cpu", nac);
+				if(asm_cpu) {
+					if(!strstr(nac, asm_cpu)) {
```
radare: add spaces before (

radare:
this change makes 20 tests to fail, can you confirm my order theory?

ickromwerk:
Actually I though about evaluation order before lurking down more. It was first though that when you change arch CPU reset to default one. And this is quite logical that you can specify CPU variant during arch evaluation.

> Many tests fails
That's why I call it quick and dirty :) I have noticed strange behaviour too at the end of the day. Is it possible to know which tests failed?

radare:
another option and easier to solve would be to just hold the value of asm.cpu when set while parsing the script, and run it again after loading the project. can you update the PR in this way?

ickromwerk:
No. I can't update it that way.

I noticed that loading logic is quite complex. First load vars, initializes assembler, then for unknown reason asm.arch bumps. And more and more.

Doing it the way you ask increases complexity and creates spaghetti code.

It should work this way: arch initializes -> loads asm plugin -> initialize asm plugin -> initialize cpu. if 'asm.cpu' already set during asm plugin cpu initialization -- it should be used.

I will investigate that further. I think that there is no reason to haste here ;)

radare:
+1 thanks

radare:
ping?

radare:
ehoo!

ickromwerk:
I have a tight schedule right now because I missed deadline for work project and people are waiting till I finish it. So its hard to invest more time to r2 project now.

If this PR messing things a bit it iz safe to delete it and continue discussion in 'issues' section.

PR #13909, opened 2019-04-27
n/N should behave like the ones in visual which are for go next nkeys

Vane11ope:


libr/core/panels.c
```diff
@@ -169,13 +170,14 @@ static const char *help_msg_panels[] = {
 	"L",        "scroll panels right by page",
 	"m",        "select the menu panel",
 	"M",        "open new custom frame",
-	"nN",       "create new panel with given command",
-	"pP",       "seek to next or previous scr.nkey",
+	"n/N",      "seek next/prev function/flag/hit (scr.nkey)",
```
radare: wtf then how do you create new panels?, imho n is very intuitive.. so i woudlnt merge this until we find a new keys for it

Vane11ope: it's only in windows mode because that's something related to only panels and not visual. default mode is only for "visual" and all those panels related commands go to windows mode.

radare: oh i see... that makes sense.. but ill have to change my workflow..

PR #13915, opened 2019-04-27
Dword

radare:


radare:
should be green now. ready to merge

PR #13916, opened 2019-04-27
Little ELF segment labelling change ##bin

devnexen:
Name to NULL rather than UNKNOWN where no size.

### DIFF
--- a/libr/debug/p/native/linux/linux_coredump.c
+++ b/libr/debug/p/native/linux/linux_coredump.c
@@ -44,7 +44,7 @@ static bool is_a_kernel_mapping(const char *name) {
 		&& strcmp (name, "[vvar]")
 		&& strcmp (name, "[heap]")
 		&& strcmp (name, "[vectors]")
-		&& strncmp (name, "[stack", sizeof ("[stack") - 1));
+		&& strncmp (name, "[stack", strlen ("[stack")));
 }
 
 static char *prpsinfo_get_psargs(char *buffer, int len) {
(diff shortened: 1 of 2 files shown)


devnexen:
One due test failing to fix. Going to ...

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -44,7 +44,7 @@ static bool is_a_kernel_mapping(const char *name) {
 		&& strcmp (name, "[vvar]")
 		&& strcmp (name, "[heap]")
 		&& strcmp (name, "[vectors]")
-		&& strncmp (name, "[stack", strlen ("[stack")));
+		&& strncmp (name, "[stack", sizeof ("[stack") - 1));
```
radare: sizeof??

devnexen: `        size_t a = sizeof("[stack") - 1;
        size_t b = strlen("[stack");
        printf("%zu %zu\n", a, b);`

Both give 6.

radare:
on widechar strings that wont be correct, we use strlen for this everywhere

radare:
also, rebase and update tests

radare:
i have not much preferences for NULL vs UNKNOWN, but NULL looks more like an error to me

devnexen:
Ok fair enough i can do the change, not an issue to me was just to make the user comfortable with known namings.

radare:
can you make this PR green?

libr/debug/p/native/linux/linux_coredump.c
radare: yeah, but sizeof will be affected by the encoding of the string and we use strlen for this case

devnexen: Sorry I forgot to change it back...

PR #13917, opened 2019-04-28
Fixes #13880: Added dt=

### LINKED ISSUE #13880: Implement dt= like we have afl=, etc
Grouping them by contiguous blocks of code traced

deepakchethan:


libr/debug/trace.c
```diff
@@ -119,6 +239,13 @@ R_API void r_debug_trace_list (RDebug *dbg, int mode) {
 			}
 		}
 	}
+	if (flag) {
+		RCore *core = dbg->corebind.core;
+		r_list_sort (info_list, cmpaddr);
```
radare: You cant use rcore from rdebug

libr/debug/trace.c
```diff
@@ -98,16 +99,135 @@ R_API RDebugTracepoint *r_debug_trace_get (RDebug *dbg, ut64 addr) {
...
+		}
+	}
+}
+
 R_API void r_debug_trace_list (RDebug *dbg, int mode) {
 	int tag = dbg->trace->tag;
 	RListIter *iter;
+	bool flag = false;
+	RList *info_list = r_list_new ();
+	if (!info_list && mode == '=') {
```
radare: thats probably never going to happen. info_list. better use || instead of &&

libr/debug/trace.c
```diff
@@ -119,6 +239,13 @@ R_API void r_debug_trace_list (RDebug *dbg, int mode) {
 			}
 		}
 	}
+	if (flag) {
+		RCore *core = dbg->corebind.core;
+		r_list_sort (info_list, cmpaddr);
+		r_core_visual_list (core, info_list, core->offset, core->blocksize,
```
radare: i wouldn't go that picky and i would use blocksize = 1 and core->offset passed as argument, so you dont need to add extra dpeendencies

libr/debug/trace.c
```diff
@@ -98,16 +99,135 @@ R_API RDebugTracepoint *r_debug_trace_get (RDebug *dbg, ut64 addr) {
...
+				ListInfo *info = R_NEW0 (ListInfo);
+				char temp[8];
+				if (!info) {
+					return;
+				}
+				info->pitv = (RInterval) {trace->addr, trace->size};
+				info->vitv = info->pitv;
+				info->perm = -1;
+				snprintf (temp, sizeof (temp), "%d", trace->times);
+				info->name = strdup (temp);
```
radare: use r_str_newf instead of snprintf+strdup

libr/debug/trace.c
```diff
@@ -98,16 +99,135 @@ R_API RDebugTracepoint *r_debug_trace_get (RDebug *dbg, ut64 addr) {
...
 	RDebugTracepoint *trace;
 	r_list_foreach (dbg->trace->traces, iter, trace) {
 		if (!trace->tag || (tag & trace->tag)) {
 			switch (mode) {
 			case 'q':
 				dbg->cb_printf ("0x%"PFMT64x"\n", trace->addr);
 				break;
+			case '=': {
+				ListInfo *info = R_NEW0 (ListInfo);
+				char temp[8];
```
radare: overflows. just use r_strnewf and kill this var

libr/debug/trace.c
deepakchethan: > i wouldn't go that picky and i would use blocksize = 1 and core->offset passed as argument, so you dont need to add extra dpeendencies

But you said not to use RCore in RDebug. core->offset is from RCore isn't it ?

libr/debug/trace.c
deepakchethan: The idea is that info_list is only required when the mode is dt=. So I think && is the right op here?

libr/debug/trace.c
deepakchethan: Updated this

libr/debug/trace.c
deepakchethan: Okay, updated.

PR #13918, opened 2019-04-28
Initial support for tabs ##panels

Vane11ope:
Basic workflow:

1. t/T open/close a tab (And the original t/T, which were for rotating sub commands, had been safely moved to i/I, as 'i' had been only used when the cursor was on.

2. f/F forward/backward a tab, which I will improve next (...and of course follow the visual style, which means enabling t[n] keys

3. Q closes all the tabs at once and save all the settings of each panels. 

Please use this "Initial support for tabs ##panels" when you squash and merge, thanks.

GIFrecord_2019-04-28_152006


XVilka:
Please add a few notes, or at least an issue in radare2 book about that.
Thanks!

PR #13921, opened 2019-04-28
Make flag name shorter by removing DLL name from it

pelijah:


radare:
The dllname matters because in windows apps can import the same symbol name from different dlls. Because this is implicit for the linker. Theres an option to trim flag names in the disasm
That maybe its better option to solve it in there than in here. Otherwise we are losing information

pelijah:
Could you provide binary or at least link to docs?

Maijin:
Agree should be optional and with an eval, that's a useful piece of information to be able to display when needed.

pelijah:
> 
> 
> Agree should be optional and with an eval, that's a useful piece of information to be able to display when needed.

You still can do `ii~FUNCTIONNAME`

pelijah:
> Theres an option to trim flag names in the disasm That maybe its better option to solve it in there than in here.

Disagree. Replacing flag name in disasm can negatively affect UX in Cutter where you make double click on a flag's name string to jump to it.

radare:
Why close if its not solved

pelijah:
@radare So what about binary and docs?

pelijah:
Tbh current implementation will break names of functions imported by ordinal but it's easy to fix.

radare:
Which is your proposal?

pelijah:
We can make an exception for them and leave DLL name for uniqueness.

radare:
this will be useful for java/dalvik/c++/swift too where the package + class name are also in the flag name. because its useful information, but too noisy to read sometimes

radare:
@ITAYC0HEN comments on this? @pelijah can you rebase and make this thing green even if requiring a config var to toggle this option?

ITAYC0HEN:
My suggestion is going for a more generic solution which will help with removing the annoying prefixes from displayed flags. Us GUI people, hate all the verbosity of the flags and prefer a more straight-forward and quiet approach, where only the name displayed by default (or by tweaking eval) is the last name in the current flag name. So instead of `sym.imp.kernel32.VirtualAllocEx` it will be only `VirtualAllocEx`.

To do this, the idea is to treat flag names in three parts:

```
flag->name   // VirtualAllocEx
flag->prefix   // sym.imp
flag->additional   // kernel32
```

Then a function in r_flag or something would be responsible for combining these parts depends on the value of the different evals. This function would be used where want to display a flag. And each user would be able to enjoy their own preset.

Same goes for other prefixes.

CC @Maijin @thestr4ng3r @xarkes @ret2libc 




----

Quiet and Clear:
image


Ugliness & Noise & Mess:
image

thestr4ng3r:
If prefix is separate, it should probably be a bitmask, enum or something and not a string. Also, this would make identifying them much harder if we continue to use the name as the id. If the structure should be changed, I would rather go with something like this:
```
ut64 id; // unique
char *name;
ut64 flags; // bitmask of imp, sym, whatever
char *module; // or something else, like additional
```
but that's just an idea.

radare:
Flags have a realname field which may hold what you want. The problem is that the same import can be loaded trom two different libraries and withiut that libname prefix your disasm will be misleading.

Im happy to have an option to toggle that behaviour

ITAYC0HEN:
@radare so what about Florian's proposal? can we decide on this refactoring and let's begin the work on it?At the end, this would be a toggled option so people who prefer the current state would be able to get back to it easily.

radare:
i already said that im happy to have an option disabled by default to have such feature so we can test it and comment it with real use cases instead of just theorical bases

radare:
btw my last commit  (cb94f900760709bad4b7c537b069173b3e731b8d) may extend this issue to objc/c++/dalvik bins, so we need a generic solution for all of them

radare:
See the new `afii` command (function imports) this can be a generic solution for all this in a more explicit way. looking for feedback

thestr4ng3r:
This is actually not a bad idea, but I have yet to try `afii` myself

radare:
the only problem in the current approach is that "imports" are bound to the function, will be good to have some "global" imports, or "by module"

radare:
@thestr4ng3r @pelijah can we close this issue? is the afii working as you expect? improvements? comments?

thestr4ng3r:
This is not an issue, it's a pr.

radare:
I know

radare:
Can we close this pr?

thestr4ng3r:
I think afii and aii should be enough for at least the case of dll names, but that's for @pelijah and @ITAYC0HEN to decide. For other cases that might be remaining, the discussion should be moved to a new issue.

ITAYC0HEN:
@pelijah can you give Florian's solution a try?

pelijah:
How to use `afii`?

thestr4ng3r:
Bildschirmfoto vom 2019-07-17 21-43-06

pelijah:
Ok, it works. But this requires from user to call `aii` for every DLL. I would prefer to have one config var.

radare:
i was proposing to initialize those imports from RBin or in RCore.bin to automatically define them at function and file-level

PR #13932, opened 2019-04-29
Fix Buffer Overflow in ds_instruction_mov_lea()

### LINKED ISSUE #13931: Segfault in pd
[TVMLKit.tar.gz](https://github.com/radare/radare2/files/3127986/TVMLKit.tar.gz)
`r2 -c "aa; pd 1 @ 0x000e0dc3" TVMLKit`

thestr4ng3r:
Fix #13931

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -2796,13 +2796,21 @@ static void ds_instruction_mov_lea(RDisasmState *ds, int idx) {
 			RAnalValue *dst = ds->analop.dst;
 			if (dst && dst->reg && dst->reg->name) {
 				if (src->reg->name && pc && !strcmp (src->reg->name, pc)) {
-					RFlagItem *item;
-					ut8 b[8];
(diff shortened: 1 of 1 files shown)


libr/core/disasm.c
```diff
@@ -2796,13 +2796,21 @@ static void ds_instruction_mov_lea(RDisasmState *ds, int idx) {
 			RAnalValue *dst = ds->analop.dst;
 			if (dst && dst->reg && dst->reg->name) {
 				if (src->reg->name && pc && !strcmp (src->reg->name, pc)) {
-					RFlagItem *item;
-					ut8 b[8];
```
radare: its because its an xmmword.. what about making this memref to be always using the heap to read the contents instead of having that extra complexity.. in  here..

also r_mem_get_num is not handling more than 8 byte size numbers...  so that wont be useful at all in here. because xmmwords cant be used to reference memory. and there are no pointers bigger than 8 in size.. so we can probably skip that path

radare:
just fixed it in master

PR #13940, opened 2019-05-01
Decompiler default off for everything especially tabs working better …

Vane11ope:
…this way ##panels

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -38,12 +38,12 @@ typedef enum {
 } Direction;
 
 static const char *panels_dynamic [] = {
-	"Disassembly", "Decompiler", "Stack", "Registers",
+	"Disassembly", "Stack", "Registers",
 	NULL
 };
 
 static const char *panels_static [] = {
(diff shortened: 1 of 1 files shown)


radare:
wat. the decompiler output is very useful. even more if you dont use pdc.

e cmd.pdc=pdd when r2dec is installed

Vane11ope:
i know. i did not say it's not useful, i say it's wrong to open it as a default panel by asking users if they want open it or not, like severely annoying to get asked about that everytime especially when you open multiple tabs and move around them, even though it's only first time when you move to brand new tab.

the reason why im not opening pdc by default is it's sometimes severely slow. Users can choose to open them by * command or in menu. Im strongly sticking to not open it by default.

radare:
Ok but its still there the option to have it ?

Vane11ope:
i mean you can open decompiler easily with '*' command or in menu, I just disabled it to be open by default because it just slows down the users.

radare:
Lgtm!

PR #13941, opened 2019-05-01
avoid aap if it's debug

Vane11ope:


radare:
I think we should change search.in and anal.in the cfg.debug changes

Vane11ope:
agreed. this fix is like super temporal solution anyway, just to avoid r2 -dAA /bin/ls to not loop infinitely. looked into code and there should be better and proper fix for this.

radare:
Yes i know. But i would prefer to have the proper fix. Because using io.maps ranges in the debugger is kind of a pain.. in fact we should be able yo detect when io.maps.x is used and this range covers the whole 64bit address space show a warning and return an empty range instead to avoid such undesired events to happen

radare:
fixed in https://github.com/radare/radare2/pull/13957

radare:
fixed now

PR #13949, opened 2019-05-03
Filter is supported with command f/F ##panels

Vane11ope:
GIFrecord_2019-05-03_223420


radare:
f should be to set/unset flags but the feature looks cool

PR #13962, opened 2019-05-05
Add unsigned decimal output to ? command

Phude:
The non-JSON version of the command still differentiates between inputs greater/less than 32 bits and includes only the appropriate output.


radare:
conflcts with https://github.com/radare/radare2/pull/13963

can you make the PR green?

PR #13974, opened 2019-05-06
Fixed a few things

Vane11ope:
tab number starts off 1 to n

and also fixed a bug with which r2 will crash if you push 0 when selecting nth tab

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -4193,16 +4193,17 @@ static RPanels *get_cur_panels(RPanelsRoot *panels_root) {
 }
 
 static bool handle_tab(RCore *core) {
-	char msg[128];
+	char *msg;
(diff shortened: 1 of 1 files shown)


libr/core/panels.c
```diff
@@ -4193,10 +4193,16 @@ static RPanels *get_cur_panels(RPanelsRoot *panels_root) {
 }
 
 static bool handle_tab(RCore *core) {
+	char msg[128];
```
radare: dont use static buffers. use r_str_newf or so, but if u want static buffer, make it of the screen width, not a fixed 128 length

Vane11ope: ah okay hang on

Vane11ope:
done

PR #13979, opened 2019-05-06
R anal op var for arm

### LINKED ISSUE #13970: Implement RAnalOp.var support for ARM
https://github.com/radare/radare2/pull/13771#issuecomment-489566905

bart1e:
Filled op->src[0] and op->dst for LOAD, STORE, ADD and SUB in ARM. Some tests may not be passed, but I suppose the behaviour is correct (although I'm not an expert with ARM, so I would ask you to verify that), except 'db/cmd/types jni and SoP' where I'm not sure.
One more important thing: since ADD and SUB are working differently on different architectures (in x86 sub rax, rbx uses values of both registers, but on ARM SUB usually just overwrites destination without using its value), I had to add new `if `to rargs detection switch, and it checks for architecture used. If you have any better idea how to solve this problem, please share it. :) 
fix #13970

### DIFF
--- a/libr/anal/p/anal_arm_cs.c
+++ b/libr/anal/p/anal_arm_cs.c
@@ -58,8 +58,8 @@
 #define ISPREINDEX64() ((OPCOUNT64() == 3) && (ISMEM64(2)) && (ISWRITEBACK64()))
 #define ISPOSTINDEX64() ((OPCOUNT64() == 4) && (ISIMM64(3)) && (ISWRITEBACK64()))
 
-static RRegItem base_regs[2];
-static RRegItem regdelta_regs[2];
+static RRegItem base_regs[4];
+static RRegItem regdelta_regs[4];
 
 static const ut64 bitmask_by_width[] = {
 	0x1, 0x3, 0x7, 0xf, 0x1f, 0x3f, 0x7f, 0xff, 0x1ff, 0x3ff, 0x7ff,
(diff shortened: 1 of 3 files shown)


bart1e:
Support for other operations will be added today or tomorrow. Btw, there is a lot of [XX] on OSX. Do you have any idea what is going wrong there?

radare:
Probably the same that was happening in x86? Didnt had time to chk this pr yet. I have 2 more prs to finish before having time for more

libr/anal/p/anal_arm_cs.c
```diff
@@ -58,6 +58,9 @@
 #define ISPREINDEX64() ((OPCOUNT64() == 3) && (ISMEM64(2)) && (ISWRITEBACK64()))
 #define ISPOSTINDEX64() ((OPCOUNT64() == 4) && (ISIMM64(3)) && (ISWRITEBACK64()))
 
+static RRegItem base_regs[4];
```
radare: avoid globals

libr/anal/p/anal_arm_cs.c
```diff
@@ -3108,32 +3125,67 @@ static void set_opdir(RAnalOp *op) {
...
-		ZERO_FILL (reg);
-		op->dst->reg = &reg;
-		if (bits == 64) {
-			parse_reg64_name (op->dst->reg, handle, insn, 1);
-			op->dst->delta = MEMDISP64(1);
-		} else {
-			parse_reg_name (op->dst->reg, handle, insn, 1);
-			op->dst->delta = MEMDISP(1);
-		}
+		set_src_dst (op->dst, &handle, insn, 1, bits);
```
radare: uhm isnt this 0, and 1, instead of 1 and 0?

bart1e: I suppose no; str in ARM has source as first operand as far as I know.

radare:
right, its like this, so nvm

just did a cheap review without testing your code :P 

i finished 1 pr, must finish the other one and then ill chk this one, can you have another look and see why the broekneess?

bart1e:
> right, its like this, so nvm just did a cheap review without testing your code :P i finished 1 pr, must finish the other one and then ill chk this one, can you have another look and see why the broekneess?

It's because in some cases it detects new arguments, but in other ones some regs which were previously are not detected as such now. I think the behaviour is correct though. I'm only not sure about this test I mentioned in description.

libr/anal/p/anal_arm_cs.c
bart1e: OK, I'll change that.

radare:
Yeah i think the new output is correct. considering that the arg ananlysis is far from perfect . can you rebase this PR to get green appveyour and send a PR to r2r updating the tests?

thanks!

radare:
yep, fix this before updating the r2r

libr/anal/p/anal_arm_cs.c
bart1e: Is there any better way to do that other than dynamically allocating op regs? I don't want to do that this way, because it won't be freed by r_anal_op_fini (it's TODO in r_anal_value_free). Of course, if you wish, I may fix it, so it frees also RRegItems and implement it this way.

radare:
Using the heap will be slower but more correct and safer. We can also enlarge the ranalop struct or refactor it but this will take more time. 

Few places we handle more than one ranalop at the same time and having this as global may result in unexpected behaviours

bart1e:
OK, so I’ll implement RRegItem free. But there are global regs used in op_fillvals for several other architectures, so I may change them as well.

radare:
👍

bart1e:
There is one problem with this approach I'm afraid. I cannot free regs in r_anal_value_free since not all are allocated dynamically (in some places they are initialised by r_reg_get). Possible solution would be to implement something like r_anal_reg_free and call it only while there was called r_anal_op with R_ANAL_OP_MASK_VAL. Do you have any better idea how to resolve that?

radare:
uhm yeaah thats a bit scary maybe worth delaying this to 3.6, release is this sunday

radare:
merging, as long as this reg-free issue is also happening on x86 and the output of the tests look better.

PR #13993, opened 2019-05-07
Fixpanelsbug

KoWu:
when loading a .r2panels terminate the string properly.


Vane11ope:
wtf why you merged this seriously. wtf.

PR #13994, opened 2019-05-07
Improve visual mark  ##visual

Vane11ope:
GIFrecord_2019-05-08_035708

I'm not sure if you like this, but i thought it would make visual mark more usable this way, so...

### DIFF
--- a/libr/core/vmarks.c
+++ b/libr/core/vmarks.c
@@ -14,16 +14,19 @@ R_API void r_core_visual_mark_reset(RCore *core) {
 	}
 }
 
-R_API void r_core_visual_mark_dump(RCore *core) {
+R_API bool r_core_visual_mark_dump(RCore *core) {
 	int i;
+	bool out = false;
 	if (!marks_init) {
-		return;
+		return out;
 	}
 	for (i = 0; i < UT8_MAX; i++) {
 		if (marks[i] != UT64_MAX) {
 			r_cons_printf (R_CONS_CLEAR_LINE"fV %c 0x%"PFMT64x"\n", i, marks[i]);
+			out = true;
 		}
 	}
+	return out;
 }
 
 R_API void r_core_visual_mark_set(RCore *core, ut8 ch, ut64 addr) {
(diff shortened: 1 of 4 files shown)


radare:
vim works like r2 does right now, with no visual feedback.

one of the problems i see in this mode is the noise. and the problem of not being able to delete any visual mark, we can only create new ones. so the list will be longer and longer and at some point scroll the screen, not sure how much nice will be in this case.

also i have been thinking in using boxes centered in the screen with the text inside, this can be nice for the usser, but its anoying for the eye because it will hide the disasm or whataever is in the center which use to be important. But can be handy to have such thing in panels for confirm dialogs or so, what do you think?

radare:
why travis is red

Vane11ope:
ah yeah about that centered box, how about not center but bottom, like the way cutter shows console widget? and about deleting the shortcut off the list, and even all clear, shouldnt be difficult. i can do it tonight.

Vane11ope:
what would you say?

radare:
we can just have the primitive with an option to specify where to display it and do some experiments with it, but unrelated to this PR

Vane11ope:
yup exactly. so for now i guess im just adding delete shortcut key options and done with this PR

radare:
did you checked why travis is red?

Vane11ope:
yeah of course i will take care of that tonight too :)

Vane11ope:
made visual marks deletable tho, now let's see if travis is still angry

libr/core/vmarks.c
```diff
@@ -14,16 +14,19 @@ R_API void r_core_visual_mark_reset(RCore *core) {
...
 	int i;
+	bool out = false;
 	if (!marks_init) {
-		return;
+		return out;
 	}
 	for (i = 0; i < UT8_MAX; i++) {
 		if (marks[i] != UT64_MAX) {
-			r_cons_printf ("fV %d 0x%"PFMT64x"\n", i, marks[i]);
+			r_cons_printf (R_CONS_CLEAR_LINE"fV %c 0x%"PFMT64x"\n", i, marks[i]);
```
Vane11ope: ah travis is angry because of this change. I changed the output to %c from %d because I wanted to use the API to tell users which shortcuts are available at the moment and in that case, it doesn't make sense if it's shown like 97 instead of 'a' because I assume people don't memorize the whole ascii table.

I can fix the test if you confirm this change, or tell me what to do please @radare

radare:
%c is fine if its printable. Add a check for this to avoid issues when using non ascii chars in here. Probably fV needs to support int arg too

Vane11ope:
yup. done without fixing tests. in fact, the tests were right.

Vane11ope:
lol

Vane11ope:
Thanks travis love ya

PR #13995, opened 2019-05-07
Revert "Fixpanelsbug"

Vane11ope:
Reverts radare/radare2#13993


Vane11ope:
seriously. WTF

radare:
Syeah it was clearly wrong. We should ask for a reproducer

KoWu:
Just append a newline at the .r2panels and it will segfault. This happens if you edit it with vim for example. I don't get the problem.

Vane11ope:
okay thanks yeah fair point i will do something about it sooner or later :)

PR #13997, opened 2019-05-08
Code cleanup for anal.ex, make java work without it.

radare:
* Current java implementation of the analysis is not working well
* RBin parser now puts the right section sizes
* Added anal.ex option to use extensions or not
* More to go


radare:
at least now with -e anal.ex=false, the java analysis works fine

PR #14008, opened 2019-05-09
Use HtPP instead of Sdb in r_bin_filter_sym function

ret2libc:


ret2libc:
@radare please test, but I don't think it will really be that much faster than before... After all, Sdb is using Ht underneath, so...

radare:
yeah we just save some string processing in here

PR #14009, opened 2019-05-09
More ^C handling in RCons, but still not working with agf because of the nestness nature of it

radare:


libr/core/agraph.c
```diff
@@ -3083,6 +3089,7 @@ static void agraph_print_edges(RAGraph *g) {
 		}
 	}
 
+	// XXX why not use r_list_free ??
```
lowlyw: why just comment, why not do it? ;)

libr/cons/canvas_line.c
```diff
@@ -434,6 +434,9 @@ R_API void r_cons_canvas_line_back_edge (RConsCanvas *c, int x, int y, int x2, i
 		r_cons_canvas_line (c, x, y, x2, y2, style);
 		return;
 	}
+	if (r_cons_is_breaked ()) {
```
lowlyw: this isn't in a loop. why do we need it here?

radare:
@lowlyw the problem is that each consbreak/push/pop is breaking at one level only, if the caller checks again for that it will fail to interrupt all the task, only a slow loop inside. so ^C is not fixed in agf. Do you want to finish this PR?

radare:
closing for now, preformance is better now and i dont see he need to bloat those loops

PR #14012, opened 2019-05-09
Ability to download capstone git as zip

XVilka:
To solve the problem of building without `git clone`.

https://github.com/radare/radare2/issues/10152

@radare please check this one. It activated like `CS_COMMIT_ARCHIVE=1 make`

### DIFF
--- a/shlr/capstone.sh
+++ b/shlr/capstone.sh
@@ -3,6 +3,7 @@ CS_URL="$1" # url
 CS_BRA="$2" # branch name
 CS_TIP="$3" # tip commit
 CS_REV="$4" # revert
+CS_ARCHIVE="$5" # download archived tip
 CS_DEPTH_CLONE=512
 
 git --help > /dev/null 2>&1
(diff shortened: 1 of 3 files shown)


shlr/capstone.sh
```diff
@@ -32,11 +33,27 @@ parse_capstone_tip() {
 	BRANCH="$(git rev-parse --abbrev-ref HEAD)"
 }
 
-clone_capstone() {
+download_archive() {
+	echo '[capstone] Downloading capstone snapshot...' >&2
+	wget -O .cs_tmp.zip "$CS_ARCHIVE" || fatal_msg 'Cannot download archived capstone'
```
radare: wget is being removed  from many operating systems, we should support curl too

XVilka:
@radare what do you mean by "an option"?

radare:
A way for the user to choose

XVilka:
@radare - user can choose by setting the environment variable:
```bash
CS_COMMIT_ARCHIVE=1 make
```

PR #14013, opened 2019-05-09
mdmp: fix incorrect data setting in pe sections

AlexKornitzer:
This addresses what was a long standing bug but it only appeared in more recent builds of r2. I believe this is also an issue in the `bin_pe.inc` but I did not want to pollute this PR and it may have been indended.

### DIFF
--- a/libr/bin/format/mdmp/mdmp_pe.c
+++ b/libr/bin/format/mdmp/mdmp_pe.c
@@ -5,7 +5,7 @@
 
 #include "mdmp_pe.h"
 
-static void PE_(add_tls_callbacks)(struct PE_(r_bin_pe_obj_t) *bin, RList* list) {
+static void PE_(add_tls_callbacks)(struct PE_(r_bin_pe_obj_t) * bin, RList *list) {
 	char *key;
 	int count = 0;
 	PE_DWord haddr, paddr, vaddr;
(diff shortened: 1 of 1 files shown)


libr/bin/format/mdmp/mdmp_pe.c
```diff
@@ -187,32 +189,30 @@ RList *PE_(r_bin_mdmp_pe_get_sections)(struct PE_(r_bin_mdmp_pe_bin) *pe_bin) {
...
 		}
-#define X 1
-#define ROW (4 | 2)
-		if (ptr->perm & ROW && !(ptr->perm & X) && ptr->size > 0) {
-			if (!strcmp (ptr->name, ".rsrc") ||
-				!strcmp (ptr->name, ".data") ||
-				!strcmp (ptr->name, ".rdata")) {
-					ptr->is_data = true;
-				}
+		if ((ptr->perm & R_PERM_R) && !(ptr->perm & R_PERM_X) && ptr->size > 0) {
```
AlexKornitzer: This block is where the fix is, the rest of the code is from formatting fixes by `indent.sh`

AlexKornitzer:
Furthermore I was alerted by @XVilka that apparently the MDMP plugin is returning trash output, but the `iz` bug is the only issue I could find. Could someone maybe provide an example. IMO the regressions are sufficient as they are not broken...

radare:
can you add a test for this fix?

AlexKornitzer:
Will do that now, do you have any further information about your original comment of `its completely broken`?

AlexKornitzer:
https://github.com/radare/radare2-regressions/pull/1824

radare:
uhm i  dont know of any iz bug showing trassh in mdmp, it actually sohws nothing on any of the mdmp files we have, os i guess, its snot searcihng in thhe right bounds

AlexKornitzer:
So I was referring to this set of comments but I was not sure what they were related to, if you have files that are not opening correctly if you are able to share them I can take a look. It could be that the dumps being used did not have the memory streams in which case the data will be trash as only metadata will be (i.e. data needed for debugging/replicating).

image

PR #14016, opened 2019-05-09
Properly handle ELFs with no entrypoint defined ##bin

radare:


radare:
cc @ITAYC0HEN

radare:
despite the new behaviour shouuld be more correct there are a 54 XX and im not in the mood to review/fix that stuff before the release beccause i can miss something, feel free to review and test

ret2libc:
@radare do you have a test case for this new thing?

ITAYC0HEN:
[pngwutil.zip](https://github.com/radare/radare2/files/3167403/pngwutil.zip)

libr/bin/format/elf/elf.c
```diff
@@ -1828,30 +1827,30 @@ ut64 Elf_(r_bin_elf_get_init_offset)(ELFOBJ *bin) {
...
 ut64 Elf_(r_bin_elf_get_entry_offset)(ELFOBJ *bin) {
-	ut64 entry;
 	if (!bin) {
 		return 0LL;
 	}
-	entry = bin->ehdr.e_entry;
+	ut64 entry = bin->ehdr.e_entry;
 	if (!entry) {
+		return UT64_MAX;
+	}
```
kazarmy: This close bracket changes semantics if `entry != 0`.

kazarmy: Add check for ELF relocatable file (ET_REL -- this is what the test case is) or just remove the code that sets `entry` to a section start?

radare:
This fix wont come today, the proeblm i see is that it takes section.text as initial address when entry is resolved at vaddr 0, even when the paddr != 0

libr/bin/format/elf/elf.c
radare: that was kind of a poc, we may not have an entry0 if paddr is 0

ret2libc:
I'm not sure I understand what is the issue with the binary provided by @ITAYC0HEN . What' the problem? It's a .o binary, does it matter what is the entrypoint? Moreover, it seems r2 already identifies an entrypoint anyway. I don't get it, sorry.

radare:
np ill finish this isssue, focuss on finishing your rbuffer pr

ret2libc:
I will, but I want to understand anyway :) What is the current issue with the binary provided for test? Can anybody provide a reproducer?

radare:
the binary have no entrypoint at all and r2 says there's one.

thats the problem my pr is fixing, but its making many tests to fail, as well as showing other regresssions that i have to fix

ret2libc:
> the binary have no entrypoint at all and r2 says there's one.

readelf says entrypoint is at address 0, and objdump shows `png_save_uint_32` at address 0.
radare2 does the same, so IMHO radare2 is right as well.

radare:
readelf is wrong, the entrypoint set to 0 means there's no entrypoint, and png_save_uint_32 is clearly not any entrypoint

ret2libc:
> readelf is wrong, the entrypoint set to 0 means there's no entrypoint, and png_save_uint_32 is clearly not any entrypoint

ok! got it.

PR #14020, opened 2019-05-09
Fix dc and F9 ##panels

Vane11ope:
run dc nor F9 which internally runs dc does not update panels at all and not it does with this


Vane11ope:
this is just because traivs took too much time, bummer. im closing opening again to test it again....

radare:
Shouldnt it  run .dr* to update flags?

Vane11ope:
...........that's what exactly i added, as you can see in the code.

PR #14032, opened 2019-05-10
Naming tabs is supported ##panels

Vane11ope:


libr/core/panels.c
```diff
@@ -4208,29 +4223,29 @@ static RPanels *get_cur_panels(RPanelsRoot *panels_root) {
...
-		}
-		break;
-	case 'p':
-		return handle_tab_prev (core);
-	case '-':
-		return handle_tab_del (core);
-	case 't':
-		return handle_tab_new (core);
+		case 'n':
+			return handle_tab_next (core);
```
radare: Wrong indent

Vane11ope: lol it's not wrong imho but okay if you say so. hang on.

PR #14033, opened 2019-05-10
Fix lgtm.yml

samlanning:
It looks like c/c++ builds are currently failing with the current configuration, but that they succeeded in the past with this configuration: ([see build logs](https://lgtm.com/projects/g/radare/radare2/logs/rev/18b069ccf43cbb5e7ce17dd6276ea84389ecf958/lang:cpp/stage:Build%20child_18b069ccf43cbb5e7ce17dd6276ea84389ecf958), view "Analysis configuration")

```
extraction:
    cpp:
        configure:
            command: ./configure --with-syscapstone
```

At the time, it looks like the configuration was set on the LGTM side, and was not present in your repository, but since adding one, builds have started failing.

cc @XVilka

*(full disclosure, I work for Semmle)*

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -1,11 +1,4 @@
 extraction:
     cpp:
-        after_prepare:
-            - export GNU_MAKE=make
         configure:
-            command: "./configure"
-        before_index:
-            - export CS_COMMIT_ARCHIVE=1
-        index:
-            build_command:
-                - $GNU_MAKE -j2 -s
+            command: ./configure --with-syscapstone


samlanning:
Looks like the build for c/c++ was successful, which you can see here: https://lgtm.com/projects/g/radare/radare2/logs/rev/pr-2f43ed06a0a1c191cc82ec5fbd6b0c70594ca71e/lang:cpp/stage:Build%20master_8cb2ad402ba42d3e3239922e475be9352e36b425

(the check for this PR will continue to fail though, as the build of the parent (base) commit is still required so results can be compared).

samlanning:
For the record, this kind of minimal configuration for lgtm works as our autobuilder is able to detect the build system. You can find out more here: https://lgtm.com/help/lgtm/cpp-extraction

XVilka:
I disabled system capstone for a reason. Please don't merge, will fix capstone build today.

XVilka:
Now it is fixed properly (by using capstone git snapshot). Thank you!

PR #14034, opened 2019-05-11
Track pushed register (note: immediately after lea/mov) to resolve push-ret

### LINKED ISSUE #13845: Support push-ret on x86-64 and reg dereference
anal.pushret

Using this binary https://github.com/radare/radare2-regressions/commit/f601fdeb5045a453bce5c0def0f3444048c2c281

SrimantaBarua:
Somewhat fixes #13845 .

This only works if the previous instruction before push is a reg mov / lea. A more generic solution would be tracking registers inside the RAnal struct, but I think ESIL should handle that.

Also, sorry for the long absence... real life is a bitch sometimes.

### DIFF
--- a/libr/anal/p/anal_x86_cs.c
+++ b/libr/anal/p/anal_x86_cs.c
@@ -1873,8 +1873,8 @@ static void op_fillval(RAnal *a, RAnalOp *op, csh *handle, cs_insn *insn) {
 		set_src_dst (op->src[1], handle, insn, 2);
 		set_src_dst (op->src[2], handle, insn, 3);
 		break;
-	case R_ANAL_OP_TYPE_PUSH:
-		if (INSOP (0).type == X86_OP_REG) {
+	case R_ANAL_OP_TYPE_UPUSH:
+		if ((op->type & R_ANAL_OP_TYPE_REG)) {
 			CREATE_SRC_DST (op);
 			set_src_dst (op->src[0], handle, insn, 0);
 		}
@@ -2410,7 +2410,7 @@ static void anop(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len, csh 
 			op->cycles = CYCLE_REG + CYCLE_MEM;
 			break;
 		case X86_OP_REG:
-			op->type = R_ANAL_OP_TYPE_PUSH;
+			op->type = R_ANAL_OP_TYPE_RPUSH;
 			op->cycles = CYCLE_REG + CYCLE_MEM;
 			break;
 		default:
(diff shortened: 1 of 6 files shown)


radare:
Cool i was about to implement this after release, because thst works already if you do emu.write with asm.emu, but it can be solved statically without emulation in a better way.

Thanks!

radare:
I was about to add anal.pushret var. do u think it makes sense to have it or not?

SrimantaBarua:
The anal.pushet var already exists.
The static analysis alteast in this commit is not costly, the information is already available. Also I don't see a scenario where you would not want to evaluate the pointed basic block. If you are thinking of having this as a var, maybe we can have it enabled by default.

radare:
Some scary XXs

Screenshot 2019-05-11 at 17 46 33

SrimantaBarua:
I glanced over the regressions. Looks like most are upush -> push. I thought upush == unknown push or something like that. Should I keep this as upush or update the tests to be push?

radare:
yeah upush is for push reg things, because you dont know at static time the value of that register. when you are analyzing you are assuming a single execution path, so you do an assumption that this register will have a specific value, but any other part of the program can reference and jump that push reg and become a different value. so yep it should be kept as UPUSH

libr/anal/p/anal_x86_cs.c
```diff
@@ -2403,6 +2409,10 @@ static void anop(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len, csh
 			op->type = R_ANAL_OP_TYPE_PUSH;
 			op->cycles = CYCLE_REG + CYCLE_MEM;
 			break;
+		case X86_OP_REG:
+			op->type = R_ANAL_OP_TYPE_PUSH;
```
radare: this is TYPE_UPUSH, there's a bitmask for REG which is used for RJMP, and can be also used here, but will require adding the rpush type

SrimantaBarua:
The current regressions look to be changing type_num from 0x0c (upush) to  0x100000c (rpush). The type_name is still "upush", like rjmp etc. Is that correct?

libr/anal/p/anal_x86_cs.c
SrimantaBarua: Added RPUSH which is UPUSH | REG. Should printed type be "upush" (like in rjmp, printed type is "ujump")?

radare:
Should be rpush i think

SrimantaBarua:
Changed. You could look at [#1829](https://github.com/radare/radare2-regressions/pull/1829) on the r2r repository to see changes to tests

PR #14043, opened 2019-05-12
Automatically choose wget/curl/git depending on availability ##build

radare:
* wget/curl downloads use CS_ARCHIVE
* code cleanup in capstone.sh, still lot to trim


XVilka:
But why do not apply patches if capstone uses archive? Nothing prevents from applying for them.

radare:
the patches were applied from the makefile, i just removed that part from the makefile and cleaned up the capstone.sh  a bit more

PR #14053, opened 2019-05-13
Autocompletion in RCons

deepakchethan:
* Handled the case of newlines
* had to fix formatting from prev commit


libr/cons/cons.c
```diff
@@ -359,30 +359,37 @@ R_API bool r_cons_is_breaked() {
 	return I.context->breaked;
 }
 
+R_API void r_cons_add_newlines (int n) {
+	int i;
+	for (i = 0; i < n; ++i) {
```
radare: dont use preincrements unless they a re strictly  necessary

radare: What about using  printf ("%s\n", r_str_pad('\n', n));

deepakchethan: Yeah, This seems like a simpler solution. I'll go with this :)

PR #14055, opened 2019-05-13
Fixes #13767: Improves p--

### LINKED ISSUE #13767: Improve p--
* [ ] Use theme colors
* [ ] Honor scr.width
* [ ] Make it work in vertical too like scr.scrollbar=1

Screenshot 2019-04-16 at 11 05 43

deepakchethan:
* Honors scr.width
* Honors themes

* Since it honors scr.width, is scr.scroll necessary ?

Update: Screenshot after the change

Screen Shot 2019-05-14 at 6 01 01 AM


libr/cons/pal.c
```diff
@@ -54,6 +54,9 @@ static struct {
 	{ "func_var_addr", r_offsetof (RConsPrintablePalette, func_var_addr), r_offsetof (RConsPalette, func_var_addr) },
 	{ "widget_bg", r_offsetof (RConsPrintablePalette, widget_bg), r_offsetof (RConsPalette, widget_bg) },
 	{ "widget_sel", r_offsetof (RConsPrintablePalette, widget_sel), r_offsetof (RConsPalette, widget_sel) },
+	{ "bg_true", r_offsetof (RConsPrintablePalette, bg_true), r_offsetof (RConsPalette, bg_true) },
```
radare: uhm, no theme will use those new keys, i would rather reuse some other fields that we have, even if thoses have no bg color set, otherwise you'll have to update all the themes

deepakchethan: Yes you are right. Since the color is based on the permissions. I will just reuse the colors used in the visual method.

radare:
while at it.. add also the colors for the popup widget in this pr

PR #14056, opened 2019-05-13
remove err dp error message, which is triggered for no reason by afu

astuder:
When using afu on code with jumps and calls, this error message is triggered. May be related to the fact, that afu does not iterate down into the referenced functions, which is as intended in my situation.


radare:
debug printfs \o/

PR #14073, opened 2019-05-14
Fix includes from to ignoring dir.types ##types

oxcabe:
Fixes https://github.com/radare/radare2/issues/14031

### DIFF
--- a/libr/core/cmd_type.c
+++ b/libr/core/cmd_type.c
@@ -1226,6 +1226,7 @@ static int cmd_type(void *data, const char *input) {
 			r_core_cmd_help (core, help_msg_to);
 		} else if (!r_sandbox_enable (0)) {
 			if (input[1] == ' ') {
+				const char *dir = r_config_get (core->config, "dir.types");
 				const char *filename = input + 2;
 				char *homefile = NULL;
 				if (*filename == '~') {
@@ -1252,7 +1253,7 @@ static int cmd_type(void *data, const char *input) {
 					}
 				} else {
 					char *error_msg = NULL;
-					char *out = r_parse_c_file (core->anal, filename, &error_msg);
+					char *out = r_parse_c_file (core->anal, filename, dir, &error_msg);
 					if (out) {
 						//r_cons_strcat (out);
 						r_anal_save_parsed_type (core->anal, out);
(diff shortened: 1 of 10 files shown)


shlr/tcc/tccpp.c
```diff
@@ -1600,8 +1600,10 @@ ST_FUNC void preprocess(int is_bof)
 			memcpy (filepath, file->filename, filepath_len);
 			strcpy (filepath + filepath_len, buf);
 			if (tcc_open (s1, filepath) < 0) {
-				int len = snprintf (filepath, sizeof (filepath),
-					"/usr/include/%s", buf);
+				if (!dirname) {
+					dirname = "/usr/include";
```
radare: random though #1: maybe /usr/include is not where we want to look at.. maybe it's just '.' ?

libr/core/cmd_type.c
```diff
@@ -1252,7 +1253,7 @@ static int cmd_type(void *data, const char *input) {
 					}
 				} else {
 					char *error_msg = NULL;
-					char *out = r_parse_c_file (core->anal, filename, &error_msg);
+					char *out = r_parse_c_file (core->anal, filename, dir, &error_msg);
```
radare: random though #2 - what if we want to add more includedirs? or maybe other options to the parser? the rparse api needs a refactor. and this must be added to the list of things to do at some point

shlr/tcc/tccpp.c
oxcabe: I wrote that as a fallback, in case dirname wasn't already set as the value on dir.types (it would be a null pointer in that case).
Also, "/usr/include" is the default value for dir.types. Based on C include priorities, it will prioritize a header with the same name located in the current directory than the header file located in the dir.types path

oxcabe: Thats's what I thought while writing that. Might be wrong, though! :p

libr/core/cmd_type.c
oxcabe: Can we already set more than one includedir in the eval var, I mean, is that already implemented?
I believe it shouldn't be way too hard to pass a dirlist as the argument, to be parsed by tcc in order to add many include directories.

oxcabe: wdyt?

PR #14076, opened 2019-05-15
Fix Color_GRAY/BGGRAY

kazarmy:
Gray isn't one of the original 8 ANSI colors ([link](https://en.wikipedia.org/wiki/ANSI_escape_code#3/4_bit)).

### DIFF
--- a/libr/cons/pal.c
+++ b/libr/cons/pal.c
@@ -456,13 +456,13 @@ static void r_cons_pal_show_rgb() {
 
 R_API void r_cons_pal_show() {
 	int i;
-	for (i = 0; colors[i].name; i++) {
 #if __WINDOWS__
-		if (r_cons_singleton ()->ansicon
+	bool ansicon = r_cons_singleton ()->ansicon;
 #else
-		if (true
+	bool ansicon = true;
 #endif
-		    && !strcmp (colors[i].name, "gray")) {
+	for (i = 0; colors[i].name; i++) {
+		if (ansicon && !strcmp (colors[i].name, "gray")) {
 			continue;
 		}
 		r_cons_printf ("%s%s__"Color_RESET" %s\n",
--- a/libr/cons/pal.c
+++ b/libr/cons/pal.c
@@ -457,7 +457,12 @@ static void r_cons_pal_show_rgb() {
 R_API void r_cons_pal_show() {
 	int i;
 	for (i = 0; colors[i].name; i++) {
-		if (r_cons_singleton ()->ansicon && !strcmp (colors[i].name, "gray")) {
+#if __WINDOWS__
+		if (r_cons_singleton ()->ansicon
+#else
+		if (true
+#endif
+		    && !strcmp (colors[i].name, "gray")) {
 			continue;
 		}
 		r_cons_printf ("%s%s__"Color_RESET" %s\n",
--- a/libr/cons/pal.c
+++ b/libr/cons/pal.c
@@ -456,15 +456,7 @@ static void r_cons_pal_show_rgb() {
 
 R_API void r_cons_pal_show() {
 	int i;
-#if __WINDOWS__
-	bool ansicon = r_cons_singleton ()->ansicon;
-#else
-	bool ansicon = true;
-#endif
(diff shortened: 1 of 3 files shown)


libr/cons/pal.c
```diff
@@ -457,6 +457,9 @@ static void r_cons_pal_show_rgb() {
 R_API void r_cons_pal_show() {
 	int i;
 	for (i = 0; colors[i].name; i++) {
+		if (r_cons_singleton ()->ansicon && !strcmp (colors[i].name, "gray")) {
```
radare: that will hide not just the color but also the entry in the palette. it is correct?

kazarmy: no, the entry is still in the palette a.k.a. `colors` table

kazarmy: it can't be removed since at least the `basic` eco theme uses it

kazarmy: the entry for gray in the `colors` table should be changed if ansicon, but that's a different pr

libr/cons/pal.c
```diff
@@ -457,6 +457,14 @@ static void r_cons_pal_show_rgb() {
 R_API void r_cons_pal_show() {
 	int i;
 	for (i = 0; colors[i].name; i++) {
+#if __WINDOWS__
+		if (r_cons_singleton ()->ansicon
```
radare: just put the && in here and drop the true. in fact it will b e cleaner if you move this #if block outside the function and define a VAR that is ansicon or not. and this can  be done outside the loop to save some cycles.

The problem i see  is that the original PR was checking only if ansicon was set. and this changes the behaviour beacuse now its also affecting non-windows platforms and gray will  never be displayed. what's exactly the problem with this sgray color not being printed? if its a  bug in their console, should be reported

kazarmy: > beacuse now its also affecting non-windows platforms and gray will never be displayed

That's the whole point. `Color_GRAY` is currently defined as:
https://github.com/radare/radare2/blob/418c1d95c634d05b7296472eb083da93404bc8a6/libr/include/r_cons.h#L556
which is wrong (notice there's no 38 [in the table here](https://en.wikipedia.org/w/index.php?title=ANSI_escape_code&oldid=895282563#3/4_bit)). It does get shown on non-ansicon windows because there is code in `r_cons_w32_print()` that handles that (incorrect) escape sequence:
https://github.com/radare/radare2/blob/418c1d95c634d05b7296472eb083da93404bc8a6/libr/cons/output.c#L274-L276

kazarmy: > ... if you move this #if block outside the function and define a VAR that is ansicon or not ...

Need to get the value of ansicon when `r_cons_pal_show()` is executed though, and not before since it can change dynamically.

kazarmy: > define a VAR that is ansicon or not. and this can be done outside the loop to save some cycles.

ok 5767b440f22c5e0ef9400dc1b41cd746e498050e

XVilka:
LGTM. No need to overthink about this - it can be always rolled back in case of the troubles.

libr/cons/pal.c
radare: ok correct, the 38m doesnt exist.. then why are we having it? in xterm/therm/... it shows no change in the color. so if you try this: `$ printf "\x1b[32m\x1b[38mGRAY\x1b[0m"` the text is in GREEN. so the terminals basically ignore that escape code. imho you should remove this Color_GRAY. because white is the closest color we can have in here

radare: the fix is imho not skipping that thing in a loop but properly killing all the reference to Color_GRAY as an ansi 16color code

radare:
this is related https://github.com/radare/radare2/issues/14046

libr/cons/pal.c
kazarmy: The `colors` table is actually a mapping from color names to values and therefore Color_GRAY should stay since gray is a color ... fixed it in e5a4420a2b96889e4eeff4c6597e20e2cda7b8f4

kazarmy:
ok to merge?

PR #14077, opened 2019-05-15
Fix for rip-relative lea

### LINKED ISSUE #13835: LEA assembler issues
* [ ] lea rdi, [rip + 0x1011] -> should be 488d3d11100000 -> but we get 488db811100000 -> which is lea rdi, [rax + 0x1011]
* [ ] $ rasm2 -b64 -d `rasm2 'lea rax, [rip + 0xff9]'` => lea rcx, [rax + 0xff9]

SrimantaBarua:
Fixes #13835  for the asm_x86_nz assembler. 

Btw my system nasm (2.14.02) also has this bug, `lea rdi, [rip + 0x1011]` is assembled to `488db811100000`


radare:
Can you add an "asm" test in r2r?

PR #14095, opened 2019-05-17
Adds axm command

roboman2444:
Copies the cross refs pointing to ADDR to also point to curseek/AT.

Useful for dealing with wrapper functions.

Help text:

"axm addr [at]   copy data/code references pointing to addr to also point to curseek (or at)"

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -6209,7 +6209,7 @@ static bool cmd_anal_refs(RCore *core, const char *input) {
 		free (fInput);
 	} break;
 	case 'm': { // "axm"
-		RList *list = NULL;
+		RList *list;
 		RAnalRef *ref;
 		RListIter *iter;
 		char *ptr = strdup (r_str_trim_head ((char *)input + 1));
@@ -6230,12 +6230,11 @@ static bool cmd_anal_refs(RCore *core, const char *input) {
 		//get all xrefs pointing to addr
 		list = r_anal_xrefs_get (core->anal, addr);
 		r_list_foreach (list, iter, ref) {
-			r_cons_printf ("0x%p %s\n", ref->addr, r_anal_xrefs_type_tostring(ref->type));
+			r_cons_printf ("0x%"PFMT64x" %s\n", ref->addr, r_anal_xrefs_type_tostring (ref->type));
 			r_anal_xrefs_set (core->anal, ref->addr, at, ref->type);
 		}
 		r_list_free (list);
 		free (ptr);
-
 	} break;
 	case 't': { // "axt"
 		RList *list = NULL;
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -6230,9 +6230,10 @@ static bool cmd_anal_refs(RCore *core, const char *input) {
 		//get all xrefs pointing to addr
 		list = r_anal_xrefs_get (core->anal, addr);
 		r_list_foreach (list, iter, ref) {
-			r_cons_printf("0x%p %s\n", ref->addr, r_anal_xrefs_type_tostring(ref->type));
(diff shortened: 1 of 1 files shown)


radare:
Notwhich would be the use case for this. i can think on bypassing jumptables destinations.. by shortening paths in the dependency graph. But we can do this dinamically, i mean, we can have some new apis to get all the XRefs at address X, where X is the given adddress as well as the REF of it in case there's only 1. But this wont be  correct completely. So thats why im asking for a practical example of this command.

roboman2444:
I've used this when trying to reverse something that did weird stuff with dlsym and wrapper functions. R2 couldn't figure out that the wrapper function was eventually calling the real function (even with emulation), so i had to manually make the xrefs for it.

radare:
im thinking that this command/apis would fit for the need to rebase analysis information when reloading the process in the debugger or when changing the base address in static..

libr/core/cmd_anal.c
```diff
@@ -6207,6 +6208,35 @@ static bool cmd_anal_refs(RCore *core, const char *input) {
...
+			addr = r_num_math (core->num, r_str_word_get0 (ptr, 0));
+			break;
+		default:
+			free (ptr);
+			return false;
+		}
+		//get all xrefs pointing to addr
+		list = r_anal_xrefs_get (core->anal, addr);
+		r_list_foreach (list, iter, ref) {
+			r_cons_printf ("0x%p %s\n", ref->addr, r_anal_xrefs_type_tostring(ref->type));
```
radare: mnissing space before (

libr/core/cmd_anal.c
```diff
@@ -6207,6 +6208,35 @@ static bool cmd_anal_refs(RCore *core, const char *input) {
...
+		/* fall through */
+		case 1: // get addr
+			addr = r_num_math (core->num, r_str_word_get0 (ptr, 0));
+			break;
+		default:
+			free (ptr);
+			return false;
+		}
+		//get all xrefs pointing to addr
+		list = r_anal_xrefs_get (core->anal, addr);
```
radare: define and assign in the same line

libr/core/cmd_anal.c
radare: use PFMT64X instead pf %p, this is not portable

libr/core/cmd_anal.c
```diff
@@ -6207,6 +6208,35 @@ static bool cmd_anal_refs(RCore *core, const char *input) {
 		free (tInput);
 		free (fInput);
 	} break;
+	case 'm': { // "axm"
+		RList *list = NULL;
```
radare: no need to assign to null

libr/core/cmd_anal.c
```diff
@@ -6207,6 +6208,35 @@ static bool cmd_anal_refs(RCore *core, const char *input) {
...
+		}
+		//get all xrefs pointing to addr
+		list = r_anal_xrefs_get (core->anal, addr);
+		r_list_foreach (list, iter, ref) {
+			r_cons_printf ("0x%p %s\n", ref->addr, r_anal_xrefs_type_tostring(ref->type));
+			r_anal_xrefs_set (core->anal, ref->addr, at, ref->type);
+		}
+		r_list_free (list);
+		free (ptr);
+
```
radare: empty line

libr/core/cmd_anal.c
radare: we can probably p`rint the r2 commands to achieve the same instead of printing and calling the api, or at least when axm* is used

roboman2444: > we can probably p`rint the r2 commands to achieve the same instead of printing and calling the api, or at least when axm* is used

Not entirely sure what you mean by this. Can you give an example where this is done elsewhere?

PR #14102, opened 2019-05-18
Enable color prompt on Windows

kazarmy:
Works also with **cmd.exe**.


radare:
#if __WINDOWS__ here and there to fix the build

kazarmy:
sorry e101acbe0e9481189ecacbd718a0b25b28d4b31d

PR #14108, opened 2019-05-19
Implement Console frame in visual panels ##panels

radare:


Vane11ope:
in fact this doesn't keep the manners well but you can still merge this, means the functionality looks okay so i will just make this fit better in the codebase later :)

PR #14112, opened 2019-05-20
Do not double define typedef RBuffer

ret2libc:
It should be enough to define struct r_buf_t, since the typedef to
RBuffer is already specified.


ret2libc:
It fixes compilation on FreeBSD

unixfreaxjp:
> It fixes compilation on FreeBSD

Testing on NetBSD and OpenBSD too now. If any problem I will posted the detail in here.
Thank you very much!

radare:
we need a CI agent running on BSD for each commit and pullreq. btw this error was actually a warning in clang

PR #14116, opened 2019-05-20
Implement r_malloc, r_free, etc. in .c

thestr4ng3r:
This is necessary for Cutter, because we must match the right `free()` for memory allocated from r2. Just calling free directly causes crashes on Windows.


thestr4ng3r:
Closing because there is `r_mem_free()`

PR #14117, opened 2019-05-20
Initial import of the RISCV capstone plugins for asm and anal ##disasm

radare:


radare:
Before making --with-capstone5 into r2 we need a decent way to determine when captone supports X or Y, becaue right now the cstool -v | grep is kind of hacky mauinly because distros dont ship cstool and that will break --with-syscapstone

PR #14123, opened 2019-05-21
Do not close IO when RBuffer is freed

### LINKED ISSUE #14115: Calling izz closes the io fd and breaks subsequent calls
To reproduce:
```
r2 -n /bin/ls
> izz
> izz
```

ret2libc:
Fixes #14115


radare:
why freeing a buffer asasociated with an fd should close the fd? imho rbuf should not handle that.

ret2libc:
Ok it seems a good point @radare ! Maybe it's better to just avoid the free at all, given it is not the buffer that creates the file it shouldn't be the buffer to close/free it. Nice catch, I will update the PR soon.

radare:
+1

PR #14126, opened 2019-05-21
Ao pseudo

radare:
Note the new "description" and "pseudo" entries (also available in JSON)

Screenshot 2019-05-21 at 12 45 36

also added the 'mask' field for debugging `r_anal_mask()`

Screenshot 2019-05-21 at 12 51 26


radare:
This r_anal_mask thing should be inside r_anal_op when a specific flag is set R_ANAL_OP_MASK_BITS or somthing

PR #14138, opened 2019-05-22
Refactor and Complete Anal Hint Printing

thestr4ng3r:
* Removed Code Duplication between `cb()` and `r_core_anal_hint_print()`
* Using pj instead of directly printing
* jump, fail and ptr are now directly printed as integers in json
* Fixed `ahj [addr]`

Tests: https://github.com/radare/radare2-regressions/pull/1836


thestr4ng3r:
should be ok now

PR #14141, opened 2019-05-23
Fix some RBin file format crashes

ret2libc:
I also switched RBuffer to st64/ut64 instead of size_t so at least we have same sizes on both 32bit and 64bit archictures.


radare:
very red

libr/include/r_util/r_buf.h
```diff
@@ -73,30 +73,30 @@ R_API RBuffer *r_buf_new_sparse(ut8 Oxff);
 /* methods */
 R_API bool r_buf_dump(RBuffer *buf, const char *file);
 R_API bool r_buf_set_bytes(RBuffer *b, const ut8 *buf, ut64 length);
-R_API int r_buf_append_string(RBuffer *b, const char *str);
+R_API st64 r_buf_append_string(RBuffer *b, const char *str);
```
radare: st64?

ret2libc:
What don't you like about the changes? The use of st64/ut64 is to match what `read`/`write`/etc. do
See:
```
ssize_t read(int fd, void *buf, size_t count);
ssize_t write(int fd, const void *buf, size_t count);
```
But instead of using size_t/ssize_t which are arch-dependent, i'm using ut64/st64.

radare:
I dont think ut64for size  will work on reads for anyone becausE that will be a reallynhuge and slow syscall.

Also, on 32 bits, this arg will require 2 regs or more size in the stack to fit them, and thats the reasson why size_t exists

ret2libc:
> I dont think ut64for size will work on reads for anyone becausE that will be a reallynhuge and slow syscall. 

Indeed you should always check your length before passing them to a read-like functions. Unsigned integers are already used for read/write operations (see `man 2 read`/`man 2 write`).

> Also, on 32 bits, this arg will require 2 regs or more size in the stack to fit them, and thats the reasson why size_t exists

Yes, that is a good point. I thought we preferred fixed type sizes, more than arch-dependent ones like `int` or `size_t`. But if you like `size_t`, i can just use `size_t` and `ssize_t` instead of `ut64` and `st64`.

ret2libc:
BTW, I'm more for having ut64/st64 since in general it seems we do use those types instead of size_t/ssize_t. For example, most of the section/segment/symbol/etc. sizes are in ut64, so having size_t in RBuffer would just cause a lot of errors (or in the best case casts) IMO. cc @ITAYC0HEN

radare:
pls rebase and make it green

radare:
ping

ret2libc:
@radare this is green now! and it fixes all crashes you sent me

radare:
I think ill just merge that and we can keep yhe size vs ut64 discussion for later. Pls push the samples that was making r2 cry so we can be sure we dont regress again

ret2libc:
@radare already done yesterday

PR #14146, opened 2019-05-23
Fix project's asm.cpu

ickromwerk:
It actually break some tests, BUT it fix https://github.com/radare/radare2/issues/13888

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -506,9 +506,11 @@ static bool cb_asmarch(void *user, void *data) {
 			if (*newAsmCPU) {
 				char *nac = strdup (newAsmCPU);
 				char *comma = strchr (nac, ',');
-				if (comma && !strstr(nac, asm_cpu)) {
-					*comma = 0;
-					r_config_set (core->config, "asm.cpu", nac);
+				if (comma) {
+					if (!*asm_cpu || (*asm_cpu && !strstr(nac, asm_cpu))) {
+						*comma = 0;
+						r_config_set (core->config, "asm.cpu", nac);
+					}
 				}
 				free (nac);
 			} else {
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -506,7 +506,7 @@ static bool cb_asmarch(void *user, void *data) {
 			if (*newAsmCPU) {
 				char *nac = strdup (newAsmCPU);
 				char *comma = strchr (nac, ',');
-				if (comma) {
+				if (comma && !strstr(nac, asm_cpu)) {
 					*comma = 0;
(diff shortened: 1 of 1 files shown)


ickromwerk:
Can someone explain how to read this tests output?
https://travis-ci.org/radare/radare2/jobs/536443584#L2257-L2260

When i check it manually with "rasm2 -a arm -b 16 abf3b2ea"
i get "blx 0x3ab568"

should it be "invalid"?

radare:
i dont think this change is correct either. and yes this disasm is invalid

ickromwerk:
If you think that this change is invalid you can provide some information to support your opinion.

I agree that patch is dirty but it makes functionality work for sure. I can explain what my reasoning behind that but don't think it is necessary because you do not provide any help and no other people are talk here. Probably that function is not in a priority for most of them.

This patch is correct but it might be incomplete. It does not check asm_cpu variable for null string for example. But this don't looks very dangerous here. Actually it more complete than current radare2 implementation because it checks whether asm cpu exists in arch. (in master you can issue something like "e asm.cpu=blabla" and it will eat it).

I checked project loading functionality with this patch and it works great! It restores asm.cpu even after r_config_bump(..., "asm.arch", ...)

The problem that test suite finds is very hard to trace and I will be very glad if you(or someone more competent) can help with locating it, because:

I made two versions of libr_core. One with patch applied and one without. Then i run test suite exactly at asm/arm_16 like this:

```
$(node npm)/r2r db/asm/arm_16
```

and got the same output for original and patched radare2! :
```
[**]                  db/asm/arm_16   OK 1293 BR   20 XX    0 FX    4
```

So the arisen problem is much more subtle than it occurs at first glance. But this patch fixes an issue. Prove me wrong if you can.

ickromwerk:
So I figured out how to make arm_16 tests fail. If you run them directly they will work fine.

If you run them in test suite, actually db/asm they will run after arm.gnu_cortex_16 and fail then.

So I think that tests fail now because program work as intended but test suite was written when program worked bad.

Old behaviour: if you bump asm.arch it resets asm.cpu to default of that arch
New behaviout: if you bump asm.arch it keeps asm.cpu if it is set and valid for that architecture

Test suite must be updated to whether specify exact asm.cpu for each arch or reset that variable to default value("") before changing asm.arch

ickromwerk:
Here PR to fix test chain:
https://github.com/radare/node-r2r/pull/1

radare:
"I can explain what my reasoning behind that but don't think it is necessary because you do not provide any help and no other people are talk here."

what?... im helping everyone, reviewing prs, answering issues, chats, mails.. 24/7... why shouldnt I help here?

PR #14166, opened 2019-05-27
Fix ansicon detection with UTF enabled

GustavoLCR:


kazarmy:
Gah functions in example code don't have the A suffix 🤦‍♂

PR #14183, opened 2019-05-29
Implement memoization optimization for RIOMaps (50s -> 2s) ##io

radare:


ret2libc:
What is the test binary used to measure the time? Also, I'd personally prefer to make it better in general by using an rbtree, instead of relying on a cache, which would bring problems like forgetting to invalidate/update it, etc.

radare:
The 65k sections pe from r2r

ret2libc:
I see. Yes, using an rbtree would solve that particular case and it will improve the average case as well, without the need of a cache which complicates the code IMHO.

lowlyw:
yeah, imo the caching should be a worst case situation, i'd rather see the perf compared with the rbtree setup, caching is hard to use properly and will likely lead to lots of bugs that har very hard to diagnose.

radare:
Are you up to do this rbtree refactoring? Now that we know this will heavily improve the performance in many situations

Other things we can do is to resort the list to have the most requested sections on top so it will take less to resolve them but there will be still slow paths

XVilka:
It reminds me of memoization implementations in:
- [Prolog](https://www.metalevel.at/prolog/memoization)
- [Adapton](https://docs.rs/adapton/0.3.30/adapton/macro.memo.html) - implementation of [incremental computing](https://en.wikipedia.org/wiki/Incremental_computing)

For inspiration.

radare:
i have fixed the dyldcache/kernelcache/macho issues related to segments vs sections. so its now loading and running fast, becases the amount of maps is pretty low now. but still this issue is there. so will be good to fix it using rbtree or so. closing this PoC PR

PR #14185, opened 2019-05-29
#define ENABLE_VIRTUAL_TERMINAL_PROCESSING if it's not defined

kazarmy:
Blindfix for https://github.com/radare/radare2/issues/14101#issuecomment-496665065. It appears to be [an acceptable way](https://www.google.com/search?q=ENABLE_VIRTUAL_TERMINAL_PROCESSING+Windows.h) of doing things.


kazarmy:
Repro'd https://github.com/radare/radare2/issues/14101#issuecomment-496665065 and tested that the fix works on a `--backend vs2017 --xp` build.

PR #14196, opened 2019-05-31
Remove last load/load_bytes methods

ret2libc:
I guess the time has come to remove load/load_bytes from RBinPlugin :)

Hopefully this will allow to clean even further the `r_bin_object_new` function and others as well in the future.

### DIFF
--- a/libr/bin/p/bin_sms.c
+++ b/libr/bin/p/bin_sms.c
@@ -32,8 +32,8 @@ static bool check_buffer(RBuffer *b) {
 	return false;
 }
 
-static bool load_bytes(RBinFile *bf, void **bin_obj, const ut8 *buf, ut64 sz, ut64 loadaddr, Sdb *sdb) {
-	return check_buffer (bf->buf);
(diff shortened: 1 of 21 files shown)


radare:
so in total, the whole refactoring caused about -600LOC :D

libr/bin/p/bin_sms.c
```diff
@@ -103,7 +103,7 @@ RBinPlugin r_bin_plugin_sms = {
 	.name = "sms",
 	.desc = "SEGA MasterSystem/GameGear",
 	.license = "LGPL3",
-	.load_bytes = &load_bytes,
+	.load_buffer = &load_buffer,
```
radare: the next step could be to do load_buffer -> load and check_buffer -> check :?

ret2libc: yeah it seems reasonable.

ret2libc: let's wait maybe a bit, just to not cause confusion

radare:
Yeah we can wait one release for this breaking change so we can confuse api changes twice :D we have to break more apis \o/

PR #14198, opened 2019-05-31
Use _exit() in pipe child

thestr4ng3r:
Using `exit()` here leads to atexit functions and C++ destructors to be called in the child process, which is incorrect and in Cutter ultimately makes the forked child processes to be stuck forever even after exiting Cutter.

Maybe `r_sys_exit_immediately()` will need some ifdefs since `_exit()` is not part of C standard, but right now I don't know a platform we support that doesn't have it.

Also maybe it is necessary to flush some streams before the call, as `man _exit` says:
```
Open stdio(3) streams are not flushed.
```
But perhaps someone else here can say more about whether this is necessary here.

### DIFF
--- a/libr/include/r_util/r_sys.h
+++ b/libr/include/r_util/r_sys.h
@@ -26,8 +26,8 @@ R_API void r_sys_set_environ(char **e);
 R_API ut64 r_sys_now(void);
 R_API const char *r_time_to_string (ut64 ts);
 R_API int r_sys_fork(void);
-// immediately = false => exit(); true => _exit()
-R_API void r_sys_exit(int status, bool immediately);
+// nocleanup = false => exit(); true => _exit()
+R_API void r_sys_exit(int status, bool nocleanup);
 R_API bool r_sys_stop(void);
 R_API char *r_sys_pid_to_path(int pid);
 R_API int r_sys_run(const ut8 *buf, int len);
--- a/libr/include/r_util/r_sys.h
+++ b/libr/include/r_util/r_sys.h
@@ -26,7 +26,8 @@ R_API void r_sys_set_environ(char **e);
 R_API ut64 r_sys_now(void);
 R_API const char *r_time_to_string (ut64 ts);
 R_API int r_sys_fork(void);
-R_API void r_sys_exit_immediately(int status); // _exit(int status)
+// immediately = false => exit(); true => _exit()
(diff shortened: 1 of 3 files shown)


libr/include/r_util/r_sys.h
```diff
@@ -26,6 +26,7 @@ R_API void r_sys_set_environ(char **e);
 R_API ut64 r_sys_now(void);
 R_API const char *r_time_to_string (ut64 ts);
 R_API int r_sys_fork(void);
+R_API void r_sys_exit_immediately(int status); // _exit(int status)
```
radare: what about r_sys_exit() instead?

thestr4ng3r: That would suggest it to be regular `exit()`, not `_exit()`.

thestr4ng3r:
@radare ping

radare:
Do we really want the regular exit?

thestr4ng3r:
what?

radare:
wtf, there's even _Exit() O_o... what about having r_sys_exit() to  accept 2 arguments, the first with the return code and the second with an enum specifying the behaviour we want to take from it.

* exit -> standard libc 
* _Exit -> exit without calling atexit
* _exit -> the real exit. the libc exit flushes buffers, closes streams, unlinks temporary files, etc. before calling this,

radare:
why do you want to call _exit() and then flush() ? imho you should use _Exit() (uppercase)

xarkes:
In my libc it seems that `_Exit` is just an alias for `_exit`, they are the same functions, so it shouldn't matter.

thestr4ng3r:
Main difference between `_Exit()` and `_exit()` is that `_Exit()` is part of C99 standard (which we don't require). So I would suggest the following:
```
void r_sys_exit(int status, bool immediately);
```
where
* immediately = false => regular `exit()`
* immediately = true => `_exit()`/`_Exit()`, i.e. no atexit calls and no guarantee that streams are flushed.

libr/util/sys.c
```diff
@@ -116,6 +116,14 @@ R_API int r_sys_fork() {
 #endif
 }
 
+R_API void r_sys_exit(int status, bool nocleanup) {
```
radare: we should probably have a R_NORETURN at some point, but meh .lets merge this

PR #14207, opened 2019-06-02
Fix #14203 - Fix grep sorting

### LINKED ISSUE #14203: Reversed grep is wrong
### Work environment

Any

### Issue
When a user is using the `~$!` it expects radare2 to show reversed order of lines. As stated here:

```
[0x00005ae0]> ~?
...
...
|  $!           sort in inverse order
...

```

In reality, the list of lines which are passed to `r_list_reverse` is being reversed starting from the second line, while the first line stays in place.

### Expected behavior
The output should be reversed without passing the first item

### Actual behavior
Notice how the first item stayed in its place
```
[0x00005ae0]> ?e 1\n2\n3\n4
1
2
3
4

[0x00005ae0]> ?e 1\n2\n3\n4~$!
1
4
3
2
```

Or with a real output:

```
# Normal output
[0x00005ae0]> pxW 16
0x00005ae0 0xfa1e0ff3
0x00005ae4 0x8949ed31
0x00005ae8 0x89485ed1
0x00005aec 0xe48348e2

# Reversed output
[0x00005ae0]> pxW 16~$!
0x00005ae0 0xfa1e0ff3
0x00005aec 0xe48348e2
0x00005ae8 0x89485ed1
0x00005ae4 0x8949ed31
```

- Open any binary.
- Execute `?e 1\n2\n3\n4` to see a list of 4 numbers, each per line
- Now execute `?e 1\n2\n3\n4~$!` and see that the first line wasn't reversed (should be the last one)

GustavoLCR:
Fix #14203

### DIFF
--- a/libr/cons/grep.c
+++ b/libr/cons/grep.c
@@ -789,7 +789,7 @@ R_API int r_cons_grep_line(char *buf, int len) {
 		if (!unsorted_lines) {
 			unsorted_lines = r_list_newf (free);
 		}
-		if (cons->lines > grep->sort_row) {
+		if (cons->lines >= grep->sort_row) {
 			r_list_append (sorted_lines, strdup (buf));
 		} else {
 			r_list_append (unsorted_lines, strdup (buf));


radare:
Thanks! Can you add some tests?

GustavoLCR:
will do

PR #14239, opened 2019-06-07
Mark r_bin_class_new/free as internal

kazarmy:


kazarmy:
Fixes the AppVeyor build.

PR #14255, opened 2019-06-09
Initial implementation GetSingleBlock for Windows heap

GustavoLCR:
And some bug fixes

### DIFF
--- a/libr/core/windows_heap.c
+++ b/libr/core/windows_heap.c
@@ -211,7 +211,7 @@ static bool GetNextHeapBlock(PDEBUG_HEAP_INFORMATION heapInfo, PHeapBlock hb) {
 
 	WPARAM flags;
 	if (block[index].extra & EXTRA_FLAG) {
-		flags = (WPARAM)block[index].flags;
+		flags = block[index].flags;
 	} else {
 		flags = (USHORT)block[index].flags;
 	}
--- a/libr/core/windows_heap.c
+++ b/libr/core/windows_heap.c
@@ -45,10 +45,10 @@
 *		Maybe instead of using hardcoded structs we can get the offsets from ntdll.pdb
 */
 
-#define PDI_MODULES 0x01
-#define PDI_HEAPS 0x04
-#define PDI_HEAP_TAGS 0x08
-#define PDI_HEAP_BLOCKS 0x10
+#define PDI_MODULES         0x01
+#define PDI_HEAPS           0x04
+#define PDI_HEAP_TAGS       0x08
+#define PDI_HEAP_BLOCKS     0x10
 #define PDI_HEAP_ENTRIES_EX 0x200
 
 #define CHECK_INFO(heapInfo)\
--- a/libr/core/windows_heap.c
+++ b/libr/core/windows_heap.c
@@ -760,7 +760,7 @@ static PHeapBlock GetSingleSegmentBlock(RDebug *dbg, HANDLE h_proc, PSEGMENT_HEA
 	}
 	PHeapBlockExtraInfo extra = calloc (1, sizeof (HeapBlockExtraInfo));
 	if (!extra) {
-		R_LOG_ERROR ("GetSingleBlock: Allocation failed.\n");
+		R_LOG_ERROR ("GetSingleSegmentBlock: Allocation failed.\n");
 		goto err;
 	}
 	hb->extraInfo = extra;
--- a/libr/core/windows_heap.c
+++ b/libr/core/windows_heap.c
@@ -205,7 +205,12 @@ static bool GetNextHeapBlock(PDEBUG_HEAP_INFORMATION heapInfo, PHeapBlock hb) {
 		hb->index++;
 	}
 
-	WPARAM flags = block[index].flags;
+	WPARAM flags;
+	if (block[index].extra & EXTRA_FLAG) {
+		flags = (WPARAM)block[index].flags;
+	} else {
+		flags = (USHORT)block[index].flags;
+	}
 	UPDATE_FLAGS (hb, flags);
 
 	return true;
--- a/libr/core/windows_heap.c
+++ b/libr/core/windows_heap.c
@@ -276,7 +276,7 @@ static bool DecodeLFHEntry(PHEAP heap, PHEAP_ENTRY entry, PHEAP_USERDATA_HEADER 
 	}
 
 	if (heap->EncodeFlagMask) {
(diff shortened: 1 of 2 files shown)


libr/core/windows_heap.c
```diff
@@ -191,26 +226,62 @@ static void free_extra_info(PDEBUG_HEAP_INFORMATION heap) {
...
+
+static bool DecodeHeapEntry(PHEAP heap, PHEAP_ENTRY entry, int bitness) {
+	if (bitness == R_SYS_BITS_64) {
+		(WPARAM) entry += sizeof (PVOID);
+	}
 	if (heap->EncodeFlagMask && (*(UINT32 *)entry & heap->EncodeFlagMask)) {
-#if defined(_M_X64)
-		(WPARAM)heap += sizeof (PVOID);
-#endif
+		if (bitness == R_SYS_BITS_64) {
```
XVilka: While you are on it - imagine the scenario where radare2 is built for the Linux i386 machine, or ARMv7, while connecting to the Windows 10 64bit through `windbg://` remote debugging protocol. Thus the heap decoding should not depend on the host (radare2) parameters, but rather the parameters set with `asm.arch`, `asm.bits`, etc. See the example of our TCC-based C types definitions parsing code, where the size of `long` depends on the architecture and OS: https://github.com/radare/radare2/blob/master/shlr/tcc/tccgen.c#L515

libr/core/windows_heap.c
```diff
@@ -60,14 +62,42 @@
...
+	} else if ((flags) & 0x20) {\
 		hb->dwFlags = LF32_MOVEABLE;\
-	} else if ((flags & 0x0100)) {\
+	} else if ((flags) & 0x0100) {\
 		hb->dwFlags = LF32_FREE;\
 	}\
-	hb->dwFlags |= (flags >> SHIFT) << SHIFT;\
+	hb->dwFlags |= ((flags) >> SHIFT) << SHIFT;
+
+#define FILL_TYPE(type, flags)\
```
radare: Make this into a function. Macros shouldnt be that large if u want to make code properly maintainable

libr/core/windows_heap.c
```diff
@@ -60,14 +62,42 @@
...
+	memset (type, 0, sizeof (type));\
+	switch (flags & 0xFFFF) {\
+	case LF32_FIXED:\
+		strncpy (type, "(FIXED)", 8);\
```
radare: Just return that const string instead kf strcpying all the time. Also strncpy is bad idea theres no sense to have the risk to miss the length and not copy the trailing 0

libr/core/windows_heap.c
```diff
@@ -60,14 +62,42 @@
...
+		break;\
+	case LF32_FREE:\
+		strncpy (type, "(FREE)", 7);\
+		break;\
+	case LF32_MOVEABLE:\
+		strncpy (type, "(MOVEABLE)", 11);\
+		break;\
+	}\
+	if (flags & SEGMENT_HEAP_BLOCK) {\
+		strncat (type, "Segment", 8);\
```
radare: Strncat is a pretty bad call and shoukdnt be used. Also its used incorrectly. Better use rstr apis which are safer and simpler, like strnewf and strappendf..

libr/core/windows_heap.c
```diff
@@ -191,26 +226,62 @@ static void free_extra_info(PDEBUG_HEAP_INFORMATION heap) {
 	}
 }
 
-static bool DecodeHeapEntry(PHEAP heap, PHEAP_ENTRY entry) {
-#if defined(_M_X64)
-	(WPARAM)entry += sizeof (PVOID);
-#endif
+static WPARAM GetNtDllOffset(RDebug *dbg) {
+	RList *map_list = r_w32_dbg_maps (dbg);
```
radare: Can this list be null? Maybe better to rreturn to assert this

libr/core/windows_heap.c
```diff
@@ -175,7 +205,12 @@ static bool GetNextHeapBlock(PDEBUG_HEAP_INFORMATION heapInfo, PHeapBlock hb) {
 		hb->index++;
 	}
 
-	WPARAM flags = block[index].flags;
+	WPARAM flags;
+	if (block[index].extra & EXTRA_FLAG) {
```
radare: Unf? Depending in extra flag you cast differently withiut any binary mask? I dont get why that

libr/core/windows_heap.c
radare: This code only runs in target machine. So the remote debugging situation wont be possible to happen

libr/core/windows_heap.c
```diff
@@ -191,26 +226,62 @@ static void free_extra_info(PDEBUG_HEAP_INFORMATION heap) {
...
+	if (bitness == R_SYS_BITS_64) {
+		(WPARAM) entry += sizeof (PVOID);
+	}
 	if (heap->EncodeFlagMask && (*(UINT32 *)entry & heap->EncodeFlagMask)) {
-#if defined(_M_X64)
-		(WPARAM)heap += sizeof (PVOID);
-#endif
+		if (bitness == R_SYS_BITS_64) {
+			(WPARAM) heap += sizeof (PVOID);
+		}
```
radare: But i would rather use dbg->bits here baceuse thats what matters here. The debugger target bitness, not the host or compiler ones

libr/core/windows_heap.c
```diff
@@ -689,6 +746,174 @@ static PDEBUG_BUFFER GetHeapBlocks(DWORD pid, RDebug *dbg) {
...
 }
 
+static PHeapBlock GetSingleSegmentBlock(RDebug *dbg, HANDLE h_proc, PSEGMENT_HEAP heapBase, WPARAM offset) {
+	/*
+	*	TODO:
+	*		- LFH
+	*		- VS
+	*		- Backend
+	*/
+	PHeapBlock hb = calloc (1, sizeof (HeapBlock));
```
radare: Use R_NEW0

libr/core/windows_heap.c
XVilka: @radare the idea is to be able to parse heap remotely as well, or even from MDMP files. So why not to lay the ground right now?

radare:
Yes i know the same should be done for jemalloc and glibc and this code is heavily tied to windows apis. So this bitness check is probably the least of my concerns

libr/core/windows_heap.c
GustavoLCR: If `map_list` is null `r_list_foreach` won't do anything and the function will return 0, so it shouldn't be necessary.

libr/core/windows_heap.c
GustavoLCR: It's because `InitHeapInfo` compared to `GetHeapBlocks` may fill the flags field in a different way that I don't understand and interferes with my flag defines, so I just truncate the useless stuff. I could also do `& 0xFFFF` if you prefer.

Issue #14268, opened 2019-06-10
Implement ob=


radare:
Like o= or om= or iS=

XVilka:
@deepakchethan please also document it in the book: https://github.com/radare/radare2book/blob/master/basic_commands/mapping_files.md

deepakchethan:
> @deepakchethan please also document it in the book: https://github.com/radare/radare2book/blob/master/basic_commands/mapping_files.md

Will add @XVilka

Maijin:
And the self-documentation in `ob?`

deepakchethan:
> And the self-documentation in `ob?`

Done, made a PR just now. I think this can be closed now.

PR #14271, opened 2019-06-10
For #14268: Implement ob=

deepakchethan:
Output:

Screen Shot 2019-06-10 at 3 37 29 PM

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -379,23 +379,18 @@ static void cmd_open_bin(RCore *core, const char *input) {
 		{
 			RListIter *iter;
 			RList *list = r_list_new ();
-			RBinFile *binfile = NULL;
+			RBinFile *bf = NULL;
 			RBin *bin = core->bin;
-			const RList *binfiles = bin ? bin->binfiles: NULL;
-			if (!binfiles) {
+			if (!bin) {
 				return;
 			}
-			r_list_foreach (binfiles, iter, binfile) {
+			r_list_foreach (bin->binfiles, iter, bf) {
 				char temp[4];
-				ListInfo *info = R_NEW (ListInfo);
+				RInterval inter = (RInterval) {bf->o->baddr, bf->o->size};
+				RListInfo *info = r_listinfo_new (bf->file, inter, inter, -1,  sdb_itoa (bf->fd, temp, 10));
 				if (!info) {
-					return;
+					break;
 				}
-				info->name = binfile->file;
-				info->pitv = (RInterval) {binfile->o->baddr, binfile->o->size};
-				info->vitv = info->pitv;
-				info->perm = -1;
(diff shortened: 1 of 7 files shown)


libr/core/cmd_open.c
```diff
@@ -375,6 +375,33 @@ static void cmd_open_bin(RCore *core, const char *input) {
 			}
 		}
 		break;
+	case '=': // "ob="
+		{
+			RListIter *iter;
+			RList *list = r_list_new ();
+			RBinFile *binfile = NULL;
```
radare: Use bf instead of binfile. its shorter and im doing this change in all the codebase

libr/core/cmd_open.c
```diff
@@ -375,6 +375,33 @@ static void cmd_open_bin(RCore *core, const char *input) {
...
+			RBin *bin = core->bin;
+			const RList *binfiles = bin ? bin->binfiles: NULL;
```
radare: unnecessary localvar

libr/core/cmd_open.c
```diff
@@ -375,6 +375,33 @@ static void cmd_open_bin(RCore *core, const char *input) {
...
+			if (!binfiles) {
```
radare: should never be null, so r_lsit_foreach will assert, unnecessary if

libr/core/cmd_open.c
```diff
@@ -375,6 +375,33 @@ static void cmd_open_bin(RCore *core, const char *input) {
...
+				return;
+			}
+			r_list_foreach (binfiles, iter, binfile) {
+				char temp[4];
+				ListInfo *info = R_NEW (ListInfo);
+				if (!info) {
+					return;
```
radare: break instead of return

libr/core/cmd_open.c
```diff
@@ -375,6 +375,33 @@ static void cmd_open_bin(RCore *core, const char *input) {
 			}
 		}
 		break;
+	case '=': // "ob="
+		{
+			RListIter *iter;
+			RList *list = r_list_new ();
```
radare: r_list_newf(free), but should be (r_listinfo_free),

libr/core/cmd_open.c
```diff
@@ -375,6 +375,33 @@ static void cmd_open_bin(RCore *core, const char *input) {
...
+			const RList *binfiles = bin ? bin->binfiles: NULL;
+			if (!binfiles) {
+				return;
+			}
+			r_list_foreach (binfiles, iter, binfile) {
+				char temp[4];
+				ListInfo *info = R_NEW (ListInfo);
+				if (!info) {
+					return;
+				}
```
radare: maybe good to have an r_listinfo_new() to just pass 4 parameters instead of doing the new+struct init by hand

deepakchethan: Okay :)

PR #14288, opened 2019-06-12
Something like Ctrl w + T in vim implemented ##panels

Vane11ope:
and Fixed a few bugs

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -506,6 +506,7 @@ static bool handle_tab_next(RCore *core);
 static bool handle_tab_prev(RCore *core);
 static void handle_tab_name(RCore *core);
 static void handle_tab_new(RCore *core);
+static void handle_tab_new_with_cur_panel(RCore *core);
 static bool handle_tab_del(RCore *core);
 static void del_panels(RCore *core);
 
@@ -4915,6 +4916,10 @@ static bool handle_tab(RCore *core) {
 		break;
 	case 't':
 		handle_tab_new (core);
+		break;
+	case 'T':
+		handle_tab_new_with_cur_panel (core);
+		return true;
 	}
 	return false;
 }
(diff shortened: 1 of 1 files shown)


libr/core/panels.c
```diff
@@ -4889,7 +4889,7 @@ static RPanels *get_cur_panels(RPanelsRoot *panels_root) {
 	return get_panels (panels_root, panels_root->cur_panels);
 }
 
-static bool handle_tab(RCore *core) {
+static bool handle_tab(RCore *core, bool *del_tab) {
```
radare: Not a big fan of this new signature. It doesnt escales and its confusing to use. Im fine for merging i guess you’ll be diing more refactors over time

libr/core/panels.c
```diff
@@ -4970,6 +4970,38 @@ static void handle_tab_new(RCore *core) {
...
+
+	RPanelsRoot *root = core->panels_root;
+	if (root->n_panels >= PANEL_NUM_LIMIT) {
+		return;
+	}
+
+	RPanel *cur = getCurPanel (panels);
+
+	root->n_panels++;
+	root->panels[root->n_panels - 1] = panels_new (core);
```
radare: Chk for nul

libr/core/panels.c
```diff
@@ -5374,8 +5406,15 @@ static int panels_process(RCore *core, RPanels **r_panels, bool *force_quit) {
 		}
 		break;
 	case 't':
-		if (handle_tab (core)) {
-			return false;
+		{
+			bool del_tab = false;
+			if (handle_tab (core, &del_tab)) {
+				if (del_tab) {
+					goto exit;
```
radare: ELse after goto. No pls

libr/core/panels.c
Vane11ope: yup exactly. i will be refactoring this later on but currently '-' keys in tab emits a bug anyway without this.

libr/core/panels.c
Vane11ope: ah ok

libr/core/panels.c
Vane11ope: fair enough. a few more points just occurred to me to fix in fact :)

Vane11ope:
this indeed contains bunch of buggy codes, let me quickly fix them.

Vane11ope:
thought enough for now but hang on yet

Vane11ope:
yeah in fact didn't come up with cleaner solution yet, will do refactoring soon enough.

Issue #14254, opened 2019-06-09
Dark text color with default theme (recent regression)


astuder:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Debian x86 64
| File format of the file you reverse (mandatory)      | bin
| Architecture/bits of the file (mandatory)            | 8051
| r2 -v full output, **not truncated** (mandatory)     | rradare2 3.6.0-git 21830 @ linux-x86-64 git.1.5.0-6974-g6fce40ea0 commit: 6fce40ea04ac1c07937cba58b4fdadf3750df66b build: 2019-06-08__18:59:39

scr.color=3
eco default

### Expected behavior

The colors of the default theme are suddenly very dark / low contrast. They were much brighter in older builds. My point of reference is May 26 2019:

![r2 PR14056](https://user-images.githubusercontent.com/1166488/59154140-38eaa880-8a20-11e9-820f-43654c0d1219.PNG)

### Actual behavior

This is how it looks after pulling the latest version from git

![r2 HEAD](https://user-images.githubusercontent.com/1166488/59154141-3f792000-8a20-11e9-9737-2267f29af534.PNG)

radare:
That’s because we are living dark times and everything is so depressive

PR #14289, opened 2019-06-12
Use a less dark color scheme by default on scr.color=3 on non-Windows

kazarmy:
Admittedly, on Linux the default color scheme on scr.color=3 is somewhat dark (see #14254), so this is an attempt to lighten things up a bit.


kazarmy:
Campbell color scheme on GitHub: https://github.com/microsoft/terminal/blob/1580c1e0934f7b04dcd420df051f8604df04c670/src/tools/ColorTool/schemes/campbell.ini (MIT-licensed)

PR #14297, opened 2019-06-13
WIP: Really basic implementation of esil.timeout.

hexploitable:
Hey @trufae,

Here's a really dumb/super basic hack for the problem we discussed with aesu/aecu etc running for too much time. Not sure if you want to do something more intelligent but it works. 

Essentially, we offer the user to configure `esil.timeout` in the variables and if this value is >0 then we start a timer and each time we loop we check if the timeout has passed.

Screenshot 2019-06-13 at 20 16 12

This isn't perfect because we could be in the middle of evaluating an expression as the timeout passed. Ideally it would be handled in a thread but this gets more complex.

Is this alright, i noticed you do timeouts differently here:
https://github.com/radare/radare2/blob/master/libr/cons/cons.c#L385-L395

Do we want to use `r_sys_now()` instead of `clock()`?

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -3828,8 +3828,7 @@ R_API int r_core_esil_step(RCore *core, ut64 until_addr, const char *until_expr,
 	if (!esil) {
 		initializeEsil (core);
 	}
-	if (esiltimeout)
-	{
+	if (esiltimeout) {
 		startTime = clock();
 	}
 	ut64 addr = r_reg_getv (core->anal->reg, name);
@@ -3840,12 +3839,10 @@ repeat:
 		return_tail (0);
 	}
 	//Break if we have exceeded esil.timeout
-	if (esiltimeout)
-	{
+	if (esiltimeout) {
 		endTime = clock();
 		double elapsedTime = ((double)endTime-startTime) / CLOCKS_PER_SEC;
-		if (elapsedTime >= esiltimeout)
-		{
+		if (elapsedTime >= esiltimeout) {
 			eprintf ("[ESIL] Timeout exceeded.\n");
 			return_tail (0);
 		}
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -3841,7 +3841,8 @@ repeat:
 	//Break if we have exceeded esil.timeout
 	if (esiltimeout) {
 		ut64 elapsedTime = r_sys_now () - startTime;
-		if (elapsedTime >= (esiltimeout * 1000000)) {
+		elapsedTime >>= 20;
+		if (elapsedTime >= esiltimeout) {
 			eprintf ("[ESIL] Timeout exceeded.\n");
 			return_tail (0);
 		}
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -1206,7 +1206,7 @@ static int var_cmd(RCore *core, const char *str) {
 		r_anal_var_list_show (core->anal, fcn, type, str[1], pj);
 		r_cons_println (pj_string (pj));
 		pj_free (pj);
-	} 
+	}
 		break;
 	case '.':
 		r_anal_var_list_show (core->anal, fcn, core->offset, 0, NULL);
@@ -3403,7 +3403,7 @@ static void __anal_reg_list(RCore *core, int type, int bits, char mode) {
 // XXX dup from drp :OOO
 void cmd_anal_reg(RCore *core, const char *str) {
 	if (0) {
-		/* enable this block when dr and ar use the same code but just using 
+		/* enable this block when dr and ar use the same code but just using
 		   core->dbg->reg or core->anal->reg and removing all the debugger
 		   dependent code */
 		RReg *reg = core->dbg->reg;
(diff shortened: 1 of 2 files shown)


libr/core/cmd_anal.c
```diff
@@ -3823,16 +3823,33 @@ R_API int r_core_esil_step(RCore *core, ut64 until_addr, const char *until_expr,
...
 	const char *name = r_reg_get_name (core->anal->reg, R_REG_NAME_PC);
 	bool breakoninvalid = r_config_get_i (core->config, "esil.breakoninvalid");
+	int esiltimeout = r_config_get_i (core->config, "esil.timeout");
+	clock_t startTime, endTime;
 	if (!esil) {
 		initializeEsil (core);
 	}
+	if (esiltimeout)
+	{
+		startTime = clock();
```
hexploitable: I'll convert this to `rsysnow()`

libr/core/cmd_anal.c
```diff
@@ -3823,16 +3823,33 @@ R_API int r_core_esil_step(RCore *core, ut64 until_addr, const char *until_expr,
 	RAnalEsil *esil = core->anal->esil;
 	const char *name = r_reg_get_name (core->anal->reg, R_REG_NAME_PC);
 	bool breakoninvalid = r_config_get_i (core->config, "esil.breakoninvalid");
+	int esiltimeout = r_config_get_i (core->config, "esil.timeout");
+	clock_t startTime, endTime;
 	if (!esil) {
 		initializeEsil (core);
 	}
+	if (esiltimeout)
+	{
```
hexploitable: will migrate this to the r2 style guidelines.

libr/core/cmd_anal.c
```diff
@@ -3823,16 +3823,33 @@ R_API int r_core_esil_step(RCore *core, ut64 until_addr, const char *until_expr,
...
 	ut64 addr = r_reg_getv (core->anal->reg, name);
 	r_cons_break_push (NULL, NULL);
 repeat:
 	if (r_cons_is_breaked ()) {
 		eprintf ("[+] ESIL emulation interrupted at 0x%08" PFMT64x "\n", addr);
 		return_tail (0);
 	}
+	//Break if we have exceeded esil.timeout
+	if (esiltimeout)
+	{
```
hexploitable: will migrate to r2 style guidelines

libr/core/cmd_anal.c
```diff
@@ -3823,16 +3823,33 @@ R_API int r_core_esil_step(RCore *core, ut64 until_addr, const char *until_expr,
...
+		endTime = clock();
+		double elapsedTime = ((double)endTime-startTime) / CLOCKS_PER_SEC;
+		if (elapsedTime >= esiltimeout)
+		{
```
hexploitable: migrate to r2 style guidelines

hexploitable:
Should be ready.

libr/core/cmd_anal.c
```diff
@@ -3823,16 +3823,29 @@ R_API int r_core_esil_step(RCore *core, ut64 until_addr, const char *until_expr,
 	RAnalEsil *esil = core->anal->esil;
 	const char *name = r_reg_get_name (core->anal->reg, R_REG_NAME_PC);
 	bool breakoninvalid = r_config_get_i (core->config, "esil.breakoninvalid");
+	int esiltimeout = r_config_get_i (core->config, "esil.timeout");
+	ut64 startTime;
 	if (!esil) {
 		initializeEsil (core);
 	}
+	if (esiltimeout) {
+		startTime = r_sys_now();
```
radare: missing space before (

libr/core/cmd_anal.c
```diff
@@ -3823,16 +3823,29 @@ R_API int r_core_esil_step(RCore *core, ut64 until_addr, const char *until_expr,
...
 	r_cons_break_push (NULL, NULL);
 repeat:
 	if (r_cons_is_breaked ()) {
 		eprintf ("[+] ESIL emulation interrupted at 0x%08" PFMT64x "\n", addr);
 		return_tail (0);
 	}
+	//Break if we have exceeded esil.timeout
+	if (esiltimeout) {
+		ut64 elapsedTime = r_sys_now () - startTime;
+		if (elapsedTime >= (esiltimeout * CLOCKS_PER_SEC)) {
```
radare: dont use this CLOCKS_PER_SEC constant. in fact r_sys_now() do a <<= 20 for the seconds, so you can shift that value back to drop the lower timing info. We probably want to have a macro or a function for this instead of depending on the posix stuff which is unrelated here because r_sys_now() uses gettimeofday()

libr/core/cmd_anal.c
hexploitable: Fixed.

radare:
@condret what do you think about this PR? maybe having this startTime inside the esil vm will make it work on all the possible use cases of esil emulation? so we add RAnalEsil.start()/stop() ?

libr/core/cmd_anal.c
hexploitable: Fixed the use of the `CLOCKS_PER_SEC` const, now just shifting back. Didn't spot that, thanks.

PR #14305, opened 2019-06-15
Faster dtd

bart1e:
Printing `dtd` output faster, but not in color and without opcodes (just address + mnemonic). May be worth considering to enable that only when printing to a file (assuming that, when the output is huge, user will redirect it) and leave current behaviour while not redirecting, so it will be slow, but will not be used with huge outputs.

### DIFF
--- a/libr/core/cmd_debug.c
+++ b/libr/core/cmd_debug.c
@@ -4571,8 +4571,8 @@ static int cmd_debug(void *data, const char *input) {
 				//r_core_cmd0 (core, "pd 1 @@= `dtq`");
 				r_list_foreach (core->dbg->trace->traces, iter, trace) {
 					op = r_core_anal_op (core, trace->addr, R_ANAL_OP_MASK_BASIC | R_ANAL_OP_MASK_DISASM);
-						r_cons_printf ("0x%08"PFMT64x" %s\n", trace->addr, op->mnemonic);
-					}
+					r_cons_printf ("0x%08"PFMT64x" %s\n", trace->addr, op->mnemonic);
+				}
 			}
 			break;
 		case 'g': // "dtg"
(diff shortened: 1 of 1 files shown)


libr/core/cmd_debug.c
```diff
@@ -4547,31 +4549,30 @@ static int cmd_debug(void *data, const char *input) {
...
 			} else if (input[2] == ' ') {
-				int min = r_num_math (core->num, input + 3);
-				RListIter *iter;
-				RDebugTracepoint *trace;
 				int n = 0;
 				r_list_foreach (core->dbg->trace->traces, iter, trace) {
+				op = r_core_anal_op (core, trace->addr, R_ANAL_OP_MASK_BASIC | R_ANAL_OP_MASK_DISASM);
 					if (n >= min) {
-						r_core_cmdf (core, "pd 1 @ 0x%08"PFMT64x, trace->addr);
+						r_cons_printf ("%llx %s\n", trace->addr, op->mnemonic);
```
radare: use PFMT64x, %llx is not portable

radare: and hex numbers start with 0x

radare: we need to refactor the disasm.c code to make it easier to use from other commands instead of having to call "pd" like this. for now thats fine i think in this case

libr/core/cmd_debug.c
```diff
@@ -4547,31 +4549,30 @@ static int cmd_debug(void *data, const char *input) {
...
 					}
 					n++;
 				}
 			} else {
 				// TODO: reimplement using the api
-				r_core_cmd0 (core, "pd 1 @@= `dtq`");
+				//r_core_cmd0 (core, "pd 1 @@= `dtq`");
+				r_list_foreach (core->dbg->trace->traces, iter, trace) {
+				op = r_core_anal_op (core, trace->addr, R_ANAL_OP_MASK_BASIC | R_ANAL_OP_MASK_DISASM);
```
radare: wrong indentation

libr/core/cmd_debug.c
```diff
@@ -4547,31 +4549,30 @@ static int cmd_debug(void *data, const char *input) {
...
+					r_cons_printf ("%llx %s\n", trace->addr, op->mnemonic);
```
radare: use PFMT64x and prefix with 0x

PR #14310, opened 2019-06-15
Open disas panel if it is not there when the cursor selects some offset

Vane11ope:


libr/core/panels.c
```diff
@@ -286,7 +286,7 @@ static void set_root_state(RCore *core, RPanelsRootState state);
 static void resetScrollPos(RPanel *p);
 
 /* update */
-static void updateDisassemblyAddr (RCore *core);
+static void update_disassembly_or_open (RCore *core);
```
radare: according to the devrules, static(private) functions should start with double lower dash

libr/core/panels.c
```diff
@@ -2520,15 +2521,37 @@ static int ioCacheOffCb(void *user) {
...
+		int h0 = panel->view->pos.h;
+		int threshold_w = x0 + panel->view->pos.w;
+		int x1 = x0 + w0 / 2 - 1;
+		int w1 = threshold_w - x1;
+
+		insertPanel (core, 0, PANEL_TITLE_DISASSEMBLY, PANEL_CMD_DISASSEMBLY, false);
+		RPanel *p0 = getPanel (panels, 0);
+		set_geometry (&p0->view->pos, x0, y0, w0 / 2, h0);
+
+		RPanel *p1 = getPanel (panels, 1);
```
radare: what if there's only one panel? wouldnt this be NULL?

libr/core/panels.c
Vane11ope: you should've told me this at day one but ok. and also, wanna delete space after the name. hang on

libr/core/panels.c
Vane11ope: never. a panel is inserted in a few lines ago as you can see, and panels have at least a panel in it always.

Vane11ope:
but thank you for review anyway, would like to fix a few things :)

libr/core/panels.c
Vane11ope: way too many functions ignoring this rule in the whole code base. do you unify them into this rule after all?

radare:
np, theres lot of code in r2 that is like this, also mixed camel case and such, not blaming :)just something to have in mind when we refactor code. ive been changing this stuff every day for the last month or so if you see the commits for ex

Vane11ope:
ah ok cool i will then unify my code base into this rule at least, shouldn't be difficult and the earlier the better anyway.

PR #14311, opened 2019-06-15
Restyling

Vane11ope:


libr/core/panels.c
```diff
@@ -681,7 +681,7 @@ static void panelPrint(RCore *core, RConsCanvas *can, RPanel *panel, int color)
 	}
 }
 
-static void menuPanelPrint(RConsCanvas *can, RPanel *panel, int x, int y, int w, int h) {
+void __menuPanelPrint(RConsCanvas *can, RPanel *panel, int x, int y, int w, int h) {
```
radare: uh? why drop static?.. actually this `__` rule seems a bit verbose in this use case . maybe we shouldnt go this way, or define a conditional rule when 3rd party apis (non-r2 ones) are also used  to make clear its a local function when reading the code.

Vane11ope:
it is static in prototypes and once you declare the function as static in prototypes then you dont need to write static in actual implementation and i wanna avoid extra typing if i could.

Vane11ope:
and its you who told me to refactor like this even a commit ago and that's why i did this. if you didn't mean it then you shouldve even not started saying this AT ALL. plz, make your mind up about this first.

radare:
🙄

Issue #10613, opened 2018-07-04
Windows failing tests


XVilka:
```
[**]                db\tools\rabin2   OK   23 BR    2 XX    0 FX    0
[**]                    db\cmd\0000   OK   23 BR    3 XX    0 FX    0
[**]                   db\cmd\archs   OK   30 BR    3 XX    0 FX    0
[**]            db\cmd\backtickgrep   OK   31 BR    3 XX    0 FX    0
[**]                   db\cmd\basic   OK   36 BR    3 XX    0 FX    0
[XX] db\cmd\bitmask bitmask cmp data  31
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -i C:\Users\appveyor\AppData\Local\Temp\1\tmp-2064dVS1HnlPZwZZ.tmp -
b 16
wx 414243
e scr.null=1
cx 41..43 2>/dev/null
e scr.null=0
?! ?e thats fine
e scr.null=1
cx 41..11 2>/dev/null
e scr.null=0
?! ?e thats wrong
 thats fine
-
+thats wrong
EXPECT64=dGhhdHMgZmluZQp0aGF0cyB3cm9uZwo=
[**]                 db\cmd\bitmask   OK   36 BR    3 XX    1 FX    0
[**]                db\cmd\bug_3788   OK   40 BR    3 XX    1 FX    0
[**]            db\cmd\bug_backtick   OK   41 BR    3 XX    1 FX    0
[**]                    db\cmd\cmds   OK   52 BR    4 XX    1 FX    0
[**]                   db\cmd\cmd_0   OK  124 BR    4 XX    1 FX    0
[**]                  db\cmd\cmd_af   OK  124 BR    5 XX    1 FX    0
[**]                  db\cmd\cmd_ah   OK  127 BR    5 XX    1 FX    0
[**]                 db\cmd\cmd_ahi   OK  149 BR    5 XX    1 FX    0
[**]               db\cmd\cmd_alias   OK  154 BR    5 XX    1 FX    0
[**]                 db\cmd\cmd_ara   OK  158 BR    5 XX    1 FX    0
[**]                 db\cmd\cmd_att   OK  162 BR    5 XX    1 FX    0
[**]                  db\cmd\cmd_av   OK  163 BR    5 XX    1 FX    0
[**]                  db\cmd\cmd_Ch   OK  164 BR    5 XX    1 FX    0
[...]

PR #14313, opened 2019-06-16
Trim before comparing pseudo asm

GustavoLCR:
#10613


radare:
I think that r_str_trim should return void. Can you do this change?

radare:
tired of waiting. ill do this change

PR #14315, opened 2019-06-16
Make r2 prefix absolute on Windows

GustavoLCR:
#10613


radare:
Why? Which test is fixing this? I think this can affect cutter

radare:
again, this may affect Cutter. @ITAYC0HEN @xarkes @thestr4ng3r can you review this?

GustavoLCR:
yep, this doesn't work very well. Better get the absolute path dynamically. Will try again

PR #14323, opened 2019-06-16
Dinamically find radare2 install dir on Windows

GustavoLCR:
#10613
better #14315

Fixes types and etc not being loaded if radare2 is spawned outside of install dir

### DIFF
--- a/libr/util/sys.c
+++ b/libr/util/sys.c
@@ -1143,7 +1143,21 @@ R_API bool r_sys_tts(const char *txt, bool bg) {
 R_API const char *r_sys_prefix(const char *pfx) {
 	static char prefix[1024] = {0};
 	if (!*prefix) {
+#if __WINDOWS__
+		char *path = r_sys_cmd_str ("where radare2", NULL, NULL);
+		if (path && *path) {
+			char *dir, *tmp = dir = r_file_dirname (path);
+			dir = r_file_dirname (tmp);
+			if (dir) {
+				r_str_ncpy (prefix, dir, sizeof (prefix));
+			}
+			free (dir);
+			free (tmp);
+		}
+		free (path);
+#else
 		r_str_ncpy (prefix, R2_PREFIX, sizeof (prefix));
+#endif
 	}
 	if (pfx) {
 		if (strlen (pfx) >= sizeof (prefix) - 1) {
--- a/libr/util/sys.c
+++ b/libr/util/sys.c
@@ -1141,7 +1141,7 @@ R_API bool r_sys_tts(const char *txt, bool bg) {
 }
 
 R_API const char *r_sys_prefix(const char *pfx) {
-	static char prefix = NULL;
+	static char *prefix = NULL;
 	if (!prefix) {
 #if __WINDOWS__ && !CUTTER
 		char *path = r_sys_cmd_str ("where radare2", NULL, NULL);
--- a/libr/util/sys.c
+++ b/libr/util/sys.c
@@ -1141,29 +1141,30 @@ R_API bool r_sys_tts(const char *txt, bool bg) {
 }
 
 R_API const char *r_sys_prefix(const char *pfx) {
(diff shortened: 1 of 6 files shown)


libr/util/sys.c
```diff
@@ -1143,7 +1143,21 @@ R_API bool r_sys_tts(const char *txt, bool bg) {
 R_API const char *r_sys_prefix(const char *pfx) {
 	static char prefix[1024] = {0};
 	if (!*prefix) {
+#if __WINDOWS__ && !CUTTER
+		char *path = r_sys_cmd_str ("where radare2", NULL, NULL);
+		if (path && *path) {
+			char *dir, *tmp = dir = r_file_dirname (path);
+			dir = r_file_dirname (tmp);
+			if (dir) {
+				r_str_ncpy (prefix, dir, sizeof (prefix));
```
radare: use r_str_newf not fixed buffer strings

libr/util/sys.c
```diff
@@ -1143,7 +1143,21 @@ R_API bool r_sys_tts(const char *txt, bool bg) {
 R_API const char *r_sys_prefix(const char *pfx) {
```
radare: we should probably return char * in the heap here. this static char *thing will fail if we use more than one rsysprefix in the same function or so

GustavoLCR:
> maybe do this change in a separate PR. also i dontn like this separation of windows and cutter, i dont see why cutter windows build should be different than the standard windows build

cutter doesn't ship a radare2.exe, so if someone installs just cutter this would fail.

Also I think there is one more thing wrong with this. I just noticed that my install dir has a different structure from the one in appveyor, The exes are inside a bin folder on my system. Is this because of the --shared flag?

libr/util/sys.c
```diff
@@ -1143,7 +1143,21 @@ R_API bool r_sys_tts(const char *txt, bool bg) {
...
 	static char prefix[1024] = {0};
 	if (!*prefix) {
+#if __WINDOWS__ && !CUTTER
+		char *path = r_sys_cmd_str ("where radare2", NULL, NULL);
```
xarkes: Does the `where` command works from r2 console?
Does it return `NULL` if r2 was not in `PATH` or an empty string?

libr/util/sys.c
```diff
@@ -1143,7 +1143,21 @@ R_API bool r_sys_tts(const char *txt, bool bg) {
...
+		if (path && *path) {
+			char *dir, *tmp = dir = r_file_dirname (path);
+			dir = r_file_dirname (tmp);
+			if (dir) {
+				r_str_ncpy (prefix, dir, sizeof (prefix));
+			}
+			free (dir);
+			free (tmp);
+		}
```
xarkes: What happens if `!path`?
On Windows, since (afaik) most of the people just copy paste the binaries, r2 is most of the time not in the `PATH`.

libr/util/sys.c
GustavoLCR: yes, its built in, if r2 is not in path, the return is an empty string

libr/util/sys.c
GustavoLCR: with the installer script radare is added to path, and if radare is not in path, I guess it should fall back to searching on cwd

GustavoLCR:
I will check the broken tests

libr/util/sys.c
GustavoLCR: I'm now using `r_sys_get_src_dir_w32()` that gets the path from the module

GustavoLCR:
OK, It's working now, but Appveyor is broken on master.

Also because of 977e8ef, I think the build script for Cutter and the Windows installer will have to add `\bin` to the PATH as well like 1b52bbf

GustavoLCR:
> so is this whahts breaking the build?

@radare no, it isn't even merged on master, and it is master that's broken. I have no idea why. I cannot reproduce

PR #14345, opened 2019-06-18
Settings menu is added, and decompiler is configurable in there ##panels

Vane11ope:
works like the attached image, yet choosing "r2dec" emits a bug since

e cmd.pdc=r2dec

does not work internally yet. otherwise the menu works.

pluraldecompiler

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -53,12 +53,17 @@ static const char *panels_static [] = {
 };
 
 static const char *menus[] = {
-	"File", "Edit", "View", "Tools", "Search", "Debug", "Analyze", "Fun", "About", "Help",
+	"File", "Settings", "Edit", "View", "Tools", "Search", "Debug", "Analyze", "Fun", "About", "Help",
 	NULL
 };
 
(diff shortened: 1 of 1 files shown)


radare:
this r2dec line comes rom the r2dec plugin, i just send a PR to fix it

radare:
I think in this setttings panel you can add toggle asm.bytes, and will be good to have a list of other bool setings that are common . but right now i can think about those:

* asm.bytes
* asm.section/hex.section
* asm.cmt.right
* io.cache
* hex.pairs
* emu.str
* asm.emu
* asm.var.summary
* asm.pseudo
* asm.flags.inbytess


and group them or justs put them inside Disasm ->

Vane11ope:
yup that's my plan. also you wanna check Colors settings which is already there that works cool i think :)

Vane11ope:
im gonna add all those config as i personally always hated to go all the way back to V and e then set the config process at the first place :P

radare:
maybe add a submenu for color theme selection

PR #14347, opened 2019-06-18
Rename r_cons_get_ansicon() since it's confusing

kazarmy:


radare:
I think this function should return bool and be renamed to rcons_is_ansicon() or rcons_is_ansi()

kazarmy:
Ok 20e79e336058df03d9bfc86cd8b8d6b0e9a54d28

libr/cons/cons.c
```diff
@@ -1334,7 +1334,7 @@ R_API int r_cons_get_size(int *rows) {
 
 #if __WINDOWS__
 R_API os_info *r_sys_get_osinfo();
-R_API int r_cons_get_ansicon() {
+R_API bool r_cons_is_ansicon() {
```
radare: add void int he fucntion definition

kazarmy: ok 389a2a37dbc73657b6d24619c193fea3063b6c65

PR #14357, opened 2019-06-19
Autocompletion fixes

wargio:
- autocomplete `afn`
- `!!!$eval` says (null)
- autocomplete when there's a space between the eval var and the `=` char.


wargio:
ref: https://github.com/radare/radare2/issues/3033

PR #14377, opened 2019-06-20
Add mingw to PATH in the Appveyor test script

GustavoLCR:
#10613
This fixes the as plugin assembler tests


GustavoLCR:
It didn't

GustavoLCR:
Ok it worked

PR #14378, opened 2019-06-21
Fix spp on Windows

GustavoLCR:
#10613


XVilka:
Please send a PR to https://github.com/radare/spp

radare:
This fix must be in spp. So you need to move this getenv function inside spp. Otherwise we will break things next time we sync

radare:
is your PR exactly the same as mine? :( read it late

GustavoLCR:
Np

Em sex, 21 de jun de 2019 às 15:47, radare <notifications@github.com>
escreveu:

Issue #14293, opened 2019-06-13
db entry0 doesn't break


0ki:
alledegly... 

see: https://github.com/Luis-Hebendanz/0pack

radare:
Any sample binary?

radare:
See https://github.com/Luis-Hebendanz/0pack/issues/1

radare:
@ret2libc as u said this wont be fixed soon please move your pr and the issue to milestone 4.5 or further

PR #14385, opened 2019-06-21
WIP: Fix entrypoint on ELF when its pointing to null ##bin

radare:


radare:
This fixed loading the 0pack

XVilka:
@radare what you decided with this one?

radare:
Its not about deciding. The pr is not good yet

radare:
@ret2libc feedback/code for this?

ITAYC0HEN:
CC @ret2libc

ret2libc:
I would love to see a binary with the issue before working on a fix

radare:
This issue fixes https://github.com/radareorg/radare2/issues/14293 and the bin is there

radare:
@ret2libc https://github.com/Luis-Hebendanz/0pack/releases/download/v1.0/0pack its a 17MB binary  :(( i wish we had a smaller reproducer

XVilka:
@radare we can create a sparse binaries format for the tests - see https://github.com/radareorg/radare2/issues/16184

radare:
Isnt that supported by git? Whats the problem you sre trying to solve? Loading huge files with gaps as sparsed files? Not sure how that will improve this test but we can use gzip:// in here i guess

XVilka:
@radare not quite. I suggest to create a binary instead of series `wx .... @`, `wx ... @`, and so on. Quite often these many huge files can be reduced to just a few pieces, e.g. header and a few blocks in the selected locations.

radare:
Gzip or ihe. Works fine for those cases

XVilka:
Not sure what "ihe" is. GZip will not create sparse files, it can just compress them.

ret2libc:
@radare I think the binary you put in elf/0pack is not the right one, as that has not entry_point==0 in the file header.

radare:
The Good one is 17MB.. though the 7MB version used the same trick..

Luis-Hebendanz:
I re-uploaded the 7Mb 0Pack binary because I forgot to patch the loader path so that it works under non NixOS systems. Now it works I verified it :-)

radare:
Cc @ret2libc

ret2libc:
I will try to give this another try before 4.5.0, thought it requires a bit of refactoring here and there in elf code, which was assuming in general 0 was a "bad value".

radare:
yeah, 0 is not that bad

radare:
pls fix conflicts

ret2libc:
No need, I still need to fix many things here. No time right now.

ret2libc:
Closing as this needs anyway a lot of re-work because many things changed in elf and the current state was not working nonetheless. There is also already an issue https://github.com/radareorg/radare2/issues/14293 for this.

PR #14390, opened 2019-06-21
Upgrade spp from git

radare:


GustavoLCR:
Isn't this on my pr already? #14378

PR #14398, opened 2019-06-22
WIP: Containing RBTree for struct that live just for a short time in an rbtree

condret:


ret2libc:
1. Can you explain what you need this for first? 
2. Why do you need a struct to live in a RBTree just for a short time?
3. Where else will this struct live for the rest of the time?
4. How will you construct the RBTree?
5. How many search operations will you do?
6. Why an RBTree?

libr/util/rbtree.c
```diff
@@ -382,3 +384,117 @@ R_API void r_rbtree_iter_next(RBIter *it) {
...
+	free(container_of(node, RContRBNode, node));
+}
+
+R_API bool r_rbtree_cont_delete(RContRBTree *tree, void *data, RContRBCmp cmp, void *user) {
+	r_return_val_if_fail (tree && cmp && tree->root, false);
+	RCRBCmpWrap cmp_wrap = { cmp, tree->free, user };
+	RBNode *root_node = &tree->root->node;
+	const bool ret = r_rbtree_aug_delete(&root_node, data, cont_rbtree_free_cmp_wrapper, cont_node_free, NULL, &cmp_wrap);
+	if (root_node != (&tree->root->node)) {
+		tree->root = container_of(root_node, RContRBNode, node);	//cursed augmentation garbage
```
radare: misssing spaces before ( in different places

libr/util/rbtree.c
```diff
@@ -382,3 +384,117 @@ R_API void r_rbtree_iter_next(RBIter *it) {
 R_API void r_rbtree_iter_prev(RBIter *it) {
 	_next (it, 1);
 }
+
+
+R_API RContRBTree *r_rbtree_cont_new() {
+	RContRBTree *tree = R_NEW0 (RContRBTree);
```
radare: just return R_NEW0

libr/util/rbtree.c
```diff
@@ -382,3 +384,117 @@ R_API void r_rbtree_iter_next(RBIter *it) {
...
+	const bool ret = r_rbtree_aug_delete(&root_node, data, cont_rbtree_free_cmp_wrapper, cont_node_free, NULL, &cmp_wrap);
+	if (root_node != (&tree->root->node)) {
+		tree->root = container_of(root_node, RContRBNode, node);	//cursed augmentation garbage
+	}
+	if (!ret) {
+		eprintf ("Deletion failed\n");
+	}
+	return ret;
+}
+
```
radare: empty line

ret2libc:
At least write unit tests for this new implementation... I'm not sure how you can say whether it even works :)

radare:
Its just adding a void * user. Not reimplementing anything. But maybe i overlooked

ret2libc:
Yes indeed, he added a lot of `void *user` all around, so now existing RBtree has a new parameter that is always NULL. Why? The new data structure is not even used anywhere. Was it tested in any way? What is `void *user` and `void *data` in the context of existing RBTree?

rlaemmert:
You're about to change the data structure just for one use case? Sounds a little dumb to me...

radare:
I just wonder if we can have void*user inside the rbtree instead of passing it as argument all the time. He is mimicing the same apis as in rlist

ret2libc:
Yes it would be possible to change the rbtree data structure to not be *intrusive*, but making this change makes both data structures (the intrusive one and the non-intrusive one) a bit of a mess. I don't understand why, yet. I think everything can be done with the existing data structure in any case and then IFF it's really necessary we can change that for everything.

radare:
Well this is probably a requirement for the esil graph he is working on and having a void* user associated with esch data structure we have makes sense to me.

ret2libc:
I don't think it's a *requirement*. What can be done with the non-intrusive data structure can be done with the intrusive ones as well. At most, it may be a nice-to-have thing. But
1) There are no tests, so as far as i'm concerned, this data structure may even be completely broken
2) There is no real use case at the moment.  I have asked a bunch of question in previous comment, do we really need a RBTree? Why? How many search does he need to perform? Does he need sorting? Is an hashtable enough? I don't know.... In any case, whatever it is that @condret is doing, it can be done with existing data structures and only then we can really see whether this change is needed.

libr/include/r_util/r_rbtree.h
```diff
@@ -25,7 +25,7 @@ typedef struct r_rb_node_t {
 
 typedef RBNode* RBTree;
 
-typedef int (*RBComparator)(const void *incoming, const RBNode *in_tree);
+typedef int (*RBComparator)(const void *incoming, const RBNode *in_tree, void *user);		//here needs to be a void *user
```
radare: imho the RBComparator should take the RBTree as first arg, and this void* user should be inside the RBTree struct

radare:
ping? is  this pr dead?

condret:
closing, as it already got merged in esil_dfg and esil_cfg

PR #14422, opened 2019-06-25
WIP: Added arrow info to pdJ

xarkes:
This PR adds "arrow" information to the `pdJ` command.
The purpose is that even if `asm.lines` is false, the information of the refline is still present in the json.
I'm not totally satisfied with this solution, but that's the more clean approach I found.

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -859,7 +859,7 @@ static void ds_reflines_init(RDisasmState *ds) {
 
 	lastaddr = UT64_MAX;
 
-	if (ds->show_lines_bb) {
+	if (ds->show_lines_bb || ds->pj) {
 		ds_reflines_fini (ds);
 		anal->reflines = r_anal_reflines_get (anal,
 			ds->addr, ds->buf, ds->len, ds->l,
@@ -1195,6 +1195,17 @@ static void ds_begin_line(RDisasmState *ds) {
 	if (ds->pj) {
 		pj_o (ds->pj);
 		pj_kn (ds->pj, "offset", ds->vat);
+		if (ds->core->anal->reflines) {
+			RAnalRefline *ref;
+			RListIter *iter;
+			// XXX Probably expensive
+			r_list_foreach (ds->core->anal->reflines, iter, ref) {
+				if (ref->from == ds->vat) {
+					pj_kn (ds->pj, "arrow", ref->to);
+					break;
+				}
+			}
+		}
 		pj_k (ds->pj, "text");
 	}
 	ds->buf_line_begin = r_cons_get_buffer_len ();


libr/core/disasm.c
```diff
@@ -856,12 +856,12 @@ static void ds_reflines_fini(RDisasmState *ds) {
...
 }
 
-static void ds_reflines_init(RDisasmState *ds) {
+static void ds_reflines_init(RDisasmState *ds, bool json) {
 	RAnal *anal = ds->core->anal;
 
 	lastaddr = UT64_MAX;
 
-	if (ds->show_lines_bb) {
+	if (ds->show_lines_bb || json) {
```
radare: why not just || ds->pj? and remove this arg

xarkes: Because I'm stupid :-)

radare:
4 XX, pls update the tests

libr/core/disasm.c
```diff
@@ -1179,6 +1179,17 @@ static void ds_begin_line(RDisasmState *ds) {
 	if (ds->pj) {
 		pj_o (ds->pj);
 		pj_kn (ds->pj, "offset", ds->vat);
+		if (ds->core->anal->reflines) {
+			RAnalRefline *ref;
+			RListIter *iter;
+			// XXX Probably expensive
```
radare: thats not probably, i'm pretty sure its an expensive operation. use this:

                list = r_anal_xrefs_get (core->anal, addr);

radare:
ping

xarkes:
There's no rush, I'm not planning to do another PR to fix this one, so let us just finish the work on Cutter side and we'll update this PR accordingly. Thanks :)

radare:
Ok

radare:
ping

xarkes:
I will finish this week end, as said in my initial comment there is no rush. Thanks for your patience.

xarkes:
Hi, I tried using `r_anal_xrefs_get` as you suggested, but I think it's not suitable for our case.
Indeed, if the binary was not analyzed, we cannot get any information about xrefs, and so this won't suit our case.
I suggest we keep it like this for now, especially since `pdJ` is only used by Cutter. I think that we can still optimize in the future if it's really slowing down the GUI.

thestr4ng3r:
`r_anal_xrefs_get()` is definitely not what we want. reflines is generally correct. About the iteration, what's used for regular disasm is `r_anal_reflines_str()` and that is doing the same since at least 2011, as can be seen from this comment:
Bildschirmfoto vom 2019-07-06 15-30-34
So yeah this should definitely be optimized at some point, but as long as `r_anal_reflines_str()` is implemented like this, the pdJ solution with iterating will at least never be slower than regular pd.

radare:
i think the refs and xrefs info should be displayed in ao/aoj too.

radare:
also useful for jmptables

radare:
teh weekend have passed

xarkes:
The current state looks good to me.

radare:
but its red
you didnt updated the tests

xarkes:
Ah lol.

PR #14425, opened 2019-06-25
WIP: Initial support for handling try/catch blocks in DEX code analysis

radare:
in radare2 we get the same as IDA or GHIDRA: (which is wrong)

Screenshot 2019-06-28 at 03 36 10

JEB does it right (is the only tool that supports this)

Screenshot 2019-06-25 at 15 35 17

Ghidra (and IDA afaik) do this (same as r2 right now) , which is wrong

Screenshot 2019-06-25 at 19 57 15

We can do better by supporting node groups or just a single line with dummy [try] and [catch] blocks


radare:
maybe we need to add a field on each RBinSymbol to expose the try/catch/final information and have some commands to modify them. problem i see here is that we may propagate this info into the RAnalFunctions and change the anal loop to honor that in order to add the right control flow lines for catching the exception. not sure if we will require extra changes to support different types of exceptions or excceptions in other langs like c++

feedback on this @thestr4ng3r

radare:
that's how its looking now in r2 with this PR

Screenshot 2019-06-28 at 03 35 12

radare:
lets just merge thiss for now, we will continue in #14442

PR #14429, opened 2019-06-26
mesonbuild: Backport to python 3.5

lzutao:
Python 3.5 (the lowest version that meson supports) has no f-String
but only old %-formatting and str.format().

cc @pelijah


lzutao:
Spurious failure in https://travis-ci.org/radare/radare2/jobs/550612404

PR #14430, opened 2019-06-26
Fix warning when printing level of ">"

lzutao:


libr/anal/meta.c
```diff
@@ -234,8 +234,6 @@ static void r_meta_del_cb (RAnal *a, int type, int rad, SdbForeachCallback cb, v
 		a->cb_printf ("[");
 	}
 
-	RAnalMetaUserItem ui = { a, type, rad, cb, user, 0, NULL };
-
```
radare: Unrelated?

lzutao: Yes. Unrelated but it is unused var.

PR #14433, opened 2019-06-27
Enable ASSERTS on all travis builds, to avoid confusion

ret2libc:


ret2libc:
I don't think the red status is related to this PR. It seems also master has some red builds because it stalled somehow.

radare:
well... master have only one timeout, this one have 4. and there are other PRs with 0 or 1 timeout only and always in the same test. so not sure what's going on

XVilka:
It is general Travis breakage, see the service status.

radare:
confirmed its random by forcing some reruns

PR #14446, opened 2019-06-28
Fix Linux meson build

kazarmy:


radare:
-lmth forwat, which is the error

kazarmy:
cos_sin_err

kazarmy:
https://travis-ci.org/radare/radare2/jobs/551763733#L1101-L1104

radare:
Isnt this just -lm ? Maybe we can move this con/sin into rutil and simplify the build process

kazarmy:
> Isnt this just -lm ?

Yes.

> Maybe we can move this con/sin into rutil

Yes but they'll just be wrappers to -lm ... don't think there's much benefit to copy-pasta from https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/ieee754/dbl-64/s_sin.c;h=26799f1909b953cda9b0249cd61d15ccf42b51d4;hb=HEAD and then figure out what the heck is going on

radare:
FIxed in master

PR #14447, opened 2019-06-28
Fix #9417: pass arguments after -d to debugged program

### LINKED ISSUE #9417: Parsing -d program arguments issues
Notice that it works fine with lldb

screen shot 2018-02-18 at 02 43 01
screen shot 2018-02-18 at 02 43 31

Ka6aSH:
Hi!

I tried to use r2 as a replacement of gdb and wondered that it can't debug a program with arguments in a way that lldb or gdb can.

This PR implements the idea described here https://github.com/radare/radare2/issues/9417#issuecomment-370237123

> consider everything after the -d flag as part of the debuggee arguments.

When we met `-d` option, we stop handling arguments as r2 arguments and wait until they are handled further:
https://github.com/radare/radare2/blob/d3909cab8462816483f0d2a4b2281ddcf22c4a58/libr/main/radare2.c#L1069-L1075

This is not a code that I'm proud of but I couldn't find another more appropriate approach.

gdb uses long options and checks before switch statement:
https://github.com/bminor/binutils-gdb/blob/5bb4fed3d48e4637015a89e40672c938c7f61e72/gdb/main.c#L666-L671

But here short options are used. So we could extract a `-d` check from switch to above but it looks uglier for me than PRed approach:
```c
               if (c == 'd') {
#if DEBUGGER
                       debug = 1;
                       break;
#else
                       eprintf ("Sorry. No debugger backend available.\n");
                       return 1;
#endif
               }

               switch (c) {
```
Of course, there is always `goto` but it is unreasonable here (in my mind ofc).

I'm not sure that test is relevent or not.

If it is, could you please help to write it better?
I can't use `ARGS` for `-d` because it inserts `-i /tmp/<filename>` between `ARGS` and `FILE`
I didn't find option to pass `-l` separately from `FILE`.
Test checks that option `-l` was passed to program and wasn't consumed by radare itself.                                                                                                                                                                                          

```
NAME='r2 -d'    
FILE=-d ../bins/elf/arg -l    
CMDS=i~referer    
EXPECT=<<RUN    
referer  dbg://../bins/elf/arg -l
Hello r2-folks    
RUN    
```


radare:
Seems like there are som tests failing

radare:
can you rebase and test again in travis?

PR #14448, opened 2019-06-28
Continue with the RBin+RBuffer refactoring ##bin

radare:


libr/core/cmd_search.c
```diff
@@ -253,18 +253,21 @@ static int search_hash(RCore *core, const char *hashname, const char *hashstr, u
...
+	int fd = core->file->fd;
+	RBuffer *b = r_buf_new_with_io (&core->anal->iob, fd);
 	r_cons_break_push (NULL, NULL);
 	while (from < to) {
 		if (r_cons_is_breaked ()) {
 			break;
 		}
-		r_io_read_at (core->io, from, buf, sz);
-		plug = r_bin_get_binplugin_by_bytes (core->bin, buf, sz);
+		RBuffer *ref = r_buf_new_slice (b, from, UT64_MAX);
```
ret2libc: this should probably be freed

ret2libc: btw, great we can finally do this! :)

radare:
yeah that how we should go, but this means killing r_buf_data.. or at least stop using it in RBin becausse this is making some plugins crash

ret2libc:
I don’t understand. The base buffer b is anyway freed at the end of the function. Also, all plugins should behave the same so if it’s not please let me know what’s the exact issue and I’ll try to fix it.

radare:
Sure, probably tomorrow i will have time to check this pr again. I pushed it because there was so much to change and improve to get merged and i had to prioriize the other 5 prs i sent meanwhile. The problem i had was with the dol and omf plugins suffering a null deref because rbufdata returns null because the source is an fd not a buffer

radare:
solved conflicts, fixed crash and solved false positives.

current problem i see is that /B is tied to the fd.. which should be tied to the IO honoring all the maps and such

radare:
cc @ret2libc

ret2libc:
We could create a new rbuffer type that just relies on RIO and consider all the maps. I can work on it next days I think

radare:
Yes that will be useful

radare:
I think this can be merged now. Despite it can be improved, itss better to have this in master and work on top of it

PR #14452, opened 2019-06-28
Implement "afsr" to change function return type ##types

oxcabe:
Closes https://github.com/radare/radare2/issues/11404.

Finally!! Check for improvements/bugs.


libr/core/cmd_anal.c
```diff
@@ -2779,6 +2780,23 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
 			free (data);
+		} else if (input[2] == 'r') { // "afsr"
+			if (input[3] == ' ') {
+				const char *ntype = ((input[4]) != '\0')
+					? (input + 4)
+					: NULL;
+				if (ntype) {
+					char *sig = r_core_cmd_str (core, "afs");
+					char *sig_notype = strchr (sig, ' ');
+					char *data = (char *)malloc (strlen (ntype) + strlen (sig_notype) + 1);
```
radare: just use r_str_newf. but ideally we shuold use the api not an r2 command

oxcabe: > just use r_str_newf
Damn. This simplifies it a lot.

> but ideally we shuold use the api not an r2 command
We can leave it like this since it works, then refactor it. But, you know, nobody probably will do that in the near future. It could be refactored now, but it's gonna take some more time for me to deliver.

libr/core/cmd_anal.c
```diff
@@ -2779,6 +2780,21 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
 			r_core_cmdf (core, "\"afs %s\"", data);
 			free (sig);
 			free (data);
+		} else if (input[2] == 'r') { // "afsr"
+			if (input[3] == ' ') {
+				const char *ntype = ((input[4]) != '\0')
```
radare: !=0 is implicit

libr/core/cmd_anal.c
```diff
@@ -2779,6 +2780,21 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
+					? (input + 4)
```
radare: unnecessary parenthesis

libr/core/cmd_anal.c
```diff
@@ -2779,6 +2780,21 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
+					: NULL;
+				if (ntype) {
+					char *sig = r_core_cmd_str (core, "afs");
+					char *sig_notype = strchr (sig, ' ');
+					char *data = r_str_newf ("%s%s", ntype, sig_notype);
+					strtok (data, "\n");
+					r_core_cmdf (core, "\"afs %s\"", data);
+					free (sig);
```
radare: liberate in the inverse order that it was created

libr/core/cmd_anal.c
```diff
@@ -2779,6 +2780,21 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
+			if (input[3] == ' ') {
+				const char *ntype = ((input[4]) != '\0')
+					? (input + 4)
+					: NULL;
+				if (ntype) {
+					char *sig = r_core_cmd_str (core, "afs");
+					char *sig_notype = strchr (sig, ' ');
+					char *data = r_str_newf ("%s%s", ntype, sig_notype);
+					strtok (data, "\n");
+					r_core_cmdf (core, "\"afs %s\"", data);
```
radare: what about just passing %s%s in here instead of data?, the newline shouldnt be a problem if you filter the ntype before using it.

libr/core/cmd_anal.c
```diff
@@ -2779,6 +2780,21 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
 			free (sig);
 			free (data);
+		} else if (input[2] == 'r') { // "afsr"
+			if (input[3] == ' ') {
+				const char *ntype = ((input[4]) != '\0')
+					? (input + 4)
+					: NULL;
+				if (ntype) {
+					char *sig = r_core_cmd_str (core, "afs");
+					char *sig_notype = strchr (sig, ' ');
```
radare: i think that the way to go would be to find '(' and go backward until theres'a space which will be the function name, then you can trim down, but again , using the api you can skip all this text parsing

libr/core/cmd_anal.c
```diff
@@ -2779,6 +2780,21 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
+					char *data = r_str_newf ("%s%s", ntype, sig_notype);
```
radare: unnecessary if you have cmdf below. i will merge after those changes but ideally we should use the api

oxcabe: > ideally we should use the api

We should discuss this.

libr/core/cmd_anal.c
```diff
@@ -2779,6 +2780,24 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
 			r_core_cmdf (core, "\"afs %s\"", data);
 			free (sig);
 			free (data);
+		} else if (input[2] == 'r') { // "afsr"
+			if (input[3] == ' ') {
+				const char *ntype = (input[4])
```
radare: unnecessary parenthesys. but.. if input[3] have a space, input4 should be != ' ' already, because commands are chopped before reaching this point

libr/core/cmd_anal.c
```diff
@@ -2779,6 +2780,24 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
+					? input + 4
+					: NULL;
+				if (ntype) {
+					char *sig = r_core_cmd_str (core, "afs");
+					char *sig_notype = strchr (sig, '('); 
+					strtok (sig_notype, "\n");
+					sig_notype -= 2;
+					while (*sig_notype != ' ' && sig_notype > sig) {
+						--sig_notype;
```
radare: unnecessary predecrement

libr/core/cmd_anal.c
```diff
@@ -2779,6 +2780,24 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
+		} else if (input[2] == 'r') { // "afsr"
+			if (input[3] == ' ') {
+				const char *ntype = (input[4])
+					? input + 4
+					: NULL;
+				if (ntype) {
+					char *sig = r_core_cmd_str (core, "afs");
+					char *sig_notype = strchr (sig, '('); 
+					strtok (sig_notype, "\n");
+					sig_notype -= 2;
```
radare: that pointer arithmetic is wrong. if sig is "(" you'll be pointing oob

libr/core/cmd_anal.c
```diff
@@ -2779,6 +2780,24 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
+					while (*sig_notype != ' ' && sig_notype > sig) {
+						--sig_notype;
+					}
+					// TODO: Use API instead of r2 command
+					r_core_cmdf (core, "\"afs %s%s\"", ntype, sig_notype);;
```
radare: unnecessary double semicolon

libr/core/cmd_anal.c
radare: i think using r_str_rchr (reverse chr search) will be better and simpler than doing all that

radare:
ping

radare:
also fix this 
```
[0x1000011e8]> afs?
<string>:1: error: declaration expected
```

radare:
thiss is how you shsould fix this, without parsing anything

Screenshot 2019-06-30 at 16 12 24

libr/core/cmd_anal.c
```diff
@@ -2802,13 +2810,25 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
 			char *data = r_core_editor (core, NULL, sig);
 			r_core_cmdf (core, "\"afs %s\"", data);
 			free (sig);
 			free (data);
-		} else {
+			break;
+		}
+		case 'r': { // "afsr"
+			RAnalFunction *fcn = r_anal_get_fcn_in (core->anal, core->offset, -1);
+			const char *query = r_str_newf ("anal/types/func.%s.ret=%s", fcn->name, input + 4);
```
radare: This is not const and must be freed

libr/core/cmd_anal.c
```diff
@@ -2802,13 +2810,25 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
 			char *sig = r_core_cmd_str (core, "afs");
 			char *data = r_core_editor (core, NULL, sig);
 			r_core_cmdf (core, "\"afs %s\"", data);
 			free (sig);
 			free (data);
-		} else {
+			break;
+		}
+		case 'r': { // "afsr"
+			RAnalFunction *fcn = r_anal_get_fcn_in (core->anal, core->offset, -1);
```
radare: Chekc if fcn is null. Maybe it doesnt exist

PR #14474, opened 2019-07-03
Fix aho being ignored by asm.tabs ##print

oxcabe:
Closes https://github.com/radare/radare2/issues/14423


oxcabe:
Ok current changes disable asm.tabs, this implies that something else needs to be done in order to fix this. Will update ASAP.

PR #14479, opened 2019-07-03
using new bsd mmap PROT_MAX flag ##magic

devnexen:
To avoid further wider promotions.

### DIFF
--- a/libr/magic/apprentice.c
+++ b/libr/magic/apprentice.c
@@ -1801,7 +1801,10 @@ static int apprentice_map(RMagic *ms, struct r_magic **magicp, ut32 *nmagicp, co
 	}
 
 #ifdef QUICK
-	if ((mm = mmap (0, (size_t)st.st_size, PROT_READ, //OPENBSDBUG  |PROT_WRITE,
+#ifndef PROT_MAX
+#define PROT_MAX(p) 0
+#endif
+	if ((mm = mmap (0, (size_t)st.st_size, PROT_READ|PROT_MAX(PROT_READ), //OPENBSDBUG  |PROT_WRITE,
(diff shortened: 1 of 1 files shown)


libr/magic/apprentice.c
```diff
@@ -1801,7 +1801,10 @@ static int apprentice_map(RMagic *ms, struct r_magic **magicp, ut32 *nmagicp, co
 	}
 
 #ifdef QUICK
-	if ((mm = mmap (0, (size_t)st.st_size, PROT_READ, //OPENBSDBUG  |PROT_WRITE,
+#ifndef PROT_MAX
+#define PROT_MAX(p) 0
+#endif
+	if ((mm = mmap (0, (size_t)st.st_size, PROT_READ|PROT_MAX(PROT_READ), //OPENBSDBUG  |PROT_WRITE,
```
radare: i dont see why we should be calling PROT_MAX here, with PROT_READ twice aand an OR seems like a nop to me. Imho we can safely remove all this |PROT_MAX(PROT_READ*....

devnexen: hmmm PROT_MAX is definitely not a no-op but I could simplify a bit this bit tho.

radare: whats the benefit of introducing this PROT_MAX in a compile-time defined option for mmap? imho prot-max is useful when its controlled by the user. or maybe  im missing  s omething

devnexen: you re not wrong, even though it s masinly because the address is available thus I did not apply this everywhere in the base code but that s fine if this does not get merged tho, just proposal ;-)

radare:
i prefer not to merge this

devnexen:
fair enough :)

PR #14482, opened 2019-07-04
Kill MOV/LEA comments ##disasm

thestr4ng3r:
If anybody finds these useful next to asm.refptr please comment.


radare:
Can you fix the tests too?

thestr4ng3r:
I don't understand these db/formats/mach0/fatio and db/formats/mach0/fatmach0 tests. They fail for me locally on master too.

radare:
because theres a test that modifies the file itself and it works only in travis because everything is checked out from scratch everytime

PR #14486, opened 2019-07-05
Fix new exe header detection for MZ

GustavoLCR:


radare:
can you add a test please? (or @ITAYC0HEN )

PR #14492, opened 2019-07-05
Initial working implementation of bin.libs ##bin

radare:


radare:
@condret pls chk the r_io_location() api, it works for this case but its pretty weak and buggy, so if you have some time pls take a look, because ive been asking for this api for more than 2 years and i didnt wanted to hold this bin.libs feature anymore

PR #14497, opened 2019-07-06
Reduce xrefs sorting for aflj perf ##anal

### LINKED ISSUE #14494: aflj taking lot of time. I can't find the reason.
afl works fine and quite fast. I don't know what is the issue with aflj which takes a lot of time ( more than hours atleast).

thestr4ng3r:
Fix #14494 

Still kind of slow, but this makes it at least about 10 times faster. The remaining bottlenecks may not be as easy to fix or at least not without removing info.

after these changes:
Bildschirmfoto vom 2019-07-06 16-36-38


radare:
Resort all the things!

PR #14510, opened 2019-07-07
expand on FreeBSD sandbox handling.

devnexen:


radare:
I miss some ifdefs here to not break on old freebsds

devnexen:
The 9.x line is EOL https://www.freebsd.org/releases/ any reason to be compatible with these ?

XVilka:
@unixfreaxjp might give an opinion, if 9.x should be supported still.

unixfreaxjp:
Thank you @XVilka, there still are some embedded and ICS based system still using freebsd9 base, and we need radare2 to run on them. Why pushing to not supporting freebsd9 if we still can support it? 
In contrary the current r2 windows binary is running well on WindowsXP... < example.

unixfreaxjp:
Additionally if you want to add something new into newer BSD version, which we've been through about this matter a lot of times, please use version checking. The EOL for ICS devices itself can take for 20-25years FYI.

XVilka:
So yes, @devnexen, please add the proper ifdefs. Radare2 is not a usual program, and can target older devices as well. So it makes sense to keep compatibility as much as possible, unless it requires a lot of the code (like in Windows versions lesser than XP).

devnexen:
ok fair enough @unixfreaxjp @XVilka  I was just considering the package version of FreeBSD did not think of self compiled usage.

unixfreaxjp:
> (like in Windows versions lesser than XP).

oh we have that one too?!

@XVilka

PR #14503, opened 2019-07-07
WIP: Fix #14499: bin.str.validchars: Valid string chars during string detection

kazarmy:
To fix #14499 using `e bin.str.validchars=ascii`. Currently only options are `ascii` and `all`.

PR #14525, opened 2019-07-09
Fix #14499: Detect ascii substrings

### LINKED ISSUE #14499: Missing strings in rabin2 -zz

The following malware contains strings that aren't found by rabin2 -zz
Password "infected"

[Reborn.zip](https://github.com/radare/radare2/files/3364960/Reborn.zip) 

```
556938:http://pomf.cat/upload.php
557277:https://a.pomf.cat/
564205:http://pomf.cat/upload.php
564258:https://a.pomf.cat/
```

For UI requests, don't hesitate to post some draws.

rabin2 -zz find the strings and display them

via @futex

kazarmy:


radare:
pascal strings should be like this:

[len][str]

so we can chk if len == strlen(str) and take those strings as valid, that may reduce false positives

kazarmy:
how many bytes is the [len]?

kazarmy:
ok the strings in #14499 appear to use 1 byte for the [len], but if the strings don't have null terminators, how do i do strlen(str)?

kazarmy:
ok the strings in #14499 appear to use 1 byte for the [len], but if a string doesn't have a null terminator, how do i do strlen(str)?

kazarmy:
and not every wide string is a pascal string

kazarmy:
tbh this pr could just be a longwinded way of defaulting `bin.str.validchars` (#14503) to `ascii`

kazarmy:
... for wide strings

kazarmy:
Closing since #14503 (after defaulting to `ascii`) appears to be the best way to fix #14499.

kazarmy:
On second thought...

kazarmy:
now i think this pr is better than #14503 because the ascii-only constraint can only affect wide16 strings

kazarmy:
ok to merge?

PR #14538, opened 2019-07-10
Parse Mach-O exports trie structure ##bin

mrmacete:


radare:
we need a test! do you have any sample bin that can be shared?

mrmacete:
> we need a test! do you have any sample bin that can be shared?

yeah i think any Mach-O lib which is sripped but has exports could work for a test case - i'll try to find or craft one

PR #14542, opened 2019-07-10
Cache the pdc outcome for each function ##panels

Vane11ope:
GIFrecord_2019-07-11_040102

this video demonstrates the speed after the cache.


radare:
I'm merging this for now but i think it will be good to have a config var to toggle this or integrate it with the pdc command itself so the caching happens not only for panels. also we need a way to invalidate this, i was planning to add a hash with the computed info of the function but we shuold probably use events and invalidate

PR #14556, opened 2019-07-12
Use HtUP to store canvas attributes instead of a sorted array

### LINKED ISSUE #14553: agf is slow
And that's because of the edges

Screenshot 2019-07-12 at 12 02 53

ret2libc:
This should improve performace as you don't need to shift elements of
the sorted array when you need to insert a new one.

Fixes #14553

Please do test the PR before merging. I've already tested on linux on my /bin/ls, but another run would not hurt :)

### DIFF
--- a/libr/cons/canvas.c
+++ b/libr/cons/canvas.c
@@ -27,40 +27,15 @@ static int __getAnsiPiece(const char *p, char *chr) {
 	return p - q;
 }
 
-static const char **__attributeAt(RConsCanvas *c, int loc) {
-	int i, j, delta;
-	if (!c->color || c->attrslen == 0) {
+static void attribute_free_kv(HtUPKv *kv) {
+	free (kv->value);
+}
+
+static const char *__attributeAt(RConsCanvas *c, int loc) {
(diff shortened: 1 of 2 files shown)


radare:
Ill test tonight. But do you have some benchmarks? Like how much times faster or which % of time is spent in this function before the change?

ret2libc:
Tested with:
```
$ time ./binr/radare2/radare2 -qc "s main;af;agf;q" ../../radare2-regressions/bins/elf/analysis/ls-linux64 2>/dev/null 1>&
```

On master:
```
./binr/radare2/radare2 -qc "s main;af;agf;q"  2> /dev/null >&2  6.48s user 0.04s system 99% cpu 6.532 total
```
With this PR:
```
./binr/radare2/radare2 -qc "s main;af;agf;q"  2> /dev/null >&2  0.89s user 0.03s system 99% cpu 0.919 total
```

radare:
Beautiful. Ill try in few mins. Thanks!

radare:
it also improves the performance in panels

Issue #14509, opened 2019-07-07
"ic?~" returns as if searching the output of "i?"


ricardoapl:
Using the internal grep (**~**) with the output of **i?** (help of **i** command) behaves as one would expect. However, if we do the same with the help output of some sub command of **i**, e.g. **ic?**, it looks as if the search for a match is done with the help output of **i**

Below is the help output of **ic** (that is, issuing **ic?** in r2 prompt)

![r2issue-1](https://user-images.githubusercontent.com/48807108/60769984-10f96e00-a0ce-11e9-8b47-e1bb4b5c3c93.png)

### Expected behavior

`ic?~List` should return

![r2issue-0](https://user-images.githubusercontent.com/48807108/60770087-2d49da80-a0cf-11e9-8fa4-ef15c5943ce2.png)

If only searching for the word Header, i.e. `ic?~Header`, only the last of the above two lines should be shown

### Actual behavior
![r2issue-2](https://user-images.githubusercontent.com/48807108/60769998-32f2f080-a0ce-11e9-8945-5a6665ded5e0.png)

which is the same as

![r2issue-3](https://user-images.githubusercontent.com/48807108/60770004-48681a80-a0ce-11e9-873d-0cf84419b892.png)

### Notes
This is actually happening with any sub command starting with **i** (ic, ik, etc.)

It also seems that a few of the tests in radare2-regressions are not taking this expected behavior into account

radare:
I would add this issue tot he other issues related to icq iccq etc. The fix is just one line. Do you want to contribute with that line?

ricardoapl:
When you talk about icq and iccq are you referring to #14177?

Anyway, I'll take a look at this one!

radare:
Yes

PR #14563, opened 2019-07-13
Fix i subcmd help grep

ricardoapl:
This aims to solve issue #14509

### DIFF
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -707,7 +707,7 @@ R_API void r_cons_clear() {
 
 static void cons_grep_reset(RConsGrep *grep) {
 	R_FREE (grep->str);
-	ZERO_FILL(*grep);
+	ZERO_FILL (*grep);
 	grep->line = -1;
 	grep->sort = -1;
 	grep->sort_invert = false;
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -706,14 +706,11 @@ R_API void r_cons_clear() {
 }
 
 static void cons_grep_reset(RConsGrep *grep) {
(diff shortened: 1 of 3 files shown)


ricardoapl:
### Details

As told in radare2-regressions PR [#1880](https://github.com/radare/radare2-regressions/pull/1880)

> the reason for this bug is because ic? is doing i? ic internally, and when you do ic? it is overriding the previous grep exprerssion

To be more precise, issuing **ic?~List** in r2 prompt will lead to the following backtrace in gdb

github-bt-ic-igrep-List

Part of the issue is due to the fact that consecutive calls to _r_core_cmd_subst_i()_, within
the context of a same _r_core_prompt_exec()_ call, will (each) trigger a call to _parse_grep_expression()_
which is zeroing the RConsGrep structure.

This RConsGrep structure keeps, among other things, the grep strings to be used later on for filtering.

We could solve this by zeroing the RConsGrep structure on each call to _cons_grep_reset()_ instead of
_parse_grep_expression()_, after all that seems to be the job of such function.

However, this still does not solve the problem. If we apply the above changes and issue **ic?~List** we get as output

github-OR

It seems that the expressions are getting OR'ed as opposed to AND'ed, i.e., it looks for strings with
either " ic" or "List".

By prepending "&" to the first grep expression (see stack frame 5) we can force the AND of current expression with any other to come.

An alternative to all this would be to refactor the _cmd_info()_ function so as to make each sub command generate its own help message, just like what happens with _cmd_print()_ or _cmd_anal()_.

radare:
The proposed solution lgtm. Can you update the tests in db/cmd/cmd_i?

PR #14568, opened 2019-07-14
Add comment that demangles calls if asm.demangle=true

kazarmy:


libr/core/disasm.c
```diff
@@ -3946,6 +3947,25 @@ static void ds_print_ptr(RDisasmState *ds, int len, int idx) {
 #endif
 }
 
+static void ds_print_demangled(RDisasmState *ds) {
+	if (!ds->show_comments || !ds->asm_demangle) {
+		return;
+	}
+	RCore *core = ds->core;
+	RFlagItem *f = NULL;
```
radare: no need to initialize, you can assign it below too if you add some braces

libr/core/cbin.c
```diff
@@ -2067,6 +2067,7 @@ static int bin_symbols(RCore *r, int mode, ut64 laddr, int va, ut64 at, const ch
 				RFlagItem *fi = r_flag_set (r->flags, fnp, addr, symbol->size);
 				if (fi) {
 					r_flag_item_set_realname (fi, n);
+					fi->demangled = !!sn.demname;
```
radare: just a cast should be fine (bool)(size_t)

libr/core/disasm.c
kazarmy: ok f1680776356163b9c30486497b89525a92a744c1

libr/core/cbin.c
kazarmy: ok f1680776356163b9c30486497b89525a92a744c1

PR #14583, opened 2019-07-17
Right click is supported to open " widget. ##panels

Vane11ope:
GIFrecord_2019-07-18_034156
Right click opens " widget.
Left/Right click outside of the widget closes the widget.
Click on the widget is not supported just yet but coming.
Also you can use wheel to scroll up and down the widget.

### DIFF
--- a/libr/cons/input.c
+++ b/libr/cons/input.c
@@ -148,6 +148,7 @@ R_API int r_cons_arrow_to_hjkl(int ch) {
 				int vn = 0;
 				do {
 					ch = r_cons_readchar ();
+					// just for debugging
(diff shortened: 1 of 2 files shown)


libr/cons/input.c
```diff
@@ -166,9 +167,18 @@ R_API int r_cons_arrow_to_hjkl(int ch) {
 						p = 0;
 					}	
 				} while (ch != 'M' && ch != 'm');
+				// right click
+				if (atoi (vel) == 2) {
```
radare: we can probably compute atoi(vel) once and make a switch statement below instead of atoiing 3 times  the same thing

Issue #14277, opened 2019-06-10
Correctly show Relocs as flags in disassembly


ITAYC0HEN:
**Is your feature request related to a problem? Please describe.**
Calls to relocated functions are considered CALL $+1 instead of a relocated call. Radare2 successfully shows that the address is relocated but the target function is $+1 instead of the relocated name

![image](https://user-images.githubusercontent.com/20182642/59208570-ff49a700-8bb1-11e9-8611-8dd01591824d.png)

**Describe the solution you'd like**

Should look clear as in in the following screenshot. Notice the calls and the MOVs
![image](https://user-images.githubusercontent.com/20182642/59208617-1ab4b200-8bb2-11e9-93c0-13c397f56223.png)

Binary:
[tif_fax3.zip](https://github.com/radare/radare2/files/3272826/tif_fax3.zip)

PR #14587, opened 2019-07-18
Show relocs as flags in disassembly for call and jmp instr.

GustavoLCR:
#14277
This only works for call and jmp instructions

image

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -4782,7 +4782,11 @@ static char *ds_sub_jumps(RDisasmState *ds, char *str) {
 	if (fcn) {
 		name = fcn->name;
 	} else if (f) {
-		RBinReloc *rel = r_core_getreloc (ds->core, addr, ds->analop.size);
+		RBinReloc *rel;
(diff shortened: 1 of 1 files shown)


libr/core/disasm.c
```diff
@@ -4782,7 +4782,11 @@ static char *ds_sub_jumps(RDisasmState *ds, char *str) {
 	if (fcn) {
 		name = fcn->name;
 	} else if (f) {
-		RBinReloc *rel = r_core_getreloc (ds->core, addr, ds->analop.size);
+		RBinReloc *rel;
+		rel = r_core_getreloc (ds->core, ds->analop.addr, ds->analop.size);
```
radare: umf, why addr != ds->analop.addr ?

GustavoLCR: ds->analop.addr is the address of the current instruction
addr is where the instruction is telling you to jump

radare: confusing varnames then :) ok to merge

PR #14589, opened 2019-07-18
Fix eval multiple assignment and quotes

GustavoLCR:
R2R: https://github.com/radare/radare2-regressions/pull/1891


radare:
The multiple assignment feature should never come back. Wit makes the pasting very complicated and we cant do things like e cmd.prompt=om=

GustavoLCR:
You should be able to do this with `e cmd.prompt="om="` in this patch, but I guess its not really intuitive. You can close this pr if its not needed

PR #14592, opened 2019-07-18
Fix demangled comment order

kazarmy:
... to be consistent with C functions.


kazarmy:
"to be consistent" -> "to be more consistent"

PR #14596, opened 2019-07-19
Add r_parse_ctype ##parse

thestr4ng3r:
This takes strings like `const char * [0x42] const * [23]` and parses them into structs.

Test: https://github.com/radare/radare2-regressions/pull/1893


libr/parse/ctype.c
```diff
@@ -0,0 +1,195 @@
...
+				&& strcmp (child->children[2]->contents, "]") == 0) {
+			RParseCTypeType *array = R_NEW0 (RParseCTypeType);
+			if (!array) {
+				goto beach;
+			}
+			array->kind = R_PARSE_CTYPE_TYPE_KIND_ARRAY;
+			array->array.count = strtoull (child->children[1]->contents, NULL, 0);
+			array->array.type = cur;
+			cur = array;
+		}
```
radare: wrong indent

libr/parse/ctype.c
```diff
@@ -0,0 +1,195 @@
...
+				goto beach;
+			}
+			pointer->kind = R_PARSE_CTYPE_TYPE_KIND_POINTER;
+			pointer->pointer.is_const = true;
+			pointer->pointer.type = cur;
+			cur = pointer;
+		}
+
+		// <array>
+		else if (strcmp (child->tag, "array|>") == 0
```
radare: woah... wtf... can you move all this complex logic into a separate function so the if block is only 1 readable statement?

thestr4ng3r:
This is only a very small subset and not even entirely valid C since C wouldn't allow things like `char[42] a;`, but this only parses the type part without `a`, so we need to support it it.
If it's only about parsing type definitions, yeah maybe it would be possible to switch to mpc, but it will be WAY more complex than this.

libr/parse/ctype.c
thestr4ng3r: Maybe I'm blind right now, but I don't see what's wrong.

XVilka:
@radare yes, because tcc also handles preprocessor, and GNU `__attribute__(())`, so it is even more complex

thestr4ng3r:
Those lbtms are in mpc.c btw

PR #14598, opened 2019-07-19
Add rep* type in isStepOverable ##debug

Maijin:
Add rep* type in isStepOverable ##debug


kamou:
@Maijin I realised we were not talking about the same problem
dso already works fine with rep and repne prefixed instructions, the problem was with dsui,
I opened a new PR for it: https://github.com/radare/radare2/pull/14601

I guess this one can be closed. Thanks again for your help !

PR #14603, opened 2019-07-21
Fix grep in tasks ##cons

thestr4ng3r:
Tests: https://github.com/radare/radare2-regressions/pull/1895

Regression introduced by https://github.com/radare/radare2/pull/14563, though the pr itself is good.

You would get unwanted newlines before the output, which was especially annoying in Cutter's console where every command is run in a task.

Bildschirmfoto vom 2019-07-21 12-24-07


thestr4ng3r:
Haha lol looks like ?E isn't such a good idea for tests

PR #14620, opened 2019-07-22
Fix #14278 - Fix function detection on coff file

### LINKED ISSUE #14278: radare2 fails to detect relocated functions even with debug symbols
### Work environment

| File format of the file you reverse (mandatory)      | PE & COFF
| Architecture/bits of the file (mandatory)            | x64.
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.6.0-git 22334 @ linux-x86-64 git.3.5.1-210-g759ebd5
commit: 759ebd5bfd2f8572d6866fffa7e26b7c739329f4 build: 2019-06-10__18:43:56

Radare2 doesn't detect may function and fails to show the name of those which it does detect

image

### Expected behavior
Radare2 should show all the function in the binary, with their respective names.

Screenshot from IDA:

image

### Actual behavior
Radare shows only 3 functions, with not accurate names

1. Open the attached binary
2. Analyze using `aaa`
3. run `afl` to list the functions
4. observe that almost no functions detected

Binary:

[tif_tile.zip](https://github.com/radare/radare2/files/3401461/tif_tile.zip)

GustavoLCR:
Fix #14278
cc @ITAYC0HEN

### DIFF
--- a/libr/bin/p/bin_coff.c
+++ b/libr/bin/p/bin_coff.c
@@ -70,7 +70,11 @@ static bool _fill_bin_symbol(struct r_bin_coff_obj *bin, int idx, RBinSymbol **s
 		ptr->type = r_str_const (R_BIN_TYPE_SECTION_STR);
 		break;
 	case COFF_SYM_CLASS_EXTERNAL: // should be prefixed with sym.imp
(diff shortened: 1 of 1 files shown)


radare:
Nice!

Issue #13766, opened 2019-04-15
aai should also sum the Cd (data dwords) as covered code


radare:
doing aaaa in /bin/ls is making everything to be analuzed, but as long as there are jump tables, and some constants are inlined in the code results in never being 100% coverage for `aai`

PR #14633, opened 2019-07-23
Sum the meta_data_code as covered code

ycarmon:


ycarmon:
This code should fix issue #13766, it calculates the size of all R_META_TYPE_DATA (Cd) when calculating covered code in compute_coverage() in cmd_anal.c

PR #14638, opened 2019-07-24
Add help for /ac

thestr4ng3r:


radare:
i think /ac can be a subcommand of /aa

PR #14642, opened 2019-07-24
Add e_machine values

eShuttleworth:
I've run into a few cases in which rabin2 didn't recognize the e_machine value in an elf header, this PR adds a bunch from http://www.sco.com/developers/gabi/latest/ch4.eheader.html, which was the most complete source I could find.

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -1992,7 +1992,7 @@ ut64 Elf_(r_bin_elf_get_main_offset)(ELFOBJ *bin) {
 	}
 #endif
 	/* linux64 pie main -- probably buggy in some cases */
-	int bo = 29; // Begin offset may vary depending on the entry prelude
+	int bo = 29; // Begin offset may vary depending on the entry prelude 
 	if (buf[0] == 0xf3 && buf[1] == 0x0f && buf[2] == 0x1e && buf[3] == 0xfa) {
 		// Change begin offset if binary starts with 'endbr64'
 		bo = 33;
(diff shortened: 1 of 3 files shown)


libr/bin/format/elf/elf.c
```diff
@@ -2135,8 +2135,9 @@ char* Elf_(r_bin_elf_get_arch)(ELFOBJ *bin) {
 	case EM_LANAI:
 		return strdup ("lanai");
 	case EM_VIDEOCORE3:
-	case EM_VIDEOCORE4:
-		return strdup ("vc4");
```
radare: Theres no vc3 plugin. Why did you removed the vc4 one? We have no tests but this was working fine

libr/bin/format/elf/elf.c
```diff
@@ -2149,90 +2150,166 @@ char* Elf_(r_bin_elf_get_arch)(ELFOBJ *bin) {
...
-	case EM_OPENRISC:    return strdup ("OpenRISC 32-bit embedded processor");
-	case EM_ARC_A5:      return strdup ("ARC Cores Tangent-A5");
-	case EM_XTENSA:      return strdup ("Tensilica Xtensa Architecture");
-	case EM_AARCH64:     return strdup ("ARM aarch64");
-	case EM_PROPELLER:   return strdup ("Parallax Propeller");
-	case EM_MICROBLAZE:  return strdup ("Xilinx MicroBlaze");
-	case EM_RISCV:       return strdup ("RISC V");
-	case EM_VIDEOCORE3:  return strdup ("VideoCore III");
-	case EM_VIDEOCORE4:  return strdup ("VideoCore IV");
+	case EM_NONE:          return strdup ("No machine");
```
radare: Maybe good to sort alphabetically

libr/bin/format/elf/elf.c
eShuttleworth: Whoops, that's leftover from when I was first figuring out how this worked.  Fixed.

libr/bin/format/elf/elf.c
eShuttleworth: The ordering is by the actual value of the E_MACHINE defines, eg EM_NONE=0, EM_M32=1, etc.  This how it was originally ordered, and I don't think it makes much sense to sort in any other way.

PR #14649, opened 2019-07-25
Swap /ac and /aa to be correct again

thestr4ng3r:
They were swapped in https://github.com/radare/radare2/commit/ff3d6ff0a50b184bc169db64fb4d3424d856ce29


radare:
Imho ‘ac’ reminds me to case sensitive and ‘aa’ is about aproximated asm (insensitive)

radare:
so imho this change shouldnt be merged. but ive been asking for some time for feedback on how to merge /a and /c, and i think it will be good to get some proposals of changes in the /a? instead of patches or issues

thestr4ng3r:
I don't really care which way it is, I just want it to be consistent with the help. So if you prefer it, I can also just fix the help instead.

radare:
i fixed the help in master

PR #14664, opened 2019-07-26
WIP: emacs keybindings

deepakchethan:
* [x] kill-line (C-k) Kill the text from point to the end of the line.
* [x] backward-kill-line (C-x Rubout) - Kill backward from the cursor to the beginning of the current line.
* [x] kill-whole-line (C-u) Kill all characters on the current line, no matter where point is. By default, this is unbound.
* [x] kill-word (C-t) - Kill from point to the end of the current word, or if between words, to the end of the next word. Word boundaries are the same as forward-word.
* [x] backward-kill-word (M-DEL)Kill the word behind point. Word boundaries are the same as backward-word.
* [x] shell-backward-kill-word (C-o) Kill the word behind point. Word boundaries are the same as shell-backward-word.
* [x] unix-word-rubout (C-w)Kill the word behind point, using white space as a word boundary. The killed text is saved on the kill-ring.
* [x] yank (C-y) Yank the top of the kill ring into the buffer at point.
* [x] yank-pop (C-z) Rotate the kill-ring, and yank the new top. You can only do this if the prior command is yank or yank-pop.

### DIFF
--- a/libr/cons/dietline.c
+++ b/libr/cons/dietline.c
@@ -30,6 +30,8 @@ typedef enum {
 	CONTROL_MODE
 } RViMode; 
 
+bool enable_yank_pop = false;
+
 static inline bool is_word_break_char(char ch, bool mode) {
 	int i;
 	if (mode == MAJOR_BREAK) {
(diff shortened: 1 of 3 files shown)


libr/cons/dietline.c
```diff
@@ -30,6 +30,8 @@ typedef enum {
 	CONTROL_MODE
 } RViMode; 
 
+int enable_yank_pop = false;
```
XVilka: use `bool`

libr/cons/dietline.c
```diff
@@ -1657,7 +1676,7 @@ R_API const char *r_line_readline_cb(RLineReadCallback cb, void *user) {
 #endif
 	int rows, columns = r_cons_get_size (&rows) - 2;
 	const char *gcomp_line = "";
-	static int gcomp_idx = 0;
+	static int flag = 0, gcomp_idx = 0;
```
XVilka: 1. use bool
2. use meaningful name

PR #14675, opened 2019-07-27
More fixes to WinDbg

GustavoLCR:
* Fix seeking (stepping now works)
* Fix more crashes
* Fix compiler warnings

### DIFF
--- a/libr/debug/p/debug_windbg.c
+++ b/libr/debug/p/debug_windbg.c
@@ -63,7 +63,7 @@ static void wstatic_debug_break(void *u) {
 
 static RDebugReasonType r_debug_windbg_wait(RDebug *dbg, int pid) {
 	RDebugReasonType reason = R_DEBUG_REASON_UNKNOWN;
-	kd_packet_t *pkt;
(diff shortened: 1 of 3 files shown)


XVilka:
What was the reason behind removing this?
```c
if (pkt->leader == KD_PACKET_DATA) {
			kd_send_ctrl_packet (ctx->io_ptr, KD_PACKET_TYPE_ACKNOWLEDGE, 0);
		}
```

GustavoLCR:
> What was the reason behind removing this?
> 
> ```c
> if (pkt->leader == KD_PACKET_DATA) {
> 			kd_send_ctrl_packet (ctx->io_ptr, KD_PACKET_TYPE_ACKNOWLEDGE, 0);
> 		}
> ```

I added this before because I didn't see it was already being sent on kd_read_packet, so this isn't needed.

See:
https://github.com/radare/radare2/blob/f482eecb02b8bd1a1b0470dee516bd60f4a61e50/shlr/windbg/kd.c#L106-L117

Issue #7515, opened 2017-05-17
how different #!python is from a regular python shell


kamou:
I am not sure if this is the right place to ask this question. If it's not just hit me with a stick and close this issue.

I'm trying to play with [Triton](https://github.com/JonathanSalwan/Triton) from rlang (#!python)
and it seems the behavior is quite not the same as if I played with it from a regular shell.

I'm just trying to disassemble a stupid instruction.
this is how I do it and what I get in a regular ipython shell:
```
In [1]: import triton                     
In [2]: triton.setArchitecture(triton.ARCH.X86_64)
In [3]: opcode = "\x55"               
In [4]:  inst = triton.Instruction() 
In [5]:  inst.setOpcodes(opcode)
In [6]:  inst.setAddress(0x1234)
In [7]: triton.disassembly(inst)
In [8]: print inst
0x1234: push rbp
```

now with r2's #!python:
```
In [1]: import triton                     
In [2]: triton.setArchitecture(triton.ARCH.X86_64)
In [3]: opcode = "\x55"               
In [4]:  inst = triton.Instruction() 
In [5]:  inst.setOpcodes(opcode)
In [6]:  inst.setAddress(0x1234)
In [7]: triton.disassembly(inst)
In [8]: print inst
0x1234: push
```
As you can see the disassembled instruction is not correct.
I'm not asking you to fix Triton or anything, just trying to understand what could lead to a different behavior.

radare:
How many versions of python are installed on your system, which is the path of those imported modules, are all the same and the last version? Against which version do the rlang plugin was compiled? Against which version the r2api module was compiled for?

radare:
anyway this seems to be a problem with triton, not r2

kamou:
sorry, I was afk, The python versions are the same.
tracing with strace shows that both load the same triton module (only one available).
now regarding the r2api and rlang plugin it is the versions installed by r2pm. (I cleaned .config/radare2 today, and reinstalled lang-python today)
There is only one version of python available in my environment.
I'll continue investigating on Triton side if you believe this is a Triton issue.
Thanks.

kamou:
ok narrowing down the issue, it lead me to capstone.
it seems like triton is not using the system wide capstone lib but the one provided / linked with r2 when launching it from rlang.
enabling CS_RELEASE flag in the root makefile make the issue disapear.
Should I investigate the problem further and provide another patch in capston-patches ?
Or do you have a better suggestion ?

kamou:
[...]

PR #14677, opened 2019-07-27
Fix capstone symbols visibility

kamou:
related to issue #7515


kamou:
dunno why, just removing the attribute won't help, I had to explicitly set the attribute as hidden

radare:
i think what this patch is doing is making the build of the shared lib of capstone with hidden symbols, so its an empty dylib, im not sure why this should fix the issue...

kamou:
I know, it is quiet weird, but without this, capstone symbols end up in the relocation section `.rela.plt`
but maybe that's because we're linking against capstone and their header explicitly sets the default visibility ?
another weird thing is that just removing the attribute won't help. I need to explicitly set it to hidden for the symbols to get removed from libr_anal and libr_asm relocation sections...

kamou:
I can try to have a deeper look at it next week

PR #14681, opened 2019-07-28
Refactoring

Vane11ope:


libr/core/panels.c
```diff
@@ -281,8 +281,9 @@ static void __init_all_dbs(RCore *core);
 static void __init_panel_param(RCore *core, RPanel *p, const char *title, const char *cmd);
 static RPanels *__panels_new(RCore *core);
 static void __init_new_panels_root(RCore *core);
-static void __init_menu_saved_layout (RCore *core);
-static void __init_menu_disasm_settings_layout (void *_core);
+static void __init_menu_saved_layout (void *core, const char *parent);
```
ret2libc: we don't use the space between the function name and the parameters.
So it should be
```C
static void __init_menu_saved_layout(void *core, const char *parent);
```

radare:
try to use meaningful commit names please

PR #14692, opened 2019-07-29
Fix flag name demangling on local DLL symbols

GustavoLCR:


libr/bin/demangle.c
```diff
@@ -91,6 +91,13 @@ R_API char *r_bin_demangle(RBinFile *bf, const char *def, const char *str, ut64
 				break;
 			}
 		}
+		size_t len = strlen (bin->file);
+		if (!r_str_ncasecmp (str, bin->file, len)) {
```
radare: why is the filename affecting this the demangling output? can you show a screenshot about the issue?

GustavoLCR: @radare The filename is included in the symbol name:
https://github.com/radare/radare2/blob/7d30ff52fc1be6f9698b166107b8981eab6ec7ba/libr/bin/format/pe/pe.c#L3025

 but isn't on the libs list, so it doesn't get skipped
ie: `GameLogic.dll_?HideCountdown@IPlayer@@QAEXXZ`

Would it be better to change that line and remove the library name from the symbol altogether?
`?HideCountdown@IPlayer@@QAEXXZ`

GustavoLCR: image
vs

image

XVilka: I remember @Maijin asked to include library name to the symbol.

radare:
so wat du @XVilka

libr/bin/demangle.c
Maijin: ? Nop, I said, we could always keep it under an eval because it can be useful time to time when there is confusion onto where one comes from.

radare: @GustavoLCR can do do the change as suggested by @Maijin ?

radare:
ping

GustavoLCR:
Sorry, was busy last few days. I implemented a var to show the library names on the demangled symbols, @Maijin, is that what you meant? It is set to true by default, should I change it to false? 
Also, I couldn't access config from rbin, so it isn't perfect.

libr/bin/demangle.c
```diff
@@ -62,7 +62,7 @@ R_API int r_bin_demangle_type(const char *str) {
 	return R_BIN_NM_NONE;
 }
 
-R_API char *r_bin_demangle(RBinFile *bf, const char *def, const char *str, ut64 vaddr) {
+R_API char *r_bin_demangle(RBinFile *bf, const char *def, const char *str, ut64 vaddr, bool libs) {
 	int type = -1;
```
radare: this function proabbly requires an `r_return_val_if_fail (bf && def && str, NULL);`

GustavoLCR: some places call this function with def = NULL, so i've used only `bf && str`

radare:
can you check those asserts?

Screenshot 2019-08-04 at 16 00 26

GustavoLCR:
> can you check those asserts?
That was because bf was NULL because of -n.
I checked and the assertion isn't really needed, the code accepts any of the parameters to be NULL and work with what it has, so I removed it.

libr/bin/demangle.c
radare: having def = NULL lgtm, probably only chking for bf and str should be good, but you did removed it completely

GustavoLCR: bf can be null as well if bin info wasnt loaded, so do i just check str?

radare:
Thats not how asserts should be fixed but anyway. Lets merge and we can improve that later

PR #14694, opened 2019-07-30
Fix #14655 - Improve asm.pseudo for Dalvik ##disasm

### LINKED ISSUE #14655: Shortening Dalvik/Java method names
Dex code will be more readable if we can shorten the names of the methods. and not just by using the afii/aii imports things. but also by removing the arugments expected to be processed by the function. dex code is usually generated by compilers so it must be verified and not call methods that have different argument types. decompilers and type analysis can still take this info from the signature. But we need a way to make the demangled output shorter some how..

here's an example.

```
- sym.imp.Landroid_content_Intent.method.getComponent__Landroid_content_ComponentName
+ intent.getComponent
```

another example:

```
- sym.imp.Ljava-util_ArrayList.method_size_I
+ ArrayList.size
```

this "issue" applies to any dex file, no need to submit a new one

radare:
class, method and field names are now shorter and easier to read

### DIFF
--- a/libr/bin/bin.c
+++ b/libr/bin/bin.c
@@ -42,11 +42,11 @@ static int __getoffset(RBin *bin, int type, int idx) {
 	return -1;
 }
 
-static const char *__getname(RBin *bin, int type, int idx) {
+static const char *__getname(RBin *bin, int type, int idx, bool sd) {
 	RBinFile *a = r_bin_cur (bin);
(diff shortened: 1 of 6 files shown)


radare:
before:
Screenshot 2019-07-30 at 04 05 14


afterr:
Screenshot 2019-07-30 at 04 01 11

PR #14698, opened 2019-07-30
asm.cmt.fcnrefs: Show name/flag of function ref in disasm (true, false, nodup)

kazarmy:
Separate config var so that i can disable it in Visual mode specifically for functions.


radare:
funrefs?

kazarmy:
sorry, fcnrefs

radare:
What are fcnrefs? Xrefs?

ret2libc:
Can you provide a before/after example?

kazarmy:
Hmm maybe i don't need to bother with the config var after all.

PR #14703, opened 2019-07-30
Don't show function name comment in Visual mode

kazarmy:


kazarmy:
Because they are not needed.

radare:
Screenshots please

PR #14709, opened 2019-07-30
add command icqj

rfc2119:
Closes https://github.com/radare/radare2/issues/14276. `icqq` was already implemented by someone else
btw, why do we need `idx` in here ?
https://github.com/radare/radare2/blob/1f8366427d0fd9890948381d9ee94e6a482c0aa4/libr/core/cmd_info.c#L1081

whoops! seems the output of `icqj` is the same as `icj`, please don't merge


libr/core/cmd_info.c
```diff
@@ -1152,12 +1152,16 @@ static int cmd_info(void *data, const char *input) {
 						mode = R_MODE_CLASSDUMP;
 						RBININFO ("classes", R_CORE_BIN_ACC_CLASSES, NULL, r_list_length (obj->classes));
 						input = " ";
-					} else {
+					} else { // "icq"
+						if (input[2] == 'j')
```
radare: open brace in this line {, not below

PR #14713, opened 2019-07-31
Add fcn name in Visual mode if necessary when asm.filter=false and asm.jmpsub=false

kazarmy:
Before:
before_no_filter_fcn_name

After:
after_no_filter_fcn_name

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -3281,13 +3281,21 @@ static void ds_print_fcn_name(RDisasmState *ds) {
 		return;
 	}
 	RAnalFunction *f = fcnIn (ds, ds->analop.jump, R_ANAL_FCN_TYPE_NULL);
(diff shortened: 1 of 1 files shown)


radare:
thats why i was asking for a screenshot. ihave the feeling that at som epoint we will need to do another huge refactoring in this disasm.c

kazarmy:
> thats why i was asking for a screenshot.

well the screenshot wouldn't have helped since i don't usually set the 2 config vars to false ... but i did manage to remember you asking me about this particular configuration

> ihave the feeling that at som epoint we will need to do another huge refactoring in this disasm.c

it's still manageable ... it's basically about ordering and coordinating the disasm pipeline stages properly

radare:
we should handle all that stuff with proper data structures and not so many is and chks depending on order of display, but yeah, disasm.c is a key piece :D

PR #14717, opened 2019-07-31
Fix #12750 - support type propagation for SUB, ADD

### LINKED ISSUE #12750: Structure offset is not autorecognized in the SUB instruction register offset
When e.g. structs or pointers are hold in registers for a long time, I would find it helpful if i could annotate this using the registers name. E.g. renaming "rbx" to "important_struct_ptr" for a specific part of the binary.

A command, which takes 2 addresses (start, end of annotation range), the register name and the new name, would be great.

**Describe alternatives you've considered**
If there is any other clever way to annotate such things, which not involves commenting every line by hand, please let me know.

XVilka:
In the future it should be isolated into the function.

### DIFF
--- a/libr/anal/p/anal_x86_cs.c
+++ b/libr/anal/p/anal_x86_cs.c
@@ -1884,6 +1884,76 @@ static void op_fillval(RAnal *a, RAnalOp *op, csh *handle, cs_insn *insn) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


radare:
i think the tests output should be updated

radare:
screenshots pls

XVilka:
This is how it looks now
image

PR #14728, opened 2019-07-31
Update flag.c

pelijah:


radare:
your patch looks correct but seems like it's breaking a test

radare:
```
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -i /tmp/tmp-16633WSNePnOtKJIz ../bins/elf/analysis/main
aaa
e asm.cmt.refs = true
f test.flag @ 0x004003e0
CC test comment @ 0x004003e0
pd 1 @ 0x0040050f

[35m-|           0x0040050f      e8ccfeffff     call sym.imp.puts           ; int puts(const char *s) ; (test.flag) ; (test comment)[39m
[35m[39m[32m+|           0x0040050f      e8ccfeffff     call sym.imp.puts           ; int puts(const char *s) ; (test comment)[39m
[32m[39m
EXPECT=<<EOF
```

radare:
basically '(test.flag) ' is gone

radare:
ok found and fixed

PR #14730, opened 2019-07-31
Linux debug plugin, page promotion feature proposal.

devnexen:
Providing new 'dmL' debugger sub command.
This is the only system having THP feature, others are more into
 direct huge tlb equivalent ...
The system ought to be set in madvise mode and disabled on Android,
would not do any good in this environment if this is ever supported.
Once the syscall succeed, the address range ought to get the
 little 'hg' VmFlags.


radare:
This seems to be a linux-specific feature and adding a generic api in the plugins callbacks is probably a bit too much, what about adding an option argument to map_alloc?

What's the use case for this? how to deallocate them?

devnexen:
It is really Linux specific. You just deallocate with dm- as usual.

PR #14749, opened 2019-08-03
Change the look of the p=

deepakchethan:
# After Change:
### With scr.utf8=true:
Screen Shot 2019-08-03 at 2 09 21 PM

### With src.utf8=false
Screen Shot 2019-08-03 at 2 09 41 PM

# Before:
Screen Shot 2019-08-03 at 2 07 36 PM


radare:
Uhm what about making those bars out of bgcolor blocks instead of dashes? Maybe we can have an extra config bar forthis. We already have vertical graphs like this

XVilka:
I think we should keep the dashes, but also allow to switch to the blocks for both vertical and horizontal. Like this:
image

radare:
Agree

deepakchethan:
Needs better config var name, will change based on you input: 

Screen Shot 2019-08-04 at 8 09 03 AM

PR #14773, opened 2019-08-08
Fix wayward rip-relative flag

kazarmy:


radare:
Thats a hack and doesnt belongs to disasm.c. Rip is arch specific, prsing that must be in rparse not in disasm

kazarmy:
is a8bac9db939709a44e52aed82f5bc9515dc2220e acceptable?

radare:
Looks better but i have to try . Any screenshot? Im worried about messing the disasm.c more

libr/core/disasm.c
```diff
@@ -1009,6 +1009,10 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
 	core->parser->relsub = r_config_get_i (core->config, "asm.relsub");
 	core->parser->regsub = r_config_get_i (core->config, "asm.regsub");
 	core->parser->relsub_addr = 0;
+	if (core->parser->relsub && ds->analop.type == R_ANAL_OP_TYPE_LEA
+	    && (ut64)ds->analop.ptr == ds->vat + ds->analop.size + ds->analop.disp) {  // PC-relative
```
radare: imho the op->ptr should be pointing to the right address without having to do this pc relative addressing in disasm.c, the anal backend should give the right pointer. and disp shouldnt be used at all

kazarmy: ok 2a5f7c8b1a50e550e751d2b3f18a9b15a74cbcba

libr/core/disasm.c
```diff
@@ -1009,6 +1009,9 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
 	core->parser->relsub = r_config_get_i (core->config, "asm.relsub");
 	core->parser->regsub = r_config_get_i (core->config, "asm.regsub");
 	core->parser->relsub_addr = 0;
+	if (core->parser->relsub && ds->analop.type == R_ANAL_OP_TYPE_LEA && ds->analop.ptr != UT64_MAX) {
+		core->parser->relsub_addr = ds->analop.ptr;
```
radare: Is this still necessary after the change in anal?

libr/core/disasm.c
```diff
@@ -1070,7 +1073,7 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
 				core->parser->flagspace = NULL;
 			}
 		}
-		if (ds->analop.refptr) {
+		if (core->parser->relsub && ds->analop.refptr) {
```
radare: Arent we missing some references here? imho relsub shouldnt block refptr to work

libr/core/disasm.c
```diff
@@ -1009,6 +1009,9 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
 	core->parser->relsub = r_config_get_i (core->config, "asm.relsub");
 	core->parser->regsub = r_config_get_i (core->config, "asm.regsub");
 	core->parser->relsub_addr = 0;
+	if (core->parser->relsub && ds->analop.type == R_ANAL_OP_TYPE_LEA && ds->analop.ptr != UT64_MAX) {
```
radare: why restrict on lea?

libr/core/disasm.c
kazarmy: The assignment to `core->parser->relsub_addr` still needs to be done.

libr/core/disasm.c
kazarmy: `relsub` is just a proxy for the config var `asm.relsub`, and not checking it will result in situations where `asm.relsub=0` but the substitution is still done. There is a test for this.

libr/core/disasm.c
kazarmy: Umm the only test case i have uses lea -- test cases can be generated for other instructions though

PR #14777, opened 2019-08-09
Some code cleanup in RUtil.lib

radare:


libr/util/lib.c
```diff
@@ -73,120 +58,104 @@ R_API void *r_lib_dl_open(const char *libname) {
...
-	const char *ext = ".dylib";
 	env = r_str_append (env, ":/lib:/usr/lib:/usr/local/lib");
 #elif __UNIX__
 	char *env = r_sys_getenv ("LD_LIBRARY_PATH");
-	const char *ext = ".so";
 	env = r_str_append (env, ":/lib:/usr/lib:/usr/local/lib");
 #else
-	char *env = strdup (".:../../../../../../../windows/system32");
-	const char *ext = ".dll";
+	char *env = strdup (".:../../../../../../../../../../../../windows/system32");
```
xarkes: Wtf?
Use `C;\\Windows\\System32` no?

XVilka: No, the path can be different. It is better to use the API to get the address: https://stackoverflow.com/questions/3544185/how-to-get-the-full-path-to-system32-directory-with-c

radare:
Yes this is one of many ugly thjngs in the windows code that i have observed. Im just keeping it as is because “it works” but its clearly crap. So i would appreciate if anybody caring about this operating system to have a look at it

mrexodia:
I looked at Windows, basically it's *a bit* more complicated that searching the `.` (which is actually wrong on Windows, should be the application path which is different than `.`) and system32. If anyone is interested you can find a full list at: https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order

Funny thing:

> The 16-bit system directory. There is no function that obtains the path of this directory, but it is searched.

libr/util/lib.c
XVilka: @GustavoLCR @pelijah can you please add and test a proper way to get the full path here?

PR #14778, opened 2019-08-09
Fix `~??` command (grep help)

bacher09:
Docs and internal help claims that grep help available via `~??` command, but in fact it's available via `~?`. This change makes grep help available via both `~?` (for backward compatibility) and `~??`.

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -3020,7 +3020,10 @@ next2:
 escape_backtick:
 	// TODO must honor " and `
 
-	if (r_str_endswith (cmd, "~?") && cmd[2] == '\0') {
+    // match ~? and ~?? help
+	if (strncmp(cmd, "~?", 2) == 0 &&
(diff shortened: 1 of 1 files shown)


libr/core/cmd.c
```diff
@@ -3020,7 +3020,10 @@ static int r_core_cmd_subst_i(RCore *core, char *cmd, char *colon, bool *tmpseek
 escape_backtick:
 	// TODO must honor " and `
 
-	if (r_str_endswith (cmd, "~?") && cmd[2] == '\0') {
+    // match ~? and ~?? help
+	if (strncmp(cmd, "~?", 2) == 0 &&
```
radare: Thats a pretty overenginnered way to say /~?[?]/. i would say:

```
const *ask = r_str_endswith(cmd, "?") 
if (ask && (ask > cmd && ask[-1] =='~') || (ask+1>cmd && ask[-2]=='~')
```

Issue #1924, opened 2014-12-30
segfault on define (Visual mode)


DaniloNC:
Here is the asciinema http://asciinema.org/a/15086

To reproduce:
- Enter visual mode
- "d" to enter on define
- Press the "Down Arrow"

PR #14789, opened 2019-08-11
Consider endiannes when performing block write operations

cpey:
Consider the architecture endianness when performing the bytewise block operations. 

The provided ___endian_swap_ function may be used to provide a representation of the whole memory buffer with the correct endianness for each word as it would be when loaded from memory. In practice the byte order change that results from reading and writing back to memory has the same effect as changing the byte order of the literal used in the operation. Therefore I just change the byte order for the literal in the PR.

### DIFF
--- a/libr/core/cio.c
+++ b/libr/core/cio.c
@@ -106,35 +106,36 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int 
 	return true;
 }
 
-static inline void __endian_swap(ut8 *buf, ut32 blocksize, ut8 len) {
+static int __endian_swap(ut8 *buf, ut32 blocksize, ut8 len) {
 	ut32 i;
 	ut8 tmp;
+	ut16 v16;
+	ut32 v32;
+	ut64 v64;
 	if (len != 8 && len != 4 && len != 2 && len != 1) {
 		eprintf ("Invalid word size. Use 1, 2, 4 or 8\n");
-		return;
+		return false;
 	}
 	if (len == 1) {
-		return;
+		return true;
 	}
-	for (i=0; i<blocksize; i+=len) {
+	for (i = 0; i < blocksize; i += len) {
 		switch (len) {
-			case 8:
-				tmp = buf[i];	buf[i]	 = buf[i+7]; buf[i+7] = tmp;
-				tmp = buf[i+1]; buf[i+1] = buf[i+6]; buf[i+6] = tmp;
-				tmp = buf[i+2]; buf[i+2] = buf[i+5]; buf[i+5] = tmp;
-				tmp = buf[i+3]; buf[i+3] = buf[i+4]; buf[i+4] = tmp;
-				break;
-			case 4:
-				tmp = buf[i];	buf[i]	 = buf[i+3]; buf[i+3] = tmp;
-				tmp = buf[i+1]; buf[i+1] = buf[i+2]; buf[i+2] = tmp;
-				break;
-			case 2:
-				tmp = buf[i];	buf[i]	 = buf[i+1]; buf[i+1] = tmp;
-				break;
-			default:
-				return;
+		case 8:
+			v64 = r_read_be64(buf+i);
+			r_write_le64(buf+i, v64);
+			break;
+		case 4:
+			v32 = r_read_be32(buf+i);
+			r_write_le32(buf+i, v32);
+			break;
+		case 2:
+			v16 = r_read_be16(buf+i);
+			r_write_le16(buf+i, v16);
+			break;
 		}
 	}
+	return true;
 }
 
 R_API int r_core_write_op(RCore *core, const char *arg, char op) {
@@ -260,7 +261,9 @@ R_API int r_core_write_op(RCore *core, const char *arg, char op) {
 	} else {
 		bool be = r_config_get_i (core->config, "cfg.bigendian");
 		if (!be) {
-			__endian_swap (str, len, len);
+			if (!__endian_swap (str, len, len)) {
+				goto beach;
+			}
 		}
 		for (i=j=0; i<core->blocksize; i++) {
 			switch (op) {
@@ -279,9 +282,6 @@ R_API int r_core_write_op(RCore *core, const char *arg, char op) {
 			if (j >= len) {
 				j = 0; /* cyclic key */
 			}
-		}
-		if (!be) {
-			__endian_swap (str, len, len);
 		}
 	}
 
--- a/libr/core/cio.c
+++ b/libr/core/cio.c
@@ -106,6 +106,18 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int 
 	return true;
 }
 
+static inline void __endian_swap(ut8 *buf, ut32 blocksize, ut64 len) {
+	int i, j;
+	ut8 tmp;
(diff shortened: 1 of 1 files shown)


XVilka:
Please add a test case for this vital feature in radare2-regressions: https://github.com/radare/radare2-regressions

libr/core/cio.c
```diff
@@ -106,6 +106,18 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
 	return true;
 }
 
+static inline void __endian_swap(ut8 *buf, ut32 blocksize, ut64 len) {
```
XVilka: I am also this one is a bit inefficient.

radare:
Im happy about the feature. But not the impl which looks underperformant and misses some spaces before ( on function calls. Can you improve that and add some tests?

cpey:
Now it is 40% faster at the expense of constraining the length of the word. But it seems reasonable.

Also as I mentioned the change in the block write operation is not parsing the whole buffer but only the literal and consequently the performance thing was not affecting.

I will add the tests if it looks good to you @XVilka @radare

libr/core/cio.c
```diff
@@ -106,6 +106,37 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
...
 }
 
+static inline void __endian_swap(ut8 *buf, ut32 blocksize, ut8 len) {
+	ut32 i;
+	ut8 tmp;
+	if (len != 8 && len != 4 && len != 2 && len != 1) {
+		eprintf ("Invalid word size. Use 1, 2, 4 or 8\n");
+		return;
+	}
+	if (len == 1) {
```
radare: Unnecessary check.

libr/core/cio.c
```diff
@@ -106,6 +106,37 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
...
+		return;
+	}
+	for (i=0; i<blocksize; i+=len) {
```
radare: Missing spces around operators

libr/core/cio.c
```diff
@@ -106,6 +106,37 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
...
+		switch (len) {
+			case 8:
```
radare: Indent goes at the same level, switch and case

libr/core/cio.c
```diff
@@ -106,6 +106,37 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
...
+				tmp = buf[i];	buf[i]	 = buf[i+7]; buf[i+7] = tmp;
```
radare: Use r_read/write apis that already do that

libr/core/cio.c
```diff
@@ -106,6 +106,37 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
...
+				tmp = buf[i+1]; buf[i+1] = buf[i+6]; buf[i+6] = tmp;
+				tmp = buf[i+2]; buf[i+2] = buf[i+5]; buf[i+5] = tmp;
+				tmp = buf[i+3]; buf[i+3] = buf[i+4]; buf[i+4] = tmp;
+				break;
+			case 4:
+				tmp = buf[i];	buf[i]	 = buf[i+3]; buf[i+3] = tmp;
+				tmp = buf[i+1]; buf[i+1] = buf[i+2]; buf[i+2] = tmp;
+				break;
+			case 2:
+				tmp = buf[i];	buf[i]	 = buf[i+1]; buf[i+1] = tmp;
```
radare: Dont abuse multiple statments in one line

libr/core/cio.c
```diff
@@ -106,6 +106,37 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
...
+				break;
+			default:
+				return;
```
radare: Unnecessary case

libr/core/cio.c
```diff
@@ -106,35 +106,36 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
 	return true;
 }
 
-static inline void __endian_swap(ut8 *buf, ut32 blocksize, ut8 len) {
+static int __endian_swap(ut8 *buf, ut32 blocksize, ut8 len) {
```
cpey: Notice I modified the return type to notify the caller when the parameters are not valid. Then accordingly in line [264](https://github.com/radare/radare2/pull/14789/commits/e5f26cf914307a615fefa410e3cf9c08fa2b291b#diff-ca04423ab87af7e69e9adc8535309fe7R264) it does not process in case of error.

libr/core/cio.c
```diff
@@ -106,6 +106,38 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
...
+		case 8:
+			v64 = r_read_be64(buf+i);
+			r_write_le64(buf+i, v64);
+			break;
+		case 4:
+			v32 = r_read_be32(buf+i);
+			r_write_le32(buf+i, v32);
+			break;
+		case 2:
+			v16 = r_read_be16(buf+i);
```
radare: Missing spaces before ( here and there

libr/core/cio.c
```diff
@@ -106,6 +106,38 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
...
+		eprintf ("Invalid word size. Use 1, 2, 4 or 8\n");
+		return false;
+	}
+	if (len == 1) {
+		return true;
+	}
+	for (i = 0; i < blocksize; i += len) {
+		switch (len) {
+		case 8:
+			v64 = r_read_be64(buf+i);
```
radare: Also theres a r read function that takes a size arg and a endian arg that does exactly this iirc

libr/core/cio.c
```diff
@@ -106,6 +106,38 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
...
+			r_write_le64(buf+i, v64);
+			break;
+		case 4:
+			v32 = r_read_be32(buf+i);
+			r_write_le32(buf+i, v32);
```
radare: Spaces around +

libr/core/cio.c
radare: This code can do oobreada because the lengh of the buffer doesnt needs to be aligned to the word size

libr/core/cio.c
```diff
@@ -227,6 +259,12 @@ R_API int r_core_write_op(RCore *core, const char *arg, char op) {
 			}
 		}
 	} else {
+		bool be = r_config_get_i (core->config, "cfg.bigendian");
+		if (!be) {
+			if (!__endian_swap (str, len, len)) {
+				goto beach;
+			}
+		}
 		for (i=j=0; i<core->blocksize; i++) {
```
radare: Spces around operators

libr/core/cio.c
radare: Return bool, not int

libr/core/cio.c
cpey: I will double check that. It was not a problem yesterday.

cpey:
Added tests in PR [#1924](https://github.com/radare/radare2-regressions/pull/1924)

radare:
Thanks!

PR #14792, opened 2019-08-12
Improve saving layouts function

Vane11ope:


radare:
Wait to merge until i can test

libr/core/panels.c
```diff
@@ -5596,6 +5603,11 @@ bool r_load_panels_layout(RCore *core, const char *_name) {
 		}
 		p_cfg += strlen (p_cfg) + 1;
 	}
+	if (!found) {
+		char *tmp = r_str_newf ("No saved layout found for the name: %s", _name);
```
radare: Add nullchk

libr/core/panels.c
```diff
@@ -5494,7 +5501,7 @@ void r_save_panels_layout(RCore *core, const char *_name) {
 		pj_kn (pj, "h", panel->view->pos.h);
 		pj_end (pj);
 	}
-	FILE *file = fopen (config_path, "w");
+	FILE *file = fopen (config_path, "a");
```
radare: Use sandbox apis. This is a security flaw. Also, better specify “b” to avoid windows/unix newline issues

radare: also, if you save multiple times the file is always growing and it just becomes unhandlable

radare: but  i understand the "bug" you are reporting here.. which implies not being able to save more than one layout. merging, but this needs a cleaner solution, otherwise autosaving layout results in HELL

Issue #14701, opened 2019-07-30
Forbid backticks in flagnames


radare:
as stated by the is* issue

radare:
they are forbidden already..
```
$ r2 -
[0x00000000]> "f je`bar=123"
[0x00000000]> f
0x0000007b 1 je_bar
[0x00000000]> f*
fs *
f je_bar 1 0x0000007b
[0x00000000]> q
$
```
i think the problem iss when accessing the struct fields properly

thestr4ng3r:
@radare is this about `is*` only or also anywhere else?

thestr4ng3r:
So can this be closed?

thestr4ng3r:
see https://github.com/radare/radare2/pull/14797#issuecomment-520813533

radare:
Why you didnt fixed r_name_filter?

thestr4ng3r:
Because it already handles ` correctly, as wrote.

radare:
this is fixed now

PR #14797, opened 2019-08-13
Refactor is* and fix Injection harder

thestr4ng3r:
After https://github.com/radare/radare2/pull/14690 it is still possible to do injection through is* using a symbol name containing a `"`.
For example: [ls.gz](https://github.com/radare/radare2/files/3496619/ls.gz) execute `.is*` or `ood` on this bin, cc @ps1337 

This patch makes the names in is* more consistent with the actual flag names and also fixes a leak.

The diff looks a bit wild because of indentation changes, the significant changes are around the `construct_symbol_flagname()` calls.

### DIFF
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -2158,7 +2158,7 @@ static int bin_symbols(RCore *r, int mode, ut64 laddr, int va, ut64 at, const ch
 				if (!flagname) {
 					goto next;
 				}
-				r_cons_printf ("f %s%s%s %u 0x%08" PFMT64x "\n",
+				r_cons_printf ("\"f %s%s%s %u 0x%08" PFMT64x "\"\n",
 					r->bin->prefix ? r->bin->prefix : "", r->bin->prefix ? "." : "",
 					flagname, symbol->size, addr);
 				free(flagname);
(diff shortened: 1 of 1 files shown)


ps1337:
Nice catch, thanks for the update! Does this also fix #14701 ? I can't check it currently as I'm on mobile :D

@blenk92 and me didn't get around yet to fix #14701 so maybe this other issue can be closed too. 

Cheers 





-----Original Message-----

thestr4ng3r:
The issue description is kind of unclear. If it's only about removing ` from the is* output, then that is fixed:
Bildschirmfoto vom 2019-08-13 14-25-46

ps1337:
Ah yes, we wanted to fix this issue and saw it already being addressed as backticks and flag names were already replaced. We then decided to go ahead and expand our original hot fix (add quoting) in order to forbid backticks altogether in symbol names, imports and relocs. We didn't file the PR yet for further testing because we found that certain PE executables weren't handled correctly as backticks were still there. 

LGTM for your PR - After your PR gets merged @blenk92 and me can do further testing with manipulated PE files (and malicious imports/relocs) and see if they are still an issue then 

Cheers 


-----Original Message-----

radare:
Another reason for using the quote is because that skips a bunch or parsing code and make projects and other is* scripts load faster

thestr4ng3r:
Ok, so should we keep the quotes? In that case the r2r pr can be dropped.

PR #14815, opened 2019-08-16
Avoid disassembling the same instruction twice on rop search

GustavoLCR:
This improves performance a bit, but for some archs r_anal_op doesn't provide mnemonics, so we have to call r_asm_disassemble

### DIFF
--- a/libr/core/cmd_search.c
+++ b/libr/core/cmd_search.c
@@ -1126,6 +1126,7 @@ static RList *construct_rop_gadget(RCore *core, ut64 addr, ut8 *buf, int buflen,
 			goto ret;
 		}
 		free (opst);
+		aop.mnemonic = NULL;
 		nb_instr++;
 	}
 ret:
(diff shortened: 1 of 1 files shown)


libr/core/cmd_search.c
```diff
@@ -1060,19 +1061,23 @@ static RList *construct_rop_gadget(RCore *core, ut64 addr, ut8 *buf, int buflen,
...
 			goto ret;
 		}
-		const int opsz = asmop.size;
+		const int opsz = aop.size;
 		// opsz = r_strbuf_length (asmop.buf);
-		char *opst = r_asm_op_get_asm (&asmop);
+		char *opst = aop.mnemonic;
+		if (!opst) {
+			r_asm_set_pc (core->assembler, addr);
+			if (!r_asm_disassemble (core->assembler, &asmop, buf + idx, buflen - idx)) {
```
radare: Keep asmop local here

radare: Can we have a warning once for those analop archs that dont provide/handle disasm instead? This is a bug we should handle in the plugins not as a workaround here. Soon or late rasm api should be using anal plugs

radare:
Can u provide some numbers about how much fast is now?

GustavoLCR:
Before:

before

After:

after

radare:
Wow nice one :)

PR #14822, opened 2019-08-16
fix command injection even harder

ps1337:
Hey there.

This hopefully completely fixes the command injection issues of CVE-2019-14745 since bypasses (e.g. #14797) to the hotfix were possible (as expected :D). All outputs of the `*` commands are filtered using `r_name_filter` regarding:

- symbols
- imports
- relocations
- sections
- libs
- all demangled things

Let me know what you think or if any filtering can be moved to a better place.

@blenk92 will come up with some more proactive mitigations over the weekend.

Cheers.

### DIFF
--- a/libr/bin/p/bin_symbols.c
+++ b/libr/bin/p/bin_symbols.c
@@ -346,6 +346,7 @@ static RBinSymbol *newSymbol(RBinString *s, ut64 addr, ut64 size) {
 		sym->type = r_str_const (R_BIN_TYPE_FUNC_STR);
 		sym->bind = r_str_const ("NONE");
 	}
+	r_name_filter(sym->name, -1);
 	return sym;
 }
 
@@ -379,6 +380,7 @@ static RList *parseSections(RBuffer *b, int x, int n_sections, RList *strings) {
 		}
 		RBinString *name = strings? r_list_get_n (strings, i): NULL;
 		const char *namestr = name? name->string: "";
+		r_name_filter(namestr, -1);
 		ut32 A = r_read_le32 (buf + off);
 		ut32 B = r_read_le32 (buf + off + 4);
 		//ut32 C = r_read_le32 (buf + off + 8);
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -1300,6 +1300,7 @@ static char *get_reloc_name(RCore *r, RBinReloc *reloc, ut64 addr) {
 		// TODO(eddyb) implement constant relocs.
 	}
 	free (demangled_name);
+	r_name_filter(reloc_name, -1);
 	return reloc_name;
 }
 
@@ -1366,6 +1367,7 @@ static void set_bin_relocs(RCore *r, RBinReloc *reloc, ut64 addr, Sdb **db, char
 			r_meta_add (r->anal, R_META_TYPE_DATA, reloc->vaddr, reloc->vaddr+4, NULL);
 		}
 		reloc_name = reloc->import->name;
+		r_name_filter(reloc_name, -1);
 		if (r->bin->prefix) {
 			snprintf (str, R_FLAG_NAME_SIZE, "%s.reloc.%s", r->bin->prefix, reloc_name);
 		} else {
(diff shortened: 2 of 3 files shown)


ps1337:
As some tests are failing, should we use a different filter function that doesn't force the results to being strictly alphabetical?

For example:
```
-007 0x000002f4 0x000002f4 GLOBAL   FUNC   20 LWorld.method.<init>()V
```
becomes
```
+007 0x000002f4 0x000002f4 GLOBAL   FUNC   20 LWorld.method._init___V
```

The downside may however be that some chars that allow command injection may be missed

radare:
The problem of filtering chars is that it is possible to get two symbols with different names sharing the same flagname, this is suposed to be fixed in rbin at some point.

Quoting everythingseems like a more drastic solution, and this can be done in cmd intepret instead of doing it at cons level. Which imho is not the place to go.

Also, this pr doesnt follow the coding standards of r2 ( missing spaces, indentation , ..)

radare:
any reproducer?

libr/bin/p/bin_symbols.c
```diff
@@ -346,6 +346,7 @@ static RBinSymbol *newSymbol(RBinString *s, ut64 addr, ut64 size) {
 		sym->type = r_str_const (R_BIN_TYPE_FUNC_STR);
 		sym->bind = r_str_const ("NONE");
 	}
+	r_name_filter(sym->name, -1);
```
radare: missing space

libr/core/cbin.c
```diff
@@ -1300,6 +1300,7 @@ static char *get_reloc_name(RCore *r, RBinReloc *reloc, ut64 addr) {
 		// TODO(eddyb) implement constant relocs.
 	}
 	free (demangled_name);
+	r_name_filter(reloc_name, -1);
```
radare: missing space

radare: also this namefilter should be happening before printing, its fine if the reloc name contains special chars here

libr/core/cbin.c
```diff
@@ -1366,6 +1367,7 @@ static void set_bin_relocs(RCore *r, RBinReloc *reloc, ut64 addr, Sdb **db, char
 			r_meta_add (r->anal, R_META_TYPE_DATA, reloc->vaddr, reloc->vaddr+4, NULL);
 		}
 		reloc_name = reloc->import->name;
+		r_name_filter(reloc_name, -1);
```
radare: missing space and same reasoning

libr/core/cbin.c
```diff
@@ -3081,6 +3087,7 @@ static int bin_libs(RCore *r, int mode) {
 			// Nothing to set.
 			// TODO: load libraries with iomaps?
 		} else if (IS_MODE_RAD (mode)) {
+			r_name_filter(lib, -1);
```
radare: this one is fine, but missing space before (

radare:
hello? any reproducer?

ps1337:
I'll provide one tomorrow, sorry for the delay

radare:
tomorrow? the release was suposed to be today...

ps1337:
I thought this PR wasn't a fitting solution anyway because of the comment above ("The problem of filtering chars is that it is possible to get two symbols with different names sharing the same flagname"). It seems that merging this could break some stuff, doesn't it? That's why I didn't fix the indentation yet 



-----Original Message-----

radare:
yes, that's why it's not merged, but having a reproducer would let me craft a fix

ps1337:
Oh and there's a PoC attached to this PR: https://github.com/radare/radare2/pull/14827



-----Original Message-----

ps1337:
this one executes a sleep after `ood`. Also check out:

```
[0x14007e3c4]> is*~sleep
"f sym.imp.KERNEL32.dll_Unhan__sleep_10_onFilter 0 0x1400c6648"
k bin/pe/imp.KERNEL32/127=Unhan`!sleep 10`onFilter
```

[notputty.zip](https://github.com/radare/radare2/files/3563908/notputty.zip)



hope that helps

radare:
thanks! i will release on monday, and work on a proper fix, ive seen there are several spots that this can fail, so i will try to cover them all in cbin.c

radare:
we may probably want the bin to execute `exit 1`. so it will be catched by the testsuite :D but ive commited the bin in r2r and pushed a pr that may fix this specific case

ps1337:
- `exit1.exe` executes `exit 1`
- `killr2.exe` executes `killalll r2`

this makes it easier to detect the execution

[PoC.zip](https://github.com/radare/radare2/files/3564593/PoC.zip)

radare:
i have created this pr which is safer because it doesnt modify the original string. only when printing the command. there's an issue to support base64 shellwide, so we may be able to spit out base64 arguments, for now i think this is fine

https://github.com/radare/radare2/pull/14952

PR #14827, opened 2019-08-17
Add auto-quoting and harden .<..> operator

blenk92:
Hey there,

as @ps1337 already mentioned in https://github.com/radare/radare2/pull/14822 we came up with another mitigation strategy:
While replacing special characters and quoting, if done correctly, seem sufficient to prevent an attacker from injecting r2 commands, it seems to be a hard task. The hot-fix and similar commits for CVE-2019-14745 showed that this approach is error-prone because of the various file formats supported by radare2. That is, we propose to additionally add automatic quoting to all (or at least i<..>) commands that will be interpreted using the . operator. In addition we propose that if commands are quoted the . operator does only accept one command per line. That is, even if an attacker can break out of the quoting injected commands will be ignored. This PR implements this strategy.

The PR by itself already mitigates at least certain exploits of the CVE (such as [kaputty.zip](https://github.com/radare/radare2/files/3512169/kaputty.zip)).

### DIFF
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -1134,13 +1134,55 @@ club:
 	va_end (ap3);
 }
 
+char* auto_quote (const char* str) {
+	char* str_copy = strdup(str);
+	if (!str_copy) {
+		return 0;
+	}
+
+	int last = strlen (str_copy)-1;
+	if (last == -1) {
+		return str_copy;
+	}
+
+	char line_ending = '\0';
+	if (str_copy[last] == '\n') {
+		str_copy[last] = '\0';
+		last--;
+		line_ending = '\n';
+	}
+
+	bool needs_quoting;
+	if (last == -1) {
+		needs_quoting = false;
+	} else {
+		needs_quoting = (str[0] != '\"' && str[0] != '\0' && str[last] != '\"');
+	}
+
+	char *tmp = r_str_replace(str_copy, "\n", "\"\n\"", true);
+	if (!needs_quoting) {
+		str_copy = r_str_newf("%s%c", tmp, line_ending);
+	} else {
+		str_copy = r_str_newf("\"%s\"%c", tmp, line_ending);
+	}
+	free(tmp);
+	return str_copy;
+}
+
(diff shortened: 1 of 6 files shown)


radare:
do we have another reproducer for this issue? quoting all the  lines is not correct because it  also blocks  the  @ sign. as well as other special chars that will break  many scripts. so this is   not a good solution imho

blenk92:
> do we have another reproducer for this issue? quoting all the lines is not correct because it also blocks the @ sign. as well as other special chars that will break many scripts. so this is not a good solution imho
hmm, we didn't think about that. In that case this is obviously not the way to go...
What do you think about allowing only one command per line? To me, this also seems like a valuable mitigation. On the other hand, this might also break scripts...

radare:
We can enable sandbox temporarily too. To forbid escsping to the shell or opening files from disk. This can also break normal scripts but at least it will be cleaner than what it is now. But the proper fix would be to properly filter all those rabin2 -r use cases

radare:
ping?

blenk92:
> We can enable sandbox temporarily too. To forbid escsping to the shell or opening files from disk. This can also break normal scripts but at least it will be cleaner than what it is now. But the proper fix would be to properly filter all those rabin2 -r use cases
view it on GitHub, or mute the thread.

Well, does the sandbox also prevent from executing `.:<port>" command (or to execute shell commands within that)? And still, because the issue occurs within `ood` it might still be possible to inject shell code and jump to it to execute (sure it this will be much harder, but anyway)... please correct me if I'm wrong

So from my point of view the goal should be to prevent escaping, not only the execution of shell commands....

ps1337:
+1, since using r_name_filter might cause problems as described in #14822 which might also not be the best solution ever. Another point is that we have to prevent executing shell commands and r2 commands together, since `.:8080` can for example be used to open a port for further exploitation.

I'm currently not sure what the best solution might be at the moment. Filtering symbol, import, etc values was my initial idea but as described this could cause name collisions. We could use a filtering function that only removes `!`, `:` and backticks but we might still miss some things :D

PR #14829, opened 2019-08-18
Use op->disp instead of op->ptr for disp of x86 MOV (and others)

kazarmy:
Following #14773 (more specifically https://github.com/radare/radare2/pull/14773/files#diff-7f6f82199b572cdbe767aa7a1ba9b436L2391).


radare:
looks red

kazarmy:
somewhat better now but still red

kazarmy:
it's green now, ok to merge?

PR #14833, opened 2019-08-18
Fix #14816 - Properly get lib path on Windows

### LINKED ISSUE #14816: Get root path properly in `libr/util/lib.c` (and other places) for Windows
See discussion here: https://github.com/radare/radare2/pull/14777

Basically use some API instead of:
```c
char *env = strdup (".:../../../../../../../../../../../../windows/system32");
```

cc @GustavoLCR @pelijah

GustavoLCR:


XVilka:
Afaik tchar was dropped for reason to support only unicode

Issue #14755, opened 2019-08-04
Search not working on block-device


fandauchytil:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | QEMU-KVM ; Debian GNU/Linux 9.9 (stretch); 4.9.168-1+deb9u4 (2019-07-19) x86_64 GNU/Linux
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86/64
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.7.0 22586 @ linux-x86-64 git.3.7.0 commit: cb2648cbbde3cd63cb1b2a9b1d4269c97bd653b8 build: 2019-08-03__15:16:58

### Expected behavior

Working 'search' on block-device.

### Actual behavior

'Search' on block-device does not work at all. It either matches nothing or it matches "everything".

### Steps to reproduce the behavior

```
# blockdev --getsize64 /dev/vdb1
1063256064
# r2 /dev/vdb1
```

1. No "r2" virtual seek (= I'm allways on virtual offset 0):

```
[0x00000000]>  b
0x100
[0x00000000]> px
- offset -   0 1  2 3  4 5  6 7  8 9  A B  C D  E F  0123456789ABCDEF
0x00000000  0000 0000 0000 0000 0000 0000 0000 0000  ................
...
0x000000f0  0000 0000 0000 0000 0000 0000 0000 0000  ................
[0x00000000]> px 32 @ 16k
- offset -   0 1  2 3  4 5  6 7  8 9  A B  C D  E F  0123456789ABCDEF
0x00004000  0101 0000 0000 0000 0000 0001 0000 0024  ...............$
0x00004010  0000 0020 0000 0007 7665 7273 696f 6e00  ... ....version.
[0x00000000]> / version
Searching 7 bytes in [0x0-0x3f600000]
hits: 0
[0x00000000]> /x 7665
Searching 2 bytes in [0x0-0x3f600000]
hits: 0
[0x00000000]> / \x76
Searching 1 byte in [0x0-0x3f600000]
hits: 0
[0x00000000]> /!x 00
Searching 1 byte in [0x0-0x3f600000]
hits: 0
[0x00000000]> o
 3 * r-x 0x3f600000 /dev/vdb1
[0x00000000]> om
 1 fd: 3 +0x00000000 0x00000000 - 0x3f5fffff r-x 
```

[...]

PR #14848, opened 2019-08-19
Fix read-only raw search (#14755)

fandauchytil:


radare:
Good fix!

fandauchytil:
Thank you :). It is my first pull request ever, did I missed something?

radare:
Nope! Everything looks good! Thanks for contributing! Hopefully that will not be the last one :)

PR #14851, opened 2019-08-19
[WIP] Fix issue with cmp & test in x86 ESIL

### LINKED ISSUE #14249: ESIL for x86 test sets SF incorrectly for non-64 bit
Example:
```
e asm.arch=x86
e asm.bits=64
s 0
aei
aepc 0
wa test eax, eax
aer eax=-1
aes
aer all~sf
```
gives:
```
sf = 0x00000000
```
which is wrong. It does work however when using rax instead.

This is the esil for `test eax, eax`:
```
0,eax,eax,&,==,$z,zf,:=,$p,pf,:=,$s,sf,:=,0,cf,:=,0,of,:=
```
s eax is &-ed with itself and the result is compared against 0 to get the flags. The issue is that the `==` op at this point has no idea about the bitness and falls back to 64:
```c
// This function also sets internal vars which is used in flag calculations.
static int esil_cmp(RAnalEsil *esil) {
	ut64 num, num2;
	int ret = 0;
	char *dst = r_anal_esil_pop (esil);
	char *src = r_anal_esil_pop (esil);
	if (dst && r_anal_esil_get_parm (esil, dst, &num)) {
		if (src && r_anal_esil_get_parm (esil, src, &num2)) {
			esil->old = num;
			esil->cur = num - num2;
			ret = 1;
			if (r_reg_get (esil->anal->reg, dst, -1)) {
				esil->lastsz = esil_internal_sizeof_reg (esil, dst);
			} else if (r_reg_get (esil->anal->reg, src, -1)) {
				esil->lastsz = esil_internal_sizeof_reg (esil, src);
			} else {
				// default size is set to 64 as internally operands are ut64
				esil->lastsz = 64;
			}
		}
	}
	free (dst);
	free (src);
	return ret;
}
```
Imo, the fallback should be `esil->anal->bits` instead, but that won't solve `test eax, eax` in x86_64.

So to fix `test`, either one could pull out the sign bit manually from the result of the & and store it in sf or make == aware about the bitness, for example by adding ops like `==8` (lol), `==16`, `==32` and `==64`. I'd prefer to go with the second option, because if esil has a notion of signedness, it must also be explicit about how many bits it uses. This might also count some other ops apart from `==`.

bacher09:
This PR fixes #14249 by adding new operators to ESIl (`==4`, `==2` and `==1`), which might be not best solution. Maybe radare2 community can provide some feedback regarding adding new ESIL operators vs adding some bitness information to data.

Please don't merge this PR at this moment, cause at least ESIL docs should be extended with new operators.

### DIFF
--- a/libr/anal/esil.c
+++ b/libr/anal/esil.c
@@ -921,7 +921,7 @@ static bool esil_interrupt(RAnalEsil *esil) {
 }
 
 // This function also sets internal vars which is used in flag calculations.
-static bool esil_cmp(RAnalEsil *esil) {
+static bool esil_cmp(RAnalEsil *esil, int bits) {
 	ut64 num, num2;
 	bool ret = false;
 	char *dst = r_anal_esil_pop (esil);
@@ -937,13 +937,29 @@ static bool esil_cmp(RAnalEsil *esil) {
 				esil->lastsz = esil_internal_sizeof_reg (esil, src);
 			} else {
 				// default size is set to 64 as internally operands are ut64
-				esil->lastsz = 64;
+				esil->lastsz = bits;
 			}
 		}
 	}
 	free (dst);
 	free (src);
 	return ret;
+}
+
+static bool esil_cmp8(RAnalEsil *esil) {
+    return esil_cmp(esil, 64);
+}
+
+static bool esil_cmp4(RAnalEsil *esil) {
+    return esil_cmp(esil, 32);
+}
+
+static bool esil_cmp2(RAnalEsil *esil) {
+    return esil_cmp(esil, 16);
+}
+
+static bool esil_cmp1(RAnalEsil *esil) {
+    return esil_cmp(esil, 8);
 }
 
 #if 0
@@ -3102,7 +3118,10 @@ static void r_anal_esil_setup_ops(RAnalEsil *esil) {
 	OP ("$js", esil_js, 1, 0, OT_UNK);
 	OP ("$r", esil_rs, 1, 0, OT_UNK);
 	OP ("$$", esil_address, 1, 0, OT_UNK);
-	OP ("==", esil_cmp, 0, 2, OT_MATH);
+	OP ("==", esil_cmp8, 0, 2, OT_MATH);
+	OP ("==4", esil_cmp4, 0, 2, OT_MATH);
+	OP ("==2", esil_cmp2, 0, 2, OT_MATH);
+	OP ("==1", esil_cmp2, 0, 2, OT_MATH);
 	OP ("<", esil_smaller, 1, 2, OT_MATH);
 	OP (">", esil_bigger, 1, 2, OT_MATH);
 	OP ("<=", esil_smaller_equal, 1, 2, OT_MATH);
--- a/libr/anal/p/anal_x86_cs.c
+++ b/libr/anal/p/anal_x86_cs.c
@@ -908,17 +908,35 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 	case X86_INS_CMPSB:
 	case X86_INS_CMPSS:
 	case X86_INS_TEST:
+    {
+        const char *opname;
+        switch (INSOP(0).size) {
+            case 1:
+                opname="==1";
+                break;
+            case 2:
+                opname="==2";
+                break;
+            case 4:
+                opname="==4";
+                break;
+            case 8:
+            default:
+                opname="==";
+                break;
+        }
 		if (insn->id == X86_INS_TEST) {
 			src = getarg (&gop, 1, 0, NULL, SRC_AR);
 			dst = getarg (&gop, 0, 0, NULL, DST_AR);
-			esilprintf (op, "0,%s,%s,&,==,$z,zf,:=,$p,pf,:=,$s,sf,:=,0,cf,:=,0,of,:=",
-				src, dst);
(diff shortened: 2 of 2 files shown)


radare:
Im not a big fan of adding new esil commands. @condret any idea here?

bacher09:
Another possible approach would be adding bitnes information to numbers, but this will require changes to numbers parser (which IMO already enough complicated).

radare:
Yeah i understand the issue but if the birness involves a register it must be taken from rreg. Right now evrrything is ut64. So no possible casts here. So i think you should just use x,0xffff,&,y,0xffff,==

bacher09:
Unfortunately this (`x,0xffff,&,y,0xffff,==`) will not work.

Issue arises when constant is compared against memory address (`cmp dword [rsp + 4], 0` in my [example](https://github.com/radare/radare2/issues/14249#issuecomment-522268654)). ESIL representation of this will look like this:

`0,0x4,rsp,+,[4],==,$z,zf,:=,32,$b,cf,:=,$p,pf,:=,$s,sf,:=,$o,of,:=`

So, basically after some operations is done it will look like this: `0,0xffffffff,==,$s` (and other side effects).
And masking bits will not help, unless there will some way express bitnets in result that calculated after `0,0xffffffff,&`.

condret:
`$o` can be replaced by `1,<lastsize>,-,$c,2,<lastsize>,-,$c,^`, so if we compared 2 32-bit "constants" we could do `31,$c,30,$c,of,:=`.

it's a bit more difficult for `$s`, but still doable:
`1,31,0,0x4,rsp,+,[4],-,>>,&,sf,:=`

the whole expression then could look something like this:
`0,0x4,rsp,+,[4],==,$z,zf,:=,32,$b,cf,:=,$p,pf,:=,1,31,0,0x4,rsp,+,[4],-,>>,&,sf,:=,31,$c,30,$c,^,of,:=`

meanwhile the substitute for $o looks okish, the substitute for $s is problematic as it performs a redundant memread.
....
We already had a similar situation with $c and $b at the very first version of esil. Back then I changed $c to $cx (x is some number specifying the bit carried from). Later I changed that to x,$c. I think we should do the same for $s and $o now:

`0,0x4,rsp,+,[4],==,$z,zf,:=,32,$b,cf,:=,$p,pf,:=,32,$s,sf,:=,32,$o,of,:=`

condret:
protip: you can override esil-operation in the plugin, so that you don't have to run through all the other plugins and fix their esil-expressions (this is pain, been there)

radare:
Will be good to decide something here.. imho i like the simplest syntax, but i understand for the very few use cases of this we probably dont want to extend the esil standard with more operations. any middle point?

bacher09:
I think  replacing `$s` or `$o` with some bit expressions would break consistency, cause in some cases they shouldn't be replaced (for example when `CMP` has register argument) while in some other cases they should be replaced (when there is no register argument). Also it will make ESIL representation much more complicated. Although correctness is much more important then consistency.

>  Later I changed that to x,$c. I think we should do the same for $s and $o now:

IMO this is good solution, but this change isn't backward compatible so I don't know.
But if backward compatibility isn't big issue, then IMHO it's really good option.

radare:
@condret are u ok in adding those new operators? Or maybe one that pops the size?

XVilka:
I think it should be postponed after this release. Any solution will likely break things, lets not rush with this before the release and a conference.

radare:
Agree

libr/anal/esil.c
```diff
@@ -946,6 +946,22 @@ static bool esil_cmp(RAnalEsil *esil) {
 	return ret;
 }
 
+static bool esil_cmp8(RAnalEsil *esil) {
+    return esil_cmp(esil, 64);
+}
+
+static bool esil_cmp4(RAnalEsil *esil) {
+    return esil_cmp(esil, 32);
```
radare: use tabs

libr/anal/esil.c
```diff
@@ -946,6 +946,22 @@ static bool esil_cmp(RAnalEsil *esil) {
 	return ret;
 }
 
+static bool esil_cmp8(RAnalEsil *esil) {
+    return esil_cmp(esil, 64);
```
radare: use tabs, missing space bfore (

libr/anal/p/anal_x86_cs.c
```diff
@@ -908,17 +908,35 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 	case X86_INS_CMPSB:
 	case X86_INS_CMPSS:
 	case X86_INS_TEST:
+    {
+        const char *opname;
+        switch (INSOP(0).size) {
```
radare: use tabs, cases go aligned at the same column as the switch and missing space before (

libr/anal/p/anal_x86_cs.c
```diff
@@ -908,17 +908,35 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
...
+            case 1:
+                opname="==1";
```
radare: spaces around "="

radare:
can you address the comments?

radare:
ping

radare:
i want this merged/solved for next release. can you please answer about addressing the comments? cc @condret if you are happy with that approach we can go forward and merge it

condret:
I do a hotfix for this now.
Cleaner fix for $s and $o is on the way as well

PR #14864, opened 2019-08-21
Skip gadgets starting with a nop in /R ##search

radare:


radare:
Did u tested it? Because i dont. And we have few rop tests.. so maybe this is introducing false positives.. it was just a blind change. Forgot to add the WIP

PR #14867, opened 2019-08-21
Fix hexagon jumps second try

Rot127:
For reference:
The problem was discussed here: https://github.com/radare/radare2/issues/12352
Script for generation of the files can be found here: https://github.com/radareorg/r2hexagon
Tests: https://github.com/radare/radare2-regressions/pull/1931


XVilka:
Thanks!

Rot127:
Thanks for your patience with a newbie!

PR #14877, opened 2019-08-23
Fixes #14856: Changing visual seek behaviour in Vv

### LINKED ISSUE #14856: `Vv` + `g` (seek offset) doesn't work properly
`r2 /bin/ls` then `aaa`, then go into `Vv` mode. It shows us `entry0` function. If we call the `seek to offset` command (`g` key), and type `main`, it doesn't jump to main and shows it. Note, that `main` is not in the visible part of the functions list. If we seek into any function that is visible in the functions list window, e.g. `sym.imp.strrchr`, it will show `>` character before it, but will not show the function body in the corresponding window.

deepakchethan:


libr/core/vmenus.c
```diff
@@ -3361,10 +3361,20 @@ R_API void r_core_visual_anal(RCore *core, const char *input) {
 				}
 			}
 			break;
-		case 'g':
-		        r_core_visual_showcursor (core, true);
-		        r_core_visual_offset (core);
-		        r_core_visual_showcursor (core, false);
+		case 'g': {
+			r_core_visual_showcursor(core, true);
```
radare: missing space before (

libr/core/vmenus.c
```diff
@@ -3361,10 +3361,20 @@ R_API void r_core_visual_anal(RCore *core, const char *input) {
...
+			r_core_visual_offset(core);        // change the seek to selected offset
+			RListIter *iter;		   // change the curren option to selected seek
+			RAnalFunction *fcn;
+			int i = 0;
+			r_list_foreach (core->anal->fcns, iter, fcn) {
+				if (core->offset ==  fcn->addr){
```
radare: missing space ){

libr/core/vmenus.c
deepakchethan: Added

libr/core/vmenus.c
deepakchethan: Added

PR #14884, opened 2019-08-24
Fixes #14845: Use unicode settings in radiff2

### LINKED ISSUE #14845: Use unicode settings (`scr.utf8` and `scr.utf8.curvy`) in radiff2 graphs
If they are set in the config - use them by default.

Also colorize the yellow (different) nodes.

deepakchethan:
After change:

Screen Shot 2019-08-24 at 5 22 50 PM


XVilka:
AppVeyor is red,
Travis is green.
Something is broken,
More commits to be seen.

PR #14892, opened 2019-08-25
Modified pdx to disassemble hexpairs

### LINKED ISSUE #14771: `pdx` to disassembly the byte sequence
Like `tpx` instruction, to allow disassemble the byte sequence instead of writing it or calling rasm2
```
[0x0000000]> pdx 0x90
nop
[0x0000000]>
```
Just an alias to `pd @x:...`

tantei3:
Closes #14771 

Modified the `pdx` command to disassemble the hexpairs by aliasing it to `pd @x:<hexpairs>` and get the output as mentioned in the issue. Took the reference from `tpx` command for implementation.


radare:
i did that already last week fforgot to close the issue :(

but in your pr you are removing my code and redoing it in a different way

tantei3:
image

Currently for the hex size bigger than like 4 bytes, it doesn't display the disassembly as expected, but this PR takes care of moving the logic above the block size check. Also the issue mentioned about aliasing to `pd @x:` so I thought that this change was required!

radare:
Hexpairs dont start with 0x

Syntactically your example is wrong because it must be assuming endianness. And this should work fine if you remove the 0x of the begining

radare:
If broken pls fix and we probably need some tests

tantei3:
image

Well looking at these examples, I don't see any differences even if 0x is included or not (I even tried with this `31c09952682f636174682f62696e89e3526873737764682f2f7061682f65746389e1b00b52515389e1cd80` as the input)

I don't know what is the expected behaviour, and how to add test cases for example.

radare:
Then thats a bug in my impl :)

libr/core/cmd_print.c
```diff
@@ -5330,7 +5330,18 @@ static int cmd_print(void *data, const char *input) {
...
 			}
 		}
-		
+
+		if (input[1] == 'x') { // pdx
+			if (input[2] == '?') {
+				r_cons_printf ("Usage: pdx [hexpairs] - disassemble hexpairs\n");
+				return 0;
+			} else {
+				r_core_cmdf (core, "pd @x:%s", r_str_trim_ro (input + 2));
```
radare: this can result in command injection issues, will be better to use the API to do that.

libr/core/cmd_print.c
```diff
@@ -5330,7 +5330,18 @@ static int cmd_print(void *data, const char *input) {
...
+				eprintf ("pd @x:%s\n", r_str_trim_ro (input + 2));
```
radare: this seems like a debug print leftover

radare:
Can you add tests and address the comments? i can confirm the issue you mention

libr/core/cmd_print.c
tantei3: Okay so I should call `__cmd_pad` function that is correctly implemented instead of the `r_core_cmdf` function before the block computation is resolved for the `pd` command, right? 

What about the `tpx` command? That also does a similar string formatting, wouldn't it be affected by the same command injection?

tantei3:
Can I get some links and resources on how to proceed with adding test cases? It would be helpful to know which files to edit in the repository and which files to look at as reference as well, since I am first time looking into this!

radare:
see other files in the radare2-regressions repository (under new/ db/*/*). to run then node_modules/.bin/r2r -i db/path/test

radare:
i think in the book there are some pages about this

libr/core/cmd_print.c
radare: remove this line

libr/core/cmd_print.c
radare: probably better to parse the hexpair string and return it back into hexpairs from the raw buffer. this way it wont be possible to inject code

radare:
ping

radare:
adding tests is easy, if you need any help drop a msg on the irc/telegram channel, but dont hold this issue for too much, can you address the comments quickly? i plan to release this sunday

tantei3:
Sorry for the delay, I added a few basic test cases for `pad`, `pix` and `pdx`. Also noticed that the current issue is that for the hexpairs that are in the range of UINT64 get compared to block size and hence that error is thrown for pix and pdx in some cases. To fix this, I have moved the evaluation of `pdx` above the compare with blocksize part, and for the `pix` case, I have added the bypass for the `i` commands.

radare:
i have fixed the conflict and the commit message and merged. hopefully will be green , otherwise ill chk whats wrong and fix it

radare:
thanks!

PR #14895, opened 2019-08-25
Hide brackets for LEA even if operand is section flag

kazarmy:


radare:
I think that parsing ansi and such inside disasm.c is wrong. Also, consider that other archs not x86 have also lea type instructions and those dont do brackets. Imho this must be inside the rparse api

radare:
is appveyour taking the branches like travis do'

because i see this red:

Screenshot 2019-08-25 at 23 13 31

radare:
yeah seems so

PR #14897, opened 2019-08-25
WIP: Remove recursive handle_stop_reason call

### LINKED ISSUE #14872: Infinite recursive loop/stack overflow in GDBclient handle_stop_reason
### Work environment

| OS/arch/bits (mandatory)                             | Arch linux x86_64
| File format of the file you reverse (mandatory)      | OpenOCD gdb
| Architecture/bits of the file (mandatory)            | arm, but N/A
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.7.1 22603 @ linux-x86-64 git.3.7.1-49-g7f9b8bdab commit: 7f9b8bdab345b433da4e124c0c07bbb26f0dee9d build: 2019-08-22__10:59:04

### Expected behavior

Cleanly errors or warns.

### Actual behavior

Segfaults due to stack overflow.

- Connect to OpenOCD gdb session. (eg radare2 -a arm -b 32 -d gdb://127.0.0.1:3000)
- dc in r2 (it should be waiting for a breakpoint)
- Connect a telnet session to OpenOCD telnet control port.
- r2 will spam console with the same message and then segfault.

This bug is also triggered by disconnecting a telnet session, connecting an additional gdb client to another OpenOCD session, exiting the OpenOCD process... I think it happens with many things that cause a OpenOCD status message.
 I have only been able to trigger this bug if R2 is "in dc mode" and waiting for a breakpoint to be hit.

Asciinema doesn't allow me to upload a 12mb file, so i attached a zipped copy of the recording here.
[gdbcrash.zip](https://github.com/radare/radare2/files/3530804/gdbcrash.zip)

2019-08-22-112701_381x400_scrot

A backtrace shows that https://github.com/radare/radare2/blob/master/shlr/gdb/src/gdbclient/responses.c#L252-L259 is the responsible section of code.

SrimantaBarua:
Fix #14872 

@roboman2444 could you test if this fixes your problem?

### DIFF
--- a/shlr/gdb/src/gdbclient/responses.c
+++ b/shlr/gdb/src/gdbclient/responses.c
@@ -256,7 +256,10 @@ int handle_stop_reason(libgdbr_t *g) {
 		if (send_ack (g) < 0) {
 			return -1;
 		}
-		return handle_stop_reason (g); // Wait for next stop message
+		memset (&g->stop_reason, 0, sizeof (libgdbr_stop_reason_t));
+		g->stop_reason.signum = -1;
+		g->stop_reason.reason = R_DEBUG_REASON_NONE;
+		return 0;
 	case 'W':
 		return stop_reason_exit (g);
 	case 'X':


radare:
i'm adding the WIP keyword until gets tested to be merged

radare:
(wip prefix, not suffix), for readibility

PR #14902, opened 2019-08-26
Fixes #14900: xdot type commands functional now

### LINKED ISSUE #14900: radiff2 dot generates invalid output
Seems it doesn't filter the colors from the dot output, or not convert them into the dot format
```
radiff2 -g main -m d /bin/false /bin/true | xdot -
```
Then xdot returns an error.

deepakchethan:


XVilka:
Broken tests, please fix them - seems a newline added?

PR #14918, opened 2019-08-28
Syncing with r2hexagon

Rot127:
Changes have been:
- Signed values get now printed as decimal (and therefore with sign): https://github.com/radare/radare2/issues/14849
- Immediate extensions now expand also duplex containers + missing non-duplex extensions added: https://github.com/radareorg/r2hexagon/pull/7
- Calling convention added


XVilka:
Thanks!

PR #14920, opened 2019-08-28
Define argument reg profile for SuperH4

brainstorm:
As spec'd in here:

Skärmavbild 2019-08-28 kl  22 26 10


libr/anal/p/anal_sh.c
```diff
@@ -1115,6 +1115,10 @@ static int sh_set_reg_profile(RAnal* anal) {
 		"=SN	r0\n"
 		"=SP	r15\n"
 		"=BP	r14\n"
+		"=A0	r4\n"
+		"=A1	r5\n"
+		"=A2	r6\n"
+		"=A3	r7\n"
```
radare: can you also add =R0 to refer the register holding the return value?

radare:
Aka r0

PR #14924, opened 2019-08-28
Implement 'cols' RTable.query to filter by column names ##print

radare:
* r_str_split_list() rewritten to support nth limit

r2 -qq -c 'aac;aflt size/gt/200,addr/cols/name/nbbs,nbbs/sort/inc' /bin/ls


radare:
Screenshot 2019-08-28 at 22 30 23

PR #14925, opened 2019-08-29
Enable mouse only if it was prev enabled and based on scr.wheel

deepakchethan:


libr/core/visual.c
```diff
@@ -3247,7 +3252,8 @@ R_API int r_core_visual_cmd(RCore *core, const char *arg) {
 				}
 			}
 			break;
-		case '/':
+		case '/': {
+			bool mouse_state = r_cons_singleton ()->mouse;
```
radare: i would make this into a separate static function:

static bool __holdMouseState(RCore *core) {
    int m = r_cons_singleton()->mouse;
	r_cons_enable_mouse (false);
       return m;
}

static void __releaseMouseState(RCore *core, bool oldState) {
  r_cons_enable_mouse (oldState && r_config_get_i (core->config, "scr.wheel"));
}

this way the repeated logic is in a separate function ,written only once. and have a meaningful name to please the reader

deepakchethan: Yeah, will change it in some time

radare:
ping

deepakchethan:
> ping

Ping reply, Made the changes :)

PR #14934, opened 2019-08-30
Fixes #14911: Use theme colors for p= and p==

### LINKED ISSUE #14911: Bars and fire dont follow color scheme
though that was fixed already

Screenshot 2019-08-27 at 15 57 17

deepakchethan:
We have only two bg based colors in the palette :]

Screen Shot 2019-08-30 at 10 35 42 PM

### DIFF
--- a/libr/util/print.c
+++ b/libr/util/print.c
@@ -83,11 +83,11 @@ R_API void r_print_columns (RPrint *p, const ut8 *buf, int len, int height) {
 	kol[3] = pal->mov;
 	kol[4] = pal->nop;
 	const char *bgkol[5];
-	bgkol[0] = Color_BGGREEN;
-	bgkol[1] = Color_BGGREEN;
-	bgkol[2] = Color_BGGREEN;
-	bgkol[3] = Color_BGWHITE;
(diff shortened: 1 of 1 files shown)


XVilka:
What about adding new color options? Something like chart.lvl1, chart.lvl2,
etc?

radare:
if we add custom color entries we will have to update all the themes. and i prefer to do that after r2con. initially my idea was to just reuse some colors from the current theme. not sure if the ones selected are the best options. but probably better than before. so im merging and we can discuss the themes IRL in the conference, best way to get feedback i guess :D

radare:
ok thats a complete disaster, lets create the new options and update all the themes.. because the ones you selected are only available in 4 themes: and the reason for that is because those colors NEED to set a background color, in fact none of those in the current themes set that.

```
$ git grep wordhl ../cons/d
../cons/d/bright:ec wordhl rgb:ff0000
../cons/d/cutter:ec wordhl rgb:ff0000
../cons/d/gentoo:ec wordhl rgb:ff0000
../cons/d/sepia:ec wordhl rgb:ff0000
```

PR #14938, opened 2019-08-31
Add frame pointer to AVR register profile

revyakin:
Lack of BP alias causes assertion fail in `libr/anal/esil.c:internal_esil_reg_write_no_null(...)` 
`WARNING: internal_esil_reg_write_no_null: assertion 'pc && sp && bp' failed (line 409)`


radare:
I think those asserts should become warnings instead, otherwise on release builds a feeding the wrong / corrupted profile will result in a crash, whihch is not desirable, can you do this change too?

radare:
i filled an issue to not forget about this https://github.com/radare/radare2/issues/14939

PR #14943, opened 2019-09-01
Fix #14939: Replace assertions about missing RReg profile registers with warnings

### LINKED ISSUE #14939: Make regprofile warnings instead of asserts
related to https://github.com/radare/radare2/pull/14938

revyakin:
I am not sure if I did it right. Waiting for your feedback.


radare:
nope. just chk if any of those required variables is not available, show the warning and return. this function cant work well with imaginary registernames like "pc" or "sp", the regprofile must be fixed to work, but we want it to not break or behave incorrectly, just warn.

thanks! can you update the pr now? i plan to release tonight

revyakin:
Ok. But first I try to explain why I wrote it this way.

The function has been written this way originally (2e2f0f2dfbd785183290d6c2e18579d8807a222a). Those imaginary registers are there just to prevent strcmp to segfault.

If we throw warning and exit the function then ESIL VM could not set the value of any register on arches with incomplete reg profile.

PR #14969, opened 2019-09-07
Align table based on column align field

GustavoLCR:


XVilka:
@deepakchethan for your information.

XVilka:
Looks good, merging. Can you please add a test for this in radare2-regressions?

PR #14971, opened 2019-09-07
Update io->off when seeking in intel hex

dudecc:
It seems like every other io plugin updates io->off when seeking and forces all reads to be from io->off, but ihex only does the latter, which seems to be breaking most high level read_at functions.

I noticed this when disassembling a hex file that used a lot of PC relative data loads -- the helpful comments at the end of the line to show the actual data being loaded were all just showing the opcode of the instruction that I started the disassembly on.

### DIFF
--- a/libr/io/p/io_ihex.c
+++ b/libr/io/p/io_ihex.c
@@ -209,7 +209,8 @@ static ut64 __lseek(struct r_io_t *io, RIODesc *fd, ut64 offset, int whence) {
 		return -1;
 	}
 	rih = fd->data;
-	return r_buf_seek (rih->rbuf, offset, whence);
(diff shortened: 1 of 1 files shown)


XVilka:
Can you also please add the test for `ihex://` formats in [radare2-regressions](https://github.com/radare/radare2-regresssions)?

XVilka:
Also looks like the AppVeyor build stuck forever

radare:
Omg so it was the issue?? @braincode can u test again?

This issue was fixed in another plugin. Will be good to have a check in io to avoid this to happen again if other plugins dont work as expected or maybe we should update it from the api instead of the plugin

radare:
The appveyour thing should be unrelated imho

PR #14973, opened 2019-09-07
R2CON2019 Fix Deleting old Types before Load

thestr4ng3r:
When re-loading multiple types in a single file, only one was deleted before loading the new ones, here is an example: https://github.com/radare/radare2-regressions/pull/1945

### DIFF
--- a/libr/anal/type.c
+++ b/libr/anal/type.c
@@ -95,21 +95,28 @@ R_API void r_anal_remove_parsed_type(RAnal *anal, const char *name) {
 
 R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
 	r_return_if_fail (anal && parsed);
-	// First, if this exists, let's remove it.
+
+	// First, if any parsed types exist, let's remove them.
 	char *type = strdup (parsed);
 	if (type) {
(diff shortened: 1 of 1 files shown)


libr/anal/type.c
```diff
@@ -95,21 +95,28 @@ R_API void r_anal_remove_parsed_type(RAnal *anal, const char *name) {
...
-			// Now add the type to sdb.
-			sdb_query_lines (anal->sdb_types, parsed);
-			save_parsed_type_size (anal, parsed);
 		}
 		free (type);
 	}
+
+	// Now add the type to sdb.
+	sdb_query_lines (anal->sdb_types, parsed);
+	save_parsed_type_size (anal, parsed);
```
radare: If this is an internal function shoukd start with __

thestr4ng3r: Which function? `save_parsed_type_size`? Since when do we prefix our internal functions with __?

radare: yes , for a while, see panels.c for example, or my dev slides

thestr4ng3r: k fixed

PR #14983, opened 2019-09-07
esil: fix arm32 stmia post-increment

anisse:
At r2con I had an idea for an r2wars bot but it wasn't supported in ESIL. Implement it !

There are other modes that could be supported (stmib, stmda, stmdb), but
that should do for now.


anisse:
And it has a test !
https://github.com/radare/radare2-regressions/pull/1946

PR #14998, opened 2019-09-09
Add info that r2pm -i is also used for pkg update

v45k0:
It is confusing that `r2pm update` only updates the database and the packages are not updated. Information about how to update a specific package is missing.

When the user searches for information how to update a particular package, e.g. using `r2pm -h`, the help provides information only how to install a package. This commit adds info that `r2pm -i` is also used for pkg update.


radare:
thanks!

Issue #13661, opened 2019-04-08
Add types information in function zignatures too


radare:
types: `txf.`

XVilka:
Related https://github.com/radare/radare2/issues/6947

radare:
Related https://github.com/radare/radare2/issues/13677

XVilka:
@oxcabe in fact, you might be interested in this too - from the implementation size r2 zignatures are just SDB databases, so it will be easy to add, once you determine how to store the function type, argument types, local variable types too.

XVilka:
I am afraid it can't be done without proper testing and tests in 3 days.

radare:
we can probably add some metrics from afis and afisf

radare:
done

PR #15000, opened 2019-09-09
Implement types field for zignatures ##types

oxcabe:


oxcabe:
Closes #13661.

This needs discussion before merging. It would be good to also add an eval var (zign.types) and choose what information goes inside this field in zignatures.

libr/anal/sign.c
```diff
@@ -67,6 +67,12 @@ R_API RList *r_sign_fcn_vars(RAnal *a, RAnalFunction *fcn) {
 	return ret;
 }
 
+R_API RList *r_sign_fcn_types(RAnal *a, RAnalFunction *fcn) {
```
radare: why have this wrapper function if its not providing anything ? why not just call r_anal_types_from_fcn ?

libr/anal/sign.c
```diff
@@ -1252,6 +1316,47 @@ static void listVars(RAnal *a, RSignItem *it, int format) {
 	}
 }
 
+static void listTypes(RAnal *a, RSignItem *it, int format) {
+	RListIter *iter = NULL;
+	char *type = NULL;
+	int i = 0;
+
+	if (format == '*') {
```
radare: will be good to use pj instead of handwritten json

libr/include/r_sign.h
```diff
@@ -63,6 +64,7 @@ typedef struct r_sign_item_t {
 	RList *refs;
 	RList *xrefs;
 	RList *vars;
+       RList *types;
```
radare: use tabs

libr/anal/sign.c
oxcabe: I didn't know if we can fetch more types from a function aside from the ones from txf. If the current implementation is enough it's obvious that we can just call r_anal_types_from_fcn from addFcnTypes

libr/anal/sign.c
oxcabe: Can you give an example of something that does it this way?

If we do print json with some API utility, we might want to do the same with the rest of the zignature fields, because they're all handwritten.

radare: See dadb421e481ba09362edc8171140a2a23fde2abc as example

libr/anal/sign.c
radare: well, the function name seems exactly the same, but with twisted order of words. so i would avoid that wrapper for now

radare:
sgtm we can merge first without discussing

libr/anal/sign.c
oxcabe: Good. I will open another PR for this refactoring later on, so I can do this for each field and specific cases of zignatures printing.

libr/anal/type.c
```diff
@@ -120,3 +120,21 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	char *type1 = (char *) a;
+	char *type2 = (char *) b;
+	return strcmp (type1 , type2);
+}
+
+R_API RList *r_anal_types_from_fcn(RAnal *anal, RAnalFunction *fcn) {
+	RListIter *iter;
+	RAnalVar *var;
+	RList *list = r_anal_var_all_list (anal, fcn);
+        RList *type_used = r_list_new ();
```
radare: use tabs

oxcabe:
> i think this pr is not adding support for checking them when searching for matches, right?

It doesn't match yet, as you said. The approach we are going to take is to populate anal/types SDB database after function analysis, then pass key-values matching each function as strings in the types field in zignature generation, so we can do the matching afterwards.

radare:
Just chk how many % of matching types theres between the sign and the function

oxcabe:
That could do it for searching, but there's also interest in applying types if zigns match with an stripped function: https://github.com/radareorg/radare2/issues/15372

PR #15009, opened 2019-09-10
Fix gdb reg write byte order

aronsky:
Two fixes:
1. Byte order of the written value is target-defined. Use `cfg.bigendian` for now. Possibly get it from target later?
2. Register number in write command (`P`) should be hexadecimal.

Hope it doesn't ruin other archs - but it complies with the definitions of the protocol (https://sourceware.org/gdb/onlinedocs/gdb/Packets.html#Packets), so should actually fix existing problems that nobody reported.

### DIFF
--- a/libr/debug/p/debug_gdb.c
+++ b/libr/debug/p/debug_gdb.c
@@ -1,5 +1,6 @@
 /* radare - LGPL - Copyright 2009-2018 - pancake, defragger */
 
+#include <r_core.h>
 #include <r_asm.h>
 #include <r_debug.h>
 #include <libgdbr.h>
@@ -311,6 +312,21 @@ static int r_debug_gdb_reg_write(RDebug *dbg, int type, const ut8 *buf, int size
 		}
 		ut64 val = r_reg_get_value (dbg->reg, current);
 		int bytes = bits / 8;
+		if (r_config_get_i(((RCore*)dbg->corebind.core)->config, "cfg.bigendian")) {
+			// TODO: validate that it's correct for all kinds of archs
+			switch (bytes) {
+				case 2:
+					val = r_swap_ut16(val);
+					break;
+				case 4:
+					val = r_swap_ut32(val);
+					break;
+				case 8:
+				default:
+					val = r_swap_ut64(val);
+					break;
+			}
+		}
 		gdbr_write_reg (desc, current->name, (char*)&val, bytes);
 	}
 	return true;
--- a/libr/debug/p/debug_gdb.c
+++ b/libr/debug/p/debug_gdb.c
@@ -312,19 +312,19 @@ static int r_debug_gdb_reg_write(RDebug *dbg, int type, const ut8 *buf, int size
 		}
 		ut64 val = r_reg_get_value (dbg->reg, current);
 		int bytes = bits / 8;
-		if (r_config_get_i(((RCore*)dbg->corebind.core)->config, "cfg.bigendian")) {
+		if (r_config_get_i (((RCore*)dbg->corebind.core)->config, "cfg.bigendian")) {
 			// TODO: validate that it's correct for all kinds of archs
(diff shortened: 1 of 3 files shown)


libr/debug/p/debug_gdb.c
```diff
@@ -311,6 +312,21 @@ static int r_debug_gdb_reg_write(RDebug *dbg, int type, const ut8 *buf, int size
 		}
 		ut64 val = r_reg_get_value (dbg->reg, current);
 		int bytes = bits / 8;
+		if (r_config_get_i(((RCore*)dbg->corebind.core)->config, "cfg.bigendian")) {
+			// TODO: validate that it's correct for all kinds of archs
+			switch (bytes) {
+				case 2:
```
radare: Case must be Indented at the same level as switch

libr/debug/p/debug_gdb.c
```diff
@@ -311,6 +312,21 @@ static int r_debug_gdb_reg_write(RDebug *dbg, int type, const ut8 *buf, int size
...
+					val = r_swap_ut16(val);
```
radare: Add apace before(

shlr/gdb/src/gdbclient/core.c
```diff
@@ -743,7 +743,7 @@ int gdbr_write_register(libgdbr_t *g, int index, char *value, int len) {
 		return -1;
 	}
 	reg_cache.valid = false;
-	ret = snprintf (command, sizeof (command) - 1, "%s%d=", CMD_WRITEREG, index);
+	ret = snprintf (command, sizeof (command) - 1, "%s%x=", CMD_WRITEREG, index);
```
radare: Scary that it ever worked at all

libr/debug/p/debug_gdb.c
```diff
@@ -311,6 +312,21 @@ static int r_debug_gdb_reg_write(RDebug *dbg, int type, const ut8 *buf, int size
 		}
 		ut64 val = r_reg_get_value (dbg->reg, current);
 		int bytes = bits / 8;
+		if (r_config_get_i(((RCore*)dbg->corebind.core)->config, "cfg.bigendian")) {
```
radare: Add space before (.. and i dont like the idea of calling config from here. It will break thendependency hirearchy i think. Can we have a gdbwrap api to set the endian and call it from outside?

shlr/gdb/src/gdbclient/core.c
aronsky: Yup, it worked for the first 10 registers. I was testing with `r5` at first (arbitrary), and using `dr` to print the registers before and after a change. The length of the list had me scrolling up every time to see the new value, so I switched to `r10`, and suddenly it stopped working - took me a while to understand why :)

libr/debug/p/debug_gdb.c
aronsky: It did require to link against r_config, but I'm not sure it's that bad - it's already linking against many core libraries.

What's gdbwrap?

libr/debug/p/debug_gdb.c
```diff
@@ -311,6 +312,21 @@ static int r_debug_gdb_reg_write(RDebug *dbg, int type, const ut8 *buf, int size
 		}
 		ut64 val = r_reg_get_value (dbg->reg, current);
 		int bytes = bits / 8;
+		if (r_config_get_i (((RCore*)dbg->corebind.core)->config, "cfg.bigendian")) {
```
radare: core can be null and should be moved outside the loop as bool bigEndian = dbg->corebind.cfgGet()..

if you use the corebind structure it will chk for null and will not require to link against rconfig

aronsky: Done.
Default to little-endian in case core is NULL, so the behavior is consistent with previous versions.

PR #15016, opened 2019-09-10
Skip analyzing functions if the region is filled by zeros ##anal

radare:


XVilka:
Maybe make it default one? Because it makes sense to me to do that.

PR #15024, opened 2019-09-12
Fix nopskip (its not fixed at all. just random typing here and there)

radare:
e io.cache=1
	e anal.nopskip=true
	wx 9090909090909090c3
	af
	pd 13
	afi

### DIFF
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -802,6 +802,16 @@ repeat:
 			}
 			gotoBeach (R_ANAL_RET_END);
 		}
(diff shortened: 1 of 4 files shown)


radare:
its fixed now

XVilka:
One broken test: `[XX] db/anal/x86_64 overlapping basic blocks and anal.jmp.mid  239`

radare:
i have cleanedup the fix and the test should be green now. i have spotted several other places to cleanup things related to code analysis

Issue #10313, opened 2018-06-10
Rarun2 regression in cad927a


mcd1992:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Arch Linux, x86_64
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86/64
| r2 -v full output, **not truncated** (mandatory)     | radare2 2.7.0-git 18346 @ linux-x86-64 git.2.6.0-169-g4ec3ef104 commit: 4ec3ef104192639ada7b08497d9b7a16c49a5218 build: 2018-06-10__04:20:48

### Expected behavior
Rarun2 should work properly and spawn its process

### Actual behavior
Rarun2 runs with no output and return code 0
Strace: https://gist.github.com/mcd1992/06c87b59ada9f7e4f0bd8034088632c5

### Steps to reproduce the behavior 
rarun2 program=/usr/bin/ls arg1=--help

### Additional Logs, screenshots, source-code,  configuration dump, ...

This regression seems to be introduced by https://github.com/radare/radare2/commit/cad927a5912114a26dd3f78383eea1a4f7dd5875 somehow. Multiple git bisects and a simple test of using master with a git revert on that single commit fixes the issue.

mcd1992:
Interesting it looks like the removal of `#include "r_userconf.h"` in `libr/include/r_types.h` is the cause.

PR #15040, opened 2019-09-15
Fix R2_PREFIX for Meson non-Windows builds with custom prefix

kazarmy:
Also tested with `sys/user.sh`, works fine.

### DIFF
--- a/libr/include/r_types.h
+++ b/libr/include/r_types.h
@@ -3,7 +3,7 @@
 
 // defines like IS_DIGIT, etc'
 #include "r_util/r_str_util.h"
-#include "r_userconf.h"
+#include <r_userconf.h>
 #include <stddef.h>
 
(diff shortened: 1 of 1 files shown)


kazarmy:
This is because `#include "r_userconf.h"` finds `libr/include/r_userconf.h` (an artifact from a previous sys/user.sh build) before `build/r_userconf.h` (which is what the Meson build generates). Switching to `#include <r_userconf.h>` fixes this, and also other #includes of r_userconf.h use `#include <r_userconf.h>` as well.

kazarmy:
Can't repro #10313 ... if the issue happens again, just ask the person to purge everything, repull and rebuild.

kazarmy:
... purging must also include all installed r2 in the system.

kazarmy:
... and purging must also include their data files, not just their binaries.

PR #15051, opened 2019-09-16
Set rpath if non-Windows Meson shared build

kazarmy:
For non-root installs and side-by-side builds.

### DIFF
--- a/binr/r2agent/meson.build
+++ b/binr/r2agent/meson.build
@@ -7,5 +7,6 @@ executable('r2agent', 'r2agent.c',
     r_cons_dep
   ],
   install: true,
+  install_rpath: get_option('default_library') == 'shared' ? '$ORIGIN/../' + get_option('libdir') : '',
   implicit_include_directories: false
 )
--- a/binr/r2agent/meson.build
+++ b/binr/r2agent/meson.build
@@ -7,6 +7,6 @@ executable('r2agent', 'r2agent.c',
     r_cons_dep
   ],
   install: true,
-  install_rpath: get_option('default_library') == 'shared' ? '$ORIGIN/../' + get_option('libdir') : '',
+  install_rpath: rpath,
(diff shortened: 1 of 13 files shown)


kazarmy:
For non-root installs and side-by-side builds.

radare:
@GustavoLCR can you have a look at this?

kazarmy:
Meson apparently ignores `install_rpath` on Windows: https://github.com/mesonbuild/meson/blob/7e2c54c0a38788380c7f05dff3244faaa76e0220/mesonbuild/compilers/d.py#L218-L219

radare:
please rebase the pr. the rpath is unrecommended for distros becacuse they want to override all the flags, and the rpath is quite intrussive. thats why --with-rpath is not enabled by default in configure.

Also, i see some spaguetti here, can we simplify this to have it only in 1 place?

kazarmy:
> please rebase the pr.

done

> the rpath is unrecommended for distros becacuse they want to override all the flags, and the rpath is quite intrussive.

ok i'll only set rpath if there is a `--local` flag

> Also, i see some spaguetti here, can we simplify this to have it only in 1 place?

probably

kazarmy:
> Also, i see some spaguetti here, can we simplify this to have it only in 1 place?

Since Meson [does not support user-defined functions](https://mesonbuild.com/Syntax.html#userdefined-functions-and-methods) then fbe7d59bcdd3f1b0dfdd5a5d1621185f42dcd220 apparently is best possible if only involving `install_rpath`.

radare:
Is there a way to build with meson.. without the rpath?

kazarmy:
> Is there a way to build with meson.. without the rpath?

yes 50c716123e977a3ef2e0daf2755c64510cdad9be by default, rpath is not set

kazarmy:
ok to merge?

PR #15056, opened 2019-09-16
WIP: Fix infinite loop because rbtree is broken as hell

radare:


ret2libc:
What is broken in rbtree?

PR #15063, opened 2019-09-17
Simplify ret_min == ret_addr check

kazarmy:


radare:
This change was on purpose. Because release builds will crash if that happens and i have the bug located i will fix the issue for next release probably. But this shouldnt happen at all.

PR #15067, opened 2019-09-18
Fix & has lower precedence than ==

lzutao:


radare:
The ==1 can be removed too

PR #15075, opened 2019-09-19
Move Lib already loaded check before dlopen Call

thestr4ng3r:
for https://github.com/radareorg/r2ghidra-dec/issues/38

This prevents constructors, for example doing C++ static initialization, from being executed twice.

### DIFF
--- a/libr/util/lib.c
+++ b/libr/util/lib.c
@@ -239,10 +239,30 @@ R_API int r_lib_close(RLib *lib, const char *file) {
 	return -1;
 }
 
+static bool __already_loaded(RLib *lib, const char *file) {
+	const char *fileName = r_str_rstr (file, R_SYS_DIR);
+	RLibPlugin *p;
+	RListIter *iter;
+	if (fileName) {
+		r_list_foreach (lib->plugins, iter, p) {
+			const char *pFileName = r_str_rstr (p->file, R_SYS_DIR);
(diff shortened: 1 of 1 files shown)


radare:
My concern on this code was mainly to lod the same plugin twice just because its copied in multiple paths or its the same with different name. Not sure if we should use checksums add another check to avoid this to happen

thestr4ng3r:
Checksums break again if you have a slightly different version of the same plugin. One good way would be to add a unique id string per plugin to `RLibStruct` and check that after loading the lib and before registering the plugin. But we still need to make sure not to call any constructors before we haven't checked the duplication. I don't know enough about dynamic loading to say if this is possible on all platforms.

radare:
yes, thats what we must check.. but if c++/objc crap run things when you do dlopen.. not sure how can we handle this in the right way without using RBin instead

radare:
lets merge for now, most people delete all the plugins after every r2 update, as well as most people put them all in the same place. so it shouldnt be a problem

thestr4ng3r:
Yes, this patch only makes the situation better. Actually even after it there is a special case when you have a duplicate plugin and the one that is loaded first triggers the `Module version mismatch /home/florian/bin/prefix/radare2/lib/radare2/core_ghidra.so (3.8.0) vs (4.0.0-git)`, then it still breaks. I have to investigate whether this is fixable somehow.

PR #15078, opened 2019-09-19
Fix #14845: Support radare2_rc in radiff2 for graphs

### LINKED ISSUE #14845: Use unicode settings (`scr.utf8` and `scr.utf8.curvy`) in radiff2 graphs

deepakchethan:


libr/core/cconfig.c
```diff
@@ -3606,6 +3606,53 @@ R_API int r_core_config_init(RCore *core) {
 	return true;
 }
 
+R_API void radare2_rc(RCore *r) {
+	char* env_debug = r_sys_getenv ("R_DEBUG");
+	bool has_debug = false;
+	if (env_debug) {
```
radare: check if *env_debug or +env_debug=='1'

libr/core/cconfig.c
```diff
@@ -3606,6 +3606,53 @@ R_API int r_core_config_init(RCore *core) {
 	return true;
 }
 
+R_API void radare2_rc(RCore *r) {
+	char* env_debug = r_sys_getenv ("R_DEBUG");
```
radare: swap those lines, better have the env-debug closer to the if

radare: we have other cases where we only use env vars as boolean, what about having r_sys_getenv_asbool() to avoid this heap massaging and custom checks here and there?

libr/core/cconfig.c
```diff
@@ -3606,6 +3606,53 @@ R_API int r_core_config_init(RCore *core) {
 	return true;
 }
 
+R_API void radare2_rc(RCore *r) {
```
radare: wrong function name. public functions should start with `r_core_`, and i dont think `r_core_radare2_rc` is a good name either

libr/util/sys.c
```diff
@@ -511,6 +511,18 @@ R_API char *r_sys_getenv(const char *key) {
 #endif
 }
 
+R_API bool r_sys_getenv_asbool(const char *key) {
+	char *env = r_sys_getenv (key);
+	if (env) {
+		if (*env || *env=='1') {
```
radare: use free instead of r_free and add spaces around ==

PR #15082, opened 2019-09-20
Integrate types field in zignature operations ##signatures

oxcabe:
Integrated types in the most significant 'z' commands.

By the way, is the "vars" field implemented into the commands? 
When I  `git grep "r_sign_match_vars"`, can only see the definition and the implementation, but not even a single use for this fcn.


XVilka:
Ok, but it's not what I asked for. It's good to have, but far from the main goal. Main goal is:
- When we have a signature matching a raw (unanalyzed code)
- Upon match it should not only add the flag, it should create the function with a specified size (in the signature) and set the corresponding argument names, argument types, return type.

PR #15084, opened 2019-09-20
Fix waf

GustavoLCR:


radare:
use the r_io api, the idea of printing the command is because it can be used with |. and such, what is this fixing aprt from changing the bevaviour? using r_core_cmd to write few bytes is an overkill

libr/core/cmd_write.c
```diff
@@ -1555,7 +1555,7 @@ static int cmd_write(void *data, const char *input) {
 							RAsmCode *ac = r_asm_massemble (core->assembler, b);
 							char* hex = r_asm_code_get_hex (ac);
```
radare: Use ac->bytes and ac->len, but this should be rewritten to use the RBuffer API instead of having two separate fields

radare: and use `r_io_write_at (core->io, addr, ac->code, ac->len);`

PR #15092, opened 2019-09-20
Some optimization for r_str_replace() and r_str_replace_icase()

pelijah:
1) Don't use `strlen()` to calc the tail len
2) Use `memmove()` instead of `strdup() + memcpy() + free()`
3) Use `realloc` only if `vlen != klen`


radare:
Very nice one! Thanks!

PR #15102, opened 2019-09-22
Change r_sys_get_src_dir_w32() behaviour

pelijah:


radare:
Im totally against having hacks in r2 just for one app. Why cutter needs to be different

pelijah:
Agree. But somebody added it. I just wanted to improve it.
Alternatives:
1) Rename this macro to smth like `NOT_INSIDE_BIN_DIR` that sounds more universal (still too hacky)
2) Move Cutter.exe to `bin` dir (just a workaround; some people strongly against it)
3) Use a env var to redefine behaviour at runtime

pelijah:
I think 3-rd variant is the best one. Feel free to change var's name.

radare:
lgtm

PR #15108, opened 2019-09-23
Introduction of hash comparison api #mem

devnexen:
Unlike bcmp/memcmp, goes through the whole buffer and using
volatile pointers to discard possible compiler optimisations
making it more suitable for sensitive data comparisons.


radare:
Any real use case for this? i have never had any issue with memcmp

devnexen:
> Any real use case for this? i have never had any issue with memcmp

It reduces significantly false positives risk let's say you can rahash2 via scripts even though, this is usually made for hash comparisons like here, crypto matters etc ... but that s ok if it does not get merged :)

radare:
i dont think it makes much sense, memcmp should be reliable enough unless your libc is broken

PR #15116, opened 2019-09-25
Update stiv.c to remove warning

Sanyam-Agrawal:
Removed cause for warning of misleading indentation of 'if' clauses.

### DIFF
--- a/libr/cons/stiv.c
+++ b/libr/cons/stiv.c
@@ -53,12 +53,9 @@ static void render_ansi(PrintfCallback cb_printf, const ut8 *c, const ut8 *d) {
 }
 
 static int rgb(int r, int g, int b) {
-	if (r<0) r=0;
-	if (r>255) r = 255;
-	if (g<0) g=0;
-	if (g>255) g = 255;
-	if (b<0) b=0;
-	if (b>255) b = 255;
+	r = R_DIM (r, 0, 255);
+	g = R_DIM (g, 0, 255);
+	b = R_DIM (b, 0, 255);
 	r = (int)(r/50.6); 
 	g = (int)(g/50.6);
 	b = (int)(b/50.6);
--- a/libr/cons/stiv.c
+++ b/libr/cons/stiv.c
@@ -53,9 +53,12 @@ static void render_ansi(PrintfCallback cb_printf, const ut8 *c, const ut8 *d) {
 }
 
 static int rgb(int r, int g, int b) {
-	if (r<0) r=0; if (r>255) r = 255;
-	if (g<0) g=0; if (g>255) g = 255;
-	if (b<0) b=0; if (b>255) b = 255;
+	if (r<0) r=0;
+	if (r>255) r = 255;
+	if (g<0) g=0;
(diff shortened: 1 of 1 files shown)


XVilka:
Looks like the build stuck at the pulling the r2r tests:
```
cd radare2-regressions/new
$env:pull_owner=($env:APPVEYOR_PULL_REQUEST_HEAD_REPO_NAME -split "/")[0]
if (-not [string]::IsNullOrWhiteSpace($env:pull_owner)) {cmd /c "(git fetch https://github.com/%pull_owner%/radare2-regressions %APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH%:%pull_owner%_%APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH% && git checkout %pull_owner%_%APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH%) > nul 2>&1 || exit 0"} else {echo "Using master"}
```
@GustavoLCR do you have any ideas what can cause this? Since it's you who added this feature AFAIR.

GustavoLCR:
> Looks like the build stuck at the pulling the r2r tests:
> 
> ```
> cd radare2-regressions/new
> $env:pull_owner=($env:APPVEYOR_PULL_REQUEST_HEAD_REPO_NAME -split "/")[0]
> if (-not [string]::IsNullOrWhiteSpace($env:pull_owner)) {cmd /c "(git fetch https://github.com/%pull_owner%/radare2-regressions %APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH%:%pull_owner%_%APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH% && git checkout %pull_owner%_%APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH%) > nul 2>&1 || exit 0"} else {echo "Using master"}
> ```
> 
> @GustavoLCR do you have any ideas what can cause this? Since it's you who added this feature AFAIR.

The only thing I can think of is that the pr author doesn't have a radare2-regressions fork, but i dont see how it would cause it to hang, git should just exit

radare:
Thats a random travis issue. Not sure if its a bug in our side

libr/cons/stiv.c
```diff
@@ -53,9 +53,12 @@ static void render_ansi(PrintfCallback cb_printf, const ut8 *c, const ut8 *d) {
 }
 
 static int rgb(int r, int g, int b) {
-	if (r<0) r=0; if (r>255) r = 255;
-	if (g<0) g=0; if (g>255) g = 255;
-	if (b<0) b=0; if (b>255) b = 255;
+	if (r<0) r=0;
```
radare: in r2land we use braces even for single statements, but this is clearly a DIM to constraint the value between 0 and 255, so we can use `r = R_DIM (r, 0, 255);` instead of doing this (this was written like this because the original code was living outside r2land and didnt had that macro)

Sanyam-Agrawal: Made the recommended change in a new commit in this branch.

PR #15120, opened 2019-09-26
Rebase things when reopening file in debugger mode

GustavoLCR:
This is in a "better than nothing" stage.

Currently rebasing only cur bin:
* Functions (and bbs)
* Flags
* Meta items (comments, etc)
* Refs / XRefs


radare:
I like this better-than-nothing approach. Let's start to move forward rebasing breakpoints and all that stuff too. i dont expect to spend much time on debugger for this release. mainly focusing on analysis from my side, but all that stuff is clearly welcome :)

PR #15137, opened 2019-09-27
Freeing main assembly object ##rasm2

devnexen:


libr/main/rasm2.c
```diff
@@ -47,6 +47,7 @@ static void __as_free(RAsmState *as) {
 	r_asm_free (as->a);
 	r_anal_free (as->anal);
 	r_lib_free (as->l);
+    	R_FREE (as);
```
radare: just call free, no need to nullify because its not used outside the function

devnexen: True it was just to counter R_NEW0 style-wise :-)

PR #15140, opened 2019-09-28
hotfix for #14851

condret:
cleaner fix is WIP.
This one should work for now


radare:
can you update the 3 tests and add one to confirm the issue is gone?

PR #15149, opened 2019-09-29
Implementing debug info per pid on NetBSD ##debug

devnexen:
While at it fixing little code mistake about freeing const address.


unixfreaxjp:
@devnexen please elaborate my stupid head about what do you mean by “debug info per pid“ ?

devnexen:
Nothing in particular but it is all about retrieving a process state.

PR #15166, opened 2019-10-02
Fix x86 assembler output for some `test` ops

blin00:


radare:
can you add some tests?

PR #15169, opened 2019-10-02
Refactor RAnal Basic Blocks and Functions ##anal

radare:


libr/anal/block.c
```diff
@@ -22,74 +35,95 @@ R_API RAnalBlock *r_anal_block_split(RAnalBlock *bb, ut64 addr) {
...
 }
 
 R_API void r_anal_del_block(RAnal *anal, RAnalBlock *bb) {
 	r_return_if_fail (anal && bb);
 D eprintf ("del block (%d) %llx\n", bb->ref, bb->addr);
 	BBAPI_PRELUDE (NULL);
-	const ut64 k = __bbHashKey (bb->addr);
 	r_anal_block_ref (bb);
-	r_list_free (bb->fcns);
+	/*
```
radare: use #if0 to comment code

radare: or better... delete it

libr/anal/block.c
```diff
@@ -22,74 +35,95 @@ R_API RAnalBlock *r_anal_block_split(RAnalBlock *bb, ut64 addr) {
...
+		return NULL;
+	}
+	RAnalBlock *block = container_of (node, RAnalBlock, rb);
+	if (addr - block->addr < block->size) {
+		return block;
 	}
 	return NULL;
 }
 
+R_API void r_anal_get_blocks_intersect(RAnal *anal, ut64 addr, ut64 size, R_OUT RPVector *out) {
```
radare: wha about returning RList* instead of using the R_OUT parameter? im aware about the performance reasons..

libr/anal/block.c
```diff
@@ -22,74 +35,95 @@ R_API RAnalBlock *r_anal_block_split(RAnalBlock *bb, ut64 addr) {
...
+	BBAPI_PRELUDE(x)
```
radare: missing space before (

libr/anal/block.c
```diff
@@ -22,74 +35,95 @@ R_API RAnalBlock *r_anal_block_split(RAnalBlock *bb, ut64 addr) {
...
-				return b;
-			}
-		}
+R_API RAnalBlock *r_anal_get_block_at(RAnal *anal, ut64 addr) {
+	// TODO: this might be a bit faster using a ht
+	RBNode *node = r_rbtree_find (anal->bb_tree, &addr, __bb_addr_cmp, NULL);
+	if (!node) {
+		return NULL;
+	}
+	return container_of (node, RAnalBlock, rb);
```
radare: use ternary operator here: return node? container_of(...): NULL;

libr/include/r_anal.h
```diff
@@ -296,7 +296,7 @@ typedef struct r_anal_function_t {
 	RList *locs; // list of local variables
 	RList *fcn_locs; //sorted list of a function *.loc refs
 	//RList *locals; // list of local labels -> moved to anal->sdb_fcns
-	RList *bbs;
+	RList *bbs; // TODO: should be RPVector
```
radare: and just store ut64 ?

libr/include/r_anal.h
```diff
@@ -296,12 +296,14 @@ typedef struct r_anal_function_t {
...
 	RAnalFcnMeta meta;
 	RRangeTiny bbr;
 	RBNode rb;
 	RBNode addr_rb;
 	RList *imports; // maybe bound to class?
+	struct r_anal_t *anal; // this function is associated with this instance
```
radare: RAnal instead of `struct r_anal_t`

radare:
many of those because of changint the bbsize to ut64
```
canal.c: In function 'core_anal_graph_construct_nodes':
canal.c:1486:42: warning: format '%d' expects argument of type 'int', but argument 3 has type 'long long unsigned int' [-Wformat=]
                                 eprintf ("cannot allocate %d byte(s)\n", bbi->size);
```

radare:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

ret2libc:
I will review later today or on Monday

ret2libc:
I may be missing a bit of context... What’s the exact purpose of this new api? 

Also, It seems it changes a lot of places.

libr/anal/block.c
thestr4ng3r: It's checked 3 lines above.

thestr4ng3r:
The idea is to make basic blocks completely independent of functions. All basic blocks are stored in one global tree and a function is defined by only a single entry point address.
This makes it possible to have overlapping functions (for example from a tailcall), as well as simplifies different kinds of queries, like basic block containing an address.

libr/include/r_anal.h
thestr4ng3r: RAnal is further down

libr/anal/block.c
thestr4ng3r: fine for me

pelijah:
Could you rebase it?

thestr4ng3r:
rebased

thestr4ng3r:
Do we really need refcounting in the function? I think it would be enough to just have refs in the basic blocks. The primary purpose of it is to free basic blocks when they are contained in no function anymore.

thestr4ng3r:
I thought a bit about the idea of having functions only identified by their entry. This actually makes things quite complicated.
The reason is that the following invariant must always hold: For all functions, every basic block that is reachable when walking all basic blocks recursively (`r_anal_block_recurse()`) starting at the function entry must have this function in their list.
This means if you create a new bb at some addr, you must find all basic blocks that have a jump, fail or switch case going to this addr and add the new bb to their functions.

Maybe we should first start by only decoupling bbs from functions (i.e. making them global), but still keep the bbs list in every RAnalFunction. In this case, if you add a new basic block at some addr, it would not be implicitly added to every function that can reach this block.

thestr4ng3r:
So it actually is possible to have overlapping basic blocks when there is a jump into the middle of an instruction currently, as tested here: https://github.com/radareorg/radare2/blob/master/test/new/db/anal/x86_64#L2343

How do we want to handle this? If blocks can overlap, the global block tree will be much harder to handle. First, it has to be augmented and search by address gets harder and maybe slower. Also when querying by addr, there can be multiple blocks returned.

I think making RAnalBlocks actually overlappable would be a bad solution because these are quite rare cases and have huge implications as described above.
Instead, for the case of jump into the middle of an instruction, I think we can have two better kinds of solution approaches:
* Split the block into two, but display the first one differently in the graph view. The first one can have a flag set that indicates that it ends in the middle of an instruction and should be displayed coalesced with the following one.
* Keep the block unsplit and only display the jmp into the middle of it differently.

I am currently a bit in favor of the first approach, because I think it makes more sense for analysis.

ret2libc:
> How do we want to handle this? If blocks can overlap, the global block tree will be much harder to handle. First, it has to be augmented and search by address gets harder and maybe slower. Also when querying by addr, there can be multiple blocks returned.

Why slower? Isn't it just another basic block in the rbtree? 

> Instead, for the case of jump into the middle of an instruction, I think we can have two better kinds of solution approaches:
> * Split the block into two, but display the first one differently in the graph view. The first one can have a flag set that indicates that it ends in the middle of an instruction and should be displayed coalesced with the following one.
> * Keep the block unsplit and only display the jmp into the middle of it differently.

What do you mean exactly?
So, let's assume you have basic block `B`, which is composed of `i1, i2, i3, i4, i5`. You now reach a point where you jump in the middle of `i2`, so you would get a path that executes `i1, i2, i3, i4, i5` and another path that executes `i2_2, i3_2, i4, i5` (supposing that `i2_2` and `i3_2` are different from `i2` and `i3` and then, from `i4`, the disassembly re-aligns again). In such a case, I would split the basic blocks in such a way that you would have `B1` composed of `i1, i2, i3`, `B2` composed of `i2_2, i3_2` and `B3` composed of `i4, i5`, with `B1 -> B3` and `B2 -> B3`. Does this make sense?

thestr4ng3r:
> Why slower? Isn't it just another basic block in the rbtree?

No. If blocks can never overlap, a classic binary tree is enough. To get the block that contains a certain address, you just walk through the tree and take the greatest node that is smaller or equal to the given address (see upper bound pr). This is the only candidate that could potentially contain the address.

If blocks are allowed to overlap, you can have something like blocks {[1, 1337), [3, 42)} and you want to get the block containing 5. Then the above algorithm breaks because you would only consider [3, 42). You would have to use an augmented tree here. See the implementation of `r_interval_node_all_in()`, which needs two recursive calls.


> What do you mean exactly?
> So, let's assume you have basic block `B`, which is composed of `i1, i2, i3, i4, i5`. You now reach a point where you jump in the middle of `i2`, so you would get a path that executes `i1, i2, i3, i4, i5` and another path that executes `i2_2, i3_2, i4, i5` (supposing that `i2_2` and `i3_2` are different from `i2` and `i3` and then, from `i4`, the disassembly re-aligns again). In such a case, I would split the basic blocks in such a way that you would have `B1` composed of `i1, i2, i3`, `B2` composed of `i2_2, i3_2` and `B3` composed of `i4, i5`, with `B1 -> B3` and `B2 -> B3`. Does this make sense?

Yes that makes sense, but `B1` and `B2` would overlap.
My two different approaches for this example would look like this:
* `B1 = i1, i2 (first half) + continue flag set`, `B2 = i2 (second half), i3, i4, i5`. So if you show `B1` in the graph view, it would contain the disassembly of the bytes in `B1` and `B2` combined and `B2` would be shown as just its own bytes disassembled. Here you couldn't cache the exact instruction positions in the blocks anymore (`RAnalBlock.op_pos`). There would probably have to be another flag in these flags that instructions should always be disassembled on-demand.
* Simply a single block `B = i1, i2, i3, i4, i5` and a jump from somewhere in the middle of `B`. The graph view would be in charge of rendering it in some way that makes sense. But this rendering might be kind of hard and any analysis based on the basic blocks will not know directly that this block would actually be somehow split.

radare:
Blocks cant overlap.. but.. is thwre a way to detect that before its too late?

ret2libc:
@radare how so? Blocks can overlap and we have examples of that. Also, currently there is a augmented rbtree that handles basic blocks well so if the new implementation can’t support overlapping blocks that imho is worse

radare:
How? With jump in the middle things?

thestr4ng3r:
Yes, that is exactly what we have been discussing in the last comments, starting at https://github.com/radareorg/radare2/pull/15169#issuecomment-568466420

thestr4ng3r:
@ret2libc Do you have other examples in addition to jump in the middle of instructions where overlapping basic blocks can make sense in practice? It would certainly possible to support them using RIntervalTree, but if it will be only used for very few specific and rare things I would prefer to go with one of the solutions I proposed.

libr/core/canal.c
```diff
@@ -4909,7 +4944,7 @@ R_API void r_core_anal_esil(RCore *core, const char *str, const char *target) {
 		}
 		// if (op.type & 0x80000000 || op.type == 0) {
 		if (op.type == R_ANAL_OP_TYPE_ILL || op.type == R_ANAL_OP_TYPE_UNK) {
-			// i +=2;
+			i += mininstrsz - 1;
```
thestr4ng3r: @pancake Where do these changes come from? They do not seem to have anything to do with the anal block refactoring.

libr/anal/anal_ex.c
```diff
@@ -254,8 +254,9 @@ R_API RList* r_anal_ex_analysis_driver(RAnal *anal, RAnalState *state, ut64 addr
 
 R_API void r_anal_ex_op_to_bb(RAnal *anal, RAnalState *state, RAnalBlock *bb, RAnalOp *op) {
 	//ut64 cnd_jmp = (R_ANAL_EX_COND_OP | R_ANAL_EX_CODEOP_JMP);
-	bb->addr = op->addr;
-	bb->size = op->size;
+	eprintf ("this is wrong\n");
```
radare: kill this debug printf, or prefix it with "Warning: " and a better message (and honor anal->verbose)

libr/anal/bb.c
```diff
@@ -74,7 +39,9 @@ R_API int r_anal_bb(RAnal *anal, RAnalBlock *bb, ut64 addr, const ut8 *buf, ut64
 		}
 		r_anal_bb_set_offset (bb, bb->ninstr++, at - bb->addr);
 		idx += oplen;
-		bb->size += oplen;
+		if (!r_anal_block_try_resize_atomic (bb, bb->addr, bb->size + oplen)) {
+			eprintf ("overlap\n");
```
radare: same here

libr/anal/esil_dfg.c
```diff
@@ -579,6 +580,34 @@ static bool _edf_consume_2_set_reg(RAnalEsil *esil, const bool use_origin) {
...
+	RGraphNode *src_node = _edf_reg_get (edf, src);
+	if (!src_node) {
+		int src_type = r_anal_esil_get_parm_type (esil, src);
+		if (src_type == R_ANAL_ESIL_PARM_INVALID) {
+			free (dst);
+			free (src);
+			return false;
+		}
+
+		if (src_type == R_ANAL_ESIL_PARM_NUM) {
```
radare: use switch isntead of 2 ifs without an else

libr/anal/esil_dfg.c
```diff
@@ -579,6 +580,34 @@ static bool _edf_consume_2_set_reg(RAnalEsil *esil, const bool use_origin) {
...
+			RGraphNode *n_value = r_graph_add_node (edf->flow, r_anal_esil_dfg_node_new (edf, src));
+			RAnalEsilDFGNode *ec_node = r_anal_esil_dfg_node_new (edf, src);
+			r_strbuf_appendf (ec_node->content, ":const_%d", edf->idx++);
+			src_node = r_graph_add_node (edf->flow, ec_node);
+			r_graph_add_edge (edf->flow, n_value, src_node);
+		}
+		if (src_type == R_ANAL_ESIL_PARM_REG) {
+			RGraphNode *n_reg = r_graph_add_node (edf->flow, r_anal_esil_dfg_node_new (edf, src));
```
radare: can be null?

libr/anal/function.c
```diff
@@ -0,0 +1,145 @@
...
+	ht_up_delete (anal->ht_addr_fun, fcn->addr);
+	ht_pp_delete (anal->ht_name_fun, fcn->name);
+	r_list_delete_data (anal->fcns, fcn);
+	r_anal_fcn_tree_delete (anal, fcn);
+	if (!r_anal_fcn_tree_delete (anal, fcn)) {
+		return false;
+	}
+	r_list_free (fcn->bbs);
+	fcn->bbs = NULL;
+#if 0
```
radare: dead code, i think its safe to reovme this if0 code

libr/include/r_anal.h
```diff
@@ -1410,6 +1417,74 @@ R_API int r_anal_op_hint(RAnalOp *op, RAnalHint *hint);
...
+R_API bool r_anal_block_recurse(RAnalBlock *block, RAnalBlockCb cb, void *user);
+
+// same as r_anal_block_recurse, but returns the blocks as a list
+R_API RList *r_anal_block_recurse_list(RAnalBlock *block);
+
+/* function.c */
+R_API RAnalFunction *r_anal_add_function(RAnal *anal, const char *name, ut64 addr);
+R_API const RList *r_anal_get_functions(RAnal *anal, ut64 addr);
+R_API bool r_anal_del_function(RAnalFunction *fcn);
+#if 0
```
radare: remove this if0 code

libr/core/canal.c
radare: to avoid infinite loops basically

thestr4ng3r:
So if I don't get any more feedback, I will just implement this:

> * Split the block into two, but display the first one differently in the graph view. The first one can have a flag set that indicates that it ends in the middle of an instruction and should be displayed coalesced with the following one.

ret2libc:
Please wait a little bit. It’s Christmas and enotime for computer these days, busy with family. :)

thestr4ng3r:
https://www.youtube.com/watch?v=K8E_zMLCRNg

radare:
it doesnt  build

XVilka:
@thestr4ng3r I support you for the first option - splitting BB in half and add a special attribute to coalesce with another one.

radare:
lgtm, just make travis green, im happy with all those changes, and if we break something in anal we may see it by reviewing the tests. thank you!!

kazarmy:
> @ret2libc Do you have other examples in addition to jump in the middle of instructions where overlapping basic blocks can make sense in practice? It would certainly possible to support them using RIntervalTree, but if it will be only used for very few specific and rare things I would prefer to go with one of the solutions I proposed.

Jumping into the middle of instructions is used to confuse a disassembler, so even though it's a rare thing, unfortunately it's an important rare thing.

> Split the block into two, but display the first one differently in the graph view. The first one can have a flag set that indicates that it ends in the middle of an instruction and should be displayed coalesced with the following one.

The byte immediately afterwards can be part of 2 different instructions (!) and I think keeping blocks  unsplit models this better.

thestr4ng3r:
Proof of concept for overlapping blocks: #15720

radare:
green \o/

radare:
this api naming change shuold be rolled back imho because RAnalFunction apis should get RAnalFunction as first argument, otherwise it must be a getter from RAnal like it was before, or just use a long name like: r_anal_get_function_at() 

Screenshot 2019-12-31 at 01 38 37

radare:
+R_DEPRECATE R_API RAnalFunction *r_anal_get_fcn_in(RAnal *anal, ut64 addr, int type); // use r_anal_function_get_at/r_anal_function_get_in

we should always use r_anal_get_fcn_in_list() now, totally agree on deprecating this function, also what about using the longnaming for the get_fcn_in_list() ?

radare:
same for the block api

Screenshot 2019-12-31 at 01 43 55

radare:
unnecessary parenthesis in
```
--- a/libr/core/cmd_type.c
+static int bb_cmpaddr(const void *_a, const void *_b) {
+       const RAnalBlock *a = _a, *b = _b;
+       return (a->addr - b->addr);
+}
```

radare:
Benchmark results: it's slower than master (numbers are in seconds)

```
bin  cmd before after
bash aa   3.5   5.4
bash aaa  12     16
```

libr/anal/block.c
```diff
@@ -0,0 +1,582 @@
...
+
+	// check if function lists are identical
+	RAnalFunction *fcn;
+	RListIter *iter;
+	r_list_foreach (a->fcns, iter, fcn) {
+		if (!r_list_contains (b->fcns, fcn)) {
+			return false;
+		}
+	}
+	r_list_foreach (b->fcns, iter, fcn) {
```
GustavoLCR: can't you just do:
```
if (r_list_length (a->fcns) != r_list_length (b->fcns)) {
	return false;
}

r_list_foreach (a->fcns, iter, fcn) {
	if (!r_list_contains (b->fcns, fcn)) {
		return false;
	}
}
```
and avoid the double loop?

thestr4ng3r: Yes, since the fcns list must not contain duplicates, that works. Thans!

radare:
conflicts

radare:
bash: before (15s) after(16s)
awk: before (1.5s) after(2s)

so, its slower

pelijah:
**bash**
_master_: 14s
_anal-block_: 11.5s

**awk**
_master_: 6s
_anal-block_: 5.3s

thestr4ng3r:
@radare gimme binaries

radare:
Yayy ill try today and see if that bottleneck was making everything slower

pelijah:
> 
> 
> **bash**
> _master_: 14s
> _anal-block_: 11.5s
> 
> **awk**
> _master_: 6s
> _anal-block_: 5.3s

This was GNU/Linux.
Today I have tested changes on Windows.
_Subject_:  `bash` from pancake.
_master_: 13.1s
_anal-block_: 11.2s

pelijah:
**Platform**: Windows 10
**Subject**: teamspeak client (18.7 MB; https://github.com/radareorg/cutter/issues/498#issuecomment-451615605)
**Results**:
```

|   Branch   | "aaa" | "aa;aac;aar" |
|------------|-------|--------------|
| master     | 324s  | 110s         |
| anal-block | 226s  | 74s          |
```

libr/anal/block.c
```diff
@@ -0,0 +1,579 @@
...
+	if (!all_in (unwrap (node->_rb.child[0]), addr, cb, user)) {
+		return false;
+	}
+	if (!all_in (unwrap (node->_rb.child[1]), addr, cb, user)) {
+		return false;
+	}
+	return true;
+}
+
+R_API bool r_anal_block_get_in(RAnal *anal, ut64 addr, RAnalBlockCb cb, void *user) {
```
ret2libc: this name is misleading, because `get` seems like you get a AnalBlock in return. I think it makes more sense to use `r_anal_block_foreach_in`. WDYT?

libr/anal/block.c
```diff
@@ -0,0 +1,579 @@
...
+	}
+	if (addr < node->addr + node->size) {
+		cb (node, user);
+	}
+	// This can be done more efficiently by building the stack manually
+	all_intersect (unwrap (node->_rb.child[0]), addr, size, cb, user);
+	all_intersect (unwrap (node->_rb.child[1]), addr, size, cb, user);
+}
+
+R_API void r_anal_blocks_get_intersect(RAnal *anal, ut64 addr, ut64 size, RAnalBlockCb cb, void *user) {
```
ret2libc: same here. Better `r_anal_blocks_foreach_interesect`?

libr/anal/block.c
```diff
@@ -0,0 +1,579 @@
...
+	RAnal *anal = bb->anal;
+	r_anal_block_ref (bb);
+	while (!r_list_empty (bb->fcns)) {
+		r_anal_function_block_remove (r_list_first (bb->fcns), bb);
+	}
+	r_anal_block_unref (bb);
+	r_anal_block_check_invariants (anal);
+}
+
+R_API void r_anal_block_set_size(RAnalBlock *block, ut64 size) {
```
ret2libc: If we set a new size of the RAnalBlock we should probably also update `op_pos`, no?

radare:
There are a bunch of bugs and crashes when messing with bbs and fcns (this crash maybe its not exatcly using this commands. Will come back on this later when i get a chance to touch my laptop)

af
afb-*
pd 10 # crash

Also if u try:

aaa
af-*
afb # bbs are still there

radare:
i can confirm the performance improvements , but i see many crashes and bugs here and there

```
$ r2 -i crash.r2 /bin/ls
Segmentation Fault

$ cat crash.r2
af
afb
afb-+
pd 10
af-*
pd 10
```

libr/anal/anal.c
```diff
@@ -192,18 +192,23 @@ R_API void r_anal_plugin_free (RAnalPlugin *p) {
 	}
 }
 
+void __block_free_rb(RBNode *node, void *user);
```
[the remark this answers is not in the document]
thestr4ng3r: I think @radare wanted to use `_` at some point but idk. I don't like them either.

libr/anal/block.c
```diff
@@ -0,0 +1,579 @@
...
+	r_anal_switch_op_free (block->switch_op);
+	r_list_free (block->fcns);
+	free (block->label);
+	free (block->op_pos);
+	free (block->parent_reg_arena);
+	free (block);
+}
+
+// TODO: this can be moved to unit tests later
+R_API void r_anal_block_check_invariants(RAnal *anal) {
```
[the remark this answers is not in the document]
thestr4ng3r: Ignore those two check functions, they are only for debugging and will be removed from the core code before merging.

libr/anal/block.c
thestr4ng3r: sgtm

libr/anal/block.c
thestr4ng3r: Depends. This is more a low-level call that only changes the size of the block while filling `op_pos` is more of an analysis.

thestr4ng3r:
Good catch, fixed.

libr/anal/block.c
ret2libc: If it is so, I don't expect it to be a public R_API. I think it is very easy to forget updating op_pos (or other stuff that depend on the size of the bb).

radare:
Me neither but this should be initialized by a public constructor instead of using that callback here and there

libr/anal/block.c
thestr4ng3r: No, there must be an interface like that for the kind of analysis that `fcn_recurse()` does. Otherwise multiple r_anal_op passes would be necessary to do on the same op.
A good way to avoid the possibility of forgetting to update it would be to make it lazy, but that is outside the scope of this pr.

thestr4ng3r:
what callback?

radare:
The rbtree bb free cb

thestr4ng3r:
I have no idea what part of the code you are referring to, we were discussing `r_anal_block_set_size()`

libr/anal/anal_ex.c
```diff
@@ -139,18 +139,17 @@ R_API RAnalBlock * r_anal_ex_get_bb(RAnal *anal, RAnalState *state, ut64 addr) {
 	if (!op || !r_anal_state_addr_is_valid (state, addr)) {
 		return NULL;
 	}
-	current_bb = r_anal_bb_new ();
+	current_bb = r_anal_ex_op_to_bb (anal, state, op);
```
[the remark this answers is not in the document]
thestr4ng3r: This is just a compatibility change, the `r_anal_ex_op_to_bb()` has been moved up from a few lines below.

radare:
i dont see much difference in analysis times comparing master and this branch. i think we are good in merging this, but i would prefer to have it splitted in separaet commits instead of a big one. but the sooner we merge it the sooner we can test it more, we have a bunch of infinite loops and segfaults in current analysis in master which i think can be fixed with this branch.

ret2libc:
I will try to review and test this today/tomorrow.

thestr4ng3r:
> I will try to review and test this today/tomorrow.

Please today if possible

radare:
merged! time to extend our testing nd continue impmroving the analysis! thanks @thestr4ng3r for the hard work contiuing my PR.

PR #15189, opened 2019-10-04
add 8051 assembler plugin

hmht:
written from scratch, so it's entirely dependent on rasm2 to do the heavy
lifting, and perhaps lacking in features?

8051 syntax didn't make things easy; ampersand and pound signs caused
integration problems, requiring the rasm and regressiontest code to be patched.

### DIFF
--- a/libr/asm/arch/8051/8051_ass.c
+++ b/libr/asm/arch/8051/8051_ass.c
@@ -1,0 +1,1545 @@
+/* radare - LGPL - Copyright 2009-2019 - hmht */
+#include"8051_ass.h"
+
+/*****************************************************************************\
+ *              Architecture
+ *
+ * File Contents:
+ * ## Section 1. Token parsers
+ * ## Section 2: some weird datastructure
+ * ## Section 3: token classifiers
+ * ## Section 4: Generic instruction emitters
+ * ## Section 5: Specific instruction parsing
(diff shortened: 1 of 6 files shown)


astuder:
Does this already support register addressing? If so, how to I have to enter it in r2?
`wa xrl a, @r0` throws an error as @ is a reserved character in r2

radare:
Quote the whole thing, also, do the same string without the @ sign will have a different meaning? If not i would like to support both or alternative syntax without special chars

astuder:
Without @ it would be xor of the registers a and r0. @r0 is equivalent to [r0] in other architectures.

hmht:
@-addressing is supported and must be quoted, aye. Although there's a bug in the db/asm testbench that doesn't quote the lines so they show it shows up as XX.

I can add [r0] as an alternative to @r0. Is that the de-facto syntax?

radare:
yeah i like [] better than @, one of the ideas of r2 is to ease readability and make somme common behaviours work as expected across all disassemblers, but having this as fallback, not to break default outputs

astuder:
I guess it depends from which side you are coming. IMO r2, while it can support other formats, should default to the syntax defined by the ISA. Otherwise, it will be very confusing when cross-checking with vendor documentation or other tools.

PR #15202, opened 2019-10-05
Fix macOS build with latest XCode ##build

nl5887:
Building fails on MacOS 10.14.6 with the following error, added the arch and platform options to the make files.

```
15:38:42 ld: Assertion failed: (0 && "unknown objc arch"), function doPass, file /Library/Caches/com.apple.xbs/Sources/ld64/ld64-512.4/src/ld/passes/objc.cpp, line 1747.
15:38:42 ld: warning: platform not specified
15:38:42 ld: warning: -arch not specified
15:38:42 ld: warning: ignoring file zip/librz.a, building for -unknown but attempting to link with file built for macOS-x86_64
15:38:42 0  0x10d730168  __assert_rtn + 127
15:38:42 1  0x10d73a105  ld::passes::objc::fixClassAliases(ld::Atom const*) (.cold.1) + 0
15:38:42 2  0x10d70b9b4  void ld::passes::objc::doPass<x86_64, true>(Options const&, ld::Internal&) + 0
15:38:42 3  0x10d6130aa  main + 868
15:38:42 A linker snapshot was created at:
15:38:42 	/tmp/librz.a.a.o-2019-08-24-163839.ld-snapshot
15:38:42 ld: Assertion failed: (0 && "unknown objc arch"), function doPass, file /Library/Caches/com.apple.xbs/Sources/ld64/ld64-512.4/src/ld/passes/objc.cpp, line 1747.
15:38:42 make[1]: *** [libr_shlr.a] Error 1
15:38:42 make: *** [install] Error 2
```

https://github.com/Homebrew/homebrew-core/pull/44517

### DIFF
--- a/mk/clang.mk
+++ b/mk/clang.mk
@@ -16,7 +16,12 @@ CFLAGS_DEBUG=-g
 
 ifeq ($(OSTYPE),darwin)
 ARCH=$(shell uname -m)
-PARTIALLD=ld -r -all_load -platform_version macos 10.14 10.14
+XCODE_VERSION=$(shell xcodebuild -version|grep Xcode|grep -o "[\.0-9]\+")
+XCODE_VERSION_MAJOR = $(word 1, $(subst ., ,$(XCODE_VERSION)))
+PARTIALLD=${LD} -r -all_load
+ifeq ($(XCODE_VERSION_MAJOR),11)
+PARTIALLD+=-arch ${ARCH} -platform_version macos 10.14 10.14
+endif
 #CFLAGS+=-arch ${ARCH}
 #LDFLAGS+=-arch ${ARCH}
 LDFLAGS_LIB=-dynamiclib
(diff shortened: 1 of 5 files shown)


nl5887:
Build failed due to incompatibilities between the MacOS (and/or sdk and xcode versions) version 10.13 (which is the default Travis uses, https://docs.travis-ci.com/user/reference/osx/) and later MacOS versions (10.14 and xcode 11.0). It is not sure to me which version of xcode introduced the platform_version option, I can add tests for xcode (`/usr/bin/xcodebuild -version`) version 11.0. 

Will it be good to introduce (or update) travis image to more recent macos version?

radare:
can you move this repeated code into a separate .mk and include it from the other files?

radare:
Also, the 10.14 is maybe too high, because i use it on macppc which is quite older and it builds and works fine. move this version number into a separate variable

nl5887:
Alright, moved the code to darwin.mk, using sdk and macos versions 10.5 and using variables instead.

radare:
Thanks! merging! i'll test iOS before the end of this week

PR #15208, opened 2019-10-06
WIP: Don't prematurely terminate fcn analysis on call to noret fcn

kazarmy:
https://github.com/radareorg/radare2-regressions/commit/86fdbad3628622a07d67b5143c645619f587f4e9


kazarmy:
Hmm probably better instead to make the jumptable code work for `bins/elf/ls.odd` too.

radare:
adding supprot for new jump tables is always fun .p

PR #15218, opened 2019-10-07
fix hashbang for lang plugins that do not set argc/argv

condret:


radare:
Can you add a test for your #!c thing ?

radare:
tired of waiting for travis

PR #15251, opened 2019-10-11
RFC: Fix cmd `drm`

koyaan:
I went down quite a rabbit hole trying to fix `drm` without arguments:
The problems is that at https://github.com/radareorg/radare2/blob/a4c145c70872c3f034158175f026f233e03b3eaf/libr/debug/dreg.c#L26 we check to see if registers are actually in a different arena than their type would indicate which is true for `R_REG_TYPE_XMM` they are in the fpu arena. This check only happens for types which are positive.

But at https://github.com/radareorg/radare2/blob/a4c145c70872c3f034158175f026f233e03b3eaf/libr/debug/p/native/linux/linux_debug.c#L771 we see that the type argument has dual-use indicating the registers should be printed if type is negative. 

So currently it was not possible to have relocated registers and also print them that's why `drm` just prints nothing right now.

This hacky PR is my attempt to work around this by adding a new function to the debug plugin `r_debug_native_reg_read_p` which receives the original type before relocation and a flag if we want to print or not.

I am sure there is a better way to do this, that's why I am opening this Request For Comments.

What we get from this is split for fpu vs. xmm registers in printing (enabled the arena debug output too):
```
[0x55a0eb946660]> drf
---- x86-64 ----
cwd = 0x037f  ; control   swd = 0x0000  ; status
ftw = 0x0000              fop = 0x0000
rip = 0x0000000000000000  rdp = 0x0000000000000000
mxcsr = 0x00001f80        mxcr_mask = 0x0000ffff
size = 0x00000200
mm0 = 0x0000000000000000 | st0 = 00000000000000000000 0.000000e+00 0.000000
mm1 = 0x0000000000000000 | st1 = 00000000000000000000 0.000000e+00 0.000000
mm2 = 0x0000000000000000 | st2 = 00000000000000000000 0.000000e+00 0.000000
mm3 = 0x0000000000000000 | st3 = 00000000000000000000 0.000000e+00 0.000000
mm4 = 0x0000000000000000 | st4 = 00000000000000000000 0.000000e+00 0.000000
mm5 = 0x0000000000000000 | st5 = 00000000000000000000 0.000000e+00 0.000000
mm6 = 0x0000000000000000 | st6 = 00000000000000000000 0.000000e+00 0.000000
mm7 = 0x0000000000000000 | st7 = 00000000000000000000 0.000000e+00 0.000000
[0x55a0eb946660]> drm
 req = 4 arena = 2 mask = 1c search = 10 
 request arena 4 found at arena 2
---- x86-64 ----
xmm0    = 11111111 22222222 33333333 44444444
xmm1    = 04030201 08070605 0c0b0a09 100f0e0d
xmm2    = 00000000 00000000 00000000 00000000
xmm3    = 00000000 00000000 00000000 00000000
xmm4    = 0000ff00 0000ff00 0000ff00 00000000
xmm5    = 00000000 00000000 00000000 00000000
xmm6    = 00000000 00000000 00000000 00000000
xmm7    = 00000000 00000000 00000000 00000000
xmm8    = 00000000 00000000 00000000 00000000
xmm9    = 00000000 00000000 00000000 00000000
xmm10   = 00000000 00000000 00000000 00000000
xmm11   = 00000000 00000000 00000000 00000000
xmm12   = 00000000 00000000 00000000 00000000
xmm13   = 00000000 00000000 00000000 00000000
xmm14   = 00000000 00000000 00000000 00000000
xmm15   = 00000000 00000000 00000000 00000000
```
All of this is Linux 64 only right now.


radare:
lgtm

libr/debug/dreg.c
```diff
@@ -34,8 +48,9 @@ R_API int r_debug_reg_sync(RDebug *dbg, int type, int write) {
 			int v = ((int)1 << i);
 			// skip checks on same request arena and check if this arena have inside the request arena type
 			if (n != i && (mask & v)) {
-				//eprintf(" req = %i arena = %i mask = %x search = %x \n", i, n, mask, v);
-				//eprintf(" request arena %i found at arena %i\n", i, n );
+				// TODO: comment this out again, debugging right now
+				eprintf(" req = %i arena = %i mask = %x search = %x \n", i, n, mask, v);
```
radare: missing space before ( and extra space before \n

libr/debug/dreg.c
```diff
@@ -64,7 +79,11 @@ R_API int r_debug_reg_sync(RDebug *dbg, int type, int write) {
 					return false;
 				}
 				//we have already checked dbg->h and dbg->h->reg_read above
-				size = dbg->h->reg_read (dbg, i, buf, bufsize);
+				if (have_get_reg_p)	{
+					size = dbg->h->reg_read_p(dbg, i, buf, bufsize, print, original_type);
```
radare: missing space before (

libr/debug/p/debug_native.c
```diff
@@ -18,6 +18,7 @@
 
 static int r_debug_native_continue (RDebug *dbg, int pid, int tid, int sig);
 static int r_debug_native_reg_read (RDebug *dbg, int type, ut8 *buf, int size);
+static int r_debug_native_reg_read_p (RDebug *dbg, int type, ut8 *buf, int size, char print, int arena);
```
radare: remove space before ( (also in all the other static definitions

libr/debug/dreg.c
```diff
@@ -21,6 +25,16 @@ R_API int r_debug_reg_sync(RDebug *dbg, int type, int write) {
 	if (!write && !dbg->h->reg_read) {
 		return false;
 	}
+
+	if(dbg->h->reg_read_p)	{
```
radare: missing space

libr/debug/dreg.c
```diff
@@ -7,6 +7,10 @@
 
 R_API int r_debug_reg_sync(RDebug *dbg, int type, int write) {
 	int i, n, size;
+ 	char have_get_reg_p = false;
+	int original_type = type;
+	char print = false;
+
 	if (!dbg || !dbg->reg || !dbg->h) {
 		return false;
```
radare: if it returns false, the return type should be bool

koyaan:
> i would prefer to have a single reg_read function. imho the printing sohuld be done in the core module, and debug api may return a list of, a less intrussive option would be to add some new fields in RDebug to pass print + arena every time.
> 
> we must try to do it being portable. so printing is not the same asretrieving the data

Do you mean we should move all the printing stuff into `r_debug_reg_list` ? To do that would need some major refactoring also on the windows side

koyaan:
I made another attempt at  https://github.com/koyaan/radare2/commit/be864fba2ba2cb85460ec3d6b72047a26b5ff82b
Because i realized `r_debug_reg_list` should already be able to print xmm registers
This gives only the following output though:
```
xmm0 = 0x44444444333333332222222211111111
xmm1 = 0x100f0e0d0c0b0a090807060504030201
xmm2 = 0x00000000000000000000000000000000
xmm3 = 0x000000000000ff00000000ff00000000
xmm4 = 0x000000000000ff000000ff000000ff00
xmm5 = 0x00000000000000000000000000000000
xmm6 = 0x00000000000000000000000000000000
xmm7 = 0x00000000000000000000000000000000
``
maybe it would be an idea to support `drmq` `drmd` etc without register argument but i guess that would need abuse of `rad` parameter

koyaan:
Superseded by https://github.com/radareorg/radare2/pull/15354

Issue #15250, opened 2019-10-11
REGRESSION. x86 NEG instruction wrongly emulated


radare:
* add test (is the 2nd time we break it)

```
IF DEST = 0                                                                                                                                                                                                                                   
     THEN CF ← 0;                           rs                                                                                                                                                                                                 
     ELSE CF ← 1;                               es                                                                                                                                                                                             
 FI;                                                                                                                                                                                                                                           
 DEST ← [– (DEST)]                                   word
```

```
mov ebx, number
neg
jz jnz


```

basically cf can only be 0 if the argument is zero
so one test would be to aer ebx,0

test:
```
wa neg ebx
aei aeim
aes
and aer cf
```

so how do you compare something with zero
atm its comparing > 0 (wrong)
we compare signed
so if i have a negative value cf will also be 0 (wrong)
cause for neg you need to check val for 0

condret:
this what you get, when ppl who believe in the existance of signed numbers in a computer generate esil expressions ...  here is the r2r pr https://github.com/radareorg/radare2-regressions/pull/1982

PR #15252, opened 2019-10-11
anal_x86_cs fix esil expr for neg instruction

condret:
hm, not sure, but to me this looks like cf is always "zf,!,cf,:=" at the end, so maybe that would be a bit faster and produce slightly shorter esil, but not completly sure about this atm ... anyway, this pr should fix the #15250

I didn't update my own r2r repo, so expect red travis. I havent got time to update that now


radare:
can you add a test?

condret:
yes, after sleep

PR #15254, opened 2019-10-11
Json output for afo and afd

xermicus:
```
:> afoj
{"address":"0x00013adc"}
:> afdj
{"name":"fcn.00013adc"}
```

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -2850,15 +2850,30 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
 			addr = core->offset;
 		}
 		RAnalFunction *fcn = r_anal_get_fcn_in (core->anal, addr, 0);
-		if (fcn) {
-			if (fcn->addr != addr) {
-				r_cons_printf ("%s + %d\n", fcn->name,
-						(int)(addr - fcn->addr));
-			} else {
-				r_cons_println (fcn->name);
+		if (input[2] == 'j') { // afdj
+			PJ *pj = pj_new ();
+			if (!pj) {
+				return false;
 			}
+			pj_o (pj);
+			if (fcn) {
+				pj_ks (pj, "name", fcn->name);
+				pj_ki (pj, "offset", (int)(addr - fcn->addr));
+			}
+			pj_end (pj);
+			r_cons_println (pj_string (pj));
+			pj_free (pj);
 		} else {
-			eprintf ("afd: Cannot find function\n");
+			if (fcn) {
+				if (fcn->addr != addr) {
+					r_cons_printf ("%s + %d\n", fcn->name,
+							(int)(addr - fcn->addr));
+				} else {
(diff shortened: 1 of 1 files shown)


libr/core/cmd_anal.c
```diff
@@ -2841,21 +2841,33 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
+				r_cons_printf ("%s + %d", fcn->name,
 						(int)(addr - fcn->addr));
 			} else {
-				r_cons_println (fcn->name);
+				r_cons_print (fcn->name);
+			}
+			if (json) {
+				r_cons_println ("\"}");
+			} else {
+				r_cons_println ("");
```
radare: there's r_print_newline ()

libr/core/cmd_anal.c
```diff
@@ -2962,6 +2974,14 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
 			eprintf ("Usage: afo[?sj] ([name|offset])\n");
 			break;
 		case 'j':
+			{
+				ut64 addr = core->offset;
+				RAnalFunction *fcn = r_anal_get_fcn_in (core->anal, addr, R_ANAL_FCN_TYPE_NULL);
+				if (fcn) {
+					r_cons_printf ("{\"address\":\"0x%08" PFMT64x "\"}\n", fcn->addr);
```
radare: there can be more than one function in the given offset, i have opened a pr to work on that, so this may change to iterate a list, also will be good to use the pj api instead of a raw println, addresses are usually printed in decimal, not as a string.

libr/core/cmd_anal.c
```diff
@@ -2962,6 +2974,14 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
 			eprintf ("Usage: afo[?sj] ([name|offset])\n");
 			break;
 		case 'j':
+			{
+				ut64 addr = core->offset;
```
radare: unnecessary variable

libr/core/cmd_anal.c
```diff
@@ -2841,21 +2841,33 @@ static int cmd_anal_fcn(RCore *core, const char *input) {
...
 		} else {
 			addr = core->offset;
+			if (input[2] == 'j') {
+				json = true;
+			}
 		}
 		RAnalFunction *fcn = r_anal_get_fcn_in (core->anal, addr, 0);
 		if (fcn) {
+			if (json) {
+				r_cons_printf ("{\"name\":\"");
```
radare: use the pj api. fcn->name can contain special chars that will break the json. the pj api takes care of this

xermicus: thanks!

libr/core/cmd_anal.c
xermicus: Ok I implemented it now using pj API (can remove again if we rather want to wait until your PR is done, but the feature would be nice to have in my script anyways)

PR #15256, opened 2019-10-12
Fix formatting for r_bin_list_archs

GustavoLCR:
* obj->obj_size is 64 bits
* offset is unsigned


XVilka:
Now seems the only difference is the UTF-8 lines (in AppVeyor builds):
```
 nth paddr         size vaddr        vsize perm name
----------------------------------------------------
+―――――――――――――――――――――――――――――――――――――――――――――――――――
 0   0x00000600   0x600 0x00401000  0x1000 -r-x CODE
 1   0x00000c00   0x200 0x00402000  0x1000 -rw- DATA
```

PR #15268, opened 2019-10-13
Add current thread field to dptj

yossizap:
While fixing JSON support for dpt I didn't notice that the JSON packing implementation was missing a field that was [available](https://github.com/radareorg/radare2/blob/82bff79820abc796c43cf46de6359f4eed3e5b9b/libr/debug/pid.c#L56) in the printed version. This small patch takes care of it.

### DIFF
--- a/libr/debug/pid.c
+++ b/libr/debug/pid.c
@@ -86,7 +86,7 @@ R_API int r_debug_thread_list(RDebug *dbg, int pid, char fmt) {
 			switch (fmt) {
 			case 'j':
 				pj_o (j);
-				pj_ki (j, "current", dbg->tid == p->pid? true: false);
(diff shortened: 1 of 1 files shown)


libr/debug/pid.c
```diff
@@ -86,6 +86,7 @@ R_API int r_debug_thread_list(RDebug *dbg, int pid, char fmt) {
 			switch (fmt) {
 			case 'j':
 				pj_o (j);
+				pj_ki (j, "current", dbg->tid == p->pid? true: false);
```
radare: use pj_kb (j, "current", dbg->tid == p->pid);

radare: its a bool

PR #15273, opened 2019-10-14
Move UTF-8 setting logic to scr.utf8 config callback on Windows

GustavoLCR:
This should fix tests on Appveyor

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -2278,7 +2278,7 @@ static bool cb_utf8(void *user, void *data) {
 #if UNICODE
 	if (IsValidCodePage (CP_UTF8)) {
 		if (!SetConsoleOutputCP (CP_UTF8) || !SetConsoleCP (CP_UTF8)) {
-			r_sys_perror ("r_cons_new");
+			r_sys_perror ("cb_utf8");
 		}
 	} else {
 		R_LOG_WARN ("UTF-8 Codepage not installed.\n");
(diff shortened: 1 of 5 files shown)


radare:
uhm. why that? isnt the env properly set when cons_init is called? i would prefer to have this code in rcons

GustavoLCR:
> uhm. why that? isnt the env properly set when cons_init is called? i would prefer to have this code in rcons

@radare the problem is that if utf8 was available at all, r2 would **always** change the active codepage to it (even with `scr.utf8=0`), affecting the other tools output.

With this it will only automatically set `scr.utf8=1` if utf8 codepage is already active when the tools launches, and if the user sets `scr.utf8` r2 will change the codepage accordingly.

XVilka:
I agree this approach is better. LGTM.

radare:
despite the thing works as you expect, i dont like having code that checks for console stuff in core, when this was already in cons, which was the right place

libr/include/r_cons.h
```diff
@@ -822,6 +822,7 @@ R_API int r_cons_w32_print(const ut8 *ptr, int len, bool vmode);
 R_API int r_cons_win_printf(bool vmode, const char *fmt, ...);
 R_API int r_cons_win_eprintf(bool vmode, const char *fmt, ...);
 R_API int r_cons_win_vhprintf(DWORD hdl, bool vmode, const char *fmt, va_list ap);
+R_API void r_cons_win_set_cp(bool utf8);
```
radare: just wondering.. is there any reason why the win api is separated instead of being abstracted into a generic functionality for all platforms? .. something to think about and do in a separate pr

libr/core/cconfig.c
```diff
@@ -2273,6 +2273,9 @@ static bool cb_utf8(void *user, void *data) {
 	RCore *core = (RCore *) user;
 	RConfigNode *node = (RConfigNode *) data;
 	core->cons->use_utf8 = node->i_value;
+#if __WINDOWS__
+	r_cons_win_set_cp (core->cons->use_utf8);
```
radare: what about naming it r_cons_set_utf8(), so using the setter/getter pattern instead of setting the use_utf8 variable by hand

PR #15284, opened 2019-10-15
(elf) avoid truncating binary length to 32 bits

jjgriego:
I was trying r2 out on a very large binary (5.6G) [debug symbols take up a lot of space!] and noticed that rabin2 was unable to parse most of the symbol table.

This fixed the issue for me and looks sane enough. I couldn't track the blame on the 32-bit cast far enough to find a reason not to get rid of it. Let me know if that sounds Wrong and Bad.


kosigz-lyft:
😮

PR #15287, opened 2019-10-16
Fixed ptrace 'Operation not permitted'

### LINKED ISSUE #14506: radare2 : debug error
OS : Kali Linux
whoami : root
radare2 info ( radare2 3.7.0-git 22212 @ linux-x86-64 git.3.6.0-98-g681e8c1a8
commit: 681e8c1a89d6e979a1c5701c7fb1d117106d536c build: 2019-07-07__11:04:49 )
ptrace : ptrace (PT_ATTACH): Operation not permitted

ERROR

[0x00001165]> ood AAAAA-BBBB-CCCCC
Process with PID 27857 started...
File dbg:///root/binary/0x05/license_2  AAAAA-BBBB-CCCCC reopened in read-write mode
= attach 27857 27857
ptrace (PT_ATTACH): Operation not permitted
27857
[0x7f31d587c090]> 

ptrace more info : {
cat /proc/sys/kernel/yama/ptrace_scope 
0
}

here is the example 

``#include <string.h>
#include <stdio.h>

int main(int argc, char *argv[]) {
if(argc==2) {
printf("Checking : %s\n", argv[1]);
int sum=0;
for(int i=0; i<strlen(argv[1]); i++){
sum += (int)argv[1][i];

	}
	if(sum==916) {
		printf("Access Granted!\n");
	} else {
		printf("WRONG!\n");
	}
} else {
	printf("Usage: <key>\n");
}
return 0;

}
`

yossizap:
This issue is a result of running PT_ATTACH on a forked PID when you're only supposed to PT_ATTACH remote pids. linux_set_options probably always worked for you when testing in fork mode but it wouldn't work in remote attachment since you have to attach first + there's was error handling there.
When using fork_and_ptraceme_for_linux the child sets himself up to be traced for you.

I used GETSIGINFO to check the attachment status since it doesn't have any consequences(same result even if you run it multiple times) and always works on an attached process.

Also went ahead and cleaned up the code around the changes.

closes #14506

### DIFF
--- a/libr/debug/p/native/linux/linux_debug.c
+++ b/libr/debug/p/native/linux/linux_debug.c
@@ -415,10 +415,10 @@ static bool linux_attach_single_pid(RDebug *dbg, int ptid) {
 	// Attaching to a process that has already been started with PTRACE_TRACEME.
 	// sets errno to "Operation not permitted" which may be misleading.
 	// GETSIGINFO can be called multiple times and would fail without attachment.
-	if (-1 == r_debug_ptrace(dbg, PTRACE_GETSIGINFO, ptid, NULL,
-		(r_ptrace_data_t)&sig)) {
-		if (-1 == r_debug_ptrace (dbg, PTRACE_ATTACH, ptid, NULL,
-			(r_ptrace_data_t)NULL)) {
+	if (r_debug_ptrace (dbg, PTRACE_GETSIGINFO, ptid, NULL,
+		(r_ptrace_data_t)&sig) == -1) {
+		if (r_debug_ptrace (dbg, PTRACE_ATTACH, ptid, NULL,
+			(r_ptrace_data_t)NULL) == -1) {
 			perror ("ptrace (PT_ATTACH)");
 			return false;
 		}
--- a/libr/debug/p/native/linux/linux_debug.c
+++ b/libr/debug/p/native/linux/linux_debug.c
@@ -54,14 +54,14 @@ char *linux_reg_profile (RDebug *dbg) {
 
 static void linux_detach_all (RDebug *dbg);
 static char *read_link (int pid, const char *file);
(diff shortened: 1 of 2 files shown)


libr/debug/p/native/linux/linux_debug.c
```diff
@@ -408,10 +409,26 @@ static int linux_stop_process(int pid) {
...
-	linux_set_options (dbg, ptid);
-	return r_debug_ptrace (dbg, PTRACE_ATTACH, ptid, NULL,
-		(r_ptrace_data_t)(size_t)NULL);
+static bool linux_attach_single_pid(RDebug *dbg, int ptid) {
+	siginfo_t sig = {0};
+	// Safely check if the PID has already been attached to avoid printing errors.
+	// Attaching to a process that has already been started with PTRACE_TRACEME.
+	// sets errno to "Operation not permitted" which may be misleading.
+	// GETSIGINFO can be called multiple times and would fail without attachment.
+	if (-1 == r_debug_ptrace(dbg, PTRACE_GETSIGINFO, ptid, NULL,
```
radare: missing space before ( and piut the == -1 in the right side of the statement

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -408,10 +409,26 @@ static int linux_stop_process(int pid) {
...
+		(r_ptrace_data_t)&sig)) {
+		if (-1 == r_debug_ptrace (dbg, PTRACE_ATTACH, ptid, NULL,
```
radare: same here

yossizap: Updated

libr/debug/p/native/linux/linux_debug.c
yossizap: Updated

PR #15288, opened 2019-10-16
buf_ref: make sure the reference buffer does not exceed the parent size

ret2libc:
r2r PR: https://github.com/radareorg/radare2-regressions/pull/1985


ret2libc:
Mmm, it doesn't seem like the errors are because of these changes. Is travis working in other PRs?

PR #15291, opened 2019-10-16
Add support for hex escaped litterals in regular expressions

the-remmer:
Example: /e /(\x77\x30\x30\x74){2}/ searches for "w00tw00t"


radare:
can you add a test in r2r?

radare:
thanks!

PR #15300, opened 2019-10-17
Replace r_str_const* with RStrConstPool

thestr4ng3r:


thestr4ng3r:
These are many commits, but I think they still should NOT be squashed because this makes bisecting easier in case of breakage.

ret2libc:
Probably stupid question, but what's the purpose of `r_str_const*`?

thestr4ng3r:
It keeps a set of constant strings and if you ask for one, it will either return the matching existing one or create a new one and return that.
The purpose is to avoid strduping the same string over and over again if it is used many times, like in `RAnalFunction.cc`.
The original `r_str_const()` used this global `__consts` which is of course bad if there are multiple RCores.

PR #15321, opened 2019-10-21
update 8051 anal plugin

hmht:
#### handle 8051 ANAL_OP DISASM
The 8051 assembler has been R_APId, #included and called to fill
op->mnemonic, which stops a bunch of warnings from appearing
whenever a bunch of disassembly appears on the screen.

The disassembler is called because of comments in r_anal.h: op->mnemonic
should contain the entire disassembly, not just the mnemonic.
Here's hoping the mnemonics and arguments will get split eventually.

#### typdef some RAnal enums
In trying to make my analysis push out more info, it's difficult to
understand what all these ints mean, while ACTUALLY they should be
filled with enum values. By using the enum names, that's made clear
immediately.

r2's style is typedef over `enum name`, so that's what I did.

the typedef-instead-of-int I added here isn't consistently propagated,
and has caused warnings about unhandled cases-in-switch, at least some
of which should just get a default: case added, but I'd rather leave it
to the domain experts, or my future self when I become that domain
expert.

#### add more basic info
cycles, cond, eob, nopcode, family, id


radare:
Screenshot 2019-10-21 at 15 44 12

libr/anal/p/anal_8051.c
```diff
@@ -881,6 +881,10 @@ static ut32 map_direct_addr(RAnal *anal, ut8 addr) {
 static int i8051_op(RAnal *anal, RAnalOp *op, ut64 addr, const ut8 *buf, int len, RAnalOpMask mask) {
 	set_cpu_model (anal, false);
 
+	if (mask & R_ANAL_OP_MASK_ESIL) {
```
radare: Not the same. Thats wrong

hmht: I have concluded that we don't actually have a MASK_VAL implementation yet.

radare:
if the code calling ranalop needs VAL to work, it should be passing |VAL in the mask

marcograss:
this broke the build on macOS btw

radare:
I know

PR #11335, opened 2018-08-31
WIP: signal -> sigaction

nsajko:
The signal function is underspecified and deprecated in favor of
sigaction, make the replacement.

Also added <signal.h> includes wheremissing, and fixed a "__UNIX_" to
"__UNIX__".

PR #15322, opened 2019-10-21
WIP: Implement signal using sigaction

dokthar:
signal replacement with sigaction has already been proposed in the PR #11335
This is for now only a replacement for signal using sigaction.
The r_sys_sigaction API can be modified if more sigaction features are required.


radare:
i think we can close https://github.com/radareorg/radare2/pull/11335, and keep your PR up

radare:
can you rebase now that master is green? and specify what's WIP?

libr/util/sys.c
```diff
@@ -444,20 +444,19 @@ R_API int r_sys_crash_handler(const char *cmd) {
...
-	sigaddset (&sigact.sa_mask, SIGQUIT);
-	sigaddset (&sigact.sa_mask, SIGHUP);
-
-	sigaction (SIGINT, &sigact, (struct sigaction *)NULL);
-	sigaction (SIGSEGV, &sigact, (struct sigaction *)NULL);
-	sigaction (SIGBUS, &sigact, (struct sigaction *)NULL);
-	sigaction (SIGQUIT, &sigact, (struct sigaction *)NULL);
-	sigaction (SIGHUP, &sigact, (struct sigaction *)NULL);
+
+	sigemptyset (&mask);
```
radare: this is unix specific, whiñe the r_sys* apis are suposed to be generic/portable. i wont mix them

libr/util/sys.c
```diff
@@ -1263,3 +1262,20 @@ R_API const char *r_sys_prefix(const char *pfx) {
...
+
+	return sigaction (sig, &sigact, NULL);
+}
+
+R_API int r_sys_sigaction(int sig, void (*handler) (int)) {
+	sigset_t mask;
+
+	sigemptyset (&mask);
+
+	return r_sys_sigaction_mask(sig, handler, mask);
```
radare: misssing space before (

libr/include/r_util/r_sys.h
```diff
@@ -29,6 +29,8 @@ typedef struct _os_info {
 	int compilation;
 } os_info;
 
+R_API int r_sys_sigaction(int sig, void (*handler) (int));
+R_API int r_sys_sigaction_mask(int sig, void (*handler) (int), sigset_t mask);
```
thestr4ng3r: `sigset_t` is platform-specific. These functions should be inside an `#if __UNIX__`

radare:
solve conflicts and rebase

dokthar:
Detecting if sigaction is available is really a pain. Relying only on `__UNIX__` being defined seem a bit weak to me as sigaction will not be defined if using std c89 or even c99...
Then I started looking for ways to detect if the libc support sigaction and there is many way to do so, but none that will work on every libc.
The glibc have it's feature test macro but I not sure it will work with others.

libr/util/sys.c
radare: this^

libr/util/sys.c
```diff
@@ -411,20 +439,19 @@ R_API int r_sys_crash_handler(const char *cmd) {
...
-	sigaction (SIGHUP, &sigact, (struct sigaction *)NULL);
+
+	sigemptyset (&mask);
+	sigaddset (&mask, SIGINT);
+	sigaddset (&mask, SIGSEGV);
+	sigaddset (&mask, SIGBUS);
+	sigaddset (&mask, SIGQUIT);
+	sigaddset (&mask, SIGHUP);
+
+	r_sys_sigaction (SIGINT, signal_handler, mask);
```
radare: my idea for this waas to have an r_sys api to build this sigset, maybe using vararg like `r_sys_sigaction_mask (SIGINt, SIGSEGV,SIGBUS,SIGQUIT,SIGHUP, -1);`

note thaat the sigset_t type should be overriden by r2 in order to provide a generic type that can compile on non-unix systems, the ideaa is to have a generic api that works across all OS, not a unix wrapper

libr/util/sys.c
```diff
@@ -411,20 +439,19 @@ R_API int r_sys_crash_handler(const char *cmd) {
...
+	r_sys_sigaction (SIGSEGV, signal_handler, mask);
```
radare: as long as we are setting the same signal handler to all of them aybe we can just have a asingle function that takes an list of signals and the handler. like this:

```
r_sys_sigaction (signal_handler, SIGINT, SIGSEGV,SIGBUS,SIGQUIT,SIGHUP, -1);
```

this single function can hide the sigset_t need to be public/wrapped so unless im missing an usecase for this i think this will be cleaner

dokthar: Yes this seems to be a good idea. But what do you think of using an array instead of va_args ?
```
r_sys_sigaction (signal_handler, (int []){SIGINT, SIGSEGV, SIGBUS, SIGQUIT, SIGHUP, -1});
```

dokthar:
changes since previous version:
r_sys_sigaction nows takes an array of signal numbers (instead of one), all the signals will be added to the mask (this could be improved), and the handler will be registered for each signal number.

PR #15323, opened 2019-10-21
Fix the way linux debugging options are set

ret2libc:
In linux_attach_single_pid there is a check to avoid calling PT_ATTACH
multiple times on an already traced pid, which is an operation that
should be done only one time. The old code supposedly tried to set the
options only the first time you attach to a process, however when
debugging a process through the io_ptrace plugin, the process uses
TRACEME and the code never set the options of ptrace to intercept the
process before dying, amongst other things.

This patch tries to address the issue by using a boolean in RDebug that
keep tracks of whether the options were set or not for the process being
debugged.

### DIFF
--- a/libr/debug/p/native/linux/linux_debug.c
+++ b/libr/debug/p/native/linux/linux_debug.c
@@ -251,7 +251,7 @@ bool linux_set_options(RDebug *dbg, int pid) {
 	return true;
 }
 
-static void linux_detach_all (RDebug *dbg) {
+static void linux_detach_all(RDebug *dbg) {
 	RList *th_list = dbg->threads;
 	if (th_list) {
 		RDebugPid *th;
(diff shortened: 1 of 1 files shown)


ret2libc:
cc @yossizap

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -227,6 +227,10 @@ int linux_step(RDebug *dbg) {
 }
 
 bool linux_set_options(RDebug *dbg, int pid) {
+	if (dbg->options_set) {
+		return true;
+	}
```
yossizap: This change will prevent updating options for a given PID after initial attachment.

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -428,12 +433,12 @@ static bool linux_attach_single_pid(RDebug *dbg, int ptid) {
 			perror ("ptrace (PT_ATTACH)");
 			return false;
 		}
-
-		if (!linux_set_options (dbg, ptid)) {
-			return false;
-		}
```
yossizap: You won't be able to set thread options this way. I would suggest to restore this line for thread option setting and to add the following code specifically for the main, potentially TRACEME'd, tid

    else if (ptid == dbg->main_pid && !dbg->options_set) {
        if (!linux_set_options (dbg, ptid)) {
            return false;
        }
    }

ret2libc: Mmm what about we just always call linux_set_options no matter what? Is there any issue with that?

radare:
please rebase

libr/debug/p/native/linux/linux_debug.c
yossizap: I looked into it and there doesn't seem to be a side effect to constantly changing thread options, that would be the easiest solution.

radare:
ping @ret2libc

libr/debug/p/native/linux/linux_debug.c
ret2libc: ok perfect I will go with the easiest solution for now, as it doesn't have side effects and it worked/works.

ret2libc:
> ping @ret2libc

On it, I will fix later today hopefully.

ret2libc:
Rebased and done a change to always call linux_set_options, no matter what.
@yossizap please review when you can.

PR #15326, opened 2019-10-22
For utf-8 locale, check LC_ALL, LC_CTYPE and LANG in sequence instead

kazarmy:
Should fix rabin2 tests, though breakage might get worse before it gets better.


kazarmy:
Env var sequence [follows glibc](https://linux.die.net/man/3/setlocale).

radare:
<3 thank you!!

PR #15347, opened 2019-10-23
r_8051_disas should not be R_API

ret2libc:


ret2libc:
This should fix the build error on appveyor, let's see.

PR #15349, opened 2019-10-24
A bit of refactoring in RBin

ret2libc:
* introduce r_bin_open_buf
* fix r_bin_file_delete to use bf->id and not fd. This was used
  inconsistently and the function was just wrong.
* rewrite r_bin_reload to just reuse the existing bf->buf
* fix some memory leaks


radare:
Neat 👍

PR #15358, opened 2019-10-25
Fixed opcodes not being associated with the current thread in gdbr

yossizap:
Prior to these fixes, running step/continue and their variation would switch back to the original pid.

### DIFF
--- a/shlr/gdb/src/gdbclient/core.c
+++ b/shlr/gdb/src/gdbclient/core.c
@@ -100,8 +100,6 @@ static int gdbr_connect_lldb(libgdbr_t *g) {
 
 int gdbr_connect(libgdbr_t *g, const char *host, int port) {
 	const char *message = "qSupported:multiprocess+;qRelocInsn+;xmlRegisters=i386";
-	RStrBuf tmp;
-	r_strbuf_init (&tmp);
 	int ret, i;
 	if (!g || !host) {
 		return -1;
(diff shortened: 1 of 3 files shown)


shlr/gdb/src/gdbclient/core.c
```diff
@@ -689,11 +689,16 @@ int gdbr_write_memory(libgdbr_t *g, ut64 address, const uint8_t *data, ut64 len)
 }
 
 int gdbr_step(libgdbr_t *g, int tid) {
-	char thread_id[64] = {0};
+	RStrBuf tmp;
+	char thread_id[64] = { 0 };
 	if (tid <= 0 || write_thread_id (thread_id, sizeof (thread_id) - 1, g->pid, tid,
 			     g->stub_features.multiprocess) < 0) {
+		r_strbuf_init (&tmp);
+		snprintf (tmp.buf, sizeof (tmp.buf) - 1, "Hc%d", tid);
```
radare: use the strbuf apis to modify the rstrbuf struct, dont touch the internal fields like this because it may lead to crashes or issues between versions

radare: well in fact you dont need an strbuf here at all, you can use `send_vcont (g, sdb_fmt ("Hc%d", tid))` instead of allocating things that never get freed or use internal private fields of unnecessary structs

yossizap: Alright thanks for the tip. Went ahead and also fixed this in other parts of the file that gave me the idea.

PR #15362, opened 2019-10-25
On AArch64 there is a new special symbol $x

ret2libc:


radare:
lgtm, ill fix the tests and add a test for the r2baby bin

thank you!

PR #15363, opened 2019-10-25
Added a way to break while in linux_dbg_wait and fixed dpk

yossizap:
Added break functionality to around waitpid since there's no normal way to replace waitpid with a timeout-able function without using WHOHANG + sleep which isn't ideal. This change was required by cutter to suspend debug tasks that were ran with `&`.

dpk just called pid 0 so I went ahead and fixed it. Regarding `/* XXX: not for threads? signal is for a whole process!! */` - you have tkill for that, I would suggest restoring this functionality in a future PR for debugging.

### DIFF
--- a/libr/core/cmd_debug.c
+++ b/libr/core/cmd_debug.c
@@ -1075,7 +1075,7 @@ static void cmd_debug_pid(RCore *core, const char *input) {
 			ptr = strchr (ptr, ' ');
 			sig = ptr? atoi (ptr + 1): 0;
 			eprintf ("Sending signal '%d' to pid '%d'\n", sig, pid);
-			r_debug_kill (core->dbg, 0, false, sig);
+			r_debug_kill (core->dbg, pid, false, sig);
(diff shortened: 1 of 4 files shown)


radare:
It would be awesome to have more tests for the debugger, are you in the mood to add some more tests for the thjngs you are fixing?

Thanks!

yossizap:
Sure, was thinking of the same thing since so much is broken. I have a small redesign in mind and I'll look around radare2-regressions to see how it can be automated once I start working on it.

Is this PR blocked by additional tests or can those wait? Currently trying to clear small issues for Cutter's debugger.

radare:
i was about to squash to add the ##debug tag in the commitm memssage, but as long as it's 2 meaningful commits it will be good to have them by separate. can you rewrite the commit messages to include it? this way it will be included in the release notes

yossizap:
Sorry but please hold the merge for a bit. I found an issue when running threads, looking into this right now.

PR #15364, opened 2019-10-26
Improve switch-statements in disassembly ##disasm

radare:
* Fix default case flag name

before:
Screenshot 2019-10-26 at 07 14 56

after:
Screenshot 2019-10-26 at 07 11 55

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -5,6 +5,8 @@
 #define HASRETRY 1
 #define HAVE_LOCALS 1
 #define DEFAULT_NARGS 4
+#define FLAG_PREFIX "; "
+// #define FLAG_PREFIX ";-- "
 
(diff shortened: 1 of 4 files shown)


libr/core/disasm.c
```diff
@@ -2172,6 +2172,31 @@ static int flagCmp(const void *a, const void *b) {
 	return strcmp (fa->name, fb->name);
 }
 
+static void __preline_flag(RDisasmState *ds, RFlagItem *flag) {
+	ds_newline (ds);
+	ds_begin_line (ds);
+	ds_pre_line (ds);
```
kazarmy: One day there should proper documentation as to how to begin a line and how to print comments, but right now the tests will have to do.

radare:
Yep disasm.c needs a lot of love still

XVilka:
New one is better, agree

radare:
hopefully ready to merge after fixing all the tests

PR #15366, opened 2019-10-26
Fix issues in windows thread switching

yossizap:
Previously `!=pid <pid>` attempted to OpenProcess even though the main pid is already opened by __open and the fact that re-opening the main pid wouldn't do anything. This way it attaches to new threads when called by r_debug_select.

### DIFF
--- a/libr/debug/p/native/windows/windows_debug.c
+++ b/libr/debug/p/native/windows/windows_debug.c
@@ -878,7 +878,8 @@ int w32_dbg_wait(RDebug *dbg, int pid) {
 					return -1;
 				}
 				if (!__is_thread_alive (dbg, dbg->tid)) {
-					if (!w32_select (dbg, dbg->pid, dbg->tid)) {
+					ret = w32_select (dbg, dbg->pid, dbg->tid);
+					if (ret == -1) {
 						return R_DEBUG_REASON_DEAD;
 					}
 				}
@@ -1050,12 +1051,12 @@ int w32_continue(RDebug *dbg, int pid, int tid, int sig) {
 
 	// Don't continue with a thread that wasn't requested
 	if (dbg->tid != tid) {
-		return false;
+		return -1;
 	}
 
 	if (breaked) {
 		breaked = false;
-		return false;
+		return -1;
 	}
 	w32dbg_wrap_instance *inst = rio->inst;
 	inst->params->type = W32_CONTINUE;
@@ -1066,7 +1067,7 @@ int w32_continue(RDebug *dbg, int pid, int tid, int sig) {
 	if (!w32dbgw_intret (inst)) {
 		w32dbgw_err (inst);
 		r_sys_perror ("w32_continue/ContinueDebugEvent");
-		return false;
+		return -1;
 	}
 
 	return tid;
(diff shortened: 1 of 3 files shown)


libr/io/p/io_w32dbg.c
```diff
@@ -180,15 +180,14 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
...
 		if (cmd[3] == ' ') {
 			int pid = atoi (cmd + 3);
-			if (pid > 0 && pid != iop->pi.dwProcessId) {
-				iop->pi.hProcess = OpenProcess (PROCESS_ALL_ACCESS, false, pid);
-				if (iop->pi.hProcess) {
-					iop->pi.dwProcessId = iop->pi.dwThreadId = pid;
+			if (pid > 0 && pid != iop->pi.dwThreadId && pid != iop->pi.dwProcessId) {
+				iop->pi.hThread = OpenThread (PROCESS_ALL_ACCESS, FALSE, pid);
+				if (iop->pi.hThread) {
+					iop->pi.dwThreadId = pid;
```
GustavoLCR: don't manually change the `iop->pi.dwThreadId`. w32_continue needs it to always be the last tid that sent an event.

yossizap: In that case, wouldn't it be wise to also change w32_continue to work with the tid it is given? Otherwise w32_continue will always continue from whichever thread dbg_wait sets globally even though someone called `dc` after switching threads. `r_debug_native_continue` and `r_debug_continue_kill` don't seem to pay any attention to this detail.

GustavoLCR: When calling `ContinueDebugEvent`, `dwThreadId` must be the one from the last debug event received. It doesn't mean the next debug event will be from this thread.

Calling `ResumeThread` should be enough to select and signal a thread to continue , thats what `w32_select` does, and `SuspendThread` is also called in all other threads if `dbg.threads = true`

libr/debug/p/native/windows/windows_debug.c
```diff
@@ -1035,14 +1035,23 @@ int w32_continue(RDebug *dbg, int pid, int tid, int sig) {
...
-		return tid;
+		return false;
 	}
+
 	if (!ContinueDebugEvent (rio->pi.dwProcessId, rio->pi.dwThreadId, continue_status)) {
 		r_sys_perror ("w32_continue/ContinueDebugEvent");
 		return false;
 	}
+
 	return tid;
```
radare: return bool and int at the same time? this is wrong

yossizap: There was a call chain depending on this rv, fixed now.

libr/io/p/io_w32dbg.c
radare: yep. it must populate that "current pid" to r2 itself

radare:
Conflicts

libr/io/p/io_w32dbg.c
yossizap: As in `return r_str_newf ("%d", pid);`?

radare:
As in exposing the rdebugselect api call with rdebugbind from io or use corebind from io to run the dp= command to do that unless this can infiniteloop

PR #15367, opened 2019-10-27
Fix #14279 - Ignore entrypoint on non-executable ELFs ##bin

### LINKED ISSUE #14279: Radare2 falsely detects an Entry Point in an ELF object file
Make sure you are testing using the latest git version of radare2 before submitting any issue.

*If you would like to report a bug, please fill the template bellow*

### Work environment

|File format of the file you reverse (mandatory)      | ELF
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.6.0-git 22334 @ linux-x86-64 git.3.5.1-210-g759ebd5   commit: 759ebd5bfd2f8572d6866fffa7e26b7c739329f4 build: 2019-06-10__18:43:5

### Expected behavior
Radare2 should no detect an entry point in the binary and should not rename it to "entry0" instead of using the symbol provided

image

### Actual behavior

```
$ r2 pngrutil.o
[0x08000034]> afl
0x08000034    3 54           entry0
0x0800006a    1 76           sym.png_get_uint_32
0x080000b6    1 76           sym.png_get_int_32
...
...
...

````

The function detected as "entry0" should be called `sym.png_get_uint_31`. 

As you can see, radare2 correctly set a flag "sym.png_get_uint_31":

```
[0x08000034]> fd.
entry0
section..text
sym..text
sym.png_get_uint_31
eip
```

And shows it also here:

```
[0x08000034]> pd 4
            ;-- section..text:
            ;-- .text:
            ;-- png_get_uint_31:
            ;-- eip:
╭ (fcn) entry0 54
│   entry0 (int32_t arg_8h, int32_t arg_ch);
│ bp: 3 (vars 1, args 2)
│ sp: 0 (vars 0, args 0)
│ rg: 0 (vars 0, args 0)
│           0x08000034      55             push ebp                    ; [01] -r-x section size 21908 named .text
│           0x08000035      89e5           mov ebp, esp
│           0x08000037      83ec18         sub esp, 0x18
````

It has an entry point detected:
```
[0x08000034]> ie
[Entrypoints]
vaddr=0x08000034 paddr=0x00000034 haddr=0x00000018 hvaddr=0x08000018 type=program

1 entrypoints
```

1. Open pngrutil.o in radare2
2. analyze the binary
3. List the functions using "afl"

**Binary:**
[pngrutil.zip](https://github.com/radare/radare2/files/3272930/pngrutil.zip)

### PR DESCRIPTION
**Pull request by radare**

### CODE REVIEW

**Comment by ret2libc** [comment]
Why? Do we have a test case?

**Comment by radare** [comment]
Maybe we can add an if only for the object files which shouldnt have entrypoint

**Comment by ret2libc** [comment]
Oh, if this is about object files .o, then yes i think we should do this only for thos kind of files. But please, do share a reproducer if you have one.

**Comment by radare** [comment]
Its in itay’s issue. Busy irl from phone right now, just search for the file name i gave u. Ill push it to r2r and add the filetyoe chk. But shared libs in linux can have entrypoints

**Comment by radare** [comment]
@ret2libc feedback again?

File: libr/bin/format/elf/elf.c [hunk 1]
```diff
@@ -1861,10 +1861,21 @@ ut64 Elf_(r_bin_elf_get_fini_offset)(ELFOBJ *bin) {
...
+	case ET_DYN:  return true;
+	}
+	return false;
+}
+
 ut64 Elf_(r_bin_elf_get_entry_offset)(ELFOBJ *bin) {
 	r_return_val_if_fail (bin, UT64_MAX);
 	ut64 entry = bin->ehdr.e_entry;
 	if (!entry) {
+		if (!isExecutable (bin)) {
```
ret2libc: @radare I think this should be done for ET_REL only, no? In that case, using function `is_bin_etrel` could have been enough.

radare:
can you make a PR?

ret2libc:
> can you make a PR?

yes, i'll do one later today!

PR #15369, opened 2019-10-27
Implement w32dbg_wrap

GustavoLCR:
Basically the same as ptrace_wrap, but in windows, only receiving debug events is limited to the thread that started the debug session.


radare:
Next time use the ##debug

PR #15375, opened 2019-10-28
sys/asan.sh: Use -fsanitize=... instead of -lasan when linking

kazarmy:
As per the [instructions](https://clang.llvm.org/docs/AddressSanitizer.html#usage) (and as recommended [here](https://stackoverflow.com/a/40215639) and [here](https://groups.google.com/d/msg/address-sanitizer/SD590XDinfQ/NMUPj_G0BgAJ)).


kazarmy:
Works with gcc 9.2.1.

PR #15383, opened 2019-10-29
Make the generated r2.bat use bat path_relative syntax instead of hardcoding

TheGoddessInari:
This makes calling `r2.bat` from the path (or should be anywhere else) "just work" as long as radare2 is in the same directory.

This produces an `r2.bat` like:

```bat
@"%~dp0\radare2" %*
```

### DIFF
--- a/sys/meson.py
+++ b/sys/meson.py
@@ -170,7 +170,7 @@ def win_dist(args):
     r2_bat_fname = args.install + r'\bin\r2.bat'
     log.debug('create "%s"', r2_bat_fname)
     with open(r2_bat_fname, 'w') as r2_bat:
-        r2_bat.write('@"%s\\bin\\radare2" %%*\n' % os.path.abspath(args.install))
+        r2_bat.write('@"%~dp0\\radare2" %*\n')
 
     copy(r'{BUILDDIR}\libr\*\*.dll', r'{DIST}\bin')
     makedirs(r'{DIST}\{R2_LIBDIR}')


sys/meson.py
```diff
@@ -170,7 +170,7 @@ def win_dist(args):
     r2_bat_fname = args.install + r'\bin\r2.bat'
     log.debug('create "%s"', r2_bat_fname)
     with open(r2_bat_fname, 'w') as r2_bat:
-        r2_bat.write('@"%s\\bin\\radare2" %%*\n' % os.path.abspath(args.install))
+        r2_bat.write('@"~dp0\\radare2" %%*\n')
```
kazarmy: Use `'@"%~dp0\\radare2" %*\n'` instead

sys/meson.py
```diff
@@ -170,7 +170,7 @@ def win_dist(args):
     r2_bat_fname = args.install + r'\bin\r2.bat'
     log.debug('create "%s"', r2_bat_fname)
     with open(r2_bat_fname, 'w') as r2_bat:
-        r2_bat.write('@"%s\\bin\\radare2" %%*\n' % os.path.abspath(args.install))
+        r2_bat.write('@"~dp0\\radare2" %*\n')
```
kazarmy: Add also a `%` before `~dp0`, i.e. `%~dp0`

kazarmy:
Thanks!

TheGoddessInari:
Just glad I could help. I'd noticed that previously. 🦊

PR #15397, opened 2019-11-01
Sync default sanitizer list of sys/asan.sh and sys/meson.py

kazarmy:
Also, the [manual](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#available-checks) says that `-fsanitize=undefined` implies `-fsanitize=signed-integer-overflow` (confirmed with gcc 9.2.1).

### DIFF
--- a/sys/asan.sh
+++ b/sys/asan.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 # ASAN="address leak memory undefined"
-ASAN=${ASAN:="address undefined signed-integer-overflow"}
+# ASAN="address signed-integer-overflow"  # Faster build
+ASAN=${ASAN:="address undefined"}
 
 printf "\033[32m"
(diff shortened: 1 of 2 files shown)


sys/asan.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 # ASAN="address leak memory undefined"
-ASAN=${ASAN:="address undefined signed-integer-overflow"}
+# ASAN="address signed-integer-overflow"  # Faster build
+ASAN=${ASAN:="address undefined"}
```
radare: why drop intovf?

kazarmy: because `undefined` should also include `signed-integer-overflow` ... no harm in keeping it though (ff7198b16d7121212bf93a879baa5b517caf7f2e)

PR #15400, opened 2019-11-01
WIP: Use RStrBuf for storing reg profile strings

pelijah:
In the most cases string passed to `r_reg_set_profile_string()` is a string literal. And some of them are 3k characters long. We can use RStrBuf objects for storing pointer to a const string which will allow us to avoid:
1) Calling `strdup()` in `r_anal_get_reg_profile()` (just alloc a new RStrBuf object)
2) Calling `strdup()` in `r_reg_set_profile_string()` for caching reg profile string (just copy the RStrBuf object)
3) Calling `strcmp()` in `r_reg_set_profile_string()` for comparing reg profile string (just compare pointers).

But these changes gives tangible analysis speedup only for multibit binaries (e.g. ARM/Thumb) and only on Windows.

### Windows (VS 2017/release/shared)
`Measure-Command {.radare2.exe -Q -qq -A libcrackme.so}`
#### Before:
```
Minutes           : 0
Seconds           : 55
...
```
#### After:
```
Minutes           : 0
Seconds           : 49
...
```
### GNU/Linux (gcc 8.3.0/release/shared)
`time radare2 -A -qq -Q libcrackme.so`
#### Before:
```
real	0m37,300s
user	0m36,576s
sys	0m0,702s
```
#### After:
```
real	0m35,684s
user	0m34,831s
sys	0m0,820s
```


libr/anal/anal.c
```diff
@@ -265,7 +265,7 @@ R_API bool r_anal_use(RAnal *anal, const char *name) {
 	return false;
 }
 
-R_API char *r_anal_get_reg_profile(RAnal *anal) {
+R_API RStrBuf *r_anal_get_reg_profile(RAnal *anal) {
```
radare: why. whats the benefit of this.strbuf is useful when the string is suposed to be resized and modified many times, this is not the case for the regprofile, so i dont see the benefit

pelijah: We can use RStrBuf for storing pointers to constant strings. See the PR description.

ret2libc:
Isn't r_str_constpool supposed to do just that?

pelijah:
> 
> 
> Isn't r_str_constpool supposed to do just that?

I think using `r_str_constpool` will make analysis much slower than now.

ret2libc:
> > Isn't r_str_constpool supposed to do just that?
> 
> I think using `r_str_constpool` will make analysis much slower than now.

Mh, after looking at the implementation (which uses sdb_ht, when it should be using HtPP), I kinda agree...

ret2libc:
But still, I don't like too much making RStrBuf even more complicated to support this const case, which should be the perfect usecase for something like r_strconstpool (which may need to be improved). If these strings are constant and don't need to be changed, why do we strdup them all around?

pelijah:
Profile can be loaded from a file:
https://github.com/radareorg/radare2/blob/36e557aa9619c1900e0fe62d9e7048475528ca1c/libr/reg/profile.c#L238

Or from debug plugins:
https://github.com/radareorg/radare2/blob/571a080ca732ba585686a5580de2d14f18453a29/libr/debug/p/debug_io.c#L77
https://github.com/radareorg/radare2/blob/20e19d602337d29d9e2831ddf3f8b46168b9450a/libr/debug/p/debug_rap.c#L51
https://github.com/radareorg/radare2/blob/82b02e8b89aad92d879e376e519b10edd715ff55/libr/debug/p/debug_gdb.c#L454

pelijah:
Therefore you can't just remove `strdup()`.
I don't think that my changes make RStrBuf code super complicated. Anyway we need to make similar code changes in `strbuf.c` if  we want to add ref counting feature in the future.

ret2libc:
@pelijah did you have a look at r_strconstpool? it should give you exactly what you need with regard to this const thing. @thestr4ng3r have improved it a bit, so it should be fast enough IMHO. You should just use `r_str_constpool_get` and then you can compare pointers once you are sure you are dealing with constpool strings.

pelijah:
Using `r_str_constpool_get()` gives almost no effect to analysis speed (compared to current implementation with `strdup() + strcmp() + strdup()`).

pelijah:
During analysis of some binaries r2 modifies `asm.bits` (e.g. on ARM binaries it often switches between 16/32 bits) that leads to a call of `r_anal_set_reg_profile()` function. This function can implicitly invoke `r_reg_set_profile_string()` via `set_reg_profile` method of the current anal plugin, or explicitly pass  to `r_reg_set_profile_string()` a reg profile string obtained from `get_reg_profile` method. `r_reg_set_profile_string()` uses `strdup()` to cache the reg profile string and `strcmp()` to avoid parsing unchanged reg profiles.

In sum: `r_anal_set_reg_profile()` is called too many times -> `strcmp()` + `strdup()` are called too many times. Don't forget that some reg profile strings are 3k characters long. Also `get_reg_profile` method returns a pointer to a heap allocated string so don't forget about extra `strdup()`.

In the most cases a string passed to `r_reg_set_profile_string()` is a constant string. No need to spend CPU time on these `strcmp()` and `strdup()`. We can just compare pointers and save it directly. But in some cases we have to pass a heap allocated string to `r_reg_set_profile_string()`. So sometimes we stiil need to use `strcmp()` and `strdup()`. To resolve this issue we need to have some meta about the pointer to a string passed to `r_reg_set_profile_string()`. We can add a new boolean argument for `r_reg_set_profile_string()` or use some container. I think RStrBuf object is a quite suitable container for this.

I have to say I understand that one use case of the presented feature is not enough to add it.

libr/anal/anal.c
radare: strbuf changes the pointer depending of the length of the string, also, what about returning const char * and int *len? and then just delegate the logic to compare pointer in a single place

ret2libc:
> Using `r_str_constpool_get()` gives almost no effect to analysis speed (compared to current implementation with `strdup() + strcmp() + strdup()`).
> Current:
> изображение
> Constpool (here almost all CPU time of `r_reg_set_profile_string()` is spent on `r_str_constpool_get()`):
> изображение

Did you do the comparisons with pointers after you are sure the strings are coming from constpool? Or you are just calling r_strconstpool_get again and again?

radare:
ah wait, cant be const * because we need to construct them from other langs too

libr/anal/anal.c
pelijah: > strbuf changes the pointer depending of the length of the string
Can you explain more?

radare:
just read the code:

typedef struct {
        char buf[32];
        int len;
        char *ptr;
        int ptrlen;
} RStrBuf;

ptr=buf unless string length > 32

pelijah:
> Did you do the comparisons with pointers after you are sure the strings are coming from constpool? Or you are just calling r_strconstpool_get again and again?

I used `r_strconstpool_get()` in `r_reg_set_profile_string()` in that way:
```c
char *tmp = r_strconstpool_get(&reg->pool, str);
if (tmp == reg->reg_profile_str) {
    return true;
}
reg->reg_profile_str = tmp;
```

Do you mean we should assume that the string passed to `r_reg_set_profile_string()` is a string constant or a string from `r_str_constpool` so we can just store and compare given pointers? Then yep, it will work faster.

radare:
can you make it green?

libr/anal/anal.c
radare: to store pointers to constant strings we may probably want to use RStrConst instead, but this api changed recently to be thread safe so its not an option here anymore.

libr/reg/profile.c
```diff
@@ -248,7 +249,10 @@ R_API bool r_reg_set_profile(RReg *reg, const char *profile) {
 		eprintf ("r_reg_set_profile: Cannot find '%s'\n", profile);
 		return false;
 	}
-	bool ret = r_reg_set_profile_string (reg, str);
+	r_strbuf_init (&sb);
+	r_strbuf_set (&sb, str);
+	bool ret = r_reg_set_profile_string (reg, &sb);
```
radare: i dont see the beneffit of passing a strbuf to set the regprofile instead of just passing the string. you can't compare by pointers here, so its an extra workload that is unnecessary here

libr/reg/profile.c
```diff
@@ -140,7 +140,7 @@ R_API bool r_reg_set_profile_string(RReg *reg, const char *str) {
 	r_reg_arena_shrink (reg);
 
 	// Cache the profile string
-	reg->reg_profile_str = strdup (str);
+	r_strbuf_copy (&reg->reg_profile_str, str);
```
radare: shuoldnt this be r_strbuf_set ?

libr/anal/p/anal_6502.c
```diff
@@ -928,7 +929,8 @@ static int set_reg_profile(RAnal *anal) {
 		"gpr	N	.1	.31	0\n"
 		"gpr	sp	.8	4	0\n"
 		"gpr	pc	.16	5	0\n";
-	return r_reg_set_profile_string (anal->reg, p);
+	r_strbuf_init_const (&sb, p, strlen (p));
```
radare: thathts filling the string without the trailing zero

libr/anal/p/anal_6502_cs.c
```diff
@@ -199,7 +200,8 @@ static int set_reg_profile(RAnal *anal) {
 		"gpr	N	.1	.31	0\n"
 		"gpr	sp	.8	4	0\n"
 		"gpr	pc	.16	5	0\n";
-	return r_reg_set_profile_string (anal->reg, p);
+	r_strbuf_init_const (&sb, p, strlen (p));
+	return r_reg_set_profile_string (anal->reg, &sb);
```
radare: i dont see any benefit on this change of function signture, the _string should be kept as const char *

Issue #13934, opened 2019-04-30
No zignspace context in applied flags or function naming


AlexKornitzer:
**Is your feature request related to a problem? Please describe.**
This could well be a design decision and not an oversight hence the raise as a feature request over bug report. The issue here is that any zignspace context for a set of zignatures is not applied to the flags or function names when command such as `z/` are run. As a result when looking at function names or flags for this data you lose this additional context. This means that if you have two signatures with the same name in different namespaces you will not be able to tell which was matched.

**Describe the solution you'd like**
My current though is that the namespace should be applied to flags and function names, the issues arise from as to how best to approach this. For flags I am thinking that the following would fit in well and be sufficient: `sign.MATCH_TYPE.ZIGNSPACE.FUNCTION_NAME` For function names this becomes a little bit more complicated, I am currently thinking that we could us a delimiter approach which would then allow automatic namespace application but this could conflict with how people name their functions. For example if we used `:` as the delimiter, once functions had been named, when we run `zg` it would split on `:` to get the namespace if any and apply the function name (this delimiter could be a setting in the r2 config): `ZIGNSPACE:FUNCTION_NAME`

Additionally I think we should replace the `count` with the matching address, the current count values are confusing and inaccurate (as noted in the source). 

**Describe alternatives you've considered**
1. Only apply zignspace context to flags, the latest match will apply to the function name.

**Additional context**
Current State of Play:
![image](https://user-images.githubusercontent.com/2750747/56952224-bb36a180-6b31-11e9-9792-db3b4afd6f11.png)

AlexKornitzer:
If we take this forward I am happy to implement it.

radare:
can you share the binary? all those Warning messages are not good

AlexKornitzer:
Yeah can do, it is nothing sensitive just `ls` from Arch Linux: https://www.archlinux.org/packages/core/x86_64/coreutils If you would rather I just upload the binary let me know where to put it.

Any thoughts on the actual issue?

radare:
@AlexanderYukhanov are you still happy to implement it?

AlexKornitzer:
@radare if that was intended for me, I am happy to implement this was just holding off until you gave feedback on the approach above as to not waste time implementing something that was not wanted or sufficient. If you are happy with the solution above, I will try and get it done over the next couple of weeks.

radare:
Sgtm. Maybe others have comments here @ret2libc @xvilka ?

[...]

PR #15402, opened 2019-11-02
WIP: add auto zignspace generation for the zg command

AlexKornitzer:
Part 1:
This allows the 'zg' command to auto generate zignspaces from function
names by utilising the ':' convention used in zignatures. Now when 'zg' is
run if any funtion names are in the format ZIGNSPACE:FUNCTION, the
zignspace will be extracted, created and the function applied under its
scope.

### DIFF
--- a/libr/core/cmd_zign.c
+++ b/libr/core/cmd_zign.c
@@ -183,13 +183,22 @@ static char *getFcnComments(RCore *core, RAnalFunction *fcn) {
 #endif
 
 static void addFcnZign(RCore *core, RAnalFunction *fcn, const char *name) {
+	char *ptr = NULL;
+	char *zignspace = NULL;
 	char *zigname = NULL;
 	const RSpace *curspace = r_spaces_current (&core->anal->zign_spaces);
+	int len = 0;
 
 	if (name) {
 		zigname = r_str_new (name);
 	} else {
-		if (curspace) {
+		// If the user has set funtion names containing a single ':' then we assume
+		// ZIGNSPACE:FUNCTION, and for now we only support the 'zg' command
+		if ((ptr = strchr (fcn->name, ':')) != NULL) {
+			len = ptr - fcn->name;
+			zignspace = r_str_newlen (fcn->name, len);
+			r_spaces_push (&core->anal->zign_spaces, zignspace);
+		} else if (curspace) {
 			zigname = r_str_newf ("%s:", curspace->name);
 		}
 		zigname = r_str_appendf (zigname, "%s", fcn->name);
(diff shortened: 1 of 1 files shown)


AlexKornitzer:
This starts the work on #13934

AlexKornitzer:
Due to the current implementation we kind of get the zignspace in the match for free due to it being in the name. Are there any strong views on this, should the ':' be changed to '.' or are we happy with it as is?

image

Furthermore should we strip the zignspace from the function name when printing? Should be easy to do as we can just prefix check, otherwise we currently have the following output:
image

libr/core/cmd_zign.c
```diff
@@ -183,13 +183,26 @@ static char *getFcnComments(RCore *core, RAnalFunction *fcn) {
...
+		// If the user has set funtion names containing a single ':' then we assume
+		// ZIGNSPACE:FUNCTION, and for now we only support the 'zg' command
+		if (r_str_char_count (fcn->name, ':') == 1) {
+			for (i = 0; i < strlen (fcn->name); i++) {
+				if (fcn->name[i] == ':') {
+					break;
+				}
+				len++;
+			}
+			zignspace = r_str_newlen(fcn->name, len);
```
radare: missing space before (

libr/core/cmd_zign.c
```diff
@@ -183,13 +183,26 @@ static char *getFcnComments(RCore *core, RAnalFunction *fcn) {
...
+	int i, len = 0;
 
 	if (name) {
 		zigname = r_str_new (name);
 	} else {
-		if (curspace) {
+		// If the user has set funtion names containing a single ':' then we assume
+		// ZIGNSPACE:FUNCTION, and for now we only support the 'zg' command
+		if (r_str_char_count (fcn->name, ':') == 1) {
+			for (i = 0; i < strlen (fcn->name); i++) {
```
radare: put the strlen outside the loop

radare: uhm. what about just using strchr ?

libr/core/cmd_zign.c
AlexKornitzer: Done

libr/core/cmd_zign.c
AlexKornitzer: Removed

AlexKornitzer: Yeah makes much more sense my C is incredibly rust atm, worth checking I have not messed up the arithmetic there.

AlexKornitzer:
So all good to go, unless we address my queries above about:
1. Using ':' in flags
2. Removing zignspace from zignature name

radare:
sgtm, lets wait for travis to be green , did u rebased?

AlexKornitzer:
Ah yes will rebase and squash. I think I will do a new PR for addressing the zignzpace in zignature name issue as IMO is should be removed.

PR #15403, opened 2019-11-02
Automatically set the debug plugin when opening debug uris

### LINKED ISSUE #15399: Problem connecting to gdbserver using commands
### Work environment

| OS/arch/bits (mandatory)                             | Ubuntu 18.04 x86 64

### Expected behavior
```
radare2 -A ~/workspace/tests/main
[x] Check for objc references
[x] Check for vtables
 -- git checkout hamster
[0x00400a40]> o gdb://127.0.0.1:1234
= attach 50119 0
[0x00400af0]>
[0x00400af0]> dc
[0x7ffff7bbed2d]>
```

### Actual behavior
```
radare2 -A ~/workspace/tests/main
[x] Check for objc references
[x] Check for vtables
 -- The unix-like reverse engineering framework.
[0x00400a40]> o gdb://127.0.0.1:1234
= attach 50119 0
ptrace (PT_ATTACH): No such process
[0x00400af0]> dc
Cannot continue, run ood?
[0x00000000]>
```

#### Bypass

```
radare2 -A ~/workspace/tests/main
[x] Check for objc references
[x] Check for vtables
-- It's the year of desktop on radare2
[0x00400a40]> o gdb://127.0.0.1:1234
= attach 50119 0
ptrace (PT_ATTACH): No such process
[0x00400af0]> dL gdb
[0x00400af0]> o gdb://127.0.0.1:1234
= attach 50119 0
[0x00400af0]> dc
[0x7ffff7bbed2d]>
```

- Run gdbserver on any binary `gdbserver :1234 /bin/ls`
- Connect in r2 with `o gdb://127.0.0.1:1234`
- Run `dc`
- It's possible to bypass this behavior using:
    -  `o gdb://127.0.0.1:1234`
    - `dL gdb`
    -  `o gdb://127.0.0.1:1234`

Notes: 
- Setting `dL gdb` before `o gdb://...` and/or running `o gdb://..` twice in a row doesn't help.
- Gdbserver sees the connection in the first try, it's just that the debugger isn't configured.

Didn't try reproducing on windows but it might be the same with windbg.

yossizap:
closes #15399 

Solves the gdbserver issue and fixes errors that might derive from forgetting `dL` or setting the wrong debug handler.

### DIFF
--- a/libr/core/cfile.c
+++ b/libr/core/cfile.c
@@ -907,9 +907,11 @@ R_API RCoreFile *r_core_file_open(RCore *r, const char *file, int flags, ut64 lo
 			swstep = false;
 		}
 		r_config_set_i (r->config, "dbg.swstep", swstep);
-		// Set the correct debug plugin
-		if (fd->plugin && fd->plugin->name && fd->plugin->isdbg) {
-			r_debug_use (r->dbg, fd->plugin->name);
+		// Set the correct debug handle
+		if (fd->plugin && fd->plugin->isdbg) {
+			const char *dh = strndup(file, (strstr(file, "://") - file));
+			r_debug_use (r->dbg, dh);
+			free(dh);
 		}
 	}
 	//used by r_core_bin_load otherwise won't load correctly
--- a/libr/core/cfile.c
+++ b/libr/core/cfile.c
@@ -909,10 +909,10 @@ R_API RCoreFile *r_core_file_open(RCore *r, const char *file, int flags, ut64 lo
 		r_config_set_i (r->config, "dbg.swstep", swstep);
 		// Set the correct debug handle
 		if (fd->plugin && fd->plugin->isdbg) {
-			const char *dh = r_str_ndup (file, (strstr (file, "://") - file));
+			char *dh = r_str_ndup (file, (strstr (file, "://") - file));
 			if (dh) {
 				r_debug_use (r->dbg, dh);
-				free(dh);
+				free (dh);
 			}
 		}
 	}
--- a/libr/core/cfile.c
+++ b/libr/core/cfile.c
@@ -907,6 +907,10 @@ R_API RCoreFile *r_core_file_open(RCore *r, const char *file, int flags, ut64 lo
 			swstep = false;
 		}
 		r_config_set_i (r->config, "dbg.swstep", swstep);
+		// Set the correct debug plugin
+		if (fd->plugin && fd->plugin->name && fd->plugin->isdbg) {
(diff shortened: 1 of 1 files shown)


libr/core/cfile.c
```diff
@@ -907,6 +907,12 @@ R_API RCoreFile *r_core_file_open(RCore *r, const char *file, int flags, ut64 lo
 			swstep = false;
 		}
 		r_config_set_i (r->config, "dbg.swstep", swstep);
+		// Set the correct debug handle
+		if (fd->plugin && fd->plugin->isdbg) {
+			const char *dh = strndup(file, (strstr(file, "://") - file));
```
radare: What happens if theres no :// ?

yossizap: Added a check. Is there a point in printing an error or attempting to set the current plugin's name as the debugger in case the uri didn't contain the debugger backend? Seems unlikely with fd->plugin->isdbg

libr/core/cfile.c
```diff
@@ -907,6 +907,14 @@ R_API RCoreFile *r_core_file_open(RCore *r, const char *file, int flags, ut64 lo
 			swstep = false;
 		}
 		r_config_set_i (r->config, "dbg.swstep", swstep);
+		// Set the correct debug handle
+		if (fd->plugin && fd->plugin->isdbg) {
+			const char *dh = r_str_ndup (file, (strstr (file, "://") - file));
```
ret2libc: if it's freed, it should not be `const char *`, I think.

ret2libc: what if there is no `://` in `file`?

yossizap: >if it's freed, it should not be const char *, I think.
Seems logical since you want to treat it as const, this is common practice with strdup but you can't control it's usage of malloc. I could explicitly cast the pointer down to char* when freeing or just remove const since it doesn't really matter.

>what if there is no :// in file?
Nothing is set since, as far as I know, it's not supposed to happen. Is there a valid uri that would set fd->plugin->isdbg to true without having ://? radare2.c does the same with -d

radare: drop the const

libr/core/cfile.c
```diff
@@ -907,6 +907,14 @@ R_API RCoreFile *r_core_file_open(RCore *r, const char *file, int flags, ut64 lo
...
+			if (dh) {
+				r_debug_use (r->dbg, dh);
+				free(dh);
```
radare: imssing space before (

PR #15405, opened 2019-11-03
sys/meson.py: --asan accepts sanitizer list

kazarmy:


XVilka:
Nice, btw it makes sense to rename the flag `--asan` in the future, since it's not only the ASAN itself.

kazarmy:
> Nice, btw it makes sense to rename the flag --asan in the future, since it's not only the ASAN itself.

`--san`?

kazarmy:
`--sanitize` is better i suppose

PR #15413, opened 2019-11-04
Remove core->oobi because not used

ret2libc:


radare:
oobi was used to feed commands via stdin, i would like to discuss it before deleting it

ret2libc:
Ok, why was it removed? It is not used at the moment and I’m not sure who ever missed it

radare:
because some c++ programs uused '<' in the flagnames and caused mayhem

ret2libc:
What is the relation between c++ flag names and oobi?

radare:
c++ flagnames contain '<', which triggered the oobi thing and caused "hangs" because r2 was not responding waiting for a ^D and users got confused , the sky was dark and there was fire in the streets

ret2libc:
But... what's the purpose of this `<` thing? For the end user, I mean. Is it really needed? Did anybody complain that it is not working? Is it such a good feature to have?

Personally, I never used it, never felt it missing in r2 or other tools.

radare:
some commands like 'wx' used oobi when no argument was given, so you could do things like:

`wx < foo.hex`

or write an entire binary file with `w < /bin/ls` instead of `wff /bin/ls`

ret2libc:
> some commands like 'wx' used oobi when no argument was given, so you could do things like:
> 
> `wx < foo.hex`
> 
> or write an entire binary file with `w < /bin/ls` instead of `wff /bin/ls`

I think oobi is not strictly necessary then, as we already have `wff` or similar commands that get a filename instead of a content. We *could* reimplement it in the new parser, once it's done, but I don't see it as a must-have feature, even now, as it is even disabled and nobody asked for it.

What do you think? Shall we just keep all that unused code? It will always be available in the git history, if you really want to have a look at it.

radare:
its ok to kill it for me because i already deleted all the oobi code from other commands.. but imho this discussion was necessary, and maybe we can just handle the '<' in the commands itself, but imho i would preffer r2 to be more unixy related to pipes, and avoid abusing the rcons buffer that much

XVilka:
Merged then.

PR #15414, opened 2019-11-04
Very simple tree sitter integration

ret2libc:
This is very far from a functional parser, but I want to do things slowly and give everybody the opportunity to look at the code without having to look at a huge PR only at the end.

Here is the grammar: https://github.com/ret2libc/tree-sitter-r2cmd/ . There are a lot of things to improve there as well, very very far from complete.

For now, you can set `scr.newr2cmd=true` and start using the new parser, which supports:
- simple commands with arguments
- temporary seek with `@`
- interpret commands that start with `.`

I'm not saying this PR should be merged, but it's just to get some feedback and let you see the direction I'm taking. In theory, it can even be merged as all the new code is used only if `scr.newr2cmd` is true, otherwise it all works as usual.


radare:
also, for the config key what about `cfg.<something>`

ret2libc:
> also, for the config key what about `cfg.<something>`

I'm ok with any config var name. Propose something if you a better name, I'm not that good at it :(

ret2libc:
> i dont want external repos, also seems this git clone is only for meson .?

It is only for meson now, but it can be easily done for acr/makefile as well, the same way we already do for capstone.

ret2libc:
@radare what are the reasons for having tree-sitter and tree-sitter-r2cmd copied in the r2 repository? We have seen with capstone that downloading the dependencies works. Also, they are external projects and having them in the main repository just creates confusion (think about how many times people have committed changes in shlr/sdb instead of committing them in the sdb repository and only then sync the directory in r2). Also, especially at the beginning I will do many changes to the tree-sitter-r2cmd and it's much easier to just update a commit-id instead of having to sync the entire repository. Moreover, tree-sitter-r2cmd should be kept separated because it can be reused by cutter, vim, or whatever other project will want to support highlighting for r2 scripts (just an example).

ret2libc:
Also, tree-sitter itself is released quite frequently (at least now) so I think it's easier to just keep track of the commit id. Finally, I don't see any reason why we should put in our git all things we use.

XVilka:
At this point I think it's better to keep it in the separate repository, at least until it is stabilized and becomes default parsing method.

libr/core/cmd.c
```diff
@@ -4349,7 +4352,145 @@ R_API void run_pending_anal(RCore *core) {
...
+		}
+	}
+	return res;
+}
+
+static bool core_cmd_tsr2cmd(RCore *core, const char *cstr, bool log) {
+	// Create a parser.
+	TSParser *parser = ts_parser_new ();
+
+	// Set the parser's language (JSON in this case).
```
thestr4ng3r: Copypasta 😜️

shlr/meson.build
```diff
@@ -233,6 +233,92 @@ sdb_gen_cmd = [
 ]
 
 
+# handle tree-sitter dependency
+tree_sitter_path = join_paths(meson.current_source_dir(), 'tree-sitter')
```
thestr4ng3r: Should be under `meson.current_build_dir()`. Meson should never touch the source dir. If it has to be in the source dir, submodules should be used instead.

shlr/meson.build
```diff
@@ -233,6 +233,92 @@ sdb_gen_cmd = [
...
+)
+
+tree_sitter_dep = declare_dependency(
+  link_with: libtree_sitter,
+  include_directories: tree_sitter_inc
+)
+
+
+# handle tree-sitter-r2cmd dependency
+tree_sitter_r2cmd_path = join_paths(meson.current_source_dir(), 'tree-sitter-r2cmd')
```
thestr4ng3r: same

shlr/meson.build
ret2libc: i know, you are right, but... i was following what was requested to be done with capstone as well. I'm perfectly fine with moving it to current_build_dir, but I'm pretty sure when I will implement this in acr/makefile it will live in the source directory (as there is no "build" directory in that case).

libr/core/cmd.c
ret2libc: ups :D :D you got me

shlr/meson.build
thestr4ng3r: Yeah acr being in-source is bad by itself. But if it works, at least meson should stay in the build dir.

ret2libc: yes, it can easily work for meson! I'll do it.

ret2libc:
@radare could you review the acr/makefile bits please? Now tests should be green and tree-sitter integration should work both with meson and acr/makefile.

shlr/tree_sitter_r2cmd.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
```
radare: imho we should have a generic script to fetch external repos, rather than copying the same script with almost the same code. we can maybe reuse it for other things.. but all that stuff was done in makefiles for capstone, no need for shellscripts maybe

ret2libc: capstone does use a script AFAIK (capstone.sh).

I can use just one script for tree-sitter and tree-sitter-r2cmd. Capstone is a bit different because it requires patching and other complicated stuff :/

XVilka:
LGTM. I am more curious about the grammar.

PR #15424, opened 2019-11-06
Use HtPP for RStrConstPool

thestr4ng3r:
Test: https://github.com/radareorg/radare2-regressions/pull/2010

### DIFF
--- a/libr/util/str_constpool.c
+++ b/libr/util/str_constpool.c
@@ -2,19 +2,31 @@
 
 #include "r_util/r_str_constpool.h"
 
+static void kv_fini(HtPPKv *kv) {
+	free (kv->key);
+}
+
 R_API bool r_str_constpool_init(RStrConstPool *pool) {
-	pool->ht = sdb_ht_new ();
+	pool->ht = ht_pp_new (NULL, kv_fini, NULL);
 	return pool->ht != NULL;
 }
 
 R_API void r_str_constpool_fini(RStrConstPool *pool) {
-	sdb_ht_free (pool->ht);
+	ht_pp_free (pool->ht);
 }
 
 R_API const char *r_str_constpool_get(RStrConstPool *pool, const char *str) {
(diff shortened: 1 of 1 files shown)


thestr4ng3r:
@ret2libc Is it legal to return the keys as the pooled strings here, i.e. assume the keys will stay allocated and keep their contents as long as the HtPP is alive?

libr/util/str_constpool.c
```diff
@@ -2,19 +2,31 @@
...
+	ht_pp_free (pool->ht);
 }
 
 R_API const char *r_str_constpool_get(RStrConstPool *pool, const char *str) {
 	if (!str) {
 		return NULL;
 	}
-	sdb_ht_insert (pool->ht, str, str);
-	return sdb_ht_find (pool->ht, str, NULL);
+	HtPPKv *kv = ht_pp_find_kv (pool->ht, str, NULL);
```
ret2libc: this looks over complicated. What about:

```C
const char *res = ht_pp_find (pool->ht, str, NULL);
if (!res) {
  ht_pp_insert (pool->ht, str, str);
  res = str;
}
return res;
```

ret2libc:
> @ret2libc Is it legal to return the keys as the pooled strings here, i.e. assume the keys will stay allocated and keep their contents as long as the HtPP is alive?

Mh.. Yeah, it should be an ok assumption.

libr/util/str_constpool.c
thestr4ng3r: Here res would be the value, not the key. The idea is that I don't have to allocate both the key and the value, but just set the value to NULL and return the key. So I essentially use HtPP as a set, not a map.

ret2libc: Mmm I see why you are doing it. You are doing it because you want to use the key, which is strdupped in HtPP. Oki.

PR #15426, opened 2019-11-07
Search for AES 128, 192 and 256-bit keys

sylvainpelissier:


libr/search/aes-find.c
```diff
@@ -30,20 +30,56 @@ const ut8 table_sbox[256] = {
...
+	aes192 = buf[24] == (buf[0] ^ table_sbox[buf[21]] ^ 1) \
+		&& buf[25] == (buf[1] ^ table_sbox[buf[22]]) \
+		&& buf[26] == (buf[2] ^ table_sbox[buf[23]]) \
+		&& buf[27] == (buf[3] ^ table_sbox[buf[20]])?1:0;
+
+	return (aes192);
+}
+
+static int aes128_key_test(const unsigned char *buf) {
+	int aes128;
```
radare: just reutnr this... no need for a local variable

also, i think you want to return bool in here and drop the ?1:0;

radare:
please provide a better test

sylvainpelissier:
Do you mean testing the three different key sizes?

radare:
yes, but also, the actual test seems to check a null byte pattern, which .. sounds not correct to me, but maybe its just the output which is not reflecting the bytes found.

```
0x0000001e hit0_0 00000000000000000000000000000000000000000000000000000000000000

```

sylvainpelissier:
It is correct it is the zero key. But I'll update it.

radare:
ok

radare:
oooh ok so the finding shows the actual aes key, not the address where the expanded key is found?

sylvainpelissier:
0x0000001e is the address and then is the key value.

radare:
oh cool! :) thanks, then it looks good. can you update the src to address the comments?

sylvainpelissier:
Ok, done.

libr/search/aes-find.c
```diff
@@ -30,20 +30,44 @@ const ut8 table_sbox[256] = {
...
+static int aes128_key_test(const unsigned char *buf) {
+	return (buf[16] == (buf[0] ^ table_sbox[buf[13]] ^ 1) \
+		&& buf[17] == (buf[1] ^ table_sbox[buf[14]]) \
+		&& buf[18] == (buf[2] ^ table_sbox[buf[15]]) \
+		&& buf[19] == (buf[3] ^ table_sbox[buf[12]]));
 }
 
 R_API int r_search_aes_update(RSearch *s, ut64 from, const ut8 *buf, int len) {
-	int i, last = len-31;
+	int i, last = len-20;
```
radare: Missing space around -

libr/search/aes-find.c
```diff
@@ -30,20 +30,44 @@ const ut8 table_sbox[256] = {
 	0x8c, 0xa1, 0x89, 0x0d, 0xbf, 0xe6, 0x42, 0x68, 0x41, 0x99, 0x2d, 0x0f, 0xb0, 0x54, 0xbb, 0x16
 };
 
-static int aes_key_test(const unsigned char *buf) {
-	return (buf[28]==table_sbox[buf[15]] \
-	    &&  buf[29]==table_sbox[buf[12]] \
-	    &&  buf[30]==table_sbox[buf[13]] \
-	    &&  buf[31]==(table_sbox[buf[14]]^1))?1:0;
+static int aes256_key_test(const unsigned char *buf) {
```
radare: Return bool instead of int

PR #15429, opened 2019-11-07
Refactor r_flag_color() to r_flag_item_set_color()

thestr4ng3r:
To be consistent with `r_flag_item_set_realname()`, etc. The RFlag arg was completely unused.


radare:
will be good to fix the vapi files before the release :) thanks for this fix

PR #15433, opened 2019-11-07
Added break to gdbserver vCont and refactored the code to use locks

### LINKED ISSUE #15415: gdbserver debugger background tasks support
**Description**
- The current gdbr implementation doesn't allow interrupting background tasks with `&b` since send_vcon isn't properly configured with the cons api.
- Several commands can't be processed at the same time, resulting in undefined behavior and the user not being able to enter `&b [task]` which calls read_regs and breaks everything.

**Solutions**
- Add cons api - done.
- Add a mutex to all send/read combinations to avoid conflicting results/freezing.
- Add a fail state on mutex acquisition to read_regs instead of waiting since it's called all the time(including any enter) so we don't want it waiting for a lock. 

**Alternatives**
- Multiple debug sockets - tried and it didn't work. 
- Placing locks in gdbserver/src/packet.c on send/receive - error prone.
- Attempt to cancel the current task and resume after a non blocking task(reading registers, thread info etc) - need to test it, wouldn't start with this feature

Actively working on this, feedback appreciated.

yossizap:
The previous gdbr implementation didn't allow interrupting background tasks with &b since send_vcon wasn't properly configured with the cons api. In addition to that, gdbserver doesn't support processing multiple commands at the same time, resulting in undefined behavior once cons were set up for vcont.
This PR adds the relevant cons api and solves the concurrency issues by adding locks on all socket related logic.

I went ahead and fixed style issues and slightly refactored parts of the code. There are still parts that require refactoring from the original implementation but I am not sure if some of the behavior/lack of error checks is intentional. Decided to keep those parts as they were and do a proper refactor once I finish writing tests for gdbserver (incl concurrent tasks).

Testing:

- Execute `dc`, see that you can interrupt it with a regular Ctrl+C
- Execute `& dc`
    - Attempt to run a different command while dc is running with `& [cmd]`, see that you can stop it with `&b [task number]`. It should hang until `dc` is finished unless you interrupt it.
    - Run a command with `& [cmd]` and stop `& dc`'s task number, check that both are done
- Go over debug commands

closes #15415


XVilka:
Failed unit test:
```
gcc test_bin.c -o bin/test_bin -DR2_ASSERT_STDOUT=1 -I/home/travis/build/radareorg/radare2/install/include/libr -g -L/home/travis/build/radareorg/radare2/install/lib -lr_core -lssl -lcrypto -lr_config -lr_debug -lr_bin -lr_anal -lr_bp -lr_egg -lr_asm -lr_lang -lr_parse -lr_flag -lr_cons -lr_reg -lr_search -lr_syscall -lr_fs -lr_io -lr_socket -lr_magic -lr_crypto -lr_hash -lr_util -ldl
/home/travis/build/radareorg/radare2/install/lib/libr_io.so: undefined reference to `r_cons_sleep_end'
/home/travis/build/radareorg/radare2/install/lib/libr_io.so: undefined reference to `r_cons_sleep_begin'
/home/travis/build/radareorg/radare2/install/lib/libr_io.so: undefined reference to `r_cons_break_push'
/home/travis/build/radareorg/radare2/install/lib/libr_io.so: undefined reference to `r_cons_break_pop'
collect2: error: ld returned 1 exit status
make[2]: *** [bin/test_bin] Error 1
```

shlr/gdb/src/gdbclient/core.c
```diff
@@ -82,20 +84,52 @@ static void reg_cache_init(libgdbr_t *g) {
...
+	r_th_lock_enter (g->gdbr_lock);
+	r_cons_sleep_end (bed);
+	if (_isbreaked) {
+		return false;
+	}
+	return true;
+}
+
+void gdbr_lock_leave(libgdbr_t *g) {
+	r_cons_break_pop ();
```
radare: I dont think u can call rcons from here

libr/io/p/io_gdb.c
```diff
@@ -209,8 +209,8 @@ static int __gettid(RIODesc *fd) {
 	return desc ? desc->tid : -1;
 }
 
-int send_msg(libgdbr_t* g, const char* command);
-int read_packet(libgdbr_t* instance);
+extern int send_msg (libgdbr_t *g, const char *command);
```
radare: No space before ( in fcn signatures

yossizap:
Didn't have the correct pkgcfg exported when I ran the tests, should be good now.

For some reason I am getting a similar error when building Cutter and I can't figure it out, tried adding r_cons to all gdb and io makefiles/meson and it still doesn't work. Could anyone else take a look at this? 
Log: https://pastebin.com/N0viPxHd

shlr/gdb/src/gdbclient/core.c
yossizap: It does work? Try removing it and run `dc`, you'll get stuck in an infinite loop.
Not sure why this would be an invalid usage, the implementation calls [r_cons_context_break_pop](https://github.com/radareorg/radare2/blob/master/libr/cons/cons.c#L333) with I.context which should be the same in enter and leave, no?

yossizap:
For some reason I didn't have this problem while testing with sys/meson.py, this only happens when I try to compile the same way travis does.

Anyways, after a lot of attempt, this is the best solution I could come up with. You could add `link_args: 'lr_cons'` to gdb's `declare_dependency` (for some reason `-Wl,rpath` and `-Wl,soname` from `r_cons_dep` weren't enough even though they work for r_socket/r_util) but it doesn't work on Windows since it can't find r_cons.dll without `r_cons_dep`. 

Gdb requires it's own meson file to be included after r_cons is setup, otherwise it won't recognize the `r_cons` flag in `link_against` due to the include order in the main meson.build(Can't place cons before shlr since it depends on util which depends on shlr..).

pkgcfg/r_io.pc.acr
```diff
@@ -6,6 +6,6 @@ includedir=${prefix}/include
 Name: r_io
 Description: radare foundation libraries
 Version: @VERSION@
-Requires: r_util r_socket
+Requires: r_util r_socket r_cons
```
radare: u dont need rcons at all

shlr/gdb/Makefile
```diff
@@ -17,10 +17,12 @@ LD=$(CC)
 ifeq ($(LD),cccl)
 LDFLAGS+=-L${LIBR}/socket -llibr_socket
 LDFLAGS+=-L${LIBR}/util -llibr_util
+LDFLAGS+=-L${LIBR}/cons -llibr_cons
```
radare: drop this

shlr/gdb/Makefile
```diff
@@ -17,10 +17,12 @@ LD=$(CC)
...
 EXT_AR?=lib
 else
 LDFLAGS+=-L${LIBR}/socket -lr_socket
 LDFLAGS+=-L${LIBR}/util -lr_util
+LDFLAGS+=-L${LIBR}/cons -lr_cons
```
radare: both if and else blocks look the same to me and rcons shouldnt be a dependency

shlr/gdb/meson.build
```diff
@@ -0,0 +1,34 @@
...
+  include_directories(
+    './include',
+    './include/gdbclient',
+    './include/gdbserver'
+  )
+]
+
+libr2gdb = static_library('r2gdb', gdb_files,
+  include_directories: gdb_inc,
+  dependencies: [r_util_dep, r_cons_dep],
```
radare: drop cons

yossizap: Removing cons from there result in the following error in travis builds when compiling with `meson --prefix=/home/travis/build/radareorg/radare2/install -Db_coverage=true build`

```
shlr/gdb/libr2gdb.a(src_gdbclient_core.c.o): In function `gdbr_lock_enter':
/home/build/radareorg/radare2/build/../shlr/gdb/src/gdbclient/core.c:94: undefined reference to `r_cons_break_push'
/home/build/radareorg/radare2/build/../shlr/gdb/src/gdbclient/core.c:95: undefined reference to `r_cons_sleep_begin'
/home/build/radareorg/radare2/build/../shlr/gdb/src/gdbclient/core.c:97: undefined reference to `r_cons_sleep_end'
shlr/gdb/libr2gdb.a(src_gdbclient_core.c.o): In function `gdbr_lock_leave':
/home/build/radareorg/radare2/build/../shlr/gdb/src/gdbclient/core.c:105: undefined reference to `r_cons_break_pop'
shlr/gdb/libr2gdb.a(src_gdbclient_core.c.o): In function `gdbr_connect':
/home/build/radareorg/radare2/build/../shlr/gdb/src/gdbclient/core.c:190: undefined reference to `r_cons_sleep_begin'
/home/build/radareorg/radare2/build/../shlr/gdb/src/gdbclient/core.c:210: undefined reference to `r_cons_sleep_end'
shlr/gdb/libr2gdb.a(src_gdbclient_core.c.o): In function `send_vcont':
/home/build/radareorg/radare2/build/../shlr/gdb/src/gdbclient/core.c:1274: undefined reference to `r_cons_sleep_begin'
/home/build/radareorg/radare2/build/../shlr/gdb/src/gdbclient/core.c:1289: undefined reference to `r_cons_sleep_end'
collect2: error: ld returned 1 exit status
```

Issue #15279, opened 2019-10-15
Cannot analyze the IOKit function under Mac OS X


IceAmber:
When I analyze a Mac kernel extention by radare2, the IOKit function cannot analyze correctly
0x000008d4      488d3dee0600.  lea rdi, str.IOMedia
0x000008db      31f6           xor esi, esi
0x000008dd      e800000000     call 0x8e2
0x000008e2      4889c7         mov rdi, rax
0x000008e5      e800000000     call 0x8ea
0x000008ea      4989c5         mov r13, rax
0x000008ed      4d85ed         test r13, r13

and the result of ida is

00000000000008D4 48 8D 3D EE 06 00 00                                            lea     rdi, aIomedia 
00000000000008DB 31 F6                                                           xor     esi, esi
00000000000008DD E8 CE 21 00 00                                                  call    __ZN9IOService15serviceMatchingEPKcP12OSDictionary ; IOService::serviceMatching(char const*,OSDictionary *)
00000000000008E2 48 89 C7                                                        mov     rdi, rax
00000000000008E5 E8 C6 22 00 00                                                  call    __ZN9IOService19getMatchingServicesEP12OSDictionary ; IOService::getMatchingServices(OSDictionary *)
00000000000008EA 49 89 C5                                                        mov     r13, rax
00000000000008ED 4D 85 ED                                                        test    r13, r13

radare:
I see the same instructions and offsets so i dont see any issue related to analysis

What i see is some missing symbol names that dont appear in the calls. Without the bin its hard to say why they dont appear, maybe a bug in the bin parser or maybe ida have some flirt signatures that are magically applied under the hood.

But definitely this have nothing to do with analysis

IceAmber:
just as the code shows, the opcode of call in radare2 without address, just E8 00 00 00 00, while it is E8 CE 21 00 00 in ida

I can provide the bin if you need.

radare:
Thats because of the relocs and have nothing to do with analysis. The relocs are set by the bin parser. Yeah if you can share the bin i can take a look

IceAmber:
I had sent the bin to your email several days ago, had you got it?

radare:
uhm where did you send the file? can you send it again?

IceAmber:
I sent the bin to pancake@nopcode.org

radare:
Can you send it again? Because i remembwr seeing it but i cant find the mail or the file now :??

IceAmber:
[...]

PR #15445, opened 2019-11-11
Add support for Mach-O external relocs ##bin

mrmacete:
Add initial support for parsing and patching Mach-O relocs. It supports only the arch / reloc type combinations i've been able to test, but at least it's a starting point.

As i ported the patching code from ELF plugin, the way it works it's more or less the same. So to enable relocs patching, the binary should be opened like this:

```
r2 -e io.cache=true /path/to/binary
```

Fixes https://github.com/radareorg/radare2/issues/15279

### DIFF
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -2892,8 +2892,56 @@ static int reloc_comparator(struct reloc_t *a, struct reloc_t *b) {
 	return a->addr - b->addr;
 }
 
+static void parse_relocation_info (struct MACH0_(obj_t) *bin, RSkipList * relocs, ut32 offset, ut32 num) {
+	if (!num || !offset) {
+		return;
+	}
+
+	ut64 total_size = num * sizeof (struct relocation_info);
+	struct relocation_info *info = calloc (1, total_size);
+	if (!info) {
+		return;
+	}
+
+	if (r_buf_read_at (bin->b, offset, (ut8 *) info, total_size) < total_size) {
+		R_FREE (info);
+		return;
+	}
+
+	int i;
+	for (i = 0; i < num; i++) {
+		struct relocation_info a_info = info[i];
+		ut32 sym_num = a_info.r_symbolnum;
+		if (sym_num > bin->nsymtab) {
+			continue;
+		}
+
+		ut32 stridx = bin->symtab[sym_num].n_strx;
+		char *sym_name = get_name (bin, stridx, false);
+		if (!sym_name) {
+			continue;
+		}
+
+		struct reloc_t *reloc = R_NEW0 (struct reloc_t);
+		if (!reloc) {
+			return;
+		}
+
+		reloc->addr = offset_to_vaddr (bin, a_info.r_address);
+		reloc->offset = a_info.r_address;
+		reloc->ord = sym_num;
+		reloc->type = a_info.r_type; // enum RelocationInfoType
+		reloc->external = a_info.r_extern;
+		reloc->pc_relative = a_info.r_pcrel;
+		reloc->size = a_info.r_length;
+		r_str_ncpy (reloc->name, sym_name, 256);
+
+		r_skiplist_insert (relocs, reloc);
+	}
+}
+
 RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
-	RSkipList *relocs;
+	RSkipList *relocs = NULL;
 	ulebr ur = {NULL};
 	int wordsize = MACH0_(get_bits)(bin) / 8;
 	if (bin->dyld_info) {
@@ -3097,25 +3145,42 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
 			}
 		}
 		R_FREE (opcodes);
-	} else {
+	}
+
+	if (bin->symtab && bin->symstr && bin->sects && bin->indirectsyms) {
 		int j;
-		if (!bin->symtab || !bin->symstr || !bin->sects || !bin->indirectsyms) {
-			return NULL;
-		}
 		int amount = bin->dysymtab.nundefsym;
 		if (amount < 0) {
 			amount = 0;
 		}
-		relocs = r_skiplist_new ((RListFree) &free, (RListComparator) &reloc_comparator);
 		if (!relocs) {
-			return NULL;
+			relocs = r_skiplist_new ((RListFree) &free, (RListComparator) &reloc_comparator);
+			if (!relocs) {
+				return NULL;
+			}
 		}
 		for (j = 0; j < amount; j++) {
 			struct reloc_t *reloc = R_NEW0 (struct reloc_t);
+			if (!reloc) {
+				break;
+			}
 			if (parse_import_ptr (bin, reloc, bin->dysymtab.iundefsym + j)) {
 				reloc->ord = j;
+				r_skiplist_insert (relocs, reloc);
+			} else {
+				R_FREE (reloc);
 			}
 		}
+	}
+
+	if (bin->symtab && bin->dysymtab.extreloff && bin->dysymtab.nextrel) {
+		if (!relocs) {
+			relocs = r_skiplist_new ((RListFree) &free, (RListComparator) &reloc_comparator);
+			if (!relocs) {
+				return NULL;
+			}
+		}
+		parse_relocation_info (bin, relocs, bin->dysymtab.extreloff, bin->dysymtab.nextrel);
 	}
 beach:
 	return relocs;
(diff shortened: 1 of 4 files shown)


libr/bin/p/bin_mach0.c
```diff
@@ -503,6 +507,179 @@ static RBinInfo *info(RBinFile *bf) {
 	return ret;
 }
 
+static bool _patch_reloc(struct MACH0_(obj_t) *bin, RIOBind *iob, struct reloc_t * reloc, ut64 symbol_at) {
+	ut64 pc = reloc->addr;
+	ut64 ins_len = 0;
+
+	switch (bin->hdr.cputype) {
+	case CPU_TYPE_X86_64: {
+		switch (reloc->type) {
```
ret2libc: This looks quite bad, could you place this `switch` in a separate function `handle_patch_reloc_x86_64(struct reloc_t *reloc, ut64 *pc, ut64 *ins_len);`? Same for the aarch64/arm case. It will be easily inlined by the compiler.

libr/bin/p/bin_mach0.c
```diff
@@ -503,6 +507,179 @@ static RBinInfo *info(RBinFile *bf) {
...
+	struct MACH0_(obj_t) *bin = NULL;
+	RIOMap *g = NULL, *s = NULL;
+	HtUU *relocs_by_sym = NULL;
+	SdbListIter *iter;
+	RBinInfo *info;
+	int cdsz;
+	ut64 n_vaddr, vaddr, size, offset = 0;
+	RIODesc *gotr2desc = NULL;
+
+	if (!b) {
```
ret2libc: use `r_return_if_val (b, NULL);`

libr/bin/p/bin_mach0.c
```diff
@@ -503,6 +507,179 @@ static RBinInfo *info(RBinFile *bf) {
...
+			continue;
+		}
+		r_list_append (ext_relocs, reloc);
+	}
+	int num_ext_relocs = r_list_length (ext_relocs);
+	if (!num_ext_relocs) {
+		goto beach;
+	}
+
+	if (!io->cached) {
```
ret2libc: does it make sense to have this check before all the allocations? If a user does not set `io.cache=true` he cannot patch anything, so it should be better to check this asap.

libr/bin/p/bin_mach0.c
```diff
@@ -503,6 +507,179 @@ static RBinInfo *info(RBinFile *bf) {
...
+		eprintf ("Warning: run r2 with -e io.cache=true to fix relocations in disassembly\n");
+		goto beach;
+	}
+
+	info = obj ? obj->info: NULL;
```
ret2libc: `obj` can't be null, you already checked it before, so just do `obj->info`

libr/bin/p/bin_mach0.c
```diff
@@ -503,6 +507,179 @@ static RBinInfo *info(RBinFile *bf) {
...
+	if (!gotr2map) {
+		goto beach;
+	}
+	gotr2map->name = strdup (".got.r2");
+
+	if (!(ret = r_list_newf ((RListFree)free))) {
+		goto beach;
+	}
+	HtUUOptions opt = { 0 };
+	if (!(relocs_by_sym = ht_uu_new_opt (&opt))) {
```
ret2libc: I think you can just use `ht_uu_new0()` in this case.

libr/bin/p/bin_mach0.c
```diff
@@ -503,6 +507,179 @@ static RBinInfo *info(RBinFile *bf) {
...
+		goto beach;
+	}
+	vaddr = n_vaddr;
+	RListIter *liter;
+	r_list_foreach (ext_relocs, liter, reloc) {
+		ut64 sym_addr = 0;
```
ret2libc: define and assign in the same line

libr/bin/p/bin_mach0.c
```diff
@@ -503,6 +507,179 @@ static RBinInfo *info(RBinFile *bf) {
...
+		RBinImport *imp;
+		if (!(imp = import_from_name (b, (char*) reloc->name, bin->imports_by_name))) {
+			R_FREE (ptr);
+			goto beach;
+		}
+		ptr->vaddr = sym_addr;
+		ptr->import = imp;
+		r_list_append (ret, ptr);
+	}
+	if (!r_list_length (ret)) {
```
ret2libc: use `r_list_empty`

libr/bin/p/bin_mach0.c
mrmacete: The reason why the check is here is because these external relocs are pretty uncommon, so i want to avoid the user to get the warning all the time, and i moved it after we know if actually there are such type of relocs.

I can easily move it at the beginning, but then i'll have to remove the warning, which is not very user-friendly either.

ret2libc: I see. In ELF we do have the check at the beginning, but if it is uncommon for MACH0 I'm fine keeping it here.

libr/bin/p/bin_mach0.c
mrmacete: is this for avoiding the nested switch?

ret2libc: Yes. We are already full of nested switches everywhere and they become every day worse

mrmacete: let me see if i can refactor this in a way which isn't worse than that 😅

ret2libc: If you prefer nested switches I'm "ok" with that. Hopefully they will remain quite simple and short :)

libr/bin/p/bin_mach0.c
```diff
@@ -503,6 +507,180 @@ static RBinInfo *info(RBinFile *bf) {
 	return ret;
 }
 
+static bool init_patch_reloc(ut32 cpu_type, struct reloc_t * reloc, ut64 * pc, ut64 * ins_len) {
```
ret2libc: Ok sorry :D probably the double switch was just clearer than this... don't hate me :)

mrmacete: yeah i was thinking the same

mrmacete: got rid of that refactoring, i think we can refactor this when/if it gets more complex than that in the future.

ret2libc: Yep, that's fine!

libr/bin/format/mach0/mach0.c
```diff
@@ -2892,8 +2892,56 @@ static int reloc_comparator(struct reloc_t *a, struct reloc_t *b) {
 	return a->addr - b->addr;
 }
 
+static void parse_relocation_info (struct MACH0_(obj_t) *bin, RSkipList * relocs, ut32 offset, ut32 num) {
+	if (!num || !offset) {
+		return;
+	}
+
+	ut64 total_size = num * sizeof (struct relocation_info);
+	struct relocation_info *info = calloc (1, total_size);
```
radare: better use calloc (num, sizeof (...)) instead of the resuslt of a*b, it's safer to integer overflows

libr/bin/format/mach0/mach0.c
```diff
@@ -2892,8 +2892,56 @@ static int reloc_comparator(struct reloc_t *a, struct reloc_t *b) {
...
+	if (!info) {
+		return;
+	}
+
+	if (r_buf_read_at (bin->b, offset, (ut8 *) info, total_size) < total_size) {
+		R_FREE (info);
```
radare: just use 'free' no need to nullify the struct

libr/bin/format/mach0/mach0.c
ret2libc: R_news0 even better

libr/bin/p/bin_mach0.c
```diff
@@ -503,6 +507,174 @@ static RBinInfo *info(RBinFile *bf) {
...
+		return false;
+	}
+
+	ut64 val = symbol_at;
+	if (reloc->pc_relative) {
+		val = symbol_at - pc - ins_len;
+	}
+
+	ut8 buf[8];
+	r_write_ble(buf, val, false, reloc->size * 8);
```
radare: mmissing space before (

libr/bin/p/bin_mach0.c
```diff
@@ -503,6 +507,174 @@ static RBinInfo *info(RBinFile *bf) {
...
+	RBinObject *obj = NULL;
+	struct MACH0_(obj_t) *bin = NULL;
+	RIOMap *g = NULL, *s = NULL;
+	HtUU *relocs_by_sym = NULL;
+	RIODesc *gotr2desc = NULL;
+
+	r_return_val_if_fail (b, NULL);
+
+	io = b->iob.io;
+	if (!io || !io->desc)
```
radare: missing braces

PR #15470, opened 2019-11-15
Fixes to graphviz related code

erfur:
`d05e816 `: The current block in graphviz output is highlighted even when `graph.gv.current` is `false`. Added a logical check to fix it.
`f5e3f60 `: else if blocks are off by one indentation level

### DIFF
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -1595,7 +1595,7 @@ static int core_anal_graph_construct_nodes (RCore *core, RAnalFunction *fcn, int
                                                 : (current && color_current)
                                                 ? pal_curr
                                                 : pal_box4;
-                                        const char *fill_color = (current || label_color == pal_traced)? pal_traced: "white";
+                                        const char *fill_color = ((current && color_current) || label_color == pal_traced)? pal_traced: "white";
                                         nodes++;
                                         if (is_star) {
                                                 char *title = get_title (bbi->addr);
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -1319,7 +1319,7 @@ static int core_anal_graph_construct_edges (RCore *core, RAnalFunction *fcn, int
                                         free(from);
                                         free(to);
                                 } else {
-                                        r_cons_printf ("\t\"0x%08"PFMT64x"\" -> \"0x%08"PFMT64x"\" "
+                                        r_cons_printf ("        \"0x%08"PFMT64x"\" -> \"0x%08"PFMT64x"\" "
                                                         "[color=\"%s\"];\n", bbi->addr, bbi->jump,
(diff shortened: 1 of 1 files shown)


libr/core/canal.c
```diff
@@ -1311,36 +1311,36 @@ static int core_anal_graph_construct_edges (RCore *core, RAnalFunction *fcn, int
...
-                                                r_cons_printf ("age %s %s\n", from, to);
-                                        } else {
-                                                r_cons_printf ("\t\"0x%08"PFMT64x"\" -> \"0x%08"PFMT64x"\" "
-                                                               "[color=\"%s\"];\n", bbi->addr, bbi->jump,
-                                                               bbi->fail != -1 ? pal_jump : pal_trfa);
-                                                core_anal_color_curr_node (core, bbi);
-                                        }
+                        } else if (!is_json && !is_keva) {
+                                if (is_star) {
+                                        char *from = get_title (bbi->addr);
```
radare: this is leaking you may  free from and to after calling printf() below

libr/core/canal.c
```diff
@@ -1311,36 +1311,36 @@ static int core_anal_graph_construct_edges (RCore *core, RAnalFunction *fcn, int
...
-                                        } else {
-                                                r_cons_printf ("\t\"0x%08"PFMT64x"\" -> \"0x%08"PFMT64x"\" "
-                                                               "[color=\"%s\"];\n", bbi->addr, bbi->fail, pal_fail);
-                                                core_anal_color_curr_node (core, bbi);
-                                        }
+                                                    "  <img class=\"connector-end\" src=\"img/arrow.gif\"/></div>\n",
+                                                    bbi->addr, bbi->fail);
+                        } else if (!is_keva && !is_json) {
+                                if (is_star) {
+                                        char *from = get_title (bbi->addr);
```
radare: same here

libr/core/canal.c
```diff
@@ -1311,36 +1311,36 @@ static int core_anal_graph_construct_edges (RCore *core, RAnalFunction *fcn, int
...
+                                        char *to = get_title (bbi->fail);
+                                        r_cons_printf ("age %s %s\n", from, to);
+                                } else {
+                                        r_cons_printf ("\t\"0x%08"PFMT64x"\" -> \"0x%08"PFMT64x"\" "
```
radare: i would prefer to use spaces rather than tabs

libr/core/canal.c
erfur: I just wanted to fix the indentation of this block but sure why not :)

libr/core/canal.c
erfur: You mean the `\t` inside the format string?

radare:
yes. we try to avoid tabs in the output of the commands

PR #15490, opened 2019-11-21
Fix anal.timeout calculation in r_cons_break_timeout

MNayer:
anal.timeout is calculated by r_sys_now() + timeout value in seconds. Since r_sys_now() provides the current time in the format [44 bit seconds][20 bit micro seconds], the timeout value must be shifted 20 bits to the left (as done in r_sys_now) instead of multiplying by 1000000. Besides, timeout is a signed integer so even "short" timeout values like 1h (3600 seconds) result in an integer overflow when multiplied by 1000000. This causes unexpected behavior.


radare:
thx

PR #15501, opened 2019-11-25
Allow breaking for WinDbg ##debug

GustavoLCR:
Thanks @yossizap for helping


yossizap:
FYI - Tested on Gustavo's windows host with a Win XP 32bit VM and on my ubuntu linux host with Win7 SP2 64bit qemu.

PR #15504, opened 2019-11-26
Partially decouple Tasks from RCore

thestr4ng3r:
Basically just pulling everything task-related from RCore into a struct.


radare:
this is wrong.. but i understand the "partial" part of your commit message:

``	int res = r_core_task_run_sync (&core->tasks, task);``

should be r_task... not r_core_task..

thestr4ng3r:
Yes, in the long run RTask can be a standalone module. But this is only a first step.

PR #15511, opened 2019-11-27
Make Task Scheduling Explicit

thestr4ng3r:
This removes automatic scheduling from `r_cons_is_breaked()` and makes it explicit.
This means most commands will just block everything for now when run in a task, but by **carefully** adding yields, they can be made concurrent again while keeping everything safe.


radare:
yieldiful

PR #15516, opened 2019-11-27
Fix handling of many seek history items and other edge cases.

karliss:


karliss:
Yes, cmd_seek has it's own json code and I added test cases that check `sj` is working.

libr/io/undo.c
```diff
@@ -127,7 +127,7 @@ R_API RList *r_io_sundo_list(RIO *io, int mode) {
 
 	idx = io->undo.idx;
 	start = (idx - undos + R_IO_UNDOS) % R_IO_UNDOS;
-	end = (idx + redos + 1) % R_IO_UNDOS;
+	end = (idx + redos + 1 - 1) % R_IO_UNDOS; // +1 slot for current position, -1 due to inclusive end
```
radare: +1-1? just drop that

libr/io/undo.c
```diff
@@ -170,21 +163,27 @@ R_API RList *r_io_sundo_list(RIO *io, int mode) {
 			if (list) {
 				RIOUndos  *u = R_NEW0 (RIOUndos);
 				if (u) {
-					memcpy (u, undo, sizeof (RIOUndos));
+					if (!(j == undos && redos == 0)) {
+						// Current position gets pushed before seek, so there
+						// is no valid offset when we are at the end of list.
+						memcpy (u, undo, sizeof (RIOUndos));
```
radare: wlll be good to abstract thihs inito a function named r_io_undos_clone(), but with the comment i guess its fine too

libr/io/undo.c
```diff
@@ -83,9 +83,9 @@ R_API void r_io_sundo_push(RIO *io, ut64 off, int cursor) {
 	if (!io->undo.s_enable) {
 		return;
 	}
-	// the first insert
-	if (io->undo.idx > 0) {
-		undo = &io->undo.seek[io->undo.idx - 1];
+	// don't push duplicate seek
+	if (io->undo.undos > 0) {
+		undo = &io->undo.seek[(io->undo.idx - 1 + R_IO_UNDOS) % R_IO_UNDOS];
```
radare: not sure if we want to make the undo history cyclic

karliss: Are we thinking about the same thing? It was already storing history in a cyclic buffer before my changes but with bugs. There were even tests expecting this.  With fixed size history deleting the oldest entry  when there is no space for new ones is only sane behaviour.

libr/io/undo.c
karliss: I intentionally left the intermediate steps to make it clear what components are taken into account. Might be better as 'idx + (1 + redos) -1' as current element is before redo.

PR #15523, opened 2019-11-28
Improved linux native debug `dp` and added current to `dpj`

yossizap:


libr/debug/p/native/linux/linux_debug.c
```diff
@@ -611,6 +615,46 @@ RDebugPid *fill_pid_info(const char *info, const char *path, int tid) {
...
+		r_list_free (list);
+		return NULL;
+	}
+	while ((de = readdir (dh))) {
+		memset (buf, 0, sizeof (buf));
+		memset (info, 0, sizeof (buf));
+		uid = 0;
+		st = ' ';
+		// For each existing pid file
+		i = atoi (de->d_name);
```
radare: you can define and assign in the same line

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -611,6 +615,46 @@ RDebugPid *fill_pid_info(const char *info, const char *path, int tid) {
...
+	int i, thid, uid;
+	RDebugPid *pid_info;
+	dh = opendir ("/proc");
+	if (!dh) {
+		r_sys_perror ("opendir /proc");
+		r_list_free (list);
+		return NULL;
+	}
+	while ((de = readdir (dh))) {
+		memset (buf, 0, sizeof (buf));
```
radare: isnt setting the first byte to 0 enough?

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -611,6 +615,46 @@ RDebugPid *fill_pid_info(const char *info, const char *path, int tid) {
 	return pid_info;
 }
 
+RList *linux_pid_list(int pid, RList *list) {
+	list->free = (RListFree)&r_debug_pid_free;
+	DIR *dh;
+	struct dirent *de;
+	char *ptr, st, buf[1024], info[1024];
```
radare: isnt this 1024 handled by a constant somewhere? like PATHMAX or so?

yossizap: PATH_MAX is 4096 usually but it's probably wiser to switch to it since cmdline could get to that length.

libr/debug/p/native/linux/linux_debug.c
yossizap: They were reassigned in each iteration to avoid errors. Anyways, no longer needed.

libr/debug/p/native/linux/linux_debug.c
radare: use the heap then, allocate once before the loop. huge heap sizes use to be a source for problems in random operating systems

PR #15525, opened 2019-11-28
Fix overflow that broke remote debugger reopening

yossizap:
`opt->sz` is initialized with `r_buf_size` at `r_bin_open_io` using an io buffer if `r_bin_open_io` can't open a file buffer. Since the debuggers returned unsigned values on lseek(used by io buffers to determine size) to opt->sz which is signed, opt->sz would overflow and contain a negative value, causing r_bin_open_buf to fail.

Went ahead and modified SEEK_END values for all debuggers even though this should only affect remote debuggers. ST64_MAX should be enough.

Note: gdbserver worked before the changes if it was on the same machine, tested with gdbserver on a different machine and wingdb.

### DIFF
--- a/libr/io/p/io_gdb.c
+++ b/libr/io/p/io_gdb.c
@@ -156,7 +156,7 @@ static ut64 __lseek(RIO *io, RIODesc *fd, ut64 offset, int whence) {
 		io->off += offset;
 		break;
 	case R_IO_SEEK_END:
-		io->off = UT64_MAX;
+		io->off = ST64_MAX;
 	}
 	return io->off;
 }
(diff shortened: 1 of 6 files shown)


XVilka:
CI failure seems unrelated:
```
[XX] db/cmd/cmd_task & + grep  31
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -i /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp-32846uJMeQc9Zk6tv -
"& ?e Hello\nfrom\na task!~task"
&& 1
&= 1
+Hello
+from
 a task!
+Hello
+from
+a task!
EXPECT=<<EOF
Hello
from
a task!
Hello
from
a task!
```

radare:
this broken test is probably related to not being rebased

PR #15529, opened 2019-11-29
gdbclient reduces risk of use after free down the road ##gdbclient

devnexen:


radare:
i like the concept of "reduce the risk of use after free" XD

PR #15537, opened 2019-12-01
Add breaking to `ood` in linux to avoid execvp freezing ##debug

yossizap:
Execvp may hang until a signal is received on some systems(reproduced in arch linux) in case the file can't be executed specifically because of a broken symlink or a different arch(32bit on 64bit without support for example) instead of just failing.

This issue was encountered by @ITAYC0HEN, it seems to be solved with this fix. I couldn't replicate it on Ubuntu so I tested by adding a long sleep before execvp.

### DIFF
--- a/libr/io/p/io_debug.c
+++ b/libr/io/p/io_debug.c
@@ -4,6 +4,7 @@
 #include <r_io.h>
 #include <r_lib.h>
 #include <r_util.h>
+#include <r_cons.h>
 #include <r_debug.h> /* only used for BSD PTRACE redefinitions */
 #include <string.h>
 
@@ -457,6 +458,7 @@ static void fork_child_callback(void *user) {
 static int fork_and_ptraceme_for_unix(RIO *io, int bits, const char *cmd) {
 	int ret, status, child_pid;
 	bool runprofile = io->runprofile && *(io->runprofile);
+	void *bed = NULL;
 	fork_child_data child_data;
 	child_data.io = io;
 	child_data.bits = bits;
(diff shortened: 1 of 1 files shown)


ret2libc:
I don’t like the solution. I think the problem should first be understood and then fixed. Could you give the exact steps to try to reproduce?

yossizap:
The error that he got after interrupting the process was "No such file or directory" - this could be reproduced by running a 32bit binary without 32bit lib support(I couldn't reproduce it on my machine but that's what @ITAYC0HEN had issues with) or a broken symlink(doesn't happens in Ubuntu but it might be possible in some systems), any other case is already caught by r2. Only happens in Cutter.

There's another hypothesis that the recently added stdin redirection in Cutter in addition to opening an invalid file left execvp waiting for an EOF. He didn't get to test it yet.

In any case, it's better to have an async waitpid that you can break instead of an infinitely blocking wait. This is a libc/kernel level bug that we can't do much about imo.

libr/io/p/io_debug.c
```diff
@@ -4,6 +4,7 @@
 #include <r_io.h>
 #include <r_lib.h>
 #include <r_util.h>
+#include <r_cons.h>
```
radare: i dont like using rcons from rio

radare:
uhm wonder why its not red, but i guess that the use of rcons is now legit from r_io, comments anyone?

radare:
im merging for now, but i would prefer to find a better solution to this

PR #15544, opened 2019-12-04
Added `dp` processes info support to gdbr ##debug

yossizap:
Most servers/clients should have xml support by now so it should behave like `dp` in any other debugger. ThreadsInfo is used as a fallback.
vFile is the only way to get detailed pid info unfortunately.


yossizap:
Test failures seem unrelated

radare:
indeed

PR #15554, opened 2019-12-07
Remove broken Threading Code from main for #12996

thestr4ng3r:


libr/main/radare2.c
```diff
@@ -818,7 +746,7 @@ R_API int r_main_radare2(int argc, char **argv) {
 
 	prj = r_config_get (r.config, "prj.name");
 	if (prj && *prj) {
-		r_core_project_open (&r, prj, threaded);
+		r_core_project_open (&r, prj, false);
```
radare: maybe you can also kill this threaded project opening flag and we can add RCore->threads to sspecify how many threads we want to use to do things.. globally

radare:
thanks for the answer (in telegram) let's go for it

PR #15555, opened 2019-12-08
Fixes for argument/variables analysis

GustavoLCR:
Before:

image

After:

image

----
Also, is there a way to merge 2 variables? For example in the second image var_18h and var_18h_3 are the same

### DIFF
--- a/test/new/db/cmd/types
+++ b/test/new/db/cmd/types
@@ -94,11 +94,11 @@ EOF
 EXPECT=<<EOF
 | ; var struct JNINativeInterface @ sp+0x12c
 | 0x00000f74      4b93           str r3, [sp + JNINativeInterface]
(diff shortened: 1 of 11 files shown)


radare:
no, there's no command or api to merge local variables, it will be good to have it

radare:
can you update the tests after rebasing?

test/new/db/cmd/types
```diff
@@ -1259,7 +1259,7 @@ af
 aaft
 afv
 EXPECT=<<RUN
-var int32_t var_0h @ sp+0x0
+var int32_t var_1454h @ sp+0x0
```
radare: uhm is this correct ?

GustavoLCR: Stack grows to that size:

image

PR #15559, opened 2019-12-08
Fix a bug of the tab

Vane11ope:
reproduce:
tT and q then r2 crashes so fixed it.


libr/core/panels.c
```diff
@@ -1,4 +1,4 @@
-/* Copyright radare2 2014-2019 - Author: pancake, vane11ope, Kirils Solovjovs */
```
xarkes: Why would you remove one's name? :'D

Vane11ope: what has he done to the mode anwyay

PR #15608, opened 2019-12-10
Use ONE_STREAM=1 instead and assume it sets scr.onestream

kazarmy:
Clone of radareorg/radare2-regressions#2037, which apparently is frozen in Archiveland ＦＯＲＥＶＥＲ．


kazarmy:
Closes radareorg/radare2-regressions#2037 ?

PR #15609, opened 2019-12-10
Vsuite

radare:


test/new/r2r-v/r2r.v
```diff
@@ -496,7 +567,19 @@ fn (r2r mut R2R)load_tests() {
 	dirs := os.ls(db_path) or { panic(err) }
 	for dir in dirs {
 		if dir == 'archos' {
-			println('TODO: archos tests')
+			$if x64 {
```
ret2libc: isn't there a if/elseif in V?
In python, this would look much better with something like:
```
if linux:
  ....
elif macos:
  ....
elif openbsd:
  ....
else:
  print('not supported')
```

radare: not yet. im filling issues and contributing to the compiler, and this is one of the limitations i found

test/new/db/archos/linux-x64/dbg_bps
```diff
@@ -1,7 +1,7 @@
 NAME='dbg.dc+ds.swstep'
 FILE=../bins/elf/analysis/elf-nx
 ARGS=-d
-CMDS=<<EXPECT
+CMDS=<<EOF
```
ret2libc: why this change? wasn't the whole purpose of using <<EXPECT to avoid having to introduce EOF?

test/new/r2r-v/r2r.v
radare: this $if thing is a compile time conditional, so the final binary will only contian this code if arch/os matches

test/new/db/archos/linux-x64/dbg_bps
radare: the <<EXPECT thing is was just for backward compat for lazines, i want the parser to be simple and be able to move K=V easily upp and down without breaking because the next statement is not a keyword. this was kind of a hack and we must kill it, the asm suite also have a bunch of old syntaxes that im trying to avoid. the v suite is throwing warnings at them

test/new/r2r-v/r2r.v
radare: follow it here: https://github.com/vlang/v/issues/3037

ret2libc: ok

PR #15610, opened 2019-12-11
Add PIE flag check in kernelcache detection ##bin

mrmacete:
In this way 64-bit SEP kernels aren’t wrongly detected as XNU kernel caches.


libr/bin/format/mach0/mach064_is_kernelcache.c
```diff
@@ -8,6 +8,11 @@ static bool is_kernelcache_buffer(RBuffer *b) {
 		return false;
 	}
 
+	ut32 flags = r_buf_read_le32_at (b, 24);
+	if (!(flags & 0x200000)) {
```
radare: maybe worth to use a named constant instead of hardcoding 0x200000 for readability

PR #15612, opened 2019-12-11
Fix first r_anal_op call after asm.bits change

thestr4ng3r:


radare:
do you have a test case?

thestr4ng3r:
yes, look at the diff. Before the code changes, the second ao just fails hard.

radare:
travis is red

thestr4ng3r:
afl ordering in that test is a bit different and the analysis for entry0 changes (left is old, right is new):
Bildschirmfoto vom 2019-12-11 13-48-56
The new behavior makes more sense to me, the old one was probably caused by this bug too.

PR #15616, opened 2019-12-12
r2r.v: Support running of just 1 cmd testfile

kazarmy:


test/new/r2r-v/r2r.v
```diff
@@ -47,12 +47,14 @@ pub fn main() {
 	println('Loading tests')
 	os.chdir('..')
 	mut r2r := R2R{}
-	r2r.load_tests()
-	// add option to specify which tests to run
+	r2r.load_tests(targets)
```
radare: what about moving targets into r2r? instead of passing it as argument everywhere?

test/new/r2r-v/r2r.v
```diff
@@ -47,12 +47,14 @@ pub fn main() {
...
 	if run_tests {
-		r2r.run_jsn_tests(threads)
-		r2r.run_asm_tests(threads)
-		r2r.run_fuz_tests(threads)
+		if targets.len <= 1 {  // WIP
+			r2r.run_jsn_tests(threads)
+			r2r.run_asm_tests(threads)
+			r2r.run_fuz_tests(threads)
+		}
 		r2r.run_cmd_tests(threads)
```
radare: imho we need a way to specify which kind of tests we want to run too and maybe threads can also go into r2r

kazarmy:
just introduced myself to V today ... will have to study it a bit more, tomorrow

test/new/r2r-v/r2r.v
```diff
@@ -577,31 +582,31 @@ fn (r2r mut R2R)load_jsn_tests(testpath string) {
...
+						r2r.load_cmd_tests('${db_path}/${dir}/darwin-x64/', targets)
 					} $else {
 						eprintln('Warning: archos tests not supported for current platform')
 					}
 				}
 			} $else {
 				eprintln('Warning: archos tests not supported for current platform')
 			}
-		} else if dir == 'json' {
+		} else if dir == 'json' && targets.len <= 1 {
```
radare: i prefer <2 than <=1

radare:
nice!

test/new/r2r-v/r2r.v
kazarmy: > imho we need a way to specify which kind of tests we want to run too

agree

> maybe threads can also go into r2r

i think so too but that should be in a different pr

test/new/r2r-v/r2r.v
kazarmy: ok c9c89f06685e9da01d6cf9472089e608a5cfb00f

test/new/r2r-v/r2r.v
kazarmy: ok 7f51ad0671144c020a6ba2db9027dbeca85401b7

kazarmy:
changes ok?

PR #15621, opened 2019-12-13
Add 64-bit iOS SEP xtr plugin ##bin

mrmacete:
Supports only 64-bit SEP firmwares.

It's based on a single sample, i'm glad to make it more generic if someone has more samples to share.

test with `rabin2 -x sep-firmware.n841.RELEASE.im4p-dec`

### DIFF
--- a/libr/bin/p/bin_xtr_sep64.c
+++ b/libr/bin/p/bin_xtr_sep64.c
@@ -11,46 +11,46 @@
 
 /* at offset 0x10f8 (pointer to it stored right after "legion2") */
 typedef struct _RSepHdr64 {
-    ut8 kernel_uuid[16];
-    ut64 unknown0;
-    ut64 kernel_base_paddr;
-    ut64 kernel_max_paddr;
-    ut64 app_images_base_paddr;
-    ut64 app_images_max_paddr;
-    ut64 paddr_max; /* size of SEP firmware image */
-    ut64 unknown1;
-    ut64 unknown2;
-    ut64 unknown3;
-    ut64 init_base_paddr;
-    ut64 unknown4;
-    ut64 unknown5;
-    ut64 unknown6;
-    ut64 unknown7;
-    ut64 unknown8;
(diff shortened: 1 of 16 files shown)


libr/bin/p/bin_xtr_sep64.c
```diff
@@ -0,0 +1,493 @@
...
+    ut64 size_data;
+    ut64 virt;
+    ut64 entry;
+    ut64 unknown4;
+    ut64 unknown5;
+    ut64 unknown6;
+    ut32 minus_one;
+    ut32 unknown7;
+    char app_name[16];
+    ut8 app_uuid[16];
```
radare: Use tabs

libr/bin/p/bin_xtr_sep64.c
```diff
@@ -0,0 +1,493 @@
...
+		}
+		return res;
+	}
+	return NULL;
+}
+
+static RSepXtr64Ctx * sep64_xtr_ctx_new(RBuffer * buf) {
+	RSepHdr64 * hdr = NULL;
+	RSepApp64 * apps = NULL;
+	RSepXtr64Ctx * ctx = NULL;
```
radare: Define and assign in the same line below

libr/bin/p/bin_xtr_sep64.c
```diff
@@ -0,0 +1,493 @@
...
+beach:
+	free (hdr);
+	free (apps);
+	free (ctx);
+
+	return NULL;
+}
+
+static void sep64_xtr_ctx_free(void * p) {
+	if (p == NULL) {
```
radare: Use !p instead if ==NULL

libr/bin/p/bin_xtr_sep64.c
mrmacete: hmm moving the declarations down will complicate the `goto beach;` "pattern"

PR #15629, opened 2019-12-14
r2r.v: Fix test skipping

kazarmy:


test/new/r2r-v/r2r.v
```diff
@@ -497,19 +497,19 @@ fn (r2r mut R2R)run_asm_tests() {
...
 				c = r2r.threads
 			}
 		}
 		if at.mode.contains('d') {
-			if c-- > 0 {
-				r2r.wg.add(1)
-				r2r.run_asm_test(at, true)
-			} else {
+			r2r.wg.add(1)
+			r2r.run_asm_test(at, true)
```
radare: We cant thread this because its using the native api. So i dont see much sense in using the wh here. Unless we define a way to tell that a test must be executed or not bwfore calling

kazarmy: I added a check for nil `r2r.r2` (f7b6d8f25ad7c8f8ee7b99e1222fa0c271972e86). Seems a shame to waste all that code.

PR #15632, opened 2019-12-15
Some more helpful messages for anal.verbose ##anal

radare:


libr/core/canal.c
```diff
@@ -1849,6 +1849,9 @@ R_API int r_core_anal_fcn(RCore *core, ut64 at, ut64 from, int reftype, int dept
 
 	if (core->io->va) {
 		if (!r_io_is_valid_offset (core->io, at, !core->anal->opt.noncode)) {
+			if (core->anal->verbose) {
```
XVilka: Probably we need a macro for that.

radare:
We need to overengineer eprintf to make it work on cutter and r2pipe

Ideas?

xarkes:
I personally don't like `eprintf` and I believe it should be used only as a shortcut for local debugging.
I think it would be better to use `r_log` or similar API.

ret2libc:
Yes indeed we have r_log and some nice R_LOG macros as well. @radare we should use those IMHO, or build something on top of that to enable/disable logs "by module". We already have log.level, log.file, we could something like `log.modules` and if you set it to `log.modules=analysis,core,rbin.elf`

radare:
but how can we make rlog output accessible from r2pipe for example? addinga c omand to collect all that buffer? eprintf use to be "streamy" not buffered like cons, so we maybe need a separate channel to do that

PR #15634, opened 2019-12-15
Fixes for variables

GustavoLCR:


libr/reg/reg.c
```diff
@@ -202,6 +202,40 @@ static int regcmp(RRegItem *a, RRegItem *b) {
...
+	RRegItem *r;
+	RList *all = r_list_newf (NULL);
+	for (i = 0; i < R_REG_TYPE_LAST; i++) {
+		r_list_foreach (reg->regset[i].regs, iter, r) {
+			r->role = R_REG_NAME_LAST;
+		}
+	}
+	for (i = 0; i < R_REG_TYPE_LAST; i++) {
+		r_list_foreach (reg->regset[i].regs, iter, r) {
+			if (r->role != R_REG_NAME_LAST) {
```
XVilka: Maybe decouple these nested loops? Make a separate small functions probably?

ret2libc:
> Not clear to me that the new tests output is correct. Can someone else take a look?

@sivaramaaa  maybe?

libr/reg/reg.c
radare: im just caring about correctness of the result

libr/reg/reg.c
```diff
@@ -202,6 +202,40 @@ static int regcmp(RRegItem *a, RRegItem *b) {
 	return offa > offb;
 }
 
+R_API void r_reg_rerole(RReg *reg) {
```
ret2libc: This looks like an expensive operation. Shall it be always done? Also, please consider if existing API like `r_reg_get` or others can be used instead of hand-made for-loops.

radare:
cant find any obvious regression in my tests, ok to merge? someone else wants to test it a bit' do you have any example of how this change is improving the var analysis? can you add a test? i'll merge it now before my flight takes off

libr/reg/reg.c
radare: oh wait, ill wait for this to be fixed, my plan is to release later today or tomorrow. ill be fixing more things during the flight

GustavoLCR: i killed it, found another way

GustavoLCR:
> do you have any example of how this change is improving the var analysis?

@radare This mainly fixes cases where SP changes mid function, and the analysis would get confused because it used the offset to SP to name/identify vars, and would result in variables created for the same area in the stack, or it would think different areas were the same variable

PR #15637, opened 2019-12-15
Fix warnings in or1k_disas.c

thestr4ng3r:


radare:
Will be good if we can get the warnings of the compilation as a comment for each pr. Something more to add to the ci

PR #15645, opened 2019-12-16
Function r_anal_cc_arg should just return NULL when nothing is available

ret2libc:
Callers should be able to handle NULLs


ret2libc:
This should fix https://travis-ci.com/radareorg/radare2/jobs/267480273#L2239

ret2libc:
Closing, see https://github.com/radareorg/radare2/pull/15646

PR #15646, opened 2019-12-16
Function r_anal_cc_arg should just return NULL when nothing is available

ret2libc:
Callers should be able to handle NULLs

(a new PR, sorry... this should enable the ASAN tests because of the `asan-*` branch name)


radare:
lgtm

ret2libc:
ASAN job is now green. See  https://github.com/radareorg/radare2/pull/15646/checks?check_run_id=350445884

PR #15655, opened 2019-12-16
Fix native debug not working with files that were opened as RW

### LINKED ISSUE #15650: Can't debug files that were opened in write-mode
### Work environment

| File format of the file you reverse (mandatory)      | Any
| Architecture/bits of the file (mandatory)            | Any
| r2 -v full output, **not truncated** (mandatory)     | recent from git

### Expected behavior

There shouldn't be a problem to debug file that is opened in write mode. Even if the file was opened in write mode and modified, running `ood` should success in debugging the file

### Actual behavior

Debugging files that are currently opened in write mode is not possible. Probably r2 kept the fd open and blocks the execution.

```
$ r2 -w /tmp/readln 
Module version mismatch /home/beet/.local/share/radare2/plugins/core_pdd.so (3.7.0) vs (4.1.0-git)
 -- Quantum dissasemble: it's there as long as you don't observe it
[0x000010d0]> doo
Could not execvp: Text file busy
Could not execvp: Text file busy
ptrace (PT_ATTACH): No such process
r_core_file_reopen: Cannot reopen file: dbg:///tmp/readln  with perms 0x7, attempting to open read-only.
[0x00000000]> 

```

- Open file in write mode `r2 -w <file>`
- Try to re-open in deubg mode using `ood`
- Fail

yossizap:
The FD was kept open and ptrace was prevented from opening it with the error "Could not execvp: Text file busy".

closes #15650

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -1101,10 +1101,21 @@ R_API void r_core_file_reopen_remote_debug(RCore *core, char *uri, ut64 addr) {
 R_API void r_core_file_reopen_debug(RCore *core, const char *args) {
 	RCoreFile *ofile = core->file;
 	RIODesc *desc;
+
 	if (!ofile || !(desc = r_io_desc_get (core->io, ofile->fd)) || !desc->uri) {
 		eprintf ("No file open?\n");
 		return;
 	}
+
+	// Reopen the original file as read only since we can't open native debug while the
+	// file is open with write permissions
+	if (!(desc->plugin && desc->plugin->isdbg) && (desc->perm & R_PERM_W)) {
+		eprintf ("Cannot debug file (%s) with permissions set to 0x%x.\n"
+			"Reopening the original file in read-only mode.\n", desc->name, desc->perm);
+		r_io_reopen (core->io, ofile->fd, R_PERM_R, 644);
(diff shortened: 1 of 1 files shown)


yossizap:
Fails on db/archos/linux-x64/dbg_oo but I am really not sure what is the purpose of those tests. Went back as far as 7e4157bc2 and I can't get them to pass locally. This doesn't seem related to my changes.

ITAYC0HEN:
shouldn't it be restored to Write mode while the debugger was finished? I think yes (though very risky because of the instability of many things that might modify user's data

yossizap:
>though very risky because of the instability of many things that might modify user's data

That was my thought, would prefer having the user run `oo+`.

ITAYC0HEN:
if so, Let them know then. don't expect them to guess what has changed

yossizap:
Rebased and added @ITAYC0HEN's request. Hopefully the test will pass now.

ITAYC0HEN:
Still failing. Painful :(

radare:
@ITAYC0HEN if its green and works for you please merge

williamcorney:
I am having the issue described in this post in 2023.  I don't know if the issue was never fixed or has resurfaced again ?

I am working around the problem by renaming the app without a space and then functions as expected.

PR #15659, opened 2019-12-17
Refactor signal handling for FreeBSD ##r_debug

devnexen:
Solving r_debug_handle_signals build warning issue.

### DIFF
--- a/libr/debug/p/debug_native.c
+++ b/libr/debug/p/debug_native.c
@@ -110,6 +110,39 @@ static int r_debug_handle_signals(RDebug *dbg) {
 #if __linux__
 	return linux_handle_signals (dbg);
 #elif __KFBSD__
+	// Trying to figure out a bit by the signal
+	struct ptrace_lwpinfo linfo = {0};
+	siginfo_t siginfo;
+	int ret = ptrace (PT_LWPINFO, dbg->pid, (char *)&linfo, sizeof (linfo));
+	if (ret == -1) {
+		if (errno == ESRCH) {
+			dbg->reason.type = R_DEBUG_REASON_DEAD;
+			return 0;
+		}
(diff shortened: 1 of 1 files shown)


libr/debug/p/debug_native.c
```diff
@@ -110,6 +110,39 @@ static int r_debug_handle_signals(RDebug *dbg) {
...
+			return 0;
+		}
+	}
+
+	siginfo = linfo.pl_siginfo;
+	dbg->reason.type = R_DEBUG_REASON_SIGNAL;
+	dbg->reason.signum = siginfo.si_signo;
+
+	switch (dbg->reason.signum) {
+		case SIGABRT:
```
radare: Bad indent

libr/debug/p/debug_native.c
```diff
@@ -110,6 +110,39 @@ static int r_debug_handle_signals(RDebug *dbg) {
...
+	siginfo_t siginfo;
+	int ret = ptrace (PT_LWPINFO, dbg->pid, (char *)&linfo, sizeof (linfo));
+	if (ret == -1) {
+		if (errno == ESRCH) {
+			dbg->reason.type = R_DEBUG_REASON_DEAD;
+			return 0;
+		}
+		r_sys_perror ("ptrace PTRACE_LWPINFO");
+		return -1;
+	} else {
```
radare: No need to else after return

libr/debug/p/debug_native.c
devnexen: You sure ? seems similar to other switch/cases.

libr/debug/p/debug_native.c
devnexen: true

PR #15660, opened 2019-12-17
Fix typo

kazarmy:
Only fixes the typo.

### DIFF
--- a/libr/main/rabin2.c
+++ b/libr/main/rabin2.c
@@ -597,7 +597,7 @@ R_API int r_main_rabin2(int argc, char **argv) {
 			r_config_unserialize (core.config, config_sdb, NULL);
 			sdb_free (config_sdb);
 		} else {
-			printf ("Cannot open file specified in R2_CONFIG\n");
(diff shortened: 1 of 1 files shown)


libr/main/rabin2.c
```diff
@@ -597,7 +597,7 @@ R_API int r_main_rabin2(int argc, char **argv) {
 			r_config_unserialize (core.config, config_sdb, NULL);
 			sdb_free (config_sdb);
 		} else {
-			printf ("Cannot open file specified in RADARE2_CONFIG\n");
+			printf ("Cannot open file specified in R2_CONFIG\n");
```
radare: Use eprintf

ret2libc: Why? I see also https://github.com/radareorg/radare2/pull/15660/files#diff-f80754f1fb53839b3da69480345affbbR761 uses printf. What were the reasons to use eprintf? I don't remember.

kazarmy: > Use eprintf

ok e410c28ff1badc07ac43ac9029b73a9a316cfa87

PR #15665, opened 2019-12-17
Refactor debug native's wait function ##debug

yossizap:
Separate windows, linux and *bsd/apple since they have vastly different logic and the ifdef jungle implementation was too confusing. 

Tested in Linux and Windows, just in case but there aren't any code changes.

@devnexen Didn't really change anything but I would appreciate it if you could check that it's still ok in BSD. Also, we should start moving bsd to native/bsd/debug_bsd to make debug_native more easily maintainable.

### DIFF
--- a/libr/debug/p/debug_native.c
+++ b/libr/debug/p/debug_native.c
@@ -475,15 +475,19 @@ static bool tracelib(RDebug *dbg, const char *mode, PLIB_ITEM item) {
  *
  * Returns R_DEBUG_REASON_*
  */
+#if __WINDOWS__
 static RDebugReasonType r_debug_native_wait(RDebug *dbg, int pid) {
 	RDebugReasonType reason = R_DEBUG_REASON_UNKNOWN;
-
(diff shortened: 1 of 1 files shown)


XVilka:
Agree on separating the BSD debug into the independent file - will make it easier.

devnexen:
> 
> @devnexen Didn't really change anything but I would appreciate it if you could check that it's still ok in BSD. Also, we should start moving bsd to native/bsd/debug_bsd to make debug_native more easily maintainable.

In fact I just refactored the code here :-) and tested under FreeBSD. I already thought of making a BSD part but waiting there were a bit more features (lot less than Linux) but that should happen indeed.

yossizap:
>In fact I just refactored the code here 

Saw just now, I'll rebase but I still think we should separate this function into 3 at least. Great job though :)

>waiting there were a bit more features (lot less than Linux).

It's worth starting with whatever is working to clean debug_native step by step. It's really a mess out there unfortunately.

PR #15667, opened 2019-12-17
Fix some Race Conditions

thestr4ng3r:
As reported by helgrind.
RThreadLock.refs was broken beyond and only used in one place, so I removed it.

### DIFF
--- a/libr/core/task.c
+++ b/libr/core/task.c
@@ -290,7 +290,6 @@ R_API void r_core_task_schedule(RCoreTask *current, RTaskState next_state) {
 
 	if (stop) {
 		scheduler->tasks_running--;
-		r_th_lock_leave (current->dispatch_lock);
 	}
 
 	// oneshots always have priority.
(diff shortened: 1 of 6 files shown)


XVilka:
I suggest we wait after the release. Too risky to merge before, IMHO

thestr4ng3r:
Here refs is decremented after unlock, which means the mutex may not be held anymore:
https://github.com/radareorg/radare2/blob/6cfdafdf5113495ff2c8a7e938a5303c5f07b540/libr/util/thread_lock.c#L59-L69
This can race with any `r_th_lock_enter()` call from another thread.

It can be fixed by touching refs only while the mutex is guaranteed to be held. However in my opinion, these constructs in r_th should only be thin wrappers around the os-specific primitives and not add anything else. But I am open for discussion.

radare:
yeah, debugging refcounters is a pain. we can probably use those as helpers to debug deadlocks instead of blocking them to work as we expect, so totally agree with you

PR #15675, opened 2019-12-18
Debug, BSD systems refactoring ##r_debug

devnexen:
Separate pure BSD calls into its separate compilation unit.

### DIFF
--- a/libr/debug/Makefile
+++ b/libr/debug/Makefile
@@ -43,6 +43,12 @@ OBJS+= p/${COREDUMP_OBJS}
 endif
 endif
 
+ifneq (,$(findstring bsd,$(OSTYPE))$(findstring dragonfly,$(OSTYPE)))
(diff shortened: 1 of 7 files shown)


libr/debug/Makefile
```diff
@@ -43,6 +43,12 @@ OBJS+= p/${COREDUMP_OBJS}
 endif
 endif
 
+ifneq (,$(findstring bsd,$(OSTYPE))$(findstring dragonfly,$(OSTYPE)))
+ifneq (${NATIVE_OBJS},)
+OBJS+= p/${NATIVE_OBJS}
```
radare: uhm i dont think this will work if native objs is more than one word. use $(addprefix instead)

XVilka:
Will be merged after 4.1.1 release: https://github.com/radareorg/radare2/milestone/53

PR #15678, opened 2019-12-19
Fix #15331 - rabin2 -rk code injection issue

### LINKED ISSUE #15331: r2 passes path containing an @ as an address to rabin2
### Work environment
| OS/arch/bits (mandatory)                             | Ubuntu 18.04.3 x64
| File format of the file you reverse (mandatory)      | Binary file (passed a PE file with -nn as option)
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.0.0-git 23088 @ linux-x86-64 git.3.9.0-245-g85b291e53
commit: 85b291e5352666bc424951ac5b444c9fad183783 build: 2019-10-23__08:30:38

### Expected behavior
File loads without an error.

### Actual behavior
An error appears that rabin2 got an invalid address, but analysis is possible nevertheless.
```
r2 -nn "/home/firstname.lastname@domain.com/test.bin"
Invalid address (domain.com/test.bin")
|ERROR| Invalid command '.!rabin2 -rk. "/home/firstname.lastname@domain.com/test.bin"' (0x2e)
[0x00000000]>
```
No special sample needed, works with any binary. In this case with a file called "test.bin":

```
mkdir /tmp/test@test
mv test.bin /tmp/test@test 
r2 -nn /tmp/test@test/test.bin 
Invalid address (test/test.bin")
|ERROR| Invalid command '.!rabin2 -rk. "/tmp/test@test/test.bin"' (0x2e)
 -- That's embarrassing.
[0x00000000]> 
```

radare:


libr/core/cbin.c
```diff
@@ -154,6 +154,28 @@ R_API int r_core_bin_set_by_fd(RCore *core, ut64 bin_fd) {
 	return false;
 }
 
+// XXX use API, not cmdf()
+R_API bool r_core_bin_load_structs(RCore *core, const char *file) {
```
radare: probably bad api namings

PR #15679, opened 2019-12-19
Remove RAnalBlock members only used in Java

thestr4ng3r:
I have no idea what these were for, let's see what the tests say...


libr/anal/anal_ex.c
```diff
@@ -265,14 +221,7 @@ R_API RList* r_anal_ex_analysis_driver(RAnal *anal, RAnalState *state, ut64 addr
 		if (!r_anal_ex_get_bb (anal, state, state->current_addr)) {
 			break;
 		}
-		if (!state->current_bb_head) {
-			state->current_bb_head = state->current_bb;
-			if (state->current_bb_head) {
-				state->current_bb_head->type |= R_ANAL_BB_TYPE_HEAD;
```
thestr4ng3r: This seems to be the only place where these members actually affect anything...

XVilka:
We don't have much Java tests actually

radare:
I would prefer to merge the anal blocks pr before tweaking that struct more. Didnt checked this pr yet but there r tons of things to cleanup here. Also theres stuff used for diffing here that we should takeninto considerarion vecause i dont think we have much tests on this either

thestr4ng3r:
Actually if you read the diff precisely, you can see that this is actually all dead code (Unless I missed anything). All these members are tracked and updated but never used for anything.

Actually, this should be merged BEFORE anal-blocks because that one more or less depends on this. This head, body, tail stuff assigned on basic block level will simply not be applicable anymore when basic blocks can be contained within multiple functions.

radare:
The diff lgtm and i guess we can mill the bb type concept for anal blocks actually

thestr4ng3r:
Exactly. These types are implied automatically after the refactor.

PR #15500, opened 2019-11-25
asm.cmt.off: Accept 0 & 1 as aliases for false & true respectively

kazarmy:

PR #15681, opened 2019-12-19
When setting non-bool config var, set false/true if r_str_is_false()/_is_true()

kazarmy:
For https://github.com/radareorg/radare2/pull/15500#issuecomment-558248872.

### DIFF
--- a/libr/config/config.c
+++ b/libr/config/config.c
@@ -428,7 +428,7 @@ R_API RConfigNode* r_config_set(RConfig *cfg, const char *name, const char *valu
 					node->value = strdup (value);
 					node->i_value = 2;
 				}
-				if (IS_DIGIT (*value)) {
+				if (IS_DIGIT (*value) || (value[0] == '-' && IS_DIGIT (value[1]))) {
 					if (strchr (value, '/')) {
 						node->i_value = r_num_get (cfg->num, value);
 					} else {
--- a/libr/config/config.c
+++ b/libr/config/config.c
@@ -417,17 +417,23 @@ R_API RConfigNode* r_config_set(RConfig *cfg, const char *name, const char *valu
 				if (node->value == value) {
 					goto beach;
 				}
-				char *tmp = node->value;
-				node->value = strdup (value);
-				free (tmp);
+				free (node->value);
+				if (r_str_is_false (value)) {
+					node->value = strdup ("false");
+					node->i_value = 0;
+				} else if (r_str_is_true (value)) {
+					node->value = strdup ("true");
+					node->i_value = 1;
+				} else {
+					node->value = strdup (value);
+					node->i_value = 2;
+				}
 				if (IS_DIGIT (*value)) {
 					if (strchr (value, '/')) {
 						node->i_value = r_num_get (cfg->num, value);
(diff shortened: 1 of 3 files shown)


XVilka:
Can you please rebase? Test errors look unrelated:
```
[XX] db/cmd/cmd_open oob consider laddr 32bit  150
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -a x86 -b 32 -i /tmp/tmp-16974QZG0xr3aPLCH ../bins/mach0/fatmach0-3true
e bin.laddr = 0x5000
e bin.baddr = -1
oob
iI~laddr[1]
iI~baddr[1]
 0x0
-0x1000
+0x2
EXPECT=<<RUN
0x0
0x2
[XX] db/cmd/cmd_open oob consider laddr 64bit  213
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -a x86 -b 64 -i /tmp/tmp-16974Uktel4UV85m8 ../bins/mach0/fatmach0-3true
e bin.laddr = 0x5000
e bin.baddr = -1
oob
iI~laddr[1]
iI~baddr[1]
 0x0
-0x100000000
+0x2
EXPECT=<<RUN
0x0
0x2
[XX] db/cmd/cmd_print print raw lines  109
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -i /tmp/tmp-16974UHsLrFmpTqBQ ../bins/other/projects/resetc64
sl 5
prl 1~[0]
e lines.abs=true
prl 1~[0]
 +1
-5
+256
EXPECT=<<EOF
+1
256
```

kazarmy:
Actually they are related

kazarmy:
Currently idk why the cmd_print test is suddenly related

ret2libc:
This seems wrong to me. If i have a var that is an integer, why do I want to see `true`/`false` instead of 1/0?

For example, see `log.level` (but also many others like `dbg.bpsize` and all the others SETICB). If I now do `e log.level=1;e log.level` I see `true`, which does not make any sense in the context of `log.level` (same applies for `dbg.bpsize`, which sould just contain the size of a breakpoint, so "true" and "false" do not make sense there).

I believe this PR should be reverted.

ret2libc:
@radare @kazarmy ^

kazarmy:
The default getter for integer vars probably needs to be changed to something like what `scr.color` has.

ret2libc:
Why?? If you do that for all integer vars, then your change in the first place is not useful, as `1` would be converted in `true` and then, when printed back, converted again into `1`. Why? If the env var is an integer one, I don't see why we should automatically converting into true/false at all.

kazarmy:
Yes but the original change was not for integer vars, it was for non-integer string vars that need to support multiple-valued logic.

ret2libc:
> Yes but the original change was not for integer vars, it was for non-integer string vars that need to support multiple-valued logic.

Ok, so I think this automatic conversion should only be done for non-integer string vars and not for all non-bool vars, right?

kazarmy:
Yes that could work too.

PR #15690, opened 2019-12-20
Cursor mode: JK works as it should, g for moving to the first line, and some improvements

Vane11ope:


Vane11ope:
cool let me merge this

PR #15693, opened 2019-12-20
Fix #15682 - Fix getting noreturn info from invalid addr

### LINKED ISSUE #15682: Several instructions not being displayed in sym.__libc_csu_init
I'm working on a ROP Emporium challenge (https://ropemporium.com/challenge/ret2csu.html) and when I disassembled the __libc_csu_init function, it skips a few instructions that can be seen with GDB or objdump.

### Work environment

| OS/arch/bits (mandatory)                             | Arch Linux x64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.0.0 0 @ linux-x86-64 git.4.0.0 commit: unknown build: 2019-10-31__10:47:02

### Expected behavior
Radare2 shows the full disassembly of a function with the "pdf" command. 

### Actual behavior
Radare2 skips a few instructions, as can be seen by the two dots in the margin (see screenshot below).

- Download the binary from here: https://ropemporium.com/binary/ret2csu.zip
- Disassemble the __libc_csu_init function as shown:
image
- Same thing with "pdr":
image
- Disassembling the same function in GDB:
image

GustavoLCR:


libr/anal/fcn.c
```diff
@@ -1237,7 +1237,7 @@ static int fcn_recurse(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 len, int
 			// XXX: this is TYPE_MCALL or indirect-call
 			(void) r_anal_xrefs_set (anal, op.addr, op.ptr, R_ANAL_REF_TYPE_CALL);
 
-			if (op.ptr != UT64_MAX && r_anal_noreturn_at (anal, op.ptr)) {
+			if (op.ptr && op.ptr != UT64_MAX && r_anal_noreturn_at (anal, op.ptr)) {
```
radare: Can we have this check inside the noreturnat?

PR #15694, opened 2019-12-20
fix some crash of widget

Vane11ope:
reproduce:

V!
g
tab
select any candidate from widget
crash

this fixes it


Vane11ope:
merging

PR #15695, opened 2019-12-21
Debug: bsd_thread_list FreeBSD implementation proposal ##r_debug

devnexen:


libr/debug/p/native/bsd/bsd_debug.c
```diff
@@ -498,3 +498,39 @@ RList *bsd_desc_list(int pid) {
...
+	for (i = 0; i < max; i ++) {
+		RDebugPid *pid_info;
+		pid_info = r_debug_pid_new (NULL, kp[i].ki_tid, kp[i].ki_uid, 's', 0);
+		r_list_append (list, pid_info);
+	}
+
+	free (kp);
+	return list;
+#else
+	r_list_free (list);
```
yossizap: Maybe add a warning saying "bsd_thread_list: Unsupported architecture" for anything that isn't KFBSD?

libr/debug/p/native/bsd/bsd_debug.c
```diff
@@ -498,3 +498,39 @@ RList *bsd_desc_list(int pid) {
...
+	if (sysctl (mib, 4, kp, &len, NULL, 0) == -1) {
+		free (kp);
+		r_list_free (list);
+		return NULL;
+	}
+
+	max = len / sizeof(*kp);
+	for (i = 0; i < max; i ++) {
+		RDebugPid *pid_info;
+		pid_info = r_debug_pid_new (NULL, kp[i].ki_tid, kp[i].ki_uid, 's', 0);
```
yossizap: According to [kernel sources](https://github.com/freebsd/freebsd/blob/7e4664b0c30ad2ab36b196af2e02847cec202700/sys/sys/user.h#L175), it seems possible to obtain path from ki->ocomm(should also be available in comm [based on the implementation](https://github.com/freebsd/freebsd/blob/7e4664b0c30ad2ab36b196af2e02847cec202700/sys/kern/kern_proc.c#L738)) and status from ki->wchan instead of setting it to a default 's'. Would be nice to have those in `dpt` to stay consistent with the other debug implementations.

devnexen: true but ocomm is the not the full path.

yossizap: Better than an empty space, it will show the process' name for clarity.

devnexen: fair enough

libr/debug/p/native/bsd/bsd_debug.c
devnexen: unsupported platform most likely I might say but yes.

yossizap: Yeah wrong phrasing on my part

libr/debug/p/native/bsd/bsd_debug.c
```diff
@@ -498,3 +498,41 @@ RList *bsd_desc_list(int pid) {
...
+		free (kp);
+		r_list_free (list);
+		return NULL;
+	}
+
+	max = len / sizeof(*kp);
+	for (i = 0; i < max; i ++) {
+		RDebugPid *pid_info;
+		pid_info = r_debug_pid_new (kp[i].ki_comm, kp[i].ki_tid,
+			kp[i].ki_uid, kp[i].ki_wchan ? 's' : '*', 0);
```
yossizap: I was thinking of something like linux's status converter from /proc/%pid/status to R2 statuses - https://github.com/radareorg/radare2/blob/master/libr/debug/p/native/linux/linux_debug.c#L636

Would the lack of ki_wchan imply that this is the currently selected thread (Indicated by '*')? That info should be set by comparing dbg->tid to the currently set tid.

devnexen: ki_wchan is just the sleep pointer address. I felt it was enough for starter...

yossizap: All threads will usually return wchan since they are suspended when dpt is ran though, unless it's done in background which will the ones that are running as `*` currently selected.

On further examination of [kernel sources](https://github.com/freebsd/freebsd/blob/506c867c6e0e0ef3477c28e18b97f50207c7b230/sys/sys/proc.h#L255) it seems that the thread's wchan is filled with the sleep address which could be appended to the path with ':'(planned for the other debuggers) and ki_stat has the relevant status that you can convert. There's already an implementation that converts it in procstat's [source code]( https://github.com/freebsd/freebsd/blob/master/usr.bin/procstat/procstat_threads.c#L87) that you can use

I would at least fix the * marking to only mark the current thread and create an issue for appropriate statuses if you feel that it's out of the scope of this issue.

yossizap: Changes look great! Thanks for also including ki_comm in the path as well, we finally have a plugin with proper paths :)

Just add `kp[i].ki_tid == dbg.tid ? '*'` to the status field to mark selected as other debug plugins do.

yossizap: Actually correction - your previous implementation was ok, * will be marked in debug/pid.c. You can revert, sorry

radare:
can you fix the build for openbsd? and rebase

PR #15696, opened 2019-12-22
Add sys/r2env.sh: Set up env for r2 in non-standard loc

kazarmy:


radare:
learned a new posix thing today thanks :P

Issue #1961, opened 2015-01-14
Default fortunes may be NSFW/NSFC


voltagex:
We've run across a problem with radare2 where the default fortunes may violate our code of conduct at LCA2015* (e.g. displaying r2 at the front of a room in a tutorial)

Would it be possible to split the fortunes into safe/unsafe or default them to off?

*https://linux.conf.au/cor/code_of_conduct

radare:
Sure, there have been some discussion about it, and i'm ok with it, having an option to enable offensive fortunes can be cool instead of removing them.

radare:
Just to be clear, me, or the project doesnt support such offensive or exclusive fortunes, they are just to make fun and joy, nobody should take them seriously. But it's clear that we have 3 kind of fortunes that maybe we should split in different files:

* usage/tips
* fun/word games
* offensive

voltagex:
Fully agree, thanks for the response!

crowell:
@voltagex please take a look at #1966 and let me know if you disagree with any of the locations of the fortunes.

Keep in mind the new default will be `fun` and `tips`

radare:
I'm closing this bug after reviewing your commit which splits the fortunes by categories. It looks safe for all publics now :)

PR #15697, opened 2019-12-22
Fix debug and 'o' not being able to open paths with spaces #debug

yossizap:
ref #1961

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -1413,6 +1413,8 @@ static int cmd_open(void *data, const char *input) {
 			return 0;
 		}
 		if (argv) {
+			// Unescape spaces from the path
+			r_str_replace (argv[0], "\\ ", " ", true);
 			if (argc == 2) {
 				if (r_num_is_valid_input (core->num, argv[1])) {
 					addr = r_num_math (core->num, argv[1]);
@@ -1427,8 +1429,6 @@ static int cmd_open(void *data, const char *input) {
 		}
 		{
 			const char *argv0 = argv ? argv[0] : ptr;
-			// Unescape spaces from the path
-			r_str_replace ((char *)argv0, "\\ ", " ", true);
 			if ((file = r_core_file_open (core, argv0, perms, addr))) {
 				fd = file->fd;
 				core->num->value = fd;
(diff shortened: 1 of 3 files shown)


libr/core/cmd_open.c
```diff
@@ -1427,6 +1427,8 @@ static int cmd_open(void *data, const char *input) {
 		}
 		{
 			const char *argv0 = argv ? argv[0] : ptr;
+			// Unescape spaces from the path
+			r_str_replace ((char *)argv0, "\\ ", " ", true);
```
radare: this (char*) cast is terribly scary

yossizap: Unfortunately r_str_replace works on the given string, ptr is const and r_core_file_open expects const. I could make a new function just for r_str_path_unescape but is it really scary at all?

libr/core/cmd_open.c
```diff
@@ -1413,6 +1413,8 @@ static int cmd_open(void *data, const char *input) {
 			return 0;
 		}
 		if (argv) {
+			// Unescape spaces from the path
+			r_str_replace (argv[0], "\\ ", " ", true);
```
radare: We should use escape and unescape apis not replace. Otherwise this replacing may be unsafe

radare: But at least it would be a bit better like this i think

radare:
.

libr/core/cmd_open.c
yossizap: My initial implementation used the escaping and unescaping apis but those broke filenames with valid escapes like the cmd_i test(https://travis-ci.com/radareorg/radare2/jobs/269637812).

The only alternative is to implement a new function for path escaping and unescaping but it would be the same as only replacing the space format. Would you prefer that?

XVilka: @yossizap yes, please do it as the inline function, with making a string copy.

libr/core/cmd_open.c
radare: It doesn’t looks safe. It may return in a different heap and cause an UAF

libr/core/cmd_open.c
yossizap: Wouldn't it be wiser to just use r_str_replace_in?

radare:
Yes

PR #15698, opened 2019-12-22
Change Semantics of r_rbtree_upper_bound() and simplify Iter Fcns

thestr4ng3r:
For `r_rbtree_upper_bound()`,  "the smallest node that is greater than data" would mean "**smallest** node that is **outside** of the set given by the upper bound". For me, "**greatest** node that is **inside** the set given by the upper bound" makes more sense and is also much more useful next to the lower bound function.
The function was used nowhere before and had no tests, so that semantics change should be fine.

I also adjusted `r_rbtree_upper_bound_backward()` to go along with this meaning and removed `lower_bound_backward` and `upper_bound_forward` because they are complicated and I can't imagine a single use case for them right now. These functions also are not used anywhere except tests.

### DIFF
--- a/libr/util/rbtree.c
+++ b/libr/util/rbtree.c
@@ -36,23 +36,28 @@ static inline RBNode *zig_zag(RBNode *x, int dir, RBNodeSum sum) {
 	return z;
 }
 
-static inline RBIter bound_iter(RBNode *x, void *data, RBComparator cmp, bool upper, bool backward, void *user) {
+static inline RBIter bound_iter(RBNode *x, void *data, RBComparator cmp, bool upper, void *user) {
 	RBIter it;
 	it.len = 0;
 	while (x) {
 		int d = cmp (data, x, user);
-		if (upper ? d < 0 : d <= 0) {
-			if (!backward) {
+		if (d == 0) {
+			it.path[it.len++] = x;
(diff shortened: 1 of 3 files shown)


libr/util/rbtree.c
```diff
@@ -36,23 +36,28 @@ static inline RBNode *zig_zag(RBNode *x, int dir, RBNodeSum sum) {
...
 	it.len = 0;
 	while (x) {
 		int d = cmp (data, x, user);
-		if (upper ? d < 0 : d <= 0) {
-			if (!backward) {
+		if (d == 0) {
+			it.path[it.len++] = x;
+			return it;
+		}
+		if (d <= 0) {
```
ret2libc: just `<`

ret2libc: also, use a `else if`, it makes more sense here

thestr4ng3r: The first if is a bit of a special case and an early return so I prefer not to have an else here.

thestr4ng3r:
For anal-block, I need `r_rbtree_upper_bound()` in the sense of "greatest node that is less or equal to the given val". The rest is only making everything consistent with what "upper bound" means.

radare:
🎉

PR #15702, opened 2019-12-24
Improve realname support for symbols

ITAYC0HEN:
These changes aim to modify the way symbols are structured and look on the disassembly and across different places. More specifically, when `asm.flags.real` is set to true.

For example, this PR is aimed to change this:
```
call  dword   [sym.imp.user32.dll_CreateWindowExA]
```
To this:

```
call  dword  [CreateWindowExA]
```

In addition, it also adds the "libname" to each import. In general, it improves consistency across symbols in r2.


ITAYC0HEN:
### Current status
**Disassembly**:
Full flag looks like this: `imp.<libname>.<funcname>`  
Realflag looks like this: `<funcname>`

image

**Imports (iij)**
"name" reduced to the function name. "libname" member was added to describe the library. This will later be used to differentiate between imports of the same name from different libraries.

image

**Flags**
The flag shows the full symbol name with library and symbol type. The real name is set only to the function name
image

XVilka:
Could you please also add a few tests?

ITAYC0HEN:
What do you think of this direction? before we start the work on all plugins

libr/bin/p/bin_pe.inc
```diff
@@ -227,7 +228,8 @@ static RList* symbols(RBinFile *bf) {
 				break;
 			}
 			//strncpy (ptr->name, (char*)symbols[i].name, R_BIN_SIZEOF_STRINGS);
-			ptr->name = r_str_newf ("imp.%s", imports[i].name);
+			ptr->name = r_str_newf ("%s", imports[i].name);
```
pelijah: strdup?

ret2libc:
I’m very in favor of this change! I like it

ITAYC0HEN:
If the user will see `call     LoadLibraryA` we need to find a way to let users know how to navigate to this because `s LoadLibraryA` won't work. What do you think?

A suggested seek mechanism can be found below:

If only one flag contains this substring (e.g. only `imp.kernel32_dll.LoadLibraryA` will contain `LoadLibraryA`), then the seek will be to this address.

If there are more then one option, then r2 will show:

```
The flag 'LoadLibraryA' wasn't found. Did you mean to one of the following flags that contain this substring:
0x4001234    imp.kernel32_dll.LoadLibraryA
0x7e01133    imp.someOther_dll.LoadLibraryA
```

radare:
iij should contain flagname

radare:
pls rebase

radare:
+1 for this

radare:
some of the failing tests.. this doesnt looks good to me, but the idea to have an option to shorten those symbol names lgtm, about navigation, just use hotkeys

Screenshot 2020-01-07 at 23 16 45
Screenshot 2020-01-07 at 23 16 53
Screenshot 2020-01-07 at 23 17 06

libr/core/cbin.c
```diff
@@ -1704,12 +1709,12 @@ static int bin_relocs(RCore *r, int mode, int va) {
 }
 
 #define MYDB 1
-/* this is a hacky workaround that needs proper refactoring in Rbin to use Sdb */
+/* this is a VERY VERY VERY hacky and bad workaround that needs proper refactoring in Rbin to use Sdb */
```
ITAYC0HEN: lollll

thestr4ng3r:
There is a slight difference in flagnames with demangling currently:
master:
```
name (bin.demangle=false): sym.imp.MSVCP100.dll___Osfx___basic_ostream_DU__char_traits_D_std___std__QAEXXZ
name (bin.demangle=true): sym.public:_void___thiscall_std::basic_ostream_char__struct_std::char_traits_char__::_Osfx_void
```

this pr:
```
name (bin.demangle=false): sym.imp.MSVCP100.dll___Osfx___basic_ostream_DU__char_traits_D_std___std__QAEXXZ
name (bin.demangle=true): sym.imp.MSVCP100.dll_public:_void___thiscall_std::basic_ostream_char__struct_std::char_traits_char__::_Osfx_void
```
Note that the `imp.MSVCP100.dll_` part is missing on master. I think the new behavior makes much more sense since demangling should have no influence on what prefixes there are. This is important to understand regarding the test updates.

ITAYC0HEN:
> Note that the `imp.MSVCP100.dll_` part is missing on master. I think the new behavior makes much more sense since demangling should have no influence on what prefixes there are. This is important to understand regarding the test updates.


I completely agree :)

XVilka:
Please rebase on top of the currently green master, thanks!

PR #15705, opened 2019-12-24
Fix some warnings

radare:


libr/util/str.c
```diff
@@ -674,6 +674,7 @@ R_API const char *r_str_rstr(const char *base, const char *p) {
 }
 
 R_API const char *r_str_rchr(const char *base, const char *p, int ch) {
+	r_return_val_if_fail (base, NULL);
```
XVilka: You forgot to remove the original lines.

PR #15714, opened 2019-12-27
Remove dupped analysis in core

thestr4ng3r:
I don't see how this code should be useful for anything. `r_anal_fcn()` is always called before this and it is the function that does the actual analysis. Maybe @radare knows more about this.


radare:
It’s surprisingly not breaking any test. Where is this code duped from?

Is this Jmp mid thing working without this code? Looks like it does.. is ranalblock filled properly?

thestr4ng3r:
It is not directly code duplication. But basically `r_core_anal_bb()` seems to have the same purpose as `r_anal_fcn_bb()` (analyze a basic block and its successors recursively).

So I tried with `aaa` on a few different binaries and `r_core_anal_bb()` **always** just early-exists at this point without actually doing anything:
https://github.com/radareorg/radare2/blob/06d9632a3d72378361e9849b6d88cd4b3a563f8d/libr/core/canal.c#L1744-L1747
This makes sense because the address it has been given already has a basic block analyzed by `r_anal_fcn()`/`r_anal_fcn_bb()` called before.

So yeah, unless we find a special case where it does not early-exit here, everything will be filled correctly. But even if there is a case, it should be fixed in the "real" analysis loop instead of keeping two loops basically trying to do the same thing.

XVilka:
Good catch! I will vote for merging this.

PR #15716, opened 2019-12-28
eprintln instead of eprint

radare:


radare:
this macro is not used anywhere, but at least is not breaking the build of the V suite

radare:
What do u think about eliminating eprint() instead of renaming it to eprintln()? it is just used in base85.c cc @ret2libc @thestr4ng3r

thestr4ng3r:
Yeah, why not just replace the calls by eprintf?

radare:
yeah kiss

PR #15718, opened 2019-12-28
Force read permissions on all load maps in ELF ##bin

radare:


ITAYC0HEN:
are there maps in any file formats that should not have read perms? Otherwise, this can be added to all formats

radare:
yep, this is kind of a "hack", maybe we should have an option in RBinOptions to specify if we want to set 'r' on all maps having write or exec perms set.

ITAYC0HEN:
This could be useful. It will prevent people from get confusing why they don't see what they expect. Merging :)

GurTelem:
Spent 7 hours last week on this, thanks :-)

Issue #15691, opened 2019-12-20
Segfault in libcapstone.so.4 after renaming variable in visual mode


DoITCreative:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | ArchLinux x86_64
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86/32
| r2 -v full output, **not truncated** (mandatory)     | radare2 2.4.0-git 17284 @ darwin-x86-64 git.2.2.0-476-gf8cf84e06 commit: f8cf84e0653642d9ad34e760e0e56dd81860e799 build: 2018-02-17__11:08:27

### Expected behavior
Variable should be renamed
### Actual behavior
Segfault
### Steps to reproduce
```
aaa
s <address>
V
<space>
dv <last hex digits of variable>
<enter>
```
### Additional Logs
I've managed to trace the bug to the file:
```
./libr/core/vmenus.c:4060
```

Error happens in libcapstone.so.4 when trying to dereference pointer, that points to the address, that is stored in variable ```try_off``` on line 4060 in ```./libr/core/vmenus.c```

### Stacktrace
```
#0  0x00007f57405ededd in ?? () from /usr/lib/libcapstone.so.4
#1  0x00007f57405ec3da in ?? () from /usr/lib/libcapstone.so.4
#2  0x00007f57405ee109 in ?? () from /usr/lib/libcapstone.so.4
#3  0x00007f5740552a86 in cs_disasm () from /usr/lib/libcapstone.so.4
#4  0x00007f5740ba4eb7 in analop (a=0x55b4ca940930, op=0x7ffd83738690, addr=404799108, 
    buf=0x55b4e86abde0 <error: Cannot access memory at address 0x55b4e86abde0>, len=32, 
    mask=<optimized out>) at ../libr/anal/p/anal_x86_cs.c:3001
#5  0x00007f5740c60965 in r_anal_op (anal=0x55b4ca940930, op=op@entry=0x7ffd83738690, 
    addr=addr@entry=404799108, 
    data=0x55b4e86abde0 <error: Cannot access memory at address 0x55b4e86abde0>, 
    len=len@entry=32, mask=mask@entry=R_ANAL_OP_MASK_ALL) at ../libr/anal/op.c:154
#6  0x00007f5740f70ad1 in r_core_visual_define (core=<optimized out>, args=0x7f57410be12c "", 
[...]

PR #15721, opened 2019-12-28
Fix segfault on variable rename

DoITCreative:
Fix for issue #15691

### DIFF
--- a/libr/core/vmenus.c
+++ b/libr/core/vmenus.c
@@ -4055,7 +4055,8 @@ onemoretime:
 		if ((off & mask) <= N) {
 			start_off = (off & ~mask) ^ N;
 		} else {
-			start_off = ((off & ~mask) ^ incr) ^ N;
+			//TODO: Please test this
+			start_off = N & mask;
 		}
 
 		ut64 try_off;


radare:
I don’t think this change fixes the actual root cause of the segfault. The last bytes of the address are computed in the way its there.

The ranalop loop below is not checking for the return value so if it analyzes wrongly it keeps doing which may cause bad things. Can u try this? Also, if the value is starting with 0x maybe we can just use the whole value instead of part of it

radare:
i fixed the ranalop thing pls retry

DoITCreative:
Tried on:
```
radare2 4.2.0-git 23434 @ linux-x86-64 git.4.1.1-40-g265e0d1f6
commit: 265e0d1f6672f0023a3d63e45407cd4945d2d910 build: 2019-12-30__16:16:38
```
Still segfaults inside `r_anal_op(...)` on line 4065. This happens, because incorrect address is passed in `try_off` variable. I had specified last hex digits of the variable in `dv` command like so `dv 820be84`. So `try_off` should contain number 0x0820be84, but it has 0x1820be84 instead. That leads to the segfault on pointer dereference further. `try_off` depends on `start_off` variable, which is calculated incorrectly at line 4059.

Variables before calculation:
```
off = 0x0820be94
mask = 0x0fffffff
incr = 0x10000000
N = 0x0820be84
```

Correct formula looks like `start_off = (off & ~mask) ^ N;` but we go to the else branch of if case and calculate `start_off` like so: `start_off = ((off & ~mask) ^ incr) ^ N;` which is leading to incorrect result. Maybe if statement should check for `(off & ~mask) <= N` instead of `(off & mask) <= N`? 
It works in my case.
Anyways, the segfault check is not working. If incorrect value is passed in dv - command like so `dv 0x0820be94` then we still have segfault.
I found the way to catch segfaults though. With it and with if statement fix everything works correctly. I will make pr soon.

radare:
An invalid address shoukdnt result in a segfault at all; the issue must be somewhere else but i cant reproduce and i didnt get your asan crashlog yet

DoITCreative:
How to get asan crashlog?

radare:
Sys/asan.sh

DoITCreative:
There is no such script in sys/

DoITCreative:
I have created pr #15732 with working fix, I'm closing this pr.

PR #15722, opened 2019-12-28
RFC: Some minor cleanup in test/

radare:


test/Makefile
```diff
@@ -70,6 +59,10 @@ symstall:
 	ln -fs $(PWD)/r2-v $(BINDIR)/r2-v
 	ln -fs $(PWD)/r2r $(BINDIR)/r2r
 
+
+t:
```
radare: must be deleted

test/Makefile
```diff
@@ -23,41 +18,35 @@ fuzz/targets:
 
 R2R=$(shell cd new ; npm bin)/r2r
 
+RUNTEST=cd new && npm install ; $(R2R)
```
radare: should use 'v' instead

test/Makefile
radare: solved

test/Makefile
radare: almost there \o/

PR #15724, opened 2019-12-29
Fix is.j command output when doing r2 --

radare:


radare:
required for https://github.com/radareorg/radare2/pull/15722

radare:
iAj and others are problematic (see `v/r2r json`)

PR #15725, opened 2019-12-29
Refactoring ptrace register write for BSD ##r_debug

devnexen:


radare:
btw the hashtag is ##debug not ##r_debug

PR #15726, opened 2019-12-29
Fix var rebasing ##debug

yossizap:
After starting debug the register deltas weren't mapped to the same registers in remote and native debug. This made wrong registers appear as vars in native debug and no registers were shown in remote gdb since the remote's list is much shorter than r2 reg lists.

Rebasing before the fix:
```
┌ 101: int main (int argc, char **argv, char **envp);
│           ; arg int argc @ rdi
│           ; arg char **argv @ rsi

┌ 101: int main (int argc, char **argv, char **envp);
│           ; arg int argc @ cf
│           ; arg char **argv @ rip
```

Rebasing in remote gdb before the fix:
```
┌ 101: int main (int argc, char **argv, char **envp);
│           ; arg int argc @ rdi
│           ; arg char **argv @ rsi

┌ 101: int main (int argc, char **argv, char **envp);
│           ; 
│           ; 
```

Now pre and post `ood`/`oodf` output should be the same.

### DIFF
--- a/libr/anal/var.c
+++ b/libr/anal/var.c
@@ -12,9 +12,10 @@ struct VarType {
 	char *type;
 	int size;
 	char *name;
+	char *regname;
 };
 
-#define SDB_VARTYPE_FMT "bzdz"
+#define SDB_VARTYPE_FMT "bzdzz"
 
 #define EXISTS(x, ...) snprintf (key, sizeof (key) - 1, x, ## __VA_ARGS__), sdb_exists (DB, key)
 #define SETKEY(x, ...) snprintf (key, sizeof (key) - 1, x, ## __VA_ARGS__);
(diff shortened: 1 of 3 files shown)


libr/anal/var.c
```diff
@@ -359,6 +411,7 @@ R_API RAnalVar *r_anal_var_get(RAnal *a, ut64 addr, char kind, int scope, int de
 	av->name = vt.name? strdup (vt.name): strdup ("unkown_var");
 	av->size = vt.size;
 	av->type = vt.type? strdup (vt.type): strdup ("unkown_type");
+	av->regname = vt.regname? strdup (vt.regname): strdup ("unkown_regname");
```
radare: Put the ternary iperator inside the strdup call. Dont call strdup twice qhen its just one arg that changes

libr/anal/var.c
```diff
@@ -370,8 +423,15 @@ R_API RAnalVar *r_anal_var_get(RAnal *a, ut64 addr, char kind, int scope, int de
 
 R_API void r_anal_var_free(RAnalVar *av) {
 	if (av) {
-		free (av->name);
-		free (av->type);
+		if (av->name) {
+			free (av->name);
```
radare: Dont do that. Free(null) is totally legit

libr/anal/var.c
```diff
@@ -370,8 +423,15 @@ R_API RAnalVar *r_anal_var_get(RAnal *a, ut64 addr, char kind, int scope, int de
...
+		}
+		if (av->regname) {
+			free (av->regname);
+		}
+		if (av->type) {
+			free (av->type);
+		}
 		R_FREE (av);
```
radare: Just free() not need to use r_free because its a local var

PR #15728, opened 2019-12-29
More improvements for the cmd tests using r2r.v

radare:


XVilka:
There is an error with the old testsuite then:
```
[XX] Wrong asm test format in db/asm/wasm:d "i64.extend_u/i32"
[XX] db/asm/wasm wasm: "i64.const 32" => 4220 (assemble)  
$ r2 
-4220
+42
```

radare:
Its fixed in a pr

PR #15732, opened 2019-12-30
Fix for issue #15691

DoITCreative:
If statement fix and segfault catch implemented.

### DIFF
--- a/libr/core/vmenus.c
+++ b/libr/core/vmenus.c
@@ -2,6 +2,9 @@
 
 #include <r_core.h>
 #include <r_util.h>
+
+#include <setjmp.h>
+#include <signal.h>
 #include <string.h>
 
 #define MAX_FORMAT 3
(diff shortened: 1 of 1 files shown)


radare:
quite elaborated joke

libr/core/vmenus.c
```diff
@@ -4061,8 +4073,26 @@ R_API void r_core_visual_define(RCore *core, const char *args, int distance) {
...
+			sigemptyset(&sa.sa_mask);
+			sa.sa_flags = SA_NODEFER;
+			sa.sa_sigaction = segfaulty_action_handler;
+			sigaction(SIGSEGV, &sa, NULL);
+			
+			int ret;
+			if (setjmp(point) == 0)
+			{
+			    // Call segfaulty function
+			    ret = r_anal_op (core->anal, &op, try_off, core->block + try_off - core->offset, 32, R_ANAL_OP_MASK_ALL);
```
radare: the problem is when try_off - core->offset is < 0 or > core->blocksize, using r_core_anal_op() should be fine because we just need to provide the address and chk for null instead

DoITCreative:
Yeah, I do not like goto stuff, either, but it looks like the only way to catch segfaults. Because it happens inside the external library libcapstone.so

DoITCreative:
Could you send me asan.sh script, so I could make crashlogs

radare:
it was renamed to sanitize.sh

DoITCreative:
Ah, ok

radare:
buti think the bug is quite obvious, let me submit a ffix so u can try

DoITCreative:
Ok

DoITCreative:
[crashlog.txt](https://github.com/radareorg/radare2/files/4010221/crashlog.txt)
Log is a little bit messy because of visual mode

radare:
yeah pls try with master and confirm the bug is gone, i think this crashlog makes sense with my understanding of the issue now. thanks

PR #15737, opened 2019-12-31
Fix issues in iCj, drlj and arlj ##json

radare:


radare:
axgj needs to be fully rewritten to get this green, the reason why its failing in the node suite is because the v one is testing all those commands wihtout opening any file

PR #15738, opened 2019-12-31
Fix #15734 - Automatically download PDB file if pdb.autoload=1

### LINKED ISSUE #15734: Windows PDB server not available anymore. Any other option?
I am trying to debug Win32 program but the PDB server from Microsoft is not available anymore.
What other option do I have to get Windows symbols. The first PDB that radare2 is complaining is ntdll.pdb.

Be dependent on MS was always a headache.

GustavoLCR:
Fix #15734

### DIFF
--- a/libr/bin/pdb/pdb_downloader.c
+++ b/libr/bin/pdb/pdb_downloader.c
@@ -31,6 +31,7 @@ static int download(struct SPDBDownloader *pd) {
 	char *curl_cmd = NULL;
 	char *extractor_cmd = NULL;
 	char *abspath_to_archive = NULL;
+	char *abspath_to_file = NULL;
 	char *archive_name = NULL;
 	size_t archive_name_len = 0;
 	char *symbol_store_path = NULL;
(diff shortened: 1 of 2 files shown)


ITAYC0HEN:
our king of windows things ♥

XVilka:
@GustavoLCR could you please rebase on top of the latest green commit - https://github.com/radareorg/radare2/commit/0db3d9a0fc0ed20dcf496f77705ca839331ce068

So we can be sure your PR doesn't break anything (it shouldn't, but just in case) before merging.

GustavoLCR:
> @GustavoLCR could you please rebase on top of the latest green commit - [0db3d9a](https://github.com/radareorg/radare2/commit/0db3d9a0fc0ed20dcf496f77705ca839331ce068)
> 
> So we can be sure your PR doesn't break anything (it shouldn't, but just in case) before merging.

@XVilka Seems like CI always use latest master, travis is still red even after rebasing

XVilka:
@GustavoLCR thanks, that's really sad then. Anyway, checked the PR - it looks fine, lets merge it.

radare:
Reverting my r2r update should be enough to get it green again but i dont see why my change should produce those errors

XVilka:
I didn't revert because I wasn't sure. It looks like might be a capstone problem again.

PR #15742, opened 2020-01-01
Add RISC-V 64 word instructions and test

sylvainpelissier:


XVilka:
In the test, it's easier to write a hex string as a whole rather than splitting every byte.

sylvainpelissier:
Done.

PR #15743, opened 2020-01-01
Fix reading on invalid process memory ##windows

GustavoLCR:
This fixes situations that if a read contained even a byte inside invalid memory, the entire read would fail

### DIFF
--- a/libr/io/p/io_w32dbg.c
+++ b/libr/io/p/io_w32dbg.c
@@ -18,17 +18,65 @@
 #undef R_IO_NFDS
 #define R_IO_NFDS 2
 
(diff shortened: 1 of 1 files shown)


radare:
it is causing any slowdown' can you rebase after having master green? @sanguinawer can you test?

GustavoLCR:
> it is causing any slowdown' can you rebase after having master green? @sanguinawer can you test?

No slowdowns if read was successful in the first place. If it fails and enters the "recovery mode" it's technically slower but its by a negligible amount

radare:
Looks good to merge for me

PR #15744, opened 2020-01-01
Add location in file to linux dpt output ##debug

yossizap:
Shows the module, offset and function name instead of showing the executable's
path for all threads.

```
 * 33225 s /home/yossi/workspace/tests/main (0x5585485d2cc5) in main+0x95
 - 33226 s /lib/x86_64-linux-gnu/libc-2.27.so (0x7f64539da9d0)
 - 33227 s /lib/x86_64-linux-gnu/libc-2.27.so (0x7f64539da9d0)
```


radare:
This pr broke a test. Can you fix it? The archos one with the grep in dpt output

yossizap:
Will do, missed it because of the node test issue sorry

PR #15746, opened 2020-01-02
Apply types on matching zignature data ##signatures

### LINKED ISSUE #15372: Create and apply types from zignatures
For example if you have a zignature with [embedded type information](https://github.com/radareorg/radare2/issues/13661) when there is a match it should create a function with a proper arguments names, types and return type.

oxcabe:
This is still WIP - don't merge yet, there are unnecessary comments and prints.

PR status:
- [x] Serialize types fields into zignatures:
Example: `types: main.ret=int, main.args=2, main.arg.0="int,argc", main.arg.1="char **,argv"`

- [x] Deserialize zignature types field into key-value strings. Done in `RList *zign_types_to_list(RAnal *a, char *types)`. 

- [x] Apply types information on matching functions.

Edit: Closes #15372

### DIFF
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -67,6 +67,40 @@ R_API RList *r_sign_fcn_vars(RAnal *a, RAnalFunction *fcn) {
 	return ret;
 }
 
+R_API RList *r_sign_fcn_types(RAnal *a, RAnalFunction *fcn) {
+
+	// From anal/types/*:
+	// Get key-value types from sdb matching "func.%s", fcn->name
+	// Get func.%s.args (number of args)
+	// Get type,name pairs
+	// Put everything in RList following the next format:
+	// types: main.ret=%type%, main.args=%num%, main.arg.0="int,argc", ...
+
+	r_return_val_if_fail (a && fcn, NULL);
+
+	RList *ret = r_list_newf ((RListFree) free);
+	char *args_expr = r_str_newf ("func.%s.args", fcn->name), *arg = NULL;
+	const char *ret_type = sdb_const_get (a->sdb_types, r_str_newf ("func.%s.ret", fcn->name), 0);
+	const char *fcntypes = sdb_const_get (a->sdb_types, args_expr, 0);
+	int argc = 0;
+
+	if (fcntypes) {
+		if (ret_type) {
+			r_list_append (ret, r_str_newf ("func.%s.ret=%s", fcn->name, ret_type));
+		}
+		argc = atoi (fcntypes);
+		r_list_append (ret, r_str_newf ("func.%s.args=%d", fcn->name, argc));
+		for (int i = 0; i < argc; i++) {
+			arg = sdb_get (a->sdb_types, r_str_newf ("func.%s.arg.%d", fcn->name, i), 0);
+			r_list_append (ret, r_str_newf ("func.%s.arg.%d=\"%s\"", fcn->name, i, arg));
+		}
+	}
+
+	free (arg);
+	free (args_expr);
+	return ret;
+}
+
 R_API RList *r_sign_fcn_xrefs(RAnal *a, RAnalFunction *fcn) {
 	RListIter *iter = NULL;
 	RAnalRef *refi = NULL;
@@ -119,12 +153,35 @@ R_API RList *r_sign_fcn_refs(RAnal *a, RAnalFunction *fcn) {
 	return ret;
 }
 
+static RList *zign_types_to_list(RAnal *a, char *types) {
+	RList *ret = r_list_newf ((RListFree) free);
+	unsigned int i = 0, prev = 0, len = strlen (types);
+	bool quoted = false;
+	char *token = NULL;
+
+	for (i = 0; i <= len; i++) {
+		if (types[i] == '"') {
+			quoted = !quoted;
+		}
+		else if ((types[i] == ',' && !quoted) || types[i] == '\0') {
+			token = r_str_ndup (types + prev, i - prev);
+			if (token) {
+				prev = i + 1;
+				r_list_append (ret, strdup (token));
+			}
+		}
+	}
+
+	free (token);
+	return ret;
+}
+
 R_API bool r_sign_deserialize(RAnal *a, RSignItem *it, const char *k, const char *v) {
 	char *refs = NULL;
 	char *vars = NULL;
 	char *types = NULL;
 	const char *token = NULL;
-	int i = 0, n = 0, nrefs = 0, nvars = 0, ntypes = 0, size = 0, w = 0;
+	int i = 0, n = 0, nrefs = 0, nvars = 0, size = 0, w = 0;
 
 	r_return_val_if_fail (a && it && k && v, false);
 
@@ -225,13 +282,7 @@ R_API bool r_sign_deserialize(RAnal *a, RSignItem *it, const char *k, const char
 			break;
 		case R_SIGN_TYPES:
 			types = r_str_new (token);
-			ntypes = r_str_split (types, ',');
-			if (ntypes > 0) {
-				it->types = r_list_newf ((RListFree) free);
-				for (i = 0; i < ntypes; i++) {
-					r_list_append (it->types, r_str_newf (r_str_word_get0 (types, i)));
-				}
-			}
+			it->types = zign_types_to_list (a, types);
 			break;
 		case R_SIGN_BBHASH:
 			if (token[0] != 0) {
@@ -2081,7 +2132,7 @@ static int typesMatchCB(RSignItem *it, void *user) {
 		return 1;
 	}
 	// TODO(nibble | oxcabe): slow operation, add cache
-	types = r_anal_types_from_fcn (ctx->anal, ctx->fcn);
+	types = r_sign_fcn_types (ctx->anal, ctx->fcn);
 	if (!types) {
 		return 1;
 	}
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -833,6 +833,11 @@ static void __add_vars_sdb(RCore *core, RAnalFunction *fcn) {
 			arg_count++;
 		}
 	}
+	if (arg_count > 0) {
+		char *query = r_str_newf ("anal/types/func.%s.args=%d", fcn->name, arg_count);
+		sdb_querys (core->sdb, NULL, 0, query);
+		free (query);
+	}
 }
 
 static bool cmd_anal_aaft(RCore *core) {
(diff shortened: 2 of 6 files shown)


oxcabe:
I'm not really sure at which point of the function matching process should we integrate the return type and arguments. Name is also included into the types section, so we can also include that in the merging.

XVilka:
Yes, please also apply the names of the function and arguments along with types. Moreover, it would be nice to apply the comment at the address of the function if it was stored in the signature files.

libr/anal/sign.c
```diff
@@ -67,6 +67,44 @@ R_API RList *r_sign_fcn_vars(RAnal *a, RAnalFunction *fcn) {
...
+	RList *ret = r_list_newf ((RListFree) free);
+	SdbKv *kv = NULL;
+	SdbListIter *it = NULL;
+	char *args_expr = r_str_newf ("func.%s.args", fcn->name);
+	SdbList *fcntypes = sdb_foreach_match (a->sdb_types, args_expr, 0);
+
+	if (fcntypes) {
+		ls_foreach (fcntypes, it, kv) {
+			if (strcmp (sdbkv_key (kv), args_expr) == 0) {
+				char *ret_type = sdb_get (a->sdb_types, r_str_newf ("func.%s.ret", fcn->name), 0);
```
thestr4ng3r: The `r_str_newf ("func.%s.ret"` is leaked. Since that string is used multiple times, it makes sense to allocate it once outside the loop, just like `args_expr` and free it at the end of the function.

The returned value of `sdb_get()` is leaked too. Here you can just use `sdb_const_get()` instead of regular get to fix it.

libr/anal/sign.c
```diff
@@ -67,6 +67,44 @@ R_API RList *r_sign_fcn_vars(RAnal *a, RAnalFunction *fcn) {
...
+				if (ret_type) {
+					r_list_append (ret, r_str_newf ("func.%s.ret=%s", fcn->name, ret_type));
+				}
+				int argc = atoi (sdb_get (a->sdb_types, args_expr, 0));
+				r_list_append (ret, r_str_newf ("func.%s.args=%d", fcn->name, argc));
+				for (int i = 0; i < argc; i++) {
+					char *arg = sdb_get (a->sdb_types, r_str_newf ("func.%s.arg.%d", fcn->name, i), 0);
```
thestr4ng3r: 2 leaks here

libr/anal/sign.c
```diff
@@ -119,12 +157,33 @@ R_API RList *r_sign_fcn_refs(RAnal *a, RAnalFunction *fcn) {
...
 
+static RList *zign_types_to_list(RAnal *a, char *types) {
+	RList *ret = r_list_newf ((RListFree) free);
+	unsigned int i = 0, prev = 0, len = strlen (types);
+	bool quoted = false;
+	char token[1024];
+
+	for (i = 0; i <= len; i++) {
+		if (types[i] == '"') {
+			quoted ^= 1;
```
thestr4ng3r: better to use `quoted = !quoted`, it fits the bool type better.

libr/anal/sign.c
```diff
@@ -535,7 +589,8 @@ static bool addItem(RAnal *a, RSignItem *it) {
 		mergeItem (curit, it);
 		serialize (a, curit, key, val);
 	}
-	sdb_set (a->sdb_zigns, key, val, 0);
+	char *query = r_str_newf ("%s=%s", key, val);
+	sdb_querys (a->sdb_zigns, NULL, 0, query);
```
thestr4ng3r:
```suggestion
	sdb_querys (a->sdb_zigns, NULL, 0, query);
	free (query);
```

libr/core/cmd_anal.c
```diff
@@ -835,6 +835,9 @@ static void __add_vars_sdb(RCore *core, RAnalFunction *fcn) {
 			arg_count++;
 		}
 	}
+	if (arg_count > 0) {
+		sdb_querys (core->sdb, NULL, 0, r_str_newf ("anal/types/func.%s.args=%d", fcn->name, arg_count));
```
thestr4ng3r:
```suggestion
		char *query = r_str_newf ("anal/types/func.%s.args=%d", fcn->name, arg_count)
		sdb_querys (core->sdb, NULL, 0, query);
		free (query);
```

libr/anal/sign.c
```diff
@@ -67,6 +67,44 @@ R_API RList *r_sign_fcn_vars(RAnal *a, RAnalFunction *fcn) {
...
+	// Put everything in RList following the next format:
+	// types: main.ret=%type%, main.args=%num%, main.arg.0="int,argc", ...
+
+	r_return_val_if_fail (a && fcn, NULL);
+
+	RList *ret = r_list_newf ((RListFree) free);
+	SdbKv *kv = NULL;
+	SdbListIter *it = NULL;
+	char *args_expr = r_str_newf ("func.%s.args", fcn->name);
+	SdbList *fcntypes = sdb_foreach_match (a->sdb_types, args_expr, 0);
```
thestr4ng3r: Wouldn't a single `sdb_get_const()` work here instead of the match, iterating through the list and strcmp?

oxcabe: Yes, looking at my code now I see you are right. I don't even know why I did it this way to begin with, lol

libr/anal/sign.c
```diff
@@ -119,12 +153,33 @@ R_API RList *r_sign_fcn_refs(RAnal *a, RAnalFunction *fcn) {
...
+	unsigned int i = 0, prev = 0, len = strlen (types);
+	bool quoted = false;
+	char token[1024];
+
+	for (i = 0; i <= len; i++) {
+		if (types[i] == '"') {
+			quoted = !quoted;
+		}
+		else if ((types[i] == ',' && !quoted) || types[i] == '\0') {
+			if (snprintf (token, i - prev + 1, "%s", types + prev)) {
```
radare: not null terminated. use r_str_ndup()

libr/anal/sign.c
```diff
@@ -535,7 +585,9 @@ static bool addItem(RAnal *a, RSignItem *it) {
 		mergeItem (curit, it);
 		serialize (a, curit, key, val);
 	}
-	sdb_set (a->sdb_zigns, key, val, 0);
+	char *query = r_str_newf ("%s=%s", key, val);
```
radare: wat. why do a query, when you can just use the api?

oxcabe: At that point I thought I had a bug in the code that was caused by the API when setting values containing commas.

It ended up being something unrelated and I forgot to undo this change.

radare:
please rebase and squash, the build failure is fixed in master

oxcabe:
Done - I expect to have the last check done over this weekend

XVilka:
Would be nice to have this ready for 4.2.0 btw

radare:
Yep

libr/core/cmd_zign.c
```diff
@@ -731,9 +731,80 @@ struct ctxSearchCB {
...
+
+// TODO (oxcabe):
 static void addFlag(RCore *core, RSignItem *it, ut64 addr, int size, int count, const char* prefix, bool rad) {
+	RAnalFunction *fcn = NULL;
 	const char *zign_prefix = r_config_get (core->config, "zign.prefix");
 	char *name = r_str_newf ("%s.%s.%s_%d", zign_prefix, prefix, it->name, count);
+	char *fcnstr = NULL;
+
+	if (it->types) {
+		fcnstr = types_list_to_fcnstr (it->types);
```
radare: define and assign the variable in here. and this is const char *fcnstr

libr/core/cmd_zign.c
```diff
@@ -731,9 +731,80 @@ struct ctxSearchCB {
...
+		fcn = r_anal_get_fcn_at (core->anal, it->addr, 0);
+		r_anal_str_to_fcn (core->anal, fcn, fcnstr);
+	}
 	if (!name) {
```
radare: construct name in here instead of donig it before it->types block. you dont need it before and you can just wrap it with a positive condition like if (name) { ..} so you can avoid the return

radare: in line 812 below. the `r_cons_printf ("f %s %d @ 0x%08"PFMT64x"\n", name, size, addr);` can execute ugly code because name is not filtered and can contain special chars.

libr/core/cmd_zign.c
```diff
@@ -731,9 +731,80 @@ struct ctxSearchCB {
 	const char *prefix;
 };
 
+static char *types_list_to_fcnstr(RList *types) {
+	char *type_kv = NULL, *k = NULL, *v = NULL;
+	char *field = NULL, *name = NULL, *rettype = NULL;
+	char **args, *ret = NULL;
```
radare: any reason not to use RList instead of char ** here? mmemory management and code readability will be better

libr/core/cmd_zign.c
```diff
@@ -731,9 +731,80 @@ struct ctxSearchCB {
...
+	int nargs = 0, i = 0, j = 0;
+	RListIter *iter;
+
+	r_list_foreach(types, iter, type_kv) {
```
radare: missing space before (

libr/core/cmd_zign.c
```diff
@@ -731,9 +731,80 @@ struct ctxSearchCB {
...
+		k = strtok (type_kv, "=");
+		v = strtok (NULL, "\0");
+
+		strtok (k, ".");
+		name = strtok (NULL, ".");
+		field = strtok (NULL, ".");
+
+		while (strcmp (field, "ret") && strcmp (field, "args") &&
+		  strcmp (field, "arg") && strcmp (field, r_str_newf ("%d", i))) {
```
radare: * dont indent with spaces.
* this condition is quite complex and will be better to move into a separate function like `while (__validField())`
* the r_str_newf is leaking

libr/core/cmd_zign.c
```diff
@@ -731,9 +731,80 @@ struct ctxSearchCB {
...
+				}
+			}
+			i++;
+		}
+
+		if (!rettype) {
+			rettype = strdup ("");
+		}
+
+		ret = r_str_newf ("%s %s(", rettype, name);
```
radare: ret is redefined on each iteration of the loop. so its leaking too

oxcabe:
I haven't fixed all of the review notes yet, but decided to commit since matching actually seems to work now.

I also have to refactor some new things I introduced from the "afs" command that perhaps belong in a R_API function.

radare:
Make travis green and i’ll merge

You can continue in a separate pr after merging

radare:
Lgtm!!

radare:
thanks!!!1

Issue #1975, opened 2015-01-17
Replace ${flagname} by its value in ?e


radare:
cmd_help.c 294

this way we can make echos much simpler:
```
now:
> ?e Size: `?v size`

future:
> ?e Size ${size}
or
> ?e Size $size
```

jvoisin:
I prefer $size, since

radare:
Both syntaxes should be valid imho

PR #15750, opened 2020-01-02
Validate bp addr on rebase and restore instead of preventing creation

yossizap:
This way it is possible to set breakpoints before starting debug through 'db' and the user will be notified when a breakpoint points to an invalid map.

ref [#1975](https://github.com/radareorg/cutter/pull/1975)

### DIFF
--- a/libr/bp/bp.c
+++ b/libr/bp/bp.c
@@ -1,6 +1,8 @@
 /* radare2 - LGPL - Copyright 2009-2018 - pancake */
 
 #include <r_bp.h>
+#include <r_core.h>
+#include <r_debug.h>
 #include <config.h>
 
 R_LIB_VERSION (r_bp);
@@ -203,6 +205,13 @@ static RBreakpointItem *r_bp_add(RBreakpoint *bp, const ut8 *obytes, ut64 addr, 
 		}
 		b->recoil = ret;
 	}
+	// Check if the breakpoint is set in a valid location
+	if (r_bp_is_valid (bp, b)) {
+		b->valid = true;
+	} else {
+		b->valid = false;
+	}
+
 	bp->nbps++;
 	r_list_append (bp->bps, b);
 	return b;
@@ -288,7 +297,7 @@ R_API int r_bp_list(RBreakpoint *bp, int rad) {
 		switch (rad) {
 		case 0:
 			bp->cb_printf ("0x%08"PFMT64x" - 0x%08"PFMT64x \
-				" %d %c%c%c %s %s %s cmd=\"%s\" cond=\"%s\" " \
+				" %d %c%c%c %s %s %s %s cmd=\"%s\" cond=\"%s\" " \
 				"name=\"%s\" module=\"%s\"\n",
 				b->addr, b->addr + b->size, b->size,
 				((b->perm & R_BP_PROT_READ) | (b->perm & R_BP_PROT_ACCESS)) ? 'r' : '-',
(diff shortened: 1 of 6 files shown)


libr/bp/bp.c
```diff
@@ -1,6 +1,8 @@
 /* radare2 - LGPL - Copyright 2009-2018 - pancake */
 
 #include <r_bp.h>
+#include <r_core.h>
```
radare: this is wrong you cant use rcore or rdebug from rbp

libr/bp/bp.c
```diff
@@ -409,3 +420,22 @@ R_API int r_bp_size(RBreakpoint *bp) {
 	}
 	return bpsize;
 }
+
+// Check if the breakpoint is inside a valid map
+R_API bool r_bp_is_valid(RBreakpoint *bp, RBreakpointItem *b) {
+	RDebugMap *map;
+	RListIter *iter;
+	RCore *core = (RCore*)bp->corebind.core;
+	if (!r_config_get_i (core->config, "dbg.bpinmaps")) {
```
radare: add bp->inmaps instead of using rconfig here

libr/bp/bp.c
```diff
@@ -409,3 +420,22 @@ R_API int r_bp_size(RBreakpoint *bp) {
...
+		return true;
+	}
+
+	r_debug_map_sync (core->dbg);
```
radare: use corebind api like bp->corebind.mapSync() but this is not implemented . see libr/include/r_bind.h and r_core_bind function

radare: but.. is this call really needed? cant we just call it from rdebug? also, this is a very expensive call to be called in isvalid()

libr/bp/bp.c
```diff
@@ -409,3 +420,22 @@ R_API int r_bp_size(RBreakpoint *bp) {
...
+	r_list_foreach (core->dbg->maps, iter, map) {
```
radare: we should use rbtree cc @thestr4ng3r

libr/include/r_bp.h
```diff
@@ -48,6 +48,7 @@ typedef struct r_bp_item_t {
 	int size; /* size of breakpoint area */
 	int recoil; /* recoil */
 	bool swstep; 	/* is this breakpoint from a swstep? */
+	bool valid; /* Whether the breakpoint in a valid map */
```
radare: this is kind of cache, because the real value can change at any time, wouldnt be better to have an api to determine if its in a valid map or not?

libr/bp/bp.c
radare: or maybe corebind.isMapped(addr)

radare:
i have merged and reverted your pr.

libr/bp/bp.c
yossizap: Added getDebugMaps since isMapped doesn't check file permissions and also calls r_debug_map_sync for each addr.

> but.. is this call really needed? cant we just call it from rdebug?

Limited it to once per restore (if bpinmaps is enabled).

libr/bp/bp.c
yossizap: Elaborate?

libr/include/r_bp.h
yossizap: Removed it and replaced with calls to r_bp_is_valid whenever it needs to be verified.

libr/bp/bp.c
yossizap: This was slightly problematic but I added new methods to bypass the issue. Is adding RList to corebind ok? Prefer not to return void *

yossizap: Also, couldn't remove r_debug.h because I need to be aware of RDebugMap, should I add a new method to corebind that checks if the addr is in a map and if that map is valid to bypass this? Is it even an issue if I am not using any methods from debug?

PR #15754, opened 2020-01-03
Refactor breakpoint validation

yossizap:
Based on the comments at #15750, history doesn't show up there anymore.

### DIFF
--- a/libr/bp/bp.c
+++ b/libr/bp/bp.c
@@ -1,6 +1,7 @@
 /* radare2 - LGPL - Copyright 2009-2018 - pancake */
 
 #include <r_bp.h>
+#include <r_debug.h>
 #include <config.h>
 
 R_LIB_VERSION (r_bp);
@@ -288,7 +289,7 @@ R_API int r_bp_list(RBreakpoint *bp, int rad) {
 		switch (rad) {
 		case 0:
 			bp->cb_printf ("0x%08"PFMT64x" - 0x%08"PFMT64x \
-				" %d %c%c%c %s %s %s cmd=\"%s\" cond=\"%s\" " \
+				" %d %c%c%c %s %s %s %s cmd=\"%s\" cond=\"%s\" " \
 				"name=\"%s\" module=\"%s\"\n",
 				b->addr, b->addr + b->size, b->size,
 				((b->perm & R_BP_PROT_READ) | (b->perm & R_BP_PROT_ACCESS)) ? 'r' : '-',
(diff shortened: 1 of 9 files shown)


libr/bp/bp.c
```diff
@@ -1,6 +1,7 @@
 /* radare2 - LGPL - Copyright 2009-2018 - pancake */
 
 #include <r_bp.h>
+#include <r_debug.h>
```
radare: Nope

libr/bp/bp.c
```diff
@@ -409,3 +412,26 @@ R_API int r_bp_size(RBreakpoint *bp) {
...
+// Check if the breakpoint is in a valid map
+R_API bool r_bp_is_valid(RBreakpoint *bp, RBreakpointItem *b) {
+	RDebugMap *map;
+	RListIter *iter;
+	RList *maps;
+	if (!bp->bpinmaps) {
+		return true;
+	}
+
+	if (!(maps = bp->corebind.getDebugMaps (bp->corebind.core))) {
```
radare: Unnecessary if you see the code in is,apped younwill see that it already does all the debugger iomaps? Debug,aps checks

libr/bp/bp.c
yossizap: So should I add isValidMap in corebind that checks address + permissions?

libr/core/core.c
```diff
@@ -282,6 +282,20 @@ static bool __isMapped(RCore *core, ut64 addr) {
 	return r_io_map_is_mapped (core->io, addr);
 }
 
+static RList *__getDebugMaps(RCore *core) {
+	if (r_config_get_i (core->config, "cfg.debug")) {
+		return core->dbg->maps;
+	}
+	return NULL;
+}
+
```
radare: Unnecessary callbacks

libr/include/r_bind.h
```diff
@@ -39,6 +43,8 @@ typedef struct r_core_bind_t {
 	RCoreConfigGet cfgGet;
 	RCoreNumGet numGet;
 	RCoreIsMapped isMapped;
+	RCoreDebugMapsGet getDebugMaps;
```
radare: Both callbacks are unnecessary

libr/bp/bp.c
yossizap: Autocorrect butchered your sentence but I assume you meant to say that isMapped does the job?

In that case, it doesn't since it only checks if the address is in a map. It doesn't check the permissions as I wrote in a comment a few lines before. Plus, isMapped runs r_debug_maps_sync for each address which isn't desirable as we agreed earlier.

libr/include/r_bind.h
yossizap: syncDebugMaps is needed to sync only once instead of syncing in each isMapped call.

PR #15757, opened 2020-01-03
Improve gdb dynamic register creation and fix drt ##debug

yossizap:
`drt` didn't print most registers due to the arch size restriction.

Gdbr - Some xmm registers were printed as fpu and bnd registers were shown in all drt categories since they were marked as flags, moved to seg since we don't have a more appropriate section.


radare:
we shuoldnt be using tabs in the output of commands. its not handled propery in rcons and may lead to glitches when displaying that output in visual or panels

yossizap:
Those tabs are only used for register profile formatting before handing it over to r_reg_set_profile_string, same as all other register profiles. `drp` will format it with spaces when printing.

PR #15758, opened 2020-01-04
Pass 64-bit sdb_fmt parameters as 64-bit

tbodt:
sdb_fmt(PFMT64x, 0) is undefined behavior if int is 32 bits, because the integer constant 0 is an int and vsnprintf will try to read a long long off the stack, resulting in overflow.

### DIFF
--- a/libr/bin/filter.c
+++ b/libr/bin/filter.c
@@ -98,7 +98,7 @@ R_API void r_bin_filter_sym(RBinFile *bf, HtPP *ht, ut64 vaddr, RBinSymbol *sym)
 	}
 	sym->dup_count = 0;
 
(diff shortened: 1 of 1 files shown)


tbodt:
I'm also curious to know why a custom PFMT64x macro is defined instead of using PRIx64 from inttypes.h.

radare:
good catch

radare:
PFMT64x was defined before PRIx64 existed on windows, same goes for all the other inttypes overrides we have in r2land, we may probably want to standarize it at some point, but its kind of a habit right now

radare:
actually whats the point on formatting a zero at all? you can just inline it in the format string

tbodt:
Agreed putting it in the format string makes more sense in this case.

radare:
So please update the pr

tbodt:
Thanks for reminding me.

PR #15763, opened 2020-01-04
Fix dbg_bps tests that always succeeded, regardless of output

yossizap:


test/new/db/archos/linux-x64/dbg_bps
```diff
@@ -197,23 +197,27 @@ EOF
 RUN
 
 NAME='valid db'
-FILE=../bins/elf/analysis/x86-helloworld-gcc
+FILE=../bins/elf/analysis/x64-loop
```
radare: Why use a different bin?

yossizap: Hello world printed
```
valid
+Hello world!
```

When it was ran

test/new/db/archos/linux-x64/dbg_bps
```diff
@@ -197,23 +197,27 @@ EOF
...
 NAME='invalid db'
-FILE=../bins/elf/analysis/x86-helloworld-gcc
+FILE=../bins/elf/analysis/x64-loop
 ARGS=-d
 CMDS=<<EOF
-db entry0
-db~1[12]
-EXPECT=<<RUN
-"valid"
+e scr.color=0
```
radare: I do t see the point on setting a breakpoint and not running the code to hit it

yossizap: I am only testing that the fields are properly updated, you will never hit an invalid breakpoint anyways and we have tests for valid breakpoints.

radare: the output matters imho

XVilka:
Good to merge? Please decide

ret2libc:
I don't understand what's the purpose of this PR honestly. As the title says, however, it seems these dbg tests always pass, no matter what. However, even with these changes they always seem to pass, which is strange :/

yossizap:
>I don't understand what's the purpose of this PR honestly. 
>these changes they always seem to pass, which is strange :/

'Always succeeded' as in 'no matter what', the check there was for `db~1[12]` and succeeded even though that field doesn't print `invalid`. With these changes the tests actually fail on wrong output and check the correct fields.

XVilka:
@radare why remove? It should be fine to merge if you agree, otherwise - just close.

ret2libc:
Let's merge this, @yossizap is right. Current tests accept any value no matter what, which is of course wrong. This PR improves the situation.

That said, dbg tests need a lot of love!

radare:
I dont think the change is right because its not doing whst it was suposed to do but i dont wanna block things but i don’t think its improving anything either. Anyway i think i wont care about this until we switch to the new V suite completely

PR #15776, opened 2020-01-08
Add null to pj and use for relocs

thestr4ng3r:
`"N/A"` makes no sense in json

### DIFF
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -1618,8 +1618,6 @@ static int bin_relocs(RCore *r, int mode, int va) {
 				// check if name is available
 				if (relname && *relname) {
 					pj_ks (pj, "name", relname);
-				} else {
-					pj_knull (pj, "name");
 				}
 				pj_ks (pj, "demname", mn ? mn : "");
 				pj_ks (pj, "type", bin_reloc_type_name (reloc));
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -1616,7 +1616,11 @@ static int bin_relocs(RCore *r, int mode, int va) {
 				}
 
 				// check if name is available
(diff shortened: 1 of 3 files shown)


libr/core/cbin.c
```diff
@@ -1616,7 +1616,11 @@ static int bin_relocs(RCore *r, int mode, int va) {
 				}
 
 				// check if name is available
-				pj_ks (pj, "name", (relname && strcmp (relname, "")) ? relname : "N/A");
+				if (relname && *relname) {
+					pj_ks (pj, "name", relname);
+				} else {
+					pj_knull (pj, "name");
```
radare: what about not showing this field just in case there's no name, programatically it will be handled the same when parsing it and it is how its done in other places in r2

libr/include/r_util/pj.h
```diff
@@ -28,13 +28,15 @@ R_API PJ *pj_o(PJ *j);
 R_API PJ *pj_a(PJ *j);
 /* keys, values */
 R_API PJ *pj_k(PJ *j, const char *k);
+R_API PJ *pj_knull(PJ *j, const char *k);
```
radare: i dont like to use "null" here wwhen all the other cases have 1 letter to define the type.

thestr4ng3r: propose a better name

libr/core/cbin.c
thestr4ng3r: It will not be handled the same in almost all parsers. But whether it is better in this case to omit the entire name or use null is a subjective question. If you prefer to have it removed, I can do that.

radare:
I cant. Thats why i didnt :D

radare:
I prefer removal. But this conflicts with json schemas i think

thestr4ng3r:
I don't know the exact specification of json schema, but I am pretty sure it is possible to model both optional fields and nullable fields with it.

PR #15782, opened 2020-01-09
Linux multithread debug fixes

yossizap:
* New threads weren't properly attached to on creation, causing issues with breakpoints and other ptrace events until they were manually attached.
* The plugin wouldn't properly trace breakpoint hits on different threads since they weren't switched to and the events weren't always handled.

Since the breakpoints are removed after they are found in one of the threads, it's best to stop all threads for now even if dbg.threads is false.


libr/debug/p/native/linux/linux_debug.c
```diff
@@ -491,15 +504,35 @@ static bool linux_kill_thread(int tid, int signo) {
...
+bool linux_stop_threads(RDebug *dbg, int except) {
+	// Stop all currently debugged threads
+	RList *list = dbg->threads;
+	bool ret = true;
+	if (list) {
+		RDebugPid *th;
+		RListIter *it;
+		r_list_foreach (list, it, th) {
+			if (th->pid && th->pid != except) {
+				if (!linux_stop_thread (th->pid)) {
```
ITAYC0HEN: extra `if`, can be combined

radare: i think it's fine like this. it's more readable

yossizap: It would print an unnecessary error if th->pid == except or if th->pid was undefined.

PR #15785, opened 2020-01-10
Fix broken AppVeyor due to rejected apostrophes

kazarmy:


XVilka:
In the future probably it's better to put this function in r_util and reuse.

PR #15792, opened 2020-01-11
Fix crash caused by rebasing a file without sections ##debug

### LINKED ISSUE #15791: Segfault when entering debug on malloc://
### Work environment

| OS/arch/bits (mandatory)                             | Arch Linux 64bit
| File format of the file you reverse (mandatory)      | N/A
| Architecture/bits of the file (mandatory)            | N/A
| r2 -v full output, **not truncated** (mandatory)     |r2 -v
radare2 4.1.1 24238 @ linux-x86-64 git.4.1.1-21-g63d25a553

#### Description

When opening r2 with `malloc://` and entering debug mode using `ood` or `doo`, radare2 segfaults.

### Expected behavior
Radare2 should print an error that debug is unavailable instead of crashing.

For example, when using `--` (note the double dash)` radare2 will block entering debug mode and write a warning.
```
$ r2 --

[0x00000000]> doo
No file open?
```

### Actual behavior
Radare2 crashes
```
$ r2 -

[0x00000000]> doo
Cannot debug file (malloc://512) with permissions set to 0x7.
Reopening the original file in read-only mode.
```

- Open r2 in malloc mode (`r2 -`)
- Enter `doo`
- Crash

image

yossizap:
closes #15791 
Had this happen while remote debugging a firmware image without sections

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -888,11 +888,17 @@ R_API void r_core_file_reopen_in_malloc(RCore *core) {
 }
 
 static RList *__save_old_sections(RCore *core) {
-	// Do I really need this?
 	RList *sections = r_bin_get_sections (core->bin);
 	RListIter *it;
 	RBinSection *sec;
 	RList *old_sections = r_list_new ();
+
+	// Return an empty list
+	if (!sections) {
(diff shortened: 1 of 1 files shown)


radare:
What about using debug maps instead of bin sections? We want all flags to be rebased not just the ones from rbin

PR #15796, opened 2020-01-13
Enable V testsuite on GitHub CI

XVilka:


XVilka:
@radare is there anything should be added for this to run? Or we should wait until https://github.com/radareorg/radare2/issues/15797 is fixed?

radare:
A quick solution would be to make a symlink installation and change some env vars to tell pkgconfig find the libs there

XVilka:
Depends on https://github.com/vlang/v/issues/3500 to be fixed.

PR #15800, opened 2020-01-13
Add register comment support and drC/j to view them

yossizap:
We are planning to add documentation on register hover to Cutter's register widget((Mainly for FPU regs), this change enables it.

The profile comment command was moved from drC to drcp.


ITAYC0HEN:
The changes look good! Maybe add a test or two?

yossizap:
Need to add register comments first to have something to test. I am working on windows fpu support and I'll add comments + tests to it once it's done.

radare:
Lgtm! You can add the tests in a separate pr!

PR #15802, opened 2020-01-14
Some coverity fixes

XVilka:


XVilka:
Ignore the OpenBSD failure - it's the network problem on their side.

Issue #8966, opened 2017-12-05
gdbserver reverse continue/step support


Manouchehri:
```
vagrant@vagrant:~$ r2 -D gdb gdb://127.0.0.1:4444 `which curl`
= attach 6371 1
= attach 6371 0
 -- Get a free shell with 'ragg2 -i exec -x'
[0x00006dc0]> dc
= attach 6371 0
[+] signal 9 aka SIGKILL received 0
[0x70000002]> dcb
cannot continue back
[0x70000002]>
```
```
vagrant@vagrant:~$ r2 -D gdb gdb://127.0.0.1:4444 `which curl`
= attach 6371 1
= attach 6371 0
 -- Run .dmm* to load the flags of the symbols of all modules loaded in the debugger
[0x00006dc0]> ds 16
= attach 6371 6371
[0x7f4c386849d6]> dsb
cannot step back
```

Example session with gdb:

```
vagrant@vagrant:~$ gdb -q `which curl`
Reading symbols from /usr/bin/curl...(no debugging symbols found)...done.
(gdb) target remote localhost:4444
Remote debugging using localhost:4444
0x00007f4c38683c30 in ?? ()
(gdb) c
Continuing.

Program received signal SIGKILL, Killed.
0x0000000070000002 in ?? ()
(gdb) rc
Continuing.

Program stopped.
0x00007f4c38683c30 in ?? ()
(gdb) c
Continuing.

Program received signal SIGKILL, Killed.
0x0000000070000002 in ?? ()
(gdb) rsi 1
0x00007f4c3845a3ed in ?? ()
(gdb) si
0x0000000070000000 in ?? ()
(gdb) rsi 16
0x00007f4c38456be3 in ?? ()
```

radare:
The gdb thing doesnt works like that but the crash must be fixed

XVilka:
@radare no, it's just a matter of supporting reverse debugging in gdb protocol, and connecting it with reverse step/continue commands.

yossizap:
>no, it's just a matter of supporting reverse debugging in gdb protocol, and connecting it with reverse step/continue commands

@XVilka why should r2 depend on gdbserver's `record` implementation, which isn't always available, when it's possible to simply set the state from r2's history as in any other debugger?

It seems that a wrong implementation/usage of gdbserver's '[G](https://sourceware.org/gdb/onlinedocs/gdb/Packets.html#write-registers-packet)' (write registers) is the issue. Currently looking into this.

[...]

PR #15814, opened 2020-01-14
Added reverse step and continue support to gdbr ##debug

yossizap:
This feature only works with server implementations that have ReverseStep and ReverseContinue enabled, such as mozilla's [rr](https://github.com/mozilla/rr)(used to test this feature). The official gdbserver doesn't support it. `=!rd` was added so that external tools will be able to check support.

ref #8966, don't close it though since these commands only work for special gdbservers.


XVilka:
Could you please add a few notes here?
- https://github.com/radareorg/radare2book/blob/master/debugger/remote_gdb.md
- https://github.com/radareorg/radare2book/blob/master/debugger/revdebug.md

Issue #13626, opened 2019-04-04
Plugin version check shouldnt check minor version numbers


radare:
3.4.0, 3.4.1, 3.4.2... should be ABI stable

PR #15816, opened 2020-01-15
Upgrade to acr-1.8 to get semver support ##core

radare:
- R2_VERSION_MAJOR
- R2_VERSION_MINOR
- R2_VERSION_PATCH


radare:
@thestr4ng3r i would like to kill r_version.h.in because its meson specific. and this r_version.h.in need to support those MAJOR, MINOR, PATCH things.. otherwise meson builds wont get this feature.

in addition those variables are inside a `R_MESON_VERSION` block and they are never set with configure.acr, so it seems like there are more inconsistencies between meson and acr builds

```
$ grep @ libr/include/r_userconf.h    
#define R2_WWWROOT "@WWWROOT@"
#define R2_PLUGINS "@PLUGINS@"
#define R2_EXTRAS "@EXTRAS@"
#define R2_BINDINGS "@BINDINGS@"
#define R2_DATDIR_R2 "@DATADIR_R2@"
#define R2_SDB "@SDB@"
#define R2_ZIGNS "@ZIGNS@"
#define R2_THEMES "@THEMES@"
#define R2_FORTUNES "@FORTUNES@"
#define R2_FLAGS "@FLAGS@"
#define R2_HUD "@HUD@"
```

xarkes:
> in addition those variables are inside a R_MESON_VERSION block and they are never set with configure.acr

If it's never set with configure.acr, why are they present in `configure.h.acr`?

I agree that it sucks to have to maintain two kind of files, but when you think about it the format looks very similar so it doesn't look too hard to me to use only one for both acr and meson.

ret2libc:
Instead of killing `r_version.h.in` I think the ACR build should just use it. It makes more sense to me, because the Makefile creates anyway the r_version.h file during the build.

configure/acr:
- r_userconf.h.acr becomes r_userconf.h at build time
- r_version.h is generated from nowhere at build time

meson:
- r_userconf.h.acr becomes r_userconf.h at build time
- r_version.h.in becomes r_version.h at build time

I think the meson behaviour is better and we should just adapt configure/acr to use r_version.h.in instead of deleting it.

Moreover, I think it's better to put R2_VERSION_MAJOR, R2_VERSION_MINOR, and R2_VERSION_PATCH in the rule `libr/include/r_version.h` in the Makefile instead of putting them in r_userconf.h.acr. So I think this PR should be fixed.

radare:
related https://github.com/radareorg/radare2/issues/13626

radare:
agree on putting those vars in r_version.h but this file is generated in a makefile because people run make more frequently than configure, on meson that would be the same. anyway, can you make meson fill those major, minor, patch vars?

ret2libc:
> agree on putting those vars in r_version.h but this file is generated in a makefile because people run make more frequently than configure, on meson that would be the same. anyway, can you make meson fill those major, minor, patch vars?

Yes, I can do that for meson.

radare:
ill wait for you to update this PR with the meson part to get this merged

radare:
appveyour looks red

```
21136
21137
21138       "C:\projects\radare2\build\radare2.sln" (default target) (1) ->
21139       "C:\projects\radare2\build\libr\core\f105261@@r_core@sta.vcxproj" (default target) (41) ->
21140         c:\projects\radare2\build\r_userconf.h(15): warning C4005: 'R2_VERSION_MAJOR': macro redefinition [C:\projects\radare2\build\libr\core\f105261@@r_core@sta.vcxproj]
21141         c:\projects\radare2\build\r_userconf.h(16): warning C4005: 'R2_VERSION_MINOR': macro redefinition [C:\projects\radare2\build\libr\core\f105261@@r_core@sta.vcxproj]
21142         c:\projects\radare2\build\r_userconf.h(17): warning C4005: 'R2_VERSION_PATCH': macro redefinition [C:\projects\radare2\build\libr\core\f105261@@r_core@sta.vcxproj]
21143         c:\projects\radare2\libr\core\agraph.c(315): warning C4566: character represented by universal-character-name '\u25EF' cannot be represented in the current code page (1252) [C:\projects\radare2\build\libr\core\f105261@@r_core@sta.vcxproj]
21144
21145
21146       "C:\projects\radare2\build\radare2.sln" (default target) (1) ->
21147
```

those are warnings but at the end it shows cnt find the bins... those warnings can be may be fixed with just an ifdef. what do you think?

ret2libc:
Again, before merging this you should move the R2_VERSION_MAJOR,MINOR,PATCH to the r_version.h file which is generated in the Makefile.

libr/include/r_userconf.h.acr
```diff
@@ -12,6 +12,12 @@
 #define HAVE_CLOCK_NANOSLEEP @HAVE_CLOCK_NANOSLEEP@
 #define HAVE_SIGACTION @HAVE_SIGACTION@
 
+#ifndef R2_VERSION_MAJOR
```
ret2libc: Why?

config-user.mk.acr
```diff
@@ -34,6 +34,10 @@ INSTALL_MAN=@INSTALL_MAN@
 INSTALL_LIB=@INSTALL_LIB@
 
 VERSION=@VERSION@
+R2_VERSION=@R2_VERSION@
```
ret2libc: This is not working! After you run `./configure` it just remains `R2_VERSION=@R2_VERSION@` and the text doesn't get replaced with the actual value.

config-user.mk.acr
```diff
@@ -34,6 +34,10 @@ INSTALL_MAN=@INSTALL_MAN@
...
+R2_VERSION_MAJOR=@R2_VERSION_MATCH@
```
ret2libc: Same for all these vars.

radare:
should be grin now. waiting. do u want to add the "int" version too? or maybe in a sepraate pr?

Issue #15815, opened 2020-01-15
mips remote debugging is not working


jishoon:
### Abstract

It seems that `radare2` has byte `endianness` & `register profile` problem for mips remote debugging. The byte endian is always `little endian` even though the target is `big endian`.
 
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Ubuntu 18.04 x86 64
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            |  mips, mipsel
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.2.0-git 23593 @ linux-x86-64 git.4.1.1-120-gabb30be72 commit: abb30be72abf54abab055eb28fd35b23791cd9dc build: 2020-01-15__11:43:38

### Expected behavior

debugging mips binary remotely via gdb remote serial protocol.

#### MIPSEL binary debugging using qemu-mipsel-static

debugging is working. I can dump virtual memory & register. But when open the `visual mode` the gdb connection closed with below qemu message.

```
$ qemu-mipsel-static -g 31337 ./bof.mipsel hello
qemu-mipsel-static: /build/qemu-XrmZRw/qemu-2.11+dfsg/accel/tcg/translate-all.c:2114: page_check_range: Assertion `start < ((target_ulong)1 << L1_MAP_ADDR_SPACE_BITS)' failed.
qemu:handle_cpu_signal received signal outside vCPU context @ pc=0x60166552
```

#### MIPS binary debugging using qemu-mips-static

Debugging is not working. This is the register dump.

```
$ r2 -a mips -d gdb://127.0.0.1:31337
Module version mismatch /home/user/.local/share/radare2/plugins/debug_unicorn.so (4.1.1) vs (4.2.0-git)
= attach 1 0
 -- Disassemble?! No Disassemble Johnny No. 5!!!
[0x40034000]> dr
zero = 0x00000000
at = 0x00000000
v0 = 0x00000000
v1 = 0x00000000
a0 = 0x00000000
a1 = 0x00000000
a2 = 0x00000000
a3 = 0x00000000
t0 = 0x00000000
t1 = 0x00000000
t2 = 0x00000000
t3 = 0x00000000
t4 = 0x00000000
t5 = 0x00000000
t6 = 0x00000000
t7 = 0x00000000
s0 = 0x00000000
s1 = 0x00000000
s2 = 0x00000000
[...]

PR #15820, opened 2020-01-16
WIP: Fix r_list_set_n() to allow empty element

jishoon:
`r_list_set_n` is used to replace `n`'th element in a given list. If one element in the list is empty, the iteration will be terminated and no chance to compare tailing indexes.

(see related issue: https://github.com/radareorg/radare2/issues/15815)

Thus, I changed this function to allow an `empty` list object. and compare the `data` just before `free`.

### DIFF
--- a/libr/util/list.c
+++ b/libr/util/list.c
@@ -418,7 +418,7 @@ R_API int r_list_set_n(RList *list, int n, void *p) {
 	r_return_val_if_fail (list, false);
 	for (it = list->head, i = 0; it ; it = it->n, i++) {
 		if (i == n) {
-			if (list->free && it->data) {
+			if (list->free) {
 				list->free (it->data);
 			}
 			it->data = p;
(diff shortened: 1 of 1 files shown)


radare:
why would we hold nulls in a list? is this any specific use casae for this situtaion? calling free(NULL) shouldnt be a problem btw

libr/util/list.c
```diff
@@ -416,9 +416,9 @@ R_API int r_list_set_n(RList *list, int n, void *p) {
 	int i;
 
 	r_return_val_if_fail (list, false);
-	for (it = list->head, i = 0; it && it->data; it = it->n, i++) {
+	for (it = list->head, i = 0; it ; it = it->n, i++) {
 		if (i == n) {
-			if (list->free) {
+			if (list->free && it->data) {
```
ret2libc: this is not necessary.

jishoon:
@radare Yes, there is a use case that NULL data is pushed to a list. And In that case, `r_list_set_n` is not working correctly. Maybe there is another list APIs that affected same issue. I will check it.

libr/util/list.c
jishoon: yep, I removed this change.

PR #15822, opened 2020-01-16
Fix incorrect jump table size #13812

### LINKED ISSUE #13812: Incorrect switch table sizes
### Work environment

| OS/arch/bits (mandatory)                             | Fedora 29 x86_64
| Architecture/bits of the file (mandatory)            | x86_64 and x86
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.5.0-git 21454 @ linux-x86-64 git.3.4.1-161-ge10d7df60 commit: e10d7df60051b6d326e8ee7366d497e1e0692336 build: 2019-04-17__11:29:17

### Expected behavior
When switch input is in memory, properly determine the number of cases.

### Actual behavior
If the input value is on the stack, the third path is taken in this code from try_get_delta_jmptbl_info. This is used

```
		if (tmp_aop.val == UT64_MAX && tmp_aop.refptr == 0) {
			isValid = true;
			*table_size = 0;
		} else if (tmp_aop.refptr == 0) {
			isValid = tmp_aop.val < 0x200;
			*table_size = tmp_aop.val + 1;
		} else {
			isValid = tmp_aop.refptr < 0x200;
			*table_size = tmp_aop.refptr + 1;
		}
```

This might be intended for a case that calculates the table offset and then compares the final value to the end of the table? It would be helpful to have sample code that utilizes this case. Maybe we could check that the base register is IP.

The following function was compiled with gcc version 8.3.1

```
void a_switch(int n)
{
  switch(n) {
  case 1:
    printf("one");
    break;
  case 2:
    printf("two");
    break;
  case 3:
    printf("three");
    break;
  case 4:
    printf("four");
    break;
  case 5:
    printf("five");
    break;
  case 6:
    printf("six");
    break;
  }
}
```

In the case of x86/32, the function begins as follows. The refptr is 8 and this is not recognized as a switch table.
```
08049186 <a_switch>:
 8049186:       55                      push   %ebp
 8049187:       89 e5                   mov    %esp,%ebp
 8049189:       83 ec 08                sub    $0x8,%esp
 804918c:       83 7d 08 06             cmpl   $0x6,0x8(%ebp)
 8049190:       77 7a                   ja     804920c <a_switch+0x86>
```

For x86/64, the refptr is 4 and radare2 interprets this as a 5 case switch (4+1).
```
0000000000401136 <a_switch>:
  401136:       55                      push   %rbp
  401137:       48 89 e5                mov    %rsp,%rbp
  40113a:       48 83 ec 10             sub    $0x10,%rsp
  40113e:       89 7d fc                mov    %edi,-0x4(%rbp)
  401141:       83 7d fc 06             cmpl   $0x6,-0x4(%rbp)
  401145:       77 72                   ja     4011b9 <a_switch+0x83>
```

Doesn't repro with ARM since cmp isn't a memory access function in that architecture.

d4em0n:
Fix #13812

### DIFF
--- a/libr/anal/jmptbl.c
+++ b/libr/anal/jmptbl.c
@@ -322,7 +322,7 @@ R_API bool try_get_jmptbl_info(RAnal *anal, RAnalFunction *fcn, ut64 addr, RAnal
 		if (tmp_aop.val == UT64_MAX && tmp_aop.refptr == 0) {
 			isValid = true;
 			*table_size = 0;
-		} else if (tmp_aop.refptr == 0) {
+		} else if (tmp_aop.refptr == 0 || tmp_aop.val != UT64_MAX) {
 			isValid = tmp_aop.val < 0x200;
 			*table_size = tmp_aop.val + 1;
 		} else {


radare:
there are 4 broken tests. canyou check if the new output is because the analysis have improved or not? search for "XX]" in the the travis logs. thanks!

also, can you rename the commiit to be like "Fix #... - message"? this way github will close the issue when merged.

d4em0n:
I can only confirm the test for cmu_binary_bomb is improved. I'm not sure the other and don't know how that's calculated

radare:
Update if needed

radare:
ping, ill be testting this pr now, because i know itst important to get this right for 4.3, but rls is in 11h, so i would appreciate more tetsting and feedback from everyone else in r2land. thanks

radare:
just confirmed the cmu bin is properly analized now. going to chk the other 3

radare:
i can confirm all the other cases are good too. comparing with source code it founds the right amount of cases

radare:
ill fix the tests after merging. thanks

Issue #15468, opened 2019-11-15
Broken tests on PowerPC64 (Little Endian) and System Z architecturs


XVilka:
119 failed tests for PPC64LE - https://travis-ci.com/radareorg/radare2/jobs/256856419

XVilka:
107 now https://travis-ci.com/radareorg/radare2/jobs/269167852
Many of them just assume x86 arch is set, should be fixed

XVilka:
Around 80 left

radare:
yay \o/

radare:
ive fixed some, now the numbers are:

ARM64 18
PPC64 22
S390X > 1000
newShell 96

radare:
see https://github.com/radareorg/radare2/pull/16435

XVilka:
12 failing tests for PPC now: https://travis-ci.com/github/radareorg/radare2/jobs/423694412#L794

XVilka:
4 failing tests for PPC now: https://travis-ci.com/github/radareorg/radare2/jobs/431985568#L549

cc  @eagleoflqj

XVilka:
See also build warnings:
```
p/debug_native.c: In function ‘r_debug_native_bp’:
p/debug_native.c:1421:10: warning: unused variable ‘dbg’ [-Wunused-variable]
  RDebug *dbg = bp->user;
          ^~~
p/native/linux/linux_debug.c: In function ‘print_fpu’:
p/native/linux/linux_debug.c:1058:2: warning: #warning print_fpu not implemented for this platform [-Wcpp]
 #warning print_fpu not implemented for this platform
  ^~~~~~~
p/native/linux/linux_debug.c: In function ‘linux_reg_read’:
p/native/linux/linux_debug.c:1063:7: warning: variable ‘showfpu’ set but not used [-Wunused-but-set-variable]
  bool showfpu = false;
       ^~~~~~~
p/native/linux/linux_debug.c: In function ‘linux_reg_write’:
p/native/linux/linux_debug.c:1268:55: warning: passing argument 4 of ‘r_debug_ptrace’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
   int ret = r_debug_ptrace (dbg, PTRACE_SETREGS, pid, buf, NULL);
                                                       ^~~
In file included from p/native/linux/linux_debug.c:6:0:
[...]

PR #15834, opened 2020-01-17
Fix wrong tests

XVilka:
Addressing this https://github.com/radareorg/radare2/issues/15468

There are many left still.


XVilka:
Cool, 20 broken tests less on PPC/ARM64

PR #15837, opened 2020-01-17
Update liberty from git for new C++, Rust and D demanglers ##bin

radare:


karliss:
Looks like this one is too new, it has removed support for gcc 2.x mangling scheme. See https://github.com/gcc-mirror/gcc/commit/6c8120c5ff130e03d32ff15a8f0d0e703592a2af#diff-5ec9e1522a9e8d70c2b7d0e39b854223

radare:
do we want to keep supporting the gcc 2.x demangling support?

radare:
@karliss we really want to keep old gcc 2.x demangling, so we probably need a way to specify the flavour /version of the symbol demangler. i still think that we should move the demangler thing into a separte library

XVilka:
In ideal case it would be nice to autotedect which demangler to call.

karliss:
@radare at the moment there is nothing to keep.  When I looked at the current code it was stripped down with implementations for many mangling schemes removed.

karliss:
Situations like this is why I am not a big fan of r2 approach - copy pasting parts of code from other projects without clear indication what, which version, from where and with what modifications is taken. It later causes confusion wether the bugs or lack of features is caused by original project, bad version of original project or modifications that are not part of original project.  Upgrade process also becomes more difficult as it is unknown what are the changes that need to be preserved.

radare:
if we dont merge this we have gcc2 demangling iirc . Thats what im refering to

karliss:
Do you have any proof confirming this? Here are the reasons I think current code doesn't support it:
* The code explicitly calls method named demangle_v3
* The bug report complaining that string looking like gcc 2 mangled name isn't demangled
* File containing old demangling code cplus-dem.c which was removed in the commit I linked isn't present in r2 repository
* Constants/enum values referencing to older demangling modes aren't used anywhere

radare:
Nothing is removed this is just a pr im proposing to keep the old and the new parsers into a separate lib we can use from r2 and not merge the code in here because its not really our code. Just a ripoff from gnu

ret2libc:
@karliss @radare so what's the status of this? If this PR in its current state is not ok, we should probably close it and re-think about it when needed.

karliss:
There might be some miscommunication.  They way I see it.
* Current code in r2 master doesn't support old demangling schemes because previously when the code was imported some parts where removed (I guess this was done to reduce size)
* Code in this PR doesn't support old demangling schemes because upstream version of the library has removed them due them being old

Assuming the code in this PR contains clean version (without unknown modifications) of updated upstream code I have no objections against it. It doesn't make things worse and improves demangling of new C++, Rust, D code. Things that didn't work still don't work but due to different reasons.

I have no objections against having the old library as separate lib to support old schemes. But @radare comment about proposing to do that with this PR doesn't match with the first commit message.

radare:
i want to remove all the demangling code in r2 and just use libdemangle, which will handle all the versions of c++ and such. i ust need to find some time to finish it and integrate it here. i keep this pr open to copypaste the code in there

radare:
I have moved this pr into the ldmg library that im splitting out from r2 rbin code. Feel free to help on that external library to get it fully tested, maintained and usable from r2 as well as bindings and other projects.

https://github.com/radareorg/ldm

radare:
@karliss please take a look at the ldmg library, can you send a PR adding the code to demangle cxx2 symbols there and submit some test cases for it?

PR #15841, opened 2020-01-18
Remove realname from strings

ITAYC0HEN:
After discussions, this PR will make sure that real name isn't shown for strings.

This will return the good old `
```
mov ebx, str.Think_you_can_make_it
```
instead of 
```
mov ebx, "Think you can make it????!!!!>>?"
```
which was added recently in the realname PR.


radare:
red

PR #15843, opened 2020-01-18
CBM Basic Support

thestr4ng3r:


XVilka:
Please rebase.

radare:
Lgtm but the thing is.. why not in extras?

thestr4ng3r:
More like, why in extras?

radare:
i dont think there's much need on having CBM support by the 99% of the r2 users. so we probably dont want to ship in the default installation because its making builds slower and bigger (not a lot, but a bit)

thestr4ng3r:
Will be moved to another repo + r2pm package

thestr4ng3r:
Code moved here: https://github.com/thestr4ng3r/r2commodore

PR #15845, opened 2020-01-18
Atl flagtable

radare:


libr/core/cmd_debug.c
```diff
@@ -2347,6 +2348,52 @@ static void cmd_debug_reg_print_packed_reg(RCore *core, RRegItem *item, char exp
...
+	r_table_add_column (t, typeBoolean, "float", 0);
+	r_table_add_column (t, typeString, "name", 0);
+	r_table_add_column (t, typeString, "flags", 0);
+	r_table_add_column (t, typeString, "comment", 0);
+	for (i = 0; i < R_REG_TYPE_LAST; i++) {
+		RList *list = r_reg_get_list (reg, i);
+		RListIter *iter;
+		r_list_foreach (list, iter, e) {
+			r_table_add_row (t,
+					sdb_fmt ("%d", e->offset),
```
ret2libc: I'm on phone right now so I cannot check myself, but does sdb_fmt have enough static vars for all these calls at the same time?

libr/core/cmd_debug.c
```diff
@@ -2347,6 +2348,52 @@ static void cmd_debug_reg_print_packed_reg(RCore *core, RRegItem *item, char exp
...
+					);
+		}
+	}
+	const char fmt = *str++;
+	const char *q = str;
+	if (r_table_query (t, q)) {
+		char *s = (fmt == 'j')
+			? r_table_tojson (t)
+			: r_table_tofancystring (t);
+		// char *s = r_table_tostring (t);
```
ret2libc: Just remove

libr/core/cmd_flag.c
```diff
@@ -420,6 +421,47 @@ static int flag_to_flag(RCore *core, const char *glob) {
 	return 0;
 }
 
+typedef struct {
+	RTable *t;
+} FlagTableData;
+
+static bool __tableItemCallback(RFlagItem *flag, void *user) {
+	FlagTableData *ftd = user;
+	if (flag->name && *flag->name) {
```
ret2libc: There is already a macro for this. It should be called R_STR_EMPTY or something like that

libr/core/cmd_debug.c
radare: yep it does, but its not thread safe :{

PR #15846, opened 2020-01-18
Fix #14424 - Handle ~ operator in the `ms` shell ##cons

### LINKED ISSUE #14424: ms shell doesnt handle ~or |
Will be good to be able to do thing like "ls | grep" or "ls~foo"

Example:
```
$ r2 -
> m posix /
> ms
[/]> ls | grep usr
[/]> ls~usr
```

radare:


libr/fs/fs.c
```diff
@@ -82,13 +82,12 @@ R_API void r_fs_add(RFS* fs, RFSPlugin* p) {
 		p->init ();
 	}
 	RFSPlugin* sp = R_NEW0 (RFSPlugin);
-	if (!sp) {
-		return;
-	}
-	if (p) {
-		memcpy (sp, p, sizeof (RFSPlugin));
+	if (sp) {
```
ret2libc: Why these kind of changes? I think it's easier to read when you have early returns instead of wrapping things inside if inside if. If you do this when you have even just 2 allocations you end up with 2 levels of indentation, which is very hard to read

radare: imho it depends on the length of the function, in this case it is more readable like this imho, there's only one level of nestinng the 5 lines of code

ret2libc: Ok np! Though it needs to be changed if/when you add a new allocation

Issue #15819, opened 2020-01-16
"afs" wont change signature if the return type is not specified


oxcabe:
```
[0x00001179]> afs
fcn.00001179 (int64_t arg1, int64_t arg3);

[0x00001179]> "afs test(int tost)"
[0x00001179]> afs
fcn.00001179 (int64_t arg1, int64_t arg3);

[0x00001179]> "afs void test(int tost)"
[0x00001179]> afs
void test (int32_t tost);
```

Putting it on an issue so I remember to fix it

radare:
in time for 4.2? (3 days left)

PR #15847, opened 2020-01-19
Fix afs not working without rettype ##anal

oxcabe:
Fixes https://github.com/radareorg/radare2/issues/15819.


XVilka:
Can you please add a test? Otherwise, LGTM.

PR #15859, opened 2020-01-20
Treat empty string as false when boolifying

kazarmy:


radare:
9 XX, 1 of them fixed in mastter, pls rebase

radare:
@ret2libc comments here?

kazarmy:
rebased, don't merge yet

libr/config/config.c
```diff
@@ -55,8 +55,8 @@ static void config_print_value_json(RConfig *cfg, RConfigNode *node) {
 		if (!strncmp (val, "0x", 2)) {
 			ut64 n = r_num_get (NULL, val);
 			cfg->cb_printf ("%"PFMT64d, n);
-		} else if (r_str_isnumber (val) || r_str_is_bool (val)) {
-			cfg->cb_printf ("%s", val);
+		} else if (r_str_isnumber (val) || (*val /* HACK */ && r_str_is_bool (val))) {
```
ret2libc: what is this hack? what does it do? it's not clear to me

libr/config/config.c
```diff
@@ -55,8 +55,8 @@ static void config_print_value_json(RConfig *cfg, RConfigNode *node) {
...
+			cfg->cb_printf ("%s", val);  // TODO: always use true/false for bool json str
```
ret2libc: what does this TODO mean? why not do it now?

kazarmy:
@ret2libc 

1. Output from a Travis build on the first commit is [here](https://travis-ci.com/radareorg/radare2/jobs/277561192). If you search for "`:,`", you will find 2 offending config vars: `cmd.esil.intr` and `cmd.pdc`.
1. The proper way to fix this is to implement the TODO, _but_ ... `cmd.esil.intr` and `cmd.pdc` are **not** boolean vars, so it doesn't make sense to assign `true`/`false` to them.
1. This suggests that there is something wrong with the r_config code, since both the config vars are not declared as boolean vars but are treated as such.
1. But for now, the hack will suffice to maintain previous behavior as encoded by the tests.

libr/config/config.c
kazarmy: see https://github.com/radareorg/radare2/pull/15859#issuecomment-576605121

libr/config/config.c
kazarmy: see https://github.com/radareorg/radare2/pull/15859#issuecomment-576605121

PR #15860, opened 2020-01-20
Check version for pkgname correctly

thestr4ng3r:
* dot can be null
* Using R2_VERSION_MAJOR/MINOR is wrong because they will be continuously updated and we want to check for exactly 4.2.0
* The version check logic itself was wrong


radare:
thanks, i need more sleep

Issue #15855, opened 2020-01-19
arm64 does not assemble "cbnz reg, label"


x0urc3:
```
$> rasm2 -a arm -b 64 "cbnz w0, 0x5c"
Cannot assemble 'cbnz w0, 0x5c' at line 3
invalid
$> rasm2 -a arm.ks -b 64 "cbnz w0, 0x5c"
e0020035
```

PR #15871, opened 2020-01-21
Add cbnz and cbz instruction for arm64 assembler

x0urc3:
Resolving #15855. Add both instructions, updated unit test and opcode description


radare:
Awesome pr! With description and test! Thank you

PR #15876, opened 2020-01-22
Refactor Anal Hints

### LINKED ISSUE #15877: Clarify Semantics of Bits/Arch Anal Hints
The situation currently is as follows:
* Most anal hints (immbase, jump/fail, ...) only apply for a single address, not a range.
* bits is assigned to a single address too without a size, but an rbtree is build to top of it and when querying for the bits at an address x, the hint with the highest addr less or equals to x is used. This essentially means all bits hints reach until the end of the address space or until the next hint.
* arch has the following syntax: `aha ppc 51        // set arch for a range of N bytes`, but the range size arg is completely ignored and the hint will only apply for a single address, so it's basically useless.

For these ranged hints, we have a few different options:
1. Make all have a size that must explicitly be specified and store in an augmented rbtree - This would not work with a lot of the analysis that sets the bits value since the size of the code that uses the bits value is generally not known.
2. Same data structure as above but add helper functions that create ranged hints at an address and automatically calculate the size to extend until the maximum addr. This would fix the analysis.
3. Make both arch and bits work like bits currently do (implicitly extend) and use one non-augmented tree for each of them for querying.

I started to do something like 1/2 here: https://github.com/radareorg/radare2/pull/15876

It would be good to get some feedback on what behavior is expected/most useful @radare @ret2libc @XVilka

thestr4ng3r:
- [x] Command help updated (I've updated the documentation and the [radare2 book](../../radare2book) with the relevant information (if needed))

**Detailed description**

This addresses hints that should affect an entire range instead of just a single address. Previously, all hints were stored in SDB (which only provides direct lookup by key) and bits had an extra tree built on top of it to provide greatest lower bound queries while arch was just broken.
Instead of further building stuff on top of SDB, I refactored the storage completely to use one `Ht<ut64, RVector<RAnalAddrHintRecord>>` and one `RBTree<ut64, ...>` for arch and bits each.
The semantics of ranged hints are as discussed in #15877: "A ranged hint affects the entire memory range >= its address until the next hint or the end of the address space". To reset the value at some address to the global setting, a reset value of NULL or 0 can be used.

**Test plan**

* Use ahb and aha and observe that all addresses after the hint show the hint on `ah.`
* Run the newly added unit tests

**Closing issues**

Fix #15877


radare:
No longer wip ? Lgtm

thestr4ng3r:
I want to add one more small thing

PR #15878, opened 2020-01-22
Don't print r2pm -ci <pkgname> if pkg version is 4.2.0-git

kazarmy:
`4.2.0-git` comes before `4.2.0`, right?

### DIFF
--- a/libr/util/lib.c
+++ b/libr/util/lib.c
@@ -308,7 +308,7 @@ R_API int r_lib_open_ptr(RLib *lib, const char *file, void *handler, RLibStruct 
 				int major = atoi (stru->version);
 				int minor = dot ? atoi (dot + 1) : 0;
 				// The pkgname member was introduced in 4.2.0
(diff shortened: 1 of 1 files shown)


kazarmy:
There's a possibility that the package/module/library/plugin won't have the `pkgname` member if its version is `4.2.0-git`.

radare:
4.2.0 is not the same as 4.2.0-git. Actually the -git means the abi can change at anytime u update the code. Versions not containing -git are basically releases

kazarmy:
Yes since the abi can change at anytime, pkgname might not exist. If it doesn't exist (and I've checked and it doesn't exist on my r2ghidra-dec installation right now), garbage can be printed as shown below:

r2ghidra-dec_no_pkgname

This shocked me the first time I saw this.

radare:
thats just a transitional issue imho, it shouldnt happen if ppl use any release or > 4.2 from git

kazarmy:
ok

PR #15879, opened 2020-01-22
Run a travis job to use the radare2-shell-parser

ret2libc:


XVilka:
LGTM. I recommend to add a keyword for these kind of FIXME comments - so you will be able to find it with simple grep after.

PR #15889, opened 2020-01-23
Add undocumented waF* command and make it a dry-run as wa*

phra:


radare:
looks red

phra:
i don't think that the failed CI status is caused by this particular PR.

PR #15890, opened 2020-01-24
r2r.v: Allow '=' in keyword argument

kazarmy:
Otherwise the following test in `tools/rahash2`, which should fail since current r2r.v doesn't support '...' tests yet:

```
NAME=rahash2 -D base64 -s "YWRtaW4="
FILE=-
CMDS=!rahash2 -D base64 -s "YWRtaW4="
EXPECT='admin'
RUN
```

actually succeeds.

### DIFF
--- a/test/src/r2r.v
+++ b/test/src/r2r.v
@@ -178,10 +178,10 @@ mut:
 }
 
 fn get_kw_arg(line string) (string,string) {
-	pos_eq := line.index('=') or {
+	i := line.index('=') or {
 		return line,''
 	}
-	return line.substr(0, pos_eq),line.substr(pos_eq + 1, line.len)
+	return line.substr(0, i),line.substr(i + 1, line.len)
 }
 
 fn (test R2RCmdTest) parse_slurp(v string) (string,string) {
--- a/test/src/r2r.v
+++ b/test/src/r2r.v
@@ -177,6 +177,13 @@ mut:
 	name string
 }
 
+fn (test R2RCmdTest) get_kw_arg(line string) (string,string) {
+	pos_eq := line.index('=') or {
+		return line,''
+	}
+	return line.substr(0, pos_eq),line.substr(pos_eq + 1, line.len)
+}
+
 fn (test R2RCmdTest) parse_slurp(v string) (string,string) {
 	mut res := ''
 	mut slurp_token := ''
@@ -233,14 +240,11 @@ fn (r2r mut R2R) load_cmd_test(testfile string) {
 			}
 			continue
 		}
-		kv := line.split_nth('=', 2)
-		if kv.len == 0 {
-			continue
-		}
-		match kv[0] {
+		kw,arg := test.get_kw_arg(line)
+		match kw {
 			'CMDS' {
-				if kv.len > 1 {
-					a,b := test.parse_slurp(kv[1])
+				if arg.len > 0 {
+					a,b := test.parse_slurp(arg)
 					test.cmds = a
 					slurp_token = b
 					if slurp_token.len > 0 {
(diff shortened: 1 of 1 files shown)


test/src/r2r.v
```diff
@@ -177,6 +177,13 @@ mut:
 	name string
 }
 
+fn (test R2RCmdTest) get_kw_arg(line string) (string,string) {
```
radare: Its not a method of r2rcmdtest. This is judt a static helper method. So remove the part betweeh () here

test/src/r2r.v
```diff
@@ -177,6 +177,13 @@ mut:
...
+	pos_eq := line.index('=') or {
+		return line,''
+	}
+	return line.substr(0, pos_eq),line.substr(pos_eq + 1, line.len)
```
radare: Use v fmt

test/src/r2r.v
```diff
@@ -233,14 +240,11 @@ fn (r2r mut R2R) load_cmd_test(testfile string) {
 			}
 			continue
 		}
-		kv := line.split_nth('=', 2)
-		if kv.len == 0 {
-			continue
-		}
-		match kv[0] {
+		kw,arg := test.get_kw_arg(line)
```
radare: Imho its k, v not kw, arg

test/src/r2r.v
```diff
@@ -177,6 +177,13 @@ mut:
 	name string
 }
 
+fn (test R2RCmdTest) get_kw_arg(line string) (string,string) {
+	pos_eq := line.index('=') or {
```
radare: I would say just “i” instead of “pos_eq”

test/src/r2r.v
radare: This is what split_nth should do. If not then this is a bug in vlib. Whats the problem u have here? Can u report the issue in v or fix this api call in vlib?

test/src/r2r.v
kazarmy: That's what v fmt wants, though I don't think v fmt generally does a good job right now

test/src/r2r.v
kazarmy: ok 7445e4ab6e9c1930f27c72314bd73c593bc9a12d

test/src/r2r.v
kazarmy: ok 956f10a3c8c14cd2edf5a22470825714a90fab9e

test/src/r2r.v
kazarmy: > Its not a method of r2rcmdtest. This is judt a static helper method. So remove the part betweeh () here

ok 42777f3546111c0b7f90842d9ab0ea12ba3a5a0b

kazarmy: > Whats the problem u have here? Can u report the issue in v or fix this api call in vlib?

The problem is, for:

```
CMDS=!rahash2 -D base64 -s "YWRtaW4="
```

split_nth('=', 2) gives:

```
kv.len: 2
kv[0]: CMDS
kv[1]: !rahash2 -D base64 -s "YWRtaW4
```

i.e. `kv[1]` is cut off at the end where the 2nd `'='` is.

Not sure whether this is a bug in split_nth or another api function is needed, e.g. split_n.

kazarmy: > This is what split_nth should do.

> Not sure whether this is a bug in split_nth or another api function is needed, e.g. split_n.

Oh it was you who introduced split_nth.

kazarmy: Submitted issue https://github.com/vlang/v/issues/3551.

radare:
lgtm for now, but will be good to fix vlib and upgrade v when the issue is gone in their side.

thanks!

PR #15891, opened 2020-01-24
sys/install.sh uses always /usr/local + rpath now ##build

radare:
Fixes https://github.com/radareorg/radare2/issues/9240

### DIFF
--- a/sys/build.sh
+++ b/sys/build.sh
@@ -52,27 +52,21 @@ if [ -z "${MAKE}" ]; then
 fi
 
 [ -z "${MAKE_JOBS}" ] && MAKE_JOBS=12
-
 [ -z "${CERTID}" ] && CERTID=org.radare.radare2
 
 # find root
 A=$(dirname "$PWD/$0")
 cd "$A" && cd .. || exit 1
 
+DEFAULT_PREFIX=/usr/local
 if [ "`uname`" = Darwin ]; then
-	DEFAULT_PREFIX=/usr/local
 	# purge previous installations on other common paths
 	if [ -f /usr/bin/r2 ]; then
 		type sudo || NOSUDO=1
 		[ "$(id -u)" = 0 ] || SUDO=sudo
 		[ -n "${NOSUDO}" ] && SUDO=
-		# purge first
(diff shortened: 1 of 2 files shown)


radare:
The problems with this change are:

* Previous installation will still be in /usr
* Some Linux distros don't have /usr/local/bin in $PATH

We can solve the first by checking if the resolved symlink of /usr/bin/radare2 is the same as `/usr/local/bin/radare2` and then perform a `make purge PREFIX=/usr`.

The second one is something that we can tell the user to deal with.. and maybe add a WARNING..

ret2libc:
Why would you want to change this? What are the reasons behind this change?

I would say that if it currently works, then it's best to keep it as is instead of breaking the user expectation and deal with users reports that will see a different behaviour compared to previous r2 versions.

Anyway, I personally never use `./sys/install.sh` because it does a bunch of things I don't have control over (like the PREFIX) and I prefer using `configure;make` or just `meson`. Distributions who package radare2 are probably doing the same.

kazarmy:
Never used `sys/install.sh` because I had the option to use `sys/user.sh` and thus better contain catastrophic errors, but switched from that to meson a long time ago because meson felt faster, especially for small changes that don't require full builds.

radare:
ping @EvanCarroll

radare:
i would love to get some feedback from people who actually ... uses this script, otherwise your comments are useless

radare:
im testing and ill merge soon, tired of waiting

ret2libc:
Why?

ret2libc:
This is actually not a real fix of the reference issue https://github.com/radareorg/radare2/issues/9240 . The referenced issue rightfully complains that ./sys/build.sh should just use the prefix set by `./configure` and not set a hard-coded prefix. `./configure` should, in theory, be able to generate the right default prefix where users can install their things. Playing with hardcoded prefix can, of course, break the system and/or the custom radare2 installation.

Anyway... for those who really care about having a clean install system with no magic scripts behind (like all those `./sys/*.sh` scripts) can use `meson` and it should work well for most distros.

PR #15906, opened 2020-01-27
Add Python bytecode magic number

x0urc3:
- Majority of the magic was taken from [here](https://github.com/radareorg/radare2-extras/blob/5eb18b3e3b358e814edf33e4cbc79fd8dca8ee1c/libr/bin/format/pyc/pyc_magic.)
- Added Python 3.7.5 bytecode magic number


x0urc3:
Too many magic number is messing up '/m' test

PR #15910, opened 2020-01-28
r2r.v: Add support for single-line '..' in tests

kazarmy:
Towards making r2r.v a drop-in replacement for node-r2r. Should be temporary. There's a bunch of tests in `tools/rahash2` that use '..'.

### DIFF
--- a/test/src/r2r.v
+++ b/test/src/r2r.v
@@ -180,8 +180,15 @@ mut:
 fn (test R2RCmdTest) parse_slurp(v string) (string,string) {
 	mut res := ''
 	mut slurp_token := ''
-	if v.starts_with("'") || v.starts_with("'") {
+	if v.starts_with("'") {
 		eprintln('Warning: Deprecated syntax, use <<EOF in ${test.source} @ ${test.name}')
+		right_quote := v.last_index("'") or {
+			// Can't happen
+			panic("Can't find right quote")
+		}
+		if right_quote != 0 {
+			res = v[1..right_quote]
+		}
 	}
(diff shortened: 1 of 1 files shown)


test/src/r2r.v
```diff
@@ -180,8 +180,15 @@ mut:
 fn (test R2RCmdTest) parse_slurp(v string) (string,string) {
 	mut res := ''
 	mut slurp_token := ''
-	if v.starts_with("'") || v.starts_with("'") {
+	if v.starts_with("'") {
 		eprintln('Warning: Deprecated syntax, use <<EOF in ${test.source} @ ${test.name}')
+		right_quote := v.last_index("'") or {
```
radare: this is deprecated syntax, cant we just fix the tests instead of extending this deprecated syntax?

kazarmy: we can fix the tests but unless you merge in https://github.com/radare/node-r2r/pull/13, current node-r2r will reject them

kazarmy: the tests can be fixed after r2r.v completely replaces node-r2r on Travis and AppVeyor, and then this code can be disabled or removed

kazarmy: the converter code in https://github.com/radare/node-r2r/pull/12/commits/656c4c412401aeb3aa2a17eed7124bd40113b762 seems to work, but has only been tested on `cmd_i` tests

radare: so can we close this pr?

kazarmy: yes

PR #15911, opened 2020-01-28
fixing memory leaks

marcograss:
we forgot to free the bin object leaking it if we bail out, plus improved the r_bin_le_free to actually free the root object

### DIFF
--- a/libr/bin/format/le/le.c
+++ b/libr/bin/format/le/le.c
@@ -609,9 +609,7 @@ static bool __init_header(r_bin_le_obj_t *bin, RBuffer *buf) {
 }
 
 void r_bin_le_free(r_bin_le_obj_t *bin) {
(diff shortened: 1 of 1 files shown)


libr/bin/format/le/le.c
```diff
@@ -609,9 +609,19 @@ static bool __init_header(r_bin_le_obj_t *bin, RBuffer *buf) {
 }
 
 void r_bin_le_free(r_bin_le_obj_t *bin) {
-	free (bin->header);
-	free (bin->objtbl);
-	free (bin->filename);
+	if (!bin) {
```
XVilka: see `r_return_if_fail()` macro.

marcograss: thanks, I changed it

PR #15924, opened 2020-02-04
Fix a bug of clicking

Vane11ope:
Clicking does not properly work in panels and now this fixes it.

### DIFF
--- a/libr/cons/input.c
+++ b/libr/cons/input.c
@@ -211,9 +211,7 @@ R_API int r_cons_arrow_to_hjkl(int ch) {
 				// M is mouse down , m is mouse up
 				if (ch == 'M' || ch == 'm') {
 					r_cons_set_click (x, y);
-					if (ch == 'm') {
-						return INT8_MAX - 1;
-					}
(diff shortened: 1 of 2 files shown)


radare:
let me test on mac before merging. thanks! what was the bug about?

ITAYC0HEN:
> what was the bug about?

@Vane11ope try following the PR template next time, it's there for a good reason :)
It'd help us understand the issue and the solution. Also, the test case to review the PR

radare:
@Vane11ope ping?

radare:
just tested this on mac, and this PR basically makes the mouse not to work on mac at all, so i wont merge this. we need to clarify what is the bug in linux and we can temporarily aadd this option for non mac builds, but this is aa hack because mac users can ssh into linux machines and be unable to use the mouse obecause of this ifdef hack.

Vane11ope:
hang on Im taking this back then. the problem is, the mouse does not properly click on the menu in linux and this fixes it. but since mac does not work with this then let me take this back

PR #15956, opened 2020-02-12
Add support for Swift calling convention

### LINKED ISSUE #14657: Add Swift calling convention
https://github.com/trufae/swift-frida/blob/master/calling-convention.js

abcSup:

**Detailed description**

```
$ r2 ~/tmp/swift-test
[0x00000f10]> aaa
[x] Check for objc references
[x] Check for vtables
[0x00000f10]> s sym.error.Thrower.throwFunc.allocator
[0x00001420]> afc swift
[0x00001420]> afva
[0x00001420]> afv
var int64_t var_48h @ rbp-0x48
var int64_t var_40h @ rbp-0x40
var int64_t var_38h @ rbp-0x38
var int64_t var_30h @ rbp-0x30
var int64_t var_28h @ rbp-0x28
var int64_t var_20h @ rbp-0x20
var int64_t var_18h @ rbp-0x18
var int64_t var_10h @ rbp-0x10
arg int64_t self @ r13
arg int64_t error @ r12
arg int64_t arg1 @ rdi
arg int64_t arg2 @ rsi
arg int64_t arg3 @ rdx
```

Currently, `error` is considered as argument because underlying functions do not support registers to be variables.

**Test plan**

A new binary compiled with the latest Swift is required to test the calling convention. I would like to ask how I should proceed with adding a new binary to the testbins

**Closing issues**

close #14657


radare:
Why WIP?

abcSup:
> Why WIP?

Currently, there is no binary in the testbins to test the functionality. Existing mach0 swift binaries, `swift-main` and `swift-main2` were compiled with older swift. I would like to know how I should proceed with adding a new binary to the test.

libr/anal/cc.c
```diff
@@ -104,6 +113,20 @@ R_API const char *r_anal_cc_arg(RAnal *anal, const char *convention, int n) {
 	return ret? r_str_constpool_get (&anal->constpool, ret): NULL;
 }
 
+R_API const char *r_anal_cc_self(RAnal *anal, const char *convention) {
+	r_return_val_if_fail (anal && convention, NULL);
+	char *query = sdb_fmt ("cc.%s.self", convention);
```
thestr4ng3r: What about calling it "this" instead of "self"? I think it's slightly more common, especially considering the term "thiscall".

thestr4ng3r:
You can build a small binary and submit it as a pull request to https://github.com/radareorg/radare2-testbins. Maybe call it `swift5.1-helloworld` or something.

radare:
self if this for swift, imho self is fine because it wont conflict with c++ because its a keyword if we end up compiling or using this field from other languages outside C/SDB.

abcSup:
> You can build a small binary and submit it as a pull request to https://github.com/radareorg/radare2-testbins. Maybe call it `swift5.1-helloworld` or something.

Sure, I will create a simple hello world program including a class that throws exception to test `error` register.

> What about calling it "this" instead of "self"? I think it's slightly more common, especially considering the term "thiscall".

> self if this for swift, imho self is fine because it wont conflict with c++ because its a keyword if we end up compiling or using this field from other languages outside C/SDB.

I can create another pull request after this to update `thiscall` cc to support the new register argument and maybe make a condition to check `bin.lang` and rename it accordingly.

radare:
feel free to submit the test now that the binary is merged in r2r-bins

PR #15957, opened 2020-02-12
r2r.v: Fix wg race

kazarmy:
- [X] I've read the [guidelines for contributing](../DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](../DEVELOPERS.md#code-style)

**Detailed description**

Seen with `./r2r -j 1 tools /r2` on 2fad5ae3c7ebb6794f30ec41a70654573b451b3c:

Total_13

Total_11

Total_11

Total_12

:scream: :scream: :scream:

This pr fixes the above and keeps everything consistent and sane (i.e. Total: 13). Yes I know it breaks CTRL&#8209;C but the above is more problematic and enotime right now to figure out how to make CTRL&#8209;C work.

**Test plan**

Do `./r2r -j 1 tools /r2` 5 times and make sure it's `Total: 13` every time. Do it again according to levels of patience and risk tolerance.

**Closing issues**

See description.

### DIFF
--- a/test/src/r2r.v
+++ b/test/src/r2r.v
@@ -514,7 +514,6 @@ fn handle_control_c() {
 }
 
 fn (r2r mut R2R) run_cmd_test(test R2RCmdTest) {
-	r2r.wg.add(1)
 	time_start := time.ticks()
 	tmp_dir := mktmpdir('')
 	tmp_script := filepath.join(tmp_dir,'script.r2')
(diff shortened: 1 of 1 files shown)


radare:
i dont see why those changes should be breaking ^C :? but lgtm

kazarmy:
i think ^C in V is only affecting the running process ... don't know how node is handling it

radare:
yes, and radare.r2 is not able to capture the sigint from the child right now, so.... json, asm tests work as expected because no spawn is involved, but in cmd and fuzz ones ^C only works when not in a spawn, just press ^C for a while and it will eventually stop before 1-2s, not perfect, but better than being unable to break

kazarmy:
i actually use `/usr/bin/kill r2r` in a different console ... faster than it looks since i only need to type it once and then just tap the up arrow key every time i need it

> Do ./r2r -j 1 tools /r2 5 times and make sure it's Total: 13 every time. Do it again according to levels of patience and risk tolerance.

Should calculate a probability based on the binomial distribution. Might be interesting.

PR #15966, opened 2020-02-13
Use state struct and start handling cmd_substition_arg in tree-sitter

ret2libc:
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible). This is still a work in progress and there are many things that need to be improved. There is however a travis job that uses the new parser, where you can see what works and what fails as of right now.
- [ ] I've updated the documentation and the [radare2 book](../../radare2book) with the relevant information (if needed). Not needed.

**Detailed description**
This PR adds support for command substitution in the new shell parser implemented with tree-sitter. We agreed on allowing command substitutions (backticks and `$( ... )`) only where an argument is expected. The expansion of the inner command can at most generate new arguments, but it cannot change the structure of the command (e.g. `pd $("?e 10;pd 10")` will not execute two times the `pd` command, but it will be interpreted as `pd "10;pd 10").

**Test plan**

After setting `e cfg.newshell=true` you can try things like:
```
> pd $(?e 10)
> om `?e 4` 0x1000$(?e 00 $(?v 100)) 0x5555
> pd 10 @ $(afl~main[0]:2)
```

**Additional information**
As usual, this is far from complete. I need to improve how arguments are handled and there are going to be some differences between old parser and new one with regards to argument wrapped in `"`.

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -4409,6 +4409,30 @@ R_API void run_pending_anal(RCore *core) {
 static int run_cmd_depth(RCore *core, char *cmd);
 
 #if USE_TREESITTER
+
+struct tsr2cmd_state {
+	TSParser *parser;
+	RCore *core;
+	char *input;
+	TSTree *tree;
+	bool log;
+};
+
+struct tsr2cmd_edit {
+	char *new_text;
+	char *old_text;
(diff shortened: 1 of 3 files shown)


radare:
awesome

PR #15971, opened 2020-02-13
new parser: fix multiple words in grep and add support for > $alias

ret2libc:

**Detailed description**

This PR adds support for redirect to alias (e.g. `om* > $myalias`) in new parser and fix grep with multiple words (e.g. `pd 10~xor~eax`)

**Test plan**

Some of the regression tests use these features, so they should now pass with the new parser.

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -5024,7 +5024,10 @@ DEFINE_HANDLE_TS_FCN(grep_command) {
 	char *specifier_str = ts_node_sub_string (specifier, state->input);
 	bool res = handle_ts_command (state, command);
 	R_LOG_DEBUG ("grep_command specifier: '%s'\n", specifier_str);
-	char *specifier_str_processed = r_cons_grep_strip (specifier_str, "`");
+	RStrBuf *sb = r_strbuf_new (specifier_str);
+	r_strbuf_prepend (sb, "~");
+	char *specifier_str_processed = r_cons_grep_strip (r_strbuf_get (sb), "`");
+	r_strbuf_free (sb);
 	R_LOG_DEBUG ("grep_command processed specifier: '%s'\n", specifier_str_processed);
 	r_cons_grep_process (specifier_str_processed);
 	free (specifier_str);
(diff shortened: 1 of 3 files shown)


libr/core/cmd.c
```diff
@@ -4994,8 +5023,14 @@ DEFINE_HANDLE_TS_FCN(grep_command) {
 	TSNode specifier = ts_node_child_by_field_name (node, "specifier", strlen ("specifier"));
 	char *specifier_str = ts_node_sub_string (specifier, state->input);
 	bool res = handle_ts_command (state, command);
-	// FIXME: this does not work for nested grep commands
-	r_cons_grep_process (specifier_str);
+	R_LOG_DEBUG ("grep_command specifier: '%s'\n", specifier_str);
+	RStrBuf *sb = r_strbuf_new (specifier_str);
+	r_strbuf_prepend (sb, "~");
+	char *specifier_str_processed = r_cons_grep_strip (r_strbuf_get (sb), "`");
```
radare: Toolongvarname

libr/core/cmd.c
```diff
@@ -4631,7 +4633,13 @@ static char *do_handle_ts_unescape_arg(struct tsr2cmd_state *state, TSNode arg)
 		return do_handle_ts_unescape_arg (state, ts_node_named_child (arg, 0));
 	} else if (is_ts_arg_identifier (arg)) {
 		return unescape_arg (state, arg, SPECIAL_CHARS_REGULAR);
-	} else if (is_ts_quoted_arg (arg)) {
+	} else if (is_ts_single_quoted_arg (arg)) {
+		char *c = unescape_arg (state, arg, SPECIAL_CHARS_SINGLE_QUOTED);
+		c[strlen (c) - 1] = '\0';
```
radare: If len ==0 oobread

radare: Err oobwrite

ret2libc: yeah let me re-write this part. len == 0 can't actually happen though because being a quoted argument, `c` contains at least `''`.

ret2libc:
@radare thanks for review. I have fixed those things. Let's see what travis said. Step by step we should move to all green tests.

Only thing that worries me so far is this:
old parser: `?e "Hello World"` is just interpreted as `"Hello World"`
new parser: `?e "Hello World"` , `"Hello World"` is interpreted as a single argument, so `argv[1] = Hello World` (without the wrapping `"`), which results in some commands returning slightly different content. I think the new parser is better as it resembles more bash, but for now I may try to find a workaround for this and make the new parser work like the old one. However, once commands will accept `argv` instead of just `input`, I think it's better to use the new behaviour.

PR #15984, opened 2020-02-17
RFC: WIP: Fix #15944 - Cache last fcn args/vars in asm.var.sub ##disasm

### LINKED ISSUE #15944: Varsub is very slow
### Work environment

| OS/arch/bits (mandatory)                             | Archlinux x86 64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.3.0-git 23793 @ linux-x86-64 git.4.2.1-24-gbe10bbe7b commit: be10bbe7b08ebada8eee41bd3844ce383946e3ec build: 2020-02-09__10:57:34

### Expected behavior
Printing disassembly with varsub enabled doesn't take significantly longer than without it.

### Actual behavior

On large function varsub can make commands like agfj 10x slower.

- Open medium size executable like cmake
- `aaa` less analysis might be sufficient, I don't know
- 'aflt nbbs/sort` find biggest functions
- `e asm.var.sub=true` enable varsub
- `?t agfj function_name`  write down the time
- `e asm.var.sub=false` enable varsub
- `?t agfj function_name` compare with previous time

In my case with  with cmake and function containing ~700 bbs and 200 local vars varsub made it almost 10x slower -  2.24s vs 0.28s.

### Observations about code and potential improvements

* List of sp and bp variables gets created for each asm line
* To find the variables which might be used in instruction varsub iterates over all variables instead of using infromation from disassembler
* Variable repalcement is done as string replacement and guessing how disassembler printed reference to the variable

Potential improvements starting with with best and less likely to be done in near time, ending with simpler solutions that could make it less bad until better fix.

* Disassembler returns asm text with annotations describing which segment of text is instruction name, argument 1, argument 2, register name, expression using a register. Disassembler already has this information but currently there is no mapping between it and the text. This would reduce amount of architecture specific assembly text parsing code. 
* Use the structured instruction arguments information from disassembler so that only relevant variables need to be searched in instruction text
* store information about variables in a structure that allows efficient lookup from instruction description without iterating over all of them
* Basic variable list cache so that it doesn't get created from scratch for each instruction

radare:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/parse/p/parse_x86_pseudo.c
+++ b/libr/parse/p/parse_x86_pseudo.c
@@ -8,109 +8,122 @@
 #include <r_util.h>
 #include <r_anal.h>
 #include <r_parse.h>
+
+#define MAXPSEUDOOPS 10
+// ?t pd 50000 >/dev/null # 3.2/3.8  #  2.9/3.5
+#define CACHE_VARS 1
+// if enabled, i cant see any speedup benefit
+#define USE_HT 0
+
+struct x86op {
+	char *op;
+	char *str;
+	int args[MAXPSEUDOOPS];  // XXX can't use flex arrays, all unused will be 0
+};
+
+struct x86op ops[] = {
+	// sort by frequency to optimize search
+	{ "jmp",  "goto #", {1}},
+	{ "mov",  "# = #", {1, 2}},
+	{ "pop",  "pop #", {1}},
+	{ "push", "push #", {1}},
+	{ "test", "var = # & #", {1, 2}},
+	{ "call", "# ()", {1}},
+	{ "lea",  "# = #", {1, 2}},
+	{ "nop",  "", {0}},
+	{ "sub",  "# -= #", {1, 2}},
+	{ "xor",  "# ^= #", {1, 2}},
+	{ "ret",  "return", {0}},
+
+	// ---
+	{ "adc",  "# += #", {1, 2}},
+	{ "add",  "# += #", {1, 2}},
+	{ "and",  "# &= #", {1, 2}},
+	{ "cmove", "if (!var) # = #", {1, 2}},
+	{ "cmovl","if (var < 0) # = #", {1, 2}},
+	{ "cmp", "var = # - #", {1, 2}},
+	{ "cmpsq", "var = # - #", {1, 2}},
+	{ "cmpsb", "while (CX != 0) { var = *(DS*16 + SI) - *(ES*16 + DI); SI++; DI++; CX--; if (!var) break; }", {0}},
+	{ "cmpsw", "while (CX != 0) { var = *(DS*16 + SI) - *(ES*16 + DI); SI+=4; DI+=4; CX--; if (!var) break; }", {0}},
+	{ "dec",  "#--", {1}},
+	{ "div",  "# /= #", {1, 2}},
+	{ "fabs",  "abs(#)", {1}},
+	{ "fadd",  "# = # + #", {1, 1, 2}},
+	{ "fcomp",  "var = # - #", {1, 2}},
+	{ "fcos",  "# = cos(#)", {1, 1}},
+	{ "fdiv",  "# = # / #", {1, 1, 2}},
+	{ "fiadd",  "# = # / #", {1, 1, 2}},
+	{ "ficom",  "var = # - #", {1, 2}},
+	{ "fidiv",  "# = # / #", {1, 1, 2}},
+	{ "fidiv",  "# = # * #", {1, 1, 2}},
+	{ "fisub",  "# = # - #", {1, 1, 2}},
+	{ "fnul",  "# = # * #", {1, 1, 2}},
+	{ "fnop",  " ", {0}},
+	{ "frndint",  "# = (int) #", {1, 1}},
+	{ "fsin",  "# = sin(#)", {1, 1}},
+	{ "fsqrt",  "# = sqrt(#)", {1, 1}},
+	{ "fsub",  "# = # - #", {1, 1, 2}},
+	{ "fxch",  "#,# = #,#", {1, 2, 2, 1}},
+	{ "idiv",  "# /= #", {1, 2}},
+	{ "imul",  "# = # * #", {1, 2, 3}},
+	{ "in",   "# = io[#]", {1, 2}},
+	{ "inc",  "#++", {1}},
+	{ "ja", "if (((unsigned) var) > 0) goto #", {1}},
+	{ "jb", "if (((unsigned) var) < 0) goto #", {1}},
+	{ "jbe", "if (((unsigned) var) <= 0) goto #", {1}},
+	{ "je", "if (!var) goto #", {1}},
+	{ "jg", "if (var > 0) goto #", {1}},
+	{ "jge", "if (var >= 0) goto #", {1}},
+	{ "jle", "if (var <= 0) goto #", {1}},
+
+	{ "jne", "if (var) goto #", {1}},
+	{ "movq",  "# = #", {1, 2}},
+	{ "movaps",  "# = #", {1, 2}},
+	{ "movups",  "# = #", {1, 2}},
+	{ "movsd",  "# = #", {1, 2}},
+	{ "movsx","# = #", {1, 2}},
+	{ "movsxd","# = #", {1, 2}},
+	{ "movzx", "# = #", {1, 2}},
+	{ "movntdq", "# = #", {1, 2}},
+	{ "movnti", "# = #", {1, 2}},
+	{ "movntpd", "# = #", {1, 2}},
+	{ "pcmpeqb", "# == #", {1, 2}},
+
+	{ "movdqu", "# = #", {1, 2}},
+	{ "movdqa", "# = #", {1, 2}},
+	{ "pextrb", "# = (byte) # [#]", {1, 2, 3}},
+	{ "palignr", "# = # align #", {1, 2, 3}},
+	{ "pxor", "# ^= #", {1, 2}},
+	{ "xorps", "# ^= #", {1, 2}},
+	{ "mul",  "# = # * #", {1, 2, 3}},
+	{ "mulss",  "# = # * #", {1, 2, 3}},
+	{ "neg",  "# ~= #", {1, 1}},
+	{ "not",  "# = !#", {1, 1}},
+	{ "or",   "# |= #", {1, 2}},
+	{ "out",  "io[#] = #", {1, 2}},
+	{ "sal",  "# <<= #", {1, 2}},
+	{ "sar",  "# >>= #", {1, 2}},
+	{ "sete",  "# = e", {1}},
+	{ "setne",  "# = ne", {1}},
+	{ "shl",  "# <<<= #", {1, 2}},
+	{ "shld",  "# <<<= #", {1, 2}},
+	{ "sbb",  "# = # - #", {1, 1, 2}},
+	{ "shr",  "# >>>= #", {1, 2}},
+	{ "shlr",  "# >>>= #", {1, 2}},
+	//{ "strd",  "# = # - #", {1, 2, 3}},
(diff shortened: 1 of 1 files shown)


radare:
cc #15944

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -358,12 +364,58 @@ static inline void mk_reg_str(const char *regname, int delta, bool sign, bool at
...
+	__args (p, NULL, -1);
+	return 0;
+}
+
+static bool varsub(RParse *p, RAnalFunction *f, ut64 addr, int oplen, char *data, char *str, int len) {
 	RAnal *anal = p->analb.anal;
 	RAnalVar *bparg, *sparg;
 	RListIter *bpargiter, *spiter;
 	char oldstr[64], newstr[64];
+	if (!p) {
```
ret2libc: use `r_return_val_if_fail (p, false);`

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -358,12 +364,58 @@ static inline void mk_reg_str(const char *regname, int delta, bool sign, bool at
...
 }
 
-static bool varsub (RParse *p, RAnalFunction *f, ut64 addr, int oplen, char *data, char *str, int len) {
-	RList *bpargs, *spargs;
+static const RList *__args(RParse *p, RAnalFunction *f, int type) {
+#if CACHE_VARS
+	static RAnalFunction *old_f = NULL;
+	static RList *__cache_spargs_list = NULL;
+	static RList *__cache_bpargs_list = NULL;
+	if (type == -1 && !p && !f) {
```
ret2libc: maybe you meant `||` ? I'm not sure this condition can ever be met otherwise, as at https://github.com/radareorg/radare2/pull/15984/files#diff-18af017c7f2f2d244eb96f6f6daa48efR407 p is probably *not null*.

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -358,12 +364,58 @@ static inline void mk_reg_str(const char *regname, int delta, bool sign, bool at
...
+		// free to issue mamleak
+		r_list_free (__cache_bpargs_list);
+		r_list_free (__cache_spargs_list);
+		old_f = NULL;
+		return NULL;
+	}
+	r_return_val_if_fail (p, NULL);
+	if (!f || !f->anal) {
```
ret2libc: do a `r_return_val_if_fail (f && f->anal, NULL)`. If you are calling `__args` without a function or with broken anal, you have already screwed things up a while ago.

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -358,12 +364,58 @@ static inline void mk_reg_str(const char *regname, int delta, bool sign, bool at
 	r_strbuf_free (sb);
 }
 
-static bool varsub (RParse *p, RAnalFunction *f, ut64 addr, int oplen, char *data, char *str, int len) {
-	RList *bpargs, *spargs;
+static const RList *__args(RParse *p, RAnalFunction *f, int type) {
```
ret2libc: add also a `r_return_val_if_fail` to check that type is either 's' or 'b'. Any other value should be refused.

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -358,12 +364,58 @@ static inline void mk_reg_str(const char *regname, int delta, bool sign, bool at
...
+#if CACHE_VARS
+	static RAnalFunction *old_f = NULL;
+	static RList *__cache_spargs_list = NULL;
```
ret2libc: aren't names with `__` prefix reserved? I think we should use names without it.

libr/parse/p/parse_x86_pseudo.c
ret2libc: actually, instead of using 'b', 's', etc. you should use the defines `R_ANAL_VAR_KIND_BPV`, `R_ANAL_VAR_KIND_SPV`, etc.

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -358,12 +364,58 @@ static inline void mk_reg_str(const char *regname, int delta, bool sign, bool at
...
+	if (!f || !f->anal) {
+		return NULL;
+	}
+	if (f == old_f) {
+		if (type == 'b') {
+			return __cache_bpargs_list;
+		}
+		return __cache_spargs_list;
+	}
+	RList *l = p->varlist (f->anal, f, 'b');
```
ret2libc: no need for an additional variable `l`. Just do:
```C
r_list_free (__cache_bpargs_list);
__cache_bpargs_list = p->varlist (f->anal, f, 'b');`
```

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -358,12 +364,58 @@ static inline void mk_reg_str(const char *regname, int delta, bool sign, bool at
...
+	r_list_free (__cache_bpargs_list);
+	__cache_bpargs_list = l;
+	l = p->varlist (f->anal, f, 's');
```
ret2libc: same

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -358,12 +364,58 @@ static inline void mk_reg_str(const char *regname, int delta, bool sign, bool at
...
+		r_list_free (__cache_bpargs_list);
+		r_list_free (__cache_spargs_list);
+		old_f = NULL;
+		return NULL;
+	}
+	r_return_val_if_fail (p, NULL);
+	if (!f || !f->anal) {
+		return NULL;
+	}
+	if (f == old_f) {
```
ret2libc: Do:
```C
if (f != old_f) {
   ... get new cached lists ...
}
if (type == 'b') {
     return __cache_bpargs_list;
}
return __cache_spargs_list;
```
So the code to return the right list is not duplicated.

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -358,12 +364,58 @@ static inline void mk_reg_str(const char *regname, int delta, bool sign, bool at
 	r_strbuf_free (sb);
 }
 
-static bool varsub (RParse *p, RAnalFunction *f, ut64 addr, int oplen, char *data, char *str, int len) {
-	RList *bpargs, *spargs;
+static const RList *__args(RParse *p, RAnalFunction *f, int type) {
+#if CACHE_VARS
```
ret2libc: if CACHE_VARS brings any substantial improvement, please just keep the CACHE_VARS version and remove the rest of the code before PR's approval.

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -512,8 +569,8 @@ static bool varsub (RParse *p, RAnalFunction *f, ut64 addr, int oplen, char *dat
 	}
 	free (tstr);
 	free (ireg);
-	r_list_free (spargs);
-	r_list_free (bpargs);
+	//r_list_free (spargs);
```
ret2libc: just remove

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -8,130 +8,159 @@
 #include <r_util.h>
 #include <r_anal.h>
 #include <r_parse.h>
+
+#define MAXPSEUDOOPS 10
+// ?t pd 50000 >/dev/null # 3.2/3.8  #  2.9/3.5
+#define CACHE_VARS 1
+// if enabled, i cant see any speedup benefit
```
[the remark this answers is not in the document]
ret2libc: As for the code, I prefer the Ht version as it's clearer what it does.

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -8,130 +8,159 @@
...
-	};
 
 	if (argc > 2 && !strcmp (argv[0], "xor")) {
 		if (!strcmp (argv[1], argv[2])) {
 			argv[0] = "mov";
 			argv[2] = "0";
 		}
 	}
+#if USE_HT
+	static HtPP *ht = NULL;
```
thestr4ng3r: No static please.

radare:
Yep this a very bad cache method but just trying to approach a possible optimization after finsing out the bottleneck. This rlist should not be that slow to be fetched and as long as it csn change at any time i was expecting it to be red.

libr/parse/p/parse_x86_pseudo.c
radare: yeah, totally aware of this. the idea behind this pr was to open a discussion on how this optimization affects performance and where to cache/hold this info instead of creating it everytime we request it.

libr/parse/p/parse_x86_pseudo.c
```diff
@@ -8,130 +8,159 @@
...
+#define MAXPSEUDOOPS 10
+// ?t pd 50000 >/dev/null # 3.2/3.8  #  2.9/3.5
+#define CACHE_VARS 1
+// if enabled, i cant see any speedup benefit
+#define USE_HT 0
+
+struct x86op {
+	char *op;
+	char *str;
+	int args[MAXPSEUDOOPS];  // XXX can't use flex arrays, all unused will be 0
```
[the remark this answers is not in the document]
radare: that comment is not from me, but i assume its refereing to use a dynamically allocated array or rlist instead of fixed size

libr/parse/p/parse_x86_pseudo.c
radare: i would prefer to focus on the cache in this PR, this code have a lot of things to improve

ret2libc: I *think* the cache method is ok ebough, but you should tell us what the benchmark says.

radare:
This makes pd 3 times faster

radare:
The cache is not fine because while analyzing the function the vars-args can be added and theres nothing yhat invalidates this. Also its leaking because its not freed

ret2libc:
I had another look at this and... Why don't we have the variables directly in the RAnalFunction?? The slowness comes from the fact that `var_generate_list` re-creates the list of variables from sdb every time it is called.

As I always say, sdb is not the right place to store things. sdb should be used only as a way to serialie/deserialize objects.

In this case, having variables directly stored as objects inside RAnalFunction would make the `varlist` operation a very quick one (as it would just return something like `fcn->variables`). You wouldn't even have the problem of stale cache, etc. On the other hand, a solution like this one probably requires changes in a bit of places.

thestr4ng3r:
@ret2libc agreed

XVilka:
Better to target for the next release.

radare:
imho fcn->args should be taken from the fcn->signature.. the fcn->vars should be specific to the function and stored as linked list, not in Sdb

radare:
do you want to do this change @thestr4ng3r ?

XVilka:
What about using RVector instead of the RList - this kind of cache fits perfectly the case?

radare:
No. We shouldnt cache at all. The current storage needs traversing it too much and its innefficient

thestr4ng3r:
> imho fcn->args should be taken from the fcn->signature.. the fcn->vars should be specific to the function and stored as linked list, not in Sdb

this isn't directly related to this pr, is it?

radare:
this PR will be closed. because the correct solution is to not use sdb to hold the locals and not hold vars in the function itself, but in the function signature, fcn needs to add .>signame... i just kept the pr open as a remindner

radare:
anyone up for a proper refactor of all the var stuff for 4.4?

I dont want to include this speedup because caching it its a hack

thestr4ng3r:
I'm up for refactor but without an eta

radare:
4.4 is april 13th. I think its feasible. But this is a huge speedup in pd and in analysis and i would be happy to help on doing the change in case we find a way to slowly do the change or split the subtasks

Issue #921, opened 2014-05-12
META - Portable Executable


Maijin:
# All Portable Executables

## Related

* [ ] Doc - TBD
* Tests - https://github.com/radare/radare2-regressions/tree/master/t.formats/pe
* Commands involved - N/A
* [ ] BlogPost - TBD
* .c involved:
    * https://github.com/radare/radare2/blob/master/libr/bin/p/bin_pe.c

## General Information

- [ ] Add DLL Characteristics in `iI`, `iIj` (Already parsed in SDB) #6785 Or in `iH` , `iHj` https://github.com/radare/radare2/issues/6759
- [ ] VTABLE detection for MSVC https://github.com/radare/radare2/issues/6851
- [x] Add authentihash and authentihash check in `it` and `itj`

## R_BIN Enhancement

 radare2-regressions files that are not properly parsed by r2 currently:
- [ ] d_resource.dll
- [ ] d_tiny.dll
- [ ] dosZMXP.exe
- [ ] exe2pe.exe
- [ ] no0code.exe

## Analysis

- [ ]  detect and rename the elements/structures accordingly. (See all the +ms_exc.registration.TryLevel etc...) see http://pastebin.com/aLrkiwuM
[reversing-demo.zip](https://github.com/radare/radare2/files/801155/reversing-demo.zip)

## Resources

- [ ] Add/remove/modify/dump Resources `rabin2 -O` && `iO`

## Import/Export Table

- [ ] Ability to move the import table `rabin2 -O` && `iO`
- [ ] Ability to add/remove import/export library (with prior checks to see if possible) `rabin2 -O` && `iO`
- [ ] Ability to add/remove function to a library (with prior checks to see if possible) `rabin2 -O` && `iO`
- [ ] Ability to dump a process and reconstruct/manipulate IAT (ollydump + IMPrec tool / x64dbg) (https://github.com/OALabs/PyIATRebuild/blob/master/pyiatrebuild.py)
- [ ] Add PE unmapper https://github.com/hasherezade/malware_analysis/tree/master/pe_unmapper {as a r2e module}
- [ ] Implement the Windows API import recovery functionality shown in the apiscout: https://github.com/danielplohmann/apiscout -- (API import recovery: http://byte-atlas.blogspot.de/2017/04/apiscout.html) (https://github.com/danielplohmann/apiscout/issues/1)
- [ ] Add imphash https://github.com/Neo23x0/ImpHash-Generator  {As a r2e module}

## Rich Header

[...]

PR #15987, opened 2020-02-17
Add Certificate Table parsing and Authentihash support for PE

abcSup:

**Detailed description**
Add Authentihash and Authentihash check support for Portable Executable. Currently, the Authentihash information are stored in the `pe` sdb.

**Missing steps**
- [x] A signed binary with valid Authentihash to test true case. There seems to be only `signature.exe` in testbins which has an invalid Authentihash.
- [x] Include Authentihash in `it` output (?)

Imo, Authentihash information should be displayed at somewhere specific to the file format. For now, `it` seems to generate general hashes (md5, sha1, sha256). Adding Authentihash into the command will complicate the code (checking if `rclass == pe`, get authentihash with `sdb_querys`, output directly) while the existing code is using a `RList` of `RBinHash`, which makes me wonder if there is a cleaner way to do this. For Authentihash check, I thought about adding it into `RBinInfo`, but that makes the structure less general, so I am looking forward to some feedbacks.

**Closing issues**
#921

### DIFF
--- a/libr/bin/format/pe/pe.c
+++ b/libr/bin/format/pe/pe.c
@@ -3,6 +3,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <r_hash.h>
 #include <r_types.h>
 #include <r_util.h>
 #include "pe.h"
@@ -836,6 +837,66 @@ int PE_(bin_pe_get_actual_checksum)(struct PE_(r_bin_pe_obj_t)* bin) {
 	// add filesize
 	computed_cs += bin->size;
 	return computed_cs;
+}
+
+static char* PE_(bin_pe_get_claimed_authentihash)(struct PE_(r_bin_pe_obj_t)* bin) {
+	if (!bin->spcinfo) {
+		return NULL;
+	}
+	RASN1Binary *digest = bin->spcinfo->messageDigest.digest;
+	return r_hex_bin2strdup (digest->binary, digest->length);
+}
+
+static char* PE_(bin_pe_get_actual_authentihash)(struct PE_(r_bin_pe_obj_t)* bin) {
+	if (!bin->spcinfo) {
+		return NULL;
+	}
+
+	const char *hashtype = bin->spcinfo->messageDigest.digestAlgorithm.algorithm->string;
+	ut64 algobit = r_hash_name_to_bits (hashtype);
+	if (!(algobit & (R_HASH_MD5 | R_HASH_SHA1))) {
+		eprintf ("Authenticode only supports md5, sha1.\n");
+		return NULL;
+	}
+
+	ut32 checksum_paddr = bin->nt_header_offset + 4 + sizeof (PE_(image_file_header)) + 0x40;
+	ut32 security_entry_offset =  bin->nt_header_offset + sizeof (PE_(image_nt_headers)) - 96;
+	PE_(image_data_directory) *data_dir_security = &bin->data_directory[PE_IMAGE_DIRECTORY_ENTRY_SECURITY];
+	PE_DWord security_dir_offset = data_dir_security->VirtualAddress;
+	ut32 security_dir_size = data_dir_security->Size;
+
+	RBuffer *buf = r_buf_new ();
+	r_buf_append_buf_slice (buf, bin->b, 0, checksum_paddr);
+	r_buf_append_buf_slice (buf, bin->b,
+		checksum_paddr + 4,
+		security_entry_offset - checksum_paddr - 4);
+	r_buf_append_buf_slice (buf, bin->b,
+		security_entry_offset + 8,
+		security_dir_offset - security_entry_offset - 8);
+	r_buf_append_buf_slice (buf, bin->b,
+		security_dir_offset + security_dir_size,
+		r_buf_size (bin->b) - security_dir_offset - security_dir_size);
+
+	ut64 len;
+	const ut8 *data = r_buf_data (buf, &len);
+	char *hashstr = NULL;
+	RHash *ctx = r_hash_new (true, algobit);
+	r_hash_do_begin (ctx, algobit);
+	int digest_size = r_hash_calculate (ctx, algobit, data, len);
+	r_hash_do_end (ctx, algobit);
+	hashstr = r_hex_bin2strdup (ctx->digest, digest_size);
+
+	r_buf_free (buf);
+	r_hash_free (ctx);
+	return hashstr;
+}
+
+const char* PE_(bin_pe_get_authentihash)(struct PE_(r_bin_pe_obj_t)* bin) {
+	return bin->authentihash;
+}
+
+int PE_(bin_pe_is_authhash_valid)(struct PE_(r_bin_pe_obj_t)* bin) {
+	return bin->is_authhash_valid;
 }
 
 static void computeOverlayOffset(ut64 offset, ut64 size, ut64 file_size, ut64* largest_offset, ut64* largest_size) {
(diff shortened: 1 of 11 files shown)


libr/bin/format/pe/pe.c
```diff
@@ -838,6 +839,69 @@ int PE_(bin_pe_get_actual_checksum)(struct PE_(r_bin_pe_obj_t)* bin) {
...
+	}
+
+	hashalgo = bin->spcinfo->messageDigest.digestAlgorithm.algorithm->string;
+	checksum_paddr = bin->nt_header_offset + 4 + sizeof(PE_(image_file_header)) + 0x40;
+	security_entry_offset =  bin->nt_header_offset + sizeof(PE_(image_nt_headers)) - 96;
+	data_dir_security = &bin->data_directory[PE_IMAGE_DIRECTORY_ENTRY_SECURITY];
+	security_dir_offset = bin_pe_rva_to_paddr (bin, data_dir_security->VirtualAddress);
+	security_dir_size = data_dir_security->Size;
+
+	buf = r_buf_new ();
```
radare: define and assign in the same line, it is more readable (RBuffer *buf = r_buf_new())

libr/bin/format/pe/pe.c
```diff
@@ -838,6 +839,69 @@ int PE_(bin_pe_get_actual_checksum)(struct PE_(r_bin_pe_obj_t)* bin) {
...
+	ut64 len;
+	PE_(image_data_directory) * data_dir_security;
+	PE_DWord security_dir_offset;
+	RBuffer *buf;
+	RHash *ctx;
+	if (!bin->spcinfo) {
+		return NULL;
+	}
+
+	hashalgo = bin->spcinfo->messageDigest.digestAlgorithm.algorithm->string;
```
radare: same here and all the other variables

libr/bin/format/pe/pe.c
```diff
@@ -838,6 +839,69 @@ int PE_(bin_pe_get_actual_checksum)(struct PE_(r_bin_pe_obj_t)* bin) {
...
+		checksum_paddr + 4,
+		security_entry_offset - checksum_paddr - 4);
+	r_buf_append_buf_slice (buf, bin->b, 
+		security_entry_offset + 8,
+		security_dir_offset - security_entry_offset - 8);
+	r_buf_append_buf_slice (buf, bin->b,
+		security_dir_offset + security_dir_size, 
+		r_buf_size (bin->b) - security_dir_offset - security_dir_size);
+	block = r_buf_data (buf, &len);
+	if (!strcmp (hashalgo, "md5")) {
```
radare: that's a bit hardcoded and it will be good to support multiple hashes here, not just sha1 and md5, i think there's an api in RHash to get the hash bits from the string name

libr/bin/format/pe/pe.c
```diff
@@ -838,6 +839,69 @@ int PE_(bin_pe_get_actual_checksum)(struct PE_(r_bin_pe_obj_t)* bin) {
...
+	PE_DWord security_dir_offset;
+	RBuffer *buf;
+	RHash *ctx;
+	if (!bin->spcinfo) {
+		return NULL;
+	}
+
+	hashalgo = bin->spcinfo->messageDigest.digestAlgorithm.algorithm->string;
+	checksum_paddr = bin->nt_header_offset + 4 + sizeof(PE_(image_file_header)) + 0x40;
+	security_entry_offset =  bin->nt_header_offset + sizeof(PE_(image_nt_headers)) - 96;
```
radare: missing space before (

libr/bin/format/pe/pe.c
```diff
@@ -838,6 +839,69 @@ int PE_(bin_pe_get_actual_checksum)(struct PE_(r_bin_pe_obj_t)* bin) {
...
+		ctx = r_hash_new (true, R_HASH_SHA1);
+		hash = r_hash_do_sha1 (ctx, block, len);
+		hashstr = r_hex_bin2strdup (hash, R_HASH_SIZE_SHA1);
+	}
+	r_hash_free (ctx);
+	r_buf_free (buf);
+	return hashstr;
+}
+
+char* PE_(bin_pe_get_authentihash)(struct PE_(r_bin_pe_obj_t)* bin) {
```
radare: maybe better to make those functions return const char *

libr/bin/format/pe/pe.c
```diff
@@ -2736,37 +2800,93 @@ R_API void PE_(bin_pe_parse_resource)(struct PE_(r_bin_pe_obj_t) *bin) {
...
+	offset = paddr;
+	while (offset < paddr + size) {
+		Pe_certificate** tmp = (Pe_certificate **) realloc (security_directory->certificates, (security_directory->length + 1) * sizeof(cert));
+		if (!tmp) {
+			r_sys_perror ("realloc (certificates)");
+			return false;
+		}
+		cert = R_NEW0 (Pe_certificate);
+		if (!cert) {
+			r_sys_perror ("R_NEW0 (security_directory)");
```
radare: no need to eprintf the error in every alloc function unless its something that can fail on normal situations

ret2libc:
> **Missing steps**
> 
>     * [ ]  A signed binary with valid Authentihash to test true case. There seems to be only `signature.exe` in testbins which has an invalid Authentihash.

I do not use Windows. Do you have any of those binary on your system? If so, you can upload it to https://github.com/radareorg/radare2-testbins .

> 
>     * [ ]  Include Authentihash in `it` output (?)
> 
> 
> Imo, Authentihash information should be displayed at somewhere specific to the file format. For now, `it` seems to generate general hashes (md5, sha1, sha256). Adding Authentihash into the command will complicate the code (checking if `rclass == pe`, get authentihash with `sdb_querys`, output directly) while the existing code is using a `RList` of `RBinHash`, which makes me wonder if there is a cleaner way to do this. For Authentihash check, I thought about adding it into `RBinInfo`, but that makes the structure less general, so I am looking forward to some feedbacks.
> 

Maybe you can fix a bit the code around the `RBinPlugin->info` method. In `bin_pe.inc:info()` you could compute the authentihash, create a RBinHash structure and put it in `info->file_hashes` (you should adapt a bit the code in `bfile.c:r_bin_file_hash()` to consider the case when `info->file_hashes` is already initialized, or maybe you should always initialize `info->file_hashes` somewhere else). By doing so the hash will be shown in `it` and any file format can easily add its own specific hash.

libr/bin/format/pe/pe.c
abcSup: At first, I was trying to shorten that line to not exceed the 80 characters style, but code readability is more important.

libr/bin/format/pe/pe.c
abcSup: Gotcha! I missed that because `clang-format` keeps suggesting space after `PE_` instead, so I kinda skimmed through those suggestions.

libr/bin/format/pe/pe.c
abcSup: I am wondering when to and not to return `const char *`, because I saw some `TODO` comment suggesting to change other API to `const char *` as well, but then freeing those memory later will require re-casting them to `(char *)` which seems like code smell.

libr/bin/format/pe/pe.c
abcSup: Thanks for the info! I will refactor it to use the RHash api, but still only support `md5` and `sha1` since the Authenticode only supports them.

abcSup:
> I do not use Windows. Do you have any of those binary on your system? If so, you can upload it to https://github.com/radareorg/radare2-testbins .

Currently, I am testing it on commercial programs. I will look it up how to sign one without a proper cert.

> Maybe you can fix a bit the code around the RBinPlugin->info method. In bin_pe.inc:info() you could compute the authentihash, create a RBinHash structure and put it in info->file_hashes (you should adapt a bit the code in bfile.c:r_bin_file_hash() to consider the case when info->file_hashes is already initialized, or maybe you should always initialize info->file_hashes somewhere else). By doing so the hash will be shown in it and any file format can easily add its own specific hash.

Thanks for the green light, I will proceed to refactor `bfile.c:r_bin_file_hash()` to make it more general.

libr/bin/format/pe/pe.c
abcSup: Guess I was wrong about that, since `PE_(bin_pe_get_authentihash)` is just a public interface, and the underlying `authentihash` property is still `char *`, so freeing it won't require re-casting.

radare:
thanks! i think we are good to remove the WIP after you do this refactor so we will be ready to merge

XVilka:
Binary was merged, please update the PR and we are good to go.

abcSup:
Appveyor and one of the Travis tests passed, but others failed because of:
```
v/v install radare.r2
Skipping module "radare.r2", since it is missing name or url information.
Makefile:19: recipe for target 'v/v' failed
make[1]: Leaving directory '/home/travis/build/radareorg/radare2/test/src'
make[1]: *** [v/v] Error 1
make[1]: Target 'all' not remade because of errors.
make: *** [src/r2r] Error 2
Makefile:14: recipe for target 'src/r2r' failed
make: Target 'all' not remade because of errors.
+ exit 1
```
I assume that if we rebuild the tests, it should be fine.

libr/main/radare2.c
```diff
@@ -1221,8 +1221,15 @@ R_API int r_main_radare2(int argc, char **argv) {
 			if (compute_hashes && iod) {
 				// TODO: recall with limit=0 ?
 				ut64 limit = r_config_get_i (r->config, "bin.hashlimit");
-				(void)r_bin_file_hash (r->bin, limit, iod->name, NULL);
-				//eprintf ("WARNING: File hash not calculated\n");
+				RList *file_hashes = r_bin_file_hash (r->bin, limit);
+				if (file_hashes) {
+					if (r->bin->cur->o->info->file_hashes) {
+						r_list_join (r->bin->cur->o->info->file_hashes, file_hashes);
+						free (file_hashes);
```
ret2libc: this probably causes a Use After Free, because it also frees each element in file_hashes, which are now owned by info->file_hashes.

libr/core/cmd_info.c
```diff
@@ -603,66 +603,52 @@ static int cmd_info(void *data, const char *input) {
...
-				}
 				if (!info) {
-					info = r_bin_get_info (core->bin);
+					eprintf ("r_bin_get_info: Cannot get bin info\n");
+					return 0;
 				}
-				if (!info) {
-					eprintf ("r_bin_get_info: Cannot get bin info");
-					r_list_free (old_file_hashes);
+				RList *new_file_hashes = r_bin_file_hash (core->bin, limit);
```
ret2libc: this does not re-compute the securihash, does it?

abcSup: You are right, it does not.

libr/main/radare2.c
abcSup: From my understanding, `r_list_join` join the list by pointing the tail of the first list to the head of the second list, so I would need to free the struct RList after that and there is another `r_list_free` which free the elements.

ret2libc: Oooops sorry I misread. I thought you were calling r_list_free, but you are correctly calling free!

abcSup:
> Looking at it now, I'm not sure adding the securihash to the `info` method was the best approach :(
> 
> I have not run your branch, but I believe `r_file_bin_hash` will not compute the format-specific hashes, so SecuriHash will be computed only the first time and it will never be updated, right? Also, I think the API is not very good, as it forces users of `r_bin_file_hash` to save the result in `o->info->file_hashes` and that's error prone (because you have to remember to manually update that value every time you use r_bin_file_hash) and not very encapsulating.
> 
> For the `r_bin_file_hash` APIs, what about:
> 
> ```c
> // compute hashes of the current rbinfile and return them
> RList *r_bin_file_compute_hashes(RBin *bin, int limit);
> // set `hashes` as the last computed hashes for the current rbinfile. `hashes` becomes owned by the current rbinfile, so you don't need to free it. If other hashes were previously set for the current rbinfile, they are returned and they should be freed by the caller when not needed anymore.
> RList *r_bin_file_set_hashes(RBin *bin, RList *hashes);
> ```
> 
> So for `it` you can do something like:
> 
> ```c
> RList *new_hashes = r_bin_file_compute_hashes (bin, limit);
> RList *old_hashes = r_bin_file_set_hashes (bin, new_hashes);
> .... do the diff, print, etc ....
> r_list_free (old_hashes);
> ```
> 
> And in radare2.c something like:
> 
> ```c
> r_bin_file_set_hashes (bin, r_bin_file_compute_hashes (bin, limit));
> ```
> 
> With regard to where to place the new securihash, maybe the best thing was indeed to add a `RBinPlugin.hashes` method and call it at the end of `r_bin_file_compute_hashes` function.
> 
> What do you think?

Initially, I was thinking that `RBinInfo` should always keep the original hashes in `info->file_hashes` and subsequent calls to `r_bin_file_hashes` will compute new hashes, so the users can always know that the file has been changed previously instead of discarding the old hashes and replacing it with the new ones. And `bfile.c:r_bin_file_hashes` will not hash files larger than the `limit` argument and discard the old hashes in `info->file_hashes` (Authentihash), so keeping the original hashes will be more appropriate.

I think your suggestions are good by including the binary specific hash in `RBinInfo` as a method and call it from `r_bin_compute_hash` and having a new method `r_bin_file_set_hashes` to set and return the old hashes. These help resolving previous design problems I faced.

ret2libc:
> I think your suggestions are good by including the binary specific hash in `RBinInfo` as a method and call it from `r_bin_compute_hash` and having a new method `r_bin_file_set_hashes` to set and return the old hashes. These help resolving previous design problems I faced.

I guess you meant `RBinPlugin`, right?

abcSup:
> I guess you meant RBinPlugin, right?
Yeah, sorry for the typo. I will work on it first and see how it goes. Thanks for the help! :)

ret2libc:
No problem, thank you for working on this! The base stuff that calculates the securihash is actually good already I think, it's just a matter of integrating it with r2 code base ;)

abcSup:
I have clicked wrong for the revert pull request. I guess I will create another pull request for the refactor in two hours.

radare:
Yep better improve in a separate pr than reverting. The contrib is good 👍

PR #15992, opened 2020-02-17
Making install call not dir dependent

bannsec:

**Detailed description**
When I attempted to build sanitize version previously I got an error about not finding sys/install.sh. Looks like there's an assumption that sanitize is called from the root dir. I've removed this assumption so it can be called from anywhere.

**Test plan**
```bash
cd sys
./sanitize.sh # This should work
```

**Closing issues**

### DIFF
--- a/sys/sanitize.sh
+++ b/sys/sanitize.sh
@@ -3,6 +3,8 @@
 # SANITIZE="address signed-integer-overflow"  # Faster build
 SANITIZE=${SANITIZE:="address undefined signed-integer-overflow"}
 
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
 printf "\033[32m"
 echo "========================================================================="
 printf "\033[33m"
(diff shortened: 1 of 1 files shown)


radare:
all the scripts in sys/ are suposed to be executed from the root directory of r2

sys/sanitize.sh
```diff
@@ -3,6 +3,8 @@
 # SANITIZE="address signed-integer-overflow"  # Faster build
 SANITIZE=${SANITIZE:="address undefined signed-integer-overflow"}
 
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
radare: this is an ugly bashism

bannsec: It's ugly, but it's the most cross platform solution in bash to reliably get the working directory of the script. I have yet to have it fail on me.

bannsec:
Eh, prolly just best to leave it as it is *shrug*

radare:
nope. its good to fix it, but not using bashisms :)

Issue #15954, opened 2020-02-11
New function signature cant contain dots


radare:
```
$ r2 -
[0x00000000]> afs
No function defined at 0x00000000
[0x00000000]> af
[0x00000000]> afs
fcn.00000000 ();
[0x00000000]> afs foo.bar
<string>:1: error: ';' expected (got ".")
<string>:1: error: declaration expected
[0x00000000]> afs foo.bar()
<string>:1: error: ';' expected (got ".")
<string>:1: error: declaration expected
[0x00000000]> afs foo_bar()
[0x00000000]>
```

PR #15993, opened 2020-02-17
Implement function names with dots in signatures ##anal

oxcabe:

- [x] I've read the guidelines for contributing to this repository
- [x] I made sure to follow the project's coding style
- [ ] I've updated the documentation and the radare2 book with the relevant information (if needed)

**Detailed description**

Merging this Pull Request would allow setting function names containing dots in function signatures.

Current state of this PR is:
- [x] User can set function signatures with dots in the function name using `afs`.
- [ ] Signatures with dots in function name are stored properly in the anal/types SDB.
The current issue (addressed by the second milestone) now is that functions with dots don't get the function type and arguments correctly.

~~In order to fix this, we might need some kind of structure that annotates where the dots were replaced by underscores in the function name, so it can find the proper keys in SDB when querying it.~~. 

Fix functions not getting right arguments in a **new** issue and PR.

**Test plan**
Include a test that asserts if adding dots in the function name of a signature works as intended.

**Closing issues**
Closes https://github.com/radareorg/radare2/issues/15954

### DIFF
--- a/shlr/tcc/tccpp.c
+++ b/shlr/tcc/tccpp.c
@@ -59,7 +59,7 @@ static int unget_saved_buffer[TOK_MAX_SIZE + 1];
 static int unget_buffer_enabled;
 static TokenSym *hash_ident[TOK_HASH_SIZE];
 static char token_buf[STRING_MAX_SIZE + 1];
-/* true if isid(c) || isnum(c) */
+/* true if isid(c) || isnum(c) || isdot(c) */
 static unsigned char isidnum_table[256 - CH_EOF];
 
 static const char tcc_keywords[] =
@@ -2353,7 +2353,7 @@ maybe_newline:
 	case 'Q': case 'R': case 'S': case 'T':
 	case 'U': case 'V': case 'W': case 'X':
 	case 'Y': case 'Z':
-	case '_':
+	case '_': case '.':
 parse_ident_fast:
 		p1 = p;
 		h = TOK_HASH_INIT;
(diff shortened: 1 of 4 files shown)


radare:
maybe we should fix this in RAnalFunction.setName() instead.. but this wont let users to use dots in the function names in r2.. which imho it should be valid char, but it is not valid for tcc. so maybe better to fix tcc instead

oxcabe:
So, allow dots in function names in TCC?

radare:
yep. hopefully it wont break too many things :D what do you think xvilka? ret2libc?

ret2libc:
Please do use the PR template next time.

oxcabe:
Description is updated now, using the PR template

thestr4ng3r:
> maybe we should fix this in RAnalFunction.setName() instead.. but this wont let users to use dots in the function names in r2.. which imho it should be valid char, but it is not valid for tcc. so maybe better to fix tcc instead

I agree with this, we shouldn't limit the zignatures functionality by only the limitations of the parser. We also considered switching from tcc to tree sitter before, which would let us change the parser more easily.

radare:
better approach but still red

shlr/tcc/tccpp.c
```diff
@@ -2353,18 +2353,29 @@ static inline void next_nomacro1(void)
...
+	case '_': case '.':
 parse_ident_fast:
 		p1 = p;
 		h = TOK_HASH_INIT;
 		h = TOK_HASH_FUNC (h, c);
 		p++;
 		for (;;) {
 			c = *p;
-			if (!isidnum_table[c - CH_EOF]) {
+			if (!isidnum_table[*p - CH_EOF]) {
```
radare: this is the same, isnt?

Issue #12061, opened 2018-11-05
Missing Exports and Imports on ELF


enovella:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Ubuntu x86 64
| File format of the file you reverse (mandatory)      |  ELF 
| Architecture/bits of the file (mandatory)            | Aarch64.
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.1.0-git 20076 @ linux-x86-64 git.3.0.1-140-ga0844ef2c commit: a0844ef2c3a2e2852e975634686f0eca4a447093 build: 2018-11-04__22:23:18

### Expected behavior
Show exports (1)  and imports(5 or so from libc)

- Imports:
```c
Address	Ordinal	Name	Library
0000000000020068		realloc	
0000000000020070		__cxa_finalize	
0000000000020078		__stack_chk_fail	
0000000000020080		malloc	
0000000000020088		memcpy	
0000000000020090		memset	
0000000000020098		free	
00000000000200A0		__cxa_atexit	
00000000000200A8		_bss_end__	
00000000000200B0		__bss_start	
00000000000200B8		__end__	
00000000000200C0		__bss_start__	
00000000000200C8		_edata	
00000000000200D0		__bss_end__	
00000000000200D8		_end	
```

- Exports:
```c
Name	Address	Ordinal
dc_::Java_o__003dc_e	000000000000AA20	
start	0000000000000730	[main entry]
```
### Actual behavior
Empty showup

### Steps to reproduce the behavior 
See attached file
[libarm64.zip](https://github.com/radare/radare2/files/2550264/libarm64.zip)

radare:
$ xcrun --sdk iphoneos otool -tv libarm64.so
llvm-objdump: 'libarm64.so': Object is not a Mach-O file type.

enovella:
it is NOT macho file

Quoting radare <notifications@github.com>:

> $ xcrun --sdk iphoneos otool -tv libarm64.so
> llvm-objdump: 'libarm64.so': Object is not a Mach-O file type.
>

radare:
the arch is totally unrelated. this issue is related to the ELF parser

enovella:
Is this then related to the segments within the binary?

radare:
I think so

ret2libc:
[...]

PR #15999, opened 2020-02-18
Fix ELF symbols for names just before the end of strtab ##bin

thestr4ng3r:
**Detailed description**

This bin (`elf/switch-hello-world.elf` in testbins) has a `.dynstr` section of size 1 with only one zero byte and a symbol in `.dynsym` whose name refers to `.dynstr + 0`.
The `sym[k].st_name + 2 > strtab_section->sh_size` check (I assume +2 meant 1 readable char + 1 null byte, but why not allow empty strings with just the null byte?) would reject this and break out of the entire symbol loading loop, so no symbols would be loaded.

**Test plan**

`r2 -c "is" elf/switch-hello-world.elf`
Before: no symbols, after: symbols \0/

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -3528,9 +3528,9 @@ static RBinElfSymbol* Elf_(_r_bin_elf_get_symbols_imports)(ELFOBJ *bin, int type
 					}
 				}
 				ret[ret_ctr].size = tsize;
-				if (sym[k].st_name + 2 > strtab_section->sh_size) {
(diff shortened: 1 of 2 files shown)


radare:
Fixes https://github.com/radareorg/radare2/issues/12061 ?

thestr4ng3r:
No

radare:
Wy so negativ :(

thestr4ng3r:
Positive attitude doesn't fix bugs

radare:
Finally someone agrees! 

Thank you. I mean Fuck you

PR #16014, opened 2020-02-20
Refactor RBin file hashes

abcSup:

**Detailed description**

* Add r_bin_file_compute_hashes
* Add r_bin_file_set_hashes
* Refactor `it` `itj` commands
* Introduce `hashes` method to RBinPlugin

I refactor the `r_bin_file` hashes functionality as suggested @ret2libc in my previous PR #15987

### DIFF
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -856,24 +856,17 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
 	return false;
 }
 
-R_API RList *r_bin_file_hash(RBin *bin, ut64 limit) {
-	r_return_val_if_fail (bin, NULL);
-
-	char hash[128];
-	RHash *ctx;
+R_API RList *r_bin_file_compute_hashes(RBin *bin, ut64 limit) {
+	r_return_val_if_fail (bin && bin->cur && bin->cur->o, NULL);
 	ut64 buf_len = 0, r = 0;
 	RBinFile *bf = bin->cur;
-	if (!bf) {
-		return NULL;
-	}
 	RBinObject *o = bf->o;
-	if (!o || !o->info) {
-		return NULL;
-	}
+
 	RIODesc *iod = r_io_desc_get (bin->iob.io, bf->fd);
 	if (!iod) {
 		return NULL;
 	}
+
 	buf_len = r_io_desc_size (iod);
 	// By SLURP_LIMIT normally cannot compute ...
 	if (buf_len > limit) {
@@ -887,7 +880,8 @@ R_API RList *r_bin_file_hash(RBin *bin, ut64 limit) {
 		return NULL;
 	}
 
-	ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1 | R_HASH_SHA256);
+	char hash[128];
+	RHash *ctx = r_hash_new (false, R_HASH_MD5 | R_HASH_SHA1 | R_HASH_SHA256);
 	while (r + blocksize < buf_len) {
 		r_io_desc_seek (iod, r, R_IO_SEEK_SET);
 		int b = r_io_desc_read (iod, buf, blocksize);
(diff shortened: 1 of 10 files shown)


libr/bin/bfile.c
```diff
@@ -936,13 +936,34 @@ R_API RList *r_bin_file_hash(RBin *bin, ut64 limit) {
...
 	// TODO: add here more rows
 
 	free (buf);
 	r_hash_free (ctx);
 	return file_hashes;
 }
 
+R_API RList *r_bin_file_set_hashes(RBin *bin, RList/*<RBinFileHash*/ *new_hashes) {
+	RBinFile *bf = bin->cur;
+	if (!bf || !bf->o || !bf->o->info) {
```
XVilka: maybe better to use `r_return_if_fail()`?

abcSup: Sure, thanks for the suggestion!

libr/bin/bfile.c
```diff
@@ -936,13 +930,32 @@ R_API RList *r_bin_file_hash(RBin *bin, ut64 limit) {
...
+		free (plugin_hashes);
+	}
 	// TODO: add here more rows
 
 	free (buf);
 	r_hash_free (ctx);
 	return file_hashes;
 }
 
+R_API RList *r_bin_file_set_hashes(RBin *bin, RList/*<RBinFileHash*/ *new_hashes) {
```
ret2libc: please add a comment here, explaining that the returned RList should be freed by the caller and the new one will be owned by RBin.

libr/bin/bfile.c
```diff
@@ -856,24 +856,17 @@ R_API bool r_bin_file_close(RBin *bin, int bd) {
 	return false;
 }
 
-R_API RList *r_bin_file_hash(RBin *bin, ut64 limit) {
-	r_return_val_if_fail (bin, NULL);
-
-	char hash[128];
-	RHash *ctx;
+R_API RList *r_bin_file_compute_hashes(RBin *bin, ut64 limit) {
+	r_return_val_if_fail (bin && bin->cur && bin->cur->o, NULL);
```
ret2libc: Actually this may happen sometimes, it should be checked.
See https://travis-ci.com/radareorg/radare2/jobs/289601576#L739 . I'm not sure why it would happen though.

Same for `r_bin_file_set_hashes`.

libr/bin/bfile.c
```diff
@@ -936,13 +930,32 @@ R_API RList *r_bin_file_hash(RBin *bin, ut64 limit) {
 		sha256h->hex = strdup (hash);
 		r_list_push (file_hashes, sha256h);
 	}
+
+	if (o->plugin && o->plugin->hashes) {
+		RList *plugin_hashes = o->plugin->hashes (bf);
+		if (!r_list_empty (plugin_hashes)) {
```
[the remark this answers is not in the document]
abcSup: Right! :D

libr/bin/bfile.c
abcSup: I found out that those tests use the `-n` flag, so `r_bin` analysis is not performed and `r->bin` will be null.

abcSup: but it should output to `stderr` instead of `stdout` since it is a `assert` (?)

ret2libc: I think these checks were right with just an if then.

In general assert prints to stderr but in Travis we use an env variable to redirect that to stdout so we can catch these kind of problems

libr/main/radare2.c
```diff
@@ -1196,6 +1196,11 @@ R_API int r_main_radare2(int argc, char **argv) {
 				}
 			}
 		}
+		if (o && compute_hashes) {
```
ret2libc: Why moving this here?

abcSup: Previously, `r_bin_file_hash` requires the `iod` argument which is now no longer needed, so I thought it is best to move it outside the `if (iod && !strstr (iod->uri, "://"))`.

abcSup: Now, i see the problem. Currently, in a debugging session the `iod` will have a size of `-1`. Every time users open r2 in debug mode `-d` they will see the warnings about the hashlimits now. Even if putting the statements back into the original if-else, users will see the warnings when they try to do `it` `itj`.

PR #16018, opened 2020-02-21
Random fixes2

ret2libc:

**Detailed description**

Right now if you do `env     AAAA     =    BBBB` it will consider all the spaces, which I think is not right. Commands should not consider spaces, unless explicitly told so (e.g. in the new parser you can do things like `env "    AAAAAA    " = "    BBB"` if yoou really want to include spaces).

**Test plan**

There is a test in the changes, but you can also just do `env    AAA    =   BBB; env AAA` and check that `BBB` is the result.

### DIFF
--- a/libr/core/cmd_eval.c
+++ b/libr/core/cmd_eval.c
@@ -356,6 +356,8 @@ static int cmd_eval(void *data, const char *input) {
 			v = strchr (k, '=');
 			if (v) {
 				*v++ = 0;
+				r_str_trim_head_tail (k);
+				r_str_trim_head_tail (v);
 				r_sys_setenv (k, v);
 			}
 			free (k);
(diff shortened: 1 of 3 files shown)


libr/core/cmd_eval.c
```diff
@@ -356,6 +356,8 @@ static int cmd_eval(void *data, const char *input) {
 			v = strchr (k, '=');
 			if (v) {
 				*v++ = 0;
+				r_str_trim_head_tail (k);
```
radare: one of the issues i see in those trim apis is that sometimes they realloc the heap. and it end up in random uaf situations, this is not the case for this function because this one calls head and tail, which one memmoves and the other appends a null byte.

in this case this function should return void. to avoid this confusion. what do you think?

ret2libc: Yeah it was indeed a bit confusing at first. I can change those API if you want, but it has to wait for Monday

PR #16019, opened 2020-02-21
Fix Spaces Interference in r_flag_get_at()

thestr4ng3r:
**Detailed description**

This fixes a case where `r_flag_get_at()`/`fd` would return NULL/print nothing when there was a flag in a non-selected flagspace at exactly the query addr.

**Test plan**

```r2 -Qc "fs x; f test.x @ 0x42; fs y; f test.y @ 0x0; fd @ 0x42" -```
Before: nothing
After: `test.y + 66`

### DIFF
--- a/libr/flag/flag.c
+++ b/libr/flag/flag.c
@@ -600,13 +600,14 @@ static bool isFunctionFlag(const char *n) {
 R_API RFlagItem *r_flag_get_at(RFlag *f, ut64 off, bool closest) {
 	r_return_val_if_fail (f, NULL);
 
-	RFlagItem *item, *nice = NULL;
+	RFlagItem *nice = NULL;
 	RListIter *iter;
 	const RFlagsAtOffset *flags_at = r_flag_get_nearest_list (f, off, -1);
 	if (!flags_at) {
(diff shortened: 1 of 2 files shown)


radare:
i thought flagspaces were suposed to work like this

thestr4ng3r:
Read the example again

radare:
Thats what i did

ret2libc:
LGTM

thestr4ng3r:
In the last command, `fd @ 0x41` gave `test.y + 65`, `fd @ 0x43` gave `test.y + 67` but `fd @ 0x42` gave nothing.

radare:
And this is expected

Unset the flagspace if u want to see other flagspaces

ret2libc:
No it is not expected. You should see test.y + 66 (the same you would see if the x space was not defined)

thestr4ng3r:
Exactly. Space `y` is the currently selected and has the flag `test.y` at `0`. So flags from `y` should be considered and everything else should be ignored completely.
Thus `fd @ 0x42` should show `test.y + 66`, not nothing.

PR #16024, opened 2020-02-22
Make fd respect asm.flags.real

thestr4ng3r:
**Test plan**

Run:
```
f pingas @ 0x42
fN pingas Snooping as usual I see...
e asm.flags.real=0
fd @ 0x42
fd @ 0x1337
e asm.flags.real=1
fd @ 0x42
fd @ 0x1337
```

Result:
```
pingas
pingas + 4853
Snooping as usual I see...
Snooping as usual I see... + 4853
```

### DIFF
--- a/libr/core/cmd_flag.c
+++ b/libr/core/cmd_flag.c
@@ -1479,21 +1479,24 @@ rep:
 			}
 			f = r_flag_get_at (core->flags, addr, !strict_offset);
 			if (f) {
+				const char *fname = core->flags->realnames && f->realname ? f->realname : f->name;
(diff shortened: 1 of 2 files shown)


thestr4ng3r:
Closed in favor of #16027

radare:
Btw i think we should rename fd to f. And rename f. To ff for function flags

What do u think? I hate that most commands use the . Suffix as an alias for $$ but f. Have a different meaning

XVilka:
Looks fine to me (the rename). Book then should be updated too.

PR #16031, opened 2020-02-22
Make r_table_columns() faster, leak- and double free free

thestr4ng3r:
**Detailed description**

For any case where the new columns were not a bijection of the old ones, it would either leak (when columns were removed) or cause a double free later (when columns were duplicated). This fixes it and makes way more efficient as a nice side effect.

**Test plan**

Run the following examples with valgrind or asan:
`f,name/cols`
`f,name/cols/name`

### DIFF
--- a/test/unit/test_table.c
+++ b/test/unit/test_table.c
@@ -4,35 +4,100 @@
 //TODO test r_str_chop_path
 
 bool test_r_table(void) {
-        RTable *t = r_table_new ();
+	RTable *t = r_table_new ();
 
-        // r_table_fromcsv (t, csv);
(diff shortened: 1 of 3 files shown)


test/unit/test_table.c
```diff
@@ -4,35 +4,100 @@
...
-        r_table_free (t);
+		free (j);
+	}
+	r_table_free (t);
 	mu_end;
 }
 
+bool test_r_table_columns() {
+	RTable *t = NULL;
+#define CREATE_TABLE \
```
radare: i hate mulitline macros

PR #16038, opened 2020-02-23
Add support for @* commands in new r2-shell-parser

ret2libc:

**Detailed description**

This PR improves the new shell parsers in the following ways:
1) it implements all temporary changes commands (`@`, `@a:`, `@b:`, etc.)
2) it uses an hashmap to quickly find the right handler for a given command type
3) it uses TSSymbol (which is an integer) to check whether a node in the syntax tree is of a given type. This is much faster than previous approach that used string comparisons. To do this, the first time the new parser is used, it loads all TSSymbols in variables and then we use those variables.

**Test plan**
Test a bit temporary changes  commands like `@a:`, `@e:`, etc. and try to concatenate them.


radare:
I dont see why a hashtable should be needed when all those commands are one char. Maybe we can do a ht api for one char keys and just hold everything in a 255 element array

ret2libc:
This is not for the commands.. it is for the type of "statements". When you parse
`pd 10 @ 0xdeadbeef @a:x86 @e:asm.bytes=false`, the parser parse this as 
```
(commands [0, 0] - [1, 0]
  (tmp_eval_command [0, 0] - [0, 44]
    (tmp_arch_command [0, 0] - [0, 25]
      (tmp_seek_command [0, 0] - [0, 18]
        (arged_command [0, 0] - [0, 5]
          command: (cmd_identifier [0, 0] - [0, 2])
          args: (args [0, 3] - [0, 5]
            (arg [0, 3] - [0, 5]
              (arg_identifier [0, 3] - [0, 5]))))
        (arg [0, 8] - [0, 18]
          (arg_identifier [0, 8] - [0, 18])))
      (arg [0, 22] - [0, 25]
        (arg_identifier [0, 22] - [0, 25])))
    (tmp_eval_args [0, 29] - [0, 44]
      (tmp_eval_arg [0, 29] - [0, 44]))))
```

`tmp_eval_command`, `tmp_arch_command`, `tmp_seek_command`, `arged_command` (and others) are all type of "commands" that needs to be parsed differently, so the hashtable allows to retrieve the handler of a given type of commands (e.g. tmp_eval_command has a specific TSSymbol(an integer), that is used as a key in the hashtable to retrieve the function handler that temporarily sets the variables)

libr/core/cmd.c
```diff
@@ -5003,23 +5000,323 @@ DEFINE_HANDLE_TS_FCN(tmp_blksz_command) {
...
+}
+
+DEFINE_HANDLE_TS_FCN(tmp_string_command) {
+	TSNode command = ts_node_named_child (node, 0);
+	TSNode arg = ts_node_named_child (node, 1);
+	char *arg_str = ts_node_handle_arg (state, arg);
+	int sz;
+	bool res = false;
+
+	sz = strlen (arg_str);
```
radare: define and assign  in the same line

libr/core/cmd.c
```diff
@@ -5003,23 +5000,323 @@ DEFINE_HANDLE_TS_FCN(tmp_blksz_command) {
...
+DEFINE_HANDLE_TS_FCN(tmp_hex_command) {
+	TSNode command = ts_node_named_child (node, 0);
+	TSNode arg = ts_node_named_child (node, 1);
+	char *arg_str = ts_node_handle_arg (state, arg);
+	int sz;
+	bool res = false;
+
+	size_t len = strlen (arg_str);
+	ut8 *buf = R_NEWS (ut8, len + 1);
+	sz = r_hex_str2bin (arg_str, buf);
```
radare: define and assign in the same line

PR #16043, opened 2020-02-24
Fix segfault in Authenticode hash check (#16042)

### LINKED ISSUE #16042: segfault when loading bin w/ authenticode hash w/ format other than MD5/SHA-1
### Work environment

| OS/arch/bits (mandatory)                             | Arch X86_64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.3.0-git 23895 @ linux-x86-64 git.4.2.1-136-g397c8697e
commit: 397c8697ec1c82c8433ced608e6099ce68441e19 build: 2020-02-24__08:57:36

### Expected behavior
`r2` does not segfault when loading a PE file.

### Actual behavior

`r2` segfaults when loading a PE file w/ Authenticode hash value of type other than MD5/SHA1.

- Find a binary with a SHA-256 (in my case) Authenticode hash value
- Open it with r2

haystack-ia:
- [ X] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [ X] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [X ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [X ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**
First off, I apologize for no tests atm. I can add later but it will take more time than the patch.

I noticed that `r2` will crash when loading a PE file with
Authenticode digest algorithm other than SHA-1 or MD5. I traced
it down to the `PE_(bin_pe_compute_authentihash)` function returning
`NULL` if it encounters an unsupported digest function. This results
in `NULL` being passed to `strcmp` which causes the segfault.

Solution was to add a check for `PE_(bin_pe_compute_authentihash)`
returning `NULL` and to set `bin->is_authhash_valid` to `NULL`.
The real solution is to add support for more algorithms but this will
stop crashes for now.

...

**Test plan**

Find a PE file with an Authenticode digest format other than SHA-1/MD5 and open it.
...

**Closing issues**

closes #16042

### DIFF
--- a/libr/bin/format/pe/pe.c
+++ b/libr/bin/format/pe/pe.c
@@ -2857,8 +2857,14 @@ static int bin_pe_init_security(struct PE_(r_bin_pe_obj_t) * bin) {
 	if (bin->cms && bin->spcinfo) {
 		const char *actual_authentihash = PE_(bin_pe_compute_authentihash) (bin);
 		const char *claimed_authentihash = PE_(bin_pe_get_claimed_authentihash) (bin);
-		bin->is_authhash_valid = !strcmp (actual_authentihash, claimed_authentihash);
-		free ((void *)actual_authentihash);
+		if (actual_authentihash && claimed_authentihash) {
+			bin->is_authhash_valid = !strcmp (actual_authentihash, claimed_authentihash);
+		} else {
+			bin->is_authhash_valid = NULL;
+		}
+		if (actual_authentihash) {
+			free ((void *)actual_authentihash);
+		}
 		free ((void *)claimed_authentihash);
 	}
 	bin->is_signed = bin->cms != NULL;


abcSup:
Thank you for the PR and it was my fault. The reason I wrote the code to support only md5, sha1 is because based on the [Authenticode documentation](https://download.microsoft.com/download/9/c/5/9c5b2167-8017-4bae-9fde-d599bac8184a/Authenticode_PE.docx), it states that it only supports those algorithms. So, I think your code is great and there is no need to support other algorithms (?) I am confused if Microsoft missed the update in their docs.

radare:
86D2311E-9BE6-4751-931B-399409A51F6F

radare:
Seems unrelated

PR #16052, opened 2020-02-25
Replace O(n2) ELF symbol matching with hashmaps

karliss:

**Detailed description**

Previously when loading symbols from ELF file code was checking for duplicate symbols in two list by comparing every element in one list with every element in second list resulting in worst case O(n^2) complexity. Just opening a large file could easily take 10-15 minutes or more.

...

**Test plan**

Just having a large executable isn't enough (self extracting archive will not work) it needs to have many symbols in multiple lists.
I was locally testing using Blender executable from my distro which was ~70MB and had ~100k symbols in each list. Before changes it took ~15 minutes just to open it `r2 /usr/bin/blender` without any additional commands.  After changes it opened in a few seconds.

...

**Closing issues**

...

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -3225,23 +3225,33 @@ static RBinElfSymbol *Elf_(get_phdr_symbols)(ELFOBJ *bin, int type) {
 
 static int Elf_(fix_symbols)(ELFOBJ *bin, int nsym, int type, RBinElfSymbol **sym) {
 	int count = 0;
+	int result = -1;
 	RBinElfSymbol *ret = *sym;
 	RBinElfSymbol *phdr_symbols = Elf_(get_phdr_symbols) (bin, type);
 	RBinElfSymbol *tmp, *p;
+	HtUP *phd_offset_map = ht_up_new0 ();
+	HtUP *phd_ordinal_map = ht_up_new0 ();
 	if (phdr_symbols) {
 		RBinElfSymbol *d = ret;
+		p = phdr_symbols;
+		while (!p->last) {
+			ht_up_insert (phd_offset_map, p->offset, p);
+			ht_up_insert (phd_ordinal_map, p->ordinal, p);
+			p++;
+		}
 		while (!d->last) {
 			/* find match in phdr */
-			p = phdr_symbols;
-			while (!p->last) {
-				if (d->offset == p->offset || p->ordinal == d->ordinal) {
-					p->in_shdr = true;
-					if (*p->name && *d->name && r_str_startswith (d->name, "$")) {
-						strcpy (d->name, p->name);
-					}
-				}
-				p++;
+			p = ht_up_find (phd_offset_map, d->offset, NULL);
+			if (!p) {
+				p = ht_up_find (phd_ordinal_map, d->ordinal, NULL);
 			}
+			if (p) {
+				p->in_shdr = true;
+				if (*p->name && *d->name && r_str_startswith (d->name, "$")) {
+					strcpy (d->name, p->name);
+				}
+			}
+
 			d++;
 		}
 		p = phdr_symbols;
@@ -3257,7 +3267,8 @@ static int Elf_(fix_symbols)(ELFOBJ *bin, int nsym, int type, RBinElfSymbol **sy
 			/*what happens if a shdr says it has only one symbol? we should look anyway into phdr*/
 			tmp = (RBinElfSymbol*)realloc (ret, (nsym + count + 1) * sizeof (RBinElfSymbol));
 			if (!tmp) {
-				return -1;
+				result = -1;
+				goto done;
 			}
 			ret = tmp;
 			ret[nsym--].last = 0;
@@ -3271,9 +3282,14 @@ static int Elf_(fix_symbols)(ELFOBJ *bin, int nsym, int type, RBinElfSymbol **sy
 			ret[nsym + 1].last = 1;
 		}
 		*sym = ret;
-		return nsym + 1;
+		result = nsym + 1;
+		goto done;
 	}
-	return nsym;
+	result = nsym;
+done:
+	ht_up_free (phd_offset_map);
+	ht_up_free (phd_ordinal_map);
+	return result;
 }
 
 static bool is_section_local_sym(ELFOBJ *bin, Elf_(Sym) *sym) {
@@ -3390,6 +3406,7 @@ static RBinElfSymbol* Elf_(_r_bin_elf_get_symbols_imports)(ELFOBJ *bin, int type
 	Elf_(Sym) *sym = NULL;
 	ut8 s[sizeof (Elf_(Sym))] = { 0 };
 	char *strtab = NULL;
+	HtPP *symbols_by_name = NULL;
 
 	if (!bin || !bin->shdr || !bin->ehdr.e_shnum || bin->ehdr.e_shnum == 0xffff) {
 		return Elf_(get_phdr_symbols) (bin, type);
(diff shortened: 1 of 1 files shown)


radare:
❤️

karliss:
Looks like one of the simplifications I did got caught by the tests. Will take a look.

karliss:
Looks like one of the simplifications I did got caught by the tests. Will take a look.

thestr4ng3r:
:heart_decoration:

ITAYC0HEN:
love

libr/bin/format/elf/elf.c
```diff
@@ -3502,14 +3519,20 @@ static RBinElfSymbol* Elf_(_r_bin_elf_get_symbols_imports)(ELFOBJ *bin, int type
 			memset (ret + ret_size, 0, nsym * sizeof (RBinElfSymbol));
 			prev_ret_size = ret_size;
 			ret_size += nsym;
+			symbols_by_name = ht_pp_new0 ();
+			for (k = 0; k < prev_ret_size; k++) {
```
ret2libc: doesn't this loop re-evaluate elements that were already evaluated in the previous iteration?

libr/bin/format/elf/elf.c
```diff
@@ -3546,21 +3569,18 @@ static RBinElfSymbol* Elf_(_r_bin_elf_get_symbols_imports)(ELFOBJ *bin, int type
...
 						bool found = false;
-						j = -1;
-						while (!ret[++j].last && j < prev_ret_size) {
-							if (ret[j].offset == ret[ret_ctr].offset &&
-									strcmp (ret[j].name, "") != 0 && strcmp (ret[j].name, &strtab[st_name]) == 0
-									&& strcmp (ret[j].type, type2str (NULL, NULL, &sym[k])) == 0) {
-								found = true;
-								break;
-							}
+						RBinElfSymbol *previous = ht_pp_find (symbols_by_name, &strtab[st_name], NULL);
```
ret2libc: Is a HT by name enough? I *think* you could have multiple symbols with the same name but different properties (e.g. offset, type, etc.) and you would have to consider them differently.

In theory, we could still use a HtPP with a custom HashFunction that would accept as key name+offset+type.

libr/bin/format/elf/elf.c
```diff
@@ -3546,21 +3569,18 @@ static RBinElfSymbol* Elf_(_r_bin_elf_get_symbols_imports)(ELFOBJ *bin, int type
...
+						if (previous && previous->offset == ret[ret_ctr].offset &&
```
ret2libc: this additional check is not enough, because if you have symbols with same name but different other properties, only one of them would be stored in the Ht, the others would be "overwritten" because they all have the same key.

thestr4ng3r:
51gve-KcvyL _SY355_

libr/bin/format/elf/elf.c
karliss: You mean previous iteration of section loop? Yes, but `ret` array gets relocated which invalidates the old pointers. It shouldn't be a problem unless there is large amount of symbol sections, can there be more than 2?

libr/bin/format/elf/elf.c
karliss: This is the same discussion as one in comment bellow, right?

libr/bin/format/elf/elf.c
karliss: I consider this case when writing the code, but wasn't sure if it's an actual problem?

Assuming there are only 2 symbol sections (SHT_SYMTAB and SHT_DYNSYM), it would mean that first one must contain multiple symbols with identical nonempty name but different offsets or types. From what i saw identical symbols in SYMBTAB and DYNSYM are common so it's important to deduplicate those, but how likely are almost identical symbols within single table? 
If such situation does happen what's the negative effect,? Any code using the data is supposed to handle multiple symbols with identical name but different properties already due to previous deduplication logic. For human inspecting the data it would mean that in case where there are two symbols with identical name there will be 3 instead.

ret2libc: look at `test/bins/elf/analysis/rust_full`, and search for `str.1`. There are many different symbols with that same name. IIRC there are other cases I saw in the past, I'll try to look for other examples, but that's already one.

karliss: Ok, will take a look.

radare:
And rebase

libr/bin/format/elf/elf.c
karliss: Weird for some of those str symbols numbers keep increasing so they are mostly unique but low number like `str.1` get repeated. Will try to use all three properties as hashmap key like you suggested above.

ret2libc: You should probably use something like:
```C

ut32 myhash(const void *k) {
  RBinElfSymbol *sym = (RBinElfSymbol *)k;
  return dosomehashwith(sym.name, sym.offset, sym.type);
}

int mycmp(const void *k1, const void *k2) {
  RBinElfSymbol *sym1 = k1, *sym2 = k2;
  return strcmp(sym1->name, sym2->name) == 0 && sym1->offset == sym2->oiffset && ...;
}

ut32 mycalcsizeK(const void *k) {
  RBinElfSymbol *sym = k;
  return strlen (k->name);
}

HtPPOptions opt = {
  .cmp = mycmp,
  .hashfn = myhash,
  .dupkey = NULL,
  .calcsizeK = mycalcsizeK,
  .freefn = myfree,
};
HtPP *ht = ht_pp_new_opt (&opt);
ht_pp_insert (ht, ret[k], ret[k]);
```

ret2libc: `dupKey` should not be necessary because the `ret` objects are always valid while you use the ht (at least in this case).

ret2libc:
@karliss it seems travis is already green, however I still think the change of the hashtable is necessary. What do you think?

karliss:
@ret2libc I agree, just finished writing it now need to test it.

libr/bin/format/elf/elf.c
```diff
@@ -3390,6 +3427,16 @@ static RBinElfSymbol* Elf_(_r_bin_elf_get_symbols_imports)(ELFOBJ *bin, int type
 	Elf_(Sym) *sym = NULL;
 	ut8 s[sizeof (Elf_(Sym))] = { 0 };
 	char *strtab = NULL;
+	HtPP *symbol_map = NULL;
+	HtPPOptions symbol_map_options = {
+		.cmp = (HtPPListComparator)cmp_RBinElfSymbol,
+		.hashfn = hashRBinElfSymbol,
+		.dupkey = NULL,
+		.calcsizeK = NULL,
```
ret2libc: you may want to implement calcsizeK as well (maybe just use the strlen of the name symbol) as it is used as a quick check to see if two keys are equal.

karliss: I can try it, but I doubt it will give much improvement on top offset check which is done before any string comparisons.

ret2libc: good point ;) ok. How is it going btw? Is the test binary still fast to analyze with this change to Ht?

karliss: There is no noticeable difference, with and without strlen.

ret2libc: Good. Do you have other changes to do?

karliss: No

ret2libc:
I've seen that the original code in `r_bin_elf_get_symbols_imports` was introduced in https://github.com/radareorg/radare2/pull/12713 . You can see https://github.com/radareorg/radare2/issues/12654 for an example where that's really needed. 

Your code still make the example work as it should, so I guess it's fine :)

However, I noticed a problem that probably was already there.
If you try to look at rust_full binary, there is `free` in both .dynsym and .symtab, exactly it is `free@@GLIBC_2.2.5`. However, by adding some eprintfs, you can see that in the hashtable `free` is inserted and then, later on, `free@@GLIBC_2.2.5` is searched (which of course does not find anything). I do not think this problem is directly related to your PR, but I wanted to keep note of it.

I think I'm going to merge your PR, thanks a lot for this!

PR #16053, opened 2020-02-25
Minor Fixes and Tests for NSO/NRO

thestr4ng3r:
**Detailed description**

`readLE32()` didn't load explicitly as LE and these two functions are redundant too.


radare:
Rebase

PR #16056, opened 2020-02-26
Avoidify the RUtil.strTrim() APIs + cleanup/refactor ##util

radare:
…l() and add asserts in RConfig ##refactor" (#16055)"

This reverts commit 9ae41ae69f89476cd9e0f946920e37bc79edbb5c.

 

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


libr/core/anal_tp.c
```diff
@@ -73,17 +73,10 @@ static void var_rename(RAnal *anal, RAnalVar *v, const char *name, ut64 addr) {
 	r_anal_var_rename (anal, fcn->addr, 1, v->kind, v->name, name, false);
 }
 
-static void var_retype(RAnal *anal, RAnalVar *var, const char *vname, char *type, ut64 addr, bool ref, bool pfx) {
-	if (!type || !var) {
-		return;
-	}
-	char *trim = type;
-	r_str_trim (trim);
```
ret2libc: Why did you remove this? I think it is necessary in some cases. You should use `r_str_trim_dup (trim)` and free the resulting string at the end of the function.

libr/core/anal_tp.c
```diff
@@ -73,17 +73,10 @@ static void var_rename(RAnal *anal, RAnalVar *v, const char *name, ut64 addr) {
...
-	if (!*trim) {
-		return;
-	}
+static void var_retype(RAnal *anal, RAnalVar *var, const char *vname, const char *type, ut64 addr, bool ref, bool pfx) {
+	r_return_if_fail (anal && var && type);
```
ret2libc: this is causing multiple warnings in travis.

radare:
so passing null type was actually useful for this :?

Screenshot 2020-02-27 at 12 23 58

libr/core/anal_tp.c
```diff
@@ -75,6 +75,8 @@ static void var_rename(RAnal *anal, RAnalVar *v, const char *name, ut64 addr) {
 
 static void var_retype(RAnal *anal, RAnalVar *var, const char *vname, const char *type, ut64 addr, bool ref, bool pfx) {
 	r_return_if_fail (anal && var && type);
+	// XXX types should be passed without spaces to trim
+	type = r_str_trim_head_ro (type);
```
ret2libc: if you do this, you re-introduce the same bug we had before, where you are writing to read-only memory. Please use r_str_trim_dup instead.

radare:
Nope. The ro means read only

libr/core/anal_tp.c
radare: where

libr/core/anal_tp.c
radare: in which cases it is necessary? the argument is char * so it can be modified, and we should change it to const char *, but we should also not trim anything from that string because it shouldnt be receiving a type with trailing spaces for example...

libr/core/anal_tp.c
radare: nope i dont because _ro means read-only

radare:
.

PR #16061, opened 2020-02-26
Fix r_cons_rgb_parse() harder

### LINKED ISSUE #16060: Cutter color themes showed wrongly on updated radare2 submodule
**Environment information**
* Operating System: Arch
* Cutter version: Newest from git
* Obtained from:
  - Built from source
* File format: Any

**Describe the bug**

After updating radare2, Cutter's color themes (e.g Ayu) looks very weird on the widgets and on the integrated console. Interestingly, themes look just fine on radare2 from the terminal.

**Ayu color theme on Cutter:**
Disassembly:
image

Cutter's console widget:
image

Radare2 console from the regular terminal:
image

**To Reproduce**

1. Update the radare1 submodule
2. Build Cutter with the updated version of radare2
3. Open a binary in Cutter
4. See that the themes look weird

**Expected behavior**
Themes should look as before, with no broken colors

thestr4ng3r:
- [ ] ~~I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)~~

**Detailed description**

This fixes the broken color themes in Cutter and some read beyond the given string.

**Test plan**

* Use Cutter
* Run the added unit tests

**Closing issues**

Fix #16060

### DIFF
--- a/libr/anal/meta.c
+++ b/libr/anal/meta.c
@@ -767,7 +767,7 @@ R_API void r_meta_print(RAnal *a, RAnalMetaItem *d, int rad, PJ *pj, bool show_f
 				break;
 			case 'H':
 				{
-					ut8 r, g, b, A;
+					ut8 r = 0, g = 0, b = 0, A = 0;
 					const char *esc = strchr (d->str, '\x1b');
 					r_cons_rgb_parse (esc, &r, &g, &b, &A);
 					a->cb_printf ("%s rgb:%02x%02x%02x @ 0x%08"PFMT64x"\n",
--- a/libr/cons/rgb.c
+++ b/libr/cons/rgb.c
@@ -116,28 +116,47 @@ R_API int r_cons_rgb_parse(const char *p, ut8 *r, ut8 *g, ut8 *b, ut8 *a) {
 	}
 	if (*p == 0x1b) {
 		p++;
-		if (*p != '[') {
-			p--;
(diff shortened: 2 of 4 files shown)


karliss:
What do you think of fixing call sites? Last time I checked there were only two of them.

thestr4ng3r:
Yes, there are only two. But what is there to fix about them except what I just pushed?

karliss:
Making them pass pointer to the same position of color code so that r_cons_rgb_parse doesn't have to guess if it was given pointer "0x1b" '[" or symbol after "[". And documenting what r_cons_rgb_parse expects as input.

Either it expect pointer to 0x1b and r_cons_html_filter needs to be updated and it should return an error if the input doesn't start with "0x1b[" or it expects pointer after "[" and r_meta_print needs to be updated.

By the way I am not fully convinced by r_meta_print having to parse terminal color code. Storing line color annotations as terminal color sequence makes  the project less portable. In my opinion it would be better to save them in r2 specific format and convert it to terminal color during disassembly printing using current terminal settings.

thestr4ng3r:
I don't really have a strong negative opinion on the \x1b and [ part. I think the way it is now (skip over \x1b if found, then skip over [ if found) is quite clear, without subtracting from the pointer.

About meta, I agree, but this should be done in another pr.

PR #16062, opened 2020-02-26
Properly null-terminate a string

pelijah:


pelijah:
I have found more places where the same code snippet is used. Do we really need to manually null-terminate a string returned by `fgets()`? @ret2libc @radare

ret2libc:
In theory we do not need to terminate them as they are already terminated. I'm also noticing that in many places we call fgets with size-1 but that's not needed because fgets reads at most size-1 and always terminates with a null byte .

PR #16078, opened 2020-02-29
Rename the Function Flag on afn

thestr4ng3r:
**Detailed description**

Before, when you renamed a function, you would get something like this:
Screenshot from 2020-02-29 17-54-11
The `fcn...` flag is useless at this point and you would have to manually delete it.
This pr automatically renames it, but only if the matching flag is in the "functions" namespace. Otherwise we might rename flags that could have another purpose, like for symbols. In such a case, a new flag is created.

I think that the entire flag setting code for functions should be refactored, maybe using REvent, but this patch is simple and clear and helps us define the desired behavior in the form of tests.

**Test plan**

see screenshot


radare:
is afs also renaming the flag?

thestr4ng3r:
I didn't try it, but it should. Both use `__setFunctionName()`

PR #16080, opened 2020-03-01
Make Strings in RIOPlugin const char *

thestr4ng3r:
**Detailed description**

This avoids warnings when writing a plugin in C++.


radare:
This will make the python io plugins fail

PR #16108, opened 2020-03-02
Build LinuxMeson with -O2

radare:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ jobs:
     # Linux with meson and coverage and asserts
     - os: linux
       name: Linux-meson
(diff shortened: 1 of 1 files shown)


radare:
we still need to know why meson builds are that slow

radare:
@ret2libc can u open a PR with just build+install with meson to compare build times instead of running the whole suite? maybe the bottleneck is there

ret2libc:
As said, meson build uses COVERAGE=1, so there are additional instrumentation points and reports being done and uploaded when running meson. So I don't think 17min of meson vs 14min of gcc build is that much of a problem.

radare:
ok, then it will be good to rename the job to say something about coverage

PR #16111, opened 2020-03-03
tree-sitter: support iter commands

ret2libc:

**Detailed description**

This PR does a bunch of things:
- add support for iter commands (@@?)
- clean a bit the code that does command substitution, as it was necessary for some iter commands like `@@c:` which gets converted into `@@=`
- fixes `@@i` in the new parser to use `r_anal_bb_size_i` to compute the size of an instruction. this is important because the old way to compute it is not stable, as it directly uses internal fields that may not have been initialized or contain wrong values.
- add support for recursive help and fixes the detail option in the new parser (e.g. `pf??*` does not work in the old parser, but it should print both basic pf help and detail1 help)
- update radare2-shell-parser, which has many small improvements (e.g. better parsing of macro definition commands like `(foo x y, pd $0 @ $1)`)

**Test plan**

As usual, compile with `meson -Duse_treesitter=true build; ninja -C build`. Set `cfg.newshell=true`.
Try some iter commands (e.g. `@@i`, `@@b`, `@@f`, etc.) and they should work as before.
Try recursive help. Commands should mostly work as before.


ret2libc:
@thestr4ng3r your command `fd @F:flagspace1 @@ sym.* @F:flagspace2` should work well now.

ret2libc:
>     * fixes `@@i` in the new parser to use `r_anal_bb_size_i` to compute the size of an instruction. this is important because the old way to compute it is not stable, as it directly uses internal fields that may not have been initialized or contain wrong values.

Actually, `pd @@i` does not work as i would expect yet, but that's very likely related to some black magic that `pd` does with the blocksize when the blocksize is small. Nevermind.

ret2libc:
Ok, so I broke everything, but I think it's for good!
The failing test is:
```
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -i /tmp/tmp-17117HD9F1xmZWxdc ../bins/elf/analysis/ls-linux64
e asm.comments=false
e asm.functions=false
e asm.lines.bb=false
e asm.flags=false
ahb 16
aha arm
pd 1
e anal.ignbithints=1
pd 1 @b:32
e anal.ignbithints=0
pd 1 @b:16
pd 1 @b:32
ah*
pd 1 @a:x86
ah*
pd 1 @a:x86:64
ah*

 0x00005c20      31ed4989       ldc p9, c8, [r1, -0x124]!
 0x00005c20      31ed4989       stmdbhi sb, {r0, r4, r5, r8, sl, fp, sp, lr, pc} ^
 0x00005c20      31ed4989       ldc p9, c8, [r1, -0x124]!
 0x00005c20      31ed4989       stmdbhi sb, {r0, r4, r5, r8, sl, fp, sp, lr, pc} ^
 aha arm @ 0x5c20
 ahb 16 @ 0x5c20
 0000:5c20     31ed           xor bp, bp
 aha arm @ 0x5c20
 ahb 16 @ 0x5c20
-0x00005c20      31ed           xor bp, bp
+0x00005c20      31ed           xor ebp, ebp
 aha arm @ 0x5c20
 ahb 16 @ 0x5c20
```

If you do `pd 1 @a:x86:64`, you would expect the bytes to be interpreted as 64bit, so `xor ebp, ebp` is better than `xor bp, bp`. This is cause by a "consistency-fix" I did in cmd.c. Before this patch, when you used `@b:32` `anal.ignbithints` was set to 1. However, it was not set when you used `@a:x86:64`. This PR also makes sure that `@a:arch:bits` sets `anal.ignbithints` as well.

I'm going to fix the test.

PR #16112, opened 2020-03-03
Initial implementation of the RStrf API

radare:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/include/r_util/r_str.h
+++ b/libr/include/r_util/r_str.h
@@ -41,13 +41,24 @@ R_NULLABLE static inline char *r_strf_(r_strf__ *s, const char *fmt, ...) {
 	if (s->idx >= s->size) {
 		s->idx = 0;
 	}
-	va_list ap;
+	va_list ap, ap2;
 	va_start (ap, fmt);
+	va_copy (ap2, ap);
+
 	char *p = s->buf + s->idx;
 	size_t left = s->size - s->idx - 1;
 	int res = vsnprintf (p, left, fmt, ap);
-	r_return_val_if_fail (res != left, NULL);
-	s->idx += res + 1;
+	if (res < left) {
+		s->idx += res + 1;
+	} else {
+		if (res < s->size) {
+			s->idx = 0;
+			p = s->buf;
+			s->idx += vsnprintf (p, s->size, fmt, ap2) + 1;
+		} else {
+			r_return_val_if_reached (NULL);
+		}
+	}
 	va_end (ap);
 	return p;
 }
--- a/libr/include/r_util/r_str.h
+++ b/libr/include/r_util/r_str.h
@@ -3,6 +3,7 @@
 
 #include <wchar.h>
 #include "r_str_util.h"
+#include "r_assert.h"
 #include "r_list.h"
 
 #ifdef __cplusplus
@@ -23,11 +24,37 @@ typedef int (*RStrRangeCallback) (void *, int);
 
 static inline void r_str_rmch(char *s, char ch) {
 	for (;*s; s++) {
-		if (*s==ch) {
+		if (*s == ch) {
 			memmove (s, s + 1, strlen (s));
 		}
 	}
 }
+
+//  -- strf
+typedef struct r_strf_t {
+	size_t idx, size;
+	char buf[0];
+} r_strf__;
+
+R_NULLABLE static inline char *r_strf_(r_strf__ *s, const char *fmt, ...) {
+	if (s->idx >= s->size) {
+		s->idx = 0;
+	}
+	r_return_val_if_fail (s && fmt, NULL);
+	va_list ap;
+	va_start (ap, fmt);
+	char *p = s->buf + s->idx;
+	size_t left = s->size - s->idx - 1;
+	int res = vsnprintf (p, left, fmt, ap);
+	r_return_val_if_fail (res != left, NULL);
+	s->idx += res + 1;
+	va_end (ap);
+	return p;
+}
+
+#define r_strf_frame(x) struct r_strf_t { size_t idx, size; char buf[x]; } r_strf_var = {0, x, 0}
+#define r_strf(x,...) r_strf_((r_strf__*)&r_strf_var, x, __VA_ARGS__)
+//  -- strf
 
 #define R_STR_ISEMPTY(x) (!(x) || !*(x))
 #define R_STR_ISNOTEMPTY(x) ((x) && *(x))
--- a/libr/include/r_util/r_str.h
+++ b/libr/include/r_util/r_str.h
@@ -37,10 +37,10 @@ typedef struct r_strf_t {
 } r_strf__;
 
 R_NULLABLE static inline char *r_strf_(r_strf__ *s, const char *fmt, ...) {
+	r_return_val_if_fail (s && fmt, NULL);
 	if (s->idx >= s->size) {
 		s->idx = 0;
 	}
-	r_return_val_if_fail (s && fmt, NULL);
 	va_list ap;
 	va_start (ap, fmt);
 	char *p = s->buf + s->idx;
--- a/test/unit/test_str.c
+++ b/test/unit/test_str.c
@@ -20,6 +20,17 @@ bool test_r_str_format(void) {
 	mu_end;
 }
 
+bool test_r_str_format_limits(void) {
+        r_strf_frame (9);
+        const char *a = r_strf ("%s.%d", "hello", 42);
+        mu_assert_streq (a, "hello.42", "error in r_strf hello.42");
+
+        const char *b = r_strf ("%s.%d", "world", 24);
+        mu_assert_streq (b, "world.24", "error in r_strf world.24");
+
+        mu_end;
+}
+
 bool test_r_str_replace_char_once(void) {
 	char* str = strdup ("hello world");
 	(void) r_str_replace_char_once (str, 'l', 'x');
@@ -415,6 +426,7 @@ bool all_tests() {
 	mu_run_test(test_r_str_unescape);
 	mu_run_test(test_r_str_constpool);
 	mu_run_test(test_r_str_format);
+	mu_run_test(test_r_str_format_limits);
 	return tests_passed != tests_run;
 }
 
--- a/test/unit/test_str.c
+++ b/test/unit/test_str.c
@@ -3,6 +3,23 @@
 
 //TODO test r_str_chop_path
 
+bool test_r_str_format(void) {
+	r_strf_frame (32);
+	const char *a = r_strf ("%s.%d", "hello", 42);
+	mu_assert_streq (a, "hello.42", "error in r_strf hello.42");
+
+	const char *b = r_strf ("%s.%d", "world", 24);
+	mu_assert_streq (b, "world.24", "error in r_strf world.24");
+
+	const char *c = r_strf ("%s.%d", "finds", 23);
+	mu_assert_streq (c, "finds.23", "error in r_strf finds.23");
+
+	mu_assert_streq (a, "hello.42", "error in r_strf hello.42");
+	mu_assert_streq (b, "world.24", "error in r_strf world.24");
+
+	mu_end;
+}
+
 bool test_r_str_replace_char_once(void) {
 	char* str = strdup ("hello world");
 	(void) r_str_replace_char_once (str, 'l', 'x');
@@ -397,6 +414,7 @@ bool all_tests() {
 	mu_run_test(test_r_str_sanitize_sdb_key);
 	mu_run_test(test_r_str_unescape);
 	mu_run_test(test_r_str_constpool);
+	mu_run_test(test_r_str_format);
 	return tests_passed != tests_run;
 }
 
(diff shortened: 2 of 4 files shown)


test/unit/test_str.c
```diff
@@ -3,6 +3,20 @@
 
 //TODO test r_str_chop_path
 
+bool test_r_str_format(void) {
+	r_strf_frame (4, 32);
```
ret2libc: i really don't like this :( it is 1second that you initially have

r_str_frame (4), then someone needs another string and just use again r_strf but he does not increment r_str_frame and you end up with overwriting other strings pointers :/ it seems fragile to me.

ret2libc: Also, it's easy that you then, after 100 commits, need another string (which may be longer than 32) and you end up with truncated strings.

ret2libc:
If we really want to use the stack, I would go with something like: https://github.com/systemd/systemd/blob/master/src/basic/string-util.h#L131

The additional problem is that we want to do format string, so computing the total size required is harder. Also, we should put a limit for sure on the max size allocated on the stack.

test/unit/test_str.c
radare: 1 second?

it works exactly like sdb_fmt, the 4 is the amount of strings you will be using at the same time, its a cyclic array. well that's up to the developer, and we can add asserts to catch this

radare: and 99% of the usecases this is not necessary, and 99% of the situations it is pretty under control the kind of strings and sizes you will manage in a function, remember, this is local to the function, its not global like sdb_fmt

radare:
your proposal is uglier and slower, a simple index to the array of buffers is more than enough and we usually just need small strings to construct keys for sdb, ht, etc..

radare:
we can ahve a single buffer and increment by  strlen, but it will make the thing slower to save just few bytes in stack. imoh its an unnecessary complication

ret2libc:
> your proposal is uglier and slower, a simple index to the array of buffers is more than enough and we usually just need small strings to construct keys for sdb, ht, etc..

small is a very subjective thing. In a situation, small may be 10 bytes, in other it could be 100 or more. I don't like having fixed strings. If we want to allocate things on the stack, then it's more direct doing simply something like:
```C
char s1[100];
char s2[32];
char s3[50];
snprintf(s1, sizeof(s1), "%s.%d", "hello", 32);
snprintf(s2, sizeof(s2), "%s.%d", "hello", 32);
snprintf(s3, sizeof(s3), "%s.%d", "hello", 32);
```
It's more direct, less magic behind it. And it has the same advantages. You can also be sure that people won't risk reusing strings by mistake, because if you need another string, you have to manually create another `char s4[...]`.

ret2libc:
> your proposal is uglier and slower, a simple index to the array of buffers is more than enough and we usually just need small strings to construct keys for sdb, ht, etc..

Slower... We should implement in ASM, then. Who cares about slower of 0.00000000000001 seconds. If systemd can deal with these kind of things, I bet radare2 can as well.

Also, it has the advantage of allocating the string of the right size, avoiding the problem of truncated strings.

thestr4ng3r:
Another simple possibility:
```
r_strf (ulumulu, 0x30, "hello %d", 42);
```
which will expand to:
```
char ulumulu[0x30];
snprintf (ulumulu, sizeof (ulumulu), "hello %d", 42);
```
However with this, you couldn't just do
```
somefunc(r_strf(whatever));
```

thestr4ng3r:
The fact that you have to call `r_strf_reweird()` manually and that `r_strf()` can fail (or return null as in the previous commit) kind of defeats the purpose of the whole thing, doesn't it? Since one call might use %s with some potentially long user input and the next will fail.

radare:
obv yes and this is expected behaviour

radare:
oops

/home/travis/build/radareorg/radare2/install/include/libr/r_util/r_str.h:40:2: error: implicit declaration of function 'r_return_val_if_fail' [-Werror=implicit-function-declaration]

ret2libc:
I think it's important to understand what we want with this API. Some questions:
- do we want to avoid having to free memory allocated for short-living strings?
- do we want to be able to specify them directly as arguments of another function? (e.g. `hello(r_str_f ("%s.%d", "hello", 42))`
- do we want it to be super-fast?
- do we care about heap memory fragmentation?

Based on the answers, we may come up with different solutions.

radare:
> On 6 Mar 2020, at 09:14, Riccardo Schirone <notifications@github.com> wrote:
> 
> ﻿
> I think it's important to understand what we want with this API. Some questions:
> 
> do we want to avoid having to free memory allocated for short-living strings?
Yes
> do we want to be able to specify them directly as arguments of another function? (e.g. hello(r_str_f ("%s.%d", "hello", 42))
No
> do we want it to be super-fast?
Yes
> do we care about heap memory fragmentation?
Yes

thestr4ng3r:
> do we want to be able to specify them directly as arguments of another function? (e.g. hello(r_str_f ("%s.%d", "hello", 42))
> No

Isn't that the number 1 purpose for `sdb_fmt()`? To quickly format keys for sdb so you can do stuff like `sdb_set(sdb_fmt("%s.%d", "hello", 42), sdb_fmt("whatever"))` in a single line?

Because if that requirement is dropped, I don't see any reason to make an extra api at all since
`char mystr[0x10]; snprintf(mystr, sizeof(mystr), ...);` will work just fine for all the other points.

ret2libc:
Go ahead with the first approach if you want.

ret2libc:
But something should be done to detect truncated strings and re-use of existing strings. Otherwise, the code base will become very fragile. I am still worried about the maximum length of a string. This r_strf could be easily used with Symbol names, function names or whatever, and I am not so sure those have strict limits on their lengths.

radare:
Just assert

radare:
To sumarize:

* we need to replace sdb_fmt with this
* we want the non-strlen approach
* we want this r_strf api in r2 or in sdb?
* we want to use this api asap to fix all the thready issues of sdb

please confirm all this and ill do the change and merge. it's important to have it asap

ret2libc:
> we want this r_strf api in r2 or in sdb?

In r2.

> it's important to have it asap
yes, but please don't rush it. We had sdb_fmt for so long, it won't really matter if it stays another week  or not. So, wait for approvals before merging anything.

libr/include/r_util/r_str.h
```diff
@@ -23,12 +24,39 @@ typedef int (*RStrRangeCallback) (void *, int);
...
+	size_t x;
+	size_t y;
+	char *buf[0];
+} r_strf__;
+
+static inline char *r_strf_(struct r_strf_t *s, const char *fmt, ...) {
+	if (s->idx >= (s->y * s->x)) {
+		s->idx = 0;
+	}
+	r_return_val_if_fail (s && fmt, NULL);
```
ret2libc: `s` is already used above, it's useless to check it here.

ret2libc: @radare this comment was done 7 days ago and it is still not resolved AFAIS

libr/include/r_util/r_str.h
```diff
@@ -23,12 +24,38 @@ typedef int (*RStrRangeCallback) (void *, int);
...
+	char *p = s->buf + s->idx;
+	size_t left = s->size - s->idx - 1;
+	int res = vsnprintf (p, left, fmt, ap);
+	r_return_val_if_fail (res != left, NULL);
+	s->idx += res + 1;
+	va_end (ap);
+	return p;
+}
+
+#define r_strf_frame(x) struct r_strf_t { size_t idx, size; char buf[x]; } r_strf_var = {0, x, 0}
```
ret2libc: no need to give the struct name, I think. We can keep it anonymous. Moreover, this `r_strf_t` is different from what defined above.

radare:
If the string is truncated it will assert

radare:
Its different on purpose because we cant declare forward

libr/include/r_util/r_str.h
ret2libc: @radare just use anonymous structs. The struct you are defining here is:
```C
struct r_strf_t {
  size_t idx, size;
  char buf[x];
}
```
but then you are actually using another struct which is:
```C
struct r_strf_t {
  size_t idx, size;
  char buf[0];
}
```

Since you don't need the first structure's name, just make it anonymous.

(please try to answer in the comments below mine, otherwise the review becomes a mess).

ret2libc:
> If the string is truncated it will assert

Ok, right, it will assert, but I think this is anyway bad, as it is very common to use something like:
```C
sdb_set (db, "img.base", r_strf("0x%s", "hello"));
sdb_set (db, "img.offset", r_strf("0x%s", "hello"));
sdb_set (db, "img.flags", r_strf("0x%s", "hello"));
sdb_set (db, "img.whatever", r_strf("0x%s", "hello"));
```
Now, either you do `r_str_frame (4 * MAX_LENGTH_OF_EACH_STRING)` (which i found a huge waste of stack space), or you will for sure trigger an assert and have truncated strings.

radare:
you didnt read the code or you didnt understood what i wrote, but the current code is not X *  maxstrsize, sum(allstr)

radare:
i can change the conditional assert with a real asssert, and then everyone will be happy  and we can merge and start to fix all those thread unsafety issues?

ret2libc:
> you didnt read the code or you didnt understood what i wrote, but the current code is not X * maxstrsize, sum(allstr)

I did read the code. What I am saying is that this code will break many times and you won't notice. There are 2 problems:
1) `r_return_val_if_fail (res != left, NULL);` is wrong for multiple reasons... it should be `r_return_val_if_fail (res == left, NULL);`... but that would be wrong as well. Maybe you meant `r_return_val_if_fail (res < left, NULL);`.
2) As I previously said, there will be truncated strings very often, so I wouldn't trust running this thing.

Try the following test:
```C
bool test_r_str_format2(void) {
	r_strf_frame (32);
	const char *a = r_strf ("%s.%d", "AAAAAAAAAAAAAAAAAA", 42);
	mu_assert_streq (a, "AAAAAAAAAAAAAAAAAA.42", "error in r_strf AAAAAAAAAA.42");
	const char *b = r_strf ("%s.%d", "BBBBBBBBBBBBBBBBBB", 42);
	mu_assert_notnull (b, "b should not be null");
	mu_assert_streq (b, "BBBBBBBBBBBBBBBBBB.42", "error in r_strf BBBBBBBBBB.42");
	mu_end;
}
```

you will see that `b` will be null. Now, that seems to me very useless and it would break a lot of places that currently use `sdb_fmt` (e.g. see usages of sdb_fmt in ctype.c or bin_art.c or other places).

libr/include/r_util/r_str.h
radare: solved, sorry i dont use the web all the time

libr/include/r_util/r_str.h
radare: thats what happens when i reply from mail, it was working before it was bought my microsoft... so anyways, im using the web now.

radare: i know, and the reason for having [0] is because its an empty field, which means the struct size wont handle this field of any size and it is actually correct

ret2libc: I know! I'm just saying don't use the struct name r_strf_t in this line because it is useless. And it is anyway different from the same `struct r_strf_t` defined above (as you said, one uses [0], the other does not).

libr/include/r_util/r_str.h
```diff
@@ -23,12 +24,38 @@ typedef int (*RStrRangeCallback) (void *, int);
...
+	r_return_val_if_fail (s && fmt, NULL);
+	if (s->idx >= s->size) {
+		s->idx = 0;
+	}
+	va_list ap;
+	va_start (ap, fmt);
+	char *p = s->buf + s->idx;
+	size_t left = s->size - s->idx - 1;
+	int res = vsnprintf (p, left, fmt, ap);
+	r_return_val_if_fail (res != left, NULL);
```
ret2libc: As said in another comment, this is wrong. It should probably be `res < left`

radare: ths is fixed

test/unit/test_str.c
```diff
@@ -3,6 +3,34 @@
...
+
+	mu_assert_streq (a, "hello.42", "error in r_strf hello.42");
+	mu_assert_streq (b, "world.24", "error in r_strf world.24");
+
+	mu_end;
+}
+
+bool test_r_str_format_limits(void) {
+        r_strf_frame (9);
+        const char *a = r_strf ("%s.%d", "hello", 42);
```
ret2libc: what if you have to use both strings at the same time? e.g. `sdb_set(sdb, r_strf(...), r_strf(...))`

You would have no way to know whether one `r_strf` overwrites the content of the other.

radare:
it should assert as i said, or use the rewind thing that i proposed some weeks ago to make that splicit

test/unit/test_str.c
ret2libc: Yes and once it asserts, what can you do? Asserting means you have done something wrong at *programming time*, but that's not really true. It's just that, sometimes, this implementation will fail. That's why I don't like this.

PR #16132, opened 2020-03-04
Support iter_hit_command and update radare2-shell-parser

ret2libc:

**Detailed description**

Add support for iter_hit_command (`@@/`) and fix parser to recognize arguments with '#' (e.g. `wx 7158 #ldr r1, [r6, r1]`).


test/new/db/formats/mangling/rust
```diff
@@ -44,7 +44,7 @@ EXPECT=<<EOF
 <core::fmt::Write::write_fmt::Adapter<'a, T> as core::fmt::Write>::write_str::he4f4768a2f446fac
 EOF
 CMDS=<<EOF
-!rabin2 -D rust _ZN96_\$LT\$core..fmt..Write..write_fmt..Adapter\$LT\$\$u27\$a\$C\$\$u20\$T\$GT\$\$u20\$as\$u20\$core..fmt..Write\$GT\$9write_str17he4f4768a2f446facE
+!rabin2 -D rust "_ZN96_\$LT\$core..fmt..Write..write_fmt..Adapter\$LT\$\$u27\$a\$C\$\$u20\$T\$GT\$\$u20\$as\$u20\$core..fmt..Write\$GT\$9write_str17he4f4768a2f446facE"
```
radare: is this escaping right? i mean  for the \$ stuff

ret2libc: yes. So, the reason why I need to wrap the string in `"` is that the new parser, opposed to the old one, does allow escape and so it unescape arguments before passing them to commands handlers.

So, if you do `?e Hello\;World` it prints `Hello;World` (the old parser would just complain about having two commands). Now, that escaping/unescaping is done for some "special characters", which are ```@;~$#|`"'()<>```.

Back to why I need to wrap stuff. If you do:
```
!rabin2 -D rust _ZN96_\$LT\$core..fmt..Write..write_fmt..Adapter\$LT\$\$u27\$a\$C\$\$u20\$T\$GT\$\$u20\$as\$u20\$core..fmt..Write\$GT\$9write_str17he4f4768a2f446facE
```
The parser would convert this into:
```
rabin2 -D rust _ZN96_$LT$core..fmt..Write..write_fmt..Adapter$LT$$u27$a$C$$u20$T$GT$$u20$as$u20$core..fmt..Write$GT$9write_str17he4f4768a2f446facE
```
which would make the shell interpret $LT and such as env variables.

Since when you use `!` you are passing strings to `system()`, the shell is already doing is escaping/unescaping, so we do need to pass the `\` before `$`, to make sure the shell does not consider `$` an env var.

So, by wrapping the string in `"..."` we make sure radare2 parser does not unescape the string,  passing it as-is to `system()`.

Does that make sense?

XVilka:
LGTM

test/new/db/formats/mangling/rust
radare: ok

radare: can we have '' to avoid having to escape things? except \' ?

ret2libc: what do you mean?

The way new parser works right now is:
`?e "$ello;World"` consider `argv[1]` = `$ello;World`  
`?e "$ello;\"World\""` consider `argv[1]` = `$ello;"World"`  
`?e '$ello;\'World\''` consider `argv[1]` = `$ello;'World'`  
`?e \$ello "World All"` consider `argv[1]` = `$ello`, `argv[2]` = `World All`  

So, things inside `"` or `'` do not need to be escaped (apart from `"` or `'` themselves, which of course need to be escaped to avoid confusing them with the end of the wrapped string).

In case you are using `!rabin2 ...`, the reason why `\$` is needed inside the `"...."` is because all those strings are passed to `system()`, which does additional unescaping and the shell should not confuse those `$` as the beginning of an shell env var.

PR #16139, opened 2020-03-04
Deprecate cmdtail and kill r_anal_case()

thestr4ng3r:
**Detailed description**

This `cmdtail` us an ugly hack, we should move away from it. Since there aren't too many places where it is used, this shouldn't be too hard.
`r_anal_case()` was one of the places, but it has been dead code since 2018: adaf135dc66d3004237aed9b857da65d62e77190


radare:
i agree on killing r_anal_case because is not used at all, but deprecating cmdtail is too early and not the place to do the change

thestr4ng3r:
`¯\_(ツ)_/¯`

PR #16156, opened 2020-03-06
Strip minuses from the hash names

l29ah:

**Detailed description**

So sha256 and probably other authentihashes work
Only tested with a sha256-hashed PE

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/bin/format/pe/pe.c
+++ b/libr/bin/format/pe/pe.c
@@ -852,13 +852,15 @@ const char* PE_(bin_pe_compute_authentihash)(struct PE_(r_bin_pe_obj_t)* bin) {
 		return NULL;
 	}
 
-	const char *hashtype = bin->spcinfo->messageDigest.digestAlgorithm.algorithm->string;
+	char *hashtype = strdup (bin->spcinfo->messageDigest.digestAlgorithm.algorithm->string);
+	r_str_replace_char (hashtype, '-', 0);
 	ut64 algobit = r_hash_name_to_bits (hashtype);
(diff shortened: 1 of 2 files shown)


l29ah:
Damn `hub`.

radare:
which hashes have dashes?

radare:
can you provide a test or example?

l29ah:
I've observed "sha-256".
I don't have a way to produce signed binaries, and the one i've checked it on is proprietary.

libr/bin/format/pe/pe.c
```diff
@@ -854,8 +854,8 @@ const char* PE_(bin_pe_compute_authentihash)(struct PE_(r_bin_pe_obj_t)* bin) {
 
 	const char *hashtype = bin->spcinfo->messageDigest.digestAlgorithm.algorithm->string;
 	ut64 algobit = r_hash_name_to_bits (hashtype);
```
radare: i think you must pass the filtered string in here, like this: r_str_replace_char (hashtype, '-', 0); (hashtype should be char* and strdupped from that .string

libr/hash/hash.c
```diff
@@ -302,8 +302,12 @@ R_API ut64 r_hash_name_to_bits(const char *name) {
 
 	do {
 		/* Eat everything up to the comma */
-		for (i = 0; *ptr && *ptr != ',' && i < sizeof (tmp) - 1; i++) {
-			tmp[i] = tolower ((ut8)*ptr++);
+		for (i = 0; *ptr && *ptr != ',' && i < sizeof (tmp) - 1;) {
```
radare: remove this logic, this should be implemented using RStr apis and not in here

PR #16164, opened 2020-03-08
Remove some unused members

thestr4ng3r:
These `RAnalCaseOp` members were only set by java (apparently to values that aren't even all correct) and never used for anything except bloating the heap.


XVilka:
Mac OS Travis is mad, ignore it.

PR #16165, opened 2020-03-08
Fix r_anal_switch_op_new() args

thestr4ng3r:
Previous declaration was
```RAnalSwitchOp *r_anal_switch_op_new(ut64 addr, ut64 min_val, ut64 max_val);```
but with the definition
```RAnalSwitchOp *r_anal_switch_op_new(ut64 addr, ut64 min_val, ut64 def_val)```
and `def_val` was just ignored and everything set to `min_val` 🤯


XVilka:
Good catch! Thank you!

PR #16170, opened 2020-03-09
Fix afs not showing signatures correctly with preloaded sdb types ##types

oxcabe:

**Detailed description**

`afs` was not showing return type and arguments for preloaded functions that start with `sym.imp`. This Pull Request fixes this.

**Test plan**

Write a test that shows the output of afs on a preloaded function signature, like `sym.imp.puts`.

**Closing issues**

...

### DIFF
--- a/libr/bin/format/pe/pe.c
+++ b/libr/bin/format/pe/pe.c
@@ -852,13 +852,15 @@ const char* PE_(bin_pe_compute_authentihash)(struct PE_(r_bin_pe_obj_t)* bin) {
 		return NULL;
 	}
 
-	const char *hashtype = bin->spcinfo->messageDigest.digestAlgorithm.algorithm->string;
+	char *hashtype = strdup (bin->spcinfo->messageDigest.digestAlgorithm.algorithm->string);
+	r_str_replace_char (hashtype, '-', 0);
 	ut64 algobit = r_hash_name_to_bits (hashtype);
-	if (!(algobit & (R_HASH_MD5 | R_HASH_SHA1))) {
-		eprintf ("Authenticode only supports md5, sha1. This PE uses %s\n", hashtype);
+	if (!(algobit & (R_HASH_MD5 | R_HASH_SHA1 | R_HASH_SHA256))) {
+		eprintf ("Authenticode only supports md5, sha1, sha256. This PE uses %s\n", hashtype);
+		free (hashtype);
 		return NULL;
 	}
-
+	free (hashtype);
 	ut32 checksum_paddr = bin->nt_header_offset + 4 + sizeof (PE_(image_file_header)) + 0x40;
 	ut32 security_entry_offset =  bin->nt_header_offset + sizeof (PE_(image_nt_headers)) - 96;
 	PE_(image_data_directory) *data_dir_security = &bin->data_directory[PE_IMAGE_DIRECTORY_ENTRY_SECURITY];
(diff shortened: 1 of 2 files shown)


oxcabe:
I'll write a test asap

libr/anal/fcn.c
```diff
@@ -1516,9 +1516,13 @@ R_API int r_anal_fcn_cc(RAnal *anal, RAnalFunction *fcn) {
 
 R_API char *r_anal_function_get_signature(RAnalFunction *function) {
 	RAnal *a = function->anal;
+	char *fcnname = strdup (function->name);
+	if (!strncmp (function->name, "sym.imp.", 8)) {
```
radare: use r_str_startswith() its more idiomatic, even if strncmp is faster

libr/anal/fcn.c
```diff
@@ -1516,9 +1516,13 @@ R_API int r_anal_fcn_cc(RAnal *anal, RAnalFunction *fcn) {
...
+		fcnname = strdup (function->name + 8);
```
radare: fcnname is leaking here

libr/anal/fcn.c
```diff
@@ -1516,9 +1516,13 @@ R_API int r_anal_fcn_cc(RAnal *anal, RAnalFunction *fcn) {
 
 R_API char *r_anal_function_get_signature(RAnalFunction *function) {
 	RAnal *a = function->anal;
+	char *fcnname = (r_str_startswith (function->name, "sym.imp."))
```
ITAYC0HEN: I think we have another way to check for i. ported functions. There's a property telling you whether it's an import. I'm afk, CC @thestr4ng3r probably remember

oxcabe: We have that for symbols, but I couldn't find anything like that but for functions. There's another API function, `r_anal_fcn_format_sig`, that parses it in a similar fashion.

ITAYC0HEN: can an imported function not be a symbol ? I think we made all imported functions symbols in the realname pr.

can't guarantee though

libr/anal/fcn.c
radare: imho we should differentitiate between fcnname and signname.. not sure how we should manage this, maybe holding the original signature info...

libr/anal/fcn.c
radare: this logic is not working everywhere, it was half implemented. thats why this is broken and why this pr

radare: actually you can do const char * and avoid the strdup, just use fcn->name + prefixskip and drop the free

ITAYC0HEN: Let me request a review from @thestr4ng3r , I remember that we tried to avoid comparing to a substring prefix.  Maybe this can be changed in another way

radare:
ideally no hardcoded prefixes should happen. actually no dots should be in the function signature names. so it should be just taking whats after the last dot. r_str_lchr('.')

libr/anal/fcn.c
thestr4ng3r: Yes, the way is to get the flag for the function, then check if it is in the imports flagspace.

radare: can you point to file:line? or code snippet?

radare:
please resolve conflicts

oxcabe:
I'm on it

radare:
Screenshot 2020-03-10 at 16 22 47

oxcabe:
ikr I'm fighting with git right now

PR #16179, opened 2020-03-10
Change macro syntax to use `;` instead of `,`

ret2libc:
- [x] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed). See https://github.com/radareorg/radare2book/pull/236

**Detailed description**

This is a change to ease the implementation of the new parser. Right now the way commands are parsed inside a macro and outside of it is very different. For example, `?e Hello,World)` outside of a macro would print the string `Hello,World)`. However, when you are inside the definition of a macro, the character `)` is the end of the macro body and `,` is the separator between commands. This difference is an inconsistency that changes how you write r2 commands based on what you are doing and it also complicates a lot the parser because arguments and commands inside the macro body should be parsed in a completely different way than what happens outside of a macro body.

For these reasons, I think it's better to just use `;` as a separator for commands inside a macro body. This way everything can be parsed as usual, because `;` is a separator normally.

**Test plan**

Try to define some macros, call them, delete them, etc.

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -2518,6 +2518,30 @@ static char *parse_tmp_evals(RCore *core, const char *str) {
 	return res;
 }
 
+static bool is_macro_command(const char *ptr) {
+	ptr = r_str_trim_head_ro (ptr);
+	while (IS_DIGIT (*ptr)) {
+		ptr++;
+	}
+	return *ptr == '(';
+}
+
+static char *find_ch_after_macro(char *ptr, char ch) {
+	int depth = 0;
+	while (*ptr) {
+		if (depth == 0 && *ptr == ch) {
+			return ptr;
+		}
+		if (*ptr == '(') {
+			depth++;
+		} else if (*ptr == ')') {
+			depth--;
+		}
+		ptr++;
+	}
+	return NULL;
+}
+
 static int r_core_cmd_subst(RCore *core, char *cmd) {
 	ut64 rep = strtoull (cmd, NULL, 10);
 	int ret = 0, orep;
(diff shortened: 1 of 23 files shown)


libr/core/cmd.c
```diff
@@ -2518,6 +2518,32 @@ static char *parse_tmp_evals(RCore *core, const char *str) {
 	return res;
 }
 
+static bool is_macro_command(const char *ptr) {
+	while (isspace (*ptr)) {
```
radare: r_str_trim_head_ro dupe

libr/core/cmd_api.c
```diff
@@ -394,52 +394,15 @@ R_API int r_cmd_macro_add(RCmdMacro *mac, const char *oname) {
...
-			if (buf[lidx]==')' && buf[lidx-1]!='(') {
-				buf[lidx]='\0';
-				memcpy (macro->code+codelen, bufp, lbufp+1);
-				break;
-			}
-			if (*buf != '\n') {
-				memcpy (macro->code+codelen, bufp, lbufp+1);
-				codelen += lbufp;
-			}
+	for (lidx=0; pbody[lidx]; lidx++) {
```
radare: missing space around=

test/new/db/cmd/cmd_macros
```diff
@@ -89,19 +89,19 @@ EXPECT=<<EOF
...
+NAME=3(foo;?e a;?e b)()
 FILE=-
 EXPECT=<<EOF
 a
 b
 a
 b
 EOF
-CMDS=2(foo,?e a,?e b)()
+CMDS=2(foo;?e a;?e b)()
```
radare: that wont work on the old parser, you have to quote the whole thign

ret2libc: no, i should have fixed this. no need to quote the whole thing because i made it look for `;` only after the `)`.

ret2libc: See https://github.com/radareorg/radare2/pull/16179/files#diff-40aa0bc93d40f6d1334c0272c061cafbR2529

test/new/db/cmd/cmd_macros
```diff
@@ -130,6 +130,24 @@ CMDS=<<EOF
 f foo.one = 1
 f foo.bar = 2
 f foo.cow = 3
-(,?v $$)() @@ foo*
+(;?v $$)() @@ foo*
```
radare: I will miss those anonymous macro calls with just the comma. But overall the change looks good to me. Will solve some issues for assembling and such

PR #16208, opened 2020-03-13
Improve ts

ret2libc:

**Detailed description**

Improvements to new parser. It adds support for the new macro syntax (with `;`), it correctly prints `~?` and it makes sure "comments" are ignored when evaluating command substitutions (this is a bit fuzzy as a thing, but existing code relies on that).

### DIFF
--- a/test/new/db/anal/arm
+++ b/test/new/db/anal/arm
@@ -822,7 +822,7 @@ NAME=arm jump table
 FILE=../bins/elf/analysis/callback.elf
 CMDS=<<EOF
 af @ sym.input_handler2
-CC. @ 0x000105b8~?(7
+CC. @ 0x000105b8~?\(7
 EOF
 EXPECT=<<EOF
 1
@@ -833,7 +833,7 @@ NAME=arm jump table
 FILE=../bins/elf/analysis/mobile_bank.45115ff5f655d94fc26cb5244928b3fc
 CMDS=<<EOF
 af @ 0x11284
-CC. @ 0x000112b0~?(8
+CC. @ 0x000112b0~?\(8
 EOF
 EXPECT=<<EOF
 1
--- a/test/new/db/cmd/cmd_afn
+++ b/test/new/db/cmd/cmd_afn
@@ -35,7 +35,7 @@ f@F:functions~483f4,48540
 f@F:*~483f4,48540
 afn mymain @ 0x08048540
 afn @ 0x08048540
-?e The main flag shouldn't be renamed, it comes from bin:
+"?e The main flag shouldn't be renamed, it comes from bin:"
 f@F:functions~483f4,48540
 ?e --
 f@F:*~483f4,48540
--- a/test/new/db/cmd/cmd_macros
+++ b/test/new/db/cmd/cmd_macros
@@ -130,7 +130,7 @@ CMDS=<<EOF
 f foo.one = 1
 f foo.bar = 2
 f foo.cow = 3
-(;?v $$)() @@ foo*
+(_;?v $$)() @@ foo*
 EOF
 RUN
 
--- a/test/new/db/cmd/cmd_open
+++ b/test/new/db/cmd/cmd_open
@@ -64,14 +64,14 @@ NAME=o-[0]
 FILE=malloc://1024
 EXPECT=<<EOF
 EOF
-CMDS=o-`o~[0]`;o
+CMDS=o- `o~[0]`;o
 RUN
 
 NAME=oj-
 FILE=malloc://1024
 EXPECT=<<EOF
 EOF
-CMDS=o-`oj~{[0].fd}`;o
+CMDS=o- `oj~{[0].fd}`;o
 RUN
 
 NAME=o -
--- a/test/new/db/cmd/cmd_pd
+++ b/test/new/db/cmd/cmd_pd
@@ -1838,7 +1838,7 @@ e asm.arch=x86
 e asm.bits=32
 e asm.bytes=false
 s 0x1000
-wz path"C:\\Program\n
+"wz path\"C:\\Program\n"
 f ansi_str
 s 0x2000
 "wa mov eax, 0x1000"
@@ -1849,7 +1849,7 @@ e str.escbslash=false
 pd 1
 pdJ 1
 s 0x3000
-ww path"C:\\Program\r
+"ww path\"C:\\Program\r"
 f wide_str
 s 0x4000
 "wa mov ebx, 0x3000"
--- a/test/new/db/cmd/cmd_repeats
+++ b/test/new/db/cmd/cmd_repeats
@@ -3,14 +3,10 @@ FILE=-
 EXPECT=<<EOF
 16
 18
-16
-18
 EOF
 CMDS=<<EOF
 @@?~?
 @@@?~?
-?@@?~?
-?@@@?~?
 EOF
 RUN
 
--- a/test/new/db/cmd/feat_arroba
+++ b/test/new/db/cmd/feat_arroba
@@ -9,29 +9,29 @@ s @ 0x42
 EOF
 RUN
 
-NAME=@ e:
+NAME=@e:
 FILE=-
 EXPECT=<<EOF
 1
 EOF
 CMDS=<<EOF
 e scr.color=false
-e scr.color @ e:scr.color=true
+e scr.color @e:scr.color=true
 EOF
 RUN
 
-NAME=@ a:
+NAME=@a:
 FILE=-
 EXPECT=<<EOF
 arm
 EOF
 CMDS=<<EOF
 e asm.arch=x86
-e asm.arch @ a:arm
+e asm.arch @a:arm
 EOF
 RUN
 
-NAME=@ e: @ addr
+NAME=@e: @ addr
 FILE=-
 EXPECT=<<EOF
 1
@@ -40,12 +40,12 @@ EOF
 CMDS=<<EOF
 e scr.color=false
 s 0
-e scr.color @ e:scr.color=true @ 0x1337
-s @ e:scr.color=true @ 0x1337
+e scr.color @e:scr.color=true @ 0x1337
+s @e:scr.color=true @ 0x1337
 EOF
 RUN
 
-NAME=@ addr @ e
+NAME=@ addr @e
 FILE=-
 EXPECT=<<EOF
 0
@@ -54,14 +54,14 @@ EXPECT=<<EOF
 0x1337
 EOF
 CMDS=<<EOF
-e scr.color @ 0x42 @ e:scr.color=false 
-e scr.color @ 0x1337 @ e:scr.color=true
-s @ 0x42 @ e:scr.color=false 
-s @ 0x1337 @ e:scr.color=true 
+e scr.color @ 0x42 @e:scr.color=false 
+e scr.color @ 0x1337 @e:scr.color=true
+s @ 0x42 @e:scr.color=false 
+s @ 0x1337 @e:scr.color=true 
 EOF
 RUN
 
-NAME=@ a: @ addr @ e:
+NAME=@a: @ addr @e:
 FILE=-
 EXPECT=<<EOF
 x86
@@ -72,12 +72,12 @@ arm
 1
 EOF
 CMDS=<<EOF
-e asm.arch @ a:x86 @ 0x42 @ e:scr.color=false
-e asm.arch @ a:arm @ 0x1337 @ e:scr.color=true
-s @ a:x86 @ 0x42 @ e:scr.color=false
-s @ @ a:arm @ 0x1337 @ e:scr.color=true
-e scr.color @ a:x86 @ 0x42 @ e:scr.color=false
-e scr.color @ @ a:arm @ 0x1337 @ e:scr.color=true
+e asm.arch @a:x86 @ 0x42 @e:scr.color=false
+e asm.arch @a:arm @ 0x1337 @e:scr.color=true
+s @a:x86 @ 0x42 @e:scr.color=false
+s @a:arm @ 0x1337 @e:scr.color=true
+e scr.color @a:x86 @ 0x42 @e:scr.color=false
+e scr.color @a:arm @ 0x1337 @e:scr.color=true
 EOF
 RUN
 
@@ -89,7 +89,7 @@ false
 EOF
 CMDS=<<EOF
 e asm.pseudo=false
-e asm.pseudo @ e:asm.pseudo=true,asm.pseudo=true
+e asm.pseudo @e:asm.pseudo=true,asm.pseudo=true
 e asm.pseudo
 EOF
 RUN
@@ -102,7 +102,7 @@ false
 EOF
 CMDS=<<EOF
 e asm.pseudo=false
-e asm.pseudo @ e:asm.pseudo=true @ e:asm.pseudo=true
+e asm.pseudo @e:asm.pseudo=true @e:asm.pseudo=true
 e asm.pseudo
 EOF
 RUN
--- a/libr/config/hold.c
+++ b/libr/config/hold.c
@@ -13,6 +13,18 @@ static void r_config_hold_num_free(RConfigHoldNum *hc) {
 	free (hc);
 }
 
+static int key_cmp_hold_s(const void *a, const void *b) {
+	const char *a_s = (const char *)a;
+	const RConfigHoldChar *b_s = (const RConfigHoldChar *)b;
+	return strcmp (a_s, b_s->key);
+}
+
+static int key_cmp_hold_i(const void *a, const void *b) {
+	const char *a_s = (const char *)a;
+	const RConfigHoldNum *b_s = (const RConfigHoldNum *)b;
+	return strcmp (a_s, b_s->key);
+}
+
 R_API bool r_config_hold_s(RConfigHold *h, ...) {
 	va_list ap;
 	char *key;
@@ -25,6 +37,9 @@ R_API bool r_config_hold_s(RConfigHold *h, ...) {
 		}
 	}
 	while ((key = va_arg (ap, char *))) {
+		if (r_list_find (h->list_char, key, key_cmp_hold_s)) {
+			continue;
+		}
 		const char *val = r_config_get (h->cfg, key);
 		if (!val) {
 			continue;
@@ -54,6 +69,9 @@ R_API bool r_config_hold_i(RConfigHold *h, ...) {
 	}
 	va_start (ap, h);
 	while ((key = va_arg (ap, char *))) {
+		if (r_list_find (h->list_num, key, key_cmp_hold_i)) {
+			continue;
+		}
 		RConfigHoldNum *hc = R_NEW0 (RConfigHoldNum);
 		if (!hc) {
 			continue;
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -4672,7 +4672,9 @@ static char *do_handle_substitution_cmd(struct tsr2cmd_state *state, TSNode inn_
 	int value = core->num->value;
 
 	// execute the sub command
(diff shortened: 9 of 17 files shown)


test/new/db/anal/arm
```diff
@@ -822,7 +822,7 @@ NAME=arm jump table
 FILE=../bins/elf/analysis/callback.elf
 CMDS=<<EOF
 af @ sym.input_handler2
-CC. @ 0x000105b8~?(7
+CC. @ 0x000105b8~?\(7
```
ret2libc: these are necessary because `(`, `)` and other chars are considered special chars in the new parser. the old parser should work well because of recent change.

test/new/db/cmd/cmd_afn
```diff
@@ -35,7 +35,7 @@ f@F:functions~483f4,48540
 f@F:*~483f4,48540
 afn mymain @ 0x08048540
 afn @ 0x08048540
-?e The main flag shouldn't be renamed, it comes from bin:
+"?e The main flag shouldn't be renamed, it comes from bin:"
```
ret2libc: Once commands will start be porting to accept `(argc, argv)` instead of doing their own parsing, this should become `?e "The main flag shouldn't be renamed, it comes from bin:"`. The problem is that without any quotes, the char `'` is supposed to be used to wrap strings (e.g. `?e 'Hello    World'`).

For now, let's just use the "legacy" ways of wrapping commands.

test/new/db/cmd/cmd_macros
```diff
@@ -130,7 +130,7 @@ CMDS=<<EOF
 f foo.one = 1
 f foo.bar = 2
 f foo.cow = 3
-(;?v $$)() @@ foo*
+(_;?v $$)() @@ foo*
```
ret2libc: Let's always have a name for macros. Anyway, even when you do `(;?v $$)`, a macro with an empty name will be created. This also solves a problem in the parser, where there may be an ambiguity when evaluating `(;?v $$)`.

One interpreation is `(` and `?v $$)`, another is `(;?v $$)`, but the parser has no way to understand which one you want. By forcing a macro name we can solve this problem.

test/new/db/cmd/cmd_open
```diff
@@ -64,14 +64,14 @@ NAME=o-[0]
 FILE=malloc://1024
 EXPECT=<<EOF
 EOF
-CMDS=o-`o~[0]`;o
+CMDS=o- `o~[0]`;o
```
ret2libc: Those syntax that accept a parameter attached to the command name are just weird and makes the splitting of the command name and the arguments hard! Let's just type an additional character and make sure most of the commands follow the syntax `<command-name> <argument1> <argument2> ...`

test/new/db/cmd/cmd_pd
```diff
@@ -1811,7 +1811,7 @@ pd 1
 e scr.color=3
 ec linehl none
 pd 1
-ec linehl #002f00
+"ec linehl #002f00"
```
ret2libc: `#whatever` is considered a comment. `#` is a special char, so in the new parser you can easily escape it `\#`, but for now let's just quote the entire command.

test/new/db/cmd/cmd_pd
```diff
@@ -1838,7 +1838,7 @@ e asm.arch=x86
 e asm.bits=32
 e asm.bytes=false
 s 0x1000
-wz path"C:\\Program\n
```
ret2libc: This is a `"` without the matching closing `"`, so it is an error in the new parser. Let's just quote everything for now, but later this will become `wz "path\"C:\\Program\n"`

test/new/db/cmd/cmd_repeats
```diff
@@ -3,14 +3,10 @@ FILE=-
 EXPECT=<<EOF
 16
 18
-16
-18
 EOF
 CMDS=<<EOF
 @@?~?
 @@@?~?
-?@@?~?
```
ret2libc: these additional help commands are not necessary. Help commands follow a very strict rule which is `<command-name>?`, so having `?@@?` is not consistent.

test/new/db/cmd/feat_arithmetic
```diff
@@ -55,7 +55,7 @@ EOF
 CMDS=<<EOF
 ?v 0x10+0x30
 ?v 1+3*2
-?v (1+3)*2
+"?v (1+3)*2"
```
ret2libc: again, parenthesis are special chars, so in the new parser they should be quoted. We could also think about giving them a special meaning, like the `$((...)` in bash. But for now, let's just quote everything.

test/new/db/cmd/feat_arroba
```diff
@@ -9,29 +9,29 @@ s @ 0x42
 EOF
 RUN
 
-NAME=@ e:
+NAME=@e:
```
ret2libc: Having the parser accept `@     e:` is just additional complxity with no real benefit. The new parser just accepts `@e:` without any space.

test/new/db/cmd/cmd_open
radare: o-1 should be a valid command as well as o- 1

test/new/db/cmd/cmd_macros
radare: why. anonymous macros are nice and those don't need to be recorded anywhere. can be reword your explanation? i cant understand the reasoning

ret2libc: Try:
```
[0x00000000]> (;?v $$)()
0x0
[0x00000000]> (
0 ( ; ?v $$)
```
As you can see, the "anonymous" macro is recorded anyway. And with an empty name. But that's not entirely a problem.

The thing is that if you have a parser that look at one or two tokens ahead, you cannot easily parse `(;?v $$)` because it could be interpreted as `(` + `?v $$)` (remember that `;` is a separator of commands and `(` is a command on its own. I understand this is not a strong reason, but I think at the moment making the parser more complicated just to support "nameless" macros is not good. Having `_` as a name is just one char more to type and it ensures the macro has always a name.

test/new/db/cmd/cmd_open
ret2libc: I know that *now* it is a valid command, but I think it's not consistent. When you are in bash, for example, `/bin/ls~` is not valid. `/bin/ls ~` is valid. I think we should stick to this way, where the command name is separated from the arguments.

test/new/db/cmd/feat_arroba
radare: totally agree

test/new/db/cmd/feat_arithmetic
radare: wtf why

test/new/db/cmd/cmd_repeats
radare: agree, but it was confusing for some people was expecting to get the help of the modifiers as arguments of the ? command. but i agree

test/new/db/cmd/cmd_pd
radare: well.. use backslash then to escape that

test/new/db/cmd/cmd_pd
radare: i dont see why we should use # in here. the syntax should be rgb:002f00

radare: this should be fixed in r2, the test is wrong

test/new/db/cmd/cmd_pd
ret2libc: That's what I'm doing.

test/new/db/cmd/cmd_pd
ret2libc: I agree but I don't want to touch too much existing commands as they need to be changed anyway in the future to accept argc,argv

radare:
Why ( is a special char?

radare:
I know that, but the output should be rgb:... not #...

radare:
Parenthesis and brackets are used by rmath. I dont see any reason to make them special chars at all

>

radare:
Do this rgb change please. This is wrong

>

radare:
ping

test/new/db/cmd/cmd_afn
radare: ok

test/new/db/cmd/cmd_macros
radare: i don't see why unnamed macros are a problem because if the ( command should ignore the ; until the )  only as a command, not in the middle of the blue

test/new/db/cmd/cmd_open
radare: think about other use cases like f-* and such.. will thiss be a command or will need to be splitted as f- *. because that looks ugly to me. and i see this as a regression. i know r2 commands without spaces can be cryptic but they are so handy and easy to type.

test/new/db/cmd/cmd_pd
radare: then we must fix this in master and u should rebase

test/new/db/cmd/cmd_open
ret2libc: `f-*` is a command. But there's a difference:
- `f-*` is just `f-*`
- `o-`, on the other hand, could be `o-1` `o-2`, `o-12312321321`.

So, as you can see, `*` is just part of the command name, while the number in `o-` is just really an argument, it's not part of the cmd name

radare:
Then what about f-foo*

test/new/db/cmd/cmd_macros
ret2libc: I have already explained why, let me try to rephrase.

`(` is a command on its own, right? (it should list all registered macros)
`(;` is just the command `(` plus the command delimiter `;`.
`(; ?e hello` means `list all registered macros and then print the string "hello"`.
`(; ?e hello; ?e hi; ?v $$; pd 10; af; pdf` means `list all registered macros, print hello, print hi, show offset value; print disassembly 10, analyze function, print disassembly function`.

Now, consider that a parser (for what I know), can choose which rule a given set of tokens belong to only by looking at a couple of tokens ahead, so when you see:

`(; ?e` it is:
`<command-name> <command-delimiter> <command name>`
Now, this can be reduced only to:
`<command> <command>`
`<macro-command>` indeed MUST have a `)` at the end, so you cannot know whether there is a `)` just by looking at few tokens ahead.

On the other hand, by forcing a macro name you would have:
`<command-name> <arg> <command>`
which would match the rule for `<macro-command>` so the parser can immediately know that what's following is a macro command (whether it will contain the ending `)` or not. If it does not contain the ending `)`, then it's a parsing error).

XVilka: IMHO, `(_` is as short as `(`, so LGTM to me.

ret2libc:
> Parenthesis and brackets are used by rmath. I dont see any reason to make them special chars at all

They are special chars the same reason why they are so in bash, for example.
The problem is that if you don't consider them special chars, you could have different ways of parsing arguments based on the context. For example, right now the old "syntax" allows you to do: `?e HelloWorld)`. But then, if you are e.g. inside a macro definition, you cannot use anymore the `)` because it would be used as a terminator for the macro body.

These inconsistencies are bad and make the parsing (and the user experience) hard for no real reason.

However, we could agree that whatever is inside `(...)` is considered like a math expression, so it will be replaced as-text in the command. So if you do `?e Hello(12+3)` it will print `Hello15`.

ret2libc:
> Then what about f-foo*

`f-foo*` has no meaning. What you really mean is `f- foo*`, I don't see why we should support this hard syntax (and then again you would have to manually parse command/arguments for each different command) just to avoid typing one space. `f-` is the command name. `foo*` is the argument. I really don't see why we should make things harder just to avoid one space.

You can't do `/bin/ls/root`, for example. You do `/bin/ls /root`.

radare:
Ok for (_;

test/new/db/cmd/cmd_pd
ret2libc: I have changed the test to use `rgb:`

radare:
Too late, u just had to rebase

>

ret2libc:
So, @radare @XVilka I see there are still a couple of points in this PR, which I believe are important to agree on before we move on.  In particular, see:

- test `"?v (1+3)*2"`
- `o- 1` vs `o-1`

Please try to reply in the right thread on this PR, so we keep things clear.

radare:
Thanks for clarifying the remaining points of discussion, my  opinion on those two things:

* i dont understand why we have to quote parenthesis when they  are only  used as part of the argument.
* i understand the o- 1 space requirement, its a bit sad for me. maybe w e can  handle this like the posix  shell do. with an alias command but this will require to have a terminator character for the alias variable, which makes the oerall thing even worst,  so  i dont see a nice solution to this syntax thing either

XVilka:
@radare you have no way to know what is the argument without some parsing first, I think it's a necessary breaking change, also UNIX people already got used to quoting things. We can try to solve this later, if there will be a strong need, once the complete switch is done, imho.

ret2libc:
> i understand the o- 1 space requirement, its a bit sad for me. maybe w e can  handle this like the posix  shell do. with an alias command but this will require to have a terminator character for the alias variable, which makes the oerall thing even worst,  so  i dont see a nice solution to this syntax thing either

we need a proper way to split the command name from the arguments. The parser does not know each single command when it "tokenize" the input string (the same as bash does not now each available command in PATH when it tries to parse the input). How can the parser know `o-1` is `command_name: o-, argument: [1]` and not, for example, `command_name: o-1` or `command_name: o, argument [-1]`? After all, `-` (but most of other characters) can either be in the command name or in the argument, so splitting the command name and the arguments is impossible. The space is required to make sure the parser knows how to split those things. Moreover, it is also for consistency... How many commands support this no-space thing? I don't think a char more for the space would really hurt anybody.

> i dont understand why we have to quote parenthesis when they  are only  used as part of the argument.

I understand that you are used to the existing "syntax", but that is not well defined and it is full of quirks. Some examples:

```
> ?e Hello)
Hello)
> (foo; ?e Hello))()
# ??? how can the parser know that the first `)` is not the closing parenthesis for the macro body? You would have to escape it with `\)`. But if you do it there, you would have a big difference in how you write commands when inside a macro body and when outside
> ?e $(?e Hello))
# ??? how can the parser know that the first `)` is not the closing parenthesis for the command substitution? Or, even worse, since `)` could be part of the argument, how do you know that this it not just a malformed command? It could well be the user wanted to print `Hello))`, but he missed the closing parenthesis for the command substitution. To solve this, you would have to escape parentheses. But if you do it there, you would have a big difference in how you write commands when inside a command substitution and when outside
> ?v (12+3
# shall the parser expect a closing parenthesis? If the parser simply consider `(` and `)` as part of the argument and does not parse them, then how can it know in `?e $(?v (12+3))` that one of those `)` is actually the closing parenthesis for `$(...)` and not just part of an argument of `?v`?
```

Again, right now I think you are still thinking in terms of "each command does its own parsing", but that leads to very inconsistent behaviour across each command and whenever you change the way a command parses something, you have to change it everywhere (good luck :D).

The new parser would work mostly like bash: bash does not know what commands are available, but it does know how to split the arguments and pass those to the command. The fact that you need to have a parser that works in general, means that you have to be careful with the characters accepted in each token. Again, if `(` and `)` are just considered regular characters of an argument, a parser would have no way to know `)` is not part of the argument, but, maybe, the ending of a command substitution/macro body.

We could make the parser directly consider whatever is in `(....)` as a sort of "math substitution". So if you do `f Hello(12+3)`, the parser would identify `(12+3)`, would call RNum on that, and at the end it would execute `f Hello15`. This is something we can do and it would work the same way *everywhere*.

XVilka:
Can you please rebase on top of the current master? To solve the r2r.v issue?

ret2libc:
@radare ?

radare:
i dont agree about the parenthesis thing. you should be escaping the separator characters like you do in any other language, for example:

```
> ?e Hello)
Hello)
> (foo; ?e Hello\))()
Hello)
> ?e $(?e Hello\))
Hello)
> ?v (12+3
that's a syntax error because the r_num_math() will fail
```

The same goes with `"...\"..."` as long as " is the separator, it should be escaped. in the case of parenthesis which can be used by macros or by $() evaluations this char must be escaped.

Also, we are using $(), but...what about also supporting ${} ?

i dont want to wait too much to merge this even if the discussion is still going on unless the current command parser is changing because of this.

ret2libc:
> Also, we are using $(), but...what about also supporting ${} ?

Why? What would be the benefit? We already have two ways of doing command substitution, so I don't see a reason to add yet another one.

> you should be escaping the separator characters like you do in any other language

I am escaping the separator characters indeed! The thing is that instead of having two ways of writing a command (based on whether you already already inside a macro body/command substitution or not) you always need to escape the parenthesis. So the first command in your example would be: `> ?e Hello\)`.

Now, of course you could have different ways of parsing arguments based on whether you are inside a macro body, inside a command substitution, or elsewhere, but IMHO this complicates things (both from an implementation point of view and from the user perspective).

radare:
i dont like to ALWAYS have to escape parenthesis it makes math expressions look like crap unless you quote them which imho is totally unnecessary

ret2libc:
I agree math expressions looks bad, that's why I suggested we use `(...)` to always indicate a math expression. So, when the parser sees `(....)` it knows it has to pass whatever is inside the parentheses to RNum. That would mean, however, that you can't do things like `?e Hello)`, but you would have to escape the parenthesis `?e Hello\)`.

So, if we consider `(...)` a way to "call" the RNum, these things would work:
```
> ?v (13+4)*2
34
> ?e Hello\)
Hello)
> (foo, ?v (13+4)*2)()
34
> (bar, ?e Hello\))()
Hello)
> ?e Hello(13+4)
Hello17
> f flag_number_(2+5)
# this would define a flag named `flag_number_7`
```

This solution would make math expressions easy to type, but at the same time you would still have an easy parser no matter whether you are inside a macro-body/command-substitution/whatever.

radare:
Wtf no. This is even worst

ret2libc:
> Wtf no. This is even worst

What is worst exactly?

Anyway, ok, then it's a mess. I don't understand why you accepted the change in https://github.com/radareorg/radare2/pull/16185 .

The underneath reasons is exactly the same. If we don't consider `(` and `)` special chars, then we have to parse arguments, command names, grep_specifier differently based on the *context* we are in (e.g. in macro body, in command substitution, in the `@(from to)`).

This is a pity, because tree-sitter is able to read context-free grammars and for the context-dependent tokens you have to create special functions in an external scanner and handle the scanning yourself. Now, this is all possible of course. It just complicates the implementation of the new parser and I wonder if it is really that important/consistent to have different ways of writing commands based on the context (some chars are escaped in one case, not in other cases).

radare:
grmpf, i understand the complication of this in the treesitter side. i dont think the ( should be a special char, and imho arguments should be separated by spaces unless you pass '\ ' to it, no?

ret2libc:
> grmpf, i understand the complication of this in the treesitter side. i dont think the ( should be a special char, and imho arguments should be separated by spaces unless you pass '\ ' to it, no?

Arguments are already separated by spaces, but unless you want to change the syntax of command substitution and of macros, that doesn't really matter. The fact is that if you consider `(` and `)` as regular characters of an argument, then when the parser sees `hello)` the `)` will be just part of the argument token.

Unless you are suggesting that there is always a space before the closing parenthesis of a command substitution, of a macro body and of `@(from to)`, then I think the space thing does not really matter.

In particular, what you are saying makes sense if you want to have this kind of syntax:
```
> ?e $(?e Hello) )
> (foo;?v 3*($$+2) )()
```
Both seems a bit "weird" to me.

ret2libc:
@radare assuming Travis will be green, I think this PR can be merged. I have removed the controversed tests about `(` and `)`, so that we can keep discussing the thing, but I think this can be merged anyway as it doesn't really change anything with that regard (or in the existing parser).

PR #16216, opened 2020-03-14
Initial Implementation of R2R in C

thestr4ng3r:


binr/r2r/load.c
```diff
@@ -0,0 +1,160 @@
...
+		const char *endtoken = val + 2;
+		if (!*endtoken) {
+			eprintf ("Error: Missing opening end token after <<\n");
+			return NULL;
+		}
+		RStrBuf *buf = r_strbuf_new ("");
+		char *line = NULL;
+		size_t linesz = 0;
+		ssize_t len;
+		while ((len = getline (&line, &linesz, f)) >= 0) {
```
radare: getline wtf

radare: this thing is not portable, use r2 apis please

radare: basically r_file_slurp()  and then r_str_split() or strchr(\n). its much faster and portable

binr/r2r/load.c
```diff
@@ -0,0 +1,160 @@
...
+			(*linenum)++;
+			char *end = strstr (line, endtoken);
```
radare: end token can be only EOF. we enforced this for simplicity. if the token is not  EOF you may throw a warning like i did  in r2r.v

binr/r2r/load.c
```diff
@@ -0,0 +1,160 @@
...
+	if (!test) {
+		r_pvector_free (ret);
+		return NULL;
+	}
+
+	ut64 linenum = 0;
+	char *line = NULL;
+	size_t linesz = 0;
+	ssize_t len;
+	while ((len = getline (&line, &linesz, f)) >= 0) {
```
radare: getline is not portable

binr/r2r/load.c
```diff
@@ -0,0 +1,160 @@
...
+		if (strcmp (line, "RUN") == 0) {
+			r_pvector_push (ret, test);
+			test = r2r_cmd_test_new ();
+			if (!test) {
+				goto beach;
+			}
+			continue;
+		}
+
+#define DO_KEY_STR(key, field) \
```
radare: multiline macros are crap. get rid of this mess

binr/r2r/load.c
```diff
@@ -0,0 +1,160 @@
...
+				eprintf (LINEFMT "Warning: Duplicate key \"%s\"\n", file, linenum, key); \
+			} \
+			if (strcmp (val, "1") != 0) { \
+				eprintf (LINEFMT "Error: Invalid value for boolean key \"%s\", only \"1\" allowed.\n", file, linenum, key); \
+			} \
+			continue; \
+		}
+
+		R2R_CMD_TEST_FOREACH_RECORD(DO_KEY_STR, DO_KEY_BOOL)
+#undef DO_KEY_STR
```
radare: you cant undef or ifdef macros. only k=v defines

binr/r2r/load.c
```diff
@@ -0,0 +1,160 @@
...
+#undef DO_KEY_BOOL
+
+		eprintf (LINEFMT "Unknown key \"%s\".\n", file, linenum, line); \
```
radare: wtf is this backlash doing here

binr/r2r/load.c
thestr4ng3r: What if you have a multiline string that contains EOF?

binr/r2r/load.c
thestr4ng3r: right it's posix, not C

binr/r2r/load.c
thestr4ng3r: source?

radare:
then write the test differently, because right now none of the 15.000 tests we have in our testsuite use it. its just avoid ato have EOF in the output in the whole line of the output

radare:
source = try it yourself

thestr4ng3r:
> source = try it yourself

Screenshot from 2020-03-14 17-26-50

thestr4ng3r:
> then write the test differently, because right now none of the 15.000 tests we have in our testsuite use it. its just avoid ato have EOF in the output in the whole line of the output

That's stupid. Supporting arbitrary eof tokens is super easy.

radare:
i feel old when i realize shit like this was not working when i learned C

kazarmy:
> > then write the test differently, because right now none of the 15.000 tests we have in our testsuite use it. its just avoid ato have EOF in the output in the whole line of the output
>
> That's stupid. Supporting arbitrary eof tokens is super easy.

I don't think we want crap like:

```
CMDS=<<q
pd 1
q
```

You can argue that only stupid people would do that, and that is probably true, but there's no point in enforcing rules through convention when we can enforce them through code.

> What if you have a multiline string that contains EOF?

If there's a real use case for this, we'll find a way to support that, perhaps by redirecting to file.

thestr4ng3r:
What about checking whether EOF is part of the payload string and rejecting other tokens if it is not? Ok, maybe we should just enforce EOF for everyting for now and keep this as an option for later in case we really need it.

binr/r2r/load.c
thestr4ng3r: The alternative is only code duplication and replicating the same list of fields across multiple code parts while `R2R_CMD_TEST_FOREACH_RECORD()` allows them to be specified in a single place. So I prefer it this way.

binr/r2r/load.c
thestr4ng3r: making a mess

kazarmy:
> What about checking whether EOF is part of the payload string and rejecting other tokens if it is not?

If a test writer forgets the EOF then either r2r encounters a truly unexpected end-of-file or a test is going to fail spectacularly. I don't quite see how a 3rd option is possible.

thestr4ng3r:
The 3rd option is that you want EOF to be inside the string. For example:
```
EXPECTED=<<FOE
FACEOFF THE TAKEOFF
FOE
```

kazarmy:
Oh you mean you want to check whether the output from CMDS contains the EOF string at the end? Yeah that could work too:

```
CMDS=?e EOF
EXPECT=<<EOF
EOF
EOF
```

thestr4ng3r:
Not necessarily at the end, it could be anywhere. How should the parser know when to stop in your example? It would jus read "\n" as the expected string and error at the second EOF.

kazarmy:
It doesn't seem that it would be worth the additional complexity though.

kazarmy:
The parser doesn't error at the second EOF and just adds the previous line to what is EXPECTed (in essence reverting its previous decision). This additional complexity doesn't appear to be worth it though.

kazarmy:
previous line -> previous line(s)

kazarmy:
> previous line -> previous line(s)

```
CMDS=?e EOF\nRUN
EXPECT=<<EOF
EOF
RUN
EOF
```

Yeah you're right this requires insane amounts of complexity to process correctly.

XVilka:
Have you seen [Criterion](https://github.com/snaipe/criterion) by the way?

XVilka:
@thestr4ng3r what about creating a separate GitHub Action for this, allowing it to fail. It will allow to merge and test/improve this as early as possible.

kazarmy:
> Have you seen Criterion by the way? Not for taking as is, but for inspiration.

Just saw it but simpler systems are easier to teach to others.

radare:
Please rebase to be green

radare:
Can yous eparate the changes related to size_t from this pr?

thestr4ng3r:
They have been separated and merged already: #16221 So rebase should do the trick.

radare:
can we merge it or what's left to be done? i would prefer to not hold such big PRs for long time

thestr4ng3r:
It breaks the build on a certain shitty platform for now. But I think after I add some #ifdefs we can merge already.

XVilka:
Lol, V panics again:

```
./v self
V panic: V tool "/home/runner/work/radare2/radare2/test/src/v/cmd/tools/vself.v" could not be compiled
make[2]: *** [all] Error 1
vlib/strconv/atof.v:499:11: cannot use type `int` as type `u64` in assignment
make[1]: *** [v/v] Error 2
  497| 	if binexp > 2046 {
make: *** [src/r2r] Error 2
  498| 		if pn.negative {
  499| 			result = DOUBLE_MINUS_INFINITY
             			 ^
  500| 		}
  501| 		else {
                                               | 0x55b38a48fd82 | ./v(+0x4d82) 
                                               | 0x55b38a54a63c | ./v(+0xbf63c) 
                                               | 0x55b38a54ab80 | ./v(+0xbfb80) 
                                               | 0x55b38a555f94 | ./v(+0xcaf94) 
                                               | 0x7f885714db97 | /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xe7) 
                                               | 0x55b38a48db6a | ./v(+0x2b6a) 
Makefile:51: recipe for target 'all' failed
make[2]: Leaving directory '/home/runner/work/radare2/radare2/test/src/v'
Makefile:19: recipe for target 'v/v' failed
make[1]: Leaving directory '/home/runner/work/radare2/radare2/test/src'
Makefile:14: recipe for target 'src/r2r' failed
```

thestr4ng3r:
It knows its days are counted and tries to sabotage this pr now.

binr/r2r/r2r.c
```diff
@@ -0,0 +1,271 @@
...
+			exit (-1);
+		}
+		r_pvector_push (&workers, th);
+	}
+
+	ut64 prev_completed = UT64_MAX;
+	while (true) {
+		ut64 completed = (ut64)r_pvector_len (&state.results);
+		if (completed != prev_completed) {
+			print_state(&state, prev_completed);
```
radare: missing space before (

binr/r2r/r2r.c
```diff
@@ -0,0 +1,271 @@
...
+
+	r_th_lock_enter (state.lock);
+
+	RPVector workers;
+	r_pvector_init (&workers, NULL);
+	int i;
+	for (i = 0; i < workers_count; i++) {
+		RThread *th = r_th_new (worker_th, &state, 0);
+		if (!th) {
+			exit (-1);
```
radare: maybe worth throwing an error here if he system cant create more threads

binr/r2r/r2r.c
```diff
@@ -0,0 +1,271 @@
...
+	case R2R_TEST_TYPE_ASM:
+		// TODO
+		break;
+	case R2R_TEST_TYPE_JSON:
+		break;
+	}
+}
+
+static void print_state(R2RState *state, ut64 prev_completed) {
+	printf ("\r\x1b[2K");
```
radare: Color_RESET

binr/r2r/r2r.c
```diff
@@ -0,0 +1,271 @@
...
+		printf (" %s "Color_YELLOW"%s"Color_RESET"\n", result->test->path, name);
+		if (result->result == R2R_TEST_RESULT_FAILED || (state->verbose && result->result == R2R_TEST_RESULT_BROKEN)) {
+			print_result_diff (result);
+		}
+		free (name);
+	}
+
+	// [x/x] OK  42 BR  0 ...
+	int w = printf ("[%"PFMT64u"/%"PFMT64u"]", completed, (ut64)r_pvector_len (&state->db->tests));
+	while (w >= 0 && w < 20) {
```
radare: use r_str_pad

configure
```diff
@@ -189,21 +189,18 @@ Optional Features:
   --with-ostype          Choose OS type ( gnulinux windows darwin haiku ) (USEROSTYPE=auto)
   --with-libversion      specify different libversion (LIBVERSION=xxx)
   --without-jemalloc     build without jemalloc
-  --with-checks-level    value between 0 and 3 to enable different level of assert (see R_CHECKS_LEVEL) (R_CHECKS_LEVEL=2)
-"
-printf "
-Some influential environment variables:
+  --without-r2r          build without r2r for regression testing
```
radare: why

configure
```diff
@@ -189,21 +189,18 @@ Optional Features:
...
+  --with-checks-level    value between 0 and 3 to enable different level of assert (see R_CHECKS_LEVEL) (R_CHECKS_LEVEL=2)\n"
```
radare: why

thestr4ng3r:
It is not useful for every user, distributions would probably not want to ship a regression testing tool. It's also not a single .c file.

configure
thestr4ng3r: wat? There is no change here.

binr/r2r/r2r.c
thestr4ng3r: No, `r_str_pad()` uses a static variable and must die.

binr/r2r/r2r.c
thestr4ng3r: This is not a color reset.

radare:
then just dont call it from root's make, but adding configure flag for this looks bad to me, we had 3 testsuites already and none of them touched the configure

radare:
also rebase to be green

thestr4ng3r:
> we had 3 testsuites already and none of them touched the configure

That's because none of them were build within the same build system as r2.

Not calling it at all in root make would be super annoying for devs and inconsistent with meson. I don't see a problem with the configure flag, so it should stay as it is here.

Issue #16048, opened 2020-02-25
Windows kernel dumps loading support


XVilka:
**Is your feature request related to a problem? Please describe.**
Currently, radare has support for user-level minidumps, but nothing for loading the kernel dumps

**Describe the solution you'd like**

Something like MDMP loader, but for loading these

**Describe alternatives you've considered**

**Additional context**
See:
- https://github.com/radareorg/radare2/issues/2598
- https://github.com/0vercl0k/kdmp-parser

abcSup:
# File Structure
Header structure (Notes: there are paddings in between properties)
https://github.com/volatilityfoundation/volatility/wiki/Crash-Address-Space
https://github.com/google/rekall/blob/master/rekall-core/rekall/plugins/overlays/windows/crashdump.py

How Physical Memory Runs work
https://computer.forensikblog.de/en/2006/03/dmp-file-structure.html

List of DumpType
https://crashdmp.wordpress.com/crash-mechanism/configuration/

# Address Space
https://github.com/google/rekall/blob/master/rekall-core/rekall/plugins/addrspaces/crash.py

# Header
`PsLoadedModuleList` - LDR_LOAD_TABLE_ENTRY
`KdDebuggerDataBlock` - _KDDEBUGGER_DATA64
`Kernel Base` address = `ntoskrnl` base address which is also the first entry in `PsLoadedModuleList`

XVilka:
@abcSup since you added some basic support, could you please describe what is missing?

abcSup:
> @abcSup since you added some basic support, could you please describe what is missing?

There are a few more rich information we can extract:
- [ ] Loaded modules/drivers. `PsLoadedModuleList` which is a vaddr to `LDR_LOAD_TABLE_ENTRY` struct.
- [ ] Permission data of each memory page from `PfnDataBase` https://rayanfam.com/topics/inside-windows-page-frame-number-part1/, which is a vaddr
- [ ] Debugger data from `KdDebuggerDataBlock`, which is a vaddr to `_KDDEBUGGER_DATA64` struct
- [ ] Active processes from `PsActiveProcessHead`, which is a vaddr to `_EPROCESS.ActiveProcessLinks`

To extract those information, we need a way to translate those vaddr to physical memory address based on memory paging. Currently, the phyiscal memory pages from the dump are mapped. We should map the virtual memory space of the kernel instead of the physical memory pages. Here are a few examples that emulate the address space:
[...]

PR #16223, opened 2020-03-15
Fix all Warnings for GCC 9

thestr4ng3r:
@abcSup Can you elaborate whether the `vtop()` function should be used anywhere? It is currently dead code.

### DIFF
--- a/libr/anal/p/anal_ppc_cs.c
+++ b/libr/anal/p/anal_ppc_cs.c
@@ -67,27 +67,6 @@ static const char* cmask32(const char *mb_c, const char *me_c) {
 	return cmask;
 }
 
-static const char *getreg(struct Getarg *gop, int n) {
-	cs_insn *insn = gop->insn;
-	csh handle = gop->handle;
-
-	if (n < 0 || n >= 8) {
-		return NULL;
-	}
-	cs_ppc_op op = INSOP (n);
-	switch (op.type) {
-	case PPC_OP_REG:
-		return cs_reg_name (handle, op.reg);
-	case PPC_OP_MEM:
-		return cs_reg_name (handle, op.mem.base);
-	case PPC_OP_INVALID:
-	case PPC_OP_IMM:
-	case PPC_OP_CRX: // Condition Register field
-		return NULL;
-	}
-	return NULL;
-}
-
 static char *getarg2(struct Getarg *gop, int n, const char *setstr) {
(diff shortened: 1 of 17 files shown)


libr/anal/p/anal_ppc_cs.c
```diff
@@ -67,6 +67,7 @@ static const char* cmask32(const char *mb_c, const char *me_c) {
 	return cmask;
 }
 
+#if 0
```
radare: dont if0, just rm

libr/anal/p/anal_ppc_cs.c
```diff
@@ -223,7 +225,9 @@ static void opex(RStrBuf *buf, csh handle, cs_insn *insn) {
 #define PPCSPR(n) getspr(&gop, n)
 #define ARG(n) getarg2(&gop, n, "")
 #define ARG2(n,m) getarg2(&gop, n, m)
+#if 0
```
radare: same

libr/bin/format/dmp/dmp64.c
```diff
@@ -67,6 +67,7 @@ static int r_bin_dmp64_init_header(struct r_bin_dmp64_obj_t *obj) {
 	return true;
 }
 
+#if 0
```
radare: same

libr/include/r_vector.h
```diff
@@ -69,6 +69,9 @@ R_API void r_vector_free(RVector *vec);
 R_API RVector *r_vector_clone(RVector *vec);
 
 static inline bool r_vector_empty(const RVector *vec) {
+	if (!vec) {
```
radare: assert

abcSup:
> @abcSup Can you elaborate whether the `vtop()` function should be used anywhere? It is currently dead code.

It is currently dead code and was meant to use for vaddr to paddr address translation later on.

libr/bin/format/dmp/dmp64.c
XVilka: No, not in this case. As far I understand this code might be useful in the future, for implementing more features: https://github.com/radareorg/radare2/issues/16048

radare:
ping @thestr4ng3r

thestr4ng3r:
I've removed the dead code. In case it will be needed later it can be grabbed from the git history.

PR #16234, opened 2020-03-17
Visual bit editor now shows bits up and down ##visual

radare:
before:
IMAGE 2020-03-17 01:31:37

after:
IMAGE 2020-03-17 01:31:46


radare:
maybe we can have a key to toggle these two representations, and maybe also add colors

PR #16236, opened 2020-03-17
Add pj_r api to print raw data

radare:
not sure if it will be better to print raw data as an array of numbers like [2,34,32,159,..] and just keep that string function as is, with null terminated strings only.

another option would be to have another api for non-null terminated strings

### DIFF
--- a/libr/include/r_util/pj.h
+++ b/libr/include/r_util/pj.h
@@ -39,6 +39,8 @@ R_API PJ *pj_kd(PJ *j, const char *k, double d);
 R_API PJ *pj_kf(PJ *j, const char *k, float d);
 R_API PJ *pj_kb(PJ *j, const char *k, bool v);
 R_API PJ *pj_null(PJ *j);
+R_API PJ *pj_r(PJ *j, const unsigned char *v, size_t v_len);
+R_API PJ *pj_kr(PJ *j, const char *name, const unsigned char *v, size_t v_len);
 R_API PJ *pj_b(PJ *j, bool v);
 R_API PJ *pj_s(PJ *j, const char *k);
 R_API PJ *pj_n(PJ *j, ut64 n);
--- a/libr/include/r_util/pj.h
+++ b/libr/include/r_util/pj.h
@@ -40,7 +40,7 @@ R_API PJ *pj_kf(PJ *j, const char *k, float d);
 R_API PJ *pj_kb(PJ *j, const char *k, bool v);
 R_API PJ *pj_null(PJ *j);
 R_API PJ *pj_r(PJ *j, const unsigned char *v, size_t v_len);
-R_API PJ *pj_kr(PJ *j, const char *name, const unsigned char *v, size_t v_len);
+R_API PJ *pj_kr(PJ *j, const char *k, const unsigned char *v, size_t v_len);
 R_API PJ *pj_b(PJ *j, bool v);
 R_API PJ *pj_s(PJ *j, const char *k);
 R_API PJ *pj_n(PJ *j, ut64 n);
--- a/libr/util/pj.c
+++ b/libr/util/pj.c
@@ -194,9 +194,27 @@ R_API PJ *pj_s(PJ *j, const char *k) {
 		pj_raw (j, ek);
 		free (ek);
 	} else {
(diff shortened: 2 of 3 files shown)


ret2libc:
Why? How do you expect to use this? unit tests please?

radare:
like this:

unsigned char *foo = "\x10\x20\x00\x30\x50";
pj_kd(pj, "data", foo, 5);

ret2libc:
And what's the expected output?

Please, either create a unit test later or fill out a proper PR template so that everybody can understand.

radare:
thats what i am asking, what would you expect?

ret2libc:
> thats what i am asking, what would you expect?

I have no idea :D you proposed the change. Do you need it for something or is it "just to have it"? If it's the latter, just drop the PR imho :) If you need it for something, I guess you know what is the output you expect.

radare:
well i need it for fixing that hacky json constructed with printfs and use a generic function that serializes binary data in json

radare:
yes, when fixing some json bugs i found there's a call in zj that was doing this and making the whole thing bad

radare:
@ret2libc review again

libr/include/r_util/pj.h
```diff
@@ -39,6 +39,8 @@ R_API PJ *pj_kd(PJ *j, const char *k, double d);
 R_API PJ *pj_kf(PJ *j, const char *k, float d);
 R_API PJ *pj_kb(PJ *j, const char *k, bool v);
 R_API PJ *pj_null(PJ *j);
+R_API PJ *pj_r(PJ *j, const unsigned char *k, size_t len);
```
ret2libc: `k` seems like a very misleading name, as it is usually used for `key`. However this is a value, so i suggest you rename it as `v`.

libr/include/r_util/pj.h
```diff
@@ -39,6 +39,8 @@ R_API PJ *pj_kd(PJ *j, const char *k, double d);
...
+R_API PJ *pj_kr(PJ *j, const char *name, const unsigned char *k, size_t len);
```
ret2libc: same here. `name` is actually the key AFAIU and `k` is actually the value, so it should be `v` and `len` -> `v_len`.

libr/include/r_util/pj.h
ret2libc: and maybe `len` -> `v_len`

libr/util/pj.c
```diff
@@ -194,12 +194,28 @@ R_API PJ *pj_s(PJ *j, const char *k) {
...
+	size_t i;
+	pj_a (j);
+	for (i = 0; i < len; i++) {
+		pj_i (j, k[i]);
+	}
+	pj_end (j);
+	return j;
+}
+
+R_API PJ *pj_kr(PJ *j, const char *name, const unsigned char *k, size_t len) {
```
ret2libc: Add `r_return_val_if_fail (j && k && v, j);`

libr/util/pj.c
```diff
@@ -194,12 +194,28 @@ R_API PJ *pj_s(PJ *j, const char *k) {
...
 		free (ek);
 	} else {
-		eprintf ("damn\n");
+		eprintf ("cannot escape string\n");
 	}
 	pj_raw (j, "\"");
 	return j;
 }
 
+R_API PJ *pj_r(PJ *j, const unsigned char *k, size_t len) {
```
ret2libc: same here, add `r_return_val_if_fail (j && k, j);`

radare:
lets wait for travis and merge

libr/include/r_util/pj.h
```diff
@@ -39,6 +39,8 @@ R_API PJ *pj_kd(PJ *j, const char *k, double d);
 R_API PJ *pj_kf(PJ *j, const char *k, float d);
 R_API PJ *pj_kb(PJ *j, const char *k, bool v);
 R_API PJ *pj_null(PJ *j);
+R_API PJ *pj_r(PJ *j, const unsigned char *v, size_t v_len);
+R_API PJ *pj_kr(PJ *j, const char *name, const unsigned char *v, size_t v_len);
```
ret2libc: s/name/k

PR #16247, opened 2020-03-18
Fix dmht for glibc caused by wrong tcache offset and definition

x0urc3:

**Detailed description**

Related to PR #16239, `dmht` does not show freed tcache chunk  due to wrong tcache size and definition. 
...

**Test plan**
- Result has been compared with gdb-gef
- Pass added test

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -2936,7 +2936,7 @@ R_API int r_core_config_init(RCore *core) {
 #endif
 #if __x86_64__
 	SETI ("dbg.glibc.ma_offset", 0x000000, "Main_arena offset from his symbol");
-	SETI ("dbg.glibc.fc_offset", 0x00240, "First chunk offset from brk_start");
+	SETI ("dbg.glibc.fc_offset", 0x00280, "First chunk offset from brk_start");
 #else
 	SETI ("dbg.glibc.ma_offset", 0x1bb000, "Main_arena offset from his symbol");
 	SETI ("dbg.glibc.fc_offset", 0x148, "First chunk offset from brk_start");
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -45,6 +45,18 @@ static bool GH(is_tcache)(RCore *core) {
 	return (v > 2.25);
 }
 
+static GHT GH(tcache_chunk_size)(RCore *core, GHT brk_start) {
+	GHT sz = 0;
+
+	GH(RHeapChunk) *cnk = R_NEW0 (GH(RHeapChunk));
+	if (!cnk) {
+		return sz;
+	}
+	r_io_read_at (core->io, brk_start, (ut8 *)cnk, sizeof (GH(RHeapChunk)));
+	sz = (cnk->size >> 3) << 3; //clear chunk flag
+	return sz;
+}
+
 static void GH(update_arena_with_tc)(GH(RHeap_MallocState_tcache) *cmain_arena, MallocState *main_arena) {
 	int i = 0;
 	main_arena->mutex = cmain_arena->mutex;
@@ -663,7 +675,7 @@ static int GH(print_single_linked_list_bin)(RCore *core, MallocState *main_arena
 	bin = m_arena + offset + SZ * bin_num;
 	r_io_read_at (core->io, bin, (ut8 *)&next, SZ);
 
-	GH(get_brks)(core, &brk_start, &brk_end);
+	GH(get_brks) (core, &brk_start, &brk_end);
 	if (brk_start == GHT_MAX || brk_end == GHT_MAX) {
 		eprintf ("No Heap section\n");
 		free (cnk);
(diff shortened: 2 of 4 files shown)


x0urc3:
Hmmm. both `dmht` and test are working on my machine

XVilka:
@x0urc3 different version of glibc?

x0urc3:
> @x0urc3 different version of glibc?

Might explain it. How could I find the glibc version from Travis?

XVilka:
It's 2.23 in Travis (Ubuntu Xenial) it seems: https://repology.org/project/glibc/versions

libr/core/cconfig.c
```diff
@@ -2936,7 +2936,7 @@ R_API int r_core_config_init(RCore *core) {
 #endif
 #if __x86_64__
 	SETI ("dbg.glibc.ma_offset", 0x000000, "Main_arena offset from his symbol");
-	SETI ("dbg.glibc.fc_offset", 0x00240, "First chunk offset from brk_start");
+	SETI ("dbg.glibc.fc_offset", 0x00280, "First chunk offset from brk_start");
```
XVilka: I think this depends from the Glibc version. Ideally, it should be set in runtime, not in the compilation time.

x0urc3:
One more thing is the expected output does not follow the sequence of command
```
r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -d -c "dcu main;dmht~Tcache;20dso;dmht~Tcache" ../bins/elf/simple_malloc_x86_64 2>/dev/null
p=3
Tcache main arena @ 0x7f6d1e4b2b80
Tcache main arena @ 0x7f6d1e4b2b80
```
-`p=3` should come after the first `dmht~Tcache`

libr/core/cconfig.c
x0urc3: - The changes I made find the fc_offset during runtime and does not depend on dbg.glibc.fc_offset

x0urc3:
> It's 2.23 in Travis (Ubuntu Xenial) it seems: https://repology.org/project/glibc/versions

This caused issue because dbg.glibc.tcache should only be enabled for glibc >2.26

radare:
can you fix this?

x0urc3:
> can you fix this?

Yes. My plan as follows
- Create another PR for glibc version detection to set dbg.glibc.tcache
- Install OLD glibc without tcache for testing

radare:
resolve the conflicts

radare:
still conflicting

radare:
still conflicting

radare:
Screenshot 2020-03-20 at 12 12 09

x0urc3:
Forgot to rebase

x0urc3:
Failed one of the added test. I'm recreating a ubuntu xenial image to test locally

radare:
Screenshot 2020-03-20 at 13 30 43

x0urc3:
Could not reproduce using docker debian/xenial:latest image. Any suggestion?

radare:
Trusty?

XVilka:
Xenial is the default Travis configuration, and we do not specify the distro in YAML, so should be it.

XVilka:
Ideally, it should be tested with a sequence of `om` and `wx` commands, for different glibc versions. Relying on the runtime for testing is wrong.

radare:
ping

x0urc3:
I'm creating a test as suggested by @XVilka. May take some time

radare:
keep us updated

radare:
red

x0urc3:
> Ideally, it should be tested with a sequence of `om` and `wx` commands, for different glibc versions. Relying on the runtime for testing is wrong.

glibc struct for malloc requires around 3kB. Should I just add the memory dump to radare2-testbins rather than using `wx`?

x0urc3:
> > Ideally, it should be tested with a sequence of `om` and `wx` commands, for different glibc versions. Relying on the runtime for testing is wrong.
> 
> glibc struct for heap requires around 3kB. Should I just add the memory dump to radare2-testbins rather than using `wx`?

XVilka:
Yes, you can add it testbins of course

XVilka:
@x0urc3 can you add tests with dumps with and without tcache? And for the binary - it's probably better to include the glibc version in the name.

x0urc3:
> @x0urc3 can you add tests with dumps with and without tcache? And for the binary - it's probably better to include the glibc version in the name.

ok. will do both in this PR

XVilka:
@x0urc3 merged, please update.

XVilka:
Well done - one missing thing left:
- Do not just replace the old value - use the new only for the newer glibc
- Determine the GLIBC version from runtime, maybe https://benohead.com/blog/2015/01/28/linux-check-glibc-version/ can be helpful
- Add test for the older glibc version too

x0urc3:
> Well done - one missing thing left:
> 
> * Do not just replace the old value - use the new only for the newer glibc
I dont't understand what you mean here. Which value?

> * Determine the GLIBC version from runtime, maybe https://benohead.com/blog/2015/01/28/linux-check-glibc-version/ can be helpful
> * Add test for the older glibc version too
Can we do this in a different PR?

XVilka:
@x0urc3 the value of `SETI ("dbg.glibc.fc_offset", 0x00280, "First chunk offset from brk_start");`
>> Determine the GLIBC version from runtime, maybe https://benohead.com/blog/2015/01/28/linux-check-glibc-version/ can be helpful
>> Add test for the older glibc version too

> Can we do this in a different PR?
Yes, just please add a #ifdef for the GLIBC version meanwhile for the `dbg.glibc.fc_offset`, so it will not suddenly break for the users with older glibc versions installed.

x0urc3:
`dbg.glibc.fc_offset` does not depend on glibc version but host bit. The #ifdef to handle these host bit difference have already been in the code. This PR just reuse those

PR #16260, opened 2020-03-19
Fix minunit.h to compute macro arguments just one time

ret2libc:

**Detailed description**

Many of the minunit macros used in unit tests were using the arguments multiple time, thus in case the argument of the macro is not a variable but a function call, the function would be called two times.

**Test plan**

Existing tests should still work.


radare:
this file is not from us. is this an update from src or a hacky patch here?

ret2libc:
This file is from us. We just copied it and that's it. When @crowell got this, long time ago, it was very different from what it is now. We have already done many changes.

PR #16279, opened 2020-03-22
Update yara test

sylvainpelissier:
**Detailed description**
Previous test was not deterministic depending the version of yara.

**Test plan**
New test should be passing for all versions of yara.


test/new/db/extras/cmd/yara
```diff
@@ -1,18 +1,8 @@
 NAME=yara: scanS
```
radare: the test is named scanS on purpose

sylvainpelissier:
I have made two tests.

PR #16280, opened 2020-03-22
Don't use r2 shell commands to implemnt an.

karliss:
Avoids some of the problems with command injection.

 

- [x] I've added tests that prove my fix is effective or that my feature works (if possible) - TODO: find out if there are any relevant existing tests

**Detailed description**

For "an" command implementation use r2 api functions directly instead of building a command string and evaluating it using r2 shell.

Reduces problems with command injection.

**Test plan**

* run "an name" on an instruction where referenced address has no flag -> create flag
* run "an name" on an instruction where referenced address has  flag -> rename flag
* run "an name" on function call instruction -> rename function
...

**Closing issues**

...

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -1013,9 +1013,9 @@ static void list_vars(RCore *core, RAnalFunction *fcn, int type, const char *nam
 
 static int cmd_an(RCore *core, bool use_json, const char *name)
 {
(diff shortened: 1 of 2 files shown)


radare:
i can't merge if it's a draft

karliss:
@radare That's intentional. I wanted to look if it is missing tests before merging.

PR #16298, opened 2020-03-24
Fix cursor visibility after leaving visual graph

gur111:
It's a single line that I copied. Tested it's fixing the issue. Think it's enough

### DIFF
--- a/libr/core/agraph.c
+++ b/libr/core/agraph.c
@@ -4858,7 +4858,6 @@ R_API int r_core_visual_graph(RCore *core, RAGraph *g, RAnalFunction *_fcn, int 
 	}
 	r_config_hold_restore (hc);
 	r_config_hold_free (hc);
-	showcursor(core, true);
 	if (update_seek) {
 		return -1;
 	}
--- a/libr/core/agraph.c
+++ b/libr/core/agraph.c
@@ -4858,6 +4858,7 @@ R_API int r_core_visual_graph(RCore *core, RAGraph *g, RAnalFunction *_fcn, int 
 	}
 	r_config_hold_restore (hc);
 	r_config_hold_free (hc);
+	showcursor(core, true);
 	if (update_seek) {
 		return -1;
 	}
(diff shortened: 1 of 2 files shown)


radare:
Screenshot 2020-03-24 at 20 02 50

libr/core/agraph.c
```diff
@@ -4858,6 +4858,7 @@ R_API int r_core_visual_graph(RCore *core, RAGraph *g, RAnalFunction *_fcn, int
 	}
 	r_config_hold_restore (hc);
 	r_config_hold_free (hc);
+	showcursor(core, true);
```
radare: add missing space, and only do this if color is set

gur111: Regarding the space, there wasn't a space before either but I'll add it.
Regarding the color. I'm not sure what you mean or how to do that. I merely looked for the place where the cursor is set to hidden and after the graph view closes I added a line to make it visible (it was made invisible earlier in the function).
I can't say I understand the project, merely was able to mimic the existing logic and thought to save someone some time fixing the issue I opened a little before.

radare:
The cursor is only hidden when in visual. So you should only call this from there. This is why it’s my making the testsuite fed fail

radare:
And the space is before the (

libr/core/panels.c
```diff
@@ -7248,6 +7249,9 @@ void __panels_process(RCore *core, RPanels *panels) {
 	}
 	goto repeat;
 exit:
+    if (!originVmode) {
```
radare: use tabs, not spacs

libr/core/panels.c
```diff
@@ -6680,6 +6680,7 @@ void __panels_process(RCore *core, RPanels *panels) {
 
 	bool o_interactive = r_cons_is_interactive ();
 	r_cons_set_interactive (true);
+    r_core_visual_showcursor (core, false);
```
radare: dont indent with spacs

gur111: done

libr/core/panels.c
gur111: done

PR #16300, opened 2020-03-24
Remove the --without-r2r configure flag

radare:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


ret2libc:
I don't use configure/make, but I don't see the problem of keeping the flag. Not everybody probably wants to build r2r-c (e.g. building radare2 for packaging)

radare:
Exactly is what we want to have everywhere because we want every user to be able to write tests and run tests. Imho a 10KB extra executable doesnrt hurt. Yes, having an exrta flag doesnt hurt, but i want to reduce complexity, reduce questions, and reduce problems or different suer environments. Otherwise we will have to teach users to buildl from git to teest a. Thing if they are using a packaged version and so on. Which is a mess. Same env for everyone simplifies the problems a lot

ret2libc:
AFAIU r2r-c is enabled by default, but it can be optionally disabled. So I see only package maintainers using this, so they don't build extra stuff not needed for the distro package.

radare:
Like having an extra small .c file woul dbe aproblem for any distro.

I WANT everyone to be able to write, run and check r2, even distros, debian will patch and change everything like theyalways do, but they wil do that even if we have this flag. I know the default is building it .

Also all the binaries are implemented in libr/main except this one

>

PR #16310, opened 2020-03-25
R2R in C Enhancements

thestr4ng3r:
see commit descriptions for info.

r2r-c reads a string specified without `<<EOF` syntax like `EXPECT_ERR=Missing file.` as `"Missing file."` without a trailing newline while r2r.js doesn't seem to care at all. So I had to change these to use `<<EOF` instead.

### DIFF
--- a/binr/r2r/load.c
+++ b/binr/r2r/load.c
@@ -529,6 +529,10 @@ static bool database_load(R2RTestDatabase *db, const char *path, int depth) {
 			if (*subname == '.') {
 				continue;
 			}
+			if (strcmp (subname, "extras") == 0) {
+				// Only load "extras" dirs if explicitly specified
+				continue;
+			}
 			r_strbuf_setf (&subpath, "%s%s%s", path, R_SYS_DIR, subname);
 			if (!database_load (db, r_strbuf_get (&subpath), depth - 1)) {
 				ret = false;
--- a/binr/r2r/load.c
+++ b/binr/r2r/load.c
@@ -92,6 +92,7 @@ static char *read_string_val(char **nextline, const char *val, ut64 *linenum) {
 		r_strbuf_free (buf);
 		return NULL;
 	}
+
 	return strdup (val);
 }
 
@@ -172,9 +173,9 @@ R_API RPVector *r2r_load_cmd_test_file(const char *file) {
 				eprintf (LINEFMT "Warning: Duplicate key \"%s\"\n", file, linenum, key); \
 			} \
 			test->field.set = true; \
-			if (strcmp (val, "1") != 0) { \
+			if (strcmp (val, "1") == 0) { \
 				test->field.value = true; \
-			} else if (strcmp (val, "0") != 0) { \
+			} else if (strcmp (val, "0") == 0) { \
                 test->field.value = false; \
 			} else { \
 				eprintf (LINEFMT "Error: Invalid value \"%s\" for boolean key \"%s\", only \"1\" or \"0\" allowed.\n", file, linenum, val, key); \
--- a/binr/r2r/load.c
+++ b/binr/r2r/load.c
@@ -173,9 +173,9 @@ R_API RPVector *r2r_load_cmd_test_file(const char *file) {
 				eprintf (LINEFMT "Warning: Duplicate key \"%s\"\n", file, linenum, key); \
 			} \
 			test->field.set = true; \
-			if (strcmp (val, "1") == 0) { \
+			if (!strcmp (val, "1")) { \
 				test->field.value = true; \
-			} else if (strcmp (val, "0") == 0) { \
+			} else if (!strcmp (val, "0")) { \
                 test->field.value = false; \
 			} else { \
 				eprintf (LINEFMT "Error: Invalid value \"%s\" for boolean key \"%s\", only \"1\" or \"0\" allowed.\n", file, linenum, val, key); \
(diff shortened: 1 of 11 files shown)


test/new/db/tools/rahash2
```diff
@@ -405,7 +407,9 @@ CMDS=!rahash2 -a md5 -s "admin" -c 21232f297a57a5a743894a0e4a801fc4
 EXPECT=<<EOF
 0x00000000-0x00000004 md5: 21232f297a57a5a743894a0e4a801fc3
 EOF
-EXPECT_ERR=rahash2: Computed hash doesn't match the expected one.
+EXPECT_ERR=<<EOF
```
radare: <<EOF is not mandatory if its just one line

radare:
what about fixint r2r.c instead o changing the tests unnecessarily?

thestr4ng3r:
Because I think it makes more sense to read `EXPECT_ERR=Missing file.` as `"Missing file."` and not `"Missing file.\n"`.

In fact, I did try it out to make it append the newline, but then that doesn't make sense for keys like `FILE`, so instead of changing the semantics depending on which key it is, I think it is better to have oneliners always without a newline and `<<EOF` for everything else.

binr/r2r/load.c
```diff
@@ -172,9 +173,9 @@ R_API RPVector *r2r_load_cmd_test_file(const char *file) {
 				eprintf (LINEFMT "Warning: Duplicate key \"%s\"\n", file, linenum, key); \
 			} \
 			test->field.set = true; \
-			if (strcmp (val, "1") != 0) { \
+			if (strcmp (val, "1") == 0) { \
 				test->field.value = true; \
-			} else if (strcmp (val, "0") != 0) { \
+			} else if (strcmp (val, "0") == 0) { \
```
radare: use ! instead of ==

binr/r2r/r2r.c
```diff
@@ -179,19 +179,49 @@ static RThreadFunctionRet worker_th(RThread *th) {
...
 
 static void print_diff(const char *actual, const char *expected) {
-	// TODO: do an actual diff
+#define DO_DIFF !__WINDOWS__
+#if DO_DIFF
+	RDiff *d = r_diff_new ();
+	char *uni = r_diff_buffers_to_string (d, (const ut8 *)expected, (int)strlen (expected), (const ut8 *)actual, (int)strlen (actual));
+	r_diff_free (d);
+
+	RList *lines = r_str_split_duplist(uni, "\n");
```
radare: mssing space

binr/r2r/r2r.c
```diff
@@ -179,19 +179,49 @@ static RThreadFunctionRet worker_th(RThread *th) {
...
+	RListIter *it;
+	char *line;
+	r_list_foreach (lines, it, line) {
+		char c = *line;
+		switch (c) {
+			case '+':
```
radare: wrong indent

binr/r2r/r2r.c
```diff
@@ -179,19 +179,49 @@ static RThreadFunctionRet worker_th(RThread *th) {
...
+				printf (Color_GREEN);
```
radare: use "%s", Color_GREEN, or dont use printf

binr/r2r/load.c
```diff
@@ -529,6 +530,10 @@ static bool database_load(R2RTestDatabase *db, const char *path, int depth) {
 			if (*subname == '.') {
 				continue;
 			}
+			if (strcmp (subname, "extras") == 0) {
```
radare: use !strcmp like we use everyewher else

libr/main/rabin2.c
```diff
@@ -1173,7 +1173,7 @@ R_API int r_main_rabin2(int argc, char **argv) {
 		rabin_do_operation (bin, op, rad, output, file);
 	}
 	if (isradjson) {
-		r_cons_print ("}");
+		r_cons_print ("}\n");
```
radare: we should use pj

test/new/db/tools/r2
```diff
@@ -104,7 +104,9 @@ RUN
 NAME=no-nl-at-eof script
 FILE=-
 CMDS=!radare2 -i scripts/no-nl-at-eof.r2 -NQ -
-EXPECT=1
+EXPECT=<<EOF
```
radare: nope. dont use <<EOF for 1 char

radare:
The testsuite trims the trailing newlines and empty spaces

kazarmy:
> Because I think it makes more sense to read `EXPECT_ERR=Missing file.` as `"Missing file."` and not `"Missing file.\n"`.

r2r.js reads `EXPECT_ERR=Missing file.` as both `"Missing file."` and `"Missing file.\n"` i.e. it doesn't care about the trailing newline at all. @radare apparently prefers simplicity over strictness, and this probably isn't a big deal since everyone should whitespace-trim r2 output.

Btw, r2r.js also reads

```
EXPECT_ERR=<<EOF
Missing file.
EOF
```

as both `"Missing file."` and `"Missing file.\n"` as well.

----

And

```
EXPECT_ERR=Missing file.
```

takes less time to read and write, and less lines compared to

```
EXPECT_ERR=<<EOF
Missing file.
EOF
```

so I think it should stay.

libr/main/rabin2.c
thestr4ng3r: yes, but that is out of the scope of this pr.

radare:
cc @kazarmy

kazarmy:
Travis integration eta wen?

PR #16334, opened 2020-03-28
Super simple r_strf()

thestr4ng3r:
Extremely simple `r_strf()` proposal.
What's dangerous about this is that one might be tempted to do something like this:
```
char *mybuf = malloc(2);
r_strf(mybuf, "aaaaaaaaaaaaaaaaa");
```
Not sure if there is anyway how we could statically prevent that.

### DIFF
--- a/libr/include/r_util/r_str.h
+++ b/libr/include/r_util/r_str.h
@@ -19,6 +19,12 @@ typedef enum {
 	R_STRING_ENC_GUESS = 'g',
 } RStrEnc;
 
+#define r_strf(buf, ...) ( \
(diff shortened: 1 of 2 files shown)


radare:
I think condret didnt pushed his rstrf implementation. Who else wants to join the reinvent the wheel party?

thestr4ng3r:
Maybe we should rewrite r_strf in another language...

radare:
Java?

ret2libc:
If the purpose is to have something as simple to use as sdb_fmt then I don't think this solution is the right one. If instead we want just some wrappers around snprintf than it's ok.

Issue #414, opened 2013-12-03
Custom Charset translating commands


condret:
I'm thinking about to add some st-commands(`psT`, `psTz`, `wsT`, `/sT`, `./sT`) for supporting unconventional code-charts like this [one](http://bulbapedia.bulbagarden.net/wiki/Save_data_structure_in_Generation_I#Text_data).
Maybe you remember [this](http://runas-racer.com/foo/incredible.png). It is nasty to overwrite the original string, if you can only use `wx` for this instead of `w`.
I think that there might be a lot of files using unconventional code-charts.

Should this be a cmd-plugin?
Where should I put code-charts? (I think it would be good if they were defined in files in a place like ~/.radare2/ )

condret:
yes, /e is allready used now

gogo2464:
Hello. Can you tell me more about the syntax of `psT` please?

gogo2464:
I finished to made a new command to support gaeboy encoding.

I made an array of structs to store the custom chart encoding of the gameboy. I imagine we want more custom text encoding. Where can I store the arrays?

radare:
please don't say "chart", chart encoding is nothing, but a typo, it's "char", from "character"

also i dont see why this stuff should be in sT for several reasons:

* s is for seeking
* uppercase letters in commands shouldnt be used
* none of the letters have any relationship with encoding chars

kazarmy have been working on character encoding things for the string search functions, he can probably share some ideas in this thread

bast233:
I read your post again and I noticed `pe pokemon  # uses last selected encoding`.

gogo2464:
I noticed your command `pe pokemon  # uses last selected encoding`.

kazarmy:
I think it's fine that you want to do something like the table in https://en.wikipedia.org/wiki/Code_page_437#Character_set but for your own custom encodings. I think what @radare is saying that you should use `pst` instead of `psT` where `t` stands for table (or maybe even `te` and `pe`?). You do have to figure out suitable syntax yourself though, for both the command(s) needed and the data files. You can see the implementations of `eco` and `pfo` for example to see how commands with data files should be implemented.

Don't think this is a task for first-timers but I don't know what you're capable of so good luck!

gogo2464:
@kazarmy I want to load values from a file. Do I implement my own file format and do I make a new function to parse it or do I use an existing function if it exists?


[...]

PR #16357, opened 2020-03-29
commands pse and pst added

gogo2464:

**Detailed description**

My modifications permit to:
-list plugins
-load char encoding plugin
-use the plugin to print a string with a different char encoding

The command `pse` lists all the plugin on the specific folder for char encoding plugins.
The plugin works by using the command `k`.
Once the plugin is loaded we can convert and see encoding with `pst`.

Some new folders have been added in the project with the makefile to handle plugins.

There are still some stuff we may make better:
-change the current folder to store the plugins from `radare2/libr/enco/d/` to something like `radare2/libr/bin/e` or something different. I do not have opinion. If the road is fine, then, I have to modify the file generator of `r_userconf.h`. Currently the file `r_userconf.h` is directly modified. It works when built with `make` but not with `./sys/user.sh`
-I have to remove superfluous functions:
- `r_print_gameboy_string()`. I have removed this function but I wanted to show you it. I used to call it instead of `r_print_string()` here: https://github.com/gogo2464/radare2/blob/pse-and-pst-added-with-superflous-functions/libr/core/cmd_print.c#L5879
- the struct array `gameboy_encoding_chart` should be removed.
-the plugin commands write the sdb in the meta folder wich can be used for another purpose.

-I did not have upated the documentation of radare2 yet
-my plugin does not handle special chars and does not handle spaces. I use `-` instead

...

**Test plan**

First download the gameboy rom from https://github.com/MrCheeze/pokered-self-replicator
`radare2 sav.dat`
`pse` to list plugins
`pse pokered` to load pokemon game boy red char encoding plugin
`pst @ 0x2528` to read a string on the gameboy rom as example
You should see 

> Selfreplicating----arbitrary-code------by-MrCheeze-:)---------------------\\\\AvvJOHN\\\\]V

Then the test is OK.
...

**Closing issues**

will close https://github.com/radareorg/radare2/issues/414
...


libr/core/cmd_print.c
```diff
@@ -5852,6 +5852,93 @@ l = use_blocksize;
...
+				ut8 out[len];
+				ut8 c;
+				int i = 0;
+
+				for (int j = 0; j < len; j ++) {
+					ls_foreach (sdbls, iter, kv) {
+						char hex_str[2];
+						memcpy(hex_str, sdbkv_key (kv)+2, sizeof(char)*2);
+						int hex = (int)strtol(hex_str, NULL, 16);
+
```
radare: Thats an overkill to load a byte array. We can define those charsets in .h and set a pointer to the array

libr/enco/d/pokered
```diff
@@ -0,0 +1,62 @@
+"k anal/meta/0x50=\0"
```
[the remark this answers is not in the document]
gogo2464: @radare 

> * use “charset” rather than “encoding”
do you mean I put them into libr/util/d/charset?

> * make a simple C api to resolve that char instead of sdb
do you want that I make a new file radare2/libr/anal/charset.c

> * loading the byte=char instead of running an r2 script
do you expect I parse it with radare2/libr/anal/charset.c ?

> * what about multibyte charsets?
no idea what I can do

libr/core/cmd_print.c
```diff
@@ -5852,6 +5852,93 @@ l = use_blocksize;
...
+						if (core->block[j] == hex) {
+							c = (ut8) sdbkv_value (kv)[0];
+							out[i] = c;
+							i++;
+						} else {
+						}
+					out[i] = '\0';
```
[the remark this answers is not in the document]
gogo2464: @radare good point. I did not focus on yet.

libr/core/cmd_print.c
```diff
@@ -5852,6 +5852,93 @@ l = use_blocksize;
...
+					ls_foreach (sdbls, iter, kv) {
+						char hex_str[2];
+						memcpy(hex_str, sdbkv_key (kv)+2, sizeof(char)*2);
+						int hex = (int)strtol(hex_str, NULL, 16);
+
+						if (core->block[j] == hex) {
+							c = (ut8) sdbkv_value (kv)[0];
+							out[i] = c;
+							i++;
+						} else {
```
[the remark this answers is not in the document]
gogo2464: @radare noted

libr/Makefile
```diff
@@ -45,6 +45,9 @@ endif
 ifneq ($(LIBS8),)
 	$(MAKE) $(LIBS8)
 endif
+ifneq ($(LIBS9),)
```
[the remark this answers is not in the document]
gogo2464: @radare ok

libr/core/cmd_print.c
```diff
@@ -5852,6 +5852,93 @@ l = use_blocksize;
...
+		case 't': // "pst"
+			if (l > 0) {
+				SdbListIter *iter;
+				SdbKv *kv;
+				SdbList *sdbls = sdb_foreach_list (core->anal->sdb_meta, true);
+				ut8 out[len];
+				ut8 c;
+				int i = 0;
+
+				for (int j = 0; j < len; j ++) {
```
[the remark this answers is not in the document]
gogo2464: @radare ok it is a bad c++ reflex

libr/core/cmd_print.c
gogo2464: @radare do you just want I move char hex_str[2]; on a .h? Do you expect something more?

libr/core/cmd_print.c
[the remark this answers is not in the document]
radare: its supported by c99 too but we dont use it for backward compat reasons

libr/enco/d/pokered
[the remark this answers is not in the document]
radare: dont put those files in libr/enco, this module shouldne exist at all. imho charset encodings should be in libr/util/d

charsets are part of text processing stuff, not analysis

libr/core/cmd_print.c
radare: just fn translate(char ch) {return charset_table[ch]}

libr/enco/d/pokered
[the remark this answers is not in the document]
gogo2464: @radare  what do you mean by "make a simple C api to resolve that char instead of sdb" I can not write it in libr/enco/d/pokered it is a config file.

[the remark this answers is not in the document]
gogo2464: Do you have an example of line that should be in pokered please?

libr/util/print.c
```diff
@@ -12,6 +12,76 @@
 
 static const char hex[16] = "0123456789ABCDEF";
 
+typedef struct { unsigned char ch; unsigned char hx } chart;
+
+chart gameboy_encoding_chart[60] = {
```
condret: this is wrong, there is no such thing as "the gameboy encoding/charset". Games bring their own set, some of them are quite unique. Please do not hard code this here

libr/enco/d/pokered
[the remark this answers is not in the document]
radare: i think that we dont need to use sdb to store this information we can just expose a function in r_util to encode and decode passing a transformation function which represents the encoding. this function will take every char and convert them to ascii (or viceversa), we can hardcode those gameboy encodings and maybe other mame ones for now. and then add an api to dynamically load others via plugins.

radare:
I think that with the fact that there can be many encodings and it can be of any length from 1 byte to 4 we should handle this in a more generic way. i agree that it is mandatory to allow users to load custom charsets. and sdb is probably the easiest way to generalize this, because this way we can handle char or word from/to conversions.

we may want to specify a configuration variable like `e scr.charset` or so to tell all the print commands to use the given charset to decode before printing to the user. and same goes for the write commands to encode. maybe this way we can kill the `ww` (write wide) command for example

im moving this issue for 4.5 for proper implementation/discussion. but im happy that this is moving forward. we will need tests for this.

radare:
Maybe the api should look like: (renaming current utf* apis to have a consistent naming)

```c
r_utf8_encode_str -> r_charset_encode(const char *asciistr, R_CHARSET_UTF8);
r_pokered_encode_str -> r_charset_encode(const char *asciistr, R_CHARSET_POKERED);
```

the thing is.. we don't want to have a hardcoded encoding, so we need a way to load that dynamically, and i thnk that sdb is fine for custom ones because doing a native plugin for just trying charsets is a little bit too much maybe.

gogo2464:
@radare yes and then replace this function here https://github.com/radareorg/radare2/blob/1020cfb3cfe4ea329d30329a1921968b3a79eff9/libr/core/core.c#L2243

radare:
Yep

radare:
Ping

radare:
why you cant import?

gogo2464:
@radare I do not need to import anymore. I wrote r_charset_encode differently. This issue is solved.

radare:
Ok!

gogo2464:
I have finished to edit the function r_charset_encode_str(). It works but it requires a struct for charsets. So I will update `pse` to  load datas and fill this struct with the loaded plugin. I put all in the draft pull request.

ret2libc:
@gogo2464 Hi! Could you merge master into your branch? I see your PR is based on a quite old master. Also, please try to "marked as resolved" discussions with @radare about the PR and to fix issues like `for (int j = .....)`.

libr/core/cmd_print.c
```diff
@@ -5852,6 +5852,93 @@ l = use_blocksize;
...
+						}
+					out[i] = '\0';
+					}
+				}
+				r_print_string (core->print, core->offset, out, len, R_PRINT_STRING_ZEROEND);
+			}
+			break;
+		case 'e': // "pse"
+			if (input[2] == ' ') {//work in progress
+				const char *fname = r_str_trim_head_ro (input + 3);
```
radare, line 5884: move this block into a separate function

libr/core/cmd_print.c
```diff
@@ -5852,6 +5852,93 @@ l = use_blocksize;
...
+						} else {
+						}
+					out[i] = '\0';
+					}
+				}
+				r_print_string (core->print, core->offset, out, len, R_PRINT_STRING_ZEROEND);
+			}
+			break;
+		case 'e': // "pse"
+			if (input[2] == ' ') {//work in progress
```
radare, line 5883: also handle '?'

libr/core/cmd_print.c
```diff
@@ -5852,6 +5852,93 @@ l = use_blocksize;
...
+						if (!r_core_cmd_file (core, input + 3)) {
+							eprintf ("pse: cannot open format file at '%s'\n", path);
+						}
+					}
+				}
+				free (home);
+				free (path);
+				free (tmp);
+			} else {//not ok
+				RList *files;
```
radare, line 5912: move this into a separate function

libr/core/cmd_print.c
radare: it will be very interesting to have a subcommand to list them like `psel`.

Makefile
```diff
@@ -14,7 +14,7 @@ BUILDSEC=$(shell date -u -d "@$(SOURCE_DATE_EPOCH)" "+__%H:%M:%S" 2>/dev/null ||
 else
 BUILDSEC=$(shell date "+__%H:%M:%S")
 endif
-DATADIRS=libr/cons/d libr/flag/d libr/bin/d libr/asm/d libr/syscall/d libr/magic/d libr/anal/d
+DATADIRS=libr/cons/d libr/flag/d libr/bin/d libr/asm/d libr/syscall/d libr/magic/d libr/anal/d libr/enco/d
```
radare, line 17: imho this should be implemented in `r_util` and the data files into libr/util/d

libr/Makefile
radare, line 48: no need to add another module. just implement this in `libr/util`

libr/include/r_util/r_print.h
```diff
@@ -192,6 +192,7 @@ R_API void r_print_offset_sg(RPrint *p, ut64 off, int invert, int offseg, int se
 #define R_PRINT_STRING_WIDE32 16
 #define R_PRINT_STRING_ESC_NL 32
 R_API int r_print_string(RPrint *p, ut64 seek, const ut8 *str, int len, int options);
+R_API int r_print_gameboy_string(RPrint *p, ut64 seek, const ut8 *buf, int len, int options);
```
radare, line 195: imho this shouldnt be a different function. We can hold the selected encoder inside RPrint and just call the decoder function before printing it in `r_print_string`.

The `pse` command can be replaced with a config variable that works globally. For temporal testing encodings we can use the `ps@e:cfg.encoding=pokemon`. (im proposing cfg.encoding, but it can be anything else

libr/util/print.c
```diff
@@ -701,6 +771,74 @@ R_API int r_print_string(RPrint *p, ut64 seek, const ut8 *buf, int len, int opti
 	return i;
 }
 
+R_API int r_print_gameboy_string(RPrint *p, ut64 seek, const ut8 *buf, int len, int options) {
```
radare, line 774: this function should be encode/decode from/to utf8/ascii. this way we can reuse the rest of functions and commands in r2 to process the given string

libr/util/print.c
```diff
@@ -12,6 +12,76 @@
 
 static const char hex[16] = "0123456789ABCDEF";
 
+typedef struct { unsigned char ch; unsigned char hx } chart;
```
radare, line 15: use ut8 instead of unsigned char. and fix the indentation

libr/enco/d/Makefile
```diff
@@ -0,0 +1,58 @@
+include ../../../global.mk
```
radare, line 1: rm

gogo2464:
@radare, @condret, @ret2libc I changed the name of the branch. The one and only branch that you should care is here: https://github.com/radareorg/radare2/pull/16621.

PR #16365, opened 2020-03-30
Implement automatic test directory for r2r

radare:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/binr/r2r/r2r.c
+++ b/binr/r2r/r2r.c
@@ -1,7 +1,6 @@
 /* radare - LGPL - Copyright 2020 - thestr4ng3r */
 
 #include "r2r.h"
-
 #include <r_cons.h>
 
 #define WORKERS_DEFAULT        8
@@ -83,7 +82,9 @@ static bool r2r_chdir_fromtest(const char *test_path) {
 			*last_slash = 0;
 		}
 	}
-	(void)chdir (abs_test_path);
+	if (chdir (abs_test_path) == -1) {
+		return false;
+	}
 	bool found = false;
 	char *cwd = NULL;
 	char *old_cwd = NULL;
@@ -129,7 +130,7 @@ int main(int argc, char **argv) {
 		case 'v':
 			verbose = true;
 			break;
-		case 'j': {
+		case 'j':
 			workers_count = atoi (opt.arg);
 			if (workers_count <= 0) {
 				eprintf ("Invalid thread count\n");
@@ -152,13 +153,13 @@ int main(int argc, char **argv) {
 			free (json_test_file);
 			json_test_file = strdup (opt.arg);
 			break;
-		}
 		default:
 			ret = help (false);
 			goto beach;
 		}
 	}
 
+	char *cwd = r_sys_getdir ();
 	if (r2r_dir) {
 		chdir (r2r_dir);
 	} else {
@@ -202,11 +203,14 @@ int main(int argc, char **argv) {
 		// Manually specified path(s)
 		int i;
 		for (i = opt.ind; i < argc; i++) {
-			if (!r2r_test_database_load (state.db, argv[i])) {
-				eprintf ("Failed to load tests from \"%s\"\n", argv[i]);
+			char *tf = (argv[i][0] == '/')? strdup (argv[i]): r_str_newf ("%s"R_SYS_DIR"%s", cwd, argv[i]);
+			if (!r2r_test_database_load (state.db, tf)) {
+				eprintf ("Failed to load tests from \"%s\"\n", tf);
 				r2r_test_database_free (state.db);
+				free (tf);
 				return -1;
 			}
+			free (tf);
 		}
 	} else {
 		// Default db path
@@ -216,7 +220,7 @@ int main(int argc, char **argv) {
 			return -1;
 		}
 	}
-
+	R_FREE (cwd);
 	r_pvector_insert_range (&state.queue, 0, state.db->tests.v.a, r_pvector_len (&state.db->tests));
 
 	bool jq_available = r2r_check_jq_available ();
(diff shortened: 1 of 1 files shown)


binr/r2r/r2r.c
```diff
@@ -41,23 +41,45 @@ static int help(bool verbose) {
...
 		"\n"
 		"OS/Arch for archos tests: "R2R_ARCH_OS"\n");
 	}
 	return 1;
 }
 
+static void r2r_chdir() {
+#if __UNIX__
+	char src_path[PATH_MAX];
+	char *r2r_path = r_file_path ("r2r");
```
thestr4ng3r: main's `argv[0]` should probably be used instead.

binr/r2r/r2r.c
```diff
@@ -41,23 +41,78 @@ static int help(bool verbose) {
...
+	char *r2r_path = r_file_path (argv0);
+	bool found = false;
+	if (readlink (r2r_path, src_path, sizeof (src_path)) != -1) {
+		char *p = strstr (src_path, R_SYS_DIR "binr"R_SYS_DIR"r2r"R_SYS_DIR"r2r");
+		if (p) {
+			*p = 0;
+			strcat (src_path, R_SYS_DIR"test"R_SYS_DIR"new");
+			if (r_file_is_directory (src_path)) {
+				(void)chdir (src_path);
+				eprintf ("Running from %s\n", src_path);
```
thestr4ng3r: We should print in all cases when the cwd was changed from the original one. Otherwise it will be confusing.

binr/r2r/r2r.c
```diff
@@ -41,23 +41,78 @@ static int help(bool verbose) {
...
+				found = true;
+			}
+		}
+	}
+	free (r2r_path);
+	return found;
+#else
+#	warning r2r_chdir not yet supported on Windows
```
thestr4ng3r: I don't think we need a warning, I doubt this will ever be supported on windows.

binr/r2r/r2r.c
```diff
@@ -41,23 +41,78 @@ static int help(bool verbose) {
...
+		char *last_slash = (char *)r_str_lchr (abs_test_path, R_SYS_DIR[0]);
+		if (last_slash) {
+			*last_slash = 0;
+		}
+	}
+	(void)chdir (abs_test_path);
+	bool found = false;
+	while (true) {
+		char *cwd = r_sys_getdir ();
+		if (!strcmp (cwd, R_SYS_DIR)) {
```
thestr4ng3r: This will probably create an infinite loop on windows. Do fixpoint iteration or something.

binr/r2r/r2r.c
```diff
@@ -94,6 +152,18 @@ int main(int argc, char **argv) {
 		}
 	}
 
+	if (r2r_dir) {
```
thestr4ng3r: You have to fix all paths given as args because if they are relative, they will be broken after cd'ing, making them absolute will probably be enough.

binr/r2r/r2r.c
radare: wat is fixpoint iteration?

binr/r2r/r2r.c
radare: then we need to chdir multiple times and chdir-back to the original one after each path assuming the user can pass multiple test files from different working directories

binr/r2r/r2r.c
radare: done

radare:
pls review again

binr/r2r/r2r.c
thestr4ng3r: I think one run-path for all given paths is enough. But as it is right now, you will be unable to load the tests at all.

binr/r2r/r2r.c
```diff
@@ -41,23 +41,88 @@ static int help(bool verbose) {
...
+		if (last_slash) {
+			*last_slash = 0;
+		}
+	}
+	(void)chdir (abs_test_path);
+	bool found = false;
+	char *cwd = NULL;
+	while (true) {
+		cwd = r_sys_getdir ();
+		if (r_file_is_root (cwd)) {
```
thestr4ng3r: I wouldn't check for root at all. If chdir fails for whatever reason you will still loop forever. Just do this instead:
```c
char *last_cwd = NULL;

while (true) {
    last_cwd = cwd;
    cwd = r_sys_getdir();
    if (last_cwd && !strcmp (last_cwd, cwd)) {
        free (last_cwd);
        free (cwd);
        break;
    }
    free (last_cwd);

....
```

radare:
done

binr/r2r/r2r.c
thestr4ng3r: What is implemented now, looping until the value does not change anymore.

thestr4ng3r:
The relative path issue is still not fixed, other than that lgtm.

radare:
which relative path issue? passing a test in relative works fine

thestr4ng3r:
Screenshot from 2020-03-30 20-01-34
It takes the relative path strings given to load the database. But after chdir'ing, you are somewhere else and the relative paths don't work anymore.

radare:
Ah ok got it didnt tried with directories

radare:
fixed:
```
pmb:~ pancake$ r2r prg/radare2/test/new/db/cmd/cmd_env
Running from /Users/pancake/prg/radare2/test/new
[3/3]                       3 OK         0 BR        0 XX        0 FX
Finished in 0 seconds.
pmb:~ pancake$ r2r $PWD/prg/radare2/test/new/db/cmd/cmd_env
Running from /Users/pancake/prg/radare2/test/new
[3/3]                       3 OK         0 BR        0 XX        0 FX
Finished in 0 seconds.
pmb:~ pancake$
```

PR #16369, opened 2020-03-30
Tweak arm64 ldr ESIL for var access ##anal

mrmacete:
Tweak the esil expression for arm64 LDR in a way that makes the local variable access detection work.

**Detailed description**

This fixes an issue with detection of local variable usages.

Disasm before:

```
str wzr, [sp + var_c40h]
ldr x8, [sp, 0x3a8]
```

Disasm after:

```
str wzr, [sp + var_c40h]
ldr x8, [sp + var_c08h]
```

The reason is very unfortunate. Apparently the `extract_arg()` function in var.c assumes a specific structure of the ESIL code (https://github.com/radareorg/radare2/blob/master/libr/anal/var.c#L655).

Therefore, this pull request just tweaks the ESIL code of the LDR variant of interest in order to conform the ESIL expression to the expected structure, even if it was already mathematically correct.

Generated ESIL before:

```
sp,936,+,DUP,tmp,=,[8],x8,=,
```

Generated ESIL after:

```
936,sp,+,DUP,tmp,=,[8],x8,=,
```

To summarise, this doesn't solve the root cause, but it's a fairly easy fix with no impact at all, and spares people a lot of headache when reversing complex functions with a lot of local variables in arm64.

### DIFF
--- a/libr/anal/p/anal_arm_cs.c
+++ b/libr/anal/p/anal_arm_cs.c
@@ -1086,8 +1086,8 @@ static int analop64_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int l
 					r_strbuf_appendf (&op->esil, "%"PFMT64d",%s,-,DUP,tmp,=,[%d],%s,=,",
 							-(int)MEMDISP64(1), MEMBASE64(1), size, REG64(0));
 				} else {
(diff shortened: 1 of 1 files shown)


radare:
Lets merge for now, because i assume it will be helpful but ideally we may have a test and the root cause fixed.

Thanks!!

PR #16371, opened 2020-03-30
Add Timeout to R2R

thestr4ng3r:


binr/r2r/run.c
```diff
@@ -274,8 +281,16 @@ R_API R2RSubprocess *r2r_subprocess_start(
 	return NULL;
 }
 
-R_API void r2r_subprocess_wait(R2RSubprocess *proc) {
-	int r;
+R_API bool r2r_subprocess_wait(R2RSubprocess *proc, ut64 timeout_ms) {
+	struct timespec timeout_abs;
+	if (timeout_ms != UT64_MAX) {
+		clock_gettime(CLOCK_MONOTONIC, &timeout_abs);
```
radare: missing space

binr/r2r/run.c
```diff
@@ -274,8 +281,16 @@ R_API R2RSubprocess *r2r_subprocess_start(
...
+		timeout_abs.tv_nsec += timeout_ms * 1000000;
```
radare: maybe use a constant define instead of 1000000

binr/r2r/run.c
```diff
@@ -302,15 +317,32 @@ R_API void r2r_subprocess_wait(R2RSubprocess *proc) {
...
+		struct timeval *timeout = NULL;
+		if(timeout_ms != UT64_MAX)
+		{
+			struct timespec now;
+			clock_gettime(CLOCK_MONOTONIC, &now);
+			st64 usec_diff = ((st64)timeout_abs.tv_sec - now.tv_sec) * 1000000 + ((st64)timeout_abs.tv_nsec - now.tv_nsec) / 1000;
+			if (usec_diff <= 0) {
+				break;
+			}
+			timeout_s.tv_sec = usec_diff / 1000000;
```
radare: same here

binr/r2r/run.c
```diff
@@ -302,15 +317,32 @@ R_API void r2r_subprocess_wait(R2RSubprocess *proc) {
 			}
 		}
 		nfds++;
-		r = select (nfds, &rfds, NULL, NULL, NULL);
+
+		struct timeval timeout_s;
+		struct timeval *timeout = NULL;
+		if(timeout_ms != UT64_MAX)
```
radare: missing sace and open brace {

binr/r2r/run.c
```diff
@@ -334,12 +367,21 @@ R_API void r2r_subprocess_wait(R2RSubprocess *proc) {
...
+		}
 	}
 	if (r < 0) {
 		perror ("select");
 	}
+	return child_dead;
+}
+
+R_API void r2r_subprocess_kill(R2RSubprocess *proc) {
+	kill (proc->pid, SIGKILL);
```
radare: #if __UNIX__

radare: or use r_sys_kill

binr/r2r/run.c
ret2libc: yes please

ret2libc: even better, some define like `MS_TO_NSEC` and `NSEC_TO_SEC`

thestr4ng3r:
I was way too tired when I wrote this yesterday, which is also why I made it a draft.

XVilka:
I suggest then the new status for such tests - not `[XX]` but `[TM]`, so it will be pretty clear from the logs.

radare:
this is very important to support the fuzz testsuite, also add a commandline flag to specify the timeout value (maybe -t)

radare:
what is TM? trademark?

radare:
i prefer XX+ TM makes no sense

XVilka:
You have to make a distinction between failed and timed out tests

thestr4ng3r:
It should be enough to print it as timeout when showing the diff and then count it as XX.

radare:
Agree with florian

radare:
But i agree about the usefulness of having TM for timeouted tests

binr/r2r/run.c
thestr4ng3r: This is already in the `#else` branch of `#if __WINDOWS__` where tons of unix api is used so should be fine.

thestr4ng3r:
This is what they look like now:
Screenshot from 2020-04-02 11-22-15

radare:
i think 60s timeout is not enough for some fuzz tests but we can tweak it later when the fuzz tests run from here

thestr4ng3r:
It's apparently not even enough for some regular tests.

radare:
conflicts

binr/r2r/run.c
```diff
@@ -2,6 +2,11 @@
 
 #include "r2r.h"
 
+#define NSEC_PER_SEC  1000000000
```
radare, line 5: so many things that shouldnt be in r2r but in util :______

PR #16381, opened 2020-03-31
Fix #16347: o+ sets maps as writable like oo+

### LINKED ISSUE #16347: o+ does not set maps writeable
Screenshot from 2020-03-29 15-27-13

The first `om` result (after `o+`) should be the same as the second (after `oo+`) and the first write should succeed.

kazarmy:

**Detailed description**

`o+` now sets maps as writable like `oo+` (https://github.com/radareorg/radare2/commit/dee13edd96162248e5ebbb4b447448691897cc9f#diff-9423a86a7236062c0e9fbbf805a079cc). Code is from `oo+`.

**Test plan**

Travis and AppVeyor are green with attached test.

**Closing issues**

Closes #16347.


libr/core/cmd_open.c
```diff
@@ -1422,6 +1423,15 @@ static int cmd_open(void *data, const char *input) {
 					addr = UT64_MAX;
 				}
 				r_core_bin_load (core, argv0, addr);
+				if (*input == '+') { // "o+"
+					SdbListIter *iter;
+					RIOMap *map;
+					ls_foreach_prev (core->io->maps, iter, map) {
+						if (map->fd == fd) {
+							map->perm |= R_PERM_WX;
```
radare: if the underlying fd doesnt permit W this will not be correct

kazarmy: ok bcbec03e5cb25a72264e8ac7c702d9d38ffe185c ... btw, tried this with a `chmod -w` bin and the error actually happened in `r_core_file_open()`, but the extra check appears harmless

PR #16382, opened 2020-03-31
hotfix of recursion stack overflow in Vbc/

HoundThe:

I've noticed a segfault due StackOverflow because of a recursion https://github.com/radareorg/radare2/blob/master/libr/core/vmenus.c#L1138 

I was digging around it and then saw that Pancake already hot-fixed part of this 9 days ago, but only the part for classes, the methods and fields still had the segfaulting recursion, so I've mimicked the fix for now.

To replicate, just open a binary that has an object and a some methods to them, then just
```
Vbc
// pick a class to show methods
// grep anything that matches nothing, for example:
/asdasdas 
```


radare:
thanks, its not even the right fix but at least we short the path earlier

Issue #15109, opened 2019-09-24
Integrate `join`, `sort`, `uniq` commands with the RTable API


XVilka:
We have `sort`, `join`, `uniq` commands for the files, but would be awesome to be able to apply them to the RTable data too.

08A:
I understand the sort command, but for join and uniq how do you want the user to use it?

radare:
thats something to discuss and decide. the sort operation is already implemented, but join and uniq are not.

uniq should be easy to be implemented so you can start in there, for join we will need to think further on how to import other commands' table into one table which is a bit more complex to discuss.

radare:
sort is implemented already, and join is more or less implemented in my comma PR.

radare:
The comma (nd) allows to load multiple table outputs into a single table, so it can be postprocessed in a handy way

radare:
So imho this pr can be closed when https://github.com/radareorg/radare2/pull/16824 gets merged

XVilka:
See https://github.com/radareorg/radare2/pull/17701

trufae:
join is done in the other pr of me implementing the, command

Issue #13763, opened 2019-04-15
Support CSV


radare:
CSV is a pretty nice format for those who like to use SQL or Excel to process the data. Those tools dont support JSON, so having CSV will give us the ability to extend the table widget.

* Import a CSV to render a table (adjust column sizes)
* Export any relevant info from r2 into CSV
* Create a CSV api like the pj_ one we have (maybe reuse the future table api) and create and exporter
* We need to define which char to use to print in that format, imho it should be a comma :D
* Columnize as a RCons filter (using the table api (_from_csv_string()))

For example:

```
afl, > functions.csv # print function list in csv format
```

Stuff that may be interesting as CSV:

* Function list
* Function Instructions
* Symbols
* Sections
* Imports
* Diffing results
* Zignatures
* ..

XVilka:
Will overlap with `CC,` for linking the text file as a comment.

radare:
This is supported by the RTable api, so we just need to make more use of it and improve the support for creating custom tables, joining them, filter, etc

radare:
we can also use aflt, but this should be supported in other commands like `i` ones

trufae:
Done via tables in 7df28e1fb253fa18d4ae9a544a1ddf65123d4c5f

PR #16385, opened 2020-03-31
Implement  r_table_uniq

08A:

**Detailed description**
prototype the function  r_table_uniq

**Test plan**
Run unit_test

**Work on issues**
#15109


libr/util/table.c
```diff
@@ -598,14 +598,70 @@ R_API void r_table_sortlen(RTable *t, int nth, bool dec) {
...
+
+		tmp = item_col->type->cmp (item_lhs, item_rhs);
+
+		if (tmp)
+			return tmp;
+	}
+
+	if (iter_lhs) {
+		return 1;
+	} else if (iter_rhs) {
```
radare: no need to else after return

libr/util/table.c
```diff
@@ -598,14 +598,70 @@ R_API void r_table_sortlen(RTable *t, int nth, bool dec) {
...
+		iter_lhs && iter_rhs && iter_col;
+		iter_lhs = iter_lhs->n, iter_rhs = iter_rhs->n, iter_col = iter_col->n) {
+
+		item_lhs = iter_lhs->data;
+		item_rhs = iter_rhs->data;
+		item_col = iter_col->data;
+
+		tmp = item_col->type->cmp (item_lhs, item_rhs);
+
+		if (tmp)
```
radare: missing braces

test/unit/test_table.c
```diff
@@ -106,15 +106,58 @@ bool test_r_table_sort1(void) {
...
+	r_table_add_row (t, "c", "99", NULL);
+	r_table_add_row (t, "b", "98", NULL);
+	r_table_add_row (t, "c", "99", NULL);
+	r_table_add_row (t, "d", "99", NULL);
+	r_table_add_row (t, "b", "98", NULL);
+	r_table_add_row (t, "d", "99", NULL);
+	r_table_add_row (t, "c", "99", NULL);
+	r_table_add_row (t, "c", "100", NULL);
+
+	r_table_uniq (t);
```
radare: imho this API should specify the column to be "uniq" instead of comparing them all. and there must be a r_table_query syntax for this. ideas here?

XVilka:
See also https://github.com/radareorg/radare2/blob/master/libr/util/syscmd.c for more

radare:
also see https://github.com/radareorg/radare2/issues/13763 after this pr gets merged i you are interested in contributing to the table api. thanks!

08A:
I understand your point of view for uniq but the main problem is wich row should be selected?

```
a | b | c
a | c | d
```

if it is selection on the first column
```
a | b | c
```
or 
```
a | c | d
```

radare:
maybe it should just take the first or do: 

```
a | b c | c d
```

not sure which would be the expected behaviour here.

08A:
Maybe we could have a prototype like that
```
R_API void r_table_uniq(RTable *t, int column_number, RSelection function);
```

if column_number < 0 we compare all column, if column_number > 0 we compare the nth column.

The selection function return an int. If the result of the selection is < 0 we keep the first row, > 0 the second and == 0 we free two row.

selection function prototype:
```
int (*foo)(int (column_numbre), void *(first_row), void *(second_row))
```

The role of selection function is select the row between 2 matches. The selection function could help merge two row.

radare:
Comments on this PR? @XVilka @ret2libc @thestr4ng3r

thestr4ng3r:
If a column number should be specified, the function should be called `group` (like in SQL) and I agree with the prototype `...(RTable *t, int column_number, RSelection function)` for this. Imo both could be useful so I am ok with the `uniq` as it is here.

radare:
Please do the query syntax and ill merge. i think the syntax can be like this:

```
aflt col0/uniq # get the first row of each that col0 is unique
aflt col0/uniq/col1 # get the first row where col0 and col1 are unique
aflt /uniq # same as | uniq (match all columns)
```

test/unit/test_table.c
ret2libc: mmm, this API seems like is the "DISTINCT" in sql, right? if so, I think it is alright to not have any column.

What you mean is  something like "GROUP BY" in  sql, I guess.

ret2libc:
As I said elsewhere before reading @thestr4ng3r 's message, I agree with him that `_uniq` as it is here is already good. Having a list of columns would make the function `group_by`. Anyway, this LGTM.

radare:
It needs to be integrated with rtable.query before being aproved

libr/include/r_util/r_table.h
```diff
@@ -2,6 +2,7 @@
 #define R_UTIL_TABLE_H
 
 #include <r_util.h>
+typedef int (*RTableSelector)(void *a, void *b, int nth);
```
ret2libc, line 5: I think `a` and `b` can be `RList`

test/unit/test_table.c
```diff
@@ -106,15 +106,163 @@ bool test_r_table_sort1(void) {
...
+	free (stri);
+	r_table_free (t);
+	mu_end;
+}
+
+static int delete_first(void *a, void *b, int nth) {
+	return -1;
+}
+
+static int merge_second(void *a, void *b, int nth) {
```
ret2libc, line 156: it should be made clearer what is the "accumulator" element. Is it `a` or `b`?

libr/util/table.c
```diff
@@ -598,14 +598,98 @@ R_API void r_table_sortlen(RTable *t, int nth, bool dec) {
...
+	r_table_group (t, -1, NULL);
+}
+
+static void select_and_delete_the_row(RList *rows, RTableRow *uniq_row, RTableRow *row,
+	RListIter *iter_inner, RListIter *iter, RTableSelector fcn, int nth) {
+	int select_flag;
+	if (!fcn) {
+		r_list_delete (rows, iter);
+	} else {
+		select_flag = fcn (uniq_row->items, row->items, nth);
```
ret2libc, line 661: why can `fcn` choose which element to delete? shouldn't it just merge the two rows together? I was thinking about a merge function that has an `accumulator` argument and a `new_row` argument.

XVilka:
@08A please rebase on top of the master so there will be no r2r.v-related problems on CI side.

test/unit/test_table.c
```diff
@@ -106,15 +106,157 @@ bool test_r_table_sort1(void) {
...
+		"c     99\n"
+		"d     99\n"
+		"c     100\n",
+		"uniq delete some rows");
+	free (stri);
+	r_table_free (t);
+	mu_end;
+}
+
+static void merge_second(RList *acc, RList *new_row, int nth) {
```
ret2libc, line 152: maybe even better to directly pass `RTableRow` instead of just the `items` part.

ret2libc: also, i'm not sure `nth` is really necessary as I can imagine when you implement the merge function you already know the column(s) you want to work on.

08A: ```nth``` could help to create reusable function.
For example a common merge function with a default separator.

08A:
> aflt col0/uniq/col1 # get the first row where col0 and col1 are unique

With the actual prototype of r_table_uniq and r_table_group that seems to be impossible.

libr/util/table.c
```diff
@@ -811,7 +884,7 @@ R_API bool r_table_query(RTable *t, const char *q) {
 		if (col == -1) {
 			if (*columnName == '[') {
 				col = atoi (columnName + 1);
-			} else {
+			} else if (columnName == NULL && strcmp(operation, "uniq")) {
```
radare, line 887: missing space before (

radare:
Amazing! thanks

Issue #16387, opened 2020-03-31
DEX loading is slow


radare:
Optimizations to speedup the dex loading:

* [ ] Use r_buf_data once and then r_read APIs to avoid all the rbuf backend messing and unnecessary memory copies.
* [ ] Cache the result of r_buf_data() to speedup consecutive calls
* [ ] Use buffer backend instead of the io one for RBuffer from RIO
* [ ] Use the RBuffer everywhere. maybe worth adding an uleb methods in rbuffer for dwarf and dex
* [ ] Avoid strdup() in getstr() this duplicates lot of elements in memory

Difference is this patch:
```
--- a/libr/bin/bin.c
+++ b/libr/bin/bin.c
@@ -280,7 +280,8 @@ R_API bool r_bin_open_buf(RBin *bin, RBuffer *buf, RBinOptions *opt) {
                }
        }
        if (!bf) {
-               bf = r_bin_file_new_from_buffer (bin, bin->file, buf, bin->rawstr,
+               RBuffer *b = r_buf_new_slurp (bin->file);
+               bf = r_bin_file_new_from_buffer (bin, bin->file, b, bin->rawstr,
                        opt->baseaddr, opt->loadaddr, opt->fd, opt->pluginname);
                if (!bf) {
                        return false;
```

<img width="328" alt="Screenshot 2020-03-31 at 23 41 01" src="https://user-images.githubusercontent.com/917142/78077752-3d19f500-73a9-11ea-9fd0-beac8f7bc400.png">

radare:
This is because the IO buffer backend plugin is 2x slower than the byte buffer one and the whole parsing depends on RBuffer. 20s is still too much and the dex parser should do bigger reads and then r_read endian APIs instead of depending so much with RBuffer.

radare:
Another optimization reduces the time to 14s

```
$ git diff buf.c
diff --git a/libr/util/buf.c b/libr/util/buf.c
index 71e1ad632..20a69e64a 100644
--- a/libr/util/buf.c
+++ b/libr/util/buf.c
@@ -65,6 +65,9 @@ static ut8 *get_whole_buf(RBuffer *b, ut64 *sz) {
        if (bsz == UT64_MAX) {
                return NULL;
        }
+       if (b->whole_buf) {
+               return b->whole_buf;
+       }
        free (b->whole_buf);
        b->whole_buf = R_NEWS (ut8, bsz);
        if (!b->whole_buf) {
```

radare:
[...]

PR #16388, opened 2020-03-31
Do not use r_buf_data in DEX results in 1.5x faster parsing (22s vs 33s) ##bin

radare:
* Fix two memleaks in the dex parser

 

**Detailed description**

Loading DEX files is slow for several reasons, most of them described, with their proposed solutions in https://github.com/radareorg/radare2/issues/16387

This PR just make RBuffer.data() cache the contents to avoid having to re-read, alloc and copy everytime this is called, which in the case of DEX is a lot of times. I know its a bad practice to use this API, and even more when used with the IO backend. but there are at least 3 bin plugins using it right now and i think it's worth the simple fix.

As I said, it will be good to add a new API `r_buf_data_cache` and use it with this behaviour to avoid surprises in the case the backend changes the data and the frontend can't be updated because of the non-invalidated cache. This situation can happen when using the IO backend for RBuffer because the RBuffer is not the only api that can be used to modify the underlying memory contents.

**Test plan**

Grab this dex [classes.dex.zip](https://github.com/radareorg/radare2/files/4414087/classes.dex.zip)

And run that line
```
time r2 -zqcq classes.dex
```

**Closing issues**

Fixes this issue https://github.com/radareorg/radare2/issues/16387

### DIFF
--- a/libr/bin/p/bin_dex.c
+++ b/libr/bin/p/bin_dex.c
@@ -1100,7 +1100,7 @@ static char *dex_class_super_name(RBinDexObj *bin, RBinDexClass *c) {
 static ut64 peek_uleb(RBuffer *b, bool *err, size_t *nn) {
 	ut64 n = UT64_MAX;
 	int len = r_buf_uleb128 (b, &n);
-	if (len < 0) {
+	if (len < 1) {
 		if (err) {
 			*err |= true;
 		}
@@ -1119,11 +1119,11 @@ static size_t parse_dex_class_fields(RBinFile *bf, RBinDexClass *c, RBinClass *c
 	int total, tid;
 	DexField field;
 	const char* type_str;
-	size_t i, skip = 0;
+	size_t i, skip = 0, skipped = 0;
 
-	r_buf_seek (bf->buf, offset, R_BUF_SET);
 	for (i = 0; i < fields_count; i++) {
 		bool err = false;
+		r_buf_seek (bf->buf, offset + skip, R_BUF_SET);
 		ut64 fieldIndex = peek_uleb (bf->buf, &err, &skip);
 		ut64 accessFlags = peek_uleb (bf->buf, &err, &skip);
 		if (err) {
@@ -1208,23 +1208,27 @@ static size_t parse_dex_class_method(RBinFile *bf, RBinDexClass *c, RBinClass *c
 		DM = 4096;
 	}
 	size_t skip = 0;
+	size_t skipped = 0;
 	ut64 bufsz = r_buf_size (bf->buf);
-	r_buf_seek (bf->buf, offset, R_BUF_SET);
+	ut64 encoded_method_addr;
+	bool err = false;
+	ut64 MI, MA, MC;
 	for (i = 0; i < DM; i++) {
-		ut64 encoded_method_addr = offset; //  + skip;
-		bool err = false;
+		err = false;
 		skip = 0;
+		encoded_method_addr = offset; //  + skip;
 		r_buf_seek (bf->buf, offset, R_BUF_SET);
-		ut64 MI = peek_uleb (bf->buf, &err, &skip);
+		MI = peek_uleb (bf->buf, &err, &skip);
 		MI += omi;
 		omi = MI;
-		ut64 MA = peek_uleb (bf->buf, &err, &skip);
-		ut64 MC = peek_uleb (bf->buf, &err, &skip);
+		MA = peek_uleb (bf->buf, &err, &skip);
+		MC = peek_uleb (bf->buf, &err, &skip);
 		if (err) {
 			eprintf ("Error\n");
 			break;
 		}
 		offset += skip;
+		skipped += skip;
 		// TODO: MOVE CHECKS OUTSIDE!
 		if (MI < dex->header.method_size) {
 			if (methods) {
@@ -1305,7 +1309,6 @@ static size_t parse_dex_class_method(RBinFile *bf, RBinDexClass *c, RBinClass *c
 						R_FREE (signature);
 						break;
 					}
-					ut64 bufsz = r_buf_size (bf->buf);
 					if (bufsz < offset || bufsz < offset + 8) {
 						R_FREE (signature);
 						break;
@@ -1425,7 +1428,6 @@ static size_t parse_dex_class_method(RBinFile *bf, RBinDexClass *c, RBinClass *c
 			sym->method_flags = get_method_flags (MA);
 			sym->ordinal = (*sym_count)++;
 			if (MC > 0) {
-				ut64 bufsz = r_buf_size (bf->buf);
 				if (bufsz < MC || bufsz < MC + 16) {
 					R_FREE (sym);
 					R_FREE (signature);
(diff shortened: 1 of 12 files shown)


radare:
Maybe add r_buf_const_data() api to make that caching implicit, otherwise i think it can affect some other usecases like the ones depending on IO that can change without using the rbuf.write api

ret2libc:
Please fill the PR template, otherwise it's hard for me to test the changes...

radare:
done

libr/util/buf_bytes.c
```diff
@@ -1,3 +1,5 @@
+/* radare - LGPL - Copyright 2009-2019 - pancake */
```
ret2libc, line 1: please put my name as well on all these `buf_*` files :)

libr/util/buf.c
```diff
@@ -65,6 +68,12 @@ static ut8 *get_whole_buf(RBuffer *b, ut64 *sz) {
 	if (bsz == UT64_MAX) {
```
ret2libc, line 68: could you remove also the comment before this line as well since you are at it? i don't think it makes any sense :/

libr/util/buf.c
```diff
@@ -190,6 +194,17 @@ R_API RBuffer *r_buf_new(void) {
 	return new_buffer (R_BUFFER_BYTES, &u);
 }
 
+R_API const ut8 *r_buf_data_cached(RBuffer *b, ut64 *size) {
```
ret2libc, line 197: Why a new API when there is already r_buf_data that does already the same thing? I think it doesn't make a lot of sense from an API point of view exposing the fact that the data internally is cached

ret2libc:
As said, I think r_buf_data is not needed here and it doesn't need to be changed either. We should just read enough data to parse the single class and not get the entire buffer every time

libr/bin/p/bin_pe.inc
```diff
@@ -420,7 +420,7 @@ static int haschr(const RBinFile* bf, ut16 dllCharacteristic) {
 	if (!bf) {
 		return false;
 	}
-	buf = r_buf_data (bf->buf, &sz);
+	buf = r_buf_data_cached (bf->buf, &sz);
```
ret2libc, line 423: this should not use `r_buf_data` either.

indeed this function just reads few bytes from the buffer, so you can use `r_buf_read_at` (or `r_buf_read_le16_at`, etc.) to read the bytes and `r_buf_size` to get the size. There is no need to read the whole buffer.

libr/include/r_util/r_buf.h
```diff
@@ -101,6 +102,7 @@ R_API st64 r_buf_fwrite_at(RBuffer *b, ut64 addr, const ut8 *buf, const char *fm
 // entire buffer in memory. Consider using the r_buf_read* APIs instead and read
 // only the chunks you need.
 R_API const ut8 *r_buf_data(RBuffer *b, ut64 *size);
+R_API const ut8 *r_buf_data_cached(RBuffer *b, ut64 *size);
```
ret2libc, line 105: remove this API, cache status should be just a "private" property and callers should not be aware of the fact that things are cached.

radare:
It does because the underlying buf can change and we cant invalidate the cache without this api

ret2libc:
> It does because the underlying buf can change and we cant invalidate the cache without this api

Why? What does this API adds? If the underlying buffer can change when using `r_buf_data`, it can change as well when using `r_buf_data_cached`, so you end up with possible inconsistencies in any case. As said, I think this API is not needed at the moment, especially because those multiple `r_buf_data` should really be `r_buf_read_at`.

radare:
I agree that Rbufdata should die. But the conversion is not easy in the dex code. And also making many calls to rbufreadat is also slow when the backend is not buf

ret2libc:
> I agree that Rbufdata should die. But the conversion is not easy in the dex code. And also making many calls to rbufreadat is also slow when the backend is not buf

I think it won't be *that* slow. We can try though, I just need to first include https://github.com/radareorg/radare2/pull/16401 . I think I can help with converting bin_dex from r_buf_data to r_buf_read ;)

radare:
if you can help on this im all for killing r_buf_data

ret2libc:
> do you agree on having this PR as a temporal fix to get 2x speedup in DEX files for now until we get rid of the r_buf_data() function?

I would avoid adding code that should be changed/removed anyway, unless it is really necessary. This dex thing has been like this for a long time, so I don't think it's so urgent to deserve a temporary fix. IMHO it can wait for an optimal fix.

radare:
You dont think anything is urgent because you dont use r2. I bet everyone loading dex files even more from android phones will appreciate a 2x speedup even if that implies removing few lines of code added for this stupid cache

radare:
seems like i broke the parser

https://api.travis-ci.com/v3/job/314184908/log.txt

libr/bin/p/bin_dex.c
```diff
@@ -1190,14 +1207,24 @@ static const ut8 *parse_dex_class_method(RBinFile *bf, RBinDexClass *c, RBinClas
...
+	ut64 bufsz = r_buf_size (bf->buf);
+	r_buf_seek (bf->buf, offset, R_BUF_SET);
 	for (i = 0; i < DM; i++) {
-		encoded_method_addr = p;
-		ut64 MI, MA, MC;
-		p = r_uleb128 (p, p_end - p, &MI);
+		ut64 encoded_method_addr = offset; //  + skip;
+		bool err = false;
+		skip = 0;
+		r_buf_seek (bf->buf, offset, R_BUF_SET);
```
ret2libc, line 1217: i do not think it is necessary to do this. It seems to me there are no reads in the middle that would change the seek.

libr/bin/p/bin_dex.c
```diff
@@ -1190,14 +1207,24 @@ static const ut8 *parse_dex_class_method(RBinFile *bf, RBinDexClass *c, RBinClas
...
 		DM = 4096;
 	}
+	size_t skip = 0;
+	ut64 bufsz = r_buf_size (bf->buf);
+	r_buf_seek (bf->buf, offset, R_BUF_SET);
 	for (i = 0; i < DM; i++) {
-		encoded_method_addr = p;
-		ut64 MI, MA, MC;
-		p = r_uleb128 (p, p_end - p, &MI);
+		ut64 encoded_method_addr = offset; //  + skip;
```
ret2libc, line 1214: i think this can become just `r_buf_tell`

libr/bin/p/bin_dex.c
```diff
@@ -1567,85 +1574,42 @@ static void parse_class(RBinFile *bf, RBinDexClass *c, int class_index, int *met
...
-		c->class_data->static_fields_size = eof;
+		c->class_data = dc;
 
-		p = r_uleb128 (p, p_end - p, &eof);
-		if (p >= p_end) {
-			free (c->class_data);
-			return;
-		}
-		c->class_data->instance_fields_size = eof;
+		ut64 offset = c->class_data_offset + skip;
```
ret2libc, line 1598: why is this offset needed at all? it's probably enough to seek the rbuf to the right offset and there is no need to pass offset to other functions.

radare:
im done with this pr, please @ret2libc aprove or review again

libr/util/buf.c
```diff
@@ -42,6 +42,9 @@ static st64 buf_read(RBuffer *b, ut8 *buf, size_t len) {
 
 static st64 buf_write(RBuffer *b, const ut8 *buf, size_t len) {
 	r_return_val_if_fail (b && b->methods, -1);
+	if (b->whole_buf) {
```
ret2libc, line 45: remove this, it is not needed

radare:
ready to merge now

PR #16401, opened 2020-04-02
r_buffer: do not move seek when using _at APIs

ret2libc:
This allows to use r_buf_read/r_buf_write to read/write sequentially but
at the same time having random reads/writes in the buffer if needed,
without the need to save/restore the seek every time.

 

**Detailed description**

Sometimes you may want to use the sequential API like `r_buf_read`/`r_buf_write` but you may also need to read/write in random places (e.g. in bin_dex.c). This makes sure that the read_at/write_at APIs restore the original seek after reading/writing data.

**Test plan**

Everything should still work as usual.

### DIFF
--- a/libr/util/buf.c
+++ b/libr/util/buf.c
@@ -537,11 +537,14 @@ R_API st64 r_buf_fread(RBuffer *b, ut8 *buf, const char *fmt, int n) {
 
 R_API st64 r_buf_fread_at(RBuffer *b, ut64 addr, ut8 *buf, const char *fmt, int n) {
 	r_return_val_if_fail (b && buf && fmt, -1);
+	st64 o_addr = r_buf_seek (b, 0, R_BUF_CUR);
 	int r = r_buf_seek (b, addr, R_BUF_SET);
 	if (r < 0) {
 		return r;
 	}
-	return r_buf_fread (b, buf, fmt, n);
+	r = r_buf_fread (b, buf, fmt, n);
+	r_buf_seek (b, o_addr, R_BUF_SET);
+	return r;
 }
 
 R_API st64 r_buf_fwrite(RBuffer *b, const ut8 *buf, const char *fmt, int n) {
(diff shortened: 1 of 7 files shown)


radare:
any unit test?

ret2libc:
there are already unit tests to check the API for this, but the errors are probably due to some users of those API that wrongly assumed seek was changed after a `read_at`/`write_at`. I'll check those out.

radare:
looks red

libr/util/buf.c
```diff
@@ -555,31 +558,40 @@ R_API st64 r_buf_fwrite(RBuffer *b, const ut8 *buf, const char *fmt, int n) {
...
 R_API st64 r_buf_write_at(RBuffer *b, ut64 addr, const ut8 *buf, ut64 len) {
 	r_return_val_if_fail (b && buf && !b->readonly, -1);
+	st64 o_addr = r_buf_seek (b, 0, R_BUF_CUR);
 	st64 r = r_buf_seek (b, addr, R_BUF_SET);
 	if (r < 0) {
 		return r;
 	}
 
-	return r_buf_write (b, buf, len);
+	r = r_buf_write (b, buf, len);
```
radare, line 592: isnt r_buf_write suposed to increase the seek ? why do we have to call rbufseek in so many places?

ret2libc: yes, `r_buf_write` increase the seek, but the following `r_buf_seek` restores the original seek to the place it was before doing the `r_buf_write_at`. Suppose you are doing:
```C
r_buf_seek(b, 0, R_BUF_SET);
r_buf_read(b, buf, 10);
r_buf_read(b, buf2, 1);
r_buf_read(b, buf3, 5);
```

then you need to read at some random place, so you would need to do `r_buf_read_at`, but without this patch `r_buf_read_at` would change the seek, so you can't go back to reading sequentially. With this patch, instead, you are sure that after doing a `r_buf_read_at`/`r_buf_write_at` the seek is not changed.

ret2libc:
@radare WDYT? Do you have concerns about this? Otherwise, it's green :)

radare:
yes i have concerns about repeated code and performance

ret2libc:
> yes i have concerns about repeated code and performance

I don't see repeated code.
About performance, I know it's not a good metric, but travis seem in line with the other builds so it doesn't seem like something important.

radare:
You cant use travis as a performance metric. Even less when we have no performance tests. But its clear to me that your patch is adding seek calls to every single bufread which adds a payload that will make the io backend even more slow this is what i was exposing ib my dex performance issue with more details

But obviously this additional time wont be noticeable in proportion and its preferibly to be consistent in behaviour

PR #16402, opened 2020-04-02
Fix a leak in r2r

thestr4ng3r:


thestr4ng3r:
appveyor canceled on purpose

PR #16409, opened 2020-04-02
Add Windows argv Formatting

thestr4ng3r:
- [ ] ~~I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)~~

**Detailed description**

This is needed because Windows processes do not receive an array of arguments like on unix but only a single string which is then split by the called program. So to call a subprocess which should get exactly the argv we want, this formatter is needed.

Reference:
http://daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULES
https://docs.microsoft.com/en-us/cpp/cpp/main-function-command-line-args?redirectedfrom=MSDN&view=vs-2019#parsing-c-command-line-arguments

**Test plan**

run the tests


radare:
This function is not used anywhere

thestr4ng3r:
It's for r2r, just splitting this off to have a better overview of the changes.

PR #16410, opened 2020-04-02
R2R for Windows

thestr4ng3r:
inet
err
sm
fail
palm
ie


radare:
So much red

ret2libc:
Can we have a look from someone who usually use windows? @pelijah @xarkes  ?

Issue #12732, opened 2019-01-10
do not rely on section info when DT_DYNAMIC is available


ret2libc:
r2r broken test: https://github.com/radare/radare2-regressions/commit/19ccc45154ca69041b8432147197738a453419e5

ELF parser should not rely on sections to compute relocations because they are not used by the loader. The loader uses the info provided in the DT_DYNAMIC segment.

08A:
Hi could you indicate where i should look in the code if i want override this behavior with a lookup on the DT_DYNAMIC segment.

ret2libc:
@08A Hi, I would start from `Elf_(r_bin_elf_get_relocs)` defined in `libr/bin/format/elf/elf.c`. As you can see it iterates over the sections looking for `.rela` or `.rel` sections. As said, loaders don't really care about sections, but they do use the information in the DT_DYNAMIC segment to find out relocations (I cannot tell you the exact details as I don't remember them anymore, but I guess the entries PLTRELSZ, PLTREL, JMPREL, RELA, RELASZ, RELAENT, RELACOUNT may help). Also, looking at the source code of the linux dynamic loader may help.

08A:
@ret2libc There is in libr/bin/format/elf/elf.c an interesting function init_dynamic_section. I was wondering if i could modify this function and the struct ELFOBJ. I could add DT_RELA, DT_REL, ..

08A:
I move here the list of possible refacto:

 - rel_cache_new and Elf_(r_bin_elf_get_relocs) (same behaviour different data structure, maybe replace by the new implementation or reused inside rel_cache_new)
- move struct dynamic_relocation_section inside ELFOBJ
- move the function get_dynamic_info inside init_dynamic_section
- remove is_rela inside ELFOBJ

08A:
I am trying to remove the `struct ht_rel_t` but in the function `get_import_addr_ppc` k is used:
`rel->k`.
What is the purpose of the field?

08A:
`get_import_addr_ppc` get the vaddr?

ret2libc:
> I am trying to remove the `struct ht_rel_t` but in the function `get_import_addr_ppc` k is used:
> `rel->k`.
> What is the purpose of the field?

AFAIS that is the position of the reloc entry in the relocations. IIRC ppc has some kind of relocation types that compute the final address based on the position of the relocation entry (or something like that, I don't remember exactly sorry).

08A:
[...]

PR #16419, opened 2020-04-05
When computing relocations, use DYNAMIC segment if available

08A:

**Detailed description**
I start working on the dynamic section, the ir command found all the relocations. I don't know why the pd 15 command doesn't use all the relocs (new/db/formats/elf/reloc). I used readelf to test the relocation. Maybe later we could move the struct dynamic_relocation_section in the struct ELFOBJ?

**Test plan**
```
ir
```

**Closing issues**

#12732

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -2676,7 +2676,7 @@ static int read_reloc(ELFOBJ *bin, RBinElfReloc *r, int is_rela, ut64 offset) {
 
 	if (is_rela == DT_RELA) {
 		reloc_info.r_addend = READWORD (buf, i);
-	    r->addend = reloc_info.r_addend;
+		r->addend = reloc_info.r_addend;
 	}
 
 	r->is_rela = is_rela;
@@ -2707,30 +2707,38 @@ static size_t get_num_relocs(struct dynamic_relocation_section *info) {
 	return res;
 }
 
+static size_t populate_relocs_record_from_dynamic(ELFOBJ *bin,
+	struct dynamic_relocation_section *info, RBinElfReloc *relocs, size_t pos) {
+	size_t size = info->plt_mode == DT_RELA? sizeof (Elf_(Rela)): sizeof (Elf_(Rel));
+
+	for (size_t offset = 0; offset < info->rela_size; offset += info->relaent) {
+		read_reloc (bin, relocs + pos, DT_RELA, info->addr_rela + offset);
+		fix_rva_and_offset (bin, relocs + pos, info->addr_rela);
+		pos++;
+	}
+
+	for (size_t offset = 0; offset < info->rel_size; offset += info->relent) {
+		read_reloc (bin, relocs + pos, DT_REL, info->addr_rel + offset);
+		fix_rva_and_offset (bin, relocs + pos, info->addr_rel);
+		pos++;
+	}
+
+	for (size_t offset = 0; offset < info->jmprel_size; offset += size) {
+		read_reloc (bin, relocs + pos, info->plt_mode, info->addr_jmprel + offset);
+		fix_rva_and_offset (bin, relocs + pos, info->addr_jmprel);
+		pos++;
+	}
+
+	return pos;
+}
+
 static RBinElfReloc *populate_relocs_record(ELFOBJ *bin, struct dynamic_relocation_section *info) {
 	size_t num_relocs = get_num_relocs (info);
 	bin->reloc_num = num_relocs;
 	RBinElfReloc *relocs = calloc (num_relocs + 1, sizeof (RBinElfReloc));
+
 	size_t i = 0;
-
-	for (size_t offset = 0; offset < info->rela_size; offset += info->relaent) {
-		read_reloc (bin, relocs + i, DT_RELA, info->addr_rela + offset);
-		fix_rva_and_offset (bin, relocs + i, info->addr_rela);
-		i++;
-	}
-
-	for (size_t offset = 0; offset < info->rel_size; offset += info->relent) {
-		read_reloc (bin, relocs + i, DT_REL, info->addr_rel + offset);
-		fix_rva_and_offset (bin, relocs + i, info->addr_rel);
-		i++;
-	}
-
-	size_t size = info->plt_mode == DT_RELA? sizeof (Elf_(Rela)): sizeof (Elf_(Rel));
-	for (size_t offset = 0; offset < info->jmprel_size; offset += size) {
-		read_reloc (bin, relocs + i, info->plt_mode, info->addr_jmprel + offset);
-		fix_rva_and_offset (bin, relocs + i, info->addr_jmprel);
-		i++;
-	}
+	i = populate_relocs_record_from_dynamic (bin, info, relocs, i);
 
 	relocs[num_relocs].last = 1;
 	return relocs;
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -2642,6 +2642,97 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
 	return ret;
 }
 
+static void fix_rva_and_offset(ELFOBJ *bin, RBinElfReloc *r, ut64 offset_table) {
+	if (is_bin_etrel (bin)) {
+		r->rva = offset_table + r->offset;
+		r->rva = Elf_(r_bin_elf_p2v) (bin, r->rva);
+	} else {
+		r->rva = r->offset;
+		r->offset = Elf_(r_bin_elf_v2p) (bin, r->offset);
+	}
+}
+
+static int read_reloc(ELFOBJ *bin, RBinElfReloc *r, int is_rela, ut64 offset) {
+	size_t size_struct;
+
+	if (is_rela) {
+		size_struct = sizeof (Elf_(Rela));
+	} else {
+		size_struct = sizeof (Elf_(Rel));
+	}
+
+	// TODO 2nd check overflow?
+	if ((offset + size_struct) > bin->size || (offset + size_struct) < offset) {
+		return -1;
+	}
+
+	ut8 buf[sizeof (Elf_(Rela))] = { 0 };
+	int res = r_buf_read_at (bin->b, offset, buf, size_struct);
+	if (res != size_struct) {
+		return -1;
+	}
+
+	// TODO make a single read and work with the buffer
+	size_t i = 0;
+	Elf_(Rela) reloc_info;
+
+#if R_BIN_ELF64
+	reloc_info.r_offset = READ64 (buf, i);
+	reloc_info.r_info = READ64 (buf, i);
+#else
+	reloc_info.r_offset = READ32 (buf, i);
+	reloc_info.r_info = READ32 (buf, i);
+#endif
+
+	r->offset = reloc_info.r_offset;
+	r->type = ELF_R_TYPE (reloc_info.r_info);
+	r->sym = ELF_R_SYM (reloc_info.r_info);
+	r->is_rela = is_rela;
+	r->last = 0;
+
+	if (is_rela == DT_RELA) {
+#if R_BIN_ELF64
+		reloc_info.r_addend = READ64 (buf, i);
+#else
+		reloc_info.r_addend = READ32 (buf, i);
+#endif
+		r->addend = reloc_info.r_addend;
+	}
+
+	return 1;
+}
+
+static void populate_relocs_record(ELFOBJ *bin, RBinElfReloc *relocs,
+	struct dynamic_relocation_section *info) {
+	size_t i = 0;
+
+	for (size_t offset = 0; offset < info->rela_size; offset += info->relaent) {
+		read_reloc (bin, relocs + i, DT_RELA, info->addr_rela + offset);
+		fix_rva_and_offset (bin, relocs + i, info->addr_rela);
+		i++;
+	}
+
+	for (size_t offset = 0; offset < info->rel_size; offset += info->relent) {
+		read_reloc (bin, relocs + i, DT_REL, info->addr_rel + offset);
+		fix_rva_and_offset (bin, relocs + i, info->addr_rel);
+		i++;
+	}
+}
+
+static size_t get_num_relocs(struct dynamic_relocation_section *info) {
+	size_t res = 0;
+
+	if (info->relaent) {
+		res += info->rela_size / info->relaent;
+	}
+
+	if (info->relent) {
+		res += info->rel_size / info->relent;
+	}
+
+	return res;
+}
+
 static struct dynamic_relocation_section *get_dynamic_info(ELFOBJ *bin) {
 	struct dynamic_relocation_section *res = calloc (1, sizeof (struct dynamic_relocation_section));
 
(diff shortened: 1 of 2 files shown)


libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2642,163 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
+	RBinElfReloc *ret = NULL;
+
+	if (!bin || !bin->g_sections) {
+		return NULL;
+	}
+
+	struct dynamic_relocation_section *info = get_dynamic_info (bin);
+	size_t num_relocs = get_num_relocs (info);
+	ret = calloc (num_relocs + 1, sizeof (RBinElfReloc));
+	populate_relocs_record (bin, ret, info);
```
ret2libc, line 2798: better do something like:
```C
ret = populate_relocs_record (bin, info);
```

and let `populate_relocs_record` allocate the memory.

08A:
I am testing to see why it failed, and i was wandering if one test could be wrong?

file: ./test/new/db/cmd/cmd_i
```
3182 NAME=ir (file x86_64, ifunc)
3183 FILE=../bins/elf/ifunc_rel64
3184 EXPECT=<<EOF
3185 [Relocations]
3186
3187 vaddr      paddr      type   name
3188 ---------------------------------
3189 0x006b17d0 0x000b17d0 SET_64 ifunc_416530 0x00416530 (ifunc)
3190 0x006b17d8 0x000b17d8 SET_64 ifunc_4193a0 0x004193a0 (ifunc)
3191 0x006b17e0 0x000b17e0 SET_64 ifunc_4199b0 0x004199b0 (ifunc)
3192 0x006b17e8 0x000b17e8 SET_64 ifunc_418a50 0x00418a50 (ifunc)
3193 0x006b17f0 0x000b17f0 SET_64 ifunc_418de0 0x00418de0 (ifunc)
3194 0x006b17f8 0x000b17f8 SET_64 ifunc_419ad0 0x00419ad0 (ifunc)
3195 0x006b1800 0x000b1800 SET_64 ifunc_4193f0 0x004193f0 (ifunc)
3196 0x006b1808 0x000b1808 SET_64 ifunc_414300 0x00414300 (ifunc)
3197 0x006b1810 0x000b1810 SET_64 ifunc_479800 0x00479800 (ifunc)
3198 0x006b1818 0x000b1818 SET_64 ifunc_4140b0 0x004140b0 (ifunc)
```
../bins/elf/ifunc_rel64 hasn't any .dynamic section.
```
❯ readelf -d ./test/bins/elf/ifunc_rel64

There is no dynamic section in this file.
```
Readelf believe that there are some relocations, but the ld don't do any relocation on this file.

08A:
And there is the problem of the .obj file...
There is no relocation, but in the future if the file is linked there will be some relocation.

08A:
In the case of ./test/bins/elf/radare2.c.obj which relocations r2 need?
'readelf -r' list a great number of relocations.

ret2libc:
I think what happens in this case is that the function is statically linked (as `file` can confirm) so there is no dynamic segment as the executable doesn't need to pass any info to the dynamic loader. However the compiler still included those additional information.

The thing is that we still want the information provided by the sections because, as you said, in some cases they are the only information we have and in other cases they may enhance what we already have. However, I think care must be taken to ensure that first we load info from the dynamic segment and only later we *enhance* those info with the section information (if there are conflicts between info provided by DYNAMIC segment and .rel sections, DYNAMIC should win).

Does that make sense? What do you think?

08A:
Okay, that doesn't seems to be to complicated but how could we detect collision?
We could ignore sections with the same offset than informations inside the .dynamic.

The second problem is detecting valid rel or rela section. If a section use 'rel' or 'rela' in his name it doesn't guarantee that the section hold relocation entries. Maybe we could list all the section's names with a special meaning?

ret2libc:
> Okay, that doesn't seems to be too complicated but how could we detect collision?
> We could ignore sections with the same offset than informations inside the .dynamic.

I think you can first get relocations from DYNAMIC (like what you are doing right now), then extract them from sections (just get the old code before your change, it should work well enough). While getting relocations from sections, check if there is already one at the same `vaddr`. If there is, ignore the one from sections and trust the one in DYNAMIC. If there isn't, add the new relocation to the list of relocations.

> The second problem is detecting valid rel or rela section. If a section use 'rel' or 'rela' in his name it doesn't guarantee that the section hold relocation entries. Maybe we could list all the section's names with a special meaning?

There is nothing you can do about this. The only way to differentiate sections is with their names AFAIK.

08A:
The problem with the vaddr is that some debug reloc could have the same addr than .dynamic reloc. 
We could check that the entrie's offset is not inside the range of a previous relocation entries.

08A:
There is two similar function 'Elf_(r_bin_elf_get_relocs)' and 'rel_cache_new'. Should i merge them?

08A:
We could use the new array of reloc to generate the hash map.

ret2libc:
@08A I think this is not working as expected. I tried the tests in https://github.com/radareorg/radare2-regressions/commit/19ccc45154ca69041b8432147197738a453419e5 and it is still failing. Are you using the info from dynamic?

radare:
can you add a test?

08A:
The https://github.com/radareorg/radare2-regressions/commit/19ccc45154ca69041b8432147197738a453419e5 can't work on __libc_start_main because there is no relocation related to this symbol.

08A:
```python
import sys
import lief

def rename_section_rela(filename, output):
    binary = lief.parse(filename)
    for i, section in enumerate(binary.sections):
        if ("rel" in section.name):
            section.name = f"section_number_{i}"
    binary.write(output);

if __name__ == "__main__":
    if len(sys.argv) != 3:
        print("Usage: {} <elf binary> <output>".format(sys.argv[0]))
        sys.exit(1)

    rename_section_rela(sys.argv[1], sys.argv[2])
```

I used this small script on this simple binary
```c
#include <stdio.h>

int main(void) {
    printf("Hello World!!!");
    return 0;
}
```

I will send a pr on https://github.com/radareorg/radare2-testbins

ret2libc:
> The [radareorg/radare2-regressions@19ccc45](https://github.com/radareorg/radare2-regressions/commit/19ccc45154ca69041b8432147197738a453419e5) can't work on __libc_start_main because there is no relocation related to this symbol.

Somehow I uploaded a wrong binary in that commit :/ As you say, that bin does not work, sorry about that. Could you please update it so that it works or just remove it if you already have added another test for this behaviour? I will review your code today/tomorrow, but the results look good! Thanks :)

08A:
I need some advice for this pr. I believe some refacto can be done.

- rel_cache_new and Elf_(r_bin_elf_get_relocs) (same behaviour different data structure, maybe replace by the new implementation or reused inside rel_cache_new)
- move struct dynamic_relocation_section inside ELFOBJ
- move the function get_dynamic_info inside init_dynamic_section
- remove the is_rela inside ELFOBJ

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,276 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
+		size_struct = sizeof (Elf_(Rel));
+	}
+
+	ut8 buf[sizeof (Elf_(Rela))] = { 0 };
+	int res = r_buf_read_at (bin->b, offset, buf, size_struct);
+	if (res != size_struct) {
+		return -1;
+	}
+
+	// TODO make a single read and work with the buffer
```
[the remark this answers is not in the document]
ret2libc: maybe it was an old comment. If so, just delete it.

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,276 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
 	return ret;
 }
 
+static void fix_rva_and_offset(ELFOBJ *bin, RBinElfReloc *r, ut64 offset_table) {
+	if (is_bin_etrel (bin)) {
+		r->rva = offset_table + r->offset;
```
ret2libc, line 2640: where did you get this from? I think it will never be executed as you are calling `fix_rva_and_offset` only from `populate_relocs_record_from_dynamic` and ET_REL objects do not have dynamic segment AFAIK.

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,276 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
+					relocs[pos].rva = bin->shdr[bin->g_sections[i].info].sh_offset + relocs[pos].offset;
+					relocs[pos].rva = Elf_(r_bin_elf_p2v) (bin, relocs[pos].rva);
 				} else {
-					ret[rel].rva = ret[rel].offset;
+					relocs[pos].rva = relocs[pos].offset;
 				}
 			} else {
-				ret[rel].rva = ret[rel].offset;
-				ret[rel].offset = Elf_(r_bin_elf_v2p) (bin, ret[rel].offset);
+				relocs[pos].rva = relocs[pos].offset;
```
ret2libc, line 2834: this code is the same as the one in `fix_rva_and_offset`. Maybe by changing a bit that function we can just call it here as well, instead of the whole `if(is_bin_etrel (bin)) ....`

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,276 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
-RBinElfReloc* Elf_(r_bin_elf_get_relocs)(ELFOBJ *bin) {
-	int res, rel, rela, i, j;
-	RBinElfReloc *ret = NULL;
+	return pos;
+}
 
-	if (!bin || !bin->g_sections) {
-		return NULL;
+static size_t get_section_mode(ELFOBJ *bin, size_t pos) {
+	if (!strncmp (bin->g_sections[pos].name, ".rela.", strlen (".rela."))) {
```
ret2libc, line 2763: use `r_str_startswith` instead of `strncmp`

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,276 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
+	} else {
+		r->rva = r->offset;
+		r->offset = Elf_(r_bin_elf_v2p) (bin, r->offset);
+	}
+}
+
+static int read_reloc(ELFOBJ *bin, RBinElfReloc *r, int is_rela, ut64 offset) {
+	size_t size_struct;
+
+	if (is_rela) {
```
ret2libc, line 2651: actually, i'm not sure this works as expected, because `read_reloc` receives `DT_REL` and `DT_RELA` and not a boolean. I think you can probably do `size_t size_struct = get_size_rel_mode (is_rela)`.

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,276 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
+	if (is_bin_etrel (bin)) {
+		r->rva = offset_table + r->offset;
+		r->rva = Elf_(r_bin_elf_p2v) (bin, r->rva);
+	} else {
+		r->rva = r->offset;
+		r->offset = Elf_(r_bin_elf_v2p) (bin, r->offset);
+	}
+}
+
+static int read_reloc(ELFOBJ *bin, RBinElfReloc *r, int is_rela, ut64 offset) {
```
ret2libc, line 2648: maybe change the name from `is_rela` to `rel_mode`, because `is_XXXX` seems like a boolean value.

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,276 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
 	if (!bin->g_sections) {
 		return 0;
 	}
-	size = bin->is_rela == DT_REL ? sizeof (Elf_(Rel)) : sizeof (Elf_(Rela));
+
 	for (i = 0; !bin->g_sections[i].last; i++) {
 		if (sectionIsInvalid (bin, &bin->g_sections[i])) {
 			continue;
 		}
 		if (!strncmp (bin->g_sections[i].name, ".rela.", strlen (".rela."))) {
```
ret2libc, line 2721: use `r_str_startswith`.

libr/bin/format/elf/elf.c
ret2libc: the if part can be useful if you use `fix_rva_and_offset` in `populate_relocs_record_from_section` as well, but it needs to be changed a bit I think.

libr/bin/format/elf/elf.c
ret2libc, line 2648: also, this seems to return a `bool` now.

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,276 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
+
+static bool is_reloc_section(size_t rel_mode) {
+	return rel_mode == DT_REL || rel_mode == DT_RELA;
+}
+
+static bool has_valid_section_header(ELFOBJ *bin, size_t pos) {
+	return bin->g_sections[pos].info < bin->ehdr.e_shnum && bin->shdr;
+}
+
+static size_t get_next_not_analysed_offset(size_t section_offset, size_t offset,
```
ret2libc, line 2780: i have an hard time following this function :(

can it be simplified a bit? also, i'm not sure it's right that we check for all the `if` cases when we are really working only with one type of rel (e.g. if i'm iterating over `rela` entries, this function may still check `info->addr_jmprel - base_addr <= g_offset && g_offset < info->addr_jmprel + info->jmprel_size - base_addr`.

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,276 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
-			if (!bin->is_rela) {
-				size = sizeof (Elf_(Rela));
-			}
+			size = sizeof (Elf_(Rela));
 			ret += NUMENTRIES_ROUNDUP (bin->g_sections[i].size, size);
-		} else if (!strncmp (bin->g_sections[i].name, ".rel.", strlen (".rel."))){
-			if (!bin->is_rela) {
-				size = sizeof (Elf_(Rel));
-			}
+		} else if (!strncmp (bin->g_sections[i].name, ".rel.", strlen (".rel."))) {
```
ret2libc, line 2724: `r_str_startswith`

libr/bin/format/elf/elf.c
ret2libc, line 2721: maybe even better you can reuse `get_section_mode` and `get_size_rel_mode` here.

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,276 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
 		}
 	}
-	ret[reloc_num].last = 1;
+
+	return pos;
+}
+
+static RBinElfReloc *populate_relocs_record(ELFOBJ *bin, struct dynamic_relocation_section *info) {
+	size_t i = 0;
+	size_t num_relocs = get_num_relocs_approx (bin, info);
```
ret2libc, line 2846: this number is always `>= real_num_relocs`, right?

libr/bin/format/elf/elf.c
08A: The main problem is that we could have obfuscated binary like that
```
section: rel.plt
false_relocation | real rela table | jmprel table | false rel table
```
We can't trust the type of relocation in the name.
Maybe we could had an option to load or not  the relocations record from section?
It could be an start's option.

libr/bin/format/elf/elf.c
08A: yes

libr/bin/format/elf/elf.c
08A: Indeed i will split this function in two sub function.

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,265 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
+	return rel_mode == DT_RELA? sizeof (Elf_(Rela)): sizeof (Elf_(Rel));
+}
+
+static bool read_reloc(ELFOBJ *bin, RBinElfReloc *r, size_t rel_mode, ut64 offset) {
+	size_t size_struct = get_size_rel_mode (rel_mode);
+
+	ut8 buf[sizeof (Elf_(Rela))] = { 0 };
+	int res = r_buf_read_at (bin->b, offset, buf, size_struct);
+	if (res != size_struct) {
+		return 0;
```
ret2libc, line 2674: `return false`

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,265 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
+		r->addend = reloc_info.r_addend;
+	}
+
+	r->is_rela = rel_mode;
+	r->last = 0;
+	r->offset = reloc_info.r_offset;
+	r->sym = ELF_R_SYM (reloc_info.r_info);
+	r->type = ELF_R_TYPE (reloc_info.r_info);
+
+	return 1;
```
ret2libc, line 2694: `return true`

libr/bin/format/elf/elf.c
```diff
@@ -2627,150 +2635,265 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
...
-				size = sizeof (Elf_(Rela));
-			}
-			ret += NUMENTRIES_ROUNDUP (bin->g_sections[i].size, size);
-		} else if (!strncmp (bin->g_sections[i].name, ".rel.", strlen (".rel."))){
-			if (!bin->is_rela) {
-				size = sizeof (Elf_(Rel));
-			}
-			ret += NUMENTRIES_ROUNDUP (bin->g_sections[i].size, size);
-		}
+		rel_mode = get_section_mode (bin, i);
```
ret2libc, line 2740: we shold check `is_reloc_section(rel_mode)` and when it is false we should just continue to the next section, otherwise the result will be wrong.

08A:
I am satisfied with the state of the PR, later today or tomorrow i will send an other PR to add some refactorization.

PR #16417, opened 2020-04-05
Move CMDS= before EXPECT= in tests

kazarmy:
**Your checklist for this pull request**
- [ ] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [ ] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

This pr moves `CMDS=` before `EXPECT=` in tests so that cause-and-effect is more apparent.

~~Conversion is done using the following:~~ (Fixed regex in update below)

```bash
(REGEX='((?:(?s)EXPECT=<<EOF\s*?\n.*?EOF\n)|(?:EXPECT=.*\n))((?:(?!RUN).)*\n)*?((?:(?s)CMDS=<<EOF\s*?\n.*?EOF\n)|(?:CMDS=.*\n))' && grep -rPzl "$REGEX" test/new | xargs -n 1 perl -0777 -i -pe "s/$REGEX/\$3\$1\$2/g")
```

There's a trailing `?` in `((?:(?!RUN).)*\n)*?` so that (theoretically) in the event of a missing `RUN`, the intended `CMDS=` is moved rather than the actual `CMDS=`.

**Test plan**

Travis and AppVeyor are green. Diff too large for visual inspection of everything.

**Closing issues**

...

----

**Update**

Had to tighten up the regexes used for the `EXPECT=<<EOF` and `CMDS=<<EOF` bodies. Tightened up the central regex as well:

```bash
(REGEX='((?:EXPECT=<<EOF\s*?\n(?:(?!EOF).*\n)*?EOF\n)|(?:EXPECT=.*\n))((?!RUN).*\n)*?((?:CMDS=<<EOF\s*?\n(?:(?!EOF).*\n)*?EOF\n)|(?:CMDS=.*\n))' && grep -rPzl "$REGEX" test/new | xargs -n 1 perl -0777 -i -pe "s/$REGEX/\$3\$1\$2/g")
```

PR #16439, opened 2020-04-06
WIP: Run full tests on fewer jobs during PR

ret2libc:
**Detailed description**

In PRs:
- only OS X and Linux-CLANG-build will run full tests, the other will just check the compilation
- on master branch all tests as of right now are run.

**Test plan**

Unfortunately some things can be tested only once this will be merged in master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,46 +20,60 @@ os: osx
 jobs:
   fast_finish: true
   include:
-    # Linux with GCC
-    - os: linux
+    # Linux with GCC - full tests
+    - if: type != pull_request AND branch = master
+      os: linux
       name: Linux-GCC
       env: COMPILER_NAME=gcc CXX=g++ CC=gcc CFLAGS="-DR2_ASSERT_STDOUT=1"
-    # Linux with clang
+    # Linux with meson and coverage and asserts - full tests
+    - if: type != pull_request AND branch = master
+      os: linux
+      name: Linux-meson
+      env: COMPILER_NAME=gcc CXX=g++ CC=gcc INSTALL_SYSTEM=meson COVERAGE=1 CFLAGS="-DR2_ASSERT_STDOUT=1"
+    # Linux with GCC - compile  only
+    - if: type == pull_request
+      os: linux
+      name: Linux-GCC-compile
+      env: COMPILER_NAME=gcc CXX=g++ CC=gcc CFLAGS="-DR2_ASSERT_STDOUT=1" R2_TESTS_DISABLE=1
+    # Linux with clang - full tests
     - os: linux
       name: Linux-CLANG
       env: COMPILER_NAME=clang CXX=clang++ CC=clang CFLAGS="-DR2_ASSERT_STDOUT=1"
-    # Linux with meson and coverage and asserts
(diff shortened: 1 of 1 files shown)


ret2libc:
I also removed ASAN from each master commit... I don't think we are looking at those "unchecked" jobs often enough to make it worth having that (long) extra time to each commit. 

ASAN will still be run when you create a branch with the name `release-XXXX` or `prerelease-XXXXXX`. Those branches names are useful to have better testing before releases.

ret2libc:
Not ready yet, wait.

thestr4ng3r:
Tests on Linux-meson are important since that job does the coverage. But it will probably be faster on github actions anyway.

ret2libc:
> Tests on Linux-meson are important since that job does the coverage. But it will probably be faster on github actions anyway.

I know, I just disabled them on the PR, but it will still run on master branch on every commit. I just think we don't really need to check the difference in coverage for each PR, WDYT?

kazarmy:
> I just think we don't really need to check the difference in coverage for each PR, WDYT?

Emm the coverage check is only done on PRs afaik. No point in doing them on master right?

ret2libc:
> > I just think we don't really need to check the difference in coverage for each PR, WDYT?
> 
> Emm the coverage check is only done on PRs afaik. No point in doing them on master right?

What? Why? I'm pretty sure it runs on every master commit as I had configured it time ago (see https://codecov.io/gh/radareorg/radare2/branch/master ). It should still work on every commit AFAIK. Why would you want the coverage only on PRs and not on master? I think the purpose of coverage is for us to see which part of the code are not well tested and, in theory, improve the testing for those parts.

kazarmy:
Oh ok then. Are we still going to get those Codecov reports (e.g. https://github.com/radareorg/radare2/pull/16417#issuecomment-609379506)?

ret2libc:
If we don't run on PR, no, but... did we ever use them? :/ If there is any requests for them, we can keep the meson build as the one that runs all tests, but only if we really care. Otherwise let's run them only on master.

kazarmy:
> If we don't run on PR, no, but... did we ever use them?

We should care about them but I don't think anybody did. Don't think anybody cares about https://codecov.io/gh/radareorg/radare2/branch/master as well, at least for now.

kazarmy:
Though if we then did care about it, it would be rather tricky to fix coverage since we wouldn't know whether our fixes worked until they are already in master, and if they didn't actually work...?

But like you said, enable on request.

ret2libc:
> Though if we then did care about it, it would be rather tricky to fix coverage since we wouldn't know whether our fixes worked until they are already in master, and if they didn't actually work...?
> 
> But like you said, enable on request.

Yeah, when someone will want to go into this direction, we can easily have travis builds (or GH actions) that are run when specific branch names are used in PR, for example. For the moment I think we don't really need them on every PR as nobody is caring about those.

kazarmy:
Ok

ret2libc:
Actually since I'm already working on https://github.com/radareorg/radare2/pull/16442 I think this can be almost be closed/changed.

PR #16440, opened 2020-04-06
Try to fix spurious errors in AppVeyor

ret2libc:

**Detailed description**

This tries to fix some errors we see often enough in AppVeyor (e.g. https://ci.appveyor.com/project/radareorg/radare2/builds/31977704). See https://help.appveyor.com/discussions/problems/1272-fatal-reference-is-not-a-tree-sha-on-git-checkout .

**Test plan**

Well.. we have to try it for a bit.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,9 +4,6 @@ version: git-{build}
 # Skip Github tags
 skip_tags: true
 
-# Clone depth
-clone_depth: 1
-
 # Branches to clone
 branches:
   only:


ret2libc:
@XVilka https://ci.appveyor.com/project/radareorg/radare2/builds/31981212 can you restart those? Why were they canceled?

XVilka:
@ret2libc canceled those because it will not affect the CI at all. The change is fine, lets merge it.

kazarmy:
Will this result in more AppVeyor builds on master? Do we need to enable [rolling builds](https://www.appveyor.com/docs/build-configuration/#rolling-builds)?

kazarmy:
more AppVeyor builds -> more lengthy AppVeyor builds

ret2libc:
> Will this result in more AppVeyor builds on master? 

How would this change affect the number of builds?

> Do we need to enable [rolling builds](https://www.appveyor.com/docs/build-configuration/#rolling-builds)?

I don't know... that could be an idea, though one cool feature about having tests on each commit is that you can see when something breaks.

kazarmy:
> How would this change affect the number of builds?

Sorry I meant more lengthy AppVeyor builds, since [the referenced error](https://help.appveyor.com/discussions/problems/1272-fatal-reference-is-not-a-tree-sha-on-git-checkout) aborts the build almost immediately after it starts.

> ... though one cool feature about having tests on each commit is that you can see when something breaks.

True, though waiting a few hours for AppVeyor to return results is not for the impatient.

Issue #16329, opened 2020-03-27
Noisy continous releases


radare:
following the feeds, it seems like there's a new release in every single commit

https://github.com/radareorg/radare2/releases.atom

radare:
![Screenshot 2020-03-31 at 17 23 25](https://user-images.githubusercontent.com/917142/78046524-8cdfc880-7377-11ea-82b3-39a9989086be.png)

kazarmy:
I think it's best that we delete these draft releases manually. Sure, a script could be possibly be written that deletes them using the GitHub API within a Travis job, but a mistake/bug could spectacularly delete normal releases as well with no way of getting them back unless they are saved somewhere. Even then, it would be difficult to restore them and the timestamps would be off.

ret2libc:
Isn't this fixed @trufae ? I see only one of those continuous build right now. Please re-open if you think otherwise.

karliss:
I'd say its still somewhat noisy. It shouldn't be necessary to create a tag and fake release.

ret2libc:
@karliss What would be the right way to distribute those in your opinion?

Personally, I think we should not generate those things at all. I'm not sure someone is actually using the .deb packages... those .deb packages probably work only on the exact debian version that generated them (from .travis.yml it seems it is debian:stretch, which is quite old and I doubt there are that many oldstable debian user to justify having a continuous build just for them) because of the libc version requirement. @trufae do you remember whether we included those as part of some actual user request or just because we thought it was valuable? I'd vote to drop them.

XVilka:
I suggest not to drop them but package fully static musl-based binaries in
these Deb packages. So they will be compatible with almost everything.

ret2libc:
> I suggest not to drop them but package fully static musl-based binaries in these Deb packages. So they will be compatible with almost everything.

I don't like doing things without reasons. If there is a a real request from more than 1 person it is fine for me to do that. Otherwise, I'm for dropping it. Can't those people use git? Would those people even download those binaries?  As I said, given they are currently built only for debian:stretch, I *highly doubt* anyone is using those. So I don't want to create myself more things to do/bother with if not necessary.

PR #16442, opened 2020-04-06
WIP: Create more GitHub action jobs for meson and gcc/clang

ret2libc:
**Detailed description**

Let's try to move some jobs from Travis to GHActions. In theory, this should add GH actions to test meson, acr, gcc, clang, osx, linux in various combinations. Maybe it's worth doing something like what I've done in https://github.com/radareorg/radare2/pull/16439, where the full tests are run only one for each OS.


XVilka:
I suggest to move also Debian packaging here:
- https://github.com/radareorg/radare2-release/issues/35
- https://github.com/radareorg/radare2-release/issues/34
- https://github.com/radareorg/radare2/issues/16329

ret2libc:
Please assign those issues to me, I will do in separate PRs.

ret2libc:
@radare @XVilka ok I think this is a good set of GH actions for PR (see the actions in this PR). If this works well, I think we can remove Travis from PR completely.

For commits on master branch, instead, I will add later another workflow file that runs all the tests on all builds and also have Coverage. Travis will still be needed in this case to test ARM64, ppc64 and s390x archs.

ret2libc:
Much faster!

radare:
Maybe we can keep travis for the non-x86 platforms? or we can do everything in gh-actions?

XVilka:
Yes, we can keep Travis only for non-x86 platforms

PR #16446, opened 2020-04-06
Fix COFF symbols/imports info

pelijah:

**Detailed description**
1. Now plugin fills symbol visibility info (`bind` field) based on its class value.
2. Fixed imports detection logic (symbol is marked as imported if its section number value is 0).
3. Fixed function detection logic. Now plugin uses symbols's derived type value for this.

### DIFF
--- a/libr/bin/format/coff/coff_specs.h
+++ b/libr/bin/format/coff/coff_specs.h
@@ -133,7 +133,7 @@
 
 R_PACKED(
 struct coff_hdr {
-	ut16 f_magic;	/* Magic number */	
+	ut16 f_magic;	/* Magic number */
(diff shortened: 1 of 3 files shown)


pelijah:
Ok lets assume that name `main` is reserved so we can make an exception for it and mark related symbol as function.  But these `sym.imp....` imports names in `tif_dir.obj` are really weird.

radare:
it shouldnt be detecvted as a swift symbol, so its an issue in the code checking for which mangling is using

pelijah:
Ok I have updated tests.

ITAYC0HEN:
you requested review and merged two hours later :(

Issue #14024, opened 2019-05-10
Slow O(n) in R_IO


radare:
<img width="579" alt="Screenshot 2019-05-10 at 11 57 02" src="https://user-images.githubusercontent.com/917142/57519065-c603ea00-731a-11e9-88db-f78ab4ba3d1f.png">

radare:
reproducible with the '../bins/pe/65535sects.exe' bin.. but probably affects dyldcaches too

cc @mrmacete

radare:
alsso this is a pretty shitty test because its parsing the bin TWICE and not catching the segfault of rabin2 if any

```
[XX] db/formats/pe/96emptysections PE: corkami 96workingsections.exe - sections list  38
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -n -i /var/folders/12/517hk27n0pxbnbv9yxlh4jxm0000gn/T/tmp-84586AKJGluc2TTNF ../bins/pe/96emptysections.exe
!rabin2 -pS ../bins/pe/96emptysections.exe
```

radare:
in fact there are many more O(n) issues in io/map.c

radare:
Fixing this bottleneck willmake r2 load this bin in few < 2s instead of 50s

radare:
this means that we can reduce the testsuite time in travis for about 3s or so

ret2libc:
True that this list can be replaced with something like a RBTree, but having a binary with soo many sections is not that common and it won't probably improve the average case. So I'm not sure it's *that* important.

radare:
It is very common on ios

mrmacete:
yes, the kernelcache with all KETXs has easily 4000+ sections. same or worse for dyldcache (which is a ~1GB file with all system libs packed in and linked together)

thestr4ng3r:
RBTree might be overkill indeed. Generally the skyline is the optimization for quick access. From what I understand, the way `r_io_map_get()` is implemented is only useful when the skyline is not there yet.

thestr4ng3r:
![flamegraph](https://user-images.githubusercontent.com/1460997/78586092-35d86880-783b-11ea-9d57-d8ac42f5d086.png)

thestr4ng3r:
[...]

PR #16447, opened 2020-04-06
Use RPVector for io->maps

thestr4ng3r:
**Detailed description**

Replace the SdbList of RIO.maps by an RPVector for better cache-friendliness in #14024.
Of course this does not change the `O(n^2)` complexity in that test but at least removes the linked list overhead. There are many other places where this list is linearly iterated, so using a vector is a good choice.
RVector instead of RPVector would be even better, but then the pointers wouldn't be stable anymore.

**Test plan**

Run the `db/formats/pe/65535sects` test.
Before: 26s
After: 15s

**Related issues**

#14024


ret2libc:
Wow, it's awesome the way just replacing lists with RPVector improves performance!

radare:
What about using rpvector in sdb? :p

PR #16454, opened 2020-04-07
Add GitHub actions for push to master as well

ret2libc:
Instead of using a full matrix, just add the builds we care about


XVilka:
Yes, similar approach to AppVeyor.
Can you please also ignore commits if only done on other CI files (if it's possible):
- `.builds/`
- `.travis*`
- `.appveyor.yml`

ret2libc:
@XVilka ok, done! I think this is good!

PR #16456, opened 2020-04-07
Run Fuzz Tests with r2r

thestr4ng3r:
**Detailed description**

Usage: `r2r -F bins/fuzzed @fuzz` or `make fuzz-tests` in `test/`
Unfortunately it is currently necessary to manually pass the path to the tests since we only search for `db` and going from db to the fuzz tests automatically would be `../bins/fuzzed` which seems very dirty to me.
There are still many segfaulting files there so this should be fixed before going in the CI.

### DIFF
--- a/binr/r2r/load.c
+++ b/binr/r2r/load.c
@@ -467,6 +467,14 @@ R_API RPVector *r2r_load_json_test_file(const char *file) {
 	return ret;
 }
 
+R_API void r2r_fuzz_test_free(R2RFuzzTest *test) {
+	if (!test) {
+		return;
+	}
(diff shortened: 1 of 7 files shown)


radare:
which segfaults? why it is green ? (not yet running in ci?)

previous suites considered that a different type of tests, so just using the @testgroup thing i did should be good here as well i think. lgtm we can run that in a separate CI task to track those segfaults we didnt spotted yet :?

thestr4ng3r:
I think they are way too many to make sense in a CI job yet. They are easy to reproduce locally so I will open an issue.

Issue #15662, opened 2019-12-17
Autofix feature for C suite tests


XVilka:
Exactly like the `r2r -i` of the NodeJS one.

PR #16466, opened 2020-04-07
Add interactive mode to r2r

thestr4ng3r:
The final nail in r2r.js' coffin.

Fixes https://github.com/radareorg/radare2/issues/15662

### DIFF
--- a/binr/r2r/r2r.c
+++ b/binr/r2r/r2r.c
@@ -881,14 +881,20 @@ static void interact_commands(R2RTestResultInfo *result, RPVector *fixup_results
 	if (!test->cmds.value) {
 		return;
 	}
-	char *name;
+	char *name = NULL;
 	int fd = r_file_mkstemp ("r2r-cmds", &name);
 	if (fd == -1) {
 		free (name);
 		eprintf ("Failed to open tmp file\n");
 		return;
 	}
-	write (fd, test->cmds.value, strlen (test->cmds.value));
+	size_t cmds_sz = strlen (test->cmds.value);
+	if (write (fd, test->cmds.value, cmds_sz) != cmds_sz) {
+		eprintf ("Failed to write to tmp file\n");
+		free (name);
+		close (fd);
+		return;
+	}
 	close (fd);
 
 	char *editor = r_sys_getenv ("EDITOR");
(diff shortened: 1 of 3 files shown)


binr/r2r/r2r.c
```diff
@@ -387,6 +397,19 @@ int main(int argc, char **argv) {
 	}
 	r_pvector_clear (&workers);
 
+	ut64 seconds = (r_sys_now () - time_start) / 1000000;
```
radare, line 400: use the constants!

thestr4ng3r: you wrote that code in the first place :-P

binr/r2r/r2r.c
```diff
@@ -631,3 +646,83 @@ static void print_log(R2RState *state, ut64 prev_completed, ut64 prev_paths_comp
...
+			break;
+		}
+		if (strlen (buf) != 2) {
+			goto inval;
+		}
+		switch (buf[0]) {
+		case 'f':
+			interact_fix (result, &failed_results);
+			break;
+		case 'i':
```
radare, line 700: run cfg.editor (EDITOR|vi) with a file containing the test in the right line (or maybe just do like r2r-i that deserializes and serializes the test files

binr/r2r/r2r.c
thestr4ng3r: I think we should do that change in another pr, where we move some stuff to r_util.

binr/r2r/r2r.c
thestr4ng3r: We have no `cfg.editor` in r2r because we have no core. I don't understand what you mean by the rest.

binr/r2r/r2r.c
```diff
@@ -631,3 +646,277 @@ static void print_log(R2RState *state, ut64 prev_completed, ut64 prev_paths_comp
...
+	replace_cmd_kv (result->test->path, line_begin, line_end, "BROKEN", "1", fixup_results);
+}
+
+static void interact_commands(R2RTestResultInfo *result, RPVector *fixup_results) {
+	assert (result->test->type == R2R_TEST_TYPE_CMD);
+	R2RCmdTest *test = result->test->cmd_test;
+	if (!test->cmds.value) {
+		return;
+	}
+	char *name;
```
radare, line 873: maybe worth = NULL; in theory when returning -1 name will not be initialized, so the free below it will crash

binr/r2r/r2r.c
```diff
@@ -631,3 +646,277 @@ static void print_log(R2RState *state, ut64 prev_completed, ut64 prev_paths_comp
...
+	int fd = r_file_mkstemp ("r2r-cmds", &name);
+	if (fd == -1) {
+		free (name);
+		eprintf ("Failed to open tmp file\n");
+		return;
+	}
+	write (fd, test->cmds.value, strlen (test->cmds.value));
```
radare, line 880: (void) the return on purpose or check the return value and report proper error in case it cant write

binr/r2r/r2r.c
```diff
@@ -631,3 +646,277 @@ static void print_log(R2RState *state, ut64 prev_completed, ut64 prev_paths_comp
...
+	}
+
+	char *trimmed = newcmds;
+	// strip all leading newlines
+	while (*trimmed == '\n') {
+		trimmed++;
+	}
+	size_t len = strlen (trimmed);
+	// strip all trailing newlines
+	while (len && trimmed[len - 1] == '\n') {
```
radare, line 907: that wont work on CRLF files, also we have r_str apis for that

binr/r2r/r2r.c
```diff
@@ -631,3 +646,277 @@ static void print_log(R2RState *state, ut64 prev_completed, ut64 prev_paths_comp
...
+		trimmed[len - 1] = '\0';
+		len--;
+	}
+
+	// if it's multiline we want exactly one trailing newline
+	if (strchr (trimmed, '\n')) {
+		char *tmp = newcmds;
+		newcmds = r_str_newf ("%s\n", trimmed);
```
radare, line 915: newcmds can be null and then strchr will crash

radare:
there are cornflakes in r2r.c now pls fix too

thestr4ng3r:
Cornflakes are eaten and comets are addressed.

PR #16469, opened 2020-04-07
Make dyldcache accelerator info optional ##bin

mrmacete:

**Detailed description**

This allows to open dyld cache files from iOS 13.4 for which apparently the accelerator info are missing.

What’s lost when this info is missing is the ability to efficiently detect which single library depends on which other libs when performing the filtering.

That means that for iOS 13.4 library caches, the user has to manually specify all the dependencies in the filter.

### DIFF
--- a/libr/bin/p/bin_dyldcache.c
+++ b/libr/bin/p/bin_dyldcache.c
@@ -828,35 +828,41 @@ static RList *create_cache_bins(RBinFile *bf, RBuffer *cache_buf, cache_hdr_t *h
 			return NULL;
 		}
 
-		ut16 *depArray = R_NEWS0 (ut16, accel->depListCount);
-		if (!depArray) {
-			r_list_free (target_lib_names);
-			R_FREE (target_libs);
-			r_list_free (bins);
-			R_FREE (deps);
-			R_FREE (img);
-			return NULL;
-		}
+		ut16 *depArray = NULL;
+		cache_imgxtr_t *extras = NULL;
+		if (accel) {
+			depArray = R_NEWS0 (ut16, accel->depListCount);
+			if (!depArray) {
+				r_list_free (target_lib_names);
+				R_FREE (target_libs);
+				r_list_free (bins);
+				R_FREE (deps);
+				R_FREE (img);
+				return NULL;
+			}
 
-		if (r_buf_fread_at (cache_buf, accel->depListOffset, (ut8*) depArray, "s", accel->depListCount) != accel->depListCount * 2) {
-			r_list_free (target_lib_names);
-			R_FREE (target_libs);
-			r_list_free (bins);
-			R_FREE (deps);
-			R_FREE (depArray);
-			R_FREE (img);
-			return NULL;
-		}
+			if (r_buf_fread_at (cache_buf, accel->depListOffset, (ut8*) depArray, "s", accel->depListCount) != accel->depListCount * 2) {
+				r_list_free (target_lib_names);
+				R_FREE (target_libs);
+				r_list_free (bins);
+				R_FREE (deps);
+				R_FREE (depArray);
+				R_FREE (img);
+				return NULL;
+			}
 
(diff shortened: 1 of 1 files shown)


libr/bin/p/bin_dyldcache.c
```diff
@@ -869,16 +875,18 @@ static RList *create_cache_bins(RBinFile *bf, RBuffer *cache_buf, cache_hdr_t *h
...
+				ut32 j;
+				for (j = extras[i].dependentsStartArrayIndex; depArray[j] != 0xffff; j++) {
+					bool upward = depArray[j] & 0x8000;
+					ut16 dep_index = depArray[j] & 0x7fff;
+					if (!upward) {
+						deps[dep_index]++;
+
+						char *dep_name = get_lib_name (cache_buf, &img[dep_index]);
+						eprintf ("-> %s\n", dep_name);
+						R_FREE (dep_name);
```
radare, line 888: you can use free (dep_name), no need to NULL it later because the scope is clear

libr/bin/p/bin_dyldcache.c
```diff
@@ -1713,24 +1717,26 @@ static void header(RBinFile *bf) {
...
-	bin->cb_printf ("dofSectionsCount: 0x%"PFMT64x"\n", cache->accel->dofSectionsCount);
-	bin->cb_printf ("reExportListOffset: 0x%"PFMT64x"\n", cache->accel->reExportListOffset);
-	bin->cb_printf ("reExportCount: 0x%"PFMT64x"\n", cache->accel->reExportCount);
-	bin->cb_printf ("depListOffset: 0x%"PFMT64x"\n", cache->accel->depListOffset);
-	bin->cb_printf ("depListCount: 0x%"PFMT64x"\n", cache->accel->depListCount);
-	bin->cb_printf ("rangeTableOffset: 0x%"PFMT64x"\n", cache->accel->rangeTableOffset);
-	bin->cb_printf ("rangeTableCount: 0x%"PFMT64x"\n", cache->accel->rangeTableCount);
-	bin->cb_printf ("dyldSectionAddr: 0x%"PFMT64x"\n", cache->accel->dyldSectionAddr + slide);
+	if (cache->accel) {
+		bin->cb_printf ("\nacceleration info:\n");
```
radare, line 1721: Maybe Add a local variable PrintfCallback p = bin->cb_printf; and just do p("");

libr/bin/p/bin_dyldcache.c
```diff
@@ -1713,24 +1717,26 @@ static void header(RBinFile *bf) {
...
+		bin->cb_printf ("version: 0x%"PFMT64x"\n", cache->accel->version);
```
radare, line 1722: what about printing this in json format using the pj api? it can be formated like k=v easily and you dont need to mess with the format strings

libr/bin/p/bin_dyldcache.c
```diff
@@ -828,35 +828,41 @@ static RList *create_cache_bins(RBinFile *bf, RBuffer *cache_buf, cache_hdr_t *h
...
-			R_FREE (deps);
-			R_FREE (img);
-			return NULL;
-		}
+		ut16 *depArray = NULL;
+		cache_imgxtr_t *extras = NULL;
+		if (accel) {
+			depArray = R_NEWS0 (ut16, accel->depListCount);
+			if (!depArray) {
+				r_list_free (target_lib_names);
```
radare, line 836: better use goto, thise collection of frees can be written once at the bottom of the function

libr/bin/p/bin_dyldcache.c
radare: imho no need to do this change in this pr,j ust to have it in mind, because this way the data can be imported with r2pipe easily

mrmacete: yeah totally, there's also way more stuff we could expose here

PR #16476, opened 2020-04-08
Allow to run travis jobs if special branch names are used

ret2libc:
It may be useful from time to time to see the results of Travis jobs in PRs too.
These are the rules:
- if the branch name starts with `s390x-`, the S390X travis job is run
- if the branch name starts with `wasm-`, the WASM travis job is run
- if the branch name starts with `ppc64-`, the PPC64 travis job is run
- if the branch name starts with `arm64-`, the ARM64 travis job is run


radare:
ok

PR #16481, opened 2020-04-08
Run newshell tests when PR is newshell-*

ret2libc:
Split the workflow in a separate file


ret2libc:
I am not sure why it runs two times the number of GH jobs... Maybe to test both the old workflow and the one modified in this PR? :/

Issue #16493, opened 2020-04-09
Radare2 should fix various issue found by lgtm


08A:
### Work environment
LGTM, i just create this issue so we can talk how to solve various problem found by this tools

### Expected behavior
Green badge

### Actual behavior
Yellow badge

### Steps to reproduce the behavior 
click on the lgtm badge

### Additional Logs, screenshots, source-code,  configuration dump, ...
https://lgtm.com/projects/g/radareorg/radare2/alerts/?mode=list

08A:
What is the acceptable range for this allocation.
maybe 1-4096?

- https://lgtm.com/projects/g/radareorg/radare2/snapshot/e0dc22a95aa72cac584423f8d558cde3bb73159a/files/shlr/sdb/src/sdb.c?sort=name&dir=ASC&mode=heatmap#x876a12223fc17e2e:1
- https://lgtm.com/projects/g/radareorg/radare2/snapshot/e0dc22a95aa72cac584423f8d558cde3bb73159a/files/shlr/sdb/src/ht_inc.c?sort=name&dir=ASC&mode=heatmap#xd2e2d3f23ed695db:1
- https://lgtm.com/projects/g/radareorg/radare2/snapshot/e0dc22a95aa72cac584423f8d558cde3bb73159a/files/libr/socket/run.c?sort=name&dir=ASC&mode=heatmap#x348f353161ff0fc5:1

XVilka:
See also https://github.com/radareorg/radare2/issues/16130

08A:
I will focus on lgtm, the feedback loop is shorter.

eagleoflqj:
I'm doing the same. imho PRs should be small which makes them easy to get merged.

radare:
Can we close this issue?

https://lgtm.com/projects/g/radareorg/radare2/alerts/?mode=list&severity=error

those are probably the important ones but i dont see them that scary because all the heap implementations i know return NULL on failure, .. well except debian :D but those are not blocking the release, so moving to 4.5 or closing. because lgtm is always there reporting all the issues

08A:
Small warning can create weird bug I believe that fix all minor lgtm warning will improve the code base's quality. It's why for me this issue should rest open until all the warning are solved (or reviewed).
This issue was open to centralized the work on the lgtm warning.

radare:
Totally agree

eagleoflqj:
[...]

PR #16497, opened 2020-04-09
Fix various lgtm warning

08A:

**Detailed description**
Start solving lgtm's warning. Some file don't respect coding style, so there is a big diff.

**Test plan**
Rerun lgtm

**Work on issues**
 #16493

### DIFF
--- a/libr/asm/asm.c
+++ b/libr/asm/asm.c
@@ -617,8 +617,8 @@ R_API RAsmCode* r_asm_mdisassemble(RAsm *a, const ut8 *buf, int len) {
 	ut64 pc = a->pc;
 	RAsmOp op;
 	ut64 idx;
-	int ret, slen;
-	const int addrbytes = a->user ? ((RCore *)a->user)->io->addrbytes : 1;
+	size_t ret, slen;
+	const size_t addrbytes = a->user? ((RCore *)a->user)->io->addrbytes: 1;
 
 	if (!(acode = r_asm_code_new ())) {
 		return NULL;
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -2826,7 +2826,8 @@ const struct symbol_t *MACH0_(get_symbols)(struct MACH0_(obj_t) *bin) {
 }
 
 static int parse_import_ptr(struct MACH0_(obj_t) *bin, struct reloc_t *reloc, int idx) {
-	int i, j, sym, wordsize;
+	int i, j;
+	size_t sym, wordsize;
 	ut32 stype;
 	wordsize = MACH0_(get_bits)(bin) / 8;
 	if (idx < 0 || idx >= bin->nsymtab) {
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -2826,8 +2826,7 @@ const struct symbol_t *MACH0_(get_symbols)(struct MACH0_(obj_t) *bin) {
 }
 
 static int parse_import_ptr(struct MACH0_(obj_t) *bin, struct reloc_t *reloc, int idx) {
-	int i, j;
-	ssize_t sym;
+	int i, j, sym;
 	size_t wordsize;
 	ut32 stype;
 	wordsize = MACH0_(get_bits)(bin) / 8;
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -2827,7 +2827,8 @@ const struct symbol_t *MACH0_(get_symbols)(struct MACH0_(obj_t) *bin) {
 
 static int parse_import_ptr(struct MACH0_(obj_t) *bin, struct reloc_t *reloc, int idx) {
 	int i, j;
-	size_t sym, wordsize;
+	ssize_t sym;
+	size_t wordsize;
 	ut32 stype;
 	wordsize = MACH0_(get_bits)(bin) / 8;
 	if (idx < 0 || idx >= bin->nsymtab) {
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -4671,7 +4671,7 @@ static int cmd_print(void *data, const char *input) {
 	ut32 tbs = core->blocksize;
 	ut64 n, off, from, to, at, ate, piece;
 	ut64 tmpseek = UT64_MAX;
-	const size_t addrbytes = core->io->addrbytes;
+	const int addrbytes = core->io->addrbytes;
 	i = l = len = ret = 0;
 	n = off = from = to = at = ate = piece = 0;
 	PJ *pj = NULL;
@@ -5590,7 +5590,7 @@ l = use_blocksize;
 						core->num->value = r_core_print_disasm (core->print,
 							core, addr, block1, addrbytes * l, l, 0, 1, formatted_json, NULL, NULL);
 					} else {
-						eprintf ("Cannot allocate %ld byte(s)\n", addrbytes * l);
+						eprintf ("Cannot allocate %d byte(s)\n", addrbytes * l);
 					}
 				} else {
 					ut8 *buf = core->block;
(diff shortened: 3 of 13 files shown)


libr/bin/format/mach0/mach0.c
```diff
@@ -2826,7 +2826,8 @@ const struct symbol_t *MACH0_(get_symbols)(struct MACH0_(obj_t) *bin) {
 }
 
 static int parse_import_ptr(struct MACH0_(obj_t) *bin, struct reloc_t *reloc, int idx) {
-	int i, j, sym, wordsize;
+	int i, j;
+	size_t sym, wordsize;
```
ret2libc, line 2830: Is this right? It seems to me the value of `j` is assigned to `sym` few lines below, so they should be of the same time. Also, `sym` initially has -1 as value, so size_t is probably not the right type.

libr/core/cmd_print.c
```diff
@@ -4671,7 +4671,7 @@ static int cmd_print(void *data, const char *input) {
 	ut32 tbs = core->blocksize;
 	ut64 n, off, from, to, at, ate, piece;
 	ut64 tmpseek = UT64_MAX;
-	const int addrbytes = core->io->addrbytes;
+	const size_t addrbytes = core->io->addrbytes;
```
ret2libc, line 4674: `addrbytes` is `int` in `RIO`, why this change?

libr/asm/asm.c
```diff
@@ -617,8 +617,8 @@ R_API RAsmCode* r_asm_mdisassemble(RAsm *a, const ut8 *buf, int len) {
 	ut64 pc = a->pc;
 	RAsmOp op;
 	ut64 idx;
-	int ret, slen;
-	const int addrbytes = a->user ? ((RCore *)a->user)->io->addrbytes : 1;
+	size_t ret, slen;
+	const size_t addrbytes = a->user? ((RCore *)a->user)->io->addrbytes: 1;
```
ret2libc, line 621: addrbytes in RIO is int, why this change?

libr/core/visual.c
```diff
@@ -254,7 +254,7 @@ static bool __core_visual_gogo (RCore *core, int ch) {
 				break;
 			}
 			(void)p->consbind.get_size (&scr_rows);
-			int scols = r_config_get_i (core->config, "hex.cols");
+			size_t scols = r_config_get_i (core->config, "hex.cols");
```
ret2libc, line 257: `r_config_get_i` returns a `ut64`, not `size_t`

libr/socket/run.c
```diff
@@ -162,15 +162,16 @@ static char *getstr(const char *src) {
...
-				rep = atoi (src + 1);
+				rep = strtol (src + 1, NULL, 10);
 				len = strlen (pat);
 				if (rep > 0) {
-					char *buf = malloc (rep);
+					char *buf = malloc (rep * sizeof(char));
 					if (buf) {
-						for (i = 0; i < rep; i++) {
-							buf[i] = pat[i%len];
+						for (long i = 0; i < rep; i++) {
```
ret2libc, line 173: do not declar the var inside the for-loop. Apparently there are compilers which complain and we don't have the right flags in place to fix that at the moment.

libr/bin/format/mach0/mach0.c
08A: Indeed i will replace his type by ssize_t

libr/core/cmd_print.c
08A: We have a possible overflow bellow, if the type is int.

```
size_t ... = int * int;
                 ^
                 | -- possible overflow
```

libr/asm/asm.c
08A: same than before

libr/core/visual.c
08A: size_t >= ut64 i believe modern C should use size_t when there is a choice.

eagleoflqj: No size_t is 32-bit on i386. Try `gcc -m32 a.c` to print sizeof(size_t).

ret2libc: what @eagleoflqj says is true, moreover even just for consistency if a function returns `ut64` you should use the same type if possible.

08A: Okay, it is fixed

08A:
I was wandering why we use ut64, ut32 if there is stdint.h.

ret2libc:
> I was wandering why we use ut64, ut32 if there is stdint.h.

@radare  ^

XVilka:
Because stdint isn't (wasn't) on the every supported platform when we
started to use it* types. Radare2 supports A LOT of compilers and
platforms, keep this in mind.

libr/core/cmd_print.c
ret2libc: I see, but I'm not sure that's the proper way to fix it (since you store the result in a `int` anyway). Maybe just check for overflows before doing the multiplication. It is probably wrong anyway if the number is so big.

eagleoflqj: What about a conversion like (size_t)int * int? Or change addrbytes to some ut if it doesn't use -1 to indicate error.

libr/asm/asm.c
radare: ? you should set RIO->addrbytes as size_t or whatever then

libr/bin/format/mach0/mach0.c
```diff
@@ -2826,7 +2826,9 @@ const struct symbol_t *MACH0_(get_symbols)(struct MACH0_(obj_t) *bin) {
 }
 
 static int parse_import_ptr(struct MACH0_(obj_t) *bin, struct reloc_t *reloc, int idx) {
-	int i, j, sym, wordsize;
+	int i, j;
+	ssize_t sym;
```
radare, line 2830: sym is signed?

08A: The default value is -1.

libr/core/cmd_print.c
radare: that wont solve the problem. we have the UT64_ADD_OVFCHK macros that detect the overflow before it happens. but we have it for + not for *.  imho addrbytes shouldnt be larger than 4 or 8, so we probably dont need to be that paranoic on handling overflows in this operation

libr/core/cmd_print.c
```diff
@@ -5590,7 +5590,7 @@ l = use_blocksize;
 						core->num->value = r_core_print_disasm (core->print,
 							core, addr, block1, addrbytes * l, l, 0, 1, formatted_json, NULL, NULL);
 					} else {
-						eprintf ("Cannot allocate %d byte(s)\n", addrbytes * l);
+						eprintf ("Cannot allocate %ld byte(s)\n", addrbytes * l);
```
radare, line 5593: as i said, addrbytes is int in r_io, i would probably go for casting (int) instead of changing this format, the value of addrbytes shouldnt be larger than 8

libr/socket/run.c
```diff
@@ -162,15 +162,16 @@ static char *getstr(const char *src) {
...
-				int i, len, rep;
+				size_t len;
+				long i, rep;
 				*pat++ = 0;
-				rep = atoi (src + 1);
+				rep = strtol (src + 1, NULL, 10);
 				len = strlen (pat);
 				if (rep > 0) {
-					char *buf = malloc (rep);
+					char *buf = malloc (rep * sizeof(char));
```
radare, line 171: missing space before (

libr/core/cmd_print.c
08A: swap back to the old type

radare:
This pull request introduces 1 alert

radare:
thats the introduced alert

Screenshot 2020-04-11 at 20 19 52

radare:
this should be merged tomorrow or before, lets focus on fixing issues for the release now

thanks!

08A:
> 
> 
> lgtm except for the remaining issue . bear in mind that sizeof(char) is always 1, so the multiplication doesnt solves anything there

I replaced atoi with strtol, which is a better function.
I didn't use sizeof(char) to solve the issue, it was just for consistency.

radare:
It breaks the windows build

08A:
The problem was the use of ssize_t.

radare:
This pull request introduces 1 alert

PR #16521, opened 2020-04-10
Remove unused RAnalFunction Members

thestr4ng3r:


libr/core/carg.c
```diff
@@ -193,16 +191,7 @@ R_API void r_core_print_func_args(RCore *core) {
 				argcnt++;
 			}
 		} else {
```
radare, line 193: What about fixing the function or removing it completely?

libr/anal/function.c
```diff
@@ -93,11 +89,9 @@ R_API void r_anal_function_free(void *_fcn) {
 	ht_pp_delete (anal->ht_name_fun, fcn->name);
 
 	free (fcn->name);
-	free (fcn->attr);
 	fcn->bbs = NULL;
 	free (fcn->fingerprint);
 	r_anal_diff_free (fcn->diff);
-	free (fcn->args);
```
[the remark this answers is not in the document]
thestr4ng3r: I don't think it will be a problem to re-add a single line that says `RList *args;`, so not worth keeping imo.

libr/core/carg.c
thestr4ng3r: I think the void fallback like this is ok for the current implementation.

libr/anal/function.c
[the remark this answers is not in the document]
radare: ok

libr/core/carg.c
radare: what about printing nargs = 4 so at least we can get the values? anyway im fine for removing this beacuse all this stuff needs to be refactored and cleaned up

thestr4ng3r: sgtm

radare: So wat do

thestr4ng3r: Nothing, just did it

PR #16523, opened 2020-04-10
Improve COFF symbol info

pelijah:

**Detailed description**
1. Mark sections symbols
2. Mark absolute symbols
3. Set paddr = UT64_MAX for imports

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/bin/format/coff/coff_specs.h
+++ b/libr/bin/format/coff/coff_specs.h
@@ -81,6 +81,10 @@
 #define COFF_SCN_MEM_READ		0x40000000
 #define COFF_SCN_MEM_WRITE		0x80000000
 
+#define COFF_SYM_SCNUM_UNDEF 		0
+#define COFF_SYM_SCNUM_ABS		((unsigned short)-1)
+#define COFF_SYM_SCNUM_DEBUG		((unsigned short)-2)
+
 #define COFF_SYM_TYPE_NULL		0
 #define COFF_SYM_TYPE_VOID		1
 #define COFF_SYM_TYPE_CHAR		2
--- a/libr/bin/p/bin_coff.c
+++ b/libr/bin/p/bin_coff.c
@@ -46,6 +46,7 @@ static RBinAddr *binsym(RBinFile *bf, int sym) {
 static bool _fill_bin_symbol(RBin *rbin, struct r_bin_coff_obj *bin, int idx, RBinSymbol **sym) {
 	RBinSymbol *ptr = *sym;
 	struct coff_symbol *s = NULL;
+	struct coff_scn_hdr *sc_hdr = NULL;
 	if (idx < 0 || idx > bin->hdr.f_nsyms) {
 		return false;
 	}
@@ -61,6 +62,11 @@ static bool _fill_bin_symbol(RBin *rbin, struct r_bin_coff_obj *bin, int idx, RB
 	ptr->forwarder = "NONE";
 	ptr->bind = R_BIN_BIND_LOCAL_STR;
 	ptr->is_imported = false;
+	if (s->n_scnum < bin->hdr.f_nscns + 1 && s->n_scnum > 0) {
+		//first index is 0 that is why -1
+		sc_hdr = &bin->scn_hdrs[s->n_scnum - 1];
+		ptr->paddr = sc_hdr->s_scnptr + s->n_value;
+	}
 
 	switch (s->n_sclass) {
 	case COFF_SYM_CLASS_FUNCTION:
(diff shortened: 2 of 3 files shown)


test/new/db/formats/coff
```diff
@@ -6,34 +6,34 @@ iS
...
-19  0x00000000 NONE FUNC     sym.imp.BitData
-20  0x00000000 NONE FUNC     sym.imp.BitData
-21  0x00000000 NONE FUNC     sym.imp.BitData
-22  0x00000000 NONE FUNC     _TIFFSetCompressionScheme
-23  0x00000000 NONE FUNC     __TIFFDataSize
-24  0x00000000 NONE FUNC     __TIFFCheckMalloc
-25  0x00000000 NONE UNK      __fltused
+nth vaddr bind type lib name
+----------------------------
+0   -1    NONE FUNC     __TIFFmalloc
```
pelijah, line 11: Is it ok?

libr/bin/p/bin_coff.c
```diff
@@ -73,25 +79,40 @@ static bool _fill_bin_symbol(RBin *rbin, struct r_bin_coff_obj *bin, int idx, RB
 		ptr->type = R_BIN_TYPE_SECTION_STR;
 		break;
 	case COFF_SYM_CLASS_EXTERNAL:
-		ptr->is_imported = !s->n_scnum;
-		ptr->bind = ptr->is_imported? "NONE": R_BIN_BIND_GLOBAL_STR;
+		if (s->n_scnum == COFF_SYM_SCNUM_UNDEF) {
+			ptr->is_imported = true;
+			ptr->paddr = UT64_MAX;
+			ptr->bind = "NONE";
+		}
```
radare, line 86: Wrong indent

libr/bin/format/coff/coff_specs.h
```diff
@@ -81,6 +81,10 @@
 #define COFF_SCN_MEM_READ		0x40000000
 #define COFF_SCN_MEM_WRITE		0x80000000
 
+#define COFF_SYM_SCNUM_UNDEF 		0
+#define COFF_SYM_SCNUM_ABS		((unsigned short)-1)
```
radare, line 85: Better use 0xfff syntax imho

test/new/db/formats/coff
radare: Hows that cant be resolved?

libr/bin/p/bin_coff.c
```diff
@@ -73,25 +79,40 @@ static bool _fill_bin_symbol(RBin *rbin, struct r_bin_coff_obj *bin, int idx, RB
...
 		break;
 	case COFF_SYM_CLASS_STATIC:
-		ptr->type = DTYPE_IS_FUNCTION (s->n_type)
-			? R_BIN_TYPE_FUNC_STR
-			: R_BIN_TYPE_UNKNOWN_STR;
+		if (s->n_scnum == COFF_SYM_SCNUM_ABS) {
+			ptr->type = "ABS";
+			ptr->paddr = UT64_MAX;
+			ptr->name = r_str_newf ("%s-0x%08x", coffname, s->n_value);
+			if (ptr->name) {
```
radare, line 99: I think this condition is wrong

libr/bin/p/bin_coff.c
```diff
@@ -73,25 +79,40 @@ static bool _fill_bin_symbol(RBin *rbin, struct r_bin_coff_obj *bin, int idx, RB
...
+				R_FREE (coffname);
+			} else {
+				ptr->name = coffname;
+			}
+		} else if (sc_hdr && !memcmp (sc_hdr->s_name, s->n_name, 8)) {
```
radare, line 104: Isn strncmp better here?

libr/bin/p/bin_coff.c
pelijah: Why? If `r_str_newf` returns NULL I just restore previous value. Otherwise heap allocated `coffname` should be released.

libr/bin/p/bin_coff.c
pelijah: No, `name` field is a union.

test/new/db/formats/coff
pelijah: What can't be resolved?

radare:
The address of the imports so the disassembly can use it to show meaningful calls. Imma bit out of coff and maybe those imports are loaded as relocs by the runtime linker. But in case the binary have a place to hold the trampolines to jump to the import like the elf plt this should be resolved there

radare:
I dont see why strncmp shouldnt work on an union

radare:
In case strnewf fails its probably better to quit this loop and shutdown your pc. So coffname is never used?

libr/bin/p/bin_coff.c
pelijah: First 4 bytes of that field might be `0`.

libr/bin/p/bin_coff.c
pelijah: Ok I will remove this check. `coffname` is saved to `ptr->name`.

test/new/db/formats/coff
pelijah: > The address of the imports so the disassembly can use it to show meaningful calls. Imma bit out of coff and maybe those imports are loaded as relocs by the runtime linker. But in case the binary have a place to hold the trampolines to jump to the import like the elf plt this should be resolved there

In case of `obj` files imports are just names for relocs but binary doesn't have trampolines. Linker (like `ld`; not a runtime linker) modifies call/jump instructions directly inside of place of use.

pelijah:
But... but I didn't update my PR...

radare:
ouch, sorry misread

PR #16526, opened 2020-04-11
Fix dup af+ removing function from hts

thestr4ng3r:
**Detailed description**

This fixes the following issue:
```
$ r2 /bin/ls
> af;.afi*
# cannot add basic block errors
> af;.afi*
# no errors at all
> af;.afi*
# cannot add basic block errors again
```

or more narrowed down:
```
[0x00000000]> af+ 0x1337 func
[0x00000000]> af+ 0x1337 func
Invalid function name 'func' at 0x00001337
Cannot add function (duplicated)
[0x00000000]> af+ 0x1337 func
[0x00000000]> af+ 0x1337 func
Invalid function name 'func' at 0x00001337
Cannot add function (duplicated)
[0x00000000]>
```

The expected behavior is that all subsequent `af+` calls show the error.
This is however not only about error printing, in fact the duplicated `af+` would remove the function from the address and name hashtables which is a problem.

**Closing issues**

#15453 (doesn't directly solve the project-related issue but I think we can close with r2db coming soon^TM)

### DIFF
--- a/libr/anal/function.c
+++ b/libr/anal/function.c
@@ -89,8 +89,12 @@ R_API void r_anal_function_free(void *_fcn) {
 	r_list_free (fcn->bbs);
 
 	RAnal *anal = fcn->anal;
-	ht_up_delete (anal->ht_addr_fun, fcn->addr);
-	ht_pp_delete (anal->ht_name_fun, fcn->name);
+	if (ht_up_find (anal->ht_addr_fun, fcn->addr, NULL) == _fcn) {
+		ht_up_delete (anal->ht_addr_fun, fcn->addr);
+	}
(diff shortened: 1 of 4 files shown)


radare:
looks red. pls confirm if the referenced issue can be closed

radare:
Screenshot 2020-04-11 at 20 15 37

IMAGE 2020-04-11 20:15:49

thestr4ng3r:
Yeah it slightly changes analysis behavior. But that is because before there were some "leaked" duplicate functions at the same address, so it is more correct now.

Well as I have written, the issue itself is about projects, which is not fixed by this. But since trying to fix current projects at this point would be a waste of time, I think we can close it.

PR #16530, opened 2020-04-11
Refactor of elf.c and remove reference to section name in get_import_addr

08A:

**Detailed description**
~2 tests fail.~
`symbols with no sections header information 3`: i believe the test is wrong, before the refacto the return addr of get_import_addr was always -1 because there was no section name.

`ELF: imports partial`: i believe the return value (`get_import_addr_x86_manual`) when to .plt.got entry can't be found should be -1 not a garbage value.

PS: if the return value of `get_import_addr_x86_manual` is the variable `plt_sym_addr` another test fail.

**Test plan**
Run tests

**Closing issues**
#12732


libr/bin/format/elf/elf.h
```diff
@@ -80,19 +99,19 @@ typedef struct r_bin_elf_lib_t {
 struct Elf_(r_bin_elf_obj_t) {
 	Elf_(Ehdr) ehdr;
 	Elf_(Phdr)* phdr;
-	Elf_(Shdr)* shdr;
+	Elf_(Shdr) * shdr;
```
ret2libc, line 102: either keep it as it was, or fix it with `Elf_(Shdr) *shdr;`

libr/bin/p/bin_elf.inc
```diff
@@ -591,7 +591,7 @@ static RBinReloc *reloc_convert(struct Elf_(r_bin_elf_obj_t) *bin, RBinElfReloc
 	r->paddr = rel->offset;
 
 	#define SET(T) r->type = R_BIN_RELOC_ ## T; r->additive = 0; return r
-	#define ADD(T, A) r->type = R_BIN_RELOC_ ## T; r->addend += A; r->additive = !rel->is_rela; return r
+	#define ADD(T, A) r->type = R_BIN_RELOC_ ## T; r->addend += A; r->additive = !rel->rel_mode; return r
```
ret2libc, line 594: since `rel_mode` is not a boolean any more, I'm not sure `!rel_mode` makes sense.

libr/bin/format/elf/elf.c
```diff
@@ -26,42 +26,47 @@
...
+	if (bin->verbose) \
+	R_LOG_WARN
 
 #define MAX_REL_RELA_SZ (sizeof (Elf_(Rel)) > sizeof (Elf_(Rela))? sizeof (Elf_(Rel)): sizeof (Elf_(Rela)))
 
-#define READ8(x, i) r_read_ble8((x) + (i)); (i) += 1
-#define READ16(x, i) r_read_ble16((x) + (i), bin->endian); (i) += 2
-#define READ32(x, i) r_read_ble32((x) + (i), bin->endian); (i) += 4
-#define READ64(x, i) r_read_ble64((x) + (i), bin->endian); (i) += 8
+#define READ8(x, i)              \
```
ret2libc, line 35: don't do all this style fixes here please... if you want to clean the file do a separate PR with just whitespace/style fixes, otherwise it is hard for me to review what you did really change.

ret2libc:
Ah, @08A please rebase on top of current master as well.

libr/bin/format/elf/elf.h
08A: How this seems to be the fault of `./sys/pre-commit-indent.sh`

ret2libc: yeah... unfortunately it does not work as it should :(

08A:
What do you think of `ELF: imports partial` test?

08A:
Should i rewrite `symbols with no sections header information 3`?

ret2libc:
> Should i rewrite `symbols with no sections header information 3`?

Yes, you should fix the test. Your results seem good.

> What do you think of ELF: imports partial test?

I see these changes:
```
--- /tmp/old    2020-04-14 12:15:46.398818567 +0200
+++ /tmp/new    2020-04-14 12:15:13.556235229 +0200
@@ -17,7 +17,7 @@
 15  0x00051fc8 GLOBAL FUNC       unlink
 16  0x00051fd0 GLOBAL FUNC       strncpy
 17  0x00051fd8 GLOBAL FUNC       strncmp
-18  0x00443a32 WEAK   NOTYPE     _ITM_deregisterTMCloneTable
+18  0x00000000 WEAK   NOTYPE     _ITM_deregisterTMCloneTable
 19  0x00051fe0 GLOBAL FUNC       _exit
 20  0x00051fe8 GLOBAL FUNC       strcpy
 21  0x00051ff0 GLOBAL FUNC       mkdir
@@ -85,7 +85,7 @@
 83  0x000521e0 GLOBAL FUNC       ftello64
 84  0x000521e8 GLOBAL FUNC       acos
 85  0x000521f0 GLOBAL FUNC       read
-86  0x00443a32 GLOBAL FUNC       __libc_start_main
+86  0x00000000 GLOBAL FUNC       __libc_start_main
 87  0x000521f8 GLOBAL FUNC       srand
 88  0x00052200 GLOBAL FUNC       memcmp
 89  0x00052208 GLOBAL FUNC       fgets
@@ -106,7 +106,7 @@
 104 0x00052280 GLOBAL FUNC       sigemptyset
 105 0x00052288 GLOBAL FUNC       ftell
 106 0x00052290 GLOBAL FUNC       feof
-107 0x00443a32 WEAK   NOTYPE     __gmon_start__
+107 0x00000000 WEAK   NOTYPE     __gmon_start__
 108 0x00052298 GLOBAL FUNC       umask
 109 0x000522a0 GLOBAL FUNC       fopen64
 110 0x000522a8 GLOBAL FUNC       strtol
@@ -168,7 +168,7 @@
 166 0x00052468 GLOBAL FUNC       perror
 167 0x00052470 GLOBAL FUNC       strtok
 168 0x00052478 GLOBAL FUNC       sysconf
-169 0x00443a32 WEAK   NOTYPE     _Jv_RegisterClasses
+169 0x00000000 WEAK   NOTYPE     _Jv_RegisterClasses
 170 0x00052480 GLOBAL FUNC       towlower
 171 0x00052488 GLOBAL FUNC       rename
 172 0x00052490 GLOBAL FUNC       tgetstr
@@ -190,7 +190,7 @@
 188 0x00052510 GLOBAL FUNC       fwrite
 189 0x00052518 GLOBAL FUNC       lseek64
 190 0x00052520 GLOBAL FUNC       __fprintf_chk
-191 0x00443a32 WEAK   NOTYPE     _ITM_registerTMCloneTable
+191 0x00000000 WEAK   NOTYPE     _ITM_registerTMCloneTable
 192 0x00052528 GLOBAL FUNC       sqrt
 193 0x00052530 GLOBAL FUNC       sigaltstack
 194 0x00052538 GLOBAL FUNC       strerror
```
I think your changes are still good even in this case. The old values may interpreted as real addresses, but they are not. They are just useless values, so I think it's better to have 0s. IDA creates a fake section for those entries, but I think 0 for now is good. Does that sound good?

08A:
> I think your changes are still good even in this case. The old values may interpreted as real addresses, but they are not. They are just useless values, so I think it's better to have 0s. IDA creates a fake section for those entries, but I think 0 for now is good. Does that sound good?

perfect i will fix that

libr/bin/format/elf/elf.c
ret2libc: Please see this before I review.

08A: You want to revert this change?

ret2libc: If possible, yes. There are already many changes and I don't want to risk overlooking something because of those style fixes.

08A: > If possible, yes. There are already many changes and I don't want to risk overlooking something because of those style fixes.
I fixed some macro, i you see other style fixes i will try to fix them fast

radare:
what about -1 instead 0 ?

radare:
otherwise lgtm

ret2libc:
I'm not sure whether you already did that, but what I meant was to remove all style fixes, not only the macro ones. The diff is full of whitespace changes which make the review very hard. Please try to keep the PR focused on one "thing" so it's easier to understand what's changing ;)

08A:
> I'm not sure whether you already did that, but what I meant was to remove all style fixes, not only the macro ones. The diff is full of whitespace changes which make the review very hard. Please try to keep the PR focused on one "thing" so it's easier to understand what's changing ;)

I tried to fix some white space problem.
Some change in indentation can't be revert, at the beginning this was not indented with tab.

08A:
> what about -1 instead 0 ?

You talk about the default value of dyn_info entries?

08A:
For lgtm i was thinking about removing the issue and add an issue for the last section name references in get_import_addr

libr/bin/format/elf/elf.c
```diff
@@ -1277,454 +1296,499 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
-static ut64 get_import_addr_ppc(ELFOBJ *bin, struct ht_rel_t *rel, RBinElfSection *plt_section, int nrel) {
-	ut8 buf[4] = { 0 };
-	if (!plt_section) {
-		return -1;
-	}
-	int len = r_buf_read_at (bin->b, plt_section->offset, buf, sizeof (buf));
-	if (len < 4) {
-		return -1;
-	}
+static bool sectionIsInvalid(ELFOBJ *bin, RBinElfSection *sect) {
```
ret2libc, line 1379: it's usually better, AFAIK, to use positive checks. So `sectionIsValid` seems better and then if you need the opposite you just use `!sectionIsValid`, which "reads" better IMO.

libr/bin/format/elf/elf.c
```diff
@@ -1277,454 +1296,499 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
-	return UT64_MAX;
+	free (relocs);
+	return rel_cache;
 }
 
-static ut64 get_import_addr(ELFOBJ *bin, int sym) {
-	int nrel;
-
-	if ((!bin->shdr || !bin->strtab) && !bin->phdr) {
+static ut64 get_got_entrie(ELFOBJ *bin, RBinElfReloc *rel) {
```
ret2libc, line 1546: typo? `get_got_entries`?

libr/bin/format/elf/elf.c
```diff
@@ -1277,454 +1296,499 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
-	case EM_SPARC32PLUS:
-		return get_import_addr_sparc (bin, rel, plt_section);
-	case EM_RISCV:
-		return get_import_addr_riscv (bin, rel, plt_section);
+static bool is_thumb_symbol(ut64 plt_addr) {
+	return plt_addr & 1;
+}
+
+static ut64 get_import_addr_arm(ELFOBJ *bin, RBinElfReloc *rel) {
+	ut64 got_addr = bin->dyn_info.dt_pltgot;
```
ret2libc, line 1575: what if we are working with a ET_REL object (`*.o` files)? They don't have dynamic info, so we should still rely on sections, in some cases, right?

libr/bin/format/elf/elf.c
```diff
@@ -2642,295 +2706,7 @@ char *Elf_(r_bin_elf_get_rpath)(ELFOBJ *bin) {
 	return ret;
 }
 
-static bool has_valid_section_header(ELFOBJ *bin, size_t pos) {
```
ret2libc, line 2645: why did you move all this code up? It makes review very hard :( there are way too many changes in this PR and some are just "moving around" others are real changes.

libr/bin/format/elf/elf.c
```diff
@@ -151,8 +138,6 @@ static bool init_ehdr(ELFOBJ *bin) {
 			" EM_ARC_COMPACT=93, EM_XTENSA=94, EM_VIDEOCORE=95, EM_TMM_GPP=96, EM_NS32K=97,"
 			" EM_TPC=98, EM_SNP1K=99, EM_ST200=100, EM_IP2K=101, EM_MAX=102, EM_CR=103,"
 			" EM_F2MC16=104, EM_MSP430=105, EM_BLACKFIN=106, EM_SE_C33=107, EM_SEP=108,"
-			" EM_ARCA=109, EM_UNICORE=110, EM_EXCESS=111, EM_DXP=112, EM_ALTERA_NIOS2=113,"
```
ret2libc, line 154: why?

ret2libc:
@08A could you cherry pick https://github.com/ret2libc/radare2/commit/53d56cc7f847344280bb6d07e9f3e8e4c1ed25b1 please? It will remove a lot of unneeded changes and will make the diff smaller. For the future, it's better if you separate fixing stuff from doing mass style-fixes.

libr/bin/format/elf/elf.c
08A: I get the value of a got entrie, in some arch the addr of the plt is in the got.

ret2libc: yep, I'm just saying this should probably be either `get_got_entry` (singular) or `get_got_entries` (plural) ;)

libr/bin/format/elf/elf.c
```diff
@@ -16,7 +16,7 @@
 #define IFDBG if (DO_THE_DBG)
 #define IFINT if (0)
 
-#define MIPS_PLT_OFFSET 108
+#define MIPS_PLT_OFFSET 0x20
```
ret2libc, line 19: why this change?

libr/bin/format/elf/elf.c
```diff
@@ -1277,454 +1296,499 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
+		return UT64_MAX;
+	}
+
+	ut64 plt_addr = get_got_entrie (bin, rel);
+	if (plt_addr == UT64_MAX) {
+		return UT64_MAX;
+	}
+
+	ut64 pos = (rel->rva - got_addr - 0x2 * WORDSIZE) / WORDSIZE;
+	return plt_addr + MIPS_PLT_OFFSET + pos * 0x10;
```
ret2libc, line 1676: why are you using `MIPS_PLT_OFFSET` considering the function is `get_import_addr_riscv`? I think it was previously using `RISCV_PLT_OFFSET`.

libr/bin/format/elf/elf.c
08A: Could you had a tests for this case? It will simplify the development for me.

libr/bin/format/elf/elf.c
08A: the layout was 
-----
rel_cache_new
---
Elf_(r_bin_elf_get_relocs)
---

My only choice was move up or down some function

libr/bin/format/elf/elf.c
08A: I mess up with vim

libr/bin/format/elf/elf.c
08A: This value is incorrect for the current implementation

ret2libc:
@08A do you need anything from me on this? If you have any concern, please raise them ;)

08A:
> @08A do you need anything from me on this? If you have any concern, please raise them ;)

I just found that i forget about the SPARC architecture, but we don't have any test for this architecture. If you could craft on executable for testing purpose, it will help me implement this feature faster.

You talk about *.o files do you have an example? (a test)

ret2libc:
> > @08A do you need anything from me on this? If you have any concern, please raise them ;)
> 
> I just found that i forget about the SPARC architecture, but we don't have any test for this architecture. If you could craft on executable for testing purpose, it will help me implement this feature faster.

I'm not really familiar with SPARC and I don't have access to such arch, but I think we have some SPARC ELFs in tests. See
```
./elf/analysis/elf-sparc-execstack: ELF 32-bit MSB executable, SPARC32PLUS, V8+ Required, total store ordering, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.2, for GNU/Linux 2.6.18, BuildID[sha1]=27daaa46422285919a67b3e25c3851313a2697db, stripped
./elf/elf-Linux-SparcV8-bash: ELF 32-bit MSB executable, SPARC32PLUS, V8+ Required, total store ordering, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.2, for GNU/Linux 2.6.15, BuildID[sha1]=dd5aab02bdf7b8c649fb24bc0bd5544a79c7aec0, stripped
./elf/elf-solaris-sparc-ls: ELF 32-bit MSB executable, SPARC, version 1 (SYSV), dynamically linked, interpreter /usr/lib/ld.so.1, with debug_info, not stripped
./fuzzed/r2-SEGV-d48-c31-e3c.dms: ELF 64-bit LSB no file type, SPARC, (SYSV)
```

> 
> You talk about *.o files do you have an example? (a test)

Mmm... maybe I was wrong on this.

Anyway, please before adding/fixing anything else, try to remove all those whitespaces/unrelated changes. As mentioned above, you should be able to cherry-pick https://github.com/ret2libc/radare2/commit/53d56cc7f847344280bb6d07e9f3e8e4c1ed25b1 . Also, rebase on top of current master (or merge master into your branch).

08A:
@ret2libc Are you satisfied with the diff size?
For the spark implementation i will try to do that later this week.

ret2libc:
> @ret2libc Are you satisfied with the diff size?
> For the sparc implementation i will try to do that later this week.

Much better now, yes! Thought there are still some changes that don't belong here, probably some wrong merge or old rebase ( see .github files in the diff). Thanks!

libr/bin/format/elf/elf.c
ret2libc: I will have a better look later at this, but I may be wrong :)

08A:
@ret2libc I added support for the sparc arch, but i have not enough knowledge (never reverse sparc exec) to add a test, do you know someone that could verify the result and maybe create a new test?

libr/bin/format/elf/elf.c
```diff
@@ -1420,6 +1420,34 @@ static size_t get_num_relocs_dynamic_plt(ELFOBJ *bin) {
 	return 0;
 }
 
+static ut64 get_import_addr_riscv(ELFOBJ *bin, RBinElfReloc *rel) {
```
ret2libc, line 1423: @08A you do like to make my life harder by moving code around? :D Joking, but on a more serious note, please avoid moving too many stuff when you are doing a big PR, as that really makes it harder to read the diff.

08A: Sorry it was just that the function wasn't in alphabetical order.

radare:
What's left to review/change in here? imho lgtm

libr/bin/format/elf/elf.h
```diff
@@ -49,7 +49,7 @@ typedef struct r_bin_elf_symbol_t {
 typedef struct r_bin_elf_reloc_t {
 	int sym;
 	int type;
-	int is_rela;
+	int rel_mode;
```
radare, line 52: can this be a bool?

test/db/formats/elf/elf-relro
```diff
@@ -131,7 +131,7 @@ nth vaddr      bind   type   lib name
 104 0x00052280 GLOBAL FUNC       sigemptyset
 105 0x00052288 GLOBAL FUNC       ftell
 106 0x00052290 GLOBAL FUNC       feof
-107 0x00443a32 WEAK   NOTYPE     __gmon_start__
+107 0x00000000 WEAK   NOTYPE     __gmon_start__
```
radare, line 134: imho this should be -1

libr/bin/format/elf/elf.c
```diff
@@ -1277,228 +1299,189 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
-			case DT_RELASZ:
-			case DT_PLTRELSZ:
-				return bin->dyn_buf[i].d_un.d_val;
-			default:
-				r_warn_if_reached ();
-				break;
-			}
-		}
-	}
+static size_t get_size_rel_mode(size_t rel_mode) {
```
radare, line 1412: rel_mode is size_t here, in other places is int, make ths consistent

libr/bin/format/elf/elf.c
```diff
@@ -1277,228 +1299,189 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
+	return plt_addr + RISCV_PLT_OFFSET + pos * 0x10;
+}
+
+static ut64 get_import_addr_sparc(ELFOBJ *bin, RBinElfReloc *rel) {
+	if (rel->type != R_SPARC_JMP_SLOT) {
+		bprintf ("Unknown sparc reloc type %d\n", rel->type);
+		return UT64_MAX;
+	}
+	ut64 tmp = get_got_entry (bin, rel);
+	if (tmp == UT64_MAX) {
```
radare, line 1444: just use a single return statemnt like this: `return (tmp==UT64_MAX)? tmp-6:UT64_MAX

libr/bin/format/elf/elf.c
```diff
@@ -1507,177 +1490,71 @@ static ut64 get_import_addr_x86(ELFOBJ *bin, struct ht_rel_t *rel, RBinElfSectio
...
+		  ff253a152000   JMP QWORD [RIP + 0x20153A]
+		  6690		     NOP
+		  ----
+		  ff25ec9f0408   JMP DWORD [reloc.puts_236]
+		  plt_addr + 2 to remove jmp opcode and get the imm reading 4
+		  and if RIP (plt_addr + 6) + imm == rel->offset
+		  return plt_addr, that will be our sym addr
+		  perhaps this hack doesn't work on 32 bits
+		  */
+		int res = r_buf_read_at (bin->b, plt_addr + 2, buf, 4);
```
radare, line 1518: use sizeof instead of the hardcoded 4

08A:
> What's left to review/change in here? imho lgtm

@radare a test for the sparc arch

libr/bin/format/elf/elf.c
```diff
@@ -2759,74 +2626,57 @@ static size_t get_num_relocs_sections(ELFOBJ *bin) {
...
-	}
-	if (info->addr_rel - base_addr <= g_offset && g_offset < info->addr_rel + info->rel_size - base_addr) {
-		return info->addr_rel + info->rel_size - g_offset - base_addr;
-	}
-	if (info->addr_jmprel - base_addr <= g_offset && g_offset < info->addr_jmprel + info->jmprel_size - base_addr) {
-		return info->addr_jmprel + info->jmprel_size - g_offset - base_addr;
+	if (bin->dyn_info.dt_rela - base_addr <= g_offset && g_offset < bin->dyn_info.dt_rela + bin->dyn_info.dt_relasz - base_addr) {
+		diff = bin->dyn_info.dt_rela + bin->dyn_info.dt_relasz - g_offset - base_addr;
+		return diff;
+	} else if (bin->dyn_info.dt_rel - base_addr <= g_offset && g_offset < bin->dyn_info.dt_rel + bin->dyn_info.dt_relsz - base_addr) {
```
radare, line 2668: dont else after return

libr/bin/format/elf/elf.c
```diff
@@ -2759,74 +2626,57 @@ static size_t get_num_relocs_sections(ELFOBJ *bin) {
...
-			return pos;
-		}
 	}
 
 	return pos;
 }
 
-static size_t get_next_not_analysed_offset(size_t section_offset, size_t offset,
-	struct dynamic_relocation_section *info, size_t base_addr) {
+static size_t get_next_not_analysed_offset(ELFOBJ *bin, size_t section_offset,
```
radare, line 2659: function signature should fit 1 line

libr/bin/format/elf/elf.c
```diff
@@ -2842,95 +2692,47 @@ static size_t populate_relocs_record_from_section(ELFOBJ *bin,
...
+			j < bin->g_sections[i].size;
+			j = get_next_not_analysed_offset (bin, bin->g_sections[i].offset, j + size, bin->baddr)) {
 
 			if (!read_reloc (bin, relocs + pos, rel_mode, bin->g_sections[i].offset + j)) {
 				break;
 			}
 
 			fix_rva_and_offset (bin, relocs + pos, i);
-			pos++;
+			++pos;
```
radare, line 2704: dont do preincrements

libr/bin/format/elf/elf.h
radare: use size_t at least to be consistent. but i think this shuold be Elf_(..)

test/db/formats/elf/elf-relro
08A: The problem is that in all tests when get_import_addr can't find any addr the value is set to 0.
It is strange to change this behavior.

libr/bin/format/elf/elf.c
```diff
@@ -1277,228 +1299,185 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
+		return UT64_MAX;
 	}
 
-	if (bin->is_rela == DT_REL) {
-		rel_sec = get_rel_sec (bin, rel_sect);
-		tsize = sizeof (Elf_(Rel));
-	} else if (bin->is_rela == DT_RELA) {
-		rel_sec = get_rel_sec (bin, rela_sect);
-		tsize = sizeof (Elf_(Rela));
+	ut64 pos = (rel->rva - got_addr - 0x3 * WORDSIZE) / WORDSIZE;
```
ret2libc, line 1363: it's not clear what it does. 
For example, why `3 * WORDSIZE`? (I guess there are 3 fields of WORDSIZE size, but...)
Also, why didn't use you store `pos` field in the relocation entry like we had before with the `k` field?

libr/bin/format/elf/elf.c
```diff
@@ -1277,228 +1299,185 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
+	ut64 addr_jmprel = bin->dyn_info.dt_jmprel;
+	ut64 addr_got = bin->dyn_info.dt_mips_pltgot;
+
+	if (!addr_jmprel || !addr_got) {
+		return UT64_MAX;
 	}
 
-	const int htsize = R_MIN (nrel, 1024);
-	rel_cache = ht_up_new_size (htsize, NULL, rel_cache_free, NULL);
+	ut64 pos = (rel->rva - addr_got - 0x2 * WORDSIZE) / WORDSIZE;
```
ret2libc, line 1390: I still think it would be much easier to just store the position from the very beginning, instead of having to computing it every time with "hard" computations.

libr/bin/format/elf/elf.c
```diff
@@ -1277,228 +1299,185 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
 	}
 
-	ret = Elf_(r_bin_elf_get_section_addr) (bin, ".got.plt");
-	if (ret) {
-		return ret;
+	ut64 pos = (rel->rva - got_addr - 0x2 * WORDSIZE) / WORDSIZE;
+	return plt_addr + RISCV_PLT_OFFSET + pos * 0x10;
+}
+
+static ut64 get_import_addr_sparc(ELFOBJ *bin, RBinElfReloc *rel) {
```
ret2libc, line 1438: this looks very different from the existing `get_import_addr_sparc`. Why?

libr/bin/format/elf/elf.c
```diff
@@ -1507,177 +1486,70 @@ static ut64 get_import_addr_x86(ELFOBJ *bin, struct ht_rel_t *rel, RBinElfSectio
...
-		}
-		base = r_mem_mem_aligned (buf, sizeof (buf), (const ut8 *)"\x3c\x0f\x00", 3, 4);
-		if (base) {
-			plt_addr += (int)(size_t) (base - buf);
-		} else {
-			plt_addr += MIPS_PLT_OFFSET + 8; // HARDCODED HACK
-		}
-		plt_addr += rel->k * 16;
-		return plt_addr;
+static ut64 get_import_addr_x86(ELFOBJ *bin, RBinElfReloc *rel) {
```
ret2libc, line 1535: I think you changed too many things... Also, there were some nice comments in the code for x86, like the new layout for endbr64 in plt. Please, try to revert as much as possible to existing code and just do the minimum number of changes to make it work better, without rewriting everything from scratch.

08A:
Okay, thanks for your feedback.

For the position, i choose to not include this information inside the struct, because i didn't find any documentation which explain that the position inside the relocation table is the same that the position inside the got. So we need to recompute every time the position (the cost is not really expensive).

I understand that the majority of the code can be confusing so i will try to explain my choices.

### get_import_addr_arm
In arm elf, the plt_addr is stored inside the got_entry, so i can read a random entry and get the based addr of the plt section.
The position is compute with this formule:
(relocation_rva - got_base_addr - nb_empty_entry * size_got_entry) / size_got_entry
Indeed the entry number 1, 2, 3 are not relevant.

### get_import_addr_mips
In mips the got addr is stored inside a special dynamic entry anmed DT_MIPS_PLTGOT (when there is some relocation)
The position computation is like the arm one (expect that there is only 2 irrelevant entry).

### get_import_addr_riscv
Riscv is like arm (get_got_entry).

### get_import_sparc
We only support R_SPARC_JMP_SLOT. If the relocation is supported the entry inside the got return almost the exact position in the plt of the import_addr we only need some small adjustment (-size of the instruction to jump). It is cleaner than use some position.

### get_import_ppc
Almost the same, pos is from the base addr of the plt (since the rva has a ref inside the plt).

### get_import_x86
less modified, the code is self-explanatory.

### Conclusion
As you can see i tried to be more logic. I understand that your are scared that the new code is broken.
The main problem is not that we have a new code but that you don't trust the test suite. Maybe it is a small sign that the test suite is not exhaustive. I personally don't have all the competence to add tests but if you know someone that is willing to do, let me know.

I will try to reduce again the diff size.

08A:
There is to much dependencies between to many function, it is why the size a the diff is so big.
I don't know how to make it smaller.

08A:
I believe i revert all the esthetic change, now all we have is algorithm modification. imho algorihm that should have been changed long time ago.

libr/bin/format/elf/elf.c
ret2libc: this probably deserves a separate function or a macro with a meaningful name (e.g. `COMPUTE_PLTGOT_POSITION (rel, pltgot_addr, n_initial_unused_entries)`) and you can reuse this in many places to make it clearer.

libr/bin/format/elf/elf.c
```diff
@@ -1277,228 +1299,185 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
+	ut64 plt_addr = get_got_entry (bin, rel);
+	if (plt_addr == UT64_MAX) {
+		return UT64_MAX;
 	}
 
-	ret = Elf_(r_bin_elf_get_section_addr) (bin, ".got.plt");
-	if (ret) {
-		return ret;
+	ut64 pos = (rel->rva - got_addr - 0x2 * WORDSIZE) / WORDSIZE;
+	return plt_addr + RISCV_PLT_OFFSET + pos * 0x10;
```
ret2libc, line 1435: use a define or a const var to hold 0x10, which is AFAIU the size of the plt entry.

libr/bin/format/elf/elf.c
```diff
@@ -1277,228 +1299,185 @@ ut64 Elf_(r_bin_elf_get_section_addr_end)(ELFOBJ *bin, const char *section_name)
...
+
+static ut64 get_import_addr_sparc(ELFOBJ *bin, RBinElfReloc *rel) {
+	if (rel->type != R_SPARC_JMP_SLOT) {
+		bprintf ("Unknown sparc reloc type %d\n", rel->type);
+		return UT64_MAX;
 	}
+	ut64 tmp = get_got_entry (bin, rel);
 
-	return -1;
+	return (tmp == UT64_MAX) ? UT64_MAX : tmp - 0x6;
```
ret2libc, line 1445: what is this `-0x6`? Can we have a macro with a meaningful name also for this? I think it's used in a couple of places at least.

libr/bin/format/elf/elf.c
```diff
@@ -1507,177 +1486,70 @@ static ut64 get_import_addr_x86(ELFOBJ *bin, struct ht_rel_t *rel, RBinElfSectio
...
-		const int sizeOfProcedureLinkageTable = 32;
-		const int sizeOfPltEntry = 16;
-		return plt_section->rva + sizeOfProcedureLinkageTable + (rel->k * sizeOfPltEntry);
+
+	RBinElfSection *pltsec_section = get_section_by_name (bin, ".plt.sec");
+
+	if (pltsec_section) {
+		ut64 got_addr = bin->dyn_info.dt_pltgot;
+		ut64 pos = (rel->rva - got_addr - 0x3 * WORDSIZE) / WORDSIZE;
+		return pltsec_section->rva + pos * 0x10;
```
ret2libc, line 1546: define 0x10 = sizeofPltEntry.

08A:
> Of course I don't trust the test suite :) It's hard to cover everything. I think, for example, we don't have any MIPS bin with relocations.

I found one bin with relocation.
```
readelf --use-dynamic -r test/bins/elf/analysis/mipsbe-ubusd
```

ret2libc:
> > Of course I don't trust the test suite :) It's hard to cover everything. I think, for example, we don't have any MIPS bin with relocations.
> 
> I found one bin with relocation.
> 
> ```
> readelf --use-dynamic -r test/bins/elf/analysis/mipsbe-ubusd
> ```

Nice catch! I did not find it because I was using `rabin2` :) And rabin2, both on master and in your PR, does not find them :( Anyway, as that's not a regression I don't consider it blocking for this PR

08A:
> ****
> 
>     * 2 for Comparison result is always the same

I don't see why lgtm isn't happy? Each comparison are valid.

ret2libc:
> > ```
> > * 2 for Comparison result is always the same
> > ```
> 
> I don't see why lgtm isn't happy? Each comparison are valid.

Oh, tricky one :D 

```C
ut64 base = r_buf_read_ble32_at (bin->b, p_plt_addr, bin->endian);
if (base == UT64_MAX) {
```
`r_buf_read_ble32_at ` returns a ut32 value, not a ut64 one. And in case there are errors, it returns UT32_MAX, not UT64_MAX. So LGTM is correctly complaining about this.

```C
ut64 addr = BREADWORD (bin->b, p_sym_got_addr);
return (!addr || addr == UT64_MAX) ? UT64_MAX : addr;
```
When R_BIN_ELF64 is not defined, BREADWORD will use the r_buf_read_ble32_at, which, again, returns UT32_MAX in case of error and not UT64_MAX.

08A:
@ret2libc It is fixed.

ret2libc:
I will review this later and merge if ok.

PR #16547, opened 2020-04-12
Refactor Variables out of SDB ##anal

thestr4ng3r:
**Detailed description**

Refactor variables to get them out of sdb.

**Closing issues**

Does not fixx #15944 yet

PR #16548, opened 2020-04-12
Remove some useless members from RAnalVar

thestr4ng3r:
This should probably be merged only AFTER the release.

**Detailed description**

`RAnalVar.addr` was always set to the function's addr. No need to keep it there. Especially as a preparation for #16547.

**Test plan**

Check that rebasing of variables still works (Open file, `aaa`, `ood`, check vars)


radare:
ok to merge before 4.5?

thestr4ng3r:
before 4.5 yes, but not before 4.4

radare:
lgtm

PR #16561, opened 2020-04-13
Proper multiplication overflow test

eagleoflqj:

**Detailed description**

Even `a * 4 >= a`, the multiplication may still overflow.
Eg: `a = 0x58000000, a * 4 = 0x60000000`
Instead, we should use division to test whether a*b overflows.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/bin/p/bin_bflt.c
+++ b/libr/bin/p/bin_bflt.c
@@ -117,13 +117,13 @@ static RList *patch_relocs(RBin *b) {
 	return list;
 }
 
-static int get_ngot_entries(struct r_bin_bflt_obj *obj) {
+static ut32 get_ngot_entries(struct r_bin_bflt_obj *obj) {
 	ut32 data_size = obj->hdr->data_end - obj->hdr->data_start;
-	int i = 0, n_got = 0;
+	ut32 i = 0, n_got = 0;
 	if (data_size > obj->size) {
 		return 0;
 	}
-	for (i = 0, n_got = 0; i < data_size; i += 4, n_got++) {
+	for (; i < data_size; i += 4, n_got++) {
 		ut32 entry, offset = obj->hdr->data_start;
 		if (offset + i + sizeof (ut32) > obj->size ||
 		    offset + i + sizeof (ut32) < offset) {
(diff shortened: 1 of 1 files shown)


karliss:
Instead of doing potentially overflowing multiplication which may be undefined behavior depending n argument type and trying to check after the fact. It is possible to do
```c
if (n > INT32_MAX / sizeof(foo)) {
    error;
}
amount = n * sizeof(foo);
```

Since INT32_MAX and  sizeof is compile time constant as additional bonus division gets avoided.

libr/bin/p/bin_bflt.c
```diff
@@ -181,22 +180,17 @@ static RList *relocs(RBinFile *bf) {
...
 		if (!reloc_table) {
 			goto out_error;
 		}
 		amount = n_reloc * sizeof (ut32);
-		if (amount < n_reloc || amount > UT32_MAX) {
-			free (reloc_table);
-			goto out_error;
-		}
-		ut32 *reloc_pointer_table = calloc (1, amount + 1);
+		ut32 *reloc_pointer_table = calloc (1, amount);
```
radare, line 193: amount can be 0, so this will be UB

libr/bin/p/bin_bflt.c
```diff
@@ -144,20 +144,19 @@ static int get_ngot_entries(struct r_bin_bflt_obj *obj) {
...
 		if (n_got) {
-			amount = n_got * sizeof (ut32);
-			if (amount < n_got || amount > UT32_MAX) {
+			if (n_got > UT32_MAX / sizeof (struct reloc_struct_t)) {
 				goto out_error;
 			}
-			struct reloc_struct_t *got_table = calloc (
-				1, n_got * sizeof (struct reloc_struct_t));
+			amount = n_got * sizeof (struct reloc_struct_t);
+			struct reloc_struct_t *got_table = calloc (1, amount);
```
radare, line 159: same

radare: just noticed this after the 2 aprovals. can you have a look at it?

libr/bin/p/bin_bflt.c
eagleoflqj:
```c
if (obj->hdr->reloc_count > 0) {
		ut32 n_reloc = obj->hdr->reloc_count;
		if (n_reloc > UT32_MAX / sizeof (struct reloc_struct_t)) {
			goto out_error;
		}
		amount = n_reloc * sizeof (struct reloc_struct_t);
```
Here n_reloc>=1, so amount can't be zero.

libr/bin/p/bin_bflt.c
eagleoflqj: `if(ngot)`
guaranteed ngot>=1 and amount >=1

PR #16565, opened 2020-04-13
Fix off by one when checking for "zzz"

disconnect3d:
The string literal "zzz" has a length of 3 so the `strcnmp` used should have size argument of 3 instead of 2.


disconnect3d:
> can u add it and confirm the fix works
Unfortunately, no. I found this issue in an automated way and I do not have more time to contribute more to the project, sorry.

radare:
we have 1 test for izzzj so i assume the patch is correct :) thanks for

radare:
the fix

PR #16567, opened 2020-04-13
Fix off by one when checkinf for .rdata section

disconnect3d:
The string literal `".rdata"` has a length of 6 so the `strncmp` used should have size argument of 6 instead of 5.

### DIFF
--- a/libr/bin/format/mdmp/mdmp_pe.c
+++ b/libr/bin/format/mdmp/mdmp_pe.c
@@ -198,7 +198,7 @@ RList *PE_(r_bin_mdmp_pe_get_sections) (struct PE_(r_bin_mdmp_pe_bin) * pe_bin) 
 		if ((ptr->perm & R_PERM_R) && !(ptr->perm & R_PERM_X) && ptr->size > 0) {
 			if (!strncmp (ptr->name, ".rsrc", 5) ||
 				!strncmp (ptr->name, ".data", 5) ||
-				!strncmp (ptr->name, ".rdata", 5)) {
+				!strncmp (ptr->name, ".rdata", 6)) {
 				ptr->is_data = true;
 			}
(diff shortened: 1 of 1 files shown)


libr/bin/format/mdmp/mdmp_pe.c
```diff
@@ -198,7 +198,7 @@ RList *PE_(r_bin_mdmp_pe_get_sections) (struct PE_(r_bin_mdmp_pe_bin) * pe_bin)
 		if ((ptr->perm & R_PERM_R) && !(ptr->perm & R_PERM_X) && ptr->size > 0) {
 			if (!strncmp (ptr->name, ".rsrc", 5) ||
 				!strncmp (ptr->name, ".data", 5) ||
-				!strncmp (ptr->name, ".rdata", 5)) {
+				!strncmp (ptr->name, ".rdata", 6)) {
```
radare, line 201: use strcmp if you want to check the whole string instead of strncmp

disconnect3d: The `strcmp` call will also check a null byte and here, I think we don't want to check for the null byte and if we used `strcmp` it will not match if the first string contains more than just the `.rdata` string:
```
In [5]: import ctypes

In [6]: libc = ctypes.CDLL('libc.so.6')

In [7]: libc.strcmp(".rdata some bytes further", ".rdata")
Out[7]: 32
```

disconnect3d: Depending on the context this is used this could be further bugged and accept e.g. `.rodataaaaa` sections. However, if this is the case here, I would say it is out of scope of the reported issue since there are other cases that would have this case too. And the reported issue is just about the off-by-one size argument.

PR #16569, opened 2020-04-13
Fix off size argument in strncmp call in http server

disconnect3d:
The `"------------------------------"` string literal has a length of 30, so it seems the `strncmp` should use the size argument of 30 instead of 10.


libr/socket/socket_http_server.c
```diff
@@ -140,7 +140,7 @@ R_API ut8 *r_socket_http_handle_upload(const ut8 *str, int len, int *retlen) {
 	if (retlen) {
 		*retlen = 0;
 	}
-	if (!strncmp ((const char *)str, "------------------------------", 10)) {
+	if (!strncmp ((const char *)str, "------------------------------", 30)) {
```
radare, line 143: not sure whats the ponit of this strncmp here.. @thestr4ng3r ?

thestr4ng3r: same

ret2libc: What to do with this? It seems it was introduced by @radare at https://github.com/radareorg/radare2/commit/3fb4c7dd5e85827908e5ba146464a5442d8c5f21#diff-2cb60c536b11b668639979ef9886e233R76 , but the purpose is not clear. Why would `str` contain `----------------------`? Is this function even working?

radare: well its to handle the mime separators, and 10 chars is fine, maybe i was just lazy to count how many dashes, so to merge this safely i would go for chopping the string rather than increasing the length number

ret2libc: Yeah, that's probably better as at https://github.com/radareorg/radare2/pull/16569/files#diff-9aac2ba872feec932a128e264b9855b9R146 it does `str + 10`

ret2libc:
@radare please review and let's just merge this. I've shortened the `----` string as we agreed on. Thanks @disconnect3d ! I hope you don't mind me adding a couple of commits to close this (I'm trying to clean a bit existing PRs).

PR #16585, opened 2020-04-14
Fix some memory leaks.

karliss:

**Detailed description**

Some memory leaks found by address sanitizers.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -3108,11 +3108,16 @@ static void cmd_print_pv(RCore *core, const char *input, bool useBytes) {
 	}
 }
 
-static int cmd_print_blocks(RCore *core, const char *input) {
+static bool cmd_print_blocks(RCore *core, const char *input) {
+	bool result = false;
 	char mode = input[0];
+	RList *list = NULL;
+	RCoreAnalStats *as = NULL;
+	RTable *t = NULL;
+	PJ *pj = NULL;
 	if (mode == '?') {
 		r_core_cmd_help (core, help_msg_p_minus);
-		return 0;
+		return false;
 	}
 
 	if (mode && mode != ' ') {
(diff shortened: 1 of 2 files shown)


libr/core/cmd_print.c
```diff
@@ -3273,26 +3283,30 @@ static int cmd_print_blocks(RCore *core, const char *input) {
...
-			r_cons_printf ("\n%s\n", r_table_tofancystring (t));
-			r_table_free (t);
-			break;
+	case 'j':
+		pj_end (pj);
+		pj_end (pj);
+		r_cons_println (pj_string (pj));
+		break;
+	case 'h': {
+		r_cons_printf ("\n%s\n", r_table_tofancystring (t));
```
radare, line 3292: Memory leak

libr/core/cmd_print.c
```diff
@@ -3132,11 +3137,11 @@ static int cmd_print_blocks(RCore *core, const char *input) {
 	ut64 off = core->offset;
 	ut64 from = UT64_MAX;
 	ut64 to = 0;
-	RTable *t = r_core_table (core);
-	t->showSum = true;
-	RList *list = r_core_get_boundaries_prot (core, -1, NULL, "search");
+
+	list = r_core_get_boundaries_prot (core, -1, NULL, "search");
 	if (!list) {
-		return 1;
```
radare, line 3139: Is result having more values than 0 and 1? Because maybe we can boolify it

karliss: Doesn't seem like exact value is used anywhere. In the success case it returns len - number of blocks. But len gets calculated only for json which is the case where count is least useful so most likely the result doesn't get properly used at all and it can be boolified.

libr/core/cmd_print.c
karliss: For what?

karliss: The shared cleanup should deal with json.

libr/core/cmd_print.c
radare: imho in case of confusion, better go for bool

libr/core/cmd_print.c
radare: tofancystring() returns a string in the heap. you must free it

libr/core/cmd_print.c
karliss: Done.

libr/core/cmd_print.c
karliss: Fixed.

wargio:
@karliss you broke something https://builds.sr.ht/~xvilka/job/187591#task-test-506

karliss:
Doing rebase fixed at least one of them locally, let's see if it will help in CI as welll.

libr/core/cmd_print.c
```diff
@@ -3273,26 +3283,32 @@ static int cmd_print_blocks(RCore *core, const char *input) {
...
-			r_cons_printf ("\n%s\n", r_table_tofancystring (t));
-			r_table_free (t);
-			break;
+	case 'j':
+		pj_end (pj);
+		pj_end (pj);
+		r_cons_println (pj_string (pj));
+		break;
+	case 'h': {
+		char *table_string = r_table_tofancystring (t);
```
radare, line 3292: maybe worth checking for null

karliss: Makes sense, tofancystring sounds like nontrivial operation which might fail. It doesn't seem to have much error checking right now but might in future.

Issue #6350, opened 2016-12-20
pf enhacements


XVilka:
- [ ] Support `pf.` nested struct/union name autocompletion (currently works only for the 1st level)
 - [ ] Add custom terminator support (for strings, etc)
 - [ ] Improve (and sort, may be group) help for the command

Maijin:
```
 Improve (and sort, may be group) help for the command
 Better API to load/store pf. formats (may be parse?) from bindings (and r2pipe)
```
Could you details those ? The first two are dup i've pointed to existing open issues

XVilka:
About improving help for the command - group the similar things, like byte-dword, chars-strings, etc.
About API - not sure how it should look exactly, but something more useful.

XVilka:
@oxcabe can you please help also to improve the `pf??` and `pf???` output, so it will be grouped by contexts (it will help you to understand the `pf` commands as well).
See also [Print Modes](https://radare.gitbooks.io/radare2book/content/basic_commands/print_modes.html) chapter of the book, so that it should be synchronized with the new `pf` output.

gogo2464:
Can I be assigned to this issue please?

Can you give me the syntax of the command related to https://github.com/radareorg/radare2/issues/414 please?

radare:
you are now in charge of it :)

gogo2464:
Thank you @radare.

gogo2464:
@radare do I use the command `pfe`? I read your post https://github.com/radareorg/radare2/issues/414.

ret2libc:
This issue has been moved from radareorg/radare2 to radareorg/ideas as we are trying to clean our backlog and this issue has probably been created a long while ago. This is an effort to help contributors understand what are the actionable items they can work on, prioritize issues better and help users find active/duplicated issues more easily. If this is not an enhancement/improvement/general idea but a bug, feel free to ask for re-transfer to main repo. Thanks for your understanding and contribution with this issue.

PR #16586, opened 2020-04-14
newshell - Add support for `pf` commands

ret2libc:
Add support for `pf` commands in newshell.
This is going to be a mess, unfortunately... `pf` commands have almost any characters inside. My 2cents is that they need to be rethought a bit, once this is done. For now I am trying to consider the arguments of `pf` as one single string, though that probably will require changes in tests.

For example, things like `(foo; pf (3xx))` will probably need to become `(foo; "pf (3xx)")` to make sure to distinguish the ending parenthesis `)` correctly.


XVilka:
In theory, `pf` commands should be parsed by the new parser as well - it will help to solve the issues like nested `pf.xxx.yyy.zzz` autocompletion: https://github.com/radareorg/radare2/issues/6350#issue-196599016

@radare what do you think about changing `pf` syntax a bit while we're on it?

ret2libc:
@XVilka do we have a full "specification" of the pf syntax? I can work on it.

XVilka:
Not really, but most of the syntax should be covered in `pf?` and `pf??`

ret2libc:
master: `71 XX`, this PR: `31 XX` yeah!

ret2libc:
> master: `71 XX`, this PR: `31 XX` yeah!

Actually, PR: `24 XX`!

libr/core/cmd.c
```diff
@@ -4889,7 +4908,7 @@ static bool substitute_args_do(struct tsr2cmd_state *state, RList *edits, TSNode
 static bool substitute_args(struct tsr2cmd_state *state, TSNode args, TSNode *new_command) {
 	RList *edits = r_list_newf ((RListFree)free_tsr2cmd_edit);
 
-	if (is_ts_args (args) || is_ts_arg (args)) {
+	if (is_ts_args (args) || is_ts_arg (args) || is_ts_pf_dot_cmd_args (args) || is_ts_pf_new_args (args) || is_ts_pf_args (args)) {
```
XVilka, line 4911: Please split the long line

PR #16588, opened 2020-04-14
Improve newshell.yaml and use better names for GH builds

ret2libc:
Unfortunately GH action does not directly support filtering of PRs by branch name (the name you specify in `on.branches` is the "base" branch. With this PR, newshell is always built on every PR, but only those that are named `newshell-*` run the full tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,23 +12,23 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


ret2libc:
@XVilka can you merge this and change the required set of actions on PR? I've changed the names, as you can see. It should be clearer now what each build does (whether it runs tests or not).

XVilka:
Which ones should be required now?

ret2libc:
- linux-acr-gcc-tests 
- linux-acr-clang-build
- linux-meson-gcc-build
- macos-clang-tests

PR #16601, opened 2020-04-15
Fix #15598 - Remove cherrypull.sh

### LINKED ISSUE #15598: Merge all the *.sh in the root directory into one
* r2r is the name of the r2r.js (npm i -g node-r2r)
* cherrypull.sh must die
* overlay.sh i guess its used by travis, so may be worth
* r2-v is useful, but 
* we can probably replace those with just makefiles

comments, feedback, ideas?

thestr4ng3r:
Fix #15598


radare:
Use “Fix #number - issue description ##tagifneeded”

Issue #16600, opened 2020-04-15
x @@@C:cmd is wrong


ret2libc:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Fedora 31 - x86_64
| File format of the file you reverse (mandatory)      | any
| Architecture/bits of the file (mandatory)            | x64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.4.0 27759 @ linux-x86-64 git.4.4.0-4-gf6528e171
commit: f6528e171e9caaef4d19e0a4bb506cd196e7d417 build: 2020-04-15__09:41:36

### Issue
The help message of `@@@C` says:
```
| x @@@C:cmd   comments matching
```
I think this is wrong, as `@@@C` probably does not need any `cmd` at the end. This is probably a wrong copy-paste from `@@@c:`. 

### Steps to reproduce
```
$ radare2 -
> Csa 10 @0xdeadbeef
> C
0xdeadbeef latin1[10] "\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff"
> pd 1 @@@C
```

### Actual behaviour
Nothing

### Expected behaviour
```
            0xdeadbeef     .string "\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff" ; len=10
```

ret2libc:
@thestr4ng3r this is fixed, isn't it? AFAIR you fixed it not long ago. If not, please re-open.

thestr4ng3r:
Yes, fixed in edf1be10d114e98b95aa1f6ddf7bbebf5279aa7e

PR #16607, opened 2020-04-16
Add support for foreach commands ##newshell

ret2libc:
Adds support for all `@@@?` commands in the newshell.
This is mostly copy-paste code from existing `@@@?` code.


ret2libc:
By the way, this also fixes the issue in https://github.com/radareorg/radare2/issues/16600 . It does not fix the help msg though.

libr/core/cmd.c
```diff
@@ -4636,6 +4636,36 @@ DEFINE_IS_TS_FCN(double_quoted_arg)
...
+		if (!ptr || !*ptr) {
+			return 1;
+		}
+		ptr = strchr (ptr + 1, ',');
+		if (!ptr || !*ptr) {
+			return 1;
+		}
+		char *cmt = (char *)sdb_decode (ptr + 1, 0);
+		if (cmt) {
+			r_core_cmdf (core, "s %s", k + 7);
```
radare, line 4661: use the API

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
 	return res;
 }
 
+DEFINE_HANDLE_TS_FCN(foreach_addrsize_command) {
+	RCore *core = state->core;
+	TSNode command = ts_node_named_child (node, 0);
+	uint32_t i = 1;
```
radare, line 5989: use size_t

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+	ut64 obs = core->blocksize;
+	if (!fcn) {
+		return false;
+	}
+
+	RListIter *iter;
+	RAnalBlock *bb;
+	bool ret = true;
+	r_list_foreach (fcn->bbs, iter, bb) {
+		r_core_seek (core, bb->addr, 1);
```
radare, line 6024: use true instead of 1, the arg expects bool

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+	return ret;
+}
+
+DEFINE_HANDLE_TS_FCN(foreach_cmd_command) {
+	// convert @@@c: command into a @@@= one, by using the output of the
+	// in_cmd as addr/blksz of @@@=
+	TSNode in_cmd = ts_node_named_child (node, 1);
+	substitute_args_init (state, node);
+
+	RList *edits = r_list_newf ((RListFree)free_tsr2cmd_edit);
```
radare, line 6039: check for null

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+
+	char *in_cmd_out = do_handle_substitution_cmd (state, in_cmd);
+	in_cmd_out = escape_special_chars (in_cmd_out, SPECIAL_CHARS_REGULAR);
+	struct tsr2cmd_edit *e = create_cmd_edit (state, in_cmd, in_cmd_out);
+	r_list_append (edits, e);
+
+	TSNode op = ts_node_child (node, 1);
+	e = create_cmd_edit (state, op, strdup ("@@@="));
```
radare, line 6047: passing strdup? isnt it leaking imho it should take a const char *

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+	r_list_append (edits, e);
+
+	TSNode new_command;
+	bool ok = substitute_args_do (state, edits, &new_command);
```
radare, line 6051: unnecessary temporal constant variable which is read only once

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+	r_meta_list_cb (state->core->anal, R_META_TYPE_COMMENT, 0, foreach_comment_newshell, (void *)&cmt_t, UT64_MAX);
+	return true;
+}
+
+DEFINE_HANDLE_TS_FCN(foreach_import_command) {
+	RCore *core = state->core;
+	TSNode command = ts_node_named_child (node, 0);
+	RBinImport *imp;
+	ut64 offorig = core->offset;
+	RList *list = r_bin_get_imports (core->bin);
```
radare, line 6078: can return null , better to do an early return in that case before allocating

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+	RList *lost = r_list_newf (free);
+	RListIter *iter;
+	r_list_foreach (list, iter, imp) {
+		char *impflag = r_str_newf ("sym.imp.%s", imp->name);
+		ut64 addr = r_num_math (core->num, impflag);
+		ut64 *n = R_NEW (ut64);
```
radare, line 6084: use ut64_new()

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+		*n = addr;
+		r_list_append (lost, n);
+		free (impflag);
+	}
+	ut64 *naddr;
+	bool res = true;
+	r_list_foreach (lost, iter, naddr) {
+		ut64 addr = *naddr;
+		if (addr && addr != UT64_MAX) {
+			r_core_seek (core, addr, 1);
```
radare, line 6094: use true instead of 1

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+
+DEFINE_HANDLE_TS_FCN(foreach_import_command) {
+	RCore *core = state->core;
+	TSNode command = ts_node_named_child (node, 0);
+	RBinImport *imp;
+	ut64 offorig = core->offset;
+	RList *list = r_bin_get_imports (core->bin);
+	RList *lost = r_list_newf (free);
+	RListIter *iter;
+	r_list_foreach (list, iter, imp) {
```
radare, line 6081: maybe better to just use RPVector or RVector

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+			}
+			if (item->type != i) {
+				continue;
+			}
+			r_list_append (list, strdup (item->name));
+		}
+		const char *item_name;
+		r_list_foreach (list, iter, item_name) {
+			value = r_reg_getv (core->dbg->reg, item_name);
+			r_core_seek (core, value, 1);
```
radare, line 6130: use true instead of 1

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+	if (!obj) {
+		return false;
+	}
+	bool res = true;
+	ut64 offorig = core->offset;
+	ut64 bszorig = core->blocksize;
+	RBinSection *sec;
+	RListIter *iter;
+	r_list_foreach (obj->sections, iter, sec) {
+		r_core_seek (core, sec->vaddr, 1);
```
radare, line 6210: true vs 1

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+DEFINE_HANDLE_TS_FCN(foreach_dbgmap_command) {
+	RCore *core = state->core;
+	TSNode command = ts_node_named_child (node, 0);
+	RDebug *dbg = core->dbg;
+	bool res = true;
+	if (dbg && dbg->h && dbg->maps) {
+		RDebugMap *map;
+		RListIter *iter;
+		r_list_foreach (dbg->maps, iter, map) {
+			r_core_seek (core, map->addr, 1);
```
radare, line 6248: true vs 1

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+	r_flag_foreach_glob (core->flags, glob, copy_into_flagitem_list, flags);
+	RListIter *iter;
+	RFlagItem *f;
+	bool res = true;
+	r_list_foreach (flags, iter, f) {
+		r_core_block_size (core, f->size);
+		r_core_seek (core, f->offset, 1);
+		res &= handle_ts_command (state, command);
+	}
+	r_core_seek (core, off, 0);
```
radare, line 6275: imho we should kill this 3rd argument to RCore.seek, but well.

radare: but can be a bit tricky to use RIO curseek instead of RCore seek

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+	if (dbg && dbg->h && dbg->h->threads) {
+		int origpid = dbg->pid;
+		RDebugPid *p;
+		RList *list = dbg->h->threads (dbg, dbg->pid);
+		if (!list) {
+			return false;
+		}
+		RListIter *iter;
+		r_list_foreach (list, iter, p) {
+			r_core_cmdf (core, "dp %d", p->pid);
```
radare, line 6326: use the api

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+			r_cons_printf ("PID %d\n", p->pid);
```
radare, line 6327: unnecessary printf

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+			res &= handle_ts_command (state, command);
+		}
+		r_core_cmdf (core, "dp %d", origpid);
```
radare, line 6330: use the api

libr/core/cmd.c
ret2libc: tree-sitter uses `uint32_t` so I thought I should use it as well to avoid problems. `size_t` may be ut64 on some arch, while uint32_t is always 32bit.

libr/core/cmd.c
ret2libc: good point... however I would feel a bit inconsistent in doing this here, as there are like ~250 uses of this function and probably ~200 use 0/1. Can I do a separate PR where I change ALL of them at once?

libr/core/cmd.c
ret2libc: It does not as `e` is then freed by `free_tsr2cmd_edit`, which expects the string to be free-able.

libr/core/cmd.c
ret2libc: that function is only defined/used in cmd_debug.c. I don't feel like it's that much used to make it generic at this point though.

libr/core/cmd.c
ret2libc: sure, but... that's completely out of scope for this PR, which simply reuses existing code to implement existing commands in the newparser

libr/core/cmd.c
ret2libc: mmm.. I'm just mimicking existing commands and mostly copy-pasted existing code. It was already doing this printf and I copied it.

libr/core/cmd.c
```diff
@@ -5953,6 +5983,356 @@ DEFINE_HANDLE_TS_FCN(iter_hit_command) {
...
+			res &= handle_ts_command (state, command);
+		}
+	}
+	r_cons_break_pop ();
+	r_core_block_size (core, obs);
+	r_core_seek (core, offorig, 1);
+	return res;
+}
+
+DEFINE_HANDLE_TS_FCN(foreach_thread_command) {
```
ret2libc, line 6312: @radare I fail to see how this was working at all...
First, a semantic question: what does it mean to execute a command "foreach thread"?
Second: `dp %d` just executes `r_debug_pid_list`, which prints a bunch of PIDs on the screen, but does not really change anything (no seek, no block size, no nothing), so executing a command after this is just like executing a command X times (where X is the number of threads in the system).

radare:
I know this and we ve discussed this already. Iterating with ut64 on 32 bit machines is a waste of time. Thats why loops should be using sizt for counters

radare:
As long as its new code i dont see why not change it here

radare:
I dont ser anything wrong in fixing. Opypasted code

radare:
Dbt @@@t will show the bscktrace of each thread. Which is selected on each itsration with dpt

radare:
The threads are the ones in the child proxess u debug

ret2libc:
> Dbt @@@t will show the bscktrace of each thread. Which is selected on each itsration with dpt
> The threads are the ones in the child proxess u debug

I fail to see how it worked, as this is implemented in https://github.com/radareorg/radare2/blob/master/libr/core/cmd.c#L3923 and it uses `dp <pid>` (which just "List children of pid") and not `dp= <pid>`(which "Select pid").

ret2libc:
> I know this and we ve discussed this already. Iterating with ut64 on 32 bit machines is a waste of time. Thats why loops should be using sizt for counters

Indeed using a `size_t` when `uint32_t` is enough is a waste of time, why are you suggesting it? I'm using an external API that already uses `uint32_t`, I don't see why I should use `size_t` and then cast it to make tree-sitter accept it, when I can just directly use `uint32_t`.

radare:
not really. 64bt machines work better with 64bit registers

radare:
so wha'ts wht you dont get from @@@t ?

ret2libc:
```
0x7f96ab0af33d]> pd 1 @@@t
 * 59084 ppid:59082 uid:102913 s /usr/local/bin/r2r
 - 59105 ppid:59084 uid:102913 S jq
PID 59084
            ;-- rip:
            0x7f96ab0af33d      c3             ret
PID 59104
            ;-- rip:
            0x7f96ab0af33d      c3             ret
 * 59084 ppid:59082 uid:102913 s /usr/local/bin/r2r
 - 59105 ppid:59084 uid:102913 S jq
```
As you can see, `pd 1` is executed two times on the same address, in the same process. The process is not changed.

radare:
then dp= is not executed? well i think the only problem here is missing ".dr*;sr PC" to reload the flags

XVilka:
Also related - it would be nice to have the textual command for `foreach` as well, in the future.

libr/core/cmd.c
XVilka: Completely agree 👍

libr/core/cmd.c
```diff
@@ -4636,6 +4636,36 @@ DEFINE_IS_TS_FCN(double_quoted_arg)
...
+
+static int foreach_comment_newshell(void *user, const char *k, const char *v) {
+	RAnalMetaUserItem *ui = user;
+	RCore *core = ui->anal->user;
+	struct foreach_comment_newshell_t *cmt_t = ui->user;
+	TSNode *cmd = cmt_t->command;
+	struct tsr2cmd_state *state = cmt_t->state;
+	if (!strncmp (k, "meta.C.", 7)) {
+		char *ptr = strchr (v, ',');
+		if (!ptr || !*ptr) {
```
XVilka, line 4652: Don't we have a macro for this pattern? It's quite common.

libr/core/cmd.c
XVilka: +1 for the separate PR.

libr/core/cmd.c
XVilka: For each thread of the debugging process. It's helpful to go through threads for diplaying the information or setting a breakpoints.

ret2libc: Yes.. my point is that this has never worked :D

ret2libc: I have fixed this in the newshell implementation to really work on all threads.

ret2libc:
@XVilka @radare I have addressed the comments apart from:
- replacing `r_core_seek (...., 0/1)` with `r_core_seek (...., false/true)`. I'm going to do that in a separate PR, it's a completely separated problem from the purpose of this PR and it's useless to just fix this in 3 places and keep all the other 250 uses the same. I'm going to fix this once and for all separately.
- using size_t instead of uint32_t. I see no reason for doing this. The compiler will complain because the `i` is passed to a function that accepts `uint32_t`. Why shall we cast things around, when we can just use the right type?

Apart from that, I think the PR is good.

PR #16608, opened 2020-04-16
WIP: Initial PoC of the new r_arch API

radare:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


libr/asm/asm.c
```diff
@@ -327,6 +323,13 @@ R_API bool r_asm_use(RAsm *a, const char *name) {
 	if (!a || !name) {
 		return false;
 	}
+#if 1
```
ret2libc, line 326: remove

libr/asm/asm.c
```diff
@@ -327,6 +323,13 @@ R_API bool r_asm_use(RAsm *a, const char *name) {
...
+	if (r_arch_use (a->arch, name)) {
+		// pull info 
+eprintf ("AUSING ARCH\n");
```
ret2libc, line 329: remove

libr/arch/p/arch_bf.c
```diff
@@ -134,24 +202,26 @@ static int assemble(RAsm *a, RAsmOp *op, const char *buf) {
...
 	.version = "4.0.0",
 	.arch = "bf",
 	.license = "LGPL3",
 	.bits = 16 | 32 | 64,
 	.endian = R_SYS_ENDIAN_NONE,
 	.desc = "Brainfuck",
-	.disassemble = &disassemble,
-	.assemble = &assemble
+	.decode = decode,
+//	.encode = encode
```
ret2libc, line 217: remove

libr/arch/p/arch_bf.c
```diff
@@ -53,12 +56,77 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
...
+	r_strbuf_setbin (&ins->data, buf, len);
+	ins->size = len;
+}
+
+R_API RArchInstruction *r_arch_instruction_new() {
+	RArchInstruction *ins = R_NEW0 (RArchInstruction);
+	return ins;
+}
+
+#if 0
```
ret2libc, line 127: remove all the code that is #if 0ed.

libr/arch/p/arch_bf.c
```diff
@@ -53,12 +56,77 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
...
+
+static bool encode(RArch *a, RArchInstruction *ins, RArchDecodeOptions opt) {
+	bool ret = false;
+	if (opt & R_ARCH_OPTION_DISASM) {
+		ret = disassemble (a, ins);
+	}
+	return ret;
+}
+
+#if 0
```
ret2libc, line 104: remove

libr/arch/p/arch_bf.c
```diff
@@ -53,12 +56,77 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
...
-	r_strbuf_set (&op->buf_asm, buf_asm);
-	op->size = rep;
-	return rep;
+	r_strbuf_set (&ins->code, buf_asm);
+	ins->size = rep;
+	return true;
 }
 
-static int assemble(RAsm *a, RAsmOp *op, const char *buf) {
+#if 0
```
ret2libc, line 64: remove this, add a unit test instead.

libr/include/r_arch.h
```diff
@@ -0,0 +1,87 @@
...
+
+typedef struct r_arch_t {
+	struct r_arch_plugin_t *cur;
+	RArchEndian endian;
+	RArchBits bits;
+	char *cpu;
+	RList *plugins;
+	RCoreBind corebind;
+	size_t syntax;
+} RArch;
```
thestr4ng3r, line 28: Having plugins and cur+bits+endian+... in the same struct means you can only use one configuration at a time and have to switch around if you want another one. I think it would be better to have one `RArch` managing all the plugins and stuff and you can create multiple `RArchInstance`s (or whatever we call it) from it that will do the actual work.

libr/arch/p/arch_bf.c
radare: no. this needs to be implemented, this is why its WIP

libr/include/r_arch.h
radare: changing bits is not that expensive imho but i agree we should discuss about this configuration, my plan was to move arch/cpu/bits from asm/anal into arch, so there will be only one common place for this, and after this we can see if we want to remove bits from the configuration, this is only affecting ARM, and its only affecting ARM because we are misusing bits when we should be naming it `mode`.

libr/arch/p/arch_bf.c
radare: yes. will do unit test

libr/include/r_arch.h
thestr4ng3r: Yeah if RArch is considered the r2 module and these are the points for the asm.arch/... vars then it makes sense somehow. However if you think of RArch as a generic disassembly library (which I think it could be extremely useful as) then it make absolutely no sense to couple the plugin list with a single arch configuration.
So in that sense maybe we should have an architecture like this:
* RArch has all the plugins, maybe initializes them, etc.
* RArchInstance is created from RArch, contains exactly one immutable configuration
* some module (can be RAnal maybe) contains the vars, manages RArch and one RArchInstance, can create more, temporary, RArchInstances on demand.
So in the end RArch really is a 100% self-contained lib that maybe even only depends on RUtil.

test/unit/test_arch.c
```diff
@@ -0,0 +1,43 @@
...
+	RArch *a = r_arch_new ();
+	mu_assert_null (a, "r_arch_new");
+
+	res = r_arch_use (a, "bf");
+	mu_assert ("r_arch_use", res);
+
+	r_arch_instruction_init (&ins);
+
+	r_strbuf_setf (&ins.code, "nop");
+	res = r_arch_encode (a, &ins, R_ARCH_OPTION_DISASM);
```
ret2libc, line 17: maybe call the option `R_ARCH_OPTION_ASM_DISASM`? When reading this I was a bit confused by the fact that you are using `R_ARCH_OPTION_DISASM` in both `r_arch_encode` and `r_arch_decode`. Now I see that `R_ARCH_OPTION_DISASM` is used both for assembler and disassembler, but it may be confusing IMHO.

libr/include/r_arch.h
```diff
@@ -0,0 +1,95 @@
...
+
+#include <r_util.h>
+#include <r_bind.h>
+
+typedef int RArchBits;
+typedef int RArchEndian;
+
+typedef enum r_arch_decode_options_t {
+	R_ARCH_OPTION_DISASM = 1,
+	R_ARCH_OPTION_SIZE = 2,
```
ret2libc, line 14: can you provide examples of these as well in tests please? I'm asking because I'm not sure dealing with options inside each arch plugin is the best approach. For example, to get the size of an instruction you probably need to disassemble it anyway, so maybe it can be done once and for all in `r_arch_encode`/`r_arch_decode` instead of per-plugin. Similar for the other values in this enum probably, but some use cases as tests would help.

libr/include/r_arch.h
ret2libc: On the other hand, I think capstone does more or less the same, so it kinda works either way I guess.

radare:
Thats how it works tight now

radare:
Maybe CODE and DATA instrad?

libr/include/r_arch.h
ret2libc: Right, sorry! I was misreading the capstone doc :/ yeah it uses indeed a handle, so maybe better to do as @thestr4ng3r said.

radare:
why's the build for meson failing? ive dded the r_arch_dep

libr/anal/anal.c
```diff
@@ -213,12 +213,27 @@ R_API bool r_anal_use(RAnal *anal, const char *name) {
 }
 
 R_API char *r_anal_get_reg_profile(RAnal *anal) {
-	return (anal && anal->cur && anal->cur->get_reg_profile)
+	r_return_val_if_fail (anal, NULL);
+	if (anal->arch && anal->arch->cur && anal->arch->cur->registers) {
```
XVilka, line 217: This part might become a macro too, since will be the same across many r2 plugins

radare:
Bool r_arch_can_encode()

radare:
The PR should be green now. so I would appreciate a deeper look on the new API, adaptations and cleanups main discussion points are:

* Instruction descriptions shuold be decoded in the instruction instead of a separate api? (move asm/d into arch/d)
* arch/bits/endian should be in RArch (at least for now, otherwise the transition from asm/anal will be a pain)
* We need to map RAnalOpMask and RArchOption, can we get rid of one of them? imho the anal ones should remain because it handles HINT and other stuff that is not going to be moved into Arch
* Use RVector instead of jump/fail + RList of destinations? (thats how i did it now)
* Should we move all the R_ANAL_OP_TYPE into R_ARCH_INSTRUCTION_TYPE_* ?
* Find a better(shorter) name for RArch.Instruction (maybe RArch.Instr) ?
* Should it depend on RIOBind? it is needed right now.. because we need to read further sometimes to know the instruction meaning.. ive also encountered the situation that maybe we need to hold/access some metadata (?.HINT), for example the LOOP instruction in brainfuck, it will jump to the previously stored WHILE address
* It shuold communicate with RBin (dex, java.. needs this because it needs to resolve offsets and names, which are not actual flags)
* Use getregprofile or setregprofile? (imho getregprofile is better, but then we need to return in the heap and we have no way to check if it has changed or its still the same)
* use bits for thumb/arm mode changes
* How many other ways we want to use .encode()? imho it will be good to support instruction modification to change ins->dest[0] = 0x4000; then call .encode(.ANAL) to encode that new jump destination into the given insturction. this way we can finally reimplement the `write-hack` commands into a meaningful api.
* Endian is no longer bi (to support middle endian)
* RStrBuf doesn't support holding a weak pointer to an already allocated or unowned string or byte data array. this is needed to avoid unnecessary memory copies.
* Do we want to keep using the OPEX json string to export unstructured data?
* Do we want to provide a low level interface to expose the internal disasssembler information like capstone struct or whatever lib its under there

I would like to solve all those questions before merging the PR, porting other plugins in the future will expose other questions

XVilka:
> use bits for thumb/arm mode changes

I think the mode should be decoupled from bits.After all, even Thumb can have non-16bit instructions and registers. It might lead to a deeper bugs, when something relies on the bit size, while technically bitness is different from the real one.

libr/include/r_arch.h
radare: i did one unit test, will add more over time

radare:
technically thumb its just 16 bit instructions. the problem appeared in thumb2

radare:
we can use asm.cpu

test/unit/test_arch.c
radare: im calling it .CODE now, but see last comment with all my questions to follow

radare:
The next transition plans should go like this:

* Make a PR to merge every asm/anal plugin pair
* Slowly remove RAsmOp and RAnalOp, we may use RArchInstruction everywhere in a generic way
* Be consistent and add tests to make encode(decode()) in sync
* Move asm/d into arch/d
* Move parse into asm

Everyone is invited to this refactoring party

libr/arch/arch.c
```diff
@@ -0,0 +1,161 @@
...
+	case R_ASM_SYNTAX_ATT:
+	case R_ASM_SYNTAX_JZ:
+		a->syntax = syntax;
+		return true;
+	default:
+		return false;
+	}
+}
+
+R_API bool r_arch_set_endian(RArch *a, RArchEndian endian) {
```
ret2libc, line 86: this should probably check whether the plugin supports such endian, otherwise it should fail.

libr/arch/arch.c
```diff
@@ -0,0 +1,161 @@
...
+	case R_SYS_ENDIAN_BIG:
+	case R_SYS_ENDIAN_NONE:
+	case R_SYS_ENDIAN_BI:
+		a->endian = endian;
+		return true;
+	}
+	return false;
+}
+
+R_API bool r_arch_set_bits(RArch *a, int bits) {
```
ret2libc, line 99: this should probably check whether the plugin supports such bits value otherwise it should fail.

ret2libc: also, use RArchBits and not int

libr/include/r_arch.h
```diff
@@ -0,0 +1,100 @@
...
+R_API void r_arch_free(RArch *arch);
+
+R_API bool r_arch_encode(RArch *a, RArchInstruction *ins, RArchOptions opt);
+R_API bool r_arch_decode(RArch *a, RArchInstruction *ins, RArchOptions opt);
+
+R_API bool r_arch_setup(RArch *a, const char *arch, RArchBits bits, RArchEndian endian);
+R_API bool r_arch_set_syntax(RArch *a, int syntax);
+R_API bool r_arch_set_cpu(RArch *a, const char *cpu);
+R_API bool r_arch_set_endian(RArch *a, RArchEndian endian);
+R_API bool r_arch_set_bits(RArch *a, int bits);
```
ret2libc, line 93: RArchBits

libr/include/r_arch.h
```diff
@@ -0,0 +1,100 @@
...
+R_API void r_arch_instruction_fini(RArchInstruction *ins);
+
+R_API RArch *r_arch_new();
+R_API void r_arch_free(RArch *arch);
+
+R_API bool r_arch_encode(RArch *a, RArchInstruction *ins, RArchOptions opt);
+R_API bool r_arch_decode(RArch *a, RArchInstruction *ins, RArchOptions opt);
+
+R_API bool r_arch_setup(RArch *a, const char *arch, RArchBits bits, RArchEndian endian);
+R_API bool r_arch_set_syntax(RArch *a, int syntax);
```
ret2libc, line 90: define a RArchSynta

libr/include/r_arch.h
```diff
@@ -0,0 +1,100 @@
...
+R_API void r_arch_instruction_free(RArchInstruction *ins);
+R_API void r_arch_instruction_init(RArchInstruction *ins);
+R_API void r_arch_instruction_init_data(RArchInstruction *ins, ut64 addr, const ut8 *buf, size_t len);
+R_API void r_arch_instruction_init_code(RArchInstruction *ins, ut64 addr, const char *opstr);
+R_API void r_arch_instruction_fini(RArchInstruction *ins);
+
+R_API RArch *r_arch_new();
+R_API void r_arch_free(RArch *arch);
+
+R_API bool r_arch_encode(RArch *a, RArchInstruction *ins, RArchOptions opt);
```
ret2libc, line 86: `r_arch_encode` always uses the `ins.code`, right? I don't really like the fact that you have somehow to know that you have to setup ins.code before calling this function. I think I would prefer something like:
```C
ins = r_arch_decode (arch, "nop", opt);
```

libr/include/r_arch.h
```diff
@@ -0,0 +1,100 @@
...
+R_API bool r_arch_decode(RArch *a, RArchInstruction *ins, RArchOptions opt);
```
ret2libc, line 87: same here. `r_arch_decode` always use `ins.data`, why not making the data just a argument of this function. Something like
```C
ins = r_arch_decode (arch, (const ut8 *)"\x90\x90", opt);
```

libr/arch/p/arch_bf.c
```diff
@@ -0,0 +1,338 @@
...
+	ins->size = n;
+	return true;
+}
+
+// --------------------
+
+static bool decode(RArch *a, RArchInstruction *ins, RArchOptions opt) {
+	bool ret = false;
+	// TODO: merge disasm+analyze into 1 function to avoid duplicating logic code. just use the opt
+	if (opt & R_ARCH_OPTION_CODE || opt & R_ARCH_OPTION_SIZE) {
```
ret2libc, line 286: i'm not a huge fan of this `opt` approach.
I think this would be cleaner if each RArchPlugin would have:
```C
.size
.esil
.anal
.get_description
.opex
```

Under the hood of course those can share some code in each plugin, but from an API point of view I think it's cleaner.

libr/arch/arch.c
```diff
@@ -0,0 +1,161 @@
...
+                        a->cur = h;
+                        return true;
+                }
+        }
+        return false;
+}
+
+R_API bool r_arch_encode(RArch *a, RArchInstruction *ins, RArchOptions opt) {
+	r_return_val_if_fail (a, false);
+	if (a->cur && a->cur->encode) {
```
ret2libc, line 130: I think the options should be parsed at this point and then separate plugin callbacks should be called.

libr/include/r_arch.h
```diff
@@ -0,0 +1,100 @@
...
+#define R2_ARCH_H
+
+#include <r_util.h>
+#include <r_bind.h>
+#include <r_io.h> // Just for RIOBind which is needed for some plugins
+
+typedef int RArchBits;
+typedef int RArchEndian;
+
+typedef enum r_arch_decode_options_t {
```
ret2libc, line 13: is it really useful to have all these options as one enum?
What about having separate APIs for each operation?

The reason I'm saying this is that by having a big enum like this, if you pass a RArchInstruction around you never know whether it was initialized with code, data, size, anal, opex, desc or nothing. So you have to do checks all over the places to ensure that the thing you want (e.g. anal) is actually there.

libr/arch/arch.c
```diff
@@ -0,0 +1,161 @@
...
+	return false;
+}
+
+// plugins
+
+R_API bool r_arch_use(RArch *a, const char *name) {
+	r_return_val_if_fail (a && a->plugins, false);
+
+	if (!name || !*name) {
+		a->cur = NULL;
```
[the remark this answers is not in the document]
radare: yes, let me show you different situations:

* when building RArch without any plugin, it shouldnt crash, this happened already in 
* right now arch->cur == null means ther'es no arch plugin for the given arch because we are in a transitional state to let asm/anal use arch and their own plugins as fallback
* Once the transition is done ->cur null could mean that setup() or use() have failed. the thing is... if we do arch.use('404') what do u expect to happen? if you do RArch.new().cur what should it contain?

libr/arch/arch.c
```diff
@@ -0,0 +1,161 @@
...
+		return true;
```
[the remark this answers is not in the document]
radare: its returning true because null is a valid value there because it was passing "" or NULL on purpose to unset the current plugin

libr/arch/arch.c
```diff
@@ -0,0 +1,161 @@
...
+                if (!strcmp (h->name, name) && h->arch) {
+                        a->cur = h;
+                        return true;
+                }
+        }
+        return false;
+}
+
+R_API bool r_arch_encode(RArch *a, RArchInstruction *ins, RArchOptions opt) {
+	r_return_val_if_fail (a, false);
```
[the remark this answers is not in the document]
radare: imho no, we cant aassume this, at least until all the asm/anal plugins are moved here

libr/include/r_arch.h
radare: actually, this is not decoding, it would  be encoding.

this is how the intiial design was, and there was a bunch of limitations, thats why i decided to go for passing an instruction instead of just the instruction to **encode**

libr/include/r_arch.h
radare: thats wrong. you should pass the length too, and this was my initial design as i said. but i want to encode decoded instructions too, whta i have in mind its not just bytes <-> disasm.

radare: you also have to pass the program counter for that call to work, so you  end up with 5 arguments, also:

* using a struct you don't have to copy the bytes into the struct because when you decode you expect those bytes to be available in the struct to show them in hexdump: "90 nop", this 0x90 must be stored in the struct of the instruction too, so you end up filling the struct soon or late. what you are doing is what RAsm.disasemble() does and imho this is a higher level design on top of RArch

libr/arch/p/arch_bf.c
radare: if you do this you will end up analyzing the same bytes 5 times. the reason for having a bitfield option argument is to decode the instruction ONCE and fill the structure fields that are requested. lke RAnal.op does. which resulted to make analysis 2x faster by just picking the information we need instead of everything all the time

libr/arch/arch.c
radare: imho this is a bad decision because of how ALL the disassembler libraries work, its a bad design for performance reasons

libr/include/r_arch.h
radare: already replied this question 3 times above

libr/arch/arch.c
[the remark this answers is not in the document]
ret2libc: I'm ok if this is as a transitional state, but after that I think cur should always be set.
When you do r_arch_new it can either automatically sets a default plugin, which can even be a dummy one (like all bytes are invalid, or something like that) or just sets NULL (but if you set NULL, that means that the RArch instance is not in the right state to be used yet). Then, if you do `r_arch_use("404")` and then someone calls `r_arch_encode()` it is doing it wrong! It means the user of the API did not check for errors before, which is bad, so having a `r_return_val_if_fail` on `a->cur` will detect those bad uses of the API, since you should always check for errors.

libr/arch/arch.c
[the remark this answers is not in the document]
ret2libc: why would you unset the current plugin if you don't have a new one? What do you expect `encode/decode` to return when there is no plugin set?

libr/include/r_arch.h
ret2libc: what were the limitations? IMHO it is confusing having to set just few parts of a struct when you actually just care about the code field. Just directly pass that each time, it makes the API much clearer.

libr/include/r_arch.h
ret2libc: I don't see `PC` being set in the unit tests.
Also, if you have other usecases in mind please implement them as unit tests to show them off, so we can evaluate those as well (you can mark the tests as broken if those things are not implemented yet).
With regard to copying bytes, you have anyway to do that sooner or later, in both approaches.
With your approach, you copy bytes before calling r_arch_decode, with my approach you copy bytes inside r_arch_decode. I don't see a difference in that regard.

libr/arch/p/arch_bf.c
ret2libc: good point, I will think about this!

libr/arch/arch.c
ret2libc: what do you mean? can you add a bit more on this?

thestr4ng3r: Every callback would have to re-disassemble the bytes. Usually you can handle multiple RArchOptions in one pass.

libr/include/r_arch.h
thestr4ng3r: It's ok to bundle the input data in one struct. But encoded and decoded instruction should be clearly separated. If you want to have only single parameters for each, you can use two structs `RArchDecodedInstruction` and `RArchEncodedInstruction` but not everything in one.

libr/include/r_arch.h
radare: done

libr/include/r_arch.h
radare: done

radare:
This PR also introduces two new  goodies. that can/may be splitted into separate PRs or commits.

* Brainfuck arch plugin now computes the branch of the ] instructions
* Fallback branch lines are now displayed if they are not the expected ones

This PR also spots A LOT of places where the flags passed to RAnal.op() are wrong and this is not going to be covered in this  PR

libr/include/r_arch.h
thestr4ng3r: Yes, kind of. What you wrote there already looks really good:
```c
RArch *a = r_arch_new ();
RArchSetup setup = {
	.arch = "x86",
	.endian = R_SYS_ENDIAN_LITTLE,
	.bits = 32,
};
RArchInstance *ai = r_arch_instance (a, &setup);
RArchInstruction ins = {	};
r_arch_instance_encode (ai, ...
```

I wouldn't keep a cur pointer at all in the entire RArch lib. Imo this should be done in RAnal for example and RAnal should also be responsible for the pooling if we want that.

libr/include/r_vector.h
```diff
@@ -65,6 +65,7 @@ R_API void r_vector_clear(RVector *vec);
 
 // frees the vector and calls vec->free on every element if set.
 R_API void r_vector_free(RVector *vec);
+R_API void r_vector_fini(RVector *vec);
```
thestr4ng3r, line 68: Not necessary, `r_vector_clear()` does the same thing. You could make fini an alias though, to make it more intuitive.

radare: what about calling it r_vector_fini and make an alias named r_vector_clear? because having init without fini makes no sense

thestr4ng3r: Ok for me too

libr/arch/arch_session.c
```diff
@@ -0,0 +1,120 @@
...
+	return false;
+}
+
+R_API void r_arch_session_free(RArchSession *as) {
+	// somethng more
+	free (as);
+}
+
+//RArchPlugin *ap = r_arch_get_plugin (a, name);
+R_API RArchSession *r_arch_session_new(RArch *a, RArchPlugin *ap, RArchSetup *setup) {
```
ret2libc, line 95: this is never incrementing the refcount, so you can't use `unref` on this instance.

libr/arch/arch_session.c
```diff
@@ -0,0 +1,120 @@
...
+	r_return_val_if_fail (a && ap, NULL);
+	RArchSession *ai = R_NEW0 (RArchSession);
+	if (ai) {
+		ai->cur = ap;
+		if (!setup) {
+			RArchSetup _setup = {
```
ret2libc, line 101: i think it's better to allow each plugin to have a default setup, as it probably doesn't make a lot of sense to setup 32|64 bits for brainfuck or no endian for x86.

shlr/Makefile
```diff
@@ -34,7 +34,7 @@ TS_TIP=f049ba350f3f6019ce9a1cbb0975ebd154ef7ad3
 # NOTE: when you update SHELLPARSER_TIP or SHELLPARSER_BRA, also update them in shlr/meson.build
 SHELLPARSER_URL=https://github.com/ret2libc/radare2-shell-parser.git
 SHELLPARSER_BRA=master
-SHELLPARSER_TIP=167f8beca060bfab89f4b8e3ef47ffbb071437f2
+SHELLPARSER_TIP=3d82cad9d865cb6e65364f66f038d1d1d4d8818a
```
ret2libc, line 37: ??? what happened here?

thestr4ng3r: Everything happened here.

radare: im rebasing to master because of the conflicts and wtf is going on

shlr/meson.build
```diff
@@ -295,7 +295,7 @@ if get_option('use_treesitter')
     endif
 
     # NOTE: when you update SHELLPARSER_TIP or SHELLPARSER_BRA, also update them in shlr/Makefile
-    SHELLPARSER_TIP = '167f8beca060bfab89f4b8e3ef47ffbb071437f2'
+    SHELLPARSER_TIP = '3d82cad9d865cb6e65364f66f038d1d1d4d8818a'
```
radare, line 298: wtf here too

shlr/Makefile
ret2libc: too many changes in too many files, so as soon as there is something in master, you have conflicts :)

radare:
Please, review this https://github.com/radareorg/radare2/pull/16643 instead . im closing this pr because of conflicts

PR #16611, opened 2020-04-16
Add a test for noreturn propagation

thestr4ng3r:
The `tn 0x00001050` should not be necessary, this seems to be a bug.
However the purpose of this test is to test how the propagation itself works.


thestr4ng3r:
cc @kamou

PR #16612, opened 2020-04-16
Chop Basic Blocks instead of re-analyzing in noreturn propagation

thestr4ng3r:


libr/anal/block.c
```diff
@@ -543,3 +543,98 @@ R_API R_NULLABLE RList/*<RAnalBlock *>*/ *r_anal_block_shortest_path(RAnalBlock
...
+		}
+		ht_up_foreach (succs, noreturn_remove_unreachable_cb, fcn);
+	}
+
+	r_anal_block_unref (block);
+	ht_up_free (succs);
+}
+
+R_API bool r_anal_block_op_starts_at(RAnalBlock *bb, ut64 addr) {
+	int i;
```
radare, line 627: size_t

libr/core/canal.c
```diff
@@ -5637,46 +5646,59 @@ R_API void r_core_anal_propagate_noreturn(RCore *core, ut64 addr) {
...
+					goto kontinue;
+				}
+			} else {
+				r_anal_block_chop_noreturn (block, chop_addr);
+			}
+
+			RListIter *fit;
+			r_list_foreach (block->fcns, fit, f) {
+				bool found = false;
+				found = ht_uu_find (done, f->addr, &found);
```
radare, line 5690: bool found = ht_uu_find . instead of initializing it twice

libr/core/canal.c
```diff
@@ -5637,46 +5646,59 @@ R_API void r_core_anal_propagate_noreturn(RCore *core, ut64 addr) {
...
+				if (f->addr && !found && analyze_noreturn_function (core, f)) {
+					f->is_noreturn = true;
+					r_anal_noreturn_add (core->anal, NULL, f->addr);
+					ut64 *n = malloc (sizeof (ut64));
```
radare, line 5694: use ut64_new()

libr/core/canal.c
```diff
@@ -5637,46 +5646,59 @@ R_API void r_core_anal_propagate_noreturn(RCore *core, ut64 addr) {
...
+					*n = f->addr;
+					r_list_append (todo, n);
+					ht_uu_insert (done, *n, 1);
+				}
 			}
+kontinue:
```
radare, line 5700: kool

libr/anal/block.c
thestr4ng3r: +1, I just moved the code.

radare:
is new output better?

thestr4ng3r:
This is the difference. The new code does not yet merge contiguous basic blocks after the chopping. So both are correct but the old is a bit nicer. I am currently implementing the auto-merge, but it might be a bit tricky.
Screenshot from 2020-04-16 19-30-28

libr/core/canal.c
XVilka: Could have just used cyrillic "c"

libr/core/canal.c
thestr4ng3r: There is no such thing, just internally in cmd_debug.c

libr/anal/block.c
```diff
@@ -461,6 +462,20 @@ R_API void r_anal_block_add_switch_case(RAnalBlock *block, ut64 switch_addr, ut6
 	r_anal_switch_op_add_case (block->switch_op, case_addr, 0, case_addr);
 }
 
+R_API bool r_anal_block_op_starts_at(RAnalBlock *bb, ut64 addr) {
+	if (!r_anal_block_contains (bb, addr)) {
+		return false;
+	}
+	ut16 off = addr - bb->addr;
```
radare, line 469: i think you should:
* use ut64
* check if distance is bigger than UT16_MAX
* check if addr > bb->addr

libr/anal/block.c
```diff
@@ -543,3 +558,246 @@ R_API R_NULLABLE RList/*<RAnalBlock *>*/ *r_anal_block_shortest_path(RAnalBlock
...
+	succ->block = block;
+	succ->reachable = false; // reset for first iteration
+	ht_up_insert (succs, block->addr, succ);
+	return true;
+}
+
+static bool noreturn_successors_reachable_cb(RAnalBlock *block, void *user) {
+	HtUP *succs = user;
+	NoreturnSuccessor *succ = ht_up_find (succs, block->addr, NULL);
+	if (!succ) {
```
radare, line 588: be positive, use if (succ) { ... succ->reachable } return true

only one return statement here

libr/anal/block.c
```diff
@@ -543,3 +558,246 @@ R_API R_NULLABLE RList/*<RAnalBlock *>*/ *r_anal_block_shortest_path(RAnalBlock
...
+	AutomergeCtx *ctx = user;
+	ctx->cur_succ_count++;
+	RAnalBlock *block = ht_up_find (ctx->blocks, addr, NULL);
+	if (!block) {
+		// we shouldn't merge this one so GL_DONT_CARE
+		return true;
+	}
+	bool found;
+	RAnalBlock *pred = ht_up_find (ctx->predecessors, (ut64)block, &found);
+	if (!found) {
```
radare, line 695: be positive,use if (found) { first.

also. if ht_up_find fails always sets found? otherwse may be UB

libr/anal/block.c
```diff
@@ -543,3 +558,246 @@ R_API R_NULLABLE RList/*<RAnalBlock *>*/ *r_anal_block_shortest_path(RAnalBlock
...
+		// no predecessor found yet, this is the only one until now
+		ht_up_insert (ctx->predecessors, (ut64)block, ctx->cur_pred);
+	} else {
+		if (pred) {
+			// only one predecessor found so far, but we are the second so there are multiple now
+			ht_up_update (ctx->predecessors, (ut64)block, NULL);
+		} else {
```
radare, line 702: empty  unnecessary else block

libr/anal/block.c
```diff
@@ -543,3 +558,246 @@ R_API R_NULLABLE RList/*<RAnalBlock *>*/ *r_anal_block_shortest_path(RAnalBlock
...
+		ctx->cur_succ_count = 0;
+		r_anal_block_successor_addrs_foreach (block, automerge_predecessor_successor_cb, ctx);
+		ht_up_insert (ctx->visited_blocks, (ut64)block, (void *)ctx->cur_succ_count);
+	}
+	return true;
+}
+
+// Try to find the contiguous predecessors of all given blocks and merge them if possible,
+// i.e. if there are no other blocks that have this block as one of their successors
+R_API void r_anal_block_automerge(RList *blocks) {
```
radare, line 730: use r_return apis in all the R_API functions

libr/core/canal.c
radare: just do `HEAPTYPE(ut64);` on top of your file. and you will get this magic ut64_new

libr/include/r_anal.h
```diff
@@ -1316,6 +1316,10 @@ static inline bool r_anal_block_contains(RAnalBlock *bb, ut64 addr) {
 // The returned block will always be refd, i.e. it is necessary to always call r_anal_block_unref() on the return value!
 R_API RAnalBlock *r_anal_block_split(RAnalBlock *bb, ut64 addr);
 
+static inline bool r_anal_block_is_contiguous(RAnalBlock *a, RAnalBlock *b) {
+	return a->addr + a->size == b->addr;
```
radare, line 1320: add parenthesis to make clear you are not comparing the size, but the result of the addition

test/db/cmd/cmd_afb
```diff
@@ -8,7 +8,7 @@ afb.
 EOF
 EXPECT=<<EOF
 33
-main
+entry0
```
radare, line 11: looks like a regression

thestr4ng3r: Nope. That's what I thought too first, but this is actually more correct. `aa` has this special logic where it forces the name of the function at `entry0` to be called "entry0":
https://github.com/radareorg/radare2/blob/4afa4296ea32a314f1aad22b35926fb892223bf9/libr/core/canal.c#L4072-L4075
So this is the behavior that you would expect and also the one that is there after this pr:
Screenshot from 2020-04-17 17-22-53
But because before, aanr would re-analyze without this special entry0 treatment, it would call it differently:
Screenshot from 2020-04-17 17-23-11
This obviously makes no sense since noreturn propagation shouldn't just randomly change function names.

About this special treatment of entry0, I personally don't like it and I would remove it. But this should be done in another pr.

libr/anal/block.c
thestr4ng3r: >     * check if addr > bb->addr

This is done by `r_anal_block_contains()`
Agree on the other points.

libr/anal/block.c
thestr4ng3r: Yes, it always sets found. That is the sole purpose of the found parameter. To distinguish between `NULL` that is assigned to the given key and `NULL` when the key does not exist.

test/db/cmd/cmd_afb
radare: nice

PR #16623, opened 2020-04-17
Fix a segfault of rabin2 on rpi4

eagleoflqj:

**Detailed description**

A segmentation fault I met when using Cutter on rpi4.

**Test plan**

Buy an rpi4 and build r2.
```sh
rabin2 -Vj /bin/bunzip2
```
Get a segmentation fault.
Apply this change and rebuild.
No segmentation fault (at least here).

**Closing issues**

...

### DIFF
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -3676,7 +3676,7 @@ static void bin_elf_versioninfo(RCore *r, int mode) {
 			r_cons_printf ("{\"section_name\":\"%s\",\"address\":%"PFMT64u",\"offset\":%"PFMT64u",",
 					section_name, (ut64)addr, (ut64)offset);
 			r_cons_printf ("\"link\":%"PFMT64u",\"link_section_name\":\"%s\",\"entries\":[",
-					(ut32)link, link_section_name);
+					link, link_section_name);
 		} else {
 			r_cons_printf ("Version symbols section '%s' contains %"PFMT64u" entries:\n", section_name, num_entries);
 			r_cons_printf (" Addr: 0x%08"PFMT64x"  Offset: 0x%08"PFMT64x"  Link: %x (%s)\n",
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -3649,7 +3649,7 @@ static void bin_pe_versioninfo(RCore *r, int mode) {
 }
 
 static void bin_elf_versioninfo(RCore *r, int mode) {
-	const char *format = "bin/cur/info/versioninfo/%s%d";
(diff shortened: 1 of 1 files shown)


ret2libc:
Do you have any idea why the segfault? It's not exactly clear to me what this fixes a crash

eagleoflqj:
Arm use register to pass arg. On rpi4 (32 bit), if the arg is 64bit, it uses 2 registers. Seeing a "%llu", the _IO_vfprintf_internal picks 2 registers, but the second register is for the second argument which should be parsed by "%s".

ret2libc:
Cool now it makes sense! Thanks for the explanation!!

radare:
better use the PJ api so this kind of problems dont appear when someone changes a type of a random struct

eagleoflqj:
Now the code seems cleaner.

libr/core/cbin.c
```diff
@@ -3649,52 +3649,57 @@ static void bin_pe_versioninfo(RCore *r, int mode) {
 }
 
 static void bin_elf_versioninfo(RCore *r, int mode) {
-	const char *format = "bin/cur/info/versioninfo/%s%d";
+	const char *const format = "bin/cur/info/versioninfo/%s%d";
```
radare, line 3652: Double const is not needed

libr/core/cbin.c
```diff
@@ -3649,52 +3649,57 @@ static void bin_pe_versioninfo(RCore *r, int mode) {
...
 	int num_versym;
 	int num_verneed = 0;
 	int num_version = 0;
 	Sdb *sdb = NULL;
 	const char *oValue = NULL;
-	bool firstit_for_versym = true;
+	PJ *pj;
 	if (IS_MODE_JSON (mode)) {
-		r_cons_printf ("{\"versym\":[");
+		pj = pj_new ();
```
radare, line 3660: Check for null

libr/core/cbin.c
```diff
@@ -3649,52 +3649,57 @@ static void bin_pe_versioninfo(RCore *r, int mode) {
...
+		pj_o (pj);
+		pj_ka (pj, "versym");
 	}
 	for (num_versym = 0;; num_versym++) {
-		const char *versym_path = sdb_fmt (format, "versym", num_versym);
+		const char *const versym_path = sdb_fmt (format, "versym", num_versym);
```
radare, line 3665: Same

libr/core/cbin.c
eagleoflqj: Simply returns if null, because other calls to pj_new either don't check or just do so.

PR #16625, opened 2020-04-17
Fix cmd_substitution in grep_specifier ##newshell

ret2libc:
This should mostly fix all tests for newshell.

### DIFF
--- a/test/db/cmd/metadata
+++ b/test/db/cmd/metadata
@@ -21,7 +21,7 @@ CMDS=<<EOF
 af
 CCf-*~?
 CC hello
-CC world @ $$ + 32
+CC world @ $$+32
 CCf~?
 CCf-*
 CCf~?
--- a/test/db/cmd/sugar
+++ b/test/db/cmd/sugar
@@ -2,7 +2,7 @@ NAME=t/sugar
 FILE=-
 CMDS=<<EOF
 ph md5@s:hello
-ph md5 @ s:hello
+ph md5 @s:hello
 ph md5@x:01020304
 EOF
 EXPECT=<<EOF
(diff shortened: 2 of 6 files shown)


ret2libc:
This is it!!!

just 2 errors, but they are just wrong tests:

```
[XX] db/cmd/types tk - crash
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'tk -!!!! d'B%CCCC!9!!!!_' -
-- stdout
--- .a	2020-04-17 23:00:07.690751596 +0000
+++ .b	2020-04-17 23:00:07.690751596 +0000
@@ -1 +0,0 @@
-0x0


-- stderr
Error while parsing command: `tk -!!!! d'B%CCCC!9!!!!_`
```

```
[XX] db/cmd/cmd_hash hash_subcommands_not_noop
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '?! `#! ~?`
?v
?! `#? ~? `
?v
?! `#a ~? `
?v
' -
-- stdout
--- .a	2020-04-17 23:00:08.598831009 +0000
+++ .b	2020-04-17 23:00:08.598831009 +0000
@@ -1,4 +1,3 @@
0
0
-0x0
0


-- stderr
Error while parsing command: `?! `#? ~? ``
Error while parsing command: `?! `#a ~? ``
```

test/db/cmd/metadata
```diff
@@ -21,7 +21,7 @@ CMDS=<<EOF
 af
 CCf-*~?
 CC hello
-CC world @ $$ + 32
+CC world @ $$+32
```
radare, line 24: do this mean that spaces are no longer supported in math expressions?

test/db/cmd/sugar
```diff
@@ -2,7 +2,7 @@ NAME=t/sugar
 FILE=-
 CMDS=<<EOF
 ph md5@s:hello
-ph md5 @ s:hello
+ph md5 @s:hello
```
radare, line 5: this test was wrong.

test/db/cmd/metadata
ret2libc: It depends... `cmd @ arg` expects just one argument after @ , so if you put a space it becomes 3 arguments... We agreed on not having the concept of "math expressions" in the parser. I may try to fix this, but I'm not 100% sure it will be easy

PR #16628, opened 2020-04-18
Fix segfault when passing null string to "r_listinfo_new"

abcSup:

**Detailed description**
Using `om=` in debug mode will cause segfault.
```
$ r2 -d rax2
Process with PID 31203 started...
= attach 31203 31203
bin.baddr 0x56254d888000
Using 0x56254d888000
asm.bits 64
Warning: r_bin_file_hash: file exceeds bin.hashlimit
 -- I C dead bugs everywhere!
[0x7f1769e54090]> om
 1 fd: 3 +0x00000000 0x00000000 - 0x7ffffffffffffffe rwx 
[0x7f1769e54090]> om=
```
Because a default `RIOMap` is created without setting its name when r2 is in debug mode.
https://github.com/radareorg/radare2/blob/35290dd51d815ed0e36623b1fba63c6f8a93bd84/libr/core/cfile.c#L688-L691

Thefore, when calling `om=`, it passes `map->name` (NULL) to `r_listinfo_new` and the function calls `strdup (NULL)`, which causes the segfault.
https://github.com/radareorg/radare2/blob/35290dd51d815ed0e36623b1fba63c6f8a93bd84/libr/core/cmd_open.c#L829-L838

A quick fix is to check the `name` and `extra` args.

### DIFF
--- a/libr/core/visual.c
+++ b/libr/core/visual.c
@@ -4337,11 +4337,11 @@ dodo:
 R_API RListInfo *r_listinfo_new(const char *name, RInterval pitv, RInterval vitv, int perm, const char *extra) {
 	RListInfo *info = R_NEW (RListInfo);
 	if (info) {
-		info->name = strdup (name);
+		info->name = name ? strdup (name) : NULL;
 		info->pitv = pitv;
 		info->vitv = vitv;
 		info->perm = perm;
-		info->extra = strdup (extra);
+		info->extra = extra ? strdup (extra) : NULL;
 	}
 	return info;
 }


radare:
Imho its wrong to name a null named map that should be asserting way before reaching this point

abcSup:
Are you suggesting that instead of assigning `info->name` to `NULL`, there should be an assertion to check for `NULL` arguments and return failure if that is so?

libr/core/visual.c
```diff
@@ -4337,11 +4337,11 @@ R_API int r_core_visual(RCore *core, const char *input) {
 R_API RListInfo *r_listinfo_new(const char *name, RInterval pitv, RInterval vitv, int perm, const char *extra) {
 	RListInfo *info = R_NEW (RListInfo);
 	if (info) {
-		info->name = strdup (name);
+		info->name = name ? strdup (name) : NULL;
```
radare, line 4340: imho its better to do strdup (name?name:"")

radare:
i dont know why this map have null as name, and imho this should never happen. cc @ret2libc @thestr4ng3r

thestr4ng3r:
Yeah might be better to check where this map comes from and fix it there. If that is done, `r_return_if_fail()` can be used here.

abcSup:
There are two places that create `RIOMap`s with null name.

One is in `r_core_bin_load` when r2 is in debug mode.
https://github.com/radareorg/radare2/blob/35290dd51d815ed0e36623b1fba63c6f8a93bd84/libr/core/cfile.c#L688-L691
Another one is in `radare2.c` when r2 is in debug mode.
https://github.com/radareorg/radare2/blob/35290dd51d815ed0e36623b1fba63c6f8a93bd84/libr/main/radare2.c#L1073-L1076

> Yeah might be better to check where this map comes from and fix it there. If that is done, r_return_if_fail() can be used here.

@thestr4ng3r could we just assign `map->name` to `desc->name`?

radare:
i dont think a null name map should be valid, but i prefer to not have a crash for now

PR #16629, opened 2020-04-18
Remove r_pvector_upper_bound() and add r_vector_lower_bound()

thestr4ng3r:
**Detailed description**

The upper bound was more wrong and confusing than actually helpful, also used nowhere until now.
Lower bound is quite helpful for RVector and I will need it for variable accesses.

**Test plan**

run the unit tests


radare:
you missed the oportunity to add r_vector_fini

PR #16630, opened 2020-04-18
x86_64: Support mov r64, 0xffffffffffffffff

kazarmy:

**Detailed description**

This pr enables support for mov r64, 0xffffffffffffffff in the x86_64 assembler.

**Test plan**

**Closing issues**

...

### DIFF
--- a/libr/asm/asm.c
+++ b/libr/asm/asm.c
@@ -191,16 +191,20 @@ static void plugin_free(RAsmPlugin *p) {
 	}
 }
 
-R_API RAsm *r_asm_new() {
+R_API RAsm *r_asm_new(RNum *num) {
 	int i;
 	RAsm *a = R_NEW0 (RAsm);
 	if (!a) {
 		return NULL;
 	}
-	a->num = r_num_new (NULL, NULL, NULL);
-	if (!a->num) {
-		free (a);
-		return NULL;
+	if (num) {
+		a->num = num;
+	} else {
+		a->num = r_num_new (NULL, NULL, NULL);
+		if (!a->num) {
+			free (a);
+			return NULL;
+		}
 	}
 	a->dataalign = 1;
 	a->bits = R_SYS_BITS;
--- a/libr/asm/asm.c
+++ b/libr/asm/asm.c
@@ -199,6 +199,7 @@ R_API RAsm *r_asm_new() {
 	}
 	a->num = r_num_new (NULL, NULL, NULL);
 	if (!a->num) {
+		free (a);
 		return NULL;
 	}
 	a->dataalign = 1;
--- a/libr/asm/asm.c
+++ b/libr/asm/asm.c
@@ -197,6 +197,10 @@ R_API RAsm *r_asm_new() {
 	if (!a) {
 		return NULL;
 	}
+	a->num = r_num_new (NULL, NULL, NULL);
+	if (!a->num) {
+		return NULL;
+	}
 	a->dataalign = 1;
 	a->bits = R_SYS_BITS;
 	a->bitshift = 0;
@@ -248,6 +252,7 @@ R_API void r_asm_free(RAsm *a) {
 	if (!a) {
 		return;
 	}
+	r_num_free (a->num);
 	if (a->cur && a->cur->fini) {
 		a->cur->fini (a->cur->user);
 	}
--- a/libr/asm/asm.c
+++ b/libr/asm/asm.c
@@ -199,6 +199,7 @@ R_API RAsm *r_asm_new(RNum *num) {
 	}
 	if (num) {
 		a->num = num;
+		r_num_ref (num);
 	} else {
 		a->num = r_num_new (NULL, NULL, NULL);
 		if (!a->num) {
--- a/libr/asm/asm.c
+++ b/libr/asm/asm.c
@@ -197,11 +197,6 @@ R_API RAsm *r_asm_new() {
 	if (!a) {
 		return NULL;
 	}
-	a->num = r_num_new (NULL, NULL, NULL);
-	if (!a->num) {
-		free (a);
-		return NULL;
-	}
 	a->dataalign = 1;
 	a->bits = R_SYS_BITS;
 	a->bitshift = 0;
@@ -253,7 +248,6 @@ R_API void r_asm_free(RAsm *a) {
 	if (!a) {
 		return;
 	}
-	r_num_unref (a->num);
 	if (a->cur && a->cur->fini) {
 		a->cur->fini (a->cur->user);
 	}
--- a/libr/asm/asm.c
+++ b/libr/asm/asm.c
@@ -258,7 +258,7 @@ R_API void r_asm_free(RAsm *a) {
 	if (!a) {
 		return;
 	}
-	// r_num_free (a->num);
+	r_num_free (a->num);
 	if (a->cur && a->cur->fini) {
 		a->cur->fini (a->cur->user);
 	}
--- a/libr/asm/asm.c
+++ b/libr/asm/asm.c
@@ -191,21 +191,16 @@ static void plugin_free(RAsmPlugin *p) {
 	}
 }
 
-R_API RAsm *r_asm_new(RNum *num) {
+R_API RAsm *r_asm_new() {
 	int i;
 	RAsm *a = R_NEW0 (RAsm);
 	if (!a) {
 		return NULL;
 	}
-	if (num) {
-		a->num = num;
-		r_num_ref (num);
-	} else {
-		a->num = r_num_new (NULL, NULL, NULL);
-		if (!a->num) {
-			free (a);
-			return NULL;
-		}
+	a->num = r_num_new (NULL, NULL, NULL);
+	if (!a->num) {
+		free (a);
+		return NULL;
 	}
 	a->dataalign = 1;
 	a->bits = R_SYS_BITS;
@@ -258,7 +253,7 @@ R_API void r_asm_free(RAsm *a) {
 	if (!a) {
 		return;
 	}
-	r_num_free (a->num);
+	r_num_unref (a->num);
 	if (a->cur && a->cur->fini) {
 		a->cur->fini (a->cur->user);
 	}
--- a/libr/asm/asm.c
+++ b/libr/asm/asm.c
@@ -252,7 +252,7 @@ R_API void r_asm_free(RAsm *a) {
 	if (!a) {
 		return;
 	}
-	r_num_free (a->num);
+	// r_num_free (a->num);
 	if (a->cur && a->cur->fini) {
 		a->cur->fini (a->cur->user);
 	}
--- a/libr/asm/p/asm_x86_nz.c
+++ b/libr/asm/p/asm_x86_nz.c
@@ -1793,7 +1793,7 @@ static int opmov(RAsm *a, ut8 *data, const Opcode *op) {
 		if (!op->operands[1].is_good_flag) {
 			return -1;
 		}
-		if (op->operands[1].immediate == -1) {
+		if (op->operands[1].immediate == -1 && a->num && a->num->nc.errors > 0) {
 			return -1;
 		}
 		immediate = op->operands[1].immediate * op->operands[1].sign;
--- a/libr/asm/p/asm_x86_nz.c
+++ b/libr/asm/p/asm_x86_nz.c
@@ -4912,9 +4912,8 @@ static int parseOperand(RAsm *a, const char *str, Operand *op, bool isrepop) {
 		op->immediate = getnum (a, str);
 		if (op->immediate == -1) {
 			op->is_good_flag = false;
-			char *num_str = strdup (str);
+			char *num_str = r_str_trim_dup (str);
 			if (num_str) {
-				r_str_trim (num_str);
 				op->is_good_flag = !strcmp (num_str, "0xffffffffffffffff");
 				free (num_str);
 			}
--- a/libr/asm/p/asm_x86_nz.c
+++ b/libr/asm/p/asm_x86_nz.c
@@ -1793,6 +1793,9 @@ static int opmov(RAsm *a, ut8 *data, const Opcode *op) {
 		if (!op->operands[1].is_good_flag) {
 			return -1;
 		}
+		if (op->operands[1].immediate == -1 && a->num && a->num->nc.errors > 0) {
+			return -1;
+		}
 		immediate = op->operands[1].immediate * op->operands[1].sign;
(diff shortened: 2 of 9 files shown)


libr/asm/asm.c
```diff
@@ -246,6 +250,7 @@ R_API void r_asm_free(RAsm *a) {
 	if (!a) {
 		return;
 	}
+	// r_num_free (a->num);
```
radare, line 253: leak?

libr/asm/asm.c
```diff
@@ -195,6 +195,10 @@ R_API RAsm *r_asm_new() {
 	if (!a) {
 		return NULL;
 	}
+	a->num = r_num_new (NULL, NULL, NULL);
```
radare, line 198: this is initialized from RCore afaik, i think its fine to have this here, but we should cover the proper reinit from core

libr/asm/asm.c
```diff
@@ -195,6 +195,10 @@ R_API RAsm *r_asm_new() {
...
+	if (!a->num) {
+		return NULL;
```
radare, line 200: memory leak

libr/asm/asm.c
radare: RNum should be refcounted, can you add support for it and just do r_num_unref here? this way we can have shared instances between different modules and not leak or double free them

libr/util/unum.c
```diff
@@ -240,6 +242,9 @@ R_API ut64 r_num_get(RNum *num, const char *str) {
 			ret = strtoull (str + 2, NULL, 16);
 			// sscanf (str+2, "%"PFMT64x, &ret);
 		}
+		if (errno == ERANGE) {
```
radare, line 245: nice, didnt knew strotull filled the errno

libr/asm/asm.c
kazarmy: sorry 4dafc48265ad332f9553a6f6b1c5b2efa20f8180

libr/asm/asm.c
kazarmy: If uncommenting that results in a double-free, still need to find out where the previous free is done

radare:
Use refcounting and u wont need to find that. Also if u build with sys/sanitize you will get that info

kazarmy:
eoutoftime today, building with asan is slow and there's a pathogen that's making my nose miserable (though thankfully not much else), so maybe in the next 24h?

radare:
Sure no hurry. Stay safe

radare:
ready to go?

kazarmy:
not yet ... still haven't done the refcounting

kazarmy:
> ready to go?

i think so

libr/asm/asm.c
kazarmy: > RNum should be refcounted ...

ok 79d65d9a4ad0d9245339c2e9ca6527f2ae6ecf9a

libr/asm/asm.c
kazarmy: ok d615cdb614476938681cc0a6fd477604d2e94f9f

libr/asm/asm.c
```diff
@@ -189,12 +189,22 @@ static void plugin_free(RAsmPlugin *p) {
...
 	int i;
 	RAsm *a = R_NEW0 (RAsm);
 	if (!a) {
 		return NULL;
 	}
+	if (num) {
+		a->num = num;
+		r_num_ref (num);
+	} else {
+		a->num = r_num_new (NULL, NULL, NULL);
```
radare, line 202: there's no need for this, the rnum api works with NULL as instance, it will just not resolve any non-numeric name

libr/asm/asm.c
```diff
@@ -189,12 +189,22 @@ static void plugin_free(RAsmPlugin *p) {
 	}
 }
 
-R_API RAsm *r_asm_new() {
+R_API RAsm *r_asm_new(RNum *num) {
```
radare, line 192: i wont touch that constructor. the r-arch api will affect all this and you can just instantaite num like we did all the time from core. it will be good to add refcounting to the RNum too

libr/asm/asm.c
```diff
@@ -246,6 +256,7 @@ R_API void r_asm_free(RAsm *a) {
 	if (!a) {
 		return;
 	}
+	r_num_free (a->num);
```
radare, line 259: use ref/unref

libr/egg/egg.c
```diff
@@ -48,7 +48,7 @@ R_API REgg *r_egg_new() {
 	if (!egg->syscall) {
 		goto beach;
 	}
-	egg->rasm = r_asm_new ();
+	egg->rasm = r_asm_new (NULL);
```
radare, line 51: uhm thats an issue because it will be inconsistent. you should create that instance from outside or pass it as argument like in the other constructor. but all the class constructors in r2 get 0 parameters because they should not have side effects

libr/util/unum.c
```diff
@@ -73,7 +76,7 @@ R_API RNum *r_num_new(RNumCallback cb, RNumCallback2 cb2, void *ptr) {
 }
 
 R_API void r_num_free(RNum *num) {
-	R_FREE (num);
+	r_unref (num, free);
```
radare, line 79: use r_num_unref (num); instead. the r_num_free callback is the one that will be used and should be used

libr/asm/asm.c
kazarmy: ?? ... 908b0e042e935b2324aa699c6229adf83d47b5bd resolves this?

libr/asm/asm.c
kazarmy: ok 908b0e042e935b2324aa699c6229adf83d47b5bd

libr/asm/asm.c
kazarmy: ok 908b0e042e935b2324aa699c6229adf83d47b5bd

libr/egg/egg.c
kazarmy: ok 908b0e042e935b2324aa699c6229adf83d47b5bd

libr/util/unum.c
kazarmy: ok 908b0e042e935b2324aa699c6229adf83d47b5bd

kazarmy:
I do think that with 908b0e042e935b2324aa699c6229adf83d47b5bd the api becomes more error-prone, since now one must remember to use `r_num_unref()` with `core->num` instead of `r_num_free()`.

libr/asm/asm.c
```diff
@@ -195,6 +195,11 @@ R_API RAsm *r_asm_new() {
 	if (!a) {
 		return NULL;
 	}
+	a->num = r_num_new (NULL, NULL, NULL);
```
radare, line 198: i dont think RAsm should creat that instance here, mainly because it is leaking because asm->num is set from core and rasm already.

kazarmy: ok 810ea16d67cd4b26f319b3d36092a7c6ca1d3b16 ... also removed refcounting from RNum since it's clearly not needed now cc9f9433733ce25bf5ba761bedd8807d6a35a3dd

libr/main/rasm2.c
```diff
@@ -34,6 +34,9 @@ static RAsmState *__as_new() {
 	if (as) {
 		as->l = r_lib_new (NULL, NULL);
 		as->a = r_asm_new ();
+		if (as->a) {
```
radare, line 37: This is wrong. Becasue core creates the instance, and owns it. So the num instance you create here is leaking and lost.

libr/main/rasm2.c
```diff
@@ -42,6 +45,9 @@ static RAsmState *__as_new() {
 }
 
 static void __as_free(RAsmState *as) {
+	if (as->a) {
```
radare, line 48: This may cause a double free because asm doesnt owns this instance when used from Core

libr/asm/p/asm_x86_nz.c
```diff
@@ -1793,7 +1793,7 @@ static int opmov(RAsm *a, ut8 *data, const Opcode *op) {
 		if (!op->operands[1].is_good_flag) {
 			return -1;
 		}
-		if (op->operands[1].immediate == -1) {
+		if (op->operands[1].immediate == -1 && a->num && a->num->nc.errors > 0) {
```
radare, line 1796: Will be good to not depend on RNum at all. So this conditional should work when a.num is null

libr/main/rasm2.c
kazarmy: We are not passing core into `r_main_rasm2()`. And asan shows no relevant leaks.

libr/main/rasm2.c
kazarmy: Again, this doesn't happen because of https://github.com/radareorg/radare2/pull/16630#discussion_r411462961.

libr/main/rasm2.c
radare: Im talking about r2. I know this is fine for rasm2

libr/main/rasm2.c
radare: Ah ok this code is for rasm2

libr/asm/p/asm_x86_nz.c
radare: Fix this and ill merge

libr/main/rasm2.c
kazarmy: r2 doesn't call that function. `__as_new()` is a static function only visible in `libr/main/rasm2.c` and only called from `r_main_rasm2()`.

libr/asm/p/asm_x86_nz.c
kazarmy: I'm not that desperate to merge this one in but ... you have a deal though I need at least another 24h

radare:
Not in a hurry to merge now but will be good to merge it soon or late :)

Thanks!

test/db/asm/x86_64
```diff
@@ -54,10 +54,8 @@ a "mov rax, 0x100000000" 48b80000000001000000
 ad "movabs rax, 0x100000000" 48b80000000001000000
 ad "mov rax, 0" 48c7c000000000
 a "mov rax, -1" 48c7c0ffffffff
-adB "mov rax, 0xffffffffffffffff" 48c7c0ffffffff
-a "mov rax, -2" 48c7c0feffffff
```
ret2libc, line 58: why removing this test? I'd keep it if it still works. If it doesn't mark it as broken.

kazarmy: it was removed since -2 is not much different from -1 ... but keeping it is also fine 0257ab8b5241e8f17c990680145041b17ebf9ed1

libr/asm/p/asm_x86_nz.c
kazarmy: > Will be good to not depend on RNum at all

ok 22cf08b6f0cce706372a3932d8622071c0912d88 ... i haven't reverted the changes to `libr/util/unum.c` since they are improvements

libr/asm/p/asm_x86_nz.c
```diff
@@ -4913,6 +4910,15 @@ static int parseOperand(RAsm *a, const char *str, Operand *op, bool isrepop) {
 			str = ++p;
 		}
 		op->immediate = getnum (a, str);
+		if (op->immediate == -1) {
+			op->is_good_flag = false;
+			char *num_str = strdup (str);
```
radare, line 4915: use r_str_trim_dup() and save one line

libr/asm/p/asm_x86_nz.c
```diff
@@ -4913,6 +4910,15 @@ static int parseOperand(RAsm *a, const char *str, Operand *op, bool isrepop) {
...
+			if (num_str) {
+				r_str_trim (num_str);
+				op->is_good_flag = !strcmp (num_str, "0xffffffffffffffff");
```
radare, line 4918: that will fail if the string is 0xFFffFFffFF

libr/asm/p/asm_x86_nz.c
kazarmy: ok 185eec4a44cd94f7a9175a7d41fcdca1020898ad

libr/asm/p/asm_x86_nz.c
kazarmy: ok ab09af7ff29158b2e78fa4a5dc2861b094fc4638

kazarmy:
> there must be a better way tha comparing a string to compare a number

it's only 1 number, and it's an edge case

kazarmy:
besides there doesn't seem to be a use case of either a human or a machine to write 18446744073709551615 instead of 0xffffffffffffffff or -1

libr/asm/p/asm_x86_nz.c
```diff
@@ -4913,6 +4910,14 @@ static int parseOperand(RAsm *a, const char *str, Operand *op, bool isrepop) {
 			str = ++p;
 		}
 		op->immediate = getnum (a, str);
+		if (op->immediate == -1) {
+			op->is_good_flag = false;
+			char *num_str = r_str_trim_dup (str);
+			if (num_str) {
+				op->is_good_flag = !r_str_casecmp (num_str, "0xffffffffffffffff");
```
radare, line 4917: i still dont see why are you comparing the string, if the sole reason is to differenciate -1 of error in getnum with UT64_MAX, just check the error bit in RNum. comparing 0xfffff seems wrong to me

radare:
btw  just as an example: getnum handles -1U as an alias for 0xfffff and with your code that will not work

libr/asm/p/asm_x86_nz.c
kazarmy: >  comparing 0xfffff seems wrong to me

it feels wrong to me too but technically not a blocker and the offending code is small ... and depending on RNum's error bit means depending on RNum's error handling code

kazarmy:
actually -1U works, probably for the same reason why -1 works, but 0xffffffffffffffffU doesn't so fixing 2efe65cf1d46c45e9646264149e3ff2502019129 d0fdd13e83c2f235d944720cdc16f7631c8e69fa

radare:
what about 1u64 and 1u32 ?

kazarmy:
1u64_1u32

They don't seem to have the proper definitions, especially if you accept how [Rust defines 1u64 and 1u32](https://doc.rust-lang.org/reference/tokens.html#integer-literals). But even with the above (negative) definitions 1u64 and 1u32 does work with 2efe65cf1d46c45e9646264149e3ff2502019129.

PR #16632, opened 2020-04-18
Refactor Variable Accesses out of SDB

### LINKED ISSUE #15944: Varsub is very slow

thestr4ng3r:
Fix #15944

### DIFF
--- a/libr/anal/function.c
+++ b/libr/anal/function.c
@@ -201,9 +201,14 @@ R_API bool r_anal_function_relocate(RAnalFunction *fcn, ut64 addr) {
 			acc->offset -= delta;
 		}
 	}
-	HtUP *inst_vars_new = ht_up_new0 ();
-	if (inst_vars_new) {
-		ht_up_foreach (fcn->inst_vars, inst_vars_relocate_cb, inst_vars_new);
+	InstVarsRelocateCtx ctx = {
+		.inst_vars_new  = ht_up_new0 (),
+		.delta = delta
+	};
+	if (ctx.inst_vars_new) {
+		ht_up_foreach (fcn->inst_vars, inst_vars_relocate_cb, &ctx);
+		ht_up_free (fcn->inst_vars);
+		fcn->inst_vars = ctx.inst_vars_new;
 	}
 
 	fcn->addr = addr;
--- a/libr/anal/function.c
+++ b/libr/anal/function.c
@@ -70,6 +70,7 @@ R_API RAnalFunction *r_anal_function_new(RAnal *anal) {
 	fcn->is_noreturn = false;
 	fcn->meta._min = UT64_MAX;
 	r_pvector_init (&fcn->vars, (RPVectorFree)r_anal_var_free);
+	fcn->inst_vars = ht_up_new0 ();
 	return fcn;
 }
 
@@ -95,6 +96,8 @@ R_API void r_anal_function_free(void *_fcn) {
 		ht_pp_delete (anal->ht_name_fun, fcn->name);
 	}
 
+	ht_up_free (fcn->inst_vars);
+	fcn->inst_vars = NULL;
 	r_pvector_clear (&fcn->vars);
 
 	free (fcn->name);
@@ -168,6 +171,17 @@ R_API RAnalFunction *r_anal_get_function_at(RAnal *anal, ut64 addr) {
 	return NULL;
 }
 
+typedef struct {
+	HtUP *inst_vars_new;
+	st64 delta;
+} InstVarsRelocateCtx;
+
+static bool inst_vars_relocate_cb(void *user, const ut64 k, const void *v) {
+	InstVarsRelocateCtx *ctx = user;
+	ht_up_insert (ctx->inst_vars_new, k - ctx->delta, (void *)v);
+	return true;
+}
+
 R_API bool r_anal_function_relocate(RAnalFunction *fcn, ut64 addr) {
 	if (fcn->addr == addr) {
 		return true;
@@ -176,6 +190,22 @@ R_API bool r_anal_function_relocate(RAnalFunction *fcn, ut64 addr) {
 		return false;
 	}
 	ht_up_delete (fcn->anal->ht_addr_fun, fcn->addr);
+
+	// relocate the var accesses (their addrs are relative to the function addr)
+	st64 delta = (st64)addr - (st64)fcn->addr;
+	void **it;
+	r_pvector_foreach (&fcn->vars, it) {
+		RAnalVar *var = *it;
+		RAnalVarAccess *acc;
+		r_vector_foreach (&var->accesses, acc) {
+			acc->offset -= delta;
+		}
+	}
+	HtUP *inst_vars_new = ht_up_new0 ();
+	if (inst_vars_new) {
+		ht_up_foreach (fcn->inst_vars, inst_vars_relocate_cb, inst_vars_new);
+	}
+
 	fcn->addr = addr;
 	ht_up_insert (fcn->anal->ht_addr_fun, addr, fcn);
 	return true;
--- a/libr/anal/function.c
+++ b/libr/anal/function.c
@@ -69,7 +69,7 @@ R_API RAnalFunction *r_anal_function_new(RAnal *anal) {
 	fcn->bp_frame = true;
 	fcn->is_noreturn = false;
 	fcn->meta._min = UT64_MAX;
-	r_pvector_init (&fcn->vars, (RPVectorFree)r_anal_var_free);
+	r_pvector_init (&fcn->vars, NULL);
 	fcn->inst_vars = ht_up_new0 ();
 	return fcn;
 }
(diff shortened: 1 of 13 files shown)


thestr4ng3r:
```
==32261== Conditional jump or move depends on uninitialised value(s)
==32261==    at 0x598CDCE: var_accesses_list (cmd_anal.c:978)
==32261==    by 0x598D017: list_vars (cmd_anal.c:1016)
==32261==    by 0x598DBFA: var_cmd (cmd_anal.c:1199)
==32261==    by 0x5996F44: cmd_anal_fcn (cmd_anal.c:3380)
==32261==    by 0x59AF9B7: cmd_anal (cmd_anal.c:9958)
==32261==    by 0x5A09FD0: r_cmd_call (cmd_api.c:248)
==32261==    by 0x5A0464B: r_core_cmd_subst_i (cmd.c:3767)
==32261==    by 0x5A006E8: r_core_cmd_subst (cmd.c:2685)
==32261==    by 0x5A00834: r_core_cmd_subst (cmd.c:2717)
==32261==    by 0x5A00834: r_core_cmd_subst (cmd.c:2717)
==32261==    by 0x5A00834: r_core_cmd_subst (cmd.c:2717)
==32261==    by 0x5A073A1: run_cmd_depth (cmd.c:6627)
==32261==
```

libr/anal/function.c
```diff
@@ -176,6 +190,27 @@ R_API bool r_anal_function_relocate(RAnalFunction *fcn, ut64 addr) {
 		return false;
 	}
 	ht_up_delete (fcn->anal->ht_addr_fun, fcn->addr);
+
+	// relocate the var accesses (their addrs are relative to the function addr)
+	st64 delta = (st64)addr - (st64)fcn->addr;
```
radare, line 195: sounds like this wont work with kernel addresses

libr/anal/var.c
```diff
@@ -262,35 +259,59 @@ R_API RAnalVar *r_anal_function_get_var(RAnalFunction *fcn, char kind, int delta
...
-		} else if (v1->kind == R_ANAL_VAR_KIND_SPV) {
-			regname = r_reg_get_name (a->reg, R_REG_NAME_SP);
+	return r_reg_getv (anal->reg, regname) + var->delta;
+}
+
+R_API st64 r_anal_function_get_var_stackptr_at(RAnalFunction *fcn, int delta, ut64 addr) {
+	st64 offset = (st64)addr - (st64)fcn->addr;
+	RPVector *inst_accesses = ht_up_find (fcn->inst_vars, offset, NULL);
+	if (!inst_accesses) {
+		return INT_MAX;
```
radare, line 278: INT64_MAX

radare: err ST64_MAX

libr/anal/var.c
```diff
@@ -262,35 +259,59 @@ R_API RAnalVar *r_anal_function_get_var(RAnalFunction *fcn, char kind, int delta
...
+	}
+	RAnalVar *var = r_pvector_at (inst_accesses, 0);
+	size_t index;
+	r_vector_lower_bound (&var->accesses, offset, index, ACCESS_CMP);
+	RAnalVarAccess *acc = NULL;
+	if (index < var->accesses.len) {
+		acc = r_vector_index_ptr (&var->accesses, index);
+	}
+	if (!acc || acc->offset != offset) {
+		return INT_MAX;
```
radare, line 300: same

libr/anal/var.c
```diff
@@ -262,35 +259,59 @@ R_API RAnalVar *r_anal_function_get_var(RAnalFunction *fcn, char kind, int delta
...
-	if (v1) {
-		if (v1->kind == R_ANAL_VAR_KIND_BPV) {
-			regname = r_reg_get_name (a->reg, R_REG_NAME_BP);
-		} else if (v1->kind == R_ANAL_VAR_KIND_SPV) {
-			regname = r_reg_get_name (a->reg, R_REG_NAME_SP);
+	return r_reg_getv (anal->reg, regname) + var->delta;
+}
+
+R_API st64 r_anal_function_get_var_stackptr_at(RAnalFunction *fcn, int delta, ut64 addr) {
+	st64 offset = (st64)addr - (st64)fcn->addr;
```
radare, line 275: looks wrong. that wont work when the higher bit is set

libr/core/disasm.c
```diff
@@ -975,6 +975,7 @@ static char *get_op_ireg (void *user, ut64 addr) {
 }
 
 static int get_ptr_at(void *user, RAnalFunction *fcn, int delta, ut64 addr) {
+	return (int)r_anal_function_get_var_stackptr_at (fcn, delta, addr);
```
radare, line 978: dead code and scary cast

libr/anal/var.c
thestr4ng3r: It seems to work fine. I added tests for exactly these cases.

libr/anal/var.c
```diff
@@ -262,35 +256,58 @@ R_API RAnalVar *r_anal_function_get_var(RAnalFunction *fcn, char kind, int delta
...
+	} else if (var->kind == R_ANAL_VAR_KIND_SPV) {
+		regname = r_reg_get_name (anal->reg, R_REG_NAME_SP);
 	}
-	RAnalVar *v1 = r_anal_function_get_var_byname (fcn, name);
-	if (v1) {
-		if (v1->kind == R_ANAL_VAR_KIND_BPV) {
-			regname = r_reg_get_name (a->reg, R_REG_NAME_BP);
-		} else if (v1->kind == R_ANAL_VAR_KIND_SPV) {
-			regname = r_reg_get_name (a->reg, R_REG_NAME_SP);
+	return r_reg_getv (anal->reg, regname) + var->delta;
```
radare, line 268: Regname can be null at this point i think. And this will make it assert

libr/anal/var.c
```diff
@@ -346,38 +362,48 @@ struct VarUsedType {
...
-	RAnalFunction *fcn = NULL;
-	if (sdb_fmt_tobin (var_def, SDB_VARUSED_FMT, &vut) == 4) {
-		fcn = r_anal_get_function_at (anal, vut.fcn_addr);
-		if (fcn) {
-			res = r_anal_function_get_var (fcn, vut.type[0], vut.delta);
-		}
-		sdb_fmt_free (&vut, SDB_VARUSED_FMT);
+R_API R_BORROW RPVector *r_anal_function_get_vars_used_at(RAnalFunction *fcn, ut64 op_addr) {
+	r_return_val_if_fail (fcn, NULL);
+	return ht_up_find (fcn->inst_vars, (st64)op_addr - (st64)fcn->addr, NULL);
```
radare, line 367: Ill probably put that substraction in a local const variable for readability. And im still unhappy with that signed cast but we can work on thst later in a separate pr if we see issues. Thanks for writing the tests

libr/include/r_parse.h
```diff
@@ -32,7 +32,7 @@ typedef struct r_parse_t {
 	// RAnal *anal; // weak anal ref XXX do not use. use analb.anal
 	RList *parsers;
 	RAnalVarList varlist;
-	int (*get_ptr_at)(void *user, RAnalFunction *fcn, int delta, ut64 addr);
+	st64 (*get_ptr_at)(void *user, RAnalFunction *fcn, st64 delta, ut64 addr);
```
radare, line 35: I assume the delta is stack relative.. and i wont expect a 63bit delta for a stackframe

PR #16640, opened 2020-04-19
just tests for r_arch

ret2libc:

Issue #16400, opened 2020-04-02
ESIL suggestions for improval


XVilka:
- [ ] Separate `R_API r_anal_esil_*` into `libr/include/r_esil.h`
- [ ] Consider to isolate the ESIL handling into the separate library, like siol was
- [ ] Provide the better API to distinct between various ESIL operation kinds
- [ ] Provide the API for "un-RPN"-ing of the ESIL expressions, for printing purpose, maybe more readable representation, useful for debugging/etc
- [ ] Provide the way for easier distinction between memory accesses and register accesses in API

condret:
I thought about this ... the problem is that esil depends on anal-plugins, which is why esil is kinda related to that. you get my point?

PR #16643, opened 2020-04-19
WIP: Initial implementation of the r_arch unified API with the bf plugin ##asm

radare:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


libr/arch/p/arch_bf.c
```diff
@@ -0,0 +1,343 @@
...
+			bool tried = false;
+			len--;
+			while (i < len && *p) {
+				if (*p == '[') {
+					lev++;
+				}
+				if (*p == ']') {
+					lev--;
+					if (lev == -1) {
+						ut64 dst = ins->addr + i + 1;
```
XVilka, line 60: Should be extracted into the separate function, imho.

libr/arch/p/arch_bf.c
```diff
@@ -0,0 +1,343 @@
...
+		break;
+	case ']':
+		buf_asm = "loop";
+		ins->type = R_ANAL_OP_TYPE_UJMP;
+		// XXX This ESIL experssion should walk backward until the matching [ is found
+		if (has (ESIL)) {
+			r_strbuf_set (&ins->esil, "brk,--=,brk,[1],pc,=");
+		}
+		if (true) { // has (ANAL)) {
+			if (a->arch && a->cbs->read_at) {
```
XVilka, line 106: Same here - please extract to the separate function.

libr/core/cmd.c
```diff
@@ -32,9 +32,9 @@
 //       whitespaces, because we let cmd_substitution_arg create
 //       new arguments
 static const char *SPECIAL_CHARS_REGULAR = "@;~$#|`\"'()<>";
-static const char *SPECIAL_CHARS_PF = "@;~$#|`\"'<>";
 
 #if USE_TREESITTER
+static const char *SPECIAL_CHARS_PF = "@;~$#|`\"'<>";
```
XVilka, line 37: There is no need to move it here, or move `SPECIAL_CHARS_REGULAR` too.

libr/core/libs.c
```diff
@@ -102,6 +103,7 @@ R_API void r_core_loadlibs_init(RCore *core) {
 	DF (LANG, "language plugins", lang);
 	DF (ANAL, "analysis plugins", anal);
 	DF (ASM, "(dis)assembler plugins", asm);
+	DF (ARCH, "unified architecture plugins", arch);
```
XVilka, line 106: No need for "unified", just "architecture plugins.

libr/include/r_arch.h
```diff
@@ -0,0 +1,157 @@
...
+R_API const char *r_arch_instruction_get_esil(RArchInstruction *ins);
+R_API const char *r_arch_instruction_get_esil(RArchInstruction *ins);
+R_API const char *r_arch_instruction_get_string(RArchInstruction *ins);
+R_API const char *r_arch_instruction_get_string(RArchInstruction *ins);
+R_API void r_arch_free(RArch *arch);
+R_API void r_arch_info_free(RArchInfo *info);
+R_API void r_arch_instruction_fini(RArchInstruction *ins);
+R_API void r_arch_instruction_free(RArchInstruction *ins);
+R_API void r_arch_instruction_init(RArchInstruction *ins);
+R_API void r_arch_instruction_init_code(RArchInstruction *ins, ut64 addr, const char *opstr);
```
ret2libc, line 145: are these `init_code`, `init_data` still used?

libr/include/r_arch.h
```diff
@@ -0,0 +1,157 @@
...
+R_API RArch *r_arch_new(void);
+R_API RArchInfo *r_arch_info_new(void);
+R_API RArchInstruction *r_arch_instruction_new();
+R_API RArchPlugin *r_arch_get_plugin(RArch *a, const char *name);
+R_API RArchSession *r_arch_session_new(RArch *a, RArchPlugin *ap, RArchSetup *setup);
+R_API bool r_arch_add(RArch *a, RArchPlugin *foo);
+R_API bool r_arch_del(RArch *a, RArchPlugin *ap);
+R_API bool r_arch_session_can_decode(RArchSession *ai);
+R_API bool r_arch_session_can_encode(RArchSession *ai);
+R_API bool r_arch_session_decode(RArchSession *ai, RArchInstruction *ins, RArchOptions opt);
```
ret2libc, line 130: this API still remains quite unclear... I think we agreed on having a RArchInputOptions and RArchOutputOptions to make it clearer. I tried something like that in https://github.com/radareorg/radare2/pull/16640/files#diff-3e056affb549bf222a411096b3492212R109 . WDYT?

libr/include/r_arch.h
```diff
@@ -0,0 +1,157 @@
...
+
+R_API RArch *r_arch_new(void);
+R_API RArchInfo *r_arch_info_new(void);
+R_API RArchInstruction *r_arch_instruction_new();
+R_API RArchPlugin *r_arch_get_plugin(RArch *a, const char *name);
+R_API RArchSession *r_arch_session_new(RArch *a, RArchPlugin *ap, RArchSetup *setup);
+R_API bool r_arch_add(RArch *a, RArchPlugin *foo);
+R_API bool r_arch_del(RArch *a, RArchPlugin *ap);
+R_API bool r_arch_session_can_decode(RArchSession *ai);
+R_API bool r_arch_session_can_encode(RArchSession *ai);
```
ret2libc, line 129: Whether you can encode/decode something depends also on the Options you provide to encode/decode/xxcode. It is not said that every plugin will be able to handle ESIL/OPAX/etc..

libr/include/r_arch.h
```diff
@@ -0,0 +1,157 @@
...
+	R_REF_TYPE;
+} RArchSession;
+
+
+typedef bool (*RArchPluginCallback)(RArch *a);
+typedef bool (*RArchSessionCallback)(RArchSession *a);
+typedef bool (*RArchEncodeCallback)(RArchSession *a, RArchInstruction *ins, RArchOptions options);
+typedef bool (*RArchDecodeCallback)(RArchSession *a, RArchInstruction *ins, RArchOptions options);
+
+R_API RArchSession *r_arch_session_new(RArch *a, RArchPluginP ap, RArchSetup *setup);
```
ret2libc, line 87: dupped methods. remove.

libr/core/core.c
```diff
@@ -2624,6 +2630,16 @@ R_API bool r_core_init(RCore *core) {
 	core->lang->cb_printf = r_cons_printf;
 	r_lang_define (core->lang, "RCore", "core", core);
 	r_lang_set_user_ptr (core->lang, core);
+	{
+		core->arch = r_arch_new ();
+		// just move this into RCore
+		RArchCallbacks *cbs = &core->arch->cbs;
```
ret2libc, line 2636: :+1:

libr/core/cmd.c
ret2libc: this was because when you compile without tree-sitter, compiler complains about SPECIAL_CHARS_PF not being used. SPECIAL_CHARS_REGULAR, instead, is used even without tree-sitter AFAIK. But please @radare , don't do this change here. I will commit a separate PR for this.

libr/core/cconfig.c
```diff
@@ -33,6 +33,37 @@ static void set_options(RConfigNode *node, ...) {
...
+		RArchSetup setup = {
+			.bits = r_config_get_i (c->config, "asm.bits"),
+			.cpu = strdup (r_config_get (c->config, "asm.cpu")),
+			.endian = r_config_get_i (c->config, "cfg.bigendian")? R_SYS_ENDIAN_BIG: R_SYS_ENDIAN_LITTLE,
+			.syntax = r_config_get_i (c->config, "asm.syntax"),
+		};
+		RArchSession *as = r_arch_session_new (a, ap, &setup);
+		if (as) {
+			r_arch_session_unref (c->assembler->asd);
+			c->assembler->asd = as;
```
ret2libc, line 49: it took me a while to understand what `asd` meant. I thought it was random name :P

libr/core/cconfig.c
```diff
@@ -33,6 +33,37 @@ static void set_options(RConfigNode *node, ...) {
...
+			r_arch_session_ref (c->assembler->asd);
+			//
+			r_arch_session_unref (c->assembler->asd);
```
ret2libc, line 52: I guess this should be `asa`, not `asd`.

libr/core/cconfig.c
```diff
@@ -33,6 +33,37 @@ static void set_options(RConfigNode *node, ...) {
...
+			c->assembler->asa = as;
+			r_arch_session_ref (c->assembler->asd);
```
ret2libc, line 54: also this.

libr/asm/asm.c
```diff
@@ -575,11 +611,20 @@ R_API int r_asm_assemble(RAsm *a, RAsmOp *op, const char *buf) {
...
-		r_parse_parse (a->ifilter, buf, b);
+		r_parse_parse (a->ifilter, b, b);
 	}
 	r_str_case (b, 0); // to-lower
 	memset (op, 0, sizeof (RAsmOp));
-	if (a->cur) {
+	if (a->asa && r_arch_session_can_encode (a->asa)) {
+		/* RArchInstruction -> RAsmOp */
+		RArchInstruction ins = {0};
+		r_strbuf_setf (&ins.code, buf);
```
ret2libc, line 621: `r_arch_instruction_init_code`

libr/arch/arch_session.c
```diff
@@ -0,0 +1,152 @@
...
+	free (as);
+}
+
+R_API RArchSession *r_arch_session_new(RArch *a, RArchPlugin *ap, RArchSetup *setup) {
+	r_return_val_if_fail (a && ap, NULL);
+	RArchSession *ai = R_NEW0 (RArchSession);
+	if (ai) {
+		ai->cur = ap;
+		if (!setup) {
+			RArchSetup _setup = {
```
ret2libc, line 106: as said in previous PR, I think each plugin should provide a `default_setup` with a callback. See https://github.com/radareorg/radare2/pull/16640/files#diff-6923b12d4ceb725eb8d67413f6807fafR4 for an example. I think it makes sense for each RArchPlugin to have its own default, because R_SYS_ENDIAN and R_SYS_BITS may not make sense for everything and this way when you call `r_arch_session_new (a, plugin, NULL)` you automatically get the default setup (e.g. 64bit mode/little endian for x86_arch_plugin, etc.)

libr/arch/arch_session.c
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return false;
+}
+
+R_API void r_arch_session_free(RArchSession *as) {
+	// something more
+	free (as);
+}
+
+R_API RArchSession *r_arch_session_new(RArch *a, RArchPlugin *ap, RArchSetup *setup) {
```
ret2libc, line 100: shouldn't it call `r_arch_session_ref` before returning? Otherwise you may have return a reference that is already ready to be freed. In general, I think, `_new` should set refcount to 1. See https://github.com/radareorg/radare2/pull/16640/files#diff-6923b12d4ceb725eb8d67413f6807fafR106 for example. That test case, IMHO, should work. And it uses `r_arch_session_unref` at the end.

libr/arch/arch_session.c
```diff
@@ -0,0 +1,152 @@
...
+	default:
+		return false;
+	}
+	return false;
+}
+
+static bool r_arch_session_set_endian(RArchSession *ai, RArchEndian endian) {
+	r_return_val_if_fail (ai && ai->cur, false);
+	switch (endian) {
+	case R_SYS_ENDIAN_LITTLE:
```
ret2libc, line 43: shouldn't we have `R_ARCH_ENDIAN_LITTLE`, etc?

libr/arch/arch_session.c
```diff
@@ -0,0 +1,152 @@
...
+		}
+		return true;
+	}
+	return false;
+}
+
+static bool r_arch_session_set_syntax(RArchSession *ai, RArchSyntax syntax) {
+	r_return_val_if_fail (ai, false);
+	switch (syntax) {
+	case R_ASM_SYNTAX_REGNUM:
```
ret2libc, line 24: shouldn't we have `R_ARCH_SYNTAX_*`?

libr/arch/arch.c
```diff
@@ -0,0 +1,80 @@
...
+	r_return_val_if_fail (a, false);
+	if (ap->fini) {
+		ap->fini (a);
+	}
+	return r_list_delete_data (a->plugins, ap);
+}
+
+R_API RArchPlugin *r_arch_get_plugin(RArch *a, const char *name) {
+	r_return_val_if_fail (a && a->plugins, false);
+        RArchPlugin *h;
```
ret2libc, line 59: indentation.

libr/anal/vtable.c
```diff
@@ -111,13 +111,14 @@ static int vtable_is_addr_vtable_start(RVTableContext *context, ut64 curAddress)
 		// section in which currenct xref lies
 		if (vtable_addr_in_text_section (context, xref->addr)) {
 			ut8 buf[VTABLE_BUFF_SIZE];
-			int res = context->anal->iob.read_at (context->anal->iob.io, xref->addr, buf, sizeof(buf));
+			if (!context->anal->iob.read_at (context->anal->iob.io, xref->addr, buf, sizeof (buf))) {
```
ret2libc, line 114: unrelated change actually...

libr/anal/reflines.c
```diff
@@ -1,4 +1,4 @@
-/* radare - LGPL - Copyright 2009-2019 - pancake, nibble */
+/* radare - LGPL - Copyright 2009-2020 - pancake, nibble */
```
ret2libc, line 1: unrelated changes as well.

libr/anal/op.c
```diff
@@ -86,12 +86,70 @@ static int defaultCycles(RAnalOp *op) {
...
+	int r = R_ARCH_OPTION_SIZE; // R_ANAL_OP_MASK_BASIC
+	r |= R_ARCH_OPTION_ANAL;
+	if (m & R_ANAL_OP_MASK_ESIL) {
+		r |= R_ARCH_OPTION_ESIL;
+	}
+	if (m & R_ANAL_OP_MASK_DISASM) {
+		r |= R_ARCH_OPTION_CODE;
+	}
+	return r;
+#if 0
```
ret2libc, line 99: remove

ret2libc:
I think an important thing to discuss is how we handle multiple assembler/disassembler backends for the same "architecture" (e.g. x86 has many disassemblers and assemblers). Now, I think you said on Telegram that you want to keep them split (correct me if I got it wrong).

I believe that if we do it so, we lose a lot of the potential of RArch, because the whole point of RArch is that you are easily able to move from one form to another (e.g. code->data, data->code, data->esil, code->esil, code->description, data->description, etc.). These transformations, I believe, can be achieved only if you have both assembler and disassembler in the same Plugin, otherwise there would be tons of duplicated code (if you end up implementing all operations).

For this reason, I believe we should think about supporting multiple assembler/disassembler backends in the same RArchPlugin, otherwise we won't achieve our purpose of having multiple kinds of transformations, but we will still have just assemblers and disassemblers (separated).

test/unit/test_arch.c
```diff
@@ -0,0 +1,107 @@
...
+	bool res;
+	RArch *a = r_arch_new ();
+	RArchPlugin *ap = r_arch_get_plugin (a, "bf");
+	if (ap) {
+		RArchSession *as = r_arch_session_new (a, ap, NULL);
+		r_arch_instruction_init (&ins);
+
+		r_arch_session_encode_instruction (as, &ins, 0x8080, "nop");
+		mu_assert ("r_arch_session_encode", ins.data.len == 1);
+		if (r_arch_session_decode_esil (as, &ins, 0x8080, (const ut8*)"+++", 3)) {
```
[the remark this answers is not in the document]
XVilka: There is no human representation of EZIL yet, see https://github.com/radareorg/radare2/issues/16400 for this

XVilka:
The thing is - for some architectures having a separate assembler/disassembler might be a good thing, like we currently have with x86 or ARM. But for many having them in one plugin might be good. Moreover, some parts of the assembler parsing can be reused by many assemblers - e.g. most RISC assemblers looks similar in the text form.

ret2libc:
> The thing is - for some architectures having a separate assembler/disassembler might be a good thing, like we currently have with x86 or ARM. 

Why? I would love for my x86 RArchPlugin to be able to handle code->data, data->code, data->esil, code->esil, data->description, code->description and you can do all these operations only if your RArchPlugin is able to disassemble AND assemble.

XVilka:
@ret2libc because we have multiple x86 and ARM assemblers - e.g. one built-in, and a couple wrappers around Nasm, etc.

ret2libc:
> @ret2libc because we have multiple x86 and ARM assemblers - e.g. one built-in, and a couple wrappers around Nasm, etc.

That's what I am saying. What about handling them in the same RArchPlugin? After all, the description is the same, no matter if you use nasm, as, cs, hand-made. ESIL is the same, no matter if you use asm, nasm, cs, hand-made. Even analysis is probably the same. If you have them separate, you end up with a lot of dupped code and probably you lose the ability of doing multiple conversions (as I said, code->data, data->code, data->esil, code->esil, data->description, code->description) which after all is one of the strongest point in this new RArch AFAIU.

libr/arch/p/arch_bf.c
radare: out of the scope of this pr imho, ive already improved and cleanedup the bf plugin to make it work as an arch plugin here, but i dont want to make this PR bigger focusing in the unrelated changes to the core logic of the new api which is the main purpose of this pr

libr/core/cmd.c
radare: yes, thats because its unused witohut treesitter. pls commit this as a separate PR, otherwise ill do it today , as well as ill do with the rest of unrelated/random changes in this pr

shlr/grub/fs/reiserfs.c
```diff
@@ -59,7 +59,7 @@
 
 static grub_dl_t my_mod;
 
-#define assert(boolean) real_assert (boolean, GRUB_FILE, __LINE__)
```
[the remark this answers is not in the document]
radare: because warnings

pkgcfg/r_arch.pc.acr
```diff
@@ -0,0 +1,11 @@
+prefix=@PREFIX@
+exec_prefix=${prefix}
+libdir=@LIBDIR@
+includedir=${prefix}/include
+
+Name: r_arch
+Description: radare foundation libraries
+Version: @VERSION@
+Requires: r_util r_parse r_flag r_cons r_reg
```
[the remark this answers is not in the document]
radare: `make pkgcfg` fixed this

libr/util/strbuf.c
```diff
@@ -119,6 +120,17 @@ R_API bool r_strbuf_slice(RStrBuf *sb, int from, int len) {
 	return true;
 }
 
+R_API bool r_strbuf_setptr(RStrBuf *sb, char *s, int len) {
```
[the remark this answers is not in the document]
radare: will do today. ill just use the slow setbin in this PR until rebased to get the setptr api ready

libr/arch/arch_session.c
radare: yes, as well as many other constants from r_util, r_asm and r_anal, that should be rebranded underr the r_arch umbrella. but imho we shuold do this change later, otherwise this PR can grow too much

libr/arch/arch.c
radare: fixed

libr/anal/reflines.c
radare: not completely unrelated, the brainfuck plugin now handles this new reflines feature, so this code was updated to show it

libr/anal/op.c
radare: this is totally WIP because the usage of this api is wrong in many places

libr/include/r_arch.h
thestr4ng3r: Yeah I thought we agreed on xxcode too.

ret2libc:
Closing this as there is already https://github.com/radareorg/radare2/pull/16669

Issue #16590, opened 2020-04-14
Cannot open large files


radare:
# Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Debian arm 64, Ubuntu x86 32
| File format of the file you reverse (mandatory)      | PE, ELF etc.
| Architecture/bits of the file (mandatory)            | PPC, x86/32, x86/64 etc.
| r2 -v full output, **not truncated** (mandatory)     | radare2 2.4.0-git 17284 @ darwin-x86-64 git.2.2.0-476-gf8cf84e06 commit: f8cf84e0653642d9ad34e760e0e56dd81860e799 build: 2018-02-17__11:08:27

### Expected behavior

r2 large.file
r2 mmap://large.file
r2 posix://large.file

### Actual behavior

Right now its quite hidden whats the actual backend, mmap is faster, but wont work for mapping HUGE files, in the test.

```
r2 mmap://recorder_image.raw
[r] Cannot open 'mmap://recorder_image.raw'
```

### Steps to reproduce the behavior 
* buy an rpi
* install debian
* install r2
* generate an 8GB file
* open it with r2

### Additional Logs, screenshots, source-code,  configuration dump, ...

cc @shaddy

Shaddy:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Debian arm 32, Raspbian GNU/Linux 10 (buster)
| File format of the file you reverse (mandatory)      | N/A
| Architecture/bits of the file (mandatory)            | N/A (8GB File)
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.4.0 24351 @ linux-arm-32 git.4.4.0-2-g08210f3d0 commit: 08210f3d0eefaebde8134ddf3d33599c47694b99 build: 2020-04-14__14:23:53


* 4GB of RAM

```
shaddy@rasp01:~/analysis $ cat /proc/meminfo
MemTotal:        3999756 kB
MemFree:          183528 kB
MemAvailable:    3503348 kB
```

* Commands output

```
shaddy@rasp01:~/analysis $ r2 posix://recorder_image.raw
[r] Cannot open 'posix://recorder_image.raw'

shaddy@rasp01:~/analysis $ r2 recorder_image.raw
[r] Cannot open 'recorder_image.raw'

[...]

PR #16648, opened 2020-04-20
Fix regression loading large files (>2GB) on 32bit systems ##io

radare:
This PR does nothing, just aesthetics for now

 

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -3,10 +3,6 @@
 #ifndef R2_IO_H
 #define R2_IO_H
 
-#define _GNU_SOURCE 1
-#define _LARGEFILE64_SOURCE 1
-#define _FILE_OFFSET_BITS 64
-
 #include "r_list.h"
 #include <r_util.h>
 #include "r_socket.h"
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -3,6 +3,10 @@
 #ifndef R2_IO_H
 #define R2_IO_H
 
+#define _GNU_SOURCE 1
+#define _LARGEFILE64_SOURCE 1
+#define _FILE_OFFSET_BITS 64
+
 #include "r_list.h"
 #include <r_util.h>
 #include "r_socket.h"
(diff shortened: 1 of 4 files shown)


ret2libc:
this is super WIP, right? :)

radare:
Yes, no need to review this pr , im waiting for feedback from the friend who foundnu this issue i didnt had time to review/check but the issue is there, and after reviewing the code i found no issues, but just few aesthetical changes. Ignore this pr until i remove the WIP bit

radare:
Ref https://github.com/radareorg/radare2/issues/16590

radare:
i think the problem is a regression. and should be fixed by bring back this https://users.suse.com/~aj/linux_lfs.html

radare:
this is imho no longer a draft, this PR should be good for testing and then merge when fix is confirmed

radare:
should be fixed now,, wonder when this regression got introduced. maybe after siol?

libr/include/r_io.h
```diff
@@ -3,6 +3,9 @@
 #ifndef R2_IO_H
 #define R2_IO_H
 
+#define _GNU_SOURCE 1
```
ret2libc, line 6: This produces a lot of warnings while compiling, see https://github.com/radareorg/radare2/pull/16648/checks?check_run_id=622122422#step:7:364

Shaddy:
The only way got this working was defining it at `libr/util/sandbox.c`

`#define _FILE_OFFSET_BITS 64`

Unfortunately seems that io_default.c/r_io.h does not apply for sandbox.c as it is not using those headers, and open seems is not getting the definition.

libr/include/r_io.h
ret2libc: This is still no good, it produces a lot of warnings.
Maybe do:
```C
#ifndef _GNU_SOURCE
#define _GNU_SOURCE 1
#endif
```

ret2libc:
@radare just make sure _GNU_SOURCE is not creating 123231 warnings, then since the issue is fixed according to https://github.com/radareorg/radare2/issues/16590#issuecomment-620262094 , we can probably merge.

Shaddy:
@radare I've tested it and works perfect

radare:
YAY

PR #16649, opened 2020-04-20
Implement r_strbuf_setptr API ##util

radare:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/include/r_util/r_strbuf.h
+++ b/libr/include/r_util/r_strbuf.h
@@ -7,7 +7,7 @@ extern "C" {
 
 typedef struct {
 	char buf[32];
-	size_t len; // string length in chars
+	size_t len; // string length in chars or binary buffer size
 	char *ptr; // ptr replacing buf in case strlen > sizeof(buf)
 	size_t ptrlen; // string length + 1 or binary buffer size
 	bool weakref; // ptr is not owned
@@ -17,12 +17,12 @@ typedef struct {
 R_API RStrBuf *r_strbuf_new(const char *s);
 R_API bool r_strbuf_set(RStrBuf *sb, const char *s);
 R_API bool r_strbuf_slice(RStrBuf *sb, int from, int len);
-R_API bool r_strbuf_setbin(RStrBuf *sb, const ut8 *s, int len);
+R_API bool r_strbuf_setbin(RStrBuf *sb, const ut8 *s, size_t len);
 R_API ut8* r_strbuf_getbin(RStrBuf *sb, int *len);
 R_API bool r_strbuf_setf(RStrBuf *sb, const char *fmt, ...);
 R_API bool r_strbuf_vsetf(RStrBuf *sb, const char *fmt, va_list ap);
 R_API bool r_strbuf_append(RStrBuf *sb, const char *s);
-R_API bool r_strbuf_append_n(RStrBuf *sb, const char *s, int l);
+R_API bool r_strbuf_append_n(RStrBuf *sb, const char *s, size_t l);
 R_API bool r_strbuf_prepend(RStrBuf *sb, const char *s);
 R_API bool r_strbuf_appendf(RStrBuf *sb, const char *fmt, ...);
 R_API bool r_strbuf_vappendf(RStrBuf *sb, const char *fmt, va_list ap);
--- a/libr/include/r_util/r_strbuf.h
+++ b/libr/include/r_util/r_strbuf.h
@@ -36,7 +36,7 @@ R_API void r_strbuf_fini(RStrBuf *sb);
 R_API void r_strbuf_init(RStrBuf *sb);
 R_API bool r_strbuf_copy(RStrBuf *dst, RStrBuf *src);
 R_API bool r_strbuf_equals(RStrBuf *sa, RStrBuf *sb);
-R_API bool r_strbuf_reserve(RStrBuf *sb, int len);
+R_API bool r_strbuf_reserve(RStrBuf *sb, size_t len);
 R_API bool r_strbuf_is_empty(RStrBuf *sb);
 R_API bool r_strbuf_setptr(RStrBuf *sb, char *p, int l);
 
--- a/libr/include/r_util/r_strbuf.h
+++ b/libr/include/r_util/r_strbuf.h
@@ -7,10 +7,10 @@ extern "C" {
 
 typedef struct {
 	char buf[32];
-	size_t len;
-	char *ptr;
-	size_t ptrlen;
-	bool weakref;
+	size_t len; // string length in chars
+	char *ptr; // ptr replacing buf in case strlen > sizeof(buf)
+	size_t ptrlen; // string length + 1 or binary buffer size
+	bool weakref; // ptr is not owned
 } RStrBuf;
 
 #define R_STRBUF_SAFEGET(sb) (r_strbuf_get (sb) ? r_strbuf_get (sb) : "")
@@ -30,6 +30,7 @@ R_API char *r_strbuf_get(RStrBuf *sb);
 R_API char *r_strbuf_drain(RStrBuf *sb);
 R_API char *r_strbuf_drain_nofree(RStrBuf *sb);
 R_API int r_strbuf_length(RStrBuf *sb);
+R_API int r_strbuf_size(RStrBuf *sb);
 R_API void r_strbuf_free(RStrBuf *sb);
 R_API void r_strbuf_fini(RStrBuf *sb);
 R_API void r_strbuf_init(RStrBuf *sb);
--- a/libr/include/r_util/r_strbuf.h
+++ b/libr/include/r_util/r_strbuf.h
@@ -10,7 +10,7 @@ typedef struct {
 	size_t len;
 	char *ptr;
 	size_t ptrlen;
-	bool ro;
+	bool weakref;
 } RStrBuf;
 
 #define R_STRBUF_SAFEGET(sb) (r_strbuf_get (sb) ? r_strbuf_get (sb) : "")
--- a/libr/include/r_util/r_strbuf.h
+++ b/libr/include/r_util/r_strbuf.h
@@ -7,9 +7,10 @@ extern "C" {
 
 typedef struct {
 	char buf[32];
-	int len;
+	size_t len;
 	char *ptr;
-	int ptrlen;
+	size_t ptrlen;
+	bool ro;
 } RStrBuf;
 
 #define R_STRBUF_SAFEGET(sb) (r_strbuf_get (sb) ? r_strbuf_get (sb) : "")
--- a/libr/include/r_util/r_strbuf.h
+++ b/libr/include/r_util/r_strbuf.h
@@ -36,6 +36,7 @@ R_API bool r_strbuf_copy(RStrBuf *dst, RStrBuf *src);
 R_API bool r_strbuf_equals(RStrBuf *sa, RStrBuf *sb);
 R_API bool r_strbuf_reserve(RStrBuf *sb, int len);
 R_API bool r_strbuf_is_empty(RStrBuf *sb);
+R_API bool r_strbuf_setptr(RStrBuf *sb, char *p, int l);
 
 #ifdef __cplusplus
 }
--- a/libr/util/strbuf.c
+++ b/libr/util/strbuf.c
@@ -53,8 +53,8 @@ R_API bool r_strbuf_copy(RStrBuf *dst, RStrBuf *src) {
 	return true;
 }
 
-R_API bool r_strbuf_reserve(RStrBuf *sb, int len) {
-	r_return_val_if_fail (sb && len > 0, false);
+R_API bool r_strbuf_reserve(RStrBuf *sb, size_t len) {
+	r_return_val_if_fail (sb, false);
 
 	if ((sb->ptr && len < sb->ptrlen) || (!sb->ptr && len < sizeof (sb->buf))) {
 		return true;
--- a/libr/util/strbuf.c
+++ b/libr/util/strbuf.c
@@ -73,7 +73,9 @@ R_API bool r_strbuf_reserve(RStrBuf *sb, int len) {
 
 R_API bool r_strbuf_setbin(RStrBuf *sb, const ut8 *s, int l) {
 	r_return_val_if_fail (sb && s, false);
-	r_return_val_if_fail (l >= 0, false);
+	if (l < 0) {
+		l = strlen ((const char *)s) + 1;
+	}
 
 	if (l >= sizeof (sb->buf)) {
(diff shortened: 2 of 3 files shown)


libr/include/r_util/r_strbuf.h
```diff
@@ -7,9 +7,10 @@ extern "C" {
 
 typedef struct {
 	char buf[32];
-	int len;
+	size_t len;
```
thestr4ng3r, line 10: ❤️

libr/include/r_util/r_strbuf.h
```diff
@@ -7,9 +7,10 @@ extern "C" {
...
 	char *ptr;
-	int ptrlen;
+	size_t ptrlen;
+	bool ro;
```
thestr4ng3r, line 13: Maybe add a comment here what ro means (weak vs. strong reference).

libr/util/strbuf.c
```diff
@@ -293,16 +311,25 @@ R_API ut8 *r_strbuf_getbin(RStrBuf *sb, int *len) {
 	return (ut8 *)(sb->ptr ? sb->ptr : sb->buf);
 }
 
+// TODO: expose aa r_strbuf_getdup() ?
+static inline char *get_strdup(RStrBuf *sb) {
+	return sb->ptr
+		? sb->ro
+			? r_mem_dup (sb->ptr, sb->len)
```
thestr4ng3r, line 318: If the weak pointer is a null-terminated string then you would get a non-null terminated one from here, right? Probably better to use ptrlen.

libr/util/strbuf.c
```diff
@@ -119,6 +120,17 @@ R_API bool r_strbuf_slice(RStrBuf *sb, int from, int len) {
 	return true;
 }
 
+R_API bool r_strbuf_setptr(RStrBuf *sb, char *s, int len) {
+	r_return_val_if_fail (sb, false);
+	if (len < 0) {
+		len = strlen (s);
+	}
+	sb->ptr = s;
+	sb->ptrlen = len;
```
thestr4ng3r, line 129: Something like this maybe?
```
sb->ptrlen = len < 0 ? strlen (s) + 1 : len;
```
Though then we have to be careful not to do something like `r_strbuf_setptr(sb, "fuck", 4)` but to always pass -1 if it's supposed to be a string vs. bytes.

libr/include/r_util/r_strbuf.h
radare: maybe better to call it `weak_ref` and avoid the comment

libr/util/strbuf.c
radare: yeah sgtm

radare: but y our example is wrong, because the string is 5 char length . you missed the \0, thats why the strlen() is there to avoid this common human problem

thestr4ng3r: This is exactly what I am saying. Most functions that take args like `char *str, int len` where you can set len to -1 expect len to be either -1 or `strlen(str)` which would be 4 here. But here `r_strbuf_setptr(sb, "fuck", 4)` is wrong so you have to pass -1 (or 5) and this has to be explicitly specified/documented.

libr/include/r_util/r_strbuf.h
thestr4ng3r: ok too

libr/util/strbuf.c
radare: weak have nothing to do with null or non null terminated strings. thats completely unrelated.

for simplicity we can nullterminate all the things

libr/util/strbuf.c
radare: well, the idea is that string functions may get no length at all, and bin ones must specify that

libr/include/r_util/r_strbuf.h
radare: done

radare:
Ill continue in r-arch after this pr gets merged

libr/util/strbuf.c
thestr4ng3r: That's what I am saying.

libr/util/strbuf.c
thestr4ng3r: Yes it has something to do with it here. If you currently do
```
r_strbuf_setptr(sb, "fuck", -1);
char *s = r_strbuf_drain(sb);
```
then s will not be null-terminated.

thestr4ng3r:
Please add some unit tests for this stuff. Also so we can run them with asan/valgrind.

radare:
Yeah this line needs a +1

libr/util/strbuf.c
```diff
@@ -293,29 +312,40 @@ R_API ut8 *r_strbuf_getbin(RStrBuf *sb, int *len) {
 	return (ut8 *)(sb->ptr ? sb->ptr : sb->buf);
 }
 
+// TODO: expose aa r_strbuf_getdup() ?
+static inline char *get_strdup(RStrBuf *sb) {
+	return sb->ptr
```
ret2libc, line 317: please just use `if` instead of nested ternary operators :/

test/unit/test_strbuf.c
```diff
@@ -11,6 +11,45 @@ bool test_r_strbuf_slice(void) {
 	mu_end;
 }
 
+bool test_r_strbuf_append(void) {
```
ret2libc, line 14: you forgot to add this to the `all_tests`

test/unit/test_strbuf.c
```diff
@@ -11,6 +11,45 @@ bool test_r_strbuf_slice(void) {
...
+
+	mu_end;
+}
+
+bool test_r_strbuf_ptr(void) {
+	RStrBuf *sa = r_strbuf_new ("");
+	RStrBuf *sb = r_strbuf_new (NULL);
+
+	r_strbuf_set (sa, "food");
+	mu_assert_eq (r_strbuf_length (sa), 5, "r_strbuf_set:food");
```
ret2libc, line 30: this should be 4, shouldn't it? the string is 4 chars long.

test/unit/test_strbuf.c
```diff
@@ -11,6 +11,45 @@ bool test_r_strbuf_slice(void) {
...
+	r_strbuf_set (sa, "VERYLONGTEXTTHATDOESNTFITINSIDETHESTRUCTBUFFER");
+	char *a = sa->ptr;
+	mu_assert_eq (r_strbuf_length (sa), 47, "r_strbuf_set:food");
```
ret2libc, line 33: 46 instead of 47

test/unit/test_strbuf.c
```diff
@@ -11,6 +11,45 @@ bool test_r_strbuf_slice(void) {
...
+	char *b = sa->ptr;
+	mu_assert_eq (a, b, "sa->ptr after setbin");
```
ret2libc, line 35: you called `set`, not `setbin`.

libr/util/strbuf.c
```diff
@@ -73,7 +73,14 @@ R_API bool r_strbuf_reserve(RStrBuf *sb, int len) {
 
 R_API bool r_strbuf_setbin(RStrBuf *sb, const ut8 *s, int l) {
 	r_return_val_if_fail (sb && s, false);
-	r_return_val_if_fail (l >= 0, false);
+	if (l < 0) {
```
ret2libc, line 76: why do we want to support this thing? If you are using `r_strbuf_setbin`, IMHO, you *must* pass a valid size as well. -1 would assume that `s` is a string, but if you know that you may as well directly call `r_strbuf_set` instead of this.

So, IMHO this can be simplified by making `l` a `size_t` and assuming `l` is always valid.

libr/util/strbuf.c
```diff
@@ -71,10 +71,10 @@ R_API bool r_strbuf_reserve(RStrBuf *sb, int len) {
 	return true;
 }
 
-R_API bool r_strbuf_setbin(RStrBuf *sb, const ut8 *s, int l) {
+R_API bool r_strbuf_setbin(RStrBuf *sb, const ut8 *s, size_t l) {
 	r_return_val_if_fail (sb && s, false);
-	r_return_val_if_fail (l >= 0, false);
-
+	sb->len = l;
+	sb->ptrlen = l;
```
ret2libc, line 77: mmm this smells odd. 
1) both len and ptrlen are set again before returning from this function
2) sb->ptrlen is used at line 80, but the condition `l + 1 > sb->ptrlen` would always be true with this.

libr/util/strbuf.c
ret2libc: @thestr4ng3r see this pls

libr/util/strbuf.c
```diff
@@ -293,29 +319,39 @@ R_API ut8 *r_strbuf_getbin(RStrBuf *sb, int *len) {
...
+		? sb->weakref
+			? r_mem_dup (sb->ptr, sb->ptrlen)
+			: sb->ptr
+		: strdup (sb->buf);
+}
+
 R_API char *r_strbuf_drain(RStrBuf *sb) {
 	r_return_val_if_fail (sb, NULL);
-	char *ret = sb->ptr ? sb->ptr : strdup (sb->buf);
+	char *ret = drain (sb);
```
ret2libc, line 332: can't we just return sb->ptr in case of weakref?

thestr4ng3r: No, drain should always give you a heap-allocated string that you own after the return. At least that is my understanding of it. Most of the time you don't want to drain a weak strbuf, but I think it makes more sense to always pass the ownership to the caller instead of only in some cases.

libr/util/strbuf.c
thestr4ng3r: Yes you are right, this looks broken.

libr/util/strbuf.c
```diff
@@ -119,14 +121,32 @@ R_API bool r_strbuf_slice(RStrBuf *sb, int from, int len) {
 	return true;
 }
 
-R_API bool r_strbuf_set(RStrBuf *sb, const char *s) {
+R_API bool r_strbuf_setptr(RStrBuf *sb, char *s, int len) {
```
pelijah, line 124:
```suggestion
R_API bool r_strbuf_setptr(RStrBuf *sb, char *s, ssize_t len) {
```

ret2libc: why?

libr/util/strbuf.c
```diff
@@ -119,14 +121,32 @@ R_API bool r_strbuf_slice(RStrBuf *sb, int from, int len) {
...
 
+R_API bool r_strbuf_set(RStrBuf *sb, const char *s) {
+	r_return_val_if_fail (sb, false);
 	if (!s) {
 		r_strbuf_init (sb);
 		return true;
 	}
-	return r_strbuf_setbin (sb, (const ut8*)s, strlen (s));
+	size_t len = strlen (s);
+	if (!r_strbuf_setbin (sb, (const ut8*)s, len + 1)) {
```
pelijah, line 145: Why `len + 1`? 
`r_strbuf_setbin()` puts 0 at the end of buffer.
https://github.com/radareorg/radare2/blob/79fcf4fbff5cc8fb2711dd4c1e7f5521a43c531d/libr/util/strbuf.c#L90

libr/util/strbuf.c
pelijah: Why not?

pelijah: And why not even to use `size_t`?

radare:
i think that setptr can be used for substrings, but then we wont be able to expect it to be null terminated. and as long as its a weak ref you cant write on it. So i think that setptr should get const ut8*arg, size_t len, isntead of char*,int

libr/include/r_util/r_strbuf.h
```diff
@@ -7,35 +7,38 @@ extern "C" {
...
-R_API bool r_strbuf_append_n(RStrBuf *sb, const char *s, int l);
+R_API bool r_strbuf_append_n(RStrBuf *sb, const char *s, size_t l);
 R_API bool r_strbuf_prepend(RStrBuf *sb, const char *s);
 R_API bool r_strbuf_appendf(RStrBuf *sb, const char *fmt, ...);
 R_API bool r_strbuf_vappendf(RStrBuf *sb, const char *fmt, va_list ap);
 R_API char *r_strbuf_get(RStrBuf *sb);
 R_API char *r_strbuf_drain(RStrBuf *sb);
 R_API char *r_strbuf_drain_nofree(RStrBuf *sb);
 R_API int r_strbuf_length(RStrBuf *sb);
+R_API int r_strbuf_size(RStrBuf *sb);
```
radare, line 33: return size_t

libr/util/strbuf.c
radare: what about having r_strbuf_setptr(const ut8* , size_t len); and r_strbuf_setstr(const char *str); instead of abusing the length with negative values

libr/util/strbuf.c
```diff
@@ -119,14 +121,32 @@ R_API bool r_strbuf_slice(RStrBuf *sb, int from, int len) {
...
 
-R_API bool r_strbuf_set(RStrBuf *sb, const char *s) {
+R_API bool r_strbuf_setptr(RStrBuf *sb, char *s, int len) {
 	r_return_val_if_fail (sb, false);
+	if (len < 0) {
+		sb->len = strlen (s);
+		sb->ptrlen = sb->len + 1;
+	} else {
+		sb->ptrlen = len;
+		sb->len = len;
```
pelijah, line 131: What if `s[len] != '\0'`?

libr/util/strbuf.c
```diff
@@ -293,29 +319,39 @@ R_API ut8 *r_strbuf_getbin(RStrBuf *sb, int *len) {
 	return (ut8 *)(sb->ptr ? sb->ptr : sb->buf);
 }
 
+static inline char *drain(RStrBuf *sb) {
```
radare, line 322: maybe set the arg as const as long as we are not touching it at all

thestr4ng3r:
I fixed the setbin with the idea that it always allocated +1 and puts 0 at the end.

thestr4ng3r:
I fixed the setbin with the idea that it always allocated +1 and puts 0 at the end.

ret2libc:
@thestr4ng3r still marking this as waiting-for-author as there are unaddressed comments.

thestr4ng3r:
Who is the author at this point?

ret2libc:
I guess that's you :D :D Or I could create a label named `waiting-for-thestranger` :P

ret2libc:
Why merge? THere were comments not dealt with.... like `r_strbuf_length` that should return `size_t` or `r_strbuf_getbin` that should get `size_t`...

PR #16657, opened 2020-04-20
Rasm clean

radare:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/main/rasm2.c
+++ b/libr/main/rasm2.c
@@ -306,7 +306,7 @@ static int rasm_show_help(int v) {
 	return 0;
 }
 
-static int rasm_disasm(RAsmState *as, char *buf, ut64 offset, int len, int bits, int ascii, int bin, int hex) {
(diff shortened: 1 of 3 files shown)


libr/main/rasm2.c
```diff
@@ -320,16 +320,14 @@ static int rasm_disasm(RAsmState *as, char *buf, ut64 offset, int len, int bits,
 		}
 		clen = len; // XXX
 		data = (ut8 *)buf;
-	} else if (ascii) {
```
ret2libc, line 323: Mmm isn't this going to change the behaviour? Without this it will go through the `else` branch and it will do `r_hex_str2bin` which expects an hexstring, doesn't it?

radare: yes, it changes the behaviour. but this behaviour was only used for the brainfuck plugin for unnecessary reasons. because it works well without this and having an arch-speciric if in an assembler is a bad idea

PR #16658, opened 2020-04-20
Some random fixes here and there

radare:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/core/p/core_a2f.c
+++ b/libr/core/p/core_a2f.c
@@ -126,6 +126,7 @@ static ut64 analyzeStackBased(RCore *core, Sdb *db, ut64 addr, RList *delayed_co
 	bool block_end = false;
 	RStack *stack = r_stack_newf (10, free);
 	addTarget (core, stack, db, addr);
+	const ut64 maxfcnsize = 4096;
 
 	while (!r_stack_is_empty (stack)) {
 		block_end = false;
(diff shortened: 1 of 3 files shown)


libr/core/p/core_a2f.c
```diff
@@ -126,6 +126,7 @@ static ut64 analyzeStackBased(RCore *core, Sdb *db, ut64 addr, RList *delayed_co
 	bool block_end = false;
 	RStack *stack = r_stack_newf (10, free);
 	addTarget (core, stack, db, addr);
+	const ut64 maxfcnsize = 4096;
```
ret2libc, line 129: why? i mean, tests are passing already, why this random limit?

radare: because of timeout running `a2f` in some fuzzed binaries. nobody uses a2f except me, and im using it to identify bugs in the default analysis loop. what other solution would you do here to detect when analyzing malloc://4G for example? because 0x00 is  a  valid instruction and the fcn grows without control

ret2libc: i guess `af` already has this kind of limit, right? ok

radare: The limit in af depends on the basic block, not the function iirc, but yeah

PR #16660, opened 2020-04-20
Move radare2-shell-parser/tree-sitter into shlr and enable by default

ret2libc:
So, this is the big day :D

I'm moving radare2-shell-parser repository into shlr itself and from now on my own repository will cease to exist, we can keep the grammar just here.

Also, I'm copying the lib code of tree-sitter into shlr. There are tree-sitter-sync options to sync it when needed. This should also solve the problem raised by some, time ago, where they were not happy about the build system downloading stuff from the internet.

As part of this, I'm an enabling the compilation of the new parser by default (but it will be disabled unless you use `cfg.newshell=true`) since it is now ready enough to be played with.

TODO:
- [x] add README file in shlr/radare2-shell-parser to explain how to re-generate parser files from the grammar.js file.
- [x] refactor GH action for newshell-tests and make it mandatory for PRs from now on.
- [x] added tests for tree-sitter test whenever files in shlr/tree-sitter and shlr/radare2-shell-parser are changed.


thestr4ng3r:
Would be good to indicate what code is generated in `.gitattributes` like this:
https://github.com/thestr4ng3r/chiaki/blob/master/.gitattributes

ret2libc:
Mmm... not sure appveyor is failing because of me. I see errors like:
```
error LNK2019: unresolved external symbol dyn_init referenced in function r_run_config_env [C:\projects\radare2\build\binr\r2agent\a05c824@@r2agent@exe.vcxproj]
```

radare:
* Fix the LGTM issues
* How big (in KB) is this thing?

radare:
Rebase on top of master. Ive fixed this build issue. Sorry

radare:
Can you add (if its not there) a make rule to clone/update this directory like we do for capstone?

Can we trim down this thing to be smaller. At the end we are not going to maintain/touch this code at all because its an external project

ret2libc:
> Can you add (if its not there) a make rule to clone/update this directory like we do for capstone?

It is already in the shlr/Makefile and in shlr/meson.build files.

> Can we trim down this thing to be smaller. At the end we are not going to maintain/touch this code at all because its an external project

I don't think I really can. I had to import all tree-sitter/lib directory and radare2-shell-parser into shlr. Those are the only files I'm touching. How do you want me to split this PR?

ret2libc:
@radare consider that you can completely ignore all the files in `shlr/tree-sitter/*`.

ret2libc:
>     * Fix the LGTM issues

Done, for my code at least. tree-sitter has a bunch of those issues as well, but we can't do anything unless contributing upstream.
 
>     * How big (in KB) is this thing?

libr_core size in release mode on master: 2340288 bytes (2.3M)
libr_core size in release mode with this PR: 2784072 bytes (2.7M)

but please consider that there is a lot of code that I *had* to duplicate, because now there are 2 "parsers", the old one and the new one. Once we will switch again to just one, we'll be able to remove all the old code.

Of course a slightly increase in the size has to be expected, as we bring in tree-sitter lib.

ret2libc:
About sizes:
shlr/libshell_parser.a: 216K
shlr/libtree_sitter.a: 189K

ret2libc:
> **new alerts:**
>     * 3 for Declaration hides parameter
>     * 3 for Comparison of narrow type with wide type in loop condition

They are in tree-sitter, not a big deal. Let's just wait for new releases and keep tree-sitter updated and/or push fixes upstream. Low prio IMHO.

>     * 1 for FIXME comment

It's in my code for the external scanner, I believe right now with the new parser it does not support things like:
```
> /* hello
> world */
```
though it supports multi-line comments if they are provided in one single shot (e.g. through a file).

PR #16665, opened 2020-04-20
Remove Variable Linking and replace with Info from Accesses

thestr4ng3r:
**Detailed description**

Previously, this "linking" was used to add information that in a case like this, where `ecx` is an argument and it is directly saved in `var_3ch`, these variables are linked together.
```
0x00083e83      894c240c       mov dword [var_3ch], ecx    ; arg4
```
However this is now completely unnecessary and we can recover the info directly from the accesses by checking which is a read and which is a write.

This is also the last step to remove everything var-related from sdb, unless I overlooked something.

**Test plan**

`db/cmd/types` contains tests for exactly these cases

### DIFF
--- a/libr/anal/var.c
+++ b/libr/anal/var.c
@@ -5,8 +5,6 @@
 #include <r_cons.h>
 #include <r_list.h>
 
-#define DB a->sdb_fcns
-
 #define ACCESS_CMP(x, y) ((x) - ((RAnalVarAccess *)y)->offset)
 
 R_API bool r_anal_var_display(RAnal *anal, int delta, char kind, const char *type) {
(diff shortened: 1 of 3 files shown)


radare:
Do this mean that we cannot link manually a variabel? It is always tied to the analysis accesses?

thestr4ng3r:
> Do this mean that we cannot link manually a variabel?

No, completely unrelated.

> It is always tied to the analysis accesses?

I don't understand the question.

thestr4ng3r:
It should, as said, despite the naming, this has absolutely nothing to do with `tl` (confused me initially too)

radare:
Ok

PR #16679, opened 2020-04-22
tree-sitter: add option to build with system library

eli-schwartz:
As of https://github.com/tree-sitter/tree-sitter/pull/602 it is possible to install tree-sitter as a system shared library, and distros want to be able to disable vendored code.

/cc @anthraxx @FFY00 @SantiagoTorres (who want this for packaging radare2 on Arch Linux)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,4 +1,4 @@
-project('radare2', 'c', license: 'LGPL3', meson_version: '>=0.47.0')
+project('radare2', 'c', license: 'LGPL3', meson_version: '>=0.50.1')
 
 py3_exe = import('python').find_installation('python3')
 git_exe = find_program('git', required: false)
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -26,6 +26,7 @@ option('use_sys_zlib', type: 'boolean', value: false)
 option('use_sys_lz4', type: 'boolean', value: false)
 option('use_sys_xxhash', type: 'boolean', value: false)
 option('use_sys_openssl', type: 'boolean', value: false)
+option('use_sys_tree_sitter', type: 'boolean', value: false)
 option('use_libuv', type: 'boolean', value: true)
 option('debugger', type: 'boolean', value: true)
 
--- a/shlr/meson.build
+++ b/shlr/meson.build
@@ -233,56 +233,62 @@ sdb_gen_cmd = [
 ]
 
 # handle tree-sitter
-tree_sitter_path = join_paths(meson.current_source_dir(), 'tree-sitter')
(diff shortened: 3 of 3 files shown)


radare:
Add the same for acr

ret2libc:
Awesome! So quick!!

Is there already a distro that ships tree-sitter as system library where I can try this PR?

FFY00:
Yes.

```
pacman -U https://pkgbuild.com/~ffy00/repo/tree-sitter-git-0.16.5.r2995.c393591e-2-x86_64.pkg.tar.zst
```

ret2libc:
Not sure if it works with your package, but just with the tree-sitter.pc file provided by upstream, meson fails because it cannot find `tree_sitter_inc`, used by `libshell_parser`. Am I doing anything wrong?

FFY00:
No, I can reproduce.

FFY00:
It works fine if you just define it outside the if.

```diff
diff --git a/shlr/meson.build b/shlr/meson.build
index b36680531..bde5cc28b 100644
--- a/shlr/meson.build
+++ b/shlr/meson.build
@@ -233,6 +233,7 @@ sdb_gen_cmd = [
 ]

 # handle tree-sitter
+tree_sitter_inc = []
 tree_sitter_dep = dependency('tree-sitter', required: false)
 if not tree_sitter_dep.found() or not get_option('use_sys_tree_sitter')
   message('Use bundled tree-sitter')
```

eli-schwartz:
@FFY00 This would be wrong and result in empty includes, which relies on the tree-sitter includedir being /usr/include as that is already available for other reasons.

I used a partial_dependency instead, and bumped the minimum meson version due to a warning that sounds like we might be affected by (for partial_dependency on a declare_dependency result).

FFY00:
You are right, I should have gotten `includedir` from pkg-config. Either way, your approach works too.

eli-schwartz:
> Add the same for acr

This (configure.acr I assume?) seems to be a custom grammar for configure scripts, and I'm not entirely sure how this will normally toggle between system and vendored deps.

e.g. autotools would do PKG_CHECK_MODULES() and define TREE_SITTER_CFLAGS and TREE_SITTER_LIBS, and if you don't find the module, you'd define those variables some other way. I'm not sure I follow how to override `SHLRS+=tree-sitter/libtree-sitter.a` and other static archives.

ret2libc:
@eli-schwartz I can actually do the acr part, don't worry... I understand nobody uses/know acr :) (I'm not very familiar with either and I try to stay away from it eheh)

XVilka:
Just for the record - `acr` sources are here: https://github.com/radare/acr

PR #16681, opened 2020-04-22
Handle 'q' to interrumpt prompt in newshell ##newshell

### LINKED ISSUE #16668: Fix 'q' command in cfg.newshell
### Work environment

| OS/arch/bits (mandatory)                             | Debian arm 64, Ubuntu x86 32
| Architecture/bits of the file (mandatory)            | PPC, x86/32, x86/64 etc.

### Expected behavior
It exit.

### Actual behavior
It does not exit

```
> e cfg.newshell=true
> q
```

ret2libc:

**Detailed description**

Before this patch the code was assuming a command could return only true or false, but it is also necessary to handle a third state, which makes the prompt stop.

**Test plan**

Just try `q` or `quit` with `cfg.newshell=true`.

**Closing issues**

closes #16668


libr/core/cmd.c
```diff
@@ -5044,7 +5052,14 @@ DEFINE_HANDLE_TS_FCN(arged_command) {
 }
 
 DEFINE_HANDLE_TS_FCN(legacy_quoted_command) {
-	return run_cmd_depth (state->core, node_string) != -1;
+	int res = run_cmd_depth (state->core, node_string);
```
ret2libc, line 5055: maybe this could just be `return run_cmd_depth (state->core, node_string)` as the values are anyway the same.

ret2libc: actually no... :D

libr/core/cmd.c
```diff
@@ -4602,19 +4602,27 @@ static char *ts_node_sub_parent_string(TSNode parent, TSNode node, const char *c
...
 		free (node_string);					\
 		return res;						\
 	} \
-	static bool handle_ts_##name##_internal(struct tsr2cmd_state *state, TSNode node, char *node_string)
+	static RCoreCmdStatus handle_ts_##name##_internal(struct tsr2cmd_state *state, TSNode node, char *node_string)
 
-static bool handle_ts_command(struct tsr2cmd_state *state, TSNode node);
-static bool handle_ts_command_tmpseek(struct tsr2cmd_state *state, TSNode node);
-static bool core_cmd_tsr2cmd(RCore *core, const char *cstr, bool split_lines, bool log);
+#define UPDATE_CMD_STATUS_RES(res, cmd_res, label) \
```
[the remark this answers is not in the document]
ret2libc: yeah it's just that if there is a `quit` command you want it to exit immediately, if a command does not work well for some reason, i try anyway to go ahead with the others. We could also stop immediately though...

ret2libc:
> `ts_node_child_by_field_name`

Why? The field name may be useful as it could happen the index are not always the same in all types of commands.

libr/core/cmd.c
```diff
@@ -5000,7 +5008,7 @@ DEFINE_HANDLE_TS_FCN(arged_command) {
 		char *exec_string = r_str_newf (".%s", cmd_str);
 		free (cmd_str);
 		free (command_str);
-		bool res = r_core_cmd (state->core, exec_string, false) != -1;
+		res = r_core_cmd (state->core, exec_string, false) != -1;
```
karliss, line 5011: r_core_cmd returns the {0, -1, -2, true, false} thing. Assigning to RCoreCmdStatus might cause problems.

libr/core/cmd.c
```diff
@@ -5034,7 +5042,7 @@ DEFINE_HANDLE_TS_FCN(arged_command) {
 	bool command_arg_space = !ts_node_is_null (args) && ts_node_end_byte (command) < ts_node_start_byte (args);
 	char *exec_string = create_exec_string (command_str, pr_args, command_arg_space);
 	R_LOG_DEBUG ("arged_command exec_string = '%s'\n", exec_string);
-	res = r_cmd_call (state->core->rcmd, exec_string) != -1;
+	res = r_cmd_call (state->core->rcmd, exec_string);
```
karliss, line 5045: {true(1), -1}

karliss:
> Why? The field name may be useful as it could happen the index are not always the same in all types of commands.

Aren't TSFieldId handled the same way as TSSymbol where it gets initialized together with TS language? You would still use a named variable, so the clarity of code wouldn't suffer. You might even get a compile time or at least initialization time error if the field gets renamed in syntax description file compared to child_by_field_name error that you would only get when executing specific code path.

ret2libc:
> > Why? The field name may be useful as it could happen the index are not always the same in all types of commands.
> 
> Aren't TSFieldId handled the same way as TSSymbol where it gets initialized together with TS language? You would still use a named variable, so the clarity of code wouldn't suffer. You might even get a compile time or at least initialization time error if the field gets renamed in syntax description file compared to child_by_field_name error that you would only get when executing specific code path.

Yeah sorry I was confusing that method with `ts_node_named_child`. Yeah sounds good, thanks! In another PR.

libr/core/cmd.c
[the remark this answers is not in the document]
ret2libc: Ok, I made it goto label even when cmd_res is INVALID. That should work as well!

PR #16692, opened 2020-04-22
Avoid defining unused is_ts_* functions

ret2libc:
Prevents warnings in the Mac build like:
```
cmd.c:4988:1: warning: unused function 'is_ts_arged_command' [-Wunused-function]
2
```


ret2libc:
> Will this code be gone upon migration? Name `DEFINE_IS_TS_FCN_AND_SYMBOL` looks a bit too long, IMHO. But overall LGTM.

Well not really... We can certainly define those things a bit better though, as right now there are a bit of indirections and a lot of logging necessary for debugging purposes. We can find something better later I guess, I'm open to suggestions.

PR #16694, opened 2020-04-22
Do not redefine assert in dis.c

eagleoflqj:

**Detailed description**

Suppress warnings.

**Test plan**

...

**Closing issues**

...


XVilka:
It's the second place that redefines assert. Could you please to either switch to the functions from `libr/include/r_util/r_assert.h` or to add a new one here and use it everywhere in those places?

XVilka:
Can you switch it to `r_assert` in other places too? Like in https://github.com/radareorg/radare2/pull/16656/files and other places?

eagleoflqj:
> Can you switch it to `r_assert` in other places too? Like in https://github.com/radareorg/radare2/pull/16656/files and other places?

You mean every other C source that includes the system assert.h?

XVilka:
No, just the ones that have "#define assert(" and "#define assert*" and use them.

radare:
better make this code work without any assert imho

radare:
that will make the program crash, but will be better to fork grub code and make it better for our usecase. filesystems dont change that much and we just want to be able to determine which file is in a specific offset. thats only implemented for FAT. this way we can probably stop thinking in syncing with grub. but then we are stuck with the license problem. so a full rewrite of the most common filesystems will be interesting too (or maybe find a way to  get netbsd filesystem (maybe from puffs) to be loaded from RFS too

radare:
we may want to fuzz those and probably fork them completely so we can forget about syncing code. comments?

ret2libc:
> that will make the program crash, but will be better to fork grub code and make it better for our usecase. filesystems dont change that much and we just want to be able to determine which file is in a specific offset. thats only implemented for FAT. this way we can probably stop thinking in syncing with grub. but then we are stuck with the license problem. so a full rewrite of the most common filesystems will be interesting too (or maybe find a way to get netbsd filesystem (maybe from puffs) to be loaded from RFS too

IMHO this is a no-go. Not other code to rewrite/maintain please.

This patch won't crash, it will just work as before. And anyway if an assert triggers it means there was something really wrong, it is not something that should usually happen. This PR seems good to me actually.

radare:
it will crash, clean your glasses and read again:

Old assert = return
New assert = assert

aka, it will crash

ret2libc:
> it will crash, clean your glasses and read again:
> 
> Old assert = return
> New assert = assert
> 
> aka, it will crash

New assert is `r_return_val_if_fail` and `r_warn_if_fail`, isn't it? If so, from `r_assert.h`:
```C

#define r_return_val_if_fail(expr, val) \
	do { \
		if (!(expr)) { \
			H_LOG_ (R_LOGLVL_WARN, "%s: assertion '%s' failed (line %d)\n", R_FUNCTION, #expr, __LINE__); \
			return (val); \
		} \
	} while (0)

#define r_warn_if_fail(expr) \
	do { \
		if (!(expr)) { \
			r_assert_log (R_LOGLVL_WARN, "WARNING (%s:%d):%s%s runtime check failed: (%s)\n", \
				__FILE__, __LINE__, R_FUNCTION, R_FUNCTION[0] ? ":" : "", #expr); \
		} \
	} while (0)

```

I see a return there as well :)

radare:
when doing --with-checks-level=3, this assert goes away , i know that our asserts return by default, but release builds wont behave like this

```
/*
 * R_CHECKS_LEVEL determines the behaviour of the r_return_* set of functions.
 *
 * 0: completely disable every function and make them like no-operation
 * 1: silently enable checks. Check expressions and do return, but do not log anything
 * 2: enable checks and logging (DEFAULT)
 * 3: transform them into real assertion
 */
```

ret2libc:
Yes and you have to enable that on purpose, which is perfectly fine. Even releases won't have that enabled. Asserts are anyway conditions that should not happen, so I see no problem at all. We are also using the same in many other places that before were just returning, so what's the problem?

radare:
My only concern here is that we should fuzz filesystem formats and improve this code to a point that it will be a fork to be considered fine for our standards. But if conflicts get fixed im fine for merging

ret2libc:
Yes please @eagleoflqj fix conflicts and then we can merge, at least we can remove that annoying warning when compiling for now

libr/asm/arch/xap/dis.c
```diff
@@ -521,7 +521,7 @@ static void own(struct state *s)
 		}
 
 		if (s->s_nop) {
-			assert(!s->s_nopd);
+			r_return_if_fail(!s->s_nopd);
```
radare, line 524: missing space before (

libr/asm/arch/xap/dis.c
```diff
@@ -561,7 +561,7 @@ static void own(struct state *s)
 	}
 	if (l) {
 		print_label(s, l);
-		assert(!l->l_next);
+		r_return_if_fail(!l->l_next);
```
radare, line 564: same

PR #16703, opened 2020-04-23
Pass parsed arguments directly to RCmd

ret2libc:

**Detailed description**

Right now the new shell just parse the arguments and then put them back together to call `r_cmd_call`. However, in the future  our command handlers (defined in RCmd) will receive argc/argv, so we need RCmd to be able to merge argv together only when needed (that is when we want to call an "old" command handler that does its own parsing).

This PR introduces RCmdParsedArgs and moves the responsability of creating a single "exec string" to RCmd and not in arged_command function.

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -4561,12 +4561,6 @@ struct tsr2cmd_edit {
 	TSPoint end_point;
 };
 
-struct parsed_args {
-	char *argv_str;
-	int argc;
-	char **argv;
-};
-
 typedef RCoreCmdStatus (*ts_handler)(struct tsr2cmd_state *state, TSNode node);
 
 struct ts_data_symbol_map {
@@ -4750,20 +4744,6 @@ void free_tsr2cmd_edit(struct tsr2cmd_edit *edit) {
 	free (edit);
 }
 
-static void parsed_args_free(struct parsed_args *a) {
-	if (!a) {
-		return;
-	}
-
-	int i;
-	for (i = 0; i < a->argc; i++) {
-		free (a->argv[i]);
-	}
-	free (a->argv);
-	free (a->argv_str);
-	free (a);
-}
-
 static char *do_handle_substitution_cmd(struct tsr2cmd_state *state, TSNode inn_cmd) {
 	RCore *core = state->core;
 	char *inn_str = ts_node_sub_parent_string (state->substitute_cmd, inn_cmd, state->input);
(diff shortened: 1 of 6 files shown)


ret2libc:
@karliss I tried to split commits in a logic way, so you could try to review them commit-by-commit. Hope it helps.

libr/core/cmd.c
```diff
@@ -4877,51 +4857,27 @@ static char *do_handle_ts_unescape_arg(struct tsr2cmd_state *state, TSNode arg)
...
+			free (unescaped_args[i]);
 		}
-		a->argv_str = create_argv_str (a);
-		return a;
+		free (unescaped_args);
+		return res;
 	} else {
-		return handle_ts_unescape_arg (state, args);
+		char *unescaped_args[] = { do_handle_ts_unescape_arg (state, args) };
+		return r_cmd_parsed_args_newargs (1, unescaped_args);
```
ret2libc, line 4879: this probably leaks, i should free unescaped_args[0].

libr/core/cmd.c
```diff
@@ -5172,7 +5165,27 @@ DEFINE_HANDLE_TS_FCN(help_command) {
...
+		RCmdParsedArgs *pr_args = NULL;
+		int res = -1;
+		if (!ts_node_is_null (args)) {
+			pr_args = ts_node_handle_arg_prargs (state, node, args, 1);
+			if (!pr_args) {
+				goto err_else;
+			}
+			r_cmd_parsed_args_setcmd (pr_args, command_str);
+		} else {
+			pr_args = r_cmd_parsed_args_newcmd (command_str);
```
karliss, line 5180: Doesn't this need has_space_after_cmd handling?

ret2libc: no, i'm handling this in exec_string so it does not print the space if there are no additional args apart from argv[0].

libr/core/cmd.c
```diff
@@ -4986,39 +4992,26 @@ DEFINE_HANDLE_TS_FCN(arged_command) {
...
+		}
+	}
+
+	int i;
+	const char *s;
+	r_cmd_parsed_args_foreach_arg (pr_args, i, s) {
+		R_LOG_DEBUG ("parsed_arg %d: '%s'\n", i, s);
 	}
 
 	pr_args->has_space_after_cmd = !ts_node_is_null (args) && ts_node_end_byte (command) < ts_node_start_byte (args);
```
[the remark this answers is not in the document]
ret2libc: yes, I think I'm handling those with the second part of the `&&` `ts_node_end_byte (command) < ts_node_start_byte (args);`.

So, `arged_command` is composed of `<command> + <args>`. So if  `<command>` is ending before the start of `<args>` it means there is at least a space. If instead it is not, it means there is no space.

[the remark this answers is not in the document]
karliss: Could you show me an example of command with args that don't have space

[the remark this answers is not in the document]
ret2libc: https://github.com/radareorg/radare2/pull/16703/files#diff-6165d904c738c087415c0aeca0e91d17R40

`.` is the command. The rest `dr*` is the argument (which is parsed as a command indeed, but then we just need to go back to string).

libr/core/cmd.c
karliss: Then what about the case with args and no space?

ret2libc: That's handled at https://github.com/radareorg/radare2/pull/16703#discussion_r413905943

karliss: How does handling it for completely different command "arged_command" help handling it for "help_command"?

PR #16705, opened 2020-04-23
Unify the relocations loading

08A:

**Detailed description**
The relocations are loaded 3 times... I try to fix that. 

**Test plan**
...

**Closing issues**
#12732
...

### DIFF
--- a/libr/bin/p/bin_elf.inc
+++ b/libr/bin/p/bin_elf.inc
@@ -688,7 +688,7 @@ static RList* relocs(RBinFile *bf) {
 		}
 	}
 	if (bf->o) {
-		if (!(relocs = Elf_(r_bin_elf_get_relocs) (bin))) {
+		if (!(relocs = bin->relocs)) {
 			return ret;
 		}
 		for (i = 0; !relocs[i].last; i++) {
@@ -697,7 +697,6 @@ static RList* relocs(RBinFile *bf) {
 			}
 			r_list_append (ret, ptr);
 		}
-		free (relocs);
 	}
 	return ret;
 }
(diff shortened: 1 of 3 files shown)


08A:
@ret2libc Again i move stuff up... sorry
I was thinking about using Ht_(foreach) on rel_cache when used inside bin_elf.inc?
```
  2 libr/bin/p/bin_elf.inc
  3 691:            if (!(relocs = Elf_(r_bin_elf_get_relocs) (bin))) {
  4 907:    if (!(relcs = Elf_(r_bin_elf_get_relocs) (bin))) {
```

Is this function well optimized?

ret2libc:
> Is this function well optimized?

I guess that depends on what you mean by "well optimized" :) It is a foreach, it iterates over each element of the hash table and run a callback function. You can use that to convert the hashtable to an array. It will be O(n) and that is what matters for now I think. If we need even more optimization, we can take care of that later if/when needed.

08A:
An other alternative could be
```
bin->relocs (array relocs)
bin->rel_cache (Hash Table)
```

08A:
In the idea bin->relocs would hold all the data and rel_cache reference the array

08A:
@ret2libc For me this refactorization is done

libr/bin/p/bin_elf.inc
```diff
@@ -688,7 +688,7 @@ static RList* relocs(RBinFile *bf) {
 		}
 	}
 	if (bf->o) {
-		if (!(relocs = Elf_(r_bin_elf_get_relocs) (bin))) {
+		if (!(relocs = bin->relocs)) {
```
ret2libc, line 691: I'd prefer to keep the access to relocations confined in `Elf_(r_bin_elf_get_relocs)` instead of having to directly access `bin->relocs`.

libr/bin/p/bin_elf.inc
```diff
@@ -944,7 +943,6 @@ static RList* patch_relocs(RBin *b) {
 		r_list_append (ret, ptr);
 	}
 	ht_uu_free (relocs_by_sym);
-	free (relcs);
```
ret2libc, line 947: there are other `free (relcs)` in this function.

08A:
I removed a check for bin->g_section because all the subfunctions do already the check.

PR #16706, opened 2020-04-24
Reduce memory usage in DEX files by caching the getstr calls ##bin

radare:

**Benchmark results**

```
type    top    vsz     rss      time
------------------------------------
calloc  339M   4732640 365452   48s
pvector 338M   4731012 371484   48s // XXX fails somehow with trash indexes
htup    338M   4731556 371620   58s
nocache 350M   4742936 376156   52s
```

nocache -> calloc = 8% faster and 3% less memory usage

IMAGE 2020-04-27 20:02:31
IMAGE 2020-04-27 20:02:35

### DIFF
--- a/libr/bin/format/dex/dex.c
+++ b/libr/bin/format/dex/dex.c
@@ -1,4 +1,4 @@
-/* radare - LGPL - Copyright 2009-2019 - pancake, h4ng3r */
+/* radare - LGPL - Copyright 2009-2020 - pancake, h4ng3r */
 
 #include <r_types.h>
 #include <r_util.h>
@@ -14,6 +14,12 @@ char* r_bin_dex_get_version(RBinDexObj *bin) {
 	return NULL;
 }
 
+// XXX this is never called
+void r_bin_dex_free(RBinDexObj *dex) {
+	r_pvector_fini (&dex->vec_strings);
+	free (dex);
+}
+
 RBinDexObj *r_bin_dex_new_buf(RBuffer *buf) {
 	r_return_val_if_fail (buf, NULL);
 	RBinDexObj *bin = R_NEW0 (RBinDexObj);
@@ -25,6 +31,8 @@ RBinDexObj *r_bin_dex_new_buf(RBuffer *buf) {
 	bin->size = r_buf_size (buf);
 	bin->b = r_buf_ref (buf);
 	/* header */
+	r_pvector_init (&bin->vec_strings, free);
+	bin->htup_strings = ht_up_new0 ();
 	if (bin->size < sizeof (struct dex_header_t)) {
 		goto fail;
 	}
@@ -63,34 +71,36 @@ RBinDexObj *r_bin_dex_new_buf(RBuffer *buf) {
 
 	/* strings */
 	#define STRINGS_SIZE ((dexhdr->strings_size + 1) * sizeof (ut32))
+	if (dexhdr->strings_size > bin->size) {
+		goto fail;
+	}
 	bin->strings = (ut32 *) calloc (dexhdr->strings_size + 1, sizeof (ut32));
 	if (!bin->strings) {
 		goto fail;
 	}
-	if (dexhdr->strings_size > bin->size) {
-		free (bin->strings);
-		goto fail;
-	}
+	r_buf_read_at (bin->b, dexhdr->strings_offset, (ut8*)bin->strings, dexhdr->strings_size * sizeof (ut32));
+	// TODO: this is unnecessary on Big endian machines
 	for (i = 0; i < dexhdr->strings_size; i++) {
-		ut64 offset = dexhdr->strings_offset + i * sizeof (ut32);
+		ut64 offset = dexhdr->strings_offset + (i * sizeof (ut32));
 		if (offset + 4 > bin->size) {
-			free (bin->strings);
-			goto fail;
+			break;
 		}
-		bin->strings[i] = r_buf_read_le32_at (bin->b, offset);
+		bin->strings[i] = r_read_le32 (&bin->strings[i]);
 	}
 	/* classes */
 	// TODO: not sure about if that is needed
-	int classes_size = dexhdr->class_size * DEX_CLASS_SIZE;
+	size_t classes_size = dexhdr->class_size * DEX_CLASS_SIZE;
 	if (dexhdr->class_offset + classes_size >= bin->size) {
-		classes_size = bin->size - dexhdr->class_offset;
-	}
-	if (classes_size < 0) {
-		classes_size = 0;
+		if (dexhdr->class_offset < bin->size) {
+			classes_size = bin->size - dexhdr->class_offset;
+		} else {
+			classes_size = 0;
+		}
 	}
 
 	dexhdr->class_size = classes_size / DEX_CLASS_SIZE;
-	bin->classes = (struct dex_class_t *) calloc (dexhdr->class_size, sizeof (struct dex_class_t));
+	bin->classes = (struct dex_class_t *) calloc (dexhdr->class_size + 1,
+		sizeof (struct dex_class_t));
 	for (i = 0; i < dexhdr->class_size; i++) {
 		ut64 offset = dexhdr->class_offset + i * DEX_CLASS_SIZE;
 		if (offset + 32 > bin->size) {
@@ -110,12 +120,13 @@ RBinDexObj *r_bin_dex_new_buf(RBuffer *buf) {
 	}
 
 	/* methods */
-	int methods_size = dexhdr->method_size * sizeof (struct dex_method_t);
+	size_t methods_size = dexhdr->method_size * sizeof (struct dex_method_t);
 	if (dexhdr->method_offset + methods_size >= bin->size) {
-		methods_size = bin->size - dexhdr->method_offset;
-	}
-	if (methods_size < 0) {
-		methods_size = 0;
+		if (dexhdr->method_offset < bin->size) {
+			methods_size = bin->size - dexhdr->method_offset;
+		} else {
+			methods_size = 0;
+		}
 	}
 	dexhdr->method_size = methods_size / sizeof (struct dex_method_t);
 	bin->methods = (struct dex_method_t *) calloc (methods_size + 1, 1);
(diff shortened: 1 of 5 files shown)


radare:
The memory allocated is leaking, but it was actually leaking more before.
On performance i dont see much difference in times. this pr introduced 2 optimizations that result in less cpu usage but not really noticeable, but worth cleaning up imho:

* when calling getstr() twice with the same idx the previously allocated string is returned.
* Reduce the calls to RBuf.read_le32_at, with just one rbuf_read call and many r_read_le32. to avoid the amount of internal work in RBuffer with IO backend (seek, etc..) which adds many syscalls taht are not needed.

Ill rewrite the hashtable with just an RPVector which is exactly what we need in here

libr/bin/format/dex/dex.c
```diff
@@ -14,6 +14,12 @@ char* r_bin_dex_get_version(RBinDexObj *bin) {
 	return NULL;
 }
 
+// XXX this is never called
```
XVilka, line 17: Why not to remove it then, if its never called?

libr/bin/p/bin_dex.c
```diff
@@ -1620,7 +1634,7 @@ static void parse_class(RBinFile *bf, RBinDexClass *c, int class_index, int *met
 			rbin->cb_printf ("  source_file_idx   : %d (%s)\n\n",
 					 c->source_file, source_file);
 		}
-		free (source_file);
+		// free (source_file);
```
XVilka, line 1637: Why not to remove free then?

libr/bin/p/bin_dex.c
```diff
@@ -1551,9 +1565,9 @@ static void parse_class(RBinFile *bf, RBinDexClass *c, int class_index, int *met
 			if (t > 0 && t < dex->header.types_size ) {
 				int tid = dex->types[t].descriptor_id;
 				if (dexdump) {
-					char *cn = getstr (dex, tid);
+					const char *cn = getstr (dex, tid);
 					rbin->cb_printf ("    #%d              : '%s'\n", z, cn);
-					free (cn);
+					// free (cn);
```
XVilka, line 1570: Same here, why not to remove free?

libr/bin/p/bin_dex.c
```diff
@@ -1372,7 +1385,7 @@ static void parse_dex_class_method(RBinFile *bf, RBinDexClass *c, RBinClass *cls
 									s,
 									handler_addr);
 							}
-							free (s);
+							s = NULL; // free (s);
```
XVilka, line 1388: memory leak?

libr/bin/p/bin_dex.c
```diff
@@ -78,31 +78,43 @@ static ut64 dex_field_offset(RBinDexObj *bin, int fid) {
...
 	if (!len || len >= dex->size) {
 		return NULL;
 	}
 	ut8 *ptr = malloc (len + 1);
-	if (!ptr) {
-		return NULL;
+	if (ptr) {
+		r_buf_read_at (dex->b, string_index + uleblen, ptr, len);
+		ptr[len] = 0;
+		if (dex->vec_strings.v.len) {
```
ret2libc, line 109: i don't understand why you didn't use `HtUP`. I honestly don't understand this code. So... if you have more than one element in the vector, you initialize it, right? Then, few lines below, if you have more than one element in the vector, you insert `ptr` ad `idx`.

What happens if the vector is empty and then you try to insert an element in the RPVector at index=10?  IMHO, an HtUP perfectly fits what you need. I don't see why not using it.

libr/bin/format/dex/dex.c
radare: because its a not for my future self to fix the memory leaks that have been there for ever

libr/bin/p/bin_dex.c
radare: it happens that it returns null, htup is unnecessary overengineering when we just need an array

libr/bin/p/bin_dex.c
radare: nope, because its not freed until the the rbindex obj is freed

libr/bin/p/bin_dex.c
radare: same as above

radare: getstr returns const char*, not owned, so not leaking

libr/bin/p/bin_dex.c
radare: ok will do now

libr/bin/format/dex/dex.c
radare: ill fix add the callback in all the plugins to free the rbinobj instance and all the related leaks in a separate PR

ret2libc:
@radare I'm not sure this works. Can you provide a testcase where the new rpvector is used? I put breakpoints on the lines where the rpvector is used and it never breaks (I tried dex/test.dex, dex/hello.dex, java/Main.dex)

Also, why a cache reduce the memory usage? Doesn't it increase it?

radare:
all the DEX tests in the testsuite use this code. if the code wasnt correct it will crash or fail

radare:
cache improves memory usage because the string pool is queries 999999 times, 800000 of them for the same strings. as long as you cant read or test the PR ill explain again:

before:

* getstr = returns NEW string in the heap with strdup

after:

* getstr = caches the index -> string and just returns a weak pointer, the heap pointer is owned by the rpvector, so its not freed until the RBinObject is freed, which right now it doesnt happen because its a bug in the RBin api design that i plan to fix in another pr when this gets merged

libr/util/vector.c
```diff
@@ -143,10 +143,13 @@ R_API void r_vector_remove_at(RVector *vec, size_t index, void *into) {
 
 R_API void *r_vector_insert(RVector *vec, size_t index, void *x) {
 	r_return_val_if_fail (vec, NULL);
-	if (vec->len >= vec->capacity) {
+	if (R_MAX (vec->len, index) >= vec->capacity) {
```
ret2libc, line 146: why this? index >= vec->capacity and you use NEXT_VECTOR_CAPACITY you may end up anyway with a vector that is not big enough for index.

radare: damn this shouldnt be commited, this is wrong. ill remove it and fix the CI i pushed the code because i had to run afk. and ill open an issue for all the problems i found in RVector and RPVector

ret2libc:
So what did you choose here? I'm ok with calloc, but we should ensure strings size cannot be too big.

libr/include/r_vector.h
```diff
@@ -158,6 +158,7 @@ R_API void *r_vector_shrink(RVector *vec);
 // RPVector
 
 R_API void r_pvector_init(RPVector *vec, RPVectorFree free);
+R_API void r_pvector_fini(RPVector *vec);
```
radare, line 161: this is unrelated

libr/bin/p/bin_dex.c
```diff
@@ -1988,11 +1999,11 @@ static RList *sections(RBinFile *bf) {
 			ptr->size = ptr->vaddr; // fix size
 		}
 		ptr->vsize = ptr->size;
+		// Commenting this line speedups loading from 4.5s to 3s
```
radare, line 2002: @thestr4ng3r is on this

libr/include/r_vector.h
radare: but needed because it was not defined :(

radare:
fixed the lgtm

libr/bin/p/bin_dex.c
```diff
@@ -76,33 +76,42 @@ static ut64 dex_field_offset(RBinDexObj *bin, int fid) {
...
-	ut64 len;
-	int uleblen;
-	if (!dex || idx < 0 || idx >= dex->header.strings_size || !dex->strings) {
+	if (idx < 0 || idx >= dex->header.strings_size || !dex->strings) {
 		return NULL;
 	}
-	if (dex->strings[idx] >= dex->size) {
+	if (dex->cal_strings) {
+		const char *p = dex->cal_strings[idx];
+		if (p && *p) {
```
ret2libc, line 86: `R_STR_ISNOTEMPTY`

libr/bin/format/dex/dex.c
```diff
@@ -63,34 +74,36 @@ RBinDexObj *r_bin_dex_new_buf(RBuffer *buf) {
...
 	/* strings */
 	#define STRINGS_SIZE ((dexhdr->strings_size + 1) * sizeof (ut32))
-	bin->strings = (ut32 *) calloc (dexhdr->strings_size + 1, sizeof (ut32));
-	if (!bin->strings) {
+	if (dexhdr->strings_size > bin->size) {
 		goto fail;
 	}
-	if (dexhdr->strings_size > bin->size) {
-		free (bin->strings);
+	bin->strings = (ut32 *) calloc (dexhdr->strings_size + 1, sizeof (ut32));
```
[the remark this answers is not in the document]
radare: done

libr/bin/format/dex/dex.c
```diff
@@ -134,13 +148,10 @@ RBinDexObj *r_bin_dex_new_buf(RBuffer *buf) {
 	}
 
 	/* types */
-	int types_size = dexhdr->types_size * sizeof (struct dex_type_t);
+	size_t types_size = dexhdr->types_size * sizeof (struct dex_type_t);
```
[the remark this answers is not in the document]
radare: yes, but not in this pr because its unrelated

radare:
as i said 3 times, the leak introduced here by allocating cal_strings will be fixed in another PR because it affects ALL the bin plugins. i even filled an issue with more details

ret2libc:
> as i said 3 times, the leak introduced here by allocating cal_strings will be fixed in another PR because it affects ALL the bin plugins. i even filled an issue with more details

Ok :/ then there is no need for that function. but ok.

Fix CI and then I can approve.

libr/bin/p/bin_dex.c
thestr4ng3r: wat am I?

libr/bin/format/dex/dex.c
```diff
@@ -63,34 +74,36 @@ RBinDexObj *r_bin_dex_new_buf(RBuffer *buf) {
...
 	/* classes */
 	// TODO: not sure about if that is needed
-	int classes_size = dexhdr->class_size * DEX_CLASS_SIZE;
+	size_t classes_size = dexhdr->class_size * DEX_CLASS_SIZE;
 	if (dexhdr->class_offset + classes_size >= bin->size) {
-		classes_size = bin->size - dexhdr->class_offset;
-	}
-	if (classes_size < 0) {
-		classes_size = 0;
+		if (dexhdr->class_offset < bin->size) {
```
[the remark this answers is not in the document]
radare: unrelated to this pr

radare:
let's wait for the CI

PR #16716, opened 2020-04-25
Refactor Meta out of SDB

thestr4ng3r:
This removes meta from SDB and instead stores it in an RIntervalTree. More info on the storage can be found in the comments in `r_anal.h`.

One important change in the API is that it now returns weak references into the existing RAnalMetaItems, so callers must not free the returned stuff anymore!

This also includes some minor convenience changes in RIntervalTree and fixes for `@@@C:` in both old and new shells because that had to be refactored somehow to conform to the new API but it was completely broken so I fixed it as a side effect.

### DIFF
--- a/libr/anal/anal.c
+++ b/libr/anal/anal.c
@@ -64,6 +64,18 @@ static void zign_rename_for(REvent *ev, int type, void *user, void *data) {
 void r_anal_hint_storage_init(RAnal *a);
 void r_anal_hint_storage_fini(RAnal *a);
 
+static void r_meta_item_fini(RAnalMetaItem *item) {
+	free (item->str);
+}
+
+static void r_meta_item_free(void *_item) {
+	if (_item) {
+		RAnalMetaItem *item = _item;
+		r_meta_item_fini (item);
+		free (item);
+	}
+}
+
 R_API RAnal *r_anal_new(void) {
 	int i;
 	RAnal *anal = R_NEW0 (RAnal);
@@ -95,8 +107,8 @@ R_API RAnal *r_anal_new(void) {
 	r_event_hook (anal->zign_spaces.event, R_SPACE_EVENT_COUNT, zign_count_for, NULL);
 	r_event_hook (anal->zign_spaces.event, R_SPACE_EVENT_RENAME, zign_rename_for, NULL);
 	anal->sdb_fcns = sdb_ns (anal->sdb, "fcns", 1);
-	anal->sdb_meta = sdb_ns (anal->sdb, "meta", 1);
 	r_anal_hint_storage_init (anal);
+	r_interval_tree_init (&anal->meta, r_meta_item_free);
 	anal->sdb_types = sdb_ns (anal->sdb, "types", 1);
 	anal->sdb_fmts = sdb_ns (anal->sdb, "spec", 1);
 	anal->sdb_cc = sdb_ns (anal->sdb, "cc", 1);
@@ -147,6 +159,7 @@ R_API RAnal *r_anal_free(RAnal *a) {
 	ht_pp_free (a->ht_name_fun);
 	set_u_free (a->visited);
 	r_anal_hint_storage_fini (a);
+	r_interval_tree_fini (&a->meta);
 	free (a->cpu);
 	free (a->os);
 	free (a->zign_path);
@@ -428,8 +441,9 @@ R_API bool r_anal_op_is_eob(RAnalOp *op) {
 
 R_API int r_anal_purge (RAnal *anal) {
 	sdb_reset (anal->sdb_fcns);
-	sdb_reset (anal->sdb_meta);
 	r_anal_hint_clear (anal);
+	r_interval_tree_fini (&anal->meta);
+	r_interval_tree_init (&anal->meta, r_meta_item_free);
 	sdb_reset (anal->sdb_types);
 	sdb_reset (anal->sdb_zigns);
 	sdb_reset (anal->sdb_classes);
--- a/libr/anal/reflines.c
+++ b/libr/anal/reflines.c
@@ -122,12 +122,32 @@ R_API RList *r_anal_reflines_get(RAnal *anal, ut64 addr, const ut8 *buf, ut64 le
 		}
 		addr += sz;
 		{
-			RAnalMetaItem *mi = r_meta_find_any_except (anal, addr, R_META_TYPE_COMMENT, 0);
-			if (mi) {
-				ptr += mi->size;
-				addr += mi->size;
-				r_meta_item_free (mi);
-				goto __next;
+			RPVector *metas = r_meta_get_all_at (anal, addr);
+			if (metas) {
+				void **it;
+				ut64 skip = 0;
+				r_pvector_foreach (metas, it) {
+					RIntervalNode *node = *it;
+					RAnalMetaItem *meta = node->data;
+					switch (meta->type) {
+					case R_META_TYPE_DATA:
+					case R_META_TYPE_STRING:
+					case R_META_TYPE_HIDE:
+					case R_META_TYPE_FORMAT:
+					case R_META_TYPE_MAGIC:
+						skip = r_meta_node_size (node);
+						goto do_skip;
+					default:
+						break;
+					}
+				}
+do_skip:
+				r_pvector_free (metas);
+				if (skip) {
+					ptr += skip;
+					addr += skip;
+					goto __next;
+				}
 			}
 		}
 		if (!anal->iob.is_valid_offset (anal->iob.io, addr, 1)) {
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -876,12 +876,12 @@ static int __core_anal_fcn(RCore *core, ut64 at, ut64 from, int reftype, int dep
 					if (i == nexti) {
 						ut64 at = r_anal_function_max_addr (fcn);
 						while (true) {
-							RAnalMetaItem *mi = r_meta_find (core->anal, at, R_META_TYPE_ANY, 0);
+							ut64 size;
+							RAnalMetaItem *mi = r_meta_get_at (core->anal, at, R_META_TYPE_ANY, &size);
 							if (!mi) {
 								break;
 							}
-							at += mi->size;
-							r_meta_item_free (mi);
+							at += size;
 						}
 						// TODO: ensure next address is function after padding (nop or trap or wat)
 						// XXX noisy for test cases because we want to clear the stderr
@@ -3868,8 +3868,8 @@ static bool found_xref(RCore *core, ut64 at, ut64 xref_to, RAnalRefType type, in
 				r_flag_space_pop (core->flags);
 				free (str_flagname);
 				if (len > 0) {
-					r_meta_add (core->anal, R_META_TYPE_STRING, xref_to,
-							xref_to + len, (const char *)str_string);
+					r_meta_set (core->anal, R_META_TYPE_STRING, xref_to,
+								len, (const char *) str_string);
 				}
 				free (str_string);
 			}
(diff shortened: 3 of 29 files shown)


ret2libc:
@thestr4ng3r what's the state of this? Do you want us to start reviewing something or what?

thestr4ng3r:
No please don't, it's still way too broken and wip.

thestr4ng3r:
I have yet to re-check all the callers into meta for memory handling in detail so this is still wip, but you can start to review now.

thestr4ng3r:
Start with the changes in r_anal.h, the API is defined there.

thestr4ng3r:
When loading dex:
Old:
Screenshot from 2020-05-08 11-21-16
New:
Screenshot from 2020-05-08 11-21-04

It is clearly much faster, but obviously the RBTree isn't completely free either.

thestr4ng3r:
Memory usage before: 383.9MB, after: 317.5MB (measured with massif)

libr/core/cmd_open.c
```diff
@@ -1019,14 +1019,7 @@ static void __rebase_everything(RCore *core, RList *old_sections, ut64 old_base)
...
 	// META
-	RList *meta_list = r_meta_enumerate (core->anal, R_META_TYPE_ANY);
-	RAnalMetaItem *item;
-	r_list_foreach (meta_list, it, item) {
-		r_meta_del (core->anal, item->type, item->from, item->size);
-		item->from += diff;
-		r_meta_add_with_subtype (core->anal, item->type, item->subtype, item->from, item->from + item->size, item->str);
-	}
-	r_list_free (meta_list);
+	r_meta_rebase (core->anal, diff);
```
thestr4ng3r, line 1022: fyi @yossizap

thestr4ng3r:
Ready now. @radare in case you want to merge, DO NOT squash. The commits are already cleaned up to be merged directly.

libr/anal/anal.c
```diff
@@ -64,6 +64,8 @@ static void zign_rename_for(REvent *ev, int type, void *user, void *data) {
 void r_anal_hint_storage_init(RAnal *a);
 void r_anal_hint_storage_fini(RAnal *a);
 
+R_IPI void r_meta_item_free(void *_item);
```
ret2libc, line 67: does it need `R_IPI`? It seems to be used only in this file, `static`?

test/unit/test_anal_meta.c
```diff
@@ -0,0 +1,615 @@
+
+#include <r_anal.h>
+
+#include "minunit.h"
+
+bool test_meta_set() {
+	RAnal *anal = r_anal_new ();
+
+	r_meta_set (anal, R_META_TYPE_DATA, 0x100, 4, NULL);
+	r_meta_set_string (anal, R_META_TYPE_COMMENT, 0x100, "summer of love");
```
ret2libc, line 10: I was slightly confused at first, because it may seem that `r_meta_set_string` would set a `R_META_TYPE_STRING`.

test/unit/test_anal_meta.c
```diff
@@ -0,0 +1,615 @@
...
+
+	items = r_meta_get_all_in (anal, 0xff, R_META_TYPE_ANY);
+	mu_assert_eq (r_pvector_len (items), 0, "all count");
+	r_pvector_free (items);
+
+	items = r_meta_get_all_in (anal, 0x101, R_META_TYPE_COMMENT);
+	mu_assert_eq (r_pvector_len (items), 0, "all count");
+	r_pvector_free (items);
+
+	items = r_meta_get_all_in (anal, 0x103, R_META_TYPE_DATA);
```
ret2libc, line 307: haven't you already done this test at https://github.com/radareorg/radare2/pull/16716/commits/0c3cddc779dc59219e87b20b0d75bc4a78055bc7#diff-2bdaf63af99d8302c5148cd3897de99cR293 ?

libr/anal/reflines.c
```diff
@@ -122,12 +122,32 @@ R_API RList *r_anal_reflines_get(RAnal *anal, ut64 addr, const ut8 *buf, ut64 le
...
-				r_meta_item_free (mi);
-				goto __next;
+			RPVector *metas = r_meta_get_all_at (anal, addr);
+			if (metas) {
+				void **it;
+				ut64 skip = 0;
+				r_pvector_foreach (metas, it) {
+					RIntervalNode *node = *it;
+					RAnalMetaItem *meta = node->data;
+					switch (meta->type) {
```
ret2libc, line 132: isn't it better to do
```C
switch (meta->type) {
case R_META_COMMENT:
   break;
default:
  skip = ..
  goto do_skip;
}
```
It seems to me it would be more in line with the old semantic.

libr/include/r_anal.h
```diff
@@ -1727,33 +1717,77 @@ R_API void r_anal_data_free (RAnalData *d);
...
+	return end - start + 1;
+}
+
+static inline ut64 r_meta_node_size(RIntervalNode *node) {
+	return r_meta_item_size (node->start, node->end);
+}
+
+// Set a meta item at addr with the given contents in the current space.
+// If there already exists an item with this type and space at addr (regardless of its size) it will be overwritten.
+R_API bool r_meta_set(RAnal *a, RAnalMetaType type, ut64 addr, ut64 size, const char *str);
```
ret2libc, line 1740: Old function was `r_meta_add`. Was it always adding a new meta?

How can i set multiple comments at the same address with this new API?

libr/anal/anal.c
thestr4ng3r: Indeed, it can be moved.

test/unit/test_anal_meta.c
thestr4ng3r: yep

libr/anal/reflines.c
thestr4ng3r: In theory yes, but after the refactoring all metas have a size of at least 1 (i.e. they cover at least one byte. Before they could have 0). This can lead to some brokenness here with the types not listed in the switch.

libr/include/r_anal.h
thestr4ng3r: No, `r_meta_add()` was also overwriting, despite the name. For example run this on master with #!c, then `C`:
```
#include <r_core.h>

void entry(RCore *core, int argc, const char **argv)
{
	r_meta_add(core->anal, R_META_TYPE_COMMENT, 0x100, 0x100, "comment a");
	r_meta_add(core->anal, R_META_TYPE_COMMENT, 0x100, 0x100, "comment b");
}
```

You can't set multiple comments at the same address right now (except in different spaces), which corresponds to how it was before.

libr/core/canal.c
```diff
@@ -4275,25 +4275,29 @@ R_API RCoreAnalStats* r_core_anal_get_stats(RCore *core, ut64 from, ut64 to, ut6
...
-		}
-		piece = (M->from - from) / step;
-		switch (M->type) {
-		case R_META_TYPE_STRING:
-			as->block[piece].strings++;
-			break;
-		case R_META_TYPE_COMMENT:
-			as->block[piece].comments++;
-			break;
+	RPVector *metas = to >= from ? r_meta_get_all_intersect (core->anal, from, to - from, R_META_TYPE_ANY) : NULL;
```
radare, line 4278: You can replace the ternary operator with an if and an early return, the rest of the function wont do a thing if the check fails

libr/core/cbin.c
```diff
@@ -1236,8 +1236,8 @@ static int bin_entry(RCore *r, int mode, ut64 laddr, int va, bool inifin) {
 			}
 			r_flag_set (r->flags, str, at, 1);
 			if (is_initfini (entry) && hvaddr != UT64_MAX) {
-				r_meta_add (r->anal, R_META_TYPE_DATA, hvaddr,
-				            hvaddr + entry->bits / 8, NULL);
+				r_meta_set (r->anal, R_META_TYPE_DATA, hvaddr,
```
radare, line 1239: That fits in. 1 line

libr/core/cbin.c
```diff
@@ -2759,7 +2759,7 @@ static int bin_sections(RCore *r, int mode, ut64 laddr, int va, ut64 at, const c
 				str = r_str_newf ("[%02d] %s %s size %" PFMT64d" named %s%s%s",
 				                  i, perms, type, size,
 				                  pfx? pfx: "", pfx? ".": "", section->name);
-				r_meta_add (r->anal, R_META_TYPE_COMMENT, addr, addr, str);
+				r_meta_set (r->anal, R_META_TYPE_COMMENT, addr, 1, str);
```
radare, line 2762: Why 1 and not size?

libr/core/cmd.c
```diff
@@ -6100,13 +6067,31 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(foreach_cmd_command) {
...
+	RCore *core = state->core;
 	TSNode command = ts_node_named_child (node, 0);
-	struct foreach_comment_newshell_t cmt_t = {
-		.command = &command,
-		.state = state,
-	};
-	r_meta_list_cb (state->core->anal, R_META_TYPE_COMMENT, 0, foreach_comment_newshell, (void *)&cmt_t, UT64_MAX);
-	return R_CMD_STATUS_OK;
+	TSNode filter_node = ts_node_named_child (node, 1);
+	char *glob = NULL;
```
radare, line 6073: Use ternary operator to avoid double assignment

libr/core/cmd_search.c
```diff
@@ -1924,10 +1922,22 @@ static void do_ref_search(RCore *core, ut64 addr,ut64 from, ut64 to, struct sear
 			r_parse_filter (core->parser, ref->addr, core->flags, hint, r_strbuf_get (&asmop.buf_asm),
 				str, sizeof (str), core->print->big_endian);
 			r_anal_hint_free (hint);
-			comment = r_meta_get_string (core->anal, R_META_TYPE_COMMENT, ref->addr);
+			const char *comment = r_meta_get_string (core->anal, R_META_TYPE_COMMENT, ref->addr);
+			char *print_comment = NULL;
+			if (comment && strchr (comment, '\n')) { // display only until the first newline
+				print_comment = strdup (comment);
```
radare, line 1928: Use r_str_ndup and avoid the code below

libr/core/visual.c
```diff
@@ -2935,17 +2936,17 @@ R_API int r_core_visual_cmd(RCore *core, const char *arg) {
...
+					ut64 amisize;
+					RAnalMetaItem *ami = r_meta_get_at (core->anal,
+														core->offset, R_META_TYPE_DATA,
+														&amisize);
 					if (!ami) {
-						ami = r_meta_find (core->anal,
-								core->offset, R_META_TYPE_STRING,
-								R_META_WHERE_HERE);
+						ami = r_meta_get_at (core->anal,
+											 core->offset, R_META_TYPE_STRING,
```
radare, line 2945: Shitty indent

libr/core/vmenus.c
```diff
@@ -917,7 +898,16 @@ R_API bool r_core_visual_hudstuff(RCore *core) {
 	}
 	list->free = free;
 	r_flag_foreach (core->flags, hudstuff_append, list);
-	sdb_foreach (core->anal->sdb_meta, cmtcb, list);
+	RIntervalTreeIter it;
+	RAnalMetaItem *mi;
+	r_interval_tree_foreach (&core->anal->meta, it, mi) {
+		if (mi->type == R_META_TYPE_COMMENT) {
+			char *s = r_str_newf ("0x%"PFMT64x" %s", r_interval_tree_iter_get (&it)->start, mi->str);
```
radare, line 905: Use %08

libr/core/canal.c
thestr4ng3r: Normally I would agree about the early return, but here it is only one part of the function that might fail, not the entire function. Someone might add another type of info after the meta stuff that shouldn't be skipped if the meta check fails and early return would break that.
But the `to >= from` condition is actually wrong. `r_meta_get_all_intersect() asserts for size >= 1`.

libr/core/cbin.c
thestr4ng3r: I think comments always have size 1, that's also how it has been before. Or well `addr, addr` is size 0 to be exact.

radare:
Agree. Missed the context. What happens with 0 sized metas? Are those possible?

thestr4ng3r:
> Agree. Missed the context. What happens with 0 sized metas? Are those possible?

No, as documented in `r_anal.h`:
> Meta uses Condret's Klemmbaustein Priciple, i.e. intervals are defined inclusive/inclusive.
> A meta item from 0x42 to 0x42 has a size of 1. Items with size 0 do not exist.
> Meta items are allowed to overlap and the internal data structure allows for multiple meta items
> starting at the same address.

This implies that it is possible to have a meta item covering the entire memory range.

radare:
Perfect 🤘

PR #16718, opened 2020-04-25
Avoid the use of 'for (int' constructions ##build

radare:

**Detailed description**

without this r2 doesnt build on some compilers

**Test plan**

see r2frida travis or try any random old bsd

**Closing issues**

nope

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -5057,7 +5057,8 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(repeat_command) {
 	}
 
 	RCoreCmdStatus res = R_CORE_CMD_STATUS_OK;
-	for (int i = 0; i < rep; i++) {
+	size_t i;
+	for (i = 0; i < rep; i++) {
 		RCoreCmdStatus cmd_res = handle_ts_command (state, command);
 		UPDATE_CMD_STATUS_RES (res, cmd_res, err);
 	}
--- a/libr/core/windows_heap.c
+++ b/libr/core/windows_heap.c
@@ -272,7 +272,8 @@ static bool GetHeapGlobalsOffset(RDebug *dbg, HANDLE h_proc) {
 	char *ntdllopen = dbg->corebind.cmdstrf (dbg->corebind.core, "ob~%s", ntdll);
 	if (*ntdllopen) {
 		char *saddr = strtok (ntdllopen, " ");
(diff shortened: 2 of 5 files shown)


ret2libc:
Can we wait a little bit and I try to push these changes to tree sitter upstream? I'd like to avoid changes to external repositories

XVilka:
What about merging without tree-sitter changes?

XVilka:
```
test_list.c:388:6: error: redefinition of 'i'
        int i;
            ^
test_list.c:380:6: note: previous definition is here
        int i;
            ^
test_list.c:416:6: error: redefinition of 'i'
        int i;
            ^
test_list.c:408:6: note: previous definition is here
        int i;
            ^
2 errors generated.
```

radare:
Imho we cant wait. Ive been receiving complains in DM for 3 days. Dont know how fast are treesitter people in merging and releasing, but i dont see a problem in having an early fix. Ill fix that compilation error

XVilka:
@radare they are fast and PR is already there: https://github.com/tree-sitter/tree-sitter/pull/604

ret2libc:
I think there is an easy workaround for those who need to compile things right now. CFLAGS=-std=c99 ./configure; make . Also, instead of DM we would like to see issues opened on GitHub with full details. Thanks

shlr/tree-sitter/lib/src/stack.c
```diff
@@ -171,7 +171,8 @@ static bool stack__subtree_is_equivalent(Subtree left, Subtree right) {
 static void stack_node_add_link(StackNode *self, StackLink link, SubtreePool *subtree_pool) {
   if (link.node == self) return;
 
-  for (int i = 0; i < self->link_count; i++) {
+  int i;
```
ret2libc, line 174: didn't you say that your PR was not touching tree-sitter?

radare: nope, i didnt said this

radare:
i said we should fix the build until we can update treesitter when their code is compilable

shlr/tree-sitter/lib/src/stack.c
ret2libc: this is not needed. remove it please. tree-sitter is already compiled with `-std=c99` so it does not complain. https://github.com/radareorg/radare2/blob/master/shlr/Makefile#L334

radare:
done

PR #16720, opened 2020-04-26
Fix #16543 - bin.str.enc affects izz + ascii is now alias for latin1

### LINKED ISSUE #16543: Missing strings on iOS binary with izz
### Work environment
OS/arch/bits (mandatory) : Arch Linux x86_64
File format of the file you reverse (mandatory): Mach-O 64-bit arm64 executable
Architecture/bits of the file (mandatory): Arm64
r2 -v full output: radare2 4.3.1 0 @ linux-x86-64 git.4.3.1/ commit: unknown build: 2020-03-20__16:02:58

### Expected behavior
izz find all the strings of the binary

### Actual behavior
izz missed the 'encryptPassword:' string: 
r24
that is present both in "strings" command:
r26
and is shown also in the binary:
R21
r23

The string has not strange encoding. Is a simple ASCII string.

-> r2 ipa://DemoHITB.ipa
-> aaa
-> e emu.str=true
-> aaa
-> izz~encryptPassword  # String 'encryptPassword:' is missing
1638 0x0004d016 0x10004d016 29  30                             ascii   +[LoginUtil encryptPassword:]
2409 0x00056ef0 0x100056ef0 29  30                             ascii   +[LoginUtil encryptPassword:]
-> s 0x100012fc4  # In the next three instruction you can see missing string 'encryptPassword:'
-> s 0x100029749  # In this location there is the missing string. The string view fail to print it correctly:
r22
but it is a simple string, as show in the HEX view:
r23

[ipaBugRadare.zip](https://github.com/radareorg/radare2/files/4466122/ipaBugRadare.zip)

kazarmy:

**Detailed description**

This pr allows `bin.str.enc` to affect `izz` (though not `iz` for some reason), and `ascii` is now an alias for `latin1`.

**Test plan**

**Closing issues**

Fixes #16543.

### DIFF
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -407,7 +407,23 @@ static void get_strings_range(RBinFile *bf, RList *list, int min, int raw, ut64 
 			return;
 		}
 	}
-	string_scan_range (list, bf, min, from, to, -1, raw, section);
+	int type;
+	const char *enc = bf->rbin->strenc;
+	if (!enc) {
+		type = R_STRING_TYPE_DETECT;
+	} else if (!strcmp (enc, "latin1")) {
+		type = R_STRING_TYPE_ASCII;
+	} else if (!strcmp (enc, "utf8")) {
+		type = R_STRING_TYPE_UTF8;
+	} else if (!strcmp (enc, "utf16le")) {
+		type = R_STRING_TYPE_WIDE;
+	} else if (!strcmp (enc, "utf32le")) {
+		type = R_STRING_TYPE_WIDE32;
+	} else { // TODO utf16be, utf32be
+		eprintf ("ERROR: encoding %s not supported\n", enc);
+		return;
+	}
+	string_scan_range (list, bf, min, from, to, type, raw, section);
 }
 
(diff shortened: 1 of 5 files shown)


kazarmy:
Can do `r2 -e bin.str.enc=ascii binary` if one wants the flags to match up.

libr/bin/bfile.c
```diff
@@ -407,7 +407,23 @@ static void get_strings_range(RBinFile *bf, RList *list, int min, int raw, ut64
 			return;
 		}
 	}
-	string_scan_range (list, bf, min, from, to, -1, raw, section);
+	int type;
+	const char *enc = bf->rbin->strenc;
```
XVilka, line 411: What about extracting it into the separate function?

kazarmy: Could be done but right now the code is unique afaik. There's another code structure like it in disasm.c but the enum involved is different. Probably will merge the two enums when the need arises.

kazarmy: Also have to decide whether it's best for the encoding to be stored as a string or an enum.

Issue #16624, opened 2020-04-17
Types API


thestr4ng3r:
Types are stored in sdb and there is a lot of code in r2 that uses it.
But there is almost no API for types and currently all of this code directly accesses the sdb, resulting in huge amounts of code duplication and highly fragile code in case the format will change in the future. Even r2ghidra has to manually pull out information from the sdb:
https://github.com/radareorg/r2ghidra-dec/blob/3773ed1a33f1e0aec77d4975c02ce573651eb7ad/src/R2TypeFactory.cpp#L42-L85

This should be made into a nice API. I am thinking about something like this (rough sketch):
```c
typedef enum {
    R_ANAL_BASE_TYPE_KIND_STRUCT,
    R_ANAL_BASE_TYPE_KIND_UNION,
    R_ANAL_BASE_TYPE_KIND_ENUM,
    ...
} RAnalBaseTypeKind;

typedef struct r_anal_base_type_struct {
    RVector members; // describe the struct in some way
} RAnalBaseTypeStruct;

typedef struct r_anal_base_type_enum {
    RVector cases; // list of all the enum cases
} RAnalBaseTypeEnum;

typedef struct r_anal_base_type {
    RAnalBaseTypeKind kind;
    union {
        RAnalBaseTypeStruct ztruct;
        RAnalBaseTypeEnum inum;
        ...
    };
} RAnalBaseType;

R_API RAnalBaseType *r_anal_get_base_type(const char *name);
```

Note: I am explicitly calling this `RAnalBaseType` because these are basically the "atomic" types that are stored in the database. These base types are definitions of structs, unions, enums, int types, etc. but NOT something like `Something *[]`.

XVilka:
Yes, good idea and will make the types code more readable

HoundThe:
Shouldn't the second **r_anal_base_type_struct** be an **r_anal_base_type_enum** ?

thestr4ng3r:
Yes, good catch.

XVilka:
Since the PR was merged, please update the issue text.

thestr4ng3r:
The scope of the issue is done. For further steps we should make follow-up issues.

Issue #10031, opened 2018-05-07
C types - structures alignment


XVilka:
Currently we perform a dumb size and offset calculation for the structures. We need the way to set, store and use the information about alignment of the whole structure, some of its members or parts of it.

Issue #7710, opened 2017-06-09
C types - SDB properties


XVilka:
- [ ] Add support for bitfields
 - [ ] Add support for alignment property of the type
 - [ ] Support for C++ objects

radare:
C++?

XVilka:
@radare yes, not sure how to do that, but I think we need to support more complex types, for languages like C++/D, Rust and Go, Haskell, etc.

XVilka:
For this we need to think/discuss first how it should be done exactly @sivaramaaa

XVilka:
@sivaramaaa for your information - in tcc parser and SDB database some metainformation about type stored in "meta" field, which can be enhanced, e.g. used as a bitfield to store more different properties (you can grep `shlr/tcc/tccgen.c` for "meta" keyword to see the code creating it).

thestr4ng3r:
About C++ types, currently I am implementing the anal classes completely separate from types to be saved in sdb under the `anal/classes` namespace. The reason is that classes are quite independent of actual types in many cases (static methods, vtables, rtti, etc.).
The idea would be to later add references between types (structs) and classes to for example describe the memory layout of a class.
However I am open for better suggestions if anybody has a better idea.

PR #16723, opened 2020-04-26
Types API

HoundThe:

Based on #16624 
This is first rough draft of the types API implementation, this is my first time doing something more with SDB so I probably missed some proper style/conventions there. 

Right now I've implemented it only for Enum to see if I am going in the right direction.

Got a few questions, should `RAnalBaseType` maybe have a name member with the type name?
I also made the base_type members a `RVector *` instead of `RVector`.
Should I do error handling on incorrect state of the type in Sdb? For example:
`enum.foo=firstCase,secondCase` but without `enum.foo.firstCase=0x1`

Example with debug print
```
[0x00001120]> "td enum foo { firstCase=1, secondCase=2,};"
[0x00001120]> debug foo
BaseType: 2
Case --- name: firstCase value: 1
Case --- name: secondCase value: 2
[0x00001120]> tk~foo
enum.foo=firstCase,secondCase
enum.foo.0x1=firstCase
enum.foo.0x2=secondCase
enum.foo.firstCase=0x1
enum.foo.secondCase=0x2
foo=enum
```


libr/anal/type.c
```diff
@@ -122,18 +122,108 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+
+	RAnalBaseType *base_type = malloc (sizeof(RAnalBaseType));
+	if (!base_type) {
+		return NULL;
+	}
+	base_type->kind = R_ANAL_BASE_TYPE_KIND_ENUM;
+	
+	RAnalBaseTypeEnum base_enum;
+	
+	char *key = sdb_fmt ("%s.%s", "enum", sanitized_name);
```
thestr4ng3r, line 158: For your info, we are trying to get rid of `sdb_fmt()` in the long run because it uses globals. We have been discussing a new design, but didn't yet agree on the concrete approach: https://github.com/radareorg/radare2/pulls?q=is%3Apr+r_strf+is%3Aunmerged

libr/anal/type.c
```diff
@@ -122,18 +122,108 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+		free (base_type);
+		r_vector_free (cases);
+		return NULL;
+	}
+
+	char *cur;
+	void *element;
+	sdb_aforeach (cur, members) {
+		const char *value = sdb_get (anal->sdb_types, sdb_fmt ("%s.%s.%s", "enum", sanitized_name, cur), NULL);
+		RAnalEnumCase cas = {.name = strdup (cur), .val = strtol (value, NULL, 16)};
```
thestr4ng3r, line 177: Regarding your question about error handling, here I would check for `value != NULL && *value`. I would say in general if the sdb contains garbage data, it's ok to also get garbage in C (for example if `value == "NotANumberLiteral"`, that the enum value silently becomes 0), but it must not lead to segfaults.

libr/anal/type.c
```diff
@@ -122,18 +122,108 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+		free (value);
+		
+		element = r_vector_push (cases, &cas); // returns null if no space available
```
thestr4ng3r, line 180: Just a small tip: you can pass `NULL` to the push, which will allocate the new element, but not initialize it. Then you can initialize it in-place and avoid a memcpy. Whether you want to do that or not depends on the case.

libr/include/r_anal.h
```diff
@@ -215,6 +215,33 @@ enum {
...
+} RAnalEnumCase;
+
+typedef enum {
+	R_ANAL_BASE_TYPE_KIND_STRUCT,
+	R_ANAL_BASE_TYPE_KIND_UNION,
+	R_ANAL_BASE_TYPE_KIND_ENUM,
+} RAnalBaseTypeKind;
+
+typedef struct r_anal_base_type_struct {
+	RVector *members; // describe the struct in some way
```
thestr4ng3r, line 230: Any particular reason you made it `RVector *`? It's just a very minor optimization, but it's usually better to not have a pointer and use init/fini.

libr/include/r_anal.h
```diff
@@ -215,6 +215,33 @@ enum {
...
+
+typedef struct r_anal_base_type_enum {
+	RVector/*<RAnalEnumCase*/ *cases; // list of all the enum cases
+} RAnalBaseTypeEnum;
+
+typedef struct r_anal_base_type {
+	RAnalBaseTypeKind kind;
+	union {
+		RAnalBaseTypeStruct ztruct;
+		RAnalBaseTypeEnum inum;
```
thestr4ng3r, line 241: Those zstruct and inum names were more of a joke tbh, we should probably find better names for them. Maybe `struct_data` and `enum_data` or something.

thestr4ng3r:
> should `RAnalBaseType` maybe have a name member with the type name?

Maybe. I think it will depend on how these structs will be used in the rest of the code, so it's a bit hard to tell right now. I would prefer to avoid having the name there since essentially it will just be a strdup of the string you already passed to the function in many cases. I think we should not add it yet and see if it is necessary later.

libr/anal/type.c
radare: the r_strf PR should be finished and merged

libr/anal/type.c
```diff
@@ -122,18 +122,108 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	if (!base_type) {
+		return NULL;
+	}
+	base_type->kind = R_ANAL_BASE_TYPE_KIND_ENUM;
+	
+	RAnalBaseTypeEnum base_enum;
+	
+	char *key = sdb_fmt ("%s.%s", "enum", sanitized_name);
+	char *members = sdb_get (anal->sdb_types, key, NULL);
+	RVector *cases = r_vector_new (sizeof(RAnalEnumCase), enum_type_fini, NULL);
```
radare, line 160: missing space before (

libr/anal/type.c
```diff
@@ -122,18 +122,108 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
 }
+
+static void enum_type_fini(void *e, void *user) {
+	(void)user;
+	RAnalEnumCase *cas = e;
+	free ((char *)cas->name);
+}
+
+static RAnalBaseType *get_enum_type(RAnal *anal, const char *sanitized_name) {
+	assert (sanitized_name != NULL);
```
radare, line 148: use r_return_val_if_fail() instead of assert

libr/anal/type.c
HoundThe: Why *value? Wouldn't an empty string result in strtol returning 0 aswell?

libr/include/r_anal.h
HoundThe: Just for an ease of use, because all the functions worked with `RVector *`, changed it back to a non pointer

libr/include/r_anal.h
HoundThe: Ye, I've figured, but I didn't come up with a better name so I've kept that. Changed it to what you've suggested now.

libr/anal/type.c
thestr4ng3r: It would, yes. Not checking *value is ok too.

libr/anal/type.c
```diff
@@ -122,18 +122,107 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
 static int typecmp(const void *a, const void *b) {
 	char *type1 = (char *) a;
```
radare, line 123: this is const char * remove the cast

libr/anal/type.c
```diff
@@ -122,18 +122,107 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
 }
+
+static void enum_type_fini(void *e, void *user) {
+	(void)user;
+	RAnalEnumCase *cas = e;
+	free ((char *)cas->name);
+}
+
+static RAnalBaseType *get_enum_type(RAnal *anal, const char *sanitized_name) {
+	r_return_val_if_fail(sanitized_name != NULL, NULL);
```
radare, line 148: missing space

libr/anal/type.c
radare: remove the assert

libr/anal/type.c
```diff
@@ -122,18 +122,107 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	assert (sanitized_name != NULL);
+
+	RAnalBaseType *base_type = malloc (sizeof(RAnalBaseType));
```
radare: missing spaces before (

libr/anal/type.c
HoundThe: I merged master, but I don't see any r_strf(), only r_str_newf() not sure if you meant that, but I changed to using that for now.

radare: then use r_str_newf() in the heap for now, its the only safe alternative. also if "enum" is always "enum", dont format it, just put it in the string

libr/anal/type.c
```diff
@@ -122,18 +122,107 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	}
+	base_type->kind = R_ANAL_BASE_TYPE_KIND_ENUM;
+	
+	RAnalBaseTypeEnum base_enum;
+	
+	char *key = sdb_fmt ("%s.%s", "enum", sanitized_name);
+	char *members = sdb_get (anal->sdb_types, key, NULL);
+
+	RVector cases;
+	r_vector_init(&cases, sizeof(RAnalEnumCase), enum_type_fini, NULL);
```
radare, line 162: missing space bfore (

libr/anal/type.c
```diff
@@ -122,18 +122,107 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	base_enum.cases = cases;
+	base_type->enum_data = base_enum;
+
+	return base_type;
+}
+// returns NULL if name is not found or 
+R_API RAnalBaseType *r_anal_get_base_type(RAnal *anal, const char *name) {
+	char *name_sanitized = r_str_sanitize_sdb_key (name);
+
+	char *type = sdb_get (anal->sdb_types, name_sanitized, NULL);
```
radare, line 207: use sdb_const_get, no need to heap it

libr/anal/type.c
```diff
@@ -122,18 +122,107 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	if (!type || !(strcmp (type, "enum") || strcmp (type, "struct") || strcmp (type, "union"))) {
+		free (type);
+		return NULL;
+	}
+	// Taking advantage that all 3 types start with distinct letter
+	// because the strcmp condition guarantees that only those will get to this flow
+	RAnalBaseType *base_type;
+	
+	switch (type[0]) {
+		case 's': // struct
```
radare, line 219: wrong indent, cases go in the same level as switch

libr/anal/type.c
```diff
@@ -122,18 +122,107 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+			break;
+		case 'e': // enum
+			base_type = get_enum_type (anal, name_sanitized);
+			// DEBUG print, TODO not forget to remove
+				r_cons_printf ("BaseType: %d\n", base_type->kind);
```
radare, line 224: wrong indentation too

libr/include/r_anal.h
```diff
@@ -215,6 +215,33 @@ enum {
 	R_ANAL_DIFF_TYPE_UNMATCH = 'u'
 };
 
+typedef struct enum_case {
+	const char *name;
```
radare, line 219: i think this should be char *

libr/anal/type.c
```diff
@@ -122,18 +122,107 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	r_vector_init(&cases, sizeof(RAnalEnumCase), enum_type_fini, NULL);
+
+	if (!r_vector_reserve (&cases, (size_t) sdb_alen (members))) {
+		free (base_type);
+		r_vector_free (&cases);
+		return NULL;
+	}
+
+	char *cur;
+	void *element;
```
radare, line 171: define in the same line that you assign i t

libr/anal/type.c
```diff
@@ -122,18 +122,107 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	sdb_aforeach (cur, members) {
+		const char *value = sdb_get (anal->sdb_types, sdb_fmt ("%s.%s.%s", "enum", sanitized_name, cur), NULL);
+		RAnalEnumCase cas = {.name = strdup (cur), .val = strtol (value, NULL, 16)};
+		free (value);
+		
+		element = r_vector_push (&cases, &cas); // returns null if no space available
+		if (!element){
+			eprintf ("Allocation failed\n");
```
radare, line 186: no need to eprintf when malloc fails

libr/anal/type.c
```diff
@@ -122,18 +122,107 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+
+	char *cur;
+	void *element;
+	sdb_aforeach (cur, members) {
+		const char *value = sdb_get (anal->sdb_types, sdb_fmt ("%s.%s.%s", "enum", sanitized_name, cur), NULL);
+		RAnalEnumCase cas = {.name = strdup (cur), .val = strtol (value, NULL, 16)};
+		free (value);
+		
+		element = r_vector_push (&cases, &cas); // returns null if no space available
+		if (!element){
```
radare, line 185: missing space before (

libr/anal/type.c
```diff
@@ -122,18 +122,107 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+				RAnalEnumCase *it;
+				r_vector_foreach (&base_type->enum_data.cases, it) {
+					r_cons_printf ("Case --- name: %s value: %d\n", it->name, it->val);
+				}
+			break;
+		case 'u': // union
+			break;
+	}
+
+	free(type);
```
radare, line 234: missing space before (

libr/anal/type.c
HoundThe: Oh, that's not the code I wrote, but I guess I can still remove it

libr/anal/type.c
```diff
@@ -122,18 +122,119 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	
+	switch (type[0]) {
+	case 's': // struct
+		break;
+	case 'e': // enum
+		base_type = get_enum_type (anal, name_sanitized);
+		// DEBUG print, TODO not forget to remove
+		r_cons_printf ("BaseType: %d\n", base_type->kind);
+		RAnalEnumCase *it;
+		r_cons_printf("Cases:\n");
```
radare, line 229: space before (

libr/anal/type.c
```diff
@@ -122,18 +122,119 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+		r_vector_foreach (&base_type->enum_data.cases, it) {
+			r_cons_printf ("\tname: %s, value: %d\n", it->name, it->val);
+		}
+		break;
+	case 'u': // union
+		break;
+	}
+
+	free(name_sanitized);
```
radare, line 238: missin space

libr/anal/type.c
```diff
@@ -122,18 +122,119 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	case 's': // struct
+		break;
+	case 'e': // enum
+		base_type = get_enum_type (anal, name_sanitized);
+		// DEBUG print, TODO not forget to remove
+		r_cons_printf ("BaseType: %d\n", base_type->kind);
+		RAnalEnumCase *it;
+		r_cons_printf("Cases:\n");
+		r_vector_foreach (&base_type->enum_data.cases, it) {
+			r_cons_printf ("\tname: %s, value: %d\n", it->name, it->val);
```
radare, line 231: dont use tabs in output it will break visual panels

libr/anal/type.c
```diff
@@ -122,18 +122,119 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	return base_type;
+}
+
+// returns NULL if name is not found or any failure happened
+R_API RAnalBaseType *r_anal_get_base_type(RAnal *anal, const char *name) {
+	r_return_val_if_fail (name != NULL, NULL);
+	r_return_val_if_fail (anal != NULL, NULL);
+
+	char *name_sanitized = r_str_sanitize_sdb_key (name);
+
```
radare, line 209: no need for this empty line

libr/anal/type.c
```diff
@@ -122,18 +122,119 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	if (!r_vector_reserve (&cases, (size_t) sdb_alen (members))) {
+		free (base_type);
+		r_vector_free (&cases);
+		return NULL;
+	}
+
+	char *cur;
+	sdb_aforeach (cur, members) {
+		char *val_key = r_str_newf ("enum.%s.%s", sanitized_name, cur);
+		char *value = sdb_const_get (anal->sdb_types, val_key, NULL);
```
radare, line 177: move after the if (!val_key) check

libr/anal/type.c
```diff
@@ -122,18 +122,119 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+static void enum_type_fini(void *e, void *user) {
+	(void)user;
+	RAnalEnumCase *cas = e;
+	free ((char *)cas->name);
+}
+
+static RAnalBaseType *get_enum_type(RAnal *anal, const char *sanitized_name) {
+	r_return_val_if_fail (sanitized_name != NULL, NULL);
+
+	RAnalBaseType *base_type = malloc (sizeof (RAnalBaseType));
```
radare, line 150: use R_NEW0()

libr/anal/type.c
```diff
@@ -122,18 +122,119 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+
+	base_enum.cases = cases;
+	base_type->enum_data = base_enum;
+
+	return base_type;
+}
+
+// returns NULL if name is not found or any failure happened
+R_API RAnalBaseType *r_anal_get_base_type(RAnal *anal, const char *name) {
+	r_return_val_if_fail (name != NULL, NULL);
```
radare, line 205: != NULL is implicit just (name, NULL)

libr/anal/type.c
```diff
@@ -122,18 +122,119 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	r_return_val_if_fail (anal != NULL, NULL);
+
+	char *name_sanitized = r_str_sanitize_sdb_key (name);
+
+	char *type = sdb_const_get (anal->sdb_types, name_sanitized, NULL);
+
+	// Right now just types: struct, enum, union are supported
+	if (!type || !(strcmp (type, "enum") || strcmp (type, "struct") || strcmp (type, "union"))) {
+		free(name_sanitized);
```
radare, line 214: missing space

libr/anal/type.c
HoundThe: What are the benefits of such macros instead of callocing directly?

libr/anal/type.c
HoundThe: It's just temporary print to see the function result for development

HoundThe:
Finally had time to move futher with the PR. I've added Unions and Structs, There is a lot of duplicity now, I'll refactor it when there will be settled functionality. Right now union and struct have members with `name` and `type`, I see there is more information in the Sdb struct format, but I don't really know what does the rest of the things mean.

Examples:
```
[0x00001080]> "td union kappa {int bar;int cow;};"
[0x00001080]> tk~kappa
kappa=union
union.kappa=bar,cow
union.kappa.!size=32
union.kappa.bar=int32_t,0,0
union.kappa.bar.meta=0
union.kappa.cow=int32_t,0,0
union.kappa.cow.meta=0
[0x00001080]> debug kappa
BaseType: 1
Cases:
	int32_t bar;
	int32_t cow;
```

```
[0x00001080]> "td struct kappa {int bar;int cow;};"
[0x00001080]> tk~kappa
kappa=struct
struct.kappa=bar,cow
struct.kappa.!size=64
struct.kappa.bar=int32_t,0,0
struct.kappa.bar.meta=0
struct.kappa.cow=int32_t,4,0
struct.kappa.cow.meta=0
[0x00001080]> debug kappa
BaseType: 0
Members:
	int32_t bar;
	int32_t cow;
```

XVilka:
I suggest also to add:
size (for bitfields)
offset
alignment

libr/anal/type.c
thestr4ng3r: They are a bit shorter and more expressive. If you don't want to pre-initialize with 0, you can also use `R_NEW()`.

thestr4ng3r:
@XVilka we shouldn't add stuff here at this time that is not yet in sdb.

From what I can tell from the code, given this:
```
struct.kappa.cow=int32_t,4,0
```
The 4 is the offset in bytes and the 0 is an optional count if the type was an array, see this code:
https://github.com/radareorg/radare2/blob/21faee260a7be9d20b9143258cc8984a09e4e5de/libr/util/utype.c#L216-L227

In my opinion, storing the array size like this is bad design, instead I would rather just make the type string something like `int32_t[4]`, but for now I'd say lets deserialize it for the API as-is.

About the .meta thing, I don't quite know its purpose myself, it seems tcc related and probably not relevant for the API:
https://github.com/radareorg/radare2/blob/21faee260a7be9d20b9143258cc8984a09e4e5de/shlr/tcc/tccgen.c#L1179-L1184
@XVilka Do you know more about this?

By the way, a good way to find usages of this sdb stuff is to grep for the format strings, e.g. `git grep '"struct\.'`

XVilka:
`meta` is used by tcc mostly, yes - it stores the internal type of the member.

kazarmy:
Guys, how is struct padding handled? https://en.cppreference.com/w/c/language/object#Alignment Does tcc handle this? Does that `!size` sdb key include padding as well?

thestr4ng3r:
Afaik currently not at all. All members are packed and if you want padding, you would have to manually add it as members.

XVilka:
It was done like that on purpose - the tcc recognizes packing and attributes. But in the reverse engineering tasks you often have to figure out the structure fields, padding, alignment by yourself. Following the compiler rules, which are different between compilers, might mislead the reverse engineer.

XVilka:
For the alignment in C types see:
- https://github.com/radareorg/radare2/issues/10031
- https://github.com/radareorg/radare2/issues/7710

kazarmy:
> It was done like that on purpose - the tcc recognizes packing and attributes. But in the reverse engineering tasks you often have to figure out the structure fields, padding, alignment by yourself. 

Ok sounds reasonable. Perhaps there can be a config var that determines whether r_types assumes structs as packed or padded by default.

> Following the compiler rules, which are different between compilers, might mislead the reverse engineer.

There seems to be only one best way to do alignment-related padding if structure member order has to be maintained, and I think compilers are required to preserve that order: https://en.cppreference.com/w/c/language/struct "Within a struct object, addresses of its elements (and the addresses of the bit field allocation units) increase in order in which the members were defined.".

libr/anal/type.c
```diff
@@ -122,18 +122,266 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+		free (type_key);
+
+		if (!values) {
+			free (base_type);
+			r_vector_fini (&members);
+			return NULL;
+		}
+		char *value = sdb_anext (values, NULL);
+		RAnalUnionMember cas = {.name = strdup (cur), .type = strdup (value)};
+		free(value);
```
radare, line 311: missing space before (

libr/anal/type.c
```diff
@@ -122,18 +122,266 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+		break;
+	case 'u': // union
+	{
+		base_type = get_union_type (anal, name_sanitized);
+		// DEBUG print, TODO not forget to remove
+		r_cons_printf ("BaseType: %d\n", base_type->kind);
+		RAnalUnionMember *it;
+		r_cons_printf ("Cases:\n");
+		r_vector_foreach (&base_type->union_data.members, it) {
+			r_cons_printf ("\t%s %s;\n", it->type, it->name);
```
radare, line 379: don't use tabs in the output it will break visual/panels/..

libr/anal/type.c
```diff
@@ -122,18 +122,266 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	switch (type[0]) {
+	case 's': // struct
+	{
+		base_type = get_struct_type (anal, name_sanitized);
+		// DEBUG print, TODO not forget to remove
+		r_cons_printf ("BaseType: %d\n", base_type->kind);
+		RAnalStructMember *it;
+		r_cons_printf ("Members:\n");
+		r_vector_foreach (&base_type->struct_data.members, it) {
+			r_cons_printf ("\t%s %s;\n", it->type, it->name);
```
radare, line 355: no tabs

libr/anal/type.c
```diff
@@ -122,18 +122,266 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	char *type = sdb_const_get (anal->sdb_types, name_sanitized, NULL);
+
+	// Right now just types: struct, enum, union are supported
+	if (!type || !(strcmp (type, "enum") || strcmp (type, "struct") || strcmp (type, "union"))) {
+		free (name_sanitized);
+		return NULL;
+	}
+	// Taking advantage that all 3 types start with distinct letter
+	// because the strcmp condition guarantees that only those will get to this flow
+	RAnalBaseType *base_type;
```
radare, line 344: = NULL;

libr/anal/type.c
```diff
@@ -122,18 +122,266 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	}
+
+	base_enum.cases = cases;
+	base_type->enum_data = base_enum;
+
+	return base_type;
+}
+
+static RAnalBaseType *get_struct_type(RAnal *anal, const char *sanitized_name) {
+	r_return_val_if_fail (sanitized_name, NULL);
```
radare, line 204: && anal

libr/anal/type.c
```diff
@@ -122,18 +122,266 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
 }
+
+static void enum_type_fini(void *e, void *user) {
+	(void)user;
+	RAnalEnumCase *cas = e;
+	free ((char *)cas->name);
+}
+
+static RAnalBaseType *get_enum_type(RAnal *anal, const char *sanitized_name) {
+	r_return_val_if_fail (sanitized_name, NULL);
```
radare, line 148: && name

HoundThe:
Sorry for the delay, I had to finish last uni projects. I've tackled rest of the comments. 

I'll add tests for this later today, unless you want to add something like the offset for struct members?

XVilka:
Offset can be helpful, calculating it only once upon types loading or changing architecture. Offset and size are used during the type analysis on converting immediates to offsets: `mov eax, [ebx + 0x20]`

libr/anal/type.c
```diff
@@ -122,18 +122,238 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+		free (base_type);
+		r_vector_free (&cases);
+		return NULL;
+	}
+
+	char *cur;
+	sdb_aforeach (cur, members) {
+		char *val_key = r_str_newf ("enum.%s.%s", sanitized_name, cur);
+		if (!val_key) {
+			free (base_type);
```
radare, line 178: return free_locals(base_type, &cases); or use goto to avoid repeating the same frees before each return statement

libr/anal/type.c
```diff
@@ -122,18 +122,238 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	}
+
+	base_enum.cases = cases;
+	base_type->enum_data = base_enum;
+
+	return base_type;
+}
+
+static RAnalBaseType *get_struct_type(RAnal *anal, const char *sanitized_name) {
+	r_return_val_if_fail (sanitized_name && anal, NULL);
```
radare, line 204: swap the order of the &&

libr/anal/type.c
HoundThe: So I can call strcmp directly with a and b?

radare:
Yes

HoundThe:
So I've added offsets to struct members, if it's okay I'll add unit tests for the functions.

```c
typedef struct struct_member {
	char *name;
	char *type;
	const int offset;
} RAnalStructMember;
```

thestr4ng3r:
Looks nice, I am looking forward to the tests!

XVilka:
It breaks Windows build https://ci.appveyor.com/project/radareorg/radare2/builds/32827163/job/u3fd2vwa3hejgyto

XVilka:
Windows build is still broken https://ci.appveyor.com/project/radareorg/radare2/builds/32837405/job/x8c3247k54nkdojd

libr/anal/type.c
```diff
@@ -120,20 +120,258 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	RAnalBaseType *base_type = R_NEW0 (RAnalBaseType);
+	if (!base_type) {
+		return NULL;
+	}
+	base_type->kind = R_ANAL_BASE_TYPE_KIND_ENUM;
+
+	RAnalBaseTypeEnum base_enum;
+	char *key = r_str_newf ("enum.%s", sanitized_name);
+	if (!key) {
+		free (base_type);
```
radare, line 171: goto error

libr/include/r_anal.h
```diff
@@ -215,6 +215,49 @@ enum {
...
+	char *type;
+	const int offset;
+} RAnalStructMember;
+
+typedef struct union_member {
+	char *name;
+	char *type;
+} RAnalUnionMember;
+
+typedef enum {
```
radare, line 234: what about naming it RAnalType. instead of the BaseType? do that conflicts? why is KIND needed? this enum looks too long

libr/include/r_anal.h
```diff
@@ -215,6 +215,49 @@ enum {
 	R_ANAL_DIFF_TYPE_UNMATCH = 'u'
 };
 
+typedef struct enum_case {
+	char *name;
+	const int val;
+} RAnalEnumCase;
+
+typedef struct struct_member {
```
radare, line 223: we use r_anal_struct_member_t

libr/include/r_anal.h
```diff
@@ -215,6 +215,49 @@ enum {
 	R_ANAL_DIFF_TYPE_UNMATCH = 'u'
 };
 
+typedef struct enum_case {
```
radare, line 218: same here

libr/anal/type.c
```diff
@@ -120,20 +120,258 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	free (sdb_members);
+	r_vector_fini (&members);
+	return NULL;
+}
+
+// returns NULL if name is not found or any failure happened
+R_API RAnalBaseType *r_anal_get_base_type(RAnal *anal, const char *name) {
+	r_return_val_if_fail (anal && name, NULL);
+
+	char *name_sanitized = r_str_sanitize_sdb_key (name);
```
radare, line 351: lexically should be sanitized_name, but you can use  `sname` which is shorter and looks readable to me

test/unit/test_anal_types.c
```diff
@@ -0,0 +1,139 @@
...
+	mu_assert_streq (member->type, "int32_t", "Incorrect type for union member");
+	mu_assert_streq (member->name, "cow", "Incorrect name for union member");
+
+	r_vector_fini (&base->union_data.members);
+	r_anal_free (anal);
+	free (base);
+	mu_end;
+}
+
+bool test_anal_get_base_type_enum() {
```
radare, line 88: make it static

libr/anal/type.c
```diff
@@ -120,20 +120,258 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
...
+	return base_type;
+
+error:
+	free (base_type);
+	free (members);
+	r_vector_fini (&cases);
+	return NULL;
+}
+
+static RAnalBaseType *get_struct_type(RAnal *anal, const char *sanitized_name) {
```
radare, line 215: see? santized name is used here, below its using name_sanitized. better be consistent and follow the same naming across the code

libr/include/r_anal.h
thestr4ng3r: About the name, see https://hackmd.io/kKvaIscWQk6_zFLYcLYWnw
There must be a clear distinction between "C-types" and "base types" with the current design of the database. The "Kind" comes from the typedefd name of the enum itself.

libr/anal/type.c
HoundThe: Sure, I've went with `sname` then

libr/include/r_anal.h
kazarmy: > About the name, see https://hackmd.io/kKvaIscWQk6_zFLYcLYWnw

Why should the Ghidra solution be "easier to do it if the above solution with base types and c types is implemented already"?

libr/include/r_anal.h
HoundThe: r_anal_struct_member_t as a Typedef or the stuct name? If struct name then the `_t` suffix shouldn't be there?

thestr4ng3r:
```c
typedef struct r_anal_struct_member_t { ... } RAnalStructMember
```

libr/include/r_anal.h
thestr4ng3r: Because then the entire code will already use C structs everywhere as opposed to strings. Refactoring these C structs into a common pool would be easier than doing the same from strings.

libr/include/r_anal.h
HoundThe: But why the _t ? doesn't the suffix mean a typedef? but its a struct name and not typedef

libr/anal/type.c
```diff
@@ -67,6 +67,17 @@ static void __save_parsed_type_size(RAnal *anal, const char *parsed) {
...
 }
 
+static char *get_type_data(Sdb *sdb_types, char *type, char *sname) {
+	char *key = r_str_newf ("%s.%s", type, sname);
+	if (!key) {
+		return NULL;
+	}
+	char *members = sdb_get (sdb_types, key, NULL);
+	free (key);
+
```
radare, line 77: empty line

libr/anal/type.c
```diff
@@ -67,6 +67,17 @@ static void __save_parsed_type_size(RAnal *anal, const char *parsed) {
 	}
 }
 
+static char *get_type_data(Sdb *sdb_types, char *type, char *sname) {
```
radare, line 70: type and sname are const char *

libr/include/r_anal.h
thestr4ng3r: Yes, some projects do consider it to mean "typedef", but we don't. We want to have the typedefd name to be as nice as possible and the struct-name with "\_t" to indicate "raw type" (or something ¯\\\_(ツ)\_/¯).

PR #16724, opened 2020-04-26
Fix assembling of xchg and bswap

atodekangae:

**Detailed description**

Previously, rasm2 incorrectly assembled `bswap` and `xchg` for x86-64 as shown below.

```
$ rasm2 -b 64 -
bswap r15
480fcf
bswap r15d
0fcf
xchg al, dl
92
xchg dl, al
92
xchg ax, dx
92
xchg rax, rdx
92
xchg rdx, rbx
87da
xchg rbx, rdx
87d3
xchg r8, r15
97
xchg r15, r8
97
xchg r8d, r15d
97
xchg r15d, r8d
97
```

While the expected behavior is as follows.

```
$ rasm2 -b 64 -
bswap r15
490fcf
bswap r15d
410fcf
xchg al, dl
86c2
xchg dl, al
86d0
xchg ax, dx
6692
xchg rax, rdx
4892
xchg rdx, rbx
4887d3
xchg rbx, rdx
4887da
xchg r8, r15
4d87c7
xchg r15, r8
4d87f8
xchg r8d, r15d
4587c7
xchg r15d, r8d
4587f8
```

This pull request fixes this problem.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/test/db/asm/x86_32
+++ b/test/db/asm/x86_32
@@ -2147,25 +2147,25 @@ aB "xadd byte [eax], al" 0fc000
 aB "xadd dword [eax], eax" 0fc100
 aB "xchg byte [eax], al" 8600
 a "xchg dword [eax], eax" 8700
-a "xchg al, dl" 86c2
-a "xchg dl, al" 86d0
-a "xchg ax, dx" 6692
+ad "xchg al, dl" 86d0
+ad "xchg dl, al" 86c2
+ad "xchg ax, dx" 6692
 a "xchg dx, ax" 6692
-a "xchg ah, dh" 86e6
-a "xchg dh, ah" 86f4
+ad "xchg ah, dh" 86f4
+ad "xchg dh, ah" 86e6
 a "xchg eax, eax" 90
-a "xchg eax, ebp" 95
-a "xchg eax, ebx" 93
-a "xchg eax, ebx" 93
-a "xchg eax, ecx" 91
-a "xchg eax, edi" 97
-a "xchg eax, edx" 92
-a "xchg eax, esi" 96
-a "xchg eax, esp" 94
-a "xchg ebx, ecx" 87d9
-a "xchg ecx, ebp" 87cd
-a "xchg ecx, ebx" 87cb
-a "xchg ecx, ecx" 87c9
+ad "xchg eax, ebp" 95
+ad "xchg eax, ebx" 93
+ad "xchg eax, ebx" 93
+ad "xchg eax, ecx" 91
+ad "xchg eax, edi" 97
+ad "xchg eax, edx" 92
+ad "xchg eax, esi" 96
+ad "xchg eax, esp" 94
+ad "xchg ebx, ecx" 87cb
+ad "xchg ecx, ebp" 87e9
+ad "xchg ecx, ebx" 87d9
+ad "xchg ecx, ecx" 87c9
 a "xgetbv" 0f01d0
 a "xlatb" d7
 a "xor al, 0" 3400
--- a/test/db/asm/x86_32
+++ b/test/db/asm/x86_32
@@ -2147,6 +2147,12 @@ aB "xadd byte [eax], al" 0fc000
 aB "xadd dword [eax], eax" 0fc100
 aB "xchg byte [eax], al" 8600
 a "xchg dword [eax], eax" 8700
+a "xchg al, dl" 86c2
+a "xchg dl, al" 86d0
+a "xchg ax, dx" 6692
+a "xchg dx, ax" 6692
+a "xchg ah, dh" 86e6
+a "xchg dh, ah" 86f4
 a "xchg eax, eax" 90
 a "xchg eax, ebp" 95
 a "xchg eax, ebx" 93
@@ -2156,9 +2162,9 @@ a "xchg eax, edi" 97
 a "xchg eax, edx" 92
 a "xchg eax, esi" 96
 a "xchg eax, esp" 94
-a "xchg ebx, ecx" 87cb
-a "xchg ecx, ebp" 87e9
-a "xchg ecx, ebx" 87d9
+a "xchg ebx, ecx" 87d9
+a "xchg ecx, ebp" 87cd
+a "xchg ecx, ebx" 87cb
 a "xchg ecx, ecx" 87c9
 a "xgetbv" 0f01d0
 a "xlatb" d7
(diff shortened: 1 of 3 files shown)


test/db/asm/x86_32
```diff
@@ -2156,9 +2162,9 @@ a "xchg eax, edi" 97
 a "xchg eax, edx" 92
 a "xchg eax, esi" 96
 a "xchg eax, esp" 94
-a "xchg ebx, ecx" 87cb
-a "xchg ecx, ebp" 87e9
-a "xchg ecx, ebx" 87d9
+a "xchg ebx, ecx" 87d9
+a "xchg ecx, ebp" 87cd
+a "xchg ecx, ebx" 87cb
```
atodekangae, line 2167: Although the original output was not incorrect per Intel's Instruction Set Reference (`xchg r32, r32` can be interpreted either as an instance of `xchg r32, r/m32` or as `xchg r/m32, r32`), I fixed these lines and the corresponding code to make it in line with the output of other assemblers (NASM and FASM).

ret2libc: Mmmm they do seem like different instructions to me. Why the change? Wasn't it working already before?

atodekangae: The primary reason for the change is to make the resulting machine code the same as the one produced by other assemblers, namely, NASM and FASM.

Technically, the original output was not incorrect as I stated. It is about which to choose when there are two options. Previously, when encoding `xchg r32, r32`, rasm2 chose the encoding of `xchg r/m32, r32`, whose `Op/En` is `MR`, which means that the left register operand is encoded in the `r/m` part of `ModR/M` and the right in the `reg` part. However, NASM and FASM choose `xchg r32, r/m32` when encoding `xchg r32, r32`. I thought it was a good idea to follow the assemblers widely in use.

The code making the choice is as follows. Flipping 0 and 1 would undo the change, which I can do if desired.

https://github.com/atodekangae/radare2/blob/2f5b8f1d672de7f1933c865293b51e0aac83222d/libr/asm/p/asm_x86_nz.c#L2791-2792

ret2libc: @radare opinions?

radare: uhm the swap of arguments in test and exchagne instructions on x86 assemnblers and disassemblers have been always a known issue. the thing is.. this change seems to return a different output than capstone, 
``
$ rasm2 -a x86  -d 87cd
xchg ebp, ecx
```

File: test/db/asm/x86_64 [hunk 2]
```diff
@@ -956,6 +956,24 @@ a "dec rdi" 48ffcf
...
+a "xchg rax, r8" 4990
+a "xchg r8, rax" 4990
+a "xchg rdx, rbx" 4887d3
+a "xchg rbx, rdx" 4887da
+a "xchg r8, r15" 4d87c7
+a "xchg r15, r8" 4d87f8
+a "xchg r8d, r15d" 4587c7
+a "xchg r15d, r8d" 4587f8
+a "xchg rdx, r8" 4987d0
+a "xchg r15, rdx" 4c87fa
```
radare, line 976: looks wrong to me:
```
$  rasm2 -a x86 -b64  -d 4c87fa
xchg rdx, r15
```

atodekangae:
Thank you for the feedback. I made a change so that encoding is consistent with Capstone and it roundtips.

atodekangae:
Thank you for the feedback. I made a change so that encoding is consistent with Capstone and it roundtrips.

radare:
Looks good now! thanks!

test/db/asm/x86_64
```diff
@@ -956,6 +956,24 @@ a "dec rdi" 48ffcf
...
+a "bswap rax" 480fc8
+a "bswap r15" 490fcf
+a "bswap eax" 0fc8
+a "bswap r15d" 410fcf
+ad "xchg eax, r8d" 4190
+a "xchg r8d, eax" 4190
+ad "xchg rax, rdx" 4892
+a "xchg rdx, rax" 4892
+ad "xchg rax, r8" 4990
+a "xchg r8, rax" 4990
```
radare, line 968: this doesnt roundtrip

atodekangae: `xchg rax, rdx` and `xchg rdx, rax` have exactly the same encoding, `4892`. So either of the two cannot roundtrip. The same goes for the pair `xchg rax, r8` and `xchg r8, rax`. In general, there is an encoding for the cases where one of the operands is `ax`, `eax`, or `rax`, where the order of the operands is lost. 

Following is an excerpt from the Instruction Set Reference Manual. As you see, the encoding does not retain the order if an `ax` register is present.

<table>
<tr>
<td>90+<em>rw</em></td>
<td>XCHG AX, <em>r16</em></td>
<td>O</td>
<td>Valid</td>
<td>Valid</td>
<td>Exchange <em>r16</em> with AX.</td></tr>
<tr>
<td>90+<em>rw</em></td>
<td>XCHG <em>r16</em>, AX</td>
<td>O</td>
<td>Valid</td>
<td>Valid</td>
<td>Exchange AX with <em>r16.</em></td></tr>
<tr>
<td>90+<em>rd</em></td>
<td>XCHG EAX, <em>r32</em></td>
<td>O</td>
<td>Valid</td>
<td>Valid</td>
<td>Exchange <em>r32</em> with EAX.</td></tr>
<tr>
<td>REX.W + 90+<em>rd</em></td>
<td>XCHG RAX, <em>r64</em></td>
<td>O</td>
<td>Valid</td>
<td>N.E.</td>
<td>Exchange <em>r64</em> with RAX.</td></tr>
<tr>
<td>90+<em>rd</em></td>
<td>XCHG <em>r32</em>, EAX</td>
<td>O</td>
<td>Valid</td>
<td>Valid</td>
<td>Exchange EAX with <em>r32.</em></td></tr>
<tr>
<td>REX.W + 90+<em>rd</em></td>
<td>XCHG <em>r64</em>, RAX</td>
<td>O</td>
<td>Valid</td>
<td>N.E.</td>
<td>Exchange RAX with <em>r64.</em></td></tr>
</table>

atodekangae: As a side note, I noticed there is a test present to check if `xchg eax, eax` assembles to `90`. This does not roundtrip, but I kept the original behavior and test unchanged.

radare:
👍👍

PR #16725, opened 2020-04-26
Use rpvector for storing esil pc trace

pelijah:

**Detailed description**

The main idea here is to use RPVector for storing some parts of esil trace instead of sdb since they anyway are accessed by index.
Also by using that vector as cache for RAnalOp objects we can improve speed of type analysis (`aaft`). But I don't know how it can affect the esil emulaton. Also these changes breaks some commands for quering trace db.

Analysis time (`aaa`) of a 18MB binary: 3m10s vs 2m25s (master vs current branch).

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/anal/esil.c
+++ b/libr/anal/esil.c
@@ -96,6 +96,7 @@ R_API RAnalEsil *r_anal_esil_new(int stacksize, int iotrap, unsigned int addrsiz
 	r_anal_esil_interrupts_init (esil);
 	esil->sessions = r_list_newf (r_anal_esil_session_free);
 	esil->addrmask = genmask (addrsize - 1);
+	r_pvector_init (&esil->trace_vec, r_anal_op_free);
 	return esil;
 }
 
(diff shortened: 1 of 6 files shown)


pelijah:
Well caching should be optional. In that case it would be better to use RVector of `ut64 + RAnalOp*`. But there are other trace entries which are accessed by index. So eventually this will require more complicated struct.

radare:
please provide performance proof of this change.

btw, we should start to think about having test/bench to run on every commit and graph the performance gains/loss over time

pelijah:
> please provide performance proof of this change.

Existing one is not enough? Any specific binary?

thestr4ng3r:
Any specific binary with concrete commands and maybe a graph from a profiler would be nice.

pelijah:
`time radare2 -Q -qq -A /usr/bin/snap`

_master_: 65.6s
_current_: 59.4s

pelijah:
`radare2 -qq -c "aa;aac;aar;?t aaft" /usr/bin/snap`

_master_: 41.4s
_current_: 34.7s

radare:
Is this affecting aae too? Which other aa subcommands make use of this?

Can you rebase to make it green?

Goo work!

pelijah:
It fails 1 test.

radare:
Why close

pelijah:
Well I have found that it is not really possible to store trace events info in a vector. So I see no reason to continue work on this PR. And caching can be done in other way. I will try to do implement it later.

radare:
ok :)

PR #16738, opened 2020-04-28
Extend RCmdDesc with different types

ret2libc:

**Detailed description**

The idea is to use RCmdDesc to hold info about different kinds of commands. RCmdDesc will be structured in a tree, used to keep the structure of the commands as we have now. Moreover, some of RCmdDesc are also stored in a hashtable for quick retrieval when you need to execute a command.

To keep supporting existing command handlers, there will be different kinds of RCmdDesc.
* `R_CMD_DESC_TYPE_OLDINPUT`: this is used to store info about the existing handlers. Those handlers accept `(void *user, const char *input)` and they do the parsing of input themselves. In a distant future, we should probably convert all these to something else, but it will require time.

* `R_CMD_DESC_TYPE_ARGV`: this store info about the new kind of command handlers, that accept argc/argv. There will be one RCmdDesc of this type for each command e.g. `agf`, `afl`, `af`, `wx`, `wv8`, etc.

* `R_CMD_DESC_TYPE_INNER`: this will represent an inner node in the help tree. For example, when you do `?` you see that `a[?]    analysis commands`. But then if you do `a?` you see that there is another "a", which is `a   alias for aai - analysis information`. The first "a" will be an inner node, which does not do anything special apart from providing some help msg and group together other nodes. The second "a" will be a R_CMD_DESC_TYPE_ARGV with the right handler for that command.

**Test plan**

So far I have just added some unit tests and adapted a bit the existing code to make it compile, but apart from that everything should still be working.

### DIFF
--- a/libr/core/cmd_api.c
+++ b/libr/core/cmd_api.c
@@ -20,7 +20,7 @@ static bool cmd_desc_set_parent(RCmdDesc *cd, RCmdDesc *parent) {
 	return true;
 }
 
-static RCmdDesc *create_cmd_desc(RCmdDesc *parent, RCmdDescType type, const char *name, const char *help) {
+static RCmdDesc *create_cmd_desc(RCmdDesc *parent, RCmdDescType type, const char *name) {
 	RCmdDesc *res = R_NEW0 (RCmdDesc);
 	if (!res) {
 		return NULL;
@@ -29,12 +29,6 @@ static RCmdDesc *create_cmd_desc(RCmdDesc *parent, RCmdDescType type, const char
 	res->name = strdup (name);
 	if (!res->name) {
 		goto err;
-	}
-	if (help) {
-		res->help = strdup (help);
-		if (!res->help) {
-			goto err;
-		}
 	}
 	res->n_children = 0;
 	r_pvector_init (&res->children, (RPVectorFree)r_cmd_desc_free);
@@ -62,7 +56,7 @@ R_API RCmd *r_cmd_new () {
 		cmd->cmds[i] = NULL;
 	}
 	cmd->nullcallback = cmd->data = NULL;
-	cmd->root_cmd_desc = create_cmd_desc (NULL, R_CMD_DESC_TYPE_INNER, "", NULL);
+	cmd->root_cmd_desc = create_cmd_desc (NULL, R_CMD_DESC_TYPE_INNER, "");
 	cmd->ht_cmds = ht_pp_new0 ();
 	r_core_plugin_init (cmd);
 	r_cmd_macro_init (&cmd->macro);
@@ -939,10 +933,10 @@ R_API const char *r_cmd_parsed_args_cmd(RCmdParsedArgs *a) {
 
 /* RCmdDescriptor */
 
-R_API RCmdDesc *r_cmd_desc_argv_new(RCmd *cmd, RCmdDesc *parent, const char *name, const char *help, RCmdArgvCb cb) {
-	r_return_val_if_fail (cmd && parent && name && help && cb, NULL);
+R_API RCmdDesc *r_cmd_desc_argv_new(RCmd *cmd, RCmdDesc *parent, const char *name, RCmdArgvCb cb) {
+	r_return_val_if_fail (cmd && parent && name && cb, NULL);
 	r_return_val_if_fail (parent->type == R_CMD_DESC_TYPE_INNER, NULL);
-	RCmdDesc *res = create_cmd_desc (parent, R_CMD_DESC_TYPE_ARGV, name, help);
+	RCmdDesc *res = create_cmd_desc (parent, R_CMD_DESC_TYPE_ARGV, name);
 	if (!res) {
 		return NULL;
 	}
@@ -955,16 +949,16 @@ R_API RCmdDesc *r_cmd_desc_argv_new(RCmd *cmd, RCmdDesc *parent, const char *nam
 	return res;
 }
 
-R_API RCmdDesc *r_cmd_desc_inner_new(RCmd *cmd, RCmdDesc *parent, const char *name, const char *help) {
-	r_return_val_if_fail (cmd && parent && name && help, NULL);
+R_API RCmdDesc *r_cmd_desc_inner_new(RCmd *cmd, RCmdDesc *parent, const char *name) {
+	r_return_val_if_fail (cmd && parent && name, NULL);
 	r_return_val_if_fail (parent->type == R_CMD_DESC_TYPE_INNER, NULL);
-	return create_cmd_desc (parent, R_CMD_DESC_TYPE_INNER, name, help);
+	return create_cmd_desc (parent, R_CMD_DESC_TYPE_INNER, name);
 }
 
 R_API RCmdDesc *r_cmd_desc_oldinput_new(RCmd *cmd, RCmdDesc *parent, const char *name, RCmdCb cb) {
 	r_return_val_if_fail (cmd && parent && name && cb, NULL);
 	r_return_val_if_fail (parent->type == R_CMD_DESC_TYPE_INNER, NULL);
-	RCmdDesc *res = create_cmd_desc (parent, R_CMD_DESC_TYPE_OLDINPUT, name, NULL);
+	RCmdDesc *res = create_cmd_desc (parent, R_CMD_DESC_TYPE_OLDINPUT, name);
 	if (!res) {
 		return NULL;
 	}
@@ -980,7 +974,6 @@ R_API void r_cmd_desc_free(RCmdDesc *cd) {
 	if (!cd) {
 		return;
 	}
-	free (cd->help);
 	free (cd->name);
 	free (cd);
 }
--- a/libr/core/cmd_api.c
+++ b/libr/core/cmd_api.c
@@ -325,9 +325,11 @@ R_API int r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
 	RCorePlugin *cp;
 	char *exec_string = r_cmd_parsed_args_execstr (args);
 	r_list_foreach (cmd->plist, iter, cp) {
-		res = cp->call (cmd->data, exec_string);
-		if (res) {
-			break;
+		if (cp->call) {
+			res = cp->call (cmd->data, exec_string);
+			if (res) {
+				break;
+			}
 		}
 	}
 	free (exec_string);
--- a/libr/core/cmd_api.c
+++ b/libr/core/cmd_api.c
@@ -12,6 +12,38 @@ static int value = 0;
 #define NCMDS (sizeof (cmd->cmds)/sizeof(*cmd->cmds))
 R_LIB_VERSION (r_cmd);
 
+static bool cmd_desc_set_parent(RCmdDesc *cd, RCmdDesc *parent) {
+	r_return_val_if_fail (cd && parent && !cd->parent, false);
+	cd->parent = parent;
+	r_pvector_push (&parent->children, cd);
+	parent->n_children++;
+	return true;
+}
+
+static RCmdDesc *create_cmd_desc(RCmdDesc *parent, RCmdDescType type, const char *name, const char *help) {
+	RCmdDesc *res = R_NEW0 (RCmdDesc);
+	if (!res) {
+		return NULL;
+	}
+	res->type = type;
+	res->name = strdup (name);
+	if (!res->name) {
+		goto err;
+	}
+	if (help) {
+		res->help = strdup (help);
+		if (!res->help) {
+			goto err;
+		}
+	}
+	res->n_children = 0;
+	r_pvector_init (&res->children, (RPVectorFree)r_cmd_desc_free);
+	cmd_desc_set_parent (res, parent);
+	return res;
+err:
+	r_cmd_desc_free (res);
+	return NULL;
+}
 
 R_API void r_cmd_alias_init(RCmd *cmd) {
 	cmd->aliases.count = 0;
@@ -30,6 +62,8 @@ R_API RCmd *r_cmd_new () {
 		cmd->cmds[i] = NULL;
 	}
 	cmd->nullcallback = cmd->data = NULL;
+	cmd->root_cmd_desc = create_cmd_desc (NULL, R_CMD_DESC_TYPE_INNER, "", NULL);
+	cmd->ht_cmds = ht_pp_new0 ();
 	r_core_plugin_init (cmd);
 	r_cmd_macro_init (&cmd->macro);
 	r_cmd_alias_init (cmd);
@@ -53,8 +87,37 @@ R_API RCmd *r_cmd_free(RCmd *cmd) {
 			R_FREE (cmd->cmds[i]);
 		}
 	}
+	r_cmd_desc_free (cmd->root_cmd_desc);
 	free (cmd);
 	return NULL;
+}
+
+R_API RCmdDesc *r_cmd_get_root(RCmd *cmd) {
+	return cmd->root_cmd_desc;
+}
+
+static bool is_valid_desc(RCmdDesc *cd, const char *name) {
+	return cd->type == R_CMD_DESC_TYPE_OLDINPUT ||
+		(cd->type == R_CMD_DESC_TYPE_ARGV && !strcmp (cd->name, name));
+}
+
+R_API RCmdDesc *r_cmd_get_desc(RCmd *cmd, const char *cmd_identifier) {
+	r_return_val_if_fail (cmd && cmd_identifier, NULL);
+	char *cmdid = strdup (cmd_identifier);
+	char *end_cmdid = cmdid + strlen (cmdid);
+	RCmdDesc *res = NULL;
+	// match longer commands first
+	while (*cmdid) {
+		res = ht_pp_find (cmd->ht_cmds, cmdid, NULL);
+		r_warn_if_fail (!res || res->type == R_CMD_DESC_TYPE_OLDINPUT || res->type == R_CMD_DESC_TYPE_ARGV);
+		if (res && is_valid_desc (res, cmd_identifier)) {
+			goto out;
+		}
+		*(--end_cmdid) = '\0';
+	}
+out:
+	free (cmdid);
+	return res;
 }
 
 R_API char **r_cmd_alias_keys(RCmd *cmd, int *sz) {
@@ -173,7 +236,7 @@ R_API int r_cmd_set_data(RCmd *cmd, void *data) {
 	return 1;
 }
 
-R_API int r_cmd_add(RCmd *c, const char *cmd, const char *desc, r_cmd_callback(cb)) {
+R_API int r_cmd_add(RCmd *c, const char *cmd, RCmdCb cb) {
 	int idx = (ut8)cmd[0];
 	RCmdItem *item = c->cmds[idx];
 	if (!item) {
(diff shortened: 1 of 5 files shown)


libr/include/r_cmd.h
```diff
@@ -52,59 +59,76 @@ typedef struct r_cmd_macro_t {
...
+		} oldinput;
+		struct {
+			const char **help_msg;
+			const char **help_detail;
+			const char **help_detail2;
+			struct r_cmd_descriptor_t *sub[127];
+		} oldhelp;
+		struct {
+			RCmdArgvCb cb;
+		} argv;
```
thestr4ng3r, line 105: argv sounds like a bad name to me for this

karliss:
From the very high level "go ahead in this direction"  and "this is completely wrong" level I would say go ahead. Looking at the exact implementation and part's that are not implemented yet I expect there to be changes structures used and method interfaces. Not sure how much of those details you wanted to discuss at this point.

Since the current changes don't allow replacing old recursive help yet. It's probably better to revert those changes and keep old recursive help as it was. I expect that getting the new structure in working state will require modifying those macros once more making the current wall of changes useless.

The way I see this moving forward is:
1. Get the code in running state where it can do command dispatching
2. Convert the top level cmds as R_CMD_DESC_TYPE_OLDINPUT in command tree so that all command dispatch goes through the new mechanism
3. Convert the recursive help to use new command tree
4. Incrementally split the old commands into smaller parts and convert them to new format.

karliss:
Parts that are note implemented yet or might require changes and could be useful to discuss.

* Additional structure for command lookup
* Logic for doing command lookup when command token contains more symbols than the name associated with callback (either roots of OLDINPUT commands or commands with arguments  included in the command name).
* Recursive and noncrecursive help printing
* builtin aliases (x/px) - new RCmdDescType kind ?
* oldinput command input offset -  how much of the command gets skipped when passing to callback
* components of the new help (summary, child command summary generated automatically, additional examples, additional description)

libr/core/cmd_anal.c
```diff
@@ -724,41 +724,41 @@ static const char *help_msg_ax[] = {
...
-	DEFINE_CMD_DESCRIPTOR (core, ah);
-	DEFINE_CMD_DESCRIPTOR (core, ahi);
-	DEFINE_CMD_DESCRIPTOR (core, ao);
-	DEFINE_CMD_DESCRIPTOR (core, ar);
-	DEFINE_CMD_DESCRIPTOR (core, ara);
-	DEFINE_CMD_DESCRIPTOR (core, arw);
-	DEFINE_CMD_DESCRIPTOR (core, as);
-	DEFINE_CMD_DESCRIPTOR (core, ax);
+static void cmd_anal_init(RCore *core, RCmdDesc *parent) {
+	DEFINE_CMD_DESCRIPTOR (core, parent, a);
```
radare, line 728: why not use core and "parent" inside the macro so you just need to specify the command name instead of repeating over and over. also this macro shhould be local to this file, so the name can be shorter

libr/include/r_cmd.h
ret2libc: is it better now? otherwise, do you have other ideas for the name?

XVilka:
It should also handle the cases when the command is implemented by the core plugin, e.g. `yara` command added by the `r_cmd_add_call()`: https://github.com/radareorg/radare2-extras/blob/master/yara/yara/core_yara.c#L362

ret2libc:
> It should also handle the cases when the command is implemented by the core plugin, e.g. `yara` command added by the `r_cmd_add_call()`: https://github.com/radareorg/radare2-extras/blob/master/yara/yara/core_yara.c#L362

Good point @XVilka , I have added that in my last commit. Now also things like `a2f` should work (though apparently we did not have tests for those things).

XVilka:
Not sure if related (https://builds.sr.ht/~xvilka/job/198532#task-test-528):
```
[XX] db/cmd/cmd_meta vars commenting
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e asm.arch=x86
e asm.bits=64

s 0x10
af

afvb 8 arg_8h int
afvb -8 var_8h uint32_t
afvs 8 var_sp_8h uint32_t
afvr rax var_rax uint64_t

Cvb 0x8 comment for arg_8h
Cvb -8 comment for var_8h
Cvr var_rax comment for var_rax
Cvs var_sp_8h comment for var_sp_8h

pd1~comment
?e --
Cvb
Cvb*
?e --
Cvr
Cvr*
?e --
Cvs
Cvs*
?e --

Cvb- arg_8h
pd1~comment
?e --
Cvb- -8
pd1~comment
?e --
Cvs- 8
pd1~comment
?e --
Cvr- var_rax
pd1~comment
?e --

' malloc://1024
-- stdout
--- .a	Thu Apr 30 07:54:46 2020
+++ .b	Thu Apr 30 07:54:46 2020
@@ -11,16 +11,3 @@
var_rax : comment for var_rax
"Cvr var_rax base64:Y29tbWVudCBmb3IgdmFyX3JheA== @ 0x00000010"
--
-var_sp_8h : comment for var_sp_8h
-"Cvs var_sp_8h base64:Y29tbWVudCBmb3IgdmFyX3NwXzho @ 0x00000010"
---
-|           ; var uint32_t var_8h @ rbp-0x8    ; comment for var_8h
-|           ; arg uint32_t var_sp_8h @ rsp+0x8    ; comment for var_sp_8h
-|           ; arg uint64_t var_rax @ rax    ; comment for var_rax
---
-|           ; arg uint32_t var_sp_8h @ rsp+0x8    ; comment for var_sp_8h
-|           ; arg uint64_t var_rax @ rax    ; comment for var_rax
---
-|           ; arg uint64_t var_rax @ rax    ; comment for var_rax
---
---


-- exit status: -1
```

ret2libc:
>     * Additional structure for command lookup

Done in the current state of the PR.

>     * Logic for doing command lookup when command token contains more symbols than the name associated with callback (either roots of OLDINPUT commands or commands with arguments  included in the command name).

If I understood your point correctly, that's done in the PR as well. I just try to match the cmd_identifier and if that does not work, i remove the last char and lookup again, until i terminate the string. It can probably be improved in terms of performance (e.g. if the longest command registered is 10 char long we probably don't care to look at all 100 chars in a cmd_identifier).

>     * Recursive and noncrecursive help printing

I'm starting to play with this. Do you prefer to look at this in this same PR or in a later stage?

>     * builtin aliases (x/px) - new RCmdDescType kind ?

Makes sense, but I think that's can be easily done later.

>     * oldinput command input offset -  how much of the command gets skipped when passing to callback

Done in current state.

>     * r_core_plugin_t call. Currently all the command calls go through all the plugins using oldinput string. How to deal with that? The new interface would allow plugins to register their commands properly. Should the call field be made optional and deprecated?

In theory we should update existing(and new) RCorePlugins to use r_cmd_desc_*_new (or r_cmd_add at least) so that RCmd is aware of those commands and can treat them as usual. However, in the current PR I'm keeping compatibility with existing RCorePlugins that do not register new commands but just use a "call" method. But yes, that method should probably be deprecated.

>     * components of the new help (summary, child command summary generated automatically, additional examples, additional description)

This is an important part indeed. Do we want to create a new Issue and discuss things there?

libr/core/cmd_api.c
```diff
@@ -311,22 +312,40 @@ static int cmdstatus2int(RCmdStatus s) {
...
+
+	// As old RCorePlugin do not register new commands in RCmd, we have no
+	// way of knowing if one of those is able to handle the input, so we
+	// have to pass the input to all of them before looking into the
+	// RCmdDesc tree
+	RListIter *iter;
+	RCorePlugin *cp;
+	char *exec_string = r_cmd_parsed_args_execstr (args);
+	r_list_foreach (cmd->plist, iter, cp) {
+		res = cp->call (cmd->data, exec_string);
```
karliss, line 334: This should check if call is not NULL, that way plugins using new API can not set it and do the command registration within init and fini. Unless plugins have other init method, function where they can do the command registration.  In that case those plugins could just not define RCorePlugin.

libr/core/cmd_api.c
```diff
@@ -236,11 +308,57 @@ R_API int r_cmd_call(RCmd *cmd, const char *input) {
...
+		return 0;
+	case R_CMD_STATUS_INVALID:
+		return -1;
+	case R_CMD_STATUS_EXIT:
+	default:
+		return -2;
+	}
+}
+
 R_API int r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
```
ret2libc, line 323: I actually intend to transform this into:
```C
R_API RCmdStatus r_cmd_call_parsed_args(....)
```
as it makes it clearer what's the return value. I think right now the `int` returned is used in unconsistent ways.

libr/include/r_cmd.h
```diff
@@ -113,14 +139,24 @@ R_API int r_core_plugin_add(RCmd *cmd, RCorePlugin *plugin);
...
-R_API int r_cmd_add(RCmd *cmd, const char *command, const char *desc, r_cmd_callback(callback));
+R_API int r_cmd_add(RCmd *cmd, const char *command, RCmdCb callback);
 R_API int r_core_del(RCmd *cmd, const char *command);
 R_API int r_cmd_call(RCmd *cmd, const char *command);
 R_API int r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args);
+R_API RCmdDesc *r_cmd_get_root(RCmd *cmd);
+R_API RCmdDesc *r_cmd_get_desc(RCmd *cmd, const char *cmd_identifier);
+
+/* RCmdDescriptor */
+R_API RCmdDesc *r_cmd_desc_inner_new(RCmd *cmd, RCmdDesc *parent, const char *name, const char *help);
```
ret2libc, line 153: actually, since I'm currently not using this `help` (and the others) anywhere I may as well just remove these. I think we are going to change these API a bit anyway once we agree on what to have as help. @karliss are you ok with me removing the `help` field and `help` in argument?

karliss: Sounds good..

libr/core/cmd_api.c
ret2libc: I have not done this as it would require many changes in cmd.c as well and I don't want to add too many changes in this PR. Going to do a separate pR after this.

libr/core/cmd_api.c
```diff
@@ -30,12 +56,23 @@ R_API RCmd *r_cmd_new () {
 		cmd->cmds[i] = NULL;
 	}
 	cmd->nullcallback = cmd->data = NULL;
+	cmd->root_cmd_desc = create_cmd_desc (NULL, R_CMD_DESC_TYPE_INNER, "");
+	cmd->ht_cmds = ht_pp_new0 ();
```
karliss, line 60: Doesn't look like ht_cmds is freed anywhere.

libr/core/cmd_api.c
```diff
@@ -53,10 +90,39 @@ R_API RCmd *r_cmd_free(RCmd *cmd) {
...
 	return NULL;
 }
 
+R_API RCmdDesc *r_cmd_get_root(RCmd *cmd) {
+	return cmd->root_cmd_desc;
+}
+
+static bool is_valid_desc(RCmdDesc *cd, const char *name) {
+	return cd->type == R_CMD_DESC_TYPE_OLDINPUT ||
+		(cd->type == R_CMD_DESC_TYPE_ARGV && !strcmp (cd->name, name));
```
karliss, line 104: Why is strcmp required? I can imagine situations where it would be useful to share the callback and differentiate the modes using command name.

libr/core/cmd_api.c
```diff
@@ -53,10 +90,39 @@ R_API RCmd *r_cmd_free(RCmd *cmd) {
...
+
+R_API RCmdDesc *r_cmd_get_desc(RCmd *cmd, const char *cmd_identifier) {
+	r_return_val_if_fail (cmd && cmd_identifier, NULL);
+	char *cmdid = strdup (cmd_identifier);
+	char *end_cmdid = cmdid + strlen (cmdid);
+	RCmdDesc *res = NULL;
+	// match longer commands first
+	while (*cmdid) {
+		res = ht_pp_find (cmd->ht_cmds, cmdid, NULL);
+		r_warn_if_fail (!res || res->type == R_CMD_DESC_TYPE_OLDINPUT || res->type == R_CMD_DESC_TYPE_ARGV);
```
karliss, line 115: Is this debug code? Wouldn't it print warning under normal use case where loop performs more than one iteration.

libr/include/r_cmd.h
```diff
@@ -15,8 +15,15 @@ extern "C" {
 #define MACRO_LABELS 20
 #define R_CMD_MAXLEN 4096
 
-#define r_cmd_callback(x) int (*x)(void *data, const char *input)
-#define r_cmd_nullcallback(x) int (*x)(void *data)
+typedef enum r_cmd_status_t {
+	R_CMD_STATUS_OK = 0,
+	R_CMD_STATUS_INVALID,
+	R_CMD_STATUS_EXIT
+} RCmdStatus;
```
karliss, line 22: How does RCmdStatus differ from RCoreCmdStatus? Looking at current code it seems that RCoreCmdStatus is result of evaluating shell expression but RCmdStatus is result of command callback. Are those really different things?

ret2libc: no they aren't, but I want to remove RCoreCmdStatus and just bring it in r_cmd.h because I think it belongs there. Changing this here, however, would have required a lot of changes in cmd.c. Do you want me to do it here or in separate PR?

libr/core/cmd_api.c
ret2libc: it is "just to be sure". It shouldn't print anything in regular situation, it's only a kind of protection against me (the developer) from doing stupid thing

libr/core/cmd_api.c
ret2libc: I think for that we may need a different type in the future, otherwise how do you differentiate when the CmdDesc should handle it and when it should not?

e.g.

`afllll` should not be handled by `afl`, it should just say "command not found" (so r_cmd_get_desc should return NULL).

Maybe I can make this a little bit more clear, let me try to refactor this piece of code.

libr/core/cmd_api.c
```diff
@@ -813,3 +928,60 @@ R_API char *r_cmd_parsed_args_execstr(RCmdParsedArgs *a) {
...
+	if (!ht_pp_insert (cmd->ht_cmds, name, res)) {
+		r_cmd_desc_free (res);
+		return NULL;
+	}
+	return res;
+}
+
+R_API RCmdDesc *r_cmd_desc_inner_new(RCmd *cmd, RCmdDesc *parent, const char *name) {
+	r_return_val_if_fail (cmd && parent && name, NULL);
+	r_return_val_if_fail (parent->type == R_CMD_DESC_TYPE_INNER, NULL);
```
karliss, line 957: I am not sure if this condition will still hold once more commands get defined in the tree. Same applies to this condition in  `r_cmd_desc_oldinput_new`.

libr/core/cmd_api.c
```diff
@@ -30,12 +56,23 @@ R_API RCmd *r_cmd_new () {
...
 	cmd->nullcallback = cmd->data = NULL;
+	cmd->root_cmd_desc = create_cmd_desc (NULL, R_CMD_DESC_TYPE_INNER, "");
+	cmd->ht_cmds = ht_pp_new0 ();
 	r_core_plugin_init (cmd);
 	r_cmd_macro_init (&cmd->macro);
 	r_cmd_alias_init (cmd);
 	return cmd;
 }
 
+static void free_cmd_desc_tree(RCmdDesc *cd) {
```
karliss, line 67: Wouldn't recursive free be useful for plugins which register  multiple commands grouped under single inner node?

libr/core/cmd_api.c
```diff
@@ -813,3 +928,60 @@ R_API char *r_cmd_parsed_args_execstr(RCmdParsedArgs *a) {
...
+	r_return_val_if_fail (cmd && parent && name && cb, NULL);
+	r_return_val_if_fail (parent->type == R_CMD_DESC_TYPE_INNER, NULL);
+	RCmdDesc *res = create_cmd_desc (parent, R_CMD_DESC_TYPE_ARGV, name);
+	if (!res) {
+		return NULL;
+	}
+
+	res->d.argv_data.cb = cb;
+	if (!ht_pp_insert (cmd->ht_cmds, name, res)) {
+		r_cmd_desc_free (res);
```
karliss, line 949: Doesn't this leave a dangling pointer in parent node?

libr/core/cmd_api.c
```diff
@@ -236,11 +302,60 @@ R_API int r_cmd_call(RCmd *cmd, const char *input) {
...
+		}
+	}
 	free (exec_string);
+	if (res) {
+		return 1;
+	}
+
+	RCmdDesc *cd = r_cmd_get_desc (cmd, r_cmd_parsed_args_cmd (args));
+	if (!cd) {
+		return -1;
```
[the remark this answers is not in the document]
ret2libc: I agree, but I don't think RCmd should print anything. I think we should think about a separate RCmdStatus value and the caller should print whatever is needed in that case. What do you think?

libr/core/cmd_api.c
ret2libc: Right now I'd like to keep this invariant, where if a node has children it must be a INNER node. All commands that actually do something should be ARGV/OLDINPUT.

libr/core/cmd_api.c
karliss: Sounds reasonable. I agree that having separate type or a flag for that would be better if it helps detecting wrong command.

libr/core/cmd_api.c
karliss: I think I misunderstood the how warn_if_fail condition works.

libr/include/r_cmd.h
karliss: If there are plans to unify then it's fine. Just wanted to clarify it.

libr/core/cmd_api.c
ret2libc: it's a sort of assert :) i just put it there to be sure to handle new cases as i will add new R_CMD_DESC_TYPE_*

libr/core/cmd_api.c
ret2libc: What do you mean?

libr/include/r_cmd.h
ret2libc: Yeah, I will probably follow up with a small PR that does a bit of a "cleaning".

libr/core/cmd_api.c
karliss: I think we agreed that in future plugins should prefer the new cmd for registering their commands. That means plugins should be also able to remove and free their commands. I initially thought that calling free_cmd_desc_tree would convenient way of cleaning up. That is during initialization plugin registers "plug" "pluga" and "plugc" and then during cleanup calls recursive free on "plug".

After thinking about this more that's not exactly right API. Better API would be having `cmd_remove(RCmd*, RCmdDesc*)` or `cmd_remove(RCmd*, const char*)` method which does all the cleanup. It should not only free RCmdDesc but also data from other structures like `ht_cmds` and pointer in the parent command.

ret2libc: yes, I think everything should pass through RCmd.

ret2libc:
> Lgtm, whats left to merge this?

I'm waiting for @karliss approval and then I will merge this.

PR #16752, opened 2020-04-30
Improve itanium RTTI parsing ##anal

mrmacete:

**Detailed description**

- Improve itanium RTTI parsing, to work on MachO arm/arm64, to allow class recovery and to work even without symbols
- Improve vtable search, making it compatible with arm/arm64 and using simpler heuristics for carving the tables from the data section (inspired by paper "Devil is Virtual: Reversing Virtual Inheritance in C++ Binaries" (Rukayat, Erinfolami 2020) [arXiv:2003.05039](https://arxiv.org/abs/2003.05039))
- Fixed a bug in Mach-O parsing of relocs, for which on arm/thumb the pointers were wrongly aligned to 16 bits
- Fixed a NULL dereference in class.c for which incomplete vtables caused a segfault

NOTE: added tests, but the binaries are here: https://github.com/radareorg/radare2-testbins/pull/18

### DIFF
--- a/libr/anal/class.c
+++ b/libr/anal/class.c
@@ -815,10 +815,13 @@ R_API RAnalClassErr r_anal_class_vtable_get(RAnal *anal, const char *class_name,
 
 	vtable->offset = r_num_math (NULL, cur);
 
-	cur = next;
-	sdb_anext (cur, NULL);
-
-	vtable->size = r_num_math (NULL, cur);
+	if (next) {
+		cur = next;
+		sdb_anext (cur, NULL);
+		vtable->size = r_num_get (NULL, cur);
+	} else {
+		vtable->size = 0;
+	}
 
 	free (content);
 
--- a/libr/anal/rtti.c
+++ b/libr/anal/rtti.c
@@ -8,8 +8,7 @@ R_API char *r_anal_rtti_demangle_class_name(RAnal *anal, const char *name) {
 	if (context.abi == R_ANAL_CPP_ABI_MSVC) {
 		return r_anal_rtti_msvc_demangle_class_name (&context, name);
 	}
-	// TODO: implement class name demangling for itanium
-	return NULL;
+	return r_anal_rtti_itanium_demangle_class_name (&context, name);
 }
 
 R_API void r_anal_rtti_print_at_vtable(RAnal *anal, ut64 addr, int mode) {
@@ -31,14 +30,17 @@ R_API void r_anal_rtti_print_at_vtable(RAnal *anal, ut64 addr, int mode) {
 	}
 }
 
-static void rtti_msvc_print_all(RVTableContext *context, int mode) {
+R_API void r_anal_rtti_print_all(RAnal *anal, int mode) {
+	RVTableContext context;
+	r_anal_vtable_begin (anal, &context);
+
 	bool use_json = mode == 'j';
 	if (use_json) {
 		r_cons_print ("[");
 	}
 
 	r_cons_break_push (NULL, NULL);
-	RList *vtables = r_anal_vtable_search (context);
+	RList *vtables = r_anal_vtable_search (&context);
 	RListIter *vtableIter;
 	RVTableInfo *table;
 
@@ -49,12 +51,15 @@ static void rtti_msvc_print_all(RVTableContext *context, int mode) {
 			if (r_cons_is_breaked ()) {
 				break;
 			}
-
 			if (use_json && success) {
 				r_cons_print (",");
 				comma = true;
 			}
-			success = r_anal_rtti_msvc_print_at_vtable (context, table->saddr, mode, true);
+			if (context.abi == R_ANAL_CPP_ABI_MSVC) {
+				success = r_anal_rtti_msvc_print_at_vtable (&context, table->saddr, mode, true);
+			} else {
+				success = r_anal_rtti_itanium_print_at_vtable (&context, table->saddr, mode);
+			}
 			if (success) {
 				comma = false;
 				if (!use_json) {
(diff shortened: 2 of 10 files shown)


libr/anal/vtable.c
```diff
@@ -90,53 +94,56 @@ static bool vtable_section_can_contain_vtables(RVTableContext *context, RBinSect
...
 	}
-	// total xref's to curAddress
-	RList *xrefs = r_anal_xrefs_get (context->anal, curAddress);
-	if (r_list_empty (xrefs)) {
-		r_list_free (xrefs);
+
+	if (!context->read_addr (context->anal, curAddress - context->word_size, &value)) {
+		return false;
+	}
+	if (value != 0 && (value < data_section_start || value >= data_section_end)) {
```
thestr4ng3r, line 110: I am not sure whether this holds in MSVC when RTTI is disabled. We should check that.

libr/anal/vtable.c
```diff
@@ -90,53 +94,56 @@ static bool vtable_section_can_contain_vtables(RVTableContext *context, RBinSect
...
+		return false;
+	}
+	if (!context->read_addr (context->anal, curAddress - 2 * context->word_size, &value)) {
+		return false;
+	}
+	if ((st32)value > 0) {
```
thestr4ng3r, line 116: This almost never holds with MSVC. I think it doesn't have a second value before the vtables at all. If we remove this condition, we get a lot of false positives unfortunately.

mrmacete:
ok, thanks! will look deeper into MSVC

libr/anal/vtable.c
mrmacete: yeah that's right, also for itanium this check is wrong and should allow the value to be NULL

thestr4ng3r: It does allow it, the condition is inverse.

libr/anal/vtable.c
mrmacete: ok i think the safest way to go is doing different checks for different ABIs.

To my perspective in general it's better having false positives than nothing at all at this level, though. Because false positives can be ruled out when trying to actually use the data produced here, while false negatives makes the user completely blind.

libr/anal/vtable.c
mrmacete: hah yeah sorry i need some coffee

radare:
Florian: Put that bin in testbins and add a test for it

thestr4ng3r:
Tests: #16754

mrmacete:
rebased to get the new test, the logic is splitted now so it seems to work for both ABIs.

libr/anal/class.c
```diff
@@ -815,10 +815,13 @@ R_API RAnalClassErr r_anal_class_vtable_get(RAnal *anal, const char *class_name,
 
 	vtable->offset = r_num_math (NULL, cur);
 
-	cur = next;
-	sdb_anext (cur, NULL);
-
-	vtable->size = r_num_math (NULL, cur);
+	if (next != NULL) {
```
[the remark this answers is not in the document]
radare: drop the != NULL

libr/anal/class.c
```diff
@@ -815,10 +815,13 @@ R_API RAnalClassErr r_anal_class_vtable_get(RAnal *anal, const char *class_name,
...
+		cur = next;
+		sdb_anext (cur, NULL);
+		vtable->size = r_num_math (NULL, cur);
```
radare, line 821: i think r_num_get is enough, no need to compute math expressions here.

libr/anal/vtable.c
```diff
@@ -28,42 +28,46 @@ R_API void r_anal_vtable_info_free(RVTableInfo *vtable) {
...
-		case 32:
-			context->read_addr = anal->big_endian ? vtable_read_addr_be32 : vtable_read_addr_le32;
-			break;
-		case 64:
-			context->read_addr = anal->big_endian ? vtable_read_addr_be64 : vtable_read_addr_le64;
-			break;
-		default:
-			return false;
+	context->word_size = (ut8) (anal->bits / 8);
+	bool is_arm = anal->cur->arch && !strncmp (anal->cur->arch, "arm", 3);
```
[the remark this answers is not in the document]
radare: and use r_str_startswith()

libr/anal/vtable.c
```diff
@@ -90,8 +93,33 @@ static bool vtable_section_can_contain_vtables(RVTableContext *context, RBinSect
...
+	if (value != 0 && (value < data_section_start || value >= data_section_end)) {
+		return false;
+	}
+	if (!context->read_addr (context->anal, curAddress - 2 * context->word_size, &value)) {
+		return false;
+	}
+	if ((st32)value > 0) {
+		return false;
+	}
+
```
radare, line 118: empty line

libr/anal/vtable.c
radare: also !=0 is implicit, just use `value`

radare:
Screenshot 2020-05-03 at 13 38 37

libr/anal/class.c
mrmacete: hmm i really don't know this stuff i just enclosed it in an IF, but i can try to change it and see that nothing breaks as long as it's in r2r

libr/anal/rtti.c
```diff
@@ -49,12 +53,15 @@ static void rtti_msvc_print_all(RVTableContext *context, int mode) {
 			if (r_cons_is_breaked ()) {
 				break;
 			}
-
 			if (use_json && success) {
 				r_cons_print (",");
 				comma = true;
 			}
-			success = r_anal_rtti_msvc_print_at_vtable (context, table->saddr, mode, true);
+			if (context.abi == R_ANAL_CPP_ABI_MSVC) {
```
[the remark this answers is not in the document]
mrmacete: i'm not sure i got what you mean here

mrmacete:
weird that test doesn't fail for me, does it depend on the OS on which the test is being run?

mrmacete:
ahh probably it's for the new shell, i'm doing something weird in the test

radare:
looks good for merging, can you squash some commits so i can rebase instead of squash them all into a singe commit? thanks!

mrmacete:
squashed

radare:
Great! good work!

PR #16747, opened 2020-04-29
Add VT sequences input support ##windows

GustavoLCR:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

This makes radare2 compatible with the new Windows Terminal and semi-working under MinTTY

* Fix radare2 immediately exiting when console is redirected to pipe
* Hack to get viewport size under MinTTY
* Fix mouse when VT input disabled and buffer scrolled
* Fix line wrapping when VT output disabled

**Test plan**

- On ConHost (Windows 10 1909)
- [x] Test mouse/arrows with `e scr.vtmode=0`
- [x] Test mouse/arrows with `e scr.vtmode=1`
- [x] Test mouse/arrows with `e scr.vtmode=2`

- On Windows Terminal/Mintty
- [x] Test mouse/arrows with `e scr.vtmode=2`

**Closing issues**

Maybe #15674 cc @pandasauce

PR #16753, opened 2020-05-01
Windows sys/meson.py: Add r2 shell script

kazarmy:

**Detailed description**

During Windows installation, this pr adds an `r2` shell script that calls `radare2`, for shells that don't recognize .bat or .cmd extensions (e.g. git bash). This is to support #16747.

**Test plan**

Running `r2` on the git bash prompt works (assuming that it's in PATH ofc).

**Closing issues**

...

### DIFF
--- a/sys/meson.py
+++ b/sys/meson.py
@@ -172,6 +172,11 @@ def win_dist(args):
     with open(r2_bat_fname, 'w') as r2_bat:
         r2_bat.write('@"%~dp0\\radare2" %*\n')
 
+    r2_sh_fname = args.install + r'\bin\r2'
+    log.debug('create "%s"', r2_sh_fname)
+    with open(r2_sh_fname, 'w') as r2_sh:
+        r2_sh.write('#!/bin/sh\n$(dirname "$0")/radare2 "$@"\n')
+
     copy(r'{BUILDDIR}\libr\*\*.dll', r'{DIST}\bin')
     makedirs(r'{DIST}\{R2_LIBDIR}')
     if args.shared:


sys/meson.py
```diff
@@ -172,6 +172,11 @@ def win_dist(args):
     with open(r2_bat_fname, 'w') as r2_bat:
         r2_bat.write('@"%~dp0\\radare2" %*\n')
 
+    r2_sh_fname = args.install + r'\bin\r2'
```
radare, line 175: uhm isnt this meson.py suposed to run on windows too?

sys/meson.py
```diff
@@ -172,6 +172,11 @@ def win_dist(args):
...
+    log.debug('create "%s"', r2_sh_fname)
+    with open(r2_sh_fname, 'w') as r2_sh:
+        r2_sh.write('#!/bin/sh\n$(dirname "$0")/radare2 "$@"\n')
```
radare, line 178: why not just use a symlink?

sys/meson.py
kazarmy: it does ... this pr is for people running POSIX shells on Windows (e.g. the bash shell that comes with [MSYS2](https://www.msys2.org/))

sys/meson.py
kazarmy: on Windows, symlinks require their creators to jump through hoops ([either Administrator or Developer Mode](https://www.howtogeek.com/howto/16226/complete-guide-to-symbolic-links-symlinks-on-windows-or-linux/)) and i don't think `sys/meson.py` can assume that either is enabled

PR #16764, opened 2020-05-04
Use HtPP for esil ops

pelijah:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/anal/esil.c
+++ b/libr/anal/esil.c
@@ -105,7 +105,7 @@ R_API RAnalEsil *r_anal_esil_new(int stacksize, int iotrap, unsigned int addrsiz
 }
 
 R_API bool r_anal_esil_set_op(RAnalEsil *esil, const char *op, RAnalEsilOpCb code, ut32 push, ut32 pop, ut32 type) {
-	r_return_val_if_fail(code && R_STR_ISNOTEMPTY (op) && esil && esil->ops, false);
+	r_return_val_if_fail (code && R_STR_ISNOTEMPTY (op) && esil && esil->ops, false);
 	RAnalEsilOp *eop = ht_pp_find (esil->ops, op, NULL);
 	if (!eop) {
 		eop = R_NEW (RAnalEsilOp);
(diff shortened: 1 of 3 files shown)


radare:
Performance diff now?

libr/anal/esil.c
```diff
@@ -74,6 +74,11 @@ static bool popRN(RAnalEsil *esil, ut64 *n) {
 
 /* R_ANAL_ESIL API */
 
+static void esil_ops_free(HtPPKv *kv) {
+	free (kv->key);
```
radare, line 78: Check if kv is null

libr/anal/esil.c
```diff
@@ -100,18 +105,15 @@ R_API RAnalEsil *r_anal_esil_new(int stacksize, int iotrap, unsigned int addrsiz
 }
 
 R_API bool r_anal_esil_set_op(RAnalEsil *esil, const char *op, RAnalEsilOpCb code, ut32 push, ut32 pop, ut32 type) {
-	r_return_val_if_fail(code && op && strlen (op) && esil && esil->ops, false);
-	char t[128];
-	char *h = sdb_itoa (sdb_hash (op), t, 16);
-	RAnalEsilOp *eop = (RAnalEsilOp *)(size_t)sdb_num_get (esil->ops, h, 0);
+	r_return_val_if_fail(code && R_STR_ISNOTEMPTY (op) && esil && esil->ops, false);
```
radare, line 108: Missing space before (

libr/anal/esil.c
radare: Current usage is correct but for consistency eith all the free methods should work when passing null as arg

ret2libc: Actually I think there is no ht_ free method so far that accepts NULL. In any case, Ht will never call the free method on a NULL object (at least it shouldn't, unless there are bugs and in that case it would be better to catch them anyway, but I don't think there are as I haven't seen those kind of issues so far).

radare:
before aae: 1.8s
after aae: 1.6s

totally random binary without much testing, but its a tiny but welcome optimization

PR #16769, opened 2020-05-05
When getting terminal size on mintty, reuse prev vals if rows=0

kazarmy:

**Detailed description**

Occasionally on mintty, `__xterm_get_size()` will get a value of 0 for rows which can disrupt the Visual mode display. This pr fixes that by reusing previous values if so, and assumes that all is fine and dandy at time = 0.

**Test plan**

Expand the mintty terminal window to full size, run `r2 --`, go to Visual mode (`V`) and hold down the arrow-down key. The display should not flicker.

I use the mintty from git bash.

**Closing issues**

...

### DIFF
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -1332,8 +1332,13 @@ static bool __xterm_get_size(void) {
 	if (write (I.fdout, R_CONS_CURSOR_SAVE, sizeof (R_CONS_CURSOR_SAVE)) < 1) {
 		return false;
 	}
+	int rows, columns;
 	(void)write (I.fdout, "\x1b[999;999H", sizeof ("\x1b[999;999H"));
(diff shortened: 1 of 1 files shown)


XVilka:
Shouldn't the bug be reported to https://github.com/mintty/mintty ?

kazarmy:
Perhaps but I need to examine `__xterm_get_cur_pos()` first and that will take some time, and it's unclear whether a minimal reproducible test case can be produced.

This pr is a quick fix that should tide things over while the above is done.

kazarmy:
It is possible that in `__xterm_get_cur_pos()`, the code is racy between the write of `R_CONS_GET_CURSOR_POSITION` and the reading of the console reply. This might be difficult, if not impossible to fix.

PR #16774, opened 2020-05-06
Remove RCoreCmdStatus and just use RCmdStatus

ret2libc:

**Detailed description**

RCoreCmdStatus is the same as RCmdStatus. That enum belongs more to r_cmd.h than to core in general, so this PR removes RCoreCmdStatus and just uses RCmdStatus. It also changes the API of r_cmd_call_parsed_args so that it directly returns RCmdStatus.

### DIFF
--- a/libr/core/cmd_api.c
+++ b/libr/core/cmd_api.c
@@ -308,20 +308,18 @@ R_API int r_cmd_call(RCmd *cmd, const char *input) {
 	return ret;
 }
 
-static int cmdstatus2int(RCmdStatus s) {
-	switch (s) {
-	case R_CMD_STATUS_OK:
-		return 0;
-	case R_CMD_STATUS_INVALID:
-		return -1;
-	case R_CMD_STATUS_EXIT:
-	default:
-		return -2;
+static RCmdStatus int2cmdstatus(int v) {
+	if (v == -2) {
+		return R_CMD_STATUS_EXIT;
+	} else if (v < 0) {
+		return R_CMD_STATUS_INVALID;
+	} else {
+		return R_CMD_STATUS_OK;
 	}
 }
 
-R_API int r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
-	int res = 0;
+R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
+	RCmdStatus res = R_CMD_STATUS_INVALID;
 
 	// As old RCorePlugin do not register new commands in RCmd, we have no
 	// way of knowing if one of those is able to handle the input, so we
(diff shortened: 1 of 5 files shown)


libr/core/cmd_api.c
```diff
@@ -332,33 +330,34 @@ R_API int r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+				res = int2cmdstatus (ires);
 				break;
 			}
 		}
 	}
 	R_FREE (exec_string);
-	if (res) {
-		return 1;
+	if (res != R_CMD_STATUS_INVALID) {
+		return res;
```
karliss, line 342: If a plugin handles a command regardless of command failing or succeeding further commands should not be executed.

Theory 1 - return value of `cp->call` is a boolean indicating if plugin handled the command.
In case of this theory a plugin can't properly report if command succeeded or not.
This matches with the old code inside loop breaking when cp->call returns non 0 value.  This also matches with behaviour of r2ghidra-dec plugin. It returns true when the command starts with "pdg" and false otherwise.

Theory 2 - return value of `cp->call` is {0 - not handled, 1 - handled and succeeded, <0 - error}
Theory 3 - there is no consistency in what the plugins do, everything is bad.

It would be good to check what most cmd plugins do.

ret2libc: Yes, indeed I do agree that we should stop regardless of command failing or not, but I don't have access to all core plugins :/

a2f and java plugins seem to return true/false indeed, but at the same time one could have written a plugin that returns -2 to exit completely from r2... This seemed the most compatible way to still support existing plugins as, i'm afraid, we are in "theory 3" :(

If we want to just support true/false at the risk of breaking some existing coreplugins (that anyway will probably be broken some time in the future), I can fix this.

karliss: Went through first 12 pages of searching RCorePlugin in Github. Almost all of them matcheded the theory 1, by returning true if prefix matches and false otherwise.

R2Ghidra-dec - good
kriw/rust_r2_plugin  returns void most likely resulting in garbage
core_test.c lalalala - returns false, good
thestr4ng3r/r2db - good
MaskRay/r2cLEMENCy  - good
radare2-extras/bcl/core_bcl.c - good
wargio/r2dec-wasm - good
MohamadMansouri/FIRST-plugin-radare2 - good
radare2-au/adare2-gamepad/core_gamepad.c  - good
0ki/r2lrn golf-plugin - good
radare2-extras core_ssdeep.c - good
r2wakare - always returns false
retdec-r2plugin - good
rarare2-extras/r2tox - good

Old code in `r_cmd_call` was doing
```
r_list_foreach (cmd->plist, iter, cp) {
   if (cp->call (cmd->data, input)) {
	free (nstr);
	return true;
    }
}
```
So I would say it's quite safe to assume that return value means handled/not handled. Any plugins that don't work with that would have been equally broken before.

karliss: This would be a good moment to document such assumptions by adding short comment in the header next to struct fields.

ret2libc: Done, thanks!

Issue #16147, opened 2020-03-05
To think about error codes for the r2 commands


XVilka:
Basically ERROR/SUCCESS/TIMEOUT

PR #16782, opened 2020-05-07
Generate help automatically ##newshell

ret2libc:

**Detailed description**

Basic draft to set/get help messages with the tree structure.
This is very basic and I think it depends a lot on what we agree on for the new help. I believe there are redundant info in the current help and we should try to provide something that is accurate and to the point at the same time.

For example:
```
> w?
Usage: w[x] [str] [<file] [<<EOF] [@addr]  
...
```
Does the first line with `Usage` give you any real help on how to use `w` subcommands? I don't think so. IMHO we can remove this.

```
> p?
| pb[?] [n]               bitstream of N bits
| pB[?] [n]               bitstream of N bytes
| ph[?][=|hash] ([len])   calculate hash for a block
| pj[?] [len]             print as indented JSON
| pm[?] [magic]           print libmagic data (see pm? and /m?)
| po[?] hex               print operation applied to block (see po?)
```
Is it really useful to have `[?]` on each command? IMHO it can be removed, you can use `?` after any command.

**Test plan**

1. discuss the desidered output
2. discuss the proposed API in this PR to get/set the help
3. move `wv` sub-tree to new ARGV type and see how that works.

**TODO**

- [ ] add colors to help messages.


libr/core/cmd_api.c
```diff
@@ -363,6 +371,105 @@ R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+	case R_CMD_DESC_TYPE_OLDINPUT:
+		return oldinput_get_help (cmd, cd, args);
+	default:
+		r_warn_if_reached ();
+		return NULL;
+	}
+}
+
+R_API char *r_cmd_get_recursive_help (RCmd *cmd) {
+	return "";
```
radare, line 470: this is const char ? or strdup()

libr/core/cmd_api.c
```diff
@@ -363,6 +371,105 @@ R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+	}
+	char *res = strdup (s? s: "");
+	r_cons_pop ();
+	return res;
+}
+
+R_API char *r_cmd_get_help(RCmd *cmd, RCmdParsedArgs *args) {
+	char *cmdid = strdup (r_cmd_parsed_args_cmd (args));
+	char *cmdid_p = cmdid + strlen (cmdid) - 1;
+	int detail = 0;
```
radare, line 435: size_t

libr/core/cmd_api.c
```diff
@@ -363,6 +371,105 @@ R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+	while (cmdid_p >= cmdid && *cmdid_p == '?') {
+		*cmdid_p = '\0';
+		cmdid_p--;
+		detail++;
+	}
+
+	if (detail <= 0) {
+		// there should be at least one `?`
```
radare, line 443: cmdid leak

libr/core/cmd_api.c
```diff
@@ -363,6 +371,105 @@ R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+}
+
+static char *oldinput_get_help(RCmd *cmd, RCmdDesc *cd, RCmdParsedArgs *a) {
+	const char *s = NULL;
+	r_cons_push ();
+	RCmdStatus status = r_cmd_call_parsed_args (cmd, a);
+	if (status == R_CMD_STATUS_OK) {
+		s = r_cons_get_buffer ();
+	}
+	char *res = strdup (s? s: "");
```
radare, line 427: `char *res = strdup ((status == R_CMD_STATUS_OK)? r_cons_get_buffer (): ""): `

libr/core/cmd_api.c
```diff
@@ -363,6 +371,105 @@ R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+	char *cmdid = strdup (r_cmd_parsed_args_cmd (args));
+	char *cmdid_p = cmdid + strlen (cmdid) - 1;
+	int detail = 0;
+	while (cmdid_p >= cmdid && *cmdid_p == '?') {
+		*cmdid_p = '\0';
+		cmdid_p--;
+		detail++;
+	}
+
+	if (detail <= 0) {
```
radare, line 442: can't be < 0

karliss:
Please rebase and resolve conflicts once  #16784 is merged. Looks like there are some conflicting changes.

libr/core/cmd_api.c
ret2libc: i did it this way because `s` could be NULL.

radare:
What about adding an option to specify this argument string by the programmer instead of automatically generating it? because i think this string should represent the most common subcommands, and maybe that requires some human handling

ret2libc:
> What about adding an option to specify this argument string by the programmer instead of automatically generating it? because i think this string should represent the most common subcommands, and maybe that requires some human handling

Which string are you referring to exactly?

radare:
The one after Usage: command

radare:
ping

libr/core/cmd_api.c
```diff
@@ -363,6 +371,105 @@ R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+	return r_strbuf_drain (sb);
+}
+
+static char *argv_get_help(RCmd *cmd, RCmdDesc *cd, RCmdParsedArgs *a, int detail) {
+	RStrBuf *sb = r_strbuf_new (NULL);
+	const char **e;
+
+	switch (detail) {
+	case 1:
+		r_strbuf_appendf (sb, "%s %s # %s\n", cd->name, cd->help.args_str, cd->help.summary);
```
karliss: What's the point of short help?

libr/core/cmd_write.c
```diff
@@ -534,6 +520,75 @@ static void cmd_write_value (RCore *core, const char *input) {
...
 
+static RCmdStatus common_wv_handler(RCore *core, int argc, const char **argv, int type) {
+	ut64 off = 0LL;
+	ut8 buf[sizeof(ut64)];
+	int wseek = r_config_get_i (core->config, "cfg.wseek");
+	bool be = r_config_get_i (core->config, "cfg.bigendian");
+
+	core->num->value = 0;
+	if (argc != 2) {
+		return R_CMD_STATUS_INVALID;
```
ret2libc, line 531: shall we already think about introducing a new `R_CMD_STATUS_WRONG_ARGS`? I think it makes sense, as you don't want a general error in this case, but probably you want the RCore to print the help of this command or a more meaningful message anyway. @karliss wdyt?

libr/core/cmd_write.c
```diff
@@ -534,6 +520,75 @@ static void cmd_write_value (RCore *core, const char *input) {
...
+	}
+
+	off = r_num_math (core->num, argv[1]);
+	if (core->file) {
+		r_io_use_fd (core->io, core->file->fd);
+	}
+
+	ut64 res = r_io_seek (core->io, core->offset, R_IO_SEEK_SET);
+	if (res == UT64_MAX) {
+		return R_CMD_STATUS_INVALID;
```
ret2libc, line 541: also in this case, I guess, it is not exactly clear what's the meaning.

There should probably be something like:
COMMAND_NOT_FOUND,
WRONG_ARGS,
OK,
ERROR (error returned by the right command handler), like in this case.

Others?

libr/core/cmd_write.c
```diff
@@ -1788,3 +1843,63 @@ static int cmd_write(void *data, const char *input) {
...
+	DEFINE_CMD_DESCRIPTOR (core, we);
+	DEFINE_CMD_DESCRIPTOR (core, wo);
+	DEFINE_CMD_DESCRIPTOR (core, wop);
+	DEFINE_CMD_DESCRIPTOR (core, wp);
+	DEFINE_CMD_DESCRIPTOR (core, wt);
+	DEFINE_CMD_DESCRIPTOR (core, wv);
+	DEFINE_CMD_DESCRIPTOR (core, wx);
+
+	RCmdDesc *wv_cd = r_cmd_desc_argv_new (core->rcmd, parent, "wv", wv_handler);
+	RCmdDescHelp wv_help = {
```
ret2libc, line 1861: FIX: need to initialize to 0 first and then set fields

libr/core/cmd_write.c
```diff
@@ -1788,3 +1843,63 @@ static int cmd_write(void *data, const char *input) {
...
+	DEFINE_CMD_DESCRIPTOR (core, wc);
+	DEFINE_CMD_DESCRIPTOR (core, we);
+	DEFINE_CMD_DESCRIPTOR (core, wo);
+	DEFINE_CMD_DESCRIPTOR (core, wop);
+	DEFINE_CMD_DESCRIPTOR (core, wp);
+	DEFINE_CMD_DESCRIPTOR (core, wt);
+	DEFINE_CMD_DESCRIPTOR (core, wv);
+	DEFINE_CMD_DESCRIPTOR (core, wx);
+
+	RCmdDesc *wv_cd = r_cmd_desc_argv_new (core->rcmd, parent, "wv", wv_handler);
```
karliss, line 1860: Shouldn't the parent of `wv` be `w` instead of root?

libr/core/cmd_api.c
```diff
@@ -355,6 +355,157 @@ R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+
+	switch (cd->type) {
+	case R_CMD_DESC_TYPE_ARGV:
+		if (detail == 1 && !r_pvector_empty (&cd->children)) {
+			if (args->argc > 1) {
+				return NULL;
+			}
+			return inner_get_help (cmd, cd);
+		}
+		return argv_get_help (cmd, cd, args, detail);
```
karliss, line 495: I'd say that "cmd?" should include the description and examples if children list is empty.

libr/core/cmd_write.c
karliss: Seems somewhat reasonable. Will need to update the error update macro.

karliss:
What's your latest opinion of inner nodes always having "inner_node" type vs using inner nodes only if no callback is attached, or using argv without callback?

karliss:
I think I already asked before but will it be possible to convert only one subtree from the old command to argv?

Will it be possible to convert only the help structure to new structure keeping oldinput for the actual command handling? The help messages are already separate so initially converting only them seems like more manageable early target than converting all commands fully to argv.

karliss:
What's the plan for help output highlighting?  Looks like the old help relies on it being split in parts.

libr/core/cmd_api.c
```diff
@@ -355,6 +355,157 @@ R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+	}
+	return max_len;
+}
+
+static void print_child_help(RStrBuf *sb, RCmdDesc *cd, size_t max_len) {
+	size_t str_len = strlen (cd->name) + strlen0 (cd->help.args_str);
+	size_t padding = str_len < max_len ? max_len - str_len : 0;
+	const char *cd_args_str = cd->help.args_str ? cd->help.args_str : "";
+	const char *cd_summary = cd->help.summary ? cd->help.summary : "";
+	r_strbuf_appendf (sb, "| %s %s %*s# %s\n", cd->name, cd_args_str, padding, "", cd_summary);
```
karliss, line 390: If `args_str` was printed without implicit space it could also be used for specifying "[a] <arg>" part from "cmd[a] <arg>" when it makes sense.

karliss: I am also not sure if it's necessary to distinguish "usage" from "{cmd}{args_str}".

libr/core/cmd_write.c
ret2libc: There is no `w` node. Everything is handled by the root node, apart from `wv` (for now).
Of course we could also create a new old_input node `w`, and make `wv` a child of `w`, but at this point I did not want to touch too many things. Mainly because if I'm going to change things, I would prefer to directly invest the time in converting them to arg nodes.

libr/core/cmd_api.c
ret2libc: Yeah the only "problem" is that e.g. `wv?` would display the list of subcommands, so in those cases you would need to use `wv??` while for others you would use `wv8?`. But yeah, I agree with you though I'm not sure how to solve this issue.

libr/core/cmd_api.c
ret2libc: Let me try this.

ret2libc:
> What's the plan for help output highlighting? Looks like the old help relies on it being split in parts.

I'm going to work now on this, I should be able to include it in this PR.The splitting is still there, so it should not be a problem to use the right colors.

> I think I already asked before but will it be possible to convert only one subtree from the old command to argv?

Isn't this what I have done with `wv`? If not, what do you mean exactly?

> Will it be possible to convert only the help structure to new structure keeping oldinput for the actual command handling? The help messages are already separate so initially converting only them seems like more manageable early target than converting all commands fully to argv.

I can try working on this, but I would for a separate PR as it is another thing, ok?

libr/core/cmd_write.c
karliss: The reason I asked because in my opinion it is a valuable example for how partially converting things to new system will work. More or less the same as question I asked lower. You could put effort into fully converting `w`, but it will not be possible everything at the same time. So partial conversion process should be validated early to ensure that design can support it well.

ret2libc: So you want to try having `w` as a oldinput node as child of `root`? With `wv` as child of `w`?

libr/core/cmd_api.c
ret2libc: I am not sure this is possible as currently the "usage" part may be different from the command name sometimes. Also, what do you mean with `If args_str was printed without implicit space it could also be used for specifying "[a] " part from "cmd[a] " when it makes sense.`?

ret2libc:
@karliss please have a look again! I have implemented the various things we have talked about.

libr/core/cmd_write.c
```diff
@@ -1788,3 +1843,66 @@ static int cmd_write(void *data, const char *input) {
...
+	r_return_if_fail (wv_cd);
+	RCmdDescHelp wv_help;
```
ret2libc, line 1862: actually, instead of working on a brand new `RCmdDescHelp` and then having to write it with `r_cmd_desc_set_help` I could just get `cd->help` and directly work on it. What do you think?

libr/core/cmd.c
```diff
@@ -4998,6 +5000,13 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(arged_command) {
 
 	pr_args->has_space_after_cmd = !ts_node_is_null (args) && ts_node_end_byte (command) < ts_node_start_byte (args);
 	res = r_cmd_call_parsed_args (state->core->rcmd, pr_args);
+	if (res == R_CMD_STATUS_WRONG_ARGS) {
+		const char *cmdname = r_cmd_parsed_args_cmd (pr_args);
+		eprintf ("Wrong number of arguments passed to `%s`, see its help with `%s?`\n", cmdname, cmdname);
+	} else if (res == R_CMD_STATUS_ERROR) {
+		const char *cmdname = r_cmd_parsed_args_cmd (pr_args);
+		eprintf ("Something wrong during the execution of `%s` command.\n", cmdname);
```
karliss, line 5008: Aren't many commands printing their own error messages making "Something wrong during the execution of " redundant and very verbose.

libr/include/r_cmd.h
```diff
@@ -16,9 +16,11 @@ extern "C" {
 #define R_CMD_MAXLEN 4096
 
 typedef enum r_cmd_status_t {
-	R_CMD_STATUS_OK = 0,
-	R_CMD_STATUS_INVALID,
-	R_CMD_STATUS_EXIT
+	R_CMD_STATUS_OK = 0, // command handler exited in the right way
+	R_CMD_STATUS_WRONG_ARGS, // command handler could not handle the arguments passed to it
+	R_CMD_STATUS_ERROR, // command handler had issues while running (e.g. allocation error, etc.)
+	R_CMD_STATUS_INVALID, // command could not be found (e.g. it does not exist)
```
karliss, line 22: Plenty of existing code although all from newshell that uses STATUS_INVALID as general purpose error not necessarily command could not be found.  It might be fine if the description gets updated. STATUS_INVALID meaning any shell level error command doesn't exist, bad expression in the `@` modifier and others.

libr/core/cmd_write.c
karliss: Do you mean setting fields inside `RCmdDesc->help` directly? I am fine with that.

libr/core/cmd_api.c
```diff
@@ -967,3 +1138,18 @@ R_API RCmdDesc *r_cmd_desc_parent(RCmdDesc *cd) {
...
+	help->group_summary = NULL;
+	help->args_str = NULL;
+	help->description = NULL;
+	r_vector_init (&help->examples, sizeof (RCmdDescExample), NULL, NULL);
+}
+
+R_API void r_cmd_desc_set_help(RCmdDesc *cd, RCmdDescHelp *help) {
+	r_return_if_fail (cd && help);
+	r_return_if_fail (help->summary && help->args_str);
+	cd->help = *help;
```
karliss, line 1154: Value copying `RCmdDescHelp` which contains nontrivial member `RVector` is bad.

libr/core/cmd.c
XVilka: Also overlaps with https://github.com/radareorg/radare2/issues/16147

libr/core/cmd_api.c
```diff
@@ -355,6 +353,174 @@ R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+	size_t str_len = strlen (cd->name) + strlen0 (cd->help.args_str);
+	size_t padding = str_len < max_len? max_len - str_len: 0;
+	const char *cd_args_str = cd->help.args_str? cd->help.args_str: "";
+	const char *cd_summary = cd->help.summary? cd->help.summary: "";
+
+	RCons *cons = r_cons_singleton ();
+	const char *pal_args_color = use_color? cons->context->pal.args: "",
+		   *pal_help_color = use_color? cons->context->pal.help: "",
+		   *pal_reset = use_color? cons->context->pal.reset: "";
+	r_strbuf_appendf (sb, "| %s %s%s %*s%s# %s%s\n", cd->name, pal_args_color, cd_args_str, padding, "", pal_help_color, cd_summary, pal_reset);
```
XVilka, line 399: Can you split this this line? Looks too long

libr/core/cmd_api.c
```diff
@@ -355,6 +353,174 @@ R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+}
+
+static char *inner_get_help(RCmd *cmd, RCmdDesc *cd, bool use_color) {
+	if (!cd->help.args_str || !cd->help.summary) {
```
XVilka, line 403: Why not `r_return_if_fail()`?

libr/core/cmd_api.c
```diff
@@ -355,6 +353,174 @@ R_API RCmdStatus r_cmd_call_parsed_args(RCmd *cmd, RCmdParsedArgs *args) {
...
+		return r_strbuf_drain (sb);
+	case 2:
+		if (cd->help.description) {
+			r_strbuf_appendf (sb, "\n%s\n", cd->help.description);
+		}
+		if (!r_vector_empty (&cd->help.examples)) {
+			r_strbuf_appendf (sb, "\n%sExamples:%s\n", pal_label_color, pal_reset);
+			RCmdDescExample *it;
+			r_vector_foreach (&cd->help.examples, it) {
+				r_strbuf_appendf (sb, "| %s%s%s %s# %s%s\n", pal_input_color, it->example, pal_reset, pal_help_color, it->comment, pal_reset);
```
XVilka, line 457: Please split this one too.

libr/core/cmd_write.c
XVilka: Sounds good to me as well.

libr/core/cmd_write.c
```diff
@@ -1788,3 +1843,66 @@ static int cmd_write(void *data, const char *input) {
...
+	wv_help.usage = "wv[size] [value]";
+	wv_help.summary = "Write value as 4-bytes/8-bytes based on value";
+	wv_help.args_str = "[value]";
+	wv_help.description = "Write the number passed as argument at the current offset as a 4-bytes value or 8-bytes value if the input is bigger than UT32_MAX, respecting the cfg.bigendian variable";
+	wv_help.group_summary = "Write value of given size";
+	RCmdDescHelp wv_ex1 = { "wv 0xdeadbeef", "Write the value 0xdeadbeef at current offset" };
+	r_vector_push (&wv_help.examples, &wv_ex1);
+	r_cmd_desc_set_help (wv_cd, &wv_help);
+
+	RCmdDesc *wv1_cd = r_cmd_desc_argv_new (core->rcmd, wv_cd, "wv1", wv1_handler);
```
XVilka, line 1873: Maybe adding a small title in a comment like `// "wv1" - bla-bla` would help visually separate these?

XVilka: Also can be compactified with a macro, I think.

libr/include/r_cmd.h
XVilka: See also https://github.com/radareorg/radare2/issues/16147

test/unit/test_static.c
```diff
@@ -0,0 +1,36 @@
+#include <r_util.h>
```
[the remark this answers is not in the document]
ret2libc: absolutely unrelated :D got added by mistake from another branch i was playing with, sorry

libr/core/cmd_api.c
ret2libc: I did not want to enforce help for all commands right now as in this transition phase some stuff may be left out and added at some point later.

libr/core/cmd.c
ret2libc: I have used a R_LOG_DEBUG, so when you set `log.level=1` you can still see this message. I think it may be useful for debugging in some cases, but it won't show up during a regular session.

libr/include/r_cmd.h
ret2libc: Please check now, I have improved the comment, should be good.

libr/core/cmd_write.c
ret2libc: Probably this can be "macrofied" a little more, but there are some problems with the addition of examples I could not figure out right now. However, this looks already much better this way, what do you think?

By the way, by doing this we can put all the define (https://github.com/radareorg/radare2/pull/16782/files#diff-8a598c2e45468fc2d7a1e00496426b29R1847) for all commands in a single (or multiple) `.h` files, so if someone wants to change the description/summary/etc. of a command he doesn't need to mess with the code.

libr/include/r_cmd.h
karliss: Looks good.

libr/core/cmd_write.c
karliss: How about using a format more similar to the existing help description arrays?

karliss:
Is https://github.com/radareorg/radare2/blob/d1b39b244d7e567bdcf180012bf486b4e8b7ced0/libr/core/cmd.c#L6576
still supposed to check CMD_STATUS_INVALID or any error now?

libr/core/cmd_write.c
ret2libc: I can do it!

PR #16784, opened 2020-05-07
Remove INNER nodes and merge inner node with the command with same name

ret2libc:
Maybe after all we don't need this inner nodes :D

While working on PR #16782 I realized I could have multiple RCmdDesc with the same name in the hashtable by having both inner and argv nodes. Shall we just remove the inner nodes and consider them as ARGV node?

### DIFF
--- a/libr/core/cmd_api.c
+++ b/libr/core/cmd_api.c
@@ -20,14 +20,7 @@ static bool cmd_desc_set_parent(RCmdDesc *cd, RCmdDesc *parent) {
 	return true;
 }
 
-static bool cmd_desc_remove_parent(RCmdDesc *cd) {
-	r_return_val_if_fail (cd && cd->parent, false);
-	r_pvector_remove_data (&cd->parent->children, cd);
-	cd->parent = NULL;
-	return NULL;
-}
-
-static RCmdDesc *create_cmd_desc(RCmdDesc *parent, RCmdDescType type, const char *name) {
+static RCmdDesc *create_cmd_desc(RCmd *cmd, RCmdDesc *parent, RCmdDescType type, const char *name) {
 	RCmdDesc *res = R_NEW0 (RCmdDesc);
 	if (!res) {
 		return NULL;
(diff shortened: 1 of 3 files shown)


karliss:
Do I understand correctly that you are replacing inner nodes with argv nodes that have NULL callback?

I am not sure what you meant by
>  multiple RCmdDesc with the same name in the hashtable by having both inner and argv nodes
How can you have multiple items in hashtable with the same key?

ret2libc:
> Do I understand correctly that you are replacing inner nodes with argv nodes that have NULL callback?

That's right

> 
> I am not sure what you meant by
> 
> > multiple RCmdDesc with the same name in the hashtable by having both inner and argv nodes
> 
> How can you have multiple items in hashtable with the same key?

So, suppose you have:
```
a
-> af
-> al
-> ax
```

On master, `a` would be a INNER node. If now you want to get the help of `a?`, you would need to find the RCmdDesc for `a` and then proceed from there. However, on master INNER nodes are not placed in the hashtable. So I would need to put `a` in the hashtable.

However, a very common case in r2 is:
```
a
-> a
-> af
---> af
---> afl
-> al
-> ax
```
where you have a command with the same name as the inner node. If you think about that case, you would end up with two RCmdDesc that have the same name but different type.

So, do we really need the inner type after all? Maybe not. After all, an "inner" RCmdDesc may be identified just as one that has `n_children > 0`

karliss:
That's why previously asked you about asserts enforcing that only nodes with inner node type can be inner node.

I initially thought that any node can have children and inner node is used only in cases when node is used only for grouping without any runnable callback.

What will happen in situation like this:
```
a - old_input handles a bunch of a* commands
ab - no meaningfull callback
abc - oldinput handled by a
abd - argv
```
That is situation where a was a big old type command, for which the abd subtree has been converted to argv.

ret2libc:
I think the example you mentioned is the same as https://github.com/radareorg/radare2/pull/16784/files#diff-6165d904c738c087415c0aeca0e91d17R167 , right?

In that test case, you have `ae` (oldinput), `aeir` (argv), `ae*` handled by `ae`(oldinput). Do you see problems here?

PR #16786, opened 2020-05-07
Reset anal regs in every iteration of aaft

pelijah:

**Detailed description**
Anal regs is not reseted between iterations of `aaft`. For example this can cause unnecessary loops during emulation of `rep stos*` instructions.
In my case `rdx = 0xffffffffffffffff` on entry of this function:
```
/ 21: sym.r_util.dll_r_mem_memzero (int64_t arg1, int64_t arg2);
|           ; var int64_t var_20h @ rsp+0x8
|           ; arg int64_t arg1 @ rcx
|           ; arg int64_t arg2 @ rdx
|           0x180014f90      48897c2408     mov qword [var_20h], rdi
|           0x180014f95      488bf9         mov rdi, rcx               ; arg1
|           0x180014f98      33c0           xor eax, eax
|           0x180014f9a      488bca         mov rcx, rdx               ; arg2
|           0x180014f9d      f3aa           rep stosb byte [rdi], al
|           0x180014f9f      488b7c2408     mov rdi, qword [var_20h]
\           0x180014fa4      c3             ret
```
...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -886,7 +886,6 @@ static bool cmd_anal_aaft(RCore *core) {
 		}
 		__add_vars_sdb (core, fcn);
 	}
-	r_core_cmd0 (core, "aeim-;aei-");
 	r_core_seek (core, seek, true);
 	r_reg_arena_pop (core->anal->reg);
 	r_config_set_i (core->config, io_cache_key, io_cache);
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -869,26 +869,28 @@ static bool cmd_anal_aaft(RCore *core) {
 	}
 	seek = core->offset;
 	r_reg_arena_push (core->anal->reg);
+	r_reg_arena_zero (core->anal->reg);
+	r_core_cmd0 (core, "aei;aeim");
+	ut8 *saved_arena = r_reg_arena_peek (core->anal->reg);
 	// Iterating Reverse so that we get function in top-bottom call order
 	r_list_foreach_prev (core->anal->fcns, it, fcn) {
-		r_core_cmd0 (core, "aei");
-		r_core_cmd0 (core, "aeim");
 		int ret = r_core_seek (core, fcn->addr, true);
 		if (!ret) {
 			continue;
 		}
+		r_reg_arena_poke (core->anal->reg, saved_arena);
 		r_anal_esil_set_pc (core->anal->esil, fcn->addr);
 		r_core_anal_type_match (core, fcn);
-		r_core_cmd0 (core, "aeim-");
-		r_core_cmd0 (core, "aei-");
 		if (r_cons_is_breaked ()) {
 			break;
 		}
 		__add_vars_sdb (core, fcn);
 	}
+	r_core_cmd0 (core, "aeim-;aei-");
 	r_core_seek (core, seek, true);
 	r_reg_arena_pop (core->anal->reg);
 	r_config_set_i (core->config, io_cache_key, io_cache);
+	free (saved_arena);
 	return true;
 }
 
--- a/test/db/cmd/cmd_fnj
+++ b/test/db/cmd/cmd_fnj
@@ -2,8 +2,8 @@ NAME=fnj shows demangled symbols
 FILE=bins/elf/demangle-test-cpp
 CMDS=<<EOF
 aaa
-fj~{268}
-fnj~{268}
+fj~{269}
+fnj~{269}
 EOF
 EXPECT=<<EOF
 {"name":"reloc.operator_delete_void","realname":"reloc.operator delete(void*)","size":8,"offset":16432}


ret2libc:
Not sure I understand. What was the original (wrong) behaviour and what is the expected behaviour?

pelijah:
Original: anal regs have random values before calling `r_core_anal_type_match()`
Expected: anal regs is reseted before calling `r_core_anal_type_match()`

ret2libc:
Ok but is there a difference in end-result? You say:

> Anal regs is not reseted between iterations of aaft. For example this can cause unnecessary loops during emulation of rep stos* instructions.

Are these unnecessary loops causing a difference in the end result? Just wondering, as I'm not that familiar with this code, but I'm trying to understand :)

pelijah:
It doesn't affect the analysis result but can cause some overhead due to a tracing of memory access. On the other hand, in general `aaft` will become a little bit slower. I understand that it is a controversial solution.

ret2libc:
@pelijah it is ok for me, but fix CI  first :)

test/db/cmd/cmd_fnj
```diff
@@ -2,8 +2,8 @@ NAME=fnj shows demangled symbols
 FILE=bins/elf/demangle-test-cpp
 CMDS=<<EOF
 aaa
-fj~{268}
-fnj~{268}
+fj~{269}
```
ret2libc, line 5: do you know what is the additional flag it finds and why?

pelijah: Just other order.

pelijah:
```diff
--- a1.txt	2020-05-10 13:48:04.756031447 +0300
+++ a2.txt	2020-05-10 13:51:17.398666647 +0300
@@ -39,7 +39,12 @@
 0x00000000 1 class.int__std::uninitialized_copy_std
 0x00000000 1 class.std::allocator_traits_std::allocator_int
 0x00000000 1 class.int___std
+0x00000000 8 rax
 0x00000000 8 rbx
+0x00000000 8 rcx
+0x00000000 8 rdx
+0x00000000 8 rsi
+0x00000000 8 rdi
 0x00000000 8 r8
 0x00000000 8 r9
 0x00000000 8 r10
@@ -48,12 +53,8 @@
 0x00000000 8 r13
 0x00000000 8 r14
 0x00000000 8 r15
-0x00000000 8 rdx
-0x00000000 8 rsi
-0x00000000 8 rcx
-0x00000000 8 rax
+0x00000000 8 rflags
 0x00000040 616 segment.PHDR
-0x00000044 8 rflags
 0x000002a8 28 section..interp
 0x000002a8 28 segment.INTERP
 0x000002a8 0 .interp
@@ -286,7 +287,6 @@
 0x00004080 0 _edata
 0x00004080 0 __TMC_END__
 0x00004080 0 __bss_start
-0x00004080 8 rdi
 0x00004088 0 _end
 0x00178000 8 rbp
 0x00178000 8 rsp
```

pelijah: Well `rdi` and `rflags` offsets are zero now.

radare:
What about saving and restoring the regstate instead of zeroing out? So it will keep the user regstate and let the thing to be analyzed from a controlled/stable state

pelijah:
I am not sure what do you mean under "user regstate" but `cmd_anal_aaft()` already saves `anal->reg` state and restores it on exit.
https://github.com/radareorg/radare2/blob/794cc01c8b71fdffcb13e1a758eeef272cb8e1f9/libr/core/cmd_anal.c#L871
https://github.com/radareorg/radare2/blob/794cc01c8b71fdffcb13e1a758eeef272cb8e1f9/libr/core/cmd_anal.c#L891

radare:
yes i think that just doing (push reset { ...aaft... } pop) in order to keep the state and run aaft on a stable state and not losing the reg values set by the user or the debugger. is that the current case?

radare:
Ok then lgtm :)

libr/core/cmd_anal.c
```diff
@@ -871,6 +871,7 @@ static bool cmd_anal_aaft(RCore *core) {
 	r_reg_arena_push (core->anal->reg);
 	// Iterating Reverse so that we get function in top-bottom call order
 	r_list_foreach_prev (core->anal->fcns, it, fcn) {
+		r_reg_arena_zero (core->anal->reg);
```
radare, line 874: i think it would be better to do that after `aei`

radare:
Ping

libr/core/cmd_anal.c
pelijah: What is the difference?

libr/core/cmd_anal.c
```diff
@@ -871,6 +871,7 @@ static bool cmd_anal_aaft(RCore *core) {
...
 		r_core_cmd0 (core, "aei");
```
radare, line 875: actually i dont see the point on running aei and aeim on every function. we can do that once before the loop. so my pseudocode will look like this:

push,aei,zero,aeim ( aaft loop ) pop

libr/core/cmd_anal.c
radare: by reading the code doesnt seems like there's anything in aeim that modifies the registers. so its fine to call .zero before aei and aeim. but we can speedup the whole thing by moving the aei/aeim outside, and just push/pop the zeroed state inside the loop. because we want to keep SP,BP registers

pelijah:
Ok

pelijah:
Well now `aaft` on ~5% faster.

libr/core/cmd_anal.c
```diff
@@ -869,26 +869,28 @@ static bool cmd_anal_aaft(RCore *core) {
...
 		r_anal_esil_set_pc (core->anal->esil, fcn->addr);
 		r_core_anal_type_match (core, fcn);
-		r_core_cmd0 (core, "aeim-");
-		r_core_cmd0 (core, "aei-");
 		if (r_cons_is_breaked ()) {
 			break;
 		}
 		__add_vars_sdb (core, fcn);
 	}
+	r_core_cmd0 (core, "aeim-;aei-");
```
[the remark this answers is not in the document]
radare: Screenshot 2020-05-13 at 11 14 37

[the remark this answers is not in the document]
pelijah: How it's related to my PR?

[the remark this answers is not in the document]
radare: Because you added this line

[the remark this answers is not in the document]
pelijah: I just moved it out of the loop.

radare:
Its wrong to do that. It shoudlnt even run

radare:
Yeah you moved it out of the loop. but actually it should never be called because it may cause problems if you use aaft after startup

radare:
So please remove the aei;aeim line and confirm it still works, also how fast it is now after all the changes?

pelijah:
But `aeim` initializes `anal->esil` object without which `r_core_esil_*()` API are useless.

radare:
I only want u to remove the deinitialization calls

pelijah:
cmd `aa;aac;aar;?t aaft`:
`/bin/ls` (138KB): 1.135s vs 1.035s --> 8.8%
`/bin/busybox` (2MB): 20.28s vs 19.33s --> 4.7%

PR #16792, opened 2020-05-09
Implement the `pci` command print the C byte array + instruction comments ##disasm

radare:

**Detailed description**

IMAGE 2020-05-09 18:45:52

**Test plan**

pci #nbytes 

**Closing issues**

None, but its been a request from @oleavr 

**Comments**

I think some ways to improve this would be:

* Argument should be number of instructions, not bytes
* Honor asm.cmt.align instead of hardcoded column for comments


radare:
also seems like disasm() doesnt return the instruction size

Screenshot 2020-05-09 at 18 52 24

XVilka:
I think it's probably better not to print "// invalid" as a comment - it's useless.

radare:
well the thing is that disasm prints and returns the lenght, so i cant print the bytes before the instruction, and i cant use anal from rprint, so this weak disasm() callback should be improved to get it right.

libr/core/cmd_print.c
```diff
@@ -4354,13 +4355,56 @@ static const char* bits_to_c_code_fmtstr(int bits) {
...
 
+static void print_c_instructions(RPrint *p, ut64 addr, const ut8 *buf, int len, int ws) {
+	size_t i;
+
+	ws = R_MAX (1, R_MIN (ws, 8));
+	int bits = ws * 8;
+	const char *fmtstr = bits_to_c_code_fmtstr (bits);
+	len /= ws;
+
+	p->cb_printf ("#define _BUFFER_SIZE %d\n", len);
```
ret2libc, line 4366: why do we need `cb_printf` in RCore? Can't you just use `r_cons_printf` ?

libr/core/cmd_print.c
```diff
@@ -4354,13 +4355,56 @@ static const char* bits_to_c_code_fmtstr(int bits) {
...
+	p->cb_printf ("const uint%d_t buffer[_BUFFER_SIZE] = {\n  ", bits);
+	p->interrupt = false;
+
+	int left = 0;
+	const int orig_align = p->coreb.cfggeti (p->coreb.core, "asm.cmt.col") - 40;
```
ret2libc, line 4371: same here. why do we need`cfggeti`? just directly use the API to get the var

libr/core/cmd_print.c
```diff
@@ -4354,13 +4355,56 @@ static const char* bits_to_c_code_fmtstr(int bits) {
...
+	int align = orig_align;
+	char *instr = NULL;
+	for (i = 0; !p->interrupt && i < len; i++) {
+		if (left == 0) {
+			ut64 at = addr + i;
+			char *is = p->coreb.cmdstrf (p->coreb.core, "ao @ 0x%08"PFMT64x"~^size[1]", at);
```
ret2libc, line 4377: 1) no cmdstrf because you can use directly `r_core_cmdf` afaik.
2) just use the disasm api to get the size of the instruction.

libr/core/cmd_print.c
```diff
@@ -4354,13 +4355,56 @@ static const char* bits_to_c_code_fmtstr(int bits) {
...
+			free (is);
+			if (instr) {
+				while (align-- > 0) {
+					p->cb_printf (" ");
+				}
+				p->cb_printf (" /* %s */\n", instr);
+				free (instr);
+				p->cb_printf ("  ");
+			}
+			instr = p->coreb.cmdstrf (p->coreb.core, "pi 1 @ 0x%08"PFMT64x, at);
```
ret2libc, line 4388: no cmdstrf, just use the api

ret2libc:
BTW, there is something wrong:
```
[0x100001060]> pd 2
               ;-- main:
               ;-- entry0:
               ;-- func.100001060:
               ;-- rip:
               0x100001060      55             push rbp
               0x100001061      4889e5         mov rbp, rsp
[0x100001060]> pci 4
#define _BUFFER_SIZE 4
const uint8_t buffer[_BUFFER_SIZE] = {
  0x55,                           /* push rbp */
  0x48, 0x89, 0xe5,
};
```

It should also print the comments of the second instruction, but it does not.

ret2libc:
Also:
```
Oops at 0x100005038 (c7 4d 00 ...)
#define _BUFFER_SIZE 20
const uint8_t buffer[_BUFFER_SIZE] = {
  0x02, 0x00,  /* add al, byte [rax] */
  0x00, 0x00,  /* add byte [rax], al */
  0x00, 0x00,  /* add byte [rax], al */
  0x00, 0x00,  /* add byte [rax], al */
  0xc7, 0x4d, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00,
};
```

It seems that as soon as it finds an invalid instruction it does not print the comments anymore... is this intended?

radare:
no need to review yet, sorry missed the wip bit, i have many more changes in local and also tests <3

libr/core/cmd_print.c
radare: sorry, but util cant use the api

libr/core/cmd_print.c
radare: no, i cant
this is r_utl, not r_core
no i cant sue the disasm api either

libr/core/cmd_print.c
radare: no, i cant

libr/core/cmd_print.c
radare: no i cant,

radare:
Its ready to review again and merge. All issues addressed. I cant call the real apis because this code must be moved to rprint.. actually i can do that in this same pr. But i would prefer to do it separately or maybe in another commit

libr/core/cmd_print.c
```diff
@@ -4354,13 +4355,77 @@ static const char* bits_to_c_code_fmtstr(int bits) {
 	}
 }
 
+static void print_c_instructions(RPrint *p, ut64 addr, const ut8 *buf, int len, int ws) {
```
ret2libc, line 4358: if this needs to be moved in RPrint, do it now, otherwise keep it here but it needs some cleaning. `ws`, for example, can only have 1 as value, so all the operations on ws are useless.

libr/core/cmd_print.c
```diff
@@ -4354,13 +4355,77 @@ static const char* bits_to_c_code_fmtstr(int bits) {
...
+	len /= ws;
+
+	p->cb_printf ("#define _BUFFER_SIZE %d\n", len);
+	p->cb_printf ("const uint%d_t buffer[_BUFFER_SIZE] = {\n", bits);
+	p->interrupt = false;
+
+	int oleft, left = 0;
+	const int orig_align = p->coreb.cfggeti (p->coreb.core, "asm.cmt.col") - 40;
+	int align = orig_align;
+	int li = 0;
```
ret2libc, line 4373: unused var

libr/core/cmd_print.c
ret2libc: though I fail to see how this would be useful in a library function such as RPrint, since it implements a very particular command which I find it hard to reuse anywhere else.

libr/core/cmd_print.c
```diff
@@ -4354,13 +4355,77 @@ static const char* bits_to_c_code_fmtstr(int bits) {
...
+	size_t i;
+
+	ws = R_MAX (1, R_MIN (ws, 8));
+	int bits = ws * 8;
+	const char *fmtstr = bits_to_c_code_fmtstr (bits);
+	len /= ws;
+
+	p->cb_printf ("#define _BUFFER_SIZE %d\n", len);
+	p->cb_printf ("const uint%d_t buffer[_BUFFER_SIZE] = {\n", bits);
+	p->interrupt = false;
```
ret2libc, line 4368: interrupt is never set to true in the entire codebase AFAIU, so this can be completely remove (you can start by just not using it in your function I guess)

libr/core/cmd_print.c
radare: the function takes rprint as argument and it performs a very basic thing like hexdumping with a comment

ive moved the thing into r_util

libr/core/cmd_print.c
radare: it is, done by rcons_break_begin() and so on. its how rprint handle user interrupt events

radare:
@ret2libc again

libr/core/cmd_print.c
ret2libc: where? I can't find it, could you point me to the exact line where `p->interrupt` is set to true?

radare: oh, you are right, its just using a callback, fixed. thanks!

radare:
Screenshot 2020-05-12 at 00 34 50

libr/util/print.c
```diff
@@ -1,5 +1,6 @@
 /* radare - LGPL - Copyright 2007-2020 - pancake */
 
+// WTF :D
```
ret2libc, line 3: remove

libr/util/print_code.c
```diff
@@ -0,0 +1,281 @@
...
+		return "0x%04x";
+	case 32:
+		return "0x%08xU";
+	case 64:
+		return "0x%016" PFMT64x "ULL";
+	}
+	return "0x%02x";
+}
+
+static void print_c_instructions(RPrint *p, ut64 addr, const ut8 *buf, int len) {
```
ret2libc, line 18: please see https://github.com/radareorg/radare2/pull/16792#pullrequestreview-408955758, the implementation of this function is (in my opinion) very hard to understand.

test/db/cmd/cmd_pc
```diff
@@ -259,3 +259,67 @@ const uint8_t buffer[_BUFFER_SIZE] = {
...
+ 0x90,                                 /* nop */
+};
+EOF
+RUN
+
+NAME=pci
+FILE=bins/elf/analysis/hello-linux-x86_64
+CMDS=<<EOF
+pci 29
+pci -10
```
ret2libc, line 283: then someone has to explain me why `pci -10` should be converted to `pci 10` :D but that's a question for another day i guess.

radare:
Because pD -10 is the same as pD 10 @$$-10

libr/util/print.c
radare: i prefer to fix the incllude instead of removing the comment

libr/util/print_code.c
radare: To do what you want we need more apis accessible from rprint. Thats the only way we can do that here. I csn refactor the code to look better if u want. But its pretty clean as it is right now. We have zillions of things to refactor and cleanup everywhere in r2land. So this pr can become and endless monster of unrelated changes just to add a command

ret2libc: I know, but this is new code so it has to be good. No reason to introduce code that is already hard and already needs a lot of refactoring.

I think something like https://pastebin.com/6seAgYV6 looks much easier to read.

ITAYC0HEN: Fantastic suggestion @ret2libc. The code is very clean and clear

ITAYC0HEN:
Love this command! I can see how it becomes very useful to me <3

XVilka:
Yes, it's very useful for articles and reports.

ITAYC0HEN:
absolutely. Also for extracting shellcodes form a binary and executing it externally. Same for analyzing exploits. Really, very nice

radare:
with @ret2libc impl the comments are not properly aligned. will fix that after lunch

Screenshot 2020-05-13 at 13 57 19

libr/util/print_code.c
```diff
@@ -0,0 +1,274 @@
...
+
+		size_t j;
+		for (j = 0; j < inst_size && i + j < len; ++j) {
+			r_print_cursor (p, i + j, 1, 1);
+			p->cb_printf (fmtstr, r_read_ble (buf++, p->big_endian, 8));
+			r_print_cursor (p, i + j, 1, 0);
+			p->cb_printf (", ");
+		}
+
+		int align = orig_align - ((j - 1) * 6);
```
ret2libc, line 53: @radare yeah probably something to fix here in this calculation

radare:
fixed now, pls review again

Issue #3574, opened 2015-10-23
Supporting PCAP file capture for testing debuggers and network activity


XVilka:
So just record for example a GDB session with GDB protocol and use the session as a replay in a test -> open the pcap file, send the command, read the answer from pcap file -> if the answer is different - then return the error or so.

Maijin:
Dat would be support useful and many people are interested in the pcap support withing radare2 during hack.lu

radare:
Gdb already have a protocol debug flag that logs everything. This would be much easier to handle than a pcap.

Also i remember that i addrd support to r1 for loading pcap files. But this was so many years ago. Dunno whre this code is now

radare:
Srsly? This is damn simple to do

Maijin:
It was not gdb related interest but really about PCAP support. Especially circl.lu people

XVilka:
@radare - it is not about debugging, it is about autotesting gdb debuggers on CI - so gdb code in radare2 will be able simply replay the PCAPs of gdbserver sessions. That was the whole point of my suggestion. But the general support of PCAP format also included in this proposal.

radare:
Just write a small c program using libpcap that prints r2 commands to set flags to specify where each structure of the capture is in the pcap file.

chinmaydd:
@radare I would like to take up this issue and work on it.

radare:
All yours! I cant assign it to you because i can only do that enabling commit perms to you.

radare:
any update on this @chinmaydd ?

radare:
I think that just setting a flag at the beginning of each packet should be enough for an initial implementation.. An RBin plugin that loads the pcap, identifies the header and sets flags, it can also expose some pf expressions to parse different packet headers

chinmaydd:
@radare Sorry. I have been a little busy. Will be working on it soon.

radare:
Cool!

XVilka:
@chinmaydd @zamaudio - this feature can be very useful for writing automatic debugger tests for running on Travis-CI/AppVeyor without actual running debug mode - just send commands (and compare with PCAP out and read PCAP in and compare with expected), may be script for recording those PCAPs would be useful too. So it would be easier to check your endian patches too.

radare:
This must go into the r2e repo and available via r2pm

Maijin:
Isn't that done by https://github.com/radare/radare2-extras/pull/116 ? What remains to be done, since it's listed in GSOC task please describe :p

[...]

PR #16794, opened 2020-05-10
Implement #6431 to support gdb:// with debug_io handler

### LINKED ISSUE #6431: Implement dh io for gdb://
the io.gdb plugin must provide more debugger commands via =! or \.. this is what the debugger will use in case of `dh io` (instead of the dh gdb), both methods should be implemented imho, the io one is lower level

abcSup:
…bugger handler

 

**Detailed description**
I have added the support for the following commands in order to use debug_io handler `dL io` with gdb:
`dm`, `dr8`, `drp`, `ds`, `dc`, `dck`

```
r2 -d gdb://localhost:4444
= attach 2699 1
= attach 2699 0
 -- Good morning, pal *<:-)
[0x7ffff7dd6ea1]> =!?
Usage: =!cmd args
 =!dm              - list memory maps of target process
 =!dr8             - display hexdump of gpr arena
 =!drp             - get register profles
 =!ds              - step one instruction
 =!dsb             - step backwards
 =!dc              - continue execution
 =!dcb             - continue backwards
 =!dck <sig>       - continue execution with sig signal
 =!pid             - show targeted pid
 =!pkt s           - send packet 's'
 =!rd              - show reverse debugging availability
 =!monitor cmd     - hex-encode monitor command and pass to target interpreter
 =!detach [pid]    - detach from remote/detach specific pid
 =!inv.reg         - invalidate reg cache
 =!pktsz           - get max packet size used
 =!pktsz bytes     - set max. packet size as 'bytes' bytes
 =!exec_file [pid] - get file which was executed for current/specified pid
[0x7ffff7dd6ea1]> =!dm
0x555555554000 - 0x555555573000 r-xp /bin/ls
0x555555772000 - 0x555555775000 rw-p /bin/ls
0x555555775000 - 0x555555776000 rw-p [heap]
0x7ffff7dd5000 - 0x7ffff7dfc000 r-xp /lib/x86_64-linux-gnu/ld-2.27.so
0x7ffff7ff8000 - 0x7ffff7ffb000 r--p [vvar]
0x7ffff7ffb000 - 0x7ffff7ffc000 r-xp [vdso]
0x7ffff7ffc000 - 0x7ffff7ffe000 rw-p /lib/x86_64-linux-gnu/ld-2.27.so
0x7ffff7ffe000 - 0x7ffff7fff000 rw-p 
0x7ffffffde000 - 0x7ffffffff000 rw-p [stack]
0xffffffffff600000 - 0xffffffffff601000 --xp [vsyscall]
```

Users can also switch to debug.io `dL io` and perform debugging.
```
$ r2 -d gdb://localhost:4444
= attach 2699 1
= attach 2699 0
 -- Connection lost with the license server, your r2 session will terminate in 30 minutes.
[0x7ffff7dd6ea1]> dL io
[0x7ffff7dd6ea1]> dm
0x0000555555554000 - 0x0000555555573000 - usr   124K s r-x /bin/ls ? ; loc.imp._ITM_registerTMCloneTable
0x0000555555772000 - 0x0000555555775000 - usr    12K s rw- /bin/ls ?
0x0000555555775000 - 0x0000555555776000 - usr     4K s rw- [heap] ?
0x00007ffff7dd5000 - 0x00007ffff7dfc000 * usr   156K s r-x /lib/x86_64-linux-gnu/ld-2.27.so ?
0x00007ffff7ff8000 - 0x00007ffff7ffb000 - usr    12K s r-- [vvar] ?
0x00007ffff7ffb000 - 0x00007ffff7ffc000 - usr     4K s r-x [vdso] ?
0x00007ffff7ffc000 - 0x00007ffff7ffe000 - usr     8K s rw- /lib/x86_64-linux-gnu/ld-2.27.so ?
0x00007ffff7ffe000 - 0x00007ffff7fff000 - usr     4K s rw-  ?
0x00007ffffffde000 - 0x00007ffffffff000 - usr   132K s rw- [stack] ?
0xffffffffff600000 - 0xffffffffff601000 - usr     4K s --x [vsyscall] ?
```

**Test plan**
Currently, there is no way to test remote gdb features until #3574 is implemented. To validate the result of the io.gdb system commands, we can compare it with `d` debug commands which uses debug.gdb

Start a gdbserver
```
$ gdbserver host:4444 /bin/ls
```
Connect to gdbserver and test the followings commands and their equivalent `io` commands to confirm they produce the same output:
* `dm`  `=!dm`
* `dr8`  `=!dr8`
* `drp`  `=!drp`
```
$ r2 -d gdb://localhost:4444
= attach 2723 1
= attach 2723 0
 -- Enhance your graphs by increasing the size of the block and graph.depth eval variable.
[0x7ffff7dd6090]> dr8
000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000080dfffffff7f0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000009060ddf7ff7f000000020000330000002b0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002220000062690000efea000032380000653d22313d2200007479706569020000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000006e7431323b000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
[0x7ffff7dd6090]> =!dr8
000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000080dfffffff7f0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000009060ddf7ff7f000000020000330000002b0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002220000062690000efea000032380000653d22313d2200007479706569020000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000006e7431323b000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```
We can step a few instructions and check the `PC` register to confirm `=!ds` is working. For `dc`, we could call the command to confirm the process continue to end or a breakpoint.
* `ds` = `=!ds`
* `dc` = `=!dc`
* `dck 9` = `=!dck 9`

**Closing issues**
closes #6431

### DIFF
--- a/libr/debug/p/debug_io.c
+++ b/libr/debug/p/debug_io.c
@@ -123,7 +123,13 @@ static int __reg_read(RDebug *dbg, int type, ut8 *buf, int size) {
 
 // "dc" continue execution
 static int __io_continue(RDebug *dbg, int pid, int tid, int sig) {
-	dbg->iob.system (dbg->iob.io, "dc");
+	const char *cmd;
+	if (sig != 0) {
+		cmd = sdb_fmt ("dck %d %d", sig, pid);
+	} else {
+		cmd = "dc";
+	}
+	dbg->iob.system (dbg->iob.io, cmd);
 	r_cons_flush ();
 	return true;
 }
--- a/libr/debug/p/debug_io.c
+++ b/libr/debug/p/debug_io.c
@@ -123,12 +123,7 @@ static int __reg_read(RDebug *dbg, int type, ut8 *buf, int size) {
 
 // "dc" continue execution
 static int __io_continue(RDebug *dbg, int pid, int tid, int sig) {
-	const char *cmd;
-	if (sig != 0) {
-		cmd = sdb_fmt ("dck %d %d", sig, pid);
-	} else {
-		cmd = "dc";
-	}
+	const char *cmd = sig > 0 ? sdb_fmt ("dck %d %d", sig, pid) : "dc";
 	dbg->iob.system (dbg->iob.io, cmd);
 	r_cons_flush ();
 	return true;
--- a/libr/io/p/io_gdb.c
+++ b/libr/io/p/io_gdb.c
@@ -268,14 +268,14 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
 				gdbr_lock_leave (desc);
 				return NULL;
 			case 'k': // "dck"
-				if (cmd[3] != '\0') {
-					int sig = r_num_math (NULL, cmd + 3);
-					gdbr_continue (desc, desc->pid, -1, sig);
-					if (desc->stop_reason.is_valid && desc->stop_reason.thread.present) {
-						desc->tid = desc->stop_reason.thread.tid;
-					}
+			{
+				int sig = r_num_math (NULL, cmd + 3);
+				gdbr_continue (desc, desc->pid, -1, sig);
+				if (desc->stop_reason.is_valid && desc->stop_reason.thread.present) {
+					desc->tid = desc->stop_reason.thread.tid;
 				}
 				return NULL;
+			}
 			default:
 				break;
 			}
@@ -289,13 +289,7 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
 			ut64 buflen = 16384;
 			snprintf (path, sizeof (path) - 1, "/proc/%d/maps", desc->pid);
 
-#ifdef _MSC_VER
-#define GDB_FILE_OPEN_MODE (_S_IREAD | _S_IWRITE)
-#else
-#define GDB_FILE_OPEN_MODE (S_IRUSR | S_IWUSR | S_IXUSR)
-#endif
-
-			if (gdbr_open_file (desc, path, O_RDONLY, GDB_FILE_OPEN_MODE) < 0) {
+			if (gdbr_open_file (desc, path, O_RDONLY, 0) < 0) {
 				return NULL;
 			}
 			if (!(buf = malloc (buflen))) {
@@ -316,18 +310,15 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
 			char name[512];
 			for (;;) {
 				char *nl = strchr (ptr, '\n');
-				if (nl) {
-					*nl = 0;
-					*name = 0;
-					*perm = 0;
-					sscanf (ptr, "%"PFMT64x"-%"PFMT64x" %s %*s %*s %*s %[^\n]",
-						&map_start, &map_end, perm, name);
-					r_strbuf_appendf (obuf,"0x%"PFMT64x" - 0x%"PFMT64x" %s %s\n",
-							  map_start, map_end, perm, name);
-					ptr = nl + 1;
-				} else {
+				if (!nl) {
 					break;
 				}
+				*nl = 0;
+				*name = 0;
+				*perm = 0;
+				sscanf (ptr, "%"PFMT64x"-%"PFMT64x" %31s %*s %*s %*s %[^\n]", &map_start, &map_end, perm, name);
+				r_strbuf_appendf (obuf,"0x%"PFMT64x" - 0x%"PFMT64x" %s %s\n", map_start, map_end, perm, name);
+				ptr = nl + 1;
 			}
 			gdbr_close_file (desc);
 			free (buf);
@@ -360,7 +351,7 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
 				}
 				gdbr_lock_leave (desc);
 				return NULL;
-				// TODO "dso"
+			// TODO "dso"
 			default:
 				break;
 			}
@@ -372,14 +363,14 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
 					return NULL;
 				}
 				return strdup (desc->target.regprofile);
-				break;
 			case '8': // "dr8"
 				gdbr_read_registers (desc);
 				if (!desc->data) {
 					return NULL;
 				}
-				char *dr8_str = r_hex_bin2strdup ((const ut8*)desc->data, desc->data_len);
-				return dr8_str;
+				return r_hex_bin2strdup ((const ut8*)desc->data, desc->data_len);
+			default:
+				break;
 			}
 			break;
 		default:
--- a/libr/io/p/io_gdb.c
+++ b/libr/io/p/io_gdb.c
@@ -214,11 +214,17 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
 	}
 	if (cmd[0] == '?' || !strcmp (cmd, "help")) {
 		eprintf ("Usage: =!cmd args\n"
+			 " =!dm              - list memory maps of target process\n"
+			 " =!dr8             - display hexdump of gpr arena\n"
+			 " =!drp             - get register profles\n"
+			 " =!ds              - step one instruction\n"
+			 " =!dsb             - step backwards\n"
+			 " =!dc              - continue execution\n"
+			 " =!dcb             - continue backwards\n"
+			 " =!dck <sig>       - continue execution with sig signal\n"
 			 " =!pid             - show targeted pid\n"
 			 " =!pkt s           - send packet 's'\n"
 			 " =!rd              - show reverse debugging availability\n"
-			 " =!dsb             - step backwards\n"
-			 " =!dcb             - continue backwards\n"
 			 " =!monitor cmd     - hex-encode monitor command and pass"
 			                     " to target interpreter\n"
 			 " =!detach [pid]    - detach from remote/detach specific pid\n"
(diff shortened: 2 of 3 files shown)


libr/debug/p/debug_io.c
```diff
@@ -123,7 +123,13 @@ static int __reg_read(RDebug *dbg, int type, ut8 *buf, int size) {
 
 // "dc" continue execution
 static int __io_continue(RDebug *dbg, int pid, int tid, int sig) {
-	dbg->iob.system (dbg->iob.io, "dc");
+	const char *cmd;
```
radare, line 126: Use the ternary operator

libr/io/p/io_gdb.c
```diff
@@ -229,6 +235,157 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
...
+							eprintf ("Thread (%d) stopped for an invalid reason: %d\n",
+								desc->stop_reason.thread.tid, desc->stop_reason.reason);
+						}
+					}
+					gdbr_invalidate_reg_cache ();
+				}
+				gdbr_lock_leave (desc);
+				return NULL;
+			case 'k': // "dck"
+				if (cmd[3] != '\0') {
```
radare, line 271: Implicit condition. Just drop the !=0

libr/io/p/io_gdb.c
```diff
@@ -229,6 +235,157 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
...
+				break;
+			}
+			break;
+		case 'm': // "dm"
+		{
+			char path[128];
+			ut8 *buf;
+			int ret;
+			// TODO refer debug_gdb.c:r_debug_gdb_map_get
+			ut64 buflen = 16384;
```
radare, line 289: Why 16k

libr/io/p/io_gdb.c
```diff
@@ -229,6 +235,157 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
...
+				if (nl) {
+					*nl = 0;
+					*name = 0;
+					*perm = 0;
+					sscanf (ptr, "%"PFMT64x"-%"PFMT64x" %s %*s %*s %*s %[^\n]",
+						&map_start, &map_end, perm, name);
+					r_strbuf_appendf (obuf,"0x%"PFMT64x" - 0x%"PFMT64x" %s %s\n",
+							  map_start, map_end, perm, name);
+					ptr = nl + 1;
+				} else {
```
radare, line 328: Swap this condition and break earlier and remove on level of indentstion

libr/io/p/io_gdb.c
```diff
@@ -229,6 +235,157 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
...
+					return NULL;
+				}
+				return strdup (desc->target.regprofile);
+				break;
+			case '8': // "dr8"
+				gdbr_read_registers (desc);
+				if (!desc->data) {
+					return NULL;
+				}
+				char *dr8_str = r_hex_bin2strdup ((const ut8*)desc->data, desc->data_len);
```
radare, line 381: Just return. Dont hold it in a var

libr/io/p/io_gdb.c
```diff
@@ -229,6 +235,157 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
...
+			}
+			break;
+		case 'r':
+			switch (cmd[2]) {
+			case 'p': // "drp"
+				if (!desc->target.valid || !desc->target.regprofile) {
+					return NULL;
+				}
+				return strdup (desc->target.regprofile);
+				break;
```
radare, line 375: Unnecessary break

libr/io/p/io_gdb.c
```diff
@@ -229,6 +235,157 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
...
+		{
+			char path[128];
+			ut8 *buf;
+			int ret;
+			// TODO refer debug_gdb.c:r_debug_gdb_map_get
+			ut64 buflen = 16384;
+			snprintf (path, sizeof (path) - 1, "/proc/%d/maps", desc->pid);
+
+#ifdef _MSC_VER
+#define GDB_FILE_OPEN_MODE (_S_IREAD | _S_IWRITE)
```
radare, line 293: Why open in write?

libr/io/p/io_gdb.c
```diff
@@ -229,6 +235,157 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
...
+#define GDB_FILE_OPEN_MODE (S_IRUSR | S_IWUSR | S_IXUSR)
+#endif
+
+			if (gdbr_open_file (desc, path, O_RDONLY, GDB_FILE_OPEN_MODE) < 0) {
+				return NULL;
+			}
+			if (!(buf = malloc (buflen))) {
+				gdbr_close_file (desc);
+				return NULL;
+			}
```
radare, line 304: Why not use the slurp api? Depending on  hardcoded file size is not nice

libr/io/p/io_gdb.c
radare: Why should this depend on msvc?

libr/io/p/io_gdb.c
```diff
@@ -229,6 +235,157 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
...
+				gdbr_close_file (desc);
+				free (buf);
+				return NULL;
+			}
+			buf[ret] = '\0';
+
+			char *ptr = (char *)buf;
+			RStrBuf *obuf = r_strbuf_new (NULL);
+			ut64 map_start, map_end;
+			char perm[32];
```
radare, line 315: Despite being enough i dont like fixed size buffers . Maybe worth to restrict the sscanf read size in the format string

ret2libc:
> **Test plan**
> Currently, there is no way to test remote gdb features until #3574 is implemented

Even if there is no automatic way to test this in CI, could you provide a manual way for us to test the new features? How did you test they were working well?

yossizap:
> **Test plan**
> Currently, there is no way to test remote gdb features until #3574 is implemented

Maybe try launching gdbserver(might require wget) on a test binary in the CI machine with `!`? Might be useful for future tests as well. Linux machines only for now.

Will review later today

libr/io/p/io_gdb.c
abcSup: I am first following what is done in `debug.gdb`, which defines it that way and has a TODO. I will look into it and see how to refactor that.

abcSup:
Thanks everyone for the feedback!

@radare I will look into the fixed buffer and `#ifdef _MSC_VER` stuffs. I took a few shortcut and followed some the original implementation in `debug_gdb`. Those seem weird to me as well.  

> Even if there is no automatic way to test this in CI, could you provide a manual way for us to test the new features? How did you test they were working well?

@ret2libc You are right that was my bad. I have updated the PR to explain how it works and manual ways to test it.

> Maybe try launching gdbserver(might require wget) on a test binary in the CI machine with !? Might be useful for future tests as well. Linux machines only for now.

@yossizap Thanks. I will try to write test in that way for now.

ret2libc:
@radare @yossizap why do we need to duplicate some commands that are already under `d` under `=!` as well?

radare:
Those are called when the debug.io plugin is set

yossizap:
Why would anyone set the debug.io plugin during debug? As I wrote in our discussion in telegram - 

> My impression was that the io commands were only for protocol debug and things that aren't generic(won't fit in `d*`). For example, being able to send a manual gdb packet, or sending reverse step/continue remotely when debugging with RR.

I don't any justification for maintaining these commands in `=!` and `d*`.

radare:
You dont what

This is how you can use the brainfuck debugger or r2frida as debugger backend by just implementing an io plugin.

Other io plugins also handle debug commands. And why would anyone do that/ well. That makes mainly sense for targets that dont have a debug bakcend or when you want to directly interact with the low level or remote commands from the specific io plugin to msnipulate debug things

Its just a very short list of commands that need to be implemented in io to be supported by debug.io

yossizap:
>This is how you can use the brainfuck debugger or r2frida as debugger backend by just implementing an io plugin.

But gdb is already fully integrated with regular debug commands, we don't need these commands for the io plugin.

>That makes mainly sense for targets that dont have a debug bakcend or when you want to directly interact with the low level or remote commands from the specific io plugin to msnipulate debug things

That's what I am saying, gdb already has a debug backend and the io plugin already has low level/remote commands. This PR isn't needed.

>Its just a very short list of commands that need to be implemented in io to be supported by debug.io

No? It's specifically for gdb io, just a duplicate of existing commands.

ret2libc:
Not to mention that it is very confusing even for me to understand what's the real difference between these things.... I don't want to imagine what it would be for a "regular"/new user :/

radare:
i doubt any new user would even notice or care about this.

ret2libc:
yeah so, what's the advantage of point in having two set of commands that do the same thing?

radare:
Io system plugins can implement stuff needed by r2 in there. The usefulness comes when u think out of the box and understand that you can open a remote session of r2 via a plain io and use the remote shell in local. Also, happens when using debug.io for backends that are just based on io. Like the r2frida one. The commands inside the io.system callbsck are not suposed to be complete. Just a small subset for compat. And its not documented or specified but may be good to have this conversation ina different place. This pr is not

libr/io/p/io_gdb.c
```diff
@@ -229,6 +235,148 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
...
+			free (buf);
+			return r_strbuf_drain (obuf);
+		}
+		case 's':
+			switch (cmd[2]) {
+			case '\0': // "ds"
+				gdbr_step (desc, desc->tid);
+				return NULL;
+			case 'b': // "dsb"
+				if (!desc->stub_features.ReverseStep) {
```
XVilka, line 333: This is very nested. Probably better to isolate into the separate function.

libr/io/p/io_gdb.c
```diff
@@ -229,6 +235,148 @@ static char *__system(RIO *io, RIODesc *fd, const char *cmd) {
...
+		case 'c':
+			switch (cmd[2]) {
+			case '\0': // "dc"
+				gdbr_continue (desc, desc->pid, -1, 0);
+				if (desc->stop_reason.is_valid && desc->stop_reason.thread.present) {
+					desc->tid = desc->stop_reason.thread.tid;
+				}
+				return NULL;
+			case 'b': // "dcb"
+				if (!desc->stub_features.ReverseContinue) {
```
XVilka, line 249: This is too nested. Probably better to isolate into the separate function.

PR #16802, opened 2020-05-10
Add pj_reset()

thestr4ng3r:
**Detailed description**

This can be used to re-use the same pj instance for creating multiple jsons without re-allocating the string buffer all the time.

### DIFF
--- a/libr/include/r_util/pj.h
+++ b/libr/include/r_util/pj.h
@@ -19,6 +19,7 @@ typedef struct pj_t {
 /* lifecycle */
 R_API PJ *pj_new(void);
 R_API void pj_free(PJ *j);
+R_API void pj_reset(PJ *j); // clear the pj contents, but keep the buffer allocated to re-use it
 R_API char *pj_drain(PJ *j);
 R_API const char *pj_string(PJ *pj);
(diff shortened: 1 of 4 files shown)


thestr4ng3r:
Test added, use case is r2db.

radare:
So the use case is not recreating new instances for PJ, just reuse the already created one and create multiple strings out of it? i didnt encountered the need for that because creating a new instance is not that heavy. Also what about just making pj.drain() to reset it? also, isnt reset just a fini+init?

thestr4ng3r:
No, fini+init would free the RStrBuf and drain would remove the string out of it:
```c
 R_API void pj_reset(PJ *j); // clear the pj contents, but keep the buffer allocated to re-use it
```
The point is to reuse the buffer.
Usage is here:
https://github.com/thestr4ng3r/r2db/blob/master/src/anal.c#L1254

radare:
Sounds like a tiny optimization for a very corner use case. But its fine for me to merge

thestr4ng3r:
I think it can be quite effective if you serialize 10000 small jsons and re-use the buffer instead of mallocing and reallocing multiple times per json.

PR #16812, opened 2020-05-12
When configuring root RCmdDesc node, parent may be NULL

ret2libc:

**Detailed description**

`parent` of `RCmdDesc` is NULL when creating the root node. This was not detected because the command initialization happen before cfg.level takes place, so the assertion was not printed.

### DIFF
--- a/libr/core/cmd_api.c
+++ b/libr/core/cmd_api.c
@@ -13,10 +13,12 @@ static int value = 0;
 R_LIB_VERSION (r_cmd);
 
 static bool cmd_desc_set_parent(RCmdDesc *cd, RCmdDesc *parent) {
-	r_return_val_if_fail (cd && parent && !cd->parent, false);
-	cd->parent = parent;
-	r_pvector_push (&parent->children, cd);
-	parent->n_children++;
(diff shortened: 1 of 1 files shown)


libr/main/radare2.c
```diff
@@ -392,6 +392,10 @@ R_API int r_main_radare2(int argc, const char **argv) {
 		r_sys_crash_handler ("gdb --pid %d");
 		free (tmp);
 	}
+	if ((tmp = r_sys_getenv ("R_LOG_INIT_LVL"))) {
```
radare, line 395: what about R_LOG_LEVEL

radare: also this seems unrelated to this PR

radare: Document this env var in the manpage and r2 -hh

libr/main/radare2.c
```diff
@@ -392,6 +392,10 @@ R_API int r_main_radare2(int argc, const char **argv) {
...
+		r_log_set_level (atoi (tmp));
```
radare, line 396: are negative values also valid? maybe worth adding a check and an error message if dimmed

PR #16817, opened 2020-05-13
Fix r2r -i and add Tests

thestr4ng3r:
Fixing tests from `r2r -i` was messing up the line numbers when multiple tests were fixed in the same file, resulting in broken tests.

### DIFF
--- a/binr/r2r/r2r.c
+++ b/binr/r2r/r2r.c
@@ -879,7 +879,7 @@ static void replace_cmd_kv_file(const char *path, ut64 line_begin, ut64 line_end
 		eprintf ("Failed to read file \"%s\"\n", path);
 		return;
 	}
-	char *newc = replace_cmd_kv(path, content, line_begin, line_end, key, value, fixup_results);
+	char *newc = replace_cmd_kv (path, content, line_begin, line_end, key, value, fixup_results);
 	free (content);
 	if (!newc) {
 		return;
--- a/binr/r2r/r2r.c
+++ b/binr/r2r/r2r.c
@@ -782,8 +782,8 @@ static char *format_cmd_kv(const char *key, const char *val) {
 	return r_strbuf_drain_nofree (&buf);
 }
 
-static char *replace_lines(char *src, ut64 from, ut64 to, char *news) {
-	char *begin = src;
+static char *replace_lines(const char *src, ut64 from, ut64 to, char *news) {
+	const char *begin = src;
 	ut64 line = 1;
 	while (line < from) {
 		begin = strchr (begin, '\n');
@@ -797,7 +797,7 @@ static char *replace_lines(char *src, ut64 from, ut64 to, char *news) {
 		return NULL;
 	}
 
-	char *end = begin;
+	const char *end = begin;
 	while (line < to) {
 		end = strchr (end, '\n');
 		if (!end) {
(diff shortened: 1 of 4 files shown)


binr/r2r/r2r.c
```diff
@@ -853,20 +853,33 @@ static void fixup_tests(RPVector *results, const char *edited_file, ut64 start_l
...
+	return newc;
+}
+
+static void replace_cmd_kv_file(const char *path, ut64 line_begin, ut64 line_end, const char *key, const char *value, RPVector *fixup_results) {
+	char *content = r_file_slurp (path, NULL);
+	if (!content) {
+		eprintf ("Failed to read file \"%s\"\n", path);
+		return;
+	}
+	char *newc = replace_cmd_kv(path, content, line_begin, line_end, key, value, fixup_results);
```
radare, line 882: missing space

binr/r2r/r2r.c
```diff
@@ -853,20 +853,33 @@ static void fixup_tests(RPVector *results, const char *edited_file, ut64 start_l
...
+		delta++;
+	}
+	fixup_tests (fixup_results, path, line_end, delta);
+	return newc;
+}
+
+static void replace_cmd_kv_file(const char *path, ut64 line_begin, ut64 line_end, const char *key, const char *value, RPVector *fixup_results) {
+	char *content = r_file_slurp (path, NULL);
+	if (!content) {
+		eprintf ("Failed to read file \"%s\"\n", path);
```
radare, line 879: path must be filtered before being printed. i know we dont do this anywhere, but i can think of a proper "printf" that actually honors quotes to escape strings inside and such. no need to change this but its a random though i had

test/unit/test_r2r.c
```diff
@@ -0,0 +1,190 @@
...
+		"RUN", "fixed contents");
+
+	free (content);
+
+	r2r_test_database_free (db);
+	mu_end;
+}
+
+int all_tests() {
+	mu_run_test(test_r2r_database_load_cmd);
```
radare, line 183: missing space before (

binr/r2r/r2r.c
```diff
@@ -853,20 +853,33 @@ static void fixup_tests(RPVector *results, const char *edited_file, ut64 start_l
...
 	}
-	ut64 kv_lines = r_str_char_count (kv, '\n');
+	ut64 kv_lines = r_str_char_count (kv, '\n') + 1;
 	char *newc = replace_lines (content, line_begin, line_end, kv);
 	free (kv);
+	if (!newc) {
+		return NULL;
+	}
+	ut64 lines_before = line_end - line_begin;
+	st64 delta = (st64)kv_lines - lines_before;
```
radare, line 868: i dont think delta can be negative, so make this signed and chk if kv_lines < lines_before and return early in that case

binr/r2r/r2r.c
```diff
@@ -853,20 +853,33 @@ static void fixup_tests(RPVector *results, const char *edited_file, ut64 start_l
...
+	if (line_end == line_begin) {
+		delta++;
+	}
+	fixup_tests (fixup_results, path, line_end, delta);
+	return newc;
+}
+
+static void replace_cmd_kv_file(const char *path, ut64 line_begin, ut64 line_end, const char *key, const char *value, RPVector *fixup_results) {
```
radare, line 876: i think line counts should be size_t instead of ut64

binr/r2r/r2r.c
```diff
@@ -782,8 +782,8 @@ static char *format_cmd_kv(const char *key, const char *val) {
 	return r_strbuf_drain_nofree (&buf);
 }
 
-static char *replace_lines(char *src, ut64 from, ut64 to, char *news) {
-	char *begin = src;
+static char *replace_lines(const char *src, ut64 from, ut64 to, char *news) {
```
radare, line 785: use size_t for lines

binr/r2r/r2r.c
thestr4ng3r: Yes, delta can be negative, for example if you fix from:
```
EXPECT=<<EOF
a
lot
of
broken
stuff
EOF
```
to
```
EXPECT=<<EOF
correct stuff
EOF
```

binr/r2r/r2r.c
thestr4ng3r: The line counts are ut64 everywhere in r2r. The idea is that in theory line numbers could be more that 32bit because they are part of a file on the disk. Of course this will never happen as long as they are slurped into memory but imo it makes more sense this way.

binr/r2r/r2r.c
thestr4ng3r: Agreed

binr/r2r/r2r.c
radare: news is const char *

binr/r2r/r2r.c
radare: ok

binr/r2r/r2r.c
radare: i know, but i think is more correct to go for size_t to count lines, thats why im suggesting you to do the change

binr/r2r/r2r.c
```diff
@@ -966,7 +972,7 @@ static void interact_commands(R2RTestResultInfo *result, RPVector *fixup_results
 		}
 	}
 
-	replace_cmd_kv (result->test->path, test->cmds.line_begin, test->cmds.line_end, "CMDS", newcmds, fixup_results);
+	replace_cmd_kv_file (result->test->path, test->cmds.line_begin, test->cmds.line_end, "CMDS", newcmds, fixup_results);
```
radare, line 975: what about kvfile() instead of kv_file()

thestr4ng3r: File has nothing to do with the kv directly. Theoretically `replacecmdkv_file` would make more sense than `replace_cmd_kvfile`.

binr/r2r/r2r.c
thestr4ng3r: Replaced ut64 by size_t in the places where it is directly operating on memory as opposed to files.

binr/r2r/r2r.c
radare: it is a key-value file. anyway just a matter of taste

binr/r2r/r2r.c
radare: exactly

Issue #9336, opened 2018-02-07
Fix rasign2 tool


XVilka:
Currently code to generate signatures is commented out. Would be nice to port it on top of the current signatures API.
Also expose the FLIRT features, see all commands `z?`.

Maijin:
Why do you assign me to this

radare:
Because you like signatures

PR #16819, opened 2020-05-13
Move rasign2 into RMain

swoops:

**Detailed description**

This is the first steps toward making rasign2 useful. This updates the build so rasign2 will be compiled but not installed to the users path. The rasign2 binary will once again be able to make signature files. This almost closes issue #9336.  

**Test plan**

I pretty much stole the zignature code from `libr/main/radiff.c`. Here are some examples of the tool running:

help menu
```
[rasign2]$ ./rasign2
Usage: rasign2 [options] [file]
 -a [-a]          add extra 'a' to analysis command
 -o sigs.sdb      add signatures to file, create if it does not exist
 -r               show output in radare commands
Examples:
  rasign2 -o libc.sdb libc.so.6
```

Without flags the number of signatures are displayed. This also shows extra analysis.
```
[rasign2]$ ./rasign2 /usr/lib/libc.so.6 
generated zignatures: 2918
```

-r to output radare2 commands, extra analysis used
```
[rasign2]$ ./rasign2 -ar /usr/lib/libc.so.6 |head -n4 
[x] Check for objc references
[x] Check for vtables
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
generated zignatures: 3020
zs *
za sym.fgetsgent_r b f30f1efa4157....
za sym.fgetsgent_r g cc=26 nbbs=43 edges=67 ebbs=1 bbsum=661
za sym.fgetsgent_r o 0x00106b30
```
The -o flag writes a signature file
```# writes a signature file
[rasign2]$ ./rasign2 -aa -o /tmp/out.sdb /usr/lib/libc.so.6 
[x] Check for objc references
[x] Check for vtables
[x] Finding function preludes
[x] Enable constraint types analysis for variables
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
generated zignatures: 3020
```
**Closing issues**

...

### DIFF
--- a/libr/main/rasign2.c
+++ b/libr/main/rasign2.c
@@ -84,7 +84,7 @@ R_API int r_main_rasign2(int argc, const char **argv) {
 	}
 
 	const char *ifile = NULL;
-	if ( opt.ind < argc ) {
+	if ( opt.ind >= argc ) {
 		eprintf("must provide a file\n");
 		return rasign_show_help ();
 	}
--- a/libr/main/rasign2.c
+++ b/libr/main/rasign2.c
@@ -64,7 +64,7 @@ R_API int r_main_rasign2(int argc, const char **argv) {
 	const char *ifile = NULL;
 	int c;
 	int a_cnt = 0;
-	int rad = 0;
+	bool rad = false;
 	RGetopt opt;
 
 	r_getopt_init (&opt, argc, argv, "ao:rvh");
@@ -77,7 +77,7 @@ R_API int r_main_rasign2(int argc, const char **argv) {
 			ofile = opt.arg;
 			break;
 		case 'r':
-			rad = 1;
+			rad = true;
 			break;
 		case 'v':
 			return r_main_version_print ("rasign2");
--- a/libr/main/rasign2.c
+++ b/libr/main/rasign2.c
@@ -53,7 +53,6 @@ static int find_functions(int count, RCore *core) {
 		rasign_show_help ();
 		return -1;
 	}
-	r_cons_flush ();
 	r_core_cmd0 (core, cmd);
 	return 0;
 }
--- a/libr/main/rasign2.c
+++ b/libr/main/rasign2.c
@@ -1,3 +1,4 @@
+/* radare - LGPL - Copyright 2009-2020 - pancake */
 #include <stdio.h>
 #include <string.h>
 #include <r_getopt.h>
--- a/libr/main/rasign2.c
+++ b/libr/main/rasign2.c
@@ -2,7 +2,7 @@
 #include <r_main.h>
 #include <r_core.h>
 
-static int rasign_show_help() {
+static void rasign_show_help() {
 	printf ("Usage: rasign2 [options] [file]\n"
 		" -a [-a]          add extra 'a' to analysis command\n"
 		" -o sigs.sdb      add signatures to file, create if it does not exist\n"
@@ -13,7 +13,6 @@ static int rasign_show_help() {
 		" -h               help menu\n"
 		"Examples:\n"
 		"  rasign2 -o libc.sdb libc.so.6\n");
-	return 0;
 }
 
 static RCore *opencore(const char *fname) {
@@ -79,8 +78,12 @@ R_API int r_main_rasign2(int argc, const char **argv) {
 			break;
 		case 'v':
 			return r_main_version_print ("rasign2");
+		case 'h':
+			rasign_show_help ();
+			return 0;
 		default:
-			return rasign_show_help ();
+			rasign_show_help ();
+			return -1;
 		}
 	}
 
--- a/libr/main/rasign2.c
+++ b/libr/main/rasign2.c
@@ -13,6 +13,10 @@ static int rasign_show_help() {
 		" -a [-a]          add extra 'a' to analysis command\n"
 		" -o sigs.sdb      add signatures to file, create if it does not exist\n"
 		" -r               show output in radare commands\n"
+		" -j               show signatures in json\n"
+		" -q               quiet mode\n"
+		" -v               show version information\n"
+		" -h               help menu\n"
 		"Examples:\n"
 		"  rasign2 -o libc.sdb libc.so.6\n");
 	return 0;
@@ -57,9 +61,11 @@ R_API int r_main_rasign2(int argc, const char **argv) {
 	int c;
 	size_t a_cnt = 0;
 	bool rad = false;
+	bool quiet = false;
+	bool json = false;
 	RGetopt opt;
 
-	r_getopt_init (&opt, argc, argv, "ao:rvh");
+	r_getopt_init (&opt, argc, argv, "ao:rjqvh");
 	while ((c = r_getopt_next (&opt)) != -1) {
 		switch (c) {
 		case 'a':
@@ -70,6 +76,12 @@ R_API int r_main_rasign2(int argc, const char **argv) {
 			break;
 		case 'r':
 			rad = true;
+			break;
+		case 'j':
+			json = true;
+			break;
+		case 'q':
+			quiet = true;
 			break;
 		case 'v':
 			return r_main_version_print ("rasign2");
@@ -98,6 +110,13 @@ R_API int r_main_rasign2(int argc, const char **argv) {
 		return -1;
 	}
 
+	// quiet mode
+	if (quiet) {
+		r_config_set (core->config, "scr.interactive", "false");
+		r_config_set (core->config, "scr.prompt", "false");
+		r_config_set_i (core->config, "scr.color", COLOR_MODE_DISABLED);
+	}
+
 	// run analysis to find functions
 	find_functions (core, a_cnt);
 
@@ -111,6 +130,10 @@ R_API int r_main_rasign2(int argc, const char **argv) {
 
 	if (rad) {
 		r_core_flush (core, "z*");
+	}
+
+	if (json) {
+		r_core_flush (core, "zj");
 	}
 
 	r_core_free (core);
--- a/libr/main/rasign2.c
+++ b/libr/main/rasign2.c
@@ -16,6 +16,7 @@ static void rasign_show_help() {
 }
 
 static RCore *opencore(const char *fname) {
+	RCoreFile * rfile = NULL;
 	RCore *c = r_core_new ();
 	if (!c) {
 		eprintf ("Count not get core\n");
@@ -25,9 +26,14 @@ static RCore *opencore(const char *fname) {
 	r_config_set_i (c->config, "scr.interactive", false);
 	if (fname) {
 #if __WINDOWS__
-		fname = r_acp_to_utf8 (fname);
+		char *winf = r_acp_to_utf8 (fname);
+		rfile = r_core_file_open (c, winf, 0, 0);
+		free (winf);
+#else
+		rfile = r_core_file_open (c, fname, 0, 0);
 #endif
-		if (!r_core_file_open (c, fname, 0, 0)) {
+
+		if (!rfile) {
 			eprintf ("Could not open file %s\n", fname);
 			r_core_free (c);
 			return NULL;
--- a/libr/main/rasign2.c
+++ b/libr/main/rasign2.c
@@ -31,8 +31,8 @@ static RCore *opencore(const char *fname) {
 		fname = r_acp_to_utf8 (fname);
 #endif
 		if (!r_core_file_open (c, fname, 0, 0)) {
-			r_core_free (c);
 			eprintf ("Could not open file %s\n", fname);
+			r_core_free (c);
 			return NULL;
 		}
 		(void)r_core_bin_load (c, NULL, UT64_MAX);
--- a/libr/main/rasign2.c
+++ b/libr/main/rasign2.c
@@ -19,7 +19,6 @@ static int rasign_show_help() {
 }
 
 static RCore *opencore(const char *fname) {
-	const ut64 baddr = UT64_MAX;
 	RCore *c = r_core_new ();
 	if (!c) {
 		eprintf ("Count not get core\n");
@@ -36,7 +35,7 @@ static RCore *opencore(const char *fname) {
 			eprintf ("Could not open file %s\n", fname);
 			return NULL;
 		}
-		(void)r_core_bin_load (c, NULL, baddr);
+		(void)r_core_bin_load (c, NULL, UT64_MAX);
 		(void)r_core_bin_update_arch_bits (c);
 		r_cons_flush ();
 	}
--- a/libr/main/rasign2.c
+++ b/libr/main/rasign2.c
@@ -1,0 +1,119 @@
+#include <stdio.h>
+#include <string.h>
+#include <r_getopt.h>
+#include <r_main.h>
+#include <r_sign.h>
+#include <r_core.h>
+
+#include "r_userconf.h"
+
+static int rasign_show_help() {
+	printf ("Usage: rasign2 [options] [file]\n"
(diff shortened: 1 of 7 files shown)


radare:
Please run r2 with R_DEBUG_ASSERT=1 and a debugger and fix the assert you see in the output

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
+#include <stdio.h>
```
radare, line 1: missing copyright header

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
...
+	r_core_cmd0 (core, cmd);
+	return 0;
+}
+
+R_API int r_main_rasign2(int argc, const char **argv) {
+	const char *ofile = NULL;
+	const char *ifile = NULL;
+	int c;
+	int a_cnt = 0;
+	int rad = 0;
```
radare, line 66: this is bool, not int

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+
+	ifile = (opt.ind < argc)? argv[opt.ind]: NULL;
+	if (ifile == NULL) {
+		return rasign_show_help ();
+	}
+
+	// get the core
+	if ((core = opencore (ifile)) == NULL) {
```
radare, line 95: define and assign core in here. dont use == NULL, put the condition below:
```
RCore *core = opencore (ifile);
if (!core) {
  return -1;
}
```

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
...
+		r_core_free (core);
+		return -1;
+	}
+
+	// create zignatures
+	r_core_cmd0 (core, "zg");
+
+	// write sigs to file
+	if (ofile) {
+		r_core_cmdf (core, "zos %s", ofile);
```
radare, line 110: quote the command like this: `"\"zos %s\"", ofile); otherwise this is a src of command injection bugs

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
...
+	if (fname) {
+#if __WINDOWS__
+		fname = r_acp_to_utf8 (fname);
+#endif
+		if (!r_core_file_open (c, fname, 0, 0)) {
+			r_core_free (c);
+			eprintf ("Could not open file %s\n", fname);
+			return NULL;
+		}
+		(void)r_core_bin_load (c, NULL, baddr);
```
radare, line 38: just pass UT64_MAX here and remove the const variable

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
...
+	RCore *c = r_core_new ();
+	if (!c) {
+		eprintf ("Count not get core\n");
+		return NULL;
+	}
+	r_core_loadlibs (c, R_CORE_LOADLIBS_ALL, NULL);
+	r_config_set_i (c->config, "scr.interactive", false);
+	if (fname) {
+#if __WINDOWS__
+		fname = r_acp_to_utf8 (fname);
```
radare, line 31: this will be leaking because the function returns char * and fname is const char *

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
...
+#endif
+		if (!r_core_file_open (c, fname, 0, 0)) {
+			r_core_free (c);
```
radare, line 34: better eprintf before RCore.free() in case of segfault we want to see the actual error before everything goes wrong

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
...
+			break;
+		case 'v':
+			return r_main_version_print ("rasign2");
+		default:
+			return rasign_show_help ();
+		}
+	}
+
+	ifile = (opt.ind < argc)? argv[opt.ind]: NULL;
+	if (ifile == NULL) {
```
radare, line 90: ifile can only be null in case opt.ind<argc. so just move this condition above and remove the ternary operator

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
...
+			rad = 1;
+			break;
+		case 'v':
+			return r_main_version_print ("rasign2");
+		default:
+			return rasign_show_help ();
+		}
+	}
+
+	ifile = (opt.ind < argc)? argv[opt.ind]: NULL;
```
radare, line 89: define and assign ifile in here

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
...
+			return NULL;
+		}
+		(void)r_core_bin_load (c, NULL, baddr);
+		(void)r_core_bin_update_arch_bits (c);
+		r_cons_flush ();
+	}
+	return c;
+}
+
+static int find_functions(int count, RCore *core) {
```
radare, line 45: should return void, the result is unused

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
...
+	switch (count) {
+	case 0: cmd = "aa"; break;
+	case 1: cmd = "aaa"; break;
+	case 2: cmd = "aaaa"; break;
+	default:
+		eprintf ("Invalid analysis (too many -a's?)\n");
+		rasign_show_help ();
+		return -1;
+	}
+	r_cons_flush ();
```
radare, line 56: flushing before running the command :? weird imho those 2 lines should be swapped

libr/main/rasign2.c
radare, line 45: int count -> size_t count

also, RCore must be the 1st argument

libr/main/rasign2.c
```diff
@@ -0,0 +1,119 @@
...
+		" -a [-a]          add extra 'a' to analysis command\n"
+		" -o sigs.sdb      add signatures to file, create if it does not exist\n"
+		" -r               show output in radare commands\n"
+		"Examples:\n"
+		"  rasign2 -o libc.sdb libc.so.6\n");
+	return 0;
+}
+
+static RCore *opencore(const char *fname) {
+	const ut64 baddr = UT64_MAX;
```
radare, line 21: remove this variable

libr/main/rasign2.c
swoops: Ah, this was left over from old code. The command itself should cause the flush, so I will remove the line.

libr/main/rasign2.c
swoops: I was concerned about this too but was unsure what to do. This code is taken from `radiff2` [here](https://github.com/radareorg/radare2/blob/master/libr/main/radiff2.c#L76).

How do you suggest I resolve this? Add another `#if __WINDOWS__` macro to `free(fname)` after it is used in `r_core_file_open`? Should I change the `const` type?

Is there an OS agnostic function I could use so the string is always allocated with the correct character width, then it can always be free'd. The macro could then be removed.

swoops:
> Please run r2 with R_DEBUG_ASSERT=1 and a debugger and fix the assert you see in the output

This appears to be a bug in `aaa` and independent of this PR.
```
> r2 -v 
radare2 4.5.0-git 24961 @ linux-x86-64 git.4.4.0-172-g7451b6744
commit: 7451b67448ba005886b7f42d8354a6cdf16a605a build: 2020-05-13__19:30:24
> r2 /lib/x86_64-linux-gnu/libc.so.6
 -- V is for Visual
[0x00026f20]> aaa
[x] Check for objc references
[x] Check for vtables
[0x00026f20]> zg
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
WARNING: r_sign_add_anal: assertion 'a && name && size > 0 && bytes' failed (line 736)
generated zignatures: 2497
^D
```
This appears to happen because `aaa` produces functions of zero length. This causes the `size` to fail in the above assertion. I will open a separate issue for this.

ITAYC0HEN:
**Good work!** :rocket: 

Please write a test plan that instructs us, the reviewers, what we should do in order to verify that your code works well. Lead us toward the edge cases you addressed, so we can verify that they're indeed been solved.

For example:

> 1. take the attached file library.so
> 2. execute `./rasign2 library.so`
> 3. make sure that this and that happened
> 4. compare the number of generated signatures to the number of functions detected by `aaa` on this library
> 5. Open in r2 the following stripped binary xyz.bin
> 6. Load the generated signatures files using `... .. ...`
> 7. See that all the found functions are marked,
> etc....

libr/main/rasign2.c
radare: nope, if the string is in the heap and must be freed it should be done like that. you can have two variables, one const and the other non-const, and use one to determine if yo have to free it or not

libr/main/rasign2.c
```diff
@@ -0,0 +1,118 @@
...
+#include <r_getopt.h>
+#include <r_main.h>
+#include <r_sign.h>
+#include <r_core.h>
+
+#include "r_userconf.h"
+
+static int rasign_show_help() {
+	printf ("Usage: rasign2 [options] [file]\n"
+		" -a [-a]          add extra 'a' to analysis command\n"
```
radare, line 13: add -v, -q and -h to the help and to the getopt loop

libr/main/rasign2.c
```diff
@@ -0,0 +1,118 @@
+/* radare - LGPL - Copyright 2009-2020 - pancake */
+#include <stdio.h>
```
radare, line 2: we can probably drop all those, just add r_main.h and r_sign.h and maybe r_core.h.. reduce the imports

libr/main/rasign2.c
```diff
@@ -0,0 +1,118 @@
...
+#include <string.h>
+#include <r_getopt.h>
+#include <r_main.h>
+#include <r_sign.h>
+#include <r_core.h>
+
+#include "r_userconf.h"
+
+static int rasign_show_help() {
```
radare, line 11: add (void)

libr/main/rasign2.c
swoops: Done, and added -j for json.

swoops:
I fixed the windows leak here: https://github.com/swoops/radare2/commit/b5e98b56b5ccdf33e74f75c45e897cd6f40bc59c

No one likes macros so I don't really want to add an `#else` but it seemed the shortest way to fix it without adding worthless junk to the non-windows compiles. 

Alternatively, to avoid any more macros I could do something like this. 
```c
	if (fname) {
		RCoreFile * rfile = NULL;
		char *winf = NULL;
#if __WINDOWS__
		winf = r_acp_to_utf8 (fname);
#endif
		if ( winf ) {
			rfile = r_core_file_open (c, winf, 0, 0);
			free (winf);
		} else {
			rfile = r_core_file_open (c, fname, 0, 0);
		}

		if (!rfile) {
```

swoops:
> Please write a test plan that instructs us, the reviewers, what we should do in order to verify that your code works well. Lead us toward the edge cases you addressed, so we can verify that they're indeed been solved.
> 

Forgive my ignorance, I don't know much about the testing zignatures yet. This code really just uses r_core to run `zg`, `zj`, `z*`, `zos` and `aa`-`aaaa` commands in the proper order. So I won't be of any help regarding zignature or analysis edge cases. 

Any existing zignature tests for `zg`, `zj` and `z*` should be applied to rasign2 and the same output should be expected. If there are not any tests yet, rasign2 might present a simpler harness for them.

That said, here is what I would suggest:

1. Run `rasign -v` and ensure version is printed and 0 exit code.
2. Run `rasign -h` and ensure help is printed and 0 exit code.
3. Run `rasign -qaa lib.so` and ensure the quiet flag is observed. Nothing should be printed to stderr (I think).
3. Run `rasign -jr -o outputfile.sdb lib.so` and parse the output. This uses all 3 output methods at once.
  - 1st found zignatures will be saved to the outputfile.sdb file in sdb format
  - 2nd found zignatures in radare2 format will be printed to stdout
  - 3rd found zignatures in JSON format will  be printed to stdout 
4. Parse all three output formats and ensure each contains the same information. One format should not contain more or less zignatures then another.
7. Repeat steps 3 through 6 but adding `-a`, `-aa` and `-aaa` commands to perform more analysis before creating output.

Currently rasign2 will can create all three output formats at the same time. Since anaysis is the slow part, doing all formats at once would save a lot of time. Let me know if you would like a boundary put between the radare2 command output and JSON output to make parsing easier.

radare:
See the tests/db/tools and add a rasign2 file in there . you can run rasign2 from inside r2 by using the bang! or if you make it work via RMain as an r2 command directly from r2 test that too.

After having some tests i'll merge! thanks!

ITAYC0HEN:
what about updating r2 book? this should come together in parallel to this PR

XVilka:
Having the documentation in the book is good, but out of the scope of this PR. I created an issue in the book repo: https://github.com/radareorg/radare2book/issues/244
For this PR tests are necessary though. Good job!

radare:
I've added a test as an example in here https://github.com/radareorg/radare2/commit/00d933f81244b5a7e2496c4ded77cb7ae006db3a

feel free to submit another PR adding some more tests while you learn more about it :)

thanks!

Issue #12745, opened 2019-01-11
Generic "tree" command/API


XVilka:
Like with tables but for creating hierarchical  output. Might be beneficial in multiple commands, `pf`, `tp`, `dmh`, etc

radare:
what about implementing all that stuff on top of json?

Issue #7519, opened 2017-05-18
Generic "table" command/api


XVilka:
Like for graphs, but for making ASCII/UTF-8 tables instead.

radare:
wat

XVilka:
For example you'll want to make some table in r2 output, something like that:
```
 *   31  30       26 25                                  0
 * +----+-----------+-------------------------------------+
 * | op | 0 0 1 0 1 |                 imm26               |
 * +----+-----------+-------------------------------------+
```
But having a generic API to do that, like graph API/commands will make it easier to improve some r2 commands output. E.g. some commands can have this table output.

radare:
Ah you want an api to do ascii art tables? Ok

XVilka:
This seems useful for prettifying also "dbta" command.

XVilka:
@ret2libc @radare for the table API syntax ideas/changes for the tree-sitter parser.

radare:
uhm we just need to define commands, but the api is there, suggestions?

XVilka:
Postponing to 4.2.0

XVilka:
@ret2libc since you mentioned the table API in the another PR - here the feedback/proposals are welcome.

ret2libc:
@XVilka isn't it already implemented? Why is the issue still open?

XVilka:
@ret2libc API is implemented, but not the commands, like we have the generic commands for graphs for example.

XVilka:
Some API functions are not implemented yet too: https://github.com/radareorg/radare2/blob/master/libr/util/table.c#L858

radare:
@XVilka i have updated the issue text and title to be in sync with the current situation

PR #16824, opened 2020-05-14
Implement the comma command to load,dump,query tables ##util

radare:

**Detailed description**

This command allows to load tables (in ascii art for now, but should support csv and json in the future) and display or filter/query them by using subcommands of it. It aims to be an agn/age replacement for tables (instead of graphs), the JOIN operation can be implemented in this user table

**Test plan**

For example:

```
> af;aflt > $a
> , $a        # load the aflt table
> ,/size/sum  # sum the size of all the functions
```

**Closing issues**

There's one but im lazy to search for it now

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -148,6 +148,20 @@ static const char *help_msg_star[] = {
 	NULL
 };
 
+static const char *help_msg_comma[] = {
+	"Usage:", ",[/] [file]", "# load table data",
+	",", "", "display table",
+	".", "$foo", "aflt > $foo (files starting with '$' are saved in memory)",
+	".", " file.csv", "load the csv file into a table",
+	",-", "", "reset table",
+	",/", "?", "query",
+	",j", "", "print table in json format",
+	",h", "xxd foo bar cow", "define header column names and types",
+	",r", "xxd 1 2 foo", "adds a row using the given format string",
+	",,", "", "print table in csv format",
+	NULL
+};
+
 static const char *help_msg_dot[] = {
 	"Usage:", ".[r2cmd] | [file] | [!command] | [(macro)]", "# define macro or interpret r2, r_lang,\n"
 	"    cparse, d, es6, exe, go, js, lsp, pl, py, rb, sh, vala or zig file",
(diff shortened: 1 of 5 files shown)


XVilka:
Comma will clash with the `CC,` command to link the text files as the comments.

ret2libc:
Can't this wait? It's it really a feature needed now?? We need to implement this in the new parser as well....

Btw, tests missing.

radare:
Obv can wait. Its not merged :p

radare:
The comma should be the suffix to specify table format, aflt should be renamed to afl, and we already user the comma in f and i to list the flags and bininfo in table format.

imho the , is nice becuase it resembles to the dot to load files but this is to interpret them as data, for now it supports ascii art tables, but will support json and csv formats too. this way we can replace excel with r2 :)

radare:
@XVilka @ret2libc @thestr4ng3r please review/add feedback

@ret2libc is there any documentation on how to add this to the new parser?

radare:
@XVilka @ret2libc @thestr4ng3r please review/add feedback

@ret2libc is there any documentation on how to add this to the new parser?

radare:
I tried to add ',' to the tree-sitter newshell but i didnt succed much mainly by how are the commands defined and comma cannot be used as an argument for a macro. What do you think @ret2libc ?

ret2libc:
> I tried to add ',' to the tree-sitter newshell but i didnt succed much mainly by how are the commands defined and comma cannot be used as an argument for a macro. What do you think @ret2libc ?

Honestly, I think we don't need all these commands, as probably you can get much more flexibility and control by doing a simple script yourself....

But I'm pretty sure you are going to implement this anyway, because another command is always good. Anyway, you should probably add `','` to `is_start_of_command` in `radare2-shell-parser/src/scanner.c`.

radare:
A simple script of what for what. Maybe we dont see the same useCases but i dont see any way you can do some stuff without this command.

- import info from diferent tables
- merge, filter and query the data
- create custom tables mike we do with graphs (agn/age)
- support csv, json, ascii art as input and output
- join tables (the only op we cant do with rtable api)
- used for testing table api from commands

I can probably write a script, but we have been looking for years to have this command as well as the table api. To allow the user to create their reports and data directly from r2.

ret2libc:
> A simple script of what for what. Maybe we dont see the same useCases but i dont see any way you can do some stuff without this command.
> 
>     * import info from diferent tables
> 
>     * merge, filter and query the data
> 
>     * create custom tables mike we do with graphs (agn/age)
> 
>     * support csv, json, ascii art as input and output
> 
>     * join tables (the only op we cant do with rtable api)
> 
>     * used for testing table api from commands

What's the purpose of all this? The final purpose, not just "we can't do these things just for the sake of doing them". I'm saying that you can easily do a r2pipe script to do all these things yourselves, since we already provide json for all commands. Probably what we are going to do will be limited anyway in some ways, so people will end up writing their own things anyway IMHO.

ret2libc:
In any case, I'm not going to argue too much about this. We already have 123132131 commands that are used just once in 5 years, so I guess having 123132139 won't change too much :)

Just make sure to implement that in `cfg.newshell` as well.

XVilka:
@ret2libc see:
- https://github.com/radareorg/radare2/issues/7519
- https://github.com/radareorg/radare2/issues/15109
- https://github.com/radareorg/radare2/issues/12745
Radare2 works a lot with graph, table, and tree data. Having a unified API and low-level commands for tables will be beneficial like it was for the graph API.

ret2libc:
> @ret2libc see:
> 
>     * #7519
> 
>     * #15109
> 
>     * #12745
>       Radare2 works a lot with graph, table, and tree data. Having a unified API and low-level commands for tables will be beneficial like it was for the graph API.

I know there are opened issues. They don't really explain the benefit of those new commands though. A 200% of the times, I will just create my own script instead of fighting with learning/making this table thing work well. It seems like we always want to introduce new things.... we are really past the point of "do one thing and do it well".  IMHO all these are unnecessary complications, but as said before, go ahead. Just make sure it works well with newshell as well please.

XVilka:
What I really think is problematic is the usage of a comma as a part of the command. It's really counter intuitive and makes parsing more complicated - e.g. you can't use these commands as a part of the macro, etc.

ret2libc:
> What I really think is problematic is the usage of a comma as a part of the command. It's really counter intuitive and makes parsing more complicated - e.g. you can't use these commands as a part of the macro, etc.

It should not be a problem, because the `,` appears only at the beginning of the command. Also, macros don't use `,` anymore, they use `;` now.

radare:
im happy to pick the 't' suffix, but then we have to change the 't' command which is for types and a lot of 't' subcommands in every other single command. comma is a "special" char that we didnt used, and i undestand it can be a problem for parsing and with combination with other commands. so im open to suggestions here, i just wanted to have a working implementation to have a discussion instead of just theorical concepts and pointless discussions

ret2libc:
They are not pointless... I know you think adding a new command can always be done, but it's never that simple. There is also user experience to consider, but that was never considered :D

XVilka:
Comma commands are pointless, indeed - they are commafull.

radare:
FUCK YOU

trufae:
It would be amazing if we can have some feedback and agreement on which subcommand use for the table output to be consistent across all the commands using this API. I proposed the comma, but i understand the problems that this cause to your treesitter thing.

So please @XVilka @ret2libc @condret and everyone else. share your ideas on this

XVilka:
It's ours tree-sitter thing ;)

ret2libc:
@trufae have you tried what suggested in https://github.com/radareorg/radare2/pull/16824#issuecomment-630001050  to implement it in tree-sitter ?

trufae:
@ret2libc nope i didnt tried, last time i tried to do something with treesitter i had to spend too much time to get very small thing to work. I prefer to wait for that when we decide if comma is the command we want to use for tables, when decided i'll do that and hopefully get this merged at some point

trufae:
Im not sure if comma should be the right command suffix, but i would like to know if this may result on more complex parsing. If u think comma is valid i would like to finish this ultra old PR for once.

trufae:
i thnk the semicolon is the current command separator even for macros.. so maybe it shouldnt such an issue.

trufae:
Ok @ret2libc i did the newshell integration, and it seems to work, lets wait for the CI. anything else blocking this to be merged?

libr/core/cmd.c
```diff
@@ -1260,6 +1274,193 @@ static int cmd_stdin(void *data, const char *input) {
...
+	RList *lines = r_str_split_list (data, "\n", 0);
+	bool expect_header = false;
+	bool expect_rows = false;
+	const char *separator = "|";
+	int ncols = 0;
+	r_list_foreach (lines, iter, line) {
+		if (!expect_rows) {
+			if (r_str_startswith (line, ".--")) {
+				expect_header = true;
+				separator = "|";
```
ret2libc, line 1303: `separator` is never changed.

libr/core/cmd.c
```diff
@@ -1260,6 +1274,193 @@ static int cmd_stdin(void *data, const char *input) {
 	return r_core_run_script (core, "-");
 }
 
+static void load_table(RCore *core, RTable *t, const char *file) {
```
ret2libc, line 1277: A small comment showing how a very simple table looks like? It's hard to review this function without seeing how a table is.

libr/core/cmd.c
trufae: Yes its changing, use an hexadecimal editor instead of a visual one and you will see that | and | are not the same

libr/core/cmd.c
trufae: really? arent tests enough for this?

libr/core/cmd.c
ret2libc: woah :D

libr/core/cmd.c
ret2libc: I don't see any test that shows how a table looks like. Could you point me to one?

trufae: see https://github.com/radareorg/radare2/pull/17701/files#diff-057701ccb34a9d99f8fa2a0c0d53e59c

libr/util/table.c
```diff
@@ -826,6 +826,26 @@ R_API void r_table_filter_columns(RTable *t, RList *list) {
 	}
 }
 
+R_API const char *r_table_help(void) {
```
trufae, line 829: uhm this is duplicated i think.. see the other PR

libr/core/cmd.c
trufae: also test/db/cmd/cmd_table

trufae:
Moved to https://github.com/radareorg/radare2/pull/17702

PR #16825, opened 2020-05-14
Use more size_t in util/str

radare:

**Detailed description**
size_t vs int 

**Test plan**
wait for travis

**Closing issues*

not an issue, just a standard to follow


libr/util/str.c
```diff
@@ -1982,7 +1979,7 @@ R_API char *r_str_ansi_crop(const char *str, ut32 x, ut32 y, ut32 x2, ut32 y2) {
 	return ret;
 }
 
-R_API int r_str_utf8_codepoint (const char* s, int left) {
+R_API size_t r_str_utf8_codepoint(const char* s, size_t left) {
 	bool safe = left >= 0;
```
thestr4ng3r, line 1983: Useless check, as reported by lgtm

XVilka:
`size_t` limits the values to uint32, what happens if you meet 5Gb string from the file compiled on the 64bit computer but in radare2 compiled on 32bit one?

karliss:
> `size_t` limits the values to uint32, what happens if you meet 5Gb string from the file compiled on the 64bit computer but in radare2 compiled on 32bit one?

In the situation you described `const char *` is also limited to 32 bit address space. All of these functions operate with strings that are already in memory. If you really wanted to operate with strings that don't fit into RAM you would need to write functions which accept position inside RIO address space instead of `char*`.

libr/util/str.c
radare: fixed

Issue #16130, opened 2020-03-04
Various issues found by Facebook Infer (from latest master)


XVilka:
See the details in the attached logs

## Buffer overrun checkers

```
Found 7221 issues
                                    Issue Type(ISSUED_TYPE_ID): #
                              Shell Injection(SHELL_INJECTION): 1511
                      Integer Overflow L2(INTEGER_OVERFLOW_L2): 1333
                          Buffer Overrun L3(BUFFER_OVERRUN_L3): 973
                                      Memory Leak(MEMORY_LEAK): 866
                            Null Dereference(NULL_DEREFERENCE): 517
                                        Dead Store(DEAD_STORE): 498
                      Uninitialized Value(UNINITIALIZED_VALUE): 355
            Inferbo Alloc May Be Big(INFERBO_ALLOC_MAY_BE_BIG): 336
                          Buffer Overrun L2(BUFFER_OVERRUN_L2): 329
                          Buffer Overrun L1(BUFFER_OVERRUN_L1): 309
                                Use After Free(USE_AFTER_FREE): 73
                      Integer Overflow L1(INTEGER_OVERFLOW_L1): 62
                          Buffer Overrun S2(BUFFER_OVERRUN_S2): 35
                    Inferbo Alloc Is Big(INFERBO_ALLOC_IS_BIG): 9
                                  Resource Leak(RESOURCE_LEAK): 8
                        Use After Lifetime(USE_AFTER_LIFETIME): 3
  Inferbo Alloc May Be Negative(INFERBO_ALLOC_MAY_BE_NEGATIVE): 2
                            Unreachable Code(UNREACHABLE_CODE): 1
                  Inferbo Alloc Is Zero(INFERBO_ALLOC_IS_ZERO): 1
```
[r2-4.6.0-infer-bo-report.txt.zip](https://github.com/radareorg/radare2/files/5128724/r2-4.6.0-infer-bo-report.txt.zip)

XVilka:
If you think there are any false positives - please write here which are, so we can report them back to the Infer developers in [their repo](https://github.com/facebook/infer/issues).

radare:
how did u got this log? can we have it in CI?

XVilka:
I built and run it myself. It took 56 CPU cores and 36 minutes of time. I doubt we have a suitable CI for this thing.

XVilka:
Updated the report.

XVilka:
Updated the report.

XVilka:
[...]

PR #16826, opened 2020-05-14
Fix bignum issues from facebook infer

FXTi:

**Detailed description**

* Check returned pointer from `calloc`

**Test plan**

...

**Related issues**

https://github.com/radareorg/radare2/issues/16130#issuecomment-628579351

### DIFF
--- a/libr/util/big.c
+++ b/libr/util/big.c
@@ -153,12 +153,11 @@ R_API void r_big_from_hexstr(RNumBig *n, const char *str) {
 
 R_API char *r_big_to_hexstr(RNumBig *b) {
 	r_return_val_if_fail (b, NULL);
-	size_t size;
 
 	int j = R_BIG_ARRAY_SIZE - 1; /* index into array - reading "MSB" first -> big-endian */
-	int i = 0; /* index into string representation. */
-	int k = 0; /* Leading zero's amount */
-	int z;
+	size_t i = 0; /* index into string representation. */
+	size_t k = 0; /* Leading zero's amount */
+	size_t z, last_z = 2 * R_BIG_WORD_SIZE;
 
 	for (; b->array[j] == 0 && j >= 0; j--) {
 	}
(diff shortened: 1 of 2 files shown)


XVilka:
Please fix a typo in the commit message

libr/util/big.c
```diff
@@ -170,6 +168,9 @@ R_API char *r_big_to_hexstr(RNumBig *b) {
 
 	size = 3 + 2 * R_BIG_WORD_SIZE * (j + 1) + ((b->sign > 0)? 0: 1);
 	char *ret_str = calloc (size, sizeof (char));
```
radare, line 170: sizeof(char) is always 1, i dont think we need this indirection. also this shuold be using malloc, because the whole buffer is later written, so no UB can happen

libr/util/big.c
```diff
@@ -180,11 +181,11 @@ R_API char *r_big_to_hexstr(RNumBig *b) {
 	sprintf (&ret_str[i], R_BIG_SPRINTF_FORMAT_STR, b->array[j--]);
```
radare, line 181: dont use sprintf, use snprintf and dont use &ret_str[i], ret_str+i is more readable

radare: also make, i, j, z and k of type size_t

libr/util/big.c
```diff
@@ -180,11 +181,11 @@ R_API char *r_big_to_hexstr(RNumBig *b) {
...
 	for (; ret_str[i + k] == '0' && k < 2 * R_BIG_WORD_SIZE; k++) {
 	}
-	for (z = k; ret_str[i + z]; z++) {
+	for (z = k; ret_str[i + z] && z < 2 * R_BIG_WORD_SIZE; z++) {
```
radare, line 184: this 2 * R_BIG_WORD_SIZE its a bit shady, better use a constant defined before with a meaningful name and use it on both loops

radare: meaningful name: `last_z` for example

libr/util/big.c
radare: define and assign size in the right place

radare: "0x0" is not an hex string

FXTi: > sizeof(char) is always 1, i dont think we need this indirection. also this shuold be using malloc, because the whole buffer is later written, so no UB can happen

No, in testing, `malloc` return a buffer already be written with a very long string. But the leading zero is omitted, so the buffer is not fully written in `r_big_to_hexstr`. In the end, there's some unexpected chars after actual hexstring.


> "0x0" is not an hex string

```python
hex(0) -> '0x0'
```

radare:
If you want to test what malloc do, use openbsd or asan builds. Malloc returns an uninitialized buffer, not a string, that atring was in the previously allocated zone. Calloc is slower than malloc. And when you are filling that buffer i see a null byte at the end. So i dont see the problem. Anyway.. its merged now

PR #16834, opened 2020-05-16
Fix check in r_core_anal_fcn() ##anal

### LINKED ISSUE #16833: r_core_anal_fcn() doesnt analyze some symbols
### Work environment

| OS/arch/bits (mandatory)                             | any
| File format of the file you reverse (mandatory)      | any
| Architecture/bits of the file (mandatory)            | any
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.5.0-git 24650 @ linux-x86-64 git.4.4.0-184-g00d933f81 commit: 00d933f81244b5a7e2496c4ded77cb7ae006db3a build: 2020-05-16__19:39:49

### Expected behavior
`af` command creates function at the given address
### Actual behavior
no new function
1) `radare2 libtest.so`
2) `af @ sym.sym_lvl1`
3) `af @ sym.sym_lvl0`
4) `afl~lvl0` (output is empty) 
`sym.sym_lvl1` just jumps to `sym.sym_lvl0`
image
Since bb related to `sym.sym_lvl0` was already analyzed, execution of `r_core_anal_fcn()` is stopped after this check: 
https://github.com/radareorg/radare2/blob/89cfe05d2d5a3899d4dd846f6e853f9a12769ae1/libr/core/canal.c#L1934-L1939
[libtest.zip](https://github.com/radareorg/radare2/files/4639342/libtest.zip)

Blocker of #7310.

pelijah:

**Detailed description**

...

**Test plan**

...

**Closing issues**

Fixes #16833
...


pelijah:
Nice... "fix" reveals another issue.

libr/core/canal.c
```diff
@@ -1926,12 +1926,21 @@ R_API int r_core_anal_esil_fcn(RCore *core, ut64 at, ut64 from, int reftype, int
 	return 0;
 }
 
+static int find_sym_flag(RAnalFunction *fcn, RFlagItem *f) {
+	return f->space && !strcmp (f->space->name, R_FLAGS_FS_SYMBOLS)? 0: 1;
+}
+
+static bool is_invalid_addr(RCore *core, ut64 addr) {
```
radare, line 1933: just this function just call r_io_is_valid api to determine if its mapped or not? if theres a function or a flag doesnt matters much imho. also i think this slowdowns the analysis

XVilka:
Very red.

libr/core/canal.c
pelijah: IDK who wrote this check and why it prints error "Invalid address...". I just extended it, moved it to a function and named it `is_invalid_addr`.

pelijah:
> 
> 
> Very red.

@XVilka Yeah, because handling of delayed slots in r2 is a crap.

pelijah:
I want to find better binary for test.

libr/core/canal.c
pelijah: I can rename it to avoid confusions. Maybe `skip_addr`?

pelijah: I don't see performance loss on the results of performed tests. The check is involved only when `r_core_anal_fcn()` is called inside `aa`, `af` and `afu`.

ret2libc:
@pelijah is this ready for review?

pelijah:
Yep

pelijah:
Well it would be better to find other solution which will solve #16979 too.

libr/core/canal.c
trufae: what this function is doing is not checking if the address is valid. this is what RIO.is_valid_addr() do.

In this case this is chcking if there's a function or a flag in there. Fix the function name to clarify this and LGTM

trufae:
still red, but thanks for addressing the comments

libr/core/canal.c
```diff
@@ -1926,12 +1926,28 @@ R_API int r_core_anal_esil_fcn(RCore *core, ut64 at, ut64 from, int reftype, int
 	return 0;
 }
 
+static int find_sym_flag(RAnalFunction *fcn, RFlagItem *f) {
```
ret2libc, line 1929: the right signature is `(void *, void *)`

libr/core/canal.c
```diff
@@ -1926,12 +1926,28 @@ R_API int r_core_anal_esil_fcn(RCore *core, ut64 at, ut64 from, int reftype, int
...
+
+static bool is_skippable_addr(RCore *core, ut64 addr) {
+	RAnalFunction *fcn = r_anal_get_fcn_in (core->anal, addr, 0);
+	if (!fcn) {
+		return false;
+	}
+	if (fcn->addr == addr) {
+		return true;
+	}
+	RList *flags = r_flag_get_list (core->flags, addr);
```
ret2libc, line 1941: `const RList *`

XVilka:
@pelijah could you please address the feedback, so we can merge it before the release? It looks important.

pelijah:
To the first commit?

libr/core/canal.c
```diff
@@ -1926,12 +1926,29 @@ R_API int r_core_anal_esil_fcn(RCore *core, ut64 at, ut64 from, int reftype, int
...
+
+static bool is_skippable_addr(RCore *core, ut64 addr) {
+	RAnalFunction *fcn = r_anal_get_fcn_in (core->anal, addr, 0);
+	if (!fcn) {
+		return false;
+	}
+	if (fcn->addr == addr) {
+		return true;
+	}
+	const RList *flags = r_flag_get_list (core->flags, addr);
```
trufae, line 1942: what about using rbin instead of flags to determine if there's a symbol? cc @ret2libc @pelijah ?

trufae:
to the pr title, i just added it. im not 100% for checking flags to find symbols, but at least is faster than iterating over rbin symbol lists.

libr/core/canal.c
ret2libc: I'm not sure how the change would affect the analysis, it should be tried.

XVilka: I suggest we merge this as is, but consider this change in a new PR.

PR #16835, opened 2020-05-17
Fix 31bit shift UB on a signed 32bit variable

radare:
Only reproducible with asan

**Detailed description**

**Test plan**

**Closing issues**

...

### DIFF
--- a/libr/util/utype.c
+++ b/libr/util/utype.c
@@ -103,7 +103,7 @@ R_API char *r_type_enum_getbitfield(Sdb *TDB, const char *name, ut64 val) {
 		if (!(val & n)) {
 			continue;
 		}
-		q = sdb_fmt ("enum.%s.0x%x", name, n)
+		q = sdb_fmt ("enum.%s.0x%x", name, n);
 		res = sdb_const_get (TDB, q, 0);
 		if (isFirst) {
 			isFirst = false;
--- a/libr/util/utype.c
+++ b/libr/util/utype.c
@@ -99,10 +99,11 @@ R_API char *r_type_enum_getbitfield(Sdb *TDB, const char *name, ut64 val) {
 	bool isFirst = true;
 	ret = r_str_appendf (ret, "0x%08"PFMT64x" : ", val);
 	for (i = 0; i < 32; i++) {
-		if (!(val & (1 << i))) {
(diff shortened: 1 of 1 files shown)


libr/util/utype.c
```diff
@@ -99,7 +99,7 @@ R_API char *r_type_enum_getbitfield(Sdb *TDB, const char *name, ut64 val) {
 	bool isFirst = true;
 	ret = r_str_appendf (ret, "0x%08"PFMT64x" : ", val);
 	for (i = 0; i < 32; i++) {
-		if (!(val & (1 << i))) {
+		if (!(val & ((ut32)1 << i))) {
 			continue;
 		}
 		q = sdb_fmt ("enum.%s.0x%x", name, (1<<i));
```
karliss, line 105: Same left shift here and one more lower.

libr/util/utype.c
```diff
@@ -99,7 +99,7 @@ R_API char *r_type_enum_getbitfield(Sdb *TDB, const char *name, ut64 val) {
 	bool isFirst = true;
 	ret = r_str_appendf (ret, "0x%08"PFMT64x" : ", val);
 	for (i = 0; i < 32; i++) {
-		if (!(val & (1 << i))) {
+		if (!(val & ((ut32)1 << i))) {
```
karliss, line 102: Wouldn't it be simpler to use the appropriate unsigned integer literal `1u` instead of cast? Although in the case shifting gets expanded to 64 it's more obvious that `(ut32)` cast needs to be replaced `(ut64)` than changing `1u` to `1ull`. The decision is up to you.

libr/util/utype.c
karliss: Github messed up comment position. I meant the shift in `sdb_fmt`.

libr/util/utype.c
radare: yeah agree, i fixed the thing. chek it now

libr/util/utype.c
karliss: Not resolved , r_type_enum_getbitfield contains two more `(1<<i)` expressions those are as UB as one you initially modified.

radare:
@karliss thanks for the reviews

PR #16838, opened 2020-05-17
Add pxe test + convert prc_256.py to pure r2

kazarmy:

**Detailed description**

This pr adds a test for `pxe`. Also moves and renames scripts.

**Test plan**

**Closing issues**

...


test/db/cmd/cmd_print
```diff
@@ -744,8 +744,8 @@ RUN
 NAME=prc text color
 FILE=-
 CMDS=<<EOF
-. bins/other/palette.r2
-.. bins/src/prc_256.py
+. scripts/palette.r2
+.. scripts/0-to-0xff.py
```
radare, line 748: wtf? why not just use woe

test/scripts/0-to-0xff.py
```diff
@@ -0,0 +1,8 @@
+import sys
```
radare, line 1: i dont think we need python for this. r2 can do that already

test/scripts/palette.r2
```diff
@@ -0,0 +1,65 @@
+ec comment rgb:c50f1f
```
radare, line 1: why this concrete palete and not test one of the one shipped in r2?

test/db/cmd/cmd_print
kazarmy: oh sorry d11f441e29a9b02f11f237037782aa04fb48dc33

test/scripts/0-to-0xff.py
kazarmy: the script was originally for the `prc` test and it wasn't just doing `woe` but also adding flags https://github.com/radareorg/radare2/pull/16838/commits/d11f441e29a9b02f11f237037782aa04fb48dc33#diff-a5c0ff5da9addea0d27f857148c32f57 so unless r2 can do loops, it will have to stay

test/scripts/palette.r2
kazarmy: it's robust against changes to the shipped palettes

test/scripts/prc_256.py
```diff
@@ -0,0 +1,5 @@
+import sys
```
radare, line 1: what about doing this script in pure r2 instead of depending on python?

also im not sure why are we accepting flags with numbers as name, this seems like regression to me. the flagnames should be foo.%02x. also why are the flags needed when px doesnt needs them?

kazarmy: > what about doing this script in pure r2 instead of depending on python?

oh ok 6c95e7a4e95dd2b496f63dc25d573b2c34531ceb ... i think `@@s` has an off-by-one error

> also why are the flags needed when px doesnt needs them?

the flags are needed for the test itself i.e. for its fancy color table ... the hexpairs below are actually flag names:

prc_color_table

> also im not sure why are we accepting flags with numbers as name, this seems like regression to me.

emm why not? ... anyway currently `prc` doesn't cut off any prefixes when displaying flag names ... probably it should, but that's a topic for another pr

radare:
Screenshot 2020-05-17 at 21 05 03

kazarmy:
newshell is green ... red openbsd appears unrelated ... merge?

kazarmy:
ok openbsd back to green ... was watching the log like a hawk lol

PR #16844, opened 2020-05-17
Fix an oobread when pressing the backspace in the r2 shell

radare:
str.c:2043:8: runtime error: addition of unsigned offset to 0x00010c7de67f overflowed to 0x00010c7de67e
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior str.c:2043:8 in

 

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/util/str.c
+++ b/libr/util/str.c
@@ -2036,8 +2036,9 @@ R_API size_t r_str_utf8_charsize(const char *str) {
  */
 R_API size_t r_str_utf8_charsize_prev(const char *str, int prev_len) {
 	r_return_val_if_fail (str, 0);
-	size_t size = 0, pos = 0;
-	while (size < prev_len && size < 5) {
+	int pos = 0;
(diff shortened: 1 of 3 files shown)


libr/util/str.c
```diff
@@ -2036,8 +2036,9 @@ R_API size_t r_str_utf8_charsize(const char *str) {
  */
 R_API size_t r_str_utf8_charsize_prev(const char *str, int prev_len) {
 	r_return_val_if_fail (str, 0);
-	size_t size = 0, pos = 0;
-	while (size < prev_len && size < 5) {
+	int pos = 0;
+	size_t size = 0, maxsize = R_MIN (5, prev_len);
```
ret2libc, line 2040: maybe `minsize` is a better name

radare:
If pos is zero will only read bytes back if the size is >0 so its not a problem because this function works with negative indexes. The ocmmit with the sizet change introduced this regression

PR #16847, opened 2020-05-18
Do not print malformed table if no columns are set

radare:

**Detailed description**

Printing a table in fancy mode results in a glitchy ascii art, this patch makes it not print anything

Screenshot from 2020-05-18 10-52-48

**Test plan**

I have to create some unit tests for it

**Closing issues**

I have spotted this bug while doing the comma(nd)

### DIFF
--- a/libr/util/table.c
+++ b/libr/util/table.c
@@ -278,6 +278,9 @@ static void __computeTotal(RTable *t) {
 }
 
 R_API char *r_table_tofancystring(RTable *t) {
+	if (r_list_length (t->cols) == 0) {
+		return NULL;
+	}
 	RStrBuf *sb = r_strbuf_new ("");
 	RTableRow *row;
 	RTableColumn *col;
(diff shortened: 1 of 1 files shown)


libr/util/table.c
```diff
@@ -278,17 +278,20 @@ static void __computeTotal(RTable *t) {
 }
 
 R_API char *r_table_tofancystring(RTable *t) {
+	if (r_list_length (t->cols) == 0) {
+		return NULL;
```
ret2libc, line 282: What about `""`? I would expect NULL to be returned when there are errors, while no columns looks more like a valid table, just empty.

trufae: more about strdup(""), but as long as strdup can also return NULL, i think it's fine to return null in case there's nothing to be printed or there's an error

trufae:
@ret2libc can you answer my comment?

libr/util/table.c
ret2libc: I think there is a semantic difference between return NULL or "". NULL means there was an error (if strdup returns NULL it means it can't allocate memory, so it is definitely an error), while "" means all was good, but there is nothing to print. As a user of the `r_table_tofancystring` API it is important to distinguish the two things.

trufae: done

ret2libc: done where?

ret2libc:
BTW, I think you are using a command that does not exist. Does `,` exist?

libr/util/table.c
trufae: working with 3 ghusers confuses me sometimes, i pushed this into the trufae fork. its now in here

trufae:
The screenshot is taken from the comma command implemented in another PR that is not merged in master, but you can reproduce the same behaviour with other comma subcommands like f, aflt ... we may probably decide a different char other than comma suffix, (there's an issue for that i think)

trufae:
it happens when no columns are defined

trufae:
ping @ret2libc

PR #16849, opened 2020-05-18
Put ret2libc and radare as owners of cmd.c

ret2libc:


CODEOWNERS
```diff
@@ -15,6 +15,7 @@ libr/bin/format/objc/ @radare
 libr/bin/format/wasm/ @wargio
 libr/config/ @radare @ret2libc
 libr/cons/ @thestr4ng3r @radare
+libr/core/cmd.c @ret2libc
```
kazarmy, line 18:
```suggestion
libr/core/cmd.c @ret2libc @radare 
```

It's important that we don't duplicate existing functionality.

XVilka:
What about `libr/core/cmd_*.c` instead?

ret2libc:
> What about `libr/core/cmd_*.c` instead?

having a single owner for that would be a bit hard right now, as there is a lot of very different code related to each command. I guess we can keep those as they are now.

PR #16921, opened 2020-05-18
Use HtPP for accessing dbg trace entries

pelijah:

**Detailed description**

`radare2 -qq -c "aa;aac;aar;?t aaft" /bin/busybox`: 18.41s vs 18.15s 

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/debug/trace.c
+++ b/libr/debug/trace.c
@@ -1,7 +1,6 @@
 /* radare - LGPL - Copyright 2008-2019 - pancake */
 
 #include <r_debug.h>
-#define R_DEBUG_HT_TRACES 1
 
 // DO IT WITH SDB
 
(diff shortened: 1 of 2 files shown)


XVilka:
What about dropping the old RList and removing the `#ifdef` while you are on it?

ret2libc:
> What about dropping the old RList and removing the `#ifdef` while you are on it?

Yeah the `#ifdef` can probably be removed. For the RList... do traces need to be in the order they are added? Because the Ht does not keep any order.

pelijah:
The list is used by `dtd` command:
https://github.com/radareorg/radare2/blob/f8a9011924c3aff5c8509f6e90bee0326e3745e3/libr/core/cmd_debug.c#L4823-L4864

Issue #16843, opened 2020-05-17
Flags should not start with numbers


radare:
This is something that is valid now for some reason. dunno when this changed but imho flags should not start with digits and we should have tests for this

![Screenshot 2020-12-11 at 13 49 16](https://user-images.githubusercontent.com/6431515/101905387-aea47700-3bb7-11eb-907a-732d83195b7e.png)

![Screenshot 2020-12-11 at 13 50 16](https://user-images.githubusercontent.com/6431515/101905472-d1369000-3bb7-11eb-989d-1cf643ebbb59.png)

Done in https://github.com/radareorg/radare2/pull/18020

thestr4ng3r:
Why not?

radare:
For the same reason we dont like variables starting with numbers.

kazarmy:
Agree

ITAYC0HEN:
> Why not?
Currently, you can define flags which are only numbers (decimal, hex) and they will conflict with decimal and hex number in the commands. How would the program know what the user (or the automation) meant

thestr4ng3r:
> Currently, you can define flags which are only numbers (decimal, hex) and they will conflict with decimal and hex number in the commands. How would the program know what the user (or the automation) meant

Depends on the parser. If we want to support args in commands that can be both a number or a flagname then I agree on this suggestion and I think forbidding leading digits is a good way to do so since you also catch cases like `0x1337`.

radare:
Yes. Thats how r2 worked always with rnum

ITAYC0HEN:
Yes, so we agree :)
That's a good suggestion

XVilka:
Personally, I don't think starting from numbers is a good idea. It's against on how many programming languages work.

ITAYC0HEN:
Yeah we all agree, flags shouldn't start with leading digits.
I improved the title so it will be more clear

karliss:
> Depends on the parser. If we want to support args in commands that can be both a number or a flagname then I agree

The simplest case is expressions. Is `123+321` flag `123` with offset 321, flag `321` with offset 123 or just 444?

thestr4ng3r:
> > Depends on the parser. If we want to support args in commands that can be both a number or a flagname then I agree
> 
> The simplest case is expressions. Is `123+321` flag `123` with offset 321, flag `321` with offset 123 or just 444?

444 is illegal in my country.

trufae:
Fixed in https://github.com/radareorg/radare2/pull/18020

PR #16932, opened 2020-05-21
prc: Remove num. prefix

kazarmy:

**Detailed description**

This pr patches `prc` by having it remove the `num.` prefix if it is encountered in a flag, otherwise current `prc` behavior is preserved (except for single-char flags but that's a minor fix). This was done mainly so that the color table in https://github.com/radareorg/radare2/pull/16838#discussion_r426268189 will still appear even after the restriction of #16843 is imposed. It's possible that there are other situations where (ab)using this is useful.

**Test plan**

**Closing issues**

...

### DIFF
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -687,8 +687,22 @@ static void cmd_prc(RCore *core, const ut8* block, int len) {
 				if (show_flags) {
 					RFlagItem *fi = r_flag_get_i (core->flags, core->offset + j);
 					if (fi) {
-						ch = fi->name[0];
-						ch2 = fi->name[1];
+						size_t fst = 0, snd = 1;
+						if (r_str_startswith (fi->name, "num.")) {
+							fst = 4;
+							snd = 5;
+						}
+						if (fi->name[fst]) {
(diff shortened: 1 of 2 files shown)


ret2libc:
I'm not exactly in favour of filtering random flag names. Does `num.*` has any particular meaning? I would wait to see how https://github.com/radareorg/radare2/issues/16843 will be done before filtering random flags. Right now this does not seem to be necessary, though in theory it should be detected while enforcing the non-leading-num-flags.

kazarmy:
> Does num.* has any particular meaning?

No apart from number, that's why it was chosen.

> I would wait to see how #16843 will be done before filtering random flags. Right now this does not seem to be necessary

Ok. It is quite possible that #16843 need not be done, since we can probably assume that users are smart enough not to define flags with leading numbers unless they know what they are doing, and we can get away with merely printing a warning.

kazarmy:
> though in theory it should be detected while enforcing the non-leading-num-flags.

> It is quite possible that #16843 need not be done ...

By this I meant that we resolve #16843 strictly according to its title, and do "should" instead of "must".

PR #16936, opened 2020-05-22
Fix @@s 'to' param so that it's inclusive

kazarmy:

**Detailed description**

This pr fixes the `@@s` `to` param so that it's inclusive, bringing it in line with its own help message (`run 'x' on all offsets from, to incrementing by step`) and `woe`. Also, UT64_MAX + 1 < UT64_MAX which is mathematically improper.

**Test plan**

**Closing issues**

...

### DIFF
--- a/test/db/cmd/cmd_print
+++ b/test/db/cmd/cmd_print
@@ -747,7 +747,7 @@ CMDS=<<EOF
 . scripts/palette.r2
 woe 0 0xff 1
 (test_flag; f `p8 1`)
-.(test_flag) @@s:0 0x100 1
+.(test_flag) @@s:0 0xff 1
 e scr.color=3
 prc 256
 EOF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -4254,7 +4254,7 @@ R_API int r_core_cmd_foreach(RCore *core, const char *cmd, char *each) {
 				ut64 from = r_num_math (core->num, r_str_word_get0 (str, 0));
 				ut64 to = r_num_math (core->num, r_str_word_get0 (str, 1));
 				ut64 step = r_num_math (core->num, r_str_word_get0 (str, 2));
-				for (cur = from; cur < to; cur += step) {
+				for (cur = from; cur <= to; cur += step) {
 					(void) r_core_seek (core, cur, true);
 					r_core_cmd (core, cmd, 0);
 					if (r_cons_is_breaked ()) {
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -5898,7 +5898,7 @@ DEFINE_HANDLE_TS_FCN_AND_SYMBOL(iter_step_command) {
 	free (step_str);
 
 	ut64 cur;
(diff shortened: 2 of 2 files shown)


test/db/cmd/cmd_print
```diff
@@ -747,7 +747,7 @@ CMDS=<<EOF
 . scripts/palette.r2
 woe 0 0xff 1
 (test_flag; f `p8 1`)
-.(test_flag) @@s:0 0x100 1
```
ret2libc, line 750: mmm @thestr4ng3r @XVilka what do you think about this? I'm actually used, in general, to have [start, end) behaviour about loops. Thought it is also true that this way you can't iterate up to `UT64_MAX`.

kazarmy: @ret2libc Could you clarify what you meant by:

https://github.com/radareorg/radare2/blob/3d2f02be246999bb98190559029bbce21bdd4b28/shlr/radare2-shell-parser/corpus/iter_commands.txt#L164

?

ret2libc: It's also true that `woe` uses inclusive limits :/

ret2libc: > @ret2libc Could you clarify what you meant by:
> 
> https://github.com/radareorg/radare2/blob/3d2f02be246999bb98190559029bbce21bdd4b28/shlr/radare2-shell-parser/corpus/iter_commands.txt#L164
> 
> ?

What do you mean?
That is a tree-sitter test to make sure the newshell is able to parse 
`x @@s:from to step  run 'x' on all offsets from, to incrementing by step`.

So in that case it would run `p8` on `0xd000`, `0xd100`, .... up to 0xe000 (excluded in master, included in your pr)

kazarmy: > So in that case it would run `p8` on `0xd000`, `0xd100`, .... up to 0xe000 (excluded in master, included in your pr)

So if I wanted to go from `0xd000` to `0xe000` inclusive, I would have to do:

```
p8 @@s:0xd000 0xe001 0x100
```
?

ret2libc: right now yes.. I don't have super strong opinions on which way is better

kazarmy: > I'm actually used, in general, to have [start, end) behaviour about loops.

This is not a for loop that we are talking about, it's a range from a to b.

Consider the range `['A', 'Z']`. How are you going to represent it using [start, end)? By using `['A', '[')`? I certainly didn't remember that '[' comes after 'Z'.

> Though it is also true that this way you can't iterate up to UT64_MAX.

I understand that this is pedantic and that possibly nobody is going to need to iterate up to UT64_MAX. But are you 100% sure? What happens if suddenly there is a need for it?

> I don't have super strong opinions on which way is better

The help message says:

```
| x @@s:from to step  run 'x' on all offsets from, to incrementing by step
```

**all offsets** ... please bring the code in line with the help message.

ret2libc: Yeah I guess you were convincing enough :D

PR #16942, opened 2020-05-23
Fix prc handling of single-char flags

kazarmy:

**Detailed description**

This pr fixes the abrupt termination of `prc` output when it encounters a single-char flag (e.g. https://github.com/radareorg/radare2/commit/e57356049669f868fd79522143b3caa694867ed6#diff-8d4ae0f0641439928790561fd63ece2b).

**Test plan**

**Closing issues**

...


libr/core/cmd_print.c
```diff
@@ -686,9 +686,14 @@ static void cmd_prc(RCore *core, const ut8* block, int len) {
 			if (square) {
 				if (show_flags) {
 					RFlagItem *fi = r_flag_get_i (core->flags, core->offset + j);
-					if (fi) {
-						ch = fi->name[0];
-						ch2 = fi->name[1];
+					if (fi && fi->name && fi->name[0]) {
```
ret2libc, line 689: use `!R_STR_ISEMPTY (fi->name)`

ret2libc: btw, when can this happen? I think in theory this should never happen, as `name` should never be NULL nor empty.  Indeed function `r_flag_set` checks `r_return_val_if_fail (f && name && *name, NULL); `. I think that if you found some cases where this is not respected, there is an error somewhere else and we should fix that, not adding `if`s around.

kazarmy: yes in theory it won't happen, both in practical use and with regard to the code, so 4d6bf841f6ed0a71bc09667aa252e6157d01ba29 should resolve this. Even if `name` was NULL, the worse that could happen is a segfault (loss of work) and that appears better than silent failure and possible output corruption.

kazarmy:
However, if `name[0] == '\0'`, then the code goes OOB (of the string) which results in UB. Therefore, I've added an assert. Is that ok with you, @ret2libc?

ret2libc:
> However, if `name[0] == '\0'`, then the code goes OOB (of the string) which results in UB. Therefore, I've added an assert. Is that ok with you, @ret2libc?

It should not happen though. As said before r_flag_set and other functions have an "assert" to check that the name is not NULL and not empty (so name[0] should not be '\0') AFAIU. If that happens, it's probably a bug i'd say, as you don't want a flag with an empty name.

BTW, instead of raw asserts we usually use `r_return_val_if_fail` and similar, which are defined in `r_assert.h`

kazarmy:
oh ok i'll revert

PR #16950, opened 2020-05-25
Do the prefix check after prefix argument has been processed.

karliss:

**Detailed description**

Do the check for need to use rpath based on PREFIX after PREFIX has been set by command line argument.

**Test plan**
* Do `sys/user.sh` in a VM preferably on Ubuntu 18.04 or other distro that uses `--enable-new-dtags` by default
* make sure running ~/bin/prefix/radare2/bin/r2 directly works
* using `readelf -d binary` check that RUNPATH is correctly set for r2 and lib/libr_core.so
* compile Cutter against ~/bin/prefix/radare2 r2 and make sure it opens without problems loading any of r2 libraries

**Closing issues**

### DIFF
--- a/sys/build.sh
+++ b/sys/build.sh
@@ -66,9 +66,6 @@ if [ "`uname`" = Darwin ]; then
 		[ "$(id -u)" = 0 ] || SUDO=sudo
 		[ -n "${NOSUDO}" ] && SUDO=
 	fi
-else
-	[ -n "${PREFIX}" -a "${PREFIX}" != /usr ] && \
-		CFGARG="${CFGARG} --with-rpath"
 fi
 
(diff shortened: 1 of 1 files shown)


ITAYC0HEN:
good catch!

ret2libc:
Just to know, is `sys/build.sh` how you suggest cutter users to install radare2?

karliss:
@ret2libc Not really.  For dev-builds the recommend way is CUTTER_USE_BUNDLED_R2. At the moment there are no recommendations for best way of system wide installation of r2.

But from the issues reported it seems that some people use sys/user.sh -> sys/build.sh which results in runtime linker errors on some systems unless RUNPATH is correctly set during r2 compilation..

Issue #2188, opened 2015-03-16
List local/args in proper order


radare:
Alphabetically sort them.

First list args, and then locals

Probably needs to implement `sdb_array_sort()` in [sdb]( https://github.com/radare/sdb ).

alvarofe:
What does mean this issue?

The local/args that the issue refers are those that belong to a function as the example below and saved in the database alphabetically? If that is, I will try to close the issue. It was only to be sure ;)

```
[0x00002430]> pdf@sym._ssl2_read
╒ (fcn) sym._ssl2_read 1072
│           ; var int local_10_4   @ rbp-0x54
│           ; var int local_12     @ rbp-0x60
│           ; var int local_6      @ rbp-0x30
│           ; var int local_10     @ rbp-0x50
│           ; var int local_0      @ rbp-0x0
```

radare:
yep that is. arguments should be listed first and symbols later, and both should be sorted by delta.

alvarofe:
I'll keep you updated @radare

PR #16953, opened 2020-05-26
Fix half-assed comparators

ulidtko:
> radare2 is WYSIWYF - what you see is what you fix

For those slower users like me, who can't get at it in a single r2 run, smooth `Ps`/`Po` is important.

This patch fixes one annoying bug in that (maybe also several others). Illustrating unit test included!

### DIFF
--- a/libr/anal/var.c
+++ b/libr/anal/var.c
@@ -1092,12 +1092,12 @@ R_API RList *r_anal_function_get_var_fields(RAnalFunction *fcn, int kind) {
 
 static int var_comparator(const RAnalVar *a, const RAnalVar *b){
 	// avoid NULL dereference
-	return (a && b)? a->delta > b->delta: false;
+	return (a && b)? (a->delta > b->delta) - (a->delta < b->delta) : 0;
 }
 
 static int regvar_comparator(const RAnalVar *a, const RAnalVar *b){
 	// avoid NULL dereference
-	return (a && b)? a->argnum > b->argnum: false;
+	return (a && b)? (a->argnum > b->argnum) - (a->argnum < b->argnum): 0;
 }
 
 R_API void r_anal_var_list_show(RAnal *anal, RAnalFunction *fcn, int kind, int mode, PJ *pj) {
--- a/test/unit/test_list.c
+++ b/test/unit/test_list.c
@@ -220,6 +220,44 @@ bool test_r_list_sort5(void) {
 	mu_end;
 }
 
+// 3-valued comparator -> {LT,EQ,GT}.
+static int pintcmp(int* a, int* b) {
+	return (int)(*a > *b) - (int)(*b > *a);
+}
+
+bool test_r_list_mergesort_pint() {
+	// 47 items
+	int data[] = {-440, -468, -444, -80, -568, -564, -396, -404, -436, -420,
+		-428, -388, -356, -324, -292, -464, -260, -252, -204, -196, -212, -76,
+		-160, -540, -216, -536, -532, -148, -116, -560, -556, -244, -460, -448,
+		-236, -156, -228, -456, -552, -548, -544, -220, -180, -188, -84, -172,
+		-164};
+	int expected[] = {-568, -564, -560, -556, -552, -548, -544, -540, -536,
+		-532, -468, -464, -460, -456, -448, -444, -440, -436, -428, -420, -404,
+		-396, -388, -356, -324, -292, -260, -252, -244, -236, -228, -220, -216,
+		-212, -204, -196, -188, -180, -172, -164, -160, -156, -148, -116, -84,
+		-80, -76};
+
+	RList* list = r_list_new();
+	for (size_t i = 0; i < R_ARRAY_SIZE(data); ++i) {
+		r_list_append (list, (void*)&data[i]);
+	}
+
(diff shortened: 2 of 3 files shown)


libr/anal/var.c
```diff
@@ -1092,12 +1092,12 @@ R_API RList *r_anal_function_get_var_fields(RAnalFunction *fcn, int kind) {
 
 static int var_comparator(const RAnalVar *a, const RAnalVar *b){
 	// avoid NULL dereference
-	return (a && b)? a->delta > b->delta: false;
+	return (a && b)? (a->delta > b->delta) - (a->delta < b->delta) : 0;
```
ret2libc, line 1095: Why not:
```suggestion
	return (a && b)? (a->delta - b->delta) : 0;
```
?

libr/anal/var.c
```diff
@@ -1092,12 +1092,12 @@ R_API RList *r_anal_function_get_var_fields(RAnalFunction *fcn, int kind) {
...
 }
 
 static int regvar_comparator(const RAnalVar *a, const RAnalVar *b){
 	// avoid NULL dereference
-	return (a && b)? a->argnum > b->argnum: false;
+	return (a && b)? (a->argnum > b->argnum) - (a->argnum < b->argnum): 0;
```
ret2libc, line 1100: same here

test/unit/test_list.c
```diff
@@ -220,6 +220,53 @@ bool test_r_list_sort5(void) {
 	mu_end;
 }
 
+// To illustrate the original bug.
```
ret2libc, line 223: Although we appreciate a lot unit tests, I'm not sure in this case they make much sense, as the wrong behaviour is not in the RList API, but just in the wrong implementation of the some specific sort comparators. There are already tests that check the correctness of `r_list_sort`.

karliss: It would be good to add a comment clarifying the expected kind of comparison operator expected either next to  r_list_sort declaration or RListComparator typedef. Even though some c strandard library functions expect 3 value comparison operator, using int instead of boolean in c isn't that unusual either and 2 value comparison operator is also used by some libraries especially in C++ standard library. Even worse due to the way r_list_sort is implemented it will work correctly using 2 value comparison operator if the list has less than 43 elements, so it's easy to miss during development.

libr/anal/var.c
karliss: @ret2libc What is the expected value range for delta and argnum? Do we know that they will never exceed half of integer range?

thestr4ng3r:
Sidenote about projects: You may want to check out https://github.com/thestr4ng3r/r2db

libr/anal/var.c
ret2libc: yeah good point actually... maybe it is ok as it was, sorry for the mistake ;)

karliss:
Those null pointer checks look suspicious.  If values can never be null there is no need for them. If they can be null everything is equal to null pointer values likely breaking sorting.

test/unit/test_list.c
ulidtko: Yes, it's exactly the 43-elements threshold why I'm adding another test. The others test on smaller lists.

I can remove the `#if FIXED || true` thing, no problem.

Adding a comment to `RListComparator` is good idea, will do.

ulidtko:
@karliss I agree; and I tested a couple variations of "one argument is null" comparators. Didn't have any effect in this case, obviously, since the list contains no null's.

test/unit/test_list.c
ulidtko: @ret2libc adding a unit test was helpful as an easy approach to isolate the issue, and to confirm the fix is working. It does lose that purpose when the fix is done; however, mergesorting of bigger lists isn't tested at all, so I left it there.

ulidtko: Pardon; "bigger lists *of int ptrs*" that is. 52-strings list is covered in `test_r_list_sort5`.

ulidtko:
Pushed rebased patch.

Re NULL checks:  `r_list_append` ignores NULL "elements", does not insert them. So it looks like *r_list elements are non-null* invariant is actually enforced. From the current `r_list.h` API, I see no ways to violate that invariant.

This null check goes back to 2015; #2188. It's kinda risky to remove it, as such invariants can change in the future — but presently, can be done if you guys insist on that.

test/unit/test_list.c
ret2libc: Ok, thanks!

ret2libc:
> Pushed rebased patch.
> 
> Re NULL checks: `r_list_append` ignores NULL "elements", does not insert them. So it looks like _r_list elements are non-null_ invariant is actually enforced. From the current `r_list.h` API, I see no ways to violate that invariant.
> 
> This null check goes back to 2015; #2188. It's kinda risky to remove it, as such invariants can change in the future — but presently, can be done if you guys insist on that.

IMHO we should wait to remove it right now. I do agree those kinds of checks are ugly and they should not be there, but r_list still allows NULL values AFAIS (e.g. through `r_list_prepend` or `r_list_insert`). I think the right thing to do first would be to add checks there as well, but that's definitely something to be done in a separate PR. This one btw looks good to me.

PR #16955, opened 2020-05-27
Fix ahi help msgs

kazarmy:

**Detailed description**

This pr with regard to the ahi help messages:

- Removes 1 as a numeric base (base s still exists).
- Fixes ahi topline.
- Uses `<base>` instead of `[base]` since it's required.
- Adds some missing items to man ahi help message i.e. `i, p, S`.
- Sorts ahi help items according to main ahi help message `(2, 8, 10, 16, i, p, S, s)`.

**Test plan**

Emm ahi help is more accurate and consistent?

**Closing issues**

...

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -580,7 +580,7 @@ static const char *help_msg_ah[] = {
 	"ahf", " 0x804840", "override fallback address for call",
 	"ahF", " 0x10", "set stackframe size at current offset",
 	"ahh", " 0x804840", "highlight this address offset in disasm",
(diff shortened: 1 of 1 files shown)


kazarmy:
> LGTM, but is there really any check on the number values you can pass to ahi? It seems to me it is just doing a raw r_num_math without additional checks, so in theory you could pass whatever value you want.

No. Unfortunately error checking is low on my TODO list generally, unless the potential for error is high and/or it has irreversible effects. And in this specific case, it appears that unexpected values merely cause the output value to be in hex:

https://github.com/radareorg/radare2/blob/e968c9442af1091ab696f16c8575bc60be3e4586/libr/parse/filter.c#L504-L506

PR #16959, opened 2020-05-27
Fix crash and warnings when opening an iOS/arm64e DSC

oleavr:

**Detailed description**
Running r2 on the `dyld_shared_cache_arm64e` from my iOS/arm64e device running iOS 13.5, I ran into the following issues:
1. It crashes because the `THREADED` opcode is encountered before the usually preceding `SET_SEGMENT_AND_OFFSET_ULEB` opcode. Apple's state machinery handles this case.
2. The reason 1) happens is because parsing doesn't stop when encountering the `DONE` opcode, so it goes ahead and parses “junk”.
3. With 1) and 2) out of the way, a lot of warnings were printed to the console: `WARNING: r_meta_set_with_subtype: assertion 'm && size' failed (line 185)`. This happens because of [this](https://github.com/radareorg/radare2/blob/f255a82a1bb45614a921e3cf156ac37b2b40811d/libr/core/cbin.c#L2206-L2209) logic in `bin_symbols`, which attempts to add an `R_META_TYPE_COMMENT` using a symbol where `symbol->size == 0`.

I've solved these three issues in two separate commits:

1. Improve handling of Mach-O bind opcodes
    * Reset state between regular binds, lazy binds, and weak binds.
    * Stop parsing when encountering the DONE opcode, except for when
      processing lazy binds.
    * Rework handling of THREADED opcode to match Apple's behavior, where
      the segment index isn't used. Also fix the page_count calculation.
    * Eliminate redundant boundary checks to improve clarity.

2. Fix symbol comment logic for zero-sized symbols

**Test plan**
```sh
$ R_DYLDCACHE_FILTER=OSAnalytics r2 dyld_shared_cache_arm64e-ios-13.5-iphone-xsmax
```

**Closing issues**
None that I'm aware of.

### DIFF
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -1540,7 +1540,7 @@ static bool reconstruct_chained_fixup(struct MACH0_(obj_t) *bin) {
 	size_t wordsize = get_word_size (bin);
 	ulebr ur = {NULL};
 	size_t j, count, skip, bind_size;
-	int seg_idx = -1;
+	int seg_idx = 0;
 	ut64 seg_off = 0;
 	bind_size = bin->dyld_info->bind_size;
 	if (!bind_size || bind_size < 1) {
@@ -1564,14 +1564,13 @@ static bool reconstruct_chained_fixup(struct MACH0_(obj_t) *bin) {
 	struct r_dyld_chained_starts_in_segment *cur_seg = NULL;
 	size_t cur_seg_idx = 0;
 	ut8 *end;
-	for (ur.p = opcodes, end = opcodes + bind_size; ur.p < end; ) {
+	bool done = false;
+	for (ur.p = opcodes, end = opcodes + bind_size; !done && ur.p < end;) {
 		ut8 imm = *ur.p & BIND_IMMEDIATE_MASK, op = *ur.p & BIND_OPCODE_MASK;
 		ur.p++;
-		if (ur.p >= end) {
-			break;
-		}
 		switch (op) {
 		case BIND_OPCODE_DONE:
+			done = true;
 			break;
 		case BIND_OPCODE_THREADED: {
 			switch (imm) {
@@ -1580,7 +1579,8 @@ static bool reconstruct_chained_fixup(struct MACH0_(obj_t) *bin) {
 				break;
 			}
 			case BIND_SUBOPCODE_THREADED_APPLY: {
-				if ((!cur_seg || cur_seg_idx != seg_idx) && seg_idx >= 0) {
+				const size_t ps = 0x1000;
+				if (!cur_seg || cur_seg_idx != seg_idx) {
 					cur_seg_idx = seg_idx;
 					cur_seg = bin->chained_starts[seg_idx];
 					if (!cur_seg) {
@@ -1590,8 +1590,8 @@ static bool reconstruct_chained_fixup(struct MACH0_(obj_t) *bin) {
 						}
 						bin->chained_starts[seg_idx] = cur_seg;
 						cur_seg->pointer_format = DYLD_CHAINED_PTR_ARM64E;
-						cur_seg->page_size = 0x1000;
-						cur_seg->page_count = (bin->segs[seg_idx].vmsize + (0x1000 - 1)) & ~(0x1000 - 1);
+						cur_seg->page_size = ps;
+						cur_seg->page_count = ((bin->segs[seg_idx].vmsize + (ps - 1)) & ~(ps - 1)) / ps;
 						if (cur_seg->page_count > 0) {
 							cur_seg->page_start = malloc (sizeof (ut16) * cur_seg->page_count);
 							if (!cur_seg->page_start) {
@@ -1601,7 +1601,7 @@ static bool reconstruct_chained_fixup(struct MACH0_(obj_t) *bin) {
 						}
 					}
 				}
-				ut32 page_index = (ut32)(seg_off / 0x1000);
+				ut32 page_index = (ut32)(seg_off / ps);
 				cur_seg->page_start[page_index] = seg_off & 0xfff;
 				break;
 			}
@@ -2488,16 +2488,16 @@ static int walk_exports(struct MACH0_(obj_t) *bin, RExportsIterator iterator, vo
 	do {
 		RTrieState * state = r_list_get_top (states);
 		ur.p = state->node;
-		ut64 len = ULEB();
+		ut64 len = ULEB ();
 		if (len == UT64_MAX) {
 			break;
 		}
 		if (len) {
-			ut64 flags = ULEB();
+			ut64 flags = ULEB ();
 		if (flags == UT64_MAX) {
 			break;
 		}
-			ut64 offset = ULEB();
+			ut64 offset = ULEB ();
 		if (offset == UT64_MAX) {
 			break;
 		}
@@ -2505,7 +2505,7 @@ static int walk_exports(struct MACH0_(obj_t) *bin, RExportsIterator iterator, vo
 			bool isReexport = flags & EXPORT_SYMBOL_FLAGS_REEXPORT;
 			bool hasResolver = flags & EXPORT_SYMBOL_FLAGS_STUB_AND_RESOLVER;
 			if (hasResolver) {
-				ut64 res = ULEB();
+				ut64 res = ULEB ();
 				if (res == UT64_MAX) {
 					break;
 				}
@@ -2548,7 +2548,7 @@ static int walk_exports(struct MACH0_(obj_t) *bin, RExportsIterator iterator, vo
 				count++;
 			}
 		}
-		ut64 child_count = ULEB();
+		ut64 child_count = ULEB ();
 		if (child_count == UT64_MAX) {
 			goto beach;
 		}
@@ -2572,7 +2572,7 @@ static int walk_exports(struct MACH0_(obj_t) *bin, RExportsIterator iterator, vo
 			R_FREE (next);
 			goto beach;
 		}
-		ut64 tr = ULEB();
+		ut64 tr = ULEB ();
 		if (tr == UT64_MAX) {
 			goto beach;
 		}
@@ -3156,19 +3156,17 @@ static void parse_relocation_info(struct MACH0_(obj_t) *bin, RSkipList * relocs,
 	}
 }
 
+static bool is_valid_ordinal_table_size(ut64 size) {
+	return size > 0 && size <= UT16_MAX;
+}
+
 RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
 	RSkipList *relocs = NULL;
 	RPVector *threaded_binds = NULL;
-	ulebr ur = {NULL};
 	size_t wordsize = get_word_size (bin);
 	if (bin->dyld_info) {
-		ut8 *opcodes,*end, type = 0, rel_type = 0;
-		int lib_ord, seg_idx = -1, sym_ord = -1;
-		char *sym_name = NULL;
-		size_t j, count, skip, bind_size, lazy_size, weak_size;
-		st64 addend = 0;
-		ut64 segmentAddress = 0LL;
-		ut64 addr = 0LL;
+		ut8 *opcodes, rel_type = 0;
+		size_t bind_size, lazy_size, weak_size;
 
 #define CASE(T) case ((T) / 8): rel_type = R_BIN_RELOC_ ## T; break
 		switch (wordsize) {
@@ -3227,44 +3225,63 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
 			return NULL;
 		}
 
-		for (ur.p = opcodes, end = opcodes + amount ; ur.p < end; ) {
-			ut8 imm = *ur.p & BIND_IMMEDIATE_MASK, op = *ur.p & BIND_OPCODE_MASK;
-			++ur.p;
-			if (ur.p >= end) {
-				break;
-			}
-			switch (op) {
-#define ULEB() read_uleb128 (&ur,end)
-#define SLEB() read_sleb128 (&ur,end)
-			case BIND_OPCODE_DONE:
-				break;
-			case BIND_OPCODE_THREADED: {
-				switch (imm) {
-				case BIND_SUBOPCODE_THREADED_SET_BIND_ORDINAL_TABLE_SIZE_ULEB: {
-					ut64 table_size = ULEB();
-					if (table_size == 0 || table_size > 65535) {
-						bprintf ("Error: BIND_SUBOPCODE_THREADED_SET_BIND_ORDINAL_TABLE_SIZE_ULEB size is wrong\n");
-						break;
-					}
-					if (threaded_binds) {
-						r_pvector_free (threaded_binds);
-					}
-					threaded_binds = r_pvector_new ((RPVectorFree) &free);
-					if (threaded_binds) {
-						r_pvector_reserve (threaded_binds, table_size);
-						threaded_binds->v.len = table_size;
-						sym_ord = 0;
+		size_t partition_sizes[] = {bind_size, lazy_size, weak_size};
+		size_t pidx;
+		int opcodes_offset = 0;
+		for (pidx = 0; pidx < R_ARRAY_SIZE (partition_sizes); pidx++) {
+			size_t partition_size = partition_sizes[pidx];
+
+			ut8 type = 0;
+			int lib_ord = 0, seg_idx = -1, sym_ord = -1;
+			char *sym_name = NULL;
+			size_t j, count, skip;
+			st64 addend = 0;
+			ut64 addr = bin->segs[0].vmaddr;
+			ut64 segment_end_addr = addr + bin->segs[0].vmsize;
+
+			ulebr ur = {opcodes + opcodes_offset};
+			ut8 *end = ur.p + partition_size;
+			bool done = false;
+			while (!done && ur.p < end) {
+				ut8 imm = *ur.p & BIND_IMMEDIATE_MASK;
+				ut8 op = *ur.p & BIND_OPCODE_MASK;
+				ur.p++;
+				switch (op) {
+#define ULEB() read_uleb128 (&ur, end)
+#define SLEB() read_sleb128 (&ur, end)
+				case BIND_OPCODE_DONE: {
+					bool in_lazy_binds = pidx == 1;
+					if (!in_lazy_binds) {
+						done = true;
 					}
 					break;
 				}
-				case BIND_SUBOPCODE_THREADED_APPLY:
-					if (threaded_binds) {
-						if (addr && seg_idx >= 0) {
-							ut64 cursor = addr;
-							int n_threaded_binds = (int) r_pvector_len (threaded_binds);
-							while (cursor < segmentAddress) {
+				case BIND_OPCODE_THREADED: {
+					switch (imm) {
+					case BIND_SUBOPCODE_THREADED_SET_BIND_ORDINAL_TABLE_SIZE_ULEB: {
+						ut64 table_size = ULEB ();
+						if (!is_valid_ordinal_table_size (table_size)) {
+							bprintf ("Error: BIND_SUBOPCODE_THREADED_SET_BIND_ORDINAL_TABLE_SIZE_ULEB size is wrong\n");
+							break;
+						}
+						if (threaded_binds) {
+							r_pvector_free (threaded_binds);
+						}
+						threaded_binds = r_pvector_new ((RPVectorFree) &free);
+						if (threaded_binds) {
+							r_pvector_reserve (threaded_binds, table_size);
+							threaded_binds->v.len = table_size;
+							sym_ord = 0;
+						}
+						break;
+					}
+					case BIND_SUBOPCODE_THREADED_APPLY:
+						if (threaded_binds) {
+							int cur_seg_idx = (seg_idx != -1)? seg_idx: 0;
+							size_t n_threaded_binds = r_pvector_len (threaded_binds);
+							while (addr < segment_end_addr) {
 								ut8 tmp[8];
-								ut64 paddr = cursor - bin->segs[seg_idx].vmaddr + bin->segs[seg_idx].fileoff;
+								ut64 paddr = addr - bin->segs[cur_seg_idx].vmaddr + bin->segs[cur_seg_idx].fileoff;
 								bin->rebasing_buffer = true;
 								if (r_buf_read_at (bin->b, paddr, tmp, 8) != 8) {
 									break;
@@ -3296,11 +3313,11 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
 											(struct dyld_chained_ptr_arm64e_rebase *) &raw_ptr;
 									delta = p->next;
 								}
-								if (ordinal >= n_threaded_binds) {
-									bprintf ("Error: Malformed bind chain\n");
-									break;
-								}
 								if (ordinal != -1) {
+									if (ordinal >= n_threaded_binds) {
+										bprintf ("Error: Malformed bind chain\n");
+										break;
+									}
 									struct reloc_t *ref = r_pvector_at (threaded_binds, ordinal);
 									if (!ref) {
 										bprintf ("Error: Inconsistent bind opcodes\n");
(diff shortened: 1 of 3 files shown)


oleavr:
Seeing this test failure:

```
[XX] db/formats/mach0/thumb mach0 non-thumb symbol on thumb bin
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'pi 1@ sym.imp.strcmp' bins/mach0/thumb-test
-- stdout
--- .a	2020-05-27 23:01:32.987922284 +0000
+++ .b	2020-05-27 23:01:32.987922284 +0000
@@ -1 +1 @@
-ldr pc, [reloc.strcmp]
+ldr pc, [segment.DATA]
```

I'm wondering if my assumption about `BIND_OPCODE_DONE` is wrong?

/cc @mrmacete

XVilka:
Still red:
```
[**]                              db/formats/mach0/objc     1461 OK        28 BR        0 XX        1 FX


[XX] db/formats/mach0/swift mach0 swift-x86-64 aav
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'C*~Cd?
aav
C*~Cd?
Cd~?
' bins/mach0/swift-main
-- stdout
--- .a	2020-05-28 03:38:01.808797816 +0000
+++ .b	2020-05-28 03:38:01.808797816 +0000
@@ -1,3 +1,3 @@
-114
-134
+115
+135
134


-- stderr
[ ] Finding xrefs in noncode section with anal.in=io.maps
```

oleavr:
@XVilka Thanks! Investigating.

oleavr:
@XVilka Looks like the test got broken by the `bin_symbols()` introduction of `R_META_TYPE_COMMENT`, but this was hidden because of the bug that I fixed here causing: `WARNING: r_meta_set_with_subtype: assertion 'm && size' failed (line 185)`. Will update the test and take it for a new spin on the CI.

Here's the diff between `C*~Cd` before and after the fix for 3):

```diff
--- before	2020-05-28 05:55:42.000000000 +0200
+++ after	2020-05-28 05:54:55.000000000 +0200
@@ -1,3 +1,4 @@
+CCu base64:U3dpZnQuU3RyaW5nLmluaXQoX2J1aWx0aW5TdHJpbmdMaXRlcmFsOiBCdWlsdGluLlJhd1BvaW50ZXIsIGJ5dGVTaXplOiBCdWlsdGluLldvcmQsIGlzQVNDSUk6IEJ1aWx0aW4uSW50MSkgLT4gU3dpZnQuU3RyaW5n @ 0x10000440a
 Cd 8 @ 0x100005000
 Cd 8 @ 0x100005010
 Cd 8 @ 0x100005018
```

XVilka:
@mrmacete do you approve this PR?

libr/bin/format/mach0/mach0.c
```diff
@@ -3227,44 +3220,61 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
-		for (ur.p = opcodes, end = opcodes + amount ; ur.p < end; ) {
-			ut8 imm = *ur.p & BIND_IMMEDIATE_MASK, op = *ur.p & BIND_OPCODE_MASK;
-			++ur.p;
-			if (ur.p >= end) {
-				break;
-			}
-			switch (op) {
+		int partition_sizes[] = {bind_size, lazy_size, weak_size};
+		int opcodes_offset = 0;
+		for (int partition_idx = 0; partition_idx < R_ARRAY_SIZE (partition_sizes); partition_idx++) {
```
ret2libc, line 3225: do not define this inside the for-loop. We don't use `std=c99` for these files AFAIK, so it would not work on some compilers.

oleavr: Oops, great catch! Fixed.

libr/bin/format/mach0/mach0.c
```diff
@@ -3311,92 +3321,89 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
-								continue;
+				case BIND_OPCODE_SET_DYLIB_ORDINAL_ULEB:
+					lib_ord = ULEB();
+					break;
+				case BIND_OPCODE_SET_DYLIB_SPECIAL_IMM:
+					lib_ord = imm? (st8)(BIND_OPCODE_MASK | imm) : 0;
+					break;
+				case BIND_OPCODE_SET_SYMBOL_TRAILING_FLAGS_IMM: {
+					sym_name = (char*)ur.p;
+					while (*ur.p++ && ur.p<end) {
```
radare, line 3354: spaces around <

libr/bin/format/mach0/mach0.c
```diff
@@ -3311,92 +3321,89 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
-				if (bin->symtab && bin->dysymtab.nundefsym < 0xffff) {
-					for (j = 0; j < bin->dysymtab.nundefsym; j++) {
-						int stridx = 0;
-						int iundefsym = bin->dysymtab.iundefsym;
-						if (iundefsym >= 0 && iundefsym < bin->nsymtab) {
-							int sidx = iundefsym + j;
-							if (sidx < 0 || sidx >= bin->nsymtab) {
-								continue;
+				case BIND_OPCODE_SET_DYLIB_ORDINAL_ULEB:
+					lib_ord = ULEB();
```
radare, line 3347: missing space before (

libr/bin/format/mach0/mach0.c
```diff
@@ -3311,92 +3321,89 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+					break;
+				case BIND_OPCODE_SET_ADDEND_SLEB:
+					addend = SLEB();
+					break;
+				case BIND_OPCODE_SET_SEGMENT_AND_OFFSET_ULEB:
+					seg_idx = imm;
+					if (seg_idx >= bin->nsegs) {
+						bprintf ("Error: BIND_OPCODE_SET_SEGMENT_AND_OFFSET_ULEB"
+							" has unexistent segment %d\n", seg_idx);
+						R_FREE (opcodes);
```
radare, line 3394: just use free, because there's a return nearby and there's no need to nullify the pointer

libr/bin/format/mach0/mach0.c
```diff
@@ -3425,56 +3432,60 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
-				addr += (ut64)imm * (ut64)wordsize + wordsize;
-				break;
-			case BIND_OPCODE_DO_BIND_ULEB_TIMES_SKIPPING_ULEB:
-				count = ULEB();
-				skip = ULEB();
-				for (j = 0; j < count; j++) {
-					if (addr >= segmentAddress) {
-						bprintf ("Error: Malformed ULEB TIMES bind opcode\n");
+				case BIND_OPCODE_DO_BIND:
+					if (!threaded_binds && addr >= segment_end_addr) {
```
radare, line 3436: imho the second condition is >, not >=

libr/bin/format/mach0/mach0.c
```diff
@@ -3425,56 +3432,60 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+					}
+					DO_BIND ();
+					addr += ULEB () + wordsize;
+					break;
+				case BIND_OPCODE_DO_BIND_ADD_ADDR_IMM_SCALED:
+					if (addr >= segment_end_addr) {
+						bprintf ("Error: Malformed IMM SCALED bind opcode\n");
+						goto beach;
+					}
+					DO_BIND();
```
radare, line 3460: missing space

libr/bin/format/mach0/mach0.c
```diff
@@ -3425,56 +3432,60 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+					addr += (ut64)imm * (ut64)wordsize + wordsize;
+					break;
+				case BIND_OPCODE_DO_BIND_ULEB_TIMES_SKIPPING_ULEB:
+					count = ULEB();
```
radare, line 3464: missing space  same in the line below

libr/core/cbin.c
```diff
@@ -2205,7 +2205,7 @@ static int bin_symbols(RCore *r, int mode, ut64 laddr, int va, ut64 at, const ch
 			}
 			if (sn.demname) {
 				r_meta_set (r->anal, R_META_TYPE_COMMENT,
-							addr, symbol->size, sn.demname);
+							addr, symbol->size? symbol->size: 1, sn.demname);
```
radare, line 2208: imho is better to have a separate variable instead of making that line that long.

```
const size_t sym_size = symbol->size > 0? symbol->size: 1;
```

test/db/formats/mach0/swift
```diff
@@ -18,8 +18,8 @@ C*~Cd?
 Cd~?
 EOF
 EXPECT=<<EOF
-114
-134
+115
```
radare, line 21: which is the line that is new? can we have a test that provides an insight of the fix instead of just a row count?

libr/bin/format/mach0/mach0.c
```diff
@@ -3311,92 +3321,89 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+				case BIND_OPCODE_SET_SEGMENT_AND_OFFSET_ULEB:
+					seg_idx = imm;
+					if (seg_idx >= bin->nsegs) {
+						bprintf ("Error: BIND_OPCODE_SET_SEGMENT_AND_OFFSET_ULEB"
+							" has unexistent segment %d\n", seg_idx);
+						R_FREE (opcodes);
+						r_skiplist_free (relocs);
+						r_pvector_free (threaded_binds);
+						return NULL; // early exit to avoid future mayhem
+					} else {
```
radare, line 3398: don't else after return

libr/bin/format/mach0/mach0.c
```diff
@@ -3227,44 +3220,61 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
 		}
 
-		for (ur.p = opcodes, end = opcodes + amount ; ur.p < end; ) {
-			ut8 imm = *ur.p & BIND_IMMEDIATE_MASK, op = *ur.p & BIND_OPCODE_MASK;
-			++ur.p;
-			if (ur.p >= end) {
-				break;
-			}
-			switch (op) {
+		int partition_sizes[] = {bind_size, lazy_size, weak_size}, partition_idx;
```
radare, line 3223: partition_idx must be size_t because it cant be negative

libr/bin/format/mach0/mach0.c
```diff
@@ -1591,7 +1590,7 @@ static bool reconstruct_chained_fixup(struct MACH0_(obj_t) *bin) {
 						bin->chained_starts[seg_idx] = cur_seg;
 						cur_seg->pointer_format = DYLD_CHAINED_PTR_ARM64E;
 						cur_seg->page_size = 0x1000;
-						cur_seg->page_count = (bin->segs[seg_idx].vmsize + (0x1000 - 1)) & ~(0x1000 - 1);
+						cur_seg->page_count = ((bin->segs[seg_idx].vmsize + (0x1000 - 1)) & ~(0x1000 - 1)) / 0x1000;
```
radare, line 1593: maybe better to reuse the pagesize instead of hardcoding the 4KB (0x1000) 3 times. you can defne a variable above like this:

`const size_t ps = 0x1000;` and use `ps` (from `pagesize`) (just trying to make it shorter)

libr/bin/format/mach0/mach0.c
```diff
@@ -3159,16 +3158,10 @@ static void parse_relocation_info(struct MACH0_(obj_t) *bin, RSkipList * relocs,
...
 	size_t wordsize = get_word_size (bin);
 	if (bin->dyld_info) {
-		ut8 *opcodes,*end, type = 0, rel_type = 0;
-		int lib_ord, seg_idx = -1, sym_ord = -1;
-		char *sym_name = NULL;
-		size_t j, count, skip, bind_size, lazy_size, weak_size;
-		st64 addend = 0;
-		ut64 segmentAddress = 0LL;
-		ut64 addr = 0LL;
+		ut8 *opcodes, rel_type = 0;
```
radare, line 3163: it's a pointer, so it must use NULL, not 0

libr/bin/format/mach0/mach0.c
```diff
@@ -3227,44 +3220,61 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
 
-		for (ur.p = opcodes, end = opcodes + amount ; ur.p < end; ) {
-			ut8 imm = *ur.p & BIND_IMMEDIATE_MASK, op = *ur.p & BIND_OPCODE_MASK;
-			++ur.p;
-			if (ur.p >= end) {
-				break;
-			}
-			switch (op) {
+		int partition_sizes[] = {bind_size, lazy_size, weak_size}, partition_idx;
+		int opcodes_offset = 0;
```
radare, line 3224: this variable is never modified. can be removed

libr/bin/format/mach0/mach0.c
```diff
@@ -3227,44 +3220,61 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+		for (partition_idx = 0; partition_idx < R_ARRAY_SIZE (partition_sizes); partition_idx++) {
```
radare, line 3225: i prefer `pidx` as its shorter to read as an iterator variable

libr/bin/format/mach0/mach0.c
```diff
@@ -3227,44 +3220,61 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+			int partition_size = partition_sizes[partition_idx];
```
radare, line 3226: must be size_t

libr/bin/format/mach0/mach0.c
```diff
@@ -3227,44 +3220,61 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+			st64 addend = 0;
+			ut64 addr = bin->segs[0].vmaddr;
+			ut64 segment_end_addr = addr + bin->segs[0].vmsize;
+
+			ulebr ur = {opcodes + opcodes_offset};
+			ut8 *end = ur.p + partition_size;
+			bool done = false;
+			while (!done && ur.p < end) {
+				ut8 imm = *ur.p & BIND_IMMEDIATE_MASK, op = *ur.p & BIND_OPCODE_MASK;
+				++ur.p;
```
radare, line 3241: do not preincrement if its not necessary. `ur.p++`

libr/bin/format/mach0/mach0.c
```diff
@@ -3227,44 +3220,61 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+			size_t j, count, skip;
+			st64 addend = 0;
+			ut64 addr = bin->segs[0].vmaddr;
+			ut64 segment_end_addr = addr + bin->segs[0].vmsize;
+
+			ulebr ur = {opcodes + opcodes_offset};
+			ut8 *end = ur.p + partition_size;
+			bool done = false;
+			while (!done && ur.p < end) {
+				ut8 imm = *ur.p & BIND_IMMEDIATE_MASK, op = *ur.p & BIND_OPCODE_MASK;
```
radare, line 3240: better use two separated variable definitions. it ill be cleaner to read (s/,/;\n/)

libr/bin/format/mach0/mach0.c
```diff
@@ -3227,44 +3220,61 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
 				}
-				case BIND_SUBOPCODE_THREADED_APPLY:
-					if (threaded_binds) {
-						if (addr && seg_idx >= 0) {
-							ut64 cursor = addr;
+				case BIND_OPCODE_THREADED: {
+					switch (imm) {
+					case BIND_SUBOPCODE_THREADED_SET_BIND_ORDINAL_TABLE_SIZE_ULEB: {
+						ut64 table_size = ULEB();
+						if (table_size == 0 || table_size > 65535) {
```
radare, line 3256: use constants or move this logic into a separate function to make the code more readable (`if (is_valid (table_size)) {`). Also use UT16_MAX instead of 65535

libr/bin/format/mach0/mach0.c
```diff
@@ -3227,44 +3220,61 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
 					break;
 				}
-				case BIND_SUBOPCODE_THREADED_APPLY:
-					if (threaded_binds) {
-						if (addr && seg_idx >= 0) {
-							ut64 cursor = addr;
+				case BIND_OPCODE_THREADED: {
+					switch (imm) {
+					case BIND_SUBOPCODE_THREADED_SET_BIND_ORDINAL_TABLE_SIZE_ULEB: {
+						ut64 table_size = ULEB();
```
radare, line 3255: missing space

libr/bin/format/mach0/mach0.c
```diff
@@ -3227,44 +3220,61 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+							r_pvector_reserve (threaded_binds, table_size);
+							threaded_binds->v.len = table_size;
+							sym_ord = 0;
+						}
+						break;
+					}
+					case BIND_SUBOPCODE_THREADED_APPLY:
+						if (threaded_binds) {
+							int cur_seg_idx = (seg_idx != -1)? seg_idx: 0;
 							int n_threaded_binds = (int) r_pvector_len (threaded_binds);
```
radare, line 3274: this cant be negative, must be size_t

libr/bin/format/mach0/mach0.c
```diff
@@ -3227,44 +3220,61 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+						if (threaded_binds) {
+							r_pvector_reserve (threaded_binds, table_size);
+							threaded_binds->v.len = table_size;
+							sym_ord = 0;
+						}
+						break;
+					}
+					case BIND_SUBOPCODE_THREADED_APPLY:
+						if (threaded_binds) {
+							int cur_seg_idx = (seg_idx != -1)? seg_idx: 0;
```
radare, line 3273: this is wrong. seg_idx cant be -1 after your change. but imho the wronginess is in not initializing that index with -1

libr/bin/format/mach0/mach0.c
```diff
@@ -3311,92 +3321,89 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+					while (*ur.p++ && ur.p<end) {
+						/* empty loop */
+					}
+					if (threaded_binds) {
+						break;
+					}
+					sym_ord = -1;
+					if (bin->symtab && bin->dysymtab.nundefsym < 0xffff) {
+						for (j = 0; j < bin->dysymtab.nundefsym; j++) {
+							int stridx = 0;
```
radare, line 3363: maybe initialize with -1 in the case the symbol is not found it will be using the first one (index 0), which is imho wrong

libr/bin/format/mach0/mach0.c
```diff
@@ -3311,92 +3321,89 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
...
+								}
+								stridx = bin->symtab[sidx].n_strx;
+								if (stridx < 0 || stridx >= bin->symstrlen) {
+									continue;
+								}
 							}
-							stridx = bin->symtab[sidx].n_strx;
-							if (stridx < 0 || stridx >= bin->symstrlen) {
-								continue;
+							if (!strcmp ((char *)bin->symstr + stridx, sym_name)) {
```
radare, line 3375: the cast must be const char * ,not char *

libr/bin/format/mach0/mach0.c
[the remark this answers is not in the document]
oleavr: Note that I only changed the indentation level of this code to keep the scope as narrow as possible – but if we're changing these lines we might as well audit them 👌

```C
int stridx = 0;
int iundefsym = bin->dysymtab.iundefsym;
if (iundefsym >= 0 && iundefsym < bin->nsymtab) {
	int sidx = iundefsym + j;
	if (sidx < 0 || sidx >= bin->nsymtab) {
		continue;
	}
	stridx = bin->symtab[sidx].n_strx;
	if (stridx < 0 || stridx >= bin->symstrlen) {
		continue;
	}
}
if (!strcmp ((char *)bin->symstr + stridx, sym_name)) {
	sym_ord = j;
	break;
}
```

So `stridx` is declared and initialized to zero, but never read until it's assigned a new value here:

```C
stridx = bin->symtab[sidx].n_strx;
```

Here, `n_strx` is a `uint32_t`. And on the next line we check:

```C
	if (stridx < 0 || stridx >= bin->symstrlen) {
```

Where `stridx < 0` is impossible, so that should be removed. So we're left with this check:

```C
	if (stridx >= bin->symstrlen) {
```

Where `symstrlen` is an `int`. So ISTM we caan just drop the variable and inline that into the check:

```C
if (bin->symtab[sidx].n_strx >= bin->symstrlen) {
```

I also noticed that `iundefsym` is assigned to an `int` here, also coming from an `uint32_t`, and checking if it's `>= 0` (anything else is impossible), and used to compute `sidx`, also an `int` that cannot be `< 0`, but still there's a check for that.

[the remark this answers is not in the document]
oleavr: Ignore that, `ENOCOFFEE` so I totally misread that code 😂

libr/bin/format/mach0/mach0.c
oleavr: The end address is exclusive, so `>=` means it will reject addresses outside the segment. Does it make sense to allow a bind address that's pointing at the first byte after the segment?

test/db/formats/mach0/swift
oleavr: Previously all lines looked like this:

```
Cd 8 @ 0x100005000
Cd 8 @ 0x100005010
Cd 8 @ 0x100005018
Cd 8 @ 0x100005020
```

Now there's also this line:

```
CCu base64:U3dpZnQuU3RyaW5nLmluaXQoX2J1aWx0aW5TdHJpbmdMaXRlcmFsOiBCdWlsdGluLlJhd1BvaW50ZXIsIGJ5dGVTaXplOiBCdWlsdGluLldvcmQsIGlzQVNDSUk6IEJ1aWx0aW4uSW50MSkgLT4gU3dpZnQuU3RyaW5n @ 0x10000440a
```

So in hindsight we should probably instead make the "grep" pattern stricter so it doesn't result in false positives when `Cd` just happens to be in an unrelated line's base64-encoded data.

oleavr: Fixed it by adding `^` to make this test less brittle.

libr/bin/format/mach0/mach0.c
oleavr: Only `opcodes` is, `rel_type` is a `ut8`.

libr/bin/format/mach0/mach0.c
oleavr: It's modified at the end of the loop.

libr/bin/format/mach0/mach0.c
oleavr: It can here, because here we start out with `-1` like Apple's state machine does. This is used for error-checking for the non-threaded opcodes. But for `reconstruct_chained_fixup()` we don't care about those non-threaded opcodes, so there we initialize it to `0`.

This is a bit confusing because Apple's state machine requires setting the current segment, except for when handling the threaded opcodes – there they don't care about the index and use the segment start + end, which are initially set based on the start/end of the segment with index 0.

oleavr:
@radare Thanks! Believe I've addressed all of the issues except for adding a test.

Test coverage for this seems like it will be really hard as the `dyld_shared_cache` that I reproduced this on is 1.6 GB and not redistributable. I've started a side project with [dyld-tools](https://github.com/oleavr/dyld-tools) that I hope will make it possible to generate a miniature `dyld_shared_cache` that can be added to radare2-testbins and used for writing tests.

I don't know whether it's worth keeping this PR in limbo until I get to that point though, as it might take a long time to get there.

oleavr:
Seems the refactoring may have broken this test:

```
[XX] db/anal/classes anal classes arm64e
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'avrr
acl
acll~\@
avra~0x100008370:0
avra~vmi:0
?e `avraj~{[0]}`~{:
' bins/mach0/TestRTTI-arm64e
-- stdout
--- .a	2020-06-01 18:10:30.279451134 +0000
+++ .b	2020-06-01 18:10:30.279451134 +0000
@@ -15,7 +15,6 @@
virtual_8 @ 0x1000060a0 (vtable + 0x8)
virtual_16 @ 0x10000615c (vtable + 0x10)
Type Info at 0x100008370:
-  Type Info type: __vmi_class_type_info
type: __class_type_info
found_at: 4295000200
class_vtable: 4295000176
```

@mrmacete Anything obvious that I screwed up that could explain this? 😅

XVilka:
Because adding test is not so easy, I suggest we merge this PR without waiting for them, since it fixes the crash.

mrmacete:
LGTM

PR #16966, opened 2020-05-29
Fix pdb cmd inj

GustavoLCR:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


XVilka:
Seems to fail more often:
https://builds.sr.ht/~xvilka/job/221987#task-test-0
```
[XX] db/formats/pdb PDB downloader check
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '!!rabin2 -PP ${R2_FILE} ~PDB' bins/pdb/user32.dll
-- stdout
--- .a	2020-05-29 04:37:20.548603000 +0000
+++ .b	2020-05-29 04:37:20.548687000 +0000
@@ -1 +1 @@
-PDB "user32.pdb" download success
+PDB "user32.pdb" download failed
```

https://github.com/radareorg/radare2/pull/16966/checks?check_run_id=719199580#step:11:130
```
[XX] db/formats/pdb PDB downloader check
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '!!rabin2 -PP ${R2_FILE} ~PDB' bins/pdb/user32.dll
-- stdout
```

libr/bin/pdb/pdb_downloader.c
```diff
@@ -18,89 +23,130 @@ static bool checkExtract() {
...
-	return true;
+#if __WINDOWS__
+	int ret = true;
+	HINTERNET hInternet = InternetOpenA (opt->user_agent, INTERNET_OPEN_TYPE_DIRECT, NULL, NULL, 0);
+	if (!hInternet) {
+		r_sys_perror ("InternetOpenA");
+		free (dir);
+		return false;
+	}
+	char *url = r_str_newf ("%s/%s/%s/%s", opt->symbol_server, opt->dbg_file, opt->guid, file);
```
ret2libc, line 43: isn't this the same as done in the `#else`? If so, it can be extracted before the `#if/#else`.

libr/socket/socket_http.c
```diff
@@ -3,35 +3,41 @@
...
+	size_t i;
+	ut8 c;
+	r_socket_block_time (s, 1, 1, 0);
+	if (r_socket_read_block (s, &c, 1) != 1) {
 		return 0;
 	}
-	for (i = 1; i < bufsz; i += chsz) {
-		buf[i] =0;
+	r_buf_append_bytes (buf, &c, 1);
+	for (i = 1;; i += 1) {
```
ret2libc, line 14: What if the server keep sending data? Can this result in infinite allocation of data on the client side?

libr/socket/socket_http.c
```diff
@@ -83,7 +115,7 @@ static char *r_socket_http_answer (RSocket *s, int *code, int *rlen) {
 
 R_API char *r_socket_http_get(const char *url, int *code, int *rlen) {
 	char *curl_env = r_sys_getenv ("R2_CURL");
-	if (curl_env && *curl_env) {
+	if (curl_env && atoi (curl_env)) {
```
HoundThe, line 118: Why the `atoi (curl_env)`? Maybe `atoi (*curl_env)` ?

GustavoLCR:
> Do you have some reproducer steps to test the various things you fixed?

@ret2libc updated pr description with test steps.

Also, newshell is failing this:
`%R2_CURL=1;idpd`
But this works
`%R2_CURL=1 ;idpd`

ret2libc:
> > Do you have some reproducer steps to test the various things you fixed?
> 
> @ret2libc updated pr description with test steps.
> 
> Also, newshell is failing this:
> `%R2_CURL=1;idpd`
> But this works
> `%R2_CURL=1 ;idpd`

Thanks for reporting this, it should be fixed with https://github.com/radareorg/radare2/pull/16988 .

libr/socket/socket_http.c
```diff
@@ -3,45 +3,81 @@
...
-	// r_socket_block_time (s, 1, 1, 0);
-	if (r_socket_read_block (s, (ut8 *) buf, 1) != 1) {
+#if __WINDOWS__
+#include <WinInet.h>
+#endif
+
+static size_t __socket_slurp(RSocket *s, RBuffer *buf) {
+	size_t i;
+	ut8 c;
+	r_socket_block_time (s, 1, 1, 0);
```
ret2libc, line 13: why this `r_socket_block_time` has different parameters than the one in the for-loop?

ret2libc: Also, if both calls can use the same parameters you can probably just put everything inside the for-loop.

libr/socket/socket.c
```diff
@@ -391,18 +341,39 @@ R_API bool r_socket_connect(RSocket *s, const char *host, const char *port, int
...
 			return false;
 		}
 		s->sfd = SSL_new (s->ctx);
 		SSL_set_fd (s->sfd, s->fd);
-		if (SSL_connect (s->sfd) != 1) {
-			r_socket_free (s);
+		int ret = SSL_connect (s->sfd);
+		if (ret != 1) {
+			int error = SSL_get_error (s->sfd, ret);
+			while (ret && (error == 2 || error == 3)) {
```
ret2libc, line 357: probably a hard-limit here would be beneficial to avoid infinite loops.

libr/socket/socket.c
```diff
@@ -260,80 +260,30 @@ R_API bool r_socket_spawn(RSocket *s, const char *cmd, unsigned int timeout) {
...
 	WSADATA wsadata;
-	TIMEVAL Timeout;
-	Timeout.tv_sec = timeout;
-	Timeout.tv_usec = 0;
 
 	if (WSAStartup (MAKEWORD (1, 1), &wsadata) == SOCKET_ERROR) {
 		eprintf ("Error creating socket.");
 		return false;
 	}
-	s->fd = socket (AF_INET, SOCK_STREAM, 0);
```
ret2libc, line 274: why removing all this code? Was it not working? Does this function still work without all this? I don't have a Windows machine at the moment to check.

libr/bin/pdb/pdb_downloader.c
```diff
@@ -18,89 +18,77 @@ static bool checkExtract() {
...
 	}
-	return true;
+	char *url = r_str_newf ("%s/%s/%s/%s", opt->symbol_server, opt->dbg_file, opt->guid, file);
+	int len;
+	char *file_buf = r_socket_http_get (url, NULL, &len);
+	free (url);
+	if (len && file_buf && *file_buf) {
+		char *path = r_str_newf ("%s%s%s", dir, R_SYS_DIR, opt->dbg_file);
+		FILE *f = fopen (path, "wb");
+		free (dir);
```
ret2libc, line 37: `dir` should be freed outside of this `if`, otherwise it may leak.

libr/bin/pdb/pdb_downloader.c
```diff
@@ -18,89 +18,77 @@ static bool checkExtract() {
...
+	if (!r_sys_mkdirp (dir)) {
+		free (dir);
 		return false;
 	}
-	return true;
+	char *url = r_str_newf ("%s/%s/%s/%s", opt->symbol_server, opt->dbg_file, opt->guid, file);
+	int len;
+	char *file_buf = r_socket_http_get (url, NULL, &len);
+	free (url);
+	if (len && file_buf && *file_buf) {
```
ret2libc, line 34: use `!R_STR_ISEMPTY` or `R_STR_ISNOTEMPTY`

ret2libc: I'd make the error path nested, and leave the "main part" of the function outside the if, but that's just a personal taste actually :)

```suggestion
	if (!len || R_STR_ISEMPTY (file_buf)) {
		free (file_buf);
		return false;
	}
	char *path = ...
```

libr/bin/pdb/pdb_downloader.c
```diff
@@ -18,89 +18,77 @@ static bool checkExtract() {
...
-	archive_name_escaped = r_str_escape (archive_name);
-	user_agent = r_str_escape (opt->user_agent);
-	symbol_server = r_str_escape (opt->symbol_server);
 
-	abspath_to_archive = r_str_newf ("%s%s%s%s%s%s%s",
-			    symbol_store_path, R_SYS_DIR,
-			    dbg_file, R_SYS_DIR,
-			    guid, R_SYS_DIR,
-			    archive_name_escaped);
+	char *abspath_to_file = r_str_newf ("%s%s%s%s%s%s%s",
```
ret2libc, line 62: why did you remove the `r_str_escape` calls? I think we should at least keep a `r_str_escape (abspath_to_file)` because otherwise I think there may be injections when executing the `extractor_cmd`

libr/util/sys.c
```diff
@@ -680,20 +681,29 @@ R_API int r_sys_cmd_str_full(const char *cmd, const char *input, char **output,
...
-				if (len) {
-					*len += bytes;
+				char *tmp = realloc (outputptr, out_len + bytes + 1);
+				if (!tmp) {
+					R_FREE (outputptr);
+					break;
 				}
-				outputptr = r_str_append (outputptr, buffer);
+				outputptr = tmp;
+				memcpy (outputptr + out_len, buffer, bytes);
```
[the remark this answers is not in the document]
GustavoLCR: `r_str_append` will stop at null bytes, so the output gets chopped

test/db/formats/pdb
```diff
@@ -1,6 +1,9 @@
 NAME=PDB downloader check
 FILE=bins/pdb/user32.dll
-CMDS=!!rabin2 -PP ${R2_FILE} ~PDB
+CMDS=<<EOF
```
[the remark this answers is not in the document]
GustavoLCR: CI doesn`t build with SSL enabled, so this test would just fail

libr/socket/socket.c
GustavoLCR: it worked, but it was basically the same as the posix version, so i just removed the common part and used the posix code.
It still works fine, was able to connect to a gdbserver and a rap instance normally

libr/bin/pdb/pdb_downloader.c
```diff
@@ -18,150 +18,110 @@ static bool checkExtract() {
 	return true;
 }
 
-static bool checkCurl() {
-	const char nul[] = R_SYS_DEVNULL;
-	if (r_sys_cmdf ("curl --version > %s", nul) != 0) {
+static bool download_and_write(SPDBDownloaderOpt *opt, const char *file) {
+	char *dir = r_str_newf ("%s%s%s%s%s",
```
ret2libc, line 22: do `dbg->file` and `guid` come from the user or from the binary? If from the latter, they should be assumed untrusted so we should escape them.

libr/bin/pdb/pdb_downloader.c
```diff
@@ -18,150 +18,110 @@ static bool checkExtract() {
...
-	abspath_to_archive = r_str_newf ("%s%s%s%s%s%s%s",
-			    symbol_store_path, R_SYS_DIR,
-			    dbg_file, R_SYS_DIR,
-			    guid, R_SYS_DIR,
-			    archive_name_escaped);
-
-	abspath_to_file = strdup (abspath_to_archive);
-	abspath_to_file[strlen (abspath_to_file) - 1] = 'b';
+
+	char *abspath_to_file = r_str_newf ("%s%s%s%s%s%s%s",
```
ret2libc, line 61: actually.. I just realized that we should also check individual parts instead of only escaping the entire result string, because otherwise you may end up writing outside of the intended directory (e.g. if the dbg_file is `../../../../etc/passwd` or something like that)

libr/util/str.c
```diff
@@ -1351,6 +1351,36 @@ R_API char *r_str_escape(const char *buf) {
...
+	r_return_val_if_fail (buf, NULL);
+	char *new_buf = malloc (1 + strlen (buf) * 2);
+	if (!new_buf) {
+		return NULL;
+	}
+	const char *p = buf;
+	char *q = new_buf;
+	while (*p) {
+		switch (*p) {
+#if __WINDOWS__
```
ret2libc, line 1365: so, in case of WINDOWS, a string like `hello "world"` becomes `hello "\"world"\"`? Is that expected? BTW, it would be awesome if you could as well add a short unit test for this in `unit/test/test_str.c`, if that helps you checking the output of this function.

libr/util/str.c
```diff
@@ -1351,6 +1351,36 @@ R_API char *r_str_escape(const char *buf) {
...
+		case '"':
+			*q++ = '"';
+#else
+		case '$':
+		case '"':
+		case '`':
+		case '\\':
+			*q++ = '\\';
```
ret2libc, line 1373: if fallthrough to the next case is expected, add `/* FALLTHRU */`

libr/util/sys.c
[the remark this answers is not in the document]
ret2libc: I see, good point! :+1:

test/db/formats/pdb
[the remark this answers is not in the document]
ret2libc: Right, that's a pity :(

libr/bin/pdb/pdb_downloader.c
GustavoLCR: dbg_file gets stripped with `r_file_basename`, so it should be safe, and for the guid I added `is_valid_guid` that only allows hex digits.

libr/bin/pdb/pdb_downloader.c
GustavoLCR: see https://github.com/radareorg/radare2/pull/16966#discussion_r435585946

libr/bin/pdb/pdb_downloader.c
```diff
@@ -191,6 +151,19 @@ void deinit_pdb_downloader(SPDBDownloader *pd) {
 	pd->download = 0;
 }
 
+static bool is_valid_guid(const char *guid) {
```
ret2libc, line 154: GUID are 32 chars long, aren't they? I'm not sure this function works correctly.
e.g.:
```
is_valid_guid('12345678901234567890123456789012') should be true
is_valid_guid('123456789012345678901234567890123') should be false
```
I think that currently this function does not return the right values.

libr/bin/pdb/pdb_downloader.c
```diff
@@ -18,150 +18,110 @@ static bool checkExtract() {
...
-		archive_name_escaped[strlen (archive_name_escaped) - 1] = 'b';
-
-		curl_cmd = r_str_newf ("curl -sfLA \"%s\" \"%s/%s/%s/%s\" --create-dirs -o \"%s\"",
-		                       opt->user_agent,
-		                       opt->symbol_server,
-		                       opt->dbg_file,
-		                       opt->guid,
-		                       archive_name_escaped,
-		                       abspath_to_file);
 		eprintf ("Attempting to download uncompressed pdb in %s\n", abspath_to_file);
```
ret2libc, line 118: is it still true that the code below downloads the file in `abspath_to_file`?

libr/bin/pdb/pdb_downloader.c
```diff
@@ -18,150 +18,110 @@ static bool checkExtract() {
...
-		if ((cmd_ret = r_sys_cmd (curl_cmd) != 0)) {
-			eprintf("curl exited with error %d\n", cmd_ret);
-			res = 0;
+		res = download_and_write (opt, opt->dbg_file);
+		if (!res) {
```
ret2libc, line 120: how is the download of `opt->dbg_file` related to the removal of the file `abspath_to_file`?

libr/bin/pdb/pdb_downloader.c
GustavoLCR: tested it, function works correctly

libr/bin/pdb/pdb_downloader.c
GustavoLCR: removed, wasnt necessary

libr/bin/pdb/pdb_downloader.c
GustavoLCR: yes, first try is `abspath_to_archive`, second is to `abspath_to_file`

ret2libc: Ah, I see now! The computation is done separately two times, one in `download_and_write` and another one in this function.

libr/bin/pdb/pdb_downloader.c
ret2libc: Mmm, for me it doesn't. I'm using this test, where `is_valid_guid` is copied-pasted from your PR.
```C
#include <stdio.h>
#include <stdbool.h>
#include <ctype.h>

static bool is_valid_guid(const char *guid) {
	if (!guid) {
		return false;
	}
	size_t i;
	for (i = 0; guid[i] && i <= 33 ; i++) {
		if (!isxdigit (guid[i])) {
			return false;
		}
	}
	return i == 33;
}

static void test_s(const char *s) {
	if (is_valid_guid (s)) {
		printf("'%s' is valid guid\n", s);
	} else {
		printf("'%s' is NOT valid guid\n", s);
	}
}

	test_s("12345678901234567890123456789012");
	test_s("123456789012345678901234567890123");
	return 0;
}
```
If I run this program, it says:
```
'12345678901234567890123456789012' is NOT valid guid
'123456789012345678901234567890123' is valid guid
```
but the expected output is:
```
'12345678901234567890123456789012' is valid guid
'123456789012345678901234567890123' is NOT valid guid
```
Please let me know if I'm missing something.

libr/socket/socket_http.c
```diff
@@ -122,7 +220,7 @@ R_API char *r_socket_http_get(const char *url, int *code, int *rlen) {
 	host += 3;
 	port = strchr (host, ':');
 	if (!port) {
-		port = ssl? "443": "80";
+		port = ssl ? "443" : "80";
```
ret2libc, line 223: very minor stuff, but this was alright as it was.

libr/socket/socket_http.c
```diff
@@ -147,17 +245,22 @@ R_API char *r_socket_http_get(const char *url, int *code, int *rlen) {
...
 		response = NULL;
 	}
 	free (uri);
 	r_socket_free (s);
 	return response;
+#endif
+}
+
+R_API char *r_socket_http_get(const char *url, int *code, int *rlen) {
+	return socket_http_get_recursive (url, code, rlen, 5);
```
ret2libc, line 260: please define a macro somewhere named something like `R_SOCKET_HTTP_MAX_REDIRECTS`

libr/socket/socket_http.c
```diff
@@ -3,45 +3,72 @@
...
 		return 0;
 	}
-	for (i = 1; i < bufsz; i += chsz) {
-		buf[i] =0;
-		r_socket_block_time (s, 1, 0, 1000);
-		int olen = r_socket_read_block (s, (ut8 *) buf + i , chsz);
-		if (olen != chsz) {
-			buf[i] = 0;
+	r_socket_block_time (s, 1, 0, 1000);
+	for (i = 0; i < 0x2000; i += 1) {
```
ret2libc, line 16: please extract this in a macro, something like `R_SOCKET_MAX_ANSWER_LENGTH`

libr/socket/socket_http.c
```diff
@@ -71,48 +99,118 @@ static char *r_socket_http_answer (RSocket *s, int *code, int *rlen) {
...
+	HINTERNET hOpenUrl = InternetOpenUrlA (hInternet, url, NULL, 0, 0, 0);
+	if (!hOpenUrl) {
+		r_sys_perror ("InternetOpenUrlA");
+		InternetCloseHandle (hInternet);
+		return NULL;
+	}
+
+	char *ret = NULL;
+	size_t read_sz = 0x100000;
+	DWORD read = 0, written = 0;
```
trufae, line 128: dont use `read` as variable names because it is already a posix symbol and makes the reading confusing, not to mention that deleting the variable definition will make crash to happen and no compile time errors, you can use a `struct {size_t r, w;} stats` and do stats.r+= read for example, or just use `r` and `w`

libr/socket/socket_http.c
```diff
@@ -3,45 +3,72 @@
...
+	/* Follow redirects */
+	p = r_str_casestr (buf, "Location:");
+	if (p) {
+		if (!redirections) {
+			eprintf ("Too many redirects\n");
+			goto exit;
+		}
+		p += strlen ("Location:");
+		int url_len = strchr (p, '\n') - p;
+		char *url = r_str_ndup (p, url_len);
```
trufae, line 64: strchr can return NULL, and then url_len will be negative, and the whole thing will break.

libr/socket/socket_http.c
```diff
@@ -3,45 +3,72 @@
...
+	int ret, len = 0, delta = 0;
+	char *dn;
+	RBuffer *b = r_buf_new ();
+	if (!b) {
 		return NULL;
 	}
 	char *res = NULL;
-	int olen = __socket_slurp (s, (ut8*)buf, bufsz);
+	size_t olen = socket_slurp (s, b);
+	char *buf = malloc (olen + 1);
```
trufae, line 41: check for null

libr/socket/socket.c
```diff
@@ -794,28 +761,24 @@ R_API int r_socket_read(RSocket *s, unsigned char *buf, int len) {
...
-#endif
 }
 
 R_API int r_socket_read_block(RSocket *s, ut8 *buf, int len) {
 	int ret = 0;
 	for (ret = 0; ret < len; ) {
 		int r = r_socket_read (s, buf + ret, len - ret);
 		if (r == -1) {
+#if HAVE_LIB_SSL
+			if (SSL_get_error (s->sfd, r) == 2) {
```
trufae, line 776: whats the meaning of error == 2? can we use a constant/define or comment?

libr/socket/socket.c
```diff
@@ -678,19 +654,10 @@ R_API int r_socket_flush(RSocket *s) {
...
-#if __UNIX__
-	//int msecs = (1000 * secs) + (usecs / 1000);
-	int msecs = (usecs / 1000);
-	struct pollfd fds[1];
-	fds[0].fd = s->fd;
-	fds[0].events = POLLIN | POLLPRI;
-	fds[0].revents = POLLNVAL | POLLHUP | POLLERR;
-	return poll ((struct pollfd *)&fds, 1, msecs);
-#elif __WINDOWS__
+#if __UNIX__ || __WINDOWS__
```
trufae, line 660: isnt this the same as `#if 1` ? :D i dont think we cover more platforms than unix (variants) and windows. just remove the #if block

libr/socket/socket.c
```diff
@@ -391,18 +341,44 @@ R_API bool r_socket_connect(RSocket *s, const char *host, const char *port, int
...
 		}
 		s->sfd = SSL_new (s->ctx);
 		SSL_set_fd (s->sfd, s->fd);
-		if (SSL_connect (s->sfd) != 1) {
-			r_socket_free (s);
+		int ret = SSL_connect (s->sfd);
+		if (ret != 1) {
+			int error = SSL_get_error (s->sfd, ret);
+			int tries = 10;
+			while (tries && ret && (error == 2 || error == 3)) {
```
trufae, line 358: same here, whats the meaning of error2 and error3?

libr/socket/socket.c
```diff
@@ -391,18 +341,44 @@ R_API bool r_socket_connect(RSocket *s, const char *host, const char *port, int
...
+				struct timeval tv;
```
trufae, line 359: inititalize tv_sec and usec in the same struct definition

trufae: = {1,0}

libr/bin/pdb/pdb_downloader.c
```diff
@@ -191,6 +148,19 @@ void deinit_pdb_downloader(SPDBDownloader *pd) {
 	pd->download = 0;
 }
 
+static bool is_valid_guid(const char *guid) {
+	if (!guid) {
+		return false;
+	}
+	size_t i;
+	for (i = 0; guid[i] && i <= 33 ; i++) {
```
trufae, line 156: use a define/constant instead of 33 to make the loop condition more readable

libr/bin/pdb/pdb_downloader.c
GustavoLCR: turns out guid string has appended to it the "age", so assuming it even has a fixed length was wrong, fixed it now

libr/bin/pdb/pdb_downloader.c
```diff
@@ -18,150 +18,107 @@ static bool checkExtract() {
...
+static bool download_and_write(SPDBDownloaderOpt *opt, const char *file) {
+	char *dir = r_str_newf ("%s%s%s%s%s",
+		opt->symbol_store_path, R_SYS_DIR,
+		opt->dbg_file, R_SYS_DIR,
+		opt->guid);
+	if (!r_sys_mkdirp (dir)) {
+		free (dir);
 		return false;
 	}
+	char *url = r_str_newf ("%s/%s/%s/%s", opt->symbol_server, opt->dbg_file, opt->guid, file);
```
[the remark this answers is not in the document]
GustavoLCR: thats the url, not a path

libr/bin/pdb/pdb_downloader.c
ret2libc: mh, that looks weird, but it seems it is something a bit more deeper in the existing code. Guid are 32 chars AFAIK. Nevermind.

PR #16970, opened 2020-05-29
ahi 10 does signed decimal with 32-bit gp regs + ahi 10u for unsigned decimal

kazarmy:

**Detailed description**

This pr does 2 things:

1. `ahi 10` or `ahi d` does signed decimal if it detects an x86 32-bit register (see https://github.com/radareorg/radare2/compare/master...kazarmy:ahi-10-signed-ahi-10u?expand=1#diff-0147cd662e9a03b59a8deb98c5e3b400R375 for an example).
1. `ahi 10u` or `ahi du` forces unsigned decimal if needed.

Method used is somewhat hacky but since this is a deliberate hint, it isn't enabled by default and the hint can be removed. Unfortunately only supporting x86 for now -- support for other archs is somewhere in the distant future.

**Test plan**

**Closing issues**

...

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -580,7 +580,7 @@ static const char *help_msg_ah[] = {
 	"ahf", " 0x804840", "override fallback address for call",
 	"ahF", " 0x10", "set stackframe size at current offset",
 	"ahh", " 0x804840", "highlight this address offset in disasm",
-	"ahi", "[?] 10", "define numeric base for immediates (2, 8, 10, 16, i, p, S, s)",
+	"ahi", "[?] 10", "define numeric base for immediates (2, 8, 10, 10u, 16, i, p, S, s)",
 	"ahj", "", "list hints in JSON",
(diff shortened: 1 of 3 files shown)


ret2libc:
Unfortunately the RParse code is a mess and it is hard to do stuff properly... In theory, there is RReg that could be used to get the 32-bit sized registers of any architecture. Do you think it would be possible to use RReg to get only the 32-bit sized register names instead of hard-coding them?

kazarmy:
> Unfortunately the RParse code is a mess and it is hard to do stuff properly

This should be a disassembler option, but since there is more than 1 choice for the disassembler, it is going to be rather time consuming to add such an option if it doesn't exist or to enable it if it does, for each disassembler.

> Do you think it would be possible to use RReg to get only the 32-bit sized register names instead of hard-coding them?

I have just found out that it is possible 77f332f9d78fc2d3bb4aee21655d6c701445d4ac. Thanks for the hint!

PR #16974, opened 2020-05-30
Fix non-working drt all command

### LINKED ISSUE #16634: drtj all broken
# Work environment

| OS/arch/bits (mandatory)                             |  Ubuntu x86 64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.5.0-git 1 @ linux-x86-64 git.4.5.0-git commit: dae5bb43ee8e3c6ba9964caf6569ce28918dd000 build: 2020-04-16__23:30:26

### Expected behavior
So the command `drtj all` used to return all registers r2 knows about, as a json object. At some point, this stopped working and it has broken my angr symbion integration.

### Actual behavior
```
drt?
Usage: drt   [type] [size]    # debug register types
| drt                List all available register types
| drt [size]         Show all regs in the profile of size
| drt 16             Show 16 bit registers
| drt [type]         Show all regs in the profile of this type
| drt all            Show all registers
| drt fpu            Show fpu registers
| drt [type] [size]  Same as above for type and size
| drt [type] [size]  Same as above for type and size
| drt*               List flags in r commands
[0x7f1634629090]> drt all
[0x7f1634629090]> drtj all
```

 - Just try to use `drtj all` on any binary. In my case, the example is the vanilla /bin/ls.

abcSup:

**Detailed description**
`r_debug_reg_list` does not handle `type == R_REG_TYPE_ALL` case. This PR supports the new case by cloning and joining the returned `RList` (`RRegItem`) from `r_reg_get_list` into the `head` variable. The implementation uses `r_list_clone` so it does not mess with the original `RList` when doing `r_list_join`.

**Test plan**
Check `drt all` or `drtj all` commands are producing outputs. Currently, this cannot be implemented in a command test because of the non-constant register values in `rsp` and others and a unit test is hard for this. I am open to any suggestions on this.

**Closing issues**

closes #16634


libr/debug/dreg.c
```diff
@@ -83,6 +83,7 @@ R_API int r_debug_reg_sync(RDebug *dbg, int type, int write) {
 }
 
 R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char *use_color) {
+	int i;
```
trufae, line 86: use size_t for loops, there's no need for negative indexes

libr/debug/dreg.c
```diff
@@ -128,9 +129,27 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char
...
 	int itmidx = -1;
 	dbg->creg = NULL;
-	head = r_reg_get_list (dbg->reg, type);
-	if (!head) {
-		return false;
+	if (type == R_REG_TYPE_ALL) {
+		head = r_list_new ();
+		for (i = 0; i < R_REG_TYPE_LAST; i++) {
+			RList *tmp = r_reg_get_list (dbg->reg, i);
+			if (!tmp || !r_list_length (tmp)) {
```
trufae, line 136: if length is 0 it will leak

libr/debug/dreg.c
```diff
@@ -128,9 +129,27 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char
...
+				continue;
+			}
+			tmp = r_list_clone (tmp);
+			if (!r_list_join (head, tmp)) {
+				r_list_free (head);
+				r_list_free (tmp);
+				return false;
```
trufae, line 143: this function returns int, not bool. so it will be good to change the return type to `bool` and fix the rest of return statemnts to return n != 0 instead of n

libr/debug/dreg.c
```diff
@@ -128,9 +129,27 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char
...
-	if (!head) {
-		return false;
+	if (type == R_REG_TYPE_ALL) {
+		head = r_list_new ();
+		for (i = 0; i < R_REG_TYPE_LAST; i++) {
+			RList *tmp = r_reg_get_list (dbg->reg, i);
+			if (!tmp || !r_list_length (tmp)) {
+				continue;
+			}
+			tmp = r_list_clone (tmp);
```
trufae, line 139: dont override the same variable, use an intermediate one and check for null before going forward

libr/debug/dreg.c
trufae: then r_list_free() should not free the items, and its better to free the iters than leaking them

trufae:
no need to change where this info is printed in this PR, just thinking out loudly.

abcSup:
> Another question. is this fixing art all?
No, it does not.
 
> Also, use the ##debug tag in the commit message, and change dtr with drt
Gotcha. I just noticed the typo and will change it now.

> no need to change where this info is printed in this PR, just thinking out loudly.
Ok, thanks for the feedback!

libr/debug/dreg.c
abcSup: I mean the returned `RList` itself is still in use, so freeing it without setting the corresponding `RRegSet.regs` to `NULL` will cause a use-after-free. Let me know if I understand this wrongly.

libr/debug/dreg.c
ret2libc: Why was this marked as resolved @abcSup ?

libr/debug/dreg.c
ret2libc: Here as well, why resolved? Also, consider moving this inside the `if`, since it is only used there.

libr/debug/dreg.c
abcSup: I am going to push a new commit.

libr/debug/dreg.c
abcSup: Ok

libr/debug/dreg.c
```diff
@@ -128,9 +129,32 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char
...
+		head = r_list_new ();
+		for (i = 0; i < R_REG_TYPE_LAST; i++) {
+			RList *reglist = r_reg_get_list (dbg->reg, i);
+			if (!reglist || r_list_length (reglist) == 0) {
+				r_list_free (reglist);
+				continue;
+			}
+			RList *tmp = r_list_clone (reglist);
+			if (!tmp) {
+				r_list_free (reglist);
```
ret2libc, line 142: why this? I think you said that `r_reg_get_list` does not return a newly allocated list, so freeing it will cause issues I think.

libr/debug/dreg.c
```diff
@@ -82,7 +82,8 @@ R_API int r_debug_reg_sync(RDebug *dbg, int type, int write) {
 	return true;
 }
 
-R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char *use_color) {
+R_API bool r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char *use_color) {
+	size_t i;
```
ret2libc, line 86: move this inside the `if`.

libr/debug/dreg.c
abcSup: you are right. To output all register types, I decided to clone and join all the lists together, so I can keep the outputting part of the code the same. If we change `r_reg_get_list` API to always return a newly allocated list, we can make `r_reg_get_list` to directly return a RList of the R_REG_TYPE_LAST type instead of doing it in `r_debug_reg_list`, which can be reused elsewhere.

libr/debug/dreg.c
abcSup: Thanks for the reminder. I am sorry that I resolved those requests before actually committing the changes as I was keeping track of them here.

libr/debug/dreg.c
ret2libc: I don't have a strong opinion. You can either change the API so that it directly returns `R_REG_TYPE_ALL` or just do the chaining here as you are already doing :) I'm personally fine with both approaches for now.

abcSup: Ok, great. I think I will do it this way for now.

libr/debug/dreg.c
abcSup: I just tried and it caused segfault when exiting r2. I assume that is the case because of what I mentioned earlier that the `RList` is still referenced by the `RRegSet`.
```
RList *tmp = r_reg_get_list (dbg->reg, i);
if (!tmp || !r_list_length (tmp)) {
    r_list_free (reglist);
   continue
}
```
Let me know what you think and I hope I am looking at the right thing as well. Thank you!

abcSup:
I just found out that `RReg->allregs` already contains a list of all registers `RRegItem`, so there is no need to clone and join them again. I will update the PR soon to make use of that instead.

libr/debug/dreg.c
abcSup: I am going to resolve this as I am using `RReg->allregs` instead

Issue #9126, opened 2018-01-03
Bad LEA disassembled opcodes


kartone:
### Work environment

| Questions                                | Answers
|------------------------------------------|--------------------
| OS/arch/bits                             | Linux kali 4.14.0-kali1-amd64 
| File format of the file you reverse      | ELF
| Architecture/bits of the file.           | x86/64
| r2 -v full output                        | radare2 2.3.0-git 16905 @ linux-x86-64 git.2.2.0-96-ge854df5c0

### Expected behavior

    8d 45 fc             	lea    eax,[ebp-0x4]

### Actual behavior

    8d 45 fc              lea eax, ebp - 4 

### Steps to reproduce the behavior 

https://asciinema.org/a/X2XKN4pvQjZub0ys9uzhJ2pLY

### Compiled binaries

Bin: crackme0x02 from [IOLI crackmes serie](https://dustri.org/b/files/IOLI-crackme.tar.gz)

radare:
Thats expected. It has been discussed recently to make the syntax more readable, we can bring back the old syntax for lea, but as long as lea is not accessing memory the [] is confusing. We should probably relay this for an asm filter or asm.syntax.

neeocf:
I agree that [] would be confusing, but the "[r|e]bp - X" is not replaced by "local_Xh" for lea as it used to be (and it was very helpful).

radare:
Oh! Thats a bug then

kartone:
@neeocf That is exactly what I meant! Completely agree with you.

radare:
FIxed by rollbacking the [] change. also it was inconsistent with the assembler and syntax of [] is memory reference which is base+delta, so other syntax would be wrong for a generic assembler

PR #16980, opened 2020-05-31
x86 pseudo for LEA is wrong

### LINKED ISSUE #16777: x86 pseudo for LEA is wrong
### Work environment

| OS/arch/bits (mandatory)                             | Windows 10/Arch linux amd64
| Architecture/bits of the file (mandatory)            | x86 (32-bit)
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.5.0-git 24573 @ linux-x86-64 git.4.4.0-138-g64e6df500 commit: 64e6df5004b2718526761503a48437e82be3471e build: 2020-05-06__18:01:55

### Expected behavior

```
> e asm.pseudo = false
0x0804843c      8d45f8         lea eax, [var_8h]
```

```
> e asm.pseudo = true
0x0804843c      8d45f8         eax = var_8h
```

### Actual behavior

```
> e asm.pseudo = true
0x0804843c      8d45f8         eax = [var_8h]
```

The brackets are unneeded for pseudo, and therefore this syntax is inconsistent with arithmetic operators (which outputs like `eax = eax * edx`).

It is possible to tell the difference between a `mov` and `lea` in pseudo as the latter has no `dword` prefix, but this is still very confusing.

It was supposed to be fixed by #6618 but then reverted for #9126 which left the pseudocode part broken.

Bin: crackme0x02 from [IOLI crackmes series](https://dustri.org/b/files/IOLI-crackme.tar.gz) (from #9126)

zawwwu:

**Detailed description**
PR alters pseudo code for lea operator to not include brackets. Variable resolution is preserved.

Example:
```
> e asm.pseudo = false
0x0804843c      8d45f8         lea eax, [var_8h]
```
```
> e asm.pseudo = true
0x0804843c      8d45f8         eax = var_8h
```

**Test plan**

Use e.g. crackme0x02 from IOLI crackmes series. Run ```e asm.pseudo = true``` and confirm pseudo code for lea looks as described in example.

**Closing issues**

closes #16777

### DIFF
--- a/libr/parse/p/parse_x86_pseudo.c
+++ b/libr/parse/p/parse_x86_pseudo.c
@@ -261,6 +261,10 @@ static int parse(RParse *p, const char *data, char *str) {
 
 		replace (nw, wa, str);
 
+	} else if (strstr (w0, "lea")) {
+		r_str_replace_char (w2, '[', 0);
+		r_str_replace_char (w2, ']', 0);
+		replace (nw, wa, str);
 	} else if ((strstr (w1, "ax") || strstr (w1, "ah") || strstr (w1, "al")) && !p->retleave_asm) {
 		if (!(p->retleave_asm = (char *) malloc (sz))) {
 			return false;
@@ -303,9 +307,9 @@ static void parse_localvar (RParse *p, char *newstr, size_t newstr_len, const ch
 			r_strbuf_setf (sb, " + %s", ireg);
 		}
 		if (p->localvar_only) {
-			snprintf (newstr, newstr_len - 1, "[%s%s]", var, r_strbuf_get (sb));
+			snprintf (newstr, newstr_len - 1, "%s%s", var, r_strbuf_get (sb));
 		} else {
-			snprintf (newstr, newstr_len - 1, "[%s%s %c %s]", reg, r_strbuf_get (sb), sign, var);
+			snprintf (newstr, newstr_len - 1, "%s%s %c %s", reg, r_strbuf_get (sb), sign, var);
 		}
 	}
 	r_strbuf_free (sb);
@@ -327,9 +331,9 @@ static inline void mk_reg_str(const char *regname, int delta, bool sign, bool at
 			r_strbuf_setf (sb, " + %s", ireg);
 		}
 		if (delta < 10) {
-			snprintf (dest, len - 1, "[%s%s %c %d]", regname, r_strbuf_get (sb), sign ? '+':'-', delta);
+			snprintf (dest, len - 1, "%s%s %c %d", regname, r_strbuf_get (sb), sign ? '+':'-', delta);
 		} else {
-			snprintf (dest, len - 1, "[%s%s %c 0x%x]", regname, r_strbuf_get (sb), sign ? '+':'-', delta);
+			snprintf (dest, len - 1, "%s%s %c 0x%x", regname, r_strbuf_get (sb), sign ? '+':'-', delta);
 		}
 	}
 	r_strbuf_free (sb);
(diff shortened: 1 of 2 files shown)


libr/parse/p/parse_x86_pseudo.c
```diff
@@ -485,6 +485,13 @@ static bool varsub (RParse *p, RAnalFunction *f, ut64 addr, int oplen, char *dat
 		}
 	}
 
+    // Strip brackets from lea pseudo
+	if (r_regex_match ("[a-z]{3} = \\[[a-z0-9_]+\\]$", "ne", tstr) &&
```
trufae, line 489: There's no need to use a regex to achieve this, actually this was the actual behaviour.

See 17b17d91c2926e2900bfa4b34c83b88aa69bcc4b

it just a matter of reverting 06311f718b16176e8c1bbcf4ca36da23286c60a0

trufae: Also, please add a test for this

zawwwu: > There's no need to use a regex to achieve this, actually this was the actual behaviour.
> 
> See [17b17d9](https://github.com/radareorg/radare2/commit/17b17d91c2926e2900bfa4b34c83b88aa69bcc4b)
> 
> it just a matter of reverting [06311f7](https://github.com/radareorg/radare2/commit/06311f718b16176e8c1bbcf4ca36da23286c60a0)

Not quite sure what you mean, because simply reverting [06311f7](https://github.com/radareorg/radare2/commit/06311f718b16176e8c1bbcf4ca36da23286c60a0) indeed strips brackets, but at the same time breaks variable name resolution as mentioned in [#9126](https://github.com/radareorg/radare2/issues/9126). As I was tinkering with this issue I found out that substitution of combination `[reg +/- offset]` by variable name depends on brackets being there, so if brackets are removed **before** above mentioned combination is substituted by name of the variable, then (as far as I understood) part of varsub function that iterates over ebp/esp variables doesn't consider this a variable and thus substitution doesn't occur. If you're referring to changing disassembly completely that is
```
> e asm.pseudo = false
0x0804843c      8d45f8         lea eax, var_8h
```
```
> e asm.pseudo = true
0x0804843c      8d45f8         eax = var_8h

```
Then for sure we can try this (as I understood from comments it was tried and it worked), but in my opinion it breaks expectation of how lea source operand should look.

ret2libc: I don't think we should change how `lea` looks in the disassembly, because it may be confusing, as in general it appears with `[..]`

zawwwu:
I tinkered with that variable resolution loop I mentioned a bit more and came up with a much better solution. It works on a first glance, but I will write tests for it and update PR if it works for edge cases as well.

XVilka:
By the way, this one would benefit from a small test.

test/db/anal/x86_64
```diff
@@ -3474,3 +3474,85 @@ EOF
...
+|           ; arg int64_t arg_8h @ rbp+0x8
+|           ; arg int64_t arg_bh @ rbp+0xb
+|           ; var int64_t var_4h @ rsp+0x4
+|           ; arg int64_t arg_ch @ rsp+0xc
+|           0x00000000      54             push rsp
+|           0x00000001      4889e5         mov rbp, rsp
+|           0x00000004      b402           mov ah, 2
+|           0x00000006      66b83333       mov ax, 0x3333
+|           0x0000000a      b855554444     mov eax, 0x44445555
+|           0x0000000f      48b8bebafeca.  movabs rax, 0xdeadbeefcafebabe
```
ret2libc, line 3502: Could you add the `movabs` op in the `ops` array in `replace` function? It should be just a matter of adding `{ "movabs", "# = #", {1, 2}},` to the array. I know it's not strictly related to your change, but at least we can create a correct test ;)

ret2libc:
@trufae do you have changes to request? If not, I think this is good to merge.

PR #16983, opened 2020-05-31
Fix type matching with loaded PDB ##anal

GustavoLCR:

**Detailed description**

This fixes getting a flag in the wrong flagspace

Use r_flag_get_by_spaces() instead of manually comparing the flagspace name

**Test plan**

Tests pass

### DIFF
--- a/libr/core/anal_tp.c
+++ b/libr/core/anal_tp.c
@@ -394,8 +394,8 @@ static void type_match(RCore *core, char *fcn_name, ut64 addr, ut64 baddr, const
 							sdb_fmt ("%s%s%s", type, r_str_endswith (type, "*") ? "" : " ", name));
 					cmt_set = true;
 					if ((op->ptr && op->ptr != UT64_MAX) && !strcmp (name, "format")) {
-						RFlagItem *f = r_flag_get_i (core->flags, op->ptr);
-						if (f && f->space && !strcmp (f->space->name, R_FLAGS_FS_STRINGS)) {
+						RFlagItem *f = r_flag_get_by_spaces (core->flags, op->ptr, R_FLAGS_FS_STRINGS, NULL);
+						if (f) {
 							char formatstr[0x200];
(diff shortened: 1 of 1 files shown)


ret2libc:
> This PR contains A LOT of changes in different places not just in PDB, like improving the HTTP apis, so it will be good if next time we refactor/squash the commits accordingly. and have multiple separate commits with grouped changes

BTW, this is not an easy thing to do and it is something that should be done by those who submit PRs. It is, in some cases, a lot of work to do and something that, for some people (especially if not very familiar with git) can become annoying and frustrating, when it involves rebases, fixes due to comments in PRs, etc.

Do we really want to enforce that? I am not sure personally. Anyway, we have never logically split commits, because usually we just do squash+merge.

trufae:
If the PR gets big enough and the changes can be easily grouped i think
it's worth doing it, adjusting the commit message to explain all the
changes in 1 line and add the ## tag should be the minimum requirement, and
this is task of the one merging the PR.

I understand requesting this to the contributors can be a pain, and this
won't apply to all the PRs, but it makes reading git log much easier that
way.

ret2libc:
> If the PR gets big enough and the changes can be easily grouped i think it's worth doing it, adjusting the commit message to explain all the changes in 1 line and add the ## tag should be the minimum requirement, and this is task of the one merging the PR. I understand requesting this to the contributors can be a pain, and this won't apply to all the PRs, but it makes reading git log much easier that way.

Using a good 1-line message and adding the ##tag can be done by the one merging the PR. Splitting the commits, no, as it is too much work usually. It would require a lot of extra work, because you need to carefully rebasing, re-ordering commits, fixup, etc. IMO it shall either be done by the one who writes the PR or not done at all. In the Linux Kernel, at least, they require people who submit changes to carefully arranging the commits.

That said, IMHO we are not at the point of requesting this kind of dedication from contributors, as it would make contributing much harder and painful.

trufae:
👍

Issue #6971, opened 2017-03-10
META - RADIFF2


Maijin:
# Related

* DOC: TBD
* BLOG: TBD
* Tests : https://github.com/radare/radare2-regressions/blob/master/t.tools/radiff2/radiff2
* .c involved:
    * https://github.com/radare/radare2/blob/master/libr/core/cmd_cmp.c
    * https://github.com/radare/radare2/blob/master/binr/radiff2/radiff2.c

# Fixes

* [ ] radiff2 -g color issues https://github.com/radare/radare2/issues/3614
* [ ] radiff2 malware diffing bad results https://github.com/radare/radare2/issues/5362

# Enhancement

* [ ] Do not autoname when radiff2 -C
* [ ] radiff2 -jC - JSON output for the graphdiff
* [ ] radiff2 -x binary1 binary2 should print diff in readable hex output https://github.com/radare/radare2/issues/561
* [ ] Obey colourscheme in diff graphs https://github.com/radare/radare2/issues/2141
* [ ] Implement -r mode for radiff2 -C
* [ ] radiff2 -g should output in ascii art too https://github.com/radare/radare2/issues/3508
* [ ] Colours radiff2 graphdiff output https://github.com/radare/radare2/issues/6021
* [ ] radiff2 -u should support buffers of different size ; it is currently chopping the buffers to R_MIN(lenA, lenB)

# Compatibility with other softwares

* [ ] Output for BinNavi https://github.com/radare/radare2/issues/3873
* [ ] Support all analysis info required for Diaphora https://github.com/radare/radare2/issues/2742

XVilka:
See also https://github.com/radare/radare2/issues/9128

trufae:
take ideas to diff assembly https://github.com/simonlindholm/asm-differ

PR #16986, opened 2020-06-01
radiff2: drop MODE_DIST

MaskRay:
MODE_DIST (-s) is the same as MODE_DIST_LENVENSTEIN (-ss) but much slower.
Drop it. Make MODE_DIST_MYERS (-sss, faster than MODE_DIST_LENVENSTEIN) take its place (-s).

The original comment was incorrect (-s is not Eugene W. Myer's
algorithm). It is correct now.

 

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/main/radiff2.c
+++ b/libr/main/radiff2.c
@@ -7,9 +7,8 @@ enum {
 	MODE_DIFF,
 	MODE_DIFF_STRS,
 	MODE_DIFF_IMPORTS,
-	MODE_DIST,
 	MODE_DIST_MYERS,
-	MODE_DIST_LEVENSTEIN,
+	MODE_DIST_LEVENSHTEIN,
 	MODE_CODE,
 	MODE_GRAPH,
 	MODE_COLS,
@@ -17,15 +16,15 @@ enum {
 };
 
 enum {
-        GRAPH_DEFAULT_MODE,
-        GRAPH_SDB_MODE,
-        GRAPH_JSON_MODE,
-        GRAPH_JSON_DIS_MODE,
-        GRAPH_TINY_MODE,
-        GRAPH_INTERACTIVE_MODE,
-        GRAPH_DOT_MODE,
-        GRAPH_STAR_MODE,
-        GRAPH_GML_MODE
+	GRAPH_DEFAULT_MODE,
+	GRAPH_SDB_MODE,
+	GRAPH_JSON_MODE,
+	GRAPH_JSON_DIS_MODE,
+	GRAPH_TINY_MODE,
+	GRAPH_INTERACTIVE_MODE,
+	GRAPH_DOT_MODE,
+	GRAPH_STAR_MODE,
+	GRAPH_GML_MODE
 };
 
 typedef struct {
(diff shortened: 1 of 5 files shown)


kuqadk3:
Yeah the old code was wrong, the commit looks good.

libr/main/radiff2.c
```diff
@@ -1236,7 +1233,6 @@ R_API int r_main_radiff2(int argc, const char **argv) {
 		}
 		r_diff_free (d);
 		break;
-	case MODE_DIST:
```
ret2libc, line 1239: Since you are removing this, could you remove the
```
} else {
   d->type = 0;
}
```
few linew below? I don't think it can be ever executed.

XVilka:
We probably need tests for those. @MaskRay or @kuqadk3

libr/main/radiff2.c
```diff
@@ -1006,12 +1005,10 @@ R_API int r_main_radiff2(int argc, const char **argv) {
 		case 'h':
 			return show_help (1);
 		case 's':
-			if (mode == MODE_DIST) {
+			if (mode == MODE_DIST_MYERS) {
```
trufae, line 1008: this sounds wrong to me, why dist-myers should be enabling the dist-levenstein mode? and use myers as alternative? the namings are confusing.

trufae: imho those MODEs should be defined in r_util  and use the same identifiers in `r_diff_buffers_distance` . Also your change is making `r_diff_buffers_distance_original` to be dead code. so there's no way to compare those myers/levenstein algorithms with the plain simple implementation (despite being slower)

MaskRay: 'l' is buggy (r_diff_buffers_distance_levenshtein) is buggy. I left a FIXME.

XVilka:
@MaskRay what about the test?

MaskRay:
> @MaskRay what about the test?

`test/unit/test_diff.c` has unit tests. 'l' is commented out.

XVilka:
I think it's fine to merge as is then. I opened a new issue for fixing Levenstein distance calculation https://github.com/radareorg/radare2/issues/17237

See also more on https://github.com/radareorg/radare2/issues/6971

libr/main/radiff2.c
```diff
@@ -1246,12 +1246,12 @@ R_API int r_main_radiff2(int argc, const char **argv) {
 			RDiff *d = r_diff_new ();
 			if (d) {
 				d->verbose = verbose;
-				if (mode == MODE_DIST_LEVENSTEIN) {
-					d->type = 'l';
-				} else if (mode == MODE_DIST_MYERS) {
+				if (mode == MODE_DIST_MYERS) {
```
trufae, line 1249: if levenstein is catched in https://github.com/radareorg/radare2/pull/16986/files#diff-da3b7af0a797b3cf25e710584fb60571R1244 it should be throwing a warning or something instead of using the default method instead.

Also i would like to undestand and maybe get documnetd about the differences of all the distance diffing algorithms and why an user may pick one or another, this should be clear in the manpage too.

trufae: also the title is drop MODE_DIST: and mode_dist is still there.

trufae:
ping

trufae:
Also, please update the help message and the manpage are not updated.

trufae:
ping

trufae (approved):

trufae:
Thank you!!!

Issue #16766, opened 2020-05-04
Update tree-sitter as soon as a new release is done


ret2libc:
tree-sitter got support for old compilers (e.g. the one in CentOS 6/RHEL 6), but there was no upstream release yet. As soon as a release will be done (or anyway before new radare2 release) update it.

ret2libc:
See https://github.com/tree-sitter/tree-sitter/releases/tag/0.16.7 . With this, radare2 should compile on CentOS 6 and centos6-CI should be green.

PR #16991, opened 2020-06-01
Update tree-sitter to 0.16.8

ret2libc:
This should allow to compile radare2 again on very old compilers like
the one shipped in CentOS6.

**Test plan**

All should be green, including centos6 CI

**Closing issues**

Closes https://github.com/radareorg/radare2/issues/16766


XVilka:
1. It's red.
2. Why not update to 0.16.8 then? https://github.com/tree-sitter/tree-sitter/releases/tag/0.16.8

ret2libc:
Yeah, 0.16.8 just came out 16h ago, after I created this PR

XVilka:
Could you please update to the latest then? ;)

PR #16993, opened 2020-06-01
rabin2 could not detect go in macho binaries

DharmaOfCode:
- [] I've added tests that prove my fix is effective or that my feature works (if possible)
- [] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

rabin2 only detects go language for ELF binaries. This update now detects golang when using the `-I` flag with rabin2 against a go binary built for macho.

...

**Test plan**

1. Build a go binary with any of the following: `GOOS=darwin GOARCH=amd64 go build`
2. Run `rabin2 -I <binary> | grep lang`
3. The result should be `lang     go` (previously it returned `lang     c`

...

**Closing issues**

N/A

...

### DIFF
--- a/libr/bin/blang.c
+++ b/libr/bin/blang.c
@@ -42,10 +42,7 @@ static bool check_swift(RBinSymbol *sym) {
 }
 
 static bool check_golang(RBinSymbol *sym) {
-	if (!strncmp (sym->name, "go.", 3)) {
-		return true;
-	}
-	return false;
+	return !strncmp (sym->name, "go.", 3);
 }
 
 static inline bool is_cxx_symbol (const char *name) {
@@ -114,11 +111,9 @@ R_API int r_bin_load_languages(RBinFile *binfile) {
 				return R_BIN_NM_RUST;
 			}
 		}
-		if (!cantbe.go) {
-			if (check_golang (sym)) {
-				info->lang = "go";
-				return R_BIN_NM_GO;
-			}
+		if (check_golang (sym)) {
+			info->lang = "go";
+			return R_BIN_NM_GO;
 		}
 		if (!cantbe.swift) {
 			bool hasswift = false;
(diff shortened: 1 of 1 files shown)


libr/bin/blang.c
```diff
@@ -106,6 +114,12 @@ R_API int r_bin_load_languages(RBinFile *binfile) {
 				return R_BIN_NM_RUST;
 			}
 		}
+		if (!cantbe.go) {
```
ret2libc, line 117: I think you don't need this, because it is never set to any value.

libr/bin/blang.c
```diff
@@ -40,6 +41,13 @@ static bool check_swift(RBinSymbol *sym) {
 	return false;
 }
 
+static bool check_golang(RBinSymbol *sym) {
+	if (!strncmp (sym->name, "go.", 3)) {
```
ret2libc, line 45: just do:
```suggestion
	return !strncmp (sym->name, "go.", 3);
```

libr/bin/blang.c
nVisium-Alex-Useche: this makes sense. Looks like we could eventually remove a couple other similar checks in the same file. For instance, `cantbe.rust` is never set either. I will update this particular case for now. Thank you

libr/bin/blang.c
nVisium-Alex-Useche: I will update this accordingly. Eventually, we could do the same type of update for other similar checks in this file as well.

libr/bin/blang.c
```diff
@@ -9,6 +9,7 @@ typedef struct {
 	bool swift;
 	bool cxx;
 	bool msvc;
+	bool go;
```
ret2libc, line 12: remove this

libr/bin/blang.c
ret2libc: Yeah, absolutely. If you are up for it, feel free to improve those other cases as well ;)

nVisium-Alex-Useche: for sure! I will submit as a separate PR later this week

libr/bin/blang.c
nVisium-Alex-Useche: removed in the latest commit

PR #16829, opened 2020-05-15
WIP: Fix ELF relocs parsing overflow

radare:
**TODO**

* [x] Add type `R_AARCH64_JUMP_SL` for those relocs
* [x] Fix the relocs listing regression that affects 4-5 tests

**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

Loading the librsjni_androix.so bin from the testsuite into r2 makes it crash.

This PR fixes the parsing of the relocs by fixing the overflow and getting  back the relocs to work

**Test plan**

I have added the binary that makes r2 crash into the testsuite and just loading with r2 makes it crash. But also after the fix, the output of the relocs is empty. while it shuold be what readelf shows:

```
$ readelf -r librsjni_androix.so

Relocation section '.rela.plt' at offset 0xb600 contains 17 entries:
  Offset          Info           Type           Sym. Value    Sym. Name + Addend
00000000e210  000100000402 R_AARCH64_JUMP_SL 0000000000000000 __cxa_finalize@LIBC + 0
00000000e218  000600000402 R_AARCH64_JUMP_SL 0000000000000000 __stack_chk_fail@LIBC + 0
00000000e220  000a00000402 R_AARCH64_JUMP_SL 0000000000000000 dlopen@LIBC + 0
00000000e228  001100000402 R_AARCH64_JUMP_SL 000000000000736c _Z11loadSymbolsPvR13di + 0
00000000e230  000900000402 R_AARCH64_JUMP_SL 0000000000000000 dlerror@LIBC + 0
00000000e238  000500000402 R_AARCH64_JUMP_SL 0000000000000000 __android_log_print + 0
00000000e240  000800000402 R_AARCH64_JUMP_SL 0000000000000000 dlclose@LIBC + 0
00000000e248  001200000402 R_AARCH64_JUMP_SL 000000000000851c _Z14loadIOSuppSymsPvR8 + 0
00000000e250  000d00000402 R_AARCH64_JUMP_SL 0000000000000000 malloc@LIBC + 0
[...]

PR #17004, opened 2020-06-03
Fix elf reloc crash

ret2libc:

**Detailed description**

This PR supersedes https://github.com/radareorg/radare2/pull/16829 . It should fix a crash while loading the librsjni_androix.so bin from the testsuite.

It does so by:
- fixing wrong computation of the offset in function `get_next_not_analysed_offset`
- using virtual address instead of offset files in `read_reloc`. All relocations values are expressed in vaddr, so using offsets+-baddr was wrong
- passing `num_relocs` around so that we can use it both as a defensive measure in case `get_num_relocs_approx` returns wrong value and also as a way to limit the number of relocation entries to read in case the section info are fake and too large.

It also adds support for AARCH64 reloc entries.

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -2659,19 +2659,25 @@ static size_t populate_relocs_record_from_dynamic(ELFOBJ *bin, RBinElfReloc *rel
 	size_t offset;
 	size_t size = get_size_rel_mode (bin->dyn_info.dt_pltrel);
 
-	for (offset = 0; offset < bin->dyn_info.dt_pltrelsz && pos < num_relocs; offset += size, pos++) {
-		read_reloc (bin, relocs + pos, bin->dyn_info.dt_pltrel, bin->dyn_info.dt_jmprel + offset);
-		fix_rva_and_offset_exec_file (bin, relocs + pos);
+	for (offset = 0; offset < bin->dyn_info.dt_pltrelsz && pos < num_relocs; offset += size) {
+		if (read_reloc (bin, relocs + pos, bin->dyn_info.dt_pltrel, bin->dyn_info.dt_jmprel + offset)) {
+			fix_rva_and_offset_exec_file (bin, relocs + pos);
+			pos++;
+		}
 	}
 
-	for (offset = 0; offset < bin->dyn_info.dt_relasz && pos < num_relocs; offset += bin->dyn_info.dt_relaent, pos++) {
-		read_reloc (bin, relocs + pos, DT_RELA, bin->dyn_info.dt_rela + offset);
-		fix_rva_and_offset_exec_file (bin, relocs + pos);
+	for (offset = 0; offset < bin->dyn_info.dt_relasz && pos < num_relocs; offset += bin->dyn_info.dt_relaent) {
+		if (read_reloc (bin, relocs + pos, DT_RELA, bin->dyn_info.dt_rela + offset)) {
+			fix_rva_and_offset_exec_file (bin, relocs + pos);
+			pos++;
+		}
 	}
 
-	for (offset = 0; offset < bin->dyn_info.dt_relsz && pos < num_relocs; offset += bin->dyn_info.dt_relent, pos++) {
-		read_reloc (bin, relocs + pos, DT_REL, bin->dyn_info.dt_rel + offset);
-		fix_rva_and_offset_exec_file (bin, relocs + pos);
+	for (offset = 0; offset < bin->dyn_info.dt_relsz && pos < num_relocs; offset += bin->dyn_info.dt_relent) {
+		if (read_reloc (bin, relocs + pos, DT_REL, bin->dyn_info.dt_rel + offset)) {
+			fix_rva_and_offset_exec_file (bin, relocs + pos);
+			pos++;
+		}
 	}
 
 	return pos;
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -2562,7 +2562,12 @@ static void fix_rva_and_offset(ELFOBJ *bin, RBinElfReloc *r, size_t pos) {
 	}
 }
 
-static bool read_reloc(ELFOBJ *bin, RBinElfReloc *r, Elf_(Xword) rel_mode, ut64 offset) {
+static bool read_reloc(ELFOBJ *bin, RBinElfReloc *r, Elf_(Xword) rel_mode, ut64 vaddr) {
+	ut64 offset = Elf_(r_bin_elf_v2p_new) (bin, vaddr);
+	if (offset == UT64_MAX) {
+		return false;
+	}
+
 	size_t size_struct = get_size_rel_mode (rel_mode);
 
 	ut8 buf[sizeof (Elf_(Rela))] = { 0 };
@@ -2657,20 +2662,20 @@ static size_t populate_relocs_record_from_dynamic(ELFOBJ *bin, RBinElfReloc *rel
 
 	for (offset = 0; offset < bin->dyn_info.dt_pltrelsz; offset += size) {
 		read_reloc (bin, relocs + pos, bin->dyn_info.dt_pltrel,
-				bin->dyn_info.dt_jmprel + offset - bin->baddr);
+				bin->dyn_info.dt_jmprel + offset);
 		fix_rva_and_offset_exec_file (bin, relocs + pos);
 		pos++;
 		++i;
 	}
 
 	for (offset = 0; offset < bin->dyn_info.dt_relasz; offset += bin->dyn_info.dt_relaent) {
-		read_reloc (bin, relocs + pos, DT_RELA, bin->dyn_info.dt_rela + offset - bin->baddr);
+		read_reloc (bin, relocs + pos, DT_RELA, bin->dyn_info.dt_rela + offset);
 		fix_rva_and_offset_exec_file (bin, relocs + pos);
 		pos++;
 	}
 
 	for (offset = 0; offset < bin->dyn_info.dt_relsz; offset += bin->dyn_info.dt_relent) {
-		read_reloc (bin, relocs + pos, DT_REL, bin->dyn_info.dt_rel + offset - bin->baddr);
+		read_reloc (bin, relocs + pos, DT_REL, bin->dyn_info.dt_rel + offset);
 		fix_rva_and_offset_exec_file (bin, relocs + pos);
 		pos++;
 	}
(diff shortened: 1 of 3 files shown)


ret2libc:
cc @08A , if you can have a look please it would be appreciated! Thanks

libr/bin/format/elf/elf.c
```diff
@@ -2650,57 +2655,56 @@ static size_t get_num_relocs_approx(ELFOBJ *bin) {
...
-	size_t i = 0;
 	size_t size = get_size_rel_mode (bin->dyn_info.dt_pltrel);
 
-	for (offset = 0; offset < bin->dyn_info.dt_pltrelsz; offset += size) {
-		read_reloc (bin, relocs + pos, bin->dyn_info.dt_pltrel,
-				bin->dyn_info.dt_jmprel + offset - bin->baddr);
-		fix_rva_and_offset_exec_file (bin, relocs + pos);
-		pos++;
-		++i;
+	for (offset = 0; offset < bin->dyn_info.dt_pltrelsz && pos < num_relocs; offset += size) {
```
08A, line 2662: Why did you had pos < num_relocs.
num_relocs is compute with the information inside the dynamic section.
So if `offset < bin->dyn_info.dt_pltrelsz` is false => `pos < num_relocs` is false

libr/bin/format/elf/elf.c
```diff
@@ -2650,57 +2655,56 @@ static size_t get_num_relocs_approx(ELFOBJ *bin) {
...
+			fix_rva_and_offset_exec_file (bin, relocs + pos);
+			pos++;
+		}
 	}
 
 	return pos;
 }
 
-static size_t get_next_not_analysed_offset(ELFOBJ *bin, size_t section_offset, size_t offset, size_t base_addr) {
+static size_t get_next_not_analysed_offset(ELFOBJ *bin, size_t section_vaddr, size_t offset) {
```
08A, line 2686: Indeed the old computation was wrong

libr/bin/format/elf/elf.c
```diff
@@ -2717,11 +2721,11 @@ static size_t populate_relocs_record_from_section(ELFOBJ *bin, RBinElfReloc *rel
 
 		size = get_size_rel_mode (rel_mode);
 
-		for (j = get_next_not_analysed_offset (bin, bin->g_sections[i].offset, 0, bin->baddr);
-			j < bin->g_sections[i].size;
-			j = get_next_not_analysed_offset (bin, bin->g_sections[i].offset, j + size, bin->baddr)) {
+		for (j = get_next_not_analysed_offset (bin, bin->g_sections[i].rva, 0);
+			j < bin->g_sections[i].size && pos < num_relocs;
```
08A, line 2725: Again i don't understand the use of num_relocs

libr/bin/format/elf/elf.c
```diff
@@ -2650,57 +2655,56 @@ static size_t get_num_relocs_approx(ELFOBJ *bin) {
...
+			pos++;
+		}
 	}
 
-	for (offset = 0; offset < bin->dyn_info.dt_relsz; offset += bin->dyn_info.dt_relent) {
-		read_reloc (bin, relocs + pos, DT_REL, bin->dyn_info.dt_rel + offset - bin->baddr);
-		fix_rva_and_offset_exec_file (bin, relocs + pos);
-		pos++;
+	for (offset = 0; offset < bin->dyn_info.dt_relsz && pos < num_relocs; offset += bin->dyn_info.dt_relent) {
+		if (read_reloc (bin, relocs + pos, DT_REL, bin->dyn_info.dt_rel + offset)) {
```
08A, line 2677: if it failed you should break => the only reason of read_reloc failing is a wrong read, all next read will be invalid.

08A:
> cc @08A , if you can have a look please it would be appreciated! Thanks
done

libr/bin/format/elf/elf.c
ret2libc: That's true in theory. As explained above, I added it for 2 reasons:
1) as a defensive measure to avoid overflowing the array in case of something else wrong (weak justification, but it is what happened with the wrong computation)
2) because in case you analyze a particular binary, the relocation section may be soo big that you can't allocate the required memory for all the possible relocations, so we may just read the dynamic relocation entries (see https://github.com/radareorg/radare2/pull/17004/files#diff-7c0443a3dd0679f5b11dfd70dd990d93R2745)

trufae:
this is related to https://github.com/radareorg/radare2/pull/16829 and i have more changes in local that didn't even cared to push or finish them, ill just find some motivation to update mine after rebasing everything when all your stuff is merged

ret2libc:
> this is related to #16829 and i have more changes in local that didn't even cared to push or finish them, ill just find some motivation to update mine after rebasing everything when all your stuff is merged

Yes, it is just the minimal set of changes to properly fix the crash (+ small AARCH64 relocation support, copied from your PR, to make the test work).

kazarmy:
Does this "minimal set of changes" include the `num_relocs` stuff?

ret2libc:
> Does this "minimal set of changes" include the `num_relocs` stuff?

Yes it does, see https://github.com/radareorg/radare2/pull/17004/files#diff-7c0443a3dd0679f5b11dfd70dd990d93R2755 .

kazarmy:
That very-innocent-looking librsjni_androix.so causes r2 to enter the following `if` block? https://github.com/radareorg/radare2/pull/17004/files#diff-7c0443a3dd0679f5b11dfd70dd990d93R2744-R2749

ret2libc:
> That very-innocent-looking librsjni_androix.so causes r2 to enter the following `if` block? https://github.com/radareorg/radare2/pull/17004/files#diff-7c0443a3dd0679f5b11dfd70dd990d93R2744-R2749

No, that's a "protection" I added instead of just checking for NULL and exiting.  So, we could just do the usual `if (!relocs) return NULL`, but I think we can be a bit more resilient to errors here.

Indeed `get_num_relocs_approx` computes both the number of reloc entries according to the dynamic segment and according to the sections. If I want to create an ELF that somehow hides information from r2, I could try to create a binary with the right relocation entries (that is, the right entries in the dynamic segment), but with a very very large relocation section. By doing so, `get_num_relocs_approx` may return a very high number that could not be allocated on some systems. When that happens, we can try to allocate just enough space for the reloc entries identified in the dynamic segment (which are the only one actually used by the loader), so instead of having no information, we have some information.

Do you think I need to add a better comment in the code explaining all this?

libr/bin/format/elf/elf.c
```diff
@@ -2736,14 +2740,23 @@ static size_t populate_relocs_record_from_section(ELFOBJ *bin, RBinElfReloc *rel
...
 	size_t i = 0;
 	size_t num_relocs = get_num_relocs_approx (bin);
-	RBinElfReloc *relocs = calloc (num_relocs + 1, sizeof (RBinElfReloc));
+	RBinElfReloc *relocs = R_NEWS0 (RBinElfReloc, num_relocs + 1);
+	if (!relocs) {
+		// In case we can't allocate enough memory for all the claimed
+		// relocation entries, try to parse only the ones specified in
+		// the dynamic segment.
+		num_relocs = get_num_relocs_dynamic (bin);
+		relocs = R_NEWS0 (RBinElfReloc, num_relocs + 1);
```
kazarmy, line 2749: Please warn the user that you're truncating the relocs table and provide also the reason for truncation.

kazarmy:
> Do you think I need to add a better comment in the code explaining all this?

No but the user needs to be informed.

libr/bin/format/elf/elf.c
ret2libc: Good point!

kazarmy:
> - fixing wrong computation of the offset in function get_next_not_analysed_offset
> - using virtual address instead of offset files in read_reloc. All relocations values are expressed in vaddr, so using offsets+-baddr was wrong

librsjni_androix.so looks even more innocent under `readelf`:

program headers

Are you sure you're not getting a bad `baddr`? What am I missing?

ret2libc:
> Are you sure you're not getting a bad `baddr`? What am I missing?

The fact is that we should not use `baddr` to convert between virtual and physical (or viceversa) because that is not the right approach. We should rely on the segment mappings, which is what `r_bin_elf_v2p_new` does and that is the function I am using in this PR, instead of just using `baddr`.

What is exactly that is not clear?

kazarmy:
> What is exactly that is not clear?

That is very clear. I'm just asking whether the cause of the crash was due to an incorrect `baddr`, which if so should probably be fixed in a different pr.

ret2libc:
Ah no, the cause of the crash was fixed in https://github.com/radareorg/radare2/pull/17004/commits/49ae28d5d63607eb09a65cc7dc99b560ceb6c7cc and it was due to the wrong computation in `get_next_not_analysed_offset`. That function was supposed to give an offset, starting from the beginning of the section, where to read a relocation entry. However, in the cases https://github.com/radareorg/radare2/pull/17004/commits/49ae28d5d63607eb09a65cc7dc99b560ceb6c7cc#diff-7c0443a3dd0679f5b11dfd70dd990d93L2685 , https://github.com/radareorg/radare2/pull/17004/commits/49ae28d5d63607eb09a65cc7dc99b560ceb6c7cc#diff-7c0443a3dd0679f5b11dfd70dd990d93L2690 and https://github.com/radareorg/radare2/pull/17004/commits/49ae28d5d63607eb09a65cc7dc99b560ceb6c7cc#diff-7c0443a3dd0679f5b11dfd70dd990d93L2695 , it wrongly computed the offset starting from the beginning of the relocations as specified in the dynamic *segment*, and not from the start of the section. 

In some cases, like with this test library, this resulted in a overflow, because the offsets returned by `get_next_not_analysed_offset` would go something like: 0x100, 0x110, 0x120, 0x130, 0x140 and then, when the flow satisfies one of those `if` conditions, the offset would go back to some other values, like 0x120, 0x130, 0x140, etc. That way you end up with adding more and more relocation entries to the `relocs` array, resulting in an overflow.

kazarmy:
Ok if you say so. Not familiar enough with the code or the elf dynamic section to decide independently.

08A:
@ret2libc Your solution with num_relocs can help detecting an invalid dynamic section.

see the binary => bins/elf/analysis/dynamic-poffset

libr/bin/format/elf/elf.c
kazarmy: I lack a test bin for this so dropping the requirement.

PR #17006, opened 2020-06-04
Fix runaway scrolling in Visual mode after mashing down movement key on Windows Terminal

kazarmy:

**Detailed description**

This pr (basically 9ac00d93f98a2d4ada66cd1ce4bf9fdd9391ddbd) prevents runaway scrolling in Visual mode after mashing down hjkl or an arrow key on Windows Terminal, by clearing the stdin buffer. It is #12761, #12764 and #12823 ported over to Windows.

**Test plan**

Tested with keyboard and mouse wheel with default `scr.vtmode` and the following:

- Conhost Windows 10 1903
- Windows Terminal 1.0.1401.0
- Alacritty 0.5.0-dev (e6475c6) -- apparently currently no mouse support even with `e scr.vtmode = 2`
- mintty 3.1.6

**Closing issues**

...

### DIFF
--- a/libr/core/visual.c
+++ b/libr/core/visual.c
@@ -4265,34 +4265,34 @@ dodo:
 #ifdef __WINDOWS__
 			if (I->vtmode == 2 && !I->term_xterm) {
 #endif
-			if (IS_PRINTABLE (ch) || ch == '\t' || ch == '\n') {
+				if (IS_PRINTABLE (ch) || ch == '\t' || ch == '\n') {
 #ifdef __WINDOWS__
-				while (r_cons_readchar_timeout (1) != -1) ;
+					while (r_cons_readchar_timeout (1) != -1) ;
 #else
-				tcflush (STDIN_FILENO, TCIFLUSH);
+					tcflush (STDIN_FILENO, TCIFLUSH);
 #endif
-			} else if (ch == 0x1b) {
-				char chrs[2];
-				int chrs_read = 1;
-				chrs[0] = r_cons_readchar ();
-				if (chrs[0] == '[') {
-					chrs[1] = r_cons_readchar ();
-					chrs_read++;
-					if (chrs[1] >= 'A' && chrs[1] <= 'D') { // arrow keys
+				} else if (ch == 0x1b) {
+					char chrs[2];
+					int chrs_read = 1;
+					chrs[0] = r_cons_readchar ();
+					if (chrs[0] == '[') {
+						chrs[1] = r_cons_readchar ();
+						chrs_read++;
+						if (chrs[1] >= 'A' && chrs[1] <= 'D') { // arrow keys
 #ifdef __WINDOWS__
-						while (r_cons_readchar_timeout (1) != -1) ;
+							while (r_cons_readchar_timeout (1) != -1) ;
 #else
-						tcflush (STDIN_FILENO, TCIFLUSH);
-						// Following seems to fix an issue where scrolling slows
-						// down to a crawl after some time mashing the up and down
-						// arrow keys
-						r_cons_set_raw (false);
-						r_cons_set_raw (true);
+							tcflush (STDIN_FILENO, TCIFLUSH);
+							// Following seems to fix an issue where scrolling slows
+							// down to a crawl after some time mashing the up and down
+							// arrow keys
+							r_cons_set_raw (false);
+							r_cons_set_raw (true);
 #endif
+						}
 					}
+					(void)r_cons_readpush (chrs, chrs_read);
 				}
-				(void)r_cons_readpush (chrs, chrs_read);
-			}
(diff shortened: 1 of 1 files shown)


libr/core/visual.c
```diff
@@ -4262,26 +4262,38 @@ R_API int r_core_visual(RCore *core, const char *input) {
...
+						chrs_read++;
+						if (chrs[1] >= 'A' && chrs[1] <= 'D') { // arrow keys
+#ifdef __WINDOWS__
+							while (r_cons_readchar_timeout (1) != -1) ;
+#else
+							tcflush (STDIN_FILENO, TCIFLUSH);
+							// Following seems to fix an issue where scrolling slows
+							// down to a crawl after some time mashing the up and down
+							// arrow keys
+							r_cons_set_raw (false);
```
ret2libc, line 4289: Don't you need also on windows?

libr/core/visual.c
```diff
@@ -4262,26 +4262,38 @@ R_API int r_core_visual(RCore *core, const char *input) {
...
-						// Following seems to fix an issue where scrolling slows
-						// down to a crawl after some time mashing the up and down
-						// arrow keys
-						r_cons_set_raw (false);
-						r_cons_set_raw (true);
+#ifdef __WINDOWS__
+			if (I->vtmode == 2 && !I->term_xterm) {
+#endif
+				if (IS_PRINTABLE (ch) || ch == '\t' || ch == '\n') {
+#ifdef __WINDOWS__
```
ret2libc, line 4269: Not a big fan of #ifdef all around... In the kernel they usually try to ifdef entire functions, so it would become something like:
```C
#ifdef __WINDOWS__
static void flush_stdin(void) {
   while (r_cons_readchar_timeout (1) != -1) ;
}
#else
static void flush_stdin(void) {
   tcflush (STDIN_FILENO, TCIFLUSH);
}
#endif
```

then you could just use that function both here and few lines below.

libr/core/visual.c
```diff
@@ -4262,26 +4262,38 @@ R_API int r_core_visual(RCore *core, const char *input) {
...
-					chrs_read++;
-					if (chrs[1] >= 'A' && chrs[1] <= 'D') { // arrow keys
-						tcflush (STDIN_FILENO, TCIFLUSH);
-						// Following seems to fix an issue where scrolling slows
-						// down to a crawl after some time mashing the up and down
-						// arrow keys
-						r_cons_set_raw (false);
-						r_cons_set_raw (true);
+#ifdef __WINDOWS__
+			if (I->vtmode == 2 && !I->term_xterm) {
```
ret2libc, line 4266: What is vtmode == 2? What does it represent?

Also here I would try to extract this in a separate function, so the flow is easier to follow without all the ifdefs.
Something like:
```C
#ifdef __WINDOWS__
static bool is_vt_compatible(void) {
   return I->vtmode == 2 && !I->term_xterm
}
#else
static bool is_vt_compatible(void) {
   return true;
}
```

and then here you can just do `if (is_vt_compatible ()) {` and it works both on linux and on windows. The name `is_vt_compatible` is probably not appropriate, as I'm not sure what `I->vtmode == 2` does.

libr/core/visual.c
kazarmy:
```c
r_cons_set_raw (false);
r_cons_set_raw (true);
```
is basically equivalent to power cycling and thus didn't actually solve the cause of the issue. Issue might be due to `tcflush()` which doesn't exist on Windows, and so far issue has not been seen on Windows.

libr/core/visual.c
kazarmy: > What is vtmode == 2? What does it represent?

I'm not sure either (ask @GustavoLCR) but what I know is that it enables mouse support in Visual panels (`V!`).

GustavoLCR: > > What is vtmode == 2? What does it represent?
> 
> I'm not sure either (ask @GustavoLCR) but what I know is that it enables mouse support in Visual panels (`V!`).

vtmode == 2 means the terminal sends vt escape sequences as input

kazarmy: This raises the question of what happens when vtmode != 2, but https://docs.microsoft.com/en-us/windows/console/setconsolemode and https://docs.microsoft.com/en-us/windows/console/reading-input-buffer-events seems to explain sufficiently.

kazarmy: > Also here I would try to extract this in a separate function, so the flow is easier to follow without all the ifdefs.

Ok 19f1e733ccbf5b652c7e5dc1737969c6507ca472

libr/core/visual.c
kazarmy: Ok 43cbde42cb1544ac872bf3e59ae442b0676c1e7b

PR #17015, opened 2020-06-08
Try clang-format with Clang 10

XVilka:
With Meson it is simple:
```
meson build
cd build
ninja clang-format
```
I used Clang 10.
Some parts are off, but we can fence them with `// clang-format off`, so far it looks good and can handle most of the situations. If something we don't like or there are problems - we can report the bug in Clang/LLVM itself.

More options are available at https://clang.llvm.org/docs/ClangFormatStyleOptions.html

For a couple things I opened Clang->Formatter component bugs:
- https://bugs.llvm.org/show_bug.cgi?id=46240
- https://bugs.llvm.org/show_bug.cgi?id=46241

Also Meson bugs:
- https://github.com/mesonbuild/meson/issues/7268
- https://github.com/mesonbuild/meson/issues/7271


trufae:
I would be happy to have an automated solution for indenting code as a pre-commit hook (dunno if github allows pre-merge actions), which are the current differences/missing options to support the r2 style with clang-format? Also, clang-format have different behaviour and features depending on the version. May we ask contributors to use a specific version of it? Or do you propose to have automatic indentation and stop asking contributors to take care of what they type.

Also, there are some directories (shlr) where we probably dont want to run that.

XVilka:
1. We will require particular version or above
2. We will run the formatter as a CI action, so contributors will not need to install clang-format if they do not want.

ret2libc:
I think we can easily exclude some directories. Also, even having a requirement on some particular version is better than having no automatic solution at all IMHO.

thestr4ng3r:
Currently building clang-format from scratch to check if we can add support for our `function (stuff)` and `function(stuff)` duality. I think if we have a convenient automation for PRs we don't have to ask users at all anymore to run any formatting tool.

thestr4ng3r:
Important function apparently is `TokenAnnotator::spaceRequiredBetween()`. Will continue later.

XVilka:
@MaskRay is the LLVM/Clang expert, maybe can give an advice.

kazarmy:
> Currently building clang-format from scratch to check if we can add support for our function (stuff) and function(stuff) duality.

If this is too much trouble, it appears the regexes (e.g. for `r_sandbox_opendir`) are:

Definition: `git grep "^[^[:blank:]].*r_sandbox_opendir *(.*{"`
Declaration: `git grep "^[^[:blank:]].*r_sandbox_opendir *(.*;"`
Both: `git grep "^[^[:blank:]].*r_sandbox_opendir *(.*[{;]"`

Ugly yes but the ugliness can be hidden within a script. This might even be faster than something like `git grep "r_sandbox_opendir("` because the first char requirement eliminates a lot of lines.

trufae:
@kazarmy try git grep, is faster than grep.
also @XVilka @ret2libc i think requesting a specific version of clang to all the contributors it is a very complex thing to achieve and will be worst for contributors than having to type following the guidelines

kazarmy:
> Declaration: git grep "^[^[:blank:]].*r_sandbox_opendir *(.*;"
Both: git grep "^[^[:blank:]].*r_sandbox_opendir *(.*[{;]"

If there are pesky `#define` lines, use instead:

Declaration: `git grep "^[^[:blank:]#].*r_sandbox_opendir *(.*;"`
Both: `git grep "^[^[:blank:]#].*r_sandbox_opendir *(.*[{;]"`

Sorry WIP.

> @kazarmy try git grep, is faster than grep.

Yeah git grep is very fast.

XVilka:
Ripgrep is faster than grep or git grep.

https://github.com/BurntSushi/ripgrep

wargio:
@XVilka that's a lot of changes.

condret:
we should also use clang-tidy, clang-format does not add new tokens, so it won't touch single line if-statements that have no braces

ret2libc:
@XVilka what to do with this? I would love to have an automatic way to check/enforce code style, but it looks some people are not willing to let some bits of the current style go.

Shall we close this PR?

ret2libc:
Closing this. There is nothing to review, nothing to see. We shall just choose whether we want it or not. Apparently not.

Issue #16248, opened 2020-03-18
dmh is SLOWER


x0urc3:
| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | Ubuntu x86
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | x86/64
| r2 -v full output, **not truncated** (mandatory)     |  radare2 4.2.0-git 23640 @ linux-x86-64 git.4.1.1-153-gd0a915458 commit: d0a915458abaa429926840c9fa41e2cd1fedfb1b build: 2020-01-22__11:37:15

### Expected behavior
`dmh` produce result in reasonable amout of time

### Actual behavior
`dmh` is slow even though there's only one chunk in the heap

### Steps to reproduce the behavior 
```
$ r2 -d test/bins/elf/simple_malloc_x86_64 
Process with PID 15946 started...
= attach 15946 15946
bin.baddr 0x55b6ab962000
Using 0x55b6ab962000
asm.bits 64
Warning: r_bin_file_hash: file exceeds bin.hashlimit
 -- This shell has been seized by the Internet's Police.
[0x7f6041d2d100]> dcu main
Continue until 0x55b6ab96270a using 1 bpsize
hit breakpoint at: 55b6ab96270a
[0x55b6ab96270a]> ?t dmh
10.411503

  Malloc chunk @ 0x55b6ad0e8290 [size: 0x11c10][allocated]
  Top chunk @ 0x55b6ad0f9ea0 - [brk_start: 0x55b6ad0e8000, brk_end: 0x55b6ad109000]

[0x55b6ab96270a]> 
```

radare:
didnt tried bu ti  guess those 10s are spent parsing the libc binary and loading all the symbol information and metadata into r2

x0urc3:
> didnt tried but i guess those 10s are spent parsing the libc library and loading all the symbol information and metadata into r2

Your right. Brute force search to find main_arena. Any other method to find main_arena address in memory?

https://github.com/radareorg/radare2/blob/14977a6828b47ee5a70e6c41c6d5dce873b78d48/libr/core/linux_heap_glibc.c#L343-L353

meowmeowxw:
pwndbg checks main_arena in the symbols of the libc: https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/heap/ptmalloc.py#L73
[...]

Issue #10398, opened 2018-06-19
dmh not shows correctly freed chunk


Goyo2600:
binary with trhee mallocs: A, B, C
dmh shows chunks correctly
free(B)
dmh shows C chunk freed instead of B

https://asciinema.org/a/Pr165M3Sb0f8LEb5ztd8StufZ

Tested on Ubuntu x64
binary x64
radare2 2.7.0-git 18461 @ linux-x86-64 git.2.6.0-283-g8638b83
commit: 8638b8395fbda29c9441100122de618994c29855 build: 2018-06-18__19:48:58

### Expected behavior
It should show [free] on B chunk
### Actual behavior
It shows [free] on chunk C

soez:
The function had to detect the double free the distance of a node, if it was greater no. Now if it detects it, although it has a triple while.. :/ the other functions will be changed if this likes pancake (more will be optimized)

https://github.com/radare/radare2/pull/10419

Goyo2600:
Hello soez, I've seen the modified code in linux_heap_glibc.c, but there are something wrong.
As you can see in this simple binary, there are three mallocs, A, B, C (three large continuous chunks), and a free(B).
After free(B), r2  dmh and dmhg commands doesn't show chunk size modified neither [free] text in freed chunk.

[prueba2.gz](https://github.com/radare/radare2/files/2118870/prueba2.gz)

Goyo2600:
Hi, now is perfect without loading libc with LD_PRELOAD, but using LD_PRELOAD r2 crash after dmh 

Warning: glibc library with symbol main_arena could not be found. Is libc6-dbg installed?
*** Error in `r2': free(): invalid pointer: 0x00007f5cec737c68 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc-64.so.6(+0x777e5)[0x7f5cec3eb7e5]
/lib/x86_64-linux-gnu/libc-64.so.6(+0x7fe0a)[0x7f5cec3f3e0a]
/lib/x86_64-linux-gnu/libc-64.so.6(cfree+0x4c)[0x7f5cec3f798c]
/usr/lib/libr_util.so(r_list_pop_head+0x9c)[0x7f5ce83653ff]
/usr/lib/libr_core.so(r_core_task_schedule+0x7a)[0x7f5cec0f2caf]
/usr/lib/libr_core.so(r_core_task_continue+0x1d)[0x7f5cec0f2fc8]
[...]

PR #17018, opened 2020-06-08
Heap glibc resolve main_arena using symbol

meowmeowxw:

**Detailed description**

With this feature the heap commands for the glibc run faster, since they resolve the main_arena using the symbol from the glibc (if the glibc is compiled with symbols, otherwise it uses the bruteforce method which is slower).

I measured the time of the commands `dmh[ft]`:

1. Using the symbol: ~0.07 seconds
2. Without the symbol: ~1.10 seconds

Before I had an hard time to use the commands `dmht` and `dmhf` in `V!` mode because it took more than a second to refresh the view at every step, now it works fine.

This is an improvement but it doesn't solve the problem completely.
In fact on my machine if I try the command `dmh` before `dmh[tf]` is very slow, while the other way round is super fast.

Related to "#16248"

### DIFF
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -25,6 +25,10 @@
 static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname) {
 	RBin *bin = core->bin;
 	GHT vaddr = GHT_MAX;
+	if (!bin) {
+		return vaddr;
+	}
+
 	RListIter *iter;
 	RBinSymbol *s;
 	const char *bin_file = bin->file;
@@ -32,7 +36,6 @@ static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname)
 	ut64 laddr = r_config_get_i (core->config, "bin.laddr");
 	int fd = r_io_fd_get_current (core->io);
 	int rawstr = bin->rawstr;
-
 	RBinOptions opt;
 	r_bin_options_init (&opt, -1, 0, 0, false);
 	r_bin_open (bin, path, &opt);
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -961,7 +961,7 @@ static void GH(print_heap_segment)(RCore *core, MallocState *main_arena,
 	if (m_arena == m_state) {
 		GH(get_brks) (core, &brk_start, &brk_end);
 		if (tcache) {
-			// tcache_initial_brk = ((brk_start >> 12) << 12) + GH(HDR_SZ);
+			//tcache_initial_brk = ((brk_start >> 12) << 12) + GH(HDR_SZ);
 			GHT fc_offset = GH(tcache_chunk_size) (core, brk_start);
 			initial_brk = ((brk_start >> 12) << 12) + fc_offset;
 		} else {
@@ -1049,6 +1049,7 @@ static void GH(print_heap_segment)(RCore *core, MallocState *main_arena,
 	}
 
 	const char *comma = "";
+	bool first_time = true;
 	while (next_chunk && next_chunk >= brk_start && next_chunk < main_arena->GH(top)) {
 		if (size_tmp < min_size || next_chunk + size_tmp > main_arena->GH(top)) {
 			const char *status = "corrupted";
@@ -1131,6 +1132,10 @@ static void GH(print_heap_segment)(RCore *core, MallocState *main_arena,
 		}
 
 		if (tcache) {
+			if (!first_time) {
+				tcache_initial_brk = ((brk_start >> 12) << 12) + GH(HDR_SZ);
+			}
+			first_time = false;
 			GH(RHeapTcache) *tcache_heap = R_NEW0 (GH(RHeapTcache));
 			if (!tcache_heap) {
 				r_cons_canvas_free (can);
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -25,10 +25,6 @@
 static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname) {
 	RBin *bin = core->bin;
 	GHT vaddr = GHT_MAX;
-	if (!bin) {
-		return vaddr;
-	}
-
 	RListIter *iter;
 	RBinSymbol *s;
 	const char *bin_file = bin->file;
@@ -36,6 +32,7 @@ static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname)
 	ut64 laddr = r_config_get_i (core->config, "bin.laddr");
 	int fd = r_io_fd_get_current (core->io);
 	int rawstr = bin->rawstr;
+
 	RBinOptions opt;
 	r_bin_options_init (&opt, -1, 0, 0, false);
 	r_bin_open (bin, path, &opt);
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -26,10 +26,10 @@ static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname)
 	GHT vaddr = GHT_MAX;
 	RListIter *iter;
 	RBinSymbol *s;
-	// RBin *bin = r_bin_new ();
-	// RIO *io = r_io_new ();
-	// r_io_bind (io, &bin->iob);
-	RBin *bin = core->bin;
+	RBin *bin = r_bin_new ();
+	RIO *io = r_io_new ();
+	r_io_bind (io, &bin->iob);
+	// RBin *bin = core->bin;
 	
 	if (!bin) {
 		return vaddr;
@@ -49,8 +49,8 @@ static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname)
 		}
 	}
 
-	// r_bin_free (bin);
-	// r_io_free (io);
+	r_bin_free (bin);
+	r_io_free (io);
 	return vaddr;
 }
 
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -350,7 +350,7 @@ static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
 		RDebugMap *map;
 		r_debug_map_sync (core->dbg);
 		r_list_foreach (core->dbg->maps, iter, map) {
-			if (strstr (map->name, "/libc-") && map->perm == 4 && main_arena_sym == GHT_MAX) {
+			if (strstr (map->name, "/libc-") && map->perm == R_PERM_R && main_arena_sym == GHT_MAX) {
 				libc_path = map->name;
 				libc_addr = map->addr;
 				if (!libc_path) {
@@ -365,7 +365,7 @@ static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
 				}
 				free (path);
 			}
-			if (strstr (map->name, "/libc-") && map->perm == 6) {
+			if (strstr (map->name, "/libc-") && map->perm == R_PERM_RW) {
 				libc_addr_sta = map->addr;
 				libc_addr_end = map->addr_end;
 				break;
@@ -961,7 +961,7 @@ static void GH(print_heap_segment)(RCore *core, MallocState *main_arena,
 	if (m_arena == m_state) {
 		GH(get_brks) (core, &brk_start, &brk_end);
 		if (tcache) {
-			tcache_initial_brk = ((brk_start >> 12) << 12) + GH(HDR_SZ);
+			// tcache_initial_brk = ((brk_start >> 12) << 12) + GH(HDR_SZ);
 			GHT fc_offset = GH(tcache_chunk_size) (core, brk_start);
 			initial_brk = ((brk_start >> 12) << 12) + fc_offset;
 		} else {
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -21,8 +21,18 @@
 #define GHT_MAX UT64_MAX
 #endif
 
-
-static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname) {
+/**
+ * \brief Find the address of a given symbol
+ * \param core RCore Pointer to the r2's core
+ * \param path Pointer to the binary path in which to look for the symbol
+ * \param sym_name Pointer to the symbol's name to search for
+ * \return address
+ *
+ * Used to find the address of a given symbol inside a binary
+ *
+ * TODO: Stop using deprecated functions like r_bin_cur
+ */
+static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *sym_name) {
 	GHT vaddr = GHT_MAX;
 	RBin *bin = core->bin;
 	RBinFile *current_bf = r_bin_cur (bin);
@@ -31,11 +41,14 @@ static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname)
 
 	RBinOptions opt;
 	r_bin_options_init (&opt, -1, 0, 0, false);
-	r_bin_open (bin, path, &opt);
-	RList *syms = r_bin_get_symbols (bin);
+	bool res = r_bin_open (bin, path, &opt);
+	if (!res) {
+		return vaddr;
+	}
 
+	RList *syms = r_bin_get_symbols (bin);
 	r_list_foreach (syms, iter, s) {
-		if (!strcmp (s->name, symname)) {
+		if (!strcmp (s->name, sym_name)) {
 			vaddr = s->vaddr;
 			break;
 		}
@@ -348,6 +361,7 @@ static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
 		RDebugMap *map;
 		r_debug_map_sync (core->dbg);
 		r_list_foreach (core->dbg->maps, iter, map) {
+			/* Try to find the main arena address using the glibc's symbols. */
 			if (strstr (map->name, "/libc-") && map->perm == R_PERM_R && main_arena_sym == GHT_MAX) {
 				libc_path = map->name;
 				libc_addr = map->addr;
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -23,19 +23,15 @@
 
 
 static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname) {
+	RBin *bin = core->bin;
 	GHT vaddr = GHT_MAX;
 	RListIter *iter;
 	RBinSymbol *s;
-	const char *bin_file = core->bin->file;
+	const char *bin_file = bin->file;
 	ut64 baddr = r_config_get_i (core->config, "bin.baddr");
 	ut64 laddr = r_config_get_i (core->config, "bin.laddr");
 	int fd = r_io_fd_get_current (core->io);
-	int rawstr = core->bin->rawstr;
-	RBin *bin = core->bin;
-	
-	if (!bin) {
-		return vaddr;
-	}
+	int rawstr = bin->rawstr;
 
 	RBinOptions opt;
 	r_bin_options_init (&opt, -1, 0, 0, false);
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -25,9 +25,9 @@
 static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname) {
 	GHT vaddr = GHT_MAX;
 	RBin *bin = core->bin;
+	RBinFile *current_bf = r_bin_cur (bin);
 	RListIter *iter;
 	RBinSymbol *s;
-	ut64 baddr = r_config_get_i (core->config, "bin.baddr");
 
 	RBinOptions opt;
 	r_bin_options_init (&opt, -1, 0, 0, false);
@@ -42,14 +42,8 @@ static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname)
 	}
 
 	RBinFile *libc_bf = r_bin_cur (bin);
-	RCoreFile *cf = r_core_file_cur (core);
-	if (cf) {
-		RBinFile *bf = r_bin_file_find_by_fd (core->bin, cf->fd);
-		if (bf) {
-			r_bin_reload (core->bin, bf->id, baddr);
-		}
-	}
 	r_bin_file_delete (bin, libc_bf->id);
+	r_bin_file_set_cur_binfile (bin, current_bf);
 	return vaddr;
 }
 
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -26,10 +26,12 @@ static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname)
 	GHT vaddr = GHT_MAX;
 	RListIter *iter;
 	RBinSymbol *s;
-	RBin *bin = r_bin_new ();
-	RIO *io = r_io_new ();
-	r_io_bind (io, &bin->iob);
-	// RBin *bin = core->bin;
+	const char *bin_file = core->bin->file;
+	ut64 baddr = r_config_get_i (core->config, "bin.baddr");
+	ut64 laddr = r_config_get_i (core->config, "bin.laddr");
+	int fd = r_io_fd_get_current (core->io);
+	int rawstr = core->bin->rawstr;
+	RBin *bin = core->bin;
 	
 	if (!bin) {
 		return vaddr;
@@ -49,8 +51,8 @@ static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname)
 		}
 	}
 
-	r_bin_free (bin);
-	r_io_free (io);
+	r_bin_options_init (&opt, fd, baddr, laddr, rawstr);
+	r_bin_open (core->bin, bin_file, &opt);
 	return vaddr;
 }
 
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -31,7 +31,7 @@ static GHT GH(get_va_symbol)(const char *path, const char *symname) {
 	r_io_bind (io, &bin->iob);
 
 	if (!bin) {
-		return -1;
+		return 0;
 	}
 
 	RBinOptions opt;
@@ -39,7 +39,7 @@ static GHT GH(get_va_symbol)(const char *path, const char *symname) {
 	r_bin_open (bin, path, &opt);
 	syms = r_bin_get_symbols (bin);
 	if (!syms) {
-		return -1;
+		return 0;
 	}
 	r_list_foreach (syms, iter, s) {
 		if (!strcmp (s->name, symname)) {
@@ -360,7 +360,7 @@ static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
 				char *path = r_str_newf ("%s", libc_path);
 				if (r_file_exists (path)) {
 					GHT vaddr = GH (get_va_symbol) (path, main_arena_str);
-					if (libc_addr != GHT_MAX && vaddr != -1) {
+					if (libc_addr != GHT_MAX && vaddr != 0) {
 						main_arena_sym = libc_addr + vaddr;
 						free (path);
 					}
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -3,7 +3,6 @@
 #ifndef INCLUDE_HEAP_GLIBC_C
 #define INCLUDE_HEAP_GLIBC_C
 #define HEAP32 1
-bool main_arena_resolved = 0;
 #include "linux_heap_glibc.c"
 #undef HEAP32
 #endif
@@ -23,13 +22,11 @@ bool main_arena_resolved = 0;
 #endif
 
 
-static GHT GH(get_va_symbol)(const char *path, const char *symname) {
+static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname) {
 	GHT vaddr = GHT_MAX;
 	RListIter *iter;
 	RBinSymbol *s;
-	RBin *bin = r_bin_new ();
-	RIO *io = r_io_new ();
-	r_io_bind (io, &bin->iob);
+	RBin *bin = core->bin;
 
 	if (!bin) {
 		return vaddr;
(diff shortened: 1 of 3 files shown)


libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,37 @@
...
+	r_io_bind (io, &bin->iob);
+
+	if (!bin) {
+		return 0;
+	}
+
+	RBinOptions opt;
+	r_bin_options_init (&opt, -1, 0, 0, false);
+	r_bin_open (bin, path, &opt);
+	syms = r_bin_get_symbols (bin);
```
trufae, line 40: define and assign syms in one line

libr/core/linux_heap_glibc.c
```diff
@@ -307,15 +338,34 @@ static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
...
 		RDebugMap *map;
 		r_debug_map_sync (core->dbg);
 		r_list_foreach (core->dbg->maps, iter, map) {
+			if (strstr (map->name, "/libc-") && map->perm == 4 && !main_arena_sym) {
+				libc_path = map->name;
+				libc_addr = map->addr;
+				if (!libc_path) {
+					break;
+				}
+				char *path = r_str_newf ("%s", libc_path);
```
trufae, line 360: no need for strnewf() . just strdup()

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,37 @@
...
+	RList *syms = NULL;
+	GHT vaddr = 0LL;
+	r_io_bind (io, &bin->iob);
+
+	if (!bin) {
+		return 0;
+	}
+
+	RBinOptions opt;
+	r_bin_options_init (&opt, -1, 0, 0, false);
```
trufae, line 38: uhm, using va=0 to add the aslr baddr, sgtm!

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,37 @@
...
+	RListIter *iter;
+	RBinSymbol *s;
+	RBin *bin = r_bin_new ();
+	RIO *io = r_io_new ();
+	RList *syms = NULL;
+	GHT vaddr = 0LL;
+	r_io_bind (io, &bin->iob);
+
+	if (!bin) {
+		return 0;
```
trufae, line 34: i think GHT_MAX should be returned in case of error. i know address 0 shouldnt be valid place. but -1 should never be a valid address to place any symbol. and it makes the code more consistent to read

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,37 @@
 #define GHT_MAX UT64_MAX
 #endif
 
+static GHT GH(get_va_symbol)(const char *path, const char *symname) {
+	RListIter *iter;
+	RBinSymbol *s;
+	RBin *bin = r_bin_new ();
+	RIO *io = r_io_new ();
+	RList *syms = NULL;
+	GHT vaddr = 0LL;
```
trufae, line 30: use GHT_MAX

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,37 @@
 #define GHT_MAX UT64_MAX
 #endif
 
+static GHT GH(get_va_symbol)(const char *path, const char *symname) {
+	RListIter *iter;
+	RBinSymbol *s;
+	RBin *bin = r_bin_new ();
```
trufae, line 27: what about reusing the instances from core->rbin and core->io instead of creating new ones which make all the plugins load twice?

meowmeowxw: https://github.com/radareorg/radare2/blob/a934a39236b3604e0e6ecebbee2b3b4d9c16bf06/libr/include/r_core.h#L241

I didn't found any rbin inside core, I tried to iterate the binfiles inside bin, but I could only found the path to the program and not the path to the shared library (glibc). Probably I'm missing something.

radare:
Yes the rbin instance is there but the glibc is not loaded. Im suggesting to reuse the instance and open the glibc get the symbol and unload the rbinobj

meowmeowxw:
I cached the `main_arena` address if it was found using the symbol from the glibc.
Now all the commands run immediately after the first time.

XVilka:
You might also check https://github.com/radareorg/radare2/issues/10398 while you are on it.

libr/core/linux_heap_glibc.c
```diff
@@ -3,6 +3,7 @@
 #ifndef INCLUDE_HEAP_GLIBC_C
 #define INCLUDE_HEAP_GLIBC_C
 #define HEAP32 1
+bool main_arena_resolved = 0;
```
trufae, line 6: use false instead of 0.

Also, don't use global variables, hold this variable inside RDebug for example

libr/core/linux_heap_glibc.c
trufae: just reuse the core->rbin, and load/close the bin parsing, you dont need to instantiate io because it is already instantiated in core.bin, that will make it faster because loading allt he io and rbin plugins takes a while

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +22,39 @@
...
+
+static GHT GH(get_va_symbol)(const char *path, const char *symname) {
+	GHT vaddr = GHT_MAX;
+	RListIter *iter;
+	RBinSymbol *s;
+	RBin *bin = r_bin_new ();
+	RIO *io = r_io_new ();
+	r_io_bind (io, &bin->iob);
+
+	if (!bin) {
```
trufae, line 34: you probably want to move this check above, right after creating the instance

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +22,39 @@
...
+		return vaddr;
+	}
+
+	RBinOptions opt;
+	r_bin_options_init (&opt, -1, 0, 0, false);
+	r_bin_open (bin, path, &opt);
+	RList *syms = r_bin_get_symbols (bin);
+
+	if (!syms) {
```
trufae, line 43: be positive. use if (syms) { ... } so you can reuse the same return statement

libr/core/linux_heap_glibc.c
```diff
@@ -303,19 +337,43 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
...
+		return true;
+	}
+
+	const char *main_arena_str = "main_arena";
 	GHT brk_start = GHT_MAX, brk_end = GHT_MAX;
 	GHT libc_addr_sta = GHT_MAX, libc_addr_end = 0;
 	GHT addr_srch = GHT_MAX, heap_sz = GHT_MAX;
+	const char *libc_path = NULL;
+	GHT libc_addr = GHT_MAX;
+	GHT main_arena_sym = 0;
```
trufae, line 355: use GHT_MAX by default. that will never be a valid address

libr/core/linux_heap_glibc.c
```diff
@@ -303,19 +337,43 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
...
+				libc_addr = map->addr;
+				if (!libc_path) {
+					break;
+				}
+				char *path = strdup(libc_path);
+				if (r_file_exists (path)) {
+					GHT vaddr = GH (get_va_symbol) (path, main_arena_str);
+					if (libc_addr != GHT_MAX && vaddr != GHT_MAX) {
+						main_arena_sym = libc_addr + vaddr;
+						free (path);
```
trufae, line 373: uhm this free seems to be placed in the wrong line. it will be leaking its value in some situations

libr/core/linux_heap_glibc.c
```diff
@@ -303,19 +337,43 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
...
 		RDebugMap *map;
 		r_debug_map_sync (core->dbg);
 		r_list_foreach (core->dbg->maps, iter, map) {
+			if (strstr (map->name, "/libc-") && map->perm == 4 && !main_arena_sym) {
+				libc_path = map->name;
+				libc_addr = map->addr;
+				if (!libc_path) {
+					break;
+				}
+				char *path = strdup(libc_path);
```
trufae, line 368: missing space before (

libr/core/linux_heap_glibc.c
```diff
@@ -303,19 +337,43 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
 }
 
 static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
+
 	if (!core || !core->dbg || !core->dbg->maps) {
```
trufae, line 341: use r_return_val_if_fail instead

trufae: not having dbg.maps is fine, so just chk for core and coredbg with the returniffail and keep the dbg->maps with a simple if

libr/core/linux_heap_glibc.c
```diff
@@ -355,6 +413,16 @@ static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
 	if (!ta) {
 		return false;
 	}
+
+	if (main_arena_sym) {
+		GH (update_main_arena)
+		(core, main_arena_sym, ta);
+		*m_arena = main_arena_sym;
+		main_arena_resolved = 1;
```
trufae, line 421: use `true` instead of 1 and dont use globals

libr/core/linux_heap_glibc.c
```diff
@@ -355,6 +413,16 @@ static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
 	if (!ta) {
 		return false;
 	}
+
+	if (main_arena_sym) {
+		GH (update_main_arena)
```
trufae, line 418: uhm. the line below are the arguments for the function call in here. merge those 2 lines

libr/core/linux_heap_glibc.c
meowmeowxw: I use dbg->maps to find the libc, so It's required to check that is not empty, or not ?

meowmeowxw:
I think that the latest commit solves #10398 and #16248. 
Someone forgot to assign the address of `tcache_initial_brk` and `dmh` tried to print the tcache from the adress GHT_MAX, so the output of `dmh` was both slow and incorrect.

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,39 @@
 #define GHT_MAX UT64_MAX
 #endif
 
+
+static GHT GH(get_va_symbol)(RCore *core, const char *path, const char *symname) {
+	GHT vaddr = GHT_MAX;
+	RListIter *iter;
+	RBinSymbol *s;
+	// RBin *bin = core->bin;
+	RBin *bin = r_bin_new();
```
trufae, line 30: dont create new instances of rbin, just use core->rbin

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,39 @@
...
+ 	RIO *io = r_io_new ();
```
trufae, line 31: unnecessary. its already instantiated by rcore

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,39 @@
...
+
+	if (syms) {
+		r_list_foreach (syms, iter, s) {
+			if (!strcmp (s->name, symname)) {
+				vaddr = s->vaddr;
+				break;
+			}
+		}
+	}
+
```
trufae, line 51: call r_bin_close() and dont free rbin or io

libr/core/linux_heap_glibc.c
```diff
@@ -303,19 +336,45 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
...
+					break;
+				}
+				char *path = strdup (libc_path);
+				if (r_file_exists (path)) {
+					GHT vaddr = GH (get_va_symbol) (core, path, main_arena_str);
+					if (libc_addr != GHT_MAX && vaddr != GHT_MAX) {
+						main_arena_sym = libc_addr + vaddr;
+					}
+				}
+				if (path) {
```
trufae, line 374: unnecessary if, free(NULL) is fine

libr/core/linux_heap_glibc.c
```diff
@@ -303,19 +336,45 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
...
 		r_debug_map_sync (core->dbg);
 		r_list_foreach (core->dbg->maps, iter, map) {
+			if (strstr (map->name, "/libc-") && map->perm == 4 && main_arena_sym == GHT_MAX) {
+				libc_path = map->name;
+				libc_addr = map->addr;
+				if (!libc_path) {
+					break;
+				}
+				char *path = strdup (libc_path);
+				if (r_file_exists (path)) {
```
trufae, line 368: if (path && ...

libr/core/linux_heap_glibc.c
```diff
@@ -303,19 +336,45 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
...
 
 	if (r_config_get_i (core->config, "cfg.debug")) {
 		RListIter *iter;
 		RDebugMap *map;
 		r_debug_map_sync (core->dbg);
 		r_list_foreach (core->dbg->maps, iter, map) {
+			if (strstr (map->name, "/libc-") && map->perm == 4 && main_arena_sym == GHT_MAX) {
+				libc_path = map->name;
+				libc_addr = map->addr;
+				if (!libc_path) {
```
trufae, line 364: if (!map->name) break..

make this conditional before the assignment

libr/core/linux_heap_glibc.c
```diff
@@ -355,13 +414,23 @@ static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
 	if (!ta) {
 		return false;
 	}
+
+	if (main_arena_sym) {
+		GH (update_main_arena) (core, main_arena_sym, ta);
+		*m_arena = main_arena_sym;
+		// core->dbg->main_arena_resolved = true;
```
trufae, line 421: dont comment this chk

libr/core/linux_heap_glibc.c
```diff
@@ -355,13 +414,23 @@ static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
...
+		free (ta);
+		return true;
+	}
+
 	while (addr_srch < libc_addr_end) {
 		GH (update_main_arena) (core, addr_srch, ta);
 		if ( ta->GH(top) > brk_start && ta->GH(top) < brk_end &&
```
trufae, line 428: unnecessary space

libr/core/linux_heap_glibc.c
```diff
@@ -303,19 +336,45 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
 }
 
 static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
-	if (!core || !core->dbg || !core->dbg->maps) {
-		return false;
+
+	r_return_val_if_fail (core && core->dbg && core->dbg->maps, false);
+
+	/*
+	if (core->dbg->main_arena_resolved) {
```
trufae, line 343: uh?

libr/core/linux_heap_glibc.c
```diff
@@ -303,19 +336,45 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
...
+		return true;
 	}
+	*/
 
+	const char *main_arena_str = "main_arena";
```
trufae, line 348: only used once, just use this string in place remove this variable

libr/core/linux_heap_glibc.c
meowmeowxw: Yeah I was just trying to see if with r_bin_new the tests worked, as I did in a1dca99 is fine?

libr/core/linux_heap_glibc.c
meowmeowxw: right!

libr/core/linux_heap_glibc.c
```diff
@@ -303,19 +331,40 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
...
+	const char *libc_path = NULL;
+	GHT libc_addr = GHT_MAX;
+	GHT main_arena_sym = GHT_MAX;
 
 	if (r_config_get_i (core->config, "cfg.debug")) {
 		RListIter *iter;
 		RDebugMap *map;
 		r_debug_map_sync (core->dbg);
 		r_list_foreach (core->dbg->maps, iter, map) {
+			if (strstr (map->name, "/libc-") && map->perm == 4 && main_arena_sym == GHT_MAX) {
```
trufae, line 353: replace the '4' with R_PERM_R

meowmeowxw:
All test passes now a part:
`[XX] db/archos/linux-x64/dbg_dmh use of dmh, then any command which use grep and index`

The test is:

```
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -d -Qc 'db sym.main
dc
dmh > /dev/null
aa > /dev/null
pdf @sym.main~:1
' bins/elf/simple_malloc_x86_64
```

and the error is that it can't find the function main, however I tried to run the analysis first and it works fine:

```
R2_NOPLUGINS=1 ./radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -d -Qc 'aa > /dev/null                     
db sym.main
dc
dmh > /dev/null
pdf @sym.main~:1
' ../../test/bins/elf/simple_malloc_x86_64
```

The error resides in https://github.com/meowmeowxw/radare2/blob/59d1d76f31f0e094cc668d5528754b816bf6892c/libr/core/linux_heap_glibc.c#L29, in fact with `r_bin_new()` there isn't this problem. 

@trufae you said in https://github.com/radareorg/radare2/pull/17018#discussion_r439750772 to use r_bin_close(), however I didn't found that function in the API.

meowmeowxw:
All test passes now a part:
`[XX] db/archos/linux-x64/dbg_dmh use of dmh, then any command which use grep and index`

The test is:

```
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -d -Qc 'db sym.main
dc
dmh > /dev/null
aa > /dev/null
pdf @sym.main~:1
' bins/elf/simple_malloc_x86_64
```

and the error is that it can't find the function main, however I tried to run the analysis first and it works fine:

```
R2_NOPLUGINS=1 ./radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -d -Qc 'aa > /dev/null                     
db sym.main
dc
dmh > /dev/null
pdf @sym.main~:1
' ../../test/bins/elf/simple_malloc_x86_64
```

The error resides in https://github.com/meowmeowxw/radare2/blob/11165cd08a96fd73117d5a00c68e5c9b6b0452a5/libr/core/linux_heap_glibc.c#L32, in fact with `r_bin_new()` there isn't this problem. 

@trufae you said in https://github.com/radareorg/radare2/pull/17018#discussion_r439750772 to use r_bin_close(), however I didn't found that function in the API.

libr/core/linux_heap_glibc.c
trufae: ^

trufae:
good work!

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,37 @@
...
+	RBinSymbol *s;
+	const char *bin_file = bin->file;
+	ut64 baddr = r_config_get_i (core->config, "bin.baddr");
+	ut64 laddr = r_config_get_i (core->config, "bin.laddr");
+	int fd = r_io_fd_get_current (core->io);
+	int rawstr = bin->rawstr;
+
+	RBinOptions opt;
+	r_bin_options_init (&opt, -1, 0, 0, false);
+	r_bin_open (bin, path, &opt);
```
ret2libc, line 38: @trufae @meowmeowxw This use of RBin looks wrong IMHO. You end up keeping the binary file for libc opened, moreover you then re-open the binary under analysis, but there is already a binfile opened for it.

If you do `dmh` and then `ob`, you can see the opened binary files:
```
1 3 x86-64 ba:0x556f0b9c8000 sz:144970 /usr/bin/ls
0 4 x86-64 ba:0x00000000 sz:3178154 /usr/lib64/libc-2.30.so
2 3 x86-64 ba:0x556f0b9c8000 sz:144970 /usr/bin/ls
```

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,37 @@
...
+	RList *syms = r_bin_get_symbols (bin);
+
+	if (syms) {
```
ret2libc, line 41: no need for this check, you can just use `r_list_foreach` directly as it already checks for NULL.

libr/core/linux_heap_glibc.c
meowmeowxw: Do you suggests any alternative? I was using `r_bin_new` and then `r_bin_free` to close the bin but it forced the loading of the plugin twice: https://github.com/radareorg/radare2/pull/17018#discussion_r436851925

trufae: i didn't forced you :P just suggested that creating new instances is unnecessary payload when the plugins and libs are already there. you are calling RBin.open() twice, but never calling RBin.close(), this is why its leaking files

ret2libc: I think it should probably look something like:
```C
RBinFile *current_bf = r_bin_cur (bin);
r_bin_open (bin, path, &opt);
// get symbols as done now
RBinFile *libc_bf = r_bin_cur (bin);
r_bin_file_close(bin, bf->id);
r_bin_file_set_obj (bin, bf, bf->o);
```

I have not tested it, but unless there are bugs somewhere, I think these are the API we should use.

libr/core/linux_heap_glibc.c
```diff
@@ -303,20 +334,42 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
 }
 
 static bool GH(r_resolve_main_arena)(RCore *core, GHT *m_arena) {
-	if (!core || !core->dbg || !core->dbg->maps) {
-		return false;
+
```
ret2libc, line 337: unnecessary empty line

libr/core/linux_heap_glibc.c
```diff
@@ -303,20 +334,42 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
...
+		return true;
 	}
 
 	GHT brk_start = GHT_MAX, brk_end = GHT_MAX;
 	GHT libc_addr_sta = GHT_MAX, libc_addr_end = 0;
 	GHT addr_srch = GHT_MAX, heap_sz = GHT_MAX;
+	const char *libc_path = NULL;
+	GHT libc_addr = GHT_MAX;
+	GHT main_arena_sym = GHT_MAX;
+	bool is_debugged = true;
```
ret2libc, line 350: I think you could do
```suggestion
	bool is_debugged = r_config_get_i (core->config, "cfg.debug");
```
and use `is_debugged` everywhere in this function instead of always getting the value from `r_config_get_i`.

libr/core/linux_heap_glibc.c
ret2libc: Probably this check can be removed completely, as I think `map->name` should always be notnull.
There is a check in `r_debug_map_new` (though it misses the check of `map->name`).

libr/core/linux_heap_glibc.c
```diff
@@ -303,20 +334,42 @@ static void GH(print_arena_stats)(RCore *core, GHT m_arena, MallocState *main_ar
...
+			if (strstr (map->name, "/libc-") && map->perm == R_PERM_R && main_arena_sym == GHT_MAX) {
+				libc_path = map->name;
+				libc_addr = map->addr;
+				if (!libc_path) {
+					break;
+				}
+				char *path = strdup (libc_path);
+				if (path && r_file_exists (path)) {
+					GHT vaddr = GH (get_va_symbol) (core, path, "main_arena");
+					if (libc_addr != GHT_MAX && vaddr != GHT_MAX) {
```
ret2libc, line 366: probably `libc_addr != GHT_MAX` should be checked before, because if `libc_addr == GHT_MAX` we can avoid calling `r_file_exists` and `get_va_symbol`.

libr/core/linux_heap_glibc.c
ret2libc: I think that if `main_arena_resolved` this function should set the right value in `*m_arena`, otherwise you are already relying on the caller having the right value.

libr/core/linux_heap_glibc.c
meowmeowxw: Make sense

meowmeowxw:
> Why did you choose to use `r_core_file_cur` and `r_bin_reload` instead of what suggested in https://github.com/radareorg/radare2/pull/17018/files/7b8290f9ebc78346b01dc025342423fadfa4051a..e37aa952e2c0a8b3581458b8efa4adc4eae133e8#r441442438 ? Didn't it work?

@ret2libc Oh I didn't see that you commented, I found that snippet of code in https://github.com/radareorg/radare2/blob/aeca50166d3efa4d3339f8eb7a2bf676941f1614/libr/core/cmd_info.c#L421 and I check that was working as expected with `ob`.

Your method would be:

```c
RBin *bin = core->bin;
RBinFile *current_bf = r_bin_cur (bin); // this is the binary

RBinOptions opt;
r_bin_options_init (&opt, -1, 0, 0, false);
r_bin_open (bin, path, &opt);
// get symbols as done now

RBinFile *libc_bf = r_bin_cur (bin); // this is the libc
r_bin_file_close(bin, libc_bf->id);
r_bin_file_set_obj (bin, current_bf, current_bf->o);
```

Right?
However `r_bin_file_set_obj` is an R_IPI  function instead of R_API, should i use it anyway?

Do you see any problem with the current method?

ret2libc:
> > Why did you choose to use `r_core_file_cur` and `r_bin_reload` instead of what suggested in https://github.com/radareorg/radare2/pull/17018/files/7b8290f9ebc78346b01dc025342423fadfa4051a..e37aa952e2c0a8b3581458b8efa4adc4eae133e8#r441442438 ? Didn't it work?
> 
> @ret2libc Oh I didn't see that you commented, I found that snippet of code in
> 
> https://github.com/radareorg/radare2/blob/aeca50166d3efa4d3339f8eb7a2bf676941f1614/libr/core/cmd_info.c#L421
> and I check that was working as expected with `ob` (only 1 file open).
> 
> Your method would be:
> 
> ```c
> RBin *bin = core->bin;
> RBinFile *current_bf = r_bin_cur (bin); // this is the binary
> 
> RBinOptions opt;
> r_bin_options_init (&opt, -1, 0, 0, false);
> r_bin_open (bin, path, &opt);
> // get symbols as done now
> 
> RBinFile *libc_bf = r_bin_cur (bin); // this is the libc
> r_bin_file_close(bin, libc_bf->id);
> r_bin_file_set_obj (bin, current_bf, current_bf->o);
> ```
> 
> Right?
> However `r_bin_file_set_obj` is an R_IPI function instead of R_API, should i use it anyway?

Good point! Use `r_bin_file_set_cur_binfile` instead, that is the right R_API function ;)

> ~Do you see any problem with the current method?~ ok nvm it doesn't pass the test for the openbsd build, which is weird

Mmm weird, maybe it is not related to your changes. Fix these things first, then we'll have a look at bsd. Lately that enviroment is failing a lot :(

Thanks!

meowmeowxw:
`r_file_close` trigger a double free in my environment:

```
[0x0040135d]> ob
0 3 x86-64 ba:0x00400000 sz:20526 /home/meowmeow/CTF/test/heap/tcache
1 1050045120 -64 ba:0x7f5d3e966cc0 sz:12770042 �v�ȼU
^D
Do you want to quit? (Y/n) Y
Do you want to kill the process? (Y/n) Y
free(): double free detected in tcache 2
fish: “./radare2 -i script.rr2 -d ~/CT…” terminated by signal SIGABRT (Abort)
```

so I used `r_file_delete`:
```
[0x0040135d]> ob
0 3 x86-64 ba:0x00400000 sz:20526 /home/meowmeow/CTF/test/heap/tcache
```

ret2libc:
@meowmeowxw  Would you be willing to add a small unit test in `test/unit/test_bin.c` to try to reproduce this (as a separate PR) ?

meowmeowxw:
@ret2libc I think that I was calling `r_bin_file_close` with the bad argument that's why was crashing.

`r_bin_file_close (bin, libc_bf->id);`


The prototype is: `R_API bool r_bin_file_close(RBin *bin, int bd);` and there is no comment on what is `bd`. https://github.com/radareorg/radare2/blob/b18f7e4a2af78c23b878598c21d579af032c8b8c/libr/include/r_bin.h#L744

I grepped in the code `r_bin_file_close` but it's not used anywhere :(

Is `r_bin_file_delete` ok?

ret2libc:
> Is `r_bin_file_delete` ok? (It's also used inside `r_bin_reload`)

I guess so, yes. Those API will need some unit tests :(

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,32 @@
...
+	r_list_foreach (syms, iter, s) {
+		if (!strcmp (s->name, symname)) {
+			vaddr = s->vaddr;
+			break;
+		}
+	}
+
+	RBinFile *libc_bf = r_bin_cur (bin);
+	r_bin_file_delete (bin, libc_bf->id);
+	r_bin_file_set_cur_binfile (bin, current_bf);
```
trufae, line 46: this is another reason why cur must die. current-bf can be invalidated at any time. so using "safe" fd style identifiers would make that more error-prone and there wont be a need to tell r2 which is the current bin loaded, because the selected binfile can be held by RCore or just be selected depènding on the offset like its done by r2frida or when `e bin.at` is set

libr/core/linux_heap_glibc.c
```diff
@@ -21,6 +21,32 @@
...
+	RList *syms = r_bin_get_symbols (bin);
+
+	r_list_foreach (syms, iter, s) {
+		if (!strcmp (s->name, symname)) {
+			vaddr = s->vaddr;
+			break;
+		}
+	}
+
+	RBinFile *libc_bf = r_bin_cur (bin);
```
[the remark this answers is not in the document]
meowmeowxw: Ok I'll keep like this for now then

meowmeowxw:
@trufae I've checked the return of `r_bin_open`. I added the documentation to the new function and I left a TODO for the deprecated functions.
I would be happy to continue improving the heap viewer in the following months. :+1:

radare:
Great! Looking forward your future contributions on this!

PR #17021, opened 2020-06-09
Fixed hang when analysing objc binaries

gkpln3:

**Detailed description**

Fixed an issue when analysing objc binaries,  `count` can be 0xffffffff in some places, which in turn can lead to integer overflow, causing radare2 to try and analyse the whole memory space, effectively causing it to hang.

**Test plan**

**Closing issues**


trufae:
Can you provide a reproducer test file?

libr/core/anal_objc.c
```diff
@@ -194,6 +194,10 @@ static bool objc_find_refs(RCore *core) {
 		}
 
 		int count = readDword (&objc, classMethodsVA + 4);
+		if (count == -1) {
```
trufae, line 197: i think it will be better to make count ut32 (instead of int) and check for UT32_MAX instead of -1

trufae: and probably we can `break` instead of `continue` because such read may result in longer loading times

trufae: Also i have observed the return value of `r_io_read_at` inside the `readDword` function is ignored, it will be good to check for that and return UT32_MAX in case of failure. As long as ut32-max is a valid ut32 value maybe worth to return an error as 3rd argument as reference

gkpln3: So correct me if I am wrong but it is possible to get an invalid count before the end of the analysis, but getting a valid count afterwards, doesn't it?
If we `break` there it will stop the analysis all at once possibly missing the rest of the file.

As for the other comments, fixing now :)

trufae: it's not the same reading 0xffffffff than being unable to read from a specific address, so the errors shouldnt be combined with the data logic.

Thanks for the fixes!

trufae:
Please provide a test binary

gkpln3:
>

gkpln3:
This was the binary I was working on:
[Sketch.zip](https://github.com/radareorg/radare2/files/4753610/Sketch.zip)

libr/core/anal_objc.c
gkpln3: Oh I get what you mean.
Fixed.

trufae:
uhm we cannot use this binary for the testsuite, can you cook one small binary.

Also i have noticed that ^C is not handled in those loops, so it is not possible to stop it. Can you add a r_cons_is_breaked() check?

libr/core/anal_objc.c
```diff
@@ -182,18 +188,27 @@ static bool objc_find_refs(RCore *core) {
 
 	int total = 0;
 	ut64 off;
+	bool readSuccess = true;
 	for (off = 0; off < objc._data->vsize ; off += objc2ClassSize) {
+		if (!readSuccess) {
+			break;
+		}
+
```
trufae, line 196: allow the usr 

```c
if r_cons_is_breaked() {
  break;
}
```

libr/core/anal_objc.c
```diff
@@ -208,7 +223,11 @@ static bool objc_find_refs(RCore *core) {
 			if (isMsgRef) {
 				selRefVA -= 8;
 			}
-			ut64 funcVA = readQword (&objc, va2 + objc2ClassMethImpOffs);
```
trufae, line 211: also check for ^C in this inner loop

libr/core/anal_objc.c
```diff
@@ -62,12 +62,18 @@ static void objc_analyze(RCore *core) {
 }
 
 static ut64 getRefPtr(RCoreObjc *objc, ut64 classMethodsVA, bool *res) {
-	ut64 namePtr = readQword (objc, classMethodsVA);
+	*res = false;
+
+	bool readSuccess;
+	ut64 namePtr = readQword (objc, classMethodsVA, &readSuccess);
+	if (!readSuccess) {
+		return 0LL;
```
trufae, line 70: use UT64_MAX as invalid value . not 0

gkpln3:
> uhm we cannot use this binary for the testsuite, can you cook one small binary.
> 
> Also i have noticed that ^C is not handled in those loops, so it is not possible to stop it. Can you add a r_cons_is_breaked() check?

Yes, I'll try to make one today.
I've also added the `r_cons_is_breaked()` checks.

libr/core/anal_objc.c
```diff
@@ -182,23 +188,36 @@ static bool objc_find_refs(RCore *core) {
 
 	int total = 0;
 	ut64 off;
+	bool readSuccess = true;
 	for (off = 0; off < objc._data->vsize ; off += objc2ClassSize) {
```
mrmacete, line 192: The root cause of this issue is that this for is assuming that `__objc_data` is uniformly filled with objc classes which is not true.

So it ends up reading pointers in the middle of other data types and the -1 is ultimately caused by `classRoVA` not being an actual address, but some random else.

mrmacete: so maybe it's a good idea to just merge this to avoid the infinite loop, then fix the root cause for instance by iterating `__objc_classlist` and `__objc_catlist` instead and pull methods from there

gkpln3:
@trufae I've added the test, but not quite sure how to perform the check, there is no different output expected, I just expect the test to not take forever, is there a way to perform this kind of test?

trufae:
There's a timeout in r2r, so if the tests finishes, no output is expected, if it takes too long or crashes it will fail. Can you please commit the test in this PR?

trufae:
Can I have another approval and someone to merge this?

libr/core/anal_objc.c
ret2libc, line 192:
```suggestion
	for (off = 0; off < objc._data->vsize && readSuccess; off += objc2ClassSize) {
```
I think this it makes it easier to understand `readSuccess` makes the loop interrupt, instead of having it in a separate `if` just below.

libr/core/anal_objc.c
```diff
@@ -182,23 +188,36 @@ static bool objc_find_refs(RCore *core) {
 
 	int total = 0;
 	ut64 off;
-	for (off = 0; off < objc._data->vsize ; off += objc2ClassSize) {
+	bool readSuccess = true;
+	for (off = 0; off < objc._data->vsize && readSuccess; off += objc2ClassSize) {
+		if (!readSuccess || r_cons_is_breaked ()) {
```
ret2libc, line 193: @XVilka this `!readSuccess` needed to be removed :(

PR #17026, opened 2020-06-09
Use (void) instead of () in function signatures

trufae:

**Detailed description**

It's incorrect to define functions with no parameters like (), it is necessary to use (void) instead

This PR updates SDB ( i fixed that in sdb too), and the bug is also happening in tree-sitter, so @ret2libc may push the fix in the official repo or my fix will break on the next sync.

**Test plan**

```
git grep -e ^R_API -e ^static | grep '() {'
```

**Closing issues**

None, just 428 warnings less


trufae:
Beautiful

Screenshot 2020-06-09 at 20 19 53

shlr/tree-sitter/lib/src/query.c
```diff
@@ -259,7 +259,7 @@ static void stream_scan_identifier(Stream *stream) {
  * CaptureListPool
  ******************/
 
-static CaptureListPool capture_list_pool_new() {
+static CaptureListPool capture_list_pool_new(void) {
```
ret2libc, line 262: It's better if we don't touch these files but just work upstream. I will work with tree-sitter to include these fixes.

radare:
I know. I warned about this when i filled the issue. I touched them because the test will fail otherwise. Can you submit this fix in treesitter?

ret2libc:
> I know. I warned about this when i filled the issue. I touched them because the test will fail otherwise. Can you submit this fix in treesitter?

Apparently they already fixed it, 7 days ago. https://github.com/tree-sitter/tree-sitter/commit/81d533d2d1b580fdb507accabc91ceddffb5b6f0  . They haven't done a release yet though.

XVilka:
Can you just ignore that directory in the test? If not, and the required changes in the tree-sitter already, I think it's fine to merge as is, until the tree-sitter will make a new release.

ret2libc:
If we can't wait for a release, update the `TS_TIP` in shlr/makefile and shlr/meson.build and run a `make tree-sitter-sync`. It's better than applying random fixes.

trufae:
i've reverted the treesitter fixes and the test is only checking in libr/

trufae:
anything left?

PR #17031, opened 2020-06-09
Fix signedness issue in the DWARF line parser on ARM64

HoundThe:
(tests are already there)

I am abusing PR so I can run tests on ARM to check if I found the error  for #15688 
I think it might be because signedness of `char` is implementation defined and the line_base is supposed to be -5, but instead it's 251 which is -5 on 8bit 2's complement


HoundThe:
Seems that it worked, it was the signedness issue
https://travis-ci.com/github/radareorg/radare2/builds/170593567

trufae:
scary that a char can be unsigned somewhere  :S good finding!

XVilka:
Maybe use the usual int instead of `signed char` (convert upon reading)?

PR #17046, opened 2020-06-11
Fix compiler warnings

ret2libc:

**Detailed description**

There was a "function not used" warning related to pyc_dis and then a bunch of warnings because of recent changes to Sdb callback.

**Test plan**

Check the log of CI to see if those warnings are still there or not.

### DIFF
--- a/libr/parse/code.c
+++ b/libr/parse/code.c
@@ -23,9 +23,9 @@ static void __appendString(const char *msg, char **s) {
 	}
 }
 
-static int __typeLoad(void *p, const char *k, const char *v) {
+static bool __typeLoad(void *p, const char *k, const char *v) {
 	if (!p) {
-		return -1;
+		return false;
 	}
 	int btype = 0;
 	RAnal *anal = (RAnal*)p;
@@ -71,7 +71,7 @@ static int __typeLoad(void *p, const char *k, const char *v) {
 		}
 		tcc_sym_push ((char *)typename, typesize, btype);
 	}
-	return 0;
+	return true;
 }
 
 static void __errorFunc(void *opaque, const char *msg) {
--- a/libr/anal/class.c
+++ b/libr/anal/class.c
@@ -740,7 +740,7 @@ typedef struct {
 	const char *class_name;
 } DeleteClassCtx;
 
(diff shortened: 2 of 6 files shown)


trufae:
Screenshot 2020-06-11 at 13 14 14

libr/parse/code.c
```diff
@@ -71,7 +71,7 @@ static int __typeLoad(void *p, const char *k, const char *v) {
 		}
 		tcc_sym_push ((char *)typename, typesize, btype);
 	}
-	return 0;
+	return true;
```
trufae, line 74: this condition seems inverted. so the behaviour has changed

wargio: if everything went well, it should return true, not false. before false was `-1` and true was `0` in this func

radare:
the test fails because of this change i think

wargio:
```
[XX] db/cmd/types to error.h
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'to bins/other/error.h
t~char,double,float,int,long,short,size,uid,void # basic universal types
t~addr,date,dox,name
ts~addr,date,dox,name
t addr
' -
-- stderr
@@ -1 +1,7 @@
+Not found
+Not found
+Not found
+Not found
+Not found
+Not found
bins/other/error.h:2: warning: #error your compiler doesn't have support to my API
```
Failed test

libr/parse/code.c
ret2libc: yeah i noticed that while writing, however the logic in this function seems a bit weird, as it should `return false` so it would just get the first element.

ret2libc: Indeed I think this code is never really executed...  See https://codecov.io/gh/radareorg/radare2/src/fab89a61d9bc969c15d006e950eb0e4f2d5101bd/libr/parse/code.c#L26 . The real part of the function is never tested and I doubt it worked well, because the loop was interrupted immediately due to this `return 0`.

We have 2 options:
1) adjust the tests to include the `Not found` message
2) remove the `Not found` message from tccgen.c

Preferences?

trufae: 2

wargio: 2

PR #17048, opened 2020-06-11
Fix a fatmach0 test under aarch64 ##test

thestr4ng3r:

**Detailed description**

When the host where r2 runs is aarch64, the 64 bit part of the binary is selected by default and the test fails, so I select the arch/bits explicitly.
Since I never worked with fatmach0 stuff I am not entirely sure whether switching with `oa` is the correct approach here so please review.

**Test plan**

Run the tests on a 64 bit arm system.

### DIFF
--- a/test/db/formats/mach0/mach0
+++ b/test/db/formats/mach0/mach0
@@ -136,10 +136,8 @@ RUN
 
 NAME=mach0 headers2
 FILE=bins/mach0/ired
-CMDS=<<EOF
-oa arm 32 bins/mach0/ired
-iH
-EOF
+ARGS=-aarm -b32
+CMDS=iH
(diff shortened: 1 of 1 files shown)


ret2libc:
Can't you specify the bits to select with `-b 32`? I remember there was some option for fat bins.

trufae:
Yes, from commandline you can select the slice with -a and -b, by using the RBin api you can select by index too

trufae:
I would say to use ARGS instead of reopening. but maybe good to test both ways

thestr4ng3r:
Related: #17049

thestr4ng3r:
Updated to use ARGS.
Yes I think oa needs tests in general.

PR #17052, opened 2020-06-12
Fix NULL deref and heap-buffer-overflow

GustavoLCR:

**Test plan**

Tests pass

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -1373,10 +1373,10 @@ static int var_cmd(RCore *core, const char *str) {
 	}
 	switch (str[1]) { // afv[bsr]
 	case '\0':
-	case '*':
+	case '*': // "afv[bsr]*"
 		r_anal_var_list_show (core->anal, fcn, type, str[1], NULL);
 		break;
-	case 'j': {
+	case 'j': { // "afv[bsr]j"
 		PJ *pj = pj_new ();
 		if (!pj) {
 			return -1;
@@ -1386,7 +1386,7 @@ static int var_cmd(RCore *core, const char *str) {
 		pj_free (pj);
 	}
 		break;
-	case '.':
+	case '.': // "afv[bsr]."
 		r_anal_var_list_show (core->anal, fcn, core->offset, 0, NULL);
 		break;
 	case '-': // "afv[bsr]-"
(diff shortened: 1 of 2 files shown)


libr/core/cmd_meta.c
```diff
@@ -905,9 +905,8 @@ void r_comment_vars(RCore *core, const char *input) {
 		eprintf ("Can't find function here\n");
 		return;
 	}
-	oname = name = strdup (input + 2);
-	while (*name == ' ') {
-		name++;
+	if (input[1]) {
+		name = r_str_trim_dup (input[1] != ' ' ? input + 2 : input + 1);
```
trufae, line 909: this condition seems wrong/unnecessary, checking for a space when passing the string to a function that trim spaces is unnecessary

libr/core/cmd_anal.c
```diff
@@ -1460,6 +1460,10 @@ static int var_cmd(RCore *core, const char *str) {
 			var_help (core, type);
 			break;
 		}
+		if (!fcn) {
```
trufae, line 1463: maybe worth using r_str_trim apis instead of this for+strchr, also the for loop is not using braces {}

libr/core/cmd_anal.c
```diff
@@ -1488,14 +1492,10 @@ static int var_cmd(RCore *core, const char *str) {
 		}
 		if ((type == 'b') && delta > 0) {
 			isarg = true;
-		} else if (type == 's' && fcn && delta > fcn->maxstack) {
+		} else if (type == 's' && delta > fcn->maxstack) {
```
trufae, line 1495: cant add a comment above, but "name" is just used as an alias for `p`.  rename the variable.

vartype must be const char * and size is always 4, so just make it const. also add a comment right to the case with the command name that is handling there

libr/core/cmd_meta.c
```diff
@@ -895,7 +895,7 @@ void r_comment_var_help(RCore *core, char type) {
 void r_comment_vars(RCore *core, const char *input) {
 	//TODO enable base64 and make it the default for C*
 	RAnalFunction *fcn = r_anal_get_fcn_in (core->anal, core->offset, 0);
-	char *oname = NULL, *name = NULL;
+	char *name = NULL;
 
 	if (input[1] == '?' || (input[0] != 'b' && input[0] != 'r' && input[0] != 's') ) {
```
trufae, line 900: asuming input[0] is != 0

PR #17055, opened 2020-06-12
Add missing ELF DT entries ##bin

XVilka:

**Detailed description**

Add missing ELF dynamic tag entries to be parsed.

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -512,6 +512,7 @@ static void set_default_value_dynamic_info(ELFOBJ *bin) {
 	bin->dyn_info.dt_pltrelsz = 0;
 	bin->dyn_info.dt_pltgot = ELF_ADDR_MAX;
 	bin->dyn_info.dt_hash = ELF_ADDR_MAX;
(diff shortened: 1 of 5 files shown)


trufae:
i think it's fine to initialize them even if we are not using them, but this should be displayed with `iH`

trufae:
Please fix the tests, new output is better

ret2libc:
> i think it's fine to initialize them even if we are not using them, but this should be displayed with `iH`

@XVilka yes, either you print or use them somehow or don't store the value at all (just handle them in the switch).

trufae:
add ##bin

ret2libc:
@XVilka ping . please just remove the additional unused variables and let's get rid of the common missing dynamic tags.

trufae:
ping

ret2libc:
@XVilka Sorry, I messed up somehow with your branch while trying to push this forward. Please have a look and let's merge this so we can remove those warnings that are common and I can go ahead with https://github.com/radareorg/radare2/pull/17195 . 

I have just kept the warnings for some uncommon MIPS dt entries which would be cool in the future to properly support though.

PR #17063, opened 2020-06-13
Adding support for BA2 ELF binaries ##bin

Jerome-PS:
- [?] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
>> but I could not use clang-format-diff...

**Detailed description**
Adding a new CPU architecture

...

**Test plan**
Opened several elf files compiled for this CPU.


trufae:
What about adding this architecture in radare2-extras to not make the core r2 bigger?

Jerome-PS:
> What about adding this architecture in radare2-extras to not make the core r2 bigger?

Yeah, sure it sounds like a good idea, is there a way to automatically download it or at least tell the user to do it when the elf decoder sees EM_BA2?

trufae:
Sure, as long as the arch type is registered in the bin plugin will try to find an asm/anal plugin that matches the given arch name.

Jerome-PS:
So what should I do to add it to radare2-extras?
Also someone mentioned "Meson changes and tests" on IRC.
Finally I'm not sure about clang-format-diff.py (I copied the recommit hook, but I'm not 100% it worked)...

trufae:
r2e (radare2-extras) doesnt have meson integration, so if you move the plugin there (which is where i think it should belong, until the plugin is complete and the community decides is quite widely used to get it merged into the main repo) you dont need to do meson stuff. and yes. having tests will be interesting too, you can write them into the r2e repo and run them from the ci. but r2e is not yet tied to the mandatory test pain

Jerome-PS:
This PR is still needed (even though the code has been placed in radare2-extras), because NXP uses a non-standard EM_ constant in the elf header...

libr/bin/format/elf/elf_specs.h
```diff
@@ -159,6 +159,7 @@
 #define EM_56800EX             200        /* Freescale 56800EX Digital Signal Controller (DSC) */
 #define EM_BA1                 201        /* Beyond BA1 CPU architecture */
 #define EM_BA2                 202        /* Beyond BA2 CPU architecture */
+#define EM_BA2_NON_STANDARD 0x8472        /* Beyond BA2 CPU architecture */
```
wargio, line 162: I would suggest to add in the comment a link to NXP doc that defines this.

Jerome-PS: Actually I extracted this value from elf files generated by the NXP toolchain.
I'll check if I find a documentation somewhere....

wargio: Ah. ok, if this is not mapped anywhere, let me know, and i'll merge this.

Jerome-PS: I found a reference in the binutils version 2.22:
`binutils-2.22-ba-r33675/include/elf/common.h:362:#define EM_BA			0x8472`
A tarball of the corresponding binutils can be found here: https://github.com/alephsecurity/BA2-toolchain

wargio: hmm. then don't worry.

PR #17065, opened 2020-06-13
Bugfix: Zignatures corrupt JSON output in types

xermicus:

**Detailed description**

In the function `print_list_type_body` two new JSON contexts are created if there are 2 or more items to print (see function `print_list_separator`). However, before return these two contexts are closed anyways, without respecting whether they were actually created, resulting in corrupt JSON under some circumstances.

**Test plan**

`rasign2 -aj /bin/ls | jq .`

*Before patch*
```
Error: Parse error on line 1:
...args","type":"1"}]}],"hash":{"bbhash":"f
-----------------------^
Expecting 'EOF', got ','
```
*After patch*
```
rasign2 -aj ls | jq . ; echo $?`
...
0
```

**Closing issues**

...

### DIFF
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -1493,6 +1493,10 @@ static void print_function_args_json(RAnal *a, PJ *pj, char *arg_type) {
 }
 
 static void print_type_json(RAnal *a, char *type, PJ *pj, size_t pos) {
+	if (pos == 0) {
+		return;
+	}
+
 	char *str_type = strchr (type, '=');
 
 	if (str_type == NULL) {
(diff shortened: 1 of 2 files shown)


libr/anal/sign.c
```diff
@@ -1545,7 +1545,7 @@ static void print_list_type_body(RAnal *a, RSignItem *it, PJ *pj, int format) {
 		i++;
 	}
 
-	if (format == 'j') {
+	if (format == 'j' && i > 2) {
```
trufae, line 1548: uhm seems hacky. whats going on inside print_type_json that only works if there are two types printed first?

xermicus: It's in [`print_list_seperator`](https://github.com/radareorg/radare2/blob/0adbb4abec422ac5432d177d12232948bcf94eb2/libr/anal/sign.c#L1520)

I'm not sure why this can be an array in any case for JSON output. And apparently I think this is more broken. For example I have a function with 3 function args, while `z` shows them correctly, `zj` does not... I'll try to fix this and provide some test cases.

xermicus:
@trufae I had a quick dive into this and tried to refactor it a bit to make it less hacky. Here's what I can tell:
* The thing looks hacky because the first element in the types list is how many args there are
* This gets special treatment in JSON output (this is what `only works if there are two types printed first`)
* Why skip the first two? I think this a bug (IMHO should just skip the first one, or is the second list element also special?) and the reason why function args are missing in JSON output. In the funciton `print_type_json`: If pos == 1 nothing is printed. Or is there a special reason for this?

I removed some clutter and the behavior now is:
* Skip the first type in the list. The first "type" in the list tells how many args there are and is technically not a type/arg itself.
* After that, every element in the list just gets appended to the JSON array.

If you're happy with that I can add a test case too

libr/anal/sign.c
```diff
@@ -1493,6 +1493,10 @@ static void print_function_args_json(RAnal *a, PJ *pj, char *arg_type) {
 }
 
 static void print_type_json(RAnal *a, char *type, PJ *pj, size_t pos) {
+	if (pos == 0) {
```
ret2libc, line 1496: why doesn't it print anything if `pos == 0`? It seems weird. As said by trufae, please add a test for these changes.

xermicus: As stated:
> Skip the first type in the list. The first "type" in the list tells how many args there are and is technically not a type/arg itself.

This is redundant information to beavoided (the length of the json array should tell you how many there are)

ret2libc: Ah yeah sorry, read it just now!

xermicus: np & thx for the review

ret2libc:
@XVilka please restart build.sr.ht.... it is failing quite often lately btw... Also, i'm not sure it should really be required to merge PR, at least not the testing part.

PR #17066, opened 2020-06-14
r2r: Support char diffing

kazarmy:

**Detailed description**

This pr adds support for char diffing in r2r. Example:

diffchar

This is useful if:
1. The test output has lines that are very long (e.g. JSON output).
1. You need to be absolutely sure as to what has changed.

This is for stdout only. Also, the return type of `r_str_rstr()` has been changed from `const char *` to `char *` in line with [strstr()](https://linux.die.net/man/3/strstr).

**Test plan**

Mess around with other tests and check the diffchar output.

**Closing issues**

...

**TODO** (if this pr is accepted)

- [ ] Use `--word-diff=porcelain` instead to totally eliminate ambiguity.

### DIFF
--- a/libr/util/str.c
+++ b/libr/util/str.c
@@ -631,7 +631,7 @@ R_API const char *r_str_rsep(const char *base, const char *p, const char *sep) {
 	return NULL;
 }
 
-R_API char *r_str_rstr(const char *base, const char *p) {
+R_API const char *r_str_rstr(const char *base, const char *p) {
 	char *s = strdup (base);
 	char *k = strdup (p);
 	r_str_reverse (s);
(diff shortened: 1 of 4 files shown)


libr/util/str.c
```diff
@@ -631,7 +631,7 @@ R_API const char *r_str_rsep(const char *base, const char *p, const char *sep) {
 	return NULL;
 }
 
-R_API const char *r_str_rstr(const char *base, const char *p) {
+R_API char *r_str_rstr(const char *base, const char *p) {
```
trufae, line 634: uhm, nope input is const, so output must be const

kazarmy: ok 2f842891fae5d3a4310c7a70c80ecde45e01ecd4

PR #17067, opened 2020-06-14
Implement RLang.spp for templated scripting ##lang

trufae:

**Detailed description**

SPP is integrated into r2 already, but wasn't yet used as a templated lang extension, this way it is possible to write template files for markdown, html or just plain scripts with conditionals and run r2 commands inline.

Screenshot 2020-06-14 at 11 49 47

I will refactor/rewrite SPP and add a proper testsuite for it, because the current code was supposed to be used as a standalone program and extended in a hacky way with globals to work inside r2 as a library. But it's not thread safe. But this will take a long run and its unrelated to this pr.

**Test plan**

I added a test, and the screenshot shows how to use that feature.

**Closing issues**

none. its a new feature

### DIFF
--- a/libr/lang/Makefile
+++ b/libr/lang/Makefile
@@ -5,7 +5,4 @@ OBJS=lang.o
 R2DEPS=r_util r_cons
 
 include ../rules.mk
-
-pop:
-	make
-	r2 -c '#!rust' -
+CFLAGS+=-I$(SHLR)/spp
--- a/libr/lang/lang.c
+++ b/libr/lang/lang.c
@@ -9,6 +9,7 @@ R_LIB_VERSION(r_lang);
 #include "p/vala.c"  // hardcoded
 #include "p/rust.c"  // hardcoded
 #include "p/zig.c"   // hardcoded
+#include "p/spp.c"  // hardcoded
 #include "p/c.c"     // hardcoded
 #include "p/v.c"     // hardcoded
 #include "p/lib.c"
@@ -50,6 +51,7 @@ R_API RLang *r_lang_new(void) {
 	r_lang_add (lang, &r_lang_plugin_vala);
 	r_lang_add (lang, &r_lang_plugin_rust);
 	r_lang_add (lang, &r_lang_plugin_zig);
+	r_lang_add (lang, &r_lang_plugin_spp);
 	r_lang_add (lang, &r_lang_plugin_pipe);
 	r_lang_add (lang, &r_lang_plugin_lib);
 	r_lang_add (lang, &r_lang_plugin_v);
@@ -257,6 +259,7 @@ R_API int r_lang_prompt(RLang *lang) {
 
 	/* foo */
 	for (;;) {
+		r_cons_flush ();
 		snprintf (buf, sizeof (buf)-1, "%s> ", lang->cur->name);
 		r_line_set_prompt (buf);
 #if 0
--- a/libr/lang/meson.build
+++ b/libr/lang/meson.build
@@ -8,12 +8,15 @@ r_lang_sources = [
   #'p/rust.c',
   #'p/vala.c'
   #'p/zig.c',
+  #'p/spp.c',
 ]
 
+spp_inc = [platform_inc, include_directories('../../shlr/spp')]
+
 r_lang = library('r_lang', r_lang_sources,
-  include_directories: [platform_inc],
+  include_directories: [platform_inc, spp_inc],
   c_args: library_cflags,
-  dependencies: [r_util_dep, r_cons_dep],
+  dependencies: [r_util_dep, spp_dep, r_cons_dep],
   install: true,
   implicit_include_directories: false,
   soversion: r2_libversion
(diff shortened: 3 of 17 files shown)


XVilka:
What about making it work with the rest of `r_lang` or `*.r2` scripts as well? IMHO, that would be more useful.

trufae:
SPP is used for assembling files, to handle the include statements and such, but for RLang that's a bit more complex because each language can have different quirks, and the spp evaluations will only work in the provided file, not across all loaded files by the interpreter. Not to mention this works only by slurping files, not for executing them, so the runfile() callback will make no sense to exist because it will need to slurp the file into rlang to evaluate it before calling the lang eval instead of just telling the interpreter to run a file

trufae:
any review/comments here @ret2libc @thestr4ng3r @wargio ?

wargio:
are you sure you want all that code inside a .h file? @trufae

trufae:
@wargio this is up to the spp refactoring im doing in the spp repository. all the other spp preprocessors are in .h files in the current spp. so imho this should change, but in spp first. i just tried to follow the same logic here. as a temporary solution i can name it .inc like we did for rbin (but i dont like this .inc thing because it makes my vim not detct filetype well)

libr/lang/p/spp_r2.h
```diff
@@ -0,0 +1,470 @@
...
+	spp_r2_switch_str = NULL;
+	return -1;
+}
+
+static TAG_CALLBACK(spp_r2_endpipe) {
+#if HAVE_SYSTEM
+	/* TODO: Get output here */
+	int ret = 0, len = 0;
+	int outlen = 4096;
+	char *str = (char *)malloc (4096);
```
wargio, line 366: no check on the malloc output.

libr/lang/p/spp_r2.h
```diff
@@ -0,0 +1,470 @@
...
+}
+
+static int spp_r2_var_set(const char *var, const char *val) {
+	return r_sys_setenv (var, val);
+}
+
+#if HAVE_SYSTEM
+/* Should be dynamic buffer */
+static char *cmd_to_str(const char *cmd) {
+	char *out = (char *)calloc (4096, 1);
```
wargio, line 18: missing check on calloc output

libr/lang/p/spp_r2.h
```diff
@@ -0,0 +1,470 @@
...
+	return 0;
+}
+
+static TAG_CALLBACK(spp_r2_include) {
+	if (!state->echo[state->ifl]) {
+		return 0;
+	}
+	char *incdir = getenv("SPP_INCDIR");
+	if (incdir) {
+		char *b = strdup (incdir);
```
wargio, line 199: missing check on strdup output

libr/lang/p/spp_r2.h
```diff
@@ -0,0 +1,470 @@
...
+		else state->echo[state->ifl + 1] = 0;
+//fprintf(stderr, "IFEQ(%s)(%s)=%d\n", buf, value, echo[ifl]);
+	}
+	return 1;
+}
+
+static TAG_CALLBACK(spp_r2_hex) {
+	int i;
+	for(i = 0; buf[i]; i++) {
+		if (buf[i] >= '0' && buf[i] <= '9') {
```
wargio, line 244: don't we have IS_NUMERIC macro ?

libr/lang/p/spp_r2.h
```diff
@@ -0,0 +1,470 @@
...
+}
+
+static char *spp_r2_switch_str = NULL;
+
+static TAG_CALLBACK(spp_r2_switch) {
+	char *var = spp_r2_var_get (buf);
+	if (var) {
+		spp_r2_switch_str = strdup (var);
+	} else {
+		spp_r2_switch_str = strdup ("");
```
wargio, line 345: maybe place a `strdup` check?

wargio:
> @wargio this is up to the spp refactoring im doing in the spp repository. all the other spp preprocessors are in .h files in the current spp. so imho this should change, but in spp first. i just tried to follow the same logic here. as a temporary solution i can name it .inc like we did for rbin (but i dont like this .inc thing because it makes my vim not detct filetype well)

no, do not worry about that.

libr/lang/p/spp_r2.h
trufae: it's named isdigit in posixland

libr/lang/p/spp_r2.h
trufae: i dont think its needed in case of being null something really terrible should be going on and the worst case we will get the var assigned to null

trufae:
@wargio all comments addressed

wargio:
Tests are failing. Please check them.

trufae:
https://twitter.com/trufae/status/1277235437564813312?s=21

wargio:
@trufae looks like your tests are failing now.

trufae:
Tests where faiuling because of a bug in r2r -i. i manually fixed them,. should be green again

wargio:
Appveyor issue is related to r2pipe not this.

PR #17073, opened 2020-06-15
Fix invalid json output for drtj command

### LINKED ISSUE #16635: drtj fpu returns invalid json
### Work environment

| OS/arch/bits (mandatory)                             | Ubuntu x86 64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.5.0-git 1 @ linux-x86-64 git.4.5.0-git commit: dae5bb43ee8e3c6ba9964caf6569ce28918dd000 build: 2020-04-16__23:30:26

### Expected behavior
`drtj fpu` should return valid json.

### Actual behavior
it does not. Instead, at least for registers `st<x>` r2 decides to return hex, which is not valid json.

```
[0x7f1634629090]> drtj fpu
{"cwd":895,"swd":0,"ftw":0,"fop":0,"frip":0,"frdp":0,"mxcsr":8064,"mxcr_mask":0,"st0":0x00000000000000000000,"st1":0x00000000000000000000,"st2":0x00000000000000000000,"st3":0x00000000000000000000,"st4":0x00000000000000000000,"st5":0x00000000000000000000,"st6":0x00000000000000000000,"st7":0x00000000000000000000,"xmm0h":0,"xmm0l":0,"xmm1h":0,"xmm1l":0,"xmm2h":0,"xmm2l":0,"xmm3h":0,"xmm3l":0,"xmm4h":0,"xmm4l":0,"xmm5h":0,"xmm5l":0,"xmm6h":0,"xmm6l":0,"xmm7h":0,"xmm7l":0,"xmm8h":0,"xmm8l":0,"xmm9h":0,"xmm9l":0,"xmm10h":0,"xmm10l":0,"xmm11h":0,"xmm11l":0,"xmm12h":0,"xmm12l":0,"xmm13h":0,"xmm13l":0,"xmm14h":0,"xmm14l":0,"xmm15h":0,"x64":0}
```

Note the "0x" indicating hex where hex is not valid.

- Simply try using `drtj fpu` to notice hex. Even more apparent if you attempt `r2.cmdj('drtj fpu')` as it will cause an exception

abcSup:

**Detailed description**
Putting double quotes between hexadecimal big numbers to make them JSON strings, because big numbers are not generally supported in JSON and string is good enough.

**Test plan**
Test that `drtj fpu` or any other register types outputs a valid json. I am thinking of including it in the json test. However, every time r2r runs, it says `Skipping json tests because jq is not available.`, although I have `jq` installed and tested. The `jq` process seems to always exit with an error code. I have found that the sourcehut ci does not run the json test as well. Is there a bug or configuration error?

**Closing issues**

closes #16635

### DIFF
--- a/libr/debug/dreg.c
+++ b/libr/debug/dreg.c
@@ -92,6 +92,7 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char 
 	RList *head;
 	ut64 diff;
 	char strvalue[256];
+	bool isJson = (rad == 'j' || rad == 'J');
 	if (!dbg || !dbg->reg) {
 		return false;
 	}
@@ -120,8 +121,15 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char 
 	if (dbg->regcols) {
 		cols = dbg->regcols;
 	}
-	if (rad == 'j') {
-		dbg->cb_printf ("{");
+	PJ *pj;
+	if (isJson) {
+		pj = pj_new ();
+		if (!pj) {
+			return false;
+		}
+		if (rad == 'j') {
+			pj_o (pj);
+		}
 	}
 	// with the new field "arena" into reg items why need
 	// to get all arenas.
@@ -171,9 +179,9 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char 
 			r_reg_arena_swap (dbg->reg, false);
 			diff = r_reg_get_value (dbg->reg, item);
 			r_reg_arena_swap (dbg->reg, false);
-			delta = value-diff;
-			if (tolower ((ut8)rad) == 'j') {
-				snprintf (strvalue, sizeof (strvalue),"%"PFMT64u, value);
+			delta = value - diff;
+			if (isJson) {
+				pj_kn (pj, item->name, value);
 			} else {
 				if (pr && pr->wide_offsets && dbg->bits & R_SYS_BITS_64) {
 					snprintf (strvalue, sizeof (strvalue),"0x%016"PFMT64x, value);
@@ -200,16 +208,17 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char 
 				default:
 					snprintf (strvalue, sizeof (strvalue), "ERROR");
 			}
+			if (isJson) {
+				pj_ks (pj, item->name, strvalue);
+			}
 			delta = 0; // TODO: calculate delta with big values.
 		}
 		itmidx++;
 
+		if (isJson) {
+			continue;
+		}
 		switch (rad) {
-			case 'J':
-			case 'j':
-				dbg->cb_printf ("%s\"%s\":%s",
-						n?",":"", item->name, strvalue);
-				break;
 			case '-':
 				dbg->cb_printf ("f-%s\n", item->name);
 				break;
@@ -276,10 +285,12 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char 
 		n++;
 	}
 beach:
-	if (rad == 'j') {
-		dbg->cb_printf ("}\n");
-	} else if (rad == 'J') {
-		// do nothing
+	if (isJson) {
+		if (rad == 'j') {
+			pj_end (pj);
+		}
+		dbg->cb_printf ("%s\n", pj_string (pj));
+		pj_free (pj);
 	} else if (n > 0 && (rad == 2 || rad == '=') && ((n%cols))) {
 		dbg->cb_printf ("\n");
 	}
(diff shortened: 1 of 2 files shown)


ret2libc:
cc @thestr4ng3r could you double check if/how r2r does not run js tests?

abcSup:
> cc @thestr4ng3r could you double check if/how r2r does not run js tests?

I found out that `jq` always exits with code 2 and outputs its help message to `stderr`. This is because `r2r` writes the json to its stdin after the process ran. I can submit a PR to fix this by running `jq .` instead with the extra argument `.`, so `jq` will wait for `stdin` input before exiting with code 2 (usage error).

> Even if this change fixes the output, the PR doesn't have any test to verify that, and the whole drtj logic should be rewritten to use the safe PJ api instead of formatting it by hand.

You are right. I am thinking about using `r2`'s json test to validate the output, but I just realized that `drtj fpu` only outputs the registers when in debug mode. Otherwise, it will just output the available register types. Any suggestions on how to test this without getting into debug mode because I cannot think of a way to validate the json in a `cmd` test?

XVilka:
I suggest to omit the test due to the complexity of testing debug mode, but proceed with PJ API, which is usually better tested anyway.

abcSup:
> I suggest to omit the test due to the complexity of testing debug mode, but proceed with PJ API, which is usually better tested anyway.

Great. I will update the PR to use the pj API instead for now.

libr/debug/dreg.c
```diff
@@ -276,10 +279,10 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char
 		n++;
 	}
 beach:
-	if (rad == 'j') {
-		dbg->cb_printf ("}\n");
-	} else if (rad == 'J') {
-		// do nothing
+	if (rad == 'j' || rad == 'J') {
+		pj_end (pj);
+		r_cons_printf ("%s\n", pj_string (pj));
```
ret2libc, line 284:
```suggestion
		dbg->cb_printf ("%s\n", pj_string (pj));
```

libr/debug/dreg.c
```diff
@@ -120,8 +120,10 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char
 	if (dbg->regcols) {
 		cols = dbg->regcols;
 	}
-	if (rad == 'j') {
-		dbg->cb_printf ("{");
+	PJ *pj = NULL;
+	if (rad == 'j' || rad == 'J') {
+		pj = pj_new ();
```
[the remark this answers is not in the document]
trufae: i think that we can introduce some custom nullchks to avoid having them on release builds. What do you think @ret2libc ? i think that small allocations like this shuold not fail except if things are very wrong somewhere, so maybe we can shortten some more paths by asuming fixed small allocations wont fail

libr/debug/dreg.c
```diff
@@ -171,9 +173,9 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char
 			r_reg_arena_swap (dbg->reg, false);
 			diff = r_reg_get_value (dbg->reg, item);
 			r_reg_arena_swap (dbg->reg, false);
-			delta = value-diff;
-			if (tolower ((ut8)rad) == 'j') {
-				snprintf (strvalue, sizeof (strvalue),"%"PFMT64u, value);
+			delta = value - diff;
+			if (rad == 'j' || rad == 'J') {
```
trufae, line 177: so many checks for rad=='j' || 'J'. add a const bool isJson = (rad == 'j' ||| rad == 'J') on top and just chk for this variable in the rest of places

libr/debug/dreg.c
```diff
@@ -276,10 +279,10 @@ R_API int r_debug_reg_list(RDebug *dbg, int type, int size, int rad, const char
...
 beach:
-	if (rad == 'j') {
-		dbg->cb_printf ("}\n");
-	} else if (rad == 'J') {
-		// do nothing
+	if (rad == 'j' || rad == 'J') {
+		pj_end (pj);
+		dbg->cb_printf ("%s\n", pj_string (pj));
+		pj_free (pj);
 	} else if (n > 0 && (rad == 2 || rad == '=') && ((n%cols))) {
```
trufae, line 286: spaces around %

abcSup:
> I'm not sure why there are checks for `J`, but anyway... this looks good! Thanks!

@ret2libc I realized that too. It seems like when using 'J' `r_debug_reg_list` should not enclose the object with braces { } based on the usage here.
https://github.com/radareorg/radare2/blob/2f78b9f5df2004a509b095e3b873df61fd91fe3f/libr/core/cmd_anal.c#L4046-L4055

libr/debug/dreg.c
abcSup: You are right. I think I will just track down all the usage of 'J' and replace all of them with 'j'.

libr/debug/dreg.c
abcSup: Do you mean writing `dbg->cb_printf (" %s \n", pj_string (pj));` instead? But, why and is it necessary for a json string?

PR #17075, opened 2020-06-15
Run fuzzed-tests in github actions ##tests

ret2libc:
Apparently fuzzed tests were not run on CI.
This should fix it, let's check.


ret2libc:
https://github.com/radareorg/radare2/pull/17110 needed for this

XVilka:
Please rebase on top of the master.

.github/workflows/pr.yaml
```diff
@@ -102,7 +102,7 @@ jobs:
         fi
         cd test
         radare2 -N -Qc 'e cfg.newshell' -
-        make
+        make && make fuzz-tests
```
ret2libc, line 105: @XVilka @trufae maybe we don't want to run fuzz-tests on each PR ? it does ~11 minutes to -test builds.

trufae:
So.. this is the reason why r2r.c is faster than r2r.js ? :D

trufae:
we can use rand() to determine when to run the fuzztests

thestr4ng3r:
Nah, r2r.js never ran fuzz tests either. Sounds like something for an action once per week maybe.

ret2libc:
> So.. this is the reason why r2r.c is faster than r2r.js ? :D

I don't think so, but I didn't look too much into r2r.js and r2r.c. However, https://travis-ci.com/github/radareorg/radare2/jobs/261901458#L2827 is from 7 months ago and it doesn't seem to execute fuzzed tests. https://travis-ci.org/github/radareorg/radare2/jobs/579965623#L2242 is from 10 months ago and it doesn't seem to read fuzzed tests. So I would say that, at least way before r2r.c was introduced, we weren't running fuzzed tests on every commit.

> we can use rand() to determine when to run the fuzztests

Is it ok if we run it on each commit on master but not on PRs? Or we can run it before releases.

wargio:
@trufae @ret2libc i think some test should really be skipped in PRs. i actually agree that it might break stuff, but fuzz bins are tricky. i would prefer to see them only on radareorg branches not on PRs

PR #17085, opened 2020-06-16
Use `stale` label instead of `wontfix`

ret2libc:


XVilka:
Should have used from the beginning.

PR #17087, opened 2020-06-16
r2r: For diffchar, use --word-diff=porcelain instead

kazarmy:

**Detailed description**

Diff output is the same as #17066 except that a stray `{+`, `[-`, `+}` or `-]` in the test output does not break the diff.

Btw, an example of `--word-diff=porcelain` output is as follows:

porcelain

**Test plan**

Mess around with tests and check that the diffchar output is the same as for #17066.

**Closing issues**

...


XVilka:
Looks great, also simplifies the code. I use porcelain in my scripts as well. 👍

PR #17102, opened 2020-06-19
Fix #17093 - Fix uaf in var accesses

### LINKED ISSUE #17093: Segfault after hit search
### Work environment

| OS/arch/bits (mandatory)                             | Fedora 31 x86_64
| File format of the file you reverse (mandatory)      | ELF 64-bit
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.5.0-git 4220 @ linux-x86-64 git.4.4.0-300-ge3baa737 commit: e3baa7376f9d7ac6ad8a28b3ce461a7e11b0325a build: 2020-06-17__08:55:52

### Actual behavior
r2 segfaults after hitting search pattern.
### Steps to reproduce the behaviour
```
r2 get_it
aa
s main
pd15
            ;-- main:
            0x004005c7      55             push rbp
            0x004005c8      4889e5         mov rbp, rsp
            0x004005cb      4883ec30       sub rsp, 0x30
            0x004005cf      897ddc         mov dword [rbp - 0x24], edi
            0x004005d2      488975d0       mov qword [rbp - 0x30], rsi
            0x004005d6      bf8e064000     mov edi, str.Do_you_gets_it ; 0x40068e ; "Do you gets it??"
            0x004005db      e890feffff     call sym.imp.puts
            0x004005e0      488d45e0       lea rax, [rbp - 0x20]
            0x004005e4      4889c7         mov rdi, rax
            0x004005e7      b800000000     mov eax, 0
            0x004005ec      e8affeffff     call sym.imp.gets
            0x004005f1      b800000000     mov eax, 0
            0x004005f6      c9             leave
            0x004005f7      c3             ret
            0x004005f8      0f1f84000000.  nop dword [rax + rax]
db 0x004005f1
ood
dc
Do you gets it??
12435678
hit breakpoint at: 4005f1
/ 12435678
Searching 8 bytes in [0x0-0x7fffffffffffffff]
[# ]^C0x00adff00 < 0x7fffffffffffffff  hits = 1   

hits: 1
0x008ea6b0 hit0_0 .12435678.
```
When r2 got first hit I did Ctrl+C
Now, after I try to `V` r2 segfaults (s0metimes segfaults even trying `s`eek to hit0_0)
[bin_and_core.tar.gz](https://github.com/radareorg/radare2/files/4790758/bin_and_core.tar.gz)

GustavoLCR:

**Test plan**

Tests pass

**Closing issues**

Fixes #17093

### DIFF
--- a/libr/anal/var.c
+++ b/libr/anal/var.c
@@ -438,7 +438,7 @@ R_API void r_anal_var_set_access(RAnalVar *var, const char *reg, ut64 access_add
 		acc = r_vector_insert (&var->accesses, index, NULL);
 		acc->offset = offset;
 		acc->type = 0;
-		acc->reg = reg;
+		acc->reg = r_str_constpool_get (&var->fcn->anal->constpool, reg);
 	}
 
 	acc->type |= (ut8)access_type;


ret2libc:
@GustavoLCR I have not reproduced the mentioned issue, could you briefly explain how your change fixes the issue? From the ASAN trace in the issue it is not clear to me how this fixes the UAF :(

trufae:
The fix looks good. But constpool is O(n) which will introduce a slowdown (and those strings are never freed, which may end up into always growing memory problem).

The constpool can be improved by using a hashtable and a refcounting. What do you think about this design improvement of the constpool api?

ret2libc:
@trufae 
```
R_API void r_str_constpool_fini(RStrConstPool *pool) {
	ht_pp_free (pool->ht);
}

R_API const char *r_str_constpool_get(RStrConstPool *pool, const char *str) {
	if (!str) {
		return NULL;
	}
	HtPPKv *kv = ht_pp_find_kv (pool->ht, str, NULL);
	if (kv) {
		return kv->key;
	}
	ht_pp_insert (pool->ht, str, NULL);
	kv = ht_pp_find_kv (pool->ht, str, NULL);
	if (kv) {
		return kv->key;
	}
	return NULL;
}

```
Seems to be already implemented with a ht.

trufae:
oh i misschked libr/util/constr.c which seems to be unused. i will send a
PR killing it now

GustavoLCR:
> @GustavoLCR I have not reproduced the mentioned issue, could you briefly explain how your change fixes the issue? From the ASAN trace in the issue it is not clear to me how this fixes the UAF :(

@ret2libc var access was storing the `reg` pointer directly to a `RRegItem->name`, so if the regprofile changed, `reg` would be freed

PR #17104, opened 2020-06-19
Fix root help in newshell

ret2libc:

**Detailed description**

If root_cmddesc has `help` enabled, it will try to generate the help automatically. However, for now we don't have all the correct messages in place, so it would show wrong stuff.

By setting `help` to NULL we ensure that `?` just fallbacks to old command handler.

### DIFF
--- a/libr/core/cmd_api.c
+++ b/libr/core/cmd_api.c
@@ -72,6 +72,7 @@ R_API RCmd *r_cmd_new(void) {
 	cmd->nullcallback = cmd->data = NULL;
 	cmd->ht_cmds = ht_pp_new0 ();
 	cmd->root_cmd_desc = create_cmd_desc (cmd, NULL, R_CMD_DESC_TYPE_ARGV, "", NULL);
+	cmd->root_cmd_desc->help = NULL;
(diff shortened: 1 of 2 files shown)


karliss:
Shouldn't that be done in the command initializer function?

ret2libc:
> Shouldn't that be done in the command initializer function?

Do You mean in `create_cmd_desc`?

karliss:
Ok, just read the code in context. `create_cmd_desc` initializes it with `not_defined_help` if NULL get's passed.

karliss:
I am slightly confused why I didn't notice it during review. Was it broken during last minute changes after PR was approved?

ret2libc:
> I am slightly confused why I didn't notice it during review. Was it broken during last minute changes after PR was approved?

Exactly :D :D sorry about that.

ret2libc:
> I am slightly confused why I didn't notice it during review. Was it broken during last minute changes after PR was approved?

I thought "well, instead of doing the same assignment twice in new_argv and new_oldinput, let's just do it in create_cmd_desc", but that broke stuff :(

PR #17107, opened 2020-06-19
Allow disabling non-ascii char truncation from mostly ascii strings (str.truncnonascii) ##strings

kazarmy:

**Detailed description**

This pr allows the disabling of the truncation of non-ascii chars from ascii strings using str.truncnonascii.

Also, this pr removes the minstrlen check from r_bin_reset_strings() since it has no business checking it.

**Test plan**

**Closing issues**

...

### DIFF
--- a/libr/bin/bfile.c
+++ b/libr/bin/bfile.c
@@ -260,39 +260,41 @@ static int string_scan_range(RList *list, RBinFile *bf, int min,
 					}
 				}
 			}
-			switch (str_type) {
-			case R_STRING_TYPE_UTF8:
-			case R_STRING_TYPE_WIDE:
-			case R_STRING_TYPE_WIDE32:
-				num_blocks = 0;
-				block_list = r_utf_block_list ((const ut8*)tmp, i - 1,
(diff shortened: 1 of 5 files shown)


ret2libc:
I'm not sure I understand the purpose... If you are looking for ascii-strings, why non-ascii chars should be included. Would that mean you are actually looking for some other type of string?

kazarmy:
Sometimes there are (mostly) ascii strings that have a non-ascii char or 2 somewhere in the middle, for example currency symbols:

"The price of the cupboard is €10,20."

or have emoji:

"Hi there! 🥳 Going to the beach? 🏖️"

ret2libc:
Isn't there probably some issues with the type detection of strings? I'm not sure adding yet another var really solves the issue that the string detection may be broken.

XVilka:
But these are just Unicode UTF-8 strings.

kazarmy:
The string detection code is currently built for cases like #14499. The reason for the var is that it is impossible to implement string detection code that can satisfy all possible use cases.

kazarmy:
But perhaps detecting the entire string is not particularly needed for my purposes, so I'm closing this.

PR #17111, opened 2020-06-19
DWARF - type parsing into RAnalBaseTypes and saving into sdb

HoundThe:

**Description**

I am starting to prototype out the type parsing out of dwarf information, right now I got a big ugly and naive solution, but I am getting there. I wanted to show the approach I started with, because I am really unsure with it, so I started implementing something to have something to kick off. I didn't know better way to discuss this than in PR so I made one.

Right now I am parsing the types to RAnalBaseType format and my plan is to make appropriate type setting API inside anal/type.c as I've done with type getting.

I've kept the parsing methods inside dwarf.c as I think they belong there and exported the symbol so someone else will actually do the parsing.

* right now I am parsing atomic types into strings, I am not sure if this is correct, it might be because they are going to be stored into sdb anyway?
~~* there is a lot of hardcoded stuff I have to search for, for example for every attribute that I am looking for I have to search if the Die has such attribute.~~
~~* Lot of type stuff is parsed every time it's referenced, maybe implementing some kind of cache or map for already parsed types would be fine.~~ (Might not be necessary, a lot of the the references might fit into one cache line anyway as there is high spatial and reference locality?)
etc... I am sure I'll find a lot of additional stuff along the way
* Just a warning, there is a lot of ugly stuff so I can just try things faster, I will make it better once I am happy with the overall structure
What I am able to get right now from a simple C binary:

```
Struct, name: human
 Members:
  int : height
  int : weight
  char * : name
Struct, name: donkey
 Members:
  int : leg_count
  int : face_length
  char * : name
  human * : owner
```


HoundThe:
So I've added some more things to parse. Now I am curious. The form I am going to parse the types into, for example function pointers, what form should the type be? Proper C declaration? Example of the result now

```cpp
Struct, name: Bird
 Members:
  int () * * : _vptr.Bird
Struct, name: Dog
 Members:
Struct, name: Mammal
 Members:
  int () * * : _vptr.Mammal
```
vptrs are (in DWARF) pointer_type -> pointer_type -> subroutine -> int
the declaration is of course wrong, but should it be a valid C declaration? Or what should the format be for function pointers? I guess similar problems are gonna happen especially for other than C/C++ languages, how the types are going to be represented :thinking:

HoundThe:
I extended some of the information parsing, to properly parse function pointers, I must parse functions anyway, where and in which format should I save them into the Db? Is there some interface already?

Example of what we can print (we can parse more, so what we can print):
```cpp
struct MyStruct {
	unsigned char const * volatile a; 
	volatile unsigned char * const b; 
	MyStruct(volatile unsigned char * const arg) : b(arg){}
};

enum MyEnum {
	first = 1,
	third = 3,
	last = 99,
};

enum class ScopedEnum : unsigned char {
	a = 'a',
	b = 'b',
	z = 'z',
};

void func1();
void func2(int a);
int func3(int a);
void *func4(int a);
char func5(int *a, const char *str);

class MyClass {       
  public:             
	int myNum;        
	const char *myString;  
	const char &charRef;  
	void (*funPtr1)();
	void (*funPtr2)(int);
	int &&(*funPtr3)(int);
	void *(*funPtr4)(int, volatile int,const char);
	char &(*funPtr5)(int*, const char*);
	MyClass(const char &ref) : charRef(ref) {}
};
```
out
```
Structured type, name: MyStruct
 Members:
  unsigned char const * volatile : a;
  unsigned char volatile * const : b;
Enum, name: MyEnum
 Cases:
  first : 1;
  third : 3;
  last : 99;
Enum, name: ScopedEnum
 Cases:
  a : 97;
  b : 98;
  z : 122;
Structured type, name: MyClass
 Members:
  int : myNum;
  char const * : myString;
  char const & : charRef;
   () * : funPtr1;
   () * : funPtr2;
  int && () * : funPtr3;
   * () * : funPtr4;
  char & () * : funPtr5;
```

HoundThe:
I've added saving of structs into the sdb, I made inspiration by taking the source code of the binary and saving it with `to` and then trying to make it identical. 

I am unsure tho how to deal with some types, how should I save multidimensional arrrays? 
This is for example how `to` stored `char buffer[100][50]` - `struct.MyClass.buffer=char,9,100`
Only the leftmost dimension is there on the right and I guess storing it into a type (char [100][50]) wouldn't work.
What about storing C++ specialities, like reference and rvalue reference (&, &&).

I also think about adding size and maybe a name directly to the BaseType structure, for one it will mean the size of the whole type in memory (structure size = all members + padding, enum = how big is single case). There is also possibility to store a type of the enum itself, like enum classes in C++ - `enum class MyEnum : unsigned char {}` but that doesn't seem that useful, the size itself would be the most important. Then size of each member in the structure/unit and their offset in the structured type, after all thats saved in the sdb always aswell. 

Then a name would be valuable, because afterall it's always necessary to be saved (because it's a key in sdb) so why always pass it as an argument. With this comes up a problem, what about multiple identical types, DWARF types are per unit, so there can be multiple types defined in the whole binary composed of multiple units. Do I save all of them (maybe they differ?) or only one? What about unnamed structures? Should I makeup name for them so I can reference them, maybe using an offset in the section would work, it's unique for in the binary?

thestr4ng3r:
For more complex C types like multidimensional arrays, I don't think this is something that we should make part of the "base type" database format directly so things that are saved in such a way right now should probably be refactored. As you already noticed, doing things like `struct.MyClass.buffer=char,9,100` for arrays breaks down as soon as you need more than just primitive arrays of a single type. Instead I would go back again to the strict separation between base types and c types from https://hackmd.io/@thestr4ng3r/H1gMW5-K8. So `struct.MyClass.buffer=char,9,100` conceptually becomes `struct.MyClass.buffer=char[100],9` and we can arbitrarily extend the c types structures to support everything we need.

HoundThe:
So right now I am missing:
* Dealing with encoding, I think we will need extra member to AttrValue to know which one is set, because some types can be one of many encodings, this will be probably complicated thing
* Extracting size and offset attributes, this is also dependant on ^^
* Saving a typedef
* Tests

I was thinking of some things like implementing some sort of associative cache based on offset, so we don't parse types repeatedly, but thanks to this parsing being quite cache friendly it's still quite fast.
I also think, because Union and Struct members are indistinguishable from each other, I was thinking to restructure the BaseType this way: and maybe extending the kinds for typedef and atomic kind? Something like a real base_type, int, long int etc. we can extract from the DWARF binary aswell
```c
typedef enum {
	R_ANAL_BASE_TYPE_KIND_STRUCT,
	R_ANAL_BASE_TYPE_KIND_UNION,
	R_ANAL_BASE_TYPE_KIND_ENUM,
	R_ANAL_BASE_TYPE_KIND_TYPEDEF,
	R_ANAL_BASE_TYPE_KIND_ATOMIC, // For real atomic base types
} RAnalBaseTypeKind;

typedef struct r_anal_base_type_t {
	RAnalBaseTypeKind kind;
	char *name;
	char *type; // Used by typedef
	ut64 size; // size of the whole type in bytes;
	union {
		RVector cases;
                RVector members;
	};
} RAnalBaseType;
```

HoundThe:
I added doxygen, storing and parsing of additional stuff and now I am just missing a tests and dealing with the encoding, but thats more dwarf related than type parsing.

What do you think? I guess it's kinda lot to review

HoundThe:
Added test case, will add more when there will be settled what are we going to actually store and how etc.

From the testing on various binaries I found TODO things to deal with:

* name collisions
* What if struct is only typedefed so it's unnamed
* What if only declaration
* Why are enum values stored in hex format?
* parsing templates?
* associative cache (hash map) we don't parse same shit all the time?

But overall I think its kinda solid? I'll let you decide

libr/anal/type.c
```diff
@@ -384,3 +385,302 @@ R_API RAnalBaseType *r_anal_get_base_type(RAnal *anal, const char *name) {
...
+	r_strbuf_fini (&arglist);
+	r_strbuf_fini (&param_key);
+	r_strbuf_fini (&param_val);
+}
+
+static void save_enum(const RAnal *anal, const RAnalBaseType *type) {
+	r_return_if_fail (anal && type && type->name 
+		&& type->kind == R_ANAL_BASE_TYPE_KIND_ENUM);
+	/*
+		name=enum
```
XVilka, line 512: For these comments, could you please add also the C syntax representation of the type?

HoundThe: Sure! Done

XVilka:
@thestr4ng3r union types should have a "priority" or "preferred" field that is set by default.

PR #17116, opened 2020-06-20
Fix debugee to inherit original environment variables ##debug

### LINKED ISSUE #16359: Radare2 shouldn't add envrionment variable when debugging
Hello,

During exploit development I have spend quite some time before realizing that radare2 was adding environment variables when opening a program with -d. This modify the stack and make it hard to develop local exploit that rely on environment variable for storing values and addresses.

### Work environment
[printenv.tar.gz](https://github.com/radareorg/radare2/files/4402023/printenv.tar.gz)

| OS/arch/bits (mandatory)                             | Ubuntu derivative (elementary OS 5.1.2 Hera) x86 64
| Architecture/bits of the file (mandatory)            | x86/32, x86/64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.4.0-git 10 @ linux-x86-64 git.4.4.0-git commit: 1bf0fed6cc753f26905e97ba903cbe6dc959950a build: 2020-03-30__11:30:21

### Expected behavior
No variable environment are added by r2 when debugging a program with -d.

### Actual behavior
r2 add the following environment variables:
```sh
PATH=/tmp/.local/share/radare2/prefix/bin:(null)
HOME=/
TMPDIR=
```

The following code print environment variables:
``` C
#include <stdio.h>
#include <stdlib.h>

// gcc -o printenv printenv.c

int main(int argc, char **argv, char **envp)
{
  for (char **env = envp; *env != 0; env++)
  {
    char *thisEnv = *env;
    printf("%s\n", thisEnv);    
  }
  return 0;
}
```

Running the program with a clean environment return no result:
``` bash
$ env -i ./printenv
```

Running it in r2 with debug mode add variable environment :
``` bash
$ env -i /usr/local/bin/r2 -qc dc -d ./printenv
Process with PID 23320 started...
= attach 23320 23320
bin.baddr 0x55ea2dfe4000
Using 0x55ea2dfe4000
asm.bits 64
Warning: r_bin_file_hash: file exceeds bin.hashlimit
PATH=/tmp/.local/share/radare2/prefix/bin:(null)
HOME=/
TMPDIR=
```

abcSup:

**Detailed description**

Duplicate and save the original `environ` and use it for `execvpe` later in `io_debug` so the child process will not inherit the `environ` of `r2`. 

**Test plan**
Run `env -i /usr/local/bin/r2 -c dc -d /usr/bin/env` and check for empty outputs. However, afaik there is no way to run command like that in the test suite. This test should be done in debug mode, so a custom binary is required to check for environment variables set by `r2` and one of which is in `PATH`. But, this way will make the test depends on the binary and any changes in future will require the binary to be replaced to make the test valid which is too troublesome. I would be glad to receive any suggestion on this.

**Closing issues**

closes #16359

### DIFF
--- a/libr/core/cmd_debug.c
+++ b/libr/core/cmd_debug.c
@@ -296,6 +296,8 @@ static const char *help_msg_do[] = {
 	"Usage:", "do", " # Debug (re)open commands",
 	"do", "", "Open process (reload, alias for 'oo')",
 	"dor", " [rarun2]", "Comma separated list of k=v rarun2 profile options (e dbg.profile)",
+	"doe", "", "Show rarun2 startup profile",
+	"doe!", "", "Edit rarun2 startup profile with $EDITOR",
 	"doo", " [args]", "Reopen in debug mode with args (alias for 'ood')",
 	"doof", " [args]", "Reopen in debug mode from file (alias for 'oodf')",
 	"doc", "", "Close debug session",
@@ -5382,6 +5384,26 @@ static int cmd_debug(void *data, const char *input) {
 		switch (input[1]) {
 		case '\0': // "do"
 			r_core_file_reopen (core, input[1] ? input + 2: NULL, 0, 1);
+			break;
+		case 'e': // "doe"
+			switch (input[2]) {
+			case '\0': // "doe"
+				if (core->io->envprofile) {
+					r_cons_println (core->io->envprofile);
+				}
+				break;
+			case '!': // "doe!"
+			{
+				char *out = r_core_editor (core, NULL, core->io->envprofile);
+				if (out) {
+					free (core->io->envprofile);
+					core->io->envprofile = out;
+					eprintf ("%s\n", core->io->envprofile);
+				}
+			} break;
+			default:
+				break;
+			}
 			break;
 		case 'r': // "dor" : rarun profile
 			if (input[2] == ' ') {
--- a/libr/io/p/io_debug.c
+++ b/libr/io/p/io_debug.c
@@ -267,6 +267,12 @@ static RRunProfile* _get_run_profile(RIO *io, int bits, char **argv) {
 		if (strstr (io->runprofile, R_SYS_DIR ".rarun2.")) {
 			(void)r_file_rm (io->runprofile);
 		}
+	} else if (io->envprofile) {
+		if (!r_run_parse (rp, io->envprofile)) {
+			eprintf ("Can't parse default rarun2 profile\n");
+			r_run_free (rp);
+			return NULL;
+		}
 	}
 	if (bits == 64) {
 		r_run_parseline (rp, expr=strdup ("bits=64"));
@@ -299,7 +305,6 @@ static void handle_posix_redirection(RRunProfile *rp, posix_spawn_file_actions_t
 
 // __UNIX__ (not windows)
 static int fork_and_ptraceme_for_mac(RIO *io, int bits, const char *cmd) {
-	bool runprofile = io->runprofile && *(io->runprofile);
 	pid_t p = -1;
 	char **argv;
 	posix_spawn_file_actions_t fileActions;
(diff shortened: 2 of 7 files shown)


abcSup:
> This pull request **introduces 1 alert** when merging [9114064](https://github.com/radareorg/radare2/commit/9114064bc5b2475a15cf9d256a9a55d5258a02c6) into [8540156](https://github.com/radareorg/radare2/commit/8540156f7397372968064157d7ad1473457cfa31) - [view on LGTM.com](https://lgtm.com/projects/g/radareorg/radare2/rev/pr-a717969132dddad6854aa970003f6fa3bbd6482e)
> 
> **new alerts:**
> 
> * 1 for Implicit function declaration

I found out that I should include the `#define _GNU_SOURCE` feature test to use `execvpe`. I just added that at the top of `io_debug.c`. Let me know if that works. 

Although the current `io_debug` does not support FreeBSD, I included a check for it because FreeBSD's equivalent of `execvpe` is `exect` to satisfy the FreeBSD build test.

libr/util/sys.c
```diff
@@ -1199,10 +1199,32 @@ R_API char **r_sys_get_environ (void) {
...
+
+	// save original environ for debugging
+	if (env) {
+		size_t count, i;
+
+		for (i = 0; env[i]; i++)
+			;
+		count = i;
+
+		origenv = malloc (sizeof (char*) * (count + 1));
```
trufae, line 1211: use calloc and avoid the multiplication. Also this global allocation is never free'd and this can be a problem as long as nothing ensures this is called only once and there's no way to deallocate it.

Also, env is writable. and its linearly allocated in memory, we can just do a single allocation for the whole buffer and then another allocation for the pointer storage. instead of having one strdup per-env. But this can result on some other problems we probably dont want to handle.

What are your thoughs on this @ret2libc ?

libr/util/sys.c
```diff
@@ -26,6 +26,7 @@
 #include <r_lib.h>
 
 static char** env = NULL;
+static char** origenv = NULL;
```
ret2libc, line 29: I don't particularly like this global `origenv`... even though there was already a global `env` as well :(

libr/util/sys.c
```diff
@@ -1199,10 +1199,32 @@ R_API char **r_sys_get_environ (void) {
 	if (!env) {
 		env = r_lib_dl_sym (NULL, "environ");
 	}
+
+	// save original environ for debugging
+	if (env) {
+		size_t count, i;
+
+		for (i = 0; env[i]; i++)
```
ret2libc, line 1207: just use `count` directly

libr/util/sys.c
ret2libc: Use `R_NEWS`

ret2libc: Although in theory what @trufae suggested is probably the best thing memory-wise, I would stick with this simpler solution for the beginning, unless we realize those additional strdup are really a problem.

libr/io/p/io_debug.c
```diff
@@ -1,5 +1,6 @@
 /* radare - LGPL - Copyright 2007-2019 - pancake */
 
+#define _GNU_SOURCE
```
ret2libc, line 3: I'm not sure we want to depend on this. @trufae ? This could break on some non-gnu platforms? I'm not sure what they are, but i think there are some we want to support.

trufae: i assume this is needed for `execvpe` which is not available on bsd or mac for example. isnt execve enough?

libr/util/sys.c
trufae: Me neither but i don't see any other option here

abcSup:
Thanks for the feedbacks! I will get back to this soon after finishing another PR for testing remote debugging.

trufae:
What about saving the startup environment as a rarun profile? and add a command to print and edit it with cfg.editor? this way we don't need to hold that environment differently than when the user specifies a rarun profile and will let us modify the execution environment easily for the next run

abcSup:
> What about saving the startup environment as a rarun profile? and add a command to print and edit it with cfg.editor? this way we don't need to hold that environment differently than when the user specifies a rarun profile and will let us modify the execution environment easily for the next run

That sounds better and cleaner. I will check out rarun profile and implement it with that then. Thanks!

trufae:
Great! I would love to see that!

XVilka:
@abcSup are there any updates on this PR? It would be awesome to include the fix in the imminent 4.5.0 release.

abcSup:
> @abcSup are there any updates on this PR? It would be awesome to include the fix in the imminent 4.5.0 release.

Ok, I will finish this now to be included in the new release. I will also try to test this on a Windows and Mac to make sure it work for most platforms.

abcSup:
**Changes**
- Retrieve original environment and save it into `io->envprofile` in rarun2 profile format
- In Linux & Mac, always spawn new process with rarun2 profile from either user-defined profile file `dbg.profile` or the default rarun2 profile in `io->envprofile`.
- Changed `env` command to display the default rarun2 profile saved in `io->envprofile`
- Added `env!` command to edit default rarun2 profile with `cfg.editor`

Let me know:
- If I should change the command `env` to something more intuitive (maybe something starts with `o` or `d`).
- If I should change `r2` to setup those environment variables on Mac & Windows (which is the current implementation).
  - On Windows, 7th argument is `NULL`
https://github.com/radareorg/radare2/blob/4b029876660c5cac00d79066e27e0539895d3946/libr/io/p/io_debug.c#L106-L108
  - On Mac, last argument is `NULL`
https://github.com/radareorg/radare2/blob/4b029876660c5cac00d79066e27e0539895d3946/libr/io/p/io_debug.c#L396

abcSup:
I keep the original `env` command and introduce two new commands `doe` and `doe!` to show and edit the rarun2 startup profile.

libr/core/cmd_debug.c
```diff
@@ -5383,6 +5385,24 @@ static int cmd_debug(void *data, const char *input) {
 		case '\0': // "do"
 			r_core_file_reopen (core, input[1] ? input + 2: NULL, 0, 1);
 			break;
+		case 'e': // "doe"
+			switch (input[2]) {
+			case '\0': // "doe"
+				r_cons_println (core->io->envprofile);
```
trufae, line 5391: envprofile can be null, add a chk

yossizap:
>If I should change r2 to setup those environment variables on Mac & Windows native debugger (the current implementation does not include environment variables when spawning a new process). 

That would be nice for windows but it can be done in a future PR since custom environment variables are very rarely used, the created process uses system default which is fine.

trufae:
@thestr4ng3r can you review this pr?

XVilka:
Can we merge this?

abcSup:
I just noticed and fixed a logic error in `r_run_parse`, waiting for review.

PR #17119, opened 2020-06-21
Optimize r_file_proc_size() reading by blocks instead of chars ##util

trufae:

**Detailed description**

Reading char by char is very slow

**Test plan**

I could write a unit test for that function, but it's an internal function and internal functions shuoldnt be tested according to http://shoulditestprivatemethods.com

**Closing issues**

None, its just a performance improvement for a very rare case


trufae:
Here's my test/benchmark program:

```c
#include <stdio.h>

int getmapsize() {
        char buf[1000];
        int fs = 0;
        FILE *fd = fopen ("/proc/self/maps", "rb");
        int i = 0;
#if 0
        // 5s
        long size = 0;
        while (fgetc (fd) != EOF) {
                fs++;
        }
#else
        do {
                write(1, "a", 1);
                fs += fread (NULL, 1, 1024, fd);
                write(1, "b", 1);
                // fs += fread (buf, 1, sizeof (buf), fd);
        } while (!feof (fd));
#if 0
        // 3s
        while (1) {
                int res = fread (buf, 1, sizeof (buf), fd);
                if (res < 1) break;
                fs += res;
        }
#endif
#endif
        fclose (fd);
        return fs;
}


main() {
        int i;
#if 0
        for (i = 0; i< 300000; i++) {
                getmapsize();
        }
#endif
        printf ("%d\n", getmapsize());
}
```

libr/util/file.c
```diff
@@ -181,11 +181,12 @@ R_API bool r_file_exists(const char *str) {
 	return S_IFREG == (S_IFREG & buf.st_mode);
 }
 
-R_API long r_file_proc_size(FILE *fd) {
+static inline long r_file_proc_size(FILE *fd) {
```
ret2libc, line 184: rename it to `file_proc_size`, because we use `r_` usually for exposed functions.

ret2libc:
Ooops, conflicting files. Anyway, LGTM. Fix conflicts and we merge

PR #17124, opened 2020-06-22
Use API to get file formats instead of calling rabin2 ##bin

### LINKED ISSUE #16958: r2 does not work with files containing space
### Expected behavior

r2 does work with files containing spaces

### Actual behavior

r2 does not work with files containing spaces

```
$ cp /bin/bash bash
$ cp /bin/bash "ba sh"
$ r2 -nn -Nqc "pfj.elf_header @ 0" bash
[{ ... }]
$r2 -nn -Nqc "pfj.elf_header @ 0" "ba sh"
Cannot find 'elf_header' format.
```

ret2libc:

**Detailed description**

~I'm not sure why the quotes were around the entire command. I've moved the quotes around the filename because it may contain spaces, as demonstrated in the related issue.~ Ok, now I know why it was wrapped. In the old parser, if the filename contains characters like `@` they would be interpreted as temporary seek and things would break. I have used `'` to wrap the filename and `"` to wrap the entire command. Both old and new parsers should work well.

**Closing issues**

Closes #16958


XVilka:
Doesn't work:
```
[XX] db/cmd/cmd_pf2 -nn filename with @
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -nn -Qc 'e io.va=0
pf.elf_header @ elf_header
' bins/elf/analysis/tiny1@invalid_addr
-- stdout
@@ -1,19 +0,0 @@
-     ident : 
-                struct<elf_ident>
-           magic : 0x00000000 = "\x7fELF"
-           class : 0x00000004 = class (enum elf_class) = 0x1 ; ELFCLASS32
-            data : 0x00000005 = data (enum elf_data) = 0x1 ; ELFDATA2LSB
-         version : 0x00000006 = version (enum elf_hdr_version) = 0x1 ; EV_CURRENT
-      type : 0x00000010 = type (enum elf_type) = 0x2 ; ET_EXEC
-   machine : 0x00000012 = machine (enum elf_machine) = 0x3 ; EM_386
-   version : 0x00000014 = version (enum elf_obj_version) = 0x1 ; EV_CURRENT
-     entry : 0x00000018 = 0x08048054
-     phoff : 0x0000001c = 0x00000034
-     shoff : 0x00000020 = 0x00000000
-     flags : 0x00000024 = 0x00000000
-    ehsize : 0x00000028 = 52
- phentsize : 0x0000002a = 32
-     phnum : 0x0000002c = 1
- shentsize : 0x0000002e = 0
-     shnum : 0x00000030 = 0
-  shstrndx : 0x00000032 = 0


-- stderr
@@ -0,0 +1,4 @@
+Invalid address (invalid_addr")
+|ERROR| Invalid command '.!rabin2 -rk. "bins/elf/analysis/tiny1@invalid_addr"' (0x2e)
+Invalid address (elf_header)
+|ERROR| Invalid command 'pf.elf_header @ elf_header' (0x70)
```

XVilka:
It fails on Windows: https://ci.appveyor.com/project/radareorg/radare2/builds/33661817/job/clv02hwrij9tg9u5#L3698

libr/core/cbin.c
```diff
@@ -171,7 +171,7 @@ R_API bool r_core_bin_load_structs(RCore *core, const char *file) {
 	}
 	// TODO use the RBin API, not cmdf()
 	// r_core_bin_export_info_rad (core);
-	r_core_cmdf (core, "\".!rabin2 -rk. %s\"", file);
+	r_core_cmdf (core, "\".!rabin2 -rk. '%s'\"", file);
```
trufae, line 174: '' doesnt work on windows. so you must escape the file with the `shell_escaper` and use "" quotes around the %s

ret2libc:
While using `r_core_bin_export_info_rad` to loads pf formats when using `-nn` the function was printing some error message, which by the way may be confusing. I think there is no clean way to avoid printing stderr messages on screen, so I just redirect temporarily stderr to /dev/null, but if there is a better way please speak up.

test/db/cmd/cmd_pf2
```diff
@@ -496,7 +496,6 @@ EXPECT=<<EOF
      shnum : 0x00000030 = 0
   shstrndx : 0x00000032 = 0
 EOF
-EXPECT_ERR=
```
ret2libc, line 499: FYI, this is because the `r_core_bin_export_info_rad` is generating an eprintf. This was always generated, however without this PR rabin2 is directly called and the stderr of rabin2 is not caught, so the eprintf was never actually printed. With this PR we directly use the right api function, however I could not find an easy way to ignore stderr (and probably we should not ignore it, but we should fix somehow the code in `r_core_bin_export_info_rad`.

That however goes beyond my current knowledge of the `pf` code and this PR anyway fixes the linked issue and it improves compared to master because it does not rely on calling another binary.

ret2libc:
Please review when you can

libr/core/cbin.c
```diff
@@ -169,10 +169,19 @@ R_API bool r_core_bin_load_structs(RCore *core, const char *file) {
...
-	r_core_cmdf (core, "\".!rabin2 -rk. %s\"", file);
-	return true;
+	RBinOptions opt = { 0 };
+	r_bin_open (core->bin, file, &opt);
+	RBinFile *bf = r_bin_cur (core->bin);
+	r_cons_push ();
+	r_core_bin_export_info_rad (core);
+	r_cons_filter ();
+	const char *s = r_cons_get_buffer ();
+	char *res = strdup (s? s: "");
```
trufae, line 179: if (!R_STR_ISEMPTY(s)) { ... instead of strdup ("")

test/db/cmd/cmd_pf2
trufae: ok

ret2libc:
> The proper fix will be to make `r_core_bin_export_info_rad` accept a "mode" argument to just run the r2 apis instead of printing the commands and then grabbing the console buffer, filter it and run the commands, it's innefficient and scary because it keeps running commands.

Sorry, that's  a bit too much for me. I just wanted to ensure pf file formats are loaded even when spaces are in the filename, as said in the linked issue. Rewriting the whole `r_core_bin_export_info_rad` is a bit too much and out of scope to me. I think the current solution, even though not perfect, does a good use of existing API and fixes the issue + possible command injection, thus it's definitely a good improvement over master.

> If you are not in the mood to go for this change i'll do it, feel free to merge the PR like this and open an issue to refactor RCore.bin_export() assigned to me.

I will, but I can't merge with your `red`. Please @XVilka or someone else either approve if you are ok with the current status of the PR or request changes and in such case I will leave the actual work to @trufae for better refactoring of the entire thing.

trufae:
My +1, please create an issue assigned to me to refactor this function

PR #17126, opened 2020-06-22
Bail early if path string is an empty string

### LINKED ISSUE #17100: Error opening empty string files
### Work environment

| OS/arch/bits (mandatory)                             | macOS 10.15
| Architecture/bits of the file (mandatory)            | PPC, x86/32, x86/64 etc.
| r2 -v full output, **not truncated** (mandatory)     | 50de58171e92f57b191a4f28cf3a0cdcf96e6e2e

### Expected behavior

**

### Actual behavior

```
**$ r2 ""
WARNING: r_io_def_mmap_check_default: assertion 'filename && *filename' failed (line 160)
WARNING: r_io_def_mmap_check_default: assertion 'filename && *filename' failed (line 160)
Missing file to open
```

other tools also fail in the same way:

```
$ rabin2 -s ""
WARNING: r_io_def_mmap_check_default: assertion 'filename && *filename' failed (line 160)
Couldn't open bin for file ''
r_bin: Cannot open file
WARNING: r_core_file_free: assertion 'cf' failed (line 964)
```

Open the shell and run the tools using "" as filename

nothing so far

phakeobj:

**Detailed description**

...

**Test plan**

...

**Closing issues**

Closes #17100

### DIFF
--- a/libr/main/rabin2.c
+++ b/libr/main/rabin2.c
@@ -895,7 +895,7 @@ R_API int r_main_rabin2(int argc, const char **argv) {
 	file = argv[opt.ind];
 
 	if (file && !*file) {
-		eprintf ("Couldn't open empty path\n");
+		eprintf ("Cannot open empty path\n");
 		r_core_fini (&core);
 		return 1;
 	}	
@@ -1008,6 +1008,7 @@ R_API int r_main_rabin2(int argc, const char **argv) {
 			r_core_fini (&core);
 			return 1;
 		}
+
 		eprintf ("FMT %s\n", format);
 		eprintf ("PKG %s\n", file);
 		for (i = opt.ind + 2; i < argc; i++) {
--- a/libr/main/rabin2.c
+++ b/libr/main/rabin2.c
@@ -894,7 +894,7 @@ R_API int r_main_rabin2(int argc, const char **argv) {
 	}
 	file = argv[opt.ind];
 
-	if (file && file[0] == '\0') {
+	if (file && !*file) {
 		eprintf ("Couldn't open empty path\n");
 		r_core_fini (&core);
 		return 1;
--- a/libr/main/rabin2.c
+++ b/libr/main/rabin2.c
@@ -896,6 +896,7 @@ R_API int r_main_rabin2(int argc, const char **argv) {
 
 	if (file && file[0] == '\0') {
 		eprintf ("Couldn't open empty path\n");
+		r_core_fini (&core);
 		return 1;
 	}	
 
--- a/libr/main/rabin2.c
+++ b/libr/main/rabin2.c
@@ -898,7 +898,7 @@ R_API int r_main_rabin2(int argc, const char **argv) {
 		eprintf ("Cannot open empty path\n");
 		r_core_fini (&core);
 		return 1;
-	}	
+	}
 
 	if (!query) {
 		if (action & R_BIN_REQ_HELP || action == R_BIN_REQ_UNK || !file) {
--- a/libr/main/rabin2.c
+++ b/libr/main/rabin2.c
@@ -893,6 +893,12 @@ R_API int r_main_rabin2(int argc, const char **argv) {
 		return 1;
 	}
 	file = argv[opt.ind];
+
+	if (file && file[0] == '\0') {
+		eprintf ("Couldn't open empty path\n");
+		return 1;
+	}	
+
 	if (!query) {
 		if (action & R_BIN_REQ_HELP || action == R_BIN_REQ_UNK || !file) {
 			r_core_fini (&core);
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -715,7 +715,7 @@ R_API int r_main_radare2(int argc, const char **argv) {
 		eprintf ("Cannot open empty path\n");
 		ret = 1;
 		goto beach;
-	}	
+	}
 
 	if (do_list_io_plugins) {
 		if (r_config_get_i (r->config, "cfg.plugins")) {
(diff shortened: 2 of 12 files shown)


libr/main/rabin2.c
```diff
@@ -893,6 +893,12 @@ R_API int r_main_rabin2(int argc, const char **argv) {
 		return 1;
 	}
 	file = argv[opt.ind];
+
+	if (file && file[0] == '\0') {
```
ret2libc, line 897: why && and not || ? I think it should be || and in that case use R_STR_ISEMPTY

phakeobj: because then it means that the if will handle the case where the user did not insert a file at all, where I would expect the program to print the print help and not complain that the path string is an empty string. the case of no file at all is handled in the if block right under my patch.

btw, I should probably call r_core_fini(&core) as in other places, before returning, noted.

phakeobj: or maybe this _should_ be treated as when the user passes no file at all?

ret2libc: Right, no then I guess it is fine as it is. And yes, you should probably call `r_core_fini` as you said. Thanks!

libr/main/rabin2.c
```diff
@@ -893,6 +893,13 @@ R_API int r_main_rabin2(int argc, const char **argv) {
...
```
trufae, line 897: Use R_STR_IS_EMPTY

libr/main/radare2.c
```diff
@@ -690,6 +690,13 @@ R_API int r_main_radare2(int argc, const char **argv) {
 			pfile = argv[opt.ind] ? strdup (argv[opt.ind]) : NULL;
 		}
 	}
+
+	if (pfile && pfile[0] == '\0') {
```
trufae, line 694: Same here

libr/main/rabin2.c
```diff
@@ -893,6 +893,13 @@ R_API int r_main_rabin2(int argc, const char **argv) {
...
+		eprintf ("Couldn't open empty path\n");
+		r_core_fini (&core);
```
trufae, line 899: Why return here and goto below? Would be better to have a single exit point

libr/main/rabin2.c
ret2libc: Please see discussion in https://github.com/radareorg/radare2/pull/17126/files#r443814345

libr/main/rabin2.c
ret2libc: The `goto` below is in another file though.

phakeobj: I was trying to follow existing conventions in `r_main_radare2` and `r_main_rabin2`

libr/main/radare2.c
trufae, line 694: I prefer !*pfile than pfile[0] == 0. Shorter snd concise. Can you change this in both places?

trufae:
More issues to be fixed:

* r2 -i '' -
* r2 -P '' -
* r2 -R ''
* radiff2 '' ''

i dont think im missing any other, but please give it a try and see if you can find another confusing error/assert

Thanks!

phakeobj:
I will check it for every argument that takes a file path, in every tool in the suite. `radiff2` behavior, as an example, seems to handle empty paths without triggering any assert,

```
 ~ $ radiff2 '' ''
radiff2: Cannot open
```

Do you think I should add a test for empty string and not let it try to `opencore` the empty path strings?

trufae:
Yes, the earlier we check for invalid input filenames the faster we can spit out an error message. But this "Cannot open" message could be the same as the other cases

phakeobj:
I'm not satisfied with 4dfc203 but that's just WIP. I changed "couldn't" to "cannot" as it seems like this is the preferred tense in these errors in r2 suite. I feel bad about the code duplication, and I wonder if you have any r2 style (marco-ish maybe?) idea of how to make it less disturbing.

I've added the bails in rafind2 and rahash2 as well, didn't cover all the toolset yet, just wanted to get some feedback regarding to the direction of this issue, if this is going to where you intended.

libr/main/radare2.c
```diff
@@ -592,6 +607,12 @@ R_API int r_main_radare2(int argc, const char **argv) {
 			r_config_set (r->config, "dbg.profile", opt.arg);
 			break;
 		case 'R':
+			file = opt.arg;
```
ret2libc, line 610: this looks wrong.
` -R [rr2rule] specify custom rarun2 directive`

`-R` is supposed to be used like `radare2 -R aslr=false -d /bin/ls`. It does not have a file as a argument. `-r` is probably what you wanted to check.

phakeobj: truth is that now I'm not even sure if `-r` should be handled. The other two are OK as they're validating the input before appending it to the pre/post files list. I can't remember why I added `-R`, I think it's a leftover I forgot to delete.

phakeobj: OK now I remember, and I think that the description, and the functionality of `-R` and `-r` is inverted. '-R' sets `customRarunProfile` which is later opened and appended to `dbg.profile` while, `-r` simply appends the given string to `dbg.profile`. Thoughts?

cc: @trufae

ret2libc: So, `-R` creates a list of rarun2 directives, then it creates a temporary file, dump those entries in the temporary file and load dbg.profile with the temporary file.
`-r` directly gets a file and sets dbg.profile to that temporary file. So you should check whether the argument of `-r` is not an empty file.

phakeobj: gotcha.

PR #17127, opened 2020-06-22
Update record & replay features to trace reg/mem changes ##debug

abcSup:

**Detailed description**

This PR rewrites most parts of the reverse-debugging functionality in `r2`. Instead of restoring back the memory maps to their initial states and running the debugger until the desired point, this implementation keeps track of registers and memory write changes and applies them when returning back to a point.

How it works:

`dts+` Start a reverse-debugging session. This is to save the current memory mappings and the initial value of all registers.

**Recording**

After `RDebugSession` is initialized, every time `r_debug_step` runs, the debugger will analyze the current instruction to extract registers and memory access information. I make use of [the operand access information provided by Capstone](https://www.capstone-engine.org/op_access.html) to obtain registers write changes.

However, Capstone only provides the memory operands information. For memory writes changes, we can keep track of them in a few scenarios:
1. When the destination operand is a memory location in x86 instructions, there is a memory write to that location.
2. When the instructions affect the stack pointer `push` `pushf` `pusha` `call`, there are memory writes in the stack.

We save all the write changes of the same register or memory location in a `RVector` together with a `cnum` value to keep track of when it happens during the session. Then, we store the `RVector` into either the register or the memory hashtable `Ht_up` to refer to later.

**Replay**

To step back n times, `RDebugSession` will restore the registers back to their initial values saved during the session initialization. Then, it will look up the latest registers changes before a point `cnum` and apply those changes. The same thing goes for memory by going through the whole hashtable for all memory changes and find their latest changes.

To continue back, it will look through the `RVector` of the PC register in backward to check if its value contains a breakpoint and jump that to point `cnum`.

**Added stuffs**
- a new `RList` of `RAnalValue` called `access` in `RAnalOp` to store all the access information because `dst` and `src[3]` are not enough.
- two `Ht_up` hashtables for registers and memory in `RDebugSession` to store those `RVector` of changes
- `RDebugChangeReg` `RDebugChangeMem` structs
- `r_debug_trace_ins_before` to resolve memory address in all `RAnalValue` based on the current registers.
- `r_debug_trace_ins_after` functions to obtain all the write changes made to registers and memory locations
- Support continue and continue back in two modes, which can be set using `dbg.trace_continue`:
  - Trace every instruction when doing continue (continuously stepping until user interrupt or breakpoints)
  - Add a checkpoint at where the continue stops. User can travel between checkpoints so the debugger does not need to record all the changes executed in between two points.

**Removed stuffs**
- `RPageData`
- `RDebugSnapDiff`
- Most of the functions in `libr/debug/snap.c`

**Room for improvement**
- Update `RReg` `r_reg_get_value` to support register of size bigger than 64 bits
- Store read access of registers and memory for other analysis
- ~~Support tracing in continue. The current implementation only tracks changes when stepping.~~

**Test plan**

Unit testing is a little complicated for this because `RDebug` alone is not enough to test the features and requires initialization of `RCore`. 

Therefore, I am thinking of making `cmd` tests to compare registers/memory states outputs. For example: 
1. Start a session `dts+`
2. Step into a particular point and output the registers/memory states `ds` `dr` `dtsm`
3. Step for n times `ds`
4. Step back into the previous point and output the registers/memory states `dsb` `dr` `dtsm`
5. Compare both results


libr/debug/p/native/linux/linux_debug.c
```diff
@@ -111,14 +111,17 @@ int linux_handle_signals (RDebug *dbg, int tid) {
...
+							break;
 						} else if (r_str_startswith (p, "dbg.unlibs")) {
 							dbg->reason.type = R_DEBUG_REASON_EXIT_LIB;
+							break;
 						}
 					}
 				}
 			}
-			if (dbg->reason.type != R_DEBUG_REASON_NEW_LIB &&
-				dbg->reason.type != R_DEBUG_REASON_EXIT_LIB) {
```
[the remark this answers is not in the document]
abcSup: I think it is my mistake. I will move the if-else into
```
if (dbg->reason.type != R_DEBUG_REASON_NEW_LIB &&
	dbg->reason.type != R_DEBUG_REASON_EXIT_LIB) {
}
```

abcSup:
Currently, there is a bug that causes random segfault when doing `dc` and interrupting repeatedly.
```
[0x00000000]> dc                                                                                               
                                                                                                                             
Thread 1 "r2" hit Breakpoint 1, r_debug_continue (dbg=0x5555557ee8c0) at debug.c:1258                    
1258            return r_debug_continue_kill (dbg, 0); //dbg->reason.signum);                                                     
(gdb)                                                                                                 
Continuing.                                                                                                                    
trace_before: 524                                                                            
native-singlestep: Success                                                                              
Stepping failed!                                                                                     
Error: stepping                                                                                           
                                                                                                        
Thread 1 "r2" received signal SIGSEGV, Segmentation fault.                                                                         
0x00007ffff7306536 in __GI__IO_default_xsputn (f=0x7fffffffa1c0, data=<optimized out>, n=3) at genops.c:412                         
412     genops.c: No such file or directory.                                                                                    
```

There is also a bug that triggers stack smashing detection which I believe is also related.

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -386,8 +391,8 @@ RDebugReasonType linux_dbg_wait(RDebug *dbg, int my_pid) {
 	// Ignore keyboard interrupt while waiting to avoid signaling the child process twice on
 	// break(SIGSTOP is sent by wait_break) which forced the user to continue an extra time
 	// to handle SIGINT
-	r_sys_signal (SIGINT, SIG_IGN);
-	r_cons_break_push ((RConsBreak)linux_dbg_wait_break, dbg);
+	// r_sys_signal (SIGINT, SIG_IGN);
```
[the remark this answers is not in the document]
abcSup: Anyway, I believe using `r_sys_signal` here is still wrong, especially at the end it calls `r_sys_signal (SIGINT, SIG_DFL)` which restore the default action when receiving `SIGINT` back to terminating the process. Therefore, interrupting `r2` with ^C next time will exit it instead of calling the callback in `r_cons_breaked_stack`. `r_cons` already set the default action for `SIGINT` to be `__break_signal` in `r_cons.c`, so other code should not change that behavior.

In addition, `linux_dbg_wait_break` callback does not always work in situations when the debugee is in a stopped state. For example, interrupting `r2` will call `linux_dbg_wait_break` which sends the `SIGSTOP` to the debugee, but in our case we are calling `r_debug_step` repeatedly so the signal will not be delivered to the process. Then, the next time we continue the debugee, it will stop immediately because of the pending signal.

libr/debug/debug.c
```diff
@@ -1145,6 +1097,19 @@ R_API int r_debug_continue_kill(RDebug *dbg, int sig) {
 	if (!dbg) {
 		return 0;
 	}
+
+	if (dbg->session) {
```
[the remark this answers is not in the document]
abcSup: Do you mean toggling between a->b or repeatedly stepping mode?

libr/io/p/io_debug.c
```diff
@@ -211,6 +211,7 @@ static void trace_me(void) {
 	ptrace (PT_SIGEXC, getpid(), NULL, 0);
 #endif
 #else
+	setsid ();
```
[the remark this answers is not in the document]
abcSup: You are right. For now, the debugee `stdout` still output to the console if that is what you mean by console redirection. I am still testing this out, but I believe this can be removed because I do not see any effect of fixing the bug with this.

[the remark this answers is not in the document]
yossizap: >For now, the debugee stdout still output to the console if that is what you mean by console redirection
I meant redirecting data into the program's stdin with Cutter

libr/debug/debug.c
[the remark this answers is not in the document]
yossizap: Yes

libr/debug/p/native/linux/linux_debug.c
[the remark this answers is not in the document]
abcSup: This will be fixed in another PR.

libr/io/p/io_debug.c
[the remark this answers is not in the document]
abcSup: I will remove this because I think this is no longer necessary.

libr/debug/debug.c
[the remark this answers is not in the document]
abcSup: It will be `dbg.trace_continue` and it is set to true as default.

shlr/ptrace-wrap/src/ptrace_wrap.c
```diff
@@ -78,6 +79,11 @@ static void wrap_func(ptrace_wrap_instance *inst) {
 }
 
 static void *th_run(ptrace_wrap_instance *inst) {
+	sigset_t set;
+	sigemptyset (&set);
+	sigaddset (&set, SIGINT);
```
trufae, line 84: uhm not sure if we want a ^C handler here

libr/include/r_vector.h
```diff
@@ -129,6 +129,10 @@ R_API void *r_vector_shrink(RVector *vec);
 	if (!r_vector_empty (vec)) \
 		for (it = (void *)(vec)->a; (char *)it != (char *)(vec)->a + ((vec)->len * (vec)->elem_size); it = (void *)((char *)it + (vec)->elem_size))
 
+#define r_vector_foreach_prev(vec, it) \
```
trufae, line 132: @thestr4ng3r can u review this?

libr/debug/trace.c
```diff
@@ -43,6 +43,114 @@ R_API int r_debug_trace_tag (RDebug *dbg, int tag) {
...
+				r_list_delete (dbg->cur_op->access, it);
+			}
+		default:
+			break;
+		}
+	}
+	return true;
+}
+
+R_API int r_debug_trace_ins_after(RDebug *dbg) {
```
trufae, line 110: return bool or void

libr/debug/trace.c
```diff
@@ -43,6 +43,114 @@ R_API int r_debug_trace_tag (RDebug *dbg, int tag) {
 	return (dbg->trace->tag = (tag>0)? tag: UT32_MAX);
 }
 
+R_API int r_debug_trace_ins_before(RDebug *dbg) {
```
trufae, line 46: return bool or void

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -1172,7 +1177,12 @@ int linux_reg_write (RDebug *dbg, int type, const ut8 *buf, int size) {
 			size = sizeof (R_DEBUG_REG_T);
 		}
 #endif
-		return (ret != 0) ? false : true;
+		if (ret == -1) {
+			r_sys_perror ("reg_write");
+			return false;
+		} else {
```
trufae, line 1183: dont else after return

libr/debug/debug.c
```diff
@@ -1145,6 +1105,45 @@ R_API int r_debug_continue_kill(RDebug *dbg, int sig) {
...
+
+			r_debug_goto_cnum (dbg, dbg->session->maxcnum);
+			return dbg->tid;
+		}
+
+		if (dbg->trace_continue) {
+			// Continuously stepping to trace each instruction until user interrupt ^C
+			r_cons_break_push (NULL, NULL);
+			while (!r_cons_is_breaked ()) {
+				r_debug_step (dbg, 1);
```
yossizap, line 1137: You should also check r_debug_step's return value, it returns the number of steps done on success.

libr/debug/debug.c
```diff
@@ -1145,6 +1105,45 @@ R_API int r_debug_continue_kill(RDebug *dbg, int sig) {
...
+				if (dbg->session->reasontype != R_DEBUG_REASON_STEP) {
```
yossizap, line 1138: While testing I have attempted to `dc` to main in trace mode and I kept being stopped on the way there because reasontype was R_DEBUG_REASON_BREAKPOINT (while not stepping on any breakpoint on the PC).
This should be looked into. Maybe the session section should be implemented within continue's "repeat" loop below and use it's existing logic to determine whether `dc` is finished instead of re-implementing it.

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -386,8 +391,8 @@ RDebugReasonType linux_dbg_wait(RDebug *dbg, int my_pid) {
 	// Ignore keyboard interrupt while waiting to avoid signaling the child process twice on
 	// break(SIGSTOP is sent by wait_break) which forced the user to continue an extra time
 	// to handle SIGINT
-	r_sys_signal (SIGINT, SIG_IGN);
-	r_cons_break_push ((RConsBreak)linux_dbg_wait_break, dbg);
```
yossizap, line 390: These wrappers were only replaced in continue+trace, keep in mind you have to implement it for stepping(even though it's a single step you may interrupt a system call, right now it breaks debug) and regular continue.
Also, I am still not convinced it had to be removed from here. And you will need to remove it from window/freebsd/macos's debug loop if you intend to have the interrupt wrap in debug.c.

libr/debug/debug.c
abcSup: I have encountered this which is caused by syscall trap because PTRACE_O_TRACESYSGOOD is on by default when attached. You are right and thanks for the suggestion. I will them move into the existing loop.

libr/debug/p/native/linux/linux_debug.c
abcSup: Right, this is not a complete fix to the real problem and the complete solution is in the multithreaded debugging PR.

libr/debug/debug.c
abcSup: Then, user has to set `e dbg.bpsysign=true` to ignore those system breakpoints which is already implemented in the loop.

libr/debug/p/native/linux/linux_debug.c
abcSup: > And you will need to remove it from window/freebsd/macos's debug loop if you intend to have the interrupt wrap in debug.c.
The interrupt wrap already exists in `cmd_debug.c` and the callback is `static_debug_stop`. However, `r_sys_signal (SIGINT, SIG_IGN);` prevents that from working.

shlr/ptrace-wrap/src/ptrace_wrap.c
abcSup: This is no longer necessary and will be removed.

libr/debug/p/native/linux/linux_debug.c
abcSup: Several commands have wrap the function with `static_debug_stop` to break the debugger out.
https://github.com/radareorg/radare2/blob/261783336c1114e796f8ec6b74149140f73bab10/libr/core/cmd_debug.c#L5170-L5175

abcSup: I have put them back and use `linux_dbg_wait_break` to send the `SIGINT` instead of ignoring and passing the `SIGINT` to the debuggee

abcSup:
> Lots of changes in here, can you add a checkbox list in the PR to understand whatas been done and what's left to not be WIP?
Most of the changes are to fix obvious errors and most of the functionality are already implemented. Thanks for the review!

XVilka:
Can you please either rebase or merge master? So BSD test will be green.

PR #17129, opened 2020-06-23
Draft of DEB packaging

XVilka:
Addressing https://github.com/radareorg/radare2-release/issues/34 and https://github.com/radareorg/radare2-release/issues/35

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -1,0 +1,54 @@
+name: Packaging release assets
+
+on:
+  push:
+    branches:
+    - 'pre-release*'
+    - 'prerelease*'
+    - 'release-*'
+
+jobs:
+  build-deb:
+    name: Build DEB package
+    runs-on: ubuntu-latest
+    container: debian:jessie
+    steps:
+    - name: Install tools
+      run: apt-get update && apt-get install --yes patch unzip git gcc make curl pkg-config
+    - name: Checkout r2
+      run: |
+        git clone https://github.com/${{ github.repository }}
+        cd radare2
+        git fetch origin ${{ github.ref }}
+        git checkout -b local_branch FETCH_HEAD
+    - name: Making the static build
+      run: sys/static.sh
+      working-directory: radare2
+    - name: Cleaning up the unnecessary files
+      run: git clean -xdf
+      working-directory: radare2
+    - name: Preparing the Debian package
+      run: sys/debian.sh
+      working-directory: radare2
+    - name: Create Release
+      id: create_release
+      uses: actions/create-release@v1
+      env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/packaging.yml
```diff
@@ -0,0 +1,54 @@
+name: Packaging release assets
+
+on:
+  push:
```
ret2libc, line 4: I think we will also need to use `on.release` soon, though I need to check exactly how it works. 

When we choose to start a release we should probably branch with `prerelease-X.Y.Z` and then start all the additional tests, check integration with cutter, r2dec, extras, etc. At that point, we should fix possible issues in master and cherry-pick the commits in the `prerelease-X.Y.Z` branch as well (which will run again all the tests and integration with other repositories). Once all integration tests are fixed, we can create `release-X.Y.Z` branch.

.github/workflows/packaging.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: Making the static build
+      run: sys/static.sh
+      working-directory: radare2
+    - name: Cleaning up the unnecessary files
+      run: git clean -xdf
+      working-directory: radare2
+    - name: Preparing the Debian package
+      run: sys/debian.sh
+      working-directory: radare2
+    - name: Create Release
```
ret2libc, line 33: Do you know what happens when we will have multiple jobs that do "Create release"? I'm thinking about jobs for uploading cydia package, android apk, etc. Will it generate multiple releases or re-use the existing one?

ret2libc: Ops sorry, ignore this comment.

.github/workflows/packaging.yml
```diff
@@ -0,0 +1,54 @@
...
+
+on:
+  push:
+    branches:
+    - 'pre-release*'
+    - 'prerelease*'
+    - 'release-*'
+
+jobs:
+  build-deb:
```
ret2libc, line 11: I think there should be multiple jobs. `build-dep` should be one, then we should have a `create-release` job and a `publish-assets` job that depend on `build-dep`. As done right now, this `build-dep` also create the release and publish the deb package. I think soon we will also have other jobs like `build-android`, `build-cydia`, etc. At that point, each of those jobs will need to be run on a different container.

trufae:
though having debs using my scripts and such was rejected by all you and PPA was the way to go

ret2libc:
> though having debs using my scripts and such was rejected by all you and PPA was the way to go

I still think DEB built by us in this way are not that useful as probably debian maintainers are going to re-create them anyway, but we can't change everything in one shot. I think moving things from bash scripts to gh actions with whatever we currently have is a good first step. This is probably more about a draft for moving release stuff to gh action than strictly about the deb package in itself.

XVilka:
It wasn't rejected - it already builds but with Travis. Also Debian is just
a part of the upcoming release automation.

trufae:
Amazing

ret2libc:
I'm closing this as i'm already working on a more complete solution! I'm going to do a PR when some basic things are done. Thanks!

PR #17148, opened 2020-06-25
rasign2 refactoring ##signatures

swoops:

**Detailed description**

This is a refactor of some sign code. It will result in a speed increase on signature creation. It will also help me to expand the `zb` command soon.

Currently, `zaf` on a single function will result in around 8 `RSignItem`'s being create, saved and then freed. Saving each individually causes a name collision which requires reallocating old items and merging.

The `r_sign_addto_item` function was created so a single `RSignitem` can easily have signature types added to it. The `addItem` function was renamed and added to the `r_sign` API. Now the `zaf` command will only allocate a single `RSignitem`, signatures are added to the item with `r_sign_addto_item`. Since only one item is saved there are no collisions under normal circumstances. 

I also updated the code for `zb` to use `r_sign_addto_item` which saves some lines of code. The interface for `r_sign_find_closest_sig` was updated to be more friendly.

As a result of this update `rasign2 libc.so` completes 1 second faster on my computer.

**Test plan**

The commands `zaf`, `zaF` and `zb` are running on new code but should function identical to before. All current `r2r` tests for zignatures pass. 

Feel free to suggest better variable and function names. I am bad at naming things :( 
Also, I am not completely sure how `r_cons_break` works, so it maybe should be moved around in the `cmd_zign.c:bestmatch` function.

**Closing issues**

None...

### DIFF
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -177,6 +177,14 @@ static RList *zign_types_to_list(RAnal *a, char *types) {
 	return ret;
 }
 
+static void bytes_sig_free(RSignBytes *bytes) {
+	if (bytes) {
+		free (bytes->bytes);
+		free (bytes->mask);
+		free (bytes);
+	}
+}
+
 R_API bool r_sign_deserialize(RAnal *a, RSignItem *it, const char *k, const char *v) {
 	char *refs = NULL;
 	char *vars = NULL;
@@ -325,7 +333,7 @@ R_API bool r_sign_deserialize(RAnal *a, RSignItem *it, const char *k, const char
 			// allocate
 			size = atoi (token);
 			if (size > 0) {
-				free (it->bytes);
+				bytes_sig_free (it->bytes);
 				it->bytes = R_NEW0 (RSignBytes);
 				if (!it->bytes) {
 					goto out;
@@ -517,11 +525,7 @@ static void mergeItem(RSignItem *dst, RSignItem *src) {
 	char *ref, *var, *type;
 
 	if (src->bytes) {
-		if (dst->bytes) {
-			free (dst->bytes->bytes);
-			free (dst->bytes->mask);
-			free (dst->bytes);
-		}
+		bytes_sig_free (dst->bytes);
 		dst->bytes = R_NEW0 (RSignBytes);
 		if (!dst->bytes) {
 			return;
@@ -530,14 +534,13 @@ static void mergeItem(RSignItem *dst, RSignItem *src) {
 		dst->bytes->size = src->bytes->size;
 		dst->bytes->bytes = malloc (src->bytes->size);
 		if (!dst->bytes->bytes) {
-			free (dst->bytes);
+			bytes_sig_free (dst->bytes);
 			return;
 		}
 		memcpy (dst->bytes->bytes, src->bytes->bytes, src->bytes->size);
 		dst->bytes->mask = malloc (src->bytes->size);
 		if (!dst->bytes->mask) {
-			free (dst->bytes->bytes);
-			free (dst->bytes);
+			bytes_sig_free (dst->bytes);
 			return;
 		}
 		memcpy (dst->bytes->mask, src->bytes->mask, src->bytes->size);
@@ -604,7 +607,7 @@ static void mergeItem(RSignItem *dst, RSignItem *src) {
 	}
 }
 
-static bool addItem(RAnal *a, RSignItem *it) {
+R_API bool r_sign_anal_additem(RAnal *a, RSignItem *it) {
 	char key[R_SIGN_KEY_MAXSZ], val[R_SIGN_VAL_MAXSZ];
 	const char *curval = NULL;
 	bool retval = true;
@@ -654,7 +657,7 @@ static bool addHash(RAnal *a, const char *name, int type, const char *val) {
 	switch (type) {
 	case R_SIGN_BBHASH:
 		it->hash->bbhash = strdup (val);
-		retval = addItem (a, it);
+		retval = r_sign_anal_additem (a, it);
 		r_sign_item_free (it);
 		break;
 	}
@@ -672,19 +675,11 @@ static bool addBBHash(RAnal *a, RAnalFunction *fcn, const char *name) {
 	if (!it->name) {
 		goto beach;
 	}
-	it->hash = R_NEW0 (RSignHash);
 	it->space = r_spaces_current (&a->zign_spaces);
-	if (!it->hash) {
-		goto beach;
-	}
 
-	char *digest_hex = r_sign_calc_bbhash (a, fcn);
-	if (!digest_hex) {
-		free (digest_hex);
-		goto beach;
+	if (r_sign_addto_item (a, it, fcn, R_SIGN_BBHASH)) {
+		retval = r_sign_anal_additem (a, it);
 	}
-	it->hash->bbhash = digest_hex;
-	retval = addItem (a, it);
 beach:
 	r_sign_item_free (it);
 	return retval;
@@ -724,16 +719,13 @@ static bool addBytes(RAnal *a, const char *name, ut64 size, const ut8 *bytes, co
 		goto fail;
 	}
 	memcpy (it->bytes->mask, mask, size);
-	retval = addItem (a, it);
+	retval = r_sign_anal_additem (a, it);
 	r_sign_item_free (it);
 	return retval;
 fail:
 	if (it) {
 		free (it->name);
-		if (it->bytes) {
-			free (it->bytes->bytes);
-			free (it->bytes);
-		}
+		bytes_sig_free (it->bytes);
 	}
 	free (it);
 	return false;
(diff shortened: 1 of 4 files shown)


libr/anal/sign.c
```diff
@@ -774,6 +766,146 @@ R_API bool r_sign_add_anal(RAnal *a, const char *name, ut64 size, const ut8 *byt
...
+	}
+	hash->bbhash = digest_hex;
+	return hash;
+}
+
+R_API bool r_sign_addto_item(RAnal *a, RSignItem *it, RAnalFunction *fcn, RSignType type) {
+	r_return_val_if_fail (a && it && fcn, false);
+	switch (type) {
+	case R_SIGN_GRAPH:
+		if (!it->graph) {
```
trufae, line 843: use && and save 2 lines on each case statement

trufae: actually you can do this in one line like this:

```c
return !it->graph && it->graph = r_sign_fcn_graph (fcn);
```

trufae:
>> As a result of this update rasign2 libc.so completes 1 second faster on my computer.

How long does it take to rasign libc? 2s? or 10s? it is good to know how much % of the total work this second means

libr/anal/sign.c
```diff
@@ -774,6 +766,146 @@ R_API bool r_sign_add_anal(RAnal *a, const char *name, ut64 size, const ut8 *byt
...
+	ut8 *bytes = malloc (size);
+	if (!bytes) {
+		return NULL;
+	}
+	if (!r_io_is_valid_offset (core->io, fcn->addr, 0)) {
+		eprintf ("error: cannot read at 0x%08" PFMT64x "\n", fcn->addr);
+		free (bytes);
+		return NULL;
+	}
+	(void)r_io_read_at (core->io, fcn->addr, bytes, size);
```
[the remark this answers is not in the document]
swoops: I saw a comment in `r_io_read_at` that says it is deprecated. So I will also switch to using `r_io_nread_at`.

swoops:
Master takes about 11 seconds
```
$ time rasign2 test/bins/elf/libc.so.6
...
real	0m11.278s
user	0m11.197s
sys	0m0.063s
```

This branch takes ~10:
```
$ time rasign2 test/bins/elf/libc.so.6
...
real	0m10.227s
user	0m10.129s
sys	0m0.081s
```

Most of the time is spent in analysis. Just timing the `zg` command alone (diff of two  `gettimeofday` calls) I get:
```
master: 1.863645 s
refactor: .747524 s
```

swoops:
Sorry, last commit was a guess at trying to fix the Meson build. I don't have Meson and I don't see any of the build errors on my Linux/gcc system. I will have to figure out how I can go about getting those errors so I don't waste your time.

In the mean time, if you have any suggestions I am happy to hear them.

ret2libc:
> Sorry, last commit was a guess at trying to fix the Meson build. I don't have Meson and I don't see any of the build errors on my Linux/gcc system. I will have to figure out how I can go about getting those errors so I don't waste your time.
> 
> In the mean time, if you have any suggestions I am happy to hear them.

I am not sure why the acr/make build is not failing, but it should, as RAnal does not depend on RIO (which provides the r_io functions). RAnal should use `anal->iob` to read things and in general to access functions that are provided by RIO, not the `r_io_*` functions directly. Hope it helps :)

libr/anal/sign.c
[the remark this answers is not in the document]
trufae: actually r_io_read_at should be replaced with r_io_nread_at, iirc nread returns the amount of bytes read, and r_io_read_at does the same but returns bool depending if it can read all the bytes or none.

So we can make RIO.read_at call RIO.nread_at and save some lines in r_io..

* r_io_read_at : 251 refs
* r_io_nread_at: 6 refs

So it's clear that in most places in r2 it's checking for a complete read on pa/va depending on io.va, this probably requires further discussion with @condret and others to decide the way to go. I see useful behaviours in both apis, not sure if it should be deprecated or just overriden by nread

libr/anal/sign.c
```diff
@@ -774,6 +767,109 @@ R_API bool r_sign_add_anal(RAnal *a, const char *name, ut64 size, const ut8 *byt
...
+		graph->bbsum = r_anal_function_realsize (fcn);
+	}
+	return graph;
+}
+
+static RSignBytes *r_sign_fcn_bytes(RAnal *a, RAnalFunction *fcn) {
+	r_return_val_if_fail (a && fcn, false);
+	RCore *core = a->coreb.core;
+	int maxsz = a->coreb.cfggeti (core, "zign.maxsz");
+	int fcnlen = r_anal_function_realsize (fcn);
```
trufae, line 786: this only works with linear functions. it must iterate over all the basic blocks instead

libr/anal/sign.c
```diff
@@ -774,6 +767,109 @@ R_API bool r_sign_add_anal(RAnal *a, const char *name, ut64 size, const ut8 *byt
...
+		eprintf ("error: Negative size for bytes signature\n");
+		return NULL;
+	}
+
+	// get bytes part of signature
+	ut8 *bytes = malloc (size);
+	if (!bytes) {
+		return NULL;
+	}
+	if (r_io_nread_at (core->io, fcn->addr, bytes, size) < 0) {
```
trufae, line 800: Use a->iob.read_at () instead

trufae, line 800: check must be == size or take the result and use it to avoid analyzing invalid data

libr/anal/sign.c
```diff
@@ -774,6 +767,109 @@ R_API bool r_sign_add_anal(RAnal *a, const char *name, ut64 size, const ut8 *byt
...
+		free (bytes);
+		return NULL;
+	}
+
+	// get mask part of signature
+	ut8 *mask = r_anal_mask (a, size, bytes, fcn->addr);
+
+	// get bytes sig holder
+	RSignBytes *sig = R_NEW0 (RSignBytes);
+	if (!mask) {
```
trufae, line 811: move this check one line above, othewise its leaking `sig`

libr/anal/sign.c
```diff
@@ -1,5 +1,6 @@
 /* radare - LGPL - Copyright 2009-2020 - pancake, nibble */
 
+#include <r_io.h>
```
trufae, line 3: remove this line. RAnal, doesn't depend on RIO

libr/anal/sign.c
swoops: I meant that to be `sig`. If `sig` is NULL then `sig->size = size;` will segfault. Mask should probably be checked too...

To simplify cleanup and save some lines of code I created `bytes_sig_free`. I then use that function in a few other places. This is done in commit a959981f5.

swoops:
I think I fixed the linear function size problem. I am not so sure about `R_MIN (core->io->addrbytes * fcnlen, maxsz);` still being correct? I stole that code from what was previously there.

Non-linear functions raise another problem. The `bytes` and `mask` part of the signature both appear to just be read linearly from the function start. I have an idea of how to fix this but I probably should leave that for another PR. Do you want me to make an issue for this?

libr/anal/sign.c
```diff
@@ -774,6 +766,120 @@ R_API bool r_sign_add_anal(RAnal *a, const char *name, ut64 size, const ut8 *byt
...
+		}
+	}
+
+	int size = R_MIN (core->io->addrbytes * fcnlen, maxsz);
+
+	RSignBytes *sig = R_NEW0 (RSignBytes);
+	if (!sig) {
+		goto bytes_failed;
+	}
+	sig->size = size;
```
trufae, line 803: no this is not correct either, because you are counting bbs that are not lineart and then reading linearly. i think the way to go for this would be to create a map with a mask for all the linear bytes of the function starting from the entry. so if a function starts with a jmpback it wont be detected but if there's data in the middle of the code it will be a valid match. can you redo this to follow this logic?

libr/anal/sign.c
```diff
@@ -774,6 +766,120 @@ R_API bool r_sign_add_anal(RAnal *a, const char *name, ut64 size, const ut8 *byt
...
+
+	// hopefully not necissary, but just in case
+	if (size <= 0) {
```
trufae, line 806: this check can be moved above, right before the R_NEW0

libr/anal/sign.c
swoops: Sorry, I misunderstood earlier. To make sure I got it now I made a simple example.
```
_jumpback:
    inc rbx
    jmp _next

 ; not part of function
_ignoreme:
    inc r10

;; entry point of function
_entry:           ;<-- start matching signature
    inc rax       ;   |                        
    jmp _jumpback ;   |                              
                  ;   |             
_next:            ;   |                   
    inc rdx       ;   |                        
    jmp _end      ;<--+ end matching signature                       


;; hole in function
_junk:
    inc rbp

_end:
    ret
```

I can correct the size to implement this logic.
Let me know if I am still misunderstanding and tell me which blocks in the above example should be included.

swoops:
I went ahead and updated the size according to my comment above. Let me know if you want the logic changed at all.

This update causes one regression test to fail because the regression test is wrong.
```
$ r2r db/cmd/cmd_zignature 
Running from /home/bemodtwz/ctf/tools/radare2/test
Loaded 48 tests.
[XX] radare2/test/db/cmd/cmd_zignature zj producing valid types
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'aaa
zaf fcn.00484d40 fcn.00484d40
zj
' bins/elf/static-glibc-2.27
-- stdout
@@ -1 +1 @@
...
```

Function  `fcn.00484d40` contains a 3 byte hole  at  `0x00484d72` for  instruction `nop dword [rax]`. The old logic kept these 3 bytes in the signature and missed the last 3 bytes of the function. You can see the regression signature lacks a `ret` (0xc3) at the end.

The current logic will stop creating the signature just before the hole. So the signature won't contain the hole or the instructions on the other side of the hole. I could update this easily to allow small holes. Maybe as long as they are smaller then the max instruction size? This would mean, in the case of `fcn.00484d40` here, the signature would fill to the end, including the last basic block.

With a bit more work I could zero the bytes of the holes in the signature mask. This might be overkill for alignment though. 

What do you think? What would you prefer?

XVilka:
> So the signature won't contain the hole or the instructions on the other side of the hole. I could update this easily to allow small holes. Maybe as long as they are smaller then the max instruction size? This would mean, in the case of fcn.00484d40 here, the signature would fill to the end, including the last basic block.

I think it makes sense. @ret2libc @trufae what do you think?

trufae:
I dont think it is correct. this 3 byte desviation is a random case. But in other binaries like MIPS for example.. the amount of data inside a function can be pretty large, even more on switch tables or other obfuscated binaries. So, imho the way to go is the following:

```go
for fun in functions {
  sorted_blocks := fun.blocks.sort(.by_offset)
  ea := fun.entry
  last_bb := sorted_blocks[-1]
  funbuf := malloc(lastbb.addr-ea)
  for bb in sorted_bblocks {
    if bb.addr > ea {
      delta := bb.addr - ea
      memset (mask + delta, 0xff, bb.len)
      core.io.read_at (bb.addr, funbuf + delta, bb.size)
    } else {
      eprintf("Warning: skipping")
    }
  }
}

then you have the byte array with the binary mask to apply in the signatures for searching. and that will work in all the situations except for the functions that start jumping backward before the entrypoint of the function
```

swoops:
I agree with @trufae. I did some tests and including the gaps in the functions produces better results. The only problem with allowing the gaps is a change in gap length will cause a failed signature match. I have an idea for how to handle this case, but it is best left for another commit, if at all.

I updated my code to reflect the proposed algorithm.  Please review.

I updated the `zj` test so that it would be correct. I don't understand why it is failing. In the code block below I copped just the bytes out of the `-` and `+` entries of the failed test, as I see it on github. I also copied the bytes out of my the  #ae6b50c commit and labeled that `*`.
```
- 55534883ec08488b1f4885db741b4889fd4889dfe857b7f7ff4801d8803800751748c74500000000004883c4084889d85b5dc30f1f440000c600004883c001488945004883c40848
+ 55534883ec08488b1f4885db741b4889fd4889dfe857b7f7ff4801d8803800751748c74500000000004883c4084889d85b5dc3
* 55534883ec08488b1f4885db741b4889fd4889dfe857b7f7ff4801d8803800751748c74500000000004883c4084889d85b5dc30000000000c600004883c001488945004883c4084889d85b5dc3 
```
None of the byte signatures match... I don't understand. If I should fix something let me know.

swoops:
Sorry, I made a small change that was bugging me. Please take a peek at #4b1f354 @trufae.  The change saves a call to `memset` and reduces the number of calls to `a->iob.read_at`. As a result of the change, the bytes part of the signature will show all the original bytes, instead of having zeros in the function gaps. The gaps will still be masked out. I think this is desirable for someone auditing the signature.

> Amazing! Good work!

Thank you! And thanks for all the help.

> Yeah i agree that we should have more tests, but the current state of this PR is quite better than in master. Imho it's fine to merge this but some stuff is left:
> 
>     * add more tests
I have some plans for tests already, it shouldn't be too hard to add. I probably will get to it later today.

>     * update PR title to match the reality (or split the code in commits to be rebased on master with refactor+improve bytemask signatures, etc..)

If this is something I need to do on my end, then I might need some help. My git skills are pretty weak.

swoops:
I rebased to master and squashed everything into two commits with better names. Let me know if this is sufficient or if you want me to do something else?

Thanks for being patient with me.

XVilka:
So far looks good, but could you please add a couple more tests for other architectures? e.g. ARM7, ARMv8, MIPS, SPARC, PowerPC?

PR #17149, opened 2020-06-25
Add command to load all info for lib ##debug

GustavoLCR:

**Detailed description**

* Add 'dmia'
* Accept more print modes in 'dmi'
* Don't call rabin2 to load info

### DIFF
--- a/libr/core/cmd_debug.c
+++ b/libr/core/cmd_debug.c
@@ -251,7 +251,7 @@ static const char *help_msg_dm[] = {
 static const char *help_msg_dmi[] = {
 	"Usage: dmi", "", " # List/Load Symbols",
 	"dmi", "[j|q|*] [libname] [symname]", "List symbols of target lib",
-	"dmia", "[j|q|*] [libname] [symname]", "List all info of target lib",
+	"dmia", "[j|q|*] [libname]", "List all info of target lib",
 	"dmi*", "", "List symbols of target lib in radare commands",
 	"dmi.", "", "List closest symbol to the current address",
 	"dmiv", "", "Show address of given symbol for given lib",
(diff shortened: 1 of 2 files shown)


libr/core/cmd_debug.c
```diff
@@ -1571,6 +1572,34 @@ static int r_debug_heap(RCore *core, const char *input) {
 	return true;
 }
 
+static bool get_bin_info(RCore *core, const char *file, ut64 baseaddr, int mode, bool symbols_only, const char *symname) {
+	int fd;
+	if ((fd = r_io_fd_open (core->io, file, R_PERM_R, 0)) != -1) {
```
ret2libc, line 1577: just do early return with
```C
if ((fd = r_io_fd_open (....) == -1) {
   return false;
}
```
so all the rest of the code doesn't need to be indented.

libr/core/cmd_debug.c
```diff
@@ -1571,6 +1572,34 @@ static int r_debug_heap(RCore *core, const char *input) {
...
+		opt.sz = r_io_fd_size (core->io, fd);
+		opt.baseaddr = baseaddr;
+		RBinFile *obf = r_bin_cur (core->bin);
+		if (r_bin_open_io (core->bin, &opt)) {
+			int action = R_CORE_BIN_ACC_ALL & ~R_CORE_BIN_ACC_INFO;
+			if (symbols_only || symname) {
+				action = R_CORE_BIN_ACC_SYMBOLS;
+			}
+			if (symname) {
+				r_cons_grep (symname);
```
ret2libc, line 1589: is there any test for this? In other code I'm using `r_cons_grep_process`, though I'm not sure how they are different. Also, shall you use `r_cons_push` + `r_cons_pop`?

libr/core/cmd_debug.c
```diff
@@ -250,7 +250,8 @@ static const char *help_msg_dm[] = {
 
 static const char *help_msg_dmi[] = {
 	"Usage: dmi", "", " # List/Load Symbols",
-	"dmi", "[libname] [symname]", "List symbols of target lib",
+	"dmi", "[j|q|*] [libname] [symname]", "List symbols of target lib",
+	"dmia", "[j|q|*] [libname] [symname]", "List all info of target lib",
```
ret2libc, line 254: I tried the patch and it looks good, however I'm not sure why `dmia` should accept a `symname` as well. Isn't that just a duplicate of `dmi`? I would remove the `symname` argument from `dmia` and let it just print all info. If a user wants to reduce the output he can anyway use `~`.

ret2libc:
@Xvilla could you restart bsd ci?

PR #17154, opened 2020-06-27
Optimize and cleanup the dex_get_proto function ##bin

trufae:

**Detailed description**

reading function arguments is causing many r_io_read calls for methods with > 1 arg. this can be simplified to make the code more readable as well as use strbuf instead of manual pointer mangling.

There's almost no real speedup gain, but the code is more optimal.

**Test plan**

i added a test

**Closing issues**
noine

### DIFF
--- a/libr/bin/p/bin_dex.c
+++ b/libr/bin/p/bin_dex.c
@@ -293,10 +293,11 @@ static char *dex_get_proto(RBinDexObj *bin, int proto_id) {
 		ut16 type_idx = r_read_le16 (typeidx_buf + off);
 		ut16 type_desc_id = type_desc (bin, type_idx);
 		if (type_desc_id == UT16_MAX) {
-			break;
+			r_strbuf_append (sig, "?;");
+		} else {
+			const char *buff = getstr (bin, type_desc_id);
+			r_strbuf_append (sig, buff? buff: "?;");
 		}
-		const char *buff = getstr (bin, type_desc_id);
-		r_strbuf_append (sig, buff? buff: "?");
 	}
 	r_strbuf_appendf (sig, ")%s", return_type);
 	free (typeidx_buf);
--- a/libr/bin/p/bin_dex.c
+++ b/libr/bin/p/bin_dex.c
@@ -275,32 +275,31 @@ static char *dex_get_proto(RBinDexObj *bin, int proto_id) {
 	}
 	// size of the list, in 16 bit entries
 	ut32 list_size = r_read_le32 (params_buf);
-	if (list_size * sizeof (ut16) >= bin->size) {
-		return NULL;
-	}
-	size_t typeidx_bufsize = (list_size * 2);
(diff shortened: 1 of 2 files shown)


libr/bin/p/bin_dex.c
```diff
@@ -271,48 +273,34 @@ static char *dex_get_proto(RBinDexObj *bin, int proto_id) {
...
-		buff_len = strlen (buff);
-		size += buff_len + 1;
-		char *newsig = realloc (signature, size);
-		if (!newsig) {
-			eprintf ("Cannot realloc to %d\n", size);
-			free (signature);
-			break;
+	// size of the list, in 16 bit entries
+	ut32 list_size = r_read_le32 (params_buf);
+	size_t typeidx_bufsize = (list_size * sizeof (ut16));
```
ret2libc, line 278: no need for parentheses

ret2libc: why `size_t`? list_size is ut32 and `params_off`, used in the next line, is `ut32` as well. Maybe we could use `ut64` if we want to be sure the result of the multiplication does not overflow.

trufae: ut32 * 2 can result in integer overflow. Also typeidx_bufsize is later used to specify the size of the allocation, on 64bit systems size_t is 64bit wide, on 32bit it may fail.

One solution to this is to check for the overflow or clamp the list_size to avoid overflowing the 32bit boundary. Also, i doubt any dex is that large to hold 4 million types when the method number limit is 65,536. I may check dvm's implementation to be more precise, but i don't think it actually matters that much on real life use cases.

So options are:

* use ut64
* avoid the overflow and chop the size or fail early

by checking the google implementation:

```
128         int returnIdx = file.getTypeIds().indexOf(prototype.getReturnType());
129         int paramsOff = OffsettedItem.getAbsoluteOffsetOr0(parameterTypes);
```

it's using `int` 🤦  so I would say the trimming with warning is the best option here. But i'm already checking against file->size. So that may be enough probably.

libr/bin/p/bin_dex.c
```diff
@@ -271,48 +273,38 @@ static char *dex_get_proto(RBinDexObj *bin, int proto_id) {
...
-		buff_len = strlen (buff);
-		size += buff_len + 1;
-		char *newsig = realloc (signature, size);
-		if (!newsig) {
-			eprintf ("Cannot realloc to %d\n", size);
-			free (signature);
-			break;
+	// size of the list, in 16 bit entries
+	ut32 list_size = r_read_le32 (params_buf);
+	if (list_size >= ST32_MAX) {
```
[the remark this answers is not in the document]
ret2libc: Ops, I didn't commit the review apparnetly :/ never mind.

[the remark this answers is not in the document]
trufae: ive opened a pr to improve the add/sub/mul overflow/underflow macros. there are more places where this issue can be problematic and we may address it in a generic way.

PR #17158, opened 2020-06-27
Initial implementation of io.fd plugin ##io

trufae:
Useful for self:// and r2frida when injecting the lib inside a target process

* [x] Handle ^C
* [x] Honor sandbox
* [x] Build in meson
* [x] Support Windows

 

**Detailed description**

This plugin was in the issues for a very long time, i saw it when it was pinged by the stalebot, so it was easy to implement

**Test plan**

Uhm maybe opening a file and assuming we get the right fd from the OS to manipualte it in live.. but there are no tests for libr2preload or self:// or for r2frida

**Closing issues**

Fixes https://github.com/radareorg/ideas/issues/116

### DIFF
--- a/libr/io/meson.build
+++ b/libr/io/meson.build
@@ -9,6 +9,7 @@ r_io_sources = [
   'undo.c',
   'p_cache.c',
   'p/io_ar.c',
+  'p/io_fd.c',
   'p/io_bfdbg.c',
   'p/io_bochs.c',
   'p/io_debug.c',
--- a/libr/io/p/io_fd.c
+++ b/libr/io/p/io_fd.c
@@ -7,7 +7,11 @@
 #include <r_cons.h>
 #include <sys/types.h>
 
+#if __WINDOWS__
+#define FDURI "handle://"
+#else
 #define FDURI "fd://"
+#endif
 
 typedef struct {
 	int fd;
@@ -67,7 +71,10 @@ static RIODesc *__open(RIO *io, const char *pathname, int rw, int mode) {
 	}
 	RIOFdata *fdd = R_NEW0 (RIOFdata);
 	if (fdd) {
-		fdd->fd = r_num_math (NULL, pathname + 5);
+		fdd->fd = r_num_math (NULL, pathname + strlen (FDURI));
+#if __WINDOWS__
+		fdd->fd = _open_osfhandle (fdd->fd, 0);
+#endif
 		if (fdd->fd < 0) {
 			free (fdd);
 			eprintf ("Invalid filedescriptor.\n");
@@ -78,10 +85,15 @@ static RIODesc *__open(RIO *io, const char *pathname, int rw, int mode) {
 }
 
 RIOPlugin r_io_plugin_fd = {
+#if __WINDOWS__
+	.name = "handle",
+	.desc = "Local process file handle IO",
+#else
 	.name = "fd",
 	.desc = "Local process filedescriptor IO",
+#endif
 	.uris = FDURI,
-	.license = "LGPL3",
+	.license = "MIT",
 	.open = __open,
 	.close = __close,
 	.read = __read,
--- a/libr/io/p/io_fd.c
+++ b/libr/io/p/io_fd.c
@@ -4,6 +4,7 @@
 #include "r_lib.h"
 #include <stdio.h>
 #include <stdlib.h>
+#include <r_cons.h>
 #include <sys/types.h>
 
 typedef struct {
@@ -30,7 +31,10 @@ static bool __resize(RIO *io, RIODesc *desc, ut64 count) {
 static int __read(RIO *io, RIODesc *desc, ut8 *buf, int count) {
 	RIOMalloc *mal = (RIOMalloc*)desc->data;
 	if (mal) {
-		return read (mal->fd, buf, count);
+		r_cons_break_push (NULL, NULL);
+		int res = read (mal->fd, buf, count);
+		r_cons_break_pop ();
+		return res;
 	}
 	return -1;
 }
@@ -54,6 +58,10 @@ static bool __check(RIO *io, const char *pathname, bool many) {
 
 static RIODesc *__open(RIO *io, const char *pathname, int rw, int mode) {
 	if (__check (io, pathname, 0)) {
+		if (r_sandbox_enable (false)) {
+			eprintf ("Do not permit fd:// in sandbox mode.\n");
+			return NULL;
+		}
 		RIOMalloc *mal = R_NEW0 (RIOMalloc);
 		if (mal) {
 			mal->fd = r_num_math (NULL, pathname + 4);
--- a/libr/io/p/io_fd.c
+++ b/libr/io/p/io_fd.c
@@ -58,23 +58,23 @@ static bool __check(RIO *io, const char *pathname, bool many) {
 }
 
 static RIODesc *__open(RIO *io, const char *pathname, int rw, int mode) {
-	if (__check (io, pathname, 0)) {
-		if (r_sandbox_enable (false)) {
-			eprintf ("Do not permit " FDURI " in sandbox mode.\n");
+	if (r_sandbox_enable (false)) {
+		eprintf ("Do not permit " FDURI " in sandbox mode.\n");
+		return NULL;
+	}
+	if (!__check (io, pathname, 0)) {
+		return NULL;
+	}
+	RIOFdata *fdd = R_NEW0 (RIOFdata);
+	if (fdd) {
+		fdd->fd = r_num_math (NULL, pathname + 5);
+		if (fdd->fd < 0) {
+			free (fdd);
+			eprintf ("Invalid filedescriptor.\n");
 			return NULL;
 		}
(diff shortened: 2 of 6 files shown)


XVilka:
I am not sure if it will work on Windows since there are file handles instead of descriptors.

trufae:
@GustavoLCR what do you think about supporting fd:// in Windows? another useful feature is that we could implement fd listing in fd:// (without any actual number). so without having to use the \dd from self://

GustavoLCR:
> @GustavoLCR what do you think about supporting fd:// in Windows? another useful feature is that we could implement fd listing in fd:// (without any actual number). so without having to use the \dd from self://

@trufae I don't think theres a easy way to get a list of C runtime fd on windows, even inside the same process, so fd:// would have to get `HANDLE` values and use [_open_osfhandle](https://docs.microsoft.com/pt-br/cpp/c-runtime-library/reference/open-osfhandle?view=vs-2019) on `__open` to translate it to C fds so that you can use` write`, `read` etc

trufae:
@ret2libc correct

@GustavoLCR so maybe it's better to just make fd:// work on non-windows platforms like other unix-specific io plugins?

trufae:
@ret2libc how do you want this information to be added? as a comment in the source? documentation? help message?

ret2libc:
> @ret2libc how do you want this information to be added? as a comment in the source? documentation? help message?

As we have no "long help" for all these plugins, I would say a small comment at the top of the file about how this plugin is expected to be used would be useful.

trufae:
Yes, we do have a long help, but not for all plugins, try: `r2 frida://?`

libr/io/p/io_fd.c
```diff
@@ -0,0 +1,99 @@
...
+static RIODesc *__open(RIO *io, const char *pathname, int rw, int mode) {
+	if (__check (io, pathname, 0)) {
+		if (r_sandbox_enable (false)) {
+			eprintf ("Do not permit fd:// in sandbox mode.\n");
+			return NULL;
+		}
+		RIOMalloc *mal = R_NEW0 (RIOMalloc);
+		if (mal) {
+			mal->fd = r_num_math (NULL, pathname + 4);
+			if (((int)mal->fd) < 0) {
```
ret2libc, line 68: no need for this cast

libr/io/p/io_fd.c
```diff
@@ -0,0 +1,99 @@
...
+	}
+	return 0;
+}
+
+static bool __check(RIO *io, const char *pathname, bool many) {
+	return !strncmp (pathname, "fd://", 4);
+}
+
+static RIODesc *__open(RIO *io, const char *pathname, int rw, int mode) {
+	if (__check (io, pathname, 0)) {
```
ret2libc, line 60: just check the epposite and do an early return. Reading nested things is harder.

libr/io/p/io_fd.c
```diff
@@ -0,0 +1,99 @@
...
+static ut64 __lseek(RIO* io, RIODesc *desc, ut64 offset, int whence) {
+	RIOMalloc *mal = (RIOMalloc*)desc->data;
+	if (mal) {
+		return lseek (mal->fd, offset, whence);
+	}
+	return 0;
+}
+
+static bool __check(RIO *io, const char *pathname, bool many) {
+	return !strncmp (pathname, "fd://", 4);
```
ret2libc, line 56: define the prefix string somewhere and use it here
```C
#define FD_PREFIX "fd://"

return !strncmp (pathname, FD_PREFIX, strlen (FD_PREFIX));
```
It is clearer and less error prone than having to count bytes manually (4 vs 5)

libr/io/meson.build
```diff
@@ -9,6 +9,7 @@ r_io_sources = [
   'undo.c',
   'p_cache.c',
   'p/io_ar.c',
+  'p/io_fd.c',
```
ret2libc, line 12: this is not enough at the moment to enable a plugin in meson. Add it to the `io_plugins` array in `libr/meson.build`.

libr/io/p/io_fd.c
```diff
@@ -0,0 +1,99 @@
...
+
+static RIODesc *__open(RIO *io, const char *pathname, int rw, int mode) {
+	if (__check (io, pathname, 0)) {
+		if (r_sandbox_enable (false)) {
+			eprintf ("Do not permit fd:// in sandbox mode.\n");
+			return NULL;
+		}
+		RIOMalloc *mal = R_NEW0 (RIOMalloc);
+		if (mal) {
+			mal->fd = r_num_math (NULL, pathname + 4);
```
ret2libc, line 67: this is wrong as well, it is +5. BTW, could you add a test for this plugin? As it is right now I don't think it can work, as `r_num_math` would always return 0.

libr/io/p/io_fd.c
```diff
@@ -0,0 +1,99 @@
...
+	if (mal) {
+		return ftruncate (mal->fd, count) == 0;
+	}
+	return false;
+}
+
+static int __read(RIO *io, RIODesc *desc, ut8 *buf, int count) {
+	RIOMalloc *mal = (RIOMalloc*)desc->data;
+	if (mal) {
+		r_cons_break_push (NULL, NULL);
```
ret2libc, line 34: Stupid question: why is `cons_break_push`/`cons_break_pop` needed? What do they do?

libr/io/p/io_fd.c
ret2libc: use `pathname + strlen(FD_PREFIX)`

libr/io/p/io_fd.c
```diff
@@ -0,0 +1,99 @@
...
+		int res = read (mal->fd, buf, count);
+		r_cons_break_pop ();
+		return res;
+	}
+	return -1;
+}
+
+static int __close(RIODesc *desc) {
+	// dont close, could be problematic in self://
```
ret2libc, line 43: I think we should at least free the RIODesc allocated in `__open`

libr/io/meson.build
trufae: Thanks for the tip! i missed that

libr/io/p/io_fd.c
trufae: because read of a random filedescriptor of the current running process may result on lock states, let's say, reading from stdin, this is needed in order to handle ^C

libr/io/p/io_fd.c
trufae: this is done by r_io_close, but by the plugin callback. im freeing the ->data inside now

libr/io/p/io_fd.c
trufae: indeed. it was a stupid mistake :)

libr/io/p/io_fd.c
trufae: done

trufae:
Thanks for the review @ret2libc i think the only remaining part is now to make this plugin __UNIX__ specific like it's done in shm://, is windows not going to work with this? i mean, open/read/close is available in there, it's not the winapi, but can be supported too i think, enowindows here, so i cant test. i would appreciate a decision on this from the windows devs here in r2land

GustavoLCR:
@trufae something like this would make more sense on windows:
```diff
diff --git a/libr/io/p/io_fd.c b/libr/io/p/io_fd.c
index 35dd19e18..93622201b 100644
--- a/libr/io/p/io_fd.c
+++ b/libr/io/p/io_fd.c
@@ -7,7 +7,11 @@
 #include <r_cons.h>
 #include <sys/types.h>

+#if __WINDOWS__
+#define FDURI "handle://"
+#else
 #define FDURI "fd://"
+#endif

 typedef struct {
        int fd;
@@ -66,6 +70,9 @@ static RIODesc *__open(RIO *io, const char *pathname, int rw, int mode) {
                RIOFdata *fdd = R_NEW0 (RIOFdata);
                if (fdd) {
                        fdd->fd = r_num_math (NULL, pathname + 5);
+#if __WINDOWS__
+                       fdd->fd = _open_osfhandle (fdd->fd, 0);
+#endif
                        if (fdd->fd < 0) {
                                free (fdd);
                                eprintf ("Invalid filedescriptor.\n");
```

trufae:
Is this function honoring the original handle permissions if 0 is passed as arg? I’ll make it green, rebase and apply your patch later today. Thanks! Didnt knew this windows api

libr/io/p/io_fd.c
```diff
@@ -0,0 +1,100 @@
...
+	if (r_sandbox_enable (false)) {
+		eprintf ("Do not permit " FDURI " in sandbox mode.\n");
+		return NULL;
+	}
+	if (!__check (io, pathname, 0)) {
+		return NULL;
+	}
+	RIOFdata *fdd = R_NEW0 (RIOFdata);
+	if (fdd) {
+		fdd->fd = r_num_math (NULL, pathname + 5);
```
ret2libc, line 70: Please use stolen(fduri) instead of hardcoding 5

trufae:
All comments addressed again. Thanks for the reviews!

PR #17160, opened 2020-06-27
Fix two shift overflow causing undefined behaviour

trufae:
Spotted in clusterfuzz-testcase-minimized-ia_fuzz-6301506113634304

**Detailed description**

**Test plan**
```c
$ cat shift.c



main() {
	unsigned long long a=0x12389;
	int i = 0;
	while (1) {
		a <<= i;
		i++;
	}
}

$ gcc -fsanitize=undefined shift.c
$ ./a.out
shift.c:8:5: runtime error: shift exponent 64 is too large for 64-bit type 'unsigned long long'
^C

```

**Closing issues**

two from coverity

### DIFF
--- a/libr/util/buf.c
+++ b/libr/util/buf.c
@@ -652,8 +652,12 @@ R_API st64 r_buf_uleb128(RBuffer *b, ut64 *v) {
 			return -1;
 		}
 		c = data & 0xff;
-		sum |= ((ut64) (c & 0x7f) << s);
-		s += 7;
+		if (s < 64) {
+			sum |= ((ut64) (c & 0x7f) << s);
+			s += 7;
+		} else {
+			sum = 0;
+		}
 		l++;
 	} while (c & 0x80);
 	if (v) {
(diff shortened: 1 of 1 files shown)


libr/util/buf.c
```diff
@@ -652,8 +652,12 @@ R_API st64 r_buf_uleb128(RBuffer *b, ut64 *v) {
 			return -1;
 		}
 		c = data & 0xff;
-		sum |= ((ut64) (c & 0x7f) << s);
-		s += 7;
+		if (s < 64) {
+			sum |= ((ut64) (c & 0x7f) << s);
+			s += 7;
+		} else {
+			sum = 0;
```
ret2libc, line 659: shall the function return as well? WDYT?

trufae: or spit out a warning?, maybe we need to return error somehow as 3rd argument here so the caller can catch those issues properly?

ret2libc: it seems right now errors are returned as -1. We could keep this behaviour maybe.

radare: -1 is fine for uleb, but not for sleb, imho it's conceptually wrong to mix data and error channels

ret2libc: yeah :(

Issue #6741, opened 2017-02-08
Seemingly inconsistent disassembly of a Thumb instruction


robryk:
I'm not entirely sure that this is a bug, but at the very least I don't understand the disassembly, so making it less confusing might be helpful.

The instruction in question is a pc-relative load.

The address in square brackets in the disassembly doesn't agree with the address in the comment:

```
$ r2 -
[0x00000000]> wb 054b
[0x00000000]> e asm.arch=arm
[0x00000000]> e asm.bits=16
[0x00000000]> pd
            0x00000000      054b           ldr r3, [0x00000018]        ; [0x18:4]=0x4b054b05
            0x00000002      054b           ldr r3, [0x0000001a]        ; [0x18:4]=0x4b054b05
            0x00000004      054b           ldr r3, [0x0000001c]        ; [0x1c:4]=0x4b054b05
            0x00000006      054b           ldr r3, [0x0000001e]        ; [0x1c:4]=0x4b054b05
            0x00000008      054b           ldr r3, [0x00000020]        ; [0x20:4]=0x4b054b05
            0x0000000a      054b           ldr r3, [0x00000022]        ; [0x20:4]=0x4b054b05
            0x0000000c      054b           ldr r3, [0x00000024]        ; [0x24:4]=0x4b054b05 ; '$'
            0x0000000e      054b           ldr r3, [0x00000026]        ; [0x24:4]=0x4b054b05 ; '$'
            0x00000010      054b           ldr r3, [0x00000028]        ; [0x28:4]=0x4b054b05 ; '('
            0x00000012      054b           ldr r3, [0x0000002a]        ; [0x28:4]=0x4b054b05 ; '('
            0x00000014      054b           ldr r3, [0x0000002c]        ; [0x2c:4]=0x4b054b05 ; ','
            0x00000016      054b           ldr r3, [0x0000002e]        ; [0x2c:4]=0x4b054b05 ; ','
```

Note that the address in the disassembled instructions always increases by 2, while the offset shown in the comment is rounded down to a multiple of 4.

I'm not entirely sure which is correct.

radare:
this is expected and correct. ldr r3 [0x18] is doing a read of 4 byets at 0x18. this is presented as [0x18:4], and it shows that the value in this address of memory is 0x4b054b05.

robryk:
Look at the second instruction in the disassembly. The instruction as disassembled is:

```
ldr r3, [0x1a]
```

while the comment says:

```
[...]

PR #17161, opened 2020-06-27
Add test for #6741 - to verify thumb load instruction analysis

trufae:

**Detailed description**

Added test for this thumb issue that is now fixed

**Test plan**

theres a test and an issue describing it in detail

**Closing issues**

yep the #6741


trufae:
Yes, that's how memory access alignment works in thumb mode, instructions can be 2 or 4 bytes (in thumb2 mode) but memory access is always bound to 4, so the instructions aligned to 2 use the address - 2 as reference

PR #17162, opened 2020-06-27
Fix #6467 - Implement opn/opp/opr commands to priorize next/previous/rotate file ##io

### LINKED ISSUE #6467: Add 'o' subcommands to open next/prev file, count opened files, etc

trufae:

**Detailed description**

See the #6467 where it describes these commands 

**Test plan**

see the test

**Closing issues**

Fixes #6467

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -30,7 +30,7 @@ static const char *help_msg_o[] = {
 	"om","[?]","create, list, remove IO maps",
 	"on"," [file] 0x4000","map raw file at 0x4000 (no r_bin involved)",
 	"oo","[?+bcdnm]","reopen current file (see oo?) (reload in rw or debugger)",
-	"op"," [fd]", "select the given fd as current file (see also ob)",
+	"op","[r|n|p|fd]", "select priorized file by fd (see ob), opn/opp/opr = next/previous/rotate",
 	"oq","","list all open files",
 	"ox", " fd fdx", "exchange the descs of fd and fdx and keep the mapping",
 	NULL
@@ -41,6 +41,15 @@ static const char *help_msg_o_[] = {
 	"o-*","","close all opened files",
 	"o-!","","close all files except the current one",
 	"o-3","","close fd=3",
+	NULL
+};
+
+static const char *help_msg_op[] = {
+	"Usage:", "op[rnp] [fd]", "",
+	"opr", "", "open next file rotating",
+	"opn", "", "open next file",
+	"opp", "", "open previous file",
+	"op", " [fd]", "open priorizing fd",
 	NULL
 };
 
@@ -1209,6 +1218,76 @@ static bool desc_list_json_cb(void *user, void *data, ut32 id) {
 	return true;
 }
 
+typedef struct {
+	int fd; // input and output
+	int inc;
+	int prevfd;
+	bool picknext;
+	bool found;
+	RCore *core;
+} RCmdOpData;
+
+static bool try_next(void *_user, void *_data, ut32 id) {
+	RCmdOpData *data = _user;
+	if (data->picknext) {
+		data->fd = id;
+		goto pickme;
+	}
+	if (id == data->fd) {
+		if (data->inc > 0) {
+			// pick next
+			data->picknext = true;
+		} else if (data->inc < 0) {
+			// pick prevfd
+			data->fd = data->prevfd;
+			goto pickme;
+		}
+	}
+	data->prevfd = id;
+	return true;
+pickme:
+	data->picknext = false;
+	data->found = true;
+	return false;
+	
+}
+
+static bool cmd_op(RCore *core, char mode, int fd) {
+	int cur_fd = -2;
+	if (fd != -1) {
+		cur_fd = fd;
+	} else {
+		if (core->io && core->io->desc) {
+			cur_fd = r_io_fd_get_current (core->io);
+		}
+	}
+	const int inc = (mode == 'n')? 1: -1;
+
+	if (cur_fd != -2) {
+		RCmdOpData data = {
+			.fd = cur_fd,
+			.inc = inc,
+			.core = core,
+			.picknext = mode == 0
+		};
+		r_id_storage_foreach (core->io->files, try_next, &data);
+		if (data.found) {
+			int fdn = data.fd;
+			int myfd = r_io_fd_get_current (core->io);
+			if (fdn >= 0 && fdn != myfd && r_io_use_fd (core->io, fdn)) {
+				RBinFile *bf = r_bin_file_find_by_fd (core->bin, fdn);
+				if (bf && r_core_bin_raise (core, bf->id)) {
+					r_core_block_read (core);
+					return true;
+				} else {
+					eprintf ("Invalid RBinFile.id number.\n");
+				}
+			}
+		}
+	}
+	return false;
+}
+
 static int cmd_open(void *data, const char *input) {
 	RCore *core = (RCore*)data;
 	int perms = R_PERM_R;
@@ -1366,20 +1445,38 @@ static int cmd_open(void *data, const char *input) {
 	case 'p': // "op"
 		/* handle prioritize */
 		if (input[1]) {
-			int fd = r_num_math (core->num, input + 1);
-			if (fd >= 0 || input[1] == '0') {
-				RIODesc *desc = r_io_desc_get (core->io, fd);
-				if (desc) {
-					// only useful for io.va=0
-					// load bininfo for given fd
-					r_core_cmdf (core, "obo %d", fd);
-					core->io->desc = desc; // XXX we should use fd here, not *pointer
-					r_core_block_read (core);
-				} else {
-					eprintf ("Cannot find RBinFile associated with fd %d\n", fd);
+			switch (input[1]) {
+			case 'r': // "opr" - open next file + rotate if not found
+				if (!cmd_op (core, 'n', -1)) {
+					if (!cmd_op (core, 0, -1)) {
+						eprintf ("Cannot rotate over one or less files.\n");
+					}
 				}
-			} else {
-				eprintf ("Invalid fd number\n");
+				break;
+			case 'n': // "opn" - open next file
+				if (!cmd_op (core, input[1], -1)) {
+					eprintf ("Cannot find next file\n");
+				}
+				break;
+			case 'p': // "opp" - open previous file
+				if (!cmd_op (core, input[1], -1)) {
+					eprintf ("Cannot find previous file\n");
+				}
+				break;
+			case ' ':
+				{
+					int fd = r_num_math (core->num, input + 1);
+					if (fd >= 0 || input[1] == '0') {
+						cmd_op (core, 0, fd);
+					} else {
+						eprintf ("Invalid fd number\n");
+					}
+				}
+				break;
+			default:
+			// case '?': // "op?"
+				r_core_cmd_help (core, help_msg_op);
+				break;
 			}
 		} else {
 			if (core->io && core->io->desc) {
--- a/libr/core/cmd_flag.c
+++ b/libr/core/cmd_flag.c
@@ -768,11 +768,12 @@ rep:
 			break;
 		case ' ':
 			{
-				const int n = atoi (input + 1);
-				if (n + ST8_MAX + 1 < UT8_MAX) {
+				const int ASCII_MAX = 127;
+				int n = atoi (input + 1);
+				if (n + ASCII_MAX + 1 < UT8_MAX) {
 					const char *arg = strchr (input + 2, ' ');
 					ut64 addr = arg? r_num_math (core->num, arg): core->offset;
-					r_core_visual_mark_set (core, n + ST8_MAX + 1, addr);
+					r_core_visual_mark_set (core, n + ASCII_MAX + 1, addr);
 				}
 			}
 			break;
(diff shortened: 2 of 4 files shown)


libr/core/cmd_open.c
```diff
@@ -1373,20 +1400,49 @@ static int cmd_open(void *data, const char *input) {
...
-					r_core_block_read (core);
-				} else {
-					eprintf ("Cannot find RBinFile associated with fd %d\n", fd);
+			switch (input[1]) {
+			case '?': // "op?"
+				eprintf ("Usage: op[rnp] [fd]\n");
+				eprintf (" opr   = open next file rotating if not found\n");
+				eprintf (" opn   = open next file\n");
+				eprintf (" opp   = open previous file\n");
+				eprintf (" op 3  = opep priorizing fd = 3\n");
```
ret2libc, line 1409: typo `opep`

ret2libc: Is this command really necessary? I think there is already another one that does mostly the same thing, no?

ret2libc:
As always, I'm a bit sceptical about adding commands... are they really needed? Did you ever really miss those commands in these 3 years since the issue was created? Also, I think `opp` is a duplicate of something else, as "prioritize a given fd" is something i'm pretty sure is already done by some other command. Also, I imagine myself trying to use these commands and before i use `opn`/`opp`, i'm probably going to run `o` just to see in what orders the file descriptors are and at that point a regular `op X` would be what I would do.

For me this is a :-1: to the whole addition, as I think we already have way too many commands. However I'd suggest taking reviews from others if they think differently.

trufae:
I think it's very useful command when you work with multiple files. Making the experience similar to vim switching to the very next/previous opened file with just one command, instead of having to list the opened files and specify the filedescriptor by hand, which is kind of tedious.

I agree the for() to find the closest next file opened is not ideal, but RIO doesnt provides an api for this. and i would like to get some feedback from @condret as siol author about this topic.

ret2libc:
> I think it's very useful command when you work with multiple files. Making the experience similar to vim switching to the very next/previous opened file with just one command, instead of having to list the opened files and specify the filedescriptor by hand, which is kind of tedious.

I agree vim `bnext`/`bprev` are nice, but in the case of vim you immediately see the buffer you are switching to, while in r2 you don't. So IMHO it is just way faster/easier/more consistent to just specify the fd you want to focus on, instead of moving to the next/prev fd.

trufae:
What do you mean by "viewing the buffer"?

Bind those commands to F keys and use them in visual mode, and you'll see instant feedback. is that what you are refering to?

ret2libc:
Something changes, yeah, but it is not clear at all where you are at.

image
image

In the first image, if I remember correctly, I'm at FD 3 /bin/ls, in the second i'm at FD=5 /bin/cat.

By the way, while doing `opr` if there is a `null://4784` fd, `opr` says `Invalid RBinFile.id number.`. As I said, I don't think we really need this behaviour considering that you will have to go through the `o` output anyway to have a sort of idea where you are at. If you need to look at `o`, you can as well just do `op <fd>` instead of relying on `opn`/`opp`. Anyway, ask for other reviews as well, I'm just one.

trufae:
the visual title filename is wrong, but this is unrelated to this change.

trufae:
and all the flags analysis information, etc... is *not yet* tied to the files opened. this results in confusing views when working with multiple files, but again, this new command is not introducing this issue or so. the binding of anal/flag info to rbinfiles or iofds is a wellknown issue and must be addressed properly, but not here.

i'm glad this topic is open now, maybe we can get some good ideas out of it to address it properly.

libr/core/cmd_open.c
```diff
@@ -1373,20 +1400,49 @@ static int cmd_open(void *data, const char *input) {
...
-					// only useful for io.va=0
-					// load bininfo for given fd
-					r_core_cmdf (core, "obo %d", fd);
-					core->io->desc = desc; // XXX we should use fd here, not *pointer
-					r_core_block_read (core);
-				} else {
-					eprintf ("Cannot find RBinFile associated with fd %d\n", fd);
+			switch (input[1]) {
+			case '?': // "op?"
+				eprintf ("Usage: op[rnp] [fd]\n");
```
ret2libc, line 1405: for consistency you should use `r_core_cmd_help`.

libr/core/cmd_open.c
```diff
@@ -1216,6 +1216,33 @@ static bool desc_list_json_cb(void *user, void *data, ut32 id) {
...
+	}
+	int limit = 1024;
+	int inc = (mode == 'n')? 1: -1;
+	if (cur_fd != -2) {
+		size_t i;
+		int fdn = cur_fd + inc;
+		for (i = 0; i < limit && fdn >= 0; i++) {
+			RIODesc* desc = r_io_desc_get (core->io, fdn);
+			if (desc) {
+				r_core_cmdf (core, "obo %d", fdn);
```
ret2libc, line 1235: `obo` is deprecated apparently. Also, better not to use `r_core_cmd`.

libr/core/cmd_open.c
```diff
@@ -1216,6 +1216,33 @@ static bool desc_list_json_cb(void *user, void *data, ut32 id) {
 	return true;
 }
 
+static bool cmd_op(RCore *core, char mode) {
+	int cur_fd = -2;
+	if (mode == 0) {
+		cur_fd = -1;
+		mode = 'n';
+	} else if (core->io && core->io->desc) {
+		cur_fd = core->io->desc->fd;
```
ret2libc, line 1225: use `r_io_fd_get_current`

libr/core/cmd_open.c
```diff
@@ -1216,6 +1216,33 @@ static bool desc_list_json_cb(void *user, void *data, ut32 id) {
...
+	int limit = 1024;
+	int inc = (mode == 'n')? 1: -1;
+	if (cur_fd != -2) {
+		size_t i;
+		int fdn = cur_fd + inc;
+		for (i = 0; i < limit && fdn >= 0; i++) {
+			RIODesc* desc = r_io_desc_get (core->io, fdn);
+			if (desc) {
+				r_core_cmdf (core, "obo %d", fdn);
+				core->io->desc = desc; // XXX we should use fd here, not *pointer
```
ret2libc, line 1236: maybe `r_io_use_fd` ?

libr/core/cmd_open.c
```diff
@@ -1216,6 +1216,33 @@ static bool desc_list_json_cb(void *user, void *data, ut32 id) {
...
+		mode = 'n';
+	} else if (core->io && core->io->desc) {
+		cur_fd = core->io->desc->fd;
+	}
+	int limit = 1024;
+	int inc = (mode == 'n')? 1: -1;
+	if (cur_fd != -2) {
+		size_t i;
+		int fdn = cur_fd + inc;
+		for (i = 0; i < limit && fdn >= 0; i++) {
```
ret2libc, line 1232: instead of this, it's better if you add a `r_io_get_fds` in RIO. It should be easy to implement, because you can probably just use `r_id_storage_list (io->files)`.

libr/core/cmd_open.c
```diff
@@ -1373,20 +1400,49 @@ static int cmd_open(void *data, const char *input) {
...
+			case 'p': // "opp" - open previous file
+				if (!cmd_op (core, input[1])) {
+					eprintf ("Cannot find previous file\n");
+				}
+				break;
+			default:
+				{
+					int fd = r_num_math (core->num, input + 1);
+					if (fd >= 0 || input[1] == '0') {
+						RIODesc *desc = r_io_desc_get (core->io, fd);
```
ret2libc, line 1432: duplicated code here and in the `cmd_op` function.

libr/core/cmd_open.c
trufae: which commands do the same?

libr/core/cmd_open.c
trufae: ive done the logc inside cmdopen.c without adding new apis, it doesnt looks that bad, but still this cmd_op can have a better name and get better arguments

trufae: see the new implementation

libr/core/cmd_open.c
```diff
@@ -1216,6 +1225,76 @@ static bool desc_list_json_cb(void *user, void *data, ut32 id) {
...
+	data->prevfd = id;
+	return true;
+pickme:
+	data->picknext = false;
+	data->found = true;
+	return false;
+	
+}
+
+static bool cmd_op(RCore *core, char mode, int fd) {
```
ret2libc, line 1262: this whole function and the one before looks a bit messy :/ I have an hard time following they are doing even if I know it.
I think something like this is clearer in its intention.
```C
static bool find_and_switch_fd(RCore *core, int mode) {
	RList *descs = r_io_desc_list (core->io);
	RIODesc *cur_desc = r_io_desc_get (core->io, r_io_fd_get_current (core->io));
	RListIter *it;
	RIODesc *desc;
	r_list_foreach (descs, it, desc) {
		if (desc == cur_desc) {
			RListIter *target = get_target_switch_fd (descs, it, mode);
			if (target) {
				RIODesc *set_desc = (RIODesc *)target->data;
				switch_fd (core, set_desc->fd);
				r_list_free (descs);
				return true;
			}
		}
	}
	r_list_free (descs);
	return false;
}
```

trufae: My implementation won't need to iterate over all descs in all the cases, also your implementation is not handling the case an fd is passed. I agree that the code is messy, and i consider it wip, but i would prefer to not add new apis to r_io when we can iterate directly. I agree the ridstorage usage is not as handy as r_list. I'll try to refactor this a bit more when i get some spare time

ret2libc: it doesn't need to handle the case when the fd is passed, because that one can be done just with `switch_fd (core, fd)`.

ret2libc: How can you avoid iterating over all the descs up to the current one? I did not find any API to directly get the previous/next desc by just having the current one without the list (or without iterating with the foreach).

Also, why so many problems to add an API to r_io? It is just an addition and a one that may come handy in more than one case. If it was a change, I would have understood, but a common addition like the "to_list" method seems nice to me.

libr/core/cmd_open.c
```diff
@@ -1216,6 +1225,76 @@ static bool desc_list_json_cb(void *user, void *data, ut32 id) {
...
+	int inc;
+	int prevfd;
+	bool picknext;
+	bool found;
+	RCore *core;
+} RCmdOpData;
+
+static bool try_next(void *_user, void *_data, ut32 id) {
+	RCmdOpData *data = _user;
+	if (data->picknext) {
```
ret2libc, line 1239: having all that RCmdOpData for a simple operation such as this one looks too much to me. Getting the next/previous desc is really as simple as: 1) find the current descriptor in the list 2) get the next/prev descriptor. This `try_next` make it look much harder than what it is IMHO.

Btw, it seems `picknext` is used somehow in a unintuitive way. In this line, if data->picknext is true, you don't pick any next thing, you just set the fd and pick it, so it does the opposite than what the name suggests.

libr/core/cmd_open.c
```diff
@@ -1216,6 +1225,76 @@ static bool desc_list_json_cb(void *user, void *data, ut32 id) {
...
+typedef struct {
+	int fd; // input and output
+	int inc;
+	int prevfd;
+	bool picknext;
+	bool found;
+	RCore *core;
+} RCmdOpData;
+
+static bool try_next(void *_user, void *_data, ut32 id) {
```
condret, line 1237: I'd suggest to rename this function, so that becomes more clear what its' purpose is

libr/core/cmd_open.c
condret: You can also iterate over r_id_storage without passing a function-pointer. If you want I can create a makro for that. So far I haven't seen necessity for that as I personally find passing fcn-pointers leads to shorter functions. no need for to_list imo.

condret: you can use something like this to find the next fd (I hope I understodd correctly, that this is what you want to do)
```
ut32 next_fd = cur_fd + 1;
RIODesc *next = NULL;
for (; next_fd != cur_fd; next_fd = (next_fd + 1) % io->files->size) {
	next = r_io_desc_get (io, next_fd);
	if (next) {
		break;
	}
}
```

I think having a fcn in io and id_storage for this would be good. I can do that change for you @trufae today/evening

ret2libc:
This is waiting for https://github.com/radareorg/radare2/pull/17271

libr/core/cmd_open.c
```diff
@@ -30,7 +30,7 @@ static const char *help_msg_o[] = {
 	"om","[?]","create, list, remove IO maps",
 	"on"," [file] 0x4000","map raw file at 0x4000 (no r_bin involved)",
 	"oo","[?+bcdnm]","reopen current file (see oo?) (reload in rw or debugger)",
-	"op"," [fd]", "select the given fd as current file (see also ob)",
+	"op","[r|n|p|fd]", "select priorized file by fd (see ob), opn/opp/opr = next/previous/rotate",
```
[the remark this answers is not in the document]
trufae: op means: open priorized. what its internally doing is changing the selected io->desc. that's independent from maps in va mode. But it ends up calling obo which switches the default rbin object and reconfigures the maps.

ret2libc:
Closing in favour of https://github.com/radareorg/radare2/pull/17652.

PR #17168, opened 2020-06-28
Fix supported OS list from the README

trufae:

**Detailed description**

it's quite outdated/wrong

**Test plan**

is windows 7 the minimum windows supported?

What do you think about putting the release download links in the REAMDE?

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ You can also use [r2lrn](https://github.com/0ki/r2lrn) or r2golf for a hands-on 
 
 ## Operating Systems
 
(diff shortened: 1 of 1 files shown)


XVilka:
I think we should work on top of Windows XP afaik. @GustavoLCR or @kazarmy or @a1ext or @pelijah might know better.

eagleoflqj:
What about adding supported architecture along with OS (then architecture can link to release download)? Because Linux and Windows run on x86 and arm, and macOS will run on arm soon.

trufae:
The list of supported architectures is pretty large, the only limitation is the native debugger support which requires dedicated development to support the arch completely. As long as we know that some tests are not passing on big endian machines or arm/mips/ppc.. we could say those archs are % complete.

PR #17173, opened 2020-06-28
Fix #16781 - Add --without-dylink configure flag to disable libdl features ##build

### LINKED ISSUE #16781: Add configure option to make r2 builds without dlopen support
This is very useful for static builds that may crash trying to load plugins or you just want faster startup times without having to use the -N or the R2_NOPLUGINS env var

trufae:

**Detailed description**

Allow to build r2 without support for dynamic linking libraries, useful for portability and solve static build issues

* libdl is usually not available in static form
* static build are usually not in sync with the system install of r2, and will try to load many plugins that can make it crash
* loads faster

**Test plan**

...

**Closing issues**

...


ret2libc:
> and solve static build issues

What are the static build issues you are referring to?

trufae:
If you use a static build of r2 from any version that is not the one you have in your system. r2 will try to load the plugins, if they are from the same version (git) it will crash because the version check will pass, but the code has changed. But also, the time spent trying to load all the dynamic plugins and traversing the plugin directories is quite unnecessary if you dont really need that. this configure option aims to address that.

Before you ask, this is an issue i have encountered a couple of times recently

PR #17178, opened 2020-06-29
Initial support for git tags in r2pm ##r2pm

trufae:

**Detailed description**

When using r2 from non-git (release builds), installing packages can result in build problems. This PR aims to enable r2pm with the ability to pick the right commit from the tagged version from 3rd party repos if available.

Afaik this is done by r2ghidra-dec and r2dec-js, it may be good to define a different tag name than just the r2 version string, otherwise it will confuse users with the version of the decompiler and the veresion of the r2 required to build. so my suggestion is to use `r2-` as prefix in the 3rd party repos

**Test plan**

Get r2-4.4, replace the r2pm shellscript with the one in this PR, and type the following command:

```
$ r2pm -ci r2ghidra-dec
```

As you can see the PR also introduces r2 -H R2_VERSION, but r2 -qv is used for now, so backward compat may be handled in here.

**Closing issues**

none, but this thing has been requested for years, and i dont think those few lines will hurt anyone

### DIFF
--- a/binr/r2pm/r2pm
+++ b/binr/r2pm/r2pm
@@ -336,7 +336,7 @@ R2PM_GIT() {
 	# r2>=4.5 : R2_VERSION=$(r2 -H R2_VERSION)
 	R2_VERSION=$(r2 -qv)
 	if [ -z "`echo $R2_VERSION | grep git`" ]; then
-		git checkout v"${R2_VERSION}"
+		git checkout "r2-${R2_VERSION}"
 		git pull
 	fi
 }
--- a/binr/r2pm/r2pm
+++ b/binr/r2pm/r2pm
@@ -327,10 +327,17 @@ R2PM_GIT() {
 	cd "${R2PM_GITDIR}"
 	if [ -d "${DIR}" ]; then
 		cd "${DIR}"
+		git checkout master
 		git pull
 	else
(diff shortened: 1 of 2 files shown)


trufae:
Can you please agree on what i'm suggesting in the PR description? (the r2- prefix vs the `v` one)? basically because this must be specified for all the 3rd party projects using r2pm

thestr4ng3r:
I am not a fan of the `r2-` prefix for versions, but it is possible to tag a single commit with multiple tags, so doing both `v` and `r2-` can be done, so it's fine for me.

thestr4ng3r:
Actually having both `r2-` and `v` makes even more sense than r2pm directly using `v` since the r2 tags can be re-tagged and they don't have to be bount to plugin releases.

ret2libc:
For me it is ok to have `r2-X.Y.Z` as tags in various r2 plugins to mark them as working with specific r2 version. This is a great feature!

trufae:
Change done, waiting for testing + updated plugins using the tags (at least for 4.4)

thestr4ng3r:
Pushed `r2-4.4.0` to r2ghidra.

wargio:
i'm ok with `r2-x.y.z`

PR #17181, opened 2020-06-29
Fix r2r crash on Windows if test doesn't have CMDS

GustavoLCR:

### DIFF
--- a/binr/r2r/run.c
+++ b/binr/r2r/run.c
@@ -942,6 +942,9 @@ R_API R2RProcessOutput *r2r_run_cmd_test(R2RRunConfig *config, R2RCmdTest *test,
 		}
 		r_list_push (files, "-");
 	}
+	if (!test->cmds.value) {
+		test->cmds.value = strdup ("");
+	}
 	R2RProcessOutput *out = run_r2_test (config, test->cmds.value, files, extra_args, test->load_plugins, runner, user);
 	r_list_free (extra_args);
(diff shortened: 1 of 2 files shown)


binr/r2r/run.c
```diff
@@ -942,6 +942,9 @@ R_API R2RProcessOutput *r2r_run_cmd_test(R2RRunConfig *config, R2RCmdTest *test,
 		}
 		r_list_push (files, "-");
 	}
+	if (!test->cmds.value) {
+		test->cmds.value = strdup ("");
+	}
 	R2RProcessOutput *out = run_r2_test (config, test->cmds.value, files, extra_args, test->load_plugins, runner, user);
```
trufae, line 948: i don't think we want to run a test that have no commands. imho is better to return early and add r_return_val_if_fail everywhere to avoid those errors in the future.

So please:

* Add r_return in run_r2_test
* if cmds.value is null, throw a warning and return, this is not a valid test, so better not to run an empty thing

kazarmy: I would prefer to throw an error and abort, since clearly something has gone wrong. Throwing warnings assumes that someone is actually reading them, and that's not guaranteed especially if the log output is a gazillion lines long.

trufae: no need to be that drastic otherwise r2r couldnt be run from inside r2 itself or used as a library.

PR #17184, opened 2020-06-30
Fix covs from 2020-06-29

trufae:
* CIDs: 1293553 1408695 1430058 1430059 1430060
1430061 1430062 1430063 1430064 1430065 1430067

 

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -1489,5 +1489,6 @@ beach:
 	free (file);
 	r_cons_free ();
 	LISTS_FREE ();
+	R_FREE (pfile);
 	return ret;
 }
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -1486,7 +1486,6 @@ beach:
 	//r_core_file_close (r, fh);
 	r_core_free (r);
 	r_cons_set_raw (0);
-	free (file);
 	r_cons_free ();
 	LISTS_FREE ();
 	R_FREE (pfile);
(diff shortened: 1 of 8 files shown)


libr/socket/run.c
```diff
@@ -453,10 +453,11 @@ static int handle_redirection(const char *cmd, bool in, bool out, bool err) {
 	}
 }
 
-R_API int r_run_parsefile(RRunProfile *p, const char *b) {
+R_API bool r_run_parsefile(RRunProfile *p, const char *b) {
+	r_return_val_if_fail (p && b, 0);
```
ret2libc, line 457: `s/0/false`

libr/main/radare2.c
```diff
@@ -1489,5 +1489,6 @@ R_API int r_main_radare2(int argc, const char **argv) {
 	free (file);
 	r_cons_free ();
 	LISTS_FREE ();
+	R_FREE (pfile);
```
ret2libc, line 1492: I'm not 100% sure this is correct. In some paths in this function there is `pfile = file` and file is already freed. So I'm not sure this won't cause double frees in some cases.

radare: check now, the free(file) was a bit weird. im not 100% sure either

ret2libc: still not super sure... it could be that now `file` will leak :D but the pR is ok to me. let's see. this code is a bit messy and hard to follow.

PR #17188, opened 2020-06-30
Cleanup the ELF code a bit

trufae:

**Detailed description**

A little cleanup in the ELF parser code

**Test plan**

wait for the ci

**Closing issues**

This code is stripped from the elf-relocs pr

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -1799,11 +1799,11 @@ ut64 Elf_(r_bin_elf_get_main_offset)(ELFOBJ *bin) {
 	if (entry == UT64_MAX) {
 		return UT64_MAX;
 	}
+	ut8 buf[256];
 	if (entry > bin->size || (entry + sizeof (buf)) > bin->size) {
 		return UT64_MAX;
 	}
 	// unnecessary to read 512 bytes imho
-	ut8 buf[256];
 	if (r_buf_read_at (bin->b, entry, buf, sizeof (buf)) < 1) {
 		bprintf ("read (main)\n");
 		return UT64_MAX;
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -2808,7 +2808,7 @@ RBinElfLib* Elf_(r_bin_elf_get_libs)(ELFOBJ *bin) {
 	return ret;
 }
 
-static void create_section_from_phdr(ELFOBJ *bin, RBinElfSection *ret, int *i, const char *name, ut64 addr, ut64 sz) {
+static void create_section_from_phdr(ELFOBJ *bin, RBinElfSection *ret, size_t *i, const char *name, ut64 addr, ut64 sz) {
 	if (!addr) {
 		return;
 	}
@@ -2864,7 +2864,7 @@ static RBinElfSection *get_sections_from_phdr(ELFOBJ *bin) {
 		return NULL;
 	}
 
-	i = 0;
+	size_t i = 0;
 	create_section_from_phdr (bin, ret, &i, ".rel.dyn", reldyn, reldynsz);
 	create_section_from_phdr (bin, ret, &i, ".rela.plt", relava, pltgotsz);
 	create_section_from_phdr (bin, ret, &i, ".rel.plt", relva, relasz);
@@ -3904,18 +3904,17 @@ static bool get_nt_file_maps (ELFOBJ *bin, RList *core_maps) {
 			int len_str = 0;
 			while (n_maps > 0) {
 				ut64 addr;
-				char str[512] = {0};
 				if (bits == 64) {
 					addr = BREAD64 (bin->b, i);
 				} else {
 					addr = BREAD32 (bin->b, i);
 				}
 				if (addr == UT64_MAX) {
-					eprintf ("ffbreak\n");
 					break;
 				}
+				char str[512] = {0};
 				r_buf_read_at (bin->b, jump + len_str, (ut8*)str, sizeof (str) - 1);
-				str[sizeof(str) - 1] = 0; // null terminate string
+				str[sizeof (str) - 1] = 0; // null terminate string
 				RListIter *iter;
 				RBinMap *p;
 				r_list_foreach (core_maps, iter, p) {
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -1430,7 +1430,7 @@ static ut64 get_import_addr_mips(ELFOBJ *bin, RBinElfReloc *rel) {
 	}
 
 	const ut8 *base = r_mem_mem_aligned (buf, sizeof (buf), (const ut8 *)"\x3c\x0f\x00", 3, 4);
-	plt_addr += base ? (int)(size_t) (base - buf) :  MIPS_PLT_OFFSET + 8; // HARDCODED HACK
+	plt_addr += base? (int)(size_t) (base - buf):  MIPS_PLT_OFFSET + 8; // HARDCODED HACK
 	plt_addr += pos * 16;
 
 	return plt_addr;
@@ -3785,15 +3785,15 @@ ut64 Elf_(r_bin_elf_p2v) (ELFOBJ *bin, ut64 paddr) {
 
 /* Deprecated temporarily. Use r_bin_elf_v2p_new in new code for now. */
 ut64 Elf_(r_bin_elf_v2p) (ELFOBJ *bin, ut64 vaddr) {
-	size_t i;
-
-	r_return_val_if_fail (bin, 0); // UT64_MAX or vaddr?
+	r_return_val_if_fail (bin, 0);
 	if (!bin->phdr) {
 		if (is_bin_etrel (bin)) {
 			return vaddr - bin->baddr;
 		}
 		return vaddr;
 	}
+
+	size_t i;
 	for (i = 0; i < bin->ehdr.e_phnum; i++) {
 		Elf_(Phdr) *p = &bin->phdr[i];
 		if (p->p_type == PT_LOAD && is_in_vphdr (p, vaddr)) {
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -8,14 +8,6 @@
 #include <r_util.h>
 #include "elf.h"
 
-#ifdef IFDBG
-#undef IFDBG
-#endif
-
-#define DO_THE_DBG 0
-#define IFDBG if (DO_THE_DBG)
-#define IFINT if (0)
-
 #define MIPS_PLT_OFFSET 0x20
 #define RISCV_PLT_OFFSET 0x20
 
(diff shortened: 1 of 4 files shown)


libr/bin/format/elf/elf.c
```diff
@@ -1434,20 +1422,15 @@ static ut64 get_import_addr_mips(ELFOBJ *bin, RBinElfReloc *rel) {
...
 	}
 
-	base = r_mem_mem_aligned (buf, sizeof (buf), (const ut8 *)"\x3c\x0f\x00", 3, 4);
-	if (base) {
-		plt_addr += (int)(size_t) (base - buf);
-	} else {
-		plt_addr += MIPS_PLT_OFFSET + 8; // HARDCODED HACK
-	}
+	const ut8 *base = r_mem_mem_aligned (buf, sizeof (buf), (const ut8 *)"\x3c\x0f\x00", 3, 4);
+	plt_addr += base ? (int)(size_t) (base - buf) :  MIPS_PLT_OFFSET + 8; // HARDCODED HACK
```
ret2libc, line 1433: style

libr/bin/format/elf/elf.c
```diff
@@ -1459,7 +1442,9 @@ static size_t get_size_rel_mode(Elf_(Xword) rel_mode) {
 
 static size_t get_num_relocs_dynamic_plt(ELFOBJ *bin) {
 	if (bin->dyn_info.dt_pltrelsz) {
-		return bin->dyn_info.dt_pltrelsz / get_size_rel_mode (bin->dyn_info.dt_pltrel);
+		const size_t size = bin->dyn_info.dt_pltrelsz;
```
ret2libc, line 1445: I think this should be `ut64`, otherwise on 32bit machine it's going to fail.

libr/bin/format/elf/elf.c
```diff
@@ -2625,10 +2612,10 @@ static bool sectionIsValid(ELFOBJ *bin, RBinElfSection *sect) {
 static Elf_(Xword) get_section_mode(ELFOBJ *bin, size_t pos) {
 	if (r_str_startswith (bin->g_sections[pos].name, ".rela.")) {
 		return DT_RELA;
-	} else if (r_str_startswith (bin->g_sections[pos].name, ".rel.")) {
+	}
```
ret2libc, line 2615: I fail to understand why this style :( although the `return` means nothing can be executed after the `if`, the `else if` makes it very clear than it is all a single "logical block". (it is also more compact, but this is a very relative benefit)

libr/bin/format/elf/elf.c
```diff
@@ -3802,9 +3785,9 @@ ut64 Elf_(r_bin_elf_p2v) (ELFOBJ *bin, ut64 paddr) {
 
 /* Deprecated temporarily. Use r_bin_elf_v2p_new in new code for now. */
 ut64 Elf_(r_bin_elf_v2p) (ELFOBJ *bin, ut64 vaddr) {
-	int i;
+	size_t i;
 
-	r_return_val_if_fail (bin, 0);
+	r_return_val_if_fail (bin, 0); // UT64_MAX or vaddr?
```
ret2libc, line 3790: no need for this comment. This function shall be removed at some point. As it can be seen, the `_new` version shall be used for all future code.

libr/bin/p/bin_elf.inc
```diff
@@ -677,27 +677,20 @@ static RBinReloc *reloc_convert(struct Elf_(r_bin_elf_obj_t) *bin, RBinElfReloc
...
 		return NULL;
 	}
 	/* FIXME: This is a _temporary_ fix/workaround to prevent a use-after-
 	 * free detected by ASan that would corrupt the relocation names */
 	r_list_free (imports (bf));
-	if ((got_addr = Elf_(r_bin_elf_get_section_addr) (bin, ".got")) == -1) {
+	ut64 got_addr = Elf_(r_bin_elf_get_section_addr) (bin, ".got");
+	if (got_addr == UT64_MAX) {
 		got_addr = Elf_(r_bin_elf_get_section_addr) (bin, ".got.plt");
-		if (got_addr == -1) {
```
ret2libc, line 698: why?

libr/bin/p/bin_elf.inc
```diff
@@ -860,39 +854,38 @@ static void _patch_reloc(ut16 e_machine, RIOBind *iob, RBinElfReloc *rel, ut64 S
...
+	RBinObject *obj = r_bin_cur_object (b);
 	if (!obj) {
 	   	return NULL;
 	}
-	bin = obj->bin_obj;
+	struct Elf_(r_bin_elf_obj_t) *bin = obj->bin_obj;
 	if (bin->ehdr.e_type != ET_REL) {
 		return NULL;
 	}
+	size = bin->g_reloc_num * 4;
```
ret2libc, line 878: why? also, size == 0 means g_reloc_num was 0 in the first place. Moreover, `g_reloc_num * 4` is already computed few lines below.

libr/bin/format/elf/elf.c
trufae: it makes clear that there's nothing after that. the else is dead code, and visually it looks longer but cleaner

libr/bin/format/elf/elf.c
trufae: this must be added into the elf-ut64-max pr

libr/bin/p/bin_elf.inc
trufae: because it's checked below already

libr/bin/p/bin_elf.inc
trufae: Not exactly, int overflow can happen :}

```
$ cat a.c

int main() {
	short i = 16384;
	printf ("%hd\n", i * 4);
}
```

^- the output of this is `0`

libr/bin/p/bin_elf.inc
ret2libc: if you mean that it is checked by https://github.com/radareorg/radare2/pull/17188/files#diff-be1d0d79bce9095b796c70fcfc154c1eL702 I'm not sure it works, because got_addr is ut64 and with your patch got_addr would remain UT64_MAX (-1). When the next `if` checks got_addr < 1 it would find it false.

On the other hand, on master when got_addr is UT64_MAX(-1) it gets replaced with 0, thus it enters the next if.

libr/bin/p/bin_elf.inc
ret2libc: true, good point. we should define a macro or something in r_types_base.h to detect these kind of things

ret2libc: however, if things overflow they can even become something different than 0, can't they? so i'm not sure size == 0 is the right check to check overflows.

trufae: There's UT64_ADD_OVFCHK , but not for MUL, only for ADD

ret2libc: Would something like https://stackoverflow.com/posts/1514309/revisions work?

trufae: yep. these lines should be added as apis in r2 (or macros as UT64_MUL_OVFCHK)..

libr/bin/p/bin_elf.inc
trufae: Addr 0 is a valid address. -1 is not. Its green so it seems that is passing. I have more prs with tweaks in the elf parser to make it work better in such cases. And also making -1 become 0 is a bad idea.

ret2libc: Why is it a bad idea? And why just removing this if should be better idea? I agree this code is not ideal, but I don't see a real reason for removing this `if` here and not, for example, few linew below at line 697. Also, removing this `if`, as I said in previous comment, changes the behaviour without an actual reason. I would reintroduce the `if`, unless you can improve the entire codebase of this function or prove this is better.

libr/bin/p/bin_elf.inc
ret2libc: Good, waiting for https://github.com/radareorg/radare2/pull/17216

libr/bin/p/bin_elf.inc
trufae: Because 0 is a valid address and ut64-max not. which other reason do you need to keep the old behaviour? there's another PR introducing more changes to remove the assumption that 0 is not valid, but i wanted to add this change in here to confirm if it was breaking something in the suite. and seems like not, so when this gets merged i will continue the changes in the other pr

ret2libc: I don't see why you have 3/4 elf-related PRs opened that are more or less working all on the same code and you keep referencing one from another. If all those changes are related, just do them in one PR. for example this line, I think, was also touched in another PR (where however you have also removed the other if and done some other things..)

trufae: well, i have splitted the original PR into 5 different PRs. this is why i reference them and also why there are some mixed changes

trufae: i can move this into the other ut64max pr if you want, i was trying to reduce the size of the original PR to get it into a shape that can be reviewed properly and i dont see whats wrong in this change here.

libr/bin/p/bin_elf.inc
trufae: im fine to hold this merge until 17216 is in. but bin parsers are full of the need for this ovfchk, so maybe worth just address them together

trufae:
ping @ret2libc

libr/bin/format/elf/elf.c
ret2libc: this is still not addressed.

libr/bin/format/elf/elf.c
```diff
@@ -1709,14 +1694,14 @@ ut64 Elf_(r_bin_elf_get_baddr)(ELFOBJ *bin) {
...
 ut64 Elf_(r_bin_elf_get_boffset)(ELFOBJ *bin) {
-	int i;
 	ut64 tmp, base = UT64_MAX;
 	r_return_val_if_fail (bin, 0);
 
 	if (!bin->phdr) {
 		return 0;
 	}
 
+		size_t i;
```
ret2libc, line 1704: wrong indent

libr/bin/p/bin_elf.inc
ret2libc: please remove this change from this PR

libr/bin/format/elf/elf.c
trufae: pltrelsz is a small number, could be 8, 32 or 2. making it 64bit wide wont fix anything

libr/bin/p/bin_elf.inc
trufae: done

trufae:
please review again

libr/bin/format/elf/elf.c
ret2libc: I don't see where is the check that pltrelsz should be 8, 2 or 32, since pltrelsz comes directly from the binary. Also, in glibc_elf.h you see:
```
typedef uint64_t Elf32_Xword;
typedef uint64_t Elf64_Xword;
```
So I don't see why this should be cast to size_t for no reason.

libr/bin/format/elf/elf.c
ret2libc: indent

trufae:
done

PR #17150, opened 2020-06-25
Add support for binr/blob and fix android build ##meson

ret2libc:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

This is necessary to build static package similar to busybox, with different tools being just links to the main one. It also fixes meson.build on android.

**Test plan**

I will add soon a gh workflow to test multiple things, this included as part of the android build.

EDITED to add repro steps:
```
meson --prefix=/tmp/radare2-blob-install -Dlocal=true -Dblob=true build
ninja -C build
ninja -C build install
ls -la /tmp/radare2-blob-install/bin/
```
This should display only one binary and the rest of the binaries should be links to radare2 binary. radare2/rabin2/rasm2 should work as expected.

PR #17192, opened 2020-07-01
Release pipeline on Github action

ret2libc:
  - [x] mention in radare2 book how to build stuff for android with meson

**Detailed description**

So far releases have been done by @trufae on his own machine, with the radareorg/radare2-release repository. This required the guy who did the release to have a bunch of software installed on his machine and to check everything was ok. Also, it required to have osx + linux, at least. Checking that those things were actually working needed to be done manually.

That is quite some work, so this PR aims to create a github workflow that is automatically triggered whenever some commits are pushed to a `release-X.Y.Z` branch. Later, we can extend this workflow to perform additional jobs to ensure a release works well in many different environment that we can't test on every commit. Currently the workflow does the following:
- creates the source tarball for the release (e.g. it includes the capstone code, but in the future we can include other stuff as well, for example if we are going to use submodules for external repository like sdb, spp, tree-sitter, etc.)
- build debian packages on debian:buster
- build macos pkg
- build windows zip file and windows installer (FIXME: the zip file probably should provide the static compiled binaries, not the dynamic ones)
- build ios cydia packages
- build android packages with meson/ninja
- build some r2pm plugins that are shipped as part of radare2-extras and radare2-pipe
- test the debian deb package on debian:buster, to ensure r2 can be run and r2 libraries can be used
- test the macos pkg, to ensure r2 can be run and r2 lbiraries can be used
- test windows clang_cl installer, to ensure r2 can be run and r2 libraries can be used
- create a draft release on github and it uploads all the generated artifacts

With this script, we can have `release-X.Y.Z` branches. The workflow I have in mind is something like this:
1) at some point we choose it is time to make a release, so someone with write access to the repository creates a release branch (e.g. `release-4.5.0`) and updates the VERSION tag
2) at that point, the release code has been branched, so master can move one with whatever breaking change we are going to ship (e.g. enable cfg.newshell by default, change some commands or stuff like that).
3) github CI will do all the tests above and if all goes well, it is going to create a draft release with all the right artifacts
4) someone with write access writes the release notes and at some point it publish the draft release
5) if something goes wrong during the github CI tests or at some point after the release, we can fix stuff in master and cherry-pick only the minimal changes to the `release-X.Y.Z` branch. Github CI will run again and if necessary we can make a new release with minor fixes and stuff. In this way doing a minor fix will just include that minor fix, not other 123 commits that may or may not break everything.

Probably we need a well-written document on how to make a release, as multiple steps are needed anyway (e.g. something that explains the correct branch name to use, how to cherry-pick commits to the release branch, how to update irc title, what to write on twitter, etc.). Some steps, at the moment, are known only to @trufae AFAIK, like uploading the ios package to cydia repository.

**Test plan**

Once the PR is merged, create a branch with something like `release-test` and see if things correctly work and a draft release is generated. It should look something like https://github.com/ret2libc/radare2/releases/tag/test-4.5.0-git .

**Closing issues**

None.

**Additional info**

This is based on https://github.com/radareorg/radare2/pull/17150, I will update this once that other PR gets merged.

This in Travis/Appveyor that are not ported yet, for future PR:
- on Travis: COMPILER_NAME=gcc CXX=g++ CC=gcc INSTALL_SYSTEM=meson R2_SYS_MAGIC=true R2_SYS_ZIP=true R2_SYS_LZ4=true R2_SYS_ZLIB=true R2_SYS_OPENSSL=true
- on Travis: COMPILER_NAME=clang CXX=clang++ CC=clang CFLAGS="-DR2_ASSERT_STDOUT=1" ASAN=1 ASAN_OPTIONS=detect_odr_violation=0 LSAN_OPTIONS=detect_leaks=0 RELEASE=1
- appveyor can checks multiple versions of visualstudio, GH seems to allow only the latest one. So we may still need appveyor.


sys/ios-cydia.sh
```diff
@@ -18,9 +18,13 @@ export BUILD=1
 
 if [ ! -d sys/ios-include/mach/vm_behavior.h  ]; then
 (
-	cd sys && \
-	wget http://lolcathost.org/b/ios-include.tar.gz && \
-	tar xzvf ios-include.tar.gz
+	if [ ! -f sys/ios-include.tar.gz ]; then
```
ret2libc, line 21: FYI, I included `sys/ios-include.tar.gz` directly in the radare2 codebase as lolcathost was often unreachable.

libr/debug/p/debug_native.c
```diff
@@ -1455,13 +1455,11 @@ static int r_debug_native_bp(RBreakpoint *bp, RBreakpointItem *b, bool set) {
...
 			: drx_del (dbg, bp, b);
-#elif __arm64__ || __aarch64__
-#if __linux__
+#elif (__arm64__ || __aarch64__) && __linux__
 		return set
 			? arm64_hwbp_add (dbg, bp, b)
 			: arm64_hwbp_del (dbg, bp, b);
-#endif
-#elif __arm__
+#elif __arm__ && __linux__
```
ret2libc, line 1462: this was needed because `arm32_*` functions are not defined for ios-cydia32 build.

ret2libc:
You can look at https://github.com/ret2libc/radare2/releases/tag/test-4.5.0-git for an example of what it generated on my fork.

ret2libc:
I am willing to work on a proper document to explain how to handle releases, but I'm going to need some help from @trufae about this. Right now there is just some, probably outdated, document at https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#how-to-release and https://github.com/radareorg/radare2/blob/master/doc/releases.md .

wargio:
WOAH. GREAT JOB.

ret2libc:
I'd like to mention that most of the scripts to build the artifacts were taken from the work done by @trufae in radare2-release repository.

XVilka:
@ret2libc what about publishing a Docker image with the release tag as well? As we do in https://github.com/radareorg/radare2/blob/master/.github/workflows/docker.yml but with a new tag.

XVilka:
@ret2libc also, what about killing DEB packaging from Travis, and moving DEB packaging to GitHub Actions too, even for every commit?

ret2libc:
> @ret2libc what about publishing a Docker image with the release tag as well? As we do in https://github.com/radareorg/radare2/blob/master/.github/workflows/docker.yml but with a new tag.

yes!



> @ret2libc also, what about killing DEB packaging from Travis, and moving DEB packaging to GitHub Actions too, even for every commit? Or just nightly.

As said on Telegram, I have many doubts about the usefulness of this. The package we generate work only on the version of debian we generate them on (debian:buster in this case, and by the way this should probably be made clearer from the package name). So, how many of our users are debian:buster users? Not sure... But I would prefer to have something like AppImage or docker or something similar that works on multiple systems instead of a .deb that works only for those that use debian (and a particular version of it).

> Moreover, we can use musl-based static build of these *.deb packages, so it will work everywhere: https://ers35.com/statically-link-c-programs-with-musl-gcc.html

This would still be limited to debian only I think. AppImage or something would be better IMHO.

ret2libc:
So, I think the current state is quite good for review and inclusion. Of course there are many other things that can be done, but I'd like to add them incrementally later and/or in following releases, unless there is something in particular you think we should definitely add that was there before and it is not here now (I think before we were doing nothing automatically before release, so in theory this is just a plain improvement over the current status).

TODO:
- merge ci.yml, pr.yml, centos6 and debian.yml in this file and enable/disable jobs based on the type of event/branch name. This will have the benefit that the ci.yml/pr.yml/centos6/debian tests will run also on releases-X.Y.Z branches and that we can make the various jobs depend on each other.
- add ASAN build job
- add capstone 3/4/5 build job
- add various sys_* jobs (sys_magic, sys_capstone, sys_xxhash, etc.) for example on fedora to check whether things still compile well when using system libraries instead of ours
- add tests for r2lang plugins
- include r2-bindings as well (when it will work :D)
- maybe the create-release + publish-docker-image jobs should be run only when creating a tag and not when pushing to the branch... not sure.

XVilka:
Eventually we can create a GitHub Action to simplify plugins checking (for 3rd party plugins).

XVilka:
@ret2libc why still draft? Isn't this ready for the review? Or you are waiting for https://github.com/radareorg/radare2/pull/17150 to be merged first?

ret2libc:
Yes.

XVilka:
Meson Android was merged, please rebase, test again, and lets merge this

XVilka:
Some tests failed still at https://travis-ci.com/github/radareorg/radare2/jobs/357707197:
```
[**]                                    db/formats/bflt     1075 OK        33 BR        0 XX       13 FX
[XX] db/formats/nes NES: Pong - open
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -A -Qc q! bins/nes/Pong.nes
-- stdout
@@ -0,0 +1,18 @@
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
+WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103)
-- stderr
[Cannot find function at 0x00008000
[x] Check for objc references
Warning: aao experimental on 32bit binaries
[x] Check for vtables
[x] Finding xrefs in noncode section with anal.in=io.maps
[x] Analyze value pointers (aav)
[x] Value from 0x00010000 to 0x00014000 (aav)
[x] 0x00010000-0x00014000 in 0x10000-0x14000 (aav)
[x] 0x00010000-0x00014000 in 0xc000-0x10000 (aav)
[x] 0x00010000-0x00014000 in 0x8000-0xc000 (aav)
[x] Value from 0x0000c000 to 0x00010000 (aav)
[x] 0x0000c000-0x00010000 in 0x10000-0x14000 (aav)
[x] 0x0000c000-0x00010000 in 0xc000-0x10000 (aav)
[x] 0x0000c000-0x00010000 in 0x8000-0xc000 (aav)
[x] Value from 0x00008000 to 0x0000c000 (aav)
[x] 0x00008000-0x0000c000 in 0x10000-0x14000 (aav)
[x] 0x00008000-0x0000c000 in 0xc000-0x10000 (aav)
[x] 0x00008000-0x0000c000 in 0x8000-0xc000 (aav)
[Warning: No SN reg alias for current architecture.
Warning: No SN reg alias for current architecture.
Warning: No SN reg alias for current architecture.
[x] Emulate code to find computed references (aae)
```

Then another error at https://travis-ci.com/github/radareorg/radare2/jobs/357707196 or https://travis-ci.com/github/radareorg/radare2/jobs/357707201:
```
The Meson build system
Version: 0.49.2
Source dir: /home/travis/build/radareorg/radare2
Build dir: /home/travis/build/radareorg/radare2/build
Build type: native build
meson.build:1:0: ERROR:  Meson version is 0.49.2 but project requires >=0.50.1.
A full log can be found at /home/travis/build/radareorg/radare2/build/meson-logs/meson-log.txt
+ exit 1
```

ret2libc:
@XVilka those errors in travis has nothing to do with this PR. If you create a new branch with the name `release` in it, you will see those errors anyway.

XVilka:
Then it's LGTM as is.

ret2libc:
I think this is ready. Please review.

.github/meson-android-aarch64.ini
```diff
@@ -0,0 +1,18 @@
+[binaries]
+c       = '/usr/local/lib/android/sdk/ndk-bundle/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android28-clang'
```
trufae, line 2: why assume those paths? this looks like very wrong to me because:

* each distro/OS have the ndk installed in different path
* each ndk use to drop the bins in different places
* ndk provides a script that builds the environment to use "gcc, make, ranlib,..." without having to use those aliases

trufae: can you let the ndk prefix path be in a separate variable?

.github/meson-android-arm.ini
```diff
@@ -0,0 +1,18 @@
+[binaries]
+c       = '/usr/local/lib/android/sdk/ndk-bundle/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi28-clang'
```
trufae, line 2: same

sys/ios-cydia.sh
trufae: ideally we shuoldnt be distributing this, mainly because it comes from apple, so not legal and this is why its downloaded from a different source. ideally we shouldnt depend on having this tarball around

sys/ios-cydia.sh
```diff
@@ -18,9 +18,13 @@ export BUILD=1
...
+		set -e
+		cd sys && \
+		wget http://lolcathost.org/b/ios-include.tar.gz && \
```
trufae, line 24: use wget -c unless you want it to fail if you run this outside a clean env

sys/ios-cydia.sh
```diff
@@ -18,9 +18,13 @@ export BUILD=1
...
+		set +e
+	fi
+	cd sys && tar xzvf ios-include.tar.gz
```
trufae, line 27: what about untaring it after wgeting it? this may file if running it twice

.github/meson-android-aarch64.ini
ret2libc: 1) this files are useful only for github, so they don't need to be good for all distros.
2+3) the script we are using in sys/android-build has been deprecated for a while and they suggest to use these paths instead. Again, we don't need these paths to be good for user usecases, just for github ci.

sys/ios-cydia.sh
ret2libc: the problem is that the host that have this binary can be reached 1 out of 10 times, so the script was not reliable. Do you have a better alternative (which does not involve me rewriting all the code related to IOS to not depend on this tar..)?

sys/ios-cydia.sh
ret2libc: because the way i've done it, the file is not always downloaded, but it is directly embedded within r2 git repo, so the wget is not always done.

trufae:
i'm seeing this +WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103) lately. this must be fixed, it happens because the cc dbs are not loaded in some situations and it results in errors when analyzing because the fcn->cc is NULL. we must find a way to solve that before the release. r2 should work without any SDB file loaded at all. maybe we can just assume defaults somehow, or fix the cases where it is not loaded. but the cc are important

ret2libc:
> i'm seeing this +WARNING: r_anal_cc_arg: assertion 'anal && convention' failed (line 103) lately. this must be fixed, it happens because the cc dbs are not loaded in some situations and it results in errors when analyzing because the fcn->cc is NULL. we must find a way to solve that before the release. r2 should work without any SDB file loaded at all. maybe we can just assume defaults somehow, or fix the cases where it is not loaded. but the cc are important

agree, but unrelated to this PR.

trufae:
Can someone open an issue with those cc warnings for 4.5? Im busy now

XVilka:
Yay, merged since there is a consensus already. Any further improvement are good as separate PRs.

PR #17193, opened 2020-07-01
Add Android boot image magic

enovella:


libr/magic/d/default/linux
```diff
@@ -47,11 +47,11 @@
 4086	string		SWAPSPACE2	Linux/i386 swap file (new style)
 >0x400	long		x		%d (4K pages)
 >0x404	long		x		size %d pages
->>4086	string		SWAPSPACE2	
+>>4086	string		SWAPSPACE2
 >>>1052	string		>\0		Label %s
 # Linux kernel boot images, from Albert Cahalan <acahalan@cs.uml.edu>
 # and others such as Axel Kohlmeyer <akohlmey@rincewind.chemie.uni-ulm.de>
-# and Nicol�s Lichtmaier <nick@debian.org>
```
ret2libc, line 54: what was changed in these lines?

enovella: Surely trimming whitespaces...

ret2libc: I'm honestly not sure whether some spaces were important (maybe for the system magic library?) so I would prefer to not change those things and just add the android-specific bits. Is that ok with you?

enovella: It's fine sure!

enovella: The trimmed whitespaces are definitely a leftover and they can/should be removed. The non-ASCII modified lines can be reverted.

enovella: Comparing with `Binwalk`'s output, I think we can do it way better:

```sh
$ binwalk /tmp/boot.img 

DECIMAL       HEXADECIMAL     DESCRIPTION
--------------------------------------------------------------------------------
0             0x0             Android bootimg, kernel size: 19976538 bytes, kernel addr: 0x8000, ramdisk size: 10533373 bytes, ramdisk addr: 0x1000000, product name: ""
```

Dont know if we want/need that level of detail though.

ret2libc:
Thanks!

PR #17195, opened 2020-07-01
Print a warning when the entry point is automatically set by r2 ##bin

### LINKED ISSUE #8960: Do not set entry point when the information is invalid
For example, if the entrypoint is intentionally set to 0x0, r2 will fallback to the beginning of .text section.

We should print an error and do not set entrypoint in such cases.

Also, it doesn't seem to care about static/dynamic. Static binary entrypoints are absolute address, while dynamic linked executables has relative entrypoint.

ret2libc:

**Detailed description**

...

**Test plan**

...

**Closing issues**

Closes #8960


ret2libc:
> The commit message is confusing, try to make it shorter and add ##bin

Ok

> Also, if libs have no entrypoint (which they don't have), how's r2 suposed to behave? iirc right now it just seeks to the first executable region

This PR just prints a warning, making it clear that the entry0 was automatically set to something, but it was not taken from file. Libraries, at least on my system, often have an entrypoint set in the ELF header anyway. So they should not be a problem. In case they don't have an entrypoint set, as you said r2 just sek to the first executable region. This PR still behaves the same, but it warns the user about this behaviour as well.

> Can you add a test for this?

Yes.

test/db/formats/elf/main
```diff
@@ -4,6 +4,16 @@ CMDS=?v main
 EXPECT=<<EOF
 0x401126
 EOF
+EXPECT_ERR=<<EOF
```
trufae, line 7: can we pls merge @XVilka pr to get rid of all those warnings?

test/db/formats/elf/obj
```diff
@@ -5,6 +5,8 @@ CMDS=ie~addr
 EXPECT=<<EOF
 vaddr=0x00000000 paddr=0x00000040 baddr=0x00000000 laddr=0x00000000 haddr=0x00000018 type=program
 EOF
+EXPECT_ERR=<<EOF
```
trufae, line 8: is this needed if its empty?

libr/bin/p/bin_elf.inc
```diff
@@ -406,6 +406,11 @@ static RList* entries(RBinFile *bf) {
 		ptr->vaddr = Elf_(r_bin_elf_p2v) (obj, ptr->paddr);
 		ptr->hpaddr = 0x18;  // e_entry offset in ELF header
 		ptr->hvaddr = UT64_MAX;
+
+		if (ptr->vaddr != (ut64)obj->ehdr.e_entry && Elf_(r_bin_elf_is_executable) (obj)) {
+			eprintf ("Entry point could not be determined from file, it was automatically set to 0x%" PFMT64x ".\n", ptr->vaddr);
```
trufae, line 411: make it shorter, i would say:

`Cannot determine entrypoint, using 0x%08"PFMT64x"`

libr/bin/format/elf/elf.c
```diff
@@ -4021,3 +4013,11 @@ char *Elf_(r_bin_elf_compiler)(ELFOBJ *bin) {
 	free (buf);
 	return res;
 }
+
+bool Elf_(r_bin_elf_is_executable)(ELFOBJ *bin) {
+	switch (bin->ehdr.e_type) {
```
trufae, line 4018: probably this will be shorter and more readable (and wont need to make the compiler decide which cf construction to use):

```c
const int t = bin->ehdr.e_type;
return (t == ET_EXEC || t == ET_DYN);
```

trufae:
Also imho its not correct to set an entrypoint to a binary that have no entrypoint. actually i dont underdstand the need to have a warning for something that is expected

test/db/formats/elf/obj
ret2libc: Yes, It tests that stderr is empty afaik. If you don't put it stderr is ignored, isn't it?

test/db/formats/elf/main
ret2libc: Yes @xvilka please remove those variables or use them somehow in your pr. Whatever is easier faster for you works for me.

ret2libc:
What is expected? The issue linked to this PR explains that it is not expected to see entry0 set to some value which was not set by the binary. It is r2 that automatically sets entry0 to .text.

The problem without this PR is that if you see  r2 entry0 somewhere, you know whether it is really an entrypoint or just r2 that has chosen to put entry0 somewhere

trufae:
I'm testing it on macOS and the result is not good right now (macho):
```
$ r2 /usr/local/lib/libr_util.dylib
 -- License server overloaded (ETOOMUCHCASH)
[0x00000000]> ie
[Entrypoints]
vaddr=0x00000000 paddr=0x00000000 haddr=-1 type=program

1 entrypoints

[0x00000000]>
```

i would expect it to not return any entrypoint ('ill fix macho in a separate pr), and r2 may take the first executable map as initial seek. not sure if its fair to have a fake entry0 flag

trufae:
```
[0x00000000]> iS paddr/sort,perm/str/-r-x,vaddr/cols~:3
0x00000c30
[0x00000000]> iS paddr/sort,perm/str/-r-x,vaddr/cols
[Sections]

vaddr
―――――
0x00000c30
0x000a2370
0x000a276c
0x000a2e20
0x000a7310
0x000c2ef8


[0x00000000]>
```

Maijin:
Awesome fix

PR #17197, opened 2020-07-02
sdb: Strip UTF8 BOM when creating db

kazarmy:

**Detailed description**

This pr prevents the following:

BOM_in_sdb_with_oval

i.e. a UTF8 BOM in the sdb.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/shlr/sdb/src/main.c
+++ b/shlr/sdb/src/main.c
@@ -265,9 +265,13 @@ static int createdb(const char *f, const char **args, int nargs) {
 	insertkeys (s, args, nargs, '=');
 	sdb_config (s, options);
 	for (; (line = stdin_slurp (NULL));) {
-		if ((eq = strchr (line, '='))) {
+		size_t off = 0;
+		if (line[0] == '\xef' && line[1] == '\xbb' && line[2] == '\xbf') {
+			off = 3;
+		}
+		if ((eq = strchr (line + off, '='))) {
 			*eq++ = 0;
-			sdb_disk_insert (s, line, eq);
+			sdb_disk_insert (s, line + off, eq);
 		}
 		free (line);
 	}


trufae:
Some questions here:

* Why is having such sequence in a binary file is a problem?
* Why apply this fix in r2 insterad of the sdb repo directly?

ret2libc:
Yeah this should be done in https://github.com/radareorg/sdb . Closing this, please move it there.

trufae:
My guess is that he pushed this change in here to ensure r2 suite was not breaking after the change

kazarmy:
> Why is having such sequence in a binary file is a problem?

It's a problem because it attaches itself to the following characters (it's a bona fide unicode char after all) and in this case, it causes the following to fail:

https://github.com/radareorg/radare2/blob/cdd039a38ae9d28b81fe820a7c85289085d14952/libr/anal/cc.c#L161

bringing down at least type propagation with it.

> Why apply this fix in r2 insterad of the sdb repo directly?

Because r2 is an sdb client and it doesn't pull from the sdb repo automatically? I'm honestly confused here. Yes, the fix should be in the sdb repo as well.

> My guess is that he pushed this change in here to ensure r2 suite was not breaking after the change

That's one reason.

ret2libc:
> > Why apply this fix in r2 insterad of the sdb repo directly?
> 
> Because r2 is an sdb client and it doesn't pull from the sdb repo automatically? I'm honestly confused here. Yes, the fix should be in the sdb repo as well.

Things should land first in sdb, as it is mainly an independent project (:D) and then you should get the changes by using `make sdb-sync `.

kazarmy:
oh ok https://github.com/radareorg/sdb/pull/209

PR #17198, opened 2020-07-02
Fix rlang scripts and cpipe when non standard paths are used ##lang

ret2libc:

**Detailed description**

Before this patch, radare2 was checking the extension defined in RLangPlugin.ext, so even if the help says
```
[0x00000000]> #!
c: (LGPL) C language extension
cpipe: (LGPL) r2pipe scripting in C
vala: (LGPL) Vala language extension
rust: (MIT) Rust language extension
zig: (MIT) Zig language extension
spp: (MIT) SPP template programs
pipe: (LGPL) Use #!pipe node script.js
lib: (LGPL) Load libs directly into r2
python: (LGPL) Python language extension
[0x00000000]> #?
Usage #!interpreter [<args>] [<file] [<<eof]
|  #                     comment - do nothing
|  #!                    list all available interpreters
|  #!python              run python commandline
|  #!python foo.py       run foo.py python script (same as '. foo.py')
|  #!python arg0 a1 <<q  set arg0 and arg1 and read until 'q'
```
some plugins did not work as expected. In particular, those that had a .ext != .name (like python in rlang, cpipe, rust). Before this PR you had to do `#!rs` instead of `#!rust` to execute the rust interpreter (or `#!c2` instead of `#!cpipe`, etc.)

Moreover, it fixes the compilation/execution of cpipe programs by setting PKG_CONFIG_PATH properly and ensuring LD_LIBRARY_PATH is correctly set when executing the program.

**Test plan**

Try `#!` commands. In particular, `#!rust` should execute the rust interpreter and `#!cpipe` the C-pipe one. Moreover, `#!cpipe` should work even when you install radare2 libs in non standard dirs (e.g. i have them installed in ~/.local/lib64 on my system and i don't have that dir in LD_LIBRARY_PATH).

**Closing issues**

Closes https://github.com/radareorg/radare2/issues/17239


test/db/cmd/cmd_hash
```diff
@@ -40,4 +40,28 @@ CMDS=<<EOF
...
+EOF
+EXPECT=<<EOF
+Hello
+EOF
+RUN
+
+NAME=#! with ext != name
+FILE=-
+CMDS=<<EOF
+"?e #include <stdio.h>\nint main(){printf(\"Hello\\n\");}" > .tmp-hashextname.c
```
trufae, line 60: what about adding those .c files instead of creating them at test time?

ret2libc: as you prefer actually. i don't have strong opinions, this way was just faster for me because i don't really care about the content of the file, i just wanted to make sure things compiled properly.

ret2libc:
@trufae @xarkes @pelijah @XVilka do `#!cpipe` and `!c` only work on Linux? :(

condret:
afaik it also works on BrokenBSD (MacOS) and windows with wsl

ret2libc:
> afaik it also works on BrokenBSD (MacOS) and windows with wsl

Yes, I can see it working on MacOS, as the macos-build is green.
And windows with WSL is just linux for us, AFAIK.

trufae:
may be worth adding support to cl/vs for windows if available or show a better error if no compiler is available. another option would be to embed tcc inside this plugin so we can fallback to it if user wants or no option is available. frida embeds it

libr/lang/p/cpipe.c
```diff
@@ -38,13 +37,17 @@ static int lang_cpipe_file(RLang *lang, const char *file) {
...
+		" $(PKG_CONFIG_PATH=%s pkg-config --cflags --libs r_socket)",
+		cc, file, libpath, libname, R2_LIBDIR "/pkgconfig");
 	free (cc);
 	if (r_sandbox_system (buf, 1) == 0) {
+		char *o_ld_path = r_sys_getenv ("LD_LIBRARY_PATH");
+		r_sys_setenv ("LD_LIBRARY_PATH", R2_LIBDIR);
 		char *binfile = r_str_newf ("%s/bin%s", libpath, libname);
 		lang_pipe_run (lang, binfile, -1);
 		r_file_rm (binfile);
+		free (o_ld_path);
```
trufae, line 49: move this free to the line below or you get an UAF

libr/lang/p/cpipe.c
```diff
@@ -38,13 +37,17 @@ static int lang_cpipe_file(RLang *lang, const char *file) {
 		cc = strdup ("gcc");
 	}
 	char *buf = r_str_newf ("%s %s -o %s/bin%s"
```
trufae, line 39: can you use the shell escape function to avoid code injection issues? r2 can be installed on paths with spaces which may make this line fail

libr/lang/p/c.c
```diff
@@ -42,7 +42,6 @@ static int lang_c_file(RLang *lang, const char *file) {
 		libpath = ".";
 		libname = name;
 	}
-	r_sys_setenv ("PKG_CONFIG_PATH", R2_LIBDIR"/pkgconfig");
 	p = strstr (name, ".c");
 	if (p) {
 		*p=0;
```
trufae, line 47: spaces around =

libr/lang/p/c.c
```diff
@@ -51,8 +50,9 @@ static int lang_c_file(RLang *lang, const char *file) {
 	if (!cc || !*cc) {
```
trufae, line 50: strisempty

libr/lang/p/cpipe.c
ret2libc: Oooops

XVilka:
@ret2libc did you check that both `#!python qwe.py` and `. qwe.py` work with the same result?

ret2libc:
> @ret2libc did you check that both `#!python qwe.py` and `. qwe.py` work with the same result?

Yes.

libr/lang/p/cpipe.c
```diff
@@ -84,3 +99,10 @@ static RLangPlugin r_lang_plugin_cpipe = {
 	.fini = NULL,
 	.run_file = (void*)lang_cpipe_file,
 };
+#else
+#ifdef _MSC_VER
+#pragma message("Warning: cpipe RLangPlugin is not implemented on this platform")
```
trufae, line 104: what about implementing support for windows instead of exposing that this is not supported? the only reason for that is imho the system call to gcc. which may be replaced with the msvc compiler. the rest of the code must be generic and shuold work fine

ret2libc: This was never working, as you can see from https://github.com/radareorg/radare2/pull/17198/files#diff-4ff9c2d98fc735f1bc2745b65c380cf4L48 . Honestly, working with windows compiler is a real pain and you also have to support PKG_CONFIG_PATH and LD_LIBRARY_PATH . IMHO this PR is a big *fix*, it doesn't intend to be a new feature, as it would be adding support for windows.

XVilka: I think it's also out of the scope of this PR.

PR #17199, opened 2020-07-02
Various fixes based on fuzz bins for rabin2

wargio:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/bin/format/dex/dex.c
+++ b/libr/bin/format/dex/dex.c
@@ -17,8 +17,10 @@ char* r_bin_dex_get_version(RBinDexObj *bin) {
 void r_bin_dex_free(RBinDexObj *dex) {
 	size_t i;
 	struct dex_header_t *dexhdr = &dex->header;
-	for (i = 0; i < dexhdr->strings_size; i++) {
-		free (dex->cal_strings[i]);
+	if (dex->cal_strings) {
+		for (i = 0; i < dexhdr->strings_size; i++) {
+			free (dex->cal_strings[i]);
+		}
 	}
 	free (dex->cal_strings);
 	free (dex);
--- a/libr/bin/format/dex/dex.c
+++ b/libr/bin/format/dex/dex.c
@@ -15,9 +15,9 @@ char* r_bin_dex_get_version(RBinDexObj *bin) {
 }
 
 void r_bin_dex_free(RBinDexObj *dex) {
-	size_t i;
 	struct dex_header_t *dexhdr = &dex->header;
 	if (dex->cal_strings) {
+		size_t i;
 		for (i = 0; i < dexhdr->strings_size; i++) {
 			free (dex->cal_strings[i]);
 		}
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -3034,6 +3034,9 @@ static RBinElfSymbol* get_symbols_from_phdr(ELFOBJ *bin, int type) {
 
 	addr_sym_table = Elf_(r_bin_elf_v2p) (bin, bin->dyn_info.dt_symtab);
 	sym_size = bin->dyn_info.dt_syment;
+	if (sym_size < 1) {
+		goto beach;
+	}
 
 	//since ELF doesn't specify the symbol table size we may read until the end of the buffer
 	nsym = (bin->size - addr_sym_table) / sym_size;
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -3034,7 +3034,7 @@ static RBinElfSymbol* get_symbols_from_phdr(ELFOBJ *bin, int type) {
 
 	addr_sym_table = Elf_(r_bin_elf_v2p) (bin, bin->dyn_info.dt_symtab);
 	sym_size = bin->dyn_info.dt_syment;
-	if (sym_size < 1) {
+	if (!sym_size) {
 		goto beach;
 	}
 
--- a/libr/bin/format/pyc/marshal.c
+++ b/libr/bin/format/pyc/marshal.c
@@ -183,7 +183,7 @@ static pyc_object *get_long_object(RBuffer *buffer) {
 
 	st32 ndigits = get_st32 (buffer, &error);
 	if (ndigits < -SIZE32_MAX) {
-		eprintf ("bad marshal data (long size out of range)");
+		eprintf ("bad marshal data (long size out of range)\n");
 		return NULL;
 	}
 	if (error) {
@@ -540,7 +540,7 @@ static pyc_object *get_tuple_object(RBuffer *buffer) {
 
 	n = get_ut32 (buffer, &error);
 	if (n > SIZE32_MAX) {
-		eprintf ("bad marshal data (tuple size out of range)");
+		eprintf ("bad marshal data (tuple size out of range)\n");
 		return NULL;
 	}
 	if (error) {
@@ -561,7 +561,7 @@ static pyc_object *get_list_object(RBuffer *buffer) {
 
 	n = get_ut32 (buffer, &error);
 	if (n > SIZE32_MAX) {
-		eprintf ("bad marshal data (list size out of range)");
+		eprintf ("bad marshal data (list size out of range)\n");
 		return NULL;
 	}
 	if (error) {
@@ -622,7 +622,7 @@ static pyc_object *get_set_object(RBuffer *buffer) {
 
 	n = get_ut32 (buffer, &error);
 	if (n > SIZE32_MAX) {
-		eprintf ("bad marshal data (set size out of range)");
+		eprintf ("bad marshal data (set size out of range)\n");
 		return NULL;
 	}
 	if (error) {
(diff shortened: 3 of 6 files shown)


libr/bin/format/dex/dex.c
```diff
@@ -17,8 +17,10 @@ char* r_bin_dex_get_version(RBinDexObj *bin) {
 void r_bin_dex_free(RBinDexObj *dex) {
 	size_t i;
 	struct dex_header_t *dexhdr = &dex->header;
-	for (i = 0; i < dexhdr->strings_size; i++) {
-		free (dex->cal_strings[i]);
+	if (dex->cal_strings) {
+		for (i = 0; i < dexhdr->strings_size; i++) {
```
trufae, line 21: Move i definition here

libr/bin/format/pyc/marshal.c
```diff
@@ -540,7 +540,7 @@ static pyc_object *get_tuple_object(RBuffer *buffer) {
 
 	n = get_ut32 (buffer, &error);
 	if (n > SIZE32_MAX) {
```
trufae, line 542: Size32 max wtf? Wheres this define coming from? We have st32_max instead

libr/bin/format/elf/elf.c
```diff
@@ -3028,6 +3028,9 @@ static RBinElfSymbol* get_symbols_from_phdr(ELFOBJ *bin, int type) {
 
 	addr_sym_table = Elf_(r_bin_elf_v2p) (bin, bin->dyn_info.dt_symtab);
 	sym_size = bin->dyn_info.dt_syment;
+	if (sym_size < 1) {
```
trufae, line 3031: Can simsize be negative? If so fix it, otherwise just check for == 0

libr/bin/format/pyc/marshal.c
```diff
@@ -1046,10 +1046,16 @@ static pyc_object *get_object(RBuffer *buffer) {
...
 	case TYPE_CODE_v0:
 		ret = get_code_object (buffer);
+		if (!ret) {
+			return NULL;
+		}
 		ret->type = TYPE_CODE_v0;
 		break;
 	case TYPE_CODE_v1:
 		ret = get_code_object (buffer);
+		if (!ret) {
```
trufae, line 1056: What about if (ret) so it will return something, even partial at least?

libr/bin/format/pyc/marshal.c
```diff
@@ -1152,6 +1158,9 @@ static bool extract_sections_symbols(pyc_object *obj, RList *sections, RList *sy
 	if (!cobj->name->data) {
 		return false;
 	}
+	if (!cobj->consts) {
```
trufae, line 1161: Collapse into a single line

libr/util/vector.c
```diff
@@ -19,10 +19,15 @@
...
 		size_t new_capacity = next_capacity; \
-		void **new_a = realloc (vec->a, vec->elem_size * new_capacity); \
+		size_t new_capacity_size = vec->elem_size * new_capacity; \
+		void **new_a = realloc (vec->a, new_capacity_size); \
 		if (!new_a) { \
 			return NULL; \
 		} \
+		if (vec->capacity < new_capacity) { \
+			size_t old_capacity_size = vec->elem_size * vec->capacity; \
+			memset(&new_a[vec->capacity], 0, new_capacity_size - old_capacity_size); \
```
trufae, line 29: Add a space before (. A review from @thestr4ng3r may be good here

libr/bin/format/pyc/marshal.c
ret2libc: do you mean `if (!cobj->consts) { return false; }`??

libr/util/vector.c
ret2libc: why is this necessary @wargio ? AFAIK you should not access the elements after `vec->len`, so it doesn't really matter what is the value inside those elements.

wargio: ok, the binary i have does something that should not happen. essentially resizes the vector (old < new) and then fails to add stuff. the `realloc` does not clean the memory when you resize with a bigger chunk of memory and the vector free fails to free garbage pointer (obviously).

libr/bin/format/pyc/marshal.c
wargio: why? what about the others? maybe i should use r2 macros for this. if_null_return_value.

libr/bin/format/pyc/marshal.c
wargio: the issue is on the null deref. i need to have a return NULL anyway. if you do not like this, i can use r2 macros.

libr/bin/format/elf/elf.c
wargio: i'll fix this since is `ut32`

libr/util/vector.c
```diff
@@ -233,6 +233,22 @@ R_API RPVector *r_pvector_new(RPVectorFree free) {
 	return v;
 }
 
+R_API RPVector *r_pvector_new_with_len(RPVectorFree free, size_t length) {
+	RPVector *v = R_NEW (RPVector);
```
ret2libc, line 237: use `r_pvector_new` instead of dupping its code.

libr/bin/format/pyc/marshal.c
```diff
@@ -182,8 +183,8 @@ static pyc_object *get_long_object(RBuffer *buffer) {
 	char digist2hex[] = "0123456789abcdef";
 
 	st32 ndigits = get_st32 (buffer, &error);
-	if (ndigits < -SIZE32_MAX) {
-		eprintf ("bad marshal data (long size out of range)");
+	if (ndigits < -ST32_MAX) {
```
trufae, line 186: whats the point of this chewck? if ndigits is st32 ..

libr/bin/format/pyc/marshal.c
```diff
@@ -1046,11 +1047,11 @@ static pyc_object *get_object(RBuffer *buffer) {
 		break;
 	case TYPE_CODE_v0:
 		ret = get_code_object (buffer);
-		ret->type = TYPE_CODE_v0;
+		if (ret) { ret->type = TYPE_CODE_v0; }
 		break;
 	case TYPE_CODE_v1:
 		ret = get_code_object (buffer);
-		ret->type = TYPE_CODE_v1;
+		if (ret) { ret->type = TYPE_CODE_v1; }
```
trufae, line 1054: multiline pls

libr/bin/format/pyc/marshal.c
```diff
@@ -1139,19 +1140,15 @@ static bool extract_sections_symbols(pyc_object *obj, RList *sections, RList *sy
 	RListIter *i = NULL;
 
 	//each code object is a section
-	if (!obj || (obj->type != TYPE_CODE_v1 && obj->type != TYPE_CODE_v0)) {
-		return false;
-	}
+	if_true_return (!obj || (obj->type != TYPE_CODE_v1 && obj->type != TYPE_CODE_v0), false);
```
trufae, line 1143: wasn't this suposed to be a r_return_val_if_fail?

libr/bin/format/pyc/marshal.c
```diff
@@ -1139,19 +1140,15 @@ static bool extract_sections_symbols(pyc_object *obj, RList *sections, RList *sy
...
-		return false;
-	}
-	if (cobj->name->type != TYPE_ASCII && cobj->name->type != TYPE_STRING && cobj->name->type != TYPE_INTERNED) {
-		return false;
-	}
-	if (!cobj->name->data) {
-		return false;
-	}
+
+	if_true_return (!cobj || !cobj->name, false);
```
trufae, line 1147: same

libr/bin/format/pyc/marshal.c
wargio: don't ask me. i'm not the one that has written the code.

libr/bin/format/pyc/marshal.c
wargio: you asked me previously to have inline if.

libr/bin/format/pyc/marshal.c
wargio: no. i decided to not use them because they can be disabled completely and these checks must be in place.

libr/bin/format/pyc/marshal.c
wargio: see above.

libr/bin/format/pyc/marshal.c
wargio: i removed the condition since it was weird.

libr/bin/format/pyc/marshal.c
wargio: to be clear, this conditions is needed. so i do not want to use something that can be removed at compilation time due macros.

```
#if R_CHECKS_LEVEL == 0

#define r_return_if_fail(expr) do { ; } while(0)
#define r_return_val_if_fail(expr, val) do { ; } while(0)
```

ret2libc: absolutely! r_return_ are supposed to be used for checking programmatic errors (e.g. misuing an API, or things like that). In this case it seems more a runtime error, so it is something that should always be properly checked!

ret2libc:
@trufae could you check whether the changes you requested are ok now? IMO this is good now.

libr/bin/format/pyc/marshal.c
trufae: well, what i dont like about this approach is that this `if_true_return` thing is defined locally here, where in the rest of the codebase we use the multiline ifs. im happy to introduce this macro r2-wide, but not like this https://github.com/radareorg/radare2/pull/17199/files#diff-b1ff7ca2270551aa1b19a80b3116c91bR9

i'm fine to merge this too. because this pr fixes several more things

PR #17200, opened 2020-07-02
Correct output value for AES and private key search

sylvainpelissier:

**Detailed description**
If the function **r_search_privkey_update** return 0 when nothing is found, the search will stop after one block because the return value match with **r_search_hit_new** return value.

**Test plan**
I will update test binary in the repository to test that properly.


ret2libc:
It would be great to have unit tests for the RSearch API

PR #17206, opened 2020-07-04
Fix tcache parsing

### LINKED ISSUE #17205: Wrong tcache parsing between different glibc version
# Work environment

| OS/arch/bits (mandatory)                             | Arch Linux x86-64
| Architecture/bits of the file (mandatory)            | x86-64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.5.0-git 24844 @ linux-x86-64 git.4.4.0-353-g114a5a594
commit: 114a5a59438be07e5ffea8f757ab0df242f0be41 build: 2020-07-03__17:08:09

Currently the tcache is parsed using the tcache_perthread_struct:  https://github.com/radareorg/radare2/blob/80202e6ad5b45efc3de5f2210019eb948dd797b6/libr/include/r_heap_glibc.h#L148
in: https://github.com/radareorg/radare2/blob/80202e6ad5b45efc3de5f2210019eb948dd797b6/libr/core/linux_heap_glibc.c#L877

This works fine for glibc version >= 2.30,  however in 2.29 and in precedent versions the struct is defined as: 
https://elixir.bootlin.com/glibc/glibc-2.29/source/malloc/malloc.c#L2916

With char instead of u_int16_t for `count`

### Expected behavior
```
$ r2 -e dbg.profile=profile.rr2 -i script.rr2 -d ./tcache-2.27

Tcache main arena @ 0x7fd2785589e0
bin : 0, items : 6, fd :0xd02740->0xd02720->0xd02700->0xd026e0->0xd026c0->0xd026a0
```

### Actual behavior

```
$ r2 -e dbg.profile=profile.rr2 -i script.rr2 -d ./tcache-2.27

Do you want to print 4 lines? (y/N)
bin :33, items :351, fd :0xfffffffffffffff0->0xffffffffffffffef->0xfffffffffffffff0->0xffffffffffffffef->0xfffffffffffffff0->0xffffffffffffffef->0xfffffffffffffff0->0xffffffffffffffef->0xfffffffffffffff0->0xffffffffffffffef
[...] more output
```

I provide a zip with a simple program that malloc and free some memory (to put chunk inside tcache),  to use the version for glibc2.27 use:

```
$ r2 -e dbg.profile=profile.rr2 -i script.rr2 -d tcache-2.27
```

And it should work, if not try:

```
$ gcc tcache.c -o tcache-2.27 -no-pie
$ patchelf --set-interpreter ./ld-2.27.so tcache-2.27
$ # RUN
$ LD_PRELOAD=./libc-2.27.so ./tcache-2.27
```

I would like to work on it, however I don't know what approach should I use to solve the problem.
Currently we are reading the memory where the tcache_perthread_struct is, and we pass the memory values inside our tcache_perthread_struct, which makes sense if we have only one type of tcache_perthread_struct (but we need two different type).

So I thought about adding a new struct inside r_heap_glibc.h:

```c
typedef struct r_tcache_perthread_struct_old_32 {
	ut8 counts[TCACHE_MAX_BINS];
	ut32 entries[TCACHE_MAX_BINS];
} RHeapTcacheOld_32;

typedef struct r_tcache_perthread_struct_old_64 {
	ut8 counts[TCACHE_MAX_BINS];
	ut64 entries[TCACHE_MAX_BINS];
} RHeapTcacheOld_64;
```

And then create two function inside linux_heap_glibc.c:

```
static void GH(parse_tcache_old)(RCore *core, GH(RHeapTcacheOld) *tcache_heap);
static void GH(parse_tcache)(RCore *core, GH(RHeapTcache) *tcache_heap);
```

The two function would have the same exact code so it wouldn't be a super clean solution. The codeof the function would be this `for`: https://github.com/radareorg/radare2/blob/80202e6ad5b45efc3de5f2210019eb948dd797b6/libr/core/linux_heap_glibc.c#L882

Another approach is to not parse all the memory at once, but parsing the tcache step by step like gef is doing: 
https://github.com/hugsy/gef/blob/1d5cffd459bc25e0b02657d0300cb51609b8df6c/gef.py#L6425 and https://github.com/hugsy/gef/blob/1d5cffd459bc25e0b02657d0300cb51609b8df6c/gef.py#L722

In this way they only check the glibc version twice which is not a bad idea.

Another approach would be to use `union` and implement some getter and setter, example:

```c
#include <stdio.h>
#include <sys/types.h>
#include <unistd.h>
#include <stdlib.h>
#include <time.h>
#include <string.h>
#include <sys/types.h>

enum tcache_type {NEW, OLD};

typedef struct r_heap_old {
	char counts[10];
	void* entries[10];
} RHeapTcacheOld;

typedef struct r_heap_t {
	u_int16_t counts[10];
	void* entries[10];
} RHeapTcache;

typedef struct {
	enum tcache_type type;
	union {
		RHeapTcache heap_tcache;
		RHeapTcacheOld heap_tcache_old;
	} RHeapUnion;
} RHeap;

int getCount(RHeap *heap, int index) {
	if (heap->type == OLD) {
		return heap->RHeapUnion.heap_tcache_old.counts[index];
	} else {
		return heap->RHeapUnion.heap_tcache.counts[index];
	}
}

void setCount(RHeap *heap, int index, int value) {
	if (heap->type == OLD) {
		heap->RHeapUnion.heap_tcache_old.counts[index] = value;
	} else {
		heap->RHeapUnion.heap_tcache.counts[index] = value;
	}
}

	RHeap *heap = malloc(sizeof(RHeap));
	heap->type = NEW;
	setCount(heap, 0, 10);
	printf("get: %d\n", getCount(heap, 0));
	return 0;
}
```

What approach should I use, or do you have any other suggestion?

[test-tcache.zip](https://github.com/radareorg/radare2/files/4872992/test-tcache.zip)

meowmeowxw:

**Detailed description**

This change is needed to parse correctly the tcache struct (at the moment for `dmht`), discussed in https://github.com/radareorg/radare2/issues/17205

**Test plan**

The test are linked and explained in the issue

**Issues**

Close #17205

### DIFF
--- a/libr/core/linux_heap_glibc.c
+++ b/libr/core/linux_heap_glibc.c
@@ -10,15 +10,18 @@
 #undef GH
 #undef GHT
 #undef GHT_MAX
+#undef read_le
 
 #if HEAP32
 #define GH(x) x##_32
 #define GHT ut32
 #define GHT_MAX UT32_MAX
+#define read_le(x) r_read_le##32(x)
 #else
 #define GH(x) x##_64
 #define GHT ut64
 #define GHT_MAX UT64_MAX
+#define read_le(x) r_read_le##64(x)
 #endif
 
 /**
@@ -80,6 +83,7 @@ static bool GH(is_tcache)(RCore *core) {
 	}
 	if (fp) {
 		v = r_num_get_float (NULL, fp + 5);
+		core->dbg->glibc_version = (int) (v * 100);
 	}
 	return (v > 2.25);
 }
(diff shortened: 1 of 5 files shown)


libr/core/linux_heap_glibc.c
```diff
@@ -849,35 +850,38 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
-	GH(get_brks) (core, &brk_start, &brk_end);
-	GHT tcache_tmp = GHT_MAX, tcache_start = GHT_MAX;
+static void GH(parse_tcache)(RCore *core, GH(RHeapTcache) *tcache_heap) {
+	GHT tcache_fd = GHT_MAX;
+	GHT tcache_tmp = GHT_MAX;
 	RConsPrintablePalette *pal = &r_cons_singleton ()->context->pal;
+	int i;
+	for (i = 0; i < TCACHE_MAX_BINS; i++) {
+		if (tcache_heap->counts[i] > 0) {
+			PRINT_GA("bin :");
```
trufae, line 860: space before ( and use `size_t i ` instead of int

libr/core/linux_heap_glibc.c
```diff
@@ -849,35 +850,38 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
-	if (!tcache_heap) {
-		return;
+			PRINTF_BA("0x%"PFMT64x, (GHT)(tcache_heap->entries[i]) - GH(HDR_SZ));
+			if (tcache_heap->counts[i] > 1) {
+				tcache_fd = (GHT)tcache_heap->entries[i];
+				int n;
+				for(n=1; n < tcache_heap->counts[i]; n++) {
+					(void)r_io_read_at (core->io, tcache_fd, (ut8 *)&tcache_tmp, sizeof (ut64));
+					PRINTF_BA("->0x%"PFMT64x, tcache_tmp - TC_HDR_SZ);
+					tcache_fd = tcache_tmp;
```
trufae, line 873: add an (ut64) cast

libr/core/linux_heap_glibc.c
```diff
@@ -849,35 +850,38 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
-		return;
-	}
-
-	GH(RHeapTcache) *tcache_heap = R_NEW0 (GH(RHeapTcache));
-	if (!tcache_heap) {
-		return;
+			PRINTF_BA("0x%"PFMT64x, (GHT)(tcache_heap->entries[i]) - GH(HDR_SZ));
+			if (tcache_heap->counts[i] > 1) {
+				tcache_fd = (GHT)tcache_heap->entries[i];
+				int n;
```
trufae, line 869: size_t and respect spacings

libr/core/linux_heap_glibc.c
trufae, line 873: use the endian apis to read numbers from memory. .this code will only work on machines of the same endian as the target bin

libr/core/linux_heap_glibc.c
```diff
@@ -900,6 +904,52 @@ static void GH(print_tcache_instance)(RCore *core, GHT m_arena, MallocState *mai
 			PRINT_BA ("\n");
 		}
 	}
+}
+
+static void GH(print_tcache_instance)(RCore *core, GHT m_arena, MallocState *main_arena) {
+	if (!core || !core->dbg || !core->dbg->maps) {
```
trufae, line 910: r_return_if_fail

libr/core/linux_heap_glibc.c
```diff
@@ -900,6 +904,52 @@ static void GH(print_tcache_instance)(RCore *core, GHT m_arena, MallocState *mai
...
+
+	double glibc_version = core->dbg->glibc_version;
+	GH(RHeapTcache) *tcache_heap;
+	GH(RHeapTcacheOld) *tcache_heap_old;
+	if (glibc_version > 2.29) {
+		tcache_heap = R_NEW0(GH(RHeapTcache));
+		if (!tcache_heap) {
+			return;
+		}
+		(void)r_io_read_at (core->io, tcache_start, (ut8 *)tcache_heap, sizeof (GH(RHeapTcache)));
```
trufae, line 938: check return of read.at

libr/core/linux_heap_glibc.c
```diff
@@ -900,6 +904,52 @@ static void GH(print_tcache_instance)(RCore *core, GHT m_arena, MallocState *mai
...
+	} else {
+		tcache_heap_old = R_NEW0 (GH(RHeapTcacheOld));
+		if (!tcache_heap_old) {
+			return;
+		}
+		(void)r_io_read_at (core->io, tcache_start, (ut8 *)tcache_heap_old, sizeof (GH(RHeapTcacheOld)));
```
trufae, line 944: check return of read.at

libr/debug/debug.c
```diff
@@ -386,6 +386,7 @@ R_API RDebug *r_debug_new(int hard) {
 	dbg->q_regs = NULL;
 	dbg->call_frames = NULL;
 	dbg->main_arena_resolved = false;
+	dbg->glibc_version = 2.31; /* default version ubuntu 20 */
```
trufae, line 389: i see this double/float solution a bit tricky and could be probelmatic on some situations. i see different options. and having an api to handle `r_semver_new (2, 31, 0)` whjich may result in a struct or packed values into ut32 dword. 

r_semver_gt() to compare < or > between them.

libr/core/linux_heap_glibc.c
trufae: btw, how this would work on 32bit systems?

meowmeowxw:
@trufae is the code duplication of the functions parse_tcache and parse_tcache_old  ok or should I use an alternative way?

libr/core/linux_heap_glibc.c
meowmeowxw: PFMT64x works fine with 32 bit system, I changed the `sizeof (ut64)`  to `sizeof (GHT)` and then I added a macro to convert the bytes in little endian using `r_read_le`.

libr/include/r_heap_glibc.h
```diff
@@ -150,6 +150,17 @@ typedef struct r_tcache_perthread_struct_64 {
 	ut64 entries[TCACHE_MAX_BINS];
 } RHeapTcache_64;
 
+typedef struct r_tcache_perthread_struct_old_32 {
```
ret2libc, line 153: as said in the issue, I would give this a more descriptive name than "old/new". "pre_227_" seems better to me.

libr/debug/debug.c
ret2libc: I agree the usage of double to store the version seem hacky. Better to replace it with something else.

libr/core/linux_heap_glibc.c
```diff
@@ -849,56 +853,114 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
+			PRINTF_BA ("%2d",tcache_heap->counts[i]);
+			PRINT_GA (", fd :");
+
+			PRINTF_BA ("0x%"PFMT64x, (GHT)(tcache_heap->entries[i]) - GH(HDR_SZ));
+			if (tcache_heap->counts[i] > 1) {
+				tcache_fd = (GHT) tcache_heap->entries[i];
+				size_t n;
+				for (n = 1; n < tcache_heap->counts[i]; n++) {
+					bool r = r_io_read_at (core->io, tcache_fd, (ut8 *)&tcache_tmp, sizeof (GHT));
+					if (r) {
```
ret2libc, line 875: just do `if (!r) { break } ` and then keep the "main"code outside the if.

libr/core/linux_heap_glibc.c
```diff
@@ -849,56 +853,114 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
+						break;
+					}
+				}
+			}
+			PRINT_BA ("\n");
+		}
+	}
+}
+
+static void GH(parse_tcache_old)(RCore *core, GH(RHeapTcacheOld) *tcache_heap) {
```
ret2libc, line 889: this is a lot of duplicated code. We should find a way to write it just one time, otherwise it will be easy in the future to partially modify one code but not the other. Maybe use a macro?

libr/core/linux_heap_glibc.c
```diff
@@ -849,56 +853,114 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
 	if (brk_start == GHT_MAX || brk_end == GHT_MAX || initial_brk == GHT_MAX) {
 		eprintf ("No heap section\n");
 		return;
 	}
 
-	GH(RHeapTcache) *tcache_heap = R_NEW0 (GH(RHeapTcache));
-	if (!tcache_heap) {
-		return;
+	double glibc_version = core->dbg->glibc_version;
+	GH(RHeapTcache) *tcache_heap;
```
ret2libc, line 943: Mh.. I still think having an union would have been easier to implement/read :( it would have avoided having 2 variables to deal with.

libr/core/linux_heap_glibc.c
```diff
@@ -849,56 +853,114 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
+	GH(RHeapTcacheOld) *tcache_heap_old;
+	if (glibc_version > 2.29) {
```
ret2libc, line 945: use a define for this 2.29 version.

ret2libc: also, if you were using a union, you could have done something like:
```C
tcache_heap = read_tcache(core->io, tcache_start);
```
and inside `read_tcache` you would have read the right structure. However, everywhere else you would just have used one variable that would have hidden the internal details.

libr/core/linux_heap_glibc.c
```diff
@@ -849,56 +853,114 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
-				int n;
-				for(n=1; n < tcache_heap->counts[i]; n++) {
-					(void)r_io_read_at (core->io, tcache_fd, (ut8 *)&tcache_tmp, sizeof (ut64));
-					PRINTF_BA("->0x%"PFMT64x, tcache_tmp - TC_HDR_SZ);
-					tcache_fd = tcache_tmp;
-				}
-			}
-			PRINT_BA ("\n");
-		}
+	if (glibc_version > 2.29) {
```
ret2libc, line 960: same here. a union would have allowed to have `GH (parse_tcache) (core, tcache_heap)` and just deal with the differences inside the function and not also outside.

libr/core/linux_heap_glibc.c
meowmeowxw: I'm working on union right now, I should also modify the code of `dmh` so it's not a clean solution to duplicate the code another time.

libr/core/linux_heap_glibc.c
meowmeowxw: Ok

libr/include/r_heap_glibc.h
meowmeowxw: Ok I prefer old and put a comment on the version, but I will change it to pre_230 since it's for versions <= 2.29

libr/core/linux_heap_glibc.c
```diff
@@ -10,15 +10,18 @@
 #undef GH
 #undef GHT
 #undef GHT_MAX
+#undef r_read_le
```
trufae, line 13: why redefine an r2-wise macro for something local? use a different name like `read_le` instead

libr/core/linux_heap_glibc.c
```diff
@@ -849,63 +853,123 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
 	}
 }
 
-static void GH(print_tcache_instance)(RCore *core, GHT m_arena, MallocState *main_arena) {
-	if (!core || !core->dbg || !core->dbg->maps) {
-		return;
+static GH(RTcache)* GH(tcache_new) (tcache_type type) {
+	GH(RTcache) *tcache = R_NEW0 (GH(RTcache));
```
trufae, line 857: add a space before every (

libr/core/linux_heap_glibc.c
```diff
@@ -849,63 +853,123 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
+	if (tcache->type == NEW) {
+		tcache->RHeapTcache.heap_tcache = R_NEW0(GH(RHeapTcache));
+	} else {
+		tcache->RHeapTcache.heap_tcache_old = R_NEW0(GH(RHeapTcacheOld));
+	}
+	return tcache;
+}
+
+static void GH(tcache_free) (GH(RTcache)* tcache) {
+	free(tcache);
```
trufae, line 868: missing space

libr/core/linux_heap_glibc.c
```diff
@@ -849,63 +853,123 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
+}
+
+static bool GH(tcache_read) (RCore *core, GHT tcache_start, GH(RTcache)* tcache) {
+	bool r = false;
```
trufae, line 872: return with ternary operator, but i dont see which is the difference between the true and false blocks

libr/core/linux_heap_glibc.c
```diff
@@ -849,63 +853,123 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
+	if (tcache->type == NEW) {
+		r = r_io_read_at (core->io, tcache_start, (ut8 *)tcache->RHeapTcache.heap_tcache, sizeof (GH(RHeapTcache)));
+	} else {
+		r = r_io_read_at (core->io, tcache_start, (ut8 *)tcache->RHeapTcache.heap_tcache, sizeof (GH(RHeapTcache)));
+
+	}
+	return r;
+}
+
+static int GH(tcache_get_count) (GH(RTcache)* tcache, int index) {
```
trufae, line 882: return ternary operator

libr/core/linux_heap_glibc.c
```diff
@@ -849,63 +853,123 @@ void GH(print_heap_fastbin)(RCore *core, GHT m_arena, MallocState *main_arena, G
...
+	if (tcache->type == NEW) {
+		return tcache->RHeapTcache.heap_tcache->counts[index];
+	} else {
+		return tcache->RHeapTcache.heap_tcache_old->counts[index];
+	}
+}
+
+static GHT GH(tcache_get_entry) (GH(RTcache)* tcache, int index) {
```
trufae, line 890: same

libr/debug/debug.c
trufae: We need the r_semver api

trufae: but this must be done in a separate PR and this PR will be blocked until then. Who will proceed on implementing such API?

ret2libc: maybe if it's needed only here for now we can just use something small and local. The minimal things for this PR, to avoid blocking @meowmeowxw from going on with this.

libr/core/linux_heap_glibc.c
meowmeowxw: right!

libr/core/linux_heap_glibc.c
meowmeowxw: done

libr/debug/debug.c
meowmeowxw: Should I keep it double for this pull request or change to something else?

libr/core/linux_heap_glibc.c
meowmeowxw: In the code there are `GH(` (which is more readable IMHO) instead of `GH (`, should I change them all?

meowmeowxw:
Ok this PR now closes https://github.com/radareorg/radare2/issues/17205.
The tcache is parsed correctly for the various glibc version, and It also parse correctly the tcache in 32 bit system (it was broken before).

trufae:
Good work! i also think it's good to merge even without solving this double/float/r_semver topic, but maybe we should fill an issue for that. this r_semver api must be used from plugin loading apis, and should be used from r2pm if ever gets rewritten in C, so for me it makes sense to have it implemented at some point

libr/debug/debug.c
XVilka: @meowmeowxw just shift it to be int `231` for now maybe? The proper API can be introduced later and in a separate PR imho.

XVilka:
A good addition to 4.5.0 since fix.

libr/debug/debug.c
ret2libc: yeah use some int or something, but not double/float .

libr/core/linux_heap_glibc.c
ret2libc: I also think `GH(` should have no space and it is more readable... We also don't put space in `Elf_` in the elf code, which uses a similar structure.

meowmeowxw:
@ret2libc I need to add two binaries to this repo: https://github.com/radareorg/radare2-testbins, and `glibc-2.27` + ld-2.27`, should I create a pull request in the repo?

ret2libc:
> @ret2libc I need to add two binaries to this repo: https://github.com/radareorg/radare2-testbins, and `glibc-2.27` + `ld-2.27`, should I create a pull request in the repo?

Yes, add the binaries to that repository, then in this PR add a test in `test/db/`. See https://github.com/radareorg/radare2/tree/master/test .

meowmeowxw:
How can I set dbg.profile in a test (I couldn't find examples in the tests) ?

I neet to execute: `r2 -e dbg.profile=profile.rr2 -d bin/elf/tcache-2.27`

My test script is `test/db/archos/linux-x64/dbg_dmht`:

```
NAME=check tcache list on older version
FILE=bins/elf/tcache-2.27
ARGS=-e dbg.profile=./profile.rr2 -d
CMDS=<<EOF
db 0x004011d0
dc
dmht~?items : 6
EOF
EXPECT=<<EOF
1
EOF
RUN
```

And `profile.rr2`:

```
#!/usr/bin/env rarun2
setenv=LD_PRELOAD=bins/elf/libc-2.27.so
```

ret2libc:
Could you try with `-Rsetenv=LD_PRELOAD=bins/elf/libc-2.27.so` instead of `-e dbg.profile`?

meowmeowxw:
yes it works, I've added the relative test

XVilka:
I merged the testbins PR. Please rebase - it should be completely green. Apart from that looks fine 👍

PR #17209, opened 2020-07-05
macho libraries don't have an entrypoint and 0 is not a valid address ##bin

trufae:
* in this case the lib is mapped at 0 so the first executable section

	> iS paddr/sort,perm/str/-r-x,vaddr/cols~:3
	0x00000c30

 

**Detailed description**

see https://github.com/radareorg/radare2/pull/17195, macho libs show 1 entrypoint at adderss 0 when typing ie:

```
$ r2 /usr/local/lib/libr_util.dylib
 -- License server overloaded (ETOOMUCHCASH)
[0x00000000]> ie
[Entrypoints]
vaddr=0x00000000 paddr=0x00000000 haddr=-1 type=program

1 entrypoints

[0x00000000]>
```

**Test plan**

need to add a test opening a lib and chking the entry0 if it exists and whcih is the initial seek address. we can add another flag for this instead of entry0 which is tied to the bin headers.

**Closing issues**

kind of related to https://github.com/radareorg/radare2/pull/17195

### DIFF
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -3558,7 +3558,7 @@ struct addr_t *MACH0_(get_entrypoint)(struct MACH0_(obj_t) *bin) {
 	r_return_val_if_fail (bin, NULL);
 
 	ut64 ea = entry_to_vaddr (bin);
-	if (ea == 0 || ea == UT64_MAX) {
+	if (ea == 0 || ea == UT64_MAX) {
 		return NULL;
 	}
 	struct addr_t *entry = R_NEW0 (struct addr_t);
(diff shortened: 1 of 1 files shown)


mrmacete:
setting the entry point to the address of the first library's initializer (if present) could also be an option here?

trufae:
I would say the lowest symbol/initializer on an executable region (because there can be symbols for data) as well as executable sections mapped at 0, so a good/generic solution for all the fileformats not having a specific to macho

ret2libc:
It seems when the entrypoint could not be determined correctly from the file, each RBinPlugin does its own stuff... I agree a common solution at the RBin level and not at RBinPlugin level would be better.

trufae:
I have pushed libr_flag.dylib to bins, for testing this PR as well as exposing more bugs

ret2libc:
@trufae @mrmacete what to do with this PR? Do you think it's ok?

radare:
Imho the PR is fine and should be good to merge if @mrmacete aproves, but it exposes other bugs that imho should be addressed in a separate PR because it affects all the bin formats, not just macho

PR #17210, opened 2020-07-05
Deprecate `obb` command and make `obo` use the file descriptor ##bin

trufae:

**Detailed description**

This command was marked as deprecated for enough time

**Test plan**

Check if its not misused internally via rcore_cmd

**Closing issues**

dunno

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -70,11 +70,10 @@ static const char *help_msg_ob[] = {
 	"oba", " [addr] [baddr]", "Open file and load bin info at given address",
 	"oba", " [addr] [filename]", "Open file and load bin info at given address",
 	"oba", " [addr]", "Open bin info from the given address",
-	"obb", " [bfid]", "Switch to open binfile by fd number (Same as op)",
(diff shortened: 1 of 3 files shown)


libr/core/cmd_open.c
```diff
@@ -334,6 +323,18 @@ static void cmd_open_bin(RCore *core, const char *input) {
 			value = input[2] ? input + 2 : NULL;
 		}
 		break;
+#if 0
```
ret2libc, line 326: remove

trufae:
The thing is: we can raise the fd and the bfd by specifying the file-descriptor number or the binfile id.

* ob raises by binid
* obo raises by io-fd

the deprecation message was because the feature was implemented already and there was two commands doing the same. but obb was more about rbin, and obo more about io, imho, so after thinking a bit i think this is better approach.

I would like to know what do you think despite the fact of the DEPRECATE message was here or there, as well as get feedback on everyone on the community who can share their thoughs on this change

ret2libc:
Ok, seems to make sense!

trufae:
it's ready to merge imho

PR #17213, opened 2020-07-05
Read More ELF Dynamic Tags and Remove Warning

thestr4ng3r:

**Detailed description**

The new tags are just parsed and put in the structure like the others.

This warning here imo is quite useless, it tells a user nothing without reading the r2 source. Common elfs can easily contain some entries that by default we do not care about, for example my /bin/ls has DT_GNU_HASH which is 1879047925. It would make more sense to have a command that dumps the entire array to be examined manually if desired.

**Test plan**

r2r tests


XVilka:
Dupe of https://github.com/radareorg/radare2/pull/17055
But LGTM anyway.

thestr4ng3r:
Actually I like yours better

PR #17220, opened 2020-07-06
Fix #17137 - Support multiple symbol servers with ';'

### LINKED ISSUE #17137: Support multiple Symbol Servers
There is certainly a case for supporting multiple symbol servers. If an organization is writing Windows software, for example, and publishes its own symbols internally. Debugging the application requires loading both the public Microsoft symbols and the internally published ones. 

WinDBG supports multiple symbol servers using a parsable syntax: 

`SRV*d:\symcache*http://msdl.microsoft.com/download/symbols;SRV*d:\symcache*\\internal\DebuggingSymbols\App1`

GustavoLCR:

**Detailed description**

* Add support for UNC paths as symbol servers on Windows

**Closing issues**

Closes #17137

### DIFF
--- a/libr/bin/pdb/pdb_downloader.c
+++ b/libr/bin/pdb/pdb_downloader.c
@@ -23,15 +23,28 @@ static bool download_and_write(SPDBDownloaderOpt *opt, const char *file) {
 		return false;
 	}
 	char *url = r_str_newf ("%s/%s/%s/%s", opt->symbol_server, opt->dbg_file, opt->guid, file);
+	char *path = r_str_newf ("%s%s%s", dir, R_SYS_DIR, opt->dbg_file);
+#if __WINDOWS__
+	if (r_str_startswith (url, "\\\\")) { // Network path
+		LPCWSTR origin = r_utf8_to_utf16 (url);
+		LPCWSTR dest = r_utf8_to_utf16 (path);
+		BOOL ret = CopyFileW (origin, dest, FALSE);
+		free (dir);
+		free (path);
+		free (origin);
+		free (dest);
+		return ret;
+	}
+#endif
 	int len;
 	char *file_buf = r_socket_http_get (url, NULL, &len);
 	free (url);
 	if (!len || R_STR_ISEMPTY (file_buf)) {
 		free (dir);
 		free (file_buf);
+		free (path);
 		return false;
 	}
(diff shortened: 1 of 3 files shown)


libr/bin/pdb/pdb_downloader.c
```diff
@@ -23,15 +23,28 @@ static bool download_and_write(SPDBDownloaderOpt *opt, const char *file) {
 		return false;
 	}
 	char *url = r_str_newf ("%s/%s/%s/%s", opt->symbol_server, opt->dbg_file, opt->guid, file);
+	char *path = r_str_newf ("%s%s%s", dir, R_SYS_DIR, opt->dbg_file);
+#if __WINDOWS__
+	if (r_str_startswith (url, "\\\\")) { // Network path
```
ret2libc, line 28: so, this is in case you are on windows and the pdb file is somewhere on the network of the host? In that case, windows is capable of directly getting the file with a CopyFileW instead of using httpget? Did I get it right?

libr/bin/pdb/pdb_downloader.c
```diff
@@ -199,5 +212,5 @@ int r_bin_pdb_download(RCore *core, int isradjson, int *actions_done, SPDBOption
 	}
 	deinit_pdb_downloader (&pdb_downloader);
 
-	return 0;
+	return !ret;
```
ret2libc, line 215: this is the usual int vs bool mess :/ I'm never sure what functions are supposed to return. Is 0 the right code? Is this function returning just 1 and 0? If so, can we make it return bool instead?

libr/bin/pdb/pdb_downloader.c
xarkes: Yes indeed, you can check all the valid file names [here](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file) that CopyFileW can take.

libr/bin/pdb/pdb_downloader.c
GustavoLCR: @ret2libc yeah, this function returns 0 on success, 1 on failure. probably should fix that in another PR and change it to a bool

XVilka:
@ret2libc please re-review and if you are ok - lets merge this.

PR #17233, opened 2020-07-07
Use more size_t in the MACHO parser ##bin

trufae:

**Detailed description**

Th eoriginal purpose of this PR has changed. it is now a cleanup and random cleanup

* Use more size_t for loop variables
* Fix a memleak in checkSym
* remove unused variables

There are more bugs like using globals and such, will be slowly addressing them all

**Test plan**

nothing must break after merging this

**Closing issues**

nope but the code is full of bugs

### DIFF
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -2387,7 +2387,7 @@ static bool parse_import_stub(struct MACH0_(obj_t) *bin, struct symbol_t *symbol
 				symbol->addr = bin->sects[i].addr + delta;
 				symbol->size = 0;
 				stridx = bin->symtab[idx].n_strx;
-				if (stridx >= 0 && stridx < bin->symstrlen) {
+				if (stridx < bin->symstrlen) {
 					symstr = (char *)bin->symstr + stridx;
 				} else {
 					symstr = "???";
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -2084,9 +2084,9 @@ void *MACH0_(mach0_free)(struct MACH0_(obj_t) *mo) {
 }
 
 void MACH0_(opts_set_default)(struct MACH0_(opts_t) *options, RBinFile *bf) {
-	r_return_if_fail (options && bf);
+	r_return_if_fail (options && bf && bf->rbin);
 	options->header_at = 0;
-	options->verbose = bf && bf->rbin && bf->rbin->verbose;
+	options->verbose = bf->rbin->verbose;
 }
 
 static void *duplicate_ptr(void *p) {
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -2356,7 +2356,7 @@ struct section_t *MACH0_(get_sections)(struct MACH0_(obj_t) *bin) {
 }
 
 static bool parse_import_stub(struct MACH0_(obj_t) *bin, struct symbol_t *symbol, int idx) {
-	int i, j, nsyms, stridx;
+	size_t i, j, nsyms, stridx;
 	const char *symstr;
 	if (idx < 0) {
 		return false;
@@ -2436,7 +2436,7 @@ static int inSymtab(HtPP *hash, const char *name, ut64 addr) {
 
 static char *get_name(struct MACH0_(obj_t) *mo, ut32 stridx, bool filter) {
 	int i = 0;
-	if (stridx >= mo->symstrlen) {
+	if (!mo->symstr || stridx >= mo->symstrlen) {
 		return NULL;
 	}
 	int len = mo->symstrlen - stridx;
@@ -2637,7 +2637,8 @@ const RList *MACH0_(get_symbols_list)(struct MACH0_(obj_t) *bin) {
 	static RList * cache = NULL; // XXX DONT COMMIT WITH THIS
 	struct symbol_t *symbols;
 	int j, s, stridx, symbols_size, symbols_count;
-	ut32 to, from, i;
+	ut32 to, from;
+	size_t i;
 
 	r_return_val_if_fail (bin, NULL);
 	if (cache) {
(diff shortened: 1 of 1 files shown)


trufae:
For testing just type `is` in libr_flag.dylib from testbins

libr/bin/format/mach0/mach0.c
```diff
@@ -2949,32 +2946,42 @@ const struct symbol_t *MACH0_(get_symbols)(struct MACH0_(obj_t) *bin) {
...
 				char *sym_name = get_name (bin, stridx, false);
 				if (sym_name) {
 					symbols[j].name = sym_name;
 				} else {
-					symbols[j].name = r_str_newf ("entry%d\n", i);
-					//symbols[j].name[0] = 0;
+					size_t k = 0;
+					bool named_symbol_found = false;
+					symbols[j].last = true;
+					for (k = 0; !symbols[k].last; k++) {
```
ret2libc, line 2970: depending on the binary this could slow down a lot the parsing of mach0 files, but probably empty `sym_name` is rare enough to not care about this case for now.

trufae: Yes, this is not a good solution. Actually i couldnt find a bin with unnamed symbols so not sure if we could just get rid of the whole loop entirely without losing any valuable information. Maybe stripped?

mrmacete:
So, all this is apparently caused by the `N_STAB` kind of symbols.

Excerpt from `<mach-o/nlist.h>`:

```
#define	N_STAB	0xe0  /* if any of these bits set, a symbolic debugging entry */
#define	N_PEXT	0x10  /* private external symbol bit */
#define	N_TYPE	0x0e  /* mask for the type bits */
#define	N_EXT	0x01  /* external symbol bit, set for external symbols */

/*
 * Only symbolic debugging entries have some of the N_STAB bits set and if any
 * of these bits are set then it is a symbolic debugging entry (a stab).  In
 * which case then the values of the n_type field (the entire field) are given
 * in <mach-o/stab.h>
 */
```

which means that if `(st->n_type & N_TAB) != 0` the entire `n_type` field must be interpreted as defined in `<mach-o/stab.h>` and that gives another meaning to the entire nlist structure:

```
#define	N_GSYM	0x20	/* global symbol: name,,NO_SECT,type,0 */
#define	N_FNAME	0x22	/* procedure name (f77 kludge): name,,NO_SECT,0,0 */
#define	N_FUN	0x24	/* procedure: name,,n_sect,linenumber,address */
#define	N_STSYM	0x26	/* static symbol: name,,n_sect,type,address */
#define	N_LCSYM	0x28	/* .lcomm symbol: name,,n_sect,type,address */
#define N_BNSYM 0x2e	/* begin nsect sym: 0,,n_sect,0,address */
#define N_AST	0x32	/* AST file path: name,,NO_SECT,0,0 */
#define N_OPT	0x3c	/* emitted with gcc2_compiled and in gcc source */
#define	N_RSYM	0x40	/* register sym: name,,NO_SECT,type,register */
#define	N_SLINE	0x44	/* src line: 0,,n_sect,linenumber,address */
#define N_ENSYM 0x4e	/* end nsect sym: 0,,n_sect,0,address */
#define	N_SSYM	0x60	/* structure elt: name,,NO_SECT,type,struct_offset */
#define	N_SO	0x64	/* source file name: name,,n_sect,0,address */
#define	N_OSO	0x66	/* object file name: name,,(see below),0,st_mtime */
			/*   historically N_OSO set n_sect to 0. The N_OSO
			 *   n_sect may instead hold the low byte of the
			 *   cpusubtype value from the Mach-O header. */
#define	N_LSYM	0x80	/* local sym: name,,NO_SECT,type,offset */
#define N_BINCL	0x82	/* include file beginning: name,,NO_SECT,0,sum */
#define	N_SOL	0x84	/* #included file name: name,,n_sect,0,address */
#define	N_PARAMS  0x86	/* compiler parameters: name,,NO_SECT,0,0 */
#define	N_VERSION 0x88	/* compiler version: name,,NO_SECT,0,0 */
#define	N_OLEVEL  0x8A	/* compiler -O level: name,,NO_SECT,0,0 */
#define	N_PSYM	0xa0	/* parameter: name,,NO_SECT,type,offset */
#define N_EINCL	0xa2	/* include file end: name,,NO_SECT,0,0 */
#define	N_ENTRY	0xa4	/* alternate entry: name,,n_sect,linenumber,address */
#define	N_LBRAC	0xc0	/* left bracket: 0,,NO_SECT,nesting level,address */
#define N_EXCL	0xc2	/* deleted include file: name,,NO_SECT,0,sum */
#define	N_RBRAC	0xe0	/* right bracket: 0,,NO_SECT,nesting level,address */
#define	N_BCOMM	0xe2	/* begin common: name,,NO_SECT,0,0 */
#define	N_ECOMM	0xe4	/* end common: name,,n_sect,0,0 */
#define	N_ECOML	0xe8	/* end common (local name): 0,,n_sect,0,address */
#define	N_LENG	0xfe	/* second stab entry with length information */
```

as probably this doesn't add much to the existing symbol information, for now we can probably just skip the `N_STAB` symbols to avoid noise and performance loss, and maybe add proper support for them in the future?

trufae:
Sounds like a plan to me. Do you have some time to make a PR with this change? I will probably close this PR because the solution is not correct, but there are some changes that are unrelated to this approach and should bemerged too.

You can commit to this branch if you want too. Otherwise i will do it when i find some time. But this is an importsnt fix

The library withuot entrypoints issue is also important. I have some more queued issues to fix related to macho and dex so i will keep pushing improvements over time

trufae:
@mrmacete added the if, can you check it out again?

trufae:
Ive stripped downt he minimum needed to fix the issue in https://github.com/radareorg/radare2/pull/17277. ill update this PR when 17277 is merged

trufae:
I'll open an issue to use generic uleb128 apis instead of duplicating the implementation in here https://github.com/radareorg/radare2/issues/17353

libr/bin/format/mach0/mach0.c
```diff
@@ -2095,16 +2084,9 @@ void *MACH0_(mach0_free)(struct MACH0_(obj_t) *mo) {
...
-	}
-
+	r_return_if_fail (options && bf);
 	options->header_at = 0;
-	if (bf && bf->rbin) {
-		options->verbose = bf->rbin->verbose;
-	} else {
-		options->verbose = false;
-	}
+	options->verbose = bf && bf->rbin && bf->rbin->verbose;
```
ret2libc, line 2089: no need to check `bf`, it is not-null for sure at this point

ret2libc: actually, we don't even need to check if bf->rbin exist, because RBinFile should always be created with a reference to the RBin.

PR #17243, opened 2020-07-08
Update spp to 1.2.0

ret2libc:


XVilka:
Looks like it has broken A LOT of tests: https://github.com/radareorg/radare2/pull/17243/checks?check_run_id=849728884

trufae:
Only affecting the newshell builds. Weird

PR #17249, opened 2020-07-09
Avoid sdb double headers

ret2libc:

**Detailed description**

Some time in the future we may want to finally rely on git submodules/subtree or something like that, that would allow us to use external dependencies without the need to actually copy all the code in the radare2 repository. To do this, however, subprojects should be self-contained. SDB is heavily used in r2, however for simplicity (I think) its headers were copied and mixed with r2 headers (you could find SDB headers both in shlr/sdb/src and in libr/include/sdb/ + libr/include/sdb.h). In theory we should have only one place where files are kept, not two.

**Test plan**

...

**Closing issues**

...


trufae:
I think we should change the way includes are located in sdb before doing this change in here, the reason why sdb iuncludes are deployed in libr/ is because r2 depends on a specific version of sdb and therefor the includes must be distributed with it in ordedr to work.

I did this for capstone a while ago, because the path of the sdb includes was different depending if you used the srcdir or the install path.

ret2libc:
> I think we should change the way includes are located in sdb before doing this change in here, the reason why sdb iuncludes are deployed in libr/ is because r2 depends on a specific version of sdb and therefor the includes must be distributed with it in ordedr to work.

What do you mean? I think they can stay where they are. When compiling r2 against a system-sdb, it's enough to have something like `-I/usr/include/sdb` in your cflags. When compiling against the embedded version of sdb, you shall have something like `-I../../shlr/sdb/src`. Moreover, r2 should install the SDB headers only when it is not using the system SDB headers.

libr/Makefile
```diff
@@ -183,7 +183,7 @@ install-includes:
 		done ; \
 	done)
 	${INSTALL_DIR} "${DESTDIR}${INCLUDEDIR}/libr/sdb"
-	(cd include/sdb && for FILE in * ; do \
+	(cd ../shlr/sdb/src && for FILE in *.h ; do \
```
trufae, line 186: can you fix the symstall?

ret2libc:
I will check, but I think this should still be the case.

libr/Makefile
ret2libc: done

shlr/zip/Makefile
```diff
@@ -1,5 +1,6 @@
 NAME=rz
 LIBAR=librz.$(EXT_AR)
+CFLAGS+=-I../sdb/src/
```
trufae, line 3: why libzip needs sdb?

shlr/windbg/Makefile
```diff
@@ -6,6 +6,7 @@ include ../../mk/${COMPILER}.mk
 CFLAGS+=${PIC_CFLAGS}
 
 CFLAGS+=-I../../libr/include
+CFLAGS+=-I../sdb/src
```
trufae, line 9: same

shlr/qnx/Makefile
```diff
@@ -8,7 +8,7 @@ RANLIB?=ranlib
 LIBNAME=libqnxr
 LIBFILE=$(LIBNAME).$(EXT_SO)
 CFLAGS+=-I$(LIBR)/include $(PIC_CFLAGS)
-CFLAGS+=-Iinclude -I${LIBR}/include
+CFLAGS+=-Iinclude -I${LIBR}/include -I../sdb/src/
```
trufae, line 11: is this relaly needed?

shlr/grub/Makefile
```diff
@@ -60,7 +60,7 @@ KERNFILES+=partmap/bsdlabel.c
 KERNFILES+=partmap/sunpc.c
 
 KERNOBJS=$(subst .c,.o,${KERNFILES})
-CFLAGS+=-Iinclude
+CFLAGS+=-Iinclude -I../sdb/src/
```
trufae, line 63: i dont think grub needs sdb :?

shlr/ar/Makefile
```diff
@@ -26,7 +26,7 @@ endif
 endif
 endif
 
-CFLAGS+=-I../../libr/include
+CFLAGS+=-I../../libr/include -I../sdb/src/
```
trufae, line 29: ideally i would prefer to include sdb/deps.mk insteada of having this everywhere duplicated

ret2libc:
> See comments, the problem i see, is that, once installed the sdb headders wont be available in the system, so 3rd party plugins wont be able to compile. unless i missed the place where those files are installed separately

SDB headers are installed at https://github.com/radareorg/radare2/pull/17249/files#diff-e9de9044d8293a96538b393a6c9612faR186 and https://github.com/radareorg/radare2/pull/17249/files#diff-e9de9044d8293a96538b393a6c9612faR203 for symstall installation.

Also, SDB is needed for most of the shlr code because they were modified to use `r_util` and `r_util` needs SDB (indeed `r_list.h` includes `ht.h`, which is defined by SDB.

trufae:
Ill review once all my comments are addressed/commented

PR #17251, opened 2020-07-10
Improve multi-threaded debugging on Linux ##debug

### LINKED ISSUE #8749: Continuing radare2 hangs debuggee (and probably itself)
Steps to reproduce:

1. Run `kate` *(a graphical application, but it should work with anything — I initially found it with Skype)*.
2. Attach radare2 using `radare2 -d $(pgrep kate)`
3. Type `dc` to continue executing process.

What happens:
1. kate doesn't respond, it just stops updating its window as if I didn't type `dc`
2. Pressing <kbd>Ctrl</kbd>+<kbd>c</kbd> in radare2 doesn't interrupt it either, looks like it hanged too.

What should happen:
1. kate should continue working as usual
2. Pressing <kbd>Ctrl</kbd>+<kbd>c</kbd> in radare2 should interrupt execution.

Additional info:
radare version is 2.0.1
Pressing <kbd>Ctrl</kbd>+<kbd>c</kbd> in terminal with Kate results in radare2 interrupting execution with 

	child stopped with signal 2
	[+] SIGNAL 2 errno=0 addr=0x00000000 code=128 ret=0 got signal...

So it seems like radare2 somewhat continues working.

abcSup:

**Detailed description**
## Issues
- Failure to continue a multithreaded process after attaching it because `dbg->n_threads` is 0.
- Missing important signals `SIGTRAP` `SIGSTOP` when debugging a multithreaded process because `r_debug_continue` always continue all threads without checking if the thread is already stopped or has changed status.
- ^C does not interrupt the debugger because `r_sys_signal (SIGINT, SIG_DFL)` removes the signal handler (`__break_signal`) set by `RCons`.
- Debugger stuck in a blocking `waitpid` call because all signals are already consumed by the debugger and all threads are already stopped.

## Summary
When users interrupt `r2` with SIGINT (^C):
1. `RCons` will use the callback `static_debug_stop` in `cmd_debug.c` to call `r_debug_stop`.

2. `r_debug_stop` calls `linux_stop_threads`.

    - To avoid additional `SIGSTOP` signals, we only stop threads that are running by using `PTRACE_GETSIGINFO` to check if the thread is running.

3. Avoid calling `r_debug_stop` in `r_debug_continue_kill` again, to prevent duplicate `SIGSTOP` sent by debugger which will interrupt `r_debug_wait` later.

---

Continuing after user interrupt (`dbg->continue_all_threads == true`):
1. Continue the previous pid/tid reported by `r_debug_wait` in `dbg->tid` as usual.

2. Loop through `dbg->threads` and only continue threads that are stopped by `SIGSTOP` sent by the debugger.
    - Retrieve signal information with `PTRACE_GETSIGINFO` and look for `siginfo.si_signo == SIGSTOP`
    - Find`siginfo.si_pid == 0` (stopped by `PTRACE_ATTACH`) or `siginfo.si_pid == dbgpid` (stopped by the debugger)
    - Before `PTRACE_CONT`, we must consume the signal by calling `waitpid` on the tid first. 

---

`SIGSTOP` can happen when:
- Newly created process and task starts with `SIGSTOP`
- `PTRACE_ATTACH`
- User interrupt which causes the debugger to send `SIGSTOP` to all running threads.

Therefore, when `linux_dbg_wait` encounters `SIGSTOP`, we should check if it is caused by thread/process creation:
1. Loop through `dbg->threads` and search for threads stopped by `SIGTRAP`. We use `PTRACE_GETSIGINFO` instead of `waitpid` with `WNOHANG` flag to check each thread because we do not want to consume `SIGTRAP` that is not caused by `PTRACE_EVENT_*`, for example, `SIGTRAP` caused by single-step and breakpoints.

2. Consume the `SIGTRAP` with `waitpid` and call `linux_ptrace_event` to analyze the events and act accordingly

3. Return the reason from `linux_ptrace_event`.

---

`linux_dbg_wait` should only report status changes and set `dbg->pid` and `dbg->tid`:
- Remove `goto repeat` loop in `linux_dbg_wait` because `r_debug` should be the one calling `r_debug_wait` again when certain event happens.
- `linux_dbg_wait` should not `PTRACE_CONT` an exiting thread, when `R_DEBUG_REASON_EXIT_TID` is reported by `linux_ptrace_event`, because `r_debug_continue_kill` will handle that and decide if it should continue running.

---

- `linux_stop_thread` should return if the thread is already stopped to prevent duplicate `SIGSTOP`
- Instead of `linux_add_and_attach_new_thread `, we use `linux_add_new_thread` when `R_DEBUG_REASON_NEW_TID` happens because newly threads are already attached due to the `PTRACE_O_TRACECLONE` option.
- `linux_set_options` will always succeed on an already attached thread/process. But, `PTRACE_SETOPTIONS` will sometimes fail because of the `ptrace` asynchronous nature, so we will call it in a loop until it succeed and sleep for awhile if it fails.

## Remaining tasks
- [x] Clear SIGSTOP caused by user interrupt ^C before stepping
- [x] Remove all debugging outputs

## Future improvements
- Managing multiple processes, not only threads. The current implementation is buggy.
- `linux_select` should not use `linux_attach` to select a thread because all threads are already attached. This will only generate more error outputs when `PTRACE_ATTACH` fails.
- `linux_thread_list` should not try to attach already traced threads. This will also generate more error outputs when `PTRACE_ATTACH` fails.

I will update the description with any missing details. I am open for discussion.

**Test plan**

Currently, I am testing this with debug outputs on `kate` and some programs written in C that do multithreading with blocking calls and exits in different orders.

Generally, we need to test:
- [ ] Works for programs calling `clone` and `fork` in different orders and nested threads.
- [ ] Make sure breakpoints work across different threads.
- [ ] Check if ^C works.
- [ ] Attaching works for multithreaded programs.

**Closing issues**

closes #8749

### DIFF
--- a/libr/debug/debug.c
+++ b/libr/debug/debug.c
@@ -772,8 +772,6 @@ R_API RDebugReasonType r_debug_wait(RDebug *dbg, RBreakpointItem **bp) {
 			}
 		}
 	}
-	eprintf ("r_debug_wait reason: %d %s\n", reason, r_debug_reason_to_string (reason));
-	eprintf ("pid: %d tid: %d\n", dbg->pid, dbg->tid);
 	return reason;
 }
 
@@ -910,7 +908,18 @@ R_API int r_debug_step_hard(RDebug *dbg, RBreakpointItem **pb) {
 	if (!dbg->h->step (dbg)) {
 		return false;
 	}
+
+#if __linux__
+	// Turn off continue_all_threads to make sure linux_dbg_wait
+	// only waits for one target for a single-step or breakpoint trap
+	bool prev_continue = dbg->continue_all_threads;
+	dbg->continue_all_threads = false;
+#endif
 	reason = r_debug_wait (dbg, pb);
+#if __linux__
+	dbg->continue_all_threads = prev_continue;
+#endif
+
 	if (reason == R_DEBUG_REASON_DEAD || r_debug_is_dead (dbg)) {
 		return false;
 	}
@@ -1288,7 +1297,6 @@ repeat:
 	// Stop all threads when there is no user interrupt caused by statc_debug_stop
 	// to prevent queueing duplicate SIGSTOP signals
 	if (!r_cons_is_breaked () && dbg->continue_all_threads) {
-		eprintf ("STOP AFTER CONTINUE\n");
 		r_debug_stop (dbg);
 	}
 #endif
--- a/libr/debug/debug.c
+++ b/libr/debug/debug.c
@@ -711,6 +711,15 @@ R_API RDebugReasonType r_debug_wait(RDebug *dbg, RBreakpointItem **bp) {
 			return R_DEBUG_REASON_DEAD;
 		}
 
+#if __linux__
+		// Letting other threads running after the debugger has breaked will cause
+		// ptrace commands to fail when writing the process memory to set/unset
+		// breakpoints and is problematic in Linux.
+		if (dbg->continue_all_threads) {
+			r_debug_stop (dbg);
+		}
+#endif
+
 		/* propagate errors from the plugin */
 		if (reason == R_DEBUG_REASON_ERROR) {
 			return R_DEBUG_REASON_ERROR;
@@ -1292,14 +1301,8 @@ repeat:
 	}
 #if __WINDOWS__
 	r_cons_break_pop ();
-#elif __linux__
-	// Letting threads continue after the debugger breaks is currently problematic in linux
-	// Stop all threads when there is no user interrupt caused by statc_debug_stop
-	// to prevent queueing duplicate SIGSTOP signals
-	if (!r_cons_is_breaked () && dbg->continue_all_threads) {
-		r_debug_stop (dbg);
-	}
 #endif
+
 	// Unset breakpoints before leaving
 	if (reason != R_DEBUG_REASON_BREAKPOINT) {
 		r_bp_restore (dbg->bp, false);
--- a/libr/debug/debug.c
+++ b/libr/debug/debug.c
@@ -712,9 +712,9 @@ R_API RDebugReasonType r_debug_wait(RDebug *dbg, RBreakpointItem **bp) {
 		}
 
 #if __linux__
-		// Letting other threads running after the debugger has breaked will cause
-		// ptrace commands to fail when writing the process memory to set/unset
-		// breakpoints and is problematic in Linux.
+		// Letting other threads running will cause ptrace commands to fail
+		// when writing to the same process memory to set/unset breakpoints
+		// and is problematic in Linux.
 		if (dbg->continue_all_threads) {
 			r_debug_stop (dbg);
 		}
--- a/libr/debug/debug.c
+++ b/libr/debug/debug.c
@@ -1203,18 +1203,16 @@ repeat:
 	}
 
 #if __linux__
-	if (reason == R_DEBUG_REASON_NEW_PID) {
+	if (reason == R_DEBUG_REASON_NEW_PID && dbg->follow_child) {
 #if DEBUGGER
-		if (dbg->follow_child) {
-			/// if the plugin is not compiled link fails, so better do runtime linking
-			/// until this code gets fixed
-			static bool (*linux_attach_new_process) (RDebug *dbg, int pid) = NULL;
-			if (!linux_attach_new_process) {
-				linux_attach_new_process = r_lib_dl_sym (NULL, "linux_attach_new_process");
-			}
-			if (linux_attach_new_process) {
-				linux_attach_new_process (dbg, dbg->forked_pid);
-			}
+		/// if the plugin is not compiled link fails, so better do runtime linking
+		/// until this code gets fixed
+		static bool (*linux_attach_new_process) (RDebug *dbg, int pid) = NULL;
+		if (!linux_attach_new_process) {
+			linux_attach_new_process = r_lib_dl_sym (NULL, "linux_attach_new_process");
+		}
+		if (linux_attach_new_process) {
+			linux_attach_new_process (dbg, dbg->forked_pid);
 		}
 #endif
 		goto repeat;
--- a/libr/debug/debug.c
+++ b/libr/debug/debug.c
@@ -772,6 +772,8 @@ R_API RDebugReasonType r_debug_wait(RDebug *dbg, RBreakpointItem **bp) {
 			}
 		}
 	}
+	eprintf ("r_debug_wait reason: %d %s\n", reason, r_debug_reason_to_string (reason));
+	eprintf ("pid: %d tid: %d\n", dbg->pid, dbg->tid);
 	return reason;
 }
 
@@ -1183,16 +1185,18 @@ repeat:
 	}
 
 #if __linux__
-	if (reason == R_DEBUG_REASON_NEW_PID && dbg->follow_child) {
+	if (reason == R_DEBUG_REASON_NEW_PID) {
 #if DEBUGGER
-		/// if the plugin is not compiled link fails, so better do runtime linking
-		/// until this code gets fixed
-		static bool (*linux_attach_new_process) (RDebug *dbg, int pid) = NULL;
-		if (!linux_attach_new_process) {
-			linux_attach_new_process = r_lib_dl_sym (NULL, "linux_attach_new_process");
-		}
-		if (linux_attach_new_process) {
-			linux_attach_new_process (dbg, dbg->forked_pid);
+		if (dbg->follow_child) {
+			/// if the plugin is not compiled link fails, so better do runtime linking
+			/// until this code gets fixed
+			static bool (*linux_attach_new_process) (RDebug *dbg, int pid) = NULL;
+			if (!linux_attach_new_process) {
+				linux_attach_new_process = r_lib_dl_sym (NULL, "linux_attach_new_process");
+			}
+			if (linux_attach_new_process) {
+				linux_attach_new_process (dbg, dbg->forked_pid);
+			}
 		}
 #endif
 		goto repeat;
@@ -1281,7 +1285,10 @@ repeat:
 	r_cons_break_pop ();
 #elif __linux__
 	// Letting threads continue after the debugger breaks is currently problematic in linux
-	if (dbg->continue_all_threads) {
+	// Stop all threads when there is no user interrupt caused by statc_debug_stop
+	// to prevent queueing duplicate SIGSTOP signals
+	if (!r_cons_is_breaked () && dbg->continue_all_threads) {
+		eprintf ("STOP AFTER CONTINUE\n");
 		r_debug_stop (dbg);
 	}
 #endif
--- a/libr/debug/p/debug_native.c
+++ b/libr/debug/p/debug_native.c
@@ -207,9 +207,8 @@ static void interrupt_process(RDebug *dbg) {
 
 static int r_debug_native_stop(RDebug *dbg) {
 #if __linux__
-	// Stop all running threads except the current thread
-	// which is already stopped by linux_dbg_wait_break
-	return linux_stop_threads (dbg, dbg->tid);
+	// Stop all running threads except the thread reported by waitpid
+	return linux_stop_threads (dbg, dbg->reason.tid);
 #else
 	return 0;
 #endif
@@ -242,35 +241,21 @@ static int r_debug_native_continue(RDebug *dbg, int pid, int tid, int sig) {
 		r_cons_break_push ((RConsBreak)interrupt_process, dbg);
 	}
 
-	ret = r_debug_ptrace (dbg, PTRACE_CONT, tid, NULL, (r_ptrace_data_t)(size_t)contsig);
-	if (ret == -1) {
-		r_sys_perror ("PTRACE_CONT");
-	}
 
-	// Continue all threads that are stopped by the debugger only
-	// to keep other signals in the queue for the next waitpid
-	// by checking for SIGSTOP signals sent by the debugger
-	if (dbg->continue_all_threads && dbg->threads) {
+	if (dbg->continue_all_threads && dbg->n_threads && dbg->threads) {
 		RDebugPid *th;
 		RListIter *it;
-		pid_t dbgpid = getpid ();
 		r_list_foreach (dbg->threads, it, th) {
-			if (th->pid == tid) {
-				continue;
-			}
-			siginfo_t siginfo = { 0 };
-			ret = r_debug_ptrace (dbg, PTRACE_GETSIGINFO, th->pid, 0, (r_ptrace_data_t)(size_t)&siginfo);
-			// Skip if the thread is already running to prevent error
+			ret = r_debug_ptrace (dbg, PTRACE_CONT, th->pid, 0, 0);
 			if (ret == -1) {
-				continue;
+				eprintf ("Error: (%d) is already running.\n", th->pid);
+				r_sys_perror ("PTRACE_CONT");
 			}
-			if (siginfo.si_signo == SIGSTOP && siginfo.si_pid == dbgpid) {
-				ret = r_debug_ptrace (dbg, PTRACE_CONT, th->pid, 0, 0);
-				if (ret == -1) {
-					eprintf ("Error: %d is not traced or already running.\n", th->pid);
-					r_sys_perror ("PTRACE_CONT");
-				}
-			}
+		}
+	} else {
+		ret = r_debug_ptrace (dbg, PTRACE_CONT, tid, NULL, (r_ptrace_data_t)(size_t)contsig);
+		if (ret) {
+			r_sys_perror ("PTRACE_CONT");
 		}
 	}
 	//return ret >= 0 ? tid : false;
--- a/libr/debug/p/debug_native.c
+++ b/libr/debug/p/debug_native.c
@@ -207,7 +207,7 @@ static void interrupt_process(RDebug *dbg) {
 
 static int r_debug_native_stop(RDebug *dbg) {
 #if __linux__
-	return linux_stop_threads (dbg, dbg->reason.tid);
+	return linux_stop_threads (dbg, dbg->pid);
 #else
 	return 0;
 #endif
@@ -240,26 +240,43 @@ static int r_debug_native_continue(RDebug *dbg, int pid, int tid, int sig) {
 		r_cons_break_push ((RConsBreak)interrupt_process, dbg);
 	}
 
-	if (dbg->continue_all_threads && dbg->n_threads) {
-		RList *list = dbg->threads;
+	ret = r_debug_ptrace (dbg, PTRACE_CONT, tid, NULL, (r_ptrace_data_t)(size_t)contsig);
+	eprintf ("Continue: %d\n", tid);
+	if (ret == -1) {
+		r_sys_perror ("PTRACE_CONT");
+	}
+	// Continue threads that are stopped by the debugger to keep other signals
+	// by checking the siginfo.si_pid:
+	// - 0		= stopped by PTRACE_ATTACH
+	// - dbgpid = stopped by debugger
+	if (dbg->continue_all_threads && dbg->threads) {
 		RDebugPid *th;
 		RListIter *it;
-
-		if (list) {
-			r_list_foreach (list, it, th) {
-				if (th->pid) {
-					ret = r_debug_ptrace (dbg, PTRACE_CONT, th->pid, NULL, (r_ptrace_data_t)(size_t)contsig);
-					if (ret) {
-						perror ("PTRACE_CONT");
-					}
+		int dbgpid = getpid ();
+		r_list_foreach (dbg->threads, it, th) {
+			// Check if thread is already running
+			siginfo_t siginfo = { 0 };
+			ret = r_debug_ptrace (dbg, PTRACE_GETSIGINFO, th->pid, 0, (r_ptrace_data_t)(size_t)&siginfo);
+			if (ret == -1) {
+				continue;
+			}
+			eprintf ("tid: %d si_signo: %d siginfo.si_pid: %d\n", th->pid, siginfo.si_signo, siginfo.si_pid);
+			if (siginfo.si_signo == SIGSTOP &&
+				(siginfo.si_pid == dbgpid || siginfo.si_pid == 0)) {
+				int status;
+				ret = waitpid (th->pid, &status, 0);
+				eprintf ("Consume: %d status %d\n", th->pid, WSTOPSIG (status));
+				if (ret != th->pid) {
+					eprintf ("Error: waitpid fails to consume SIGSTOP\n");
+					continue;
+				}
+				eprintf ("Continue: %d\n", th->pid);
+				ret = r_debug_ptrace (dbg, PTRACE_CONT, th->pid, 0, 0);
+				if (ret == -1) {
+					eprintf ("Error: %d is not traced or already running.\n", th->pid);
+					r_sys_perror ("PTRACE_CONT");
 				}
 			}
-		}
-	}
-	else {
-		ret = r_debug_ptrace (dbg, PTRACE_CONT, tid, NULL, (r_ptrace_data_t)(size_t)contsig);
-		if (ret) {
-			perror ("PTRACE_CONT");
 		}
 	}
 	//return ret >= 0 ? tid : false;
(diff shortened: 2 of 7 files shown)


libr/debug/debug.c
```diff
@@ -1179,19 +1181,22 @@ R_API int r_debug_continue_kill(RDebug *dbg, int sig) {
...
-			if (!linux_attach_new_process) {
-				linux_attach_new_process = r_lib_dl_sym (NULL, "linux_attach_new_process");
-			}
-			if (linux_attach_new_process) {
-				linux_attach_new_process (dbg, dbg->forked_pid);
+			if (dbg->follow_child) {
+				/// if the plugin is not compiled link fails, so better do runtime linking
+				/// until this code gets fixed
+				static bool (*linux_attach_new_process) (RDebug *dbg, int pid) = NULL;
+				if (!linux_attach_new_process) {
```
trufae, line 1190: this dlsym thing is a hack, and wont work on stripped builds of r2. This code can be inside an #if __linux__ instead of asuming the symbol is named like this and fail at runtime, the comment above must be removed and the code fixed

libr/debug/debug.c
```diff
@@ -1239,11 +1244,13 @@ R_API int r_debug_continue_kill(RDebug *dbg, int sig) {
 			goto repeat;
 		}
 
+#if !__linux__
```
trufae, line 1247: uhm, can you confirm in which other OS is not the same?

abcSup: Sure, I will go find out why this line is necessary. Imho, the `dbg->tid` returned by `r_debug_continue` does not necessarily mean which thread is stopped and it should rely on `r_debug_wait` to tell which thread changes its states, so these lines are not necessary on Linux.

This is just a quick way to not break other debugger plugins which rely on these lines. For Linux native debugger, it relies on `r_debug_wait` to change the `dbg->tid` because the `dbg->tid` returned by `continue` which will always be the same as before `continue`.

libr/debug/p/debug_native.c
```diff
@@ -240,28 +240,45 @@ static int r_debug_native_continue(RDebug *dbg, int pid, int tid, int sig) {
...
 		RListIter *it;
-
-		if (list) {
-			r_list_foreach (list, it, th) {
-				if (th->pid) {
-					ret = r_debug_ptrace (dbg, PTRACE_CONT, th->pid, NULL, (r_ptrace_data_t)(size_t)contsig);
-					if (ret) {
-						perror ("PTRACE_CONT");
-					}
+		int dbgpid = getpid ();
```
trufae, line 255: pid_t

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -223,11 +233,17 @@ RDebugReasonType linux_ptrace_event (RDebug *dbg, int pid, int status) {
 			r_sys_perror ("ptrace GETEVENTMSG");
 			return R_DEBUG_REASON_ERROR;
 		}
-		linux_add_and_attach_new_thread (dbg, (int)data);
+		linux_add_new_thread (dbg, (int)data);
 		if (dbg->trace_clone) {
 			r_debug_select (dbg, dbg->pid, (int)data);
 		}
 		eprintf ("(%d) Created thread %d\n", dbg->pid, (int)data);
+		// Continue the parent task
```
yossizap, line 241: This seems to be duplicated in other cases as well, it should be handled outside of linux_ptrace_event using the debug reason.

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -494,7 +533,7 @@ static void linux_add_and_attach_new_thread(RDebug *dbg, int tid) {
 	} else {
 		tid_info = r_debug_pid_new ("new_path", tid, uid, 's', 0);
 	}
-	(void) linux_attach_single_pid (dbg, tid);
```
yossizap, line 497: The new child is attached at this point but it has a pending SIGSTOP and we can't affect it until it arrives. We should waitpid here, you might have added more sigstop handling elsewhere because this wasn't handled properly.

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -510,32 +549,30 @@ static bool linux_kill_thread(int tid, int signo) {
...
-	if (linux_kill_thread (tid, SIGSTOP)) {
-		if ((ret = waitpid (tid, &status, WUNTRACED)) == -1) {
-			perror("waitpid");
-		}
+static bool linux_stop_thread(RDebug *dbg, int tid) {
+	int ret;
+	siginfo_t siginfo = { 0 };
+	// Return if thread is already stopped to prevent an extra SIGSTOP signal
+	ret = r_debug_ptrace (dbg, PTRACE_GETSIGINFO, tid, 0, (r_ptrace_data_t)(intptr_t)&siginfo);
+	if (ret != -1) {
```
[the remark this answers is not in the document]
abcSup: Based on the ptrace manuals, it says that most of the ptrace commands will fail if the target is not stopped and I believe `PTRACE_GETSIGINFO` is included. I am not checking the content because of that and if it contains other signals, we can receive those signals later in the next `waitpid`.

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -472,19 +514,16 @@ RDebugReasonType linux_dbg_wait(RDebug *dbg, int my_pid) {
...
-	dbg->reason.tid = pid;
 	return reason;
 }
 
 int match_pid(const void *pid_o, const void *th_o) {
 	int pid = *(int *)pid_o;
-	RDebug *th = (RDebug *)th_o;
-	return pid == th->pid;
+	RDebugPid *th = (RDebugPid *)th_o;
+	return (pid == th->pid) ? 0 : 1;
```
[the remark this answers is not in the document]
abcSup: It seems like `r_list_find` will return when the comparator returns false.

https://github.com/radareorg/radare2/blob/3b76cc25c815d9f97d711cfc033e001a7c264cb7/libr/util/list.c#L459-L471

libr/debug/p/native/linux/linux_debug.c
abcSup: This is handled in the `linux_dbg_wait` which will look for `SIGTRAP` in other threads when it hits a `SIGSTOP` caused by newly created thread/process. I could probably move that handling to here because iirc there is a weird bug forces me to do it in `linux_dbg_wait`. I will check it again.

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -584,37 +624,17 @@ static RList *get_pid_thread_list(RDebug *dbg, int main_pid) {
...
 	// First time we run: We try to attach to all "possible" threads and to the main pid
 	if (!dbg->threads) {
 		dbg->threads = get_pid_thread_list (dbg, pid);
-		linux_attach_all (dbg);
 	} else {
 		// This means we did a first run, so we probably attached to all possible threads already.
 		// So check if the requested thread is being traced already. If yes: skip
 		if (dbg->threads && !r_list_find (dbg->threads, &pid, &match_pid)) {
-			goto out;
+			linux_attach_single_pid (dbg, pid);
```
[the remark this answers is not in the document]
abcSup: You are right. I was confused by the function of `match_pid` at first.

libr/debug/p/native/linux/linux_debug.c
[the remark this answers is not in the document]
yossizap: You're right, sorry was confused by the previous implementation. We should look into this on rlist's side though.

libr/debug/p/native/linux/linux_debug.c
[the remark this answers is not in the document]
yossizap: > will fail if the target is not stopped

So you will return from the function without stopping the thread in the current implementation.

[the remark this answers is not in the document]
yossizap: Checking signals with getsignals won't remove them, you still need to waitpid for that. Relying on this error sounds like undefined behaviour for any case except for checking if the debugger is attached. You can check errno to see if actually fails because the process is running.

libr/debug/p/native/linux/linux_debug.c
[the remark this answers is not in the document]
abcSup: The statement actually runs when it did not find an existing thread because `!r_list_find (dbg->threads, &pid, &match_pid)` will be true when `r_list_find` returns `NULL` and there is no existing thread with the same pid.

libr/debug/p/native/linux/linux_debug.c
abcSup: This is actually checked in r_debug_native_continue for SIGTRAP sent with `si_pid == 0` when attaching.

libr/debug/p/native/linux/linux_debug.c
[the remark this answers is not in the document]
abcSup: > So you will return from the function without stopping the thread in the current implementation.

That was my bad and I see the problem now. It should return when success and `PTRACE_GETSIGINFO` returns 0.

> Checking signals with getsignals won't remove them, you still need to waitpid for that. Relying on this error sounds like undefined behaviour for any case except for checking if the debugger is attached. You can check errno to see if it actually fails because the process is running.

I agree with you on this. `PTRACE_GETSIGINFO` will not remove the signals and I will need to `waitpid` to consume them. Basically, `r_debug_native_continue` will check for them and consume them accordingly. This also answers your another comment regarding the `linux_attach_single_pid` in which the thread always starts with a SIGSTOP after being attached. This make me realize that I could have done the removing signal part in here instead which will be included in my next commit. Thanks

libr/debug/p/native/linux/linux_debug.c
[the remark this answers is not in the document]
abcSup: Before this, `match_pid` is implemented differently and correcting it reverses the condition. I will update the comment instead.

libr/debug/p/native/linux/linux_debug.c
abcSup: Because every `fork` and `clone` will stop the parent and its child together, I will `PTRACE_CONT` one of them based on the debugger option `dbg->trace_clone` and `dbg->trace_fork` and another will be assigned to `dbg->tid` to be handled outside.

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -435,20 +449,48 @@ RDebugReasonType linux_dbg_wait(RDebug *dbg, int my_pid) {
 				eprintf ("child received signal %d\n", WTERMSIG (status));
 				reason = R_DEBUG_REASON_SIGNAL;
 			} else if (WIFSTOPPED (status)) {
-				if (WSTOPSIG (status) != SIGTRAP &&
-					WSTOPSIG (status) != SIGSTOP) {
-					if (WSTOPSIG (status) == SIGINT) {
-						reason = R_DEBUG_REASON_USERSUSP;
+				// A new task is created with SIGSTOP and its parent is stopped with SIGTRAP
+				// Find SIGTRAP in other threads to see if that is the case
+				if (WSTOPSIG (status) == SIGSTOP) {
```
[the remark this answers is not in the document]
abcSup: > Can you add a test? will be good to have more coverage

@trufae I wish to add some tests for this which will help testing the debugger, but I will need to think of a way to send `SIGINT` to `r2` to check if the user interrupt is working which might require some time on implementing that in `r2r`.

abcSup:
> Can you add a test? will be good to have more coverage

@trufae I wish to add some tests for this which will help testing the debugger, but I will need to think of a way to send `SIGINT` to `r2` to check if the user interrupt is working which might require some time on implementing that in `r2r`.

XVilka:
@abcSup @yossizap what is the status of this PR?

abcSup:
> @abcSup @yossizap what is the status of this PR?

There is a recent change in the RR PR to fix RCons break in `linux_debug.c`, so I would like to wait for the RR PR to be merged first and deal with the merge conflict later in this PR.

XVilka:
@abcSup it was merged.

abcSup:
> @abcSup it was merged.

Thank you! I will push an update to resolve the conflict soon today.

libr/debug/debug.c
abcSup: After reviewing this agian, `r_debug_select` should be called here actually. However, some code in the debug_native plugin call this directly to switch to a thread/process and get information from it, which is messing with `RDebug`'s state and should be avoided.

libr/debug/p/native/linux/linux_debug.c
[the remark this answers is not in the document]
abcSup: A new function to deal with new task that is not already in the `dbg->threads` will be called `linux_handle_new_task`.

abcSup:
Updating the PR to make tests green and resolve radareorg/cutter#1931 issue.

abcSup:
A new commit to fix the debugger freezes when interrupted by the user in an attach session `r2 -d <pid>` and a side effect. `linux_dbg_wait` always send a SIGINT signal to the debuggee when interrupted. However, this will have a side effect as noted in the code before my RR PR, which is the debugger will break two times in a row due to SIGINT sent by the debugger and SIGINT sent by the user when interrupting the debugger. To prevent that side effect and make sure the debugger breaks out of the wait when interrupted, we check for the process group of the debugger and the debuggee and send the SIGINT if the debuggee is not in the same process group. The same explanation is documented in the code for future reference.

I have also tested radareorg/cutter#1931 in the following way with the sample code in the issue:
```
$ r2 -d issue-1931
Process with PID 97347 started...
= attach 97347 97347
bin.baddr 0x559df2e5f000
Using 0x559df2e5f000
asm.bits 64
 -- THIS IS NOT FUNNY
[0x7f2fdb552090]> db sym.thread_func+48
[0x7f2fdb552090]> dc
(97347) Created thread 97348
(97347) Created thread 97349
hit breakpoint at: 0x559df2e5fe2a
[0x559df2e5fe2a]> dpt
 - 97347 s /lib/x86_64-linux-gnu/libpthread-2.27.so (0x7f2fdad99d2d)
 - 97348 s /home/abcsup/test-c/1931 (0x559df2e5fe2b)
 * 97349 s /home/abcsup/test-c/1931 (0x559df2e5fe2a)
[0x559df2e5fe2a]> dpt=97348
= attach 97347 97348
[0x559df2e5fe2b]> ds
[0x559df2e5fe2b]> dc
hit breakpoint at: 0x559df2e5fe2a
[0x559df2e5fe2a]> dpt
 - 97347 s /lib/x86_64-linux-gnu/libpthread-2.27.so (0x7f2fdad99d2d)
 * 97348 s /home/abcsup/test-c/1931 (0x559df2e5fe2a)
 - 97349 s /home/abcsup/test-c/1931 (0x559df2e5fe2a)
[0x559df2e5fe2a]> 

```

abcSup:
I hope to add some tests for this test, but we will need a way to send interrupt to the debugger in `r2r` during a session, and to select a tid from the thread list`dpt` and switch to it with `dpt=<tid>`.

XVilka:
@yossizap @ret2libc please take a look again.

libr/debug/p/debug_native.c
```diff
@@ -240,28 +242,34 @@ static int r_debug_native_continue(RDebug *dbg, int pid, int tid, int sig) {
...
-					}
+		pid_t dbgpid = getpid ();
+		r_list_foreach (dbg->threads, it, th) {
+			siginfo_t siginfo = { 0 };
+			ret = r_debug_ptrace (dbg, PTRACE_GETSIGINFO, th->pid, 0, (r_ptrace_data_t)(size_t)&siginfo);
+			// Skip if the thread is already running
+			if (ret == -1) {
+				continue;
+			}
+			if (siginfo.si_signo == SIGSTOP && siginfo.si_pid == dbgpid) {
```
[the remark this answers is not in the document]
abcSup: > This will fail forever for the original thread after switching to a different thread with dpt=. The thread won't resume even if you re-attach to it.

You are right.

> Why not continue if there's another signal? It should be waitpid'd and handled.

I have noticed that even if there is a pending signal in the thread, PTRACE_CONT will continue the thread without errors and the pending signal will be lost after the that thread continues running.

For example when, a thread gets signaled before it is stopped by the debugger's SIGSTOP during linux_stop_threads, the pending signal is in the queue, so continuing all threads even if there is another signal will cause the pending signal to be lost:

1. Thread A creates Thread B.
2. `linux_dbg_wait` breaks and returns `R_DEBUG_REASON_NEW_TID`
2. `r_debug_wait` calls `r_debug_stop` and `linux_stop_threads` which tries to stop all running threads.
2. Thread B exits immediately and Thread A traps into `SIGTRAP` with `PTRACE_EVENT_EXIT` before it gets stopped by `linux_stop_threads`. `linux_stop_threads` will not send another `SIGSTOP` to Thread A because it is stopped.
3.  User continues the process. If we continue all thread even if there is a pending signal in Thread A (`SIGTRAP`), it will continue it without errors. The pending signal in Thread A is lost after it is continued by `PTRACE_CONT`.

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -126,12 +130,11 @@ int linux_handle_signals (RDebug *dbg, int tid) {
 				} else {
 					dbg->reason.bp_addr = (ut64)(size_t)siginfo.si_addr;
 					dbg->reason.type = R_DEBUG_REASON_BREAKPOINT;
-					// Switch to the thread that hit the breakpoint
-					r_debug_select (dbg, dbg->pid, tid);
-					dbg->tid = tid;
 				}
 			}
-		}
+		} break;
```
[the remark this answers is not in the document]
abcSup: It can be inside as well. That `break` is to prevent it from falling through and running code below.

I usually see it outside, for example:

https://github.com/radareorg/radare2/blob/154416c8fd4ce0a399c42e27ba88a0c535e320a0/libr/core/cmd_debug.c#L4859-L4864

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -172,7 +175,7 @@ static void linux_remove_fork_bps(RDebug *dbg) {
 	// Set dbg tid to the new child temporarily
 	dbg->pid = dbg->forked_pid;
 	dbg->tid = dbg->forked_pid;
-	r_debug_select (dbg, dbg->forked_pid, dbg->forked_pid);
+	// r_debug_select (dbg, dbg->forked_pid, dbg->forked_pid);
```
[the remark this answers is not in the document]
abcSup: This should be removed because it will call `linux_attach_new_process` which calls `linux_detach_all` before attaching to the forked process. This is problematic because `PTRACE_DETACH` will continue the parent process before removing its breakpoints and the parent process might terminate before we have removed all breakpoints in the forked process and return to it.

The best way to get access to registers and memory is to temporarily change the `dbg->pid` and `dbg->tid`, so calls like `r_debug_reg_sync` can succeed.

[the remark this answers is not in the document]
yossizap: Please run a test program with a fork and a breakpoint after the fork, will child will simply die. This still has to be solved.

If you don't select the forked process, the code will still run and remove breakpoints from the parent that shouldn't be removed, I am not sure you understood this function. If debug_select was changed to include detach_all you need to call linux_attach_new_process directly to solve this issue or create a new function to handle forked processes.

libr/debug/p/debug_native.c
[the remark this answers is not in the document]
yossizap: The new solution simply overwrites the pending signal in the child, isn't there a chance we'll miss a signal if two threads had an event at the same time? I think that you should handle the signal properly instead of just resetting it to SIGSTOP.

libr/debug/p/native/linux/linux_debug.c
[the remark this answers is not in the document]
abcSup: I tested what you said and you are right. Besides, what I said is totally wrong, since `dbg->pid` is already set to the forked `pid` before doing `r_debug_select`, so it will attach the forked process as a new process and detaching will not happen. I actually forgot the actual reason I commented this out, but I know that it will cause some ptrace calls to fail and printing errors in a specific situation. I will just revert this back to what it was before then, since it does not affect my PR in any way. After all, thanks for pointing this out as well.

libr/debug/p/native/linux/linux_debug.c
abcSup: `linux_add_and_attach_new_thread` is only used in `linux_ptrace_event` when `PTRACE_EVENT_CLONE` happens to add the new thread into the thread list. The initial `SIGSTOP` is handled in `linux_ptrace_event` or by `waitpid` depending on whether `waitpid` got the `SIGTRAP` from the parent or the initial `SIGSTOP` from the child first.

libr/debug/p/debug_native.c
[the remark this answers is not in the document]
abcSup: I will open a new issue to triage this problem, since it requires more than handling the signals properly and also dealing with thread switching. Everything happens in an asynchronous manner. For example, stopping and continuing threads cannot be done to all threads at once and new signals can be sent in between calls. When every thread is not properly tracked, we might accidentally continue a thread that has a pending signal, which causes the signal to be lost and sometimes the debugger freezes.

libr/debug/p/debug_native.c
```diff
@@ -242,36 +241,22 @@ static int r_debug_native_continue(RDebug *dbg, int pid, int tid, int sig) {
 		r_cons_break_push ((RConsBreak)interrupt_process, dbg);
 	}
 
-	ret = r_debug_ptrace (dbg, PTRACE_CONT, tid, NULL, (r_ptrace_data_t)(size_t)contsig);
-	if (ret == -1) {
-		r_sys_perror ("PTRACE_CONT");
-	}
```
yossizap, line 244: Extra newline

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -96,7 +95,7 @@ int linux_handle_signals(RDebug *dbg, int tid) {
 		dbg->stopaddr = (ut64)siginfo.si_addr;
 		//dbg->errno = siginfo.si_errno;
 		// siginfo.si_code -> HWBKPT, USER, KERNEL or WHAT
-		// TODO: DO MORE RDEBUGREASON HERE
+// TODO: DO MORE RDEBUGREASON HERE
```
yossizap: Fix indent

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -85,9 +85,8 @@ int linux_handle_signals(RDebug *dbg, int tid) {
 	}
 
 	if (siginfo.si_signo > 0) {
-		// Switch the target thread to the signaled thread to allow access
-		// to its registers later.
-		dbg->tid = tid;
+		// Select the signaled thread to allow access to its info later.
+		r_debug_select (dbg, dbg->pid, tid);
```
yossizap: Are you sure that this is necessary? Seems to work without it and this causes the thread to always return to the main pid after an interrupt.

libr/debug/p/native/linux/linux_debug.c
```diff
@@ -272,6 +315,49 @@ RDebugReasonType linux_ptrace_event (RDebug *dbg, int pid, int status) {
...
+			// Skip if PTRACE_GETSIGINFO fails when the thread is running.
+			if (ret == -1) {
+				continue;
+			}
+			if (siginfo.si_signo == SIGTRAP) {
+				// si_code = (SIGTRAP | PTRACE_EVENT_* << 8)
+				int pt_evt = siginfo.si_code >> 8;
+				// Handle PTRACE_EVENT_* that creates a new task (fork/clone)
+				switch (pt_evt) {
+				case PTRACE_EVENT_CLONE:
```
XVilka, line 347: Since you started to use `PTRACE_EVENT_*`, could you please add them in [libr/debug/p/native/linux/linux_ptrace.h](https://github.com/radareorg/radare2/blob/master/libr/debug/p/native/linux/linux_ptrace.h)? So it will still be correct for older Glibc, like the one from Debian Etch (see https://github.com/radareorg/radare2/pull/17459 for more information).

abcSup: They were used before this as well, but sure!

libr/debug/p/native/linux/linux_debug.c
abcSup: `r_debug_reg_get` will fail if the debugger is targeting a running thread instead of the one returned by `waitpid` when the code below runs:
```
if (dbg->glob_libs || dbg->glob_unlibs) {
	ut64 pc_addr = r_debug_reg_get (dbg, "PC");
	if (b && b->internal) {
```
I realized that setting `dbg->tid` directly is ugly and also using `r_debug_select` will cause other issues. I will just remove the code here then as well and create an issue for that.

abcSup: Some parts of the code call `r_debug_reg_get` without switching to the target thread and will fail in a multi-threaded process.

libr/debug/p/native/linux/linux_debug.c
abcSup: Added PTRACE_EVENT_FORK, PTRACE_EVENT_VFORK, PTRACE_EVENT_CLONE, PTRACE_EVENT_EXEC, PTRACE_EVENT_VFORK_DONE, PTRACE_EVENT_EXIT

libr/debug/p/native/linux/linux_debug.c
yossizap: Alrighty. Don't forget to open a new issue for this :)

Issue #10115, opened 2018-05-17
Meson + clang-cl with ASAN, UBSAN, etc build for Windows


XVilka:
We have enough power at our paid AppVeyor account. Would be nice to have the version for Windows with all sanitizers enabled, which will help us to catch the bugs easier.

See https://clang.llvm.org/docs/MSVCCompatibility.html and Google Chrome success story http://blog.llvm.org/2018/03/clang-is-now-used-to-build-chrome-for.html

XVilka:
@pelijah it was fixed by https://github.com/mesonbuild/meson/pull/4250

pelijah:
```
C:\Program Files (x86)\Windows Kits\10\include\10.0.17134.0\um\winnt.h:2653:1: error: conflicting types for '_bittest'
_bittest (
^
C:\Program Files (x86)\Windows Kits\10\include\10.0.17134.0\um\winnt.h:2653:1: note: '_bittest' is a builtin with type 'unsigned char (const long *, long)'
C:\Program Files (x86)\Windows Kits\10\include\10.0.17134.0\um\winnt.h:2677:1: error: conflicting types for '_interlockedbittestandset'
_interlockedbittestandset (
^
C:\Program Files (x86)\Windows Kits\10\include\10.0.17134.0\um\winnt.h:2677:1: note: '_interlockedbittestandset' is a builtin with type 'unsigned char (volatile long *, long)'
C:\Program Files (x86)\Windows Kits\10\include\10.0.17134.0\um\winnt.h:2683:1: error: conflicting types for '_interlockedbittestandreset'
_interlockedbittestandreset (
^
C:\Program Files (x86)\Windows Kits\10\include\10.0.17134.0\um\winnt.h:2683:1: note: '_interlockedbittestandreset' is a builtin with type 'unsigned char (volatile long *, long)'
C:\Program Files (x86)\Windows Kits\10\include\10.0.17134.0\um\winnt.h:2689:1: error: conflicting types for '_bittest64'
_bittest64 (
^
C:\Program Files (x86)\Windows Kits\10\include\10.0.17134.0\um\winnt.h:2689:1: note: '_bittest64' is a builtin with type 'unsigned char (const long long *, long long)'
C:\Program Files (x86)\Windows Kits\10\include\10.0.17134.0\um\winnt.h:2713:1: error: conflicting types for '_interlockedbittestandset64'
_interlockedbittestandset64 (
^
[...]

PR #17252, opened 2020-07-10
It seems we don't really require such a high version of meson

ret2libc:
Some systems still ship older meson version and since it doesn't seem we
require such a high version, let's just require an older one.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,4 +1,4 @@
-project('radare2', 'c', license: 'LGPL3', meson_version: '>=0.50.1')
+project('radare2', 'c', license: 'LGPL3', meson_version: '>=0.49.2')
 
 py3_exe = import('python').find_installation('python3')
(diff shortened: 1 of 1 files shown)


ret2libc:
> I think it was `clang-cl` problems that were fixed in 0.50 release:
> 
>     * [mesonbuild/meson#4250](https://github.com/mesonbuild/meson/pull/4250)
> 
>     * #10115

I see.. Maybe we just need to keep 0.50.1 then. Thanks for the info.

trufae:
I will try to find some time to understand why meson builds always fail with funky errors in my systems and fix the symstall dirs issue too

trufae:
I have upgraded from 0.49.2 to 0.54.3 and it seems to build fine, will do some more tests and report if i find anything failing again

ret2libc:
Let's close this given what said by @xvilka

PR #16621, opened 2020-04-17
WIP: Unconventional code char encoding ##print

gogo2464:
New draft pull request to show changes from https://github.com/radareorg/radare2/pull/16357. I have implemented the function r_charset_encode_str().

Now I need to modify the structure used in r_charset_encode_str() with pst by the command pse.

**PAD** https://hackmd.io/axFbkdadQ3qML85nDgFzHQ

  

**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

My modifications permit to:
-list plugins
-load char encoding plugin
-use the plugin to print a string with a different char encoding

The command `pse` lists all the plugin on the specific folder for char encoding plugins.
The plugin works by using the command `k`.
Once the plugin is loaded we can convert and see encoding with `pst`.

Some new folders have been added in the project with the makefile to handle plugins.

I did not have upated the documentation of radare2 yet
my plugin does not handle special chars and does not handle spaces. I use `-` instead

...

**Test plan**

First download the gameboy rom from https://github.com/MrCheeze/pokered-self-replicator
`radare2 sav.dat`
`pse` to list plugins
`pse pokered` to load pokemon game boy red char encoding plugin
`pst @ 0x2528` to read a string on the gameboy rom. The charset will not be loaded from `pse` yet.
You should see 

> Selfreplicating----arbitrary-code------by-MrCheeze-:)---------------------\\\\AvvJOHN\\\\]V

Then the test is OK.
...

**Closing issues**

will close https://github.com/radareorg/radare2/issues/414
...

PR #17256, opened 2020-07-11
Add rebasing instructions

ulidtko:
This just adds a section to CONTRIBUTING.md teaching how to rebase. I'm also put in JeisonWi's link to a git guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ try to keep the codebase consistent and clean.
 
 ## Getting Started
 
-* Make sure you have a GitHub account.
+* Make sure you have a GitHub account and solid ability to use `git`.
 * Fork the repository on GitHub.
 * Create a topic branch from master. Please avoid working directly on the `master` branch.
 * Make commits of logical units.
@@ -29,6 +29,49 @@ try to keep the codebase consistent and clean.
 * Submit the Pull Request(PR) on Github.
 * When relevant, write a test in [test/](test).
 
+## Rebasing onto updated master
+
+Every so often, your PR will lag behind `master` and get conflicts.
+
+To "update" your branch `my-awesome-feature`, you *rebase* it onto
+the latest `radareorg/master`, and *force-push* the result into your fork.
+
+#### Step 1: Switch to `master` branch.
+
+    $ git checkout master
+
+#### Step 2: Pull new commits published to radareorg repo.
+
+    $ git pull https://github.com/radareorg/radare2
+
+#### Step 3: Switch back to `my-awesome-feature` branch.
+
+    $ git checkout my-awesome-feature
+
+#### Step 4: Rebase the `my-awesome-feature` branch.
+
+    $ git rebase master
+
+Optionally, use the alternative mode "interactive rebase". It allows
+to `squash` your commits all into one, reorder, reword them, etc.
+
+    $ git rebase -i master
+
+Follow git instructions when conflicts arise.
+
+#### Step 5: publish your updated local branch.
+
+    $ git push -f
+
+This `-f` *force-flag* is needed because git commits are immutable: rebasing
+creates newer versions of them. git needs to confirm the destruction of
+previous incarnations.
+
+When afraid to touch force and risk losing your work (do backups!..),
+try *merging master into your branch* instead of rebasing onto it.
+This is discouraged, as it produces ugly hard-to-maintain commit history.
+
+
 # Additional resources
 
  * [README.md][]
@@ -36,3 +79,6 @@ try to keep the codebase consistent and clean.
 
 [README.md]: https://github.com/radareorg/radare2/blob/master/README.md
 [DEVELOPERS.md]: https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md
+
+If you need more confidence in your git skills, check out this quick guide:
+<https://learnxinyminutes.com/docs/git/>
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,13 +19,20 @@ try to keep the codebase consistent and clean.
 
 * Make sure you have a GitHub account.
 * Fork the repository on GitHub.
-* Create a topic branch from master. Please avoid working directly on the ```master``` branch.
+* Create a topic branch from master. Please avoid working directly on the `master` branch.
 * Make commits of logical units.
-* Check for coding style issues with ```git diff master..mybranch | ./sys/clang-format-diff.py -p1``` and be sure to follow the CODINGSTYLE (more on this in [DEVELOPERS.md](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md)).
+* Check for coding style issues with:
+
+      git diff master..mybranch | ./sys/clang-format-diff.py -p1
+
+  and be sure to follow the CODINGSTYLE (more on this in [DEVELOPERS.md][]).
 * Submit the Pull Request(PR) on Github.
 * When relevant, write a test in [test/](test).
 
 # Additional resources
 
-* [README.md](https://github.com/radareorg/radare2/blob/master/README.md)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -17,15 +17,63 @@ try to keep the codebase consistent and clean.
...
+
+Every so often, your PR will lag behind `master` and get conflicts.
+
+To "update" your branch `my-awesome-feature`, you *rebase* it onto
+the latest `radareorg/master`, and *force-push* the result into your fork.
+
+Be sure to understand *git remotes* as you need working with 3 git repos
+simultaneously: 1: your local clone, 2: radareorg GH repo, 3: your GH fork.
+
+For example, user Joe rebases his `my-awesome-feature` and using
```
ret2libc, line 42: why do you need `git remote rename` and `git remote set-url`? I think mentioning them just creates confusion as I've never used them for r2.

CONTRIBUTING.md
```diff
@@ -17,15 +17,63 @@ try to keep the codebase consistent and clean.
...
+
+  and be sure to follow the CODINGSTYLE (more on this in [DEVELOPERS.md][]).
 * Submit the Pull Request(PR) on Github.
 * When relevant, write a test in [test/](test).
 
+## Rebasing onto updated master
+
+Every so often, your PR will lag behind `master` and get conflicts.
+
+To "update" your branch `my-awesome-feature`, you *rebase* it onto
```
ret2libc, line 36: Actually it is also possible to merge master into your branch, which is IMHO easier/faster at times, because you just need to resolve the conflicts one time instead of, possibly, N times, one for each of your commit.

CONTRIBUTING.md
wargio: for me this should be simplified by saying: 

1. Fork the radareorg/radare2 repository
2. Clone the forked repository
3. Create a new branch
4. Do the changes and commit them
5. Push everything on the forked repository
6. Open a PR and wait to be merged
7. Once merged, checkout to master
8. Add remote to radareorg/radare2 repo (step that needs to be done only once)
9. Git fetch, rebase and push

additional steps in case of needed rebase on the PR

1. Add remote to radareorg/radare2 repo (step that needs to be done only once)
2. Git fetch
3. Rebase interactively and fix any conflicts
4. Git push `-f`

ulidtko: @ret2libc because people get confused about what `origin` is; when cloning the fork-repo, `origin` won't point to the "logical" origin -- it will point to the fork. As I'm describing specifically the case of conflicts with master: adding the second `radareorg` remote will be unavoidable. For clueless folks, it gets really messy real quick; that's why I'm pointing out the basics of managing remotes, and (indirectly) what they even are.

You didn't hit this stuff — simply because you're collaborator. You can push branches directly to `radareorg`, and need no fancy remote manipulations. Outside folks do need to deal with that; see #16621.

@wargio my list has 3 steps, yours is 9... in what sense is that simplification? :)
Please notice the section header and 1st paragraph. No need to duplicate the already existing "Getting started" thing. This section is *specifically* for rebasing old PRs with conflicts.

CONTRIBUTING.md
ulidtko: Well yes, merging-vs-rebasing is a hot topic in itself, there're both pros & cons to each...

But from the perspective of project maintainer — I find that *merging master into feature branch* produces absolutely disgusting history graphs. It takes way more effort to make any sense of such history (should anything need go wrong and need reverting, cherry-picking, cross-merging, etc) — compared to the neat linear rebased-feature-branches.

CONTRIBUTING.md
ret2libc: > You didn't hit this stuff — simply because you're collaborator. You can push branches directly to `radareorg`, and need no fancy remote manipulations. Outside folks do need to deal with that; see #16621.

Even though I'm a collaborator I hardly, if ever, push directly to branches. I do work with multiple remotes: mine, radareorg one, several other contributors. But I never used `git remote rename` or never had to use `git remote set-url` (said in other words, I did use them, but for doing very different stuff. And if a user needs to read this basic git stuff, I think mentioning these things would just confuse them even more, because they are probably not very familiar with git in the first place).

CONTRIBUTING.md
ret2libc: Yes, however 99% of the time we do not merge multiple commits in master, but we do squash+merge. This means that even if you have your super clean history of commits on top of master (and I assure you that 99% of the times you don't), everything will be reduced to one single commit.

For this reason using rebasing or merging, when updating an old PR, doesn't really matter, because at the end of the day everything will be unified in one single commit. So, knowing that, a `git merge origin/master` in my PR-branch is what I usually do as I need to resolve conflicts just one time and the end result is the same (plus for me that means I don't have to use `--force` ever, because as I have permission to write on multiple branches, a small mistake somewhere could make me force-push wrong stuff on some branches).

CONTRIBUTING.md
wargio: @ulidtko i actually always push to my fork, even if i'm a collab. and all these steps allows people who are not used to git to understand each step.

CONTRIBUTING.md
trufae: Both reasonings make sense to me, but i'm more with @ulidtko , it requires more work to rebase+cherrypick.. than just merge master .. which only looks fine in github if you squash+rebase from the github web. otherwise it's a mess and it becomes much more confusing to reference commits by index and merge commits appear empty while they group other ones.. so imho we can say that:

* merge is fine if the whole PR will end up being 1 squashed commit and you let others to commit or your PR will take a lot to be finished
* rebase if you know what you are the only commiter in there and you have proper git skills to not loose your changes

Additionally we can also explain how

CONTRIBUTING.md
trufae: My steps use to be like this:

* git checkout -b my-cool-feature
* do changes in code and commit
* git checkout master
* git pull https://github.com/radareorg/radare2
* git checkout my-cool-feature
* git rebase master

trufae: without remotes or origins, its more implicit and may be less confusing for the reader

ulidtko: Ah, I often forget that git accepts full URLs in place of `remote` args.

Also yes, updating local master should be part of the steps, as not doing it (and not saying explicitly `origin/master`) is very often made mistake.

CONTRIBUTING.md
ulidtko: OK so mentioning `merge master` here will require mentioning `squash` -- which requires going more in-depth into interactive rebase. That's rather big detour off the flow. You sure want to put this into `CONTRIBUTING.md` ? I can do it no problem, it's just a bit too much IMO.

ulidtko:
Gave it one more spin (git remotes stuff dropped), please re-review.

CONTRIBUTING.md
ret2libc: merging master into the pr doesn't require using squash at all. Never used `squash`. The way we work and we have always done in the last X years (don't remember exactly how much) is that we squash+merge PRs into master. So even if your PR has 10 commits, they will appear as one single commit in master. We even have squash+merge as the default button for merging PRs.

ret2libc:
> Looks good to me, i agree with @ret2libc
> 
> I also think it will be good to explain about git reset @^^ and git reset --hard, as well as cherry-pick... so not sure if we should have kind of doc/git.md pointed out from CONTRIBUTING.md for those who want to learn the basics of git without having to use google, but we end up wanting our contributors to follow some specific workflows, so i guess that will help us point them to a resource we all agree

I wouldn't go into the path of git-explanation too much. We are not a git-learning framework, people should google :D If they can't, we provide a link to some learning resource, but that's more than enough IMHO. If they can't find out how to do things even at that point... Well, maybe github/opensource is not for them :)

PR #17259, opened 2020-07-13
Check for ferror(), not only for feof()

### LINKED ISSUE #17257: Visual mode freeze when entering debugger panel
### Work environment

| OS/arch/bits (mandatory)                             | Archlinux
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.5.0-git 24916 @ linux-x86-64 git.4.4.0-411-g7ee93fb45
|commit | 7ee93fb4573a7c165100483ff8848e3e3c84880e build: 2020-07-11__14:27:27

### Expected behavior

Enter debugger panel and show memory graph. om should show correct mapped spaces.

### Actual behavior

r2 hang when using Vpp. om shows only one block with rwx permission. strace shows it hangs at futex. 

[target.zip](https://github.com/radareorg/radare2/files/4907555/target.zip)
```
r2 -d a.out
om
Vpp
```

_strace log when it hangs:_

> [pid 249427] read(8, 0x558bf3558120, 1024) = -1 EACCES (Permission denied)
> [pid 249427] read(8, 0x558bf3558120, 1024) = -1 EACCES (Permission denied)
> [pid 249427] read(8, 0x558bf3558120, 1024) = -1 EACCES (Permission denied)
> [pid 249427] read(8, 0x558bf3558120, 1024) = -1 EACCES (Permission denied)
> [pid 249427] read(8, 0x558bf3558120, 1024) = -1 EACCES (Permission denied)
> [pid 249427] read(8, 0x558bf3558120, 1024) = -1 EACCES (Permission denied)
> [pid 249427] read(8, 0x558bf3558120, 1024) = -1 EACCES (Permission denied)
> [pid 249427] read(8, 0x558bf3558120, 1024) = -1 EACCES (Permission denied)
> ....

ret2libc:
Some files could not be read and would create an infinite loop during
the execution of r_file_slurp. This patch fixes that issue: when an
error occur at read time, it detects it with ferror(fd) and it returns
an empty buffer.

 

**Detailed description**

Among the info showed by `di` there is the "Kernel stack". On linux debugger this is taken by reading the `/proc/<pid>/stack` file, which is read with a `r_file_slurp` call. `r_file_slurp` continuously reads data with `fread` until the end of file is reached. However, it does not consider errors. On some systems you can open `/proc/<pid>/stack` but you are not allowed to read it, so the `fread` call generates an error that is never caught and the loop never ends.

This patch checks for the error and returns an empty buffer in such a case.

**Test plan**

Just see the issue, there are steps to reproduce.

**Closing issues**

Closes #17257 .


wargio:
shouldn't be the check after `fseek` at line 342? what about trying to read a byte and check the value of ferror?

ret2libc:
> shouldn't be the check after `fseek` at line 342?

the fseek manpage does not mention to check ferror, while fread manpage does. I'm not sure.

>  what about trying to read a byte and check the value of ferror?

Why the additional complexity of checking one byte + storing one byte in the buffer + updating the size of one byte + reading in chunks, when you can just do one general code to handle everything?

XVilka:
We should handle missing "/proc/<pid>/stack" better though, on a higher level - maybe show the warning to the user or something like that.

ret2libc:
> We should handle missing "/proc//stack" better though, on a higher level - maybe show the warning to the user or something like that.

I agree. I'm going to open an issue for this. It's still not clear to me how that works in recent kernels, because I think they have restricted the set of processes who can access that file.

libr/util/file.c
```diff
@@ -360,7 +360,11 @@ R_API char *r_file_slurp(const char *str, R_NULLABLE size_t *usz) {
 				buf = nbuf;
 				size_t r = fread (buf + size, 1, BS, fd);
 				size += r;
-			} while (!feof (fd));
+			} while (!feof (fd) && !ferror (fd));
```
trufae, line 363: call ferror once inside the loop and break or return NULL directly? in there before incrementing the size? usz doesnt needs to be set if the read fails

libr/util/file.c
```diff
@@ -360,7 +360,11 @@ R_API char *r_file_slurp(const char *str, R_NULLABLE size_t *usz) {
...
+			if (ferror (fd)) {
+				size = 0;
+				R_FREE (buf);
+			}
 			if (usz) {
 				*usz = size;
 			}
```
trufae, line 370: fclose(fd) is never called

trufae:
Screenshot 2020-07-14 at 18 41 06

trufae:
Please fix the PR title. it doesn't explain where the fix is or for what it is and which ##tag covers

trufae:
Something like: "Fix infinite loop regression in Linux debugger by using ferror in addition to feof() ##debug"

PR #17156, opened 2020-06-27
#!pipe: Fix stray command output on Windows and add r2pipe.py test

kazarmy:
**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [X] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

This pr fixes a `. <script>` (aka #!pipe) problem on Windows where the commands received were erroneously printed out after the script had executed. Also, adds an r2pipe.py test. 

**Test plan**

All builds are green.

**Closing issues**

...

PR #17269, opened 2020-07-13
Id next

condret:

**Detailed description**
Let me be honest with you: From my perspective this code is mostly useless, except for the use case of #17162 . RIDStorage does not provide any specific ordering of the stored values, so if you look at a specific value, that is stored in an id_storage, there is no such thing as next or previous value. However this code pr is useful for cycling through values in an unspecific order, just as in #17156.

So why is there r_io_fd_get_next and r_io_desc_get_next? Wouldn't it be enough to have only one of them?
The reason for that is, that I want the io api to be used as painless as possible. Imo that means, that projects that only use the map and fd api shouldn't be forced to switch between desc and fd to use this.

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/util/idpool.c
+++ b/libr/util/idpool.c
@@ -163,6 +163,35 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
 	return storage->data[id];
 }
 
+// usually these shall not be used
+// id-storage doesn't provide sorting things in specific order
+// there are only a very few reasonable usecases, such as cycling through RIODescs
+R_API bool r_id_storage_get_next(RIDStorage *storage, ut32 *id) {
+	r_return_val_if_fail (id && storage && storage->data, false);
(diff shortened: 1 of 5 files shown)


libr/util/idpool.c
```diff
@@ -163,6 +163,35 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+}
+
+R_API bool r_id_storage_get_prev(RIDStorage *storage, ut32 *id) {
+	r_return_val_if_fail (id && storage && storage->data, false);
+	if (storage->id <= *id) {
+		return false;
+	}
+
+	do {
+		*id = (*id + (storage->size - 1)) % storage->size;
```
trufae, line 190: it is usually faster and safer to hold the *id value in a local variable until the function returns to save the result back. so it doesnt needs to access the stack to dot hat. compilers cant optimize this because behaviour differs, and affects threading

Issue #16353, opened 2020-03-29
I Can't Use A Saved Project


RHL120:
I am new to Radare so this is probably my fault not Radare's.
To save a project I use Ps proejectname
but whenever I open a project with r2 -p projectname or Po project name (within Radare) I get
afb+: Cannot add basic block at 0x00001040
afB: Cannot find function to set bits at 0x00001040
afc: Cannot find function here
Missing function at 0x00001040
afb+: Cannot add basic block at 0x00001070
afb+: Cannot add basic block at 0x00001083
afb+: Cannot add basic block at 0x0000108f
afb+: Cannot add basic block at 0x00001098
afB: Cannot find function to set bits at 0x00001070
afc: Cannot find function here
afb+: Cannot add basic block at 0x000010a0
afb+: Cannot add basic block at 0x000010c4
afb+: Cannot add basic block at 0x000010d0
afb+: Cannot add basic block at 0x000010d8
afB: Cannot find function to set bits at 0x000010a0
afc: Cannot find function here
afb+: Cannot add basic block at 0x000010e0
afb+: Cannot add basic block at 0x000010ed
afb+: Cannot add basic block at 0x000010fb
afb+: Cannot add basic block at 0x00001108
afb+: Cannot add basic block at 0x00001120
afB: Cannot find function to set bits at 0x000010e0
afc: Cannot find function here
afb+: Cannot add basic block at 0x00001130
afB: Cannot find function to set bits at 0x00001130
afc: Cannot find function here
afb+: Cannot add basic block at 0x00001000
afb+: Cannot add basic block at 0x00001014
afb+: Cannot add basic block at 0x00001016
afB: Cannot find function to set bits at 0x00001000
afc: Cannot find function here
afb+: Cannot add basic block at 0x000011c0
afB: Cannot find function to set bits at 0x000011c0
afc: Cannot find function here
afb+: Cannot add basic block at 0x000011c8
afB: Cannot find function to set bits at 0x000011c8
afc: Cannot find function here
afb+: Cannot add basic block at 0x00001150
afb+: Cannot add basic block at 0x00001187
afb+: Cannot add basic block at 0x00001190
afb+: Cannot add basic block at 0x000011a6
afB: Cannot find function to set bits at 0x00001150
afc: Cannot find function here
Missing function at 0x00001150
Missing function at 0x00001150
Missing function at 0x00001150
afb+: Cannot add basic block at 0x00001139
afB: Cannot find function to set bits at 0x00001139
[...]

PR #17270, opened 2020-07-13
Fix non-debug projects by reopening the file ##projects

RHL120:
- [*] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [*] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [*] I've added tests that prove my fix is effective or that my feature works (if possible)
- [*] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**
normally when you save a project it and reaload it r2 doesn't reopen the file so I added a o [binary file path] into the r2 project's rc file. the projects loads successfully if not saved in debug mode. because of PIE 
...

**Test plan**
I have recompiled r2 on archlinux with /sys/install.sh and everything works prefectly fine including the Ps (project save) command
...
...
# #16353

### DIFF
--- a/libr/core/project.c
+++ b/libr/core/project.c
@@ -630,7 +630,10 @@ static bool simpleProjectSaveScript(RCore *core, const char *file, int opts) {
 
 static bool projectSaveScript(RCore *core, const char *file, int opts) {
 	char *filename, *hl, *ohl = NULL;
+	char *binpath = r_file_abspath(core->bin->file);
+	char  *reopen = (char *) r_malloc(strlen(binpath) + 32);
 	int fd, fdold;
+	sprintf(reopen, "\"o %s\"\n", binpath);
 
 	if (!file || *file == '\0') {
 		return false;
@@ -653,6 +656,9 @@ static bool projectSaveScript(RCore *core, const char *file, int opts) {
 	r_cons_singleton ()->fdout = fd;
 	r_cons_singleton ()->context->is_interactive = false;
 	r_str_write (fd, "# r2 rdb project file\n");
+        r_str_write(fd, reopen);
+	free(reopen);
+	free(binpath);
 	// Set file.path and file.lastpath to empty string to signal
 	// new behaviour to project load routine (see io maps below).
 	r_config_set (core->config, "file.path", "");
@@ -779,6 +785,7 @@ R_API bool r_core_project_save(RCore *core, const char *prjName) {
 #if TRANSITION
 			r_file_rm (scriptPath);
 			r_sys_mkdirp (prjDir);
+
 			eprintf ("Please remove: rm -rf %s %s.d\n", prjName, prjName);
 			char *rc = r_str_newf ("%s" R_SYS_DIR "rc", prjDir);
 			if (!rc) {
--- a/libr/core/project.c
+++ b/libr/core/project.c
@@ -630,7 +630,7 @@ static bool simpleProjectSaveScript(RCore *core, const char *file, int opts) {
 
 static bool projectSaveScript(RCore *core, const char *file, int opts) {
 	char *filename, *hl, *ohl = NULL;
-	char  *reopen = r_str_newf ("\"o %s\"\n", r_file_abspath(core->bin->file);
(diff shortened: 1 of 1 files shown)


libr/core/project.c
```diff
@@ -630,7 +630,10 @@ static bool simpleProjectSaveScript(RCore *core, const char *file, int opts) {
 
 static bool projectSaveScript(RCore *core, const char *file, int opts) {
 	char *filename, *hl, *ohl = NULL;
+	char *binpath = r_file_abspath(core->bin->file);
+	char  *reopen = (char *) r_malloc(strlen(binpath) + 33);
```
trufae, line 634: remove space in `char  *`

libr/core/project.c
```diff
@@ -630,7 +630,10 @@ static bool simpleProjectSaveScript(RCore *core, const char *file, int opts) {
...
 	int fd, fdold;
+	sprintf(reopen, "\"o %s\"\n", binpath);
```
trufae, line 636: just use r_str_newf() and avoid r_malloc+sprintf

libr/core/project.c
```diff
@@ -653,6 +656,9 @@ static bool projectSaveScript(RCore *core, const char *file, int opts) {
 	r_cons_singleton ()->fdout = fd;
 	r_cons_singleton ()->context->is_interactive = false;
 	r_str_write (fd, "# r2 rdb project file\n");
+	r_str_write(fd, reopen);
+	free(reopen);
```
trufae, line 660: see CODING_STYLE. add a space before every ( in calls

trufae:
I think this will be breaking debug projects. there's no if (!cfg.debug) {..

libr/core/project.c
trufae: also, can you describe which is the exact issue and what do this solves? the idea of projects is that they are tied to one specific file which must be copied into the projects directory imho.

the problem is that bininfo will not be loaded otherwise.

maybe we can have an o command that loads the file the bininfo, but doesnt sets any flag from them. so the projects may be able to load and map all the files and then just reload the flags from the previous session and keep consistency

RHL120:
> I think this will be breaking debug projects. there's no if (!cfg.debug) {..

The thing is that projects are already broken (at least in my experience)

RHL120:
> Can we have a test for this? there are few tests for projects

What exactly do you want me to do?

trufae:
See test/db/cmd/cmd_project and add another test to verify your patch fixes an issue, this way we can ensure the behaviour is kept over time and nothing breaks

trufae:
Also, this PR breaks a test. didnt chk if its for better or worst

Screenshot 2020-07-15 at 11 19 53

RHL120:
> Also, this PR breaks a test. didnt chk if its for better or worst
> 
> Screenshot 2020-07-15 at 11 19 53

I ran the test with make and you are right it fails and I also ran it manually and got what is in the screenshot
(left is r2 but without saving and reopenning the project)
(top right is r2 with the patch and the regular test
r2-out
)
(bottom right is regular r2 with the regular test)

RHL120:
> See test/db/cmd/cmd_project and add another test to verify your patch fixes an issue, this way we can ensure the behaviour is kept over time and nothing breaks

and then I should push it to the main repo ?

trufae:
What do you mean?

RHL120:
> What do you mean?

Do you want me to add a test and then push it to the repository ?

trufae:
push the test into this same PR, actually there's a broken test in cmd_project in your PR, can you confirm if its a regression and fix it?

RHL120:
> push the test into this same PR, actually there's a broken test in cmd_project in your PR, can you confirm if its a regression and fix it?

actuelly there is a bit of a problem, I can't push to this PR would you like me to close (and possibly open another one)

RHL120:
> push the test into this same PR, actually there's a broken test in cmd_project in your PR, can you confirm if its a regression and fix it?

The pull request is pointing to an unkown repository

PR #17271, opened 2020-07-13
Fix #17162 - Implement next/prev fd helpers for ##io

condret:

**Detailed description**
Let me be honest with you: From my perspective this code is mostly useless, except for the use case of #17162 . RIDStorage does not provide any specific ordering of the stored values, so if you look at a specific value, that is stored in an id_storage, there is no such thing as next or previous value. However this code pr is useful for cycling through values in an unspecific order, just as in #17156.

So why is there r_io_fd_get_next and r_io_desc_get_next? Wouldn't it be enough to have only one of them?
The reason for that is, that I want the io api to be used as painless as possible. Imo that means, that projects that only use the map and fd api shouldn't be forced to switch between desc and fd to use this.

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -393,6 +393,8 @@ R_API void r_io_desc_free (RIODesc *desc);
 R_API bool r_io_desc_add (RIO *io, RIODesc *desc);
 R_API bool r_io_desc_del (RIO *io, int fd);
 R_API RIODesc *r_io_desc_get (RIO *io, int fd);
+R_API RIODesc *r_io_desc_get_next (RIO *io, RIODesc *desc);
+R_API RIODesc *r_io_desc_get_prev (RIO *io, RIODesc *desc);
 R_API ut64 r_io_desc_seek (RIODesc *desc, ut64 offset, int whence);
 R_API bool r_io_desc_resize (RIODesc *desc, ut64 newsize);
 R_API ut64 r_io_desc_size (RIODesc *desc);
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -460,6 +460,8 @@ R_API int r_io_fd_get_current(RIO *io);
 R_API bool r_io_use_fd (RIO *io, int fd);
 R_API int r_io_fd_get_next (RIO *io, int fd);
 R_API int r_io_fd_get_prev (RIO *io, int fd);
+R_API int r_io_fd_get_highest (RIO *io);
+R_API int r_io_fd_get_lowest (RIO *io);
 
 
 #define r_io_range_new()	R_NEW0(RIORange)
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -395,6 +395,8 @@ R_API bool r_io_desc_del (RIO *io, int fd);
 R_API RIODesc *r_io_desc_get (RIO *io, int fd);
 R_API RIODesc *r_io_desc_get_next (RIO *io, RIODesc *desc);
 R_API RIODesc *r_io_desc_get_prev (RIO *io, RIODesc *desc);
+R_API RIODesc *r_io_desc_get_highest (RIO *io);
+R_API RIODesc *r_io_desc_get_lowest (RIO *io);
 R_API ut64 r_io_desc_seek (RIODesc *desc, ut64 offset, int whence);
 R_API bool r_io_desc_resize (RIODesc *desc, ut64 newsize);
 R_API ut64 r_io_desc_size (RIODesc *desc);
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -456,6 +456,8 @@ R_API bool r_io_fd_get_base (RIO *io, int fd, ut64 *base);
 R_API const char *r_io_fd_get_name (RIO *io, int fd);
 R_API int r_io_fd_get_current(RIO *io);
 R_API bool r_io_use_fd (RIO *io, int fd);
+R_API int r_io_fd_get_next (RIO *io, int fd);
+R_API int r_io_fd_get_prev (RIO *io, int fd);
 
 
 #define r_io_range_new()	R_NEW0(RIORange)
--- a/libr/io/io_desc.c
+++ b/libr/io/io_desc.c
@@ -73,6 +73,24 @@ R_API RIODesc* r_io_desc_get(RIO* io, int fd) {
 	return (RIODesc*) r_id_storage_get (io->files, fd);
 }
 
+R_API RIODesc *r_io_desc_get_next(RIO *io, RIODesc *desc) {
+	r_return_val_if_fail (desc && io && io->files, NULL);
+	const int next_fd = r_io_fd_get_next (io, desc->fd);
+	if (next_fd == -1) {
+		return NULL;
+	}
+	return (RIODesc*) r_id_storage_get (io->files, next_fd);
+}
+
+R_API RIODesc *r_io_desc_get_prev(RIO *io, RIODesc *desc) {
+	r_return_val_if_fail (desc && io && io->files, NULL);
+	const int prev_fd = r_io_fd_get_prev (io, desc->fd);
+	if (prev_fd == -1) {
+		return NULL;
+	}
+	return (RIODesc*) r_id_storage_get (io->files, prev_fd);
+}
+
 R_API RIODesc *r_io_desc_open(RIO *io, const char *uri, int perm, int mode) {
 	r_return_val_if_fail (io && uri, NULL);
 	RIOPlugin *plugin = r_io_plugin_resolve (io, uri, 0);
--- a/libr/io/io_desc.c
+++ b/libr/io/io_desc.c
@@ -91,6 +91,14 @@ R_API RIODesc *r_io_desc_get_prev(RIO *io, RIODesc *desc) {
 	return (RIODesc*) r_id_storage_get (io->files, prev_fd);
 }
 
+R_API RIODesc *r_io_desc_get_highest(RIO *io) {
+	return r_io_desc_get (io, r_io_fd_get_highest (io));
+}
+
+R_API RIODesc *r_io_desc_get_lowest(RIO *io) {
+	return r_io_desc_get (io, r_io_fd_get_lowest (io));
+}
+
 R_API RIODesc *r_io_desc_open(RIO *io, const char *uri, int perm, int mode) {
 	r_return_val_if_fail (io && uri, NULL);
 	RIOPlugin *plugin = r_io_plugin_resolve (io, uri, 0);
--- a/libr/io/io_fd.c
+++ b/libr/io/io_fd.c
@@ -144,3 +144,24 @@ R_API int r_io_fd_get_prev(RIO *io, int fd) {
 	}
 	return ret;
 }
+
+R_API int r_io_fd_get_highest(RIO *io) {
+	r_return_val_if_fail (io, -1);
+	int fd = 0;
+	if (!r_id_storage_get_prev (io->files, (ut32 *)&fd)) {
+		return -1;
+	}
+	return fd;
+}
+
+R_API int r_io_fd_get_lowest(RIO *io) {
+	r_return_val_if_fail (io, -1);
+	int fd = 0;
+	if (!r_id_storage_get_prev (io->files, (ut32 *)&fd)) {
+		return -1;
+	}
+	if (!r_id_storage_get_next (io->files, (ut32 *)&fd)) {
+		return -1;
+	}
+	return fd;
+}
--- a/libr/io/io_fd.c
+++ b/libr/io/io_fd.c
@@ -126,3 +126,21 @@ R_API int r_io_fd_get_current(RIO *io) {
 	}
 	return -1;
 }
+
+R_API int r_io_fd_get_next(RIO *io, int fd) {
+	r_return_val_if_fail (io, -1);
+	int ret = fd;
+	if (!r_id_storage_get_next (io->files, (ut32 *)&ret)) {
+		return -1;
+	}
+	return ret;
+}
+
+R_API int r_io_fd_get_prev(RIO *io, int fd) {
+	r_return_val_if_fail (io, -1);
+	int ret = fd;
+	if (!r_id_storage_get_prev (io->files, (ut32 *)&ret)) {
+		return -1;
+	}
+	return ret;
+}
--- a/libr/util/idpool.c
+++ b/libr/util/idpool.c
@@ -163,6 +163,48 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
 	return storage->data[id];
 }
 
+// usually these shall not be used
+// id-storage doesn't provide sorting things in specific order
+// there are only a very few reasonable usecases, such as cycling through RIODescs
+R_API bool r_id_storage_get_next(RIDStorage *storage, ut32 *id) {
+	r_return_val_if_fail (id && storage && storage->data, false);
+	// it does not indicate a bug, if the passed id goes beyond the boundaries of the storage
+	// therefor no r_return_val_if_fail for this case
+	if (storage->size <= *id) {
+		return false;
+	}
+
+	ut32 _id = *id;
+	const ut32 start_id = _id;
+	do {
+		_id = (_id + 1) % storage->size;
+		if (_id == start_id) {
+			// in case when there is no other element in the storage
+			return false;
+		}
+	} while (!storage->data[_id]);
+	*id = _id;
+	return true;
+}
+
+R_API bool r_id_storage_get_prev(RIDStorage *storage, ut32 *id) {
+	r_return_val_if_fail (id && storage && storage->data, false);
+	if (storage->size <= *id) {
+		return false;
+	}
+
+	ut32 _id = *id;
+	const ut32 start_id = _id;
+	do {
+		_id = (_id + storage->size - 1) % storage->size;
+		if (_id == start_id) {
+			return false;
+		}
+	} while (!storage->data[_id]);
+	*id = _id;
+	return true;
+}
+
 R_API void r_id_storage_delete(RIDStorage* storage, ut32 id) {
 	if (!storage || !storage->data || (storage->size <= id)) {
 		return;
--- a/test/unit/test_id_storage.c
+++ b/test/unit/test_id_storage.c
@@ -1,0 +1,62 @@
+#include <r_util.h>
+#include "minunit.h"
+
+
+bool test_r_id_storage_prev_next_eq_0(void) {
+	// test if next reverts prev
+	ut32 id, _id;
+	RIDStorage *ids = r_id_storage_new (0, 15);
+	r_id_storage_add (ids, "aaa", &_id);
+	r_id_storage_add (ids, "bbb", &id);
+	r_id_storage_add (ids, "ccc", &_id);
+	_id = id;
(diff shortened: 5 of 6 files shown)


trufae:
Will be good to have an fd iterator under @@@ to walk them all using this api this way you can add a test with commands instead of unit and use this api inside r2 for something rather than a future PR

ret2libc:
> Will be good to have an fd iterator under @@@ to walk them all using this api this way you can add a test with commands instead of unit and use this api inside r2 for something rather than a future PR

Adding a new command just to test an API does not seem like the best thing to do.

trufae:
Agree @ret2libc having a new command will require more tests and wont be that nice as a unit.

My PR will test those APIs from the command pov, and then ill create another PR adding the @@@ fd iterator. we still have a discussion pending on all the @@@ commands

trufae:
Also @ret2libc i want to know your opinion on the function signature model returning bool or the fd, using ut32_MAX or not, etc..

ret2libc:
> Also @ret2libc i want to know your opinion on the function signature model returning bool or the fd, using ut32_MAX or not, etc..

I think it depends on the kind of API. When UT32_MAX (or other values) are not possible valid values, it may make sense having UT32_MAX used as error. When, however, UT32_MAX is a valid value it would be best to separate those things (e.g. like the example of the uleb/sleb we discussed few days ago).

libr/include/r_io.h
```diff
@@ -395,6 +395,8 @@ R_API void r_io_desc_free (RIODesc *desc);
 R_API bool r_io_desc_add (RIO *io, RIODesc *desc);
 R_API bool r_io_desc_del (RIO *io, int fd);
 R_API RIODesc *r_io_desc_get (RIO *io, int fd);
+R_API RIODesc *r_io_desc_get_next (RIO *io, RIODesc *desc);
```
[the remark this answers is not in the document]
trufae: We can address that in a separate PR, because for some reason the whole r_io.h is liekt his

libr/include/r_io.h
```diff
@@ -458,6 +460,8 @@ R_API bool r_io_fd_get_base (RIO *io, int fd, ut64 *base);
 R_API const char *r_io_fd_get_name (RIO *io, int fd);
 R_API int r_io_fd_get_current(RIO *io);
 R_API bool r_io_use_fd (RIO *io, int fd);
+R_API int r_io_fd_get_next (RIO *io, int fd);
```
trufae, line 463: fds are int or ut32? :D

libr/util/idpool.c
```diff
@@ -163,6 +163,35 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+R_API bool r_id_storage_get_next(RIDStorage *storage, ut32 *id) {
+	r_return_val_if_fail (id && storage && storage->data, false);
+	// it does not indicate a bug, if the passed id goes beyond the boundaries of the storage
+	// therefor no r_return_val_if_fail for this case
+	if (storage->size <= *id) {
+		return false;
+	}
+
+	do {
+		*id = (*id + 1) % storage->size;
```
trufae, line 178: hold *id in a local var until the function returns, then set its value

libr/io/io_fd.c
```diff
@@ -126,3 +126,21 @@ R_API int r_io_fd_get_current(RIO *io) {
...
+	if (!r_id_storage_get_next (io->files, (ut32 *)&ret)) {
+		return -1;
+	}
+	return ret;
+}
+
+R_API int r_io_fd_get_prev(RIO *io, int fd) {
+	r_return_val_if_fail (io, -1);
+	int ret = fd;
+	if (!r_id_storage_get_prev (io->files, (ut32 *)&ret)) {
```
trufae, line 142: fds in idstorage are ut32, but in io are int?

trufae:
ping. my pr is waiting for this

libr/include/r_io.h
condret: int, they are int, and ids are ut32. There is no point in having int ids, but it's useful to have int fds

libr/io/io_fd.c
condret: yes, it's intended that way.

libr/util/idpool.c
condret: why

condret: I changed that, even thought I don't see any point in this

libr/util/idpool.c
```diff
@@ -163,6 +163,39 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+	// it does not indicate a bug, if the passed id goes beyond the boundaries of the storage
+	// therefor no r_return_val_if_fail for this case
+	if (storage->size <= *id) {
+		return false;
+	}
+
+	ut32 _id = *id;
+	do {
+		_id = (_id + 1) % storage->size;
+	} while (!storage->data[_id]);
```
ret2libc, line 180: I haven't checked whether that's true or not, but this code assumes there will always be a NULL element in `storage->data`. Is that the case? Because if not, there could be an infinite loop.

libr/util/idpool.c
```diff
@@ -163,6 +163,39 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
 	return storage->data[id];
 }
 
+// usually these shall not be used
+// id-storage doesn't provide sorting things in specific order
```
ret2libc, line 167: Is it really true? I think there is indeed sorting provided by id, afaik, as things are stored by id.

libr/util/idpool.c
condret: The code assumes, that there is at least 1 non-NULL element in the storage... but you're right, this can escalate to an endless loop (think about "r2 --") . Nice catch. I fixed that now

libr/util/idpool.c
condret: idstorage was created during siol as a replacement for sdb. I felt like using sdb to store int-values from idpool was just unnecessary perf consuming (sprintf), for no reason. idstorage is a thing that you pass a pointer that you want to store and it, and it grabs an id from idpool and uses that as a key, which it passes back to the caller. Cool thing about it, is that it always will use the same "pattern" of keys/ids, whenever you run r2 and do the same commands. This makes creating r2-scripts much easier, as you can just hardcode some values, bc you know, it will always with these values.
Ofc, you could do sort operations on idstorage, but I felt like this could weaken the scripting reliability of io. Therefor there is no "r_id_storage_sort". if one wants to sort things in an idstorage, i suggest using oidstorage, which is basically an idstorage with a permutation array, on which one can do sorting operations.

I hope this explains, why I don't see these ids as a sorting

ret2libc: it's a bit fuzzy but... ok :D

ret2libc:
@condret could you please merge master into your pr or rebase on top of master? Everything should be green in theory.

ret2libc:
>     * [ ]  get first fd
> 
>     * [ ]  get last fd

Why do you need these @trufae ? You always start somewhere AFAIK.

trufae:
yes i need them because there's a command to rotate, so when the last fd is reached it jumps back to the begining.

ret2libc:
> yes i need them because there's a command to rotate, so when the last fd is reached it jumps back to the begining.

This is already done by the API @condret implemented AFAIU. They go back to the beginning automatically.

trufae:
Uhm so its returning false, but returning the first? or how that goes?

ret2libc:
> Uhm so its returning false, but returning the first? or how that goes?

No, AFAIU if there are at least 2 elements it will be able to wrap around and return the next/prev even when you are at the last/first item. It returns false only when there is just one element in the storage. Maybe this is not the best behaviour though.

@condret could you add unit tests for these 2 API please?

This is what I expect, please guys let me know if you agree:
```
id_storage = [2, 4, 5, 6]
r_id_storage_get_next (id_storage, 2) == 4, true
r_id_storage_get_next (id_storage, 4) == 5, true
r_id_storage_get_next (id_storage, 5) == 6, true
r_id_storage_get_next (id_storage, 6) == 2, true
r_id_storage_get_prev (id_storage, 2) == 6, true
r_id_storage_get_prev (id_storage, 6) == 5, true
r_id_storage_get_prev (id_storage, 5) == 4, true
r_id_storage_get_prev (id_storage, 4) == 2, true

id_storage = [2]
r_id_storage_get_next (id_storage, 2) == 2, true
r_id_storage_get_prev (id_storage, 2) == 2, true

id_storage = []
r_id_storage_get_next (id_storage, 1) == false
r_id_storage_get_prev (id_storage, 1) == false
```

condret:
no, I add unit test now, I thought this was already merged. Don't assume any particular order in idstorage. idstorage guarantees that if you do the same thing in 2 idstorages, that were initialized the same way, you get the same results. id-pool currently produces ids on ascending order, but that might change in the future, so don't implement constraints in tests here.

condret:
ok, maybe one could do that, if one uses r_id_storage_set instead of add...

test/unit/test_id_storage.c
```diff
@@ -0,0 +1,62 @@
+#include <r_util.h>
+#include "minunit.h"
+
+
+bool test_r_id_storage_prev_next_eq_0(void) {
```
ret2libc, line 5: these aren't really testing much apart from the fact that next(prev(x)) == x ... if you don't want to assume any sorting at all on the IDs (and I don't understand the reasons) then this is the wrong structure for this. I don't understand why we can't assume a sorting, given that id_storage have numerical IDs and you *can* sort numbers. sure, id_pool could give numbers in different orders, but that does not matter much, because id_storage maintains an array indexed by the id, so in any case we have a sorting on the ids.

condret: the only assumption here, is that prev inverts next. this doesn't mean the same as ids are given out in an ascending/descending or any other specific order. just because the initial pattern provided by r_id_pool is ascending atm, doesn't mean this will stay like that forever.

XVilka:
@condret could you please rebase on top of the current master?

test/unit/test_id_storage.c
ret2libc: As already explained over telegram, it is essential that if fds are listed in a given order to users then next/prev respect that order, whatever that is.

In other words, if I do `o` and I see 5, 3, 7, 4 and 5 is the current fd, then executing next one time should return 3. Executing next another time should return 7. Then 4 and then 5 again. This is a very important property of this API, otherwise it would be useless.

libr/util/idpool.c
```diff
@@ -163,6 +163,48 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
 	return storage->data[id];
 }
 
+// usually these shall not be used
```
ret2libc, line 166: Also, providing an API and saying from its very first day "hey you should not use them" seems wrong. Moreover, if you have a next/prev API that means there is indeed some sort of order. Saying there isn't one is just wrong and not in line with this API. I would just remove this comment.

trufae:
The primitives por my pr are:

* go next/prev fd (and stop if its the first or the last)
* go next/prev (but rotate when reaching the limit)

i dont really need to know which is the first or the last.. can I assume numeric order and check if X < Y ?
the tests should reflect this usage

ret2libc:
Where are we at with this? @trufae I think it would be much easier if you just provide a small unit test that does mostly what you'd like to have. It shoud require 1 minute. At least @condret can know what exactly he needs to implement and he can help properly. 

In an effort to try to help with this, this could be a test:
```c
static bool test_prevnext(void) {
	RIDStorage *ids = r_id_storage_new (5, 10);
	ut32 id[3];
	r_id_storage_add (ids, "a", &id[0]);
	r_id_storage_add (ids, "b", &id[1]);
	r_id_storage_add (ids, "c", &id[2]);

	RList *ids_list = r_id_storage_list (ids);
	const char *val0list = r_list_get_n (ids_list, 0);
	ut32 id0list = id[*val0list - 'a'];
	const char *val1list = r_list_get_n (ids_list, 1);
	ut32 id1list = id[*val1list - 'a'];
	const char *val2list = r_list_get_n (ids_list, 2);
	ut32 id2list = id[*val2list - 'a'];

	ut32 idw;
	idw = r_id_storage_get_next (ids, id0list);
	mu_assert_eq (idw, id1list, "next of id0 should be id1");
	idw = r_id_storage_get_next (ids, id1list);
	mu_assert_eq (idw, id2list, "next of id1 should be id2");
	idw = r_id_storage_get_next (ids, id2list);
	mu_assert_eq (idw, -1, "next of id2 (last) should not exist (-1)");

	idw = r_id_storage_get_prev (ids, id2list);
	mu_assert_eq (idw, id1list, "prev of id2 should be id1");
	idw = r_id_storage_get_prev (ids, id1list);
	mu_assert_eq (idw, id0list, "prev of id1 should be id0");
	idw = r_id_storage_get_prev (ids, id0list);
	mu_assert_eq (idw, -1, "prev of id0 (first) should not exist (-1)");

	idw = r_id_storage_get_lowest (ids);
	mu_assert_eq (idw, id0list, "lowest should be id0");
	idw = r_id_storage_get_highest (ids);
	mu_assert_eq (idw, id2list, "highest should be id2");

	r_list_free (ids_list);
	r_id_storage_free (ids);

	mu_end;
}
```

By the way, with this test I'm not even assuming any particular order within id_storage, but i'm just assuming that the order returned by r_id_storage_list is the same as the one you get by traversing the ids with _next/_prev. How does that look like? I think this is everything that is needed to correctly implement `opn`, `opp`, etc.

libr/io/io_desc.c
```diff
@@ -73,6 +73,32 @@ R_API RIODesc* r_io_desc_get(RIO* io, int fd) {
...
+R_API RIODesc *r_io_desc_get_prev(RIO *io, RIODesc *desc) {
+	r_return_val_if_fail (desc && io && io->files, NULL);
+	const int prev_fd = r_io_fd_get_prev (io, desc->fd);
+	if (prev_fd == -1) {
+		return NULL;
+	}
+	return (RIODesc*) r_id_storage_get (io->files, prev_fd);
+}
+
+R_API RIODesc *r_io_desc_get_highest(RIO *io) {
```
radare, line 94: object is `io`. so the api should be r_io_get_highest_desc()

libr/io/io_desc.c
```diff
@@ -73,6 +73,32 @@ R_API RIODesc* r_io_desc_get(RIO* io, int fd) {
...
+	return r_io_desc_get (io, r_io_fd_get_highest (io));
+}
+
+R_API RIODesc *r_io_desc_get_lowest(RIO *io) {
+	return r_io_desc_get (io, r_io_fd_get_lowest (io));
```
radare, line 99: same here

libr/util/idpool.c
```diff
@@ -163,6 +163,48 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+// id-storage doesn't provide sorting things in specific order
+// there are only a very few reasonable usecases, such as cycling through RIODescs
+R_API bool r_id_storage_get_next(RIDStorage *storage, ut32 *id) {
+	r_return_val_if_fail (id && storage && storage->data, false);
+	// it does not indicate a bug, if the passed id goes beyond the boundaries of the storage
+	// therefor no r_return_val_if_fail for this case
+	if (storage->size <= *id) {
```
radare, line 173: i think its more readable if you put the subject first: `if (id > storage->size) {..`

test/unit/test_id_storage.c
```diff
@@ -0,0 +1,62 @@
...
+
+int all_tests() {
+	mu_run_test(test_r_id_storage_prev_next_eq_0);
+	mu_run_test(test_r_id_storage_prev_next_eq_1);
+	mu_run_test(test_r_id_storage_prev_next_eq_2);
+	return tests_passed != tests_run;
+}
+
+int main(int argc, char **argv) {
+	return all_tests();
```
radare, line 61: missing space (

test/unit/test_id_storage.c
```diff
@@ -0,0 +1,62 @@
...
+	r_id_storage_get_prev (ids, &_id);
+	r_id_storage_get_next (ids, &_id);
+	r_id_storage_free (ids);
+	
+	mu_assert_eq (id, _id, "r_id_storage_{next/prev} reversal 2");
+	mu_end;
+}
+
+int all_tests() {
+	mu_run_test(test_r_id_storage_prev_next_eq_0);
```
radare, line 54: missing spaces

test/unit/test_id_storage.c
radare, line 5: static

test/unit/test_id_storage.c
```diff
@@ -0,0 +1,62 @@
...
+	_id = id;
+	r_id_storage_get_prev (ids, &_id);
+	r_id_storage_get_next (ids, &_id);
+	r_id_storage_free (ids);
+	
+	mu_assert_eq (id, _id, "r_id_storage_{next/prev} reversal 0");
+	mu_end;
+}
+
+bool test_r_id_storage_prev_next_eq_1(void) {
```
radare, line 21: static

test/unit/test_id_storage.c
```diff
@@ -0,0 +1,62 @@
...
+	_id = id;
+	r_id_storage_get_prev (ids, &_id);
+	r_id_storage_get_next (ids, &_id);
+	r_id_storage_free (ids);
+	
+	mu_assert_eq (id, _id, "r_id_storage_{next/prev} reversal 1");
+	mu_end;
+}
+
+bool test_r_id_storage_prev_next_eq_2(void) {
```
radare, line 37: static

libr/include/r_io.h
```diff
@@ -393,6 +393,10 @@ R_API void r_io_desc_free (RIODesc *desc);
 R_API bool r_io_desc_add (RIO *io, RIODesc *desc);
 R_API bool r_io_desc_del (RIO *io, int fd);
 R_API RIODesc *r_io_desc_get (RIO *io, int fd);
+R_API RIODesc *r_io_desc_get_next (RIO *io, RIODesc *desc);
```
radare, line 396: functin definitions shoudlnt have a space before (

libr/io/io_fd.c
```diff
@@ -126,3 +126,42 @@ R_API int r_io_fd_get_current(RIO *io) {
 	}
 	return -1;
 }
+
+R_API int r_io_fd_get_next(RIO *io, int fd) {
```
radare, line 130: r_io_get_next_fd maybe?

radare (changes requested):

libr/util/idpool.c
trufae: ignore, you are just following the same style as in other places in the same file. ill rather go for doing this in a separate pr

ret2libc:
Closing in favour of https://github.com/radareorg/radare2/pull/17652.

libr/include/r_io.h
condret: this issue is not part of this pr, imo there should be a dedicated pr to fix that file

libr/io/io_fd.c
condret: no, it's r_<big_api>_<small_api>_<function_name>

condret:
might reopen later, imo this implementation is almost perfect, except it needs some more tests

trufae:
Your commits are in my PR, ill carefully squash them when ready. i can give you access to my branch if you are in the mood. i just joined them to move further

PR #17275, opened 2020-07-14
Fix #17264 - use r_types.h in r2r

### LINKED ISSUE #17264: Compile failure: radare2 4.4.0 does not compile on MacOS 10.9 Mavericks
I am running MacOS 10.9. Yes, I am aware that it is no longer officially supported, but I am still running it on this machine, and I would like to run the very nice looking rasm2 on it as well.

When running `sys/install.sh` as instructed in the readme (as well as when trying to install via the last supported version of brew), compilation fails because the project uses `clock_gettime` and `CLOCK_MONOTONIC`, which are not supported in MacOS 10.11 and earlier, without using any compatibility layer such as [PosixMachTiming](https://github.com/pineapplemachine/PosixMachTiming/tree/improve-degenerate-case-behavior).

It would be fantastic if such a compatibility layer could be added so that radare2 can be compiled on MacOS 10.9.

As a workaround, though, I was able to get 3.1.3 installed and apparently running fine. I don't suppose anyone can let me know what was the last supported version on MacOS 10.9? I could not find anything in the release changelogs that indicated dropping support for older versions of MacOS.

Thanks!

XVilka:

**Detailed description**

A blind fix by adding `<r_types.h>` into the `binr/r2r/r2r.h` to define `CLOCK_MONOTONIC` if undefined for the current platform.

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/binr/r2r/r2r.h
+++ b/binr/r2r/r2r.h
@@ -3,6 +3,7 @@
 #ifndef RADARE2_R2R_H
 #define RADARE2_R2R_H
 
+#include <r_types.h>
 #include <r_util.h>
 
 #if defined (__FreeBSD__) || defined (__FreeBSD_kernel__)


trufae:
This is included from rutil. It wont fix the problem. R2r should use r_sys_now() instead of getclock

trufae:
Actuallly rutil have profiling primitives too

thestr4ng3r:
`r_sys_now()` is wall clock time, so it would be wrong. But we should have a similar function in util with monotonic time.

thestr4ng3r:
Superseded by #17276

PR #17276, opened 2020-07-14
Add r_sys_now_mono() and use in r2r ##util

### LINKED ISSUE #17264: Compile failure: radare2 4.4.0 does not compile on MacOS 10.9 Mavericks

thestr4ng3r:

**Detailed description**

Wall clock time is conceptually wrong for measuring time deltas. This adds a portable function for monotonic time.

**Test plan**

r2r should still show the elapsed time correctly on exit.

**Closing issues**

Fix #17264

### DIFF
--- a/binr/r2r/run.c
+++ b/binr/r2r/run.c
@@ -2,12 +2,6 @@
 
 #include "r2r.h"
 
-#define NSEC_PER_SEC  1000000000
-#define NSEC_PER_MSEC 1000000
-#define USEC_PER_SEC  1000000
-#define NSEC_PER_USEC 1000
-#define USEC_PER_MSEC 1000
-
 #if __WINDOWS__
 struct r2r_subprocess_t {
 	HANDLE stdin_write;
@@ -236,20 +230,6 @@ error:
 	goto beach;
 }
 
-static ut64 now_us(void) {
-	LARGE_INTEGER f;
-	if (!QueryPerformanceFrequency (&f)) {
-		return 0;
-	}
-	LARGE_INTEGER v;
-	if (!QueryPerformanceCounter (&v)) {
-		return 0;
-	}
-	v.QuadPart *= 1000000;
-	v.QuadPart /= f.QuadPart;
-	return v.QuadPart;
-}
-
 R_API bool r2r_subprocess_wait(R2RSubprocess *proc, ut64 timeout_ms) {
 	OVERLAPPED stdout_overlapped = { 0 };
 	stdout_overlapped.hEvent = CreateEvent (NULL, TRUE, FALSE, NULL);
@@ -265,7 +245,7 @@ R_API bool r2r_subprocess_wait(R2RSubprocess *proc, ut64 timeout_ms) {
 
 	ut64 timeout_us_abs = UT64_MAX;
 	if (timeout_ms != UT64_MAX) {
-		timeout_us_abs = now_us () + timeout_ms * USEC_PER_MSEC;
+		timeout_us_abs = r_sys_now_mono () + timeout_ms * USEC_PER_MSEC;
 	}
 
 	ut8 stdout_buf[0x500];
@@ -307,7 +287,7 @@ R_API bool r2r_subprocess_wait(R2RSubprocess *proc, ut64 timeout_ms) {
 		
 		DWORD timeout = INFINITE;
 		if (timeout_us_abs != UT64_MAX) {
-			ut64 now = now_us ();
+			ut64 now = r_sys_now_mono ();
 			if (now >= timeout_us_abs) {
 				return false;
 			}
@@ -635,12 +615,9 @@ error:
 }
 
 R_API bool r2r_subprocess_wait(R2RSubprocess *proc, ut64 timeout_ms) {
-	struct timespec timeout_abs;
+	ut64 timeout_abs;
 	if (timeout_ms != UT64_MAX) {
-		clock_gettime (CLOCK_MONOTONIC, &timeout_abs);
-		timeout_abs.tv_nsec += timeout_ms * NSEC_PER_MSEC;
-		timeout_abs.tv_sec += timeout_abs.tv_nsec / NSEC_PER_SEC;
-		timeout_abs.tv_nsec = timeout_abs.tv_nsec % NSEC_PER_SEC;
+		timeout_abs = r_sys_now_mono () + timeout_ms * R_USEC_PER_MSEC;
 	}
 
 	int r = 0;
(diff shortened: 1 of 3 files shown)


libr/util/sys.c
```diff
@@ -218,6 +218,27 @@ R_API ut64 r_sys_now(void) {
...
+	LARGE_INTEGER v;
+	if (!QueryPerformanceCounter (&v)) {
+		return 0;
+	}
+	v.QuadPart *= 1000000;
+	v.QuadPart /= f.QuadPart;
+	return v.QuadPart;
+#else
+	struct timespec now;
+	clock_gettime (CLOCK_MONOTONIC, &now);
```
trufae, line 236: I still see issues here:

* on modern unix, timespec/timeval is 64bit for seconds, this means we cant fit seconds+usecs in 64bit numbers
* there's an overflow in this multiplication
* we should use the timeradd/timersub/timercmp posix macros
* the CLOCK_MONOTONIC is not everywhere, it must be checked at configure stage or with macros carefully chking each OS version

```
#if __APPLE__ && defined(MAC_OS_X_VERSION_10_12)
```

mac manpage says that:

```
HISTORY
     These functions first appeared in Mac OSX 10.12

STANDARDS
     The clock_gettime(), clock_settime(), and clock_getres() system calls conform to IEEE Std
     1003.1b-1993 (``POSIX.1'').  cleck_gettime_nsec_np() is a non-portable Darwin extension.  The
     clock IDs CLOCK_MONOTONIC_RAW and CLOCK_UPTIME_RAW are extensions to the POSIX interface.
```

this post shows an alternative for pre-monotonic systems, but it depends on globals, that imho can be defined as static to avoid scope polution.

* https://stackoverflow.com/questions/5167269/clock-gettime-alternative-in-mac-os-x

Here's a better alternative:

* https://gist.github.com/jbenet/1087739

libr/util/sys.c
```diff
@@ -218,6 +218,27 @@ R_API ut64 r_sys_now(void) {
...
+	return now.tv_sec * R_USEC_PER_SEC
```
trufae, line 237: this multiplication can overflow. tv_sec is 64bit wide

binr/r2r/run.c
```diff
@@ -674,15 +651,12 @@ R_API bool r2r_subprocess_wait(R2RSubprocess *proc, ut64 timeout_ms) {
 		struct timeval timeout_s;
 		struct timeval *timeout = NULL;
 		if (timeout_ms != UT64_MAX) {
-			struct timespec now;
-			clock_gettime (CLOCK_MONOTONIC, &now);
-			st64 usec_diff = ((st64)timeout_abs.tv_sec - now.tv_sec) * USEC_PER_SEC
-					+ ((st64)timeout_abs.tv_nsec - now.tv_nsec) / NSEC_PER_USEC;
+			st64 usec_diff = timeout_abs - r_sys_now_mono ();
```
trufae, line 654: check if now_mono is > than timeout_abs instead of making it signed

thestr4ng3r:
All the internal overflows could be fixed, but the thing to clarify is what return type to use for the util function itself. `ut64` in microseconds could hold up to about 584942 years. Alternatively we would have to use something like `struct timespec`, but would have to define it ourselves because it doesn't exist on windows.

trufae:
yeah the main problem here is portability, i assume ut64+ut64 inline struct definition could work for windows, we can use a similar appraoch as RDebugRegState to have the same name pointing to timeval on unix and use our custom struct on windows. we can redefine those posix timeradd/sub/cmp macros as APIs taking this type as argument to process the differences

trufae:
Also, windows is unhappy about that

Screenshot 2020-07-14 at 19 25 08

trufae:
Btw i think you can solve the r2r by just filling the timeval struct with the expected timeout, no need to be that picky by adding the diff of the monotonic computation, the difference will be small imho

trufae:
ping

thestr4ng3r:
Didn't forget about it, will finish when I have the time.

kazarmy:
> `ut64` in microseconds could hold up to about 584942 years

I would prefer a single `ut64` value for simplicity and I would question as to whether r2r even needs microsecond precision, but a (normalized) 2-level timeval structure doesn't seem particularly difficult to implement and could be useful in other situations

kazarmy:
Could be tedious to implement though :weary:

libr/util/sys.c
pineapplemachine: Note also:

https://github.com/pineapplemachine/PosixMachTiming/tree/improve-degenerate-case-behavior

trufae:
As an inspiration, this is how it's done in V:

Screenshot 2020-07-22 at 16 15 33
Screenshot 2020-07-22 at 16 16 21

so it's not monotonic

thestr4ng3r:
Typedefing timespec and providing an alternative where it's not available could be a bit tricky in some cases since the members have different layout and sizes depending on the arch, so you get different types depending on what platform you are on:
20200729_13h47m37s_grim
@trufae What do you think about just using ut64 for microseconds? 584942 years sounds reasonable to me and `r_sys_now()` already does the same thing.

thestr4ng3r:
> Btw i think you can solve the r2r by just filling the timeval struct with the expected timeout, no need to be that picky by adding the diff of the monotonic computation, the difference will be small imho

No, it must do the diff. `select()` here can return any time, including right before the timeout expires. If we would just keep the contents of the timeval struct, we would extend the timeout further and further and possibly infinitely.

thestr4ng3r:
> Also i think this should be moved to libr/util/time.c and use r_time as prefix (note we have libr/util/date.c too)

Should we do the move right now? I think it would better fit in a new pr since the old `r_sys_now()` would have to be moved too and all callsites changed.

trufae:
We dont need that much precission. Its just for few seconds or maybe minute time diffing

thestr4ng3r:
I do think we should have at least microsecond precision available since milliseconds are quite long. And since `r_sys_now()` does the same thing. What I would propose is the following:
* Merge this as-is, ignoring the overflow after 584942 years
* Make another pr moving the API into `r_time.h`
* Make another pr splitting `r_sys_now(_mono)?` into `r_sys_now(_mono)?_ms` and `r_sys_now(_mono)?` where the latter will give you a full timespec-like struct.

XVilka:
Cherry-picked in `release-4.5.0` branch for the future [4.5.1 release](https://github.com/radareorg/radare2/milestone/57).

PR #17281, opened 2020-07-14
Annotations for Constant Variables and Global Variables for the decompiler

NirmalManoj:

**Detailed description**
This PR implements annotations for constant and global variables. As of now, the only information that is stored in the annotation is the address of the constant/global variable. The actions in Cutter that we will be having using these annotations is to add/rename/delete flags at the address of constant/global variables. See the corresponding PR[ #124](https://github.com/radareorg/r2ghidra-dec/pull/124) in r2ghidra-dec

...

**Test plan**

- [ ] 1. Compile PR[ #124](https://github.com/radareorg/r2ghidra-dec/pull/124) in r2ghidra-dec after compiling this PR.
- [ ] 2. Make sure JSON output is working as expected.
- [ ] 3. Look at code.
- [ ] 4. Think if we need more information in annotations for having the action described above in Cutter.

...

**Closing issues**

...

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,7 +7,7 @@ skip_tags: true
 # Branches to clone
 branches:
   only:
-    - master
+    - decompiler-refactoring
 
 # Environment variables
 environment:
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,6 +7,7 @@ skip_tags: true
 # Branches to clone
 branches:
   only:
+    - master
     - decompiler-refactoring
 
 # Environment variables
--- a/libr/include/r_util/r_annotated_code.h
+++ b/libr/include/r_util/r_annotated_code.h
@@ -35,7 +35,9 @@ typedef enum r_syntax_highlight_type_t {
 typedef enum r_code_annotation_type_t {
 	R_CODE_ANNOTATION_TYPE_OFFSET,
 	R_CODE_ANNOTATION_TYPE_SYNTAX_HIGHLIGHT,
-	R_CODE_ANNOTATION_TYPE_FUNCTION_NAME
+	R_CODE_ANNOTATION_TYPE_FUNCTION_NAME,
+	R_CODE_ANNOTATION_TYPE_GLOBAL_VARIABLE,
+	R_CODE_ANNOTATION_TYPE_CONSTANT_VARIABLE
 	// ...
 } RCodeAnnotationType;
 
@@ -56,6 +58,14 @@ typedef struct r_code_annotation_t {
 			char *name;
 			ut64 offset; // offset of the function with the name specified
 		} function_name;
+
+		struct {
+			ut64 offset;
+		} global_variable;
+
+		struct {
+			ut64 offset;
+		} constant_variable;
 	};
 } RCodeAnnotation;
 
--- a/libr/include/r_util/r_annotated_code.h
+++ b/libr/include/r_util/r_annotated_code.h
@@ -35,7 +35,6 @@ typedef enum r_syntax_highlight_type_t {
 typedef enum r_code_annotation_type_t {
 	R_CODE_ANNOTATION_TYPE_OFFSET,
 	R_CODE_ANNOTATION_TYPE_SYNTAX_HIGHLIGHT,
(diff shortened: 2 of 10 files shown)


libr/include/r_util/r_annotated_code.h
```diff
@@ -56,6 +58,14 @@ typedef struct r_code_annotation_t {
...
 			ut64 offset; // offset of the function with the name specified
 		} function_name;
+
+		struct {
+			ut64 offset;
+		} global_variable;
+
+		struct {
+			ut64 offset;
+		} constant_variable;
```
karliss, line 68: Instead of having separate enum variants for function, global variable and constant I suggest merging them into single one called reference. In many cases using it don't care if it's a reference to function, global variable or read only global variable, just the fact that it points to some fixed memory address. That way the code using it can be simplified. You can keep the separate RCodeAnnotationType values so that they can be differentiated if necessary.

NirmalManoj: @karliss I think it's a good idea to merge them. Functions have `char *name`, so when the type is not function, it will point to `nullptr`.

NirmalManoj: @karliss Please have a look now.

karliss: I don't think it's necessary to have two nested type specifiers unless the structure gets significantly more complicated, but that seems unlikely. If you decide to keep both of them don't reuse the same enum type for both. If you keep single type field there should be a comment explaining what RCodeAnnotationType use the reference field.

NirmalManoj: It is to differentiate between function annotations and variable annotations. It's useful for printing JSON also.

> If you decide to keep both of them don't reuse the same enum type for both.

I didn't understand this. I will add a comment that explains what RCodeAnnotationType values are used by `annotation.reference.type`.

karliss: You can still differentiate between the 3 if you store R_CODE_ANNOTATION_TYPE_FUNCTION_NAME, R_CODE_ANNOTATION_TYPE_GLOBAL_VARIABLE, R_CODE_ANNOTATION_TYPE_CONSTANT_VARIABLE directly in the top level type field without having a second type field inside union and R_CODE_ANNOTATION_TYPE_REFERENCE.

Currently you have two type fields one directly inside the RAnnotatedCode and one within the union reference case.  In your current design top one will always be _OFFSET, _SYNTAX_HIGHLIGHT or _REFERENCE but never FUNCTION_NAME or VARIABLE. The type inside reference struct will always be one of the last 3 but never one of the first 3. If it is used that way you should define two enums instead of using single type for two different purpose.

Out of the two approaches described above I recommend the first one.

NirmalManoj: I'll go with the first one:)

.appveyor.yml
```diff
@@ -7,7 +7,7 @@ skip_tags: true
 # Branches to clone
 branches:
   only:
-    - master
+    - decompiler-refactoring
```
karliss, line 10: You can probably add decompiler-refactoring to a list instead of replacing master. That way it will be easier to cleanup once everything is done.

NirmalManoj: Changed

libr/include/r_util/r_annotated_code.h
karliss: Is there need to keep R_CODE_ANNOTATION_TYPE_REFERENCE type instead of using the more specific ones? Also the current situation where multiple annotation types use the same enum field was for which I suggested to add a comment. Explaining the mapping.

NirmalManoj: Removed R_CODE_ANNOTATION_TYPE_REFERENCE and explained the mapping with a comment.

NirmalManoj:
Please merge.

karliss:
@NirmalManoj The PR has been accepted and tests are green. Don't you have button for doing it yourself?

karliss:
Just don't forget to cleanup commit message when squashing.

NirmalManoj:
Yes, I will merge!! It's just that I was not sure if I am supposed to do it. I have never merged a PR before.

PR #17284, opened 2020-07-15
Fixes for jump tables

GustavoLCR:

**Detailed description**

* Fix case detection for jumptables with case indirection
* Add info of switch case value
* Don't count switch cases that point to default
* Fix printing case ranges

**Test plan**

Tests pass


libr/anal/jmptbl.c
```diff
@@ -42,8 +42,105 @@ R_API bool r_anal_jmptbl(RAnal *anal, RAnalFunction *fcn, RAnalBlock *block, ut6
 	return try_walkthrough_jmptbl (anal, fcn, block, depth, jmpaddr, table, table, tablesize, tablesize, default_addr, false);
 }
 
+R_API bool try_walkthrough_casetbl(RAnal *anal, RAnalFunction *fcn, RAnalBlock *block, int depth, ut64 ip, ut64 jmptbl_loc, ut64 casetbl_loc, ut64 jmptbl_off, ut64 sz, ut64 jmptbl_size, ut64 default_case, bool ret0) {
+	bool ret = ret0, default_in_switch = false;
+	if (jmptbl_size == 0) {
+		jmptbl_size = JMPTBL_MAXSZ;
+	}
+	if (jmptbl_loc == UT64_MAX) {
+		if (anal->verbose) {
```
ret2libc, line 51: a `verbose_printf` or a `aprintf` macro would be nice in this file

test/db/anal/x86_64
```diff
@@ -3348,6 +3336,33 @@ EXPECT=<<EOF
 EOF
 RUN
 
+NAME=test_switch_indirect.exe switch/case flags
```
ret2libc, line 3339: nice!

ret2libc: only thing is that IDA is able to recognize that ebx comes from `ecx - 1` so all the cases are actually +1. But I'd say this is already a very nice improvement :)

test/db/anal/x86_32
```diff
@@ -1005,18 +1005,18 @@ aaa
 CC. @@ switch
 EOF
 EXPECT=<<EOF
-switch table (275 cases) at 0x4140b0
-switch table (121 cases) at 0x413cc0
+switch table (58 cases) at 0x4140b0
```
ret2libc, line 1008: I'm not sure this is what we want. This switch actually has 275 cases, it's just that most of them points to the default case. IDA as well shows "275 cases" and I think it may be misleading seeing less in r2.

libr/anal/jmptbl.c
```diff
@@ -42,8 +42,105 @@ R_API bool r_anal_jmptbl(RAnal *anal, RAnalFunction *fcn, RAnalBlock *block, ut6
...
+			jmpptr = r_read_le16 (jmptbl + jmpptr_idx * 2);
+			break;
+		case 4:
+			jmpptr = r_read_le32 (jmptbl + jmpptr_idx * 4);
+			break;
+		default:
+			jmpptr = r_read_le64 (jmptbl + jmpptr_idx * 8);
+			break;
+		}
+		if (jmpptr == 0 || jmpptr == UT32_MAX || jmpptr == UT64_MAX) {
```
ret2libc, line 97: ok I admit i have not read the whole PR yet, but why do you check for jmpptr == 0? Can't it be a valid value?

test/db/anal/x86_32
GustavoLCR: reverted it

libr/anal/jmptbl.c
GustavoLCR: @ret2libc see this comment:
https://github.com/radareorg/radare2/blob/27a56e848689a87e6f3b27855567148419803e88/libr/anal/jmptbl.c#L171-L174

libr/anal/fcn.c
```diff
@@ -830,7 +850,12 @@ static int fcn_recurse(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 len, int
 					// handled with try_get_jmptbl_info
 					if (try_get_jmptbl_info (anal, fcn, jmp_aop.addr, bb, &table_size, &default_case)
 						|| try_get_delta_jmptbl_info (anal, fcn, jmp_aop.addr, op.addr, &table_size, &default_case)) {
-						ret = try_walkthrough_jmptbl (anal, fcn, bb, depth, jmp_aop.addr, jmptbl_addr, op.ptr, 4, table_size, default_case, 4);
+						if (casetbl_addr != op.ptr) {
```
trufae, line 853: i would prefer == instead of !=, makes it more readable. and maybe ternary operator:?

PR #17292, opened 2020-07-16
DWARF - fixed OOB read and memory leak

HoundThe:
**Detailed description**

Fixed OOB read and leak in dwarf code base based on ASAN and valgrind findings when executing 
"radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'CL 0x0000a24e
' bins/mach0/TestRTTI-armv7-dSYM"

### DIFF
--- a/libr/bin/dwarf.c
+++ b/libr/bin/dwarf.c
@@ -481,6 +481,7 @@ static const ut8 *parse_line_header_source(RBinFile *bf, const ut8 *buf, const u
 					hdr->file_names[count].file_len = file_len;
 				}
 				free (comp_dir);
+				free (include_dir);
 				free (pinclude_dir);
 			}
 			count++;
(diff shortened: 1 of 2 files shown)


libr/bin/dwarf.c
```diff
@@ -480,8 +480,11 @@ static const ut8 *parse_line_header_source(RBinFile *bf, const ut8 *buf, const u
 					hdr->file_names[count].mod_time = mod_time;
 					hdr->file_names[count].file_len = file_len;
 				}
-				free (comp_dir);
-				free (pinclude_dir);
+				if (comp_dir) {
```
ret2libc, line 483: can't you just do `R_FREE(include_dir); R_FREE(comp_dir);` without checking for comp_dir? If they are null, it's going to be a no-op anyway.

HoundThe: I can't, because based on the code, **include_dir** and **pinclude_dir** point to the same thing in case **comp_dir** isn't NULL, that's why I check if it is NULL, if it's not, I free **include_dir** aswell, its to prevent double free in the case they both point to the same thing.

ret2libc: :+1:

PR #17293, opened 2020-07-16
Remove test/unit/legacy_unit files

### LINKED ISSUE #15575: Convert all unit legacy tests in the new minunit format
Convert all unit legacy tests in the new minuet format https://github.com/radare/radare2/issues/5314

ret2libc:
These files were created long ago, some of them don't compile anymore
and API has changed a lot since then anyway. No need to keep those
things around, just create new unit tests as needed.

 

**Detailed description**

...

**Test plan**

...

**Closing issues**

Closes #15575 , sort of. All API should be unit tested, keeping an issue opened won't make that happen and it would get stale easily. We are already trying to create unit tests for new APIs that are created.


trufae:
imho those "tests" are more like examples. im fine for removing them because i doubt they compile at all, but there are lot of .r (ragg) example programs that afaik they are nowhere else, and those should be tested properly. i didnt look carefully at all the removed c files in here. but i would prefer to wait a bit to merge this until all this stuff is reviewed. another option is to move this into extras

test/unit/legacy_unit/cons/editor.c
```diff
@@ -1,6 +0,0 @@
-#include <r_cons.h>
```
trufae, line 1: this could be `rvi` in extras :P

test/unit/legacy_unit/egg/t.sh
```diff
@@ -1,53 +0,0 @@
...
-	ragg2 t.r
-	;;
--r)
-	cat t.r
-	;;
--x)
-	ragg2 -FO t.r
-	./t
-	;;
--s)
```
trufae, line 32: all the egg scripts and programs must remain, and should be used/tested

test/unit/legacy_unit/socket/client.c
```diff
@@ -1,14 +0,0 @@
-#include <r_socket.h>
-
-int main() {
```
trufae, line 3: all those progrmas are useful examples. could be written as units, but imho its fine to move them to extras

test/unit/legacy_unit/cons/editor.c
ret2libc: `rvi`?

trufae: radare visual (editor) like vi, but with r2

XVilka:
In fact, I think we can do a separate repository for examples, rather than abusing already overweight `radare2-extras`.

trufae:
The .r programs are not just "examples".. those must be compiled and executed by the testsuite.

trufae:
imho the examples can live in r2-extras/examples perfeclty. no need to have 9000 repositories, the extras repos needs more cleanup still, but splitting up in more repos without that work is not a good idea imho

test/unit/legacy_unit/anal/test_x86im.c
```diff
@@ -1,72 +0,0 @@
-#include <stdio.h>
-#include <stdlib.h>
-#include <r_util.h>
-#include <r_types.h>
-#include "x86/x86im/x86im.h"
```
trufae, line 5: this code requires x86im which is in extras. so imho it should be moved there

test/unit/legacy_unit/config/test.c
```diff
@@ -1,19 +0,0 @@
-#include "r_config.h"
-
-int main () {
```
trufae, line 3: can you confirm we have unity tests for RCOnfig?

test/unit/legacy_unit/egg/exit.r
```diff
@@ -1,7 +0,0 @@
-main@global(128) {
```
trufae, line 1: imho this can be into libr/egg/t and installed as sample ragg programs in the share/ directory

test/unit/legacy_unit/egg/t-fun3.sh
```diff
@@ -1,24 +0,0 @@
-#!/bin/sh
```
trufae, line 1: this is actually a test for r_egg that should pass

test/unit/legacy_unit/util/array.c
```diff
@@ -1,27 +0,0 @@
-/* radare - LGPL - Copyright 2009-2011 pancake<nopcode.org> */
-#include "r_util.h"
-
-int test_flist () {
```
trufae, line 4: do we have unit for r_flist?

test/unit/legacy_unit/util/threads.c
```diff
@@ -1,53 +0,0 @@
...
-	//th = r_th_new (&looper, &ctr, 0);
-
-#if __i386__ || __x86_64__
-	asm ("int3");
-#endif
-	//r_th_start (th, true);
-	while (r_th_wait_async (th)) {
-		printf ("\nwaiting...\n");
-		fflush (stdout);
-		r_sys_usleep (400);
```
trufae, line 35: do we have unit for threads api?

ret2libc:
Closing as this is not my main prio right now. Let's leave those things there for now.. I have moved test_graph.c recently, but that's it.

PR #17294, opened 2020-07-16
Reintroduce --without-r2r configure option ##build

ret2libc:
This reverts commit 8758fa6b40b83d74abf043e2dcbbafc19846ff82.

I don't see real reasons to not provide the *option* to not compile r2r, which is not a real utility but just useful for development. The default is anyway to compile r2r, so 99.9% of users won't even notice it.


XVilka:
Broke bad builds:
```
+ gmake -k all
r2r -L
gmake: r2r: No such file or directory
gmake: *** [Makefile:12: r2r-tests] Error 127
gmake -C unit
```

PR #17308, opened 2020-07-17
Fix build without -std=c99

mscherer:
- [X] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

radare2 do not build without -std=c99 on the version of GCC I used. But, plot twist, I tried to add -std=c99, and it fail to build for others reasons. So I am fixing this code for the time being, while trying to fix it for good later.

### DIFF
--- a/libr/anal/type_dwarf.c
+++ b/libr/anal/type_dwarf.c
@@ -46,8 +46,9 @@ static void enum_type_fini(void *e, void *user) {
  * @return st32 Index, -1 if nothing found
  */
 static st32 find_attr_idx(const RBinDwarfDie *die, st32 attr_name) {
+	st32 i;
 	r_return_val_if_fail (die, -1);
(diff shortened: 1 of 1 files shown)


libr/anal/type_dwarf.c
```diff
@@ -46,8 +46,9 @@ static void enum_type_fini(void *e, void *user) {
  * @return st32 Index, -1 if nothing found
  */
 static st32 find_attr_idx(const RBinDwarfDie *die, st32 attr_name) {
+	st32 i;
```
trufae, line 49: Why signed? Use size_t

trufae: Oh i see.. it returns -1 on error. Better use int. So it will fit arch reg size instead. Anyway the pr lgtm as it fixes the build

PR #17309, opened 2020-07-18
Fix r_anal_get_bbaddr to include internal addrs ##anal

swoops:

**Detailed description**

The `r_anal_get_bbaddr` function was just a wrapper for `r_anal_get_block_at` . This is incorrect since `r_anal_get_block_at` wont return a block if the provided address is not the block entry point. This mistake causes `VV` to get lost if you current seek is in the middle of a block. 

This PR fixes the problem and ads a regression test using the `sb` command.

**Test plan**

The `sb` command relies on  `r_anal_get_bbaddr`, so you can test for off by one errors by seeking to tricky addresses. Below shows an example of `sb` performing correctly when at main-1, main, main+1, main+blocksize-1 and when you seek to the second block.

```
$ r2 test/bins/elf/vim 
 -- ---8<--------------------8<------------------8<-----------------8<------
[0x00055a40]> s main
[0x000543a0]> af
[0x000543a0]> sb
[0x000543a0]> s-1
[0x0005439f]> sb
[0x0005439f]> s main+1
[0x000543a1]> sb
[0x000543a0]> s +`afbi ~size[1]`-1 # last addr before next block
[0x0005441d]> sb
[0x000543a0]> s +`afbi ~size[1]`# start of next block
[0x0005441e]> sb
[0x0005441e]> 
```

**Closing issues**

None that I could find.

### DIFF
--- a/libr/anal/bb.c
+++ b/libr/anal/bb.c
@@ -112,24 +112,9 @@ R_API ut64 r_anal_bb_size_i(RAnalBlock *bb, int i) {
 	return idx_next != UT16_MAX? idx_next - idx_cur: bb->size - idx_cur;
 }
 
-static int __bb_cmp_contains(const void *in, const RBNode *tree_node, void *user) {
-	ut64 *ret = (ut64 *)user;
-	ut64 addr = *(ut64 *)in;
-	const RAnalBlock *bb = container_of (tree_node, const RAnalBlock, _rb);
-	if (addr < bb->addr) {
-		return -1;
-	}
-	if (addr >= bb->addr + bb->size) {
-		return 1;
-	}
-	*ret = bb->addr;
-	return 0;
-}
-
 /* returns the address of the basic block that contains addr or UT64_MAX if
  * there is no such basic block */
 R_API ut64 r_anal_get_bbaddr(RAnal *anal, ut64 addr) {
-	ut64 ret = UT64_MAX;
-	r_rbtree_find (anal->bb_tree, &addr, __bb_cmp_contains, &ret);
-	return ret;
+	RAnalBlock *bb = r_anal_bb_from_offset (anal, addr);
+	return bb? bb->addr: UT64_MAX
 }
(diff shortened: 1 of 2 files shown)


libr/anal/bb.c
```diff
@@ -112,9 +112,24 @@ R_API ut64 r_anal_bb_size_i(RAnalBlock *bb, int i) {
...
+	return 0;
+}
+
 /* returns the address of the basic block that contains addr or UT64_MAX if
  * there is no such basic block */
 R_API ut64 r_anal_get_bbaddr(RAnal *anal, ut64 addr) {
-	RAnalBlock *bb = r_anal_get_block_at (anal, addr);
-	return bb? bb->addr: UT64_MAX;
+	ut64 ret = UT64_MAX;
+	r_rbtree_find (anal->bb_tree, &addr, __bb_cmp_contains, &ret);
```
ret2libc, line 133: instead of reimplementing the search I think you can either use `r_anal_bb_from_offset` or `r_anal_get_blocks_in` and get the first returned `RAnalBlock`.

trufae:
two basic blocks can overlap, but only one can be on one offset. this is similar to using RAnal.get_function_in vs get_function.at() the first one expects any adddress, the second just the function entrypoint which is unique.

libr/anal/bb.c
swoops: Good call, I chose `r_anal_bb_from_offset`. It looks to function nearly identical to the code I wrote.

swoops:
ugh... forgot a `;`. I thought my build worked... 

Add a commit to fix, let me know if you want me to squash my embarrassments away.

libr/anal/bb.c
```diff
@@ -115,6 +115,6 @@ R_API ut64 r_anal_bb_size_i(RAnalBlock *bb, int i) {
 /* returns the address of the basic block that contains addr or UT64_MAX if
  * there is no such basic block */
 R_API ut64 r_anal_get_bbaddr(RAnal *anal, ut64 addr) {
-	RAnalBlock *bb = r_anal_get_block_at (anal, addr);
+	RAnalBlock *bb = r_anal_bb_from_offset (anal, addr);
```
trufae, line 118: what i was about to say is that this function should be renamed to r_anal_get_block_in() instead of r_anal_bb_from_offset(). but this can go into a separate PR

there's a bunch of stuff duplicated between bb.c and blocks.c still maybe good to clean it up

PR #17314, opened 2020-07-19
Only run aao in aaa on apple targets ##bin

trufae:

**Detailed description**

aaa runs aao on any binary, aao is smart enough to not do anything if there are missing sections and such, but aaa is not that smart. this piece of code makes that work only on macho-arm,ppc-x86 bins. any other target or objc programs? (i know there's opensource objc, but this is probably using a different format for that objc section.

**Test plan**

Screenshot 2020-07-19 at 04 17 20

**Closing issues**

...

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -9059,6 +9059,16 @@ static bool is_unknown_file(RCore *core) {
 	return true;
 }
 
+static bool is_apple_target(RCore *core) {
+	const char *arch = r_config_get (core->config, "asm.arch");
+	if (!strstr (arch, "ppc") && !strstr (arch, "arm") && !strstr (arch, "x86")) {
+		return false;
+	}
+	RBinObject *bo = r_bin_cur_object (core->bin);
+	r_return_val_if_fail (!bo || (bo->plugin && bo->plugin->name), false);
+	return bo? strstr (bo->plugin->name, "mach"): false;
+}
+
 static int cmd_anal_all(RCore *core, const char *input) {
 	switch (*input) {
 	case '?':
@@ -9260,9 +9270,11 @@ static int cmd_anal_all(RCore *core, const char *input) {
 				if (r_cons_is_breaked ()) {
 					goto jacuzzi;
 				}
-				oldstr = r_print_rowlog (core->print, "Check for objc references");
-				r_print_rowlog_done (core->print, oldstr);
-				cmd_anal_objc (core, input + 1, true);
+				if (is_apple_target (core)) {
+					oldstr = r_print_rowlog (core->print, "Check for objc references");
+					r_print_rowlog_done (core->print, oldstr);
+					cmd_anal_objc (core, input + 1, true);
+				}
 				r_core_task_yield (&core->tasks);
 				oldstr = r_print_rowlog (core->print, "Check for vtables");
 				r_core_cmd0 (core, "avrr");


libr/core/cmd_anal.c
```diff
@@ -9260,9 +9272,12 @@ static int cmd_anal_all(RCore *core, const char *input) {
 				if (r_cons_is_breaked ()) {
 					goto jacuzzi;
 				}
-				oldstr = r_print_rowlog (core->print, "Check for objc references");
-				r_print_rowlog_done (core->print, oldstr);
-				cmd_anal_objc (core, input + 1, true);
+				const bool apple_target = is_apple_target (core);
+				if (apple_target) {
```
ret2libc, line 9276:
```suggestion
				if (is_apple_target (core)) {
```
I would say this is short enough to avoid the use of another var

libr/core/cmd_anal.c
```diff
@@ -9059,6 +9059,18 @@ static bool is_unknown_file(RCore *core) {
 	return true;
 }
 
+static bool is_apple_target(RCore *core) {
+	const char *arch = r_config_get (core->config, "asm.arch");
+	if (!strstr (arch, "ppc") && !strstr (arch, "arm") && !strstr (arch, "x86")) {
```
ret2libc, line 9064: shall this be `||`?

libr/core/cmd_anal.c
```diff
@@ -9059,6 +9059,18 @@ static bool is_unknown_file(RCore *core) {
...
+		return false;
+	}
+	RBinObject *bo = r_bin_cur_object (core->bin);
+	if (bo && bo->plugin && bo->plugin->name) {
```
ret2libc, line 9068: I think we don't need all these checks.
Function `r_bin_object_new` is where `o->plugin` is assigned and there `plugin` is guaranteed to be not-null. Also `plugin->name` should never be null, so these checks should not be necessary (apart from the `if (bo)` check.

libr/core/cmd_anal.c
trufae: No. translating this to V:

`if ["ppc", "arm", "ppc"].index(asm_arch) != -1 {`

objc on apple platforms is only avalble on ppc,arm and ppc. so if arch is not any of those it must return

trufae: note its "strstr", not "strcmp"

libr/core/cmd_anal.c
```diff
@@ -9065,10 +9065,8 @@ static bool is_apple_target(RCore *core) {
 		return false;
 	}
 	RBinObject *bo = r_bin_cur_object (core->bin);
-	if (bo && bo->plugin && bo->plugin->name) {
-		return strstr (bo->plugin->name, "mach");
-	}
-	return false;
+	r_return_val_if_fail (bo && bo->plugin && bo->plugin->name, false);
```
ret2libc, line 9068:
```suggestion
	r_return_val_if_fail (!bo || (bo->plugin && bo->plugin->name), false);
```

otherwise if `bo` is NULL it will create a warning, while it is a valid case.

trufae: my bad

libr/core/cmd_anal.c
ret2libc: yeah sorry i misread the function indeed. LGTM

PR #17299, opened 2020-07-16
Add tests for creating #signatures

swoops:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)
- [x] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

I added the following tests to cmd_zignature:
x86-64 function with hole
x86-64 function with jumpback
x86-64 r_sign_fcn_bytes bounds check
80386 function with hole
80386 function with jumpback
80386 r_sign_fcn_bytes bounds check
MIPS R3000 function with hole
MIPS R3000 function with jumpback
MIPS R3000  r_sign_fcn_bytes bounds check
PowerPC function with hole
PowerPC function with jumpback
PowerPC r_sign_fcn_bytes bounds check
ARM function with hole
ARM function with jumpback
ARM r_sign_fcn_bytes bounds check
ARM aarch64 function with hole
ARM aarch64 function with jmpback
ARM aarch64 r_sign_fcn_bytes bounds check

Mostly this checks edge cases when creating byte signatures. I manually double checked that the output was expected. I did not include any tests for sparc yet.

**Test plan**

Ensure the tests are relevant and useful and well named. Make sure I am not attempting to match too much information. 
I might of gotten carried away, feel free to tell me to reduce things.

All tests pass on my machine:
```
r2r db/cmd/cmd_zignature 
Running from ....test
Loaded 66 tests.
Skipping json tests because jq is not available.
[66/66]                    66 OK         0 BR        0 XX        0 FX
Finished in 12 seconds.
```

**Closing issues**
Does not close anything.

...

PR #17319, opened 2020-07-20
Mark test as broken to avoid red PRs

ret2libc:
With #17299 and other PRs merged together with it, master became red. I'm not sure just adjusting the test case is the right thing to do, because of the reasons explained in https://github.com/radareorg/radare2/pull/17299#discussion_r457243994 .

It seems to me there is something very wrong with that feature.

This should at least make master/PRs green so we can deal with that problem with more calm while knowing it is broken.

cc @swoops


XVilka:
Could you please also fix this one?
```
[XX] db/formats/coff patching REL32 amd64
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -e io.cache=1 -Qc 'pd 2 @ sym.main+6
' bins/coff/coff.obj
-- stdout
@@ -1,3 +1,3 @@
-            0x0000006a      4c8d051a0000.  lea r8, [0x0000008b]        ; sym..data ; "Win64 assembly"; RELOC 32 .data @ 0x0000008b + 0x0
+            0x0000006a      4c8d051a0000.  lea r8, [0x0000008b]        ; sym..data ; "Win64 assembly"; RELOC 32 .data @ 0x0000008b
 (reloc..data)
             0x00000071      488d15220000.  lea rdx, [0x0000009a]       ; str.Coffee_time ; "Coffee time!"; RELOC 32 .data @ 0x0000008b + 0xf
```

swoops:
I made these tests too general. I shouldn't be testing that analysis is setting variable names/types correctly, analysis tests should do that. 

I will remove the `t`, `h`, and `r` signatures from these tests, that will cause tests to pass again. These tests were really to ensure the bytes signatures are being created properly. So ref type signatures don't need to be in there.

I will also double check that the removed signature types are well tested somewhere else. They should be tested to ensure analysis information is properly captured without judging the correctness of analysis information.

Does that sound good?

PR #17324, opened 2020-07-21
Fix O(n) in r_io_map_get() ##io

### LINKED ISSUE #14024: Slow O(n) in R_IO

GustavoLCR:

**Detailed description**

* Always add new map to skyline

##### `r2 -qq test/bins/pe/65535sects.exe  1.73s user 0.14s system 99% cpu 1.877 total`

----

**Test plan**

Tests pass

**Closing issues**

Closes #14024

### DIFF
--- a/libr/include/r_util/r_itv.h
+++ b/libr/include/r_util/r_itv.h
@@ -43,6 +43,11 @@ static inline ut64 r_itv_end(RInterval itv) {
 	return itv.addr + itv.size;
 }
 
+// Returns true if itv equals itv2
+static inline ut64 r_itv_eq(RInterval itv, RInterval itv2) {
+	return itv.addr == itv2.addr && itv.size == itv2.size;
+}
+
 // Returns true if itv contained addr
 static inline bool r_itv_contain(RInterval itv, ut64 addr) {
 	const ut64 end = itv.addr + itv.size;
--- a/libr/include/r_util/r_itv.h
+++ b/libr/include/r_util/r_itv.h
@@ -44,7 +44,7 @@ static inline ut64 r_itv_end(RInterval itv) {
 }
 
 // Returns true if itv equals itv2
-static inline ut64 r_itv_eq(RInterval itv, RInterval itv2) {
+static inline bool r_itv_eq(RInterval itv, RInterval itv2) {
 	return itv.addr == itv2.addr && itv.size == itv2.size;
 }
 
--- a/libr/io/io_map.c
+++ b/libr/io/io_map.c
@@ -71,61 +71,15 @@ static bool add_map_to_skyline(RIO *io, RIOMap *map) {
 
 // Store map parts that are not covered by others into io->map_skyline
 void io_map_calculate_skyline(RIO *io) {
-	bool *deleted = NULL;
 	r_pvector_clear (&io->map_skyline);
-	r_pvector_clear (&io->map_skyline_shadow);
-
-	int i = 0;
 	// Last map has highest priority (it shadows previous maps)
 	void **it;
 	r_pvector_foreach (&io->maps, it) {
 		add_map_to_skyline (io, (RIOMap *)*it);
 	}
-
-	const RPVector *skyline = &io->map_skyline;
-	RPVector *shadow = &io->map_skyline_shadow;
-	RInterval *cur_itv = NULL;
-
-	for (i = 0; i < r_pvector_len (skyline); i++) {
-		const RIOMapSkyline *part = r_pvector_at (skyline, i);
-		if (!part) {
-			continue;
-		}
-
-		ut64 part_from = part->itv.addr;
-		ut64 part_size = part->itv.size;
-
-		if (!cur_itv) {
-			cur_itv = r_itv_new (part_from, part_size);
-			if (!cur_itv) {
-				break;
-			}
-			continue;
-		}
-
-		if (part_from == r_itv_end (*cur_itv)) {
-			cur_itv->size += part_size;
-		} else {
-			if (!r_pvector_push (shadow, cur_itv)) {
-				R_FREE (cur_itv);
-				break;
-			}
-			cur_itv = r_itv_new (part_from, part_size);
-			if (!cur_itv) {
-				break;
-			}
-		}
-	}
-	if (cur_itv) {
-		if (!r_pvector_push (shadow, cur_itv)) {
-			R_FREE (cur_itv);
-		}
-	}
-out:
-	free (deleted);
 }
 
-RIOMap* io_map_new(RIO* io, int fd, int perm, ut64 delta, ut64 addr, ut64 size, bool do_skyline) {
+RIOMap* io_map_new(RIO* io, int fd, int perm, ut64 delta, ut64 addr, ut64 size) {
 	if (!size || !io || !io->map_ids) {
 		return NULL;
 	}
@@ -138,7 +92,7 @@ RIOMap* io_map_new(RIO* io, int fd, int perm, ut64 delta, ut64 addr, ut64 size, 
 	map->delta = delta;
 	if ((UT64_MAX - size + 1) < addr) {
 		/// XXX: this is leaking a map!!!
-		io_map_new (io, fd, perm, delta - addr, 0LL, size + addr, do_skyline);
+		io_map_new (io, fd, perm, delta - addr, 0LL, size + addr);
 		size = -(st64)addr;
 	}
 	// RIOMap describes an interval of addresses (map->from; map->to)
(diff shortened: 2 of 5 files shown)


libr/include/r_util/r_itv.h
```diff
@@ -43,6 +43,11 @@ static inline ut64 r_itv_end(RInterval itv) {
 	return itv.addr + itv.size;
 }
 
+// Returns true if itv equals itv2
+static inline ut64 r_itv_eq(RInterval itv, RInterval itv2) {
```
trufae, line 47: why not return bool?

libr/io/io_map.c
```diff
@@ -14,142 +14,81 @@
...
+	// `part` is the first RIOMapSkyline with part->itv.addr > new_part->itv.addr
+	r_pvector_lower_bound (skyline, new_part->itv.addr, slot, CMP_BEGIN_GTE_PART);
+	RIOMapSkyline *part = slot < r_pvector_len (skyline) ? r_pvector_at (skyline, slot) : NULL;
+	if (slot) {
+		RIOMapSkyline *prev_part = r_pvector_at (skyline, slot - 1);
+		if (r_itv_eq (new_part->itv, prev_part->itv)) {
+			// `new_part` == `prev_part`, just remove `prev_part`
+			r_pvector_remove_at (skyline, --slot);
+			part = NULL; // Skip other checks
+		} else if (r_itv_end (prev_part->itv) > r_itv_begin (new_part->itv)) {
```
trufae, line 40: this is called 3 times. cache the result in a const once

Screenshot 2020-07-21 at 09 53 16

libr/io/io_map.c
```diff
@@ -14,142 +14,81 @@
...
+			r_pvector_remove_at (skyline, slot);
+			part = slot < r_pvector_len (skyline) ? r_pvector_at (skyline, slot) : NULL;
 		}
-		part1->map = map;
-		part1->itv = (RInterval){ UT64_MAX, 1 };
-		if (!r_pvector_push (map_skyline, part1)) {
-			free (part1);
+		if (part && r_itv_overlap (new_part->itv, part->itv)) {
+			// Chop start of last `part` that intersects `new_part`
+			const size_t oaddr = r_itv_begin (part->itv);
```
trufae, line 67: why size_t instead of ut64?

libr/io/io_map.c
```diff
@@ -342,12 +282,15 @@ R_API RIOMap* r_io_map_get_paddr(RIO* io, ut64 paddr) {
...
-			return map;
-		}
+	const RPVector *skyline = &io->map_skyline;
+	size_t i, len = r_pvector_len (skyline);
+	r_pvector_lower_bound (skyline, addr, i, CMP_END_GTE_PART);
+	if (i == len) {
+		return NULL;
+	}
+	const RIOMapSkyline *sky = r_pvector_at (skyline, i);
+	if (sky->itv.addr <= addr) {
```
trufae, line 292: compress those 4 lines in a single return statement with the ternary operator

trufae:
Can you post the results of aaaa as well? io.map_get() is also used in analysis

libr/io/io_map.c
```diff
@@ -14,142 +14,81 @@
...
+
+	RIOMapSkyline *new_part = R_NEW (RIOMapSkyline);
+	new_part->map = map;
+	new_part->itv = map->itv;
+
+	// `part` is the first RIOMapSkyline with part->itv.addr > new_part->itv.addr
+	r_pvector_lower_bound (skyline, new_part->itv.addr, slot, CMP_BEGIN_GTE_PART);
+	RIOMapSkyline *part = slot < r_pvector_len (skyline) ? r_pvector_at (skyline, slot) : NULL;
+	if (slot) {
+		RIOMapSkyline *prev_part = r_pvector_at (skyline, slot - 1);
```
karliss, line 35: Something is wrong. lower_bound should return first value >= than  x. Either comment, comparison operator or r_pvector_lower_bound is incorrect.

karliss: `test_pvector_lower_bound` tests this. So that leaves one of the first two cases.

karliss: I'd say COMP_BEGIN_GTE_PART is at fault. Passing a comparison operator which claims that either `x > x` or `x < x` to algorithm which expects a comparison operator will likely give you a wrong result at best and memory corruption in the worst case.

Just don't try to be smart and use the known correct pattern for comparing `(x > y) - (x < y)`. Not only is it correct in all cases but can even generate code that is equally good if not better than `x < y ? -1 : 1`. https://godbolt.org/z/Teo97M

libr/io/io_map.c
```diff
@@ -9,147 +9,82 @@
...
+	RIOMapSkyline *new_part = R_NEW (RIOMapSkyline);
+	new_part->map = map;
+	new_part->itv = map->itv;
+	const ut64 new_part_end = r_itv_end (new_part->itv);
+
+	// `part` is the first RIOMapSkyline with part->itv.addr >= new_part->itv.addr
+	r_pvector_lower_bound (skyline, new_part->itv.addr, slot, CMP_BEGIN_GTE_PART);
+	RIOMapSkyline *part = slot < r_pvector_len (skyline) ? r_pvector_at (skyline, slot) : NULL;
+	if (slot) {
+		RIOMapSkyline *prev_part = r_pvector_at (skyline, slot - 1);
```
karliss, line 34: Doesn't this need to check if slot is 0?

libr/io/io_map.c
```diff
@@ -342,14 +278,14 @@ R_API RIOMap* r_io_map_get_paddr(RIO* io, ut64 paddr) {
...
 	r_return_val_if_fail (io, NULL);
-	void **it;
-	r_pvector_foreach_prev (&io->maps, it) {
-		RIOMap *map = *it;
-		if (r_itv_contain (map->itv, addr)) {
-			return map;
-		}
+	const RPVector *skyline = &io->map_skyline;
+	size_t i, len = r_pvector_len (skyline);
+	r_pvector_lower_bound (skyline, addr, i, CMP_END_GTE_PART);
```
karliss, line 279: Looks wrong. I would expect to check `upper_bound(x)-1` which should be the same as `lower_bound(x+1)-1`. Unit tests need to be expanded since this wasn't cought.

libr/io/io_map.c
GustavoLCR: yeah, it was wrong. got confused because the other macros are for non-inclusive end, while begin is inclusive.

libr/io/io_map.c
GustavoLCR: isnt that `if (slot)` doing it?

karliss: Sorry, not sure why I was thinking that it is `if (part)` not `if (slot)`.

GustavoLCR:
@trufae I gave up on the 'before' `aaaa` after waiting for it for an hour

Before:
`r2 -qc "aaa" test/bins/pe/65535sects.exe  129.61s user 0.30s system 100% cpu 2:09.90 total`

After:
`r2 -qc "aaa" test/bins/pe/65535sects.exe  14.00s user 0.47s system 100% cpu 14.467 total`
`r2 -qc "aaaa" test/bins/pe/65535sects.exe  58.95s user 0.27s system 99% cpu 59.234 total`

libr/io/io_map.c
GustavoLCR: @kariss couldnt find a way that this fails, so I guess its right

XVilka:
@ret2libc please take a look when you have time. For me it's LGTM.

libr/io/io_map.c
```diff
@@ -219,6 +149,8 @@ RIOMap* io_map_new(RIO* io, int fd, int perm, ut64 delta, ut64 addr, ut64 size,
 	r_pvector_push (&io->maps, map);
 	if (do_skyline) {
 		io_map_calculate_skyline (io);
+	} else {
+		add_map_to_skyline (io, map);
```
ret2libc, line 153: What is the difference between `io_map_calculate_skyline` and this new function? It seems a bit weird to see
```C
if (do_skyline) {
  // compute skyline
} else  {
  // do again something on skyline
}
```

GustavoLCR: @ret2libc `io_map_calculate_skyline()` resets everything and calculates both the skyline and the skyline shadow from scratch, while calling `add_map_to_skyline()` just adds the map to the skyline to make sure we can read from it.
This changes the behavior of `io_map_new()` as before you had to set `do_skyline` to `true` or call `r_io_update()` to read from a newly added map.

test/unit/test_io.c
```diff
@@ -5,7 +5,9 @@ bool test_r_io_mapsplit (void) {
 	RIO *io = r_io_new ();
 	io->va = true;
 	r_io_open_at (io, "null://2", R_PERM_R, 0LL, UT64_MAX);
+	mu_assert ("UT64_MAX not mapped", r_io_map_is_mapped (io, UT64_MAX));
```
ret2libc, line 8: you can use `mu_assert_true`  and `mu_assert_false` if you want.
There is also `mu_assert_null` and `mu_assert_notnull`.

PR #17326, opened 2020-07-21
Make root a argv RCmdDesc and adapt to form results ##newshell

ret2libc:

**Detailed description**

This is another step in better help generation and better shell.
The form I sent out some time ago received 23 responses (not a huge number, but not few either :)

43.5% of people don't like to always see `[?]` at the end of each command. I agree with this, as `?` should be available everywhere, so if something is always there, I don't see why we need to print `[?]` every time.

56.5% of people don't find the general arguments useful (by "general arguments" i mean those arguments that are shown at some top-level commands. For example if you do `?` it shows ` s[?] [addr]`, however not all subcommands of `s` accept an argument and some of them don't accept an address but something else).

56.5% never used the recursive help `?*`. This is interesting :)

This PR removes the general arguments and the `[?]` from each help.
image
image

Due to the fact that some commands are still handled in the old way, you can still find the current look in other commands:
image

**Test plan**

Use `cfg.newshell=true` and navigate the help.


ret2libc:
There are of course some broken tests like:
```

[XX] db/cmd/cmd_help "base help"
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '?~print
' -
-- stdout
@@ -1,2 +1,2 @@
-| p[?] [len]              print current block with format and length
-| x[?] [len]              alias for 'px' (print hexadecimal)
+| p  # print commands
+| x  # alias for 'px' (print hexadecimal)
``` but I'll adjust them later. I want to get some feedback about this first. So feel free to have a look anyway, even if CI is red.

File: libr/core/cmd_helps.help [hunk 1]
```diff
@@ -0,0 +1,231 @@
+#include <r_cmd.h>
+
+// root helps
+
+static const RCmdDescHelp system_help = {
```
ret2libc, line 5: BTW, later if we want to make this "easier" to contribute to for non-C devs or we just want to avoid all the boilerplate code, we may consider generating this file from a json or something and converting it to this format at build-time. This way people will be able to update the help without having to worry about C code. WDYT?

Maijin: +1

libr/core/cmd_write.c
```diff
@@ -850,16 +850,134 @@ static void cmd_write_pcache(RCore *core, const char *input) {
...
+	return w_incdec_handler (data, argc, argv, 4);
+}
+
+static RCmdStatus w8_incdec_handler(void *data, int argc, const char **argv) {
+	return w_incdec_handler (data, argc, argv, 8);
+}
+
 /* TODO: simplify using r_write */
-static int cmd_write(void *data, const char *input) {
+static int cmd_write (void *data, const char *input) {
```
ret2libc, line 973: space

libr/core/cmd.c
```diff
@@ -96,6 +96,8 @@ static int bb_cmpaddr(const void *_a, const void *_b) {
 }
 
 static void cmd_debug_reg(RCore *core, const char *str);
+#include "cmd_helps.help"
```
karliss, line 99: Please don't do this nonstandard code editors don't like non-standard extensions. Other problem is that now cmd_write.c depends on being included in a file which includes `cmd_helps.help` this also confuses code editors. I was hoping that after conversion to newshell we can get rid of including  cmd_x.c files and compile it properly.

karliss:
How would `wB` look in once `w?` uses the new mechanism. Wouldn't it just print `wB [value]   # Set or unset bits with given value`. 
From what I understand looking at `wv` help it is possible to `.usage = "wB[-]"`.

ret2libc:
> How would `wB` look in once `w?` uses the new mechanism. Wouldn't it just print `wB [value] # Set or unset bits with given value`.

It will actually be:
```
wB  # Set or unset bits with given value
```

And this makes me think that maybe for the time being we still want to see `[?]` for the inner nodes, to let users understand that there are additional options under a given node.

> From what I understand looking at `wv` help it is possible to `.usage = "wB[-]"`.

This is an option, yes, which can be used in case there are some particular sub-commands that we want to show from the top command.

ret2libc:
@karliss please have another look. There are still many things I'd like to improve on how commands are shown and probably document/explain even better how to use `RCmdDescHelp`, but I'm not sure I want to put everything in here...

This PR, at least, converts `w` to a argv command with a mix of oldinput/argv sub-commands. 

Few things I'm considering:
- using some kind  of automatic folding to show sub-commands, similar to what's done now. For example, `w6[de]` because `we` has just 2 sub-commands, but `wo[?]` because `wo` has many subcommands. 
- when it makes sense to show args_str and when it doesn't. For example, `wB` has 2 sub-commands `wB` and `wB-`, but both accept the same argument `<value>`, so it probably makes sense to show `wB[-] <value>` when showing it in `w?`. Maybe also `w6[de] <base64>|<hexstring>` has some sense. However, for `we` probably it doesn't make sense to show any argument when doing `w?`.

XVilka:
@ret2libc looks like it broke a few tests with a newshell: https://github.com/radareorg/radare2/pull/17326/checks?check_run_id=1090484378#step:13:405

XVilka (approved):

trufae:
What's holding this PR to be merged?

ret2libc:
> What's holding this PR to be merged?

Just waiting for karliss to have a look.

libr/core/cmd_helps.h
```diff
@@ -0,0 +1,104 @@
...
+extern const RCmdDescHelp y_help;
+extern const RCmdDescHelp z_help;
+
+// w0 helps
+
+extern const RCmdDescHelp w0_help;
+
+// w[1248][+-] helps
+
+extern const RCmdDescExample w_incdec_help_examples[];
```
karliss, line 61: No need to have example arrays available here.

karliss:
Overall seems reasonable there two things worth considering, not necessarily something to do within this PR.

* As pointed out by @ret2libc the current form of defining the help isn't perfect. Especially with having to modify 3 files. One more potential approach instead of having a generator could be having a parser load the help messages. Preferably help messages should be embedded within executable and not rely on external resource loading. No good way of doing that. I doubt we want to deal with linker scripts for all the targets. The other approach would be having a generator script which defeats the point of having a parser inside r2 instead of generator.

* How about changing RCmdArgvCb first argument from `void*` to `RCore *` ? It can be forward declared to deal with header loop problem. It's not really user data if we always pass RCore* and there is no API to configure per command user data when creating or adding a new command. It is very unlikely that rcmd will be reused outside r2. It's too late to change something for oldinput, but  not RCmdArgvCb. The benefits would be additional error checking and having to copy unnecessary line to all the command callbacks. It's not much, but it was already annoying for the old command handlers, with new ones there will be even more of them. So I think it's worth it.

ret2libc:
> As pointed out by @ret2libc the current form of defining the help isn't perfect. Especially with having to modify 3 files. 

Yes, adding a new commands requires some changes here and there.

> One more potential approach instead of having a generator could be having a parser load the help messages. Preferably help messages should be embedded within executable and not rely on external resource loading. No good way of doing that.

Do you mean there is no good way to "load" help messages without having them in external files? Not sure I got what you mean.

> The other approach would be having a generator script which defeats the point of having a parser inside r2 instead of generator.

I'm not sure I understand this either. When I think about a "generator" I think about a script that converts something expressed in a simpler(non C) form into cmd_helps.c + cmd_helps.h, so one does not have to deal with those two files. Are you referring to something else?

> How about changing RCmdArgvCb first argument from `void*` to `RCore *` ? It can be forward declared to deal with header loop problem. It's not really user data if we always pass RCore* and there is no API to configure per command user data when creating or adding a new command. It is very unlikely that rcmd will be reused outside r2. It's too late to change something for oldinput, but  not RCmdArgvCb. The benefits would be additional error checking and having to copy unnecessary line to all the command callbacks. It's not much, but it was already annoying for the old command handlers, with new ones there will be even more of them. So I think it's worth it.

Done in last commit.

karliss:
> Do you mean there is no good way to "load" help messages without having them in external files? Not sure I got what you mean.

I was referring to there being no good way for embedding an arbitrary data file within executable. As for generators I was referring to generating a c file from data file as one of approaches for embedding it into executable.

Issue #17237, opened 2020-07-08
Fix MODE_LEVENSTEIN diffing algorithm


XVilka:
See commented test in [`test/unit/test_diff.c`](https://github.com/radareorg/radare2/blob/master/test/unit/test_diff.c):
```
// Broken r_diff_buffers_distance_levenshtein, uncomment and see why it is incorrect
	// diff->type = 'l';
	// for (i = 0; i < R_ARRAY_SIZE (tests); i++) {
	// 	size_t la = strlen (tests[i].a), lb = strlen ((const char *)tests[i].b);
	// 	r_diff_buffers_distance (diff, tests[i].a, la, tests[i].b, lb, &distance, NULL);
	// 	snprintf (msg, sizeof msg, "levenshtein %s/%s distance", tests[i].a, tests[i].b);
	// 	mu_assert_eq (distance, tests[i].dis_distance, msg);
	// }
```
More information for the context: https://github.com/radareorg/radare2/pull/16986

PR #17330, opened 2020-07-21
Fix Levenstein distance on empty buffers ##radiff2

swoops:

**Detailed description**

I saw issue #17237 and took a look. I could not get the unit test to run but I spotted an unhanded edge case which the test uses. This PR will fix the case when one or both provided buffers are of length 0.

**Test plan**

I couldn't get the unit tests to run, but this works now:
```
radiff2 -ss /tmp/tests/1/*
File size differs 5 vs 0
similarity: 0.000
distance: 5
```

Previously distance was reported as 0.

**Closing issues**

Maybe #17237 ?

### DIFF
--- a/libr/util/udiff.c
+++ b/libr/util/udiff.c
@@ -239,8 +239,8 @@ R_API bool r_diff_buffers_distance_levenstein(RDiff *d, const ut8 *a, ut32 la, c
 	stop = bLen;
 	// Preliminary tests
 
-	//Do we have both files a & b, with positives sizes?
-	if (!aBufPtr || !bBufPtr || aLen < 0 || bLen < 0) {
+	//Do we have both files a & b
+	if (!aBufPtr || !bBufPtr) {
 		return false;
 	}
 
--- a/libr/util/udiff.c
+++ b/libr/util/udiff.c
@@ -168,6 +168,7 @@ R_API int r_diff_buffers(RDiff *d, const ut8 *a, ut32 la, const ut8 *b, ut32 lb)
 }
 
 R_API bool r_diff_buffers_distance_levenstein(RDiff *d, const ut8 *a, ut32 la, const ut8 *b, ut32 lb, ut32 *distance, double *similarity) {
+	r_return_val_if_fail (a && b, false);
 	const bool verbose = d? d->verbose: false;
 	/*
 	More memory efficient version on Levenshtein Distance from:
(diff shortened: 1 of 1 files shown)


ret2libc:
Just to understand: were you able to run the test but not to make the Levenstein commented code work or weren't you able to run the test at all?

libr/util/udiff.c
```diff
@@ -239,11 +239,22 @@ R_API bool r_diff_buffers_distance_levenstein(RDiff *d, const ut8 *a, ut32 la, c
 	stop = bLen;
 	// Preliminary tests
 
-	//Do we have both files a & b, and are they at least one byte?
-	if (!aBufPtr || !bBufPtr || aLen < 1 || bLen < 1) {
+	//Do we have both files a & b
+	if (!aBufPtr || !bBufPtr) {
```
trufae, line 243: uhm, it is valid to pass those nulls? shouldnt be asserting because of bad api usage?

swoops:
> Just to understand: were you able to run the test but not to make the Levenstein commented code work or weren't you able to run the test at all?

I don't think I got the tests to run at all. I ran `r2r unit` from the test directory because of a comment seen in `run_unit.sh` but output seemed to indicate I was not doing things correctly. I was excited to learn the diff algorithm anyways, so I went looking at that instead of r2r code. I spotted the missed edge case for when a zero length buffer. So I figured I would send a quick fix. If the test is only failing on `R("", "zzz", 3, 3)` and `R("meow", "", 4, 4)` then this PR should fix it. This code should also cover the untested `R("", "", 0, 0)` case.


> Fix the PR title, follow the rules described in #17335
I updated the PR title. If someone can confirm this is the fix for  #17237 then I will update the PR title to "Fix #17237 zero length buffer ##radiff2" or the like. Sorry my PR titles are always terrible, I appreciate the direction in  #17335.

libr/util/udiff.c
swoops: I agree, I updated to check that at function entry with `r_return_val_if_fail`.

ret2libc:
> I don't think I got the tests to run at all. I ran `r2r unit` from the test directory because of a comment seen in `run_unit.sh` but output seemed to indicate I was not doing things correctly.

I think if you do `cd test/unit; make` you should be able to compile the unit test. Then just run `./bin/test_diff`.

PR #17338, opened 2020-07-22
Add tests for “$foo=#!pipe“

trufae:

**Detailed description**

The old syntax doesnt work. this test proves the new syntax doesnt breaks

**Test plan**

i just wrote a test

**Closing issues**

change introduced in 765f2b6bd4098b8e90e46a45c174a71099bc0d56

### DIFF
--- a/test/db/cmd/cmd_interpret
+++ b/test/db/cmd/cmd_interpret
@@ -30,3 +30,22 @@ Eron
 
 EOF
 RUN
+
+NAME=$foo=#!pipe
+FILE=-
+CMDS=<<EOF
+#!pipe echo hello
+$foo=#!pipe echo hello
+$foo
+$-*
+e cfg.newshell=true
+$foo=#!pipe echo hello
+$foo
+$-*
+EOF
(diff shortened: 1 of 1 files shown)


test/db/cmd/cmd_interpret
```diff
@@ -30,3 +30,30 @@ Eron
...
+#!pipe echo hello
+$foo=#!pipe echo hello
+$foo
+?e --
+"$foo=#!pipe echo hello"
+$foo
+?e --
+$-*
+e cfg.newshell=true
+$foo=\#!pipe echo hello
```
ret2libc, line 46: ops, wait. I need to check one thing. There is something wrong here with newshell.

ret2libc: https://github.com/radareorg/radare2/issues/17351

ret2libc: that said, I think this test is ok. After all, an alias is just a string, so the only way to pass `#` is to either escape it or wrap it. For now, escaping is the only supported option though.

PR #17343, opened 2020-07-23
r_cons_print fix for very large output ##cons

HoundThe:

**Detailed description**

There was a segfault linked to a very large binary with DWARF information (http://xvilka.me/vmlinux_ida_dwarf_fail.zip) when using `id`. It was caused by the information being over several GiB of text and r_cons_context using `int` for it's size and length, changing it to ut64 fixed the segfaulting behaviour.

### DIFF
--- a/libr/include/r_cons.h
+++ b/libr/include/r_cons.h
@@ -447,8 +447,8 @@ typedef struct r_cons_context_t {
 	RConsGrep grep;
 	RStack *cons_stack;
 	char *buffer;
-	ut64 buffer_len;
-	ut64 buffer_sz;
+	size_t buffer_len;
+	size_t buffer_sz;
 
 	bool breaked;
 	RStack *break_stack;
--- a/libr/include/r_cons.h
+++ b/libr/include/r_cons.h
@@ -447,8 +447,8 @@ typedef struct r_cons_context_t {
 	RConsGrep grep;
 	RStack *cons_stack;
 	char *buffer;
-	int buffer_len;
-	int buffer_sz;
+	ut64 buffer_len;
+	ut64 buffer_sz;
 
 	bool breaked;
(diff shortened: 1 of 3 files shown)


libr/include/r_cons.h
```diff
@@ -447,8 +447,8 @@ typedef struct r_cons_context_t {
 	RConsGrep grep;
 	RStack *cons_stack;
 	char *buffer;
```
trufae, line 449: This should be done using RBuffer ideally.

libr/include/r_cons.h
```diff
@@ -447,8 +447,8 @@ typedef struct r_cons_context_t {
...
-	int buffer_len;
-	int buffer_sz;
+	ut64 buffer_len;
```
trufae, line 450: What about size_t?

HoundThe: Sure, could be size_t aswell probably, I think size_t is guaranteed to hold size of any object so it should cover capacity of whole possible memory.

HoundThe:
> Which situation the buffer is thst big ?

As mentioned in the PR, for a large binary with a lot of DWARF debugging information, this happens when printing all the information out.

libr/include/r_cons.h
ret2libc: however size_t could be 32bit on 32bit system. Would this risk to reintroduce the wrong behaviour on those platforms?

HoundThe: well 32bit systems wouldn't fit more 2^32 bytes of output anyways

ret2libc: Good point :D :D

trufae: use size_t and u get my +1

Issue #16535, opened 2020-04-12
Build all radare2 plugins weekly in the cronjob of GitHub Actions


XVilka:
To check if something is broken, we can set the job every few days or weekly, like Coverity one.

radare:
Thsi is r2extras

XVilka:
@radare no, extras fires only on the change in the extras. This one should reflect ongoing radare2 changes and check if all plugins are compatible still.

ret2libc:
I doubt I'm gonna be able to do it for 4.5.0, moving already to 4.6.0

ret2libc:
I will be able to do this for next release, it's already in plan. Also, this is done on the `release-*` branches, so it's already something.

ret2libc:
In theory this should be done, since the merge of #17355 . I'm going to close this once I see it working. https://github.com/radareorg/radare2/actions?query=workflow%3ACI+event%3Aschedule should list something in next days. I'll keep an eye on this and close if everything works right.

ret2libc:
Ok, looks like this ran correctly. https://github.com/radareorg/radare2/actions?query=workflow%3ACI+event%3Aschedule There are problems with the build-static build, but everything runs. Closing.

PR #17355, opened 2020-07-24
Merge all things in one workflow file, reducing duplication of CI code

ret2libc:
### Reasons for this change
Whenever we want to test some new configuration, we may want to test it both on PRs and/or on the master branch. Moreover, some configurations may be tested only on releases. However, we also want to test some particular things on PRs with particular names. For example, the iOS package is built only on release, but if you are doing some changes related to that, you may want to test the iOS package built also in PR. The same is true for ASAN, static, osx, etc.

At the same time, when doing a release we want to ensure that all "regular" checks are performed and are green and after that performs additional operations (e.g. built iOS package, build windows installer, etc.)

### Current state
Right now the only solution was to create multiple workflow files and duplicate the code needed to do the tests in both files. This may result in a lot of duplicated "devops" code, having the risk of testing things differently when on PR and when doing releases or commits on master branch.

### Solution proposed in this PR
By merging everything in one single file we can avoid the duplication of the "devops" code and just choose when to run some particular jobs based on the kind of triggering event (e.g. some jobs may be run only on schedule, others only when doing a release, etc.). At the same time, it allows to create dependencies between jobs (e.g. the create-release can only run if all the various build and tests jobs on various configurations pass).

#### Possible issues of this PR
- some jobs will be skipped, but they are going to be shown anyway in the list of jobs. For example, in this PR the job "Build on CentOS 6" will not run, but you will see anyway the entry below, though it will be marked as "skip".
- on PRs, linux-acr-clang-build runs instead of linux-acr-clang-tests, but you will see anyway linux-acr-clang-tests (though nothing will be run there).


XVilka:
To be honest, I don't like this approach. While it reduces some code duplication the resulting file is massive and hard to navigate. Currently CI configuration neatly split allowing easier understanding and improvements.

ret2libc:
> To be honest, I don't like this approach. While it reduces some code duplication the resulting file is massive and hard to navigate. Currently CI configuration neatly split allowing easier understanding and improvements.

You are right. I'm not very convinced by this either, though I've read that's the way github workflows are supposed to work right now :( But yeah, I agree it does not look ideal. The fact is that I would like to have things dependent on one another and this seems the only way.

To summarize, this is what I'm trying to achieve:
- make release steps depend on build-and-test jobs (linux-acr-clang-test, linux-meson-gcc-test, etc.)
- make sure releases steps do not work if some of the build steps do not work well.
- reduce duplication of "devops" code
- allow to run some steps of the release process (e.g. building cydia package or android) when a particular branch name is used. This would be useful to check changes to those things more easily.


I want to first see how this works, we can then choose what to do. I won't push too hard for this, I understand it has some pros and cons.

ret2libc:
BTW, this approach should allow us to easily perform things like https://github.com/radareorg/radare2/issues/16535 without having duplicate yaml code around. Of course, it would be ideal if we could split the big yaml in smaller ones and "include" them in one file, but this is still not possible in github ci :(

ret2libc:
So @XVilka I think I'm done with this. WDYT? Again, I understand it is not a perfect solution, though the alternative is to duplicate a lot of steps.

For example, one of the requests from @trufae is that we want to be able to test the ios build on PRs that have the `ios` word in the branch name. However, if we don't put everything in one file we would need to create a new workflow file and duplicate the ios build steps in that file. Same for all the other checks (debian, osx, android, etc.).

ret2libc:
Another downside of this is that it adds anyway some checks to the PR/commit list, though you can see them as "skipped". Those checks are only performed in some cases (e.g. when releasing or when doing particular PRs).

.github/workflows/ci.yml
```diff
@@ -330,7 +560,7 @@ jobs:
 
   create-release:
     runs-on: ubuntu-latest
-    needs: [create-tarball, test-deb, build-android, test-osx-pkg, test-windows-clang_cl, build-ios-cydia, build-extras]
+    needs: [build-and-test, build-centos6, build-debian, build-static, test-deb, build-android, test-osx-pkg, test-windows-clang_cl, build-ios-cydia, build-extras]
```
XVilka, line 563: Maybe make this a multiline list?

XVilka:
I don't 100% like this, but lets proceed then. Maybe we can ask GitHub to allow includes in GitHub Actions.

XVilka:
One more thing - what about better naming of the actions?
Right now it's a bit inconsistent then:
image

wargio:
can i suggest to have the suffix `-notests` instead of `-tests-false` and remove `-tests-true` or even better `-tests` `-tests-true` and remove `-tests-false`

wargio:
look ok, but why there are some cancelled jobs?

ret2libc:
> look ok, but why there are some cancelled jobs?

They are not cancelled, they are skipped. IMO github should just *not* show those skipped jobs at all in the list, but apparently it is not possible at the moment... :(

XVilka:
@thestr4ng3r @trufae are you ok with this change? Please take a look, otherwise lets merge it.

ret2libc:
Ok, conflicts solved.
I have tried to summarize, again, the reasons for this PR. See https://github.com/radareorg/radare2/pull/17355#issue-456370071

In particular, this would have been useful for recent PRs done by @trufae related to sdk.sh, ios, static fixes.

trufae (approved):

ret2libc:
As an example, you can see https://github.com/radareorg/radare2/pull/17525 and https://github.com/radareorg/radare2/actions/runs/225017795 . The PR was named `improve-workflow-ios-osx-debian` and indeed ios, osx and debian jobs ran.

PR #17357, opened 2020-07-26
r2pm: Support cleaning of package cloned to <package>.git ##tools

kazarmy:

**Detailed description**

There is at least 1 r2pm package that clones itself into a \<package\>.git folder. This pr allows cleaning of those kind of packages.

**Test plan**

`r2pm -c r2ghidra-dec`

**Closing issues**

...

### DIFF
--- a/binr/r2pm/r2pm
+++ b/binr/r2pm/r2pm
@@ -584,8 +584,11 @@ case "$1" in
 			if [ -d "${R2PM_GITDIR}/$2" ]; then
 				echo "Cleaning up ${R2PM_GITDIR}/$2..."
 				rm -rf "${R2PM_GITDIR}/$2"
+			elif [ -d "${R2PM_GITDIR}/$2.git" ]; then
+				echo "Cleaning up ${R2PM_GITDIR}/$2.git..."
+				rm -rf "${R2PM_GITDIR}/$2.git"
 			else
-				echo "Cannot find $2 in ${R2PM_GITDIR}"
+				echo "Cannot find $2 or $2.git in ${R2PM_GITDIR}"
 				RC=1
 			fi
 			shift


trufae:
Maybe it should clean Wrkdir instead. Also, which packages are behaving like this?

kazarmy:
> Maybe it should clean Wrkdir instead.

sometimes the only quick way to fix a problem is to clean everything ... besides, cleaning Wrkdir can be done during installation e.g. [here](https://github.com/radareorg/radare2-pm/blob/dafdd43b6bae6dc166afbe30bdca7f16041ef989/db/r2ghidra-dec#L10) ... in this case r2ghidra-dec is a pretty large download but an `r2pm -ci` due to a release shouldn't happen too often

> Also, which packages are behaving like this?

apart from r2ghidra-dec, i didn't check ... anyway i think in this case r2pm should be accommodating

trufae:
I would rather prefer to fix the r2ghidra pkg instead of doing this

kazarmy:
```
db]$ git grep '\.git"'
emscripten:R2PM_GIT "https://github.com/juj/emsdk.git"
os9:R2PM_GIT "https://github.com/thestr4ng3r/r2-os9.git"
r2commodore:R2PM_GIT "https://github.com/thestr4ng3r/r2commodore.git"
r2ghidra-dec:R2PM_GIT "https://github.com/radareorg/r2ghidra-dec.git"
retdec-r2plugin:R2PM_GIT "https://github.com/avast/retdec-r2plugin.git"
```

I think cloning to \<package\>.git is frequent enough that we should support \<package\>.git cleaning as well.

trufae:
Frequent enough? It never happens to me. On which environment/situation is this mandatory? It can be removed also r2pm_git accepts a 2nd arg with the name of the directory to clone it

PR #17359, opened 2020-07-26
Change Itanium VTable detection heuristic ##anal

HoundThe:

**Detailed description**

I've been testing out the VTable, RTTI detection and parsing, I've tried some exotic Arches and found that sometimes only some Vtables are found, so far we assumed that vtable always have to be in the same section as a vtable, but in some binaries (PowerPC one, some RTTIs were in the .rodata and others were in .sdata) that wasn't the case, I've relaxed the requirement so we only check if the RTTI pointer points to an initialized data section which improved the detection. But I don't know how to properly detect such section in Radare, I got inspired by the implementation for possible Vtable section.

I've looked into Retdec way and they only seem to check if RTTI is a valid pointer somewhere in the binary and then they parse the RTTI, if any crucial step of the parsing fails, they take it as an invalid RTTI and thus invalid Vtable

We can maybe try the same approach, that would also help us to not rely on flags for knowing the TypeInfoClass, also because in some examples relying on the flags didn't work and we lost more RTTI information because of it, because it assumes that anything that doesn't have the flag is the default TypeInfo_Class

### DIFF
--- a/libr/anal/vtable.c
+++ b/libr/anal/vtable.c
@@ -90,6 +90,7 @@ static bool vtable_section_can_contain_vtables(RBinSection *section) {
 	return !strcmp (section->name, ".rodata") ||
 		!strcmp (section->name, ".rdata") ||
 		!strcmp (section->name, ".data.rel.ro") ||
+		!strcmp (section->name, ".data.rel.ro.local") ||
 		r_str_endswith (section->name, "__const");
 }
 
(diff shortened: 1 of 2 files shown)


XVilka:
@HoundThe can you please add a test for this?

HoundThe:
@XVilka Should that be in a form of Unit tests? Or the class db test like the rest?

XVilka:
class db test is fine.

PR #17362, opened 2020-07-27
Add `acg` - printing class hierarchy graph ##anal

HoundThe:

**Detailed description**

So I was thinking about adding ability to print out class hierarchy graph, based on class information in anal/classes
I've made a rough idea, I still need to look into agraph implementation more so I can make it look better, maybe add other options, this was made just by looking at the agraph interface to sketch it out. What do you think?

PS: I know dogs don't come out of cats

Example now:
```
[0x00001120]> acl
Bat
Bird
Cat
Dog
Kappa
Mammal
[0x00001120]> acb Bat Bird
[0x00001120]> acb Bat Mammal
[0x00001120]> acb Kappa Cat
[0x00001120]> acb Dog Cat
[0x00001120]> acb Cat Mammal
[0x00001120]> acg
┌────────────────────┐    ┌────────────────────┐
│  Bird              │    │  Mammal            │
└────────────────────┘    └────────────────────┘
    v                           t
     └                      ───╯  f
      └                 ───╯      │
       └            ───╯          │
        └       ───╯              │
         └   ──╯                  │
    ┌────────────────────┐    ┌────────────────────┐
    │  Bat               │    │  Cat               │
    └────────────────────┘    └────────────────────┘
                                    t f
                                 ──╯   └─
                             ───╯        └─
                         ───╯              └─
                      ──╯                    └─
                 ┌────────────────────┐    ┌────────────────────┐
                 │  Dog               │    │  Kappa             │
                 └────────────────────┘    └────────────────────┘
```

### DIFF
--- a/libr/anal/class.c
+++ b/libr/anal/class.c
@@ -3,6 +3,7 @@
 #include <r_anal.h>
 #include <r_vector.h>
 #include "../include/r_anal.h"
+#include "../include/r_agraph.h"
 
 static void r_anal_class_base_delete_class(RAnal *anal, const char *class_name);
 static void r_anal_class_method_delete_class(RAnal *anal, const char *class_name);
@@ -1217,4 +1218,42 @@ R_API void r_anal_class_list_vtable_offset_functions(RAnal *anal, const char *cl
 		}
 		ls_free (classes);
 	}
+}
+
+/**
+ * @brief Prints an ASCII graph from all classes and their inheritance information
+ * 
+ * @param anal 
+ * @return R_API 
+ */
+R_API void r_anal_class_print_inheritance_graph(RAnal *anal) {
+	RAGraph *class_graph = r_agraph_new (r_cons_canvas_new (1,1));
+	SdbList *classes = r_anal_class_get_all (anal, true);
+	HtPP/*<char *name, RANode *node>*/ *hashmap = ht_pp_new0 ();
+	SdbListIter *iter;
+	SdbKv *kv;
+	ls_foreach (classes, iter, kv) {
+		const char *name = sdbkv_key (kv);
+		RANode *curr_node = ht_pp_find (hashmap, name, NULL);
+		if (!curr_node) { // If not visited yet
+			curr_node = r_agraph_add_node (class_graph, name, NULL);
+			ht_pp_insert (hashmap, name, curr_node);
+		}
+		RVector *bases = r_anal_class_base_get_all (anal, name);
+		RAnalBaseClass *base;
+		r_vector_foreach (bases, base) {
+			bool base_found = 0;
+			RANode *base_node = ht_pp_find (hashmap, base->class_name, &base_found);
+			if (!base_found) {
+				// Speed it up and already add base classes if not visited yet
+				base_node = r_agraph_add_node (class_graph, base->class_name, NULL);
+				ht_pp_insert (hashmap, name, curr_node);
+			}
+			r_agraph_add_edge (class_graph, base_node, curr_node);
+		}
+		r_vector_free (bases);
+	}
+	ls_free (classes);
+	r_agraph_print (class_graph);
+	r_agraph_free (class_graph);
 }
--- a/libr/anal/class.c
+++ b/libr/anal/class.c
@@ -1226,7 +1226,7 @@ R_API void r_anal_class_list_vtable_offset_functions(RAnal *anal, const char *cl
  * @param anal 
  * @return R_API 
  */
-R_API void r_anal_class_print_inheritance_graph(RAnal *anal) {
+R_API RAGraph *r_anal_class_print_inheritance_graph(RAnal *anal) {
 	RAGraph *class_graph = r_agraph_new (r_cons_canvas_new (1,1));
 	SdbList *classes = r_anal_class_get_all (anal, true);
 	HtPP/*<char *name, RANode *node>*/ *hashmap = ht_pp_new0 ();
@@ -1254,6 +1254,5 @@ R_API void r_anal_class_print_inheritance_graph(RAnal *anal) {
 		r_vector_free (bases);
 	}
 	ls_free (classes);
-	r_agraph_print (class_graph);
-	r_agraph_free (class_graph);
+	return class_graph;
 }
--- a/libr/anal/class.c
+++ b/libr/anal/class.c
@@ -3,7 +3,7 @@
 #include <r_anal.h>
 #include <r_vector.h>
 #include "../include/r_anal.h"
-#include "../include/r_agraph.h"
+#include "../include/r_util/r_graph.h"
 
 static void r_anal_class_base_delete_class(RAnal *anal, const char *class_name);
 static void r_anal_class_method_delete_class(RAnal *anal, const char *class_name);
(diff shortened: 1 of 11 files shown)


XVilka:
Idea is good. I recommend to try it on a real binaries with a huge inheritance tree, to check how it might look like. Also it might be of particular interest for Cutter folks @ITAYC0HEN @karliss

karliss:
For this to to be usable by Cutter it needs to expose proper API.
Ideally I would like to see something like this, it would be nice to refactor the existing graph functions to use similar structure.
* r_anal_class_get_inheritance_graph -> creates and returns RAGraph*
* r_core_agraph_print refactored or split into two functions so that it can be passed any RAgraph* instead of reading global graph instance from core->graph
* r2 command would looks something like
```
graph = r_anal_class_get_inheritance_graph()
r_core_agraph_print(graph, input); // input potentially containing extra arguments for choosing output type
r_agraph_free(graph);
```
The benefits to such structure that all r2 graph output modes are easily available and it also makes the C API more usable.
As a general principle what I would like to see more in R2 to is better separation of core logic from UI code - console printing, r2 command argument parsing, default values based on global variables representing text UI state.

ITAYC0HEN:
I very much like the idea!! If it will be under the `ag` commands and implement the required information of a graph we can rather easily parse them and present them to the user

karliss:
Correction it might be better for r_anal_class_get_inheritance_graph to return `RGraph` or something similar which is abstract representation of graph without any drawing logic. RAGraph should be created from RGraph when you decide to draw something unless unless it gets used in format that don't require Ascii graphic output like dot, gml or json.

karliss:
Most of the refactoring I mentioned should probably be done in separate PR, but some of the smaller things to make more of the modes for acg and make it easier to update it's code once graph refactoring is done could be done within this PR.

I don't think all of the graph commands need be bellow `ag`. As long as interface is consistent it can be anywhere. Placing everything under `ag` is also limited by number of available characters and logical naming limiting even sooner.

HoundThe:
Looking at it now, I think maybe creating RAGraph as a start and then parsing the info from that might be better I guess? Or making both, RAGraph and RGraph and using the API you want. Transforming RGraph to RAGraph every time for `acg` seems a bit unnecessary. But transforming RAGraph into RGraph might be very easy as RAGraph contains RGraph that you can just take and use with the nodes that have RANode as the generic data.

libr/anal/meson.build
```diff
@@ -206,7 +206,8 @@ pkgconfig_mod.generate(r_anal,
     'r_syscall',
     'r_search',
     'r_cons',
-    'r_flag'
+    'r_flag',
+	'r_agraph'
```
ret2libc, line 210: use the right spaces

libr/anal/meson.build
```diff
@@ -207,7 +207,7 @@ pkgconfig_mod.generate(r_anal,
     'r_search',
     'r_cons',
     'r_flag',
-	'r_agraph'
+    'r_agraph'
```
trufae, line 210: there's no such thing

libr/anal/class.c
```diff
@@ -3,6 +3,7 @@
 #include <r_anal.h>
 #include <r_vector.h>
 #include "../include/r_anal.h"
+#include "../include/r_agraph.h"
```
trufae, line 6: this api is in RCore, so it cannot be included from RAnal

libr/core/cmd_anal.c
```diff
@@ -9902,6 +9902,12 @@ static void cmd_anal_classes(RCore *core, const char *input) {
 	case 'm': // "acm"
 		cmd_anal_class_method (core, input + 1);
 		break;
+	case 'g': // "acg"
+	{
+		RAGraph *graph = r_anal_class_print_inheritance_graph (core->anal);
+		r_agraph_print (graph);
+		r_agraph_free (graph);
+	} break;
```
trufae, line 9910: bad indent

libr/core/cmd_anal.c
```diff
@@ -9902,6 +9902,12 @@ static void cmd_anal_classes(RCore *core, const char *input) {
 	case 'm': // "acm"
 		cmd_anal_class_method (core, input + 1);
 		break;
+	case 'g': // "acg"
+	{
+		RAGraph *graph = r_anal_class_print_inheritance_graph (core->anal);
+		r_agraph_print (graph);
```
trufae, line 9908: check for null

libr/include/r_anal.h
```diff
@@ -2052,6 +2053,7 @@ R_API void r_anal_class_list(RAnal *anal, int mode);
 R_API void r_anal_class_list_bases(RAnal *anal, const char *class_name);
 R_API void r_anal_class_list_vtables(RAnal *anal, const char *class_name);
 R_API void r_anal_class_list_vtable_offset_functions(RAnal *anal, const char *class_name, ut64 offset);
+R_API RAGraph *r_anal_class_print_inheritance_graph(RAnal *anal);
```
trufae, line 2056: this is why libr/core/canal.c exists

libr/anal/class.c
```diff
@@ -1245,45 +1255,45 @@ R_API RGraph *r_anal_class_get_inheritance_graph(RAnal *anal) {
 	if (!classes) {
 		r_graph_free (class_graph);
 	}
+	// O(1) lookup cache for processed nodes
+	// to lookup parents to create edges
 	HtPP /*<char *name, RGraphNode *node>*/ *hashmap = ht_pp_new0 ();
 	if (!hashmap) {
 		r_graph_free (class_graph);
 		ls_free (classes);
```
trufae, line 1263: Missing return

libr/anal/class.c
```diff
@@ -1218,3 +1219,99 @@ R_API void r_anal_class_list_vtable_offset_functions(RAnal *anal, const char *cl
...
+static void free_class_node_info(void *ptr) {
+	RAnalClassNodeInfo *info = ptr;
+	free (info->body);
+	free (info->title);
+	free (info);
+}
+
+static RAnalClassNodeInfo *create_class_node_info(char *title, char *body, ut64 offset) {
+	RAnalClassNodeInfo *data = R_NEW0 (RAnalClassNodeInfo);
+	if (!data) {
```
trufae, line 1232: Inverse if and just return data

libr/anal/class.c
```diff
@@ -1218,3 +1219,99 @@ R_API void r_anal_class_list_vtable_offset_functions(RAnal *anal, const char *cl
...
+				goto failure;
+			}
+			curr_node->free = free_class_node_info;
+			ht_pp_insert (hashmap, name, curr_node);
+		}
+		// Travel all bases to create edges between parents and child
+		RVector *bases = r_anal_class_base_get_all (anal, name);
+		RAnalBaseClass *base;
+		r_vector_foreach (bases, base) {
+			bool base_found = 0;
```
trufae, line 1288: False not 0

XVilka:
It looks better now. Could you please add a test?

libr/include/r_anal.h
```diff
@@ -2052,6 +2059,7 @@ R_API void r_anal_class_list(RAnal *anal, int mode);
 R_API void r_anal_class_list_bases(RAnal *anal, const char *class_name);
 R_API void r_anal_class_list_vtables(RAnal *anal, const char *class_name);
 R_API void r_anal_class_list_vtable_offset_functions(RAnal *anal, const char *class_name, ut64 offset);
+R_API RGraph *r_anal_class_get_inheritance_graph(RAnal *anal);
```
karliss, line 2062: The type of user data inside RGraph is part of interface which needs to be documented the same way it is for RList or RVector.

libr/include/r_anal.h
```diff
@@ -2020,6 +2021,12 @@ typedef enum {
 	R_ANAL_CLASS_ERR_OTHER
 } RAnalClassErr;
 
+typedef struct r_anal_class_node_info_t {
+	char *title;
+	char *body;
+	ut64 offset;
+} RAnalClassNodeInfo;
```
karliss, line 2028: The purpose of having user data structure with these fields was so that isn't specific to single use case and can be reused by large portion of other graphs. This means that the structure needs to be declared in rgraph.h or soemwhere else in util, the structure name needs to be more generic. Similar changes apply to  rgraph->ragraph conversion function, but that one should probably be in ragraph.

libr/include/r_anal.h
```diff
@@ -18,6 +18,7 @@
 #include <set.h>
 #include <r_flag.h>
 #include <r_bin.h>
+#include <r_agraph.h>
```
karliss, line 21: This should probably be rgraph instead of r_agraph now.

libr/include/r_anal.h
HoundThe: Yes, moving the function to agraph module and the generic data node into rgraph is much better.

libr/anal/class.c
```diff
@@ -1218,3 +1219,99 @@ R_API void r_anal_class_list_vtable_offset_functions(RAnal *anal, const char *cl
 		ls_free (classes);
 	}
 }
+
+static void free_node_info(void *ptr) {
```
karliss, line 1223: Since RGraphNodeInfo is declared in r_graph.h shouldn't this and the constructor be there as well?

HoundThe: It could be, but I am not sure. Because in this specific case the free function abides the RListFree signature, so it's untyped with the void *.

libr/anal/class.c
```diff
@@ -3,6 +3,7 @@
 #include <r_anal.h>
 #include <r_vector.h>
 #include "../include/r_anal.h"
+#include "../include/r_util/r_graph.h"
```
ret2libc, line 6:
```suggestion
#include <r_graph.h>
```

test/unit/test_agraph.c
```diff
@@ -0,0 +1,109 @@
...
+	r_core_cmd0 (core, "ac C");
+	r_core_cmd0 (core, "ac D");
+	r_core_cmd0 (core, "acb B A");
+	r_core_cmd0 (core, "acb C A");
+	r_core_cmd0 (core, "acb D B");
+	r_core_cmd0 (core, "acb D C");
+
+	RGraph *graph = r_anal_class_get_inheritance_graph (core->anal);
+	mu_assert_notnull (graph, "Couldn't create the graph");
+	mu_assert_eq (graph->nodes->length, 4, "Wrong node count");
```
ret2libc, line 20: JFYI, you could have used `graph->n_nodes`. Also in the other places.

libr/include/r_util/r_graph.h
```diff
@@ -7,6 +7,15 @@
 extern "C" {
 #endif
 
+/**
```
ret2libc, line 10: This looks wrong to me, from an API point of view. It is not generic IMO. actually, it assumes the graph needs to be drawn and that it has an offset a body and a title, which may very well not be true. Also, by having this thing here it seems like all nodes of the rgraph are of this type, which is again not true.

IMHO this should go somewhere else. Anal maybe.

As I previously said, I like to think about this as simple data structures, like RList, RVector or similar. We don't have particular data definitions in the header files rlist.h, r_vector.h and I don't see why we should have this here in this case.

XVilka:
Sorry, seems page wasn't updated. @HoundThe please send addressing @ret2libc feedback in a separate PR.

libr/include/r_util/r_graph.h
HoundThe: :D Ye, that's true also, then I am unsure where to move it, I guess there isn't much of other option than r_anal.h middleground

ret2libc: `r_anal.h` is the right option i think, since the RGraph is returned from a r_anal method it should be that module to define the data contained within. In the same way as RBin defines RBinSymbol when it returns a RList of symbols of a binary.

Also, IMHO it would be better to have a sort of RAnalClass structure rather than a "generic" graph-drawing structure (at least for now). It may make sense to think about something much more generic for drawing graphs once we realize we have multiple points in the code where this is required. The fact is that in r2 maybe the title is just the class name, but we can't know if Cutter, for example, would like to also show something else in the graph title (maybe it wants to show both the demangled and mangled name? or maybe make it configurable? I don't know, but IMO it's better to return the class data and let the actual users of the API handle the drawing in the way it prefers, rather than forcing a view of the data to the users of the r_anal_class_get_inheritance_graph function). Also, note that maybe I want the relationships between the classes but I actually don't want to draw them, so forcing a draw-like structure does not seem the best option from API PoV.

karliss: Drawable graph is better than old situation of no API/getting drawn graph and creating AGraph even when it isn't necessary.  In most cases Cutter doesn't need to reinvent the wheel and can use whatever r2 decides is the best label. 

If you want class hierarchy information for non drawing purposes using class specific structures there is already API for that -> `r_anal_class_base_get_all`, `r_anal_class_get_all` and the other related methods.

If you really don't like RGraphNodeInfo in rgraph then it can be split into 3 parts: rgraph, drawable graph, ragraph.

ret2libc: If I understand it right, `r_anal_class_base_get_all` just returns the list of classes identified, not their relationships. On the other hand, `r_anal_class_get_inheritance_graph` as it is now just returns a "drawing graph", but no direct access to the class info. One would have to then do additional calls to get the classes info from RGraphNodeInfo. What's worse is that you would have to *assume* that, for example, the title of the node is the class name or something like that.

I'm very against RGraphNodeInfo in r_graph, yeah. The same as I don't see a RListNodeInfo in r_list or a RTreeNodeInfo in r_tree or something else in other places. All of these data structures may require some drawing, sure, but it is not the responsibility of r_graph(nor r_list, nor r_tree, etc.) to provide a particular way to draw the provided info (think about the RAGraph drawn with `VV`. You can see that in multiple ways by pressing p/P keys. If the logic of what to put in the body of the nodes would be in r_util/r_graph you would have to do a mess to change even a small thing). Those data structures should just provide the structure. In this case I think you have chosen a graph because there are relationships between classes like multiple parent/children. 

We may want to have a RGraphDrawable structure or something as a "subclass" of RGraph, but personally I would go that way only if there is more than one case where this is actually needed. If right now this is only used by the class usecase, it seems a bit overkill to me.

PR #17364, opened 2020-07-27
Add bytes to zb command ##signatures

swoops:

**Detailed description**

This adds the bytes signature into the the `zb` command to find the top `n` closest matching signatures. The bytes comparison is done by the Levenshtien distance. The Levenshtien algorithm is slow, so the difference in signature lengths is considered first to see if the Levenshtien algorythm is even necessary.

Example on known `strdup` and `fclose` functions in a statically compiled binary against signatures from the known libc version.
```
[0x0041e390]> z* ~ g cc=?
3149
[0x0041e390]> s sym.strdup 
[0x0041e390]> zb
1.00000  1.00000 B  1.00000 G   sym.__strdup
0.95000  0.91176 B  0.98824 G   sym.wcsdup
0.92754  0.86957 B  0.98551 G   sym.strndup
0.77465  0.55556 B  0.99375 G   sym.envz_remove
0.73906  0.51562 B  0.96250 G   sym.getlogin_r
[0x0041e390]> s sym.fclose 
[0x0040fc10]> zb
0.96032  0.92400 B  0.99664 G   sym.fclose
0.65971  0.35600 B  0.96342 G   sym._nl_expand_alias
0.65770  0.37800 B  0.93740 G   sym.fdopen
0.65112  0.35000 B  0.95225 G   sym.__run_exit_handlers
0.62532  0.34800 B  0.90264 G   sym.__cxa_finalize
```

The `zb` command on `strdup` completes near instantly. The same command on `fclose` takes about a second to complete. Notice that the byte score improves upon the graph score, increasing the distance between the correct answer and the second best answer.

You can see [this](https://github.com/swoops/r2_zb_stats/) repo for some quick and dirty statistics on how this can help when a perfect match fails. 

**Test plan**

See the update to the regression tests.

**Closing issues**

...

### DIFF
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -1261,8 +1261,7 @@ static bool closest_match_callback(void *a, const char *name, const char *value)
 			// bytes distance is slow. To avoid it, we can do quick maths to
 			// see if the highest possible score would be good enough to change
 			// results
-			double maxscore = maxscore = R_MIN (sizea, sizeb) /
-				R_MAX (sizea, sizeb);
+			double maxscore = R_MIN (sizea, sizeb) / R_MAX (sizea, sizeb);
 			if (div > 0) {
 				maxscore = (maxscore + score) / div;
 			}
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -1114,10 +1114,8 @@ R_API bool r_sign_delete(RAnal *a, const char *name) {
 }
 
 static ut8 * build_combined_bytes(RSignBytes *bsig) {
-	if (!bsig || !bsig->bytes || !bsig->mask) {
-		return NULL;
-	}
-	ut8 *buf = (ut8 *)malloc(bsig->size);
+	r_return_val_if_fail (bsig && bsig->bytes && bsig->mask, NULL);
+	ut8 *buf = (ut8 *)malloc (bsig->size);
 	if (!buf) {
 		return NULL;
 	}
@@ -1129,9 +1127,7 @@ static ut8 * build_combined_bytes(RSignBytes *bsig) {
 }
 
 static double cmp_bytesig_to_buff(RSignBytes *sig, ut8 *buf, int len) {
-	if (!sig || !buf || len < 0) {
-		return -1.0;
-	}
+	r_return_val_if_fail (sig && buf && len >= 0, (double)-1.0);
 	ut8 *sigbuf = build_combined_bytes (sig);
 	if (!sigbuf) {
 		return -1.0;
@@ -1320,22 +1316,15 @@ R_API void r_sign_close_match_free(RSignCloseMatch *match) {
 }
 
 R_API RList *r_sign_find_closest_sig(RAnal *a, RSignItem *it, int count, double score_threshold) {
-	r_return_val_if_fail (a && it && count > 0, NULL);
+	r_return_val_if_fail (a && it && count > 0 && score_threshold >= 0 && score_threshold <= 1, NULL);
 
 	// need at least one acceptable signature type
-	if (!it->bytes && !it->graph) {
-		return NULL;
-	}
+	r_return_val_if_fail (it->bytes || it->graph, NULL);
 
 	ClosestMatchData data;
 	RList *output = r_list_newf ((RListFree)r_sign_close_match_free);
 	if (!output) {
 		return NULL;
-	}
-
-	// impossible standard, return empty list
-	if (score_threshold > 1.0) {
-		return output;
 	}
 
 	data.output = output;
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -1113,6 +1113,36 @@ R_API bool r_sign_delete(RAnal *a, const char *name) {
 	return sdb_remove (a->sdb_zigns, k, 0);
 }
 
+static ut8 * build_combined_bytes(RSignBytes *bsig) {
+	if (!bsig || !bsig->bytes || !bsig->mask) {
+		return NULL;
+	}
+	ut8 *buf = (ut8 *)malloc(bsig->size);
+	if (!buf) {
+		return NULL;
+	}
+	int i;
+	for (i = 0; i < bsig->size; i++) {
+		buf[i] = bsig->bytes[i] & bsig->mask[i];
+	}
+	return buf;
+}
+
+static double cmp_bytesig_to_buff(RSignBytes *sig, ut8 *buf, int len) {
+	if (!sig || !buf || len < 0) {
+		return -1.0;
+	}
+	ut8 *sigbuf = build_combined_bytes (sig);
+	if (!sigbuf) {
+		return -1.0;
+	}
+
+	double sim = -1.0;
+	r_diff_buffers_distance (NULL, sigbuf, sig->size, buf, len, NULL, &sim);
+	free (sigbuf);
+	return sim;
+}
+
 static double matchBytes(RSignItem *a, RSignItem *b) {
 	double result = 0.0;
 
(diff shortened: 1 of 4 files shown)


ret2libc:
@swoops we have some tests about `r_diff_buffers_distance_levenstein` which seem to imply that algorithm is broken right now. You may be interested in having a look to avoid wrong results here as well (see https://github.com/radareorg/radare2/blob/master/test/unit/test_diff.c#L43 ).

swoops:
> @swoops we have some tests about r_diff_buffers_distance_levenstein which seem to imply that algorithm is broken right now. 

There seems to be two implementations of the same algorithm. One is indeed broken (when `diff->type = 'l'`). The other (`diff == NULL || diff->type = 'l'`) seems to work properly. This code relies on the implementation that works. The `zb match 5 by default` test will fail if you switch to the broken algorythm because it will place the fourth place signature above the third place (if I remember right).

libr/anal/sign.c
```diff
@@ -1113,21 +1113,51 @@ R_API bool r_sign_delete(RAnal *a, const char *name) {
...
+	}
+	int i;
+	for (i = 0; i < bsig->size; i++) {
+		buf[i] = bsig->bytes[i] & bsig->mask[i];
+	}
+	return buf;
+}
+
+static double cmp_bytesig_to_buff(RSignBytes *sig, ut8 *buf, int len) {
+	if (!sig || !buf || len < 0) {
```
ret2libc, line 1132: use `r_return_val_if_fail`. These checks actually should always be true, otherwise there is something very wrong and just ignoring it is not good because it hides the real problem (e.g. someone is passing wrong values to this function).

libr/anal/sign.c
```diff
@@ -1113,21 +1113,51 @@ R_API bool r_sign_delete(RAnal *a, const char *name) {
 	return sdb_remove (a->sdb_zigns, k, 0);
 }
 
+static ut8 * build_combined_bytes(RSignBytes *bsig) {
+	if (!bsig || !bsig->bytes || !bsig->mask) {
```
ret2libc, line 1117: `r_return_val_if_fail` here as well

libr/anal/sign.c
```diff
@@ -1205,75 +1236,126 @@ static bool closest_match_callback(void *a, const char *name, const char *value)
...
-	free (match->name);
-	free (match);
+	if (match) {
+		r_sign_item_free (match->item);
+		free (match);
+	}
 }
 
 R_API RList *r_sign_find_closest_sig(RAnal *a, RSignItem *it, int count, double score_threshold) {
-	r_return_val_if_fail (a && it && count > 0 && score_threshold <= 1 && score_threshold >= 0, false);
```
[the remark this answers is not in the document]
swoops: You are correct, currently `score_threshold` is always 0. I intend on adding variable in `e` to set this value to a nice default. The variable now, even though it can't be changed, to ensure the algorythm does not need a complicated change later.

I changed the `r_return_val_if_fail` to try and draw a better distinction between what is erroneous input and what is just an edge case. I guess if the caller is sending an unreasonable threshold it should probably emit a warning though. I will add score_threshold back into `r_return_val_if_fail`.

libr/anal/sign.c
```diff
@@ -1205,75 +1236,126 @@ static bool closest_match_callback(void *a, const char *name, const char *value)
...
+	r_return_val_if_fail (a && it && count > 0, NULL);
+
+	// need at least one acceptable signature type
+	if (!it->bytes && !it->graph) {
```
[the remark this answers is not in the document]
swoops: Yeah, this should probably be the responsibility of the caller.

libr/anal/sign.c
swoops: Sorry, I thought, for some reason, `r_return_val_if_fail` was used for `R_API` functions. I will fix these.

ret2libc: You can use it whenever you want to ensure there is no "programmatic error" (eg to check some conditions that you assume should always be true at some point of the program)

libr/anal/sign.c
```diff
@@ -1113,21 +1113,47 @@ R_API bool r_sign_delete(RAnal *a, const char *name) {
 	return sdb_remove (a->sdb_zigns, k, 0);
 }
 
+static ut8 * build_combined_bytes(RSignBytes *bsig) {
+	r_return_val_if_fail (bsig && bsig->bytes && bsig->mask, NULL);
+	ut8 *buf = (ut8 *)malloc (bsig->size);
+	if (!buf) {
+		return NULL;
+	}
+	int i;
```
trufae, line 1122: use size_t

libr/anal/sign.c
```diff
@@ -1113,21 +1113,47 @@ R_API bool r_sign_delete(RAnal *a, const char *name) {
 	return sdb_remove (a->sdb_zigns, k, 0);
 }
 
+static ut8 * build_combined_bytes(RSignBytes *bsig) {
+	r_return_val_if_fail (bsig && bsig->bytes && bsig->mask, NULL);
+	ut8 *buf = (ut8 *)malloc (bsig->size);
+	if (!buf) {
```
trufae, line 1119: use if (buf) { ... instead of having two return statements

libr/anal/sign.c
```diff
@@ -1113,21 +1113,47 @@ R_API bool r_sign_delete(RAnal *a, const char *name) {
...
+	for (i = 0; i < bsig->size; i++) {
+		buf[i] = bsig->bytes[i] & bsig->mask[i];
+	}
+	return buf;
+}
+
+static double cmp_bytesig_to_buff(RSignBytes *sig, ut8 *buf, int len) {
+	r_return_val_if_fail (sig && buf && len >= 0, (double)-1.0);
+	ut8 *sigbuf = build_combined_bytes (sig);
+	if (!sigbuf) {
```
trufae, line 1132: same

libr/core/cmd_zign.c
```diff
@@ -983,43 +983,74 @@ TODO: add useXRefs, useName
...
+		return false;
 	}
 
-	if (!r_sign_addto_item (core->anal, item, fcn, R_SIGN_GRAPH)) {
-		r_sign_item_free (item);
-		return NULL;
+	if (r_config_get_i (core->config, "zign.bytes")) {
+		r_sign_addto_item (core->anal, item, fcn, R_SIGN_BYTES);
+		RSignBytes *b = item->bytes;
+		int minsz = r_config_get_i (core->config, "zign.minsz");
```
trufae, line 1029: move this one line below

libr/core/cmd_zign.c
```diff
@@ -983,43 +983,74 @@ TODO: add useXRefs, useName
...
+		if (b && b->size < minsz) {
+			eprintf (
```
trufae, line 1031: just 1 line for this eprintf, not 3. Also there's a typo `to` must be `too`

libr/core/cmd_zign.c
```diff
@@ -983,43 +983,74 @@ TODO: add useXRefs, useName
...
+				"[!!] Function signature to small (%d < %d (zign.minsz))\n",
```
trufae, line 1032: dont nest parenthesis here. use a msg like:

`Warning: Function signature is too small (%d < %d) See e zign.minsz`

libr/core/cmd_zign.c
swoops: I am not sure what you mean here. I can't move `minsz` declaration inside of the `if` body because I use `minsz` in the condition. I could call `r_config_get_i` in the condition and then again in `eprintf` to avoid a variable since it only happens on the unlikely event that the function is too small.

Or were you talking about moving `RSignBytes *b` down below `minsz`?

PR #17365, opened 2020-07-27
Don't depend on flags when parsing RTTI ##anal

HoundThe:

**Detailed description**

Right now we're depending on flags to tell us what kind of __class_type_info the RTTI is, but from what I've seen that isn't ideal, because whenever we can't find the flag, we assume it's the most basic one and we could lose information about the classes. I've seen this in a binaries where the flag wasn't for deeper levels of inheritance (I am still not sure where the flags come from and their reasoning so I don't know why). So I've made a fallback option that if we can't find such a flag that would tell us, we just try to parse as much information we can. Based on the things we parse out we decide which class_type it is.

(It's very rough prototype, I'll refactor it if I know we want this)


libr/anal/rtti_itanium.c
```diff
@@ -439,41 +459,201 @@ static RTypeInfoType rtti_itanium_type_info_type_from_flag(RVTableContext *conte
...
 			return NULL;
 		}
 		follow -= 2 * context->word_size;
 		type = rtti_itanium_type_info_type_from_flag (context, follow);
 	}
 
 	if (type == R_TYPEINFO_TYPE_UNKNOWN) {
-		type = R_TYPEINFO_TYPE_CLASS;
-	}
+		/* RTTI
```
XVilka, line 561: I think this part should be a separate function

HoundThe:
If we'll wait to merge #17405 before this, then the feature workings will be visible on existing RTTI tests because the new parsing fallback is able to detect more of the base class inheritance in the binary

thestr4ng3r:
It's merged, please rebase on top of master.

HoundThe:
I've merged the master and yeah, now we detect

```
A
B: A
C: A
D: B, C
```
In the TestRTTI binaries, instead of just:
```
A
B
C
D: B, C
```
I've confirmed it from the sources of the binary by mrmacete, source
```cpp
#include "TestRTTI.hpp"

struct A {
  virtual void f ();
  virtual void g ();
  virtual void h ();
  int ia;
};
void A::f() { printf ("This is A::f()\n"); }
void A::g() { printf ("This is A::g()\n"); }
void A::h() { printf ("This is A::h()\n"); }

struct B: public virtual A {
  void f ();
  void h ();
  int ib;
};
void B::f() { printf ("This is B::f()\n"); }
void B::h() { printf ("This is B::h()\n"); }

struct C: public virtual A {
  void g ();
  void h ();
  int ic;
};
void C::g() { printf ("This is C::g()\n"); }
void C::h() { printf ("This is C::h()\n"); }

struct D: public B, public C {
  void h ();
  int id;
};
```
Tho the way Obj-C generates vtables is very unusual for me, still works tho

XVilka:
See also https://builds.sr.ht/~xvilka/job/275634#task-build-70 :
```
rtti_itanium.c:592:10: warning: enumeration values 'R_TYPEINFO_TYPE_UNKNOWN' and 'R_TYPEINFO_TYPE_CLASS' not handled in switch [-Wswitch]
        switch (cti->type) {
                ^
rtti_itanium.c:622:18: warning: incompatible pointer types assigning to 'RListFree' (aka 'void (*)(void *)') from 'void (class_type_info *)' (aka 'void (struct class_type_info_t *)') [-Wincompatible-pointer-types]
        rtti_list->free = rtti_itanium_type_info_free;
                        ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 warnings generated.
```

PR #17387, opened 2020-07-31
Fix #17379 - Avoid buffer overflow while identifying imports

### LINKED ISSUE #17379: [x86_64][ELF] Segmentation fault opening binary.
### Work environment

| OS/arch/bits (mandatory)                             | Arch Linux x86_64
| Architecture/bits of the file (mandatory)            | x86_64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.5.0-git 24750 @ linux-x86-64 git.4.4.0-247-ga934a3923
commit: a934a39236b3604e0e6ecebbee2b3b4d9c16bf06 build: 2020-06-09__09:28:44

### Expected behavior

`$ radare2 libpcl_segmentation.so.1.10.1` opens the file in radare2 and displays the r2 shell to the user.

### Actual behavior

```
$ radare2 libpcl_segmentation.so.1.10.1
r_uleb128: undefined behaviour in 70 shift on ut32
r_uleb128: undefined behaviour in 70 shift on ut32
r_uleb128: undefined behaviour in 77 shift on ut32
r_uleb128: undefined behaviour in 84 shift on ut32
r_uleb128: undefined behaviour in 91 shift on ut32
r_uleb128: undefined behaviour in 70 shift on ut32
r_uleb128: undefined behaviour in 70 shift on ut32
r_uleb128: undefined behaviour in 77 shift on ut32
r_uleb128: undefined behaviour in 70 shift on ut32
```

- Open the file. 
- I cannot drag and drop to github so the file is hosted temporarily [here](https://ufile.io/0xhfd2en). (Please download it and re-upload it before the link expires)

ret2libc:
  Not sure where the file comes from, can we add it to our tests? @SjRNMzU 

**Detailed description**

So, let's start by saying that I think the whole ELF symbols/imports retrieval is really really hacky and probably wrong in some cases, always because of the same issue that things are taken from the sections first and then "adjusted" with the info of phdr (it should be the opposite).

This problem was quite hard to debug, because ASAN/valgrind for some reason could not detect the issue at all, but there was buffer overflow. Function `_r_bin_elf_get_symbols_imports` has an initial loop where it computes `import_ret_ctr`, then it calls `fix_symbols` and then it allocates an array `import_ret` of size `import_ret_ctr`, copying elements from `ret`. However in some cases `fix_symbols` may add additional elements to `ret`, some of which apparently could be treated as imported symbols as well. This means that the original calculation done to compute `import_ret_ctr` may actually miss some symbols. When the loop at https://github.com/radareorg/radare2/blob/master/libr/bin/format/elf/elf.c#L3654 copies elements, it may actually copy more element than what originally allocated and overwrite data around in the heap.

The patch simply tries to consider this case, by incrementing `import_ret_ctr` as needed.

**Test plan**

...

**Closing issues**

Closes https://github.com/radareorg/radare2/issues/17379

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -3623,6 +3623,12 @@ static RBinElfSymbol* Elf_(_r_bin_elf_get_symbols_imports)(ELFOBJ *bin, int type
 	if (nsym == -1) {
 		goto beach;
 	}
+
+	// Elf_(fix_symbols) may find additional symbols, some of which could be
+	// imported symbols. Let's reserve additional space for them.
+	r_warn_if_fail (nsym >= ret_ctr);
+	import_ret_ctr += nsym - ret_ctr;
+
 	aux = ret;
 	while (!aux->last) {
 		if ((int)aux->ordinal > max) {


XVilka:
Probably worth cherry-picking to 4.5.1

PR #17396, opened 2020-08-03
Fix rabin2 -Pj and idpij (json PDB output format) ##bin

emuresan:

**Detailed description**

The information packed in the `pdb` field of the json, such as the `types` and `gvars` should be objects inside an array but, with recent changes in the json/pdb area, those were ommited.
The result was an invalid json when calling, for example `rabin2 -Pj some.pdb > some.pdb.json` or `idpij` in the radare shell.

**Test plan**

On the master branch, the issue can be reproduced as follows:
```
$ rabin2 -Pj ntdll.pdb | jq . 1>/dev/null
parse error: ':' not as part of an object at line 1, column 15
```
Also, the issue can easily be seen by inspecting the first items in the resulting json ( the missing `[{` before the `"types"` array):
```
{"pdb":"types":[{"type":"structure","name"
```

With the following fix, the resulting json is valid:
```
$ rabin2 -Pj ntdll.pdb | jq . 1>/dev/null ; echo $?
0
```
Also, it can be seen in the json as well:
```
{"pdb":[{"types":[{"type":"structure",
```

**Closing issues**

### DIFF
--- a/test/db/tools/rabin2
+++ b/test/db/tools/rabin2
@@ -1061,6 +1061,15 @@ EXPECT=<<EOF
 EOF
 RUN
 
+NAME=rabin2 -Pj
+FILE=bins/pdb/minimal.pdb
+CMDS=!rabin2 -Pj ${R2_FILE}
+EXPECT=<<EOF
+{"pdb":[{"types":[{"type":"structure","name":"struct_typedef","size":7,"members":[{"member_type":"char","member_name":"a","offset":0},{"member_type":"uint16_t","member_name":"b","offset":1},{"member_type":"int32_t","member_name":"c","offset":3}]},{"type":"structure","name":"<unnamed-tag>","size":4,"members":[{"member_type":"char[4]","member_name":"a","offset":0},{"member_type":"int32_t","member_name":"b","offset":0}]},{"type":"structure","name":"unnamed_member_types_typedef","size":8,"members":[{"member_type":"union <unnamed-tag>","member_name":"a","offset":0},{"member_type":"int32_t","member_name":"b","offset":4}]},{"type":"structure","name":"bitfield_typedef","size":1,"members":[{"member_type":"bitfield uint8_t : 1","member_name":"a","offset":0},{"member_type":"bitfield uint8_t : 2","member_name":"b","offset":0},{"member_type":"bitfield uint8_t : 3","member_name":"c","offset":0}]},{"type":"structure","name":"union_typedef","size":4,"members":[{"member_type":"char","member_name":"a","offset":0},{"member_type":"uint16_t","member_name":"b","offset":0},{"member_type":"int32_t","member_name":"c","offset":0}]}]},{"gvars":[{"address":12292,"symtype":0,"section_name":".data","gdata_name":"_uninitialized_global"},{"address":12288,"symtype":0,"section_name":".data","gdata_name":"_initialized_global"},{"address":4096,"symtype":2,"section_name":".text","gdata_name":"_function"},{"address":4182,"symtype":2,"section_name":".text","gdata_name":"_mainCRTStartup"},{"address":4120,"symtype":2,"section_name":".text","gdata_name":"_main"}]}]
+}
+EOF
+RUN
+
 NAME=rabin2 -S ""
 FILE=-
 CMDS=!rabin2 -S ""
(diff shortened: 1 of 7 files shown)


trufae:
PD. i have fixed the PR title

emuresan:
> Rabin2 edit ruins the json validity, but the json is valid for `idpij`, your PR fixes the Rabin2, but creates invalid json for idpij

The issue is a little bit tricky since `r_main_rabin2` ,via the `run_action` macro, expects a valid json from `r_core_pdb_info` ( with heading and trailing brackets ), while `cmd_info` expects an invalid json ( without heading and trailing brackets ). That being said, it's harder than it seems to fix one without breaking the other.

XVilka:
I suggest just remove "pdb" part

emuresan:
> I suggest just remove "pdb" part

That will fix the apparent issue, but there's better ways to do it imho.
For one, "removing the \"pdb\" part" sugests neglecting the existence of the `run_action` interface in `r_main_rabin2` which will lead to further inconsistency.
And again, having an api that has a json output and intentionally expecting that json to be invalid seems wrong.
Fixing the issue in this manner seems like we're simply paving the way for future bugs.

trufae (approved):
lgtm, we can further improve this later in a separate pr

ret2libc:
Hi @emuresan , is there any news on this? I see there were some doubts around the expected output. I think a test in this case would help a lot for avoiding future issues, for making sure your PR works as intended and for ensuring what is intended it is actually what everybody expects as well :) I think you can easily add a new test in `test/db/tools/rabin2`. Let us know if you need anything.

trufae:
Some tests are not passing

emuresan:
Yeah, the tests I added failed because they are relying on the `minimal.pdb` file that I added here: https://github.com/radareorg/radare2-testbins/pull/50 . A rerun of the tests should hopefully have them passing.

ret2libc:
Hi @emuresan , as @HoundThe I have some doubts about the resulting json with this change. Is there a reason to put things in an array?

I'm no windows/PDB expert at all, so I may be missing some important details, but I would expect something like:
```
> idpij ./test.pdb
{
   "types": [.....],
   "gvars": [....],
}
```
and for the rabin2 case something like:
```
$ rabin2 -Pj ./test.pdb
{
   "pdb": {
      "types": [....],
      "gvars": [....],
   }
}
```
Does this make sense?

Basically, I would expect we keep the object of idpij and we wrap it around "pdb", which I guess was the original intent of the code.

emuresan:
Well, the original code had all of those in arrays (I'm talking about the ages old build that I had on my machine before I upraded to latest master and spotted the bug).
The reason why it also makes sense to have all the information in array entries is because of the structure of the pdb files. Basically the file is grouped in streams and each stream will hold different data: https://llvm.org/docs/PDB/index.html .
Now, we don't necessarly care about said streams or data in the json output since only care about types and gvars so I can change it if you'd like.

ret2libc:
@HoundThe do you have an opinion on this? array vs object? In any case, however, @emuresan a valid JSON is better than an invalid one :D So whatever we choose it must be valid. Right not, it should be very easy to make it valid json in all cases (rabin2 and r2) by keeping the object json structure.

trufae:
@ret2libc i think this is good to merge

ret2libc:
> @ret2libc i think this is good to merge

I'm not sure actually.. I still don't understand why the structure is:
```json
"pdb": [
   {
      "types": [ ... ]
   },
   {
      "gvars": [ ... ]
   }
]
```
instead of:
```json
"pdb": {
      "types": [ ... ]
      "gvars": [ ... ]
},
```

trufae:
uhm yeah. that shouldnt be like this

trufae:
ping @emuresan , can you change the code to construct the proposed structure in https://github.com/radareorg/radare2/pull/17396#issuecomment-698172821

test/db/tools/rabin2
```diff
@@ -1045,6 +1045,15 @@ EXPECT=<<EOF
 EOF
 RUN
 
+NAME=rabin2 -Pj
+FILE=bins/pdb/minimal.pdb
+CMDS=!rabin2 -Pj ${R2_FILE}
```
trufae, line 1050: append `~j` to get the JSON indent output, its easier to read and check the diff in case the output changes.

trufae (changes requested):
Please see @ret2libc comment. just removing the pj_o pj_end calls will fix the issue i guess. in which situations is the output incorrect without those calls?

trufae:
Ping

trufae (approved):

PR #17397, opened 2020-08-03
Update docs in annotate code API 

NirmalManoj:

**Detailed description**
Earlier, I had used kernel-docs documentation style for documenting new API functions created for the annotations in the decompiler. This PR replaces these documentations by Doxygen. 

...

**Test plan**
1. Make sure this compiles properly.
2. Read the documentation and make sure nothing is wrong.

...

**Closing issues**

...

### DIFF
--- a/libr/include/r_core.h
+++ b/libr/include/r_core.h
@@ -927,28 +927,31 @@ extern RCorePlugin r_core_plugin_a2f;
 
 /* DECOMPILER PRINTING FUNCTIONS */
 /**
- * The function will print the output in console using the function r_cons_printf();
  * @brief Prints the data contained in the specified RAnnotatedCode in JSON format.
+ * 
+ * The function will print the output in console using the function r_cons_printf();
  * 
  * @param code Pointer to a RAnnotatedCode.
  */
 R_API void r_core_annotated_code_print_json(RAnnotatedCode *code);
 /**
+ * @brief Prints the decompiled code from the specified RAnnotatedCode.
+ * 
  * This function is used for printing the output of commands pdg and pdgo.
  * It can print the decompiled code with or without offsets. If line_offsets is a null pointer,
  * the output will be printed without offsets (pdg), otherwise, the output will be
  * printed with offsets.
  * This function will print the output in console using the function r_cons_printf();
- * @brief Prints the decompiled code from the specified RAnnotatedCode.
  * 
  * @param code Pointer to a RAnnotatedCode.
  * @param line_offsets Pointer to a RVector that contains offsets for the decompiled code.
  */
 R_API void r_core_annotated_code_print(RAnnotatedCode *code, RVector *line_offsets);
 /**
+ * @brief  Prints the decompiled code as comments
+ * 
  * This function is used for the output of command pdg*
  * Output will be printed in console using the function r_cons_printf();
- * @brief  Prints the decompiled code as comments
  * 
  * @param code Pointer to a RAnnotatedCode.
  */
--- a/libr/include/r_core.h
+++ b/libr/include/r_core.h
@@ -944,7 +944,7 @@ R_API void r_core_annotated_code_print_json(RAnnotatedCode *code);
  * This function will print the output in console using the function r_cons_printf();
  * 
  * @param code Pointer to a RAnnotatedCode.
- * @param line_offsets Pointer to a RVector that contains offsets for the decompiled code.
+ * @param line_offsets Pointer to a @ref RVector that contains offsets for the decompiled code.
  */
 R_API void r_core_annotated_code_print(RAnnotatedCode *code, RVector *line_offsets);
 /**
--- a/libr/include/r_core.h
+++ b/libr/include/r_core.h
@@ -927,9 +927,8 @@ extern RCorePlugin r_core_plugin_a2f;
 
 /* DECOMPILER PRINTING FUNCTIONS */
 /**
- * Prints the data contained in RAnnotatedCode represented by the pointer 'code' in JSON format. 
  * The function will print the output in console using the function r_cons_printf();
- * @brief Prints the data contained in RAnnotatedCode *code in JSON format.
+ * @brief Prints the data contained in the specified RAnnotatedCode in JSON format.
  * 
  * @param code Pointer to a RAnnotatedCode.
  */
@@ -940,14 +939,13 @@ R_API void r_core_annotated_code_print_json(RAnnotatedCode *code);
  * the output will be printed without offsets (pdg), otherwise, the output will be
  * printed with offsets.
  * This function will print the output in console using the function r_cons_printf();
- * @brief Prints the decompiled code in the passed argument 'code'.
+ * @brief Prints the decompiled code from the specified RAnnotatedCode.
  * 
  * @param code Pointer to a RAnnotatedCode.
- * @param line_offsets Pointer to a RVector that containes offsets for the decompiled code.
+ * @param line_offsets Pointer to a RVector that contains offsets for the decompiled code.
  */
 R_API void r_core_annotated_code_print(RAnnotatedCode *code, RVector *line_offsets);
 /**
- * This functions prints the decompiled code as comment.
  * This function is used for the output of command pdg*
  * Output will be printed in console using the function r_cons_printf();
  * @brief  Prints the decompiled code as comments
--- a/libr/include/r_core.h
+++ b/libr/include/r_core.h
@@ -927,39 +927,33 @@ extern RCorePlugin r_core_plugin_a2f;
 
 /* DECOMPILER PRINTING FUNCTIONS */
 /**
-* r_core_annotated_code_print_json() - Prints the data contained in RAnnotatedCode *code in JSON format.
-* @code: Pointer to a RAnnotatedCode
-*
-* Prints the data contained in RAnnotatedCode represented by the pointer 'code' in JSON format. 
-* The function will print the output in console using the function r_cons_printf();
-* 	
-* Return: Nothing
-*/
+ * Prints the data contained in RAnnotatedCode represented by the pointer 'code' in JSON format. 
+ * The function will print the output in console using the function r_cons_printf();
+ * @brief Prints the data contained in RAnnotatedCode *code in JSON format.
+ * 
+ * @param code Pointer to a RAnnotatedCode.
+ */
 R_API void r_core_annotated_code_print_json(RAnnotatedCode *code);
 /**
-* r_core_annotated_code_print() - Prints the decompiled code in the passed argument 'code'.
-* @code: Pointer to a RAnnotatedCode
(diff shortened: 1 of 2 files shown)


libr/include/r_util/r_annotated_code.h
```diff
@@ -79,105 +79,92 @@ typedef struct r_annotated_code_t {
...
+ * @return Pointer to the RPVector created.
  */
 R_API RPVector *r_annotated_code_annotations_in(RAnnotatedCode *code, size_t offset);
 /**
- * r_annotated_code_annotations_range() - Returns all annotations with range that overlap with the given range.
- * @code: Pointer to a RAnnotatedCode.
- * @start: Start of the range(inclusive).
- * @end: End of the range(exclusive).
- *
  * Creates an RPVector and inserts the pointers to all annotations whose
```
karliss, line 150: In comments don't describe how, describe what the code does. "Creates an RPVector and inserts pointers" is more of literal description of code written. The brief part should be sufficient. in this case. Range endpoint behavior is already described in parameter description.

libr/include/r_util/r_annotated_code.h
```diff
@@ -79,105 +79,92 @@ typedef struct r_annotated_code_t {
...
  */
 R_API bool r_annotation_is_variable(RCodeAnnotation *annotation);
 /**
- * r_annotated_code_add_annotation() - Inserts *annotation in *code.
- * @code: Pointer to a RAnnotatedCode.
- * @annotation: Pointer to a annotation.
- *
  * This functions inserts the annotation represented by the pointer 'annotation' to the vector
  * of annotations in the RAnnotatedCode represented by 'code'. To be more precise,
  * annotation will be added to code->annotations, which is a RVector<RCodeAnnotation> annotations.
```
karliss, line 132: There is no need to go into implementation details like "added to code->annotations" which is RVector<RCodeAnnotation>" unless it affects the way this function used. The whole point of this function is to hide implementation detail like this so that it in future it can be easily replaced with different container or multiple of them if necessary to answer queries about annotated code. For example it might replaced with a specialized tree structure which allows implementing `r_annotated_code_annotations_in` more efficiently.

libr/include/r_core.h
```diff
@@ -927,39 +927,33 @@ extern RCorePlugin r_core_plugin_a2f;
...
-* @code: Pointer to a RAnnotatedCode
-*
-* Prints the data contained in RAnnotatedCode represented by the pointer 'code' in JSON format. 
-* The function will print the output in console using the function r_cons_printf();
-* 	
-* Return: Nothing
-*/
+ * Prints the data contained in RAnnotatedCode represented by the pointer 'code' in JSON format. 
+ * The function will print the output in console using the function r_cons_printf();
+ * @brief Prints the data contained in RAnnotatedCode *code in JSON format.
```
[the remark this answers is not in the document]
NirmalManoj: I have removed the parts that say the same thing twice. I think it will be better if we keep the detailed description first and then have brief description and information regarding `parameters` and `return`.  This way, we can see the brief and parameters together. I think a detailed description won't be required every time to understand what the function does, so this is probably more readable. Also, this style is followed by many functions in other places.

libr/include/r_util/r_annotated_code.h
```diff
@@ -79,105 +79,92 @@ typedef struct r_annotated_code_t {
...
- *
  * Creates an RVector and inserts the offsets for every seperate line of decompiled code in
  * code->code (code->code is a character array).
  * If a line of decompiled code doesn't have a unique offset, UT64_MAX is inserted as its offset.
- * 	
- * Return: Pointer to the RVector created.
+ * @brief Returns the offset for every line of decompiled code in RAnnotatedCode *code.
+ * 
+ * @param code Pointer to a RAnnotatedCode.
+ * @return Pointer to the RVector created.
```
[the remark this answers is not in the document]
NirmalManoj: Done.

libr/include/r_util/r_annotated_code.h
```diff
@@ -79,105 +79,92 @@ typedef struct r_annotated_code_t {
...
- *
  * Creates an RPVector and inserts the pointers to all annotations whose 
  * range overlap with range [start, end-1] (both inclusive).
+ * @brief Returns all annotations with range that overlap with the given range.
  * 
- * Return: Pointer to the RPVecrtor created.
+ * @param code Pointer to a RAnnotatedCode.
+ * @param start Start of the range(inclusive).
+ * @param end End of the range(exclusive).
+ * @return Pointer to the RPVecrtor created.
```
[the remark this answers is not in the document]
NirmalManoj: Done.

libr/include/r_util/r_annotated_code.h
NirmalManoj: Thanks, I have updated this.

NirmalManoj:
> It would probably be more suitable to document each enum value inline.

Done.

NirmalManoj:
> It would be good to document the inclusiveness or exclusiveness r_code_annotation_t start and end positions.

Done.

libr/include/r_util/r_annotated_code.h
```diff
@@ -21,44 +21,45 @@ typedef enum r_syntax_highlight_type_t {
...
+	R_CODE_ANNOTATION_TYPE_FUNCTION_NAME, /*!< Specified range in annotation represents a function name. */ 
+	R_CODE_ANNOTATION_TYPE_GLOBAL_VARIABLE, /*!< Specified range in annotation represents a global variable. */ 
+	R_CODE_ANNOTATION_TYPE_CONSTANT_VARIABLE, /*!< Specified range in annotation represents a constant variable with an address. */ 
+	R_CODE_ANNOTATION_TYPE_LOCAL_VARIABLE, /*!< Specified range in annotation represents a local variable. */ 
+	R_CODE_ANNOTATION_TYPE_FUNCTION_PARAMETER, /*!< Specified range in annotation represents a function parameter. */ 
 	// ...
 } RCodeAnnotationType;
 
+/**
+ * \struct r_code_annotation_t r_annotated_code.h libr/include/r_util/r_annotated_code.h
```
karliss, line 37: \struct tag is only necessary in case of detached documentation. It is unnecessary when the comment is immediately before the thing it documents.

libr/include/r_util/r_annotated_code.h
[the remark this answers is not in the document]
karliss: Information like this is something that you could have included in the `@return` description, but current approach is also fine.

libr/include/r_util/r_annotated_code.h
```diff
@@ -74,110 +75,99 @@ typedef struct r_code_annotation_t {
...
- * @e: Pointer to the annotation
- * @user: Always NULL for this function. Present here for this function to be of the type RVectorFree.
- * 
  * This function recognizes the type of the specified annotation and
  * frees memory that is dynamically allocated for it.
+ * @brief Deallocates dynamically allocated memory for the specified annotation.
  * 
- * Return: Nothing.
+ * @param e Pointer to the annotation.
+ * @param user Always NULL for this function. Present here for this function to be of the type RVectorFree.
```
karliss, line 108: When mentioning other types like `RVectorFree` you can mark them using `@ref` like this `@ref RVectorFree`. It will cause doxygen to insert a link to the documentation of corresponding element.

Screenshot from 2020-08-04 12-06-13
Screenshot from 2020-08-04 12-12-39

libr/include/r_util/r_annotated_code.h
NirmalManoj: Removed struct tag.

libr/include/r_core.h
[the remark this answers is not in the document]
karliss: There are two arguments for having brief before more detailed description:
* That's the order Doxygen and other tools will display it. `By default the brief descriptions become the first sentence of the detailed descriptions`.  Moving it to the end means that it becomes less readable as single text.
* Typical way of marking end of brief and start of detailed description is using an empty line. Some tools like like VSCode get confused if it isn't in this order and ignores everything before brief.

> Also, this style is followed by many functions in other places.

By searching for `[@\\]brief` I was able to find only one such function in r2 which wasn't just written by you - `parse_attr_value`. And even in that case it was less of detailed description more of TODO comment noting that function is unfinished.

libr/include/r_util/r_annotated_code.h
```diff
@@ -21,44 +21,43 @@ typedef enum r_syntax_highlight_type_t {
...
- *
- * There are two kinds of RCodeAnnotation. One for offset, which of the type 
- * R_CODE_ANNOTATION_TYPE_OFFSET and other one is for syntax highlight, which is
- * of the type R_CODE_ANNOTATION_TYPE_SYNTAX_HIGHLIGHT.
- * R_CODE_ANNOTATION_TYPE_OFFSET is for representing annotations that gives an offset for 
- * a range while R_CODE_ANNOTATION_TYPE_SYNTAX_HIGHLIGHT is for representing the
- * kind of data the range represents. Here, range refers to the range of annotation.
- */
-
+//! Represents the type of annnotation.
```
karliss, line 24: In most cases Doxygen treats `//!` `///` `/**` and `/*!` almost identically so it's a question of chosen style. Whatever style is chosen use it consistently. Everywhere else in this file you used `/**` version so you should use the same for this enum instead of `//!` or `/*!`.

libr/include/r_core.h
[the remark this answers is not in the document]
NirmalManoj: @karliss I saw such examples somewhere and was also sure that I have seen it in radare2. My mistake I said many! Now I understand the problems with this and I will change it as you suggested.

libr/include/r_util/r_annotated_code.h
NirmalManoj: Done.

libr/include/r_util/r_annotated_code.h
NirmalManoj: Done.

libr/include/r_core.h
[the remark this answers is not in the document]
NirmalManoj: Done.

NirmalManoj:
> Some bad examples:
> 
> * `RCodeAnnotationType type; /**< Type of the annotation. */` - who could have guessed that code annnotation type named type is type of the annotation.
> * `R_CODE_ANNOTATION_TYPE_FUNCTION_PARAMETER, /*!< Specified range in annotation represents a function parameter. */`

@karliss I'm not sure what all are bad examples. The second one here doesn't seem useless to me. If it is useless, I don't see how any documentation for the `enum r_code_annotation_type_t` is useful.

karliss:
>  If it is useless, I don't see how any documentation for the enum r_code_annotation_type_t is useful.

Yes most of the descriptions for r_code_annotation_type_t don't provide any information. I think one or two had unique words providing little bit of additional information.

Issue #17383, opened 2020-07-31
[ELF] Segmentation fault by opening a binary (Bug in DWARF parsing)


S01den:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | 5.6.16-1-Manjaro x86_64, Kubuntu x86 32
| File format of the file you reverse (mandatory)      |  ELF
| Architecture/bits of the file (mandatory)            |  x86/32, x86/64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.6.0-git 25006 @ linux-x86-64 git.4.4.0-481-geac93216e
commit: eac93216ecb0819dd05264e44347b2bbb931e7d9 build: 2020-07-26__21:25:02

### Expected behavior

`radare2 ./test_crash` opens the file in radare2 and displays the r2 shell to the user.

### Actual behavior

`$ r2 test_crash `
`Segmentation fault (core dumped)`

### Steps to reproduce the behavior 
We, Architect (@CitadelArcho) and me, discovered this bug and dug a bit into it.
It is caused by malformed DWARF information (DW_AT_name), in the .debug_info section.
So we wrote a small PoC script which turns any ELF into a binary which makes radare2 crash.

```Python
#!/usr/bin/python3
from elftools.elf.elffile import ELFFile
from elftools.elf.enums import ENUM_E_MACHINE
import sys
import struct
import argparse
import os
import base64

# trigger a segfault in radare2 by modifing a DW_FORM_strp (a reference to a string in the dwarf debug format) (modify the shift in DW_AT_name)
# bug found by S01den and Architect (with custom fuzzing)

def build_parser():
	parser = argparse.ArgumentParser(description="Trigger a segfault in radare2 by modifing a DW_FORM_strp in .debug_info")
	parser.add_argument("-f", "--file", 
		type=str, default="main",
		help="select the file to patch")

	return parser

print("__________                       _____  ________   _____  _________                      .__     ")
print("\______   \_______  ____   _____/ ____\ \_____  \_/ ____\ \_   ___ \____________    _____|  |__  ")
print("|     ___/\_  __ \/  _ \ /  _ \   __\   /   |   \   __\  /    \  \/\_  __ \__  \  /  ___/  |  \  ")
print("|    |     |  | \(  <_> |  <_> )  |    /    |    \  |    \     \____|  | \// __ \_\___ \|   Y  \ ")
[...]

PR #17399, opened 2020-08-03
Fix malformed DWARF crash due invalid .debug_str reference ##bin

HoundThe:

**Detailed description**

Fix for #17383 

In case the string pointer into debug_str isn't valid, the string can be NULL which wasn't accounted for in the future type parsing, I've added the handling for this situation.


XVilka:
Btw, you might want to create a macros or a small inline function for this pattern:
```c
		if (!value->string.content) {
				name = create_type_name_from_offset (die->offset);
			} else {
				name = strdup (value->string.content);
			}
			if (!name) {
				goto cleanup;
			}
```

PR #17411, opened 2020-08-06
r2pm: Allow package script to declare cleaning is done

kazarmy:

**Detailed description**

This pr allows r2pm package scripts to do their own cleaning of their git dir.

**Test plan**

Add the following to the r2dec package script before R2PM_END:

```sh
if [ "$ACTION" = clean ]; then
	rm -rf "${R2PM_GITDIR}/r2dec-js"
	DONE=1
fi
```

and then run `r2pm -ci r2dec`. There should be no error message stating that the r2dec git dir cannot be found.

This is currently necessary for cleaning because the git dirname (`r2dec-js`) is hardcoded in the r2dec source.

**Closing issues**

...

### DIFF
--- a/binr/r2pm/r2pm
+++ b/binr/r2pm/r2pm
@@ -573,25 +573,26 @@ case "$1" in
 			if [ -f "${FILE}" ]; then
 				NAME="$2"
 				ACTION=clean
-				DONE=0
 				export PATH="${R2PM_HOMEBINDIR}:${PATH}"
 				. "${FILE}"
+				SKIP_GITCLEAN=$?
 				echo $FILE TGZ=$R2PM_TGZ
-				[ "${DONE}" = 1 ] && exit 0
 			else
 				echo "Cannot find $FILE"
 				RC=1
 			fi
 
-			if [ -d "${R2PM_GITDIR}/$2" ]; then
-				echo "Cleaning up ${R2PM_GITDIR}/$2..."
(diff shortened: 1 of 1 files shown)


trufae:
What about returning a value and checking for error code instead of using the arbitrary DONE var that doesnt seems to specify anything related to cleaning or r2pm? Reading the pkg script doesnt makes it clear what the DONE var will do

kazarmy:
Checking the return value can be done (b69d59393fd6fcb62963bfca3a7dd7713cf0898c with
```sh
if [ "$ACTION" = clean ]; then
	rm -rf "${R2PM_GITDIR}/r2dec-js"
	R2PM_END
	return 1
fi
```
) but is somewhat awkward (also, the `echo` in R2PM_END is not guaranteed to return 0) and I think renaming the DONE var to perhaps CLEAN_GITDIR is better.

trufae:
Yes. Using a meaningful var is more readable. Maybe starting it with R2PM_ for consistency like all the other r2pm functions

kazarmy:
> Yes. Using a meaningful var is more readable. Maybe starting it with R2PM_ for consistency like all the other r2pm functions

Ok bbc1bafedeaf7afc1ff8f1dca1c6cc1ab902e4d9 (and

```sh
if [ "$ACTION" = clean ]; then
	rm -rf "${R2PM_GITDIR}/r2dec-js"
	R2PM_CLEAN_GITDIR=0
fi
```
)

PR #17421, opened 2020-08-08
Do not use 0b... integer constants

XVilka:
The form of 0b01101 appeared only in GCC 4.3.
It wasn't defined in the C standard and was only
an extension.

 

**Detailed description**

The binary literals in form of `0b01101` appeared only in GCC 4.3 and was a non-standard C extension. Switching to hexadecimal constants will give the ability to be compiled C99-compatible older GCC compilers.


wargio:
you can merge this once is green

PR #17423, opened 2020-08-08
r2pm: R2PM_CLEAN_GITDIR states (sub)dir to clean

kazarmy:

**Detailed description**

This pr changes `R2PM_CLEAN_GITDIR` so that it states the git (sub)dir to clean. I think this is better than #17411 because more r2pm code can be reused.

Supersedes #17411. Reverts #17357. Also removes the trailing `/` in `R2PM_GITDIR`.

This git (sub)dir can't be derived directly from `R2PM_GIT` because many plugins use:

```sh
R2PM_GIT "https://github.com/radareorg/radare2-extras"
```

**Test plan**

Add the following anywhere in the `r2dec` package script:

```sh
R2PM_CLEAN_GITDIR=r2dec-js
```

**Closing issues**

...

### DIFF
--- a/binr/r2pm/r2pm
+++ b/binr/r2pm/r2pm
@@ -98,7 +98,7 @@ R2PM_ETCD="${R2CONFIGHOME}/radare2rc.d"
 
 # TODO. support system plugin installs R2PM_PLUGDIR="${R2PM_SYSLIBDIR}/radare2/last"
 if [ -z "${R2PM_GITDIR}" ]; then
-	R2PM_GITDIR="${R2PM_USRDIR}/git/"
+	R2PM_GITDIR="${R2PM_USRDIR}/git"
 fi
 if [ -z "${R2PM_DBDIR}" ]; then
 	if [ -d "${R2DATAHOME}/r2pm/db" ]; then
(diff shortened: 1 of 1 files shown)


trufae:
Why not just pass the 2nd argument to the git url?

kazarmy:
> Why not just pass the 2nd argument to the git url?

Because if a 2nd argument is passed, it changes where the package is installed, and at least r2dec doesn't like that (i.e. it explicitly tries to find a `r2dec-js` folder).

Now you may argue that r2dec should change from using `r2dec-js` to using `r2dec`, and ideally that should be done, but fixing something that ain't broken doesn't seem very appealing (not to mention it requires juggling 2 prs and also might require cooperation from the package owner).

kazarmy:
requires juggling 2 prs -> requires the simultaneous merging of 2 different prs

kazarmy:
> I think this is better than #17411 because more r2pm code can be reused.

Hmm another way of doing that is to refactor out the gitdir-cleaning code into a separate function and then call it from inside an `if [ "$ACTION" = clean ]` block. Probably more proper than having yet another global var, so closing this.

PR #17430, opened 2020-08-09
Implement save and load functions for debug session ##debug

abcSup:

**Detailed description**

* Add dtst and dtsf commands
* Change Rangstr to unsigned int to increase json length limit

Store `RDebugSession` in sdb and save it to disk in the following format:
```
maxcnum=<maxcnum>

reglist=<addr>,<addr>,...
registers={"<addr>": {"size":<size_t>, "a":[<RDebugChangeReg>]}, ...}

memorylist=<addr>,<addr>,...
memory={"<addr>": {"size":<size_t>, "a":[<RDebugChangeMem>]}, ...}

chkptlist=<cnum>,<cnum>,...
checkpoints={
  "<cnum>": {
        registers: {"<RRegisterType>": <RRegArena>, ...},
        snaps: {"size":<size_t>, "a":[<RDebugSnap>]},
  },
  ...
}

RDebugChangeReg JSON:
{"cnum":<int>, "data":<ut64>}

RDebugChangeMem JSON:
{"cnum":<int>, "data":<ut8>}

RRegArena JSON:
{"size":<int>, "bytes":"<base64>"}

RDebugSnap JSON:
{"name":<str>, "addr":<ut64>, "addr_end":<ut64>, "size":<ut64>,
 "data":"<base64>", "perm":<int>, "user":<int>, "shared":<bool>}
```
The serialization style mostly follows the WIP `r2db` for r2 project. This PR uses the `sdb_json` json parser and a list of keys (`reglist`, `memorylist`, `chkptlist`) to ease iteration over dictionary (`registers`, `memory`, `checkpoints`).

**Test plan**

I have added the test case `dbg_dts` to test saving, loading and debugging after loading in one session.

### DIFF
--- a/libr/debug/dsession.c
+++ b/libr/debug/dsession.c
@@ -2,6 +2,10 @@
 
 #include <r_debug.h>
 
+#define CMP_CNUM_REG(x, y) ((x) >= ((RDebugChangeReg *)y)->cnum ? 1 : -1)
+#define CMP_CNUM_MEM(x, y) ((x) >= ((RDebugChangeMem *)y)->cnum ? 1 : -1)
+#define CMP_CNUM_CHKPT(x, y) ((x) >= ((RDebugCheckpoint *)y)->cnum ? 1 : -1)
+
 R_API void r_debug_session_free(RDebugSession *session) {
 	if (session) {
 		r_vector_free (session->checkpoints);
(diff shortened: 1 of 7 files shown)


XVilka:
@thestr4ng3r please take a look to the serialization part since you're the creator of `r2db`.

libr/debug/dsession.c
```diff
@@ -231,3 +235,375 @@ R_API bool r_debug_session_add_mem_change(RDebugSession *session, ut64 addr, ut8
...
+ *
+ *   memorylist=<addr>,<addr>,...
+ *   memory={"<addr>": {"size":<size_t>, "a":[<RDebugChangeMem>]}, ...}
+ *
+ *   chkptlist=<cnum>,<cnum>,...
+ *   checkpoints={
+ *     "<cnum>": {
+ *       registers: {"<RRegisterType>": <RRegArena>, ...},
+ *       snaps: {"size":<size_t>, "a":[<RDebugSnap>]},
+ *     },
```
thestr4ng3r, line 400: Having such a big dictionary in a single sdb value can become an issue. SDB by design has a maximum value size of 16MB (please correct me if I'm wrong @trufae) and I would expect these checkpoints to commonly exceed that. The same might be the case for `registers` and `memorylist`.
What I would do here is to use one `sdb_ns` for each of these lists and use the sdb structure of the sub-namespaces as the dict. Instead of `chkptlist` you can then just list all the keys of the respective namespace.

libr/debug/dsession.c
```diff
@@ -231,3 +235,375 @@ R_API bool r_debug_session_add_mem_change(RDebugSession *session, ut64 addr, ut8
...
+		RDebugCheckpoint checkpoint = { 0 };
+		checkpoint.cnum = (int)sdb_atoi (cnum);
+
+		// Extract RRegArena's from "registers"
+		for (i = 0; i < R_REG_TYPE_LAST; i++) {
+			int size = _json_num_get (db, "checkpoints", REGATTR (size), 0);
+			if (size == 0) {
+				continue;
+			}
+			char *edata = sdb_json_get (db, "checkpoints", REGATTR (bytes), 0);
```
thestr4ng3r, line 540: `sdb_json_get` is extremely inefficient, it re-parses the entire json on every call so here you would have O(n^2) parsing. After the format change mentioned above, it will probably be O(n) so as a temporary solution `sdb_json` might be ok until nxjson is available in r2.

thestr4ng3r:
When the format is finalized, please also add extensive unit test in the style of for example https://github.com/thestr4ng3r/r2db/blob/master/test/unit/test_flag.c

XVilka:
Have you considered to use it for ESIL-based session recording as well? This way it can be better tested without invoking debugger.

abcSup:
> Have you considered to use it for ESIL-based session recording as well? This way it can be better tested without invoking debugger.

The session format here could support it after I added support for ESIL tracing in a debug session, which could be done since ESIL already provides API to access information regarding reg/mem read/write.

libr/debug/dsession.c
abcSup: > Having such a big dictionary in a single sdb value can become an issue. SDB by design has a maximum value size of 16MB (please correct me if I'm wrong @trufae) and I would expect these checkpoints to commonly exceed that. The same might be the case for registers and memorylist.

You are right. I should just dump the memory snapshots into their own files rather than saving them in base64 and store them in `sdb`. Now, I see the reason why `Rangstr` is `unsigned short` which could be the max limit you mentioned. 

> What I would do here is to use one sdb_ns for each of these lists and use the sdb structure of the sub-namespaces as the dict. Instead of chkptlist you can then just list all the keys of the respective namespace.

That sounds better and I will just store them separately into its own db like how `r2db` does. I could wait for `r2db`, so I do not have to re-implement  the sdb archiving functions.

thestr4ng3r: It still makes sense to extend Rangstr since `unsigned short` has a max of 64k instead of 16MB. I think it depends if the memory snapshot base64s themselves can exceed 16MB whether it makes sense to dump them into separate binary files outside sdb. If there is only one per sdb key it might work out.

abcSup: Dumping them into their own binaries could save time for encoding and decoding them in base64. Besides, the only advantage of storing them in sdb is having everything in one place. I will just keep them in sdb for now.

abcSup: For now, I will just save each of them into its own `sdb` files. It would be great to have a function such as `sdb_archive_save` in `r2db` to store everything in one archive. Are you planning to add that in future?

XVilka:
```
[XX] db/tools/r2 use void on functions without parameters
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '"!git grep -e ^R_API -e ^static ../libr | grep -e '[a-z]() {' -e '[a-z]();'"' -
-- stdout
@@ -0,0 +1,2 @@
+../libr/debug/dsession.c:R_API RDebugSession *r_debug_session_new() {
+../libr/include/r_debug.h:R_API RDebugSession *r_debug_session_new();
```

thestr4ng3r:
r_json is now merged so it can also be used for this.

XVilka:
Don't forget please to update the R2 book chapter after these changes btw: https://github.com/radareorg/radare2book/blob/master/src/debugger/revdebug.md

yossizap (approved):
LGTM. Will open an issue for the `dtst` suggestion.

Issue #3607, opened 2015-10-28
Issues with ESIL flag computation


sushant94:
There are some issues when esil computes flags. This bug is to consolidate these and work towards a fix.

1. Carry flag for subtraction is set using `$c,cf,=` rather than using `$b,cf,=`
2. Overflow flag is not computed correctly (for subtraction)
3. ESIL never sets Adjust Flag (AF)
4. x86 add (`+=`) does not set flags.
5. x86 sub  does not set `pf`

I'm working towards a fix (and respective tests). Will update this issue when I find more bugs.

radare:
Cc @condret

sushant94:
Some more points to consider:
- How do we go about setting the CF for SHL/SHR (and other variants) [[See Link]](http://x86.renejeschke.de/html/file_module_x86_id_285.html)
- Even though new operations `>>>` and `<<<` are in ESIL, these are not currently used to implement ROR/ROL of x86. It is possible to implement these using this, however, the other variants (RCL/RCR) include the carry flag in the shifts and hence we cannot use this directly. [[See Link]](http://x86.renejeschke.de/html/file_module_x86_id_273.html)

condret:
````$b,cf,=```` will never work, it must be ````$b63,cf,=```` for example

condret:
the shl/shr thing is easy. it works similiar in gb-land: srl a
``1,a,&,C,=,1,a,>>=,$z,Z,=,0,N,=,0,H``

condret:
i really don't like ````>>>```` and ````<<<```` because they need 3 args pushed on the stack, if i remember right

sushant94:
@condret
Why do you say just `$b` dies not work? Cause this seems to be working. Then again, I might be wrong and it might have just worked for this case.

 I don't know about gb, but for x86, the carry flag is not defined if the number of bits shifted out is more than the width of the destination. I'm not sure on how to account for this.

And I don't know where `>>>` and `<<<` are used and who introduced it. It __could__ be used here but we need to find a way around for RCL/RCR.

XVilka:
Added the 0.10 milestone, since I think ESIL is now widely (more or less) used, so it is better to provide fixes in the release.

condret:
@sushant94  $b means borrow-from-bit, if you just do $b, you do not define from what bit you want to check the borrow. So $b performs undefined behaviour, which may end up in analdeamons eating your pet/child/keyboard, or launching nethack.
[...]

Issue #17395, opened 2020-08-03
OpenBSD - failing "db/anal/x86_64 anal hasnext" test


XVilka:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | OpenBSD CI (SourceHut)
| File format of the file you reverse (mandatory)      | -
| Architecture/bits of the file (mandatory)            | x86/64
| r2 -v full output, **not truncated** (mandatory)     | https://github.com/radareorg/radare2/commit/2a00f5fee1a4b71fc087444bb0f307224aacb354

See full log here: https://builds.sr.ht/~xvilka/job/271607#task-test-0

```
[XX] db/anal/x86_64 anal hasnext on main
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e anal.hasnext=true
af@main
afi@main~size
' bins/elf/true32
-- stdout
@@ -1 +0,0 @@
-size: 261

-- stderr
0x08048ed9 > 0x08048efe 9943
0x08048f30 > 0x08049197 9947
0x08048b80 > 0x08048b86 9949
0x08048b86 > 0x08048b90 9951
0x08048b90 > 0x08048b96 9953
0x08048ba0 > 0x08048ba6 9957
0x08048bb0 > 0x08048bb6 9961
0x08048bc0 > 0x08048bc6 9965
0x08048bd0 > 0x08048bd6 9969
0x08048be0 > 0x08048be6 9973
0x08048bf0 > 0x08048bf6 9977
0x08048c00 > 0x08048c06 9981
0x08048c10 > 0x08048c16 9985
0x08048c20 > 0x08048c26 9989
0x08048c30 > 0x08048c36 9993
0x08048c40 > 0x08048c46 9997
0x08048cd0 > 0x08048cd6 9949
0x08048ce0 > 0x08048ce6 9953
0x08048cf0 > 0x08048cf6 9957
0x08048d00 > 0x08048d06 9961
0x08048d10 > 0x08048d16 9965
0x08048d20 > 0x08048d26 9969
0x08048d30 > 0x08048d36 9973
0x08048d40 > 0x08048d46 9977
0x08048d50 > 0x08048d56 9981
0x08048c70 > 0x08048c76 9949
0x08048c80 > 0x08048c86 9953
0x08048c90 > 0x08048c96 9957
0x08048ca0 > 0x08048ca6 9961
[...]

PR #17436, opened 2020-08-10
ESIL signedness and other extensions ##emu ##anal

aemmitt-ns:
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible) (tested with esilcheck, I can make other tests also)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed) - Will update the r2book ESIL page 

**Detailed description**

- esil.c 
  - added sign extension operator: `~`
  - added signed division and signed modulus operators: `~/` and `~%`

- x86 ESIL
  - fixed `IDIV` so that it uses the sign extension and signed division / modulus operators. It still needs a fix to work with the full 128 bit RDX:RAX values though
  - fixed `IMUL` so it sign extends the values first, otherwise signed multiplication is the same as unsigned
  - Changed `MOVSX` to sign extend the source value properly
  - Changed `SUB` and `CMP` so that they set the flags right when second operand is 2**(bits-1) 
  - Added AF flag setting to relevant instructions

- ARM 64 ESIL
  - Added expression for `MNEG` based on VEX2ESIL
  - Added expression for `SBC` but have not been able to verify it 
  - Differentiated `SDIV` from `UDIV`, used sign extension and signed division
  - Added expression for `CLZ`, first using GOTO and then something better
  - Added case for `STNP` which for ESIL is equivalent to `STP`
  - Added expression for `EXTR` using VEX2ESIL, and verified with ESILCheck 
  - Added insanely long expression for `RBIT` based on VEX2ESIL, there has to be a better way to do this one
  - Added case for `MVN` which right now is the same as `MOVN`, need to verify if this is accurate
  - Added expressions for `SBFIZ`, `UBFIZ`, and `SBFX`, expressions that insert variable length bitstrings with and without sign extension
  - Added support for the signed `LDR` instructions by splitting them into a second case and adding sign extensions
  - Changed `OPCALL` to not use `opchar=` pattern since some operators did not have that form and were silently failing
  - Fixed a tricky issue in the `REV*` instructions where something like `rev x0, x0` would produce incorrect results because partial results were stored in the dst register causing issue when it was also the src register
  - Added expression for `SMADDL` to do the sign extension 
  - Added sign extension to `SMULL` and `SDIV`
  - Fix `LSL`, `LSR`, `ASR` to use the remainder as the shift amount 
  - Add shift to `MOVN`
  - Change the `SXT*` instructions to use the new signext operator and fix some inconsistencies 
  - Made many other changes to `LDR` and `STR` to properly evaluate the different shifts and extensions of the memory offset 
    - Added first register set to 0 for `STXR`, as it always succeeds for ESIL
    - Completely redo code for `STR` to mirror the code for `LDR` which handles the shifts and extensions properly
    - Fix pre and post indexing so it uses the previously DUPed values
  - Add shift / extensions to `CM*` instructions 
  - Added decode_extension function to be used in the modified shifted_register_append  

**Test plan**

These changes were tested with ESILCheck, a tool derived from ESILSolve which I am going to open source imminently. I can create other tests for the esil improvements but I am unsure what the protocol is there, as I am not sure what happened to the ESIL unit tests. I also made sure to check that some previous programs that emulated arm64 still worked the same

**Closing issues**

Does not close any issue that I could find, though they address many TODOs previously in the code. 

Actually it does address part of this (very old) issue involving the adjust flag https://github.com/radareorg/radare2/issues/3607


aemmitt-ns:
working on getting the failing tests fixed

XVilka:
Amazing work, thank you! Will get through later.

libr/anal/esil.c
```diff
@@ -3229,7 +3318,7 @@ static void r_anal_esil_setup_ops(RAnalEsil *esil) {
 	OP ("--=[2]", esil_mem_deceq2, 0, 1, OT_MATH | OT_MEMR | OT_MEMW);
 	OP ("--=[4]", esil_mem_deceq4, 0, 1, OT_MATH | OT_MEMR | OT_MEMW);
 	OP ("--=[8]", esil_mem_deceq8, 0, 1, OT_MATH | OT_MEMR | OT_MEMW);
-        OP ("<<=[]", esil_mem_lsleq, 0, 2, OT_MATH | OT_MEMR | OT_MEMW);
+    OP ("<<=[]", esil_mem_lsleq, 0, 2, OT_MATH | OT_MEMR | OT_MEMW);
```
XVilka, line 3321: Broken indentation it seems?

XVilka:
By the way you might be interested in https://github.com/sushant94/esil-tests
See also https://github.com/radareorg/radare2/issues/16400

aemmitt-ns:
the tools referred to in the PR are [here](https://github.com/aemmitt-ns/esilsolve/tree/master/tools ). They use symbolic execution of ESIL to check whether there are any values where the ESIL implementation differs from the VEX implementation in angr. They are still a WIP and are pretty janky but they are useful.

FXTi:
ESIL part LGTM.

libr/anal/esil.c
```diff
@@ -544,6 +544,42 @@ R_API int r_anal_esil_reg_read(RAnalEsil *esil, const char *regname, ut64 *num,
 	return ret;
 }
 
+// sign extension operator for use in idiv, imul, movsx* 
+// and other instructions involving signed values
```
FXTi, line 548: Better add some comments like examples:
```
<bit-width of value>,<value>,~
"ae 8,0xfb,~"
```

FXTi: Actually, more detailed can also be added: Extend signed integer to the size of st64 on machine running this ESIL vm.
Since my opinion come from the view of P-code to ESIL translation, I will care about something like this while others may take this as granted. So whether to add is up to you.
And `~/` and `~%` do not need extra comments as long as them works just like normal `/` and `%`.

FXTi: Anyway, I hope this pr will be merged sooner.

aemmitt-ns: Yea having a signext operator will help with all different kinds of IR translation. I am not sure if the VEX2ESIL tool I have in my ESILSolve repo could be useful to the pcode stuff but there are probably similar problems involved

XVilka:
Ignore the OpenBSD error, it's https://github.com/radareorg/radare2/issues/17395

libr/anal/p/anal_x86_cs.c
```diff
@@ -1462,8 +1475,8 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 			// Set OF, SF, ZF, AF, PF, and CF flags.
 			// We use $b rather than $c here as the carry flag really
 			// represents a "borrow"
-			esilprintf (op, "%s,%s,%d,$o,of,:=,%d,$s,sf,:=,$z,zf,:=,$p,pf,:=,%d,$b,cf,:=",
-				src, dst, bitsize - 1, bitsize - 1, bitsize);
+			esilprintf (op, "%s,%s,%s,%u,-,!,%d,$o,^,of,:=,%d,$s,sf,:=,$z,zf,:=,$p,pf,:=,%d,$b,cf,:=,3,$b,af,:=",
+				src, dst, src, (1U << (bitsize - 1)), bitsize - 1, bitsize - 1, bitsize);
```
trufae, line 1479: bitsize can overflow this shift (undefined behaviour)

libr/anal/p/anal_x86_cs.c
```diff
@@ -795,10 +797,21 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
...
+					else {
+						esilprintf (op, "%s,%s,=", src, dst64);
+					}
+
 				} else {
-					esilprintf (op, "%s,%s,=", src, dst);
-				}
+					if (insn->id == X86_INS_MOVSX || insn->id == X86_INS_MOVSXD) {
+						esilprintf (op, "%d,%s,~,%s,=", width*8, src, dst);
+					}
```
trufae, line 810: else goes after the }, not below

libr/anal/esil.c
```diff
@@ -3132,6 +3220,7 @@ static void r_anal_esil_setup_ops(RAnalEsil *esil) {
 	OP ("$js", esil_js, 1, 0, OT_UNK);
 	OP ("$r", esil_rs, 1, 0, OT_UNK);
 	OP ("$$", esil_address, 1, 0, OT_UNK);
+	OP ("~", esil_signext, 1, 2, OT_MATH);
```
trufae, line 3223: not sure if ~ is the best char for this because it's technically used for NOT in other contexts. but i cant think of a better char. comments?

libr/anal/esil.c
```diff
@@ -544,6 +544,44 @@ R_API int r_anal_esil_reg_read(RAnalEsil *esil, const char *regname, ut64 *num,
...
+	if (!p_src) {
+		return false;
+	}
+
+	if (!r_anal_esil_get_parm (esil, p_src, &src)) {
+		ERR ("esil_of: empty stack");
+		free (p_src);
+		return false;
+	}
+	free (p_src);
```
trufae, line 564: use goto or else to avoid having this free (p_src) twice, like it happens below

libr/anal/esil.c
```diff
@@ -544,6 +544,44 @@ R_API int r_anal_esil_reg_read(RAnalEsil *esil, const char *regname, ut64 *num,
...
+	if (!r_anal_esil_get_parm (esil, p_dst, &dst)) {
+		ERR ("esil_of: empty stack");
+		free (p_dst);
+		return false;
+	}
+	free (p_dst);
+
+	ut64 m = 1U << (dst - 1);
+	// dst = (dst & ((1U << src_bit) - 1)); // clear upper bits
+	src = ((src ^ m) - m); // sign-extended
```
trufae, line 580: define src and assign in the same line

libr/anal/esil.c
aemmitt-ns: Yea I had this internal debate for a long time and thought about some other possible characters but i think ~ is still the best. Its twos-complement in many languages so I think it makes sense to associate with sign extension. I also don't like that its grep in r2 but a bunch of the other operators are also special in r2 so that doesn't change much. I think the other option is just to have it be "SIGN", which I would be fine with but it would make expressions bulkier.

libr/anal/p/anal_x86_cs.c
aemmitt-ns: is there a preference between using (ut64)1 or 1ULL? Other places use 1LL.

trufae: Better use 1ULL imho

test/db/cmd/cmd_print
```diff
@@ -431,7 +431,7 @@ EXPECT=<<EOF
   },
   {
     "offset": 5,
-    "esil": "rcx,rdx,==,$z,zf,:=,64,$b,cf,:=,$p,pf,:=,63,$s,sf,:=,63,$o,of,:=",
+    "esil": "rcx,rdx,==,$z,zf,:=,64,$b,cf,:=,$p,pf,:=,63,$s,sf,:=,rcx,9223372036854775808,-,!,63,$o,^,of,:=,3,$b,af,:=",
```
trufae, line 434: uh this 9223372036854775808 is 0x8000000000000000.. couldnt be represented in a better way? esil supports hex values too in numbers.

libr/anal/p/anal_x86_cs.c
```diff
@@ -1462,8 +1473,8 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 			// Set OF, SF, ZF, AF, PF, and CF flags.
 			// We use $b rather than $c here as the carry flag really
 			// represents a "borrow"
-			esilprintf (op, "%s,%s,%d,$o,of,:=,%d,$s,sf,:=,$z,zf,:=,$p,pf,:=,%d,$b,cf,:=",
-				src, dst, bitsize - 1, bitsize - 1, bitsize);
+			esilprintf (op, "%s,%s,%s,%"PFMT64u",-,!,%d,$o,^,of,:=,%d,$s,sf,:=,$z,zf,:=,$p,pf,:=,%d,$b,cf,:=,3,$b,af,:=",
```
trufae, line 1476: if bitsize > 63 { UB }

libr/anal/esil.c
```diff
@@ -3169,8 +3258,10 @@ static void r_anal_esil_setup_ops(RAnalEsil *esil) {
 	OP ("--", esil_dec, 1, 1, OT_MATH);
 	OP ("--=", esil_deceq, 0, 1, OT_MATH | OT_REGW);
 	OP ("/", esil_div, 1, 2, OT_MATH);
+	OP ("~/", esil_signed_div, 1, 2, OT_MATH);
```
trufae, line 3261: we need signed multiplication, addition and substraction, i guess @condret would prefer a SIGN operator instead of adding so many new instructions to the esil vm.

aemmitt-ns: After sign extension multiplication, addition, subtraction are all the same as unsigned, and yea I also added the signed division operator (~/) and signed modulus (~%). Also ~ (SIGN) pops two values from the stack, the value to sign extend and the size of the value before extension, and it pushes the sign extended 64 bit value to the stack. I could modify it to have a target bit size also but I haven't found a place where that would be necessary.

aemmitt-ns: I also want to try to minimize the number of operators, I could possibly eliminate the need for the signed mod but it would make expressions for instrs like IDIV even wilder than they are already. I think otherwise they are the only new operators required to cover all simple math with signed integers.

XVilka:
@aemmitt-ns have you considered to check other common architectures? E.g. MIPS or PowerPC? @junchao-loongson and @RinGotou might be interested in improving ESIL precision for MIPS e.g.

aemmitt-ns:
@XVilka  Yes my tools in ESILSolve (VEX2ESIL and ESILCheck) will also work for MIPS and PPC so I am gonna work on them soon. I still want to do more work on x86, though accurately emulating every x86 instruction is probably 200 years of work. I would definitely encourage checking out VEX2ESIL, it should work right now for MIPS (32 and 64). I haven't tried PPC but vex does support it.

libr/anal/p/anal_x86_cs.c
aemmitt-ns: Ok should be fixed now, all shifts have an explicit check for < 64

aemmitt-ns:
Is this ready to move on from draft?

libr/anal/esil.c
```diff
@@ -544,6 +544,48 @@ R_API int r_anal_esil_reg_read(RAnalEsil *esil, const char *regname, ut64 *num,
...
+		free (p_dst);
+	}
+
+	ut64 m = 0;
+	if (dst < 64) {
+		m = 1ULL << (dst - 1);
+	}
+
+	// dst = (dst & ((1U << src_bit) - 1)); // clear upper bits
+	return r_anal_esil_pushnum (esil, ((src ^ m) - m));
```
condret, line 586: so technically this is just xor and subtraction?

condret:
this seems to be quite focused on x86 ... what do you think about making these operations customops in the anal x86-cs plugin?

libr/anal/esil.c
aemmitt-ns: Yea it can be done with existing operators and has been done in a bunch of ad hoc ways in different archs. But that blows up the size and complexity of expressions and I think there is a lot of value in just having the operator so we can standardize it.

aemmitt-ns:
@condret most of the changes are actually in the arm64 plugin. and I think a couple mips instrs need the sign extension fixed so I made add it there in the future as well.

trufae:
Other architectures also have signed arithmetic instructions. in mips for example:

`In MIPS assembly language, there is a multiplication instruction for signed integers, mult, and
for unsigned integers multu.`

on arm:

Screenshot 2020-08-16 at 10 27 26

This PR focus on x86, but imho it is applicable to RISC architectures too

some ref for riscv which is more interesting to read: https://stackoverflow.com/questions/59234619/riscv-rv32im-mulhsu-which-operand-is-the-signed-one

Screenshot 2020-08-16 at 10 29 14

trufae (approved):

libr/anal/esil.c
```diff
@@ -544,6 +544,48 @@ R_API int r_anal_esil_reg_read(RAnalEsil *esil, const char *regname, ut64 *num,
...
+	char *p_src = r_anal_esil_pop (esil);
+	if (!p_src) {
+		return false;
+	}
+
+	if (!r_anal_esil_get_parm (esil, p_src, &src)) {
+		ERR ("esil_of: empty stack");
+		free (p_src);
+		return false;
+	} else {
```
trufae, line 563: dont else after return

libr/anal/esil.c
```diff
@@ -544,6 +544,48 @@ R_API int r_anal_esil_reg_read(RAnalEsil *esil, const char *regname, ut64 *num,
...
+	char *p_dst = r_anal_esil_pop (esil);
+	if (!p_dst) {
+		return false;
+	}
+
+	if (!r_anal_esil_get_parm (esil, p_dst, &dst)) {
+		ERR ("esil_of: empty stack");
+		free (p_dst);
+		return false;
+	} else {
```
trufae, line 576: dont else after return

junchao-loongson:
> @aemmitt-ns have you considered to check other common architectures? E.g. MIPS or PowerPC? @junchao-loongson and @RinGotou might be interested in improving ESIL precision for MIPS e.g.

Sign extension is urgently needed in MIPS ESIL

junchao-loongson:
After this PR is merged，I will modify the code corresponding to MIPS.(asm_mips_gnu.c and asm_mips_cs.c)

trufae:
@junchao-loongson here you have

trufae:
Thanks @aemmitt-ns ! good work!

Issue #3222, opened 2015-09-02
Structured document helpers


radare:
R2 already have a json tokenizer, writing an xml one should be simple, and we can also use a C one using shlr/tcc or ALT to parse the structured code and define flags for every opened bracked etc. This can be useful to find and analyze binary stuff inside structured text documents like XML or JSON. 

* Seek to source line Nth
* Seek to column Nth
* Seek to token Nth
* Seek to the end of the current scope (like % in vim ?)
* Enumerate all start/end tokens

This way we can also replace `ed`.

This stuff can be implemented in RBin as a plugin that detects XML and JSON documents by checking the first bytes of the file.

radare:
Related https://github.com/radare/alt

radare:
I have implemented this for JSON, we will need that for XML too

XVilka:
See https://github.com/radare/radare2/issues/13312

XVilka:
There is even more - we can consider to reuse [Tree Sitter](https://github.com/tree-sitter/tree-sitter) library to generate parsers for all these languages:
- [XML](https://github.com/tree-sitter/tree-sitter/issues/266)
- [JSON](https://github.com/tree-sitter/tree-sitter-json)

XVilka:
See also [CSV support request](https://github.com/radare/radare2/issues/13763)

PR #17439, opened 2020-08-10
Add the r_json json parser based on nxjson ##util

thestr4ng3r:

**Detailed description**

r_json is a json parser heavily based on https://github.com/yarosla/nxjson. It is only a parser so does not duplicate anything from pj.
I evaluated every single json parser in C that I could find against my requirements which are the following:
* In-place, i.e. no strings are copied out of the json string and instead you just get references into the string, but they should be unescaped
* No fixed upper bound on how many tokens you can parse
* Handle utf-8 correctly
* Be able to parse full 64bit signed AND unsigned integers because we need those all the time
* As small and simple as possible, no json formatter because we already have that

js0n, which we already have in sdb, is very nice for what it is, but it is not suitable at all for most purposes since it only gives back the position for a single key, so to parse a full json you have to re-parse the entire json multiple times. Moreover it does not un-escape strings, read integer values, etc.

The only suitable parser I found was nxjson (after some patches) so this is the heavily r2-ified version of it that meets all of the above requirements.

It has been part of r2db for quite some time and works very nice there. It is also relevant for #17430 and for RAnnotatedCode, so we can later parse RAnnotatedCode from json that comes from a decompiler that we don't directly access through C (like r2dec) and we will be able to take advantage of all the nice annotation features.

**Test plan**

Run the unit tests. These come directly from https://github.com/yarosla/nxjson/tree/master/tests and were brought into C using this script: https://github.com/thestr4ng3r/nxjson/blob/master/gen_tests.hs


test/unit/test_json.c
```diff
@@ -0,0 +1,338 @@
...
+		r_strbuf_init (&buf);
+		dump (json, &buf, 0);
+		r_json_free (json);
+		mu_assert_streq (r_strbuf_get (&buf), expected_output, "dump");
+		r_strbuf_fini (&buf);
+	}
+	mu_end;
+}
+
+static int all_tests() {
```
trufae, line 324: Add (void)

test/unit/test_json.c
```diff
@@ -0,0 +1,338 @@
...
+		char testname[256];
+		snprintf (testname, sizeof(testname), "json_test (%u)", (unsigned int)i);
+		mu_run_test_named (json_test, testname, i, input, tests[i].expected);
+		free (input);
+	}
+	return tests_passed != tests_run;
+}
+
+int main() {
+	return all_tests();
```
trufae, line 337: Missing space

libr/util/json.c
```diff
@@ -0,0 +1,374 @@
...
+			}
+			break;
+		default:
+			R_JSON_REPORT_ERROR ("unexpected chars", p);
+			return 0; // error
+		}
+	}
+}
+
+R_API RJson *r_json_parse(char *text) {
```
trufae, line 344: Const char *text

libr/util/json.c
```diff
@@ -0,0 +1,374 @@
...
+			}
+		} else {
+			*d++ = c;
+		}
+	}
+	R_JSON_REPORT_ERROR ("no closing quote for string", s);
+	return 0;
+}
+
+static char *skip_block_comment(char *p) {
```
trufae, line 136: Const char*

libr/include/r_util/r_json.h
```diff
@@ -0,0 +1,67 @@
...
+		struct {                // children of OBJECT or ARRAY
+			int length;
+			struct r_json_t *first;
+			struct r_json_t *last;
+		} children;
+	};
+	struct r_json_t *next;    // points to next child
+} RJson;
+
+R_API RJson *r_json_parse(char *text);
```
trufae, line 56: Const char *

libr/util/json.c
```diff
@@ -0,0 +1,374 @@
...
+#else
+#define R_JSON_REPORT_ERROR(msg, p) do { (void)(msg); (void)(p); } while (0)
+#endif
+
+static RJson *json_new(void) {
+	return R_NEW0 (RJson);
+}
+
+static RJson *create_json(RJsonType type, const char *key, RJson *parent) {
+	RJson *js = json_new();
```
trufae, line 21: Space

libr/util/json.c
```diff
@@ -0,0 +1,374 @@
...
+		R_JSON_REPORT_ERROR ("endless comment", ps);
+		return 0;
+	}
+	REPEAT:
+	p = strchr (p + 1, '/');
+	if (!p) {
+		R_JSON_REPORT_ERROR ("endless comment", ps);
+		return 0;
+	}
+	if (p[-1] != '*') goto REPEAT;
```
trufae, line 149: Braces

test/unit/test_json.c
```diff
@@ -0,0 +1,338 @@
...
+	}, { // 50
+		"\"\\u0",
+		NULL
+	}, { // 51
+		"\"\\u04\"",
+		NULL
+	}
+};
+
+static void dump(const RJson *json, RStrBuf *out, int indent) {
```
trufae, line 248: Make it public api as char *r_json_tostring(RJson)

libr/util/json.c
```diff
@@ -0,0 +1,374 @@
...
+				}
+			}
+		case ']':
+			return p;
+		case '"':
+			p++;
+			js = create_json (R_JSON_STRING, key, parent);
+			js->str_value = unescape_string (p, &p);
+			if (!js->str_value) {
+				return 0; // propagate error
```
trufae, line 252: Missing rjsonreport call here. Is this on purpose?

libr/util/json.c
```diff
@@ -0,0 +1,374 @@
...
+			*d++ = c;
+		}
+	}
+	R_JSON_REPORT_ERROR ("no closing quote for string", s);
+	return 0;
+}
+
+static char *skip_block_comment(char *p) {
+	// assume p[-2]=='/' && p[-1]=='*'
+	char *ps = p - 2;
```
trufae, line 138: This assumption is scary. It is properly checked on the caller? Or maybe it’s better to pass two pointers to this function to avoid this risk

libr/util/json.c
```diff
@@ -0,0 +1,374 @@
...
+	while (1) {
+		switch (*p) {
+		case '\0':
+			R_JSON_REPORT_ERROR ("unexpected end of text", p);
+			return 0; // error
+		case ' ':
+		case '\t':
+		case '\n':
+		case '\r':
+		case ',':
```
trufae, line 212: Comma is not always valid in all the contexts. This case should have extra logic to check for that

libr/include/r_util/r_json.h
```diff
@@ -0,0 +1,67 @@
...
+} RJsonType;
+
+typedef struct r_json_t {
+	RJsonType type;             // type of json node, see above
+	const char *key;            // key of the property; for object's children only
+	union {
+		const char *str_value;  // text value of STRING node
+		struct {
+			union {
+				ut64 u_value;   // the value of INTEGER or BOOL node
```
trufae, line 42: Yayyy json-ut64 \o/

libr/include/r_util/r_json.h
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+typedef enum r_json_type_t {
+	R_JSON_NULL,
+	R_JSON_OBJECT,  // properties can be found in child nodes
+	R_JSON_ARRAY,   // items can be found in child nodes
+	R_JSON_STRING,  // value can be found in the str_value field
+	R_JSON_INTEGER, // value can be found in the num.u_value/num.s_value fields
+	R_JSON_DOUBLE,  // value can be found in the num.dbl_value field
+	R_JSON_BOOL     // value can be found in the num.u_value field
```
trufae, line 32: Name it BOOLEAN. To be consistent with the rest of the enum

XVilka:
Related issue: https://github.com/radareorg/radare2/issues/3222

trufae:
I dont see how this implementation can solve or be related to #3222

XVilka:
> I dont see how this implementation can solve or be related to #3222
> Structured document helpers

It's all about making easy interface for JSON, XML, CSV that can be reused everywhere in r2. So definitely related. Moreover, it is related also to https://github.com/radareorg/radare2/issues/12870.

libr/util/json.c
thestr4ng3r: All of these must be non-const because the parser works in-place and modifies the string. If you want to parse a const char * you have to strdup it yourself before.

test/unit/test_json.c
thestr4ng3r: This isn't actually producing json, it's just some random format used for comparing test results, so not really useful in the API. For testing this format actually makes more sense than printing actual json because you can compare possibly invalid utf-8 strings on the binary level.

libr/util/json.c
thestr4ng3r: Yes, unescape_string does more detailed reporting about what is wrong with the string itself.

libr/util/json.c
thestr4ng3r: Changed to passing p - 2 directly.

libr/include/r_util/r_json.h
thestr4ng3r: Yes, BOOLEAN is also more consistent with Java :-P

libr/util/json.c
thestr4ng3r: Agreed, I have to check the best way to fix this.

libr/util/json.c
trufae: Or use pascal style strings like js0n do

test/unit/test_json.c
trufae: Would be better to stick to json. Maybe using the pj api

trufae:
@xvilka for me the 3222 issue is about having error prone, interruptable and stream based parsers for common formats inside r2 that work on any io backend and provide apis to seek to the next element, find matching bracket etc

libr/util/json.c
thestr4ng3r: What do pascal style strings have to do with this?

test/unit/test_json.c
thestr4ng3r: As said for this special case the raw dumping is slightly better and has less behavorial dependencies. But a printing for actual usage could easily be built with pj. I'm just not sure if it is really practically useful.

libr/util/json.c
thestr4ng3r: Good catch, turns out it was just ignoring all commas and treating them as whitespace, accepting stuff like `{{}{}}` or `[,,,,,,42,,,,,1337,,,,]`. I made it strict about commas now.

thestr4ng3r: Also added tests for that stuff.

thestr4ng3r:
> Where do you thibj this api can be useful for outside projects?

As mentioned in the description, it can be made part of the future decompiler api so decompilers can just supply annotated json. Maybe we will also find usages for it in the HTTP server or something like json-rpc for commands.

> What do you think about supporting binary json?

If you mean this: http://bsonspec.org/spec.html
Should be possible, at least for a subset of it, or by extending the `RJson` struct for more types.

XVilka:
There are multiple Binary JSON formats, while it might be usable, do we have any use case already? For the PR part it's LGTM.

libr/util/json.c
```diff
@@ -0,0 +1,391 @@
...
+		}
+	}
+	R_JSON_REPORT_ERROR ("no closing quote for string", s);
+	return NULL;
+}
+
+static char *skip_block_comment(char *ps) {
+	// ps is at "/* ..."
+	// caller must ensure that ps[0], ps[1] and ps[2] are valid.
+	char *p = ps + 2;
```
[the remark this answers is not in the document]
trufae: it will be good to run this code thru radamsa, i think there are other bugs like this in other places

thestr4ng3r:
> another interesting addition to this api to cover the #3222 issue is to make it interruptable, this is.. when parsing huge or truncated jsons from corrupted dumps it is useful to be able to parse it until it fails. how this api handle invalid jsons?

It fails and returns NULL on the first error. You would need to make it a streaming parser to parse truncated json which it currently isn't. I think that's doable by refactoring it a bit but the concept here is different.

> can we identify at which offset inside the string the element is like js0n do?

Theoretically adding this functionality would be trivial, but it would take up more memory for a feature that I think will be only used in rare cases.

> otherwise i dont see much benefit on this api compared to js0n apart from a cleaner api.

Please read the description of the pr: js0n, which we already have in sdb, is very nice for what it is, but it is not suitable at all for most purposes since it only gives back the position for a single key, so to parse a full json you have to re-parse the entire json multiple times. Moreover it does not un-escape strings, read integer values, etc.

> also, the testsuite is just comparing strings. it will be good to check for objects and types as well as some invalid json inputs. jq have a nice testsuite for that

Many of the tests are testing for invalid jsons. Take a look at the `JsonTest tests[]` array. All the entries that have `NULL` as the second string are testing for invalid.

libr/util/json.c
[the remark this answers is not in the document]
thestr4ng3r: Yes, it is checked. The only callsite is in `skip_whitespace()` where the call to this function is enclosed in `if (*p == '/' && p[1] == '*')` so `p[2]` is automatically valid. If you really want to have checks there, I can add them but it would be redundant.

Actually I also thought about fuzzing this api before. It might even be a nice target to run klee on since it's small enough to get some good results.

libr/util/json.c
```diff
@@ -0,0 +1,391 @@
...
+			if (*p == ',') {
+				char *commapos = p;
+				p++;
+				p = skip_whitespace (p);
+				if (*p == '}') {
+					R_JSON_REPORT_ERROR ("trailing comma", commapos);
+					return NULL;
+				}
+			} else if (*p == '}') {
+				return p + 1; // end of object
```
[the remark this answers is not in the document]
thestr4ng3r: Not in this case. if `*p == ','`, it will not necessarily return in the if and you don't want to fall into the bottom else in that case.

test/unit/test_json.c
thestr4ng3r: What would you say about instead of comparing these dumped strings at all, doing it all directly in C by `mu_assert(...` calls? I could probably write a code generator for all these cases so I don't have to write them from scratch. If we would go through pj I think we would test too much interdependent code here when we just want to test the parser and nothing else.

thestr4ng3r:
> > another interesting addition to this api to cover the #3222 issue is to make it interruptable, this is.. when parsing huge or truncated jsons from corrupted dumps it is useful to be able to parse it until it fails. how this api handle invalid jsons?
> 
> It fails and returns NULL on the first error. You would need to make it a streaming parser to parse truncated json which it currently isn't. I think that's doable by refactoring it a bit but the concept here is different.

I thought a bit more about it and it should be easy to make it parse truncated json without making it a streaming parser. You would just get the parsed json up until that point + the pointer of the first error. Depends on the use-case I would say if that's what you want since it would parse the entire string before you get a result.

libr/util/json.c
trufae: instead of null terminate the strings specify the length

libr/util/json.c
trufae: Awesome! Good work

libr/include/r_util/r_json.h
```diff
@@ -0,0 +1,67 @@
...
+		struct {                // children of OBJECT or ARRAY
+			int length;
+			struct r_json_t *first;
+			struct r_json_t *last;
+		} children;
+	};
+	struct r_json_t *next;    // points to next child
+} RJson;
+
+R_API RJson *r_json_parse(char *text);
```
trufae, line 56: Another option is to make a one-level json parser like this: `RJson *r_json_parse_once(char *s);` 

Also, i dont like having mutable string by default in here, what do you think about having two parsers, keeping the pointer inside RJson for later freeing.

```
RJson *r_json_parse_mut(char *s);
RJson *r_json_parse(const char *s);
```

thestr4ng3r: Possible, but `RJson` is used for all nested values too while you will only want the complete string at the top-level so you would have to bloat it hard.
Alternatively, you could do this:

```c
RJson *r_json_parse_mut(char *s);

typedef struct {
    char *s;
    RJson *j;
} RParsedJson;

RParsedJson *r_json_parse(const char *s) {
    RParsedJson *r = R_NEW(RParsedJson);
    r->s = strdup(s);
    r->j = r_json_parse_mut (r->s);
    return r;
}

void r_json_parsed_json_free(RParsedJson *j) {...}
```
which is a bit ugly.

This mutable in-place parsing is actually a major benefit whenever you can take advantage of it. It saves you a lot of copying.

test/unit/test_json.c
thestr4ng3r: Removed the non-standard dumping stuff and replaced with direct tests.

thestr4ng3r:
> as long as strings are referencing the original buffer string it should be possible to determine the position inside the buffer of any specific element, right? didnt tested that but it should work

Yes, but only for string values and keys.

thestr4ng3r:
> Im happy to merge this but i do think:
> 
>     * r_json_tostring() must be implemented and use the pj api
> 
>     * mutable string as argument can be problematic, but i understand its easier way to go if used properly

Yep, these things can be added on top.

PR #17441, opened 2020-08-11
Escape environment vars in the default rarun profile

abcSup:

**Detailed description**
Escape the environment variable values with `'` in the startup rarun profile to prevent misinterpreting special characters as rarun prefixes.

For example, an env var with special characters like `DISPLAY=:0` should not be parsed as a hexpair string by `run.c:getstr`

**Test plan**

...

**Closing issues**

### DIFF
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -386,26 +386,8 @@ R_API int r_main_radare2(int argc, const char **argv) {
 
 	r_sys_env_init ();
 	// Create rarun2 profile with startup environ
-	char *envprofile = NULL;
-	char **e = r_sys_get_environ ();
-	if (e) {
-		RStrBuf *sb = r_strbuf_new (NULL);
-		while (e && *e) {
-			char *k = strdup (*e);
-			char *v = strchr (k, '=');
-			if (v) {
-				*v++ = 0;
-				v = r_str_escape (v);
-				if (v) {
-					r_strbuf_appendf (sb, "setenv=%s=\"%s\"\n", k, v);
-					free (v);
-				}
-			}
-			free (k);
-			e++;
-		}
-		envprofile = r_strbuf_drain (sb);
-	}
+	char **env = r_sys_get_environ ();
+	char *envprofile = r_run_get_environ_profile (env);
 
 	if (r_sys_getenv_asbool ("R2_DEBUG")) {
 		char *sysdbg = r_sys_getenv ("R2_DEBUG_TOOL");
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -395,7 +395,11 @@ R_API int r_main_radare2(int argc, const char **argv) {
 			char *v = strchr (k, '=');
 			if (v) {
 				*v++ = 0;
-				r_strbuf_appendf (sb, "setenv=%s='%s'\n", k, v);
+				v = r_str_escape (v);
+				if (v) {
+					r_strbuf_appendf (sb, "setenv=%s=\"%s\"\n", k, v);
+					free (v);
+				}
 			}
 			free (k);
 			e++;
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -391,7 +391,13 @@ R_API int r_main_radare2(int argc, const char **argv) {
 	if (e) {
 		RStrBuf *sb = r_strbuf_new (NULL);
 		while (e && *e) {
-			r_strbuf_appendf (sb, "setenv=%s\n", *e);
+			char *k = strdup (*e);
+			char *v = strchr (k, '=');
+			if (v) {
+				*v++ = 0;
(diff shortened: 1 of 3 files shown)


karliss:
What if the environment variable contains `'` symbol or any other special symbol interpreted by rarun profile mechanism. From what I understood `r_run_parseline` uses run.c/getstr which deals s with a few special symbols.

abcSup:
> What if the environment variable contains `'` symbol or any other special symbol interpreted by rarun profile mechanism. From what I understood `r_run_parseline` uses run.c/getstr which deals s with a few special symbols.

You are right. I am trying to copy all original env vars from the environment as they are exactly and not interpret them as special characters. For example, `:0` in a line `setenv=DISPLAY=:0` will be parsed by `getstr` as a hexpair string which ends up setting `DISPLAY` to `"\0"`, so I escape all original env vars from the environment with `'` to prevent `getstr` from interpreting them differently.

karliss:
As a general rule of thumb working string escaping system need at least these 3 parts:
* quotes
* mechanism to escape quotes
* mechanism to escape the escape mechanism

If you have only first one or first two it can't be a working string escape system which supports all symbols. In your code I only see the first one. Without testing this makes me think that current code doesn't properly solve the problem.

In some use cases the whole problem can be sidestepped if there are string bounds established using external information or some symbol that can never appear in the text (with limitation that this symbol can't be used in the text).

karliss:
In this case it seems that `'` doesn't cause problem because parser tries to sidestep the problem by using end of line for locating end of string. This means that single quotes don't support newline symbol. Just tested using environment variable that contains newline and it indeed broke the parser. This is reasonable compromise for hand written rarun profiles but not good enough when we create one automatically from existing environment which could contain anything.

Rarun also treats doublequotes `"` differently. It uses `r_str_unescape` looks. Just need to choose correct `r_str_escape_*` method.

abcSup:
Looking at the code, `getstr` returns the exact string if it is wrapped between single quotes `'` and does not try to interpret single quotes or other special symbols inside that string again. I find that wrapping the original values with single quotes `'` is enough and there is no need to further escaping any special symbols inside the env variable value.

I understand that you are trying to teach me the proper way to deal with this. What I mean by "enough" is not that I try to ignore what you just said and insist on my own way, but I tried and cannot think of a better way to handle this because there seems to be no way a malicious string could get away with if I just escape it with `'` from the code in `run.c`.

abcSup:
> This means that single quotes don't support newline symbol. Just tested using environment variable that contains newline and it indeed broke the parser. This is reasonable compromise for hand written rarun profiles but not good enough when we create one automatically from existing environment which could contain anything.

Ok, I see it now.

libr/main/radare2.c
```diff
@@ -391,7 +391,17 @@ R_API int r_main_radare2(int argc, const char **argv) {
 	if (e) {
 		RStrBuf *sb = r_strbuf_new (NULL);
 		while (e && *e) {
-			r_strbuf_appendf (sb, "setenv=%s\n", *e);
+			char *k = strdup (*e);
```
trufae, line 394: may be worth moving this logic into a separate function

libr/main/radare2.c
```diff
@@ -391,7 +391,17 @@ R_API int r_main_radare2(int argc, const char **argv) {
 	if (e) {
 		RStrBuf *sb = r_strbuf_new (NULL);
 		while (e && *e) {
```
trufae, line 393: e should not be null, this while can be just `while (*e)`

trufae (dismissed):

libr/main/radare2.c
```diff
@@ -391,7 +391,17 @@ R_API int r_main_radare2(int argc, const char **argv) {
...
-			r_strbuf_appendf (sb, "setenv=%s\n", *e);
+			char *k = strdup (*e);
+			char *v = strchr (k, '=');
+			if (v) {
+				*v++ = 0;
+				v = r_str_escape (v);
```
karliss, line 398: `r_str_escape` strips anything that look like terminal color codes. 

Weird I thought I made this comment twice already but it wasn't here.

abcSup: Thanks again and sorry for sounding a bit rude previously as I did not think of the scenario of having newline as values in env vars. I will look into this again.

trufae:
ping

libr/main/radare2.c
abcSup: I have moved it to a new function called r_run_get_environ_profile

libr/main/radare2.c
abcSup: I have tested color codes and escape sequence and it should work now.

radare (approved):
+1!

karliss (approved):

ret2libc:
@XVilka I did not follow this changes, sorry. If you are ok with them, feel free to merge, there are already 2 approvals.

Issue #13969, opened 2019-05-06
Add * and j in all the missing commands


radare:
For consistency

* [ ] b*  bj
* [ ] y* yj
* [ ] cj c*
* [ ] rj r*
* [ ] P* 
* [ ] pj p* - what do here
* [ ] a* aj - whta info to show?
* [ ] uj u*
* [ ] wj w* - does it make sense?
* [ ] lj l* - json version of ls
* [ ] #!j #!*
* [ ] Lj Llj, ....
* [ ] kj have some newlnes
* [ ] k* is not working

Missing helps

* [ ] v?

jiteshm17:
Hi, 
I am Jitesh. I am a newbie in Open Source and want to work towards solving this issue. However, I am not really able to understand the problem and how to start working on it. Can you please explain in which file of the repository should I start making changes and what exactly I should do?

XVilka:
@jiteshm17 see the `libr/core/cmd_write.c` for `w...` commands for example, `libr/core/cmd_anal.c` for `aoq` command, see `aoj` and `ao*` on how to implement the JSON and r2 commands output, `libr/cmd.c` for more generic commands.

ret2libc:
I'm moving this out of 4.6.0 milestone because fixing *all* commands is really an optimistic view :D

officialcjunior:
What is the desired output for `rj`, if `r` returns, say `5240`?

officialcjunior:
While implementing `uj`, how can I make it compatible with Windows and other OS-s?
For `u`, currently we have something like:
```
#elif __WINDOWS__
	r_cons_printf ("windows\n");
#else
	r_cons_printf ("unknown\n");
```
Does `uj` needs to cover them too? 
If yes, is there a better way to do the check?

trufae:
Yes it should support all platforms. actually i think we shold have uname and uname -j instead of u and uj. the u command should be for undo things. also im not sure if we have tests for these commands. can u add them?

officialcjunior:
> i think we shold have uname and uname -j instead of u and uj

Yes, it makes sense as, `u` is kind of an odd one now under `u?`, but `uname` is already a command in r2.

>  if we have tests for these commands. can u add them?

Aren't these commands system dependent? Should I really add them?

PR #17443, opened 2020-08-11
Implement `rj` ##json

officialcjunior:

**Detailed description**

Issue #13969 suggests to implement `rj`for consistency and standardization.

**Test plan**

Currently, through this commit, I'm not sure that `rj` has been implemented perfectly, as on a test binary, it there's a difference in the outputs, as shown:
```
[0x08048360]> r
7580
[0x08048360]> rj
7.4K
```
Any other high level or specific guidance is hereby requested and appreciated.

- [ ] Add the tests for `rj`, after everything is implemented correctly and ready for review

**Closing issues**

As discussed in #13969

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -266,6 +266,7 @@ static const char *help_msg_k[] = {
 static const char *help_msg_r[] = {
 	"Usage:", "r[+-][ size]", "Resize file",
 	"r", "", "display file size",
+	"rj", "", "display the file size in JSON format",
 	"r", " size", "expand or truncate file to given size",
 	"r-", "num", "remove num bytes, move following data down",
 	"r+", "num", "insert num bytes, move following data up",
@@ -1821,15 +1822,17 @@ static int cmd_resize(void *data, const char *input) {
 			}
 		}
 		return true;
-	case 'j': // "rj"
-		if (core->file) {
+	case 'j': { // "rj"
 			PJ * pj = pj_new ();
 			pj_o (pj);
 			if (oldsize != -1) {
-				pj_n(pj, oldsize);
+				pj_n (pj, oldsize);
 			}
 			pj_end (pj);
-			pj_free (pj);
+			char *s = pj_drain (pj);
+			r_cons_printf ("%s\n", s);
+			free (s);
+			break;
 		}
 	case 'h':
 		if (core->file) {
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -1826,7 +1826,7 @@ static int cmd_resize(void *data, const char *input) {
 			PJ * pj = pj_new ();
 			pj_o (pj);
 			if (oldsize != -1) {
-				pj_n (pj, oldsize);
+				pj_kn (pj, "size", oldsize);
 			}
 			pj_end (pj);
 			char *s = pj_drain (pj);
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -1830,7 +1830,7 @@ static int cmd_resize(void *data, const char *input) {
 			}
 			pj_end (pj);
 			char *s = pj_drain (pj);
-			r_cons_printf ("%s\n", s);
+			r_cons_println (s);
 			free (s);
 			break;
 		}
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -1821,6 +1821,16 @@ static int cmd_resize(void *data, const char *input) {
 			}
 		}
 		return true;
+	case 'j': // "rj"
+		if (core->file) {
+			PJ * pj = pj_new ();
+			pj_o (pj);
+			if (oldsize != -1) {
+				pj_n(pj, oldsize);
+			}
+			pj_end (pj);
+			pj_free (pj);
+		}
 	case 'h':
 		if (core->file) {
 			if (oldsize != -1) {


libr/core/cmd.c
```diff
@@ -1821,6 +1821,16 @@ static int cmd_resize(void *data, const char *input) {
...
+	case 'j': // "rj"
+		if (core->file) {
+			PJ * pj = pj_new ();
+			pj_o (pj);
+			if (oldsize != -1) {
+				pj_n(pj, oldsize);
+			}
+			pj_end (pj);
+			pj_free (pj);
+		}
```
trufae, line 1833: Misisng break

libr/core/cmd.c
```diff
@@ -1821,6 +1821,16 @@ static int cmd_resize(void *data, const char *input) {
...
 		return true;
+	case 'j': // "rj"
+		if (core->file) {
+			PJ * pj = pj_new ();
+			pj_o (pj);
+			if (oldsize != -1) {
+				pj_n(pj, oldsize);
+			}
+			pj_end (pj);
+			pj_free (pj);
```
trufae, line 1832: Drain+print+free the pj string

libr/core/cmd.c
```diff
@@ -1821,6 +1821,16 @@ static int cmd_resize(void *data, const char *input) {
 			}
 		}
 		return true;
+	case 'j': // "rj"
+		if (core->file) {
```
trufae, line 1825: Is this check really necessary?

libr/core/cmd.c
```diff
@@ -1821,6 +1821,16 @@ static int cmd_resize(void *data, const char *input) {
 			}
 		}
 		return true;
+	case 'j': // "rj"
```
trufae, line 1824: Add rj to the help message in r?

libr/core/cmd.c
trufae: Or return

libr/core/cmd.c
officialcjunior: Yeah, I thought so too, but when I remove that, I get a:
```
error: a label can only be part of a statement and a declaration is not a statement
 1826 |   PJ * pj = pj_new ();
```

libr/core/cmd.c
officialcjunior: How about
```
		char *s = pj_drain (pj);
		if (s) {
			r_cons_println (s);
			free (s);
		}
```
after the pj_end?

libr/core/cmd.c
officialcjunior: When I simply put them, `rj` returned: 
`r_io_resize: cannot resize`

libr/core/cmd.c
trufae: if you call pj_drain you dont need to call pj_free,

libr/core/cmd.c
trufae: what do you mean by "put them"?

officialcjunior: Sorry.
What I meant was, when I had used them without the check, inside a curly bracket, that particular error was displayed.

Now, I reckon that, this error is not related to checking for `core->file`.

officialcjunior:
Added the internal help and drain.

Now, `rj` returns the size just fine, but there's one more thing to fix.
```
[0x08048440]> r
3928
[0x08048440]> rj
r_io_resize: cannot resize
{3928}
```

trufae:
Btw {123} is not valid json. the output expected probably needs to display the current filename and size like `{"file":"/bin/ls","size":132}`

and this info is already under `i`, so you can see the code there how it shows that info in json

trufae:
actually i dont see anything wrong in making the output of `rj` be the same as `ij~{core}`.. try it like this:

```
$ r2 -n /bin/ls
> ij~{}
```

the 'core' object inside ij should be the output of `rj`, so we can nest info without duplicating the logic.. `i` shuold be only for printing info from RBin not from RIO imho

officialcjunior:
This looks a bit hard to me, at the moment as it's different from what I've did so far.

Couple of questions :
How exactly does `ij` work? It does not use PJ, does it?
Also, what did you mean by 'so we can nest info without duplicating the logic' ?

trufae:
ij doesnt uses pj.. yet, but everything in r2land generating json should use pj, so any help on that direction is welcome.

for this PR i think we are good if you add the filename info to the output, like ij~{core} do.

about nesting, if we have an object inside a json that is the same as the one generated by another command, this object should be generated by calling the other command instead of having the same code in 2 places doing the same

officialcjunior:
Understood.
How do I do that? Is there an already implemented example from which I can try to learn?

trufae:
If you make ij use the PJ api you can just create a function that takes the PJ instance as argument and build the nested object, this function can then be reused from ij and rj

ret2libc (changes requested):
Hi! Did you check whether the JSON output you generate is a correct json? As in the other PR, you are producing something like `{12345}`, which should not be valid json. A valid json would be something like `{"size": 12345}`. To obtain this second output, you should use `pj_kn ("size", oldsize)` instead of `pj_n`.

officialcjunior:
Replaced `pj_n` with `pj_kn`.
Now I'll try the nesting part.

libr/core/cmd.c
```diff
@@ -1821,6 +1822,18 @@ static int cmd_resize(void *data, const char *input) {
...
 		return true;
+	case 'j': { // "rj"
+			PJ * pj = pj_new ();
+			pj_o (pj);
+			if (oldsize != -1) {
+				pj_kn (pj, "size", oldsize);
+			}
+			pj_end (pj);
+			char *s = pj_drain (pj);
+			r_cons_printf ("%s\n", s);
```
trufae, line 1833: Use r_cons_println instead, no need for formatting.

trufae (approved):

libr/core/cmd.c
officialcjunior: Replaced.

ret2libc (approved):

officialcjunior:
Hey, we still have a problem here, `rj` also prints `r_io_resize: cannot resize` along with it.

```
[0x08048320]> rj
r_io_resize: cannot resize
{"size":5240}
```
And I found two instances of that string:
 https://github.com/radareorg/radare2/blob/549784d73fed420e51c4c4caeeb942806200ab08/libr/core/cmd.c#L1875

How do I fix this?

I apologize for not posting it before as I got carried away with some academic tests.

I'll add the fix to this and the test for this on an another PR.

officialcjunior:
Hey @trufae @ret2libc , we still seem to have a problem here, rj also prints r_io_resize: cannot resize along with it.

```
[0x08048320]> rj
r_io_resize: cannot resize
{"size":5240}
```
Is it an issue? If yes, how can I fix this?
It's getting eprinted right below the place where the switch for `r_` ends.

ret2libc:
> Hey @trufae @ret2libc , we still seem to have a problem here, rj also prints r_io_resize: cannot resize along with it.
> 
> ```
> [0x08048320]> rj
> r_io_resize: cannot resize
> {"size":5240}
> ```
> 
> Is it an issue? If yes, how can I fix this?
> It's getting eprinted right below the place where the switch for `r_` ends.

Yes it is an issue. For how to solve it, I'd like you to come up with it :)
`rj` and `r` should do almost the same things, but one should be in JSON, the other in "regular" mode. So try to look at the code executed by `r` and `rj` and see if you can find some differences.

PR #17451, opened 2020-08-12
Fix 2 last critic error lgtm

08A:

**Detailed description**
I fixed the 2 last critic error discover by lgtm.

1. Is a false positif, cmd is type ut8 and can't overflow because the standart defined his value between 0 and 255. But i add a condition to prevent regression.
2. Add a threshold of 5000 -> https://en.wikipedia.org/wiki/Maximum_transmission_unit

**Test plan**
Run the test suite

**Closing issues**
work on https://github.com/radareorg/radare2/issues/16493

### DIFF
--- a/libr/core/core.c
+++ b/libr/core/core.c
@@ -3154,6 +3154,10 @@ reaccept:
 				}
 				goto out_of_function;
 			}
+			// prevent overflow in uncontrolled allocation size
+			if (cmd > 255) {
+				cmd = 255;
+			}
 			switch (cmd) {
 			case RAP_PACKET_OPEN:
 				r_socket_read_block (c, &flg, 1); // flags
@@ -3161,7 +3165,6 @@ reaccept:
 				r_socket_read_block (c, &cmd, 1); // len
 				pipefd = -1;
 				ptr = malloc (cmd + 1);
-				//XXX cmd is ut8..so <256 if (cmd<RAP_PACKET_MAX)
 				if (!ptr) {
 					eprintf ("Cannot malloc in rmt-open len = %d\n", cmd);
 				} else {
--- a/libr/core/core.c
+++ b/libr/core/core.c
@@ -3154,7 +3154,7 @@ reaccept:
 				}
 				goto out_of_function;
 			}
-			switch ((ut8)cmd) {
+			switch (cmd) {
 			case RAP_PACKET_OPEN:
 				r_socket_read_block (c, &flg, 1); // flags
 				eprintf ("open (%d): ", cmd);
--- a/libr/core/core.c
+++ b/libr/core/core.c
@@ -3154,17 +3154,13 @@ reaccept:
 				}
 				goto out_of_function;
 			}
-			// prevent overflow in uncontrolled allocation size
-			if (cmd > 255) {
-				cmd = 255;
-			}
 			switch (cmd) {
 			case RAP_PACKET_OPEN:
 				r_socket_read_block (c, &flg, 1); // flags
 				eprintf ("open (%d): ", cmd);
 				r_socket_read_block (c, &cmd, 1); // len
 				pipefd = -1;
-				ptr = malloc (cmd + 1);
+				ptr = malloc ((int)cmd + 1);
 				if (!ptr) {
 					eprintf ("Cannot malloc in rmt-open len = %d\n", cmd);
 				} else {
--- a/libr/core/core.c
+++ b/libr/core/core.c
@@ -3160,7 +3160,10 @@ reaccept:
 				eprintf ("open (%d): ", cmd);
 				r_socket_read_block (c, &cmd, 1); // len
 				pipefd = -1;
-				ptr = malloc ((int)cmd + 1);
+				if (UT8_ADD_OVFCHK (cmd, 1)) {
+					goto out_of_function;
+				}
+				ptr = malloc (cmd + 1);
 				if (!ptr) {
 					eprintf ("Cannot malloc in rmt-open len = %d\n", cmd);
 				} else {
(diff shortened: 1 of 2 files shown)


libr/core/core.c
```diff
@@ -3154,14 +3154,17 @@ R_API bool r_core_serve(RCore *core, RIODesc *file) {
 				}
 				goto out_of_function;
 			}
-			switch ((ut8)cmd) {
+			// prevent overflow in uncontrolled allocation size
+			if (cmd > 255) {
+				cmd = 255;
```
trufae, line 3159: cmd is ut8, the only value possible here is 256. and i dont see why it should be handled specially if there's no valid packet for it

libr/socket/socket_http_server.c
```diff
@@ -73,6 +73,9 @@ R_API RSocketHTTPRequest *r_socket_http_accept (RSocket *s, RSocketHTTPOptions *
 				hr->host = strdup (buf + 6);
 			} else if (!strncmp (buf, "Content-Length: ", 16)) {
 				content_length = atoi (buf + 16);
+				if (content_length > 5000) {
```
trufae, line 76: why 5000? why truncating the contents?

libr/core/core.c
08A: Okay so we don't touch this lgtm error.

08A:
imho if we don't need threshold for the first issue and we can't have a threshold for the 2 issue. We can consider the fact that there no critic lgtm error anymore.

libr/core/core.c
karliss: ut8 max value is 255=0xff not 256=0x100 so it can't even be 256.

XVilka:
We probably can just mark them as "false positive" then. Maybe reach LGTM team as well to point this out.

trufae:
Both lgtm are different issues. I wouldnt put them in the same bag.

08A:
> Both lgtm are different issues. I wouldnt put them in the same bag.

What should i do?

08A:
Maybe we could use https://help.semmle.com/lgtm-enterprise/user/help/alert-suppression.html#alert-suppression-examples

ret2libc:
See my 2 comments at https://github.com/radareorg/radare2/issues/16493#issuecomment-680100070

ret2libc:
@08A do you intend to still work on this? I think the content_length issue can be resolved as suggested in https://github.com/radareorg/radare2/issues/16493#issuecomment-680103876 . Maybe the other can be left as is for now, or maybe just forcing the cast as `malloc ((int)cmd + 1)` will make LGTM see an overflow is not possible and remove the warning.

libr/socket/socket_http_server.c
```diff
@@ -105,6 +105,11 @@ R_API RSocketHTTPRequest *r_socket_http_accept (RSocket *s, RSocketHTTPOptions *
 	}
 	if (content_length>0) {
 		r_socket_read_block (hr->s, (ut8*)buf, 1); // one missing byte wtf
+		if (ST32_ADD_OVFCHK (content_length, 1)) {
```
ret2libc, line 108: I think you should create a new `INT_ADD_OVFCHK`, because it is not said that int == st32 everywhere, AFAIK.

libr/core/core.c
```diff
@@ -3154,14 +3154,13 @@ R_API bool r_core_serve(RCore *core, RIODesc *file) {
...
-			switch ((ut8)cmd) {
+			switch (cmd) {
 			case RAP_PACKET_OPEN:
 				r_socket_read_block (c, &flg, 1); // flags
 				eprintf ("open (%d): ", cmd);
 				r_socket_read_block (c, &cmd, 1); // len
 				pipefd = -1;
-				ptr = malloc (cmd + 1);
-				//XXX cmd is ut8..so <256 if (cmd<RAP_PACKET_MAX)
+				ptr = malloc ((int)cmd + 1);
```
ret2libc, line 3163: it seems like this is not enough to make codeql happy. Please add a check similar to the one for content-length with UT8_ADD_OVFCHK.

libr/socket/socket_http_server.c
08A: look at the file libr/include/r_types_base.h
`11 #define st32 int"

I made the same mistake

ret2libc: ah mmm :D ok!

ret2libc (approved):

libr/core/core.c
```diff
@@ -3154,14 +3154,16 @@ R_API bool r_core_serve(RCore *core, RIODesc *file) {
...
+			switch (cmd) {
 			case RAP_PACKET_OPEN:
 				r_socket_read_block (c, &flg, 1); // flags
 				eprintf ("open (%d): ", cmd);
 				r_socket_read_block (c, &cmd, 1); // len
 				pipefd = -1;
+				if (UT8_ADD_OVFCHK (cmd, 1)) {
+					goto out_of_function;
+				}
 				ptr = malloc (cmd + 1);
```
radare, line 3166: missing (size_t)cmd + 1

radare (changes requested):

libr/core/core.c
ret2libc, line 3166:
```suggestion
				ptr = malloc ((size_t)cmd + 1);
```

ret2libc:
I committed the change requested by @trufae myself. Let's see what CI says and let's merge then. Maybe we'll have to flag as false positives some warnings in LGTM/CodeQL if they still come out.

XVilka (approved):

PR #17461, opened 2020-08-13
Variable printf arguments: let compiler help with formatting

anisse:
This generates many warnings, it might not be suitable to merge as-is
without many fixes first.

Attribute supported by both gcc and clang.

 

**Detailed description**

While playing with ESIL, I found that type and argument verification wasn't properly checked at compile time with r_strbuf_appendf ; this will give more compiler warnings in case of wrong arguments size, bad order, forgotten arguments, too many arguments, etc.

**Test plan**

Build, and you'll find many new warnings. There may be bugs there.

**Closing issues**

N/A


XVilka:
It breaks MSVC build because there is no such thing for it: https://ci.appveyor.com/project/radareorg/radare2/builds/34634627/job/p8yamuogkauejlem

anisse:
I can add compiler checking for this. It would be too bad to miss on these checks.

trufae:
For portability reasons i would say to have this wrapped as a macro so we can disable it source-wide in one place. Also may be worth add this to other vararg-format functions. Cool!

XVilka:
You might create something like `R_PACKED` macro to switch this automatically depending on the compiler: https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#packed-structures

anisse:
I might have reached too far, looks like I broke some builds, let me check again.

XVilka:
I recommend to fix these warnings in the same PR - we can then merge commit without squashing.

libr/include/r_types.h
```diff
@@ -210,6 +210,13 @@
   #define FUNC_ATTR_ALWAYS_INLINE
 #endif
 
+/* printf format check attributes */
+#if defined(__clang__) || defined(__GNUC__)
+#define R_PRINTF_CHECK(fmt, dots) __attribute__ ((format (printf, fmt, dots)))
+#else
+#define R_PRINTF_CHECK(format, dots)
```
ret2libc, line 217: use `fmt`, `dots` as in the previous definition.

libr/include/r_util/r_strbuf.h
```diff
@@ -19,12 +19,14 @@ R_API bool r_strbuf_set(RStrBuf *sb, const char *s);
 R_API bool r_strbuf_slice(RStrBuf *sb, int from, int len);
 R_API bool r_strbuf_setbin(RStrBuf *sb, const ut8 *s, size_t len);
 R_API ut8* r_strbuf_getbin(RStrBuf *sb, int *len);
-R_API bool r_strbuf_setf(RStrBuf *sb, const char *fmt, ...);
+R_API bool r_strbuf_setf(RStrBuf *sb, const char *fmt, ...)
```
ret2libc, line 22: I know it becomes a bit longer, but I think putting it on the same line would be better. @trufae opinions?

ret2libc (commented):
I really like this! Overall, this is a strong :+1: from me. Just have some comments, but I'd love to see this integrated as well! Thanks.

anisse:
Thanks for your feedback @ret2libc . For this PR, I was thinking of splitting it in a first patch that adds the define, and maybe a few checks that have no warnings. And in followup PRs, adding the check and associated fixes for each of those functions. I'll fix your comments then.

ret2libc:
@XVilka I'm ok with merging this with just the defines and fixing the reported warnings in separate PRs. The fact that this PR adds a bunch of warnings at compilation time is not @anisse 's fault, so waiting until everything is fixed just hides the real problems. WDYT?

ret2libc:
> Thanks for your feedback @ret2libc . For this PR, I was thinking of splitting it in a first patch that adds the define, and maybe a few checks that have no warnings. And in followup PRs, adding the check and associated fixes for each of those functions. I'll fix your comments then.

Ok ignore what I said in the previous comment. @XVilka made me notice that if we merge this as-is everything will be red because of CodeQL check. So, I guess your plan is a better plan :)

libr/include/r_types.h
anisse: Fixed

libr/include/r_util/r_strbuf.h
anisse: Fixed as well, it will be in a separate PR because it generates many warnings.

anisse:
As discussed, I've tried to put the new warnings in separate PR #17529

anisse:
It seems the macos tests timed-out. Should I rebase on master and retry ?

PR #17424, opened 2020-08-08
Use meson install instead of manual installation in meson.py

yossizap:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

The recent SDB header PR #17249 required an update to meson.py's manual installation paths. This PR replaces the dist building functions in the script with meson's install command to avoid modifying file paths each time include dirs are added.

**Test plan**

Build r2 using sys/meson.py. This solves an issue when building r2 with Cutter so you can also run Cutter's prepare_r2.bat and then run build.bat.

PR #17463, opened 2020-08-13
meson install: Create shortcut/symlink to radare2

kazarmy:

**Detailed description**

This pr recreates r2.bat on Windows under Meson that went missing, probably after #17424. Also enables the r2 tests, and creates a symlink to radare2 on non-Windows Meson.

**Test plan**

All builds should be green.

**Closing issues**

...

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -708,6 +708,4 @@ install_data(
   install_dir: r2_zsh_compdir
 )
 
-if host_machine.system() == 'windows'
-  meson.add_install_script('sys/create_r2.bat')
-endif
+meson.add_install_script(host_machine.system() == 'windows' ? 'sys/create_r2.bat' : 'sys/create_r2.sh')
--- a/meson.build
+++ b/meson.build
@@ -707,3 +707,7 @@ install_data(
   'doc/zsh/_rax2',
   install_dir: r2_zsh_compdir
 )
+
+if host_machine.system() == 'windows'
+  meson.add_install_script('sys/create_r2.bat')
+endif
--- a/sys/create_r2.bat
+++ b/sys/create_r2.bat
@@ -1,0 +1,1 @@
+@echo @"%%~dp0\radare2" %%*> %MESON_INSTALL_PREFIX%\bin\r2.bat
--- a/sys/create_r2.sh
+++ b/sys/create_r2.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-cd "$(MESON_INSTALL_PREFIX)/bin" && ln -fs radare2 r2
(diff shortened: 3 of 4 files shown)


XVilka:
[Breaks linux-meson-newshell](https://github.com/radareorg/radare2/pull/17463/checks?check_run_id=981158583#step:13:66):
```
[XX] db/tools/r2 r2 -L
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e cfg.newshell=false
?q `!!r2 -L~default?`
?+ ?vi $?
?! ?e `!!r2.bat -L~default?`
' -
-- stdout
@@ -1 +1 @@
-1
+0

-- stderr
sh: 1: r2: not found
sh: 1: r2.bat: not found
```

trufae:
sh: 1: r2.bat.covid19: not found

kazarmy:
> Breaks linux-meson-newshell:

I think the reason is that meson's r2 shell script went missing as well

kazarmy:
... or was never there in the first place

kazarmy:
Description updated. The newshell incompatibility should be fixed in a later pr (things become strange at the edges of r2land).

ret2libc:
@kazarmy you can use log.level=1 to get some more info if something is wrong with cfg.newshell

ret2libc:
If you won't work on that in the short time, please file an issue and assign that to me so I can have a look next week

karliss:
Assuming the purpose of most of those tests is testing r2 command line flags not the r2 symlink or r2.bat couldn't they use the long executable name `radare2`?  Wouldn't that solve the naming difference between Windows and other platforms?

kazarmy:
> Assuming the purpose of most of those tests is testing r2 command line flags not the r2 symlink or r2.bat couldn't they use the long executable name radare2? Wouldn't that solve the naming difference between Windows and other platforms?

Yes they could use `radare2` but now they are currently doing double duty. This is not ideal ofc, 1 reason being that a test could be modified without the modifier being aware of what duties the test is currently performing, but I don't think it's likely that:

1. The tests will be modified.
2. `r2.bat` will go missing again.

kazarmy:
> If you won't work on that in the short time, please file an issue and assign that to me so I can have a look next week

Ok I'll file the issues after #17488 goes through

kazarmy:
@ret2libc All yours #17494 #17495 

Regarding the commands, they could be buggy in the old parser as well and they certainly have the feel of being well off the beaten path, but then I needed only one working combination for the tests.

PR #17462, opened 2020-08-13
Add minimal armv7 and aarch32 VFP and NEON support ##esil

anisse:
The register file starts at offset 68, after the flags.

We bootstrap instructions with register dumping and loading.

 

**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [X] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

VFP and NEON instructions for ARMv7/AARCH32 are missing in ESIL.

**Test plan**

See tests for newly supported armv7 instructions in ESIL.

**Closing issues**

N/A

PR #17473, opened 2020-08-16
Bring merge sort and insertion sort cmp function semantics together

anisse:
Merge sort uses cmp (a, b) < 0 for its first test branch, and insertion
sort cmp (a, b) > 0 ; which means the 0 boundary goes in one case in one
branch, and in the other sort function in the other branch.

We keep the semantics of the insertion sort, because it allows stability between
the two sort functions for equal elements.

Update tests that were broken because of wrong register ordering.

 

**Detailed description**

Register order changed on arm32 after adding more register at the end: everything was seemingly in a random order. It's because the list went over 43 elements, and started using merge sort instead of insertion sort. This PR make merge sort behave properly with compare functions that return a bool.

**Test plan**

...

**Closing issues**

This should unblock test regression from PR #17462


shlr/sdb/src/ls.c
```diff
@@ -45,7 +45,7 @@ static SdbListIter *_merge(SdbListIter *first, SdbListIter *second, SdbListCompa
 		} else if (!first) {
 			next = second;
 			second = second->n;
-		} else if (cmp (first->data, second->data) < 0) {
+		} else if (cmp (first->data, second->data) <= 0) {
```
ret2libc, line 48: @xvilka this should not be merged here, it should be changed in sdb code base first and then sdb needs to be updated in r2.

anisse: You mean two different commits instead of updating both at the same time like here ?

ret2libc: Sdb is actually taken from another repo, see https://github.com/radareorg/sdb . Changes to sdb files should first be merged there and then you can use `make sdb-sync` to update sdb

anisse: I sent a separate PR for sdb and removed this.

anisse:
I rebased on master and removed sdb modifications after sending https://github.com/radareorg/sdb/pull/212

XVilka:
Please rebase on top of the master and sync SDB here since that PR was merged.

XVilka:
Seems it [has broken the reverse debugger](https://github.com/radareorg/radare2/pull/17473/checks?check_run_id=1003414481):
```
[XX] db/archos/linux-x64/dbg_step_back dbg.stepback
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -d -e dbg.bpsysign=true -Qc 'db main
db 0x004028fe
dc
dts+
dc
dsb
dsb
dr rbx,rcx,rdx,r12,rip
dk 9
' bins/elf/analysis/ls-linux-x86_64-zlul
-- stdout
@@ -1,5 +1,5 @@
 0x00000001
 0x00000001
-0x00000001
+0x0000a401
 0x00404870

-- stderr
Process with PID 33840 started...
= attach 33840 33840
bin.baddr 0x00400000
Using 0x400000
asm.bits 64
hit breakpoint at: 0x4028a0
Reading 4096 byte(s) from 0x0061c000...
Reading 4096 byte(s) from 0x0061d000...
Reading 135168 byte(s) from 0x02466000...
Reading 4096 byte(s) from 0x7fa88960b000...
Reading 16384 byte(s) from 0x7fa88960c000...
Reading 4096 byte(s) from 0x7fa889814000...
Reading 4096 byte(s) from 0x7fa889a18000...
Reading 4096 byte(s) from 0x7fa889c89000...
Reading 8192 byte(s) from 0x7fa88a075000...
Reading 16384 byte(s) from 0x7fa88a077000...
Reading 4096 byte(s) from 0x7fa88a282000...
Reading 4096 byte(s) from 0x7fa88a4a8000...
Reading 4096 byte(s) from 0x7fa88a282000...
Reading 4096 byte(s) from 0x7fa88a4a8000...
Reading 8192 byte(s) from 0x7fa88a4a9000...
Reading 28672 byte(s) from 0x7fa88a6b9000...
Reading 4096 byte(s) from 0x7fa88a6d3000...
Reading 4096 byte(s) from 0x7fa88a6d4000...
Reading 135168 byte(s) from 0x7ffe012d9000...
r_reg_get_value: Bit size 256 not supported
r_reg_get_value: Bit size 256 not supported
r_reg_get_value: Bit size 256 not supported
r_reg_get_value: Bit size 256 not supported
r_reg_get_value: Bit size 256 not supported
r_reg_get_value: Bit size 256 not supported
r_reg_get_value: Bit size 256 not supported
r_reg_get_value: Bit size 256 not supported
r_reg_get_value: Bit size 256 not supported
...
```

anisse:
Yes, I didn't notice it at first because the test is also failing on my machine (Fedora 32), but differently. I saw the same result before and an after the PR. I'll install Ubuntu to try to reproduce it.

anisse:
I think it may be because I don't have AVX extensions on my (old) CPU, see logs of a github action on my account, with additional debug info:
```
[…]
r_reg_get_value: Bit size 256 not supported for reg ymm10 
r_reg_get_value: Bit size 256 not supported for reg ymm11 
r_reg_get_value: Bit size 256 not supported for reg ymm12 
r_reg_get_value: Bit size 256 not supported for reg ymm13 
r_reg_get_value: Bit size 256 not supported for reg ymm14 
[…]
```

https://github.com/anisse/radare2/pull/3/checks?check_run_id=1005021519

Now I need to find a machine with those or debug through github actions :-/

anisse:
Apparently glibc has a function to do a faster memset using avx512: vzeroupper. That's what the trace debugger is hitting, trying to get the value of the ymm0-15 registers, and failing because this isn't implemented.

I'm pretty sure this PR accidentally fixed something (I don't know what) which is now making this trace debugger test fail. I wouldn't mind a second pair of eyes for this.

I could add a fake case for getting the value of a 256bits register, just like 128bits (which isn't really implemented). What do you think ?

ret2libc:
I'm not sure about this. By definition in r_list.h, RListComparator should return -1, 0, 1, so I think it is just wrong to pass a RListComparator that returns a bool. IMHO it is wrong to use `cmp_order` as in test_list.c and assume to have good results by returning a bool. @thestr4ng3r what do you think?

anisse:
> I'm not sure about this. By definition in r_list.h, RListComparator should return -1, 0, 1, so I think it is just wrong to pass a RListComparator that returns a bool. IMHO it is wrong to use `cmp_order` as in test_list.c and assume to have good results by returning a bool. @thestr4ng3r what do you think?

I've thought about this as well, but decided against this, for the following reasons:

- there are comparators that rely on the "wrong" behaviour. And this behaviour works for insertion sort.
- both insertion sort and merge sort should have the same behaviour
- historically, there was only insertion sort, and merge sort was added later, with a different semantic at the 0 boundary. The "definition" in r_list.h was also added later, without fixing all the comparators.

I could provide a different PR that changes the comparators, but IMHO it would require changing the insertion sort as well, to stop working as it always did, since the "code is the API", and the behaviour should be the same for both functions to avoid any surprise once the list size changes; ideally this should be propagated

Right now there are lists in the code that aren't properly sorted: I've found the register lists for x86, aarch64, and the new one for arm32 in PR #17462 .

ret2libc:
> Right now there are lists in the code that aren't properly sorted: I've found the register lists for x86, aarch64, and the new one for arm32 in PR #17462 .

I think they are not properly sorted *because* of this bool/int mess. If you convert the comparator functions to use -1/0/1, they should work. 


> > I'm not sure about this. By definition in r_list.h, RListComparator should return -1, 0, 1, so I think it is just wrong to pass a RListComparator that returns a bool. IMHO it is wrong to use `cmp_order` as in test_list.c and assume to have good results by returning a bool. @thestr4ng3r what do you think?
> 
> I've thought about this as well, but decided against this, for the following reasons:
> 
>     * there are comparators that rely on the "wrong" behaviour. And this behaviour works for insertion sort.

The fact that it works is just a side effect, IMO. However, in general in C comparator functions like `strcmp` and similar return -1/0/1 (a negative, 0, positive number actually), this is why having the same semantics seems better to me.

> 
>     * both insertion sort and merge sort should have the same behaviour

I agree, and I think they do, if the right comparator function (that is one that returns -1/0/1) is used in both cases. If merge or insertion sort don't work properly, provided with a -1/0/1 comparator function, than that is definitely a bug.

>     * historically, there was only insertion sort, and merge sort was added later, with a different semantic at the 0 boundary. The "definition" in r_list.h was also added later, without fixing all the comparators. It was also probably added as consequence of this issue, without fixing the core issue.

I didn't dig into the history, but the different semantic at the 0 boundary is probably only a "small issue" and probably should affect only the relative sorting of two elements with the same value according to `cmp`. Anyway, RListComparator is used not only in insertion/merge_sort and the same comparator function should work well in all those functions (e.g. r_list_sort, but also r_list_find). A comparator function that returns `a > b` would return `false` both when `a == b` and when `a < b`. False is interpreted as 0 and `r_list_find` uses `!cmp` to check whether an element was found in a list. So `r_list_find` would not work correctly. I think this is just one example, but in general I do not see it as a good thing to have a boolean disguised as a int. 

All this to say that if there are problems when moving from insertion to merge sort, it is probably just because the comparator function passed to insertion_sort abused the way the sorting is implemented, but *those* are actually wrong. We should change those comparator functions. You can also see that you are actually casting a boolean (`a > b`) to int (which is the return value of `RListComparator`) and this usually should raise a red flag. That said, it is also ok to ensure that the behaviour when two elements are evaluated to be the same (RListComparator returns 0) is consistent between merge and insertion sort.

anisse:
> > ```
> > * historically, there was only insertion sort, and merge sort was added later, with a different semantic at the 0 boundary. The "definition" in r_list.h was also added later, without fixing all the comparators. It was also probably added as consequence of this issue, without fixing the core issue.
> > ```
> 
> I didn't dig into the history, but the different semantic at the 0 boundary is probably only a "small issue" and probably should affect only the relative sorting of two elements with the same value according to `cmp`. Anyway, RListComparator is used not only in insertion/merge_sort and the same comparator function should work well in all those functions (e.g. r_list_sort, but also r_list_find). A comparator function that returns `a > b` would return `false` both when `a == b` and when `a < b`. False is interpreted as 0 and `r_list_find` uses `!cmp` to check whether an element was found in a list. So `r_list_find` would not work correctly. I think this is just one example, but in general I do not see it as a good thing to have a boolean disguised as a int.

It's not a small issue, once merge sort is used, lists aren't sorted at all. You can look at the test fixes to see how the order changes.

> 
> All this to say that if there are problems when moving from insertion to merge sort, it is probably just because the comparator function passed to insertion_sort abused the way the sorting is implemented, but _those_ are actually wrong. We should change those comparator functions. You can also see that you are actually casting a boolean (`a > b`) to int (which is the return value of `RListComparator`) and this usually should raise a red flag. That said, it is also ok to ensure that the behaviour when two elements are evaluated to be the same (RListComparator returns 0) is consistent between merge and insertion sort.

I agree it's a comparator bug, but this PR has the advantage of entirely fixing this bug class. It will reappear again, because casting the function once to (RListComparator) is slightly easier than casting both const void*.

Here is my proposal:
 - keep this behaviour change for merge sort, but also fix the bad RListComparators. I've found about ~14 entries, I'll fix them in a separate patch in the PR.
 - I'll remove the test to show that it isn't acceptable use of r_list_sort

anisse:
I've updated the PR with my proposal and changed the main description.

libr/core/agraph.c
```diff
@@ -1332,11 +1332,11 @@ static void place_single(const RAGraph *g, int l, const RGraphNode *bm, const RG
 }
 
 static int RM_listcmp(const struct len_pos_t *a, const struct len_pos_t *b) {
-	return a->pos < b->pos;
+	return (a->pos < b->pos) - (a->pos > b->pos);
 }
 
 static int RP_listcmp(const struct len_pos_t *a, const struct len_pos_t *b) {
-	return a->pos >= b->pos;
+	return (a->pos >= b->pos) - (a->pos < b->pos);
```
ret2libc, line 1339:
```suggestion
	return (a->pos > b->pos) - (a->pos < b->pos);
```
or am i missing anything?

ret2libc:
> I've updated the PR with my proposal and changed the main description.

Thanks! Much better now ;) Let's wait for CI. Please just fix that small comment (unless there is a reason to keep it like that), then it's ok for me.

libr/core/agraph.c
anisse: This one I wasn't sure about because of the weird compare. It's used in r_list_add_sorted. I kept it as-is to prevent any regression, but I can also remove it entirely, because I don't know how to test it properly.

anisse: By "weird compare", I mean that is used as a mirror to RM_listcmp with < vs >=. I think what I did would keep the behaviour closest to what was previously done.

ret2libc: I'm not sure it would. The cmp function should return 0 in case the elements are the same, but with this it does not.

anisse: Fixed

test/db/anal/vars
```diff
@@ -151,8 +151,8 @@ afv
 EOF
 EXPECT=<<EOF
 arg int64_t arg3 @ r8
-arg int64_t arg2 @ rdx
```
ret2libc, line 154: do you know how these arguments are sorted? and why the sorting has changed here?

test/db/archos/linux-x64/dbg_drt
```diff
@@ -135,188 +135,188 @@ CMDS=<<EOF
 drt all~[0]
 EOF
 EXPECT=<<EOF
-ymm15
```
ret2libc, line 138: same here. Do you know why?

test/db/anal/vars
anisse: Yes, there are sorted by register index, which corresponds to the register order. x86-64 registers where sorted by merge sort, and where in a semi-random order. Now it matches the register order.

ret2libc: Mmm could be, though it seems also that it tries to list things in the order of the ABI, could it be? (except for r8/r9 which seems to be wrong).

test/db/archos/linux-x64/dbg_drt
ret2libc: In this case in particular it seems even worse, as there are registers that seem completely out of place (e.g. `cwd` in between `r15w` and `r15d`). What's happening?

anisse: That's because r15w, r15d and cwd are all at offset 0 in libr/anal/p/anal_x86_cs.c: line ~3519 and ~3577. That's the fourth column of the ESIL register file description.

test/db/anal/vars
anisse: That's also the register order. r8 is at offset 72, rcx 88, and rdx 96.

libr/core/cconfig.c
```diff
@@ -51,28 +51,28 @@ static void print_node_options(RConfigNode *node) {
 }
 
 static int compareName(const RAnalFunction *a, const RAnalFunction *b) {
-	return a && b && a->name && b->name && strcmp (a->name, b->name);
+	return (a && b && a->name && b->name ?  strcmp (a->name, b->name) : 0);
 }
 
 static int compareNameLen(const RAnalFunction *a, const RAnalFunction *b) {
-	return a && b && a->name && b->name && strlen (a->name) > strlen (b->name);
+	return (a && b && a->name && b->name  ? strlen (a->name) - strlen (b->name) : 0);
```
ret2libc, line 58: this and all the following functions should use the usual body: `(a > b) - (a < b)`

ret2libc (changes requested):

libr/core/cconfig.c
anisse: Fixed and rebased on master.

trufae (commented):
Change the commit title and add ##anal i think this is an important change and it shuold be in the changelog, the message should make clear the function variable and reflines order is what is mainly affected. Do you have a screenshot of the change in reflines after this change? is the change in sdb correct or should be changed again because of the cmp <=

libr/core/cconfig.c
```diff
@@ -51,28 +51,40 @@ static void print_node_options(RConfigNode *node) {
...
 
 static int compareName(const RAnalFunction *a, const RAnalFunction *b) {
-	return a && b && a->name && b->name && strcmp (a->name, b->name);
+	return (a && b && a->name && b->name ?  strcmp (a->name, b->name) : 0);
 }
 
 static int compareNameLen(const RAnalFunction *a, const RAnalFunction *b) {
-	return a && b && a->name && b->name && strlen (a->name) > strlen (b->name);
+	size_t la, lb;
+	if (!a || !b || !a->name || !b->name)
```
ret2libc, line 59: we always use parentheses.

libr/core/cconfig.c
```diff
@@ -51,28 +51,40 @@ static void print_node_options(RConfigNode *node) {
...
+		return 0;
+	la = strlen(a->name);
```
ret2libc, line 61: and spaces after function
```suggestion
	la = strlen (a->name);
```

libr/core/cconfig.c
```diff
@@ -51,28 +51,40 @@ static void print_node_options(RConfigNode *node) {
...
-	return a && b && a->diff->type && b->diff->type && a->diff->type > b->diff->type;
+	return (a && b && a->diff->type && b->diff->type ?
+			(a->diff->type > b->diff->type) - (a->diff->type < b->diff->type) : 0);
 }
 
 static int compareSize(const RAnalFunction *a, const RAnalFunction *b) {
+	ut64 sa, sb;
 	// return a && b && a->_size < b->_size;
-	return a && b && r_anal_function_realsize (a) > r_anal_function_realsize (b);
+	if (!a || !b)
```
ret2libc, line 78: same here

test/db/cmd/types
```diff
@@ -1334,9 +1334,9 @@ var char ** var_20h @ rbp-0x20
 var int64_t var_14h @ rbp-0x14
 var void * var_10h @ rbp-0x10
 var int64_t var_4h @ rbp-0x4
-arg int argc @ rdi
```
ret2libc, line 1337: I still think there is something wrong here. I would expect variables and arguments to be shown in the "right order", not just sorted by register names. I know this depends on aBI, etc. but it kinda made sense to have `rdi`, `rsi`, `rdx`, while I find this new sorting inappropriate in this particular context. @XVilka @thestr4ng3r @kazarmy WDYT?

anisse:
> Change the commit title and add ##anal i think this is an important change and it shuold be in the changelog, the message should make clear the function variable and reflines order is what is mainly affected.

Should it be for all commits ?

> Do you have a screenshot of the change in reflines after this change?

I'm not sure how to test, but here is a small example; before:
```
$ ./binr/radare2/radare2 -e "asm.describe=false" -e "scr.color=0" -Qc "pd 12 @ 0x00065f70" test/bins/mach0/Alamofire-stripped
            0x00065f70      81040f58       ldr x1, 0x84000
            0x00065f74      60070094       bl 0x67cf4
        ┌─< 0x00065f78      e00000b4       cbz x0, 0x65f94
       ┌──> 0x00065f7c      1f0013eb       cmp x0, x19
      ┌───< 0x00065f80      80000054       b.eq 0x65f90
      │╎│   0x00065f84      ff060094       bl 0x67b80
      │└──< 0x00065f88      a0ffffb5       cbnz x0, 0x65f7c
      │┌──< 0x00065f8c      02000014       b 0x65f94
      └───> 0x00065f90      e0030032       orr w0, wzr, 1
       └└─> 0x00065f94      fd7b41a9       ldp x29, x30, [sp, 0x10]
            0x00065f98      f44fc2a8       ldp x20, x19, [sp], 0x20
            0x00065f9c      c0035fd6       ret
```

after:

```
$ ./binr/radare2/radare2 -e "asm.describe=false" -e "scr.color=0" -Qc "pd 12 @ 0x00065f70" test/bins/mach0/Alamofire-stripped
            0x00065f70      81040f58       ldr x1, 0x84000
            0x00065f74      60070094       bl 0x67cf4
        ┌─< 0x00065f78      e00000b4       cbz x0, 0x65f94
       ┌──> 0x00065f7c      1f0013eb       cmp x0, x19
      ┌───< 0x00065f80      80000054       b.eq 0x65f90
      │╎│   0x00065f84      ff060094       bl 0x67b80
      │└──< 0x00065f88      a0ffffb5       cbnz x0, 0x65f7c
      │┌──< 0x00065f8c      02000014       b 0x65f94
      └───> 0x00065f90      e0030032       orr w0, wzr, 1
       └└─> 0x00065f94      fd7b41a9       ldp x29, x30, [sp, 0x10]
            0x00065f98      f44fc2a8       ldp x20, x19, [sp], 0x20
            0x00065f9c      c0035fd6       ret
```
(no change)

> is the change in sdb correct or should be changed again because of the cmp <=

Yes, I've made sure we keep the same behaviour with sdb.

test/db/cmd/types
anisse: I agree, the argument order feels reversed. I'm not sure how to address that here though.

libr/core/cconfig.c
anisse: Thanks, I did it for the other function, but forgot strlen.

libr/core/cconfig.c
anisse: I'll add the braces for the if.

libr/core/cconfig.c
anisse: Thanks, fixed.

anisse:
> > Change the commit title and add ##anal i think this is an important change and it shuold be in the changelog, the message should make clear the function variable and reflines order is what is mainly affected.
> 
> Should it be for all commits ?

I did it for commits that impact sort order and test fixes.

test/db/cmd/types
ret2libc: this needs a little bit more investigation. Maybe it's just the way it is, but maybe we are missing something.

libr/reg/reg.c
```diff
@@ -200,7 +200,7 @@ R_API void r_reg_free_internal(RReg *reg, bool init) {
 static int regcmp(RRegItem *a, RRegItem *b) {
 	int offa = (a->offset * 16) + a->size;
 	int offb = (b->offset * 16) + b->size;
-	return offa > offb;
+	return (offa > offb) - (offa < offb);
```
ret2libc, line 203: I believe the problem of the registers in the wrong order is because of this.
Try this:
```diff
-       int offa = (a->offset * 16) + a->size;
-       int offb = (b->offset * 16) + b->size;
-       return (offa > offb) - (offa < offb);
+       int off_diff = (a->offset < b->offset) - (a->offset > b->offset);
+       if (off_diff) {
+               return off_diff;
+       }
+       return (a->size > b->size) - (a->size < b->size);
```

anisse: It's just sorted differently (registers in reverse order by offset, smallest size first), I'm not convinced it's a correct solution. Look at how the zignatures change for example. And for architectures that have less than 43 registers, it will be reversed. IMHO sorting the anal vars by name on print might be better. Don't forget the previous order was accidental.

Another possiblity: assign "delta" of register anal vars to be specific to the current call convention, and not derivative of the register offset.

ret2libc: Yes it just stores the registers in reverse order, but it seems to me that's more or less what was intended before (though ofc the ordering was wrong at the end).

> And for architectures that have less than 43 registers, it will be reversed.

Why? It should behave the same for all number of registers.

> IMHO sorting the anal vars by name on print might be better

No, not by name because you want to see what is the actual order of the vars/argument. You want to know that the first argument you see is actualy the first argument passed to the function.

> assign "delta" of register anal vars to be specific to the current call convention, and not derivative of the register offset.

I think this is the proper solution long term, yes.

> I'm not convinced it's a correct solution. Look at how the zignatures change for example. 

I will have a look, but I'm not convinced of current PR's output either :(

anisse: If this fix is appropriate for the x86_64 call conventions, that's fine for me. My initial motivation for this PR was to keep the zignatures stable for arm32 after PR #17462 ; I then discovered the discrepancy between the two sorts, and attempted to fix it. It would have been shorter to just fix the zignatures in the previous PR. I don't have a preference as I don't rely on this feature.

Note: reversing the register order, will just change the zignatures for all architectures, but that's fine if it's a one-time thing, I guess.

anisse: Also, for architectures that pass arguments in an other order (e.g arm32/aarch64), the argument order will be reversed.

ret2libc: > Also, for architectures that pass arguments in an other order (e.g arm32/aarch64), the argument order will be reversed.

good point!

I did not intend to slow you down, but what sometimes seems like a simple fix then risk breaking other stuff, that's why I was worried about the tests changes (and way too often we accidentally fix tests in the wrong way :( )

Anyway, it seems there is no simple fix here, so I'm ok as it is. Thanks!

ret2libc (approved):

libr/reg/reg.c
anisse: Please note that I really appreciated your reviews. Maintaining constant high quality in an open-source project is very hard, and I appreciate you taking the time to look at this PR.

ret2libc:
@XVilka @thestr4ng3r @trufae someone please have another look at this and merge if you think it's ok!

XVilka (approved):
The comparison change looks good to merge. What looks wrong is the reversed order of the arguments/etc like @ret2libc pointed out.

test/db/archos/linux-x64/dbg_drt
```diff
@@ -135,188 +135,188 @@ CMDS=<<EOF
...
+xmm0h
+xmm0
+ymm5
+fs
+xmm0l
+gs
+xmm1h
+xmm1
+ds
+xmm1l
```
radare, line 265: i think this order is messed up

test/db/archos/linux-x64/dbg_drt
```diff
@@ -135,188 +135,188 @@ CMDS=<<EOF
...
 r13
+dr2
+frdp
 r12b
 r12w
 r12d
+mxcsr
 r12
+dr3
+mxcr_mask
```
radare, line 166: random order :?

radare (changes requested):

ret2libc:
@XVilka @trufae as said in previous comments, we think those "random" orders are just because they were never really sorted in the first place. If you look at those tests, the original order was not really better than the new one. It seems like somehow they seemed sorted, but there is no real sorting underneath. Probably those commands should sort the things before printing them themselves.

test/db/archos/linux-x64/dbg_drt
anisse: This corresponds to the offset of the register: dr3 is at 24 and mxcr_mask at 28:
dr3: https://github.com/radareorg/radare2/blob/master/libr/anal/p/anal_x86_cs.c#L3576
mxcr_mask: https://github.com/radareorg/radare2/blob/master/libr/anal/p/anal_x86_cs.c#L3604

anisse: Sorry, wrong links, this is the file: https://github.com/radareorg/radare2/blob/2dfa75cc470261e45030fc4dac1fdba2c57d3dcd/libr/debug/p/native/linux/reg/linux-x64.h (but same offsets)

test/db/archos/linux-x64/dbg_drt
```diff
@@ -135,188 +135,188 @@ CMDS=<<EOF
 drt all~[0]
 EOF
 EXPECT=<<EOF
-ymm15
-ymm14
-ymm13
-ymm12
-ymm11
```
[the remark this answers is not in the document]
anisse: Considering the original functions wants to sort in order, this is the issue this PR fixes:
https://github.com/radareorg/radare2/blob/master/libr/reg/reg.c#L200-L203

You can verify this by always using insertion sort instead here:
https://github.com/radareorg/radare2/blob/master/libr/util/list.c#L573-L577

test/db/archos/linux-x64/dbg_drt
anisse: It is, but that's because the register offsets are the same for xmm1l and ds : 184
https://github.com/radareorg/radare2/blob/2dfa75cc470261e45030fc4dac1fdba2c57d3dcd/libr/debug/p/native/linux/reg/linux-x64.h#L159-L161
https://github.com/radareorg/radare2/blob/2dfa75cc470261e45030fc4dac1fdba2c57d3dcd/libr/debug/p/native/linux/reg/linux-x64.h#L101

trufae (approved):

trufae:
lgtm

ret2libc:
@anisse can you make the PR ready if it's ok to merge? Just to be sure you don't intend to add anything to this. Thanks again for this!

PR #17476, opened 2020-08-16
Fix race condition and remove unnecessary SDB build targets ##build

trufae:

**Detailed description**

sdb_version.h is removed in make clean step which happens when runnig those two targets in parallel.

**Test plan**

make -j

**Closing issues**

none

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ endif
 endif
 
 all: plugins.cfg libr/include/r_version.h
-	${MAKE} -C shlr targets
+	${MAKE} -C shlr sdbs
 	${MAKE} -C shlr/zip
 	${MAKE} -C libr/util
 	${MAKE} -C libr/socket
--- a/libr/util/Makefile
+++ b/libr/util/Makefile
@@ -64,7 +64,6 @@ ifeq (${BUILD_OS},haiku)
   LDFLAGS+=-lexecinfo
 endif
 
-EXTRA_PRE+=sdb_version
 EXTRA_PRE+=spp_config
 
 include ../rules.mk
--- a/libr/util/sdb.mk
+++ b/libr/util/sdb.mk
@@ -1,6 +1,7 @@
 SDBPATH=../../shlr/sdb/src/
 SDBLIB=${SDBPATH}/libsdb.a
 EXTRA_TARGETS+=${SDBLIB}
+EXTRA_PRE+=$(SDBLIB)
 
 SDB_OBJS=
 SDB_OBJS+=buffer.o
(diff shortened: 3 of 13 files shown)


XVilka:
Maybe we can call `make -j` in one of our CI workers to catch such problems automatically.

trufae:
The actual race condition was fixed already i find out after rebasing my master branch, this PR aims to remove unused/duplicated/unnecessary logic related to building the SDB library and native executable. im testing this branch from 3rd party projects. and make -j is already tested, but the problem appears when calling it from different source paths and using it differently from what we are using in CI. So i dont think we need to test that in here.

Let me fix the PR title. wait for me to remove the WIP: when i verify the fix is ready

trufae:
Ok i can confirm tthe race condition is still there :( guess it was also introduced in the sdb refactoring pr because make -j8 was working fine before

ret2libc:
Isn't sdb_version.h generated at build time? If so, ahouldnt it also be removed by make clean?

trufae:
Should be good to go now, the race condition is gone. But i'm still doing some tests on this branch. Btw as long as SDB may be distributed as a shared lib every single module should link against it, not just libr_util. unless we go for dynamic linking. thats kind of a pain, but wont move yet because there are still build regressions to fix

wargio:
Why j1?

trufae:
Wargio: -j1 was a way to proof this was a race condition
Ret2libc: yes. And this is why it was failing. And removing this file is not a task of anyone else than the makefiles inside the sdb repo

wargio:
@trufae how i can re-create the race condition on the build?

trufae:
I am reproducing it by calling sys/sdk.sh which is doing make -j8 on a custom CI machine that have quite a lot CPUs. But the issue is clear by checking the diff. I can show you build error logs and the explain which lines in the diff cause the problem.

* On ubuntu20 make -j8 with a machine with more than 4 cores may be a good candidate to repro the issue
* Running sys/sdk.sh
* See the commits in the sdb repo where main.o depends on sdb_version.h and its built in parallel
* Seems like some of those races wasnt reproducible on ubuntu18.

```
=== Building r2
Cloning into 'radare2'...
cp: './plugins.cfg' and 'plugins.cfg' are the same file
In file included from cdb_make.c:5:
sdb.h:19:10: fatal error: sdb_version.h: No such file or directory
   19 | #include "sdb_version.h"
      |          ^~~~~~~~~~~~~~~
```

btw i can confirm this PR fixes the 2 race conditions and cleanups the build a little bit. So it's ready to merge for me.

trufae:
And this is in macOS right now with master 🤦 using sys/install.sh

Screenshot 2020-08-18 at 20 13 38

XVilka (approved):

wargio (approved):

PR #17481, opened 2020-08-17
Dont assert when no calling convention defined ##anal

trufae:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/anal/cc.c
+++ b/libr/anal/cc.c
@@ -1,4 +1,4 @@
-/* radare - LGPL - Copyright 2011-2019 - pancake, Oddcoder */
+/* radare - LGPL - Copyright 2011-2020 - pancake, Oddcoder */
 
 /* Universal calling convention implementation based on sdb */
 
@@ -136,6 +136,9 @@ R_API int r_anal_cc_max_arg(RAnal *anal, const char *cc) {
 	static char *oldCC = NULL;
 	static int oldArg = 0;
 	if (!cc) {
+		if (anal->verbose) {
+			eprintf ("Warning: r_anal_cc_max_arg: cc == NULL\n");
+		}
 		return 0;
 	}
 	if (oldDB == DB && !strcmp (cc, oldCC)) {
--- a/libr/anal/cc.c
+++ b/libr/anal/cc.c
@@ -131,10 +131,13 @@ R_API const char *r_anal_cc_error(RAnal *anal, const char *convention) {
 
 R_API int r_anal_cc_max_arg(RAnal *anal, const char *cc) {
 	int i = 0;
-	r_return_val_if_fail (anal && DB && cc, 0);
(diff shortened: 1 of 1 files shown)


thestr4ng3r:
Do you have some context about from where passing NULL here would make valid sense?

thestr4ng3r:
Does it really make sense to make passing NULL valid here? Wouldn't it make more sense to check at the callsites?

trufae:
Static builds of r2 dont have the calling conventions because there were moved from inside r2 to external SDB files. This is a known issue for a very long time, at least for me, i know the analysis of arguments will be wrong when no CC is found, but having a runtime assert for something that is loaded at runtime is a bad idea imho

XVilka:
- https://gcc-renesas.com/manuals/SH-ABI-Specification.html
- https://devblogs.microsoft.com/oldnewthing/20190820-00/?p=102792

thestr4ng3r:
Sure, there can always be cases where you don't have a cc. But the question is where exactly to check this. If the assert is removed there, meaning NULL is a "valid cc" for this API, then the asserts in all the other functions in this file must be removed too.

ret2libc:
So, what have we chosen on this?

I'm personally against considering cc=NULL a valid cc, because when cc is null there is probably nothing we can really do, and probably a lot of features (all the vars/args stuff, etc.) can't work at all. Having cc=NULL as a valid cc would just mean that we "ignore" the problem and pretend everything is alright, when it actually isn't. I think we should find a better solution for this.

radare:
R2 should have at least one CC per-arch hardcoded, not depending on .sdb files to work imho. i agree that getting rid of this "warning" is not a good idea, but maybe it should be an eprintf instead of a soft assert. that can be hidden behind anal.verbose

kazarmy:
Pedantic me says that one should not even think of passing a NULL cc to a function named `r_anal_cc_arg` / `r_anal_cc_max_arg`. Practical me says that one should not be so pedantic.

trufae:
Uhm, i see this warning now. shouldnt this pr get merged to be consistent?

```
$ r2 -a tms320 -
WARNING: No calling convention defined for this file, analysis may be inaccurate.
 -- Prove you are a robot to continue ...
[0x00000000]> 

```

ret2libc:
@trufae do you see double warnings?

trufae:
No, asserts are not happening anymore. But imho we should cover all generic calling conventrions inside r2 without depending on loading runtime sdb files

thestr4ng3r:
One option would be to automatically try to derive a fallback cc from the register profile.

trufae:
Yeah that would be easy to do and work in some cases. But ideally we may be able to embed any arbitrary list of calling conventions and type info into the binary

trufae:
Let's close this, the right fix is somewhere else

trufae:
aka in another castle

PR #17485, opened 2020-08-18
Handle COND_AL and COND_NV in analysis ##anal

lowlyw:

**Detailed description**

if a RAnalOp is properly filled out, in some archs, there can be ALWAYS or NEVER conditionals.
anal should take this into consideration and treat them as either a NOP (never) or as a non conditional (ALWAYS).

...

**Test plan**

i'm adding support for another arch that supports this and will pr this soon

...

**Closing issues**
n/a

### DIFF
--- a/libr/anal/op.c
+++ b/libr/anal/op.c
@@ -110,6 +110,17 @@ R_API int r_anal_op(RAnal *anal, RAnalOp *op, ut64 addr, const ut8 *data, int le
 		if (ret < 1) {
 			op->type = R_ANAL_OP_TYPE_ILL;
 		}
+		/* special case some conditionals */
+		if (op->cond == R_ANAL_COND_AL) {
+			/* Always conditional. */
+			op->cond = 0;
+			op->type &= ~R_ANAL_OP_TYPE_COND;
+			op->fail = 0;
+		} else if (op->cond == R_ANAL_COND_NV) {
+			/* Never conditional. */
+			op->cond = 0;
+			op->type = R_ANAL_OP_TYPE_NOP;
+		}
 		op->addr = addr;
 		/* consider at least 1 byte to be part of the opcode */
 		if (op->nopcode < 1) {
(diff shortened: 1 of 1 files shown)


lowlyw:
holy cow... how the hell did this break so many tests?

XVilka:
Most of these tests look autogenerated, so it makes sense, e.g. signatures/FLIRT ones.

kazarmy:
I think this suggests that `op->cond` doesn't always have a valid value. Also,
https://github.com/radareorg/radare2/blob/fb1190d248b09849fd29c9c5854bb32919d83929/libr/include/r_anal.h#L493
Maybe check `op->type` first  before `op->cond`?

lowlyw:
@kazarmy that shouldn't matter tbh, it's only clearing the cond bit if it is already set, i can change it to be more clear, but in theory it doesn't matter.

kazarmy:
It matters because
https://github.com/radareorg/radare2/blob/5f0be49fd16fadb8c6338d0807f3de65c5fb1e0e/libr/include/r_anal.h#L500
combines really well with, e.g.:
https://github.com/radareorg/radare2/blob/5f0be49fd16fadb8c6338d0807f3de65c5fb1e0e/libr/core/canal.c#L982

Unfortunately another reason is that even for `op->type & R_ANAL_OP_TYPE_COND`, `op->cond` may not be set or not set properly. For example

bb_je

there's a `je` at the end of the basic block at 0x3960 and it's clearly neither COND_AL nor COND_NV.

radare:
Any chance to check this @lowlyw ?

trufae:
ping

trufae:
@lowlyw  any time for this?

trufae:
can you rebase this PR on top of master? we are now using capstone5 and many tests has changed. will be good what are the current results

trufae:
@aemmitt-ns @crowell what are the plans for this PR? can we close it? i think this was also affecting the esilsolver, so if it can help to simplify some future logics we may think on handling those conditional instruction

trufae:
Closing, timeout

PR #17489, opened 2020-08-18
Fix #17353 - Remove duplicated implementation of uleb128 in mach0 ##bin

### LINKED ISSUE #17353: Remove duplicated implementation of uleb128 in the MACHO parser
There was a time in which r2 had 4 different implementations of uleb128, right now there's only 2.

```c
$ git grep ^static | grep leb128
libr/bin/format/mach0/mach0.c:static ut64 read_uleb128(ulebr *r, ut8 *end) {
libr/bin/format/mach0/mach0.c:static st64 read_sleb128(ulebr *r, ut8 *end) {
```

* Time to get rid of it!
* Probably good to add some tests

Related: https://github.com/radareorg/radare2/issues/15479

phakeobj:

**Detailed description**

I removed the duplicate implementations of read_{u,s}leb128 in mach0.c and wrote wrappers to use the implementations from RUtil. I also removed ulebr structure as I found it useless.

**Test plan**

...

**Closing issues**

Closes #17353

### DIFF
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -36,11 +36,6 @@ static ut64 read_uleb128(ut8 **p, ut8 *end) {
 	return v;
 }
 
-static st64 read_sleb128(ut8 **p, ut8 *end) {
-	st64 result = r_sleb128 ((const ut8 **)p, end);
-	return result;
-}
-
 static ut64 entry_to_vaddr(struct MACH0_(obj_t) *bin) {
 	switch (bin->main_cmd.cmd) {
 	case LC_MAIN:
@@ -1635,7 +1630,7 @@ static bool reconstruct_chained_fixup(struct MACH0_(obj_t) *bin) {
 			}
 			break;
 		case BIND_OPCODE_SET_ADDEND_SLEB:
-			read_sleb128 (&p, end);
+			r_sleb128 (&p, end);
 			break;
 		case BIND_OPCODE_SET_SEGMENT_AND_OFFSET_ULEB:
 			seg_idx = imm;
@@ -3264,7 +3259,7 @@ RSkipList *MACH0_(get_relocs)(struct MACH0_(obj_t) *bin) {
 				p++;
 				switch (op) {
 #define ULEB() read_uleb128 (&p, end)
-#define SLEB() read_sleb128 (&p, end)
+#define SLEB() r_sleb128 (&p, end)
 				case BIND_OPCODE_DONE: {
 					bool in_lazy_binds = pidx == 1;
 					if (!in_lazy_binds) {
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -34,53 +34,17 @@ typedef struct {
 // USE THIS: int ws = bf->o->info->big_endian;
 #define mach0_endian 1
 
-// TODO: Use the implementation from RUtil
 static ut64 read_uleb128(ulebr *r, ut8 *end) {
-	ut64 result = 0;
-	int bit = 0;
-	ut64 slice = 0;
-	ut8 *p = r->p;
-	do {
-		if (p == end) {
-			eprintf ("malformed uleb128\n");
-			return UT64_MAX;
-		}
-		slice = *p & 0x7f;
-		if (bit > 63) {
-			eprintf ("uleb128 too big for uint64, bit=%d, result=0x%"PFMT64x"\n", bit, result);
-			return UT64_MAX;
-		} else {
-			result |= (slice << bit);
-			bit += 7;
-		}
-	} while (*p++ & 0x80);
-	r->p = p;
-	return result;
+	int datalen = end - r->p;
+	ut64 v;
+	r->p = (ut8 *)r_uleb128 (r->p, datalen, &v);
+	return v;
 }
 
 static st64 read_sleb128(ulebr *r, ut8 *end) {
-	st64 result = 0;
-	int bit = 0;
-	ut8 byte = 0;
-	ut8 *p = r->p;
-	do {
-		if (p == end) {
-			eprintf ("malformed sleb128\n");
-			break;
-		}
-		byte = *p++;
-		result |= (((st64)(byte & 0x7f)) << bit);
-		bit += 7;
-		if (bit > 63) {
-			eprintf ("too large sleb128 shift\n");
-			return result;
-		}
-	} while (byte & 0x80);
-	// sign extend negative numbers
-	if ((byte & 0x40)) {
-		result |= (-1LL) << bit;
-	}
-	r->p = p;
+	const ut8 *t = r->p;
+	st64 result = r_sleb128 (&t, end);
+	r->p = (ut8 *)t;
 	return result;
 }
 
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -31,8 +31,9 @@ typedef struct {
 #define mach0_endian 1
 
 static ut64 read_uleb128(ut8 **p, ut8 *end) {
+	bool success = true;
 	ut64 v;
-	*p = (ut8 *)r_uleb128 (*p, end - *p, &v);
+	*p = (ut8 *)r_uleb128 (*p, end - *p, &v, &success);
 	return v;
 }
 
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -31,10 +31,12 @@ typedef struct {
 #define mach0_endian 1
 
 static ut64 read_uleb128(ut8 **p, ut8 *end) {
-	bool success = true;
+	const char *error = NULL;
 	ut64 v;
-	*p = (ut8 *)r_uleb128 (*p, end - *p, &v, &success);
-	if (!success) {
+	*p = (ut8 *)r_uleb128 (*p, end - *p, &v, &error);
+	if (error) {
+		eprintf (error);
+		R_FREE (error);
 		return UT64_MAX;
 	}
 	return v;
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -10,10 +10,6 @@
 #define bprintf if (bin->verbose) eprintf
 #define Eprintf if (mo->verbose) eprintf
 
-typedef struct _ulebr {
-	ut8 *p;
-} ulebr;
-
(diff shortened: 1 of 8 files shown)


libr/bin/format/mach0/mach0.c
```diff
@@ -34,53 +30,14 @@ typedef struct {
...
-		} else {
-			result |= (slice << bit);
-			bit += 7;
-		}
-	} while (*p++ & 0x80);
-	r->p = p;
-	return result;
+static ut64 read_uleb128(ut8 **p, ut8 *end) {
+	ut64 v;
+	*p = (ut8 *)r_uleb128 (*p, end - *p, &v);
```
ret2libc, line 35: I think we should check for NULL and when the return value is NULL `read_uleb128` should return `UT64_MAX` to match old behaviour.

libr/bin/format/mach0/mach0.c
```diff
@@ -34,53 +30,14 @@ typedef struct {
...
-			return result;
-		}
-	} while (byte & 0x80);
-	// sign extend negative numbers
-	if ((byte & 0x40)) {
-		result |= (-1LL) << bit;
-	}
-	r->p = p;
+static st64 read_sleb128(ut8 **p, ut8 *end) {
+	st64 result = r_sleb128 ((const ut8 **)p, end);
```
ret2libc, line 40: no need for this function, just use `r_sleb128` and remove function `read_sleb128`.

ret2libc (changes requested):

libr/bin/format/mach0/mach0.c
phakeobj: r_uleb128 from rutil does not return NULL in all scenarios that read_uleb128 did, so we won't cover all cases by just checking for NULL rv

phakeobj: maybe we can refactor walk_exports as they did in llvm (https://reviews.llvm.org/rL308690)

ret2libc: IMHO `r_uleb128` should be slightly modified to return NULL (aka error) when something goes wrong, like it was done in `read_uleb128`. Right now `r_uleb128` always tries to return something, even if there were errors, but in this way there is no way for the caller of `r_uleb128` to know whether there were problems or not. @trufae do you agree?

phakeobj: that's definitely possible, but I have to check that it doesn't break anything. I tried to return with v = UT64_MAX when r_uleb128 and there seems to be a regression in other places.

ret2libc: yeah probably some try-and-error will be needed :(

phakeobj:
found this binary, [macho-trie-bad-export-info-malformed-uleb128.zip](https://github.com/radareorg/radare2/files/5138821/macho-trie-bad-export-info-malformed-uleb128.zip), in llvm's test suite that causes master branch to yell 'malformed uleb128' twice, while on my branch it seems like everything is working smooth.

ret2libc:
Did I miss anything or I saw a message from you with regard to a mach0 binary that is incorrectly accepted without problems with this new PR? @phakeobj

phakeobj:
I had some problems with making sure this binary is actually doing what I hope it is, so I removed to comment to avoid confusion. I will restore it once I will figure things out.

phakeobj:
I'm not sure if it's a problem on my end or not, but are you able to break on read_uleb128 on master branch when running any of these binaries? 
[macho-malformed-uleb128.zip](https://github.com/radareorg/radare2/files/5152743/macho-malformed-uleb128.zip)

ret2libc:
> I'm not sure if it's a problem on my end or not, but are you able to break on read_uleb128 on master branch when running any of these binaries? [macho-malformed-uleb128.zip](https://github.com/radareorg/radare2/files/5152743/macho-malformed-uleb128.zip)

Yes. `gdb --args radare2 ./macho-rebase-uleb-malformed-uleb128`. Then I run `r` and `CTRL+C` immediately. Then `b read_uleb128` and `r` again. It should break.

phakeobj:
> > I'm not sure if it's a problem on my end or not, but are you able to break on read_uleb128 on master branch when running any of these binaries? [macho-malformed-uleb128.zip](https://github.com/radareorg/radare2/files/5152743/macho-malformed-uleb128.zip)
> 
> Yes. `gdb --args radare2 ./macho-rebase-uleb-malformed-uleb128`. Then I run `r` and `CTRL+C` immediately. Then `b read_uleb128` and `r` again. It should break.

So the problem is confirmed to be on my side. Thanks @ret2libc, I will update this thread once I have a solution that will allow me to debug the code like a sane human being.

libr/bin/format/mach0/mach0.c
radare: yes, but this is probably something to care in a separate PR imho, the behaviour doesnt seems to be different from current code.

phakeobj:
it differs if you try to run one of the binaries I talked with @ret2libc about.

libr/bin/format/mach0/mach0.c
phakeobj: it is different when running these binaries, [macho-malformed-uleb128.zip](https://github.com/radareorg/radare2/files/5152743/macho-malformed-uleb128.zip). master prints a malformed uleb error while this branch doesn't.

trufae:
Is this file in testbins? Can we have the ci red ?

phakeobj:
> 
> 
> Is this file in testbins? Can we have the ci red ?

In progress @ https://github.com/radareorg/radare2-testbins/pull/51

ret2libc:
Hey @phakeobj how is it going with this? Any luck debugging the issue?

phakeobj:
> 
> 
> Hey @phakeobj how is it going with this? Any luck debugging the issue?

yeah! I'll be pushing some code for review pretty soon :-)

trufae:
👏

libr/util/uleb128.c
```diff
@@ -25,13 +26,24 @@ R_API const ut8 *r_uleb128(const ut8 *data, int datalen, ut64 *v) {
...
 				} else {
 					sum |= ((ut64) (c & 0x7f) << s);
 				}
 				if (!(c & 0x80)) {
+					malformed_uleb = false;
 					break;
 				}
 			}
+			if (malformed_uleb) {
+				eprintf ("malformed uleb128\n");
```
ret2libc, line 42: This `eprintf` should not be here IMO. At most, it can be printed by the callers, but utilities functions should not print anything.

phakeobj: I agree. My initial intention was to make error a char ** argument that would store the error message in case of an error, but since there was already an eprintf in r_uleb128,
https://github.com/radareorg/radare2/blob/eab60612287c2d7680c9b4a80e1b7489bcc2dd5a/libr/util/uleb128.c#L27
I decided to comply with the current implementation and eprintf from within r_uleb128.

So, do you suggest I should refactor the code to eprintf from the caller, in both eprintf cases?

ret2libc: Yes, just move the eprintf in mach0.c. Don't care about all other callers for now ;)

libr/include/r_util/r_uleb128.h
```diff
@@ -5,7 +5,7 @@
 // compression used to store an arbitrarily large integer in a small number of
 // bytes. LEB128 is used in the DWARF debug file format.
 
-R_API const ut8 *r_uleb128(const ut8 *data, int datalen, ut64 *v);
+R_API const ut8 *r_uleb128(const ut8 *data, int datalen, ut64 *v, bool *success);
```
trufae, line 8: if its not successful can we return NULL and avoid changing the signature i think.

libr/util/uleb128.c
```diff
@@ -4,10 +4,11 @@
 
 /* dex/dwarf uleb128 implementation */
 
-R_API const ut8 *r_uleb128(const ut8 *data, int datalen, ut64 *v) {
+R_API const ut8 *r_uleb128 (const ut8 *data, int datalen, ut64 *v, bool *success) {
```
trufae, line 7: function definitions dont have a space before (. only when calling them

trufae (changes requested):

libr/include/r_util/r_uleb128.h
phakeobj: this will break the concept of caller calling eprintf with the relevant error message (thus, will break ci)

phakeobj: and I cannot guarantee that none of the other callers to r_uleb128 aren't counting on valid return value even if decoding fails.

trufae (approved):

libr/include/r_util/r_uleb128.h
```diff
@@ -5,9 +5,9 @@
 // compression used to store an arbitrarily large integer in a small number of
 // bytes. LEB128 is used in the DWARF debug file format.
 
-R_API const ut8 *r_uleb128(const ut8 *data, int datalen, ut64 *v);
+R_API const ut8 *r_uleb128(const ut8 *data, int datalen, ut64 *v, const char **error);
```
ret2libc, line 8: Is it really necessary to return a full heap-allocated string? What about:
```suggestion
typedef enum {
   R_ULEB_UNDEFINED_SHIFT_ERR,
   R_ULEB_MALFORMED_ERR,
} RUlebError;
R_API const ut8 *r_uleb128(const ut8 *data, int datalen, ut64 *v, RUlebError *error);
```

and let the caller just choose what to print exactly? You have to think these functions *may* be called by other callers as well and it's not said all of them actually want those strings you have provided.

phakeobj: > Is it really necessary to return a full heap-allocated string?

Most definitely not.

> and let the caller just choose what to print exactly? You have to think these functions _may_ be called by other callers as well and it's not said all of them actually want those strings you have provided.

The idea was that if someone passes something other than NULL in `error` he knows what he's doing and what should he do if `error` is not NULL when the utility function returns.

I like your idea and if you think that it will be better in any way than the one currently suggested, I will go for it.

ret2libc: @thestr4ng3r @trufae WDYT?

trufae: Yes, i think it's fine this way. At least as an intermediate step for another refactoring, because as @phakeobj said, there are many other calls that may not be checking for errors and this should be done in another PR.

ret2libc (approved):

ret2libc:
Thanks!

PR #17141, opened 2020-06-23
Fix #3929 - Enable LTO in sys/static.sh ##build

trufae:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

PR #17497, opened 2020-08-20
Don't use LTO for Debian package builds

kazarmy:

**Detailed description**

This pr attempts to fix the following problem regarding the Debian package build (i.e. the build is taking longer than 50 minutes) by disabling LTO:

debpkg-50min
https://travis-ci.com/github/radareorg/radare2/jobs/375579914#L3074

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ jobs:
       os: linux
       name: Debian
       dist: stretch
-      env: COMPILER_NAME=gcc DEBPKG=1 CC=gcc CFLAGS="-O2 -DR2_ASSERT_STDOUT=1 -fno-lto" LDFLAGS="-fno-lto"
+      env: COMPILER_NAME=gcc DEBPKG=1 CC=gcc CFLAGS="-O2 -DR2_ASSERT_STDOUT=1"
     # Run Fuzzing Sanity Test on every push
     # XXX its broken it complains about invalid key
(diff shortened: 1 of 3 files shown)


kazarmy:
Description updated

XVilka:
It was enabled by @trufae in https://github.com/radareorg/radare2/pull/17141 on purpose. Thus calling him to check this one.

kazarmy:
Ok, but 50 minutes isn't enough

trufae:
LTO should be used only for static builds, not for debian ones because those are not static. instead of passing nolto it willll be better to just not pass lto in that case. i though i only enabled it in a PR thta was not merged yet and it was only affecting sys/static.sh iirc

kazarmy:
> instead of passing nolto it willll be better to just not pass lto in that case

okay 272ecca60b1a1d441f03e97a10342038008895c8

> LTO should be used only for static builds, not for debian ones because those are not static

I'm not a consumer of the Debian package (I can just git pull, and this pr is just to make the build green), but as far as I can see, the build does call sys/static.sh

Debian_static
https://travis-ci.com/github/radareorg/radare2/jobs/375579914#L329

and

https://github.com/radareorg/radare2/blob/3c889001347fb764db98594b97e65b4e22cf9600/travis-script#L90-L95

PR #17501, opened 2020-08-20
WIP: Add user configurable JSON encoding options

### LINKED ISSUE #17368: izj not escaping json data properly.
### Work environment

| OS/arch/bits (mandatory)                             | macOS Catalina 10.15.6
| File format of the file you reverse (mandatory)      | Dex
| Architecture/bits of the file (mandatory)            | Dalvik
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.6.0-git 25007 @ darwin-x86-64 git.4.4.0-479-gb25c3275d commit: b25c3275dee4543e2464de60f58e8f279d9eed87 build: 2020-07-24__13:00:45

### Expected behavior
When running the following command `izj`, the expected outcome is properly escaped JSON data. 

### Actual behavior
The outputted JSON has not escaped characters appropriately, resulting in malformed JSON.

1. r2 -a dalvik [classes.dex.zip](https://github.com/radareorg/radare2/files/4988149/classes.dex.zip)
2. aa
3. izj > string_output.json
4. View string_ouput.json has incorrectly escaped characters.

```
 {
        "vaddr": 6230364,
        "paddr": 6230364,
        "ordinal": 3,
        "size": 16,
        "length": 16,
        "section": "data",
        "type": "ascii",
        "string":"\u0001\u0002\u0000\u0001\u0001\u0002\u0002\u0000\u0000\u0000\u0001\xe1"
    },
```
Identified this while trying to use r2pipe to pull string data from the binary in json format, but appears to be an issue within r2. 

R2pipe error due to the above: `r2pipe.cmdj.Error: Invalid \escape: line 1 column 343 (char 342)`

Tried enabling the following but the issue still persists.

```
e str.escbslash=true

e scr.strconv=asciidot

```
[string_output.json.zip](https://github.com/radareorg/radare2/files/4988190/string_output.json.zip)
[r2pipe_test.py.zip](https://github.com/radareorg/radare2/files/4988306/r2pipe_test.py.zip)

hexploitable:

**Detailed description**

Currently, the JSON output for various commands can include forbidden characters including \x. This is outside of the JSON spec so requires some adjustments. Rather than assume the best case for the user, we can afford them multiple choices.

**Encoding options:**
- [ ] base64
- [ ] unicode
- [ ] backslash-x hex values
- [ ] hexpairs
- [ ] strip non-printable
...

**Test plan**

...

**Closing issues**
closes #17368 
...

### DIFF
--- a/libr/anal/anal.c
+++ b/libr/anal/anal.c
@@ -200,7 +200,6 @@ R_API bool r_anal_use(RAnal *anal, const char *name) {
 	RAnalPlugin *h;
 
 	if (anal) {
-		bool change = anal->cur && strcmp (anal->cur->name, name);
 		r_list_foreach (anal->plugins, it, h) {
 			if (!h->name || strcmp (h->name, name)) {
 				continue;
@@ -213,9 +212,6 @@ R_API bool r_anal_use(RAnal *anal, const char *name) {
 #endif
 			anal->cur = h;
 			r_anal_set_reg_profile (anal);
-			if (change) {
-				r_anal_set_fcnsign (anal, NULL);
-			}
 			return true;
 		}
 	}
@@ -241,27 +237,6 @@ R_API bool r_anal_set_reg_profile(RAnal *anal) {
 		free (p);
 	}
 	return ret;
-}
-
-R_API bool r_anal_set_fcnsign(RAnal *anal, const char *name) {
-	const char *dirPrefix = r_sys_prefix (NULL);
-	const char *arch = (anal->cur && anal->cur->arch) ? anal->cur->arch : R_SYS_ARCH;
-	const char *file = (name && *name)
-		? sdb_fmt (R_JOIN_3_PATHS ("%s", R2_SDB_FCNSIGN, "%s.sdb"), dirPrefix, name)
-		: sdb_fmt (R_JOIN_3_PATHS ("%s", R2_SDB_FCNSIGN, "%s-%s-%d.sdb"), dirPrefix,
-			anal->os, arch, anal->bits);
-	if (r_file_exists (file)) {
-		sdb_close (anal->sdb_fcnsign);
-		sdb_free (anal->sdb_fcnsign);
-		anal->sdb_fcnsign = sdb_new (0, file, 0);
-		sdb_ns_set (anal->sdb, "fcnsign", anal->sdb_fcnsign);
-		return (anal->sdb_fcnsign != NULL);
-	}
-	return false;
-}
-
-R_API const char *r_anal_get_fcnsign(RAnal *anal, const char *sym) {
-	return sdb_const_get (anal->sdb_fcnsign, sym, 0);
 }
 
 R_API bool r_anal_set_triplet(RAnal *anal, const char *os, const char *arch, int bits) {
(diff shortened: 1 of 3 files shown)


kazarmy:
The problem with #17368 is that JSON assumes that all strings are Unicode, but not all byte sequences are valid Unicode strings. For example, if you decode "\u00e1" using UTF-8, you will get `[ 0xc3, 0xa1 ]` and not `[ 0xa1 ]`.

The workaround for #17368 is probably to use `r2pipe.cmd` instead of `r2pipe.cmdj` and do the string->byte sequence conversion manually. The proper fix is probably to make `izj`, `iVj` and their ilk return byte arrays instead of strings.

kazarmy:
> ... if you decode "\u00e1" using UTF-8, you will get [ 0xc3, 0xa1 ] and not [ 0xa1 ]

The more precise sentence is to say that if you convert "\u00e1" into a byte sequence assuming a UTF-8 encoding, you will get `[ 0xc3, 0xa1 ]` and not `[ 0xa1 ]`.

kazarmy:
Oh sloppy me, I meant to say:

... and not `[ 0xe1 ]`.

XVilka:
> The proper fix is probably to make `izj`, `iVj` and their ilk return byte arrays instead of strings.

I second this.

hexploitable:
Good point, ok ... I guess knowing that *j commands are probably used for scripting and tooling - byte arrays can work, but they're not exactly great for an interactive user inside r2.

Some questions:
1. Do you think the byte arrays should be returned in all instances, i.e. for every string - or only for those entries with unrepresentable bytes?
2. If we do it for every string, most tools people have built around this in r2pipe etc. will break. Not that it will be hard for them to fix, but still - it will be a breaking change.
3. I'm sure there are hundreds of commands that utilize j at the end, fixing this centrally will affect the others right - are you aware of any problems in doing that?

kazarmy:
@hexploitable Your plans for this?

hexploitable:
@kazarmy seems like you guys are still not sure what the fix should be, so as the collaborators I'd defer to whatever way you decide it should be fixed. Once we have consensus, I'll make the changes - does that sound ok?

kazarmy:
Don't worry just do whatever seems reasonable -- @trufae will need to have a look at the code anyway. It's good experience any which way.

ret2libc:
@hexploitable Hi! How is this going? Do you need assistance with anything? If so, feel free to ask :)

hexploitable:
Just afk currently, feel free to reassign it if priority. Otherwise i will make the PR changes on the weekend.

- thanks

ret2libc:
That's fine, thanks!

trufae:
Seems like theres no formal solution for this because json doesnt support binary data as is. only text which is bound  to the utf8 verifier and this is bad for portability. What we can do is imho to have a config var to tweak pj behaviour to be able to embed binary data inside json using any of the selected encoders:

* base64
* unicode
* backslash-x hex values
* hexpairs
* strip non-printable

I think this way we can delegate the responsability to pick the most optimal format for their usecases. we had base64 in the past and it was a huge bottleneck for izj

hexploitable:
I agree - this would be a really good way to solve this issue. As noted before, I think blind-encoding all json outputs could have adverse effects and providing multiple mechanisms for users to consume is a more solid path forward.

hexploitable:
For cleanliness, opening a new MR since this is all changing.

PR #17502, opened 2020-08-20
Implement `lj` ##json

officialcjunior:

**Detailed description**
Issue #13969 suggests to implement `lj` for consistency and standardization.

**Test plan**

Right now, there's extra spaces and a `\n` at the end, I would need guidance on how to fix that, if needed.
```
[{"             file1                 ../                  ./               file2\n"}] 
```

**Closing issues**

As discussed in #13969

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -1188,6 +1188,22 @@ static int cmd_ls(void *data, const char *input) { // "ls"
 			eprintf ("Usage: less [file]\n");
 		}
 		break;
+	case 'j': { // "lj"
+		PJ * pj = pj_new();
+		pj_a (pj);
+		pj_o (pj);
+		char * res = r_syscmd_ls (arg);
+		pj_s (pj, res);
+		pj_end (pj);
+		pj_end (pj);
+		char * j = pj_drain (pj);
+		if (j) {
+			r_cons_printf ("%s \n", j);
+			free (j);
+		}
+		free (res);
+		break;
+		}
 	default: // "ls"
 		if (!arg) {
 			arg = "";
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -1189,18 +1189,24 @@ static int cmd_ls(void *data, const char *input) { // "ls"
 		}
 		break;
 	case 'j': { // "lj"
-		PJ * pj = pj_new();
-		pj_a (pj);
-		pj_o (pj);
-		char * res = r_syscmd_ls (arg);
-		pj_s (pj, res);
-		pj_end (pj);
-		pj_end (pj);
-		char * j = pj_drain (pj);
-		if (j) {
-			r_cons_printf ("%s \n", j);
-			free (j);
+		if (!arg) {
+			arg = "";
 		}
+		if (r_fs_check (core->fs, arg)) {
+			r_core_cmdf (core, "md %s", arg);
+		} else {
+			PJ * pj = pj_new ();
+			pj_a (pj);
+			pj_o (pj);
+			RList *res = r_sys_dir (arg);
+			pj_s (pj, res);
+			pj_end (pj);
+			pj_end (pj);
+			char * j = pj_drain (pj);
+			if (j) {
+				r_cons_println (j);
+				free (j);
+			}
 		free (res);
 		break;
 		}


officialcjunior:
Hello, can I have some of your opinion on this?

ret2libc (changes requested):
I see multiple problems with this PR:
1. I believe that in json an object `{..}` needs keys and values. Right now you are just putting a value inside an object (`{"             file1                 ../                  ./               file2\n"}`) which, I think, is wrong.
2. What is the benefit of having `lj` if it just returns the whole string returned by `l`? The right approach here would be to return a json array containing separate elements for each file. Something like: `["file1", "file2", ....]`.

Does it make sense?

libr/core/cmd.c
```diff
@@ -1188,6 +1188,22 @@ static int cmd_ls(void *data, const char *input) { // "ls"
...
+		PJ * pj = pj_new();
+		pj_a (pj);
+		pj_o (pj);
+		char * res = r_syscmd_ls (arg);
+		pj_s (pj, res);
+		pj_end (pj);
+		pj_end (pj);
+		char * j = pj_drain (pj);
+		if (j) {
+			r_cons_printf ("%s \n", j);
```
trufae, line 1201: why this space between the json and the newline? you can also use r_cons_println and save the formattting

libr/core/cmd.c
```diff
@@ -1188,6 +1188,22 @@ static int cmd_ls(void *data, const char *input) { // "ls"
 			eprintf ("Usage: less [file]\n");
 		}
 		break;
+	case 'j': { // "lj"
+		PJ * pj = pj_new();
+		pj_a (pj);
+		pj_o (pj);
+		char * res = r_syscmd_ls (arg);
```
trufae, line 1195: can return null, on sandbox mode for example, so better stop early

libr/core/cmd.c
```diff
@@ -1188,6 +1188,22 @@ static int cmd_ls(void *data, const char *input) { // "ls"
 			eprintf ("Usage: less [file]\n");
 		}
 		break;
+	case 'j': { // "lj"
+		PJ * pj = pj_new();
```
trufae, line 1192: missing space before (

trufae (changes requested):

libr/core/cmd.c
officialcjunior: Stop early, in the sense?
I did not understand that part.

trufae: well. instead of early stop its probably better to show the actual error or just an empty listr of files.

libr/core/cmd.c
officialcjunior: Replaced it with `r_cons_println ("%s", j);`

libr/core/cmd.c
officialcjunior: > I see multiple problems with this PR:
> 
>     1. I believe that in json an object `{..}` needs keys and values. Right now you are just putting a value inside an object (`{"             file1                 ../                  ./               file2\n"}`) which, I think, is wrong.
> 
>     2. What is the benefit of having `lj` if it just returns the whole string returned by `l`? The right approach here would be to return a json array containing separate elements for each file. Something like: `["file1", "file2", ....]`.
> 
> 
> Does it make sense?

Yes, it does.
How should I do that? I tried some ways, but couldn't make it happen.
I'm not sure, but I think it should work if there's a comma in between each of the file names in the string? Currently, PJ thinks that this is a one big object.

officialcjunior:
> I see multiple problems with this PR:
> 
>     1. I believe that in json an object `{..}` needs keys and values. Right now you are just putting a value inside an object (`{"             file1                 ../                  ./               file2\n"}`) which, I think, is wrong.
> 
>     2. What is the benefit of having `lj` if it just returns the whole string returned by `l`? The right approach here would be to return a json array containing separate elements for each file. Something like: `["file1", "file2", ....]`.
> 
> 
> Does it make sense?

Yes, it does.
How should I do that? I tried some ways, but couldn't make it happen.
I'm not sure, but I think it should work if there's a comma in between each of the file names in the string? Currently, PJ treats this is a one big object.

ret2libc:
One way of doing it could be to get the output of `r_sys_cmd_ls ("-l " + arg)` and parsing it line by line. Maybe @trufae has some other inputs on how to do this correctly.

To create the correct json you should do something like:
```C
pj_a (j)
for each file in list_of_files:
    pj_o (j)
    pj_ks ("perms", file.permissions)
    pj_kn ("uid", file.uid)
    pj_kn ("gid", file.gid)
    pj_ks ("filename", file.name)
    pj_end (j)
pj_end(j);
```

libr/core/cmd.c
radare: Get the list of files instead of getting a string and construct a proper JSON with the information of each file (size, permissions, ...)

radare (changes requested):
z

officialcjunior:
Couple of things I am having trouble with:

1. 
> for each file in list_of_files:

How do I implement this.
As we are going to feed it to PJ only after parsing the `char *` array, how do I parse and implement that?

Right now, `list_of_files`, the thing  `r_syscmd_ls (arg)` returns, is a giant string array.

2.
```
    pj_ks ("perms", file.permissions)
    pj_kn ("uid", file.uid)
    pj_kn ("gid", file.gid)
    pj_ks ("filename", file.name)
```
I understand what you mean by this, but how do I make `file.permissions` and the similar ones, actually point to the required data?
Is it similar to [0d701a3](https://github.com/radareorg/radare2/commit/0d701a3b7972ced7eb450504ffbdffbe8273c303)?

ret2libc:
> Couple of things I am having trouble with:
> 
> 
> > for each file in list_of_files:
> 
> How do I implement this.
> As we are going to feed it to PJ only after parsing the `char *` array, how do I parse and implement that?
> 
> Right now, `list_of_files`, the thing `r_syscmd_ls (arg)` returns, is a giant string array.

As said in above comment, one way to do that could be:
> > One way of doing it could be to get the output of r_sys_cmd_ls ("-l " + arg) and parsing it line by line

libr/core/cmd.c
radare: Use r_sys_dir(arg) and iterate over the list

libr/core/cmd.c
radare: i dont see any change

ret2libc:
Ok, @trufae has provided a much better alternative, which is to use `r_sys_dir`!

officialcjunior:
Thank you.
How do I get the path, the variable to pass as an argument to `r_sys_dir`, into this function?

I apologize for asking too many questions. I'm such a noob and a learner at all of this.

trufae:
See how the ls function is implemented and copypaste the code from there

trufae:
You are already passing the path to ls, just use the same

officialcjunior:
Now, I was trying to do the iteration part done.

karliss:
@trufae  Is there a practical use case for `lj` command? The way I see it json output is for consumption by some kind of scripting language. Any such scripting language most likely already has a filesystem API, which provides more functionality, is better documented  and easier to use. 

Only situai


ls has some use when using r2 interactively as it allows you to check the directory content and locate a file before opening it without switching to proper shell or other program.

trufae:
I think lj should be an alias for `ls -j`. And this code must be implemented in the `ls` api instead

officialcjunior:
Alright, so I think it's better if I just close this PR now.
I'll try to do it the way mentioned above when I'm more used to everything.

trufae:
Up to you :) you can keep updating this one if you want. Feel free to close it if you prefer

Issue #9384, opened 2018-02-13
Radare2 do not properly detect branches with IT on Arm Thumb 2


mscherer:
### Work environment

| Questions                                            | Answers
|------------------------------------------------------|--------------------
| OS/arch/bits (mandatory)                             | RHEL x86_64 64
| File format of the file you reverse (mandatory)      | ELF
| Architecture/bits of the file (mandatory)            | Arm 6 in thumb2 mode
| r2 -v full output, **not truncated** (mandatory)         | radare2 2.3.0-git 16927 @ linux-x86-64 git.2.2.0-9-g469e811
| r2 -V full output in a pastebin document (mandatory) | https://gist.github.com/mscherer/c2f49a88cec5f7e4a025cab756a66ec3

### Expected behavior

IT for more than 1 instruction is correctly detected and show proper jump and conditional.

### Actual behavior

Basic block is not correctly detected, thus graphs are wrong.

### Steps to reproduce the behavior 

I was fiddling with a bigger binary, but I wrote a small assembly code for that:
```
.data

msg_a:
    .ascii      "R0 = 8\n"
len_a = . - msg_a

msg_b:
    .ascii      "R0 != 8\n"
len_b = . - msg_b

.text

.globl _start
.syntax unified

.code 16
some_IT_func:
    push {LR}
    cmp r0, r1
    itte eq
    addeq r0, r0, #1
    addeq r0, r0, #1
    movne r0, #4
    pop {pc}


.code 32
_start:
    mov r0, #6
    mov r1, r0

    blx some_IT_func

    ldreq r1, =msg_b
    ldreq r2, =len_b
 
    cmp r0, #8
    ldreq r1, =msg_a
    ldreq r2, =len_a
    mov r0, #1
    mov r7, #4   
    swi #0 

    mov     r0, #0  
    mov     r7, #1  
    swi     #0         
```

To compile (using the cross compiler of Fedora):
```
arm-linux-gnu-as -o test.o test.s
arm-linux-gnu-ld -s -o test.bin test.o 
```

Then use test.bin. Running on a RPI 2, this display:
```
R0 = 8
```

However, if we look at the disassembly of some_IT_func, the arrow do not show that instructions in 0x0001007e is skipped.

```
[0x00010074]> afb
0x00010074 0x0001007a 00:0000 6 j 0x0001007a f 0x0001007e
0x0001007a 0x0001007e 00:0000 4 j 0x0001007e
0x0001007e 0x00010082 00:0000 4
[...]

PR #17509, opened 2020-08-23
WIP: FIX ARM IT block (disass + anal)

0x8008135:

**Detailed description**
In ARM (Thumb) the IT instruction makes up to 4 following instructions conditional.
As the disassembly and analyze was performed one instruction at a time, the instructions part of an IT block were not disassembled correctly and the generated ESIL was also wrong.

This patch forces capstone to analyze previous instruction to know if the current instruction is part of an IT block and apply the correct conditions.

- function lookahead() was removed as unused
- ARM_INS_IT corrected in anal (line 2662 in libr/anal/p/anal_arm_cs.c)

Note:
Some work needs to be done on the analysis part to instruct ESIL to set the proper flags (IT) in the CPSR.

**Closing issues**
https://github.com/radareorg/radare2/issues/9384


XVilka:
Breaks ARM tests: https://github.com/radareorg/radare2/pull/17509/checks?check_run_id=1018914035#step:13:99

0x8008135:
> Breaks ARM tests: https://github.com/radareorg/radare2/pull/17509/checks?check_run_id=1018914035#step:13:99

My bad, should work now:
r2r test/db/anal/arm*  
Loaded 82 tests.
[82/82]                    77 OK         5 BR        0 XX        0 FX
Finished in 0 seconds.

r2r test/db/asm/arm_* 
Loaded 1698 tests.
[1698/1698]              1573 OK       120 BR        0 XX        5 FX
Finished in 2 seconds.
r2r test/db/asm/arm_*  7.62s user 10.07s system 779% cpu 2.268 total

libr/anal/p/anal_arm_cs.c
```diff
@@ -3288,6 +3271,29 @@ static int analop(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len, RAn
 				analop64_esil (a, op, addr, buf, len, &handle, insn);
 			}
 		} else {
+			if (a->bits == 16) {
+				//patch analysis if insn in IT block
```
XVilka, line 3275: It's a quite long (horizontally) construction, I think it could be extracted into a separate function.

XVilka (changes requested):

libr/anal/p/anal_arm_cs.c
0x8008135: Done

XVilka (approved):

0x8008135:
Unfortunately, I noticed that there is one case where this patch does work as intended...when seeking to an address within an IT block and performing disassembly (with pd)

Is there a way to increase the buffer to encompass the 4 previous instructions when disassembling from a specific address ?

Thanks in advance

0x8008135:
Following a discussion with @ret2libc, below is a detailed explanation of the issue.

If we perform the disassembly of the following hex pairs : 00281abf491c921cd21c121d
Results **without this PR**:
```
[0x00000000]> wx 00281abf491c921cd21c121d;pd 6
0x00000000      0028           cmp r0, 0
0x00000002      1abf           itte ne
0x00000004      491c           adds r1, r1, 1
0x00000006      921c           adds r2, r2, 2
0x00000008      d21c           adds r2, r2, 3
0x0000000a      121d           adds r2, r2, 4
```

Results **with this PR**:
```
[0x00000000]> wx 00281abf491c921cd21c121d;pd 6
0x00000000      0028           cmp r0, 0
0x00000002      1abf           itte ne
0x00000004      491c           addne r1, r1, 1
0x00000006      921c           addne r2, r2, 2
0x00000008      d21c           addeq r2, r2, 3
0x0000000a      121d           adds r2, r2, 4

[0x00000000]> wx 00281abf491c921cd21c121d;pdt 6
0x00000000      0028           cmp r0, #0               0,r0,==,$z,zf,:=,31,$s,nf,:=,32,$b,!,cf,:=,31,$o,vf,:=                0x00000000 
0x00000002      1abf           itte ne                  zf,!,?{,2,$$,+,pc,=,}                                                            
0x00000004      491c           addne r1, r1, #1         zf,!,?{,1,r1,+,0xffffffff,&,r1,=,}                                               
0x00000006      921c           addne r2, r2, #2         zf,!,?{,2,r2,+,0xffffffff,&,r2,=,}                                               
0x00000008      d21c           addeq r2, r2, #3         zf,?{,3,r2,+,0xffffffff,&,r2,=,}                                                 
0x0000000a      121d           adds r2, r2, #4          4,r2,+,0xffffffff,&,r2,=,$z,zf,:=,31,$s,nf,:=,31,$c,cf,:=,31,$o,vf,:=            
```
However, if we seek to an address after the IT instruction, the conditions on the adds are totally ignored:
```
[0x00000000]> wx 00281abf491c921cd21c121d;pd 6 @0x04
0x00000004      491c           adds r1, r1, 1 <-- should be addne
0x00000006      921c           adds r2, r2, 2 <-- should be addne
0x00000008      d21c           adds r2, r2, 3 <-- should be addeq
0x0000000a      121d           adds r2, r2, 4
```
I have been able to identify that this problem exists because the buffer on which the disassm/analysis is performed contains only the instructions from the current seek (or specified address). My investigation demonstrated that capstone and ESIL are working properly as long as the buffer include the 4 previous instructions.

Below is an example using capstone in python:

```
from capstone import * 
CODE = b"\x00\x28\x1a\xbf\x49\x1c\x92\x1c\xd2\x1c\x12\x1d" 
md = Cs(CS_ARCH_ARM, CS_MODE_THUMB) 
for i in md.disasm(CODE, 0x1000): 
    print("0x%x:\t%s\t%s" %(i.address, i.mnemonic, i.op_str)) 
```
Output:   
```
0x1000:	cmp	r0, #0
0x1002:	itte	ne
0x1004:	addne	r1, r1, #1
0x1006:	addne	r2, r2, #2
0x1008:	addeq	r2, r2, #3
0x100a:	adds	r2, r2, #4
```

For ARM (16 bits) this buffer should always include the 4 previous instructions to know if it is in an ITBLOCK or not.

If there is a good solution to load additional bytes when performing a disassembly/analysis or to adapt the buffer for ARM 16bits, then the current patch might not be required as the output will be directly correct. (AFAIK)

However, if the issue is too difficult to patch I will create an issue and propose this patch as-is.

In any cases, help appreciated :)

libr/anal/p/anal_arm_cs.c
```diff
@@ -2584,15 +2584,30 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
-	cs_insn *insn = NULL;
-	int n = cs_disasm (handle, buf, len, addr, distance, &insn);
-	if (n < 1) {
-		return UT64_MAX;
+static void anal_ITblock(csh handle, RAnalOp *op, cs_insn *insn, const ut64 addr, const ut8 *buf, int len) {	
+	//patch analysis if insn in IT block
+	cs_insn *ITinsn = NULL;
+	int ITcounter = 0;
+	int s = cs_disasm (handle, buf-8, len+8, addr-8, 5, &ITinsn);
+	for (int i = 0; i < s; i++) {
```
trufae, line 2592: * dont define vars inside the for() loop.
* use size_t for loop variables.

libr/anal/p/anal_arm_cs.c
```diff
@@ -2584,15 +2584,30 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
-static ut64 lookahead(csh handle, const ut64 addr, const ut8 *buf, int len, int distance) {
-	cs_insn *insn = NULL;
-	int n = cs_disasm (handle, buf, len, addr, distance, &insn);
-	if (n < 1) {
-		return UT64_MAX;
+static void anal_ITblock(csh handle, RAnalOp *op, cs_insn *insn, const ut64 addr, const ut8 *buf, int len) {	
+	//patch analysis if insn in IT block
+	cs_insn *ITinsn = NULL;
+	int ITcounter = 0;
+	int s = cs_disasm (handle, buf-8, len+8, addr-8, 5, &ITinsn);
```
trufae, line 2591: add spaces around + and -

libr/anal/p/anal_arm_cs.c
```diff
@@ -2584,15 +2584,30 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
 
-static ut64 lookahead(csh handle, const ut64 addr, const ut8 *buf, int len, int distance) {
-	cs_insn *insn = NULL;
-	int n = cs_disasm (handle, buf, len, addr, distance, &insn);
-	if (n < 1) {
-		return UT64_MAX;
+static void anal_ITblock(csh handle, RAnalOp *op, cs_insn *insn, const ut64 addr, const ut8 *buf, int len) {	
+	//patch analysis if insn in IT block
+	cs_insn *ITinsn = NULL;
+	int ITcounter = 0;
```
trufae, line 2590: dont use uppercase or camelcase variables. just lower stuff like `insn`

libr/anal/p/anal_arm_cs.c
```diff
@@ -2584,15 +2584,30 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
+	int s = cs_disasm (handle, buf-8, len+8, addr-8, 5, &ITinsn);
+	for (int i = 0; i < s; i++) {
+		if(ITinsn[i].id == ARM_INS_IT) {
+			ITcounter = r_str_nlen (ITinsn[i].mnemonic, 5);
+		}
+		if(ITcounter > 0) {
```
trufae, line 2596: missing space before (

libr/anal/p/anal_arm_cs.c
```diff
@@ -3288,6 +3297,10 @@ static int analop(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len, RAn
 				analop64_esil (a, op, addr, buf, len, &handle, insn);
 			}
 		} else {
+			if (a->bits == 16) {
+				anal_ITblock(handle, op, insn, addr, buf, len);
```
trufae, line 3301: missing space

libr/asm/p/asm_arm_cs.c
```diff
@@ -98,6 +98,25 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
 	}
 
 	n = cs_disasm (cd, buf, R_MIN (4, len), a->pc, 1, &insn);
+
+	if (a->bits == 16) {
+		//patch disassembly if insn in IT block
+		cs_insn *ITinsn = NULL;
+		int ITcounter = 0;
+		int s = cs_disasm (cd, buf-8, len+8, a->pc-8, 5, &ITinsn);
+		for (int i = 0; i < s; i++) {
```
trufae, line 107: same as above. also this code seems copypasta fromt he one above. can you move this to a separate function?

libr/asm/p/asm_arm_cs.c
```diff
@@ -98,6 +98,25 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
 	}
 
 	n = cs_disasm (cd, buf, R_MIN (4, len), a->pc, 1, &insn);
+
+	if (a->bits == 16) {
+		//patch disassembly if insn in IT block
+		cs_insn *ITinsn = NULL;
+		int ITcounter = 0;
+		int s = cs_disasm (cd, buf-8, len+8, a->pc-8, 5, &ITinsn);
```
trufae, line 106: you are calling cs_disasm twice in here. this is a worrying performance problem

libr/asm/p/asm_arm_cs.c
```diff
@@ -98,6 +98,25 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
...
+		for (int i = 0; i < s; i++) {
+			if(ITinsn[i].id == ARM_INS_IT) {
+				ITcounter = r_str_nlen (ITinsn[i].mnemonic, 5);
+			}
+			if(ITcounter > 0) {
+				if(insn->address == ITinsn[i].address) {
+					memcpy(insn->mnemonic, ITinsn[i].mnemonic, sizeof(insn->mnemonic));
```
trufae, line 113: more missing spaces. also , use strcpy (or r_str_ncpy) to avoid copyinng more than one null byte

libr/asm/p/asm_arm_cs.c
```diff
@@ -98,6 +98,25 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
...
+				}
```
trufae, line 114: add a break to stop iterating if the address is found?

libr/asm/p/asm_arm_cs.c
```diff
@@ -98,6 +98,25 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
 	}
 
 	n = cs_disasm (cd, buf, R_MIN (4, len), a->pc, 1, &insn);
+
+	if (a->bits == 16) {
+		//patch disassembly if insn in IT block
+		cs_insn *ITinsn = NULL;
+		int ITcounter = 0;
```
trufae, line 105: dont use capitalized vars

trufae (changes requested):

XVilka (approved):

ret2libc (commented):
I'm not sure whether I like this solution or not. It seems very inefficient when dealing with thumb mode, since it *always* require to read back some bytes. Also, how that would work when, for example, using `pad`? Is that going to work well?

I just got another idea, by the way, but let's see what you and others think: when disassembling/analyzing a IT instruction, you create new RAnalHints for the following instructions. Then when displaying those following instructions you apply the hint to show them in the right way. This is used for example to show immediates in a given base or to mark something as thumb vs non-thumb mode. It has the advantage of not requiring to "go back" (which may be imo risky even on arm, if we are crossing a thumb/no-thumb boundary) and it doesn't require to *always* read some bytes, but only when a IT instruction is encountered. On the other hand, it may require some thinking before implementing it, because RAnalHint are not immediately available to RAsm... What do you think?

That said, I don't work on arm so I leave the decision to others.

ret2libc:
Or, instead of RAnalHints, just keep a cache, as @trufae suggested, inside the ARM RAsmPlugin to store these kind of "hints".

0x8008135:
@ret2libc @trufae can you review and pinpoint where to invalidate the HT once a new analysis start.

libr/anal/p/anal_arm_cs.c
```diff
@@ -2584,21 +2587,36 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
-		return UT64_MAX;
+static void anal_itblock(RAnal *a, cs_insn *insn) {
+	if (!ht_it) {
+		ht_it = ht_uu_new0 ();
+	}
+	size_t i;
+	for (i = 1; i < 5; i++) {
+		switch (insn->mnemonic[i]) {
+		case 0x74: //'t'
+			if (!ht_uu_insert (ht_it, insn->address + (i * insn->size), insn->detail->arm.cc)) {
```
ret2libc, line 2598: just use `ht_uu_update` if you want to insert or update.
From the ht_inc.h file:
```
// Insert a new Key-Value pair into the hashtable, or updates the value if the key already exists.
SDB_API bool Ht_(update)(HtName_(Ht)* ht, const KEY_TYPE key, VALUE_TYPE value);
```

libr/anal/p/anal_arm_cs.c
```diff
@@ -2584,21 +2587,36 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
+				ht_uu_update (ht_it, insn->address + (i * insn->size), insn->detail->arm.cc);
+			}
+			break;
+		case 0x65: //'e'
+			if (!ht_uu_insert (ht_it, insn->address + (i * insn->size), (insn->detail->arm.cc % 2)? insn->detail->arm.cc + 1: insn->detail->arm.cc - 1)) {
```
ret2libc, line 2603: same here

libr/asm/p/asm_arm_cs.c
```diff
@@ -40,13 +42,72 @@ static bool check_features(RAsm *a, cs_insn *insn) {
...
+
+static void disass_itblock(RAsm *a, cs_insn *insn) {
+	if (!ht_it) {
+		ht_it = ht_uu_new0 ();
+	}
+	size_t i;
+	for (i = 1; i < 5; i++) {
+		switch (insn->mnemonic[i]) {
+		case 0x74: //'t'
+			if (!ht_uu_insert (ht_it, a->pc + (i * insn->size), insn->detail->arm.cc)) {
```
ret2libc, line 88: also here

libr/asm/p/asm_arm_cs.c
```diff
@@ -40,13 +42,72 @@ static bool check_features(RAsm *a, cs_insn *insn) {
...
+				ht_uu_update (ht_it, a->pc + (i * insn->size), insn->detail->arm.cc);
+			}
+			break;
+		case 0x65: //'e'
+			if (!ht_uu_insert (ht_it, a->pc + (i * insn->size), (insn->detail->arm.cc % 2)?insn->detail->arm.cc + 1:insn->detail->arm.cc - 1)) {
```
ret2libc, line 93: and here

libr/asm/p/asm_arm_cs.c
```diff
@@ -113,7 +175,21 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
...
+		itcond = ht_uu_find (ht_it,  a->pc, &found);
+		if (found){
+			insn->detail->arm.cc = itcond;
+			insn->detail->arm.update_flags = 0;
+			r_str_cpy(insn->mnemonic, 
+				r_str_newf ("%s%s",
+					cs_insn_name (cd, insn->id),
+					cc_name (itcond)));
+		}
+		buf_asm = sdb_fmt ("%s%s%s",
```
ret2libc, line 192: just define it here as `char *buf_asm = ...` as it was before.

libr/asm/p/asm_arm_cs.c
```diff
@@ -113,7 +175,21 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
...
+			disass_itblock (a, insn);
+		}
+		char *buf_asm;
+		ut64 itcond;
+		itcond = ht_uu_find (ht_it,  a->pc, &found);
+		if (found){
+			insn->detail->arm.cc = itcond;
+			insn->detail->arm.update_flags = 0;
+			r_str_cpy(insn->mnemonic, 
+				r_str_newf ("%s%s",
```
ret2libc, line 188: if you allocate this here on the fly you can't free it, so we will leak a lot of memory.

libr/asm/p/asm_arm_cs.c
```diff
@@ -40,13 +42,72 @@ static bool check_features(RAsm *a, cs_insn *insn) {
...
+		return "";
+	}
+}
+
+static void disass_itblock(RAsm *a, cs_insn *insn) {
+	if (!ht_it) {
+		ht_it = ht_uu_new0 ();
+	}
+	size_t i;
+	for (i = 1; i < 5; i++) {
```
ret2libc, line 85: you should stop the for-loop before 5 if `insn->mnemonic` is shorter than that, otherwise you'll read out-of-bound memory.

ret2libc (changes requested):
One way to initialize/free the ht could be to use `RAsmPlugin.init/fini` methods. However, that's not exactly the same as "when an analysis starts/ends". And it won't work well when the user changes the bytes. However I'm not aware of a better solution. @trufae ?

libr/anal/p/anal_arm_cs.c
0x8008135: Done

libr/anal/p/anal_arm_cs.c
```diff
@@ -3775,6 +3795,19 @@ static RList *anal_preludes(RAnal *anal) {
...
 
+static int init (void* user) {
+	if (!ht_it) {
+		ht_it = ht_uu_new0 ();
+	}
+	return 0;
+}
+
+static int fini (void* user) {
+	free(ht_it);
```
ret2libc, line 3806:
```suggestion
	ht_uu_free (ht_it);
```

0x8008135: My bad, corrected

libr/asm/p/asm_arm_cs.c
0x8008135: see latest commit

0x8008135:
Last push added a routine to track and cleanup ITblock if an it instruction is overwritten.

Example:
> [0x00000000]> wx 00281abf491c921cd21c121d;
[0x00000000]> pd 6
            0x00000000      0028           cmp r0, 0                   ; compare
            0x00000002      1abf           itte ne
            0x00000004      491c           addne r1, r1, 1             ; add two values if Z clear (not equal)
            0x00000006      921c           addne r2, r2, 2             ; add two values if Z clear (not equal)
            0x00000008      d21c           addeq r2, r2, 3             ; add two values if Z set (equal)
            0x0000000a      121d           adds r2, r2, 4              ; add two values and update flags


Overwrite instruction at 0x4 with adds r2, r2, 4
> [0x00000000]> wx 121d @ 0x4; pd 6
            0x00000000      0028           cmp r0, 0                   ; compare
            0x00000002      1abf           itte ne
            0x00000004      121d           addne r2, r2, 4             ; add two values if Z clear (not equal)
            0x00000006      921c           addne r2, r2, 2             ; add two values if Z clear (not equal)
            0x00000008      d21c           addeq r2, r2, 3             ; add two values if Z set (equal)
            0x0000000a      121d           adds r2, r2, 4              ; add two values and update flags


Overwrite instruction at 0x2 with adds r2, r2, 4
> [0x00000000]> wx 121d @ 0x2; pd 6
            0x00000000      0028           cmp r0, 0                   ; compare
            0x00000002      121d           adds r2, r2, 4              ; add two values and update flags
            0x00000004      121d           adds r2, r2, 4              ; add two values and update flags
            0x00000006      921c           adds r2, r2, 2              ; add two values and update flags
            0x00000008      d21c           adds r2, r2, 3              ; add two values and update flags
            0x0000000a      121d           adds r2, r2, 4              ; add two values and update flags

Same applies to ESIL.

Should be good to merge after some more testing/comment from yours (@ret2libc / @trufae)

libr/anal/p/anal_arm_cs.c
```diff
@@ -2584,21 +2588,43 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
 
-static ut64 lookahead(csh handle, const ut64 addr, const ut8 *buf, int len, int distance) {
-	cs_insn *insn = NULL;
-	int n = cs_disasm (handle, buf, len, addr, distance, &insn);
-	if (n < 1) {
-		return UT64_MAX;
+static void anal_itblock(cs_insn *insn) {
+	size_t i;
+	ht_uu_update (ht_itblock, insn->address,  r_str_nlen (insn->mnemonic, 5));
+	for (i = 1; i < r_str_nlen (insn->mnemonic, 5); i++) {
```
radare, line 2594: compute this str_nlen outside the loop. one line above

libr/anal/p/anal_arm_cs.c
```diff
@@ -2584,21 +2588,43 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
+		switch (insn->mnemonic[i]) {
+		case 0x74: //'t'
+			ht_uu_update (ht_it, insn->address + (i * insn->size), insn->detail->arm.cc);
+			break;
+		case 0x65: //'e'
+			ht_uu_update (ht_it, insn->address + (i * insn->size), (insn->detail->arm.cc % 2)? insn->detail->arm.cc + 1: insn->detail->arm.cc - 1);
```
radare, line 2600: newline after '?'

libr/anal/p/anal_arm_cs.c
```diff
@@ -2584,21 +2588,43 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
+			break;
+		}
+	}
+}
+
+static void check_itblock(cs_insn *insn) {
+	size_t x;
+	ut64 itlen;
+	bool found;
+	itlen = ht_uu_find (ht_itblock, insn->address, &found);
```
radare, line 2612: define and assign here

libr/anal/p/anal_arm_cs.c
```diff
@@ -2584,21 +2588,43 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
-	cs_free (insn, n);
-	return result;
 }
 
 static void anop32(RAnal *a, csh handle, RAnalOp *op, cs_insn *insn, bool thumb, const ut8 *buf, int len) {
 	const ut64 addr = op->addr;
-	const int pcdelta = thumb? 4 : 8;
+	const int pcdelta = thumb? 4: 8;
 	int i;
+	bool found = 0;
```
radare, line 2625: = false

libr/anal/p/anal_arm_cs.c
```diff
@@ -3775,6 +3812,24 @@ static RList *anal_preludes(RAnal *anal) {
...
+	if (!ht_it) {
+		ht_it = ht_uu_new0 ();
+	}
+	if (!ht_itblock) {
+		ht_itblock = ht_uu_new0 ();
+	}
+	return 0;
+}
+
+static int fini (void* user) {
```
radare, line 3825: remove space before (

libr/anal/p/anal_arm_cs.c
```diff
@@ -3775,6 +3812,24 @@ static RList *anal_preludes(RAnal *anal) {
 	return l;
 }
 
+static int init (void* user) {
```
radare, line 3815: remove space before (

libr/anal/p/anal_arm_cs.c
```diff
@@ -2625,6 +2651,11 @@ static void anop32(RAnal *a, csh handle, RAnalOp *op, cs_insn *insn, bool thumb,
 	} else {
 		op->family = R_ANAL_OP_FAMILY_CPU;
 	}
+
+	if (insn->id != ARM_INS_IT) {
+		check_itblock(insn);
```
radare, line 2656: space before (

libr/anal/p/anal_arm_cs.c
```diff
@@ -60,6 +62,8 @@
 
 static RRegItem base_regs[4];
 static RRegItem regdelta_regs[4];
+static HtUU *ht_itblock;
```
radare, line 65: = NULL

libr/asm/p/asm_arm_cs.c
```diff
@@ -183,6 +270,24 @@ static int assemble(RAsm *a, RAsmOp *op, const char *buf) {
 	return opsize;
 }
 
+static bool init (void* user) {
```
radare, line 273: im not a big fan of using globals. Also, remove space before ( ^.

Maybe it will be good to make use of *user.. or well, pass RAsm*  argument to init/fini so we can get rid of all those globals

libr/asm/p/asm_arm_cs.c
```diff
@@ -113,6 +186,20 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
...
+		if (insn->id == ARM_INS_IT) {
+			disass_itblock (a, insn);
+		} else {
+			check_itblock (a, insn);
+		}
+		itcond = ht_uu_find (ht_it,  a->pc, &found);
+		if (found) {
+			insn->detail->arm.cc = itcond;
+			insn->detail->arm.update_flags = 0;
+			tmpstr = r_str_newf ("%s%s",
```
radare, line 198: char *tmpstr = r_str_newf...

libr/asm/p/asm_arm_cs.c
```diff
@@ -113,6 +186,20 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
...
+				cs_insn_name (cd, insn->id),
+				cc_name (itcond));
+			r_str_cpy (insn->mnemonic, tmpstr);
```
radare, line 201: free (tmpstr)

libr/asm/p/asm_arm_cs.c
```diff
@@ -40,13 +43,82 @@ static bool check_features(RAsm *a, cs_insn *insn) {
...
+}
+
 static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len) {
 	static int omode = -1;
 	static int obits = 32;
 	bool disp_hash = a->immdisp;
 	cs_insn* insn = NULL;
 	cs_mode mode = 0;
 	int ret, n = 0;
+	bool found = 0;
```
radare, line 118: = false

radare (changes requested):
The problem of this approach is that it have no way to invalidate the cache. Which may result in very scary results when opening multiples and swap between them. maybe we should reinitialize the plugins to reset this cache.

thanks!

libr/asm/p/asm_arm_cs.c
0x8008135: I agree and attempted to correct this as suggested, however doing so impacts other plugins (e.g. snes,tms320_fini,...) and requires some change in the logic of core component (e.g. plugin_free() in asm.c).

The [previous commit](https://github.com/radareorg/radare2/pull/17509/commits/693d709ee2eb91c6388995bd061ff983b83bcefd) cleans up the entries in the HTs if something is changed, resulting in low footprint.

I have seen other plugins having globals, so maybe we can go forward with this proposal (for now) and open issue to implement this change globally. What do you think ?

ret2libc:
@trufae what would you like to see to move this forward?
@anisse could you please update the PR to avoid conflicts with master?

anisse:
@ret2libc I think you meant to ask @0x8008135 to rebase the PR on master.

0x8008135:
> could you please update the PR to avoid conflicts with master?
Done

libr/anal/p/anal_arm_cs.c
```diff
@@ -2,10 +2,14 @@
 
 #include <r_anal.h>
 #include <r_lib.h>
+#include <ht_uu.h>
+#include <capstone/capstone.h>
```
trufae, line 6: shouldnt this be just capstone.h after rebase?

libr/anal/p/anal_arm_cs.c
```diff
@@ -2962,21 +2968,44 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
 }
 
-static ut64 lookahead(csh handle, const ut64 addr, const ut8 *buf, int len, int distance) {
-	cs_insn *insn = NULL;
-	int n = cs_disasm (handle, buf, len, addr, distance, &insn);
-	if (n < 1) {
-		return UT64_MAX;
+static void anal_itblock(cs_insn *insn) {
+	size_t i, size;
+	size =  r_str_nlen (insn->mnemonic, 5);
```
trufae, line 2973: define and assign in the same line

trufae (changes requested):
my main concern here is about that cache that have no way to be invalidated or reset or limited, so size in memory usage can grow quite a lot, and it will result on false results when switching between different files.

libr/anal/p/anal_arm_cs.c
ret2libc: yes
```suggestion
#include <capstone.h>
```

ret2libc:
> my main concern here is about that cache that have no way to be invalidated or reset or limited, so size in memory usage can grow quite a lot, and it will result on false results when switching between different files.

Yeah, this is known limitation of this solution. Do you have suggestions/alternatives? I honestly could not come up with something better so far, but this seemed a bit better than the previous implementation that always analyzed 5 instructions before each instruction.

Maybe one way could really be to use, somehow, RAnalHints, but that would require quite a bit of refactoring and binds here and there to let RAsmPlugin access RAnalHints... And I'm not sure it would really be that much better.

BTW, @trufae are RAnalHint per-file? Do they work well if you switch file/maps? If not, I would keep this implementation.

ret2libc:
In addition to what said before, I think that not even regular analysis works well with multiple files. So, all in all, this solution does not seem that bad. It is not perfect, but I think proper multi-file is still not there, not for this feature but not even for many others.

libr/anal/p/anal_arm_cs.c
```diff
@@ -2,10 +2,13 @@
 
 #include <r_anal.h>
 #include <r_lib.h>
+#include <ht_uu.h>
+#include <capstone/arm.h>
```
ret2libc, line 6:
```suggestion
#include <arm.h>
```

ret2libc:
@trufae ping. See previous messages that are replies to what you said above, please.

trufae:
Just address the two remaining comments and lgtm for merging

libr/anal/p/anal_arm_cs.c
```diff
@@ -2962,21 +2967,44 @@ static void anop64(csh handle, RAnalOp *op, cs_insn *insn) {
...
 }
 
-static ut64 lookahead(csh handle, const ut64 addr, const ut8 *buf, int len, int distance) {
-	cs_insn *insn = NULL;
-	int n = cs_disasm (handle, buf, len, addr, distance, &insn);
-	if (n < 1) {
-		return UT64_MAX;
+static void anal_itblock(cs_insn *insn) {
+	size_t i, size;
+	size =  r_str_nlen (insn->mnemonic, 5);
```
ret2libc, line 2972:
```suggestion
	size_t i, size =  r_str_nlen (insn->mnemonic, 5);
```

trufae (approved):

PR #17513, opened 2020-08-24
Fix #17492 - EOF when alt+e is pressed on Linux systems ##cons

### LINKED ISSUE #17492: Alt+key triggers quit
Pressing any <kbd>Alt+key</kbd> makes r2 quit, except the ones that needed for line editing (like <kbd>Alt+f/b</kbd>; maybe these are the only two exceptions). Issue _does not_ seem to be terminal dependent.

### Work environment

| OS/arch/bits (mandatory)                             | Arch Linux
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.4.0 0 @ linux-x86-64 git.4.4.0 commit: unknown build: 2020-05-08__21:16:55
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.5.0 24955 @ linux-x86-64 git.4.4.0-426-g9d7eda5ec  commit: 9d7eda5ec7367d1682e489e92d1be8e37e459296 build: 2020-07-17__11:05:37 (also tried with r2 from the prebuilt deb package)
| Terminal Emulators | Alacritty, Kitty, rxvt-unicode, Termite, Linux console
| Config | None

trufae:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/cons/dietline.c
+++ b/libr/cons/dietline.c
@@ -1593,7 +1593,7 @@ R_API const char *r_line_readline_cb(RLineReadCallback cb, void *user) {
 					buf[1] = r_cons_readchar_timeout (50);
 					if (buf[1] == -1) {
 						r_cons_break_pop ();
-						return NULL;
+						return I.buffer.data;
 					}
 				}
 				if (buf[0] == 0x5b) {	// [
--- a/libr/cons/dietline.c
+++ b/libr/cons/dietline.c
@@ -1591,9 +1591,10 @@ R_API const char *r_line_readline_cb(RLineReadCallback cb, void *user) {
 			default:
 				if (I.vtmode == 2) {
 					buf[1] = r_cons_readchar_timeout (50);
-					if (buf[1] == -1) {
+					if (buf[1] == -1) { // alt+e
 						r_cons_break_pop ();
-						return I.buffer.data;
+						__print_prompt ();
+						continue;
 					}
 				}
 				if (buf[0] == 0x5b) {	// [


libr/cons/dietline.c
```diff
@@ -1593,7 +1593,7 @@ R_API const char *r_line_readline_cb(RLineReadCallback cb, void *user) {
 					buf[1] = r_cons_readchar_timeout (50);
 					if (buf[1] == -1) {
 						r_cons_break_pop ();
-						return NULL;
+						return I.buffer.data;
```
ret2libc, line 1596: I'm not sure this is the right fix. Why shall we return something here, but not in e.g. https://github.com/radareorg/radare2/pull/17513/files#diff-1b276f78ec258bb5c8baaa8447300b2dL1607 or https://github.com/radareorg/radare2/pull/17513/files#diff-1b276f78ec258bb5c8baaa8447300b2dL1668 .

From what I can see the issue seems to be in `core.c:r_core_prompt_loop` and/or `core.c:r_core_prompt`. Indeed in `r_core_prompt` the function returns `false`(aka 0) when `ret == -1` and it returns `R_CORE_CMD_EXIT` when `ret == -2`. Now, `r_core_prompt_loop` should exit from the prompt loop *only* when `err == R_CORE_CMD_EXIT`, but the current check is ` if (err < 1 ) // handle ^D .... ; break;` so also the case where `ret` is `false` is considered the same as `R_CORE_CMD_EXIT`.

ret2libc (changes requested):

libr/cons/dietline.c
trufae: this returns the current buffer instead of NULL which was causing the program to be closed

ret2libc: I'm not sure I understand your reply. I do know you are trying to fix the issue of radare2 being closed, but as I explained in the comment above, I do not think this is the right solution (again, for the reasons explained above).

trufae: Yes, that breaks the line and we probably just want to get rid of the trashy chars and continue in the read loop. ill look into that

ret2libc (approved):

PR #17514, opened 2020-08-24
Fix #17506 - r2 http://example.org works again ##io

### LINKED ISSUE #17506: Can not open file via http
### Work environment

| OS/arch/bits (mandatory)                             | Ubuntu x86 64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.6.0-git 25587 @ linux-x86-64 git.4.4.0-574-g3c8890013 commit: 3c889001347fb764db98594b97e65b4e22cf9600 build: 2020-08-21__14:19:08

### Expected behavior
r2 opens file via http

```
r2 http://host/path
[r] Cannot open '...
```

trufae:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/cons/input.c
+++ b/libr/cons/input.c
@@ -143,6 +143,7 @@ R_API int r_cons_arrow_to_hjkl(int ch) {
 	}
 	switch (ch) {
 	case 0x1b:
+eprintf ("thats a quit\n");
 		ch = 'q'; // XXX: must be 0x1b (R_CONS_KEY_ESC)
 		break;
 	case 0x4f: // function keys from f1 to f4
@@ -164,6 +165,7 @@ R_API int r_cons_arrow_to_hjkl(int ch) {
 	case '[': // function keys (2)
 		ch = r_cons_readchar ();
 #endif
+eprintf ("rechDOWN\n");
 		switch (ch) {
 		case '<':
 			{
@@ -246,6 +248,7 @@ R_API int r_cons_arrow_to_hjkl(int ch) {
 		case '3':
 			// handle mouse down /up events (35 vs 32)
 			__parseMouseEvent();
+eprintf ("MOUSE DOWN\n");
 			return 0;
 			break;
 		case '2':
--- a/libr/cons/input.c
+++ b/libr/cons/input.c
@@ -143,7 +143,6 @@ R_API int r_cons_arrow_to_hjkl(int ch) {
 	}
 	switch (ch) {
 	case 0x1b:
-eprintf ("thats a quit\n");
 		ch = 'q'; // XXX: must be 0x1b (R_CONS_KEY_ESC)
 		break;
 	case 0x4f: // function keys from f1 to f4
@@ -165,7 +164,6 @@ eprintf ("thats a quit\n");
 	case '[': // function keys (2)
 		ch = r_cons_readchar ();
 #endif
-eprintf ("rechDOWN\n");
 		switch (ch) {
 		case '<':
 			{
@@ -248,7 +246,6 @@ eprintf ("rechDOWN\n");
 		case '3':
 			// handle mouse down /up events (35 vs 32)
 			__parseMouseEvent();
-eprintf ("MOUSE DOWN\n");
 			return 0;
 			break;
 		case '2':
--- a/libr/socket/socket.c
+++ b/libr/socket/socket.c
@@ -19,31 +19,31 @@ R_LIB_VERSION(r_socket);
 
 #if NETWORK_DISABLED
 /* no network */
-R_API RSocket *r_socket_new (bool is_ssl) {
+R_API RSocket *r_socket_new(bool is_ssl) {
 	return NULL;
 }
-R_API bool r_socket_is_connected (RSocket *s) {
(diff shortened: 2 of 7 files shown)


XVilka:
Breaks the [test on Windows](https://ci.appveyor.com/project/radareorg/radare2/builds/34827477/job/ttm9ao30172yauuw#L6497):
```
[XX] C:\projects\radare2\test\db\io\http r2 http://example.org
ANSICON=1 R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'r
b 64
ps
' http://example.org
-- stdout
@@ -1,3 +1,6 @@
 1256
-    <meta charset="utf-8" />
-    <meta http-equiv="Content-type"
+<!doctype html>
+<html>
+<head>
+    <title>Example Domain</title>
+
[**]                
```

libr/cons/input.c
```diff
@@ -143,6 +143,7 @@ R_API int r_cons_arrow_to_hjkl(int ch) {
 	}
 	switch (ch) {
 	case 0x1b:
+eprintf ("thats a quit\n");
```
ret2libc, line 146: please remove these before merging.

ret2libc (changes requested):
Just to ensure debugging printf are removed before merging.

libr/socket/socket.c
```diff
@@ -772,7 +771,7 @@ R_API void r_socket_printf(RSocket *s, const char *fmt, ...) {
 	if (s->fd != R_INVALID_SOCKET) {
 		va_start (ap, fmt);
 		vsnprintf (buf, BUFFER_SIZE, fmt, ap);
-		r_socket_write (s, buf, strlen (buf));
+		(void) r_socket_write (s, buf, strlen (buf));
```
trufae, line 774: r_socket_printf should return int to remove this void cast

libr/socket/socket.c
```diff
@@ -311,7 +311,7 @@ R_API bool r_socket_connect(RSocket *s, const char *host, const char *port, int
 					s->fd = -1;
 					continue;
 				}
-				r_socket_block_time (s, 0, 0, 0);
+				r_socket_block_time (s, 10000, 1000, 0);
```
trufae, line 314: i would like to have some defines to specify some default timeout values here, (and reuse them in the other cases below). which time do you think it shuold be fine to define for waiting until the connection succeds?

from internet:

* In Windows the value is dynamic for established conections, though the default for initial connections is 72 seconds. 

imho thats too much, i would go for having lower timeouts like 5s

ret2libc: No real preference here. 5s seems like a good random value for now!

ret2libc (approved):

libr/socket/socket.c
trufae: Don't aprove it until the implementation is done. you can review it again now

trufae:
Screenshot 2020-08-28 at 11 16 15

GustavoLCR:
@trufae This fixes it:
```diff
diff --git a/libr/socket/socket_http.c b/libr/socket/socket_http.c
index 3aaf7f0ee..9b4efba45 100644
--- a/libr/socket/socket_http.c
+++ b/libr/socket/socket_http.c
@@ -34,7 +34,7 @@ static char *socket_http_answer(RSocket *s, int *code, int *rlen, ut32 redirecti
        r_return_val_if_fail (s, NULL);
        const char *p;
        int ret, len = 0, delta = 0;
-       char *dn;
+       char *dn = NULL;
        RBuffer *b = r_buf_new ();
        if (!b) {
                return NULL;
@@ -47,11 +47,21 @@ static char *socket_http_answer(RSocket *s, int *code, int *rlen, ut32 redirecti
        }
        r_buf_read_at (b, 0, (ut8 *)buf, olen);
        buf[olen] = 0;
-       if ((dn = (char*)r_str_casestr (buf, "\n\n"))) {
-               delta += 2;
-       } else if ((dn = (char*)r_str_casestr (buf, "\r\n\r\n"))) {
-               delta += 4;
+       char *dnn = (char *)r_str_casestr (buf, "\n\n");
+       char *drn = (char *)r_str_casestr (buf, "\r\n\r\n");
+       if (dnn) {
+               if (drn && (drn < dnn)) {
+                       dn = drn;
+                       delta = 4;
+               } else {
+                       dn = dnn;
+                       delta = 2;
+               }
        } else {
+               dn = drn;
+               delta = 4;
+       }
+       if (!dn) {
                goto exit;
        }

diff --git a/test/db/io/http b/test/db/io/http
index 9ba707cb0..ea3f5776b 100644
--- a/test/db/io/http
+++ b/test/db/io/http
@@ -7,8 +7,11 @@ ps
 EOF
 EXPECT=<<EOF
 1256
-    <meta charset="utf-8" />
-    <meta http-equiv="Content-type"
+<!doctype html>
+<html>
+<head>
+    <title>Example Domain</title>
+
 EOF
 RUN

```

radare:
Why red if all CI is green? windows test fixed by @GustavoLCR pr, so imho its good to merge. pls review

PR #17517, opened 2020-08-25
allow dN to work in cursor mode

lowlyw:
(N/A)
N/A

**Detailed description**
before, dN to change the function signature would not work if you are in cursor mode, this fixes it
...

**Test plan**

nope

...

**Closing issues**

nope
...

### DIFF
--- a/libr/core/vmenus.c
+++ b/libr/core/vmenus.c
@@ -3811,7 +3811,7 @@ onemoretime:
 	wordsize = 4;
 	switch (ch) {
 	case 'N':
-		r_core_cmd0 (core, "afs!");
+		r_core_cmdf (core, "afs! @ 0x%08"PFMT64x, off);
 		break;
 	case 'F':
 		{


ret2libc:
What is not working exactly in cursor mode? To me `VdN` is working the same in cursor mode and in non-cursor mode. (Apart from the fact that if I change the name of the variables it does not *actually* change them in the variables section).

lowlyw:
> What is not working exactly in cursor mode? To me `VdN` is working the same in cursor mode and in non-cursor mode. (Apart from the fact that if I change the name of the variables it does not _actually_ change them in the variables section).

@ret2libc the issue is that currently it uses the seek, which is not the same as the offset. offset is computed from the cursor address.

so in this case

image

`afs!` would be run at `0x100001037` instead of at `0x100001066` as the user intended.

you can see this yourself, it only matters if the top of the screen is not at the same function as your cursor is within.

ret2libc:
Oh I see now what you meant. I was thinking about a different case.

ret2libc (approved):

Issue #17518, opened 2020-08-25
r_reg_get_name_idx doesn't check string length


FXTi:
Problem:
When my `get_reg_profile` is called:
> Ignoring duplicated register definition 'PCR'

Here's functions result:
```
=PC	PC
=SP	SP
=R0	D0
...
gpr	PC	.32	72	0
...
gpr	PCR	.32	176	0
...
```

`PC` and `PCR` register should not collide. 
According to my understanding, the issue is that: In `parse_def(RReg *reg, char **tok, const int n)`, it calls `r_reg_get` to verify if given register is already exists. 
And when `PCR` is passed to `r_reg_get`, `=PC	PC` is processed.
So `r_reg_get(..., PCR, -1)` -> `r_reg_get_name_idx(PCR)`, and return `R_REG_NAME_PC` since it doesn't check string length.
In the end, `r_reg_get` find `PC` and `parse_def` think `PCR` is already exists.

---

Also, I will use `r_reg_get` in my code, consider this:

```
=PC RIP
=SP RSP
gpr RSP .64 32  0
gpr ESP .32 32  0
gpr SP  .16 32  0
```

When I try to get `SP`, if I call `r_reg_get(..., SP, -1)`, I will get `RSP`. So I will specify third argument of `r_reg_get` to avoid alias lookup process. Then the only thing need to change is make `r_reg_get_name_idx` check string length == 2.

trufae:
As discussed in Telegram:

* Registers must be lowercase
* Reg aliases must be 2 char in length and uppercase

So, using uppercase registers will work for X86, but not for ARM for example, so my suggestion here is:

* Make all the registers lowercase in your side
* Add the if (!name[0] || !name[1] || name[2]) check in this function to ensure length = 2

PR #17519, opened 2020-08-25
Add length check to r_reg_get_name_idx()

FXTi:

**Detailed description**

Add length check to r_reg_get_name_idx()

**Test plan**

...

**Closing issues**

https://github.com/radareorg/radare2/issues/17518

### DIFF
--- a/libr/reg/reg.c
+++ b/libr/reg/reg.c
@@ -88,7 +88,8 @@ R_API void r_reg_item_free(RRegItem *item) {
 }
 
 R_API int r_reg_get_name_idx(const char *type) {
-	if (type && *type && type[1] && !type[2])
+        r_return_val_if_fail (R_STR_ISNOTEMPTY (type), -1);
+	if (type[1] && !type[2])
 	switch (*type | (type[1] << 8)) {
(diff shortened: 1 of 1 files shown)


libr/reg/reg.c
```diff
@@ -88,7 +88,7 @@ R_API void r_reg_item_free(RRegItem *item) {
 }
 
 R_API int r_reg_get_name_idx(const char *type) {
-	if (type)
+	if (type && *type && type[1] && !type[2])
```
ret2libc, line 91:
```suggestion
        r_return_val_if_fail (R_STR_ISNOTEMPTY (type), -1);
	if (type[1] && !type[2])
```
then I would add a `default` case in the below switch, instead of having a return at the end of function on its own.

ret2libc (changes requested):

libr/reg/reg.c
FXTi: Great. This feature is also great. It's my first time to try this.

ret2libc (approved):
Thanks!

XVilka (approved):

PR #17520, opened 2020-08-25
Do not return a buffer when slurping a empty file

mscherer:
- [X] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

While running r2 from git, I was greeted by:
    
    |ERROR| Invalid command ��Y�' (0x08)
    
I quickly found this came from:
    
    $ ls -l ~/.radare2rc
    -rw-r--r--. 1 misc misc 0 19 avril  2017 /home/misc/.radare2rc
    
Debugging showed that if the file is empty, we return the buffer
that was reallocate earlier with uninitialized memory.
    
Since r_core_cmd_file do not check the size (it pass NULL as a
2nd parameter for r_file_slurp), it doesn't see the file is empty,
and so try to interpret 6 bytes as a command.
    
Given that grep find 54 calls to r_file_slurp with NULL
as a 2nd parameter, this will also fix a bit more than just this
error message.

**Test plan**

* Create a empty file ~/.radare2rc
* Run r2
* See no error message about "|ERROR| Invalid command"

### DIFF
--- a/libr/util/file.c
+++ b/libr/util/file.c
@@ -367,7 +367,7 @@ R_API char *r_file_slurp(const char *str, R_NULLABLE size_t *usz) {
 			} while (!feof (fd));
 			char *nbuf = realloc (buf, size + 1);
 			if (!nbuf) {
-				free(buf);
+				free (buf);
 				return NULL;
 			}
 			buf = nbuf;
--- a/libr/util/file.c
+++ b/libr/util/file.c
@@ -365,6 +365,13 @@ R_API char *r_file_slurp(const char *str, R_NULLABLE size_t *usz) {
 				}
 				size += r;
 			} while (!feof (fd));
+			char *nbuf = realloc (buf, size + 1);
+			if (!nbuf) {
+				free(buf);
+				return NULL;
+			}
+			buf = nbuf;
+			buf[size] = '\0';
 			if (usz) {
 				*usz = size;
 			}
--- a/test/unit/test_util.c
+++ b/test/unit/test_util.c
@@ -113,6 +113,41 @@ bool test_autonames(void) {
 	mu_end;
 }
 
+bool test_file_slurp(void) {
+
+#ifdef __WINDOWS__
+#define S_IRWXU _S_IREAD | _S_IWRITE
+#endif
+
+	const char* test_file =  "./empty_file";
+	size_t s;
+	const char* some_words = "some words";
+
+	int f = open (test_file, O_CREAT, S_IRWXU);
+	mu_assert_neq (f, -1, "cannot create empty file");
+	close (f);
+
+	char* content = r_file_slurp (test_file, &s);
+	mu_assert_eq (s, 0, "size should be zero");
+	mu_assert_eq (strlen (content), 0, "returned buffer should be empty");
+	free (content);
+
+	f = open (test_file, O_WRONLY, S_IRWXU);
+	mu_assert_neq (f, -1, "cannot reopen empty file");
+	write (f, some_words, strlen (some_words));
+	close (f);
+
+	content = r_file_slurp (test_file, &s);
+	mu_assert_eq (s, strlen (some_words), "size should be correct");
+	mu_assert_eq (strlen (content), strlen (some_words), "size for the buffer should be correct");
+	mu_assert_streq (content, some_words, "content should match");
+	free (content);
+
+	unlink (test_file);
+
+	mu_end;
+}
+
 bool test_initial_underscore(void) {
 	Sdb *TDB = setup_sdb ();
 	char *s;
@@ -168,6 +203,7 @@ int all_tests() {
 	mu_run_test (test_dll_names);
 	mu_run_test (test_references);
 	mu_run_test (test_autonames);
+	mu_run_test (test_file_slurp);
 	mu_run_test (test_initial_underscore);
 	return tests_passed != tests_run;
 }


ret2libc (changes requested):
I think this fix is not good from an API point of view. If I call `r_file_slurp` on an empty file I would expect an empty buffer to be returned, not NULL (which is usually used to indicate there was an error). I think the issue here is that `buf` is not NULL terminated, so random stuff is read from there.

I believe the correct fix is to ensure that buf gets NULL terminated even in the `if (!sz) {` path. So callers that do not pass/use `usz` can still work correctly (assuming the file is not a binary file and does not contain \x00 bytes).

ret2libc:
> Ok, I will fix that. I will also add a unit test for that, as soon as I find the right place for them (it seems that ./test/unit/legacy_unit/ is a good place, but "legacy" do not look good, and didn't looked at the documentation yet for how to do it).

Please add the test in `test/unit/test_util.c`. Thanks!

libr/util/file.c
```diff
@@ -365,6 +365,9 @@ R_API char *r_file_slurp(const char *str, R_NULLABLE size_t *usz) {
 				}
 				size += r;
 			} while (!feof (fd));
+			if (!size) {
```
ret2libc, line 368: probably there should always be `buf[size] = '\0';`, no matter what's the value of size.

test/unit/test_util.c
```diff
@@ -113,6 +113,34 @@ bool test_autonames(void) {
...
+	int f = open(nonempty_file, O_CREAT|O_WRONLY);
+	write(f, some_words);
+	close(f);
+
+	content = r_file_slurp(nonempty_file, &s);
+
+	mu_assert_eq (s, strlen(some_words), "size should be correct");
+	mu_assert_eq (strlen(content), strlen(some_words), "size for the buffer should be correct");
+	mu_assert_streq (s, some_words, "content should match");
+}
```
ret2libc, line 142: add `mu_end` before final `}`

mscherer: Done

ret2libc (approved):

test/unit/test_util.c
```diff
@@ -113,6 +113,42 @@ bool test_autonames(void) {
...
+#ifdef __WINDOWS__
+#define S_IRWXU _S_IREAD | _S_IWRITE
+#endif
+
+	char* test_file =  "./empty_file";
+	size_t s;
+	char* content;
+	char* some_words = "some words";
+
+	close(open(test_file, O_CREAT, S_IRWXU));
```
trufae, line 127: missing space before (. also may be worth checking the return value of open to avoid calling close(-1) with an assert

test/unit/test_util.c
```diff
@@ -113,6 +113,42 @@ bool test_autonames(void) {
...
+
+	content = r_file_slurp(test_file, &s);
```
trufae, line 129: defnie and assign the variable in the same line and add a space bfore (

test/unit/test_util.c
```diff
@@ -113,6 +113,42 @@ bool test_autonames(void) {
...
+
+	mu_assert_eq (s, 0, "size should be zero");
+	mu_assert_eq (strlen(content), 0, "returned buffer should be empty");
+
+	free(content);
+
+	int f = open(test_file, O_CREAT|O_WRONLY, S_IRWXU);
```
trufae, line 136: check for the return value of open

trufae: the file wasnt removed, so o_creat shuoldnt be correct tob e used here

libr/util/file.c
```diff
@@ -365,6 +365,7 @@ R_API char *r_file_slurp(const char *str, R_NULLABLE size_t *usz) {
 				}
 				size += r;
 			} while (!feof (fd));
+			buf[size] = '\0';
```
trufae, line 368: change the code of the function to just do a single malloc(sz+1) if you want to null terminate the buffer. the current implementation do realloc for no reason.

test/unit/test_util.c
```diff
@@ -113,6 +113,42 @@ bool test_autonames(void) {
...
+bool test_file_slurp(void) {
+
+#ifdef __WINDOWS__
+#define S_IRWXU _S_IREAD | _S_IWRITE
+#endif
+
+	char* test_file =  "./empty_file";
+	size_t s;
+	char* content;
+	char* some_words = "some words";
```
trufae, line 125: this is const

test/unit/test_util.c
```diff
@@ -113,6 +113,42 @@ bool test_autonames(void) {
 	mu_end;
 }
 
+bool test_file_slurp(void) {
+
+#ifdef __WINDOWS__
+#define S_IRWXU _S_IREAD | _S_IWRITE
+#endif
+
+	char* test_file =  "./empty_file";
```
trufae, line 122: this is also const

test/unit/test_util.c
```diff
@@ -113,6 +113,42 @@ bool test_autonames(void) {
...
+	size_t s;
+	char* content;
+	char* some_words = "some words";
+
+	close(open(test_file, O_CREAT, S_IRWXU));
+
+	content = r_file_slurp(test_file, &s);
+
+	mu_assert_eq (s, 0, "size should be zero");
+	mu_assert_eq (strlen(content), 0, "returned buffer should be empty");
```
trufae, line 132: well. assuming the file is zero, the buffer shuold not contain anything, not even a null byte, because files are binary data not text. But i can get this assumption could be "handy" for some usages of the api

trufae (changes requested):

test/unit/test_util.c
ret2libc: There are many cases where `r_file_slurp` is used assuming that it returns a null-terminated string. This is why I suggested this behaviour (which by the way already always holds except when `sz == 0`)

test/unit/test_util.c
mscherer: done

test/unit/test_util.c
mscherer: also done

test/unit/test_util.c
mscherer: done (both)

test/unit/test_util.c
mscherer: done, even if I am unsure on how we want to react exactly if open fail. I used some assert on the value after a quick reading of others tests cases, but I will dig further.

test/unit/test_util.c
mscherer: done

libr/util/file.c
```diff
@@ -365,6 +365,11 @@ R_API char *r_file_slurp(const char *str, R_NULLABLE size_t *usz) {
 				}
 				size += r;
 			} while (!feof (fd));
+			char *nbuf = realloc (buf, size + 1);
+			if (nbuf) {
```
ret2libc, line 369: I think you should invert the condition and return Null when the reallocation fails.

test/unit/test_util.c
```diff
@@ -113,6 +113,41 @@ bool test_autonames(void) {
...
+    mu_assert_neq (f, -1, "cannot create empty file");
+    close (f);
+
+	char* content = r_file_slurp (test_file, &s);
+	mu_assert_eq (s, 0, "size should be zero");
+	mu_assert_eq (strlen(content), 0, "returned buffer should be empty");
+	free (content);
+
+	f = open (test_file, O_WRONLY, S_IRWXU);
+    mu_assert_neq (f, -1, "cannot reopen empty file");
```
ret2libc, line 136: Pay attention to style please

ret2libc (changes requested):

test/unit/test_util.c
mscherer: Oups, I keep forgetting due to my vim setting :/

libr/util/file.c
ret2libc: I believe it does `realloc` because the size is not known. If it was known, we would not enter this path.

ret2libc (approved):

libr/util/file.c
trufae: the size should be known if its a regular file, only edge case are the files in /proc i think. i didnt tried the code, but it was a blind guess and its unrelated to the PR

test/unit/test_util.c
```diff
@@ -113,6 +113,41 @@ bool test_autonames(void) {
...
+	mu_assert_eq (strlen(content), 0, "returned buffer should be empty");
+	free (content);
+
+	f = open (test_file, O_WRONLY, S_IRWXU);
+	mu_assert_neq (f, -1, "cannot reopen empty file");
+	write (f, some_words, strlen(some_words));
+	close (f);
+
+	content = r_file_slurp(test_file, &s);
+	mu_assert_eq (s, strlen(some_words), "size should be correct");
```
trufae, line 141: some missing spaces before ( in this file

libr/util/file.c
ret2libc: yes indeed the general case is not this path. In general it just does `malloc(sz + 1)`, but this function tries to deal also with some files where it can't determine the size ahead of time, so continuous reallocation are needed.

ret2libc (approved):
@trufae please review again. It looks good to me.

libr/util/file.c
```diff
@@ -365,6 +365,13 @@ R_API char *r_file_slurp(const char *str, R_NULLABLE size_t *usz) {
 				}
 				size += r;
 			} while (!feof (fd));
+			char *nbuf = realloc (buf, size + 1);
+			if (!nbuf) {
+				free(buf);
```
radare, line 370: missing space before (

radare (approved):
Fix the indent and lgtm

libr/util/file.c
ret2libc, line 370:
```suggestion
				free (buf);
```

ret2libc:
I did fix the space. Let's wait for CI and then let's merge.

ret2libc:
Thanks!!

PR #17521, opened 2020-08-25
Fix resource leaks in dwarf.c and dwarf_process.c

HoundThe:

**Detailed description**
Fixes of few leaks found by coverity or me with valgrind.

### DIFF
--- a/libr/anal/dwarf_process.c
+++ b/libr/anal/dwarf_process.c
@@ -1188,9 +1188,11 @@ static st32 parse_function_args_and_vars(Context *ctx, ut64 idx, RStrBuf *args, 
 					}
 					var->type = strdup (r_strbuf_get (&type));
 					r_list_append (variables, var);
+				} else {
+					free (var->location);
+					free (var);
 				}
 				argNumber++;
-				r_strbuf_fini (&type);
 			} else if (child_depth == 1 && child_die->tag == DW_TAG_unspecified_parameters) {
 				r_strbuf_appendf (args, "va_args ...,");
 			} else if (child_die->tag == DW_TAG_variable) { /* child_depth == 1 &&  */
@@ -1216,7 +1218,7 @@ static st32 parse_function_args_and_vars(Context *ctx, ut64 idx, RStrBuf *args, 
 						break;
 					// abstract origin is supposed to have omitted information
 					case DW_AT_abstract_origin:
-						parse_abstract_origin (ctx, val->reference, &type, &name);
+						parse_abstract_origin (ctx, val->reference, &var_type, &name);
 						break;
 					case DW_AT_location:
 						var->location = parse_dwarf_location (ctx, val, find_attr (die, DW_AT_frame_base));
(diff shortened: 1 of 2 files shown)


XVilka (approved):

libr/bin/dwarf.c
```diff
@@ -2447,6 +2460,11 @@ static HtUP *parse_loc_raw(HtUP/*<offset, List *<LocListEntry>*/ *loc_table, con
 			range = NULL;
 		}
 	}
+	/* if for some reason end of list is missing, then loc_list would leak */
+	if (loc_list) { /* looks ugly, recycling existing function so I don't dupe code */
```
ret2libc, line 2464: instead of this ugliness + comment of ugliness you can do:
```C
static void free_loc_table_list(RBinDwarfLocList *loc_list) {
	RListIter *iter;
	RBinDwarfLocRange *range;
	r_list_foreach (loc_list->list, iter, range) {
		free (range->expression->data);
		free (range->expression);
		free (range);
	}
	r_list_free (loc_list->list);
	free (loc_list);
}

static void free_loc_table_entry(HtUPKv *kv) {
	if (kv) {
		free_loc_table_list (kv->value);
	}
}
```

and here just do:
```C
if (loc_list) {
	free_loc_table_list (loc_list);
}
```

libr/anal/dwarf_process.c
```diff
@@ -1188,9 +1188,11 @@ static st32 parse_function_args_and_vars(Context *ctx, ut64 idx, RStrBuf *args,
 					}
 					var->type = strdup (r_strbuf_get (&type));
 					r_list_append (variables, var);
+				} else {
+					free (var->location);
+					free (var);
 				}
 				argNumber++;
```
ret2libc, line 1195: if the argument is ignored (thus we entered the `else` branch) shall this be incremented anyway?

libr/anal/dwarf_process.c
```diff
@@ -1188,9 +1188,11 @@ static st32 parse_function_args_and_vars(Context *ctx, ut64 idx, RStrBuf *args,
 					}
 					var->type = strdup (r_strbuf_get (&type));
 					r_list_append (variables, var);
+				} else {
+					free (var->location);
```
ret2libc, line 1192: i'd prefer to see a local function `variable_free (Variable *)` that correctly frees the whole structure and that is used here, later and also in `parse_function`.

libr/anal/dwarf_process.c
HoundThe: I think yes, because that just means that the argument is there, but we just wasn't able to apply that information, so I'd keep the correct number of the argument so I wouldn't misleading the user

ret2libc: ok ;)

libr/bin/dwarf.c
HoundThe: Yes, that's better

libr/anal/dwarf_process.c
HoundThe: True, done

ret2libc (approved):

PR #17522, opened 2020-08-25
parse_dwarf_location: Guard against possible OOB read

kazarmy:

**Detailed description**

Fixes the following (https://travis-ci.com/github/radareorg/radare2/jobs/377445325#L2921-L2961):

asan-parse_dwarf_location

**Test plan**

The test above in the asan build is fixed. All non-asan builds are green.

**Closing issues**

...

### DIFF
--- a/libr/anal/dwarf_process.c
+++ b/libr/anal/dwarf_process.c
@@ -1083,9 +1083,6 @@ static VariableLocation *parse_dwarf_location (Context *ctx, const RBinDwarfAttr
 			reg_name = get_dwarf_reg_name (ctx->anal->cpu, reg_num, &kind, ctx->anal->bits);
 		} break;
 		case DW_OP_addr: {
-			if (i == block.length - 1) {
-				return NULL;
-			}
 			/* The DW_OP_addr operation has a single operand that encodes a machine address and whose
 			size is the size of an address on the target machine.  */
 			const int addr_size = ctx->anal->bits / 8;
--- a/libr/anal/dwarf_process.c
+++ b/libr/anal/dwarf_process.c
@@ -969,10 +969,13 @@ static VariableLocation *parse_dwarf_location (Context *ctx, const RBinDwarfAttr
 		case DW_OP_fbreg: {
 		/* TODO sometimes CFA is referenced, but we don't parse that yet
 		   just an offset involving framebase of a function*/
+			if (i == block.length - 1) {
+				return NULL;
+			}
 			const ut8 *dump = &block.data[++i];
-			offset = r_sleb128 (&dump, &block.data[loc->block.length]);
+			offset = r_sleb128 (&dump, &block.data[block.length]);
 			if (frame_base) {
-				/* recursive parsing, but frame_base should be only one, but someone 
+				/* recursive parsing, but frame_base should be only one, but someone
 				   could make malicious resource exhaustion attack, so a depth counter might be cool? */
 				VariableLocation *location = parse_dwarf_location (ctx, frame_base, NULL);
 				if (location) {
@@ -1054,6 +1057,9 @@ static VariableLocation *parse_dwarf_location (Context *ctx, const RBinDwarfAttr
 		case DW_OP_breg29:
 		case DW_OP_breg30:
 		case DW_OP_breg31: {
+			if (i == block.length - 1) {
+				return NULL;
+			}
 			/* The single operand of the DW_OP_bregn operations provides 
 			signed LEB128 offset from the specified register.  */
 			reg_num = block.data[i] - DW_OP_breg0; // get the reg number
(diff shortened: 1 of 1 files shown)


libr/anal/dwarf_process.c
```diff
@@ -1062,6 +1068,9 @@ static VariableLocation *parse_dwarf_location (Context *ctx, const RBinDwarfAttr
 			reg_name = get_dwarf_reg_name (ctx->anal->cpu, reg_num, &kind, ctx->anal->bits);
 		} break;
 		case DW_OP_bregx: {
+			if (i == block.length - 1) {
+				return NULL;
+			}
 			/* 2 operands, reg_number, offset*/
 			/* I need to find binaries that uses this so I can test it out*/
 			const ut8 *buffer = &block.data[++i];
```
HoundThe, line 1076: I think there should be an additional if check before the next uleb128() as there are two in this case.

HoundThe (changes requested):
Oh I thought the sleb128() that takes `but_end` as an argument won't read past the bounds, but it's `do {} while {}` so it reads the first byte even if it's out of bounds. I should do a proper expression parsing in future. With comments LGTM.

libr/anal/dwarf_process.c
```diff
@@ -1071,6 +1080,9 @@ static VariableLocation *parse_dwarf_location (Context *ctx, const RBinDwarfAttr
 			reg_name = get_dwarf_reg_name (ctx->anal->cpu, reg_num, &kind, ctx->anal->bits);
 		} break;
 		case DW_OP_addr: {
+			if (i == block.length - 1) {
```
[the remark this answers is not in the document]
kazarmy: Why is it unnecessary? Are you depending on the `if (block.length - i < addr_size)` check?

libr/anal/dwarf_process.c
kazarmy: I'm eoutoftime right now -- will check `r_uleb128()` tomorrow.

libr/anal/dwarf_process.c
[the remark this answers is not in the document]
HoundThe: But I think maybe the check is off by one?

libr/anal/dwarf_process.c
kazarmy: > I think there should be an additional if check before the next uleb128() as there are two in this case.

Okay 604f3715007fe241bf00933cc6a14fd1ca5572a6 ... I think the first check is unnecessary but it's harmless and its psychological impact should not be underestimated.

libr/anal/dwarf_process.c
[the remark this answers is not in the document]
kazarmy: > because there is no leb128() function

Emm the `r_read_ble##()` functions, they don't do bounds-checking (as can be seen from their argument list).

> But I think maybe the check is off by one?

Suppose `i == block.length - 1` and the check is
https://github.com/radareorg/radare2/blob/1cae744fd5487d1b06b37b4bbbdc7e6273e353c6/libr/anal/dwarf_process.c#L1078-L1081

Substituting `block.length - 1` for `i`:

```c
if (block.length - (block.length - 1) < addr_size) {
        return NULL;
}
```

Simplifying:

```c
if (1 < addr_size) {
        return NULL;
}
```

The `return NULL` must happen, otherwise there's going to be an OOB read. Now, `addr_size` is defined as:

https://github.com/radareorg/radare2/blob/1cae744fd5487d1b06b37b4bbbdc7e6273e353c6/libr/anal/dwarf_process.c#L1076

So the question is whether `addr_size` can be 0 i.e. `ctx->anal->bits` is less than 8.

Now I think this can happen, whether by accident or malice, so I'd like to keep the `if (i == block.length - 1)` check (line 1086 above). Do you agree?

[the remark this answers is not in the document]
kazarmy: > So the question is whether `addr_size` can be 0 i.e. `ctx->anal->bits` is less than 8.

This should be:

So the question is whether `addr_size` can be 0 or 1 i.e. `ctx->anal->bits` is less than 16.

But the above makes the case for keeping the check even stronger imo.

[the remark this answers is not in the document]
kazarmy: Ah, silly me

> I don't think it's necessary here

Okay 8cfe33c28c708076e626715c94e04e92a5107f94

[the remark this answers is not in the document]
HoundThe: Sure, makes sense. I agree

libr/anal/dwarf_process.c
HoundThe: I think both checks are useful here outside of the psychological impact, one for each `leb128` as they don't do bound checking on the first byte they read.

kazarmy: I agree, could you approve the pr?

HoundThe (approved):

libr/anal/dwarf_process.c
kazarmy: Actually `r_uleb128()` does check whether the length is 0 but users shouldn't need to know that and there is a need to `return NULL` anyway.

HoundThe: Oh true, might be cool to have same interfaces for boh `leb` and `uleb` in the future and have them to make the same checks :D

kazarmy: Have to verify correctness of each change though :dizzy_face:

PR #17527, opened 2020-08-26
Fix oob read in getFunctionName()

thestr4ng3r:

**Detailed description**

If the function name is less than 4 chars, this will strdup out of bounds. `asm.flags.real` is only for changing the display of things, not how analysis names things, so I think this can be removed completely. Moreover, it is almost always directly overwritten later in `set_fcn_name_from_flag()`.

**Test plan**

Run `r2 -Qc "f a; af" -` with valgrind or asan. The added test triggers the bug too but doesn't crash directly for me because it's just a read in "usually" valid memory. The tests purpose is primarily to ensure this change doesn't alter the functionality.

### DIFF
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -109,13 +109,7 @@ static char *getFunctionName(RCore *core, ut64 addr) {
 			break;
 		}
 	}
-	if (name) {
-		if (r_config_get_i (core->config, "asm.flags.real")) {
-			name += 4;
-		}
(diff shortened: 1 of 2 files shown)


ret2libc (approved):
Honestly... I have no idea why there was that `asm.flags.real` check :/ Anyway, LGTM

thestr4ng3r:
Pretty sure it was for skipping over "sym.", hence the +4.

PR #17531, opened 2020-08-27
Don't try to parse the function if null ##dwarf

lowlyw:
fixes crashes on gcc 4.9.4 and ld 2.35 with score7 target "aaa"

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/anal/dwarf_process.c
+++ b/libr/anal/dwarf_process.c
@@ -1651,15 +1651,15 @@ R_API void r_anal_dwarf_integrate_functions(RAnal *anal, RFlag *flags, Sdb *dwar
 				r_flag_unset_off (flags, offset);
 				r_flag_set_next (flags, global_name, offset, 4);
 				free (global_name);
-			} else if (*kind == 's') {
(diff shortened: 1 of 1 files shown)


ret2libc:
@HoundThe question for you: does it make sense to execute the `sdb_aforeach` loop when no function is found? If yes, is it because of the `g` kind?

HoundThe:
@ret2libc
Oh, the loop should have been inside the 
`if (fcn) {` statement

HoundThe (commented):
Thanks! I think it might be better to move the whole variable loop inside the 
```c
if (fcn) {
    .....
}
```

As the whole loop deals only variables/arguments of that one function

ret2libc:
@HoundThe what about the `type=g` case? There doesn't seem to be reference to `fcn` there.

HoundThe:
@ret2libc `type=g` is a variable of global storage, so it's probably some static variable inside the function. Even it creates

HoundThe (approved):

PR #17537, opened 2020-08-27
Don't add already existing parent class ##anal

HoundThe:

**Detailed description**
I was going through my notes of details, TODOs etc. and I added help string to print help to a new `acg` command with `ac?`, I've also noticed that since we are adding RTTI inheritance info with `aaa` or `avrr` if you do it multiple times, the same base will be added twice, so you might end up with:
```
[0x00001120]> acl
Bat: Bird, Mammal, Bird, Mammal
Bird
Cat: Mammal, Mammal
Dog: Cat, Cat
Kappa: Cat, Cat
Mammal
```
So I've added a check that confirms the base doesn't already exists for a class before adding it

### DIFF
--- a/libr/anal/class.c
+++ b/libr/anal/class.c
@@ -727,8 +727,20 @@ R_API RAnalClassErr r_anal_class_base_set(RAnal *anal, const char *class_name, R
 		free (base_class_name_sanitized);
 		return R_ANAL_CLASS_ERR_NONEXISTENT_CLASS;
 	}
+	RVector /*<RAnalBaseClass>*/ *bases = r_anal_class_base_get_all (anal, class_name);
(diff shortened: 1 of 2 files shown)


trufae:
lgtm but im afraid about the performance because of the O(n) cost of this

HoundThe:
Yeah, I was thinking about that and many languages don't even have multiple inheritance and if they do there aren't more than couple of parents. I'll try to find a large binary with lot of inheritance (in RTTI) and test it.

HoundThe:
I've tried it on a C++ binary with 600 classes and the `avrr` was pretty much instant.

ret2libc (approved):

PR #17544, opened 2020-08-28
Fix compilation warnings ##dwarf

HoundThe:

**Detailed description**
```
./libr/anal/dwarf_process.c:54:8: warning: passing argument 1 of 'free' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  free (var->name);

../test/unit/test_dwarf_integration.c:42:8: warning: unused variable 'object_name' [-Wunused-variable]
  char *object_name = "_cairo_status";
```

### DIFF
--- a/libr/anal/dwarf_process.c
+++ b/libr/anal/dwarf_process.c
@@ -1148,17 +1148,13 @@ static st32 parse_function_args_and_vars(Context *ctx, ut64 idx, RStrBuf *args, 
 		bool get_linkage_name = prefer_linkage_name (ctx->lang);
 		bool has_linkage_name = false;
 		int argNumber = 1;
-		/* TODO deal with lexical block ?? */
-		/* TODO merge the formal_parameter and variable parts */
 		size_t j;
 		for (j = idx; child_depth > 0 && j < ctx->count; j++) {
 			child_die = &ctx->all_dies[j];
 			RStrBuf type;
 			r_strbuf_init (&type);
 			const char *name = NULL;
-			// right now we skip non direct descendants of the structure
-			// TODO maybe add parsing of possible thrown exception DW_TAG_thrown_type
-			if (child_depth == 1 && child_die->tag == DW_TAG_formal_parameter) {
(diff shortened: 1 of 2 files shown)


XVilka:
I am not sure forcing to "char*" of the const pointer is a good idea.

libr/anal/dwarf_process.c
```diff
@@ -51,7 +51,7 @@ typedef struct dwarf_variable_t {
 } Variable;
 
 static void variable_free(Variable *var) {
-	free (var->name);
+	free ((char *) var->name);
```
ret2libc, line 54: I think you should change `Variable.name` to `char *`. Right now `var->name` is set with things like `r_str_newf`, `strdup`, etc., so it is definitely not `const char *`.

ret2libc (changes requested):

libr/anal/dwarf_process.c
HoundThe: Well at first it was set by `const char *` from the raw dwarf information. And because there are 2 possibilitites of the name (name and linkage name) I wanted to not waste extra duplicating and freeing, so I've kept it `const char *` and strdup it at the end when I picked which of the names I want. I've refactored it to use separate variable for this and store the strduped name into the variable so I've removed the const.

libr/anal/dwarf_process.c
```diff
@@ -1182,86 +1178,48 @@ static st32 parse_function_args_and_vars(Context *ctx, ut64 idx, RStrBuf *args,
 						parse_abstract_origin (ctx, val->reference, &type, &name);
 						break;
 					case DW_AT_location:
-						if (val->kind == DW_AT_KIND_BLOCK) {
```
ret2libc, line 1185: why this removal?

ret2libc (approved):

libr/anal/dwarf_process.c
HoundThe: Because we can parse also a location list now, not just blocks (and the handling is inside the parse_location function)

HoundThe:
Not sure why the tests fail right now, the failing ones seems to be unrelated to the PR?

XVilka (approved):

XVilka (approved):

PR #17545, opened 2020-08-28
Add zb score threshold variable ##signatures

swoops:
Book update [pending here](https://github.com/radareorg/radare2book/pull/271/commits)

**Detailed description**

Small change to finally make use of the score threshold variable in zb commands. Setting `e zign.bestthresh = 0.5` will limit requires results to have a similarity score above `0.5`.  Setting a good threshold value should increase the speed of the algorithm by reducing the number of times the Levenshtein Distance is executed.

I am not a big fan of `bestthresh`, so if you have a better name I am happy to change it.  

**Test plan**

Tests were submitted to take care of common edge cases. The bulk of this feature was previously implemented, I just added the variable.

**Closing issues**

...

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -3261,7 +3261,7 @@ R_API int r_core_config_init(RCore *core) {
 	SETBPREF ("zign.autoload", "false", "Autoload all zignatures located in " R_JOIN_2_PATHS ("~", R2_HOME_ZIGNS));
 	SETPREF ("zign.diff.bthresh", "1.0", "Threshold for diffing zign bytes [0, 1] (see zc?)");
 	SETPREF ("zign.diff.gthresh", "1.0", "Threshold for diffing zign graphs [0, 1] (see zc?)");
-	SETPREF ("zign.bestthresh", "0.0", "Threshold for zb commands");
+	SETPREF ("zign.threshold", "0.0", "Minimum similarity required for inclusion in zb output");
 
 	/* diff */
 	SETCB ("diff.sort", "addr", &cb_diff_sort, "Specify function diff sorting column see (e diff.sort=?)");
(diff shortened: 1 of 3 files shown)


libr/core/cconfig.c
```diff
@@ -3261,6 +3261,7 @@ R_API int r_core_config_init(RCore *core) {
 	SETBPREF ("zign.autoload", "false", "Autoload all zignatures located in " R_JOIN_2_PATHS ("~", R2_HOME_ZIGNS));
 	SETPREF ("zign.diff.bthresh", "1.0", "Threshold for diffing zign bytes [0, 1] (see zc?)");
 	SETPREF ("zign.diff.gthresh", "1.0", "Threshold for diffing zign graphs [0, 1] (see zc?)");
+	SETPREF ("zign.bestthresh", "0.0", "Threshold for zb commands");
```
trufae, line 3264: I dont like the var name or the description. But it seems like an useful feature.

Proposals:

- zign.filter
- use zign.threshold. As prefix for diff.b,g,min..
- find better name instead of threshold

For the dexcription i would say zign.diff.* are also poor descriptions. Maybe better to say like “ filter fir matches having minimum threshold for bytes matching”

XVilka: I would say just `zign.threshold` with a better description than just "Threshold for zb commands".

trufae: make clear the idfference with `	SETPREF ("zign.diff.bthresh", "1.0", "Threshold for diffing zign bytes [0, 1] (see zc?)");
`

libr/core/cmd_zign.c
```diff
@@ -1035,6 +1035,16 @@ static RSignItem *item_frm_signame(RAnal *a, const char *signame) {
 	return it;
 }
 
+double get_zb_threshold(RCore *core) {
```
trufae, line 1038: static

trufae (changes requested):

swoops:
Sorry was away for a bit, I was busy being anxious over r2con :)

I made an update, let me know if you like it. I am pretty bad at naming and describing things so feel free to make more suggestions.

trufae (approved):
Looks good to me now! :)

PR #17548, opened 2020-08-29
Add tests for r_id_storage

condret:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/test/unit/test_idstorage.c
+++ b/test/unit/test_idstorage.c
@@ -1,0 +1,86 @@
+#include <r_util.h>
+#include "minunit.h"
+
+bool test_r_id_storage_add0(void) {
+	char const *str = "lol";
+	RIDStorage *ids = r_id_storage_new (5, 23);
+	ut32 id;
+	bool success = r_id_storage_add (ids, str, &id);
+	void *ptr = r_id_storage_get (ids, id);
+	r_id_storage_free (ids);
+	mu_assert (success && (ptr == str), "id_storage_add 0");
+	mu_end;
+}
+
+bool test_r_id_storage_add1(void) {
+	char const *str = "lol";
+	RIDStorage *ids = r_id_storage_new (0, 4);
+	ut32 id;
+	r_id_storage_add (ids, str, &id);
+	r_id_storage_add (ids, str, &id);
+	r_id_storage_add (ids, str, &id);
+	r_id_storage_add (ids, str, &id);
+	bool success = r_id_storage_add (ids, str, &id);
+	r_id_storage_free (ids);
+	mu_assert (!success, "id_storage_add 1");
+	mu_end;
+}
+
+bool test_r_id_storage_set(void) {
+	char const *str = "lol";
+	RIDStorage *ids = r_id_storage_new (5, 23);
+	r_id_storage_set (ids, str, 1);
+	void *ptr = r_id_storage_get (ids, 1);
+	r_id_storage_free (ids);
+	mu_assert_ptreq (ptr, str, "id_storage_set");
+	mu_end;
+}
+
+bool test_r_id_storage_delete(void) {
+	RIDStorage *ids = r_id_storage_new (5, 23);
+	ut32 id;
+	r_id_storage_add (ids, "lol", &id);
+	r_id_storage_delete (ids, id);
+	void *ptr = r_id_storage_get (ids, id);
+	r_id_storage_free (ids);
+	mu_assert_ptreq (ptr, NULL, "id_storage_delete");
+	mu_end;
+}
+
+bool test_r_id_storage_take0(void) {
+	char const *str = "lol";
+	RIDStorage *ids = r_id_storage_new (5, 23);
+	ut32 id;
+	r_id_storage_add (ids, str, &id);
+	void *ptr = r_id_storage_take (ids, id);
+	r_id_storage_free (ids);
+	mu_assert_ptreq (ptr, str, "id_storage_take 0");
+	mu_end;
+}
+
+bool test_r_id_storage_take1(void) {
+	char const *str = "lol";
+	RIDStorage *ids = r_id_storage_new (5, 23);
+	ut32 id;
+	r_id_storage_add (ids, str, &id);
+	r_id_storage_take (ids, id);
+	void *ptr = r_id_storage_get (ids, id);
+	r_id_storage_free (ids);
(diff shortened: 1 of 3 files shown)


test/unit/test_idstorage.c
```diff
@@ -0,0 +1,71 @@
...
+	ut32 id;
+	r_id_storage_add (ids, str, &id);
+	void *ptr = r_id_storage_get (ids, id);
+	r_id_storage_free (ids);
+	mu_assert_ptreq (ptr, str, "id_storage_add");
+	mu_end;
+}
+
+bool test_r_id_storage_set(void) {
+	char *str = "lol";
```
trufae, line 16: This is const

test/unit/test_idstorage.c
```diff
@@ -0,0 +1,71 @@
...
+	r_id_storage_add (ids, "lol", &id);
+	r_id_storage_delete (ids, id);
+	void *ptr = r_id_storage_get (ids, id);
+	r_id_storage_free (ids);
+	mu_assert_ptreq (ptr, NULL, "id_storage_delete");
+	mu_end;
+}
+
+bool test_r_id_storage_take0(void) {
+	char *str = "lol";
```
trufae, line 37: Same

test/unit/test_idstorage.c
```diff
@@ -0,0 +1,71 @@
...
+	mu_run_test (test_r_id_storage_add);
+	mu_run_test (test_r_id_storage_set);
+	mu_run_test (test_r_id_storage_delete);
+	mu_run_test (test_r_id_storage_take0);
+	mu_run_test (test_r_id_storage_take1);
+	return tests_passed != tests_run;
+}
+
+int main(int argc, char **argv) {
+	return all_tests();
```
trufae, line 70: Missing space

test/unit/test_idstorage.c
```diff
@@ -0,0 +1,71 @@
+#include <r_util.h>
+#include "minunit.h"
+
+bool test_r_id_storage_add(void) {
+	char *str = "lol";
+	RIDStorage *ids = r_id_storage_new (5, 23);
+	ut32 id;
+	r_id_storage_add (ids, str, &id);
```
trufae, line 8: Check return value

test/unit/test_idstorage.c
```diff
@@ -0,0 +1,71 @@
...
+	r_id_storage_free (ids);
+	mu_assert_ptreq (ptr, str, "id_storage_set");
+	mu_end;
+}
+
+bool test_r_id_storage_delete(void) {
+	RIDStorage *ids = r_id_storage_new (5, 23);
+	ut32 id;
+	r_id_storage_add (ids, "lol", &id);
+	r_id_storage_delete (ids, id);
```
trufae, line 29: Check return value deleting a valid and an invalid fd

test/unit/test_idstorage.c
```diff
@@ -0,0 +1,71 @@
...
+	mu_assert_ptreq (ptr, NULL, "id_storage_delete");
+	mu_end;
+}
+
+bool test_r_id_storage_take0(void) {
+	char *str = "lol";
+	RIDStorage *ids = r_id_storage_new (5, 23);
+	ut32 id;
+	r_id_storage_add (ids, str, &id);
+	void *ptr = r_id_storage_take (ids, id);
```
trufae, line 41: Difference between get and take?

test/unit/test_idstorage.c
```diff
@@ -0,0 +1,71 @@
...
+	r_id_storage_free (ids);
+	mu_assert_ptreq (ptr, str, "id_storage_take 0");
+	mu_end;
+}
+
+bool test_r_id_storage_take1(void) {
+	char *str = "lol";
```
trufae, line 48: Const

test/unit/test_idstorage.c
trufae: It is valid to add NULL?

trufae: Do we have an R_IO_INVALID_ID UT32_MAX?

trufae (commented):
Check the comments. Looks good so far

test/unit/test_idstorage.c
condret: this would be 2 more tests, ugh

condret: delete is void. why would you care if delete returns success indicator? This test should test if an entry exists after deletion.

test/unit/test_idstorage.c
condret: Adding NULL to the storage can be done in theory, but you should do that if you want to use iteration functions

condret: and yes that limitiation of max io id exists, but it's not part of this test

test/unit/test_idstorage.c
condret: get gets you the element that is adressed by the id.
take does the same but also deletes the element from the idstorage

test/unit/test_idstorage.c
trufae: To ensure the element u tried to delete was there already. Checking and deleting can be racy

test/unit/test_idstorage.c
trufae: Ok thats what i was expecting too but is the only place in r2land where this verb is used

test/unit/test_idstorage.c
```diff
@@ -0,0 +1,86 @@
...
+}
+
+bool test_r_id_storage_delete(void) {
+	RIDStorage *ids = r_id_storage_new (5, 23);
+	ut32 id;
+	r_id_storage_add (ids, "lol", &id);
+	r_id_storage_delete (ids, id);
+	void *ptr = r_id_storage_get (ids, id);
+	r_id_storage_free (ids);
+	mu_assert_ptreq (ptr, NULL, "id_storage_delete");
```
ret2libc, line 46: you could use `mu_assert_null`

ret2libc (approved):
Thanks!

test/unit/test_idstorage.c
condret: this whole api is racy, no mutexes anywhere. we can make this threadsafe at some point, but that would be out of scope of this pr.
What I want to say is, I'm fine with changing delete. but I don't want to do this now. Feel free to modify the pr, if you really think this should be part of this pr

XVilka (approved):

PR #17554, opened 2020-08-31
meson install sdb faster

kazarmy:

**Detailed description**

On Windows with Meson 0.52.0 and Ninja 1.10.1, Meson searches for `install_name_tool` on the PATH before _**every**_ individual sdb install, e.g.:

Meson-install_name_tool

For a long PATH, this can take a significant amount of time. Idk whether this is a bug, and whether the bug is in Meson or Ninja.

This pr short-circuits the above by using an install script. Note that as per https://github.com/mesonbuild/meson/pull/1215, arguments can be passed to install scripts.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/libr/anal/d/meson.build
+++ b/libr/anal/d/meson.build
@@ -81,9 +81,12 @@ foreach file : sdb_files
     command: sdb_gen_cmd,
     depends: sdb_exe,
     build_by_default: true,
-    install: false
+    install: host_machine.system() == 'windows' ? false : true,
+    install_dir: join_paths(r2_sdb, 'fcnsign')
   )
 endforeach
 
-meson.add_install_script('../../../sys/copy_sdb_files' + (host_machine.system() == 'windows' ? '.bat' : '.sh'),
-                         meson.current_build_dir(), join_paths(r2_sdb, 'fcnsign'))
+if host_machine.system() == 'windows'
+  meson.add_install_script('../../../sys/copy_sdb_files.bat',
+                           meson.current_build_dir(), join_paths(r2_sdb, 'fcnsign'))
+endif
--- a/libr/anal/d/meson.build
+++ b/libr/anal/d/meson.build
@@ -81,7 +81,9 @@ foreach file : sdb_files
     command: sdb_gen_cmd,
     depends: sdb_exe,
     build_by_default: true,
-    install: true,
-    install_dir: join_paths(r2_sdb, 'fcnsign')
+    install: false
   )
 endforeach
+
+meson.add_install_script('../../../sys/copy_sdb_files' + (host_machine.system() == 'windows' ? '.bat' : '.sh'),
+                         meson.current_build_dir(), join_paths(r2_sdb, 'fcnsign'))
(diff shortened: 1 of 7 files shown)


kazarmy:
[Description](https://github.com/radareorg/radare2/pull/17554#issue-476028424) updated

ret2libc:
I'm against this PR actually because install scripts does not allow you to uninstall things, so those sdb files won't be deleted by ninja uninstall. Moreover the use of install/install_dir seems to be much more "mesonic".

karliss:
> We don't look for this on import because it will do a useless PATH lookup on non-mac platforms. That can be expensive on some Windows machines (upto 30ms),

Discussion https://github.com/mesonbuild/meson/pull/6612#discussion_r378581401 and the code https://github.com/mesonbuild/meson/blob/7d0393342ac4c94488b9629c4d0f110fe28b2221/mesonbuild/scripts/depfixer.py#L471

It's expensive to do it on import so we do it for each file 🤦‍♂️

I assume caching fails if there is no install_name_tool in path as it can't distinguish between cache not being built and not having install_name_tool.

kazarmy:
> I'm against this PR actually because install scripts does not allow you to uninstall things, so those sdb files won't be deleted by ninja uninstall.

You don't need `ninja uninstall` on Windows right? I personally uninstall on Windows by deleting the folder where r2 is installed.

> Moreover the use of install/install_dir seems to be much more "mesonic".

The installation on Windows is seriously painfully slow, and trimming the PATH seems just as painful :anguished:

ret2libc:
> > I'm against this PR actually because install scripts does not allow you to uninstall things, so those sdb files won't be deleted by ninja uninstall.
> 
> You don't need `ninja uninstall` on Windows right? I personally uninstall on Windows by deleting the folder where r2 is installed.

Maybe not on windows, but this PR uses an install-script also for linux, so I won't be able to correctly uninstall those files even there.

karliss:
How about sending PR to meson? Should be quite simple. Extend the cache from storing only path to path + boolean flag so that in case of install_name_tool missing it doesn't retry the lookup.

kazarmy:
> ... but this PR uses an install-script also for linux ...

It doesn't now 59b3af8f4bcf9378e008ef4d3d3bc01524ae952a

> How about sending PR to meson? Should be quite simple. Extend the cache from storing only path to path + boolean flag so that in case of install_name_tool missing it doesn't retry the lookup.

Sgtm, though this PR should still be faster because it copies the sdbs in one go rather than do the sdb install step a gazillion times. Also, the Meson version on GitHubCI is stuck at 0.52.0.

Anyway, I think this PR can be merged first, and when/if the above Meson PR is merged, this pr should be reverted and then we check whether there is any noticeable performance difference.

ret2libc:
@kazarmy it's not only that... it's that if we start doing with scripts what meson can do by itself, we'll end up with messy things like the current makefile, where you have no idea where something is done or how. I would prefer as well having a PR sent to meson instead and keeping things on our side as they are. Even if this may be slightly faster, as long as meson way is fast enough I think we should be ok.

kazarmy:
@ret2libc Oh ok ... missed the fact that https://github.com/mesonbuild/meson/pull/6612 was merged after Meson 0.52.0 ... Meson 0.55.1 performs much better so I don't think there's any need for that meson PR.

@karliss Thanks for the link!

ret2libc:
Thanks @kazarmy !

PR #17540, opened 2020-08-28
Fixes for <=glibc-2.2 versions

XVilka:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

A continuation of #17468. Continuing to bringing radare2 to the older platforms - older GCC and Glibc.
Old version of Glibc don't have support for `cpu_set_t` types and `CPU_SET` macro. Moreover, the type of `usleep()` is different - it returns void in older versions.

**Test plan**

It should fix some build errors on Debian Potato: https://github.com/XVilka/debian-oldies/tree/master/potato

See more errors at: https://github.com/XVilka/debian-oldies/runs/1040090876?check_suite_focus=true

PR #17564, opened 2020-09-01
Fixes for old glibc (Debian Etch)

XVilka:

**Detailed description**

A continuation of #17540. Continuing to bringing radare2 to the older platforms - older GCC and Glibc.
Old version of Glibc doesn't export `PTRACE_GETEVENTMSG' [ptrace() API](https://www.man7.org/linux/man-pages/man2/ptrace.2.html) types but the Linux kernel >= 2.5.46 provides it. There is no reliable and portable way to check the kernel version during the compilation, thus it's defined for sufficiently old glibc unconditionally.

**Test plan**

It should fix some build errors on Debian Etch: https://github.com/XVilka/debian-oldies/tree/master/etch

### DIFF
--- a/libr/debug/p/native/linux/linux_ptrace.h
+++ b/libr/debug/p/native/linux/linux_ptrace.h
@@ -18,8 +18,8 @@
 // Sadly, there is no reliable and portable way to check the linux kernel
 // version from headers, so we assume it's supported.
 #if defined(__GLIBC__) && defined(__GLIBC_MINOR__) && (__GLIBC__ <= 2) && (__GLIBC_MINOR__ <= 3)
-#if !defined(PTRACE_GETEVENTMSG) && !defined(PTRACE_GETEVENTMSG)
-#define PTRACE_SETOPTIONS 0x4201
+#if !defined(PT_GETEVENTMSG) && !defined(PTRACE_GETEVENTMSG)
+#define PTRACE_GETEVENTMSG 0x4201
 #define PT_GETEVENTMSG PTRACE_GETEVENTMSG
 #endif
 #endif
--- a/libr/debug/p/native/linux/linux_ptrace.h
+++ b/libr/debug/p/native/linux/linux_ptrace.h
@@ -14,6 +14,16 @@
 #define PTRACE_SETSIGINFO 0x4203
 #endif
 
+// A special case of the older Glibc but the kernel newer than 2.5.46
+// Sadly, there is no reliable and portable way to check the linux kernel
+// version from headers, so we assume it's supported.
+#if defined(__GLIBC__) && defined(__GLIBC_MINOR__) && (__GLIBC__ <= 2) && (__GLIBC_MINOR__ <= 3)
+#if !defined(PTRACE_GETEVENTMSG) && !defined(PTRACE_GETEVENTMSG)
+#define PTRACE_SETOPTIONS 0x4201
+#define PT_GETEVENTMSG PTRACE_GETEVENTMSG
+#endif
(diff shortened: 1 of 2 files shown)


libr/debug/p/native/linux/linux_ptrace.h
```diff
@@ -14,6 +14,16 @@
 #define PTRACE_SETSIGINFO 0x4203
 #endif
 
+// A special case of the older Glibc but the kernel newer than 2.5.46
+// Sadly, there is no reliable and portable way to check the linux kernel
+// version from headers, so we assume it's supported.
+#if defined(__GLIBC__) && defined(__GLIBC_MINOR__) && (__GLIBC__ <= 2) && (__GLIBC_MINOR__ <= 3)
+#if !defined(PTRACE_GETEVENTMSG) && !defined(PTRACE_GETEVENTMSG)
+#define PTRACE_SETOPTIONS 0x4201
+#define PT_GETEVENTMSG PTRACE_GETEVENTMSG
```
ret2libc, line 23: if few lines above you are checking `!defined(PTRACE_GETEVENTMSG)` why at this line you assign PT_GETEVENTMSG=PTRACE_GETEVENTMSG ?

libr/debug/p/native/linux/linux_ptrace.h
```diff
@@ -14,6 +14,16 @@
 #define PTRACE_SETSIGINFO 0x4203
 #endif
 
+// A special case of the older Glibc but the kernel newer than 2.5.46
+// Sadly, there is no reliable and portable way to check the linux kernel
+// version from headers, so we assume it's supported.
+#if defined(__GLIBC__) && defined(__GLIBC_MINOR__) && (__GLIBC__ <= 2) && (__GLIBC_MINOR__ <= 3)
+#if !defined(PTRACE_GETEVENTMSG) && !defined(PTRACE_GETEVENTMSG)
```
yossizap: Duplicated check

libr/debug/p/native/linux/linux_ptrace.h
XVilka: A typo, sorry.

libr/debug/p/native/linux/linux_ptrace.h
XVilka: It's outdated.

ret2libc (commented):
Wouldn't it be better to just disable the feature that relies on this macro? I'm not a fan of randomly defining macros that are supposed to be defined by someone else.

XVilka:
It's supported by the kernel, just not exposed in headers of older glibc.

libr/debug/p/native/linux/linux_ptrace.h
yossizap: Phone GitHub is misleading :/

yossizap:
Gdb simply defines these values if they aren't defined without checking kernel/glibc for compatability. Do we really want to maintain ancient kernels with minimal ptrace support beyond fixing the build?

XVilka:
Yes, it it will not require too much effort. I think it shouldn't.

yossizap (approved):

trufae:
Ideally this shouldl be a runtime check, because r2 can be compiled on a different glibc/kernel than the place it runs, but i guess that when you build with a specific sdk you assume the binary will run on systems having the current or superior versions, not the other way around, so it may be a good fix too

XVilka:
Agree about the runtime checks. But that would need much more changes. It's a long term goal.

PR #17569, opened 2020-09-01
Rename new json parser file to json_indent to avoid collisions

ret2libc:
libr.a is built with all the files required to build against r2, so that includes also SDB. As SDB already has a file named json.c, the new json parser in r_util collides with it.


XVilka:
Will solve this problem https://github.com/radareorg/radare2/pull/17563/checks?check_run_id=1056598254#step:4:2392 in https://github.com/radareorg/radare2/pull/17563

XVilka (approved):

trufae (approved):

PR #17594, opened 2020-09-05
Fix asan r2pipe test by updating libasan and using LD_PRELOAD

kazarmy:

**Detailed description**

This pr attempts to fix the following (https://travis-ci.com/github/radareorg/radare2/jobs/381284399#L3420):

asan-r2pipe

by updating to libasan 6 and using LD_PRELOAD.

**Test plan**

The above asan test is fixed. All non-asan builds are green.

**Closing issues**

...


ret2libc:
I will review once it is not draft anymore. I think @kazarmy is still testing what works and what not

kazarmy:
There appears to be at least 1 debug test that can block r2r frequently enough to be a problem.

XVilka:
@abcSup @yossizap  the problematic tests are for the reverse debugging, could you please take a look?

abcSup:
I will take a look soon.

kazarmy:
> @abcSup @yossizap the problematic tests are for the reverse debugging, could you please take a look?

The tests work if there's no blanket `LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libasan.so.6`, so the problem might be with libasan.

binr/r2r/run.c
```diff
@@ -1203,6 +1203,19 @@ R_API R2RTestResultInfo *r2r_run_test(R2RRunConfig *config, R2RTest *test) {
 	}
 	}
 	bool broken = r2r_test_broken (test);
+#if ASAN
+# if !R2_ASSERT_STDOUT
```
ret2libc, line 1207: why is this test required? What happens if R2_ASSERT_STDOUT is not defined?

kazarmy: Assertion output goes to stderr which then invalidates an assumption made by a line further down:

```c
	        (!strstr (out->out, "WARNING:") && !strstr (out->out, "ERROR:") && !strstr (out->out, "FATAL:")))
```

XVilka:
Looks like `ttyname(0)` returns NULL:
```
 [ENOTTY]
    The fildes argument does not refer to a terminal.
```
in that piece of code (`libr/lang/p/pipe.c`):
```c
safe_in = open (ttyname(0), O_RDONLY);
```
Probably just handle this situation instead? Not sure why `ttyname(0)` fails though.

kazarmy:
> Probably just handle this situation instead?

Handled in 82b8b9f1109aef76ac4c1ff4536a46b3261458c7.

binr/r2r/run.c
```diff
@@ -1203,6 +1203,19 @@ R_API R2RTestResultInfo *r2r_run_test(R2RRunConfig *config, R2RTest *test) {
...
+# error R2_ASSERT_STDOUT undefined or 0
+# endif
+	R2RProcessOutput *out = ret->proc_out;
+	if (!success && test->type == R2R_TEST_TYPE_CMD && strstr (test->path, "/dbg")
+	    && (!out->out ||
+	        (!strstr (out->out, "WARNING:") && !strstr (out->out, "ERROR:") && !strstr (out->out, "FATAL:")))
+	    && (!out->err ||
+	        (!strstr (out->err, "Sanitizer") && !strstr (out->err, "runtime error:")))) {
+		broken = true;
```
ret2libc, line 1216: why set to broken and not to fail?

kazarmy: if it's set to fail, the build is not going to be green

ret2libc: maybe i'm missing something...
What is this check about? What kind of tests should fall into this `if`?

kazarmy: Last time when I used Ubuntu 20.04 instead of the Travis Docker image, the output of some of the debug tests changed. I suspect that the differences are due to the tests being dependent on shared libraries from the environment, which can change depending on the distribution used.

Unfortunately, libasan6 [is only available on Ubuntu 20.04 and above](https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=libasan6&searchon=names) so I don't think there's much option regarding the Ubuntu version.

Only 7 tests should be affected by this.

ret2libc: So by doing this we do expect some tests to fail, right?

kazarmy: See https://github.com/radareorg/radare2/pull/17594/checks?check_run_id=1096771366#step:13:49 (starting from line 49 of the "Run tests" section) for the failing debug tests.

ret2libc (approved):

binr/r2r/run.c
kazarmy: If `ubuntu-latest` == `ubuntu-20.04` then I think this check can and should be dropped.

XVilka (approved):

kazarmy:
Actually this pr did not update to libasan 6. Sorry about that.

PR #17599, opened 2020-09-05
Add API for getting Atomic and Typedef Types and remove broken !size

thestr4ng3r:
Filling this template is fun.

**Detailed description**

see commit titles

**Test plan**

run the tests

### DIFF
--- a/libr/anal/type.c
+++ b/libr/anal/type.c
@@ -18,56 +18,6 @@ static char *is_type(char *type) {
 	return NULL;
 }
 
-/*!
- * \brief Save the size of the given datatype in sdb
- * \param sdb_types pointer to the sdb for types
- * \param name the datatype whose size if to be stored
- */
-static void save_type_size(Sdb *sdb_types, char *name) {
-	const char *type = NULL;
-	r_return_if_fail (sdb_types && name);
-	if (!sdb_exists (sdb_types, name) || !(type = sdb_const_get (sdb_types, name, 0))) {
-		return;
-	}
-	char *type_name_size = r_str_newf ("%s.%s.%s", type, name, "!size");
-	r_return_if_fail (type_name_size);
-	int size = r_type_get_bitsize (sdb_types, name);
-	sdb_set (sdb_types, type_name_size, sdb_fmt ("%d", size), 0);
-	free (type_name_size);
-}
-
-/*!
- * \brief Save the sizes of the datatypes which have been parsed
- * \param core pointer to radare2 core
- * \param parsed the parsed c string in sdb format
- */
-static void __save_parsed_type_size(RAnal *anal, const char *parsed) {
-	r_return_if_fail (anal && parsed);
-	char *str = strdup (parsed);
-	if (str) {
-		char *ptr = NULL;
-		int offset = 0;
-		while ((ptr = strstr (str + offset, "=struct\n")) ||
-			(ptr = strstr (str + offset, "=union\n"))) {
-			*ptr = 0;
-			if (str + offset == ptr) {
-				break;
-			}
-			char *name = ptr - 1;
-			while (name > str && *name != '\n') {
-				name--;
-			}
-			if (*name == '\n') {
-				name++;
-			}
-			save_type_size (anal->sdb_types, name);
-			*ptr = '=';
-			offset = ptr + 1 - str;
-		}
-		free (str);
-	}
-}
-
 static char *get_type_data(Sdb *sdb_types, const char *type, const char *sname) {
 	char *key = r_str_newf ("%s.%s", type, sname);
 	if (!key) {
@@ -127,7 +77,6 @@ R_API void r_anal_save_parsed_type(RAnal *anal, const char *parsed) {
 
 	// Now add the type to sdb.
 	sdb_query_lines (anal->sdb_types, parsed);
-	__save_parsed_type_size (anal, parsed);
 }
 
 static int typecmp(const void *a, const void *b) {
@@ -372,7 +321,6 @@ static void save_struct(const RAnal *anal, const RAnalBaseType *type) {
 		Sdb:
 		name=struct
 		struct.name=param1,param2,paramN
-		struct.name.!size=96
 		struct.name.param1=type,0,0
 		struct.name.param2=type,4,0
 		struct.name.paramN=type,8,0
@@ -404,12 +352,6 @@ static void save_struct(const RAnal *anal, const RAnalBaseType *type) {
 	char *key = r_str_newf ("%s.%s", kind, sname);
 	sdb_set (anal->sdb_types, key, r_strbuf_get (&arglist), 0);
 	free (key);
-	// struct.name.!size=96
-	key = r_str_newf ("%s.%s.!size", kind, sname);
-	char *val = r_str_newf ("%" PFMT64u "", type->size);
-	sdb_set (anal->sdb_types, key, val, 0);
-	free (val);
-	free (key);
 
 	free (sname);
 
@@ -428,7 +370,6 @@ static void save_union(const RAnal *anal, const RAnalBaseType *type) {
 	Sdb:
 	name=union
 	union.name=param1,param2,paramN
-	union.name.!size=32
 	union.name.param1=type,0,0
 	union.name.param2=type,0,0
 	union.name.paramN=type,0,0
@@ -459,12 +400,6 @@ static void save_union(const RAnal *anal, const RAnalBaseType *type) {
 	// union.name=arg1,arg2,argN
 	char *key = r_str_newf ("%s.%s", kind, sname);
 	sdb_set (anal->sdb_types, key, r_strbuf_get (&arglist), 0);
-	free (key);
-
-	key = r_str_newf ("%s.%s.!size", kind, sname);
-	char *val = r_str_newf ("%" PFMT64u "", type->size);
-	sdb_set (anal->sdb_types, key, val, 0);
-	free (val);
 	free (key);
 
 	free (sname);
@@ -518,12 +453,6 @@ static void save_enum(const RAnal *anal, const RAnalBaseType *type) {
 	// enum.name=arg1,arg2,argN
 	char *key = r_str_newf ("enum.%s", sname);
 	sdb_set (anal->sdb_types, key, r_strbuf_get (&arglist), 0);
-	free (key);
-
-	key = r_str_newf ("enum.%s.!size", sname);
-	char *val = r_str_newf ("%" PFMT64u "", type->size);
-	sdb_set (anal->sdb_types, key, val, 0);
-	free (val);
 	free (key);
 
 	free (sname);
--- a/libr/anal/type.c
+++ b/libr/anal/type.c
@@ -279,35 +279,81 @@ error:
 	return NULL;
 }
 
+static RAnalBaseType *get_typedef_type(RAnal *anal, const char *sname) {
+	r_return_val_if_fail (anal && sname, NULL);
+
+	RAnalBaseType *base_type = r_anal_base_type_new (R_ANAL_BASE_TYPE_KIND_TYPEDEF);
+	if (!base_type) {
+		return NULL;
+	}
+
+	base_type->type = get_type_data (anal->sdb_types, "typedef", sname);
+	if (!base_type->type) {
+		goto error;
+	}
+	return base_type;
+
+error:
+	r_anal_base_type_free (base_type);
+	return NULL;
+}
+
+static RAnalBaseType *get_atomic_type(RAnal *anal, const char *sname) {
+	r_return_val_if_fail (anal && sname, NULL);
+
+	RAnalBaseType *base_type = r_anal_base_type_new (R_ANAL_BASE_TYPE_KIND_ATOMIC);
+	if (!base_type) {
+		return NULL;
+	}
+
+	base_type->type = get_type_data (anal->sdb_types, "type", sname);
+	if (!base_type->type) {
+		goto error;
+	}
+
+	RStrBuf key;
+	r_strbuf_init (&key);
+	r_strbuf_setf (&key, "type.%s.size", sname);
+	base_type->size = sdb_num_get (anal->sdb_types, r_strbuf_get (&key), 0);
+	r_strbuf_fini (&key);
+
+	return base_type;
+
+error:
+	r_anal_base_type_free (base_type);
+	return NULL;
+}
+
 // returns NULL if name is not found or any failure happened
 R_API RAnalBaseType *r_anal_get_base_type(RAnal *anal, const char *name) {
 	r_return_val_if_fail (anal && name, NULL);
 
 	char *sname = r_str_sanitize_sdb_key (name);
 	const char *type = sdb_const_get (anal->sdb_types, sname, NULL);
-
-	// Right now just types: struct, enum, union are supported
-	if (!type || !(strcmp (type, "enum") || strcmp (type, "struct") || strcmp (type, "union"))) {
+	if (!type) {
 		free (sname);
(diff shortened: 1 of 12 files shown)


libr/anal/type.c
```diff
@@ -330,35 +279,85 @@ static RAnalBaseType *get_union_type(RAnal *anal, const char *sname) {
...
+
+	base_type->type = get_type_data (anal->sdb_types, "type", sname);
+	if (!base_type->type) {
+		goto error;
+	}
+
+	RStrBuf key;
+	r_strbuf_init (&key);
+	r_strbuf_setf (&key, "type.%s.size", sname);
+	const char *sdb_size = sdb_const_get (anal->sdb_types, r_strbuf_get (&key), 0);
```
trufae, line 317: Why not getnum?

trufae (changes requested):

test/unit/test_sdb.h
```diff
@@ -0,0 +1,21 @@
+
```
ret2libc, line 1:
```suggestion
```

test/unit/test_sdb.h
```diff
@@ -0,0 +1,21 @@
...
+#ifndef TEST_SDB_H
+#define TEST_SDB_H
+
+static void diff_cb(const SdbDiff *diff, void *user) {
```
ret2libc, line 5: what about just putting these in min_unit.h?

thestr4ng3r: I'm not a huge fan of this small extra-header either, but putting it in `minunit.h` would mean that every test must include `sdb.h`.

libr/anal/type.c
thestr4ng3r: done

test/unit/test_sdb.h
ret2libc: Good point yeah. Though any unit test probably needs sdb.h anyway, because sdb.h is required by RUtil and anything in r2world requires RUtil.

test/unit/test_anal_types.c
```diff
@@ -125,9 +205,120 @@ static bool test_anal_get_base_type_enum(void) {
...
+	mu_assert_notnull (anal->sdb_types, "Couldn't create new RAnal.sdb_types");
+
+	setup_sdb_for_atomic (anal->sdb_types);
+
+	RAnalBaseType *base = r_anal_get_base_type (anal, "char");
+	mu_assert_notnull (base, "Couldn't create get base type of atomic type \"char\"");
+
+	mu_assert_eq (R_ANAL_BASE_TYPE_KIND_ATOMIC, base->kind, "Wrong base type");
+	mu_assert_streq (base->name, "char", "type name");
+	mu_assert_streq (base->type, "c", "atomic type type");
```
ret2libc, line 289: sorry for the stupid question... `c` stands for what? And the `base->size` is in bits?

libr/anal/type.c
```diff
@@ -330,35 +279,81 @@ static RAnalBaseType *get_union_type(RAnal *anal, const char *sname) {
...
+
+	base_type->type = get_type_data (anal->sdb_types, "type", sname);
+	if (!base_type->type) {
+		goto error;
+	}
+
+	RStrBuf key;
+	r_strbuf_init (&key);
+	r_strbuf_setf (&key, "type.%s.size", sname);
+	base_type->size = sdb_num_get (anal->sdb_types, r_strbuf_get (&key), 0);
```
ret2libc, line 317:
```suggestion
	base_type->size = sdb_num_get (anal->sdb_types, sdb_fmt ("type.%s.size", sname), 0);
```

ret2libc (approved):
For what I understand about the code, it looks very good :) just nitpicking on a couple of things.

test/unit/test_anal_types.c
thestr4ng3r: `c` is char, which comes from the sdb format. Yes, `base->size` is in bits, there is a comment about it in the struct definition.

libr/anal/type.c
thestr4ng3r: `sdb_fmt()` is bad

ret2libc: Yeah we know, but it is used everywhere anyway. It won't be this single place to save r2world :D (nor to ruin it btw)

libr/anal/type.c
```diff
@@ -330,35 +279,81 @@ static RAnalBaseType *get_union_type(RAnal *anal, const char *sname) {
...
+	}
+
+	base_type->type = get_type_data (anal->sdb_types, "type", sname);
+	if (!base_type->type) {
+		goto error;
+	}
+
+	RStrBuf key;
+	r_strbuf_init (&key);
+	r_strbuf_setf (&key, "type.%s.size", sname);
```
trufae, line 316: whats the point on using strbuf here?

libr/anal/type.c
```diff
@@ -330,35 +279,81 @@ static RAnalBaseType *get_union_type(RAnal *anal, const char *sname) {
...
+	}
+	return base_type;
+
+error:
+	r_anal_base_type_free (base_type);
+	return NULL;
+}
+
+static RAnalBaseType *get_atomic_type(RAnal *anal, const char *sname) {
+	r_return_val_if_fail (anal && sname, NULL);
```
trufae, line 302: also check if *sname

libr/anal/type.c
```diff
@@ -330,35 +279,81 @@ static RAnalBaseType *get_union_type(RAnal *anal, const char *sname) {
 	return NULL;
 }
 
+static RAnalBaseType *get_typedef_type(RAnal *anal, const char *sname) {
+	r_return_val_if_fail (anal && sname, NULL);
```
trufae, line 283: check for *sname

libr/anal/type.c
```diff
@@ -555,6 +530,10 @@ static void save_atomic_type(const RAnal *anal, const RAnalBaseType *type) {
 	r_strbuf_setf (&val, "%" PFMT64u "", type->size);
 	sdb_set (anal->sdb_types, r_strbuf_get (&key), r_strbuf_get (&val), 0);
 
+	r_strbuf_setf (&key, "type.%s", sname);
```
trufae, line 533: same here. i assume you can just malloc once because you know the size already

test/unit/test_sdb.h
```diff
@@ -0,0 +1,20 @@
...
+	char buf[2048];
+	if (sdb_diff_format (buf, sizeof(buf), diff) < 0) {
```
trufae, line 6: missing space before (

test/unit/test_dwarf_integration.c
```diff
@@ -257,4 +254,4 @@ int all_tests(void) {
 
 int main(int argc, char **argv) {
 	return all_tests();
```
trufae, line 256: missing space

test/unit/test_anal_types.c
trufae: c is char in pf format, not sdb

trufae (changes requested):
Can you add an r2 test (not unit) getting atomic and typedef types? i dont get how that's suposed to work by parsing the C expr instead of using the API? Also, how is the size of types accessible now that the !size attribute is not there?

libr/anal/type.c
ret2libc: if so, use `R_STR_ISNOTEMPTY` or `!R_STR_ISEMPTY`

libr/anal/type.c
thestr4ng3r: It's slightly more efficient than r_str_newf() when the string fits in the small stack buffer.

libr/anal/type.c
thestr4ng3r: Well we want to get rid of it eventually, so better not use it at all in new code.

test/unit/test_anal_types.c
thestr4ng3r: It is also used in sdb, for example here:
https://github.com/radareorg/radare2/blob/fe5b80d615881bd1c31e0a99d655f4af90acf126/libr/anal/d/types.sdb.txt#L41
In the long run, this should probably be made an enum (and types should be outside of sdb) but that is the current format.

thestr4ng3r:
> Can you add an r2 test (not unit) getting atomic and typedef types?

Just pushed a test for typedef, but it's not directly related to this yet because the commands aren't using the API yet. They should be refactored to use the API too but I think that would bloat the pr too much.

> i dont get how that's suposed to work by parsing the C expr instead of using the API?

What do you mean by C expr? There is no C expression parsing involved in this pr.

> Also, how is the size of types accessible now that the !size attribute is not there?

The most correct way to get the size currently is this:
https://github.com/radareorg/radare2/blob/fe5b80d615881bd1c31e0a99d655f4af90acf126/libr/util/utype.c#L122
Its implementation is quite broken though so it needs fixing. Previously the `!size` was just the return value of this function, possibly outdated when things changed so you shouldn't have a regression compared to before when using this function directly.

Because this function might get a bit heavy depending on how hard your types recurse, it might make sense to have something like a lazy cache and invalidating depending on dependencies between types, but I would only implement that if it turns out to be a bottleneck in practivce., otherwise it's premature optimization.

libr/anal/type.c
trufae: I know. But we had 5 different implementations of it and end up to the trash

libr/anal/type.c
trufae: What about having an initf vonstructor with format?

thestr4ng3r: Sounds like a good idea to me, but out of scope here.

thestr4ng3r: done: https://github.com/radareorg/radare2/pull/17634

trufae: yeah, please update this pr when the initf gets merged, it lgtm, waiting for @ret2libc to review it

libr/anal/type.c
trufae: im totally fine to replace sdb_fmt with strbuf, so for me this thread is resollved

libr/anal/type.c
```diff
@@ -555,6 +530,10 @@ static void save_atomic_type(const RAnal *anal, const RAnalBaseType *type) {
...
+	r_strbuf_setf (&val, "%s", type->type);
```
trufae, line 534: unneessary format, and get. just pass type->type below

test/unit/test_anal_types.c
trufae: yep i was just trying to clarify

libr/anal/type.c
```diff
@@ -330,35 +279,81 @@ static RAnalBaseType *get_union_type(RAnal *anal, const char *sname) {
...
-		break;
-	case 'e': // enum
+	} else if (!strcmp (type, "enum")) {
 		base_type = get_enum_type (anal, sname);
-		break;
-	case 'u': // union
+	} else if (!strcmp (type, "union")) {
 		base_type = get_union_type (anal, sname);
-		break;
+	} else if (!strcmp (type, "typedef")) {
```
trufae, line 345: right now `struct C` is the same as `typedef struct {} C` in r2 types. both are accessiblel under the name "C", i assume this change wont affect the current C parser. but do you think it should be treated differently?

libr/anal/type.c
```diff
@@ -330,35 +279,81 @@ static RAnalBaseType *get_union_type(RAnal *anal, const char *sname) {
...
+		base_type = get_typedef_type (anal, sname);
+	} else if (!strcmp (type, "type")) {
```
trufae, line 347: uhm how that would look in C? i mean for the cparse side.

test/unit/test_anal_types.c
```diff
@@ -98,6 +146,37 @@ static bool test_anal_get_base_type_union(void) {
...
 
+static bool test_anal_save_base_type_union(void) {
+	RAnal *anal = r_anal_new ();
+	mu_assert_notnull (anal, "Couldn't create new RAnal");
+	mu_assert_notnull (anal->sdb_types, "Couldn't create new RAnal.sdb_types");
+
+	RAnalBaseType *base = r_anal_base_type_new (R_ANAL_BASE_TYPE_KIND_UNION);
+	base->name = strdup ("kappa");
+
+	RAnalUnionMember member;
```
trufae, line 157: use = { .offset = 0, type = ... in multiline

trufae (changes requested):

libr/anal/type.c
thestr4ng3r: I think the behavior that `struct C` must always be used as `struct C` instead of just `C` when it is not a typedef is very C-specific (C++ already doesn't enforce that) and not necessarily relevant for reverse engineering generic binaries written in any language.
Also, currently `typedef struct {} C` isn't directly representable in r2 because we can't have anonymous structs before maybe refactoring the types database out of sdb later.

libr/anal/type.c
thestr4ng3r: cparse as in parsing a type string like "char **[]" or parsing a C header? Atomic types can't be created in C, they are only for specifying the exact properties of `char`, `int`, etc. and usually you will have them loaded by default already.

thestr4ng3r:
The failing test comes from this change:
https://github.com/radareorg/radare2/pull/17599/files#diff-a36bd1a27956d24560ff432fde115fc5R534

Because the behavior that is tested depends on this to be in sdb:
20200913_21h42m03s_grim

libr/anal/type.c
trufae: for me atomic type have a different meaning, which is basically data types that cant introduce race codnitions. not sure if you are refering to the same or you are just talking about the basic types which have nothing to do with atomic. in that case i would suggest a different name.

libr/anal/type.c
trufae: it is representable right now because it uses the typedefed name and the struct name is ignored. but it works as i said. you can test it. i did that on purpose for simplicity btw, so im aware of that behaviour and this is why im asking for feedback

trufae (changes requested):
Make it green

libr/anal/type.c
HoundThe: I think I was the one who named them atomic types, as basic types would kinda collide, I thought of atomic as a further undivisable (internally unstructured).

thestr4ng3r: I think atomic is actually a very good name here. "atomic" indeed can refer to the multi-threading things, but imo having this meaning for a typekind here doesn't make much sense so it kind of rules itself out.
Instead, "atomic" or "atoms" is also a very common term for entities in formulas that can't be split up further: https://en.wikipedia.org/wiki/Atomic_formula
which is pretty much exactly what we have here.

libr/anal/type.c
thestr4ng3r: Yes, I think considering the current database design, this behavior is good.

test/unit/test_anal_types.c
thestr4ng3r: Ok, now I understand what you actually meant by that. I think we shouldn't save the concrete pf format in the sdb but just something like "this is some char", "this is some float", "this is signed/unsigned" from which in combination with the bitsize the pf format could be easily derived. I think I'll open an issue for that.

thestr4ng3r:
After close inspection I can say that the new behavior of the test is the correct one.

test/unit/test_anal_types.c
thestr4ng3r: https://github.com/radareorg/radare2/issues/17649

thestr4ng3r:
No idea what's wrong with the `db/cmd/cmd_tc tc typedef`, I have yet to check that.

trufae (approved):

PR #17603, opened 2020-09-06
Create new GitHub workflow for continuous build ##build

### LINKED ISSUE #17598: New 'continuous' CI release breaks version/release tracking

New "continuous" CI release breaks release and version tracking through GitHub API. Release and version tracking is useful when automating tool packaging for third-party providers and keeping it up-to-date with official releases instead of master branch.
Example:
```
curl https://api.github.com/repos/radareorg/radare2/releases/latest
```
Returns release with tag 'continuous', which is not useful and will be always the same. Instead, it would be useful to see the latest real release, in this case the one with tag 4.5.1.

Continuous release could be marked as 'pre-release`. Then it is not included, when making the request to see the latest actual release.

ret2libc:
Avoid polluting the Releases page with a lot of "continuous build"
releases, which also break the "Latest release" shown by API/UI. This
job will upload as artifacts the .deb packages created on each
commit and create a pre-release "Continuous build" instead of an actual one.

**Closing issues**

Closes #17598

### DIFF
--- a/.github/upload.sh
+++ b/.github/upload.sh
@@ -1,0 +1,135 @@
+#!/bin/bash
+
+set +x # Do not leak information
+
+# Exit immediately if one of the files given as arguments is not there
+# because we don't want to delete the existing release if we don't have
+# the new files that should be uploaded 
+for file in "$@"
+do
+    if [ ! -e "$file" ]
+    then echo "$file is missing, giving up." >&2; exit 1
+    fi
+done
+
+if [ $# -eq 0 ]; then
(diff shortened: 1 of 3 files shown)


XVilka (approved):

ret2libc:
I've tested it a bit on my personal fork. It seems to work. I had to create pre-releases anyway because artifacts cannot be downloaded by unregistered users, so it's quite useless. Also, I ended up using a modified version of upload.sh script(that was already download+used by travis) because there is no official github action to overwrite existing releases.

PR #17607, opened 2020-09-08
Fix multiple bugs in matching ##signatures

swoops:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [x] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

This refactors out multiple bugs in the matching of signatures. When a signature is matched, either `fcnMatchCB` or `searchHitCB` is called. This previous lead to `addFlag` where there were a few bugs. These bugs prevented the function from being renamed, having types applied and resulted in some leaks.

There is still a bug where types are not applied because the `RSignItem` passed to `searchHitCB` does not have a `types` member. This is due to a bug in `r_sign_item_dup` and is addressed in #17608.

**Test plan**
```
$ r2 static 
 -- Did you know that radare2 can decompile to assembly?
[0x00401c60]> s 0x0041dec0
[0x0041dec0]> af
[0x0041dec0]> afs
void fcn.0041dec0 (int64_t arg1);
[0x0041dec0]> zo ./out.sdb 
[0x0041dec0]> z*
zs *
za sym.__strdup b f30f1efa4154554889fd4883ec08e83d69f9ff4c8d60014c89e7e8d167f9ff4885c074154883c4084c89e24889ee4889c75d415ce98769f9ff5a31c05d415cc3:ffffffffffffffffffffffffffffff00000000ffffffffffffffff00000000ffffffff00ffffffffffffffffffffffffffffffffff00000000ffffffffffffff
za sym.__strdup g cc=2 nbbs=3 edges=3 ebbs=1 bbsum=64
za sym.__strdup o 0x0008fb30
za sym.__strdup v r110
za sym.__strdup t func.sym.__strdup.args=1 func.sym.__strdup.arg.0="uint32_t,AAAA"
za sym.__strdup h d8b82299783a827712112871062c12f183da2579815bbb644bec13b71fac5618
[0x0041dec0]> z.
[+] searching 0x0041dec0 - 0x0041dfc0
[+] searching function metrics
hits: 1
[0x0041dec0]> afs
void sym.__strdup (uint32_t AAAA);
[0x0041dec0]> 
```
[...]

PR #17608, opened 2020-09-08
Remove r_sign_item_dup ##signatures

swoops:

**Detailed description**

The `r_sign_item_dup` appears to only be used in one place. This is to set up a list of `RSignItems` for `r_search`. The items don't need to be duplicated if they are never free'd. This should result in a performance boost. 

The `r_sign_item_dup` function also has a bug where it does not properly duplicate the `items` signature. So by refactoring this function out, the `searchHitCB` should now get the necessary information to apply the type information to a function.

**Test plan**

See the command executed in #17607 PR.

**Closing issues**

...

### DIFF
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -2184,6 +2184,7 @@ R_API void r_sign_space_rename_for(RAnal *a, const RSpace *space, const char *on
 struct ctxForeachCB {
 	RAnal *anal;
 	RSignForeachCallback cb;
+	bool freeit;
 	void *user;
 };
 
@@ -2200,13 +2201,21 @@ static bool foreachCB(void *user, const char *k, const char *v) {
 	} else {
 		eprintf ("error: cannot deserialize zign\n");
 	}
-	r_sign_item_free (it);
+	if (ctx->freeit) {
+		r_sign_item_free (it);
+	}
 	return true;
+}
+
+static bool r_sign_foreach_nofree(RAnal *a, RSignForeachCallback cb, void *user) {
+	r_return_val_if_fail (a && cb, false);
+	struct ctxForeachCB ctx = { a, cb, false, user };
+	return sdb_foreach (a->sdb_zigns, foreachCB, &ctx);
 }
 
 R_API bool r_sign_foreach(RAnal *a, RSignForeachCallback cb, void *user) {
 	r_return_val_if_fail (a && cb, false);
-	struct ctxForeachCB ctx = { a, cb, user };
+	struct ctxForeachCB ctx = { a, cb, true, user };
 	return sdb_foreach (a->sdb_zigns, foreachCB, &ctx);
 }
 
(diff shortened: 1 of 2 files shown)


XVilka (approved):

ret2libc (commented):
I'm not so sure the `_dup` is not required. Those `RSignItem`s get added to `ss->items` and `ss->items` is freed in `r_sign_search_free`, for example.

trufae (approved):

ret2libc (changes requested):
Just to make sure none merge the PR without first checking what said in previous comment.

swoops:
> I'm not so sure the `_dup` is not required. Those `RSignItem`s get added to `ss->items` and `ss->items` is freed in `r_sign_search_free`, for example.

The `_dup` was required previously because the only `foreach` function would allow the callback to *borrow* the `RSignItem`, but not keep it. This is because the `foreach` kept a reference to each `RSignItem` and free'd that item on each loop.

This code ads a `foreach_no_free` function which never free's the `RSignItem`s it produces. This means the item can be given away and put inside of `ss->items`. The end result is the same, `ss->items` contains all the `RListItems`. They are just the items given away by the `foreach_no_free` function instead duplicates of the items borrowed from `foreach`.

ret2libc:
@swoops if you look at `cmd_zign.c:searchRange` there are calls, in this order, to: `r_sign_search_new`+`r_sign_search_init`+`r_sign_search_update`+`r_sign_search_free`. In `r_sign_search_new` `ret->items` is initialized with `r_list_newf (r_sign_item_free)` (see https://github.com/radareorg/radare2/blob/master/libr/anal/sign.c#L2217). Then, in `r_sign_search_free` there is a call to `r_list_free (ss->items)` (see https://github.com/radareorg/radare2/blob/master/libr/anal/sign.c#L2227) which not only free the rlistiterator, but it also calls the function `r_sign_item_free`. If the RSignItems are not duplicated, as in your PR, this ends up in double free.

Am I missing anything?

swoops:
Everything you said is correct, I think you are missing the removal of a free from whiten `r_sign_search_innit` call chain. Here is how it would work before the PR:

1. Create item `it` (bt: `foreachCB`, `r_sign_foreach`, `r_sign_search_init`)
2. Duplicate `it` into `it2`(bt: `addSearchKwCB`, `foreachCB`, `r_sign_foreach`, `r_sign_search_init`)
3. Add `it2` to `ss->items` (bt: `addSearchKwCB`, `foreachCB`, `r_sign_foreach`, `r_sign_search_init`)
4. free `it` (bt: `foreachCB`, `r_sign_foreach`, `r_sign_search_init`)
5. use `ss->items` in search
6. Free `ss->items` in `r_sign_search_free`

Now it looks like this:

1. Create item `it` (bt: `foreachCB`, `r_sign_foreach_nofree`, `r_sign_search_init`)
2. REMOVED
3. Modified: Add *original* `it` to `ss->items` (bt: `addSearchKwCB`, `foreachCB`, `r_sign_foreach_nofree`, `r_sign_search_init`)
4. REMOVED 
5. use `ss->items` in search
6. Free `ss->items` in `r_sign_search_free`

Notice the `r_sign_foreach_nofree` is used instead of `r_sign_foreach`. The `r_sign_foreach_nofree` sets `ctx.freeit == false` causing the `foreachCB` to skip the call to `r_sign_item_free`. As a result the `it` is not free'd in step 4. So it will still exist for steps 5 and 6.

ret2libc (approved):
Ok I think I kinda understand now! Thanks for the explanation.

PR #17612, opened 2020-09-08
Implement ?== operator to compare strings ##shell

trufae:

**Detailed description**

needed because:

* theres no way to strcmp in r2
* x86.as tests are x86-specific

other solutions for the 2nd problem are:

- add ARCHONLY= in r2r
- do some r2 scripting magic
- make arch-specific tests in a separate directory
- install x86 toolchain on non-x86 testsuite platforms
- fallback to x86.nz inside the plugin

but also bear in mind that there are more failing x86.as tests that must be addressed, this PR is only fixing one

```
$ git grep x86.as test
test/db/esil/x86_32:e asm.assembler=x86.as
test/db/esil/x86_32:e asm.assembler=x86.as
test/db/tools/rasm2:CMDS=!rasm2 -s att -a x86.as -b 64 "test %rbx, %rax" 
test/db/tools/rasm2:CMDS=!rasm2 -s intel -a x86.as -b 64 "test rax, rbx" 
test/db/tools/rasm2:CMDS=!rasm2 -s att -a x86.as -b 64 "add \$33, %rbx"
test/db/tools/rasm2:CMDS=!rasm2 -s intel -a x86.as -b 64 "add rbx, 33" 
test/db/tools/rasm2:NAME=rasm2 -s att -a x86.as -b 64 movq %rdx, %rax
test/db/tools/rasm2:CMDS=!rasm2 -s att -a x86.as -b 64 "movq %rdx, %rax"
test/db/tools/rasm2:NAME=rasm2 x86.asm.32
```

**Test plan**

just see travis for arm64

**Closing issues**

there are some issues tracking XX on non-x86, lazy to find now, and this is just fixing some of them

### DIFF
--- a/libr/core/cmd_help.c
+++ b/libr/core/cmd_help.c
@@ -127,7 +127,7 @@ static const char *help_msg_question[] = {
 	"?+", " [cmd]", "run cmd if $? > 0",
 	"?-", " [cmd]", "run cmd if $? < 0",
 	"?=", " eip-0x804800", "hex and dec result for this math expr",
-	"?==", "x86==`e asm.arch`", "strcmp two strings separated by ==",
+	"?==", "x86 `e asm.arch`", "strcmp two strings separated by ==",
 	"??", " [cmd]", "run cmd if $? != 0",
 	"??", "", "show value of operation",
 	"?_", " hudfile", "load hud menu with given file",
@@ -748,13 +748,12 @@ static int cmd_help(void *data, const char *input) {
 	case '=': // "?=" set num->value
 		if (input[1] == '=') { // ?==
 			char *s = strdup (input + 2);
-			char *e = strstr (s, "==");
+			char *e = strchr (s, ' ');
 			if (e) {
-				*e = 0;
-				e += 2;
+				*e++ = 0;
 				core->num->value = strcmp (s, e);
 			} else {
-				eprintf ("Missing == in expression\n");
+				eprintf ("Missing secondary word in expression to compare\n");
 			}
 			free (s);
 		} else {
--- a/libr/core/cmd_help.c
+++ b/libr/core/cmd_help.c
@@ -127,6 +127,7 @@ static const char *help_msg_question[] = {
 	"?+", " [cmd]", "run cmd if $? > 0",
 	"?-", " [cmd]", "run cmd if $? < 0",
 	"?=", " eip-0x804800", "hex and dec result for this math expr",
+	"?==", "x86==`e asm.arch`", "strcmp two strings separated by ==",
 	"??", " [cmd]", "run cmd if $? != 0",
 	"??", "", "show value of operation",
 	"?_", " hudfile", "load hud menu with given file",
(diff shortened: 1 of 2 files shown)


test/db/esil/x86_32
```diff
@@ -2370,7 +2370,9 @@ CMDS=<<EOF
 # Common parts
 #
 e asm.arch=x86
-e asm.assembler=x86.as
+?==x86==`e asm.arch`
```
ret2libc, line 2373: this looks awful :(

```
?== x86 `e asm.arch`
``` 
Looks better imo. Probably the `==` were just done to make the parsing easy.

trufae: ?== x y implies that no spaces are supported. i can change for that. and i bet that would work with the new parser. any hint how to do that?

trufae: or do you want to handle this when the new command gets accepted, reviewed and tested?

trufae: updated

test/db/esil/x86_32
```diff
@@ -2370,7 +2370,9 @@ CMDS=<<EOF
 # Common parts
 #
 e asm.arch=x86
-e asm.assembler=x86.as
+?==x86 `e asm.arch`
```
ret2libc, line 2373: also, let's assume a space between the command name and the arguments.
```suggestion
?== x86 `e asm.arch`
```
otherwise it becomes harder for the parser to know where the command ends and the arguments begins.

test/db/esil/x86_32
ret2libc: Yeah, I think it looks much better. New shell will be able to deal with spaces as well just by doing
```
?== "first string" "second string"
```

test/db/esil/x86_32
trufae: do you want this space to be mandatory?

ret2libc: I think yes. There should be a clear separation between commands and arguments (the same as you can't do `echoCiao` but you have to do `echo ciao`).

XVilka:
I wonder if we should just make `==` polymorphic in the future in the new parser (able to compare integers, hexadecimals, strings, whatever else).

libr/core/cmd_help.c
```diff
@@ -127,6 +127,7 @@ static const char *help_msg_question[] = {
 	"?+", " [cmd]", "run cmd if $? > 0",
 	"?-", " [cmd]", "run cmd if $? < 0",
 	"?=", " eip-0x804800", "hex and dec result for this math expr",
+	"?==", " x86 `e asm.arch`", "strcmp two strings separated by ==",
```
ret2libc, line 130:
```suggestion
	"?==", " x86 `e asm.arch`", "strcmp two strings",
```

ret2libc (approved):
I commited a small fix to help mmessage. LGTM

trufae:
May i fix the other x86.as tests like this? actually, the tests we have shuold be assembled with x86.nz too. so probably this conditional may work for the rest of them

PR #17617, opened 2020-09-08
Avoid close(-1) in r2pipe

trufae:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/socket/r2pipe.c
+++ b/libr/socket/r2pipe.c
@@ -304,8 +304,6 @@ R_API R2Pipe *r2pipe_open(const char *cmd) {
 			fflush (stderr);
 			// trigger the blocking read
 			write (1, "\xff", 1);
-			// previously set to -1 // 
-			// XXX close (r2p->output[0]);
 			close (r2p->output[1]);
 			close (0);
 			close (1);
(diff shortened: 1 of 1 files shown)


ret2libc (approved):

libr/socket/r2pipe.c
```diff
@@ -304,7 +304,6 @@ R_API R2Pipe *r2pipe_open(const char *cmd) {
 			fflush (stderr);
 			// trigger the blocking read
 			write (1, "\xff", 1);
-			close (r2p->output[0]);
```
XVilka, line 307: It's created only if `#if __UNIX__` is defined in https://github.com/radareorg/radare2/pull/17617/files#diff-29c03378d6a5a5012e9a4b4ba0daaadaR201
What about wrapping this in `#if !__UNIX__` instead?

XVilka (changes requested):

libr/socket/r2pipe.c
ret2libc: Mmm, no r2p->output[0] is set to -1 just few lines above, at line 301, so I think it is ok here.

XVilka (approved):

PR #17618, opened 2020-09-08
Implement arcc command and get_cc_from_regprofile ##types

trufae:

**Known Bugs**

* No way to set default calling convention
* Add (`e anal.cc=reg`) .. overrides the info from types (func.main.cc=amd64)
* Not yet used as fallback. will do a separate pr for this
* Invalid profiles break the regstate and result in runtime assert. Fix will be submited in a separate PR
* using a space before the ( results in no error, but not registers (`tcc eax failure (eax)`). Will submit a fix in another PR

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/anal/cc.c
+++ b/libr/anal/cc.c
@@ -6,38 +6,13 @@
 #define DB anal->sdb_cc
 
 R_API void r_anal_cc_del(RAnal *anal, const char *name) {
-	int i;
+	size_t i;
 	sdb_unset (DB, sdb_fmt ("%s", name), 0);
 	sdb_unset (DB, sdb_fmt ("cc.%s.ret", name), 0);
 	sdb_unset (DB, sdb_fmt ("cc.%s.argn", name), 0);
 	for (i = 0; i < R_ANAL_CC_MAXARG; i++) {
 		sdb_unset (DB, sdb_fmt ("cc.%s.arg%d", name, i), 0);
 	}
-}
-
-R_API char *r_anal_cc_from_regprofile(RAnal *anal) {
-	const char *r0 = r_reg_get_name_by_type (anal->reg, "R0");
-	const char *a0 = r_reg_get_name_by_type (anal->reg, "A0");
-	const char *a1 = r_reg_get_name_by_type (anal->reg, "A1");
-	const char *a2 = r_reg_get_name_by_type (anal->reg, "A2");
-	const char *a3 = r_reg_get_name_by_type (anal->reg, "A3");
-
-	// it is mandatory to have at least =A0 defined in the reg profile
-	// this will be enforced in reg/profile at parsing time
-	r_return_val_if_fail (a0, NULL);
-	if (!r0) {
-		r0 = a0;
-	}
-	if (a3 && a2 && a1) {
-		return r_str_newf ("%s reg(%s, %s, %s, %s)", r0, a0, a1, a2, a3);
-	}
-	if (a2 && a1) {
-		return r_str_newf ("%s reg(%s, %s, %s)", r0, a0, a1, a2);
-	}
-	if (a1) {
-		return r_str_newf ("%s reg(%s, %s)", r0, a0, a1);
-	}
-	return r_str_newf ("%s reg(%s)", r0, a0);
 }
 
 R_API void r_anal_cc_set(RAnal *anal, const char *expr) {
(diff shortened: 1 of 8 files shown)


trufae:
Related to https://github.com/radareorg/radare2/pull/17481

libr/anal/cc.c
```diff
@@ -15,6 +15,39 @@ R_API void r_anal_cc_del(RAnal *anal, const char *name) {
 	}
 }
 
+static const char *getreg(RAnal *anal, const char *alias) {
```
ret2libc, line 18: sounds like a good `r_reg_get_name_by_type`

libr/core/cmd_anal.c
```diff
@@ -654,6 +654,7 @@ static const char *help_msg_ar[] = {
 	"ar?", " <reg>", "Show register value",
 	"arb", " <type>", "Display hexdump of the given arena",
 	"arc", " <name>", "Conditional flag registers",
+	"arcc", "", "Show calling convention defined from the register profile",
```
thestr4ng3r, line 657: I think it should be very explicit about the fact that the cc is just generically derived. So maybe:
```suggestion
	"arcc", "", "Show (fallback) calling convention automatically derived from the register profile",
```

libr/anal/cc.c
```diff
@@ -15,6 +15,39 @@ R_API void r_anal_cc_del(RAnal *anal, const char *name) {
...
+	const char *r0 = getreg (anal, "R0");
+	const char *a0 = getreg (anal, "A0");
+	const char *a1 = getreg (anal, "A1");
+	const char *a2 = getreg (anal, "A2");
+	const char *a3 = getreg (anal, "A3");
+	if (!r0) {
+		r0 = a0;
+	}
+	if (!a0) {
+		return NULL;
```
thestr4ng3r, line 37: Maybe still select some register here? Otherwise we will run into NULL problems again.

trufae: which register? if the regprofile doesn't have A0 it's a problem in the regprofile. i can't magically guess that if the reg profile is incomplete.

libr/anal/cc.c
trufae: You want me to move this into RReg?

trufae: i will take your silence as a yes

ret2libc: sorry, I missed the notification I guess :D but yeah

libr/anal/cc.c
thestr4ng3r: ¯\\_(ツ)_/¯  I guess if you consider a regprofile without A0 invalid it's fine.

trufae: indeed. this can be good pre-check to avoid null cases later

libr/anal/cc.c
```diff
@@ -15,6 +15,30 @@ R_API void r_anal_cc_del(RAnal *anal, const char *name) {
 	}
 }
 
+R_API char *r_anal_cc_from_regprofile(RAnal *anal) {
```
thestr4ng3r, line 18: Why not this:
```suggestion
R_API char *r_anal_cc_from_regprofile(RReg *reg) {
```

trufae:
making failed regprofiles to not break the whole r2is not that easy. ill just drop an assert here and implement this in a separate PR

libr/anal/cc.c
trufae: No. This will break the naming rules of public APIs. this is under RAnal, so it must use such type as 1st argument. if we move this into reg/profile.c then that would be fine to use RReg. but it's not there right now

trufae:
bsd builds are failing. https://builds.sr.ht/~xvilka/job/300915

i moved the api from anal to reg and imho its ready to go. i will start working on better integration of this as well as better error handling in separte PRs

libr/anal/cc.c
trufae: Done

libr/anal/dwarf_process.c
```diff
@@ -57,6 +57,7 @@ static void variable_free(Variable *var) {
 	free (var);
 }
 
+#if 0
```
ret2libc, line 60: remove if not needed

trufae: Well im not touching the dwarf code but this apits out a warning because unused. I can remove but i would prefer to get the feedback from the one who wrote this (in the phone now, no git blame handy)

ret2libc: @HoundThe ok to remove this?

HoundThe: Yes, it is

thestr4ng3r (approved):

PR #17621, opened 2020-09-09
Refactor Labels out of SDB ##anal

thestr4ng3r:
Filling this template is cool and good.

**Detailed description**

What the title says. I wanted to refactor this before implementing labels into r2db.

**Test plan**

run the tests

### DIFF
--- a/libr/anal/labels.c
+++ b/libr/anal/labels.c
@@ -1,98 +1,52 @@
-/* radare - LGPL - Copyright 2014-2018 - pancake */
+/* radare - LGPL - Copyright 2014-2020 - pancake, thestr4ng3r */
 
 #include <r_anal.h>
 
-#define DB anal->sdb_fcns
-
-// list of all the labels for a specific function
-#define LABELS sdb_fmt ("fcn.%"PFMT64x".labels", fcn->addr)
-// value of each element in the labels list
-#define ADDRLABEL(x,y) sdb_fmt ("0x%"PFMT64x"/%s", x,y)
-// resolve by name
-#define LABEL(x) sdb_fmt ("fcn.%"PFMT64x".label.%s", fcn->addr, x)
-// resolve by addr
-#define ADDR(x) sdb_fmt ("fcn.%"PFMT64x".label.0x%"PFMT64x, fcn->addr,x)
-// SDB looks like fcn.0x80480408.labels=0x8048480/patata,0x0405850/potro
-
-R_API ut64 r_anal_fcn_label_get (RAnal *anal, RAnalFunction *fcn, const char *name) {
-	return (anal && fcn)? sdb_num_get (DB, LABEL (name), NULL): UT64_MAX;
+R_API ut64 r_anal_function_get_label(RAnalFunction *fcn, const char *name) {
+	r_return_val_if_fail (fcn, UT64_MAX);
+	ut64 *addr = ht_pp_find (fcn->label_addrs, name, NULL);
+	return addr ? *addr : UT64_MAX;
 }
 
-R_API const char *r_anal_fcn_label_at (RAnal *anal, RAnalFunction *fcn, ut64 addr) {
-	return (anal && fcn)? sdb_const_get (DB, ADDR (addr), NULL): NULL;
+R_API const char *r_anal_function_get_label_at(RAnalFunction *fcn, ut64 addr) {
+	r_return_val_if_fail (fcn, NULL);
+	return ht_up_find (fcn->labels, addr, NULL);
 }
 
-R_API int r_anal_fcn_label_set (RAnal *anal, RAnalFunction *fcn, const char *name, ut64 addr) {
-	if (!anal || !fcn) {
+R_API bool r_anal_function_set_label(RAnalFunction *fcn, const char *name, ut64 addr) {
+	r_return_val_if_fail (fcn && name, false);
+	if (ht_pp_find (fcn->label_addrs, name, NULL)) {
 		return false;
 	}
-	if (sdb_add (DB, ADDR (addr), name, 0)) {
-		if (sdb_num_add (DB, LABEL (name), addr, 0)) {
-			sdb_array_add (DB, LABELS, ADDRLABEL (addr, name), 0);
-			return true;
-		} else {
-			sdb_unset (DB, ADDR (addr), 0);
-		}
-	} else {
-		eprintf ("Cannot add\n");
-	}
-	return false;
-}
-
-R_API int r_anal_fcn_label_del (RAnal *anal, RAnalFunction *fcn, const char *name, ut64 addr) {
-	if (!anal || !fcn || !name) {
+	char *n = strdup (name);
+	if (!ht_up_insert (fcn->labels, addr, n)) {
+		free (n);
 		return false;
 	}
-	sdb_array_remove (DB, LABELS, ADDRLABEL (addr, name), 0);
-	sdb_unset (DB, LABEL (name), 0);
-	sdb_unset (DB, ADDR (addr), 0);
+	ut64 *a = R_NEW (ut64);
+	*a = addr;
+	ht_pp_insert (fcn->label_addrs, name, a);
 	return true;
 }
 
-R_API int r_anal_fcn_labels(RAnal *anal, RAnalFunction *fcn, int rad) {
-	if (!anal) {
-		return 0;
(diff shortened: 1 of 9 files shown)


libr/anal/labels.c
```diff
@@ -1,98 +1,52 @@
...
-		}
-	} else {
-		eprintf ("Cannot add\n");
+	char *n = strdup (name);
+	if (!ht_up_insert (fcn->labels, addr, n)) {
+		free (n);
+		return false;
 	}
-	return false;
+	ut64 *a = R_NEW (ut64);
```
trufae, line 26: you can use ut64_new()

trufae (changes requested):
Looks good to merge for me after those few changes

libr/core/cmd_flag.c
```diff
@@ -682,6 +682,78 @@ static int flag_space_stack_list(RFlag *f, int mode) {
...
+	}
+	PrintFcnLabelsCtx ctx = { rad, pj, fcn };
+	ht_up_foreach (fcn->labels, print_function_labels_cb, &ctx);
+	if (rad == 'j') {
+		pj_end (pj);
+	}
+}
+
+static void print_function_labels(RAnal *anal, RAnalFunction *fcn, int rad) {
+	r_return_if_fail (anal || fcn);
```
trufae, line 725: && not ||

libr/core/cmd_flag.c
```diff
@@ -682,6 +682,78 @@ static int flag_space_stack_list(RFlag *f, int mode) {
...
+	PJ *pj = NULL;
+	if (rad == 'j') {
+		pj = pj_new ();
+	}
+	if (fcn) {
+		print_function_labels_for (fcn, rad, pj);
+	} else {
+		if (rad == 'j') {
```
trufae, line 733: const bool json = rad == 'j';
json && pj_o(o);
...

trufae (changes requested):
This small change will make the code shorter and the comparisons may be better with const

libr/anal/labels.c
```diff
@@ -1,98 +1,52 @@
...
-// resolve by addr
-#define ADDR(x) sdb_fmt ("fcn.%"PFMT64x".label.0x%"PFMT64x, fcn->addr,x)
-// SDB looks like fcn.0x80480408.labels=0x8048480/patata,0x0405850/potro
-
-R_API ut64 r_anal_fcn_label_get (RAnal *anal, RAnalFunction *fcn, const char *name) {
-	return (anal && fcn)? sdb_num_get (DB, LABEL (name), NULL): UT64_MAX;
+R_API ut64 r_anal_function_get_label(RAnalFunction *fcn, const char *name) {
+	r_return_val_if_fail (fcn, UT64_MAX);
+	ut64 *addr = ht_pp_find (fcn->label_addrs, name, NULL);
+	return addr ? *addr : UT64_MAX;
```
ret2libc, line 10:
```suggestion
	return addr? *addr: UT64_MAX;
```

libr/anal/labels.c
```diff
@@ -1,98 +1,52 @@
...
-		}
-	} else {
-		eprintf ("Cannot add\n");
+	char *n = strdup (name);
+	if (!ht_up_insert (fcn->labels, addr, n)) {
+		free (n);
+		return false;
 	}
-	return false;
+	ht_pp_insert (fcn->label_addrs, name, ut64_new (addr));
```
ret2libc, line 28: Isn't there a check here because you assume the insert in `label_addrs` cannot fail if the insert in `labels` didn't?

libr/parse/filter.c
```diff
@@ -222,7 +222,7 @@ static bool filter(RParse *p, ut64 addr, RFlag *f, RAnalHint *hint, char *data,
 				if (flag && !strncmp (flag->name, "section.", 8)) {
 					flag = r_flag_get_i (f, off);
 				}
-				const char *label = p->label_get (p->analb.anal, fcn, off);
+				const char *label = fcn ? p->label_get (fcn, off) : NULL;
```
ret2libc, line 225:
```suggestion
				const char *label = fcn? p->label_get (fcn, off): NULL;
```

test/unit/test_anal_function.c
```diff
@@ -70,9 +70,63 @@ bool test_r_anal_function_relocate() {
 	mu_end;
 }
 
+bool test_r_anal_function_labels() {
+	RAnal *anal = r_anal_new ();
+
+	RAnalFunction *f = r_anal_create_function (anal, "do_something", 0x1337, 0, NULL);
+
+	bool s = r_anal_function_set_label (f, "smartfriend", 0x1339);
+	mu_assert ("set label", s);
```
ret2libc, line 79: may i suggest to use `mu_assert_true`? (note `mu_assert_true` has the actual value as first arg, like all the other `mu_assert_*` macros.

libr/include/r_anal.h
```diff
@@ -1955,13 +1957,12 @@ R_API RAnalCaseOp* r_anal_switch_op_add_case(RAnalSwitchOp * swop, ut64 addr, ut
 R_API RAnalCycleFrame* r_anal_cycle_frame_new (void);
 R_API void r_anal_cycle_frame_free (RAnalCycleFrame *cf);
 
-
 /* labels */
```
ret2libc, line 1960: stupid question, but... What are exactly function labels? Would it make sense to have a small description here?

trufae: they are like flags but local to the function, like the labels starting with dollars in the gnu assembler. this was a request from xvilka many years ago. he can provide more examples and details

libr/parse/filter.c
trufae: how do u do these commit suggestions?

libr/include/r_anal.h
thestr4ng3r: beach is a label

libr/parse/filter.c
thestr4ng3r: 20200914_19h58m07s_grim
Unfortunately only works on single lines

libr/core/cmd_flag.c
thestr4ng3r: No, || is correct. It either expects `fcn` in which case it prints only this function and ignores `anal` or `anal` in which case it prints all functions and ignores `fcn`.

libr/anal/labels.c
thestr4ng3r: The check is here:
https://github.com/radareorg/radare2/pull/17621/files#diff-635a23ca710c8cf67d67f28d75cdcf86R20
```c
if (ht_pp_find (fcn->label_addrs, name, NULL)) {
	return false;
}
```

libr/core/cmd_flag.c
trufae: thats quite weird, and fcn have a pointer to anal. but its an internal api, so its probably not tied to clear usage arguments.

trufae: also, anal is never used in this function

libr/core/cmd_flag.c
```diff
@@ -682,6 +682,80 @@ static int flag_space_stack_list(RFlag *f, int mode) {
...
+				pj_k (pj, f->name);
+			}
+			print_function_labels_for (f, rad, pj);
+		}
+		if (json) {
+			pj_end (pj);
+		}
+	}
+	if (json) {
+		r_cons_printf ("%s\n", pj_string (pj));
```
trufae, line 754: use r_cons_println

libr/core/cmd_flag.c
thestr4ng3r: Yep, the main reason to do this and not multiple functions is to avoid code duplication.

trufae (changes requested):

libr/core/cmd_flag.c
thestr4ng3r: anal is used here: https://github.com/radareorg/radare2/pull/17621/files#diff-1c86fab8e892153e8e316d8fc8ce810eR740

trufae: yes i understand the code dup reason, but imho it would be better to specify what do in an argument or function name rather than doing black magic depending on which argument is null

ret2libc (approved):

trufae (approved):

PR #17624, opened 2020-09-10
Implement RCons.setTitle() for Windows and add ?et command ##cons

trufae:
DO NOT SQUASH

**Detailed description**

there's a w32api for that

**Test plan**

requires humans

**Closing issues**

none registered, this api was not used anywhere

### DIFF
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -1678,7 +1678,15 @@ R_API void r_cons_set_last_interactive(void) {
 
 R_API void r_cons_set_title(const char *str) {
 #if __WINDOWS__
+#  if defined(_UNICODE)
+	wchar_t* wstr = r_utf8_to_utf16_l (str, strlen (str));
+	if (wstr) {
+		SetConsoleTitleW (wstr);
+		R_FREE (wstr);
+	}
+#  else // defined(_UNICODE)
 	SetConsoleTitle (str);
+#  endif // defined(_UNICODE)
 #else
 	r_cons_printf ("\x1b]0;%s\007", str);
 #endif
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -1010,7 +1010,7 @@ R_API void r_cons_print_fps (int col) {
 		if (diff < 0) {
 			fps = 0;
 		} else {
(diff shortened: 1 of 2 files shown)


libr/core/cmd_help.c
```diff
@@ -949,6 +949,9 @@ static int cmd_help(void *data, const char *input) {
 		break;
 	case 'e': // "?e" echo
 		switch (input[1]) {
+		case 't': // "?e=t newtitle"
```
ret2libc, line 952:
```suggestion
		case 't': // "?et newtitle"
```

ret2libc (approved):

wargio (approved):

libr/cons/cons.c
```diff
@@ -1677,7 +1677,11 @@ R_API void r_cons_set_last_interactive(void) {
 }
 
 R_API void r_cons_set_title(const char *str) {
+#if __WINDOWS__
+	SetConsoleTitle (str);
```
a1ext, line 1681: You have to use "wide" version of this API and you need to convert the input string. Actually it displays garbage (treats str as `wchar_t*`:
image

a1ext (changes requested):

libr/cons/cons.c
a1ext: I suggest it should look like:
```
#if __WINDOWS__
#  if defined(_UNICODE)
wchar_t* wstr = r_utf8_to_utf16_l(str, strlen(str));
if (wstr) {
    SetConsoleTitleW (wstr);
    R_FREE(wstr);
}
#  else // defined(_UNICODE)
SetConsoleTitle (str);
# endif
#endif // __WINDOWS__
```

trufae: applied your patch

a1ext: I have compiled your recent commit with the following meson command to turn on the unicode support:
`python sys\meson.py --shared --install --options "c_args=-D_UNICODE -DUNICODE"`

and it works well with utf8 console charset in default Windows terminal:
image

a1ext (approved):
👍

XVilka (approved):

PR #17627, opened 2020-09-10
Increase timeout for macos-clang-tests job

ret2libc:


ret2libc:
I have force-pushed this because it just increments the timeout of the macos job and the only way to test if it works is on master. On PRs everything runs smoothly so far, it's just on master that probably jobs go a bit slower due to more jobs running.

PR #17631, opened 2020-09-10
Graph refactor

karliss:

**Detailed description**

This continuation of what was started #17362 in.

* Support all graph output formats for `acg`
* Convert `agx` and `agi` to use the same mechanism

There are a bunch more more or less manual graph printing that could be refactored this way, but I wanted to to keep first PR smaller and discuss the concept of changes before refactoring everything.

Goal of changes (within this PR and follow up PR) are:
* reduce usage of global state for implementing commands (core->graph)
* have a C api for obtaining drawable graph structure that is somewhere between exact drawing and Ranal structures.
* Include information object's addres when it makes sense so that it can be used for interactive functionality. In many of the r2 graphs each node represents something that have address.
* Use agraph only when drawing a graph in terminal.

@ret2libc Had a strong opinion related to where exactly this code should be located. This might be a good chance to discuss that further.

**Test plan**

* r2r tests work
* Interactive graph modes agiv, agxv, acgv work

**Closing issues**

...


libr/core/cmd_anal.c
```diff
@@ -8058,35 +8058,71 @@ static char *getViewerPath(void) {
 	return NULL;
 }
 
-static char* graph_cmd(RCore *core, char *r2_cmd, const char *save_path) {
+static char *dot_executable_path() {
```
ret2libc, line 8061: tests say this should be `...(void)`

libr/util/graph.c
```diff
@@ -162,6 +162,16 @@ R_API RGraphNode *r_graph_add_node(RGraph *t, void *data) {
 	return n;
 }
 
+R_API RGraphNode *r_graph_add_nodef(RGraph *graph, void *data, RListFree user_free) {
+	RGraphNode *node = r_graph_add_node (graph, data);
+	if (node) {
+		node->free = user_free;
+	} else if (user_free) {
+		user_free (data);
```
ret2libc, line 170: I'm not sure about this. If the allocation goes wrong, `data` should still be owned by the caller, no?
The caller would do something like:
```C
n = r_graph_add_nodef (g, data, ufree);
if (!n) {
   ufree (data);
   // free other stuff defined in this function
}
```

libr/util/graph.c
```diff
@@ -316,6 +328,23 @@ R_API RGraphNodeInfo *r_graph_create_node_info(char *title, char *body, ut64 off
 		data->title = title;
 		data->body = body;
 		data->offset = offset;
+	} else {
+		free (title);
```
ret2libc, line 332: same here. Do we really want these free here? Maybe I'm wrong, but I think the "ownership" should be transferred only when everything goes right. If something goes wrong the objects still belong to the user, who can still do whatever it wants with those data.

libr/util/graph.c
```diff
@@ -316,6 +328,23 @@ R_API RGraphNodeInfo *r_graph_create_node_info(char *title, char *body, ut64 off
...
 	}
 	return data;
-}
\ No newline at end of file
+}
+
+R_API RGraphNode *r_graph_add_node_info(RGraph *graph, char *title, char *body, ut64 offset) {
+	RGraphNodeInfo *data;
+	if (!graph) {
+		free (title);
```
ret2libc, line 341: same ownership thing.

libr/core/agraph.c
```diff
@@ -4939,6 +4939,7 @@ R_API RAGraph *create_agraph_from_graph(const RGraph/*<RGraphNodeInfo>*/ *graph)
 	if (!result_agraph) {
 		return NULL;
 	}
+	result_agraph->need_reload_nodes = false;
```
ret2libc, line 4942: Why was this required?

libr/include/r_core.h
```diff
@@ -578,16 +578,16 @@ R_API void r_core_anal_esil(RCore *core, const char *str, const char *addr);
...
 R_API int r_core_anal_search(RCore *core, ut64 from, ut64 to, ut64 ref, int mode);
 R_API int r_core_anal_search_xrefs(RCore *core, ut64 from, ut64 to, int rad);
-R_API int r_core_anal_data (RCore *core, ut64 addr, int count, int depth, int wordsize);
+R_API int r_core_anal_data(RCore *core, ut64 addr, int count, int depth, int wordsize);
 R_API void r_core_anal_datarefs(RCore *core, ut64 addr);
 R_API void r_core_anal_coderefs(RCore *core, ut64 addr);
-R_API void r_core_anal_codexrefs(RCore *core, ut64 addr);
-R_API void r_core_anal_importxrefs(RCore *core);
+R_API RGraph/*<RGraphNodeInfo>*/ *r_core_anal_codexrefs(RCore *core, ut64 addr);
+R_API RGraph /*<RGraphNodeInfo>*/ *r_core_anal_importxrefs(RCore *core);
```
ret2libc, line 590:
```suggestion
R_API RGraph/*<RGraphNodeInfo>*/ *r_core_anal_importxrefs(RCore *core);
```

ret2libc (commented):
Some initial comments, I still need to go through the whole thing.

libr/core/cmd_anal.c
```diff
@@ -8286,25 +8323,248 @@ R_API void r_core_agraph_print (RCore *core, int use_utf, const char *input) {
...
+		case 'k': // "ag_k"
+		{
+			Sdb *db = r_agraph_get_sdb (agraph);
+			char *o = sdb_querys (db, "null", 0, "*");
+			r_cons_print (o);
+			free (o);
+			break;
+		}
+		case 'v': // "ag_v"
+		case 'i': // "ag_i" - open current core->graph in interactive mode
```
ret2libc, line 8489: Is this missing from the help? I see `v Interactive ascii art` but I can't find `i`.

libr/core/cmd_anal.c
```diff
@@ -8286,25 +8323,248 @@ R_API void r_core_agraph_print (RCore *core, int use_utf, const char *input) {
...
+		{
+			RANode *ran = r_agraph_get_first_node (agraph);
+			if (ran) {
+				ut64 oseek = core->offset;
+				r_agraph_set_title (agraph, r_config_get (core->config, "graph.title"));
+				r_agraph_set_curnode (agraph, ran);
+				agraph->force_update_seek = true;
+				agraph->need_set_layout = true;
+				agraph->layout = r_config_get_i (core->config, "graph.layout");
+				int ov = r_cons_is_interactive ();
```
ret2libc, line 8499:
```suggestion
				bool ov = r_cons_is_interactive ();
```

libr/core/cmd_anal.c
```diff
@@ -8623,25 +8871,21 @@ static void cmd_anal_graph(RCore *core, const char *input) {
...
                                 char *cmdargs = r_str_newf ("agdd 0x%"PFMT64x, core->offset);
-                                char *cmd = graph_cmd (core, cmdargs, input + 2);
-                                if (cmd && *cmd) {
-                                        r_core_cmd0(core, cmd);
-                                }
-                                free(cmd);
-                                free(cmdargs);
-                                break;
+                                convert_dotcmd_to_image (core, cmdargs, input + 2);
+				free (cmdargs);
```
ret2libc, line 8887: tabs

libr/core/cmd_anal.c
```diff
@@ -8058,35 +8058,71 @@ static char *getViewerPath(void) {
...
-	return cmd;
+	r_core_cmd0 (core, cmd);
+	result = true;
+end:
+	free (cmd);
+	free (dot);
+	return result;
+}
+
+static bool convert_dotcmd_to_image(RCore *core, char *r2_cmd, const char *save_path) {
```
ret2libc, line 8107: r2_cmd is never used

test/db/cmd/cmd_ag
```diff
@@ -0,0 +1,337 @@
+NAME=agi
+FILE=bins/elf/hello_world
+CMDS=<<EOF
+aa 2> /dev/null
+agi*
+EOF
+EXPECT=<<EOF
+agn "sym.imp.free" base64:
```
ret2libc, line 8: master currently doesn't print the empty string. Which, according to the help of `agn` seems wrong because a body is expected. However currently `agn <title> base64:` fails.

libr/core/agraph.c
karliss: TLDR: `need_reload_nodes` causes agraph to attempt loading either callgraph or basic block graph, which doesn't make sense in this case. 

Because agraph is mixing multiple responsibilities within single module. On hand it deals with task of making exact drawing on other it also handles reloading of data from r2 for some kinds of graphs like function CFG. Many of agraph properties are specific to one kind of graph.

test/db/cmd/cmd_ag
```diff
@@ -0,0 +1,337 @@
...
+.--------------------.
+|  main              |
+`--------------------'
+digraph code {
+rankdir=LR;
+outputorder=edgesfirst
+graph [bgcolor=azure];
+edge [arrowhead=normal, color="#3030c0" style=bold weight=2];
+node [fillcolor=white, style=filled shape=box fontname="Courier" fontsize="8"];
+0 [URL="sym.imp.free", color="lightgray", label="sym.imp.free"]
```
ret2libc, line 79: master here uses "sym.imp.free" instead of 0. Is it ok?

ret2libc (commented):
So, if I get it right, the final purpose is to also move all those other things that currently work on `core->graph` directly in `r_core_agraph_print` to `r_core_graph_print` and make somehow core->graph be a graph of RGraphNodeInfo, right?

Would it be already possible to rewrite `r_core_agraph_print` in terms of `r_core_graph_print` or do you prefer to do it in a separate PR?

Anyway, the general idea looks ok to me. However, as said previously, I think `RGraphNodeInfo` doesn't really belong to `r_util/r_graph.h`. I'd prefer to have a separate header for it `r_graph_drawable` or something, just for clear separation.

libr/util/graph.c
karliss: Yes I was having some doubts on which of the two strategies to choose. Here are the reasons I chose this strategy:
* "This function argument takes ownership, ALWAYS!" seemed like a simpler rule than "It takes ownership but sometimes it doesn't"
* Less error handling code and cleaner code in function using the API to build a graph.
* Handling of the case when construction consists of two steps and first step succeeds transfering the ownership, but second step fails. Here is an example to make the problem clearer:
```c
foo_bar *make_foobar(char *a)
{
    foo * foo_obj = make_foo(a);
    if (!foo_obj) {
       return NULL;
   }
   foo_x(foo_obj);
   foo_bar * foo_bar_obj = malloc(...);
   if (!foo_bar) {
       // What to do now? Ownership of a was already transfered to foo_obj. Can't call the normal destructor of foo, because that would also free a. What if foo_x or even make_foo can modify or replace `a`?
   }
   return foo_bar_obj;
}
```

It seems that neither of approaches is widely used and in most case the approach you suggested means that no one in case of erorr memory is leaked because caller also doesn't release it.  If you consider that my approach is likely to cause confusion I would prefer the third option of not doing the ownership transfer at all and doing a copy when necessary. That way at least there is less conditional code on most levels.

libr/core/agraph.c
ret2libc: ok! ;)

karliss:
> So, if I get it right, the final purpose is to also move all those ...

Most cases except the cases maybe some cases that do custom stuff in a way that provides additional value. Replacing core->graph with RGraph<RGraphNodeInfo>+dictionary<string, node> would be one of the last steps. Yes I would preffer doing r_core_agraph_print rewrite separately. I hope that doing things in multiple smaller PRs will make both review process and further refactoring easier once there is less code directly depending core->graph. I expect there to be a bit of extra work to untangle the parts important for `agg` from `afg` specific code.

An approximate plan is:
*) Initial PR (this one), introduce the concept, demonstrate it in one or two cases. During review debate various design choices. Don't try to convert everything so that in case review requires rewriting some parts only the 1-3 examples need to be redone.
*) Convert most of the cases that don't need special handling. Convert the cases that currently have special handling but don't need it.
*) agg refactoring, reduce code duplication between mechanism I introduced and agg. See what can be cleaned up within agraph.

I will move the relevant code from r_graph to r_graph_drawable within this PR.

test/db/cmd/cmd_ag
karliss: I hav

karliss: No problems removing it. The behavior in current master is a bit inconsistent. If you do `agi*; .agi*; agg*` first one prints without `base64:` but last one prints it.

libr/util/graph.c
ret2libc: Probably both approaches have their own pros/cons, but so far I think we are used to the approach "take ownership only if there's no error". At least, this is what I think is used in r_list and other data structures. @thestr4ng3r @trufae opinions?

libr/core/cmd_anal.c
karliss: What's exactly you suggestion? Keep spaces for this block or reindent it whole?

ret2libc: Oh damn... it's all messed up... I thought this was "the wrong one", instead this is the right one and all others are wrong :D I would say to just make it wrong as well :PP

test/db/cmd/cmd_ag
karliss: Why not? Its valid dot file. Less copies of the the same string, less place in code that needs to properly print those strings. Allow this function to be used with graphs that don't have unique titles.

libr/core/cmd_anal.c
karliss: Fixed

libr/core/canal.c
```diff
@@ -2280,43 +2280,67 @@ R_API void r_core_anal_coderefs(RCore *core, ut64 addr) {
...
+	RGraphNode *curr_node;
+	char *me;
+	if (!graph) {
+		return;
+	}
+	me = (f && f->offset == addr)
+		? r_str_new (f->name)
+		: r_str_newf ("0x%" PFMT64x, addr);
+
+	curr_node = r_graph_add_node_info (graph, me, NULL, addr);
```
trufae, line 2294: define and assign in the same line

libr/core/canal.c
```diff
@@ -2280,43 +2280,67 @@ R_API void r_core_anal_coderefs(RCore *core, ut64 addr) {
 	}
 }
 
-R_API void r_core_anal_importxrefs(RCore *core) {
+static void add_single_addr_xrefs(RCore *core, ut64 addr, RGraph *graph) {
+	RFlagItem *f = r_flag_get_at (core->flags, addr, false);
+	RGraphNode *curr_node;
+	char *me;
+	if (!graph) {
```
trufae, line 2287: use r_return_if_fail

libr/core/canal.c
```diff
@@ -2280,43 +2280,67 @@ R_API void r_core_anal_coderefs(RCore *core, ut64 addr) {
...
+	bool lit = info? info->has_lit: false;
 	bool va = core->io->va || core->bin->is_debugger;
 
 	RListIter *iter;
 	RBinImport *imp;
 	if (!obj) {
-		return;
+		return NULL;
+	}
+	graph = r_graph_new ();
```
trufae, line 2322: define and assign in the same line

libr/util/graph.c
```diff
@@ -316,6 +328,23 @@ R_API RGraphNodeInfo *r_graph_create_node_info(char *title, char *body, ut64 off
 		data->title = title;
 		data->body = body;
```
trufae, line 329: i think this is a setter and it should be responsible for owning the passed string, so better use const char* for body and title. just strdup in here and remove the frees below

libr/util/graph.c
```diff
@@ -305,6 +314,9 @@ R_API void r_graph_dfs(RGraph *g, RGraphVisitor *vis) {
 
 R_API void r_graph_free_node_info(void *ptr) {
 	RGraphNodeInfo *info = ptr;
+	if (!info) {
```
trufae, line 317: if (!ptr) one line above?

trufae (changes requested):

libr/util/graph.c
ret2libc: BTW, see also https://github.com/radareorg/radare2/pull/17621/files#diff-635a23ca710c8cf67d67f28d75cdcf86R24 . I implemented `ht_`, so I'm not sure how much that qualifies as a valid example, but even there, if `n` is not correctly inserted in the ht, it is still freed by the caller.

libr/core/cmd_anal.c
```diff
@@ -8521,7 +8708,7 @@ static void cmd_anal_graph(RCore *core, const char *input) {
 		case 'w':
 		case ' ': {
 			core->graph->is_callgraph = true;
-			r_core_cmdf (core, "ag-; .agc* @ %"PFMT64u"; agg%s;", core->offset, input + 1);
+			r_core_cmdf (core, "ag-; .agc* @ %" PFMT64u "; agg%s;", core->offset, input + 1);
```
trufae, line 8711: Code injection

libr/core/canal.c
```diff
@@ -2280,43 +2281,65 @@ R_API void r_core_anal_coderefs(RCore *core, ut64 addr) {
 	}
 }
 
-R_API void r_core_anal_importxrefs(RCore *core) {
+static void add_single_addr_xrefs(RCore *core, ut64 addr, RGraph *graph) {
+	RFlagItem *f = r_flag_get_at (core->flags, addr, false);
+	char *me;
+	r_return_if_fail (graph);
+	me = (f && f->offset == addr)
```
trufae, line 2288: Define and assign in the same line

libr/core/cmd_anal.c
```diff
@@ -8286,25 +8324,186 @@ R_API void r_core_agraph_print (RCore *core, int use_utf, const char *input) {
...
+		}
+		break;
+	}
+	case '*': // "ag_*" -
+		print_graph_agg (graph);
+		break;
+	case 'J':
+	case 'j': {
+		PJ *pj = pj_new ();
+		if (!pj) {
```
trufae, line 8469: I would invert the check and remove the return

libr/util/graph_drawable.c
```diff
@@ -0,0 +1,100 @@
...
+R_API RGraphNode *r_graph_add_node_info(RGraph *graph, const char *title, const char *body, ut64 offset) {
+	r_return_val_if_fail (graph, NULL);
+	RGraphNodeInfo *data = r_graph_create_node_info (title, body, offset);
+	if (!data) {
+		return NULL;
+	}
+	return r_graph_add_nodef (graph, data, r_graph_free_node_info);
+}
+
+char *r_graph_drawable_to_dot (RGraph /*RGraphNodeInfo*/ *graph, const char *node_properties, const char *edge_properties) {
```
trufae, line 33: If its public use R_API

Also im not 100% happy eith the drawable name but im fine for all the changes proposed here

trufae (changes requested):

libr/core/cmd_anal.c
karliss: *) was there before this issue
*) might get removed in the follow up issues which convert more graphs to the mechanism introduce in this issue

libr/util/graph_drawable.c
karliss: Do you have suggestions for better name?

XVilka (approved):
Looks very good to me!

libr/core/cmd_anal.c
```diff
@@ -8258,14 +8295,15 @@ R_API void r_core_agraph_print (RCore *core, int use_utf, const char *input) {
 	case 'd': { // "aggd" - dot format
 		const char *font = r_config_get (core->config, "graph.font");
 		r_cons_printf ("digraph code {\nrankdir=LR;\noutputorder=edgesfirst\ngraph [bgcolor=azure];\n"
-			"edge [arrowhead=normal, color=\"#3030c0\" style=bold weight=2];\n"
-			"node [fillcolor=white, style=filled shape=box "
-			"fontname=\"%s\" fontsize=\"8\"];\n", font);
+			       "edge [arrowhead=normal, color=\"#3030c0\" style=bold weight=2];\n"
```
ret2libc, line 8298: no need to change these lines, they were right as they were before AFAIK.

libr/util/graph.c
ret2libc: You said this reply was not required any more. So, is this going away? Are changes still missing from this PR?

libr/include/r_util/r_graph_drawable.h
```diff
@@ -0,0 +1,53 @@
...
+	 * @brief Optional offset for the object corresponding to node.
+	 * 
+	 * Interactive output modes can use it to provide actions like seeking to
+	 * this position or modify the object.
+	 */
+	ut64 offset;
+} RGraphNodeInfo;
+
+R_API void r_graph_free_node_info(void *ptr);
+R_API RGraphNodeInfo *r_graph_create_node_info(const char *title, const char *body, ut64 offset);
```
ret2libc, line 30: I guess we can start by hiding this from the API as it's not used by any module outside the r_graph_add_node_info function.

libr/util/graph.c
karliss: Missed that one due to it being separate file. Removed.

libr/util/graph_drawable.c
karliss: Adding missing R_API to function definitions.

libr/core/cmd_anal.c
karliss: Clang format decided to change that, reverted.

XVilka:
@ret2libc please take a look again and lets merge this.

libr/core/cmd_anal.c
```diff
@@ -8286,25 +8324,186 @@ R_API void r_core_agraph_print (RCore *core, int use_utf, const char *input) {
...
+
+static void print_graph_agg(RGraph /*RGraphNodeInfo*/ *graph) {
+	RGraphNodeInfo *print_node;
+	RGraphNode *node, *target;
+	RListIter *it, *edge_it;
+	r_list_foreach (graph->nodes, it, node) {
+		char *encbody;
+		int len;
+		print_node = node->data;
+		if (print_node->body && print_node->body[0]) {
```
ret2libc, line 8357:
```suggestion
		if (R_STR_ISNOTEMPTY (print_node->body)) {
```

libr/util/graph_drawable.c
```diff
@@ -0,0 +1,104 @@
...
+		return NULL;
+	}
+	RGraphNode *node = r_graph_add_nodef (graph, data, r_graph_free_node_info);
+	if (!node) {
+		r_graph_free_node_info (data);
+	}
+	return node;
+}
+
+R_API char *r_graph_drawable_to_dot (RGraph /*RGraphNodeInfo*/ *graph, const char *node_properties, const char *edge_properties) {
```
ret2libc, line 37:
```suggestion
R_API char *r_graph_drawable_to_dot(RGraph /*RGraphNodeInfo*/ *graph, const char *node_properties, const char *edge_properties) {
```

libr/util/graph_drawable.c
```diff
@@ -0,0 +1,104 @@
...
+		}
+		r_list_foreach (node->out_nodes, itt, target) {
+			r_strbuf_appendf (&buf, "%d -> %d\n", node->idx, target->idx);
+		}
+	}
+	r_strbuf_append (&buf, "}\n");
+	return r_strbuf_drain_nofree (&buf);
+}
+
+R_API void r_graph_drawable_to_json (RGraph /*RGraphNodeInfo*/ *graph, PJ *pj, bool use_offset) {
```
ret2libc, line 69:
```suggestion
R_API void r_graph_drawable_to_json(RGraph /*RGraphNodeInfo*/ *graph, PJ *pj, bool use_offset) {
```

ret2libc:
Please address opened comments by pancake and me, but overall I'm ok with the PR. :+1:

libr/core/canal.c
karliss: I am fine with whatever style you choose, but shouldn't code style guide be updated in that case. https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md requests not to use C99 variable declarations.

XVilka (approved):

ret2libc (approved):

XVilka:
One unit test is still failing: https://github.com/radareorg/radare2/pull/17631/checks?check_run_id=1205972042#step:13:825

libr/anal/class.c
```diff
@@ -2,6 +2,7 @@
 
 #include <r_anal.h>
 #include <r_vector.h>
+#include <r_util/r_graph_drawable.h>
```
trufae, line 5: this file is not in libr/meson.build. this is why ci is red.

trufae (changes requested):

PR #17635, opened 2020-09-12
r2pm: Add new command that installs package without updating git

zamaudio:

Previously, r2pm seems to switch branches to master and git pull, causing havoc for CI trying to test changes.
With this change in radare2, the following pull request in radare2-extras enables a working GitHub CI:
https://github.com/radareorg/radare2-extras/pull/265
(However the WIP change needs to be dropped so it does not use my own radare2 repo for each clone, once the r2 PR is merged).

The proof can be seen here:
https://github.com/zamaudio/radare2-extras/runs/1106054548

### DIFF
--- a/binr/r2pm/r2pm
+++ b/binr/r2pm/r2pm
@@ -327,16 +327,20 @@ R2PM_GIT() {
 	cd "${R2PM_GITDIR}"
 	if [ -d "${DIR}" ]; then
 		cd "${DIR}"
-		git checkout master
-		git pull
+		if [ "$ACTION" = Install ]; then
+			git checkout master
+			git pull
+		fi
 	else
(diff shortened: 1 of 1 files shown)


trufae:
Why

trufae:
This is breaking version compatibility tags support

zamaudio:
OK, I will try force-pushing a new patch that adds a new flag `r2pm -ti <pkgname>` for test-install, and not to interfere with install target.

zamaudio:
Latest r2 PR works here: https://github.com/zamaudio/radare2-extras/runs/1107386303

trufae:
If its just for testing i think its better to use an env var or just change the code for testing

PR #17639, opened 2020-09-13
Set RPATH/RUNPATH on r2 libs too if local

kazarmy:

**Detailed description**

The linker of Ubuntu 20.04 (WSL2) by default sets RUNPATH instead of RPATH if requested, but its dynamic linker treats RUNPATH and RPATH differently resulting in the following error when running r2 with a meson local installation:

```
r2: error while loading shared libraries: libr_config.so.4.6.0-git: cannot open shared object file: No such file or directory
```

This pr fixes that by setting RPATH/RUNPATH on r2 libs too.

**Test plan**

Under Ubuntu 20.04 (WSL2) in r2's git root dir:

```sh
rm -rf build
sys/meson.py --release --shared --local --prefix ~/bin/radare2-git --install
~/bin/radare2-git/bin/r2 -
```

The above error should not occur.

This pr was only tested under Ubuntu 20.04 (WSL2), but should be harmless for distribution linkers that set RPATH by default (e.g. Fedora).

**Closing issues**

...

----

The renaming of `rpath` to `rpath_exe` was done using:

```sh
REGEX="install_rpath: rpath" bash -c 'git grep -l "$REGEX" | xargs -n 1 sed -i "s/$REGEX/&_exe/"'
```

The setting of `r2_lib` was done using:

```sh
REGEX="^\([[:blank:]]*\)\(soversion: r2_libversion\)" bash -c 'git grep -l "$REGEX" | xargs -n 1 sed -i "s/$REGEX/\1install_rpath: rpath_lib,\n\1\2/"'
```

### DIFF
--- a/binr/blob/meson.build
+++ b/binr/blob/meson.build
@@ -5,7 +5,7 @@ executable('radare2', 'main.c',
     r_util_dep,
   ],
   install: true,
-  install_rpath: rpath,
+  install_rpath: rpath_exe,
   implicit_include_directories: false
 )
 
--- a/binr/r2agent/meson.build
+++ b/binr/r2agent/meson.build
@@ -7,6 +7,6 @@ executable('r2agent', 'r2agent.c',
     r_cons_dep
   ],
   install: true,
-  install_rpath: rpath,
+  install_rpath: rpath_exe,
   implicit_include_directories: false
 )
--- a/binr/r2r/meson.build
+++ b/binr/r2r/meson.build
@@ -7,7 +7,7 @@ if get_option('enable_r2r')
       lrt,
     ],
     install: true,
-    install_rpath: rpath,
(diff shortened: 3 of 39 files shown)


karliss:
Wouldn't a more proper fix be setting rpath for all the r2 libraries? Currently it is only set for the executables. That way it should also work with RUNPATH.

kazarmy:
Hmm, now I realize that libraries can load other libraries so that certainly sounds reasonable

trufae (commented):
lgtm, this was done already for make

trufae (approved):

karliss (approved):

trufae:
@ret2libc can you review this one?

ret2libc (approved):
Nice!

PR #17640, opened 2020-09-13
Use sdb_set instead of sdb_querys.

karliss:

**Detailed description**

Use sdb_set instead of sdb_querys. sdb_querys takes a lot of time doing some json stuff, there is also risk of command injection.

This was discovered when profiling r2. After the changes time for doing aaa on Cutter debug executable is reduced from ~840s to 
125s.

flamegraph_before_after

Doing the changes also improved the argument analysis for two rust tests. I am guessing that this was due to function name getting interpreted as part of query.

**Test plan**

Covered by existing r2r testsuite. Having the 2 changed lines commented out or using wrong sdb command breaks ~ 20 tests.

**Closing issues**

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -855,16 +855,20 @@ static void __add_vars_sdb(RCore *core, RAnalFunction *fcn) {
 
 	r_list_foreach (all_vars, iter, var) {
 		if (var->isarg) {
-			char *query = r_str_newf ("anal/types/func.%s.arg.%d=%s,%s", fcn->name, arg_count, var->type, var->name);
-			sdb_querys (core->sdb, NULL, 0, query);
-			free (query);
+			char *key = r_str_newf ("func.%s.arg.%d", fcn->name, arg_count);
+			char *value = r_str_newf ("%s,%s", var->type, var->name);
+			sdb_set (core->anal->sdb_types, key, value, 0);
+			free (key);
+			free (value);
 			arg_count++;
 		}
 	}
 	if (arg_count > 0) {
-		char *query = r_str_newf ("anal/types/func.%s.args=%d", fcn->name, arg_count);
-		sdb_querys (core->sdb, NULL, 0, query);
-		free (query);
+		char *key = r_str_newf ("func.%s.args", fcn->name);
+		char *value = r_str_newf ("%d", arg_count);
+		sdb_set (core->anal->sdb_types, key, value, 0);
+		free (key);
+		free (value);
 	}
 	r_anal_fcn_vars_cache_fini (&cache);
 }
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -853,23 +853,30 @@ static void __add_vars_sdb(RCore *core, RAnalFunction *fcn) {
 	r_list_join (all_vars, cache.bvars);
 	r_list_join (all_vars, cache.svars);
 
+	RStrBuf key, value;
+	r_strbuf_init (&key);
+	r_strbuf_init (&value);
+
 	r_list_foreach (all_vars, iter, var) {
 		if (var->isarg) {
-			char *key = r_str_newf ("func.%s.arg.%d", fcn->name, arg_count);
-			char *value = r_str_newf ("%s,%s", var->type, var->name);
-			sdb_set (core->anal->sdb_types, key, value, 0);
-			free (key);
-			free (value);
+			if (!r_strbuf_setf (&key, "func.%s.arg.%d", fcn->name, arg_count) ||
+				!r_strbuf_setf (&value, "%s,%s", var->type, var->name)) {
+				goto exit;
+			}
+			sdb_set (core->anal->sdb_types, r_strbuf_get (&key), r_strbuf_get (&value), 0);
 			arg_count++;
 		}
 	}
 	if (arg_count > 0) {
-		char *key = r_str_newf ("func.%s.args", fcn->name);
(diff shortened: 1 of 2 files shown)


thestr4ng3r (approved):

trufae (approved):

libr/core/cmd_anal.c
```diff
@@ -855,16 +855,20 @@ static void __add_vars_sdb(RCore *core, RAnalFunction *fcn) {
...
 			arg_count++;
 		}
 	}
 	if (arg_count > 0) {
-		char *query = r_str_newf ("anal/types/func.%s.args=%d", fcn->name, arg_count);
-		sdb_querys (core->sdb, NULL, 0, query);
-		free (query);
+		char *key = r_str_newf ("func.%s.args", fcn->name);
+		char *value = r_str_newf ("%d", arg_count);
+		sdb_set (core->anal->sdb_types, key, value, 0);
```
trufae, line 869: what about using sdb_num_set and avoid the strnewf/free of the value?

libr/core/cmd_anal.c
```diff
@@ -855,16 +855,20 @@ static void __add_vars_sdb(RCore *core, RAnalFunction *fcn) {
 
 	r_list_foreach (all_vars, iter, var) {
 		if (var->isarg) {
-			char *query = r_str_newf ("anal/types/func.%s.arg.%d=%s,%s", fcn->name, arg_count, var->type, var->name);
-			sdb_querys (core->sdb, NULL, 0, query);
-			free (query);
+			char *key = r_str_newf ("func.%s.arg.%d", fcn->name, arg_count);
+			char *value = r_str_newf ("%s,%s", var->type, var->name);
+			sdb_set (core->anal->sdb_types, key, value, 0);
```
trufae, line 860: no null checks so far.. also maybe is good to go for strbuf_initf. and prefix the call with (void) or check for return value

libr/core/cmd_anal.c
```diff
@@ -855,16 +855,20 @@ static void __add_vars_sdb(RCore *core, RAnalFunction *fcn) {
...
+			free (key);
+			free (value);
 			arg_count++;
 		}
 	}
 	if (arg_count > 0) {
-		char *query = r_str_newf ("anal/types/func.%s.args=%d", fcn->name, arg_count);
-		sdb_querys (core->sdb, NULL, 0, query);
-		free (query);
+		char *key = r_str_newf ("func.%s.args", fcn->name);
```
trufae, line 867: same here. no null chks and maybe use the new strbuf_initf

test/db/cmd/dwarf
```diff
@@ -256,8 +256,8 @@ EXPECT=<<EOF
 |           0x000052b4      488b442450     mov rax, qword [n]
 |           0x000052b9      48c744246001.  mov qword [var_60h], 1
 |           0x000052c2      4889442468     mov qword [var_68h], rax
-|           0x000052c7      488b7c2460     mov rdi, qword [var_60h]
-|           0x000052cc      488b742468     mov rsi, qword [var_68h]
+|           0x000052c7      488b7c2460     mov rdi, qword [var_60h]    ; int64_t arg1
```
trufae, line 259: woot! why is this working now?

trufae (changes requested):

libr/core/cmd_anal.c
karliss: Tried it before. That sets it in hexadecimal form and breaks some tests.

test/db/cmd/dwarf
karliss: As stated in issue description sdb_querys used previosly was vulnerable to query injection. Seems likely that something broke due to that especially with rust names which can have a bunch of special symbols.

libr/core/cmd_anal.c
trufae: The numbers in sdb should be valid in any base. so if that fails it means that somewhere else is not using the api and should be fixed. can you try to find out the place?

libr/core/cmd_anal.c
```diff
@@ -855,16 +855,20 @@ static void __add_vars_sdb(RCore *core, RAnalFunction *fcn) {
...
+		char *value = r_str_newf ("%d", arg_count);
```
trufae, line 868: is arg_count signed? imho it shouldnt

trufae (changes requested):
Add a better PR title. imho the important info from this change is:

* speedup X times of aaa
* use the sdb api instead of query
* fixed sdb injection vulnerability.

Also use a ## tag to make it highlight in the release notes

trufae:
ping

karliss:
Pong.

Don't worry I will address the review comments.

libr/core/cmd_anal.c
karliss: There are quite a few of those. From what I understood in telegram discussion goal of Analysis project and changes done by @thestr4ng3r is to get rid of direct sdb access for such operations and maybe even sdb. I don't think it's worth refactoring within this issue knowing that such code will be removed. This PR supposed to be a small change to avoid a large performance issue without changing behavior.

libr/core/cmd_anal.c
karliss: Better?

libr/core/cmd_anal.c
karliss: Changed to strbuf.

libr/core/cmd_anal.c
karliss: See comment for sdb_num_set

trufae (approved):

PR #17642, opened 2020-09-13
Use PJ in `oij` ##json

officialcjunior:

**Detailed description**

`oij` currently works by constructing a JSON string and printing it. Here, I've reimplemented it using the PJ API.

**Test plan**
Using this, `oij` on one of the test binary prints:
`[{"raised":false,"fd":3,"uri":"bins/elf/crackme0x05","from":0,"writable":false,"size":7656},{"raised":true,"fd":5,"uri":"bins/elf/crackme0x00b","from":0,"writable":false,"size":7375}]`

**Closing issues**

None

### DIFF
--- a/libr/core/cfile.c
+++ b/libr/core/cfile.c
@@ -1061,8 +1061,14 @@ R_API int r_core_file_list(RCore *core, int mode) {
 	RListIter *it;
 	RBinFile *bf;
 	RListIter *iter;
-	PJ * pj = pj_new ();
-	pj_a (pj);
+	PJ *pj;
+	if (mode == 'j') {
+		pj = pj_new ();
+		if (!pj) {
+			return 0;
+		}
+		pj_a (pj);
+	}
 	r_list_foreach (core->files, iter, f) {
 		desc = r_io_desc_get (core->io, f->fd);
 		if (!desc) {
@@ -1072,21 +1078,14 @@ R_API int r_core_file_list(RCore *core, int mode) {
 		from = 0LL;
 		switch (mode) {
 		case 'j': {  // "oij"
-			if (!pj) {
-				free (desc);
-				break;
-			}
 			pj_o (pj);
-			pj_kb (pj, "raised", r_str_bool (core->io->desc->fd == f->fd));
-			pj_ki (pj, "fd", (int) f->fd);
+			pj_kb (pj, "raised", core->io->desc->fd == f->fd);
+			pj_ki (pj, "fd", f->fd);
 			pj_ks (pj, "uri", desc->uri);
 			pj_kn (pj, "from", (ut64) from);
-			pj_kb (pj, "writable", r_str_bool (desc->perm & R_PERM_W));
+			pj_kb (pj, "writable", desc->perm & R_PERM_W);
 			pj_ki (pj, "size", (int) r_io_desc_size (desc));
 			pj_end (pj);
-			pj_end (pj);
-			r_cons_println (pj_string (pj));
-			pj_free (pj);
 			break;
 		}
 		case '*':
@@ -1151,6 +1150,11 @@ R_API int r_core_file_list(RCore *core, int mode) {
 		break;
 		}
 		count++;
+	}
+	if (mode == 'j') {
+		pj_end (pj);
+		r_cons_println (pj_string (pj));
+		pj_free (pj);
 	}
 	return count;
 }
--- a/libr/core/cfile.c
+++ b/libr/core/cfile.c
@@ -1075,7 +1075,6 @@ R_API int r_core_file_list(RCore *core, int mode) {
 		case 'j': {  // "oj"
 			PJ * pj = pj_new ();
 			pj_o (pj);
-			pj_a (pj);
 			pj_ks (pj, "raised", r_str_bool (core->io->desc->fd == f->fd));
 			pj_ki (pj, "fd", (int) f->fd);
 			pj_ks (pj, "uri", desc->uri);
@@ -1084,7 +1083,7 @@ R_API int r_core_file_list(RCore *core, int mode) {
 			pj_ki (pj, "size", (int) r_io_desc_size (desc));
 			pj_end (pj);
 			pj_end (pj);
-			r_cons_printf ("%s\n", pj_string (pj));
+			r_cons_println (pj_string (pj));
 			pj_free (pj);
 			break;
 		}
--- a/libr/core/cfile.c
+++ b/libr/core/cfile.c
@@ -1072,15 +1072,22 @@ R_API int r_core_file_list(RCore *core, int mode) {
 		}
 		from = 0LL;
 		switch (mode) {
-		case 'j':
-			r_cons_printf ("{\"raised\":%s,\"fd\":%d,\"uri\":\"%s\",\"from\":%"
-				PFMT64d ",\"writable\":%s,\"size\":%d}%s",
-				r_str_bool (core->io->desc->fd == f->fd),
-				(int) f->fd, desc->uri, (ut64) from,
-				r_str_bool (desc->perm & R_PERM_W),
-				(int) r_io_desc_size (desc),
-				iter->n? ",": "");
+		case 'j': {  // "oj"
+			PJ * pj = pj_new ();
+			pj_o (pj);
+			pj_a (pj);
+			pj_ks (pj, "raised", r_str_bool (core->io->desc->fd == f->fd));
+			pj_ki (pj, "fd", (int) f->fd);
+			pj_ks (pj, "uri", desc->uri);
+			pj_kn (pj, "from", (ut64) from);
+			pj_ks (pj, "writable", r_str_bool (desc->perm & R_PERM_W));
+			pj_ki (pj, "size", (int) r_io_desc_size (desc));
+			pj_end (pj);
+			pj_end (pj);
+			r_cons_printf ("%s\n", pj_string (pj));
+			pj_free (pj);
 			break;
+		}
 		case '*':
 		case 'r':
 			// TODO: use a getter
--- a/libr/core/cfile.c
+++ b/libr/core/cfile.c
@@ -1061,9 +1061,6 @@ R_API int r_core_file_list(RCore *core, int mode) {
 	RListIter *it;
 	RBinFile *bf;
 	RListIter *iter;
-	if (mode == 'j') {
-		r_cons_printf ("[");
-	}
 	r_list_foreach (core->files, iter, f) {
 		desc = r_io_desc_get (core->io, f->fd);
 		if (!desc) {
@@ -1072,8 +1069,9 @@ R_API int r_core_file_list(RCore *core, int mode) {
 		}
 		from = 0LL;
 		switch (mode) {
-		case 'j': {  // "oj"
+		case 'j': {  // "oij"
 			PJ * pj = pj_new ();
+			pj_a (pj);
 			pj_o (pj);
 			pj_ks (pj, "raised", r_str_bool (core->io->desc->fd == f->fd));
 			pj_ki (pj, "fd", (int) f->fd);
@@ -1149,9 +1147,6 @@ R_API int r_core_file_list(RCore *core, int mode) {
 		break;
 		}
 		count++;
-	}
-	if (mode == 'j') {
-		r_cons_printf ("]\n");
 	}
 	return count;
 }
--- a/libr/core/cfile.c
+++ b/libr/core/cfile.c
@@ -1071,7 +1071,6 @@ R_API int r_core_file_list(RCore *core, int mode) {
 		switch (mode) {
 		case 'j': {  // "oij"
 			PJ * pj = pj_new ();
-			pj_a (pj);
 			pj_o (pj);
 			pj_ks (pj, "raised", r_str_bool (core->io->desc->fd == f->fd));
 			pj_ki (pj, "fd", (int) f->fd);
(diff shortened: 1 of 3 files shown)


karliss:
iAj or oj?

karliss:
Please add a test as asked in the checklist. If you did that, you would see that the code isn't working as you described.

libr/core/cfile.c
```diff
@@ -1072,15 +1072,22 @@ R_API int r_core_file_list(RCore *core, int mode) {
...
+			pj_a (pj);
+			pj_ks (pj, "raised", r_str_bool (core->io->desc->fd == f->fd));
+			pj_ki (pj, "fd", (int) f->fd);
+			pj_ks (pj, "uri", desc->uri);
+			pj_kn (pj, "from", (ut64) from);
+			pj_ks (pj, "writable", r_str_bool (desc->perm & R_PERM_W));
+			pj_ki (pj, "size", (int) r_io_desc_size (desc));
+			pj_end (pj);
+			pj_end (pj);
+			r_cons_printf ("%s\n", pj_string (pj));
```
trufae, line 1087: Use r_cons_println

libr/core/cfile.c
```diff
@@ -1072,15 +1072,22 @@ R_API int r_core_file_list(RCore *core, int mode) {
...
-				PFMT64d ",\"writable\":%s,\"size\":%d}%s",
-				r_str_bool (core->io->desc->fd == f->fd),
-				(int) f->fd, desc->uri, (ut64) from,
-				r_str_bool (desc->perm & R_PERM_W),
-				(int) r_io_desc_size (desc),
-				iter->n? ",": "");
+		case 'j': {  // "oj"
+			PJ * pj = pj_new ();
+			pj_o (pj);
+			pj_a (pj);
```
trufae, line 1078: an array inside an object? this doesnt seems good json to me. i think pj api should check for those invalid json serialization problems

trufae (changes requested):

libr/core/cfile.c
officialcjunior: Oh, okay.
I just went with that, as that's what I've seen commonly implemented, and to make it similar to the current output.

What should be changed?

trufae: {[1,2,3]} is invalid json. Just remove the pj_a

karliss:
Still wrong, still no test. And that part of code isn't even executed by `oj`.

officialcjunior:
Okay. I had mistook this as the code of `oj` as it has similar outputs. I still have no idea what this command is. I realize that I am very short-sighted and senseless, at the moment. If you could tell me a way to find what this command is, I can rename this PR and add a test for it.

karliss:
When writing code and doing local testing you should notice that making changes to code have no effect on output. A dumb but effective way for verifying that you are working with the right code is adding extra print statements, if their output is not visible then something is wrong.

As with all other commands there two simple approaches that work most of the time for locating required code. One is to search by command name something like "oj". I assume this is what you did, but it didn't work out this time. Just searching for such comment comment doesn't guarantee that it's correct code so it's worth verifying the function containing it and where it's called.
If you look at search results carefully you can see that there are two results, this is a good indicator you have to start paying attention and reading the code more carefully. This can be considered bottom to top approach. 

Other approach is following the path from top to bottom. Start with the top level command handler starting with cmd_open and follow the the switch statements and function calls handling command letters one by one until you reach `oj`. In case of `oj` that's only two steps.

How could you know this being new to r2 source code Follow the function calls. You can apply this approach to learning any code base.
* Get your bearing. Locate code handling the part you are interested in or something similar. Good approach for this searching for ui strings, help messages, warnings that appear when interacting with corresponding feature. Especially for command line tools like r2.
* Follow the function calls and other references up. Sometimes the string you found might not be directly next the function you want or if you started with some similar method you might need to go a few levels up. Follow the function calls and variable names. Until you reach the common code. While doing this also pay attention of code structure. This will help in next step when you follow the code in opposite direction.
* Follow the function calls top to bottom. Use the things you learned previously for which path is more likely.

IDE features for finding function declaration or references to function are nice but even with regular you can usually find the right thing easily as global function names are typically sufficiently long and unique. Most code editors have a feature for searching text in all files inside project directory. Learn to use it. In case of less integrated development process you can use tools like grep.

Remember that software development isn't just writing code but also reading and navigating code written by others. And you need to practice all of those skills.

In this specific case it seems that the code you modified used to be used by `oj` command but not anymore so it should be cleaned up so that there aren't two implementations.

officialcjunior:
Thank you so much, @karliss.
And I did find it. It's `oij`.

Also, about removing `pj_a()`:
Currently it r_cons_prints the square brackets at line [L1064](
  https://github.com/radareorg/radare2/blob/c06d82eb9b3ba75f29940a7550af0ed68db8f66b/libr/core/cfile.c#L1064) and [L1147](https://github.com/radareorg/radare2/blob/c06d82eb9b3ba75f29940a7550af0ed68db8f66b/libr/core/cfile.c#L1147).

Removing the `pj_a` and `r_cons_println`-ing the `pj_string` will output `oij` like this: (as there's a newline) 
```
[{"raised":"true","fd":3,"uri":"../ch18.bin","from":0,"writable":"false","size":5240}
]
```

So, I think it's better to remove printing square brackets and get a `pj_a()`, to fix this.

What should I do?

karliss:
> What should I do?

1) Learn how to run testsuite, see README in test folder
2) Create a test, doesn't need to be final, you can adjust it later. Try adding a bit more than just basic case, for example add a case where there are two files in the list.
3) Run the test
4) Observe that for test with two opened files it doesn't properly print them
5) Get rid of manual `[` printing and use appropriate PJ api for describing list of elements as you suggested yourself. The whole point of this PR was to replace manual JSON printing with use of PJ. Using PJ for parts of single JSON output and then gluing them together like it's done currently defeats the point of this PR.

officialcjunior:
Removed the manual printing part in the output.

I also added a basic test and ran it. I'm not sure about whether I have done it correctly and whether this is what you meant by adding two files as, I didn't encounter any failed tests.
 
```
$ r2r -i db/cmd/cmd_open 
Running from /home/cjunior/Downloads/programs/radare2/test
Loaded 30 tests.
[30/30]                    29 OK         1 BR        0 XX        0 FX
Finished in 1 seconds.

```

officialcjunior:
Removed the 'manual printing' part

Added a basic test for this. I'm not sure whether I've done it right, as:
1. I'm slightly confused whether this is what you meant by opening two files.
2. The test didn't encounter any errors:

```
$ r2r -i db/cmd/cmd_open 
Running from /home/cjunior/Downloads/programs/radare2/test
Loaded 30 tests.
[30/30]                    29 OK         1 BR        0 XX        0 FX
Finished in 0 seconds.

```

karliss:
> I'm slightly confused whether this is what you meant by opening two files.

Two files at the same time so that the list contains more than one entry instead of two files. Two files one at a time doesn't add any value to the test. See `o?` for commands that can be used to open additional files.

officialcjunior:
Under, `o?`, I tried `o [filename]`. It doesn't accept two arguments, so I opened two files twice, but that seems to have the same effect as opening two files normally.

Then, there's `oc`, `of` and `oi`, which I'm not sure what it is used for.

Please share the syntax to open two files the way you mean,

karliss:
Looks like it the same.Not sure if it's an intended way of using r2r but does what I wanted and demonstrates that output isn't valid json.

officialcjunior:
```
[0x08048320]> oj
[{"raised":true,"fd":3,"uri":"ch18.bin","from":0,"writable":false,"size":5240},{"raised":false,"fd":4,"uri":"null://8","from":0,"writable":true,"size":8},{"raised":false,"fd":5,"uri":"malloc://983040","from":0,"writable":true,"size":983040}]
[0x08048320]> oij
[{"raised":"true","fd":3,"uri":"ch18.bin","from":0,"writable":"false","size":5240}]
```
So, the JSON notation is correct now, isn't it?

karliss:
Take a look at expected output in the test, that isn't valid JSON.

officialcjunior:
I hope that's fixed now, if not, please request changes.

Maijin:
Add a test in JSON test db/json/

officialcjunior:
Yeah, is that it @Maijin ? Just adding the word?

Maijin:
yes

libr/core/cfile.c
officialcjunior: Replaced with r_cons_println.
Resolving this as marked

libr/core/cfile.c
```diff
@@ -1072,15 +1069,21 @@ R_API int r_core_file_list(RCore *core, int mode) {
...
-				(int) r_io_desc_size (desc),
-				iter->n? ",": "");
+		case 'j': {  // "oij"
+			PJ * pj = pj_new ();
+			pj_o (pj);
+			pj_ks (pj, "raised", r_str_bool (core->io->desc->fd == f->fd));
+			pj_ki (pj, "fd", (int) f->fd);
+			pj_ks (pj, "uri", desc->uri);
+			pj_kn (pj, "from", (ut64) from);
+			pj_ks (pj, "writable", r_str_bool (desc->perm & R_PERM_W));
```
trufae, line 1079: writable should be a bool. use pj_kb, not a string

libr/core/cfile.c
```diff
@@ -1072,15 +1069,21 @@ R_API int r_core_file_list(RCore *core, int mode) {
...
-				PFMT64d ",\"writable\":%s,\"size\":%d}%s",
-				r_str_bool (core->io->desc->fd == f->fd),
-				(int) f->fd, desc->uri, (ut64) from,
-				r_str_bool (desc->perm & R_PERM_W),
-				(int) r_io_desc_size (desc),
-				iter->n? ",": "");
+		case 'j': {  // "oij"
+			PJ * pj = pj_new ();
+			pj_o (pj);
+			pj_ks (pj, "raised", r_str_bool (core->io->desc->fd == f->fd));
```
trufae, line 1075: raised must be a bool. not a string

libr/core/cfile.c
```diff
@@ -1072,15 +1069,21 @@ R_API int r_core_file_list(RCore *core, int mode) {
...
-		case 'j':
-			r_cons_printf ("{\"raised\":%s,\"fd\":%d,\"uri\":\"%s\",\"from\":%"
-				PFMT64d ",\"writable\":%s,\"size\":%d}%s",
-				r_str_bool (core->io->desc->fd == f->fd),
-				(int) f->fd, desc->uri, (ut64) from,
-				r_str_bool (desc->perm & R_PERM_W),
-				(int) r_io_desc_size (desc),
-				iter->n? ",": "");
+		case 'j': {  // "oij"
+			PJ * pj = pj_new ();
```
trufae, line 1073: check for null

trufae (changes requested):

libr/core/cfile.c
officialcjunior: Did you mean:
```
if (!pj) {
  free (out);
  break;
}
```

trufae: yes

libr/core/cfile.c
trufae: ping

libr/core/cfile.c
officialcjunior: Replaced with `pj_kb`

libr/core/cfile.c
officialcjunior: Replaced with `pj_kb`

officialcjunior:
So, I did what you asked me to, but the bool for `writable` got swapped, as you can see in the tests.
Is that alright?

karliss:
Output still isn't a valid JSON. `{}{}` is not JSON.

trufae (changes requested):
Move the pj_new outside the loop and you are opening one object and closing it twice. You should open an array outside the loop and an object inside...

officialcjunior:
I think I've correctly did what you had asked. Now, the output is :

`[{"raised":true,"fd":3,"uri":"test/bins/elf/crackme0x05","from":0,"writable":true,"size":7656}`

But as @karliss had pointed out earlier, this fails when we open two files.

trufae (approved):

officialcjunior:
Hey @XVilka, this seems to be done.

test/db/cmd/cmd_open
```diff
@@ -105,8 +105,7 @@ bins/elf/crackme0x00b
 EOF
 CMDS=oij
 EXPECT=<<EOF
-{"raised":true,"fd":3,"uri":"bins/elf/crackme0x05","from":0,"writable":true,"size":7656}
-{"raised":true,"fd":5,"uri":"bins/elf/crackme0x00b","from":0,"writable":true,"size":7375}
+[{"raised":false,"fd":3,"uri":"bins/elf/crackme0x05","from":0,"writable":false,"size":7656},{"raised":true,"fd":5,"uri":"bins/elf/crackme0x00b","from":0,"writable":false,"size":7375}]
```
Maijin, line 108: don't forget to also add a json test in https://github.com/radareorg/radare2/tree/master/test/db/json

officialcjunior: Hey, I have added that:

https://github.com/radareorg/radare2/pull/17642/files#diff-11f08680e054bf1235c7c156ec3962f4272ae558b64c44158f965f1d287cbf49

officialcjunior: Marking as resolved.

eagleoflqj:
Ping @XVilka

XVilka (approved):

XVilka:
Sorry for the delay, didn't notice it before.

PR #17643, opened 2020-09-13
Move coverage testing to  meson-gcc-newshell-tests

karliss:

**Detailed description**

Previously coverage was measured on meson-gcc-tests which isn't run on pull requests. meson-gcc-newshell-tests should do more or less the same code except part of command parsing code.

**Test plan**

Wait for pull request tests to finish. Make sure test coverage for whole project not the diff is ~40% same as when master is being tested.

**Closing issues**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,9 +45,7 @@ jobs:
             build_system: meson
             compiler: gcc
             run_tests: true
-            meson_options: -Db_coverage=true
-            coverage: true
-            enabled: ${{ github.event_name != 'test__pull_request' }}
+            enabled: ${{ github.event_name != 'pull_request' }}
             timeout: 45
           - name: linux-acr-clang-build
             os: ubuntu-latest
@@ -59,9 +57,7 @@ jobs:
             os: ubuntu-latest
             build_system: meson
             compiler: gcc
-            meson_options: -Db_coverage=true
-            coverage: true
-            enabled: ${{ github.event_name == 'test__fpull_request' }}
+            enabled: ${{ github.event_name == 'pull_request' }}
             timeout: 45
           - name: linux-meson-gcc-newshell-tests
             os: ubuntu-latest
@@ -72,7 +68,7 @@ jobs:
             meson_options: -Db_coverage=true
             coverage: true
             enabled: true
-            timeout: 60
+            timeout: 50
           - name: macos-clang-tests
             os: macos-latest
             build_system: acr
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
             meson_options: -Db_coverage=true
             coverage: true
             enabled: true
-            timeout: 45
+            timeout: 60
           - name: macos-clang-tests
             os: macos-latest
             build_system: acr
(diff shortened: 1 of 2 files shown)


karliss:
No coverage report. Maybe it doesn't report anything if no C files are modified.

karliss:
It succeeded on first run although with 39.9% coverage running ~26 minutes and on the second run it managed to hit 45 minute timeout.

karliss:
Looks like meson-gcc-tests with coverage testing in the master runs ~40 minutes which is close to the 45 minute time limit. In that case it's a bit weird that it managed to run in 27 minutes the first time.

Are we fine with test run time increase from 20 to 40 minutes for pull requests? Didn't expect coverage measurement to cause such a slowdown. Although windows tests on appveyor also take 45 minutes so it shouldn't increase the longest running test performed on pull requests.

karliss:
Correction 40 minutes in master include fuzz tests which take ~10 minutes. 40-10 ~ 27 minutes makes sense. It shouldn't time out as it did.

kazarmy:
Note that ci.yml sets `-Db_coverage=true` but `b_coverage` isn't checked anywhere in the Meson build code.

ret2libc:
@kazarmy what do you mean? That option should be a built in one.

ret2libc:
So... do we really want the codecov run on every PR? I'm not sure how much things can improve from PR to PR and how much we would look at it. Maybe it is ok to just run this on master, what do you think?

kazarmy:
> @kazarmy what do you mean? That option should be a built in one.

Orly? (ok just found https://mesonbuild.com/Builtin-options.html#base-options ... sure wish it was more transparent though)

ret2libc:
> > @kazarmy what do you mean? That option should be a built in one.
> 
> Orly? (ok just found https://mesonbuild.com/Builtin-options.html#base-options ... sure wish it was more transparent though)

I think all the `b_*` options are built-in.

kazarmy:
Don't think CMake has built-in options to that extent, much less GNU make

karliss:
I think I was able to repeat testsuite getting stuck. Need to investigate more as that shouldn't happen even if we don't want to run coverage on PR.

> So... do we really want the codecov run on every PR? I'm not sure how much things can improve from PR to PR 

Absolute numbers probably won't change PR to PR but code cov also reports the coverage for diff. That can be useful to quickly evaluate if newly added code has matching amount of tests added.

karliss:
I have no more ideas why it times out now. It didn't timeout on the first run. It also doesn't time out in master.

karliss:
This time linux-meson-gcc-tests timed out so it isn't specific to combination of newshell and coverage measurement.

karliss:
One percent difference between this branch and master could be due to fuzzing being run only on master.

ret2libc:
> This time linux-meson-gcc-tests timed out so it isn't specific to combination of newshell and coverage measurement.

This kind of random timeouts are probably due to bugs in r2r. Or at least, they started appearing when we introduced r2r. They seemed to happen mainly in macos though. Just a random thought, maybe it's not that :)

trufae:
There are 2 or 3 tests that take almost all the time of the testsuite to run. its just wrongly written tests or too heavy ones. will be good to show them up in the r2r output (top 10 slowest tests for example and their times). this will help us debug and fix those bottlenecks

ret2libc:
Also master sometimes fail and this is bad :(

karliss:
Looks acceptable?

ret2libc (approved):

PR #17644, opened 2020-09-13
Better use of the R_SOCKET_PROTO_DEFAULT and add _NONE

trufae:
DO NOT SQUASH. JUST REBASE

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


ret2libc (commented):
Is this related to the recent PR about R_SOCKET_PROTO_TCP or it is a different thing?

trufae:
It is the same fix. but done by me and a week ago. for soem reason github didnt spotted this branch as a PR so it didnt make it into the repo :(( i just opened this PR so the other contributor can see the fixes and get the best of both. ill close once merged the other

ret2libc:
> It is the same fix. but done by me and a week ago. for soem reason github didnt spotted this branch as a PR so it didnt make it into the repo :(( i just opened this PR so the other contributor can see the fixes and get the best of both. ill close once merged the other

I see. So you don't need a review on this one, right?

trufae:
Yes, the other pr was merged. and i rebased this one and applied further improvements here.. ill add a test.. but -C is broken too..

libr/socket/socket.c
```diff
@@ -494,12 +494,12 @@ R_API bool r_socket_listen(RSocket *s, const char *port, const char *certfile) {
 		return false;
 	}
 #endif
-
-	switch (s->proto) {
-	case R_SOCKET_PROTO_DEFAULT:
+	if (s->proto == R_SOCKET_PROTO_NONE) {
```
ret2libc, line 497: just add this to the switch below

trufae: NONE is not a proto by itself is kind of pre-check before entering the switch, so imho its cleaner to have this early if instead. but i an do the fallback thing if u like

ret2libc: Yeah I think I prefer
```C
switch (s->proto) {
case R_SOCKET_PROTO_NONE:
case R_SOCKET_PROTO_TCP:
    s->proto = R_SOCKET_PROTO_TCP;
    .....
}
```
BTW, the `s->proto = R_SOCKET_PROTO_TCP` is not needed if you don't do as I suggested above.

trufae:
Anyone have a magic idea to test the webserver? maybe using rarun2 to spawn the child with a timeout :?

libr/socket/socket.c
ret2libc: Why marked as resolved?

ret2libc:
Nope, sorry.

libr/socket/socket.c
trufae: your solution is wrong because its asuming that TCP is the DEFAULT, when we already have a DEFAULT variable that can change at any time and shouldnt break the rest of the code. so having NONE as an alias for TCP is wrong. putting NONE in the swith table is also wrong because it forces to refactor the code in the switch table everytime the default protocol is changed. shouldnt happenfrequently, but its an unnecessary weak point. and actually imho it should NEVER be NONE. because the socket constructor must use DEFAULT, and the connect/listen methods are tied to the protocol so that field may be set before reaching that point.

trufae:
I solved the testing of the http server using a new feature in rarun2. Not perfect yet. but at least it works reliabily and we have tests to avoid regressions. i would like to get this merged when the windows CI gets green. i will work on further web improvements but without a working webserver i cant do anything

libr/socket/socket.c
```diff
@@ -494,12 +494,12 @@ R_API bool r_socket_listen(RSocket *s, const char *port, const char *certfile) {
...
-
+	if (s->proto == R_SOCKET_PROTO_NONE) {
+		s->proto = R_SOCKET_PROTO_DEFAULT;
+	}
 	switch (s->proto) {
-	case R_SOCKET_PROTO_DEFAULT:
-		s->proto = R_SOCKET_PROTO_TCP;
-		/* fallthrough */
 	case R_SOCKET_PROTO_TCP:
+		s->proto = R_SOCKET_PROTO_TCP;
```
ret2libc, line 502: unnecessary

GustavoLCR:
This patch should fix windows:

[patch.txt](https://github.com/radareorg/radare2/files/5258977/patch.txt)

trufae:
Thanks @GustavoLCR ! just applied your patch. let's see how this ends now

trufae:
half windows are red with this error now:

Screenshot 2020-09-22 at 15 53 27

XVilka:
Still red with timeout:
```
[XX] TIMEOUT C:\projects\radare2\test\db\io\http r2 remote webserver two
ANSICON=1 R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc '!rarun2 timeout=2 daemon=1 system="r2 -e http.root=/404 -e http.port=9292 -c=h -qcq -"
sleep 1
=+ http://127.0.0.1:9292/cmd/
=
=0 ?e hello remote world
' --
-- stderr
@@ -1,7 +1,8 @@
-Cannot find http.root (/404) or http.homeroot (~/.local/share/radare2/www)
+PID: Cannot determine pid with 'system' directive. Use 'program'.
+Cannot find http.root (/404) or http.homeroot (~\.local\share\radare2\www)
 Starting http server...
 open http://localhost:9292/
 r2 -C http://localhost:9292/cmd/
-Connected to: 'http://127.0.0.1'
 
 rarun2: Interrupted by timeout
+Connected to: 'http://127.0.0.1'
```

libr/socket/run.c
```diff
@@ -1111,7 +1125,59 @@ R_API int r_run_start(RRunProfile *p) {
...
+				}
+#else
+				eprintf ("timeout not supported for this platform\n");
+#endif
+			}
+#endif
+#if __UNIX__
+			close(0);
+			close(1);
+			exit (execl ("/bin/sh","/bin/sh", "-c", p->_system, NULL));
```
ret2libc, line 1171: Could you just call `r_sys_cmd` both here and in the `#else`? (and just move it outside of the `#if`)

test/db/io/http
```diff
@@ -15,3 +15,47 @@ EXPECT=<<EOF
 EOF
 RUN
 
+# NAME=r2 remote webserver
```
ret2libc, line 18: why is this commented?

libr/socket/run.c
trufae: no. exec != system. one replaces the current process. the other creates a new. its needed to handle the pid properly

test/db/io/http
trufae: i can remove it, i was about to add more http tests. but i think its fine to go one by one :D

trufae:
windows:

```
c:\projects\radare2\libr\util\sys.c(180): error C2065: 'sig_t': undeclared identifier [C:\projects\radare2\build\libr\util\dc150f5@@r_util@sta.vcxproj]
1491c:\projects\radare2\libr\util\sys.c(180): error C2146: syntax error: missing ';' before identifier 'ret' [C:\projects\radare2\build\libr\util\dc150f5@@r_util@sta.vcxproj]
1492c:\projects\radare2\libr\util\sys.c(180): error C2065: 'ret': undeclared identifier [C:\projects\radare2\build\libr\util\dc150f5@@r_util@sta.vcxproj]
1493c:\projects\radare2\libr\util\sys.c(181): error C2065: 'ret': undeclared identifier [C:\projects\radare2\build\libr\util\dc150f5@@r_util@sta.vcxproj]
```

trufae:
Windows is now green, and this PR should be making master green too

ret2libc:
You should rebase/merge to make sure you are working on latest master and check whether this really fixes broken tests.

trufae:
it was green before rebasing, and will be green after rebase. but i wont get the responsability to merge this. this PR can probably be splitted up into 20 different ones.

test/scripts/test-webserver.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/sh
 
-r2 -qq -c=h - &
+r2 -N -e http.port=9393 -qq -c=h bins/elf/arg > /dev/null 2>&1 &
 CHILD=$!
 sleep 2
```
kazarmy, line 5: unless r2 retries, this could be racy

libr/core/rtr_http.c
```diff
@@ -23,10 +23,8 @@ static int r_core_rtr_http_run(RCore *core, int launch, int browse, const char *
 
 	if (!r_file_is_directory (root)) {
 		if (!r_file_is_directory (homeroot)) {
-			eprintf ("Cannot find http.root (%s) or http.homeroot (%s)\n", root, homeroot);
-			return false;
+			eprintf ("Cannot find http.root or http.homeroot\n");
```
ret2libc, line 26: can this function run even withoutn http.root/homeroot?

libr/util/sys.c
```diff
@@ -73,6 +73,7 @@ extern char **environ;
 #include <psapi.h>
 #include <process.h>  // to allow getpid under windows msvc compilation
 #include <direct.h>  // to allow getcwd under windows msvc compilation
+#define sig_t int
```
ret2libc, line 76: I think this is wrong. `signal` does not return a int and it was already fixed in https://github.com/radareorg/radare2/commit/83d3e7bf1872b93af059f1f912d2e5199c8b51dd

test/scripts/test-webserver.sh
```diff
@@ -1,7 +1,9 @@
...
-curl -s --retry 6 --retry-connrefused http://localhost:9090/ | head -n 8
```
ret2libc, line 6: why commented?

test/db/anal/mach0
```diff
@@ -2,42 +2,49 @@ NAME=aav find xrefs
 FILE=bins/mach0/macho-bash
 CMDS=<<EOF
 e anal.loads=true
-aa;afna;aac;aar;aan;aav
```
ret2libc, line 5: this does not seem to be related to R_SOCKET, http, or anything like that.

libr/socket/run.c
```diff
@@ -1041,13 +1048,17 @@ R_API int r_run_config_env(RRunProfile *p) {
 			}
 			sleep (p->_timeout);
 			if (!kill (mypid, 0)) {
-				eprintf ("\nrarun2: Interrupted by timeout\n");
+				// eprintf ("\nrarun2: Interrupted by timeout\n");
```
ret2libc, line 1051: commented code

libr/socket/run.c
```diff
@@ -706,6 +710,13 @@ static int redirect_socket_to_stdio(RSocket *sock) {
 	return 0;
 }
 
+#if __WINDOWS__
+static RThreadFunctionRet exit_process(RThread *th) {
+	// eprintf ("\nrarun2: Interrupted by timeout\n");
```
ret2libc, line 715: commented code

libr/socket/run.c
```diff
@@ -865,11 +876,7 @@ R_API int r_run_config_env(RRunProfile *p) {
 				is_child = true;
 
 				if (p->_dofork && !p->_dodebug) {
-#ifdef _MSC_VER
-					int child_pid = r_sys_fork ();
-#else
 					pid_t child_pid = r_sys_fork ();
```
ret2libc, line 879: `pid_t` is not necessary and this should be `int child_pid`, because `r_sys_fork` returns int.

libr/socket/run.c
```diff
@@ -54,6 +54,7 @@
 #ifdef _MSC_VER
 #include <direct.h>   // to compile chdir in msvc windows
 #include <process.h>  // to compile execv in msvc windows
+#define pid_t int
```
ret2libc, line 57: not necessary

libr/parse/meson.build
```diff
@@ -8,12 +8,12 @@ r_parse_sources = [
   'p/parse_att2intel.c',
   'p/parse_avr_pseudo.c',
   'p/parse_chip8_pseudo.c',
-  'p/parse_tms320_pseudo.c',
```
ret2libc, line 11: unneeded and unrelated change

kazarmy (approved):

libr/core/rtr_http.c
kazarmy: the webserver apparently has a REST api

kazarmy:
Oh no I accidentally squashed the pr ... sorry @trufae

test/scripts/test-webserver.sh
kazarmy: probably because that line might come back into play

trufae:
You dont need to say sorry @kazarmy , you are not pancake.

Also there was a bunch of comments pending from @ret2libc, i will eventually answer them, but i would prefer this PR to be reverted, properly rebased and splitted into 20 different PRs, because what started as a simple http server fix end up finding and fixing more bugs most of them depending between them

ret2libc:
To me It seems like now it's too late and reverting+splitting would be additional work for unclear advantage.

PR #17650, opened 2020-09-14
Fix typo in r2pm

bazuchan:


binr/r2pm/r2pm
```diff
@@ -334,7 +334,7 @@ R2PM_GIT() {
 		cd "${DIR}" || R2PM_FAIL "Oops"
 	fi
 	r2 -qv | grep -q git || { # r2 >= 4.5
-		git checkout "r2-${R2_VERSION}"
+		git checkout "r2-${R2VERSION}"
```
trufae, line 337: oh! i see, this variable is set at the begining of the script. but i think the right var name should be R2_VERSION. i would suggest using something like this:

```
$ eval $(r2 -H)
$ echo $R2_VERSION
4.6.0-git
$
```

trufae (changes requested):
Can you try this eval line and update the PR? Nice catch

trufae:
ping

bazuchan:
will do tomorrow

binr/r2pm/r2pm
zamaudio: So is this PR a no-op then?

trufae (approved):

PR #17651, opened 2020-09-14
Fix spacing bugs when parsing the calling convention signatures

trufae:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/test/db/cmd/cmd_tcc
+++ b/test/db/cmd/cmd_tcc
@@ -25,3 +25,49 @@ tfc rax ms (rcx, rdx, r8, r9, stack);
 tfc rax swift (rdi, rsi, rdx, rcx, r8, r9, xmm0, xmm1, xmm2, xmm3, xmm4, r13, r12);
 EOF
 RUN
+
+NAME=tcc
+FILE=-
+CMDS=<<EOF
+e asm.arch=x86
+e asm.bits=32
+?e 1
+tcc eax jeje(ebx, ecx)
+tcc jeje
+tcc-jeje
+?e 2
+tcc eax    jeje(ebx, ecx)
+tcc   jeje
+tcc-jeje
+?e 3
+tcc eax  jeje   (ebx, ecx)
+tcc   jeje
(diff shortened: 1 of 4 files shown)


trufae:
@ret2libc is this a bug in newshell?

ret2libc:
> @ret2libc is this a bug in newshell?

As the bug seems to be in the tcc function itself, I would say that the same problem applies there as well. However, once we move the `tcc` command handler to accept argv-style arguments, spaces are ignored, unless you wrap everything in `"..."`.

trufae:
So quoting the whole command will be good for the newparser in this case?

test/db/cmd/cmd_tcc
```diff
@@ -25,3 +25,49 @@ tfc rax ms (rcx, rdx, r8, r9, stack);
...
+
+NAME=tcc
+FILE=-
+CMDS=<<EOF
+e asm.arch=x86
+e asm.bits=32
+?e 1
+tcc eax jeje(ebx, ecx)
+tcc jeje
+tcc-jeje
```
ret2libc, line 37:
```suggestion
tcc- jeje
```
otherwise the newshell won't be able to execute this properly.
newshell parses the line and it splits it into `<cmd-name>` and `<arguments>`. Since we have all kinds of commands and `-` is not in all cases the last character of the command name, it is required to have a space to separate the `<cmd-name>` from `<arguments>`.

ret2libc: This should fix the newshell tests.

test/db/cmd/cmd_tcc
```diff
@@ -34,7 +34,7 @@ e asm.bits=32
 ?e 1
 tcc eax jeje(ebx, ecx)
 tcc jeje
-tcc-jeje
+tcc- jeje
```
ret2libc, line 37: same for all the others of course ;)

ret2libc (approved):

PR #17652, opened 2020-09-14
Fix #6467 - Add opn/opp/opr commands to priorize next/previous/rotate file ##io

### LINKED ISSUE #6467: Add 'o' subcommands to open next/prev file, count opened files, etc

trufae:
**DO NOT SQUASH**
================

Based on https://github.com/radareorg/radare2/pull/17271 

* I fixed the review comments i wrote there
* fixed next/prev/lowest/highest apis
* confirm behaviour with r2 tests
* this is ready to review, but ill squash and rebase all the commits properly when the review is done

 

**Detailed description**

See https://github.com/radareorg/radare2/pull/17162 and https://github.com/radareorg/radare2/pull/17161

**Test plan**

There are unit and r2 tests

**Closing issues**

None. thats a feature

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -30,7 +30,7 @@ static const char *help_msg_o[] = {
 	"om","[?]","create, list, remove IO maps",
 	"on"," [file] 0x4000","map raw file at 0x4000 (no r_bin involved)",
 	"oo","[?+bcdnm]","reopen current file (see oo?) (reload in rw or debugger)",
-	"op"," [fd]", "select the given fd as current file (see also ob)",
+	"op","[r|n|p|fd]", "select priorized file by fd (see ob), opn/opp/opr = next/previous/rotate",
 	"oq","","list all open files",
 	"ox", " fd fdx", "exchange the descs of fd and fdx and keep the mapping",
 	NULL
@@ -41,6 +41,15 @@ static const char *help_msg_o_[] = {
 	"o-*","","close all opened files",
 	"o-!","","close all files except the current one",
 	"o-3","","close fd=3",
+	NULL
+};
+
+static const char *help_msg_op[] = {
+	"Usage:", "op[rnp] [fd]", "",
+	"opr", "", "open next file rotating",
+	"opn", "", "open next file",
+	"opp", "", "open previous file",
+	"op", " [fd]", "open priorizing fd",
 	NULL
 };
 
@@ -1209,6 +1218,37 @@ static bool desc_list_json_cb(void *user, void *data, ut32 id) {
 	return true;
 }
 
+static bool cmd_op(RCore *core, char mode, int fd) {
+	int cur_fd = r_io_fd_get_current (core->io);
+	int next_fd = cur_fd;
+	switch (mode) {
+	case 0:
+		next_fd = fd;
+		break;
+	case 'n':
+		next_fd = r_io_fd_get_next (core->io, cur_fd);
+		break;
+	case 'p':
+		next_fd = r_io_fd_get_prev (core->io, cur_fd);
+		break;
+	case 'r':
+		next_fd = r_io_fd_get_next (core->io, cur_fd);
+		if (next_fd == -1) {
+			next_fd = r_io_fd_get_lowest (core->io);
+		}
+		break;
+	}
+	if (next_fd >= 0 && next_fd != cur_fd && r_io_use_fd (core->io, next_fd)) {
+		RBinFile *bf = r_bin_file_find_by_fd (core->bin, next_fd);
+		if (bf && r_core_bin_raise (core, bf->id)) {
+			r_core_block_read (core);
+			return true;
+		}
+		eprintf ("Invalid RBinFile.id number.\n");
+	}
+	return false;
+}
+
 static int cmd_open(void *data, const char *input) {
 	RCore *core = (RCore*)data;
 	int perms = R_PERM_R;
@@ -1366,20 +1406,27 @@ static int cmd_open(void *data, const char *input) {
 	case 'p': // "op"
 		/* handle prioritize */
 		if (input[1]) {
-			int fd = r_num_math (core->num, input + 1);
-			if (fd >= 0 || input[1] == '0') {
-				RIODesc *desc = r_io_desc_get (core->io, fd);
-				if (desc) {
-					// only useful for io.va=0
-					// load bininfo for given fd
-					r_core_cmdf (core, "obo %d", fd);
-					core->io->desc = desc; // XXX we should use fd here, not *pointer
-					r_core_block_read (core);
-				} else {
-					eprintf ("Cannot find RBinFile associated with fd %d\n", fd);
+			switch (input[1]) {
+			case 'r': // "opr" - open next file + rotate if not found
+			case 'n': // "opn" - open next file
+			case 'p': // "opp" - open previous file
+				if (!cmd_op (core, input[1], -1)) {
+					eprintf ("Cannot find file\n");
 				}
-			} else {
-				eprintf ("Invalid fd number\n");
+				break;
+			case ' ':
+				{
+					int fd = r_num_math (core->num, input + 1);
+					if (fd >= 0 || input[1] == '0') {
+						cmd_op (core, 0, fd);
+					} else {
+						eprintf ("Invalid fd number\n");
+					}
+				}
+				break;
+			default:
+				r_core_cmd_help (core, help_msg_op);
+				break;
 			}
 		} else {
 			if (core->io && core->io->desc) {
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -199,38 +199,38 @@ typedef struct r_io_desc_cache_t {
 
 struct r_io_bind_t;
 
-typedef bool (*RIODescUse) (RIO *io, int fd);
-typedef RIODesc *(*RIODescGet) (RIO *io, int fd);
-typedef ut64 (*RIODescSize) (RIODesc *desc);
-typedef RIODesc *(*RIOOpen) (RIO *io, const char *uri, int flags, int mode);
-typedef RIODesc *(*RIOOpenAt) (RIO *io, const  char *uri, int flags, int mode, ut64 at);
-typedef bool (*RIOClose) (RIO *io, int fd);
-typedef bool (*RIOReadAt) (RIO *io, ut64 addr, ut8 *buf, int len);
-typedef bool (*RIOWriteAt) (RIO *io, ut64 addr, const ut8 *buf, int len);
-typedef char *(*RIOSystem) (RIO *io, const char* cmd);
-typedef int (*RIOFdOpen) (RIO *io, const char *uri, int flags, int mode);
-typedef bool (*RIOFdClose) (RIO *io, int fd);
-typedef ut64 (*RIOFdSeek) (RIO *io, int fd, ut64 addr, int whence);
-typedef ut64 (*RIOFdSize) (RIO *io, int fd);
-typedef bool (*RIOFdResize) (RIO *io, int fd, ut64 newsize);
-typedef ut64 (*RIOP2V) (RIO *io, ut64 pa);
-typedef ut64 (*RIOV2P) (RIO *io, ut64 va);
-typedef int (*RIOFdRead) (RIO *io, int fd, ut8 *buf, int len);
-typedef int (*RIOFdWrite) (RIO *io, int fd, const ut8 *buf, int len);
-typedef int (*RIOFdReadAt) (RIO *io, int fd, ut64 addr, ut8 *buf, int len);
-typedef int (*RIOFdWriteAt) (RIO *io, int fd, ut64 addr, const ut8 *buf, int len);
-typedef bool (*RIOFdIsDbg) (RIO *io, int fd);
-typedef const char *(*RIOFdGetName) (RIO *io, int fd);
-typedef RList *(*RIOFdGetMap) (RIO *io, int fd);
-typedef bool (*RIOFdRemap) (RIO *io, int fd, ut64 addr);
-typedef bool (*RIOIsValidOff) (RIO *io, ut64 addr, int hasperm);
-typedef RIOMap *(*RIOMapGet) (RIO *io, ut64 addr);
-typedef RIOMap *(*RIOMapGetPaddr) (RIO *io, ut64 paddr);
-typedef bool (*RIOAddrIsMapped) (RIO *io, ut64 addr);
-typedef RIOMap *(*RIOMapAdd) (RIO *io, int fd, int flags, ut64 delta, ut64 addr, ut64 size);
+typedef bool (*RIODescUse)(RIO *io, int fd);
+typedef RIODesc *(*RIODescGet)(RIO *io, int fd);
+typedef ut64(*RIODescSize)(RIODesc *desc);
+typedef RIODesc *(*RIOOpen)(RIO *io, const char *uri, int flags, int mode);
+typedef RIODesc *(*RIOOpenAt)(RIO *io, const  char *uri, int flags, int mode, ut64 at);
+typedef bool (*RIOClose)(RIO *io, int fd);
+typedef bool (*RIOReadAt)(RIO *io, ut64 addr, ut8 *buf, int len);
+typedef bool (*RIOWriteAt)(RIO *io, ut64 addr, const ut8 *buf, int len);
+typedef char *(*RIOSystem)(RIO *io, const char* cmd);
+typedef int (*RIOFdOpen)(RIO *io, const char *uri, int flags, int mode);
+typedef bool (*RIOFdClose)(RIO *io, int fd);
+typedef ut64 (*RIOFdSeek)(RIO *io, int fd, ut64 addr, int whence);
+typedef ut64 (*RIOFdSize)(RIO *io, int fd);
+typedef bool (*RIOFdResize)(RIO *io, int fd, ut64 newsize);
+typedef ut64 (*RIOP2V)(RIO *io, ut64 pa);
+typedef ut64 (*RIOV2P)(RIO *io, ut64 va);
+typedef int (*RIOFdRead)(RIO *io, int fd, ut8 *buf, int len);
+typedef int (*RIOFdWrite)(RIO *io, int fd, const ut8 *buf, int len);
+typedef int (*RIOFdReadAt)(RIO *io, int fd, ut64 addr, ut8 *buf, int len);
+typedef int (*RIOFdWriteAt)(RIO *io, int fd, ut64 addr, const ut8 *buf, int len);
+typedef bool (*RIOFdIsDbg)(RIO *io, int fd);
+typedef const char *(*RIOFdGetName)(RIO *io, int fd);
+typedef RList *(*RIOFdGetMap)(RIO *io, int fd);
+typedef bool (*RIOFdRemap)(RIO *io, int fd, ut64 addr);
+typedef bool (*RIOIsValidOff)(RIO *io, ut64 addr, int hasperm);
+typedef RIOMap *(*RIOMapGet)(RIO *io, ut64 addr);
+typedef RIOMap *(*RIOMapGetPaddr)(RIO *io, ut64 paddr);
+typedef bool (*RIOAddrIsMapped)(RIO *io, ut64 addr);
+typedef RIOMap *(*RIOMapAdd)(RIO *io, int fd, int flags, ut64 delta, ut64 addr, ut64 size);
 #if HAVE_PTRACE
-typedef long (*RIOPtraceFn) (RIO *io, r_ptrace_request_t request, pid_t pid, void *addr, r_ptrace_data_t data);
-typedef void *(*RIOPtraceFuncFn) (RIO *io, void *(*func)(void *), void *user);
+typedef long (*RIOPtraceFn)(RIO *io, r_ptrace_request_t request, pid_t pid, void *addr, r_ptrace_data_t data);
+typedef void *(*RIOPtraceFuncFn)(RIO *io, void *(*func)(void *), void *user);
 #endif
 
 typedef struct r_io_bind_t {
@@ -273,13 +273,13 @@ typedef struct r_io_bind_t {
 
 //map.c
 R_API RIOMap *r_io_map_new(RIO *io, int fd, int flags, ut64 delta, ut64 addr, ut64 size);
-R_API void r_io_map_init (RIO *io);
-R_API bool r_io_map_remap (RIO *io, ut32 id, ut64 addr);
-R_API bool r_io_map_remap_fd (RIO *io, int fd, ut64 addr);
+R_API void r_io_map_init(RIO *io);
+R_API bool r_io_map_remap(RIO *io, ut32 id, ut64 addr);
+R_API bool r_io_map_remap_fd(RIO *io, int fd, ut64 addr);
 R_API ut64 r_io_map_location(RIO *io, ut64 size);
-R_API bool r_io_map_exists (RIO *io, RIOMap *map);
-R_API bool r_io_map_exists_for_id (RIO *io, ut32 id);
-R_API RIOMap *r_io_map_resolve (RIO *io, ut32 id);
+R_API bool r_io_map_exists(RIO *io, RIOMap *map);
+R_API bool r_io_map_exists_for_id(RIO *io, ut32 id);
+R_API RIOMap *r_io_map_resolve(RIO *io, ut32 id);
 R_API RIOMap *r_io_map_add(RIO *io, int fd, int flags, ut64 delta, ut64 addr, ut64 size);
 // same as r_io_map_add but used when many maps need to be added. Call r_io_update when all maps have been added.
 R_API RIOMap *r_io_map_add_batch(RIO *io, int fd, int flags, ut64 delta, ut64 addr, ut64 size);
@@ -292,13 +292,13 @@ R_API void r_io_map_reset(RIO* io);
 R_API bool r_io_map_del(RIO *io, ut32 id);
 R_API bool r_io_map_del_for_fd(RIO *io, int fd);
 R_API bool r_io_map_depriorize(RIO* io, ut32 id);
-R_API bool r_io_map_priorize (RIO *io, ut32 id);
-R_API bool r_io_map_priorize_for_fd (RIO *io, int fd);
-R_API void r_io_map_cleanup (RIO *io);
-R_API void r_io_map_fini (RIO *io);
-R_API bool r_io_map_is_in_range (RIOMap *map, ut64 from, ut64 to);
-R_API void r_io_map_set_name (RIOMap *map, const char *name);
-R_API void r_io_map_del_name (RIOMap *map);
+R_API bool r_io_map_priorize(RIO *io, ut32 id);
+R_API bool r_io_map_priorize_for_fd(RIO *io, int fd);
+R_API void r_io_map_cleanup(RIO *io);
+R_API void r_io_map_fini(RIO *io);
+R_API bool r_io_map_is_in_range(RIOMap *map, ut64 from, ut64 to);
+R_API void r_io_map_set_name(RIOMap *map, const char *name);
+R_API void r_io_map_del_name(RIOMap *map);
 R_API RList* r_io_map_get_for_fd(RIO *io, int fd);
 R_API bool r_io_map_resize(RIO *io, ut32 id, ut64 newsize);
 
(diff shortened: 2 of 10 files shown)


test/unit/test_id_storage.c
```diff
@@ -0,0 +1,62 @@
...
+	r_id_storage_get_prev (ids, &_id);
+	r_id_storage_get_next (ids, &_id);
+	r_id_storage_free (ids);
+	
+	mu_assert_eq (id, _id, "r_id_storage_{next/prev} reversal 2");
+	mu_end;
+}
+
+static int all_tests(void) {
+	mu_run_test (test_r_id_storage_prev_next_eq_0);
```
ret2libc, line 54: Could you add something like https://github.com/radareorg/radare2/pull/17271#issuecomment-683700546 please?

trufae: Yes i can do that, the problem is that idpool doenst have such highest/lowest concept thing. this is implemented in io on top of the next/prev primitives. i can move that too

libr/util/idpool.c
```diff
@@ -163,6 +162,71 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
 	return storage->data[id];
 }
 
+R_API bool r_id_storage_get_lowest(RIDStorage *storage, ut32 *id) {
+	r_return_val_if_fail (storage, false);
+	int fd = 0;
+	if (!r_id_storage_get_prev (storage, (ut32 *)&fd) && fd == 0) {
```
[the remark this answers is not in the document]
condret: yep, 0 is a valid id. id_storage has no ids indicating errors. returnvalue indicates errors

test/db/io/dupfd
```diff
@@ -27,3 +27,37 @@ EXPECT=<<EOF
 1001
 EOF
 RUN
+
+NAME=dupfd hard
+FILE=-
+CMDS=<<EOF
+op
+o -
```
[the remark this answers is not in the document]
trufae: no. it's opening "-", which is the same as `r2 -`

condret:
thx for doing this, I just don't have time for finishing this pr atm

libr/util/idpool.c
```diff
@@ -163,6 +162,71 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+		return false;
+	}
+	if (!r_id_storage_get_next (storage, (ut32 *)&fd) && fd == 0) {
```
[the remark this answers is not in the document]
condret: it uses modulo arithmetics that is implemented in id_next and id_prev

libr/util/idpool.c
```diff
@@ -163,6 +162,71 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+	ut32 _id = *id;
+	const ut32 start_id = _id;
+	do {
+		_id = (_id + storage->size - 1) % storage->size;
+		if (_id == start_id) {
+			break;
+		}
+	} while (!storage->data[_id]);
+	*id = _id;
+	return (_id <= start_id); //true;
```
condret, line 227: this is wrong. should always return true

libr/util/idpool.c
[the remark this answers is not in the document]
trufae: this is used to get the highest and lowest fds, fd is set to 0 and then calls to get the next and the prev. to find the first and last of them. so in those cases fd == 0 means that there's no lower or higher fds, but also, there's no next or prev (otherwise that would be reported in the fd !=0)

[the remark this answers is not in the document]
trufae: im not 100% happy about this api design, but it works for me

libr/util/idpool.c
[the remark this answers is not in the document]
trufae: see my explanation above. the previous implementation was not working well and it was evenb more confusing.

libr/util/idpool.c
trufae: nope. then theres no way to know if we are in the latest or lowest fd. in your previous implementation rotate and next was doing the same for example. and it was never returning false unless there was no elements in the storage.

libr/util/idpool.c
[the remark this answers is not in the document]
condret: what you want to check is if the storage is empty, right?

libr/util/idpool.c
condret: yes, that's intended, so you can rotate forward, like in a ring-buffer. if you have ids 3,4,5 and do id_next for 5 the resulting id should be 3

libr/util/idpool.c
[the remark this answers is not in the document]
condret: id_next and id_prev do not require a valid id as parameter, except the id should be in the possible range of the idstorage.
if the id is not valid, but in that range, prev and next are intended to provide the next/previous valid id, as in a ringbuffer.

libr/util/idpool.c
trufae: it does that, but returns false, so you can deterine if the ring buffer have looped or not

condret: but the return value should indicate an error

libr/io/io_desc.c
```diff
@@ -92,11 +92,19 @@ R_API RIODesc *r_io_desc_get_prev(RIO *io, RIODesc *desc) {
...
+	int fd = r_io_fd_get_highest (io);
+	if (fd == -1) {
+		return NULL;
+	}
+	return r_io_desc_get (io, fd);
 }
 
 R_API RIODesc *r_io_desc_get_lowest(RIO *io) {
-	return r_io_desc_get (io, r_io_fd_get_lowest (io));
+	int fd = r_io_fd_get_lowest (io);
```
condret, line 103: I'm sure this is not needed as r_io_desc_get should return NULL if fd is not valid

libr/util/idpool.c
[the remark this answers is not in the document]
ret2libc: In any case, I don't think `fd == 0` is right. If there is no "lowest" you should get `false` from `r_id_storage_get_prev` and that is a good enough error condition to check.

libr/io/io_desc.c
ret2libc: well, it is always good to check for errors after calling a function. I'm very fine with having a `fd == -1` check after getting `fd`.

libr/util/idpool.c
```diff
@@ -163,6 +162,70 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+	r_return_val_if_fail (storage, false);
+	int fd = 0;
+	if (!r_id_storage_get_prev (storage, (ut32 *)&fd) && fd == 0) {
+		return false;
+	}
+	*id = fd;
+	return true;
+}
+
+// usually these shall not be used
```
ret2libc, line 188: remove these comments. We can't provide an API and say "bla bla you should not use it... this is wrong.. bla bla... there is no sorting but we provide prev/next". Either you provide the prev/next API and you expect developers to use it (and also you assume there is some sort of sorting in id-storage) or you don't provide an API.

libr/util/idpool.c
```diff
@@ -163,6 +162,70 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+// id-storage doesn't provide sorting things in specific order
+// there are only a very few reasonable usecases, such as cycling through RIODescs
+R_API bool r_id_storage_get_next(RIDStorage *storage, ut32 *idref) {
+	r_return_val_if_fail (id && storage && storage->data, false);
```
ret2libc, line 192:
```suggestion
	r_return_val_if_fail (idref && storage && storage->data, false);
```

libr/util/idpool.c
```diff
@@ -163,6 +162,70 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+	// it does not indicate a bug, if the passed id goes beyond the boundaries of the storage
+	// therefor no r_return_val_if_fail for this case
+	if (storage->size <= *id) {
```
ret2libc, line 195:
```suggestion
	if (storage->size <= *idref) {
```

libr/core/cmd_open.c
```diff
@@ -1209,6 +1218,45 @@ static bool desc_list_json_cb(void *user, void *data, ut32 id) {
...
+		next_fd = fd;
+		break;
+	case 'n':
+		next_fd = r_io_fd_get_next (core->io, cur_fd);
+		break;
+	case 'p':
+		next_fd = r_io_fd_get_prev (core->io, cur_fd);
+		break;
+	case 'r':
+		next_fd = r_io_fd_get_next (core->io, cur_fd);
```
ret2libc, line 1235: judging from these few lines, it looks like you are assuming that `r_io_fd_get_next` does not rotate by itself, and I'm perfectly fine with this. However, I think you should change `r_id_storage_get_next`/`prev` to actually not rotate. The rotation can be easily done as you did, by checking for `-1` and getting the lowest (or highest, if you are rotating backward).

ret2libc (changes requested):

libr/io/io_desc.c
trufae: can we make fds signed to handle errors with -1? right now there's an inconsistency between idstorage and iofds, first are ut32 , second are int.

libr/util/idpool.c
trufae: agree

libr/util/idpool.c
```diff
@@ -163,6 +162,65 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+		return false;
+	}
+	if (!r_id_storage_get_next (storage, (ut32 *)&fd) && fd == 0) {
+		return false;
+	}
+	*id = fd;
+	return true;
+}
+
+R_API bool r_id_storage_get_highest(RIDStorage *storage, ut32 *id) {
```
ret2libc, line 178: btw @condret , this really implies that the "highest" id is the one with a bigger value and that the "lowest" is the one with the smaller value. Really, all this just says that there is an ascending order. And that is also confirmed by `o`, which always shows stuff in ascending order.

libr/util/idpool.c
```diff
@@ -163,6 +162,65 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+	r_return_val_if_fail (storage, false);
+	int fd = 0;
+	if (!r_id_storage_get_prev (storage, (ut32 *)&fd) && fd == 0) {
+		return false;
+	}
+	*id = fd;
+	return true;
+}
+
+R_API bool r_id_storage_get_next(RIDStorage *storage, ut32 *idref) {
```
ret2libc, line 188: Can't this become a bit shorter?
```C
for (id = *idref + 1; id < storage->size && !storage->data[id]; id++);
if (id < storage->size) {
  *idref = id;
  return true;
}
return false;
```

libr/util/idpool.c
```diff
@@ -163,6 +162,65 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+		id = (id + 1) % storage->size;
+		if (id == start_id) {
+			break;
+		}
+	} while (!storage->data[id]);
+	*idref = id;
+	return (id >= start_id);
+}
+
+R_API bool r_id_storage_get_prev(RIDStorage *storage, ut32 *idref) {
```
ret2libc, line 206: Can't this become a bit shorter as well?
```C
if (*idref >= stoarege->size || *idref == 0) {
  return -1;
}
for (id = *idref - 1; id > 0 && !storage->data[id]; id--);
if (storage->data[id]) {
  *idref = id;
  return true;
}
return false;
```

libr/util/idpool.c
```diff
@@ -163,6 +162,65 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
 	return storage->data[id];
 }
 
+R_API bool r_id_storage_get_lowest(RIDStorage *storage, ut32 *id) {
```
ret2libc, line 165: What about:
```C
for (i = 0; i < storage->size && !storage-data[i]; i++);
*id = i;
return i < storage->size;
```

libr/util/idpool.c
ret2libc, line 178: What about:
```C
for (i = storage->size - 1; i > 0; i--);
*id = i;
return storage->data[i] != NULL;
```

ret2libc (commented):
Let me be honest... I think this whole RIDStorage thing is becoming more complicated than it should be. It is simply an array indexed by id. So, a natural sorting of this is just the ascending sorting by id. All those wraparounds and hypothetical discussions seems just unnecessary complications because "we" (@condret ) do not want to give any information at all to users of the API. What is wrong with letting users of this API assume that the data is sorted by id (as it already is, by the way)?

test/unit/test_id_storage.c
```diff
@@ -0,0 +1,107 @@
...
+	mu_assert (h, "get-lowest must return true");
+	mu_assert_eq (idw, id2list, "highest should be id2");
+
+	r_list_free (ids_list);
+	r_id_storage_free (ids);
+
+	mu_end;
+}
+
+static int all_tests(void) {
```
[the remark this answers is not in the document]
trufae: Im not in the mood to write more unit tests for an API that seems it needs changes still, but yes. will do

libr/util/idpool.c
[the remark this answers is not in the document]
trufae: yes i know this is a hack because i was trying to use that idstorage api, without changing it a lot. thats why i didnt liked the API at first and wanted to have it merged for testing with my PR. for me i was happy to get the functionality, but the api was duplicating the same logic in multiple placs.. i mean. iofd an riodesc are just frontends for idpool which end up just giving a different api usage,

libr/util/idpool.c
trufae: yes, the error is that there´s no such next or prev fd because you need to cicle back to the lower fds to get the next one and this is what its returned by reference.

libr/core/cmd_open.c
trufae: i dont think next/prev should assume rotation at all. for me fds are a list, if idpool uses a ring internally thats out of the concept of FDs, and also, fd =0 should be valid, which doesnt seems to be valid in the current implementation

my current implementation does not rotate. just return false and give u the next/prev in case it rotates

libr/util/idpool.c
trufae: agree

libr/util/idpool.c
trufae: yep

libr/include/r_io.h
```diff
@@ -146,28 +146,28 @@ typedef struct r_io_plugin_t {
...
-	int (*gettid)(RIODesc *desc);
-	bool (*getbase)(RIODesc *desc, ut64 *base);
-	bool (*resize)(RIO *io, RIODesc *fd, ut64 size);
-	int (*extend)(RIO *io, RIODesc *fd, ut64 size);
-	bool (*accept)(RIO *io, RIODesc *desc, int fd);
-	int (*create)(RIO *io, const char *file, int mode, int type);
-	bool (*check)(RIO *io, const char *, bool many);
+	RIODesc*(*open)(RIO *io, const char *, int perm, int mode);
+	RList* /*RIODesc* */(*open_many)(RIO *io, const char *, int perm, int mode);
+	int(*read)(RIO *io, RIODesc *fd, ut8 *buf, int count);
```
ret2libc, line 157: ???? they were alright as they were before.

libr/include/r_io.h
```diff
@@ -199,38 +199,38 @@ typedef struct r_io_desc_cache_t {
...
-typedef bool (*RIOFdIsDbg) (RIO *io, int fd);
-typedef const char *(*RIOFdGetName) (RIO *io, int fd);
-typedef RList *(*RIOFdGetMap) (RIO *io, int fd);
-typedef bool (*RIOFdRemap) (RIO *io, int fd, ut64 addr);
-typedef bool (*RIOIsValidOff) (RIO *io, ut64 addr, int hasperm);
-typedef RIOMap *(*RIOMapGet) (RIO *io, ut64 addr);
-typedef RIOMap *(*RIOMapGetPaddr) (RIO *io, ut64 paddr);
-typedef bool (*RIOAddrIsMapped) (RIO *io, ut64 addr);
-typedef RIOMap *(*RIOMapAdd) (RIO *io, int fd, int flags, ut64 delta, ut64 addr, ut64 size);
+typedef bool(*RIODescUse)(RIO *io, int fd);
```
ret2libc, line 202: same for these...

libr/util/idpool.c
```diff
@@ -163,6 +162,54 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+	return i < storage->size;
+}
+
+R_API bool r_id_storage_get_highest(RIDStorage *storage, ut32 *id) {
+	r_return_val_if_fail (storage, false);
+	size_t i = 0;
+	if (storage->size > 0) {
+		for (i = storage->size - 1; !storage->data[i] && i > 0; i--);
+		*id = i;
+		return storage->data[i];  // != NULL;
```
ret2libc, line 179: just put != NULL in the code instead of the comment... if it is not clear enough, it's worth putting it in the code, as it gets compiled.

libr/util/idpool.c
```diff
@@ -163,6 +162,54 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+	}
+	*id = i;
+	return false;
+}
+
+R_API bool r_id_storage_get_next(RIDStorage *storage, ut32 *idref) {
+	r_return_val_if_fail (idref && storage && storage->data, false);
+	ut32 id = *idref;
+	if (storage->size < 1 || id >= storage->size || id == 0) {
```
ret2libc, line 188: why return false if `id == 0`?

test/unit/test_id_storage.c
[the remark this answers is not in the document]
ret2libc: the API seems quite stable, just the internal implementation changed so far.

ret2libc (changes requested):

libr/include/r_io.h
trufae: that regex went too far

ret2libc: :D :D i guess so

libr/util/idpool.c
trufae: because id=0 is treated as an invalid id in other places. there are no tests for this, but its the case. and i dont think this is the place to discuss/address this. so i just put it there for consistency with current behaviour

test/unit/test_id_storage.c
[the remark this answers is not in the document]
trufae: there are a bunch of tests already i think

libr/util/idpool.c
condret: 0 is a valid id in fact

condret:
tbh, I don't think that wrapping around is really complicated. It's just simple modulo arithmetics.

I'd like to not have id_next and prev without having the wrapping. The original uses case, as i understood it was iterating over elements starting at an id somewhere in an id_storage. (imo )From the perspective of io, id_next should provide lowest id, if the highest id was passed to id_next, just so that io can do id_next until it gets to the starting point again without needing to care for what id_storage is doing below.

On the other side, one could argue, that it should behave similar to r_list, in which r_list_iter_get_next does not wrap around.

I think the stance, that RIDStorage should behave like RList is wrong. They are similar in some way, but also intended to satisfy different requirements:
- RLIst: get fast to next element
- RIDStorage: get element by numerical keyvalue fast and produce such unique keys

I don't see a necessity for RIDStorage to behave like RList here.

ret2libc:
> tbh, I don't think that wrapping around is really complicated. It's just simple modulo arithmetics.
> 
> I'd like to not have id_next and prev without having the wrapping. The original uses case, as i understood it was iterating over elements starting at an id somewhere in an id_storage. (imo )From the perspective of io, id_next should provide lowest id, if the highest id was passed to id_next, just so that io can do id_next until it gets to the starting point again without needing to care for what id_storage is doing below.
> 
> On the other side, one could argue, that it should behave similar to r_list, in which r_list_iter_get_next does not wrap around.
> 
> I think the stance, that RIDStorage should behave like RList is wrong. They are similar in some way, but also intended to satisfy different requirements:
> 
>     * RLIst: get fast to next element
> 
>     * RIDStorage: get element by numerical keyvalue fast and produce such unique keys
> 
> 
> I don't see a necessity for RIDStorage to behave like RList here.

What are the issues you see by having next/prev return NULL/false when reaching the first/last id? If I want to iterate all elements in the RIDstorage with no rotation it would be something like:
```C
ut32 id;
bool has_element = r_id_storage_get_lowest (ids, &id);
while (has_element) {
   // do something with `id`
   has_element = r_id_storage_get_next (ids, &id);
}
```
or, if the caller wants to actually rotate:
```C
ut32 first_id, id;
bool has_element = r_id_storage_get_lowest (ids, &id);
first_id = id;
while (has_element) {
   // do something with `id`
   has_element = r_id_storage_get_next (ids, &id) || r_id_storage_get_lowest (ids, &id);
   has_element &= id != first_id;
}
```

with the rotation, it would probably be something like (i'm not 100% sure, as I did not fully get how the API in #17271 would work):
```C
ut32 first_id = 0, id;
bool has_element = r_id_storage_get_lowest (ids, &id);
first_id = id;
while (has_element) {
    // do something with `id`
   r_id_storage_get_next (ids, &id); // useless to check the return value because according to that API it would return always true if there is at least one element
   has_element = id != first_id;
}
```
In this case you always need to keep track yourselves of the first element, because there is no way for the API to tell you "hey, you have looked at all the elements".

It is not just about behaving like RList, but in general most data structures have a start and an end. It's just the most common thing to do. Even with circular queue, where the implementation uses an array that wrap-arounds, the API still provides a consistent view of the data with a "first"/top element and a "last"/bottom element, no matter whether the "first" element in the queue is actually positioned in the first/second or last element of the array.

libr/io/io_fd.c
```diff
@@ -126,3 +126,39 @@ R_API int r_io_fd_get_current(RIO *io) {
...
+	if (!r_id_storage_get_prev (io->files, (ut32 *)&ret)) {
+		return -1;
+	}
+	return ret;
+}
+
+R_API int r_io_fd_get_highest(RIO *io) {
+	r_return_val_if_fail (io, -1);
+	int fd = 0;
+	if (!r_id_storage_get_highest (io->files, (ut32 *)&fd) && fd == 0) {
```
ret2libc, line 151:
```suggestion
	if (!r_id_storage_get_highest (io->files, (ut32 *)&fd)) {
```

libr/io/io_fd.c
```diff
@@ -126,3 +126,39 @@ R_API int r_io_fd_get_current(RIO *io) {
...
+		return -1;
+	}
+	return fd;
+}
+
+R_API int r_io_fd_get_lowest(RIO *io) {
+	r_return_val_if_fail (io, -1);
+	int fd = 0;
+	if (!r_id_storage_get_lowest (io->files, (ut32 *)&fd) && fd == 0) {
```
ret2libc, line 160:
```suggestion
	if (!r_id_storage_get_lowest (io->files, (ut32 *)&fd)) {
```

libr/util/idpool.c
ret2libc: @trufae Could you point to where id=0 is treated as invalid id?

ret2libc (changes requested):
Fix all the spaces issues in r_io.h and the opened comments please.

condret:
@ret2libc to be clear, I don't see this as a problem (yet). But I also don't see wrapping around as complicated.
You could do things like:

`const ut32 start_id = id;
do {
	/* some stuff */
while ((r_id_storage_next(st, &id)) && (id =! start_id));`

I don't see this as mandatory to have, but it would be a really cool api feature.

libr/io/io_fd.c
condret: I imagine this is about to check if io->files is empty

condret: @trufae if this is the intention here, we can add r_id_storage_is_empty

libr/io/io_desc.c
condret: they are signed in the context of io. negative values are used to indicate errors by io, not id_storage. id_storage used as io->files is initialized with parameters that ensure that no negative fd can be inserted

libr/util/idpool.c
```diff
@@ -163,6 +162,54 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
 	return storage->data[id];
 }
 
+R_API bool r_id_storage_get_lowest(RIDStorage *storage, ut32 *id) {
+	r_return_val_if_fail (storage, false);
+	size_t i;
```
condret, line 167: why size_t, ids are ut32 and i is used as id here

libr/util/idpool.c
```diff
@@ -163,6 +162,54 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
+	if (storage->size > 0) {
+		for (i = storage->size - 1; !storage->data[i] && i > 0; i--);
+		*id = i;
+		return storage->data[i];  // != NULL;
+	}
+	*id = i;
+	return false;
+}
+
+R_API bool r_id_storage_get_next(RIDStorage *storage, ut32 *idref) {
```
condret, line 185: can we rename idref as id. Id is used as parameter in other id_storage functions. don't break that

ret2libc:
> @ret2libc to be clear, I don't see this as a problem (yet). But I also don't see wrapping around as complicated.
> You could do things like:
> 
> ```
> do {
> 	/* some stuff */
> while ((r_id_storage_next(st, &id)) && (id =! start_id));
> ```
> 
> I don't see this as mandatory to have, but it would be a really cool api feature.

Yes it would be an additional cool API, not the core essential one though. All other APIs for data structure have a begin/end approach, not an "infinite" walk.
Also, your example is very incomplete.
```C
bool has_element = r_id_storage_next (st, &start_id);
if (has_element) {
  do {
     /* some stuff */
  while ((r_id_storage_next(st, &id)) && (id =! start_id));
}
```

for a regular loop this seems quite complicated to me.

libr/util/idpool.c
trufae: fixed

libr/util/idpool.c
trufae: no. this way we avoid having to dereference a dword many unnecessary times inside the loops. the value is written back at the end.

libr/io/io_fd.c
trufae: done

libr/io/io_fd.c
trufae: yeah, thats not really needed

ret2libc:
@trufae see https://github.com/radareorg/radare2/pull/17652#discussion_r496576948

ret2libc (changes requested):
Just to prevent merging.

libr/util/idpool.c
trufae: i removed that assumption. see r_oids_to_rear for example in uses id=0 to detect the loop has failed to find a match, cant find that ref anywhere else. but still i think i need to do further reviews and cleanups on all this API. but i dont think this PR is the place to do it

libr/core/cmd_open.c
```diff
@@ -1209,6 +1218,44 @@ static bool desc_list_json_cb(void *user, void *data, ut32 id) {
...
+		if (next_fd == -1) {
+			next_fd = r_io_fd_get_lowest (core->io);
+		}
+		break;
+	}
+	if (cur_fd != next_fd) {
```
ret2libc, line 1241: this check is already performed in the next line

libr/io/io_desc.c
```diff
@@ -1,10 +1,11 @@
-/* radare2 - LGPL - Copyright 2017-2019 - condret, pancake, alvaro */
+/* radare2 - LGPL - Copyright 2017-2020 - condret, pancake, alvaro */
 
 #include <r_io.h>
 #include <sdb.h>
 #include <string.h>
 
-//shall be used by plugins for creating descs
+// shall be used by plugins for creating descs
+// calling this function many times result in always fd=0
```
ret2libc, line 8: why?

libr/io/io_fd.c
ret2libc: I still see `fd == 0` @trufae

libr/util/idpool.c
```diff
@@ -163,6 +162,67 @@ R_API void* r_id_storage_get(RIDStorage* storage, ut32 id) {
...
 
+R_API bool r_id_storage_get_lowest(RIDStorage *storage, ut32 *id) {
+	r_return_val_if_fail (storage, false);
+	ut32 i;
+	for (i = 0; i < storage->size && !storage->data[i]; i++);
+	*id = i;
+	return i < storage->size;
+}
+
+R_API bool r_id_storage_is_empty(RIDStorage *storage) {
```
ret2libc, line 173: this is not used right now, is it?

libr/io/io_desc.c
trufae: i think we can discuss the whole rio api in a separate place, not in this PR, i just wanted a single command without having to create new APIs or mess with low level discussions of APIs, which i'm totally happy with them, but i dont think it should block the PR because the discussion goes beyond the actual feature, the new apis and the already defined APIs.

libr/io/io_desc.c
trufae: just read the code, is quite obvious, in short:

* fd32 is only updated when io->files != NULL
* this function doersnt call or set anything to change io->files
* idpool_grab_id is never called

but again, this is far from the scope of this PR and the discussion can go on and on for years, i dont think it should be addressed in here

libr/util/idpool.c
trufae: removed. also, the fd == 0 i was refering to the ridstorage code, i have fixed this remaining one in iofd .

libr/core/cmd_open.c
trufae: i'm not sure if i follow you. i dont see this check anywhere

trufae:
It's only red because of big sur's xcode. another review?

test/unit/test_id_storage.c
```diff
@@ -0,0 +1,187 @@
...
+	mu_assert ("id1", id1 == 2);
+	r = r_id_storage_add (ids, "ccc", &id2);
+	mu_assert ("id2", r);
+	mu_assert ("id2", id2 == 3);
+	ut32 id = id1;
+	r_id_storage_get_prev (ids, &id);
+	r_id_storage_get_next (ids, &id);
+	r_id_storage_free (ids);
+	
+	mu_assert_eq (id, id1, "r_id_storage_{next/prev} reversal 0");
```
ret2libc, line 45: what is the difference between this test and the ones at https://github.com/radareorg/radare2/pull/17652/files#diff-8dacb57a2e0ae9396828e3bdd407a6b77c779b6db28117f84c904eab2613ebc0R61-R68 and at https://github.com/radareorg/radare2/pull/17652/files#diff-8dacb57a2e0ae9396828e3bdd407a6b77c779b6db28117f84c904eab2613ebc0R84-R90 

It seems to me they are testing exactly the same thing.

test/unit/test_id_storage.c
```diff
@@ -0,0 +1,187 @@
...
+	r_id_storage_free (ids);
+	mu_assert_eq (id, id1, "r_id_storage_{next/prev} reversal 1");
+	mu_end;
+}
+
+static bool test_r_id_storage_empty(void) {
+	// test if next reverts prev (modulo wrap 2)
+	ut32 id, _id;
+	RIDStorage *ids = r_id_storage_new (0, 15);
+	bool r = r_id_storage_get_prev (ids, &_id);
```
ret2libc, line 98: `_id` should probably be initialized.

ret2libc: It doesn't really matter in this case because it won't probably be used since `ids` is empty, but..

test/unit/test_id_storage.c
```diff
@@ -0,0 +1,187 @@
...
+	r_id_storage_add (ids, "aaa", &_id);
+	r_id_storage_add (ids, "ccc", &_id);
+	r_id_storage_add (ids, "bbb", &id);
+	_id = id;
+	r_id_storage_get_prev (ids, &_id);
+	r_id_storage_get_next (ids, &_id);
+	r_id_storage_free (ids);
+	
+	mu_assert_eq (id, _id, "r_id_storage_{next/prev} reversal 2");
+	mu_end;
```
ret2libc, line 124: this is again like all the other tests i've mentioned before.. we are always testing the exact same code paths, I think.

libr/core/cmd_open.c
trufae: omg im blind. fixed

libr/io/io_desc.c
trufae: i was speaking from the unit-test pov, because this io->files chk seems wrong to me. but as i said, i think this should be discussed and addressed in a different place. not in this pr

test/unit/test_id_storage.c
trufae: im testing the lower limit is 3 and not 0. so maybe the same code path but different behaviour

trufae:
@ret2libc ping

XVilka:
Unit test build failure:
```
clang test_id_storage.c -o bin/test_id_storage -I/home/build/include/libr -I/home/build/include/libr/sdb  -g -L/home/build/lib -lr_core -lssl -lr_config -lr_debug -lr_bin -lr_anal -lr_bp -lr_egg -lr_asm -lr_lang -lr_parse -lr_flag -lr_reg -lr_search -lr_syscall -lr_fs -lr_io -lr_socket -lr_cons -lr_magic -lr_crypto -lr_hash -lr_util 
test_id_storage.c:88:42: error: too few arguments provided to function-like macro invocation
        mu_assert_true ("get_prev(1) must exist");
                                                ^
./minunit.h:69:9: note: macro 'mu_assert_true' defined here
#define mu_assert_true(actual, message) do { \
        ^
test_id_storage.c:88:2: error: use of undeclared identifier 'mu_assert_true'
        mu_assert_true ("get_prev(1) must exist");
        ^
2 errors generated.
```
https://builds.sr.ht/~xvilka/job/330578#task-test-750

libr/io/io_desc.c
ret2libc: I don't see that as a problem in unit tests either.

Anyway, I agree to not discuss this in this PR, but you can't add *random* comments and then say "this is not the place to discuss it". If it's not the place to discuss it, just don't add the comment in the first place.

trufae:
@ret2libc its green (except for mac), please check it out again. i have removed the comments

ret2libc (approved):

PR #17659, opened 2020-09-16
Make codecov/project informational on master as well

ret2libc:
I don't think we want to see master red because we coverage is not good. It is definitely something we want to look at from time to time, but seeing master red just because of codecov does not seem right. Hopefully this will make it green.


ret2libc:
I have forced the merge to master, as that change won't do anything on PR.

PR #17660, opened 2020-09-16
Correctly check whether esil_str is available

ret2libc:
Fix issues like https://github.com/radareorg/radare2/runs/1122620100?check_suite_focus=true#step:13:1961 .

This hopefully makes asan build green again.

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -1807,7 +1807,7 @@ static void core_anal_bytes(RCore *core, const ut8 *buf, int len, int nops, int 
 				free (opname);
 			}
 		} else if (fmt == 'e') {
-			if (*esilstr) {
+			if (R_STR_ISNOTEMPTY (esilstr)) {
 				if (use_color) {
 					r_cons_printf ("%s0x%" PFMT64x Color_RESET " %s\n", color, core->offset + idx, esilstr);
 				} else {
@@ -1929,7 +1929,7 @@ static void core_anal_bytes(RCore *core, const ut8 *buf, int len, int nops, int 
 				}
 
 			}
-			if (esilstr) {
+			if (R_STR_ISNOTEMPTY (esilstr)) {
 				int ec = esil_cost (core, addr, esilstr);
 				pj_ki (pj, "esilcost", ec);
 			}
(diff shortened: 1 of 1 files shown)


ret2libc:
There are 2 failing tests, because in both cases esil is not implemented. Do we want to show `esilcost = 0` anyway in those cases? cc @XVilka @trufae 
```
[XX] db/anal/tms320.c64x_32 320c64 OPEX
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e asm.arch=tms320
e asm.cpu=c64x
wx 0041bd10
aoj~{}
' malloc://512280
-- stdout
@@ -21,7 +21,6 @@
     "bytes": "0041bd10",
     "size": 4,
     "type": "jmp",
-    "esilcost": 0,
     "scale": 0,
     "refptr": 0,
     "cycles": 0,
```
```
[XX] db/anal/arm64 ao mte irg addg
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'ao@0x100007f10
?e --
ao@0x100007f14
' bins/mach0/hello-mte
-- stdout
@@ -1,6 +1,5 @@
 address: 0x100007f10
 opcode: irg x8, sp, x8
-esilcost: 0
 disasm: irg x8, sp, x8
 pseudo: irg x8,sp,x8 
 mnemonic: irg
@@ -17,7 +16,6 @@
 --
 address: 0x100007f14
 opcode: addg x9, x8, 0x20, 0x0
-esilcost: 0
 disasm: addg x9, x8, 0x20, 0x0
 pseudo: addg x9,x8,0x20, 0x0 
 mnemonic: addg
```

trufae:
I was thinking in having +1 for each word parsed by the esil engine. Not just considering memory accesses as costful things. If the esil expr is not implemented i think its fine to have this as 0

ret2libc:
> I was thinking in having +1 for each word parsed by the esil engine. Not just considering memory accesses as costful things. If the esil expr is not implemented i think its fine to have this as 0

So you think it's better to see `0` instead of nothing at all?

trufae:
json objects should be consistent imho. esilstr should be there even if empty i think. so yes. for me 0 is fine

kazarmy:
Does `nop` have an esil implementation? Is its esil cost 1 or 0?

trufae:
nop should be 0 esil cost imho. the implementation is just a comma. this is different from no esil expression and results in the same: nothing

trufae (approved):

ret2libc:
> json objects should be consistent imho. esilstr should be there even if empty i think. so yes. for me 0 is fine

Ok good I will fix the PR a bit to make this happen.

kazarmy:
> I was thinking in having +1 for each word parsed by the esil engine.

ok

kazarmy (approved):

PR #17663, opened 2020-09-17
Automatically group sub-commands ##newshell

ret2libc:

**Detailed description**

Improve a bit the help as shown by newshell. In particular, this groups sub-commands automatically so, when you do `w?` you will see `wv[1248]` because `wv` is a command, but you also have `wv1`, `wv2`, etc. If there are too many sub-commands, you see `wv[?]`. Also, if `wv` by itself is not a command, you would see `wv<1248>`, to make sure a user can tell the difference between a command that can work by itself and one that cannot.

Below you can see some screens of the current state. (Please note that `w<1248><+->` is not automatically generated :(, but I let commands define their own list of options, when that's required in some cases).
image
image
image
image
image


libr/core/cmd_api.c
```diff
@@ -1162,6 +1235,19 @@ R_API RCmdDesc *r_cmd_desc_parent(RCmdDesc *cd) {
 	return cd->parent;
 }
 
+R_API bool r_cmd_desc_has_handler(RCmdDesc *cd) {
+	r_return_val_if_fail (cd, NULL);
```
trufae, line 1239: false not null

libr/core/cmd_api.c
```diff
@@ -1162,6 +1235,19 @@ R_API RCmdDesc *r_cmd_desc_parent(RCmdDesc *cd) {
 	return cd->parent;
 }
 
+R_API bool r_cmd_desc_has_handler(RCmdDesc *cd) {
```
trufae, line 1238: this function only can return true if the .cb is set on argv mode, all the other cases is false, can you simplify this logic to just handle this?

libr/core/cmd_api.c
```diff
@@ -418,29 +420,95 @@ static void fill_usage_strbuf(RStrBuf *sb, RCmdDesc *cd, bool use_color) {
...
+static bool show_group_args(RCmdDesc *cd, RCmdDesc *parent) {
+	return !show_args (cd, parent) && cd->help->group_args_str;
+}
+
+static size_t calc_padding_len(RCmdDesc *cd, RCmdDesc *parent) {
+	size_t name_len = strlen (cd->name);
+	size_t args_len = 0;
+	size_t children_length = 0;
+	if (show_children_shortcut (cd, parent)) {
+		char *children_s = children_chars (cd);
```
trufae, line 471: check for null

libr/core/cmd_api.c
```diff
@@ -418,29 +420,95 @@ static void fill_usage_strbuf(RStrBuf *sb, RCmdDesc *cd, bool use_color) {
...
+		   *pal_opt_color = use_color? cons->context->pal.reset: "",
 		   *pal_help_color = use_color? cons->context->pal.help: "",
 		   *pal_input_color = use_color? cons->context->pal.input: "",
 		   *pal_reset = use_color? cons->context->pal.reset: "";
 
-	r_strbuf_appendf (sb, "| %s%s%s%s %*s%s# %s%s\n", pal_input_color, cd->name,
-		pal_args_color, cd_args_str, padding, "", pal_help_color, cd_summary, pal_reset);
+	r_strbuf_appendf (sb, "| %s%s", pal_input_color, cd->name);
+	if (show_children_shortcut (cd, parent)) {
+		char *children_s = children_chars (cd);
```
trufae, line 502: check for null

trufae (changes requested):

libr/core/cmd_api.c
ret2libc: No there is also the old input case that can return true

XVilka (approved):
Good idea.

test/unit/test_cmd.c
```diff
@@ -288,8 +288,8 @@ bool test_cmd_help(void) {
 	r_cmd_desc_oldinput_new (cmd, p_cd, "px", px_handler, &px_help);
 
 	const char *p_help_exp = "Usage: p-usage   # p summary\n"
-		"| pd <num>                 # pd summary\n"
-		"| px <verylongarg_str_num> # px summary\n";
+		"| pd <num>                    # pd summary\n"
+		"| px[?] <verylongarg_str_num> # px summary\n";
```
trufae, line 292: can this verylongarg string be shorter?

libr/core/cmd_api.c
```diff
@@ -7,6 +7,8 @@
 #include <r_cmd.h>
 #include <r_util.h>
 
+#define MAX_CHILDREN_SHOW 7
```
trufae, line 10: why seven? can you add a comment explaining the use of this?

trufae (changes requested):

test/unit/test_cmd.c
ret2libc: what do you mean? this is a test to make sure the summaries (the text on the right) are aligned. What's the problem with having a long string in the tests??

libr/core/cmd_api.c
ret2libc: Random number... I had to set a limit somewhere. I will add a comment.

trufae (approved):

test/unit/test_cmd.c
trufae: oh nvmits just a unit test

trufae:
make it grin

PR #17665, opened 2020-09-17
x86 cmp/sub: Properly generate esil for bitsize of 64 ##anal

kazarmy:

**Detailed description**

This pr fixes the following (https://github.com/radareorg/radare2/runs/1128444655#step:15:67):

bitsize_0_error

**Test plan**

The error above should not appear in the asan fuzz test output. All other builds are green.

**Closing issues**

...

### DIFF
--- a/libr/anal/p/anal_x86_cs.c
+++ b/libr/anal/p/anal_x86_cs.c
@@ -986,7 +986,7 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 			} else {
 				esilprintf (op,
 					"%s,%s,==,$z,zf,:=,%d,$b,cf,:=,$p,pf,:=,%d,$s,sf,:=,%s,0x%"PFMT64x",-,!,%d,$o,^,of,:=,3,$b,af,:=",
-					src, dst, bitsize, bitsize - 1, src, (1ULL << (bitsize - 1)), bitsize - 1);
+					src, dst, bitsize, bitsize - 1, src, bitsize ? 1ULL << (bitsize - 1) : 0, bitsize - 1);
 			}
 		}
 		break;
@@ -1484,7 +1484,7 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 			// We use $b rather than $c here as the carry flag really
 			// represents a "borrow"
 			esilprintf (op, "%s,%s,%s,0x%"PFMT64x",-,!,%d,$o,^,of,:=,%d,$s,sf,:=,$z,zf,:=,$p,pf,:=,%d,$b,cf,:=,3,$b,af,:=",
-				src, dst, src, (1ULL << (bitsize - 1)), bitsize - 1, bitsize - 1, bitsize);
+				src, dst, src, bitsize ? 1ULL << (bitsize - 1) : 0, bitsize - 1, bitsize - 1, bitsize);
 		}
 		break;
 	case X86_INS_SBB:
--- a/libr/anal/p/anal_x86_cs.c
+++ b/libr/anal/p/anal_x86_cs.c
@@ -976,7 +976,7 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 			src = getarg (&gop, 1, 0, NULL, SRC_AR, NULL);
 			dst = getarg (&gop, 0, 0, NULL, DST_AR, &bitsize);
 
-			if (bitsize > 63) {
+			if (bitsize > 64) {
 				bitsize = 0;
 			}
 
@@ -1476,7 +1476,7 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 			src = getarg (&gop, 1, 0, NULL, SRC_AR, NULL);
 			dst = getarg (&gop, 0, 1, "-", DST_AR, &bitsize);
 
-			if (bitsize > 63) {
+			if (bitsize > 64) {
 				bitsize = 0;
 			}
 
--- a/libr/anal/p/anal_x86_cs.c
+++ b/libr/anal/p/anal_x86_cs.c
@@ -976,7 +976,7 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 			src = getarg (&gop, 1, 0, NULL, SRC_AR, NULL);
 			dst = getarg (&gop, 0, 0, NULL, DST_AR, &bitsize);
 
-			if (bitsize > 64) {
+			if (!bitsize || bitsize > 64) {
 				break;
 			}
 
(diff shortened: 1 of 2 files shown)


trufae:
That was added to avoid ub on >63 shifts spotted by asan. Are you testing this too?

trufae:
Imho this is wrong:

```c
$ cat a.c
#include <stdint.h>

int main ()
{
	uint64_t a = 1LL << 64;
	uint64_t b = 1LL << 63;
	return 0;
}
$ gcc a.c
a.c:5:19: warning: shift count >= width of type [-Wshift-count-overflow]
        uint64_t a = 1LL << 64;
                         ^  ~~
1 warning generated.
$
```

kazarmy:
Ok getting the warning

kazarmy:
But both
https://github.com/radareorg/radare2/blob/3d9ce08c19f0144855c2510d7990148f3b5fd562/libr/anal/p/anal_x86_cs.c#L989 and
https://github.com/radareorg/radare2/blob/3d9ce08c19f0144855c2510d7990148f3b5fd562/libr/anal/p/anal_x86_cs.c#L1487 use
```c
(1ULL << (bitsize - 1))
```
so a `bitsize` of 64 should be fine, and there's nothing regarding this on the [asan build](https://github.com/radareorg/radare2/pull/17665/checks?check_run_id=1128789452) too.

kazarmy:
> Ok getting the warning

When I said this, I meant I got the warning for the a.c example code, not for r2 code.

trufae:
Uhm, but still if we know the shift operation will result in 0, we can just return 0 without doing the actual shift. I dont know why shifting 64 bits left in a 64 bit value may result in UB, the only reason is in case the variable is signed. But its not in this case.

kazarmy:
Emm the shift operation doesn't result in 0, because if `bitsize` is 64

```c
(1ULL << (bitsize - 1)) == (1ULL << (64 - 1))
                        == (1ULL << 63)
                        == 0x8000000000000000
```

And it's highly unlikely that the intention of the shift is to zero out the value, since there are easier ways of doing that.

kazarmy:
> Uhm, but still if we know the shift operation will result in 0, we can just return 0 without doing the actual shift. I dont know why shifting 64 bits left in a 64 bit value may result in UB, the only reason is in case the variable is signed. But its not in this case.

If the code attempts to shift 64 bits left in a 64 bit value, then 4edb84caa7bd096676da4fcb7dcb5243cc11bbea will substitute 0 for the shift.

XVilka:
It's not ASAN but UBSAN, because shift to the type size is the undefined behavior.

libr/anal/p/anal_x86_cs.c
```diff
@@ -986,7 +986,7 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 			} else {
 				esilprintf (op,
 					"%s,%s,==,$z,zf,:=,%d,$b,cf,:=,$p,pf,:=,%d,$s,sf,:=,%s,0x%"PFMT64x",-,!,%d,$o,^,of,:=,3,$b,af,:=",
-					src, dst, bitsize, bitsize - 1, src, (1ULL << (bitsize - 1)), bitsize - 1);
+					src, dst, bitsize, bitsize - 1, src, bitsize ? 1ULL << (bitsize - 1) : 0, bitsize - 1);
```
ret2libc, line 989: `bitsize-1` is used also as the last argument and it would be 0xfffffff if bitsize is 0. I'm not sure what this code does, but this seems wrong.

libr/anal/p/anal_x86_cs.c
```diff
@@ -1476,15 +1476,15 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
...
+			if (bitsize > 64) {
 				bitsize = 0;
 			}
 
 			// Set OF, SF, ZF, AF, PF, and CF flags.
 			// We use $b rather than $c here as the carry flag really
 			// represents a "borrow"
 			esilprintf (op, "%s,%s,%s,0x%"PFMT64x",-,!,%d,$o,^,of,:=,%d,$s,sf,:=,$z,zf,:=,$p,pf,:=,%d,$b,cf,:=,3,$b,af,:=",
-				src, dst, src, (1ULL << (bitsize - 1)), bitsize - 1, bitsize - 1, bitsize);
+				src, dst, src, bitsize ? 1ULL << (bitsize - 1) : 0, bitsize - 1, bitsize - 1, bitsize);
```
ret2libc, line 1487: same here.

ret2libc (commented):
Also, `bitsize` is `ut32` but then it seems to be printed as `st32`. Maybe it should be `st32` in the first place?

kazarmy:
> Also, `bitsize` is `ut32` but then it seems to be printed as `st32`. Maybe it should be `st32` in the first place?

Doesn't make sense to have a `bitsize` that's negative.

libr/anal/p/anal_x86_cs.c
kazarmy: the original esil apparently worked as expected despite its wrongness, but i guess it's better to have no esil at all if `bitsize` is invalid (44eef896a1f0c3da2e626185332cab983a97c074) since it becomes clearer that something needs to be done

libr/anal/p/anal_x86_cs.c
kazarmy: see https://github.com/radareorg/radare2/pull/17665#discussion_r493507451

ret2libc:
> > Also, `bitsize` is `ut32` but then it seems to be printed as `st32`. Maybe it should be `st32` in the first place?
> 
> Doesn't make sense to have a `bitsize` that's negative.

Then bitsize should be printed as `%u` instead of `%d` and there shall never be `bitsize - 1` without first checking if bitsize is already 0.

libr/anal/p/anal_x86_cs.c
kazarmy: `bitsize` invalid here means `bitsize` > 64

kazarmy:
> Then bitsize should be printed as %u instead of %d

ok 45ed826566f2f9f0a27fdc77ee9cefd0945144ef

> there shall never be bitsize - 1 without first checking if bitsize is already 0.

The code aborts if bitsize is 0, otherwise the esil makes even less sense than before.

ret2libc:
> The code aborts if bitsize is 0, otherwise the esil makes even less sense than before.

where?

kazarmy:
https://github.com/radareorg/radare2/pull/17665/files#diff-7f6f82199b572cdbe767aa7a1ba9b436R979-R980
https://github.com/radareorg/radare2/pull/17665/files#diff-7f6f82199b572cdbe767aa7a1ba9b436R1479-R1480

Sorry, i should have said "The pr now aborts if ..."

ret2libc:
> https://github.com/radareorg/radare2/pull/17665/files#diff-7f6f82199b572cdbe767aa7a1ba9b436R979-R980
> https://github.com/radareorg/radare2/pull/17665/files#diff-7f6f82199b572cdbe767aa7a1ba9b436R1479-R1480
> 
> Sorry, i should have said "The pr now aborts if ..."

Am I missing anything or that just breaks only if bitsize > 64, but not if bitsize == 0?

kazarmy:
Well bitsize can't be 0 there but you do have a point, therefore dc42f73cd036297192422d28e1c7967a6807c06a

ret2libc (approved):
Thanks for the changes, LGTM!

trufae (approved):

PR #17668, opened 2020-09-17
Allow spaces for some commands like <cmd> @ <arg-with-spaces>

### LINKED ISSUE #17666: newshell: support <cmd> @ <arg-with-spaces>
It was reported by some that having `pd 10 @ entry0 + 10` not working when `cfg.newshell=true` is annoying. This bug is about adding support for this, so `pd 10 @ entry0 + 10` will be the same as `pd 10 @ entry0+10`.

Spaces shall be accepted only in few cases,  for example after `@` , `@!` and maybe few others.
Stuff like `@@=` already accept spaces of course, because when you do `pd 1 @@=1 2 3`, `1`, `2`, and `3` are correctly considered by the parser as separate arguments.

**Describe alternatives you've considered**
Just deny the use of spaces after `@`, but this may cause issues for users switching to newshell, who are used to the lazy/incosistent checks of oldshell.

**Additional context**
The grammar used for newshell splits arguments by spaces (and by few other special chars). For example,
```
o /usr/bin/cat 0x1000 rx

---

(commands [0, 0] - [1, 0]
  (arged_command [0, 0] - [0, 24]
    command: (cmd_identifier [0, 0] - [0, 1])
    args: (args [0, 2] - [0, 24]
      (arg [0, 2] - [0, 14]
        (arg_identifier [0, 2] - [0, 14]))
      (arg [0, 15] - [0, 21]
        (arg_identifier [0, 15] - [0, 21]))
      (arg [0, 22] - [0, 24]
        (arg_identifier [0, 22] - [0, 24])))))
```

In the grammar we have the following definition:
```
tmp_seek_command: $ => seq($._simple_command, '@', $.arg),
```
As you can see, only one argument is expected after `@`. For consistency, the `arg` accepted by `tmp_seek_command`, is the same `arg` that is used by `arged_command` and all other sorts of commands. Probably we can change this a little bit so that `tmp_seek_command`(and maybe few others) accept also spaces. This would mean that things like `pd 10 @ entry0 test1 test2` would be accepted by the parser and the handler will need to show the proper error message in this case.

ret2libc:

**Detailed description**

More than one has complained/was surprised about newshell not supporting things like `pd 10 @ entry0 + 3` (note the spaces are important. Without spaces in `entry0 + 3` everything would work well even in newshell). This PR adds support for spaces in some cases, for example in case like `<cmd> @ <arg>` and few others.

However, the reason why I'm not a big fan of this is that then we start messing with the parser and it won't be consistent any more. For example, take `/x @{<from> <to>}`. Now, in this case I can't use spaces anymore in `<from>`, because otherwise it will be impossible to distinguish `<from>` and `<to>` (e.g. `/x @{10 + 3 - 2}` what is `<from>`? what is `<to>`?). This is already not consistent, because in one case spaces are accepted, in another they cannot be.

Apart from that, by accepting this PR, we also mess up a bit with argv-style concept. For example, `s` handler can be thought as accepting one single argument, which would be the string representing the argument  to seek to. However, if users now want to do `s entry0 + 10` (with spaces), the `s` handler will be called with 3 arguments: `entry0`, `+`, and `10`. To solve this, each handler would need to add code to get just one string from `["entry0", "+", "10"]`.

So, before going this route, I think it is important to think about the implications. If `entry + 10` is accepted in one place, it should also be accepted in most other places, which may imply additional code/cost(in terms of cleanliness) in command handlers.

**Test plan**

Try commands like `pd 10 @ something + something` or `pd 10 @o:3+2` and `pd 10 @i:1 + 1`.

**Closing issues**

Closes #17666


trufae (approved):

thestr4ng3r (approved):

radare (approved):

kazarmy:
I don't see what is wrong with `/x @{<expression> <expression>}` instead of `/x @{<numeral-or-operator> <numeral-or-operator>}`.

ret2libc:
> I don't see what is wrong with `/x @{<expression> <expression>}` instead of `/x @{<numeral-or-operator> <numeral-or-operator>}`.

@kazarmy right now the grammar for that operator is: `/x @{<arg> <arg>}`, where `<arg>` does not accept spaces. What I'm saying is that by allowing users to do `pd 10 @ entry0 + 10` (with spaces), they may expect to be able to use spaces everywhere, but that can't work for a command like `/x @{<arg> <arg>}`. The reason it can't work is that if you have things like `/x @{30 + 10 - 3}` you have no idea what is the first expression and what is the second one. Is it clearer?

thestr4ng3r:
> Apart from that, by accepting this PR, we also mess up a bit with argv-style concept. For example, `s` handler can be thought as accepting one single argument, which would be the string representing the argument to seek to. However, if users now want to do `s entry0 + 10` (with spaces), the `s` handler will be called with 3 arguments: `entry0`, `+`, and `10`. To solve this, each handler would need to add code to get just one string from `["entry0", "+", "10"]`.
> 
> So, before going this route, I think it is important to think about the implications. If `entry + 10` is accepted in one place, it should also be accepted in most other places, which may imply additional code/cost(in terms of cleanliness) in command handlers.

I think it makes sense to, whenever an argument is an expression, to have that one as the last one (whenever possible) with index i and in the handler use the concatenation of all args >=i with spaces for this argument. This will indeed need extra code in each handler, but I think concatenating all args >=i can be easily put in a helper function and then it will also be consistent across all commands that use it.

> However, the reason why I'm not a big fan of this is that then we start messing with the parser and it won't be consistent any more. For example, take `/x @{<from> <to>}`. Now, in this case I can't use spaces anymore in `<from>`, because otherwise it will be impossible to distinguish `<from>` and `<to>` (e.g. `/x @{10 + 3 - 2}` what is `<from>`? what is `<to>`?). This is already not consistent, because in one case spaces are accepted, in another they cannot be.

In the case of `@{10 + 3 - 2}` you would then interpret it as `["10", "+ 3 - 2"]`, or alternatively output an error and enforce quotes.

Overall I think that the r2 shell is mainly one that is meant to be used by humans so supporting spaces there is quite important imo for convenience. If it is done as described above, I think the behavior should still be simple enough to understand and machine-writable.

thestr4ng3r (approved):

kazarmy:
> ... and enforce quotes.

Tcl-style braces/brackets/parentheses should work too I suppose ... https://riptutorial.com/tcl/example/11461/the-problems-with-unbraced-expressions

kazarmy (approved):

trufae (approved):

XVilka (approved):

ret2libc:
@thestr4ng3r it could be a nice idea to let the last argument be an expression with spaces. In the handler for `w` command I've already used `r_str_array_join` to merge multiple arguments, see https://github.com/radareorg/radare2/blob/master/libr/core/cmd_write.c#L1424 .

PR #17672, opened 2020-09-18
Prep work to allow compiling and loading calling convention in extras

zamaudio:

**Detailed description**
This PR adds ability for radare2-extras to compile calling convention sdbs, and have them auto loaded when installed into R2_HOME_SDB_FCNSIGN directory.
Also, previously, there was a bug preventing r_file_exists() from statting $HOME because it needs abspath to expand tilde.
This is now fixed.

**Test plan**
Once this is merged, I can do a PR to r2e to fix the calling convention of atombios.


ret2libc:
@zamaudio could you tell exactly what you want to do, at the end, that is not working now?

zamaudio:
> @zamaudio could you tell exactly what you want to do, at the end, that is not working now?

I am trying to push changes to radare2-extras for atombios to complete the port, but I keep running into issues:
1) There is currently no Github CI for r2e - I am fixing that in r2e
2) There is a bug that prevents home directory files from being tested for existence so it is blocking me from loading calling conventions from ~/.local/share/radare2 - I am fixing that in this changeset
3) There is no sdb binary installed with r2 so r2e cannot compile .sdb files such as a calling convention - I am fixing that in this changeset
4) There is no calling convention directory defined in r2 for local plugins - I am fixing that in this changeset

shlr/sdb/src/Makefile
```diff
@@ -21,6 +22,9 @@ json.o: json/api.c json/js0n.c json/path.c json/rangstr.c json/indent.c
 install:
 	$(MAKE) -C .. install
 
+install-r2sdb:
```
trufae, line 25: this should be done in the root Makefile

trufae (changes requested):
Meson support is also needed

meson.build
```diff
@@ -520,6 +520,8 @@ sdb_exe = executable('sdb', ['shlr/sdb/src/main.c'] + sdb_files,
   ],
   implicit_include_directories: false,
   native: true,
+  install: true,
```
ret2libc, line 523: why this? I'm not sure it is required to make r2 work, this is why it was not installed previously.

zamaudio: Its needed if you want radare2-extras to be able to compile sdb texts into sdb binaries.

Makefile
```diff
@@ -218,7 +218,11 @@ install-doc-symlink:
 	for FILE in $(shell cd doc ; ls) ; do \
 		ln -fs "$(PWD)/doc/$$FILE" "${DESTDIR}${DOCDIR}" ; done
 
-install love: install-doc install-man install-www install-pkgconfig
+install-r2sdb:
+	mkdir -p "${DESTDIR}${BINDIR}"
+	cp -f shlr/sdb/sdb "${DESTDIR}${BINDIR}/r2sdb"
```
trufae, line 223: use INSTALL_PROGRAM instead of cp

Also, add support for symlink install of r2sdb

trufae (changes requested):
lgtm so far, but needs symstall support for me to merge

libr/util/file.c
```diff
@@ -160,9 +160,10 @@ R_API bool r_file_fexists(const char *fmt, ...) {
 }
 
 R_API bool r_file_exists(const char *str) {
+	char *absfile = r_file_abspath (str);
 	struct stat buf = {0};
 	r_return_val_if_fail (!R_STR_ISEMPTY (str), false);
-	if (file_stat (str, &buf) == -1) {
+	if (file_stat (absfile, &buf) == -1) {
```
trufae, line 166: absfile leaks

Makefile
trufae: also use $(SHLR)/sdb instead of shlr/sdb

trufae (approved):

ret2libc (approved):

meson.build
```diff
@@ -520,6 +520,8 @@ sdb_exe = executable('sdb', ['shlr/sdb/src/main.c'] + sdb_files,
...
+  rename: 'r2sdb'
```
ret2libc, line 524: @zamaudio 
```
../meson.build:517: WARNING: Passed invalid keyword argument "rename".
WARNING: This will become a hard error in the future.
WARNING: Installing target build for the build machine. This will fail in a cross build.
```

ret2libc: It seems like `rename` is not ok and also it is not ok to install this executable because this is compiled for the build machine, not the target machine (which may be different if you cross compile for android, for example). Could you fix this? I'm not sure what to do, but r2sdb should not be installed. Maybe radare2-extras should just require sdb installed, but it should get it directly from sdb repository.

zamaudio: Hmm, that's weird.  Is there a repo for sdb I can use for CI testing radare2-extras that require sdb?

ret2libc: What do you mean? https://github.com/radareorg/sdb ?

zamaudio: OK thanks, I just submitted a partial revert for the r2sdb stuff.

PR #17673, opened 2020-09-18
Fix leak in dwarf processing.

karliss:

**Detailed description**

Store read files with line information in in a dictionary. Simplifies the logic within `bin_dwarf` and prevents memory leak. This also reduces execution time by not having to process the same file multiple times. Previous version only stored last 1-2 files. 

As usual there is room for further optimizations but it isn't the worst part anymore. There is little value for doing it right now before improve other parts first.

When testing on Cutter debug executable
* prevent 6.7 GB memory leak
* reduce peak memory usage from 6.7GB to 1.9GB
* reduce execution time 49s -> 20s
memory consumption screenshot

**Test plan**

* comparisons performed by running `r2 -ci -qq executable`

Wait for coverage report to check if modifed code is executed during tests, based on that decide need for more tests.

**Closing issues**

### DIFF
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -1016,9 +1016,9 @@ static int bin_info(RCore *r, int mode, ut64 laddr) {
 }
 
 typedef struct {
-	int *line_starts;
+	size_t *line_starts;
 	char *content;
-	int line_count;
+	size_t line_count;
 } FileLines;
 
 static void file_lines_free(FileLines *file) {
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -1015,6 +1015,41 @@ static int bin_info(RCore *r, int mode, ut64 laddr) {
 	return true;
 }
 
+typedef struct {
+	int *line_starts;
+	char *content;
+	int line_count;
+} FileLines;
+
+static void file_lines_free(FileLines *file) {
+	if (!file) {
+		return;
+	}
+	free (file->line_starts);
+	free (file->content);
+	free (file);
+}
+
+FileLines *read_file_lines(const char *path) {
+	FileLines *result = R_NEW0 (FileLines);
+	if (!result) {
+		return result;
+	}
+	result->content = r_file_slurp (path, NULL);
+	if (result->content) {
+		result->line_starts = r_str_split_lines (result->content, &result->line_count);
+	}
+	if (!result->content || !result->line_starts) {
+		R_FREE (result);
+	}
+	return result;
+}
+
+static void file_lines_free_kv(HtPPKv *kv) {
+	free (kv->key);
+	file_lines_free (kv->value);
+}
+
 static int bin_dwarf(RCore *core, int mode) {
 	RBinDwarfRow *row;
 	RListIter *iter;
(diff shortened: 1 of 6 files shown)


libr/core/cbin.c
```diff
@@ -1015,6 +1015,41 @@ static int bin_info(RCore *r, int mode, ut64 laddr) {
...
+static void file_lines_free(FileLines *file) {
+	if (!file) {
+		return;
+	}
+	R_FREE (file->line_starts);
+	R_FREE (file->content);
+	R_FREE (file);
+}
+
+FileLines *read_file_lines (const char *path) {
```
ret2libc, line 1033:
```suggestion
FileLines *read_file_lines(const char *path) {
```

ret2libc (commented):
LGTM

karliss:
The reported test coverage looks good, but it seems that none of them use source code lines read from file. I tried replacing reading of file with returning fixed string and nothing failed. I will try to investigate it a bit more to see if test detecting it can be easily added.

libr/core/cbin.c
```diff
@@ -1015,6 +1015,41 @@ static int bin_info(RCore *r, int mode, ut64 laddr) {
...
+	int *line_starts;
+	char *content;
+	int line_count;
+} FileLines;
+
+static void file_lines_free(FileLines *file) {
+	if (!file) {
+		return;
+	}
+	R_FREE (file->line_starts);
```
trufae, line 1028: just free, without nullifying the pointers

libr/core/cbin.c
```diff
@@ -1015,6 +1015,41 @@ static int bin_info(RCore *r, int mode, ut64 laddr) {
 	return true;
 }
 
+typedef struct {
+	int *line_starts;
```
trufae, line 1019: size_t

libr/core/cbin.c
```diff
@@ -1015,6 +1015,41 @@ static int bin_info(RCore *r, int mode, ut64 laddr) {
...
+	char *content;
+	int line_count;
```
trufae, line 1021: size_t

trufae (changes requested):

libr/core/cbin.c
karliss: That also means changing the signature of r_str_split_lines and all the code using it. I can do it if you want, there aren't too many such places.

trufae: Yes i understand, just trying to be more formal and correct on every PR we merge. Not sure what @ret2libc thinks about that

karliss:
Looked into testing. It's somewhat problematic because dwarf test executables have absolute paths to the environement in which they were compiled. And options for source file prefix and substitution are lacking. Manually verified using `idj~{}` that it reads source lines for locally compiled executable.

Although codecov reported that lines are executed. So maybe there is an appropriate executable in testsuite with correct relative paths.

libr/core/cbin.c
karliss: Refactored in separate commit.

trufae (approved):

PR #17675, opened 2020-09-18
Do not let ASAN crash on large allocations

ret2libc:

**Detailed description**

It tries to "fix" asan build by allowing it to return NULL when large chunks of memory are requested.

**Test plan**

...


.github/workflows/ci.yml
```diff
@@ -83,7 +83,7 @@ jobs:
             cflags: '-DR2_ASSERT_STDOUT=1'
             meson_options: -Db_sanitize=address,undefined
             asan: true
-            asan_options: 'detect_leaks=0,detect_odr_violation=0'
+            asan_options: 'detect_leaks=0,detect_odr_violation=0,allocator_may_return_null=1'
```
kazarmy, line 86: Wasn't the original plan to clamp the allocation so that any useful information in the binary can be salvaged even though the binary is malformed? Not that I disapprove of this but others might.

libr/bin/p/bin_elf.inc
```diff
@@ -912,7 +912,7 @@ static RList* patch_relocs(RBin *b) {
 	struct Elf_(r_bin_elf_obj_t) *bin = obj->bin_obj;
 	info = obj ? obj->info: NULL;
 	size_t cdsz = info? (info->bits == 64? 8: info->bits == 32? 4: info->bits == 16 ? 4: 0): 0;
-	if (bin->ehdr.e_type != ET_REL && bin->ehdr.e_type != ET_DYN) {
+	if (bin->ehdr.e_type != ET_REL) {
```
trufae, line 915: this is breaking my previous fix

.github/workflows/ci.yml
trufae: yes. malformed binaries should be loaded

libr/bin/p/bin_elf.inc
trufae: test this in the r2pay arm64 binary

test/db/cmd/elf
```diff
@@ -16,3 +16,13 @@ EXPECT=<<EOF
 0x0804841d
 EOF
 RUN
+
+NAME=et_dyn no warnings for relocations
```
trufae, line 20: do something here. like listing relocs with ir.

.github/workflows/ci.yml
trufae: but i dont understand whats this asan option about. it is assuming that malloc can return null?

test/db/cmd/elf
ret2libc: nah, i just wanted to make sure `Warning: use -eio.cache=true to fix relocations` message was not shown. There is no need to do anything.

libr/bin/p/bin_elf.inc
ret2libc: what is r2pay binary?

.github/workflows/ci.yml
ret2libc: When you enable ASAN, if you do a malloc with a size larger than X (i think X by default is 0x100000000 or something like that) ASAN will complain about the size being too large. Now, what happens in real world, without ASAN, is that when you pass such a large value malloc will just return NULL. This option prevent ASAN from blocking the program execution when such malloc-"errors" happen.

libr/bin/p/bin_elf.inc
ret2libc: found it ;)

.github/workflows/ci.yml
trufae: so with this enabled, malloc will return null instead of warning and continue the execution?

test/db/cmd/elf
trufae: ok

.github/workflows/ci.yml
ret2libc: yes

kazarmy:
Is it possible to change the branch name so that GitHubCI runs the fuzz tests, or is that unnecessary?

ret2libc:
Closing in favour of https://github.com/radareorg/radare2/pull/17683 .

PR #17681, opened 2020-09-21
Add user configurable JSON encoding options 

### LINKED ISSUE #17368: izj not escaping json data properly.

hexploitable:

**Detailed description**

Currently, the JSON output for various commands can include forbidden characters including \x. This is outside of the JSON spec so requires some adjustments. Rather than assume the best case for the user, we can afford them multiple choices.

Whenever the config is updated for `json.encoding` we now set the encoding enum value inside `pj` and `pj_ke` (new) will use the associated encoding when generating the JSON string. There are many many places where `pj_ks` is used, and not all of them make sense to be converted to `pj_ke` - I don't have the intricate knowledge of r2, all the architectures and bin types etc to know all of those that make sense but I converted some easy ones that I think make sense - let me know if they don't make sense. Additionally, there are locations which I guessed would be candidates for this such as `iVj` and `icj` but these do not use `pj` yet, and instead print JSON strings the old way - these would also be good places to use `pj_ke` but I'm wary I'm dragging this out too long due to stuff I have going on.

**Encoding options:**
- [X] base64
- [X] hexpair string
- [X] array of bytes
- [X] strip non-printable
- [X] none, i.e. current backslash-x hex values
...

**Test plan**

...

**Closing issues**
closes #17368 
...


libr/util/str.c
```diff
@@ -1505,8 +1506,142 @@ R_API char *r_str_escape_utf32be(const char *buf, int buf_size, bool show_asciid
...
+				break;
+			case '\r':
+				*q++ = '\\';
+				*q++ = 'r';
+				break;
+			case '\\':
+				*q++ = '\\';
+				*q++ = '\\';
+				break;
+#if 0
```
ret2libc, line 1570: remove commented code please.

ret2libc (commented):
Nice! Some notes:
- I like the idea of passing the configuration of the encoding to `pj` and let it automatically print stuff with the right encoding.
- Please add unit tests for `r_str_encoded_json` and `r_str_escape_utf8_for_json_strip`
- Add integration test in `test/db/` to show how this fixes a problem in `iz`

Thanks!

libr/util/str.c
```diff
@@ -1505,8 +1506,142 @@ R_API char *r_str_escape_utf32be(const char *buf, int buf_size, bool show_asciid
 	return r_str_escape_utf (buf, buf_size, R_STRING_ENC_UTF32BE, show_asciidot, esc_bslash, false);
 }
 
-// JSON has special escaping requirements
-// TODO: merge with r_str_escape_utf() and r_str_byte_escape() using RStrEsc
+R_API char *r_str_encoded_json(const char *buf, int buf_size, const char *encoding) {
```
thestr4ng3r, line 1509: `encoding` should be an enum. You can make a helper function to quickly get the enum from the `r_config_get (r->config, "json.encoding")` string.

thestr4ng3r (changes requested):
> I like the idea of passing the configuration of the encoding to `pj` and let it automatically print stuff with the right encoding.

This would make pj (from RUtil) depend on RConfig, but pj itself should be independent.
Instead, I would add a function `r_config_pj_ks_e (RConfig *config, PJ *pj, const char *k, const char *e)` (better naming suggestions welcome) that can be part of config or core.

ret2libc:
> > I like the idea of passing the configuration of the encoding to `pj` and let it automatically print stuff with the right encoding.
> 
> This would make pj (from RUtil) depend on RConfig, but pj itself should be independent.
> Instead, I would add a function `r_config_pj_ks_e (RConfig *config, PJ *pj, const char *k, const char *e)` (better naming suggestions welcome) that can be part of config or core.

No, I did not mean passing RConfig, just the string. You would have `R_API PJ *pj_new(const char *encoding)`

libr/include/r_util/pj.h
```diff
@@ -37,6 +37,7 @@ R_API PJ *pj_knull(PJ *j, const char *k);
 R_API PJ *pj_kn(PJ *j, const char *k, ut64 n);
 R_API PJ *pj_kN(PJ *j, const char *k, st64 n);
 R_API PJ *pj_ks(PJ *j, const char *k, const char *v);
+R_API PJ *pj_ks_e(PJ *j, const char *k, const char *v, const char *e);
```
trufae, line 40: i think it will be better to use pj_ke as name (basically because what we are encoding here is not really a string. it's raw data, and must be encoded somehow.

so better to use `e` and use const  ut8* and size_t len

libr/util/str.c
trufae: agree. PJ_ENCODING_BASE64 and so on

libr/util/str.c
```diff
@@ -1505,8 +1506,142 @@ R_API char *r_str_escape_utf32be(const char *buf, int buf_size, bool show_asciid
...
+	if (!buf) {
+		return NULL;
+	}
+
+	if (!strcmp (encoding, "base64")) {
+		return r_base64_encode_dyn (buf, buf_size);
+	} else if (!strcmp (encoding, "bytearray")) {
```
trufae, line 1516: dont else after return

thestr4ng3r:
> > > I like the idea of passing the configuration of the encoding to `pj` and let it automatically print stuff with the right encoding.
> > 
> > 
> > This would make pj (from RUtil) depend on RConfig, but pj itself should be independent.
> > Instead, I would add a function `r_config_pj_ks_e (RConfig *config, PJ *pj, const char *k, const char *e)` (better naming suggestions welcome) that can be part of config or core.
> 
> No, I did not mean passing RConfig, just the string. You would have `R_API PJ *pj_new(const char *encoding)`

Yep, that's better. Only that encoding should be an enum.
And mabye in addition something like:
```
PJ *r_core_pj_new (RCore *core);
```
which will wrap the above `pj_new` and set everything up correctly.

libr/util/str.c
```diff
@@ -1505,8 +1506,142 @@ R_API char *r_str_escape_utf32be(const char *buf, int buf_size, bool show_asciid
...
+		int loop = 0;
+		int i = 0;
+		int len = buf_size < 0 ? strlen (buf) : buf_size;
+		int new_sz = (len * 2) + 1;
+
+		char *string_bytes = malloc (new_sz * sizeof (char));
```
trufae, line 1522: check for null, sizeof(char) is always 1

libr/util/str.c
```diff
@@ -1505,8 +1506,142 @@ R_API char *r_str_escape_utf32be(const char *buf, int buf_size, bool show_asciid
...
+		while(buf[loop] != '\0' && i < (new_sz-1)) {
```
trufae, line 1523: missing spaces

libr/util/str.c
```diff
@@ -1505,8 +1506,142 @@ R_API char *r_str_escape_utf32be(const char *buf, int buf_size, bool show_asciid
...
+R_API char *r_str_encoded_json(const char *buf, int buf_size, const char *encoding) {
+	if (!buf) {
+		return NULL;
+	}
+
+	if (!strcmp (encoding, "base64")) {
+		return r_base64_encode_dyn (buf, buf_size);
+	} else if (!strcmp (encoding, "bytearray")) {
+		int loop = 0;
+		int i = 0;
```
trufae, line 1518: size_t

libr/core/cconfig.c
```diff
@@ -853,6 +853,23 @@ static bool cb_emuskip(void *user, void *data) {
...
 }
 
+static bool cb_jsonencoding(void *user, void *data) {
+	RConfigNode *node = (RConfigNode*) data;
+	if (*node->value == '?') {
+		if (strlen (node->value) > 1 && node->value[1] == '?') {
+			r_cons_printf ("choose either: \n"\
+			"none (default)\n" \
+			"base64 - encode the json string values as base64\n" \
+			"bytearray - convert the string to hexpairs\n" \
```
trufae, line 863: i would say "hex" instead of bytearray.

also i expect also the type: array of numbers as encoder for that for example:

pj_ke (pj, "word", "foo", 3);

will result in:

{"word": [102, 102, 111]}

libr/core/cconfig.c
```diff
@@ -853,6 +853,23 @@ static bool cb_emuskip(void *user, void *data) {
 	return true;
 }
 
+static bool cb_jsonencoding(void *user, void *data) {
+	RConfigNode *node = (RConfigNode*) data;
+	if (*node->value == '?') {
+		if (strlen (node->value) > 1 && node->value[1] == '?') {
```
trufae, line 859: no need to walk all the string to chk only one char. do it like this: if(node->value[1] &&..

libr/core/cbin.c
```diff
@@ -455,7 +456,11 @@ static void _print_strings(RCore *r, RList *list, int mode, int va) {
 			pj_kn (pj, "length", string->length);
 			pj_ks (pj, "section", section_name);
 			pj_ks (pj, "type", type_string);
-			pj_ks (pj, "string", string->string);
+			if (!strcmp(encoding, "none")) {
```
trufae, line 459: imho strings dont need to be encoded, only `e` elements

trufae (changes requested):

libr/util/str.c
trufae: also add PJ_ENCODING_DEFAULT

libr/core/cbin.c
ret2libc: good point. we may want anyway a `pj_ke` as @trufae suggested, but I think a global configuration of the encoding at the PJ level is required anwyay.

trufae: we can have a global because that shouldnt be racy, but i would prefer not to change the constructor

hexploitable:
Awesome feedback, thanks all - I think I have enough information to close it out now.

karliss:
On the topic of having mechanism for configuring PJ based on RConfig. I'd like to remind about 64 bit integer printing. JSON spec doesn't require supporting 64bit integers and many parsers don't as it allows storing all numbers in double. If the parser supports 64bit or bigger integers (python) it's nicer if they are handled as numbers, but if it doesn't it would be good to have option for printing 64bit integers as strings so that the consumer of JSON objects can convert them to numbers manually.

So when designing the mechanism for PJ config please keep mind possibility that it might extend with other config options.

trufae:
Totally agree. We should opt out to encode the addresses as strings instead. Frida does that and its quite easy to manage with other libraries on top. So im fine for having more options under json.

libr/core/cbin.c
ret2libc: why?? Globals are bad, changing the constructor is the best way to go to provide options afaik.

thestr4ng3r: Agree with @ret2libc. Also it is highly important to also be able to serialize with fixed local-only settings. For example you wouldn't want these settings to affect projects for example, which should always use specific settings.

thestr4ng3r: If you don't want to change the ctor, we can find an alternative solution, for example an arg-less pj constructor that uses some default settings.

ret2libc: yeah of course we can provide both arg-less constructore and option-based constructor, but the constructor is where those things shall be configured imo, not at the caller site like how it is done in this pr.

libr/core/cconfig.c
hexploitable: I added both, personally, I would like to use a hexpair string, because then i dont need to do extra work on the python side (e.g. using r2pipe) to rebuild the string. But I also think it makes sense to encode it as an array like you said, so there is both :)

hexploitable:
Whenever the config is updated for `json.encoding` we now set the encoding enum value inside `pj` and `pj_ke` (new) will use the associated encoding when generating the JSON string. There are many many places where `pj_ks` is used, and not all of them make sense to be converted to `pj_ke` - I don't have the intricate knowledge of r2, all the architectures and bin types etc to know all of those that make sense but I converted some easy ones that I think make sense - let me know if they don't make sense. Additionally, there are locations which I guessed would be candidates for this such as `iVj` and `icj` but these do not use `pj` yet, and instead print JSON strings the old way - these would also be good places to use `pj_ke` but I'm wary I'm dragging this out too long due to stuff I have going on.

libr/util/pj.c
```diff
@@ -3,6 +3,8 @@
 #include <r_util.h>
 #include <r_util/r_print.h>
 
+static int pj_encoding = PJ_ENCODING_DEFAULT;
```
ret2libc, line 6: Globals are very bad, in particular when used inside a utility library like `pj`.

ret2libc (changes requested):
For me this is a no-go, as we shouldn't use globals unless really necessary. In this particular case, I don't think they are at all. I understand there would be many places to update if we are going to change the constructor, but I think setting this at constructor time is the right thing to do. Moreover, we could just add a new constructor e.g. `pj_new_with_encoding (const ut8 *encoding)` (or something with a better name, I'm not good at naming) and keep the default one `pj_new (void)` in most places. This would avoid having to change many existing places and it would also avoid the use of globals.

libr/util/pj.c
hexploitable: Ok, so we're saying that we need to make an additional constructor (e.g. `pj_new_with_encoding `) and in locations where we deem it necessary to encode, we create the `pj` instance using that? 

Just means we're going to have a lot of places that fetch the value from the config and then update it inside pj.
Will make the adjustment (I originally did this before pushing but it felt really strained) and then let you all decide where we want `pj_new` to be swapped with `pj_new_with_encoding`

hexploitable: Adjusted.

ret2libc:
Ok it seems much better now! If we want to avoid the duplication of code to get the cfg var and make sure everything in RCore is consistent, we can create a new API like `r_core_pj_new (RCore *core)`, which automatically get the config var and pass it to `r_pj_new_with_encoding`

libr/anal/fcn.c
```diff
@@ -1716,7 +1718,7 @@ R_API char *r_anal_function_get_json(RAnalFunction *function) {
 	int argc = argc_str? atoi (argc_str): 0;
 
 	pj_o (pj);
-	pj_ks (pj, "name", function->name);
+	pj_ke (pj, "name", function->name);
```
trufae, line 1721: i dont think we should change that. a string implies a specifric encoding.

libr/core/cbin.c
```diff
@@ -383,7 +383,8 @@ static void _print_strings(RCore *r, RList *list, int mode, int va) {
 	bin->minstrlen = minstr;
 	bin->maxstrlen = maxstr;
 	if (IS_MODE_JSON (mode)) {
-		pj = pj_new ();
+		const ut8 *encoding = r_config_get(r->config, "json.encoding");
```
trufae, line 386: missing space

libr/core/cbin.c
```diff
@@ -455,7 +456,9 @@ static void _print_strings(RCore *r, RList *list, int mode, int va) {
 			pj_kn (pj, "length", string->length);
 			pj_ks (pj, "section", section_name);
 			pj_ks (pj, "type", type_string);
-			pj_ks (pj, "string", string->string);
+			// data itself may be encoded so use pj_ke
+			pj_ke (pj, "string", string->string);
```
trufae, line 460: keep using pj_ks imho theres no benefit in adding a new api for just strings

libr/core/cbin.c
```diff
@@ -1694,8 +1699,8 @@ static int bin_relocs(RCore *r, int mode, int va) {
 		r_cons_println ("[Relocations]");
 		r_table_set_columnsf (table, "XXss", "vaddr", "paddr", "type", "name");
 	} else if (IS_MODE_JSON (mode)) {
-		// start a new JSON object
-		pj = pj_new ();
+		const ut8 *encoding = r_config_get(r->config, "json.encoding");
```
trufae, line 1702: same here, write a separate function that takes RCore and constructs this pj instance with the right settings as @ret2libc said

libr/util/pj.c
```diff
@@ -26,6 +26,23 @@ R_API PJ *pj_new(void) {
...
 		r_strbuf_init (&j->sb);
 		j->is_first = true;
+		j->encoding = PJ_ENCODING_DEFAULT;
+	}
+	return j;
+}
+
+R_API PJ *pj_new_with_encoding(const ut8 *encoding) {
+	PJ *j = pj_new ();
+	if (!strcmp("base64", encoding)) {
```
trufae, line 36: missing space, see all the other strcmps below too

libr/util/pj.c
```diff
@@ -26,6 +26,23 @@ R_API PJ *pj_new(void) {
 	if (j) {
 		r_strbuf_init (&j->sb);
 		j->is_first = true;
+		j->encoding = PJ_ENCODING_DEFAULT;
+	}
+	return j;
+}
+
+R_API PJ *pj_new_with_encoding(const ut8 *encoding) {
+	PJ *j = pj_new ();
```
trufae, line 35: check for nul

libr/util/str.c
```diff
@@ -1505,9 +1506,48 @@ R_API char *r_str_escape_utf32be(const char *buf, int buf_size, bool show_asciid
 	return r_str_escape_utf (buf, buf_size, R_STRING_ENC_UTF32BE, show_asciidot, esc_bslash, false);
 }
 
-// JSON has special escaping requirements
-// TODO: merge with r_str_escape_utf() and r_str_byte_escape() using RStrEsc
-R_API char *r_str_escape_utf8_for_json(const char *buf, int buf_size) {
+R_API char *r_str_encoded_json(const char *buf, int buf_size, int encoding) {
+	if (!buf) {
```
trufae, line 1510: use r_return_val_if_fail, also avoid negative buf_size to come

libr/util/str.c
```diff
@@ -1505,9 +1506,48 @@ R_API char *r_str_escape_utf32be(const char *buf, int buf_size, bool show_asciid
...
+		}
+		if (encoding == PJ_ENCODING_ARRAY) {
+			// get rid of the trailing comma
+			encodedString[i - 1] = '\0';
+		} else {
+			encodedString[i] = '\0';
+		}
+	} else if (encoding == PJ_ENCODING_STRIP) {
+		encodedString = r_str_escape_utf8_for_json_strip (buf, buf_size);
+	}
```
trufae, line 1543: else goes }

libr/util/str.c
```diff
@@ -1505,9 +1506,48 @@ R_API char *r_str_escape_utf32be(const char *buf, int buf_size, bool show_asciid
...
+			return NULL;
+		}
+		while (buf[loop] != '\0' && i < (new_sz - 1)) {
+			snprintf (encodedString + i, sizeof (encodedString), format, (ut8) buf[loop]);
+			loop += 1;
+			i += increment;
+		}
+		if (encoding == PJ_ENCODING_ARRAY) {
+			// get rid of the trailing comma
+			encodedString[i - 1] = '\0';
```
trufae, line 1537: use `es` instead of camelcased var name. al si i think i can be 0 in this line and harm the heap

test/unit/test_str.c
```diff
@@ -470,6 +470,23 @@ bool test_r_str_str_xy(void) {
 	mu_end;
 }
 
+bool test_r_str_encoded_json(void) {
+	char *invalidJsonString = "This is my \xe2 sample © string\n";
+	size_t len = strlen (invalidJsonString);
+
+	const char *array = r_str_encoded_json(invalidJsonString, len, PJ_ENCODING_ARRAY);
+	mu_assert_streq(array, "084,104,105,115,032,105,115,032,109,121,032,226,032,115,097,109,112,108,101,032,194,169,032,115,116,114,105,110,103,010", "string as array of uchar");
```
trufae, line 478: space before (

trufae (changes requested):

libr/core/cbin.c
hexploitable: Fixed everything else in the PR, this is the last thing I'm not clear on.

`pj_ks` is used for strings yes but the main point of this PR was that `pj_ks` can produce invalid json strings (e.g. including `\x` and the user will now get a new way to force those strings to be encoded as base64, hex, stripped, etc.

I guess, either:

1. `pj_ke` is new and is used only in the locations we think the source data could be invalid per the JSON spec (e.g. class names, method names, strings, version info)
2. We just keep using `pj_ks` but add some logic at the start of the function to check if there's an encoding configured, and branch to that logic instead of proceeding with the default string builder.

Option 2. would mean that we are encoding every current data point used by pj_ks which may or may not be desirable - thoughts?

hexploitable:
All changes were made. Let me know of any i's or t's that need dotted and crossed 👍

libr/include/r_bind.h
```diff
@@ -41,6 +42,7 @@ typedef struct r_core_bind_t {
 	RCoreNumGet numGet;
 	RCoreIsMapped isMapped;
 	RCoreDebugMapsSync syncDebugMaps;
+	RCorePJWithEncoding	pjWithEncoding;
```
ret2libc, line 45:
```suggestion
	RCorePJWithEncoding pjWithEncoding;
```

libr/util/pj.c
```diff
@@ -168,7 +207,11 @@ R_API PJ *pj_ka(PJ *j, const char *k) {
 R_API PJ *pj_ks(PJ *j, const char *k, const char *v) {
 	r_return_val_if_fail (j && k && v, j);
 	pj_k (j, k);
-	pj_s (j, v);
+	if (j->str_encoding != PJ_ENCODING_DEFAULT) {
```
ret2libc, line 210: so every `pj_ks` actually get encoded? Is that done on purpose?

libr/include/r_util/pj.h
```diff
@@ -14,10 +14,24 @@ typedef struct pj_t {
...
+	PJ_ENCODING_STR_HEX,
+	PJ_ENCODING_STR_ARRAY,
+	PJ_ENCODING_STR_STRIP,
+	PJ_ENCODING_NUM_STR,
+	PJ_ENCODING_NUM_HEX
+};
+
 /* lifecycle */
 R_API PJ *pj_new(void);
+R_API PJ *pj_new_with_encoding(const char *str_encoding, const char *num_encoding);
```
ret2libc, line 34: from an API point of view I would just pass the enum value instead of strings. Let `RCore` deal with converting strings to enum values. Also, give a name to the enum above and maybe split it in `enum PJEncodingStr` and `enum PJEncodingNum`.

libr/util/pj.c
hexploitable: It is on purpose yeah. Ultimately, there are trade-offs here. 

If we encode only some things then the user now needs to selectively decode certain things only, and that involves them writing code to check if it's encoded or not, or they need to try out all of the commands ahead of time to work out what fields need to be decoded. On the flip-side, if we encode everything, we encode things that are already safe JSON strings (per the spec) so its added work - albeit shouldn't be that noticeable.

IMO, the default behaviour (no encoding) is unchanged, so everyone is unaffected. Users will opt into encoding to fit their circumstances, and it may be better longer-term to extend this functionality to allow them to configure which fields explicitly they want to be encoded (via config file or something), but I think as the first version of this, we probably want to encode all. Plus if someone opts in for encoding, they'll know that they need to decode each thing and we can always adjust it if it's undesirable.

libr/include/r_util/pj.h
hexploitable: Ok, pushing the change in a second.

radare:
the json is wrong for pvj

Screenshot 2020-10-22 at 22 50 58

libr/util/pj.c
```diff
@@ -277,7 +330,6 @@ R_API char *pj_fmt(PrintfCallback p, const char *fmt, ...) {
 	char ch[2] = { 0 };
 	PJ *j = pj_new ();
 	while (*fmt) {
-		j->is_first = true;
```
ret2libc, line 280: ??

hexploitable: Yeah not sure how that was removed honestly, wasn't deliberate, re-adding it now - mb.

hexploitable:
> the json is wrong for pvj
> 
> Screenshot 2020-10-22 at 22 50 58

Fixed in 2a588331c - must have accidentally removed that LoC, definitely wasn't intentional.

hexploitable:
57b70a414 should also fix the unsigned long long vs int format specifier.

libr/util/pj.c
```diff
@@ -287,7 +287,7 @@ R_API PJ *pj_ne(PJ *j, ut64 n) {
 	if (j->num_encoding == PJ_ENCODING_NUM_STR) {
 		pj_raw (j, sdb_fmt ("\"%" PFMT64u "\"", n));
 	} else if (j->num_encoding == PJ_ENCODING_NUM_HEX) {
-		pj_raw (j, sdb_fmt ("\"0x%x\"", n));
+		pj_raw (j, sdb_fmt ("\"0x%llx\"", n));
```
ret2libc, line 290: nope, please use `PFMT64x`

hexploitable: done dbe56da8c

ret2libc (approved):

trufae (approved):

libr/core/cconfig.c
```diff
@@ -3534,6 +3568,16 @@ R_API int r_core_config_init(RCore *core) {
 	SETBPREF ("esil.nonull", "false", "Prevent memory read, memory write at null pointer");
 	SETCB ("esil.mdev.range", "", &cb_mdevrange, "Specify a range of memory to be handled by cmd.esil.mdev");
 
+	/* json encodings */
+	n = NODECB ("json.encoding.strings", "none", &cb_jsonencoding);
```
trufae, line 3572: there's no root `json` config variable, so i would suggest using `cfg.json.str` and `cfg.json.num`. it's shorter and fits into `cfg`. no need to highlight the "encoding" thing imho

trufae (changes requested):

trufae (changes requested):
Also, can you add a couple of tests using r2 commands instead of just the API? to confirm izj works as expected for example with different json encodings,

hexploitable:
> Also, can you add a couple of tests using r2 commands instead of just the API? to confirm izj works as expected for example with different json encodings,

Just pushed some, let's see that the build passes (expected as tests completed locally).

trufae (approved):

PR #17685, opened 2020-09-21
Always register the derived CC from the reg profile ##anal

trufae:

**Detailed description**

Some archs lack any definition of calling conventions. In addition, r2 can start without cc defined when running without proper installation or just using the single static binary. This patch registers the derived regprofile right after setting the arch

**Test plan**

this will probably break some tests,so its worth reviewing and updating them to include the new reg profile

**Closing issues**

...


libr/core/cbin.c
```diff
@@ -751,6 +752,10 @@ R_API void r_core_anal_cc_init(RCore *core) {
 		sdb_concat_by_path (core->anal->sdb_cc, dbpath);
 		core->anal->sdb_cc->path = strdup (dbpath);
 	}
+	if (sdb_isempty (core->anal->sdb_cc)) {
+		eprintf ("Warning: Missing calling conventions for '%s' / '%s'. Deriving it from the regprofile.\n", anal_arch, asm_cpu);
+	}
+	r_core_cmd0 (core, "tcc `arcc`");
```
thestr4ng3r, line 758: Are you sure this will never cmd-inject itself? I would rather use api here.

thestr4ng3r (changes requested):

thestr4ng3r:
lgtm when green

trufae:
I think the warning gets printed several times in some cases. Maybe its better to remove it. Wait a bit

trufae:
It's green. can we merge this? (the red is because the http test)

PR #17692, opened 2020-09-22
Add Global Import API

thestr4ng3r:

**Detailed description**

Again, needed for testing r2db.

**Test plan**

Run the tests

### DIFF
--- a/libr/anal/anal.c
+++ b/libr/anal/anal.c
@@ -130,6 +130,7 @@ R_API RAnal *r_anal_new(void) {
 	anal->lineswidth = 0;
 	anal->fcns = r_list_newf (r_anal_function_free);
 	anal->leaddrs = NULL;
+	anal->imports = r_list_newf (free);
 	r_anal_set_bits (anal, 32);
 	anal->plugins = r_list_newf ((RListFree) r_anal_plugin_free);
 	if (anal->plugins) {
@@ -694,16 +695,12 @@ R_API bool r_anal_is_prelude(RAnal *anal, const ut8 *data, int len) {
 }
 
 R_API void r_anal_add_import(RAnal *anal, const char *imp) {
-	if (anal->imports) {
-		RListIter *it;
-		const char *eimp;
-		r_list_foreach (anal->imports, it, eimp) {
-			if (!strcmp (eimp, imp)) {
-				return;
-			}
+	RListIter *it;
+	const char *eimp;
+	r_list_foreach (anal->imports, it, eimp) {
+		if (!strcmp (eimp, imp)) {
+			return;
 		}
-	} else {
-		anal->imports = r_list_newf (free);
 	}
 	char *cimp = strdup (imp);
 	if (!cimp) {
@@ -713,9 +710,6 @@ R_API void r_anal_add_import(RAnal *anal, const char *imp) {
 }
 
 R_API void r_anal_remove_import(RAnal *anal, const char *imp) {
-	if (!anal->imports) {
-		return;
-	}
 	RListIter *it;
 	const char *eimp;
 	r_list_foreach (anal->imports, it, eimp) {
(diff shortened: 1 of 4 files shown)


libr/anal/anal.c
```diff
@@ -690,3 +692,41 @@ R_API bool r_anal_is_prelude(RAnal *anal, const ut8 *data, int len) {
...
+	if (anal->imports) {
+		RListIter *it;
+		const char *eimp;
+		r_list_foreach (anal->imports, it, eimp) {
+			if (!strcmp (eimp, imp)) {
+				return;
+			}
+		}
+	} else {
+		anal->imports = r_list_newf (free);
```
ret2libc, line 706: Why did you opt for lazy allocation of the list? Can't we just allocate `anal->imports` when creating `RAnal` object like it's done for other stuff and in the `r_anal_purge_imports` just do a `r_list_purge`.

ret2libc (changes requested):

trufae:
This is falling into the same mistake as the anal classes. imho all this stuff should be stored and accessed from RBin. 

Also rlist is probably the worst storage for this. we need rbin things to be accessed by name and offset. that why i did rdb, but i guess everything fits into a hashtable now

libr/anal/anal.c
thestr4ng3r: Because it was like that before, this pr does not change the approach (yet).

thestr4ng3r:
> This is falling into the same mistake as the anal classes. imho all this stuff should be stored and accessed from RBin.

This has nothing to do with bin and even less with "imports" as coming from a binary. It is the feature **you** introduced here: https://github.com/radareorg/radare2/commit/231bbf02a7ef1bff083bed9e65168e93d47b61d0

> Also rlist is probably the worst storage for this. we need rbin things to be accessed by name and offset. that why i did rdb, but i guess everything fits into a hashtable now

What is rdb?

libr/anal/anal.c
ret2libc: yeah I saw that later. I think it makes sense to do this change directly here if you want.

thestr4ng3r: Agreed.

ret2libc (approved):

trufae:
RDB is a data structure i did 10 years ago based on multiple byte trees indexed by different offsets of different structs moved to Attic in SDB because nobody was looking at it, and a hashtable is just a more generic solution.

How imports have nothing to do with RBin? This API just provides an api to register names without any other value in it.

Sumarizing my complains on this issue:

* imports are not just a name
* imports are not related to analysis, they must be created in a binobject
* imports must be able to be tied to an address and optionally a library/context
* imports are wrongly stored in a list, because there use the be thousands of them
* this is O(n) which is not what we want.
* analysis and binary information should be tied to maps/bins/files, global things are bad
* the whole patch is a step back in several wip or well discussed topics on the rbin and anal design

My questions:

* why do you need this?
* i never used aii, but its pretty badly implemented, the command lacks help, usage, documentation and tests, and its conceptually wrong with the concept of "analysis"
* lacks json and other useful formatters
* the delete is just wrong missing* and such
* what is this namespace thing for imports? 
* also discovered afii, dunno who implemented this , but itlooks totally wrong to me. it also have the same problems as described in the rest of this comment.

Current help looks like this:

```
[0x00000000]> ai?
Usage: ai  [j*] [sz] # analysis/address information/imports
| aii [namespace]  global import (like afii, but global)
| aii-             delete all global imports
| ai @addr         show address information
[0x00000000]>
[0x00000000]> afi?
Usage: afi[jlp*]   <addr>
| afi              show information of the function
| afii[-][import]  show/add/delete imports used in function
| afi.             show function name in current offset
| afi*             function, variables and arguments
| afij             function info in json format
| afil             verbose function info
| afip             show whether the function is pure or not
| afis             show function stats (opcode, meta)
[0x00000000]> afii?
[0x00000000]>
```

trufae (approved):
Ok, let's go for it, but I think this feature should be redesigned to not be global and be more generic across RCons, RBin and RAsm for replacing words here and there. Won't block r2db

PR #17698, opened 2020-09-23
Kill RAnal.sdb_fcns

thestr4ng3r:
**DO NOT SQUASH!!!!!!!!!!!!!!!!!!!!!!!!!!!!**

Filling this template is power, France is bacon.

### Detailed description

I'll go through the previously remaining uses of `sdb_fcns` and how they have been killed.

#### Variables

... were saved in this sdb until #16547 which was merged months ago. Apparently the only remaining use of vars there was the `exists_var` callback for RPrint, which I rewrote for the correct variable storage.

#### Variable XRefs

`axt` queried accesses to variables from `sdb_fmt`, but there was no place that actually wrote these accesses there so it was basically a nop. The correct way to query accesses to vars is by using `afv=`.

#### Type propagation

This one is a bit more complicated. This is roughly how the relevant type matching part worked before:

```
foreach f in all functions {
    foreach call to callee in f {
        if callee is a user-defined function {
            sdb_fcns["arg i of callee has type"] = "type coming from f's vars" foreach i in len(callee args)
        } else { ... }
    }
    foreach i where "arg i of f has type" in sdb_fcns {
        f.arg[i].type = sdb_fcns["arg i of f has type"]
    }
}
```

So whenever types should be propagated from a caller into a callee's args, they weren't directly set, but rather temporarily saved by abusing `sdb_fmt` and only later actually applied to the callee. Do not confuse `f` and `callee` above! In the bottom loop, the types for `f` are applied, which are generally not the ones saved in the same loop iteration of the outermost loop.
This had the negative side-effect that some callee's just wouldn't get their args propagated, depending on their position in the list of functions.

I changed it to the more intuitive solution, just applying the types directly:
```
foreach f in all functions {
    foreach call to callee in f {
        if callee is a user-defined function {
            callee.arg[i].type = "type coming from f's vars" foreach i in len(callee args)
        } else { ... }
    }
}
```

This gets rid of the abuse of the global sdb as local temporary variables and fixes the order-dependence mentioned above.
The fixed order-dependence does slightly change the results of a few tests. I checked them in detail and some now have better results, some a bit worse with the worse results being the result of bugs in the backtracking of types and ignoring of pointer derefs and refs, but that part is not touched by this pr so it is overall more correct now.

**tl;dr**: The sdb was abused for local storage, this was fixed by fixing a bug, which affects some tests positively and some slightly negatively, but overall it's better now and definitely more correct.

### Test plan

run the tests

### DIFF
--- a/libr/core/anal_tp.c
+++ b/libr/core/anal_tp.c
@@ -235,10 +235,49 @@ static RList *parse_format(RCore *core, char *fmt) {
 	return ret;
 }
 
+static void retype_callee_arg(RAnal *anal, const char *callee_name, bool in_stack, const char *place, int size, const char *type) {
+	RAnalFunction *fcn = r_anal_get_function_byname (anal, callee_name);
+	if (!fcn) {
+		return;
+	}
+	if (in_stack) {
+		RAnalVar *var = r_anal_function_get_var (fcn, R_ANAL_VAR_KIND_BPV, size - fcn->bp_off + 8);
+		if (!var) {
+			return;
+		}
+		__var_retype (anal, var, NULL, type, false, false);
+	} else {
+		RRegItem *item = r_reg_get (anal->reg, place, -1);
+		if (!item) {
+			return;
+		}
+		RAnalVar *rvar = r_anal_function_get_var (fcn, R_ANAL_VAR_KIND_REG, item->index);
+		if (!rvar) {
+			return;
+		}
+		__var_retype (anal, rvar, NULL, type, false, false);
+		RAnalVar *lvar = r_anal_var_get_dst_var (rvar);
+		if (lvar) {
+			__var_retype (anal, lvar, NULL, type, false, false);
+		}
+	}
+}
+
 #define DEFAULT_MAX 3
 #define REGNAME_SIZE 10
 #define MAX_INSTR 5
 
+/**
+ * type match at a call instruction inside another function
+ *
+ * \param fcn_name name of the callee
+ * \param addr addr of the call instruction
+ * \param baddr addr of the caller function
+ * \param cc cc of the callee
+ * \param prev_idx index in the esil trace
+ * \param userfnc whether the callee is a user function (affects propagation direction)
+ * \param caddr addr of the callee
+ */
 static void type_match(RCore *core, char *fcn_name, ut64 addr, ut64 baddr, const char* cc,
 		int prev_idx, bool userfnc, ut64 caddr) {
 	Sdb *trace = core->anal->esil->trace->db;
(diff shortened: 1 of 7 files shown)


libr/core/anal_tp.c
```diff
@@ -412,6 +449,7 @@ static int bb_cmpaddr(const void *_a, const void *_b) {
 }
 
 R_API void r_core_anal_type_match(RCore *core, RAnalFunction *fcn) {
+	eprintf ("r_core_anal_type_match (%s @ 0x%"PFMT64x")\n", fcn->name, fcn->addr);
```
trufae, line 452: r_return_if_fail

libr/core/anal_tp.c
```diff
@@ -235,12 +235,52 @@ static RList *parse_format(RCore *core, char *fmt) {
...
+ * \param addr addr of the call instruction
+ * \param baddr addr of the caller function
+ * \param cc cc of the callee
+ * \param prev_idx index in the esil trace
+ * \param userfnc whether the callee is a user function (affects propagation direction)
+ * \param caddr addr of the callee
+ */
 static void type_match(RCore *core, char *fcn_name, ut64 addr, ut64 baddr, const char* cc,
 		int prev_idx, bool userfnc, ut64 caddr) {
+	eprintf ("  type_match(%s @ 0x%"PFMT64x", baddr: 0x%"PFMT64x", userfnc: %d, 0x%"PFMT64x")\n", fcn_name, addr, baddr, userfnc, caddr);
```
trufae, line 283: remove debug message

libr/core/anal_tp.c
thestr4ng3r: This is all debug stuff, I am still testing.

test/db/cmd/types
```diff
@@ -1260,7 +1260,7 @@ var char * var_28h @ rbp-0x28
 var char * var_20h @ rbp-0x20
 var int64_t var_18h @ rbp-0x18
 var int64_t var_8h @ rbp-0x8
-arg char ** arg2 @ rsi
+arg char * arg2 @ rsi
```
trufae, line 1263: is this change correct?

trufae (approved):
LGTM, just wondering about the char ** -> char * change if its a fix or a regression

XVilka (approved):

trufae:
Newshell red

trufae:
Redshell new

test/db/cmd/types
thestr4ng3r: It is a slightly worse result for how the code looks here but with respect to how the algorithm works it is correct, so yes it's a fix.

PR #17702, opened 2020-09-24
Implement the comma command to manage user-defined tables ##util

trufae:
Continuation of: https://github.com/radareorg/radare2/pull/16824

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


libr/core/cmd.c
```diff
@@ -1266,6 +1282,241 @@ static int cmd_stdin(void *data, const char *input) {
...
+			r_list_append (t->rows, row);
+		}
+	}
+	r_list_free (lines);
+	free (data);
+}
+
+static int cmd_table(void *data, const char *input) {
+	RCore *core = (RCore*)data;
+	static RTable *t = NULL;
```
trufae, line 1387: NOTE TO SELF: move into core->table

libr/core/cmd.c
```diff
@@ -1266,6 +1282,241 @@ static int cmd_stdin(void *data, const char *input) {
...
+				r_cons_printf ("%s\n", ts);
+				free (ts);
+			}
+		}
+		break;
+	case 0:
+		// print table
+		{
+			char *ts = r_table_tofancystring (t);
+			// char *ts = r_table_tostring (t);
```
trufae, line 1487: better to just call tostring()

libr/util/table.c
```diff
@@ -438,6 +438,37 @@ R_API char *r_table_tostring(RTable *t) {
...
+	}
+	r_list_foreach (t->cols, iter, col) {
+		r_strbuf_appendf (sb, " %s",  col->name);
+	}
+	r_strbuf_append (sb, "\n");
+
+	r_list_foreach (t->rows, iter, row) {
+		char *item;
+		int c = 0;
+	r_strbuf_appendf (sb, ",r ");
```
trufae, line 460: wrong indent

trufae (commented):
.

XVilka:
I propose the following syntax:
`[` as a command instead of `,`, and put the rest of the command inside `[]`:
```
[0x00000000]> [h xxd foo bar cow]
.--------------------.
| foo  | bar  | cow  |
)--------------------(
`--------------------'
[0x00000000]> [r 123 456 789]
.--------------------.
| foo  | bar  | cow  |
)--------------------(
| 123  | 456  | 789  |
`--------------------'
```

trufae:
i like the [] idea, but imnot sure if that will work the same for other commands, and this will be probably more complicated to support In TS. imho im good to merge this when CI decides to go green, and i can switch to [] to all the other commands using rtable instead of , or t. wdyt?

XVilka:
@ret2libc what do you think? I don't think using `[...]` syntax will break the tree-sitter parser.

XVilka:
Also it's red:
```
[XX] db/cmd/cmd_table comma table
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'afr
e scr.utf8=true
aflt > .tt
cat .tt~?
, .tt
rm .tt
,~?
' bins/mach0/mac-ls
-- stdout
@@ -1,2 +1,5 @@
 37
-37
+
+
+
+0

[XX] db/cmd/cmd_table comma table
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'afr
e scr.utf8=false
aflt > .tt2
cat .tt2~?
, .tt2
rm .tt2
,~?
' bins/mach0/mac-ls
-- stdout
@@ -1,2 +1,5 @@
 37
-37
+
+
+
+0
```

ret2libc:
@trufae @XVilka what is exactly the syntax you want to use with `[..]`? Also, I don't think it makes sense to merge `,` and change everything few commits later with `[`. Let's just choose now what we are going to use and let's go with it from the beginning.

XVilka:
@ret2libc see https://github.com/radareorg/radare2/pull/17702#issuecomment-698254970

libr/core/cmd.c
```diff
@@ -7086,6 +7332,7 @@ R_API void r_core_cmd_init(RCore *core) {
 		{"*",        "pointer read/write", cmd_pointer, NULL, &pointer_help},
 		{"-",        "open cfg.editor and run script", cmd_stdin, NULL, &stdin_help},
 		{".",        "interpret", cmd_interpret, NULL, &interpret_help},
+		{",",        "create and manipulate tables", cmd_table, NULL, &table_help},
```
ret2libc, line 7335: this probably cannot work as `table_help` is not of type `RCmdDescHelp`.

libr/core/cmd.c
```diff
@@ -153,6 +153,22 @@ static const char *help_msg_star[] = {
 	NULL
 };
 
+static const char *table_help[] = {
+	"Usage:", ",[/] [file]", "# load table data",
+	",", "", "display table",
+	".", "$foo", "aflt > $foo (files starting with '$' are saved in memory)",
```
ret2libc, line 159: why the `.` commands? I don't think they belong here.

libr/core/cmd.c
```diff
@@ -1266,6 +1282,239 @@ static int cmd_stdin(void *data, const char *input) {
...
+	}
+	r_list_free (lines);
+	free (data);
+}
+
+static int cmd_table(void *data, const char *input) {
+	RCore *core = (RCore*)data;
+
+	if (!core->table) {
+		core->table = r_table_new ();
```
ret2libc, line 1389: could you just initialize it together with all `RCore` fields and avoid the lazy initialization?

libr/core/cmd.c
```diff
@@ -1266,6 +1282,239 @@ static int cmd_stdin(void *data, const char *input) {
...
+	}
+
+	switch (*input) {
```
ret2libc, line 1392: could you split each case like how it is done in https://github.com/radareorg/radare2/blob/master/libr/core/cmd_write.c#L1981 ?

libr/core/cmd.c
```diff
@@ -1266,6 +1282,239 @@ static int cmd_stdin(void *data, const char *input) {
...
+				}
+				const char *type_name = (type_char == 's')
+					? "string": "number";
+				RTableColumnType *typeString = r_table_type (type_name);
+				r_table_add_column (core->table, typeString, s, 0);
+				i++;
+			}
+			r_list_free (list);
+			free (format);
+	//		free (s);
```
ret2libc, line 1416:
```suggestion
			free (s);
```

libr/core/cmd.c
```diff
@@ -153,6 +153,22 @@ static const char *help_msg_star[] = {
...
+	",", "", "display table",
+	".", "$foo", "aflt > $foo (files starting with '$' are saved in memory)",
+	". ", " [table-query]", "filter and print table. See ,? for more details",
+	"..", " file.csv", "load the csv file into a table",
+	",-", "", "reset table",
+	",/", "?", "query/filter current table",
+	",*", ">$foo", "print table as r2 commands",
+	",j", "", "print table in json format",
+	",h", "xxd foo bar cow", "define header column names and types",
+	",r", "xxd 1 2 foo", "adds a row using the given format string",
```
ret2libc, line 167: `xxd` is probably not required to add a row, just the elements are used AFAIS

libr/core/cmd.c
```diff
@@ -153,6 +153,22 @@ static const char *help_msg_star[] = {
...
+	"Usage:", ",[/] [file]", "# load table data",
+	",", "", "display table",
+	".", "$foo", "aflt > $foo (files starting with '$' are saved in memory)",
+	". ", " [table-query]", "filter and print table. See ,? for more details",
+	"..", " file.csv", "load the csv file into a table",
+	",-", "", "reset table",
+	",/", "?", "query/filter current table",
+	",*", ">$foo", "print table as r2 commands",
+	",j", "", "print table in json format",
+	",h", "xxd foo bar cow", "define header column names and types",
```
ret2libc, line 166:
```suggestion
	",h", " xxd foo bar cow", "define header column names and types",
```
same for other lines

libr/util/table.c
```diff
@@ -462,6 +462,37 @@ R_API char *r_table_tosimplestring(RTable *t) {
...
+	}
+	r_list_foreach (t->cols, iter, col) {
+		r_strbuf_appendf (sb, " %s",  col->name);
+	}
+	r_strbuf_append (sb, "\n");
+
+	r_list_foreach (t->rows, iter, row) {
+		char *item;
+		int c = 0;
+		r_strbuf_appendf (sb, ",r ");
```
ret2libc, line 484:
```suggestion
		r_strbuf_appendf (sb, ",r");
```

libr/util/table.c
```diff
@@ -462,6 +462,37 @@ R_API char *r_table_tosimplestring(RTable *t) {
...
+		r_list_foreach (row->items, iter2, item) {
+			RTableColumn *col = r_list_get_n (t->cols, c);
```
ret2libc, line 486: why is this required? I think that if a row has 3 items, there are at least 3 columns in the table, otherwise something would be very wrong in the table. Am I missing anything?

libr/core/cmd.c
```diff
@@ -1266,6 +1282,239 @@ static int cmd_stdin(void *data, const char *input) {
...
+				free (ts);
+			}
+		}
+		break;
+	case 'j':
+		// print table
+		if (input[1] == ' ') {
+			const char *q = r_str_trim_head_ro (input + 2);
+			if (r_table_query (core->table, q)) {
+				char *ts = r_table_tojson (core->table);
```
ret2libc, line 1495: this code is the same as few lines below. You can invert the `if` and write this code just one time and get it executed in both cases (input[1] ==  ' ' and the else case).

libr/core/cmd.c
```diff
@@ -153,6 +153,22 @@ static const char *help_msg_star[] = {
 	NULL
 };
 
+static const char *table_help[] = {
+	"Usage:", ",[/] [file]", "# load table data",
+	",", "", "display table",
+	".", "$foo", "aflt > $foo (files starting with '$' are saved in memory)",
+	". ", " [table-query]", "filter and print table. See ,? for more details",
+	"..", " file.csv", "load the csv file into a table",
```
ret2libc, line 161: I think this is not implemented.

libr/core/cmd.c
ret2libc: Maybe a typo

libr/core/cmd.c
```diff
@@ -1266,6 +1282,239 @@ static int cmd_stdin(void *data, const char *input) {
...
+		if (r_table_query (core->table, input + 1)) {
+			char *ts = r_table_tostring (core->table);
+			r_cons_printf ("%s", ts);
+			free (ts);
+		}
+		break;
+	case '$':
+		load_table (core, core->table, input);
+		break;
+	case ' ':
```
ret2libc, line 1445: what is the difference between this case and the `/` case? They look the same to me apart from the `r_str_trim_head_ro`.

libr/core/cmd.c
```diff
@@ -1266,6 +1282,239 @@ static int cmd_stdin(void *data, const char *input) {
...
+	case '/':
+		// query here
+		if (r_table_query (core->table, input + 1)) {
+			char *ts = r_table_tostring (core->table);
+			r_cons_printf ("%s", ts);
+			free (ts);
+		}
+		break;
+	case '$':
+		load_table (core, core->table, input);
```
ret2libc, line 1443: get the `data` from the alias before calling `load_table` and call it just with the data, so `load_table` doesn't need to perform the loading from files and it can potentially get data from everywhere, not just aliases and files.

libr/core/cmd.c
```diff
@@ -1266,6 +1282,239 @@ static int cmd_stdin(void *data, const char *input) {
...
+				char *ts = r_table_tostring (core->table);
+				if (ts) {
+					r_cons_printf ("%s\n", ts);
+					free (ts);
+				}
+			}
+		}
+		break;
+	case '.':
+		load_table (core, core->table, r_str_trim_head_ro (input + 1));
```
ret2libc, line 1458: same here, get the data from the file before calling `load_table` and just pass the data.

libr/core/cmd.c
```diff
@@ -1266,6 +1282,239 @@ static int cmd_stdin(void *data, const char *input) {
...
+		return;
+	}
+	RListIter *iter;
+	char *line;
+	RList *lines = r_str_split_list (data, "\n", 0);
+	bool expect_header = false;
+	bool expect_rows = false;
+	const char *separator = "|";
+	int ncols = 0;
+	r_list_foreach (lines, iter, line) {
```
ret2libc, line 1307: since the table has a well defined format:
```
<line to determine the separator>
<headers>
<line to determine start of rows>
<rows>
```
it's much easier to read if those things are done separately instead of using one big loop with multiple booleans that completely change the behaviour of the loop. In particular, because it seems you can't really go ahead if any of those pieces is missing from the table.

libr/core/cmd.c
```diff
@@ -1266,6 +1282,239 @@ static int cmd_stdin(void *data, const char *input) {
...
+			char *arg;
+			RList *args = r_str_split_list (line + strlen (separator), separator, 0);
+			RListIter *iter2;
+			ncols = 0;
+			if (r_list_length (t->cols) > 0) {
+				eprintf ("Warning: Not re-adding headers. Use ,- to reset the table.\n");
+				continue;
+			}
+			r_list_foreach (args, iter2, arg) {
+				char *s = strchr (arg, ' ');
```
ret2libc, line 1343: why is `strchr` needed here?

ret2libc (changes requested):

ret2libc:
> @ret2libc what do you think? I don't think using `[...]` syntax will break the tree-sitter parser.

Mh, I don't like having `]` at the end of the command, it adds no real value and it requires the intervention of the parser for no real reason. If you want to do things like `[...]` there are 2 ways and I don't like either:
1) you change the grammar and parse the table commands at the parser level (so it would be like a special case and not like any other command)
2) you treat it like a regular command, but then `]` would need to be parsed and filtered by the command handler, which is very dirty and bad.

So, all in all, I think I prefer `,` or just `[...`, without the ending `]`.

XVilka:
Ok, then I think just `[...` is good enough.

libr/core/cmd.c
trufae: Yeah, the / command can be removed i think, just using a space is fine

libr/core/cmd.c
trufae: done

libr/core/cmd.c
trufae: oh right, my bad.fixed

libr/util/table.c
trufae: would you prefer a runtime assert?

libr/core/cmd.c
trufae: it skips lines that dont match those rules

libr/core/cmd.c
trufae: to skip the first word

libr/core/cmd.c
trufae: r_str_split_list owns the pointer. this would cause a double free

libr/core/cmd.c
trufae: done

libr/core/cmd.c
trufae: but it allows to show the help.. maybe one of them should be non destructive

trufae: yeah, done. review again

libr/core/cmd.c
trufae: done

libr/core/cmd.c
trufae: Yes, but, if the reason is readability it will be better to move the whole thing into a separate .c file. if its for api accessibility thats a different topic, all of them or only the ones that hold local vars between braces?

libr/core/cmd.c
trufae: Correct. r_table_from_csv is not implemented and this command just tries to guess the format to load, which from my tests it seems to not work very reliabiliy. ill write tests for this and fix the behaviour

libr/core/cmd.c
trufae: thanks

libr/core/cmd.c
trufae: ok

libr/core/cmd.c
trufae: done

trufae:
@XVilka i have been trying to use , and [ and it feels weird that theres no ] at the end. maybe its my OCD. im sticking with comma for now. what do you think?

libr/core/cmd.c
ret2libc: the list generated by `r_str_split_list` doesn't free the string and since you are freeing the list a couple of lines above, if you don't free the string `s` will leak.

libr/util/table.c
ret2libc: if we don't think this is 100% true yeah, maybe better, though I would expect this to be always true, if the table API is ok.

libr/core/cmd.c
```diff
@@ -153,6 +153,22 @@ static const char *help_msg_star[] = {
 	NULL
 };
 
+static const char *cmd_table_help[] = {
+	"Usage:", ",[/] [file]", "# load table data",
+	",", "", "display table",
+	",", "$foo", "aflt > $foo (files starting with '$' are saved in memory)",
+	", ", " [table-query]", "filter and print table. See ,? for more details",
+	",.", " file.csv", "load the csv file into a table",
```
ret2libc, line 161: it is not clear that `,.` can be used to load regular files as well, not only .csv

libr/core/cmd.c
ret2libc: This function seems still very messy to me.

libr/core/cmd.c
ret2libc: why do you want to skip it?
```
.-------------------------.
| foo fuffa | bar  | cow  |
)-------------------------(
| ciao      | zoo  | boh  |
`-------------------------'
```
By doing so, if you load the above table `foo fuffa` becomes `fuffa`. Is that done on purpose?

ret2libc (changes requested):
Much better apart from some small things (and the `load_table` function which seems messy to me, but that's just my opinion so don't worry too much about it, just adjust the other stuff).

libr/core/cmd.c
```diff
@@ -1276,6 +1292,208 @@ static int cmd_stdin(void *data, const char *input) {
...
+		display_table (r_table_tocsv (core->table));
+		break;
+	case '*':
+		display_table (r_table_tor2cmds (core->table));
+		break;
+	case 0:
+		display_table (r_table_tofancystring (core->table));
+		break;
+	case 'j':
+		// print table
```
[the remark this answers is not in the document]
trufae: Fixed

libr/core/cmd.c
trufae: so do you agree on adding cmd_table.c and adding all those static functions in there? @ret2libc

libr/util/table.c
trufae: See line :249 .. it is not reliable so better keep this safe if until we get proper unit and r2 tests for this

libr/core/cmd.c
```diff
@@ -153,6 +153,22 @@ static const char *help_msg_star[] = {
 	NULL
 };
 
+static const char *cmd_table_help[] = {
+	"Usage:", ",[/] [file]", "# load table data",
+	",", "", "display table",
+	",", "$foo", "aflt > $foo (files starting with '$' are saved in memory)",
+	", ", " [table-query]", "filter and print table. See ,? for more details",
```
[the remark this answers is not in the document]
trufae: done

libr/core/cmd.c
ret2libc: It's also ok as-is, don't worry.

XVilka (changes requested):
4 broken tests
```
[XX] db/cmd/cmd_table comma table
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'afr
aflt > $a
cat $a~?
, $a
rm $a
,~?
' bins/mach0/mac-ls
-- stdout
@@ -1,2 +1,5 @@
 37
-37
+
+
+
+0

[XX] db/cmd/cmd_table comma table
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'afr
e scr.utf8=false
aflt > .tt2
cat .tt2~?
, .tt2
rm .tt2
,~?
' bins/mach0/mac-ls
-- stdout
select: Bad file descriptor
@@ -1,2 +1,5 @@
 37
-37
+
+
+
+0

[XX] TIMEOUT db/cmd/write Write misaligns jump instruction, control flow is reanalyzed
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -a x86 -b 64 -e anal.detectwrites=true -Qc 'wx e80200000075020000c3
af
pdr
wa jne 9
pdr
wx e802
pdr
' -
-- stdout
@@ -7,24 +7,3 @@
 | ----------- true: 0x00000009
 \ 0x00000009      c3             ret
 
-/ 10: fcn.00000000 (int64_t arg_2h);
-| ; arg int64_t arg_2h @ rbp+0x2
-| 0x00000000      7507           jne 9
-| ----------- true: 0x00000009  false: 0x00000002
-| 0x00000002      0000           add byte [rax], al
-| 0x00000004      007502         add byte [arg_2h], dh
-| ; CALL XREF from fcn.00000000 @ 
-| 0x00000007      0000           add byte [rax], al
-| ----------- true: 0x00000009
-\ 0x00000009      c3             ret
-
-/ 10: fcn.00000000 (int64_t arg_2h);
-| ; arg int64_t arg_2h @ rbp+0x2
-| 0x00000000      e802000000     call 7
-| 0x00000005      7502           jne 9
-| ----------- true: 0x00000009  false: 0x00000007
-| ; CALL XREF from fcn.00000000 @ 
-| 0x00000007      0000           add byte [rax], al
-| ----------- true: 0x00000009
-\ 0x00000009      c3             ret
-


-- exit status: 
[XX] db/cmd/cmd_table comma table
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'afr
e scr.utf8=true
aflt > .tt
cat .tt~?
, .tt
rm .tt
,~?
' bins/mach0/mac-ls
-- stdout
@@ -1,2 +1,5 @@
 37
-37
+
+
+
+0
```

PR #17707, opened 2020-09-24
Add r_str_escape_r2 helper function ##util

trufae:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...


ret2libc (changes requested):
A lot of unit tests please

libr/util/str.c
```diff
@@ -1351,6 +1351,32 @@ R_API char *r_str_escape(const char *buf) {
...
+	const char *p = buf;
+	char *q = new_buf;
+	while (*p) {
+		switch (*p) {
+		case ';':
+		case '$':
+		case '`':
+		case '\\':
+		case '"':
+			*q++ = ' ';
```
thestr4ng3r, line 1369: That's not really escaping, it's sanitizing.

libr/util/str.c
```diff
@@ -1351,6 +1351,32 @@ R_API char *r_str_escape(const char *buf) {
...
+			p++;
+			/* FALLTHRU */
```
thestr4ng3r, line 1371: Why fallthrough? Won't this make `pd 42 $$(!rm -rf /)` into `pd 42 $(!rm -rf /)` then?

libr/util/str.c
```diff
@@ -1351,6 +1351,32 @@ R_API char *r_str_escape(const char *buf) {
...
+		default:
+			*q++ = *p++;
+			break;
+		}
+	}
+	*q = '\0';
+	return new_buf;
+}
```
XVilka, line 1379: Please also have an empty line here.

trufae:
As long as there's no such thing as escaping in the oldshell. maybe this function should take an argument to specify filtering or escaping and use cfg.newshell value for it. what do you think @ret2libc ?

ret2libc:
> As long as there's no such thing as escaping in the oldshell. maybe this function should take an argument to specify filtering or escaping and use cfg.newshell value for it. what do you think @ret2libc ?

I'm not sure it is worth doing too complicated stuff for filter/escaping in oldshell as it was never done and doing it now would be wasted time probably?

trufae:
Thats why im doing only filtering for oldshell. Because there's no way to escape things properly in the current shell. So the only safe way would be to remove those bad chars

libr/util/str.c
trufae: fixed dunno what i smoked

PR #17726, opened 2020-09-28
Split GROUP RCmdDesc and regular ones ##newshell

ret2libc:

**Detailed description**

Some background:
when you do `a` as a command can mean 2 things: 1) `alias for aai - analysis information` 2) the group which includes all analysis information. In the help, you see them in 2 places: when you do `?` you will see `| a[?]                    analysis commands`, when you do `a?`, you will see `Usage: a     analysis commands` and then one line below `| a                  alias for aai - analysis information`.

As you can see, the same string can reference two different "concepts" in the help. Before this PR, we had merged the help strings in one single structure, however most of the other commands did not use those additional fields (group_summary, group_args_str) and they required special handling in the cmd_api code.

This PR splits the two concepts, instead of keeping everything in one single RCmdDesc. The R_CMD_DESC_TYPE_GROUP will hold the information about the "group" (e.g. "analysis commands" group), while the `d.group_data.exec_cd` will hold the info about the actual command that execute something (e.g. `alias for aai - analysis information`). This makes the code easier and the RCmdDescHelp structure easier to explain, I think. 

**Test plan**

Play with the help and `e cfg.newshell=true`.

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -7161,8 +7161,8 @@ R_API void r_core_cmd_init(RCore *core) {
 		case R_CMD_DESC_TYPE_ARGV:
 			cd = r_cmd_desc_argv_new (core->rcmd, root, cmds[i].cmd, cmds[i].argv_cb, cmds[i].help);
 			break;
-		case R_CMD_DESC_TYPE_GROUP:
-			cd = r_cmd_desc_group_new (core->rcmd, root, cmds[i].cmd, cmds[i].help);
+		case R_CMD_DESC_TYPE_INNER:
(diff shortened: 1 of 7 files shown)


XVilka (approved):
Sounds reasonable to me.

trufae (approved):
This is just boring to review, i dont have any complain except that it seems a huge overengineering, but I guess that would be a thing at some point. and its better to be formal

ret2libc:
> This is just boring to review, i dont have any complain except that it seems a huge overengineering, but I guess that would be a thing at some point. and its better to be formal

It is actually already a thing. You can just do `e cfg.newshell=true` and you see everything I'm touching here.

PR #17728, opened 2020-09-28
Do not free ds->opstr after r_parse_immtrim

ret2libc:
There is no new allocation in r_parse_immtrim, so it is not required to
free the "previous" string.

 

**Detailed description**

r_parse_immtrim just moves bytes but it does not allocate a new string, so the free is wrong.

**Test plan**

Run `e asm.imm.trim=true; pd 10` and you will get a double free.

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -1051,7 +1051,6 @@ static void ds_build_op_str(RDisasmState *ds, bool print_color) {
 	if (ds->immtrim) {
 		char *res = r_parse_immtrim (ds->opstr);
 		if (res) {
-			free (ds->opstr);
 			ds->opstr = res;
 		}
 		return;
--- a/test/db/cmd/cmd_pd
+++ b/test/db/cmd/cmd_pd
@@ -2217,6 +2217,7 @@ RUN
 NAME=asm.imm.trim
 FILE=bins/elf/crackme0x05
 CMDS=<<EOF
+s 0x080483d5
 pi 1 @e:asm.imm.trim=false
 pi 1 @e:asm.imm.trim=true
 EOF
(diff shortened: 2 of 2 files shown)


trufae:
I guess this is causing a segfault somehow? do you have a reproducer? the RParse API needs a redesign too. the current usage/behaviour is not properly defined, and it works.. but functions take so much arguments and its not easy to follow/extend.

ret2libc:
> I guess this is causing a segfault somehow? do you have a reproducer? the RParse API needs a redesign too. the current usage/behaviour is not properly defined, and it works.. but functions take so much arguments and its not easy to follow/extend.

See the first comment, Test plan:

> Test plan
> 
> Run `e asm.imm.trim=true; pd 10` and you will get a double free.

trufae (changes requested):
Can you add a test for this? it will be helpful for the rparser refactoring

trufae:
ping

ret2libc:
> ping

pong

trufae:
looks red

XVilka (approved):

ret2libc:
> looks red:
> 
> ```
> [XX] db/cmd/cmd_pd asm.imm.trim
> R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 's 0x080483d5
> pi 1 @e:asm.imm.trim=false
> pi 1 @e:asm.imm.trim=true
> ' bins/elf/crackme0x05
> -- stdout
> @@ -1,2 +1,2 @@
>  and esp, 0xfffffff0
> -and esp,
> +and esp, 
> ```

Green now!

trufae (approved):

PR #17732, opened 2020-09-28
Unwrap quoted strings when passing args to newer handlers ##newshell

ret2libc:
When a user wraps an argument, it means like in SH that he wants the
command to consider the argument as one single arg, even if composed by
multiple words. The same applies for r2newshell. So far we did not
unwrap quoted strings to maintain compatibility with old shell, but
doing this unwrapping is one of the nice feature of newshell.

 

**Detailed description**

When using old shell,
```
w Hello   World
```
would write to file the string 
```
Hello   World
```
because it simply takes as argument whatever is found after `w `. However, with `cfg.newshell=true`, arguments are parsed and split by the parser itself, which is not aware of how each command handlers wants to parse its arguments. To be generic, it splits arguments by space, but if a user wants to pass multiple words as a single argument, he can do that in the same way he is used to do that in bash/sh, by wrapping things in `"..."` or `'...'`.

When wrapping strings in quotes, the command should just get the quoted string without the quotes, because those quotes were just a mean to tell the parser the string had to be considered as a single argument. So far however we also passed the quotes to the command handlers, because this way we were compatible with existing handlers, which do the parsing themselves. So if you did:
```
e cfg.newshell=true
w "Hello   World"
```
the `w` handler would get the string
```
"Hello   World"
```
which is not what we wanted. This PR removes the quotes when the command handler is not an old one, which does the parsing by itself.

**Test plan**

Try writing things like `Hello;World` with the `w` command.
With oldshell, you had to quote the entire command, which was counter-intuitive and different from anything the user is usually used to.
With newshell you should be able to just do `w "Hello;World"`.

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -4830,28 +4830,42 @@ static void handle_substitution_args(struct tsr2cmd_state *state, TSNode args, R
 	}
 }
 
-static char *unescape_arg(struct tsr2cmd_state *state, TSNode arg, const char *special_chars) {
-	char *arg_str = ts_node_sub_string (arg, state->input);
+static char *unescape_arg_str(struct tsr2cmd_state *state, const char *arg_str, const char *special_chars) {
 	char *unescaped_arg = unescape_special_chars (arg_str, special_chars);
 	R_LOG_DEBUG ("original arg = '%s', unescaped arg = '%s'\n", arg_str, unescaped_arg);
+	return unescaped_arg;
+}
+
+static char *unescape_arg(struct tsr2cmd_state *state, TSNode arg, const char *special_chars) {
+	char *arg_str = ts_node_sub_string (arg, state->input);
+	char *unescaped_arg = unescape_arg_str (state, arg_str, special_chars);
 	free (arg_str);
 	return unescaped_arg;
 }
 
-static char *do_handle_ts_unescape_arg(struct tsr2cmd_state *state, TSNode arg) {
+static char *do_handle_ts_unescape_arg(struct tsr2cmd_state *state, TSNode arg, bool do_unwrap) {
 	if (is_ts_arg (arg)) {
-		return do_handle_ts_unescape_arg (state, ts_node_named_child (arg, 0));
+		return do_handle_ts_unescape_arg (state, ts_node_named_child (arg, 0), do_unwrap);
 	} else if (is_ts_arg_identifier (arg)) {
 		return unescape_arg (state, arg, SPECIAL_CHARS_REGULAR);
 	} else if (is_ts_single_quoted_arg (arg) || is_ts_double_quoted_arg (arg)) {
 		const char *special = is_ts_single_quoted_arg (arg)? SPECIAL_CHARS_SINGLE_QUOTED: SPECIAL_CHARS_DOUBLE_QUOTED;
-		return unescape_arg (state, arg, special);
+		char *o_arg_str = ts_node_sub_string (arg, state->input);
+		char *arg_str = o_arg_str;
+		if (do_unwrap) {
+			// remove quotes
+			arg_str[strlen (arg_str) - 1] = '\0';
+			arg_str++;
+		}
+		char *res = unescape_arg_str (state, arg_str, special);
+		free (o_arg_str);
+		return res;
 	} else if (is_ts_concatenation (arg)) {
 		uint32_t i, n_children = ts_node_named_child_count (arg);
 		RStrBuf *sb = r_strbuf_new (NULL);
 		for (i = 0; i < n_children; i++) {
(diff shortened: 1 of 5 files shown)


ret2libc:
By the way, of course if you actually want to write the `"`, in newshell you can do `w "\"Hello World\""`.

XVilka:
SH is Shangai (上海) 😉

ret2libc:
> SH is Shangai (上海) wink
> 
> Please add a test for this - it seems a core parser feature anyway, should be well-tested.

Tests added!

ret2libc:
```
[XX] db/cmd/cmd_pd pdJ string
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'w Snoo"ping as" usual,
Cs 30
pdJ 1
' malloc://128
-- stdout
@@ -1 +1 @@
-[{"offset":0,"text":"            0x00000000     .string \"Snoo\\\"ping as\\\" usual,\" ; len=30"}]
+[{"offset":0,"text":"            0x00000000     .string \"Snooping as usual,\" ; len=30"}]




[XX] db/cmd/regexp "/e /t\wst\d\d\d\s\w\w/i"
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'w "test123 ab"
w "Test123 ab" @444
?e
e search.in=block
b 777
"/e /t\wst\d\d\d\s\w\w/i"
' malloc://1024
-- stdout
@@ -1,3 +1,3 @@
 
-0x00000001 hit0_0 ""test123 ab""
-0x000001bd hit0_1 ""Test123 ab""
+0x00000000 hit0_0 "test123 ab"
+0x000001bc hit0_1 "Test123 ab"
```

I hope we agree `w "test123 ab"` should just write `test123 ab` (without the quotes)
and that if you want to write `Snoo"ping as" usual,` you should do `w "Snoo\"ping as\" usual,"`. Are we all ok with this? If yes, I'm going to change the tests (for `cfg.newshell=true` at least).

XVilka:
> I hope we agree w "test123 ab" should just write test123 ab (without the quotes)
and that if you want to write Snoo"ping as" usual, you should do w "Snoo\"ping as\" usual,". Are we all ok with this?

Absolutely.

thestr4ng3r:
> I hope we agree `w "test123 ab"` should just write `test123 ab` (without the quotes)
> and that if you want to write `Snoo"ping as" usual,` you should do `w "Snoo\"ping as\" usual,"`. Are we all ok with this? If yes, I'm going to change the tests (for `cfg.newshell=true` at least).

yes.

https://www.youtube.com/watch?v=LyM1UGD8sqM

libr/core/cmd.c
```diff
@@ -4830,28 +4830,42 @@ static void handle_substitution_args(struct tsr2cmd_state *state, TSNode args, R
...
 	} else if (is_ts_arg_identifier (arg)) {
 		return unescape_arg (state, arg, SPECIAL_CHARS_REGULAR);
 	} else if (is_ts_single_quoted_arg (arg) || is_ts_double_quoted_arg (arg)) {
 		const char *special = is_ts_single_quoted_arg (arg)? SPECIAL_CHARS_SINGLE_QUOTED: SPECIAL_CHARS_DOUBLE_QUOTED;
-		return unescape_arg (state, arg, special);
+		char *o_arg_str = ts_node_sub_string (arg, state->input);
+		char *arg_str = o_arg_str;
+		if (do_unwrap) {
+			// remove quotes
+			arg_str[strlen (arg_str) - 1] = '\0';
```
trufae, line 4857: can strlen be 0?

ret2libc: No, because `single_quoted_arg` and `double_quoted_arg` in grammar.js always have `'...'` or `".."`, so there are at least 2 characters, which are the single/double quotes.

test/db/cmd/regexp
```diff
@@ -41,6 +42,24 @@ b 777
 EOF
 EXPECT=<<EOF
 
+0x00000001 hit0_0 ""test123 ab""
```
ret2libc, line 45: what???

test/db/cmd/regexp
```diff
@@ -41,6 +42,24 @@ b 777
...
+0x000001bd hit0_1 ""Test123 ab""
+EOF
+RUN
+
```
ret2libc, line 49:
```suggestion
```

XVilka:
On OpenBSD:
```
[**]                                     db/esil/x86_64     1468 OK        45 BR        0 XX        8 FX
r2r(13245) in free(): use after free 0x56a26913600
gmake: *** [Makefile:12: r2r-tests] Abort trap (core dumped)
gmake -C unit
```

PR #17529, opened 2020-08-26
Checking format arguments in variadic functions

anisse:
**Your checklist for this pull request**
- [X] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [X] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

This PR adds many new warnings, they should be fixed before merging. This could be automated (gcc gives a correct suggestion), but it still needs thorough testing.

**Test plan**

...

**Closing issues**

N/A

PR #17733, opened 2020-09-29
Do not enforce A0 to be present as it is missing from many arch

ret2libc:

**Detailed description**

Although it should be like that, A0 is missing from many architectures and until that is properly implemented in all architectures we cannot keep this warning (`WARNING: r_reg_profile_to_cc: assertion 'a0' failed (line 415)`) everywhere.

**Test plan**

asan should be green.

**Closing issues**

...

### DIFF
--- a/libr/reg/profile.c
+++ b/libr/reg/profile.c
@@ -410,9 +410,12 @@ R_API char *r_reg_profile_to_cc(RReg *reg) {
 	const char *a2 = r_reg_get_name_by_type (reg, "A2");
 	const char *a3 = r_reg_get_name_by_type (reg, "A3");
 
-	// it is mandatory to have at least =A0 defined in the reg profile
+	// TODO: it is mandatory to have at least =A0 defined in the reg profile
 	// this will be enforced in reg/profile at parsing time
-	r_return_val_if_fail (a0, NULL);
+	// r_return_val_if_fail (a0, NULL);
+	if (!a0) {
+		return NULL;
+	}
 	if (!r0) {
 		r0 = a0;
 	}


trufae (changes requested):
A week ago we decided to actually ENFORCE this, to avoid nulls in the calling conventions derived from the reg profile when none can be loaded from sdb. So -1 for this. the proper fix is to update the regprofiles

ret2libc:
> A week ago we decided to actually ENFORCE this, to avoid nulls in the calling conventions derived from the reg profile when none can be loaded from sdb. So -1 for this. the proper fix is to update the regprofiles

Yes but enforcing something when it is hardly ready is not the way to go. The same could be said about https://github.com/radareorg/radare2/pull/17529 . The right thing to do would be to enable the printf-like warnings, but that would create 12321321 warnings, so we haven't enabled that. First, you have to fix most of the warnings, then we can enable the check. The same for this issue. There are many architectures that don't have A0 defined and in the meantime ASAN build is very red. Once someone will fix all the regprofiles, we can re-enable this.

trufae:
Took me 5min before entering a meeting to read the documentation for tricore, 6502, and PIC to get some understanding of the subroutine call conventions and with a simple oneliner i updated all the regprofiles. That may fix the drama.

- I understand its not good to enforce something that is not fixed
- Everyone in the discussion was happy about that because we understood the fix was easy and quick
- i agree that an assert here is not good because profiles can be feeded by the user, so it can cause crashes at runtime without control. so an early catch, like the one i have in one of my other PRs is WIP to solve this

ret2libc:
> Took me 5min before entering a meeting to read the documentation for tricore, 6502, and PIC to get some understanding of the subroutine call conventions and with a simple oneliner i updated all the regprofiles. That may fix the drama.
> 
>     * I understand its not good to enforce something that is not fixed
> 
>     * Everyone in the discussion was happy about that because we understood the fix was easy and quick
> 
>     * i agree that an assert here is not good because profiles can be feeded by the user, so it can cause crashes at runtime without control. so an early catch, like the one i have in one of my other PRs is WIP to solve this

You read fast man! Nice, closing in favour of https://github.com/radareorg/radare2/pull/17735 .

PR #17745, opened 2020-10-02
Add tests for pxd and pxd8

wasamasa:

**Detailed description**

I've revisited the tests for the pxd command while adding support for the pxdj family and noticed there were a few test cases missing for it, so I've added them.

**Test plan**

The test numbers should be verified by hand for the pxd8 case. The other default case doesn't need to be tested as it contains the same numbers as the pxd4 one (which is already handled).


trufae:
Please add tests for negative values and no argument too (by changing the block size before calling pxd)

pxd -12
pxd
pxd 0

trufae (changes requested):
see https://github.com/radareorg/radare2/pull/17745#issuecomment-703467852

XVilka (approved):

PR #17748, opened 2020-10-04
Fix #17391 - Fix unicode collision and box separators

### LINKED ISSUE #17391: Some Unicode characters encoded with base64 appear funny on the screen
### Work environment

| OS/arch/bits (mandatory)                             | Ubuntu 20.04.1 x86-64
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.6.0-git 26193 @ linux-x86-64 git.4.4.0-498-g154416c8f
commit: 154416c8fd4ce0a399c42e27ba88a0c535e320a0 build: 2020-08-02__02:36:41
### Expected behavior
base64 encoded characters should appear correctly on the screen.
### Actual behavior
Some Unicode characters encoded with base64 appear funny on the screen.
For example, the Unicode character `'\u03a6'` should appear as the Greek letter Phi, but appears as two gibberish characters on the screen. On the other hand, `'\u2660'` appropriately appears as a Spade symbol.
On bash: `$ echo -n $'\u03a6' | base64` yields: `zqY=`
However, on r2: `ag- ; agn "Phi" base64:zqY= ; agg` yields gibberish characters (see screenshot below)

Notes:
1. The character appears fine if printed directly, without being encoded to base64 first.
2. `e scr.utf8 = true`

phi

layderv:

**Detailed description**

This PR fixes https://github.com/radareorg/radare2/issues/17391. I found two ways to fix that issue:
1. check boundaries (x, y)
1. avoid utf8 collisions with box separators (such as │ and ┘)

I decided to go for (2) because:
* I did not find any reason why not to also use 0xd0-0xde vs the current 0xc8-0xd6
* I did not find a simple formula for checking the inner borders of the boxes (see below)

**Test plan**

Run `agn "Title" "ɸ"; agg` with many titles and bodies

This is the result of the proposed change in this PR:

This is what I meant above by "did not find a simple formula for inner borders": 

**Closing issues**

closes #17391


trufae (approved):
I dont know how those values as strings used in a switch/case didnt spotted any LGTM issue before.. anyway. good catch!

layderv:
@trufae please don't merge yet! The char `у ` breaks it, I'm working on a fix :)

trufae (changes requested):
Can you add tests too? :)

karliss:
Note that original range was chosen to match box characters in CP437 encoding. I don't see how shifting the range fixes anything, wouldn't that only make different characters to conflict. Looking at the screenshots I don't think I fully understand the mechanism due to which output breaks, I was expecting  `ɸ`  to be incorrectly displayed not the borders.

layderv:
@karliss you're right, I'm breaking other characters. The problem here is the code has been written with the assumption to always place one single char (I would need at least two here). There is some refactoring to do, unless you know better where to look and can point me

trufae:
So what to do?

layderv:
> So what to do?

I'm thinking of changing the code to handle fixes 2-byte chars (currently one). I don't think there's any faster way than refactoring.

If that's ok with you, I will close this PR as this commit didn't fix the issue

trufae:
Ok. You can close it, looking forward a proper fix :) thanks!

PR #17757, opened 2020-10-07
Show raw bytes in pdj when invalid

### LINKED ISSUE #10389: Bytes field in disas json when opcode invalid
# This template is meant for bug reports, if you have a feature request, please be as descriptive as possible and delete the template

Currently when an opcode is seen as invalid (in pdj), the json returned has no useful information apart from the size and the offset. However, I'd like to argue that the raw byte(s) are very useful as you can overcome some parts manually when this error occurs. Hence, it would be nice to have the 'bytes' section in the pdj command, so I can manually handle the bytes when radare2 fails to disassemble this instruction. 

Thanks!

layderv:

**Detailed description**

I've added the raw bytes along the type (invalid), size and offset fields to pdj.

**Test plan**

Manually generate a binary with an invalid opcode

**Closing issues**

closes #10389


trufae (approved):
i dont think its needed to add a test. but imho it makes sense to show the bytes even if the instruction is invalid

PR #17770, opened 2020-10-08
Add a progress bar for when scripts are running

dyc3:

**Detailed description**

Adds `r_print_progressbar_with_count`, and uses it to display progress when a script is being executed. The reason for not just changing `r_print_progressbar` is that I didn't want to break any existing functionality.

**Test plan**

```
r2 /bin/grep
Ps test
Po test
```


libr/core/cmd.c
```diff
@@ -6836,10 +6836,25 @@ R_API int r_core_cmd_lines(RCore *core, const char *lines) {
 	if (!odata) {
 		return false;
 	}
+	int line_count = 0;
+	{
+		int i = 0;
+		while (lines[i]) {
+			if (lines[i] == '\n') {
+				line_count++;
+			} else if (lines[i] == '\0') {
```
trufae, line 6845: This is already checked in while(x)

libr/core/cmd.c
```diff
@@ -6836,10 +6836,25 @@ R_API int r_core_cmd_lines(RCore *core, const char *lines) {
...
+			i++;
+		}
+	}
+
+	int current_line = 0;
 	nl = strchr (odata, '\n');
 	if (nl) {
 		r_cons_break_push (NULL, NULL);
 		do {
+			r_print_progressbar_with_count (core->print, current_line++, line_count, 80, true);
```
trufae, line 6857: I would add a config var to toggle this. I dont think we want progressbars all the time. Only for slow or large scripts maybe

libr/core/cmd.c
```diff
@@ -6866,6 +6881,7 @@ R_API int r_core_cmd_lines(RCore *core, const char *lines) {
 			r_core_task_yield (&core->tasks);
 		} while ((nl = strchr (data, '\n')));
 		r_cons_break_pop ();
+		printf ("\n");
```
trufae, line 6884: Uh?

trufae: Why not just flush after \r?

trufae (changes requested):

libr/core/cmd.c
dyc3: If this isn't present, the next line that is printed will go on the same line as the progress bar. Is there something else I should be using?

dyc3:
I think a more ideal implementation would be to allow commands to submit progress so it could be rendered out separately with less overhead. It would allow us to clear the line when a new line is printed so we don't get extra progress bars, and allow multiple progress bars to be shown at once. However, I think this would require a significant rework of how printing works.

dyc3:
I've been testing this out over the past couple of days, and one problem I'm having is that the progress bar shows up when piping r2's stdout somewhere. 2 potential solutions:
- Output the progress bar on stderr
- Disable the progress bar when output is piped

The other problem is that there's a series of 11 commands that gets run every time r2 starts up which causes the progress bar to appear, even when `scr.progressbar` is false. (Tested with `r2 -e scr.progressbar=false  /bin/ls` )

XVilka (changes requested):
It broke the Windows tests

https://ci.appveyor.com/project/radareorg/radare2/builds/35802430/job/lys5v3rm6jfehvi6

libr/core/cmd.c
```diff
@@ -6836,10 +6836,30 @@ R_API int r_core_cmd_lines(RCore *core, const char *lines) {
 	if (!odata) {
 		return false;
 	}
+	int line_count = 0;
+	{
+		int i = 0;
```
trufae, line 6841: use size_t, and remove this unnecessary nesting braces

libr/core/cmd.c
```diff
@@ -6836,10 +6836,30 @@ R_API int r_core_cmd_lines(RCore *core, const char *lines) {
 	if (!odata) {
 		return false;
 	}
+	int line_count = 0;
```
trufae, line 6839: use size_t too

libr/core/cmd.c
```diff
@@ -6836,10 +6836,30 @@ R_API int r_core_cmd_lines(RCore *core, const char *lines) {
...
+	{
+		int i = 0;
+		while (lines[i]) {
```
trufae, line 6842: Just use  `537 R_API int r_str_char_count(const char *string, char ch) {`

libr/core/cmd.c
```diff
@@ -6836,10 +6836,30 @@ R_API int r_core_cmd_lines(RCore *core, const char *lines) {
...
+			if (lines[i] == '\n') {
+				line_count++;
+			}
+			i++;
+		}
+	}
+
+#if __UNIX__
+	const bool show_progress_bar = core->print->enable_progressbar && r_config_get_i (core->config, "scr.interactive") && r_config_get_i (core->config, "scr.progressbar") && r_cons_isatty ();
```
trufae, line 6851: reduce the amount of duplicated code in here. something like this will look better:

```c
#if __UNIX__
 	 const bool istty = r_cons_isatty ();
#else
         const bool istty = true
 #endif

const bool show_progress_bar = core->print->enable_progressbar && r_config_get_i (core->config, "scr.interactive") && r_config_get_i (core->config, "scr.progressbar") && istty);
```

also move all this logic into a separate function so the condition can be checked in other places and the code will be more readable. at the end no #if code must be in here.

```
const show_progress_bar = want_progress_bar(core);
```

libr/core/cmd.c
```diff
@@ -6836,10 +6836,30 @@ R_API int r_core_cmd_lines(RCore *core, const char *lines) {
...
+#else
+	const bool show_progress_bar = core->print->enable_progressbar && r_config_get_i (core->config, "scr.interactive") && r_config_get_i (core->config, "scr.progressbar");
+#endif
+	int current_line = 0;
```
trufae, line 6855: sizet

libr/core/cmd.c
```diff
@@ -6866,6 +6886,9 @@ R_API int r_core_cmd_lines(RCore *core, const char *lines) {
 			r_core_task_yield (&core->tasks);
 		} while ((nl = strchr (data, '\n')));
 		r_cons_break_pop ();
+		if (show_progress_bar) {
+			printf ("\n");
```
trufae, line 6890: r_cons_newline(); or u dont want to use rcons buffering here?

libr/core/core.c
```diff
@@ -2731,7 +2731,9 @@ R_API bool r_core_init(RCore *core) {
 		char *a = r_str_r2_prefix (R2_FLAGS);
 		if (a) {
 			char *file = r_str_newf ("%s/tags.r2", a);
+			core->print->enable_progressbar = false;
```
trufae, line 2734: keep the old value and then set it back later, not force to true

libr/include/r_util/r_print.h
```diff
@@ -199,6 +202,8 @@ R_API int r_print_date_get_now(RPrint *p, char *str);
 R_API void r_print_zoom(RPrint *p, void *user, RPrintZoomCallback cb, ut64 from, ut64 to, int len, int maxlen);
 R_API void r_print_zoom_buf(RPrint *p, void *user, RPrintZoomCallback cb, ut64 from, ut64 to, int len, int maxlen);
 R_API void r_print_progressbar(RPrint *pr, int pc, int _cols);
+void r_print_progressbar_set_enable(const bool enable);
```
trufae, line 205: R_API

trufae (changes requested):

libr/util/print.c
```diff
@@ -1564,9 +1565,44 @@ R_API void r_print_progressbar(RPrint *p, int pc, int _cols) {
...
+	if (p->flags & R_PRINT_FLAGS_HEADER) {
+		if (enable_colors) {
+			p->cb_printf ("%s%4d%s%% %s%6d%s/%6d ", Color_GREEN, pc * 100 / total, Color_RESET, Color_GREEN, pc, Color_RESET, total);
+		} else {
+			p->cb_printf ("%4d%% %6d/%6d ", pc * 100 / total, pc, total);
+		}
+	}
+	cols -= 29;
+	enable_colors? p->cb_printf ("[%s", Color_YELLOW): p->cb_printf ("[");
+	for (i = cols * pc / total; i; i--) {
```
trufae, line 1591: uhm there are two bugs in this division:

* total can be 0 and cause divbyzero exception
* pc is signed, so a division unrepresentable result can happen too

i wrote some macros few commits ago in master to detect such issues, use that before entering the loop. and add proper checks to avoid negative values in in cols/total/..

libr/util/print.c
```diff
@@ -1564,9 +1565,44 @@ R_API void r_print_progressbar(RPrint *p, int pc, int _cols) {
...
+		p->cb_printf (block);
+	}
+	if (enable_colors) {
+		p->cb_printf (Color_RESET);
```
trufae, line 1595: use "%s", Color_RESET)

libr/util/print.c
```diff
@@ -1564,9 +1565,44 @@ R_API void r_print_progressbar(RPrint *p, int pc, int _cols) {
...
+	}
+	for (i = cols - (cols * pc / total); i; i--) {
+		p->cb_printf (h_line);
```
trufae, line 1598: first arg must be "%s"

libr/util/print.c
```diff
@@ -1564,9 +1565,44 @@ R_API void r_print_progressbar(RPrint *p, int pc, int _cols) {
...
+		if (enable_colors) {
+			p->cb_printf ("%s%4d%s%% %s%6d%s/%6d ", Color_GREEN, pc * 100 / total, Color_RESET, Color_GREEN, pc, Color_RESET, total);
+		} else {
+			p->cb_printf ("%4d%% %6d/%6d ", pc * 100 / total, pc, total);
+		}
+	}
+	cols -= 29;
+	enable_colors? p->cb_printf ("[%s", Color_YELLOW): p->cb_printf ("[");
+	for (i = cols * pc / total; i; i--) {
+		p->cb_printf (block);
```
trufae, line 1592: first arg must be "%s"

libr/util/print.c
```diff
@@ -1564,9 +1565,44 @@ R_API void r_print_progressbar(RPrint *p, int pc, int _cols) {
...
+	}
+	if (p->flags & R_PRINT_FLAGS_HEADER) {
+		if (enable_colors) {
+			p->cb_printf ("%s%4d%s%% %s%6d%s/%6d ", Color_GREEN, pc * 100 / total, Color_RESET, Color_GREEN, pc, Color_RESET, total);
+		} else {
+			p->cb_printf ("%4d%% %6d/%6d ", pc * 100 / total, pc, total);
+		}
+	}
+	cols -= 29;
+	enable_colors? p->cb_printf ("[%s", Color_YELLOW): p->cb_printf ("[");
```
trufae, line 1590: maybe join this chk witt he if above. and avoid using the ternary operator for non assignments.
```c
if (enable_colors) {
  if (FLAGS_HEADER) {
    printf ("%s%4d%s%%...")
 }
  printf("[%s",Color_YELLOW)
} else {
  ...
}
```

libr/util/print.c
dyc3: I've made most of the requested changes, but I'm a little unsure about what to do for this. How exactly should I handle these cases? I'm not sure what macros you're referring to.

dyc3: I've attempted to handle these cases adequately. The only macros I found were in `r_types_overflow.h`, and it wasn't clear to me how they would be used in this context.

dyc3:
I don't know what I did to cause those windows builds to fail. They were working before...

libr/core/cconfig.c
```diff
@@ -3706,6 +3706,7 @@ R_API int r_core_config_init(RCore *core) {
 	SETDESC (n, "Convert string before display");
 	SETOPTIONS (n, "asciiesc", "asciidot", NULL);
 	SETBPREF ("scr.confirmquit", "false", "Confirm on quit");
+	SETBPREF ("scr.progressbar", "true", "Display a progress bar when running scripts.");
```
trufae, line 3709: i think this option should be set to false by default it can be quite anoying for batch mode. or maybe make it honor the scr.interactive variable

trufae (changes requested):

libr/core/cconfig.c
dyc3: I believe it does respect `scr.interactive`:

https://github.com/radareorg/radare2/blob/bd4c52f282ac3851b473a038508c55cf783af373/libr/core/cmd.c#L6852

What is the "batch mode" you're referring to?

libr/util/print.c
```diff
@@ -1581,9 +1582,57 @@ R_API void r_print_progressbar(RPrint *p, int pc, int _cols) {
...
+		}
+		if (enable_colors) {
+			p->cb_printf ("%s", Color_RESET);
+		}
+		for (i = cols - (cols * pc / total); i; i--) {
+			p->cb_printf ("%s", h_line);
+		}
+		if (enable_colors) {
+			p->cb_printf ("%s]", Color_RESET);
+		}
```
trufae, line 1626: else goes here

libr/util/print.c
```diff
@@ -1581,9 +1582,57 @@ R_API void r_print_progressbar(RPrint *p, int pc, int _cols) {
...
+			p->cb_printf ("%s%4d%s%% %s%6d%s/%6d ", Color_GREEN, pc * 100 / total, Color_RESET, Color_GREEN, pc, Color_RESET, total, Color_YELLOW);
+		} else {
+			p->cb_printf ("%4d%% %6d/%6d ", pc * 100 / total, pc, total);
+		}
+		// TODO: determine string length of the numbers
+		cols -= 20;
+	}
+	if (cols > 0) {
+		if (enable_colors) {
+			p->cb_printf ("[%s", Color_YELLOW);
```
trufae, line 1611: just oneline with ternary operator enable_colors?Color_YELLOW:""

also, instead of hardcoding the yellow and green. you may take the colors from the selected palette

libr/util/print.c
```diff
@@ -1581,9 +1582,57 @@ R_API void r_print_progressbar(RPrint *p, int pc, int _cols) {
...
+	const char *block = p->cons->use_utf8? UTF_BLOCK: "#";
+
+	total = R_MAX (1, total);
+	pc = R_MAX (0, R_MIN (total, pc));
+	if (reset_line) {
+		p->cb_printf ("\r");
+	}
+	if (p->flags & R_PRINT_FLAGS_HEADER) {
+		if (enable_colors) {
+			p->cb_printf ("%s%4d%s%% %s%6d%s/%6d ", Color_GREEN, pc * 100 / total, Color_RESET, Color_GREEN, pc, Color_RESET, total, Color_YELLOW);
```
trufae, line 1602: newline after `total,` make shorter lines

libr/util/print.c
```diff
@@ -1581,9 +1582,57 @@ R_API void r_print_progressbar(RPrint *p, int pc, int _cols) {
...
+		// TODO: determine string length of the numbers
+		cols -= 20;
+	}
+	if (cols > 0) {
+		if (enable_colors) {
+			p->cb_printf ("[%s", Color_YELLOW);
+		} else {
+			p->cb_printf ("[");
+		}
+		for (i = cols * pc / total; i; i--) {
```
trufae, line 1615: cols * pc can result in negative value, making the loop iterate wrongly. You can solve this issue in different ways:

* use i > 0 instead of `i`
* compute cols*pc before the for body and check if the value is negative.
* use the OVFCHK apis

for this last thing you can use the following macros:

```
 65 UNSIGNED_DIV_OVERFLOW_CHECK(UT32_DIV_OVFCHK, ut32, UT32_MIN, UT32_MAX)
 66 UNSIGNED_DIV_OVERFLOW_CHECK(UT64_DIV_OVFCHK, ut64, UT64_MIN, UT64_MAX)
 71 SIGNED_MUL_OVERFLOW_CHECK(ST32_MUL_OVFCHK, st32, ST32_MIN, ST32_MAX)
 72 SIGNED_MUL_OVERFLOW_CHECK(ST64_MUL_OVFCHK, st64, ST64_MIN, ST64_MAX)
```

the problem here is that pc and cols are int and unsigned int. There are no macros for that, so probably the easiest way to go is checking if cols*pc > 0 before doing the for

libr/util/print.c
```diff
@@ -1581,9 +1582,57 @@ R_API void r_print_progressbar(RPrint *p, int pc, int _cols) {
...
+			p->cb_printf ("%s", block);
+		}
+		if (enable_colors) {
+			p->cb_printf ("%s", Color_RESET);
+		}
+		for (i = cols - (cols * pc / total); i; i--) {
```
trufae, line 1621: reuse the previous  cols*pc/total computation

libr/util/print.c
```diff
@@ -1581,9 +1582,57 @@ R_API void r_print_progressbar(RPrint *p, int pc, int _cols) {
 	p->cb_printf ("]");
 }
 
+/* TODO: handle screen width */
```
trufae, line 1585: Use `hex.cols` eval var for this. use p->coreb.cfgGet(p->coreb->core, "hex.cols)` instead of hardcode the 78 in there. Bear in mind that hex.cols is 16 by default because its used for printing hexdumps. but as long as hexdumps are a common visual metric that variable is used to determine the width of other printed stuff.

trufae (changes requested):

libr/util/print.c
dyc3: How do I grab the colors from the palette?

trufae: Use `r_cons_pal_get_i`, you can resolve the types by name with `r_cons_pal_get_name`. see cons/pal.c for all the possible values. The defines + enum in r_cons.h must be removed

trufae:
Let's wait for the CI, lgtm :)

dyc3:
I haven't made the changes you requested. Do you still want me to make those changes?

trufae:
Just fix the ci

28453E0A-B711-403E-A853-F53EBBDE3B48

libr/util/print.c
trufae:
```suggestion
			p->cb_printf ("%s%4d%s%% %s%6d%s/%6d%s ", Color_GREEN, pc * 100 / total, Color_RESET, Color_GREEN, pc, Color_RESET, total, Color_YELLOW);
```

trufae:
This branch cannot be rebased due to conflicts

trufae:
As i expected its still far from perfect. but at least after 2months people is checking it and there are some discussions in the chat.

Screenshot 2020-12-29 at 15 59 09

PR #17771, opened 2020-10-09
Update version in package.json ##json

officialcjunior:

**Detailed description**

The property `version` in `package.json` still reads 1.6.0. I've updated that to 4.5.1 here.

**Test plan**

None

**Closing issues**

None

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "pancake <pancake@nopcode.org>"
   ],
   "name": "radare2-git",
-  "version": "1.6.0-git",
+  "version": "4.5.1-git",
   "bugs": {
     "url": "https://github.com/radareorg/radare2/issues"
   }


XVilka:
Is this file necessary? Seems like no one uses it, maybe remove it completely? @trufae

trufae:
R2 is also packaged in npm

package.json
```diff
@@ -21,7 +21,7 @@
     "pancake <pancake@nopcode.org>"
   ],
   "name": "radare2-git",
-  "version": "1.6.0-git",
+  "version": "4.5.1-git",
```
trufae, line 24: thats not what configure.acr says. right now this is 4.6.0-git. anyway its a pain to update this file version. this can be done with a jq oneliner or nodejs. and should be good to npm pub at some point.. or well decide that npm is probably not the place to distribute r2. But as long as r2 is compiled with emscripten and wasm. this node module aims to provide a way to use r2 from nodejs with pure JS, no native things involved. So imho its fine to merge this if the version is the correct one, but its totally low prio unless we do emscripten builds in CI

trufae:
Also, it will be good to have a way to update the version of this file using jq or node, when building the emscripten builds. this is in the sys/emscripten.sh script

officialcjunior:
So, I was looking through the files built by `sys/emscripten.sh`, but I couldn't find understand where exactly this 'oneliner' should be written.

PR #17774, opened 2020-10-09
Set realname on all bin strings for better asm.flags.real when bin.str.real is set ##disasm

trufae:

**Detailed description**

before

Screenshot 2020-10-09 at 17 45 06

after

IMAGE 2020-10-09 17:44:52

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -432,7 +432,14 @@ static void _print_strings(RCore *r, RList *list, int mode, int va) {
 				? r_str_newf ("%s.str.%s", r->bin->prefix, string->string)
 				: r_str_newf ("str.%s", string->string);
 			r_name_filter (str, -1);
-			(void)r_flag_set (r->flags, str, vaddr, string->size);
+			RFlagItem *fi = r_flag_set (r->flags, str, vaddr, string->size);
+			if (fi) {
+				char *es = r_str_escape (string->string);
+				char *s = r_str_newf ("\"%s\"", es);
+				r_flag_item_set_realname (fi, s);
+				free (s);
+				free (es);
+			}
 			free (str);
 		} else if (IS_MODE_SIMPLE (mode)) {
 			r_cons_printf ("0x%"PFMT64x" %d %d %s\n", vaddr,
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -433,7 +433,7 @@ static void _print_strings(RCore *r, RList *list, int mode, int va) {
 				: r_str_newf ("str.%s", string->string);
 			r_name_filter (str, -1);
 			RFlagItem *fi = r_flag_set (r->flags, str, vaddr, string->size);
-			const bool realstr = r_config_get_i (core->config, "bin.str.real");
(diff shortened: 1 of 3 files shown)


XVilka:
Didn't check the code, but will it work with the ANSI control characters and hidden Unicode symbols? Like RTL control characters etc?

ITAYC0HEN (changes requested):
I think it was discussed during the real name refactoring and was decided not to be done. Or to be done via other eval.
I don't remember. 

Requesting changes as a way to block the merge 



@thestr4ng3r, do you remember?

thestr4ng3r:
Don't remember either. I think the main issue is that it can use a lot of memory.

ret2libc:
> I think it was discussed during the real name refactoring and was decided not to be done. Or to be done via other eval.
> I don't remember. 
> 
> Requesting changes as a way to block the merge 
> 
> 
> 
> @thestr4ng3r, do you remember? 

I don't remember either

ITAYC0HEN:
Thinking of Cutter, the rework on Real-Name was, among others, so we can get rid of the annoying prefix for API functions (mostly).
So `call sym.imp.user32.dll_MessageBoxA` will be shown as `call MessageBoxA`. On GUI, you mostly navigate by double-clicking so the visible naming can be truncated, unlike in r2 where you need to copy the name and to `s sym.imp.user32.dll_MessageBoxA`.

Strings are a different story, as they can be huge and messy and not necessarily improve the experience of the user.
I do think that there should be a way to show raw strings in the disassembly but am against having it shown only when `asm.flags.real` is eanbled. Otherwise, I will probably need to think of a way to do another refactoring for this part to improve the output

thestr4ng3r:
So I guess the better solution will be to add `asm.flags.strings` which will when disassembling check if the flag is in the strings flagspace and try to print the string there.

ITAYC0HEN:
I am fine with this suggestion :)

XVilka:
Agree on separation of strings and flags real names show. Thus, as @thestr4ng3r suggested - `asm.flags.strings`.

XVilka (changes requested):
Agree on separation of strings and flags real names show. Thus, as @thestr4ng3r suggested - `asm.flags.strings`.

trufae:
the memory issue problem is not triggered by the disasm, the problem is rbin loading those strings multiple times in memory. This can be resolved at runtime by calling a function that gets the string, instead of taking a field from the flagitem. Actually those strings can be built on the fly without taking more memory at load time. ill think on a better solution that doesnt involves consuming more memory. this was just a poc for discussion. thanks for the feedback.

thestr4ng3r:
> the memory issue problem is not triggered by the disasm, the problem is rbin loading those strings multiple times in memory. This can be resolved at runtime by calling a function that gets the string, instead of taking a field from the flagitem. Actually those strings can be built on the fly without taking more memory at load time.

Yep, that's exactly the proposal. sgtm.

XVilka:
@trufae could you please update the PR according to @thestr4ng3r proposal?

trufae:
Yes, but this may require some time, its not a quick fix. And right now i don't have that spare time. Hopefully soon

PR #17777, opened 2020-10-11
Implement `asm.pseudo` for RISC-V

officialcjunior:

**Detailed description**

What I'm hoping to add here, is support for `asm.pseudo` for the RISC-V architecture. I referred to [The RISC-V Instruction Set Manual](http://content.riscv.org/wp-content/uploads/2017/05/riscv-spec-v2.2.pdf), notably page 110, for the instructions.
As I am not an expert in this, I was mostly looking at the commit [e28ce](https://github.com/radareorg/radare2/commit/e28ce2990d90660a35053263555d796f034f74e3) and `parse_arm_pseudo.c` to make this possible.

**Test plan**
It seems to work, at least on layman's terms:

image

I've added a small test which succeeds, but I think its probably not enough, as it's not very big and I believe that there's lot more to fix and add in this PR.
```
$ r2r -i test/db/anal/riscv 
Loaded 4 tests.
[4/4]                       4 OK         0 BR        0 XX        0 FX
Finished in 0 seconds.
```

I'm still learning how the parsing works, so here, I've basically just used the function `parse()` from `parse_arm_pseudo`, so it probably have many useless things, that I didn't understand or thought wasn't important. So, I'm hoping to fix them all, one by one, through your requests.

**Closing issues**
None

### DIFF
--- a/libr/parse/p/parse_riscv_pseudo.c
+++ b/libr/parse/p/parse_riscv_pseudo.c
@@ -181,10 +181,11 @@ static int parse(RParse *p, const char *data, char *str) {
 					ptr = (char*)r_str_lchr (buf, ']');
 					if (n && ptr) {
 						char *rest = strdup (ptr + 1);
+						int dist = ptr - buf;
 						if (n > 0) {
-							snprintf (ptr, sizeof (ptr), "+%d]%s", n, rest);
+							snprintf (ptr, len + 1 - dist, "+%d]%s", n, rest);
 						} else {
-							snprintf (ptr, sizeof (ptr), "%d]%s", n, rest);
+							snprintf (ptr, len + 1 - dist, "%d]%s", n, rest);
 						}
 						free (rest);
 					}
--- a/libr/parse/p/parse_riscv_pseudo.c
+++ b/libr/parse/p/parse_riscv_pseudo.c
@@ -175,16 +175,16 @@ static int parse(RParse *p, const char *data, char *str) {
 					num = (char *)r_str_lchr (buf, ',');
 				}
 				if (num) {
-					n = atoi (num+1);
+					n = atoi (num + 1);
 					*ptr = '[';
-					memmove (num+1, ptr, strlen (ptr)+1);
+					r_str_cpy (num + 1, ptr);
 					ptr = (char*)r_str_lchr (buf, ']');
 					if (n && ptr) {
-						char *rest = strdup (ptr+1);
+						char *rest = strdup (ptr + 1);
 						if (n > 0) {
-							sprintf (ptr, "+%d]%s", n, rest);
+							snprintf (ptr, sizeof (ptr), "+%d]%s", n, rest);
 						} else {
-							sprintf (ptr, "%d]%s", n, rest);
+							snprintf (ptr, sizeof (ptr), "%d]%s", n, rest);
 						}
 						free (rest);
 					}
--- a/libr/parse/p/parse_riscv_pseudo.c
+++ b/libr/parse/p/parse_riscv_pseudo.c
@@ -1,0 +1,229 @@
+/* radare - LGPL - Copyright 2020 - Aswin C (officialcjunior) */
+
+#include <r_lib.h>
+#include <r_flag.h>
+#include <r_anal.h>
+#include <r_parse.h>
+
+static int replace(int argc, const char *argv[], char *newstr) {
+#define MAXPSEUDOOPS 10
+	int i, j, k, d;
+	char ch;
+	struct {
+		int narg;
+		char *op;
+		char *str;
+		int args[MAXPSEUDOOPS];
+	} ops[] = {
+		{ 0, "add", "# = # + #", { 1, 2, 3 } },
+		{ 0, "addi", "# = # + #", { 1, 2, 3 } },
+		{ 0, "and", "# = # & #", { 1, 2, 3 } },
+		{ 0, "andi", "# = # & #", { 1, 2, 3 } },
+		{ 0, "beq", "if (# == #) goto #", { 1, 2, 3 } },
+		{ 0, "bleu", "if (unsigned)# <= # goto #", { 1, 2, 3 } },
+		{ 0, "bltu", "if (unsigned)# < # goto #", { 1, 2, 3 } },
+		{ 0, "blt", "if (# < #) goto #", { 1, 2, 3 } },
+		{ 0, "beqz", "if (# == 0) goto #", { 1, 2 } },
+		{ 0, "bne", "if (# != #) goto #", { 1, 2, 3 } },
+		{ 0, "bnez", "if (# != 0) goto #", { 1, 2 } },
+		{ 0, "bgez", "if (# >= 0) goto #", { 1, 2 } },
+		{ 0, "bgtz", "if (# > 0) goto #", { 1, 2 } },
+		{ 0, "fld", "# = #", { 1, 2 } },
+		{ 0, "j", "jmp #", { 1 } },
+		{ 0, "jr", "jmp #", { 1 } },
+		{ 0, "jalr", "jmp #", { 1 } },
+		{ 0, "jal", "jmp #", { 1 } },
+		{ 0, "ld", "# = (double)[#]", { 1, 2 } },
+		{ 0, "li", "# = #", { 1, 2 } },
+		{ 0, "lh", "# = [#]", { 1, 2 } },
+		{ 0, "lui", "# = #", { 1, 2 } },
+		{ 0, "lbu", "# = (unsigned)[#]", { 1, 2 } },
+		{ 0, "lhu", "# = (unsigned)[#]", { 1, 2 } },
+		{ 0, "lw", "# = [#]", { 1, 2 } },
+		{ 0, "mv", "# = #", { 1, 2 } },
+		{ 0, "or", "# = # | #", { 1, 2, 3 } },
+		{ 0, "sd", "[#] = (double)#", { 2, 1 } },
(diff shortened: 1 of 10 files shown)


libr/parse/p/parse_riscv_pseudo.c
```diff
@@ -0,0 +1,229 @@
...
+				continue;
+			}
+		}
+		if (!strcmp (ops[i].op, argv[0])) {
+			if (newstr) {
+				d = 0;
+				j = 0;
+				ch = ops[i].str[j];
+				for (j = 0, k = 0; ch != '\0'; j++, k++) {
+					ch = ops[i].str[j];
```
XVilka, line 68: Too many levels of nesting - please split to a smaller blocks, functions, etc.

libr/parse/p/parse_riscv_pseudo.c
```diff
@@ -0,0 +1,229 @@
...
+				if (!num) {
+					num = (char *)r_str_lchr (buf, ',');
+				}
+				if (num) {
+					n = atoi (num+1);
+					*ptr = '[';
+					memmove (num+1, ptr, strlen (ptr)+1);
+					ptr = (char*)r_str_lchr (buf, ']');
+					if (n && ptr) {
+						char *rest = strdup (ptr+1);
```
XVilka, line 183: Same here.

XVilka (changes requested):

libr/parse/p/parse_riscv_pseudo.c
officialcjunior: Thank you for your reply. I really appreciate it.

So, I was looking up about functions similar to what you are referring to and saw `can_replace()` in `parse_mips_pseudo` which I believe checks if something can be replaced. Is that what you're talking about here? Or is it the `if (ch == '#')` part?

If not, can you just write what and which exact part should be done by a function?

libr/parse/p/parse_riscv_pseudo.c
officialcjunior: And here, the stuff inside the `if (n && ptr)` should be made into a function?

trufae: most of your code is copypasta from other pseudo plugins that are also made of pasta, we may probably want to implement generic parser, but all the new code should conform the rules.

libr/parse/p/parse_riscv_pseudo.c
trufae: he is refering to moving the body of the function to a separate function. so the whole function fits a single screen

libr/parse/p/parse_riscv_pseudo.c
trufae: not needed,this block is small enough.

libr/parse/p/parse_riscv_pseudo.c
```diff
@@ -0,0 +1,229 @@
...
+						if (n > 0) {
+							sprintf (ptr, "+%d]%s", n, rest);
```
trufae, line 185: dont use sprintf, use snprintf or so, otherwise this thing can overflow

libr/parse/p/parse_riscv_pseudo.c
```diff
@@ -0,0 +1,229 @@
...
+			if (ptr) {
+				*ptr = 0;
+				num = (char*)r_str_lchr (buf, ' ');
+				if (!num) {
+					num = (char *)r_str_lchr (buf, ',');
+				}
+				if (num) {
+					n = atoi (num+1);
+					*ptr = '[';
+					memmove (num+1, ptr, strlen (ptr)+1);
```
trufae, line 180: spaces around + and suhc..but you can use r_str_cpy, which is doing memmove under the hood, but you dont have to use the strlen by hand

libr/parse/p/parse_riscv_pseudo.c
trufae: but as i said in my previous comment, i dont think refactoring this is really necessary because this code is duplicated multiple times and this interesting and useful feature may take ages to be merged if this cleanup ends in the full refactoring that implies having a common pseudo parser for all archs so we dont duplicate code

trufae (changes requested):

officialcjunior:
Thanks for the reply.

Right now, I've used `r_str_cpy`, replaced `sprintf` with `snprintf` and added the missing spaces.

About `snprintf`, what should be the maximum number of bytes? 
I had used `sizeof (ptr)` here.

trufae:
Can you rebase the PR i dont see why it's not passing the CI otherwise

officialcjunior:
I think it's because I have not updated `libr/parse/meson.build`, yet.

Can I add that? I think that'll solve the issue.

Along with that, I saw `plugins.static.cfg`. Should I update that, too?

trufae:
Yep update those files

officialcjunior:
Did that, but we've got a problem with the `snprintf`: 
```
error: argument to ‘sizeof’ in ‘snprintf’ call is the same expression as the destination; did you mean to provide an explicit length? [-Werror=sizeof-pointer-memaccess]
        snprintf (ptr, sizeof (ptr), "+%d]%s", n, rest);
```
If you're around, you may suggest the changes, meanwhile I'll look up about fixing this.

libr/parse/p/parse_riscv_pseudo.c
```diff
@@ -0,0 +1,229 @@
...
+				}
+				if (num) {
+					n = atoi (num + 1);
+					*ptr = '[';
+					r_str_cpy (num + 1, ptr);
+					ptr = (char*)r_str_lchr (buf, ']');
+					if (n && ptr) {
+						char *rest = strdup (ptr + 1);
+						if (n > 0) {
+							snprintf (ptr, sizeof (ptr), "+%d]%s", n, rest);
```
trufae, line 185: sizeof(ptr) is always size_t, so its not how you should use snprintf. 

* buf contents size len + 1
* ptr is buf+x

remaining chars in ptr are:

size_t remaining = len + 1 - (ptr - buf);

trufae (changes requested):

libr/parse/p/parse_riscv_pseudo.c
officialcjunior: `snprintf (ptr, len + 1 - (ptr - buf), "%d]%s", n, rest);`
Like that?
I didn't produce any warnings.
What exactly is `(ptr - buf)`

trufae: Better use a variable one line above to make it more readable.

This ptr-buf is the distance between the current char and the beginning of the buffer

officialcjunior: Sure. Variable `dist`of type `int`?

libr/parse/p/parse_riscv_pseudo.c
```diff
@@ -181,10 +181,11 @@ static int parse(RParse *p, const char *data, char *str) {
 					ptr = (char*)r_str_lchr (buf, ']');
 					if (n && ptr) {
 						char *rest = strdup (ptr + 1);
+						int dist = ptr - buf;
```
trufae, line 184: Why not hold the result in a variable instrad of computing it twice below?

Also. The type must be size_t

officialcjunior: Alright, so, I'll make it:
`size_t dist = len + 1 - ptr - buf;`

libr/parse/p/riscv_pseudo.mk
```diff
@@ -0,0 +1,16 @@
...
+STATIC_OBJ+=${OBJ_RISCVPSEUDO}
+
+ifeq ($(CC),cccl)
+	RISCV_CFLAGS:=${CFLAGS}
+	else
+	RISCV_CFLAGS:=${CFLAGS} ${LINK}
+	endif
+
+${TARGET_RISCVPSEUDO}: ${OBJ_RISCVPSEUDO}
+	${CC} $(call libname,parse_riscv_pseudo) -L../../util -llibr_util \
```
trufae, line 14: -llibr_util? It should be -lr_util

trufae (changes requested):
The failing test it seems to be because the parse is not working well and causes it to miss a var. can u compare the input and output of that string because this plugin is also in charge of doing that.. also its possible that the current test is wrong. So maybe also an option to update the output in the broken test. Because it was replaced using the wrrong parse plugin.

Btw did u checked if theres any other missing pseudo plugin for any of the archs in default r2?

trufae:
Please run `r2r -i test/db/formats/elf/elf-riscv64`

and press `f` to fix the test and commit the changes. this PR will also fix another arm64 travis test

trufae (approved):

eagleoflqj:
https://builds.sr.ht/~xvilka/job/332435#task-test
db/cmd/cmd_list also needs to be updated

trufae:
looks good to merge for me

libr/parse/p/parse_riscv_pseudo.c
officialcjunior: Okay, marking this as resolved

libr/parse/p/parse_riscv_pseudo.c
officialcjunior: Marking as resolved.

XVilka (approved):

officialcjunior:
Also @trufae, shall I replace `sprintf` with `snprintf` on `parse_att2intel` too?

trufae:
yep!

PR #17779, opened 2020-10-11
Add initial function prelude support for v850 ##anal

trufae:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/anal/p/anal_v850.c
+++ b/libr/anal/p/anal_v850.c
@@ -541,10 +541,18 @@ static char *get_reg_profile(RAnal *anal) {
 	return strdup (p);
 }
 
+static RList *anal_preludes(RAnal *anal) {
+#define KW(d,ds,m,ms) r_list_append (l, r_search_keyword_new((const ut8*)d,ds,(const ut8*)m, ms, NULL))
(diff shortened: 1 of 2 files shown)


XVilka:
Something like this should come with test, imho.

trufae:
Yes, the problem is that we have no distributable v850 binary. the one that is in the testsuite is actually for v810, and our v850 disassembler is for v800, r2ghidra/v850 is the only able to handle that disasm/analytsis properly right now. ill try to get a toolchain to generate proper code for that. i can cook something by hand too, but i would prefer to get the right toolchains working first. From what ive seen < v850 there are no actual function preludes. and alignment is important, so those archs need more metainfo in the asm plugins too.

trufae:
Yay i found the way to get proper prepare/dispose functions with the toolchain. I have commited a bin in tests and will add some tests and more preludes.

I am also fixing the v850 disassembler and analyzer to support the last instructions in the newest E models.

XVilka (approved):

trufae:
rebased after the -Werror, lets hope its green too :)

PR #17786, opened 2020-10-15
Refactor r_sign metrics match ##signatures

swoops:

**Detailed description**

This refactors r_sign function metrics matching. Now instead of a function to match each metric type, there is a single function. This allows RSignItems to created only once, instead of once per type. Additionally, a cache was added to some of the expensive function information. This results in a significant speed increase.
 
```
$ ./r2test.sh 
Warning: run r2 with -e io.cache=true to fix relocations in disassembly
generated zignatures: 3182
[+] searching function metrics
hits: 15919
774.342817
15919
$ ./r2test.sh 
Warning: run r2 with -e io.cache=true to fix relocations in disassembly
generated zignatures: 3182
[+] searching function metrics
hits: 286318
134.880014
286318
$ cat r2test.sh 
#!/bin/bash

r2 -Qc 'aa
zg
e zign.bytes = false
?t z/
f ~sign.?' /usr/lib/libc.so.6
```

I am not sure why there is a discrepancy in hits between versions. All current tests pass.

**Test plan**

Current r2 zign tests pass. More could be added but I will have to look into where things are weak.

**Closing issues**

None

### DIFF
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -2332,242 +2332,179 @@ static bool fcnMetricsCmp(RSignItem *it, RAnalFunction *fcn) {
 	return true;
 }
 
-struct ctxFcnMatchCB {
-	RAnal *anal;
-	RAnalFunction *fcn;
-	RSignGraphMatchCallback cb;
-	void *user;
-	int mincc;
-};
-
-static int graphMatchCB(RSignItem *it, void *user) {
-	struct ctxFcnMatchCB *ctx = (struct ctxFcnMatchCB *) user;
+static bool graph_match(RSignItem *it, RSignSearchMetrics *sm) {
 	RSignGraph *graph = it->graph;
 
 	if (!graph) {
-		return 1;
+		return false;
 	}
 
-	if (graph->cc < ctx->mincc) {
-		return 1;
+	if (graph->cc < sm->mincc) {
+		return false;
 	}
 
-	if (!fcnMetricsCmp (it, ctx->fcn)) {
-		return 1;
+	if (!fcnMetricsCmp (it, sm->fcn)) {
+		return false;
 	}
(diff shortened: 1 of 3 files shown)


libr/anal/sign.c
```diff
@@ -2332,242 +2332,179 @@ static bool fcnMetricsCmp(RSignItem *it, RAnalFunction *fcn) {
...
+	return true;
 }
 
-R_API bool r_sign_match_addr(RAnal *a, RAnalFunction *fcn, RSignOffsetMatchCallback cb, void *user) {
-	r_return_val_if_fail (a && fcn && cb, false);
-	struct ctxFcnMatchCB ctx = { a, fcn, cb, user, 0 };
-	return r_sign_foreach (a, addrMatchCB, &ctx);
+static bool str_list_equals(RList *la, RList *lb) {
+	r_return_val_if_fail (la && lb, false);
+	int len = r_list_length (la);
```
trufae, line 2377: Sizet

libr/anal/sign.c
```diff
@@ -2332,242 +2332,179 @@ static bool fcnMetricsCmp(RSignItem *it, RAnalFunction *fcn) {
...
+	if (len != r_list_length (lb)) {
+		return false;
+	}
+	int i;
```
trufae, line 2381: Sizet

trufae (changes requested):
The amount of hits diffeence could be because your changes are only loading one type per zignature? Can you show the time difference?

swoops:
The time is hidden output I showed above. I used `?t z/` running on 3182 signatures for libc. The bytes search was disabled because it should not affect this. Time on original code was 774.342817, with the new code took 134.880014.

radare (approved):

PR #17788, opened 2020-10-15
Fix incorrect esil of RISC-V jalr insn ##emu

fabianfreyer:

**Detailed description**

The emitted esil for `jalr` was incorrect in the case where `rd` and `rs1` are equal, since then `rs1` is written to as `rd` before adding the immediate.

For example, `jalr ra ra 0x42` was lifted to the following esil:

      4,$$,+,ra,=,ra,66,+,pc,=

While this does correctly set rd to the return address after the call, `pc` is not correctly set according to `pc <- ra + 0x42`, but instead `pc <- pc + 4 + 66`.

This commit flips the order of setting `pc` and `rd`, so that `rd` is read from before it is written to.


fabianfreyer:
Together with #17787, this should correctly calculate jump targets generated from calls by `gcc` and `llvm`.

XVilka:
Could you please add a test for this?

fabianfreyer:
@XVilka uh... sure. Although I'm not really sure how to or what a good test case would be...

fabianfreyer:
Added a test that does this now:
```
            0x00000000      b7300100       lui ra, 0x13
            0x00000004      93800037       addi ra, ra, 880
            0x00000008      e780a002       jalr 42(ra)
```
This sets `ra` to `0x13370`, then calls `jalr 42(ra)`. After this, `ra` should point behind the `jalr` insn, i.e. `0xc`, and `pc` should be `0x13370 + 42`, so `0x1339a`.

trufae (approved):

PR #17792, opened 2020-10-16
Fixes for format string

XVilka:

**Detailed description**

Import fixes from https://github.com/radareorg/radare2/pull/17529

**Test plan**

All green

### DIFF
--- a/libr/anal/cc.c
+++ b/libr/anal/cc.c
@@ -12,7 +12,7 @@ R_API void r_anal_cc_del(RAnal *anal, const char *name) {
 	sdb_unset (DB, r_strbuf_setf (&sb, "cc.%s.ret", name), 0);
 	sdb_unset (DB, r_strbuf_setf (&sb, "cc.%s.argn", name), 0);
 	for (i = 0; i < R_ANAL_CC_MAXARG; i++) {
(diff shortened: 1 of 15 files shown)


ret2libc (approved):

libr/anal/cc.c
```diff
@@ -12,7 +12,7 @@ R_API void r_anal_cc_del(RAnal *anal, const char *name) {
 	sdb_unset (DB, r_strbuf_setf (&sb, "cc.%s.ret", name), 0);
 	sdb_unset (DB, r_strbuf_setf (&sb, "cc.%s.argn", name), 0);
 	for (i = 0; i < R_ANAL_CC_MAXARG; i++) {
-		sdb_unset (DB, r_strbuf_setf (&sb, "cc.%s.arg%d", name, i), 0);
+		sdb_unset (DB, r_strbuf_setf (&sb, "cc.%s.arg%zu", name, i), 0);
```
wargio, line 15: seems that `%zu` is not supported everywhere.

XVilka: @wargio it is supported in MSVC since 2013, so we can use it without any fear. MSVC 2017 still can produce binaries for Windows XP, thus all platforms we are interested in are covered. If there will be any particular bug reports from less mainstream compilers for arcane platforms we can fix it. Until then it should be fine as is.

wargio (approved):

PR #17798, opened 2020-10-18
Re-analyze function control flow after a write ##anal

### LINKED ISSUE #15644: Graph flow doesn't change when modifying instructions
### Work environment

| OS/arch/bits (mandatory)                             | Arch
| File format of the file you reverse (mandatory)      | Any
| Architecture/bits of the file (mandatory)            | Any
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.1.0-git 24001 @ linux-x86-64 git.4.0.0-226-g89a5320e5 commit: 89a5320e5f918327f6f528fa325eeef6567932a8 build: 2019-12-16__08:20:38

### Expected behavior

Graph flow should change after editing bytes or instructions according to the new value

### Actual behavior

The flow stays the same even when it doesn't makes sense according to the new disassembly

Before the change:
image

I will `nop`  0x00400609 and the flow will wrongly stay the same

image

- Open a binary with radare2 in write mode
- Perform analysis
- Open graph view and spot a conditional jump you want to modify or nop
- Modify it (e.j `je` -> `jne` or `nop`)
- Refresh graph and see that it stays the same

Binary: https://github.com/LiveOverflow/liveoverflow_youtube/raw/master/0x05_simple_crackme_intro_assembler/license_1

GustavoLCR:

**Detailed description**

* Re-analyze function control flow after a write
* Refresh graph after a write

**Test plan**

Test pass

**Closing issues**

Fixes #15644

### DIFF
--- a/libr/anal/block.c
+++ b/libr/anal/block.c
@@ -1,6 +1,7 @@
 /* radare - LGPL - Copyright 2019-2020 - pancake, thestr4ng3r */
 
 #include <r_anal.h>
+#include <r_hash.h>
 #include <ht_uu.h>
 
 #include <assert.h>
@@ -59,6 +60,9 @@ static RAnalBlock *block_new(RAnal *a, ut64 addr, ut64 size) {
 	block->parent_stackptr = INT_MAX;
 	block->cmpval = UT64_MAX;
 	block->fcns = r_list_new ();
+	if (size) {
+		r_anal_block_update_hash (block);
+	}
 	return block;
 }
 
@@ -241,6 +245,7 @@ R_API bool r_anal_block_relocate(RAnalBlock *block, ut64 addr, ut64 size) {
 	r_rbtree_aug_delete (&block->anal->bb_tree, &block->addr, __bb_addr_cmp, NULL, NULL, NULL, __max_end);
 	block->addr = addr;
 	block->size = size;
+	r_anal_block_update_hash (block);
 	r_rbtree_aug_insert (&block->anal->bb_tree, &block->addr, &block->_rb, __bb_addr_cmp, NULL, __max_end);
 	return true;
 }
@@ -271,6 +276,7 @@ R_API RAnalBlock *r_anal_block_split(RAnalBlock *bbi, ut64 addr) {
 	r_anal_block_set_size (bbi, addr - bbi->addr);
 	bbi->jump = addr;
 	bbi->fail = UT64_MAX;
+	r_anal_block_update_hash (bbi);
 
 	// insert the second block into the tree
 	r_rbtree_aug_insert (&anal->bb_tree, &bb->addr, &bb->_rb, __bb_addr_cmp, NULL, __max_end);
@@ -338,6 +344,7 @@ R_API bool r_anal_block_merge(RAnalBlock *a, RAnalBlock *b) {
 	a->size += b->size;
 	a->jump = b->jump;
 	a->fail = b->fail;
+	r_anal_block_update_hash (a);
 
 	// kill b completely
 	r_rbtree_aug_delete (&a->anal->bb_tree, &b->addr, __bb_addr_cmp, NULL, __block_free_rb, NULL, __max_end);
@@ -652,6 +659,41 @@ beach:
 	r_pvector_clear (&visit_a);
 	r_pvector_clear (&visit_b);
 	return ret;
+}
+
+R_API bool r_anal_block_was_modified(RAnalBlock *block) {
+	r_return_val_if_fail (block, false);
+	if (!block->anal->iob.read_at) {
+		return false;
+	}
+	ut8 *buf = malloc (block->size);
+	if (!buf) {
+		return false;
+	}
+	if (!block->anal->iob.read_at (block->anal->iob.io, block->addr, buf, block->size)) {
+		free (buf);
+		return false;
+	}
+	ut32 cur_hash = r_hash_xxhash (buf, block->size);
+	free (buf);
+	return block->bbhash != cur_hash;
+}
+
+R_API void r_anal_block_update_hash(RAnalBlock *block) {
+	r_return_if_fail (block);
+	if (!block->anal->iob.read_at) {
+		return;
+	}
+	ut8 *buf = malloc (block->size);
+	if (!buf) {
+		return;
+	}
+	if (!block->anal->iob.read_at (block->anal->iob.io, block->addr, buf, block->size)) {
+		free (buf);
+		return;
+	}
+	block->bbhash = r_hash_xxhash (buf, block->size);
+	free (buf);
 }
 
 typedef struct {
--- a/libr/anal/block.c
+++ b/libr/anal/block.c
@@ -781,6 +781,12 @@ typedef struct {
 	size_t cur_succ_count;
 } AutomergeCtx;
 
+static bool count_successors_cb(ut64 addr, void *user) {
+	AutomergeCtx *ctx = user;
+	ctx->cur_succ_count++;
+	return true;
+}
+
 static bool automerge_predecessor_successor_cb(ut64 addr, void *user) {
 	AutomergeCtx *ctx = user;
 	ctx->cur_succ_count++;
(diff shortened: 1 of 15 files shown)


trufae:
I got this request 2 years ago and i dont think this is the right way to implement this because its inneficient an inacurate:

* there are more ways to write/modify memory than the function you are catching
* writing in any address doesnt imply reanalizing the function
* Functions should have an associated bbsumhash, to detect when the function bytes has changed and suggest reanalisis
* Reanalizing will loose information if the user did some changes by hand, so this can break things
* add two tests, one writing inside, another writing outside the function
* add a config var disabled by default, and just warn the user when the fcnhash has changed

thoughs on this?

thestr4ng3r:
Yeah I agree calling `r_anal_fcn` after a write is way too much and probably discards too much existing info. Imo it should try to be smarter and detect exactly what has changed in the control flow and then update only what is necessary, a bit similar to how noreturn propagation chops up functions. This will be tricky but in the end the better solution.

GustavoLCR:
@trufae 
> there are more ways to write/modify memory than the function you are catching

Other than the write commands, I can only think of esil emu, that uses r_io directly, but I don't think it is very useful to automatically update analysis in that case

> Functions should have an associated bbsumhash, to detect when the function bytes has changed and suggest reanalisis

Wouldn't a `was_modified` bool do the trick if we add it to both `RAnalBlock` and `RAnalFunction`?
Also when/how should we warn the user?

thestr4ng3r:
There are many cases where the bytes can change without any specific action. The most important one is I think when something changes in the background in debug mode.

libr/core/cmd_write.c
```diff
@@ -740,7 +740,7 @@ static bool cmd_wfs(RCore *core, const char *input) {
 			done += rc;
 		}
 		r_socket_free (c);
-		if (r_io_write_at (core->io, core->offset, buf, done)) {
+		if (r_core_write_at (core, core->offset, buf, done)) {
```
trufae, line 743: I do t think @condret will be happy with this change

trufae (changes requested):
Maybe update aob and aobj outputs to include the bbhash? I think this was computed already somewhere else.. also, it will be better not to wrap riowrite, because we want the underlying apis to be used directly and interact with the whole core. Also what about having an option to toggle this feature?

XVilka:
Please fix the following alert:
```
static void update_analysis(RAnal *anal, RList *fcns, HtUP *reachable) {
--
                RAnal *anal = fcn->anal;
--
Local variable 'anal' hides a parameter of the same name.
```

libr/anal/block.c
```diff
@@ -59,6 +60,7 @@ static RAnalBlock *block_new(RAnal *a, ut64 addr, ut64 size) {
 	block->parent_stackptr = INT_MAX;
 	block->cmpval = UT64_MAX;
 	block->fcns = r_list_new ();
+	r_anal_block_update_hash (block);
```
trufae, line 63: uhm, ther's no data read at this point.. so not sure what should be updating here

libr/anal/block.c
```diff
@@ -654,6 +659,41 @@ R_API R_NULLABLE RList/*<RAnalBlock *>*/ *r_anal_block_shortest_path(RAnalBlock
...
+R_API void r_anal_block_update_hash(RAnalBlock *block) {
+	r_return_if_fail (block);
+	if (!block->anal->iob.read_at) {
+		return;
+	}
+	ut8 *buf = malloc (block->size);
+	if (!buf) {
+		return;
+	}
+	if (block->anal->iob.read_at (block->anal->iob.io, block->addr, buf, block->size) < 1) {
```
trufae, line 689: != block->size

libr/anal/fcn.c
```diff
@@ -2049,3 +2050,210 @@ R_API void r_anal_function_check_bp_use(RAnalFunction *fcn) {
...
+	}
+	ut64 len = to - from;
+	ut8 *buf = malloc (len);
+	if (!buf) {
+		return;
+	}
+	anal->iob.read_at (anal->iob.io, from, buf, len);
+	for (cur_addr = from; cur_addr < to; cur_addr += opsz, len -= opsz) {
+		RAnalOp op;
+		r_anal_op (anal->coreb.core, &op, cur_addr, buf, len, R_ANAL_OP_MASK_ESIL | R_ANAL_OP_MASK_VAL);
```
trufae, line 2111: check return value

libr/core/cmd_write.c
trufae: r_core_write_at should die. so dont use it. maybe adding hooks to rio is the way to go here

trufae (changes requested):

trufae:
Also im worried about the performance cost of this change. and a random though.. will be good to have a function hash omputation that we can probalby use for zignatures. but we have other techniques that may work better on fuzzy situations

GustavoLCR:
@trufae 
> Also im worried about the performance cost of this change.

before:
`r2 -e bin.dbginfo=false -Qc "aaa" libr/asm/libr_asm.so  37.80s user 0.56s system 99% cpu 38.380 total`
after:
`r2 -e bin.dbginfo=false -Qc "aaa" libr/asm/libr_asm.so  38.05s user 0.31s system 99% cpu 38.361 total`

So it seems basically negligible.


> and a random though.. will be good to have a function hash omputation that we can probalby use for zignatures. but we have other techniques that may work better on fuzzy situations

zignatures already does function hash, but it uses sha256, which is slower compared to xxhash

libr/core/cconfig.c
```diff
@@ -3286,6 +3303,7 @@ R_API int r_core_config_init(RCore *core) {
 	/* cfg */
 	SETBPREF ("cfg.r2wars", "false", "Enable some tweaks for the r2wars game");
 	SETBPREF ("cfg.plugins", "true", "Load plugins at startup");
+	SETCB ("cfg.analyzewrites", "false", &cb_cfganalyzewrites, "Re-analyze functions on writes");
```
ret2libc, line 3306: I would put it under `anal.something` as it is strictly analysis related.

wargio: `anal.update.on.change` is a terrible name.

wargio (approved):

XVilka (approved):

libr/io/io_cache.c
```diff
@@ -169,6 +169,9 @@ R_API bool r_io_cache_write(RIO *io, ut64 addr, const ut8 *buf, int len) {
 	}
 	memcpy (ch->data, buf, len);
 	r_list_append (io->cache, ch);
+	if (io->on_write) {
```
ret2libc, line 172: I think in theory an `REvent` would be better for these kind of things, as it allows more than one callback to be called when an event occurs.

libr/core/cconfig.c
```diff
@@ -1144,6 +1144,26 @@ static bool cb_binforce(void *user, void *data) {
 	return true;
 }
 
+static void rio_on_write_cb(ut64 addr, const ut8 *buf, int len, void *user) {
+	RCore *core = (RCore *)user;
+	r_anal_update_analysis_range (core->anal, addr, len);
+	if (core->cons->event_resize && core->cons->event_data) {
```
ret2libc, line 1150: why is this required?

libr/core/agraph.c
```diff
@@ -3488,6 +3488,15 @@ static int agraph_print(RAGraph *g, int is_interactive, RCore *core, RAnalFuncti
 	return true;
 }
 
+static void check_function_modified(RCore *core, RAnalFunction *fcn) {
+	if (r_anal_function_was_modified (fcn)) {
+		if (r_config_get_i (core->config, "cfg.analyzewrites")
```
ret2libc, line 3493: did you mean `anal.detectwrites`?

libr/core/cconfig.c
GustavoLCR: it forces an opened graph to update

trufae (approved):
LGTM!

we can enable this by default after some time+testing+discussions

The only thing i miss in this PR is that it will be good to add this bbhash in the output of `aob` (`aobj`...) maybe it could be useful for some usecases. But this can be done in a separate PR, what do you think?

Good work!

PR #17805, opened 2020-10-19
Add 6 minutes to cmd_aae timeout and skip aaa for swift_read fuzz test

kazarmy:

**Detailed description**

The pr is because:

1. The master asan build is still timing out on the cmd_aae test despite #17793.
2. I accidentally ran the asan+fuzz tests (probably because I used the pencil) and the following asan fuzz test is timing out (https://github.com/radareorg/radare2/runs/1274872621#step:15:20):

swift_read-fuzz

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
             meson_options: -Db_coverage=true -Duse_webui=true
             coverage: true
             enabled: true
-            timeout: 50
+            timeout: 60
             cflags: '-Werror -Wno-cpp'
           - name: macos-clang-tests
             os: macos-latest
--- a/binr/r2r/run.c
+++ b/binr/r2r/run.c
@@ -1117,9 +1117,14 @@ R_API void r2r_asm_test_output_free(R2RAsmTestOutput *out) {
 }
 
 R_API R2RProcessOutput *r2r_run_fuzz_test(R2RRunConfig *config, R2RFuzzTest *test, R2RCmdRunner runner, void *user) {
+#if ASAN
+	ut64 fuzz_timeout_ms = config->timeout_ms + 360 * 1000;
+#else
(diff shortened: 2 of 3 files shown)


trufae:
What about removing /replacing slow tests for something that runs fast even with asan? otherweise i think its preferibly to skip those tests in asan mode, such timeouts for a single test are not something good. 99% of the tests run in less than 2s, while 3 tests take 5min to run, we must squash those out, even if it's needed to test big binaries somehow, the tests should be more concrete

kazarmy:
> otherweise i think its preferibly to skip those tests in asan mode

sgtm for the fuzz test 641b4635c7d7210db4a00298cca92030339a16b0 ... for the cmd_aae test i think we should give that larger timeout a chance, so that we don't lose asan coverage for that test

kazarmy:
you're probably right though -- there's probably a combination of tests that individually normally pass their timeout thresholds yet in combination they take up enough time that when the server is under enough load, the build time goes past its timeout threshold ... squashing them would require time logging i think

trufae:
i have already identified them a month ago, fixed 2 of them, then lost interest on everything and stopped doing that, but shuold be easy to spot by running in local, doubt i have noted which were those tests sorry

kazarmy:
I think this pr should still be merged though. The problem with linux-meson-gcc-newshell-tests might be an artifact of having both "linux-meson-gcc-newshell-tests (push)" and "CI / linux-meson-gcc-newshell-tests (pull_request)", and the time of week. The asan+fuzz "linux-gcc-tests-asan (push)" build is running it close at 89 minutes but there's still another 10 minutes left.

kazarmy:
Ok I've bumped the linux-meson-gcc-newshell-tests timeout to 60 minutes, same as the macos-clang-tests timeout.

XVilka (approved):

PR #17821, opened 2020-10-22
fix msvc warning not enough args for macro

eagleoflqj:

**Detailed description**

https://ci.appveyor.com/project/radareorg/radare2/builds/35901972/job/ghq9uefp4vktn6wj?fullLog=true#L1912

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -2465,7 +2465,7 @@ static char *get_name(struct MACH0_(obj_t) *mo, ut32 stridx, bool filter) {
 }
 
 static int walk_exports(struct MACH0_(obj_t) *bin, RExportsIterator iterator, void * ctx) {
-#define ULEB(at) read_uleb128 (&p, end)
+#define ULEB() read_uleb128 (&p, end)
 	r_return_val_if_fail (bin, 0);
 	if (!bin->dyld_info) {
 		return 0;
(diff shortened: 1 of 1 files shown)


trufae (approved):

libr/bin/format/mach0/mach0.c
```diff
@@ -2465,7 +2465,7 @@ static char *get_name(struct MACH0_(obj_t) *mo, ut32 stridx, bool filter) {
 }
 
 static int walk_exports(struct MACH0_(obj_t) *bin, RExportsIterator iterator, void * ctx) {
-#define ULEB(at) read_uleb128 (&p, end)
+#define ULEB() read_uleb128 (&p, end)
```
trufae, line 2468: Uhm.. is this macro used at all? Because you changed the signature but the build didnt break

eagleoflqj: Yes it's used at https://github.com/radareorg/radare2/blob/master/libr/bin/format/mach0/mach0.c#L2509 and afterwards. But what surprised me is the original version was compiled successfully.

trufae: interesting. anyway. thanks!

XVilka:
Hiding those arguments is a bad idea, imho. It's better to get rid of this macro whatsoever. Just use `read_uleb128 (&p, end)` directly - it's not that big. Could you please do this instead?

ret2libc:
> Hiding those arguments is a bad idea, imho. It's better to get rid of this macro whatsoever. Just use `read_uleb128 (&p, end)` directly - it's not that big. Could you please do this instead? Same for `SLEB` macro.
> cc @ret2libc

I definitely agree with this. The function to call is not long at all and it is called in just few places.

XVilka (approved):

PR #17829, opened 2020-10-23
Remove R_CORE_AUTOCMPLT_END and fix UAF/crash

trufae:
```
cmd.c: In function 'cmd_autocomplete':
cmd.c:2151:29: warning: iteration 15 invokes undefined behavior [-Waggressive-loop-optimizations]
    if (autocomplete_flags[i].desc) {
        ~~~~~~~~~~~~~~~~~~~~~^~~~~
cmd.c:2150:3: note: within this loop
   for (i = 0; i < R_CORE_AUTOCMPLT_END; i++) {
   ^~~
```
 

**Your checklist for this pull request**
- [ ] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [ ] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

PR #17831, opened 2020-10-23
Fix UB in core.java plugin

trufae:
***SQUASH ME***
==============

```
libr/core/p/core_java.c: In function 'r_cmd_java_handle_help':
libr/core/p/core_java.c:390:34: warning: iteration 20 invokes undefined behavior [-Waggressive-loop-optimizations]
   help_msg[3 + (i * 3) + 0] = cmd->name;
                               ~~~^~~~~~
libr/..//libr/core/p/core_java.c:388:2: note: within this loop
  for (i = 0; i < END_CMDS; i++) {
```
 

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/core/p/core_java.c
+++ b/libr/core/p/core_java.c
@@ -272,13 +272,12 @@ enum {
 	REPLACE_CLASS_NAME_IDX = 12,
 	RELOAD_BIN_IDX = 13,
 	SUMMARY_INFO_IDX = 14,
-	LIST_CODE_REFS_IDX = 15,
-	PRINT_EXC_IDX = 16,
-	YARA_CODE_REFS_IDX = 17,
-	INSERT_MREF_IDX = 18,
-	CALC_SZ_IDX = 19,
-	ISVALID_IDX = 20,
-	END_CMDS = 21,
+	PRINT_EXC_IDX = 15,
+	YARA_CODE_REFS_IDX = 16,
+	INSERT_MREF_IDX = 17,
+	CALC_SZ_IDX = 18,
+	ISVALID_IDX = 19,
+	END_CMDS = 20,
 };
 
 static ut8 _(r_cmd_java_obj_ref)(const char *name, const char *class_name, ut32 len) {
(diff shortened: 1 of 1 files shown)


ret2libc (changes requested):
Just remove `LIST_CODE_REFS_IDX = 15,` from the enum and shift all the values after that (or even better, just remove all the values apart from `= 0`).

trufae:
I think its safer to expect null items at the end of the array than syncing it's size with the enum and such.

ret2libc:
You are not really solving anything by putting `NULL, NULL, NULL` at the end of the array. The array is not traversed until the null item, it is traversed until `END_CMDS`. And the issue is that the enum and the array are not in sync, so if you don't remove the entry i said in https://github.com/radareorg/radare2/pull/17831#pullrequestreview-515929603 all commands after `15` will be wrong.

trufae:
wtf seems like some of my fixes got lost, this pr is incomplete. let me tweak it back

ret2libc (changes requested):
This does not work well anyway, because some of the enum values are used to access the JAVA_CMDS array, so they have to be adjusted no matter what. Please remove the `LIST_CODE_REFS_IDX = 15` and shift all values after that.

XVilka:
@wargio the same pattern here, as in https://github.com/radareorg/radare2/pull/17829

wargio:
> `LIST_CODE_REFS_IDX`

Also remove lines from 215 to 218
```c
#define LIST_CODE_REFS "lcr"
#define LIST_CODE_REFS_ARGS " [addr]"
#define LIST_CODE_REFS_DESC "list all references to fields and methods in code sections"
#define LIST_CODE_REFS_LEN 3
```

trufae:
This issue was introduced in 3ead80cca8a9e03ce4c755718d6772cf583d5716 i think the custom analysis of this plugin was more precise, but it's true that the default analysis should be the one handling everything in here. so im fine for removing this as long as the rest of the code was already removed.

wargio (approved):

ret2libc (approved):

PR #17832, opened 2020-10-23
Fix UB in r_core_cmd_init

trufae:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -7178,7 +7178,7 @@ R_API void r_core_cmd_init(RCore *core) {
 	for (i = 0; i < R_ARRAY_SIZE (cmds); i++) {
 		r_cmd_add (core->rcmd, cmds[i].cmd, cmds[i].cb);
 
-		RCmdDesc *cd;
+		RCmdDesc *cd = NULL;
 		switch (cmds[i].type) {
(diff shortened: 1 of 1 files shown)


ret2libc (changes requested):
There is no UB and no bug. It is just a false positive. `cd` cannot be NULL because all enum values are considered in the switch and if you don't consider all values the compiler spits out a warning about a missing enum value in the switch.

trufae:
The warning is this:

```
cmd.c: In function 'r_core_cmd_init':
cmd.c:7197:4: warning: 'cd' may be used uninitialized in this function [-Wmaybe-uninitialized]
    cmds[i].descriptor_init (core, cd);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

By reading the code i see that type is an enum so it must be handling all the cases of the switch, so indeed, the cd variable must be initialized.. the problem is that the callbacks can return null, which may result on cd=null, and therefor there's a bug in the call, which is fixed by the `if (cd && `.. about the uniitalized i think its a false positive from the compiler, but i think it doesnt hurts to initialize it

ret2libc (approved):
`new` can indeed report NULL, right.

XVilka (approved):

PR #17833, opened 2020-10-23
Fix UB in cannotated_code

trufae:
```
cannotated_code.c: In function 'r_core_annotated_code_print_json':
cannotated_code.c:80:4: warning: 'type_str' may be used uninitialized in this function [-Wmaybe-uninitialized]
    pj_ks (pj, "syntax_highlight", type_str);
```
 

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/core/cannotated_code.c
+++ b/libr/core/cannotated_code.c
@@ -51,6 +51,7 @@ R_API void r_core_annotated_code_print_json(RAnnotatedCode *code) {
 			break;
 		case R_CODE_ANNOTATION_TYPE_SYNTAX_HIGHLIGHT:
 			pj_ks (pj, "type", "syntax_highlight");
+			type_str = NULL;
 			switch (annotation->syntax_highlight.type) {
 			case R_SYNTAX_HIGHLIGHT_TYPE_KEYWORD:
 				type_str = "keyword";
@@ -77,7 +78,9 @@ R_API void r_core_annotated_code_print_json(RAnnotatedCode *code) {
 				type_str = "global_variable";
 				break;
 			}
-			pj_ks (pj, "syntax_highlight", type_str);
+			if (type_str) {
+				pj_ks (pj, "syntax_highlight", type_str);
+			}
 			break;
 		}
 		pj_end (pj);


ret2libc (commented):
What system/compiler are you using for this? I think the code is actually fine and the only case where type_str is not initialize is if you are missing a value in the switch, but in that case the compiler should provide another warning about a missing enum value. I think this is definitely a false positive.

trufae:
Did you answered using a mail client? Because the html response was fuckedup.

Refering your question. Its ubuntu20

ret2libc:
> Did you answered using a mail client? Because the html response was fuckedup.

No, I did not. I replied as always.

> Refering your question. Its ubuntu20

Do these warnings appear by default or did you enable any particular cflag?

trufae:
by default i didnt applied any special flag

ret2libc:
@trufae could you tell me exactly how you did that? I've just tried on a ubuntu:groovy container by doing `./sys/install.sh` and also `./configure;make` but none of these reported this warning (nor the others).

trufae:
It's ubuntu:focal in gitlab's CI, didnt tried in local, but those are real warnings. Do youw ant the full log?

the one for this issue is:

```
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cannotated_code.c: In function 'r_core_annotated_code_print_json':
cannotated_code.c:80:4: warning: 'type_str' may be used uninitialized in this function [-Wmaybe-uninitialized]
    pj_ks (pj, "syntax_highlight", type_str);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

wargio (commented):
What about adding default to that switch that initialize the `type_str` to `NULL` ?

ret2libc:
> What about adding default to that switch that initialize the `type_str` to `NULL` ?

I personally prefer to not do that, and I think this is a false positive (as explained above). Given all values are handled in the switch, `type_str` should never be null. The reason why I personally prefer to not add a default value is that if you add a new value in the enum, the compiler will warn you that there is a value not handled in the switch. If you provide the default case, the compiler will not provide such a warning and you will probably end up forgetting about it.

Said that, I'm okish with this fix if it makes some compiler happier, but it is not a real bug IMO so no "Fix UB in cannotated code".

trufae:
I think it's safer this way in case that more cases are added and not handled, it wont be printing trash. But yeah, i agree that it's not a real bug. this patch just make the code safer. Feel free to decide what to do. I'm also happy to remove compiler warnings even if those are bugs in their side

ret2libc:
In any case I think you could use type_str ="" instead of null, so you don't have to put a useless if.

trufae:
The if is on purpose to not add this field in case the highlight type info is not available. Just to try to avoid cluttering the output. im fine for doing the change.

wargio (approved):

PR #17834, opened 2020-10-23
Fix UB in r_core_anal_type_match

trufae:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/core/anal_tp.c
+++ b/libr/core/anal_tp.c
@@ -471,8 +471,8 @@ R_API void r_core_anal_type_match(RCore *core, RAnalFunction *fcn) {
 	if (!hc) {
 		return;
 	}
-	RDebugTrace *dt;
-	RAnalEsilTrace *et;
+	RDebugTrace *dt = NULL;
+	RAnalEsilTrace *et = NULL;
 	if (!anal_emul_init (core, hc, &dt, &et) || !fcn) {
 		anal_emul_restore (core, hc, dt, et);
 		return;


ret2libc (commented):
I'm not sure how there could be UB here, as both `dt` and `et` are initialized in `anal_emul_init` if `core->anal->esil` is not NULL. And if we reach this point in this function it already means `core->anal->esil` is not NULL.

trufae:
some context: 
```
   ^~~~~~~~~~~~~~
anal_tp.c: In function 'r_core_anal_type_match':
anal_tp.c:477:3: warning: 'et' may be used uninitialized in this function [-Wmaybe-uninitialized]
   anal_emul_restore (core, hc, dt, et);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
anal_tp.c:477:3: warning: 'dt' may be used uninitialized in this function [-Wmaybe-uninitialized]
cannotated_code.c: In function 'r_core_annotated_code_print_json':
cannotated_code.c:80:4: warning: 'type_str' may be used uninitialized in this function [-Wmaybe-uninitialized]
    pj_ks (pj, "syntax_highlight", type_str);
```

ret2libc:
As explained in https://github.com/radareorg/radare2/pull/17834#pullrequestreview-515915142, I think this is yet another false positive. So, I'm ok with the change, but I'm not ok with all the claimed "Fix UB" where there is no UB at all.

trufae:
i dont think it's such a big issue to just initialize few vars to avoid those warnings. but its true that the only caller is ensuring this esil &&fcn nonnull. so it looks like a fals positive to me. maybe a compiler bug, i will add compiler version info in the build logs

wargio (approved):

wargio:
I'm always OK with initializing the vars.

ret2libc:
> i dont think it's such a big issue to just initialize few vars to avoid those warnings. but its true that the only caller is ensuring this esil &&fcn nonnull. so it looks like a fals positive to me. maybe a compiler bug, i will add compiler version info in the build logs

As said, I'm ok with the change, but there is no "UB in r_core_anal_type_match".

PR #17847, opened 2020-10-25
Attach to target pid/tid on remote lldb connect ##debug

yossizap:

**Detailed description**

LLDB's connect implementation didn't properly handle attaching to the current pid/tid and setting them for step/continue operations. gdbr's implementation is shared with lldb now.

**Test plan**

* Run lldb-server with:

```
lldb-server-10 g localhost:1234 -- test_binary
```

* Run the following command in r2 to connect to the server:

```
oodf gdb://127.0.0.1:1234
```

* Check that the registers are synchronized, step/continue/breakpoints work and that `dp` and `dpt` show the correct thread+process.

### DIFF
--- a/shlr/gdb/src/gdbclient/core.c
+++ b/shlr/gdb/src/gdbclient/core.c
@@ -233,8 +233,9 @@ int gdbr_connect(libgdbr_t *g, const char *host, int port) {
 		}
 	}
 	if (g->remote_type == GDB_REMOTE_TYPE_LLDB) {
-		ret = gdbr_connect_lldb (g);
-		goto end;
+		if ((ret = gdbr_connect_lldb (g)) < 0) {
+			goto end;
+		}
 	}
 	// Query the thread / process id
 	g->stub_features.qC = true;
@@ -249,7 +250,7 @@ int gdbr_connect(libgdbr_t *g, const char *host, int port) {
 	// Check if vCont is supported
 	gdbr_check_vcont (g);
 	// Set pid/thread for operations other than "step" and "continue"
-	if (gdbr_select (g, g->pid, 0) < 0) {
+	if (gdbr_select (g, g->pid, g->tid) < 0) {
 		// return -1;
 	}
 	// Set thread for "step" and "continue" operations


shlr/gdb/src/gdbclient/core.c
```diff
@@ -249,7 +250,7 @@ int gdbr_connect(libgdbr_t *g, const char *host, int port) {
 	// Check if vCont is supported
 	gdbr_check_vcont (g);
 	// Set pid/thread for operations other than "step" and "continue"
-	if (gdbr_select (g, g->pid, 0) < 0) {
+	if (gdbr_select (g, g->pid, g->tid) < 0) {
```
trufae, line 253: maybe attach to pid 0 in case tid fails? i think some gdb server implementations just ignore or only handle pid0 when debugging in real mode. can someone confirm this?

yossizap: I assume that gdbserver implementations that only handle pid0 will return false in the [`multiprocess`](https://sourceware.org/gdb/current/onlinedocs/gdb/General-Query-Packets.html#multiprocess-extensions) flag which is already handled by `write_thread_id`(used in `gdbr_select` with `Hg`) by only setting the thread id:

https://github.com/radareorg/radare2/blob/89cd9ec98374aabd85d951923618b2f44f9fbd5a/shlr/gdb/src/utils.c#L139-L145
(same as the implementation in gdb's `write_ptid`: https://github.com/bminor/binutils-gdb/blob/02855c3a546243f4a0e48dc207ebb1f9e0e04545/gdb/remote.c#L2962-L2983)

Since the thread id is returned by [`qC`](https://sourceware.org/gdb/current/onlinedocs/gdb/General-Query-Packets.html#index-qC-packet) in `gdbr_connect` it may also be 0 if the target doesn't have multiple threads. It seems that gdb's `set_thread` implementation matches this assumption: 
https://github.com/bminor/binutils-gdb/blob/02855c3a546243f4a0e48dc207ebb1f9e0e04545/gdb/remote.c#L2758-L2785

`magic_null_ptid` is set when there's no tid, `any_thread_ptid` and `minus_one_ptid` seem to be manually set to interact with all threads.

Wouldn't mind testing it if you know of a limited gdbserver implementation that might reinforce this theory.

trufae: I think avarice was one of them (the AVR emulator) that spawns a gdb server. but I may be wrong. i'm fine for merging this in case test works.

XVilka (approved):

shlr/gdb/src/gdbclient/core.c
yossizap: This [avarice](http://avarice.sourceforge.net/)? It only works with JTAG so I couldn't set it up with my Arduino but it seems that their implementation doesn't even support `Hg`(See `talkToGdb` in src/remote.cc) so our selection will simply be ignored. I tried testing with simulavr but it only works with `gdb-avr`, closes itself when connecting with r2/regular gdb.

Anyways, I don't think it's problem since this is the same implementation as before for non-multiprocess targets and we didn't receive any issues about it.

trufae (approved):

trufae:
Thanks for confirming and testing! Lgtm now!

PR #17815, opened 2020-10-21
Fix unexpected FPU exception in ESIL emulation ##esil

trufae:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

IMAGE 2020-10-21 17:57:26

**Test plan**

We need OVF macros for signed/unsigned and zero division checks to cover all possible r2 fpu exceptions.

* i cant share the binary, but its easy to reproduce with this oneliner. But the signed instructions were not documented (ae??)

```
[0x00000000]> "ae -1,0x8000000000000000,~/"
zsh: floating point exception  r2 -
$
```

**Closing issues**

PR #17856, opened 2020-10-27
Add division overflow macros in r_types ##util

trufae:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/anal/esil.c
+++ b/libr/anal/esil.c
@@ -1540,18 +1540,6 @@ static bool esil_mod(RAnalEsil *esil) {
 	return ret;
 }
 
-static bool detect_fpu_div_exception(ut64 a, ut64 b) {
-	// division by zero
-	if (b == UT64_MIN) {
-		return true;
-	}
-	// undefined result (0x80000 / -1) cant be represented
-	if (a == UT64_GT0 && b == UT64_MAX) {
-		return true;
-	}
-	return false;
-}
-
 static bool esil_signed_mod(RAnalEsil *esil) {
 	bool ret = false;
 	st64 s, d;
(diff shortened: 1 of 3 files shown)


trufae:
This PR add generic macros to use across the whole r2 codebase. Improving the state of https://github.com/radareorg/radare2/pull/17815

wargio:
having functions on .h files, is not the best. especially if you include the file twice.

ret2libc:
> having functions on .h files, is not the best. especially if you include the file twice.

It is a common way to use functions instead of macros, very useful for small pieces of code. So you still avoid calling functions (which has a cost) + you have the advantages of function type checking, easy of writing, etc.

wargio:
> > having functions on .h files, is not the best. especially if you include the file twice.
> 
> It is a common way to use functions instead of macros, very useful for small pieces of code. So you still avoid calling functions (which has a cost) + you have the advantages of function type checking, easy of writing, etc.

Nono. I'm fully aware of that, is just i do not like those functions to be defined in a header instead of a c file.

The issue I'm referring is `multiple declaration of xxxxx function` which happens when you include the same header file in multiple files even when you have ifndef/define/endif

ret2libc:
> > > having functions on .h files, is not the best. especially if you include the file twice.
> > 
> > 
> > It is a common way to use functions instead of macros, very useful for small pieces of code. So you still avoid calling functions (which has a cost) + you have the advantages of function type checking, easy of writing, etc.
> 
> Nono. I'm fully aware of that, is just i do not like those functions to be defined in a header instead of a c file.

If you want to avoid the overhead of functions calling, etc. you *must* define them in the header AFAIK.

> 
> The issue I'm referring is `multiple declaration of xxxxx function` which happens when you include the same header file in multiple files even when you have ifndef/define/endif

AFAIK if you include a file like this in multiple files you should not get that error/warning, because the function is `static`, so it is redefined (separately) for each file that includes the header.

XVilka (approved):

wargio (approved):

PR #17890, opened 2020-11-02
Add x86-16 tests for the operand-size prefix

ramikg:
In real mode – make sure the operand-size override prefix (0x66) is added when, and only when, necessary.

**Detailed description**

In x86 real mode, it is possible to access 32-bit registers by prefixing the instruction with 66h. This is exemplified in this PR's commit.
I've added tests that make sure that the operand-size prefix is added when needed, and not added otherwise.

### DIFF
--- a/test/db/asm/x86_16
+++ b/test/db/asm/x86_16
@@ -10,10 +10,10 @@ dB "jmp 0xfec50" e95bec f000:fff2
 d "jmp 0x1fec50" e95bec 0x001ffff2
 a "mov al, [0xbeef]" a0efbe
 a "mov ax, [0xbeef]" a1efbe
-a "mov bx, ax" 89c3
-a "mov ebx, eax" 6689c3
-a "sub sp, 2" 83ec02
-a "sub esp, 4" 6683ec04
+aB "mov bx, ax" 89c3
+aB "mov ebx, eax" 6689c3
+aB "sub sp, 2" 83ec02
+aB "sub esp, 4" 6683ec04
 a "test bl, 0x12" f6c312
 a "test bx, 0x1234" f7c33412
 aB "test byte [bx], 0x12" f60712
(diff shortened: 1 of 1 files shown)


trufae:
Screenshot 2020-11-11 at 16 41 36

ramikg:
Indeed, _asm.nz_ fails the new tests, as it doesn't apply the operand-size prefix correctly (as opposed to _asm.as_ & _asm.nasm_).

Hopefully I'll find the time to go through _asm.nz_'s code.

ramikg:
Indeed, _asm.nz_ fails the new tests, as it doesn't apply the operand-size prefix correctly (as opposed to _asm.as_ & _asm.nasm_).

I'll open an issue about this.

trufae:
I would prefer to update this pr to support these indtructions in .nz or mark them as broken as last option. No need to open an issue if you are planning to implement that. Nz code is quite straight forward. Let me know if you have any question.

trufae:
ping

test/db/asm/x86_16
```diff
@@ -10,6 +10,10 @@ dB "jmp 0xfec50" e95bec f000:fff2
 d "jmp 0x1fec50" e95bec 0x001ffff2
 a "mov al, [0xbeef]" a0efbe
 a "mov ax, [0xbeef]" a1efbe
+a "mov bx, ax" 89c3
+a "mov ebx, eax" 6689c3
+a "sub sp, 2" 83ec02
```
trufae, line 15: Use aB instead of a in all those 4 failing tests

trufae (changes requested):
Use aB instead of a in all those 4 failing tests

ramikg:
Changed to `aB`, but unfortunately I won't be able to update the PR or open an issue any time soon.

PR #17891, opened 2020-11-03
Use `r_str_cpy` and `snprintf` in `parse_att2intel.c` ##refractor

officialcjunior:

**Detailed description**

While working on #17777, I [was suggested](https://github.com/radareorg/radare2/pull/17777#discussion_r512959073) to replace `sprintf` with `snprintf` to avoid overflow. I've applied the same on `parse_att2intel.c` on this PR.

I've also used `r_str_cpy` on an instance of `memmove` and added some missing spaces .

**Test plan**

See if any CI fails and fix anything that went unnoticed.

**Closing issues**

None

### DIFF
--- a/libr/parse/p/parse_att2intel.c
+++ b/libr/parse/p/parse_att2intel.c
@@ -42,8 +42,8 @@ static int replace(int argc, const char *argv[], char *newstr) {
 					if (ops[i].str[j]>='0' && ops[i].str[j]<='9') {
 						const char *w = argv[ ops[i].str[j]-'0' ];
 						if (w != NULL) {
-							strcpy(newstr+k, w);
-							k += strlen(w)-1;
+							strcpy (newstr + k, w);
+							k += strlen (w) - 1;
 						}
 					} else {
 						newstr[k] = ops[i].str[j];
@@ -105,16 +105,17 @@ static int parse(RParse *p, const char *data, char *str) {
 			num = (char *)r_str_lchr (buf, ',');
 		}
 		if (num) {
-			n = atoi (num+1);
+			n = atoi (num + 1);
 			*ptr = '[';
-			memmove (num+1, ptr, strlen (ptr)+1);
+			r_str_cpy (num + 1, ptr);
 			ptr = (char*)r_str_lchr (buf, ']');
 			if (n && ptr) {
 				char *rest = strdup (ptr+1);
+				size_t dist = strlen (data) + 1 - (ptr - buf);
 				if (n > 0) {
-					sprintf (ptr, "+%d]%s", n, rest);
+					snprintf (ptr, dist, "+%d]%s", n, rest);
 				} else {
-					sprintf (ptr, "%d]%s", n, rest);
+					snprintf (ptr, dist, "%d]%s", n, rest);
 				}
 				free (rest);
 			}


XVilka:
It's `refactor` not `refractor`.

eagleoflqj:
I would suggest use only one `snprintf` with `"%+d"` to handle both postive and negative `n`, as `n` is not 0.

officialcjunior:
@trufae, what do you think?

trufae:
sgtm, i like @eagleoflqj , according to the msvc webpage it seems to be also supported on Windows, so go for it

trufae (changes requested):
sgtm, i like @eagleoflqj , according to the msvc webpage it seems to be also supported on Windows, so go for it

officialcjunior:
Alright, great.
So, we don't need the `if` condition, right?

eagleoflqj (changes requested):
`%+d` instead of `+%d`

libr/parse/p/parse_att2intel.c
```diff
@@ -42,8 +42,8 @@ static int replace(int argc, const char *argv[], char *newstr) {
 					if (ops[i].str[j]>='0' && ops[i].str[j]<='9') {
 						const char *w = argv[ ops[i].str[j]-'0' ];
```
trufae, line 43: proper spacing around - and <= operators across the whole file, not just the lines you changed there.

libr/parse/p/parse_att2intel.c
```diff
@@ -42,8 +42,8 @@ static int replace(int argc, const char *argv[], char *newstr) {
...
 						if (w != NULL) {
-							strcpy(newstr+k, w);
-							k += strlen(w)-1;
+							strcpy (newstr + k, w);
```
trufae, line 45: you can compute the strlen(w) before the strcpy. and just use memmove, to avoid computing the length twice.

also, is newstr able to contain `w` ? strcpy is usually a very bad idea to use

trufae (changes requested):

libr/parse/p/parse_att2intel.c
officialcjunior: > you can compute the strlen(w) before the strcpy. and just use memmove, to avoid computing the length twice.

Alright, so are you suggesting to use `memmove` instead of `strcpy`?
```
                                                 k = strlen (w) - 1;
                                                 memmove (newstr, w, k);
```

trufae (approved):

libr/parse/p/parse_att2intel.c
trufae: that wont copy the null byte and it is not checking the newstr size, so it wont change the situation.

trufae (changes requested):

libr/parse/p/parse_att2intel.c
eagleoflqj: > you can compute the strlen(w) before the strcpy. and just use memmove, to avoid computing the length twice.
> 
> also, is newstr able to contain `w` ? strcpy is usually a very bad idea to use

There is no way to check buffer overflow in this function. Do you want a change to the function signature to include buffer size?

trufae:
ping

officialcjunior:
> > you can compute the strlen(w) before the strcpy. and just use memmove, to avoid computing the length twice.
> > also, is newstr able to contain `w` ? strcpy is usually a very bad idea to use
> 
> There is no way to check buffer overflow in this function. Do you want a change to the function signature to include buffer size?

I hoped that you'll reply to @eagleoflqj's comment. I was waiting for that. You seem to have missed that?

trufae:
Oh yes i missed that. Yep there must be a boundary check in every buffer/string handling usage. So yes. Any pointer to a buffer should come with the length

officialcjunior:
I've been trying to wrap my head around it. I think have understood what needs to be done, here.

Okay so, when `replace()` is called from `parse()`, it should have the length of `str`/`newstr` too.
Can we just calculate that somehow in `parse()`?

Also, isn't it a part of `r_parse_plugin`? That means, dependent on the user?

libr/parse/p/parse_att2intel.c
trufae: yes

trufae (changes requested):
Please update the function signature

officialcjunior:
Understood. How should the length be calculated inside `parse()` so that it can be send over to `replace()`?
I've been looking around for that. chip8 pseudo had hardcoded the values for that.
‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

trufae (commented):
The way rparser is done right now i think the change is out of the scope of this PR. i have a half written refactoring of this library, and i think is better to keep it like this until that refactoring hits the repo. So i would say, i'm fine with the current state of the PR even if unsafe.

PR #17892, opened 2020-11-03
r_str_append always returns copy

kazarmy:

**Detailed description**

This pr might fix the AppVeyor build. If not, it should be merged in anyway for consistency.

**Test plan**

All builds are green.

**Closing issues**

...

### DIFF
--- a/libr/core/core.c
+++ b/libr/core/core.c
@@ -3492,9 +3492,7 @@ R_API char *r_core_editor(const RCore *core, const char *file, const char *str) 
 		cons->cb_editor = tmp;
 	} else {
 		if (editor && name) {
-			char *escaped_name = r_str_escape_sh (name);
-			r_sys_cmdf ("%s \"%s\"", editor, escaped_name);
-			free (escaped_name);
+			r_sys_cmdf ("%s '%s'", editor, name);
 		}
 	}
 	size_t len = 0;
--- a/libr/core/core.c
+++ b/libr/core/core.c
@@ -3492,7 +3492,9 @@ R_API char *r_core_editor(const RCore *core, const char *file, const char *str) 
 		cons->cb_editor = tmp;
 	} else {
 		if (editor && name) {
-			r_sys_cmdf ("%s '%s'", editor, name);
+			char *escaped_name = r_str_escape_sh (name);
+			r_sys_cmdf ("%s \"%s\"", editor, escaped_name);
+			free (escaped_name);
 		}
 	}
 	size_t len = 0;
--- a/libr/util/file.c
+++ b/libr/util/file.c
@@ -1084,7 +1084,7 @@ R_API int r_file_mkstemp(R_NULLABLE const char *prefix, char **oname) {
 		char *name_ = r_sys_conv_win_to_utf8 (name);
 		name_[strlen (name_) - 4] = '\0';
 		name_ = r_str_append (name_, suffix);
-		h = r_sandbox_open (name_, O_RDWR|O_CREAT|O_EXCL|O_BINARY, 0644);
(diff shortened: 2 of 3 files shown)


ret2libc (changes requested):
I think this is wrong.... who is going to free `ptr`?

kazarmy:
The caller, and it's better than a UAF.

kazarmy:
And `realloc` also requires that the caller free `ptr`.

ret2libc:
> The caller, and it's better than a UAF.

AFAIK the expected usage of this API is:
```C
str = malloc(10);
// str is filled with a string
strcpy (str, "world");
str = r_str_prepend(str, "hello ");
```

so the behaviour should be consistent. If you expect the caller to do the free (which I don't think is the right assumption for what there is so far) it should be done consistently, not only for the particular cause of empty `append string`.

> And `realloc` also requires that the caller free `ptr`.

Not that I'm aware of. The `realloc` just reallocs, if necessary. When reallocs fail, r_str_prepend already frees `ptr`, so the user does not need to do it.

Where is the UAF?

kazarmy:
The current code of `r_str_append` on master is the following:

https://github.com/radareorg/radare2/blob/a26f6c56567fee48f1657b4c666bef6d41d6716f/libr/util/str.c#L880-L901

- Line 886 returns a `strdup`'d string, which needs to be freed by the caller.
- Line 900 returns a pointer that was returned by `realloc`, courtesy of line 898 (`ptr = newptr`). This pointer needs to be freed by the caller.
- Line 889 returns the original `ptr`, which can't be freed since it might be a string literal that is stored in read-only storage.

See the inconsistency?

----

> > And realloc also requires that the caller free ptr.

> Not that I'm aware of. The realloc just reallocs, if necessary. When reallocs fail, r_str_prepend already frees ptr, so the user does not need to do it.

https://en.cppreference.com/w/c/memory/realloc states for`realloc`'s Return value:

"On success, returns the pointer to the beginning of newly allocated memory. **To avoid a memory leak, the returned pointer must be deallocated with free()** or realloc(). The original pointer ptr is invalidated and any access to it is undefined behavior **(even if reallocation was in-place)**." (emphasis mine)

----

> Where is the UAF?

Emm the following code just frees the return value of `r_str_append`, even though `suffix` can be the empty string:

https://github.com/radareorg/radare2/blob/a26f6c56567fee48f1657b4c666bef6d41d6716f/libr/util/file.c#L1086-L1097

Ok this is technically not a UAF, but I can't repro the AppVeyor problem locally so the guess here is that there is memory corruption, or there is some AppVeyor configuration that is causing this.

ret2libc:
I think the whole problem is:

> Line 889 returns the original ptr, which can't be freed since it might be a string literal that is stored in read-only storage.

which is a wrong usage of the API due to https://github.com/radareorg/radare2/blob/a26f6c56567fee48f1657b4c666bef6d41d6716f/libr/util/str.c#L881

As said, the usage should be:
```C
str = malloc(10);
strcpy (str, "world");
// NOTE: until this point, this function MUST free str
str = r_str_prepend(str, "hello ");
// NOTE: even after r_str_prepend, this function MUST free str, because it was:
// 1) strdup of "hello "
// 2) realloced by r_str_prepend, so once again it must be freed by this function
// 3) `str` itself is returned, which was supposed to be freed by this function in the first place, so it still needs to be freed by this function.
```

kazarmy:
> which is a wrong usage of the API due to
> https://github.com/radareorg/radare2/blob/a26f6c56567fee48f1657b4c666bef6d41d6716f/libr/util/str.c#L881

Ok.

Ok I have finally realized that `realloc` frees the original pointer and thus finally understood what you are trying to say so I will close this pr.

PR #17896, opened 2020-11-06
Minor optimizations to cmd_print.c

eagleoflqj:

**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -6859,7 +6859,6 @@ R_API void r_print_offset_sg(RPrint *p, ut64 off, int invert, int offseg, int se
 	char space[32] = {
 		0
 	};
-	const char *white;
 	const char *reset = p->resetbg? Color_RESET: Color_RESET_NOBG;
 	bool show_color = p->flags & R_PRINT_FLAGS_COLOR;
 	if (show_color) {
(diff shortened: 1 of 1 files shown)


libr/core/cmd_print.c
```diff
@@ -6946,8 +6941,7 @@ R_API void r_print_offset_sg(RPrint *p, ut64 off, int invert, int offseg, int se
 			} else {
 				if (offdec) {
 					snprintf (space, sizeof (space), "%"PFMT64u, off);
-					white = r_str_pad (' ', 10 - strlen (space));
-					r_cons_printf ("%s%s", white, space);
+					r_cons_printf ("%10s", space);
```
trufae, line 6944: Align left vs align right?

trufae (changes requested):

libr/core/cmd_print.c
eagleoflqj: As https://en.cppreference.com/w/c/io/fprintf says, without `-` after `%`, align right is applied.
```
-: the result of the conversion is left-justified within the field (by default it is right-justified)
(optional) integer value or * that specifies minimum field width. The result is padded with space characters (by default), if required, on the left when right-justified, or on the right if left-justified. In the case when * is used, the width is specified by an additional argument of type int. If the value of the argument is negative, it results with the - flag specified and positive field width. (Note: This is the minimum width: The value is never truncated.)
```

XVilka: We should check this behavior on Windows too, just in case, before merging. cc @GustavoLCR @kazarmy @pelijah @a1ext

eagleoflqj: FYI https://rextester.com/l/c_online_compiler_visual

XVilka: Yes, it works as intended.

XVilka (approved):

wargio (approved):
LGTM.

trufae (approved):

XVilka:
@ret2libc please take one more look, if you are good - lets merge.

ret2libc (approved):

Issue #17898, opened 2020-11-06
ARM32 build warnings


eagleoflqj:
On my rpi4 (Linux liumeo-rpi4 5.4.79-v7l+ #1373 SMP Mon Nov 23 13:27:40 GMT 2020 armv7l GNU/Linux), all test errors fixed now.
```
pi@liumeo-rpi4:~/github/radare2/test $ r2 -v
radare2 4.6.0-git 25454 @ linux-arm-32 git.
commit: 8c742b28f089385cd76fd6b7a35cb658eb911237 build: 2020-12-20__14:25:02
pi@liumeo-rpi4:~/github/radare2/test $ r2r -L
Skipping db/archos/linux-x64 because it does not match the current platform.
Skipping db/archos/darwin-x64 because it does not match the current platform.
Skipping db/archos/windows-x64 because it does not match the current platform.
Skipping db/extras because it requires additional dependencies.
Loaded 14214 tests.
...
[**]                               db/cmd/cmd_zignature    13274 OK       927 BR        0 XX       13 FX

Finished in 14 minutes and 5 seconds.
```

eagleoflqj:
Build warnings
```
CC hex.c
In file included from /home/pi/github/radare2/libr/include/r_types.h:220,
                 from str.c:3:
str.c: In function ‘r_str_encoded_json’:
/home/pi/github/radare2/libr/include/r_types_base.h:77:18: warning: conversion from ‘long long unsigned int’ to ‘size_t’ {aka ‘unsigned int’} changes value from ‘18446744073709551615’ to ‘4294967295’ [-Woverflow]
 #define UT64_MAX 0xFFFFFFFFFFFFFFFFULL
                  ^~~~~~~~~~~~~~~~~~~~~
/home/pi/github/radare2/libr/include/r_types_base.h:110:18: note: in expansion of macro ‘UT64_MAX’
 #define SZT_MAX  UT64_MAX
                  ^~~~~~~~
str.c:1525:52: note: in expansion of macro ‘SZT_MAX’
   if (!SZT_MUL_OVFCHK (((buf_sz * increment) + 1), SZT_MAX)) {
                                                    ^~~~~~~
```
```
CC fs_io.c
In file included from /home/pi/github/radare2/libr/../shlr/sdb/src/sdb.h:13,
                 from /home/pi/github/radare2/libr/include/r_list.h:6,
                 from /home/pi/github/radare2/libr/include/r_util/r_graph.h:4,
[...]

Issue #15849, opened 2020-01-19
Failing rahash2 and analysis tests on PowerPC target


XVilka:
```
[XX] db/tools/rahash2 rahash2 -E serpent-ecb 128 bit  68
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -i /tmp/tmp-19286CFdnJtspfhXW -
!rahash2 -E serpent-ecb -S "BFBFBFBFBFBFBFBFBFBFBFBFBFBFBFBF" -x "BFBFBFBFBFBFBFBFBFBFBFBFBFBFBFBF" | rax2 -S
-af39614e747b9331c38b797f527ebea6
+af78035b00000000342b1eb900000000
EXPECT=<<EOF
af78035b00000000342b1eb900000000
[XX] db/tools/rahash2 rahash2 -D serpent-ecb 128 bit  72
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -i /tmp/tmp-19286E0aj2VpHVH7S -
!rahash2 -D serpent-ecb -S "2BD6459F82C5B300952C49104881FF48" -x "EA024714AD5C4D84EA024714AD5C4D84" | rax2 -S
-beb6c069393822d3be73ff30525ec43e
+dd3116ed0000000013d1ffd200000000
EXPECT=<<EOF
dd3116ed0000000013d1ffd200000000
[XX] db/tools/rahash2 rahash2 -E serpent-ecb 192 bit  69
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -i /tmp/tmp-19286liVa8d7k07WM -
!rahash2 -E serpent-ecb -S "C6C6C6C6C6C6C6C6C6C6C6C6C6C6C6C6C6C6C6C6C6C6C6C7" -x "C6C6C6C6C6C6C6C6C6C6C6C6C6C6C6C7" | rax2 -S
-844a4ef08b4354c52593d3e907a163f7
+8ebda27400000000f40d797500000000
EXPECT=<<EOF
8ebda27400000000f40d797500000000
[XX] db/tools/rahash2 rahash2 -D serpent-ecb 192 bit  68
$ r2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Q -i /tmp/tmp-19286s0tTF2LcRSUB -
[...]

PR #17899, opened 2020-11-07
Fix UB in serpent

eagleoflqj:

**Detailed description**

UB in crypto_serpent.c: out of boundary memory read.
UB in crypto_serpent_algo.c: shift of int more than 31.
Partially solved #15849 and #17898 

**Test plan**

https://travis-ci.com/github/radareorg/radare2/jobs/427537252#L1047
https://travis-ci.com/github/eagleoflqj/radare2/jobs/430534124#L1020
db/tools/rahash2 fixed.

**Closing issues**

...

### DIFF
--- a/libr/crypto/p/crypto_serpent.c
+++ b/libr/crypto/p/crypto_serpent.c
@@ -50,11 +50,14 @@ static bool update(RCrypto *cry, const ut8 *buf, int len) {
 	}
 
 	// Construct ut32 blocks from byte stream
-	for (j = 0; j < size/4; j++) {
-		ibuf[j] = r_read_le32(&buf[4*j]);
+	for (j = 0; j < len/4; j++) {
+		ibuf[j] = r_read_le32 (&buf[4*j]);
 	}
-
-	// Zero padding.
+	if (len & 0x3) {
+		ut8 tail[4] = {0}; // Zero padding
+		memcpy (tail, buf + (len & ~0x3), len & 0x3);
+		ibuf[len/4] = r_read_le32 (tail);
+	}
 
 	if (cry->dir == 0) {
 		for (i = 0; i < blocks; i++) {
--- a/libr/crypto/p/crypto_serpent.c
+++ b/libr/crypto/p/crypto_serpent.c
@@ -9,7 +9,7 @@ static bool serpent_set_key(RCrypto *cry, const ut8 *key, int keylen, int mode, 
 	if ((keylen != 128 / 8) && (keylen != 192 / 8) && (keylen != 256 / 8)) {
 		return false;
 	}
-	st.key_size = keylen*8;
+	st.key_size = keylen * 8;
 	eprintf ("key_size: %d\n", st.key_size);
 	memcpy (st.key, key, keylen);
 	cry->dir = direction;
@@ -33,16 +33,16 @@ static bool update(RCrypto *cry, const ut8 *buf, int len) {
 	const int blocks = size / BLOCK_SIZE;
 	int i, j;
 
-	ut8 *const obuf = calloc (4, size/4);
+	ut8 *const obuf = calloc (4, size / 4);
 	if (!obuf) {
 		return false;
 	}
-	ut32 *const ibuf = calloc (4, size/4);
+	ut32 *const ibuf = calloc (4, size / 4);
 	if (!ibuf) {
 		free (obuf);
 		return false;
 	}
-	ut32 *const tmp = calloc (4, size/4);
+	ut32 *const tmp = calloc (4, size / 4);
 	if (!tmp) {
 		free (obuf);
 		free (ibuf);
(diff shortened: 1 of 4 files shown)


libr/crypto/p/crypto_serpent_algo.c
```diff
@@ -77,7 +77,7 @@ void apply_IP(ut32 in[DW_BY_BLOCK], ut32 out[DW_BY_BLOCK]) {
 	for (i = 0; i < DW_BY_BLOCK*32; i++) {
 		index = IPTable[i];
 		out[i/32] ^= (-(ut32)get_bit (index%32, in[index/32]) ^ out[i/32])
-			& (1 << i);
+			& (1 << (i & 0x1f));
```
pelijah, line 80: Wouldn't it be better to use `i % 32`?

eagleoflqj: I don't know whether there's a compiler that won't do this conversion, so I used &0x1f directly. Just checked the assembly on rpi4 and yes, %32 is treated as &31 properly. Is there a readability concern?

libr/crypto/p/crypto_serpent_algo.c
```diff
@@ -77,7 +77,7 @@ void apply_IP(ut32 in[DW_BY_BLOCK], ut32 out[DW_BY_BLOCK]) {
 	for (i = 0; i < DW_BY_BLOCK*32; i++) {
```
trufae, line 77: add a space around *

libr/crypto/p/crypto_serpent_algo.c
trufae: this is a shift left, this means that using % will result on wrong behaviour because it will be using a rotated value instead of a shifted one.

The variable `i` goes from 0 to DW_BY_BLOCK * 32, which is 4.. this is.. this code will try to shift from 0 to 128 bits. this is a clear UB. and for me it seems like using % will result in better behaviour. but its different than the current one. I'm not sure if we have tests for this or if you tried to compare this with other implementations of serpent. but if we are just xoring the buffer aganist 0s.. its probably better to break before doing useless ops..

Can you add a test and compare the result with another implementation. https://github.com/sysopfb/pyserpent this one seems like a good candidate for messing up and comparing outputs.

libr/crypto/p/crypto_serpent.c
```diff
@@ -50,11 +50,14 @@ static bool update(RCrypto *cry, const ut8 *buf, int len) {
...
 	// Construct ut32 blocks from byte stream
-	for (j = 0; j < size/4; j++) {
-		ibuf[j] = r_read_le32(&buf[4*j]);
+	for (j = 0; j < len/4; j++) {
+		ibuf[j] = r_read_le32 (&buf[4*j]);
+	}
+	if (len & 0x3) {
+		ut8 tail[4] = {0}; // Zero padding
+		memcpy (tail, buf + (len & ~0x3), len & 0x3);
+		ibuf[len/4] = r_read_le32 (tail);
```
trufae, line 59: add spaces around /, * in all the file.

also &buf[4*j] is the same as buf + (4*j) , which is imho cleaner and easy to read

trufae (changes requested):

eagleoflqj:
The implementation you give is python2 (I have limited understanding), and seems only supports CBC mode (not ECB here). So I tried the [full submission package](http://www.cl.cam.ac.uk/~rja14/Papers/serpent.tar.gz) in the [official website of serpent](https://www.cl.cam.ac.uk/~rja14/serpent.html) on ubuntu x64. It supports 256bit mode only and the input/output is given in big endian hex. I tested with cases in db/tools/rahash2 which is 256bit.
```sh
# after cd floppy1 and make
liumeo@liumeo-ubuntu:~/serpent/floppy1$ ./serpent-test -e -k 5657575757575757575757575757575757575757575757575757575757575757 -p 56575757575757575757575757575757
plainText=56575757575757575757575757575757
userKey=5657575757575757575757575757575757575757575757575757575757575757
cipherText=b37300fa7f43dfa0f2742e3dcf8855d3

liumeo@liumeo-ubuntu:~/serpent/floppy1$ ./serpent-test -d -k 48FF814810492C9500B3C5829F45D62B48FF814810492C9500B3C5829F45D62B -c 844D5CAD144702EA844D5CAD144702EA
cipherText=844d5cad144702ea844d5cad144702ea
userKey=48ff814810492c9500b3c5829f45d62b48ff814810492c9500b3c5829f45d62b
plainText=f28ab2d115b4d23f747180a0fa8d7c67

liumeo@liumeo-ubuntu:~/serpent/floppy1$ ./serpent-test -e -k fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210 -p 00000000000000000000000000311210
plainText=00000000000000000000000000311210
userKey=fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210
cipherText=fafe561b7389b081ea46818e51fc7d41
```
In both cases, official implementation produces the same result with our implementation. The third one is the case I add for testing zero padding.
And here's a python function for convenience.
```py
def switch_endian(s):
    r = ''
    n = len(s)
    for i in range(0, n, 2):
        r = s[i] + s[i + 1] + r
    return r

switch_endian("b37300fa7f43dfa0f2742e3dcf8855d3")
```
I think now our implementation is correct (at least for 256bit mode).

XVilka (approved):

trufae (approved):

PR #17904, opened 2020-11-08
Fix raspbian32 db/cmd/cmd_ac

eagleoflqj:

**Detailed description**

Random negative numbers #17898.
There remains a potential bug https://github.com/radareorg/radare2/blob/91aebb64901285f5177151604816d270d0862039/libr/anal/class.c#L512
which should be `atoll` but it's only available in C99. I notice ANSI C building is still supported now, so I didn't change this.

**Test plan**

`r2r db/cmd/cmd_ac` on rpi4.

**Closing issues**

...

### DIFF
--- a/libr/anal/class.c
+++ b/libr/anal/class.c
@@ -556,7 +556,7 @@ R_API RVector/*<RAnalMethod>*/ *r_anal_class_method_get_all(RAnal *anal, const c
 }
 
 R_API RAnalClassErr r_anal_class_method_set(RAnal *anal, const char *class_name, RAnalMethod *meth) {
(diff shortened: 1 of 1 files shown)


XVilka:
I think you can use `atoll` - we can't support non-C99 compilers due to the tree-sitter inclusion.

ret2libc:
> I think you can use `atoll` - we can't support non-C99 compilers due to the tree-sitter inclusion. cc @ret2libc @trufae

IIRC even before the introduction of tree-sitter, `std=gnu99` was used in SDB as well, so I don't think it's a problem (see https://github.com/radareorg/sdb/blob/master/config.mk#L34).

XVilka (approved):

PR #17887, opened 2020-11-01
Fix r_core_editor() on Windows ##ports

GustavoLCR:
**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [ ] I've added tests that prove my fix is effective or that my feature works (if possible)
- [ ] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Test plan**

Manually tested

PR #17908, opened 2020-11-09
Revert "Fix r_core_editor() on Windows ##ports"

XVilka:
Reverts radareorg/radare2#17887

### DIFF
--- a/libr/core/core.c
+++ b/libr/core/core.c
@@ -3492,9 +3492,7 @@ R_API char *r_core_editor(const RCore *core, const char *file, const char *str) 
 		cons->cb_editor = tmp;
 	} else {
 		if (editor && name) {
-			char *escaped_name = r_str_escape_sh (name);
-			r_sys_cmdf ("%s \"%s\"", editor, escaped_name);
-			free (escaped_name);
+			r_sys_cmdf ("%s '%s'", editor, name);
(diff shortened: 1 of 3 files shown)


libr/core/core.c
```diff
@@ -3492,9 +3492,7 @@ R_API char *r_core_editor(const RCore *core, const char *file, const char *str)
 		cons->cb_editor = tmp;
 	} else {
 		if (editor && name) {
-			char *escaped_name = r_str_escape_sh (name);
-			r_sys_cmdf ("%s \"%s\"", editor, escaped_name);
-			free (escaped_name);
+			r_sys_cmdf ("%s '%s'", editor, name);
```
kazarmy, line 3495: I think this change should **not** be reverted.

libr/util/str.c
```diff
@@ -885,7 +885,7 @@ R_API char *r_str_append(char *ptr, const char *string) {
 	if (string && !ptr) {
 		return strdup (string);
 	}
-	if (R_STR_ISEMPTY (string)) {
+	if (!string) {
```
kazarmy, line 888: I think this change should **not** be reverted too.

kazarmy (changes requested):
Maybe change the pr title?

PR #17931, opened 2020-11-17
fixed add breakpoint

HereCouldBeName:
- [+] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [+] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)

**Detailed description**

When creating a breakpoint, the r_debug_gdb_breakpoint function was not called. Therefore Breakpoints didn't work.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/libr/debug/debug.c
+++ b/libr/debug/debug.c
@@ -335,6 +335,11 @@ R_API RBreakpointItem *r_debug_bp_add(RDebug *dbg, ut64 addr, int hw, bool watch
 			bpi->name = r_str_newf ("%s+0x%" PFMT64x, module_name, m_delta);
 		}
 		bpi->module_delta = m_delta;
+		if (dbg && dbg->h && dbg->h->breakpoint) {
+			if (!dbg->h->breakpoint (dbg->bp, bpi, true)) {
+				return NULL;
+			}
+		}
 	}
 	return bpi;
 }


yossizap:
>r_debug_gdb_breakpoint function was not called. Therefore Breakpoints didn't work

It seems to work for me in the current state with software breakpoint and x86. Is this an issue with hardware breakpoints or a specific arch?

HereCouldBeName:
> > r_debug_gdb_breakpoint function was not called. Therefore Breakpoints didn't work
> 
> It seems to work for me in the current state with software breakpoint and x86. Is this an issue with hardware breakpoints or a specific arch?

I have connected to a remote Qemu debugger with AVR architecture. There weren't breakpoints in this version. I ran avr-gdb and noticed that it contains Z instructions. In radare2 these instructions were not called.

> > r_debug_gdb_breakpoint function was not called. Therefore Breakpoints didn't work
> 
> It seems to work for me in the current state with software breakpoint and x86. Is this an issue with hardware breakpoints or a specific arch?

I connected to the Qemu remote debugger using radare2.
My steps:
1) In Qemu, I ran qemu-system-avr with an avr bin file.
2) Connected to a remote debugger in radare2.
3) Used to set a breakpoint with the "db" command.
4) On entering the "dc" command, I should have jumped to a breakpoint, but it didn't.
5) I ran avr-gdb and also remote connected to Qemu to collect logs.
6) It turned out that radare2 does not send instructions 'Z' to Qemu.
7) After adding the code written above, this problem disappeared. After setting a breakpoint, you can now navigate to it.

yossizap:
> In Qemu, I ran qemu-system-avr with an avr bin file.

AVR doesn't have a plugin in `libr/bp/p` so I am not sure how `r_bp_add_[sw/hw]` even worked for you, maybe a different arch is selected in `dmh`? I think that the correct solution should check if `r_bp_add`'s failed and only run the backend's breakpoint function if a breakpoint backend doesn't exist in r2. This way you are potentially placing a breakpoint on the same place twice with two backends(r2 and the debug plugin) and then having both of them handle the breakpoint when triggered.

I tried reproducing this to look into the solution but I couldn't even receive a valid target XML from qemu and everything failed. I used `qemu-system-avr -machine mega2560 -bios demo.elf -s -S`([demo.elf](https://github.com/seharris/qemu-avr-tests/blob/master/free-rtos/Demo/AVR_ATMega2560_GCC/demo.elf)) with a local build of upstream qemu (configure --enable-debug). `oodf gdb://127.0.0.1:1234` on r2's side after analysis on demo.elf succeeded. Am I missing something?

karliss:
>  I think that the correct solution should check if `r_bp_add`'s failed and only run the backend's breakpoint function if a breakpoint backend doesn't exist in r2.

Shouldn't it be other way around? Use plugin if plugin provides functionality and if plugin doesn't implement it fallback to r2.

HereCouldBeName:
> > In Qemu, I ran qemu-system-avr with an avr bin file.
> 
> AVR doesn't have a plugin in `libr/bp/p` so I am not sure how `r_bp_add_[sw/hw]` even worked for you, maybe a different arch is selected in `dbh`? I think that the correct solution should check if `r_bp_add`'s failed and only run the backend's breakpoint function if a breakpoint backend doesn't exist in r2. With the current solution you are potentially placing a breakpoint on the same place twice with two backends(r2 bp and the debug plugin) and then having both backends handle the breakpoint when triggered.
> 
> I tried reproducing this to look into the solution but I couldn't even receive a valid target XML from qemu and everything failed. I used `qemu-system-avr -machine mega2560 -bios demo.elf -s -S`([demo.elf](https://github.com/seharris/qemu-avr-tests/blob/master/free-rtos/Demo/AVR_ATMega2560_GCC/demo.elf)) with a local build of upstream qemu (configure --enable-debug). `oodf gdb://127.0.0.1:1234` on r2's side after analysis on demo.elf succeeded. Am I missing something?

I used Cutter to work with radare2.
At the moment I have verified that the r_bp_add_sw function is being called. Also the architecture is referred to as AVR.
I attach the screen:
[http://i.imgur.com/LUdyoy9.png](url)

HereCouldBeName:
> > In Qemu, I ran qemu-system-avr with an avr bin file.
> 
> AVR doesn't have a plugin in `libr/bp/p` so I am not sure how `r_bp_add_[sw/hw]` even worked for you, maybe a different arch is selected in `dbh`? I think that the correct solution should check if `r_bp_add`'s failed and only run the backend's breakpoint function if a breakpoint backend doesn't exist in r2. With the current solution you are potentially placing a breakpoint on the same place twice with two backends(r2 bp and the debug plugin) and then having both backends handle the breakpoint when triggered.
> 
> I tried reproducing this to look into the solution but I couldn't even receive a valid target XML from qemu and everything failed. I used `qemu-system-avr -machine mega2560 -bios demo.elf -s -S`([demo.elf](https://github.com/seharris/qemu-avr-tests/blob/master/free-rtos/Demo/AVR_ATMega2560_GCC/demo.elf)) with a local build of upstream qemu (configure --enable-debug). `oodf gdb://127.0.0.1:1234` on r2's side after analysis on demo.elf succeeded. Am I missing something?

I used Cutter to work with radare2.
At the moment I have verified that the r_bp_add_sw function is being called. Also the architecture is referred to as AVR.
I attach the screen:
s1

yossizap:
>I used Cutter to work with radare2.

I get the same failure since Cutter just runs `oodf` and the target XML with my build of qemu running mega2560 is broken. 
Can you please share your qemu version and qemu repository/binary if it's a fork in addition to the line you ran qemu with?

>Shouldn't it be other way around? Use plugin if plugin provides functionality and if plugin doesn't implement it fallback to r2.

Agreed but it will require changing how breakpoint entries are added and that might be out of the scope of this PR. The debug backend breakpoint function depends on an initialized `RBreakpontItem` from `r_bp_add`.

HereCouldBeName:
> > I used Cutter to work with radare2.
> 
> I get the same failure since Cutter just runs `oodf` and the target XML with my build of qemu running mega2560 is broken.
> Can you please share your qemu version and qemu repository/binary if it's a fork in addition to the line you ran qemu with?
> 
> > Shouldn't it be other way around? Use plugin if plugin provides functionality and if plugin doesn't implement it fallback to r2.
> 
> Agreed but it will require changing how breakpoint entries are added and that might be out of the scope of this PR. The debug backend breakpoint function depends on an initialized `RBreakpontItem` from `r_bp_add`.

My branch qemu: https://github.com/HereCouldBeName/qemu_my
I used `avr-softmmu/qemu-system-avr -bios 6_2.bin -machine Atmega8  -monitor stdio -s -S -D my_log.txt` [6_2.bin](https://github.com/HereCouldBeName/Qemu-file/blob/master/6_2.bin)

yossizap:
Thanks for providing your qemu branch and binaries for testing! After testing with your binaries it seems that this fix only bypasses the placement and removal of breakpoints before and after continue/step. The sections in this binary aren't properly registered in r2, possibly due to a gdb protocol issue on qemu's side that should still be looked into(`WARNING: No sections found, functions and flags won't be rebased`), and the breakpoints are set as invalid - meaning that they won't be placed before each step/continue.

I am going to close this PR since the solution isn't relevant and your issue can be solved by turning off bpinmaps(`e dbg.bpinmaps=false`) to validate breakpoints in invalid addresses. I added a warning to avoid users being confused by this in  #17986 and fixed the breakpoint placement issue when no arch is available that we discussed here.

Note - Debug backend breakpoints already take priority over the debug backend.

PR #17945, opened 2020-11-20
Fix ASAN crash when analyzing arm64

GustavoLCR:

**Detailed description**
* Remove static `RRegItem` from anal_arm_cs.c

**Test plan**

Tests pass

### DIFF
--- a/libr/anal/p/anal_arm_cs.c
+++ b/libr/anal/p/anal_arm_cs.c
@@ -63,8 +63,6 @@
 #define ISPREINDEX64() ((OPCOUNT64() == 3) && (ISMEM64(2)) && (ISWRITEBACK64()))
 #define ISPOSTINDEX64() ((OPCOUNT64() == 4) && (ISIMM64(3)) && (ISWRITEBACK64()))
 
-static RRegItem base_regs[4];
-static RRegItem regdelta_regs[4];
 static HtUU *ht_itblock = NULL;
 static HtUU *ht_it = NULL;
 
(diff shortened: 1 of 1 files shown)


XVilka (approved):

XVilka:
There one TIMEOUT still, but ASAN crashes were fixed, good job!
```
[XX] TIMEOUT db/cmd/cmd_aae crash aae oob
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc aae bins/elf/arm32esilcrash
-- exit status: -1
```

libr/anal/p/anal_arm_cs.c
```diff
@@ -3772,9 +3759,9 @@ static void op_fillval (RAnalOp *op, csh handle, cs_insn *insn, int bits) {
 		}
 #endif
 		for (j = 0; j < 3; j++, i++) {
-			set_src_dst (op->src[j], &handle, insn, i, bits);
+			set_src_dst (op->src[j], anal->reg, &handle, insn, i, bits);
```
karliss, line 3762: @GustavoLCR Is reg_num/x actually used? Shouldn't it be `j` here instead of `i`?

trufae:
Wasnt a strdup enough? So many r_reg_gets in there is making the analysis much slower. Why merging that quickly

libr/anal/p/anal_arm_cs.c
GustavoLCR: @karliss reg_num/x is used to get the operand inside the instruction, and its `i`, because it can at an offset from `j`

GustavoLCR:
> Wasnt a strdup enough? So many r_reg_gets in there is making the analysis much slower. Why merging that quickly

@trufae x86 analysis also uses r_reg_get, so i just copied from it to kill the global array

PR #17954, opened 2020-11-21
Call convention support for x86-64 kernel interface (syscall) %rdi, %…

David-Reguera-Garcia-Dreg:
…rsi, %rdx, %r10, %r8, %r9

 

**Your checklist for this pull request**
- [x] I've read the [guidelines for contributing](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md) to this repository
- [x] I made sure to follow the project's [coding style](https://github.com/radareorg/radare2/blob/master/DEVELOPERS.md#code-style)
- [x] I've added tests that prove my fix is effective or that my feature works (if possible)
- [x] I've updated the documentation and the [radare2 book](https://github.com/radareorg/radare2book) with the relevant information (if needed)

**Detailed description**

The current amd64 call convention in r2 supported is User-level call convention: %rdi, %rsi, %rdx, **%rcx, %r8, %r9**

I add the amd64 kernel interface call convention (amd64syscall): %rdi, %rsi, %rdx, **%r10, %r8, %r9**

**Test plan**

Its impossible test it yet, reason: **libr/core/cmd_anal.c**
```
// TODO: move this to r_syscall
//TODO replace the hardcoded CC with the sdb ones
```

Just debug a x86-64-syscall-code with 4 args or more and run "as". Example with wait4 (4 args):

```
    mov rcx, 0 ; rcx is not an arg in this syscall-case
    mov rdi, 1
    mov rsi, 2
    mov rdx, 3
    mov r10, 4
    mov r8, 5
    mov r9, 6
    mov rax, 61 ; wait4(rdi, rsi, rdx, r10)
    syscall
```

**Closing issues**

PR #17960, opened 2020-11-22
Add anal.cc & anal.syscc and remove the hardcoded conventions in debug ##anal

trufae:

**Detailed description**

This PR introduces two new eval variables:

* anal.cc
* anal.syscc

those are used to set the sdb keys default.cc and default.syscc, which is used tod efine the default calling convention for analysing functions and syscalls. This change also lets the https://github.com/radareorg/radare2/pull/17954 to allow to write tests for it

**Test plan**

Fix/review all the tests that are failing and make CI green.

Expect https://github.com/radareorg/radare2/pull/17954 to add a test after rebasing

**Closing issues**

not sure if there's one, but the src had comments about it, and it was planned to kill all the hardcoded calling conventions. so i did it.


libr/core/cmd_anal.c
```diff
@@ -5978,7 +5979,7 @@ static void r_anal_aefa(RCore *core, const char *arg) {
 		int i;
 		eprintf ("NARGS %d (%s)\n", nargs, key);
 		for (i = 0; i < nargs; i++) {
-			ut64 v = r_debug_arg_get (core->dbg, R_ANAL_CC_TYPE_STDCALL, i);
+			ut64 v = r_debug_arg_get (core->dbg, "reg", i);
```
trufae, line 5982: not sure about stdcall.. maybe just using `cc` is fine

pelijah:
Then you have to kill `afc=` command.

trufae:
No. The functions take the calling convention defined by default. But you can specify a different one on every function. So afc shouldnt die. But the problem is that befofe the cc had no way to specify which was the default. Also with this im allowing to define different convwntions for syscalls and userland code

pelijah:
> No. The functions take the calling convention defined by default. But you can specify a different one on every function. So afc shouldnt die. But the problem is that befofe the cc had no way to specify which was the default. Also with this im allowing to define different convwntions for syscalls and userland code

See `afc=` https://github.com/radareorg/radare2/blob/master/libr/core/cmd_anal.c#L372

pelijah:
And `afcR` handler uses same query
https://github.com/radareorg/radare2/blob/master/libr/core/cmd_anal.c#L3846
https://github.com/radareorg/radare2/blob/master/libr/core/cmd_anal.c#L3939

pelijah:
Well in case of `afcR` it would be better to replace this string command with API call. But `afc=` duplicates `e anal.cc`.

karliss:
Is there a way for storing known os-arch-bits to syscall cc mapping so that for combinations with known syscall numbers which r2 currently defines in syscall/d/*.sdb.txt files, so that it isn't necessary to manually set `anal.syscc` each time?

trufae:
afc= is suposed to set the calling convention for the given function. anal.cc should be the default if none specified. so.. if afc= is just doing the same that must be fixed i guess. anyway what do you think about this change? adding this new variable makes sense to me, what about you?

pelijah:
`afc` (not `afc=`) is suposed to set the calling convention for the given function. 
Unfortunatelly, I am not active r2 user so I can't say how useful are these changes.

PR #17968, opened 2020-11-23
Make hints seek directly to indirect call target address

GustavoLCR:

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -147,6 +147,7 @@ typedef struct {
 	bool asm_hint_jmp;
 	bool asm_hint_cdiv;
 	bool asm_hint_call;
+	bool asm_hint_call_indirect;
 	bool asm_hint_lea;
 	bool asm_hint_emu;
 	int  asm_hint_pos;
@@ -717,6 +718,7 @@ static RDisasmState * ds_init(RCore *core) {
 	ds->show_usercomments = r_config_get_i (core->config, "asm.usercomments");
 	ds->asm_hint_jmp = r_config_get_i (core->config, "asm.hint.jmp");
 	ds->asm_hint_call = r_config_get_i (core->config, "asm.hint.call");
+	ds->asm_hint_call_indirect = r_config_get_i (core->config, "asm.hint.call.indirect");
 	ds->asm_hint_lea = r_config_get_i (core->config, "asm.hint.lea");
 	ds->asm_hint_emu = r_config_get_i (core->config, "asm.hint.emu");
 	ds->asm_hint_cdiv = r_config_get_i (core->config, "asm.hint.cdiv");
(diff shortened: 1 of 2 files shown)


XVilka (approved):

ret2libc (commented):
A test to show what this does exactly?

trufae (commented):
I think its worth having hints to follow both pointers, the destination address and the address hold in that address

libr/core/disasm.c
```diff
@@ -3565,24 +3565,20 @@ static bool ds_print_core_vmode(RDisasmState *ds, int pos) {
...
-			// ds_print_shortcut (ds, ds->analop.jump, pos);
-			ds_print_shortcut (ds, ds->analop.ptr, pos);
-		}
-#endif
 		if (ds->asm_hint_call) {
 			if (ds->analop.jump != UT64_MAX) {
 				slen = ds_print_shortcut (ds, ds->analop.jump, pos);
 			} else {
-				slen = ds_print_shortcut (ds, ds->analop.ptr, pos);
+				ut8 buf[sizeof (ut64)] = {0};
```
trufae, line 3572: ut64 callref = read_u64(core, ds->analop.ptr);

^ and put the body of this function as static. but maybe useful also for other places i think

libr/core/disasm.c
```diff
@@ -3565,24 +3565,20 @@ static bool ds_print_core_vmode(RDisasmState *ds, int pos) {
...
+				r_io_read_at (core->io, ds->analop.ptr, buf, sizeof (buf));
+				ut64 call_addr = core->rasm->bits == 64
+					? r_read_ble64 (buf, core->rasm->big_endian)
+					: r_read_ble32 (buf, core->rasm->big_endian);
+				if (r_io_is_valid_offset (core->io, call_addr, R_PERM_X)) {
+					slen = ds_print_shortcut (ds, call_addr, pos);
+				} else {
```
trufae, line 3579: just change call_addr = ds->analop.ptr; in case the offset is not valid and call ds_print_shortcut once

libr/core/disasm.c
```diff
@@ -3565,24 +3565,20 @@ static bool ds_print_core_vmode(RDisasmState *ds, int pos) {
...
 			if (ds->analop.jump != UT64_MAX) {
 				slen = ds_print_shortcut (ds, ds->analop.jump, pos);
 			} else {
-				slen = ds_print_shortcut (ds, ds->analop.ptr, pos);
+				ut8 buf[sizeof (ut64)] = {0};
+				r_io_read_at (core->io, ds->analop.ptr, buf, sizeof (buf));
+				ut64 call_addr = core->rasm->bits == 64
+					? r_read_ble64 (buf, core->rasm->big_endian)
+					: r_read_ble32 (buf, core->rasm->big_endian);
+				if (r_io_is_valid_offset (core->io, call_addr, R_PERM_X)) {
```
trufae, line 3577: i can find useful to have this behaviour as an option. this is to have another hint mode. what do you think?

trufae (changes requested):

libr/core/disasm.c
GustavoLCR: added `asm.hint.call.indirect`

trufae:
LGTM. but im starting to think that we shuold have a single asm.hint var and use the value the type of hint to ref, (as well as support numeric value to increment from visual and such) because it makes no sense to have more than one hint to be accessible at the same time.

trufae:
@GustavoLCR let's merge it like this, defaults seems reasonable and useful to me 👍 

We can discuss about having a single hint variable in a separate PR.

PR #17978, opened 2020-11-27
Update fortunes.nsfw

banesec:
Just a suggestion


ITAYC0HEN:
yes, please

trufae:
I wouldn't call censorship an "update". I'm totally fine for removing the nsfw fortunes, i'm tired of managing offended people and even if this file is not used by default, such sense of humour is no longer appreciated by anyone.

The problem I see in this PR are:

* the file is just emptied instead of removed
* referenced uses are not removed
* commit message doesnt reflects the reality of the changes
* for some reason @thestr4ng3r was the only codeowner of this file when it shuold be owned by me

I guess this is the issue in where we are suposed to discuss about puritanism in r2land. So feel free to expose your positions in here.

thestr4ng3r:
Looks like emptying out the file was done on purpose, in the sense that these fortunes are not nsfw anymore but should be added to the sfw category.
Jokes aside, while I do like these fortunes and consider them satirical, I'm fine with removing these if it makes anyone feel better.

XVilka:
I don't have a specific opinion on this. It's sad that some people find this offensive though. I guess we are living in new Cyber-Victorian times.

XVilka:
Also how did `"Our exploits production increased for 24% annually after our switch to radare2" -- An unnamed agency` and `Everytime you run radare2, a random file is removed` became NSFW? 🙄 
Also why `radare2 is a hex toy` considered harmful? Started as a hex editor toy project...

ITAYC0HEN:
I recommend supporting (if not already supported) loading fortunes from custom fortunes path and then you set it up in your machine once and forget about it.

Then, someone can create their collections of fortunes, update them to github and share with the world. Thus the fortunes wouldn't disappear for those who love them, but will not be part of the code base.

I agree with @trufae about the points that were raised (commit. message, removing the file, etc)

condret:
I don't see any reason for removing nsfw fortunes, people who cannot deal with this should simply not activate them. nsfw fortunes are not activated by default, which is already a compromise

hybriz:
muh freedoms! 

why not just do a PR to remove all of the repo?

this is github and by being a user here and having repos to attract other users one is helping their business, thus contributing to the bottom line of a company that enables DHS/ICE and their concentration camps

does anyone knows where I could file a PR to remove all of github ?

hybriz:
> Also how did `"Our exploits production increased for 24% annually after our switch to radare2" -- An unnamed agency` and `Everytime you run radare2, a random file is removed` became NSFW? 
> Moreover, why `radare2 is a hex toy` considered harmful? Started as a hex editor toy project...

these should be in the sfw list ! ! ! wtf

hybriz:
can someone provide a copy for review of the butthurt report for this?

wargio:
Moved discussion to https://github.com/radareorg/radare2/pull/17980

PR #17982, opened 2020-11-28
Use r_core_cmd_str_pipe when using `.` in a panel

trufae:
* Captures the output of the r2pipe interpreter's `print`

 

**Detailed description**

Running r2pipe scripts in visual panels results in no output captured because the `print` statements of the language are printed directly to fd=2, in order to capture them this patch checks if the command is starting with a dot and uses the cmdstr_pipe() function instead.

Known issues:

* I tried to handle #! but it doesnt works. 

```
$ cat test.py
import r2pipe

r2 = r2pipe.open()
print(r2.cmd("?e hehehehe"))
$
```

**Test plan**

run r2 panels mode like this: `r2 -cv /bin/ls`

press `e` to edit the command to run in the current panel and type `. test.py`

without this patch there's no output captured.

IMAGE 2020-11-28 01:30:46

**Closing issues**

None but reported in the chat by dreg

### DIFF
--- a/libr/core/cmd_hash.c
+++ b/libr/core/cmd_hash.c
@@ -337,13 +337,13 @@ static int cmd_hash_bang (RCore *core, const char *input) {
 					eprintf ("Error: scr.interactive required to run the rlang prompt\n");
 				}
 			}
-		} else if (av[0][0]=='?' || av[0][0]=='*') {
+		} else if (av[0][0] == '?' || av[0][0] == '*') {
 			r_lang_list (core->lang);
 		}
 	} else {
 		r_lang_list (core->lang);
 	}
-	r_str_argv_free(av);
+	r_str_argv_free (av);
 	return true;
 }
 
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -872,8 +872,9 @@ static char *__handle_cmd_str_cache(RCore *core, RPanel *panel, bool force_cache
 	if (b) {
 		core->print->cur_enabled = false;
 	}
-	char *out = r_core_cmd_str (core, cmd);
-	r_cons_echo (NULL);
+	char *out = (*cmd == '.' || *cmd == '#')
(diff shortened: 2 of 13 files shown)


David-Reguera-Garcia-Dreg:
dont works with args, ex: . /root/r2-syscall-printer/r2-syscall-printer.py /32 /extra

trufae:
Yes thats expected behaviour because the dot command includes a script. Spaces are treated as spaces in the file name itself.

trufae:
It works now but seems like i broke something else. Must check before merging. Can you test?

David-Reguera-Garcia-Dreg:
The behaviour its annoying: 

pannel_bug_partial

I try also creating r2-script with some prints and I have the same problem, I dont know why. 

This script shows nothing in pannel:
```
#!/usr/bin/env python3
import r2pipe

r2 = r2pipe.open()
print("one")
print("\ntwo\n")
```

In console it works:
```
:> #!pipe python3 /root/r2-syscall-printer/r2-syscall-printer.py /32 /extra
one

two

:>
```

trufae:
git pull and try with this example 👍 
```
$ cat ../lang/py.py
import r2pipe
import sys

a = str(sys.argv)
r2 = r2pipe.open()
print("echo")
r2.cmd("?e hello world %s"%(a))
```

trufae:
In short: 

if yuo use `print` that will be printed to stdout. and rlang cant handle that well because of buffering and other reasons. so in order to print stuff handlable by r2 you need to use rcons, and the way to use rcons from r2pipe is via `?e` to echo things

David-Reguera-Garcia-Dreg:
This dont works:

```
#!/usr/bin/env python3
import r2pipe
import sys

a = str(sys.argv)
r2 = r2pipe.open()
print("echo")
r2.cmd("?e hello world %s"%(a))
print("\nteeest\n")
r2.cmd("?e wooop")
```

PR #17986, opened 2020-11-28
Fixed breakpoint placement when no bp arch is available and added a warning for invalid bps ##debug

yossizap:

**Detailed description**

* Placing breakpoints without a selected bp plugin(the selection can be seen in `dbh`) caused breakpoint creation to fail even though the debug backend might have supported setting a breakpoint.
* It seems that issues with section mapping in avr arch/using qemu gdb information caused confusion in #17931 when breakpoints were marked as invalid and weren't placed. The second change simply adds a warning for users to be aware that their breakpoints aren't going to work without `dbg.bpinmaps` set to false in case they placed them in an invalid addresses.

**Test plan**

* See that you are able to place breakpoints without a selected bp plugin when connecting to gdbserver(you can change the x86 plugin's name in `libr/bp/p/bp_x86.c` to something else if you are testing on a x86 machine).
* See the invalid breakpoint warning using the binary given at #17931 or by setting a very high address

### DIFF
--- a/libr/bp/bp.c
+++ b/libr/bp/bp.c
@@ -173,6 +173,10 @@ static RBreakpointItem *r_bp_add(RBreakpoint *bp, const ut8 *obytes, ut64 addr, 
 	if (bp->baddr > addr) {
 		eprintf ("base addr should not be larger than the breakpoint address.\n");
 	}
+	if (bp->bpinmaps && !r_bp_is_valid (bp, b)) {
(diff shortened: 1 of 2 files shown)


trufae (commented):
Uhmm i have to test this. Isnt the recoil not saved per item now? That info should be kept

yossizap:
bp->recoil isn't used anywhere, size is used instead.

trufae (approved):

XVilka (approved):

PR #17992, opened 2020-11-29
asm_*_gnu.c: *_buffer_read_memory: honour memaddr as offset into code buffer

Spritetm:

**Detailed description**

The V850.gnu disassembler target didn't honour memory offset in reading code memory, mis-disassembling e.g. mov instructions with an IMM32 operand. This fixes that.

**Test plan**

    echo -ne '\x21\x06\x74\x06\x00\x00' > test.bin; r2 -A -a v850.gnu -c 'pd 1 @0' test.bin

Good output: 0x00000000      210674060000   mov 0x674, r1

### DIFF
--- a/libr/asm/p/asm_cris_gnu.c
+++ b/libr/asm/p/asm_cris_gnu.c
@@ -26,7 +26,14 @@ static RStrBuf *buf_global = NULL;
 static unsigned char bytes[8];
 
 static int cris_buffer_read_memory (bfd_vma memaddr, bfd_byte *myaddr, ut32 length, struct disassemble_info *info) {
-	memcpy (myaddr, bytes, length);
+	int delta = (memaddr - Offset);
+	if (delta < 0) {
+		return -1;      // disable backward reads
+	}
+	if ((delta + length) > 8) {
+		return -1;
+	}
+	memcpy (myaddr, bytes + delta, length);
 	return 0;
 }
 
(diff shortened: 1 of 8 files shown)


trufae:
This code was copypasted from another gnu plugin i dont recall right now. Can you fix the other *gnu.c files with the same fix? Also, do you have a reproducer?

Thanks!

Spritetm:
Possibly, lemme check real quick. What's a reproducer?

Spritetm:
Seems the asm_arm_gnu file already had a better working implementation, so I copy/pasted that to the other args that didn't have it.

trufae (changes requested):
Looks better now! But ideally we may try to reduce as much duplicated code as possible. Do you think you can address this in this PR? Can you add a test?

Spritetm:
I've added a test for the original issue. While I get the code copy/paste issue, the original code already was mostly a copy/paste job and me fixing that would mean architectural changes I'm not comfortable with doing... I'm barely a r2 user yet, let alone a developer, so if you don't mind, I'll leave that to the next person.

trufae (approved):
Thanks for your comments. Yeah no pressure at all. Just wanted to clarify your plans after this. The PR lgtm now!

wargio (approved):

wargio:
Thanks for the PR

PR #17998, opened 2020-12-06
Update cmd_open_map() output format for current map

Lazula:
This makes om. output consistent with om

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -570,10 +570,10 @@ static void cmd_open_map(RCore *core, const char *input) {
 	case '.': // "om."
 		map = r_io_map_get (core->io, core->offset);
 		if (map) {
-			core->print->cb_printf ("map: %i fd: %i +0x%"PFMT64x" 0x%"PFMT64x
-				" - 0x%"PFMT64x" ; %s : %s\n", map->id, map->fd,
-				map->delta, map->itv.addr, r_itv_end (map->itv),
-			r_str_rwx_i (map->perm), map->name ? map->name : "");
+			core->print->cb_printf ("%2d fd: %i +0x%08"PFMT64x" 0x%08"PFMT64x
+				" - 0x%08"PFMT64x" %s %s\n", map->id, map->fd,
+				map->delta, map->itv.addr, r_itv_end (map->itv) - 1,
+				r_str_rwx_i (map->perm), (map->name ? map->name : ""));
 		}
 		break;
 	case 'r': // "omr"
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -6,6 +6,7 @@
 #include "r_util.h"
 #include "r_bin.h"
 #include "r_debug.h"
+#include "r_io.h"
 
 static const char *help_msg_o[] = {
 	"Usage: o","[com- ] [file] ([offset])","",
@@ -422,7 +423,7 @@ static void map_list(RIO *io, int mode, RPrint *print, int fd) {
 		switch (mode) {
 		case 'q':
 			if (fd == -2) {
-				print->cb_printf ("0x%08"PFMT64x"\n", map->itv.addr);
+				print->cb_printf ("0x%08"PFMT64x"\n", r_io_map_get_from (map));
 			} else {
 				print->cb_printf ("%d %d\n", map->fd, map->id);
 			}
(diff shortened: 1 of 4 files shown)


Lazula:
Had to make a new commit as `r2r -i` did not change the tests how I expected. Updated manually.

trufae:
What was r2r doing that was not expected?

Lazula:
Just a case of new user.  I don't know exactly what happened, but one of the tests had an unintended change I noticed looking at the diff.

test/db/formats/pe/sectiondata
```diff
@@ -28,7 +28,7 @@ NAME=PE: corkami cm01.exe - section seek 1
 FILE=bins/pe/cm01.exe
 CMDS=<<EOF
 s 0x00401000
-om.~[10]
+om.~[8]
```
trufae, line 31: i think it will be good to have a test covering the actual output without any grep, so we can confirm the output doesnt changes and behaves properly

libr/core/cmd_open.c
```diff
@@ -570,10 +570,10 @@ static void cmd_open_map(RCore *core, const char *input) {
...
 		map = r_io_map_get (core->io, core->offset);
 		if (map) {
-			core->print->cb_printf ("map: %i fd: %i +0x%"PFMT64x" 0x%"PFMT64x
-				" - 0x%"PFMT64x" ; %s : %s\n", map->id, map->fd,
-				map->delta, map->itv.addr, r_itv_end (map->itv),
-			r_str_rwx_i (map->perm), map->name ? map->name : "");
+			core->print->cb_printf ("%2d fd: %i +0x%08"PFMT64x" 0x%08"PFMT64x
+				" - 0x%08"PFMT64x" %s %s\n", map->id, map->fd,
+				map->delta, map->itv.addr, r_itv_end (map->itv) - 1,
+				r_str_rwx_i (map->perm), (map->name ? map->name : ""));
```
trufae, line 576: use r_str_get2(map->name)

libr/core/cmd_open.c
```diff
@@ -570,10 +570,10 @@ static void cmd_open_map(RCore *core, const char *input) {
 	case '.': // "om."
 		map = r_io_map_get (core->io, core->offset);
 		if (map) {
-			core->print->cb_printf ("map: %i fd: %i +0x%"PFMT64x" 0x%"PFMT64x
-				" - 0x%"PFMT64x" ; %s : %s\n", map->id, map->fd,
-				map->delta, map->itv.addr, r_itv_end (map->itv),
-			r_str_rwx_i (map->perm), map->name ? map->name : "");
+			core->print->cb_printf ("%2d fd: %i +0x%08"PFMT64x" 0x%08"PFMT64x
+				" - 0x%08"PFMT64x" %s %s\n", map->id, map->fd,
+				map->delta, map->itv.addr, r_itv_end (map->itv) - 1,
```
trufae, line 575: i think we can make the code more readable if we add some macros in r_io.h like this: 
```c
#define r_io_map_get_from(map) map->itv.addr
#define r_io_map_get_to(map) r_itv_end (map->itv) - 1
...
```

What do you think @condret ? I never liked this weak usage of the RInterval structure, accessing direct methods and substracting values is a source of problems because any change in the behaviour will require updating all the uses of this api.

trufae: We already have r_itv_begin(&map->itv);, so that maybe probably better usage than acessing fields directly. but having the macros (or static functions in the r_io.h) will make it a bit cleaner maybe. That's up for discussion in here

libr/core/cmd_open.c
trufae: btw r_str_get2 is a pretty bad name for a function. any proposal to change this?

trufae (changes requested):

test/db/formats/pe/sectiondata
condret: yeah, but that is out of scope of this pr imo

libr/core/cmd_open.c
condret: r_str_get_nonnull, I can do this in a different pr, I intend doing some changes on omj anyway, I'd do it within that

libr/core/cmd_open.c
condret: having such functions like r_io_map_begin and r_io_map_end would be great, we should note that down somewhere

Lazula:
I've added the macros and r_str_get2() as requested. Discussion of r_io map address macros is ongoing with @condret. Any more changes are likely outside the scope of this PR, so it should be ready if there are no further concerns.

test/db/formats/pe/sectiondata
trufae: Having the full output of om and om. in a single test is useful and doesnt cost more than 4 lines, 3 of them copypasted

trufae (changes requested):

libr/include/r_io.h
```diff
@@ -15,6 +15,9 @@
 
 #define R_IO_UNDOS 64
 
+#define r_io_map_get_from(map) map->itv.addr
+#define r_io_map_get_to(map) ( r_itv_end (map->itv) - 1 )
```
trufae, line 19: Feedback on this change @condret ?

trufae (changes requested):

libr/include/r_io.h
condret: I like this, it kinda ressembles the old map->to and map->from

trufae (changes requested):
Just add a test with the full output of om;om.

Lazula:
Added  @condret's new tests. No problems in r2r, should be ready to merge.

trufae (approved):

PR #18005, opened 2020-12-07
Add om.j command

Lazula:
Very simple change, just borrowed the existing json output from map_list(). @condret says no tests needed for this one, so it should be ready to merge immediately after the om. update.

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -571,10 +571,17 @@ static void cmd_open_map(RCore *core, const char *input) {
 	case '.': // "om."
 		map = r_io_map_get (core->io, core->offset);
 		if (map) {
-			core->print->cb_printf ("%2d fd: %i +0x%08"PFMT64x" 0x%08"PFMT64x
-				" - 0x%08"PFMT64x" %s %s\n", map->id, map->fd,
-				map->delta, r_io_map_get_from (map), r_io_map_get_to (map),
-				r_str_rwx_i (map->perm), r_str_get2 (map->name));
+			if (input[2] == 'j') { // "om.j"
+				core->print->cb_printf ("{\"map\":%2d,\"fd\":%d,\"delta\":%"PFMT64u",\"from\":%"PFMT64u
+					",\"to\":%"PFMT64u",\"perm\":\"%s\",\"name\":\"%s\"}\n", map->id, map->fd,
+					map->delta, r_io_map_get_from (map), r_itv_end (map->itv),
+					r_str_rwx_i (map->perm), r_str_get2 (map->name));
+			} else {
+				core->print->cb_printf ("%2d fd: %i +0x%08"PFMT64x" 0x%08"PFMT64x
+					" - 0x%08"PFMT64x" %s %s\n", map->id, map->fd,
+					map->delta, r_io_map_get_from (map), r_io_map_get_to (map),
+					r_str_rwx_i (map->perm), r_str_get2 (map->name));
+			}
 		}
 		break;
 	case 'r': // "omr"
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -1,12 +1,7 @@
 /* radare - LGPL - Copyright 2009-2020 - pancake */
 
-#include "r_list.h"
-#include "r_config.h"
-#include "r_core.h"
-#include "r_util.h"
-#include "r_bin.h"
-#include "r_debug.h"
-#include "r_io.h"
+#include <r_bin.h>
+#include <r_debug.h>
 
 static const char *help_msg_o[] = {
 	"Usage: o","[com- ] [file] ([offset])","",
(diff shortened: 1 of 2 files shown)


libr/core/cmd_open.c
```diff
@@ -6,6 +6,7 @@
 #include "r_util.h"
 #include "r_bin.h"
 #include "r_debug.h"
+#include "r_io.h"
```
trufae, line 9: please reduce the includes and use `<>` instead of "" (note that rbin uses io and debug uses util, so imho just including rbin and debug shuold be enough

libr/core/cmd_open.c
```diff
@@ -453,8 +454,8 @@ static void map_list(RIO *io, int mode, RPrint *print, int fd) {
 		default:
 			print->cb_printf ("%2d fd: %i +0x%08"PFMT64x" 0x%08"PFMT64x
 					" - 0x%08"PFMT64x" %s %s\n", map->id, map->fd,
-					map->delta, map->itv.addr, r_itv_end (map->itv) - 1,
-					r_str_rwx_i (map->perm), (map->name ? map->name : ""));
+					map->delta, r_io_map_get_from (map), r_io_map_get_to (map),
```
trufae, line 457: can you do this chagne across the whole r2 codebase? this will make it more grep friendly

libr/core/cmd_open.c
```diff
@@ -570,10 +571,17 @@ static void cmd_open_map(RCore *core, const char *input) {
 	case '.': // "om."
 		map = r_io_map_get (core->io, core->offset);
 		if (map) {
-			core->print->cb_printf ("map: %i fd: %i +0x%"PFMT64x" 0x%"PFMT64x
-				" - 0x%"PFMT64x" ; %s : %s\n", map->id, map->fd,
-				map->delta, map->itv.addr, r_itv_end (map->itv),
-			r_str_rwx_i (map->perm), map->name ? map->name : "");
+			if (input[2] == 'j') { // "om.j"
+				core->print->cb_printf ("{\"map\":%2d,\"fd\":%d,\"delta\":%"PFMT64u",\"from\":%"PFMT64u
```
trufae, line 575: please use the PJ api to construct the JSON output

libr/include/r_io.h
```diff
@@ -15,6 +15,9 @@
 
 #define R_IO_UNDOS 64
 
+#define r_io_map_get_from(map) map->itv.addr
```
trufae, line 18: this seems to overlap with your other PR will cause a conflict when trying to merge, so imho is better to wait for the other pr to be merged (just add one test and wait for travis) and then rebase this pr without the dupe stuff

trufae (changes requested):

libr/include/r_io.h
condret: really, those are the same commits, afaik git can handle that, you just need to merge the other pr first

libr/core/cmd_open.c
condret: use <> is good, but why reduce the includes. What if r_bin changes in the future so that it no longer needs r_io?

libr/core/cmd_open.c
condret: which PJ api?

trufae:
Please rebase and resolve conflicts

libr/core/cmd_open.c
trufae: anything can change in the future, but for now i think its better to keep the minimum includes as possible, compilers will also appreciate doing less work

libr/core/cmd_open.c
trufae: $ git grep 'map->itv'|wc -l
     153 

this will make the PR quite big, so better do that in a separate PR. ignore this comment for now. would you like to do this change?

libr/core/cmd_open.c
trufae: the only pj api we have (libr/util/pj.c) see 046a1fe32474246e982e3bab313028f9631917b4 as an example

libr/include/r_io.h
trufae: git creates a conflict if i merge the other pr :P

trufae:
Btw the only test i would add for this PR is this:

```
echo om.j >> test/db/json/json3
```

thanks!!

libr/core/cmd_open.c
Lazula: I'll get this in the next PR. I will also include a new macro in r_io.h then, so we'll have:

```
#define r_io_map_get_from(map) map->itv.addr
#define r_io_map_get_to(map) ( r_itv_end (map->itv) - 1 )
#define r_io_map_get_end(map) r_itv_end (map->itv)
```

which should make things a lot cleaner since there's a lot of `map->itv.addr + map->itv.size`, `r_itv_end(map->itv)` and similar.

Lazula:
Rebased to master, added the test, modified the headers, and switched to using PJ in both cmd_open_map() and map_list(). Should be ready.

trufae (approved):
Perfect! 👌

PR #18020, opened 2020-12-11
Use r_name_check() in r_flag after filtering the flagname ##core

trufae:
**Checklist**

- [x] Closing issues: #issue

**Description**

With this change, flags can start with lowerdash and semicolon, but avoids the use of flags starting with '$'.

Before this patch it was possible to do things like `f 123=321` which was totally confusing and useless because RNumCalc is not resolving numbers from flags.

But having : and _ as flag prefixes seems useful to me. I added tests for all that stuff, also invalid flags are no longer created with different name, they just throw an error instead of silently messing other flags.


trufae:
Only one broken test which must be fixed because it's imho wrong assumption.

PR #18046, opened 2020-12-16
Fix #17997 - Do not use system binary for testing the debugger

### LINKED ISSUE #17997: Remove/Change tests that depend on special version of local installations
There are tests in the testsuite that depend on locally installed binaries, such as /bin/ls, which cause failing tests on different platforms. It's frustrating a lot

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**


trufae:
Only this test is failing. probably because the binary is not the exact same... but testing against ubuntu-latest makes it not reliable either, so maybe its better to just specify ubuntu-focal in the CI and use this binary updating the test.

PR #18049, opened 2020-12-17
Clean up r_num ##util ##refactor

Lazula:
**Checklist**

**Description**
* Fix function spacing for r_num functions
* Remove r_num_ntohs
* Rename r_num_to_trits to r_num_to_ternary for
  consistency with surrounding code
* Rename trits to ternary in other files
* Fix spacing in test_contrbtree.c


trufae:
Ideally there shouldnt be two files with the same name. there are two technical reasons for that:

* a backtrace with symbols is more readable with just the filename to know which specific num is refering to
* some archivers and linkers fail to repackage archives so we do flat archive linking which requires no 2 files with the same name
* there must be a test using find |grep whatever to avoid that to happen. 
* maybe we can prefix all the files with the first 1-2 letters of each library. to make it "generic"

here's a thing:

```
$ find * | grep num.c
libr/util/unum.c
shlr/sdb/src/num.c
test/unit/test_unum.c
```

trufae (changes requested):
just keep unum.c and ill merge, the rest of changes lgtm!

PR #18053, opened 2020-12-17
Fix arm32 db/cmd/cmd_pc

eagleoflqj:
**Checklist**

- [ ] Closing issues: #issue

**Description**

#17898 
The returned string are all used to format the return value of `r_read_ble`, which is ut64.
```sh
pi@liumeo-rpi4:~/github/radare2/test $ r2r db/cmd/cmd_pc
Loaded 15 tests.
[15/15]                    15 OK         0 BR        0 XX        0 FX
Finished in 0 seconds.
```

### DIFF
--- a/libr/util/print_code.c
+++ b/libr/util/print_code.c
@@ -6,13 +6,13 @@
 static const char* bits_to_c_code_fmtstr(int bits) {
 	switch (bits) {
 	case 16:
-		return "0x%04x";
+		return "0x%04" PFMT64x;
 	case 32:
-		return "0x%08xU";
+		return "0x%08" PFMT64x "U";
 	case 64:
 		return "0x%016" PFMT64x "ULL";
 	}
-	return "0x%02x";
+	return "0x%02" PFMT64x;
 }
 
 static int get_instruction_size(RPrint *p, ut64 at) {


libr/util/print_code.c
```diff
@@ -6,13 +6,13 @@
 static const char* bits_to_c_code_fmtstr(int bits) {
```
trufae, line 6: Isnt this firmat string used against int bits? Why does it needs to be 64? Maybe the fix is in the caller

eagleoflqj: https://github.com/radareorg/radare2/blob/56cdc7c112e5a51ce71715dcf84252d016155076/libr/util/print_code.c#L26-L46
https://github.com/radareorg/radare2/blob/56cdc7c112e5a51ce71715dcf84252d016155076/libr/util/print_code.c#L72-L83
These are the only 2 places that call `bits_to_c_code_fmtstr`. `fmtstr` is used to format the return value of r_read_ble, which is ut64. Either modify like this PR, or do `(int)r_read_ble` are OK, up to you.

trufae: as long as the argument is "int" i would expect the function to return a format string for `int`.  so casting is probably better and more implicit. can you update this. also its a waste of bits to represent the number 64 in 64bits :D

trufae (changes requested):

libr/util/print_code.c
eagleoflqj: Oh, I was wrong. Line 83 we cannot know the value of `bits`, so we cannot do the cast. Have to waste bits here.

> as long as the argument is "int" i would expect the function to return a format string for `int`. so casting is probably better and more implicit. can you update this. also its a waste of bits to represent the number 64 in 64bits :D

trufae: a value bigger than 64 shouldnt be valid, to specify the bits. maybe at some point we may have 128 or 256. but none of this value need 64bit variables to hold this. I mean i understand the fix, but i dont want to waste that much bits and use bigger format strings for that case. You can use & UT32_MAX to wrap the 64bit value and just get the lowerbits to get the format string if that's what worrying you.

trufae (approved):
nvm i confused int bits with the value passed to the format string. i think we are good like this.

thanks!

PR #18055, opened 2020-12-18
Fix arm32 db/cmd/regexp

eagleoflqj:
**Checklist**

- [ ] Closing issues: #issue

**Description**

#17898 
It's a preprocess problem. Due to the included headers, `libr/search/regexp.c` see `RRegexMatch`'s `off_t` field as 64-bit, while `libr/util/regex/regexec.c` see the `off_t` as 32-bit. This can be noticed by comparing the output of `gcc -I../include -I../../shlr/sdb/src -E regexp.c` and `gcc -I../../include -E regexec.c`.
The macro that produces this difference is `__USE_FILE_OFFSET64`. We can run `gcc -I../include -I../../shlr/sdb/src -E -dM regexp.c` and `gcc -I../../include -E -dM regexec.c` to see the difference.
The solution is to force 64-bit.
```sh
pi@liumeo-rpi4:~/github/radare2/test $ r2r -t 0 db/cmd/regexp
Loaded 4 tests.
[4/4]                       4 OK         0 BR        0 XX        0 FX
Finished in 0 seconds.
```

### DIFF
--- a/libr/include/r_regex.h
+++ b/libr/include/r_regex.h
@@ -13,8 +13,8 @@ typedef struct r_regex_t {
 } RRegex;
 
 typedef struct r_regmatch_t {
-	off_t rm_so;		/* start of match */
-	off_t rm_eo;		/* end of match */
+	st64 rm_so;		/* start of match */
+	st64 rm_eo;		/* end of match */
 } RRegexMatch;
 
 /* regcomp() flags */


libr/include/r_regex.h
```diff
@@ -13,8 +13,8 @@ typedef struct r_regex_t {
 } RRegex;
 
 typedef struct r_regmatch_t {
-	off_t rm_so;		/* start of match */
-	off_t rm_eo;		/* end of match */
+	st64 rm_so;		/* start of match */
```
trufae, line 16: why is this failing? the change shuoldnt introduce any regression, but i dont know which is the issue that is trying to solve

eagleoflqj: Due to other included headers, in regexp.c `__USE_FILE_OFFSET64` is defined before including r_regex.h so off_t is resolved to 4 byte, in regexec.c `__USE_FILE_OFFSET64` is not defined so off_t is resolved to 8 byte. You can run the commands I mentioned above to see the preprocess result and all visible macros on your rpi.

trufae: I see :) good finding. Thanks!

PR #18056, opened 2020-12-18
Fix arm32 db/cmd/cmd_i

eagleoflqj:
**Checklist**

- [ ] Closing issues: #issue

**Description**

#17898 
Fix all tests of db/cmd/cmd_i, for various reasons:
* `r_table_add_rowf` wrong format
* `SSIZE_MAX` not defined, so that `#else` branch is applied: https://github.com/radareorg/radare2/blob/56cdc7c112e5a51ce71715dcf84252d016155076/libr/include/r_types_base.h#L104-L114
* wrong limit `sizeof (encoded_str)` which is the size of a pointer passed to `snprintf`. Use `sprintf` is safe since the buffer is big enough and the arg `(ut8) buf[loop]` is 0 to 255.
* assign bit in test
```sh
pi@liumeo-rpi4:~/github/radare2/test $ r2r db/cmd/cmd_i
Loaded 142 tests.
[142/142]                 134 OK         7 BR        0 XX        1 FX
Finished in 6 seconds.
```

### DIFF
--- a/libr/util/str.c
+++ b/libr/util/str.c
@@ -1566,7 +1566,7 @@ R_API char *r_str_encoded_json(const char *buf, int buf_size, int encoding) {
 
 		const char *format = encoding == PJ_ENCODING_STR_ARRAY ? "%03u," : "%02X";
 		while (buf[loop] != '\0' && i < (new_sz - 1)) {
-			sprintf (encoded_str + i, format, (ut8) buf[loop]);
+			snprintf (encoded_str + i, new_sz - i, format, (ut8) buf[loop]);
 			loop++;
 			i += increment;
 		}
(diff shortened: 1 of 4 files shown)


libr/util/str.c
```diff
@@ -1566,7 +1566,7 @@ R_API char *r_str_encoded_json(const char *buf, int buf_size, int encoding) {
 
 		const char *format = encoding == PJ_ENCODING_STR_ARRAY ? "%03u," : "%02X";
 		while (buf[loop] != '\0' && i < (new_sz - 1)) {
-			snprintf (encoded_str + i, sizeof (encoded_str), format, (ut8) buf[loop]);
+			sprintf (encoded_str + i, format, (ut8) buf[loop]);
```
trufae, line 1569:
```suggestion
			snprintf (encoded_str + i,  new_sz - i, (ut8) buf[loop]);
```

trufae (changes requested):

libr/util/str.c
eagleoflqj: Why must snprintf? Here every time 2 or 4 chars plus a \0 is written, as we already calculated length no overflow happens. Checking boundary has overhead.

trufae: all static code analysis tool report using unsafe C functions as an issue. even if it's safe because you know it wont overflow, its never a good idea to use them

eagleoflqj: Sounds reasonable.

PR #18063, opened 2020-12-19
Fix arm32 test_vector_fini

eagleoflqj:
**Checklist**

- [ ] Closing issues: #issue

**Description**

#17898 
Get rid of 64bit specific things.
The macro comes from https://github.com/radareorg/radare2/blob/dcc17e7f327db213f4ad50d7fa358143b69906ad/libr/util/vector.c#L5-L13

### DIFF
--- a/test/unit/test_vector.c
+++ b/test/unit/test_vector.c
@@ -2,12 +2,6 @@
 #include <r_vector.h>
 #include "minunit.h"
 
-#if __WORDSIZE == 32
-#define INITIAL_VECTOR_LEN 3
-#else
-#define INITIAL_VECTOR_LEN 4
-#endif
-
 // allocates a vector of len ut32 values from 0 to len
 // with capacity len + padding
 static bool _init_test_vector(RVector *v, size_t len, size_t padding, RVectorFree free, void *free_user) {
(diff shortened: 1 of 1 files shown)


test/unit/test_vector.c
```diff
@@ -2,6 +2,12 @@
 #include <r_vector.h>
 #include "minunit.h"
 
+#if __WORDSIZE == 32
```
trufae, line 5: I dont see the need for this defines here. Why not just use always 4?

trufae: If the initial capacity differs in 32 and 64bit versions of it something wrong is going on somewhere else i think

trufae (changes requested):

test/unit/test_vector.c
eagleoflqj: > If the initial capacity differs in 32 and 64bit versions of it something wrong is going on somewhere else i think

I updated the description above (forgot to reference line 5, which says the choice of initial capacity is for glibc optimization.
imho if it brings trouble elsewhere, the problem comes from the user of r_vector, rather than r_vector itself. We shouldn't write code whose correctness depends on specific implementation of a library. Ironically this unit test fails on arm32 because it depends on specific implementation :-)

trufae: if its a glibc optimization this test may probably fail in other libcs like for Android or so. so its a bad idea to workaround this in the testsuite. i prefer not to include workarounds

eagleoflqj: It won't fail, since the choice of initial capacity is for every libc but is only optimal on glibc. But I'll remove the test.

trufae (approved):

trufae:
This 3 vs 4 is hardcoded in the vector.c which makes it a really bad decision, because that define just depends on word size, it have nothing to do with the version of libc, or the arch used. so imho this code is wrongly assuming some struct sizes without ensuring anything beforehand.

PR #18066, opened 2020-12-19
Revert "Fix clang-cl build"

pelijah:
My PR radareorg/radare2#17935 must be reverted since it causes a crash after opening binary in debug mode.


yossizap:
Should also set -fcommon to disable -fno-common which is now default in clang-cl 11 and broke the build.

trufae (approved):
Yeah i dont see why function signatures can be static, the revert lgtm

PR #18092, opened 2020-12-22
Add sign extension assignment operator

sylvainpelissier:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Introduce new **~=** operator doing sign extension assignment of a register. For example:
```
> "ae 0x81,a0,="
> "ae 8,a0,~="
> ar a0
0xffffff81
```
This would simplify a lot of ESIL expressions for architecture using sign extension.

### DIFF
--- a/libr/anal/esil.c
+++ b/libr/anal/esil.c
@@ -542,11 +542,8 @@ R_API int r_anal_esil_reg_read(RAnalEsil *esil, const char *regname, ut64 *num, 
 	return ret;
 }
 
-// sign extension operator for use in idiv, imul, movsx* 
-// and other instructions involving signed values, extends n bit value to 64 bit value
-// example : >"ae 8,0x81,~" ( <src bit width>,<value>,~ )
-// output  : 0xffffffffffffff81
-static bool esil_signext(RAnalEsil *esil) {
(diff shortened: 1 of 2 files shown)


trufae (approved):

libr/anal/esil.c
```diff
@@ -587,6 +587,55 @@ static bool esil_signext(RAnalEsil *esil) {
...
 
+// sign extension assignement
+// example : > "ae 0x81,a0,="
+//           > "ae 8,a0,~="   ( <src bit width>,register,~= )
+// output  : > ar a0
+//           0xffffff81
+static bool esil_signexteq(RAnalEsil *esil) {
+	ut64 src, dst;
+
+	char *p_src = r_anal_esil_pop (esil);
```
trufae, line 598: Ideally it will be good to share most code with  `esil_signext`, can you refactor this? we should be able to describe operations in a functional way without having to go into much details in the code. X= operations must be calling internally the X operation and then the = operation for example.

trufae (changes requested):

libr/anal/esil.c
sylvainpelissier: I refactored to share code but doing that in a functional way is not obvious to me.

PR #18097, opened 2020-12-23
Fix ppc opex & use pj in RAnal plugins ##refactor

eagleoflqj:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Thanks @Lazula . I fixed some small bugs, and refactored a bit (see the 3rd commit).
According to https://codecov.io/gh/radareorg/radare2/tree/master/libr/anal/p, only sysz is completely not tested, so I added a test.

### DIFF
--- a/libr/anal/p/anal_ppc_cs.c
+++ b/libr/anal/p/anal_ppc_cs.c
@@ -160,44 +160,49 @@ static const char* getspr(struct Getarg *gop, int n) {
 	return cspr;
 }
 
-//TODO PJ
 static void opex(RStrBuf *buf, csh handle, cs_insn *insn) {
 	int i;
-	r_strbuf_init (buf);
-	r_strbuf_append (buf, "{");
+	PJ *pj;
+	pj = pj_new ();
+	if (!pj) {
+		return;
+	}
 	cs_sysz *x = &insn->detail->sysz;
-	r_strbuf_append (buf, "\"operands\":[");
+	pj_o (pj);
+	pj_ka (pj, "operands");
 	for (i = 0; i < x->op_count; i++) {
 		cs_sysz_op *op = &x->operands[i];
-		if (i > 0) {
-			r_strbuf_append (buf, ",");
-		}
-		r_strbuf_append (buf, "{");
+		pj_o (pj);
 		switch (op->type) {
 		case SYSZ_OP_REG:
-			r_strbuf_append (buf, "\"type\":\"reg\"");
-			r_strbuf_appendf (buf, ",\"value\":\"%s\"", cs_reg_name (handle, op->reg));
+			pj_ks (pj, "type", "reg");
+			pj_ks (pj, "value", cs_reg_name (handle, op->reg));
 			break;
 		case SYSZ_OP_IMM:
-			r_strbuf_append (buf, "\"type\":\"imm\"");
-			r_strbuf_appendf (buf, ",\"value\":%" PFMT64d, (st64)op->imm);
+			pj_ks (pj, "type", "imm");
+			pj_kN (pj, "value", (st64)op->imm);
 			break;
 		case SYSZ_OP_MEM:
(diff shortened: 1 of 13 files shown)


trufae (approved):

libr/anal/p/anal_ppc_cs.c
```diff
@@ -160,44 +160,48 @@ static const char* getspr(struct Getarg *gop, int n) {
...
+	pj_ka (pj, "operands");
 	cs_sysz *x = &insn->detail->sysz;
-	r_strbuf_append (buf, "\"operands\":[");
 	for (i = 0; i < x->op_count; i++) {
-		cs_sysz_op *op = &x->operands[i];
-		if (i > 0) {
-			r_strbuf_append (buf, ",");
-		}
-		r_strbuf_append (buf, "{");
+		cs_sysz_op *op = x->operands + i;
```
eagleoflqj, line 173: Wait, why is this sysz?

Lazula: Probably meant to be cs_ppc_op. pancake takes the git blame for this one, it likely happened when copying over opex functions, as they share very similar code between multiple plugins. Should probably open a new PR for changing that. May want to pull the opex() funcs out of the plugins themselves, too.

eagleoflqj: > Probably meant to be cs_ppc_op. pancake takes the git blame for this one, it likely happened when copying over opex functions, as they share very similar code between multiple plugins. Should probably open a new PR for changing that. May want to pull the opex() funcs out of the plugins themselves, too.

Better fix here as it's a bug :-)
As C lacks reflection in Java, it seems impossible to have a unified opex function because we cannot dynamically know what fields a struct has.

Lazula: Could use a solution similar to CPython's PyObject with a generic struct header. So, there could exist an opcode struct header containing the type, then opex() could check the universal type member, then cast based on it, likely calling other functions based on the type.

For the record, this is well-defined behavior, not a hack. From section 3.5.2.1: Structure and union specifiers of the [C89 draft standard](http://port70.net/~nsz/c/c89/c89-draft.html): "A pointer to a structure object, suitably cast, points to its initial member (or if that member is a bit-field, then to the unit in which it resides), and vice versa. There may therefore be unnamed holes within a structure object, but not at its beginning, as necessary to achieve the appropriate alignment."

You can see how the CPython developers made their code compatible with standard C [here](https://www.python.org/dev/peps/pep-3123/).

While this may or may not be a desirable addition, it is certainly possible to add this functionality.

eagleoflqj: > Could use a solution similar to CPython's PyObject with a generic struct header. So, there could exist an opcode struct header containing the type, then opex() could check the universal type member, then cast based on it, likely calling other functions based on the type.
> 
> For the record, this is well-defined behavior, not a hack. From section 3.5.2.1: Structure and union specifiers of the [C89 draft standard](http://port70.net/~nsz/c/c89/c89-draft.html): "A pointer to a structure object, suitably cast, points to its initial member (or if that member is a bit-field, then to the unit in which it resides), and vice versa. There may therefore be unnamed holes within a structure object, but not at its beginning, as necessary to achieve the appropriate alignment."
> 
> You can see how the CPython developers made their code compatible with standard C [here](https://www.python.org/dev/peps/pep-3123/) and the PyObject header file is [here](https://github.com/python/cpython/blob/master/Include/object.h).
> 
> While this may or may not be a desirable addition, it is certainly possible to add this functionality.

I know this way to implement polymorphism in C (C++ just does so), however we still need to **define** different functions to deal with different structs, which is exactly current situation.

Lazula: I may have misunderstood you, I just think that having this output processing in the plugins themselves is not desirable.

PR #18064, opened 2020-12-19
Merge pse to ps

gogo2464:
**Checklist**

- [ ] Closing issues: #issue
- [ ] Mark this if you consider it ready to merge
- [ ] I've added tests (optional)
- [ ] I wrote some lines in the [radare2book](https://github.com/radareorg/radare2book)

**Description**

`e cfg.charset=pokered; ps` changes now the encoding. I also removed a redundant ps declaration. I preview to separate in 2 commits. I want your opinion first.

PR #18100, opened 2020-12-23
remove redundant 'ps ' with a space

gogo2464:
**Checklist**

- [ ] Closing issues: #issue

**Description**

As told here https://github.com/radareorg/radare2/pull/18064 there is a redundant call of ps. When ps is set with no space ps works fine. But with a space, ps is zeroend. It is the behaviour of psz. Not ps. It also need to be refactored for readablitity purpose.

The tests of 'ps' and 'ps ' must also be changed.


gogo2464:
make tests does not correctly work on my system. Then I am using the github tests. I will make another updates for the tests.

gogo2464:
I understand now that the programmer has zeroend the ps command to get the exact size expected by the number. I do not think it was a good idea. I am going to change that.

test/db/cmd/cmd_wox
```diff
@@ -72,7 +72,7 @@ wox 0xabcdef010203@!10
 ps 14
 EOF
 EXPECT=<<EOF
-helloworld\x00\x00\x00\x00
+helloworld
```
trufae, line 75: ps 18 should show the zeroes. can you also testt this, as well as psz as i said in my previous comment?. also this is not testing the charsets at all

trufae (changes requested):

test/db/cmd/cmd_wox
gogo2464: In my opinion ps should not have a different behavior if the space is specified. If we want 0s we should use psz.


This pull request is not for test the charset. I will test the charset in another pull request.

trufae:
Please rebase/resolve the conflicts and i'll try your branch to confirm behaviour.

trufae:
Also commit messages must start in uppercase (and PR titles)

trufae:
As I said, ps is not redundant. it will print N chars as string, ignoring if there are null bytes. think about wide strings or wanting to print all the strings in the current block. dont remove this command

test/db/cmd/cmd_wox
trufae: Exactly. i agree that `ps` and `ps $b` shuold print the same. but your PR is making ps do the same thing as psz. and this is not correct. Can you please make `ps` just take the blocksize as length instead of behaving like psz?

Also, can you make all subcommands of `ps` that are not handled to print an error and do nothing instead of doing something random?

trufae: Screenshot 2020-12-28 at 18 42 53

gogo2464: Ok I understand. I did not know the command psz before. I doubt. I think reverse the effects of ps and psz might be more intuitive. Not sure.

`ps $b` prints the block size. So it is different of course. Why should it be same?

trufae:
what im seeing is that maybe we want ps to print strings skiping the null bytes, but not escaping non printable chars. we have this already in pcs

trufae:
closing, theres a difference between ps and psz already on purpose

PR #18113, opened 2020-12-27
Run the build-spaces job

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Seems like there's no way to make GNU Make work with spaces. there's only $(wildcard) function which supports that. and handling all the other cases will make all the makefiles very ugly to read and maintain, and i think some use cases wont work at all. So i think the best way to go here to address this issue is to create a CI task that runs installs in a directory with spaces using acr/make, and builds on a directory with spaces with meson. which is what it should be supported and working.

anyway, having directories with spaces is an anti-pattern and i totally discourage people to do that.


trufae:
Screenshot 2020-12-28 at 20 11 06

PR #18120, opened 2020-12-28
SDB: In Soviet Russia, C has booleans ##refactor

ratijas:
**Checklist**

- [ ] Closing issues: None

**Description**

Change arguments type from int to bool. Those are internal static function, so no other code could possibly be affected. There are booleans all over the place in that file, so we are not losing any portability.

### DIFF
--- a/shlr/sdb/src/sdb.c
+++ b/shlr/sdb/src/sdb.c
@@ -546,7 +546,7 @@ SDB_API void sdbkv_free(SdbKv *kv) {
 	}
 }
 
-static ut32 sdb_set_internal(Sdb* s, const char *key, char *val, int owned, ut32 cas) {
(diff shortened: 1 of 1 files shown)


shlr/sdb/src/sdb.c
```diff
@@ -184,7 +184,7 @@ SDB_API int sdb_count(Sdb *s) {
 	return count;
 }
 
-static void sdb_fini(Sdb* s, int donull) {
+static void sdb_fini(Sdb* s, bool donull) {
```
trufae, line 187: those files are from a different repository. dont patch them in here. make the PR in https://github.com/radareorg/sdb

trufae:
Closing, as this PR should be done in https://github.com/radareorg/sdb instead

ratijas:
Aww, thanks for the heads up. Gotta send it there.

Issue #17034, opened 2020-06-10
Remove "projects" feature until it's implemented properly.


XVilka:
It doesn't work anyway, proper implementation will require some time, people try this feature and go away disappointed because it doesn't work in 98% of time. I think it's better to remove them completely until they are properly implemented.
- [ ] Remove the code
- [ ] Remove the commands
- [ ] Remove mentions from manpages and help
- [ ] Remove mentions from radare2book
- [ ] Go through radare2-extras and r2pm to remove all usages if any.
- [ ] Make a public announcement on blog, twitter, social networks about removal of the feature.

unixfreaxjp:
I taught people who learn r2 in beginners level in my community to create project to save their stuff.
I understand the situation, as users, so I am not against this feature's deletion due to mentioned circumstance.

However, please make a proper announcement if the project feature is completely removed. So I can refer it to documentation that I made. Adding "announcement" in these tasks will be appreciated.

>    Remove the code
>    Remove the commands
>    Remove mentions from manpages and help
>    Remove mentions from radare2book
>    Go through radare2-extras and r2pm to remove all usages if any.

@XVilka 
cc: @trufae

thestr4ng3r:
fyi a rewrite of projects is currently in process: https://github.com/thestr4ng3r/r2db

ulidtko:
This gets a strong bold **-1** from me.

Main reason: despite being imperfect and buggy (like the whole radare2 itself), it's still very useful.

Secondary reason: I fail to see what the removal will achieve besides inflicting suffering. `Ps`/`Po` can be incrementally improved IMO.

> [...] To make it work it would require huge amounts of tests, considering every corner case.

No; this is exactly the use-case for **fuzz-testing**. With not much code, it'll find all those corner cases for you. With not much code, it can find `Ps` idempotence violations, `Ps; Po` identity violations, version regressions and whatever other properties one cares to formulate.

I agree with "something needs to be done with r2 projects" — but as soon as the "something" takes the value of "remove for indeterminate time", I disagree strongly.

ret2libc:
Did we reach any conclusion on this? I'm personally for removing the feature because it is as unreliable as possible, thus something I would not trust using in any case. However, if users think it's better to keep the feature as is until a replacement will be available, I'm ok as well.

Let's try to reach an agreement in less than one month, so we can either close this issue or work on removing the feature.

PR #18134, opened 2020-12-29
Tell the user that debugging projects don't work ##projects

RHL120:
**Checklist**

- [X] Closing issues: #issue

**Description**
Now the user won't be able to save debugging/emulating projects instead they will get an error.
I think this is kindof related to this issue https://github.com/radareorg/radare2/issues/17034

### DIFF
--- a/libr/core/project.c
+++ b/libr/core/project.c
@@ -770,14 +770,10 @@ R_API bool r_core_project_save(RCore *core, const char *prj_name) {
 	SdbListIter *it;
 	SdbNs *ns;
 	char *old_prj_name = NULL;
-	r_return_val_if_fail(prj_name && *prj_name, false);
+	r_return_val_if_fail (prj_name && *prj_name, false);
 	script_path = get_project_script_path (core, prj_name);
-	if (core->bin->is_debugger) {
+	if (r_config_get_i (core->config, "cfg.debug")) {
 		eprintf ("radare2 does not support projects on debugged bins\n");
-		return false;
-	}
-	if (r_config_get_i (core->config, "asm.emu")) {
-		eprintf ("radare2 does not support projects on emulated bins\n");
 		return false;
 	}
 	if (!script_path) {
--- a/libr/core/project.c
+++ b/libr/core/project.c
@@ -659,13 +659,15 @@ static bool project_save_script(RCore *core, const char *file, int opts) {
 	r_cons_singleton ()->context->is_interactive = false;
 	r_str_write (fd, "# r2 rdb project file\n");
 	if (core->bin->file) {
-		char *fpath = r_file_abspath (core->bin->file);
-		if (fpath) {
-			char *reopen = r_str_newf ("\"o %s\"\n", fpath);
-			if (reopen) {
-				r_str_write (fd, reopen);
-				free (reopen);
-				free (fpath);
+		if (r_config_get_i(core->config, "asm.emu")) {
+			char *fpath = r_file_abspath (core->bin->file);
+			if (fpath) {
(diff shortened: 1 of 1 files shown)


trufae:
Can you join telegram/discord/irc to chat about projects?

RHL120:
> Can you join telegram/discord/irc to chat about projects?

Yeah, I already joined telegeram under the same name

libr/core/project.c
```diff
@@ -774,6 +772,14 @@ R_API bool r_core_project_save(RCore *core, const char *prj_name) {
 	char *old_prj_name = NULL;
 	r_return_val_if_fail(prj_name && *prj_name, false);
```
trufae, line 773: missing space before (

libr/core/project.c
```diff
@@ -774,6 +772,14 @@ R_API bool r_core_project_save(RCore *core, const char *prj_name) {
...
 	script_path = get_project_script_path (core, prj_name);
+	if (core->bin->is_debugger) {
+		eprintf ("radare2 does not support projects on debugged bins\n");
+		return false;
+	}
+	if (r_config_get_i (core->config, "asm.emu")) {
```
trufae, line 779: this variable just makes the disasm emulate the selected instructions to show computed references. i dont see why an emulated binary shouldnt be a valid project. Also, what is the problem for not supporting debug/emulation sessions? both are very common use cases and saving/restoring files and flags and anal info shouldnt be affected by it.

In any case i'm fine for forbiddnig to save a project when in debugger mode. but i dont see why the emulated environment shouldnt be supported. (the only reason is missing iomaps for stack and other stuff like that) and this is the main reason why we dont support projects well because rbin doesnt hold this information and core cant corrlate io and sort its operational orders. and refactoring io and bin are max prio in order to support any kind of environment configured in r2 as in projects.

libr/core/project.c
```diff
@@ -774,6 +772,14 @@ R_API bool r_core_project_save(RCore *core, const char *prj_name) {
 	char *old_prj_name = NULL;
 	r_return_val_if_fail(prj_name && *prj_name, false);
 	script_path = get_project_script_path (core, prj_name);
+	if (core->bin->is_debugger) {
```
trufae, line 775: i dont think rbin should have this variable at all. imho it must be killed in a further refactoring. use cfg.debug instead.

trufae (changes requested):

libr/core/project.c
RHL120: I see your point and I will remove the lines that forbid emulation but I would like to say that when it comes down to the general case emulation and projects are broken together (right now at least). We can allow emulation when a work around has been found.

RHL120: It is worth pointing out that even though saving a debug project is in fact very useful, it is currently broken so there would be no point in giving the user a false sense of comfort that their "project" is saved and can be reopend easily.

libr/core/project.c
```diff
@@ -658,8 +658,8 @@ static bool project_save_script(RCore *core, const char *file, int opts) {
 	r_cons_singleton ()->fdout = fd;
 	r_cons_singleton ()->context->is_interactive = false;
 	r_str_write (fd, "# r2 rdb project file\n");
-	if (!core->bin->is_debugger && !r_config_get_i (core->config, "asm.emu")) {
-		if (core->bin->file) {
+	if (core->bin->file) {
+		if (!r_config_get_i(core->config, "asm.emu")) {
```
trufae, line 662:
```suggestion
		if (!r_config_get_i (core->config, "asm.emu")) {
```

PR #18137, opened 2020-12-29
Automatically set prj.git if `git` is found in PATH ##projects

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**


RHL120:
I think you should check the ret val of system("git config --global user.name && git config --global user.email")

trufae:
> I think you should check the ret val of system("git config --global user.name && git config --global user.email")

this is not related to this PR, but i think we should have git projects enabled by default for the users. but in the CI we should test projects with and without git support. Comments on this?

trufae:
ping @RHL120

trufae:
Let's merge this, so we can start to improve the git integration and avoid mistakes that make people loose data when saving a project.

PR #18143, opened 2020-12-30
Use pj in cbin.c ##refactor

eagleoflqj:
**Checklist**

- [ ] Closing issues: #issue

**Description**


trufae:
I'm fine to merge it like this because the change is big, but i see two changes that may happen after this refactor:

* pair functions can be probably removed now that they are just a wrapper of `pj`.
* use r_return and R_STR_ISEMPTY instead of manual precondition nullchecks

Issue #18029, opened 2020-12-13
Esil Syscalls


condret:
So far, esil only has an operation for interrupts `$`, but not for syscalls. I propose `()` as operation for syscalls.
This feature requires:
- [ ] r_anal_esil_do_syscall (like r_anal_esil_fire_interrupt)
- [ ] `()` operation in esil
- [ ] load commands for syscall implementation like aeli (propose: aels)

condret:
implemented

trufae:
i marked the three tasks as done, but the last pr is not yet merged. and will be good to introduce a basic plugin with some syscalls implemented too

PR #18144, opened 2020-12-31
Implement esil syscalls #18029 ##anal

condret:
**Checklist**

- [x] Closing issues: #18029

**Description**

Adds API-functions and commands to load esil syscall handlers from shared objects. Also fixes a few bugs, i found in `aeli`-commands.

Will provide an example after committing some sleep.


libr/core/cmd_anal.c
```diff
@@ -6552,21 +6559,63 @@ static void cmd_anal_esil(RCore *core, const char *input) {
...
+						_aeli_print (0ULL, esil->sysc0);
+					}
+					dict_foreach (esil->syscalls, _aeli_iter, NULL);
+				}
+				break;
+			case 'r': // "aelsr"
+				if (esil && esil->syscalls) {
+					const ut32 sysc_num = r_num_math (core->num, input + 3);
+					if (sysc_num) {
+						RAnalEsilSyscall *syscall = dict_getu (esil->syscalls, sysc_num);
```
trufae, line 6605: Use `sc` instead of `syscall` which is defined in unistd.h and may result on compilation issues or reader confussion

libr/core/cmd_anal.c
```diff
@@ -6552,21 +6559,63 @@ static void cmd_anal_esil(RCore *core, const char *input) {
...
+						r_anal_esil_syscall_free (esil, syscall);
+					} else {
+						r_anal_esil_syscall_free (esil, esil->sysc0);
```
trufae, line 6608: just call _free once. just set RAnalEsilSyscall *sc = once

libr/core/cmd_anal.c
```diff
@@ -6552,21 +6559,63 @@ static void cmd_anal_esil(RCore *core, const char *input) {
...
+					}
 				}
 				break;
 			}
+			break;
+		case ' ': // "ael"
+			if (r_anal_esil_load_handlers_from_lib (esil, input + 2)) {
```
trufae, line 6615: should return false or NULL on error

trufae, line 6615: r_str_trim_head_ro(input + 2) to skip spaces

libr/core/cmd_anal.c
```diff
@@ -6552,21 +6559,63 @@ static void cmd_anal_esil(RCore *core, const char *input) {
...
 					dict_foreach (esil->interrupts, _aeli_iter, NULL);
 				}
 				break;
 			case 'r': // "aelir"
 				if (esil && esil->interrupts) {
-					RAnalEsilInterrupt* interrupt = dict_getu (esil->interrupts, r_num_math (core->num, input + 3));
-					r_anal_esil_interrupt_free (esil, interrupt);
+					const ut32 intr_num = r_num_math (core->num, input + 3);
+					if (intr_num) {
+						RAnalEsilInterrupt* interrupt = dict_getu (esil->interrupts, intr_num);
```
trufae, line 6577: Use `ei` as for `esil-interrupt` instead of `interrupt`.

libr/anal/esil_interrupt.c
```diff
@@ -136,12 +214,60 @@ R_API bool r_anal_esil_load_interrupts_from_lib(RAnalEsil *esil, const char *pat
...
+		return false;
+	}
+	bool ret = true;
+	if (ihandlers) {
+		ret &= r_anal_esil_load_interrupts (esil, ihandlers, src_id);
+	}
+	if (shandlers) {
+		ret &= r_anal_esil_load_syscalls (esil, shandlers, src_id);
+	}
+	return ret;
```
trufae, line 253: imho ret should be false if !ihandlers && !shandlers too

libr/anal/esil.c
```diff
@@ -995,6 +995,14 @@ static bool esil_interrupt(RAnalEsil *esil) {
 	return false;
 }
 
+static bool esil_syscall(RAnalEsil *esil) {
+	ut64 syscall;
+	if (popRN (esil, &syscall)) {
+		return r_anal_esil_do_syscall (esil, (ut32)syscall);
```
trufae, line 1001: why ut32 and dont name it `syscall` its a reserved symbol from libc

libr/anal/esil_interrupt.c
```diff
@@ -52,34 +73,62 @@ R_API RAnalEsilInterrupt *r_anal_esil_interrupt_new(RAnalEsil *esil, ut32 src_id
...
 	}
 	free (intr);
 }
 
+R_API void r_anal_esil_syscall_free(RAnalEsil *esil, RAnalEsilSyscall *sysc) {
+	if (sysc && esil) {
+		_del_syscall (esil, sysc->handler->num);
+	}
+	if (sysc) {
+		if (sysc->user) {
```
trufae, line 98: sysc->handler and handler->fini can be null and this may lead to a crash

libr/anal/esil_interrupt.c
```diff
@@ -52,34 +73,62 @@ R_API RAnalEsilInterrupt *r_anal_esil_interrupt_new(RAnalEsil *esil, ut32 src_id
...
-	return _set_interrupt(esil, intr);
+	return _set_interrupt (esil, intr);
+}
+
+R_API bool r_anal_esil_set_syscall(RAnalEsil *esil, RAnalEsilSyscall *sysc) {
+	r_return_val_if_fail (esil && esil->syscalls && sysc && sysc->handler && sysc->handler->cb, false);
+	// check if interrupt is already set
+	RAnalEsilSyscall *o_sysc = _get_syscall (esil, sysc->handler->num);
+	if (o_sysc) {
+		r_anal_esil_syscall_free (esil, o_sysc);
```
trufae, line 122: its always fine to call free with just NULL. but free functions should take only one parameter. otherwise swap the arguments and put `free` in the middle like this: `r_anal_esil_free_syscall(esil, o_sysc);` or make Syscall hold an instance to ESIL

libr/anal/esil_interrupt.c
```diff
@@ -94,18 +143,27 @@ R_API int r_anal_esil_fire_interrupt(RAnalEsil *esil, ut32 intr_num) {
...
 }
 
-R_API bool r_anal_esil_load_interrupts (RAnalEsil *esil, RAnalEsilInterruptHandler *handlers[], ut32 src_id) {
-	RAnalEsilInterrupt *intr;
-	ut32 i = 0;
+R_API int r_anal_esil_do_syscall(RAnalEsil *esil, ut32 sysc_num) {
+	r_return_val_if_fail (esil, false);
 
+	if (!esil->syscalls) {
+		eprintf ("no syscalls initialized\n");
```
trufae, line 153: "syscalls not initialized"

libr/anal/esil_interrupt.c
```diff
@@ -121,13 +179,33 @@ R_API bool r_anal_esil_load_interrupts (RAnalEsil *esil, RAnalEsilInterruptHandl
 	return true;
 }
 
+R_API bool r_anal_esil_load_syscalls(RAnalEsil *esil, RAnalEsilHandler *handlers[], ut32 src_id) {
+	r_return_val_if_fail (esil && esil->syscalls && handlers, false);
+
+	ut32 i = 0;
+	while (handlers[i]) {
+		RAnalEsilSyscall *sysc = _get_syscall (esil, handlers[i]->num);
+		if (sysc) {
```
trufae, line 188: unnecessary if

libr/anal/esil_interrupt.c
```diff
@@ -136,12 +214,60 @@ R_API bool r_anal_esil_load_interrupts_from_lib(RAnalEsil *esil, const char *pat
...
-		dict_free (esil->interrupts);
-		esil->interrupts = NULL;
+R_API bool r_anal_esil_load_syscalls_from_lib(RAnalEsil *esil, const char *path) {
+	r_return_val_if_fail (esil, false);
+	const ut32 src_id = r_anal_esil_load_source (esil, path);
+	if (!src_id) {
+		return false;
+	}
+	RAnalEsilHandler **handlers = (RAnalEsilHandler **)
+		r_lib_dl_sym (r_anal_esil_get_source (esil, src_id), "syscalls");
```
trufae, line 224: imho its a bad idea to use low level r_lib_dl this can lead to resolving symbols on NULL and it would be better if we could define syscalls without depending on native code. maybe having an rlang interface for it would be ideal to use py/js or other safe languages to do so. What do you think=?

trufae (changes requested):

trufae:
To summarize what we discussed on phone:

 Those syscall libraries should be implemented as r2 plugins, so we can use the `L` command to load the plugins and register all the arch-bits-kernel triplet info. So the commands will just pick which syscall implementation you want to use from the already loaded plugins, instead of directly loading native code at runtime which is a bit unsafe.

I can help you on doing that part i mean, i created the plugins api and did all the plugin implementations. so i understand this can be ab it scary. The problem is where shuold this implementation go. also it s better to define getters/setters rather than direct array accesses. this way the api can provide a getter to retrieve the item needed.

There must be an rlang wrapper around this, so js/py plugins can implement syscalls. (obv this stuff is out of the scope of this PR)

condret:
I doubt that rlang wrapper could be done easlily, unless you throw everything else ,except setting handlers, under the bus

libr/anal/esil_interrupt.c
condret: I want syscall handlers that can interact with libr to full extend. They're supposed to simulate lowlevel things that the OS does

libr/anal/esil.c
```diff
@@ -1022,6 +1020,14 @@ static bool esil_interrupt(RAnalEsil *esil) {
 	return false;
 }
 
+static bool esil_syscall(RAnalEsil *esil) {
+	ut64 syscall;
```
trufae, line 1024:
```suggestion
	ut64 sc;
```

libr/anal/esil.c
```diff
@@ -1022,6 +1020,14 @@ static bool esil_interrupt(RAnalEsil *esil) {
...
+	if (popRN (esil, &syscall)) {
```
trufae, line 1025:
```suggestion
	if (popRN (esil, &sc)) {
```

libr/anal/esil.c
```diff
@@ -1022,6 +1020,14 @@ static bool esil_interrupt(RAnalEsil *esil) {
...
+		return r_anal_esil_do_syscall (esil, (ut32)syscall);
```
trufae, line 1026:
```suggestion
		return r_anal_esil_do_syscall (esil, (ut32)sc);
```

trufae (approved):

trufae (approved):

PR #18180, opened 2021-01-06
Fix and refactor the ar:// plugin ##io

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This PR fixes the archive io plugin. This will let you open `ar://file.a//obj.o` files again. Tests already existed for this, but were marked broken. This PR updates the tests because formatting has changed since the tests worked. Additionally, a couple of the test appear broken from the start (ELF header should not be at entry0).

The code was refactored a lot. While I hope this makes the code more readable, the primary reason for the refactor is to add in open_many support later. Archive files have no compression. So many files could be supported by the single buffer. The `refcount` member of the `RArFp` structure is mostly unused now. When many file support is added it will ensure the underlying `RBuffer` will be cleaned up when the last reference to it is removed.

I intend on adding at least one more test to this. I don't think any of the current ar tests use a .a file that contains a filename table. I will add the next test in the morning, or on another PR.

### DIFF
--- a/libr/io/p/io_ar.c
+++ b/libr/io/p/io_ar.c
@@ -20,9 +20,9 @@ static RIODesc *r_io_ar_open(RIO *io, const char *file, int rw, int mode) {
 		filename += 2;
 	}
 
-	RBuffer *b = ar_open_file (arname, filename);
-	if (b) {
-		res = r_io_desc_new (io, &r_io_plugin_ar, filename, rw, mode, b);
+	RArFp *arf = ar_open_file (arname, filename);
+	if (arf) {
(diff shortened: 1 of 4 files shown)


trufae (approved):

trufae:
LGTM, but i'll wait for you to add a testas you mentioned.

About the support for openmany. i think this api should be rethought because it will be useful to know the source file from where the opened files are coming from.I am deleting RCoreFile, so take this into account. see https://github.com/radareorg/radare2/pull/18181

swoops:
Added two tests for a new ar file found in this PR: https://github.com/radareorg/radare2-testbins/pull/57

PR #18184, opened 2021-01-08
Fix r2 - tab crash

eagleoflqj:
**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/test/unit/test_regex.c
+++ b/test/unit/test_regex.c
@@ -9,7 +9,7 @@ static int test_regex(void) {
 		printf ("error\n");
 	} else {
 		const size_t nmatch = 32;
-		RRegexMatch pmatch[nmatch];
+		RRegexMatch pmatch[nmatch] = {0};
 		rc = r_regex_exec (&rx, "patata", nmatch, pmatch, 0);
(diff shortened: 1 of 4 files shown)


test/unit/test_regex.c
```diff
@@ -10,6 +11,7 @@ static int test_regex(void) {
 	} else {
 		const size_t nmatch = 32;
 		RRegexMatch pmatch[nmatch];
```
trufae, line 13: you can do `= {0};` instead of the memset below

test/unit/test_regex.c
```diff
@@ -10,6 +11,7 @@ static int test_regex(void) {
 	} else {
 		const size_t nmatch = 32;
```
trufae, line 12: maybe msvc dislikes array with const vars, if so we can just use a define

test/unit/test_regex.c
```diff
@@ -1,4 +1,5 @@
 #include <stdio.h>
+#include <string.h>
```
trufae, line 2: it is really needed? if you use ={0} you can remove this line too.

trufae (changes requested):

trufae:
Fixed in local and pushed :) thanks!!

PR #18186, opened 2021-01-08
Fix underflow in r_list_pop

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

If you pop from an empty list the `RList->length` value will go negative. I think this is an unexpected state that could cause unexpected behavior.

I don't know if there is any code in r2 that uses this in an unsafe way. I only found this because I did a `while (r_list_pop (list))` and ended up with a list of length -1.

### DIFF
--- a/libr/util/list.c
+++ b/libr/util/list.c
@@ -282,8 +282,8 @@ R_API void *r_list_pop(RList *list) {
 		}
 		data = iter->data;
 		free (iter);
+		list->length--;
 	}
-	list->length--;
 	return data;
 }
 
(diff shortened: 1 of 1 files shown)


swoops:
`CS-` and `fs-` over and over again will cause a list of negative length. This wont cause any problems though because the length is never really used. The r_list_* functions tend to favor checking `head` and `tail` for logic, instead of length.

I only see a hand full of places where r_list_length is used in doing math on mem allocation. I don't see a way to reach r_list_pop* in those cases in order to cause a list to have an invalid length. 

So no crash case or easy meme leak that I can show.

PR #18209, opened 2021-01-12
Fix some stuff in the CI

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**

* [ ] Add RPM builds (for redhat/fedora/...)
* [ ] Alpine-i386 builds (for iSH.app)
* [ ] Add Termux builds
* [ ] Make macOS builds fat

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,12 @@
 name: CI
 
-on: push
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+#  schedule:
+#    - cron: '0 0 * * 1'
 
 # python sys\meson.py --release --backend vs2019 --shared --install --prefix="%cd%\radar    e2_dist" --webui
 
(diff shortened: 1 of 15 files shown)


libr/anal/esil.c
```diff
@@ -10,6 +10,8 @@
 #include <float.h>
 #include <fenv.h>
 
+#define R_NUM_NAN -1 /* XXX */
```
pelijah, line 13: There is `NAN` macro in `math.h` for MSVC

pelijah:
```c
int main() {
	printf("isnan(NAN)=%d\n", isnan(NAN));
	printf("isnan(-1)=%d\n", isnan((double)-1));
	return 0;
}
```
gives
```
isnan(NAN)=1
isnan(-1)=0
```

trufae: I know. Its a bad patch. Just to check if the windows build was failing for something else. Theres a pr fixing that nan thing. But we can try with this NAN on windows too

PR #18222, opened 2021-01-13
Fix UnnamedClass issue on fatmacho.objc parsing on arm64 ##bin

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Reproducer:

* This issue happens ONLY on arm64 hardware (tested on iOS-arm64, linux-arm64 and macos-arm64)
* This bug happens only when listing objc classes on FAT MACHO binaries

```
$ r2 fatrepro
> ic
UnnamedClass ...
```

This PR is not yet ready to merge. i need to add a reproducer and think a better fix.


trufae:
This fix looks better https://github.com/radareorg/radare2/pull/18248

PR #18226, opened 2021-01-13
WIP: r2up - update from git with r2 -U

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**

- [ ] sys/install.sh leaves r2 unaccessible until the build is done, will be good to stage current installation until the new build finishes to avoid losing r2 if the build fails
- [ ] this is similar to r2pm -ci r2 but written in C
- [ ] This may be useful to unify all the update and installation methods for all OSs

Please discuss!

### DIFF
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -323,6 +323,43 @@ static void set_color_default(RCore *r) {
 	free (tmp);
 }
 
+static int r2up(const char *argv0) {
+#if __UNIX__
+	int rc = 1;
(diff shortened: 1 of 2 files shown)


trufae:
This thing reminds me that it would be handy to have r2 as a build systrerm frontend to not depend on make to build the plugins, this way we can make many plugins easily compilable without having to care about pkg-config, installation path, operating system ...

trufae:
It's better to delegate this task to r2env

Issue #18150, opened 2021-01-01
Fix all memleaks in test and enable memleak detection in ASAN


eagleoflqj:
On Ubuntu 20.10 x64
```sh
sys/sanitize.sh
export ASAN_OPTIONS=detect_odr_violation=0
export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libasan.so.6
```
Then run tests. Memory leak everywhere.

trufae:
uhm not sure if we really need to have an issue to track this

eagleoflqj:
> uhm not sure if we really need to have an issue to track this

Maybe newcomers see and know where to start? I made my first pr after noticing how many LGTM alerts there were.

trufae:
that could be

eagleoflqj:
Feel free to close. I only remember tpi stuff result in terrible memleak.

PR #18238, opened 2021-01-16
Fix memleak in db/esil/apple

eagleoflqj:
**Checklist**

- [ ] Closing issues: #issue

**Description**

#18150 
As I said in discord, there's a problem of `r_skiplist_insert`. When it meets duplicate key, it simply returns the original node. but this result is always ignored, thus a call `r_skiplist_insert (list, data)` may cause memory leak.
Should we check the result every time, or use a new function to insert and free if duplicate?
the leak can be reproduced by `r2 -Qc 'pd 2~?push' bins/mach0/arm16-ofp.lzh`

### DIFF
--- a/libr/util/skiplist.c
+++ b/libr/util/skiplist.c
@@ -198,6 +198,15 @@ R_API RSkipListNode* r_skiplist_insert(RSkipList* list, void* data) {
 	return x;
 }
 
+R_API bool r_skiplist_insert_autofree(RSkipList* list, void* data) {
+	RSkipListNode* node = r_skiplist_insert (list, data);
+	if (node && data != node->data) { // duplicate
+		free (data);
+		return false;
+	}
+	return true;
+}
+
 // Delete node with data as it's payload.
 R_API bool r_skiplist_delete(RSkipList* list, void* data) {
 	return delete_element (list, data, true);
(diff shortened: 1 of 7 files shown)


trufae:
Do you think we can merge this PR as is? or you want to address this skiplist issue you mention in te same pr?

eagleoflqj:
> Do you think we can merge this PR as is? or you want to address this skiplist issue you mention in te same pr?

You mean add the api to automatically free? If you think it’s the right way, then i’ll do it in this pr.

libr/util/skiplist.c
```diff
@@ -198,6 +198,15 @@ R_API RSkipListNode* r_skiplist_insert(RSkipList* list, void* data) {
 	return x;
 }
 
+R_API bool r_skiplist_insert_autofree(RSkipList* list, void* data) {
+	RSkipListNode* node = r_skiplist_insert (list, data);
+	if (node && data != node->data) { // duplicate
+		free (data);
```
trufae, line 204: uhm maybe this should be calling the list.free function pointer instead of just free

trufae (changes requested):

libr/util/skiplist.c
eagleoflqj: right

PR #18245, opened 2021-01-19
Minor bugfixes to radiff2

aamcintosh:
**Checklist**

- [X ] Closing issues: 18235, 18236 and 18244.
- [ X] Mark this if you consider it ready to merge

**Description**

**Issue 18236**
This fix makes changes to `libr/main/radiff2.c`  around line 1118.  I included a description of how to build a test case when I submitted this issue.
**Issue 18235**
This fix changes line 189 in  `libr/anal/diff.c.`  This is "it can't happen here" code, so it's not clear how to build a test case.  The original code is incorrect on inspection.  If either `fcn->name`  or `fcn2->name` is `NULL`, control falls into the code that will compare one function with a `NULL` name to one function with a name that may not be `NULL`.
**Issue 18244**
Fixes some minor documentation inconsistencies that I spotted.  I believe I understand the current semantics for -O, but someone else please check!

### DIFF
--- a/libr/anal/diff.c
+++ b/libr/anal/diff.c
@@ -186,7 +186,7 @@ R_API int r_anal_diff_fcn(RAnal *anal, RList *fcns, RList *fcns2) {
 	if (fcns) {
 		r_list_foreach (fcns, iter, fcn) {
 			r_list_foreach (fcns2, iter2, fcn2) {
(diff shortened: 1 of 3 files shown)


trufae:
Thanks! The changes look good to me. Just one extra hint: if you use Fix #xxx - in thr commit message github will automatically close and reference the issue.

PR #18248, opened 2021-01-19
Fix Mach-O rebase on fat slices ##bin

mrmacete:
**Checklist**

- [ ] Closing issues: #issue

**Description**

The gist of it is: compensate for the bin file offset when running the rebasing logic.

This aims to fix the same problem as https://github.com/radareorg/radare2/pull/18222 in a less hacky way.

I still think that fat slices should be mapped alone and not relying on bin file offset, but this works in the short term.


mrmacete:
ok, looks like a broke a couple of tests, will look into that later today

trufae:
I like your patch more than mine, but it seems to break a couple of tests. Not sure if those are real regressions or fixes :D can you check them out?

mrmacete:
> I like your patch more than mine, but it seems to break a couple of tests. Not sure if those are real regressions or fixes :D can you check them out?

yep! will do later today

mrmacete:
Added a test, but needs this new bin in testbins: https://github.com/radareorg/radare2-testbins/pull/59

PR #18270, opened 2021-01-23
Use pj in debug ##refactor

eagleoflqj:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Not perfect. There is still `r_cons_strcat ("[");` and manually printed comma, because the callback is not easy to deal with.

### DIFF
--- a/libr/debug/plugin.c
+++ b/libr/debug/plugin.c
@@ -50,15 +50,20 @@ R_API bool r_debug_use(RDebug *dbg, const char *str) {
 	return (dbg && dbg->h);
 }
 
-R_API int r_debug_plugin_list(RDebug *dbg, int mode) {
+R_API bool r_debug_plugin_list(RDebug *dbg, int mode) {
 	char spaces[16];
 	int count = 0;
 	memset (spaces, ' ', 15);
 	spaces[15] = 0;
 	RDebugPlugin *h;
 	RListIter *iter;
+	PJ *pj = NULL;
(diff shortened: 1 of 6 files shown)


trufae:
Do you want to fix this strcat thing in this PR? im fine to merge it in this state too. otherwise ill just change the title to WIP:

eagleoflqj:
> Do you want to fix this strcat thing in this PR? im fine to merge it in this state too. otherwise ill just change the title to WIP:

The only way I can think of is to define a struct that contains a `RDebug *` and a `PJ *`, then pass it to `sdb_foreach`. If you are ok with it, I'll do it in this pr.

libr/debug/plugin.c
```diff
@@ -50,27 +50,31 @@ R_API bool r_debug_use(RDebug *dbg, const char *str) {
...
 	char spaces[16];
 	int count = 0;
 	memset (spaces, ' ', 15);
 	spaces[15] = 0;
 	RDebugPlugin *h;
 	RListIter *iter;
+	PJ *pj = NULL;
 	if (mode == 'j') {
-		dbg->cb_printf ("[");
+		pj = pj_new ();
```
trufae, line 62: Better use core_pj_new() i thunk having a PJ instance in rdebug is the simplest solution here. So it can be initialized by core for user sefined options

trufae (changes requested):

PR #18272, opened 2021-01-23
Fix mg when file exists and cannot be created

as0ler:
- [ ] Closing issues: #issue

**Description**

When executing 'mg', the message is printed when file does not previously exist.  With this fix, error message is only printed when the file exists and cannot be truncated.

### DIFF
--- a/libr/core/cmd_mount.c
+++ b/libr/core/cmd_mount.c
@@ -331,8 +331,9 @@ static int cmd_mount(void *data, const char *_input) {
 			int total_bytes_read = 0;
 			int blocksize = file->size < core->blocksize ? file->size : core->blocksize;
 			size = size > 0 ? size : file->size;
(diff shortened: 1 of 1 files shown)


libr/core/cmd_mount.c
```diff
@@ -331,8 +331,9 @@ static int cmd_mount(void *data, const char *_input) {
 			int total_bytes_read = 0;
 			int blocksize = file->size < core->blocksize ? file->size : core->blocksize;
 			size = size > 0 ? size : file->size;
-			if (!r_sys_truncate (localFile, 0)) {
+			if (r_file_exists(localFile) && !r_sys_truncate (localFile, 0)) {
```
trufae, line 334:
```suggestion
			if (r_file_exists (localFile) && !r_sys_truncate (localFile, 0)) {
```

PR #18294, opened 2021-01-26
Avoid overwriting LDFLAGS

krytarowski:
Append, instead of setting -lkvm for BSD platforms.

**Checklist**

- [ ] Closing issues: #issue

**Description**

Overwriting LDFLAGS causes spurious fallout in the build system, as detected on NetBSD.


krytarowski:
Without this change, .so symlinks are broken.

PR #18299, opened 2021-01-26
Update debian.sh

gordboy:
The flag as given is a typo by the looks of it, as there is no such flag for GCC. As given, the build by doing

`./sys/debian.sh`

to build debian packages is broken.

This small edit corrects things.

Of course the whole thing is just a ruse to hide warnings from the compiler, it does not actually fix any underlying badness.


trufae:
Thanks!

PR #18317, opened 2021-01-31
Fix the broken debugger tests after last gha update

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**


trufae:
Uhm debug tests behave different in the tcc build because the amount of libs loaded in memory is different, im tempted to disable those tests and future rewrite them properly, but at least we get a green ci

PR #18338, opened 2021-02-08
from rizin: Fix dead nested assignment libr/bin/format/mach0/mach0.c

mrmacete:
import commit from rizin "Fix dead nested assignment libr/bin/format/mach0/mach0.c"


trufae:
The way r_buf_fread_at() works right now is by returning:

* the amount of bytes read
* -1 if can't read enough bytes

This seems to me like a bool, i do think that writing a formatted buffer doesnt make sense to read further or try back later if it has been truncated by the buffer boundaries. So to be it will be good to change the signature of this function and check for bool in all the uses around.

This will help to identify some bugs in the way its used, what do you think?

mrmacete:
agree, but yeah that would be another separated huge PR 😅

trufae:
Agree!

PR #18351, opened 2021-02-10
write multi-byte input represented by less bytes as output

gogo2464:
**Checklist**

- [ ] Closing issues: #issue

**Description**

tha commands: `e cfg.charset=pokered; w abcd` used to work with only one char for one char translation. It can now select less values in the output.

Tested with
`e cfg.charset=pokered; "w <RIVAL>abcdefg<RIVAL>"`

### DIFF
--- a/libr/include/r_util/r_str.h
+++ b/libr/include/r_util/r_str.h
@@ -33,6 +33,8 @@ typedef struct r_charset_t {
 	Sdb *db_char_to_hex;
 	RCharsetRune *custom_charset;
 	size_t remaining;
+    size_t encode_maxkeylen;
+    size_t decode_maxkeylen;
 } RCharset;
 
 #define R_STR_ISEMPTY(x) (!(x) || !*(x))
--- a/libr/util/charset.c
+++ b/libr/util/charset.c
@@ -113,24 +113,35 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const 
 
 // assumes out is as big as in_len
 R_API size_t r_charset_decode_str(RCharset *rc, ut8 *out, size_t out_len, const ut8 *in, size_t in_len) {
-	char k[32];
 	char *o = (char*)out;
-	int i;
-	for (i = 0; i < in_len; i++) {
-		ut8 ch_in = in[i];
+	size_t last_char_size;
+	size_t maxkeylen = 8;
 
-		//zero terminate the string
-		snprintf (k, sizeof (k), "%c", ch_in);//snprintf (k, sizeof (k), "0x%02x", ch_in);
-		char *v = sdb_get (rc->db_char_to_hex, k, 0);
-		memmove(v, v+2, strlen (v));
+	size_t cur, j;
+	for (cur = 0; cur < in_len; ) {
+		char *str = r_str_ndup((char *)in, maxkeylen);
+		for (j = in_len ; j > 0; j--) {
+			//zero terminate the string
+			str[j] = '\0';
 
-		//convert to ascii
-		char str_hx[32];
-		snprintf (str_hx, sizeof (str_hx), "%c", (char) strtol( v, 0, 16));
-		const char *ret = r_str_get_fail (str_hx, "?");
-		strcpy (o, ret);
+			const char *v = sdb_const_get (rc->db_char_to_hex, (char *) str+cur, 0);
+			if (v) {
+				//convert to ascii
+				char *str_hx = malloc(maxkeylen);
+				snprintf (str_hx, maxkeylen, "%c", (char) strtol(v, 0, 16));//in the future handle multiple chars output
+				const char *ret = r_str_get_fail (str_hx, "?");
 
-		o += strlen (o);
+				//concatenate
+				strcpy (o, ret);
+				o += strlen (o);
+
+				//pass for multiple chars
+				last_char_size = strlen ( (char *)str+cur);
+				cur += last_char_size;
+				free (str_hx);
+			}
+		}
+		free (str);
 	}
 	return o - (char*)out;
 }


libr/util/charset.c
```diff
@@ -113,24 +113,36 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
-
-		//convert to ascii
-		char str_hx[32];
-		snprintf (str_hx, sizeof (str_hx), "%c", (char) strtol( v, 0, 16));
-		const char *ret = r_str_get_fail (str_hx, "?");
-		strcpy (o, ret);
-
-		o += strlen (o);
+	int j;
+	ut8 str[320];
```
trufae, line 118: remove this local variable its not necessary

libr/util/charset.c
```diff
@@ -113,24 +113,36 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+	size_t last_char_size;
+
+	size_t cur;
+	for (cur = 0; cur < in_len; ) {
+		memcpy (str, in, sizeof (str));
+		for (j =( in_len+1 ); j > 0; j--) {
```
trufae, line 124: remove parenthesis and add spaces around +

libr/util/charset.c
```diff
@@ -113,24 +113,36 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+			str[j] = '\0';
+			//zero terminate the string
+			char *v = sdb_get (rc->db_char_to_hex, (char *) str+cur, 0);
+
+			if (v) {
+				memmove (v, v+2, strlen (v));
+
+				//convert to ascii
+				char str_hx[320];
```
trufae, line 133: you dont need 320 chars to save two chars

libr/util/charset.c
```diff
@@ -113,24 +113,36 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+				snprintf (str_hx, sizeof (str_hx), "%c", (char) strtol(v, 0, 16));//in the future handle multiple chars output
+				const char *ret = r_str_get_fail (str_hx, "?");
+
+				//concatenate
+				strcpy (o, ret);
```
trufae, line 138: see below

libr/util/charset.c
trufae: use 8 instead of 320

libr/util/charset.c
```diff
@@ -113,24 +113,36 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+				o += strlen (o);
```
trufae, line 139: o = strcpy (o, ret);

libr/util/charset.c
```diff
@@ -113,24 +113,36 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+	size_t cur;
+	for (cur = 0; cur < in_len; ) {
+		memcpy (str, in, sizeof (str));
+		for (j =( in_len+1 ); j > 0; j--) {
+			str[j] = '\0';
+			//zero terminate the string
+			char *v = sdb_get (rc->db_char_to_hex, (char *) str+cur, 0);
+
+			if (v) {
+				memmove (v, v+2, strlen (v));
```
trufae, line 130: remove this line

libr/util/charset.c
```diff
@@ -113,24 +113,36 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+	ut8 str[320];
+	size_t last_char_size;
+
+	size_t cur;
+	for (cur = 0; cur < in_len; ) {
+		memcpy (str, in, sizeof (str));
+		for (j =( in_len+1 ); j > 0; j--) {
+			str[j] = '\0';
+			//zero terminate the string
+			char *v = sdb_get (rc->db_char_to_hex, (char *) str+cur, 0);
```
trufae, line 127: use sdb_const_get

libr/util/charset.c
```diff
@@ -113,24 +113,36 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+
+			if (v) {
+				memmove (v, v+2, strlen (v));
+
+				//convert to ascii
+				char str_hx[320];
+				snprintf (str_hx, sizeof (str_hx), "%c", (char) strtol(v, 0, 16));//in the future handle multiple chars output
```
trufae, line 134: use v+2, so you dont need to memmove

libr/util/charset.c
```diff
@@ -113,24 +113,36 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+				const char *ret = r_str_get_fail (str_hx, "?");
+
+				//concatenate
+				strcpy (o, ret);
+				o += strlen (o);
+
+				//pass for multiple chars
+				last_char_size = strlen ( (char *)str+cur);
```
trufae, line 142: strlen(o) is the same as this str+cur. but for now i would just cur++

libr/util/charset.c
```diff
@@ -113,24 +113,36 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
-		strcpy (o, ret);
-
-		o += strlen (o);
+	int j;
+	ut8 str[320];
+	size_t last_char_size;
+
+	size_t cur;
+	for (cur = 0; cur < in_len; ) {
+		memcpy (str, in, sizeof (str));
```
trufae, line 123: remove this line is wrong.causes overflow and its unnecssary

libr/util/charset.c
trufae, line 124: why in_len+1? why overengineer this loop that can be done forward? for(j=0;j<in_len;j++)

libr/util/charset.c
```diff
@@ -113,24 +113,36 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+		for (j =( in_len+1 ); j > 0; j--) {
+			str[j] = '\0';
```
trufae, line 125: do this after the for loop.

libr/util/charset.c
trufae: and you can actually do : you can actually do *o = strtol(v+2,0,16); which is much simpler

libr/util/charset.c
trufae: its always 1 for now and strlen shouldnt be needed when implemented multichar support

trufae (changes requested):

libr/util/charset.c
trufae: or just do my *o = strtol.. suggestion and this is not necessary. multichar support is not yet there, so lets make the code simpler

libr/util/charset.c
trufae: wait i think i got your idea to find the nth chars in the database, this can be very expensive as it depends on the size of str. and 320 is still too much, i would say 8 or 16 for now.

libr/util/charset.c
trufae: how do u plan to handle multichar? i think we can check if the value starts with 0x otherwise consider it a plain ascii string instead of binary data

libr/util/charset.c
trufae: size_t retlen = strlen (ret);

and check if the string that is about to be copied fits in the output buffer

libr/util/charset.c
gogo2464: because j should be decreasing.

libr/util/charset.c
gogo2464: `strcpy (o, ret, strlen (ret));` ?

libr/util/charset.c
gogo2464: `o = strcpy (o, ret);` writes only the first char. For example in `w <RIVAL>abc`, only <RIVAL> is printed. It is a great bug.

libr/util/charset.c
trufae: yeah i got your idea after sending the review... but instead of using in_len, maybe it will be to compute the maxkeylen and maxkeylen

gogo2464: in_len + 1 looks useless and weird. I think `for (j = in_len ; j > 0; j--) {` fits perfectly.

gogo2464: No the +1 is necessary.

libr/util/charset.c
gogo2464: It actually works with 0x before too.

libr/util/charset.c
gogo2464: curr++ will only work for one char characters size.

libr/util/charset.c
gogo2464: `for (cur = 0; cur < in_len; ) {` sounds better. What do you think?

libr/util/charset.c
gogo2464: I do not think this is unnecessary. we just need to find a way to come back to the previous state and to not allocate too much memory.

gogo2464:
I have solved all what I can solve. The errors in CI are not related to my code[See](https://github.com/radareorg/radare2/pull/18354).

libr/util/charset.c
trufae: in this case then use r_str_ncpy instead of strcpy, but the 3rd argument must be R_MIN(strlen(ret),remaining_bytes); then

libr/util/charset.c
trufae: the previous state is not necessary, you just trim one char from the end of the string on each iteration. so better use
```
 char *str = r_str_ndup(in, maxkeylen);
for (j ...
free(str);
```

gogo2464: I am going to try.

libr/util/charset.c
```diff
@@ -113,24 +113,33 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
-
-		//convert to ascii
-		char str_hx[32];
-		snprintf (str_hx, sizeof (str_hx), "%c", (char) strtol( v, 0, 16));
-		const char *ret = r_str_get_fail (str_hx, "?");
-		strcpy (o, ret);
-
-		o += strlen (o);
+	size_t last_char_size;
+	size_t maxkeylen = 8;
```
trufae, line 118: this should be rc->maxkeylen; do you want to make this change in this commit?

libr/util/charset.c
```diff
@@ -113,24 +113,33 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+
+	size_t cur, j;
+	for (cur = 0; cur < in_len; ) {
+		char *str = r_str_ndup((char *)in, maxkeylen);
```
trufae, line 122: shouldnt it be in + cur, R_MIN (maxkeylen, in_len-cur)) ?

libr/util/charset.c
```diff
@@ -113,24 +113,33 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+
+				//concatenate
+				strcpy (o, ret);
+				o += strlen (o);
+
+				//pass for multiple chars
+				last_char_size = strlen ( (char *)str+cur);
+				cur += last_char_size;
+			}
+		}
```
trufae, line 142: missing free(str);

libr/util/charset.c
```diff
@@ -113,24 +113,33 @@ R_API size_t r_charset_encode_str(RCharset *rc, ut8 *out, size_t out_len, const
...
+			const char *v = sdb_const_get (rc->db_char_to_hex, (char *) str+cur, 0);
+			if (v) {
+				//convert to ascii
+				char *str_hx = malloc(maxkeylen);
+				snprintf (str_hx, sizeof (str_hx), "%c", (char) strtol(v, 0, 16));//in the future handle multiple chars output
+				const char *ret = r_str_get_fail (str_hx, "?");
+
+				//concatenate
+				strcpy (o, ret);
+				o += strlen (o);
```
trufae, line 136: missing free(str_hx);

libr/util/charset.c
gogo2464: It will impact r_charset_new so I will make another PR.

gogo2464: Because I realize I have tons of change to make on this function.

libr/util/charset.c
gogo2464: It works like that...

trufae: Yes. It works because in is luckily null terminated. But it is not correct

libr/util/charset.c
trufae: Passing strlen to strcpy should be forbidden at compile time. Is one of the most dangerous and obvious bugs

libr/util/charset.c
trufae: Sgtm, but computing the maxkeylen by iterating all the keys should be no more than 5 lines and an extra variable. But im fine if you want to make that into a separate pr. Lets finish the comments in here fitst

trufae:
Ci is red

libr/include/r_util/r_str.h
```diff
@@ -33,6 +33,8 @@ typedef struct r_charset_t {
 	Sdb *db_char_to_hex;
 	RCharsetRune *custom_charset;
 	size_t remaining;
+    size_t encode_maxkeylen;
```
trufae, line 36: Use tabs

PR #18359, opened 2021-02-13
Fix #17049 - oa whithout filename specify, add oa test ##shell ##test

### LINKED ISSUE #17049: Fix oa when not specifying the filename and add tests for it
oa expects the filename to be specified as the third argument. When not passing it, the following happens:

### Steps to reproduce + actual behavior
```
[florian@florian-pinebook test]$ r2 bins/mach0/ired
 -- What has been executed cannot be unexecuted
[0x100007e4c]> oa arm 32
WARNING: r_bin_file_find_by_name: assertion 'bin && name' failed (line 622)
Cannot find file (null)
[0x00008da8]> 

```

### Expected behavior
It should either print that it is missing an argument or (preferred) check if there is only a single open file and just use that.

RNRNRNR:
**Checklist**

- [ ] Closing issues: #issue

**Description**
Fix bug with oa command, add a case usage without filename if opened only one file. Add tests for oa.
Probably need to change command description in the book because the new use case not obvious, but not sure how

### DIFF
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -1416,9 +1416,18 @@ static int cmd_open(void *data, const char *input) {
 				bits = r_num_math (core->num, r_str_word_get0 (ptr, 1));
 				arch = r_str_word_get0 (ptr, 0);
 				r_core_bin_set_arch_bits (core, filename, arch, bits);
-				RBinFile *file = r_bin_file_find_by_name (core->bin, filename);
-				if (!file) {
-					eprintf ("Cannot find file %s\n", filename);
+				RBinFile *file;
+				if (filename != NULL) {
+					file = r_bin_file_find_by_name (core->bin, filename);
+					if (!file) {
+						eprintf ("Cannot find file %s\n", filename);
+						free (ptr);
+						return 0;
+					}
+				} else if (core->bin->binfiles->length == 1) {
+					file = (RBinFile *)core->bin->binfiles->head->data;
+				} else {
+					eprintf ("More than one file is opened, specify the filename\n");
 					free (ptr);
 					return 0;
 				}
--- a/libr/core/cmd_open.c
+++ b/libr/core/cmd_open.c
@@ -1416,18 +1416,18 @@ static int cmd_open(void *data, const char *input) {
 				bits = r_num_math (core->num, r_str_word_get0 (ptr, 1));
 				arch = r_str_word_get0 (ptr, 0);
 				r_core_bin_set_arch_bits (core, filename, arch, bits);
-				RBinFile *file;
-				if (filename != NULL) {
+				RBinFile *file = NULL;
+				if (filename) {
 					file = r_bin_file_find_by_name (core->bin, filename);
 					if (!file) {
 						eprintf ("Cannot find file %s\n", filename);
-						free (ptr);
-						return 0;
 					}
-				} else if (core->bin->binfiles->length == 1) {
-					file = (RBinFile *)core->bin->binfiles->head->data;
+				} else if (r_list_length(core->bin->binfiles) == 1) {
+					file = (RBinFile *)r_list_first(core->bin->binfiles);
 				} else {
 					eprintf ("More than one file is opened, specify the filename\n");
+				}
+				if (!file) {
 					free (ptr);
 					return 0;
 				}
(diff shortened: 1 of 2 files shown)


libr/core/cmd_open.c
```diff
@@ -1416,9 +1416,18 @@ static int cmd_open(void *data, const char *input) {
...
-					eprintf ("Cannot find file %s\n", filename);
+				RBinFile *file;
+				if (filename != NULL) {
+					file = r_bin_file_find_by_name (core->bin, filename);
+					if (!file) {
+						eprintf ("Cannot find file %s\n", filename);
+						free (ptr);
+						return 0;
+					}
+				} else if (core->bin->binfiles->length == 1) {
```
trufae, line 1427: use r_list_length () instead of accessible the internal field

libr/core/cmd_open.c
```diff
@@ -1416,9 +1416,18 @@ static int cmd_open(void *data, const char *input) {
...
+					file = (RBinFile *)core->bin->binfiles->head->data;
```
trufae, line 1428: use r_list_head

libr/core/cmd_open.c
```diff
@@ -1416,9 +1416,18 @@ static int cmd_open(void *data, const char *input) {
 				bits = r_num_math (core->num, r_str_word_get0 (ptr, 1));
 				arch = r_str_word_get0 (ptr, 0);
 				r_core_bin_set_arch_bits (core, filename, arch, bits);
-				RBinFile *file = r_bin_file_find_by_name (core->bin, filename);
-				if (!file) {
-					eprintf ("Cannot find file %s\n", filename);
+				RBinFile *file;
+				if (filename != NULL) {
```
trufae, line 1420:
```suggestion
				if (filename) {
```

libr/core/cmd_open.c
```diff
@@ -1416,9 +1416,18 @@ static int cmd_open(void *data, const char *input) {
...
+					file = r_bin_file_find_by_name (core->bin, filename);
+					if (!file) {
+						eprintf ("Cannot find file %s\n", filename);
+						free (ptr);
+						return 0;
+					}
+				} else if (core->bin->binfiles->length == 1) {
+					file = (RBinFile *)core->bin->binfiles->head->data;
+				} else {
+					eprintf ("More than one file is opened, specify the filename\n");
```
trufae, line 1430: check if (!file) and return once for the 3 failing cases above

libr/core/cmd_open.c
```diff
@@ -1416,9 +1416,18 @@ static int cmd_open(void *data, const char *input) {
 				bits = r_num_math (core->num, r_str_word_get0 (ptr, 1));
 				arch = r_str_word_get0 (ptr, 0);
 				r_core_bin_set_arch_bits (core, filename, arch, bits);
-				RBinFile *file = r_bin_file_find_by_name (core->bin, filename);
-				if (!file) {
-					eprintf ("Cannot find file %s\n", filename);
+				RBinFile *file;
```
trufae, line 1419: initialize to NULL; so you can move the if (!file) check once after trying to open the file

trufae (changes requested):

libr/core/cmd_open.c
```diff
@@ -1416,9 +1416,18 @@ static int cmd_open(void *data, const char *input) {
...
 				arch = r_str_word_get0 (ptr, 0);
 				r_core_bin_set_arch_bits (core, filename, arch, bits);
-				RBinFile *file = r_bin_file_find_by_name (core->bin, filename);
+				RBinFile *file = NULL;
+				if (filename) {
+					file = r_bin_file_find_by_name (core->bin, filename);
+					if (!file) {
+						eprintf ("Cannot find file %s\n", filename);
+					}
+				} else if (r_list_length(core->bin->binfiles) == 1) {
```
trufae, line 1425:
```suggestion
				} else if (r_list_length (core->bin->binfiles) == 1) {
```

libr/core/cmd_open.c
```diff
@@ -1416,9 +1416,18 @@ static int cmd_open(void *data, const char *input) {
...
+					file = (RBinFile *)r_list_first(core->bin->binfiles);
```
trufae, line 1426:
```suggestion
					file = (RBinFile *)r_list_first (core->bin->binfiles);
```

trufae (approved):

trufae:
Thanks! Merged

PR #18364, opened 2021-02-14
Fix bug to write the last chars

gogo2464:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Fix the bug that attempt to write the last char.

Does not fix the bug that write 2 times multiple charsets.


gogo2464:
The test has faield for capstone. Not me.

I still do not think the charset test will work because there is still the unfixed bug remaining.

gogo2464:
I know how to solve the remaining bug.

gogo2464:
I have not finished. Let me fix the second bug I have almost finished it. :)

PR #18367, opened 2021-02-14
Fix bug to not print both long chars

gogo2464:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Fix the bug that write 2 times multiple charsets.

### DIFF
--- a/libr/util/charset.c
+++ b/libr/util/charset.c
@@ -149,7 +149,8 @@ R_API size_t r_charset_decode_str(RCharset *rc, ut8 *out, size_t out_len, const 
 					o += ll - 1;
 				}
 				found = true;
-				cur += strlen (str) - 1;
+				size_t str_len = strlen (str) - 1;
+				cur += str_len > 0 ? str_len - 1: 1;
 				free (str_hx);
 				break;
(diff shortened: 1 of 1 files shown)


libr/util/charset.c
```diff
@@ -149,6 +149,7 @@ R_API size_t r_charset_decode_str(RCharset *rc, ut8 *out, size_t out_len, const
 					o += ll - 1;
 				}
 				found = true;
+				cur += strlen (str) - 1;
```
trufae, line 152: Strlen(str) is the same as j

gogo2464: No sometimes they do not have the same value.

trufae, line 152:
```suggestion
				size_t str_len = strlen (str) - 1;
				cur += str_len > 0 ? str_len - 1: 1;
```

trufae (approved):

trufae:
I just fixed the possible underflow and ill merge, but this strlen is unnecessary to me, we can optimize it later. lets just move forward. sorry for the slow reposnse, busy in the real world lately

PR #18369, opened 2021-02-16
Fix segfault in ms with blank line

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This aims to fix a segfault in `ms` shell. Just hitting enter in the shell will cause a segfault. This is because `shell->readline ()` returns a non-null pointer, pointing at an empty string in non-writable memory. Then `r_str_trim ((char *)ptr)` would cause  a `memmove(ptr, ptr, 1)` resulting in a segfault. This update makes it so the returned `ptr` buffer is never written to. It also simplifies the code a bit (imho) and removes a memory leak.


trufae:
Neat catch!

PR #18387, opened 2021-02-26
Do not write nops in the cache ##io

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -2087,6 +2087,12 @@ static bool cb_io_cache_mode(void *user, void *data) {
 	return true;
 }
 
+static bool cb_io_cache_nodup(void *user, void *data) {
+	RCore *core = (RCore *)user;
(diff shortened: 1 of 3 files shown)


trufae:
@condret what do you think about this? the idea is that some writes shouldnt probably happen if the contents in memory are the same, maybe this behaviour should be enabled/disabled during loading time, because under esil emulation we need to track all the writes even if they are the same. but for patching relocs we can simplify that quite a lot

condret:
hm, for user this makes perfectly sense, and sometimes even for esil, if you want to extract a diff

trufae:
Yeah but maybe we need a toggle to pick the behaviour instead of forcing one mode or the other

condret:
but the only usecase for tracking all writes is memory mapped hardware

condret:
I think we can pretty much force mode for userland and support different modes in APIland

PR #18392, opened 2021-02-28
Add Support for new CoreSymbolication element format ##bin

mrmacete:
**Checklist**

- [ ] Closing issues: #issue

**Description**

The new format is similar to the older one, but all string references are offsets from the start of the string table instead of relative to where the reference lives.

This change detects and reacts to this condition.

Depends on new test bin:  https://github.com/radareorg/radare2-testbins/pull/60

### DIFF
--- a/libr/bin/format/mach0/coresymbolication.c
+++ b/libr/bin/format/mach0/coresymbolication.c
@@ -247,6 +247,7 @@ RCoreSymCacheElement *r_coresym_cache_element_new(RBinFile *bf, RBuffer *buf, ut
 		}
 	}
 	bool relative_to_strings = false;
+	ut8* string_origin;
 	if (hdr->n_sections > 0) {
 		result->sections = R_NEWS0 (RCoreSymCacheElementSection, hdr->n_sections);
 		if (!result->sections) {
@@ -278,11 +279,8 @@ RCoreSymCacheElement *r_coresym_cache_element_new(RBinFile *bf, RBuffer *buf, ut
 			if (bits == 32) {
 				cursor += word_size;
 			}
-			if (relative_to_strings) {
-				sect->name = str_dup_safe (b, b + start_of_strings + (size_t)sect_name_off, end);
-			} else {
-				sect->name = str_dup_safe (b, sect_start + (size_t)sect_name_off, end);
-			}
+			string_origin = relative_to_strings? b + start_of_strings : sect_start;
+			sect->name = str_dup_safe (b, string_origin + (size_t)sect_name_off, end);
 		}
 	}
 	if (hdr->n_symbols) {
@@ -300,20 +298,14 @@ RCoreSymCacheElement *r_coresym_cache_element_new(RBinFile *bf, RBuffer *buf, ut
 			size_t name_off = r_read_le32 (cursor + 0xc);
 			size_t mangled_name_off = r_read_le32 (cursor + 0x10);
 			sym->unk2 = (st32)r_read_le32 (cursor + 0x14);
-			if (relative_to_strings) {
-				sym->name = str_dup_safe (b, b + start_of_strings + name_off, end);
-			} else {
-				sym->name = str_dup_safe (b, cursor + name_off, end);
-			}
+			string_origin = relative_to_strings? b + start_of_strings : cursor;
+			sym->name = str_dup_safe (b, string_origin + name_off, end);
 			if (!sym->name) {
 				cursor += R_CS_EL_SIZE_SYM;
 				continue;
 			}
-			if (relative_to_strings) {
-				sym->mangled_name = str_dup_safe (b, b + start_of_strings + mangled_name_off, end);
-			} else {
-				sym->mangled_name = str_dup_safe (b, cursor + mangled_name_off, end);
-			}
+			string_origin = relative_to_strings? b + start_of_strings : cursor;
+			sym->mangled_name = str_dup_safe (b, string_origin + mangled_name_off, end);
 			if (!sym->mangled_name) {
 				cursor += R_CS_EL_SIZE_SYM;
 				continue;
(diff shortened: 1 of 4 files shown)


libr/bin/format/mach0/coresymbolication.c
```diff
@@ -185,13 +185,23 @@ RCoreSymCacheElement *r_coresym_cache_element_new(RBinFile *bf, RBuffer *buf, ut
 		goto beach;
 	}
 	ut8 *end = b + hdr->size;
-	if (hdr->file_name_off) {
+	if (file_name) {
+		result->file_name = file_name;
```
trufae, line 189: use ternary operator

libr/bin/format/mach0/coresymbolication.c
```diff
@@ -261,15 +271,20 @@ RCoreSymCacheElement *r_coresym_cache_element_new(RBinFile *bf, RBuffer *buf, ut
...
 			ut64 sect_name_off = r_read_ble (cursor, false, bits);
+			if (!i && !sect_name_off) {
+				relative_to_strings = true;
+			}
 			cursor += word_size;
 			if (bits == 32) {
 				cursor += word_size;
 			}
-			sect->name = str_dup_safe (b, sect_start + (size_t)sect_name_off, end);
+			if (relative_to_strings) {
```
trufae, line 281: ternary operator, call the function once, just make the argument conditional (just as a separate variable passed to the str_dup_safe thing).

Also, is there any reason to have all thoe str_dup_ helpers? looks to me that the code can directly call r_str_ndup()

libr/bin/format/mach0/coresymbolication.c
```diff
@@ -285,20 +300,27 @@ RCoreSymCacheElement *r_coresym_cache_element_new(RBinFile *bf, RBuffer *buf, ut
 			size_t name_off = r_read_le32 (cursor + 0xc);
 			size_t mangled_name_off = r_read_le32 (cursor + 0x10);
 			sym->unk2 = (st32)r_read_le32 (cursor + 0x14);
-			sym->name = str_dup_safe (b, cursor + name_off, end);
+			if (relative_to_strings) {
```
trufae, line 303: ternary

libr/bin/format/mach0/coresymbolication.c
```diff
@@ -285,20 +300,27 @@ RCoreSymCacheElement *r_coresym_cache_element_new(RBinFile *bf, RBuffer *buf, ut
...
+				sym->name = str_dup_safe (b, b + start_of_strings + name_off, end);
+			} else {
+				sym->name = str_dup_safe (b, cursor + name_off, end);
+			}
 			if (!sym->name) {
 				cursor += R_CS_EL_SIZE_SYM;
 				continue;
 			}
-			sym->mangled_name = str_dup_safe (b, cursor + mangled_name_off, end);
+			if (relative_to_strings) {
```
trufae, line 312: const char *mangled_name = relative_to_strings? .. : ...;

libr/bin/p/bin_symbols.c
```diff
@@ -118,7 +118,7 @@ static SymbolsMetadata parseMetadata(RBuffer *buf, int off) {
 	return sm;
 }
 
-static void printSymbolsHeader(SymbolsHeader sh) {
+/*static void printSymbolsHeader(SymbolsHeader sh) {
```
trufae, line 121: comment code with #if 0 instead of multiline comments

libr/bin/p/bin_symbols.c
```diff
@@ -193,7 +193,7 @@ static RBinSymbol *bin_symbol_from_symbol(RCoreSymCacheElement *element, RCoreSy
 	return sym;
 }
 
-static RCoreSymCacheElement *parseDragons(RBinFile *bf, RBuffer *buf, int off, int bits) {
+static RCoreSymCacheElement *parseDragons(RBinFile *bf, RBuffer *buf, int off, int bits, char * file_name) {
```
trufae, line 196: const char*

libr/bin/p/bin_symbols.c
```diff
@@ -289,13 +289,24 @@ static bool load_buffer(RBinFile *bf, void **bin_obj, RBuffer *buf, ut64 loadadd
...
+	//printSymbolsHeader (sh);
 	SymbolsMetadata sm = parseMetadata (buf, 0x40);
-	RCoreSymCacheElement *element = parseDragons (bf, buf, sm.addr + sm.size, sm.bits);
+	char * file_name = NULL;
+	if (sm.namelen) {
+		file_name = malloc (sm.namelen);
+		if (!file_name) {
+			return false;
+		}
+		if (r_buf_read_at (buf, 0x50, (ut8*)file_name, sm.namelen) != sm.namelen) {
```
trufae, line 300: this can result on non-null terminated string, use calloc with namelen+1 for a simple fix

trufae (changes requested):
Damn evolving formats :D

libr/bin/p/bin_symbols.c
mrmacete: good catch!

yeah in all samples i've seen the length actually includes the NUL terminator, but better be ready for it anyways.

libr/bin/p/bin_symbols.c
mrmacete: why `const`? the `file_name` buffer is going to be owned (and later free'd) from the element we're going to create here if everything goes well. If i `const` it here, i'd have to cast the `const` away later in the code, or am i missing something?

libr/bin/format/mach0/coresymbolication.c
mrmacete: the reason for `str_dup_` helpers is that i don't know the length of the strings ahead of time, but i know they are zero-terminated and the helper just avoids me to duplicate the code needed to:
- check the start of string is inside the boundaries of the origin buffer
- calculate the length
- actually call `r_str_ndup`

libr/bin/format/mach0/coresymbolication.c
mrmacete: i'm not sure: this would be two nested ternary operators, does it look more readable without `if`?

```c
	result->file_name = file_name?
		file_name :
		(hdr->file_name_off?
			str_dup_safe (b, b + (size_t)hdr->file_name_off, end) :
			NULL);
```

libr/bin/p/bin_symbols.c
trufae: Oh i see, then its better to use the R_OWNED empty macro just as sugar to make that clear

libr/bin/format/mach0/coresymbolication.c
trufae: Kk that sounds fair enough :)

libr/bin/p/bin_symbols.c
mrmacete: how is the `R_OWN` intended to be used here? at the level of the single argument? or the whole function?

like this:
```c
static RCoreSymCacheElement *parseDragons(RBinFile *bf, RBuffer *buf, int off, int bits, R_OWN char *file_name) {
```

or like this:
```c
R_OWN  static RCoreSymCacheElement *parseDragons(RBinFile *bf, RBuffer *buf, int off, int bits, char *file_name) {
```

?

libr/bin/format/mach0/coresymbolication.h
```diff
@@ -78,7 +78,7 @@ typedef struct r_coresym_cache_element_t {
 	RCoreSymCacheElementLineInfo *line_info;
 } RCoreSymCacheElement;
 
-R_API RCoreSymCacheElement *r_coresym_cache_element_new(RBinFile *bf, RBuffer *buf, ut64 off, int bits);
+R_API RCoreSymCacheElement *r_coresym_cache_element_new(RBinFile *bf, RBuffer *buf, ut64 off, int bits, char * file_name);
```
trufae, line 81:
```suggestion
R_API RCoreSymCacheElement *r_coresym_cache_element_new(RBinFile *bf, RBuffer *buf, ut64 off, int bits, R_OWN char * file_name);
```

libr/bin/p/bin_symbols.c
```diff
@@ -193,7 +179,7 @@ static RBinSymbol *bin_symbol_from_symbol(RCoreSymCacheElement *element, RCoreSy
 	return sym;
 }
 
-static RCoreSymCacheElement *parseDragons(RBinFile *bf, RBuffer *buf, int off, int bits) {
+static RCoreSymCacheElement *parseDragons(RBinFile *bf, RBuffer *buf, int off, int bits, char *file_name) {
```
trufae, line 182:
```suggestion
static RCoreSymCacheElement *parseDragons(RBinFile *bf, RBuffer *buf, int off, int bits, R_OWN char *file_name) {
```

trufae:
Perfect! Merging!

PR #18401, opened 2021-03-01
Implement some more interrupt/syscall lifecycle functions for esil-plugins ##emu ##api

condret:
**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/libr/anal/esil_handler.c
+++ b/libr/anal/esil_handler.c
@@ -57,6 +57,21 @@ R_API bool r_anal_esil_set_interrupt(RAnalEsil *esil, ut32 intr_num, RAnalEsilHa
 	return _set_interrupt (esil, intr, intr_num);
 }
 
+R_API RAnalEsilHandlerCB r_anal_esil_get_interrupt(RAnalEsil *esil, ut32 intr_num) {
+	r_return_val_if_fail (esil && esil->interrupts, NULL);
+	RAnalEsilHandler *handler = _get_interrupt (esil, intr_num);
+	return handler ? handler->cb : NULL;
+}
+
+R_API void r_anal_esil_del_interrupt(RAnalEsil *esil, ut32 intr_num) {
+	r_return_if_fail (esil && esil->interrupts);
+	if (intr_num == 0) {
+		R_FREE (esil->intr0)
+	} else {
+		dict_del (esil->interrupts, intr_num);
+	}
+}
+
 R_API bool r_anal_esil_set_syscall(RAnalEsil *esil, ut32 sysc_num, RAnalEsilHandlerCB cb, void *user) {
 	r_return_val_if_fail (esil && esil->syscalls && cb, false);
 	RAnalEsilHandler *sysc = r_anal_esil_handler_new (cb, user);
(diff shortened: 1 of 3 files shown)


libr/anal/esil_handler.c
```diff
@@ -57,6 +57,21 @@ R_API bool r_anal_esil_set_interrupt(RAnalEsil *esil, ut32 intr_num, RAnalEsilHa
...
+	RAnalEsilHandler *handler = _get_interrupt (esil, intr_num);
+	return handler ? handler->cb : NULL;
+}
+
+R_API void r_anal_esil_del_interrupt(RAnalEsil *esil, ut32 intr_num) {
+	r_return_if_fail (esil && esil->interrupts);
+	if (!intr_num) {
+		R_FREE (esil->intr0)
+		return;
+	}
```
trufae, line 71: else instead of return

libr/anal/esil_handler.c
```diff
@@ -57,6 +57,21 @@ R_API bool r_anal_esil_set_interrupt(RAnalEsil *esil, ut32 intr_num, RAnalEsilHa
...
 
+R_API RAnalEsilHandlerCB r_anal_esil_get_interrupt(RAnalEsil *esil, ut32 intr_num) {
+	r_return_val_if_fail (esil && esil->interrupts, NULL);
+	RAnalEsilHandler *handler = _get_interrupt (esil, intr_num);
+	return handler ? handler->cb : NULL;
+}
+
+R_API void r_anal_esil_del_interrupt(RAnalEsil *esil, ut32 intr_num) {
+	r_return_if_fail (esil && esil->interrupts);
+	if (!intr_num) {
```
trufae, line 68: compare to == 0, using ! is the same, but we are actually comparing to a number, not a pointer or a boolean, using a number looks cleaner to me

trufae (changes requested):
i think esil plugins should provide such primitives as pointers in the esilplugin struct, but overall the only comment i have on this is the will to mix it with RSyscall, but i understand how untied those could be, but maybe it will be good to allow to get the esil syscall name by number.

Also, syscalls in x86 can be tricky. for example on macOS there are xnu and mach syscalls available on different 0x vectors, same goes for DOS which theres an interrupt number and the syscall number. how would this be handled in esil syscalls?

libr/anal/esil_handler.c
condret: done

libr/anal/esil_handler.c
condret: done

condret:
"i think esil plugins should provide such primitives as pointers in the esilplugin struct" @trufae wdym by that? having a pointer to an array of syscall- and interrupt-handlers?

condret:
"Also, syscalls in x86 can be tricky. for example on macOS there are xnu and mach syscalls available on different 0x vectors, same goes for DOS which theres an interrupt number and the syscall number. how would this be handled in esil syscalls?"

ok, I think maybe we need something like a namespace but for bitness. The way how I desire syscalls via interrupts to be handled goes like this:
```
static bool intr_0x80_linux32_handler (RAnalEsil *esil, ut32 intr, void *user) {
     const ut32 sysc = r_reg_getv (esil->anal->reg, "eax");
     return r_anal_esil_do_syscall (esil, sysc);
}
```

But this doesn't respect that you could have code that switches between 32 and 64 bit, ... we could just use the higher bits of the sysc num for 32 bits sysc and do some left shifts ... but that sounds like a hack ... not sure if we can completly generalize this problem

condret:
what if we do it like this
´´´
static bool intr_0x80_linux32_handler (RAnalEsil *esil, ut32 intr, void *user) {
     const ut32 sysc = r_reg_getv (esil->anal->reg, "eax");
     const ut32 bits = esil->anal->bits;
     esil->anal->bits = 32;
     const bool ret = r_anal_esil_do_syscall (esil, sysc);
     esil->anal->bits = bits;
     return ret;
}
´´´

feels to me like the bitness of a syscall should be encoded in the keys of the dictionary ... we could use the upper bits for this. atm it only uses 32 bit keys, which is already more than enough, but we could also use 64 bit keys, if needed

trufae:
lets merge and just move forward, dont wanna block for something we all agree and understand

PR #18417, opened 2021-03-04
Initial implementation for custom bit size cyclic wrapping ##io

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Reference documentation

* https://gitlab.riscosopen.org/RiscOS/Sources/Kernel/-/blob/master/Docs/32bit
* https://en.m.wikipedia.org/wiki/26-bit_computing


condret:
I have a few concerns about this:
1. It restricts user access to io, if you want to use this. Maybe this is the right way to do it, but I'm unsure
2. Seems a bit oversimplified.

trufae:
now its working as expected

PR #18419, opened 2021-03-05
Fix #16254 - grep expression parse improvement ##cons

### LINKED ISSUE #16254: Fix nested grep
See https://github.com/radareorg/radare2/issues/7336#issuecomment-600682778
```
>  pd 10~lea[3]
r8
rcx
> pd 10~lea[3]~cx
r8
rcx
```
 cc @radare @ret2libc

RNRNRNR:
**Checklist**

- [ ] Closing issues: #issue

**Description**
Now grep expression converts from `~str1[0]~str2~str3~?` into `?&str1[0],str2,str3`, and parses in series: modifier, endmodifier, column, row and at the end words. After the parsing of column or row has occurred, it is deleting col or row part by `\0` split, so we get `~str1\00]~str2~str3~?`, lost all data after and cases like a `~[0]~:1` will not work. So better way is to convert the expression to `?&str1[0]~str2~str3` create pointer to every nested grep, split by `\0` and parse every expression apart in the loop

### DIFF
--- a/libr/cons/grep.c
+++ b/libr/cons/grep.c
@@ -373,7 +373,7 @@ static char *find_next_intgrep(char *cmd, const char *quotes) {
 			return (char *)p;
 		}
 		//twiddle unescape
-		memmove (p - 1, p, strlen (p) + 1);
+		r_str_cpy (p - 1, p);
 		cmd = p + 1;
 	} while (*cmd);
 	return NULL;
--- a/libr/cons/grep.c
+++ b/libr/cons/grep.c
@@ -72,9 +72,10 @@ R_API void r_cons_grep_help(void) {
 
 static void parse_grep_expression(const char *str) {
 	static char buf[R_CONS_GREP_BUFSIZE], *ptrs[R_CONS_GREP_COUNT];
-	int wlen, len, ptrs_length, is_range, num_is_parsed, fail = 0;
+	int wlen, len, is_range, num_is_parsed, fail = false;
 	char *ptr, *optr, *ptr2, *ptr3, *end_ptr = NULL, last;
 	ut64 range_begin, range_end;
+	size_t ptrs_length;
 
 	if (!str || !*str) {
 		return;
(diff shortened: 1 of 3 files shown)


libr/cons/grep.c
```diff
@@ -203,128 +97,262 @@ static void parse_grep_expression(const char *str) {
...
+			eprintf ("to many nested greps\n");
+			return;
+		}
+	}
+
+	free (grep->str);
+	grep->str = NULL;
+	bool first = true;
+	sorted_column = 0;
+	for (int i = 0; i < ptrs_length; i++) {
```
trufae, line 118: Use sizet and move the variable definition out of the for block

libr/cons/grep.c
```diff
@@ -203,128 +97,262 @@ static void parse_grep_expression(const char *str) {
...
+						if (is_range) {
+							range_end = r_num_get (cons->num, ptr2);
+							// check for bad value, if range_end == 0, we check if ptr2 == '0'
+							if (range_end == 0 && *ptr != '0') {
+								range_end = -1; // this allow [n- ]
+							}
+						} else {
+							range_begin = range_end = r_num_get (cons->num, ptr2);
+						}
+						num_is_parsed = 1;
```
trufae, line 279: Use bool type here

libr/cons/grep.c
```diff
@@ -340,10 +368,10 @@ static char *find_next_intgrep(char *cmd, const char *quotes) {
 			break;
 		}
 		if (p == cmd || *(p - 1) != '\\') {
-			return (char*)p;
+			return (char *)p;
 		}
 		//twiddle unescape
-		memmove (p - 1, p, strlen(p) + 1);
+		memmove (p - 1, p, strlen (p) + 1);
```
trufae, line 374: r_str_cpy does the strlen inside and used memmove. Is p-1 safe? (I think so, just looking for a quick double check after your changes)

trufae (changes requested):
Lgtm! Just few changes here and there. Thanks!

PR #18440, opened 2021-03-12
Rename r_io_map_resolve and r_io_map_get ##io

condret:
**Checklist**

- [ ] Closing issues: #issue

**Description**

ksnip_20210312-164522


trufae:
Comments:

get_at vs get_in: for analysis we differ those to work in begin address or any address containing it. So imho is better to keep mapget

also. If there are multiple maps in the same vaddr (only one will be the visible, aka highwr prio,) but maybe we want a method to get a list. Thats how it works for the anal api

condret:
I'd have prefered r_io_map_get_at_vaddr and also rename r_io_map_get_paddr to r_io_map_get_at_paddr. I think those would be the most meaning full names.

The reason why I renamed r_io_map_resolve to r_io_map_get is, that map->id relates to a map like desc->fd relates to a desc. There is r_io_desc_get, which returns a desc for a given fd.

PR #18483, opened 2021-03-16
Fix some vulnerabilities

as0ler:
Fixes the following vulnerabilities found by Coverity:
- 1447031 Resource leak
- 1447034 Resource leak
- 1447092 Structurally dead code
- 1447093 Resource leak
- 1448201 Read from pointer after free
- 1448202 Out-of-bounds access

**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -8565,7 +8565,7 @@ static bool cmd_ageh(RCore *core, const char *input) {
 	char *arg = r_str_trim_dup (input + 1);
 	char *sp = strchr (arg, ' ');
 	if (!sp) {
-		free(arg);
+		free (arg);
 		return false;
 	}
 	*sp++ = 0;
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -8565,6 +8565,7 @@ static bool cmd_ageh(RCore *core, const char *input) {
 	char *arg = r_str_trim_dup (input + 1);
 	char *sp = strchr (arg, ' ');
 	if (!sp) {
+		free(arg);
 		return false;
 	}
 	*sp++ = 0;
@@ -8573,6 +8574,7 @@ static bool cmd_ageh(RCore *core, const char *input) {
 
 	const char *k = sdb_fmt ("agraph.edge.0x%"PFMT64x"_0x%"PFMT64x".highlight", a, b);
 	sdb_set (core->sdb, k, add? "true": "", 0);
+	free(arg);
 	return true;
 }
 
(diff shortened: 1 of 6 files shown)


libr/core/cmd_zign.c
```diff
@@ -686,9 +686,9 @@ static const char *getprefix(RSignType t) {
 	case R_SIGN_BBHASH:
 		return "bbhash";
 	default:
-		r_return_val_if_reached ("unknown_type");
+		break;
 	}
-	return "unk";
+	return NULL;
```
condret, line 691: why return NULL here? is there any caller that checks if the return value is NULL and behave different according to that?

libr/core/cmd_anal.c
```diff
@@ -8565,6 +8565,7 @@ static bool cmd_ageh(RCore *core, const char *input) {
 	char *arg = r_str_trim_dup (input + 1);
 	char *sp = strchr (arg, ' ');
 	if (!sp) {
+		free(arg);
```
trufae, line 8568:
```suggestion
		free (arg);
```

libr/core/cmd_anal.c
```diff
@@ -8573,6 +8574,7 @@ static bool cmd_ageh(RCore *core, const char *input) {
 
 	const char *k = sdb_fmt ("agraph.edge.0x%"PFMT64x"_0x%"PFMT64x".highlight", a, b);
 	sdb_set (core->sdb, k, add? "true": "", 0);
+	free(arg);
```
trufae, line 8577:
```suggestion
	free (arg);
```

libr/core/cmd_print.c
```diff
@@ -4689,6 +4689,7 @@ static bool cmd_pi(RCore *core, const char *input, int len, int l, ut8 *block) {
 				// TODO: add limit as arg
 				const char *instruction = r_str_word_get_first (_input + 3);
 				optype = r_anal_optype_from_string (instruction);
```
trufae, line 4691: Dont force non const cast. Change the variable to be non const

libr/fs/p/fs_r2.c
```diff
@@ -260,8 +260,9 @@ static RList *__cfg(RFSRoot *root, const char *path) {
 }
 
 static RList *__root(RFSRoot *root, const char *path) {
+	r_return_val_if_fail (root, NULL);
 	RList *list = r_list_newf (NULL);
-	r_return_val_if_fail (root && list, NULL);
+	r_return_val_if_fail (list, NULL);
```
trufae, line 265: If list fails to be allocated it will be at tuntime so better use normal if check instead of compile time macro

libr/core/cmd_zign.c
trufae: In theory this null is unreadhable code. So better have null deref than undefined behaviour or hiding the error with a valid string

trufae (changes requested):

libr/core/cmd_zign.c
trufae: @as0ler can you check for the callers of this function and see if they handle null properly?

trufae:
Thanks!

Issue #18501, opened 2021-03-23
libr/socket/run.c sets incorrect function pointer


intruder-kat:
Freshly cloned from https://github.com/radareorg/radare2

The following code in `libr/socket/run.c` is supposed to set up the `dyn_openpty`, `dyn_login_tty`, and `dyn_forkpty` function pointers. No other assignments to `dyn_login_tty` or `dyn_forkpty` exist, this is the only place that sets them.
```c
#if HAVE_PTY
static int (*dyn_openpty)(int *amaster, int *aslave, char *name, struct termios *termp, struct winsize *winp) = NULL;
static int (*dyn_login_tty)(int fd) = NULL;
static id_t (*dyn_forkpty)(int *amaster, char *name, struct termios *termp, struct winsize *winp) = NULL;
static void dyn_init(void) {
	if (!dyn_openpty) {
		dyn_openpty = r_lib_dl_sym (NULL, "openpty");
	}
	if (!dyn_login_tty) {
		dyn_openpty = r_lib_dl_sym (NULL, "login_tty");
	}
	if (!dyn_forkpty) {
		dyn_openpty = r_lib_dl_sym (NULL, "forkpty");
	}
}

#endif
```
The lines `dyn_openpty = r_lib_dl_sym (NULL, "login_tty");` and `dyn_openpty = r_lib_dl_sym (NULL, "forkpty");` are setting the `dyn_openpty` function pointer instead of the function pointers they are supposed to be setting up. Consequently, `dyn_login_tty` and `dyn_forkpty` will always be NULL and can never be used, resulting in errors if they are needed. A fixed version of this function would _possibly_ be
```c
static void dyn_init(void) {
	if (!dyn_openpty) {
		dyn_openpty = r_lib_dl_sym (NULL, "openpty");
	}
	if (!dyn_login_tty) {
		dyn_login_tty = r_lib_dl_sym (NULL, "login_tty");
	}
	if (!dyn_forkpty) {
		dyn_forkpty = r_lib_dl_sym (NULL, "forkpty");
	}
}
```
[...]

PR #18505, opened 2021-03-27
correct function pointers and allow fallback on libutil.so for dyn_init

intruder-kat:
**Checklist**

- [x] Closing issues: #18501

I've corrected the function pointers in the original dyn_init logic. If we're on linux and any symbol failed to be found, dl_open libutil.so and try to resolve any fps which are still NULL.

### DIFF
--- a/libr/socket/run.c
+++ b/libr/socket/run.c
@@ -74,11 +74,27 @@ static void dyn_init(void) {
 		dyn_openpty = r_lib_dl_sym (NULL, "openpty");
 	}
 	if (!dyn_login_tty) {
-		dyn_openpty = r_lib_dl_sym (NULL, "login_tty");
+		dyn_login_tty = r_lib_dl_sym (NULL, "login_tty");
 	}
 	if (!dyn_forkpty) {
-		dyn_openpty = r_lib_dl_sym (NULL, "forkpty");
+		dyn_forkpty = r_lib_dl_sym (NULL, "forkpty");
 	}
+
+#ifdef __linux__
+	// for linux distros which use it, attempt to fall back on libutil.so if something failed
+	if (!(dyn_openpty && dyn_login_tty && dyn_forkpty)) {
+		r_lib_dl_open("libutil.so");
+		if (!dyn_openpty) {
+			dyn_openpty = r_lib_dl_sym (NULL, "openpty");
+		}
+		if (!dyn_login_tty) {
+			dyn_login_tty = r_lib_dl_sym (NULL, "login_tty");
+		}
+		if (!dyn_forkpty) {
+			dyn_forkpty = r_lib_dl_sym (NULL, "forkpty");
+		}
+	}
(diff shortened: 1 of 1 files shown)


libr/socket/run.c
```diff
@@ -74,11 +74,27 @@ static void dyn_init(void) {
...
 	}
 	if (!dyn_forkpty) {
-		dyn_openpty = r_lib_dl_sym (NULL, "forkpty");
+		dyn_forkpty = r_lib_dl_sym (NULL, "forkpty");
 	}
+
+#ifdef __linux__
+	// for linux distros which use it, attempt to fall back on libutil.so if something failed
+	if (!(dyn_openpty && dyn_login_tty && dyn_forkpty)) {
+		r_lib_dl_open("libutil.so");
```
trufae, line 86: check if it returns NULL; if so, you can skip the lines below

libr/socket/run.c
```diff
@@ -74,11 +74,27 @@ static void dyn_init(void) {
...
+		if (!dyn_openpty) {
+			dyn_openpty = r_lib_dl_sym (NULL, "openpty");
```
trufae, line 88: better pass the returned pointer from dl_open instead of NULL, so you ensure the attempt to resolve the symbols from this specific library instead from the global namespace

libr/socket/run.c
trufae, line 86:
```suggestion
		r_lib_dl_open ("libutil.so");
```

trufae: its fine to hardcode .so here, because we are in __linux__ specific thing.. but this libutil is not just from Linux, some BSD systems also use `R_LIB_EXT` instead. and remove the ifdef __linux__ thing, because this can apply to cygwin, macOS and freebsd at least., not just linux. and afaik it all depends on how libc is compiled, so imho its better to do this thing linking at runtime

trufae (changes requested):

trufae:
Please rebase to get the CI green

libr/socket/run.c
```diff
@@ -79,19 +79,22 @@ static void dyn_init(void) {
 	if (!dyn_forkpty) {
 		dyn_forkpty = r_lib_dl_sym (NULL, "forkpty");
 	}
-
-#ifdef __linux__
-	// for linux distros which use it, attempt to fall back on libutil.so if something failed
+#ifdef R_LIB_EXT
```
trufae, line 82: This variable is always set. Better wrap this in #if __UNIX__

trufae, line 82:
```suggestion
#if __UNIX__
```

trufae (changes requested):

trufae:
Thanks!!

libr/socket/run.c
intruder-kat: thanks!

PR #18509, opened 2021-03-28
Abstract r_diff_levenshtein_path ##diff

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This abstracts the `r_diff_levenshtein_path` function so it can compare things other then byte buffers. The `RLevBuf` now holds a `void *` and the caller provides a function pointer to do comparisons between two buffers. So this could be used to compare ESIL strings. I ran my differential fuzzer on this using a simple `ut8 *` buffers and it seemed to work properly.

Double check that the function declaration and types make sense. Nothing relies on this code yet, so it's best to nit pick it now. Maybe it would be better to store both buffers and their lengths in a single struct like the following?

```
typedef struct r_lev_buf {
    void *bufa, *bufb;
    ut32 lena, lenb;
    void *user; // for callback, just in case
}
```

I am terrible at organizing this kind of stuff, let me know what you think. I am happy to change anything.

### DIFF
--- a/libr/include/r_diff.h
+++ b/libr/include/r_diff.h
@@ -52,9 +52,10 @@ typedef enum {
 } RLevOp;
 
 typedef struct r_lev_buf {
-	ut8 *buf, *mask;
+	void *buf;
 	ut32 len;
 } RLevBuf;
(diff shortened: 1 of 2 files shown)


trufae (approved):
Lgtm!

PR #18517, opened 2021-03-31
Fix pcdelta for ARM esil LDR ##anal

Taggggy:
**Checklist**

- [ ] Closing issues: #issue

**Description**
ARM LDR translation in ESIL was wrong when pcdelta was not zero.
We have to add it to the $$ value.


trufae:
Can you provide a test case for this fix? it will help to avoid a regression in case something changes in the future. thanks!

Taggggy:
I added the test, it should be ok now.

trufae:
Perfect! 👌

PR #18528, opened 2021-04-01
Update to use keys that can be programmed onto a CPS2.

el-goe:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Current MAME CPS2 roms include the keys that can be used to restore dead boards as seen in https://github.com/ArcadeHacker/ArcadeHacker_CPS2.  These are incompatible with the code currently present in Radare2.


trufae:
What about keeping the support for the 8 byte keys? i'll invoke @poliva for reviewing this. Thanks!

trufae:
Also, please add/fix tests

poliva:
Thanks for the PR @el-goe, nowadays it makes much more sense to use 20-byte keys as new romsets have them inside the key files instead of the old 8-byte keys hardcoded in old mame versions, but for the sake of retro-compatibility I'd keep both options available (8 byte & 20 byte keys).

el-goe:
Cool to me to keep both options, but afaik old hardcoded keys are not valid anymore for the current roms, so the user needs to decode the 20-byte ones anyways.

PR #18535, opened 2021-04-03
Add abstract Levenshtein dist ##diff

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Add `r_diff_levenshtein_dist` which acts like `r_diff_levenshtein_path` but does not allocate the matrix or return the path through the matrix. 

This is pretty much identical to `r_diff_buffers_distance_levenshtein` except `r_diff_levenshtein_dist` is abstract. It can work on any data type since comparisons are done via a function pointer. The `r_diff_levenshtein_dist` function also has the same `maxdist` logic from `r_diff_levenshtein_path` which can improve speed. In particular, `zb` and `zbr` should see a speed increase in using the new function because of an early exit when two signatures show differences early but are of similar size.

The new function will also attempt to calculate fewer cells as the number of changes seen increases. This could have a speedup but also comes at the expense of code complexity. I would not get rid of `r_diff_buffers_distance_levenshtein`, it would be good to compare the results of both to each other to find bugs. I did do differential fuzzing on this again and got through 9M executions of AFL without finding a difference.

If this is accepted I will go about switching `zb` and `zbr` to this function so we can have some regression testing for it.

### DIFF
--- a/libr/include/r_diff.h
+++ b/libr/include/r_diff.h
@@ -91,7 +91,6 @@ R_API int r_diff_gdiff(const char *file1, const char *file2, int rad, int va);
 R_API RDiffChar *r_diffchar_new(const ut8 *a, const ut8 *b);
 R_API void r_diffchar_print(RDiffChar *diffchar);
 R_API void r_diffchar_free(RDiffChar *diffchar);
-R_API st32 r_diff_levenshtein_dist(RLevBuf *bufa, RLevBuf *bufb, ut32 maxdst, RLevMatches levdiff);
 R_API st32 r_diff_levenshtein_path(RLevBuf *bufa, RLevBuf *bufb, ut32 maxdst, RLevMatches levdiff, RLevOp **chgs);
 #endif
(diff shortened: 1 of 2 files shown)


trufae (approved):

trufae:
What about making this function static and call it when the last arg is null?

swoops:
> What about making this function static and call it when the last arg is null?

That is a great idea. There is also a lot of repeated code easily be dropped that way. I did not want to try to drop that code before and make things difficult to read.

Give me a bit and I will have an update.

swoops:
I decided to be lazy for now. I want to remove the repeated code for sufix and prefix matches but I don't want to over complicate things. For now this will work. Then I can get `zb` and `zbr` to use this, get some tests working, then remove some of the repeated code when tests are available to ensure I am not doing something stupid.

trufae:
Looks good to me! Yeah you can remove the duplicated code later.

PR #18546, opened 2021-04-07
Fix coverity issue in Mach-O format

mrmacete:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Fix the last High coverity issue in Mach-O which is not a false positive. Slipped through previous passes.


trufae:
thanks!

PR #18526, opened 2021-04-01
Add recursive list functionality to libr/util/file.c

RHL120:
**Checklist**

- [ ] Closing issues: #issue
- [X] Mark this if you consider it ready to merge
- [ ] I've added tests (optional)
- [ ] I wrote some lines in the [radare2book](https://github.com/radareorg/radare2book)

**Description**
This patch is supposed to add the functionality given by the unix command find <dir>. In the future this functionality might enable fetures like recursive copy and remove

PR #18552, opened 2021-04-08
Tried to fix certain problems in r_file_dir_recursive and added it to a cmd.c

RHL120:
**Checklist**

- [ ] Closing issues: #issue

**Description**
A while ago I wrote this stupid function https://github.com/radareorg/radare2/pull/18526 
of course me being me I fucked up the code and unfortunately the issues went over the reviewer's head so I hopefuly fixed some of the glraing issues in the function. Now the function can be accesed by the cmdline.

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -1267,11 +1267,11 @@ static int cmd_lsr(RCore *core, const char *input) {
 	}
 	files = r_file_lsrf (arg);
 	if (!files) {
-		eprintf ("Failed to read directories");
+		eprintf ("Failed to read directories\n");
 		return 0;
 	}
 	r_list_foreach (files, iter, path) {
-		printf ("%s\n", path);
+		r_cons_println (path);
 	}
 	r_list_free (files);
 	return 0;
@@ -1286,7 +1286,7 @@ static int cmd_ls(void *data, const char *input) { // "ls"
 	}
 	switch (*input) {
 	case 'r':
-		cmd_lsr(core, input);
+		cmd_lsr (core, input);
 		break;
 	case '?': // "l?"
 		eprintf ("Usage: l[es] # ls to list files, le[ss] to less a file\n");
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -1255,18 +1255,39 @@ R_API bool r_core_run_script(RCore *core, const char *file) {
 	return ret;
 }
 
+static int cmd_lsr(RCore *core, const char *input) {
+	const char *arg;
+	char *path;
+	RList *files;
+	RListIter *iter;
+	if (!arg) {
+		arg = ".";
+	} else {
+		arg = input;
+	}
+	files = r_file_lsrf (arg);
+	if (!files) {
+		eprintf ("Failed to read directories");
+		return 0;
+	}
+	r_list_foreach (files, iter, path) {
+		printf ("%s\n", path);
+	}
+	r_list_free (files);
+	return 0;
+}
+
 static int cmd_ls(void *data, const char *input) { // "ls"
 	RCore *core = (RCore *)data;
-	RListIter *iter;
-	char *file;
-	RList *files;
-	const char *dir;
 	int ret;
 	const char *arg = strchr (input, ' ');
 	if (arg) {
 		arg = r_str_trim_head_ro (arg + 1);
 	}
 	switch (*input) {
+	case 'r':
+		cmd_lsr(core, input);
+		break;
 	case '?': // "l?"
 		eprintf ("Usage: l[es] # ls to list files, le[ss] to less a file\n");
 		break;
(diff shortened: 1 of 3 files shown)


libr/core/cmd.c
```diff
@@ -1272,6 +1277,23 @@ static int cmd_ls(void *data, const char *input) { // "ls"
 			eprintf ("Usage: less [file]\n");
 		}
 		break;
+	case 'r':
+		files = r_list_new ();
```
trufae, line 1281: better move this logic into cmd_lsr(RCore *core, const char *input); so you can put those local variables locally there instead of having more variables that are only used in one execution path in this function.

trufae (changes requested):
lgtm apart from that

libr/util/file.c
```diff
@@ -1276,29 +1276,41 @@ R_API bool r_file_copy(const char *src, const char *dst) {
...
+		absp = r_file_abspath (name);
+		if (!absp) {
+			ret = false;
+			break;
+		}
+		r_list_append (dst, absp);
 		if (r_file_is_directory (name)) {
 			ret = r_file_dir_recursive (dst, name);
+			if (!ret) {
+				ret = false;
```
trufae, line 1306: uh wait. whats the point of this? ret is false already in here

libr/util/file.c
```diff
@@ -1276,29 +1276,41 @@ R_API bool r_file_copy(const char *src, const char *dst) {
...
+				break;
+			}
 		}
 	}
+	r_list_free (files);
 	r_sys_chdir (cwd);
```
trufae, line 1312: i think it is doable to implement this function wiuthout having to chdir. Otherwise running this function in a thread can result on unexpected problems

libr/util/file.c
```diff
@@ -1276,29 +1276,41 @@ R_API bool r_file_copy(const char *src, const char *dst) {
 }
 
 R_API bool r_file_dir_recursive(RList *dst, const char *dir) {
```
trufae, line 1278: imho this function should be defined like this:

```R_API RList *r_sys_dir_recursive(const char *path) {```

and return the list instead of bool. we can later decide if dirlisting apis are related to RFile or RSys later.

File: libr/util/file.c [hunk 5]
```diff
@@ -1276,29 +1276,41 @@ R_API bool r_file_copy(const char *src, const char *dst) {
...
-	bool ret = false;
+	bool ret = true;
 	char *cwd = r_sys_getdir ();
 	if (!cwd) {
 		return false;
 	}
 	if (r_sys_chdir (dir) == false) {
```
trufae, line 1284: better get rid of this call and use ! instead of == false

libr/util/file.c
```diff
@@ -1276,29 +1276,41 @@ R_API bool r_file_copy(const char *src, const char *dst) {
...
 		free (cwd);
-		return ret;
+		return false;
 	}
 	RList *files = r_sys_dir (".");
 	RListIter *iter;
 	char *name;
 	r_return_val_if_fail (files, false);
```
trufae, line 1291: ONLY use r_return_* at the begining of functions to check function arguments. dont use that to chk what function calls return

libr/util/file.c
```diff
@@ -1276,29 +1276,41 @@ R_API bool r_file_copy(const char *src, const char *dst) {
...
 	r_list_foreach (files, iter, name) {
+		char *absp;
 		if (strcmp (name, ".") == 0 || strcmp (name, "..") == 0) {
```
trufae, line 1294: use !strcmp instead of == 0

trufae (changes requested):
.

libr/util/file.c
RHL120: The point is to prove stupidity ;)

libr/util/file.c
RHL120: Yes exactly, I tried to avoid that to make less use of memory but now that I look at I already allocate cwd so I might aswell just replace that with the next path.

libr/util/file.c
trufae: ?

libr/core/cmd.c
```diff
@@ -1255,18 +1255,39 @@ R_API bool r_core_run_script(RCore *core, const char *file) {
...
+	} else {
+		arg = input;
+	}
+	files = r_file_lsrf (arg);
+	if (!files) {
+		eprintf ("Failed to read directories");
+		return 0;
+	}
+	r_list_foreach (files, iter, path) {
+		printf ("%s\n", path);
```
trufae, line 1274: Use r_cons_println or r_cons_printf

libr/core/cmd.c
```diff
@@ -1255,18 +1255,39 @@ R_API bool r_core_run_script(RCore *core, const char *file) {
...
-	RList *files;
-	const char *dir;
 	int ret;
 	const char *arg = strchr (input, ' ');
 	if (arg) {
 		arg = r_str_trim_head_ro (arg + 1);
 	}
 	switch (*input) {
+	case 'r':
+		cmd_lsr(core, input);
```
trufae, line 1289: Missing space

trufae (changes requested):

libr/core/cmd.c
```diff
@@ -1260,7 +1260,7 @@ static int cmd_lsr(RCore *core, const char *input) {
 	char *path;
 	RList *files;
 	RListIter *iter;
-	if (!arg) {
+	if (!input) {
 		arg = ".";
```
trufae, line 1264: Maybe better use R_STR_ISEMPTY. Considering “” is the same as NULL

PR #18578, opened 2021-04-13
filename printing improvements in rafind2 ##search

tgorochowik:
**Checklist**

- [ ] Closing issues: #issue

**Description**

The current implementation of the filename header wasn't very useful when there are many files without any match, for example:
```
$ rafind2 -s R2AGENT man
File: man/rax2.1
File: man/rarun2.1
File: man/radare2.1
File: man/r2agent.1
0x15
0x2c
File: man/r2pm.1
File: man/r2-docker.1
File: man/rahash2.1
File: man/radiff2.1
File: man/r2r.1
File: man/rafind2.1
File: man/rasm2.1
File: man/esil.7
File: man/rasign2.1
File: man/ragg2.1
File: man/rabin2.1
```
This gets worse if there are more files.

And even worse when it is combined with other printing modes (like json, where the header literally breaks the json by printing filenames in between).

The first patch changes the way the header is printed o more grep-like approach - the name is printed before each match:
```
$ rafind2 -s R2AGENT man
man/r2agent.1: 0x15
man/r2agent.1: 0x2c
```
(it can still be disabled with the `-q` swtich)

The second patch adds `file` entry to the json output, it now looks like this:
```
[(null){"file":"README.md","offset":185,"type":"string","data":"radare.org"},{"file":"README.md","offset":2765,"type":"string","data":"radare.org/](https://www.radare.org/)"},{"file":"README.md","offset":2790,"type":"string","data":"radare.org/)"}]
```

Note the `(null)` at the beginning of each json. The third patch solves that by initializing `comma` to an empty string, I think it was accidentally removed in 66bd6ed8706857cd7e17e94abedc53925bffee67

Note that the search types that use `r_sys_cmdf` do not now have the header printing feature, possible approaches here:
* Restore the original filename printing here (despite the flood for large directories) - should be fairly easy to add.
* Revisit this after their TODOs (_implenet using api_) are done (I might want to look at this later although I cannot promise anything).

### DIFF
--- a/libr/main/rafind2.c
+++ b/libr/main/rafind2.c
@@ -271,8 +271,6 @@ static int rafind_open_file(RafindOptions *ro, const char *file, const ut8 *data
 				r_search_kw_add (rs, r_search_keyword_new_str (kw, ro->mask, NULL, 0));
 			}
 		}
-	} else if (ro->mode == R_SEARCH_STRING) {
-		r_search_kw_add (rs, r_search_keyword_new_hexmask ("00", NULL)); //XXX
 	}
 
 	ro->curfile = file;
--- a/libr/main/rafind2.c
+++ b/libr/main/rafind2.c
@@ -33,7 +33,7 @@ typedef struct {
 	RList *keywords;
 	const char *mask;
 	const char *curfile;
-	const char *comma;
+	PJ *pj;
 } RafindOptions;
 
 static void rafind_options_fini(RafindOptions *ro) {
@@ -47,6 +47,7 @@ static void rafind_options_init(RafindOptions *ro) {
 	ro->bsize = 4096;
 	ro->to = UT64_MAX;
 	ro->keywords = r_list_newf (NULL);
+	ro->pj = NULL;
 }
 
 static int rafind_open(RafindOptions *ro, const char *file);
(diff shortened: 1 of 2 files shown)


libr/main/rafind2.c
```diff
@@ -47,6 +47,7 @@ static void rafind_options_init(RafindOptions *ro) {
 	ro->bsize = 4096;
 	ro->to = UT64_MAX;
 	ro->keywords = r_list_newf (NULL);
+	ro->comma = "";
```
trufae, line 50: using the PJ api this field is not necessary

trufae (changes requested):
i like the grep-like output! good proposal. iirc after your patch -q is keeping the offset-per-line output, right?

tgorochowik:
Thanks for the review!

Rewrote the json output using pj. I had to add an extra check at the beginning not to do anything when using modes which use their own json output. I hope that's okay.

In addition, the format script started asking me to re-format some lines which I did not change (again), I took a look at it and in my opinion there are some issues with the detector (at least when used in the recommended way), see #18580

trufae:
Awesome! Now looks good! Ill check the clang format script too! I use to indent evrything by hand always. All the tools that indent C have issues and its not that hard for me to do it by hand. But if we can make it indent properly the r2 codebase it will be a good step forwArd but i assume will only work on some specific versions of clang. I recall reporting issues in previous versions but when the new updates came out nobody had them available in their systems.. so after all this time maybe its now good to go

PR #18589, opened 2021-04-16
Rafind regexp fixes ##search

tgorochowik:
strlen does not count the trailing \0 - the r_search_keyword_new call
uses memcpy internally to copy the string based on the length provided
in the argument.

Effectively we do not copy the trailing \0 to the keyword and the search
does not always work as intended (sometimes the garbage from the memory
is 'joined' to the keyword string).

**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/libr/main/rafind2.c
+++ b/libr/main/rafind2.c
@@ -464,7 +464,7 @@ R_API int r_main_rafind2(int argc, const char **argv) {
 	if (opt.ind + 1 == argc && !r_file_is_directory (argv[opt.ind])) {
 		ro.quiet = true;
 	}
-	if (ro.json && ro.mode == R_SEARCH_KEYWORD) {
+	if (ro.json && (ro.mode == R_SEARCH_KEYWORD || ro.mode == R_SEARCH_REGEXP)) {
 		// TODO: remove mode check when all modes use api
 		ro.pj = pj_new ();
 		pj_a (ro.pj);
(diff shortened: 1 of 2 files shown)


tgorochowik:
(looking at CI failures...)

trufae:
I think searching for strings wont work after this change. Where do u see the trashed memory without this patch? Do u have a test for it?

tgorochowik:
Yep, searching for regular strings broke, it fixes
`rafind2 -e README -e This test/bins/arm/README` for me - I was too quick submitting this patch. Give me a moment for further investigation

tgorochowik:
So.. searching for strings assumes the trailing \0 is missing is that correct? It looks as though regexes do not - I think that is the issue here

tgorochowik:
@trufae yes, that was the issue, regexp search expects the keyword to be a proper null-terminated string and that was the issue I was trying to fix. I added a fix that copies the string data to a new memory region where it can freely add the trailing null. I guess that's not perfect, but it solves the issue - is there a way to do that without copying?

I also added a fix for regexp json output in rafind2 and multiple new tests for rafind (including what triggered this issue here)

tgorochowik:
I think something still fails..

tgorochowik:
third time's a charm, should be okay now.

I guess #18586 wasn't actually correct - the code cannot be reused between string keywords and regexp keywords because they handle string termination differently - and there actually IS a proper api regexp keywords - it just requires the /canonical/ form - I did not know that...

It should be okay now - the CI is green.

Issue #18632, opened 2021-04-25
Write a not zero terminated command for `pr` (print raw)


gogo2464:
## Environment

```sh
# copypaste this script into your shell and replace it with the output
date
r2 -v
uname -ms
```

## Description

I can not do `pr 0x188 > bin.bin`; because my datas to copy contain 00 bytes.

We could make a new command such as `prz` but `prz` already exists. `prz` is supposed to solve this issue like `psz` but I still can not ignore `00` with both `pr` and `prz`.

gogo2464:
The pr implementation is here: https://github.com/radareorg/radare2/blob/master/libr/core/cmd_print.c#L6158 check out printeay https://github.com/radareorg/radare2/blob/702e1c1adae8d297ce23fab71f7c6b3140d2cc39/libr/core/cmd_print.c#L2573 and r_print_raw https://github.com/radareorg/radare2/blob/777ca01c1860582836fc2836af5d32f1a8d7bd7f/libr/util/print.c

gogo2464:
I also plan to refactor these conditions by a single switch.

gogo2464:
Ready to merge https://github.com/radareorg/radare2/pull/18634

PR #18634, opened 2021-04-26
refactor the r_print_raw function in libr/util/print.c to replace a s…

gogo2464:
…witch by a switch

**Checklist**

- [ ] Closing issues: #issue

**Description**

start to fix https://github.com/radareorg/radare2/issues/18632 . 

Not finished. Already fix a refactoring.

### DIFF
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -1268,7 +1268,7 @@ R_API int r_cons_get_column(void) {
 
 /* final entrypoint for adding stuff in the buffer screen */
 R_API int r_cons_write(const char *str, int len) {
-	if (R_STR_ISEMPTY (str) || len < 0) {
+	if (len < 0) {
 		return -1;
 	}
 	if (I.echo) {
--- a/libr/util/print.c
+++ b/libr/util/print.c
@@ -1485,7 +1485,12 @@ R_API void r_print_bytes(RPrint *p, const ut8 *buf, int len, const char *fmt) {
 }
 
 R_API void r_print_raw(RPrint *p, ut64 addr, const ut8 *buf, int len, int offlines) {
-	if (offlines == 2) {
+	switch (offlines) {
+	case 0:
+		p->write (buf, len);
+		break;
+	case 2:
+	{
 		int i, j, cols = p->cols * 4;
 		char ch;
 		for (i = 0; i < len; i += cols) {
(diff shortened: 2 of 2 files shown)


libr/util/print.c
```diff
@@ -1485,56 +1485,64 @@ R_API void r_print_bytes(RPrint *p, const ut8 *buf, int len, const char *fmt) {
...
-					r_print_cursor (p, i + j, 1, 0);
+				p->cb_printf ("\n");
+			}
+			break;
+		}
+		default:
+		{
+			const ut8 *o, *q;
+			ut64 off;
+			int i, linenum_abs, mustbreak = 0, linenum = 1;
```
trufae, line 1519: mustbreak should be `bool`, not int

trufae (changes requested):

libr/util/print.c
gogo2464: I just made a copy paste. I am going to solve.

gogo2464:
Ready to merge. Just do some tests before and wait for the automatic tests. I have removed the check for the size. It just checks the length instead. I hope it does not break nothing.

libr/cons/cons.c
```diff
@@ -1268,7 +1268,7 @@ R_API int r_cons_get_column(void) {
 
 /* final entrypoint for adding stuff in the buffer screen */
 R_API int r_cons_write(const char *str, int len) {
-	if (R_STR_ISEMPTY (str) || len < 0) {
+	if (len < 0) {
```
trufae, line 1271: can you add ` r_return_val_if_fail (str && len <1);`

and make ` if (len == 0) return 0;`?

trufae (approved):
The PR LGTM! i have tested it and it seems to work as expected. Can you address the comment and add a test? i was thinking in having something like this:
```
wx 4400440044004400
pr 10 > $x
wff $x
x8
```

but seems like the $alias files are not working well with wff. so i'll merge and add a test after fixing that. thank you!

libr/cons/cons.c
gogo2464: Yes I will do it this today.

trufae: I already did that i. Master. Do you want to extend alias files to support null bytes too? I can explain better about what are alias files and how to handle them in discord

PR #18651, opened 2021-05-03
Refactor signature matching ##signatures

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Mostly refactors code. This also makes reporting of found signatures a bit more verbose (showing specific types found). 

A very minor bug was fixed. Before two different search contexts were used. This meant the match count for byte signatures was unique, while graph, refs, offset, etc all shared a second match count. Now all signatures share the same match count. 

Before you would get flags like:
```
0x00400536 37 sign.bytes.sym.print_0
0x00400440 41 sign.graph.sym.print_0
0x00400420 6 sign.graph.sym.print_1
0x00400410 6 sign.graph.sym.print_2
0x0040055b 32 sign.graph.sym.print_3
0x00400536 37 sign.graph.sym.print_4
```

Now you get:
```
0x00400536 37 sign.bytes.sym.print_0
0x00400440 41 sign.graph.sym.print_1
0x00400420 6 sign.graph.sym.print_2
0x00400410 6 sign.graph.sym.print_3
0x0040055b 32 sign.graph.sym.print_4
0x00400536 37 sign.graph.sym.print_5
```

I could easily make it so that none of the signature types share a count, if that is preferred (just use `ctxSearchCB[graph_count, offset_count, bytes_count...]` instead of `ctxSearchCB.count`). 

This fix resulted in a minor change to to one of the test files. There are still 3 broken tests, in `test/db/cmd/cmd_zignature`. All are powerpc tests. I will make a issue for this tomorrow, if there isn't one. I suspect `PowerPC r_sign_fcn_bytes bounds check` test is correct and the new powerpc analysis is masking out a relative branch when it shouldn't. I also think `PowerPC function with jumpback` test is wrong and the new ppc analysis does better. But I don't know powerpc well enough to say for sure.

### DIFF
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -2312,198 +2312,130 @@ static int matchCount(int a, int b) {
 	return R_ABS (c) < m;
 }
 
-static bool fcnMetricsCmp(RSignItem *it, RAnalFunction *fcn) {
-	RSignGraph *graph = it->graph;
-	int ebbs = -1;
-
-	if (graph->cc != -1 && graph->cc != r_anal_function_complexity (fcn)) {
+static bool graph_match(RSignGraph *a, RSignGraph *b) {
+	if (!a || !b) {
 		return false;
 	}
-	if (graph->nbbs != -1 && graph->nbbs != r_list_length (fcn->bbs)) {
+	if (a->cc != -1 && a->cc != b->cc) {
 		return false;
 	}
-	if (graph->edges != -1 && graph->edges != r_anal_function_count_edges (fcn, &ebbs)) {
+	if (a->nbbs != -1 && a->nbbs != b->nbbs) {
 		return false;
 	}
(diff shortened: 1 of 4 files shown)


trufae:
The tests that fail arepeobably with capstone4. Rimraf shlr/capstone to repro

Any reason why results are not sorted by offset?

trufae (approved):

chinggg:
The refactor is great though it cannot improve the zignature search during `aaa` 👍

trufae:
@chinggg any reproducer/test to guide on this?

swoops:
I think @chinggg is referring to a rizin issue about multiple signatures matching a single function.

This refactor is just the first step tworard dealing with signature collisions. Next pull I send will get even closer but it might take a bit. I am quick to embarrass myself with `static` functions, I am slower to embarrass myself in front of all `r_sign` namespace.

PR #18658, opened 2021-05-05
simple fix priorities.

ihsinme:
I think you have a confusion with priorities.
`unsigned short` cannot give anything other than `zero` in this operation.

### DIFF
--- a/libr/asm/arch/v850/v850_disas.h
+++ b/libr/asm/arch/v850/v850_disas.h
@@ -179,7 +179,7 @@ static inline ut8 get_opcode(const ut16 instr) {
 
 // FIXME: XXX: Invalid for extended instruction format 4 (Format XII)!
 static inline ut8 get_subopcode(const ut16  instr) {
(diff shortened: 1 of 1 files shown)


trufae:
In this case the shift is unnecessary, just place the mask value directly. Can you provide any example of instruction that gets fixed with this? Im currently using r2ghidra disassembler for v850, because this one is pretty broken and doesnt support e2

ihsinme:
I fixed PR

Unfortunately, I do not know your product so well and I cannot give an example.
I just see an obvious error in the logic of the program.

trufae:
Dont mark the PR as draft because that way the CI is not executed. i'll check tonight if your blindfix solves the problems in the disassembler, because i had already started to write my own v850e2 disassembler some time ago, which was automatically generated from the specs pdf. But didnt had time to finish.

trufae:
Seems like its breaking all the v850 amalysis tests

ihsinme:
I cannot help you with this.

PR #18683, opened 2021-05-11
Improve the xxhash system library detection

xambroz:
The pkgconfig module is named libxxhash on Debian/Fedora/RedHat/SuSE/Arch so I added libxxhash on first place.
Keeping the xxhash checked second.

**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -237,8 +237,11 @@ if sys_magic.found() and get_option('use_sys_magic')
 endif
 
 # handle xxhash library
-sys_xxhash = dependency('xxhash', required: false)
+sys_xxhash = dependency('libxxhash', required: false)
 use_sys_xxhash = false
+if not sys_xxhash.found()
+  sys_xxhash = dependency('xxhash', required: false)
+endif
 if not sys_xxhash.found()
   sys_xxhash = cc.find_library('xxhash', required: false)
 endif


trufae:
Is the configure check working well? Do you think its worth handling it via pkgconfig in acr? https://github.com/radareorg/radare2/blob/9ad922df069487a44ff760863e70bf3ad18d4a64/configure.acr#L66

xambroz:
Hello,
in the meson configuration the check was actually also working, but not thanks to pkgconfig.
It was always checking "pkgconf xxhash" (and failing on majority of linux distributions):
```
sys_xxhash = dependency('xxhash', required: false)
```
 and then falling back to finding the library, which was working OK.
```
if not sys_xxhash.found()
   sys_xxhash = cc.find_library('xxhash', required: false)
endif
``` 
I do not know whether or not there is some platform where there would be xxhash.pc pkgconfig module.
Majority of linux distributions seems to have libxxhash.pc in the recent releases.


> Is the configure check working well? 

yes configure works OK finding the libxxhash successfully on Fedora (and probably others as well).
image

> Do you think its worth handling it via pkgconfig in acr?

Probably not critical thing to change (as options for xxhash are pretty straightforward), but it is possible.
Still on some platforms you will still need fallback to the library detection as you have now.

I have checked that Ubuntu/Debian/Fedora/RedHat/SuSE/Arch all have the pkgconfig module libxxhash.pc in the most recent release. So yes you can add/switch for the pkgconfig check even in the configure. You already use pkgconfig for openssl, capstone, libuv in configure so you could use it for xxhash (and others) as well.

BTW as Fedora is moving to pkgconfig even in defining the dependencies, I can tell there are even more of the modules, which are available on Fedora (and possibly others):
pkgconfig(libxxhash)
pkgconfig(libzip)
pkgconfig(zlib)
pkgconfig(liblz4)
pkgconfig(capstone)
pkgconfig(libuv)
pkgconfig(openssl)
pkgconfig(libmagic)
pkgconfig(bzip2)

Be aware that on RHEL7 for example file-devel and bzip2-devel didn't provide the pkgconfig modules, RHEL8 has bzip2 but also missing libmagic.pc in file-devel. I guess similar would apply for the long-term support versions of Debian/Ubuntu.

Michal Ambroz

trufae:
Lets merge it as is :) it’s already an improvement so no need to go for acr in here. Thanks for the detailed explanation!

PR #18692, opened 2021-05-14
Add collisions type to ##signatures

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**
This commit ads the R_SIGN_COLLISIONS ('C') signature type, which will later be used to resolve what an unkown function should be named when multiple signatures match. Currently, there is not enough code around this signature type to make it very useful. Additionally, the way matching is done with byte signatures and graph signatures being matched separately,  r2 can not fully benefit from this feature. I am sending this now in hopes for comment, let me know if you think I can do something better. Also, I am sending this now because it can help signature database maintainers (if there are any) trim their databases.

This pull ads the `zac` command to add zignature collisions. This is another signature type that is added to `RSignItem` and is just a `RList` of all the other signature type,name tuple that collide with the current signature.  Two graph type signatures can be created that match, running `zac` creates the collision signature for everything in the current sign space: 
```
[0x00000000]> za main1 g cc=1 nbbs=2 edges=3 ebbs=4 bbsum=5
[0x00000000]> za main2 g cc=1 nbbs=2 edges=3 ebbs=4 bbsum=5
[0x00000000]> z
main1:
  graph: cc=1 nbbs=2 edges=3 ebbs=4 bbsum=5
main2:
  graph: cc=1 nbbs=2 edges=3 ebbs=4 bbsum=5
[0x00000000]> zac
[0x00000000]> z
main1:
  graph: cc=1 nbbs=2 edges=3 ebbs=4 bbsum=5
  collisions: graph:main2
main2:
  graph: cc=1 nbbs=2 edges=3 ebbs=4 bbsum=5
  collisions: graph:main1
```

If the signature `A` matches signature `B` on some metric, and signature `A` matches a function on that same metric, then that function would also match signature `B`. Having a list of collisions computed ahead of time allows the matching algorithm to make better decisions on whether the matched function can confidently be renamed. 

Collisions signatures are the only signature type that depends on the other signatures. If a two signature databases are combined, the collisions will change. For this reason, collisions signatures are not automatically generated  with `zg`, `zaf` and `zaF`. I could use a `e zign.collisions` flag to change this behavior in the future. This means you can loop over all the `.o` files in a `libc.a` archive, add each signature to a single sdb file. Then a single `zac` will efficiently create collisions. 

This brings in the first bug that needs to be fixed. Currently `zac` only appends to the collision signatures, it should delete previous collisions and start over. Because of this behavior, and lack of utility so far, you may not want to merge this yet.

Let me know if you think this is a good ide, if something other then `zac` should be used or if you had something nice happen to you today :).

### DIFF
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -2367,6 +2367,11 @@ static int match_metrics(RSignItem *it, void *user) {
 }
 
 static int _sig_to_vec_cb(RSignItem *it, void *user) {
+	RSignItem *it = (RSignItem *)user;
+	if (it->collisions) {
+		r_list_free (it->collisions);
+		it->collisions = NULL;
+	}
 	return r_pvector_push ((RPVector *)user, it)? 1: 0;
 }
 
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -313,6 +313,10 @@ R_API bool r_sign_deserialize(RAnal *a, RSignItem *it, const char *k, const char
 			DBL_VAL_FAIL (it->types, R_SIGN_TYPES);
 			it->types = zign_types_to_list (a, token);
 			break;
+		case R_SIGN_COLLISIONS:
+			DBL_VAL_FAIL (it->collisions, R_SIGN_COLLISIONS);
+			it->collisions = zign_types_to_list (a, token);
+			break;
 		case R_SIGN_BBHASH:
 			DBL_VAL_FAIL (it->hash, R_SIGN_BBHASH);
 			if (token[0] != 0) {
@@ -499,6 +503,7 @@ static char *serialize_value(RSignItem *it) {
 	FreeRet_on_fail (serialize_str_list (it->xrefs, sb, R_SIGN_XREFS), sb);
 	FreeRet_on_fail (serialize_str_list (it->vars, sb, R_SIGN_VARS), sb);
 	FreeRet_on_fail (serialize_str_list (it->types, sb, R_SIGN_TYPES), sb);
+	FreeRet_on_fail (serialize_str_list (it->collisions, sb, R_SIGN_COLLISIONS), sb);
 
 	if (it->comment) {
 		FreeRet_on_fail (r_strbuf_appendf (sb, "|%c:%s", R_SIGN_COMMENT, it->comment), sb);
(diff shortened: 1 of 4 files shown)


swoops:
Found a bug in r_sign that conflicts with this (pull #18694), so I am going to have to rebase this at least once. May as well fix some stuff while I am at it.

libr/anal/sign.c
```diff
@@ -2277,6 +2332,184 @@ static int match_metrics(RSignItem *it, void *user) {
...
+		return true;
+	}
+	char *dup = strdup (add);
+	if (!dup) {
+		return false;
+	}
+	return r_list_append (l, dup)? true: false;
+}
+
+static bool update_collide(RPVector *sigs, ssize_t start, ssize_t end, int type) {
```
radare, line 2362: ssize_t doesnt exist on windows

libr/anal/sign.c
```diff
@@ -2277,6 +2332,184 @@ static int match_metrics(RSignItem *it, void *user) {
...
+	r_return_val_if_fail (start >= 0 && end > 0 && sigs, false);
+	ssize_t i, ii;
```
radare, line 2364: same

libr/anal/sign.c
```diff
@@ -2277,6 +2332,184 @@ static int match_metrics(RSignItem *it, void *user) {
 	return 0;
 }
 
+static int _sig_to_vec_cb(RSignItem *it, void *user) {
+	if (r_pvector_push ((RPVector *)user, it)) {
+		return 1;
+	} else {
```
radare, line 2338: no need to else after return, but you can also use the ternary operator and make oneliner here

radare (changes requested):
I like the idea but im not sure if those shuold be done inside as another type of zignature. Are you planning to reuse them as hints to reduce the amount of signatures to search for faster analysis for example? if so, maybe it's the way to go. otherwise i would probably prefer `zac` to just print those collisions instead of registering them back as zignature hints

trufae:
Great! Yeah i read your plans for improving matching collisions! Lgtm

libr/anal/sign.c
swoops: I did not know that :). I will fix it.

PR #18701, opened 2021-05-15
Make collision ##signatures more efferent

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Mostly to make tests more deterministic. A sorted list might also be easier to parse.


swoops:
Sorry for second commit, wanted to change the format of the collision signatures quick, before they are used. 

Between the sorted collisions, and a single char type specifier, doing collision logic should be a lot faster.

I updated the title to reflect the last commit.

PR #18704, opened 2021-05-17
Create function when z/ finds byte ##signatures 

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Followup to #18703, this makes `z/` create new functions when a byte signature is found outside of a function. If bytes are the only signature in use, the function will get named. Otherwise the new function will be named by the combination of matched signatures that occurs later.


radare:
Seems like the separate PRs got in conflict

swoops:
that should fix it, same code just rebased to master.

PR #18707, opened 2021-05-18
Remove assertion in pFA ##print

meme:
**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/libr/util/axml.c
+++ b/libr/util/axml.c
@@ -306,7 +306,11 @@ R_API char *r_axml_decode(const ut8 *data, const ut64 data_size) {
 	RStrBuf *sb = NULL;
 	st32 depth = 0;
 
-	r_return_val_if_fail (data && data_size, NULL);
+	if (!data_size) {
+		return NULL;
+	}
+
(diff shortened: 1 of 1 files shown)


libr/util/axml.c
```diff
@@ -306,7 +306,9 @@ R_API char *r_axml_decode(const ut8 *data, const ut64 data_size) {
 	RStrBuf *sb = NULL;
 	st32 depth = 0;
 
-	r_return_val_if_fail (data && data_size, NULL);
```
trufae, line 309: Only check for data_size in here and keep the r_return for the data which never should be null

trufae (changes requested):

libr/util/axml.c
```diff
@@ -306,7 +306,11 @@ R_API char *r_axml_decode(const ut8 *data, const ut64 data_size) {
...
+	if (!data_size) {
+		return NULL;
+	}
+
+	r_return_val_if_fail(data, NULL);
```
trufae, line 313:
```suggestion
	r_return_val_if_fail (data, NULL);
```

PR #18763, opened 2021-05-27
Fix grep cmd with neg

as0ler:
**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/libr/cons/grep.c
+++ b/libr/cons/grep.c
@@ -779,12 +779,9 @@ R_API int r_cons_grep_line(char *buf, int len) {
 				ampfail = 0;
 				continue;
 			}
-			
-			if (grep->begin) {
-				hit = grep->neg? p != in: p == in;	
-			} else {
-				hit = !grep->neg;
-			}
+
+			hit = grep->neg? p != in: p == in;
+
 			// TODO: optimize without strlen without breaking t/feat_grep (grep end)
 			if (grep->end && (strlen (grep->strings[i]) != strlen (p))) {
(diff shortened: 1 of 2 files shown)


libr/cons/grep.c
```diff
@@ -780,7 +780,10 @@ R_API int r_cons_grep_line(char *buf, int len) {
 				continue;
 			}
 			if (grep->begin) {
-				hit = (p == in);
+				hit = (p == in);
```
trufae, line 783: what about `hit = grep->neg? p != in: p == in;`

test/db/cmd/feat_grep
```diff
@@ -67,6 +67,20 @@ size     0x400
 EOF
 RUN
 
+NAME=grep neg begin
```
trufae, line 70: Please add more tests and mark them as broken if they don't pass

trufae (changes requested):

libr/cons/grep.c
```diff
@@ -779,8 +779,9 @@ R_API int r_cons_grep_line(char *buf, int len) {
 				ampfail = 0;
 				continue;
 			}
+			
 			if (grep->begin) {
-				hit = (p == in);
+				hit = grep->neg? p != in: p == in;	
 			} else {
```
trufae, line 785: Remove the if and the else . The ternary operator is enough

trufae (changes requested):
Rebase to make it green

PR #18792, opened 2021-06-02
Fix armhf/armv7 musl compatibility again

nmeum:
In e92d170a3b9fa70e2abacae73e30acfb3dcfc6bd, the include of `asm/ptrace.h` has been silently disabled (it is presently unclear to me why). Thereby basically reverting #13427. However, without `asm/ptrace.h` included the build will fail on musl armhf/armv7 targets with the following compilation failure:

	p/native/linux/linux_coredump.c: In function 'linux_get_arm_vfp_data':
	p/native/linux/linux_coredump.c:950:27: error: 'ARM_VFPREGS_SIZE' undeclared (first use in this function)
	  950 |  char *vfp_data = calloc (ARM_VFPREGS_SIZE + 1, 1);
	      |                           ^~~~~~~~~~~~~~~~

This commit fixes this build failure by including asm/ptrace.h again.

---

CC: @trufae

### DIFF
--- a/libr/debug/p/native/linux/linux_coredump.c
+++ b/libr/debug/p/native/linux/linux_coredump.c
@@ -7,9 +7,13 @@
 #if __x86_64__ || __i386__ || __arm__ || __arm64__
 #include <sys/uio.h>
 #include <sys/ptrace.h>
-// #include <asm/ptrace.h>
 #include "linux_coredump.h"
 #include "linux_ptrace.h"
+
+/* Required for ARM_VFPREGS_SIZE on musl */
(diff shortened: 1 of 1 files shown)


trufae:
This file is not found when building for amd64, so maybe we can wrap it inside some ifdefs, but afaik including asm/ is a bad practice

nmeum:
> This file is not found when building for amd64, so maybe we can wrap it inside some ifdefs […]

Done.

> […] but afaik including asm/ is a bad practice

Yeah, I agree I can't think of an easy alternative solution though.

PR #18499, opened 2021-03-20
Radare2 Version Control System

RHL120:
**Checklist**

- [ ] Closing issues: #issue
- [ ] Mark this if you consider it ready to merge
- [ ] I've added tests (optional)
- [ ] I wrote some lines in the [radare2book](https://github.com/radareorg/radare2book)

**Description**

This is supposed to be a replacement for git in radare2 (not the actual method used for contributing). Not much has been completed yet. I just wrote the nesessary code for initing (so the data types, branching and initing). This system is supposed to be accumulative meanning that patch files will be stored and every time patches reach a certain limit the base is restored.
directory structure:
.rvc
.rvc/branches
.rvc/branches/new
.rvc/branches/new/commits
.rvc/branches/new/commits/be104552aaef802cf59242763ccec7e99d1d94e07a9db66bfd97d534ef3fd0c8
.rvc/branches/master
.rvc/branches/master/commits
.rvc/branches/master/commits/be104552aaef802cf59242763ccec7e99d1d94e07a9db66bfd97d534ef3fd0c8
.rvc/blobs
.rvc/blobs/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
.rvc/blobs/a16d870d5f54fa15d80bc8de90acffcc2aa9c4c201f0fa5f09e4880887992494
Commit structure:

author:RHL120
message:Test
timestamp:1618570635
\----
blob:1:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
blob:2:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
blob:3:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
blob:4:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
blob:5:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
blob:6:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
blob:7:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
blob:8:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
blob:9:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
[...]

PR #18794, opened 2021-06-03
WIP: use sdb in rvc

RHL120:
**Checklist**

- [ ] Closing issues: #issue
**Description**

In https://github.com/radareorg/radare2/pull/18499 I helped introduce the abomination known as rvc. This pr tries to make rvc use sdb for branches instead of having them as dirs. This has many advantages most notably:
* Rvc does no longer copy & dup commits when branching since commit don't hold info about their position
* Rvc no longer has to have the "loading_functionality" since needed data is loaded on command
a disadvantage would be that now Rvc would have to load important data every time an R_API function is called
I would recommend the revs to focuse on the sdb usage because it is probably not valid.
TODOs:
* Add some proper error messages
* Add r_vc_checkout functionality
* Add some checks to know whether or not an rvc_repo exists before executing a command
Far TODOs:
* Unify The git APIS and RVC apis based on prj.vc
* Add reverting functionality

### DIFF
--- a/binr/rvc2/rvc2.c
+++ b/binr/rvc2/rvc2.c
@@ -1,7 +1,6 @@
 /* radare - LGPL - Copyright 2021 - pancake */
-
 #include <r_main.h>
 
 int main (int argc, const char *argv[]) {
-	return r_main_rvc2 (argc, argv);
+	//return r_main_rvc2 (argc, argv);
 }
--- a/binr/rvc2/rvc2.c
+++ b/binr/rvc2/rvc2.c
@@ -2,5 +2,5 @@
 #include <r_main.h>
 
 int main (int argc, const char *argv[]) {
-	return r_main_rvc2 (argc, argv);
+	//return r_main_rvc2 (argc, argv);
 }
--- a/binr/rvc2/rvc2.c
+++ b/binr/rvc2/rvc2.c
@@ -2,5 +2,5 @@
 #include <r_main.h>
 
 int main (int argc, const char *argv[]) {
-	//return r_main_rvc2 (argc, argv);
+	return r_main_rvc2 (argc, argv);
 }
--- a/libr/core/rvc.c
+++ b/libr/core/rvc.c
@@ -66,6 +66,32 @@ static char *rp2absp(const char *rp, const char *path) {
 	return r_str_appendf (arp, R_SYS_DIR "%s", path);
 }
 
+static RList *get_commits(Sdb *db, const size_t max_num) {
+	char *i;
+	RList *ret = r_list_new ();
+	if (!ret) {
+		return NULL;
+	}
+	i = sdb_get (db, sdb_const_get (db, "current_branch", 0), 0);
+	while (true) {
+		if (!r_list_prepend (ret, i)) {
+			r_list_free (ret);
+			break;
+		}
+		i = sdb_get (db, ret->tail->data, 0);
+		if (!i || !*i) {
+			r_list_free (ret);
+			ret = NULL;
+			break;
+		}
+		if ((max_num && ret->length >= max_num) || !*i) {
+			ret = NULL;
+			break;
+		}
+	}
+	return ret;
+}
+
 static bool bfadd(const char *rp, RList *dst, const char *path) {
 	RvcBlob *blob;
 	char *absp;
@@ -173,6 +199,7 @@ static RList *blobs_add(const char *rp, const RList *paths) {
 	}
 	return ret;
 }
+
 
 static char *write_commit(const char *rp, const char *message, const char *author, RList *blobs) {
 	RvcBlob *blob;
--- a/libr/core/rvc.c
+++ b/libr/core/rvc.c
@@ -72,7 +72,7 @@ static bool bfadd(RList *dst, const char *path, const char *rp) {
 		return false;
 	}
 	blob->fhash = sha256_file (absp);
-	if (blob->fhash) {
+	if (!blob->fhash) {
 		free (absp);
 		free (blob->fname);
 		free (blob);
@@ -294,8 +294,8 @@ R_API bool r_vc_new(const char *path) {
 	if (!vcp) {
 		return false;
 	}
-	commitp = r_str_newf ("%s" R_SYS_DIR "commits", path);
-	blobsp = r_str_newf ("%s" R_SYS_DIR "blobs", path);
+	commitp = r_str_newf ("%s" R_SYS_DIR "commits", vcp);
+	blobsp = r_str_newf ("%s" R_SYS_DIR "blobs", vcp);
 	if (!commitp || !blobsp) {
 		free (commitp);
 		free (blobsp);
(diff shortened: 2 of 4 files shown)


binr/rvc2/rvc2.c
```diff
@@ -1,7 +1,6 @@
 /* radare - LGPL - Copyright 2021 - pancake */
-
 #include <r_main.h>
 
 int main (int argc, const char *argv[]) {
-	return r_main_rvc2 (argc, argv);
+	//return r_main_rvc2 (argc, argv);
```
trufae, line 5:
```suggestion
	return r_main_rvc2 (argc, argv);
```

libr/include/rvc.h
```diff
@@ -18,6 +18,7 @@ R_API int r_vc_git_add(const char *path, const char *fname);
 R_API int r_vc_git_commit(const char *path, const char *message);
 
 R_API bool r_vc_commit(const char *rp, const char *message, const char *auth,
```
trufae, line 22: that fits in one line, function signatures should never be wrapped in multilines for grepping purposes

trufae (changes requested):

libr/core/rvc.c
```diff
@@ -66,491 +32,306 @@ static inline char *sha256_file(const char *fname) {
...
-	r_list_foreach (commit->blobs, iter, blob) {
-		char *tmp = r_str_appendf (commit_string, "\nblob:%s:%s",
-				blob->fname, blob->hash);
-		if (!tmp) {
-			return false;
-		}
-		commit_string = tmp;
+	char *absp;
+	char *blob_path;
+	blob = R_NEW (RvcBlob);
```
trufae, line 61: always define and assign in the same line if possible

libr/core/rvc.c
```diff
@@ -66,491 +32,306 @@ static inline char *sha256_file(const char *fname) {
...
+		free (blob);
 		return false;
 	}
-	commit_path = r_str_newf ("%s" R_SYS_DIR "branches" R_SYS_DIR
-			"%s" R_SYS_DIR "commits" R_SYS_DIR "%s",
-			repo->path, b->name, commit->hash);
-	if (!commit_path) {
-		free (commit_string);
+	blob->fname = absp2rp (path, rp);
+	if (!blob->fname) {
```
trufae, line 71: better use goto and have a single exit point with all the frees in there

libr/main/rvc2.c
```diff
@@ -1,6 +1,6 @@
 /* radare - LGPL - Copyright 2021 - pancake */
 #include <rvc.h>
-
+/*
```
trufae, line 3: why

libr/include/rvc.h
```diff
@@ -3,55 +3,26 @@
...
-R_API RList *r_vc_add(Rvc *repo, RList *files);
-R_API Rvc *r_vc_new(const char *path);
-R_API void r_vc_free(Rvc *vc);
 R_API int r_vc_git_init(const char *path);
 R_API bool r_vc_git_branch(const char *path, const char *name);
 R_API bool r_vc_git_checkout(const char *path, const char *name);
 R_API int r_vc_git_add(const char *path, const char *fname);
 R_API int r_vc_git_commit(const char *path, const char *message);
 
+R_API bool r_vc_commit(const char *rp, const char *message, const char *auth, RList *files);
```
trufae, line 22:
```suggestion
R_API bool r_vc_commit(const char *rp, const char *message, const char *author, RList *files);
```

libr/include/rvc.h
```diff
@@ -3,55 +3,26 @@
...
 extern "C" {
 #endif
 
-typedef struct blob {
-	char *fname;
-	char *hash;
+#include <r_core.h>
+#include <sdb.h>
+typedef struct RvcBlob_t {
+	char *fname, *fhash;
```
trufae, line 14: on structs, dont use the comma, one line per field makes them more readable

libr/core/rvc.c
```diff
@@ -66,491 +32,306 @@ static inline char *sha256_file(const char *fname) {
...
+			return false;
+		}
+		if (sdb_open (db, dbf) < 0) {
+			free_blobs (blobs);
+			free (commit_hash);
 			return false;
 		}
-		free (ppath);
+		free (dbf);
+		current_branch = sdb_get (db, "current_branch", 0);
```
trufae, line 225: you can use `sdb_const_get` and avoid the free later. if you dont modify the value it will save a dupped string in memory

libr/core/rvc.c
```diff
@@ -66,491 +32,306 @@ static inline char *sha256_file(const char *fname) {
...
+		if (!current_branch) {
+			free_blobs (blobs);
+			free (commit_hash);
+			sdb_free (db);
+			return false;
+		}
+		if (!sdb_array_push (db, current_branch, commit_hash, 0)) {
```
trufae, line 232: dont use sdb_array for this, its unnefficient (O(n)) and will not work the way we discussed.

trufae (changes requested):

libr/core/rvc.c
RHL120: I don't know whether or not that is wise. You see it would work now but it might risk double frees in the fututre. What do you think though

libr/core/rvc.c
RHL120: when is is it possible?

libr/core/rvc.c
RHL120: what should I use instead?

trufae (changes requested):
.

libr/main/rvc2.c
```diff
@@ -40,55 +40,93 @@ R_API int r_main_rvc2(int argc, const char **argv) {
...
+			return -8;
+		}
+		files = r_list_new ();
+		if (!files) {
+			free (rp);
+			free (auth);
+			free (message);
+			return -9;
+		}
+		for (i = 3; i < argc - 1; ++i) {
```
trufae, line 93:
```suggestion
		for (i = 3; i < argc - 1; i++) {
```

libr/main/rvc2.c
```diff
@@ -40,55 +40,93 @@ R_API int r_main_rvc2(int argc, const char **argv) {
...
+			return -4;
+		}
+		if (!r_vc_branch (rp, opt.argv[opt.ind + 1])) {
+			return -5;
+		}
+		return 0;
+	}
+	if (!strcmp (action, "commit")) {
+		char *auth, *message;
+		uint i;
```
trufae, line 69:
```suggestion
		int i;
```

libr/core/rvc.c
trufae: i mean `RvcBlob *blob = R_NEW0 (RvcBlob);`

radare:
Thanks! That become a large PR already, lets pile the next changes after merging this! Good work

Issue #18799, opened 2021-06-03
r2pipe commands containing @e aka temporary eval vars print the result directly to stdout bypassing R2PIPE_IN


ab1030:
## Description

I open an r2pipe to bash then I issue my radare2 commands and receive the expected results until I start mixing my commands with temporary eval vars using `@e:` then I stop receiving the results through `R2PIPE_IN` but instead everything is printed directly to `stdout`.

## Test

```bash
$ r2 -
-- ...
[0x00000000]> #!pipe bash
bash$ # This is the expected
bash$ r2p 'pd 1' | wc -c
256
bash$ # This is a bug
bash$ r2p 'pd 1 @e:asm.hints=false' | wc -c
            0x00000000      0000           add byte [rax], al
0
```

PR #18800, opened 2021-06-04
Fix issue #18799 - r2pipe commd output leaks to stdout

ab1030:
**Checklist**

- [x] Closing issues: #18799

**Description**

When r2pipe commands are executed the result is not supposed to be printed to the console but instead it should be returned as a string to r2pipe, while this works in general but there are some command combination that cause some leakage. This patch ensures the console isn't flushed until r_core_cmd_str is finished including the case when r_core_cmd_str is reentrant.

### DIFF
--- a/test/unit/test_cmd_str.c
+++ b/test/unit/test_cmd_str.c
@@ -4,7 +4,7 @@
 bool test_cmd_str_issue_18799() {
 	RCore *core = r_core_new ();
 	char *output = r_core_cmd_str (core, "pd 1 @e:asm.hints=false");
-	mu_assert ("command output leaked to stdout", strlen(output) > 0);
+	mu_assert ("command output leaked to stdout", strlen (output) > 0);
 	r_core_free (core);
 	mu_end;
 }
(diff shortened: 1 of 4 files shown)


test/unit/test_cmd_str.c
```diff
@@ -0,0 +1,20 @@
+#include <r_core.h>
+#include "minunit.h"
+
+bool test_cmd_str_issue_18799() {
+	RCore *core = r_core_new ();
+	char *output = r_core_cmd_str (core, "pd 1 @e:asm.hints=false");
+	mu_assert ("command output leaked to stdout", strlen(output) > 0);
```
trufae, line 7:
```suggestion
	mu_assert ("command output leaked to stdout", strlen (output) > 0);
```

trufae:
I had some feelings about a bug like this would happen after the latest refactorings, but didnt broke any test, thanks for submitting the fix and the test!

trufae:
Now i remember why this code changed several times to fix and regress the following issue: visual debugger is not working after merging your patch, you can try this with `r2 -cVpp -d r2`

trufae:
I fixed it by using r_core_cmd_str instead in https://github.com/radareorg/radare2/pull/18806

PR #18802, opened 2021-06-04
Add doclean target into libr/util/d/Makefile ##build

jmaselbas:
**Checklist**
- [x] Not closing issues
- [x] Not adding tests (optional)
- [x] Not adding documentation

**Description**

radare2 failed to build without using `--without-gperf` option,
with the following message: No rule to make target 'd/ascii.o',
needed by 'libr_util.so'.

I took a look into anal/d/Makefile which seems to be doing the same
thing. I noticed that it was using `all` as default target instead
of `alle` as done in util/d/Makefile. Moreover the .PHONY only mention
`all` an not `alle`, thus I believe this is a typo, and it was meant
to be `all`, fix this.

This could also be fixed by changing the `all` to `alle` target in `.PHONY`. not sure if it is a better fix.
---
edit: Maybe the issue I was having is only related to a dirty build directory.
The issue I was having can no longer be reproduced after cleaning the `libr/util/d/` directory

This commit now add a clean target to  `libr/util/d/Makefile`

### DIFF
--- a/libr/util/d/Makefile
+++ b/libr/util/d/Makefile
@@ -15,6 +15,9 @@ alle: $(F_SDB)
 	$(SDB) -t -C $@
 	test -f $@
 
+clean:
+	rm -f *.sdb *.c *.gperf
+
 install:
 	mkdir -p "$(R2_CHARSETS_PATH)"
 	for FILE in *.sdb ; do \
@@ -30,4 +33,4 @@ symstall install-symlink:
 		ln -fs "${CWD}/$$FILE" "${R2_CHARSETS_PATH}/$$FILE" ; \
 	done
 
-.PHONY: all install symstall
+.PHONY: alle clean uninstall install symstall install-symlink
--- a/libr/util/Makefile
+++ b/libr/util/Makefile
@@ -73,6 +73,7 @@ endif
 
 EXTRA_PRE+=spp_config
 EXTRA_PRE+=charsets
+EXTRA_CLEAN=doclean
 
 include ../rules.mk
 include sdb.mk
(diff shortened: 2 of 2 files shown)


trufae:
Which version of GNU Make are you using to not run the first target defined in a makefile? the name shouldnt matter much, the reason to use `alle` was to avoid overriding the `all` one that was taken from an external .mk.. i think I end up removing that .mk from the dependencies so using `all:` should be good now. Curious that this wasnt happening to any of my machines or in the CI

jmaselbas:
I am using `GNU Make 4.1` but the strange thing is that if make the `alle` target in `.PHONY` it works better.

jmaselbas:
also I don't thinks all the files are removed when calling the clean target, so I am running: `rm libr/util/d/*.c libr/util/d/*.d libr/util/d/*.sdb libr/util/d/*.o` between build (to test if the change actually fix the issue or not)

jmaselbas:
How strange,
it seems to work properly now, after cleaning all the intermediate files with `rm libr/util/d/*.c libr/util/d/*.d libr/util/d/*.sdb libr/util/d/*.o`.
This could explain why this works fine in CI but not if the build dir is dirty

trufae:
I do `git clean -xdf` or run the `meson newbuilddir`, could you extend this PR improving the clean to remove those stale files?

libr/util/d/Makefile
```diff
@@ -15,6 +15,9 @@ alle: $(F_SDB)
 	$(SDB) -t -C $@
 	test -f $@
 
+clean:
+	rm -f *.sdb *.c
```
trufae, line 19:
```suggestion
	rm -f *.sdb *.c *.gperf
```

trufae (changes requested):
Maybe worth checking the other libr/*/d that do the same, but i guess the issue was only in util.

thanks!

PR #18840, opened 2021-06-17
io.plugin.lseek -> .seek for portability (wasi related)

trufae:
This will break API compat for all io plugins


ilyasELBANI:
This breaks a lot of stuff, I think it should be documented somewhere? or even keep both lseek and seek at the same time?

trufae:
I updated all the io plugins i use. You can use #if directive to use the old and the new. Its a change required to build with the webassembly toolchain where lseek is a define and actually seek is a better name. This change will be documented in the next release

PR #18853, opened 2021-06-21
Fix #18852 wrong tables in graphviz dot ##print

### LINKED ISSUE #18852: Wrong jump tables edges when printing graphviz dot
## Environment

```sh
Mon Jun 21 17:17:07 JST 2021
radare2 5.4.0-git 26488 @ darwin-arm-64 git.5.3.1
commit: 6f05d09713fb83926a1f76140dcc7ae748626b25 build: 2021-06-21__17:16:00
Darwin arm64
```

## Description

When printing the Basic block function graph (`agf`) in form of Graphviz dot, every jump table is drawn incorrectly.
In particular, the targets of a jump table are represented correctly with the `VV` command. However, when exporting the graphviz dot representation, the edges "jump table" -> "jump target" are missing. The missing edges are replaced by a "jump target" -> "jump target" edge.
Surprisingly, the GML output does not present this problem.

Correct representation with `VV`
Screen Shot 2021-06-21 at 17 24 15

Incorrect representation with the exported graphviz dot (later processed with the dot command)
Note the black arrows
Screen Shot 2021-06-21 at 16 56 28

Expected output with the exported graphviz dot
Screen Shot 2021-06-21 at 16 56 10

## Test

I used this binary file for testing, but any jump table exhibit this problem.
[ls.zip](https://github.com/radareorg/radare2/files/6685177/ls.zip)

Then this command:
`r2 -qAc "s 0xA1A0;agfdm" ls`

The wrong lines are the one containing the `[color2="#c50f1f"];` part.

davidepi:
**Checklist**

- [x] Closing issues: #18852

**Description**

Details on the bug can be found in Issue #18852.
Turns out that when processing a graph for export in several formats (including graphviz), when processing a jump table, an edge `caseop->addr` -> `caseop->jump` is added for each case of the jump table. However debugging the `caseop` struct shows that these fields contain the same value, hence the reason of the self-looping edge. 
Given that the GML implementation was only slightly different and did not present the bug, I replaced `caseop->addr` and `caseop->jump` with the values used in that implementation.

Additionally:
- replaced `color2` with `color`. color2 is not a valid graphviz keyword and is ignored (tested with dot 2.47.3).
- changed the color to be consistent with all the other representations.

Formatting looks a bit weird to me, but this is the output of clang-format.py

### DIFF
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -1561,19 +1561,20 @@ static int core_anal_graph_construct_edges(RCore *core, RAnalFunction *fcn, int 
 							"bb.0x%08"PFMT64x".switch", bbi->addr);
 							sdb_array_add_num (DB, key, caseop->value, 0);
 				} else if (is_html) {
-					r_cons_printf ("<div class=\"connector _0x%08"PFMT64x" _0x%08"PFMT64x"\">\n"
+					r_cons_printf ("<div class=\"connector _0x%08" PFMT64x " _0x%08" PFMT64x "\">\n"
 							"  <img class=\"connector-end\" src=\"img/arrow.gif\"/></div>\n",
-							caseop->addr, caseop->jump);
+							bbi->addr, caseop->addr);
 				} else if (!is_json && !is_keva) {
 					if (is_star) {
-						char *from = get_title (caseop->addr);
-						char *to = get_title (caseop->jump);
+						char *from = get_title (bbi->addr);
+						char *to = get_title (caseop->addr);
 						r_cons_printf ("age %s %s\n", from ,to);
-						free(from);
-						free(to);
+						free (from);
+						free (to);
 					} else {
-						r_cons_printf ("        \"0x%08"PFMT64x"\" -> \"0x%08"PFMT64x"\" " \
-						"[color2=\"%s\"];\n", caseop->addr, caseop->jump, pal_fail);
+						r_cons_printf ("        \"0x%08" PFMT64x "\" -> \"0x%08" PFMT64x "\" "
+								"[color=\"%s\"];\n",
+								bbi->addr, caseop->addr, pal_trfa);
 						core_anal_color_curr_node (core, bbi);
 					}
 				}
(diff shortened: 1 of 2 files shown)


trufae:
May be worth having a test for this, but im happy to merge as it seems to fix the jump table issue. thanks!

libr/core/canal.c
```diff
@@ -1561,19 +1561,20 @@ static int core_anal_graph_construct_edges(RCore *core, RAnalFunction *fcn, int
 							"bb.0x%08"PFMT64x".switch", bbi->addr);
 							sdb_array_add_num (DB, key, caseop->value, 0);
 				} else if (is_html) {
-					r_cons_printf ("<div class=\"connector _0x%08"PFMT64x" _0x%08"PFMT64x"\">\n"
-							"  <img class=\"connector-end\" src=\"img/arrow.gif\"/></div>\n",
-							caseop->addr, caseop->jump);
+					r_cons_printf ("<div class=\"connector _0x%08" PFMT64x " _0x%08" PFMT64x "\">\n"
+						       "  <img class=\"connector-end\" src=\"img/arrow.gif\"/></div>\n",
```
trufae, line 1565: dont indent with spaces

libr/core/canal.c
```diff
@@ -1561,19 +1561,20 @@ static int core_anal_graph_construct_edges(RCore *core, RAnalFunction *fcn, int
...
+						char *from = get_title (bbi->addr);
+						char *to = get_title (caseop->addr);
 						r_cons_printf ("age %s %s\n", from ,to);
 						free(from);
 						free(to);
 					} else {
-						r_cons_printf ("        \"0x%08"PFMT64x"\" -> \"0x%08"PFMT64x"\" " \
-						"[color2=\"%s\"];\n", caseop->addr, caseop->jump, pal_fail);
+						r_cons_printf ("        \"0x%08" PFMT64x "\" -> \"0x%08" PFMT64x "\" "
+							       "[color=\"%s\"];\n",
```
trufae, line 1576: dont indent with spaces
```suggestion
							"[color=\"%s\"];\n",
```

davidepi:
I agree. I'm going to look how the tests are structured in this repository and add one.
Going to update this PR asap

libr/core/canal.c
```diff
@@ -1561,19 +1561,20 @@ static int core_anal_graph_construct_edges(RCore *core, RAnalFunction *fcn, int
...
+						       "  <img class=\"connector-end\" src=\"img/arrow.gif\"/></div>\n",
+						bbi->addr, caseop->addr);
 				} else if (!is_json && !is_keva) {
 					if (is_star) {
-						char *from = get_title (caseop->addr);
-						char *to = get_title (caseop->jump);
+						char *from = get_title (bbi->addr);
+						char *to = get_title (caseop->addr);
 						r_cons_printf ("age %s %s\n", from ,to);
 						free(from);
```
trufae, line 1572:
```suggestion
						free (from);
```

libr/core/canal.c
```diff
@@ -1561,19 +1561,20 @@ static int core_anal_graph_construct_edges(RCore *core, RAnalFunction *fcn, int
...
 						free(to);
```
trufae, line 1573:
```suggestion
						free (to);
```

libr/core/canal.c
trufae, line 1565:
```suggestion
							"  <img class=\"connector-end\" src=\"img/arrow.gif\"/></div>\n",
```

trufae:
there's no actual C indent tool that works well, they all need post processing, for 3 lines of patch you can just indent by hand and shouldnt be much of a hassle. feel free to apply the suggested changes. thanks!

libr/core/canal.c
davidepi: oh, sorry. first PR here and I assumed clang-format.py was going to fix everything and didn't double check. Going to fix!

trufae:
To add a test edit any file under test/db/ and add the commands you want to test by copypasting another one, then just run r2r -i (path-to-file) and press 'f' to fix the output with the new one. try to find a small or fast reproducer to avoid having slow test times.

thanks!

davidepi:
Added a test and asserted that it fails in the current master and passes with this fix.
I handcrafted the assembly so there should be no problem of slow test times

trufae:
Beautiful!

PR #18854, opened 2021-06-21
Completely eliminate the deprecated backslash command ##shell

trufae:


trufae:
Everyone seems to agree. lets kill that

PR #18878, opened 2021-07-01
Update arm.sdb.txt file with opcodes

alex-bender:
**Checklist**

**Description**

Add some opcodes generated from https://developer.arm.com/documentation/ddi0596/2021-03/Base-Instructions?lang=en
with help of
```
let doc = $('[role=document]');
var x = document.createElement("TEXTAREA");
document.body.appendChild(x);

for (var i=1; i<doc.childElementCount; i++) {
  var el = doc.children[i];
  if (el.nodeName == 'DIV') {
    x.value += ('\n' + doc.children[i].textContent);
  }
}
```
and 
```
data = open('./arm.txt').readlines()
result = []
def filter(line):
    line = line.replace(' (immediate)', '')
    line = line.replace(' (register)', '')
    line = line.replace(' (shifted register)', '')
    line = line.replace(' (extended register)', '')
    line = line.replace(' (bitmask immediate)', '')
    coma = line.find(',')
    dots = line.find(':')
    part = line[:dots]
    if ',' in part and coma < dots:

        opcodes = part.split(', ')
        descr = line[dots+1:].strip()
        for opcode in opcodes:
            res = '%s=%s' % (opcode.lower(), descr[:-1].lower())
            result.append(res)
            print(res)
    else:
        opcode = line.split(':')[0]
        descr = line[dots+1:].strip()

    res = '%s=%s' % (opcode.lower(), descr[:-1].lower())
    print(res)
    result.append(res)
```


trufae:
Maybe worth having this script somewhere.. maybe in extras?

libr/asm/d/arm.sdb.txt
```diff
@@ -1,167 +1,475 @@
...
+brab=Branch to Register, with pointer authentication
+brabz=Branch to Register, with pointer authentication
+brk=Breakpoint instruction
+bti=Branch Target Identification
+cas=Compare and Swap word or doubleword in memory
+casa=Compare and Swap word or doubleword in memory
+casab=Compare and Swap byte in memory
+casah=Compare and Swap halfword in memory
+casal=Compare and Swap word or doubleword in memory
+casalb=Compare and Swap byte in memory
```
radare, line 55: can you uncapitalize all the words in those strings? lowercase is fine

libr/asm/d/arm.sdb.txt
```diff
@@ -1,167 +1,475 @@
...
-stp=store register pairs in memory
-ldp=load a pair of registers from memory
-addlo=add if C is clear (no carry)
-mvn=move bitwise NOT
-mvns=move bitwise NOT and update flags
-blo=variant of bcc
-bhs=variant of bcs
+adc=Add with Carry
+adcs=Add with Carry, setting flags
+add=Add
```
radare, line 3: not very descriptive. old text was clearer in this case
Screenshot 2021-07-02 at 11 50 46

radare (changes requested):
did you checked if any of the old entries is not available in the new autogenerated format? this file should contain arm, thumb and arm64 instruction descriptions.

radare:
Ping

alex-bender:
> did you checked if any of the old entries
Looks like I've forgot to do that.
Will check it today and add the script for that check as well

alex-bender:
> did you checked if any of the old entries is not available in the new autogenerated format?

I've added only opcodes which was not in file already

PR #18889, opened 2021-07-06
Rework the rvc_commit functions ##projects

RHL120:
**Checklist**

- [ ] Closing issues: #issue

**Description**
This PR is supposed to make the rvc add apis abit more elegent and better suitable for directory adding.

### DIFF
--- a/libr/core/rvc.c
+++ b/libr/core/rvc.c
@@ -9,6 +9,24 @@
 #define BPREFIX "branches."
 #define NULLVAL "-"
 
+static char *strip_sys_dir(const char *path) {
+	char *ret = r_str_new ("");
+	if (!ret)  {
+		return NULL;
+	}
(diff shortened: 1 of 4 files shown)


libr/core/rvc.c
```diff
@@ -9,6 +9,24 @@
 #define BPREFIX "branches."
 #define NULLVAL "-"
 
+static char *strip_sys_dir(const char *path) {
```
RHL120, line 12: The way this function works is inefficient, a more inefficient way to do it is with strchr but this method wouldn't be as elegent.

trufae: Ready to merge then?

RHL120: Well I mean maybe it depends on wheter or not you tolarate ugly code or inefficient code

PR #18915, opened 2021-07-18
Use r_sys_whoami as the author name for r_vc_commit ##projects

RHL120:
**Checklist**

- [ ] Closing issues: #issue

**Description**
r_vc_commit now checks if author is NULL in case it is it sets author to the value of prj.vc.author. prj.vc.author is by default set to the return value of r_sys_whoami().

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -4024,6 +4024,8 @@ R_API int r_core_config_init(RCore *core) {
 	/* RVC */
 	{
 		char *p = r_file_path ("git");
+		SETBPREF ("prj.vc.author", r_sys_whoami(),
+				"What author name to use for rvc");
 		if (strcmp (p, "git")) {
 			SETCB ("prj.vc.type", "git", &cb_prjvctype, "What should projects use as a vc");
 		} else {
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -4024,7 +4024,7 @@ R_API int r_core_config_init(RCore *core) {
 	/* RVC */
 	{
 		char *p = r_file_path ("git");
-		SETBPREF ("prj.vc.user", r_sys_whoami(),
+		SETBPREF ("prj.vc.user", r_sys_whoami (),
 				"What author name to use for rvc");
 		if (strcmp (p, "git")) {
 			SETCB ("prj.vc.type", "git", &cb_prjvctype, "What should projects use as a vc");
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -4024,8 +4024,6 @@ R_API int r_core_config_init(RCore *core) {
 	/* RVC */
 	{
 		char *p = r_file_path ("git");
-		SETBPREF ("prj.vc.user", r_sys_whoami (),
-				"What author name to use for rvc");
 		if (strcmp (p, "git")) {
 			SETCB ("prj.vc.type", "git", &cb_prjvctype, "What should projects use as a vc");
 		} else {
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -4024,7 +4024,7 @@ R_API int r_core_config_init(RCore *core) {
 	/* RVC */
 	{
 		char *p = r_file_path ("git");
-		SETBPREF ("prj.vc.author", r_sys_whoami(),
+		SETBPREF ("prj.vc.user", r_sys_whoami(),
 				"What author name to use for rvc");
 		if (strcmp (p, "git")) {
 			SETCB ("prj.vc.type", "git", &cb_prjvctype, "What should projects use as a vc");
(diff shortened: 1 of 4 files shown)


libr/core/rvc.c
```diff
@@ -713,6 +713,7 @@ R_API bool r_vc_commit(const char *rp, const char *message, const char *author,
 		eprintf ("Can't commit");
 		return false;
 	}
+	author = author? author : r_config_get (core->config, "prj.vc.author");
```
trufae, line 716: Instead of calling config and core here just do this in the caller. So rvc stays independent

libr/core/rvc.c
```diff
@@ -1121,6 +1122,8 @@ R_API int r_vc_git_add(const char *path, const char *fname) {
 	return ret;
 }
 
+//Access both git and rvc functionality from one set of functions
+
 R_API int r_vc_git_commit(const char *path, const char *message) {
 	return message ? r_sys_cmdf ("git -C %s commit -m %s", path, message) :
```
trufae, line 1128: Use quotes around “” and escape the strings passed

libr/core/rvc.c
```diff
@@ -1135,7 +1138,7 @@ R_API int rvc_git_init(RCore *core, const char *rp) {
 
 R_API int rvc_git_commit(RCore *core, const char *rp, const char *message, const char *author, const RList *files) {
 	if (!strcmp (r_config_get (core->config, "prj.vc.type"), "rvc")) {
-		r_vc_commit (rp, message, author, files);
```
trufae, line 1138: Get vc.user here instead

trufae (changes requested):

libr/core/rvc.c
```diff
@@ -1090,15 +1090,17 @@ R_API bool r_vc_checkout(const char *rp, const char *bname) {
 // GIT commands as APIs
 
 R_API int r_vc_git_init(const char *path) {
-	return r_sys_cmdf ("git init %s", path);
+	return r_sys_cmdf ("git init \"%s\"", r_str_escape (path));
```
trufae, line 1093: Leaks

libr/core/rvc.c
```diff
@@ -1090,15 +1090,17 @@ R_API bool r_vc_checkout(const char *rp, const char *bname) {
...
 
 R_API bool r_vc_git_branch(const char *path, const char *name) {
-	return !r_sys_cmdf ("git -C %s branch %s", path, name);
+	return !r_sys_cmdf ("git -C \"%s\" branch \"%s\"",
+			r_str_escape (path), r_str_escape (name));
 }
 
 R_API bool r_vc_git_checkout(const char *path, const char *name) {
-	return !r_sys_cmdf ("git -C %s checkout %s", path, name);
+	return !r_sys_cmdf ("git -C \"%s\" checkout \"%s\"", r_str_escape (path),
```
trufae, line 1102: Leaks too

libr/core/cconfig.c
```diff
@@ -4024,6 +4024,8 @@ R_API int r_core_config_init(RCore *core) {
 	/* RVC */
 	{
 		char *p = r_file_path ("git");
+		SETBPREF ("prj.vc.user", r_sys_whoami(),
```
trufae, line 4027:
```suggestion
		SETBPREF ("prj.vc.user", r_sys_whoami (),
```

trufae (changes requested):

trufae:
What about reusing cfg.user?

RHL120:
> What about reusing cfg.user?

Well, I don't know one one hand it is good because it doesn't add a new (possibly pointless) variable but on the other hand it might discourage the user from configuring it because this is a variable used not just by rvc but by r2 in it's entirety so to the user changing it might cause some unintended consequences. I already committed tell me if you want me to revert.

trufae:
Yeah i have divided opinion here but i would probably pick cfg.user because we want to be in sync with r2.. but for me cfg.user contains username and vc.author should be name + mail

RHL120:
> Yeah i have divided opinion here but i would probably pick cfg.user because we want to be in sync with r2.. but for me cfg.user contains username and vc.author should be name + mail

Just to clarify are you saying you want there to be a vc.author.name & a vc.author.mail? rvc has no use for the users email and adding emails to commit would only complicate the code because either we force the user to set an email or we end up having to deal with empty values which is doable but not pleasent.

RHL120:
> Yeah i have divided opinion here but i would probably pick cfg.user because we want to be in sync with r2.. but for me cfg.user contains username and vc.author should be name + mail

Maybe we should just force the user to use their username so just r_sys_whoami without setting up any vars this would remove the need for core->config and thus make the code faster but on the other hand it will disable the customization abilities which in my opinion are already dead if we were to use cfg.user (that's because of possible side-effects)

Issue #16178, opened 2020-03-10
Function vs Fcn


radare:
We decided to go for r_anal_function insteadof r_anal_fcn... so let's deprecate all the r_anal_fcn* apis

PR #18917, opened 2021-07-18
Changed r_anal_fcn call to r_anal_function in support of depreciation ##anal

Royos90:
Changed found r_anal_fcn api calls to r_anal_function to reflect new changes.

**Checklist**

- [x] Closing issues: #16178 

**Description**

The changes done were to replace the r_anal_fcn with r_anal_function. This is due to the r_anal_fcn being depreciated.

### DIFF
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -1497,7 +1497,7 @@ R_API void r_anal_del_jmprefs(RAnal *anal, RAnalFunction *fcn) {
 }
 
 /* Does NOT invalidate read-ahead cache. */
-R_API int r_anal_fcn(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 len, int reftype) {
+R_API int r_anal_function(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 len, int reftype) {
 	RPVector *metas = r_meta_get_all_in(anal, addr, R_META_TYPE_ANY);
 	void **it;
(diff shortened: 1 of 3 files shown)


trufae:
The problem i see in this renaming is that it conflicts with the oo-style needed for the bindings to work. this is:

RAnalFunction -> r_anal_function, one is a struct, and the other a function. so.. as long as its a method for RAnal. there must be a verb before the word `function`. Like: `r_anal_create_function` etc.. 

But as long as there's still some multiple ways to create and register a function this is something that should be addressed.

Any suggestions ?

radare:
Well. Considering this syntax is in sync with r_anal_op then its fine to merge, and the vapi files may be updated anyway, so merging! thanks

Royos90:
Apologies for the delayed response. I wasn't originally sure what the best suggestion is for addressing the multiple ways "r anal syntax" is used. I want to say that the current implementation is best left as is for now. I don't have enough experience working on this repo to really have a definite opinion just yet.

Royos90:
I do like the r_anal_ <verb> syntax but I don't see a way to merge that with the already existing RAnal syntax. I understand that a decision was made to change the syntax but is this for readability of the code or simply to make it easier to use functions going forward?

trufae:
The main problem is deciding best names without picking the long ones. And in order to make those apis consistent feom other languages like vala, its good to follow some patterns like verb_element, getter/setter etc

PR #18919, opened 2021-07-19
fix afj command and afj command extension

lasek0:
fix invalid argv read for table param and others,
fix memory leak by missing free for strdup,
fix invalid elem size and elem count for jmtbl which previously was set to same value, now it can be set properly
extend afj command by adding 2 new parameters elem_sz and seg for precise jmp tbl reading.

According to our IRC conversation about afj naming it is not changed yet. Great subject for discussion about analyzing function jumptable command name.

**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -352,7 +352,7 @@ static const char *help_msg_af[] = {
 	"afd", "[addr]","show function + delta for given offset",
 	"afF", "[1|0|]", "fold/unfold/toggle",
 	"afi", " [addr|fcn.name]", "show function(s) information (verbose afl)",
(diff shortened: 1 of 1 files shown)


trufae:
The change looks good! But i have some questions/suggestions:

- maybe better to rename to something not ending with j because it implies json. What about at?
- is r_anal_jmptbl used anywhere?
- can you add a test using the new arguments for it so we can ensure nothing breaks over time.


If you have a shareable bin for testing your usecase push it to the testbins repo

trufae:
Anyway the pr is good, so just merging for now, if you have some time it would be great to have a test for it. and i would like to change the command to something not ending in `j` as discussed.

thanks

PR #18929, opened 2021-07-23
Fix *NULL derfrence in /libr/main/rvc2.c

RHL120:
**Checklist**

- [ ] Closing issues: #issue

**Description**
There was a *NULL derf in /libr/main/rvc2.c: ln 107 this PR is supposed to fix it.

### DIFF
--- a/libr/main/rvc2.c
+++ b/libr/main/rvc2.c
@@ -104,6 +104,11 @@ R_API int r_main_rvc2(int argc, const char **argv) {
 			}
 		}
 		RCore *core = r_core_new ();
+		if (!core) {
+			free (rp);
+			r_list_free (files);
+			return -11;
+		}
 		const char *author = r_config_get (core->config, "cfg.user");
 		bool ret = r_vc_commit (rp, message, author, files);
 		free (message);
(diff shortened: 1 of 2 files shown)


libr/main/rvc2.c
```diff
@@ -104,6 +104,12 @@ R_API int r_main_rvc2(int argc, const char **argv) {
 			}
 		}
 		RCore *core = r_core_new ();
+		if (!core) {
+			free (message);
+			free (rp);
+			r_list_free (files);
+			return -11;
+		}
 		const char *author = r_config_get (core->config, "cfg.user");
```
trufae, line 113: If u only need core to get the user just use the rsysuser call. Also core is never freed later

libr/main/rvc2.c
```diff
@@ -104,6 +104,12 @@ R_API int r_main_rvc2(int argc, const char **argv) {
 			}
 		}
 		RCore *core = r_core_new ();
+		if (!core) {
+			free (message);
```
trufae, line 108: Better use goto instead of freeing on wvery failure the same items

trufae (approved):

Issue #18876, opened 2021-07-01
x86.nz invalid assembly


radare:
That shouldnt be assembleable

```
$ rasm2 'push 0x3311223344899999'
6899998944
$
```

PR #18942, opened 2021-07-26
Fix x86.nz invalid assembly #18876

as0ler:


libr/asm/p/asm_x86_nz.c
```diff
@@ -2466,6 +2466,10 @@ static int oppush(RAsm *a, ut8 *data, const Opcode *op) {
 			}
 		}
 	} else {
```
trufae, line 2468: Thats a compile time check. You should check the immediate value instead of its size

libr/asm/p/asm_x86_nz.c
```diff
@@ -2466,6 +2466,10 @@ static int oppush(RAsm *a, ut8 *data, const Opcode *op) {
...
+		if (a->bits == 32 && sizeof(op->operands[0].immediate) > 4) {
+			eprintf ("Invalid Immediate value\n");
```
trufae, line 2470:
```suggestion
			eprintf ("Immediate value is out of range.\n");
```

libr/asm/p/asm_x86_nz.c
```diff
@@ -2466,6 +2466,10 @@ static int oppush(RAsm *a, ut8 *data, const Opcode *op) {
 			}
 		}
 	} else {
+		if (a->bits == 32 && (op->operands[0].immediate >> 32) > 0) {
```
trufae, line 2469: I think is better to drop the >0. Negative values could be a problem if imm is signed i think

```suggestion
		if (a->bits == 32 && (op->operands[0].immediate >> 32)) {
```

trufae (commented):
Looks good after this change! Can you add a test?

radare:
Also this bound check affects many other instructions. can you write some tests for them too?

radare:
Remove the submodule that is breaking the build please:

```
0 / 2 files viewed
Review changes 
 1  libr/asm/arch/arm/v35arm64/arch-arm64 
Viewed

 Submodule arch-arm64 added at 583791
 ```

File: libr/asm/p/asm_x86_nz.c (hunk 3 above)
Comment (trufae) [reply libr/asm/p/asm_x86_nz.c:2469]: ^

**Review by trufae — CHANGES_REQUESTED** [review]
.

File: libr/asm/p/asm_x86_nz.c [hunk 4]
```diff
@@ -1813,6 +1817,10 @@ static int opmov(RAsm *a, ut8 *data, const Opcode *op) {
 		if (op->operands[1].immediate == -1 && a->num && a->num->nc.errors > 0) {
 			return -1;
 		}
+		if(immediate_out_of_range(a->bits, op->operands[1].immediate)) {
```
trufae, line 1820:
```suggestion
		if (immediate_out_of_range (a->bits, op->operands[1].immediate)) {
```

libr/asm/p/asm_x86_nz.c
```diff
@@ -136,6 +136,10 @@ typedef struct Opcode_t {
 	bool has_bnd;
 } Opcode;
 
+static int immediate_out_of_range(int bits, ut64 immediate) {
```
trufae, line 139:
```suggestion
static bool immediate_out_of_range(int bits, ut64 immediate) {
```

libr/asm/p/asm_x86_nz.c
```diff
@@ -136,6 +136,10 @@ typedef struct Opcode_t {
...
+	return (bits == 32 && (immediate >> 32) > 0);
```
trufae, line 140:
```suggestion
	return bits == 32 && (immediate >> 32); 
```

libr/asm/p/asm_x86_nz.c
```diff
@@ -2466,6 +2474,10 @@ static int oppush(RAsm *a, ut8 *data, const Opcode *op) {
 			}
 		}
 	} else {
+		if(immediate_out_of_range(a->bits, op->operands[0].immediate)) {
```
trufae, line 2477:
```suggestion
		if (immediate_out_of_range (a->bits, op->operands[0].immediate)) {
```

libr/asm/p/asm_x86_nz.c
```diff
@@ -2500,6 +2512,10 @@ static int opout(RAsm *a, ut8 *data, const Opcode *op) {
 			return l;
 		}
 	} else if (op->operands[0].type & OT_CONSTANT) {
+		if(op->operands[0].immediate > 255) {
```
trufae, line 2515:
```suggestion
		if (op->operands[0].immediate > 255) {
```

trufae (changes requested):
see code suggestions and lgtm if grin

trufae:
the new output is correct. use r2r -i to update the test .

Screenshot 2021-07-30 at 10 50 46

Issue #18860, opened 2021-06-26
(Maybe) Incorrect ESIL for "mul rdx"


ZhangZhuoSJTU:
## Environment

```sh
# copypaste this script into your shell and replace it with the output
$ date
Sat Jun 26 01:03:32 EDT 2021

$ r2 -v
radare2 5.4.0-git 26454 @ darwin-x86-64 git.5.3.1
commit: 635fad4eb9a7b6c7f41de7f72525083fc74213c1 build: 2021-06-26__00:24:45

$ uname -ms
Darwin x86_64
```

## Description

For x64, the ESIL of `mul rdx` seems incorrect.

```sh
$ rasm2 -a "x86" -b 64 "mul rdx"
48f7e2

$ rasm2 -a "x86" -b 64 -E 48f7e2
64,rdx,rax,*,>>,rdx,=,rdx,rax,*=,rdx,?{,1,1,}{,0,0,},cf,:=,of,:=
```

Specifically, the value of _rdx_ is changed before setting the value of _rax_.

For example, if we have `rdx = 8` and `rax = 4` at the beginning. 

For the result ESIL `64,rdx,rax,*,>>,rdx,=,rdx,rax,*=,rdx,?{,1,1,}{,0,0,},cf,:=,of,:=`,

When interpreting the first`64,rdx,rax,*,>>,rdx,=`, we will have: `rdx = 0` and `rax = 4`.
Then we interpret `rdx,rax,*=`, we will have: `rdx = 0` and `rax = 0`.

However, the correct result should be: `rdx = 0` and `rax = 32`.

## Test

Following aforementioned steps.

radare:
as for reference it should be:

```c
void multiply64(uint64_t a, uint64_t b, uint64_t *h, uint64_t *l) {
	uint64_t const base = 0x100000000;
	uint64_t al = a % base;
	uint64_t ah = a / base;
	uint64_t bl = b % base;
	uint64_t bh = b / base;
	*l = al * bl;
	*h = ah * bh;
	uint64_t rlh = *l / base + al * bh;
	*h += rlh / base;
	rlh = rlh % base + ah * bl;
	*h += rlh / base;
	*l = (rlh % base) * base + *l %base;
}
```

aemmitt-ns:
I wrote this and will fix, it should be using the L* operator that exists now for 64 bit muls.

aemmitt-ns:
Though that isn't really the issue here, the esil will also fail for mul edx but I will fix that too.

PR #18950, opened 2021-07-30
fix mul and imul for *dx operands and 64 bit widths ##esil

aemmitt-ns:
**Checklist**

- [x] Closing issues: #18860

**Description**

Fixes my bad mul / imul ESIL which did not work for *dx operands (the register is written to and then used again as if it were not) and also MUL was not valid for 64 bits. 64 bit IMUL instructions will still be wrong since theres no 128 bit sign extension. I will fix this at some point. I really hate IMUL.

### DIFF
--- a/libr/anal/p/anal_x86_cs.c
+++ b/libr/anal/p/anal_x86_cs.c
@@ -1712,10 +1712,12 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 						if ( width == 1 ) {
 							esilprintf(op, "0xffffff00,eflags,&=,%s,%s,%%,eflags,|=,%s,%s,*,%s,=,0xff,eflags,&,%s,=,0xffffff00,eflags,&=,2,eflags,|=",
 								arg0, r_nume, arg0, r_nume, r_quot, r_rema);
+						} else if ( width == 8 ) { // TODO still needs to be fixed to handle correct signed 128 bit value 
+							esilprintf (op, "%s,%s,L*,%s,=,DUP,%s,=,!,!,DUP,cf,:=,of,:=",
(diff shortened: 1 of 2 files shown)


aemmitt-ns:
I am going to fix the tests and also refactor the IMUL code. its gotten real ugly, mostly because of me

trufae:
Can you update the tests and add a couple of tests emulating and checking the registers afterwards? Having yests with the esil string around is not much useful i think.

we all hate mul and div :{{ totally not pleasant to describe

aemmitt-ns:
ok still going to add some emulation tests. i agree that the string tests aren't super useful. I'd be ok with getting rid of them in general and relying only on tests of the emulation results.

aemmitt-ns:
Alright I think this is ready to go now. Added tests for the relevant instructions and everything is passing now

libr/anal/p/anal_x86_cs.c
```diff
@@ -1697,29 +1697,26 @@ static void anop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
...
-						}
-					}
-					else {
-						/* should never happen */
+				if (arg0) {
+					const char *r_quot = (width==1)?"al": (width==2)?"ax": (width==4)?"eax":"rax";
+					const char *r_rema = (width==1)?"ah": (width==2)?"dx": (width==4)?"edx":"rdx";
+					const char *r_nume = (width==1)?"ax": r_quot;
+
+					if ( width == 8 ) { // TODO still needs to be fixed to handle correct signed 128 bit value
```
trufae, line 1714:
```suggestion
					if (width == 8) { // TODO still needs to be fixed to handle correct signed 128 bit value 
```

trufae (approved):

PR #18956, opened 2021-07-30
Move CAN table search to /Du command

sylvainpelissier:
**Checklist**

- [ ] Closing issues: #issue

**Description**
Can table search was previously in /cu but since it is not related to Cryptography I moved it to /Du. Test have been moved.

### DIFF
--- a/libr/asm/arch/arm/armass64.c
+++ b/libr/asm/arch/arm/armass64.c
@@ -346,11 +346,22 @@ static ut32 cmp(ArmOp *op) {
 			return UT32_MAX;
 		}
 		k =  0x1f00006b;
+	} else if (op->operands[0].type & ARM_GPR  && op->operands[1].type & ARM_CONSTANT) {
+		if (op->operands[0].reg_type & ARM_REG64) {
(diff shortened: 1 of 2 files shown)


trufae:
Uhm, Why /Du if there's no /D. and also why use /D when there are still lowercase letters to pick? personally i would prefer to discuss a better command before moving forward on this change.

Do you have other suggestions? Another option would be to change the meaning of /c or.. juse use /ru (because at the end, we are looking for references to the UDS entries of the CAN table..) even if that's related to data and not code.

What do you think?

sylvainpelissier:
Sorry last time we discuss I thought it was the solution which was the best. Indeed keeping /c is the simplest but I do not have a global view like you.

sylvainpelissier:
After thinking about that, I think the best is to have `/ru` to keep the mnemonic `/c` for Cryptography. If you agree I'll do the change.

trufae:
Sounds good to me!

trufae:
Please solve the conflicts (aka rebase) before changing that command. thanks!

trufae:
ping

PR #18966, opened 2021-08-03
Exclude files of nested rvc repos from repo_files() ##projects

RHL120:
**Checklist**

- [ ] Closing issues: #issue

**Description**
This PR prevents repo_files() from traversing dirs with a .rvc in them (besides the root)

### DIFF
--- a/libr/core/rvc.c
+++ b/libr/core/rvc.c
@@ -918,20 +918,13 @@ R_API bool r_vc_branch(const char *rp, const char *bname) {
 R_API bool r_vc_new(const char *path) {
 	Sdb *db;
 	char *commitp, *blobsp;
-	switch (repo_exists (path)) {
-	case 0:
-		break;
-	case 1:
-		eprintf ("Repo already exists at: %s", path);
-		return false;
-	case -1:
-		eprintf ("Can't create repo\n");
-		return false;
-	case -2:
-		eprintf ("Repository exists but is corrupt\n");
-		return false;
-	}
 	char *vcp = r_str_newf ("%s" R_SYS_DIR ".rvc", path);
+	if (r_file_is_directory (vcp)) {
+		eprintf ("A directory already exists in %s\n", path);
+		free (vcp);
+		return false;
+
+	}
 	if (!vcp) {
 		return false;
 	}
--- a/libr/core/rvc.c
+++ b/libr/core/rvc.c
@@ -404,13 +404,15 @@ static bool traverse_files(RList *dst, const char *dir) {
 }
 
 static RList *repo_files(const char *dir) {
-	RList *ret = r_list_new ();
-	if (!ret && !traverse_files (ret, dir)) {
-		r_list_free (ret);
-		return NULL;
-	}
-	return ret;
-}
+ 	RList *ret = r_list_newf (free);
+ 	if (ret) {
+           if (!traverse_files (ret, dir)) {
+ 		r_list_free (ret);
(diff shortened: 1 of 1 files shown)


libr/core/rvc.c
```diff
@@ -355,46 +355,65 @@ static bool rm_empty_dir(const char *rp) {
 	return true;
 }
 
-static RList *repo_files(const char *rp) {
-	RList *files = r_file_lsrf (rp);
-	if (!files) {
-		return NULL;
+//This is why anonymous functions are cool
```
trufae, line 358: This is not anonymous. Anonymous means it have no name. And C doesnt support such things. Thats just a static/local function

libr/core/rvc.c
```diff
@@ -355,46 +355,65 @@ static bool rm_empty_dir(const char *rp) {
...
-
 	}
-	free (repo_meta);
 	r_list_free (files);
 	return ret;
+}
 
+static RList *repo_files(const char *dir) {
+	RList *ret = r_list_new ();
+	if (!ret) {
```
trufae, line 409: Remove this

libr/core/rvc.c
```diff
@@ -355,46 +355,65 @@ static bool rm_empty_dir(const char *rp) {
...
+		return NULL;
+	}
+	if (!dr (ret, dir)) {
```
trufae, line 412: Use if (ret && !dr..

libr/core/rvc.c
```diff
@@ -898,20 +917,13 @@ R_API bool r_vc_branch(const char *rp, const char *bname) {
...
-		eprintf ("Repo already exists at: %s", path);
-		return false;
-	case -1:
-		eprintf ("Can't create repo\n");
-		return false;
-	case -2:
-		eprintf ("Repository exists but is corrupt\n");
+	char *vcp = r_str_newf ("%s" R_SYS_DIR ".rvc", path);
+	if (r_file_is_directory (vcp)) {
+		eprintf ("A directory already exists in %s\n", path);
```
trufae, line 922: A repository better than a directory

libr/core/rvc.c
RHL120: yes lol, that't the point I wish C had Anonymous functions becaus if it did I would have been able to embed dr into repo_files...

trufae: Remove the comment and find a better name for the function

libr/core/rvc.c
RHL120: Remove repo_files? but if I did we would habe to alloc everytime we list the files. That would be 9 lines per repo_file.

libr/core/rvc.c
RHL120: What should that name be?

libr/core/rvc.c
```diff
@@ -355,46 +355,61 @@ static bool rm_empty_dir(const char *rp) {
...
 			break;
 		}
-
 	}
-	free (repo_meta);
 	r_list_free (files);
 	return ret;
+}
 
+static RList *repo_files(const char *dir) {
```
radare, line 406:
```c
static RList *repo_files(const char *dir) {
 	RList *ret = r_list_newf (free);
 	if (ret) {
           if ( !traverse_files (ret, dir)) {
 		r_list_free (ret);
 		ret = NULL;
           }
 	}
 	return ret;
 }
```

maybe like this?

PR #19023, opened 2021-08-27
Fix #19019 - ##asm - Added OT_GPREG flag to instructions with byte, word, qword etc.,

### LINKED ISSUE #19019: rasm2 generating wrong machine code for ```jmp qword ptr [rax - 0x2a]```
Hi folks,

## Environment

```sh
$ r2 -v
radare2 5.4.0 26506 @ linux-x86-64 git.5.4.0-git
commit: e6967804743227f6bca3da01b636ae83706d222c build: 2021-08-23__19:59:24
$ uname -ms
Linux x86_64
```

## Description

rasm2 is assembling ```jmp qword ptr [rax - 0x2a]``` to give "eb28". But this is the machine code of ```jmp 0x2a```. Verified the same with nasm as well. For ```jmp qword ptr [rax - 0x2a]```, machine code is "ff 60 d6"

Interesting part is that any ```jmp qword ptr [REG - 0x2a]``` is assembed to "eb28", REG = any of the 16 GPRs and ```rip```. For REG=```rflags```, we get "eb29" (which is actually ```jmp 0x2b```)

## Test

```
$ rasm2 -a x86 -b 64 "jmp qword ptr[rax - 0x2a]"
eb28
$ rasm2 -a x86 -b 64 -d "eb28"
jmp 0x2a
$ rasm2 -a x86 -b 64 -d "ff60d6"
jmp qword [rax - 0x2a]
```

Thanks,
Adwaith

adwait1-G:
**Checklist**

- [X] Closing issues: #19019

**Description**

1. Please check Issue [19019](https://github.com/radareorg/radare2/issues/19019) for proper description of the problem.


trufae:
Yes, go for it. I dont want the byte/word/dword/qword to be required to work, it should bound to the register size by default imho.

can you verify with nasm or gas or just rasm2. Those instructions are valid

14A0DAC0-C432-4A67-9341-99B1445E50D3

trufae:
ping?

adwait1-G:
Hi,

I have gone over the the sourcefile and have updated the instructions which used OT_GPREG and OT_MEMORY. Have added tests for those instructions to make sure nothing is broken because of these changes.

But during this, I found 2 fundamental problems with the assembler(along with a few other bugs):

1. We are not handling instructions involving the new 8 GPRs(r8-r15) of x64 properly. I think except for the most used(and thus most tested) instructions, many others have this problem. Here are a few examples:
```asm
$ rasm2 -a x86.nasm -b 64 "and r15b, 0x2a"
4180e72a
$ rasm2 -a x86 -b 64 "and r15b, 0x2a"
ff2a
-----
$ rasm2 -a x86 -b 64 "xor r15, 0x2a"
4883f72a
$ rasm2 -a x86.nasm -b 64 "xor r15, 0x2a"
4983f72a
```

2. We don't seem to be handling 16-bit instructions properly - especially when the keyword ```word``` is used.

```asm
$ rasm2 -a x86.nasm -b 64 "and word[rax], bx"
662118
$ rasm2 -a x86 -b 64 "and word[rax], bx"
Error: mismatched operand sizes
Error: mismatched operand sizes
Error: mismatched operand sizes
Cannot assemble 'and word[rax], bx' at line 3
invalid
```
I have added 'aB' type tests to track the ones I caught.

Regards,
Adwaith

trufae:
Yep its known. The initial aim for this assembler was to cover 32 and 64bit instructionsets at least the most generic ones.

the name “nz” stands for non-zero, because initially it tried to generate bytecode without null bytes. This is not the currentapproach and as long as it become the standard assembler for intel cpus on r2 i think its worth supporting 16 bit instructions too.

actually for r2wars it would make a lot of sense to open the x86-16 category too :)

Thanks for the improvements! The testsuite may ideally support assembly and disassembly for all instructionsets. So we may eventually be able to recompile portions of programs with just the text representation

keep up those contribs! Kudos

PR #19032, opened 2021-08-30
Add a status and log commands to ravc2 and add tests for rvc

RHL120:
**Checklist**

- [ ] Closing issues: #issue

**Description**
I have added a ravc2 log command which outputs the list of commits, a ravc2 status command which outputs thew current branch and unstaged files. I have also added tests.

### DIFF
--- a/libr/core/rvc.c
+++ b/libr/core/rvc.c
@@ -76,7 +76,7 @@ static Sdb *vcdb_open(const char *rp) {
 	if (!frp) {
 		return NULL;
 	}
-	Sdb *ret = sdb_new0();
+	Sdb *db = sdb_new0 ();
 	if (!ret) {
 		free (frp);
 		return NULL;
--- a/libr/core/rvc.c
+++ b/libr/core/rvc.c
@@ -437,7 +437,8 @@ static RList *repo_files(const char *dir) {
 }
 
 //shit function:
-static RList *get_uncommitted(const char *rp) {
+R_API RList *r_vc_get_uncommitted(const char *rp) {
+	CHECK_REPO (rp, "Cant get uncommitted");
 	RList *blobs = get_blobs (rp);
 	if (!blobs) {
 		return NULL;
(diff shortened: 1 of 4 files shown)


libr/core/rvc.c
```diff
@@ -57,6 +71,25 @@ static char *strip_sys_dir(const char *path) {
 	return ret;
 }
 
+static Sdb *vcdb_open(const char *rp) {
+	char *frp = r_str_newf ("%s" R_SYS_DIR ".rvc" R_SYS_DIR DBNAME, rp);
+	if (!frp) {
+		return NULL;
+	}
+	Sdb *ret = sdb_new0();
```
trufae, line 79:
```suggestion
	Sdb *db = sdb_new0 ();
```

trufae: `db` is better name than `ret`here

libr/core/rvc.c
```diff
@@ -1003,19 +937,7 @@ R_API bool r_vc_new(const char *path) {
...
-		eprintf ("No repo in %s\nCan't checkout\n", rp);
-		return false;
-	case -1:
-		eprintf ("Can't checkout\n");
-		return false;
-	case -2:
-		eprintf ("Can't checkout");
-		return false;
-	}
+	CHECK_REPO (rp, "Can't checkout");
```
trufae, line 940: if (!check_repo (rp, true)) {return false;}

test/db/tools/ravc2
```diff
@@ -0,0 +1,76 @@
...
+cd ravc2_branch_test
+!!ravc2 init
+!!ravc2 branch hello
+echo hello > hello
+!!ravc2 commit "hello world!" hello
+!!ravc2 checkout hello
+ls
+!!ravc2 status~Branch:
+!!ravc2 checkout master
+ls
```
trufae, line 51: you can use ls -q for one file per line which is more readable

trufae (changes requested):

libr/core/rvc.c
RHL120: I think I may have misnamed repo_exists the purpose of it is not to check wheter or not a repo exists but to run digonstics on it. Do you think I should keep the functionality or change it?

test/db/tools/ravc2
RHL120: yes, but the reason tests are failing is because the entries are sorted differently.

PR #19036, opened 2021-08-31
Fix the rvc asan tests and some heap issues in rvc ##vc

RHL120:
**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/libr/core/rvc.c
+++ b/libr/core/rvc.c
@@ -46,7 +46,8 @@ static char *strip_sys_dir(const char *path) {
 	for (; *path && !(*path == *R_SYS_DIR && !*(path + 1)); path++) {
 		if (!r_str_cmp (path, R_SYS_DIR R_SYS_DIR, 2)) {
 			continue;
-		} ret = r_str_appendf (ret, "%c", *path);
+		}
+		ret = r_str_appendf (ret, "%c", *path);
 		if (!ret) {
 			return NULL;
 		}
@@ -480,8 +481,7 @@ R_API RList *r_vc_get_uncommitted(const char *rp) {
 				free (blob_absp);
 				continue;
 			}
-			char *fname = r_str_new (blob_absp);
-			free (blob_absp);
+			char *fname = blob_absp;
 			if (!fname) {
 				goto fail_ret;
 			}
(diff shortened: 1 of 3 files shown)


libr/core/rvc.c
```diff
@@ -44,11 +44,9 @@ static char *strip_sys_dir(const char *path) {
 		return NULL;
 	}
 	for (; *path && !(*path == *R_SYS_DIR && !*(path + 1)); path++) {
-		if (*path == *R_SYS_DIR &&
-				*(ret + r_str_len_utf8 (ret) - 1) == *R_SYS_DIR) {
+		if (!r_str_cmp (path, R_SYS_DIR R_SYS_DIR, 2)) {
 			continue;
-		}
-		ret = r_str_appendf (ret, "%c", *path);
+		} ret = r_str_appendf (ret, "%c", *path);
```
trufae, line 49: missing newline

libr/core/rvc.c
```diff
@@ -479,9 +477,11 @@ R_API RList *r_vc_get_uncommitted(const char *rp) {
 		}
 		if (!found) {
 			if (!strcmp (NULLVAL, blob->fhash)) {
+				free (blob_absp);
 				continue;
 			}
-			char *fname = r_str_new (blob_absp) ;
+			char *fname = r_str_new (blob_absp);
```
trufae, line 483: just char *fname = blob_absp; and you save an unnecessary free and a strdup

trufae (changes requested):

libr/core/rvc.c
```diff
@@ -479,9 +478,10 @@ R_API RList *r_vc_get_uncommitted(const char *rp) {
 		}
 		if (!found) {
 			if (!strcmp (NULLVAL, blob->fhash)) {
+				free (blob_absp);
 				continue;
 			}
-			char *fname = r_str_new (blob_absp) ;
+			char *fname = blob_absp;
```
trufae, line 484: by reading the code above, i think blob_absp cant be null, so you dont need to check that if below. and as long as fname is only used in the list_append below its better to just pass the blob_basp to .append() instead of fname so you save one extra var

PR #19042, opened 2021-09-01
Correct MZ size when bytes_in_last_block = 0

guest177:
**Checklist**

- [ ] Closing issues: #issue

**Description**

`r_bin_mz_init_hdr` computes `dos_file_size` incorrectly when `bytes_in_last_block` is 0.

https://github.com/radareorg/radare2/blob/c4622b6603454bf7061d02db6926db5bd1ffc4d1/libr/bin/format/mz/mz.c#L211-L215

`bytes_in_last_block` = 0 means the last block is completely full, not completely empty. See for example http://www.delorie.com/djgpp/doc/exe/:

> If this value is zero, that means the entire last block is used (i.e. the effective value is 512).

You can see the problem using the attached [test.exe](https://github.com/radareorg/radare2/files/7088519/test.exe.gz), which is constructed to have a size that is a multiple of 512. It was made from the following assembly file:

```
; nasm -fobj test.asm
bits	16
section	code
..start:
	xor	al, al
	mov	ah, 0x4c
	int	0x21
times	512-($-$$)	nop
```

The file has 512 bytes of header and 512 bytes of program text, but r2 thinks it has 512 bytes of header and 0 bytes of program text.

```
$ rabin2 -S test.exe
Error: entry point outside load module
Error: entry point outside load module
[Sections]

nth paddr       size vaddr       vsize perm name
――――――――――――――――――――――――――――――――――――――――――――――――
0   0x00000200   0x0 0x00000000    0x0 -rwx seg_000
```

With this change, the output is:

```
$ rabin2 -S test.exe
[Sections]

nth paddr        size vaddr       vsize perm name
―――――――――――――――――――――――――――――――――――――――――――――――――
0   0x00000200    0x0 0x00000000    0x0 -rwx seg_000
1   0x00000200  0x200 0x00000000  0x200 -rwx seg_001
```


trufae:
The fix looks good! Can you add a test case and commit the sample bin in the testbins repo? Thanks!

guest177:
> Can you add a test case and commit the sample bin in the testbins repo?

All right, I did my best, pushed a revised commit, and opened https://github.com/radareorg/radare2-testbins/pull/66 in testbins. I copied the form of the db/formats/mz/unzip test.

test/db/formats/mz/cblp0
```diff
@@ -0,0 +1,32 @@
...
+FILE=bins/mz/cblp0.exe
+ARGS=-A
+CMDS=q!
+EXPECT=<<EOF
+EOF
+RUN
+
+NAME=MZ: cblp0.exe - entrypoint
+FILE=bins/mz/cblp0.exe
+CMDS=s
```
radare, line 11: better check the output of `ie` (not just the initial seek)

test/db/formats/mz/cblp0
```diff
@@ -0,0 +1,32 @@
...
+EXPECT=<<EOF
+0x0
+EOF
+RUN
+
+NAME=MZ: cblp0.exe - pi 1
```
radare, line 17: you can run pio 1 and iS* in the same test. the more tests we have the slower the CI runs, so its better to squash some tests to work together in case those cant be side-effected.

PR #19049, opened 2021-09-03
Aarch64 correct FPU and SIMD register types

sylvainpelissier:
**Checklist**

- [ ] Closing issues: #issue

**Description**
Registers Bn, Hn, Sn, Dn, and Vn are FPU and SIMD registers thus they should be labeled fpu.


trufae:
Can you do the same change for v35?

PR #19053, opened 2021-09-04
Add an rr function which does rm -rf 

RHL120:
**Checklist**

- [ ] Closing issues: #issue

this adds a command called rr which allows the removal of a non-empty directory with its contents


trufae:
Imho rr should be renamed to rmrf or rm -rf. R is a command to resize and rm is just there because unix as well as rm did

PR #19072, opened 2021-09-12
Fix const clobber in ar ##io

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Small fix to prevent the ar io plugin from truncating the provided uri.

### DIFF
--- a/libr/io/p/io_ar.c
+++ b/libr/io/p/io_ar.c
@@ -34,7 +34,8 @@ static int r_io_ar_close(RIODesc *fd) {
 
 static RIODesc *r_io_ar_open(RIO *io, const char *file, int rw, int mode) {
 	r_return_val_if_fail (r_io_ar_plugin_open (io, file, false), NULL);
-	const char *arname = strstr (file, "://") + 3;
+	char *uri = strdup (file);
+	const char *arname = strstr (uri, "://") + 3;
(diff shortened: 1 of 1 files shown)


libr/io/p/io_ar.c
```diff
@@ -34,7 +34,8 @@ static int r_io_ar_close(RIODesc *fd) {
 
 static RIODesc *r_io_ar_open(RIO *io, const char *file, int rw, int mode) {
 	r_return_val_if_fail (r_io_ar_plugin_open (io, file, false), NULL);
-	const char *arname = strstr (file, "://") + 3;
+	char *uri = strdup (file);
+	const char *arname = strstr (uri, "://") + 3;
```
trufae, line 38: Uhm i think this should be checking for null before the +3

swoops: The `r_io_ar_plugin_open` in `r_return_val_if_fail` should verify that the file is of the form `ar://...` so strstr shouldn't fail.

This is not obvious though, I was halfway through fixing it when I realized. Should I add a comment or something?

trufae: Yes i think that all functions should not segfault whethever you call them . Right now open is not called without check.. but i think -F overrides this so there shohld be ways to reach that condition. Can you make another pr fixing this?

PR #19084, opened 2021-09-14
Fix unreleased locks

ryancaicse:
The lock subprocs_mutex is not released correctly when write (proc->killpipe[1], &r, 1) != 1 (the second break) in the function static RThreadFunctionRet sigchld_th(RThread *th). Also, the lock could be released before exit (-1); for better resource management and code symmetry.

This PR is to fix them. Thank you for your checking.

### DIFF
--- a/binr/r2r/r2r.c
+++ b/binr/r2r/r2r.c
@@ -464,6 +464,7 @@ int main(int argc, char **argv) {
 		RThread *th = r_th_new (worker_th, &state, 0);
 		if (!th) {
 			eprintf ("Failed to start thread.\n");
+			r_th_lock_leave (state.lock);
 			exit (-1);
 		}
 		r_pvector_push (&workers, th);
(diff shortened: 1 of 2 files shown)


binr/r2r/r2r.c
```diff
@@ -464,6 +464,7 @@ int main(int argc, char **argv) {
 		RThread *th = r_th_new (worker_th, &state, 0);
 		if (!th) {
 			eprintf ("Failed to start thread.\n");
+			r_th_lock_leave (state.lock);
 			exit (-1);
```
trufae, line 468: exit should free all resources so it shouldnt be a problem imho. but it's formally correct to do so i guess :)

ryancaicse: Hi, indeed, it is harmless since most of the OS(some OS do not like embedded systems) would release the resources when exiting. I think releasing a lock at the source code level is good for resource management : )

ryancaicse:
Thanks

PR #19087, opened 2021-09-15
Add path to vcvarsall.bat from VS 2019 Professional

bdoner:
Currently the `preconfigure.bat` file only looks at the community edition
and the Enterprise edition of VS 2019. This change adds Professional and BuildTools 
edition to the search.

**Checklist**

- [ ] Closing issues: #issue

**Description**

The `preconfigure.bat` scripts is looking for a heap of tools needed in order to build radare2 on Windows. One of such tools is the [Microsoft C++ Compiler](https://docs.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-160#developer_command_file_locations) which can be included in the environment of your current console executing the `vcvarsall.bat` file. The `vcvarsall.bat` file can be located in a few different places depending on your edition of Visual Studio. This change looks at the other two default locations in addition to the two locations `preconfigure.bat` already looked.

### DIFF
--- a/preconfigure.bat
+++ b/preconfigure.bat
@@ -45,19 +45,28 @@ cl --help > NUL 2> NUL
 if %ERRORLEVEL% == 0 (
   echo FOUND
 ) else (
-
-if EXIST "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community" (
-  echo "Found community edition"
-  call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
-) else (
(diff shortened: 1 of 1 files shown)


trufae:
Nice! Wish there was a better way to select and initialize msvc toolchains from commandline without hardcoded paths. But that was the best way i though it could be done for now.

We will need to add new entries when new vs is out

bdoner:
You could probably loop through all the version folders, under the Visual Studio installation folder, sort them and from there look through the editions of the most recent version. But my batch-fu is not strong enough to do that confidently :)

This was just an issue I faced while building that I thought others could gain from as well.

PR #19110, opened 2021-09-19
PE sections == segments, but segments are the ones that actually map io ##bin

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**


trufae:
lets close this for now, but PE plugins behave incorrectly and we should keep io interactions clear at some point by separating segments and sections at api level, not field ("is_segment")

Issue #19118, opened 2021-09-19
If the command server is started in the background, the console becomes silent


thymol0:
## Environment

```sh
Sun Sep 19 22:22:23 GMT 2021
radare2 5.4.1 26822 @ linux-x86-64 git.5.4.0-145-g719561644
commit: 719561644889df23b4a5a475963cfb008251fccf build: 2021-09-19__21:40:03
Linux x86_64
```

## Description

When a background task is executed the console doesn't print anything until the task is finished.

## Test

```
$ r2 -
[0x00000000]> & =t9000
[0x00000000]> Listening for commands on port 9000
[0x00000000]>
[0x00000000]> pd 1
```
No result will be printed to the console. If you break the background task then the queued console buffer will be printed.
```
[0x00000000]> &b 1
Task 1 finished
[0x00000000]> 
            0x00000000      0000           add byte [rax], al
```

thymol0:
I am trying to fix this issue myself, my understanding so far:

1. There is only one global console with only one active context, this context holds the console buffer where the output is buffered before radare decides to print if the function called is `r_core_cmd` or to return it as a string if the function called is `r_core_cmd_str`.
2. Each background tasks runs in a dedicated thread.
3. Background tasks use `r_core_cmd_str`.
4. `r_core_cmd_str` disables console printing.

Since there is only one global console instance shared between all threads with only one active context, then all background tasks will be writing to the same context (i.e same buffer).

The solution is to either make an instance of the console for each thread or to make the console context different for each thread.

@trufae can you guide me here? running `grep -r -e '->context' libr| wc -l` return 269 matches, while I have no problem replacing every access to the console context with a a call to some function that returns the thread local context like `r_cons_get_context` or something, do you have something else in mind?

trufae:
Thats how tasks work. The output of the task is saved in a buffer. Use the & command to collect the output

trufae:
Oh sorry missreas the issue affects shell commands not the output of the task! Then thats an issue

thymol0:
It seems the issue is quite deep. The `core` is shared and it is not meant to be used concurrently for example if two tasks run concurrently and one task changes the config to `e asm.hints=false` while the other sets `e asm.hints=true`, the result will be incorrect for one of them. In order to solve this systematically there will be quite a bit of restructuring in `libr/core/` to make it centered around tasks instead of being centered on the `core`.

[...]

PR #19120, opened 2021-09-20
Move the console flushing decision to the console context

thymol0:
When a server is started as a background task, the console will remain silent until the server stops. This is undesirable because the intent of running the server in the background is to be able to keep working in radare while the server is running.

**Checklist**

- [x] Closing issues: #19118

### DIFF
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -116,6 +116,8 @@ static void cons_context_init(RConsContext *context, R_NULLABLE RConsContext *pa
 	context->event_interrupt_data = NULL;
 	context->pageable = true;
 	context->log_callback = NULL;
+	context->cmd_str_depth = 0;
+	context->noflush = false;
 
 	if (parent) {
 		context->color_mode = parent->color_mode;
@@ -588,7 +590,6 @@ R_API RCons *r_cons_new(void) {
 	I.force_columns = 0;
 	I.event_resize = NULL;
 	I.event_data = NULL;
-	I.noflush = false;
 	I.linesleep = 0;
 	I.fdin = stdin;
 	I.fdout = 1;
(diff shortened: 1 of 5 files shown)


trufae:
The reason for that is because the command nesting gets increased by the task running a command that doesn't ends.

Because this counter gets decremented when the command is executed. So this fix doesn't seems correct to me, despite working as expected for this use case. but could be wrong in other situations. Actually, the webserver behaves incorrectly after this change. even if started in foreground.

Sorry for the late response, ive been sick for 2 days unable to move from bed 🤒

thymol0:
Sorry to hear that. I hope for a quick recovery.

thymol0:
This doesn't decrement the nesting counter, we just ingore it. Can you expand on the web server's misbehaviour?

trufae:
I tried your patch and the webserver was working except the /cmd that returns empty results or partial

thymol0:
Yes, this patch is incorrect, I need to find a better solution.

trufae:
As an approach to find the root cause would be to find the offending commit. This was working at some point

trufae:
Any update?

thymol0:
I have been busy with other things. sorry.
Probably this portion has to be re-implemented
https://github.com/radareorg/radare2/blob/adbb29f9c2ee6679bc966e674227233f15b9866d/libr/core/cmd.c#L5520-L5533
The nesting counter should move from core struct to the console context struct, the `noflush` flag too should move to the context struct. The console should consult with the active context when it wants to flush. Since the output stored in the console buffer is contained inside the context which is unique for each task; the decision to flush the buffer should also be container inside the context.

trufae:
That makes sense, would be a good try to move this into the conscontext struct. Do you have some time this week to test it out?

thymol0:
I will try to implement this today.

thymol0:
Ok. The patch is more correct now.

trufae:
That fix looks good, but its surfacing another problem that i hope to fix soon. then i'll merge when confirmed :)

thanks!

PR #19128, opened 2021-09-22
Add AES Key Wrap Algorithm

sylvainpelissier:
**Checklist**

- [ ] Closing issues: #issue

**Description**
Implementation of AES Key Wrap Algorithm according to [RFC 3394](https://datatracker.ietf.org/doc/html/rfc3394) in radare2.


libr/crypto/p/crypto_aes_wrap.c
```diff
@@ -0,0 +1,149 @@
+/* radare - LGPL - Copyright 2021 - Sylvain Pelissier 
+   Implementation of AES Key Wrap Algorithm (RFC 3394) */
+
+#include <r_lib.h>
+#include <r_crypto.h>
+#include "crypto_aes_algo.h"
+
+#define BLOCK_SIZE 8
+
+static struct aes_state st;
```
trufae, line 10: may be good to not use globals. i think crypto plugins should have an init method to return that void* state to be used

trufae (changes requested):

libr/crypto/p/crypto_aes_wrap.c
sylvainpelissier: I do not see what is the problem with global here ? Should I change `RCryptoPlugin` to add a init function ?

trufae: The problem would be when this code gets used from two different threads or coroutines. Also another big problem with globals is that on some OS like android they are loaded once across different executions which means that the initial global state is initialized once, the subsequent executions will missbehave

sylvainpelissier: Since it require modification of `RCryptoPlugin` struct I would prefer to handle that in a separate PR.

trufae: 👍

PR #19134, opened 2021-09-23
Use non-interactive RCore in rasign2 -S

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Noticed on larger signature files that `rasign2 -S` would prompt and mess up the terminal. This should fix that.

I also notices rasig2 couldn't open /usr/libx32/libc.a even though it has 0644 perms. So I tried to change flags on the open command but it didn't work... I will look into it more later.

### DIFF
--- a/libr/main/rasign2.c
+++ b/libr/main/rasign2.c
@@ -34,7 +34,7 @@ static RCore *opencore(const char *fname) {
 		return NULL;
 	}
 	r_core_loadlibs (c, R_CORE_LOADLIBS_ALL, NULL);
-	r_config_set_i (c->config, "scr.interactive", false);
+	r_config_set_b (c->config, "scr.interactive", false);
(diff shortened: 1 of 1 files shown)


libr/main/rasign2.c
```diff
@@ -90,6 +90,7 @@ static int handle_sdb(const char *fname, struct rasignconf *conf) {
 	// can't use RAnal here because JSON output requires core, in a sneaky way
 	RCore *core = r_core_new ();
 	if (core && r_sign_load (core->anal, fname)) {
+		r_config_set_i (core->config, "scr.interactive", false);
```
trufae, line 93:
```suggestion
		r_config_set_b (core->config, "scr.interactive", false);
```

trufae (changes requested):

PR #19150, opened 2021-09-30
FIX #19149: Prevent NULL pointer dereference in line_header_fini()

### LINKED ISSUE #19149: Segmentation fault when opening elf file.
## Environment

Thu Sep 30 07:39:10 PM CEST 2021
radare2 5.4.0 0 @ linux-x86-64 git.5.4.0
commit: unknown build: 2021-09-07__19:12:53
Linux x86_64

(Arch Linux)

Also tested it with the newest code from git. Commit 688710932f8.

## Description

When opening a specific ELF file for an ARM cortex-m4 processor, radare2 closes unexpectadly with a segmentation fault in `line_header_fini` in libr/bin/dwarf.c

This is caused by an inconsistent setting of 
`hdr->file_names == NULL`
and
`hdr->file_names_count != 0`

In this case the code dereferences `hdr->file_names`

## Test

Exract the ELF from the attached zip and run `r2 -AA <path/to/ELF>

Output looks like:
`
[1]    27580 segmentation fault (core dumped)  r2 -AA test.elf
`

<!-- Steps to reproduce the issue and provide files/links/images/.. --
[test.zip](https://github.com/radareorg/radare2/files/7261534/test.zip)
>

0mhu:
hdr->file_names is set to NULL.
However, the corresponding count variable is not set to zero.
It might stay on a non-zero value and
cause the code in line_header_fini() to dereference the NULL pointer.

Setting hdr->file_names_count to zero solves that issue.

**Checklist**

- [x] Closing issues: #19149

**Description**

### DIFF
--- a/libr/bin/dwarf.c
+++ b/libr/bin/dwarf.c
@@ -612,6 +612,7 @@ static const ut8 *parse_line_header (
 	hdr->line_range = READ8 (buf);
 	hdr->opcode_base = READ8 (buf);
 
+	hdr->file_names_count = 0;
 	hdr->file_names = NULL;
 
 	if (mode == R_MODE_PRINT) {


trufae:
thanks for the fix! a test case would be ideal, but i understand where the problem comes from

PR #19157, opened 2021-10-02
Add r_core_cmd_tobuf() and overhaul alias system

Lazula:
**Checklist**

**Description**

The alias system up to this point has been pretty bare-bones and stored as string arrays, leading to slow lookups and incorrect behavior with commands that use raw bytes like `pr`. Aliases are now stored in a hashtable mapping a string key to a structure with a data buffer, size, string flag, and command flag. Aliases can now store both strings and raw bytes. To support this, API functions have been added for running commands and getting the raw output, and escaping unprintable characters in a buffer. The alias API is also greatly expanded, making it easier to work with in the future.

### DIFF
--- a/libr/core/cmd_api.c
+++ b/libr/core/cmd_api.c
@@ -1,6 +1,7 @@
 /* radare - LGPL - Copyright 2009-2021 - pancake */
 
 #include <r_core.h>
+#include "ht_pp.h"
 
 /*!
  * Number of sub-commands to show as options when displaying the help of a
@@ -101,9 +102,7 @@ err:
 }
 
 R_API void r_cmd_alias_init(RCmd *cmd) {
-	cmd->aliases.count = 0;
-	cmd->aliases.keys = NULL;
-	cmd->aliases.values = NULL;
+	cmd->aliases = ht_pp_new0 ();
 }
 
 R_API RCmd *r_cmd_new(void) {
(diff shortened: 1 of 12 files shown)


Lazula:
Updated `r_core_cmd_tobuf()` with new `core->cons->context` members

libr/core/cmd_api.c
```diff
@@ -185,114 +186,251 @@ R_API RCmdDesc *r_cmd_get_desc(RCmd *cmd, const char *cmd_identifier) {
...
 	}
-	return cmd->aliases.keys;
+
+	ht_pp_foreach (cmd->aliases, get_keys, keylist);
+
+	return keylist;
+}
+
+R_API void r_cmd_alias_free(RCmd *cmd) {
+	ht_pp_free(cmd->aliases);
```
trufae, line 206:
```suggestion
	ht_pp_free (cmd->aliases);
```

trufae:
```suggestion
	ht_pp_free (cmd->aliases);
```

libr/core/cmd_api.c
```diff
@@ -185,114 +186,251 @@ R_API RCmdDesc *r_cmd_get_desc(RCmd *cmd, const char *cmd_identifier) {
...
-		if (s) {
-			tofree = s;
-			v = (const char *)s;
-		}
+R_API RCmdAliasVal *r_cmd_alias_get(RCmd *cmd, const char *k) {
+	r_return_val_if_fail (cmd && cmd->aliases && k, NULL);
+
+	bool found;
+	RCmdAliasVal *v = ht_pp_find(cmd->aliases, k, &found);
+#if DEBUG_ALIAS
```
trufae, line 327: Maybe we can get rid of all this debugging code?

libr/core/cmd_api.c
```diff
@@ -185,114 +186,251 @@ R_API RCmdDesc *r_cmd_get_desc(RCmd *cmd, const char *cmd_identifier) {
...
+	ut8* out;
+	int out_sz;
+
+	/* If appending to a string, always overwrite the trailing \0 */
+	int bytes_from_first = first->is_str
+		? first->sz - 1
+		: first->sz;
+
+	out_sz = bytes_from_first + second_sz;
+	out = malloc (out_sz);
```
trufae, line 349: Check for null

libr/core/cmd_api.c
```diff
@@ -185,114 +186,251 @@ R_API RCmdDesc *r_cmd_get_desc(RCmd *cmd, const char *cmd_identifier) {
...
+	if (v_old) {
+		if (!v_old->is_data) {
+#if DEBUG_ALIAS
+			eprintf ("Won't append to existing command alias.\n");
+#endif
 			return 1;
 		}
+
+		int new_len;
+		ut8* new = alias_append_internal (&new_len, v_old, (ut8 *)a, strlen (a) + 1);
```
trufae, line 373: Check for null

libr/core/cmd_api.c
```diff
@@ -185,114 +186,251 @@ R_API RCmdDesc *r_cmd_get_desc(RCmd *cmd, const char *cmd_identifier) {
...
-			matches = !strcmp (k, cmd->aliases.keys[i]);
-		}
-		if (matches) {
-			return cmd->aliases.values[i];
-		}
+/* Returns a new copy of v->data. If !v->is_str, hex escaped */
+R_API char *r_cmd_alias_val_strdup(RCmdAliasVal *v) {
+	if (v->is_str) {
+		return strdup ((char *)v->data);
+	} else {
```
trufae, line 422: Unnecessary else

libr/core/cmd_api.c
```diff
@@ -185,114 +186,251 @@ R_API RCmdDesc *r_cmd_get_desc(RCmd *cmd, const char *cmd_identifier) {
...
+		return r_str_escape_raw (v->data, v->sz);
+	}
+}
+
+/* Returns a new copy of v->data. If !v->is_str, b64 encoded. */
+R_API char *r_cmd_alias_val_strdup_b64(RCmdAliasVal *v) {
+	if (v->is_str) {
+		return strdup ((char *)v->data);
+	} else {
```
trufae, line 431: Unnecessary else

libr/core/cmd_debug.c
```diff
@@ -4725,11 +4725,25 @@ static int cmd_debug_step (RCore *core, const char *input) {
...
 }
 
-static ut8*getFileData(RCore *core, const char *arg) {
+static ut8 *getFileData(RCore *core, const char *arg, int *sz) {
+	ut8 *out;
+	int size;
 	if (*arg == '$') {
-		return (ut8*) r_cmd_alias_get (core->rcmd, arg, 1);
+		RCmdAliasVal *v  = r_cmd_alias_get (core->rcmd, arg);
+		out = malloc (v->sz);
```
trufae, line 4733: Check for null

libr/core/cmd_write.c
```diff
@@ -588,10 +588,13 @@ static bool cmd_wff(RCore *core, const char *input) {
 	}
 	
 	if (*a == '$') {
-		const char *res = r_cmd_alias_get (core->rcmd, a, 1);
-		if (res) {
-			buf = (ut8*)strdup (res);
-			size = strlen (res);
+		RCmdAliasVal *v = r_cmd_alias_get (core->rcmd, a+1);
+		if (v) {
+			buf = malloc (v->sz);
```
trufae, line 593: Nullchk

libr/core/cmd_write.c
```diff
@@ -1546,27 +1549,55 @@ static int wt_handler_old(void *data, const char *input) {
...
 		}
-		tmp = *str? strchr (str + 1, ' ') : NULL;
+		size_sep = *str? strchr (str + 1, ' ') : NULL;
 		if (!filename || !*filename) {
 			const char* prefix = r_config_get (core->config, "cfg.prefixdump");
 			snprintf (_fn, sizeof (_fn), "%s.0x%08"PFMT64x, prefix, poff);
 			filename = _fn;
 		}
-		if (tmp) {
+#if 0
```
trufae, line 1558: Remove debug

libr/util/str.c
```diff
@@ -1381,6 +1381,23 @@ static char *r_str_escape_(const char *buf, int dot_nl, bool parse_esc_seq, bool
...
+R_API char *r_str_escape_raw(const ut8 *buf, int sz) {
+	r_return_val_if_fail (buf, NULL);
+
+	/* Worst case scenario, we convert every byte to a \xhh escape */
+	char *new_buf = malloc (1 + sz * 4);
+	if (!new_buf) {
+		return NULL;
+	}
+	char *q = new_buf;
+	for (int i = 0; i < sz; i++) {
```
trufae, line 1394: Move this int outside the for

trufae (changes requested):

libr/core/cmd_api.c
Lazula: That's fine with me

libr/core/cmd_api.c
Lazula: Will remove, though imo explicit else is fine in cases like this

libr/core/cmd_debug.c
Lazula: also missed doing arg+1 here in alias_get

Lazula: (aliases arent stored with $ anymore)

libr/util/str.c
Lazula: What's wrong with loop scope? Isn't it more confusing to declare outside?

trufae: It wont compile on macos-ppc and other systems with toolchains before y2k. Its a little change that doesnt hurt much and actually it looks to me like a c++ism, () shouldnt be a scope to define variables imho

trufae:
Ill have to take a look at the ci. Because there was a check to detect “for (int” and fail if found

libr/util/str.c
Lazula: It's C99, but imo being able to access a loop variable outside is not good and leads to people doing weird things with it, and smaller scope = less clutter when things do get more complicated.

If we need to do this for C89 or other weird legacy compatibility we should note that in developers/contributing docs

Lazula:
> Ill have to take a look at the ci. Because there was a check to detect “for (int” and fail if found

Seems strange, afaik any systems that dont support loop scope also don't support declarations not at top of block, and we do lots of that

libr/core/cmd_api.c
```diff
@@ -185,114 +184,200 @@ R_API RCmdDesc *r_cmd_get_desc(RCmd *cmd, const char *cmd_identifier) {
...
-				}
-				memmove (cmd->aliases.values,
-					cmd->aliases.values + 1,
-					cmd->aliases.count * sizeof (void*));
-				memmove (cmd->aliases.keys,
-					cmd->aliases.keys + 1,
-					cmd->aliases.count * sizeof (void*));
+R_API int r_cmd_alias_set_raw(RCmd *cmd, const char *k, const ut8 *v, int sz) {
+	RCmdAliasVal *val = R_NEW (RCmdAliasVal);
+	val->data = malloc (sz);
```
trufae, line 244: chekc for sz > 0 and if it returns != NULL

libr/core/cmd_api.c
```diff
@@ -185,114 +184,200 @@ R_API RCmdDesc *r_cmd_get_desc(RCmd *cmd, const char *cmd_identifier) {
...
+	memcpy (val->data, v, sz);
+	val->sz = sz;
+
+	/* If it's a string already, we speed things up later by checking now */
+	const ut8 *firstnull = NULL;
+	for (int i = 0; i < sz; i++) {
```
trufae, line 250: move thiss int outside the for()

test/db/cmd/cmd_alias
```diff
@@ -57,3 +57,77 @@ EXPECT=<<EOF
...
+CMDS=<<EOF
+wx 4141414100
+ps >$str_redirect_alias
+$str_set_alias=$BBBB
+wx 430043004300
+pr 6 >$bytes_redirect_alias
+$bytes_set_alias=$D\\x00D\\x00D\\x00
+$bytes_decode_alias=base64:RQBFAEUA
+
+$*
```
trufae, line 86: also test 'cat $alias'

trufae (changes requested):

trufae (approved):

PR #19162, opened 2021-10-04
Improve alias string detection

Lazula:
**Checklist**

**Description**

Alias string detection now considers presence of non-ASCII characters in addition to null byte locations when determining if data is an unterminated string. e.g. `\x90\x90\x90` is no longer incorrectly considered an unterminated string. Didn't get picked up by other tests because they contained nulls, which caused the string detection to recognize that they were bytes.

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -3723,6 +3723,7 @@ escape_pipe:
 			// pipe to alias variable
 			// register output of command as an alias
 
+			r_config_set_i (core->config, "scr.color", COLOR_MODE_DISABLED);
 			RBuffer *cmd_out = r_core_cmd_tobuf (core, cmd);
 			int alias_len;
 			ut8 *alias_data = r_buf_read_all (cmd_out, &alias_len);
--- a/libr/core/cmd_api.c
+++ b/libr/core/cmd_api.c
@@ -262,28 +262,36 @@ R_API int r_cmd_alias_set_raw(RCmd *cmd, const char *k, const ut8 *v, int sz) {
 
 	/* If it's a string already, we speed things up later by checking now */
 	const ut8 *firstnull = NULL;
+	bool is_binary = false;
 	for (i = 0; i < sz; i++) {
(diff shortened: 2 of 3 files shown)


libr/core/cmd.c
```diff
@@ -3723,6 +3723,7 @@ static int r_core_cmd_subst_i(RCore *core, char *cmd, char *colon, bool *tmpseek
 			// pipe to alias variable
 			// register output of command as an alias
 
+			r_config_set_i (core->config, "scr.color", COLOR_MODE_DISABLED);
```
trufae, line 3726: uhm, better save that value and restore it later

libr/core/cmd_api.c
```diff
@@ -262,18 +262,25 @@ R_API int r_cmd_alias_set_raw(RCmd *cmd, const char *k, const ut8 *v, int sz) {
 
 	/* If it's a string already, we speed things up later by checking now */
 	const ut8 *firstnull = NULL;
+	bool has_nonascii = false;
```
trufae, line 265: what about `is_binary`

libr/core/cmd_api.c
```diff
@@ -262,18 +262,25 @@ R_API int r_cmd_alias_set_raw(RCmd *cmd, const char *k, const ut8 *v, int sz) {
...
 
-	if (firstnull && firstnull == &v[sz-1]) {
+	if (firstnull == &v[sz-1] && !has_nonascii) {
 		/* Data is already a string */
 		val->is_str = true;
-	} else if (!firstnull) {
+	} else if (!firstnull && !has_nonascii) {
 		/* Data is an unterminated string */
 		val->sz++;
 		val->data = realloc(val->data, val->sz);
```
trufae, line 286:
```suggestion
		ut8 *data = realloc (val->data, val->sz);
		if (!data) {
				return 0;
		}
		val->data = data;
```

trufae (changes requested):

PR #19165, opened 2021-10-04
Add .rvc_ignore

RHL120:
**Checklist**

- [ ] Closing issues: #issue

**Description**
Adds an .rvc_ignore which is like .gitignore but doesn't work with direcotries.

### DIFF
--- a/libr/core/rvc.c
+++ b/libr/core/rvc.c
@@ -10,6 +10,7 @@
 #define COMMIT_BLOB_SEP "----"
 #define DBNAME "branches.sdb"
 #define CURRENTB "current_branch"
+#define IGNORE_NAME ".rvc_ignore"
 #define MAX_MESSAGE_LEN 80
 #define NULLVAL "-"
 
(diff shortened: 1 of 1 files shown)


libr/core/rvc.c
```diff
@@ -227,9 +228,26 @@ static RList *get_commits(const char *rp, const size_t max_num) {
...
 
-static bool update_blobs(RList *blobs, const RList *nh) {
+
+static bool in_rvc_ignore(const char *rp, const char *rpf) {
+	char *ignorep = r_str_newf ("%s" R_SYS_DIR IGNORE_NAME, rp);
+	if (ignorep) {
+		char *content = r_file_slurp (ignorep, 0);
+		free (ignorep);
+		if (content) {
+			bool ret = strstr (content, rpf);
```
trufae, line 238: Free content

trufae: Better cache the contents in rvector or rlist . Use rstr.split and just check for the given string not for strstr which may result on false positives

libr/core/rvc.c
```diff
@@ -227,9 +228,26 @@ static RList *get_commits(const char *rp, const size_t max_num) {
 	return ret;
 }
 
-static bool update_blobs(RList *blobs, const RList *nh) {
+
+static bool in_rvc_ignore(const char *rp, const char *rpf) {
+	char *ignorep = r_str_newf ("%s" R_SYS_DIR IGNORE_NAME, rp);
```
trufae, line 233: Didnt u wanted to use the rpath api?

trufae (changes requested):

libr/core/rvc.c
RHL120: I don't understand why you would want to cache it (in memory) in_rvc_ignore is only ran once per ravc2 command

libr/core/rvc.c
RHL120: It didn't work for some reason appending weird strings to the end

trufae: Last arg must be NULL

PR #19178, opened 2021-10-06
Fix a Segfault during aaft

thymol0:
Steps to reproduce the Segfault:
```sh
$ echo 'main(){}' >m.c
$ gcc -o m m.c
$ r2 -e bin.cache=1 -e anal.detectwrites=1 -A m
[x] Finding and parsing C++ vtables (avrr)
segmentation fault
```
Environment:
OS: `ubuntu-18.04 docker image`
GCC: `gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)`

**Checklist**

- [ ] Closing issues: #issue


trufae:
Uhm. I dont see where it is modified inside this foreach to be the reason for safe to be used.

trufae:
Can you add a test? may be good to see the backtrace from asan too

trufae:
Yes please, provide the reproducer file at least. ive re-read the function a couple of times but maybe im blind and i dont see any reference to `it` anywhere. i mean it doesnt hurt to use that _safe. it just makes the loop a little slower, but im trying to understand the root cause of this for a proper fix that doesnt involves saving the iterator.

thanks!

trufae:
ill look into that today

trufae:
Found some more segfaults in the aaft code. guess it needs to be rewritten, but not for 5.5.. i wanted to do so after the release, so ill fix the bugs for now. sorry for the delay. ill come back to this next week. but anyone is welcome to help on this

radare:
I decided to implement the simplest solution which is to keep a list of the bb addresses and fetch the basic block associated on each iteration, this way its able to detect that the cfg has changed and re-loops to compute the list of basic blocks again.

i used an rlist for the PoC, but rvector would work better, ill fix that and merge it.

Thanks for reporting and researching the crash details and writing the test file! I'll ping u back when i merge the PR, so you can rebase your PR with the testcase. Sounds good?

radare:
Nvm i just took your test and commited it under your name :) very nice catch! it should work fine now

PR #19186, opened 2021-10-07
Fix new defects in various files

Lazula:
Fix all the defects introduced in August or later, excluding a minor one on new experimental code for the pdr command.

* Remove unnecessary NULL check in __demangleAs()
* Remove unnecessary NULL check in r_io_fd_seek()
* Add missing break in ARM32 COFF case and fix test
* Add missing NULL checks for r_io_bank_get() result
* Add missing NULL checks in r_bin_demangle()
* Add missing NULL check for r_io_submap_new() result
* Fix improper failure check in riscv_assemble()
* Add missing break in open command
* Document fall-through case in dalvik anal


condret:
thx

PR #19198, opened 2021-10-09
Prevent merging zigns on load ##signatures

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This update will prevent signatures being loaded from a file from clobbering current signatures if they have the same name. For example signature foo will be renamed foo_2.

This also causes rasign2 -S to properly respect the -m flag.


swoops:
> ##tags must be placed at the end of the commit message
Sorry, sent pull from my phone and just assumed it used the git commit message. I fixed it I think.

PR #19353, opened 2021-11-01
Add more information to var ##signatures

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Work in progress, don't merge just yet. I need to fix tests.

The purpose of this pull is to add more information to the vars signature type. Previously vars included the `RAnalVarKind` and stack offset. This update will add the variable name and type to the information (see new `RSignVar` type). Should I add any other info from `RAnalVar`? 

Before I fix tests, I wanted to double check this syntax and serialization look good. If a user names a variable `foo:bar` then the would break the serialization. So currently the code rejects such a variable name. Variable names, with current serialization, can't have a `:` or `,` in them. Types for variables can not currently contain `,`. If you prefer a different serialization method, or seperator character[s], let me know (maybe `@` instead of `:`?). 

New serialized set of vars might look like:
```
za main v b-32:src:char *, b-24:s2:char *, b-40:var_20h:size_t, b-36:size:size_t, b-16:dest:char *
```
This token `b-32:src:char *` implies a variable named `src` that is -32 bytes from  `b`p with type `char *`.

Let me know if you have any better ideas, want more or less data in vars, etc. I am more then happy to try something else.


trufae:
I dont think , should be a valid name for vars

radare:
one of the problem i see in this serialization is that the type can contain special cahrs that can make us depend on the "" quoting to avoid command injection issues.we can filter out those chars too, they shouldnt be valid for types or varnames i think

swoops:
This code is now based off of the serialization API for vars. Still need to fix tests.

> one of the problem i see in this serialization is that the type can contain special cahrs that can make us depend on the "" quoting to avoid command injection issues.we can filter out those chars too, they shouldnt be valid for types or varnames i think

The `sanitize_var_serial` function in var.c will replace characters ``$`{}~|#@&<>,:`` with `_` in the var name, the var type will allow `:` but none of the others.

swoops:
I think this is good now, assuming tests pass and you are happy with all the code.

I left all the commits separate just so you can audit them, feel free to squish it into one. Or I can if you would like.

radare:
I can squash them. lets wait for the ci

swoops:
Hmm test was broken, give ma minute.

libr/anal/sign.c
```diff
@@ -1572,6 +1561,56 @@ static void listOffset(RAnal *a, RSignItem *it, PJ *pj, int format) {
 	}
 }
 
+static void inline list_vars_abs(RAnal *a, RSignItem *it, bool rad) {
+	if (it->vars && !r_list_empty (it->vars)) {
+		char *ser = r_anal_var_prot_serialize (it->vars, true);
+		if (ser) {
+			if (rad) {
+				a->cb_printf ("za %s %c %s\n", it->name, R_SIGN_VARS, ser);
```
radare, line 1569: filter it->name before printing it, same goes for 'ser'

libr/anal/sign.c
```diff
@@ -1572,6 +1561,56 @@ static void listOffset(RAnal *a, RSignItem *it, PJ *pj, int format) {
...
+			} else {
+				a->cb_printf ("  vars: %s\n", ser);
```
radare, line 1571: same

libr/anal/sign.c
```diff
@@ -1572,6 +1561,56 @@ static void listOffset(RAnal *a, RSignItem *it, PJ *pj, int format) {
...
+}
+
+static void inline list_vars(RAnal *a, RSignItem *it, PJ *pj, int fmt) {
+	switch (fmt) {
+	case '*':
+		list_vars_abs (a, it, true);
+		break;
+	case 'q':
+		if (it->vars) {
+			a->cb_printf (" vars(%d)", r_list_length (it->vars));
```
radare, line 1602: maybe better to use brackets? `vars[%d]`

libr/anal/sign.c
```diff
@@ -2192,6 +2226,37 @@ static int sig_hash_cmp(RSignHash *a, RSignHash *b) {
...
+	int len = r_list_length (la);
+	int dif = len - r_list_length (lb);
+	if (dif) {
+		return dif;
+	}
+	size_t i;
+	for (i = 0; i < len; i++) {
+		const RAnalVarProt *a = r_list_get_n (la, i);
+		const RAnalVarProt *b = r_list_get_n (lb, i);
+		dif = a->delta - b->delta;
```
radare, line 2245: can a or b be null :?

radare (changes requested):

libr/anal/sign.c
```diff
@@ -89,7 +89,7 @@ R_API RList *r_sign_fcn_refs(RAnal *a, RAnalFunction *fcn) {
 	return ret;
 }
 
-static inline *r_sign_vars(RAnalFunction *fcn) {
+static inline RList *r_sign_vars(RAnalFunction *fcn) {
```
radare, line 92: if it's static don't name it with the `r_` prefix

libr/anal/sign.c
swoops: I don't think so, but better to check then crash. I will update it to warn in that case too.

radare:
one of the unit tests is crashing

libr/anal/sign.c
```diff
@@ -1480,6 +1469,35 @@ struct ctxListCB {
 	PJ *pj;
 };
 
+static void list_sanitise_warn(char *s, const char *name, const char *field) {
```
radare, line 1472: see r_name_filter() and such

libr/anal/sign.c
```diff
@@ -1502,6 +1520,8 @@ static void listBytes(RAnal *a, RSignItem *it, PJ *pj, int format) {
 		return;
 	}
 
+	list_sanitise_warn (strbytes, it->name, "bytes");
+	list_sanitise_warn (strmask, it->name, "mask");
 	if (format == '*') {
 		if (masked == bytes->size) {
 			a->cb_printf ("za %s b %s\n", it->name, strbytes);
```
radare, line 1527: command injection if no name filtered

radare (changes requested):

radare:
Please rebase the PR on top of master, it should run a little bit faster

libr/anal/sign.c
swoops: I will update to use `r_name_filter` on `it->name` and `it->space->name`.

I will use this `list_sanitize_warn` for sig types, like bytes, bbhash and vars. This is because `r_name_filter` does not allow names to start with a number.

In the future I can add a `RAsignItem` validator or sanitizer so it will all be done in one place. Does that sound good?

libr/anal/sign.c
swoops: Name is sanitized in `listCB`, each of these little functions uses name. I don't like sanitizing the name so far away from the output sink, but I also don't like re-sanitizing the name for every signature type. I will do whatever you think is better.

PR #19374, opened 2021-11-04
Simplify deserialization of byte ##signatures

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This is two commits. One creates `r_hex_str2bin_until_new` see the unit tests for examples. Works like so:

```
ut8 *out;
char *str = "010203_NOTHEX";
int len = r_hex_str2bin_until_new (str, &out);
// len == 3
// *out == "\x010203"
// str + len * 2 == "_NOTHEX"
free (out);
```

The second commit uses the first. It simplifies the byte signature (de)serialization/parsing. It bothered me that `za` commands and sdb value deserialization used different parsing functions. Now they are put in the same place.

This also means bytes are stored in sdb files as `|b:0102030405:ffffffffff` now. Previously, serialized signatures looked like `|s:5|b:0102030405|m:ffffffffff` and required the `R_SIGN_SIZE` ('s') and `R_SIGN_MASK` ('m') enums. Those values are now free to be used for a new signature type, if desired.

This simplification should help later too. I hope to allow `z/` to deserialize just byte signatures for byte matching. This should have a modest speed increase to `z/` depending on the size of the signature database. It will large memory savings.

If you want any changes, or would rather this be split or anything else, let me know.

### DIFF
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -617,7 +617,7 @@ static bool validate_item(RSignItem *it) {
 			eprintf ("Signature '%s' has empty byte field\n", it->name);
 			return false;
 		}
-		if (b->mask[0] == '\0') {
+		if (!b->mask[0]) {
 			eprintf ("Signature '%s' mask starts empty\n", it->name);
 			return false;
 		}
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -89,6 +89,60 @@ R_API RList *r_sign_fcn_refs(RAnal *a, RAnalFunction *fcn) {
 	return ret;
 }
 
+static RSignBytes *des_bytes_norm(const char *in) {
+	// "444444:ffffff" or "44444444"
+	RSignBytes *b = R_NEW0 (RSignBytes);
+	if (b && (b->size = r_hex_str2bin_until_new (in, &b->bytes)) > 0) {
+		in += 2 * b->size;
+		if (*in == '\0' && (b->mask = malloc (b->size)) != NULL) {
+			// no mask, set it to f's
+			memset (b->mask, 0xff, b->size);
+			return b;
+		} else if (*in++ == ':') {
+			// get mask
+			int size = r_hex_str2bin_until_new (in, &b->mask);
+			in += 2 * size;
+			if (size == b->size && *in == '\0') {
+				return b;
+			}
+		}
+	}
+	r_sign_bytes_free (b);
+	return NULL;
+}
+
+static RSignBytes *deserialize_bytes(const char *in) {
+	RSignBytes *b = des_bytes_norm (in);
+	if (b) {
+		return b;
+	}
+
+	// "44..44.."
+	b = R_NEW0 (RSignBytes);
+	size_t len = strlen (in) + 3;
+	b->bytes = malloc (len);
+	b->mask = malloc (len);
+	if (b && b->bytes && b->mask) {
+		b->size = r_hex_str2binmask (in, b->bytes, b->mask);
+		if (b->size > 0) {
+			return b;
+		}
+	}
+	return NULL;
+}
+
+static RSignBytes *deserialize_anal(RAnal *a, const char *in) {
+	RSignBytes *b = R_NEW0 (RSignBytes);
+	if (b && (b->size = r_hex_str2bin_until_new (in, &b->bytes)) > 0) {
+		in += 2 * b->size;
+		if (*in == '\0' && (b->mask = r_anal_mask (a, b->size, b->bytes, 0))) {
+			return b;
+		}
+	}
+	r_sign_bytes_free (b);
+	return NULL;
+}
+
 static inline RList *sign_vars(RAnalFunction *fcn) {
 	RList *l = r_anal_var_get_prots (fcn);
 	if (l && r_list_empty (l)) {
@@ -174,7 +228,7 @@ R_API bool r_sign_deserialize(RAnal *a, RSignItem *it, const char *k, const char
 	// Deserialize value: |k:v|k:v|k:v|...
 	n = r_str_split (v2, '|');
 	const char *token = NULL;
-	int w, size;
+	int w;
 	for (w = 0; w < n; w++) {
 		const char *word = r_str_word_get0 (v2, w);
 		if (!word) {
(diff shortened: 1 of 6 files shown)


swoops:
I should of split this...

If it works out fine this time, I will leave it to your will. Otherwise I will close and split into two pulls.

libr/anal/sign.c
```diff
@@ -89,6 +89,60 @@ R_API RList *r_sign_fcn_refs(RAnal *a, RAnalFunction *fcn) {
...
+static RSignBytes *des_bytes_norm(const char *in) {
+	// "444444:ffffff" or "44444444"
+	RSignBytes *b = R_NEW0 (RSignBytes);
+	if (b && (b->size = r_hex_str2bin_until_new (in, &b->bytes)) > 0) {
+		in += 2 * b->size;
+		if (*in == '\0' && (b->mask = malloc (b->size)) != NULL) {
+			// no mask, set it to f's
+			memset (b->mask, 0xff, b->size);
+			return b;
+		} else if (*in++ == ':') {
```
trufae, line 101: no need to else after return

libr/anal/sign.c
```diff
@@ -89,6 +89,60 @@ R_API RList *r_sign_fcn_refs(RAnal *a, RAnalFunction *fcn) {
 	return ret;
 }
 
+static RSignBytes *des_bytes_norm(const char *in) {
+	// "444444:ffffff" or "44444444"
+	RSignBytes *b = R_NEW0 (RSignBytes);
+	if (b && (b->size = r_hex_str2bin_until_new (in, &b->bytes)) > 0) {
+		in += 2 * b->size;
+		if (*in == '\0' && (b->mask = malloc (b->size)) != NULL) {
```
trufae, line 97: use !*in and != NULL is implicit

libr/anal/sign.c
```diff
@@ -89,6 +89,60 @@ R_API RList *r_sign_fcn_refs(RAnal *a, RAnalFunction *fcn) {
...
+			// no mask, set it to f's
+			memset (b->mask, 0xff, b->size);
+			return b;
+		} else if (*in++ == ':') {
+			// get mask
+			int size = r_hex_str2bin_until_new (in, &b->mask);
+			in += 2 * size;
+			if (size == b->size && *in == '\0') {
```
trufae, line 105: same !*in

libr/anal/sign.c
```diff
@@ -89,6 +89,60 @@ R_API RList *r_sign_fcn_refs(RAnal *a, RAnalFunction *fcn) {
...
+	size_t len = strlen (in) + 3;
+	b->bytes = malloc (len);
+	b->mask = malloc (len);
+	if (b && b->bytes && b->mask) {
+		b->size = r_hex_str2binmask (in, b->bytes, b->mask);
+		if (b->size > 0) {
+			return b;
+		}
+	}
+	return NULL;
```
trufae, line 131: leaks if bytes or mask is null

trufae: i mean 'xor'

libr/anal/sign.c
```diff
@@ -89,6 +89,60 @@ R_API RList *r_sign_fcn_refs(RAnal *a, RAnalFunction *fcn) {
...
+}
+
+static RSignBytes *deserialize_anal(RAnal *a, const char *in) {
+	RSignBytes *b = R_NEW0 (RSignBytes);
+	if (b && (b->size = r_hex_str2bin_until_new (in, &b->bytes)) > 0) {
+		in += 2 * b->size;
+		if (*in == '\0' && (b->mask = r_anal_mask (a, b->size, b->bytes, 0))) {
```
trufae, line 138: same !*in

and im not a big fan of assignments in if statements, because that leads to confision and iirc some compilers will warn on that

```suggestion
		if (!*in && (b->mask = r_anal_mask (a, b->size, b->bytes, 0))) {
```

libr/anal/sign.c
```diff
@@ -588,7 +599,36 @@ R_API RSignItem *r_sign_get_item(RAnal *a, const char *name) {
...
+		return false;
+	}
+
+	if (it->bytes) {
+		RSignBytes *b = it->bytes;
+		if (!b->mask || !b->bytes || b->size <= 0) {
+			eprintf ("Signature '%s' has empty byte field\n", it->name);
+			return false;
+		}
+		if (b->mask[0] == '\0') {
```
trufae, line 620:
```suggestion
		if (!b->mask[0]) {
```

libr/anal/sign.c
```diff
@@ -588,7 +599,36 @@ R_API RSignItem *r_sign_get_item(RAnal *a, const char *name) {
...
+	}
+	return true;
+}
+
 R_API bool r_sign_add_item(RAnal *a, RSignItem *it) {
+	if (!validate_item (it)) {
+		return false;
+	}
 	char *key = item_serialize_key (it);
 	RSignItem *current = sign_get_sdb_item (a, key);
```
trufae, line 633: kcan key be null?

libr/util/hex.c
```diff
@@ -450,6 +450,31 @@ R_API int r_hex_str2bin(const char *in, ut8 *out) {
 	return nibbles / 2;
 }
 
+// get the hex chars from start of string, until first non-hex char, as a heap
+// allocated ut8* buffer
+R_API int r_hex_str2bin_until_new(const char *in, ut8 **out) {
+	r_return_val_if_fail (in && out && !*out, -1);
+	size_t len = strlen (in);
+	if (len <= 1) {
```
trufae, line 458: this only happens if the string is empty. so no need to strlen in this case, do u think that *in should be checked in the return_if_fail statement?

also, there's R_STR_ISEMPTY()) and R_STR_ISNOTEMPTY() macros to help ont hat

libr/util/hex.c
```diff
@@ -450,6 +450,31 @@ R_API int r_hex_str2bin(const char *in, ut8 *out) {
...
+		return 0;
+	}
+	len = (len + 1) / 2;
+
+	ut8 *buf = NULL;
+	size_t nibbles = 0;
+	if ((buf = malloc (len)) != NULL) {
+		while (!r_hex_to_byte (&buf[nibbles / 2], *in)) {
```
trufae, line 466: i prefer the buf + (nibbles/2) syntax instead of the &a[b] one, looks easier to read to me

libr/util/hex.c
```diff
@@ -450,6 +450,31 @@ R_API int r_hex_str2bin(const char *in, ut8 *out) {
...
+	r_return_val_if_fail (in && out && !*out, -1);
+	size_t len = strlen (in);
+	if (len <= 1) {
+		return 0;
+	}
+	len = (len + 1) / 2;
+
+	ut8 *buf = NULL;
+	size_t nibbles = 0;
+	if ((buf = malloc (len)) != NULL) {
```
trufae, line 465: dont assign in if statements and != NULL is implicit

trufae: you can define and assign in the same line `ut8 *buf = mallloc(len); if (!buf) {...}`

trufae (changes requested):
even with units! please check the comments! lgtm so far

libr/util/hex.c
swoops: I think the way it is makes life a bit easier on the caller. I might misunderstand you though.

I think `in == NULL` is indicative of a programmer error and so `r_return_val_if_fail`. The case `*out != NULL` is bad because the caller should be able to free `*out` regardless of success/failure/error. This means the caller does not have to handle as many edge cases. I could see a case for `out == NULL` being accepted, just to let the function count how many bytes of hex it would parse.

I think if `in` is "", "4", "ZZZZ" or "4Z" then the function should return 0 because it did not parse any hex from the start of the string. This shouldn't be a programing error. This input will happen when trying to get a byte mask from the invalid signatures "11:", "11:4", "11:ZZZZ" and "11:4Z". So the caller has less edge cases to handle.

So input strings of length 1 or 0 should return 0, since neither can have any hex *bytes* parsed. This function can't parse nibbles otherwise the caller wouldn't be able to know where the parser stopped in the string (via: `in + (2 * return_code)`)

libr/anal/sign.c
```diff
@@ -426,9 +440,8 @@ static char *serialize_value(RSignItem *it) {
 		if (hexbytes && hexmask) {
 			r_hex_bin2str (bytes->bytes, bytes->size, hexbytes);
 			r_hex_bin2str (bytes->mask, bytes->size, hexmask);
-			success = r_strbuf_appendf (sb, "|%c:%d|%c:%s|%c:%s",
-				R_SIGN_BYTES_SIZE, bytes->size, R_SIGN_BYTES, hexbytes,
-				R_SIGN_BYTES_MASK, hexmask);
+			success = r_strbuf_appendf (sb, "|%c:%s:%s", R_SIGN_BYTES,
```
trufae, line 443: if hexbytes is null and hexmask is not then its leaking

swoops:
All the changes to `r_hex_str2bin_until_new` are in the "r_hex_str2bin_until_new fixes" commit.

Please review that again, I ended up reworking it a bit. I decided I was being too tricky with the `nibbles` variable. I also realized if `malloc` failed I would return 0 instead of -1. So I just made everything simpler.

I also fixed some tests. Hopefully CI likes it better.

If it looks good, I can squish it for you. Probably should be two commits.

trufae (commented):
Those comments are unrelated to your current PR, so lgt merge.

The `FreeRet_on_fail` is a bad name, should be all snake case.

PR #19381, opened 2021-11-05
Fix afv outputing commands when not in "*" mode

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This should fix the failing "CI / linux-test (pull_request) " failing on "var rebase" test.

I think this is a quick way to fix it. I don't know this code well so I am not completely convinced it is right.

I have two pulls failing on this test, so if you want to merge this first (assuming it's good), then I will rebase those pulls to make sure they pass.


swoops:
This also fixes `afvj` producing invalid JSON, which causes r2pipe to fail.

Issue #19227, opened 2021-10-14
Add support for jdh-8


condret:
https://www.youtube.com/watch?v=7A1SzIIKMho

asm, anal and esil support would be great

trufae:
any doc about the bin format?

trufae:
totally low prio and imho should be done in extras, removed from milestone

trufae:
open the ticket in extras this doesnt seems like an useful arch to have in core

PR #19412, opened 2021-11-15
Jdh-8 asm support  #19227

slowhand99:
**Checklist**

- [ ] Closing issues: #issue

**Description**
##asm

Support disassembly for JDH-8 toy architecture

### DIFF
--- a/libr/asm/arch/jdh8/jdh8dis.c
+++ b/libr/asm/arch/jdh8/jdh8dis.c
@@ -1,4 +1,4 @@
-/* radare - <TODO> - Copyright 2021 - <TODO> */
+/* radare - LGPL - Copyright 2021 - condret, slowhand99 */
 
 #include <r_types.h>
 #include <r_util.h>
@@ -138,12 +138,3 @@ static int jdh8Disass(RAsmOp *op, const ut8 *buf, int len) {
 	r_strbuf_set (&op->buf_asm, buf_asm);
 	return op->size = rlen;
 }
-/*static int jdh8Disass(RAsmOp *op, const ut8 *buf, int len) {
-	if (len < 1) {
-		return 0;
-	}
-	
-	r_strbuf_setf (&op->buf_asm, "unknown(0x%02x)", buf[0]);
-	return 1;
-}
-*/
--- a/libr/asm/arch/jdh8/jdh8dis.c
+++ b/libr/asm/arch/jdh8/jdh8dis.c
@@ -63,7 +63,7 @@ static int  jdh8Disass (RAsmOp *op, const ut8 *buf, int len) {
 		if (rlen == 1) {
 			buf_asm = sdb_fmt ("jnz %c", reg[(low & 7)); break;
 		else {
-			buf_asm = sdb_fmt ("jnz 0x%02d", buf[1]); break;
+			buf_asm = sdb_fmt ("jnz 0x%02x", buf[1]); break;
 		}
 	case 7: 
 		if ( low & 8) {
--- a/libr/asm/arch/jdh8/jdh8dis.c
+++ b/libr/asm/arch/jdh8/jdh8dis.c
@@ -11,9 +11,9 @@ static const int jdh_len[16] = {
 static const char reg[6] = {'A', 'B', 'C', 'D', 'L', 'H'};
 
 
-static int i4004_get_ins_len (ut8 hex) {
+static int jdh_get_ins_len (ut8 hex) {
 	ut8 high = (hex & 0xf0)>>4;
-	int ret = i4004_ins_len[high];
+	int ret = jdh_len[high];
 	if (ret == 3)
 		ret = (hex & 8) ? 2 : 3;
 	if (ret == 1)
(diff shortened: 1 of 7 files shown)


trufae:
Can you add it to the meson too?

libr/asm/arch/jdh8/jdh8dis.c
```diff
@@ -0,0 +1,140 @@
...
+};
+
+static const char reg[6] = { 'A', 'B', 'C', 'D', 'L', 'H' };
+
+static int jdh_get_ins_len(ut8 hex) {
+	ut8 high = (hex & 0xf0) >> 4;
+	int ret = jdh_len[high];
+	if (ret == 3) {
+		ret = (hex & 8) ? 2 : 3;
+	}
```
trufae, line 18: add an "else", after the first if, the second will never be reached

libr/asm/arch/jdh8/jdh8dis.c
```diff
@@ -0,0 +1,140 @@
...
+	if (ret == 1) {
+		ret = (hex & 8) ? 1 : 2;
+	}
+	return ret;
+}
+
+static int jdh8Disass(RAsmOp *op, const ut8 *buf, int len) {
+	int rlen = jdh_get_ins_len (*buf);
+	ut8 high = (*buf & 0xf0) >> 4;
```
trufae, line 27:
```suggestion
	const ut8 high = (*buf & 0xf0) >> 4;
```

libr/asm/arch/jdh8/jdh8dis.c
```diff
@@ -0,0 +1,140 @@
...
+	ut8 low = (*buf & 0xf);
+	const char *buf_asm = "invalid";
+	if (rlen > len) {
+		return op->size = 0;
+	}
+	switch (high) {
+	case 0:
+		buf_asm = (low & 8)? sdb_fmt ("mw %c, %c", reg[low & 7], reg[buf[1]]):
```
trufae, line 35: sdb_fmt is not thread safe and we are trying to get rid of it across all the r2codebase, there are alternatives like snprintf or r_str_newf in case it must be held of the heap. in this case it would work better to have a local buffer of 64 bytes in this function and use snprintf like this:

```
char buf[64];
buf_asm = snprintf (buf, sizeof (buf), "mw %c, %c", ...);
..
```

trufae: you can add a #define to avoid having to pass buf and sizeof(buf) all the time so its easier for the eye 😄

libr/asm/arch/jdh8/jdh8dis.c
```diff
@@ -0,0 +1,140 @@
...
+		ret = (hex & 8) ? 2 : 3;
+	}
+	if (ret == 1) {
+		ret = (hex & 8) ? 1 : 2;
+	}
+	return ret;
+}
+
+static int jdh8Disass(RAsmOp *op, const ut8 *buf, int len) {
+	int rlen = jdh_get_ins_len (*buf);
```
trufae, line 26: why `rlen`? maybe `ilen` works better as a name as it stands for `instruction length`

trufae (changes requested):
as a suggestion, i think it will be better to make it as an RAnalPlugin, because you can now use analysis plugins from the RAsm api, and reuse the same code when you implement the code analysis information for it. (reg profile, instruction info, etc)

slowhand99:
On it...

libr/asm/arch/jdh8/jdh8dis.c
slowhand99: > #define to avoid having to pass buf and sizeof(buf)

How? Can you give an example?

trufae: but dont u think is better to rewrite the plugin as an analysis one so it doesnt needs to duplicate the logic for the analysis? also, are there plans to implement the assembler fr this arch too?

trufae (changes requested):
ill merge when the meson.build references it

trufae (changes requested):
Can you add this disassembler for the meson builds and add a test? see libr/asm/meson.build and for the tests will be good to have a smple file for this architecture and post it in the radare2-testbins repo. is the bin format supported? because thats probably something to do too.

Im fine to merge this asm plugin, but it should be redone as an analysis plugin to avoid duplication of parsing logic as long as asm plugins do use anal ones when needed

trufae:
Looks good to merge now! thank you! ill wait for the CI :)

libr/asm/meson.build
```diff
@@ -78,7 +78,7 @@ r_asm_sources = [
   join_paths('p','asm_xcore_cs.c'),
   # join_paths('p','asm_xtensa.c'),
   join_paths('p','asm_z80.c'),
-  join_paths('p','asm_jdh8.c'),
+  #join_paths('p','asm_jdh8.c'),
```
trufae, line 81: uhm why remove it?

slowhand99: Sorry mb... Pushed the previous commit

condret:
Thanks!

Please use meaningful commit messages next time.

PR #19417, opened 2021-11-17
Show prev nodes in graph.few ##graph

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This pull does two things. It upgrades `graph.few` to display the nodes that reference the current node (see changes to `cur_points_to_bb`). 

It also fixes a minor bug in `graph.few`. Changing nodes with tab did not refresh the nodes on the screen. I refactored the code a bit so that any change to seek when `graph.few` was enabled would result in a refresh.


swoops:
Looks like I missed `u` and `U` refreshing somehow... Give me a bit.

swoops:
Using `u` or `U` now works, but it does not re-focus on current node. Hitting `.` fixes it.

PR #19418, opened 2021-11-17
Fix NULL node warning in VV ##graph

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

I am not very familiar with this code, so please double check me. 

I had `r_agraph_add_edge` warning because `v` was `NULL`.  If `v` being `NULL` is an error, then an `else` should be added to the code to warn. Otherwise, i think it is good to merge.

### DIFF
--- a/libr/core/agraph.c
+++ b/libr/core/agraph.c
@@ -2315,16 +2315,11 @@ static void add_child(RCore *core, RAGraph *g, RANode *u, ut64 jump) {
 	RANode *v = r_agraph_get_node (g, title);
 	if (v) {
 		ut64 a = r_num_get (NULL, u->title);
-		int len = 64;
-		char key[len];
+		char key[64];
 		char *fmt = "agraph.edge.0x%" PFMT64x "_0x%" PFMT64x ".highlight";
-		if (len > snprintf (key, len, fmt, a, jump)) {
-			bool hl = sdb_exists (core->sdb, key);
-			r_agraph_add_edge (g, u, v, hl);
-		} else {
-			// shouldn't overflow but implies len is wrong
-			r_warn_if_reached ();
-		}
+		snprintf (key, sizeof (key), fmt, a, jump);
+		bool hl = sdb_exists (core->sdb, key);
+		r_agraph_add_edge (g, u, v, hl);
 	} else {
 		eprintf ("[!!] Failed to add child node 0x%" PFMT64x " to %s. Child not found\n", jump, u->title);
 	}
--- a/libr/core/agraph.c
+++ b/libr/core/agraph.c
@@ -2315,10 +2315,16 @@ static void add_child(RCore *core, RAGraph *g, RANode *u, ut64 jump) {
 	RANode *v = r_agraph_get_node (g, title);
 	if (v) {
 		ut64 a = r_num_get (NULL, u->title);
-		ut64 b = r_num_get (NULL, title);
-		const char *k = sdb_fmt ("agraph.edge.0x%" PFMT64x "_0x%" PFMT64x ".highlight", a, b);
-		bool hl = sdb_exists (core->sdb, k);
-		r_agraph_add_edge (g, u, v, hl);
+		int len = 64;
+		char key[len];
+		char *fmt = "agraph.edge.0x%" PFMT64x "_0x%" PFMT64x ".highlight";
+		if (len > snprintf (key, len, fmt, a, jump)) {
+			bool hl = sdb_exists (core->sdb, key);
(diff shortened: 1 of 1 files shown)


trufae:
Maybe worth adding an eprintf for that case?

libr/core/agraph.c
```diff
@@ -2310,12 +2310,14 @@ static void add_child(RCore *core, RAGraph *g, RANode *u, ut64 jump) {
 	}
 	char *title = get_title (jump);
 	RANode *v = r_agraph_get_node (g, title);
-	ut64 a = r_num_get (NULL, u->title);
-	ut64 b = r_num_get (NULL, title);
+	if (v) {
+		ut64 a = r_num_get (NULL, u->title);
+		ut64 b = r_num_get (NULL, title);
+		const char *k = sdb_fmt ("agraph.edge.0x%" PFMT64x "_0x%" PFMT64x ".highlight", a, b);
```
trufae, line 2316: better not to use sdb_fmt as its not thread safe, we are in the plan to remove its uses and at some point have a simpler string-in-stack api like we did 2 years ago without reaching any agreement. but it will definitvely make the code cleaner.

trufae (changes requested):
f

libr/core/agraph.c
swoops: What should I use instead? Just `r_str_newf` a string that would of been returned by sdb_fmt? 

I am not too familiar with the inner workings of sdb.

trufae: Use snprintf for now. Strnew uses the heap and there was 3 r_strf proposals that all of them ens up deleted. But will be good to bring one of them

libr/core/agraph.c
```diff
@@ -2310,12 +2310,22 @@ static void add_child(RCore *core, RAGraph *g, RANode *u, ut64 jump) {
 	}
 	char *title = get_title (jump);
 	RANode *v = r_agraph_get_node (g, title);
-	ut64 a = r_num_get (NULL, u->title);
-	ut64 b = r_num_get (NULL, title);
+	if (v) {
+		ut64 a = r_num_get (NULL, u->title);
+		int len = 64;
+		char key[len];
```
trufae, line 2316: this is a variable size array, which is bad in many ways:

* not all c compilers support this
* can cause stack exhaustion if the value is big (in this case is constant, but the compiler doesnt needs to know that)

char do char key[64] and use sizeof(key) instead of len, this way the size of the array is always consistent with the array itself. not a separate variable

libr/core/agraph.c
```diff
@@ -2310,12 +2310,22 @@ static void add_child(RCore *core, RAGraph *g, RANode *u, ut64 jump) {
...
+		char *fmt = "agraph.edge.0x%" PFMT64x "_0x%" PFMT64x ".highlight";
+		if (len > snprintf (key, len, fmt, a, jump)) {
```
trufae, line 2318: that snprintf should never fail. you know before hand how long this string will be (2 64bit numbers in hexa are 10 chars, + a bunch of text ... so you can skip this check

libr/core/agraph.c
```diff
@@ -2310,12 +2310,22 @@ static void add_child(RCore *core, RAGraph *g, RANode *u, ut64 jump) {
...
+			bool hl = sdb_exists (core->sdb, key);
+			r_agraph_add_edge (g, u, v, hl);
+		} else {
+			// shouldn't overflow but implies len is wrong
+			r_warn_if_reached ();
```
trufae, line 2323: remove this, at compile time u know how big the string will be

trufae (changes requested):

libr/core/agraph.c
swoops: Thanks, I did not know that. I have updated the code.

libr/core/agraph.c
```diff
@@ -2313,12 +2313,17 @@ static void add_child(RCore *core, RAGraph *g, RANode *u, ut64 jump) {
 	}
 	char *title = get_title (jump);
 	RANode *v = r_agraph_get_node (g, title);
-	ut64 a = r_num_get (NULL, u->title);
-	ut64 b = r_num_get (NULL, title);
+	if (v) {
+		ut64 a = r_num_get (NULL, u->title);
+		char key[64];
+		char *fmt = "agraph.edge.0x%" PFMT64x "_0x%" PFMT64x ".highlight";
```
trufae, line 2319:
```suggestion
		const char *fmt = "agraph.edge.0x%" PFMT64x "_0x%" PFMT64x ".highlight";
```

trufae (approved):

PR #19451, opened 2021-11-29
Fix #19436  - DoS analysing ELF64 binary for MIPS architecture ##analysis

### LINKED ISSUE #19436: DoS analysing ELF64 binary for MIPS architecture
## Environment
```
$ date
mar 23 nov 2021 10:31:32 -03
$ r2 -v
radare2 5.5.0 1 @ linux-x86-64 git.
commit: b50c2c35acd266f1b18bbbcfe0c63d9d0331b09d build: 2021-11-14__22:46:21
$ uname -ms
Linux x86_64
```

## Description
We found with @OctavioGalland an ELF64 binary for MIPS architecture that hangs when analysed.
We think this is caused by mapping a huge section that is interpreted as NOPs. If we modify the size of the section, the analysis doesn't hang. While this is not an infinite loop, it can be very long. And this has been acknowledged as a DoS in the past (see #18923).

## Test

```
$ base64 -d <<< f0VMRgIBAQAAAACqqqqqqqqqCgABAAAAABBAAAAAAABAAAAAAAAAACAQAAAAAAAAAAAAAEAAOAABAEAAAwACAAEAAAABAAAABIAAAAAAAAAAAAAAAABAAAAAAAAAAEAAAAAAALAAAAD//////////wAAAAAAEAAA  > hang

$ base64 -d <<< f0VMRgIBAQAAAACqqqqqqqqqCgABAAAAABBAAAAAAABAAAAAAAAAACAQAAAAAAAAAAAAAEAAOAABAEAAAwACAAEAAAABAAAABIAAAAAAAAAAAAAAAABAAAAAAAAAAEAAAAAAALAAAAABAAAAAAAAAAAAAAAAEAAA  > nohang
```

```
$ r2 ./hang
 -- Beer in mind.
[0x400000003f8ffc]> aaa
[ ] Analyze all flags starting with sym. and entry0 (aa)
```

```
$ readelf -l hang
readelf: Error: Reading 192 bytes extends past end of file for section headers
Elf file type is <unknown>: aaaa
Entry point 0x401000
There is 1 program header, starting at offset 64
Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000008004 0x0040000000000000 0x0040000000000000
                 0x000000b000000000 0xffffffffffffffff    E    0x100000000000
```
```
$ readelf -l nohang
readelf: Error: Reading 192 bytes extends past end of file for section headers
Elf file type is <unknown>: aaaa
Entry point 0x401000
There is 1 program header, starting at offset 64
Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000008004 0x0040000000000000 0x0040000000000000
                 0x000000b000000000 0x0000000000000001    E    0x100000000000
readelf: Error: the segment's file size is larger than its memory size
```
```
$ binwalk -W hang nohang
OFFSET      hang                                                                 nohang
--------------------------------------------------------------------------------
0x00000000  7F 45 4C 46 02 01 01 00 00 00 00 AA AA AA AA AA |.ELF............| \ 7F 45 4C 46 02 01 01 00 00 00 00 AA AA AA AA AA |.ELF............|
0x00000010  AA AA 0A 00 01 00 00 00 00 10 40 00 00 00 00 00 |..........@.....| / AA AA 0A 00 01 00 00 00 00 10 40 00 00 00 00 00 |..........@.....|
0x00000020  40 00 00 00 00 00 00 00 20 10 00 00 00 00 00 00 |@...............| \ 40 00 00 00 00 00 00 00 20 10 00 00 00 00 00 00 |@...............|
0x00000030  00 00 00 00 40 00 38 00 01 00 40 00 03 00 02 00 |....@.8...@.....| / 00 00 00 00 40 00 38 00 01 00 40 00 03 00 02 00 |....@.8...@.....|
0x00000040  01 00 00 00 01 00 00 00 04 80 00 00 00 00 00 00 |................| \ 01 00 00 00 01 00 00 00 04 80 00 00 00 00 00 00 |................|
0x00000050  00 00 00 00 00 00 40 00 00 00 00 00 00 00 40 00 |......@.......@.| / 00 00 00 00 00 00 40 00 00 00 00 00 00 00 40 00 |......@.......@.|
0x00000060  00 00 00 00 B0 00 00 00 FF FF FF FF FF FF FF FF |................| \ 00 00 00 00 B0 00 00 00 01 00 00 00 00 00 00 00 |................|
0x00000070  00 00 00 00 00 10 00 00 XX XX XX XX XX XX XX XX |................| / 00 00 00 00 00 10 00 00 XX XX XX XX XX XX XX XX |................|
```

ogianatiempo:
**Checklist**

- [x] Closing issues: #19436

**Description**

As suggested in https://github.com/radareorg/radare2/issues/19436#issuecomment-977865424, I added a counter to avoid analysing the function after N nops. I have done some local testing however, since I'm not fully familiarized with the project, I marked it as WIP.

### DIFF
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -27,6 +27,9 @@
 // 256KB max function size
 #define MAX_FCN_SIZE (1024 * 256)
 
+// Max NOP count to stop analysis
+#define MAX_NOP_PREFIX_CNT 1024
+
 #define DB a->sdb_fcns
 #define EXISTS(x, ...) snprintf (key, sizeof (key) - 1, x, ## __VA_ARGS__), sdb_exists (DB, key)
 #define SETKEY(x, ...) snprintf (key, sizeof (key) - 1, x, ## __VA_ARGS__);
@@ -548,6 +551,7 @@ static int fcn_recurse(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 len, int
 	bool overlapped = false;
 	int oplen, idx = 0;
 	size_t lea_cnt = 0;
+	size_t nop_prefix_cnt = 0;
 	static ut64 cmpval = UT64_MAX; // inherited across functions, otherwise it breaks :?
 	bool varset = false;
 	struct {
@@ -725,6 +729,10 @@ repeat:
 			case R_ANAL_OP_TYPE_TRAP:
 			case R_ANAL_OP_TYPE_ILL:
 			case R_ANAL_OP_TYPE_NOP:
+				nop_prefix_cnt++;
+				if (nop_prefix_cnt > MAX_NOP_PREFIX_CNT) {
+					gotoBeach (R_ANAL_RET_ERROR);
+				}
 				if (r_anal_block_relocate (bb, at + op->size, bb->size)) {
 					addr = at + op->size;
 					fcn->addr = addr;


trufae:
I was about to take a look at this issue today, so it's great to see you going ahead and submitting a fix proposal for this issue. Thanks!

The patch looks good, but i see some things to change/improve:

* i think 1024 nops in a single function is more than a reason to stop analyzing it, but i think this should be restricted to consecutive nops. so basically resetting this counter when !nop is found
* Also, i assume nops wont be splitting a basic block, so maybe a better approach is to consider avoiding basic blocks larger than a specific size (no matter which instruction is in there)

My suggestion for the 2nd point is because on mips null bytes are nops, but on other archs that could be anohter non-exception instruction. So it's probably worth considering this is not a mips specific issue.

Also, there's a maxfcnsize, just above your maxnopsinfcn variable. Which makes me think that is only used when a basic block is analyzed completely and registered as part of a function, so, what about reusing the same variable to check for instructions in basicblock?

Thanks for trying it out, reporting and talking to redhat to handle the security issue.

I plan to release 5.5.2 in a week, so this fix, as well as some other CVEs reported recently will be addressed in there.

ogianatiempo:
Thanks for taking time for reviewing this PR and making suggestions. I would like use this opportunity to also thank you for mantaining this awesome project.

About the first change/improvment: I think `nop_prefix_cnt` will only be increased when NOPs are found at the beginning of a function because of the condition on line 717: `if (anal->opt.nopskip && fcn->addr == at) {`. So this forces NOPs to be consecutive because if other type of instruction is found, then `at` and `fcn->addr` will differ.

Regarding the use of `MAX_FCN_SIZE` and discarding large basic blocks: each time the size of the basic block is increased, if its greater than this constant then the analysis fails (line 765). So I think this might already work as you suggested.

trufae:
Your approach looks like a good metric to me. So I'm merging!

Thanks for answering and taking the time to cook the fix.

PR #19453, opened 2021-11-29
Fix #19452 - Highlight grep with empty search string causes infinite loop ##util

### LINKED ISSUE #19452: Highlight grep with empty search string causes infinite loop
## Environment

No files needed. Use latest version of radare2.

## Description

Setting `scr.highlight.grep` to `true` should not cause grep column/line command to go into an infinite loop.

## Test

```sh
# first command works as expected
r2 -NN -q -c 'i~:0' -
# WARNING: this will run forever
r2 -NN -q -c 'e scr.highlight.grep=true; i~:0' -
```

richyliu:
**Checklist**

- [x] Closing issues: #19452

**Description**

Add check for empty key in `r_str_replace`, fixing an infinite loop when `scr.highlight.grep=true` and the user runs a grep command with an empty string.

### DIFF
--- a/libr/util/str.c
+++ b/libr/util/str.c
@@ -962,6 +962,9 @@ R_API char* r_str_replace(char *str, const char *key, const char *val, int g) {
 		r_str_replace_char (str, *key, *val);
 		return str;
 	}
+	if (klen == 0) {
+		return str;
+	}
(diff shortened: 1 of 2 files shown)


trufae:
Thanks!

PR #19466, opened 2021-12-04
Dalvik fix

aemmitt-ns:
**Checklist**

- [ ] Closing issues: #issue

**Description**

These changes simply fix the conditionals for dalvik analysis by just using the inequality operators (<,>,<=,...) instead of flag bits. The previous expressions were wrong, and the new ones are clearer anyway. With these changes we can now correctly ESIL emulate all static dalvik code.

### DIFF
--- a/libr/anal/p/anal_dalvik.c
+++ b/libr/anal/p/anal_dalvik.c
@@ -562,10 +562,10 @@ static int dalvik_op(RAnal *anal, RAnalOp *op, ut64 addr, const ut8 *data, int l
 	case 0x26: // filled-new-array-data
 		op->type = R_ANAL_OP_TYPE_NEW;
 		// 0x1c, 0x1f, 0x22
-		if (len > 2 && mask & R_ANAL_OP_MASK_ESIL) {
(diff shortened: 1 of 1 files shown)


aemmitt-ns:
checks are passing and I think this is all ready to go

trufae:
Yeah sorry im having a migraine day. The patch lgtm. But its on the plan to remove all the >,<,>=.. operators from esil. Flags should be enough. So i would suggest adding a test to confirm the currwnt correct behaviour doesnt breaks when switching to the flag bits thing

aemmitt-ns:
Oh ok. Personally i think we should keep the inequalities, or at least `<` from which we could derive the others easily. If we want ESIL to be easy to write in the console it would be a bad choice to remove. Its too difficult to write the correct expressions quickly with the flag ops, $c and $b are especially confusing. But i think keeping one is a good compromise.

trufae:
I implemented esil macros to support such meta instructions, they should be reimplemented, i also prefer them more readable. But internally the real opcodes should be he minimum as possible

PR #19477, opened 2021-12-07
Cleanup public API for ##search

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Trying to clean up the r_search API a bit.

A couple functions in r_earch.h were not used or were duplicates.

I moved the update functions into their own .h file to hide them from public use. They should only be used internally. I have not made this type of change in r2 before, so I probably am not doing it just right.

### DIFF
--- a/libr/include/r_search.h
+++ b/libr/include/r_search.h
@@ -115,13 +115,6 @@ R_API int r_search_range_reset(RSearch *s);
 
 R_API int r_search_bmh(const RSearchKeyword *kw, const ut64 from, const ut8 *buf, const int len, ut64 *out);
 
-// TODO: is this an internal API?
-R_API int r_search_mybinparse_update(RSearch *s, ut64 from, const ut8 *buf, int len);
-R_API int r_search_aes_update(RSearch *s, ut64 from, const ut8 *buf, int len);
-R_API int r_search_privkey_update(RSearch *s, ut64 from, const ut8 *buf, int len);
-R_API int r_search_deltakey_update(RSearch *s, ut64 from, const ut8 *buf, int len);
-R_API int r_search_strings_update(RSearch *s, ut64 from, const ut8 *buf, int len);
-R_API int r_search_regexp_update(RSearch *s, ut64 from, const ut8 *buf, int len);
 // Returns 2 if search.maxhits is reached, 0 on error, otherwise 1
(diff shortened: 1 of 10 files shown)


libr/search/aes_find.c
```diff
@@ -55,7 +55,7 @@ static bool aes128_key_test(const unsigned char *buf) {
 	return word1 && word2;
 }
 
-R_API int r_search_aes_update(RSearch *s, ut64 from, const ut8 *buf, int len) {
+int search_aes_update(RSearch *s, ut64 from, const ut8 *buf, int len) {
```
trufae, line 58: use R_IPI for internal apis (functions that are meant to be used only inside the module, static for the file-specific and R_API for the public ones. Not sure if this is documented in DEVELOPERS.md but it should be

libr/include/r_search.h
```diff
@@ -111,19 +110,7 @@ R_API void r_search_kw_reset(RSearch *s);
 R_API int r_search_range_add(RSearch *s, ut64 from, ut64 to);
 R_API int r_search_range_set(RSearch *s, ut64 from, ut64 to);
 R_API int r_search_range_reset(RSearch *s);
-R_API int r_search_set_blocksize(RSearch *s, ut32 bsize);
-
-R_API int r_search_bmh(const RSearchKeyword *kw, const ut64 from, const ut8 *buf, const int len, ut64 *out);
-
-// TODO: is this an internal API?
-R_API int r_search_mybinparse_update(RSearch *s, ut64 from, const ut8 *buf, int len);
```
trufae, line 119: maybe just use R_IPI in here :? i think its fine to make them internal. so the change lgtm.

trufae (changes requested):

swoops:
I broke something bad....

libr/include/r_search.h
```diff
@@ -16,7 +16,6 @@ enum {
 	R_SEARCH_ESIL,
 	R_SEARCH_KEYWORD,
 	R_SEARCH_REGEXP,
-	R_SEARCH_PATTERN,
```
trufae, line 19: that should route the call to `r_search_pattern`, the problem is that this 'mode' doesnt follows the same callback model. so it cant reuse the searchupdate loop. will be good to get this fixed too

PR #19487, opened 2021-12-10
Fix 8051

gogo2464:
**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/libr/asm/arch/8051/8051_ass.c
+++ b/libr/asm/arch/8051/8051_ass.c
@@ -401,34 +401,94 @@ static bool address_bit(char const* addr_str, ut8* addr_out) {
 	char const *separator = r_str_lchr (addr_str, '.');
 	ut8 byte;
 	int bit;
-	bool ret = false;
+	bool ret = true;
 	// TODO: check if symbols are resolved properly in all cases:
 	// - symbol.2
 	// - 0x25.symbol
 	// - symbol.symbol
 	// - symbol
 	if (!separator) {
+		ret = false;
 		goto end;
 	}
 	r_str_ncpy (bytepart, addr_str, separator - addr_str + 1);
 	bytepart[separator - addr_str + 1] = '\0';
 	r_str_ncpy (bitpart, separator + 1, strlen (separator));
-	if (!address_direct (bytepart, &byte)) {
-		goto end;
-	}
-	if (1 < strlen (bitpart)
-		|| bitpart[0] < '0' || '7' < bitpart[0]) {
+
+	bit = bitpart[0] - '0';
+
+	if (bit > 9) {
 		ret = false;
 		goto end;
 	}
-	bit = bitpart[0] - '0';
-	if (0x20 <= byte && byte < 0x30) {
-		*addr_out = (byte - 0x20) * 8 + bit;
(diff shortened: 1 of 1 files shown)


gogo2464:
Just warn the user when an opcode can not be asembled yet.

libr/asm/arch/8051/8051_ass.c
```diff
@@ -413,22 +413,181 @@ static bool address_bit(char const* addr_str, ut8* addr_out) {
...
 	r_str_ncpy (bitpart, separator + 1, strlen (separator));
-	if (!address_direct (bytepart, &byte)) {
-		goto end;
-	}
-	if (1 < strlen (bitpart)
-		|| bitpart[0] < '0' || '7' < bitpart[0]) {
-		ret = false;
-		goto end;
-	}
-	bit = bitpart[0] - '0';
```
Lazula, line 424: `bit` should be kept, as it is used several times below

libr/asm/arch/8051/8051_ass.c
```diff
@@ -413,22 +413,181 @@ static bool address_bit(char const* addr_str, ut8* addr_out) {
...
-	if (0x20 <= byte && byte < 0x30) {
-		*addr_out = (byte - 0x20) * 8 + bit;
-		ret = true;
-	} else if (0x80 <= byte && !(byte%8)) {
-		*addr_out = byte + bit;
-		ret = true;
+
+	if (!strcmp (bytepart, "b")) {
+		if ((bitpart[0] - '0') <= 9) {
```
Lazula, line 418: the statement `if (bit <= 9)` appears in every block here. check it once before the elif chain instead:

```
if (bit <= 0) {
        ret = false;
        goto end;
}
```

and you can remove the `else {ret = false}` in every case.

Lazula: you can also set ret to true before the elif chain and set it to false in the final else to avoid having `ret = true` in every elif block.

libr/asm/arch/8051/8051_ass.c
```diff
@@ -413,22 +413,181 @@ static bool address_bit(char const* addr_str, ut8* addr_out) {
...
+			*addr_out = 0xf0 + (bitpart[0] - '0');
```
Lazula, line 419: as the structure is the same in each place here, it may be easier to read if you put the constant part (`0xf0`) into a variable, then have one place where you set `*addr_out` after the loop

gogo2464: @Lazula  could you provide an example please?

Lazula: something like this:
```
ut8 addr_bytepart;
if (!strcmp (bytepart, "b")) {
        addr_bytepart = 0xf0;
}
```

and at the end:

```
*addr_out = addr_bytepart + bit;
```

gogo2464: fine I understand

Lazula:
looks good to me, thanks! :)

gogo2464:
Cool !

gogo2464:
My bad. Let me two minutes.

PR #19500, opened 2021-12-14
Check for an existing upstream remote in install scripts ##build

Lazula:
**Checklist**

**Description**

Follow-up to #19423. If there is an existing remote the user has tracking upstream, using the direct URL leaves the remote lagging behind and requires manual updating. Install scripts now check for an existing named remote before pulling from the direct URL, keeping the remote tracking aligned with the current `master` when updated by these scripts.

### DIFF
--- a/sys/termux.sh
+++ b/sys/termux.sh
@@ -7,7 +7,13 @@ export ANDROID=1
 # make clean > /dev/null 2>&1
 rm -f libr/include/r_version.h
 cp -f dist/plugins-cfg/plugins.termux.cfg plugins.cfg
-git pull https://github.com/radareorg/radare2 master
+# Attempt to update from an existing remote
+UPSTREAM_REMOTE=$(git remote -v | grep 'radareorg/radare2 (fetch)' | cut -f1 | head -n1)
+if [ -n "$UPSTREAM_REMOTE" ]; then
+	git pull "$UPSTREAM_REMOTE" master
+else
+	git pull https://github.com/radareorg/radare2 master
+fi
 ./preconfigure
 ./configure-plugins
 bash ./configure --with-compiler=termux --prefix=${PREFIX} || exit 1
--- a/sys/user.sh
+++ b/sys/user.sh
@@ -53,7 +53,13 @@ if [ $WITHOUT_PULL -eq 0 ]; then
 		git branch | grep "^\* master" > /dev/null
 		if [ $? = 0 ]; then
 			echo "WARNING: Updating from remote repository"
-			git pull https://github.com/radareorg/radare2 master
+			# Attempt to update from an existing remote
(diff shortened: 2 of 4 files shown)


sys/user.sh
```diff
@@ -53,7 +53,13 @@ if [ $WITHOUT_PULL -eq 0 ]; then
 		git branch | grep "^\* master" > /dev/null
 		if [ $? = 0 ]; then
 			echo "WARNING: Updating from remote repository"
-			git pull https://github.com/radareorg/radare2 master
+			# Attempt to update from an existing remote
+			UPSTREAM_REMOTE=$(git remote -v | grep 'radareorg/radare2 (fetch)' | cut -f1 | head -n1)
+			if [ -n "$UPSTREAM_REMOTE" ]; then
+				git pull "$UPSTREAM_REMOTE" master
+			else
+				git pull https://github.com/radareorg/radare2 master
```
trufae, line 61: what about taking this url from `ORIGIN` in the root of the project. also, why's a problem to just keep the `git pull` we had before? i consider r2 forks to be able to build by themselves without depending on the origin repo, this breaks a little that logic.

sys/termux.sh
```diff
@@ -7,7 +7,13 @@ export ANDROID=1
 # make clean > /dev/null 2>&1
 rm -f libr/include/r_version.h
 cp -f dist/plugins-cfg/plugins.termux.cfg plugins.cfg
-git pull https://github.com/radareorg/radare2 master
+# Attempt to update from an existing remote
+UPSTREAM_REMOTE=$(git remote -v | grep 'radareorg/radare2 (fetch)' | cut -f1 | head -n1)
+if [ -n "$UPSTREAM_REMOTE" ]; then
+  git pull "$UPSTREAM_REMOTE" master
```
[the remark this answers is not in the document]
Lazula: oops, thought i caught all of these. i have `:set expandtab` on .sh files by default

sys/user.sh
Lazula: to my understanding the conditions for `git pull` are intended to update `master` to match upstream during install scripts (based on the conditions). running `git pull` by itself has 2 problems:

1. `master` may not be tracking any upstream remote. this is why we use the remote URL, because otherwise it does nothing.
2. if an upstream remote is being tracked, using the named remote here makes sure it doesnt fall behind. e.g.: i have my upstream remote named `radareorg` and without this, the branch tracking lags behind `master` and i have to manually `git pull radareorg master` to make sure it stays aligned

the original use of the `git pull` is preserved, it just eases some common use cases. i think we shouldnt assume a raw git clone from upstream. plenty of people have origin as their own fork and this lets them fetch from upstream easily.

Lazula: i may have mixed up my phrasing here but i mean that master isnt always tracking upstream

PR #19511, opened 2021-12-17
Move /e to new search API ##search

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This moves `/e` command over to `r_search_update_read` and fixes a few bugs in `/e`. 

This fixes some bugs in `/e`.
* Less false negatives due to matches crossing buffer boundaries. This can still happen if a match is larger then 64 characters and it falls across a buffer boundary. This should be updated in the future so users can configure an expected largest match length.
* Prevent infinite loop when searching for a zero length match `/e /A*/`.
* Matches have a proper length.

This pull implements  a second callback system that should replace the first. When a match was found previously, the callback assumed the match length was equal to the keyword length. So a match of `/A{30}/` would be treated as a length of 5. The new callback accepts a size of match to fix this. The old callback system should probably be fazed out.

### DIFF
--- a/libr/core/cmd_search.c
+++ b/libr/core/cmd_search.c
@@ -568,6 +568,7 @@ static char *getstring(char *b, int l) {
 }
 
 static int _cb_hit_sz(RSearchKeyword *kw, int klen, void *user, ut64 addr) {
+	r_return_val_if_fail (kw && user, -1);
 	struct search_parameters *param = user;
 	RCore *core = param->core;
 	ut64 base_addr = 0;
@@ -682,11 +683,16 @@ static int _cb_hit_sz(RSearchKeyword *kw, int klen, void *user, ut64 addr) {
 	return true;
 }
 
-static int _cb_hit(RSearchKeyword *kw, void *user, ut64 addr) {
-	struct search_parameters *param = user;
-	const RSearch *search = param->core->search;
-	int klen = kw? kw->keyword_length + (search->mode == R_SEARCH_DELTAKEY): 0;
-	return _cb_hit_sz (kw, klen, user, addr);
+static int _cb_hit(R_NULLABLE RSearchKeyword *kw, void *user, ut64 addr) {
+	RSearchKeyword kw_fake = { 0 };
+	RSearchKeyword *kw_used = &kw_fake;
+	int klen = 0;
+	if (kw) {
+		struct search_parameters *param = user;
+		const RSearch *search = param->core->search;
+		klen = kw? kw->keyword_length + (search->mode == R_SEARCH_DELTAKEY): 0;
+	}
+	return _cb_hit_sz (kw_used, klen, user, addr);
 }
 
 static inline void print_search_progress(ut64 at, ut64 to, int n, struct search_parameters *param) {
--- a/libr/core/cmd_search.c
+++ b/libr/core/cmd_search.c
@@ -231,6 +231,7 @@ struct search_parameters {
 	bool inverse;
 	bool aes_search;
 	bool privkey_search;
+	int c; // used for progress
 };
 
 struct endlist_pair {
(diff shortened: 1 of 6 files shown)


libr/core/cmd_search.c
```diff
@@ -683,6 +681,13 @@ static int _cb_hit(RSearchKeyword *kw, void *user, ut64 addr) {
 	return true;
 }
 
+static int _cb_hit(RSearchKeyword *kw, void *user, ut64 addr) {
+	struct search_parameters *param = user;
+	const RSearch *search = param->core->search;
+	int klen = kw? kw->keyword_length + (search->mode == R_SEARCH_DELTAKEY): 0;
```
trufae, line 687: so, it is valid to run this callback with kw as null? will be better to use R_NULLABLE in the signature definition. and i understand the reason for this case is when using search backends that don't use the concept of "keywords" like patterns and such, but maybe for those cases we can just holdl a fake instance in the stackframe and pass it, so you can keep using the kw->len instead of passing more arguments

libr/core/cmd_search.c
```diff
@@ -683,6 +681,13 @@ static int _cb_hit(RSearchKeyword *kw, void *user, ut64 addr) {
...
+	return _cb_hit_sz (kw, klen, user, addr);
+}
+
 static int c = 0;
```
trufae, line 691: globals must die. move them all on top of the file and put a scary warning comment over it

libr/search/regexp.c
```diff
@@ -1,8 +1,87 @@
...
+			reflags |= R_REGEX_ICASE;
+		}
+
+		if (r_regex_init (&rx, (char *)kw->bin_keyword, reflags)) {
+			eprintf ("Cannot compile '%s' regexp\n", kw->bin_keyword);
+			return -1;
+		}
+
+		while (addr < to) { // get buffer
+			ut8 buf[512];
```
trufae, line 29: handle r_cons_is_breaked() in this inner loop and store this buffer in the heap at the begining of the function, probably better to read blocks of 4KB or honor core->blocksize

libr/search/regexp.c
```diff
@@ -1,8 +1,87 @@
...
+		ut64 addr = from;
+		int reflags = R_REGEX_EXTENDED;
+
+		if (kw->icase) {
+			reflags |= R_REGEX_ICASE;
+		}
+
+		if (r_regex_init (&rx, (char *)kw->bin_keyword, reflags)) {
+			eprintf ("Cannot compile '%s' regexp\n", kw->bin_keyword);
+			return -1;
```
trufae, line 25: uhm, that rx will leak as its freed outside the outer loop block

libr/search/search.c
```diff
@@ -476,10 +483,23 @@ R_API void r_search_pattern_size(RSearch *s, int size) {
...
+		s->r_callback = NULL;
+	}
 	s->callback = callback;
 	s->user = user;
 }
 
+R_API void r_search_set_read_cb(RSearch *s, RSearchRCb cb, void *user) {
+	if (s->callback) {
+		// prevent user from being passed to wrong function
+		s->callback = NULL;
```
trufae, line 497: no need to check for null to set a null

libr/search/search.c
```diff
@@ -476,10 +483,23 @@ R_API void r_search_pattern_size(RSearch *s, int size) {
 }
 
 R_API void r_search_set_callback(RSearch *s, RSearchCallback(callback), void *user) {
+	if (s->r_callback) {
```
trufae, line 486: no need to chk for null when set to null

trufae (changes requested):

libr/core/cmd_search.c
swoops: Check what I did and let me know.

libr/core/cmd_search.c
swoops: Moved it out of global space and put into a struct

swoops:
I can't figure out how to make windows happy...

trufae:
Maybe its better to use RConsBind or something, i dont think the search api should force the developers to instantiate rcons

PR #19513, opened 2021-12-19
Fix bug in regex searching ##search

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Searching for `/e /A*/` will cause an infinite loop. This fixes that and only allows regex matches that have a positive length.


trufae:
Can you add some tests to avoid those issues to come back?

trufae:
Merging for now as its an important fix. thanks!

PR #19538, opened 2021-12-27
Add Rabin Karp algorythm to ##search

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This adds the `/k` command to search for a single string using the Rabin Karp search algorithm. The alg is just a hash of the `n` byte search string, hash of the first `n` bytes of search space and a compare. If the compare matches a full string compare is done. The only cool part is the hash is a "rolling hash", so you don't have to re-compute the entire hash for the next search space hash, you just update the hash to add in the next byte and remove the previous byte.

Compared to the `/` command, this algorithm runs faster if there would of been backtracking. For example, on a file containing 0x200000 `A`'s, the searching for `ZAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAZ` will not cause any backtracking so `/` and `/k` run at nearly the same speed:

```
[0x00000000]> ?t /k ZAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAZ
Searching in [0x0-0x200001]
0.062433
[0x00000000]> ?t / ZAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAZ
Searching 42 bytes in [0x0-0x200001]
hits: 0
0.058487
```

But when you search for `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAZ` there will be a lot of backtracking done by the `/` algorithm. So `/k` runs an order of magnitude faster:

```
[0x00000000]> ?t /k AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAZ
Searching in [0x0-0x200001]
0.062421
[0x00000000]> ?t / AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAZ
Searching 41 bytes in [0x0-0x200001]
hits: 0
0.725914
```

In practice, I doubt this alone will be all that helpful. I want to get this basic idea in mostly so it can be used and checked for bugs. Eventually I will update this to work with multiple search inputs using a hash map and as a result reduce those search times in practice.

**About the hashing algorithm**
The hash algorithm is left a bit abstract so it can more easily be modified. This might not be necessary. It is a Galois field, which is just fancy talk for "you can think of it as math". The idea is for a search string `s` of length `n` you do `s[0] * pow(x,n) + s[1] * pow(x, n-1)...` but `pow()`, `+` are expensive so you use bit rotation and xor instead. See `hash_full` and `roll_forward`. I think a `roll_backward` should be possible as well, for doing backward searches.

In particular I tried to scramble things so that every bit of the each hashed byte is distributed evenly throughout the hash. I am hoping this will result in less collisions when used in a hashmap. I *think* it has the property that the 32 bit version of the hash can be obtained by xor'ing the most significant 4 bytes with the least significant 4 bytes. Same for the 16 bit version.  So the hash can be reduced as desired for the hashmap and the larger hash can be checked in the hashmap before doing the full string compare.

I really doubt this hash is novel. I just made it up from some stuff I read on Wikipedia. If someone finds a hash that produces less collisions, they are welcome to replace this mine.


swoops:
Found a bug just now... it misses a result if it's at the end of the search space. I will send a fix soon.

trufae:
Sorry for the delay. i was a little against the /k command because that was used in r1 to configure multiple keywords but i understand the change as something transitional, so its good for testing for now. Amazing benchmarks! good job

PR #19552, opened 2022-01-07
Fix fast jump with ahc on register calls ##visual

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This fixes what I think is a bug, but maybe I am not thinking about it properly?

For a `call sym.somewhere` you can use `ahc sym.other` to set the suggested location where the call actually lands. Then in visual modes you can get promps that will fast jump to the `sym.other` location.

When you have a `call rax` and you know `rax` is always holds `sym.other` you can *not* do `ahc sym.other` because `ds_print_core_vmode` will de-reference `sym.other`. So for `ahc` to work in this case, you must find a location that points to `sym.other`. In static analysis, such a location may not exist.

I feel like `ahc` means "use this jump location". I see the merit of de-referencing in some situations but I feel like a different command should be used.

So this pull "fixes" things but maybe something else should be done. Let me know if I can do something better and I will.

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -3814,7 +3814,6 @@ static bool ds_print_core_vmode(RDisasmState *ds, int pos) {
 	case R_ANAL_OP_TYPE_RJMP:
 	case R_ANAL_OP_TYPE_RCALL:
 		if (ds->analop.jump != UT64_MAX && ds->analop.jump != UT32_MAX) {
-			ds->analop.jump = get_ptr_ble (ds, ds->analop.jump);
 			slen = ds_print_shortcut (ds, ds->analop.jump, pos);
 			gotShortcut = true;
 		}


libr/core/disasm.c
```diff
@@ -3814,7 +3814,6 @@ static bool ds_print_core_vmode(RDisasmState *ds, int pos) {
 	case R_ANAL_OP_TYPE_RJMP:
 	case R_ANAL_OP_TYPE_RCALL:
 		if (ds->analop.jump != UT64_MAX && ds->analop.jump != UT32_MAX) {
-			ds->analop.jump = get_ptr_ble (ds, ds->analop.jump);
```
trufae, line 3817: uhm, RCALL/RJMP, is about register reference, the `get_ptr_ble` should be only for the MJMP, MCALL, .. so i think your patch is good

trufae: i think the check in 				`if (ds->asm_hint_call_indirect) {` is covering that for the UCALL/UJMP, maybe we should handle MJMP or just remove it and make it UJMP, but U is about unknown.. and there are probably more places where MJMP is misused, can you check that too?

trufae: will be good to have some tests for x86-64 and arm64 on this

trufae: diff between `call rax`, `call [rax]`, `call [0x80080]` ,..

trufae:
Let's merge for now, we can add the tests and improvethe plugins later :)

thanks for spotting and fixing this!

PR #19563, opened 2022-01-10
Fix new coverity leaks

Lazula:


Lazula:
sp->event wasn't from coverity but it was unavoidable in code flow so it cluttered up ASan output

trufae (approved):

PR #19567, opened 2022-01-12
Add 'first session' example in the README

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,21 @@ r2env init
 r2env add radare2@git
 ```
 
+## Usage
+
+These are the first steps to use r2, read the book or find tutorials for more details
+
+```sh
+$ r2 /bin/ls   # open the binary in read-only mode
+> px 32        # print 32 byte hexdump current block
+> f~foo        # filter flags with ~grep (same as |grep)
+> aaa          # same as r2 -A, analyse the binary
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,6 +37,21 @@ r2env init
...
 ```
 
+## Usage
+
+These are the first steps to use r2, read the book or find tutorials for more details
+
+```sh
+$ r2 /bin/ls   # open the binary in read-only mode
+> px 32        # print 32 byte hexdump current block
+> f~foo        # filter flags with ~grep (same as |grep)
```
alex-bender, line 47: Having that it's for  the first steps, it's better to introduce jargon slowly, so better show example on function list, for example
```
$ r2 /bin/ls
> aaa           # analyze functions and cross references, same as `r2 -A`
> afl~sym    # filter functions with `~` ...

> oo+;w hello    # write `hello` in current offset
```
would be better to introduce `seek` in that list

trufae: chk again, feel free to suggest more changes

alex-bender: looks good. Actually I'm thinking about some tutorial being in code. Kind of vimtutor

trufae:
Merging then! :) feel free to submit another PR improving it if you notice anything

PR #19581, opened 2022-01-13
Add r_core_help_match() to get help for a specific command

Lazula:
This new function makes it easier to get help text for specific commands in color. By avoiding grep and working directly on the help text array, we can pinpoint the exact line we want by using strcmp and borrowing the string pointers to craft a custom help text array. The only allocation needed is done to append spec to cmd if `spec != 0`.

```
/* Impossible to get color due to color codes interfering */
r_core_cmd0 (core, "pd?~^\| pdl");

/* These 2 are interchangable - spec makes it easier to check for ? in one place */
r_core_cmd_help_match (core, help_msg_pd, "pdl");
r_core_cmd_help_match_spec (core, help_msg_pd, "pd", 'l');

/* Get help for an appropriate subcommand */
r_core_cmd_help_match_spec (core, help_msg_pd, "pd", input[1]);
```

### DIFF
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -2146,29 +2146,9 @@ R_API void r_cons_cmd_help(const char *help[], bool use_color) {
 	}
 }
 
-/* See r_cons_cmd_help().
- * This version will only print help for a specific command.
- * Will append spec to cmd before looking for a match, if spec != 0.
- */
-R_API void r_cons_cmd_help_match(const char *help[], bool use_color, R_BORROW R_NONNULL char *cmd, char spec) {
-	const char **match = NULL;
+static void print_match(const char **match, bool use_color) {
 	const char *match_help_text[4];
-	int i;
-
-	if (spec) {
-		/* We now own cmd */
-		cmd = r_str_newf ("%s%c", cmd, spec);
-	}
-
-	for (i = 0; help[i]; i += 3) {
-		if (!strcmp (help[i], cmd)) {
-			match = &help[i];
-			break;
-		}
-	}
-
-	/* Command was not found in help text. */
-	r_return_if_fail (match);
+	size_t i;
 
 	/* Manually construct help array. No need to strdup, just borrow. */
 	match_help_text[3] = NULL;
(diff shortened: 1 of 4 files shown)


Lazula:
Fixed not initializing `match`

libr/cons/cons.c
```diff
@@ -2146,6 +2146,42 @@ R_API void r_cons_cmd_help(const char *help[], bool use_color) {
...
 }
 
+/* See r_cons_cmd_help().
+ * This version will only print help for a specific command.
+ * Will append spec to cmd before looking for a match, if spec != 0.
+ */
+R_API void r_cons_cmd_help_match(const char *help[], bool use_color, R_BORROW R_NONNULL char *cmd, char spec) {
+	const char **match = NULL;
+	const char *match_help_text[4];
+	int i;
```
trufae, line 2156:
```suggestion
	size_t i;
```

libr/cons/cons.c
```diff
@@ -2146,6 +2146,42 @@ R_API void r_cons_cmd_help(const char *help[], bool use_color) {
...
+
+	if (spec) {
+		/* We now own cmd */
+		cmd = r_str_newf ("%s%c", cmd, spec);
+	}
+
+	for (i = 0; help[i]; i += 3) {
+		if (!strcmp (help[i], cmd)) {
+			match = &help[i];
```
trufae, line 2165: only one match? whaht about strstr instead of strcmp? or add an option flag to specify that? i think it can be useful to filter the subcommands not just the specific command

libr/cons/cons.c
```diff
@@ -2146,6 +2146,42 @@ R_API void r_cons_cmd_help(const char *help[], bool use_color) {
...
+	match_help_text[3] = NULL;
+	for (i = 0; i < 3; i++) {
+		match_help_text[i] = match[i];
+	}
+	r_cons_cmd_help (match_help_text, use_color);
+
+	if (spec) {
+		free (cmd);
+	}
+}
```
trufae, line 2183: maybe return a pointer to the help string if any or null?

trufae (changes requested):
Looks good despite not being used anywhere, how do you plan to handle that? from the syntax point-of-view we can handle the chars after the '?' as filter text in color , instead of depending on   ~.

So for example that will be `pd?pdr` instead of `pd~pdr`

libr/cons/cons.c
Lazula: do you mean `return match`? im fine with changing that, its only void bc r_core_cmd_help() returns void too.

right now this function considers no match to be a programming error but now that i think of it this causes problems if you dont check spec first, so that should be changed. i can change the return to char ** and the r_return_val to `return NULL`;

Lazula: not sure what the use case for a return value is

libr/cons/cons.c
Lazula: true, would be good to get both `pdr` and `pdr.` when you give it ("pd", 'r'). can add a `bool exact`.

could work if we return void, but gets more complicated if we want to return a pointer. maybe return a list?

Lazula:
> Looks good despite not being used anywhere, how do you plan to handle that?

Use it in new code and slowly bring it in, especially when looking for commands where ? fails. In practice this should be a recursive approach, e.g. `p` -> `pd` -> `pd{b,c,e,...}` -> (go up) `px` -> `px{0,a,b,c,...}`. It's going to take a while, but it will clean up the approach to subcommand help which is currently handled in various different ways.

> So for example that will be `pd?pdr` instead of `pd~pdr`

This could be a good way to expose it to the user if we want to. I mostly had internal usage in mind. So if you run `pdr?` you get just the line for `pdr`. with the suggested `exact` option disabled you would get `pdr` and `pdr.`. Not sure what the use case for `pd?pdr` is, but I don't see why we shouldn't give the option.

libr/cons/cons.c
Lazula: though then the caller would need to handle the new memory, im not sure about it. i dont think we can do both of these things

trufae: Ok

libr/cons/cons.c
trufae: Use the api somewhere so I can see an example usage

trufae:
I understand correctly the usage will be when grepping with the last char in the cmd ==? Right?

trufae:
i think it can be useful to have and there will be places to use it. waiting for your next PRs. i think is better to merge it for now, so you can base your changes on top of master

PR #19588, opened 2022-01-15
V850 ESIL

brainstorm:
**Checklist**

- [ ] Closing issues: #issue

**Description**

Not sure if NOT is right, WIP... why isn't NOT implemented in ESIL bit arithmetic, btw?: https://github.com/radareorg/radare2/blob/master/doc/esil.md#bit-arithmetic


doc/esil.md
```diff
@@ -24,6 +24,52 @@ peeks and pokes, spawning a syscall, etc.
 
     [0x00000000]> e asm.esil = true
 
+Running ESIL
+============
+
+In visual mode, `V`, one can iterate through the instructions via the `s` (step) key
```
trufae, line 30: imho is better to document `aes` or `ae` which evaluates and steps from the shell without going into visual because visual cant be scripted that easily. also, if you want to document this, i think is better to make a PR to the book instead. this file aims to be just a very basic reference guide to the esil syntax

doc/esil.md
```diff
@@ -24,6 +24,52 @@ peeks and pokes, spawning a syscall, etc.
...
+and see how registers are changing interactively as `;-- pc` (program counter) advances,
+just like in r2's debug facilities:
+
+[0x00100004 [xaDvc]0 2% 395 bin/ired_v850]> diq;?t0;f .. @ entry0+4 # 0x100004
+dead at 0x00000000
+- offset -   0 1  2 3  4 5  6 7  8 9  A B  C D  E F  0123456789ABCDEF
+0x00200000  ffff ffff ffff ffff ffff ffff ffff ffff  ................
```
trufae, line 37: run aeim to initialize a fake stack or you'll get this ffff. but anyway, using aes and dr= should be enough for testing purposes. the visual mode is just the same as in the debugger

libr/anal/arch/v850np/opc.inc
```diff
@@ -709,8 +709,8 @@ const struct v850_opcode v850_opcodes[] = {
 	{ "add", OP (0x12), OP_MASK, IF2, 0, V850_CPU_ALL, R_ANAL_OP_TYPE_ADD, "$0,$1,+=" },
 	{ "addi", OP (0x30), OP_MASK, IF6, 0, V850_CPU_ALL, R_ANAL_OP_TYPE_ADD, "$0,$1,+,$2,=" },
 	{ "adf", two (0x07e0, 0x03a0), two (0x07e0, 0x07e1), {CCCC_NOTSA, R1, R2, R3}, 0, V850_CPU_E2_UP },
-	{ "and", OP (0x0a), OP_MASK, IF1, 0, V850_CPU_ALL, R_ANAL_OP_TYPE_AND },
-	{ "andi", OP (0x36), OP_MASK, IF6U, 0, V850_CPU_ALL, R_ANAL_OP_TYPE_AND },
+	{ "and", OP (0x0a), OP_MASK, IF1, 0, V850_CPU_ALL, R_ANAL_OP_TYPE_AND, "$0,$1,&,$1,=" }, // TODO: OV, S, Z
```
trufae, line 712: there's also &=, but i think @condret wants to kill it, and could become a macro, so whatever

PR #19598, opened 2022-01-18
Fix potential overflow in r_file_mkstemp

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

While checking out a bug in `r_file_mkstemp` I found a potential overflow. The overflow can not be reached by a user, so there is no security impact. 

The only path that comes close to reaching this overflow is through `r_core_editor` but for `r_core_editor` to call `r_file_mkstemp` the file name must start with `*`. This means the `prefix` that will be copied into the fixed size buffer with have a length of 0. Since 0 bytes fits easily in a 1024 byte buffer there is no overflow :).

There is still another bug in `r_file_mkstemp` that I plan to fix soon but it is a bit more involved. So I figured it's best to get this in first, in case some dev wants to use `r_core_editor` with a file name that has a `*` character more then 1024 bytes deep.

### DIFF
--- a/libr/util/file.c
+++ b/libr/util/file.c
@@ -1183,28 +1183,31 @@ err_r_file_mkstemp:
 	free (path_);
 	free (prefix_);
 #else
-	char pfxx[1024];
-	const char *suffix = strchr (prefix, '*');
-
-	if (suffix) {
-		suffix++;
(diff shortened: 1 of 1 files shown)


swoops:
If #19601 looks good, then this is unnecessary.

PR #19601, opened 2022-01-18
Update r_file_mkstemp to work with sandbox

swoops:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This simplifies `r_file_mkstemp` and fixes a couple bugs. If this is accepted then #19598 can just be closed.

Bugs:
On my system the macro `defined(__GLIBC__) && defined(__GLIBC_MINOR__) && 2 <= __GLIBC__ && 19 <= __GLIBC__MINOR__` does not seem to correctly detect the presence of `mkstemps`. 

The fallback to `mkstemp` causes bugs when a suffix is used. Given `r2.XXXXXX.h`, the alg truncates the name to `r2.XXXXXX` to be acceptable to `mkstemp`. This then returns a file descriptor for `r2.XXXXXX`. Afterward the alg ads the suffix back in, returning a filename of `r2.XXXXXX.h`. So running `toe struct.foo` will populate `r2.XXXXXX` file with the known `foo` struct and open the new file `r2.XXXXXX.h` in the text editor.

Also, previously the sandbox was not used for file creation here. This should fix that.

### DIFF
--- a/libr/util/file.c
+++ b/libr/util/file.c
@@ -1164,7 +1164,7 @@ R_API char *r_file_temp_ex(R_NULLABLE const char *prefix, R_NULLABLE const char 
 
 static inline char *file_fmt_split(const char *fmt) {
 	if (R_STR_ISEMPTY (fmt)) {
-		return r_file_temp_suf (NULL, NULL);
+		return r_file_temp_ex (NULL, NULL);
 	}
 	char *name = NULL;
 	char *dup = strdup (fmt);
--- a/libr/util/file.c
+++ b/libr/util/file.c
@@ -1149,7 +1149,7 @@ R_API char *r_file_temp(const char *prefix) {
 	return res;
 }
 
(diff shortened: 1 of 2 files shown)


libr/util/file.c
```diff
@@ -1149,15 +1149,51 @@ R_API char *r_file_temp(const char *prefix) {
 	return res;
 }
 
+R_API char *r_file_temp_suf(R_NULLABLE const char *prefix, R_NULLABLE const char *suf) {
```
trufae, line 1152: Suf stands for suffix but in the filename jargon that would be extension. I think there can be a better name for this function

trufae (changes requested):

libr/util/file.c
swoops: I am pretty bad at naming things. How about `r_file_temp_ex`?

Issue #18739, opened 2021-05-24
Implement pdu


trufae:
pdu - should be print disasm until (condition: address, type of instruction..)

trufae:
@Lazula wanna check this issue?

Lazula:
Will take this on now.

c-urly:
This we need to implement?

`[0x00006160]> pdu 0x0000616e   ------> until address  ` <br>` 
;-- entry0:
;-- rip:   ` <br>` 
        0x00006160      31ed           xor ebp, ebp   ` <br>` 
        0x00006162      4989d1         mov r9, rdx   ` <br>` 
        0x00006165      5e             pop rsi  ` <br>` 
        0x00006166      4889e2         mov rdx, rsp  ` <br>` 
        0x00006169      4883e4f0       and rsp, 0xfffffffffffffff0  ` <br>` 
        0x0000616d      50             push rax  ` <br>` 
         0x0000616e      54             push rsp  ` <br>` `

`[0x00006160]> pdu op push   ------> until first push INS  ` <br>` 
;-- entry0:   ` <br>` 
            ;-- rip:   ` <br>` 
            0x00006160      31ed           xor ebp, ebp   ` <br>` 
            0x00006162      4989d1         mov r9, rdx  ` <br>` 
            0x00006165      5e             pop rsi  ` <br>` 
            0x00006166      4889e2         mov rdx, rsp  ` <br>` 
            0x00006169      4883e4f0       and rsp, 0xfffffffffffffff0  ` <br>` 
            0x0000616d      50             push rax  ` <br>`

trufae:
if i understand your question properly, you are asking to include the instruction. imho it's easier, and probably more consistent if `pdu X` is the same as `pD X-$$`. so it shouldnt be including that instruction, because it doesnt know the size of it before hand.

If i consider `pdu` kind of `dcu` (as in continue until given address), i expect the debugger to stop in the next instruction. What other people think about this? @Lazula ?

Lazula:
i can open a draft PR to give an idea of my current direction. so far my thoughts are this (copy-pasting the help text):

```
static const char *help_msg_pdu[] = {
        "Usage:", "pdu[aceios]", "Disassemble instructions until condition",
        "pdua", " [addr]", "disassemble until address",
        "pduc", "", "disassemble until call",
        "pdue", " [expr]", "disassemble until esil expression",
[...]

PR #19616, opened 2022-01-22
Implement pdu

Lazula:
**Checklist**

- [X] Closing issues: #18739

**Description**

Initial implementation of `pdu` commands. All of the proposed commands have been implemented except for `pdue` and they all have tests.

`r_core_print_disasm()` has been modified to add condition type and pointer arguments. Internally, each condition type can be cast from void* and placed in their own special variable. To add a new condition type, add a new entry to the condition type enumeration in r_core.h and handle it in both cmd_pdu() and r_core_print_disasm()

### DIFF
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -1669,7 +1669,7 @@ static int core_anal_graph_construct_nodes(RCore *core, RAnalFunction *fcn, int 
 			if (buf) {
 				r_io_read_at (core->io, bbi->addr, buf, bbi->size);
 				if (is_json_format_disasm) {
-					r_core_print_disasm (core, bbi->addr, buf, bbi->size, bbi->size, true, true, pj, NULL);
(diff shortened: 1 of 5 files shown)


Lazula:
If it's preferred I could just modify `r_core_print_disasm()` with the 2 condition arguments and change the loop and retry conditions.

trufae:
Ideally we shouldnt be having more copies of the same disasm function, so maybe better to reuse a single implementation and pass a struct with custom options instead of adding more arguments

Lazula:
For now I've just added 2 arguments to `r_core_print_disasm()`, adding an option structure is out of scope here. Will need help with `pdue` but I think everything else is covered now.

Lazula:
Will merge it once tests finish - the ones that failed were due to an unused variable and I've commented out pdue stuff to be added later

PR #19618, opened 2022-01-23
Rearrange 'dij' output curly brace and commas

Talos878:
**Checklist**

- [x] Closing issues: #issue

**Description**

update dij so that the output doesn't have inconsistent use of curly braces

### DIFF
--- a/libr/core/cmd_debug.c
+++ b/libr/core/cmd_debug.c
@@ -5287,8 +5287,9 @@ static int cmd_debug(void *data, const char *input) {
 				break;
 			case 'j': // "dij"
 				P ("{");
-				P ("\"stopreason\":\"%s\"}\n", r_debug_reason_to_string (stop));
+				P ("\"stopreason\":\"%s\"", r_debug_reason_to_string (stop));
 				if (rdi) {
+					P (",");
 					const char *s = r_signal_to_string (core->dbg->reason.signum);
 					P ("\"type\":\"%s\",", r_debug_reason_to_string (core->dbg->reason.type));
 					P ("\"signal\":\"%s\",", r_str_get_fail (s, "none"));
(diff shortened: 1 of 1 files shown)


trufae:
Please, switch to use the PJ api, its safer and cleaner way to serialize json. Thanks

Talos878:
What is the PJ api?

trufae:
Check libr/util/pj.c and grep for uses in the source with `git grep r_core_pj_new`

Talos878:
Okay, I will make a separate pull request in the future in esilsolve when switching over, cause I think that uses the old api. Do you know is the person who made esilsolve still active with it?

trufae:
The author of esilsolver wrote radius which is a replacement for it. So i assume its not as maintained as radius

Talos878:
I saw radius but haven't had the time to truly learn rust, so I avoided it since I am just doing this for learning purposes. For now I am going to mark this closed, and will open another request later if I do anything with radius or esilsolve

aemmitt-ns:
esilsolve does not use dij, and only uses di to check whether a debugger is being used. same with radius. but they might use dij in the future so thanks for the fix

Talos878:
Hells yeah, that would be awesome. How much longer have you been working on radius since esilsolve? Also what made you choose to write your own solver, rather than try to append functionality to something like angr to make it be able to handle esil instructions? Thanks for the tool

aemmitt-ns:
ok i realized that i switched to just checking `di` in uncommitted changes, so they are pushed now and r200.py should work again. been working on radius for like 9 months? and i mostly made it just cuz i liked ESIL and wanted to try using it for symbolic execution. its really nice to be able to write ESIL expressions to do quick things, like in the ESILSolve r2 plugin or the `radius2` cmd line tool. `radius2` is really easy to use you can install it with `cargo install radius2` and do cool things like solve r200 with just `radius2 -p r200 -s stdin 48 str -X Incorrect -m 0x4007ba` and it takes ~0.1 seconds not ~2.4 like for ES. And you can write hooks in ESIL so like if you add `-H main SP,1,+,.` itll print out rsp+1 when it starts main. radius adds two ESIL operators `.` to print an arg and `_` which constrains the arg to be  != 0. So nearly everything you might want to do in a hook can be done with a short ESIL string. Its great for rapidly trying things in a flexible way that other frameworks like angr and manticore are too slow and annoying for.

trufae:
@Talos878 do you have the other PR using PJ ready?

Talos878:
Sorry, no. Give me a week or two. I have some things I have to sort through
before I can get to that.

PR #19619, opened 2022-01-23
Support for Capstone EVM disassembly

sylvainpelissier:
**Checklist**

- [ ] Closing issues: #issue

**Description**
Initial plugin for Capstone disassembly support.


trufae:
I commited only the anal.evm.cs plugin on purpose. RAsm api now uses RAnal plugins as disassembler backends. so there's no need to duplicate the logic here and there. If you want to help ont hat process, the plan is to move all the asm plugins and make the analop callback handle the DISASM mask

sylvainpelissier:
Sorry, I will help on that.

PR #19630, opened 2022-01-25
Dont SIGINT in RCons when used in a thready way ##cons

trufae:
**Checklist**

- [ ] Closing issues: #issue

**Description**

### DIFF
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -2240,8 +2240,8 @@ R_API void r_cons_thready(void) {
 	}
 	r_th_lock_enter (r_cons_lock);
 	r_cons_instance = R_NEW0 (RCons);
-	r_th_lock_leave (r_cons_lock);
 	C->unbreakable = true;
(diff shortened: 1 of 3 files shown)


libr/cons/cons.c
```diff
@@ -2235,6 +2241,7 @@ R_API void r_cons_thready(void) {
 	r_th_lock_enter (r_cons_lock);
 	r_cons_instance = R_NEW0 (RCons);
 	r_th_lock_leave (r_cons_lock);
+	C->unbreakable = true;
```
meme, line 2244: Shouldn't we be holding the `r_cons_lock` while setting `unbreakable = true`?

meme (changes requested):

libr/cons/cons.c
trufae: totally. and this is why i shouldnt merge things when having migraine :D 

thnks for the catch!

PR #19635, opened 2022-01-27
fix esil for bnd jmp x86 instrs and cmn arm instrs ##esil

aemmitt-ns:
**Checklist**

- [ ] Closing issues: #issue

**Description**

This PR fixes the an issue where `bnd jmp` instructions had ESIL corresponding to `repne jmp` because Intel MPX changed the `repne` prefix to mean `bnd` on control flow instructions. it was a security thing that no longer does anything so we too will ignore it 

It also fixes the `cmn` instruction ESIL which is supposed to compare the negative of the register value to the second operand, but instead was being treated as a normal cmp.

### DIFF
--- a/libr/anal/p/anal_arm_cs.c
+++ b/libr/anal/p/anal_arm_cs.c
@@ -1895,10 +1895,8 @@ static int analop64_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int l
 		break;
 	}
 	case ARM64_INS_CCMP:
-	case ARM64_INS_CCMN:
 	case ARM64_INS_TST: // cmp w8, 0xd
 	case ARM64_INS_CMP: // cmp w8, 0xd
-	case ARM64_INS_CMN: // cmp w8, 0xd
 		ARG64_APPEND(&op->esil, 1);
 		COMMA(&op->esil);
 		ARG64_APPEND(&op->esil, 0);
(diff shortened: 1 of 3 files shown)


aemmitt-ns:
not sure whats going on in the failed tests, some weird capstone thing ?

libr/anal/p/anal_arm_cs.c
```diff
@@ -1895,10 +1895,8 @@ static int analop64_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int l
 		break;
 	}
 	case ARM64_INS_CCMP:
-	case ARM64_INS_CCMN:
 	case ARM64_INS_TST: // cmp w8, 0xd
 	case ARM64_INS_CMP: // cmp w8, 0xd
-	case ARM64_INS_CMN: // cmp w8, 0xd
 		ARG64_APPEND(&op->esil, 1);
```
trufae, line 1900: Add a space before (

trufae (changes requested):
Fix the space before the ( in calls and i'll merge. The current failing tests are because of capstone-next being wrong

libr/anal/p/anal_arm_cs.c
trufae: i found some more cases like this on already existing code. so i merged and im fixing it in local .thanks for the contrib!

PR #19636, opened 2022-01-27
WIP: Add safe static lock initialization ##util

meme:
**Checklist**

- [ ] Closing issues: #issue

**Description**

* Allows RThreadLock to be initialized on acquisition, which means global locks do not need to be heap allocated

Note: MSVC does not support C11 atomics so those users should make sure they are wearing their seatbelts. I think we could use Windows APIs instead for interlocked exchange/store.


meme:
Closed in favour of https://github.com/radareorg/radare2/pull/19645

PR #19662, opened 2022-02-01
Fix Negative Offset in Hexdump Json Output

AandersonL:
**Checklist**

- [ ] Closing issues: #issue
- [ x] Mark this if you consider it ready to merge

**Description**

When the user try to get a json output of a negative relative offset in the `px` command, it will always get a fixed `256` data block.

This fix will rely in the `core->num->value` instead the `core->blocksize` when passing it to `r_print_jsondump` making the output correct.

How to reproduce:

`xj -10`

### DIFF
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -6541,7 +6541,7 @@ static int cmd_print(void *data, const char *input) {
 		r_cons_break_push (NULL, NULL);
 		switch (input[1]) {
 		case 'j': // "pxj"
-			r_print_jsondump (core->print, core->block, core->num->value, 8);
+			r_print_jsondump (core->print, core->block, len, 8);
 			break;
 		case '/': // "px/"
 			r_core_print_examine (core, input + 2);
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -6541,7 +6541,7 @@ static int cmd_print(void *data, const char *input) {
 		r_cons_break_push (NULL, NULL);
 		switch (input[1]) {
 		case 'j': // "pxj"
-			r_print_jsondump (core->print, core->block, core->blocksize, 8);
+			r_print_jsondump (core->print, core->block, core->num->value, 8);
 			break;
 		case '/': // "px/"
 			r_core_print_examine (core, input + 2);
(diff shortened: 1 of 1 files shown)


libr/core/cmd_print.c
```diff
@@ -6541,7 +6541,7 @@ static int cmd_print(void *data, const char *input) {
 		r_cons_break_push (NULL, NULL);
 		switch (input[1]) {
 		case 'j': // "pxj"
-			r_print_jsondump (core->print, core->block, core->blocksize, 8);
+			r_print_jsondump (core->print, core->block, core->num->value, 8);
```
trufae, line 6544: Can you test this patch? See other cases how it takes the data.
```suggestion
			r_print_jsondump (core->print, core->block, len, 8);
```

also can you add a test to ensure the output is correct for positive and negative offsets, as well as 0

trufae (changes requested):

libr/core/cmd_print.c
trufae: In the tests try also bigger output than the blocksize. Something like:

```
pxj 10
pxj -10
b 16
pxj
pxj 32
pxj-32
```

AandersonL:
```
b 16
pxj
pxj 32
pxj-32
```
The following tests produced a wrong output from the negative value, it cames 16 bytes instead 32

trufae: Yep, i can confirm this issue too. But i think it's fixed if you pass buf and len.

Please create a new test by adding those lines in `test/db/cmd/cmd_print` and run it like this `r2r -i test/db/cmd/cmd_print`.

libr/core/cmd_print.c
```diff
@@ -6541,7 +6541,7 @@ static int cmd_print(void *data, const char *input) {
 		r_cons_break_push (NULL, NULL);
 		switch (input[1]) {
 		case 'j': // "pxj"
-			r_print_jsondump (core->print, core->block, core->blocksize, 8);
+			r_print_jsondump (core->print, core->block, len, 8);
```
trufae, line 6544:
```suggestion
			r_print_jsondump (core->print, buf, len, 8);
```

libr/core/cmd_print.c
trufae: ping?

PR #19692, opened 2022-02-12
Fix order of operations in the ESIL code of pop with memory operand

nemarci:
**Checklist**

- [ ] Closing issues: #issue

**Description**

The ESIL code of the x86 instruction `pop [rsp]` was wrong, because the calculated memory destination was the value of `rsp` **before** the `pop` instruction; this is wrong behaviour according to the Intel manual:

If the ESP register is used as a base register for addressing a destination operand in memory, the POP instruction computes the effective address of the operand after it increments the ESP register.


trufae:
Can you add a test? that's such an important thing to take into account for not repeating that bug in the future. I think this issue has been fixed a couple of times at least :_

thanks

PR #19696, opened 2022-02-13
Even moar v850 esil

brainstorm:
IIRC, I was looking at the signextend macros for ESIL and they were not doing the right thing, need to revisit the remaining few insns that we need:

```
mulh
sar
set1
sld.hu
sld.w
sst.b
st.b
st.h
st.w
sxh
tst1
zxb
zxh
```


trufae:
superseeded by https://github.com/radareorg/radare2/pull/19698

PR #19699, opened 2022-02-13
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
**Checklist**

- [ ] Closing issues: #issue
- [X] Mark this if you consider it ready to merge
- [ ] I've added tests (optional)
- [ ] I wrote some lines in the [radare2book](https://github.com/radareorg/radare2book)

**Description**

Allow to override build date with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.
This date call only works with GNU date. *Is MacOS-support required?*

Also use UTC to be independent of timezone.

This is the equivalent of 6b260b87c3345568ebeddf57fbe95c864ee8baf2 for meson.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

PR #19705, opened 2022-02-13
Rework `SOURCE_DATE_EPOCH` usage be portable

bmwiedemann:
**Checklist**

- [ ] Closing issues: #issue

**Description**

As a follow-up of #19699,
Rework `SOURCE_DATE_EPOCH` usage in meson to be portable
to make @rofl0r and FreeBSD users happy.

Please test.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -72,11 +72,12 @@ endif
 if host_machine.system() == 'windows'
   r2birth = run_command('cmd', '/c', 'echo %date%__%time%')
 else
-  source_date_epoch = run_command('sh', '-c', 'echo $SOURCE_DATE_EPOCH').stdout().strip()
-  if source_date_epoch == ''
-     source_date_epoch = run_command('date', '+%s').stdout().strip()
-  endif
-  r2birth = run_command('date', '-u', '-d', '@' + source_date_epoch, '+%Y-%m-%d__%H:%M:%S')
+  r2birth = run_command('sh', '-c', '''
+      SOURCE_DATE_EPOCH="${SOURCE_DATE_EPOCH:-`date +%s`}";
+      FORMAT="+%Y-%m-%d__%H:%M:%S";
+      date -u -d @"$SOURCE_DATE_EPOCH" "$FORMAT" 2>/dev/null ||
+      date -u -r  "$SOURCE_DATE_EPOCH" "$FORMAT" 2>/dev/null ||
+      date -u "$FORMAT"''')
 endif
 if r2birth.returncode() != 0
   r2birth = ''


meson.build
```diff
@@ -72,7 +72,7 @@ endif
 if host_machine.system() == 'windows'
   r2birth = run_command('cmd', '/c', 'echo %date%__%time%')
 else
-  r2birth = run_command('date', '+%Y-%m-%d__%H:%M:%S')
+  r2birth = run_command('sh', '-c', 'SOURCE_DATE_EPOCH="${SOURCE_DATE_EPOCH:-`date +%s`}"; FORMAT="+%Y-%m-%d__%H:%M:%S"; date -u -d @"$SOURCE_DATE_EPOCH" "$FORMAT" 2>/dev/null || date -u -r "$SOURCE_DATE_EPOCH" "$FORMAT" 2>/dev/null || date -u "$FORMAT"')
```
trufae, line 75: that line requires bash and its pretty unreadable, maybe better to move the logic into sys/date.sh and make the code cleaner here.

trufae (changes requested):

meson.build
bmwiedemann: I thought, it should work with any POSIX-compliant shell. Or did I miss a bash-ism? A quick with Debian's `dash` looked good. A `sys/date.sh` would probably need a shell, too.
For readability, I thought about adding linebreaks, but not sure what escaping is needed in meson.

rofl0r: having it in a separate script would also make it easier to use it in the alternative configure/make build sys, i guess

bmwiedemann: I changed it into a multi-line string for now. I'm off to bed now. Feel free to push the move to `date.sh` to this PR.

trufae: Oh, i was mistaken, the ${:-} is posix and works on mksh,dash,zsh .. so definively not a bashism. its way more readable this way :) thanks for the improvement

eli-schwartz:
Alternatively, the standard `__DATE__` and `__TIME__` macros in C could be used GCC since 2017 will respect the SOURCE_DATE_EPOCH environment variable *for* you and you don't have to do anything in meson.build (in fact you can just remove complexity from both that and configure/make).

trufae:
What's your proposal then?

eli-schwartz:
https://github.com/radareorg/radare2/blob/ad749a0b1a2f3f8fcc7826639c572f52631c5a61/libr/util/str.c#L3917
https://github.com/radareorg/radare2/blob/88e8fdfbd39c092cebccae630f70ff855d143f70/libr/main/radare2.c#L675

These two locations can be modified to use `__DATE__`. They will switch from `YYYY-MM-DD` to `MMM DD YYYY`, and when SOURCE_DATE_EPOCH is set the compiler (at least GCC) will use that epoch. When it isn't set, or you use some compiler that doesn't itself support the reproducible builds standard (I assume MSVC probably doesn't?) it will just use the time that the compiler was invoked on that file.

trufae:
I'm fine to merge this PR and i understand your proposal, but im not sure if it will work as its intended to have the same timestamp across all binaries. That would be another behaviour, that maybe we can discuss and use if agreed. Would you like to make a PR to continue the discussion there?

PR #19712, opened 2022-02-14
Fix #19687 - Correct types definitions for ptrace(2) on FreeBSD ##io

### LINKED ISSUE #19687: EPIC: Improve the CI
Right now the CI takes so much time to test all the things, it's fine somehow as long as we can cancel some jobs sometimes.

General
* [x] Automatically publish artifacts in releases
* [x] Reduce time, decide better which pipelines to run and merge some

FreeBSD
* [x] run some tests (like tcc or gperf pipelines do, not the whole testsuite?)
* [x] Dont copy back the source directory via rsync (rm -rf *)
* [ ] Create dist/freebsd package https://github.com/danrue/oneoff-pkg-create
* [x] Fix warnings in io.ptrace
* [ ] Cache somehow the vbox image? (upvote the action repo!)

Windows
* [ ] Fix race condition in meson build system
* [ ] Build may fail when not using configure;make

Fedora
* [ ] Update dist/rpm from Fedora repository
  * https://github.com/radareorg/radare2/issues/19681
* [ ] Create .rpm by adding a fedora pipeline and publish the artifact

Android
* [ ] Add builds for android kitkat (32bit)

AH-DL:

**Description**

### libr/include/r_io.h
- Argument  **r_ptrace_data_t** needs to be **int** instead of **void**  on FreeBSD ptrace(2)
- Then **R_PTRACE_NODATA** should be initialized as **0** instead of **NULL**

### DIFF
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -58,7 +58,7 @@ typedef void * r_ptrace_data_t;
 #elif __APPLE__
 typedef int r_ptrace_request_t;
 typedef int r_ptrace_data_t;
-#elif __OpenBSD__
+#elif __OpenBSD__ || __FreeBSD__
 typedef int r_ptrace_request_t;
 typedef int r_ptrace_data_t;
 #define R_PTRACE_NODATA 0


trufae:
Thank you! i just fixed the other warning (ALIGN macro redefined)

PR #19720, opened 2022-02-15
CMDBREAK: Use wx+ instead of wx;s+16 in pc* command ##print

trufae:
This change introduces a breaking change because `wx+` was introduced in r2 5.6.2, so the output of `pc*` wont be usable on older instances of r2. as a workaround we can add `pc**` to keep the old behaviour, and just try to stay modern in the default command.

Opinions? Vote with emojis! The idea is to only introduce ABI, API and CMD breaking changes in "minor version" releases. So i'll hold this change until then and we have time to decide

Screenshot 2022-02-16 at 11 57 52


condret:
My opinion is, that if you downgrade r2, you're either a dev or you're doing something wrong. Just do it

trufae:
Having abi and cmd stability over a period of time reduces the amount of complains from users updating without the need to compile the plugins again and also help tooling to be safely maintained knowing that abis and commands wont break. We can have some notes on which commands and apis has changed on every minor release. Thats when we can break things. The rest of patch releases are judt for fixes.

actually with the current state of the things is pretty easy to improve the tool without having to add or change structs or functions. So its not blocking to add new features or fix things 99% of the time

PR #19728, opened 2022-02-18
Fix heap buffer overflow in dyldcache parser ##crash

radare:
* Reported by: Lazymio via huntr.dev
* Reproducer: dyldovf

```mermaid
graph LR;
   Fuzz --> Crash;
   Crash -->Verify;
   Crash --> Fuzz;
   Verify --> Report;
   Report --> Fix;
   Fix --> Verify;
   Fix --> Publish
```


wtdcode:
Verified the fix works.

PR #19735, opened 2022-02-18
Add support for powerpc, powerpc64, powerpc64le and riscv64 on FreeBSD ##arch

pkubaj:
Add support for powerpc, powerpc64, powerpc64le and riscv64 on FreeBSD ##arch

Also correct issue in vmenus.c, where ch is checked later whether it's -1,
but on ARM and POWER, char is unsigned.

**Description**


libr/core/vmenus.c
```diff
@@ -3301,7 +3301,7 @@ R_API void r_core_visual_debugtraces(RCore *core, const char *input) {
 		//int rows = r_cons_get_size (NULL);
 		r_core_cmdf (core, "dtd %d", delta);
 		r_cons_visual_flush ();
-		char ch ;
+		signed char ch ;
```
trufae, line 3304:
```suggestion
		signed char ch;
```

PR #19736, opened 2022-02-18
Fix DoS on macho parser spotted by scan coverity ##crash

trufae:

**Description**

### DIFF
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -3151,6 +3151,13 @@ static void parse_relocation_info(struct MACH0_(obj_t) *bin, RSkipList *relocs, 
 	}
 
 	ut64 total_size = num * sizeof (struct relocation_info);
+	if (offset > bin->size) {
+		return;
+	}
+	if (total_size > bin->size) {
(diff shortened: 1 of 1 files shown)


libr/bin/format/mach0/mach0.c
```diff
@@ -3151,6 +3151,13 @@ static void parse_relocation_info(struct MACH0_(obj_t) *bin, RSkipList *relocs,
 	}
 
 	ut64 total_size = num * sizeof (struct relocation_info);
+	if (offset > bin->size) {
+		return;
+	}
+	if (total_size > bin->size) {
```
gogo2464, line 3157: @trufae This definitely fixes the crash for security but I think there might be a way to make the codebase lighter.

trufae: What do you mean by making the codebase lighter?

gogo2464: I mean remove the if and let a single num allocation.

trufae: The limit is the boundary of the file size you cant use a constant here

PR #19745, opened 2022-02-21
Adding NULL Check to __func_name_from_ord to prevent segfault 

archcloudlabs:
Function [__func_name_from_ord](https://github.com/radareorg/radare2/blob/master/libr/bin/format/ne/ne.c#L51) does not perform a NULL check on module which can result in a segfault with a specially crafted binary.

A specially crafted binary will result in crashing radare2 as shown in the image below.

PUBLIC_radare2-crash


radare:
Thanks!

PR #19770, opened 2022-03-01
Add R2_DEBUG_EPRINT to disable debug var printing at compiletime or runtime

Lazula:
EPRINT macros are very useful for debugging, but interfere with output for tests when running r2r. This patch adds a check for the `R2_SUPPRESS_VAR_EPRINT` environment variable to allow the user (or the test suite) to disable them at runtime (must be `"1"`).

This includes a minor refactor to the macros - now, instead of each one being a raw eprintf, they use a wrapper that passes the name, format string, and `__VA_ARGS__` along to eprintf after checking the environment variable. The format string is partially pre-constructed to make the definitions more clear about what the specific type needs to be printed properly.

Example usage:

```
$ git diff libr/core/cmd_print.c
diff --git a/libr/core/cmd_print.c b/libr/core/cmd_print.c
index 0d77c1700..43ed406db 100644
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -6575,6 +6575,7 @@ static int cmd_print(void *data, const char *input) {
                 cmd_print_op(core, input);
                 break;
         case 'x': // "px"
+                EPRINT_STR (input);
                 if (input[1] == '-' && input[2] == '-') {
                         int rowsize = r_config_get_i (core->config, "hex.cols");
                         int ctxlines = r_num_math (core->num, input + 3);
$ sys/user.sh
$ r2 -
[0x00000000]> px >/dev/null
input: "x"
[0x00000000]> q
$ R2_SUPPRESS_VAR_EPRINT=1 r2 -
[0x00000000]> px >/dev/null
[0x00000000]> q
```

### DIFF
--- a/binr/r2r/r2r.c
+++ b/binr/r2r/r2r.c
@@ -336,6 +336,7 @@ int main(int argc, char **argv) {
 
 	r_sys_setenv ("RABIN2_TRYLIB", "0");
 	r_sys_setenv ("R2_DEBUG_ASSERT", "1");
+	r_sys_setenv ("R2_SUPPRESS_VAR_EPRINT", "1");
 	r_sys_setenv ("TZ", "UTC");
 	ut64 time_start = r_time_now_mono ();
 	R2RState state = {{0}};
--- a/binr/r2r/r2r.c
+++ b/binr/r2r/r2r.c
@@ -336,7 +336,7 @@ int main(int argc, char **argv) {
 
 	r_sys_setenv ("RABIN2_TRYLIB", "0");
 	r_sys_setenv ("R2_DEBUG_ASSERT", "1");
-	r_sys_setenv ("R2_NO_EPRINT_MACROS", "1");
+	r_sys_setenv ("R2_DEBUG_EPRINT", "0");
 	r_sys_setenv ("TZ", "UTC");
 	ut64 time_start = r_time_now_mono ();
 	R2RState state = {{0}};
--- a/binr/r2r/r2r.c
+++ b/binr/r2r/r2r.c
@@ -336,7 +336,7 @@ int main(int argc, char **argv) {
 
 	r_sys_setenv ("RABIN2_TRYLIB", "0");
 	r_sys_setenv ("R2_DEBUG_ASSERT", "1");
-	r_sys_setenv ("R2_SUPPRESS_VAR_EPRINT", "1");
+	r_sys_setenv ("R2_NO_EPRINT_MACROS", "1");
 	r_sys_setenv ("TZ", "UTC");
 	ut64 time_start = r_time_now_mono ();
 	R2RState state = {{0}};
--- a/libr/include/r_types.h
+++ b/libr/include/r_types.h
@@ -463,7 +463,10 @@ static inline void *r_new_copy(int size, void *data) {
 #define eprintf(...) fprintf (stderr, __VA_ARGS__)
 #endif
 
-#ifdef NO_EPRINT_MACROS
+#ifndef R2_DEBUG_EPRINT
+#define R2_DEBUG_EPRINT 0
+#endif
+#if !R2_DEBUG_EPRINT
 #define EPRINT_STR
 #define EPRINT_CHAR
 #define EPRINT_INT
--- a/libr/include/r_types.h
+++ b/libr/include/r_types.h
@@ -461,23 +461,34 @@ static inline void *r_new_copy(int size, void *data) {
 
 #ifndef eprintf
 #define eprintf(...) fprintf (stderr, __VA_ARGS__)
-
-#define EPRINT_STR(x) eprintf (#x ": \"%s\"\n", x)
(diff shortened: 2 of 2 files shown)


libr/include/r_types.h
```diff
@@ -462,6 +462,17 @@ static inline void *r_new_copy(int size, void *data) {
...
+/* These macros may be turned off at will. This is so that they can be left in
+ * the codebase while running tests. Disabling USE_DEBUG_VAR_PRINTS will
+ * prevent them from affecting program output so tests can be run as normal
+ * after recompiling.
+ *
+ * Note to maintainers: This macro should be left enabled in production, but
+ * the function macros should be unused. The functions are supposed to be
+ * immediately available to anyone hacking on the source.
+ */
+#define USE_DEBUG_VAR_PRINTS 1
```
trufae, line 474: uhmf the user cant override this value with an env var or configure option.. so it seems like that should be wrapped inside an #ifdef

Lazula:
freebsd ssh dying again...

trufae:
Oh noes i maybe didnt explained well what i was refering was a way to redefine that value from commandline or adding a cflags. Sysgetenv is overkill here

Lazula:
hmm, idk exactly how to do that. runtime checking is very useful though as it lets you build with the debug prints but turn it off if you want, plus r2r can always disable which imo is nice and convenient instead of needing to rebuild with it disabled to run tests

trufae:
Is what you had before but adding an #ifndef/#endif ... before defining the variable that shows/hides the debug prints

Lazula:
Has both compile-time and run-time options to disable now

binr/r2r/r2r.c
```diff
@@ -336,6 +336,7 @@ int main(int argc, char **argv) {
 
 	r_sys_setenv ("RABIN2_TRYLIB", "0");
 	r_sys_setenv ("R2_DEBUG_ASSERT", "1");
+	r_sys_setenv ("R2_SUPPRESS_VAR_EPRINT", "1");
```
trufae, line 339:
```suggestion
	r_sys_setenv ("R2_DEBUG_EPRINT", "1");
```

libr/include/r_types.h
```diff
@@ -461,24 +461,35 @@ static inline void *r_new_copy(int size, void *data) {
...
 #ifndef eprintf
 #define eprintf(...) fprintf (stderr, __VA_ARGS__)
+#endif
 
-#define EPRINT_STR(x) eprintf (#x ": \"%s\"\n", x)
-#define EPRINT_CHAR(x) eprintf (#x ": '%c' (0x%x)\n", x, x)
-#define EPRINT_INT(x) eprintf (#x ": %d (0x%x)\n", x, x)
-#define EPRINT_BOOL(x) eprintf (#x ": %s\n", x? "true": "false")
-#define EPRINT_PTR(x) eprintf (#x ": %p\n", x)
+#ifndef NO_EPRINT_MACROS
```
trufae, line 466: Else?

Lazula:
I think `NO_EPRINT_MACROS` (compile flag) and `R2_NO_EPRINT_MACROS` (env) are good names for this.

trufae:
i would keep the same name for define and env var.

libr/include/r_types.h
```diff
@@ -461,24 +461,50 @@ static inline void *r_new_copy(int size, void *data) {
...
 #ifndef eprintf
 #define eprintf(...) fprintf (stderr, __VA_ARGS__)
+#endif
 
-#define EPRINT_STR(x) eprintf (#x ": \"%s\"\n", x)
-#define EPRINT_CHAR(x) eprintf (#x ": '%c' (0x%x)\n", x, x)
-#define EPRINT_INT(x) eprintf (#x ": %d (0x%x)\n", x, x)
-#define EPRINT_BOOL(x) eprintf (#x ": %s\n", x? "true": "false")
-#define EPRINT_PTR(x) eprintf (#x ": %p\n", x)
+#ifdef NO_EPRINT_MACROS
```
trufae, line 466:
```suggestion
#ifndef R2_DEBUG_EPRINT
#define R2_DEBUG_EPRINT 0
#endif
#if !R2_DEBUG_EPRINT
```

binr/r2r/r2r.c
```diff
@@ -336,6 +336,7 @@ int main(int argc, char **argv) {
 
 	r_sys_setenv ("RABIN2_TRYLIB", "0");
 	r_sys_setenv ("R2_DEBUG_ASSERT", "1");
+	r_sys_setenv ("R2_NO_EPRINT_MACROS", "1");
```
trufae, line 339:
```suggestion
	r_sys_setenv ("R2_DEBUG_EPRINT", "0");
```

trufae (changes requested):

trufae:
lgtm after applying my suggestion changes

PR #19776, opened 2022-03-01
Partial #19687: Fix CI release notes on creation

prodrigestivill:
This pull request responses to #19687.

Add git history before generate release notes.


trufae:
That script is failing for two reasons, and this is just one of the reasons :D ill take care of the head==tag problem .

thanks!

PR #19785, opened 2022-03-03
Refactor `cp` command

Lazula:
Clean up cp command implementation and emit output in `r_file_copy()` when src and dst are the same instead of failing silently.


libr/core/cmd_cmp.c
```diff
@@ -720,7 +754,7 @@ static int cmd_cmp(void *data, const char *input) {
 
 	switch (*input) {
 	case 'p':
-		return cmd_cp (data, input);
+		return cmd_cp (data, input+1);
```
trufae, line 757:
```suggestion
		return cmd_cp (data, input + 1);
```

trufae (changes requested):

PR #19789, opened 2022-03-03
Clean up files in tests

Lazula:
Update the testsuite to not leave behind files that show up in `git status`. Wont happen in practice until the ytf fix is merged, but isnt blocked by it. The order doesnt matter once both are in.

### DIFF
--- a/test/db/cmd/cmd_fs
+++ b/test/db/cmd/cmd_fs
@@ -883,6 +883,7 @@ CMDS=<<EOF
 mg /root/bin/ls
 o ./ls
 i~size
+rm ls
 EOF
 EXPECT=<<EOF
 size     0x1ee78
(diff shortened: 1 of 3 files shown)


trufae:
This can be racy. Better get a different file or work on a different directory. Both tests can run at the same time

trufae:
Yeah the failing test seems weird. Try rebasing :?

Lazula:
> This can be racy. Better get a different file or work on a different directory. Both tests can run at the same time

maybe tests should use a mktmp for this instead of predetermined paths? seems hard to maintain all of them while needing to pay attention to that, this test uses .tmp1, this one uses .tmp2, etc. etc.

trufae:
Looks good now :D good fix!

PR #19795, opened 2022-03-06
Fix `dd` command

Lazula:
`dd` is supposed to print a `dx` command with hex bytes for you to paste, but it's failing due to the filename being unquoted and treated as an immediate instead of a string.

Additionally, the current implementation of `r_core_syscall()` has no static frame size for storing strings. It has been changed to account for a single string argument and adjust the static frame appropriately.


libr/core/cmd_debug.c
```diff
@@ -5118,12 +5118,12 @@ static int cmd_debug(void *data, const char *input) {
 		case ' ': // "dd"
 			// TODO: handle read, readwrite, append
 			{
-				RBuffer *buf = r_core_syscallf (core, "open", "%s, %d, %d", input + 2, 2, 0644);
+				RBuffer *buf = r_core_syscallf (core, "open", "\"%s\", %d, %d", input + 2, 2, 0644);
```
trufae, line 5121: Maybe good to escape the string to avoid quotes and other problematic chars before passing it.

also if the user wants to pass a numeric value with the address of the allocated string that wont work. So maybe we can also check if the passed string is a number and use pfmt64x

what do you think?

trufae (changes requested):
Good catch :)

libr/core/cmd_debug.c
```diff
@@ -4755,22 +4754,16 @@ static ut8 *getFileData(RCore *core, const char *arg, int *sz) {
...
 	int i;
+	r_return_if_fail (buf);
+
+	r_cons_print ("dx ");
 	r_buf_seek (buf, 0, R_BUF_SET);
 	for (i = 0; i < r_buf_size (buf); i++) {
 		r_cons_printf ("%02x", r_buf_read8 (buf));
 	}
-	r_cons_printf ("\n");
+	r_cons_print ("\n");
```
trufae, line 4766: r_
```suggestion
	r_cons_newline ();
```
```suggestion
	r_cons_print ("\n");
```

libr/core/cmd_debug.c
```diff
@@ -5064,32 +5062,43 @@ static int cmd_debug(void *data, const char *input) {
...
-				if (str) off = r_num_math (core->num, str+1);
-				if (str) str = strchr (str+1, ' ');
-				if (str) len = r_num_math (core->num, str+1);
-				if (len == UT64_MAX || off == UT64_MAX || \
-						!r_debug_desc_read (core->dbg, fd, off, len)) {
-					consumeBuffer (r_core_syscallf (core, "read", "%d, 0x%"PFMT64x", %d",
-								fd, off, (int)len), "dx ", "Cannot read");
+				if (str) {
+					off = r_num_math (core->num, str+1);
+					str = strchr (str+1, ' ');
```
trufae, line 5083:
```suggestion
					str = strchr (str + 1, ' ');
```

Lazula:
Further fix here: #19853

PR #19797, opened 2022-03-07
Add missing eiz/riz registers for x86 and x64 ##analysis

trufae:
* ESIL was already correct, but the regprofile was missing those

**Description**


trufae:
broken tests raise another pretty good improvement to do. i'll handle that in this same pr

Issue #19805, opened 2022-03-09
Error creating zignatures for function(s) in Mach-O binary


whoami-defnotme:
## Environment

```sh
Wed Mar  9 15:38:11 +08 2022
radare2 5.6.5 28988 @ darwin-x86-64 git.5.5.0
commit: 6a4dd942f853e77ddf7fab1ddafbc4f44a1cf3ba build: 2022-03-09__13:55:36
Darwin x86_64
```

## Description

I'm observing failures when attempting to create zignatures for a Mach-O binary. It's unclear if there's an issue with the function or if this is an r2 bug.

## Test

```
# First, file hashes
[0x100003790]> it
md5 95d76db3a8cd1bb4e2a1bdfeb54f35fb
sha1 e7204d84b60ef5a03d51680c2f0c568684a4addd
sha256 1e19e76af67ffc7249c99ef66c28f4425f79e4f97d4976bb093fda2d0fee1a2b

# Try to create the zig for the target function
[0x100003790]> zaf "method.OnePasswordNativeMessageHost.OPNMXPCConnection.setXpcConnection:" test

# Then, try to view all zigs as this was the only one created.
[0x100003790]> z
Invalid types: ```void method.OnePasswordNativeMessageHost.OPNMXPCConnection.setXpcConnection: (int64_t arg1, void *instance)``` in signatuer for zign|*|test
error: cannot deserialize zign
```

swoops:
@whoami-defnotme if you are trying to get some work done, you can just remove `:` from the function name before creating the signature and it should work. I will see about getting it to work otherwise.

whoami-defnotme:
Oh sweet as - yea, suspected it might have been a sanitization issue. I can replace semi-colons with underscores and keep going.

swoops:
#19807 should fix it completely, so I am going to close. Let us know if anything else goes bad @whoami-defnotme

PR #19807, opened 2022-03-09
Sanitize function names for prototypes ##anal

swoops:

**Description**

Potential/partial fix for #19805

This will ensure function prototypes created by `r_anal_function_get_signature` have a function name that tcc will accept. When a function has an invalid name, the invalid chars will be replaced with `_`. So in the case of #19805 the function name `method.OnePasswordNativeMessageHost.OPNMXPCConnection.setXpcConnection:` will become `method.OnePasswordNativeMessageHost.OPNMXPCConnection.setXpcConnection_` in the prototype.

For signatures this won't affect the renaming of the function on match. The matched function should still become `method.OnePasswordNativeMessageHost.OPNMXPCConnection.setXpcConnection:`. I don't know if it will have another ill affect somewhere else.

The core of the issue might be that a functions can have names with `:` in them. So feel free to deny this pull and tell me to fix it another way :)

## example with this patch

```
> r2 /bin/ls
Warning: run r2 with -e bin.cache=true to fix relocations in disassembly
 ╭──╮    ╭─────────────────────────────────────╮
 │ ╶│╶   │                                     │
 │ O o  <  r2 is meant to be read by machines. │
 │  │  ╱ │                                     │
 │ ╭┘ ╱  ╰─────────────────────────────────────╯
 │ ╰ ╱
 ╰──'
[0x00006180]> aa
[x] Analyze all functions arguments/locals
[0x00006180]> s main
[0x00004760]> afn method.OnePasswordNativeMessageHost.OPNMXPCConnection.setXpcConnection:
[0x00004760]> zaf
[0x00004760]> z.
[+] searching function metrics
bytes:  1
graph:  1
refs:   1
types:  1
bbhash: 1
total:  5
[0x00004760]> z~type
  types: int method.OnePasswordNativeMessageHost.OPNMXPCConnection.setXpcConnection_ (int argc, char **argv)
[0x00004760]> 
```

### DIFF
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -1847,6 +1847,29 @@ R_API char *r_anal_function_get_json(RAnalFunction *function) {
 	return pj_drain (pj);
 }
 
+#define ALPH(x) \
+	(x >= 'a' && x <= 'z') || (x >= 'A' && x <= 'Z')
+#define ISNUM(x) \
+	(x >= '0' && x <= '9')
+#define VALID_TOKEN_CHR(x) \
+	(ALPH (x) || ISNUM (x) || x == '_' || x == '.')
+static inline char *sanitize_fname(const char *name) {
+	char *sane = strdup (name);
+	if (sane) {
+		char *ptr = sane;
+		while (*ptr) {
+			if (!VALID_TOKEN_CHR (*ptr)) {
+				*ptr = '_';
+			}
+			ptr++;
+		}
+	}
+	return sane;
+}
+#undef ALPH
+#undef ISNUM
+#undef VALID_TOKEN_CHR
+
 R_API char *r_anal_function_get_signature(RAnalFunction *function) {
 	RAnal *a = function->anal;
 	const char *realname = NULL, *import_substring = NULL;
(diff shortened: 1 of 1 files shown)


trufae:
Arent the functions in util/filter/name enough? The idea is to reduce the amount of functions that filter text

trufae:
Check `138 R_API bool r_name_filter(char *s, int maxlen) {` or `121 R_API bool r_name_filter_print(char *s) {`

swoops:
> Check `138 R_API bool r_name_filter(char *s, int maxlen) {` or `121 R_API bool r_name_filter_print(char *s) {`

Ah yeah, I will fix this in a bit.

swoops:
Turns out both `r_name_filter` and `r_name_filter_print` accept `:` via `r_name_validate_char` ([here](https://github.com/radareorg/radare2/blob/9dea9a21e91347e4c30e2e7053efac9043118422/libr/include/r_util/r_name.h#L22)). Removing `:` from `r_name_validate_char` breaks a lot of tests.

So I used `r_name_filter2`, which does `r_name_filter` + `strdup` and then just replaced `:` in place. It's a extra pass but speed is not very important here and functions names shouldn't be huge (c++ withstanding :p). If you would rather I create a second stricter `r_name_filter` or something, I can do that.

trufae:
Looks shorter now :) thanks!

PR #19808, opened 2022-03-09
Check for .git extension in upstream remote in install scripts

Lazula:
Improvement to #19500. Now, if you used ".git" in the remote name it will still work.


trufae:
nvm thats a basic regex, no need for -e

PR #19814, opened 2022-03-12
Remove R_EMPTY and R_EMPTY2

Lazula:
Big automated search and replace.

```
$ find test binr libr -name "*.c" -exec sed -i -e 's/{{ \?0 \?}}/R_EMPTY2/' {} \; && find test binr libr -name "*.c" -exec sed -i -e 's/{ \?0 \?}/R_EMPTY/' {} \;
# exclude files without access to r_types
$ git restore libr/asm/arch/alpha/gnu/alpha-opc.c libr/asm/arch/arm/aarch64/aarch64-opc-2.c libr/asm/arch/arm/aarch64/aarch64-opc.c libr/asm/arch/mcore/mcore.c libr/asm/arch/mips/gnu/micromips-opc.c libr/asm/arch/mips/gnu/mips-opc.c libr/asm/arch/mips/gnu/mips16-opc.c libr/asm/arch/ppc/gnu/ppc-opc.c libr/asm/arch/ppc/libvle/vle.c libr/asm/arch/tricore/gnu/tricore-dis.c libr/asm/arch/v850/gnu/v850-opc.c libr/asm/p/asm_i4004.c libr/bin/mangling/cxx/cp-demangle.c
```


trufae:
I have never been a fan of R_EMPTY, for two reasons: it's longer to type than {0} or {{0}} and at the end is just an static initialization so, if we have R_NEW, it would fit better to use R_INIT. What do you think? I agree that we should define a `standard` and use it everywhere for consistency. so i'm all 👍 with the change

Lazula:
```
$ find shlr libr binr test -name "*.c" -exec sed -i -e 's/R_EMPTY2/{{0}}/g' {} 2>/dev/null \;
$ find shlr libr binr test -name "*.c" -exec sed -i -e 's/R_EMPTY/{0}/g' {} 2>/dev/null \;
```

PR #19842, opened 2022-03-19
Refactor, fix, and test `dd` command

Lazula:
(hijack of a simple dd fix that turned into a full refactor)

Major refactor of dd including:

* `dd` commands execute by default, with the `*` postfix printing commands.
* `dd+` command now has python-style automatic creation
* `ddr` and `ddw` now take the proper arguments
* r_core_syscall now uses 0 for read instead of 128. r_syscall_get_num is still bugged, it's a hack workaround
* Tests for all dd commands!
* Implementation for linux_desc_list offset. Now actually reported instead of always showing 0
* Correct reporting of file types. Normal files were being reported at sockets due to the socket mask being 2 bits and overlapping with a normal file bit
* Correct reporting of file permissions. Now correctly uses the linked file instead of the fd symlink name

Please note that this PR is reliant on #19956. Will need to rebase onto new master once it's in.


trufae:
I think `dd` in this case is used as a code generator. so it should be fine to access it without the debugger. but the actual problem is in `dx` which doesnt check or show any error if you are not in debugger mode.

Also, at some point it will be good to discuss a better way to handle emulation/debugging env switching because the `ar` command is duplicated under `dr` and some other step logic is also dubbed, not to say debug breakpoints only work with the esil debugger backend and not emulation commands, etc.. it will be good to find a better approach than the current one. thoughs on this?

Lazula:
r_debug_desc_list() dies if dbg->pid isnt alive. if we want to expose the other dd commands we can move the error message inside that function instead

```
$ r2 -
[0x00000000]> dd
opendir /proc/x/fd: No such file or directory
```

libr/core/cmd_debug.c
```diff
@@ -5037,6 +5037,10 @@ static int cmd_debug(void *data, const char *input) {
 			r_core_cmd_help (core, help_msg_dd);
 			break;
 		}
+		if (core->dbg->pid == -1) {
```
trufae, line 5040: usually we check if core->config have cfg.debug set to true, and there's also an api tlo chk if the debugger child is dead or not

trufae:
I think dx should check for this. And dd should run the code instead of printing it. The reason it was like this was for debugging purposes. To be able to copypaste that shellcode and disassemble it or test it without going into debugger mode.

Maybe we can do this change for 5.7:

* dont print the shellcode on dd commands. Just execute it
* Print it when * is suffixed
* dont allow to run dd without cfg.debug
* use else instead of two breaks in your checks

Lazula:
That all sounds good to me, I'll draft this PR till that's all ready and mark it as a breaking change

libr/core/cmd_debug.c
```diff
@@ -5039,9 +5039,17 @@ static int cmd_debug(void *data, const char *input) {
...
+			if (!r_config_get_b (core->config, "cfg.debug")) {
+				eprintf ("No child process to manage files for.\n");
+				break;
+			}
 			r_debug_desc_list (core->dbg, 0);
 			break;
 		case '*': // "dd*"
+			if (!r_config_get_b (core->config, "cfg.debug")) {
+				eprintf ("No child process to manage files for.\n");
+				break;
```
trufae, line 5051: I think is better to use else and have a single break statement for each case

trufae:
The "cool" thing of having dd to work without debugger mode is for tests . we can verify the egg generates the right assembly for example

trufae:
i dont think this change should wait for 5.7. its not that breaking.. as it was broken before

Lazula:
This PR will be updated to a full dd refactor/testing/bugfixing patch soon, there is one major blocking bug remaining in the syscall code

trufae:
Can you rebase + spuash some commits or you do you want me to squash them all into a single commit?

trufae:
Also solve conflicts :$

Lazula:
Conflicts were expected, I can squash-merge it after fixing and seeing CI pass.

Lazula:
No conflicts after rebasing, it was just GH not moving master forward on the PR. Should be good if CI still passes. archos/linux-x64 tests are still green locally.

PR #19881, opened 2022-03-27
Add an error return to r_core_project_cat ##projects

RHL120:

**Description**

### DIFF
--- a/libr/core/project.c
+++ b/libr/core/project.c
@@ -87,7 +87,7 @@ R_API bool r_core_is_project(RCore *core, const char *name) {
 }
 
 R_API void r_core_project_cat(RCore *core, const char *name) {
-	core->num->value = 1;
+				r_core_return_code (core, R_CMD_RC_FAILURE);
 	char *path = get_project_script_path (core, name);
 	if (path) {
 		char *data = r_file_slurp (path, NULL);
--- a/libr/core/project.c
+++ b/libr/core/project.c
@@ -87,7 +87,7 @@ R_API bool r_core_is_project(RCore *core, const char *name) {
 }
 
 R_API void r_core_project_cat(RCore *core, const char *name) {
-				r_core_return_code (core, R_CMD_RC_FAILURE);
+	r_core_return_code (core, R_CMD_RC_FAILURE);
 	char *path = get_project_script_path (core, name);
 	if (path) {
 		char *data = r_file_slurp (path, NULL);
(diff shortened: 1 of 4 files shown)


libr/core/project.c
```diff
@@ -87,16 +87,18 @@ R_API bool r_core_is_project(RCore *core, const char *name) {
 }
 
 R_API int r_core_project_cat(RCore *core, const char *name) {
```
trufae, line 89: I think it should return bool. The command handler who calls this function is the one that must change the return code

trufae:
Please rebase and try again when https://github.com/radareorg/radare2/pull/19897 gets merged

trufae:
I would like to move core/rvc.c into util/vc.c and rename include/rvc.h to include/r_util/vc.h

the only blocker for this seems to be only the dependency on r_hash. So i propose making this a function pointer that RCore can replace to substitute the default algorithm. Ideally there should be always the same checksum algorithm used everywhere.. so i'm not fully convinced about this solution. what do you think?

libr/core/project.c
```diff
@@ -86,17 +86,18 @@ R_API bool r_core_is_project(RCore *core, const char *name) {
...
-R_API int r_core_project_cat(RCore *core, const char *name) {
+R_API void r_core_project_cat(RCore *core, const char *name) {
+	core->num->value = 1;
 	char *path = get_project_script_path (core, name);
 	if (path) {
 		char *data = r_file_slurp (path, NULL);
 		if (data) {
 			r_cons_println (data);
 			free (data);
+			core->num->value = 0;
```
trufae, line 97:
```suggestion
			r_core_return_code (core, R_CMD_RC_SUCCESS);
```

libr/core/project.c
```diff
@@ -86,17 +86,18 @@ R_API bool r_core_is_project(RCore *core, const char *name) {
 	return ret;
 }
 
-R_API int r_core_project_cat(RCore *core, const char *name) {
+R_API void r_core_project_cat(RCore *core, const char *name) {
+	core->num->value = 1;
```
trufae, line 90:
```suggestion
				r_core_return_code (core, R_CMD_RC_FAILURE);
```

trufae (changes requested):
pls merge those changes and rebase

Lazula:
The test `Save PE project with maddr` will be fixed by #19965

PR #19885, opened 2022-03-27
Correct OS description

throwaway1037:
Also reduces file size :)

**Description**


trufae:
Thats not the same. HURD is 100% GNU, so its GNU/HURD. For Linux r2 runs on non-gnu linuxes too, like musl or other non-gnu userlands

throwaway1037:
How about "non-GNU+Linux, GNU+{Linux, Linux-libre,HURD}"?

PR #19888, opened 2022-03-28
Fix incorrect op-type assignment in anal_arm_cs

condret:

**Description**

`mov pc, lr` and conditional variants are rather rets/condrets than movs. This patch has huge impact on function analysis


trufae:
And that results in 0 breaking tests? Seems like we need more. Do you have a reproducer where analysis was failing?

Issue #19887, opened 2022-03-28
EPIC: Refactoring project - `c`


Lazula:
## Refactoring `c`

This issue tracks progress in the large project of giving the `c` command a full pass for code quality, bugfixing, and other improvements.

Following is a list of each subcommand. I'm creating the issue with only the top-level ones, but more sub-commands may be added for tracking specific progress.

## Subcommand progress

* [ ] c
* [ ] c*
* [ ] c1
* [ ] c2
* [ ] c4
* [ ] c8
* [ ] cat
* [ ] cc
* [ ] ccc
* [ ] ccd
* [ ] ccdd
* [ ] cd
* [ ] cf
* [ ] cg
* [ ] ci
* [ ] cl
* [X] cp - #19785
* [ ] cu
* [ ] cud
* [ ] cv[1248]
* [ ] cV[1248]
* [ ] cw
* [ ] cx
* [ ] cx*
* [ ] cX

trufae:
moved to 5.7.0 as its no blocker for pushing a minor release

PR #19889, opened 2022-03-28
Partial #19887 - Refactor, document, and test `cw` command

Lazula:
This command was originally implemented in 2012 but never documented or tested and the help text was sparse with information about what it actually did.

## Changelog

* Change the `R_API r_core_cmpwatch_*` functions to return bool for reporting success and failure.
  * The internal `cmd_cmp_watcher()` function now returns an int, so it can propagated through `cmd_cmp()`.
* Fix a bug that caused watchers to be partially overwritten and then duplicated in `core->watchers` when creating a new watcher at an address that already has one.
  * Watchers are now totally overwritten when this happens.
* Provide more verbose help text about watchers with `cw??`, including a reference to the book page.
* Add tests for `cw` and its subcommands.
* Ask user for confirmation before deleting/reverting all watchers - it's a potentially very destructive action that's very easy to do by accident.

### Book

I have created a page for comparison watchers in the book. See radareorg/radare2-book#334

### DIFF
--- a/libr/core/cmd_cmp.c
+++ b/libr/core/cmd_cmp.c
@@ -510,7 +510,7 @@ out_free_argv:
 	case '\0': { // "cw"
 		int mode;
 		if (*input) {
-			mode = input[1];
+			mode = *input;
 			if (input[1]) {
 				addr = r_num_get (core->num, input + 2);
 			}
--- a/libr/core/cmd_cmp.c
+++ b/libr/core/cmd_cmp.c
@@ -59,7 +59,11 @@ R_API R_BORROW RCoreCmpWatcher *r_core_cmpwatch_get(RCore *core, ut64 addr) {
 	return NULL;
 }
 
+#if R2_VERSION_MAJOR >= 5 && R2_VERSION_MINOR >= 7
 R_API bool r_core_cmpwatch_add(RCore *core, ut64 addr, int size, const char *cmd) {
+#else
+R_API int r_core_cmpwatch_add(RCore *core, ut64 addr, int size, const char *cmd) {
+#endif
 	RCoreCmpWatcher *cmpw;
 	bool found = false;
 	r_return_val_if_fail (core && cmd && size > 0, false);
@@ -104,7 +108,11 @@ R_API bool r_core_cmpwatch_add(RCore *core, ut64 addr, int size, const char *cmd
 	return true;
 }
 
+#if R2_VERSION_MAJOR >= 5 && R2_VERSION_MINOR >= 7
 R_API bool r_core_cmpwatch_del(RCore *core, ut64 addr) {
+#else
+R_API int r_core_cmpwatch_del(RCore *core, ut64 addr) {
+#endif
 	bool ret = false;
 	RCoreCmpWatcher *w;
 	RListIter *iter, *iter2;
@@ -129,7 +137,11 @@ R_API bool r_core_cmpwatch_del(RCore *core, ut64 addr) {
 	return ret;
 }
 
+#if R2_VERSION_MAJOR >= 5 && R2_VERSION_MINOR >= 7
 R_API bool r_core_cmpwatch_show(RCore *core, ut64 addr, int mode) {
+#else
+R_API int r_core_cmpwatch_show(RCore *core, ut64 addr, int mode) {
+#endif
 	RListIter *iter;
 	RCoreCmpWatcher *w;
 	PJ *pj = NULL;
@@ -206,7 +218,11 @@ static bool update_watcher(RIO *io, RCoreCmpWatcher *w) {
 }
 
 /* Replace old data with current new data, then read IO into new data */
+#if R2_VERSION_MAJOR >= 5 && R2_VERSION_MINOR >= 7
 R_API bool r_core_cmpwatch_update(RCore *core, ut64 addr) {
+#else
+R_API int r_core_cmpwatch_update(RCore *core, ut64 addr) {
+#endif
 	RCoreCmpWatcher *w;
 	RListIter *iter;
 	bool ret = false;
@@ -240,7 +256,11 @@ static bool revert_watcher(RCoreCmpWatcher *w) {
 }
 
 /* Mark the current old state as new, discarding the original new state */
+#if R2_VERSION_MAJOR >= 5 && R2_VERSION_MINOR >= 7
 R_API bool r_core_cmpwatch_revert(RCore *core, ut64 addr) {
+#else
+R_API int r_core_cmpwatch_revert(RCore *core, ut64 addr) {
+#endif
 	RCoreCmpWatcher *w;
 	RListIter *iter;
 	bool ret = false;
--- a/libr/core/cmd_cmp.c
+++ b/libr/core/cmd_cmp.c
@@ -44,6 +44,7 @@ static const char *help_msg_c[] = {
 R_API void r_core_cmpwatch_free(RCoreCmpWatcher *w) {
 	free (w->ndata);
 	free (w->odata);
+	free (w->cmd);
 	free (w);
 }
 
@@ -61,10 +62,7 @@ R_API RCoreCmpWatcher *r_core_cmpwatch_get(RCore *core, ut64 addr) {
 R_API bool r_core_cmpwatch_add(RCore *core, ut64 addr, int size, const char *cmd) {
 	RCoreCmpWatcher *cmpw;
 	bool found = false;
-	r_return_val_if_fail (core, false);
-	r_return_val_if_fail (cmd, false);
-	r_return_val_if_fail (strlen (cmd) < sizeof (cmpw->cmd), false);
-	r_return_val_if_fail (size > 0, false);
+	r_return_val_if_fail (core && cmd && size > 0, false);
 
 	cmpw = r_core_cmpwatch_get (core, addr);
 	if (!cmpw) {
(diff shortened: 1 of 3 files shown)


libr/core/cmd_cmp.c
```diff
@@ -304,59 +378,196 @@ static int radare_compare(RCore *core, const ut8 *f, const ut8 *d, int len, int
...
+	};
+	static const char *verbose_help_cw =
+		"Watchers are used to record memory at 2 different points in time, then\n"
+		"report if and how it changed. First, create one with `cw addr sz cmd`. This\n"
+		"will record sz bytes at addr. To record the second state, use `cwu`. Now, when\n"
+		"you run `cw`, the watcher will report if the bytes changed and run the command given\n"
+		"at creation with the size and address. You may overwrite any watcher by creating\n"
+		"another at the same address. This will discard the existing watcher completely.\n"
+		"\n"
+		"When you create a watcher, the data read from memory is marked as \"new\". Updating\n"
```
trufae, line 401: Use single quote ' instead of " so u dont have to escape it al the time

libr/core/cmd_cmp.c
```diff
@@ -58,103 +58,177 @@ R_API RCoreCmpWatcher *r_core_cmpwatch_get(RCore *core, ut64 addr) {
...
-R_API int r_core_cmpwatch_add(RCore *core, ut64 addr, int size, const char *cmd) {
+R_API bool r_core_cmpwatch_add(RCore *core, ut64 addr, int size, const char *cmd) {
 	RCoreCmpWatcher *cmpw;
-	if (size < 1) {
-		return false;
-	}
+	bool found = false;
+	r_return_val_if_fail (core, false);
+	r_return_val_if_fail (cmd, false);
+	r_return_val_if_fail (strlen (cmd) < sizeof (cmpw->cmd), false);
```
trufae, line 66: That cant make release builds safe from segfaults. There shouldnt be a limit imho

libr/core/cmd_cmp.c
```diff
@@ -58,103 +58,177 @@ R_API RCoreCmpWatcher *r_core_cmpwatch_get(RCore *core, ut64 addr) {
...
+	} else {
+		free (cmpw->odata);
+		free (cmpw->ndata);
+		found = true;
 	}
 	cmpw->size = size;
-	snprintf (cmpw->cmd, sizeof (cmpw->cmd), "%s", cmd);
+
+	// Size is checked in caller
+	strcpy (cmpw->cmd, cmd);
```
trufae, line 84: Dont use strcpy

libr/core/cmd_cmp.c
```diff
@@ -58,103 +58,177 @@ R_API RCoreCmpWatcher *r_core_cmpwatch_get(RCore *core, ut64 addr) {
...
+
+		ret = true;
 	}
-	return false;
+	return ret;
 }
 
-R_API int r_core_cmpwatch_update(RCore *core, ut64 addr) {
+static bool update_watcher(RIO *io, RCoreCmpWatcher *w) {
+	r_return_val_if_fail (io, false);
```
trufae, line 162: Just io && w no need for separate staments

libr/core/cmd_cmp.c
```diff
@@ -304,59 +378,196 @@ static int radare_compare(RCore *core, const ut8 *f, const ut8 *d, int len, int
...
+static int cmd_cmp_watcher(RCore *core, const char *input) {
+	static const char *help_msg_cw[] = {
+		"Usage: cw", "[args]", "Manage compare watchers; See if and how memory changes",
+		"cw??", "", "Show more info about watchers",
+		"cw ", "addr sz cmd", "Add a compare watcher",
+		"cw", "[*q] [addr]", "Show compare watchers (*=r2 commands, q=quiet)",
+		"cwd", " [addr]", "Delete watcher",
+		"cwr", " [addr]", "Revert watcher",
+		"cwu", " [addr]", "Update watcher",
+		NULL
```
trufae, line 391: Add cwj

libr/core/cmd_cmp.c
```diff
@@ -304,59 +378,196 @@ static int radare_compare(RCore *core, const ut8 *f, const ut8 *d, int len, int
...
+			} else if (size > INT_MAX) {
+				ret = 1;
+				eprintf ("Can't create a watcher with size larger than an int.\n");
+				goto out_free_argv;
+			}
+
+			// No good way to get sizeof(RCoreCmpWatcher.cmd)
+			// here, so this is hardcoded. There's an accurate
+			// debug check inside cmpwatch_add()
+			if (strlen (argv[2]) >= 32) {
```
trufae, line 450: Remove limits

trufae (changes requested):

trufae (changes requested):
Abi breaking changes must wait until 5.7. But thats a pretty good improvement . Good work!

trufae:
Considering int->bool shoudlt be breaking any abi for those symbols that arent used outside RCore, i think it's good to consider it as a non-abi-breaking change. what do you think?

Lazula:
> Considering int->bool shoudlt be breaking any abi for those symbols that arent used outside RCore, i think it's good to consider it as a non-abi-breaking change. what do you think?

Hm, I don't think it is actually, since despite using int was still using true/false to report, so only the type changed and not the actual return values. Is changing `cmd_cmp_watcher()` to return int instead of void considered breaking?

libr/core/cmd_cmp.c
Lazula: Double quotes in the output look better imo

libr/core/cmd_cmp.c
```diff
@@ -34,7 +34,7 @@ static const char *help_msg_c[] = {
 	"cud", " [addr] @at", "Unified diff disasm from $$ and given address",
 	"cv", "[1248] [hexpairs] @at", "Compare 1,2,4,8-byte (silent return in $?)",
 	"cV", "[1248] [addr] @at", "Compare 1,2,4,8-byte address contents (silent, return in $?)",
-	"cw", "[?] [us?] [...]", "Compare memory watchers",
+	"cw", "[*dqjru?] [addr]", "Compare memory watchers",
```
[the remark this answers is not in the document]
trufae: the "[?]" is mandatory, otherwise `?*` won't dig in.

libr/core/cmd_cmp.c
```diff
@@ -58,103 +59,203 @@ R_API RCoreCmpWatcher *r_core_cmpwatch_get(RCore *core, ut64 addr) {
...
 	if (!cmpw) {
 		cmpw = R_NEW (RCoreCmpWatcher);
 		if (!cmpw) {
 			return false;
 		}
 		cmpw->addr = addr;
+	} else {
+		free (cmpw->odata);
+		free (cmpw->ndata);
+		free (cmpw->cmd);
```
trufae, line 77:
```suggestion
		free (cmpw->cmd);
		free (cmpw);
		return false;
```

libr/core/cmd_cmp.c
```diff
@@ -58,103 +59,203 @@ R_API RCoreCmpWatcher *r_core_cmpwatch_get(RCore *core, ut64 addr) {
...
 	RCoreCmpWatcher *w;
 	RListIter *iter;
+	bool ret = false;
+
+	if (addr != UT64_MAX) {
+		w = r_core_cmpwatch_get (core, addr);
+		if (w) {
+			return update_watcher (core->io, w);
+		}
+
```
trufae, line 213:
```suggestion
```

libr/core/cmd_cmp.c
```diff
@@ -304,59 +405,189 @@ static int radare_compare(RCore *core, const ut8 *f, const ut8 *d, int len, int
...
+		argv = r_str_argv (input + 1, &argc);
+		if (!argv) {
+			return 1;
+		}
+
+		if (argc == 1) { // "cw [addr]"
+			addr = r_num_get (core->num, argv[0]);
+			r_core_cmpwatch_show (core, addr, 0);
+		} else if (argc == 3) { // "cw addr sz cmd"
+			addr = r_num_get (core->num, argv[0]);
```
trufae, line 461: any reason not to use r_num_math ?

trufae, line 461:
```suggestion
			addr = r_num_math (core->num, argv[0]);
```

libr/core/cmd_cmp.c
```diff
@@ -304,59 +405,189 @@ static int radare_compare(RCore *core, const ut8 *f, const ut8 *d, int len, int
...
+			size = r_num_get (core->num, argv[1]);
```
trufae, line 462:
```suggestion
			size = r_num_math (core->num, argv[1]);
```

libr/core/cmd_cmp.c
```diff
@@ -304,59 +405,189 @@ static int radare_compare(RCore *core, const ut8 *f, const ut8 *d, int len, int
...
+
+			if (size < 1) {
+				ret = 1;
+				eprintf ("Can't create a watcher with size less than 1.\n");
+				goto out_free_argv;
+			} else if (size > INT_MAX) {
```
trufae, line 468: use r2 constants INT_MAX can be different depending on compilter. kind of ST32_MAX ?

trufae, line 468: dont else after goto

trufae (dismissed):

libr/core/cmd_cmp.c
```diff
@@ -58,103 +59,203 @@ R_API RCoreCmpWatcher *r_core_cmpwatch_get(RCore *core, ut64 addr) {
...
 		free (cmpw);
 		return false;
 	}
 	r_io_read_at (core->io, addr, cmpw->ndata, size);
-	r_list_append (core->watchers, cmpw);
+
+	// Don't append a duplicate
+	if (!found) {
+		r_list_append (core->watchers, cmpw);
+	}
```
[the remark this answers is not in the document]
Lazula: if found, we're working on a borrowed cmpw from r_core_cmpwatch_get(), which is why we don't append.

libr/core/cmd_cmp.c
Lazula: we're overwriting the watcher, not removing it

libr/core/cmd_cmp.c
Lazula: get is generally preferred afaik

libr/core/cmd_cmp.c
Lazula: size is just an int, and INT_MAX should match it. its kept as an int because that's what r_io_read_at() uses and i didnt want to potentially break that

libr/core/cmd_cmp.c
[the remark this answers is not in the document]
Lazula: cmpwatch_get() is now marked as returning `R_BORROW RCoreCmpwatcher *` to clarify this

Lazula:
Will wait till we can merge the book changes at the same time.

PR #19893, opened 2022-03-29
Fix bug in z/, that creates misplaced functions ##zign

condret:

**Description**


test/db/cmd/cmd_zignature
```diff
@@ -409,7 +409,7 @@ aa
 axt 0x400536
 EOF
 EXPECT=<<EOF
-main 0x40056f [CALL] call sign.bytes.sym.print_0
+main 0x40056f [CALL] call sym.print
```
trufae, line 412: Thats a regression. The test was correct before your change imho

PR #19897, opened 2022-03-29
Ship RCmd.returnValue into RCore.num.value ##shell

trufae:

**Description**

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -4480,7 +4480,10 @@ fuji:
 		r_str_trim_head (cmd);
 		rc = r_cmd_call (core->rcmd, cmd);
 	} else {
-		rc = false;
+		rc = 0;
+	}
+	if (rc == 1) {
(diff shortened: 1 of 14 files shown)


trufae:
no checks running. thank you github actions! 👋

trufae:
There's a conceptual problem here because num->value can contain ut64 and cmd returns int. Handling negative values for errors. So maybe we should only update this value when the cmd returns 0 or 1? And let the rest of cases for the function to update the value as its will:

-2 : exit r2
-1 : error happened, breaks loops
0: all is fine (corenunval=1)
1: corenunval =1

any other number may not update the corenumval. We may also use a define/enum for all those 5 possible cases:

typedef enum {
  R_CMD_RETVAL_QUIT=-2,
  R_CMD_RETVAL_ERROR=-1
  R_CMD...
} RCmdRetval;

To solve the long name problem we can have a macro

R_CMD_RETURN (OK);

PR #19909, opened 2022-04-01
Fix autoloading of ##signatures

swoops:

**Description**

Fixes `dir->zigns` typo and avoids a warning from `r_sys_dir` if `!path`. I also pulled it as out as a static inline function to simplify `r_main_radare2`.

### DIFF
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -284,6 +284,31 @@ static bool mustSaveHistory(RConfig *c) {
 	return true;
 }
 
+static inline void autoload_zigns(RCore *r) {
+	if (r_config_get_i (r->config, "zign.autoload")) {
+		char *path = r_file_abspath (r_config_get (r->config, "dir.zigns"));
+		if (!path) {
+			return;
+		}
+		RList *list = r_sys_dir (path);
+		RListIter *iter;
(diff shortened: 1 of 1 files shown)


libr/main/radare2.c
```diff
@@ -284,6 +284,31 @@ static bool mustSaveHistory(RConfig *c) {
 	return true;
 }
 
+static inline void autoload_zigns(RCore *r) {
```
trufae, line 287: Usually is better to move this check outside the tunction. And use _b

libr/main/radare2.c
```diff
@@ -284,6 +284,31 @@ static bool mustSaveHistory(RConfig *c) {
...
+	if (r_config_get_i (r->config, "zign.autoload")) {
+		char *path = r_file_abspath (r_config_get (r->config, "dir.zigns"));
+		if (!path) {
+			return;
+		}
+		RList *list = r_sys_dir (path);
+		RListIter *iter;
+		char *file;
+		r_list_foreach (list, iter, file) {
```
trufae, line 296: Use R_STR_ISNOTEMPTY

trufae (changes requested):
.

PR #19917, opened 2022-04-04
Update byte signature flag name ##signatures

swoops:

**Description**

Byte signatures can match in one of two ways. First by searching with r_search, or by matching against a function. There are different implications and desires for each.

This update will cause function matches to get `bytes_func` flags and the r_search version will stay as `bytes`.


trufae:
@condret ?

trufae:
Please update the tests but i dont see the reasoning behind changing the name could fix

trufae:
Add a test to repro

condret:
oh, i didn't see the mention. The reasoning behind this is that I got flag duplicates and I wanted to do @@ iteration over them. the duplicates were kinda noisy

trufae:
So you think we need a @@= that does sort -u?

PR #19929, opened 2022-04-09
Honor escaping semicolons in macro definitions ##shell

trufae:
```
[0x00000000]> (x\;?e hello\;?e world)
[0x00000000]> .(x)
No macro named 'x'
[0x00000000]> (
0 (x\ ; ?e hello\; ?e world)
[0x00000000]>
```

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -1,4 +1,4 @@
-/* radare - LGPL - Copyright 2009-2021 - nibble, pancake */
+/* radare - LGPL - Copyright 2009-2022 - nibble, pancake */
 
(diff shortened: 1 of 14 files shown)


trufae:
This can be probably escaped at a highr level place, but as long as we have tests and the syntax parser is still evolving we can move forward. because that also affects other commands like this:

```
[0x100003b34]> ?e hello\; world
hello\; world
[0x100003b34]>
```

trufae:
also fixes a segfault in w and ?e because both commands were calling the old trim.args() function which had an oobread crash bug

PR #19952, opened 2022-04-13
fix loongarch meson script miss

junchao-loongson:

I found that the meson build of radare2 does not seem to include anal_loongarch_gnu
When I use r2（build with meson） to analyze the loongarch binary I get the following message
```
anal.arch: cannot find (loongarch)
anal.arch: cannot find (loongarch)
asm.arch: cannot find (loongarch)
asm.arch: cannot find (loongarch)
```
I fixed the issue with anal in the patch, but I'm not sure how to fix the issue with asm, because the asm function is included in anal in loongarch, and there is no separate asm module

### DIFF
--- a/libr/meson.build
+++ b/libr/meson.build
@@ -109,6 +109,9 @@ if not no_user_plugins
     anal_plugins += [ 'riscv' ]
     parse_plugins += [ 'riscv_pseudo' ]
   endif
+  if user_plugins.contains('loongarch')
+    anal_plugins += [ 'loongarch_gnu' ]
+  endif
   if user_plugins.contains('elf')
     bin_plugins += [ 'elf', 'elf64' ]
   endif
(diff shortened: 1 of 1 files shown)


libr/meson.build
```diff
@@ -109,6 +109,9 @@ if not no_user_plugins
     anal_plugins += [ 'riscv' ]
     parse_plugins += [ 'riscv_pseudo' ]
   endif
+  if user_plugins.contains('loongarch')
```
radare, line 112: i think it's better to include it under `use_gpl` too

junchao-loongson: ok！
and “asm.arch: cannot find (loongarch)” how do I handle it, is there a reference code?

radare: let me merge and try the error you mention.

PR #19957, opened 2022-04-13
Improve `wz` help and error handling

### LINKED ISSUE #19891: `wz` handles arguments loosely or inaccurately
`wz` with no arguments fails and incorrectly reports a permission problem

```
$ r2 -
[0x00000000]> o
 3 * rwx 0x00000200 malloc://512
[0x00000000]> wz
ERROR: Cannot write in here, check map permissions or reopen the file with oo+
```

Not using a space between `wz` and the string is ignored completely.

```
[0x00000000]> wz1A
[0x00000000]> p8 4
41000000
```

No help text from `wz?`

```
[0x00000000]> wz?
```

Lazula:
Fixes #19891.

```
$ r2 -
[0x00000000]> o
 3 * rwx 0x00000200 malloc://512
[0x00000000]> wz
| wz string  write zero terminated string (like w + \x00)
[0x00000000]> wz1A
| wz string  write zero terminated string (like w + \x00)
[0x00000000]> wz?
| wz string  write zero terminated string (like w + \x00)
```

### DIFF
--- a/libr/core/cmd_write.c
+++ b/libr/core/cmd_write.c
@@ -1424,8 +1424,7 @@ static int w_handler(RCore *core, const char *input) {
 	return 0;
 }
 
-static int wz_handler_old(void *data, const char *input) {
-	RCore *core = (RCore *)data;
+static int wz_handler(RCore *core, const char *input) {
 	int wseek = r_config_get_i (core->config, "cfg.wseek");
 	char *str = r_str_trim_dup (input);
 	int len = r_str_unescape (str) + 1;
@@ -1440,11 +1439,7 @@ static int wz_handler_old(void *data, const char *input) {
 	if (!r_core_write_at (core, core->offset, (const ut8 *)str, len)) {
 		cmd_write_fail (core);
 	}
-	if (len > 0) {
-		r_core_return_code (core, len);
-	} else {
-		r_core_return_code (core, 0);
-	}
+	r_core_return_code (core, len);
 	WSEEK (core, len + 1);
 	r_core_block_read (core);
 	free (str);
@@ -2176,7 +2171,7 @@ static int cmd_write(void *data, const char *input) {
 		break;
 	}
 	case 'z': // "wz"
-		wz_handler_old (core, input + 1);
+		wz_handler (core, input + 1);
 		break;
 	case 't': // "wt"
 		wt_handler_old (core, input + 1);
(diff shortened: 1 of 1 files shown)


libr/core/cmd_write.c
```diff
@@ -1427,10 +1427,19 @@ static int w_handler(RCore *core, const char *input) {
 static int wz_handler_old(void *data, const char *input) {
 	RCore *core = (RCore *)data;
 	int wseek = r_config_get_i (core->config, "cfg.wseek");
-	char *str = strdup (input);
+	char *str;
```
trufae, line 1430: Define and assign in the same line. Its in the coding style guidelines

libr/core/cmd_write.c
```diff
@@ -1427,10 +1427,19 @@ static int w_handler(RCore *core, const char *input) {
...
+	int len;
+
 	/* write zero-terminated string */
-	int len = r_str_unescape (str);
-	if (!r_core_write_at (core, core->offset, (const ut8 *)str + 1, len)) {
+	str = strdup (r_str_trim_head_ro (input));
```
trufae, line 1434: Theres r_str_trim_dup for this

libr/core/cmd_write.c
```diff
@@ -1427,10 +1427,19 @@ static int w_handler(RCore *core, const char *input) {
...
+	len = r_str_unescape (str) + 1;
```
trufae, line 1435: Define and assign in the same line

libr/core/cmd_write.c
```diff
@@ -1427,10 +1427,19 @@ static int w_handler(RCore *core, const char *input) {
...
+	if (*input == '?' || *input != ' ' || len == 0) {
```
trufae, line 1436: Check for len <1 as the type is signed

trufae (changes requested):

libr/core/cmd_write.c
Lazula: left over from an older patch oops

trufae:
i had the same broken test (smd) in https://github.com/radareorg/radare2/pull/19950 while triming string hits. maybe we can do this change for 5.7 or add an option. what do you think? strings can start with spaces.. but this is noisy for flags and a little unreadable. so maybe its just something to handle for printing, not internally

Lazula:
Didn't know that was being used as intended behavior. Trailing spaces are being used too. For 5.7 we could make trimming the default and add an option. Will revert that change for now.

PR #19980, opened 2022-04-19
Fix #19966 - Reset seek in r_debug_execute() to real PC

### LINKED ISSUE #19966: r_debug_execute incorrectly advances seek
## Environment

```sh
$ date && r2 -v && uname -ms
Sun 17 Apr 2022 03:13:43 PM CDT
radare2 5.6.7 30157 @ linux-x86-64 git.5.6.6
commit: 7435f879820c47a9dfe144c16a345bbff5d18cd6 build: 2022-04-17__15:10:06
Linux x86_64
```

## Description

`r_debug_execute()` does not properly reset seek after hitting the breakpoint in the overwritten code at the program counter. The program counter still rests at the original PC, but seek is left at the breakpoint.

### Actual behavior

```
$ r2 -N -Q -c 's; dx 90909090 2>/dev/null; s; doc' -d test/bins/elf/ls-focal
0x7fa48ad6e100
0x7fa48ad6e105
```

### Expected behavior

```
$ r2 -N -Q -c 's; dx 90909090 2>/dev/null; s; doc' -d test/bins/elf/ls-focal
0x7fa48ad6e100
0x7fa48ad6e100
```

Lazula:
Fix #19966

Other bugfixes and API changes:

* `r_debug_execute()` returns bool instead of ut64 to report if code was successfully run
  * Return value is now an output arg
* Fix return value coming from register A0 (arg) instead of R0 (ret)
* Fix stack restoration - always restore the full 4k

### DIFF
--- a/libr/debug/debug.c
+++ b/libr/debug/debug.c
@@ -515,9 +515,10 @@ R_API bool r_debug_set_arch(RDebug *dbg, const char *arch, int bits) {
  * The bytes overwritten at the program counter are always restored.
  *
  * TODO: Add support for reverse stack architectures
+ *
+ * XXX: This function will advance your seek to the end of the injected code.
  */
 R_API bool r_debug_execute(RDebug *dbg, const ut8 *buf, int len, R_OUT ut64 *ret, bool restore, bool ignore_stack) {
-	RCore *core;
 	ut8 stack_backup[4096];
 	ut8 *pc_backup = NULL, *reg_backup = NULL;
 	int reg_backup_sz;
@@ -525,7 +526,6 @@ R_API bool r_debug_execute(RDebug *dbg, const ut8 *buf, int len, R_OUT ut64 *ret
 	ut64 reg_sp, reg_pc, bp_addr;
 
 	r_return_val_if_fail (dbg && buf && len > 0, false);
-	core = dbg->corebind.core;
 
 	if (r_debug_is_dead (dbg)) {
 		return false;
@@ -602,11 +602,6 @@ R_API bool r_debug_execute(RDebug *dbg, const ut8 *buf, int len, R_OUT ut64 *ret
 
 	free (pc_backup);
 	free (reg_backup);
-
-	/* Rewind seek from breakpoint address to real PC */
-	if (core) {
-		core->offset = reg_pc;
-	}
 
 	return true;
 }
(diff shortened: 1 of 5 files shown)


trufae:
you cant call core apis from debug `/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/9/../../../../lib/libr_debug.so: undefined reference to `r_core_seek'
[758](https://github.com/radareorg/radare2/runs/6081628303?check_suite_focus=true#step:7:758)`

Lazula:
Would it work to just set core->offset then?

libr/debug/debug.c
```diff
@@ -516,70 +516,99 @@ R_API bool r_debug_set_arch(RDebug *dbg, const char *arch, int bits) {
...
 	ut8 stack_backup[4096];
 	ut8 *pc_backup = NULL, *reg_backup = NULL;
-	RRegItem *ri, *risp, *ripc;
-	ut64 rsp, rpc, ra0 = 0LL;
+	int reg_backup_sz;
+	RRegItem *ri_sp, *ri_pc, *ri_ret;
+	ut64 reg_sp, reg_pc, bp_addr;
+
+	r_return_val_if_fail (dbg && buf && len > 0, false);
+	core = dbg->corebind.core;
```
trufae, line 528: the idea behind corebind is that the RCore structure is opaque. so that is should be solved that way. why is the current seek changing from core? rdebug shouldnt be the one doing this if you restore the PC register after that..

Also. you can add an RCoreSeek() caallback in the RCoreBind structure and do `dbg->corebind.seek()` instead of accessing the offset field directly. That would be more legal.. but i dont think debug should be changing the core.offset

Lazula: > rdebug shouldnt be the one doing this if you restore the PC register after that

The problem is that the PC is restored, but seek is not. I believe that seek is getting forced forward when a breakpoint is hit, which seems incorrect unless we simply accept it and then move it in the command function. But it's also used elsewhere such as in native debug code, which would also result in seek being moved unexpectedly.

> but i dont think debug should be changing the core.offset

That's what I'm trying to fix. Maybe we should be able to just not move seek when a breakpoint is hit but that would probably require a lot more effort.

trufae: then you may want to restore the seek in RCore afer calling this function. Core is the only one who should be updating the core->offset variable

Lazula: We can just do it in cmd_debug() then, and note the seek behavior.

PR #19984, opened 2022-04-20
Fix for --without-pull not working in install.sh

Kruat:

**Description**

### DIFF
--- a/sys/install.sh
+++ b/sys/install.sh
@@ -66,7 +66,7 @@ cd "$(dirname $0)"/..
 pwd
 
 # update
-if [ -z $WITHOUT_PULL ]; then
+if [ -z "$WITHOUT_PULL" ]; then
 	if [ -d .git ]; then
 		git branch | grep "^\* master" > /dev/null
 		if [ $? = 0 ]; then
--- a/sys/install.sh
+++ b/sys/install.sh
@@ -34,7 +34,6 @@ while : ; do
 		;;
 	"--without-pull")
 		export WITHOUT_PULL=1
-		continue
 		;;
 	-*)
 		# just for the penguin face case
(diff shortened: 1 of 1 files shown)


sys/install.sh
```diff
@@ -32,6 +32,11 @@ while : ; do
 		shift
 		continue
 		;;
+	"--without-pull")
+		export WITHOUT_PULL=1
+		shift
```
trufae, line 37:
```suggestion
```

sys/install.sh
```diff
@@ -32,6 +32,11 @@ while : ; do
...
+		continue
```
trufae, line 38:
```suggestion
```

sys/install.sh
```diff
@@ -62,7 +67,7 @@ cd "$(dirname $0)"/..
 pwd
 
 # update
-if [ "$1" != "--without-pull" ]; then
+if [ -z $WITHOUT_PULL ]; then
```
trufae, line 70:
```suggestion
if [ -z "$WITHOUT_PULL" ]; then
```

trufae (changes requested):

sys/install.sh
Kruat: why? the original code shifts it away. Do you want to pass it on thru ARGS?

sys/install.sh
Kruat: the continue is important if we do shift, to avoid edge cases.

trufae: After the case theres a shift and the continue is implicit because it loops again

sys/install.sh
trufae: Uhm maybe i missread the patch :? Ill check after lunch

sys/install.sh
Kruat: yeah, sorry, you're in the right.

PR #20018, opened 2022-04-29
Add 8051 disassemble/reassemble tests on realtek firmware.

gogo2464:

**Description**

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -1,4 +1,4 @@
-all: r2r-tests unit_tests
+all: r2r-tests unit_tests test-8051-reassembly
 
 arm.v35:
 	${RUNTEST} db/extras/asm/arm.v35_64
@@ -16,6 +16,9 @@ RUNTEST=r2r -L -o results.json
 
 r2r-tests: bins
 	${RUNTEST}
+
+test-8051-reassembly:
+	$(SHELL) ./test-8051-reassembly.sh
 
 fuzz-tests: bins
 	r2r -LF bins/fuzzed @fuzz
--- a/test/test-8051-reassembly.sh
+++ b/test/test-8051-reassembly.sh
@@ -1,6 +1,6 @@
 BUILD_DIRECTORY=$(mktemp -d tmp.XXXXXXXXXX)
 [ -z "$BUILD_DIRECTORY" ] && exit 1
-export EXIT_CODE = 0
+export EXIT_CODE=0
 
 printf "Reassembling 8051... "
 radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_NIC.bin > $BUILD_DIRECTORY/MP_8192E_FW_NIC.asm || exit 1
--- a/test/test-8051-reassembly.sh
+++ b/test/test-8051-reassembly.sh
@@ -7,7 +7,7 @@ radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_N
 rasm2 -a 8051 -o 0x8000 -f $BUILD_DIRECTORY/MP_8192E_FW_NIC.asm -B  > $BUILD_DIRECTORY/recompiled-firmware.bin || exit 1
 
 cmp ./bins/8051/MP_8192E_FW_NIC.bin $BUILD_DIRECTORY/recompiled-firmware.bin
-if [ $? -eq 0 ]
+if [ $? -eq 0 ]; then
 	echo "OK"
 	EXIT_CODE=0
 else
--- a/test/test-8051-reassembly.sh
+++ b/test/test-8051-reassembly.sh
@@ -1,0 +1,21 @@
+BUILD_DIRECTORY=$(mktemp -d tmp.XXXXXXXXXX)
+[ -z "$BUILD_DIRECTORY" ] && exit 1
+export EXIT_CODE = 0
+
+printf "Reassembling 8051... "
+radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_NIC.bin > $BUILD_DIRECTORY/MP_8192E_FW_NIC.asm || exit 1
+rasm2 -a 8051 -o 0x8000 -f $BUILD_DIRECTORY/MP_8192E_FW_NIC.asm -B  > $BUILD_DIRECTORY/recompiled-firmware.bin || exit 1
+
+cmp ./bins/8051/MP_8192E_FW_NIC.bin $BUILD_DIRECTORY/recompiled-firmware.bin
+if [ $? -eq 0 ]
+	echo "OK"
+	EXIT_CODE=0
+else
+	radiff2 -q -D ./bins/8051/MP_8192E_FW_NIC.bin "$BUILD_DIRECTORY/recompiled-firmware.bin"
+	echo "failed"
+	EXIT_CODE=1
+fi
+
+rm -rf $BUILD_DIRECTORY
+
+exit $EXIT_CODE


test/Makefile
```diff
@@ -50,6 +62,8 @@ vc4: bins
 yara: bins
 	${RUNTEST} db/extras/cmd/yara
 
+
```
trufae, line 65: delete those empty lines

test/Makefile
```diff
@@ -17,6 +17,18 @@ RUNTEST=r2r -L -o results.json
 r2r-tests: bins
 	${RUNTEST}
 
+8051: bins
+	radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_NIC.bin > MP_8192E_FW_NIC.asm
+	rasm2 -a 8051 -o 0x8000 -f MP_8192E_FW_NIC.asm -B  > recompiled-firmware.bin
+	if [ "$(diff -s ./bins/8051/MP_8192E_FW_NIC.bin recompiled-firmware.bin)" != "Binary files ./bins/8051/MP_8192E_FW_NIC.bin and recompiled-firmware.bin differ" ]
```
trufae, line 23: use `cmp` instead. `diff` is not portable

test/Makefile
```diff
@@ -17,6 +17,18 @@ RUNTEST=r2r -L -o results.json
 r2r-tests: bins
 	${RUNTEST}
 
+8051: bins
+	radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_NIC.bin > MP_8192E_FW_NIC.asm
```
trufae, line 21: make clean should remove this .asm file

test/Makefile
```diff
@@ -17,6 +17,18 @@ RUNTEST=r2r -L -o results.json
 r2r-tests: bins
 	${RUNTEST}
 
+8051: bins
```
trufae, line 20: run this target from another make target or in the ci, so we can see if the test works before mergin

trufae (changes requested):
.

test/Makefile
gogo2464: I am 100% sure it works. When I test locally it works. I also pushed a very similar test on another repo. See: https://github.com/gogo2464/realtek-firmware-decompilation/blob/master/.github/workflows/.travis.yml#L24

gogo2464: Ah yes I just mentionned the issue with another architecture.

gogo2464:
The http file takes too long I am not responsible of that.

Makefile
```diff
@@ -391,6 +391,7 @@ shot:
 		radare.org:/srv/http/radareorg/get/shot
 
 tests test:
+	export LC_ALL=C
```
trufae, line 394: That line those nothing. The very next line wont maintain the same environment

also its wrong to override user settings. If you have warnjngs fix your system

test/Makefile
trufae: Well. You cant be sure if that will break tomorrow. Theres no sense in adding tests that never run

trufae (changes requested):
Which http are you talking about

Makefile
gogo2464: It is to ensure that local tests of everyone running tests have same setting.

gogo2464:
By http I mean the failling test.

Makefile
trufae: no. this line is useless. try this if my epxlanation above wasnt clear

```
$ make
export JEJE=123
echo $JEJE

$ cat Makefile
all:
	export JEJE=123
	echo $$JEJE
$
```

test/Makefile
```diff
@@ -17,6 +17,18 @@ RUNTEST=r2r -L -o results.json
...
+	radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_NIC.bin > ./tmp/MP_8192E_FW_NIC.asm
+	rasm2 -a 8051 -o 0x8000 -f ./tmp/MP_8192E_FW_NIC.asm -B  > ./tmp/recompiled-firmware.bin
+	if [ "$(cmp ./bins/8051/MP_8192E_FW_NIC.bin ./tmp/recompiled-firmware.bin)" != "./bins/8051/MP_8192E_FW_NIC.bin ./tmp/recompiled-firmware.bin differ:"* ]
```
trufae, line 23: thats not valid makefile. please read the other comments and address them

trufae:
The simplest way to move on here is just to use a shellscript in `test/test-8051-reassembly.sh` (and call it from the makefile). because makefiles are not meant to be used as shellscript replacements. the logic is different

Makefile
gogo2464: ah ok

test/Makefile
gogo2464: Nice catch. I decided to put the .asm file in the ./tmp folder to delete it.

gogo2464:
Hard to test with no master running.

trufae:
the ci is green again you can rebase if you like the green color

gogo2464:
I prefer the red color. But I also prefer when the code is running. Then let's rebase. :)

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,12 @@
+mkdir -p ../tmp
```
trufae, line 1:
```
#!/bin/sh

T=$(mktemp -d tmp.XXXXXX)
[ -z "$T" ] && exit 1
mkdir -p $T
....

rm -rf $T
```

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,12 @@
...
+
+radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_NIC.bin > ../tmp/MP_8192E_FW_NIC.asm
+rasm2 -a 8051 -o 0x8000 -f ../tmp/MP_8192E_FW_NIC.asm -B  > ../tmp/recompiled-firmware.bin
+if [ "$(cmp ./bins/8051/MP_8192E_FW_NIC.bin ../tmp/recompiled-firmware.bin)" != "./bins/8051/MP_8192E_FW_NIC.bin ../tmp/recompiled-firmware.bin differ:"* ]
```
trufae, line 5: just compare the return code. the output may differ between implementations

```
if [ $? = 0 ]; then ,...

```

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,12 @@
+mkdir -p ../tmp
+
+radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_NIC.bin > ../tmp/MP_8192E_FW_NIC.asm
```
trufae, line 3: check return code with `|| exit 1`

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,12 @@
...
+rasm2 -a 8051 -o 0x8000 -f ../tmp/MP_8192E_FW_NIC.asm -B  > ../tmp/recompiled-firmware.bin
```
trufae, line 4: check error code with || exit 1

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,12 @@
+mkdir -p ../tmp
+
```
trufae, line 2:
```suggestion
printf "Reassembling 8051... "

```

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,12 @@
...
+radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_NIC.bin > ../tmp/MP_8192E_FW_NIC.asm
+rasm2 -a 8051 -o 0x8000 -f ../tmp/MP_8192E_FW_NIC.asm -B  > ../tmp/recompiled-firmware.bin
+if [ "$(cmp ./bins/8051/MP_8192E_FW_NIC.bin ../tmp/recompiled-firmware.bin)" != "./bins/8051/MP_8192E_FW_NIC.bin ../tmp/recompiled-firmware.bin differ:"* ]
+then
+	echo "OK"
+else
+	radiff2 -q -D ./bins/8051/MP_8192E_FW_NIC.bin ../tmp/recompiled-firmware.bin
+	echo "failed"
+	exit -1
```
trufae, line 11:
```suggestion
	exit 1
```

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,12 @@
+mkdir -p ../tmp
+
+radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_NIC.bin > ../tmp/MP_8192E_FW_NIC.asm
+rasm2 -a 8051 -o 0x8000 -f ../tmp/MP_8192E_FW_NIC.asm -B  > ../tmp/recompiled-firmware.bin
+if [ "$(cmp ./bins/8051/MP_8192E_FW_NIC.bin ../tmp/recompiled-firmware.bin)" != "./bins/8051/MP_8192E_FW_NIC.bin ../tmp/recompiled-firmware.bin differ:"* ]
+then
+	echo "OK"
```
trufae, line 7:
```suggestion
	echo "OK"
	exit 0
```

trufae (changes requested):

gogo2464:
Ready to merge but it would be better if I did not disassemble/reassemble the offset.

test/Makefile
```diff
@@ -1,4 +1,8 @@
-all: r2r-tests unit_tests
+all:
+	export LC_ALL=C
```
trufae, line 2:
```suggestion
```

test/Makefile
```diff
@@ -1,4 +1,8 @@
...
+	$(MAKE) r2r-tests
+	$(MAKE) unit_tests
+	$(MAKE) test-8051-reassembly
```
trufae, line 5: why dont run those 3 targets in parallel as it was doing before? please change this

test/Makefile
```diff
@@ -17,6 +21,9 @@ RUNTEST=r2r -L -o results.json
 r2r-tests: bins
 	${RUNTEST}
 
+test-8051-reassembly:
+	sh ./test-8051-reassembly.sh
```
trufae, line 25:
```suggestion
	$(SHELL) ./test-8051-reassembly.sh
```

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,20 @@
+BUILD_DIRECTORY=$(mktemp -d tmp.XXXXXXXXXX)
+[ -z "$BUILD_DIRECTORY" ] && exit 1
+mkdir -p $BUILD_DIRECTORY
```
trufae, line 3: mktemp already creates the directory, this line is not necessary
```suggestion
```

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,20 @@
...
+
+printf "Reassembling 8051... "
+radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_NIC.bin > $BUILD_DIRECTORY/MP_8192E_FW_NIC.asm || exit 1
+rasm2 -a 8051 -o 0x8000 -f $BUILD_DIRECTORY/MP_8192E_FW_NIC.asm -B  > $BUILD_DIRECTORY/recompiled-firmware.bin || exit 1
+
+cmp ./bins/8051/MP_8192E_FW_NIC.bin $BUILD_DIRECTORY/recompiled-firmware.bin
+if [ $? -eq 0 ]
```
trufae, line 10:
```suggestion
if [ $? -eq 0 ]; then
```

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,20 @@
...
+then
```
trufae, line 11:
```suggestion
```

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,20 @@
...
+	echo "OK"
+	exit 0
+else
+	radiff2 -q -D ./bins/8051/MP_8192E_FW_NIC.bin "$BUILD_DIRECTORY/recompiled-firmware.bin"
+	echo "failed"
+	exit 1
+fi
+
+rm -rf $BUILD_DIRECTORY
```
trufae, line 20: This line is never executed because you exit before leaving. use RC=0 and then exit $RC

trufae (changes requested):

gogo2464:
A bit different than expected. You could merge anyway.

trufae:
Stop saying me to merge because your script is not even able to execute

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,21 @@
+BUILD_DIRECTORY=$(mktemp -d tmp.XXXXXXXXXX)
+[ -z "$BUILD_DIRECTORY" ] && exit 1
+export EXIT_CODE = 0
```
trufae, line 3:
```suggestion
export EXIT_CODE=0
```

trufae:
I think it should be fine now. Lets wait for the ci. But if the script fails to run because the reassembling is not correct then i dont see the point on having this test

test/test-8051-reassembly.sh
```diff
@@ -0,0 +1,21 @@
+BUILD_DIRECTORY=$(mktemp -d tmp.XXXXXXXXXX)
+[ -z "$BUILD_DIRECTORY" ] && exit 1
+export EXIT_CODE=0
+
+printf "Reassembling 8051... "
+radare2 -a 8051 -m 0x8000 -e "scr.color=0" -qf -c "pI" ./bins/8051/MP_8192E_FW_NIC.bin > $BUILD_DIRECTORY/MP_8192E_FW_NIC.asm || exit 1
+rasm2 -a 8051 -o 0x8000 -f $BUILD_DIRECTORY/MP_8192E_FW_NIC.asm -B  > $BUILD_DIRECTORY/recompiled-firmware.bin || exit 1
+
+cmp ./bins/8051/MP_8192E_FW_NIC.bin $BUILD_DIRECTORY/recompiled-firmware.bin
+if [ $? -eq 0 ]
```
trufae, line 10:
```suggestion
if [ $? -eq 0 ]; then
```

trufae (commented):
will be good if you try to run the script before pushing a change. or at least apply my suggestions instead of adding more breaking changes

PR #20021, opened 2022-04-30
Increase project name max length ##projects

mdolmen:

**Description**

If there is no reason not to I'd like to increase the maximum length of a project name. It's particularly annoying as of know when working with files which must respect a certain naming convention. I have some scripts where I automate the analysis and creation of the project and I'd like for the project to have the same name as the file.


trufae:
Sounds fair

PR #20031, opened 2022-05-02
Support assemble for mul, udiv, sdiv, lsl, lsr, mvn, tst arm64 instructions ##assembler

MohamedX99:
Support assemble for mul, udiv, sdiv, lsl, lsr, mvn, tst arm64 instructions


trufae:
Cherrypicked into master! thanks!

PR #20080, opened 2022-05-10
Add pcg to print array in Golang format

AandersonL:
Add a new print support for Golang variable slices, that's very useful when embedding files/shellcodes inside the source.

- [X ] Mark this if you consider it ready to merge

```go
[0x00005bc0]> pcg 10
var BUFF = [10]byte{0xf3,0xf,0x1e,0xfa,0x31,0xed,0x49,0x89,0xd1,0x5e}
```


trufae:
Looks good! I missed that one indeed!

PR #20083, opened 2022-05-10
Unconditionally compile r_magic_from_ebcdic

ypsvlq:

**Description**

was 12 seconds too slow to add this to the previous PR


trufae:
We may probably add more ci targets to check system magic and system ssl to avoid such problems in the future. thanks for taking care!

PR #20101, opened 2022-05-15
Introduce RAnalPlugin.jmpmid and replace some is_x86 ##anal

condret:

**Description**


trufae:
Maybe rhis can be taken or induced from archinfo instead of adding more fields

PR #20104, opened 2022-05-16
Remove anal.endsize

condret:
DO NOT MERGE


trufae:
lgtm. what do you think about killing the linearfcnsize info too? the only reason to keep it is imho when using fcn size without bbs added by hand. which is probably not ideal. and we can compute the linear info from the bbinfo too. and its no longer a single linked list so its not slow

trufae:
protip: if you dont want a PR to be merged by accident. prefix the title with `WIP: ` its an unwritten convention. but worth using it

PR #20136, opened 2022-05-20
This condition looks inverted, needs testing

trufae:

**Description**

### DIFF
--- a/libr/main/radare2.c
+++ b/libr/main/radare2.c
@@ -12,7 +12,7 @@
 #include <r_core.h>
 
 static bool is_valid_gdb_file(RIODesc *d) {
-	return d && strncmp (d->name, "gdb://", 6);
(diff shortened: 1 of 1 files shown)


condret:
lgtm

trufae:
Yes the change is correct. but the whole logic in the else block needs to be rewritten to make sense, and this needs testing with real gdb backends to ensure we dont break stuff. will take care of this and merge when done

condret:
renaming is_valid_gdb_file to is_valid_gdb_uri would be a good improvement imo

trufae:
it's way more readable to remove the function itself and just use r_str_startswith() there are more confusing things in this file and i've also tried the gdb backend from master and its also full of bugs, so i dont think it will hurt to merge this meaningful change.

PR #20154, opened 2022-05-22
Fix wasm custom name leak ##bin

swoops:

**Description**

Fixing a leak and improving code some. You might want to wait for a couple tests to be added before merging.

Looks like the `parse_custom_names_local` code is untested, I don't think there is any wasm files in the test files to even hit that code. I will fix add tests soon. If someone has a wasm file that hits that code, let me know. Otherwise I will find or build something.

### DIFF
--- a/libr/bin/format/wasm/wasm.c
+++ b/libr/bin/format/wasm/wasm.c
@@ -271,6 +271,41 @@ static void wasm_sec_free(RBinWasmSection *sec) {
 	}
 }
 
+static void wasm_custom_name_free(RBinWasmCustomNameEntry *cust) {
+	if (cust) {
+		switch (cust->type) {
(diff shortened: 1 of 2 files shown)


swoops:
Got some code to do some tests, it's all coming out wrong... not sure if it's this pull or old bugs but I will get it fixed.

swoops:
I introduced a off-by-one bounds check error before, it was too strict so no overflow or anything. With that fixed this code properly parses the custom section of the attached file.

I will add tests soon for this, but I want to make a better wasm test file, something that has all the sections at least. So you can merge this if you like.

[a.wasm.tar.gz](https://github.com/radareorg/radare2/files/8752767/a.wasm.tar.gz)

PR #20162, opened 2022-05-25
Correction returned code, set zero if ret isn't -1

fr0ster:

**Description**

If return code from r2 isn't -1 then it's non error and set return code in 0


trufae:
Thanks! i'm going to commit a test for this issue in the regression testsuite

PR #20188, opened 2022-05-30
Set asm.flags.inline = true by default ##disasm

trufae:

**Description**


trufae:
Screenshot 2022-05-30 at 11 57 02

PR #20197, opened 2022-05-30
Better handling of Wasm Names ##bin

swoops:

**Description**

A Wasm "Name" is valid if it's utf-8, which can include NULL bytes and all kind of other crazy things. So to prevent r2 from choking on this later in tcc or something, I encode the names on entry. This means we don't have to keep the string length and we can decode them later on output if we want.

Encoder might be a bit overkill atm, I pretty much encode every WASM name into something that can be treated as a C var. I probably don't need to encode section names so strongly, for example I could let a section name start with a number or `.`. So maybe this is just a stop gap.

Feel free to ask for better encoding or accept for now.

### DIFF
--- a/libr/bin/format/wasm/wasm.c
+++ b/libr/bin/format/wasm/wasm.c
@@ -129,7 +129,8 @@ static bool consume_encoded_name_new(RBuffer *b, ut64 bound, ut32 *len_out, char
 	}
 
 	// room for even every character getting encoded
-	char *sout = malloc (len * 4 + 2);
+	ut32 maxsize = len * 4 + 2;
+	char *sout = malloc (maxsize);
 	if (!sout) {
 		free (orig);
 		return false;
@@ -140,7 +141,7 @@ static bool consume_encoded_name_new(RBuffer *b, ut64 bound, ut32 *len_out, char
 		if (WASM_IS_OK (orig, i, len)) {
 			sout[oi++] = orig[i];
 		} else {
-			oi += snprintf (sout + oi, len - oi, "_%02x_", orig[i]);
+			oi += snprintf (sout + oi, maxsize - oi, "_%02x_", orig[i]);
 		}
 	}
 	sout[oi++] = '\0';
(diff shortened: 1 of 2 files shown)


libr/bin/format/wasm/wasm.c
```diff
@@ -105,6 +105,60 @@ static bool consume_str_new(RBuffer *b, ut64 bound, ut32 *len_out, char **str_ou
...
+		free (orig);
+		return false;
+	}
+
+	size_t i, oi = 0;
+	for (i = 0; i < len; i++) {
+		if (WASM_IS_OK (orig, i, len)) {
+			sout[oi++] = orig[i];
+		} else {
+			oi += sprintf (sout + oi, "_%02x_", orig[i]);
```
trufae, line 143: use snprintf passingn len - oi as size, maybe check for overflow may be good in the other case

swoops: Should be fixed, along with a pointer mistake

swoops: Another fix... I should use length of sout rather then length of orig.

libr/bin/format/wasm/wasm.c
```diff
@@ -105,6 +105,61 @@ static bool consume_str_new(RBuffer *b, ut64 bound, ut32 *len_out, char **str_ou
...
+#define WASM_IS_OK(str, i, max) (WASM_IS_OK_EXCEPT__ (i, str[i]) || WASM_UNDERSCORE_OK (str, i, max))
+static bool consume_encoded_name_new(RBuffer *b, ut64 bound, ut32 *len_out, char **str_out) {
+	ut32 len;
+	char *orig = NULL;
+	if (!consume_str_new (b, bound, &len, &orig)) {
+		return false;
+	}
+
+	// room for even every character getting encoded
+	ut32 maxsize = len * 4 + 2;
```
trufae, line 132: size_t
```suggestion
	size_t maxsize = len * 4 + 2;
```

libr/bin/format/wasm/wasm.c
```diff
@@ -105,6 +105,61 @@ static bool consume_str_new(RBuffer *b, ut64 bound, ut32 *len_out, char **str_ou
...
+	char *sout = malloc (maxsize);
+	if (!sout) {
+		free (orig);
+		return false;
+	}
+
+	size_t i, oi = 0;
+	for (i = 0; i < len; i++) {
+		if (WASM_IS_OK (orig, i, len)) {
```
trufae, line 141: Add check for oi +4 < maxsize break

PR #20210, opened 2022-06-02
Bpf

trufae:

**Description**


libr/meson.build
```diff
@@ -250,6 +251,7 @@ bin_plugins += [
   'art',
   'avr',
   'bf',
+  'bpf',
```
trufae, line 254:
```suggestion
```

PR #20212, opened 2022-06-02
Add bpf.cs anal plugin

terorie:
- [x] I've added tests (optional) -- Only minimally

**Description**

Adds Capstone disassembler for BPF


trufae:
Yay! I was about to do this in some hours! That was fast :) some comments before merging:

* asm plugins are in process of deprecation, can you move the code to be just an anal plugin? (Anal.op callback takes a mask that specifies if its suposed to disasm, esil, analyze...)
* Add the plugin to be compiled for meson builds too


Apart from that it lgtm!

libr/asm/p/asm_bpf_cs.c
```diff
@@ -0,0 +1,62 @@
...
+
+RAsmPlugin r_asm_plugin_bpf_cs = {
+	.name = "bpf",
+	.desc = "Capstone " CAPSTONE_VERSION_STRING " BPF disassembler",
+	.license = "BSD",
+	.author = "terorie",
+	.arch = "bpf",
+	.cpus = "bpf",
+	.bits = 64,
+	.endian = R_SYS_ENDIAN_LITTLE | R_SYS_ENDIAN_BIG,
```
terorie, line 51: Should actually pass through endianness to Capstone

trufae: Yes. But i think bpf is always big endian. Isnt?

terorie:
Got it, will fix 🙏🏻

libr/asm/p/asm_bpf_cs.c
```diff
@@ -0,0 +1,62 @@
...
+}
+
+RAsmPlugin r_asm_plugin_bpf_cs = {
+	.name = "bpf",
+	.desc = "Capstone " CAPSTONE_VERSION_STRING " BPF disassembler",
+	.license = "BSD",
+	.author = "terorie",
+	.arch = "bpf",
+	.cpus = "bpf",
+	.bits = 64,
```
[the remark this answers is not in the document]
trufae: Thats up to capstone. But sure if it does 🤔

libr/asm/p/asm_bpf_cs.c
terorie: Capstone and LLVM support both. I'm not sure what the Linux kernel does, but Solana uses little-endian.

libr/asm/p/asm_bpf_cs.c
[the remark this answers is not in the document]
terorie: It's not really just register size – Classic BPF (32-bit GPRs) and extended BPF (64-bit GPRs) are very different architectures. https://www.kernel.org/doc/html/latest/bpf/classic_vs_extended.html
So I'm not sure if bit size is an appropriate way to switch here.  🤔

libr/anal/p/anal_bpf_cs.c
```diff
@@ -0,0 +1,84 @@
...
+			cs_close (&handle);
+			handle = 0; // unnecessary
+		}
+		omode = mode;
+		obits = a->config->bits;
+	}
+	op->size = 8;
+	op->addr = addr;
+	if (handle == 0) {
+		ret = cs_open(CS_ARCH_BPF, mode, &handle);
```
trufae, line 30:
```suggestion
		ret = cs_open (CS_ARCH_BPF, mode, &handle);
```

libr/anal/p/anal_bpf_cs.c
```diff
@@ -0,0 +1,84 @@
...
+		}
+		op->size = insn->size;
+		op->id = insn->id;
+		cs_free (insn, n);
+	}
+	return op->size;
+}
+
+RAnalPlugin r_anal_plugin_bpf_cs = {
+	.name = "bpf",
```
trufae, line 58:
```suggestion
	.name = "bpf.cs",
```

trufae: Not sure which should be the 'default' bpf disassembler. as i plan to maintain both bpf plugins, the other plugin supports assembling and esil. so let's make the cs plugin non-default for now

trufae (changes requested):

libr/anal/p/anal_bpf_cs.c
```diff
@@ -0,0 +1,84 @@
...
+#if CS_API_MAJOR >= 5
+
+static R_TH_LOCAL csh cd = 0;
+
+static int analop(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len, RAnalOpMask mask) {
+	static R_TH_LOCAL csh handle = 0;
+	static R_TH_LOCAL int omode = -1;
+	static R_TH_LOCAL int obits = 32;
+	cs_insn *insn = NULL;
+	int mode = (a->config->bits==32)? CS_MODE_BPF_CLASSIC: CS_MODE_BPF_EXTENDED;
```
trufae, line 16:
```suggestion
	int mode = (a->config->bits == 32)? CS_MODE_BPF_CLASSIC: CS_MODE_BPF_EXTENDED;
```

trufae (changes requested):

PR #20213, opened 2022-06-02
Fix #20206 - pthread_create: Resource temporarily unavailable

### LINKED ISSUE #20206: pthread_create: Resource temporarily unavailable
## Environment

```sh
date
mi\u00e9 01 jun 2022 16:56:13 WEST

r2 -v
radare2 5.7.0 28112 @ linux-x86-64 git.5.7.0-5-gdd3bc7b52
commit: dd3bc7b52008ca6130f3993a8ecdfe310f50d88e build: 2022-06-01__15:55:09

uname -ms
Linux x86_64

lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04 LTS
Release:	22.04
Codename:	jammy
```

## Description
Running a long session with esil and capturing data through a pipe script causes an error in the pthread library.

## Test
```
"e cmd.esil.stepout=#!pipe node ./cmd.esil.stepout.js"

const fs = require('fs');
fs.appendFileSync("stepout.txt", `${process.argv.join(", ")}\n`);
```
```
$ ./r2.sh 
 -- Put some sugar before the release, attracts the bugs and makes it easier to squash them
[0x00000000]> aecu 0x00002848
node[117931]: pthread_create: Resource temporarily unavailable
^C[+] ESIL emulation interrupted at 0x00003378
[0x0000335c]> q
```
While the script is running a bunch of <defunct> processes are created and they usually disappear when I close r2, but not always, in some cases having to reboot the machine.
```
$ ps ax | grep "<defunct>" | grep -v grep | wc -l
4610
```
I have created a small patch that I use to complete the execution of the esil sessions and so far, after a few tests, it works and does not cause that error to appear.

```
@@ -234,7 +235,7 @@ static bool lang_pipe_run(RLang *lang, const char *code, int len) {
 		close (safe_in);
 	}
 #ifndef __wasi__
-	waitpid (child, NULL, WNOHANG);
+	waitpid (child, NULL, WUNTRACED);  
 #endif
 	return true;
 #else
diff --git a/libr/util/sandbox.c b/libr/util/sandbox.c
```

pipothebit:
Fix error that appears when running a long session with esil.
https://github.com/radareorg/radare2/issues/20206

**Description**

### DIFF
--- a/libr/lang/p/pipe.c
+++ b/libr/lang/p/pipe.c
@@ -234,7 +234,7 @@ static bool lang_pipe_run(RLang *lang, const char *code, int len) {
 		close (safe_in);
 	}
 #ifndef __wasi__
-	waitpid (child, NULL, WNOHANG);
+	waitpid (child, NULL, WUNTRACED);
 #endif
 	return true;
 #else


libr/lang/p/pipe.c
```diff
@@ -234,7 +234,7 @@ static bool lang_pipe_run(RLang *lang, const char *code, int len) {
 		close (safe_in);
 	}
 #ifndef __wasi__
-	waitpid (child, NULL, WNOHANG);
+	waitpid (child, NULL, WUNTRACED);
```
trufae, line 237:
```suggestion
	waitpid (child, NULL, WNOHANG | WUNTRACED);
```

trufae (changes requested):
Can you confirm the suggested patch works as expected for you too?

trufae:
related doc https://stackoverflow.com/questions/33508997/waitpid-wnohang-wuntraced-how-do-i-use-these

pipothebit:
I've just tested your change and it doesn't work, zombie processes still remain there on every pipe call until r2 closes.

According to the documentation the options parameter of the ```waitpid``` function can have a combination of 3 possible values: ```WCONTINUED```, ```WNOHANG``` and ```WUNTRACED```. I've been doing some tests and when ```WNOHANG``` is used as an option, if ```waitpid``` returns 0, that process goes zombie. If using ```WNOHANG``` ```waitpid``` returns 0, the state of the child process has not changed. I don't know what to do in this case, I'm just a rookie in c programming,

PR #20214, opened 2022-06-02
Add initial BPF ELF relocs

terorie:
- [x] ~~I wrote some lines in the [book](https://github.com/radareorg/radare2book) (optional)~~

**Description**

- Adds BPF relocation types
- Adds an attempt to implement said relocations

```
Enum  ELF Reloc Type     Description      BitSize  Offset        Calculation
0     R_BPF_NONE         None
1     R_BPF_64_64        ld_imm64 insn    32       r_offset + 4  S + A
2     R_BPF_64_ABS64     normal data      64       r_offset      S + A
3     R_BPF_64_ABS32     normal data      32       r_offset      S + A
4     R_BPF_64_NODYLD32  .BTF[.ext] data  32       r_offset      S + A
10    R_BPF_64_32        call insn        32       r_offset + 4  (S + A) / 8 - 1
```

Reference: https://www.kernel.org/doc/html/latest/bpf/llvm_reloc.html

### DIFF
--- a/libr/bin/format/elf/glibc_elf.h
+++ b/libr/bin/format/elf/glibc_elf.h
@@ -3964,6 +3964,10 @@ enum
 
 #define R_BPF_NONE		0	/* No reloc */
 #define R_BPF_64_64		1
+#define R_BPF_64_ABS64		2
+#define R_BPF_64_ABS32		3
+#define R_BPF_64_NODYLD32	4
+#define R_BPF_64_RELATIVE	8
 #define R_BPF_64_32		10
 
 /* Imagination Meta specific relocations. */
--- a/libr/bin/p/bin_elf.inc
+++ b/libr/bin/p/bin_elf.inc
@@ -711,8 +711,19 @@ static RBinReloc *reloc_convert(struct Elf_(r_bin_elf_obj_t) *bin, RBinElfReloc 
 		case R_PPC_GLOB_DAT:    ADD (32, 0);
 		case R_PPC_JMP_SLOT:    ADD (32, 0);
 		default:
-			eprintf ("unimplemented ELF/PPC reloc type %d\n", rel->type);
+			R_LOG_DEBUG ("unimplemented ELF/PPC reloc type %d", rel->type);
 		break;
+		}
+		break;
+	case EM_BPF: switch (rel->type) {
+		case R_BPF_NONE: break;
+		case R_BPF_64_64: r->vaddr += 4; ADD (32, 0); break;
+		case R_BPF_64_ABS64: ADD (64, 0); break;
+		case R_BPF_64_ABS32: ADD (32, 0); break;
+		case R_BPF_64_NODYLD32: ADD (32, 0); break;
+		default:
+			R_LOG_DEBUG ("unimplemented ELF/BPF reloc type %d", rel->type);
+			break;
 		}
 		break;
 	default:
--- a/libr/bin/p/bin_elf.inc
+++ b/libr/bin/p/bin_elf.inc
@@ -612,114 +612,114 @@ static RBinReloc *reloc_convert(struct Elf_(r_bin_elf_obj_t) *bin, RBinElfReloc 
 	switch (bin->ehdr.e_machine) {
 	case EM_386: switch (rel->type) {
 		case R_386_NONE:     break; // malloc then free. meh. then again, there's no real world use for _NONE.
-		case R_386_32:       ADD(32, 0);
-		case R_386_PC32:     ADD(32,-(st64)P);
-		case R_386_GLOB_DAT: SET(32);
-		case R_386_JMP_SLOT: SET(32);
-		case R_386_RELATIVE: ADD(32, B);
-		case R_386_GOTOFF:   ADD(32, -(st64)got_addr);
-		case R_386_GOTPC:    ADD(32, got_addr - P);
-		case R_386_16:       ADD(16, 0);
-		case R_386_PC16:     ADD(16,-(st64)P);
-		case R_386_8:        ADD(8,  0);
-		case R_386_PC8:      ADD(8, -(st64)P);
-		case R_386_COPY:     ADD(32, 0); // XXX: copy symbol at runtime
+		case R_386_32:       ADD(32, 0); break;
+		case R_386_PC32:     ADD(32,-(st64)P); break;
+		case R_386_GLOB_DAT: SET(32); break;
+		case R_386_JMP_SLOT: SET(32); break;
+		case R_386_RELATIVE: ADD(32, B); break;
+		case R_386_GOTOFF:   ADD(32, -(st64)got_addr); break;
+		case R_386_GOTPC:    ADD(32, got_addr - P); break;
+		case R_386_16:       ADD(16, 0); break;
+		case R_386_PC16:     ADD(16,-(st64)P); break;
+		case R_386_8:        ADD(8,  0); break;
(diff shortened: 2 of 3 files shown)


libr/bin/format/elf/glibc_elf.h
```diff
@@ -3964,6 +3964,10 @@ enum
 
 #define R_BPF_NONE		0	/* No reloc */
 #define R_BPF_64_64		1
+#define R_BPF_64_ABS64		2
+#define R_BPF_64_ABS32		3
+#define R_BPF_64_NODYLD32	4
+#define R_BPF_64_RELATIVE	8
```
terorie, line 3970: `R_BPF_64_RELATIVE` is a really odd one.

It applies on the ldxdw instruction (read "load 64-bit value from absolute 64-bit address").
This instruction is encoded using 128 bits (four 32-bit words). The first word stores the virtual address (ELF file offset) to resolve. After resolving it to a physical address, the high word of the physical address gets stored in the first word and the low word in the third word.

I don't think `RBinReloc` can express this behavior.

libr/bin/p/bin_elf.inc
```diff
@@ -715,6 +715,17 @@ static RBinReloc *reloc_convert(struct Elf_(r_bin_elf_obj_t) *bin, RBinElfReloc
...
 		}
 		break;
+	case EM_BPF: switch (rel->type) {
+		case R_BPF_NONE: break;
+		case R_BPF_64_64: r->vaddr += 4; ADD (32, 0);
+		case R_BPF_64_ABS64: ADD (64, 0);
+		case R_BPF_64_ABS32: ADD (32, 0);
+		case R_BPF_64_NODYLD32: ADD (32, 0);
+		default:
+			eprintf ("unimplemented ELF/BPF reloc type %d\n", rel->type);
```
trufae, line 725:
```suggestion
			R_LOG_DEBUG ("Unimplemented elf.bpf reloc type %d", rel->type);
```

libr/bin/p/bin_elf.inc
```diff
@@ -715,6 +715,17 @@ static RBinReloc *reloc_convert(struct Elf_(r_bin_elf_obj_t) *bin, RBinElfReloc
 		break;
 		}
 		break;
+	case EM_BPF: switch (rel->type) {
+		case R_BPF_NONE: break;
+		case R_BPF_64_64: r->vaddr += 4; ADD (32, 0);
```
trufae, line 720:
```suggestion
		case R_BPF_64_64: r->vaddr += 4; ADD (32, 0); break;
```

libr/bin/p/bin_elf.inc
```diff
@@ -715,6 +715,17 @@ static RBinReloc *reloc_convert(struct Elf_(r_bin_elf_obj_t) *bin, RBinElfReloc
...
+		case R_BPF_64_ABS64: ADD (64, 0);
```
trufae, line 721:
```suggestion
		case R_BPF_64_ABS64: ADD (64, 0); break;
```

libr/bin/p/bin_elf.inc
```diff
@@ -715,6 +715,17 @@ static RBinReloc *reloc_convert(struct Elf_(r_bin_elf_obj_t) *bin, RBinElfReloc
...
+		case R_BPF_64_ABS32: ADD (32, 0);
+		case R_BPF_64_NODYLD32: ADD (32, 0);
```
trufae, line 723:
```suggestion
		case R_BPF_64_NODYLD32: ADD (32, 0); break
```

trufae (changes requested):

libr/bin/p/bin_elf.inc
terorie, line 720: I have no idea whether I've done this correctly.
According to Linux docs:

> For example, R_BPF_64_64 relocation type is used for ld_imm64 instruction. The actual to-be-relocated data (0 or section offset) is stored at r_offset + 4 and the read/write data bitsize is 32 (4 bytes). The relocation can be resolved with the symbol value plus implicit addend. Note that the BitSize is 32 which means the section offset must be less than or equal to UINT32_MAX and this is enforced by LLVM BPF backend.

Again, the `ld_imm64` instruction uses split encoding (4x32-bit words, high 32-bits in word 1, low 32-bits in word 3). So the value to be written gets forcibly truncated to 32 bits. I think the Linux kernel does this because it has a 32-bit program counter, but this doesn't make sense for a 64-bit PC flavor.

terorie: @trufae The `ADD` macro is defined as follows:

```
#define ADD(T, A) r->type = R_BIN_RELOC_ ## T; r->addend += A; r->additive = rel->mode == DT_RELA; return r
```

And all the other relocs don't use a break either – should I change them too?

libr/bin/format/elf/glibc_elf.h
terorie: Pre-relocation
```
[ 127..96 ] [ 95..64 ] [ 63..32 ] [ 31..0 ]
              ^ virtual address bits 31..0
```
Post-relocation
```
[ 127..96 ] [ 95..64 ] [ 63..32 ] [ 31..0 ]
              ^ physical address bits 63..32 (i think??)
                                    ^ physical address bits 31..0
```

libr/bin/p/bin_elf.inc
trufae: Correct. there's no need to place a break after a return, but it seems more consistent with a break to me. otherwise it's a little confusing for the reader. (just trying to improve overall code quality, not related to your patch in here), but always feel free to make things more readable if you find them :)

trufae: I would say that if the resulting pointers look good it's fine to merge and start adding tests with more binaries.

terorie: Thanks for clarifying. Will fix 🙏🏻

trufae:
ok to merge?

trufae:
Please resolve the cornflakes^Wconflicts

trufae:
ping?

terorie:
@trufae Yes, sorry for the delay, looking now, and thanks for resolving the conflicts. When trying to load the Solana example ELF, I noticed the relocs never get applied at all though. So I'm not sure this PR is of any use because I haven't been able to test them.

From what I can tell, this is because `init_dynamic_section` bails here at `r_bin_elf_v2p_new`:

https://github.com/radareorg/radare2/blob/bd4cf6bb2e9a18e57a15a151a76853eb49776ee7/libr/bin/format/elf/elf.c#L691-L700

I think `r_bin_elf_v2p_new` expects that there is a corresponding virtual address covered loaded by `PT_LOAD` for the dynamic segment.

But the Solana ELF file `DYNAMIC` and `LOAD` segments don't overlap like this.

```
Program Header:
    LOAD off    0x00000000000000e8 vaddr 0x00000000000000e8 paddr 0x00000000000000e8 align 2**12
         filesz 0x000000000000e208 memsz 0x000000000000e208 flags r-x
    LOAD off    0x000000000000e300 vaddr 0x000000000000e300 paddr 0x000000000000e300 align 2**12
         filesz 0x0000000000001884 memsz 0x0000000000001884 flags r--
 DYNAMIC off    0x000000000000fb88 vaddr 0x000000000000fb88 paddr 0x000000000000fb88 align 2**3
         filesz 0x0000000000002540 memsz 0x0000000000002540 flags rw-
```

PR #20234, opened 2022-06-06
Fix CS BPF mistakes

aemmitt-ns:

**Description**

I made a number of mistakes in the previous PR that are now fixed, namely not using a->config->bits to determine the mode in jump instructions, calling immediates, neg32/64, and `lddw` which I still don't fully understand. LLVM seems to put the immediate in two different places. But the behaviour now matches both cases so it should be ok.

### DIFF
--- a/libr/anal/p/anal_bpf_cs.c
+++ b/libr/anal/p/anal_bpf_cs.c
@@ -21,6 +21,7 @@ static int analop(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len, RAn
 	static R_TH_LOCAL csh handle = 0;
 	static R_TH_LOCAL int omode = -1;
 	static R_TH_LOCAL int obits = 32;
+
 	cs_insn *insn = NULL;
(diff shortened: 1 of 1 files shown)


aemmitt-ns:
ok so for `lddw` they just put the least significant 4 bytes first, then 4 nulls, then the most significant 4 bytes 

`22:	18 05 00 00 0d d0 0d f0 00 00 00 00 ad de ad de	r5 = -2401018187680460787 ll // 0xdeaddeadf00dd00d`

its like they couldnt decide between LE or BE so they just decided to do something that made absolutely no sense instead

terorie:
Upstream fix: https://github.com/capstone-engine/capstone/pull/1885

terorie:
> its like they couldnt decide between LE or BE so they just decided to do something that made absolutely no sense instead

The immediate is always in the last word of an instruction. The instruction field layout is the same across BE and LE but the field encoding isn't.
Since the immediate is 32-bit, they just used two instruction slots, and the opcode of the second "fake" instruction is ignored :^)

PR #20237, opened 2022-06-06
Add Z registers

aemmitt-ns:

**Description**

Adds the z0-31 registers to the register profile. They are newer scalable registers but share the bottom 128 bits with the v0-31 regs so for right now I just identify the two.


trufae:
can you update this test?

Screenshot 2022-06-08 at 19 21 25

trufae:
Fixed and merged. thanks!

PR #20244, opened 2022-06-07
Define R_DIRTY_VAR if needed

wtdcode:

**Description**

When `r_util.h` is included alone, this macro is not defined, which make `r2libr` fails to generate bindings.


wtdcode:
> i think its better to define R_DIRTY in r_types_base, so it's defined only in one place. can you update the PR with this change and confirm it works for you?
> 
> thanks!

Done and test work with `r2libr`.

trufae:
Sorry for my explanation, what I mean, is that R_DIRTY is already defined in r_util.h, so you should remove those 3 lines from r_util.h and write them in r_types.h

```
#define R_DIRTY(x) (x)->is_dirty = true
#define R_IS_DIRTY(x) (x)->is_dirty
#define R_DIRTY_VAR bool is_dirty
```

wtdcode:
> Sorry for my explanation, what I mean, is that R_DIRTY is already defined in r_util.h, so you should remove those 3 lines from r_util.h and write them in r_types.h
> 
> ```
> #define R_DIRTY(x) (x)->is_dirty = true
> #define R_IS_DIRTY(x) (x)->is_dirty
> #define R_DIRTY_VAR bool is_dirty
> ```

Done. But I'm not very sure if it would introduce any side effects.

PR #20270, opened 2022-06-10
Initial import of the assembler fuzzer ##tests

trufae:

**Description**


dist/docker/fuzz-asm/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ubuntu
```
terorie, line 1: Maybe pin OS version? e.g. jammy

trufae: im always confused about ubuntu/debian release names, i prefer version numbers, but yeah both dockerfiles can be improved, but i dont think the version of the system really affects much in the fuzzing process. the latest the better i guess because that will end up building with a more modern asan and compiler to spot more issues, but not a blocker

PR #20276, opened 2022-06-10
Use hex for d specifier in pf

Lazula:
Might need improvement but it no longer contradicts the pf?? message, and uint32 no longer prints decimal while uint64 prints hex.


trufae:
uhm, 'd' format should be for decimal not hex.. there's 'x' for hexadecimal formating already 🤔

trufae:
i know d stands for dword in here.. is `i` for 32bit decimal then?

Lazula:
relevant lines:

```
|  d       0xHEX value (4 bytes) (see 'i' and 'x')
|  i       signed integer value (4 bytes) (see 'd' and 'x')
|  x       0xHEX value and flag (fd @ addr) (see 'd' and 'i')
```

so, d = dword (hex), i = int (decimal), x = int (hex) and flag, idk if there is any real difference here though

d was already using a function called `r_print_format_hex()`, but that was using PFMT64d

the problem this is trying to fix is that current behavior makes uint32_t print decimal and uint64_t print hex

trufae:
Lgtm!

PR #20294, opened 2022-06-15
Fix bug in io_ihex ##io

condret:

**Description**


trufae:
Add a test before merging. The ihex parser is a little messy and is better to have them for the next time we touch this code

PR #20295, opened 2022-06-15
Last of RBinWasmObj entries to RPVectors ##bin

swoops:

**Description**

Mostly refactoring things to RPVectors. This does clean up a bunch of code and fix a few bugs that could cause if double free's.

This does change how ordinals are assigned to some symbols. Wasm uses separate "index spaces" to identify functions, tables, memories and globals.  So there is a `global[0]` and a `function[0]` rather then `symbols[0]` being either a function or global. This code will better this behavior in ordinals. So there may be a global symbol with the same ordinal as a function.


swoops:
`bins/fuzzed/r2_hoobr_wasm_r_read_le32` was causing a warning. The latest commit fixes it.

I did not realize `r_pvector_at` warned when asked for an out of bound index. I could see that being a programing error in a lot of cases, but because we want to parse broken files as best as we can, some sections may not be as full as  they should be. So I shimmed the call to keep it from warning and added better errors to indicate why a file is corrupt.

PR #20312, opened 2022-06-16
Merge asm.v850 into anal.v850

terorie:

**Description**


trufae:
thanks for the tests!

trufae:
Eventually i would like to merge v850 and v850.np (my reimplementation) which is why more complete, but its targetting a different model of the chip. so i will route this code when using v850e0 cpu settings. but that's just another story

trufae:
when the meson builds fail on windows is because of a race condition bug in meson i reported a while ago, but its hard to repro and its not fixed yet. so you can just re-run the job or assume its just a bug generating .sdb files in parallel. i may find some time to look into that at some point

terorie:
Thanks, that makes sense. I don't think I have re-run job perms though.

PR #20337, opened 2022-06-20
Fix oobread in RTable exposed via an ELF reproducer ##crash

trufae:

**Description**


libr/util/table.c
```diff
@@ -50,6 +50,7 @@ static void __table_adjust(RTable *t) {
 		int ncol = 0;
 		r_list_foreach (row->items, iter2, item) {
 			int itemLength = r_str_len_utf8_ansi (item) + 1;
+eprintf ("LEN (%s) %d\n", item, itemLength);
```
trufae, line 53:
```suggestion
```

PR #20361, opened 2022-06-26
Add ESIL support for ARM-16 ROR instruction

Baldanos:

**Description**

ESIL support for ARM ´ror´ instruction was missing. This PR adds support and a test.


trufae:
thanks! even with tests :) good contrib!

PR #20374, opened 2022-06-28
Fix wasm function offset lookup ##bin

swoops:

**Description**

This code addresses two things.

First the import sectoin is now split into 4 `RPVector`'s instead of 1, there is a separate vector for functions, memories, globals and tables. The import section marks the start of the index space and since there is a different index space for each type, having them separated will simplify parsing logic. This also produces a very minor speed improvement because import parsing loops don't need switches for types.

With the separate import types I then fixed an issue where the WASM `.get_offset` API was returning the wrong function. Now that API returns the offset to the correct function. This results in `aa` finding functions in a different order, so a test had to be fixed.

I did not add a test for `get_offset` yet because it is still not quite right. I have not figured out why but the offset it returns is off by a few bytes:
```
$ r2 ~/radare2/test/bins/wasm/sections.wasm 
...
[0x00000207]> s 0x2e8
[0x000002e8]> pd 2
            0x000002e8      1007           call sym.static_int_sum     ; 0x3b9 ; "#"
            0x000002ea      2116           set_local 22

 ```
```
$ wasm-objdump -d ~/radare2/test/bins/wasm/sections.wasm|grep static_int_sum
 0002e8: 10 07                      |   call 7 <static_int_sum>
0003ac func[7] <static_int_sum>:
```

Wasm-objdump puts `static_int_sum` at 0x0003ac while r2 puts it at 0x3b9. So r2 is not quite right but it's a lot less wrong. Before this patch it would point to the wrong function:

```
[0x000002e8]> pd 3
            0x000002e8      1007           call sym.static_int_sum     ; sym.__original_main
                                                                       ; 0x4ab ; sym.__original_main
            0x000002ea      2116           set_local 22
            0x000002ec      2004           get_local 4
```

I will look into correcting the address problem next.

### DIFF
--- a/libr/bin/format/wasm/wasm.c
+++ b/libr/bin/format/wasm/wasm.c
@@ -355,6 +355,14 @@ static void free_import_entry(RBinWasmImportEntry *entry) {
 	}
 }
 
+static inline void free_all_imports(RBinWasmObj *bin) {
+	int i;
+	for (i = 0; i < R_ARRAY_SIZE (bin->g_imports_arr); i++) {
+		r_pvector_free (bin->g_imports_arr[i]);
+	}
+	memset (bin->g_imports_arr, 0, sizeof (bin->g_imports_arr));
+}
+
 static void free_export_entry(RBinWasmExportEntry *entry) {
 	if (entry) {
 		free (entry->field_str);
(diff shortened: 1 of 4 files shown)


swoops:
I am mistaken on r2 having the wrong address. I did not properly understand the output of wasm-objdump. So I will add a test in a bit.

PR #20375, opened 2022-06-28
[5.8.0] Fix free-before-use on r_io_reopen of a rbuf:// fd/desc ##io

condret:

**Description**


libr/io/p/io_rbuf.c
```diff
@@ -18,8 +18,6 @@ static int __read(RIO *io, RIODesc *fd, ut8 *buf, int count) {
 }
 
 static bool __close(RIODesc *fd) {
-	RBuffer *b = fd->data;
-	r_buf_free (b);
```
trufae, line 22: what about r_unref. and call r_ref when taking it in oen? instead

trufae:
We can merge this now

PR #20427, opened 2022-07-08
Fix ROL RCL ROR RCR with memory locations ##esil

Lahvuun:

**Description**
These instructions work fine with registers:
```assembly
mov eax,0x10
ror eax, 0x18
```
That is translated to esil as:
```
1785643824,eax,=
24,eax,>>>,eax,=
```
But when working with memory locations:
```assembly
mov dword [esp + 0x14], 0x6a6ec330
ror dword [esp + 0x14], 0x18
```
That gets translated to esil as:
```
1785643824,0x14,esp,+,=[4]
24,0x14,esp,+,[4],>>>,0x14,esp,+,[4],=
```
Which doesn't work because `0x14,esp,+,[4]` is evaluated to
`[esp+0x14]`, not `esp+0x14`. `0x14,esp,+,[4],=` then tries to assign to
`[esp+0x14]`, instead of correctly assigning to `esp+0x14`.

This change fixes the problem by translating:
```assembly
ror dword [esp + 0x14], 0x18
```
To:
```
24,0x14,esp,+,[4],>>>,0x14,esp,+,=[4]
```
So the instructions work with memory locations, as well.

---
This still doesn't emulate the instructions correctly, however, since they also set `CF` and `OF` (https://c9x.me/x86/html/file_module_x86_id_273.html). The ESIL expression for doing that gets quite complex, so I gave up on writing it for now. But this change is enough to fix emulation for my use case, so I figured I should still submit it.


trufae:
can you add tests for both destinations (reg and memory?)

Lahvuun:
Sure! I'll ping you once I push them.

trufae:
I just tested it and it works as expected. i'm fine to merge without tests, but feel free to submit another PR with them! thank you

Lahvuun:
OK! I'll submit them once I get the flags to work properly.

PR #20432, opened 2022-07-09
Refactor projects ##refactor ##projects	

RHL120:

**Description**


trufae:
I dont think wrapping up lines to 80 columns improves readability at all, for some cases it reads better with one statement per line. Other cases look like it can be refactored as an external inline function or a separate variable.

maybe we should update the guidelines to suggest 120-130 columns for lines that cant be wrapped without improving readability. This rule is actually followed by all the function definitions to make grep happy. And personally i think we can improve the code without paying off with more locs

trufae:
The developers recommendation for maxlinewodth has been updated. Can u update the pr accordingly?

libr/core/project.c
```diff
@@ -661,8 +648,7 @@ R_API bool r_core_project_save(RCore *core, const char *prj_name) {
 	if (!prj_dir) {
 		prj_dir = strdup (prj_name);
 	}
-	if (r_core_is_project (core, prj_name) &&
-			strcmp (prj_name, r_config_get (core->config, "prj.name"))) {
+	if (r_core_is_project (core, prj_name) && strcmp (prj_name, r_config_get (core->config, "prj.name"))) {
 		R_LOG_ERROR ("A project with this name already exists. Use P-%s to delete it",
```
trufae, line 652: Why 2 lines

libr/core/project.c
```diff
@@ -608,8 +596,7 @@ R_API bool r_core_project_save_script(RCore *core, const char *file, int opts) {
 	}
 	r_core_cmd (core, "wc*", 0);
 	if (opts & R_CORE_PRJ_ANAL_SEEK) {
-		r_cons_printf ("# seek\n"
-				"s 0x%08" PFMT64x "\n",
+		r_cons_printf ("# seek\n" "s 0x%08" PFMT64x "\n",
```
trufae, line 599: Why 2 lines

libr/core/project.c
```diff
@@ -478,17 +472,12 @@ static bool store_files_and_maps(RCore *core, RIODesc *desc, ut32 id) {
...
 		if ((maps = r_io_map_get_by_fd (core->io, id))) { //wtf
 			r_list_foreach (maps, iter, map) {
-				r_cons_printf ("om %d 0x%" PFMT64x " 0x%"
-						PFMT64x " 0x%" PFMT64x
-						" %s%s%s\n", fdc,
-					r_io_map_begin (map), r_io_map_size (map),
-					map->delta, r_str_rwx_i (map->perm),
-					map->name ? " " : "",
-					r_str_get (map->name));
+				r_cons_printf ("om %d 0x%" PFMT64x " 0x%" PFMT64x " 0x%" PFMT64x " %s%s%s\n", fdc, r_io_map_begin (map),
```
trufae, line 478: Even if that fits in 2 lines i think its less readable . Thats why automatic indentation will never be better than a human because it cant appreciate or determine when a construction is easier for the eye or not. Same as following rules blindly

libr/core/project.c
```diff
@@ -393,8 +388,7 @@ R_API bool r_core_project_open(RCore *core, const char *prj_path) {
 	}
 	if (ask_for_closing && r_project_is_loaded (core->prj)) {
 		if (r_cons_is_interactive ()) {
-			close_current_session = interactive? r_cons_yesno ('y',
-					"Close current session? (Y/n)"): true;
+			close_current_session = interactive? r_cons_yesno ('y', "Close current session? (Y/n)"): true;
```
trufae, line 391: Same here. Its easier to read in two lines even if it fits in one line

trufae (changes requested):

libr/core/project.c
```diff
@@ -65,7 +65,7 @@ static int make_projects_directory(RCore *core) {
 
 R_API bool r_core_is_project(RCore *core, const char *name) {
 	bool ret = false;
-	if (name && *name && *name != '.') {
+	if (!R_STR_ISEMPTY (name) && *name != '.') {
```
trufae, line 68:
```suggestion
	if (R_STR_ISNOTEMPTY (name) && *name != '.') {
```

trufae (changes requested):

PR #20433, opened 2022-07-10
[WIP] Merge asm.dalvik into anal

gogo2464:

**Description**


Lazula:
This is far from ready, so I'm marking it as a draft. Please mark it for review once it's in working order.

trufae:
I would suggest @gogo2464 to work on a simplest plugin instead. It will be essier to review too. Dalvik one is probably the most complex of the remaining ones

gogo2464:
Alright. Should I just push a last time to let someone continue on my work?

trufae:
im curious about how the new output is getting printed , because i think its more readable, but can cause problems to understand which exact overloaded function is being called. so not sure if there's an `asm.` variable to select that or how you got the different disassembly on the failing tests

Screenshot 2022-07-14 at 17 32 57

gogo2464:
Basically I just changed RAsm struct by RAnal. But the names of the attributes are different. Then I probably made a mistake.

Lazula:
its using `RAnal.binb.get_name()`. previously it used `RAsm.binb.get_name()` and passed `RAsm.pseudo` as `sd` (`simplified`), but idk which if any anal member would be used

Lazula:
Right now simplified is just always enabled instead of being configured

gogo2464:
Ah yes and I put the fourth argument of RAnal.binb.get_name() as alway true! It could be the mistake.

PR #20442, opened 2022-07-13
Remove global use for esil

iotaaxel:

**Description**

- The global variable was marked and moved. Will now use local static variable instead.


Lazula:
This does not make the variable not global, it makes it thread unsafe. "static" does not mean the same thing between variables and functions.

iotaaxel:
Refer to https://github.com/radareorg/radare2/pull/20469 for changes

PR #20454, opened 2022-07-16
Allow seek to previous result item when it is at 0 offset

richardpl:

**Description**

Fixes unable to previous seek to 0 (sp command or N key in Visual mode) offset when search hit is at 0 offset also. This one I find very irritating.


trufae:
I think that just checking for ut64max instead of 0 should be enough

libr/core/vmenus.c
```diff
@@ -3745,17 +3745,20 @@ struct seek_flag_offset_t {
...
 static bool seek_flag_offset(RFlagItem *fi, void *user) {
 	struct seek_flag_offset_t *u = (struct seek_flag_offset_t *)user;
 	if (u->is_next) {
 		if (fi->offset < *u->next && fi->offset > u->offset) {
 			*u->next = fi->offset;
+			u->found = true;
 		}
 	} else {
-		if (fi->offset > *u->next && fi->offset < u->offset) {
+		if (fi->offset >= *u->next && fi->offset <  u->offset) {
```
trufae, line 3759: if its the same (>=) wont seek anywhere, isnt?

richardpl: This is to be able to seek to 0x0 offset, currently you can not seek with N key to result at 0x0 offset. I made sure that I have not introduced new regressions by fixing this small bug, so seeking with N/n should be same as before. And regarding checking for ut64max in seeking to backward results, that can only work if it is also offset is initialized to that value. That could be alternative solution instead of adding new variable 'found', will test this alternative and if correct update this PR.

richardpl: I tried using u64max but it does not work. >= is there so previous seeking still works if search result is at 0x0. New 'found' variable is introduced instead of using initialization values to properly check that search was successful.

trufae: yes i think this code will work the same without the need of the found field. can you please confirm? thanks for the contribs!

richardpl: That was my original thought, but then I found that it seeks to 0x0 even it there is not hit there, so to my understanding 'found' variable introduction is mandatory for finding previous hits (technically only hit at 0x0). For next hits it is not needed but I changed it anyway to not add more inconsistencies or confusions. The hit at 0x0 is special as the code can not search into negative offsets (it wraps around instead), thus both 'found' and >= changes are required to handle this special 0x0 case IMO. So my conclusion is that there is no alternative solution to this bug.

trufae:
Lets merge it then! No need for more rounds of reviews. Lgtm

PR #20455, opened 2022-07-16
Merge asm.dalvik into anal.dalvik

Lazula:
Includes some minor refactors to dalvik disasm and cleans up the r_anal.h/r_asm.h macros only used in dalvik, with private functions instead. The NULL-checked get_offset was missing in a few places in dalvik_op(), which caused it to call anal->binb.get_offset() when it was NULL, causing a crash, which has been fixed.


libr/anal/p/anal_dalvik.c
```diff
@@ -1157,6 +1720,7 @@ RAnalPlugin r_anal_plugin_dalvik = {
 	.bits = 32,
 	.desc = "Dalvik (Android VM) bytecode analysis plugin",
 	.op = &dalvik_op,
+	.opasm = &dalvik_opasm,
```
Lazula, line 1723: Might want to comment out the opasm for the time being since its basically unimplemented

libr/include/r_anal.h
```diff
@@ -611,6 +598,7 @@ typedef struct r_anal_hint_cb_t {
 typedef struct r_anal_t {
 	RArchConfig *config;
 	int lineswidth; // asm.lines.width
+	bool pseudo;    // asm.pseudo
```
Lazula, line 601: Couldnt find a way to pass RAsm.pseudo into an RAnalPlugin. Will remove this if it's possible otherwise.

libr/core/cconfig.c
```diff
@@ -585,6 +585,7 @@ static bool cb_asmpseudo(void *user, void *data) {
 	RCore *core = (RCore *) user;
 	RConfigNode *node = (RConfigNode *) data;
 	core->rasm->pseudo = node->i_value;
+	core->anal->pseudo = node->i_value;
```
Lazula, line 588: See the r_anal.h comment

libr/include/r_anal.h
trufae: this change will break the ABI, so it's not suposed to be merged on a patch release.

imho asm->pseudo makes "sense", because rparse will be merged into rasm, when all the asm plugins get moved to anal. 

So the non-breaking abi solution to this is to use RAnal.coreb.cfgGetI("asm.pseudo")

trufae: but ideally the logic of "pseudo" should be separated from the disassembler logic that lives in r_anal now

libr/anal/p/dalvik.mk
```diff
@@ -6,4 +6,4 @@ TARGET_DALVIK=anal_dalvik.${EXT_SO}
 ALL_TARGETS+=${TARGET_DALVIK}
 
 ${TARGET_DALVIK}: ${OBJ_DALVIK}
-	${CC} $(call libname,anal_dalvik) ${LDFLAGS} ${CFLAGS} -o anal_dalvik.${EXT_SO} ${OBJ_DALVIK}
+	${CC} $(call libname,anal_dalvik) ${LDFLAGS} ${CFLAGS} -I../../asm/arch -o anal_dalvik.${EXT_SO} ${OBJ_DALVIK}
```
trufae, line 9: move libr/asm/arch/dalvik into libr/anal/arch and remove this extra -I

libr/anal/p/anal_dalvik.c
trufae: yeah its probably only useful for "nop" and maybe another 0 argument opcode

trufae (changes requested):
.

trufae:
lgtm! do you wanna squash+merge?

gogo2464:
Great ! GG !

PR #20456, opened 2022-07-16
Add python pickle machine (pypm) dissassembler ##anal

swoops:

**Description**
This pull allows r2 to disassemble python pickle opcodes:

```
>  python3 -c 'import pickle;import sys;sys.stdout.buffer.write(pickle.dumps(10))' > /tmp/junk.pickle
> r2 -a pypm /tmp/junk.pickle 
[0x00000000]> pid 3
0x00000000                 8004  PROTO 0x4
0x00000002                 4b0a  BININT1 0xa
0x00000004                   2e  STOP
```
All opcodes have handlers but some are completely untested or validated.  The ops with `// TODO: validate` commented have not been tested at all. Figured I would go ahead and share in case anyone wants to play/break it.


trufae:
I think it would look better if named just "pickle", without the py_. What do you think?

swoops:
> I think it would look better if named just "pickle", without the py_. What do you think?

Sounds good to me. I will rename the files. 

Is `pypm`, for "python pickle machine" good for the arch name? Like `r2 -pypm file.pickle`? I feel like the `py` helps put it close to `pyc` arch. Also, I don't feel like python pickles are cool enough to deserve a <4 letter arch name. Any thoughts?

Also I tested most of the opcodes and they seem to work. I still need to add tests but otherwise this code is ready I think.

swoops:
I think this is ready for merging now.

There are two places where this comes up short though. Both are in regards to the older opcodes that take in newline terminated strings.

1. A newline terminated argument might be really big, if it's bigger then 128 bytes then the code gives up looking for the newline and considered the op invalid.

2. Some old numerical ops, like `LONG` accept newline terminated strings that are to be parsed into num types, like `LONG "42L"` encoded as `L42L\n` corresponds to python2's `long(42)`. The disassembler here will properly disassemble this opcode, but it won't set the `RAnalOp->val` to `42`, it won't even check if the arg is a number. It just treats it like a string.

trufae:
Pypm sounds like python package manager to me. Why not just pickle? 😃

trufae:
This pr breaks the windows build. Please check the logs https://github.com/radareorg/radare2/runs/7400860775?check_suite_focus=true

trufae:
ping

swoops:
Hopefully windows likes this better. I have very little experience compiling on windows and for good reason :)

libr/anal/p/anal_pickle.c
```diff
@@ -111,7 +111,10 @@ static inline int handle_float(RAnalOp *op, const char *name, int sz, const ut8
 
 static inline char *get_line(const ut8 *buf, int len) {
 	// TODO: find end of large strings through RAnal->iob
-	char *str = strndup ((char *)buf, len);
+	char *str = (char *)malloc (len + 1);
```
trufae, line 114: Just use r_str_ndup

libr/anal/p/anal_pickle.c
```diff
@@ -111,7 +111,10 @@ static inline int handle_float(RAnalOp *op, const char *name, int sz, const ut8
...
+	strncpy (str, (char *)buf, len);
+	str[len + 1] = '\0';
```
trufae, line 116: this is an overflow. just use r_str_ndup

libr/anal/p/anal_pickle.c
swoops: I knew that existed somewhere but couldn't find it, thanks. Fixed now.

trufae:
should be good after the cast :)

PR #20462, opened 2022-07-19
Remove optyp global variable for magic

iotaaxel:

**Description**

- Remove the `optyp` global variable for magic


libr/magic/mdump.c
```diff
@@ -44,8 +44,8 @@
 #define SZOF(a)	(sizeof(a) / sizeof(a[0]))
 
 #ifndef COMPILE_ONLY
-void file_mdump(struct r_magic *m) {
-	static R_TH_LOCAL const char optyp[] = { FILE_OPS };
+void file_mdump(struct r_magic *m, const char *optyp[]) {
+	optyp = { FILE_OPS };
```
trufae, line 48: by reading the code i dont see any place where this is modified, and it's actually a const char const string defined by the preprocessor. so for simplicity i would just remove the variable completely, keep the same function signature and just use FILE_OPS instead of `optyp` everywhere.

libr/magic/mdump.c
```diff
@@ -44,8 +44,8 @@
 #define SZOF(a)	(sizeof(a) / sizeof(a[0]))
 
 #ifndef COMPILE_ONLY
-void file_mdump(struct r_magic *m) {
-	static R_TH_LOCAL const char optyp[] = { FILE_OPS };
+void file_mdump(struct r_magic *m, const char *optyp[]) {
```
trufae, line 47: rollback this function signature change as its not necessary to pass this variable because its not modified anywhere, and you can just use the FILE_OPS define directly

trufae (changes requested):

trufae:
waiting for the CI to confirm your changes 👍

PR #20463, opened 2022-07-19
Remove daylight global variable for magic

iotaaxel:

**Description**
- Remove the `daylight` global variable for magic


libr/magic/mdump.c
```diff
@@ -210,7 +211,7 @@ const char *file_fmttime(ut32 v, int local, char *pp) {
 		}
 #endif /* HAVE_TM_ISDST */
 #endif /* HAVE_DAYLIGHT */
-		if (daylight) {
+		if (&daylight) {
```
trufae, line 214:
```suggestion
		if (*daylight) {
```

you need to access the contents with *, not the address of the reference

trufae (commented):
.

PR #20466, opened 2022-07-19
Remove last_cont_level  global variable for magic

iotaaxel:

**Description**
- Remove `last_cont_level`  global variable for magic

### DIFF
--- a/libr/include/r_magic.h
+++ b/libr/include/r_magic.h
@@ -264,6 +264,8 @@ struct r_magic_set {
 	/* FIXME: Make the string dynamically allocated so that e.g.
 	   strings matched in files can be longer than MAXstring */
 	union VALUETYPE ms_value;	/* either number or string */
+
+	ut32 last_cont_level;
 };
 
 #if USE_LIB_MAGIC
--- a/libr/magic/apprentice.c
+++ b/libr/magic/apprentice.c
@@ -958,7 +958,7 @@ static int check_cond(RMagic *ms, int cond, ut32 cont_level) {
  * parse one line from magic file, put into magic[index++] if valid
  */
 static int parse(RMagic *ms, struct r_magic_entry **mentryp, ut32 *nmentryp, const char *line, size_t lineno, int action) {
-	static R_TH_LOCAL ut32 last_cont_level = 0;
+	ms->last_cont_level = 0;
 	size_t i;
 	struct r_magic_entry *me;
 	struct r_magic *m;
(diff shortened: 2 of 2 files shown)


libr/magic/apprentice.c
```diff
@@ -957,8 +957,8 @@ static int check_cond(RMagic *ms, int cond, ut32 cont_level) {
 /*
  * parse one line from magic file, put into magic[index++] if valid
  */
-static int parse(RMagic *ms, struct r_magic_entry **mentryp, ut32 *nmentryp, const char *line, size_t lineno, int action) {
-	static R_TH_LOCAL ut32 last_cont_level = 0;
+static int parse(RMagic *ms, struct r_magic_entry **mentryp, ut32 *nmentryp, const char *line, size_t lineno, int action, ut32 *last_cont_level) {
+	last_cont_level = 0;
```
trufae, line 961:
```suggestion
	*last_cont_level = 0;
```

trufae: that change is incorrect as you are nullifying the pointer instead of the contents of it. but you need the contents of the pointer below. so it should be initialized before calling it not inside the function call

iotaaxel:
@trufae  At a glance, removing the `last_cont_level` global variable would require modifying several parts of apprentice.c.

In apprentice.c: 
- parse()
- load_b()
- load_1()
- apprentice_load()
- apprentice_1()
- file_apprentice()

Also, any references outside of apprentice.c would need to be updated. I wonder if any sections of the code can be removed / redesigned to simplify this.

trufae:
you can move this var into RMagic and use it from all the functions instead of passing pointers every call

libr/include/r_magic.h
```diff
@@ -264,6 +264,8 @@ struct r_magic_set {
 	/* FIXME: Make the string dynamically allocated so that e.g.
 	   strings matched in files can be longer than MAXstring */
 	union VALUETYPE ms_value;	/* either number or string */
+
+	ut32 last_cont_level;
```
trufae, line 268: i think this variable should be initialized to 0 in the `file_apprentice()` function  (apprentice.c:280). im not 100% sure about the use of this variable, but as long as changing the RMagicSet struct implies changing the ABI, this PR may be hold until 5.7.9 (1 release away), so i think it's a good time to squash other variables in the same PR related to rmagic, this way we break abi only once

iotaaxel: Might try to incorporate changes from pr #20474 or just hold until 5.7.9

libr/magic/apprentice.c
```diff
@@ -958,7 +959,7 @@ static int check_cond(RMagic *ms, int cond, ut32 cont_level) {
  * parse one line from magic file, put into magic[index++] if valid
  */
 static int parse(RMagic *ms, struct r_magic_entry **mentryp, ut32 *nmentryp, const char *line, size_t lineno, int action) {
-	static R_TH_LOCAL ut32 last_cont_level = 0;
+	ms->last_cont_level = 0;
```
trufae, line 962:
```suggestion
```

trufae (changes requested):

PR #20447, opened 2022-07-14
Remove global use for magic

iotaaxel:
- [x] Mark this if you consider it ready to merge
- [ ] I've added tests (optional)
- [ ] I wrote some lines in the [book](https://github.com/radareorg/radare2book) (optional)

**Description**

- The global variable was marked and moved. Will now use local static variable instead.

PR #20474, opened 2022-07-20
[5.8.0] Superseed #20447 - initializing magic globals once ##search

trufae:

**Description**


trufae:
even if passing tests, those globals should be moved into RMagic, otherwise they wont be initialized when used on different threads

trufae:
merged!

PR #20479, opened 2022-07-23
Fix p=e output ##print

richardpl:

**Description**

Test with:

dd count=8196 if=/dev/urandom of=random.raw
r2 random.raw
p=e

(and compare with 'binwalk -JE random.raw' by looking at random.raw.png)

Was this code attempting to optimize writing to terminal or something else, otherwise i see no logic in it.
Moreover name of function is then misleading.


trufae:
p=e depends on blocksize , which must be set to filesize with `r2 -f` or `b $s`, and then you pass the amount of rows (how many division of the blocksize) to show the entropy of each block. Just to confirm the behaviour was understood/expected.

the output of some tests looks more correct indeed, can you verify them and update the outputs with r2r -i test/db/path/to/file ?

PR #20483, opened 2022-07-24
Allow to change entropy bars width with '[]' keys ##print

richardpl:

**Description**

For bigger terminals no more hardcoded value for bar width.
I tried to keep old look default.


libr/util/print.c
```diff
@@ -1852,7 +1852,8 @@ R_API void r_print_fill(RPrint *p, const ut8 *arr, int size, ut64 addr, int step
 	for (i = 0; i < size; i++) {
 		cols = arr[i] > cols ? arr[i] : cols;
 	}
-	cols /= 5;
+	int div = R_MAX(255 / (p->cols * 3), 1);
```
trufae, line 1855:
```suggestion
	int div = R_MAX (255 / (p->cols * 3), 1);
```

PR #20486, opened 2022-07-25
Fix io_ptrace UB when increasing NULL pointer

richardpl:
Fix possible overreads.

**Description**

FIxed also some strange ut32 * usage while here.
Hoping I have not broken something in process.


trufae:
Just tested this PR and it doesnt work well. if you seek to address 0 and 1 you'll see incorrect results.

trufae:
Maybe worth adding a test for this (can you add it in archos/linux-x64 specific directory)?

trufae:
That's still crashing for me and i dont see a reproducible way to get that null deref. do you?

richardpl:
Where it crashes?

trufae:
nvm, i just tested the pr again, and i cant reproduce the crash i was having last itme i tried so i assume you fixed my previous comment :)

trufae:
Sorry for the delay testing and merging :) and thanks for the contrib!

PR #20494, opened 2022-07-27
Fix p=F output ##print

richardpl:

**Description**

The case was never triggered. Giving only zero bars output results.

### DIFF
--- a/libr/core/cmd_print.c
+++ b/libr/core/cmd_print.c
@@ -4324,7 +4324,7 @@ static void cmd_print_bars(RCore *core, const char *input) {
 						k++;
 					}
 					break;
-				case 'f':
+				case 'F':
 					if (p[j] == 0xff) {
 						k++;
 					}
--- a/test/db/cmd/cmd_print
+++ b/test/db/cmd/cmd_print
@@ -1258,3 +1258,16 @@ offset table offset points outside of valid range
 ERROR: bplist parse error
 EOF
 RUN
+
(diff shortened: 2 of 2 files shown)


trufae:
Why not fix the help instead? Or handle both f and F? Can u add a test? Same for /rx

richardpl:
I wanted to be compatible to some degree with pz* commands, where pzf and pzF are different beast. Adding test.

libr/core/cmd_print.c
```diff
@@ -4324,7 +4324,7 @@ static void cmd_print_bars(RCore *core, const char *input) {
 						k++;
 					}
 					break;
-				case 'f':
+				case 'F':
```
trufae, line 4327: i think its better to keep p=f instead of p=F, uppercase chars shouldnt be used unless there's no other lowercase letter left. the only uppercase one is p=A and i would like to also keep it lowercase, so im open to suggestions. can you fix the help message instead ?

richardpl: pzf does count of flags, pzF does of 0xFF so why not mimic it too with prc=f for flags and prc=F for 0xFF ?
the prc=F is mentioned even in book iiuc

trufae: oh, ok then it makes sense, but we should have p=f to handle the flags case to be in sync and maybe mention the pz[*] in the help to clarify this too. merging!

PR #20506, opened 2022-07-28
Fix not calling r_asm_op_fini() when needed

richardpl:
Fixes some leaks.

**Description**

The RAsmOp is used on stack when disassembling and after its content is not need it needs to call its _fini to free possible heap memory previously allocated.


libr/core/casm.c
```diff
@@ -599,6 +602,7 @@ static RList *r_core_asm_back_disassemble_all(RCore *core, ut64 addr, ut64 len,
 		current_buf_pos--;
 		current_instr_addr--;
 		hit_count++;
+		r_asm_op_fini (&op);
 	} while ( ((int) current_buf_pos  >= 0) && (int)(len - current_buf_pos) >= 0 && hit_count <= max_hit_count);
 
 	free(buf);
```
trufae, line 608:
```suggestion
	free (buf);
```

trufae (commented):
Foo

PR #20514, opened 2022-07-30
Fix double free when shrinking vectors ##crash

richardpl:
realloc() call can return NULL for size of 0 and still free()
that memory, there was no check for this so code ends prematurely
and double free can happen later.

**Description**


trufae:
Maybe worth checking for < 1 instead so we can handle other invalid usecases.

ideally it will be good to get rid of such big macros but im fine to merge as is

richardpl:
size_t is always unsigned AFAIK. Feel free to change.

PR #20516, opened 2022-07-30
[5.8.0] Fix memory leak in r_str_list_join()

richardpl:

**Description**

### DIFF
--- a/libr/util/str.c
+++ b/libr/util/str.c
@@ -3753,13 +3753,13 @@ R_API void r_str_stripLine(char *str, const char *key) {
 
 R_API char *r_str_list_join(RList *str, const char *sep) {
 	r_return_val_if_fail (str && sep, NULL);
+	RListIter *iter;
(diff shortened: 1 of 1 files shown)


trufae:
I think we should break abi and make this function iterate over the list instead of poping it to the end. So i would keep that change for 5.8

trufae:
i think i fixed the bug you were trying to chase on windows builds :D anyway, i still prefer to fix this memleak by making this api iterate instead of destroy the list

richardpl:
Should i rename commit message?

richardpl:
The leak still happens in master, so this merge will wait for 5.8.0.

trufae:
Yep

trufae:
can you finish this pr now that we can break apis? :D

richardpl:
What is needed? Does it still apply? The CL stuff is green ignoring random unrelated problem(s).

trufae:
Right! i though the callers should be updated, but its not the case :) merged!

PR #20527, opened 2022-07-31
Add FPU/XMM/YMM panel displays ##view

richardpl:

**Description**

So they can be tracked easily.

### DIFF
--- a/libr/core/panels.c
+++ b/libr/core/panels.c
@@ -21,6 +21,9 @@ static void __panels_refresh(RCore *core);
 #define PANEL_TITLE_XREFS_HERE       "Xrefs Here"
 #define PANEL_TITLE_XREFS            "Xrefs"
 #define PANEL_TITLE_REGISTERS        "Registers"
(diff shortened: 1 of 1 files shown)


libr/core/panels.c
```diff
@@ -4391,6 +4400,8 @@ static void __set_dcb(RCore *core, RPanel *p) {
 		p->model->directionCb = __direction_disassembly_cb;
 	} else if (__check_panel_type (p, PANEL_CMD_REGISTERS)) {
 		p->model->directionCb = __direction_register_cb;
+	} else if (__check_panel_type (p, PANEL_CMD_FPU_REGISTERS)) {
```
trufae, line 4403: add case for XMM and YMM

trufae:
ill fix upstream

trufae:
What about changing the title to be Registers+TYPE instead of TYPE+Registers. This way it will show up alphabetically sorted in the panels menu

PR #20531, opened 2022-08-01
Refactor a few eprintf to R_LOG_ERROR (part 2)

terorie:
- Refactor a few eprintf 'Cannot …' to R_LOG_ERROR
- Add r_malloc_logerr wrapper

**Description**

### DIFF
--- a/libr/core/cio.c
+++ b/libr/core/cio.c
@@ -25,7 +25,7 @@ R_API int r_core_setup_debugger(RCore *r, const char *debugbackend, bool attach)
 				r_core_cmdf (r, "dpa %d", pid);
 			}
 		} else {
-			eprintf ("Cannot retrieve pid from io.\n");
+			R_LOG_ERROR ("Cannot retrieve pid from io");
 		}
 	}
 	//this makes to attach twice showing warnings in the output
@@ -77,7 +77,7 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int 
 		fd = r_sandbox_fopen (file, "wb");
 	}
 	if (!fd) {
-		eprintf ("Cannot open '%s' for writing\n", file);
+		R_LOG_ERROR ("Cannot open '%s' for writing", file);
 		return false;
 	}
 	/* some io backends seems to be buggy in those cases */
--- a/libr/anal/cc.c
+++ b/libr/anal/cc.c
@@ -116,13 +116,13 @@ R_API char *r_anal_cc_get(RAnal *anal, const char *name) {
 	int i;
 	// get cc by name and print the expr
 	if (r_str_cmp (sdb_const_get (DB, name, 0), "cc", -1)) {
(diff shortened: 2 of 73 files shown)


libr/include/r_util/r_alloc.h
```diff
@@ -52,6 +52,22 @@ R_API void r_free(void *p);
 #define _r_free r_free
 #define _r_realloc r_realloc
 
+static inline void *r_malloc_logerr (size_t n) {
```
terorie, line 55: `r_malloc_check` sounds better honestly
but it could make monkey think that the program would abort on malloc fail

libr/include/r_util/r_alloc.h
```diff
@@ -52,6 +52,22 @@ R_API void r_free(void *p);
...
+	void *buf = malloc (n);
+	if (R_UNLIKELY (!buf)) {
+		R_LOG_ERROR ("malloc(%zu) failed", n);
```
terorie, line 58: is `%zu` even portable?

libr/include/r_util/r_alloc.h
trufae: That sounds like a good addition for 5.8, because i was planning to make malloc calls configurable at compile time, to allow linking r2 with custom allocators. But we can extend this thing to also avoid null checks on malloc failure or print a message or just return. Just like we have for rreturnvaliffail

libr/include/r_util/r_alloc.h
trufae: %p looks better to me. But maybe worth waiting for 5.7.9 to rework all those

trufae:
lgtm

libr/include/r_util/r_alloc.h
trufae: well this pr is fine to merge because its burning lots of eprintfs already. but yeah, hold those malloc changes for 3.7.9

libr/core/cio.c
```diff
@@ -84,9 +84,8 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int
 	if (bs > 4096) {
 		bs = 4096;
 	}
-	buf = malloc (bs);
+	buf = r_malloc_logerr (bs);
```
trufae, line 87: if you drop those changes, the pr is good to go

libr/core/cio.c
```diff
@@ -485,9 +484,8 @@ R_API int r_core_shift_block(RCore *core, ut64 addr, ut64 b_size, st64 dist) {
 	if ((st64)b_size < 1) {
 		return false;
 	}
-	shift_buf = calloc (b_size, 1);
+	shift_buf = r_calloc_logerr (b_size, 1);
```
trufae, line 487: imho is better to have a single logerr call to chk for null and do whatever its defined at configure time instead of having a logerr wrapper for each alloc call

PR #20542, opened 2022-08-02
[5.8.0] Remove globals in anal_arm_cs

iotaaxel:

**Description**

- Remove the global variables `obits` and `omode`
- WIP: Still thinking of how to effectively remove the global variables `ht_it` , `ht_itblock`, and `handle`

### DIFF
--- a/libr/include/r_anal.h
+++ b/libr/include/r_anal.h
@@ -667,6 +667,8 @@ typedef struct r_anal_t {
 	RList *leaddrs;
 	char *pincmd;
 	R_DIRTY_VAR;
+	int omode;
+	int obits;
 } RAnal;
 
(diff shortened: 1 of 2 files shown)


libr/include/r_anal.h
```diff
@@ -889,6 +889,8 @@ typedef struct r_anal_op_t {
 	RAnalHint hint;
 	RAnalDataType datatype;
 	int vliw; // begin of opcode block.
+	int omode;
```
trufae, line 892: RAnalOp is a struct that provides information about 1 opcode. the obits/omode should be tied to the whole analysis session, not a single instruction

iotaaxel: Got it. I will move obits/omode into the RAnal struct instead.

PR #20545, opened 2022-08-02
WIP: Use abi-compatible directory for loading plugins

trufae:

**Description**


trufae:
@MewtR thoughs on this? can you finish this PR and merge it?

MewtR:
I'll take a look

MewtR:
Is there any reason why the abi version is `Major.Minor.x` as opposed to just `Major.Minor` like it's already defined in `config-user.mk.acr`? This isn't a huge deal and can be easily changed but I'm just wondering

trufae:
Irs about clarifying the version implies 3 digits. Also it should take into account that .9 is not abi stable

PR #20550, opened 2022-08-05
Fix reference counting for RAnalBlock ##anal

richardpl:
Fixes memleaks and use after free.

**Description**

Wondering why address sanitizer is not reporting pointer dereferencing after its been freed.
Lets see if this breaks anything.


richardpl:
Some tests are failing because af- will unreference blocks associated with specified function.
Previous code was not doing this, it would half remove stuff and keep RAnalBlock memory. Does it make sense to have different functions starting at same address (even after undefinition with af-)? If not, I can update those failing tests. @trufae

trufae:
A basic block should have a unique address. Functions should start on an address that is associated with a basic block. But you can have two functions sharing basic blocks.

the same refcounting should be implemented for rregitems. But it was causing lot of problems when i tried to convert and it was breaking abi so i though on doing a simpler approach just copying the regitems when needed

richardpl:
Two zignature tests that are failing use 'af-' which removes all stuff associated with function and function itself.
The anal x86-64 test that is failing is writing different hex blocks to same starting offset but of different size, and after each time calls 'af-'. How is that supposed to work at all currently? It removes previous function but graph drawing code still finds removed functions and draw stuff. Yes, more functions can use single block, thats why it is reference counted, to not waste memory with dupe stuff. But I think using same starting offset after previous function have been removed from that same offset is invalid. I will update those tests that currently fail.

libr/core/cmd_search.c
```diff
@@ -2351,9 +2351,10 @@ static bool do_anal_search(RCore *core, struct search_parameters *param, const c
 				if (inc < 0) {
 					inc = 0;
 				}
-	 			i += inc;
-	 			at += inc;
+				i += inc;
+				at += inc;
 			}
+			r_anal_op_fini(&aop);
```
trufae, line 2357:
```suggestion
			r_anal_op_fini (&aop);
```

trufae (approved):
Looks good

ValZapod:
Wow, you work with radare2, that is so cool.

Issue #20540, opened 2022-08-02
A couple of issues related to ragg2


adwait1-g:
## Environment

```sh
$ date
Tue Aug  9 09:39:30 IST 2022
$ r2 -v
radare2 5.7.7 28613 @ linux-x86-64 git.2.0.1-12317-g977d1ade8
commit: 977d1ade8ade28fe1c72e5b20058c8adbe9e405b build: 2022-08-09__09:03:41
$ uname -ms
Linux x86_64
```

### Issue 1: 'c' type shellcode not working

Consider the following C program:

```c
$ cat code5.c
int main()
{
	return 0;
}
```

Generate 'c' type shellcode.

```sh
$ cat code5
const char cstr[27] = ""\
  "\xeb\x0efffff.\x0f\x1f\x84\x00\x00\x00\x00\x001\xc0\xc7D$\xfc"\
  "\x00\x00\x00\x00\xc3";
```

Converting it into a C program and see if the above shellcode  is working or not,

```c
$ cat code5.c.c
int main()
{
	const char cstr[27] = ""\                                                       
  "\xeb\x0efffff.\x0f\x1f\x84\x00\x00\x00\x00\x001\xc0\xc7D$\xfc"\              
  "\x00\x00\x00\x00\xc3";
	
	void (*func)() = cstr;
	func();
	
	return 0;
}
dell@adwi:~/Documents/projects/r2/ragg2-doc$ gcc code5.c.c -o code5.c.elf -zexecstack
```

Run it.

```
$ ./code5.c.elf
Segmentation fault (core dumped)
```

A peak into the issue.

```c
$ ./code5.c.elf
Segmentation fault (core dumped)
$ gdb -q code5.c.elf
Reading symbols from code5.c.elf...(no debugging symbols found)...done.
gdb-peda$ run
Starting program: /home/dell/Documents/projects/r2/ragg2-doc/code5.c.elf 

Program received signal SIGSEGV, Segmentation fault.

[----------------------------------registers-----------------------------------]
RAX: 0x0 
RBX: 0x0 
RCX: 0x5555555546f0 (<__libc_csu_init>:	push   r15)
RDX: 0x7fffffffd6d0 --> 0x841f0f2effeb 
RSI: 0x7fffffffd7e8 --> 0x7fffffffdbf3 ("/home/dell/Documents/projects/r2/ragg2-doc/code5.c.elf")
RDI: 0x1 
[...]

PR #20564, opened 2022-08-09
Fixes for Issues in #20540 (ragg2 related)

adwait1-g:
- [x] I wrote some lines in the [book](https://github.com/radareorg/radare2book) (optional)

### Fix for Issue 1:  'c' type shellcode not working

```c
const char cstr[27] = ""\
  "\xeb\x0efffff.\x0f\x1f\x84\x00\x00\x00\x00\x001\xc0\xc7D$\xfc"\
  "\x00\x00\x00\x00\xc3";
```

The `\x001` is supposed to be interpreted as '\x00-0x31' (The '1' is a character), but the C compiler is talking it as '\x00-\x01'. This is a problem. Here is the compiler version.

```sh
$ gcc --version
gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Fix for this is simply emitting hex-code even for printable characters(So that '1' is emitted as \x31).

### Fix for Issue 2: PE executables not generated on Linux-x64

The switch case code which goes over all the file formats was not covering PE format. The fix has been added - its simply calling ```create``` with the right flags. PE32 binaries are created but not PE64/PE32+ binaries - looks like it is yet to be implemented. Being tracked through https://github.com/radareorg/radare2/issues/20566.

Along with these, please consider this PR https://github.com/radareorg/radare2-book/pull/347 for r2-book on ragg2.

Regards,
Adwaith


trufae:
i dont think the new output is better, you can use `pcx` for an alternative byte-only output for C

Screenshot 2022-08-09 at 20 43 56

adwait1-g:
Okay. But leaving it untouched will lead to erroneous code in the final executable. What do we do about that?

trufae:
Why it is erroneous?

libr/util/print_code.c
```diff
@@ -147,7 +147,8 @@ R_API void r_print_code(RPrint *p, ut64 addr, const ut8 *buf, int len, char lang
 						if (buf[i] == '"') {
 							p->cb_printf ("\\\"");
 						} else {
-							p->cb_printf ("%c", buf[i]);
+							p->cb_printf("\\x%02x", buf[i]);
```
trufae, line 150:
```suggestion
							p->cb_printf ("%c", buf[i]);
```

trufae:
massaged into master

Issue #19250, opened 2021-10-17
/gg command does not work as expected


nkamadan:
## Environment

1. Sun 17 Oct 2021 11:46:40 PM +03
2. radare2 5.4.2 1 @ linux-x86-64 git.
    commit: 84e6cc6a21ec1c816d4d3eb3510d2cdc94330414 build: 2021-09-20__04:53:52
3. Linux x86_64

## Description
We have an issue with /gg command. I think it does not work as expected. I am going to try to explain the issue with an example.
![This is the control flow graph of my basic program starting from the main](https://user-images.githubusercontent.com/63115827/137644458-7ba9bd9c-c763-4e96-bd55-91e11f70be96.png)
I have set my seek to the address of function "sym.foo" and wanted to get the paths from main to that function
![The output](https://user-images.githubusercontent.com/63115827/137644508-9bf96c5e-7ae0-46e7-9e43-3eea7eb95a3e.png)

The 1st and 3rd lines in the output makes sense since these are the obvious paths from main to the function sym.foo which is called in basic block 0x1164
However, the second line does not make sense. How there can be call to 'sym.foo' without going into the basic block 0x1164?

## Test
This is the binary that we have tried as an example
[gg_option_bug.zip](https://github.com/radareorg/radare2/files/7360835/gg_option_bug.zip)

trufae:
I think a better behaviour would be something like this

* /gg? Show help
* /gg show paths in text form
* /ggj in json
* /gg current output
* /ggg in graph (same as ./gg*)

Another inconsistent point for is that /gg is actually analyzing so its not really a search. And if it is maybe it should be /a or /afbp but i did it in / because conceptually its searching over analyzed data.

what makes more sense to you?

trufae:
@nkamadan let  me know if you find some time to have a look at this. Thanks!

trufae:
i think the problem can be that the refs are still tied to the old basic block while its being split. but i need to have a look at it to see what's going on in reality

trufae:
Is 9854d41cc292dfd75c8ae195a8668a07bbeacf99 fixing this problem?

PR #20565, opened 2022-08-09
Fix for /gg

richardpl:

**Description**

Fixes for #19250 and another leak.


trufae:
Can you add a test to show the fixed /gg output? the patches lgtm

richardpl:
Trying to find similar graph layout as in bug report, takes time. Can that binary from bug report be uploaded or some other fast alternative found?
I think I found out that cyclic paths are not printed at all, with this change or without.

PR #20594, opened 2022-08-17
Remove quotes from piclkle disassembly ##anal

swoops:

**Description**

The pickle disassembler added quotes to strings and the assembler then added them as part of the string. As a result you couldn't disassemble to a file, modify, then re-assemble. For example:

```
x='STRING test'; for i in {1..10}; do x=$(rasm2 -da pickle $(rasm2 -a pickle "$x" )); echo $x; done
STRING "test"
STRING ""test""
STRING """test"""
STRING """"test""""
STRING """""test"""""
STRING """"""test""""""
STRING """""""test"""""""
STRING """"""""test""""""""
STRING """""""""test"""""""""
STRING """"""""""test""""""""""
```

I don't think quotes are really necessary for parsing or reading. So I just removed them entirely.


test/db/anal/pickle
```diff
@@ -14,13 +14,13 @@ EXPECT=<<EOF
...
-0x00000025 8d040000000000000041424344  BINUNICODE8 "ABCD"
-0x00000032         4c3334354c0a  LONG "345L"
-0x00000038 63414243440a454647480a  GLOBAL "ABCD EFGH"
-0x00000043         550441424344  SHORT_BINSTRING "ABCD"
+0x00000014                 5500  SHORT_BINSTRING 
+0x00000016         550441424344  SHORT_BINSTRING ABCD
+0x0000001c   420400000041424344  BINBYTES ABCD
+0x00000025 8d040000000000000041424344  BINUNICODE8 ABCD
+0x00000032         4c3334354c0a  LONG 345L
+0x00000038 63414243440a454647480a  GLOBAL ABCD EFGH
```
trufae, line 22: that seems incorrect. its not the same "abcd efgh" than abcd efgh... its two separate arguments. also for the color highlighting its better to have the string starting with `"` imho. if you want you can use one of the ASM_SYNTAX to tweak the output

trufae: and also add a pseudo plugin that makes all this readable.

swoops: So there is no real ambiguity if you know how pickle opcodes work but that is a cop out. 
I took the easy way out instead of updating the assembler to use quotes. I will fix it. I will see about adding a pseudo plugin as well.

I will try to amend soon.

swoops:
Closing this, the better fix will just remove this commit so I will just send a new PR.

PR #20600, opened 2022-08-18
[5.8.0] Disable the default io.basemap ##io

trufae:

**Description**


condret:
i'd prefer to kill this thing entirely

trufae:
me too. but can we keep it as an option for 1-2 releases so we can let users report and workaround in case of regression?

PR #20607, opened 2022-08-19
Allow rasm2 -f to open files with r_io files ##io

swoops:

**Description**

Updates rasm2 so that if a file fails to slurp r_io is attempted.

### DIFF
--- a/libr/main/rasm2.c
+++ b/libr/main/rasm2.c
@@ -716,12 +716,13 @@ static inline char *io_slurp(const char *file, size_t *len) {
 		des = r_io_open_nomap (io, file, R_PERM_R, 0);
 		if (des) {
 			ut64 size = r_io_desc_size (des);
-			ret = malloc (size);
-			if (!ret || !r_io_read (io, ret, size)) {
+			ret = malloc (size + 1);
+			if (size >= ST32_MAX || !ret || !r_io_read (io, ret, size)) {
 				free (ret);
 				ret = NULL;
 			} else {
 				*len = size;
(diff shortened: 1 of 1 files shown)


swoops:
Don't merge yet, found a bug

trufae:
rabin2 should be able to do the same, as it did at some point in the past. There are plans to unify rfs and rio and replace the direct posix access apis overall. but that seems like an initial approach to do that. Should IO provide a `slurp` method too? so we can reuse it from different tools :? @condret

libr/main/rasm2.c
```diff
@@ -708,6 +708,28 @@ static void __load_plugins(RAsmState *as) {
...
 }
 
+static inline char *io_slurp(const char *file, size_t *len) {
+	RIODesc *des;
+	ut8 *ret = NULL;
+	RIO *io = r_io_new ();
+	if (io) {
+		des = r_io_open_nomap (io, file, R_PERM_R, 0);
+		if (des) {
+			ut64 size = r_io_desc_size (des);
```
trufae, line 718: size can be really large. i dont think we want to assemble 3GB files for example. Maybe better if rasm could use RBuffer, which can be backed by io, this way it wont need to slurp the data

swoops: I added a check for `size >= ST32_MAX` for now. I like the idea of using `RBuffer`. It would likely simplify the code.

I would have to get deeper into the code to get that working though and I don't have a lot of free time atm. If you like the update you can merge, or wait until I make it into something better.

Issue #9782, opened 2018-03-30
rarun2 input truncating on null chars


darrellenns:
rarun2 seems to truncate input when it encouters a null byte. Here is an example test case:

The rarun2 profile - just running cat here to pass the input straight through to output
```
$ cat test.rr2
#!/usr/bin/rarun2
program=/usr/bin/cat
input=!./test.sh
```

This script makes some sample input data with null bytes in it
```
$ cat test.sh 
#!/bin/sh
echo -en "AAAAAAAAAAAAAAAABBBBBBBB\x93GUUUU\0\0\xa0\x0f\x86\xf7\xff\x7f\x00\x00\x0a"
```

Piping directly to cat, the input data is passed through unaltered.
```
$ ./test.sh|cat|hexdump -C
00000000  41 41 41 41 41 41 41 41  41 41 41 41 41 41 41 41  |AAAAAAAAAAAAAAAA|
00000010  42 42 42 42 42 42 42 42  93 47 55 55 55 55 00 00  |BBBBBBBB.GUUUU..|
00000020  a0 0f 86 f7 ff 7f 00 00  0a                       |.........|
00000029
```

Using rarun2, the input is trucated.
```
$ rarun2 ./test.rr2 | hexdump -C
00000000  41 41 41 41 41 41 41 41  41 41 41 41 41 41 41 41  |AAAAAAAAAAAAAAAA|
00000010  42 42 42 42 42 42 42 42  93 47 55 55 55 55        |BBBBBBBB.GUUUU|
0000001e
```

Arch Linux, 64 bit
radare2 2.5.0-git 1 @ linux-x86-64 git.2.5.0-git
commit: c4acac59be16cf5ab11592b426b42c2ee7f32d62 build: 2018-03-29__19:29:24

imyxh:
I can reproduce this with `r2 ./program -dR input='asdf\x00'` as well.

imyxh:
Truncation also occurs with null bytes in the file specified in an `stdin` flag.

terorie:
Can we close this?

imyxh:
> Can we close this?

@terorie this is still a bug, btw. Maybe you interpreted my earlier message to mean that it was completely fixed, but I just meant that it works in the stdin flag, while still not working with `input`.

```sh
$ rarun2 input='AAAA\x00\x00AAAA' -- /bin/cat | xxd
    00000000: 4141 4141                                AAAA
[...]

PR #20622, opened 2022-08-20
Fix r_run_parseline OOB read and truncate bug ##crash

terorie:

**Description**

- Fix r_run_parseline OOB read when using `@<n>@<pat>` (missing null terminator)
- Fix truncation when using null bytes in stdin #9782

### DIFF
--- a/libr/socket/run.c
+++ b/libr/socket/run.c
@@ -162,8 +162,8 @@ static void set_limit(int n, int a, ut64 b) {
 }
 #endif
 
-static char *getstr(const char *src) {
-	int len;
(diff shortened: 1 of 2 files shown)


libr/socket/run.c
```diff
@@ -586,7 +593,7 @@ R_API bool r_run_parseline(RRunProfile *p, const char *b) {
 	} else if (!memcmp (b, "arg", 3)) {
```
trufae, line 593: r_str_
```suggestion
	} else if (!r_str_startswith (b, "arg")) {
```

trufae (changes requested):

PR #20626, opened 2022-08-21
libFuzzer support, r_run_parseline fixes ##crash ##fuzz

terorie:
`r_run_parseline` is still badly broken
Get the fuzzers out

**Description**

- More r_run_parseline fixes (UB, leak) ##crash
- add test/fuzz to Meson with libFuzzer integration ##fuzz
- add fuzz target for r_run_parseline ##fuzz
- add fuzz target for bins ##fuzz
- add fuzz target for demangler ##fuzz


test/fuzz/README.md
```diff
@@ -0,0 +1,37 @@
...
+```shell
+cd test/fuzz
+mkdir build
+$CC \
+  -fsanitize=fuzzer,address \
+  -isystem "../../libr/include" \
+  -isystem "../../shlr/sdb/src" \
+  ../../build/libr/socket/libr_socket.a \
+  ../../build/libr/util/libr_util.a \
+  -o build/fuzz_r_run_parseline fuzz_r_run_parseline.c
```
terorie, line 29: Would appreciate any help getting this wired up with meson

libr/socket/run.c
```diff
@@ -2,6 +2,7 @@
 
 /* this helper api is here because it depends on r_util and r_socket */
 /* we should find a better place for it. r_io? */
+#include <stdbool.h>
```
trufae, line 5: that should be included already in r_types.h which comes from r_util. no need to include it here

libr/socket/run.c
```diff
@@ -166,9 +167,9 @@ static char *getstr(const char *src, R_NULLABLE size_t *out_len) {
 	size_t len = 0;
 	char *ret = NULL;
 
-	switch (*src) {
+	switch (*(src++)) {
```
trufae, line 170:
```suggestion
	src++;
	switch (*src) {
```

trufae: imho its more readable in two separate lines

libr/socket/run.c
```diff
@@ -218,46 +227,63 @@ static char *getstr(const char *src, R_NULLABLE size_t *out_len) {
...
-		int msg_len = strlen (msg);
-		if (msg_len > 0) {
-			msg [msg_len - 1] = 0;
-			char *ret = r_sys_cmd_str (msg, NULL, NULL);
-			r_str_trim_tail (ret);
-			free (msg);
-			len = strlen (ret);
-			goto beach;
-		}
+		int msg_len = strlen (src);
```
trufae, line 235:
```suggestion
		size_t msg_len = strlen (src);
```

libr/socket/run.c
```diff
@@ -218,46 +227,63 @@ static char *getstr(const char *src, R_NULLABLE size_t *out_len) {
...
+		if (msg_len == 0) {
+			R_LOG_ERROR ("Invalid backtick expression in input");
+			return NULL;
+		}
+		if (src [msg_len-1] != '`') {
```
trufae, line 240:
```suggestion
		if (src [msg_len - 1] != '`') {
```

libr/socket/run.c
```diff
@@ -218,46 +227,63 @@ static char *getstr(const char *src, R_NULLABLE size_t *out_len) {
...
+			R_LOG_ERROR ("Unterminated backtick expr in input");
+			return NULL;
+		}
+		char *msg = strdup (src);
+		if (!msg) {
+			return NULL;
+		}
+		msg [msg_len-1] = 0;
```
trufae, line 248:
```suggestion
		msg [msg_len - 1] = 0;
```

libr/socket/run.c
```diff
@@ -218,46 +227,63 @@ static char *getstr(const char *src, R_NULLABLE size_t *out_len) {
...
+		int cmd_len;
+		ret = r_sys_cmd_str (msg, NULL, &cmd_len);
```
trufae, line 250: check if `!ret`. cmd_len seems to be always initialized by r_sys_cmd_str(). so i think thats not a problem

libr/socket/run.c
```diff
@@ -218,46 +227,63 @@ static char *getstr(const char *src, R_NULLABLE size_t *out_len) {
...
+			return NULL;
+		}
+		int hexlen = r_hex_str2bin (hex, NULL);
+		if (hexlen <= 0) {
+			R_LOG_ERROR ("Invalid hexpair string");
+			free (hex);
+			return NULL;
+		}
+		len = (size_t)hexlen;
+		ret = malloc (len+1);
```
trufae, line 279:
```suggestion
		ret = malloc (len + 1);
```

libr/socket/run.c
```diff
@@ -218,46 +227,63 @@ static char *getstr(const char *src, R_NULLABLE size_t *out_len) {
...
+		r_hex_str2bin (hex, (ut8*)ret);
+		free (hex);
+		}
+		break;
 	default:
-		len = r_str_unescape ((ret = strdup (src)));
+		len = r_str_unescape ((ret = strdup (src-1)));
```
trufae, line 285:
```suggestion
		len = r_str_unescape ((ret = strdup (src - 1)));
```

test/fuzz/README.md
trufae: i think its fine to just use makes for the fuzzing testsuite for now

trufae (changes requested):

libr/socket/run.c
terorie: Will change to initialize `cmd_len = 0`. But IMO `ret` can be NULL since it's not dereferenced after the break

libr/socket/run.c
terorie: This is different behavior though, I need to read first and then increment the pointer

terorie:
> i think its fine to just use makes for the fuzzing testsuite for now

@trufae Got it to work with meson :) It's now very easy to add new fuzz targets

test/fuzz/README.md
terorie: Done with Meson!

terorie:
Passing 10M fuzz iterations now

trufae:
Awesome! The code looks good! Good to see it ready for meson too!

PR #20628, opened 2022-08-22
Use rlog instead of eprintf in 300 more places

trufae:

**Description**

### DIFF
--- a/libr/core/cio.c
+++ b/libr/core/cio.c
@@ -1,4 +1,4 @@
-/* radare2 - LGPL - Copyright 2009-2021 - pancake */
+/* radare2 - LGPL - Copyright 2009-2022 - pancake */
 
 #include "r_core.h"
 
@@ -11,7 +11,7 @@ R_API int r_core_setup_debugger(RCore *r, const char *debugbackend, bool attach)
 	p = fd ? fd->data : NULL;
 	r_config_set_b (r->config, "cfg.debug", true);
 	if (!p) {
-		eprintf ("Invalid debug io\n");
+		R_LOG_ERROR ("Invalid debug io");
 		return false;
 	}
 
@@ -86,7 +86,7 @@ R_API bool r_core_dump(RCore *core, const char *file, ut64 addr, ut64 size, int 
 	}
 	buf = malloc (bs);
 	if (!buf) {
-		eprintf ("Cannot alloc %d byte(s)\n", bs);
+		R_LOG_ERROR ("Cannot alloc %d byte(s)", bs);
 		fclose (fd);
 		return false;
 	}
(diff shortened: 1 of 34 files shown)


libr/core/cio.c
```diff
@@ -487,7 +487,7 @@ R_API int r_core_shift_block(RCore *core, ut64 addr, ut64 b_size, st64 dist) {
 	}
 	shift_buf = calloc (b_size, 1);
 	if (!shift_buf) {
-		eprintf ("Cannot allocated %d byte(s)\n", (int)b_size);
+		R_LOG_ERROR ("Cannot allocate %d byte(s)", (int)b_size);
 		return false;
 	}
```
terorie, line 492: `r_return_val_if_fail (shift_buf, false);`

libr/core/cmd_debug.c
```diff
@@ -1576,20 +1575,20 @@ static int cmd_debug_map(RCore *core, const char *input) {
 				char *buf = r_file_slurp (input + 2, &sz);
 				//TODO: use mmap here. we need a portable implementation
 				if (!buf) {
-					eprintf ("Cannot allocate 0x%08"PFMT64x" byte(s)\n", map->size);
+					R_LOG_ERROR ("Cannot allocate 0x%08"PFMT64x" byte(s)", map->size);
 					return false;
 				}
```
terorie, line 1580: `r_return_val_if_fail (buf, false);`

libr/core/cmd_debug.c
```diff
@@ -2137,18 +2136,18 @@ static void cmd_reg_profile(RCore *core, char from, const char *str) { // "arp"
 						arena->bytes = newbytes;
 						arena->size = n;
 					} else {
-						eprintf ("Cannot allocate %d\n", (int)n);
+						R_LOG_ERROR ("Cannot allocate %d", (int)n);
 					}
```
terorie, line 2140: r_return_if_fail (newbytes);

terorie (changes requested):

libr/core/cmd_debug.c
trufae: no. r_return must be used for conditions that cannot change at runtime, for example. a function passing a null pointer or negative length, because those patramters can be controlled by the caller.

this case is a condition that can fail in many different ways:
* the file doesnt exist
* the file is too large
* the heap is over

none of these conditions can be controlled at runtime, so the r_return condition **disappears** when you compile r2 in release mode. which means that in those kind of builds that distros use. the code will just segfault if any of the cases above happen. and thats not what we want

libr/core/cmd_debug.c
trufae: no

libr/core/cio.c
trufae: i was thinking in adding new r_malloc_fail macros, in 5.8 to let the developers use 3rd party alloc implementations and also unify all the malloc-checks in an unified way. but i dont want those checks to go away in release builds or have invalid asserts at runtime, because those are actual errors that cannot be controlled by the programmer

PR #20630, opened 2022-08-22
Harden swift demangler ##crash

terorie:

**Description**

- Fixes OOB memory accesses on heap and stack (not exploitable)
- Fixes infinite loops (`getnum` could return negative indices, leading demangler to rewind on crafted input)
- Fuzzed for a good 30 mins with no apparent memory leaks

### DIFF
--- a/libr/bin/mangling/swift-sd.c
+++ b/libr/bin/mangling/swift-sd.c
@@ -73,7 +73,13 @@ static const struct Type flags [] = {
 
 static const char *getnum(const char* n, int *num) {
 	if (num && *n) {
-		*num = atoi (n);
+		char *endptr;
+		int _num = (int)strtoul (n, &endptr, 10);
+		if (R_UNLIKELY (n == endptr || errno == ERANGE || _num < 0)) {
+			*num = -0x40000000; // arbitrarily large number
+		} else {
+			*num = _num;
+		}
 	}
 	while (*n && *n>='0' && *n <='9') {
 		n++;
@@ -171,23 +177,34 @@ static char *swift_demangle_lib(const char *s) {
 	return NULL;
 }
 
+static const char *str_seek (const char *s, int n) {
+	int i;
+	for (i = 0; i < n && *s != 0; i++) {
+		s++;
+	}
+	return s;
+}
+
+static inline const char *str_removeprefix (const char *s, const char *prefix) {
+	if (r_str_startswith (s, prefix)) {
+		s = s + strlen (prefix);
+	}
+	return s;
+}
+
+static void demangle_alloc_fail(RStrBuf *out) {
+	R_LOG_WARN ("demangle malloc failed");
+	r_strbuf_free (out);
+}
+
 R_API char *r_bin_demangle_swift(const char *s, bool syscmd, bool trylib) {
-#define STRCAT_BOUNDS(x) if (((x) + 2 + strlen (out)) > sizeof (out)) break;
-	char out[1024];
 	int i, len, is_generic = 0;
 	int is_first = 1;
 	int is_last = 0;
 	int retmode = 0;
-	if (!strncmp (s, "imp.", 4)) {
-		s = s + 4;
-	}
-	if (!strncmp (s, "reloc.", 6)) {
-		s = s + 6;
-	}
-
-	if (!strncmp (s, "__", 2)) {
-		s = s + 2;
-	}
+	s = str_removeprefix (s, "imp.");
+	s = str_removeprefix (s, "reloc.");
+	s = str_removeprefix (s, "__");
 	char *res = NULL;
 	if (trylib) {
 		res = swift_demangle_lib (s);
@@ -195,9 +212,9 @@ R_API char *r_bin_demangle_swift(const char *s, bool syscmd, bool trylib) {
 			return res;
 		}
 	}
-	if (*s != 'T' && strncmp (s, "_T", 2) && strncmp (s, "__T", 3)) {
+	if (*s != 'T' && !r_str_startswith (s, "_T") && !r_str_startswith (s, "__T")) {
 		// modern swift symbols not yet supported in this parser (only via trylib)
-		if (strncmp (s, "$s", 2)) {
+		if (!r_str_startswith (s, "$s")) {
 			return NULL;
 		}
 	}
@@ -216,8 +233,6 @@ R_API char *r_bin_demangle_swift(const char *s, bool syscmd, bool trylib) {
 			return res;
 		}
 	}
-
-	out[0] = 0;
 
 	const char *tail = NULL;
 	if (p[0]) {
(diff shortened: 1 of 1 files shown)


terorie:
Fixes broke tests though

libr/bin/mangling/swift-sd.c
```diff
@@ -171,33 +177,44 @@ static char *swift_demangle_lib(const char *s) {
 	return NULL;
 }
 
+static const char *str_seek (const char *s, int n) {
```
terorie, line 180: @trufae this seems useful to have in `r_str.h`, wdyt?

libr/bin/mangling/swift-sd.c
```diff
@@ -171,33 +177,44 @@ static char *swift_demangle_lib(const char *s) {
...
+	int i;
+	for (i = 0; i < n && *s != 0; i++) {
+		s++;
+	}
+	return s;
+}
+
+static const char *str_removeprefix (const char *s, const char *prefix) {
```
terorie, line 188: @trufae this also seems useful to have in `r_str.h`, wdyt?

libr/bin/mangling/swift-sd.c
```diff
@@ -171,33 +177,44 @@ static char *swift_demangle_lib(const char *s) {
...
 		res = swift_demangle_lib (s);
 		if (res) {
 			return res;
 		}
 	}
-	if (*s != 'T' && strncmp (s, "_T", 2) && strncmp (s, "__T", 3)) {
+	if (*s != 'T' && !r_str_startswith (s, "_T") && !r_str_startswith (s, "__T")) {
 		// modern swift symbols not yet supported in this parser (only via trylib)
-		if (strncmp (s, "$s", 2)) {
+		if (!r_str_startswith (s, "$s")) {
```
terorie, line 217: These 3 lines make no sense to me

terorie:
[corpus_swift_fuzz.tar.gz](https://github.com/radareorg/radare2/files/9393995/corpus_swift_fuzz.tar.gz)

Corpus

libr/bin/mangling/swift-sd.c
trufae: yes, feel free to tag this with R2_580, because we cant add new apis until 5.8.. well we can, but must be added as `static inline` in the ` .h` files

trufae:
Screenshot 2022-08-22 at 14 36 21

libr/bin/mangling/swift-sd.c
trufae, line 180:
```suggestion
static const char *str_seek(const char *s, int n) {
```

trufae: well. the name looks confusing, doesnt returns null when seeking outside and dosnt handle negative values. not sure how much useful that function will be or in how many places we will be able to use this api. if its meaningful for other places, then yes, it can be added into r_str. i dont wanna expose methods that we dont need mainly

libr/bin/mangling/swift-sd.c
```diff
@@ -171,33 +177,44 @@ static char *swift_demangle_lib(const char *s) {
...
+	int i;
+	for (i = 0; i < n && *s != 0; i++) {
+		s++;
+	}
+	return s;
+}
+
+static const char *str_removeprefix (const char *s, const char *prefix) {
+	if (r_str_startswith (s, prefix)) {
+		s = s + strlen (prefix);
```
trufae, line 190:
```suggestion
		s += strlen (prefix);
```

libr/bin/mangling/swift-sd.c
```diff
@@ -171,33 +177,44 @@ static char *swift_demangle_lib(const char *s) {
...
+	}
+	return s;
+}
+
+static void demangle_alloc_fail(RStrBuf *out) {
+	r_assert_log (R_LOGLVL_WARN, __FILE__, "demangle malloc failed");
```
trufae, line 196: use R_LOG_WARN. r_assert apis arent suposed to be called directly

libr/bin/mangling/swift-sd.c
```diff
@@ -171,33 +177,44 @@ static char *swift_demangle_lib(const char *s) {
...
-	}
-	if (!strncmp (s, "reloc.", 6)) {
-		s = s + 6;
-	}
-
-	if (!strncmp (s, "__", 2)) {
-		s = s + 2;
-	}
+	s = str_removeprefix (s, "imp.");
+	s = str_removeprefix (s, "reloc.");
```
trufae, line 206: those functionsplay well with null pointers?

libr/bin/mangling/swift-sd.c
trufae: why not? if the string doesnt starts with '$s' (and its clearly not starting with `T` or _T, then its not a swift string. am i missing anything?

anyway, i wrote this code before swift was opensource, so now it would be better to rewrite it and support more constrictions. we cant reuse the ofificla implementation because its written in swift, and iirc there's another one written in c++. but for mac/ios users. r2 uses the system implementation for this which covers much more cases

trufae (changes requested):

libr/bin/mangling/swift-sd.c
terorie: @trufae This function would be useful in the Swift and MSVC demanglers, probably others too. I proposed another variant here https://github.com/radareorg/radare2/pull/20631 called `r_str_skip`. Since skipping negative bytes makes no sense.
IMHO String subslicing with clamped bounds is a common operation in other languages so I felt it's appropriate to add. But happy to leave it as static, lmk

terorie: not for now

libr/bin/mangling/swift-sd.c
terorie: not for now

libr/bin/mangling/swift-sd.c
terorie: no, but `strncmp` doesn't either. fuzzer didn't find a null pointer reach here

libr/bin/mangling/swift-sd.c
terorie: ahh thanks makes sense now, thanks for the clarification 👍🏻 I was confused because usually `r_str_startswith` checks aren't negated

trufae:
fix the following lintings: (look like trailing tabs) 

```
libr/bin/mangling/swift-sd.c:287:	
libr/bin/mangling/swift-sd.c:530:							if (is_first) {	
Error: Proc
```

and there's a regression that is breaking the parser

Screenshot 2022-08-23 at 20 14 56

trufae:
and dis?

libr/bin/mangling/swift-sd.c
```diff
@@ -217,8 +234,6 @@ R_API char *r_bin_demangle_swift(const char *s, bool syscmd, bool trylib) {
 		}
 	}
 
-	out[0] = 0;
```
trufae, line 220: uh?

libr/bin/mangling/swift-sd.c
```diff
@@ -171,33 +177,44 @@ static char *swift_demangle_lib(const char *s) {
...
 
+static const char *str_seek(const char *s, int n) {
+	int i;
+	for (i = 0; i < n && *s != 0; i++) {
+		s++;
+	}
+	return s;
+}
+
+static const char *str_removeprefix (const char *s, const char *prefix) {
```
trufae, line 188:
```suggestion
static inline const char *str_removeprefix (const char *s, const char *prefix) {
```

libr/bin/mangling/swift-sd.c
```diff
@@ -171,33 +177,44 @@ static char *swift_demangle_lib(const char *s) {
...
+	if (r_str_startswith (s, prefix)) {
+		s += strlen (prefix);
+	}
+	return s;
+}
+
+static void demangle_alloc_fail(RStrBuf *out) {
+	r_assert_log (R_LOGLVL_WARN, __FILE__, "demangle malloc failed");
```
trufae, line 196:
```suggestion
	R_LOG_WARN ("demangle malloc failed");
```

libr/bin/mangling/swift-sd.c
```diff
@@ -217,8 +234,6 @@ R_API char *r_bin_demangle_swift(const char *s, bool syscmd, bool trylib) {
 		}
 	}
```
trufae, line 236:
```suggestion
		out[0] = 0;
```

libr/bin/mangling/swift-sd.c
```diff
@@ -262,34 +277,41 @@ R_API char *r_bin_demangle_swift(const char *s, bool syscmd, bool trylib) {
...
 	if (tail) {
-		if (*p) {
-			p++;
-		}
+		p = str_seek (p, 1);
 	} else {
-		if (*p && p[1]) {
-			p += 2;
-		}
+		p = str_seek (p, 2);
```
trufae, line 282: if (*p && p[1]) is not covered in this case.
```suggestion
		if (*p && p[1]) {
			p = str_seek (p, 2);
		}
```

trufae (changes requested):

libr/bin/mangling/swift-sd.c
terorie: out changed from an array to an `RStrBuf`. Doesn't `r_strbuf_drain(r_strbuf_new (NULL))`  already result in `"\0"`?

libr/bin/mangling/swift-sd.c
terorie: no since `RStrBuf *out` is not a char array

libr/bin/mangling/swift-sd.c
terorie: thanks

terorie:
@trufae i accidentally force pushed your changes away. could you rebase and push please?

terorie:
I don't have enough time this week to fix up this PR. Afaik, it's only one breaking parsing bug that needs to be fixed. I'm closing it for now but the commit history is still there so feel free to take finish these changes. I hope they're still useful regardless.

trufae:
ok, ill take over the pr and finish it today. thanks!

PR #20631, opened 2022-08-22
[5.8.0] add r_str_skip, r_str_skip_prefix

terorie:

**Description**

Two helper functions required for memory-safe parsing in demanglers

### DIFF
--- a/libr/util/str.c
+++ b/libr/util/str.c
@@ -3972,3 +3972,20 @@ R_API bool r_str_startswith(const char *str, const char *needle) {
 	return !strncmp (str, needle, strlen (needle));
 }
 
+
+R_API R_NONNULL const char *r_str_skip(R_NONNULL const char *s, int n) {
+	int i;
+	r_return_val_if_fail (s && n >= 0, "");
+	for (i = 0; i < n && *s != 0; i++) {
+		s++;
+	}
+	return s;
+}
+
+R_API const char *r_str_skip_prefix(const char *s, const char *prefix) {
+	r_return_val_if_fail (s && prefix, NULL);
+	if (r_str_startswith (s, prefix)) {
+		s += strlen (prefix);
+	}
+	return s;
+}
--- a/libr/include/r_util/r_str.h
+++ b/libr/include/r_util/r_str.h
@@ -63,6 +63,7 @@ R_API RCharsetRune *search_from_char(RCharsetRune *rcsr, const ut8 *ch);
 
 // str
 R_API char *r_str_repeat(const char *ch, int sz);
+R_API const char *r_str_skip(const char *s, int n);
 R_API const char *r_str_pad(const char ch, int len);
 R_API const char *r_str_rstr(const char *base, const char *p);
 R_API const char *r_strstr_ansi(const char *a, const char *b);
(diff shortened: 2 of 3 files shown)


libr/util/str.c
```diff
@@ -3972,3 +3972,19 @@ R_API bool r_str_startswith(const char *str, const char *needle) {
 	return !strncmp (str, needle, strlen (needle));
 }
 
+
+R_API const char *r_str_seek(const char *s, int n) {
```
terorie, line 3976:
```suggestion
R_API const char *r_str_skip(const char *s, int n) {
```

probably a more appropriate name since seek could imply support for negative arg

libr/util/str.c
```diff
@@ -3972,3 +3972,19 @@ R_API bool r_str_startswith(const char *str, const char *needle) {
 	return !strncmp (str, needle, strlen (needle));
 }
 
+
+R_API const char *r_str_skip(const char *s, int n) {
+	int i;
+	r_return_val_if_fail (n < 0, "");
```
trufae, line 3978:
```suggestion
	r_return_val_if_fail (s && n < 0, "");
```

libr/util/str.c
```diff
@@ -3972,3 +3972,19 @@ R_API bool r_str_startswith(const char *str, const char *needle) {
...
+	for (i = 0; i < n && *s != 0; i++) {
+		s++;
+	}
+	return s;
```
trufae, line 3982: imho this function should return NULL if you want to seek beyond the end of the string. shouldn it?

libr/util/str.c
```diff
@@ -3972,3 +3972,19 @@ R_API bool r_str_startswith(const char *str, const char *needle) {
...
+}
+
+R_API const char *r_str_skip_prefix(const char *s, const char *prefix) {
+	if (r_str_startswith (s, prefix)) {
+		s = s + strlen (prefix);
```
trufae, line 3987:
```suggestion
		s += strlen (prefix);
```

libr/util/str.c
```diff
@@ -3972,3 +3972,19 @@ R_API bool r_str_startswith(const char *str, const char *needle) {
...
+	int i;
+	r_return_val_if_fail (n < 0, "");
+	for (i = 0; i < n && *s != 0; i++) {
+		s++;
+	}
+	return s;
+}
+
+R_API const char *r_str_skip_prefix(const char *s, const char *prefix) {
+	if (r_str_startswith (s, prefix)) {
```
trufae, line 3986:
```suggestion
	r_return_val_if_fail (s && prefix, NULL);
	if (r_str_startswith (s, prefix)) {
```

trufae (changes requested):
This cant be merged because it breaks api/abi. can you evaluate if those functions make sense anywhere else of swift-sd.c ? otherwise i would prefer not to add/remove functions if we dont need them. there's already some apis that needs to be removed because no one uses them

terorie:
> can you evaluate if those functions make sense anywhere else of swift-sd.c ? 

@trufae `r_str_skip` is useful in demanglers that just do `s += 3` without checking `strlen (s)`. Swift and MSVC demanglers are full of these heap overflows. I haven't looked at other parts of the code and can't think of anything outside text parsers that would benefit from `r_str_skip`.

`r_str_skip_prefix` is a common string operation that every modern language supports

libr/util/str.c
terorie: The user can already use a `strlen ()` comparison to check if an index if out of bounds.

For text parsing, it usually simplifies code to stay in non-null territory. An empty string and an invalid string sometimes take the same code paths in an algorithm. That's why I thought it'd be a good idea to add this helper.

Semantically, the developer would want to express "give me the chars N bytes back". Then, an empty string indicates "there are no chars". Since text parsers work on char streams, they don't care whether that's due to OOB or because the string really ended. This is why Python supports sub-slicing out-of-bounds and Go equates `len([]byte{})` and `len(nil)`.

We could annotate the return parameter with `R_NONNULL` to make the API more obvious.

terorie:
Closing because I think there's a way to have it inline-only or all static. We can reopen if multiple sources depend on this

PR #20633, opened 2022-08-22
Harden msvc demangler ##crash

terorie:

**Description**

Fixes countless libFuzzer findings (stack, heap, global overflows, null deref)

### DIFF
--- a/libr/bin/mangling/microsoft_demangle.c
+++ b/libr/bin/mangling/microsoft_demangle.c
@@ -134,6 +134,26 @@ static EDemanglerErr parse_function(const char *sym, STypeCodeStr *type_code_str
 static EDemanglerErr parse_microsoft_mangled_name(const char *sym, char **demangled_name, size_t *chars_read);
 static EDemanglerErr parse_microsoft_rtti_mangled_name(const char *sym, char **demangled_name, size_t *chars_read);
 
+static const char *str_skip_report(const char *s, R_NULLABLE size_t *len, size_t n) {
+	size_t i;
+	r_return_val_if_fail (s, "");
+	for (i = 0; i < n && *s != 0; i++) {
+		s++;
+	}
+	if (len) {
+		*len += i;
+	}
+	return s;
+}
+
+static inline const char *str_skip_n(const char *s, size_t n) {
+	return str_skip_report (s, NULL, n);
+}
+
+static void state_skip_n(SStateInfo *state, size_t n) {
+	state->buff_for_parsing = str_skip_report (state->buff_for_parsing, &state->amount_of_read_chars, n);
+}
+
 static void run_state(SStateInfo *state_info, STypeCodeStr *type_code_str) {
 	state_table[state_info->state](state_info, type_code_str);
 }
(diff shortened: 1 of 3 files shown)


libr/bin/mangling/demangler_types.h
```diff
@@ -8,6 +8,7 @@ typedef enum EDemanglerErr {
 	eDemanglerErrUnsupportedMangling, ///< unsupported mangling scheme yet
 	eDemanglerErrUnkown, ///< unknown mangling scheme
 	eDemanglerErrUncorrectMangledSymbol, ///< uncorrect mangled symbol
+	eDemanglerErrWTF, ///< broken code violated invariant
```
terorie, line 11: Remove this

trufae: wtf

libr/bin/mangling/microsoft_demangle.c
```diff
@@ -145,7 +172,7 @@ static int copy_string(STypeCodeStr *type_code_str, const char *str_for_copy, si
 
 	if (free_space < str_for_copy_len) {
 		int newlen = type_code_str->type_str_len + (str_for_copy_len << 1) + 1;
-		if (newlen < 1) {
+		if (newlen < 1 || newlen > 65536) {
```
trufae, line 175: use UT
```suggestion
		if (newlen < 1 || newlen > UT16_MAX) {
```

trufae (approved):

libr/bin/mangling/microsoft_demangle.c
```diff
@@ -134,6 +134,33 @@ static EDemanglerErr parse_function(const char *sym, STypeCodeStr *type_code_str
 static EDemanglerErr parse_microsoft_mangled_name(const char *sym, char **demangled_name, size_t *chars_read);
 static EDemanglerErr parse_microsoft_rtti_mangled_name(const char *sym, char **demangled_name, size_t *chars_read);
 
+static const char *str_skip_prefix(R_NONNULL const char *s, R_NONNULL const char *prefix) {
+	if (r_str_startswith (s, prefix)) {
+		s += strlen (prefix);
+	}
+	return s;
```
trufae, line 141: i think it makes more sense to return null here

trufae: you can move this function into r_util/r_str.h as `inline` so you can use it anywhere in r2land without adding a new symbol.. we will change that in 5.8

libr/bin/mangling/microsoft_demangle.c
trufae: why this limit

trufae (approved):

trufae:
Screenshot 2022-08-22 at 20 06 34

libr/bin/mangling/microsoft_demangle.c
terorie: @trufae Found frequent OOMs here. IMO 64k is a reasonable limit for any symbol names

libr/bin/mangling/microsoft_demangle.c
terorie: > i think it makes more sense to return null here

This is supposed to be more like "remove if it exists, otherwise I don't care", like
- Python `str.removeprefix()`
- Go `strings.TrimPrefix()`

but if that doesn't make sense will fix

> you can move this function into r_util/r_str.h as inline so you can use it anywhere in r2land without adding a new symbol.. we will change that in 5.8

will do

terorie:
@trufae i fixed that lgtm bot warning already but it didn't update

PR #20647, opened 2022-08-23
[5.8.0] Initial import of the RCorePriv and RAnalPriv structs

trufae:
* Reduce abi/api breakness and clarifies the need for new apis as
* well as define cleaner public interfaces to use it programatically

**Description**


trufae:
Meh

PR #20662, opened 2022-09-02
Remove occurences of $$ in arm_cs esil ##refactor

condret:

**Description**

### DIFF
--- a/libr/anal/p/anal_arm_cs.c
+++ b/libr/anal/p/anal_arm_cs.c
@@ -2375,7 +2375,7 @@ static void arm32math(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len,
 		r_strbuf_appendf (&op->esil, "%s,", ARG(3));
 	}
 	if (!strcmp (op2, "pc")) {
-		r_strbuf_appendf (&op->esil, "%d,$$,+", pcdelta);
+		r_strbuf_appendf (&op->esil, "0x%"PFMT64x, addr + pcdelta);
 	} else {
 		r_strbuf_appendf (&op->esil, "%s", op2);
 	}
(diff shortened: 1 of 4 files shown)


libr/anal/p/anal_arm_cs.c
```diff
@@ -2439,7 +2439,7 @@ static int analop_esil(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len
 		r_strbuf_appendf (&op->esil, "%s,!,?{,32,%s,=,BREAK,},0,%s,=,%s,%s,<<,0x80000000,&,!,?{,1,%s,+=,11,GOTO,}", REG (1), REG (0), REG (0), REG (0), REG (1), REG (0));
 		break;
 	case ARM_INS_IT:
-		r_strbuf_appendf (&op->esil, "2,$$,+,pc,=");
+		r_strbuf_appendf (&op->esil, "2,pc,+,%d,pc,-,pc,:=", op->size);
```
trufae, line 2442: i dont think pc is the same as $$

condret: yes, I found that out, pc is $$ + op->size.

trufae: the register is updated before stepping

trufae: But that behaviour changes depending on thumb, arm and the alignment if its 2 or 4. We have tests for this madness and thats why it was using $$

trufae: Because you cant easily compute that from pc

condret: why is this behavior changed based on thumb? anyways, I need to deprecate $$ for esil_dfg

aemmitt-ns: op->size is right regardless of whether its thumb right?

trufae: I think that we had enough tests to confirm that the behaviour was correct. We can do a blind guess and merge and hopefully we didnt missed any test. Sgty?

Issue #20671, opened 2022-09-03
##visual ##print unsigned integer dump like 'pxd'


nik6ena:
```
[0x00000000 0x00000000 [Xadvc]8 [0x0..0x0] 1]> pxd                                                                                                                                                
- offset - | 00 01  02 03  04 05  06 07  08 09  0A 0B  0C 0D  0E 0F  10 11  12 13  14 15  16 17  18 19  1A 1B  1C 1D  1E 1F  20 21  22 23  24 25  26 27| 0123456789ABCDEF0123456789ABCDEF01234567 
0x00000000 |           22            86             6             0             0             0          2785          2729          1428             9  ....V................................... 
0x00000028 |            0          3450   -1073741825          1428             9             0          1064   -1073741825          1380             9  ....z...................(.......d....... 
0x00000050 |            0            29   -1073741825           588             9             0         47153    2147483647          8604             9  ............L...........1........!...... 
```

> | pxd[?1248]        signed integer dump (1 byte, 2 and 4)

> 
> :> pf??
> Usage: pf[.k[.f[=v]]|[v]]|[n]|[0|cnt][fmt] [a0 a1 ...]
> Examples:
> | pf n2                                        print signed short (2 bytes) value. Use N instead of n for printing unsigned values

Need an inverse of this print format in the Visual hexdump mode, i.e., all those numbers should be displayed as unsigned values.

I have just stumbled upon r2 yesterday and couldn't for the life of me figure out this one by reading the [book](https://book.rada.re/visual_mode/intro.html) or the program help text.

Is this even doable without introducing a change-set?

excerpt from https://github.com/radareorg/radare2/tree/master/libr/core/visual.c

```
static const char *printHexFormats[PRINT_HEX_FORMATS] = {
	"px", "pxa", "pxr", "prx", "pxb", "pxh", "pxw", "pxq", "pxd", "pxr",
};
```

Or a `"pxu"`? as in the C format specifier `"%u"`.

If it is of any significance here, that I am using Arch linux, latest package `radare2 5.6.6-3`, and also compared master branch to this release but don't see any commit in those 923 sum commits talking about this.

trufae:
You can use the `cmd.visual` variable to choose a different command to run in visual mode. The cmd.vprompt may be also interesting for you. it runs a command before the visual prompt

nik6ena:
Hey! Good to see you there. A fast response from the lead itself.

[...]

PR #20714, opened 2022-09-09
Add 'pxu' mode to ##visual

nik6ena:
* Unsigned integers in print format list

**Description**

Issue #20671

```
[nik@viki radare2]$ binr/radare2/radare2 -v
radare2 5.7.7 28754 @ linux-x86-64 git.5.7.7
commit: e241f89b362b36d74d3491a47306eaab94652d69 build: 2022-09-09__12:54:16
```
```
[0x00000000 *0x00000000 [Xadvc]8 ($$+0x0)]> pxu                                                        
- offset - |  0  1   2  3   4  5   6  7   8  9   A  B   C  D   E  F   0  1   2  3| 0123456789ABCDEF0123
0x00000000 |           22            86             6             0             0  ....V...............
0x00000014 |            0          2785          2729          1428             9  ....................
0x00000028 |            0          3450    3221225471          1428             9  ....z...............
```

### DIFF
--- a/libr/core/visual.c
+++ b/libr/core/visual.c
@@ -52,13 +52,13 @@ static const char *printfmtColumns[NPF] = {
 };
 
 // to print the stack in the debugger view
-#define PRINT_HEX_FORMATS 10
+#define PRINT_HEX_FORMATS 11
 #define PRINT_3_FORMATS 2
 #define PRINT_4_FORMATS 9
 #define PRINT_5_FORMATS 7
 
 static const char *printHexFormats[PRINT_HEX_FORMATS] = {
-	"px", "pxa", "pxr", "prx", "pxb", "pxh", "pxw", "pxq", "pxd", "pxr",
+	"px", "pxa", "pxr", "prx", "pxb", "pxh", "pxw", "pxq", "pxu", "pxd", "pxr",
 };
 static R_TH_LOCAL int current3format = 0;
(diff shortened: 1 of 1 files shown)


trufae:
The patch looks good :) not sure for how many people will be that relevant to have it accessible in visual, but let's have it for now as i think you will made use of it! thanks!

nik6ena:
We'll see. Unsigned is more needed than signed, IMHO.

trufae:
My main use case is for signal processing (audio waves) which use to be signed, and also jump tables for switch cases, where the delta applied to the base adddress is specified as signed too. But yeah i can imagine other places where unsigned decimal fits well :)

nik6ena:
I understand a file can have both signed and unsigned values and having the
ability to switch back and forth is much appreciated.

Otherwise we'd have to resort to setting/unsetting cmd.visual every time as
opposed to current rev. (shortcut keys o, O).

Before this, I'd to manually enter hex digits in a programmer's calculator.
Also a lot of other hex editors I'd used before, GHex, wxHexEditor, had
this panel displaying various interpretation of bytes.

I am using it, don't see the view becoming any cluttered or slow.

I say we keep it until people start complaining to you : )

trufae:
Usually nobody cares about anything, so don't worry that feature wont go away, in any case it will be reorganized to make the list of formats shorter maybe in a 2 dimensions to select sign and word size in a different way. But for now is more than enough :) thanks!

nik6ena:
Cool, thanks!

PR #20730, opened 2022-09-13
use RVector for RAnalOp src and dst instead of fix-sized array

HighW4y2H3ll:

**Description**

Hi, I spent some time changing RAnalOp.src and RAnalOp.dst to use RPVector type, they are now RAnalOp.srcs and RAnalOp.dsts. The thing was that instructions like ARM ldm/stm might have multiple src or dst, and they won't be able to fit in the limited 3 src slots or 1 dst slot in the old RAnalOp.  Currently, I've changed all the places that use them so at least it compiles, and (hopefully) functions. Please double-check if I've broken anything... Thank you!

### DIFF
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -1142,6 +1142,9 @@ repeat:
 				bb->cmpreg = op->reg;
 				r_anal_cond_free (bb->cond);
 				bb->cond = r_anal_cond_new_from_op (op);
+				if (bb->cond) {
+					src0 = src1 = NULL;
+				}
 			}
 		}
 			break;
--- a/libr/anal/fcn.c
+++ b/libr/anal/fcn.c
@@ -203,10 +203,11 @@ static bool is_delta_pointer_table(RAnal *anal, RAnalFunction *fcn, ut64 addr, u
 			mov_aop = *aop;
 			o_reg_dst = cur_dst.reg;
 			RAnalValue *rval = NULL;
-			rval = r_pvector_at(mov_aop.dsts, 0);
-			if (rval)
+			rval = r_vector_index_ptr (mov_aop.dsts, 0);
+			if (rval) {
 				cur_dst = *rval;
-			rval = r_pvector_at(mov_aop.srcs, 0);
+			}
+			rval = r_vector_index_ptr (mov_aop.srcs, 0);
 			if (rval) {
 				cur_scr = *rval;
 				reg_src = cur_scr.regdelta;
@@ -547,7 +548,7 @@ static int fcn_recurse(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 len, int
 	}
 	// TODO Store all this stuff in the heap so we save memory in the stack
 	RAnalOp *op = NULL;
-	RAnalValue *dst=NULL, *src0=NULL, *src1=NULL;
+	RAnalValue *dst = NULL, *src0 = NULL, *src1 = NULL;
 	char *movbasereg = NULL;
 	const int addrbytes = anal->iob.io ? anal->iob.io->addrbytes : 1;
 	char *last_reg_mov_lea_name = NULL;
@@ -715,13 +716,13 @@ repeat:
 			// RET_END causes infinite loops somehow
 			gotoBeach (R_ANAL_RET_END);
 		}
-		dst = r_pvector_at(op->dsts, 0);
+		dst = r_vector_index_ptr (op->dsts, 0);
 		free (op_dst);
 		op_dst = (dst && dst->reg && dst->reg->name)? strdup (dst->reg->name): NULL;
-		src0 = r_pvector_at(op->srcs, 0);
+		src0 = r_vector_index_ptr (op->srcs, 0);
 		free (op_src);
 		op_src = (src0 && src0->reg && src0->reg->name) ? strdup (src0->reg->name): NULL;
-		src1 = r_pvector_at(op->srcs, 1);
+		src1 = r_vector_index_ptr (op->srcs, 1);
 
 		if (anal->opt.nopskip && fcn->addr == at) {
 			RFlagItem *fi = anal->flb.get_at (anal->flb.f, addr, false);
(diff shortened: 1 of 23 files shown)


trufae:
Uhm, probably rvector makes more sense than rpvector. as long as the reason of having a static array was to avoid unnecessary allocations. and taking into account that most archs only have 3 operands as much.. v850 also have this reglist parameters. but maybe we can handle them keeping the array, and just changing whats stored in the array instead of supporting more operands, because at the end, the list of registers in ldm/stm is just one operand.

what do you think?

HighW4y2H3ll:
Ahh, that makes sense. It was `RAnalValue*` type for both src and dst, so I thought rpvector would be a good fit. So we could do something like:
```
op->srcs = r_vector_new(sizeof(RAnalValue), r_anal_value_free, NULL);
op->dsts = r_vector_new(sizeof(RAnalValue), r_anal_value_free, NULL);
r_vector_reserve(op->srcs, 3);
r_vector_reserve(op->dsts, 1);
```
Or... um, simply adding another vector or list field in RAnalValue? But then, do we still want to have `src[3]` in RAnalOp? since we could just keep the array inside RAnalValue in a single src operand.

libr/anal/fcn.c
```diff
@@ -202,11 +202,13 @@ static bool is_delta_pointer_table(RAnal *anal, RAnalFunction *fcn, ut64 addr, u
 			omov_aop = mov_aop;
 			mov_aop = *aop;
 			o_reg_dst = cur_dst.reg;
-			if (mov_aop.dst) {
-				cur_dst = *mov_aop.dst;
-			}
-			if (mov_aop.src[0]) {
-				cur_scr = *mov_aop.src[0];
+			RAnalValue *rval = NULL;
+			rval = r_pvector_at(mov_aop.dsts, 0);
```
trufae, line 206: add a space before the ( on every funtion call

libr/anal/fcn.c
```diff
@@ -202,11 +202,13 @@ static bool is_delta_pointer_table(RAnal *anal, RAnalFunction *fcn, ut64 addr, u
...
+			if (rval)
```
trufae, line 207: add missing braces {}

libr/anal/fcn.c
```diff
@@ -545,6 +547,7 @@ static int fcn_recurse(RAnal *anal, RAnalFunction *fcn, ut64 addr, ut64 len, int
 	}
 	// TODO Store all this stuff in the heap so we save memory in the stack
 	RAnalOp *op = NULL;
+	RAnalValue *dst=NULL, *src0=NULL, *src1=NULL;
```
trufae, line 550: spaces around the =

libr/anal/p/anal_arm_cs.c
```diff
@@ -4313,6 +4313,7 @@ static void set_opdir(RAnalOp *op) {
 static void set_src_dst(RAnalValue *val, RReg *reg, csh *handle, cs_insn *insn, int x, int bits) {
 	cs_arm_op armop = INSOP (x);
 	cs_arm64_op arm64op = INSOP64 (x);
+	if (!val)	return;
```
trufae, line 4316: braces

trufae (changes requested):
Please follow the coding style guidelines (missing spaces or braces).

Any reason towards RPVector instead of RVector? i was never a big fan of the fixed array size, but it is simple and usable. having an api to access that small array is fine, but for your usecase i think you should be having that RPVector inside RAnalValue and make that value to be of type "array".

sounds good?

HighW4y2H3ll:
Updated to use RVector now! 
There's one test case failing in the CI, but I can't seem to reproduce it on my local env tho (built with ASAN)... Any idea where should I look at?

https://github.com/radareorg/radare2/actions/runs/3069656252/jobs/4958557735#step:4:519

> R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'i
> ' bins/wasm/dotnet.wasm
> -- stdout
> @@ -1,37 +0,0 @@
> -fd       3
> -file     bins/wasm/dotnet.wasm
> -size     0x2ae89a
> -humansz  2.7M
> -minopsz  1
> -maxopsz  1
> -invopsz  1
> -pcalign  1
> -mode     r-x
> -format   wasm
> -iorw     false
> -block    0x100
> -type     EXEC
> -arch     wasm
> -baddr    0x0
> -binsz    2812058
> -bintype  wasm
> -bits     32
> -canary   false
> -retguard false
> -class    module
> -crypto   false
> -endian   little
> -havecode true
> -laddr    0x0
> -linenum  false
> -lsyms    false
> -machine  wasm
> -nx       false
> -os       WebAssembly
> -pic      false
> -relocs   false
> -sanitize false
> -static   true
> -stripped false
> -subsys   wasm
> -va       false
> 
> 
> -- stderr
> malloc(): invalid size (unsorted)
> 
> -- exit status: -1

condret:
it looks like you're allocating stuff without freeing it
ksnip_20220917-040242

HighW4y2H3ll:
But ... I also built radare2 with ASAN, and it's not catching anything. Has there been any update on dotnet.wasm? This is the output from my local run:

> % R2_NOPLUGINS=1  ./binr/radare2/radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'i
> ' test/bins/wasm/dotnet.wasm 
> radare2(21607,0x10f2a9600) malloc: nano zone abandoned due to inability to preallocate reserved vm space.
> ERROR: asm.arch: cannot find 'x86'
> ERROR: e asm.bits: Cannot set value, no plugins defined yet
> ERROR: e asm.bits: Cannot set value, no plugins defined yet
> ERROR: asm.arch: cannot find 'x86'
> fd       3
> file     test/bins/wasm/dotnet.wasm
> size     0x2ae89a
> humansz  2.7M
> mode     r-x
> format   any
> iorw     false
> block    0x100

condret:
There is currently a pr in the making by @trufae to fix an UB bug, when accessing RVector. Not sure if this bug is related to this problem and messing something up here

libr/util/vector.c
```diff
@@ -24,10 +24,13 @@
 			vec->capacity = 0; \
 			break; \
 		} \
-		void **new_a = realloc (vec->a, vec->elem_size * new_capacity); \
+		void **new_a = calloc (1, vec->elem_size * new_capacity); \
```
condret, line 27: memleak here.

Please drop all changes to this file, and rebase your pr with our master branch. The problem of uninitialized memory of this macro was recently fixed

HighW4y2H3ll: awesome! rebased

Issue #20650, opened 2022-08-24
merge RHash into RCrypto


trufae:
both libs do the same, none of them provide proper plugin interface, so better merge and write a common api for both, about libsize. hash is 100k, crypto 150k, the new crypto will be 250k, and solve the current limitations

trufae:
First stage Done! now its a matter of moving plugins and files

PR #20747, opened 2022-09-15
Crypto hash

trufae:

**Description**


trufae:
ref https://github.com/radareorg/radare2/issues/20650

trufae:
* Remove globals used in some crypto plugins, still more to go
* Use R_LOG in r_crypto
* Hold an rcrypto instance in RCore and reuse it
* RCrypto.use() returns an instance of an RCryptoJob to use it
* Add .implements as an csv names instead of the infamous .use/.check method (wip deprecation)

PR #20783, opened 2022-09-19
Implement RString (20% faster than RStrBuf) and optimize RStr.trim ##api

trufae:

**Description**


trufae:
Looks like a simpler mix between RBuffer and RStrBuf, lives in the stack and the compiler can optimize it better, weak references can help reduce memory usage quite a lot. Still buf size is static. maybe good to create a string pool with some predefined conditions, so we can have strings stored in an independent instance which holds the references to the underlying buffers without the need to copy memory.

libr/include/r_util/r_str.h
```diff
@@ -21,8 +21,17 @@ typedef enum {
 	R_STRING_ENC_GUESS = 'g',
 } RStrEnc;
 
+typedef struct {
+	char *str;
+	char buf[64]; // 20s vs 11s - 2x faster if string fits in buffer
```
trufae, line 26: this 64 is constant. having a custom allocator or pool may require extra work in this data structure that im not sure if its worth, it depends on the use cases. comments here?

PR #20812, opened 2022-09-25
Fix null pointer in aflxj ##anal

MohamedX99:

### DIFF
--- a/libr/core/cmd_anal.c
+++ b/libr/core/cmd_anal.c
@@ -4082,23 +4082,27 @@ static void cmd_aflxj(RCore *core) {
 	SdbKv *kv;
 	PJ * pj = r_core_pj_new (core);
 	if (pj) {
-		pj_o(pj);
+		pj_o (pj);
 	}
 	ls_foreach (keys, liter, kv) {
-		const char *key = sdbkv_key(kv);
+		const char *key = sdbkv_key (kv);
 		const char *value = sdbkv_value (kv);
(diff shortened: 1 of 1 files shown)


libr/core/cmd_anal.c
```diff
@@ -4090,10 +4090,10 @@ static void cmd_aflxj(RCore *core) {
 		ut64 fcn_xref_addr = r_num_get (NULL, key);
 		ut64 xref_addr = r_num_get (NULL, value);
 		RAnalFunction *xref = r_anal_get_fcn_in (core->anal, fcn_xref_addr, R_ANAL_FCN_TYPE_ANY);
-		pj_kn(pj, "address", fcn->addr);
-		pj_ks(pj, "name", fcn->name);
+		pj_kn(pj, "address", (fcn)? fcn->addr: 0);
```
trufae, line 4093: first of all this seems like an invalid case. if there's no function in line 4069 you can return early beacuse the rest of the code is suposed to be taking the xrefs of that given function.

So in that case i guess that the address doesnt have a function defined because (potatoes) but it have references to it. in theory `af` subcommands work with functions, so imho in this case it should just return early at the beginig.

```
if (!fcn) {
  R_LOG_WARN ("No function defined here");
}

trufae: Also, note that you missed the space before ( in all those calls, can u update this too in the same pr fixing the null deref? thanks!

trufae:
with return early i was refering before entering the loop below, right after checking if there's a function in the 1st line of the function

PR #20822, opened 2022-09-27
Moar arch

condret:

**Description**


libr/anal/esil.c
```diff
@@ -2021,10 +2021,10 @@ static bool esil_poke_n(RAnalEsil *esil, int bits) {
 			if (bits == 128) {
 				src2 = r_anal_esil_pop (esil);
 				if (src2 && r_anal_esil_get_parm (esil, src2, &num2)) {
-					r_write_ble (b, num, esil->anal->config->big_endian, 64);
+					r_write_ble (b, num, R_ARCH_CONFIG_IS_BIG_ENDIAN (esil->anal->config), 64);
```
trufae, line 2024: just hold `bool be =` once at the begining of the function to make it more readable and avoid to compute this many times

libr/anal/p/anal_arc.c
```diff
@@ -60,7 +60,7 @@ static int disassemble(RAnal *a, RAnalOp *op, const ut8 *buf, int len) {
 	disasm_obj.symbol_at_address_func = &symbol_at_address;
 	disasm_obj.memory_error_func = &memory_error_func;
 	disasm_obj.print_address_func = &generic_print_address_func;
-	disasm_obj.endian = !a->config->big_endian;
+	disasm_obj.endian = !R_ARCH_CONFIG_IS_BIG_ENDIAN (a->config);
```
trufae, line 63: what about using `R_ARCH_CONFIG_IS_LITTLE_ENDIAN` here?

libr/anal/p/anal_lanai_gnu.c
```diff
@@ -50,7 +50,7 @@ static int lanai_op(RAnal *a, RAnalOp *op, ut64 addr, const ut8 *buf, int len, R
 	disasm_obj.symbol_at_address_func = &symbol_at_address;
 	disasm_obj.memory_error_func = &memory_error_func;
 	disasm_obj.print_address_func = &generic_print_address_func;
-	disasm_obj.endian = !a->config->big_endian;
+	disasm_obj.endian = !R_ARCH_CONFIG_IS_BIG_ENDIAN (a->config);
```
trufae, line 53: IS_LITTLE_ENDIAN

trufae: also maybe we should use shorter macros or names here. also, as long as we support multiple endians, not just two, we should avoid using that as a boolean, if its not big endian, it can be anything, not just little

trufae (changes requested):
jjj

trufae:
fix the cornflakes

PR #20835, opened 2022-10-01
Fix long1,long4 pickle opcodes ##anal

swoops:

**Description**

The long1 and long4 opcodes were implemented wrong. This updates the assembler/disassembler to fix them, mostly.

Previously I thought `long1` would push a 8 bit number to the stack, instead it can push a 2040 bit number to the stack. The first byte after `long1` op indicates how many bytes should be interpreted as a signed little endian number and pushed onto the VM stack. The `long4` op uses 4 bytes.

Giant numbers are not completely handled, but handled well enough to get through I think.

In light of this, I also output longs as decimal numbers. Signdness gets a little funny otherwise because 0x00ff is treated as 255 while 0xff is treated as -1.

### DIFF
--- a/libr/anal/p/anal_pickle.c
+++ b/libr/anal/p/anal_pickle.c
@@ -566,7 +566,7 @@ static inline int l_num_to_bytes(ut64 n, bool sign) {
 		test = -test;
 		flip = true;
 	}
-	st64 last;
+	st64 last = 0;
 	while (test) {
 		last = test;
 		test = test >> 8;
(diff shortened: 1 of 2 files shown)


libr/anal/p/anal_pickle.c
```diff
@@ -491,19 +537,72 @@ static inline bool write_num_sz(ut64 n, int byte_sz, ut8 *outbuf, int outsz) {
...
+
+static inline int l_num_to_bytes(ut64 n, bool sign) {
+	bool flip = false;
+	st64 test = n;
+	int ret = 0;
+	if (sign && test < 0) {
+		test = -test;
+		flip = true;
+	}
+	st64 last;
```
trufae, line 569:
```suggestion
	st64 last = 0;
```

PR #20838, opened 2022-10-03
Add log error when pj depth limit reached ##util

swoops:

**Description**

I hit the JSON depth limit and it took a bit to figure out why JSON failed. This may save someone in the future some debugging. Assuming `R_LOG` is allowed here.


swoops:
If you would prefer, I could try and modify pj to accept arbitrarily depth limits from the user config.

PR #20843, opened 2022-10-04
Remove io timestamps, if it fails it fails ##io

radare:

**Description**


condret:
if you do this, also remove RIOMapRef and the mapref related functions, and please do it in 1 commit, so we can easily revert later, if needed

condret:
Forget, what I said. Keep RIOMapRef and remove ts for release builds

trufae:
It's actually implemented as a counter that increments on every new map created. so it's not really a timestamp and its not much runtime load, despite we can save one pointer indirection if we remove that, but encapsulation looks cleaner at the end.. so im closing this for now

PR #20864, opened 2022-10-08
Fix r_bin_elf_get_stripped for ELF files without section header.

QKaiser:

**Description**

ELF files without sections were automatically considered as not "not stripped". Given that static symbols and MiniDebugInfo are stored in dedicated sections (.symtab, .strtab, and .gnu_debugdata, respectively) an ELF file without any section would have a difficult time providing that information.

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -2024,7 +2024,7 @@ ut64 Elf_(r_bin_elf_get_main_offset)(ELFOBJ *bin) {
 
 bool Elf_(r_bin_elf_get_stripped)(ELFOBJ *bin) {
 	if (!bin->shdr) {
-		return false;
+		return true;
 	}
 	if (bin->g_sections) {
 		size_t i;


QKaiser:
Let me know if you want me to provide integration test files of ELF without section header. I'd just need a bit of guidance tho.

trufae:
Good finding! thanks for submiting the fix. Yes, having tests will be good! there are other bugs in ELF to be fixed in case you have the time and interest to move forward on this.

To add tests do the following steps:

* Find a binary in https://github.com/radareorg/radare2-testbins that matches your needs. if you cant find any, submit a PR adding your binary in there, bear in mind that the binaries must be legally redistributable, ideally cooked by you.
* Edit any test script that matches your needs (test/db/cmd/cmd_i seems like a good place for your needs)
* Copypaste the lines from `NAME` to `RUN` and change the fields to have a unique name, point to your file and add the tests
* Run `r2r -i test/db/cmd/cmd_i` and press 'f' when asking to update the fix the test output with the new one.
* Commit that and push the change into your pr

trufae:
Let's merge it for now. The fix is important, feel free to open a new PR with the testcase! Thanks

PR #20874, opened 2022-10-10
Switch use RArchOp instead of RAnalOp, remove duplicated enums and st…

trufae:
…ructs ##arch

**Description**


libr/include/r_arch.h
```diff
@@ -51,6 +51,38 @@ typedef struct r_arch_config_t {
 
 #define	R_ARCH_CONFIG_IS_BIG_ENDIAN(cfg_)	(((cfg_)->endian & R_SYS_ENDIAN_BIG) == R_SYS_ENDIAN_BIG)
 
+typedef struct r_arch_op_hint_t {
```
condret, line 54: Why have hints in arch, feels more like they rather belong to anal

trufae: because it needs to be tied to the ArchOp struct

trufae: archophints are just overrides of the archop details

condret: does it? as far as I've seen, it is only used in arm to set bitness. So I replaced hint with ut32 in archop

PR #20926, opened 2022-10-21
Radare2 decompiler

gogo2464:

**Description**


condret:
can you make this a core plugin?

gogo2464:
Did I do something bad?

gogo2464:
@trufae I was going to listen to @condret

trufae:
@gogo2464 this pr makes no sense, it's not even building and as condret and me told u several times in the chat before doing this PR, the r2 decompiler should be implemented in a separate repository because it's will take ages until it could be considered usable and there's no need to mess the main tree with wip and broken code

gogo2464:
ok

PR #20956, opened 2022-10-28
Fix meson with sys_xxhash

schrotthaufen:

`libr/crypto/meson.build`used `files(sys_xxhash)` instead of just `sys_xxhash`, resulting in `libr/crypto/meson.build:44:2: ERROR: files argument 1 was of type "PkgConfigDependency" but should have been "str"`.


trufae:
thanks for the fix! we have no checks for xxhash in the ci, so that wasnt spotted before. feel free to add a ci job doing a build with tcc against sysxxhash to ensure this issue doesnt happens again.

PR #20987, opened 2022-11-04
Tests for #20956

schrotthaufen:
Uses all --with-sysFOO flags

**Description**

I've added building r2 with all the `--with-sys...` configure flags.

### DIFF
--- a/.github/workflows/tcc.yml
+++ b/.github/workflows/tcc.yml
@@ -177,4 +177,4 @@ jobs:
         r2r -v
         export R2R_SKIP_ASM=1
         export R2R_SKIP_ARCHOS=1
-        make -j -C test
+        make -j -C test/unit
--- a/.github/workflows/tcc.yml
+++ b/.github/workflows/tcc.yml
@@ -131,3 +131,50 @@ jobs:
       run: |
           r2pm -r muon setup b
           r2pm -r samu -j4 -C b >/dev/null
+
+  syslibs:
+    name: ubuntu-tcc-syslibs
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout TinyCC repository
+      run: |
+          git clone https://github.com/mirror/tinycc.git
+          cd tinycc && git checkout mob
+          git reset --hard 917aad3bcfbb534875aa6d66609bdb36459460a4
+          sudo apt-get --assume-yes install gperf wheel setuptools || true
+          sudo python -m pip install --upgrade pip
+          pip install r2pipe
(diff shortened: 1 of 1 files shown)


trufae:
i guess that capstone3 support must go away

schrotthaufen:
Ubuntu 22.04 ships capstone4, if that's an option for the build image. Alternatively, I could just remove `--with-syscapstone`, but that feels a bit like missing the point.

trufae:
Capatone 3 and 4 should be fine for r2. But if you want accuracy or at least you work on latest x64 or arm64 binaries where you need pac instructions and other stuff added for the spectra then you can only use capstone5 which is not released.

i will take a look and see whats up with that option. Thanks for the pr! Stuff like this must be tested in every commit :)

trufae:
Should be fixed after merging this https://github.com/radareorg/radare2/pull/20995 can you rebase when those commits hit the master branch?

trufae:
Looks like its ssl time now

.github/workflows/tcc.yml
```diff
@@ -131,3 +131,50 @@ jobs:
...
+      with:
+          repository: radareorg/radare2-testbins
+          path: test/bins
+    - name: Configure, build and install
+      env:
+          CC: tcc
+      run: |
+          sudo apt-get --assume-yes install libxxhash-dev liblz4-dev libzip-dev libmagic-dev libcapstone-dev
+          ./configure --prefix=/usr --with-compiler=tcc \
+              --with-syscapstone \
```
trufae, line 164: maybe also add --with-oenssl ?

trufae:
Should be fixed in https://github.com/radareorg/radare2/pull/21014, pls rebase the pr and retry after merged :) thank you

.github/workflows/tcc.yml
```diff
@@ -131,3 +131,50 @@ jobs:
...
+
+    - name: Run tests
+      env:
+        PKG_CONFIG_PATH: /usr/lib/x86_64-linux-gnu/pkgconfig
+      run: |
+        r2 -v
+        r2r -v
+        export R2R_SKIP_ASM=1
+        export R2R_SKIP_ARCHOS=1
+        make -j -C test
```
trufae, line 180:
```suggestion
        make -j -C test/unit
```

trufae: If you build with any version of capstone that is not v5 with the custom patches from r2 it wont pass the testsuite. so i would just run the unit tests and merge :) all bugs fixed here. thanks for the pr!

PR #20992, opened 2022-11-06
Fix pdc leaks

Lazula:
Fixes in r_type_get_enum() also apply to cmd_type.c


libr/util/utype.c
```diff
@@ -76,6 +76,12 @@ R_API RList* r_type_get_enum(Sdb *TDB, const char *name) {
 	return res;
 }
 
+R_API void r_type_enum_free(RTypeEnum *member) {
+	free (member->name);
```
trufae, line 80: All free methods must accept null. Add an if here

trufae:
Nice ones! thanks

PR #21000, opened 2022-11-06
Fix a problem with ravc2 status on git

RHL120:

**Description**


libr/main/ravc2.c
```diff
@@ -157,7 +157,7 @@ R_API int r_main_ravc2(int argc, const char **argv) {
 	} else if (!strcmp (action, "checkout") && opt.argc > 2) {
 		save =  rvc_git_checkout (rvc, opt.argv[opt.ind + 1]);
 	} else if (!strcmp (action, "status")) {
-		char *current_branch = r_vc_current_branch (rvc);
+		char *current_branch = r_vc_git_current_branch (rvc);
```
trufae, line 160: shouldnt Rvc.currentBranch() call Rvc.gitCurrentBranch() internally? maybe the bug should be somewhere else

PR #21028, opened 2022-11-13
Don't skip byte when moving left/right in cursor mode

MewtR:

**Description**

I noticed that in the disassembly panel, when trying to move left/right in cursor mode it 'skips' a byte. This commit aims to fix that.


trufae:
Oh! nice catch! the one in visual works fine! merging!

PR #21044, opened 2022-11-15
Fix subkey removal ##bin

zutle:

**Description**
This PR fixes an issue in the type removal code that was using an invalid `char *` in an `snprintf` operation to create the name of an SDB subkey to be searched for. Consequently, we ended up iterating over all SDB keys to search for a key with the incorrect format: `.<name>.`, where the correct format should have been `<type>.<name>`.

### DIFF
--- a/libr/anal/type.c
+++ b/libr/anal/type.c
@@ -40,20 +40,22 @@ R_API void r_anal_remove_parsed_type(RAnal *anal, const char *name) {
 	if (!type) {
 		return;
 	}
-	int tmp_len = strlen (name) + strlen (type);
-	char *tmp = malloc (tmp_len + 1);
+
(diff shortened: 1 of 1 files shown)


libr/anal/type.c
```diff
@@ -36,24 +36,33 @@ R_API void r_anal_remove_parsed_type(RAnal *anal, const char *name) {
...
-			}
+
+	// Add space for two dots and zero termination
+	int subkey_len = strlen (name) + strlen (type) + 3;
+	char *subkey = malloc (subkey_len);
+	if (!subkey) {
+		return;
+	}
+
+	snprintf (subkey, subkey_len, "%s.%s.", type, name);
```
trufae, line 55: use r_str_newf

PR #21070, opened 2022-11-21
Fix size returned from r_anal_op ##anal

swoops:

**Description**

When `r_arch_decode` is used `r_anal_op` returns 2 even if `op->size` is larger. This fixes that. 

Not sure if this is what you want, so feel free to close if it's wrong.

### DIFF
--- a/libr/anal/op.c
+++ b/libr/anal/op.c
@@ -87,13 +87,14 @@ R_API int r_anal_op(RAnal *anal, RAnalOp *op, ut64 addr, const ut8 *data, int le
 	int ret = R_MIN (2, len);
 	if (len > 0 && anal->arch->session) {
 		r_anal_op_set_bytes (op, addr, data, len);
-		bool bret = r_arch_decode (anal->arch, op, mask);
-		if (!bret) {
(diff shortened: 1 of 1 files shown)


libr/anal/op.c
```diff
@@ -87,13 +87,14 @@ R_API int r_anal_op(RAnal *anal, RAnalOp *op, ut64 addr, const ut8 *data, int le
...
-		bool bret = r_arch_decode (anal->arch, op, mask);
-		if (!bret) {
+		if (!r_arch_decode (anal->arch, op, mask) || op->size <= 0) {
 			op->type = R_ANAL_OP_TYPE_ILL;
 			op->size = r_anal_archinfo (anal, R_ANAL_ARCHINFO_INV_OP_SIZE);
 			if (op->size < 0) {
 				op->size = 1;
 			}
+		} else {
+			ret = op->size;
```
trufae, line 97: that seems correct because thats how r_anal_op works. so it sounds good to me, can u fix the test that shows (data) instead of `invalid` and then ill merge.

thanks!

trufae:
Yeah i can confirm the fix is good. as its fixing the riscv PR! thank you!  i have fixed the arch.sh plugin and merged!

PR #21084, opened 2022-11-23
Move anal_riscv to arch_riscv ##arch

trufae:

**Description**


trufae:
damn, the other nasty bugs were fixed, but now we have those which are harder

PR #21095, opened 2022-11-26
Fix for the double cursor issue in panels

MewtR:

**Description**

I noticed that in panels disassembly, when you enter cursor mode, you sometimes see two cursors if too many bytes are on screen. This commit aims to fix that.


trufae:
Good patch! Merged

PR #21109, opened 2022-11-30
Add esil.dfg.mapinfo and esil.dfg.maps config vars ##anal

condret:
Allows esil dfg to take into account if a specific addr is RO and read from the corresponding map instead of the treebuf fd

**Description**


trufae:
Maybe better to have an EsilDfgOptions struct and put all those vars in there, it scales better for not having to pass many args to the function

Issue #21055, opened 2022-11-18
Integrate ABI diffing into the CI scripts


trufae:
We promise abi stability, but that's just based on human reviews and a project compromise, ideally we should have a way to generate a file that describes the ABI (all public structs and all public functions must remain the same) for one specific version and then compare that file everytime we commit a change.

* https://sourceware.org/libabigail/manual/abidiff.html

My proposal will be:

* use an already existing tool or write a script that creates a file that contains all includes stripping comments and keeps the order between versions, and use that to compare the generated one that can be downloaded from the releases page (we can publish that file on every release, and ensure we dont break that contract for all the patch releases.

PR #21215, opened 2022-12-29
Integrate ABI diffing into CI scripts

MewtR:

My attempt at addressing #21055. I essentially added a new job to the CI that compares the ABI of the latest release to the one currently being built. Let me know what you think.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -520,6 +520,39 @@ jobs:
 #        path: radare2-freebsd.tgz
 
 
+  check_abi_compatibility:
+    needs:
+      - linux-static
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+        working-directory: /tmp
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get --assume-yes update
+        sudo apt install --assume-yes abigail-tools wget
+    - name: Download Build
+      uses: actions/download-artifact@v3
+      with:
+        name: linux-static
+        path: /tmp
+    - name: Download latest release
+      run: |
+        DOWNLOAD_URL=$(curl -s https://api.github.com/repos/radareorg/radare2/releases/latest | jq -r '.assets[] | select(.name | startswith("radare2")) | select(.name | endswith("static.tar.xz")) | .browser_download_url')
+        wget -O r2-static-latest.tar.xz $DOWNLOAD_URL
+    - name: Untar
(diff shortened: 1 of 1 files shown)


trufae:
That's beautiful and I didn't expect it, which makes it even more beautiful.

Will be good to know exactly what this tool considers "ABI" because for us is fine to change internal structs and such.. can you enable the job for PRs too?

MewtR:
Thanks
Is it alright if I remove the condition on the `linux-static` job? The new one depends on it (that's the reason why it's getting skipped).

> Will be good to know exactly what this tool considers "ABI"

Good point. You can check by running something like `abidw libr_main.so --out-file libr_main.abi`
The major problem I see right now is that it considers some external dependencies as part of the ABI. Stuff from libc for example. In theory, it should be possible to ignore these through a [suppression file like this one](https://github.com/endlessm/freedesktop-sdk/blob/3bfc8a401096a4a0a660a7276a8ceae734aac79e/utils/abidiff-suppressions.ini) but I haven't been able to get it to work properly yet.

Can you give an example of something that should be changeable triggering an abi incompatibility?

trufae:
Check the `doc/abi.md` file for some examples of what we consider abi breaking changes.

Try adding a new field inside the `RCore` struct in `libr/include/r_core.h` for example.

MewtR:
Yeah so doing that does seem to have caused [the job to fail](https://github.com/MewtR/radare2/actions/runs/3803767928/jobs/6470450590)...
I'll try to look further into the suppression options to see what can be done.

trufae:
Amazing, thanks for confirming it works as expected. Let me know when it's ready to merge (i think it's fine already for me)

MewtR:
Really? In that case, I suggest we merge. If it generates too many false positives it can be disabled/reworked. But yeah it'd be cool to see it in action.

trufae:
Let's merge then! We can play and tweak it later if needed :) thank you!

trufae:
Can you check this failing log? https://github.com/radareorg/radare2/actions/runs/3855015379/jobs/6574474060

Its from the last open pr

MewtR:
Only reason I can think of for why the curl command would fail is that we're being [rate-limited by Github's API](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting).
According to the doc, there's a way to bump our requests per hour to 1000 (as opposed to 60)
I'll look into it.

PR #21219, opened 2022-12-30
Rework function merging

Fraserbc:

**Description**

r2 segfaulted when I was merging functions and this potentially fixes that.

### DIFF
--- a/libr/core/canal.c
+++ b/libr/core/canal.c
@@ -4755,6 +4755,9 @@ struct r_merge_ctx_t {
 bool fcn_merge_touch_cb(ut64 addr, struct r_merge_ctx_t *ctx) {
 	RAnalBlock *bb = r_anal_get_block_at(ctx->anal, addr);
 
+	if (!bb)
+		return true;
+
 	RListIter *iter;
 	RAnalFunction *fcn;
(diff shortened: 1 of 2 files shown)


trufae:
i think the problem of the segfault is because the basic block you are merging its being associated to two functions
thats why its freed in the first loop and used in the second
we can fix this in different ways
1) adding refcounting (this will break ABI), so not an option until 5.9
2) check the list of associated functions in the bb->fcns and use that as a "refcounter", because when we remove a bb from a function that list should be modified, and if its empty it means it can be safely removed

Fraserbc:
I have made some changes to how the merge function works. It now works by looping over all the BBs in the first function and finding where they go to (ie jump, fail, etc) and looks for basic blocks at those locations. Those basic blocks are then checked to see if they're part of function 2. If they are, then they are looped over recursively and added to function 1. This process is repeated in the other direction for function 2. I'm not sure how much of this explanation is legible, but I hope the code is self explanatory enough.

trufae:
Thanks! The new logic seems more safe! Will be good to add some tests for afm, in case you have a shareable reproducer. Good patch!

PR #21221, opened 2022-12-30
Add line information for Plan 9 ##bin

meme:

**Description**

* Fix load address of Plan 9 arm64 kernels
* Add Plan 9 calling convention for arm64 and x86-64
* Add Plan 9 line number information

Additionally, since I have rewritten the Plan 9 bin plugin entirely from its original state, with your blessing pancake I'd like to relicense it to be an MIT plugin like the Plan 9 source itself (we now borrow the line recovery algorithms from the source, specifically `hline` and `pc2line` from `sym.c` in `libmach` so I've added the foundation to the copyright header).

### DIFF
--- a/libr/bin/p/bin_p9.c
+++ b/libr/bin/p/bin_p9.c
@@ -371,6 +371,7 @@ static RBinInfo *info(RBinFile *bf) {
 	ret->bclass = strdup ("program");
 	ret->rclass = strdup ("p9");
 	ret->os = strdup ("Plan9");
+	ret->default_cc = strdup ("p9");
 	ret->arch = strdup (r_sys_arch_str (arch));
 	ret->machine = strdup (ret->arch);
 	ret->subsystem = strdup ("plan9");
--- a/libr/bin/p/bin_p9.c
+++ b/libr/bin/p/bin_p9.c
@@ -9,6 +9,8 @@
 #undef P9_ALIGN
 #define P9_ALIGN(address, align) (((address) + (align - 1)) & ~(align - 1))
 
+extern struct r_bin_dbginfo_t r_bin_dbginfo_p9;
+
 static bool check_buffer(RBinFile *bf, RBuffer *buf) {
 	RSysArch arch;
 	int bits, big_endian;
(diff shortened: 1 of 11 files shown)


trufae:
The license change is totally fine for me. MIT is a good license and its compatible with r2 (there are already other mit plugins), and as long as you are the author you are free to do that.

Thanks for the headsup!

trufae:
thats the only breaking test, can you run `r2r -i` on it? squashing it into the last commit looks fine to me. ill merge after that. Good contrib!

Screenshot 2022-12-30 at 19 08 17

meme:
Thanks, missed that one. Should be fixed. I'll let CI run.

libr/bin/p/bin_p9.c
```diff
@@ -247,66 +272,217 @@ static RList *sections(RBinFile *bf) {
...
+	size += strlen (name) + 1;
+
+	sym->value = value;
+	sym->type = type;
+	sym->name = name;
+	return size;
+}
+
+static void sym_fini(void *sym, void *user) {
+	(void)user;
```
trufae, line 322: you can use the R_UNUSED attribute inline

libr/bin/p/bin_p9.c
```diff
@@ -247,66 +272,217 @@ static RList *sections(RBinFile *bf) {
...
+	}
+	size += strlen (name) + 1;
+
+	sym->value = value;
+	sym->type = type;
+	sym->name = name;
+	return size;
+}
+
+static void sym_fini(void *sym, void *user) {
```
trufae, line 321:
```suggestion
static void sym_fini(void *sym, R_UNUSED void *user) {
```

libr/bin/p/bin_p9.c
```diff
@@ -247,66 +272,217 @@ static RList *sections(RBinFile *bf) {
...
+	// start of current level
+	Sym *start;
+	// current entry
+	Sym *h;
+	// sum of size of files this level
+	st64 delta;
+	int k;
+
+	// see fline in libmach
+	start = base;
```
trufae, line 340: i prefer definition and assignment in the same line. so you can half these LOCs, its imho easier to read this way

libr/bin/p/bin_p9.c
```diff
@@ -247,66 +272,217 @@ static RList *sections(RBinFile *bf) {
...
+	if (start != base) {
+		line = line - start->value + 1;
+	} else {
+		line = line - delta + 1;
+	}
+
+	char buffer[SDB_NUM_BUFSZ] = {0};
+	char *line_info = r_str_newf ("%s|%"PFMT64d, name, line);
+	char *address = sdb_itoa (pc, 16, buffer, sizeof (buffer));
+	sdb_set (bf->sdb_addrinfo, address, line_info, 0);
```
trufae, line 383: i did new sdb apis where the key is a number, you can use `sdb_nset` and remove the sdb_itoa lines

libr/bin/p/bin_p9.c
```diff
@@ -247,66 +272,217 @@ static RList *sections(RBinFile *bf) {
...
 
-				if (data == 0) {
-					offset += j + sizeof (ut16);
+				// read indices until a zero index
+				if (index == 0) {
+					offset += i + sizeof (ut16);
 					break;
 				}
+
+				char *name = r_pvector_at (names, index - 1);
```
trufae, line 455:
```suggestion
				const char *name = r_pvector_at (names, index - 1);
```

libr/bin/p/bin_p9.c
```diff
@@ -416,30 +666,10 @@ static ut64 size(RBinFile *bf) {
 
 #if !R_BIN_P9
 
-/* inspired in http://www.phreedom.org/solar/code/tinype/tiny.97/tiny.asm */
```
trufae, line 419: dont remove the create method pls. but it will be good to extend it to support something else than i386

trufae (changes requested):
.

PR #21222, opened 2023-01-01
Debug UX improvements

meme:

**Description**

    Fix loading source files with dir.source ##debug

    * Deprecate addr2line2 because addr2line should instead do this
    * Deprecate r_bin_addr2fileline because it is unused and users
      should use addr2line instead
    * Clean up the code in addr2text to use addr2line. Now it can handle
      get_line debug directives, whereas it previously only used the
      sdb_addrinfo table
    * Preprend the FULL PATH with dir.source instead of basename
    
---

    Handle aarch64 XMM register profile ##debug

    When using aarch64 QEMU I had failing assertions in my terminal.
    I inspected the output of the register profile and found that it
    was failing on:

            xmm     v0      .128    .2144   0
            xmm     v1      .128    .2272   0
            xmm     v2      .128    .2400   0
            xmm     v3      .128    .2528   0

    The previous entries, which are `gpr`, are handled. Instead of adding
    `xmm` back to the table, we can instead convert it into the `vec128`
    counterpart. I wonder which other entries are missing.

### DIFF
--- a/libr/reg/reg.c
+++ b/libr/reg/reg.c
@@ -127,6 +127,9 @@ R_API int r_reg_type_by_name(const char *str) {
 	if (!strcmp (str, "all")) {
 		return R_REG_TYPE_ALL;
 	}
+	if (!strcmp (str, "xmm")) {
+		return R_REG_TYPE_VEC128;
+	}
 	return -1;
 }
 
--- a/libr/bin/dbginfo.c
+++ b/libr/bin/dbginfo.c
@@ -1,4 +1,4 @@
-/* radare - LGPL - Copyright 2009-2020 - nibble, pancake */
+/* radare - LGPL - Copyright 2009-2020 - nibble, pancake, keegan */
 
 #include <r_types.h>
 #include <r_bin.h>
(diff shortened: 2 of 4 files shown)


libr/reg/reg.c
```diff
@@ -127,6 +127,9 @@ R_API int r_reg_type_by_name(const char *str) {
 	if (!strcmp (str, "all")) {
 		return R_REG_TYPE_ALL;
 	}
+	if (!strcmp (str, "xmm")) {
```
trufae, line 130: xmm type was deprecated because it was intel-specific, and the new name is vec128.

We can shorten the name from vec128 to v128 imho? what do you think?

Screenshot 2023-01-02 at 01 43 09

meme: The version of QEMU I am using (7.2.50) responds with "xmm" names when I enable the gdb stub. So, is it possible we can have a mapping of old names into the newer ones?

meme:
The failing tests are related to the functionality changes from the source lines. Previously we would take only the base name of the sources path and attempt to read the file either from the current directory, or `dir.sources`.

I have changed the behaviour to be such that the entire path is relative to `dir.sources` which I believe is much more reasonable. But, I could change it so that we first look for the full path and then fallback to the basename again?

libr/reg/reg.c
trufae: Its the gdb plugin that should be mapping those names

trufae:
Looks like the dwarf tests are failing, the tests show no data at all, i dont have any preference about relative or absolute path to find those files. Relative is usually better for tests but the broken ones look like regressions to me

trufae:
If you want you can split each commit into a separate pr

meme:
Sure I can reopen as separate PRs

PR #21223, opened 2023-01-02
Register RArch plugins to be loaded dynamically ##arch

trufae:

**Description**


libr/core/libs.c
```diff
@@ -110,6 +110,7 @@ R_API void r_core_loadlibs_init(RCore *core) {
 	DF (BIN, "bin plugins", bin);
 	DF (EGG, "egg plugins", egg);
 	DF (FS, "fs plugins", fs);
+	DF (ARCH, "archplugins", arch);
```
trufae, line 113:
```suggestion
	DF (ARCH, "arch plugins", arch);
```

PR #21234, opened 2023-01-06
Cleanup in libr/anal/op.c ##analysis

jmaselbas:

---

 - Make r_anal_optype_{to,from}_string use the same optypes array ##analysis
 - Balance spacings in braces ##indent
 - Use R_ARRAY_SIZE in r_anal_op_family_from_string
 -  Use R_ANAL_OP_DIR* enum value instead of hardcoded values

### DIFF
--- a/libr/anal/op.c
+++ b/libr/anal/op.c
@@ -570,7 +570,7 @@ static const struct op_family of[] = {
 
 R_API int r_anal_op_family_from_string(const char *f) {
 	size_t i;
-	for (i = 0; i < sizeof (of) / sizeof (of[0]); i ++) {
+	for (i = 0; i < R_ARRAY_SIZE (of); i ++) {
 		if (!strcmp (f, of[i].name)) {
 			return of[i].id;
 		}
--- a/libr/anal/op.c
+++ b/libr/anal/op.c
@@ -641,9 +641,17 @@ R_API const char *r_anal_op_direction_tostring(RAnalOp *op) {
 	if (!op) {
 		return "none";
 	}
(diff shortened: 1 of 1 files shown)


trufae:
This change is breaking the ABI, so you may need to wait for 5.9 to merge this. I would suggest to wrap the breaking changes with an #if R2_590, so we can merge now and  remove it when the 5.9 time comes.

jmaselbas:
I don't understand why nor how this breaks the ABI.
I just also noticed that there is some very similar structures in `libr/arch/arch_op.c` (code duplication)

jmaselbas:
the ABI fails on:
```
DOWNLOAD_URL=$(curl -s https://api.github.com/repos/radareorg/radare2/releases/latest | jq -r '.assets[] | select(.name | startswith("radare2")) | select(.name | endswith("static.tar.xz")) | .browser_download_url')
wget -O r2-static-latest.tar.xz $DOWNLOAD_URL
shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
jq: error (at <stdin>:1): Cannot iterate over null (null)
Error: Process completed with exit code 5.
```

trufae:
Lets rerun the job and see what happened 🤞

trufae:
Ref https://github.com/radareorg/radare2/pull/21215

trufae:
The test runs fine in local and it seems like an issue in github ci, so let's merge :)

thanks! i would love if there was an idiomatic way to generate all those enum-from/to-string methods, but the patch looks good. sorry for the delay i was afk

PR #21235, opened 2023-01-06
minor update for kvx architecture ##arch

jmaselbas:

---

- Fix script used to generate opc decoding struct for kvx
- kvx: update generated opc decoding structures
- Add missing licence in header kvx.h
- Fix typos and improve comment in kvx_op


trufae:
Looks good!  We are in the process of moving all the analysis plugins into arch. Would you like to take care of this plugin, as long as you updated it i think its better if you do this change

thanks!

trufae:
Adding some tests before doing the transition may also help on reducing regressions later

jmaselbas:
Yes, I can take care of this plugin

PR #21236, opened 2023-01-07
Use an authorized API call for the abi job

MewtR:

**Description**

This should increase the number of allowed api requests from 60 to 1000. So hopefully, we don't get any failed api calls anymore.
[Link to the doc](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -539,7 +539,10 @@ jobs:
         path: /tmp
     - name: Download latest release
       run: |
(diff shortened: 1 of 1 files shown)


trufae:
Your abi checks are working fine now :) good fix!

here's an example of a failing abi test https://github.com/radareorg/radare2/actions/runs/3867218013/jobs/6592861584

PR #21259, opened 2023-01-16
Use ut64 for bin bind offsets ##bin #if R2_590

swoops:

**Description**

The `RBinBind->get_offset` returns an `int` while the code using these values cast them to `ut64`. This will update the API to ut64. 

Still not 100% confident this code is right, figured I would let the tests run while I check a few things manually.


swoops:
ABI check works!

trufae:
Uhm not really. The abi check failed. Ao we must wait for 5.9 to introduce this patch :{

swoops:
Changing wasm should of caused merge conflicts with this. So hopefully this will fix it.

PR #21273, opened 2023-01-20
WIP: Move ws plugin into arch -- XXX requires iobind

trufae:

**Description**


condret:
allow arch to use iobind then

trufae:
I would prefer not because

* breaks the abi so needs to wait for 5.9
* The current buffer passing becomes useless
* Makes decoding time non deterministic
* exposes something that is only used by 2 plugins that barely nobody uses

trufae:
Ive found the solution to all our problems, discussed in the chat, and patched here, ill move forward in this direction so we can unblock the rest of plugins affected too 🔥

PR #21290, opened 2023-01-24
Clean-up ASN.1 parser and update x509 fuzzing

sylvainpelissier:

**Description**

Move function used only by x509 fuzzer and update checks in ASN.1 lengths to reject length over 48 bits.


trufae:
https://github.com/radareorg/radare2/pull/21292 <- this is the fix for your abi break

PR #21291, opened 2023-01-25
Avoid the use of sprintf()

trufae:
* Use r_str_newf() or snprintf ensuring buffer size
* No vulns found, but better to avoid compiler warnings
* sprintf like strcpy is a bad practice

**Description**


swoops:
I would watch out for `snprintf` too, it does not null terminate if length is exceeded so it can cause some strange bugs. I think there is a r_util alternative that forces null termination.

trufae:
Its strncpy the one that doesnt null terminate. Snprintf does it right

PR #21297, opened 2023-01-28
Use correct compilation directory for DWARF binairies

MewtR:

**Description**

(Irrelevant for DWARF5 as the compilation directory is already in the line number program header)
Files in the line number program header with 0 as their directory index end up with the wrong directory,
because only the final compilation directory is stored. This PR aims to fix that by storing compilation  directories along with their offsets from the .debug_line section (DW_AT_stmt_list). 
We can then calculate this offset when parsing the line number program header to associate files with the correct directories if needed.


trufae:
Merged! Thanks for the patch! The only thing i changed in the squash was the commit message to avoid using two ##tags.

PR #21299, opened 2023-01-28
Fix memory leak when parsing DW_FORM_line_strp ##bin

MewtR:

**Description**

Sorry missed this from my previous PR #21297. Since we support DW_FORM_line_strp now we need to free those strings as well.


trufae:
Merged because master was failing already and i had this fix pending in local

MewtR:
Thanks!

PR #21322, opened 2023-02-06
Fix the type used for the fini callback in r_arch_plugin_t ##arch

jmaselbas:
The fini callback has a type (RArchPluginFiniCallback) use that.

### DIFF
--- a/libr/include/r_arch.h
+++ b/libr/include/r_arch.h
@@ -163,7 +163,7 @@ typedef struct r_arch_plugin_t {
 	RSysBits bits;
 	RSysBits addr_bits;
 	RArchPluginInitCallback init;
-	RArchPluginInitCallback fini;
+	RArchPluginFiniCallback fini;
 	RArchPluginInfoCallback info;
 	RArchPluginRegistersCallback regs;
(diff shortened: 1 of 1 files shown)


trufae:
Good catch! Unfortunately this breaks the abidiff ci test. So we should wait for 5.9 to merge this. Despite i think both init a d finive the sane signature and shouldnt be breaking anything, so maybe thats a bug that shoukd be reported to libabigail/abidiff. I will check this and report them back :)

thanks!

jmaselbas:
Yes, they both share the same signature so it is just a name, types are compatible. But it better to keep them separated if they need to be different in the future. This can easy wait for the 5.9 release

trufae:
Seems like this is properly handled with abidiff from git.. must chk with abidiff2.2, which is the last release. but ubuntu-latest... ships an ancient version (2.0) and i have to see how this can be updated in github and try again. but your change doesnt breaks the abi :)

trufae:
I can confirm it's fixed in abidiff 2.2 :) it's just coz ubuntu22.02 is too old

PR #21323, opened 2023-02-06
 Convert kvx analysis plugin into arch ##arch 

jmaselbas:
Notable change during the conversion: the arch plugin now uses init/fini for decode context which might not be the final solution but it is better than using a per thread local storage.


jmaselbas:
I am marking this as a draft, I will like to test this a bit more (I only compiled r2 and haven't tested this properly)

trufae:
The meson build is what is missing here to be fully green, i can fix it if you want, but to build with meson just do: `meson b; ninja -C b`. guess you'll need to fix libr/meson.build and libr/(anal|arch)/meson.build too. Let's make the arch transition happen!

jmaselbas:
I have this error, when running `meson b`:
```
shlr/meson.build:211:2: ERROR: File capstone/arch/BPF/BPFDisassembler.c does not exist.
```
Did I miss a setup step ?

trufae:
Looks great! merged!

PR #21341, opened 2023-02-13
WIP: Aim to add POP and DUP for dfg ##esil

trufae:

**Description**


trufae:
all yours @condret imho dfg needs to implement DUP too, but we all agree on POP, but thre's also the thing that & updates the internal flags. which wasnt clear.. so we need further discussions and more tests here

PR #21372, opened 2023-02-21
Fix two undefined behaviour bugs in esil and reg

trufae:

**Description**


hamaney:
Nice. Thanks, a lot. I will test today!

PR #21374, opened 2023-02-22
Add Inferno Dis VM ##bin ##arch

meme:

**Description**

This PR adds support for the Inferno Dis VM bytecode format.

It adds a bin plugin, which is capable of reading the Dis bytecode format, as well as a Dis disassembler which is capable of decoding all instructions available in the fourth edition, including undocumented ones. It implements basic analysis for control flow and is able to recover control flow graphs.

For now, the following features *are not yet* implemented:
* No support for computed-gotos or `case` statements
* The link section and type section are ignored for now
* If a signature is present, it is not validated, only skipped

However, the plugin is fully usable and I have tested it on a number of Dis bytecode files from the Inferno distribution. I plan on making a follow-up PR with tests for Plan 9 a.out as well as Dis.

**References**

* The Dis VM specification can be found here: https://github.com/inferno-os/inferno-os/blob/master/doc/dis.pdf

* Test binaries, such as `sh.dis` can be found here (in the `dis` folder): https://github.com/inferno-os/inferno-os/blob/master/dis/sh.dis


libr/bin/p/bin_dis.c
```diff
@@ -0,0 +1,302 @@
...
+}
+
+static bool load_buffer(RBinFile *bf, void **bin_obj, RBuffer *buf, ut64 loadaddr, Sdb *sdb) {
+	ut32 i;
+
+	if (!check_buffer (bf, buf)) {
+		return false;
+	}
+
+	RBinDisObj *o = R_NEW0 (RBinDisObj);
```
trufae, line 31: check for null

libr/bin/p/bin_dis.c
```diff
@@ -0,0 +1,302 @@
...
+#include <r_lib.h>
+#include <r_bin.h>
+
+#include "../../arch/p/dis/dis.h"
+#include "../../arch/p/dis/dis.c"
+
+static bool check_buffer(RBinFile *bf, RBuffer *buf) {
+	ut64 pos = r_buf_tell (buf);
+	r_buf_seek (buf, 0, R_BUF_SET);
+	st32 magic;
```
trufae, line 16: woah signed magic \o/

libr/bin/p/bin_dis.c
```diff
@@ -0,0 +1,302 @@
...
+		goto invalid;
+	}
+	if (o->header.stack_extent < 0) {
+		goto invalid;
+	}
+
+	if (!dis_read_operand (buf, &o->header.code_size)) {
+		goto invalid;
+	}
+	if (o->header.code_size < 0) {
```
trufae, line 63:
```suggestion
	if (o->header.code_size < 1) {
```

if codesize is 0 it should be invalid too

libr/bin/p/bin_dis.c
```diff
@@ -0,0 +1,302 @@
...
+		goto invalid;
+	}
+
+	if (!dis_read_operand (buf, &o->header.data_size)) {
+		goto invalid;
+	}
+	if (o->header.data_size < 0) {
```
trufae, line 70: same here :?

libr/bin/p/bin_dis.c
```diff
@@ -0,0 +1,302 @@
...
+		if (!dis_read_instr (buf, &instr)) {
+			R_LOG_ERROR ("Bad Dis instruction (pc is 0x%x)", i);
+			goto invalid;
+		}
+	}
+	o->code_size = r_buf_tell (buf) - addr;
+
+	// parse type section
+	addr = r_buf_tell (buf);
+	for (i = 0; i < o->header.type_size; i++) {
```
trufae, line 120: save the o->header.type_size into a const variable before entering the loop to avoid the unnecessary o-> dereference on each iteration

libr/bin/p/bin_dis.c
```diff
@@ -0,0 +1,302 @@
...
+
+	if (!(ret = r_list_new ())) {
+		return NULL;
+	}
+
+	ret->free = free;
+
+	if ((ptr = R_NEW0 (RBinAddr))) {
+		ptr->paddr = entry_address;
+		ptr->vaddr = entry_address;
```
trufae, line 187: maybe vaddr should honor the baddr, but as long as its a vm i guess that its fine tha pa == va

libr/bin/p/bin_dis.c
```diff
@@ -0,0 +1,302 @@
...
+
+static bool check_buffer(RBinFile *bf, RBuffer *buf) {
+	ut64 pos = r_buf_tell (buf);
+	r_buf_seek (buf, 0, R_BUF_SET);
+	st32 magic;
+	if (!dis_read_operand (buf, &magic)) {
+		return false;
+	}
+	r_buf_seek (buf, pos, R_BUF_SET);
+	return magic == XMAGIC || magic == SMAGIC;
```
trufae, line 21: maybe add some size limits to avoid false positives too. is the 4 byte magic the only valid signature here?

trufae (changes requested):
looks good. my concerns are:

- code section depends on instruction decoding, which results in slow loading, cant be determined in a constant time :? or
- can you add (in a separate PR or commit) the magic in the magic database? may be useful for carving memory
- i was afraid that using the disassembler from two modules would result in dupped symbols on static builds, but seems like thats fine for the ci, so i guess the visibility of those symbols is correct

meme:
> code section depends on instruction decoding, which results in slow loading, cant be determined in a constant time :? or

I don't think so. Instructions have variable length encoding, and there is no indication of how big the code section is, only the number of instructions. There are also no markers between sections, and it is not possible to parse the file backwards either. I think if we come across a binary that is slow we can investigate then, but since most modules are loaded dynamically, they are not super big in isolation.

> can you add (in a separate PR or commit) the magic in the magic database? may be useful for carving memory

Sure.

> i was afraid that using the disassembler from two modules would result in dupped symbols on static builds, but seems like thats fine for the ci, so i guess the visibility of those symbols is correct

Yup I validated it with a static build and it functions as intended.

libr/bin/p/bin_dis.c
meme: I don't implement `baddr` so it is 0... am I supposed to read the user-supplied base address somewhere for this?

libr/bin/p/bin_dis.c
meme: I think for 0 sized data this is fine. It is possible for a Dis program to simply call into another module, which does not require data declarations.

libr/bin/p/bin_dis.c
```diff
@@ -0,0 +1,302 @@
...
+		goto invalid;
+	}
+
+	o->header_size = r_buf_tell (buf);
+
+	// in order to determine the size of the code section, we must parse
+	// all instructions. if we encounter an invalid instruction we must fail
+	// because instruction sizes are variable.
+	ut64 addr = r_buf_tell (buf);
+	for (i = 0; i < o->header.code_size; i++) {
```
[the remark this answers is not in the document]
meme: Code size is *actually* the number of instructions, I am just using the same naming scheme as the specification.

For that reason, it's not possible to use it for any bounds checking since instructions can be variably sized.

trufae:
Thanks for all the replies! Looks good to merge!

PR #21384, opened 2023-02-23
Fix libs having proper SONAME on Darwin

Teutates:

**Description**

On Darwin systems, the normal SONAME convention is libexample.{ABIVERSION}.dylib. This fixes that.


trufae:
Thanks!

trufae:
Maybe, as long as now r2 is abi stable for major.minor releases we may enforce that change for non darwin targets too

Diatrus:
@trufae I only have full knowledge on how SOVER is handled on Darwin, but that sounds like it would be correct.

PR #21386, opened 2023-02-24
Fix #21380 - Hint the user with quoting the /R and /ad/ ##shell

### LINKED ISSUE #21380: improve /R command docs to include ""
## Description
for some regexp you need to surround them with quotations "likethis"

*Provide images, ascii-art, test files and anything that may help us understand your request*
Screen Shot 2023-02-22 at 08 19 54 AM

trufae:

**Description**

### DIFF
--- a/libr/core/cmd_search.c
+++ b/libr/core/cmd_search.c
@@ -161,7 +161,7 @@ static const char *help_msg_slash_a[] = {
 	"/ab", " [delta]", "search for backward jumps (usually loops)",
 	"/ac", " mov eax", "same as /aa, but case-sensitive",
 	"/ad", "[/*j] push;mov", "match ins1 followed by ins2 in linear disasm",
-	"/ad/", " ins1;ins2", "search for regex instruction 'ins1' followed by regex 'ins2'",
+	"/ad/", " ins1;ins2", "search for regex 'ins1' followed by regex 'ins2' ('""' to use ;|` in the regex)",
 	"/ad/a", " instr", "search for every byte instruction that matches regexp 'instr'",
 	"/ae", " esil", "search for esil expressions matching substring",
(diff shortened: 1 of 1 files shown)


libr/core/cmd_search.c
```diff
@@ -207,7 +207,7 @@ static const char *help_msg_slash_r[] = {
 };
 
 static const char *help_msg_slash_R[] = {
-	"Usage: /R", "", "search for ROP gadgets",
+	"Usage: /R", "", "search for ROP gadgets (hint: quote the command if your regex uses special chars)",
```
hamaney, line 210: optional : (hint:  surround your regex with a double quotation if your regex uses special chars. eg: "regex1 | regex2" ) 

I think it is more clear.

trufae: thats not correct. because its the whole command that needs to be quoted or just start the command with ""

hamaney: Yes. This was before learning it yesterday. We need the word R2 command with the quote.

should render like this:
```
search for ROP gadgets (hint: quote the whole r2 command if your regex uses special chars. eg: "/R/ regx1;regx2")
```

trufae: that's too large to fit in a 80x24 screen for something that affects all commands, imho is better to add a hint and put a proper explanation somewhere else in a generic way for all the commands that the user may want to use special foes

trufae: i mean, all shells have special characters, and none of them reminds the user to escape the chars

AbdulAbCellera: Fair enough. Would the length of you current changes allow for 2 more chars for (r2) ?

hamaney: Fair enough. Would the length of you current changes allow for 2 more chars for (r2) ?

PR #21396, opened 2023-02-24
Remove some R_TH_LOCAL globals

jmaselbas:

Remove some R_TH_LOCAL globals in two arch plugin: snes and or1k

### DIFF
--- a/libr/arch/p/or1k/plugin.c
+++ b/libr/arch/p/or1k/plugin.c
@@ -16,8 +16,21 @@ struct operands {
 };
 
 // XXX remove globals and move into init/fini data pointer or just reimplement this into an ctual anal plugin for 5.9 so maybe its not worth
-static R_TH_LOCAL ut32 cpu[32] = {0}; /* register contents */
-static R_TH_LOCAL ut32 cpu_enable; /* allows to treat only registers with known value as valid */
+struct or1k_regs {
+	ut32 cpu[32]; /* register contents */
+	ut32 cpu_enable; /* allows to treat only registers with known value as valid */
+};
+
+static bool or1k_init(RArchSession *s) {
+	r_return_val_if_fail (s && !s->data, false);
+	s->data = R_NEW0 (struct or1k_regs);
+	return s->data? true: false;
+}
+
+static bool or1k_fini(RArchSession *s) {
+	R_FREE (s->data);
+	return true;
+}
 
 static char *insn_to_str(ut64 addr, insn_t *descr, insn_extra_t *extra, ut32 insn) {
 	struct operands o = {0};
@@ -101,7 +114,7 @@ static ut64 n_oper_to_addr(ut32 n, ut32 mask, ut64 addr) {
 	return (ut64) ((st64) ((st32) (sign_extend(n, mask) << 2)) + addr);
 }
 
-static int insn_to_op(RAnalOp *op, ut64 addr, insn_t *descr, insn_extra_t *extra, ut32 insn) {
+static int insn_to_op(struct or1k_regs *regs, RAnalOp *op, ut64 addr, insn_t *descr, insn_extra_t *extra, ut32 insn) {
 	struct operands o = {0};
 	insn_type_t type = type_of_opcode (descr, extra);
 	insn_type_descr_t *type_descr = &types[INSN_X];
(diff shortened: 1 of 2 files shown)


libr/arch/p/or1k/plugin.c
```diff
@@ -16,8 +16,22 @@ struct operands {
...
 // XXX remove globals and move into init/fini data pointer or just reimplement this into an ctual anal plugin for 5.9 so maybe its not worth
-static R_TH_LOCAL ut32 cpu[32] = {0}; /* register contents */
-static R_TH_LOCAL ut32 cpu_enable; /* allows to treat only registers with known value as valid */
+struct or1k_regs {
+	ut32 cpu[32]; /* register contents */
+	ut32 cpu_enable; /* allows to treat only registers with known value as valid */
+};
+
+static bool or1k_init(RArchSession *s) {
+	r_return_val_if_fail (!s->user, false);
```
trufae, line 25: imho its better to use `->data` instead.. but im seeing some inconsistencies (wasm plugin is >user) but evm and bpf use >data instead.

imho user holds a user controlled pointer. and data its plugin-specific data. the thing is that there's no comment to clarify this, so its normal that is missused, and also i see no way to set this user pointer from the arch api.

```
130$ git grep '>user' *.c
arch_session.c:         ai->user = NULL;
0$ git grep '>data' *.c
arch_session.c:                 bool res = init (ai); // must fill ai->data
0$
```

this is the only comment referencing this. so imho we can add comments about this in the r_arch.h and so

trufae, line 25: this check seems wrong to me, s->user must be null because the user poitjner is not used yet. so better assert if the session is null instead of the user pointer

libr/arch/p/or1k/plugin.c
```diff
@@ -16,8 +16,22 @@ struct operands {
...
+	s->user = calloc (1, sizeof (struct or1k_regs));
```
trufae, line 26:
```suggestion
	s->user = R_NEW0 (struct or1k_regs);
```

libr/arch/p/or1k/plugin.c
```diff
@@ -16,8 +16,22 @@ struct operands {
...
+	return s->user? true: false;
+}
+
+static bool or1k_fini(RArchSession *s) {
+	free (s->user);
```
trufae, line 31:
```suggestion
	R_FREE (s->data);
```

libr/arch/p/or1k/plugin.c
```diff
@@ -16,8 +16,22 @@ struct operands {
...
+	s->user = NULL;
```
trufae, line 32:
```suggestion
```

libr/arch/p/snes/plugin.c
```diff
@@ -12,9 +12,19 @@ struct snes_asm_flags {
 	ut8 X;
 };
 
-static R_TH_LOCAL struct snes_asm_flags snesflags = {0};
+static bool snes_init(RArchSession *s) {
+	r_return_val_if_fail (!s->user, false);
+	s->user = calloc (1, sizeof (struct snes_asm_flags));
```
trufae, line 17:
```suggestion
	s->user = R_NEW0 (struct snes_asm_flags);
```

libr/arch/p/snes/plugin.c
```diff
@@ -12,9 +12,19 @@ struct snes_asm_flags {
 	ut8 X;
 };
 
-static R_TH_LOCAL struct snes_asm_flags snesflags = {0};
+static bool snes_init(RArchSession *s) {
+	r_return_val_if_fail (!s->user, false);
```
trufae, line 16: wrong assert

trufae (changes requested):
.

libr/arch/p/or1k/plugin.c
jmaselbas: I double check the r_return_val_if_fail macro, and I am pretty sure this will return only if `!(!s->user)`,  it asserts that `s == NULL`. I can also add `s != NULL` in the assert.

jmaselbas: ack, I didn't knew the difference, I am more used to drivers using a pointer called `priv`.

libr/arch/p/or1k/plugin.c
jmaselbas: ack

libr/arch/p/or1k/plugin.c
jmaselbas: ack

libr/arch/p/or1k/plugin.c
trufae: Yes the assert wont fail. But at some point the arch api may allow to set a value in this user pointer and that will explode. Maybe we can even remove the field if we cant find an use for it in 5.9

PR #21397, opened 2023-02-25
Use calloc in kvx arch plugin

jmaselbas:

The plugin user struct used to be in bss, thus zero initialized... Lets use calloc here, just in case.

### DIFF
--- a/libr/arch/p/kvx/plugin.c
+++ b/libr/arch/p/kvx/plugin.c
@@ -233,19 +233,18 @@ static const char *kvx_reg_profile = ""
  */
 
 static bool kvx_init(RArchSession *s) {
-	r_return_val_if_fail (!s->user, false);
-	s->user = malloc (sizeof (bundle_t));
-	return s->user? true: false;
+	r_return_val_if_fail (s && !s->data, false);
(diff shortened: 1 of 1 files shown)


libr/arch/p/kvx/plugin.c
```diff
@@ -234,7 +234,7 @@ static const char *kvx_reg_profile = ""
 
 static bool kvx_init(RArchSession *s) {
 	r_return_val_if_fail (!s->user, false);
-	s->user = malloc (sizeof (bundle_t));
+	s->user = calloc (1, sizeof (bundle_t));
```
trufae, line 237:
```suggestion
	s->user = R_NEW0 (bundle_t);
```

trufae (commented):
Im not sure if plugins should use user or data pointer here. Same for the other pr. I'll look in detail tomorrow. Also adding comments in those fields of the struct will help to clarify

trufae:
Yep. globals must die :) about analysis filling the opcode info is part of the 2nd refactoring after all those plugins are moved here. i hope to spend more time next week on this and i'll share more details about that design change

trufae:
Looks good now! 3 less globals \o/

PR #21398, opened 2023-02-25
Smallz4 support

MewtR:

**Description**
Initial support for smallz4. For now only lz4 blocks are supported, not full frames.
I tested mostly with [the nso binary from the test repo](https://github.com/radareorg/radare2-testbins/tree/master/nso) and I get the same results as with lz4. So at least it works.
Making this a draft for now because I'd like clean things up a bit and test a bit more, but this is what it could look like.
I kept all my changes to smallz4 in a separate patch file just so it's clear.
I also haven't confirmed that it works with legacy lz4 blocks yet. 

Side note: I didn't understand how to get the configure script to properly generate `config-user.mk.acr`, so I edited that file manually.


trufae:
Beautiful! ideally i would prefer to remove shlr/lz4, because that's -9000 LOC and keep the option to link against the system-wide library. Did you need to add more tests to verify those NSO files? or the ones in the testsuite are enough to check block and stream lz4 compression?

MewtR:
Really? Don't you think we should keep lz4 around at least initially 'just in case'?

For the tests it's more about testing with different binaries just to be sure we get similar results across the board. But yes I might add additional tests depending on what I discover.
By the way this PR only adds decompression support, which is all we care about right? I've only see calls to `LZ4_decompress_safe` in the code base

trufae:
Im fine to keep it optional for this upcoming release, as i dont want many moving parts. is this pr ready to review? :)

MewtR:
Give me a couple of days to make a few more changes. I'll try to have it ready before next week.

MewtR:
A little update.
I confirmed that smallz4 can decompress full lz4 frames using both the modern or legacy format.
It's ready for review, I'll keep iterating in future PRs.

(Just writing how I tested stuff in case I need to remember)
For modern lz4 frames, I just used `lz4 somefile somefile.lz4` to compress the file. Then `unlz4 somefile.lz4 > somefile-lz4` and `./smallz4cat somefile.lz4 > somefile-smallz4` to decompress the file using both tools. Then I just checked that both decompressed files had the same sha256 hash.

For legacy lz4 frames, I used `mkinitcpio` to get the compressed file with something like `sudo mkinitcpio -z lz4 -g initramfs-lz4.img`. I used `od -An --endian=little -N 4 -x initramfs-lz4.img` verify that the magic bytes were in fact the legacy ones. Then I decompressed with both tools and compared hashes.

radare:
Quick review:

* I think the code is small enough to just massage it to follow r2 coding style rules and remove the sync/patch/stuff and just consider it our own implementation, as long as you did a bunch of fixes on top of the original code (feel free to submit to the authors)
* What about meson? it seems like your pr is only affecting make builds

shlr/smallz4/smallz4cat.h
```diff
@@ -1,3 +1,3 @@
 typedef unsigned char (*GET_BYTE)  (void* userPtr);
 typedef void          (*SEND_BYTES)(const unsigned char*, unsigned int, void* userPtr);
-int unlz4Block_userPtr(GET_BYTE getByte, SEND_BYTES sendBytes, void* userPtr, unsigned int blockSize, unsigned int* position, unsigned char* hist);
+int unlz4Block_userPtr (GET_BYTE getByte, SEND_BYTES sendBytes, void *userPtr, unsigned int blockSize, unsigned int *position, unsigned char *hist);
```
trufae, line 3: Use r2 types aka ut8 and so on :) looks better

trufae: also no need for a space here, as its a function signature not a function call

radare:
lgtm to merge so we can start messsing around testing more and cleaning up code after this.

it is ok for you if i merge too?

thanks!

trufae:
ping?

trufae:
may i take your silences as a yes?

MewtR:
Hey sorry, I was a bit busy last week. I still have to make it work with meson but I'll probably only get around to it next week.

*Last online: 5 years ago*

MewtR:
But on a more serious note, I'll also address your other review points. If you would like to merge, you can of course and I'll do the rest of the work in a subsequent PR. But I do intend to do it.

trufae:
fixed and merged

PR #21426, opened 2023-03-03
ob to oi and omb to ob

Lazula:
Partial #11988

Resolves:

* Rename `ob` to `oi`
* Rename `omb` to `ob`
* Rename `objid` to `binfd` in `ob?` (now `oi?`)

### DIFF
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -4602,7 +4602,7 @@ static bool r_core_bin_file_print(RCore *core, RBinFile *bf, PJ *pj, int mode) {
 	switch (mode) {
 	case '*': {
 		char *n = __filterShell (name);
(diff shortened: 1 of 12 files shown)


radare:
Let's hold this pr for 5.9, also despite i think its a reasonable change because it reuses commands in a better way i would like to discuss about the change and explain it properly when applied because it will break more than one script

trufae:
Unfortunately this change needs to resolve the conflicts and may require more discussion to clarify those command changes to not break scripts here and there. im moving this PR for a near future

trufae:
So much conflicts and we may probably want to move slower before doing such breaking changes as it will break r2papi too.

Closing waiting for more feedback and slowly do changes that help move forward

PR #21450, opened 2023-03-08
Reduce eUsage counter

Lazula:
Get rid of a big chunk of eUsage. The only ones left are places without access to the core API, not sure what to do about those.

Also includes some help text reorganization, and fixed a bug in `tej?` where r_type_kind received a zero-length string.


libr/core/cconfig.c
```diff
@@ -1491,7 +1491,7 @@ static bool cb_dirsrc(void *user, void *data) {
 static bool cb_cfgsanbox_grain(void *user, void *data) {
 	RConfigNode *node = (RConfigNode*) data;
 	if (strstr (node->value, "?")) {
-		eprintf ("Usage: comma separated grain types to be masked out by the sandbox.\n");
+		eprintf ("comma separated grain types to be masked out by the sandbox.\n");
```
trufae, line 1494: User is an instance of core. Use rcorehelp here

trufae: Using R_LOG_INFO can be also an option in some situations but for help its better to be consistent if possible

Lazula:
There are 25 left. 2 in fs_shell.c, 9 in IO plugins, 1 in rcond.c, 1 in test_x86im.c, and 13 in syscmd.c. Maybe those should use R_LOG_INFO.

radare:
Please fix the tests and ill merge

Lazula:
The `Cf` tests look malformed so I just marked the failing one as broken. I don't think it was doing anything before. I checked it on master and it just printed out the usage text to stderr with eprintf instead of stdout with the help api, which is why the help text change made it fail.

radare:
Amazing PR! thank you!

PR #21456, opened 2023-03-09
Remove all char* casts in free

Lazula:
The 3 left in the counter are ABI breaks that will be removed in 5.9

### DIFF
--- a/libr/egg/emit_arm.c
+++ b/libr/egg/emit_arm.c
@@ -83,6 +83,7 @@ static void emit_syscall_args(REgg *egg, int nargs) {
 }
 
 static void emit_set_string(REgg *egg, const char *dstvar, const char *str, int j) {
+	char *str_escaped;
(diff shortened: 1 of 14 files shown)


libr/egg/emit_arm.c
```diff
@@ -94,8 +95,9 @@ static void emit_set_string(REgg *egg, const char *dstvar, const char *str, int
 	// XXX: does not handle \n and so on.. must use r_util
 	// use r_str_escape to handle \n
 	// do not forget mem leak
-	r_egg_printf (egg, ".string \"%s\"\n", str = r_str_escape (str));
-	free ((char *) str);
+	str_escaped = r_str_escape (str);
```
radare, line 98: you can safely define and assign in the same line, same for off and rest

radare (approved):
lgtm!

PR #21465, opened 2023-03-11
Set hit on /r research.

gogo2464:

**Description**

The command `/` already sets a flag contrary to `/r` then I put a flag on hits for `/r`. 

I am waiting for opinion as well as review.

### DIFF
--- a/libr/core/cmd_search.c
+++ b/libr/core/cmd_search.c
@@ -2192,6 +2192,9 @@ static void do_ref_search(RCore *core, ut64 addr,ut64 from, ut64 to, struct sear
 			if (from <= ref->addr && to >= ref->addr) {
 				r_cons_printf ("%s 0x%" PFMT64x " [%s] %s\n",
 						buf_fcn, ref->addr, r_anal_ref_type_tostring (ref->type), str);
+
+				r_flag_set (core->flags, "hit.prelude", ref->addr, strlen ("hit.prelude"));
+
 				if (*param->cmd_hit) {
 					ut64 here = core->offset;
 					r_core_seek (core, ref->addr, true);


libr/core/cmd_search.c
```diff
@@ -3985,7 +3988,7 @@ static int cmd_search(void *data, const char *input) {
 					}
 					if (r_cons_is_breaked ()) {
 						break;
-					}
+					}//r_flag_set (core->flags, flag, base_addr + addr, klen);
```
gogo2464, line 3991: my bad

libr/core/cmd_search.c
```diff
@@ -2192,6 +2192,9 @@ static void do_ref_search(RCore *core, ut64 addr,ut64 from, ut64 to, struct sear
 			if (from <= ref->addr && to >= ref->addr) {
 				r_cons_printf ("%s 0x%" PFMT64x " [%s] %s\n",
 						buf_fcn, ref->addr, r_anal_ref_type_tostring (ref->type), str);
+
+				r_flag_set (core->flags, "hit.prelude", ref->addr, strlen ("hit.prelude"));
```
trufae, line 2196: there are several wrong things in this line:

* why hit.prelude? /r have nothing to do with preludes, its about references
* this codepath is executed many times, one for each reference found, your pr is overwriting the same flag over and over
* why the flag size is the length of a string? thats unrelated to the reference it points to
* xrefs are stored as xrefs, thats why no flags are set when using /r

trufae: the 'r' in /r stands for reference, not for research

gogo2464:
`xa` does what I want instead of `f`. Could you close this PR please?

Sorry for the inconvenience.

Issue #21503, opened 2023-03-21
SIGSEGV at `libr/bin/p/bin_coff.c:509` in `patch_relocs()` 


echel0nn:
## Environment

```sh
Tue Mar 21 07:50:58 PM +03 2023
Linux 6.2.2-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Fri, 03 Mar 2023 15:58:29 +0000 x86_64 GNU/Linux
radare2 5.8.2 0 @ linux-x86-64 git.5.8.2
commit: unknown build: 2023-02-25__12:10:29
```

## Description
SIGSEGV at libr/bin/p/bin_coff.c:509 in patch_relocs function, found via fuzzing.

```
static bool is_imported_symbol(struct coff_symbol *s) { // <- s is 0x0
 ► 128  return s->n_scnum == COFF_SYM_SCNUM_UNDEF && s->n_sclass == COFF_SYM_CLASS_EXTERNAL;
   129 }
```
### input.bin 
```
00000000: 6603 e846 4058 6458 4036 5858 5858 5868  f..F@XdX@6XXXXXh
00000010: 5858 7063 5858 5840 0038 00de 57ff ffff  XXpcXXX@.8..W...
00000020: 7f58 5858 0600 0000 0000 0010 0038 00de  .XXX.........8..
00000030: 57ff ffff 7f58 5858 0600 0000 0000 0010  W....XXX........
00000040: 0000 00ff 7fce 0000 4000 b03c 0000 ff7f  ........@..<....
00000050: ce00 0040 00b0 3c31 4058 d5d5 d5d5 5800  ...@..<1@X....X.
00000060: 00ff 4f0f 05                             ..O..
```
```
ZgPoRkBYZFhANlhYWFhYaFhYcGNYWFhAADgA3lf///9/WFhYBgAAAAAAABAAOADeV////39YWFgGAAAAAAAAEAAAAP9/zgAAQACwPAAA/3/OAABAALA8MUBY1dXV1VgAAP9PDwU=
```

## backtrace 

```
pwndbg> backtrace
#0  patch_relocs (b=0x5555557c0be0) at ../libr/bin/p/bin_coff.c:523
#1  0x00007ffff768f2bc in r_bin_object_patch_relocs (bo=0x555555938d00, bin=0x5555557c0be0) at ../libr/bin/bobj.c:428
#2  r_bin_patch_relocs (bin=0x5555557c0be0) at ../libr/bin/bin.c:700
[...]

PR #21504, opened 2023-03-21
fix segfault when bin->symbols is NULL #21503 ##bin

echel0nn:

**Description**
extra check added before for loop if bin->symbols is not NULL in patch_relocs()


trufae:
Looks good!

PR #21508, opened 2023-03-22
Fix #21375 also for linux-arm-64.sdb syscalls with meson #build

### LINKED ISSUE #21375: syscall detection is broken in meson build
## Environment

```sh
# copypaste this script into your shell and replace it with the output
[root@3d0837c045fa /]# date
Wed Feb 22 07:10:31 UTC 2023
[root@3d0837c045fa /]# r2 -v
radare2 5.8.2 0 @ linux-x86-64 git.5.8.2
commit: unknown build: 2023-01-25__00:00:00
[root@3d0837c045fa /]# uname -ms
Linux x86_64
```
This is a fedora 37 container (on a RHEL 8 machine, probably irrelevant), using the radare2 build of fedora. At first I thought it's a packaging bug in fedora, thus reported it at https://bugzilla.redhat.com/show_bug.cgi?id=2170036 , but further investigation might imply it's a general bug in the meson build, thus current report.

## Description

syscalls detection does not work, apparently because the way the meson build generates the syscall *.sdb files is broken.

## Test

Do a meson build - e.g. something like:
```
meson build1 .
meson compile -C build1
```

Fedora's spec file is here, for reference:

https://src.fedoraproject.org/rpms/radare2/blob/rawhide/f/radare2.spec

Using such a build, run e.g.:
```
[root@3d0837c045fa /]# r2 /bin/bash
WARN: run r2 with -e bin.cache=true to fix relocations in disassembly
[0x00032ed0]> asl 77
ERROR: Unknown syscall number
[0x00032ed0]> asl ftruncate
77
```

The github linux-static build (using make) does work well on the same machine. Checking the logs of both runs:

Fedora:

https://koji.fedoraproject.org/koji/taskinfo?taskID=97322791
```
[399/1188] /builddir/build/BUILD/radare2-5.8.2/x86_64-redhat-linux-gnu/sdb libr/syscall/d/linux-x86-64.sdb == ../libr/syscall/d/linux-x86-64.sdb.txt
```

github (a recent master build picked randomly, can't find in github the action for a specific tag):

https://github.com/radareorg/radare2/actions/runs/4127735691/jobs/7131308671
```
2023-02-08T19:48:06.8914288Z "/bin/sh" gen.sh < linux-x86-64.sdb.txt | ../../..//libr/../shlr/sdb//sdb linux-x86-64.sdb =
```

The make/github build calls `gen.sh` on the text file, which "duplicates" it - adds the "opposite" direction, AFAIU - the text file stores "ftruncate=0x80,77,2," and gen.sh adds "0x80.77=ftruncate", for the above example. The meson build does not.

A somewhat shallow further checking of the git log hints that this might have been a regression which was part of https://github.com/radareorg/radare2/commit/657524aabca350db00f3bfdbf46afa587254c5b9. If so, perhaps parts of this should be reverted.

Thanks!

didib:
Specifically, see https://github.com/radareorg/radare2/issues/21375#issuecomment-1445926582


didib:
9804c9b was missing a newline.

Verified a162e0f by running
```
python3 gen.py ../../../shlr/sdb/src/sdb linux-arm-64.sdb.txt linux-arm-64.sdb
```
and comparing the result to the one generated using gen.sh - it's identical.

trufae:
thank you!

didib:
Thanks you for such a quick turnaround!

PR #21520, opened 2023-03-25
Make main function detection for 64bit PE Arm64 ##bin

kohnakagawa:

**Description**

radare2 cannot find main function of 64bit PE Arm64. This pull request adds main function detection for fixing this issue.

You can test this feature with the attached PE binaries.

[test.zip](https://github.com/radareorg/radare2/files/11068507/test.zip)


trufae:
Merging as long as it looks good. But will be good to include this binary in the testsuite

PR #21527, opened 2023-03-27
Migrate bf plugin

luc-tielen:

**Description**

### DIFF
--- a/libr/arch/p/bf/bf.c
+++ b/libr/arch/p/bf/bf.c
@@ -276,6 +276,10 @@ static bool encode(RArchSession *as, RAnalOp *op, RArchEncodeMask mask) {
 	return size > 0;
 }
 
+static int archinfo(RArchSession *as, ut32 q) {
+	return 1;
+}
+
 RArchPlugin r_arch_plugin_bf = {
 	.name = "bf",
 	.desc = "brainfuck code analysis plugin",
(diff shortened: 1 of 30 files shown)


libr/arch/p/bf/bf.c
```diff
@@ -4,10 +4,10 @@
 #include <r_types.h>
```
trufae, line 4: probably no need to include those, as r_arch includes them too

libr/arch/p/bf/bf.c
```diff
@@ -178,12 +196,12 @@ static int bf_op(RAnal *anal, RAnalOp *op, ut64 addr, const ut8 *_buf, int len,
...
 					goto beach;
 				}
-				if (i == len - 1 && anal->read_at) {
+				if (i == len - 1) {
 					int new_buf_len = len + 1 + BUFSIZE_INC;
 					ut8 *new_buf = calloc (new_buf_len, 1);
 					if (new_buf) {
 						free (buf);
-						(void)anal->read_at (anal, addr, new_buf, new_buf_len);
+						(void)read_at (bin->iob.io, addr, new_buf, new_buf_len);
```
trufae, line 204: ideally rarch shouldnt be directly accessing io, and should use only the data passed in op->bytes

PR #21557, opened 2023-04-03
Optimize macho parser

luc-tielen:

**Description**

I refactored some parts of the mach0 parser:
- Some results are now cached after first time they are needed, subsequent function calls return the cached result. Further optimizations can be made here if e.g. in `bobj.c` the free function of a `RList` is not hardcoded to `r_bin_import_free` (😱)
- Reduced memory usage by making clever usage of data-structures
- Removed the ".last hack" for the refactored parts.

**PR is still WIP**, I still want to refactor symbols and relocations list.

### DIFF
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -2179,7 +2179,7 @@ void *MACH0_(mach0_free)(struct MACH0_(obj_t) *mo) {
 	free (mo->compiler);
 #if FEATURE_SYMLIST
 	if (mo->symbols_loaded) {
-		r_pvector_fini (&mo->symbols_cache);
+		r_list_purge (&mo->symbols_cache);
 	}
 #endif
 	r_list_free (mo->sections_cache);
@@ -2864,7 +2864,7 @@ typedef struct fill_context_t {
 	ut32 *ordinal;
 } FillCtx;
 
-static void _fill_exports_pvector(struct MACH0_(obj_t) *bin, const char *name, ut64 flags, ut64 offset, void *ctx) {
+static void _fill_exports(struct MACH0_(obj_t) *bin, const char *name, ut64 flags, ut64 offset, void *ctx) {
 	FillCtx *context = ctx;
 	ut64 vaddr = offset_to_vaddr (bin, offset);
 	if (hash_find_or_insert (context->hash, name, vaddr)) {
@@ -2882,7 +2882,7 @@ static void _fill_exports_pvector(struct MACH0_(obj_t) *bin, const char *name, u
 	sym->bind = R_BIN_BIND_GLOBAL_STR;
 	sym->ordinal = (*context->ordinal)++;
 	_enrich_symbol (context->bf, bin, context->symcache, sym);
-	r_pvector_push (&bin->symbols_cache, sym);
+	r_list_append (&bin->symbols_cache, sym);
 }
 
 static void _parse_symbols(RBinFile *bf, struct MACH0_(obj_t) *bin, Sdb *symcache) {
@@ -2902,7 +2902,7 @@ static void _parse_symbols(RBinFile *bf, struct MACH0_(obj_t) *bin, Sdb *symcach
 	}
 
 	FillCtx fill_context = { .bf = bf, .symcache = symcache, .hash = hash, .boffset = obj->boffset, .ordinal = &ordinal };
-	walk_exports (bin, _fill_exports_pvector, &fill_context);
+	walk_exports (bin, _fill_exports, &fill_context);
 
 	if (!bin->symtab || !bin->symstr) {
 		ht_pp_free (hash);
@@ -2988,7 +2988,7 @@ static void _parse_symbols(RBinFile *bf, struct MACH0_(obj_t) *bin, Sdb *symcach
 			} else {
 				_enrich_symbol (bf, bin, symcache, sym);
 				sym->ordinal = ordinal++;
-				r_pvector_push (&bin->symbols_cache, sym);
+				r_list_append (&bin->symbols_cache, sym);
 			}
 		}
 	}
@@ -3017,7 +3017,7 @@ static void _parse_symbols(RBinFile *bf, struct MACH0_(obj_t) *bin, Sdb *symcach
 			sym->is_imported = symbol.is_imported;
 			sym->ordinal = ordinal++;
 			_enrich_symbol (bf, bin, symcache, sym);
-			r_pvector_push (&bin->symbols_cache, sym);
+			r_list_append (&bin->symbols_cache, sym);
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


libr/bin/p/bin_mach0.c
```diff
@@ -387,49 +385,21 @@ static RBinImport *import_from_name(RBin *rbin, const char *orig_name, HtPP *imp
...
-		if (!strcmp (name, "_NSConcreteGlobalBlock")) {
-			bin->has_blocks_ext = true;
-		}
-		r_list_append (ret, ptr);
+
+	RList *list = r_list_newf ((RListFree) r_bin_import_free);
+	void **it;
+	r_pvector_foreach (imports, it) {
+		// need to clone here, in bobj.c the list free function is forced to `r_bin_import_free`
+		// otherwise, a list with no free function could be returned here..
```
luc-tielen, line 398: This would also reuse the memory of the vector, which is sequential in memory => better performance

test/db/formats/mach0/mach0
```diff
@@ -267,8 +267,8 @@ CMDS=ik ***~uuid
 EXPECT=<<EOF
 info/mach0_uuid_command.format=[4]Ed[16]b (mach0_load_command_type)cmd cmdsize uuid
 info/mach0_cmd_8.format=mach0_uuid_command
-info/uuid.0=ff4d6c2e1dc0369d98d0d3040a8049e8
 info/mach0_cmd_8.cmd=uuid
+info/uuid.0=ff4d6c2e1dc0369d98d0d3040a8049e8
```
luc-tielen, line 271: So this is technically a change in functionality, but I think it makes sense (`info/mach0_cmd_8` lines are grouped together now).
The change is probably due to the manual memory management for `libs` that is now replaced by a `r_pvector`.

libr/bin/p/bin_mach0.c
```diff
@@ -358,16 +358,14 @@ static RBinImport *import_from_name(RBin *rbin, const char *orig_name, HtPP *imp
 
 	char *name = (char*) orig_name;
 	const char *_objc_class = "_OBJC_CLASS_$";
-	const int _objc_class_len = strlen (_objc_class);
 	const char *_objc_metaclass = "_OBJC_METACLASS_$";
-	const int _objc_metaclass_len = strlen (_objc_metaclass);
 	const char * type = "FUNC";
```
trufae, line 362: const char * const

libr/bin/format/mach0/mach0.c
```diff
@@ -3250,42 +3288,68 @@ struct import_t *MACH0_(get_imports)(struct MACH0_(obj_t) *bin) {
...
+		if (!import) {
+			free (imp_name);
+			break;  // TODO change into continue?
+		}
+
+		import->ordinal = i;
+		r_pvector_push (&bin->imports_cache, import);
+
+		num_imports++;
+		if (!strcmp (import->name, "__stack_chk_fail") ) {
```
trufae, line 3322: Move into separate function for readability

libr/bin/p/bin_mach0.c
trufae: Yes but this would break abi, and maybe it will be better to cache the list, but i dont think thats really a bottleneck. Another option is to just use rlist internally and just return the pointer, returned list shouldnt be freed/owned

libr/bin/format/mach0/mach0.c
```diff
@@ -4001,25 +4004,30 @@ void MACH0_(kv_loadlibs)(struct MACH0_(obj_t) *bin) {
...
-	if (!(libs = calloc ((bin->nlibs + 1), sizeof (struct lib_t)))) {
-		return NULL;
-	}
-	for (i = 0; i < bin->nlibs; i++) {
+
+	r_pvector_init (&bin->libs_cache, (RPVectorFree) free);
+	r_pvector_reserve (&bin->libs_cache, bin->nlibs);
+
+	char lib_flagname[128];
+	for (int i = 0; i < bin->nlibs; i++) {
```
trufae: Define "int i" before the loop

libr/bin/format/mach0/mach0.c
```diff
@@ -1199,23 +1199,22 @@ static int parse_function_starts(struct MACH0_(obj_t) *bin, ut64 off) {
...
 
 	if (off > bin->size || off + sizeof (struct dylib_command) > bin->size) {
 		return false;
 	}
-	lib = bin->nlibs - 1;
 
-	void *relibs = realloc (bin->libs, bin->nlibs * R_BIN_MACH0_STRING_LENGTH);
-	if (!relibs) {
-		r_sys_perror ("realloc (libs)");
+	char *lib = calloc (1, R_BIN_MACH0_STRING_LENGTH);
```
radare, line 1209: better to hold this thing in the stack. its just 256 bytes. `char lib[256]` and then use strdup() when appending it to the list/vector. otherwise we are wasting memory and doing unnecessary allocations

libr/bin/format/mach0/mach0.c
```diff
@@ -1199,23 +1199,22 @@ static int parse_function_starts(struct MACH0_(obj_t) *bin, ut64 off) {
...
+	if (!lib) {
+		r_sys_perror ("calloc (lib)");
```
radare, line 1211: no need to perror on simple/smol allocations you can remove this too

libr/bin/format/mach0/mach0.c
```diff
@@ -1227,17 +1226,19 @@ static int parse_dylib(struct MACH0_(obj_t) *bin, ut64 off) {
...
-		(ut8*)bin->libs[lib], R_BIN_MACH0_STRING_LENGTH - 1);
-	bin->libs[lib][R_BIN_MACH0_STRING_LENGTH - 1] = 0;
+		(ut8*) lib, R_BIN_MACH0_STRING_LENGTH - 1);
 	if (len < 1) {
 		bprintf ("Error: read (dylib str)");
+		free (lib);
 		return false;
 	}
+
+	r_pvector_push (&bin->libs_cache, lib);
```
radare, line 1241: even if vectors are better, i dont think an RList is really a big issue, and having rlist internally also avoids us to rewrite the rest of the callers, so maybe just having the RList internally and not freed later is enough at least until we decide to break abi, but this abi break will require not just rewriting all the callers, but also all the plugins, and thats imho a little out of the scope for this optimization

libr/bin/p/bin_mach0.c
luc-tielen: I don't see what would break it? The hack in `bobj.c` needs to be removed, and a free function needs to be added to the returned lists of other file formats (where needed).

Once we have this, we can just use `r_list_new`, with pointers pointing into the vector. Then `r_list_free` would do no freeing except for the list pointer itself, and the vector would manage the other memory.

libr/bin/format/mach0/mach0.c
luc-tielen: Good point.

libr/bin/format/mach0/mach0.c
luc-tielen: I've seen massive differences in the past in other codebases when switching from list to vector, which is why I also did it here.

We can't fully cache the lists though, because free is overridden in `bobj.c`, leading to double frees..

luc-tielen:
Closing this because I split it up into multiple PRs.

PR #21580, opened 2023-04-11
Refactor macho imports

luc-tielen:

**Description**

### DIFF
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -2179,6 +2179,9 @@ void *MACH0_(mach0_free)(struct MACH0_(obj_t) *mo) {
 	free (mo->compiler);
 	r_list_free (mo->symbols_cache);
 	r_list_free (mo->sections_cache);
+	if (mo->imports_loaded) {
+		r_pvector_fini (&mo->imports_cache);
+	}
 	if (mo->chained_starts) {
 		for (i = 0; i < mo->nsegs && i < mo->segs_count; i++) {
 			if (mo->chained_starts[i]) {
@@ -2593,7 +2596,7 @@ static bool parse_import_stub(struct MACH0_(obj_t) *bin, struct symbol_t *symbol
 	return false;
 }
 
-static int inSymtab(HtPP *hash, const char *name, ut64 addr) {
+static int hash_find_or_insert(HtPP *hash, const char *name, ut64 addr) {
 	bool found = false;
 	char *key = r_str_newf ("%"PFMT64x".%s", addr, name);
 	ht_pp_find (hash, key, &found);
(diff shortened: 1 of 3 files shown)


libr/bin/format/mach0/mach0.c
```diff
@@ -3263,10 +3266,45 @@ static int parse_import_ptr(struct MACH0_(obj_t) *bin, struct reloc_t *reloc, in
...
 
-struct import_t *MACH0_(get_imports)(struct MACH0_(obj_t) *bin) {
+static RBinImport *import_from_name(RBin *rbin, const char *orig_name) {
+	RBinImport *ptr = R_NEW0 (RBinImport);
+	if (!ptr) {
+		return NULL;
+	}
+
+	char *name = (char*) orig_name;
+	const char *_objc_class = "_OBJC_CLASS_$";
```
trufae, line 3276: this should be const char * const, otherwise the code below cannot be optimized because the string can be modified at any time. but imho it will look better to use defines or just inline the string in the startswith + strlen like its done for the `type` line

libr/bin/p/bin_mach0.c
```diff
@@ -358,16 +358,14 @@ static RBinImport *import_from_name(RBin *rbin, const char *orig_name, HtPP *imp
...
 	char *name = (char*) orig_name;
 	const char *_objc_class = "_OBJC_CLASS_$";
-	const int _objc_class_len = strlen (_objc_class);
 	const char *_objc_metaclass = "_OBJC_METACLASS_$";
-	const int _objc_metaclass_len = strlen (_objc_metaclass);
 	const char * type = "FUNC";
 
-	if (!strncmp (name, _objc_class, _objc_class_len)) {
-		name += _objc_class_len;
+	if (r_str_startswith (name, _objc_class)) {
```
trufae, line 364: uh, this seems dupe  :D so do the same as in mach0.c 3280

libr/bin/format/mach0/mach0.c
```diff
@@ -3263,10 +3266,45 @@ static int parse_import_ptr(struct MACH0_(obj_t) *bin, struct reloc_t *reloc, in
...
+
+const RPVector *MACH0_(load_imports)(RBinFile *bf, struct MACH0_(obj_t) *bin) {
 	r_return_val_if_fail (bin, NULL);
+	if (bin->imports_loaded) {
+		return &bin->imports_cache;
+	}
+
+	bin->imports_loaded = true;
+	r_pvector_init (&bin->imports_cache, (RPVectorFree) r_bin_import_free);
+	r_pvector_reserve (&bin->imports_cache, bin->dysymtab.nundefsym);
```
trufae, line 3306: use a local variable like `int amount = bin->dysymtab.nundefsym` . so you can check if the value is invalid (0 will cause invalid allocations and large value can be causing negative operations or invalid allocation because not enough memory if its too large).

having the var in local instead of reusing bin->dysymtab.nundefsym:
* allow to dim/trim to a decent value
* detach from the bin header information
* faster access because this is used in the for loop and you dont need to dereference pointers to know the value.

trufae (changes requested):

libr/bin/p/bin_mach0.c
luc-tielen: Yeah, that code is used by `patch_relocs`, which I haven't touched yet. Will probably also be moved over to the other file once we cache all that..

PR #21587, opened 2023-04-11
Refactor macho libs

luc-tielen:

**Description**

### DIFF
--- a/libr/bin/format/mach0/mach0.c
+++ b/libr/bin/format/mach0/mach0.c
@@ -2182,6 +2182,9 @@ void *MACH0_(mach0_free)(struct MACH0_(obj_t) *mo) {
 	if (mo->imports_loaded) {
 		r_pvector_fini (&mo->imports_cache);
 	}
+	if (mo->libs_loaded) {
+		r_pvector_fini (&mo->libs_cache);
+	}
 	if (mo->chained_starts) {
 		for (i = 0; i < mo->nsegs && i < mo->segs_count; i++) {
 			if (mo->chained_starts[i]) {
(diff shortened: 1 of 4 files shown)


libr/bin/format/mach0/mach0.c
```diff
@@ -3295,6 +3295,20 @@ static RBinImport *import_from_name(RBin *rbin, const char *orig_name) {
 	return ptr;
 }
 
+static void check_for_special_import_names(struct MACH0_(obj_t) *bin, RBinImport *import) {
+	if (import->name[0] == '_') {
```
trufae, line 3299:
```suggestion
	const char *name = import->name;
	if (*name == '_') {
```

libr/bin/format/mach0/mach0.c
```diff
@@ -3295,6 +3295,20 @@ static RBinImport *import_from_name(RBin *rbin, const char *orig_name) {
...
+		if (!strcmp (import->name, "__stack_chk_fail") ) {
```
trufae, line 3300:
```suggestion
		if (!strcmp (name, "__stack_chk_fail") ) {
```

libr/bin/format/mach0/mach0.c
```diff
@@ -3295,6 +3295,20 @@ static RBinImport *import_from_name(RBin *rbin, const char *orig_name) {
...
+			bin->has_canary = true;
+		}
+		if (!strcmp (import->name, "__asan_init") || !strcmp (import->name, "__tsan_init")) {
```
trufae, line 3303:
```suggestion
		if (!strcmp (name, "__asan_init") || !strcmp (name, "__tsan_init")) {
```

libr/bin/format/mach0/mach0.c
```diff
@@ -3295,6 +3295,20 @@ static RBinImport *import_from_name(RBin *rbin, const char *orig_name) {
...
+			bin->has_sanitizers = true;
+		}
+		if (!strcmp (import->name, "_NSConcreteGlobalBlock")) {
```
trufae, line 3306:
```suggestion
		if (!strcmp (name, "_NSConcreteGlobalBlock")) {
```

libr/bin/p/bin_mach0.c
```diff
@@ -451,23 +451,22 @@ static RList *relocs(RBinFile *bf) {
...
 	}
-	if ((libs = MACH0_(get_libs) (obj->bin_obj))) {
-		for (i = 0; !libs[i].last; i++) {
-			ptr = strdup (libs[i].name);
-			r_list_append (ret, ptr);
-		}
-		free (libs);
+
+	RList *result = r_list_new ();
+	void **it;
```
trufae, line 465: will be good to make r_pvector api foreach() work like the r_vector_foreach one. for consistency, also maybe we can have an r_pvector_tolist() function that does exactly this, because we will need to duplicate this code in a bunch of places if we want to use vector/pvector internally, but keep the public abi with rlist

trufae (changes requested):
looks good! also, will be good to compare pvector and vector performance, maybe writing some unit benchmarks in test/unit with at least a time() with the same operations and different growing mechanics.

libr/bin/p/bin_mach0.c
luc-tielen: Yeah that's a good point, I didn't even realize the foreach worked so differently for the 2 variations. 😅

Sure I can make a "tolist" function, could be useful in other places too!

PR #21650, opened 2023-04-25
Add - to tar xzvf ##r2pm

CorruptedVor:

**Description**

This breaks the tar I use, seems this "dashless" syntax is GNU-specific. This change should result in no breakage, and just improve support for various tar implementations.


trufae:
Looks good!

PR #21681, opened 2023-05-02
Fix/various memory leaks

luc-tielen:

**Description**


libr/bin/format/elf/elf.c
```diff
@@ -4520,6 +4520,10 @@ void Elf_(r_bin_elf_free)(ELFOBJ* bin) {
 	ht_up_free (bin->rel_cache);
 	bin->rel_cache = NULL;
 	sdb_free (bin->kv);
+	if (bin->inits) {
```
trufae, line 4523: free(NULL) is always valid, so no need for an if

trufae (changes requested):
.

PR #21702, opened 2023-05-04
Migrate mips gnu

luc-tielen:

**Description**


libr/anal/meson.build
```diff
@@ -65,7 +65,7 @@ r_anal_sources = [
   '../asm/arch/mips/gnu/mips16-opc.c',
   '../asm/arch/mips/gnu/micromips-opc.c',
   '../asm/arch/mips/gnu/mips-opc.c',
-  'p/anal_mips_gnu.c',
+  '../arch/p/mips_gnu/plugin.c',
```
trufae, line 68: Uh? Is anal still using this file?

trufae: if its not used that will result in duplicated compilation or unnecessarily linking or dupped symbols . sobetter remove them all . actually when we move all the plugins that list should be empty

PR #21707, opened 2023-05-04
Fix #21705 - check r_io_plugin_add return value and freed unused allocated memory ##io 

### LINKED ISSUE #21705: Memory leak in r_io_plugin_init
The r_io_plugin_init function in the libr/io/io_plugin.c source file appears to contain a memory leak. Specifically, memory is allocated using R_NEW0 macro in line 30:

static_plugin = R_NEW0 (RIOPlugin); // line 30

If r_io_plugin_add (io, static_plugin);(line 35) fails to add the allocated memory object to the io, the function will return false. In such a scenario, the allocated memory should be cleaned.

Hripsimee:

**Description**


trufae:
Awesome! thanks for contributing! can you submit a PR for the other two tickets you filled? otherwise i'll find some time when i have some spare time

PR #21708, opened 2023-05-04
Fix #17523 - Resolve internal symbols referenced from the PLT ##bin

### LINKED ISSUE #17523: PLT stub names not being resolved properly
### Work environment

| OS/arch/bits (mandatory)                             | Arch x86_64
| Architecture/bits of the file (mandatory)            | arm 32
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.6.0-git 25551 @ linux-x86-64 git.4.4.0-580-g7066f58d8
commit: 7066f58d8c7390cc118af641ae8ee0d0f00ed228 build: 2020-08-25__09:23:38

### Expected behavior

When a function exported by a shared object file is called from within that shared object file, the function name for the stub in the `plt` section should be named properly.

Example: in `libcurl`, the function `Curl_getaddrinfo_ex` is called from the function `Curl_ipv4_resolve_r`. That function call should appear as `sym.imp.Curl_getaddrinfo_ex`.

### Actual behavior

Instead of `sym.imp.Curl_getaddrinfo_ex`, the function is flagged `fcn.000074fc`. This makes analysis a bit painful. 

The address that the `plt` stub points to is correctly identified as `reloc.Curl_getaddrinfo_ex`, so the relocation is being correctly ID'd (I've also double-checked with Ghidra). 

Asciinema: https://asciinema.org/a/o4UkBFPskH5rhJZ8lwjuNTKJC

Representative file here: [libcurl.zip](https://github.com/radareorg/radare2/files/5124497/libcurl.zip)

Here is a screenshot from ghidra showing the same address (though, interestingly, it's mapped at 0x174fc instead of 0x74fc):
ghidra_curl_reloc_example

trufae:

**Description**

### DIFF
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -2081,7 +2081,6 @@ ut64 Elf_(r_bin_elf_get_main_offset)(ELFOBJ *bin) {
 		/* non-thumb entry points */
 		if (!memcmp (buf, "\x00\xb0\xa0\xe3\x00\xe0\xa0\xe3", 8)) {
 			if (buf[0x40 + 2] == 0xff && buf[0x40 + 3] == 0xeb) {
-				// nothing may happen
 			} else if (!memcmp (buf + 0x28 + 2, "\xff\xeb", 2)) {
 				return Elf_(r_bin_elf_v2p) (bin, r_read_le32 (&buf[0x34]) & ~1);
 			}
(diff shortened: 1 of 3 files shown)


libr/bin/format/elf/elf.c
```diff
@@ -4011,6 +4010,7 @@ RBinSymbol *Elf_(_r_bin_elf_convert_symbol)(struct Elf_(r_bin_elf_obj_t) *bin, s
 		ptr->bind = symbol->bind;
 		ptr->type = symbol->type;
 		ptr->is_imported = symbol->is_imported;
+		// ptr->is_internal = symbol->is_internal;
```
luc-tielen, line 4013: Don't forget to remove this line.

PR #21734, opened 2023-05-10
Add support for multiple r2pm package repos

condret:

**Description**

### DIFF
--- a/libr/main/r2pm.c
+++ b/libr/main/r2pm.c
@@ -150,6 +150,7 @@ static char *r2pm_pkgdir(void) {
 
 typedef enum {
 	TT_TEXTLINE,
+	TT_TEXTLINE_LIST,
 	TT_CODEBLOCK,
 	TT_ENDQUOTE,
 } R2pmTokenType;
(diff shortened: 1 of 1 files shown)


libr/main/r2pm.c
```diff
@@ -600,17 +600,24 @@ static int r2pm_clone(const char *pkg) {
...
-		char *url = r2pm_get (pkg, "\nR2PM_GIT ", TT_TEXTLINE);
-		if (url) {
-			char *dir = strchr (url, ' ');
-			if (dir) {
-				*dir++ = 0;
-				if (strcmp (dir, pkg)) {
-					R_LOG_WARN ("pkgname != clonedir");
+		char *urls = r2pm_get (pkg, "\nR2PM_GIT ", TT_TEXTLINE);
+		if (urls) {
+			int url_ct = r_str_split (urls, ',');
```
trufae, line 605: what about using ':' instead ? like in env vars? the proposal lgtm

trufae: iwas also thinking that || can be another good option because its not really a list, but a fallback queue

condret: I think , is a great character, because it cannot be used in urls

PR #21767, opened 2023-05-23
Make RLangPlugin.init() actually define if can be instantiated ##lang

radare:

**Description**

### DIFF
--- a/libr/lang/p/qjs.c
+++ b/libr/lang/p/qjs.c
@@ -435,6 +435,11 @@ static bool lang_quickjs_file(RLangSession *s, const char *file) {
 }
 
 static void *init(RLangSession *ls) {
+	if (ls == NULL) {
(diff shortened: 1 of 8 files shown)


libr/lang/p/qjs.c
```diff
@@ -435,6 +435,11 @@ static bool lang_quickjs_file(RLangSession *s, const char *file) {
 }
 
 static void *init(RLangSession *ls) {
+	if (ls == NULL) {
+		// XXX should return bool instead
+		return (void*)1;
+	}
+	r_return_val_if_fail (ls, NULL);
```
luc-tielen, line 442: Isn't this redundant? And conflicting with the if above?

radare: yeah leftover!

PR #21811, opened 2023-05-31
Fix reading phdr for elf ##bin

luc-tielen:

**Description**


libr/bin/format/elf/elf.c
```diff
@@ -3728,10 +3728,7 @@ static void _add_ehdr_section (RBinFile *bf, ELFOBJ *eo) {
 	ptr->vaddr = eo->baddr;
 	ptr->size = ehdr_size;
 	ptr->vsize = ehdr_size;
-	ptr->add = false;
```
trufae, line 3731: uhm i think this is wrong because it will make relocs to fail

trufae:
the only broken tests are the core ones.. 🤔

libr/bin/format/elf/elf.c
luc-tielen: How so? Look at the other lines below as well, I refactored the boolean assignment.

PR #21815, opened 2023-05-31
Fix #21813 - parsing reg profiles from android-gdbserver ##debug

### LINKED ISSUE #21813: r2 fails to parse /proc/maps from the Android device via gdb remote protocol
## Environment

```sh
Wed May 31 19:11:45 CEST 2023
radare2 5.8.6 0 @ darwin-arm-64
birth: git.5.8.6 2023-05-05__15:36:00
commit: 5.8.6
options: gpl release -O1 checks=2
Darwin arm64
```

## Description

When I connect to a `gdbserver` running on an Android Arm64 phone I get the following:

```
r_debug_gdb_map_get: Unable to parse "/proc/27546/maps"
Content:
12c00000-12cc0000 rw-p 00000000 00:00 0                                  [anon:dalvik-main space (region space)]
```

It seems to fails to parse the /proc/maps for the target process. I checked the r2 code and seems to expect something more of a filename in last place. Would it be possible to ignore the regions that don't parse or fix up their name to be able to proceed with the debugging? At this point the parsing function aborts the debugger.

## Test

I run the gdbserver on the phone like this:

```sh
gdbserver :1234 --attach $(pidof my.app.package.id)
```

I forward the the port to the host with `adb`:

```
adb forward tcp:1234 tcp:1234
```

And I simply launch `r2` and run the following command there and the following output:

```
[0x0000c500]> doof gdb://localhost:1234

WARN: Duplicated register definition for 'SP' has been ignored
r_debug_gdb_map_get: Unable to parse "/proc/28834/maps"
Content:
12c00000-12c80000 rw-p 00000000 00:00 0                                  [anon:dalvik-main space (region space)]
r_debug_gdb_map_get: Unable to parse "/proc/28834/maps"
Content:
12c00000-12c80000 rw-p 00000000 00:00 0                                  [anon:dalvik-main space (region space)]
```

No debugging is possible after this.

trufae:

**Description**

### DIFF
--- a/libr/debug/p/debug_gdb.c
+++ b/libr/debug/p/debug_gdb.c
@@ -156,7 +156,6 @@ static RList *r_debug_gdb_map_get(RDebug* dbg) { //TODO
 	char *ptr, *pos_1;
 	size_t line_len;
 	char name[1024], region1[100], region2[100], perms[5];
-	RDebugMap *map = NULL;
 	region1[0] = region2[0] = '0';
 	region1[1] = region2[1] = 'x';
 	char *save_ptr = NULL;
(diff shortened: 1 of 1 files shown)


trufae:
That shouldnt hurt, so lets merge so you can test on master

PR #21827, opened 2023-06-02
Add missing r_cons_canvas_goto_write ##cons

condret:

**Description**

### DIFF
--- a/libr/cons/canvas.c
+++ b/libr/cons/canvas.c
@@ -351,7 +351,7 @@ R_API void r_cons_canvas_write(RConsCanvas *c, const char *_s) {
 	free (os);
 }
 
-R_API void r_cons_canvas_goto_write(RConsCanvas *c, int x, int y, const char *s) {
+R_API void r_cons_canvas_write_at(RConsCanvas *c, const char *s, int x, int y) {
 	r_cons_canvas_gotoxy (c, x, y);
 	r_cons_canvas_write (c, s);
 }
--- a/libr/cons/canvas.c
+++ b/libr/cons/canvas.c
@@ -351,6 +351,11 @@ R_API void r_cons_canvas_write(RConsCanvas *c, const char *_s) {
 	free (os);
 }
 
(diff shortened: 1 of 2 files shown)


trufae:
This is an abi breaking change so will need to wait 1 week to merge. But i think we can find a better name for it. Guess the line in the .h was a leftover from a cleanup of unused functions

trufae:
If we rename this api to `r_cons_canvas_write_at` and we use it somewhere then im fine to merge. otherwise it feels like this can be replaced with a `#define W(x,y,s)`..

libr/cons/canvas.c
```diff
@@ -351,6 +351,11 @@ R_API void r_cons_canvas_write(RConsCanvas *c, const char *_s) {
 	free (os);
 }
 
+R_API void r_cons_canvas_goto_write(RConsCanvas *c, int x, int y, const char *s) {
```
trufae, line 354:
```suggestion
R_API void r_cons_canvas_write_at(RConsCanvas *c, const char *s, int x, int y) {
```

libr/include/r_cons.h
```diff
@@ -772,7 +772,7 @@ R_API void r_cons_canvas_attr(RConsCanvas *c,const char *attr);
 R_API void r_cons_canvas_write(RConsCanvas *c, const char *_s);
 R_API void r_cons_canvas_background(RConsCanvas *c, const char *color);
 R_API bool r_cons_canvas_gotoxy(RConsCanvas *c, int x, int y);
-R_API void r_cons_canvas_goto_write(RConsCanvas *c,int x,int y, const char *s);
+R_API void r_cons_canvas_goto_write(RConsCanvas *c, int x, int y, const char *s);
```
trufae, line 775:
```suggestion
R_API void r_cons_canvas_write_at(RConsCanvas *c, const char *s, int x, int y);
```

trufae:
i just did the change @condret is this ok for u? i can merge it before the 5.9 rls

PR #21837, opened 2023-06-05
Refactor some more elf code

luc-tielen:

**Description**

### DIFF
--- a/libr/bin/p/bin_elf.inc.c
+++ b/libr/bin/p/bin_elf.inc.c
@@ -180,7 +180,7 @@ static void process_constructors(RBinFile *bf, RList *ret, int bits) {
 	RListIter *iter;
 	RBinSection *sec;
 	r_list_foreach (secs, iter, sec) {
-		if (sec->size > 0xffffff) {
+		if (sec->size > ALLOC_SIZE_LIMIT) {
 			continue;
 		}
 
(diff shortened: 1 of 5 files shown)


libr/bin/p/bin_elf.inc.c
```diff
@@ -195,15 +198,14 @@ static void process_constructors(RBinFile *bf, RList *ret, int bits) {
...
 		}
-		if (sec->size > 0xffffff) {
-			continue;
-		}
 		ut8 *buf = calloc (sec->size, 1);
 		if (!buf) {
 			continue;
 		}
+
 		(void)r_buf_read_at (bf->buf, sec->paddr, buf, sec->size);
```
trufae, line 206: check return to realloc the buffer and maybe add a warning if that section->size needs to be trimmed to fit the reality

libr/bin/p/bin_elf.inc.c
```diff
@@ -195,15 +198,14 @@ static void process_constructors(RBinFile *bf, RList *ret, int bits) {
 		if (type == -1) {
 			continue;
 		}
-		if (sec->size > 0xffffff) {
```
trufae, line 198: this is done to avoid huge allocations. which is an OOM crash in the fuzzing loop

luc-tielen: I figured, but I only moved it up to save some work if a section is too large

libr/bin/p/bin_elf.inc.c
luc-tielen: can you be more clear what needs to happen here?
```
int size = r_buf_read_at(...)
if (size != sec->size) {
        if (size < sec->size) { R_LOG_WARN ("unexpected section size"); }
        realloc (buf, size);
        sec->size = size;
}
```
something like this?

libr/bin/p/bin_elf.inc.c
```diff
@@ -179,10 +179,13 @@ static void process_constructors(RBinFile *bf, RList *ret, int bits) {
 	RList *secs = sections (bf);
 	RListIter *iter;
 	RBinSection *sec;
-	int i, type;
 	r_list_foreach (secs, iter, sec) {
-		type = -1;
+		if (sec->size > 0xffffff) {
```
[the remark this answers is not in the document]
luc-tielen: good point!

libr/bin/p/bin_elf.inc.c
trufae: Yep like this

PR #21841, opened 2023-06-05
Vec re-implementation

luc-tielen:

**Description**


trufae:
Looks good!

trufae (changes requested):
those comments are probably deprecated so feel free to ignore them

shlr/sdb/include/sdb/r_vec.h
```diff
@@ -0,0 +1,344 @@
+#ifndef R2_VEC_H
+#define R2_VEC_H
+
+#include <r_util/r_assert.h>
+#include <r_types.h>
+#include <stdlib.h>
```
trufae, line 6: i think stdlib and stdbool are already included in r_types.

libr/include/r_vec.h
```diff
@@ -0,0 +1,224 @@
...
+	} \
+	static inline void R_VEC_FUNC(type, shrink_to_fit)(R_VEC(type) *vec) { \
+		r_return_if_fail (vec); \
+		const size_t num_elems = R_VEC_FUNC(type, length) (vec); \
+		if (num_elems == R_VEC_FUNC(type, capacity) (vec)) { \
+			return; \
+		} \
+		type *buf = realloc (vec->start, num_elems * sizeof (type)); \
+		if (R_UNLIKELY (!buf)) { \
+			return; \
```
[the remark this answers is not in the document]
luc-tielen: I thought about this but I think this check is better here, realloc could sometimes fail if we hit OOM error.. but probably we have bigger problems then.

shlr/sdb/include/sdb/r_vec.h
luc-tielen: Ah nice.

PR #21868, opened 2023-06-09
Introduce RIORegion as an alternative to r_io_is_valid_offset

condret:
There are situations in which anal performs consecutive calls to r_io_is_valid_offset. each of these call causes a search on the submap tree of the active iobank. RIORegion provides a different approach to this, which can be used in anal for optimization.

**Description**


trufae:
Can you use it somewhere in the code to see how this new api can improve the perf?

condret:
sure, I've also made some benchmark. first pic is before https://github.com/radareorg/radare2/pull/21868/commits/c9c339bab177d78351a14251693530615e2955b3

r_io_is_valid_offset
ksnip_20230610-030055

r_io_get_region_at
ksnip_20230610-030953

I guess one can argue that the difference is within the margin of error

condret:
gonna reopen later

PR #21876, opened 2023-06-10
Create reloc maps for elf bins (experimental) ##bin

condret:
This is more of a POC than a fully functional thing. It lacks of proper map perms, a real write and remap cb. Also RelocMaps do not need to be heap located, it's done here for simplicity. I wanted to get this done befor sleep.
Ideally a RelocMap->data would contain all the necessary information to patch the bin header if this user tries to write to the RelocMap. so if the bin is opened with write permissions. the user can patch a reloc, close r2, open the bin again and the patch would persist. This could not be achieved if we abuse io-cache for setting relocs

**Description**

### DIFF
--- a/libr/bin/p/bin_elf.inc.c
+++ b/libr/bin/p/bin_elf.inc.c
@@ -670,8 +670,55 @@ static RList* relocs(RBinFile *bf) {
 	return ret;
 }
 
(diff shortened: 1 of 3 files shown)


trufae:
i think we agreed on going for the overlay approach, so can we close this PR?

PR #21877, opened 2023-06-10
docker push

gitadvisor:
- [x] I wrote some lines in the [book](https://github.com/radareorg/radare2book) (optional)

**Description**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,16 +10,9 @@ RUN \
   apt-get install -y byobu curl git htop man unzip vim wget && \
   rm -rf /var/lib/apt/lists/*
 
-# Add files.
-ADD root/.bashrc /root/.bashrc
-ADD root/.gitconfig /root/.gitconfig
-ADD root/.scripts /root/.scripts
-
-# Set environment variables.
-ENV HOME /root
 
 # Define working directory.
-WORKDIR /root
+WORKDIR /.tmp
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:14.04
```
trufae, line 1: ubuntu14 srsly?

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+  apt-get install -y build-essential && \
+  apt-get install -y software-properties-common && \
+  apt-get install -y byobu curl git htop man unzip vim wget && \
+  rm -rf /var/lib/apt/lists/*
+
+
+# Define working directory.
+WORKDIR /.tmp
+
+RUN git clone https://github.com/gitadvisor/radare2 && \
```
trufae, line 17: uh? wrong repo

trufae:
We already have docker publish in a separate protected repo. docker credentials wont be shipped here. also the current docker ships all the plugins and its configurable and uses modern ubuntu. and its build for all the available archs and integrated with snap.

Dockerfile
gitadvisor: ubuntu:14.04

Dockerfile
gitadvisor: https://github.com/radareorg/radare2

PR #21887, opened 2023-06-12
R2_590 change ESIL pop signature

luc-tielen:

**Description**

Merge #21886 first, also this doesn't fix all warnings yet..

### DIFF
--- a/libr/anal/esil_dfg.c
+++ b/libr/anal/esil_dfg.c
@@ -630,7 +630,7 @@ beach:
 	return mem_node;
 }
 
-static RGraphNode *_edf_const_get(RAnalEsilDFG *dfg, char *const_value) {
+static RGraphNode *_edf_const_get(RAnalEsilDFG *dfg, const char *const_value) {
 	RGraphNode *orig_value_gnode = r_graph_add_node (dfg->flow, r_anal_esil_dfg_node_new (dfg, const_value));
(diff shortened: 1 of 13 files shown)


trufae:
comments on this @condret ?

trufae:
Can you rebase this PR we should target r2-6.x now

condret:
sorry for the late response. i'm not a fan of this. I don't see any necessity, and all these casts to char * feel bloaty to me. We can do this, if you really want this

trufae:
Theres a noticeable perf boost by not strduping every word. Thats the reason behind this patch. So maybe the casts are ugly but i wanted to make clear the intentions behind

condret:
can't we then just rm the strdups

trufae:
Nope i tried. Some because invalid pointers. So we need again an abstraction over strings

condret:
ok, but why does it have to be const then, if the problem is just the strdups?

trufae:
Because const is the way to tell the user that the token string is there and cant be modified. Making it clear that its not in the heap and cant be owned by the struct

trufae:
@condret can you take care of this optimiation before 6.x your proposal about making the whole string a single char* and then replace , with 0 looks good to me too

trufae:
too much conflicts closing

PR #21889, opened 2023-06-13
Map overlay ##io

condret:
Ok, so why is this better than cache abuse and reloc_maps:

reloc_maps spams a shitload of maps, rwx perms are fuzzy to handle, potentially causes crashes when a dynamic loaded plugin gets dlclosed at runtime

cache abuse cannot handle remapping, map priorization and it allows esil to ignore w perms, because iocache disables checks in io. Also it renders iocache useless for it's primary purpose

**Description**

### DIFF
--- a/libr/bin/p/bin_bflt.c
+++ b/libr/bin/p/bin_bflt.c
@@ -21,10 +21,10 @@ static RList *entries(RBinFile *bf) {
 	return ret;
 }
 
-static void __patch_reloc(RBuffer *buf, ut32 addr_to_patch, ut32 data_offset) {
+static void __patch_reloc(RIOBind *iob, ut32 addr_to_patch, ut32 data_offset) {
 	ut8 val[4] = { 0 };
 	r_write_le32 (val, data_offset);
-	r_buf_write_at (buf, addr_to_patch, (void *) val, sizeof (val));
+	iob->overlay_write_at (iob->io, addr_to_patch, val, sizeof (val));
 }
 
 static int search_old_relocation(struct reloc_struct_t *reloc_table, ut32 addr_to_patch, int n_reloc) {
@@ -40,29 +40,20 @@ static int search_old_relocation(struct reloc_struct_t *reloc_table, ut32 addr_t
 static RList *patch_relocs(RBin *b) {
 	r_return_val_if_fail (b && b->iob.io, NULL);
 	struct r_bin_bflt_obj *bin = NULL;
-	RList *list = NULL;
-	int i = 0;
-	if (!b->iob.io->desc) {
-		R_LOG_WARN ("No default descriptor loaded");
-		return NULL;
-	}
-	if (!r_io_cache_writable (b->iob.io)) {
-		R_LOG_WARN ("Please run r2 with -e io.cache=true to patch relocations");
-		return NULL;
-	}
 	RBinObject *obj = r_bin_cur_object (b);
 	if (!obj) {
 		return NULL;
 	}
 	bin = obj->bin_obj;
-	list = r_list_newf ((RListFree) free);
+	RList *list = r_list_newf ((RListFree) free);
 	if (!list) {
 		return NULL;
 	}
 	if (bin->got_table) {
 		struct reloc_struct_t *got_table = bin->got_table;
+		int i;
 		for (i = 0; i < bin->n_got; i++) {
-			__patch_reloc (bin->b, got_table[i].addr_to_patch,
+			__patch_reloc (&b->iob, got_table[i].addr_to_patch,
 				got_table[i].data_offset);
 			RBinReloc *reloc = R_NEW0 (RBinReloc);
 			if (reloc) {
@@ -77,15 +68,15 @@ static RList *patch_relocs(RBin *b) {
 
 	if (bin->reloc_table) {
 		struct reloc_struct_t *reloc_table = bin->reloc_table;
+		int i = 0;
 		for (i = 0; i < bin->hdr->reloc_count; i++) {
 			int found = search_old_relocation (reloc_table,
-				reloc_table[i].addr_to_patch,
-				bin->hdr->reloc_count);
+				reloc_table[i].addr_to_patch, bin->hdr->reloc_count);
 			if (found != -1) {
-				__patch_reloc (bin->b, reloc_table[found].addr_to_patch,
+				__patch_reloc (&b->iob, reloc_table[found].addr_to_patch,
 					reloc_table[i].data_offset);
 			} else {
-				__patch_reloc (bin->b, reloc_table[i].addr_to_patch,
+				__patch_reloc (&b->iob, reloc_table[i].addr_to_patch,
 					reloc_table[i].data_offset);
 			}
 			RBinReloc *reloc = R_NEW0 (RBinReloc);
@@ -98,9 +89,6 @@ static RList *patch_relocs(RBin *b) {
 		}
 		R_FREE (bin->reloc_table);
 	}
-	ut64 tmpsz;
-	const ut8 *tmp = r_buf_data (bin->b, &tmpsz);
-	b->iob.write_at (b->iob.io, 0, tmp, tmpsz);
 	return list;
 }
 
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -367,7 +367,8 @@ R_API RList* r_io_map_get_by_fd(RIO *io, int fd);
 R_API bool r_io_map_resize(RIO *io, ut32 id, ut64 newsize);
 R_API void r_io_map_read_from_overlay(RIOMap *map, ut64 addr, ut8 *buf, int len);
 R_API bool r_io_map_write_to_overlay(RIOMap *map, ut64 addr, const ut8 *buf, int len);
-R_IPI bool _io_map_get_overlay_intersects(RIOMap *map, RQueue *q, ut64 addr, int len);
+R_IPI bool io_map_get_overlay_intersects(RIOMap *map, RQueue *q, ut64 addr, int len);
+R_API void r_io_map_drain_overlay(RIOMap *map);
 
 // next free address to place a map.. maybe just unify
 R_API bool r_io_map_locate(RIO *io, ut64 *addr, const ut64 size, ut64 load_align);
@@ -438,6 +439,7 @@ R_API bool r_io_set_write_mask(RIO *io, const ut8 *mask, int len);
 R_API void r_io_bind(RIO *io, RIOBind *bnd);
 R_API bool r_io_shift(RIO *io, ut64 start, ut64 end, st64 move);
 R_API ut64 r_io_seek(RIO *io, ut64 offset, int whence);
+R_API void r_io_drain_overlay(RIO *io);
 R_API void r_io_fini(RIO *io);
 R_API void r_io_free(RIO *io);
 #define r_io_bind_init(x) memset (&(x), 0, sizeof (x))
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -222,26 +222,14 @@ typedef struct r_io_plugin_t {
 	bool (*check)(RIO *io, const char *, bool many);
 } RIOPlugin;
 
-struct r_io_map_t;
-
-typedef struct r_io_reloc_map_t {
-	void *data;
-	int (*read)(RIO *io, struct r_io_map_t *map, ut64 addr, ut8 *buf, int len);
-	int (*write)(RIO *io, struct r_io_map_t *map, ut64 addr, const ut8 *buf, int len);
-	bool (*remap)(RIO *io, struct r_io_map_t *map, ut64 addr);
-	void (*free)(void *data);
-} RIORelocMap;
-
 typedef struct r_io_map_t {
 	int fd;
 	int perm;
 	ut32 id;
 	ut64 ts;
 	RInterval itv;
-	union {
-		ut64 delta; // paddr = vaddr - itv.addr + delta
-		RIORelocMap *reloc_map;
-	};
+	ut64 delta; // paddr = vaddr - itv.addr + delta
+	RRBTree *overlay;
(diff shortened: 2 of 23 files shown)


trufae:
Sounds like the way to go for me too 👍 good work. we just need to massage it a little to not have regressions

libr/include/r_io.h
```diff
@@ -307,6 +296,7 @@ typedef struct r_io_bind_t {
 	RIOClose close;
 	RIOReadAt read_at;
 	RIOWriteAt write_at;
```
trufae, line 298: should not be used at all imho. bin plugins should just parse stuff, not modify the underlying data contents.

libr/io/io.c
```diff
@@ -226,6 +226,19 @@ R_API bool r_io_vwrite_at(RIO *io, ut64 vaddr, const ut8* buf, int len) {
 	return r_io_bank_write_at (io, io->bank, vaddr, buf, len);
 }
 
+R_API bool r_io_vwrite_to_overlay_at(RIO *io, ut64 vaddr, const ut8 *buf, int len) {
+	r_return_val_if_fail (io && buf && len > 0, false);
+	if ((UT64_MAX - (len - 1)) < vaddr) {
+		int _len = UT64_MAX - vaddr + 1;
```
trufae, line 232:
```suggestion
		int olen = UT64_MAX - vaddr + 1;
```

better olen than _len but meh

libr/io/io_bank.c
```diff
@@ -844,6 +865,40 @@ R_API bool r_io_bank_write_at(RIO *io, const ut32 bankid, ut64 addr, const ut8 *
 	return ret;
 }
 
+R_API bool r_io_bank_write_to_overlay_at(RIO *io, const ut32 bankid, ut64 addr, const ut8 *buf, int len) {
+	r_return_val_if_fail (io, false);
+	RIOBank *bank = r_io_bank_get (io, bankid);
+	if (!bank) {
```
trufae, line 871: maybe worth R_LOG_WARN here because if that happens we may miss data without knowing why

libr/io/io_map.c
```diff
@@ -493,3 +450,187 @@ R_API RIOMap *r_io_map_get_by_ref(RIO *io, RIOMapRef *ref) {
...
 }
+
+typedef struct map_overlay_chunk_t {
+	// itv is relative to the map vaddr
+	// this is to keep maps with overlay moveable in the va space
+	RInterval itv;
+	ut8 *buf;
+} MapOverlayChunk;
+
+static int _overlay_chunk_find (void *incoming, void *in, void *user) {
```
trufae, line 461:
```suggestion
static int _overlay_chunk_find(void *incoming, void *in, void *user) {
```

libr/io/io_map.c
```diff
@@ -493,3 +450,187 @@ R_API RIOMap *r_io_map_get_by_ref(RIO *io, RIOMapRef *ref) {
...
+			break;
+		}
+		node = prev;
+		prev = r_rbnode_prev (prev);
+	}
+	chunk = (MapOverlayChunk *)node->data;
+	do {
+		addr = R_MAX (r_itv_begin (search_itv), r_itv_begin (chunk->itv));
+		ut8 *dst = &buf[addr - r_itv_begin (search_itv)];
+		ut8 *src = &chunk->buf[addr - r_itv_begin (chunk->itv)];
```
trufae, line 499: src is const

libr/io/io_map.c
```diff
@@ -493,3 +450,187 @@ R_API RIOMap *r_io_map_get_by_ref(RIO *io, RIOMapRef *ref) {
...
+		chunk->buf = realloc (chunk->buf, r_itv_end (chunk->itv) - r_itv_begin (search_itv));
+		memmove (&chunk->buf[r_itv_size (search_itv)],
+			&chunk->buf[r_itv_end (search_itv) - r_itv_begin (chunk->itv)],
+			r_itv_end (chunk->itv) - r_itv_end (search_itv));
+		memcpy (chunk->buf, buf, r_itv_size (search_itv));
+		chunk->itv.size = r_itv_end (chunk->itv) - r_itv_begin (search_itv);
+		chunk->itv.addr = search_itv.addr;
+		return true;
+	}
+	chunk = R_NEW0 (MapOverlayChunk);
```
trufae, line 595: chk for null

trufae (changes requested):
i like the approach of having rbin writing the relocs into an overlay map because thats tied to the underlying file. So i have some questions at this point

- tests are not pasisng, so something is not working well
- how can we list/show/modify those overlayed writes from the r2shell?
- why not reuse the write_at callback and just put the map in "overlay" mode, so it redirects stuff to the layer on top?
- how much this differs from the pcache?

libr/io/io_map.c
```diff
@@ -493,3 +450,187 @@ R_API RIOMap *r_io_map_get_by_ref(RIO *io, RIOMapRef *ref) {
...
+	}
+	chunk = (MapOverlayChunk *)node->data;
+	if (r_itv_include (chunk->itv, search_itv)) {
+		ut8 *dst = &chunk->buf[r_itv_begin (search_itv) - r_itv_begin (chunk->itv)];
+		memcpy (dst, buf, r_itv_size (search_itv));
+		return true;
+	}
+	if (r_itv_begin (chunk->itv) < r_itv_begin (search_itv)) {
+		chunk->itv.size = r_itv_begin (search_itv) - r_itv_begin (chunk->itv);
+		chunk->buf = realloc (chunk->buf, r_itv_size (chunk->itv));
```
luc-tielen, line 574: check if realloc returns valid pointer

libr/io/io_map.c
```diff
@@ -493,3 +450,187 @@ R_API RIOMap *r_io_map_get_by_ref(RIO *io, RIOMapRef *ref) {
...
+	if (node) {
+		chunk = (MapOverlayChunk *)node->data;
+		while (chunk && r_itv_include (search_itv, chunk->itv)) {
+			node = r_rbnode_next (node);
+			r_crbtree_delete (map->overlay, &chunk->itv, _overlay_chunk_find, NULL);
+			chunk = node? (MapOverlayChunk *)node->data: NULL;
+		}
+	}
+	if (chunk && r_itv_end (search_itv) >= r_itv_begin (chunk->itv)) {
+		chunk->buf = realloc (chunk->buf, r_itv_end (chunk->itv) - r_itv_begin (search_itv));
```
luc-tielen, line 586: check for valid pointer

libr/io/io_map.c
condret: reloc always returns a valid pointer here, because the new size is guaranteed to be smaller than the old size

libr/include/r_io.h
```diff
@@ -398,14 +390,15 @@ R_API void r_io_bank_del_map(RIO *io, const ut32 bankid, const ut32 mapid);
...
 R_API bool r_io_bank_write_at(RIO *io, const ut32 bankid, ut64 addr, const ut8 *buf, int len);
+R_API bool r_io_bank_write_to_overlay_at(RIO *io, const ut32 bankid, ut64 addr, const ut8 *buf, int len);
 R_API int r_io_bank_read_from_submap_at(RIO *io, const ut32 bankid, ut64 addr, ut8 *buf, int len);
 R_API int r_io_bank_write_to_submap_at(RIO *io, const ut32 bankid, ut64 addr, const ut8 *buf, int len);
 R_API void r_io_bank_drain(RIO *io, const ut32 bankid);
 
 //io.c
 R_API RIO *r_io_new(void);
 R_API void r_io_init(RIO *io);
-R_API RIODesc *r_io_open_nomap(RIO *io, const char *uri, int flags, int mode);		//should return int
```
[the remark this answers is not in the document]
condret: nah

libr/io/io_bank.c
condret: ok, but it will be a funny warning, we can make it serious when breaking is over

libr/io/io.c
condret: olen implies old len, doesn't it

libr/io/io_map.c
condret: if you insist

condret:
the failing tests are most likely because relocs are not expected to work properly.

the reason why I don't reuse the writeat cb is because there are also other parts of r2, that use iobind for writing (fx esil).
so the iowrite cb needs to stay as it is.

so far I don't have any concrete plans to show what parts of a map are affected by the overlay in the r2shell.

this is different from pcache, because pcache is tied to desc, meanwhile overlay is tied to maps. The purpose of overlay is relocs, pcache is for caching user writes and creating patch files

libr/include/r_io.h
condret: the entire rbin architecture leaves a lot to be desired

libr/bin/p/bin_bflt.c
```diff
@@ -21,10 +21,10 @@ static RList *entries(RBinFile *bf) {
 	return ret;
 }
 
-static void __patch_reloc(RBuffer *buf, ut32 addr_to_patch, ut32 data_offset) {
+static void __patch_reloc(RIOBind *iob, ut32 addr_to_patch, ut32 data_offset) {
 	ut8 val[4] = { 0 };
 	r_write_le32 (val, data_offset);
-	r_buf_write_at (buf, addr_to_patch, (void *) val, sizeof (val));
+	iob->overlay_write_at (iob->io, addr_to_patch, val, sizeof (val));
```
trufae, line 27: my main concern on this is that your changes makes it impossible to test (without building a separate branch) with the current layered io cache. so i would like to make this configurable at runtime to choose if u want the relocs to be written on the overlay or in the buffer (actually rbufwrite() was silently calling the io api under the hood)

libr/io/io_bank.c
trufae: yea

test/db/cmd/cmd_i
```diff
@@ -3852,10 +3852,10 @@ EXPECT=<<EOF
             ;-- section..init_array:
             ;-- segment.LOAD1:
             ;-- segment.GNU_RELRO:
-            0x0021bfa8      .qword 0x0000000000005d20 ; entry.init0    ; [18] -rw- section size 8 named .init_array
+            0x0021bfa8      .qword 0x0000000000005d20 ; entry.init0    ; RELOC 64  ; [18] -rw- section size 8 named .init_array
 
             ;-- section..fini_array:
-            0x0021bfb0      .qword 0x0000000000005ce0 ; entry.fini0    ; [19] -rw- section size 8 named .fini_array
+            0x0021bfb0      .qword 0x0000000000005ce0 ; entry.fini0    ; RELOC 64  ; [19] -rw- section size 8 named .fini_array
```
trufae, line 3858: woot! this is actually better than before

test/db/cmd/cmd_ie
```diff
@@ -9,9 +9,7 @@ EXPECT=<<EOF
 
 0 entrypoints
 EOF
-EXPECT_ERR=<<EOF
-WARN: run r2 with -e bin.cache=true to fix relocations in disassembly
-EOF
+EXPECT_ERR=
```
trufae, line 12: omg this is making relocs on by default. which is something we dont want yet. because im aware of many bugs still in the relocs parsing/patching, and the iocache layers is good for debugging that.

but still i think bin.cache was just a hack and we need `bin.applyrelocs` or sthg?

im happy to merge your pr if the current behaviour is not modified (choose overlay vs iocache) and then we can remove the bin.cache and call it bin.patchrelocs or wathever makes sense,

test/db/formats/dwarf
```diff
@@ -701,23 +701,23 @@ line: 3
...
-DEBUG: [cbin.c:3284] (section .rela.plt) Cd 8[3] @ 0x610
-DEBUG: [cbin.c:3284] (section .init_array) Cd 8[1] @ 0x3db8
-DEBUG: [cbin.c:3284] (section .fini_array) Cd 8[1] @ 0x3dc0
-DEBUG: [cbin.c:3284] (section .dynamic) Cd 8[62] @ 0x3dc8
-DEBUG: [cbin.c:2552] Cannot resolve symbol address __libc_start_main
-DEBUG: [cbin.c:2552] Cannot resolve symbol address _ITM_deregisterTMCloneTable
-DEBUG: [cbin.c:2552] Cannot resolve symbol address __gmon_start__
-DEBUG: [cbin.c:2552] Cannot resolve symbol address _ITM_registerTMCloneTable
-DEBUG: [cbin.c:2552] Cannot resolve symbol address __cxa_finalize
+DEBUG: [cbin.c:3269] (section .dynstr) Css 141 @ 0x480
```
trufae, line 704: i hate this test :D

test/db/formats/rel
```diff
@@ -1,17 +1,16 @@
 NAME=uncached err
 FILE=bins/rel/d_profileNP.rel
-ARGS=-e bin.cache=false
+ARGS=-e bin.relocs.apply=false
```
trufae, line 3: oh!

libr/include/r_io.h
trufae: Agree, but we can always make it worse. I mean clean it up

libr/io/io.c
trufae: Or original_length

test/db/formats/dwarf
trufae: It was suposed to ensure the log apis work as expected but maybe better to do this in a unit test that we dont need to modify that often

test/db/cmd/cmd_ie
condret: I just removed the warnings, because they are annoying

condret: if your concern is debugging, i can give you some #!c scripts for that

test/db/cmd/cmd_i
condret: laptop_giga_chad

PR #21934, opened 2023-06-18
LZ4 removal

MewtR:

**Description**

This PR removes the official LZ4 implementation from radare2. Users can still it by installing it on their machines and passing the appropriate flags (`-Duse_sys_lz4` for meson and `--with-syslz4` for make) when building radare2.
radare2 will now use smallz4 as it's LZ4 implementation.


trufae:
That looks great! thanks for taking care of this change! and welcome back :)

PR #21935, opened 2023-06-18
Fix to symlink of DLL which causes link loop

dvertx:

**Description**

Without the fix, install-symlink in Makefile causes link loop of libr/bin/d/dll directory:
`lrwxrwxrwx 1 root root 40 Jun 19 01:06 libr/bin/d/dll/dll -> /home/dvertx/radare2/libr/bin/d/dll`

Output of `lr` command:
`...
./libr/bin/d/dll/dll/dll/dll/dll/dll/dll/dll/kernel32.sdb.txt
./libr/bin/d/dll/dll/dll/dll/dll/dll/dll/dll/keyboard.sdb
./libr/bin/d/dll/dll/dll/dll/dll/dll/dll/dll/keyboard.sdb.txt
./libr/bin/d/dll/dll/dll/dll/dll/dll/dll/dll/lzexpand.sdb
./libr/bin/d/dll/dll/dll/dll/dll/dll/dll/dll/lzexpand.sdb.txt
./libr/bin/d/dll/dll/dll/dll/dll/dll/dll/dll/maincp16.sdb
./libr/bin/d/dll/dll/dll/dll/dll/dll/dll/dll/maincp16.sdb.txt
...`

### DIFF
--- a/libr/bin/d/Makefile
+++ b/libr/bin/d/Makefile
@@ -51,7 +51,7 @@ CWD=$(shell pwd)
 symstall install-symlink:
 	mkdir -p "$P"
 	for FILE in * ; do \
-		if [ "$$FILE" != Makefile  && "$$FILE" != dll ]; then \
+		if [ "$$FILE" != Makefile  -a "$$FILE" != dll ]; then \
 			ln -fs "${CWD}/$$FILE" "$P/$$FILE" ; \
 		fi ; \
 	done
--- a/libr/bin/d/Makefile
+++ b/libr/bin/d/Makefile
@@ -51,7 +51,7 @@ CWD=$(shell pwd)
 symstall install-symlink:
 	mkdir -p "$P"
 	for FILE in * ; do \
-		if [ "$$FILE" != Makefile ]; then \
+		if [ "$$FILE" != Makefile  && "$$FILE" != dll ]; then \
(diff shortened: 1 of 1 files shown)


libr/bin/d/Makefile
```diff
@@ -51,7 +51,7 @@ CWD=$(shell pwd)
 symstall install-symlink:
 	mkdir -p "$P"
 	for FILE in * ; do \
-		if [ "$$FILE" != Makefile ]; then \
+		if [ "$$FILE" != Makefile  && "$$FILE" != dll ]; then \
```
trufae, line 54:
```suggestion
		if [ "$$FILE" != Makefile  -a "$$FILE" != dll ]; then \
```

trufae:
Good catch! never had a chance to find out the root cause of this

dvertx:
Glad to have helped :)

trufae:
0A8E3B83-1E8B-4688-A54A-2065AF884594

seems like it breaks stuff

trufae:
I think i did the right fix now in master, check it out and confirm :) the problem happens after the 2nd installation

dvertx:
After your revert, I checked that my "fix" actually caused pfo to lost its link to dll definition files. :(
Then I cloned radare2 to another directory, and was surprised that the clean copy was not bugged by link loop at first installation. But the problem reemerged on subsequent installs. Great that you fixed the problem

PR #21938, opened 2023-06-19
Add r_graph_dom_tree ##util

condret:
This is supposed to compute a dominance tree from a given RGraph and RGraphNode (root).

I took inspiration from https://www.cs.princeton.edu/courses/archive/fall03/cs528/handouts/a%20fast%20algorithm%20for%20finding.pdf, though I'm not sure if i understood it correctly, because I fucking HATE mathematical notation.

I also found this paper https://users-cs.au.dk/gerth/advising/thesis/henrik-knakkegaard-christensen.pdf

Review is appreciated, but please don't waste any time with cosmetic issues (coding style, unchecked allocations, etc.).
I can run the indent.sh script later.

### DIFF
--- a/libr/util/graph.c
+++ b/libr/util/graph.c
@@ -312,3 +312,121 @@ R_API void r_graph_dfs(RGraph *g, RGraphVisitor *vis) {
 		free (color);
 	}
 }
+
+typedef struct _dfs_inserter {
+	RGraph *g;
+	HtUP *reverse;	//reverse lookup of nodes
+//	RList *mo;	//multiple out nodes
+	RList *mi;	//multiple in nodes
+	ut32 idx;
+	bool fail;
+} DfsInserter;
+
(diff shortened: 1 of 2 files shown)


libr/util/graph.c
```diff
@@ -312,3 +312,121 @@ R_API void r_graph_dfs(RGraph *g, RGraphVisitor *vis) {
 		free (color);
 	}
 }
+
+typedef struct _dfs_inserter {
+	RGraph *g;
+	HtUP *reverse;	//reverse lookup of nodes
+//	RList *mo;	//multiple out nodes
+	RList *mi;	//multiple in nodes
```
luc-tielen, line 320: @trufae at some point these should be turned into Vec's for better memory locality / performance. Also for RGraph in general.

condret: tbh i think a stack or queue might be better here, but list is more flexible

PR #21939, opened 2023-06-19
Another attempt to do EsilHooks ##esil

condret:
Maybe this time

Still a bit WIP

**Description**

### DIFF
--- a/libr/esil/esil.c
+++ b/libr/esil/esil.c
@@ -114,6 +114,10 @@ R_API REsil *r_esil_new(int stacksize, int iotrap, unsigned int addrsize) {
 	esil->addrmask = genmask (addrsize - 1);
 	esil->trace = r_esil_trace_new (esil);
 	esil->hooks = r_esil_hooks_new ();
+	esil->stats_mr_handle = UT32_MAX;
+	esil->stats_mw_handle = UT32_MAX;
+	esil->stats_rr_handle = UT32_MAX;
+	esil->stats_rw_handle = UT32_MAX;
 	return esil;
 }
 
--- a/libr/esil/esil.c
+++ b/libr/esil/esil.c
@@ -113,6 +113,7 @@ R_API REsil *r_esil_new(int stacksize, int iotrap, unsigned int addrsize) {
 	r_esil_plugins_init (esil);
 	esil->addrmask = genmask (addrsize - 1);
 	esil->trace = r_esil_trace_new (esil);
+	esil->hooks = r_esil_hooks_new ();
 	return esil;
 }
 
@@ -208,6 +209,7 @@ R_API void r_esil_free(REsil *esil) {
 		esil->anal->arch->esil = NULL;
 	}
 
+	r_esil_hooks_free (esil->hooks);
 	r_esil_plugins_fini (esil);
 	r_esil_handlers_fini (esil);
 	ht_pp_free (esil->ops);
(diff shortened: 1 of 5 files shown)


libr/esil/esil.c
```diff
@@ -4011,10 +4018,16 @@ R_API bool r_esil_setup(REsil *esil, RAnal *anal, int romem, int stats, int nonu
 		esil->cb.reg_write = internal_esil_reg_write_no_null;
 		esil->cb.mem_read = internal_esil_mem_read_no_null;
 		esil->cb.mem_write = internal_esil_mem_write_no_null;
+		r_esil_set_reg_write_imp (esil, (REsilImpHookRegWriteCB)internal_esil_reg_write_no_null, esil);
```
trufae, line 4021: ideally the cast shuoldnt be needed, otherwise we can pass functions with different signature and get runtime crashes hard to debug

libr/esil/esil_hooks.c
```diff
@@ -0,0 +1,363 @@
...
+	hook->imr = imp;
+	return true;
+}
+
+R_API void r_esil_del_mem_read_imp(REsil *esil) {
+	r_return_if_fail (esil && esil->hooks);
+	R_FREE (esil->hooks->mem_read_implementation);
+}
+
+R_API bool r_esil_set_mem_write_imp(REsil *esil, REsilImpHookMemWriteCB imp, void *user) {
```
trufae, line 57: what about having only a set method and tell which event as argument, index those type of events in an fixed size array like esil->hooks[R_ESIL_EVENT_RR]..

libr/esil/esil_hooks.c
```diff
@@ -0,0 +1,363 @@
...
+	r_return_val_if_fail (esil && esil->hooks, false);
+	if (!esil->hooks->mem_write_implementation) {
+		esil->hooks->mem_write_implementation = R_NEW (REsilHook);
```
trufae, line 60: uhm, only 1 handler per event? we cant have multiple listeners? that seems like not what we want right?

libr/esil/esil_hooks.c
```diff
@@ -0,0 +1,363 @@
...
+		esil->hooks->reg_write_implementation = R_NEW (REsilHook);
+	}
+	REsilHook *hook = esil->hooks->reg_write_implementation;
+	r_return_val_if_fail (hook, false);
+	hook->user = user;
+	hook->irw = imp;
+	return true;
+}
+
+R_API void r_esil_del_reg_write_imp(REsil *esil) {
```
trufae, line 103: i would prefer to have a smaller API  and less functions, we want to do the ADD, DEL, CLEAR operations on RR,RW,MM,MW array of callbacks. this array of pointers can be an RVec, so they are executed in the right order and hold a void*user pointer. im thinking about something like this:

`esil.hook(ADD, MR, cb, user);`

ADD = enum for the supported operations: ADD, DEL, CLEAR, ..
MR = memory read, just another enum with other stuff
cb = the user provided callback for the given esil operation hook
user = user data pointer associated with the given callback

The cb is used as a uniqueid for each event, this is, you can `esil.hook(DEL, cb, NULL);`

At the end this is very similar to what REvent, what are the limitations you can think about it?

trufae (changes requested):
i like the way it works, but i think this introduces so many methods when we can make it much simpler by using a single function that takes operation as argument.

It's also very similar to how REvent works. maybe better to re-evaluate to improve this api that we dont use yet because we can have multiple event handlers around and that will be important for ranal too.

What do you think?

trufae:
ping, rebase for conflicts, and respond my comments pls

condret:
I'll look at this again in the next breaking

trufae:
can this PR be closed now?

PR #21942, opened 2023-06-19
Fix some ls cmd part

dvertx:

**Description**

Found some errors in ls cmd which are obvious. Some are not so obvious and would like more time to delve in it to create a fix in another PR.
I also put a note as a remark: tilde (~) is always received as 0x00, why is that, masked in the shell part or what?

### DIFF
--- a/libr/util/syscmd.c
+++ b/libr/util/syscmd.c
@@ -177,20 +177,15 @@ R_API char *r_syscmd_ls(const char *input, int cons_width) {
 		}
 		if ((!strncmp (input, "-e", 2))) {
 			printfmt = FMT_EMOJI;
-			path = r_str_trim_head_ro (path + 1);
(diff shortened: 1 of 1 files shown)


trufae:
~ is the internal grep you need to escape or quote to not strip it from the shell

dvertx:
Sorry, I don't understand what you mean. It's just that by never receiving ~, there is no possibility to list home dir or part thereof by tilde. And there is no use of the strncmp against ~/. What do you suggest I do?

trufae:
Yes, you can skip the r2 command syntax parsing in 3 different ways:

```
Escaping the special char (~, @, |, >, ..)
> ls \~/foo

Quoting the whole command:
> "ls ~/foo"

Prefix the command with double-quote / or single quote, to make it literal:
> ""ls ~/foo
>'ls ~/foo
```

PR #21953, opened 2023-06-22
Cleanup 590 and qjs integration

luc-tielen:

**Description**


trufae:
Screenshot 2023-06-26 at 12 06 22

this is wrong, runtime asserts shouldnt hit normal sessions, also, the full path of the error message will be different from host to host

luc-tielen:
> Screenshot 2023-06-26 at 12 06 22
> this is wrong, runtime asserts shouldnt hit normal sessions, also, the full path of the error message will be different from host to host

I know, that test wasn't passing yet 😅.

PR #21959, opened 2023-06-25
[fix]Null Pointer reference

Pikachu1412:

**Description**
When as is newed, I add the judgment to check if it is empty, if it is empty, return directly

### DIFF
--- a/libr/main/rasm2.c
+++ b/libr/main/rasm2.c
@@ -751,7 +751,7 @@ R_API int r_main_rasm2(int argc, const char *argv[]) {
 	}
 	R_FREE (log_level);
 	RAsmState *as = __as_new ();
-	if(!as){
+	if (!as) {
 		return 0;
 	}
 	// TODO set addrbytes
--- a/libr/main/rasm2.c
+++ b/libr/main/rasm2.c
@@ -191,9 +191,6 @@ static void rarch2_list(RAsmState *as, const char *arch) {
 	RListIter *iter, *iter2;
 	const char *feat;
 	PJ *pj = NULL;
-	if(!as){
-		return;
-	}
 	if (as->json) {
 		pj = pj_new ();
 		pj_a (pj);
--- a/libr/main/rasm2.c
+++ b/libr/main/rasm2.c
@@ -191,6 +191,9 @@ static void rarch2_list(RAsmState *as, const char *arch) {
 	RListIter *iter, *iter2;
 	const char *feat;
 	PJ *pj = NULL;
+	if(!as){
+		return;
+	}
 	if (as->json) {
 		pj = pj_new ();
 		pj_a (pj);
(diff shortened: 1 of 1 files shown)


libr/main/rasm2.c
```diff
@@ -748,7 +751,9 @@ R_API int r_main_rasm2(int argc, const char *argv[]) {
 	}
 	R_FREE (log_level);
 	RAsmState *as = __as_new ();
-
+	if(!as){
```
trufae, line 754: Add proper spaces
```suggestion
	if (!as) {
```

libr/main/rasm2.c
```diff
@@ -748,7 +751,9 @@ R_API int r_main_rasm2(int argc, const char *argv[]) {
...
+		return 0;
```
trufae, line 755: Return 1 because thats an error
```suggestion
		return 1;
```

libr/main/rasm2.c
```diff
@@ -191,6 +191,9 @@ static void rarch2_list(RAsmState *as, const char *arch) {
 	RListIter *iter, *iter2;
 	const char *feat;
 	PJ *pj = NULL;
+	if(!as){
```
trufae, line 194: Probably better to use r_return_if_fail because that shouldnt happen

trufae (changes requested):

Pikachu1412:
Thank you for your suggestions!




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "radareorg/radare2"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2023年6月25日(星期天) 晚上6:41
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [radareorg/radare2] [fix]Null Pointer reference (PR #21959)





 
@trufae requested changes on this pull request.
 
 
In libr/main/rasm2.c:
 &gt; @@ -748,7 +751,9 @@ R_API int r_main_rasm2(int argc, const char *argv[]) {  	}  	R_FREE (log_level);  	RAsmState *as = __as_new (); - +	if(!as){  
Add proper spaces
 ⬇️ Suggested change -	if(!as){ +	if (!as) {  
 
In libr/main/rasm2.c:
 &gt; @@ -748,7 +751,9 @@ R_API int r_main_rasm2(int argc, const char *argv[]) {  	}  	R_FREE (log_level);  	RAsmState *as = __as_new (); - +	if(!as){ +		return 0;  
Return 1 because thats an error
 ⬇️ Suggested change -		return 0; +		return 1;  
 
In libr/main/rasm2.c:
 &gt; @@ -191,6 +191,9 @@ static void rarch2_list(RAsmState *as, const char *arch) {  	RListIter *iter, *iter2;  	const char *feat;  	PJ *pj = NULL; +	if(!as){  
Probably better to use r_return_if_fail because that shouldnt happen

libr/main/rasm2.c
Pikachu1412: There is only one call and it was checked before the call, no need to check again

trufae:
Looks good now! Thanks

PR #21967, opened 2023-06-26
Remove unnecessary else, for better config descriptions

dvertx:

**Description**

Current radare2 configuration descriptions through e? command only shows the descriptions parts, but not the variable names, as follows:
```
[0x000010a0]> e?asm.bytes
display the bytes of each instruction
align bytes at right (left padding space)
show instruction bits instead of bytes
colorize bytes depending on opcode size + variant information
display the bytes at the right of the disassembly
separate hexadecimal bytes with a whitespace
```

This patch will (re-)enable showing the variable names as in [the book](https://book.rada.re/search_bytes/intro.html#search-options)
```
[0x00006ab0]> e?asm.bytes
           asm.bytes: display the bytes of each instruction
     asm.bytes.align: align bytes at right (left padding space)
    asm.bytes.asbits: show instruction bits instead of bytes
   asm.bytes.opcolor: colorize bytes depending on opcode size + variant information
     asm.bytes.right: display the bytes at the right of the disassembly
     asm.bytes.space: separate hexadecimal bytes with a whitespace
```

### DIFF
--- a/libr/config/config.c
+++ b/libr/config/config.c
@@ -163,7 +163,7 @@ R_API void r_config_list(RConfig *cfg, const char *str, int rad) {
 	const char *sfx = "";
 	const char *pfx = "";
 	int len = 0;
-	bool verbose = false;
+	bool found, verbose = false;
 	PJ *pj = NULL;
 
 	if (!IS_NULLSTR (str)) {
@@ -215,15 +215,21 @@ R_API void r_config_list(RConfig *cfg, const char *str, int rad) {
 	case 2:
 		r_list_foreach (cfg->nodes, iter, node) {
 			if (!str || (str && (!strncmp (str, node->name, len)))) {
-				if (!str || !strncmp (str, node->name, len)) {
-					if (R_STR_ISNOTEMPTY (str)) {
-						cfg->cb_printf ("%s\n", r_str_get (node->desc));
-					} else {
-						cfg->cb_printf ("%20s: %s\n", node->name,
-							r_str_get (node->desc));
-					}
-				}
+				cfg->cb_printf ("%20s: %s\n", node->name, r_str_get (node->desc));
 			}
+		}
+		break;
+	case 3:
+		found = false;
+		r_list_foreach (cfg->nodes, iter, node) {
+			if (!str || (str && (!strcmp (str, node->name)))) {
+				found = true;
+				cfg->cb_printf ("%s\n", r_str_get (node->desc));
+				break;
+			}
+		}
+		if (!found) {
+			cfg->cb_printf ("Key not found. Try e??%s\n", str);
 		}
 		break;
 	case 's':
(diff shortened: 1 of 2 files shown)


trufae:
Uhm, i think e? should only show the description of the key you give, not all the keys matching that string, thats the reason why there's an if conditional there, otherwise there's no way to query just the text of the description

libr/config/config.c
```diff
@@ -217,8 +217,6 @@ R_API void r_config_list(RConfig *cfg, const char *str, int rad) {
 			if (!str || (str && (!strncmp (str, node->name, len)))) {
 				if (!str || !strncmp (str, node->name, len)) {
```
trufae, line 218: these conditionals look like very similar ones to the ones in the line above, this can be probably improved with caching the results in variables instead

libr/config/config.c
```diff
@@ -217,8 +217,6 @@ R_API void r_config_list(RConfig *cfg, const char *str, int rad) {
...
 					if (R_STR_ISNOTEMPTY (str)) {
-						cfg->cb_printf ("%s\n", r_str_get (node->desc));
```
trufae, line 220: i think just adding a break after this will fix the issue. if you want to get the other output you can just use the internal grep. but imho the behaviour of `e?k` is to get only the description of `k`

trufae (changes requested):
.

trufae:
sorry not just the break, also needs to change the strncmp with strcmp

libr/config/config.c
dvertx: OK. Let me try a few things for the conditions and the single/double question mark parsing, because e??search also did not show the variable names. I'll update the pull request later

dvertx:
OK, I will see to that. Thanks for the heads up

trufae:
Don't take the book as reference, imho the book is the one that needs to be improved/updated and it's always better to discuss for the better behaviour rather than assuming something is good as documented

Thanks for the contrib!

dvertx:
That's true, better to update the book. But on the other hand, I personally think that e?? should provides more descriptions and gives more leniency to users who are not familiar with radare2. So, here is my take:

When the `e?` command supplied with correct key, gives out the (only correct) description:
```
[0x00006ab0]> e?asm.bytes
display the bytes of each instruction
```  

When `e?` is given incomplete key/var name. Notice the missing 's': `byte` instead of `bytes`:
```
[0x00006ab0]> e?asm.byte
Key not found. Try e??asm.byte
```
Will always suggest to try `e??`  

The `e??` will supply any key and descriptions to somewhat similar to available keys. Again notice the missing 's':
```
[0x00006ab0]> e??asm.byte
           asm.bytes: display the bytes of each instruction
     asm.bytes.align: align bytes at right (left padding space)
    asm.bytes.asbits: show instruction bits instead of bytes
   asm.bytes.opcolor: colorize bytes depending on opcode size + variant information
     asm.bytes.right: display the bytes at the right of the disassembly
     asm.bytes.space: separate hexadecimal bytes with a whitespace
```

`e??` will return nothing when supplied key is not even remotely similar to available keys:
```
[0x00006ab0]> e??asbvz
[0x00006ab0]>
```

trufae:
Your proposal sounds good to me 👍

dvertx:
Glad that you like it. That is how the current PR work :)

dvertx:
I think I botched this PR. Will close this.
Created a new PR #21969

PR #21969, opened 2023-06-27
Rework e? and e?? commands for better output

dvertx:

**Description**

 Reworked `e?` and `e??` commands to have these behavior:

When the `e?` command supplied with a correct key, gives out the (only correct) description:
```
[0x00006ab0]> e?asm.bytes
display the bytes of each instruction
```

When `e?` is given incomplete key/var name. Notice the missing 's': byte instead of bytes:
```
[0x00006ab0]> e?asm.byte
Key not found. Try e??asm.byte
```
Will always suggest to try `e??` even for non existent key

The `e??` will supply any key and descriptions when given something similar to available keys. Again notice the missing 's':
```
[0x00006ab0]> e??asm.byte
           asm.bytes: display the bytes of each instruction
     asm.bytes.align: align bytes at right (left padding space)
    asm.bytes.asbits: show instruction bits instead of bytes
   asm.bytes.opcolor: colorize bytes depending on opcode size + variant information
     asm.bytes.right: display the bytes at the right of the disassembly
     asm.bytes.space: separate hexadecimal bytes with a whitespace
```

Another `e??` example:
```
[0x00006ab0]> e??asm.b
        asm.bbmiddle: realign disassembly if a basic block starts in the middle of an instruction
            asm.bits: word size in bits at assembler
           asm.bytes: display the bytes of each instruction
     asm.bytes.align: align bytes at right (left padding space)
    asm.bytes.asbits: show instruction bits instead of bytes
   asm.bytes.opcolor: colorize bytes depending on opcode size + variant information
     asm.bytes.right: display the bytes at the right of the disassembly
     asm.bytes.space: separate hexadecimal bytes with a whitespace
```

`e??` will return nothing when supplied with a key which is not even remotely similar to available keys:
```
[0x00006ab0]> e??asbvz
[0x00006ab0]>
```

### DIFF
--- a/libr/config/config.c
+++ b/libr/config/config.c
@@ -163,7 +163,7 @@ R_API void r_config_list(RConfig *cfg, const char *str, int rad) {
 	const char *sfx = "";
 	const char *pfx = "";
 	int len = 0;
-	bool verbose = false;
+	bool found, verbose = false;
 	PJ *pj = NULL;
 
 	if (!IS_NULLSTR (str)) {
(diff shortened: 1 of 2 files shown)


trufae:
Is this pr replacing the previous one?

dvertx:
Yes. Please merge this one if you think the patch is OK. The branch of the
previous PR somehow got mangled. My bad

PR #21974, opened 2023-06-27
Remove single quotes

dvertx:

**Description**

Copy command does not work because of single quotes
Current error message:, even though several .c files exists above current working directory:
```
[0x000010a0]> cp ../*c .
cp: cannot stat '../*c': No such file or directory
```

### DIFF
--- a/libr/util/file.c
+++ b/libr/util/file.c
@@ -1387,7 +1387,7 @@ R_API bool r_file_copy(const char *src, const char *dst) {
 #else
 	char *src2 = r_str_replace (strdup (src), "'", "\\'", 1);
 	char *dst2 = r_str_replace (strdup (dst), "'", "\\'", 1);
-	int rc = r_sys_cmdf ("cp -f '%s' '%s'", src2, dst2);
+	int rc = r_sys_cmdf ("cp -f %s %s", src2, dst2);
 	free (src2);
 	free (dst2);
 	return rc == 0;


libr/util/file.c
```diff
@@ -1387,7 +1387,7 @@ R_API bool r_file_copy(const char *src, const char *dst) {
 #else
 	char *src2 = r_str_replace (strdup (src), "'", "\\'", 1);
 	char *dst2 = r_str_replace (strdup (dst), "'", "\\'", 1);
-	int rc = r_sys_cmdf ("cp -f '%s' '%s'", src2, dst2);
+	int rc = r_sys_cmdf ("cp -f %s %s", src2, dst2);
```
trufae, line 1390: this is introducing a command injection vulnerability. so better not use r_sys_cmdf here imho. so if we have to replace this better do a pure native implementation without depending on `cp`

trufae:
Also, i dont think r_file_copy is suposed to be used with glob expansions. So the actual error you are reporting is correct. unless we implement that natively in the r2 api. the use of `cp` was just a workaround to get an api on top of that functionality

libr/util/file.c
dvertx: Yes, quite right, user input sanitation and all

dvertx:
If not meant for glob expansion, I might implement it as a simple 'single file' copy function,if you want

trufae:
Yeah this api is not suposed to handle glob expressions

PR #21979, opened 2023-06-28
[fix] Free a null pointer

Pikachu1412:

**Description**
Add the check before free

### DIFF
--- a/libr/main/rasm2.c
+++ b/libr/main/rasm2.c
@@ -751,7 +751,7 @@ R_API int r_main_rasm2(int argc, const char *argv[]) {
 	}
 	R_FREE (log_level);
 	RAsmState *as = __as_new ();
-	if(!as){
+	if (!as) {
 		return 0;
 	}
 	// TODO set addrbytes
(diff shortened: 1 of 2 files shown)


libr/util/unum.c
```diff
@@ -108,7 +108,9 @@ R_API RNum *r_num_new(RNumCallback cb, RNumCallback2 cb2, void *ptr) {
 }
 
 R_API void r_num_free(RNum *num) {
-	free (num);
+	if (num) {
+		free (num);
```
trufae, line 112: free(NULL) is valid, no need to check for null twice

libr/main/rasm2.c
```diff
@@ -748,7 +748,9 @@ R_API int r_main_rasm2(int argc, const char *argv[]) {
 	}
 	R_FREE (log_level);
 	RAsmState *as = __as_new ();
-
+	if (!as) {
```
trufae, line 751: 👍

libr/util/unum.c
Pikachu1412: You are right. It's not a problem, sorry.

PR #21981, opened 2023-06-28
Update ls.c

Pikachu1412:

**Description**
`list` may be a null ptr, we should check it before use.


Pikachu1412:
resolve https://github.com/radareorg/radare2/issues/21982#issue-1778738489

Pikachu1412:
fixes #21981

PR #22006, opened 2023-07-05
WIP: Initial rework on the rbin new API

trufae:

**Description**

**Copilot**

copilot:all


libr/include/r_bin.h
```diff
@@ -282,8 +282,14 @@ typedef struct r_bin_object_t {
 	HtUP *addr2klassmethod;
 	void *bin_obj; // internal pointer used by formats
 	bool is_reloc_patched; // used to indicate whether relocations were patched or not
+	// the new stuff
+	void *priv; // RBinObjectPriv
```
luc-tielen, line 286: You could make this `RBinObjectPriv`too? Forward declare it before RBinObject, and then define it in a C file somewhere.
Only difference is that you will need extra helper functions (but that's good, we don't want people directly accessing it)

PR #22027, opened 2023-07-07
Fix x86 XOR ESIL for 64bit registers ##esil 

aemmitt-ns:

**Description**

The ESIL expressions for XOR instructions with 64 bit registers is currently wrong due to the get64from32 function returning strdup(s) instead of NULL when the register does not match a 32bit register. This change lead to the use of the 32bit ESIL instead of the correct 64bit version. 

The change in this PR simply has get64from32 return NULL when the register is already 64bit, ensuring the right ESIL is generated for the instruction. 

**Copilot**

copilot:all

### DIFF
--- a/libr/arch/p/x86/plugin_cs.c
+++ b/libr/arch/p/x86/plugin_cs.c
@@ -419,7 +419,7 @@ static char *get64from32(const char *s) {
 			return r_str_newf ("r%d", atoi (s + 1));
 		}
 	}
-	return NULL; // strdup (s);
+	return NULL; 
 }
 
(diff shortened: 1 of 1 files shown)


libr/arch/p/x86/plugin_cs.c
```diff
@@ -419,7 +419,7 @@ static char *get64from32(const char *s) {
 			return r_str_newf ("r%d", atoi (s + 1));
 		}
 	}
-	return strdup (s);
+	return NULL;
```
trufae, line 422:
```suggestion
	return NULL;
```

trufae:
That was introduced during the arch regression. We probably would find it with more tests, but porting the x86 plugin was the hardest part so i'm happy that the esil is taking more care now that arch and cs5 are out!

Good patch and great contrib <3

PR #22030, opened 2023-07-08
Fix #22029 ESIL for REP[Z] [BRANCH INSTR] ##esil 

### LINKED ISSUE #22029: repz ret ESIL
## Environment

Sat Jul  8 00:56:36 EDT 2023
radare2 5.8.9 31408 @ darwin-arm-64
birth: git.5.8.8-255-g5bae28e7cc 2023-07-07__16:18:46
commit: 5bae28e7cc6e9f17f6e089ab003b7d1b08acac1e
options: gpl -O2 cs:5 cl:2 make
Darwin arm64

## Description

In capstone 5 the bytes f3c3 are decoded as `repz ret` instead of `ret`. This leads the x86 analysis to produce the ESIL `rcx,!,?{,BREAK,},rsp,[8],rip,=,8,rsp,+=,rcx,--=,0,GOTO` which is incorrect as it should still be treated as a simple `ret`. Most significantly when `rcx` is 0 the ret instruction is effectively skipped leading to false control flow. 

I will fix this in the morning. Not sure how yet.

aemmitt-ns:

**Description**

Fix #22029 an issue with `repz ret` instruction which should not actually have the ESIL for the REP prefix as it is undefined for non-string instructions like `ret`. The `repz` prefix only serves to create a single 2 byte `ret` instruction for branch prediction optimization as noted [here](https://repzret.org/p/repzret/). Specifically these changes remove the `rep` ESIL for any of the control flow instruction types also excluded from `repne` already.

### DIFF
--- a/libr/arch/p/x86/plugin_cs.c
+++ b/libr/arch/p/x86/plugin_cs.c
@@ -2452,8 +2452,7 @@ static void anop_esil(RArchSession *as, RAnalOp *op, ut64 addr, const ut8 *buf, 
 	}
 
 	// AMD K8 optimization lead some compilation to emit REPZ RET which should be treated as RET
-	ut32 ignore_op_mask = (R_ANAL_OP_TYPE_UJMP | R_ANAL_OP_TYPE_CALL | R_ANAL_OP_TYPE_RET) & ~1;
-	if (op->prefix & R_ANAL_OP_PREFIX_REP && !(op->type & ignore_op_mask)) {
+	if (op->prefix & R_ANAL_OP_PREFIX_REP && op->type != R_ANAL_OP_TYPE_RET) {
 		r_strbuf_prepend (&op->esil, ",!,?{,BREAK,},");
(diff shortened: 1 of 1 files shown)


aemmitt-ns:
hmm looks like it broke some `rep ...` instructions that it shouldn't have. gonna mark as draft and fix

aemmitt-ns:
i feel like the repne ESIL exclusion is also probably wrong for the same reason 

`(R_ANAL_OP_TYPE_MOV & (R_ANAL_OP_TYPE_UJMP | R_ANAL_OP_TYPE_CALL | R_ANAL_OP_TYPE_RET)) != 0`

but idk how often this is emitted. i assume this fix is here for a reason so i am gonna leave it alone.

aemmitt-ns:
oh damn i wrote that code

trufae:
Good finding! Definitely this was a regression after updating to capstone5! Thanks for taking care of this issue and testing it properly!

PR #22057, opened 2023-07-14
Ioqjs

trufae:

**Description**

**Copilot**

copilot:all

### DIFF
--- a/libr/include/r_io.h
+++ b/libr/include/r_io.h
@@ -81,8 +81,8 @@ typedef struct r_io_undos_t {
 } RIOUndos;
 
 typedef struct r_io_undo_t {
-	int s_enable;
-	int w_enable;
+	int s_enable; // R2_590 - can be bool imho
+	int w_enable; // R2_590 - can be bool
 	/* write stuff */
 	RList *w_list;
 	int w_init;
@@ -90,8 +90,7 @@ typedef struct r_io_undo_t {
 	int idx;
 	int undos; /* available undos */
 	int redos; /* available redos */
-	RIOUndos seek[R_IO_UNDOS];
-	/*int fd[R_IO_UNDOS]; // XXX: Must be RIODesc* */
+	RIOUndos seek[R_IO_UNDOS]; // XXX R2_590 - use an RVec here
 } RIOUndo;
 
 typedef struct r_io_undo_w_t {
@@ -192,6 +191,7 @@ typedef struct {
 } RIORap;
 
 typedef struct r_io_plugin_t {
+	// RPluginMeta
 	const char *name;
 	const char *desc;
 	const char *version;
@@ -204,18 +204,21 @@ typedef struct r_io_plugin_t {
 	RIOUndo undo;
 	bool isdbg;
 	// int (*is_file_opened)(RIO *io, RIODesc *fd, const char *);
-	char *(*system)(RIO *io, RIODesc *fd, const char *);
+	char *(*system)(RIO *io, RIODesc *fd, const char *); // Rename to call? or cmd? unify with anal and core
 	RIODesc* (*open)(RIO *io, const char *, int perm, int mode);
 	RList* /*RIODesc* */ (*open_many)(RIO *io, const char *, int perm, int mode);
 	int (*read)(RIO *io, RIODesc *fd, ut8 *buf, int count);
 	ut64 (*seek)(RIO *io, RIODesc *fd, ut64 offset, int whence);
 	int (*write)(RIO *io, RIODesc *fd, const ut8 *buf, int count);
 	bool (*close)(RIODesc *desc);
+	// maybe just have a getinfo() that returns this struct
+	// RIOInfo * = struct { isblock, ischar, pid, tid, base, size }
 	bool (*is_blockdevice)(RIODesc *desc);
 	bool (*is_chardevice)(RIODesc *desc);
 	int (*getpid)(RIODesc *desc);
 	int (*gettid)(RIODesc *desc);
 	bool (*getbase)(RIODesc *desc, ut64 *base);
+	///
 	bool (*resize)(RIO *io, RIODesc *fd, ut64 size);
 	int (*extend)(RIO *io, RIODesc *fd, ut64 size);
 	bool (*accept)(RIO *io, RIODesc *desc, int fd);
--- a/libr/lang/p/qjs.c
+++ b/libr/lang/p/qjs.c
@@ -33,14 +33,29 @@ typedef struct qjs_arch_plugin_t {
 	// JSValue encode_func;
 } QjsArchPlugin;
 
+typedef struct qjs_io_plugin_t {
+	char *name;
+	RIOPlugin *iop;
+	R_BORROW JSContext *ctx;
+	JSValue fn_check_js;
+	JSValue fn_open_js;
+	JSValue fn_seek_js;
+	JSValue fn_read_js;
+	JSValue fn_close_js;
+	JSValue fn_system_js;
+	// JSValue encode_func;
+} QjsIoPlugin;
+
 R_GENERATE_VEC_IMPL_FOR(ArchPlugin, QjsArchPlugin);
+R_GENERATE_VEC_IMPL_FOR(IoPlugin, QjsIoPlugin);
 
 typedef struct qjs_plugin_manager_t {
 	R_BORROW RCore *core;
 	R_BORROW JSRuntime *rt;
-	QjsContext default_ctx;  // context for running normal JS code
+	QjsContext default_ctx; // context for running normal JS code
 	RVecCorePlugin core_plugins;
 	RVecArchPlugin arch_plugins;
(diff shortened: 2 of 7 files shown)


libr/include/r_io.h
```diff
@@ -192,6 +191,7 @@ typedef struct {
 } RIORap;
 
 typedef struct r_io_plugin_t {
+	// RPluginMeta
```
luc-tielen, line 194: mark as R2_590, so we can change it in a future PR

libr/include/r_io.h
```diff
@@ -204,18 +204,21 @@ typedef struct r_io_plugin_t {
 	RIOUndo undo;
 	bool isdbg;
 	// int (*is_file_opened)(RIO *io, RIODesc *fd, const char *);
-	char *(*system)(RIO *io, RIODesc *fd, const char *);
+	char *(*system)(RIO *io, RIODesc *fd, const char *); // Rename to call? or cmd? unify with anal and core
```
luc-tielen, line 207: Agreed

libr/include/r_io.h
```diff
@@ -204,18 +204,21 @@ typedef struct r_io_plugin_t {
...
 	RIODesc* (*open)(RIO *io, const char *, int perm, int mode);
 	RList* /*RIODesc* */ (*open_many)(RIO *io, const char *, int perm, int mode);
 	int (*read)(RIO *io, RIODesc *fd, ut8 *buf, int count);
 	ut64 (*seek)(RIO *io, RIODesc *fd, ut64 offset, int whence);
 	int (*write)(RIO *io, RIODesc *fd, const ut8 *buf, int count);
 	bool (*close)(RIODesc *desc);
+	// maybe just have a getinfo() that returns this struct
```
luc-tielen, line 214: good idea!

libr/lang/p/qjs.c
```diff
@@ -71,17 +87,43 @@ static void plugin_manager_add_core_plugin(QjsPluginManager *pm, const char *nam
...
+static QjsIoPlugin *plugin_manager_add_io_plugin(QjsPluginManager *pm, const char *name, JSContext *ctx, RIOPlugin *iop, JSValue func) {
+	r_return_val_if_fail (pm, NULL);
+
+	QjsIoPlugin *cp = RVecIoPlugin_emplace_back (&pm->io_plugins);
+	if (cp) {
+		cp->name = name? strdup (name): NULL;
+		cp->ctx = ctx;
+		cp->iop = iop;
+		cp->fn_check_js = func;
+		// cp->qctx.open_func = func;
```
luc-tielen, line 99: not filled in yet? or are you trying to get MVP up and running first?

libr/lang/p/qjs.c
```diff
@@ -71,17 +87,43 @@ static void plugin_manager_add_core_plugin(QjsPluginManager *pm, const char *nam
 	}
 }
 
+static QjsIoPlugin *plugin_manager_add_io_plugin(QjsPluginManager *pm, const char *name, JSContext *ctx, RIOPlugin *iop, JSValue func) {
```
luc-tielen, line 90: no remove IO plugin yet?

libr/lang/p/qjs/example-io.r2.js
```diff
@@ -0,0 +1,39 @@
+/* */
```
luc-tielen, line 1: either remove or fill in comment

libr/lang/p/qjs/io.c
```diff
@@ -0,0 +1,258 @@
...
+	QjsPluginManager *pm = R_UNWRAP4 (core, lang, session, plugin_data);
+
+	// Iterate over plugins until one returns "true" (meaning the plugin handled the input)
+	QjsIoPlugin *plugin;
+	R_VEC_FOREACH (&pm->io_plugins, plugin) {
+		JSContext *ctx = plugin->ctx;
+		JSValueConst args[1] = { JS_NewString (ctx, input) };
+		JSValue res = JS_Call (ctx, plugin->fn_system_js, JS_UNDEFINED, countof (args), args);
+		if (JS_ToBool (ctx, res)) {
+			return strdup ("true");
```
luc-tielen, line 160: this is what IO plugins usually return for a call or is this a hack?

libr/lang/p/qjs/io.c
```diff
@@ -0,0 +1,258 @@
...
+			RIOPlugin *iop = plugin->iop;
+			return r_io_desc_new (io, iop, name, perm, mode, core);
+		}
+	}
+
+	return NULL;
+}
+
+// TODO maybe add a function to call by plugin name? (is 1 extra arg)
+static char *qjs_io_system(RIO *io, RIODesc *fd, const char *input) {
```
luc-tielen, line 149: rename to qjs_io_call

libr/lang/p/qjs/io.c
```diff
@@ -0,0 +1,258 @@
...
+
+	// check if res is an object
+	JSValue res = JS_Call (ctx, argv[1], JS_UNDEFINED, countof (args), args);
+	if (!JS_IsObject (res)) {
+		return JS_ThrowRangeError (ctx, "r2.plugin function must return an object");
+	}
+
+	JSValue name = JS_GetPropertyStr (ctx, res, "name");
+	size_t namelen;
+	const char *nameptr = JS_ToCStringLen2 (ctx, &namelen, name, false);
```
luc-tielen, line 185: This string is managed by qjs right? (In case we do an early return on an error later on..)

libr/lang/p/qjs/io.c
```diff
@@ -0,0 +1,258 @@
...
+	if (!ap) {
+		return JS_ThrowRangeError (ctx, "could not allocate qjs core plugin");
+	}
+
+	JSValue desc = JS_GetPropertyStr (ctx, res, "desc");
+	JSValue license = JS_GetPropertyStr (ctx, res, "license");
+	const char *descptr = JS_ToCStringLen2 (ctx, &namelen, desc, false);
+	const char *licenseptr = JS_ToCStringLen2 (ctx, &namelen, license, false);
+
+	ap->name = nameptr;  // TODO no strdup here?
```
luc-tielen, line 223: Depends on if it nameptr is managed by qjs, but imo I would do it, and then just free it in the fini function of QjsIoPlugin

libr/lang/p/qjs/io.c
```diff
@@ -0,0 +1,258 @@
...
+
+	ap->check = qjs_io_check;
+	ap->open = qjs_io_open;
+	ap->read = qjs_io_read;
+	ap->seek = qjs_io_seek;
+#if 0
+	ap->check = qjs_io_check;
+	ap->close = qjs_io_close;
+#endif
+	ap->system = qjs_io_system;  // Technically this could all be handled by a single generic plugin
```
luc-tielen, line 239: Maybe we should do this refactor, just have a single generic "system" IO plugin (could be baked into r2)

libr/include/r_io.h
trufae: i did not marked it because it feels like we are breaking so much stuff and this will require patching more projects xD like r2frida, but yeah ideally i would like to have all the plugins this way.

libr/include/r_io.h
trufae: but im afraid to break more apis when we didnt finished stabilizing the others!

libr/lang/p/qjs.c
```diff
@@ -33,14 +33,29 @@ typedef struct qjs_arch_plugin_t {
...
+typedef struct qjs_io_plugin_t {
+	char *name;
+	RIOPlugin *iop;
+	R_BORROW JSContext *ctx;
+	JSValue fn_check_js;
+	JSValue fn_open_js;
+	JSValue fn_seek_js;
+	JSValue fn_read_js;
+	JSValue fn_close_js;
+	JSValue fn_system_js;
```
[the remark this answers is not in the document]
trufae: will do when we decide the right name for this for consistency i keep the original function name used in the plugin struct. Shall we use "run", "call", "cmd" .. ?

libr/lang/p/qjs.c
trufae: its returning the object and filling the fields later, otherwise ill need to add over9000 parameters to this function and change the signature everytime i update the api to support more callbacks

libr/lang/p/qjs/io.c
trufae: no. thats just a hack and should be fixed, io plugin system() return a string with the output of the command. like in r2pipe

libr/lang/p/qjs/io.c
trufae: only if we change the 90 io plugins to use this convention

trufae:
i agree on the refactoring changes discussed in the comments, but this must be done in separate PR, mainly because that affects other projects and plugins and i dont wanna jump into more breaking refactorings when we have some in our plate already :D

PR #22073, opened 2023-07-18
Minor fixes

luc-tielen:

**Description**

**Copilot**

copilot:all


libr/util/udiff.c
```diff
@@ -273,6 +273,12 @@ R_API char *r_diff_buffers_unified(RDiff *d, const ut8 *a, int la, const ut8 *b,
 	}
 	if (!fa || !fb) {
 		R_LOG_ERROR ("fafb nul");
+		if (fa) {
```
trufae, line 276: No need to if(x)free(x)

free accepts null

luc-tielen: it's still early, forgot that 😂 will change

PR #22114, opened 2023-07-31
Swiss xrefs

luc-tielen:

**Description**

**Copilot**


trufae:
about the linter:

```
libr/include/cwisstable.h:891:  size_t value = ++counter;
libr/include/cwisstable.h:2185:  for (size_t i = 0; i <= ctrl_bytes; ++i) {
```

(size_t must be moved before the for, and dont use a preincrement

trufae:
Screenshot 2023-08-01 at 00 53 02

luc-tielen:
@trufae I fixed linting issues, but the screenshot you posted I think are all false positives due to not having __uint128_t on another compiler that isn't TCC.

Bigger problem I think is Windows.. I think we need atleast this version of the SDK to compile atomics correctly?: https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/

luc-tielen:
@trufae only windows is broken, probably due to some things missing (such as `alignas`), probably could be fixed by ifdef'ing the code that fails, and using something like: https://code.videolan.org/videolan/dav1d/-/blob/master/include/common/attributes.h#L82

trufae:
I will take care of the windows bits :) amazing!

PR #22123, opened 2023-08-02
Fix arrow keys on some terminals ##cons

McKael:

**Description**

Some terminals (xterm, tmux, screen...) use the escape sequences `^[{A,B,C,D}` for the arrow keys.
This M.R. adds support for those escape sequences.

Additionally, it fixes the behavior of Ctrl-L which does a cursor move before clearing the screen.

### DIFF
--- a/libr/cons/dietline.c
+++ b/libr/cons/dietline.c
@@ -1783,6 +1783,9 @@ R_API const char *r_line_readline_cb(RLineReadCallback cb, void *user) {
 					buf[1] = r_cons_readchar_timeout (50);
 				}
 #endif
+				if (buf[0] == 'O' && strchr("ABCDFH", buf[1]) != NULL) { // O
+					buf[0] = '[';
+				}
 				if (buf[0] == '[') { // [
 					switch (buf[1]) {
 					case '3': // supr or mouse click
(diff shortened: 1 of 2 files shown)


trufae:
i can confirm arrow keys and ^L work as expected on all the terminals you mentioned in the issue. so im unsure what are you fixing

McKael:
Hi @trufae,  thanks for commenting,

> i can confirm arrow keys and ^L work as expected on all the terminals you mentioned in the issue. so im unsure what are you fixing

Regarding the ^L "fix", here's a way to reproduce it:
- run `r2 -`
- type a string, e.g. "foobar"
- move the cursor to the left, e.g. ^A to go to the first column
- hit ^L one or several times


For the arrow keys, I think it may be related to my terminal being stuck in "application mode", and I believe it might be related to a conflict between zsh system configuration and my own configuration (Debian sets the terminal in _application mode_ for zsh's Line Editor and it keeps turned on).

I guess I can close the issue.  Please let me know if you can reproduce the Ctrl-L weirdness and if it deserves a separate PR.

trufae:
sorry for the super duper later response. i have tested your reproducer now and i can say your patch fixes it. i will merge it via this PR https://github.com/radareorg/radare2/pull/22789 which i have just rebase and resolved conflicts.

PR #22125, opened 2023-08-03
Swiss htuu / htpu / htsu

luc-tielen:

**Description**

This refactors the HtUU hashmap type to use swisstable as the backing hashmap.
Ideally, this code should be upstreamed to sdb (with only the r2 fixes in this repo). 
**Copilot**


shlr/sdb/src/ht_uu.c
```diff
@@ -1,19 +1,112 @@
...
+SDB_API bool ht_uu_insert(HtUU *hm, const ut64 key, ut64 value) {
+	assert (hm);
+
+	HtUU_Entry entry = { .key = key, .val = value };
+	HtUU_Insert result = HtUU_insert (hm, &entry);
+	return result.inserted;
+}
+
+SDB_API bool ht_uu_update(HtUU *hm, const ut64 key, ut64 value) {
+	assert (hm);
```
trufae, line 42: dont use assert. libraries shuold never crash. use r_return_val_if_fail... but those macros shuold be moved into sdb and used there.

also we differentiate from runtime and compile time asserts. this iis a compile time chk because by running the tests we can find if this function was called with null somwehre.

shlr/sdb/src/ht_uu.c
```diff
@@ -1,19 +1,112 @@
...
-		.hashfn = NULL,
-		.dupkey = NULL,
-		.dupvalue = NULL,
-		.calcsizeK = NULL,
-		.calcsizeV = NULL,
-		.freefn = NULL
-	};
-	return Ht_(new_opt) (&opt);
+#include "sdb/heap.h"
+#include "../../../libr/include/cwisstable.h"  // XXX use cwisstable from sdb itself
```
trufae, line 6: maybe better to move this into sdb and merge it and use it later when done :)

trufae: also update the copyright header line, add your name and use 2023

trufae (changes requested):
asdf

shlr/sdb/src/ht_uu.c
luc-tielen: Yeah I know, only used assert because that macro wasn't available there yet.

shlr/sdb/src/ht_uu.c
luc-tielen: Yep. This is just something for me to iterate easily on the code from r2.

shlr/sdb/src/ht_uu.c
```diff
@@ -1,19 +1,112 @@
...
+#pragma GCC diagnostic ignored "-Wtypedef-redefinition"
+#endif
+CWISS_DECLARE_FLAT_HASHMAP(HtUU, ut64, ut64);
+#if defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
+
+SDB_API HtUU* ht_uu_new0(void) {
+	HtUU *hm = sdb_gh_calloc (1, sizeof (HtUU));
+	assert (hm);
```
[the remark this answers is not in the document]
luc-tielen: You're right, will change.

PR #22150, opened 2023-08-14
Remove global state in visual UI ##core

luc-tielen:

**Description**

**Copilot**

### DIFF
--- a/libr/include/r_core.h
+++ b/libr/include/r_core.h
@@ -232,10 +232,14 @@ typedef struct r_core_visual_t {
 	int current3format;
 	int current4format;
 	int current5format;
-	RConfigHold *hold;  // XXX should be a tab-specific var
+	RConfigHold *hold;  // TODO should be a tab-specific var
 	ut64 oldpc;
 	ut64 oseek;
 	char debugstr[512];
+
+	bool firstRun;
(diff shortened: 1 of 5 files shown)


libr/include/r_core.h
```diff
@@ -209,6 +209,33 @@ typedef struct r_core_visual_t {
 	RCoreVisualMode printidx;
 	RList *tabs;
 	int tab;
+	bool textedit_mode;
+	int obs;
+	bool __ime;
```
trufae, line 214: probably worth removing the __

libr/include/r_core.h
```diff
@@ -209,6 +209,33 @@ typedef struct r_core_visual_t {
...
+	int currentFormat;
+	int current0format;
+	char numbuf[32];
+	int numbuf_i;
+	bool splitView;
+	ut64 splitPtr;
+	int current3format;
+	int current4format;
+	int current5format;
+	RConfigHold *hold;  // XXX should be a tab-specific var
```
trufae, line 235: wanna try to make this change, also this is more like TODO than XXX

luc-tielen: The comment is very vague for me right now, I'd need more explanation. I can make it a TODO :)

PR #22156, opened 2023-08-14
ARM branching bugfix

sha0coder:
On a "bx lr" it was not considering the LSB of anddress on LR

```asm
...
284 0x0007414c push {r0, r1, lr}
285 0x0007414e bl 0x740cc    ---------- call
286 0x000740cc movs r3, 1
287 0x000740ce movs r2, 0
288 0x000740d0 push {r4}
289 0x000740d2 cmp r0, r1
290 0x000740d4 blo 0x74130
291 0x00074130 adds r0, r2, 0
292 0x00074132 pop {r4}
293 0x00074134 bx lr        -------------- return to lr
esil bug triggered
> aer pc
0x00074150
                                                                                                        
> aer lr                                                                                                
0x00074152                                                                                              
                                                                                                        
> pi 1                                                                                                  
vdup.8 d11, d14[6]  
```
is returning to LR -2 instead of LR because is not thumb mode:
>>> bin(0x00074152)
'0b1110100000101010010'


libr/arch/p/arm/plugin_cs.c
```diff
@@ -2507,7 +2507,7 @@ static int analop_esil(RArchSession *as, RAnalOp *op, ut64 addr, const ut8 *buf,
 			if (ISIMM (0)) {
 				r_strbuf_appendf (&op->esil, "%s,pc,:=", ARG (0));
 			} else {
-				r_strbuf_appendf (&op->esil, "%d,%s,-,pc,:=", thumb, ARG (0));
+				r_strbuf_appendf (&op->esil, "%s,pc,:=,-1,1,<<,pc,&", ARG (0));
```
trufae, line 2510: i think it should be this way
```suggestion
				r_strbuf_appendf (&op->esil, "-1,1,<<,%s,&=", ARG (0));
```

sha0coder:
this works for ARM32 but should be tested on ARM16:

```c
    case ARM_INS_BLX:
        r_strbuf_appendf (&op->esil, "1,%s,&,tf,:=", ARG (0));
```
```c
case ARM_INS_B: 
        if (ISREG (0) && REGID (0) == ARM_REG_PC) {
            r_strbuf_appendf (&op->esil, "0x%" PFMT64x ",pc,:=",
                (ut64)((addr & ~3LL) + pcdelta));
        } else {
            if (ISIMM (0)) {
                r_strbuf_appendf (&op->esil, "%s,pc,:=", ARG (0));
            } else {
                r_strbuf_appendf (&op->esil, "tf,%s,-,pc,:=", ARG (0));
            }
        }
        break;
```

trufae:
Can you please rebae this pr? i think one of the prlblems is that the tf register is not resetted when emulating different codepaths, causing wrong assumptions in other situations.

PR #22168, opened 2023-08-17
Fix UB in the loongarch decoder

trufae:

**Description**

**Copilot**

### DIFF
--- a/libr/arch/p/loongarch/loongarch-coder.c
+++ b/libr/arch/p/loongarch/loongarch-coder.c
@@ -110,8 +110,13 @@ loongarch_decode_imm (const char *bit_field, insn_t insn, int si)
 
   if (si)
     {
-      ret <<= sizeof (ret) * 8 - len;
-      ret >>= sizeof (ret) * 8 - len;
+      int sh = sizeof (ret) * 8 - len;
+      if (sh >= 31) {
+	ret = 0;
+      } else {
+        uint32_t t = (ret << sh);
+        ret = (t >> sh);
+      }
     }
   return ret;
 }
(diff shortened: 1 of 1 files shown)


trufae:
Hey @junchao-loongson i have found this UB via asan, but attempting to fix the (too much shifting to the left, and also, on a signed variable) causes some troubles. By reading the output of the tests seems like it was not behaving well before, there are some calls to negative addresses and adds with negative numbers..

I don't know much this arch, and i have my head in other stuff, so i would appreciate if you could take a look at this and provide a proper fix.

Here's the failing test.

Screenshot 2023-08-17 at 17 35 42

junchao-loongson:
hello~ I'm not sure I understand what UB means. Can you give me a link to the relevant information
Do you mean that the following code may cause security problems?
```
      ret <<= sizeof (ret) * 8 - len;
      ret >>= sizeof (ret) * 8 - len;
```

junchao-loongson:
okay，i got it, thanks. I'll handle it.

junchao98:
Under normal circumstances, len is always less than 32, because loongarch's instructions are encoded as 32 bit.
we can add a assert()
```
assert(sizeof(ret) * 8 > len);
ret <<= sizeof (ret) * 8 - len;
ret >>= sizeof (ret) * 8 - len;
```

junchao-loongson:
Under normal circumstances, len is always less than 32, because loongarch's instructions are encoded as 32 bit.
we can add a assert()
```
assert(sizeof(ret) * 8 > len);
ret <<= sizeof (ret) * 8 - len;
ret >>= sizeof (ret) * 8 - len;
```
This operation is designed to do symbolic expansion of data of the specified bit length

trufae:
Adding an assert will make r2 to crash. Consider dimming the value or so

PR #22172, opened 2023-08-17
WIP: Remove global from bin.kernelcache ##bin

trufae:

**Description**

**Copilot**


libr/bin/p/bin_xnu_kernelcache.c
```diff
@@ -1979,22 +1982,22 @@ static int kernelcache_io_read(RIO *io, RIODesc *fd, ut8 *buf, int count) {
...
 					}
 				}
 			}
 			break;
 		}
 	}
 
-	if (!cache) {
-		r_list_foreach (pending_bin_files, iter, bf) {
+	if (!cache && cache->pending_bin_files) {
```
mrmacete, line 1999: i think we have a chicken and egg problem here, if `cache` is null we can't search for it in `cache->pending_bin_files`... that's the reason why this was global

PR #22191, opened 2023-08-22
Fix mnemonic output in m680x plugin ##fix

mmuman:

Seems r_str_replace_in wants the buffer size, not the string size, so it was truncating the last character.

### DIFF
--- a/libr/arch/p/m680x_cs/plugin.c
+++ b/libr/arch/p/m680x_cs/plugin.c
@@ -513,7 +513,7 @@ beach:
 		} else {
 			op->mnemonic = r_str_newf ("%s%s%s", insn->mnemonic,
 					insn->op_str[0]?" ": "", insn->op_str);
-			r_str_replace_in (op->mnemonic, strlen (op->mnemonic),
+			r_str_replace_in (op->mnemonic, strlen (op->mnemonic) + 1,
 				"ptr ", "", true);
 		}
 	}
--- a/test/db/extras/asm/m680x_6805_8
+++ b/test/db/extras/asm/m680x_6805_8
@@ -1,0 +1,2 @@
+d "decx" 5A
+d "inc $ae" 3CAE


trufae:
can you add a test for this?

mmuman:
I did add a test after opening the PR (hence the force push), spent some time figuring out how to do that.

Ah ok for the style, I looked at other similar code and saw a "+1" without space but I guess not all files are up to date on that.

mmuman:
Fixed style; mentioned the test in commit log.

mmuman:
did it just fail randomly? could try to force push again…

trufae:
Yes this is a race condition bug in meson. I reported it a while ago and its hard to reproduce in real hardware. Mainly fails on the ci. But it was ignored and not fixed so we live with that :) looks good to me but i want to test a thing before merging . Sorry for the slow response, this week im quite busy irl

mmuman:
It seems the backend doesn't classify opcodes correctly, it looks like half finished on that… but I can open another PR later.

trufae:
i'm back :) ill answer faster now. ive seen other inconsistencies in this disassembler. 

for example

```
fcb $3
fcb $a
```

in r2, we use a unified syntax for all the disassembly, and this is, special chars have no room in the default asm.syntax=intel, so i would like to trim those '$' '%' '#' chars if possible (you can check the config->syntax for this)

would you like to work on this too?

mmuman:
I could have a look but I'd first want to have it usable. I want to get a disasm of several versions of a 68HC05 firmware, either straight from r2 or using r2 to list the symbols of each version to tell dasmxx where to look.

Actually it seems Capstone has syntax options, so maybe it's just about checking `asm.syntax` and passing the right one to `cs_option()` ?

trufae:
Yes. These options should be honored via the values in the arch config. But i dont think this is handled for this arch in capstone. So i would suggest to do some str.replace in the r2 plugin, and maybe in the future submit a fix for capstone to handle this in their side, but this may take longer and will require code in r2 to handle all versions of capstone.

mmuman:
Hmm ok but how do you differentiate `lda 12` from `lda 12`? :thinking:

mmuman:
Let's continue on #22196?

PR #22196, opened 2023-08-26
M680x fixes

mmuman:

Some pending fixes for the m680x plugin.


trufae:
Please rebase, those tests seems to work in local but not in the ci, so i marked them as broken for now

libr/arch/p/m680x_cs/plugin.c
```diff
@@ -515,6 +515,14 @@ static bool decode(RArchSession *as, RAnalOp *op, RAnalOpMask mask) {
 					insn->op_str[0]?" ": "", insn->op_str);
 			r_str_replace_in (op->mnemonic, strlen (op->mnemonic) + 1,
 				"ptr ", "", true);
+			// we don't really output operands in intel order but we can clean up
+			if (as->config->syntax == R_ARCH_SYNTAX_INTEL) {
```
trufae, line 519: uhm, is this $ an immediate and # a load on the given address? because if its a load we can use [addr] instead

mmuman:
I didn't touch the tests yet. Actually which syntax are they run with anyway?

libr/arch/p/m680x_cs/plugin.c
mmuman: Oh it's actually just like for 6502, # is immediate, and $ is for hex, which is the easy part.

I could go like replace all the "lda " and other insns that accept immediate with "lda [", then replace "[#" by "", then append the missing ] to those lines that still have the [ …

mmuman:
I made a naive version of the syntax stripper, but there are probably many other insns involved and side effects, like with multiple operands…

libr/arch/p/m680x_cs/plugin.c
trufae: yep, that may work

trufae:
it is ready to review/merge?

mmuman:
I'd like to finish classifying opcodes at least.

trufae:
Sure, no hurry!

libr/arch/p/m680x_cs/plugin.c
```diff
@@ -515,6 +515,25 @@ static bool decode(RArchSession *as, RAnalOp *op, RAnalOpMask mask) {
 					insn->op_str[0]?" ": "", insn->op_str);
 			r_str_replace_in (op->mnemonic, strlen (op->mnemonic) + 1,
 				"ptr ", "", true);
+			// we don't really output operands in intel order but we can clean up
+			// XXX: this is not finished
+			if (as->config->syntax == R_ARCH_SYNTAX_INTEL) {
+				op->mnemonic = r_str_replace (op->mnemonic,
```
trufae, line 521: probably better to move this logic into a separate function, so op-> is only referenced once and not 10 times

libr/arch/p/m680x_cs/plugin.c
```diff
@@ -515,6 +515,25 @@ static bool decode(RArchSession *as, RAnalOp *op, RAnalOpMask mask) {
...
+					"lda ", "lda [", true);
+				op->mnemonic = r_str_replace (op->mnemonic,
+					"ldx ", "ldx [", true);
+				op->mnemonic = r_str_replace (op->mnemonic,
+					"and ", "and [", true);
+				op->mnemonic = r_str_replace (op->mnemonic,
+					"ora ", "ora [", true);
+				op->mnemonic = r_str_replace (op->mnemonic,
+					" [#", " ", true);
+				if (strstr(op->mnemonic, "[")) {
```
trufae, line 533: missing space before (

libr/arch/p/m680x_cs/plugin.c
```diff
@@ -525,13 +544,38 @@ static bool decode(RArchSession *as, RAnalOp *op, RAnalOpMask mask) {
...
-		"gpr	pc	.16	48	0\n"
-		"gpr	sp	.16	48	0\n"
-		"gpr	a0	.16	48	0\n"
-		"gpr	a1	.16	48	0\n";
+		"=SP    s\n"
+		"gpr	pc	.16	0	0\n" ///< M6800/1/2/3/9, M6301/9
+		"gpr	s	.16	2	0\n" ///< M6809/M6309 system stack (=sp on others)
+		"gpr	cc	.8	4	0\n" ///< M6800/1/2/3/9, M6301/9
+		"flg	C		.1	4.0	0\n"
+		"flg	V		.1	4.1	0\n" // At least 6805 lacks it? Are the others shifted??
```
trufae, line 552: two tabs. should be one

trufae (changes requested):

mmuman:
busy busy, I'll try to have a look next week.

trufae:
the next week has arrived, can i help you somehow?

libr/arch/p/m680x_cs/plugin.c
mmuman: Indeed, and it's easier on the indentation.

libr/arch/p/m680x_cs/plugin.c
mmuman: ok

libr/arch/p/m680x_cs/plugin.c
mmuman: I probably copied that from some other place… fixed.

mmuman:
Added some opcode types, but we should probably check them one by one, maybe sort the known ones by type instead, and move them from the alpha-sorted list gradually? We can probably do that later, I thought it was needed for function detection but it seems it's only useful for syntax coloring.

mmuman:
bleh, JSR is not relative, had a doubt at last minute :sweat_smile:

libr/arch/p/m680x_cs/plugin.c
mmuman: Still incomplete but seems to work for now.

mmuman:
ouch, bset != brset, it's not a jump…

libr/arch/p/m680x_cs/plugin.c
```diff
@@ -67,6 +67,21 @@ static inline csh cs_handle_for_session (RArchSession *as) {
 	return pd->cs_handle;
 }
 
+static inline char *filter_intel_syntax (char *mnemonic) {
```
trufae, line 70: i think its better not to inline this
```suggestion
static char *filter_intel_syntax(char *mnemonic) {
```

trufae (changes requested):
looks good to merge fter this small style fix. you can add more improvements in a separate pr

libr/arch/p/m680x_cs/plugin.c
mmuman: ok!

mmuman:
rebased

PR #22198, opened 2023-08-26
blub blub, here comes a core plugin

condret:

**Description**

**Copilot**

### DIFF
--- a/libr/core/p/core_agD.c
+++ b/libr/core/p/core_agD.c
@@ -1,0 +1,82 @@
+/* radare - LGPL3 - Copyright 2023 - condret */
+
+#include <r_core.h>
+#include <r_config.h>
+#include <r_cons.h>
+#include <r_anal.h>
+#include <r_util.h>
+
+static char *_get_title(void *data, void *user) {
+	return r_str_newf ("0x%"PFMT64x,
+		((RAnalBlock *)(((RGraphNode *)data)->data))->addr);
+}
+
+static char *_get_body(void *data, void *user) {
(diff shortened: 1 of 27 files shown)


libr/core/p/core_agD.c
```diff
@@ -0,0 +1,82 @@
...
+#else
+	.name = "agD",
+	.desc = "agD core plugin",
+	.license = "LGPL3",
+	.author = "condret",
+#endif
+	.call = r_cmd_agD_call,
+};
+
+#ifndef CORELIB
```
trufae, line 76:
```suggestion
#ifndef R2_CORELIB
```

libr/core/p/core_agD.c
```diff
@@ -0,0 +1,82 @@
...
+	r_agraph_print (dtagraph);
+	r_agraph_free (dtagraph);
+	r_graph_free (fcn_dtgraph);
+	r_graph_free (fcn_graph);
+	r_cons_flush ();
+	return true;
+}
+
+RCorePlugin r_core_plugin_agD = {
+#if R2_VERSION_NUMBER > 50808
```
trufae, line 60: Why not just in cmd_anal or an anal plug instead of core?

condret: because this is only doing this one specific command

condret: well, i don't consider this something actually useful, outside of debugging the domtree code. nobody needs to see a domtree. this is only useful for improving analysis

PR #22209, opened 2023-09-01
Fixed ESIL PPC32 lbz/lbzcix instructions to not update register.

phix33:
- [ X] Mark this if you consider it ready to merge
- [ X ] I've added tests (optional)

**Description**
The PPC32 lbz instruction was being treated as the lbzu instruction in ESIL so the 'source' register was being updated. Fixed to mirror the lwz/lwzu instructions and added some simple tests for lbz/lbzu/lwz/lwzu.

**Copilot**


trufae:
Thanks! the fix looks good

PR #22236, opened 2023-09-10
Fix relative doc link and perma link it

ChillerDragon:
When clicking the link on github in the CI section https://github.com/radareorg/radare2/blob/94a31e97b868ead86d27031280ead2f5c64fecbd/doc/abi.md#ci

It redirects you to https://github.com/radareorg/radare2/blob/master/doc/.github/workflows/build.yml#L522

Which is wrong since .github is not in the doc folder. Also pinning line 522 without a fixed commit gets outdated quite fast.

### DIFF
--- a/doc/abi.md
+++ b/doc/abi.md
@@ -47,4 +47,4 @@ What you **MUST** do in X.Y.9:
 
 All this is checked in the CI with the `abidiff` utility.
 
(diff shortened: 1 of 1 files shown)


trufae:
yeah guess it makes sense to pin that line, but as long as its documentation maybe better to just ship the snippet or move the logic into a seaprate file, so we dont need to reference a line

PR #22286, opened 2023-09-28
Reintroduce RBinPlugin->{init/fini} for plugins that need to initialize a third party global runtime ##bin

condret:

**Description**

**Copilot**

### DIFF
--- a/libr/bin/bin.c
+++ b/libr/bin/bin.c
@@ -416,15 +416,14 @@ R_IPI RBinXtrPlugin *r_bin_get_xtrplugin_by_name(RBin *bin, const char *name) {
 	return NULL;
 }
 
-static void r_bin_plugin_free(RBinPlugin *p) {
-	R_FREE (p);
-}
-
 R_API bool r_bin_plugin_add(RBin *bin, RBinPlugin *foo) {
 	RListIter *it;
 	RBinPlugin *plugin;
 
 	r_return_val_if_fail (bin && foo, false);
+	if (foo->init) {
+		foo->init (bin);
+	}
 
 	r_list_foreach (bin->plugins, it, plugin) {
 		if (!strcmp (plugin->meta.name, foo->meta.name)) {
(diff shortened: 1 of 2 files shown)


libr/include/r_bin.h
```diff
@@ -518,6 +518,8 @@ R_API void r_bin_trycatch_free(RBinTrycatch *tc);
 
 typedef struct r_bin_plugin_t {
 	RPluginMeta meta;
+	void (*init)();
+	void (*fini)();
```
trufae, line 522: Init and fini should take rbinfile as argument because they need a context and a place to save their global state

condret: for bf related things there is load and destroy. init and fini are needed for plugins that need a global runtime, like in actual global scope. if you init the mono runtime twice, it will crash. so this needs a to live outside rbinfile (in the global scope of the plugin), or you could not load multiple dotnet bins

trufae: Load and destroy sound like a very bad lifecycle names to me

trufae:
ping? can u fix the init/fini callbacks adding the context pointer?

libr/bin/bin.c
```diff
@@ -417,6 +417,9 @@ R_IPI RBinXtrPlugin *r_bin_get_xtrplugin_by_name(RBin *bin, const char *name) {
 }
 
 static void r_bin_plugin_free(RBinPlugin *p) {
+	if (p && p->fini) {
+		p->fini ();
```
trufae, line 421: Fini should take bin and binplugin too

trufae: Maybe the binplugin is implicit

trufae: problem is that RBinPluginFree is not taking RBin* so it doesnt have any context..

libr/include/r_bin.h
condret: didn't you came up with these?

libr/bin/bin.c
condret: uhm, the information that fini should free/finitialize is not actually stored in the plugin struct

PR #22324, opened 2023-10-14
Fix warning when rasm2 fails to load core plugins

dennis-doyensec:

**Description**

Before patch
```
$ rasm2 -a x86 'xor eax,eax'
ERROR: Library handler has failed for '~/.local/share/radare2/plugins/core_ghidra.dylib'
ERROR: Library handler has failed for '/usr/local/lib/radare2/5.8.9/pickle_dec.dylib'
31c0
```

This is because rasm2 does not set up any callbacks for core plugins. So the `p->handler` for core plugins is `NULL` and `r_lib_run_handler` treats this as an error. 

A byproduct of this patch is `p->file` won't leak when the `lib->plugins` `RList` list is free'd. 

I think there are a couple other potential leaks that this patch does not address. Namely using `RLibPlugin->free` on `RLibPlugin->data` and  closing `RLibPlugin->dl_handler` with `r_lib_dl_close`. I want to double check the logic before adding these. A small leak is preferable to a double free.

For reference `RLibPlugin` is seen below.
From: `libr/include/r_lib.h:56 - 66`
```c
typedef struct r_lib_plugin_t {
	int type;
	char *file;
	void *data; /* user pointer */
	struct r_lib_handler_t *handler;
	void *dl_handler; // DL HANDLER
	void (*free)(void *data);
#if 0
	RPluginMeta meta;
#endif
} RLibPlugin;
```

### DIFF
--- a/libr/util/lib.c
+++ b/libr/util/lib.c
@@ -372,7 +372,7 @@ R_API int r_lib_open_ptr(RLib *lib, const char *file, void *handler, RLibStruct 
 
 	if (!p->handler) {
 		rlibplugin_free (p);
-		return NULL;
+		return -1;
 	}
 
 	int ret = r_lib_run_handler (lib, p, stru);
--- a/libr/util/lib.c
+++ b/libr/util/lib.c
@@ -18,6 +18,13 @@ static const char *__lib_types_get(int idx) {
 	return r_lib_types[idx];
 }
 
+static inline void rlibplugin_free(RLibPlugin *plug) {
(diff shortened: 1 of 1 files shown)


dennis-doyensec:
Got a double free from my code, I am going to take a better look and send a better pull.

trufae:
Its ok you can use github ci to verify your changes at any time. Good finding!

trufae:
Are other tools affected? Didnt tried

PR #22354, opened 2023-10-28
Split r_core_cmd_help_match and add r_core_cmd_help_contains

### LINKED ISSUE #21437: Split r_core_cmd_help_match
The current signature includes `bool exact`, but it's rarely false, so I want to make it more clear at a glance.

```
// Current r_core_cmd_help_match(..., ..., true);
R_API void r_core_cmd_help_match(const RCore *core, RCoreHelpMessage help, R_BORROW R_NONNULL char *cmd);

// Current r_core_cmd_help_match(..., ..., false);
R_API void r_core_cmd_help_contains(const RCore *core, RCoreHelpMessage help, R_BORROW R_NONNULL char *cmd);
```

Lazula:
Closes #21437. Reads much better now.

Used the following regex:

```
$ find libr -type f -exec sed -i {} -E \
  -e 's/r_core_cmd_help_match (\(.*), true/r_core_cmd_help_match \1/' \
  -e 's/r_core_cmd_help_match (\(.*), false/r_core_cmd_help_contains \1/' \
  -e 's/r_core_cmd_help_match_spec (\(.*), true/r_core_cmd_help_match_spec \1/' \
  -e 's/r_core_cmd_help_match_spec (\(.*), false/r_core_cmd_help_contains_spec \1/' \;
```

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -420,12 +420,20 @@ R_API void r_core_cmd_help(const RCore *core, RCoreHelpMessage help) {
 	r_cons_cmd_help (help, core->print->flags & R_PRINT_FLAGS_COLOR);
 }
 
-R_API void r_core_cmd_help_match(const RCore *core, RCoreHelpMessage help, R_BORROW R_NONNULL char *cmd, bool exact) {
-	r_cons_cmd_help_match (help, core->print->flags & R_PRINT_FLAGS_COLOR, cmd, 0, exact);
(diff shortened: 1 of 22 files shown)


trufae:
Thank you! I was thinking that maybe we can drop the _cmd_ part from the function name making it shorter. What do you think? Is this reading well too?

* RCoreHelp.match() // without the cmd_ part

Lazula:
Right now it's using the same pattern as r_core_cmd_help, so these should be changed if that one is.

PR #22361, opened 2023-11-02
Update install.sh

apkunpacker:
Added termux support in install.sh


sys/install.sh
```diff
@@ -4,6 +4,12 @@
 cd "$(dirname $0)"/..
 pwd
 
+if [ "${SHELL}" = "/data/data/com.termux/files/usr/bin/bash" ]; then
+    echo "Termux environment detected. Installing necessary packages"  
+    pkg update -y && pkg install git build-essential binutils pkg-config -y
+    ${PWD}/sys/termux.sh
```
trufae, line 10:
```suggestion
    ${PWD}/sys/termux.sh
    exit $?
```

PR #22362, opened 2023-11-04
Fix entry point determination and info for HUNK file parser ##bin

michaelortmann:

**Description**

**Test1**
```
radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -NN -Qc 'pd 3' radare2-testbins/hunk/mc0201
            ;-- entry0:
            0x00000024      33fc01a000..   move.w 0x1a0, 0xdff096.l
            0x0000002c      43f90000004c   lea.l 0x4c.l, a1
            0x00000032      23c900dff080   move.l a1, 0xdff080.l
```

**Test2**
```
radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -NN -Qc 'pd 3' radare2-testbins/hunk/grey-final 
            ;-- entry0:
            0x00000020      4df900dff000   lea.l 0xdff000.l, a6
            0x00000026      3e3c7fff       move.w 0x7fff, d7
            0x0000002a      3d470096       move.w d7, 0x96(a6)
```

### DIFF
--- a/libr/bin/p/bin_hunk.c
+++ b/libr/bin/p/bin_hunk.c
@@ -82,9 +82,6 @@ static RList* entries(RBinFile *bf) {
 		}
 	}
 	R_LOG_ERROR ("Cannot determine entrypoint, cannot find HUNK_CODE");
-	ptr->paddr = 0;
-	ptr->vaddr = 0;
-	r_list_append (ret, ptr);
 	return ret;
 }
 
--- a/libr/bin/p/bin_hunk.c
+++ b/libr/bin/p/bin_hunk.c
@@ -6,15 +6,17 @@
 #if 0
 https://es.wikipedia.org/wiki/Amiga_Hunk
 http://amiga-dev.wikidot.com/file-format:hunk
+https://retro-commodore.eu/files/downloads/amigamanuals-xiik.net/eBooks/AmigaDOS%20Technical%20Reference%20Manual%20-%20eBook-ENG.pdf
 #endif
 
-#define HUNK_MAGIC "\x00\x00\x03\xf3"
+#define HUNK_HEADER "\x00\x00\x03\xf3"
+#define HUNK_CODE "\x00\x00\x03\xe9"
 
 static bool check(RBinFile *bf, RBuffer *b) {
 	if (r_buf_size (b) > 4) {
 		ut8 buf[4];
 		r_buf_read_at (b, 0, buf, sizeof (buf));
-		return (!memcmp (buf, HUNK_MAGIC, sizeof (buf)));
+		return (!memcmp (buf, HUNK_HEADER, sizeof (buf)));
 	}
 	return false;
 }
@@ -29,38 +31,13 @@ static RBinInfo *info(RBinFile *bf) {
 		return NULL;
 	}
 	ret->file = strdup (bf->file);
-	ret->type = strdup ("ROM");
+	ret->type = strdup ("Hunk (Executable file)");
 	ret->machine = strdup ("Amiga");
-	ret->os = strdup ("Workbench");
+	ret->os = strdup ("AmigaOS");
 	ret->arch = strdup ("m68k");
-	ret->cpu = strdup ("68040");
-	ret->bits = 8;
+	ret->bits = 32;
 	ret->has_va = 1;
-	return ret;
-}
-
-#if 0
-static void addsym(RList *ret, const char *name, ut64 addr, ut32 size) {
-	RBinSymbol *ptr = R_NEW0 (RBinSymbol);
-	if (!ptr) {
-		return;
-	}
-	ptr->name = strdup (r_str_get (name));
-	ptr->paddr = ptr->vaddr = addr;
-	ptr->size = size;
-	ptr->ordinal = 0;
-	r_list_append (ret, ptr);
-}
-#endif
-
-static RList* symbols(RBinFile *bf) {
-	RList *ret = NULL;
-	if (!(ret = r_list_newf (free))) {
-		return NULL;
-	}
-#if 0
-	addsym (ret, "NMI_VECTOR_START_ADDRESS", NMI_VECTOR_START_ADDRESS,2);
-#endif
+	ret->big_endian = true;
 	return ret;
 }
 
@@ -73,7 +50,7 @@ static RList* sections(RBinFile *bf) {
 	if (!(ptr = R_NEW0 (RBinSection))) {
 		return ret;
 	}
-	ptr->name = strdup ("hunk");
+	ptr->name = strdup ("HUNK_HEADER");
 	ptr->paddr = 0;
 	ptr->size = r_buf_size (bf->buf);
 	ptr->vaddr = 0;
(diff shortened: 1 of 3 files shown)


libr/bin/p/bin_hunk.c
```diff
@@ -84,17 +61,31 @@ static RList* sections(RBinFile *bf) {
...
+	ut8 b[4];
 	if (!(ret = r_list_new ())) {
 		return NULL;
 	}
 	if (!(ptr = R_NEW0 (RBinAddr))) {
 		return ret;
 	}
-	ptr->paddr = 0x24;
-	ptr->vaddr = 0x24;
+	buf = bf->buf;
```
trufae, line 76: Define and assign in the same line

libr/bin/p/bin_hunk.c
```diff
@@ -84,17 +61,31 @@ static RList* sections(RBinFile *bf) {
...
+	for (addr = 0x18; addr <= (r_buf_size (buf) - 4); addr += 0x4) {
```
trufae, line 77: Define a const variable holding this bufsize-4 before entering the loop, so thats not called. On every interation

trufae: Also no need for hex 0x4. Just use 4. Its more readable and consistent with the rest of the code

libr/bin/p/bin_hunk.c
```diff
@@ -84,17 +61,31 @@ static RList* sections(RBinFile *bf) {
...
+		r_buf_read_at (buf, addr, b, sizeof (b));
+		if (!memcmp (b, HUNK_CODE, sizeof (b))) {
```
trufae, line 79: Uhm thats pretty slow. What about read once a 64kb block (orbrestrict to a sane limit and use r_mem_mem and check if the result is aligned to 4? Or just do the read once avoiding the interation on large files and keep the memcmp loop

libr/bin/p/bin_hunk.c
```diff
@@ -109,7 +100,6 @@ RBinPlugin r_bin_plugin_hunk = {
 	.check = &check,
 	.entries = &entries,
 	.sections = sections,
-	.symbols = &symbols,
```
trufae, line 112: Uh?

trufae (changes requested):
Please add a test with snother executable that the entrypoint is not working with master code

libr/bin/p/bin_hunk.c
michaelortmann: i could also "fix" this line
https://github.com/radareorg/radare2/blob/master/libr/bin/p/bin_pyc.c#L29
while at it, where i originally copied that hex style from

libr/bin/p/bin_hunk.c
michaelortmann: the hunk parser is still very naive, so for this kind of parser 1k read will be enough. im planning proper hunk parsing for iS (section info) anyway, but i wanna keep this PR here small and fix more urgent issue here first.

libr/bin/p/bin_hunk.c
michaelortmann: the bin hunk plugin doesnt handle any symbol stuff at all yet. i dont see reason to keep this "unused" line.

libr/bin/p/bin_hunk.c
michaelortmann: i did commit code that precalced the `bufsize - 4`,

but in current code, where we have a bigger buffer now, that we read before the loop,
we need to keep track of position for `memcmp()` and `addr` and so we end up having a pointer addition in that very for loop anyway. only a slightly different one. now its a `b + 4` for `memcmp()`. it could be coded in different ways, but one addition / offset remains.

michaelortmann:
i changed `ret->bits` from 8 to 32.
for the mc680xx cpu is a kinda 32 bit cpu,
but im still unsure what  `ret->bits` should be.
smd (sega master drive) bin loader uses `bits = 16` and this macine has mc680xx also
`ret->bits` does have an impact of analyzer-disassembler-output, like in testcase mc0201
`lea.l 0x4c.l, a1`
vs
`lea.l 0x4c, a1`
no idea if `bits = 8`, `16` or `32` is correct here, but it should be same for amiga and sega master drive and other mc680xx based formats/machines.

trufae:
So iirc the m68k have 32bit sized general purpose registers but the data bus is tied to 16 or 24bit. So in this case the value should be 32. Because arch.bits is tied to the the general purpose register size (except for arm/thumb).

trufae:
About the circular address space caused by the 16/24 data bus. This can be done through the io.mask eval variable. Which is something that should be added as bininfo.. or archinfo. As long as it depends on the arch.cpu selected. What do you think? I can implement this once implemented and it will be useful for all non64bit archs

libr/bin/p/bin_hunk.c
```diff
@@ -84,17 +61,29 @@ static RList* sections(RBinFile *bf) {
...
+	for (addr = 0x18; addr <= last; addr += 4) {
+		if (!memcmp (b + addr, HUNK_CODE, 4)) {
+			ptr->paddr = addr + 8;
+			ptr->vaddr = addr + 8;
+			r_list_append (ret, ptr);
+			return ret;
+		}
+	}
+	R_LOG_ERROR ("Cannot determine entrypoint, cannot find HUNK_CODE");
+	ptr->paddr = 0;
```
trufae, line 85: Free ptr and return null. Im fine with the error message but if theres no entry. Handling nullable return value here should be better than asumimg address zero

trufae (changes requested):
After this change we should be good to go and merge. Thanks for your contribution!

about the differences in the disassembly and this awkward 0x00.l syntax seems something coming from the gnu arch plugin that must be fixed in a separate pr. I can take a look at it

trufae:
Or just return an empty list

libr/bin/p/bin_hunk.c
michaelortmann: lets return empoty list, see comment below.

michaelortmann:
> Or just return an empty list

done.

michaelortmann:
> About the circular address space caused by the 16/24 data bus. This can be done through the io.mask eval variable. Which is something that should be added as bininfo.. or archinfo. As long as it depends on the arch.cpu selected. What do you think? I can implement this once implemented and it will be useful for all non64bit archs

i dont fully understand. i understand meaning of circular buffer. but 64 bit archs have something similar. i remember riscv64 does use the upper bits of 64 bit addresses as flags or something. which means, it could also be masked. i could be wrong, i only have something like this in memory from when i touched it once some while ago.

rvalles:
> About the circular address space caused by the 16/24 data bus. This can be done through the io.mask eval variable. Which is something that should be added as bininfo.. or archinfo. As long as it depends on the arch.cpu selected. What do you think? I can implement this once implemented and it will be useful for all non64bit archs

Note Amiga does use the full address space (32bit), and 68020+ CPUs (alongside a variant of 68010) do address more than 24 bit.

The first 68000-based Amiga did not, but AmigaOS Manual: Exec did warn developers never to use the upper 8 bits for other purposes. I do not believe any software made that mistake.

PR #22376, opened 2023-11-09
Fix: Popd throws an error incorrectly on final pop

bqv:

Popd throws an error incorrectly on final pop. An empty stack should not be considered an error, after emptying it.


trufae:
Good catch! thank you

PR #22396, opened 2023-11-15
Assure map boundary sanity on creation ##io

condret:

**Description**


trufae:
your fix seems more correct from the io pov, but probably not the same as what a runtime linker would do. but i understand the fix for a null byte below shuold be made from rbin instead in that case. wdyt? also , i think my patch shouldnt harm if applied on top of yours.

Screenshot 2023-11-16 at 10 23 01

condret:
I think your patch is incomplete: sure io should handle maps that exceed the underlying file on read, but it should also handle that on write. instead of fixing the concrete read/write operation, it should adjust the boundaries of the affected maps and maybe do a warning. usually this case should never happen for the average r2 user, but I also care about people who hack with io structures (me).

trufae:
Seems like a reasonable change, making that behaviour explicit in the bin plugins sounds more correct than delaying the responsability to IO

PR #22417, opened 2023-11-19
Add missing map covering all the addresspace for frida://

radare:

**Description**


trufae:
@condret i dont know at which point r2 didnt created a map like this for frida:// but it will be good to have a better fix antime in the future. without this patch it was needed to io.va=0

PR #22420, opened 2023-11-20
Improve map tying support ##io

condret:

**Description**


trufae:
guess lgtm, no complains

trufae:
Do we want to have double tie? I mean maps should resize if fd changes.. but what if the map is resized? May this always change the underlying fd?

trufae:
ok to merge? rebase to be green, the failing tests is unrelated

condret:
i added another command 'omtf' to tie maps forward to the desc. 'omt' and 'omtb' do the same, as they tie the map backwards

PR #22436, opened 2023-11-22
Add arch fslsp

eibach:
This adds support for disassembling Freescale service processor code, as used in QorIQ and Layerscape systems.

**Description**

### DIFF
--- a/libr/arch/p/fslsp/plugin.c
+++ b/libr/arch/p/fslsp/plugin.c
@@ -34,8 +34,7 @@
 #include <r_arch.h>
 
 static bool fslsp_ancmd(RArchSession *s, RAnalOp *op, RArchDecodeMask mask) {
-	RStrBuf buf_asm;
-	r_strbuf_init (&buf_asm);
+	RStrBuf *buf_asm = NULL;
 	const bool be = R_ARCH_CONFIG_IS_BIG_ENDIAN (s->config);
 	const bool disasm = mask & R_ARCH_OP_MASK_DISASM;
 	ut8 cmd = (r_read_ble32 (op->bytes, be) >> 16) & 0xff;
@@ -43,7 +42,7 @@ static bool fslsp_ancmd(RArchSession *s, RAnalOp *op, RArchDecodeMask mask) {
 	switch (cmd) {
 	case 0x00:
 		if (disasm)
-			r_strbuf_setf (&buf_asm, "Block Copy src 0x%x, from 0x%x, to 0x%x, size 0x%x",
+			buf_asm = r_strbuf_newf ("Block Copy src 0x%x, from 0x%x, to 0x%x, size 0x%x",
 				r_read_ble32 (op->bytes, be) & 0xffff,
 				r_read_ble32 (op->bytes + 4, be),
 				r_read_ble32 (op->bytes + 8, be),
@@ -54,7 +53,7 @@ static bool fslsp_ancmd(RArchSession *s, RAnalOp *op, RArchDecodeMask mask) {
 
 	case 0x02:
 		if (disasm)
-			r_strbuf_setf (&buf_asm, "CCSR Write from Address word type %d, from 0x%x, to 0x%x, size 0x%x",
+			buf_asm = r_strbuf_newf ("CCSR Write from Address word type %d, from 0x%x, to 0x%x, size 0x%x",
 				r_read_ble32 (op->bytes, be) & 3,
 				r_read_ble32 (op->bytes + 4, be),
 				r_read_ble32 (op->bytes + 8, be),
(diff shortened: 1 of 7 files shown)


libr/arch/p/fslsp/plugin.c
```diff
@@ -0,0 +1,102 @@
...
+			r_read_ble32 (op->bytes, R_ARCH_CONFIG_IS_BIG_ENDIAN (s->config)) & 0xffff,
+			r_read_ble32 (op->bytes + 4, R_ARCH_CONFIG_IS_BIG_ENDIAN (s->config)),
+			r_read_ble32 (op->bytes + 8, R_ARCH_CONFIG_IS_BIG_ENDIAN (s->config)),
+			r_read_ble32 (op->bytes + 12, R_ARCH_CONFIG_IS_BIG_ENDIAN (s->config)));
+		op->type = R_ANAL_OP_TYPE_MOV;
+		op->size = 16;
+		break;
+
+	case 0x10:
+		sprintf (buf_asm, "Load RCW with Checksum");
```
eibach, line 21: I really would like to dump the RCW block here. What would be the proper way to do so?

trufae:
Everything looks good to me :) you can move forward adding more stuff here!

eibach:
> Everything looks good to me :) you can move forward adding more stuff here!

Great, thanks. Will continue this way then.

eibach:
All commands implemented. Can be merged if ok from your side.

eibach:
@trufae anything missing from your side?

trufae:
Oops missed your message. Let me finish lunch and ill review/reply/merge

libr/arch/p/fslsp/plugin.c
```diff
@@ -0,0 +1,231 @@
...
+ * Jump								0x84	8			Unconditional jump forward in PBI command sequence
+ * Jump Conditional					0x85	12			Conditional jump forward in PBI boot sequence
+ * CRC and Stop						0x8F	8			Stop the PBI sequence and indicate the expected CRC value
+ * Stop								0xFF	8			Stop the PBI sequence
+ */
+
+#include <r_arch.h>
+
+static bool fslsp_ancmd(RArchSession *s, RAnalOp *op, RArchDecodeMask mask) {
+	char buf_asm[128];
```
trufae, line 37: Better use RStrBuf here, so you dont have to worry about buffer boundaries later

libr/arch/p/fslsp/plugin.c
```diff
@@ -0,0 +1,231 @@
...
+	ut8 cmd = (r_read_ble32 (op->bytes, R_ARCH_CONFIG_IS_BIG_ENDIAN (s->config)) >> 16) & 0xff;
+
+	switch (cmd) {
+	case 0x00:
+		sprintf (buf_asm, "Block Copy src 0x%x, from 0x%x, to 0x%x, size 0x%x",
```
trufae, line 42: use snprintf or just r_strbuf_ apis

libr/arch/p/fslsp/plugin.c
```diff
@@ -0,0 +1,231 @@
...
+			r_read_ble32 (op->bytes, R_ARCH_CONFIG_IS_BIG_ENDIAN (s->config)) & 0xffff,
```
trufae, line 43: keep a `const bool be = R_ARCH_CONFIG_IS_BIG_ENDIAN(s->config); `at the begining of the function instead of calling this so many times

libr/arch/p/fslsp/plugin.c
trufae: how big is this rcw block? imho you can do that using an anal hint later, or associating a type or a format string to the given address where the instruction is found.

libr/arch/p/fslsp/plugin.c
```diff
@@ -0,0 +1,231 @@
...
+		break;
+
+	case 0x10:
+		sprintf (buf_asm, "Load RCW with Checksum");
+		op->type = R_ANAL_OP_TYPE_MOV;
+		op->size = 136;
+		break;
+
+	case 0x11:
+		sprintf (buf_asm, "Load RCW w/o Checksum");
```
trufae, line 68: maybe use mnemonics instead of instruction descriptions :? that seems a bit too long an dprobably wont be colorized properly

libr/arch/p/fslsp/plugin.c
```diff
@@ -0,0 +1,231 @@
...
+		op->type = R_ANAL_OP_TYPE_MOV;
+		op->size = 136;
+		break;
+
+	case 0x12:
+		sprintf (buf_asm, "Load Alternate Config Window %d",
```
trufae, line 74: dont sprintf if the DISASM bit is not set in the mask. the purpose of this flag is to speedup analysis when disassembly is not required

trufae (changes requested):
review done! sorry for the delay

eibach:
> review done! sorry for the delay

Thanks, awesome.

libr/arch/p/fslsp/plugin.c
eibach: Unfortunately there are no mnemonics defined by Freescale, co maybe I have to come up with my own?
Can I put the fulltext somewhere else?

libr/arch/p/fslsp/plugin.c
eibach: Yeah, maybe chip8 was not the best blueprint after all :)

libr/arch/p/fslsp/plugin.c
eibach: It is 32x 32bit words. Is there an example how to do the anal hint?

trufae:
the linting blocks you to use format string function without format string

Screenshot 2023-11-27 at 18 48 50

eibach:
Ah, I was already wondering

pancake ***@***.***> schrieb am Mo., 27. Nov. 2023, 18:49:

libr/arch/p/fslsp/plugin.c
```diff
@@ -0,0 +1,254 @@
...
+			r_strbuf_set (&buf_asm, "Alternate Configuration Write");
+		op->type = R_ANAL_OP_TYPE_MOV;
+		op->size = 8;
+	} else {
+		return false;
+	}
+
+	if (disasm) {
+		op->mnemonic = r_strbuf_tostring (&buf_asm);
+	}
```
trufae, line 232: Strbuf_fini or it will leak. You can also use drain() so the returned pointer is passed to the one that calls

trufae (changes requested):

libr/arch/p/fslsp/plugin.c
trufae: U can also use strbuf in the heap with new/free actually that will be non null only when disasm is requested so you can use this to spot bad uses too and drain() will work well

Issue #21746, opened 2023-05-15
Update PE parser with support for this new symbolname encoding


trufae:
* https://sourceware.org/bugzilla/show_bug.cgi?id=30444
* https://git.savannah.nongnu.org/cgit/poke.git/commit/?id=22ae5dc18877c6c086b2f4dd472ec188a3a9f90f

PR #22439, opened 2023-11-23
New plugin to support bigobj

MewtR:

**Description**
In order to address #21746, r2 first had to be able to parse [coff bigobj binaries](https://learn.microsoft.com/en-us/cpp/build/reference/bigobj-increase-number-of-sections-in-dot-obj-file?view=msvc-170).
This is what support for bigobj could look like. While it isn't particularly complicated to support bigobj, I was trying to find a clean way to do it without having this much duplicated code. Unfortunately I couldn't, but I'm curious to hear your thoughts on how this could be done. This plugin is maybe 90% copied from the coff plugin.


trufae:
Maybe both file formats can be handled from the same plugin so you dont need to link twice the same underlying implementaiton?  dont know much of the differences, but if its like elf and elf objects, maybe makes sense to do it in the same way

MewtR:
Thanks for the hint, I ended up looking at how the elf64 plugin was written and tried to do something similar for bigobj.
I'm much more comfortable with this iteration as it doesn't have as much copy-pasted code, although I'm not sure if I made a mess with the defines (`R_BIN_COFF_BIGOBJ`) or not. Anyways let me know what you think.

trufae:
Can you check this pr @terorie ?

terorie:
Will check. We already stuffed XCOFF handling in the same plugin, so I would naively assume this is doable.

libr/bin/format/coff/coff_specs.h
```diff
@@ -359,6 +396,19 @@ struct xcoff64_scn_hdr {
 	char pad44[4];
 });
 
+#if R_BIN_COFF_BIGOBJ
```
terorie, line 399: I suggest calling the new data structures bigobj_symbol, etc. Having the same struct name have twice different meanings depending on flags is a bit hacky.

terorie:
@MewtR Would you be open to chatting about this change in the Radare Discord or somewhere else?

trufae:
Ready to merge then? :)

trufae (changes requested):

libr/bin/format/coff/coff.c
```diff
@@ -47,14 +50,54 @@ static bool r_coff_supported_arch_le(const ut8 *buf) {
...
+#endif
 
+#ifndef R_BIN_COFF_BIGOBJ
 R_IPI bool r_coff_supported_arch(const ut8 *buf) {
 	return r_coff_supported_arch_le (buf) || r_coff_supported_arch_be (buf);
 }
+#endif
+
+// copied from bfd
+static bool r_coff_decode_base64(const char *str, ut32 len, ut32 *res) {
```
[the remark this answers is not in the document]
MewtR: Some time ago I had tested with the r2 functions but couldn't get it to work. According to this [thread](https://sourceware.org/bugzilla/show_bug.cgi?id=30444), they omit the padding characters. But I'll give it another shot.

libr/bin/format/coff/coff.c
```diff
@@ -145,6 +212,42 @@ static RBinAddr *r_xcoff_get_entry(RBinCoffObj *obj) {
 	return addr;
 }
 
+static bool r_coff_get_entry_helper(RBinCoffObj *obj, RBinAddr *address) {
+
```
trufae, line 216:
```suggestion
```

libr/bin/format/coff/coff.c
```diff
@@ -145,6 +212,42 @@ static RBinAddr *r_xcoff_get_entry(RBinCoffObj *obj) {
...
+		symbols = (void *)obj->symbols;
+		symbol_size = sizeof (struct coff_symbol);
+		symbol_count = obj->hdr.f_nsyms;
+	}
+
+	char *name;
+	int i;
+
+	for (i = 0; i < symbol_count; i++) {
+		name = (char *)symbols + i * symbol_size;
```
trufae, line 234:
```suggestion
		const char *name = (char *)symbols + i * symbol_size;
```

libr/bin/format/coff/coff.c
```diff
@@ -145,6 +212,42 @@ static RBinAddr *r_xcoff_get_entry(RBinCoffObj *obj) {
...
+		symbols = (void *)obj->bigobj_symbols;
+		symbol_size = sizeof (struct coff_bigobj_symbol);
+		symbol_count = obj->bigobj_hdr.f_nsyms;
+	} else {
+		symbols = (void *)obj->symbols;
+		symbol_size = sizeof (struct coff_symbol);
+		symbol_count = obj->hdr.f_nsyms;
+	}
+
+	char *name;
```
trufae, line 230:
```suggestion
```

libr/include/r_bin.h
```diff
@@ -937,6 +937,7 @@ extern RBinPlugin r_bin_plugin_java;
 extern RBinPlugin r_bin_plugin_dex;
 extern RBinPlugin r_bin_plugin_dis;
 extern RBinPlugin r_bin_plugin_coff;
+extern RBinPlugin r_bin_plugin_coff_bigobj;
```
trufae, line 940:
```suggestion
```

libr/bin/format/coff/coff.c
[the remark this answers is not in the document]
trufae: its ok for now, but if there's a bug in the r2 base64 apis it will be better to fix it there instead of having another duplicated implementation. thanks for taking care of this. i think the pr can be merged now! good work!

libr/bin/format/coff/coff.c
```diff
@@ -145,6 +212,40 @@ static RBinAddr *r_xcoff_get_entry(RBinCoffObj *obj) {
...
+	for (i = 0; i < symbol_count; i++) {
+		const char *name = (const char *)symbols + (i * symbol_size);
+		if ((!strcmp (name, "_start") || !strcmp (name, "start")) &&
+			r_coff_rebase_sym (obj, address, i)) {
+			return true;
+		}
+	}
+
+	for (i = 0; i < symbol_count; i++) {
+		name = (char *)symbols + i * symbol_size;
```
trufae, line 240:
```suggestion
		const char *name = (const char *)symbols + (i * symbol_size);
```

PR #22446, opened 2023-11-26
Fix dc derivatives causing unintended seek towards PC ##shell

kyufie:

**Description**
Currently, whenever `dc` or its derivatives is invoked, radare2 will listen to `dbg.follow` and set the offset accordingly.
But this is not needed when, for instance, followed by `?`.
In that case, seeking would be undesirable as it also affects the `?*` command, throwing people's offset every time recursive help is requested.


trufae:
Looks good

Issue #21466, opened 2023-03-12
DWARF hardcodes ELF section names


terorie:
## Description

`libr/bin/dwarf.c` looks up DWARF sections by names that are ELF-specific.

https://github.com/radareorg/radare2/blob/86b5d8e4ab198619c7528c94bac977f84f722586/libr/bin/dwarf.c#L2508

This is a problem because:
1. Other binary formats (e.g. XCOFF) use different names (e.g. `.dwline` instead of `.debug_line`)
2. Other binary formats discover sections by numeric identifiers not strings (e.g. on XCOFF, where the string `.dwline` as section name is technically optional -- the flags bitfield decides the type of the section)

## Suggested Fix

There should be an abstraction for discovering DWARF sections to be implemented by the respective bin plugins. The `dwarf.c` code should retrieve this info via the abstraction and not hardcode any lookups. Possible ways to do this:

A -- Add callback function  `.dbginfo` 
B -- Add more metadata to RBinSection
C -- Change the affected bin plugins (e.g. XCOFF) to hijack the names to change `.dwline` => `.debug_line`, etc. Works without changes to core but hacky

trufae:
What about adding a field in rbinsection specifying the section contains debug information instead of adding a new callback? I think for now we can hardcode this in dwarf.c or at least make some assumptions from there because dwarf.c have access to rbin and knows the file format that is loaded

trufae:
Imho B sounds like a good solution but that will break abi. So i would go for modifying dwarf.c to use the right sectionname depending on plugin used. I dont think we wikl find much more different names to justify bigger breaking changes

trufae:
Can you do a PR with the quickfix (adding the coff dwarf names in dwarf.c depending on the plugin used?)

trufae:
ping @terorie

trufae:
Ping ...

terorie:
Super sorry @trufae Job and life things swamped me. I can't guarantee I can get to this soon... Feel free to take my PRs over if you guys are looking for work...

trufae:
Nobody is looking to contribute or so. Dont feel bad life events suck and there’s no pressure on anything. If anyone really cares about this issue it shouldnt take much time to get fixed, i was just trying to get an update to wrap up remaining tickets to cut a release and organize my time. Thanks for replying, hope things get better soon in your side <3

terorie:
@trufae A simple fix is available in #22449.

[...]

PR #22449, opened 2023-11-26
Support DWARF section names for XCOFF64 ##bin

terorie:

**Description**

Fixes https://github.com/radareorg/radare2/issues/21466

### DIFF
--- a/libr/bin/dwarf.c
+++ b/libr/bin/dwarf.c
@@ -326,15 +326,65 @@ static const char *dwarf_unit_types[] = {
 	[DW_UT_hi_user] = "DW_UT_hi_user",
 };
 
+enum {
+	DWARF_SN_ABBREV,
+	DWARF_SN_INFO,
+	DWARF_SN_FRAME,
+	DWARF_SN_LINE,
+	DWARF_SN_LOC,
+	DWARF_SN_STR,
+	DWARF_SN_LINE_STR,
+	DWARF_SN_RANGES,
+	DWARF_SN_ARANGES,
+	DWARF_SN_PUBNAMES,
+	DWARF_SN_PUBTYPES,
+
+	DWARF_SN_MAX
+};
+
+static const char *dwarf_sn_elf[DWARF_SN_MAX] = {
+	[DWARF_SN_ABBREV] = "debug_abbrev",
+	[DWARF_SN_INFO] = "debug_info",
+	[DWARF_SN_FRAME] = "debug_frame",
+	[DWARF_SN_LINE] = "debug_line",
+	[DWARF_SN_LOC] = "debug_loc",
+	[DWARF_SN_STR] = "debug_str",
+	[DWARF_SN_LINE_STR] = "debug_line_str",
+	[DWARF_SN_RANGES] = "debug_ranges",
+	[DWARF_SN_ARANGES] = "debug_aranges",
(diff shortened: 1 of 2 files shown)


libr/bin/dwarf.c
```diff
@@ -326,15 +326,65 @@ static const char *dwarf_unit_types[] = {
...
 	RListIter *iter;
 	RBinSection *section = NULL;
 	RBinFile *binfile = a ? a->cur: NULL;
 	RBinObject *o = binfile ? binfile->bo : NULL;
 	if (R_LIKELY (o && o->sections)) {
+		/* XXX: xcoff64 specific hack */
+		const char * const *name_tab = strcmp (o->info->rclass, "xcoff64")
+			? dwarf_sn_elf
+			: dwarf_sn_xcoff64;
+		assert (sn >= 0 && sn < DWARF_SN_MAX);
```
trufae, line 384: we dont do assert in r2land, you can use an r_return_if_fail here, so it's removed on release builds. but i think this is more like a precondition that shuold happen at the begining of the function. same check for a that is not nullable

libr/bin/dwarf.c
```diff
@@ -326,15 +326,65 @@ static const char *dwarf_unit_types[] = {
...
+	[DWARF_SN_PUBNAMES] = "dwpbnms",
+	[DWARF_SN_PUBTYPES] = "dwpbtyp"
+};
+
 // 1 of 20s spent in this non-mnemonized function
-static RBinSection *getsection(RBin *a, const char *sn) {
+static RBinSection *getsection(RBin *a, int sn) {
 	RListIter *iter;
 	RBinSection *section = NULL;
 	RBinFile *binfile = a ? a->cur: NULL;
```
trufae, line 377: i dont think a should be named a or be nullable. also we can pick the a->cur->binfile->bo via R_UNWRAP macros now which makes it cleaner to nullcheck sequences of pointers

libr/bin/dwarf.c
```diff
@@ -326,15 +326,65 @@ static const char *dwarf_unit_types[] = {
...
+		/* XXX: xcoff64 specific hack */
+		const char * const *name_tab = strcmp (o->info->rclass, "xcoff64")
+			? dwarf_sn_elf
+			: dwarf_sn_xcoff64;
+		assert (sn >= 0 && sn < DWARF_SN_MAX);
+		const char *name_str = name_tab[sn];
 		r_list_foreach (o->sections, iter, section) {
-			if (strstr (section->name, sn)) {
+			if (strstr (section->name, name_str)) {
 				if (strstr (section->name, "zdebug")) {
```
trufae, line 388: oh i think that should be a good thing to have at some point too. ive seen those compressed dwarf sections more freqntly these days, but not for this pr. anyway i dont see why your pr is breaking so much tests. can you rebase and see if the magic happens? because your changes look good to me

trufae (changes requested):
.

libr/bin/dwarf.c
terorie: Fixed

libr/bin/dwarf.c
terorie: > Anyway i dont see why your pr is breaking so much tests

Fixed 🙏🏻

PR #22457, opened 2023-11-28
Fix a few long-standing issues in dyldcache parsing

mrmacete:

This change gets rid of a couple of wrong assumptions:

- we assumed that entries in the images array were unique: wrong because multiple items with different names can refer to a single image in the dyld cache, for example when links are present when the cache is built (like `/usr/lib/libobjc.A.dylib` and `/usr/lib/libobjc.dylib` in iOS 9.x caches) and in that case we don't want to load and parse duplicated information
- we assumed that local symbol entries were 1:1 in terms of number and ordering with the images in the image list: wrong, symbol entries can come in any number - instead they reference the corresponding image by offset in the file, we just need to correct those offsets for some overhead we have when dealing with multi-file caches and we're now sure to not miss symbols or load more than the user is interested in

While at it, now also the old "alien dep" infinite loop is gone 😅

### DIFF
--- a/libr/bin/p/bin_dyldcache.c
+++ b/libr/bin/p/bin_dyldcache.c
@@ -19,12 +19,14 @@ typedef struct {
 typedef struct _r_dyldcache {
 	cache_hdr_t *hdr;
 	ut64 *hdr_offset;
+	ut64 *hdr_overhead;
 	ut32 *maps_index;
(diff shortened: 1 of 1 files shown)


trufae:
Woah that’s quite a huge patch and lots of good old bugs squashed <3

PR #22467, opened 2023-12-02
Fix security issues in coff plugin ##bin

terorie:

**Description**

Kudos to @trufae for reporting.

1. Fix out-of-bounds read in XCOFF ldsym handling (Affected versions: none, as this was a recent regression in https://github.com/radareorg/radare2/pull/22458)
2. Fix unbounded memory allocation in (X)COFF section handling 

The second bug has probably been there for a while. If `RBinSection::size` is not bounded, then the generic logic to find strings will calloc a large buffer. It is low severity, as the alloc failure will be gracefully handled without terminating the process. We should probably harden the RBinObject logic regardless to always truncate the section file ranges to the bounds of the file. It's better to never attempt to alloc that much memory in the first place.

### DIFF
--- a/libr/bin/format/coff/coff.c
+++ b/libr/bin/format/coff/coff.c
@@ -302,10 +302,11 @@ static bool r_bin_xcoff_init_ldhdr(RBinCoffObj *obj) {
 static bool r_bin_xcoff_init_ldsyms(RBinCoffObj *obj) {
 	int ret;
 	size_t size;
-	ut64 offset = obj->scn_hdrs[obj->x_opt_hdr.o_snloader-1].s_scnptr + sizeof (struct xcoff32_ldhdr);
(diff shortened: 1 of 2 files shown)


terorie:
Ping @radare

trufae:
I was busy sleeping at 5am. Sorry for the delay in merging it

terorie:
> I was busy sleeping at 5am. Sorry for the delay in merging it

😆 You are forgiven. Sorry, I didn't mean to make it seem time sensitive. Just CCed you since you asked for the change.

trufae:
I shouldnt sleep its a bad security practice i know. Maybe setting up some keyword sensitive in the mail filter to ring an alarm may help us get bugfixes merged earlier 😅 np! And thanks for the fix! That was fast!

PR #22478, opened 2023-12-06
Don't use /dev/stderr ##port

terorie:

**Description**

`/dev/stderr` is not supported on AIX.

### DIFF
--- a/configure
+++ b/configure
@@ -903,7 +903,7 @@ pcgen() {
 	PCNAME="${1}" ; shift
 	REQUIRES="$@"
 
-	echo "generating ${PCFILE}" > /dev/stderr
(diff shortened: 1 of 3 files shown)


configure
```diff
@@ -903,7 +903,7 @@ pcgen() {
 	PCNAME="${1}" ; shift
 	REQUIRES="$@"
 
-	echo "generating ${PCFILE}" > /dev/stderr
+	echo "generating ${PCFILE}" >&2
```
terorie, line 906: Just noticed that this was generated by acr

terorie: https://github.com/radareorg/acr/pull/32

PR #22502, opened 2023-12-12
Fix typo in dbg.px0 test

koyaan:

Fix a one byte typo in test for `dbg.px0`


trufae:
I think those tests are not running in the ci. Can you try enabling them when they all run in local? (Only for linux for now) thanks for taking care!

koyaan:
Maybe this is related to r2r thinking its on 32bit linux, or how/where would i go about enabling them?

trufae:
I think arch tests arent run unless  implicitly specified. I check the code and i found no traces of being disabled

PR #22509, opened 2023-12-16
Improve sigdb file format detection ##anal

micronn:

**Description**

Detection code in signdb_type doesn't seem to work when the correct extension is not provided. This also happens when trying to load from a gzipped sdb file (because temporary file does not have sdb extension).

I did't know that signature files created by zos should have .sdb extension and I did zos test.sig, but didn't load when I tried zo test.sig. Now it loads fine.

I don't know if the sdb format has changed over time, there are parts of the current detection code that seem odd to me. I think this commit should make detection a bit better than before.

### DIFF
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -3010,7 +3010,7 @@ static int signdb_type(const char *file) {
 				continue;
 			}
 		}
-		if (is_sdb == 0) {
+		if (is_sdb < 0) {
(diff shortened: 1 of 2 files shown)


trufae:
Can you add a test for this?

micronn:
Done basic tests. Sorry about the commit mess. I assume a squash merge could be done.

trufae:
Yeah squash is the way to go. Thank you! Looks good

PR #22510, opened 2023-12-20
document

gogo2464:
More doc on dcs. Plan to make more syscall on windows.

**Description**

not ready to merge.

### DIFF
--- a/libr/debug/debug.c
+++ b/libr/debug/debug.c
@@ -1561,7 +1561,17 @@ static int show_syscall(RDebug *dbg, const char *sysreg) {
 	return reg;
 }
 
-// continue execution until a syscall is found, then return its syscall number or -1 on error
+/* Continue execution until a syscall is found.
+ *
+ * If no plugin is selected, read the x86 sn register in order to determine the syscall. See https://www2.lauterbach.com/pdf/debugger_x86.pdf.
+ * In case of bug, it is recommended to use a debugger plugin such as -d "windbg://file.exe" on Windows.
+ *
+ * first argument, RDebug: debugger configuration structure.
+ * second argument, int sc: register value.
+ * third argument, int n_sc: syscall number (eg: sys_read = 0 in x64).
+ *
+ * return the syscall number or -1 on error.
+*/
 R_API int r_debug_continue_syscalls(RDebug *dbg, int *sc, int n_sc) {
 	r_return_val_if_fail (dbg, false);
 	int i, err, reg;


gogo2464:
why not:

```
R_API int r_debug_continue_syscalls(RDebug *dbg, int *sc, int n_sc) {
	r_return_val_if_fail (dbg, false);
	int i, err, reg;
	/* user-level syscall tracing */
	r_debug_continue_until_optype (dbg, R_ANAL_OP_TYPE_SWI, 0);
	return show_syscall (dbg, "A0");
```

instead of

```R_API int r_debug_continue_syscalls(RDebug *dbg, int *sc, int n_sc) {
	r_return_val_if_fail (dbg, false);
	int i, err, reg;
	if (!dbg->current || r_debug_is_dead (dbg)) {
		return -1;
	}
	RDebugPlugin *plugin = R_UNWRAP3 (dbg, current, plugin);
	if (plugin && !plugin->contsc) {
		/* user-level syscall tracing */
		r_debug_continue_until_optype (dbg, R_ANAL_OP_TYPE_SWI, 0);
		return show_syscall (dbg, "A0");
	}
	
...```.

**Comment by trufae** [comment]
these two codes do different things, i dont see the point in the change and its unrelated to the code documentation, which looks unnecessary and wrong

File: libr/debug/debug.c [hunk 1]
```diff
@@ -1561,7 +1561,17 @@ static int show_syscall(RDebug *dbg, const char *sysreg) {
 	return reg;
 }
 
-// continue execution until a syscall is found, then return its syscall number or -1 on error
+/* Continue execution until a syscall is found.
+ *
+ * If no plugin is selected, read the x86 sn register in order to determine the syscall. See https://www2.lauterbach.com/pdf/debugger_x86.pdf.
+ * In case of bug, it is recommended to use a debugger plugin such as -d "windbg://file.exe" on Windows.
+ *
+ * first argument, RDebug: debugger configuration structure.
```
trufae, line 1569: rdebug is the instance that holds the state of rdebug, it have nothing to do with the debugger configuration

libr/debug/debug.c
```diff
@@ -1561,7 +1561,17 @@ static int show_syscall(RDebug *dbg, const char *sysreg) {
...
+ * second argument, int sc: register value.
```
trufae, line 1570: first, second, third, those are not correct ways to refere to the arguments, varnames have names, and those should be clear anough to not need to be documented for example. its part of the developer guide

libr/debug/debug.c
```diff
@@ -1561,7 +1561,17 @@ static int show_syscall(RDebug *dbg, const char *sysreg) {
 	return reg;
 }
 
-// continue execution until a syscall is found, then return its syscall number or -1 on error
+/* Continue execution until a syscall is found.
+ *
+ * If no plugin is selected, read the x86 sn register in order to determine the syscall. See https://www2.lauterbach.com/pdf/debugger_x86.pdf.
```
trufae, line 1566: i dont see whats the relationship between this pdf of > 100 pages and this sentence.

libr/debug/debug.c
gogo2464: good point. does r2 still uses doxygen?

libr/debug/debug.c
gogo2464: the pdf document how works the sn register in Linux.

the debugger method is very obscure!!!

libr/debug/debug.c
gogo2464: I need more documentation experience.

trufae: R2 never used doxygen and never enforced documentation. See DEVELOPERS.md 

so no need to document anything. Is something is not obvious change function names and variables but documentation shouldnt be added

gogo2464: ah my bad. I suggest to add reference doc... maybe for a full PR. What do you think of this idea?

trufae:
Closing the PR, open it when you have some fixes to share here

PR #22515, opened 2023-12-27
Update docker readme

prodrigestivill:
- Disable GIT docker publish via Makefile (since this is done via github actions on [radare2-snap](https://github.com/radareorg/radare2-snap)).
- Update docker README with all the details of the 2 versions available (also fix the build commands).
- Update snap doc to include all new architectures and commands.


trufae:
Thanks!

PR #22516, opened 2023-12-28
Better way to determine the glibc version

MewtR:

**Description**
So I wrote a new function that determines the glibc version based on the strings inside the .so.
This is similar to what pwndbg/GEF do. Previously, the version was parsed out of the filename, but this is not
relevant for newer versions of libc as they will all be named `libc.so.6`

### DIFF
--- a/libr/core/cmd_debug.inc.c
+++ b/libr/core/cmd_debug.inc.c
@@ -1393,6 +1393,48 @@ beach:
 	r_list_free (list);
 }
 
+R_API double get_glibc_version(RCore *core, const char *libc_path) {
+	double version = 0.0f;
+	RBin *bin = r_bin_new ();
+	RIO *io = r_io_new ();
+	r_io_bind (io, &bin->iob);
+	RBinFileOptions opt = { 0 };
+
+	if (r_bin_open (bin, libc_path, &opt)) {
+		RBinString *string;
+		RList *strings = r_bin_get_strings (bin);
+		RListIter *iterator;
+
+		// I feel like this regex might miss some cases
+		// but it seems to be the same one used by GEF/pwndbg
+		RRegex *glibc_version_regex = r_regex_new ("[0-9]+[0-9]*\\.[0-9]+[0-9]*", "e");
+		r_list_foreach (strings, iterator, string) {
+
+			/*
+			  This logic is to search for the following two strings in glibc
+			  and parse out the version:
+			  1) "glibc <version>"
+			  2) "GNU C Library (GNU libc) stable release version <version>."
+			*/
+			if ((strstr (string->string, "glibc") || strstr (string->string, "GNU C Library"))) {
+				RList *matches = r_regex_match_list (glibc_version_regex, string->string);
+
+				// We only care about the first match
+				const char *first_match = r_list_first (matches);
+				version = r_num_get_double (core->num, first_match);
+				r_list_free (matches);
+				if (version) {
+					break;
+				}
+			}
+		}
+		r_regex_free (glibc_version_regex);
+	}
+	r_bin_free (bin);
+	r_io_free (io);
+	return version;
+}
+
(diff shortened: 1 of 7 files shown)


koyaan:
Funny that after years of stale code two people are working on the same thing, I also did some new libc version detection in https://github.com/koyaan/radare2/commit/ffcb7c2fc9e6349eff50ce389b2b447314e27e35 
1) looking for a `__libc_version` version symbol and 2) doing string searching in `.rodata` section similiar to your solution

MewtR:
Yeah... you're version seems solid. I also made a change to `regcomp.c` which could potentially be dangerous. Are you going to have a PR up soon?

libr/core/cmd_debug.inc.c
```diff
@@ -1393,6 +1393,48 @@ static void cmd_debug_modules(RCore *core, int mode) { // "dmm"
...
+			  and parse out the version:
+			  1) "glibc <version>"
+			  2) "GNU C Library (GNU libc) stable release version <version>."
+			*/
+			if ((strstr (string->string, "glibc") || strstr (string->string, "GNU C Library"))) {
+				RList *matches = r_regex_match_list (glibc_version_regex, string->string);
+
+				// We only care about the first match
+				char *first_match = r_list_first (matches);
+				version = r_num_get_float (core->num, first_match);
```
trufae, line 1424: i think this api should be renamed to get_double instead 🤔 

```unum.c:R_API double r_num_get_float(RNum *num, const char *str) {```

File: libr/core/cmd_debug.inc.c [hunk 2]
```diff
@@ -1393,6 +1393,48 @@ static void cmd_debug_modules(RCore *core, int mode) { // "dmm"
...
+			  This logic is to search for the following two strings in glibc
+			  and parse out the version:
+			  1) "glibc <version>"
+			  2) "GNU C Library (GNU libc) stable release version <version>."
+			*/
+			if ((strstr (string->string, "glibc") || strstr (string->string, "GNU C Library"))) {
+				RList *matches = r_regex_match_list (glibc_version_regex, string->string);
+
+				// We only care about the first match
+				char *first_match = r_list_first (matches);
```
trufae, line 1423:
```suggestion
				const char *first_match = r_list_first (matches);
```

libr/core/cmd_debug.inc.c
```diff
@@ -1393,6 +1393,48 @@ static void cmd_debug_modules(RCore *core, int mode) { // "dmm"
...
 }
 
+R_API double get_glibc_version(RCore *core, const char *libc_path) {
+	double version = 0.0f;
+	RBin *bin = r_bin_new ();
+	RIO *io = r_io_new ();
+	r_io_bind (io, &bin->iob);
+	RBinFileOptions opt = { 0 };
+
+	if (r_bin_open (bin, libc_path, &opt)) {
```
trufae, line 1403: maybe worth adding some extra checks about the filetype. but guess its fine for an intiaial approach.

trufae (approved):
lgtm

libr/core/cmd_debug.inc.c
MewtR: I can make this change, but I'd rather do it in a separate PR

libr/core/cmd_debug.inc.c
MewtR: Checks such as what? Like making sure it's an .so?

libr/core/cmd_debug.inc.c
trufae: Yes please

libr/core/cmd_debug.inc.c
trufae: Not sure if its worth ensuring its an elf too. But i doubt this function will take a wrong path as arg

trufae:
Should i close this pr?

koyaan:
Please leave it open till mine is resolved i will take some code from this

trufae:
Closing this pr in favour of the other one

PR #22518, opened 2023-12-29
WIP: Update pyc magic detection ##bin

dennis-doyensec:

**Description**

Does not work yet, but wanted to make sure this is acceptable before going to far. Feel free to make suggestions.

Pyc version magic are out of date and use a linear search. The linear search is done multiple times. This commit adds new versions, puts them into a switch table. Once completed, the magic compare should only be  done once.

New versions were taken from [here](https://github.com/python/cpython/blob/main/Lib/importlib/_bootstrap_external.py#L220). This seems to be the most comprehensive list, but I don't think it is complete. I did a diff with what r2 has and used r2's versions when available. I also deleted many duplicate magic signatures.


libr/bin/format/pyc/pyc_magic.c
```diff
@@ -3,226 +3,502 @@
...
+	case 0x0a0d0dc7: // 3527
+		return &(PycVer){ 3, 12, "b1", "(Add LOAD_SUPER_ATTR)" };
+	case 0x0a0d0dc8: // 3528
+		return &(PycVer){ 3, 12, "b1", "(Add LOAD_SUPER_ATTR_METHOD specialization)" };
+	case 0x0a0d0dc9: // 3529
+		return &(PycVer){ 3, 12, "b1", "(Inline list/dict/set comprehensions)" };
+	case 0x0a0d0dca: // 3530
+		return &(PycVer){ 3, 12, "b1", "(Shrink the LOAD_SUPER_ATTR caches)" };
+	case 0x0a0d0dcb: // 3531
+		return &(PycVer){ 3, 12, "b1", "(Add PEP 695 changes)" };
```
trufae, line 499: Uhm this is returning a pointer in the stack. Thats not correct. What about returning the struct and use -1 in the first field instead of null on error?

dennis-doyensec: yeah... I was thinking backwards. I will update it. Thanks.

trufae:
just update this test

Screenshot 2024-01-08 at 10 39 21

trufae:
(and will be good to merge)

dennis-doyensec:
This code ends up being subtly wrong. Python versions are hard, major and minor are not enough...

I can fix it with a python version comparison function. This will also speed up the disassembler and let the disassembler use the nearest match. If the disassembler and magic list get out of line, the disassembler can then warn but still try it's best.

dennis-doyensec:
So I added a basic python version comparison alg. It's not perfect, it does not handle `dev` and `+` versions but we shouldn't run into those, since we only care about versions where the pyc magic changed.

The switch is now sorted by python's own `from packaging import version`. I validated that the new version compare function agreed with what python said for all possible pairs of versions in the switch.

I will debug/fix tests tomorrow.

trufae:
Good work! I also think and agree that we dont have to go into git versions here

dennis-doyensec:
Yeah, I am just not sure what do about it. I don't want to do all the legwork to get the commits but I like that they exist.

If you are stuck reversing magic `0x0a0dc61b` then unique opcodes may not be well supported. Those opcodes are undocumented. I can easily find 2.0py source but 2.0b1?. Not just that, apparently opcodes (or pyc magic) changed may times over the life of 2.0b1???

I just don't want to do the legwork. It's amazing someone before me did. I was tempted to put in:
`commit: Don't know (if you do now, contribute it to __FILE__:__LINE__)`

libr/bin/format/pyc/pyc_magic.c
```diff
@@ -1,282 +1,658 @@
...
+	return ma - mb;
+}
+
+/*
+ * Bassed on PEP 440 (https://peps.python.org/pep-0440/)
+ * This algorythm is good enough for the versions we have. We shouldn't need to
+ * work with `dev` or `+` versions since we only care about versions with
+ * unique pyc magic.
+ */
+int py_version_cmp(char *va, char *vb, bool *err) {
```
trufae, line 611: arent va/vb const ?

libr/bin/format/pyc/pyc_magic.c
```diff
@@ -1,282 +1,658 @@
...
-	{ 0x0a0d0d55, "v3.8.1", "d2529ef779ce819a5ea833b264e47440efcbac29" },
-	{ 0x0a0d0d55, "v3.8.1rc1", "827f6399a61be9d14f8ccfa5be73a6030ec45f1d" },
-	{ 0x0a0d0d5c, "v3.9.0a1", "fd757083df79c21eee862e8d89aeefefe45f64a0" },
-	{ 0x0a0d0d5e, "v3.9.0a2", "bf0a31c8fb782e03e9530c2488ab2d0e29fc0495" },
-	{ 0x0a0d0d60, "v3.9.0a3", "a36ea266c6470f6c65416f24de4497637e59af23" },
-	{ 0x0a0d0d61, "v3.9.0", "123" },
+struct pyc_version get_pyc_version(ut32 magic) {
+	switch (magic) {
+	case 0x00949494:
+		return (struct pyc_version){ 0x00949494, "0.9.4b0", "77b80a91d357c1d95d8e7cd4cbbe799e5deb777e" };
```
trufae, line 8: you can also have a macro that takes 3 parameters and constructs this cast+struct . tomake it less verbose

libr/bin/format/pyc/pyc_magic.c
trufae: dunno why th elinter complains about this line :?

trufae (changes requested):

libr/bin/format/pyc/pyc_magic.c
trufae: git grep -n 'cmp(' libr | grep -v R_API | grep -v static | grep c:) && exit 1 I think thats the issue from lint.sh

dennis-doyensec:
That lint rule seems to prohibit function names that end in `cmp(` unless they are `R_API` or `static`. I don't really understand why. But the function it's angry at should probably be `R_IPI`. So I can update that and update the lint.sh to accept `R_IIPI`. I think that is the way to fix it?

If not I can try something else.

trufae:
Lgtm. Ready to merge for you?

PR #22522, opened 2024-01-02
minor fix

condret:

**Description**

### DIFF
--- a/libr/util/thread.c
+++ b/libr/util/thread.c
@@ -39,14 +39,20 @@ static void *_r_th_launcher(void *_th) {
 	}
 	r_th_lock_enter (th->lock);
 	do {
-		r_th_set_running (th, true);
+//		r_th_set_running (th, true);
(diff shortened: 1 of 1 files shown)


libr/util/thread.c
```diff
@@ -39,14 +39,20 @@ static void *_r_th_launcher(void *_th) {
...
 		if (ret < 0) {
 			th->ready = false;
+//			r_th_set_running (th, false);
+// don't call r_th_set_running, as it would unlock th->lock
+			th->running = false;
 			r_th_lock_leave (th->lock);
 			return 0;
 		}
-		r_th_set_running (th, false);
+//		r_th_set_running (th, false);
```
trufae, line 54: uhm whats the helper doing that we dont like?

condret: locking and unlocking th->lock, but it has already been locked previously. this also renders the call to r_th_lock_leave below pointless

trufae: So merge

PR #22526, opened 2024-01-06
Add initial support for vsf dialect produced by C64 Debugger app ##bin

mrmacete:

**Description**

This change avoids misinterpreting vsf files currently produced by the C64 Debugger app (https://sourceforge.net/projects/c64-debugger/) in which the header has one more field (the Vice version), and can also produce a module named `"C64MEMHACKS"` which clashes with the vital `"C64MEM"` since only the prefix is compared.

Turns out also Vice for macOS is emitting this format now.

### DIFF
--- a/libr/bin/p/bin_vsf.c
+++ b/libr/bin/p/bin_vsf.c
@@ -93,7 +93,7 @@ static bool load(RBinFile *bf, RBuffer *b, ut64 loadaddr) {
 				return false;
 			}
 #define CMP_MODULE(x) memcmp (module.module_name, x, sizeof (x) - 1)
-			if (!CMP_MODULE (VICE_C64MEM) && CMP_MODULE (VICE_C64MEMHACKS) &&!module.major) {
+			if (!CMP_MODULE (VICE_C64MEM) && CMP_MODULE (VICE_C64MEMHACKS) && !module.major) {
 				res->mem = offset + read;
 			} else if (!CMP_MODULE (VICE_C64ROM) && !module.major) {
 				res->rom = offset + read;
@@ -106,8 +106,7 @@ static bool load(RBinFile *bf, RBuffer *b, ut64 loadaddr) {
 				r_buf_read_at (bf->buf, offset + read, (ut8 *)res->maincpu, sizeof (*res->maincpu));
 			} else {
 				char safe_name[sizeof (module.module_name) + 1];
-				memcpy (safe_name, module.module_name, sizeof (module.module_name));
-				safe_name[sizeof (module.module_name)] = 0;
+				r_str_ncpy (safe_name, module.module_name, sizeof (safe_name));
 				R_LOG_TODO ("Ignoring unsupported module: %s", safe_name);
 			}
 #undef CMP_MODULE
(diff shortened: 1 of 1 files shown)


libr/bin/p/bin_vsf.c
```diff
@@ -85,7 +93,7 @@ static bool load(RBinFile *bf, RBuffer *b, ut64 loadaddr) {
 				return false;
 			}
 #define CMP_MODULE(x) memcmp (module.module_name, x, sizeof (x) - 1)
-			if (!CMP_MODULE (VICE_C64MEM) && !module.major) {
+			if (!CMP_MODULE (VICE_C64MEM) && CMP_MODULE (VICE_C64MEMHACKS) &&!module.major) {
```
trufae, line 96:
```suggestion
			if (!CMP_MODULE (VICE_C64MEM) && CMP_MODULE (VICE_C64MEMHACKS) && !module.major) {
```

libr/bin/p/bin_vsf.c
```diff
@@ -96,6 +104,11 @@ static bool load(RBinFile *bf, RBuffer *b, ut64 loadaddr) {
 			} else if (!CMP_MODULE (VICE_MAINCPU) && module.major == 1) {
 				res->maincpu = R_NEW (struct vsf_maincpu);
 				r_buf_read_at (bf->buf, offset + read, (ut8 *)res->maincpu, sizeof (*res->maincpu));
+			} else {
+				char safe_name[sizeof (module.module_name) + 1];
+				memcpy (safe_name, module.module_name, sizeof (module.module_name));
```
trufae, line 109:
```suggestion
				memcpy (safe_name, module.module_name, sizeof (safe_name));
```

its better to use the sizeof of the destination even if both are the same

libr/bin/p/bin_vsf.c
```diff
@@ -96,6 +104,11 @@ static bool load(RBinFile *bf, RBuffer *b, ut64 loadaddr) {
...
+				safe_name[sizeof (module.module_name)] = 0;
```
trufae, line 110: uhm maybe you can use r_str_ncpy instead. that function always null terminates

libr/bin/p/bin_vsf.c
```diff
@@ -75,6 +77,12 @@ static bool load(RBinFile *bf, RBuffer *b, ut64 loadaddr) {
 		}
 		// read all VSF modules
 		offset = sizeof (struct vsf_hdr);
+		ut8 vice_version[sizeof (VICE_VERSION)];
+		if (r_buf_read_at (bf->buf, offset, vice_version, sizeof (VICE_VERSION)) == sizeof (VICE_VERSION)) {
+			if (!memcmp (vice_version, VICE_VERSION, sizeof (VICE_VERSION) - 1)) {
```
trufae, line 82: not sure about this -1?

trufae (changes requested):
.

libr/bin/p/bin_vsf.c
mrmacete: oh yeah thanks i always forget that

libr/bin/p/bin_vsf.c
mrmacete: that's because the string constant `VICE_VERSION` is nul-terminated, while the string in the binary is not, same as at the `CMP_MODULE` macro a few lines below

trufae: Ok! Good to confirm

PR #22548, opened 2024-01-23
Update sign.c

gogolovefish:
**Description**
fix Memory leak
Signed-off-by: wanhui@uniontech.com


libr/anal/sign.c
```diff
@@ -1405,6 +1405,10 @@ R_API RList *r_sign_find_closest_fcn(RAnal *a, RSignItem *it, int count, double
 		// turn function into signature item
 		RSignItem *fsig = r_sign_item_new ();
 		if (!fsig) {
+			if(data.bytes_combined)
+			{
+				free (data.bytes_combined);
```
trufae, line 1410: Free already checks for null no need for the if

PR #22561, opened 2024-02-03
Fix oob write for dwarf with abbrev with count 0

Crispy-fried-chicken:
The `init_die` and `parse_die` function implemented in [dwarf.c](https://github.com/radareorg/radare2/blob/master/libr/bin/dwarf.c) is quite similiar to those in the [rizin](https://github.com/rizinorg/rizin). Thus, it's very likely that Heap-based OOB write in [CVE-2021-43814](https://nvd.nist.gov/vuln/detail/CVE-2021-43814) also affects [radare2](https://github.com/radareorg/radare2).
Here is the [patch](https://github.com/rizinorg/rizin/commit/aa6917772d2f32e5a7daab25a46c72df0b5ea406?diff=split&w=1) for [CVE-2021-43814](https://nvd.nist.gov/vuln/detail/CVE-2021-43814) is provided.
Similar to [rizin](https://github.com/rizinorg/rizin), there is a heap-based out of bounds write in parse_die when reversing an amd64 elf binary with dwarf debug info, respectively.
Try to fix it based on the patch for [CVE-2021-43814](https://nvd.nist.gov/vuln/detail/CVE-2021-43814)

### DIFF
--- a/libr/bin/dwarf.c
+++ b/libr/bin/dwarf.c
@@ -1492,9 +1492,13 @@ static bool init_die(RBinDwarfDie *die, ut64 abbr_code, ut64 attr_count) {
 	if (!die) {
 		return false;
 	}
-	die->attr_values = calloc (sizeof (RBinDwarfAttrValue), attr_count);
-	if (!die->attr_values) {
-		return false;
+	if (attr_count) {
+		die->attr_values = calloc (sizeof (RBinDwarfAttrValue), attr_count);
+		if (!die->attr_values) {
+			return false;
+		}
+	} else {
+		die->attr_values = NULL;
 	}
 	die->abbrev_code = abbr_code;
 	die->capacity = attr_count;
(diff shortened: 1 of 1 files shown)


libr/bin/dwarf.c
```diff
@@ -2176,30 +2180,32 @@ static const ut8 *parse_die(RBin *bin, const ut8 *buf, const ut8 *buf_end, RBinD
...
-		bool is_valid_string_form = is_string && attribute->string.content;
-		if (attribute->attr_name == DW_AT_stmt_list) {
-			debug_line_offset = attribute->reference;
-		}
-		if (attribute->attr_name == DW_AT_comp_dir && is_valid_string_form) {
-			comp_dir = strdup (attribute->string.content);
-			r_str_ansi_strip (comp_dir);
-			r_str_replace_ch (comp_dir, '\n', 0, true);
-			r_str_replace_ch (comp_dir, '\t', 0, true);
+	if (abbrev->count) {
```
trufae, line 2183: The consition in the loop should make this condition unnecessary

trufae (commented):
Is there a reproducer for this bug? Ideally it will be good to have it, as part of the fuzz suite, at least to confirm that bug doesnt appears or can be reproduced at least

trufae:
Ah the issue is UB because calloc(0). Merging

trufae:
Feel free to request it. I never asked for a cve and  i dont care about recognizing my work i have better things to do. So i’m unsure about the burocracy involved

PR #22562, opened 2024-02-03
Use r_str_scanf as a safe alternative to fix gdb reg profile parsing bugs

Crispy-fried-chicken:
`scanf("%s")` uses are not safe, which may cause OOB, so we need to limit the number of characters read.


libr/debug/p/debug_gdb.c
```diff
@@ -12,6 +12,13 @@ typedef struct {
 #define UNSUPPORTED 0
 #define SUPPORTED 1
 
+#define PROC_NAME_SZ   1024
```
trufae, line 15: Are those constants actually coming from the kernel? If not i would prefer to have no limits on any of this

libr/debug/p/debug_gdb.c
```diff
@@ -12,6 +12,13 @@ typedef struct {
...
+#define PROC_REGION_SZ 100
+// PROC_REGION_SZ - 2 (used for `0x`). Due to how RZ_STR_DEF works this can't be
+// computed.
+#define PROC_REGION_LEFT_SZ 98
```
trufae, line 19: Precomputed values and assumptions like this are prompt to become maintainance problems for later. Comments shouldnt be needed if the code is solid

libr/debug/p/debug_gdb.c
```diff
@@ -240,8 +247,8 @@ static RList *r_debug_gdb_map_get(RDebug* dbg) { // TODO
 		// 7ffff7dda000-7ffff7dfd000 r-xp 00000000 08:05 265428 /usr/lib/ld-2.25.so
 		// Android
 		// "12c00000-12c40000 rw-p 00000000 00:00 0                                  [anon:dalvik-main space (region space)]";
-		ret = sscanf (ptr, "%s %s %"PFMT64x" %*s %*s %[^\n]", &region1[2],
-			      perms, &offset, name);
+		ret = sscanf(ptr, "%" RZ_STR_DEF(PROC_REGION_LEFT_SZ) "s %" RZ_STR_DEF(PROC_PERM_SZ) "s %" PFMT64x " %*s %*s %" RZ_STR_DEF(PROC_NAME_SZ) "[^\n]",
```
trufae, line 250: Dont use rizinisms

libr/debug/p/debug_io.c
```diff
@@ -22,8 +25,8 @@ static RList *__io_maps(RDebug *dbg) {
 	}
 	char *ostr = str;
 	ut64 map_start, map_end;
-	char perm[32];
-	char name[512];
```
trufae, line 26: Theres no need to have a define for a size that is tied to an array. Thats defined in the secure coding guidelines as a bad practice because it makes the code more xlobbered and harder to maintain and error probe

libr/debug/p/debug_io.c
```diff
@@ -47,7 +50,7 @@ static RList *__io_maps(RDebug *dbg) {
 			if (_s_) {
 				memmove (_s_, _s_ + 2, strlen (_s_));
 			}
-			sscanf (str, "0x%"PFMT64x" - 0x%"PFMT64x" %s %s",
+			sscanf(str, "0x%" PFMT64x " - 0x%" PFMT64x " %" RZ_STR_DEF(IO_MAPS_PERM_SZ) "s %" RZ_STR_DEF(IO_MAPS_NAME_SZ) "s",
```
trufae, line 53: Sscanf is an awful function and there are actually many other issues than the size of the data stored after parsing. The return value should be checked for example

trufae: Better to just check for the space and use rnum to parse the string and strdup to catch the name. This way no limits in stack bounds are needed and will make it more flexible and simple

trufae (changes requested):

libr/debug/p/debug_gdb.c
trufae: The defines shouldnt exist either. Those constants are arbitrary and the value for that one can be computed from another one. Otherwise its a maintainance complication

trufae:
I had some time to check your PR in detail. and i think the only sane way to handle sscanf with dynamic string formating in a clean and maintainable way is to not use sscanf or build the format string at runtime, because %*s is not supported by sscanf (this is printf-specific), and its not possible to use sizeof() and concat that result into the formatting string. Therefor this is probably the cleanest way I can think of right now:

```
#include <stdio.h>
#include <r_util.h>
#include <stdlib.h>

int main() {
	char str[4];
	char str2[4];
	char buf[] = "hello world";
	r_strf_var (format, 64, "%%%ds %%%ds", (int)sizeof (str), (int)sizeof (str2));
	int ret = sscanf (buf, format, str, str2);
	printf ("FMT %s\n", format);
	printf ("RET %d == 2\n", ret);
	printf ("-> %s\n", str);
	return 0;
}
```

i dont like the fact that the sizeof and str are decoupled into separate statements, and that could be probably wrapped up with amacro that constructs the string and runs sscanf, also, considering modern compilers can do printf at compile time the final code should be as simple as the original, and the construction of the format string should be done outside any loop.

But again, if its possible i would prefer to avoid the use of sscanf, and just strchr or use r_str to split the string by spaces and such instead of depending on it.

Also this PR raises concern for other bad usecases of sscanf:

```
0$ git grep sscanf| grep '%s'
libr/debug/p/debug_gdb.c:		ret = sscanf (ptr, "%s %s %"PFMT64x" %*s %*s %[^\n]", &region1[2],
libr/debug/p/debug_io.c:			sscanf (str, "0x%"PFMT64x" - 0x%"PFMT64x" %s %s",
libr/debug/p/debug_native.c:		if (sscanf (line, "%s %s %d %d 0x%s %3s %d %d",
libr/debug/p/debug_native.c:		i = sscanf (line, "%s %s %08"PFMT64x" %*s %*s %[^\n]", &region[2], perms, &offset, name);
libr/debug/p/native/linux/linux_coredump.c:		sscanf (buff,  "%d %s %c %d %d %d %d %d %u %lu %lu %lu %lu"
libr/debug/p/native/linux/linux_coredump.c:		sscanf (buff, "%d %s %c %d %d %d %d %d %u %lu %lu %lu %lu"
libr/io/p/io_self.c:		sscanf (line, "%s %s %*s %*s %*s %[^\n]", region + 2, perms, path);
libr/reg/profile.c:		ret = sscanf (ptr, " %s %d %d %d %d %s %s", name, &number, &rel,
0$
```

and also:

```
$ git grep sscanf| grep '%\*s'
libr/debug/p/debug_gdb.c:		ret = sscanf (ptr, "%s %s %"PFMT64x" %*s %*s %[^\n]", &region1[2],
libr/debug/p/debug_native.c:		i = sscanf (line, "%s %s %08"PFMT64x" %*s %*s %[^\n]", &region[2], perms, &offset, name);
libr/io/p/io_self.c:		sscanf (line, "%s %s %*s %*s %*s %[^\n]", region + 2, perms, path);
```

thanks f

trufae:
importing this https://github.com/tusharjois/bscanf is probably the best option :) ill look into that and get it as part of r_util asap

libr/debug/p/debug_gdb.c
```diff
@@ -236,12 +238,8 @@ static RList *r_debug_gdb_map_get(RDebug* dbg) { // TODO
 		r_list_free (words);
 		free (wordstr);
 #else
-		// We assume Linux target, for now, so -
-		// 7ffff7dda000-7ffff7dfd000 r-xp 00000000 08:05 265428 /usr/lib/ld-2.25.so
-		// Android
-		// "12c00000-12c40000 rw-p 00000000 00:00 0                                  [anon:dalvik-main space (region space)]";
-		ret = sscanf (ptr, "%s %s %"PFMT64x" %*s %*s %[^\n]", &region1[2],
-			      perms, &offset, name);
+		r_strf_var (format, 64, "%%%ds %%%ds %%PFMT64x %%*s %%*s %%%d [^\\n]", (int)sizeof (region1[2]), (int)sizeof (perms), (int)sizeof (offset), (int)sizeof (name));
```
trufae, line 241: PFMT64x is a define, and needs to be defined outside the string itself, so that wont work, also the %*s is not supported in sscanf, so that wont work either because that seems to be a nonportable gnuism

```suggestion
		r_strf_var (format, 64, "%%%ds %%%ds %%PFMT64x %%*s %%*s %%%d [^\\n]", (int)sizeof (region1[2]), (int)sizeof (perms), (int)sizeof (offset), (int)sizeof (name));
```

trufae (changes requested):
wait for me to push the new sscanf implementaion, it will make things much cleaner and safe and i'll add unit tests for all that, so we wont need to use PFMT64x and such anymore and %*s will work, and %s will be banned

trufae:
Thanks for the patience and time updating this PR. it's a good finding, but i want it to be addressed in the cleanest way as possible. Good work!

trufae:
Found a couple of bugs in their bscanf implementation. fixed them in my fork and extended it to support %.s (%*s is kept for backward compat), so we cant use the attribute check sadly. added some tests that may probably be helpful for those cases :)

https://github.com/radareorg/radare2/pull/22567

let me know what do you think?

i still want to cleanup this code more, and add a lot of more tests. and probably integrate it with RNum, so we can use it properly with extended math operation expressions and other useful format strings

trufae:
The new `r_str_scanf` function is merged in, do you wanna give it a try to rewrite this function using the new API instead? will be good to test the patch and maybe add some more unit tests for it. thanks!

libr/debug/p/debug_gdb.c
trufae, line 241: uhm will be good to implement this [^\\n] part in r_str_scanf, or change the code to not make it depend on this.. for several reasons.

The code you should write to make this format string should be like this:
```suggestion
		r_strf_var (format, 64, "%%%ds %%%ds %%"PFMT64x" %%*s %%*s %%%d [^\\n]", (int)sizeof (region1[2]), (int)sizeof (perms), (int)sizeof (offset), (int)sizeof (name));
```

but still. scanf scansets introduce a bunch of issues

* output buffer is not size bounded (so it will cause overflows)
* not portable, not all libc implementations support this syntax
* even if portable, does not support wide strings and is affected by locales
* scanset size is not defined by the standard, and char ranges may vary depending on implementation

in other words, scanfs scansets are a bad idea because they introduce an unnecessary complexity to scanning strings.

After some thoughs and tries i end up implementing it in the new RStr.scanf() function you can find out the implementation+ test in here:

https://github.com/radareorg/radare2/pull/22572

libr/debug/p/debug_io.c
```diff
@@ -2,6 +2,7 @@
 
 #include <r_debug.h>
 #include <r_asm.h>
+#include <r_util.h>
```
trufae, line 5: r_util already included by r_asm iirc

libr/debug/p/debug_io.c
```diff
@@ -47,8 +48,8 @@ static RList *__io_maps(RDebug *dbg) {
 			if (_s_) {
 				memmove (_s_, _s_ + 2, strlen (_s_));
 			}
-			sscanf (str, "0x%"PFMT64x" - 0x%"PFMT64x" %s %s",
-				&map_start, &map_end, perm, name);
+			r_strf_var (format, 64, "0x%%PFMT64x - 0x%%PFMT64x %%%ds %%%ds", (int)sizeof (perm), (int)sizeof (name));
+			sscanf (str, format, &map_start, &map_end, perm, name);
```
trufae, line 52: just use r_str_scanf. some notes here:

* no need to cast (int)sizeof, because our scanf handles size_t properly
* the 0x%PFMT64x thing is just 0x%Lx

please add tests for those strings before updating this, so we can be sure the formatting and parsing works well acrsoss all platforms, this is safer and cleaner than depending on libc

libr/debug/p/debug_native.c
```diff
@@ -3,6 +3,7 @@
 #include <r_userconf.h>
 #include <r_drx.h>
 #include <r_core.h>
+#include <r_util.h>
```
trufae, line 6: not necessary

libr/debug/p/debug_native.c
```diff
@@ -1091,9 +1092,8 @@ static RList *r_debug_native_map_get(RDebug *dbg) {
 		}
 #if __KFBSD__
 		// 0x8070000 0x8072000 2 0 0xc1fde948 rw- 1 0 0x2180 COW NC vnode /usr/bin/gcc
-		if (sscanf (line, "%s %s %d %d 0x%s %3s %d %d",
-				&region[2], &region2[2], &ign, &ign,
-				unkstr, perms, &ign, &ign) != 8) {
+		r_strf_var (format, 64, "%%%ds %%%ds %%d %%d 0x%%%ds %%%ds %%d %%d", (int)sizeof (region1[2]), (int)sizeof (region2[2]), (int)sizeof (unkstr), (int)sizeof (perms));
+		if (sscanf (line, format, &region[2], &region2[2], &ign, &ign, unkstr, perms, &ign, &ign) != 8) {
```
trufae, line 1096: use r_str_scanf instead of rstrvar

libr/debug/p/debug_native.c
```diff
@@ -1109,7 +1109,8 @@ static RList *r_debug_native_map_get(RDebug *dbg) {
 #else
 		ut64 offset = 0;
 		// 7fc8124c4000-7fc81278d000 r--p 00000000 fc:00 17043921 /usr/lib/locale/locale-archive
-		i = sscanf (line, "%s %s %08"PFMT64x" %*s %*s %[^\n]", &region[2], perms, &offset, name);
+		r_strf_var (format, 64, "%%%ds %%%ds %%08PFMT64x %%*s %%*s %%%d[^\n]", (int)sizeof (region[2]), (int)sizeof (perms), (int)sizeof (name));
```
trufae, line 1112: same here, use the new r_str_scanf

trufae (changes requested):
see inline comments

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -813,9 +812,7 @@ static proc_per_process_t *get_proc_process_content(RDebug *dbg) {
 		long unsigned int no_lui;
 		long int no_li;
 		int no_num;
-		sscanf (buff, "%d %s %c %d %d %d %d %d %u %lu %lu %lu %lu"
-				"%lu %lu %ld %ld %ld %ld %ld",
-			&p->pid, no_str, &p->s_name, &p->ppid, &p->pgrp, &no_num,
+		r_str_scanf (buff, "%d %.s %c %d %d %d %d %d %u %lu %lu %lu %lu %lu %lu %ld %ld %ld %ld %ld", &p->pid, sizeof (no_str), no_str, &p->s_name, &p->ppid, &p->pgrp, &no_num,
```
trufae, line 815: i added a test for this scanf expression and find out the parsing of %c was wrong in the original bscanf code. fixed it and added some more tests. hope this work now for your pr :)

https://github.com/radareorg/radare2/pull/22578

additionally, you should check for the return value for this call, (which should return 20).

Also all those no_lui, no_num, no_li... can be removed if we just use %*d or %*lu, the `*` is the supression modifier, which may make the number of arguments much shorter in this case.

libr/reg/profile.c
```diff
@@ -2,6 +2,7 @@
 
 #include <r_reg.h>
 #include <r_lib.h>
+#include <r_util.h>
```
trufae, line 5:
```suggestion
```

libr/reg/profile.c
```diff
@@ -350,8 +351,7 @@ static char *gdb_to_r2_profile(const char *gdb) {
 			r_strbuf_free (sb);
 			return false;
 		}
-		ret = sscanf (ptr, " %s %d %d %d %d %s %s", name, &number, &rel,
-			&offset, &size, type, groups);
+		ret = r_str_scanf (ptr, "%.s %d %d %d %d %.s %.s", sizeof (name), name, &number, &rel, &offset, &size, sizeof (type), type, sizeof (groups), groups);
 		// Groups is optional, others not
```
trufae, line 355:
```suggestion
		// Groups is optional, others are not
```

libr/debug/p/debug_io.c
```diff
@@ -47,8 +48,7 @@ static RList *__io_maps(RDebug *dbg) {
 			if (_s_) {
 				memmove (_s_, _s_ + 2, strlen (_s_));
 			}
-			sscanf (str, "0x%"PFMT64x" - 0x%"PFMT64x" %s %s",
-				&map_start, &map_end, perm, name);
+			r_str_scanf (str, "0x%Lx - 0x%Lx %.s %.s", &map_start, &map_end, sizeof (perm), perm, sizeof (name), name);
```
trufae, line 51: check for return value (should be 4 or <4) - i added tests for this scanf expression too

libr/debug/p/debug_io.c
trufae, line 5:
```suggestion
```

trufae (changes requested):
Thanks!

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -813,12 +812,12 @@ static proc_per_process_t *get_proc_process_content(RDebug *dbg) {
 		long unsigned int no_lui;
 		long int no_li;
 		int no_num;
-		sscanf (buff, "%d %s %c %d %d %d %d %d %u %lu %lu %lu %lu"
-				"%lu %lu %ld %ld %ld %ld %ld",
-			&p->pid, no_str, &p->s_name, &p->ppid, &p->pgrp, &no_num,
-			&no_num, &p->sid, &p->flag, &no_lui, &no_lui, &no_lui,
+		if(r_str_scanf (buff, "%d %.s %c %*d %u %*lu %*ld", &p->pid, sizeof (no_str), no_str, &p->s_name, &p->ppid, &p->pgrp, &no_num, &no_num, &p->sid, &p->flag, &no_lui, &no_lui, &no_lui,
```
trufae, line 815:
```suggestion
		if (r_str_scanf (buff, "%d %.s %c %*d %u %*lu %*ld", &p->pid, sizeof (no_str), no_str, &p->s_name, &p->ppid, &p->pgrp, &no_num, &no_num, &p->sid, &p->flag, &no_lui, &no_lui, &no_lui,
```

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -813,12 +812,12 @@ static proc_per_process_t *get_proc_process_content(RDebug *dbg) {
...
 			&no_lui, &no_lui, &no_lui, &no_li, &no_li,
-			&no_li, &p->nice, &p->num_threads);
+			&no_li, &p->nice, &p->num_threads) < 7){
```
trufae, line 817:
```suggestion
			&no_li, &p->nice, &p->num_threads) < 7) {
```

trufae:
Looks like there's still an issue when parsing the profile: `ERROR: Unable to parse "/proc/41683/maps"` can you repro in local?

trufae:
the linting errors are because of missing spaces or wrong indentations. i fixed them in my suggested fixes

trufae:
r_util.h is not necessary because its usually included from other r_*.h

libr/debug/p/debug_gdb.c
```diff
@@ -12,6 +12,7 @@ typedef struct {
 #define UNSUPPORTED 0
 #define SUPPORTED 1
 
+
```
trufae, line 15:
```suggestion
```

libr/debug/p/debug_native.c
```diff
@@ -1091,9 +1092,7 @@ static RList *r_debug_native_map_get(RDebug *dbg) {
 		}
 #if __KFBSD__
 		// 0x8070000 0x8072000 2 0 0xc1fde948 rw- 1 0 0x2180 COW NC vnode /usr/bin/gcc
-		if (sscanf (line, "%s %s %d %d 0x%s %3s %d %d",
-				&region[2], &region2[2], &ign, &ign,
-				unkstr, perms, &ign, &ign) != 8) {
+		if (r_str_scanf (line, "%.s %.s %d %d 0x%.s %.s %d %d", sizeof (region[2]), &region[2], sizeof (region2[2]), &region2[2], &ign, &ign, sizeof (unkstr), unkstr, sizeof (perms) perms, &ign, &ign) != 8) {
```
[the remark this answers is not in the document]
trufae: i mean it was like this in the original string, but i think the unkstr can be skipped by using %*s instead and remove tha argument.

libr/debug/p/debug_native.c
```diff
@@ -1109,7 +1108,7 @@ static RList *r_debug_native_map_get(RDebug *dbg) {
 #else
 		ut64 offset = 0;
 		// 7fc8124c4000-7fc81278d000 r--p 00000000 fc:00 17043921 /usr/lib/locale/locale-archive
-		i = sscanf (line, "%s %s %08"PFMT64x" %*s %*s %[^\n]", &region[2], perms, &offset, name);
+		i = r_str_scanf (line, "%.s %.s %08Lx %*s %*s %.[^\n]", sizeof (region[2]), &region[2], sizeof (perms), perms, &offset, sizeof (name), name);
```
trufae, line 1111:
```suggestion
		i = r_str_scanf (line, "%.s %.s %Lx %*s %*s %.[^\n]", sizeof (region[2]), &region[2], sizeof (perms), perms, &offset, sizeof (name), name);
```

trufae: i think our scanf doesnt support %08Lx, because %Lx just works (i have added tests for this)

libr/io/p/io_self.c
```diff
@@ -212,7 +212,9 @@ static int update_self_regions(RIO *io, int pid) {
 		}
 		path[0]='\0';
 		strcpy (region, "0x");
-		sscanf (line, "%s %s %*s %*s %*s %[^\n]", region + 2, perms, path);
+		if (r_str_scanf (line, "%.s %.s %*s %*s %*s %.[^\n]", sizeof (region[2]), region + 2, sizeof (perms), perms, sizeof (path), path) < 6) {
```
trufae, line 215:
```suggestion
		if (r_str_scanf (line, "%.s %.s %*s %*s %*s %.[^\n]", sizeof (region) - 2, region + 2, sizeof (perms), perms, sizeof (path), path) < 6) {
```

libr/debug/p/debug_gdb.c
```diff
@@ -236,12 +236,7 @@ static RList *r_debug_gdb_map_get(RDebug* dbg) { // TODO
 		r_list_free (words);
 		free (wordstr);
 #else
-		// We assume Linux target, for now, so -
-		// 7ffff7dda000-7ffff7dfd000 r-xp 00000000 08:05 265428 /usr/lib/ld-2.25.so
-		// Android
-		// "12c00000-12c40000 rw-p 00000000 00:00 0                                  [anon:dalvik-main space (region space)]";
-		ret = sscanf (ptr, "%s %s %"PFMT64x" %*s %*s %[^\n]", &region1[2],
-			      perms, &offset, name);
+		ret = r_str_scanf (ptr, "%.s %.s %Lx %*s %*s %.[^\n]", sizeof (region1[2]), &region1[2], sizeof (perms), perms, offset, sizeof (name), name);
```
trufae, line 239:
```suggestion
		ret = r_str_scanf (ptr, "%.s %.s %Lx %*s %*s %.[^\n]", sizeof (region1)-2, &region1[2], sizeof (perms), perms, offset, sizeof (name), name);
```

libr/debug/p/debug_gdb.c
```diff
@@ -236,12 +236,7 @@ static RList *r_debug_gdb_map_get(RDebug* dbg) { // TODO
 		r_list_free (words);
 		free (wordstr);
 #else
-		// We assume Linux target, for now, so -
-		// 7ffff7dda000-7ffff7dfd000 r-xp 00000000 08:05 265428 /usr/lib/ld-2.25.so
-		// Android
-		// "12c00000-12c40000 rw-p 00000000 00:00 0                                  [anon:dalvik-main space (region space)]";
-		ret = sscanf (ptr, "%s %s %"PFMT64x" %*s %*s %[^\n]", &region1[2],
-			      perms, &offset, name);
+		ret = r_str_scanf (ptr, "%.s %.s %Lx %*s %*s %.[^\n]", sizeof (region1)-2, &region1[2], sizeof (perms), perms, offset, sizeof (name), name);
```
trufae, line 239:
```suggestion
		ret = r_str_scanf (ptr, "%.s %.s %Lx %*s %*s %.[^\n]",
			sizeof (region1)-2, &region1[2], sizeof (perms), perms, offset, sizeof (name), name);
```

trufae:
sizeof (region[2]) is 1 byte, which is clearly not enough to store a string, the argument expects to take sizeof(region1) substracting the "0x" which are the first two bytes written in that string

trufae:
let me jump into the linux machine and tryout your pr and see what else could be

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -869,7 +868,10 @@ static proc_per_process_t *get_proc_process_content(RDebug *dbg) {
 	file = r_strf ("/proc/%d/coredump_filter", dbg->pid);
 	buff = r_file_slurp (file, &size);
 	if (buff) {
-		sscanf (buff, "%hx", &filter_flags);
+		if (r_str_scanf (buff, "%hx", &filter_flags) != 1) {
+			free (p);
```
trufae, line 872:
```suggestion
			free (p);
			free (buff);
```

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -813,12 +812,12 @@ static proc_per_process_t *get_proc_process_content(RDebug *dbg) {
 		long unsigned int no_lui;
 		long int no_li;
 		int no_num;
-		sscanf (buff, "%d %s %c %d %d %d %d %d %u %lu %lu %lu %lu"
-				"%lu %lu %ld %ld %ld %ld %ld",
-			&p->pid, no_str, &p->s_name, &p->ppid, &p->pgrp, &no_num,
-			&no_num, &p->sid, &p->flag, &no_lui, &no_lui, &no_lui,
+		if (r_str_scanf (buff, "%d %.s %c %*d %u %*lu %*ld", &p->pid, sizeof (no_str), no_str, &p->s_name, &p->ppid, &p->pgrp, &no_num, &no_num, &p->sid, &p->flag, &no_lui, &no_lui, &no_lui,
```
trufae, line 815:
```suggestion
		if (r_str_scanf (buff, "%d %*s %c %d %d %*d %*d %lu %ld",
			&p->pid, &p->s_name, &p->ppid, &p->pgrp,
			&p->sid, &p->flag,
```

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -813,12 +812,13 @@ static proc_per_process_t *get_proc_process_content(RDebug *dbg) {
...
-		sscanf (buff, "%d %s %c %d %d %d %d %d %u %lu %lu %lu %lu"
-				"%lu %lu %ld %ld %ld %ld %ld",
-			&p->pid, no_str, &p->s_name, &p->ppid, &p->pgrp, &no_num,
-			&no_num, &p->sid, &p->flag, &no_lui, &no_lui, &no_lui,
-			&no_lui, &no_lui, &no_lui, &no_li, &no_li,
-			&no_li, &p->nice, &p->num_threads);
+		if (r_str_scanf (buff, "%d %*s %c %d %d %*d %*d %lu %ld",
+			&p->pid, &p->s_name, &p->ppid, &p->pgrp,
+			&p->sid, &p->flag,
+			&no_li, &p->nice, &p->num_threads) < 7) {
```
trufae, line 818:
```suggestion
			&p->nice, &p->num_threads) < 7) {
```

libr/debug/p/debug_native.c
trufae, line 6:
```suggestion
```

libr/debug/p/debug_native.c
```diff
@@ -1091,9 +1091,7 @@ static RList *r_debug_native_map_get(RDebug *dbg) {
 		}
 #if __KFBSD__
 		// 0x8070000 0x8072000 2 0 0xc1fde948 rw- 1 0 0x2180 COW NC vnode /usr/bin/gcc
-		if (sscanf (line, "%s %s %d %d 0x%s %3s %d %d",
-				&region[2], &region2[2], &ign, &ign,
-				unkstr, perms, &ign, &ign) != 8) {
+		if (r_str_scanf (line, "%.s %.s %d %d 0x%.s %.s %d %d", sizeof (region[2]), &region[2], sizeof (region2[2]), &region2[2], &ign, &ign, sizeof (unkstr), unkstr, sizeof (perms) perms, &ign, &ign) != 8) {
```
trufae, line 1094:
```suggestion
		if (r_str_scanf (line, "%.s %.s %d %d 0x%.s %.s %d %d", sizeof (region)-2, &region[2], sizeof (region2)-2, &region2[2], &ign, &ign, sizeof (unkstr), unkstr, sizeof (perms) perms, &ign, &ign) != 8) {
```

trufae: same here, the whole format string and arguments were wrong.

The meaning of * in sscanf is documented in the manpage. and bascially means to skip that argument, so you dont need to pass any argument, and we can remove all the silly variables that do nothing in this code

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -813,12 +812,13 @@ static proc_per_process_t *get_proc_process_content(RDebug *dbg) {
 		long unsigned int no_lui;
 		long int no_li;
```
trufae, line 813:
```suggestion
```

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -813,12 +812,13 @@ static proc_per_process_t *get_proc_process_content(RDebug *dbg) {
 		long unsigned int no_lui;
```
trufae, line 812:
```suggestion
```

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -157,9 +158,7 @@ static proc_per_thread_t *get_proc_thread_content(int pid, int tid) {
 		int no_num;
 		char no_char;
 		ut32 no_ui;
-		sscanf (buff,  "%d %s %c %d %d %d %d %d %u %lu %lu %lu %lu"
-			"%"PFMT64x" %"PFMT64x" %ld %lu",
-			&no_num, no_str, &no_char, &no_num, &no_num, &no_num,
+		r_str_scanf (buff, "%d %.s %c %d %d %d %d %d %u %lu %lu %lu %Lx %Lx %ld %lu", &no_num, sizeof (no_str), no_str, &no_char, &no_num, &no_num, &no_num,
```
trufae, line 161: we can get rid of all those no* variables if we use %*d in the arguments instead

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -810,15 +809,14 @@ static proc_per_process_t *get_proc_process_content(RDebug *dbg) {
...
-		sscanf (buff, "%d %s %c %d %d %d %d %d %u %lu %lu %lu %lu"
-				"%lu %lu %ld %ld %ld %ld %ld",
-			&p->pid, no_str, &p->s_name, &p->ppid, &p->pgrp, &no_num,
-			&no_num, &p->sid, &p->flag, &no_lui, &no_lui, &no_lui,
-			&no_lui, &no_lui, &no_lui, &no_li, &no_li,
-			&no_li, &p->nice, &p->num_threads);
+		if (r_str_scanf (buff, "%d %*s %c %d %d %*d %*d %lu %ld",
+			&p->pid, &p->s_name, &p->ppid, &p->pgrp,
+			&p->sid, &p->flag,
+			&p->nice, &p->num_threads) < 7) {
```
trufae, line 816:
```suggestion
			&p->nice, &p->num_threads) < 6) {
```

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -810,15 +809,14 @@ static proc_per_process_t *get_proc_process_content(RDebug *dbg) {
 	/* we only need few fields which are process-wide */
 	{
 		char no_str[128];
-		long unsigned int no_lui;
-		long int no_li;
 		int no_num;
```
trufae, line 812:
```suggestion
```

libr/debug/p/native/linux/linux_coredump.c
```diff
@@ -810,15 +809,14 @@ static proc_per_process_t *get_proc_process_content(RDebug *dbg) {
 	/* we only need few fields which are process-wide */
 	{
 		char no_str[128];
```
trufae, line 811:
```suggestion
```

trufae:
i just tested your pr in local and i missed another wrong sizeof(region[2]). i have squahed the 32 commits into this separate PR and merged it under your name.

Thank you!

trufae:
This is the PR https://github.com/radareorg/radare2/pull/22599

PR #22564, opened 2024-02-04
Flip boolean condition for di command ##debug

zutle:

**Description**
This PR flips a boolean condition checking if we are currently in debugging mode (flag `cfg.debug` set). 

Currently, the `di` command only works when we are _not_ in debug mode, and it should be the other way around :)


trufae:
Oops! Good catch

PR #22611, opened 2024-02-17
Support whole address space for IO

davispuh:

**Description**

When debugging recent Windows 10 x64 kernel then it maps kernel modules very high in virtual memory (above `0x7fffffffffffffff`, eg. `0xfffff8018f920000`) that means it's not addressable using `ST64_MAX`.

And that is causing `oba` to not work due to such high address space.

This PR fixes that by:
* Reserving `UT64_MAX` for error case in `RBufferSeek()` 
* Updating  `ST64_MAX` to `UT64_MAX - 1` for `R_IO_SEEK_END`  in `winkd`
* Updating  `UT64_MAX` to `UT64_MAX - 1` for other plugins - they're broken in similar way and would fail above ST64_MAX anyway so this actually increases addressable space
* Use `buf_seek() != -1` instead of `buf_seek() >= 0` because `buf_seek` uses `st64` while `RBufferSeek` uses `ut64` - probably better would be to convert `buf_seek()` to `ut64` aswell but that would be quite massive change...
* Fix PE plugin to be able to handle huge buffers - this is needed because when loaded from IO plugin you don't actually know true size and you get `UT64_MAX - 1 - offset` as PE size...

PS. Tracking this down was incredibly painful...

### DIFF
--- a/libr/bin/format/pe/pe.c
+++ b/libr/bin/format/pe/pe.c
@@ -642,7 +642,7 @@ error:
 
 int PE_(read_dos_header)(RBuffer *b, PE_(image_dos_header) *header) {
 	st64 o_addr = r_buf_seek (b, 0, R_BUF_CUR);
-	if (r_buf_seek (b, 0, R_BUF_SET) < 0) {
+	if (r_buf_seek (b, 0, R_BUF_SET) == -1) {
 		return -1;
 	}
 	header->e_magic = r_buf_read_le16 (b);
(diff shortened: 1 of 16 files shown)


trufae:
Would be good to use the ##io tag in the first commit and the ##bin one in the second. This way the changes will appear in the release notes.

and maybe you can add a test by using ARGS=-B 0xff000.. but apart from that the changes look good to me

trufae:
Thank you! good fixes! <3

PR #22612, opened 2024-02-17
Crosscompile for Windows arm64

trufae:

**Description**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -417,6 +417,50 @@ jobs:
         name: w64-static-2022
         path: prefix\bin\r2blob.static.exe
 
+  w64-arm64-static-2022:
+    runs-on: windows-2022
+    steps:
+    - name: Win configure Pagefile
+      uses: al-cheb/configure-pagefile-action@v1.3
+      with:
+          minimum-size: 16GB
+          maximum-size: 16GB
+          disk-root: "C:"
+    - name: Checkout
+      uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
+    - name: Preparing nmake
+      uses: ilammy/msvc-dev-cmd@v1
+      with:
+        arch: x64
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


trufae:
seems like the arm64 binary is actually an x64 one 😅

.github/workflows/build.yml
```diff
@@ -417,6 +417,50 @@ jobs:
...
+          disk-root: "C:"
+    - name: Checkout
+      uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
+    - name: Preparing nmake
+      uses: ilammy/msvc-dev-cmd@v1
+      with:
+        arch: x64
```
prodrigestivill, line 437: I think `arch` here needs to be `amd64_arm64`.

preconfigure.bat
```diff
@@ -49,7 +49,7 @@ REM: Targets: x86 amd64 x64 arm arm64
 IF "%*" == "x86" (
   set VSARCH=x86
 ) ELSE IF "%*" == "arm64" (
-  set VSARCH=x86_arm64
+  set VSARCH=x64_arm64
```
prodrigestivill, line 52: And here also `amd64_arm64` for crosscompile?

prodrigestivill:
Also using [meson cross-compile](https://mesonbuild.com/Cross-compilation.html) feature will be required to avoid a sanity check error.

.github/workflows/build.yml
trufae, line 437:
```suggestion
        arch: amd64_arm64
```

trufae:
> Also using [meson cross-compile](https://mesonbuild.com/Cross-compilation.html) feature will be required to avoid a sanity check error.

never used that before. did you?

```
meson.build:1:0: ERROR: Could not invoke sanity test executable: [WinError 216] This version of %1 is not compatible with the version of Windows you're running. Check your computer's system information and then contact the software publisher.
```

prodrigestivill:
yes, I arrived at the same point in my fork.

this is because meson is checking as if you are building for native environment.
To avoid it it is requiered to use the crosscompile functionality of meson, just like it is done for android builds (using a configuration file per env).

I tried in my fork but it would be easier to implement it having a local windows build environment.

PR #22625, opened 2024-02-19
Update documentation for macOS/arm64e ##doc

trufae:

**Description**

### DIFF
--- a/doc/macos.md
+++ b/doc/macos.md
@@ -1,70 +1,88 @@
-macOS
-===
+# macOS
 
 macOS Users need to follow some extra steps to get the radare2 program signed and ready to debug other applications without running it as root. Same happens for iOS users, read `doc/ios` for more information.
 
-Installation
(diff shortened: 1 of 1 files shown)


mrmacete:
maybe write explicitly that the problem arises only if you need to use the debugger, no need for it for static analysis only

doc/macos.md
```diff
@@ -1,70 +1,88 @@
...
-	$ sys/install.sh
+```sh
+$ sys/install.sh
+```
 
-By default it is installed in /usr/local, you can specify a different prefix like this:
+## Common Issues
 
-	$ sys/install.sh /custom/prefix
+### Arm64e targets
```
mrmacete, line 15: maybe something like "Arm64 debug targets"

trufae, line 15:
```suggestion
### Arm64e debug targets
```

PR #22634, opened 2024-02-22
refactor resolve_glibc_version, improve libc regex

koyaan:

**Description**

Breaks out the function to get libc filename from memory maps since it is gonna be needed in other functions later.
Also changes the regex for finding libc. (inspired by pwndbg)

### DIFF
--- a/libr/core/dmh_glibc.inc.c
+++ b/libr/core/dmh_glibc.inc.c
@@ -110,7 +110,7 @@ static GH(section_content) GH(get_section_content)(RCore *core, const char *path
 	RBinFileOptions opt;
 	r_bin_file_options_init (&opt, -1, 0, 0, false);
 	if (!r_bin_open (bin, path, &opt)) {
-		R_LOG_ERROR ("section_content: r_bin_open failed on path %s", path);
+		R_LOG_ERROR ("get_section_content: r_bin_open failed on path %s", path);
 		return content;
 	}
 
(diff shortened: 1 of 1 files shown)


koyaan:
```
[XX] db/cmd/cmd_dmh dmh crash without arena/heap map
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -NN -Qc 'dmh
o malloc://512 0x20000 rwx
omn 0x20000 arena
dmh
o malloc://512 0x10000 rwx
omn 0x10000 [heap]
dmh
' -
-- exit status: -1
```
i was under the impression that `r_return_val_if_fail (core && core->dbg && core->dbg->maps, NULL);` was enough to protect  `r_list_foreach (core->dbg->maps, iter, map) {` but apparently it's not.

What is the reason to have `dmh` if you are not debugging anyway? 

Should be just check `cfg.debug` and return false if not set?

libr/core/dmh_glibc.inc.c
```diff
@@ -214,13 +214,30 @@ R_API double GH(get_glibc_version)(RCore *core, const char *libc_path) {
 	return version;
 }
 
+static const char* GH(get_libc_filename_from_maps)(RCore *core) {
+	RListIter *iter;
+	RDebugMap *map = NULL;
+
+	r_return_val_if_fail (core && core->dbg && core->dbg->maps, NULL);
```
trufae, line 221: move to the first line, before variabl definitions

trufae:
Looks good now!

PR #22655, opened 2024-02-28
Fix leak in canvas.c

ashamedbit:

**Description**

This patch fixes a memory leak within `r_cons_canvas_resize`.  In one location the authors manually free memory of individual fields and even miss fields such as `bgcolor` of `RConsCanvas`. Please do note that entire `c->b` is deallocated by authors. The patch instead frees using `r_cons_canvas_free` which includes free for `bgcolor`.


trufae:
woah neat spot!

Issue #8666, opened 2017-10-09
Symbols in /usr/lib/debug aren't automatically loaded


lucianposton:
See https://github.com/radare/radare2/issues/4888 for context.

Several linux systems installs split debug symbols from executables and installs them under `/usr/lib/debug/` e.g. `/usr/path/to/bin/cmd` has symbols at `/usr/lib/debug/usr/path/to/bin/cmd`. radare2 1.6.0 doesn't appear to load these symbols. The symbols for calls are missing, attemping to seek, `s`, to a function fails, etc. On the other hand, gdb automatically loads the symbols from these files.

lucianposton:
Also, `ob /usr/lib/debug/usr/path/to/bin/cmd` doesn't appear to manually load the symbols either.

XVilka:
Code from https://github.com/radare/radare2/blob/master/libr/core/linux_heap_glibc.c#L213 should be generalized for using anywhere in RBin.

radare:
same goes for jemalloc

lucianposton:
In the meantime, is there a workaround to manually load symbols in version 2.0.0?

`o /usr/lib/debug/usr/path/to/bin/cmd` doesn't work. After issuing the preceding command, there is no error output, no new files are listed by `o`, and no symbols appear to have been loaded.

tbarabosch:
@lucianposton I care too know as well. Did you find a solution?

radare:
.!rabin2 -r ....

But you can use the ob subcommands too

lucianposton:
> I care too know as well. Did you find a solution?

Nope. The ob command didn't load any symbols as far as I could tell.

I haven't yet tried the .!rabin2 -r ... suggestion above.

radare:
Ob by itself doesnt do anything. Check the help message

lucianposton:
oops. I meant to say `o /usr/lib/debug/usr/path/to/bin/cmd` didn't load the symbols, and `ob`, which is supposed to list opened files, did lis anything new after `o`.

radare:
Nope, o opens a binary, replacing the current one, so nope

Ob doesnt lists the opened files. It lists the bin objects and u can use it to load the bin info of that bin on top of the current opened file

Also, historically this has been done via .!rabin2... since the very first versions of r1.

Which distro are u using to have test this?

radare:
related to the dmi issue

agausmann:
A workaround I've found is to use `eu-unstrip` from elfutils:

[...]

PR #22668, opened 2024-03-05
add new way to resolve main_arena ##debug

koyaan:

This adds a new way to resolve `main_arena` address  in libc `get_main_arena_offset_with_relocs` which does a clever thing with relocations directly ported from pwndbg (https://github.com/pwndbg/pwndbg/blob/da3f0f3b26ffaa3c1680340a6e13731603c1bdb9/pwndbg/heap/ptmalloc.py#L1526).
This also cuts down the functionality of `get_main_arena_with_symbol` by _only_ looking for `main_arena` and failing if it is not there. The method of just taking an offset of `__malloc_hook` produced a lot of false results, and this symbol might be even removed since malloc hooks are deprecated.

There is still work to do to make this function with different versions of libc, this only works for >= 2.27 right now.
Also non-pie and static binaries need a different approach.

### DIFF
--- a/libr/core/dmh_glibc.inc.c
+++ b/libr/core/dmh_glibc.inc.c
@@ -71,33 +71,15 @@ static inline GHT GH(get_next_pointer)(RCore *core, GHT pos, GHT next) {
 	return (core->dbg->glibc_version < 232) ? next : PROTECT_PTR (pos, next);
 }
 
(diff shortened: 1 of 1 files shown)


trufae:
@koyaan can you take a look at this issue? https://github.com/radareorg/radare2/issues/8666#issuecomment-1756662752

koyaan:
@trufae  Yes! I already had this on my radar

PR #22669, opened 2024-03-05
Fix type confusion in pyc ##arch

dennis-doyensec:

**Description**

The code assumes py objects are of a type that contains a r_list in it's data. This may not be true, so this verifies that a list is available. A deeper check may be necessary that the list contains intended objects, I did not get that far.


trufae:
Thank you! This code needs quite a bit more love but merging it as long as it’s improving the situation already

PR #22672, opened 2024-03-06
Handle URL safe base64 decoding

sylvainpelissier:

**Description**

According to [RFC 4648 §5](https://datatracker.ietf.org/doc/html/rfc4648#section-5) another way of doing base64 encoding  is using "-" instead of "+" and "_" instead of "/". This modification allows r2 to decode such strings.


trufae:
Beautiful! Merged!

PR #22674, opened 2024-03-08
Fix iS. paddr vaddr confusion ##core

mrmacete:

**Description**

This change removes the part of the `printHere` address check where it explicitly checked for `paddr` even if the paddr / vaddr dichotomy is already handled via the `addr` variable, which is the only one we have to check here.

That caused wrong behaviour in binaries where the vaddr and paddr space numerically overlap (like old DOS/4GW LE binaries) where `iS.` was in fact picking the `paddr` instead of `vaddr` to filter if such an entry came first in the list.

### DIFF
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -3088,8 +3088,7 @@ static bool bin_sections(RCore *r, PJ *pj, int mode, ut64 laddr, int va, ut64 at
 			continue;
 		}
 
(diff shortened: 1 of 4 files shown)


mrmacete:
draft, i'm going to check the tests soon

mrmacete:
new test needs https://github.com/radareorg/radare2-testbins/pull/94

PR #22681, opened 2024-03-12
Ensure termux installation script only runs from root directory

0xXA:
This fixes problems when script is invoked from directories other than the radare's root directory.


trufae:
Thank you! Merged

PR #22694, opened 2024-03-15
Guard all reallocs in io.map

trufae:
* Multiplication can cause overflow and realloc can fail

**Description**

### DIFF
--- a/libr/io/io_map.c
+++ b/libr/io/io_map.c
@@ -1,4 +1,4 @@
-/* radare2 - LGPL - Copyright 2017-2023 - condret, MaskRay */
+/* radare2 - LGPL - Copyright 2017-2024 - condret, MaskRay */
 
 #include <r_io.h>
 #include <stdlib.h>
@@ -594,8 +594,15 @@ R_API bool r_io_map_write_to_overlay(RIOMap *map, ut64 addr, const ut8 *buf, int
 	}
 	if (r_itv_begin (chunk->itv) < r_itv_begin (search_itv)) {
 		chunk->itv.size = r_itv_begin (search_itv) - r_itv_begin (chunk->itv);
-		// realloc cannot fail here because the new size is smaller than the old size
-		chunk->buf = realloc (chunk->buf, r_itv_size (chunk->itv) * sizeof (ut8));
+		st64 new_size = r_itv_size (chunk->itv) * sizeof (ut8);
+		if (new_size > 0) {
+			ut8 *chunk_buf = realloc (chunk->buf, new_size);
+			if (chunk_buf) {
+				chunk->buf = chunk_buf;
+			} else {
+				// Chunk reallocation failed
+			}
+		}
 		node = r_rbnode_next (node);
 	}
 	if (node) {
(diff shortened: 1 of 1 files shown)


libr/io/io_map.c
```diff
@@ -594,8 +594,15 @@ R_API bool r_io_map_write_to_overlay(RIOMap *map, ut64 addr, const ut8 *buf, int
 	}
 	if (r_itv_begin (chunk->itv) < r_itv_begin (search_itv)) {
 		chunk->itv.size = r_itv_begin (search_itv) - r_itv_begin (chunk->itv);
-		// realloc cannot fail here because the new size is smaller than the old size
-		chunk->buf = realloc (chunk->buf, r_itv_size (chunk->itv) * sizeof (ut8));
+		st64 new_size = r_itv_size (chunk->itv) * sizeof (ut8);
+		if (new_size > 0) {
+			ut8 *chunk_buf = realloc (chunk->buf, new_size);
+			if (chunk_buf) {
```
condret, line 600: cannot fail, because new size is always smaller than old size. realloc cannot run out of memory here

libr/io/io_map.c
```diff
@@ -606,17 +613,20 @@ R_API bool r_io_map_write_to_overlay(RIOMap *map, ut64 addr, const ut8 *buf, int
 			chunk = node? (MapOverlayChunk *)node->data: NULL;
 		}
 		if (chunk && r_itv_end (search_itv) >= r_itv_begin (chunk->itv)) {
-			ut8 *ptr = realloc (chunk->buf,
-				(r_itv_end (chunk->itv) - r_itv_begin (search_itv)) * sizeof (ut8));
-			if (!ptr) {
-				return false;
+			ut64 new_size = r_itv_end (chunk->itv) - r_itv_begin (search_itv);
+			st64 new_size2 = new_size * sizeof (ut8);
+			if (new_size > 0) {
```
condret, line 618: always true, because of check in line 615

libr/io/io_map.c
```diff
@@ -693,7 +703,8 @@ R_API void r_io_map_drain_overlay(RIOMap *map) {
 			if (!r_queue_is_empty (q)) {
 				MapOverlayChunk *start = (MapOverlayChunk *)start_n->data;
 				const ut64 new_size = r_itv_end (cur->itv) - r_itv_begin (start->itv);
-				ut8 *buf = realloc (start->buf, new_size * sizeof (ut8));
+				const st64 new_size2 = new_size * sizeof (ut8);
```
condret, line 706: realloc expects size_t which is unsigned, so let's avoid passing signed integers to it. realloc (ptr, 0) is UB, since c23. so we'd only need to check for that specific case. i suggest something like this above the original call to realloc:

```
if (!(newsize * sizeof (ut8))) {
   start_n = cur_n = next_n;
   next_n = r_rbnode_next (next_n);
   continue;
}
```

condret (commented):
Overall i'd prefer to not merge this. the additional checks seem unnecessary. Individual chunks in the tree are initially size limited to ST32_MAX * sizeof (ut8), because of the function signature of r_io_map_write_to_overlay. In theory one could create a situation in which multiple chunks get merged into a huge chunk, where size could overflow. however there'd be already several petabytes of memory allocated on the heap for that to happen. i know of no system with that amount of resources, that r2 runs on. so existing checks on the return value of realloc already stop the construction of these huge chunks.

given that realloc (ptr, 0) is UB in c23, maybe there should be some check for that. I'll investigate further to see, if this is needed and do it in case

libr/io/io_map.c
trufae: thats up to the heap implementation. better dont assume what the heap allocator does before hand. not all builds of r2 use glibc

libr/io/io_map.c
```diff
@@ -594,8 +594,15 @@ R_API bool r_io_map_write_to_overlay(RIOMap *map, ut64 addr, const ut8 *buf, int
 	}
 	if (r_itv_begin (chunk->itv) < r_itv_begin (search_itv)) {
 		chunk->itv.size = r_itv_begin (search_itv) - r_itv_begin (chunk->itv);
-		// realloc cannot fail here because the new size is smaller than the old size
-		chunk->buf = realloc (chunk->buf, r_itv_size (chunk->itv) * sizeof (ut8));
+		st64 new_size = r_itv_size (chunk->itv) * sizeof (ut8);
+		if (new_size > 0) {
```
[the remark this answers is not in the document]
trufae: uhm actually sizeof(ut8) is 1. whats the point on doing that?

libr/io/io_map.c
trufae: cant be zero? because in the sample object we tested yesterday it was donig some realloc(0) which is UB

libr/io/io_map.c
trufae: the signe dcheck is to detec the overflow in the multiplication. thing is that i didnt though about the useless multiplication, so yeah it should be unsigned indeed

libr/io/io_map.c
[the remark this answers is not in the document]
condret: oh, yeah, that's a mistake on my side, i completely forgot that we're using stdint instead of defining ut8 as unsigned char

libr/io/io_map.c
condret: uhm, ok

libr/io/io_map.c
condret: that got fixed

libr/io/io_map.c
trufae: so wat du, do you wanna finish this pr?

trufae:
@condret anything else to care  from this pr @condret ?

PR #22734, opened 2024-03-22
Use size or vsize consistently in iS. ##core

mrmacete:

**Description**

Now the size is virtual or physical consistently with the chosen address. This makes it possible to get correct ouput for `iS.` on addresses belonging to sections which have different `vsize` than `size` (like for example bss sections).


trufae:
Maybe worth adding a test with ARGS=-e io.va=0 too

mrmacete:
right! doing

Issue #22742, opened 2024-03-23
implement handling of R_X86_64_TPOFF64 relocation


koyaan:
## implement handling of R_X86_64_TPOFF64 relocation

This is "Offset in initial TLS block" relocation and newer libc have it apparently and we get a lot of spam resolving `main_arena` via relocations because this is not handled. 
 
Should go somewhere here  i guess https://github.com/radareorg/radare2/blob/180fea967b4f42ef91a4c31f5092c71989d086a1/libr/bin/p/bin_elf.inc.c#L533

more info:  https://chao-tic.github.io/blog/2018/12/25/tls

koyaan:
@trufae  i will try to whip up a test binary tonight

trufae:
Some random hints:

* This relocation type subtracts the thread pointer (usually %fs or %gs register in x86-64) from the symbol's offset within the TLS segment.
* There's also a 32 bit form of the same reloc:
* Sample instruction 'movl %fs:a@tpoff, %eax'
* In that case the instruction is encoded like this: 648b0425XXXXXXXX

We just need to save the absolute value of the reloc in the instruction + 4

the 64bit form of movl is movq, which is the same, but starting with 6448b0 instead of 64b0, and the offset is +5

in theory this offset+4 or offset+5 should be implicit by the reloc rva. but will be good to verify. so if all this theory is correct. that patch should be enough: can you try it out and update your "supress reloc warning" PR with the proper implementation? thank you!

```
diff --git a/libr/bin/p/bin_elf.inc.c b/libr/bin/p/bin_elf.inc.c
index 756eaa2db8..af2099533c 100644
--- a/libr/bin/p/bin_elf.inc.c
+++ b/libr/bin/p/bin_elf.inc.c
@@ -844,6 +844,11 @@ static void _patch_reloc(ELFOBJ *bo, ut16 e_machine, RIOBind *iob, RBinElfReloc
  			r_write_le32 (buf, v);
 			iob->overlay_write_at (iob->io, rel->rva, buf, 4);
 			}
+			break;
+		case R_X86_64_TPOFF32:
+			word = 4;
+			V = S + A;
+			break;
  		default:
  			break;
  		}
@@ -851,6 +856,10 @@ static void _patch_reloc(ELFOBJ *bo, ut16 e_machine, RIOBind *iob, RBinElfReloc
 	case EM_X86_64: {
 		int word = 0;
 		switch (rel->type) {
+		case R_X86_64_TPOFF64:
[...]

PR #22743, opened 2024-03-23
suppress R_X86_64_TPOFF64 relocation spam

koyaan:
This just suppresses error messages we get because R_X86_64_TPOFF64 relocations are not implemented and newer libc uses them. 
And since we resolve `main_arena` via relocations we get a lot of errors.

The issue to really fix this is https://github.com/radareorg/radare2/issues/22742

### DIFF
--- a/libr/bin/p/bin_elf.inc.c
+++ b/libr/bin/p/bin_elf.inc.c
@@ -727,7 +727,10 @@ static RList* relocs(RBinFile *bf) {
 		if (already_inserted) {
 			continue;
 		}
-
+		// TODO: implement this, this stops spam for glibc main_arena resolving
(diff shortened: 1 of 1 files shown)


trufae:
The idea behind spamming those relocs is to make users be anoyed enough to care about implementing them

trufae:
Hiding the warning doesnt fixes the problem

koyaan:
I know thats why i also opened the issue, it's just very annoying because i spams the same reloc like 30 times

trufae:
object files have a lot of unimplemented relocs. so supressing the warnings is not the way to go sorry. it is easy to implement, but i need a binary to repro at least

PR #22763, opened 2024-03-28
WIP: Fix #22576 - Honor asm.emu in pdfj ##disasm

### LINKED ISSUE #22576: Include refs on rcall and ucall on pdfj command
## Description

* the whole story is described in this long (sorry) issue: https://github.com/radareorg/radare2/issues/22571#issuecomment-1933797133

* basically, including refs on `rcall` / `ucall` would save us form implementing a memory/register tracer

* is it planned?

trufae:

**Description**

### DIFF
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -6978,6 +6978,49 @@ R_API int r_core_print_disasm_json(RCore *core, ut64 addr, ut8 *buf, int nb_byte
 	const bool be = R_ARCH_CONFIG_IS_BIG_ENDIAN (core->rasm->config);
 
 	r_cons_break_push (NULL, NULL);
+	{ /* used by asm.emu */
+		r_reg_arena_push (core->anal->reg);
+		ds_print_esil_anal_init (ds);
+	}
+	if (ds->show_emu) {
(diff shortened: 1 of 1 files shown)


trufae:
This is actually what we have in pdJ, isnt?  cc @attilamester

attilamester:
Not exactly. Since `pdJ` returns what pd but in json, even with `e asm.emu=true`, the instruction in question (https://github.com/radareorg/radare2/issues/22576#issuecomment-1938634219) looks like this in pdJ:
```
{"offset":4219569,"text":"│           0x004062b1      ffd7           call edi                    ; esp=0xffffffe8 ; eip=0x3020c reloc.KERNEL32.dll_HeapAlloc"}
```

which actually does contain the dynamic address, but not in the way pdf contains at `refs` section with simple hex address.

trufae:
Ok so the text is there but not the ref. This can be improved i think to extend the current json with that info. So we dont make pdj more complicated. Sgty?

trufae:
I think that works like you want  now :) but i pushed the code in another pr

Screenshot 2024-03-30 at 20 46 59

attilamester:
Thanks! Super!

PR #22764, opened 2024-03-29
Rebase LE objects independently ##bin

mrmacete:

**Description**

This change introduces the optional environment variable `R_LE_BASES` which can contain a `:`-separated list of addresses where the base of each defined object should be. This can be used to easily match the same memory layout as a running DOS system can have while debugging.

The env variable is not great, but it's needed since each object is rebased independently, instead of "rigidly translated" like is more common on modern systems where the existing concept of a single base address would suffice.

### DIFF
--- a/libr/bin/format/le/le.c
+++ b/libr/bin/format/le/le.c
@@ -56,6 +56,14 @@ static const char *__get_arch(RBinLEObj *bin) {
 	}
 }
 
+static ut64 get_object_base(RBinLEObj * bin, size_t idx) {
+	r_return_val_if_fail (bin && bin->header && idx < bin->header->objcnt, 0);
+	if (idx < bin->n_bases) {
+		return bin->obj_bases[idx];
+	}
+	return bin->objtbl[idx].reloc_base_addr;
+}
+
 static char *__read_nonnull_str_at(RBuffer *buf, ut64 *offset) {
 	ut8 size = r_buf_read8_at (buf, *offset);
 	size &= 0x7F; // Max is 127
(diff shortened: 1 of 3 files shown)


trufae:
if you want to go for the env var appraoch i would suggest to prefix it as RABIN2_ instead (this is from rabin2 -h) but maybe this feature makes sense to be used by other multibin binaries like ar, jar, multidex, .. im happy to merge it like this but if we need to break the abi i would prefer to do it asap before 5.9

Screenshot 2024-03-29 at 19 04 29

trufae:
So my proposal would be:

* Use RABIN2_LE_BASES instead of R_LE_BASES
* Make this concept "global" for all fileformats via config var and a field in a struct, etc

Any of those options are valid to me, probably the first one is fine for a quick merge.

libr/bin/format/le/le.c
```diff
@@ -721,10 +729,44 @@ R_IPI void r_bin_le_free(RBinLEObj *bin) {
...
 		free (bin->filename);
+		free (bin->obj_bases);
 		free (bin);
 	}
 }
 
+static void parse_obj_bases_from_env(RBinLEObj *bin) {
+	ut64 * bases = NULL;
+	RList * bases_str = NULL;
+	char * bases_raw = r_sys_getenv ("R_LE_BASES");
```
trufae, line 740:
```suggestion
	char * bases_raw = r_sys_getenv ("RABIN2_LE_BASES");
```

trufae (changes requested):
.

mrmacete:
> So my proposal would be:
> 
> * Use RABIN2_LE_BASES instead of R_LE_BASES
> * Make this concept "global" for all fileformats via config var and a field in a struct, etc
> 
> Any of those options are valid to me, probably the first one is fine for a quick merge.

agree! let's rename for now, the better solution would be a global concept but requires more discussion and more work which requires time and can be done after the release 🙌

trufae:
Looks good!

PR #22774, opened 2024-03-30
Fix RVec for gcc4 forward declaration not supported ##build

trufae:

**Description**


trufae:
probably not worth mainatining support for such an old and broken compiler as long as these systems have support for more modern compilers. lets keep thing simple fo rnow

PR #22820, opened 2024-04-11
Show ASAN in -v options when compiling via `gcc` ##shell

qoelet:

**Description**

This adds support for reflecting the status of sanitized builds compiled with `gcc` < 14 when running `-v`.


trufae:
Thanks for your first contrib!

PR #22840, opened 2024-04-16
Update preconfigure.bat

mriamnobody:
Made selection of Host and target platform more simple and intuitive

**Description**

In previous code, if a user didn't provide any argument with the command then x86 will be selected as Host by default which is kind of Flawed.

This change will detect the Host system on its and will allow the user to select the target system

### DIFF
--- a/preconfigure.bat
+++ b/preconfigure.bat
@@ -1,5 +1,7 @@
 :: This is a comment
 @echo off
+SETLOCAL EnableDelayedExpansion
+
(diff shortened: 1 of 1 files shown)


mriamnobody:
I'm sorry for closing this PR. Will add another PR with a simple correction. This code didnt had the logic to deal with a case if both host and target are same resulting in error. New PR with same will be added immediately with same title. I apologize for my confusion

PR #22842, opened 2024-04-16
Update preconfigure.bat

mriamnobody:
In previous code, if a user didn't provide any argument with the command then x86 will be selected as Host by default which is kind of Flawed.

This change will detect the Host system on its and will allow the user to select the target system

**Description**

### DIFF
--- a/preconfigure.bat
+++ b/preconfigure.bat
@@ -1,5 +1,7 @@
 :: This is a comment
 @echo off
+SETLOCAL EnableDelayedExpansion
+
 :: Preconfigure script for Windows
 
 echo === Finding Python...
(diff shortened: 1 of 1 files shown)


mriamnobody:
> Now that you are messing with the windows build scripts, would you like to take a look at this ticket to do the windows-arm64 builds in the CI so we can have those artifacts in the next release? #22612
> 
> thanks!

Will definitely give it a look.

mriamnobody:
> Now that you are messing with the windows build scripts, would you like to take a look at this ticket to do the windows-arm64 builds in the CI so we can have those artifacts in the next release? #22612
> 
> thanks!

I think the last update to preconfigure.bat should have solved this issue, as I ran `preconfigure.bat` with a fork, and `VSARCH` was successfully set to the selected value.

mriamnobody:
If you carefully observe the old code:

```
IF "%*" == "x86" (
  set VSARCH=x86
) ELSE IF "%*" == "arm64" (
  set VSARCH=x86_arm64
) ELSE (
  set VSARCH=x86_amd64
)
```
If the user does not specify an architecture as an argument while running `preconfigure.bat`, `x86_amd64` will be selected by default, regardless of the host platform.

Also, if you provide the argument arm64 while running preconfigure.bat, the host architecture x86 will be selected, even if you prefer or your architecture is `x64`.

The new code will avoid all this hassle and give users the ability to select the required platforms.

mriamnobody:
If a user specifies an architecture as an argument while running `preconfigure.bat`, that argument will be set as the architecture for building. If no arguments are provided, the user will be presented with a menu to select the target system.

The host architecture is detected by the script using `%PROCESSOR_ARCHITECTURE%`. This step also helps to avoid burdening or confusing new users.

trufae:
@mriamnobody i think the blocker was that meson required a config file for crosscompiling, here's the latest attempt from @prodrigestivill to get this to work https://github.com/prodrigestivill/radare2/commit/1e174f95196b6ad19b0b521cc700c26df345283d you can continue your patch from there. i'll find some time later today to try in local, but afaik it's "just" a "ci issue"

prodrigestivill:
Also is usefull this commit: 
https://github.com/prodrigestivill/radare2/commit/586ac9721abdcd26f6d37a28f1768eb92bcab702

Not sure why without this, the script doesn't detect the vscode. I don't have a windows environment to test it.

PR #22851, opened 2024-04-18
Fix concurrency issue ##test

kyufie:

### **Description**
There's a bit of a problem with r2r on Android aarch64.
```
$ r2r test/db/asm
INFO: Running from /data/data/com.termux/files/home/project/radare2/test
Already up to date.
Loaded 10298 tests.
INFO: Skipping json tests because jq is not available
[41/10298]                 41 OK         0 BR        0 XX        0 FXFORTIFY: pthread_mutex_unlock called on a destroyed mutex (0xb400007303ff884c)
[1]    6672 abort      r2r test/db/asm
```
This patch makes sure that sigchild thread will fire a signal to the subprocess wait func only after it has done with the lock.

Also a small fix about unlocking an unlocked lock.

### DIFF
--- a/binr/r2r/run.c
+++ b/binr/r2r/run.c
@@ -463,9 +463,9 @@ static RThreadFunctionRet sigchld_th(RThread *th) {
 				proc->ret = -1;
 			}
 			ut8 r = 0;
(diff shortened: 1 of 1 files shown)


trufae:
Good catch!

PR #22858, opened 2024-04-20
Fix concurrency issue (improves 71f57c6c) ##test

kyufie:

### **Description**
I realized that I left a few things out in my previous PR (#22851).
This should provide a better solution.

### DIFF
--- a/binr/r2r/run.c
+++ b/binr/r2r/run.c
@@ -463,9 +463,9 @@ static RThreadFunctionRet sigchld_th(RThread *th) {
 				proc->ret = -1;
 			}
 			ut8 r = 0;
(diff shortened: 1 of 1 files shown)


kyufie:
@trufae Completely unrelated question, what does `oba` actually do?
From what I can tell from the description, it loads binary informations (from file or memory) to a certain location (as a base address for flags and hints). Is that correct?
It doesn't seem to work properly though.
Perhaps I can fix it when have the time. It seems pretty important for debugging.

Also, I would like to add an ability to specify an optional prefix for the flags added during `oba` bin load. To avoid confusion if there are more than one identical flags across binaries.

PR #22928, opened 2024-05-13
Big endian AVR encode fix ##endian

riptl:

**Description**


libr/arch/p/avr/assemble.c
```diff
@@ -128,8 +128,11 @@ int avr_encode(RArchSession *as, ut64 pc, const char *str, ut8 *outbuf) {
 	}
 
 	// copying result to radare struct
-	if (len > 0) {
-		memcpy (outbuf, (const ut8*)&coded, len);
+	switch (len) {
```
trufae, line 131: that looks like r_write to me, but its fine too 👍

PR #22964, opened 2024-05-23
env.sh: Don't add trailing colons to paths

blin00:

If $LD_LIBRARY_PATH is originally empty, then it becomes "$pfx/lib:" - but the trailing colon adds the current directory to the path.

This breaks horribly if you try to run r2 from a directory that also happens to have a `libc.so.6` in it :P


trufae:
Good catch :D and yeah i can imagine how dramatic is the situation when having libc in cwd :D

PR #22997, opened 2024-06-04
Fix last two covs

trufae:

**Description**


libr/core/cmd_info.inc.c
```diff
@@ -886,7 +886,7 @@ static void cmd_ic0(RCore *core, RBinObject *obj, int mode, PJ *pj, bool is_arra
 			{
 				size_t len = r_list_length (obj->classes);
 				int oldmode = mode;
```
trufae, line 888:
```suggestion
```

PR #23084, opened 2024-07-02
Fix the symbol parsing in Plan9

shurizzle:

Symbol parsing for Plan9's a.out files was failing both because the content of type `Z` (which is currently ignored) was not being discarded, and because a UT8_MAX (255 or 0xff) was being read in the PC table and treated as an error when it is actually a valid value.

### DIFF
--- a/libr/bin/p/bin_p9.c
+++ b/libr/bin/p/bin_p9.c
@@ -545,8 +545,9 @@ static RList *symbols(RBinFile *bf) {
 
 		ut64 prev = line;
 
-		const ut8 b = r_buf_read8_at (bf->buf, pcs + offset);
-		if (b == UT8_MAX) {
+		ut8 b;
+		st64 r = r_buf_read_at (bf->buf, pcs + offset, &b, sizeof (b));
+		if (r != sizeof (b)) {
 			goto error;
 		}
 		offset += sizeof (ut8);
--- a/libr/bin/p/bin_p9.c
+++ b/libr/bin/p/bin_p9.c
@@ -486,7 +486,21 @@ static RList *symbols(RBinFile *bf) {
 		case 'b':
 			break;
 		// TODO: source file line offset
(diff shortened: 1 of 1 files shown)


trufae:
Can you add a test for this? Code looks good

shurizzle:
> Can you add a test for this? Code looks good

Honestly, it's a bit difficult for me to add tests for Plan9 since none currently exist and I would have to create them from scratch. I don't have the time to do that, and I used radare2 for the first time today. Maybe if you help me, I can provide you with an a.out file that fails without the patch. I'm sorry.

trufae:
There are tests for plan9 in this file `test/db/formats/plan9`. You can uploda the binary in the testbins repository and then just copypaste one of those tests changing the `FILE` and `CMDS` fields, and then just run `r2r -i test/db/formats/plan9` to fix the output. if you dont feel comfy or don't have time don't worry much and just share the executable and i'll write the tests. but as long as you did the code and tested it you probably know better than me what's wrong and what's correct

shurizzle:
The point is that without the patch, it fails "silently," meaning it gives you the decompiled assembly but discards all the symbols. Since I'm really just starting, it would be great if I could create 2 minimal binaries (just an entry point calling `exits(nil)`) and include the two bugs. Then, you could check that it correctly exports `_main` (the entry point). If you tell me which folder to place them in, I'll do it as soon as possible. Thank you.

trufae:
i've just merged i'll wait for the binaries and i can do the tests later as i said. there are other r2/plan9 users here, so maybe others can help here. thanks for the contrib! and glad r2 worked well for your needs

shurizzle:
@trufae How can I send you the 2 `a.out` files?

trufae:
Submit a pr to this repo placing the bins in the right directory https://github.com/radareorg/radare2-testbins

thank you!

PR #23090, opened 2024-07-04
Add iP command to dsc ##io

mrmacete:

**Description**

Add the iP command to the dsc io plugin, to get **i**nfo about a **P**ointer. That draws information from the underlying rebase infos which are embedded in the pointer itself (whose semantics potentially changes over different maps), which otherwise gets lost as part of rebasing / pointer cleanup.

This is especially useful for getting info about pointer authentication, like the **diversity** and the **key** to use if there's any need to ever re-sign a pointer from the dyld cache.

The usage is `:iP [size][@vaddr]` if no size is provided defaults to 8, runs at current seek (or virtual seek if provided). The output is JSON only.

Example:

```
# this is how we see the pointer stored there normally
> pxq 8@0x1f242f210
0x1f242f210  0x00000001b972985c                       \.r.....

# and this digs the hidden metadata
> :iP @0x1f242f210~{}
[
  {
    "paddr": "0x707b7210",
    "raw": "0x8009b5883972985c",
    "format": "v3",
    "has_diversity": true,
    "diversity": 46472,
    "key": "ia"
  }
]
```

**Experimental** in the sense that it's quite naive and there's no guard against the user pointing it to an unaligned address for example, where the bits will be interpreted in the wrong way.

### DIFF
--- a/libr/io/p/io_dsc.c
+++ b/libr/io/p/io_dsc.c
@@ -308,11 +308,9 @@ static char *__system(RIO *io, RIODesc *fd, const char *command) {
 
 					ut64 raw_value = r_read_le64 (raw_value_buf);
 
-					char * tmp;
-
 					pj_o (pj);
 
-					tmp = r_str_newf ("0x%"PFMT64x, trimmed->slice->start + off_local);
+					char * tmp = r_str_newf ("0x%"PFMT64x, trimmed->slice->start + off_local);
 					pj_ks (pj, "paddr", tmp);
 					free (tmp);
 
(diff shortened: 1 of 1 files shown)


trufae:
unused vars here and there. but lgtm

Screenshot 2024-07-04 at 17 26 11

libr/io/p/io_dsc.c
```diff
@@ -254,6 +255,127 @@ static ut64 __lseek_dsc(RIO *io, RIODesc *fd, ut64 offset, int whence) {
...
+				}
+			}
+			r_list_free (infos);
+		}
+
+		r_list_free (slices);
+
+		pj_end (pj);
+
+		char * result = strdup (pj_string (pj));
```
trufae, line 371: you can use pj_drain(pj); so you can avoid the strdup and the pj_free

libr/io/p/io_dsc.c
```diff
@@ -254,6 +255,127 @@ static ut64 __lseek_dsc(RIO *io, RIODesc *fd, ut64 offset, int whence) {
...
+						break;
+					}
+
+					ut64 raw_value = r_read_le64 (raw_value_buf);
+
+					char * tmp;
+
+					pj_o (pj);
+
+					tmp = r_str_newf ("0x%"PFMT64x, trimmed->slice->start + off_local);
```
trufae, line 315:
```suggestion
					char *tmp = r_str_newf ("0x%"PFMT64x, trimmed->slice->start + off_local);
```

trufae (changes requested):

libr/io/p/io_dsc.c
```diff
@@ -254,6 +255,127 @@ static ut64 __lseek_dsc(RIO *io, RIODesc *fd, ut64 offset, int whence) {
...
+					remaining_size -= 8;
+					cursor += 8;
+					if (!got_raw_value) {
+						R_LOG_ERROR ("reading raw pointer");
+						break;
+					}
+
+					ut64 raw_value = r_read_le64 (raw_value_buf);
+
+					char * tmp;
```
trufae, line 311:
```suggestion
```

libr/io/p/io_dsc.c
```diff
@@ -254,6 +255,127 @@ static ut64 __lseek_dsc(RIO *io, RIODesc *fd, ut64 offset, int whence) {
...
+					tmp = r_str_newf ("0x%"PFMT64x, raw_value);
+					pj_ks (pj, "raw", tmp);
+					free (tmp);
+
+					tmp = r_str_newf ("v%d", trimmed_info->info->info->version);
+					pj_ks (pj, "format", tmp);
+					free (tmp);
+
+					switch (trimmed_info->info->info->version) {
+						case 1:
```
trufae, line 328: bad indentation

trufae (changes requested):
pls fix the few comments

Issue #22901, opened 2024-05-07
Avoid or filter duplicated function signature entries


radare:
Way to reproduce:

```
$ r2 /bin/ls
[0x100003a84]> aaa
[0x100003a84]> zg
[0x100003a84]> zg
[0x100003a84]> zg
[0x100003a84]> z* $$~N
za main_3 N sym.func.1000047d0_3
za main_2 N sym.func.1000047d0_2
za main N sym.func.1000047d0
[0x100003a84]>
```

trufae:
Fixed now

PR #23098, opened 2024-07-08
Remove duplicated function signature entries

satk0:

**Description**

Probably solves the issue #22901. I removed the code that creates new signatures from the same functions.

### DIFF
--- a/libr/anal/sign.c
+++ b/libr/anal/sign.c
@@ -958,14 +958,8 @@ R_API int r_sign_all_functions(RAnal *a, bool merge) {
 		RSignItem *it = NULL;
 		if (merge || !name_exists (a->sdb_zigns, realname, sp)) {
 			it = item_from_func (a, fcn, realname);
-		} else {
-			char *name = get_unique_name (a->sdb_zigns, fcn->name, sp);
-			if (name) {
-				it = item_from_func (a, fcn, name);
-			}
-			free (name);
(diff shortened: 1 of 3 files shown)


libr/anal/sign.c
```diff
@@ -958,14 +958,8 @@ R_API int r_sign_all_functions(RAnal *a, bool merge) {
 		RSignItem *it = NULL;
 		if (merge || !name_exists (a->sdb_zigns, realname, sp)) {
 			it = item_from_func (a, fcn, realname);
-		} else {
```
trufae, line 961: the fix looks good and i think thats a good default behaviour, but maybe we should allow users to have the same signature twice too. maybe we can have the `zign.dups` config variable for this, and maybe we need to break the abi to add such option inside RAnalOptions. but i think is better to have this feature as an option and disabled by default rather than removing the code. what do you think?

trufae:
You can use the RCoreBind field from RAnal.coreb for now to achieve this, put the ConfigGet call outside the loop to avoid perf issues. add a comment for R2_600 to add the field in RAnalOptions this way we dont break the abi.

the option will be disabled by default . add  it in core/cconfig.c

satk0:
Allright, I will try my best to implement it. Thanks for all the tips 😁

trufae:
Looks perfect now :) thank you!

PR #23115, opened 2024-07-10
Fix Objective-C classes Swift demangling ##bin

mrmacete:

**Description**

Use `r_bin_demangle_swift` to demangle Swift Objective-C classes in the couple of places where that was missing, and ensure all parsed classes have `index` set, to allow coexistence of different categories with the same name.

### DIFF
--- a/libr/bin/format/objc/mach0_classes.c
+++ b/libr/bin/format/objc/mach0_classes.c
@@ -1101,39 +1101,10 @@ static inline const char *skipnum(const char *s) {
 	return s;
 }
 
-// TODO: split up between module + classname
-static char *demangle_classname(const char *s) {
-	const char *kstr;
-	char *ret, *klass;
+static char *demangle_classname(RBin *rbin, const char *s) {
+	char *ret;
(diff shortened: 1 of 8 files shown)


trufae:
test output should be updated. a bit more verbose, but if that fixes other usecases better sync up

Screenshot 2024-07-10 at 21 50 36

mrmacete:
yeah not sure of all this sudden `Swift.` prefixing

mrmacete:
apart from that yeah it fixes a lot of stuff, but we need better test coverage

trufae:
please rebase

mrmacete:
rebased, but also observed that the `Swift.` prefix is an artifact of r2's own demangler, because it goes away on macos by enabling `-e bin.demangle.trylib=true`

mrmacete:
i can work around it in mach0_classes or we can fix the root cause, what do you think?

trufae:
ok to merge?

mrmacete:
yeah should be "good enough" now

trufae:
merged!

PR #23148, opened 2024-07-22
Avoid over-unsetting fixed arch / bits ##core

mrmacete:

**Description**

They're already set back to the original value on the `is_arch_set` / `is_bits_set` branches above, removing the second restoration allows for actually setting them via the `anal.fixed.arch` / `anal.fixed.bits` configs.


trufae:
Looks good

PR #23167, opened 2024-08-04
Fix bug in V;! not bringing back the mouse ##visual

trufae:

**Description**


libr/cons/cons.c
```diff
@@ -657,28 +664,13 @@ R_API bool r_cons_enable_mouse(const bool enable) {
 		bool enabled = I->mouse;
```
trufae, line 664:
```suggestion
```

PR #23190, opened 2024-08-08
Improve addr wrap support in r_itv

condret:
Not sure if this is a good idea

**Description**


trufae:
let's merge it, we can have a test later, but that makes the behaviour more consistent after the chat we had

PR #23191, opened 2024-08-09
Add r2book info files to deb builds

prodrigestivill:
Add latest r2book info file to deb builds


prodrigestivill:
Currently rpm builds download an old version from source, it might need to change the rpm build system before we can add this there.

prodrigestivill:
Not sure what is prefered:

```console
$ info r2book
$ info radare2
```

As is currently implemented in this PR, only will work `info r2book` but I can change it to the other or both (symlink).

trufae:
R2book lgtm

PR #23192, opened 2024-08-10
r2r: Display the correct number of jobs

sylvainpelissier:

**Description**


trufae:
I think thats better than before but shows only the jobs of the type of tests only so the counter doesnt match the full set. Still, the output looks better than before 👍

PR #23198, opened 2024-08-15
Move CRC search to magic

sylvainpelissier:

**Description**
Move the CRC32 search from `/ck` to `/m` since it is not really Crypto material. In addition, the big/little endian selector does not really applies since a binary can contain both.


trufae:
So i like to have it in /m but lrobably i expect it to be in /ck too wdyt?

sylvainpelissier:
> /ck is faster than /m

I wonder why `/m` is so slow even when using a single pattern file.

Yes sure we can keep them both specially right now that `/m` is slow.

sylvainpelissier:
I've revert the `\ck` modifications.

libr/core/cmd_search.inc.c
```diff
@@ -4338,8 +4338,8 @@ static int cmd_search(void *data, const char *input) {
 
 				// crc32
 				kw = (le)
-					? r_search_keyword_new_hexmask ("00000000963007772c610eeeba51099919c46d078ff46a7035a563e9a395649e3288db0ea4b8dc791ee9d5e088d9d2972b4cb609bd7cb17e072db8e7911dbf90", NULL)
-					: r_search_keyword_new_hexmask ("0000000077073096ee0e612c990951ba076dc419706af48fe963a5359e6495a30edb883279dcb8a4e0d5e91e97d2d98809b64c2b7eb17cbde7b82d0790bf1d91", NULL);
+						? r_search_keyword_new_hexmask ("00000000963007772c610eeeba51099919c46d078ff46a7035a563e9a395649e3288db0ea4b8dc791ee9d5e088d9d2972b4cb609bd7cb17e072db8e7911dbf90", NULL)
```
trufae, line 4341:
```suggestion
					? r_search_keyword_new_hexmask ("00000000963007772c610eeeba51099919c46d078ff46a7035a563e9a395649e3288db0ea4b8dc791ee9d5e088d9d2972b4cb609bd7cb17e072db8e7911dbf90", NULL)
```

libr/core/cmd_search.inc.c
```diff
@@ -4338,8 +4338,8 @@ static int cmd_search(void *data, const char *input) {
...
+						: r_search_keyword_new_hexmask ("0000000077073096ee0e612c990951ba076dc419706af48fe963a5359e6495a30edb883279dcb8a4e0d5e91e97d2d98809b64c2b7eb17cbde7b82d0790bf1d91", NULL);
```
trufae, line 4342:
```suggestion
					: r_search_keyword_new_hexmask ("0000000077073096ee0e612c990951ba076dc419706af48fe963a5359e6495a30edb883279dcb8a4e0d5e91e97d2d98809b64c2b7eb17cbde7b82d0790bf1d91", NULL);
```

PR #23200, opened 2024-08-17
horisontal swipe in termux

john-peterson:
normal panning works now in termux. the graph follows the finger 

```
binr/radare2/radare2 -c "aa;s main;VV" binr/rabin2/rabin2
```

this branch emit the horizontal codes

https://github.com/john-peterson/termux-app/tree/scroll


trufae:
Are those ansi codes standard? Will termux end up supporting this by default?

PR #23250, opened 2024-08-30
Minor optimization in r_graph ##util

condret:
maybe abi breaking?

**Description**

### DIFF
--- a/libr/include/r_util/r_graph.h
+++ b/libr/include/r_util/r_graph.h
@@ -8,7 +8,7 @@ extern "C" {
 #endif
 
 typedef struct r_graph_node_t {
-	unsigned int idx;
+	ut32 idx;
 	void *data;
 	RList *out_nodes;
 	RList *in_nodes;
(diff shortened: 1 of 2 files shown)


libr/util/graph.c
```diff
@@ -12,7 +12,8 @@ static RGraphNode *r_graph_node_new(void *data) {
 	RGraphNode *p = R_NEW0 (RGraphNode);
 	if (p) {
 		p->data = data;
-		p->free = NULL;
+// R_NEW0 already sets p->free to NULL
```
trufae, line 15:
```suggestion
```

libr/util/graph.c
```diff
@@ -12,7 +12,8 @@ static RGraphNode *r_graph_node_new(void *data) {
...
+//		p->free = NULL;
```
trufae, line 16:
```suggestion
```

libr/include/r_util/r_graph.h
```diff
@@ -19,13 +19,13 @@ typedef struct r_graph_node_t {
...
 	RGraphNode *to;
-	int nth;
+	st32 nth;
 } RGraphEdge;
 
 typedef struct r_graph_t {
-	unsigned int n_nodes;
-	unsigned int n_edges;
-	int last_index;
+	ut32 n_nodes;
```
trufae, line 26: What about using size_t instead so we ensure the struct fields are properly aligned in memory without gaps? This change will break abi but we can have the new reordered fields under the newabi ifblock

libr/util/graph.c
condret: comment is so that nobody who reads this code reports a bug, where there is none

libr/include/r_util/r_graph.h
condret: n_nodes is more or less a copy of r_list_length (g->nodes), which is ut32. i see no need to switch to size_t. i'm not sure if size_t has guaranteed minimum size

PR #23257, opened 2024-08-31
print string until first non-printable character

sylvainpelissier:

**Description**

`ps` command print only the string until the first non-printable character. Not yet implemented when the charset is changed for example with `e cfg.charset = ascii`.

### DIFF
--- a/test/db/cmd/charset
+++ b/test/db/cmd/charset
@@ -31,7 +31,7 @@ ARGS=-n
 CMDS=<<EOF
 s 0x19a8
 px 16
-ps
+psx 10
 e cfg.charset=ebcdic37
 px 16
 ps 10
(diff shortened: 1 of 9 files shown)


test/db/cmd/charset
```diff
@@ -39,7 +39,7 @@ EOF
 EXPECT=<<EOF
 - offset -  A8A9 AAAB ACAD AEAF B0B1 B2B3 B4B5 B6B7  89ABCDEF01234567
 0x000019a8  c896 9381 4094 a495 8496 1500 0000 0000  ....@...........
-\xc8\x96\x93\x81@\x94\xa4\x95\x84\x96\x15
+
 - offset -  A8A9 AAAB ACAD AEAF B0B1 B2B3 B4B5 B6B7  89ABCDEF01234567
```
trufae, line 43: Use psx or maybe it should be pszx :? So we can keep those escaped bytes i. The test

sylvainpelissier: Yes, done.

PR #23271, opened 2024-09-02
WIP: Make arm plugin use capstone 6 ##arch

satk0:

**Description**

### DIFF
--- a/libr/arch/p/bpf_cs/plugin.c
+++ b/libr/arch/p/bpf_cs/plugin.c
@@ -4,7 +4,6 @@
 #include <r_esil.h>
 #include <r_lib.h>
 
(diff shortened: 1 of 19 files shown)


libr/arch/p/bpf_cs/plugin.c
```diff
@@ -4,6 +4,7 @@
 #include <r_esil.h>
 #include <r_lib.h>
 
+#define CAPSTONE_AARCH64_COMPAT_HEADER
```
trufae, line 7: is this necessary for bpf? same for the other archs

satk0: Will try to delete it whenever possible

PR #23295, opened 2024-09-06
Add ESIL translation for `movn` instruction (fixes #23286) ##emu

### LINKED ISSUE #23286: No ESIL for `movn` on MIPS
## Environment

```sh
Thu Sep  5 11:39:40 PM CEST 2024
radare2 5.9.2 0 @ linux-x86-64
birth: git.5.9.2 2024-08-01__08:35:56
options: gpl release -O1 cs:5 cl:2 meson
Linux x86_64
```

## Description

No ESIL is displayed for conditional move instructions on MIPS, i.e., they are treated like NOPs.

```
[0x80053bb8]> pd
        :   0x80053bb8      0b304300       movn a2, v0, v1
[...]
[0x80053bb8]> e asm.esil = true
[0x80053bb8]> pd
        :   0x80053bb8      0b304300       ,
[...]
```

The semantics of this particular insn are:

```
movn d,s,t if (t) d = s;
```

I'd expect to either see a notification that is insn is unimplemented or an ESIL expression that is expressing the correct semantics.

Wouldn't mind fixing this myself (I expect it to be rather small fix) but I need some help where to start.

## Test

I guess creating a file with the bytes and basing it at zero should be enough:

```
$ echo -n '\x0b\x30\x83\x00' > /tmp/bytes
$ r2 -e asm.esil=true -c 'pd 1' -q -a mips -b 32 -e cfg.bigendian=false -m 0x0 /tmp/bytes
```

vobst:

**Description**

Add missing translation for MIPS `movn` instruction.
Its a conditional move instruction with semantics

            movn d,s,t      if (t) d = s;

Which is quite similar to the existing:

            movt d,s,$fccN  if (fcc(N)) d = s;

In fact, we can reuse the exact same implementation.

PS: Would be nice if someone could explain to me what the `$z` does and why it is there.

### DIFF
--- a/libr/arch/p/mips/plugin_cs.c
+++ b/libr/arch/p/mips/plugin_cs.c
@@ -456,6 +456,7 @@ static int analop_esil(RArchSession *as, RAnalOp *op, csh *handle, cs_insn *insn
 					ARG (2), ARG (1), REG (0));
 			}
 			break;
+		case MIPS_INS_MOVN:
 		case MIPS_INS_MOVT:
 			PROTECT_ZERO () {
 				r_strbuf_appendf (&op->esil, "1,%s,==,$z,?{,%s,%s,=,}",


trufae:
Looks good :) i was about to say that the implementation is not quite the same but the way its done can be reused. Congrats for your first contrib

PR #23359, opened 2024-09-19
Fix /cp segfault on missing key

0x8008135:

When using /cp if pubkey parameter was not supplied it resulted in a segfault.

`
[0x00000000]> /cp ed25519 
WARN: r_str_trim_head_ro: assertion 'str' failed (line 110)
[1]    956100 segmentation fault (core dumped)  r2 file.bin
`

This patch displays back the help instead of crashing r2.

### DIFF
--- a/libr/core/cmd_search.inc.c
+++ b/libr/core/cmd_search.inc.c
@@ -812,7 +812,7 @@ R_API RList *r_core_get_boundaries_prot(RCore *core, R_UNUSED int perm, const ch
 	snprintf (bound_from, sizeof (bound_from), "%s.%s", prefix, "from");
 	snprintf (bound_to, sizeof (bound_to), "%s.%s", prefix, "to");
 	const ut64 search_from = r_config_get_i (core->config, bound_from),
-	      search_to = r_config_get_i (core->config, bound_to);
+		  search_to = r_config_get_i (core->config, bound_to);
 	const RInterval search_itv = {search_from, search_to - search_from};
 	if (!mode) {
 		mode = r_config_get (core->config, bound_in);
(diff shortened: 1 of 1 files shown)


libr/core/cmd_search.inc.c
```diff
@@ -812,7 +812,7 @@ R_API RList *r_core_get_boundaries_prot(RCore *core, R_UNUSED int perm, const ch
 	snprintf (bound_from, sizeof (bound_from), "%s.%s", prefix, "from");
 	snprintf (bound_to, sizeof (bound_to), "%s.%s", prefix, "to");
 	const ut64 search_from = r_config_get_i (core->config, bound_from),
-	      search_to = r_config_get_i (core->config, bound_to);
+		  search_to = r_config_get_i (core->config, bound_to);
```
trufae, line 815: split in two statements, dont do two assignemnts in one line

libr/core/cmd_search.inc.c
```diff
@@ -1286,7 +1286,7 @@ static RList *construct_rop_gadget(RCore *core, ut64 addr, ut8 *buf, int buflen,
 			r_anal_op_fini (&asmop);
 		}
 		if (!r_str_ncasecmp (opst, "invalid", strlen ("invalid")) ||
-		    !r_str_ncasecmp (opst, ".byte", strlen (".byte"))) {
+			!r_str_ncasecmp (opst, ".byte", strlen (".byte"))) {
```
trufae, line 1289: maybe i should add a case insensitive version of the r_str_startswith().. note for my future self

libr/core/cmd_search.inc.c
trufae: indent ONLY with tabs.

trufae (changes requested):
.

libr/core/cmd_search.inc.c
0x8008135: Probably a leftover from someone's commit, fixed by https://github.com/radareorg/radare2/pull/23359/commits/a32873816503db175c5ef9284f9b15d4e2263331

PR #23360, opened 2024-09-19
Throw error when python>3.11 in preconfigure.bat ##build

satk0:

**Description**


trufae:
I would like to understand better the problematic with python 3.12 before blindmerging this change

trufae:
`pip install setuptools` should fix that

satk0:
Dont merge it, batch is awful, it still is not working, will tell u when

satk0:
Ok, now it's finally working

Issue #23371, opened 2024-09-22
rahash2 -B behavior


jukuisma:
## Environment

```sh
Sun Sep 22 11:26:30 UTC 2024
radare2 5.9.5 32609 @ linux-x86-64
birth: git.5.9.4-211-g909aa376f8 2024-09-22__11:12:16
commit: 909aa376f88832dc8c93e77ddc0d135e46044930
options: gpl -O? cs:5 cl:2 make
Linux x86_64
```

## Description

`rahash2 -B` always returns an extra block. It reports the same final byte range twice, but two different hashes. The final hash is the md5sum of an empty file.
 
## Test

```sh
╭─vagrant@bookworm ~/github/radare2  ‹master› 
╰─➤  head -c 4096 /dev/urandom | rahash2 -Bb 1K -a md5 -
0x00000000-0x000003ff md5: 2c0582a026fc0f01120009f9d4c29e34
0x00000400-0x000007ff md5: 6a28c606b117a44475b97dbefb69a664
0x00000800-0x00000bff md5: 7380159fdc66c6a6baa635e49c5ee58a
0x00000c00-0x00000fff md5: 094a33b0afbcbac7670d5655ffa75024
0x00000c00-0x00000fff md5: d41d8cd98f00b204e9800998ecf8427e
```
```sh
╭─vagrant@bookworm ~/github/radare2  ‹master› 
╰─➤  rahash2 -Bb 1K -a md5 /bin/ls | tail -2
0x00024c00-0x00024f2f md5: ccf624397ab1400b10ce704fd65f33ee
0x00024c00-0x00024f2f md5: d41d8cd98f00b204e9800998ecf8427e
```
```sh
╭─vagrant@bookworm ~/github/radare2  ‹master› 
╰─➤  echo -n '' | md5sum
d41d8cd98f00b204e9800998ecf8427e  -
```

jukuisma:
Other algos seem to run into the same problem, but trigger an assertion as well:
```sh
╭─vagrant@bookworm ~  
╰─➤  head -c 4096 /dev/urandom | rahash2 -Bb 1K -a sha256 -
0x00000000-0x000003ff sha256: f40f250f58ac7181a904d9dd7e28ca302416e8978e5a5fcd81189e7a25ada103
0x00000400-0x000007ff sha256: 0791d6e8c3e36ceee04e6607b3a083bc155dc5accc9303abbe27b1b60eda5617
[...]

PR #23373, opened 2024-09-22
rahash2 -B behavior

jukuisma:

**Description**

Fixes: https://github.com/radareorg/radare2/issues/23371


trufae:
Why in draft?

jukuisma:
> Why in draft?

Because I didn't have time to thoroughly test the changes yesterday. Granted, the changes seem kind of trivial and the tests pass. Also, I did leave one open question, but it doesn't seem to be showing to anyone else than me:
image
I'm more familiar with gitlab instead of github :sweat_smile:

trufae:
iirc passing null as the buffer is used to initialize the selected hashing plugin. i would keep it commented, because it will be easier to spot later if anything else fails, but iimho if tests pass is the way to go. and i would like to refactor the hash api during the 6.0 transition so this commented line may help us find that. also, adding more tests will help us identify other issues when refactoring

jukuisma:
> iirc passing null as the buffer is used to initialize the selected hashing plugin. i would keep it commented, because it will be easier to spot later if anything else fails, but iimho if tests pass is the way to go. and i would like to refactor the hash api during the 6.0 transition so this commented line may help us find that. also, adding more tests will help us identify other issues when refactoring

I can add that line commented out and some more test cases later tonight. Maybe around 17:00 UTC.

trufae:
the world can wait ;) feel free to add more tests in case you find them useful/related to the issue. thank you!

jukuisma:
Added the removed line back, but left it commented out to fix this bug. Didn't add any more test cases. Rahash2 doesn't seem to support calculating checksums of empty files and while we could copy paste similar new test cases for other algorithms, it seems a bit redundant.

Ready to merge on my behalf.

Issue #22445, opened 2023-11-25
Store plugin path and avoid dupes


trufae:
there should be a warning when trying to open a secondary plugin with the same name+type. also Lcj should display the absolute path where each plugin is loaded from

<img width="394" alt="Screenshot 2023-11-25 at 20 54 28" src="https://github.com/radareorg/radare2/assets/6431515/90c5744c-2d22-4142-a020-398bd7fa82b9">

trufae:
or maybe checksum the plugin file which can be probably a better approach, because r2 could be reporting better error messages when finding duplicated plugin installations

trufae:
let's move this to the next abi breaking release, not really a big issue if users keep their plugin folders clean. and we may probably want to find a better balance between startup performance and so on

PR #23386, opened 2024-09-25
Store plugin's by name rather than path

MewtR:

**Description**

So I was looking at #22445 and I figured it would be interesting to store plugins in the hash table by their names instead of their paths.
It makes it easier to detect duplicates and easier to get the path from which the plugin was loaded, but this is more of a proposition.
Code is a little awkward in some places, looking for feedback.

### DIFF
--- a/libr/arch/p/avr/pseudo.c
+++ b/libr/arch/p/avr/pseudo.c
@@ -212,6 +212,8 @@ RParsePlugin r_parse_plugin_avr_pseudo = {
 	.meta = {
 		.name = "avr.pseudo",
 		.desc = "AVR pseudo syntax",
+		.author = "pancake",
+		.license = "LGPL-3.0-only",
 	},
 	.parse = parse
 };
--- a/libr/arch/p/6502/pseudo.c
+++ b/libr/arch/p/6502/pseudo.c
@@ -180,6 +180,8 @@ RParsePlugin r_parse_plugin_6502_pseudo = {
 	.meta = {
 		.name = "6502.pseudo",
 		.desc = "6502 pseudo syntax",
+		.author = "pancake",
+		.license = "LGPL-3.0-only",
 	},
 	.parse = parse,
 };
--- a/libr/arch/p/arm/pseudo.c
+++ b/libr/arch/p/arm/pseudo.c
@@ -487,6 +487,8 @@ RParsePlugin r_parse_plugin_arm_pseudo = {
 	.meta = {
 		.name = "arm.pseudo",
 		.desc = "ARM/ARM64 pseudo syntax",
+		.author = "pancake",
+		.license = "LGPL-3.0-only",
 	},
 	.parse = parse,
 	.subvar = &subvar,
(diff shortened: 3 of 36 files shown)


trufae:
This change breaks the abi. Can u use the R2_USE_NEW_ABI ?

MewtR:
Alright, wasn't aware of that define, I'll look into it in the coming days.

trufae:
As long as this is a rather large PR changing the abi of rparse i dont think it makes sense to use the new-abi define. so i think it's better to hold this PR until 5.9.9 is in dev. (we need to release 5.9.6 and 5.9.8 before reaching the abi breaking seasson).

Apart from that i think the change makes sense and seems reasonable.

I want to perform more changes in the rparse api, so i wanted to have some time before the 5.9.9 time arrives to do that. what do you think? as long as we are not gonna break abi, this pr shouldnt be affected by any change until the time arrives.

also, do u think the getname callback is necessary? because if all the plugins use the rpluginmeta now, we can just assume this struct on all of them and not depend on helper functions to do that.

MewtR:
You're right about the getname callback, I removed it.

I'll try to write some tests for this at some point.
Also, yeah we can wait until abi-breaking season I don't mind.

I'm afraid this change might break some plugins that are outside the r2 repo though. Are you aware of any bp/parse plugins "external" to r2 or how I can find them?

trufae:
Theres inly afen. An external parse plugin. Dont know any about bp

libr/arch/p/avr/pseudo.c
```diff
@@ -209,8 +209,12 @@ static int parse(RParse *p, const char *data, char *str) {
 }
 
 RParsePlugin r_parse_plugin_avr_pseudo = {
-	.name = "avr.pseudo",
-	.desc = "AVR pseudo syntax",
+	.meta = {
+		.name = "avr.pseudo",
+		.desc = "AVR pseudo syntax",
+		.author = "pancake",
+		.license = "LGPL3",
```
trufae, line 216: Please use SPDX namings. See the related commits in git log | grep SPDX for more info

trufae:
Now it’s the abi breaking moment! Feel free to help improve the state or other apis too! Glad to see this pr updated now btw

MewtR:
I think the PR is ready. I wanted to write some tests but I need a "dummy" plugin  and I'm not sure where to put them. I was initially thinking of putting it in [radare2-testbins](https://github.com/radareorg/radare2-testbins) but it might get complicated to support all platforms. Do you have any suggestions?

trufae:
Looks good to merge for me too. No need for tests imho. We have many other important refactorings to do rn and we probably want to mess a little more with the plugins architecture before we release. Would u like to split this up into three commits for bp/parse meta and another for the speicific changes to make this logic work? Otherwise in fine to merge it as 1 commit too

MewtR:
Alright, I split it into three commits. That's what you meant right? 3 commits and not 3 PRs? I can also have 3 PRs if you like

trufae:
Maybe worth using "Fix #22445 - Store plugin's by name rather than path ##core"

this way the commit will be listed in the release notes, and the ticket will be closed when merged

trufae:
one 1 with 3 commits is fine, what i care mainly is about using the ## for the relevant commits to appear in the release notes :) so just update the last commit and ill merge

MewtR:
Ah yes ok, it's done!

PR #23387, opened 2024-09-26
Refactor listing crypto algorithms into a single function ##crypto

sylvainpelissier:

**Description**

Remove the `algolist` function of `ph` command and refactor the `r_crypto_list` function to list algorithms by types.


trufae:
This change breaks the abi. Use the R2_USE_NEW_ABI define meanwhile. All that code will be swapped during the .9 release

libr/core/cmd_print.inc.c
```diff
@@ -3634,15 +3606,15 @@ static bool cmd_print_ph(RCore *core, const char *input) {
...
+		r_crypto_free (cry);
 		return true;
 	}
 	if (i0 == 'j') {
-		algolist (core, 'j');
-		return true;
-	}
-	if (i0 == '=') {
-		algolist (core, 0);
+		RCrypto *cry = r_crypto_new ();
```
trufae, line 3615: i think we should have a crypto instance inside RCore, and then use get_session() to associate the requested plugins to do the cyrpto operations over it.

libr/crypto/crypto.c
```diff
@@ -194,7 +194,7 @@ static inline void print_plugin_verbose(RCryptoPlugin *cp, PrintfCallback cb_pri
 	cb_printf ("%c %12s %5s %s %s\n", type, cp->meta.name, license, desc, author);
 }
 
-R_API void r_crypto_list(RCrypto *cry, R_NULLABLE PrintfCallback cb_printf, int mode) {
+R_API void r_crypto_list(RCrypto *cry, R_NULLABLE PrintfCallback cb_printf, int mode, RCryptoType type) {
```
trufae, line 197: to keep abi stable you can combine mode and type by shifting the value and unshifting it here

trufae (changes requested):
make this crypto list api abi safe by combining the args and wrap the calls with new-abi, this way we can properly unwrap it and merge it now

trufae:
The change looks good. but we need to keep abi stable until the 5.9.9 time arrives. maybe i should release 5.9.6 soon, so we must wait a little. which is sad, because i wanted to do this change for 5.9.0 but i couldnt wait more to do the release. in other words. im all for merging this but we need to keep the abi stable promise.

the crypto instance in core cant be done because that would be a major abi break but can be added for later in the new abi block and we can discuss the refactoring of the crypto api to work like arch does

libr/crypto/crypto.c
sylvainpelissier: Ok I implemented that, it does not break the ABI anymore and I would change it for 5.6.9.

libr/core/cmd_print.inc.c
```diff
@@ -2,6 +2,8 @@
 
 #if R_INCLUDE_BEGIN
 
+#include <r_crypto.h>
```
trufae, line 5: I dont think this is inecessary from a .inc file

PR #23398, opened 2024-09-27
Disable bin.relocs.apply when running avrr ##bin

trufae:

**Description**


trufae:
Relocs are necessary for parsing rtti ... this issue is probably because relocs are wrongly applied or messing up with an obfuscated binary..

PR #23418, opened 2024-10-01
Fix wb single byte write behavior

jukuisma:

**Description**

Fix a bug in single byte wb commands and add some new test cases that catch the old bug. Related to: https://github.com/radareorg/radare2/issues/23405, but doesn't implement the multibyte wb commands yet.


jukuisma:
TODO:
- [ ] Update `(neo)?vim` configs to use tabs instead of spaces in C
- [ ] Amend commit messages and write PR description
- [ ] Use bitwise or instead of `+=`, which should also work now that the shifts have been fixed

trufae:
Good catch and fix, TDD all the way!

PR #23421, opened 2024-10-02
fix for xrefs permission indicator: -w- r-- etc.

sha0coder:

**Description**


trufae (commented):
try this change. because op.direction can be -1, 0 or 1

PR #23499, opened 2024-10-18
Increase float and double precision ##core

rolio37:

**Description**

It would be interesting with the ? option to increase the precision of double and float to have more precise values.

Test program to solve:
https://github.com/rolio37/re/blob/main/crackme_xmm.c


trufae:
fix the tests

Issue #23490, opened 2024-10-17
Ensure nullable for all the apis


trufae:
example https://github.com/radareorg/radare2/pull/23487

PR #23518, opened 2024-10-23
Check vector API nullability for  #23490

astralia:

**Description**


astralia:
NOTE: This API should be deprecated and use instead r_vec

PR #23520, opened 2024-10-23
PE section flags

sha0coder:
Ideally the section->perm item is not necesary because can be infered from section->flags
both come from PE section Characteristics but perm is transformed in the way, we keep the orignal flags on flags item.

### DIFF
--- a/libr/core/cbin.c
+++ b/libr/core/cbin.c
@@ -3228,7 +3228,7 @@ static bool bin_sections(RCore *r, PJ *pj, int mode, ut64 laddr, int va, ut64 at
 				pj_ks (pj, "type", section->type);
 			}
 			pj_ks (pj, "perm", perms);
-            pj_kN (pj, "flags", section->flags);
+			pj_kN (pj, "flags", section->flags);
 			if (hashtypes && (int)section->size > 0) {
 				int datalen = section->size;
 				if (datalen > 0 && datalen < plimit) {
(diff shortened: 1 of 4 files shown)


libr/include/r_bin.h
```diff
@@ -303,6 +303,7 @@ typedef struct r_bin_section_t {
 	ut64 vaddr;
 	ut64 paddr;
 	ut32 perm;
+	ut32 flags;
```
trufae, line 306: This change breaks the ABI so we can't merge this until the 5.9.8 release is out. we can have R2_USE_NEW_ABI ifdef if you want.. but personall i would wait because it shuoldnt be taking long and we can probably introduce this new field to work with other fileformats

trufae: for ELF https://www.sco.com/developers/gabi/2003-12-17/ch4.sheader.html

trufae (changes requested):
maybe good to keep original output in tests

trufae:
can you rebase this pr and finish it? we are in the abi breaking season now :)

libr/core/cbin.c
```diff
@@ -3047,11 +3047,11 @@ static bool bin_sections(RCore *r, PJ *pj, int mode, ut64 laddr, int va, ut64 at
 	}
 	if (IS_MODE_NORMAL (mode)) {
 		if (hashtypes) {
-			r_table_set_columnsf (table, "dXxXxssss",
-				"nth", "paddr", "size", "vaddr", "vsize", "perm", hashtypes, "type", "name");
+			r_table_set_columnsf (table, "dXxXxssssx",
+				"nth", "paddr", "size", "vaddr", "vsize", "perm", hashtypes, "type", "name", "flags");
```
trufae, line 3051: Put flags before or after perm. But not after name

trufae (changes requested):
Then run r2r -i .. with the testfile that fails to fix them

trufae:
fixed in https://github.com/radareorg/radare2/pull/23699

PR #23558, opened 2024-10-29
Fix #23557 - Ignore types with DW_AT_declaration in DWARF

### LINKED ISSUE #23557: sdb types are overwritten from other compilation units with same name
## Environment

```sh
ter 29 out 2024 11:29:20 -03
radare2 5.9.4 0 @ linux-x86-64
birth: git.5.9.4 2024-09-28__12:07:37
options: gpl release -O1 cs:5 cl:2 meson
Linux x86_64
```

## Description

DWARF type structures that are incomplete (DW_AT_declaration) are inserted in sdb_types overwriting the concrete type that was previously added, making radare2 lose track of all struct members. There's another related issue here that even if the type is concrete but not exported in a header file, it will overwrite one from another compilation unit with the same name, because types doesn't distinguish from compilation unit atm.

## Test

https://github.com/radareorg/radare2-testbins/pull/98

be aware that the order of the arguments in gcc makes DW_AT_declaration appear first then it won't reproduce

wgrr:

### DIFF
--- a/libr/anal/type.c
+++ b/libr/anal/type.c
@@ -385,7 +385,13 @@ static void save_struct(const RAnal *anal, const RAnalBaseType *type) {
 	}
 	// struct.name=param1,param2,paramN
 	char *key = r_str_newf ("%s.%s", kind, sname);
-	sdb_set_owned (anal->sdb_types, key, r_strbuf_drain (arglist), 0);
+	if (sdb_exists (anal->sdb_types, key)) {
+		R_LOG_WARN ("Ignoring overwrite of type '%s' in sdb_types", key);
+		r_strbuf_free (arglist);
+	} else {
+		sdb_set_owned (anal->sdb_types, key, r_strbuf_drain (arglist), 0);
+	}
+
 	free (key);
 	free (sname);
 }
--- a/libr/anal/dwarf_process.c
+++ b/libr/anal/dwarf_process.c
@@ -516,6 +516,10 @@ cleanup:
 static void parse_structure_type(Context *ctx, ut64 idx) {
 	const RBinDwarfDie *die = &ctx->all_dies[idx];
 
+	if (find_attr_idx (die, DW_AT_declaration) != -1) {
+		return;
+	}
+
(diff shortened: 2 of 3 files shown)


libr/anal/type.c
```diff
@@ -385,7 +385,13 @@ static void save_struct(const RAnal *anal, const RAnalBaseType *type) {
 	}
 	// struct.name=param1,param2,paramN
 	char *key = r_str_newf ("%s.%s", kind, sname);
-	sdb_set_owned (anal->sdb_types, key, r_strbuf_drain (arglist), 0);
+	if (sdb_exists (anal->sdb_types, key)) {
+		R_LOG_WARN ("Ignoring overwrite of type \"%s\" in sdb_types.", key);
```
trufae, line 389:
```suggestion
		R_LOG_WARN ("Ignoring overwrite of type \"%s\" in sdb_types", key);
```

trufae (changes requested):
^

wgrr:
I've just noticed that I forgot to update the file name in cmd_tc test after I changed it to use underscore like other execs in bins, done

libr/anal/type.c
```diff
@@ -385,7 +385,13 @@ static void save_struct(const RAnal *anal, const RAnalBaseType *type) {
 	}
 	// struct.name=param1,param2,paramN
 	char *key = r_str_newf ("%s.%s", kind, sname);
-	sdb_set_owned (anal->sdb_types, key, r_strbuf_drain (arglist), 0);
+	if (sdb_exists (anal->sdb_types, key)) {
+		R_LOG_WARN ("Ignoring overwrite of type '%s' in sdb_types.", key);
```
trufae, line 389:
```suggestion
		R_LOG_WARN ("Ignoring overwrite of type '%s' in sdb_types", key);
```

trufae: remove the dot please. otherwise the linter will fail

trufae:
perfect 👌

PR #23583, opened 2024-11-01
Correct print strhash

sylvainpelissier:

**Description**

Correct strhash plugin according to http://www.cse.yorku.ca/~oz/hash.html and start using plugins in `ph` command.


libr/util/str.c
```diff
@@ -248,14 +248,14 @@ R_API R_MUSTUSE char *r_str_r2_prefix(const char *str) {
...
-// Compute a 64 bit DJB hash of a string.
+// Compute a 64 bit DJB hash of a string (See the XOR version in http://www.cse.yorku.ca/~oz/hash.html)
 R_API ut64 r_str_hash64(const char *s) {
 	ut64 len, h = 5381;
 	if (!s) {
 		return 0;
 	}
 	for (len = strlen (s); len > 0; len--) {
-		h = (h ^ (h << 5)) ^ *s++;
+		h = (h + (h << 5)) ^ *s++;
```
trufae, line 258: this comment in sdb explains the reason for the change. 

Screenshot 2024-11-01 at 11 52 20 PM

sylvainpelissier: Ok I corrected it and add a comment to explain.

trufae: i dont thikn the algo is the same,.. but in my tests it was behaving faster and having the same amount of collisions. but im not a mathematician so dunno

sylvainpelissier: Yes it changed with the carry propagation. But we can keep it like that.

Issue #23579, opened 2024-10-31
dwarf5 wrong file names


wgrr:
## Environment

```sh
# copypaste this script into your shell and replace it with the output
qui 31 out 2024 17:17:01 -03
radare2 5.9.4 0 @ linux-x86-64
birth: git.5.9.4 2024-09-28__12:07:37
options: gpl release -O1 cs:5 cl:2 meson
Linux x86_64
```

## Description

dwarf5 file name for the addr2line are wrong

```sh
% r2 a.out
[0x00001040]> CL~file
file: /usr/include/stdio.h
file: /usr/include/stdio.h
file: /usr/include/stdio.h
file: /usr/include/stdio.h
file: /usr/include/stdio.h
file: /usr/include/stdio.h
file: /usr/include/stdio.h
file: /usr/include/stdio.h
```

## Test

Reproducer: https://github.com/radareorg/radare2-testbins/pull/99

Issue #23560, opened 2024-10-29
dwarf5 wrong addresses


trufae:
Looks like on dwarf5 the address for the addr2line are wrong. tested on macOS Sequoia

Reproducer file attached: [test-dwarf5.zip](https://github.com/user-attachments/files/17558965/test-dwarf5.zip)

NOTE: the dSYM folder is loaded automatically when opening a.out, but it's also psosible to repro the same thing by just loading the macho inside the dSYM directory.

```
$ r2 a.out
[0x100003f28]> CL~addr
addr: 0x00000031
addr: 0x00000041
addr: 0x0000006b
addr: 0x00000051
addr: 0x00000002
addr: 0x0000007a
addr: 0x0000000b
addr: 0x00000066
addr: 0x00000082
addr: 0x00000091
addr: 0x00000016
addr: 0x0000005e
addr: 0x00000026
addr: 0x0000008d
[0x100003f28]>
```

PR #23584, opened 2024-11-02
Fixes for #23579 and #23560

wgrr:

**Description**

In dwarf5 it's not possible to use the same logic to add files to sdb_addrinfo, the 0'th index of directories is valid and emitted by compilers, currently we translate indexes from an array starting at 1 to 0, when a compiler emit a set_file opcode to 0 then the file wont be inserted in sdb_addrinfo because it would underflow, notably clang likes to emit set_file to 0. So i changed the file_names array to always have a minimum of 1 element, this makes the struct work for both dwarf versions, when `dwarf < 5` indexes start at 1 and the 0'th element is all zeros and unused, in dwarf 5 indexes start at 0 and _i think_ it's always present and is the compilation unit path.

Also `dwarf < 5` parser was inserting one extra slash between the base directory and the file, making the output inconsistent with `dwarf 5`, fixed that too and removed an out of date comment from the code.

we could also cleanup those path before inserting in sdb_addrinfo, like removing leading/trailing slash from paths, dot and dotdot that resolve to itself etc. i considered realpath(3) but Windows and i don't feel like it's a good ideia to resolve symlinks here.


trufae:
use ##bin or ##dwarf

trufae:
the file pathr esolution is another problem that is partly covered and maybe an open place to discussion because file paths can be in multiple places, and the user need to handle absolute and relative paths to them.. actually relative paths are needed for the CI to work, because we cant rely on absolute ones in CI systems.

Another issue is that there's no way to actually map the src/dst directories where to translate them. for example, maybe the bin have srces for libc in /home/blah/src/libc, but we want them in /opt/libc.. and other files are located in a different place. right now we can only override a global directory for flat files in there. not sure how complex or well we really want to cover this to not make it difficult to use

wgrr:
> the file pathr esolution is another problem that is partly covered and maybe an open place to discussion because file paths can be in multiple places, and the user need to handle absolute and relative paths to them.. actually relative paths are needed for the CI to work, because we cant rely on absolute ones in CI systems.
> 

Does radare2 attempt to do something if the file pointed by dwarf exists on the users file system? I was thinking more in making  full resolved *dwarf paths* (no file system involved) printing nicer, ie `./foo/./bar.c` be cleaned up with `/foo/bar.c` and the like, just string cleanup, no filesystem involved

wgrr:
Previous commit still using wrong output, just making sure the test passes. I will fix that later, the correct file output should be `/home/pancake/prg/radare2/libr/core/cfile.c` i think.

trufae:
About the // and /./ rvc have this code https://github.com/radareorg/radare2/blob/e25d1beb23397520603ed78d98ed468698fe0abf/libr/util/rvc_rvc.c#L143 which we may want to move into libr/util/file.c after the next release and also handle /./ in there.

maybe we can generalize this function to also handle the abspath in a portable way, but i would probably think in a way for the user to specify via `k` a key=value list of paths to be replaced, so we can have a command to list all the directories used by the dwarf files loaded and where we want to be remapped. and ideally being able to do this automatically via dir.source so the user doesnt need to map all this by hand everytime. 

check this pad i created for further discussions on this topic https://github.com/radareorg/radare2/blob/e25d1beb23397520603ed78d98ed468698fe0abf/libr/util/rvc_rvc.c#L143

PR #23611, opened 2024-11-06
Fix #23610: stop parsing compressed DWARF sections

### LINKED ISSUE #23610: endless loop when parsing compressed DWARF section
## Environment

```
Wed Nov  6 12:22:34 UTC 2024
radare2 5.8.9 31715 @ linux-x86-64
birth: git.5.8.9 2024-10-29__08:52:54
commit: 4eedc8a548bb3cfc898264ee827a086174efebc7
options: gpl -O? cs:5 cl:2 make
Linux x86_64
```

## Description

When parsing DWARF information from debug sections such as `.debug_line`, radare2 does not check whether the section is compressed or not. This means that radare2 ends up parsing compressed content as raw DWARF.

Under specific conditions, this can lead to busy looping.

In the file that triggered this, we have a `.debug_line` section that contains the following bytes:

```
00000000  01 00 00 00 89 0d 00 00  01 00 00 00 78 9c 95 57  |............x..W|

```

When decompressed, we obtain these bytes:

```
00000000  68 00 00 00 03 00 2c 00  00 00 02 01 fb 0e 0d 00  |h.....,.........|
```

Since the content is not decompressed by radare2, the DWARF version is parsed as `0x0d89` (3465) then `parse_line_header_source_dwarf5` is called, which ends up parsing a `total_entries` value of `218404728533093` and going into an (almost) endless loop.

Dumping the DWARF sections in compressed / decompressed state can be done with objcopy:

```
arm-linux-gnueabi-objcopy --dump-section .debug_line=sample.debug_line sample
arm-linux-gnueabi-objcopy --decompress-debug-sections --dump-section .debug_line=sample.debug_line.dec sample
```

## Implementation Hints

There's already a check for sections with `zdebug` in their names:

https://github.com/radareorg/radare2/blob/d56955a0d24eeeae786897a5541644f7fcb5fc10/libr/bin/dwarf.c#L387-L390

A similar approach should be implemented, checking the section flags for `SHF_COMPRESSED`.

## Test

It's unlikely that I can provide the sample, but I'll try to synthesize one.

qkaiser:
Parsing compressed DWARF sections as raw DWARF can lead to endless loops since content is invalid, leading to situations where an overly large total_entries value in `parse_line_header_source_dwarf5` causes radare2 to busyloop.

Fixes #23610

**Description**

If a DWARF section holds the `SHF_COMPRESSED` flag, we should not parse it since compressed DWARF parsing is not supported.

Not all compressed debug sections are marked with a `z`, improves on the incomplete patch at 38318bc588809d2817cdf226c8ff668a1d1087e3


trufae:
we can merge this after the release. athere's actually another PR adding this flag waiting for the ability to breaking season

libr/bin/dwarf.c
```diff
@@ -384,6 +385,10 @@ static RBinSection *getsection(RBin *bin, int sn) {
 		}
 		r_list_foreach (o->sections, iter, section) {
 			if (strstr (section->name, name_str)) {
+				if (strncmp (section->name, ".debug_", 7) == 0 && R_BIN_ELF_SCN_IS_COMPRESSED(section->flags))  {
```
trufae, line 388: use r_str_startswith and add missing space before (

trufae (commented):
can you split the PR to not break abi so we can merg enow?

libr/include/r_bin.h
```diff
@@ -303,6 +303,7 @@ typedef struct r_bin_section_t {
 	ut64 vaddr;
 	ut64 paddr;
 	ut32 perm;
+	ut32 flags;
```
trufae, line 306: wrap this within `#if R2_USE_NEW_ABI`

libr/bin/format/elf/elf.c
```diff
@@ -3747,6 +3747,7 @@ static void _store_bin_sections(ELFOBJ *eo, const RVector *elf_bin_sections) {
 		ptr->type = elf_section_type_tostring (section->type);
 		ptr->add = !eo->phdr; // Load sections if there is no PHDR
 		ptr->perm = elf_flags_to_section_perms (section->flags);
+		ptr->flags = section->flags;
```
trufae, line 3750: same here, i think we can fix the bug without adding this field for now

libr/bin/dwarf.c
qkaiser: Done

libr/include/r_bin.h
qkaiser: Done

libr/bin/format/elf/elf.c
qkaiser: Done

qkaiser:
Wrapped locations referencing `->flags` within `#if R2_USE_NEW_ABI` and tested it with:

```
./configure --with-new-abi
make -j
./binr/radare2/radare2 /tmp/cb81de7321e5f0063b87022799345f13465cbe
ERROR: reloc conversion failed for 0x12000
ERROR: reloc conversion failed for 0x12000
WARN: Compressed dwarf sections not yet supported
WARN: Relocs has not been applied. Please use `-e bin.relocs.apply=true` or `-e bin.cache=true` next time
 -- THIS IS NOT FUNNY
[0x000004e8]>
```

PR #23615, opened 2024-11-07
Check codemeta APIs nullability for #23490

astralia:

**Description**

### DIFF
--- a/libr/anal/codemeta.c
+++ b/libr/anal/codemeta.c
@@ -30,6 +30,7 @@ R_API RCodeMetaItem *r_codemeta_item_clone(RCodeMetaItem *code) {
 }
 
 R_API RCodeMeta *r_codemeta_clone(RCodeMeta *code) {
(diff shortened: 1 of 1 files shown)


libr/anal/codemeta.c
```diff
@@ -41,10 +42,11 @@ R_API RCodeMeta *r_codemeta_clone(RCodeMeta *code) {
 }
 
 R_API RCodeMeta *r_codemeta_new(const char *code) {
+	R_RETURN_VAL_IF_FAIL (code, NULL);
 	RCodeMeta *r = R_NEW0 (RCodeMeta);
 	if (r) {
 		r->tree = r_crbtree_new (NULL);
-		r->code = code? strdup (code): NULL;
+		r->code = strdup (code);
```
trufae, line 49: good catch this argument should never be null! there are many other codepaths that will fail otherwise. so lgtm!

PR #23617, opened 2024-11-07
Check class APIs nullability for #23490

astralia:

**Description**


astralia:
I have some concerns about this one, as some functions call another public functions that would do the same check so maybe those cases could be avoided in order to skip one check, but dunno :(

trufae:
I think it’s fine to merge, the extra checks in static functions are probably not necessary but we had some null derefs spotted in the internal apis in the past so it wont hurt

PR #23634, opened 2024-11-13
Check cond APIs nullability for #23490

astralia:

**Description**


libr/anal/cond.c
```diff
@@ -132,6 +137,7 @@ R_API RAnalCond *r_anal_cond_new_from_op(RAnalOp *op) {
 }
 
 R_API RAnalCond *r_anal_cond_new_from_string(const char *str) {
+	R_RETURN_VAL_IF_FAIL (str, NULL);
 	RAnalCond *cond = R_NEW0 (RAnalCond);
 	// TODO: find '<','=','>','!'...
```
trufae, line 142: would be great to close some TODOs and XXX together with the nullability fixes which look good. do you feel brave enough to give it a chance? :)

trufae: here's sopmething to start with:

```c
R_API RAnalCond *r_anal_cond_new_from_string(const char *str) {
    R_RETURN_VAL_IF_FAIL(str, NULL);
    RAnalCond *cond = R_NEW0(RAnalCond);
    if (!cond) return NULL;

    const char *operators[] = {"==", "!=", ">=", ">", "<=", "<", NULL};
    const int conditions[] = {R_ANAL_COND_EQ, R_ANAL_COND_NE, R_ANAL_COND_GE, R_ANAL_COND_GT, R_ANAL_COND_LE, R_ANAL_COND_LT};
    const char *op = NULL;
    int op_index = -1;

    for (int i = 0; operators[i]; i++) {
        op = strstr(str, operators[i]);
        if (op) {
            op_index = i;
            break;
        }
    }

    if (op_index == -1) {
        free(cond);
        return NULL;
    }

    cond->type = conditions[op_index];

    size_t left_len = op - str;
    char *left = strndup(str, left_len);
    char *right = strdup(op + strlen(operators[op_index]));

    cond->arg[0] = r_anal_value_new_from_string(left);
    cond->arg[1] = r_anal_value_new_from_string(right);

    free(left);
    free(right);

    return cond;
}
```

libr/anal/cond.c
```diff
@@ -69,6 +72,7 @@ static inline const char *condstring(RAnalCond *cond) {
 }
 
 R_API int r_anal_cond_eval(RAnal *anal, RAnalCond *cond) {
+	R_RETURN_VAL_IF_FAIL (anal && cond, false);
 	// XXX: sign issue here?
```
trufae, line 76: We may probably honor the information in arg[0]/arg[1] which are RArchValue types. that contain the /absolute/ field to notify if the field is signed or unsigned. (i know its a pretty bad name, and its not even bool, but we may also fix that abi issue in after the upcoming release

libr/anal/cond.c
astralia: done

PR #23652, opened 2024-11-17
Add plugins in hash search ##crypto

sylvainpelissier:

**Description**

Add plugin during hash search. Add progress bar and tag creation.

### DIFF
--- a/libr/core/cmd_search.inc.c
+++ b/libr/core/cmd_search.inc.c
@@ -84,7 +84,7 @@ static RCoreHelpMessage help_msg_slash_magic = {
 
 static RCoreHelpMessage help_msg_slash = {
 	"Usage:", "/[!bf] [arg]", "Search stuff (see 'e??search' for options)\n"
-	"Use io.va for searching in non virtual addressing spaces",
+				  "Use io.va for searching in non virtual addressing spaces",
 	"/", " foo\\x00", "search for string 'foo\\0'",
 	"/j", " foo\\x00", "search for string 'foo\\0' (json output)",
 	"/!", " ff", "search for first occurrence not matching, command modifier",
@@ -101,7 +101,7 @@ static RCoreHelpMessage help_msg_slash = {
 	"/F", " file [off] [sz]", "search contents of file with offset and size",
 	// TODO: add subcommands to find paths between functions and filter only function names instead of offsets, etc
 	"/g", "[g] [from]", "find all graph paths A to B (/gg follow jumps, see search.count and anal.depth)",
-	"/h", "[?][algorithm] [digest] [size]", "find block of size bytes having this digest. See ph",
+	"/h", "[?*] [algorithm] [digest] [size]", "find block of size bytes having this digest. See ph",
 	"/i", " foo", "search for string 'foo' ignoring case",
 	"/k", " foo", "search for string 'foo' using Rabin Karp alg",
 	"/m", "[?][ebm] magicfile", "search for magic, filesystems or binary headers",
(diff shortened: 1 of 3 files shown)


libr/core/cmd_search.inc.c
```diff
@@ -288,25 +301,24 @@ static int search_hash(RCore *core, const char *hashname, const char *hashstr, u
...
-				// eprintf ("0x%08"PFMT64x" %s\n", from+i, s);
 				if (!strcmp (s, hashstr)) {
-					R_LOG_INFO ("Found at 0x%"PFMT64x, from + i);
-					r_cons_printf ("f hash.%s.%s = 0x%"PFMT64x "\n",
-						hashname, hashstr, from + i);
+					char hash_short[9];
+					strncpy (hash_short, hashstr, 8);
+					r_strf_var (flag, 256, "%s.%s", hashname, hash_short);
+					r_cons_printf ("0x%" PFMT64x ": %s : %s\n", from + i, flag, hashstr);
+					r_flag_set (core->flags, flag, from + i, strlen (hashstr));
```
trufae, line 321: maybe the size should be strlen/2?

libr/core/cmd_search.inc.c
```diff
@@ -288,25 +301,24 @@ static int search_hash(RCore *core, const char *hashname, const char *hashstr, u
...
 					R_LOG_ERROR ("Hash fail");
 					break;
 				}
-				// eprintf ("0x%08"PFMT64x" %s\n", from+i, s);
 				if (!strcmp (s, hashstr)) {
-					R_LOG_INFO ("Found at 0x%"PFMT64x, from + i);
-					r_cons_printf ("f hash.%s.%s = 0x%"PFMT64x "\n",
-						hashname, hashstr, from + i);
+					char hash_short[9];
+					strncpy (hash_short, hashstr, 8);
```
trufae, line 318: i prefer to use r_str_ncpy and sizeof (hash_short) so its easier to maintain and read without having to use arbitrary numbers like 8 and 9

test/db/cmd/hash
```diff
@@ -104,7 +104,7 @@ ph crc32 @ $$+10
 EOF
 EXPECT=<<EOF
 83618b8a
-f hash.crc32.83618b8a = 0x804830a
+0x804830a: crc32.83618b8a : 83618b8a
```
trufae, line 107: uhm is /h* working like before?

trufae (changes requested):

libr/core/cmd_search.inc.c
sylvainpelissier: Right, in fact it is more appropriate to give the block length which was hashed I think.

libr/core/cmd_search.inc.c
sylvainpelissier: Ah yes thank you.

test/db/cmd/hash
sylvainpelissier: It seems `/h*` is not a correct command on master branch. What do you mean ?

trufae: when running a search command with the * subcommand should be printing the commands to set the flags. can you add it in this pr? otherwise im fine to merge it like this and then push another pr with that up 2u

sylvainpelissier: Ok but currently `/h*` does an error. What I did in this PR is that `\h` inserts the flag directly without printing it. Before `/h` simply printed the command to add a flag.

trufae: I know that in master /h behaves like /h* should do even if that command doesnt exist is what it should be done.

* keep your /h 
* Add /h* to print flags without setting any like /h did before

sylvainpelissier: Ok I have added it.

trufae:
thanks!

PR #23663, opened 2024-11-22
WIP: Start refactoring of esil engine

condret:

**Description**

### DIFF
--- a/libr/include/r_esil.h
+++ b/libr/include/r_esil.h
@@ -133,7 +133,10 @@ typedef bool (*REsilMemRead)(void *mem, ut64 addr, ut8 *buf, int len);
 typedef bool (*REsilMemWrite)(void *mem, ut64 addr, const ut8 *buf, int len);
 
 typedef struct r_esil_memory_interface_t {
-	void *mem;
+	union {
+		void *mem;
+		void *user;
+	};
 	REsilMemSwitch mem_switch;
 	REsilMemRead mem_read;
 	REsilMemWrite mem_write;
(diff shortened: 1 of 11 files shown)


libr/include/r_esil.h
```diff
@@ -148,6 +148,29 @@ typedef struct r_esil_callbacks_t {
 	bool (*reg_write)(ESIL *esil, const char *name, ut64 val);
 } REsilCallbacks;
 
+typedef bool (*REsilMemSwitch)(void *mem, ut32 idx);
+typedef bool (*REsilMemRead)(void *mem, ut64 addr, ut8 *buf, int len);
+typedef bool (*REsilMemWrite)(void *mem, ut64 addr, const ut8 *buf, int len);
+
+typedef struct r_esil_memory_interface_t {
+	void *mem;
```
trufae, line 156: userdata?

libr/include/r_esil.h
```diff
@@ -148,6 +148,29 @@ typedef struct r_esil_callbacks_t {
...
+	REsilMemSwitch mem_switch;
```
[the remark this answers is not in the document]
condret: yes, but it's optional, so you could also use something that doesn't provide banking

libr/include/r_esil.h
condret: yes, within r2 this would be RIO *

trufae:
build fails because you cant use io apis from esil library, because it hasnt been built yet. you can use RIOBind instead or try to mess with the library dependency order in library/libs.mk

libr/include/r_esil.h
trufae: Better user userdata instead of mem for cosistency

condret: will change that later

PR #23758, opened 2024-12-06
Add r_core_esil_{load/unload}_arch ##core

condret:

**Description**

### DIFF
--- a/libr/core/core_esil.c
+++ b/libr/core/core_esil.c
@@ -172,6 +172,40 @@ init_fail:
 	return false;
 }
 
+R_API void r_core_esil_load_arch(RCore *core) {
+	R_RETURN_IF_FAIL (core && core->anal && core->anal->arch);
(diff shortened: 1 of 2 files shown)


trufae:
uhm i dont get the point of those apis, can u explain?

condret:
the point is to painlessly load/unload customops from current arch session/plugin into core esil. These wrappers solely exist, because the arch won't let you do this for any instance of esil that lives outside of arch. This design is not final, and would likely change as soon as thing in arch get better. but until then i want 2 functions that are easy to use, so i can unload and load customops when asm.arch/bits/cpu/etc change.

Maybe these functions should also load register profiles

trufae:
Go for it. We have time for more refactorings

trufae:
Was wondering if this was needed to be public

condret:
It propably doesn't need to be part of the public api. however for now i keep it public so that i can call it from #!c during the esil refactoring

Issue #21117, opened 2022-12-01
PLT entries aren't flagged properly


alessandrocarminati:
## Environment

```
Thu 01 Dec 2022 05:59:48 PM CET
radare2 5.7.9 29673 @ linux-x86-64 git.5.7.8-629-gee5f9a635f
commit: ee5f9a635ff53875716bbc5d9f59e2692691db53 build: 2022-12-01__10:15:16
```

## Description

PLT trampoline functions are not named with a proper name, and are left with a name derived by their address.

## Test
```
$ r2 libc.so.6 
WARN: run r2 with -e bin.cache=true to fix relocations in disassembly
 -- Have you seen the latest radare2 TV spot?
[0x0002c980]> aaa
INFO: Analyze all flags starting with sym. and entry0 (aa)
INFO: Analyze all functions arguments/locals (afva@@@F)
INFO: Analyze function calls (aac)
INFO: Analyze len bytes of instructions for references (aar)
INFO: Finding and parsing C++ vtables (avrr)
INFO: Finding function preludes (aap)
INFO: Finding xrefs in noncode section (e anal.in=io.maps.x)
INFO: Analyze value pointers (aav)
INFO: aav: 0x00000000-0x00187ea8 in 0x0-0x187ea8
INFO: Emulate functions to find computed references (aaef)
INFO: Type matching analysis for all functions (aaft)
INFO: Propagate noreturn information (aanr)
INFO: Use -AA or aaaa to perform additional experimental analysis
[0x0002c980]> s sym.__fopen_internal
[0x0006e4f0]> pdf
            ; CODE XREF from sym.fopen @ 0x6e604(x)
┌ 272: sym.__fopen_internal (int64_t arg1, int64_t arg2, int64_t arg3);
│           ; arg int64_t arg1 @ x0
│           ; arg int64_t arg2 @ x1
│           ; arg int64_t arg3 @ x2
│           ; var int64_t var_10h @ sp-0x30
│           ; var int64_t var_20h @ sp-0x20
│           ; var int64_t var_30h @ sp-0x10
│           0x0006e4f0      fd7bbca9       stp x29, x30, [sp, -0x40]!
│           0x0006e4f4      fd030091       mov x29, sp
│           0x0006e4f8      f35301a9       stp x19, x20, [var_10h]
│           0x0006e4fc      f403022a       mov w20, w2                 ; arg3
│           0x0006e500      f55b02a9       stp x21, x22, [var_20h]
[...]

PR #23805, opened 2024-12-17
Fix #17523 - Resolve internal symbols referenced from the PLT ##bin

### LINKED ISSUE #17523: PLT stub names not being resolved properly

trufae:

**Description**

### DIFF
--- a/libr/bin/p/bin_elf.inc.c
+++ b/libr/bin/p/bin_elf.inc.c
@@ -431,7 +431,7 @@ static RList* imports(RBinFile *bf) {
 
 	RBinElfSymbol *is;
 	R_VEC_FOREACH (imports, is) {
-		if (import_symbol->is_internal) {
+		if (is->is_internal) {
 			continue;
 		}
 		RBinImport *ptr = R_NEW0 (RBinImport);
--- a/libr/bin/p/bin_elf.inc.c
+++ b/libr/bin/p/bin_elf.inc.c
@@ -431,6 +431,9 @@ static RList* imports(RBinFile *bf) {
 
 	RBinElfSymbol *is;
 	R_VEC_FOREACH (imports, is) {
+		if (import_symbol->is_internal) {
+			continue;
+		}
 		RBinImport *ptr = R_NEW0 (RBinImport);
 		if (!ptr) {
 			break;
--- a/libr/bin/format/elf/elf.c
+++ b/libr/bin/format/elf/elf.c
@@ -2165,7 +2165,7 @@ ut64 Elf_(get_main_offset)(ELFOBJ *eo) {
 		// non-thumb entry points
 		if (!memcmp (buf, "\x00\xb0\xa0\xe3\x00\xe0\xa0\xe3", 8)) {
 			if (buf[0x40 + 2] == 0xff && buf[0x40 + 3] == 0xeb) {
-				// nothing may happen
+				// nothing to do
 			} else if (!memcmp (buf + 0x28 + 2, "\xff\xeb", 2)) {
 				return Elf_(v2p) (eo, r_read_le32 (&buf[0x34]) & ~1);
 			}
@@ -4375,6 +4375,7 @@ RBinSymbol *Elf_(convert_symbol)(ELFOBJ *eo, RBinElfSymbol *symbol) {
 		ptr->bind = symbol->bind;
 		ptr->type = symbol->type;
 		ptr->is_imported = symbol->is_imported;
+		// ptr->is_internal = symbol->is_internal;
 		ptr->paddr = paddr;
 		ptr->vaddr = vaddr;
 		ptr->size = symbol->size;
--- a/libr/bin/format/elf/elf.h
+++ b/libr/bin/format/elf/elf.h
@@ -61,6 +61,7 @@ typedef struct r_bin_elf_symbol_t {
 	bool is_sht_null;
 	bool is_vaddr; /* when true, offset is virtual address, otherwise it's physical */
 	bool is_imported;
+	bool is_internal;
 } RBinElfSymbol;
 
 typedef struct r_bin_elf_reloc_t {


grgalex:
@trufae The build ('sys/install.sh') fails with:

```
In file included from /home/grg/radare2/libr/..//libr/bin/p/bin_cgc.c:4:
/home/grg/radare2/libr/..//libr/bin/p/bin_elf.inc.c: In function ‘imports’:
/home/grg/radare2/libr/..//libr/bin/p/bin_elf.inc.c:434:21: error: ‘import_symbol’ undeclared (first use in this function)
  434 |                 if (import_symbol->is_internal) {
      |                     ^~~~~~~~~~~~~
/home/grg/radare2/libr/..//libr/bin/p/bin_elf.inc.c:434:21: note: each undeclared identifier is reported only once for each function it appears in
gmake[4]: *** [/home/grg/radare2/libr/../global.mk:45: /home/grg/radare2/libr/..//libr/bin/p/bin_cgc.o] Error 1
gmake[4]: *** Waiting for unfinished jobs....
gmake[3]: *** [/home/grg/radare2/libr/..//libr/rules.mk:87: prelib-build] Error 2
gmake[2]: *** [Makefile:178: bin] Error 2
```

grgalex:
I think the fix for the compile error is to replace `import_symbol` with `is`:
```
diff --git a/libr/bin/p/bin_elf.inc.c b/libr/bin/p/bin_elf.inc.c
index 6ec451a462..607a9777bf 100644
--- a/libr/bin/p/bin_elf.inc.c
+++ b/libr/bin/p/bin_elf.inc.c
@@ -431,7 +431,7 @@ static RList* imports(RBinFile *bf) {
 
        RBinElfSymbol *is;
        R_VEC_FOREACH (imports, is) {
-               if (import_symbol->is_internal) {
+               if (is->is_internal) {
                        continue;
                }
                RBinImport *ptr = R_NEW0 (RBinImport);
```

libr/bin/p/bin_elf.inc.c
```diff
@@ -431,6 +431,9 @@ static RList* imports(RBinFile *bf) {
 
 	RBinElfSymbol *is;
 	R_VEC_FOREACH (imports, is) {
+		if (import_symbol->is_internal) {
```
trufae, line 434:
```suggestion
		if (is->is_internal) {
```

trufae:
doesnt seems to break anything. but .. @grgalex it is improving anything? maybe we should have a test for this before merging :?

grgalex:
@trufae I think there are still commits missing from this PR.

In the issue (#17523), GitHub shows multiple commits referencing the issue.

While some of those commits are duplicates, others offer distinct changes.

This PR only contains (the updated equivalent) of https://github.com/radareorg/radare2/commit/d98edd0568fb6bfce404de8d526af5cacef6e9fe.

All commits in the list from the issue are telling are part of the same story, namely:
1. https://github.com/radareorg/radare2/commit/dd62a5624dae82b72cd95158f5981e554feca5fe
2. https://github.com/radareorg/radare2/commit/b0dc8791068a9af6990cee2db0e6b2c19d2f048f
3. https://github.com/radareorg/radare2/commit/625dc4e6cec661f05352a2bff3b913064ab3613f
4. https://github.com/radareorg/radare2/commit/bbaf7db47159cef90049fb3a287b7d990d5fb418
6. https://github.com/radareorg/radare2/commit/c1d2f349eea803d518aac71a34c933fe28661002
7. https://github.com/radareorg/radare2/commit/53cab17e8fd92ebe1fc33a89d9c97bb9bf722e56
8. [PROBABLY NOT NEEDED] https://github.com/luc-tielen/radare2/commit/0b4b671494496062496570f5a0fec186294352f4
9. https://github.com/radareorg/radare2/commit/d7361366868743ac666ddf27e33f33c9df1305bb
10. https://github.com/radareorg/radare2/commit/dd62a5624dae82b72cd95158f5981e554feca5fe
11. https://github.com/radareorg/radare2/commit/f00bfbf6797b28800f7cac712682d5426b4da941
12. [THIS IS THE ONLY ONE WE INCLUDED IN THIS PR] https://github.com/radareorg/radare2/commit/d98edd0568fb6bfce404de8d526af5cacef6e9fe

Right now, with the commit that this PR has, we don't set the value of the `is_internal` field somehow, yet we check for its value, because the changes from the previous commits are missing (for example we set it in b0dc8791068a9af6990cee2db0e6b2c19d2f048f).

So, we need to apply all those commits in order to get the final proposed change.

As it is right now, with the missing content, the PR can't fix the problem :(

grgalex:
What do you think is the smoothest way to fix this?

(Btw, I think this happened because the PR branch was deleted and we forgot to merge, so now all we have of it are a bunch of lonely commits)

trufae:
I would suggest you to create a PR with the patches you mention and squash/rebase them to make it work, testing them in local and waiting for the CI, adding more tests will be good too.

Which is the PR that wasn't merged? I guess the PR was closed because of conflicts and not clearly fixing the problem, but if you can fix https://github.com/radareorg/radare2/issues/21117

grgalex:
FOUND IT!

https://github.com/radareorg/radare2/pull/21708

This is your original PR that was never merged, and whose branch was deleted.

It's a single commmit too!

So, in essence, the present PR #23805 should include the commit from #21708 and not the one it has now :)

I think it should be very straightforward for you to do this and force-push to this branch.

PR #23816, opened 2024-12-19
Improve 'pdsfq', avoid some dupes and handle fn flags ##print

trufae:

**Description**


libr/core/cmd_print.inc.c
```diff
@@ -3547,6 +3574,7 @@ static void disasm_strings(RCore *core, const char *input, RAnalFunction *fcn) {
 	bool orig_show_offset = show_offset;
 	int asm_tabs = r_config_get_i (core->config, "asm.tabs");
 	bool scr_html = r_config_get_b (core->config, "scr.html");
+	int asm_bytes = r_config_get_i (core->config, "scr.bytes");
```
trufae, line 3577:
```suggestion
	int asm_bytes = r_config_get_i (core->config, "asm.bytes");
```

PR #23859, opened 2025-01-07
Draft: Fix cw command in debug mode ##debug

jukuisma:

**Description**

### DIFF
--- a/libr/io/p/io_ptrace.c
+++ b/libr/io/p/io_ptrace.c
@@ -124,7 +124,7 @@ static int __read(RIO *io, RIODesc *desc, ut8 *buf, int len) {
 				len + sizeof (ptrace_word), aligned_addr);
 		memcpy (buf, aligned_buf + aligned_delta, len);
 		r_free_aligned (aligned_buf);
-		return res;
+		return res - sizeof (ptrace_word);
 	}
 	return -1;
 }


jukuisma:
This fixes the bug reported on discord w/
```
$ r2 -d /bin/ls
[0x7f8ec462cb40]> cw $$ 4 @@/v4 10
[# ]ERROR: Failed to add watcher39000  hits = 0
ERROR: Failed to add watcher
[# ]ERROR: Failed to add watcher39000  hits = 2
[# ]ERROR: Failed to add watcher39000  hits = 3
ERROR: Failed to add watcher
[# ]ERROR: Failed to add watcher39000  hits = 5
0x7f8ec461ad47 hit0_0 0a000000
0x7f8ec461b723 hit0_1 0a000000
0x7f8ec46256d5 hit0_2 0a000000
0x7f8ec4633b61 hit0_3 0a000000
0x7f8ec4635ed1 hit0_4 0a000000
0x7f8ec463795b hit0_5 0a000000
```
I can currently only reproduce this with `ls` and am running into into problems with ASLR when trying to formulate a robust test case for this. Hence the Draft. Let me sleep on this.

trufae:
good catch! unfortunately i think this is a better fix https://github.com/radareorg/radare2/pull/23861 ive tested it and it seems to work well but i would prefer if you could confirm and add a test at least  . thank you!

jukuisma:
```
return R_MIN (len, res);
```
so that we'll return -1 if `debug_os_read_at` fails? Otherwise this should always return the same `len - sizeof (ptrace_word)`. 

> seems to work well but i would prefer if you could confirm

I don't see why this would fail, but double checked just in case. Works as advertised!
```
╭─vagrant@testing ~/scratch/radare2 (master)
╰─➤ r2 -d /bin/ls
WARN: Relocs has not been applied. Please use `-e bin.relocs.apply=true` or `-e bin.cache=true` next time
 -- Pass '-j' to rabin2 to get the information of the binary in JSON format.
[0x7fb534899b40]> cw $$ 4 @@/v4 10
0x7fb534887d47 hit0_0 0a000000
0x7fb534888723 hit0_1 0a000000
0x7fb5348926d5 hit0_2 0a000000
0x7fb5348a0b61 hit0_3 0a000000
0x7fb5348a2ed1 hit0_4 0a000000
0x7fb5348a495b hit0_5 0a000000
```
I'll add test cases in a separate PR if I'll get same half sane test cases working reliably.

trufae:
if you put a printf showing len and res values you'll see that sometimes its way more different than just substracting the last word. i think this solution is good for now and having tests may help us dig a little more if anything goes wrong later

PR #23868, opened 2025-01-08
Fix r_io_bank_locate ##io

condret:

**Description**


trufae:
Will be good to have a test but smaller than the 9MB reproducer :D do u have one?

PR #23893, opened 2025-01-14
Add last pseudo instructions to pickle ##arch

dennis-doyensec:

**Description**

This is still not quite right but all opcodes are supported.

```
r2 -a pickle -qc 'af+ $$ main; afb+ $$ $$ $s;pdc' selfref/dict.pic
 // callconv: x0 reg (x0, x1, x2, x3);
int main (int argc, char **argv, char **envp) {
    loc_0x00000000:
        stack.append({})
        memo[len(memo)] = stack[-1]
        stack.append("selfref") // 0x4
        stack.append(memo[0x0])
        key = stack.pop()// value = stack.pop()// stack[-1][key] = value
        return stack[-1]
        return x0;
}
```
I guess I need to set up analysis to fix return. Also, I don't think I quite understand how pseudo is suppose to work. This parser is getting previous output as input?

So if you want to wait until I figure that stuff out to accept, that is fine.

### DIFF
--- a/libr/arch/p/pickle/dis_helper.inc
+++ b/libr/arch/p/pickle/dis_helper.inc
@@ -1,0 +1,167 @@
+enum opcode {
+	OP_MARK = '(',
+	OP_STOP = '.',
+	OP_POP = '0',
+	OP_POP_MARK = '1',
+	OP_DUP = '2',
+	OP_FLOAT = 'F',
+	OP_INT = 'I',
+	OP_BININT = 'J',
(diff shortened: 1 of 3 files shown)


trufae:
It’s fine to merge it like this. The double return issue is probably a bug in core/pseudo.c which is what takes the asm.pseudo output and generates this pseudocode from the whole function. You can now use r2ai.c or decai to get proper decompiler output from pickle :)

PR #23904, opened 2025-01-16
Fix #23903, compile on pi4, debian bookworm due to missing '_atomic_*' functions.

### LINKED ISSUE #23903: compile/link (r2pm) fails on raspberry pi4, debian bookwoom.
## Environment

[uname.txt](https://github.com/user-attachments/files/18429129/uname.txt)

[gcc.txt](https://github.com/user-attachments/files/18429160/gcc.txt)

[configure-5.9.6.txt](https://github.com/user-attachments/files/18429163/configure-5.9.6.txt)

[configure-5.9.8.txt](https://github.com/user-attachments/files/18429166/configure-5.9.8.txt)

## Description

compile/link fails on raspberry pi4...  debian bookworm...
missing _atomic_* during compile/link of r2pm.

compiling radare2...
v5.9.6: compiles clean.
v5.9.8...master: broken.

[r2pm.txt](https://github.com/user-attachments/files/18429207/r2pm.txt)

bsekisser:
./configure && make fails reporting missing '_atomic_*' functions
	during link of r2pm.

compiled on pi4, uname -a:
Linux raspberrypi 6.6.62+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.62-1+rpt1 (2024-11-25) aarch64 GNU/Linux

with gcc: version 12.2.0 (Raspbian 12.2.0-14+rpi1)

	modified:   mk/gcc.mk

**Description**


mk/gcc.mk
```diff
@@ -8,6 +8,7 @@ AR?=ar
 CC_AR=${AR} q ${LIBAR}
 CFLAGS+=-MD
 CFLAGS_INCLUDE=-I
+LDFLAGS+=-latomic
```
trufae, line 11: this check should detect if its needed at compile time, because this library is not necessary or available in other targets. another option is to check if the output from uname is arm32 (is your pi4 running in arm64 mode?)

```
CHECK_ATOMIC = $(shell echo "int main() { __atomic_fetch_add(0, 0, 0); return 0; }" | \
	$(CC) -x c -o /dev/null - $(CFLAGS) $(LDFLAGS) >/dev/null 2>&1 || echo "-latomic")
DFLAGS += $(CHECK_ATOMIC)
```

bsekisser: bit fuggered up on the commits (doing a github sync put me down a rabbit hole), but here is where it stands...


amended commit...  your suggested revision mostly worked....  two minor changes were needed.  as it were, test would always fail and make stops with circular loop.  also, __atomic_fetch_add by itself is fine except when referencing large values, __atomic_fetch_*_8 nails the problem.


stock bookworm install...  aarch64 kernel, 32bit user-land.

uname -m reports aarch64...  but binaries are 32bit.

so, uname is missleading but, file $(command -v file) shows 32 bit.  that was the one suggested test I found when digging in response to your question.

trufae:
Another option would be to check for libatomic in configure and meson like u did for r2dec

bsekisser:
yes, that was my thought...  while it works for me, seems like there ought to be a better place to do it during configuration...  started looking...  seen the meson build files, believe i tried putting in qjs like i did with quickjs but didnt work/help.  will keep digging, hope to get back to you with something in a few days.

trufae:
lgtm now!

trufae:
ok to remove the draft bit?

PR #23907, opened 2025-01-17
WIP: Fix asan for qjs ##build

satk0:

**Description**


trufae:
It is good to merge now?

trufae:
The whole testsuite is executed with asan. It takes more than an hour but we already check this. And some tests use qjs

satk0:
Oke, so if all the ci checks are passing, I think that's it ;)

trufae:
Seems likt this is affecting macOS only. the fix for this error is the following patch, not sure if it should be pushed upstream..

```
diff --git a/quickjs.c b/quickjs.c
index 3108855..4656d96 100644
--- a/quickjs.c
+++ b/quickjs.c
@@ -1773,9 +1773,13 @@ static inline uintptr_t js_get_stack_pointer(void)

 static inline bool js_check_stack_overflow(JSRuntime *rt, size_t alloca_size)
 {
+#if __has_feature(address_sanitizer)
+    return false;
+#else
     uintptr_t sp;
     sp = js_get_stack_pointer() - alloca_size;
     return unlikely(sp < rt->stack_limit);
+#endif
 }

 JSRuntime *JS_NewRuntime2(const JSMallocFunctions *mf, void *opaque)
```

trufae:
I'm merging this, but according to https://github.com/quickjs-ng/quickjs/issues/841 we can just handle this at runtime by initializing the jsvm with unlimited stack size on macOS, i will address that in another PR after merging this.

Thank you!

satk0:
Thanks for the merge! Time for capstone now ;)

PR #23908, opened 2025-01-17
R2600 renaming

jukuisma:

**Description**

### DIFF
--- a/libr/anal/flirt.c
+++ b/libr/anal/flirt.c
@@ -1388,7 +1388,7 @@ static inline void flirt_free(RFlirt *f) {
 
 // dump a flirt signature content on screen
 R_API void r_sign_flirt_dump(const RAnal *anal, const char *flirt_file) {
-	RBuffer *flirt_buf = r_buf_new_slurp (flirt_file);
+	RBuffer *flirt_buf = r_buf_new_from_file (flirt_file);
 	if (!flirt_buf) {
 		R_LOG_ERROR ("Can't slurp %s", flirt_file);
 		return;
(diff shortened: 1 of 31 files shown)


jukuisma:
```diff
diff --git a/libr/flag/flag.c b/libr/flag/flag.c
index 1fc8765f3d..1cfcda26ad 100644
--- a/libr/flag/flag.c
+++ b/libr/flag/flag.c
@@ -509,8 +509,7 @@ R_API RFlagItem *r_flag_get(RFlag *f, const char *name) {
 }

 /* return the first flag item that can be found at offset "off", or NULL otherwise */
-// R2_600 - rename to r_flag_get_at and r_flag_get_in ??
-R_API RFlagItem *r_flag_get_i(RFlag *f, ut64 off) {
+R_API RFlagItem *r_flag_get_item(RFlag *f, ut64 off) {
        R_RETURN_VAL_IF_FAIL (f, NULL);
        if (f->mask) {
                off &= f->mask;
```
Not exactly the suggestion, but `r_flag_get_item` seems more logical if `get_i` isn't clear enough.

jukuisma:
Renamed
```diff
-R_API RBuffer *r_buf_new_slurp(const char *file);
+R_API RBuffer *r_buf_new_from_file(const char *file);
```
But there's also:
```C
215 // rename to new?
  1 R_API RBuffer *r_buf_new_mmap(const char *filename, int perm) {
  2     R_RETURN_VAL_IF_FAIL (filename, NULL);
  3     struct buf_mmap_user u = {filename, perm};
  4     return new_buffer (R_BUFFER_MMAP, &u);
  5 }
```
without the R2_600 comment. Do we want to rename this while we're at it?

trufae:
All those changes look good to me :) thanks for taking the time to do such a painful renaming!

jukuisma:
> All those changes look good to me :) thanks for taking the time to do such a painful renaming!

No problem, the following script helps with the pain ;)
```bash
#!/usr/bin/bash

if [ "$#" -ne 2 ]; then
    echo "Must be 2 parameters: e.g. 'replace replace_this with_this'"
    return 1
fi

limit=50  # if more than 50 grep results found, do nothing
search="$1"
replace="$2"
length=$(grep -rlI --color=never "$search" | wc -l)

if [ "$length" -ne 0 ] && [ "$length" -lt $limit ]; then
    for file in $(grep -rlI --color=never "$search")
    do
        nvim -c "%s/$search/$replace/gc" -c 'wq' "$file"
    done
else
    echo "$length results found, nothing done"
fi
```

PR #23931, opened 2025-01-26
Add string formatting functions similar to asprintf()

rtgill82:

**Description**

Add a string formatting functions that automatically allocate memory similarly to asprintf(). Also provide a function that will automatically reallocate a buffer as necessary, allowing a single buffer to be reused without constantly having to malloc()/free().


rtgill82:
There's very little error handling in `my_vsnprintf()`, preventing it from being used in these functions. `vsnprintf()` is used instead, so these functions do not support the custom format specifiers.

rtgill82:
I wasn't aware of those. I'm working on a DOSEMU plugin that I was using these functions for.

PR #23963, opened 2025-02-17
r2pm: Add env for man pages.

teasp00n:
This commit adds an environment variable for `MANDIR` which can be used by `r2pm` managed plugins to determine where to place any manfiles they contain.

**Description**

Currently `r2pm` doesn't provide a consistent way to handle manfiles provided for the plugins it installs. This MR adds a directory under the `r2_prefix` for this. It'll end up being something like `$HOME/.local/share/radare2/prefix/man`.

NOTE: This relies on the behavior of how a `manpath` is constructed where `../man` (and `../share/man`) are both added to the `manpath` if they exist for a given element in `PATH`. This means manpages will only be accessible if the user has added the `R2PM_BINDIR` to their `PATH`. We deliberately do _not_ set the `MANPATH` explicitly as it will override manpage discovery based on `PATH` which most users are probably relying on for other things.

### DIFF
--- a/libr/main/r2pm.c
+++ b/libr/main/r2pm.c
@@ -430,6 +430,10 @@ static void r2pm_setenv(R2Pm *r2pm) {
 	r_sys_setenv ("R2PM_BINDIR", bindir);
 	free (bindir);
 
+	char *mandir = r_str_newf ("%s/man", r2_prefix);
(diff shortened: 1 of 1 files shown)


trufae:
a separate one works fine to me, thanks for taking care of it! merged

PR #23981, opened 2025-02-19
Fix: Potential Vulnerability in magic Library

tabudz:

**Description**
This PR fixes a security vulnerability in __magic_file_check_mem() that was cloned from file/file but did not receive the security patch. The original issue was reported and fixed under https://github.com/file/file/commit/6713ca45e7757297381f4b4cdb9cf5e624a9ad36.

This PR applies the same patch to eliminate the vulnerability.

**References**
https://nvd.nist.gov/vuln/detail/cve-2015-8865
https://github.com/file/file/commit/6713ca45e7757297381f4b4cdb9cf5e624a9ad36

### DIFF
--- a/libr/magic/funcs.c
+++ b/libr/magic/funcs.c
@@ -322,7 +322,7 @@ const char *__magic_file_getbuffer(RMagic *ms) {
 
 int __magic_file_check_mem(RMagic *ms, unsigned int level) {
 	if (level >= ms->c.len) {
-		size_t len = (ms->c.len += 20) * sizeof (*ms->c.li);
+		size_t len = (ms->c.len = 20 + level) * sizeof (*ms->c.li);
 		ms->c.li = (!ms->c.li) ? malloc (len) :
 		    realloc (ms->c.li, len);
 		if (!ms->c.li) {
--- a/libr/magic/funcs.c
+++ b/libr/magic/funcs.c
@@ -322,7 +322,8 @@ const char *__magic_file_getbuffer(RMagic *ms) {
 
 int __magic_file_check_mem(RMagic *ms, unsigned int level) {
 	if (level >= ms->c.len) {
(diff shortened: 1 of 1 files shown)


libr/magic/funcs.c
```diff
@@ -322,7 +322,7 @@ const char *__magic_file_getbuffer(RMagic *ms) {
 
 int __magic_file_check_mem(RMagic *ms, unsigned int level) {
 	if (level >= ms->c.len) {
-		size_t len = (ms->c.len += 20) * sizeof (*ms->c.li);
+		size_t len = (ms->c.len = 20 + level) * sizeof (*ms->c.li);
```
trufae, line 325: maybe better like this?

```suggestion
		ms->c.len += 20 + level;
		size_t len = ms->c.len * sizeof (*ms->c.li);
```

tabudz: Not so sure tbh. I'm running a clone detection tool and found the clone then suggested the fix based on the original fix. And the original fix is '=' not '+='. Anw, you should have a better grasp than I do about this.

trufae: ah yes, the level must be larger than c.len, there's no need to make the increment exponential. i was just splitting the line in two for readability. the code quality of libfile is pretty bad, i would like to rewrite it completely at some point, but at least having a working implementation was good. i was not planning to keep in sync

tabudz:
Hi @trufae, thanks for merging my PR. Just want to let you know that I plan to report this as a CVE. Let me know if you have any concern!

trufae:
All good

nluedtke:
Looks like this received CVE-2025-1864

Crispy-fried-chicken:
Hi, I really want to know how to reproduce it. Thank you!

kingkingkingson:
> Hi, I really want to know how to reproduce it. Thank you!
I also want to know.

PR #23989, opened 2025-02-20
Split linux-64 and linux-32 tests multilib is dead ##debug

trufae:

**Description**


jukuisma:
LGTM, should definitely help debugging if it's only `x86_32` paths that fail the tests.

PR #24011, opened 2025-02-27
Add test for aavr on async swift binary

mrmacete:

**Description**

Tests the usage of `aavr` to find references to async Swift tasks, by setting `anal.in` accordingly.

Depends on https://github.com/radareorg/radare2-testbins/pull/105


trufae:
will be good to have another test running 'aaa' instead of 'aavr', and modify aaa accordingly to run aavr internally when the target bin is swift. wdyt?

PR #24016, opened 2025-02-27
Fix type for uname bits

mrmacete:

**Description**

Before:
```
{"platform":"Darwin","arch":"arm","bits":1}
```
After:
```
{"platform":"Darwin","arch":"arm","bits":64}
```

spotted because of a compiler warning


trufae:
ooops. good catch

PR #24022, opened 2025-02-28
Skip __objc_catlist2 in dyldcache parsing ##bin

mrmacete:

**Description**

Because it's not the same as `__objc_catlist`, so attempting to parse the same structures there fails with noisy warnings.


trufae:
maybe its better to use r_str_endswith ("objc_catlist") instead? we never know if catlist3 will even hit the theaters

trufae:
as long as we have no tests. can you confirm this patch works? https://github.com/radareorg/radare2/pull/24026 and maybe good to add a reproducer/check

trufae:
thanks!

PR #24045, opened 2025-03-05
WIP: rlz4

MewtR:

**Description**

Support for rlz4 (r2's lz4):
Copy pasted the existing `lz4_decompress` into `lz4_decompress_block`. These are essentially the same, except that the latter writes output to a new buffer. I tested with the nso binary found [here](https://github.com/radareorg/radare2-testbins/blob/master/nso/application.nso) and made sure the output of rlz4 when decompressing is exactly the same as that of lz4/smallz4.

I still have to do some cleanup to do and add support for it in meson builds. Let me know what you think.


trufae:
Public apis must start with r_ and be tagged with the R_API define. Can the decompress function use the decompress_block one?

trufae:
rebase to fix the red job

MewtR:
Ok I rebased and made the function part of the API. For now r2 only has need for decompressing lz4 blocks as opposed to frames.
The new function is just because the other one kept running out of space when decompressing. As a bonus, it takes similar arguments to the lz4/smallz4 decompression functions now.

MewtR:
Yeah there's probably a way to merge the two functions. I'll do that as well as remove smallz4 sometime this week.

trufae:
Sounds good!

MewtR:
Ok, I removed smallz4. We also have only a single decompression function in rlz4. One thing I noticed though is that rlz4 can't decompress files that were compressed with lz4, it seems it doesn't recognized the format with `0x184D2204` magic.
Decompressing blocks is not a problem or files compressed by rlz4 works as expected.

trufae:
That’s weird maybe that requires some investigation

trufae:
https://github.com/lz4/lz4/blob/dev/doc/lz4_Frame_format.md search for https://github.com/lz4/lz4/blob/dev/doc/lz4_Frame_format.md

trufae:
vibe fix https://chatgpt.com/share/67dde3d0-b1e4-8011-995a-76417bd6ca0a

MewtR:
Haha yeah I was going to look into it but I've been a bit busy lately. I'll try to get it done next week

MewtR:
Alright, we should be good now. The only problem is that it fails sometimes due to lack of memory, the fix would be to increase `BLOCK_SIZE`. Note that this is only a problem when decompressing frames, when decompressing blocks directly we pass in our own max size.

trufae:
Looks good! Can you drop the draft state?

trufae:
Looks good!

trufae:
Can you fix this crash?
Screenshot from 2025-03-28 12-42-20
Screenshot from 2025-03-28 12-42-27
Screenshot from 2025-03-28 12-42-34

MewtR:
I'll take a look

PR #24112, opened 2025-04-05
Initial support for PEF/macOS(7-9) executables ##bin

trufae:

**Description**


elliotnunn:
This is a very happy surprise! From pancake themself no less!

I have been working on a plugin myself, which we can pick code from: https://github.com/elliotnunn/radare2/commits/master/

These are the outstanding problems as I see them:

### "Pidata" decompression
RW data is usually compressed in PEF. This fits awkwardly with cbin.c, which expects to be able to memory-map the file directly ± appending zeros. Could you please look over https://github.com/elliotnunn/radare2/commit/dcc10d7 as my proposed adaptation for compressed segments?

### The "table of contents" data model
The Mac OS PowerPC runtime environment is approximately copied from the AIX one, so figuring this out will be useful for that OS too. PowerPC lacks PC-relative addressing, so in PEF/XCOFF32 the "r2" general purpose register is reserved as a base register for data accesses. The "rTOC" points to the data section, often with an offset so that negative and positive offsets can be used.

The rTOC is set by the **caller** as read from a fat function descriptor called a "transition vector": `struct TVector {void *code; void *rTOC};`. Therefore, all of the function exports of a PEF or XCOFF32 are actually 8-byte structures inside the data section.

1. Should the radare bin-format plugin be responsible for chasing exported TVectors and adding function flags to the code itself?
2. Or should this be a new analysis step, to prevent code-duplication between PEF and XCOFF32?
3. Could we add a new "flagspace" for TVectors so that they can be labelled without being confused with their associated functions?

And we can use the rTOC to recognise data access inside the disassembly, but I am not sure how to do this.

4. Should a new analysis step be created to associate rTOC-based access with a memory address? (e.g. `lwz r3,0x66(r2) # access to variable at 0x100066`)

Thank you so much for your interest in adding classic Mac OS support to radare. I have a largish reverse engineering project coming up and I need radare support to get it done.

trufae:
First of all.. DAMN, i hate dupping work. I guess i may close the PR, but can you submit your implementation in exchange?

Even preliminary work it will be good to have it merged so we can improve code quality, add tests and discuss future directions from there.

Some quick thoughs on your points:

* which compression algo is used in these sections?
* can you submit sample bins (with no redistribution restrictions in the testbins repo)?
* about the compressed section maps, i think it's close to what we have already in other bin plugins like dyldcache patching, and maybe having an io plugin handling that can be an option. maybe @condret can share some thoughs on this too, other plugins with compressed sections are xalz and elf with compressed dwarf sections. but i think the usecases for those are different.

About the rTOC thing. i see different solutions to the problem:

* code duplication between bin plugins is sometimes "good" for stability
* this rtoc seems like relocs to me, something that patches instructions to add references for the given pointer. the problem here is that on ppc you can't have absolute addressing. so how did you though as a solution to this?
  * add comments and xrefs?
  * expose and rbin command (See `i:` command)
  * add a generic `aav` subcommand for that
  * add anal hints to tweak the instruction details?
  * rely on esil emulation (aae) using `r2` as global pointer?

For the last solution, which imho seems like the less hacky. i would pick arm64's flutter trick as inspiration. flutter compiler uses a register as "GP" register, which is enforced at runtime by r2 when it identifies the binary as flutter one. Then you can use `anal.gp` to set this value from rbin. You can see how that stuff works in mips, because that's replaced in the disasm without depending on emulation.

In other words. i think there's no single solution to the problem and whatever is made it's usually better to require no extra steps to work. But i think, that, considering other binary files like COFF have similar characteristics we may find a common approach for consistency and clarity.

trufae:
btw i have this tab open in my browser for a veeeery long time, and i guess you know it too. but i think its worth sharing it just for the record. https://vintageapple.org/macprogramming/

elliotnunn:
> Even preliminary work it will be good to have it merged so we can improve code quality, add tests and discuss future directions from there.

All right, I will do this to expose it to some more eyeballs. But based on your implementation I will convert my many `r_buf_read` calls to struct accesses.

> * which compression algo is used in these sections?

I think it is unique. Described here: https://dev.os9.ca/techpubs/mac/runtimehtml/RTArch-94.html#HEADING94-3

> * can you submit sample bins (with no redistribution restrictions in the testbins repo)?

Yes, I will make some.

> * about the compressed section maps, i think it's close to what we have already in other bin plugins like dyldcache patching, and maybe having an io plugin handling that can be an option. maybe @condret can share some thoughs on this too, other plugins with compressed sections are xalz and elf with compressed dwarf sections. but i think the usecases for those are different.

I would appreciate any advice.

> About the rTOC thing. i see different solutions to the problem:
>   * rely on esil emulation (aae) using `r2` as global pointer?

This might be the way to go, even if it requires "TOC-finding" code to be duplicated across PEF and XCOFF32. I will read the "flutter" code that you mentioned. In some rare cases it will be necessary to use heuristics to determine the rTOC: one PEF that I found exported only a struct full of function pointers, so it was difficult to determine what exactly was or was not a TVector.

trufae:
fixed all the things and merged to master!

trufae:
ive merged the current approach, because it works and i think we may want to do more refactorings and discussions before we reach an agreement on what would be the best way to go. thanks @condret ! it will be good to have more tests so we can ensure behaviour is correct

PR #24116, opened 2025-04-07
Pefpef

trufae:

**Description**


libr/include/r_bin.h
```diff
@@ -312,6 +312,7 @@ typedef struct r_bin_section_t {
 	bool add; // indicates when you want to add the section to io `S` command
 	bool is_data;
 	bool is_segment;
+	RIODesc *backing;
```
trufae: changed that to use `int` instead of `RIODesc` here

PR #24127, opened 2025-04-09
Forbid usage of dot in function name upon signature change afs

alex-bender:
Abort `afs` command if `.` present in function name from signature


alex-bender:
Should I remove test `db/cmd/types afs fcnname with dots` now?

alex-bender:
it would be better to support dot in functio names instead of warning. closing that PR

PR #24220, opened 2025-05-19
Fix Fish Shell Detection In The Install Script 

iteroji:

**Description**

The installation script `sys/install.sh` used `${SHELL} --help` to detect  fish shell, however the `fish` command does not have a `--help` argument. I changed it to `--version` which outputs something like `fish, version 4.0.0-g1e069b0`.

### DIFF
--- a/sys/install.sh
+++ b/sys/install.sh
@@ -165,7 +165,7 @@ fi
 umask 0002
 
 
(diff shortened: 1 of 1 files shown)


trufae:
Good catch! I’m not a fish user but this change should be good for other shells too

Issue #24217, opened 2025-05-17
`dmh`: broken in latest radare2 git version.


ishmael-42:
## Environment

```sh
Saturday 17 May 2025 XXXXXXXXXXXXXXX
radare2 5.9.9 33662 @ linux-x86-32
birth: git.5.9.8-663-g2bdbbb25b6 2025-05-15__17:33:49
commit: 2bdbbb25b6c6ba5cdd69e267a65c5986c1fd618c
options: gpl -O? cs:5 cl:2 make
Linux x86_64
```

## Description

In radare2 compiled from HEAD, the `dmh` command is broken.

In the left, is radare2 installed from package manager  (`argv[0] = /usr/bin/radare2`) `[version 5.9.8]`
In the right, is radare2 installed by `configure && make` (`argv[0] = /usr/local/bin/radare2`)

![Image](https://github.com/user-attachments/assets/bab967a5-d923-4148-8cea-51f099ecce62)

## Test

```bash
$ cat <<EOF >main.c
#include <stdlib.h>

int main(void) {
  void *a = malloc(0x0);
  void *b = malloc(0x8);
  void *c = malloc(0x10);
  void *d = malloc(0x18);
  void *e = malloc(0x20);
  void *f = malloc(0x28);
  void *g = malloc(0x30);

  free(a);
  free(b);
  free(c);
  free(d);
  free(e);
  free(f);
  free(g);
}

EOF

$ gcc main.c
```
```bash
$ r2 -d a.out -A
[0x759da42455c0]> dc
[0x759da40c4f68]> dmh
```

### Expectation
(This is with radare2 version 5.9.8, installed from OS package manager)
![Image](https://github.com/user-attachments/assets/6a6969b0-38c0-43ce-9212-2d461f418217)

### Reality
(This is with radare2 git, installed with `./configure` && `make`)
![Image](https://github.com/user-attachments/assets/71f30a64-48a9-4657-8155-b493daa09414)

ishmael-42:
I checked out the package manager's version of radare2 source

```
git worktree add ../r2-arch 5.9.8
```

package manager's version of radare2 (`~/repos/r2-arch`) 
current version of radare2 (`~/repos/radare2`).

And found the following diffs  


TL;DR, it mostly has only nullchecks for `R2_NEW0`, and the changes caused by rename of `RCore::offset` to  `RCore::addr` 

IG this is caused by some other factor.


```diff
[...]

PR #24221, opened 2025-05-21
fix SZ for glibc heap commands

koyaan:

**Description**

`SZ`  is expected by the code to be the size of pointers in bytes not bits, unsure how this ever worked before


trufae:
Ooh there was a change because bits was used as a bitmask, as bytes size and bits size in different places making everything very confusing. The refactoring to unify all this ended up breaking things that had no tests like this one!! Good catch and thanks for digging!

trufae:
Not sure how many times this SZ is used but maybe better to move it into a function that ensures the value is 64 or 32 and nothing else, because if the debug target is running a 32 bit target on a 64bit host the value can be tricky.. but i accept this patch as is, its just a recommendation to improve code quality for better maintainance

koyaan:
Ah i see!
I am already investigating why tests fail on libc 2.35

trufae:
Related to https://github.com/radareorg/radare2/issues/24217#issuecomment-2902063214

trufae:
Looks good! Can you change the commit message to follow the style:

- first letter must be capitalized
- Use ##debug hashtag so the fix appears in the chlog

trufae:
I think your patch is fixing the referenced issue

koyaan:
done.
And yes that issue is fixed by this!

```
$ r2 -A -d issue_24217 
[0x7ffff7fe3480]> dc
(142067) Process exited with status=0x0
[0x7ffff7e73f1d]> dmh
INFO: libc version 2.41 identified from .rodata banner
WARN: Could not find main_arena via symbol or relocations
WARN: Found main_arena offset with pattern matching

  Malloc chunk @ 0x555555559290 [size: 0x20][free]
  Malloc chunk @ 0x5555555592b0 [size: 0x20][free]
  Malloc chunk @ 0x5555555592d0 [size: 0x20][free]
  Malloc chunk @ 0x5555555592f0 [size: 0x20][free]
  Malloc chunk @ 0x555555559310 [size: 0x30][free]
  Malloc chunk @ 0x555555559340 [size: 0x30][free]
  Malloc chunk @ 0x555555559370 [size: 0x40][free]
  Top chunk @ 0x5555555593b0 - [brk_start: 0x555555559000, brk_end: 0x55555557a000]

[0x7ffff7e73f1d]> 
```

trufae:
Awesome!

PR #24267, opened 2025-06-06
Fix FreeBSD x64 register profile ##debug

jukuisma:

**Description**

rflags was defined twice and always printed warnings:

```
WARN: Duplicated register definition for 'rflags' has been ignored
```

### DIFF
--- a/libr/debug/p/native/reg/kfbsd-x64.h
+++ b/libr/debug/p/native/reg/kfbsd-x64.h
@@ -76,7 +76,6 @@ return strdup (
 "flg	eflags	.32	152	0	c1p.a.zstido.n.rv\n"
 "gpr	rsp	.64	160	0\n"
 "seg	ss	.64	168	0\n"
-"flg	rflags	.64	144	0	c1p.a.zstido.n.rv\n"
 "flg	cf	.1	.1216	0	carry\n"
 "flg	pf	.1	.1218	0	parity\n"
 "flg	af	.1	.1220	0	adjust\n"


trufae:
I think this rflags is ok. The problem is the seond one defined few lines below. See the offset differs

jukuisma:
> I think this rflags is ok. The problem is the seond one defined few lines below. See the offset differs

Whoops, switched back the other definition here: https://github.com/radareorg/radare2/pull/24267#issuecomment-2949636113. However, didn't have time to test it yet. I'll run some debugging to see what the registers actually look like. Gotta also check if any existing test cases catch this and consider adding new ones if not. FreeBSD isn't part of the github pipelines, but ideally the tests would catch this if running manually.

I'll get back to you on this when I have some time.

jukuisma:
> I'll run some debugging to see what the registers actually look like.

Testing with:
```
$ cat main.c
int main () {
        __asm__ volatile ("int3");
}
```
Master and current PR always give:
```
$ for i in (seq 1 100)
      r2 -dqc 'dc; dr rflags' ~/a.out 2>&1 | grep -vE 'INFO|WARN'
  end | sort | uniq
0x00000202
0x00000206
```
i.e., interrupt flag always set and parity essentially random: https://en.wikipedia.org/wiki/FLAGS_register#FLAGS. Removing the rflags definition with correct offset gives garbage as expected:
```
$ for i in (seq 1 100)
      r2 -dqc 'dc; dr rflags' ~/a.out 2>&1 | grep -vE 'INFO|WARN'
  end | sort | uniq
0x00000043
```

> Gotta also check if any existing test cases catch this and consider adding new ones if not.

There aren't many tests that check rflags and many of them use linux binaries that straight up don't work on FreeBSD. Adding FreeBSD tests could be cool, but requires a lot more work than this little fix warrants IMO. I'd say this is good to merge now on my behalf and we can consider FreeBSD testing later if needed.

trufae:
I dont think we can easily run freebsd tests in github ci without shipping a bloated qemu image. But there are external services for bsd ci, and i agree that it would be cool to have some more debugger tests for other than linux

PR #24334, opened 2025-06-20
Add claude-code

dnakov:

**Description**


trufae:
do we need to pay for that?

trufae:
Would be great to have but i dont pay for it, and the pr was already red, so maybe if that can be enabled in the future we can make it as an alternative companion

PR #24401, opened 2025-07-07
Fix install script

frukto:
Move argument parsing to the beginning. Otherwise the PREFIX is evaluated in sudosetup before it is set.

**Description**

`sys/install.sh` tries to handle a situation where no `sudo` is required, e.g. `install.sh --prefix=/home/frukto/.local`. However the script fails to properly detect a user writeable prefix, because this is checked before arguments are parsed. This patch moves the argument parsing to the beginning.

### DIFF
--- a/sys/install.sh
+++ b/sys/install.sh
@@ -36,6 +36,43 @@ sudosetup() {
 	fi
 }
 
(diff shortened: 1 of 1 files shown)


trufae:
you have sys/user.sh to install r2 in your home without sudo. not sure if we want to skip that case in sys/install

PR #24415, opened 2025-07-10
Add r2sdb to list of bin files

frukto:

**Description**

Add `r2sdb` to list of bin files in the Makefile. Otherwise it will not be removed by a `make purge`.


trufae:
good spot

PR #24429, opened 2025-07-16
Make zip a subproject ##build

satk0:

**Description**


trufae:
lets change the plan:
- lets use mrzip as a subproject
- if users want to use libzip they can link against system libzip
- this way we have full control on the code and we dont ship 3rd party code

PR #24459, opened 2025-08-11
Avoid having files twice after binparsing with uri ##bin

trufae:

**Description**

### DIFF
--- a/test/db/cmd/cmd_io
+++ b/test/db/cmd/cmd_io
@@ -129,7 +129,6 @@ EOF
 EXPECT=<<EOF
  3 * r-x 0x0002daa0 bins/mach0/ls-m1
 --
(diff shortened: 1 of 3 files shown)


test/db/cmd/cmd_io
```diff
@@ -129,7 +129,6 @@ EOF
 EXPECT=<<EOF
  3 * r-x 0x0002daa0 bins/mach0/ls-m1
 --
- 3 * r-x 0x0002daa0 bins/mach0/ls-m1
  5 - rw- 0x00000008 null://8
```
trufae:
```suggestion
 3 * r-x 0x0002daa0 bins/mach0/ls-m1
 5 - rw- 0x00000008 null://8
```

PR #24472, opened 2025-08-13
Re-enable the abi diff checks ##ci

trufae:

**Description**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -612,7 +612,7 @@ jobs:
 
 
   check_abi_compatibility:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -612,7 +612,7 @@ jobs:
 
 
   check_abi_compatibility:
-    if: contains(github.ref, 'de345ee04d0777b462f2f98ac383c883')
+#    if: contains(github.ref, 'de345ee04d0777b462f2f98ac383c883')
```
trufae:
```suggestion
    if: contains(github.ref, 'invalid-commit-just-to-skip-that-job-until-rls-fixed')
```

PR #24480, opened 2025-08-15
Unable to execute `.\preconfigure.bat` ##build

simexce:

**Description**

Solve the problems encountered in the process of compiling radare2.

### DIFF
--- a/configure.bat
+++ b/configure.bat
@@ -1,4 +1,5 @@
 @echo off
+setlocal enabledelayedexpansion
 REM call preconfigure.bat
 
 set MESON_FLAGS=-Dsdb_cgen=false -Dc_std=c11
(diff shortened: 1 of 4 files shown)


configure.bat
```diff
@@ -18,7 +18,7 @@ if "%*" == "release" (
 set PATH=%CD%\prefix\bin;%PATH%
 set WORKS=0
 if EXIST vs (
-  meson vs %MESON_FLAGS% --backend vs --reconfigure && set WORKS=1
+  meson vs %MESON_FLAGS% --reconfigure && set WORKS=1
```
trufae: I think thats not correct because that wont create the visual studio project

trufae: can you revert that change?

trufae:
```suggestion
  meson vs %MESON_FLAGS% --backend vs --reconfigure && set WORKS=1
```

trufae (changes requested):
.

configure.bat
trufae: Uhm?

trufae:
that's weird, why meson 1.8 is checked to do the `vs` backend or not? was the support removed? i cant find anything in the changelog. anyway, the `vs` directory is just a helper for those who want to use vs to build and debug r2, so it is not required to build, and imho it shuoldnt be a blocker if it fails. i prefer to remove all the boilerplate checks and keep it simple

make.bat
```diff
@@ -1,7 +1,9 @@
 :: Build (and eventually install) r2 for windows
 @echo off
 if "%*" == "clean" (
-  wsl rm -rf b vs
+  REM wsl rm -rf b vs
```
trufae:
```suggestion
```

configure.bat
```diff
@@ -18,7 +19,12 @@ if "%*" == "release" (
 set PATH=%CD%\prefix\bin;%PATH%
 set WORKS=0
 if EXIST vs (
-  meson vs %MESON_FLAGS% --backend vs --reconfigure && set WORKS=1
+  for /f "delims=" %%i in ('meson -v') do set MESON_VERSION=%%i
```
trufae, line 22: imho there's no sense in overengineering this check, because meson should be able to build the visual studio project files without depending on the version. we can just ignore the error return in any case. but meh. im fine to merge it this way and we can refine it later

trufae (approved):
wip

PR #24493, opened 2025-08-22
Implement Initial sBPF support

ulexec:
- [ ] Mark this if you consider it ready to merge
- [ ] I've added tests (optional)
- [ ] I wrote some lines in the [book](https://github.com/radareorg/radare2book) (optional)

**Description**
 These changes implement initial support for sBPF ELF binaries. Include the following features:
 - Map file appropriately
 - Applies sBPF specific ELF relocations
 - Implements heuristics to identify length of Rust strings
 - Annotates Solana Syscalls
 - Applies appropriate ESIL emulation for instructions

PR #24501, opened 2025-08-26
Add support for better Rust string hints

ulexec:

**Description**
Improve rust string hints used by sBPF arch plugin (https://github.com/radareorg/radare2/pull/24493)


ulexec:
seems some test do not pass, its likely this is not the best way to implement what I intended. Any suggestions welcomed!

trufae:
can we close this now?

PR #24511, opened 2025-08-28
Refine the core IOKit types ##types

oleavr:

**Description**

My first iteration ended up broken because midway through I renamed `_ClassName` to `ClassName_Data`, but forgot to update the instance struct references to it. This splitting of data from the instance struct was also done in cases where it wasn't needed, i.e. for structs/classes without a vtable.


trufae:
ok to merge? or what's the followup?

oleavr:
Haven't yet reached the point where I can test this, so there might be more bugs lurking. But this should be an improvement regardless, so think it can be merged.

PR #24527, opened 2025-08-31
xnu_debug: fixes for powerpc

barracuda156:
There is no such header `mach/ppc/_types.h`. There is an arch-specific `ppc/_types.h` and generic `machine/_types.h`, which includes respective arch-specific headers. Since here it is used only for powerpc, `ppc/_types.h` should be the right one.


barracuda156:
@trufae Okay, now it builds finally. With fixes from this PR plus https://github.com/radareorg/radare2/pull/24526

PR #24541, opened 2025-09-04
Plugs

trufae:

**Description**


trufae:
Now -l is redundant. but as long as its a lowercase L i dont have much appreciation for it and we can kill it

PR #24584, opened 2025-09-22
Fix #24325 - Another proposal to address this resize loop issue

### LINKED ISSUE #24325: Out of memory when using r- command with seeking
## Environment

```sh
# copypaste this script into your shell and replace it with the output
Thu Jun 19 17:29:35 UTC 2025
radare2 5.9.9 34075 @ linux-x86-64
birth: git.5.9.8-756-gbe6a57690b 2025-06-19__16:48:34
commit: be6a57690b42ba4fa1b0906281288230904a7b8f
options: gpl -O? cs:5 cl:2 make
Linux x86_64
```

## Description

sometimes r- (truncating size of binary does not work) and when it works it will leak exponentially memory.
>25GB in few seconds.

## Test

With enough big file (> 36M):

run this:

r- 304 @@s:0x0800 0x022b97a0 2048

Beware, this will usually trigger OOM killer on Linux.

The command just removes 304 bytes and skips 2048 bytes, unitl it reaches end offset.

trufae:

**Description**

### DIFF
--- a/test/db/cmd/cmd_r
+++ b/test/db/cmd/cmd_r
@@ -60,8 +60,5 @@ EOF
 EXPECT=<<EOF
 ok
 EOF
-EXPECT_ERR=<<EOF
-WARN: Cannot resize below zero -304 208
-INFO: @@s: sequence interrupted
-EOF
+EXPECT_ERR=
 RUN


trufae:
this test is "broken" on purpose. waiting for the reporter to confirm behaviour: 
```

[XX] /home/runner/work/radare2/radare2/test/db/cmd/cmd_r r-infloop
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -NN -Qc 'r- 304 @@s:0x0800 0x022b97a0 2048
?e ok
' -
-- stderr
@@ -1,2 +0,0 @@
-WARN: Cannot resize below zero -304 208
-INFO: @@s: sequence interrupted

```

PR #24597, opened 2025-09-22
Clarify immediate arguments in snes/6502 disassembly

macmv:

**Description**

I've been using radare2 to look at 6502 assembly, and I noticed that `lda #0x00` and `lda 0x00` (load immediate, and load from address) look the same in r2. I've updated the syntax to show immediate arguments with a `#` prefix.

This also modifies snes disassembly, which I haven't used, so I'm not sure if this is the correct fix, but it definitely makes things clearer when looking at 6502 assembly.

### DIFF
--- a/libr/arch/p/snes/snesdis.c
+++ b/libr/arch/p/snes/snesdis.c
@@ -38,16 +38,16 @@ static int snesDisass(int M_flag, int X_flag, ut64 pc, RAnalOp *op, const ut8 *b
 		break;
 	case SNES_OP_IMM_M:
 		if (M_flag) {
-			op->mnemonic = r_str_newf ("%s 0x%02x", s_op->name, buf[1]);
+			op->mnemonic = r_str_newf ("%s #0x%02x", s_op->name, buf[1]);
 		} else {
-			op->mnemonic = r_str_newf ("%s 0x%04x", s_op->name, ut8p_bw (buf + 1));
+			op->mnemonic = r_str_newf ("%s #0x%04x", s_op->name, ut8p_bw (buf + 1));
 		}
 		break;
 	case SNES_OP_IMM_X:
 		if (X_flag) {
-			op->mnemonic = r_str_newf ("%s 0x%02x", s_op->name, buf[1]);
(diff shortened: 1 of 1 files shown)


trufae:
Can you take a look at the differences between snes, 6502 and 6502.cs? Maybe we can rename snes to 6502.snes?

https://github.com/radareorg/radare2/blob/master/libr/arch/p/6502/6502dis.inc.c

also maybe good to have an asmparse plugin for 6502 to make pseudo decompilation work, i can write the skeleton if u can review it and test it

macmv:
It looks like "6502.cs" doesn't interpret jump destinations, so it doesn't build the arrows on the left side in `pd`. "6502" does interpret these jump destinations, so it's what I've been using. I can't tell why "6502.cs" exists, it looks like just another author's implementation of the same thing (with less features).

It looks like the "snes" is a derivative of the WDC 65C816 chip, which is a 16-bit CPU based on the 6502. I don't know the instruction set for the 65C816, but it's definitely a derivative of the 6502 instruction set.

I don't know if it would be better to split out 6502 asm parsing. I guess it'd be a little like parsing x86 vs x86_64? I'm not entirely sure. I'd be happy to review and test the 6502 stuff, and I can look into the 65C816 (although that'll take more time).

There are definitely things like `bne` and `jmp` (branch and jump) instructions, where on the 65C816 there is a 24 bit address, and on the 6502 there is a 16 bit address. So I think the disassembled code would be a little cleaner if there were 6502-specific ways of disassembling those.

trufae:
after some thinking i believe its better to keep it like "snes" for simplicity, maybe the arch plugin description can bemore clear about being 65C816, aka 16bit 6502, but the register size and bus is different too. and im pretty sure there may be more bugs because i didnt really tested much this arch. the asm parsing plugin is there already i just pushed few improvements. but its not for snes. so that may require another separate one for it. check it out and feel free to report more bugs or send patches if you find anything.

trufae:
btw, if you have binaries for the testbins repo that will be very helpful too

PR #24623, opened 2025-09-26
Revert "Parse file and  flag on autocomplete ##shell"

jukuisma:

**Description**

This reverts commit 148784001349cc5038296919214f80d9f67784f6.

The aformentioned commit breaks autocompletion on fish:
```
$ r2 -
<tab>
.
.
.
| ?$?                     show available '$' variables and aliases
| ?@?                     misc help for '@' (seek), '~' (grep) (see ~?""?)
| ?>?                     output redirection
| ?|?                     help for '|' (pipe)
fish: Unknown command: \e
fish:

ile
^
fish: Unknown command: \e
fish:

lag
^
fish: Unknown command: \e
fish:

lag
^
[0x00000000]>
```
I'm not sure if simply reverting this messes things up in r2frida. Any better ideas are welcome as well. ping @trufae @as0ler

### DIFF
--- a/libr/core/core.c
+++ b/libr/core/core.c
@@ -1403,26 +1403,6 @@ R_API void r_core_autocomplete(RCore * R_NULLABLE core, RLineCompletion *complet
 			r_line_completion_clear (completion);
 			char *s = r_core_cmd_strf (core, "%s?", buf->data);
 			eprintf ("%s%s\n%s", core->cons->line->prompt, buf->data, s);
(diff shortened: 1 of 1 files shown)


trufae:
I guess the problem is the ansi escape codes but i would prefer to fix the code instead of removing it. Ill give it a try because i dont use fish and i dont see why it should be hitting the shell at all

trufae:
correct fix in https://github.com/radareorg/radare2/pull/24624

thanks for spotting it! that was a little awful :D

PR #24635, opened 2025-09-29
meson builds, tests, bpf & elf loader

quetz:

**Description**

- r2 built with meson now passes almost all tests (it used to miss some needed plugins)
- elf loader does not require special handling for bpf files
- some tests are fixed, others made more stable (like dwarf)
- import resolution in pe files stripped 4 chars off module names even if they did not end with .dll
- r2 and r2r now add R2_BINDIR to path so that !rasm2 from within r2 shell launches correct rasm2


libr/bin/format/pe/pe.c
```diff
@@ -547,8 +548,10 @@ static int bin_pe_parse_imports(RBinPEObj* pe,
 				// remove the trailling ".dll"
 				size_t len = strlen (symdllname);
 				r_str_case (symdllname, 0);
-				len = len < 4? 0: len - 4;
-				symdllname[len] = 0;
+				if (r_str_endswith(symdllname, ".dll")) {
```
trufae: Missing space before (

libr/bin/format/elf/elf.c
```diff
@@ -62,62 +62,6 @@ static inline bool is_elfclass64(Elf_(Ehdr) *h) {
 	return h->e_ident[EI_CLASS] == ELFCLASS64;
 }
 
-static bool elf_is_sbpf_by_hdr(const ELFOBJ *eo) {
```
trufae, line 65: Maybe @ulexec needs to check that before merging

libr/bin/format/pe/pe.c
quetz: fixed

libr/bin/format/elf/elf.c
ulexec: Some early versions of Solana programs are shipped with EM_BPF, but really they should be processed as EM_SBPF. It would be nice to set the right architecture automatically by using a similar heuristic than what this function implements. In any case, it's not consequential, as the architecture can be set manually. Same goes for the default baddr --which I did not know when I wrote these changes

trufae:
Changes look good. but ill see how to fix that sbpf thing in the bininfo to provide the right architecture and base address that will be a simpler and cleaner way to adjust

PR #24649, opened 2025-10-05
help: show core plugins in root ‘?’ dynamically; remove ‘Lc’ entry

### LINKED ISSUE #23724: Append Lc root commands help in ? help message
all those commands should be listed in the help message associated, maybe some are in the root, maybe others are in subcommands. that must be chosen by the plugin itself

```
[0x00000000]> Lc
a2f        The reworked analysis from scratch thing
agD        agD core plugin
java       Suite of java commands, java help for more info
prj        Experimental binary projects
sixref     quickly find xrefs in arm64 buffer
decai      r2 decompiler based on r2ai
sarif      Manage SARIF documents
r2jadx     jadx decompiler for radare2
goresym    rename symbols with goresym
codeshare  pull and run codeshare scripts via r2frida
[0x00000000]>
```

DivInstance:
## Summary
This PR updates the root ? help in Radare2 to include a dynamic “Core plugins” section and removes the Lc entry. Users can now discover core plugin commands directly from the main help without relying on Lc.

## Changes:
- Modified libr/core/cmd_help.inc.c : 
    - Removed Lc line from help_msg_root.
    - In cmd_help() for root ?, added a dynamic help block built from core->rcmd->plist :
        - Header: “Core plugins:”
        - Per plugin: <name> [ ? ] <desc>   

## Rationale:
- Listing core plugins directly in root help improves discoverability and keeps help consistent with the loaded plugin set.
- Avoids duplicating a static Lc entry when the plugin list is readily available programmatically.

## Testing:
- Build: ./configure && make -j
- Lint: sys/lint.sh

## Impact:
- Help output improvement only; no behavioral changes to command execution.
- Deprecates the Lc row in root ? in favor of the inline dynamic list.

Resolves #23724

### DIFF
--- a/libr/core/cmd_help.inc.c
+++ b/libr/core/cmd_help.inc.c
@@ -226,6 +226,7 @@ static RCoreHelpMessage help_msg_root = {
 	"k", "[?] [query]", "evaluate an sdb query",
 	"l", "[?] [filepattern]", "list files and directories",
 	"L", "[?] [-] [plugin]", "list, unload load r2 plugins",
+	"Lc", "[j]", "list core plugins",
 	"m", "[?]", "mount filesystems and inspect its contents",
 	"o", "[?] [file] ([addr])", "open file at optional address",
 	"p", "[?] [len]", "print current block with format and length",
--- a/libr/core/cmd_help.inc.c
+++ b/libr/core/cmd_help.inc.c
@@ -226,7 +226,6 @@ static RCoreHelpMessage help_msg_root = {
 	"k", "[?] [query]", "evaluate an sdb query",
 	"l", "[?] [filepattern]", "list files and directories",
 	"L", "[?] [-] [plugin]", "list, unload load r2 plugins",
-	"Lc", "[j]", "list core plugins",
 	"m", "[?]", "mount filesystems and inspect its contents",
 	"o", "[?] [file] ([addr])", "open file at optional address",
 	"p", "[?] [len]", "print current block with format and length",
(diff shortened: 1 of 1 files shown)


trufae:
the way to make core plugins visible shpuld be listing their help in the root help instead of exposing Lc in here imho. But we must find a fast way to make the commands created by core plugins available in their subcommands helps and thats the current blocker imho

DivInstance:
I agree with the direction. Instead of exposing Lc, I’ll surface core plugin help in the root ? and in the corresponding subcommand helps.

libr/core/cmd_help.inc.c
```diff
@@ -1722,6 +1721,40 @@ static int cmd_help(void *data, const char *input) {
 		// TODO #7967 help refactor
 		r_core_cmd_help (core, help_msg_intro);
 		r_core_cmd_help (core, help_msg_root);
+		/* Dynamically list core plugins in root help */
+		{
+			RListIter *iter;
+			RCorePlugin *pl;
+			int count = 0;
```
trufae, line 1728: just r_list_length() no need to iterate

libr/core/cmd_help.inc.c
```diff
@@ -1722,6 +1721,40 @@ static int cmd_help(void *data, const char *input) {
 		// TODO #7967 help refactor
 		r_core_cmd_help (core, help_msg_intro);
 		r_core_cmd_help (core, help_msg_root);
+		/* Dynamically list core plugins in root help */
```
trufae, line 1724: move all this into a separate function

trufae (changes requested):
the code assumes the plugin name is aligned with the commands provided in the plugin, which is not always correct. we probably need a way for core plugins to expose a help message or list of commands provided instead of relying on the name. also ideally those commands shuold be listed in the right help message. but the concern for this would be about performance when doing "?*"

PR #24666, opened 2025-10-07
Fix #24658 - Better conditional commands (?eq/ne/ge/le) ##shell

### LINKED ISSUE #24658: ?? clarify
During trainings its always confusing to have to explain that ??? is the help for ? because ?? was already taken, we can simplify the problem by performing the following breaking changes in the commands:

`??` -> `?eq`
`?!` -> `?ne`
`???` -> `??`

Additionally we can also change those:

`?+` -> `?ge`
`?-` -> `?le`

those are other related subcommands:

```
| ?$                               show value all the variables ($)
| ?+ [cmd]                         run cmd if $? > 0
| ?- [cmd]                         run cmd if $? < 0
| ?= eip-0x804800                  update $? return code with result of operation
```

trufae:
* Deprecate ??, ?!, ?+ and ?-

**Description**

### DIFF
--- a/libr/core/cmd_help.inc.c
+++ b/libr/core/cmd_help.inc.c
@@ -286,16 +286,13 @@ static RCoreHelpMessage help_msg_question_e = {
 
 static RCoreHelpMessage help_msg_question = {
 	"Usage: ?[?[?]] expression", "", "",
-	"?!", " [cmd]", "run cmd if $? == 0",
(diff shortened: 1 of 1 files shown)


trufae:
thats wip because there are a bunch of failing scripts that needs some testing, scr.prompt.code is helpful but i guess we need more tests to get this right

PR #24686, opened 2025-10-10
Fix #24680 - Increase the timeout to 1s in socket_slurp to avoid read timeout when download a PDB

### LINKED ISSUE #24680: Download PDB timeout is too short
## Description

When downloading PDBs from radare2, the function `socket_slurp` sets the socket timeout to 1000 microseconds (1ms) before reading the HTTP response. If the file data isn't received within this very short interval, recv will fail with `EAGAIN`, causing the download to abort. The original timeout was 1 second, which seems to work fine for most cases.

### Fix 1: Reset timeout in `socket_slurp`

Some functions are changing the timeout of the request, and sometimes resetting it. I'm not sure who should do that and when. I'm not sure why `socket_slurp` isn't resetting the timeout to 1s like the other functions, but resetting it to 1s in `socket_slurp` is fixing the issue as 1s is generally enough to receive part of the data. We could also remove the timeout change entirely if it's not necessary.

### Fix 2: Retry on `EAGAIN`

Another fix could be in `r_socket_read_block` where we could check `errno` to see if we have `EAGAIN` and retry the read conditionally on `r_socket_ready`, e.g

```
if (errno == EAGAIN && r_socket_ready (s, 1, 0) == 1) {
    continue;
}
```

### Fix 3: `R2_CURL`

Setting the environment variable `R2_CURL` to use curl instead of the built-in HTTP client.

Quentin01:
**Description**

This timeout wasn't reset to something else after the execution, everywhere else the timeout is set to 1s. If we don't reset it, it causes issues latter when reading the content of the PDB file as it could be longer than this initial timeout of 1000 microseconds.

### DIFF
--- a/libr/socket/socket_http.c
+++ b/libr/socket/socket_http.c
@@ -15,7 +15,7 @@ static size_t socket_slurp(RSocket *s, RBuffer *buf) {
 	if (r_socket_ready (s, 1, 0) != 1) {
 		return 0;
 	}
-	r_socket_block_time (s, true, 0, 1000);
+	r_socket_block_time (s, true, 1, 0);
 	for (i = 0; i < SOCKET_HTTP_MAX_HEADER_LENGTH; i += 1) {
 		ut8 c;
 		int olen = r_socket_read_block (s, &c, 1);


trufae:
thats not pdb-specific, but i guess we are good to merge this and revert if anything wrong happens somewhere else

PR #24687, opened 2025-10-11
Remove crosscompile for linux builds ##build

prodrigestivill:
- **Add cancel previous builds on the same branch on a new push.**

- **Removed crosscompile for debian arm64.**
  Works.

- **Removed crosscompile for static builds.**
  Fails with missing `asm/ptrace.h` include file, I was unable to find one for ubuntu arm64 that include `ARM_VFPREGS_SIZE`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,11 +67,11 @@ jobs:
       matrix:
         arch:
           - amd64
-          - i386
+#         - i386
         runner: [ubuntu-22.04]
-        include:
-          - arch: arm64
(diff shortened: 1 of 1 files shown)


trufae:
the static build is failing i think we are good to remove that because this toolchain looks broken if it doesn thave sys/ptrace.h

trufae:
Thanks! Yeah i have the same feeling with this toolchain. Maybe we can enable it later and try zig-cc but i dont think that plays well on android so meh. Lets focus on what works for now

thanks for other changes too!

Issue #24631, opened 2025-09-28
`dbt` does not display correctly


tm99hjkl:
## Environment

```sh
$ date
2025年  9月 28日 日曜日 17:09:46 JST
$ r2 -v
radare2 6.0.3 34109 @ linux-x86-64
birth: git.6.0.2-179-g76255f90b8 2025-09-24__20:55:00
commit: 76255f90b8aaa8acc31bef53654a9937327d297f
options: gpl -O? cs:5 cl:2 make
$ uname -ms
Linux x86_64

```

## Description

(This issue might be related to #22891)

First, I created the following `a.c`:

```c
int f() {
        int *p = 0;
        *p = 42;
        return 0;
}

int g() {
        return f();
}

int h() {
        return g();
}

int main() {
        return h();
}
```

I compiled this and then ran `dc; dbt` inside `r2 -d /tmp/a`. The resulting backtrace looked like this (the debugger plugin is `native`, and the backtrace algorithm is set to `fuzzy`):

```
[0x7e07e1486540]> e dbg.backend
native
[0x7e07e1486540]> e dbg.btalgo
fuzzy
[0x7e07e1486540]> dc; dbt
[+] SIGNAL 11 (aka SIGSEGV) errno=0 addr=0x00000000 code=1 si_pid=0 ret=0
0  0x63736089813d     sp: 0x0                 0    [??]  sym.f+20
```

On the other hand, with gdb I get the following backtrace:

```
(gdb) backtrace
#0  0x000055555555513d in f ()
#1  0x000055555555515c in g ()
#2  0x0000555555555170 in h ()
#3  0x0000555555555184 in main ()
```

After investigating, it seems that `r_debug_ptrace_func` at [line 75 of `libr/debug/p/native/bt.c`](https://github.com/radareorg/radare2/blob/252f88b738daefb65ed02f551a939557b3ef502e/libr/debug/p/native/bt.c#L75) is not working properly. As an experiment, I rewrote it as `list = cb (dbg, at)`, and that made it work correctly.

## Test

As mentioned above

trufae:
There are actually 7 tickets reporting the same problem

trufae:
Feel free to submit a pr and fix them all :)

tm99hjkl:
Thanks, I’ll take some time to look into it and see if I can help with a fix.

trufae:
Awesome!!

tm99hjkl:
[...]

PR #24692, opened 2025-10-13
##debug Fix fuzzy backtrace to show complete call stack with correct SP values

vicky-dx:

**Description**

This PR fixes issue #24631 where the fuzzy backtrace algorithm (`dbt` command with `dbg.btalgo=fuzzy`) was only showing 1 frame with `sp: 0x0` instead of the complete call stack.

**Changes made:**

1. **Fixed `prepend_current_pc()` in `bt.c`**: Now properly initializes `frame->sp` and `frame->bp` from the SP and BP registers instead of leaving them as 0.

2. **Improved `iscallret()` in `fuzzy_all.c`**: Enhanced CALL instruction detection by:
   - Scanning backwards up to 16 bytes to find CALL instructions
   - Supporting both direct CALL (`e8` opcode) and indirect CALL (`ff /2` opcode)
   - Better handling of variable-length CALL instructions on x86/x64

**Testing:**

Tested with the reproduction case from the issue:
- **Before**: `dbt` showed only 1 frame with `sp: 0x0`
- **After**: `dbt` shows complete backtrace matching gdb output with all 4 frames (f→g→h→main) and correct SP values

Backtrace output now matches gdb's behavior.

### DIFF
--- a/libr/debug/p/native/bt.c
+++ b/libr/debug/p/native/bt.c
@@ -19,6 +19,8 @@ static void prepend_current_pc(RDebug *dbg, RList *list) {
 		if (frame) {
 			frame->addr = addr;
 			frame->size = 0;
+			frame->sp = r_reg_getv (dbg->reg, "SP");
+			frame->bp = r_reg_getv (dbg->reg, "BP");
 			r_list_prepend (list, frame);
 		}
 	}
--- a/libr/debug/p/native/bt/fuzzy_all.c
+++ b/libr/debug/p/native/bt/fuzzy_all.c
@@ -2,34 +2,63 @@
 
 static int iscallret(RDebug *dbg, ut64 addr) {
 	ut8 buf[32];
-	if (addr == 0LL || addr == UT64_MAX)
+	RAnalOp op = {0};
+	int i;
+	if (addr == 0LL || addr == UT64_MAX) {
 		return 0;
+	}
 	/* check if region is executable */
 	/* check if previous instruction is a call */
-	/* if x86 expect CALL to be 5 byte length */
+	/* try looking back up to 16 bytes to find a call instruction */
 	if (dbg->arch && !strcmp (dbg->arch, "x86")) {
-		(void)dbg->iob.read_at (dbg->iob.io, addr-5, buf, 5);
-		if (buf[0] == 0xe8) {
-			return 1;
+		// On x86, try to find CALL instructions of various lengths
+		// Check for common patterns
+		int maxdist = (dbg->bits == 64) ? 16 : 8;
+		if (!dbg->iob.read_at (dbg->iob.io, addr - maxdist, buf, maxdist)) {
+			return 0;
 		}
-		if (buf[3] == 0xff  /* bits 4-5 (from right) of next byte must be 01 */
-				&& ((buf[4] & 0xf0) == 0xd0  /* Mod is 11 */
-					|| ((buf[4] & 0xf0) == 0x10  /* Mod is 00 */
-						&& (buf[4] & 0x06) != 0x04))) {  /* R/M not 10x */
-			return 1;
+		// Look backwards for CALL instructions
+		for (i = maxdist - 1; i >= 0; i--) {
+			// Direct CALL (e8 xx xx xx xx)
+			if (buf[i] == 0xe8 && (i + 5 == maxdist)) {
+				return 1;
+			}
+			// Indirect CALL (ff /2)
+			if (buf[i] == 0xff) {
+				int modrm = (i + 1 < maxdist) ? buf[i + 1] : 0;
+				int reg = (modrm >> 3) & 7;
+				if (reg == 2) { // CALL r/m
+					// Estimate instruction length based on ModR/M
+					int mod = (modrm >> 6) & 3;
+					int rm = modrm & 7;
+					int len = 2;
+					if (mod == 0 && rm == 5) {
+						len = 6; // disp32
+					} else if (mod == 1) {
+						len = 3; // disp8
+					} else if (mod == 2) {
+						len = 6; // disp32
+					}
+					if (i + len == maxdist) {
+						return 1;
+					}
+				}
+			}
 		}
-		// IMMAMISSINGANYOP
 	} else {
-		RAnalOp op;
-		(void) dbg->iob.read_at (dbg->iob.io, addr-8, buf, 8);
-		(void) r_anal_op (dbg->anal, &op, addr-8, buf, 8, R_ARCH_OP_MASK_BASIC);
-		if (op.type == R_ANAL_OP_TYPE_CALL || op.type == R_ANAL_OP_TYPE_UCALL) {
-			return 1;
+		// For non-x86, use anal to check
+		int maxdist = 16;
+		if (!dbg->iob.read_at (dbg->iob.io, addr - maxdist, buf, maxdist)) {
+			return 0;
 		}
-		/* delay slot */
-		(void) r_anal_op (dbg->anal, &op, addr-4, buf, 4, R_ARCH_OP_MASK_BASIC);
-		if (op.type == R_ANAL_OP_TYPE_CALL || op.type == R_ANAL_OP_TYPE_UCALL) {
-			return 1;
+		// Try different positions
+		for (i = maxdist - 1; i >= maxdist - 8 && i >= 0; i--) {
+			if (r_anal_op (dbg->anal, &op, addr - (maxdist - i), buf + i, maxdist - i, R_ARCH_OP_MASK_BASIC) > 0) {
+				if ((op.type == R_ANAL_OP_TYPE_CALL || op.type == R_ANAL_OP_TYPE_UCALL) &&
+				    (addr - (maxdist - i) + op.size == addr)) {
+					return 1;
+				}
+			}
 		}
 	}
 	return 0;


trufae:
Awesome! Good work

tm99hjkl:
This does not fundamentally resolve #24631. While the stack pointer value shown in the backtrace is indeed correct, the test for #24631 still does not produce the expected result.

As I mentioned in the comment on #24631, the root cause seems to be something related to ptrace_wrap rather than the backtrace algorithm itself.

vicky-dx:
> This does not fundamentally resolve #24631. While the stack pointer value shown in the backtrace is indeed correct, the test for #24631 still does not produce the expected result.
> 
> As I mentioned in the comment on #24631, the root cause seems to be something related to ptrace_wrap rather than the backtrace algorithm itself.


Thanks for the detailed feedback and investigation, @tm99hjkl!

You're absolutely right that there's a deeper ptrace_wrap issue that needs to be addressed. However, I'd like to clarify what my PR fixes vs. what remains:

What my PR fixes:


- The immediate user-facing issue: `dbt `now shows the complete call stack with correct SP values instead of just 1 frame with `sp: 0x0`

- This works correctly in practice (both with `[dbg.wrap=false]` which is the default, and with `[dbg.wrap=true] ` in my testing on WSL2)

### What remains to be investigated:


- The ptrace_wrap behavior you discovered where `ptrace `returns `ESRCH `and addresses become `0xffffffffffffffff`

- This might be environment-specific or test-framework-specific and may not affect interactive usage

I tested with `[dbg.wrap=true]` and the backtrace still works correctly, showing all frames with proper SP values. This suggests either:


1. The ptrace_wrap issue is specific to certain Linux configurations (not WSL2)

1. The SP/BP initialization fix I made resolves most manifestations of the issue

1. There's a timing/race condition that's hard to reproduce

### Questions:


- Are you seeing the actual backtrace fail with `[dbg.wrap=true]` in interactive use, or only in automated tests?

- What Linux distribution/kernel version are you testing on?

- Could you share more details about "the test for #24631" that doesn't work?

I'm happy to continue investigating the ptrace_wrap issue, but I believe my PR does resolve the primary symptom that users were experiencing.

tm99hjkl:
@vicky-dx

Thank you for your reply. I had assumed that everyone was in a similar situation to mine since I heard there were multiple issues related to `dbt` , but it seems that it actually works fine in some environments. In my case (Ubuntu24.04 / Linux 6.14.0-32-generic) , however, it appears that even `dc` fails to complete when `dbg.wrap=true` is set.

The test for #24631 simply means running the a.c file shown at the beginning of the issue under `r2 -d` , then executing `dc; dbt` .

vicky-dx:
Hi @tm99hjkl,

Thank you for the detailed feedback and investigation! You're absolutely right that there's a deeper `ptrace_wrap` issue on certain Linux configurations that my PR doesn't address.

I'm on WSL2 (Ubuntu 24.04 / Linux 6.6.87.2-microsoft-standard-WSL2), and interestingly, both `dbg.wrap=false` and `dbg.wrap=true` work fine for me - they both show the complete 4-frame backtrace with correct SP values. This suggests the ptrace_wrap ESRCH issue is specific to native Linux kernels (particularly 6.14+).

## Request for Diagnostic Information

To help investigate the root cause, could you please run this diagnostic script on your Ubuntu 24.04 / Linux 6.14.0-32 system and share the output?

```bash
#!/bin/bash
# Simple diagnostic for ptrace_wrap issue on Linux 6.14+

echo "=========================================="
echo "Simple Backtrace Diagnostic"
echo "=========================================="
echo ""

# Create test case
if [ ! -f "test_bt" ]; then
    cat > test_bt.c <<'EOF'
int f() { int *p = 0; *p = 42; return 0; }
int g() { return f(); }
int h() { return g(); }
int main() { return h(); }
EOF
    gcc -g -O0 test_bt.c -o test_bt
fi

echo "System: $(uname -r)"
echo "Radare2: $(r2 -v | head -1)"
echo ""

echo "=========================================="
echo "Test 1: With dbg.wrap=false (default)"
echo "=========================================="
timeout 5 r2 -q -d test_bt <<'EOF'
e dbg.wrap
dc
dbt
q
EOF

echo ""
echo "=========================================="
echo "Test 2: With dbg.wrap=true"
echo "=========================================="
timeout 5 r2 -q -d test_bt <<'EOF'
e dbg.wrap=true
dc
dbt
q
EOF

echo ""
echo "=========================================="
echo "Summary"
echo "=========================================="
echo "If Test 2 shows fewer frames or hangs, that confirms the issue."
echo "Expected: Both tests should show 4 frames (f, g, h, main)"

```

This will help us understand:

1. Whether dc completes on your system with dbg.wrap=true
2. How many frames dbt actually shows in each mode
3. If there are invalid addresses (0xffffffffffffffff) as you mentioned

I think we should Open a new issue specifically for "ptrace_wrap returns ESRCH on Linux 6.14+ causing backtrace failures"
there we will Investigate the ptrace_wrap issue separately with your diagnostic data

tm99hjkl:
@vicky-dx 

Thank you for your kind and thorough response. I ran the diagnostic script, and here are the results. I also agree that this seems to be an environment-specific issue:

```
==========================================
Simple Backtrace Diagnostic
==========================================

System: 6.14.0-32-generic
Radare2: radare2 6.0.5 34379 @ linux-x86-64

==========================================
Test 1: With dbg.wrap=false (default)
==========================================
false
[+] SIGNAL 11 (aka SIGSEGV) errno=0 addr=0x00000000 code=1 si_pid=0 ret=0
0  0x5fbd00be413d     sp: 0x7ffd6d2a59e0      0    [??]  rip sym.f+20

==========================================
Test 2: With dbg.wrap=true
==========================================
PTRACE_GETREGS: No such process
PTRACE_CONT: No such process

==========================================
Summary
==========================================
If Test 2 shows fewer frames or hangs, that confirms the issue.
Expected: Both tests should show 4 frames (f, g, h, main)
```

By the way, one point to clarify: in [this comment](https://github.com/radareorg/radare2/issues/24631#issuecomment-3351408750), the term “invalid address” refers not to `sp` itself, but to `*sp`.

In any case, I also agree that we should open a new issue for this matter.

If you can infer anything from the diagnostic results above, please let me know.

vicky-dx:
Hello @tm99hjkl 

see these are my output

```bash
==========================================
Simple Backtrace Diagnostic
==========================================

System: 6.6.87.2-microsoft-standard-WSL2
Radare2: radare2 6.0.5 34313 @ linux-x86-64

==========================================
Test 1: With dbg.wrap=false (default)
==========================================
false
[+] SIGNAL 11 (aka SIGSEGV) errno=0 addr=0x00000000 code=1 si_pid=0 ret=0
0  0x60d42058a13d     sp: 0x7ffe565463f0      0    [??]  rip sym.f+20
1  0x60d42058a15c     sp: 0x7ffe565463f8      8    [??]  sym.g+18
2  0x60d42058a170     sp: 0x7ffe56546408      16   [??]  sym.h+18
3  0x60d42058a184     sp: 0x7ffe56546418      16   [??]  main+18
4  0x75d8c0a2a1ca     sp: 0x7ffe56546428      16   [??]  section..bss-1605774918
5  0x75d8c0a2a28b     sp: 0x7ffe565464c8      160  [??]  section..bss-1605774725
6  0x60d42058a065     sp: 0x7ffe56546528      96   [??]  entry0+37

==========================================
Test 2: With dbg.wrap=true
==========================================
[+] SIGNAL 11 (aka SIGSEGV) errno=0 addr=0x00000000 code=1 si_pid=0 ret=0
0  0x562dae89913d     sp: 0x7ffc5339c850      0    [??]  rip sym.f+20
1  0x562dae89915c     sp: 0x7ffc5339c858      8    [??]  sym.g+18
2  0x562dae899170     sp: 0x7ffc5339c868      16   [??]  sym.h+18
3  0x562dae899184     sp: 0x7ffc5339c878      16   [??]  main+18
4  0x7a5ef0c2a1ca     sp: 0x7ffc5339c888      16   [??]  section..bss+1111024058
5  0x7a5ef0c2a28b     sp: 0x7ffc5339c928      160  [??]  section..bss+1111024251
6  0x562dae899065     sp: 0x7ffc5339c988      96   [??]  entry0+37

==========================================
Summary
==========================================
If Test 2 shows fewer frames or hangs, that confirms the issue.
Expected: Both tests should show 4 frames (f, g, h, main)
```

**_My Inferences:_**

1. Test 1 (dbg.wrap=false):

  - Debugging works - process caught SIGSEGV correctly
  - Backtrace incomplete - only shows 1 frame (sym.f) instead of expected 4 (f→g→h→main)
  -  SP value is valid: 0x7ffd6d2a59e0 (not 0x0)
  - Conclusion: My PR fixed the SP initialization, but fuzzy backtrace algorithm still can't traverse full stack

2. Test 2 (dbg.wrap=true):

  - Complete failure - PTRACE_GETREGS: No such process
  -  Can't even continue execution - PTRACE_CONT: No such process
  -  Conclusion: ptrace_wrap has a fundamental issue on Linux 6.14+ where the wrapped process becomes unreachable

3. Environment Comparison:

  -  WSL2 (works): Linux 6.6.87.2-microsoft-standard-WSL2
  -  Native Linux (fails): Linux 6.14.0-32-generic
  -  Hypothesis: Kernel 6.14+ may have changed ptrace behavior or seccomp/namespace handling

4. Root Cause (not sure though):
    
  -  ptrace_wrap uses a wrapper process to isolate ptrace operations
  -  On kernel 6.14+, the wrapper loses track of child process (ESRCH = "No such process")

Hello, @trufae what's your thought on that?

PR #24693, opened 2025-10-14
##arch Fix ARM64 variable substitution in address calculation instructions

### LINKED ISSUE #17637: Wrong variable names substitution?
### Work environment

| OS/arch/bits (mandatory)                             | Ubuntu 20.04.1
| Architecture/bits of the file (mandatory)            | arm
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.6.0-git 25649 @ linux-x86-64 git.4.4.0-678-g5ccf9fd48
commit: 5ccf9fd4824ee54e999e92993442ba200c65ab80 build: 2020-09-13__02:36:27

### Expected behavior
variable name in function is
; var int64_t var_0h_2 @ sp+0x20
so the substitution should be
add x29, sp - var_0h_2

### Actual behavior
variable name in function is
; var int64_t var_0h_2 @ sp+0x20
so the substitution should be
add x29, sp - var_0h_2**x20**

Disassemble a binary
Now

vicky-dx:
**Description**

Fixes incorrect variable substitution in ARM64 disassembly for address calculation instructions.

## Problem
Variable substitution was incorrectly applied to `add x29, sp, 0x20` instructions that set up frame pointers, resulting in garbled output like `add x29, var_0h_2x20`.

## Solution
Modified the `subvar()` function in `libr/arch/p/arm/pseudo.c` to skip variable substitution for three-operand `add`/`sub` instructions without memory access brackets. Variable substitution now only occurs for actual memory operations (instructions with `[]`).

## Example
**Before:**
```asm
sub sp, sp, 0x30
stp x29, x30, [sp, 0x20]
add x29, var_0h_2           Incorrect garbled output
```
**After:**
```asm
sub sp, sp, 0x30
stp x29, x30, [var_20h]     Correct (memory access)
add x29, sp, 0x20           Correct (no substitution)
```
**Testing:**

All 669 ARM64 assembly tests pass (r2r test/db/asm/arm_64)
All 23 variable feature tests pass (r2r test/db/cmd/feat_variables)
Behavior now aligns with IDA Pro

`Fix #17637`

### DIFF
--- a/test/db/cmd/cmd_pseudo_arm64
+++ b/test/db/cmd/cmd_pseudo_arm64
@@ -249,3 +249,33 @@ EXPECT=<<EOF
             0x00000000      4218631e       d2 /= d3
 EOF
 RUN
+
+NAME=ARM64 no variable substitution in frame pointer setup (add x29, sp, 0x20)
+FILE=malloc://4
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=64
+e asm.pseudo=true
+wx fd830091
+pd 1
+EOF
+EXPECT=<<EOF
+            0x00000000      fd830091       x29 = sp + 0x20
+EOF
+RUN
+
+NAME=ARM64 variable substitution with frame pointer setup
+FILE=malloc://12
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=64
+e asm.pseudo=true
+wx ffc300d1fd7b02a9fd830091
+af
+pd 3~x29
+EOF
+EXPECT=<<EOF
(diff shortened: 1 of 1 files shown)


trufae:
Can you add a test? Also, the ##tags must be at the end of the commit msg.

thanks for fixing it!

trufae:
why did you closed and deleted the pr?

vicky-dx:
i am sorry i was just cleaning some branches, while deleting i must have overlooked it.

trufae:
thanks!

PR #24694, opened 2025-10-14
Fix x0 register write bug in r_reg_next_diff ##debug

### LINKED ISSUE #17637: Wrong variable names substitution?

### LINKED ISSUE #18975: Cannot set X0 register while debugging AARCH64
## Environment

```
Thu Aug 12 10:55:14 AM CEST 2021
radare2 5.4.0-git 26622 @ linux-x86-64 git.5.4.0-git
commit: 078714d0b22c6cc2bfc66e3cb18439842a2046e4 build: 2021-08-12__10:29:54
Linux x86_64
```

## Description
When debugging a AARCH64 binary the register x0 cannot be modified.

The attached [ldp.zip](https://github.com/radareorg/radare2/files/6974122/ldp.zip) file contains a compiled binary for this architecture which can be used to launch a qemu to reproduce the bug:

`qemu-aarch64-static -singlestep -g 1234  binary`

On another shell, launch radare2 with the desired backend (gdb-multiarch or aarch64-linux-gnu-gdb)

`r2 -e asm.arch=arm -e asm.bits=64 -e bin.cache=true  -D gdb-multiarch gdb://127.0.0.1:1234`

Then attempt to modify the value of the `x0` register the register is not updated 

```
[0x00400078]> dr x0
0x00000000
[0x00400078]> dr x0=1
0x00000000 ->0x00000001
[0x00400078]> dr x0
0x00000000
```

If the same is performed on `x1` register, the register gets updated

```
[0x00400078]> dr x1
0x00000000
[0x00400078]> dr x1=1
0x00000000 ->0x00000001
[0x00400078]> dr x1
0x00000001
```

By capturing the traffic on the loopback interface it is possible to confirm that the command is not even being sent to the debugger when attempting to set the `x0` register. (see [wireshark.zip](https://github.com/radareorg/radare2/files/6974230/wireshark.zip))

[ldp.zip](https://github.com/radareorg/radare2/files/6974122/ldp.zip)
[wireshark.zip](https://github.com/radareorg/radare2/files/6974230/wireshark.zip)

vicky-dx:
Fix #18975, Fix #17637

This PR fixes two related issues:

1. **#18975** - Registers at offset 0 (like ARM64 x0) were being skipped during register write operations
2. **#17637** - Wrong variable name substitution in ARM64 frame pointer setup instructions

## Problem 1: x0 register write bug (#18975)

When debugging ARM64 binaries via GDB remote protocol, setting the x0 register with `dr x0=1` would not work, while other registers (x1, x2, etc.) worked fine.

### Root Cause
The `r_reg_next_diff()` function in `libr/reg/reg.c` was using `offset > prev_offset` to iterate through registers. When `prev_ri` is NULL (first iteration), `prev_offset` is 0. Since x0 is also at offset 0, the condition `0 > 0` would evaluate to false, causing x0 to be skipped.

### Solution
Changed the condition from `>` to `>=` to include registers at offset 0.

## Problem 2: ARM64 variable substitution (#17637)

Instructions like `add x29, sp, 0x20` (frame pointer setup) were incorrectly having variables substituted, showing `add x29, var_0h_2` instead of keeping the original instruction.

### Root Cause
The pseudo disassembly code in `libr/arch/p/arm/pseudo.c` was substituting variables in all address calculation instructions, including frame pointer setup which should be left unchanged.

### Solution
- Added `is_frame_pointer_setup()` helper function using `r_str_casestr()` for case-insensitive detection
- Skip variable substitution for `add`/`sub` instructions that setup frame pointers
- Pseudo-ize `add x29, sp, 0` to `mov x29, sp` for cleaner output
- Added test case in `test/db/cmd/feat_variables`

## Testing

- ✅ Verified x0 register can now be written in debug mode
- ✅ Verified frame pointer instructions no longer get variable substitution
- ✅ All existing tests pass
- ✅ New test added for ARM64 frame pointer behavior

### DIFF
--- a/test/db/formats/elf/thumb
+++ b/test/db/formats/elf/thumb
@@ -117,7 +117,7 @@ EXPECT=<<EOF
 push {r7, lr}
 add r7, sp, 0
 push {r7, lr}
-add r7, var_0h
+add r7, sp, 0
 EOF
 RUN
 


libr/arch/p/arm/pseudo.c
```diff
@@ -563,6 +563,24 @@ static char *subvar(RAsmPluginSession *s, RAnalFunction *f, ut64 addr, int oplen
...
+	// Don't substitute variables in address calculation instructions
+	// like "add x29, sp, 0x20" which sets up frame pointers
+	if (tstr) {
+		const char *comma = strchr (tstr, ',');
+		if (comma) {
+			comma = strchr (comma + 1, ',');
+			if (comma && !strchr (tstr, '[')) {
+				// This is a three-operand instruction without memory access brackets
+				// Check if it's an add/sub instruction
+				if (strstr (tstr, "add") == tstr || strstr (tstr, "sub") == tstr ||
```
trufae: move these strstrs into a separate function to make this conditional more readable. the comments can be simplified or even removed if the function name is clear enough. you can probably tolower tstr inside this function and strstr once

trufae:
Much cleaner now! If you can provide a test it will be great, otherwise im happy to merge it as is

vicky-dx:
> Much cleaner now! If you can provide a test it will be great, otherwise im happy to merge it as is

ok sure, working on it

trufae:
run r2r -i on the broken tests. i think new output is better, because [sp+0] is the return address

trufae:
please git rebase and update the tests

trufae:
oops. please rebase again

trufae:
the tests hasnt been updated :? run 'r2r -i ' on the failing ones

trufae:
Just one left: "r2r -i db/formats/elf/thumb "

vicky-dx:
> Just one left: "r2r -i db/formats/elf/thumb "
ok working on it,

trufae:
r2r -i db/formats/elf/thumb

trufae:
r2r -i  db/formats/elf/random

trufae:
this pr is fixing two issues. it will be good that when you get all tests pass, to squash all the changes in two separate commits. if you like i can do that. otherwise i can wait. but the changes look good to me

vicky-dx:
> this pr is fixing two issues. it will be good that when you get all tests pass, to squash all the changes in two separate commits. if you like i can do that. otherwise i can wait. but the changes look good to me

please can you do it for me, everything looks good on my PC, all test

trufae:
fixed the tests , squashed changes and merged into master by hand

Issue #23096, opened 2024-07-06
`axff` didnt work as expected


ucgJhe:
## Environment

```sh
Sun Jul  7 02:59:40 CST 2024

radare2 5.9.3 32339 @ linux-x86-64
birth: git.5.7.8-3398-g3382a3e231 2024-07-07__02:43:10
commit: 3382a3e231f6a5bbf21facd8fdb2d5377354bdc1
options: gpl -O? cs:5 cl:2 make

Linux x86_64
```

## Description

I want to list all call refs from a function with `axff`, but it seems not work as expected.

First, open attached file with `r2 -A -e bin.cache=true ncc`, and thats said I want to list call refs from the function `sym.doEventInternal`, so I use `axff @ sym.doEventInternal~CALL`.

But I found that there is no `sym.readInfo_log` on the list, althought it actually existed in the code.

we can make sure by using `aae @ sym.doEventInternal ;axl~sym.readInfo_log~sym.doEventInternal~CALL`

This will take some time but will give the result as the picture below:

![image](https://github.com/radareorg/radare2/assets/10076774/6c524962-13a5-4aaf-8913-82fff721f7ca)

## Test

```sh
axff @ sym.doEventInternal~sym.readInfo_log
```
[ncc.zip](https://github.com/user-attachments/files/16117065/ncc.zip)

ucgJhe:
@radare hi, i couldn't repro this with any random bin from testbin, so I just attached the file here.

trufae:
The bin is now in the testbins and i checking for the best way to fix that bug now. thanks for reporting

trufae:
Some comments here:

* the `@` modifier does nothing on aa subcommands
* the reason why aae find those xrefs and aaa does not is because of function boundaries not handled properly
* the boundaries in this function are wrong because jump tables are not implemented for mips (working on it rn) so the coverage is incorrect, and therefor the reference emulation fails

trufae:
partial fix of the problem is here https://github.com/radareorg/radare2/pull/23099

trufae:
Can you try again with latest r2 from git?

ucgJhe:
@trufae Thx for your quick response, after test it again i think the issue still exist.

The command `axff @ sym.doEventInternal~CALL` now shows  function `sym.readInfo_log` but i found there are more missing functions that should be in the list like `sym.runProbe`, `sym.imp.strcpy`, `sym.unlockInfo_log` etc.

[...]

PR #24695, opened 2025-10-14
Fix duplicate xrefs in axff command ##anal ##xrefs

vicky-dx:
**Description**

This PR partially fixes issue #23096 by addressing the duplicate xrefs problem in the `axff` command.

### Problem

Issue #23096 reported two separate problems:
1. **Duplicate xrefs** appearing in `axff` output (showing the same function multiple times) 
2. **Missing xrefs** due to MIPS jump table issues (some functions not appearing at all)

This PR fixes **problem #1** (duplicates). Problem #2 is a separate issue related to MIPS architecture support.

### Root Cause

The `fcn_get_all_refs()` function in `libr/anal/xrefs.c` collects cross-references by iterating through all basic blocks in a function and all instructions within each block. For each instruction, it calls `collect_refs()` and appends the results to a vector.

When the same reference appears from multiple instructions (which can happen in valid code), the reference would be added multiple times to the result vector, causing duplicates in the output.

### Solution

After sorting the collected references (which was already being done), I added a deduplication pass that:
1. Walks through the sorted array with two pointers (read and write)
2. Only copies elements that differ from the previous element (using the existing `compare_ref()` function)
3. Adjusts the vector's `_end` pointer to truncate the duplicates

This is an efficient O(n) deduplication algorithm after the O(n log n) sort, and it correctly handles the case where the same xref legitimately appears from different source addresses.

### Testing

**Verified with the original test case:**
- Downloaded the `ncc` MIPS binary from issue #23096
- Ran `axff @ sym.doEventInternal~CALL`
- Result: No duplicate CALL references (each function appears exactly once)
- Previously missing `sym.readInfo_log` is now shown without duplicates

**Regression testing:**
- All 5 existing xref tests pass: `r2r test/db/cmd/cmd_ax`
- No test failures introduced

### Example Output

**Before fix:** Same function would appear multiple times

> CALL 0x004978bc 0x00523760 sym.readInfo_log
> CALL 0x004978bc 0x00523760 sym.readInfo_log <-- duplicate

**After fix:** Each function appears once

> CALL 0x004978bc 0x00523760 sym.readInfo_log
> CALL 0x004978d8 0x00523820 sym.writeInfo_log
> CALL 0x004978fc 0x005654dc sym.runSysCall
> ...

### Note

This PR does **not** address the missing xrefs issue mentioned in #23096 (functions like `sym.runProbe`, `sym.imp.strcpy`, `sym.unlockInfo_log` still not appearing). That requires fixing MIPS jump table analysis and function boundary detection, which is beyond the scope of this PR.

### DIFF
--- a/libr/anal/xrefs.c
+++ b/libr/anal/xrefs.c
@@ -541,6 +541,27 @@ static RVecAnalRef *fcn_get_all_refs(RAnalFunction *fcn, RefManager *rm, Collect
 		}
 
 		RVecAnalRef_sort (anal_refs, compare_ref);
+
+		// Remove duplicates after sorting
+		if (!RVecAnalRef_empty (anal_refs)) {
+			RAnalRef *write_ptr = anal_refs->_start;
+			RAnalRef *read_ptr = anal_refs->_start + 1;
+			RAnalRef *end_ptr = anal_refs->_end;
+
+			while (read_ptr < end_ptr) {
+				// Only keep if different from previous
+				if (compare_ref (write_ptr, read_ptr) != 0) {
+					write_ptr++;
+					if (write_ptr != read_ptr) {
+						*write_ptr = *read_ptr;
+					}
+				}
+				read_ptr++;
+			}
+
+			// Truncate by adjusting end pointer
+			anal_refs->_end = write_ptr + 1;
+		}
 	}
 
 	return anal_refs;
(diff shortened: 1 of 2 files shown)


trufae:
Looks good! yeah ideally it will be good to find the root cause of the missing xrefs, but getting some more issues spotted here is the way to go. thanks for taking care

PR #24701, opened 2025-10-14
Fix ARM MOVW+MOVT instruction xref detection #17647

### LINKED ISSUE #17647: Xrefs not found for armcompiler6 (armclang)
I have been using this tool for analyzing the machine code of binaries built with arm-gcc for some time. Now I need to perform the same kind of analysis on binaries produced with Keil armcompiler6. For some reason I cannot find the xrefs on these binaries. After some experimentation I am able to find xrefs when linking "flat" (without a linker script). When running the same linker script through the binary built with arm-gcc I am able to find the xrefs.

### Work environment

| OS/arch/bits (mandatory)                             | arm 32
| File format of the file you reverse (mandatory)      | axf, elf
| Architecture/bits of the file (mandatory)            | arm32
| r2 -v full output, **not truncated** (mandatory)     | radare2 4.6.0-git 25246 @ linux-x86-64 git.4.4.0-687-g67cc89877
commit: 67cc898770d23abfafd47b633f623eab429f02f7 build: 2020-09-14__08:19:00

### Expected behavior
Load binary, issue aaa, s obj.find_me, axt and list the xrefs.

### Actual behavior
No xrefs found (except for the binary built without a .sct linker script).

I have attached three binaries;

- armgcc_with_ld.elf which works and is linked using the same memory map as the one that doesnt work (cc6_with_sct.axf)
- cc6_with_sct.axf Binary not working as expected
- cc6_without_sct.axf Binary which works but does not have the correct memory map.

Load the binary cc6_with_sct.axf, aaa, s obj.find_me , axt.

I would be glad to help out with the code change myself if I can get some initial hints on where to look. Thanks in advance

vicky-dx:
### Description:

**_Problem_**
ARM Compiler 6 (armclang) uses MOVW+MOVT instruction pairs to load 32-bit addresses in Thumb mode, but radare2 was not detecting these as pointer references. This prevented xref (cross-reference) creation even when the target memory was properly mapped.

### Example from the issue:
```
movw r0, #0xc000
movt r0, #0x1fff    ; Combined: r0 = 0x1fffc000
```

The `aaa; axt obj.find_me` command showed no xrefs to the `obj.find_me` variable at `0x1fffc000.`

**_Root Cause_**
Two distinct issues were identified:

1. **radare2 bug (FIXED in this PR):** The ARM analysis plugin wasn't tracking `MOVW+MOVT` pairs to set `op->ptr`, which is required for xref creation.

1. **User's binary issue (separate):** The specific binary `cc6_with_sct.axf` has a linker script configuration problem where `REGION_DATA `exists as a section but is not included in any ELF LOAD segment, causing radare2 to not map it in memory.

**_Solution_**
Implemented hash table tracking in ` plugin_cs.c`

- Added HtUU *ht_movw field to PluginData for per-register tracking
- MOVW stores lower 16 bits indexed by register number
- MOVT retrieves stored value and combines: op->ptr = (upper << 16) | lower
- Handles Capstone Thumb mode quirk where MOVW/MOVT decode as ARM_INS_MOV
- Uses mnemonic string check with r_str_startswith() for detection
- Uses ht_uu_update() to properly handle register overwrites

**_Verification_**
With our fix applied:

```
$ r2 -qc 's 0x080002c4; ao 2~ptr' cc6_with_sct.axf
ptr: 0x0000c000
ptr: 0x1fffc000   Correctly computed!
```

Xrefs work when memory is mapped:

```
$ r2 -qc 'om 3 0x1fffc000 0x4 0x3a8 rw-; aaa; axt obj.find_me' cc6_with_sct.axf
# Shows 13 xrefs from 4 different functions 
```

**_Comparison with working binaries:_**

`cc6_without_sct.axf`: 19 xrefs found (same compiler, proper linker config) 
`armgcc_with_ld.elf`: 18 xrefs found (uses LDR instruction) 

**_Test Coverage_**

Added 16 comprehensive test cases to `arm`:

- Basic ptr value computation: 0x00000000, 0x1fffc000, 0x08001234
- Different registers: r0, r1, r7, lr (r14)
- Register overwrite scenarios
- Standalone MOVW/MOVT without pairs
- Edge cases: max value 0xffffffff, zero values
- Opcode and val field verification

**_Test Results:_** 
All 77 functional tests passing, 0 regressions 

Files Changed
`plugin_cs.c:` +69 lines (MOVW+MOVT tracking logic)
`arm:` +238 lines (16 new test cases)

Closes
#17647

### DIFF
--- a/test/db/anal/arm
+++ b/test/db/anal/arm
@@ -1262,3 +1262,241 @@ EOF
 EXPECT=<<EOF
 EOF
 RUN
+
+
+NAME=arm movw+movt ptr value simple
+FILE=malloc://32
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=16
+wx 40f20000 c0f20000
+s 0
+ao 2~ptr
+EOF
+EXPECT=<<EOF
+ptr: 0x00000000
+ptr: 0x00000000
+EOF
+RUN
+
+NAME=arm movw+movt ptr value 0x1fffc000
+FILE=malloc://32
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=16
+wx 4cf20000 c1f6ff70
+s 0
+ao 2~ptr
+EOF
+EXPECT=<<EOF
+ptr: 0x0000c000
+ptr: 0x1fffc000
+EOF
+RUN
+
+NAME=arm movw+movt ptr value 0x08001234
+FILE=malloc://32
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=16
+wx 41f23420 c0f60000
+s 0
+ao 2~ptr
+EOF
+EXPECT=<<EOF
+ptr: 0x00001234
+ptr: 0x08001234
+EOF
+RUN
+
+NAME=arm movw+movt different registers no combine
+FILE=malloc://32
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=16
+wx 4cf20001 c1f6ff70
+s 0
+ao 2~ptr
+EOF
+EXPECT=<<EOF
+ptr: 0x0000c000
+ptr: 0x1fff0000
+EOF
+RUN
+
+NAME=arm movw+movt to r1
+FILE=malloc://32
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=16
+wx 4cf20001 c1f6ff71
+s 0
+ao 2~ptr
+EOF
+EXPECT=<<EOF
+ptr: 0x0000c000
+ptr: 0x1fffc000
+EOF
+RUN
+
+NAME=arm movw+movt to r7
+FILE=malloc://32
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=16
+wx 4af6cd37 c1f23427
+s 0
+ao 2~ptr
+EOF
+EXPECT=<<EOF
+ptr: 0x0000abcd
+ptr: 0x1234abcd
+EOF
+RUN
+
+NAME=arm movw+movt to lr
+FILE=malloc://32
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=16
+wx 45f2786e c9f6bc2e
+s 0
+ao 2~ptr
+EOF
+EXPECT=<<EOF
+ptr: 0x00005678
+ptr: 0x9abc5678
+EOF
+RUN
+
+NAME=arm movw standalone without movt
+FILE=malloc://32
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=16
+wx 4cf20000 0000
+s 0
+ao 1~ptr
+EOF
+EXPECT=<<EOF
+ptr: 0x0000c000
+EOF
+RUN
+
+NAME=arm movt standalone without movw
+FILE=malloc://32
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=16
+wx c1f6ff70
+s 0
+ao 1~ptr
+EOF
+EXPECT=<<EOF
+ptr: 0x1fff0000
+EOF
+RUN
+
+NAME=arm movw+movt register overwrite
+FILE=malloc://32
+CMDS=<<EOF
+e asm.arch=arm
+e asm.bits=16
+wx 41f21110 42f22220 c3f23330
+s 0
+ao 3~ptr
+EOF
+EXPECT=<<EOF
+ptr: 0x00001111
+ptr: 0x00002222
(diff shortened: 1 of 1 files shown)


trufae:
these xrefs should be detected with esil. aae must be able to handle them

vicky-dx:
> these xrefs should be detected with esil. aae must be able to handle them


Hi @trufae, thanks for the feedback!

You're absolutely right that ESIL-based analysis is the more robust approach. I tested both methods and here are my findings:

### Current Situation

**With aae (ESIL emulation):**
```
$ r2 -qc 'om 3 0x1fffc000 0x4 0x3a8 rw-; aae; axt obj.find_me' cc6_with_sct.axf
# Successfully finds 10 xrefs through ESIL emulation 

```
**With aaa (standard analysis):**

- Before our PR: 0 xrefs (MOVW+MOVT not tracked)
- After our PR: 9 xrefs (using `op->ptr` static analysis)

_Key Observations_
The ESIL for MOVW+MOVT is already correct:

- MOVW: `49152,r0,= → r0 = 0xc000`
- MOVT: `16,8191,<<,r0,|= → r0 |= (0x1fff << 16)`

So `aae `can detect these xrefs through emulation, which is indeed the better architectural approach.

**_I have some questions:_** 

- Should we remove our op->ptr tracking and document that users should use aae instead of aaa for ARM binaries with MOVW+MOVT patterns?
- Or should we keep both approaches as complementary?
  - `op->ptr `for quick static analysis with `aa/aaa`
  - ESIL emulation for thorough dynamic analysis with `aae`
- Does `aaa `call `aae `internally? If not, should it for ARM architecture, or should users explicitly run `aae `for better xref detection?

**_My Understanding_**
I think you're suggesting that setting `op->ptr` in the analysis plugin goes against radare2's design principle where ESIL emulation handles dynamic value tracking. If that's the case, I'm happy to:

- Remove the hash table tracking code
- Update the PR description to document that aae should be used
- Maybe add a note in the issue about the recommended workflow
- Please let me know your preferred approach, and I'll update the PR accordingly!

Thanks again for the review! 🙏

trufae:
I think its doable to have non-emulation based xrefs analysis for performance reasons. The problem is that this approach, which has been also used in intel plugins in the past introduces non-O1 disassembler and adds context information to the decoder which is not always true because we can be swapping instructions in the same memory address and this hashtable will need to be invalidated and that causes an extra load. imho is better to move this stuff into the analysis library or plugin rather than the arch/decoder infrastructure to avoid all the problems mentioned above.

I am working on defining an infrastructure of analysis plugins that can be enabled, disabled and structured automatically depending on the type of binary and architecture selected. this approach will permit us to move cases like this into a location were O1 costs dont matter and its more flexible and tuneable. the ->ptr tracking thing as you mention is uncertain on arm because most pointers in that architecture are constructed with multiple instructions, and the movk can be used multiple times over the base movz to construct other pointers, and the only sane way to handle all those is by using emulation.

if you have a specific usecase to apply this approach without emulation or a problem (performance in analysis) it will be good to use it as a base for planing the right approach without esil. maybe aar shuold behave different for arm targets ignoring the ->ptr or at least verifying the destination twice to avoid false positive refs

vicky-dx:
Hi @trufae,

First off Sorry for late response 
Thank you for the detailed explanation! I really appreciate you taking the time to explain the architectural concerns and the reasoning behind keeping the decoder layer O(1) and stateless.

You're absolutely right - I now understand why adding state (hash table) to the decoder is problematic:

1.  Non-O(1) disassembly with hash table lookups
2.  Need to track when instructions change
3. Won't handle ARM64's more complex MOVZ/MOVK sequences (3-4 instructions)
4. Mixing decoder concerns with analysis logic

am i correct with above points?

## My Understanding so far (please correct if i am wrong anywhere):

After your feedback, I see the proper separation should be:

- **Decoder (`plugin_cs.c`)**: Simple, stateless, O(1) - just decode instructions
- **Analysis plugin**: Complex, stateful, multi-instruction tracking - handles xrefs
- **ESIL**: Already works correctly for these patterns via emulation

The ESIL approach is indeed more robust:
```bash
$ r2 -qc 'om 3 0x1fffc000 0x4 0x3a8 rw-; aae; axt obj.find_me' cc6_with_sct.axf
# Finds 10 xrefs correctly through emulation
```



---

I think i should close this PR (acknowledge it's the wrong architectural approach)
and keep the test cases in a comment (they might be useful for future analysis plugin work)

Thank you for the mentorship on radare2's architecture! This helps me understand the codebase design principles better. 🙏

trufae:
Correct

PR #24724, opened 2025-10-21
Fix p8fm mask size mismatch for instructions longer than 8 bytes #24712

### LINKED ISSUE #24712: Occasional size mismatch between function bytes and mask in `p8fm` output
## Environment
```
Fri Oct 17 22:51:31 2025

radare2 6.0.4 0 @ linux-x86-64
birth: git.6.0.4 2025-09-30__15:04:01
commit: 6.0.4
options: gpl release -O? cs:5 cl:2 make

x86
```

## Description

There is an issue with the `p8fm` command where the size of the function's byte mask doesn't match the size of the raw bytes. Often it's just like 1/2 bytes of difference, but sometimes it can be more. This doesn't happen consistently with all functions, but it's still relatively frequent.

The way I would work around this issue downstream is the following:
1. Scan blocks with `abm` until I hit that mask difference, then
2. Scan instructions with `aoj`, and 
3. When the "offending" instruction is located, zero-out the corresponding mask and finally 
4. Merge what was reconstructed with the original mask from p8fm

## Test

[vlc_3-0-20.zip](https://github.com/user-attachments/files/22980240/vlc_3-0-20.zip)

I'm attaching a VLC binary to reproduce this issue. The commands I ran were the following:
```
> r2 -e bin.cache=true /path/to/vlc_3-0-20
 -- This computer has gone to sleep.
[0x000018e0]> aaa
INFO: Analyze all flags starting with sym. and entry0 (aa)
INFO: Analyze imports (af@@@i)
INFO: Analyze entrypoint (af@ entry0)
INFO: Analyze symbols (af@@@s)
INFO: Analyze all functions arguments/locals (afva@@@F)
INFO: Analyze function calls (aac)
INFO: Analyze len bytes of instructions for references (aar)
INFO: Finding and parsing C++ vtables (avrr)
INFO: Analyzing methods (af @@ method.*)
INFO: Recovering local variables (afva@@@F)
INFO: Type matching analysis for all functions (aaft)
INFO: Propagate noreturn information (aanr)
INFO: Use -AA or aaaa to perform additional experimental analysis
[0x000018e0]> p8fm @ 6752
41554989cd41544989d455534889fb4881ecd80000004c894424404c894c244884c074370f294424500f294c24600f295424700f299c24800000000f29a424900000000f29ac24a00000000f29b424b00000000f29bc24c000000064488b042528000000488944241831c0488b2d1e250000488d542420488d842400010000c7042420000000488b7d00c74424043000000048894424084889542410e8cff8ffff8b0b83f9047627488b7d00e8cff8ffff488b44241864482b04252800000075704881c4d80000005b5d415c415dc383c101488b7d004d89e0be01000000890b488d15e3040000488d0dd404000031c0e83bf7ffff488b7d004889e14c89eabe01000000e8b7f7ffff488b4d00ba02000000be01000000488d3dbc040000e88df8ffff488b7d00e8c4f8ffffe977ffffffe81af8ffff:731e4b1f00731e4b1f00161f4f1f006f3b1c000000000f2f0000000f2f0000007d00f700627700000062770000006277000000627700000000000062770000000000006277000000000000627700000000000062770000000000001b0f0e00000000000f2f0000005f00cf1f00000000000fbf0000000fff0000000000003f0000000000004f1f00003f000000000000000f2f0000000f2f000000bf000000001f003b1c00f5004f1f0000bf000000000f3f0000001b0f010000000000ff006f3a1c0000000017177b177b1fdf3a1c004f1f00004f1f000f000000003f00cfbf0000000000cfbf00000000005f00bf000000004f1f00004f1f004f1f000f00000000bf000000004f1f00000b000000000f00000000cf3f0000000000bf000000004f1f0000bf00000000bf00000000bf00000000
```

In the above example, notice that the mask (308 bytes) is smaller than the function (310 bytes).

vicky-dx:
Fixes #24712

The aobm command was capping instruction mask generation at 8 bytes, causing p8fm to produce masks shorter than the actual function bytes when functions contained instructions longer than 8 bytes.

Changes:
- Increase last from R_MIN(8, analop.size) to R_MIN(sizeof(buf), analop.size)
- Increase finalmask buffer from 8 to 32 bytes to match input buf size
- Remove redundant '&& i < last' check in final loop

Test case: VLC binary function at 0x1a60 previously showed 310 bytes but only 308 mask bytes. Now correctly shows 310:310 match.

### DIFF
--- a/libr/core/cmd_anal.inc.c
+++ b/libr/core/cmd_anal.inc.c
@@ -9561,7 +9561,7 @@ static void cmd_anal_opcode_bits(RCore *core, const char *arg, int mode) {
 	r_anal_op_init (&analop);
 	r_anal_op_set_bytes (&analop, core->addr, buf, sizeof (ut64));
 	(void)r_anal_op (core->anal, &analop, core->addr, buf, sizeof (buf), R_ARCH_OP_MASK_DISASM);
-	int last = R_MIN (8, analop.size);
+	int last = R_MIN (sizeof (buf), analop.size);
 	PJ *pj = (mode == 'j')? r_core_pj_new (core): NULL;
 	if (last < 1) {
 		return;
@@ -9668,7 +9668,7 @@ static void cmd_anal_opcode_bits(RCore *core, const char *arg, int mode) {
 			int pi = 0;
 			char *s = r_strbuf_drain (sb);
 			char *p = s;
-			ut8 finalmask[8] = {0};
+			ut8 finalmask[32] = {0};
 			for (; *p; p++) {
 				int byte_index = (pi / 8);
 				int bit_index = (pi % 8);
@@ -9685,7 +9685,7 @@ static void cmd_anal_opcode_bits(RCore *core, const char *arg, int mode) {
 				}
 			}
 			free (s);
-			for (i = 0; i < 8 && i < last; i++) {
+			for (i = 0; i < last; i++) {
 				r_cons_printf (core->cons, "%02x", finalmask[i]);
 			}
 			r_cons_newline (core->cons);


trufae:
Clever fix. Thanks. Hopefully we wont hit instructions larger than 32… which they exist. So.. can you make another pr showing a warning when this happens? And maybe truncating bytes and mask there?

PR #24727, opened 2025-10-22
Fix line counting for large output display (issue #1973)

### LINKED ISSUE #1973: Define number of lines when output is large
When output is too big in scr.interactive there should be more options than just "Yes" or "No".

I'm proposing the following:
* p : use scr.pager (less, more, ..)
* 10 : any numeric argument will asume that it must flush only the first 10 lines and drop the rest. 

We can add r_cons_head() and r_cons_tail() functions to chop the non-flushed buffer.

vicky-dx:
Improve console output confirmation by counting display lines more accurately when output is large.

Changes:
- Remove COUNT_LINES ifdef (keep line counting, drop char counting)
- Add count_display_lines() function
- Count long lines (>80 chars) as multiple display lines
- More accurate line count prevents misleading prompts like 'print 1 line?' for a 100MB single line

As suggested by @trufae, this makes the 'Do you want to print N lines?' prompt more meaningful by considering line wrapping.

Fixes #1973 (issue open since 2015)

### DIFF
--- a/libr/cons/cons.c
+++ b/libr/cons/cons.c
@@ -10,36 +10,28 @@ R_LIB_VERSION (r_cons);
 static RCons s_cons_global = {0};
 
 static void __break_signal(int sig);
+#define MAX_DISPLAY_LINES 9000
 
-// Count display lines, treating long lines (>80 chars) as multiple lines
-// Stops counting at R_CONS_MAX_LINES to avoid performance issues with huge outputs
-#define R_CONS_MAX_LINES 10000
-
-static unsigned int count_display_lines(const char *buffer, size_t len) {
-	if (!buffer || len == 0) {
-		return 0;
+static unsigned int count_display_lines(RCons *cons, const char *buffer, size_t len) {
+	int columns = r_cons_get_size (cons, NULL);
+	if (columns < 1) {
+		columns = 80;
 	}
 	unsigned int lines = 0;
-	unsigned int line_chars = 0;
-	size_t i;
-	for (i = 0; i < len && lines < R_CONS_MAX_LINES; i++) {
-		if (buffer[i] == '\n') {
-			lines++;
-			line_chars = 0;
-		} else {
-			line_chars++;
-			// Count extra line for every 80 characters in a single line
-			if (line_chars > 0 && line_chars % 80 == 0) {
-				lines++;
-				if (lines >= R_CONS_MAX_LINES) {
-					break;
-				}
-			}
+	const char *ptr = buffer;
+	const char *end = buffer + len;
+
+	while (ptr < end && lines < MAX_DISPLAY_LINES) {
+		const char *nl = strchr (ptr, '\n');
+		if (!nl || nl >= end) {
+			nl = end;
 		}
-	}
-	// Add one more line if buffer doesn't end with newline
-	if (line_chars > 0 && lines < R_CONS_MAX_LINES) {
-		lines++;
+		size_t line_len = nl - ptr;
+		lines += line_len ? ((line_len + columns - 1) / columns) : 1;
+		ptr = nl + 1;
+		if (nl == end) {
+			break;
+		}
 	}
 	return lines;
 }
@@ -890,8 +882,8 @@ R_API void r_cons_flush(RCons *cons) {
 			r_sys_cmd_str_full (cons->pager, ctx->buffer, -1, NULL, NULL, NULL);
 			r_cons_reset (cons);
 		} else if (cons->maxpage > 0 && ctx->buffer_len > cons->maxpage) {
-			unsigned int lines = count_display_lines (ctx->buffer, ctx->buffer_len);
-			if (lines > 0 && !r_cons_yesno (cons, 'n',"Do you want to print %d lines? (y/N)", lines)) {
+			unsigned int lines = count_display_lines (cons, ctx->buffer, ctx->buffer_len);
+			if (lines > 0 && !r_cons_yesno (cons, 'n',"Do you want to print %u lines? (y/N)", lines)) {
 				r_cons_reset (cons);
 				return;
 			}


libr/cons/cons.c
```diff
@@ -13,6 +11,32 @@ static RCons s_cons_global = {0};
...
 static void __break_signal(int sig);
 
+// Count display lines, treating long lines (>80 chars) as multiple lines
+static int count_display_lines(const char *buffer, size_t len) {
+	if (!buffer || len == 0) {
+		return 0;
+	}
+	int lines = 0;
+	int line_chars = 0;
+	for (size_t i = 0; i < len; i++) {
```
trufae:
```suggestion
	size_t i;
	for (i = 0; i < len; i++) {
```

libr/cons/cons.c
```diff
@@ -13,6 +11,33 @@ static RCons s_cons_global = {0};
 
 static void __break_signal(int sig);
 
+// Count display lines, treating long lines (>80 chars) as multiple lines
+static int count_display_lines(const char *buffer, size_t len) {
```
trufae: may be worth using unsigned or setting a limit to avoid counting more than X lines to avoid integer overflow. and also avoid counting too much trash

trufae:
the reason to not have this count lines enabled was because it was too slow sometimes, but maybe this approach will work better. thanks for taking care of it

libr/cons/cons.c
```diff
@@ -13,6 +11,39 @@ static RCons s_cons_global = {0};
 
 static void __break_signal(int sig);
 
+// Count display lines, treating long lines (>80 chars) as multiple lines
+// Stops counting at R_CONS_MAX_LINES to avoid performance issues with huge outputs
+#define R_CONS_MAX_LINES 10000
```
trufae: let's use 9000, so we can make the joke and tell the user the lines are over 9000. also remove the R_CONS prefix here because its local to this file. R_CONS must be only the ones in the .h

trufae (changes requested):
almost there

libr/cons/cons.c
```diff
@@ -859,26 +890,11 @@ R_API void r_cons_flush(RCons *cons) {
...
-#if COUNT_LINES
-			char *buffer = ctx->buffer;
-			int i, lines = 0;
-			for (i = 0; buffer[i]; i++) {
-				if (buffer[i] == '\n') {
-					lines ++;
-				}
-			}
+			unsigned int lines = count_display_lines (ctx->buffer, ctx->buffer_len);
 			if (lines > 0 && !r_cons_yesno (cons, 'n',"Do you want to print %d lines? (y/N)", lines)) {
```
trufae: if lines is unsigned we should use %u instead

libr/cons/cons.c
```diff
@@ -13,6 +11,39 @@ static RCons s_cons_global = {0};
 
 static void __break_signal(int sig);
 
+// Count display lines, treating long lines (>80 chars) as multiple lines
```
trufae: remove comments

libr/cons/cons.c
```diff
@@ -13,6 +11,39 @@ static RCons s_cons_global = {0};
...
+	unsigned int line_chars = 0;
+	size_t i;
+	for (i = 0; i < len && lines < R_CONS_MAX_LINES; i++) {
+		if (buffer[i] == '\n') {
+			lines++;
+			line_chars = 0;
+		} else {
+			line_chars++;
+			// Count extra line for every 80 characters in a single line
+			if (line_chars > 0 && line_chars % 80 == 0) {
```
trufae: let's remove the hardcoded "80" value here. use const int columns = r_cons_get_size

libr/cons/cons.c
```diff
@@ -13,6 +11,39 @@ static RCons s_cons_global = {0};
 
 static void __break_signal(int sig);
 
+// Count display lines, treating long lines (>80 chars) as multiple lines
+// Stops counting at R_CONS_MAX_LINES to avoid performance issues with huge outputs
+#define R_CONS_MAX_LINES 10000
+
+static unsigned int count_display_lines(const char *buffer, size_t len) {
+	if (!buffer || len == 0) {
```
trufae: remove those checks. if ctx->buffer is null we have a problem. also if len is zero should be fine to go thru the code below without any issues

libr/cons/cons.c
```diff
@@ -13,6 +11,39 @@ static RCons s_cons_global = {0};
...
+			// Count extra line for every 80 characters in a single line
+			if (line_chars > 0 && line_chars % 80 == 0) {
+				lines++;
+				if (lines >= R_CONS_MAX_LINES) {
+					break;
+				}
+			}
+		}
+	}
+	// Add one more line if buffer doesn't end with newline
```
trufae: remove unnecessary comment

libr/cons/cons.c
trufae: use size_t or ut32 or unsigned int ^

libr/cons/cons.c
```diff
@@ -13,6 +11,39 @@ static RCons s_cons_global = {0};
...
+
+static unsigned int count_display_lines(const char *buffer, size_t len) {
+	if (!buffer || len == 0) {
+		return 0;
+	}
+	unsigned int lines = 0;
+	unsigned int line_chars = 0;
+	size_t i;
+	for (i = 0; i < len && lines < R_CONS_MAX_LINES; i++) {
+		if (buffer[i] == '\n') {
```
trufae: maybe using strchr (buffer '\n') is quite faster than comparing char by char. because of assembly optimizations done inside libc. also, we must consider that the printed buffer ends with a null byte so the 'len' is only honored when printing raw bytes

libr/cons/cons.c
```diff
@@ -13,6 +11,39 @@ static RCons s_cons_global = {0};
...
+			lines++;
+			line_chars = 0;
+		} else {
+			line_chars++;
```
trufae: you can compute how many lines a wrapped line have if you use strchr comparsing with line pointer and newline one and mod by columns

vicky-dx:
Hey @trufae , I've applied all your review feedback. Thank you for the detailed suggestions!

Changes made:

-  Changed R_CONS_MAX_LINES 10000 → MAX_DISPLAY_LINES 9000 🤣
-  Removed R_CONS_ prefix (local to file)
-  Removed unnecessary null/zero checks
-  Using r_cons_get_size(cons, NULL) to get terminal columns dynamically
-  Optimized with strchr() instead of char-by-char loop
-  Removed all comments
-  Fixed printf format from %d to %u

**Line wrapping calculation fix:**
I noticed the initial formula lines += 1 + (line_len / columns) had an off-by-one issue at boundaries:

- 80 chars was counting as 2 lines (should be 1)
- 160 chars was counting as 3 lines (should be 2)

**Fixed with ceiling division:**
```c
lines += line_len ? ((line_len + columns - 1) / columns) : 1;
```

Tested and verified:

- 80 chars = 1 line 
- 81 chars = 2 lines 
- 160 chars = 2 lines 
- 4000 chars = 50 lines

trufae:
i'm merging it because i want to move forward and get your contribs in! but i think the linelimit should be pages limit.. this way we take in consideration the amount of rows in the terminal. What do you think? which limit would make sense 

Thanks!

vicky-dx:
> i'm merging it because i want to move forward and get your contribs in! but i think the linelimit should be pages limit.. this way we take in consideration the amount of rows in the terminal. What do you think? which limit would make sense

Hi @trufae 
I think it would be great idea! using pages limit makes more sense than absolute line count. 

Instead of MAX_DISPLAY_LINES = 9000, we could do:
- Get terminal rows: `r_cons_get_size(cons, &rows)`
- Set limit as: `max_pages * rows` (e.g., 100 pages * terminal_rows)
- This way on a 40-row terminal: 100 * 40 = 4000 lines limit
- On a 60-row terminal: 100 * 60 = 6000 lines limit

Should I create a follow-up PR for this improvement?

trufae:
yes! go for it! that plan sounds good

PR #24728, opened 2025-10-22
pxa command now honors functions, not just flags Fix #2953

### LINKED ISSUE #2953: pxa doesn't honors functions
It's only using flags right now

vicky-dx:
Fixes #2953 (10-year-old issue from July 2015)

## Problem
The `pxa` (annotated hexdump) command was only displaying flags, completely ignoring actual function boundaries defined by analysis. This caused confusion as function starts were not clearly marked.

## Solution
- Added `r_anal_get_function_at()` check before flag lookup
- Display function name with `/fcn.` prefix when a function starts
- Uses Color_YELLOW for function markers for better visibility
- Maintains existing flag and comment functionality

## Testing
Before:
/entry0 <- only flag

After:
/fcn.entry0 <- function marker
/entry0 <- flag

Now users can clearly see where analyzed functions begin during hexdump annotation.


trufae:
fix the test
r2r -i db/cmd/cmd_flags

libr/core/cmd_print.inc.c
```diff
@@ -2622,6 +2622,20 @@ static void annotated_hexdump(RCore *core, const char *str, int len) {
...
+			// collect functions
+			RAnalFunction *fcn = r_anal_get_function_at (core->anal, addr + j);
+			if (fcn) {
+				if (show_section) {
+					r_cons_printf (core->cons, "%20s ", "");
+				}
+				if (usecolor) {
+					r_cons_printf (core->cons, "%s", Color_YELLOW);
+				}
+				r_cons_printf (core->cons, "             /fcn.%s\n", fcn->name);
```
trufae: these arbitrary amount of spaces seems wrong to me

libr/core/cmd_print.inc.c
```diff
@@ -2622,6 +2622,20 @@ static void annotated_hexdump(RCore *core, const char *str, int len) {
 				note[j] = r_str_newf (";%s", comment);
 				marks = true;
 			}
+			// collect functions
+			RAnalFunction *fcn = r_anal_get_function_at (core->anal, addr + j);
+			if (fcn) {
+				if (show_section) {
+					r_cons_printf (core->cons, "%20s ", "");
+				}
+				if (usecolor) {
```
trufae, line 2631: just if (usecolor) once, this way you can compact these 7 lines into just 5 (if usecolor {print} else{print})

trufae:
ping

vicky-dx:
I've fixed the failing test as requested. All 48 tests in `cmd_flags` now pass.

### What Was Wrong

 when `r2r` runs tests, it redirects stdout to capture and compare output. When stdout is **not a TTY**, radare2 automatically disables color output, regardless of the `scr.color` setting.

### Fix:

Updated test expectations to match the actual non-TTY output that `r2r` sees:
- Kept the function annotations: `/fcn.entry0` 
- Removed ANSI escape codes from expected output 

### Verification

**Tests pass:**
```bash
[48/48]  48 OK  0 BR  0 XX  0 SK  0 FX
```

Feature works correctly:
When running interactively (TTY mode), colors DO appear as expected:

PR #24732, opened 2025-10-23
Implement SDB-based command suggestions (#23406)

### LINKED ISSUE #23406: sdb/kv database with missing commands and suggestions
useful for suggesting plugins etc, isntead of hardcoding it in C:

```
r2dec=you are probably looking for the pdd command
pdd=you need to install the plugin with r2pm -ci r2dec 
...
```

vicky-dx:
- Replace hardcoded plugin error messages with SDB lookup
- Add init_cmd_suggestions() to populate default suggestions
- Unified error handling for pd{d,g,z,v} commands
- Support for custom suggestions via SDB at runtime
- Reduces code redundancy by centralizing messages

Example usage:
  pdd  → Missing plugin. Run: r2pm -ci r2dec pdg  → Missing plugin. Run: r2pm -ci r2ghidra

Users can add custom suggestions:
  k mycommand=Use 'othercommand' instead

Closes #23406

### DIFF
--- a/libr/core/cmd.c
+++ b/libr/core/cmd.c
@@ -6371,21 +6371,16 @@ static int run_cmd_depth(RCore *core, char *cmd) {
 		if (R_UNLIKELY (ret == -1)) {
 			// Check for fallback command in SDB (fallbackcmd.* namespace)
 			char *fallback_key = r_str_newf ("fallbackcmd.%s", rcmd);
-			if (fallback_key) {
-				const char *fallback_cmd = sdb_const_get (core->sdb, fallback_key, NULL);
-				if (fallback_cmd && r_str_startswith (fallback_cmd, "?e ")) {
+			const char *fallback_cmd = sdb_const_get (core->sdb, fallback_key, NULL);
+			if (fallback_cmd) {
+				if (r_str_startswith (fallback_cmd, "?e ")) {
 					// Execute safe ?e (echo) command only
 					r_core_cmd0 (core, fallback_cmd);
-				} else if (fallback_cmd) {
-					R_LOG_WARN ("Fallback command for '%s' exists but is not a safe ?e command", rcmd);
-					R_LOG_ERROR ("Invalid command '%s' (0x%02x)", rcmd, *rcmd);
-				} else {
-					R_LOG_ERROR ("Invalid command '%s' (0x%02x)", rcmd, *rcmd);
 				}
-				free (fallback_key);
 			} else {
 				R_LOG_ERROR ("Invalid command '%s' (0x%02x)", rcmd, *rcmd);
 			}
+			free (fallback_key);
 			break;
 		}
 		if (!ptr) {


trufae:
This is interesting! i think we can extend this logic to replace the '$' command with 'alias' and expose that functionality as a first class command. the behaviour we are looking for is:

* Create a command that will override other commands
* Create a command to run when no command was claiming for it

The first one looks like "alias" to me, and this alias thing is done right now as part of the '$' but only works with commands prefixed with the dollar (this is because this way we can forbid users to break the entire shell if they add new alias commands that rimraf the user home for example)

the second action is what you did. But i dont want the core->sdb to be used with "key"=value for this because we should be using a prefix like "fallbackcmd.pdd=?e install r2pm -ci r2dec"

and when this happens return the error code to the shell

this way we can do other actions not just echo a message.

to protect the user from causing hell maybe we should maybe keep using '$' for the new commands (despite i dislike the syntax). but we can reuse this core->sdb using a different prefix for that. Similar stuff is done with the macro system (under the '(' command).

so let's discuss after that and see the best approach to unify thse 3 concepts in a common clear way

vicky-dx:
@trufae Thank you for the thoughtful feedback! I love where you're taking this - unifying alias/$, fallback, and macros makes total sense.



I just need some quick clarification before I refactor:

**1. Prefix in SDB**
Use `fallbackcmd.pdd` instead of just `pdd` - ✓ clear

**2. Command execution** 
You said: `fallbackcmd.pdd=?e install r2pm -ci r2dec`

So fallback can execute ANY r2 command (not just echo)? Including:
- `?e` (echo)
- `?(topic)` (help)
- `$alias` (call user alias)
- Shell commands `!` (probably restrict this?)

Should I whitelist safe commands or trust r2's existing command sandbox?

**3. Return codes**
When fallback executes, return special exit code so scripts know "command not found but handled"? 

**4. Scope for this PR**
Start with just the `fallbackcmd.*` prefix + command execution, or also tackle the broader alias/macro unification you mentioned?

I can implement the `fallbackcmd.*` system properly now, and we can unify with `$` and `(` later as a separate effort. Sound good?

trufae:
Please resolve the conflicts. let's filter to accept only "?e" command so we maintain functionality and keep it safe and expected. we can work on doing redesign for the alias, macros later on with that as base

vicky-dx:
sorry, i am currently out of town, i will be back in 2 days

trufae:
run: r2r -i db/cmd/cmd_graph

and press 'y' to update test output. with that the ci should be green i guess

also fix the linting. run sys/lint.sh and remove empty lines with spaces and such like the ones reported:

Run sys/lint.sh
libr/core/core.c:1791:	
libr/core/core.c:1795:

libr/core/cmd.c
```diff
@@ -6369,7 +6369,23 @@ static int run_cmd_depth(RCore *core, char *cmd) {
...
+			// Check for fallback command in SDB (fallbackcmd.* namespace)
+			char *fallback_key = r_str_newf ("fallbackcmd.%s", rcmd);
+			if (fallback_key) {
+				const char *fallback_cmd = sdb_const_get (core->sdb, fallback_key, NULL);
+				if (fallback_cmd && r_str_startswith (fallback_cmd, "?e ")) {
+					// Execute safe ?e (echo) command only
+					r_core_cmd0 (core, fallback_cmd);
+				} else if (fallback_cmd) {
+					R_LOG_WARN ("Fallback command for '%s' exists but is not a safe ?e command", rcmd);
+					R_LOG_ERROR ("Invalid command '%s' (0x%02x)", rcmd, *rcmd);
```
trufae:
```suggestion
```

libr/core/cmd.c
```diff
@@ -6369,7 +6369,23 @@ static int run_cmd_depth(RCore *core, char *cmd) {
 		}
 		ret = r_core_cmd_subst (core, rcmd);
 		if (R_UNLIKELY (ret == -1)) {
-			R_LOG_ERROR ("Invalid command '%s' (0x%02x)", rcmd, *rcmd);
+			// Check for fallback command in SDB (fallbackcmd.* namespace)
+			char *fallback_key = r_str_newf ("fallbackcmd.%s", rcmd);
+			if (fallback_key) {
```
trufae: lets assume this cant be null, no need to chk for null here

libr/core/cmd.c
```diff
@@ -6369,7 +6369,23 @@ static int run_cmd_depth(RCore *core, char *cmd) {
...
+				const char *fallback_cmd = sdb_const_get (core->sdb, fallback_key, NULL);
+				if (fallback_cmd && r_str_startswith (fallback_cmd, "?e ")) {
```
trufae: just if (fallback_cmd) { if (startswith("?e) ... } else { LOG_ERROR (invalidCommand)}

libr/core/cmd.c
```diff
@@ -6369,7 +6369,23 @@ static int run_cmd_depth(RCore *core, char *cmd) {
...
+					r_core_cmd0 (core, fallback_cmd);
+				} else if (fallback_cmd) {
+					R_LOG_WARN ("Fallback command for '%s' exists but is not a safe ?e command", rcmd);
+					R_LOG_ERROR ("Invalid command '%s' (0x%02x)", rcmd, *rcmd);
+				} else {
+					R_LOG_ERROR ("Invalid command '%s' (0x%02x)", rcmd, *rcmd);
+				}
+				free (fallback_key);
+			} else {
+				R_LOG_ERROR ("Invalid command '%s' (0x%02x)", rcmd, *rcmd);
```
trufae, line 6381: this else will never be taken after the refactoring proposed above

trufae:
Merged! now I can start thinking in a cleaner way to have this as well as integrating it with the new 'alias' and '$' command, and maybe move that from Core's sdb into the RCoreCmd structure to make it more solid. if you want to give it some thoughts and propose a design plan i'll happy to read it

Issue #19003, opened 2021-08-19
Unique coloring for simmilar immediates


swoops:
## Description

*Please describe what are you missing or wanting to be improved*
Recently I read a new syntax coloring technique for programing languages. Instead of highlighting grammar, highlight the things you are likely to mistake. So make sure `strcpy` and `strncpy` are easily distinguishable by their color. It would be cool to apply this to immediate values in assembly and the `?` command.

For example, it is difficult to distinguish between 0x9d343f8b318614d1 and 0x9d345f8b318614d1. I don't even read that number all the way when I see it, I won't recognize it if I see it again later in code. If later the result of a math operation (from `?` cmd) is 0x9d343f8b318614d1, I won't remember the significance of the number. Applying a unique color to similar numbers would help the user remember if they had seen the number before. It would make it easy to see if the same number is used a lot in large chunks of code, or if numbers vary slightly.

Initial implementation could just be a simple hash of the immediate, then index that value from an accepted color space.

** If anyone knows the article I am speaking of, please link it. I can't find it anymore and I would like them to receive credit for the idea.

hlee131:
Could I take this issue? If so, could I get some pointers on which files I should be looking at to implement this?

swoops:
@hlee131  Yeah, feel free to take it.

Once you use `sys/install.sh` once, you *should* just be able to run `make` for each of your updates.

Not sure where to start. I imagine pointer printing is all over the place. I would open `/bin/ls` like you are going to reverse it, find the first pointer you see in your normal reversing processes and figure out what printed that code. It should be a `r_cons_printf` so you might break on that.

Sorry I can't be of more help there.

PR #24733, opened 2025-10-23
Add unique coloring for immediate values #19003

vicky-dx:
## Description
Implements unique color assignment for immediate values in disassembly output, making it easier to distinguish similar hex values like `0x1234567890abcdef` vs `0x1234567890abcdee`.

## Changes
- Added config option `scr.color.ops.unique` (default: false)
- Added `unique_colors` boolean field to `RPrint` structure
- Implemented hash-based deterministic color palette with 10 distinct colors
- Identical immediate values consistently receive the same color
- Different values receive different colors (with acceptable ~10% collision rate)

## Usage
```bash
e scr.color.ops.unique=true
pd 20
```

### Testing

```c
int main() {
    int a = 0xDEADBEEF;
    int b = 0xDEADBEEF;  // Same as a
    int c = 0xDEADBEEF;  // Same as a
    int d = 0xCAFEBABE;  // Different
    int e = 0xDEADBEEF;  // Same as a
    return a + b + c + d + e;
}
```

```bash
gcc -o /tmp/test_colors /tmp/test_colors.c
```

### Test 1: Without Unique Colors (Default Behavior)
```bash
r2 -qc 'e scr.color=3; e scr.color.ops=true; e scr.color.ops.unique=false; s main; pd 15' /tmp/test_colors
```
image

> Expected: All immediate values (0xdeadbeef and 0xcafebabe) appear in the SAME default color.

### Test 2: With Unique Colors (New Feature)
```bash
r2 -qc 'e scr.color=3; e scr.color.ops=true; e scr.color.ops.unique=true; s main; pd 15' /tmp/test_colors
```

image

> Expected:
> 
> All four 0xDEADBEEF values appear in the SAME color (e.g., blue)
>0xCAFEBABE appears in a DIFFERENT color (e.g., magenta)
> 

### Test 3: Similar Hex Values (Main Use Case)

```c
int main() {
    long long a = 0x1234567890ABCDEF;
    long long b = 0x1234567890ABCDEE;  // Differs by last digit only
    return 0;
}
```
```bash
gcc -o /tmp/test_similar /tmp/test_similar.c
```
image

> Expected: The two similar hex values will have DIFFERENT colors, making it easy to spot they're different.

### DIFF
--- a/libr/core/cconfig.c
+++ b/libr/core/cconfig.c
@@ -4569,6 +4569,7 @@ R_API int r_core_config_init(RCore *core) {
 	SETCB ("scr.color.grep", "false", &cb_scr_color_grep, "enable colors when using ~grep");
 	SETB ("scr.color.pipe", "false", "enable colors when using pipes");
 	SETB ("scr.color.ops", "true", "colorize numbers and registers in opcodes");
+	SETB ("scr.color.ops.unique", "false", "use unique colors for each immediate value");
 	SETCB ("scr.color.ophex", "false", &cb_scr_color_ophex, "colorize in hexdump depending on opcode type (px)");
 	SETB ("scr.color.args", "true", "colorize arguments and variables of functions");
 	SETB ("scr.color.bytes", "true", "colorize bytes that represent the opcodes of the instruction");
--- a/libr/core/disasm.c
+++ b/libr/core/disasm.c
@@ -728,6 +728,7 @@ static RDisasmState *ds_init(RCore *core) {
 	ds->show_color_bytes = r_config_get_i (core->config, "scr.color.bytes");
 	ds->show_color_args = r_config_get_i (core->config, "scr.color.args");
 	ds->colorop = r_config_get_b (core->config, "scr.color.ops");
+	core->print->unique_colors = r_config_get_b (core->config, "scr.color.ops.unique");
 	ds->show_utf8 = r_config_get_i (core->config, "scr.utf8");
 	ds->acase = r_config_get_b (core->config, "asm.ucase");
 	ds->capitalize = r_config_get_b (core->config, "asm.capitalize");
--- a/libr/include/r_util/r_print.h
+++ b/libr/include/r_util/r_print.h
@@ -110,6 +110,7 @@ typedef struct r_print_t {
 	int cols;
 	int flags;
 	bool use_comments;
+	bool unique_colors;
 	int addrmod;
 	int col;
 	int stride;
--- a/libr/util/print.c
+++ b/libr/util/print.c
@@ -2314,6 +2314,29 @@ static bool is_flag(const char *p) {
 	return len > 3;
 }
 
+// Helper function to generate unique color for immediate values
+static const char* get_unique_imm_color(ut64 value) {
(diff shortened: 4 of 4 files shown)


trufae:
i think that's what scr.rainbow does already

trufae:
can you confirm or improve that  option instead of adding more? also i dislike the scr.color.ops.unique

vicky-dx:
> can you confirm or improve that option instead of adding more? also i dislike the scr.color.ops.unique

@trufae  I tested `scr.rainbow` and it's different from this PR:

**scr.rainbow:**
- Colors based on **address** ("shows rainbow colors depending of address")
- Each instruction at a different address gets a different color
- Not related to immediate values

**This PR (scr.color.ops.unique):**
- Colors based on **immediate value content**
- Same immediate value = same color (e.g., all `0xdeadbeef` in blue)
- Different immediate values = different colors (e.g., `0xcafebabe` in magenta)
- Helps distinguish similar hex values like `0x1234567890abcdef` vs `0x1234567890abcdee`



but if you want 
i can make it part of existing `scr.color.ops`
Instead of new boolean, i'll make `scr.color.ops` an enum:

`scr.color.ops=0` → no coloring
`scr.color.ops=1` → basic coloring (current)
`scr.color.ops=2` → unique immediate coloring (new)

what's your thought on that?

trufae:
Agree its not exactly the same.

That seems interesting too, because it feels like how scr.color works. But i think .ops is not the best name we can use here. maybe something like scr.color.imm will fit better, also we can have =3 to colorize using the same addr-to-color code from the rainbow and maybe we can do the same for asm.addr and remove the scr.rainbow? buit that will be another pr! lets finish some of yours first before opening more fronts!

vicky-dx:
> Agree its not exactly the same.
> 
> That seems interesting too, because it feels like how scr.color works. But i think .ops is not the best name we can use here. maybe something like scr.color.imm will fit better, also we can have =3 to colorize using the same addr-to-color code from the rainbow and maybe we can do the same for asm.addr and remove the scr.rainbow? buit that will be another pr! lets finish some of yours first before opening more fronts!



I have few concerns 

```c
scr.color.imm = 0   → no immediate coloring (but what about scr.color.ops?)
scr.color.imm = 1   → basic coloring
scr.color.imm = 2   → unique coloring
```

don't you think it will crate ambiguity?

What if ` scr.color.ops = true` but `scr.color.imm = 0`?
Does `scr.color.imm` override `scr.color.ops` for immediates only?


what i was thinking of doing ->

```C
scr.color.ops = false   → no operand coloring at all
scr.color.ops = true    → basic operand coloring
   ─ scr.color.ops.imm = 0   → basic immediate coloring
   ─ scr.color.ops.imm = 1   → unique immediate coloring
```
this looks clearer to me, because of hierarchy obvious
no ambiguity if parent is false, children don't matter
consistent with radare2's existing config style,
and `src.color.ops` remains the mater switch for all operands

or simpler
```C
scr.color.ops.imm = 0   → basic coloring (default)
scr.color.ops.imm = 1   → unique coloring
```


so whats your thought on this?

Issue #16193, opened 2020-03-12
Colorize the `pf` output (especially in `r2 -nn binary` case)


XVilka:
**Is your feature request related to a problem? Please describe.**
Currently radare2 prints both `pf` and built-in `pf` output in `r2 -nn` mode without any colors, even if they are enabled. It would be amazing to improve the readability to colorize them according to the current color options and colorscheme.

![image](https://user-images.githubusercontent.com/203261/76514618-2449ae00-6493-11ea-8b1d-e3d365809199.png)

radare:
also i as thinking in having ad integrated into pf.. so we can tell pf to magically guess the type stored in one place by telescoping it and making some assumptions to see if the dword matches an allocated address, or printable chars to represent a string.

also what i find out is that ad is not detecting the strings when the first byte is nul.. basically "" strings in disassembly never appear.

vicky-dx:
> also i as thinking in having ad integrated into pf.. so we can tell pf to magically guess the type stored in one place by telescoping it and making some assumptions to see if the dword matches an allocated address, or printable chars to represent a string.
> 
> also what i find out is that ad is not detecting the strings when the first byte is nul.. basically "" strings in disassembly never appear.

This PR focuses on **cosmetic improvements** to enhance readability by adding colors to the existing `pf` output format. 

It does **not** implement smart type detection or telescoping (as discussed by @radare in the issue comments). The foundation laid here with color helper functions will work seamlessly with any future smart type detection features.

Looking forward to seeing smart type detection implemented in the future! 🥹😊

PR #24734, opened 2025-10-23
Add colorization to pf output #16193

vicky-dx:
- Added helper functions for maintainable color management
- Addresses use addr palette color (green)
- Values use b0xff palette color (red/magenta)
- Updated format functions: byte, word, hex, int, octal, qword
- Arrays are properly colorized
- Easy to modify color scheme by changing helper functions


trufae:
i think you are printing colors always, even when the scr.color=0

trufae:
also the colorization of pf should use the checks to verify if the value falls into a rx, rw, ro or invalid area like its done in pxr or drr commands. this way it will be easier to read the colors in the output

trufae:
ping

trufae:
pf has been fully rewritten (format2.c is now the default) so this pr can be closed

Issue #14765, opened 2019-08-06
Convert hardcoded shellcodes in libr/egg into assembly sources


XVilka:
Currently they are a byte sequence. Should be an assembly sources, preferably compilable by rasm2.

radare:
Thats exactly how you get the hexpairs. Internally its passing the text representstion to the assembler apis

XVilka:
hm? I don't see assembly sources in the repo then.

PR #24741, opened 2025-10-25
Egg/convert x86 linux binsh

vicky-dx:
## Convert shellcodes to human-readable assembly sources

This PR converts **6 out of 16** hardcoded shellcodes to rasm2-compatible assembly sources as requested in issue #14765.

### Converted in this PR:
-  `x86-linux-binsh.rasm` - 32-bit Linux /bin/sh (24 bytes)
-  `x86_64-linux-binsh.rasm` - 64-bit Linux /bin/sh (28 bytes)
-  `x86-osx-binsh.rasm` - macOS x86_64 /bin/sh with setuid(0) (45 bytes)
-  `x86-osx-suidbinsh.rasm` - macOS suid variant (identical to x86-osx-binsh)
-  `arm-linux-binsh.rasm` - 32-bit ARM Linux /bin/sh (47 bytes)
-  `thumb-linux-binsh.rasm` - ARM Thumb mode /bin/sh (31 bytes)

### Remaining shellcodes (for future PRs):
- arm-linux-bind.c
- armle-osx-reverse.c
- sparc-linux-bind4444.c
- x86-freebsd-reverse.c
- x86-linux-bind4444.c
- x86-linux-udp4444.c
- x86-openbsd-bind6969.c
- x86-osx-bind4444.c
- x86-solaris-bind4444.c
- x86-w32-tcp4444.c

### Verification:
All assembly files produce **byte-exact output** matching original shellcodes:
```bash
rasm2 -a x86 -b 32 -f libr/egg/p/sc/asm/x86-linux-binsh.rasm
# Output: 31c050682f2f7368682f62696e89e3505389e199b00bcd80

### DIFF
--- a/libr/egg/p/sc/Makefile
+++ b/libr/egg/p/sc/Makefile
@@ -25,6 +25,6 @@ test: test-shellcodes
 
 test-shellcodes:
 	@echo "Testing shellcode assembly..."
-	@cd "$(shell pwd)" && ./test-shellcodes.sh
+	@$(SHELL) ../../../../sys/test-egg-shellcodes.sh .
 
 .PHONY: all clean test test-shellcodes
--- a/sys/test-egg-shellcodes.sh
+++ b/sys/test-egg-shellcodes.sh
@@ -1,5 +1,9 @@
 #!/bin/sh
 # Test shellcode assembly files against original C versions
+# Usage: sys/test-egg-shellcodes.sh [path-to-sc-dir]
+
+SC_DIR="${1:-libr/egg/p/sc}"
+cd "$SC_DIR" || exit 1
 
 fail=0
 
@@ -18,25 +22,25 @@ for rasm in asm/*.rasm; do
 	
 	# Determine architecture and bits from filename
 	case "$base" in
-		x86-linux-binsh)
-			actual=$(rasm2 -a x86 -b 32 -f "$rasm" 2>/dev/null)
-			;;
-		x86_64-linux-binsh)
-			actual=$(rasm2 -a x86 -b 64 -f "$rasm" 2>/dev/null)
-			;;
-		x86-osx-binsh|x86-osx-suidbinsh)
-			actual=$(rasm2 -a x86 -b 64 -f "$rasm" 2>/dev/null)
-			;;
-		arm-linux-binsh)
-			actual=$(rasm2 -a arm -b 32 -f "$rasm" 2>/dev/null)
-			;;
-		thumb-linux-binsh)
-			actual=$(rasm2 -a arm -f "$rasm" 2>/dev/null)
-			;;
-		*)
-			echo "  [--] $base: Unknown architecture, skipping"
-			continue
-			;;
+	x86-linux-binsh)
+		actual=$(rasm2 -a x86 -b 32 -f "$rasm" 2>/dev/null)
+		;;
+	x86_64-linux-binsh)
+		actual=$(rasm2 -a x86 -b 64 -f "$rasm" 2>/dev/null)
+		;;
+	x86-osx-binsh|x86-osx-suidbinsh)
+		actual=$(rasm2 -a x86 -b 64 -f "$rasm" 2>/dev/null)
+		;;
+	arm-linux-binsh)
+		actual=$(rasm2 -a arm -b 32 -f "$rasm" 2>/dev/null)
+		;;
+	thumb-linux-binsh)
+		actual=$(rasm2 -a arm -f "$rasm" 2>/dev/null)
+		;;
+	*)
+		echo "  [--] $base: Unknown architecture, skipping"
+		continue
+		;;
 	esac
 	
 	if [ "$actual" = "$expected" ]; then

### CODE REVIEW

**Comment by trufae** [comment]
can you send a separate PR only with the null deref check? i dont think that we should be providing assembly versions of the shellcodes. maybe we can have a script that does that with rax2 and rasm2. but considering most shellcodes do jump-in-the-middle and have gaps to fill im unsure if that would be good to provide as is.

maybe @gum3t have some thoughts or can use those .rasm for testing chameleon

**Comment by vicky-dx** [comment]
@trufae The NULL deref fix is already merged in upstream master (commit 61a1dfcb85)! 

I'll close this PR since the main value (the NULL check) is already incorporated. The assembly shellcode conversions can be revisited later if there's a clearer use case.

Thanks for the review!

**Comment by trufae** [comment]
i think its a good start, and i think it will be very nice to use these shellcodes for more than just an asm reference. im spending some time to improve rasm and we need tests! so having a makefile that uses rasm2 to compare the output with the .h from the original shellcodes will be ideal. also check the inline comments in the files you pushed

**Comment by trufae** [comment]
please rebase so the nullderef commit goes away

File: libr/egg/p/sc/asm/arm-linux-binsh.rasm [hunk 1]
```diff
@@ -0,0 +1,22 @@
+; ARM 32-bit Linux execve /bin/sh shellcode
+; Assembly equivalent of libr/egg/p/sc/src/arm-linux-binsh.c
+;
+; Execve("/bin/sh", ["/bin/sh", NULL], NULL) using syscall 0xb
+
```
trufae, line 5: use .arch and .bits directives to make this .rasm file assemble fine with rasm2.

libr/egg/p/sc/asm/arm-linux-binsh.rasm
```diff
@@ -0,0 +1,22 @@
...
+add r3, pc, 0x1c            ; r3 = address of "/bin/sh" string
+str r3, [sp, 4]             ; argv[0] = "/bin/sh"
+str r2, [sp, 8]             ; argv[1] = NULL
+lsl r0, r3, r2              ; r0 = r3 (filename ptr)
+strb r2, [r3, 7]            ; null-terminate string at "/bin/sh\0"
+add r3, pc, 4               ; r3 = address of argv on stack
+add r1, sp, 4               ; r1 = argv pointer
+strb r2, [r3, 1]            ; another NULL byte
+svc 0x900b0b                ; syscall execve (syscall 0xb)
+.byte 0x2f
```
trufae: maybe a .string instead?

libr/egg/p/sc/asm/thumb-linux-binsh.rasm
```diff
@@ -0,0 +1,43 @@
+; ARM/Thumb Linux execve /bin/sh shellcode
+; Assembly equivalent of libr/egg/p/sc/src/thumb-linux-binsh.c
+;
+; Starts in ARM mode, switches to Thumb, then calls execve
+; Note: rasm2 doesn't support mixed ARM/Thumb in single file,
+; so this file documents the assembly but uses .byte for exact encoding
+
+; ARM mode - switch to Thumb
+.byte 0x01  ; add r3, pc, 1
```
trufae: thats as readable as the .hex. please use asm code or remove this .rasm

libr/egg/p/sc/asm/x86-linux-binsh.rasm
```diff
@@ -0,0 +1,30 @@
+; x86 (32-bit) execve /bin/sh shellcode
+; Assembly equivalent of the bytes originally stored in
+; libr/egg/p/sc/src/x86-linux-binsh.c
+; This file is rasm2-friendly (Intel syntax).
+; How to assemble and generate the corresponding C out file:
+;   rasm2 -a x86 -b 32 -c -d "<asm>"  # one-liner
+; or use the provided script (if available) to emit sc/out/*.c
+
+; xor eax, eax
```
trufae: dupped comments and code? fix that

libr/egg/p/sc/asm/x86-osx-binsh.rasm
```diff
@@ -0,0 +1,46 @@
...
+; Assembly equivalent of the bytes in libr/egg/p/sc/src/x86-osx-binsh.c
+; 
+; This shellcode first calls setuid(0) and then execve("/bin/sh", argv, NULL)
+; Uses jmp/call technique for position-independent string address
+;
+; Note: Using .byte directives for encoding-sensitive instructions
+; to match original bytes exactly (rasm2 may produce different encodings)
+
+; mov r8b, 2 - Setup syscall base
+.byte 0x41
```
trufae, line 11: if any of those instructions isnt assembled properly we may want to improve r2 assemblers to make them work. lets think about this task as a way to improve rasm2

libr/egg/p/sc/asm/x86-osx-suidbinsh.rasm
```diff
@@ -0,0 +1,47 @@
...
+; Note: This is identical to x86-osx-binsh.c
+; 
+; This shellcode first calls setuid(0) and then execve("/bin/sh", argv, NULL)
+; Uses jmp/call technique for position-independent string address
+;
+; Note: Using .byte directives for encoding-sensitive instructions
+; to match original bytes exactly (rasm2 may produce different encodings)
+
+; mov r8b, 2 - Setup syscall base
+.byte 0x41
```
trufae, line 12: same here

libr/egg/p/sc/asm/x86_64-linux-binsh.rasm
```diff
@@ -0,0 +1,31 @@
+; x86_64 (64-bit) execve /bin/sh shellcode
+; Assembly equivalent of the bytes originally stored in
+; libr/egg/p/sc/src/x86_64-linux-binsh.c
+; This file is rasm2-friendly (Intel syntax).
+
+; xor eax, eax
```
trufae: unecessary comments

vicky-dx: :D SORRY those comment were for me, i will remove it

libr/egg/p/sc/asm/x86-osx-binsh.rasm
vicky-dx: Regarding the `.byte` directives in x86-osx files:

I've kept the `.byte` directives for these specific instructions because rasm2 currently produces different (but functionally equivalent) opcodes:

**ex:**
```bash
$ rasm2 -a x86 -b 64 'shl r8, 0x18'
48c1e018    # rasm2 produces this

# But original shellcode uses:
49c1e018    # This encoding
```
oth are valid and execute the same way, but to maintain byte-exact compatibility with the original shellcodes which is important for exploitation/security contexts, I've used .byte directives for these encoding-sensitive instructions.

As you mentioned, improving rasm2 to handle these encodings correctly would be the proper fix. I can open a separate issue to track improving rasm2's x86-64 instruction encoding if you'd like?

For now, I've:

-  Added comments explaining why .byte is used
-  Used real assembly for all other instructions (xor, mov, syscall, etc.)
-  Used .ascii for strings instead of .byte
-  Verified output matches original shellcodes byte-for-byte via make test

The files are much more readable now compared to the original all-.byte version, while still maintaining exact binary compatibility.

libr/egg/p/sc/Makefile
```diff
@@ -20,4 +20,55 @@ clean:
 	rm -rf out
 	rm -f *.c
 
-.PHONY: all clean
+# Test target: verify rasm2 output matches original shellcodes
+test: test-shellcodes
+
+test-shellcodes:
```
trufae, line 26: shellscript inside makefiles sucks. move that into a shellscript and call it from themakefile instead.the schellscript can take the .rasm file name to test and the arch/bits should be implicit by the filename. lets try to keep this thing simple

libr/egg/p/sc/Makefile
```diff
@@ -20,4 +20,55 @@ clean:
...
+			if [ "$$actual" = "$$expected" ]; then \
+				echo "  ✓  $$base: PASS ($$actual)"; \
+			else \
+				echo "  ✗  $$base: FAIL"; \
+				echo "     Expected: $$expected"; \
+				echo "     Got:      $$actual"; \
+				fail=$$((fail + 1)); \
+			fi; \
+		else \
+			echo "  ⚠️  $$base: No source file found, skipping"; \
```
trufae: dont use emojis better [XX] or "[OK]" like r2r do

trufae:
fix the linter complains: Run sys/lint.sh
libr/egg/p/sc/asm/x86-osx-binsh.rasm:3:; 
libr/egg/p/sc/asm/x86-osx-suidbinsh.rasm:4:;

PR #24767, opened 2025-11-01
Change read timeout for sockets created using the connect option in r…

SverkerBerggren:

**Description**

When using rarun2 to connect the stdin and stdout to a socket so is there a 1-second timeout when reading from the socket. This makes it hard to interact with the program in an interactive way.

### DIFF
--- a/libr/socket/run.c
+++ b/libr/socket/run.c
@@ -975,6 +975,8 @@ R_API bool r_run_config_env(RRunProfile *p) {
 				r_socket_free (fd);
 				return false;
 			}
+			r_socket_block_time(fd, true, 99999, 0);
+
 			if (p->_pty) {
 				if (!redirect_socket_to_pty (fd)) {
 					R_LOG_ERROR ("socket redirection failed");


trufae:
related to 462f4aea7c4fe12f8465e8f7e00aa5db577cafb2

maybe we shuold revert the other commit, but enforcing a specific timeout when using sockets i think its worth it

libr/socket/run.c
```diff
@@ -975,6 +975,8 @@ R_API bool r_run_config_env(RRunProfile *p) {
 				r_socket_free (fd);
 				return false;
 			}
+			r_socket_block_time(fd, true, 99999, 0);
```
trufae:
```suggestion
			r_socket_block_time (fd, true, 0, 0);
```

trufae: disabling block time is probably better than having a long sleep

SverkerBerggren:
I do not quite understand your point, do you think that the sockets should have the same timeout? I think that reading from stdin should block indefinitely, because that is the standard behaviour. The linked issue seems a bit unrelated to this example because it concerns sockets that are not supposed to emulate stdin.

trufae:
i found the bug was here https://github.com/radareorg/radare2/pull/24776 if you set block_timeout to 0,0 after this pr is merged you may be able to disable the timeout.

your code is just making a long timeout which makes no sense for stdin as you said. can you rebase after merging the pr, update the pr and test to confirm the behaviour?

Issue #24769, opened 2025-11-01
`cmd.esil.trap` command not executed


sagittarius-a:
## Environment

```sh
Sat Nov  1 21:09:51 UTC 2025
radare2 5.9.8 34281 @ linux-arm-64
birth: git.5.9.8 2025-11-01__21:00:49
commit: 4eb49d5ad8c99eaecc8850a2f10bad407067c898
options: gpl -O1 cs:5 cl:1 meson
Linux aarch64
```

_I tried with version 5.7.8 as well and the following build_

```
radare2 6.0.4 0 @ darwin-arm-64
birth: git.6.0.4 2025-09-27__18:08:23
commit: 6.0.4
options: gpl release -O1 cs:5 cl:2 make
Darwin arm64
```

## Description

The command set in `cmd.esil.trap` is not executed.

I found the issue while playing with [r2wars](https://github.com/radareorg/r2wars). Some bots were triggering the ESIL TRAP but the command was not executed.

I don't know if `cmd.esil.*` are deprecated or if it was simply not implemented after some code changes.

## Test

Open any binary with r2 then:

```
# Initialize ESIL environment
aei
aeim
aeip

# set the ESIL hook
e cmd.esil.trap=f theend=2

f theend=0

# Manually trigger a trap
aer pc=0xdeadbeef
aes

# Check result
?v theend
```

trufae:
this probably needs more testing, can you add more tests in the ci? i just open a pr with one of these cases fixed now
```
[0x00006d30]> ahe TRAP
[0x00006d30]> e cmd.esil.trap =?E TRAP
[0x00006d30]> aes
 ╭──╮    ╭──────────────╮
 │ _│_   │              │ 
 │ O O  <  TRAP 27952 7 │ 
 │  │╷   │              │ 
 │  ││   ╰──────────────╯
 │ ─╯│              
 ╰───╯              
[0x00006d30]> 
```


[0x00006d30]> dr PC=0xdeadbeef
[0x00006d30]> e cmd.esil.trap =?E TRAP
[0x00006d30]> aes
INFO: [ESIL] Trap, trying to execute on non-executable memory
 ╭──╮    ╭──────────────╮
 │ _│    │              │ 
 │ O O  <  TRAP 27952 7 │ 
 │  │╭   │              │ 
 ││ ││   ╰──────────────╯
 │└─┘│              
 ╰───╯              
[0x00006d30]>

trufae:
[...]

PR #24772, opened 2025-11-02
test: allow r2r to pick an alternative r2 binary

sagittarius-a:

**Description**

As part of our work on #24769, I figured out that `r2r` is supposed to accept a custom `r2` binary, but the environment variable was not read.


trufae:
Ouch dunno when this was removed because if it was in the help message it was suposed to be implemented. And when i saw the pr i though that this thing was already done.. anyway thanks :)

sagittarius-a:
_Regression in the regression tests suite 😇_

PR #24791, opened 2025-11-05
Fix inconsistency in `pq` command help message

sagittarius-a:

**Description**

The code also mentions `pq0` in a cryptic _TODO_ comment. As the current changes only imply documentation, I did not change the logic itself.


trufae:
I assume pq0 discussion is aligned with the p8s pr, would you like to take a look and discuss about these subcommands so we can clarify it

PR #24792, opened 2025-11-05
Change the order where register aliases fetch their numeric value to prioritize r_reg_get over r_flag_get ##radare2

SverkerBerggren:

**Description**

When starting radare with the -i option to execute a script, so did the alias for registers not work correctly for math operations such as "? rbp". The reason, as I understand it, is that when a debugger is attached, so is a part of the setup to set all the registers as flags using the ".dr*" command, but that on startup just sets them all to 0x0. Later, when the token's numerical value is fetched, so does it first lookup flags rather than registers, which means it always gets the flag value. This makes it so that using "? rbp" in a -i script will always yield 0x0. 

The reason why it works normally otherwise as I understand it, is because during the regular command loop so does it always execute a certain number of commands after each user inputed command, and one of them is ".dr*" which makes the register flags always up to date.

The way to fix this issue would be either to make it so that the same "cleanup" commands are executed after each -i command in the startup script, or the suggested change that I made, which I find to be less intrusive. It does, on the other hand, change the way that "? rbp" is executed in almost all scenarios, but I do not feel that I have the knowledge to say if that is necessarily a bad thing. 

This pull request also includes a commit that was supposed to me made in this previous pull request but I did not push it before it was merged. #24767

### DIFF
--- a/libr/core/numvars.inc.c
+++ b/libr/core/numvars.inc.c
@@ -1007,14 +1007,6 @@ static ut64 num_callback(RNum *userptr, const char *str, bool *ok) {
 				...
 			}
 #endif
-			if ((flag = r_flag_get (core->flags, str))) {
-				ret = flag->addr;
-				if (ok) {
-					*ok = true;
-				}
-				return ret;
-			}
-
 			// check for reg alias
 			RRegItem *r = r_reg_get (core->dbg->reg, str, -1);
 			if (r) {
@@ -1022,6 +1014,14 @@ static ut64 num_callback(RNum *userptr, const char *str, bool *ok) {
 					*ok = true;
 				}
 				ret = r_reg_get_value (core->dbg->reg, r);
+				return ret;
+			}
+			
+			if ((flag = r_flag_get (core->flags, str))) {
+				ret = flag->addr;
+				if (ok) {
+					*ok = true;
+				}
 				return ret;
 			}
 		}


trufae:
Do not do merge commits. Always rebase

trufae:
There r 2 ways to fix that:
- use the $r:regname prefix
- Hook the rregget callback to the flagitems associated with registers

trufae:
The problem of the first is that it is different symtax, the problem of the second is that it will slowdown the thing because it will sync regs all the time unless we mark a dirty bit somewhere to avoid syncing the regs all the time

SverkerBerggren:
I tried to follow the instructions regarding rebasing but I am not a git wizard. Is there any action that needs to be taken to fix this branch now or?

I personally prefer the second approach you describe because I value that scripting works as similar to the interactive mode as possible, I think it is very hard for a new user to debug and understand the nuances in how the context of the invocation of the commands affect their output, but that is just my opinion.

Is r_reg_get and r_reg_get_value expensive operations compared to r_flag_get? The changes I have in this pull request resolved the issue for me, but do you think that this solution is undesirable because of performance?

I would otherwise be willing to test to implement the dirty bit solution, but is there a way to know which commands affect the state of the binary? Continuing the execution of the program would I guess always need to set the registers dirty, but just seeking to an adress would not.

trufae:
sorry rebase again please

trufae:
rebase

SverkerBerggren:
I have seen on some other pull requests that not all tests have been successful, but are there tests here that should pass unless my changes introduced some bugs?

trufae:
your pr introduces bugs:

```

[XX] /home/runner/work/radare2/radare2/test/db/cmd/cmd_pa asm/callflag
R2_NOPLUGINS=1 radare2 -escr.utf8=0 -escr.color=0 -escr.interactive=0 -NN -Qc 'e asm.arch=x86
f eax=33
pa call eax
f test=eax
pa call test
' -
-- stdout
@@ -1,2 +1,2 @@
 ffd0
-e81c000000
+e8fbffffff
```

this is clearly eax not being usable as a normal flag because its hooked by the rreg api so it fails and returns -1.
also your patch doesnt pass the sys/lint.sh linter with this error:

```
libr/core/numvars.inc.c:1019:			
```
^thats an empty line with trailing spaces

libr/core/numvars.inc.c
```diff
@@ -1007,21 +1007,21 @@ static ut64 num_callback(RNum *userptr, const char *str, bool *ok) {
 				...
 			}
 #endif
-			if ((flag = r_flag_get (core->flags, str))) {
-				ret = flag->addr;
+			// check for reg alias
+			RRegItem *r = r_reg_get (core->dbg->reg, str, -1);
```
trufae, line 1011: you can also use r_reg_getv

libr/socket/run.c
```diff
@@ -981,7 +981,7 @@ R_API bool r_run_config_env(RRunProfile *p) {
 				r_socket_free (fd);
 				return false;
 			}
-			r_socket_block_time(fd, true, 99999, 0);
+			r_socket_block_time(fd, true, 0, 0);
```
trufae, line 984:
```suggestion
			r_socket_block_time (fd, true, 0, 0);
```

SverkerBerggren:
Sure, I can try to fix those things. I am a bit unsure what you meant about the tests being broken. Did the pr not produce a bug, and the tests were instead not working as intended? Is the default behaviour how it is right now before the pr and the config variable you suggested should enable the change in this pr? Should I otherwise try to change the tests to what they should be?

libr/socket/run.c
trufae: This socket thing can be one pr. And the other flag reg thing in a separate one if u want.

If u add a config var to change the behaviour of num to pick values from rega instead of flags and disable it by default the tests will pass

trufae:
check my comment above. i think its worth merging this for 6.0.6, and then go try to enable this flagreg feature on top.

because flags are accessed from other places, not just via rnum, and the rflags api have a way to tie a flag value to a callback. so it should be tied this way in order to not make rnum slow by default. which is my initial concern on this feature. but as long as its optional we can maybe experiment with it and add the opt-in solution for people to try

trufae:
I commited half of the change in your name here: https://github.com/radareorg/radare2/pull/24861

i'll do the other part in a separate pr

trufae:
All done in https://github.com/radareorg/radare2/commit/4cf036c96517f6a3c63d44cd54807c180413d032

SverkerBerggren:
Aight cool, I need to be bit faster next time

PR #24850, opened 2025-11-17
Implement extended support for custom floating-point formats in the r…

0verflowme:
- Extended libr/util/cfloat.c/r_cfloat.h with the RCFloatValue helpers.
- Added the >64-bit code paths via r_cfloat_parse_ex/r_cfloat_write_ex.
- Added unt tests.
- Verified everything locally via make -C libr/util, make -C libr/reg, gcc … /tmp/test_reg, and LD_LIBRARY_PATH=… /tmp/test_reg

image

### DIFF
--- a/test/unit/test_reg.c
+++ b/test/unit/test_reg.c
@@ -5,19 +5,21 @@
 #include <string.h>
 #include "minunit.h"
 
-static void mu_assert_double_close(double actual, double expected, double eps, const char *message) {
+static int mu_assert_double_close(double actual, double expected, double eps, const char *message) {
 	double diff = fabs (actual - expected);
 	char buf[MU_BUF_SIZE];
 	snprintf (buf, sizeof (buf), "%s: expected %.16f, got %.16f (diff %.16f)", message, expected, actual, diff);
 	mu_assert (buf, diff <= eps);
+	return MU_PASSED;
 }
 
 #if !R2_NO_LONG_DOUBLE
-static void mu_assert_longdouble_close(long double actual, long double expected, long double eps, const char *message) {
+static int mu_assert_longdouble_close(long double actual, long double expected, long double eps, const char *message) {
 	long double diff = fabsl (actual - expected);
 	char buf[MU_BUF_SIZE];
 	snprintf (buf, sizeof (buf), "%s: expected %.19Lf, got %.19Lf (diff %.19Lf)", message, expected, actual, diff);
 	mu_assert (buf, diff <= eps);
+	return MU_PASSED;
 }
 #endif
 


libr/reg/double.c
```diff
@@ -3,111 +3,324 @@
...
 
-// long double = 128 bit
-R_API double r_reg_get_double(RReg *reg, RRegItem *item) {
-	R_RETURN_VAL_IF_FAIL (reg && item, 0.0f);
-	double vld = 0.0f;
-	double ret = 0.0f;
-	int off = BITS2BYTES (item->offset);
-	RRegSet *regset = &reg->regset[item->arena];
-	switch (item->size) {
+	if (success)
```
trufae: run clang-format-radare2 in this file, we dont do c++ bracket placings in r2land

0verflowme: updated

libr/esil/esil_ops.c
```diff
@@ -2596,7 +2596,24 @@ static bool esil_double_to_float(REsil *esil) {
 			ret = r_esil_pushnum (esil, f.u32);
 		} else if (s == 64) {
 			ret = r_esil_pushnum (esil, d.u64);
-		/* TODO handle 80 bit and 128 bit floats */
+		} else if (s == 80 || s == 96 || s == 128 || s == 256) {
+			// Handle >64-bit floats using cfloat API
+			RCFloatProfile profile;
```
trufae, line 2600: move this assginment into a separate function that returns the profile by taking the 's' bits as argument

libr/reg/double.c
```diff
@@ -3,111 +3,279 @@
 #include <r_reg.h>
 #include <r_util.h>
 
-// TODO: add support for 80bit floating point value
+// Helper to select appropriate float profile based on register size and endianness
+static RCFloatProfile get_float_profile(RReg *reg, RRegItem *item, bool *success) {
+	const bool be = (reg->endian & R_SYS_ENDIAN_BIG) == R_SYS_ENDIAN_BIG;
+	RCFloatProfile profile;
```
trufae, line 9: this profile = from_bits() helper function can be reused here too

trufae:
the unit tests code fails to build in windows. see the ci logs for that

libr/core/cmd_debug.inc.c
```diff
@@ -3083,7 +3083,11 @@ static void cmd_debug_reg(RCore *core, const char *str) {
 					sscanf (eq, "%Lf", &val);
```
trufae: i think that %Lf is not portable. so maybe better to use our own r_str_scanf imho

libr/reg/double.c
```diff
@@ -3,111 +3,279 @@
...
+		RCFloatValue value;
+		if (r_cfloat_parse_ex (src, size_bytes, profile, &value)) {
+			return r_cfloat_value_to_double (&value, profile);
+		}
+		return 0.0;
+	}
+
+	// For <=64-bit formats, use regular API
+	ut8 tmp[8] = { 0 };
+	if (size_bytes > (int)sizeof (tmp)) {
```
trufae: use R_MIN instead

libr/reg/double.c
```diff
@@ -3,111 +3,279 @@
...
+		size_bytes = sizeof (tmp);
+	}
+	memcpy (tmp, src, size_bytes);
```
trufae: like this: memcpy (tmp, src, R_MIN (size_bytes, sizeof (tmp));

libr/util/cfloat.c
```diff
@@ -206,3 +89,288 @@ R_API const RCFloatProfile *r_cfloat_profile_from_name(const char *name) {
...
+	if (profile->explicit_leading_bit && profile->mant_bits > 0) {
+		int leading_idx = profile->mant_bits - 1;
+		long double leading = value_get_bit (value, mant_pos + leading_idx)? 1.0L: 0.0L;
+		long double frac = fraction_from_bits (value, mant_pos, leading_idx);
+		mant_val = leading + frac;
+	} else {
+		long double frac = fraction_from_bits (value, mant_pos, profile->mant_bits);
+		mant_val = 1.0L + frac;
+	}
+	long double result = mant_val * ldexpl (1.0L, (int)exp - profile->bias);
```
trufae: is ldexpl portable? the original implementation was much simpler and portable imho

trufae (changes requested):
more comments here and there

libr/reg/double.c
trufae: actually if size_bytes is larger or smaller.. maybe its probably better not to try to work and just error

libr/util/cfloat.c
0verflowme: replaced with a portable loop that multiplies/divides by 2.0L based on the exponent delta.

libr/reg/double.c
0verflowme: updated

libr/reg/double.c
0verflowme: updated

libr/core/cmd_debug.inc.c
0verflowme: updated

0verflowme:
> the unit tests code fails to build in windows. see the ci logs for that

updated tests as well

PR #24854, opened 2025-11-17
Fix #14715: Validate pid argument in cmd_debug_continue function

### LINKED ISSUE #14715: Invalid dc parameter hangs r2
### Work environment

| OS/arch/bits (mandatory)                             | Arch Linux, x86, 64 bits
| Architecture/bits of the file (mandatory)            | x86/32, x86/64
| r2 -v full output, **not truncated** (mandatory)     | radare2 3.7.0-git 22437 @ linux-x86-64 git.3.6.0-300-g8fca9c95f commit: 8fca9c95f81d8c922723ac882fa46cc981a088f5 build: 2019-07-31__18:33:53

### Expected behavior
```
[0xf7f35120]> dc qwerty
error: no such process
[0xf7f35120]> 
```
### Actual behavior
```
[0xf7f35120]> dc qwerty
= attach 0 6969
PTRACE_CONT: No such process

```
and hangs. Ctrl-C and Ctrl-D do not affect its state.

[hello.zip](https://github.com/radare/radare2/files/3452798/hello.zip)

Compile with `gcc -o hello64 hello.c` or `gcc -m32 -o hello32 hello.c`

Run with `r2 -d hello32` and execute a command `dc <some_text>`

I know that `dc` expects a number (pid), so this is not critical, but still, it shouldn't hang the whole debugger...

0verflowme:

**Description**
image

### DIFF
--- a/libr/core/cmd_debug.inc.c
+++ b/libr/core/cmd_debug.inc.c
@@ -4926,22 +4926,26 @@ static int cmd_debug_continue(RCore *core, const char *input) {
 		}
 		break;
 	case ' ':
+	{
 		const char *pidstr = r_str_trim_head_ro (input + 2);
-		if (!pidstr || !*pidstr) {
+		if (R_STR_ISEMPTY (pidstr)) {
 			R_LOG_ERROR ("Missing pid argument");
 			break;
 		}
-		if (!r_num_is_valid_input (core->num, pidstr)) {
+		const char *err = NULL;
+		ut64 pidn = r_num_math_err (core->num, pidstr, &err);
+		if (err || r_num_failed (core->num)) {
 			R_LOG_ERROR ("Invalid pid argument: %s", pidstr);
 			break;
 		}
 		old_pid = core->dbg->pid;
-		pid = (int)r_num_math (core->num, pidstr);
+		pid = (int)pidn;
 		r_reg_arena_swap (core->dbg->reg, true);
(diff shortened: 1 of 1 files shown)


libr/core/cmd_debug.inc.c
```diff
@@ -4926,8 +4926,17 @@ static int cmd_debug_continue(RCore *core, const char *input) {
 		}
 		break;
 	case ' ':
+		const char *pidstr = r_str_trim_head_ro (input + 2);
+		if (!pidstr || !*pidstr) {
```
trufae: you can't define vars after ':' in the case. open braces, move that into a separtte function or use ":;"

trufae: use R_STR_ISEMPTY(pidstr) instead of !||!*

libr/core/cmd_debug.inc.c
```diff
@@ -4926,8 +4926,17 @@ static int cmd_debug_continue(RCore *core, const char *input) {
...
+			R_LOG_ERROR ("Missing pid argument");
+			break;
+		}
+		if (!r_num_is_valid_input (core->num, pidstr)) {
```
trufae: you can check if the math is correct or not by checking the err field. this function checks for valid number, not valid math. so tis not correct

libr/core/cmd_debug.inc.c
0verflowme: updated

PR #24872, opened 2025-11-19
Fix -H option to accept variable name as argument

0verflowme:

**Description**


trufae:
There should be a test for that

PR #24874, opened 2025-11-19
Improve UBIFS filesystem support ##fs

mikelloc:
Improve UBIFS filesystem support

- Add UBI container format support with LEB-to-PEB mapping
- Add optional LZO decompression

### DIFF
--- a/libr/fs/p/ubifs.mk
+++ b/libr/fs/p/ubifs.mk
@@ -3,7 +3,15 @@
 OBJ_UBIFS=fs_ubifs.o
 
 # Detect LZO support at build time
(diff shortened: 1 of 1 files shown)


trufae:
macos-arm64 is crosscompiled target. fails because the lzo lib is not available for arm64

trufae:
it's not about macos or arm64. its a crosscompilation problem . that thing will fail if you crosscompile to windows from linux or build for riscv or do ann uefi build. just disable the support for lzo by default unless explicitly defined via --with-lzo in the configure /meson

mikelloc:
Will rework on this in the future, for now it's not justified invest time trying to fix a pipeline problem when the code is ok.

